PR #6003, opened 2026-07-10
fix(video-editor): disable clip Split/Speed while busy instead of hiding them

### LINKED ISSUE #6004: fix(video-editor): disable clip Split/Speed while busy instead of hiding them
Auto-created from PR #6003.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6003
**Branch:** `fix/video-editor-clip-controls-busy-state`

hm21:
Closes #6004

## Description

The Split and Speed controls in the clip-editor toolbar were **removed** from the toolbar while a per-clip render (split / audio extraction) was in flight, which users found confusing — the control set visibly reshuffled mid-operation. This shows them **disabled** instead, and scopes the block to the affected clip so a render on a *different* clip no longer disables the current clip's actions.

Key changes:
- Track `splittingClipId` / `extractingAudioClipId` in `ClipEditorState` so the disable is per-clip (following the existing `reversingClipId` / `transformingClipId` pattern).
- Render Split/Speed disabled (not removed) via the shared `VideoEditorTimelineControls`.
- Don't gate **Done** during a render — leaving edit mode is safe because the extraction/split result is committed by a session-level listener in `VideoEditorScaffold`, so it survives these controls unmounting.
- Queue split & audio extraction on other clips (`droppable()` → `sequential()`) instead of silently dropping them. The two events now carry the target clip id (split also the cut position) so a queued op still targets the intended clip even if the selection changes while an earlier render runs.
- Re-read bloc state after the speed bottom sheet closes so a concurrent mute (e.g. an extraction on another clip finishing) isn't clobbered by a stale pre-`await` snapshot.

**Related Issue:** 

## Out of Scope

- True parallelism (running per-clip renders concurrently) — the queue is `sequential` (one native render at a time), matching the current single-render pipeline.
- Two known minor, undo-able edges are intentionally left: a rapid double-tap of Extract on a clip while another render is queued can enqueue a duplicate extraction; and a queued split on a clip *before* the current selection can shift the selection index by one. Fully closing these needs synchronous "pending" tracking (a `Set`-based state), deferred.

## Verification

- `flutter analyze` on the touched `lib`/`test` dirs — no issues.
- `flutter test` for the clip-editor bloc + both timeline-controls widget suites — all green (116 tests), including new regression tests asserting a queued split/extraction targets the clip named by the event, not the current selection.
- Not yet exercised in the running app — opening as a draft for manual verification.

## Type of Change

### DIFF
--- a/mobile/lib/blocs/video_editor/clip_editor/clip_editor_bloc.dart
+++ b/mobile/lib/blocs/video_editor/clip_editor/clip_editor_bloc.dart
@@ -116,7 +116,9 @@ class ClipEditorBloc extends Bloc<ClipEditorEvent, ClipEditorState> {
 
     // Split
     on<ClipEditorOriginalClipReplaced>(_onOriginalClipReplaced);
-    on<ClipEditorSplitRequested>(_onSplitRequested, transformer: droppable());
+    // sequential (not droppable) so a split requested on a different clip while
+    // one is rendering is queued and runs, rather than being silently dropped.
+    on<ClipEditorSplitRequested>(_onSplitRequested, transformer: sequential());
 
     // Trim
     on<ClipEditorTrimUpdated>(_onTrimUpdated, transformer: restartable());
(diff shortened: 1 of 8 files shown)


rabble (approved):
Reviewed the clip-bound queued split and audio-extraction flows, selection preservation, duplicate extraction guard, and stable disabled controls. Focused analysis is clean and all 119 targeted tests pass.\n\nReviewed by Codex on behalf of @rabble.

PR #6005, opened 2026-07-10
fix(video-editor): keep timeline thumbnails stable across split renders and undo/redo

### LINKED ISSUE #6006: fix(video-editor): keep timeline thumbnails stable across split renders and undo/redo
Auto-created from PR #6005.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6005
**Branch:** `fix/video-editor-split-thumbnail-stability`

hm21:
Closes #6006

## Description

Splitting a clip made the timeline strip visibly fall apart in three ways, all reported from device testing:

1. **Thumbnails reshuffled while the split rendered.** The split halves are seeded with the source clip's dense, correct frames — but as soon as the rendered segment files arrived, the first fresh extraction batch (6 frames) *wholesale-replaced* the dense seeds. Most slots lost their frame, fell back to the poster, and then refilled batch by batch — a multi-second reshuffle.
2. **Undo/redo flashed wrong thumbnails and re-extracted everything.** Removing a clip id deleted its strip (notifier + files) immediately. Undo brought the pre-split clip back to an empty notifier: every slot briefly showed the poster (the clip's first frame) while a full native re-extraction ran.
3. **The end half's frames all jumped left when its render landed.** The render rebases the clip from source time (`trimStart = splitPos`) to file time (`trimStart = 0`), which re-anchored the 48px thumbnail slot raster at the new file's zero — shifting every frame in the tile by up to one slot.

Key changes (one per defect, all in the thumbnail pipeline):

- **Merge instead of replace** (`ClipThumbnailManager._mergeCarriedFrames`): carried frames (seeds, pre-restart frames) survive as gap-fillers until a fresh frame lands within half the final frame spacing; an `onDone` sweep prunes the residue (long clips cap the generator's frame count, so the final spacing can exceed that distance). Files are only deleted once nothing references them.
- **Retired-strip cache**: removed clip ids park their strip (frames + video path + completeness) in a bounded FIFO (8 strips) instead of being deleted. When the id returns with the same file — undo re-adds the pre-split clip, redo the halves — the strip is restored instantly; complete strips skip re-extraction entirely, partial ones start a gap-filling subscription. Seeded-but-unrendered halves are deliberately *not* retired (their timestamps carry an unapplied rebase; restoring them would misalign frames).
- **Recording-anchored slot raster**: new `DivineVideoClip.sourceStartOffset` field records where a clip's file starts inside the original recording. The split service sets it on the rendered end half (`source offset + absoluteSplitPos`, so nested splits accumulate) and the bloc carries it through `onClipRendered`. The tile shifts its slot windows *and* its content translation by `sourceStartOffset mod slotSpan`, which makes each frame's on-screen position mathematically invariant across the preview → rendered swap. Slot windows also move from `duration/count` pitch to a fixed 48px pitch, fixing a pre-existing drift that pushed frames up to ~half a slot right of their true position toward a tile's end.

**Why a model field** rather than view-layer bookkeeping: the offset must survive undo/redo and draft reload (it round-trips through the clip's JSON), accumulate across nested splits, and be inherited by duplicate/trim/transform via `copyWith` — the strip widget's state dies on unmount and can't provide any of that. Old drafts without the key default to zero (unchanged behavior).

## Out of Scope

- The retired cache is capped at 8 strips; undoing further back than that re-extracts (cold start, as before).
- Reversed clips keep a file-anchored raster — their content mirrors anyway, so there is no continuity to preserve.

## Verification

- `flutter analyze` over all changed `lib`/`test` dirs — no issues.
- Full timeline-editor widget suite (364 tests) + manager/model/bloc suites green. New regression tests: batch-merge gap-filling, instant undo/redo restore (complete + partial + different-file + FIFO eviction), `sourceStartOffset` JSON round-trip, offset propagation through the split render, and a widget test asserting pixel-identical frame positions across the preview → rendered geometry swap (red without the fix).
- Device-tested on iOS: undo/redo now restores thumbnails instantly; split no longer reshuffles the strip.

## Type of Change

### DIFF
--- a/mobile/lib/services/video_editor/clip_thumbnail_manager.dart
+++ b/mobile/lib/services/video_editor/clip_thumbnail_manager.dart
@@ -53,6 +53,17 @@ class ClipThumbnailManager {
   // zero-based), the same shift is applied here so the frames stay
   // aligned until fresh thumbnails replace them.
   final Map<String, Duration> _pendingRebase = {};
+  // IDs whose subscription has delivered its final batch — the strip is at
+  // full density and only a source-file change warrants re-extraction.
+  final Set<String> _complete = {};
+  // Strips of recently removed clips, kept (files included) for an instant
+  // restore when the same id returns — undo re-adds the pre-split clip,
+  // redo re-adds the split halves. Without this the returning clip starts
+  // with an empty notifier: every slot falls back to the poster frame and
+  // the whole strip re-extracts from scratch. Insertion-ordered for FIFO
+  // eviction.
+  final Map<String, _RetiredStrip> _retired = {};
+  static const _maxRetiredStrips = 8;
 
   /// When true, newly started subscriptions begin paused and existing ones are
   /// held. See [pauseAll].
@@ -76,31 +87,74 @@ class ClipThumbnailManager {
   }) {
     final currentIds = clips.map((c) => c.id).toSet();
 
-    // Remove stale entries.
+    // Retire stale entries — the strip is kept for an instant restore if
+    // the id returns (undo re-adds the pre-split clip, redo the halves).
     final staleIds = _notifiers.keys
         .where((id) => !currentIds.contains(id))
         .toList();
     for (final id in staleIds) {
       _subscriptions.remove(id)?.cancel();
-      _videoPaths.remove(id);
-      _seeded.remove(id);
+      final videoPath = _videoPaths.remove(id);
+      final wasSeeded = _seeded.remove(id);
       _pendingRebase.remove(id);
+      final wasComplete = _complete.remove(id);
       final notifier = _notifiers.remove(id);
-      if (notifier != null) {
-        // Files borrowed by seeded clips (split halves) stay alive; only
-        // files no other notifier references are deleted.
-        _deleteUnreferencedFiles(notifier.value);
-        notifier.dispose();
+      if (notifier == null) continue;
+      final frames = notifier.value;
+      notifier.dispose();
+      // Seeded strips are not retired: their frames are borrowed from the
+      // (retired) source strip and carry a pending rebase that is lost
+      // here — restoring them later would misalign timestamps. The
+      // unreferenced check keeps the borrowed files alive through the
+      // source's retired entry.
+      if (frames.isEmpty || videoPath == null || wasSeeded) {
+        _deleteUnreferencedFiles(frames);
+        continue;
       }
+      _retire(
+        id,
+        _RetiredStrip(
+          videoPath: videoPath,
+          frames: frames,
+          complete: wasComplete,
+        ),
+      );
     }
 
     // Ensure notifiers exist and start (or restart) loading.
     for (final clip in clips) {
-      _notifiers.putIfAbsent(clip.id, () => ValueNotifier(const []));
+      final notifier = _notifiers.putIfAbsent(
+        clip.id,
+        () => ValueNotifier(const []),
+      );
       final newPath = clip.video.file?.path;
-      final currentPath = _videoPaths[clip.id];
       final hasSubscription = _subscriptions.containsKey(clip.id);
       final isSeeded = _seeded.contains(clip.id);
+
+      // Instant restore for a returning clip id (undo/redo): reuse the
+      // retired strip instead of flashing posters and re-extracting.
+      if (!hasSubscription && !isSeeded && notifier.value.isEmpty) {
+        final retired = _retired.remove(clip.id);
+        if (retired != null) {
+          if (newPath != null && newPath == retired.videoPath) {
+            notifier.value = retired.frames;
+            _videoPaths[clip.id] = newPath;
+            if (retired.complete) {
+              _complete.add(clip.id);
+              continue;
+            }
+            // Partial strip (removed mid-extraction): fall through so a
+            // fresh subscription fills the gaps — the restored frames
+            // stay visible as gap-fillers via the batch merge.
+          } else {
+            // Same id but a different source file — the frames would show
+            // stale content, so drop them (borrowed files stay protected).
+            _deleteUnreferencedFiles(retired.frames);
+          }
+        }
+      }
+
+      final currentPath = _videoPaths[clip.id];
 
       if (!hasSubscription) {
         if (isSeeded) {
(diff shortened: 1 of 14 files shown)


mobile/lib/services/video_editor/clip_thumbnail_manager.dart
```diff
@@ -248,31 +321,110 @@ class ClipThumbnailManager {
...
+            final currentPaths = {for (final thumb in merged) thumb.path};
+            _deleteUnreferencedFiles([
+              for (final thumb in replaced)
+                if (!currentPaths.contains(thumb.path)) thumb,
+            ]);
+          },
+          onDone: () {
+            final notifier = _notifiers[clip.id];
+            if (notifier == null || latestFresh.isEmpty) return;
+            _complete.add(clip.id);
```
realmeylisdev, line 365: **should-fix (error path only).** `generateStripThumbnails` catches a native `getThumbnails` failure mid-stream and `break`s (`video_thumbnail_service.dart` L651-657), so the `async*` stream closes **normally** — `onDone` can't distinguish a clean finish from an error truncation. On a decoder hiccup at batch 3 of N, `latestFresh` is partial-but-non-empty, so this block marks the clip `_complete` and hard-replaces the notifier with the partial set, dropping (and deleting) the seeded/carried gap-fillers that were showing correct content — i.e. the exact "strip falls apart" symptom this PR fixes, now with no retry (a later remove/undo even re-parks it `complete: true`). The all-batches-fail path (`latestFresh.isEmpty`, L354) correctly keeps the seeds — that asymmetry is the tell. Consider signalling clean-vs-truncated completion from the generator (or gating `_complete` + the seed-drop on reaching the expected frame count).

mobile/lib/services/video_editor/video_editor_split_service.dart
```diff
@@ -104,9 +104,14 @@ class VideoEditorSplitService {
...
       processingCompleter: Completer<bool>(),
     );
+    // The rendered end file starts at the split point, so its zero-based
+    // timeline sits absoluteSplitPos into the source recording. Recording
+    // that shift keeps the timeline thumbnail raster anchored to the
+    // original recording (no visible frame shift when the render lands).
     final renderedEndClip = previewEndClip.copyWith(
       duration: sourceClip.duration - absoluteSplitPos,
       trimStart: Duration.zero,
+      sourceStartOffset: sourceClip.sourceStartOffset + absoluteSplitPos,
```
realmeylisdev, line 114: **should-fix (coverage).** This line — including the `sourceClip.sourceStartOffset +` accumulation for nested splits — has no direct test. `video_editor_split_service_test.dart` runs the real service and captures `onClipRendered` but never asserts `sourceStartOffset`; the bloc test asserts it through a fake (`_fakeSplitClipThenRenderEnd`) that re-implements this same formula; the widget tests set the offset directly on a `copyWith` clip. Reverting this to `sourceStartOffset: absoluteSplitPos` (dropping accumulation) keeps the whole suite green. One capture+assert in the existing split-service test (`== splitPosition` for a zero-offset source, `== base + splitPosition` for a pre-offset source) closes both the first-split and nested-split gaps.

realmeylisdev (commented):
## Formal review — timeline thumbnail stability

Reviewed the full diff plus the surrounding thumbnail pipeline, split/reverse/undo paths, and draft serialization. Verified locally: `flutter analyze` clean on all 10 changed files, the four changed suites pass (incl. 151 strip tests), CI green.

**The core fix is sound.** The recording-anchored raster is mathematically correct — I worked the preview→rendered on-screen-X invariance algebraically for the general case (arbitrary split point, non-zero parent offset for nested splits, `playbackSpeed ≠ 1`) and it holds exactly; the pixel-level regression test pins it. `sourceStartOffset` is display-only — it threads split → bloc → `copyWith`/JSON and never reaches export or any Nostr/imeta path (this is a pure client-side editor change, no backend/protocol surface). The merge/retire/rebase file lifecycle protects borrowed and shared frames (no leaks or premature deletes found), and the retired-cache keying on `(id, videoPath)` restores undo/redo correctly.

Two non-blocking should-fixes (inline) + one cosmetic nit below. **Nothing blocks merge.**

### Nit — reverse of a split-reversed clip (untouched code, noted here)
`reversed → split → un-reverse` via the render path (`clip_editor_bloc.dart` `_onClipReverseRequested`, ~L895): splitting a reversed clip clears both cache paths (per the L879-883 comment), so un-reversing the half falls to the render branch, which produces a fresh zero-based mirror file but `copyWith` keeps the non-zero `sourceStartOffset`. With `reversed=false`, `_rasterAnchorOffset` then phase-shifts that strip by `offset mod 48px` against a file with no recording continuity — a static ≤47px filmstrip offset (content correct, alignment off). Consider zeroing `sourceStartOffset` whenever reverse renders a fresh mirror file.

mobile/lib/services/video_editor/clip_thumbnail_manager.dart
hm21: Fixed in 38ab1aa4e. The generator now rethrows after logging instead of breaking, so a mid-stream failure surfaces as a stream error, and the manager gates `_complete` + the carried-frame sweep on a clean close — truncated strips keep their gap-fillers and a later retire/restore starts a gap-filling re-extraction instead of re-parking `complete: true`. The cover screen (the other stream consumer) got a keep-partial `onError`. Regression tests added at both layers (manager: gap-fillers survive + restore re-extracts; service: batches then error).

mobile/lib/services/video_editor/video_editor_split_service.dart
hm21: Added in 9a9bd2af4 — the split-service test now captures the rendered halves from the real service and asserts the offset for both a zero-offset source (`== splitPosition`) and a pre-offset source with a non-zero trimStart (`base + trimStart + splitPosition`). Verified by mutation: reverting to `sourceStartOffset: absoluteSplitPos` turns the nested-split test red.

hm21:
@realmeylisdev All three findings addressed:

- **Truncated-stream should-fix** → 38ab1aa4e (generator rethrows, manager gates completion + seed-drop on clean close)
- **sourceStartOffset coverage** → 9a9bd2af4 (real-service assertions for first and nested splits)
- **Reverse nit** → d3fff01d6, with one deviation from the suggestion: the offset is zeroed only on the *un-reverse* render, not whenever reverse renders a fresh mirror file. Zeroing on the forward→reversed render would drop the offset that the cached un-reverse branch later restores together with `forwardVideoPath` — that round trip would come back mis-anchored. Both directions are pinned by bloc tests.

rabble (approved):
Reviewed the carried-frame merge, truncated-stream handling, retired-strip lifecycle, source-offset persistence/accumulation, and resolved regression coverage. CI is green; no blocking findings. (Review by Codex on behalf of @rabble)

PR #5961, opened 2026-07-08
fix(comments): hide stranded keyboard on sheet dismiss after background/resume

realmeylisdev:
## Description

The software keyboard stayed visible over the video feed after closing the comments sheet, when the app had been backgrounded and resumed while the sheet was open (in-app bug report, build 1.0.16+791, iOS 26.5.1 — the build already contained the #5616 fix, verified by dating the build via the `ingest_queue_depth` telemetry from #5906).

`UnfocusOnSheetDismiss` reacted only to `AnimationStatus.reverse` and only called `primaryFocus?.unfocus()`. Traced against the Flutter 3.44 framework and iOS engine sources, that leaves three gaps:

1. **Platform-initiated text-input teardown during background/resume (this report's repro).** iOS resigns `FlutterTextInputView` during a background/resume cycle; the engine sends `TextInputClient.onConnectionClosed`; the framework nulls its side of the connection **without any outbound `TextInput.hide`** (`connectionClosedReceived`) and unfocuses. The engine keeps the stale first responder, so on resume iOS re-presents the keyboard while the framework believes nothing is focused. At dismissal, `unfocus()` produces zero text-input channel traffic and the keyboard is stranded with no recovery affordance on the feed.
2. **Chrome drag to the route controller's 0.0 clamp.** Status goes `completed→forward→dismissed` — never `reverse` (`_direction` is still forward, both fling branches in `material/bottom_sheet.dart` are guarded by `value > 0.0`, and the follow-up pop's `reverse()` takes the zero-duration branch).
3. **Route removal without a pop** (navigator page-stack rebuild): no status change fires at all.

The fix hardens the widget on all three: also react to `AnimationStatus.dismissed`, send an explicit `TextInput.hide` alongside `unfocus()` (the engine's `hideTextInput` resigns the first responder regardless of framework connection state and is a no-op otherwise), and dismiss from `dispose()` as a fallback.

Each vector is pinned by a regression test that reproduces the split-brain in-harness via `tester.testTextInput.closeConnection()` (the exact message the engine sends), asserting the keyboard stays stranded pre-dismiss and is hidden post-dismiss. All three new tests fail on the previous implementation and pass with the fix; the two pre-existing tests pass in both states.

**Related Issue:** Closes #5959

## Out of Scope

- On-device manual verification of the exact iOS 26.5.1 trigger that resigns the first responder during backgrounding (scene snapshotting / keyboard session churn) — the framework-side split-brain state is fully reproduced in the widget tests regardless of the trigger.
[...]

Issue #5959, opened 2026-07-08
fix(comments): keyboard stranded over feed after closing comments sheet following background/resume


realmeylisdev:
### Summary

The software keyboard stays visible after closing the comments sheet and sits on top of the video feed. Reported in-app on 2026-07-07 (build 1.0.16+791, iOS 26.5.1); user log export `openvine_full_logs_2026-07-07T19-36-34.051123.txt`.

This is the same user-visible symptom as #5604, but the build already contains that fix (`UnfocusOnSheetDismiss`, #5616) — verified via the `ingest_queue_depth` memory-telemetry format in the log, which only shipped in #5906 on 2026-07-07. The existing fix has a gap.

### Timeline from the log

1. `19:26:30` comments sheet opened, comment posted (`sign_event` at 19:26:32)
2. `19:32:56`–`19:33:16` app backgrounded and resumed **twice while the sheet stayed open with the keyboard up**
3. `19:34:09` `BottomSheet closed`
4. `19:34:39`+ user swipes the feed with the keyboard stranded on top; bug report filed at 19:36

### Root cause

`UnfocusOnSheetDismiss` only reacts to `AnimationStatus.reverse` and only calls `FocusManager.instance.primaryFocus?.unfocus()`. Verified against the Flutter 3.44 framework and iOS engine sources, that leaves three gaps — the first one matches this log:

1. **Platform-initiated text-input teardown during background/resume (this repro).** When iOS resigns `FlutterTextInputView` during a background/resume cycle, the engine sends `TextInputClient.onConnectionClosed`; the framework nulls its side of the connection **without sending `TextInput.clearClient`/`TextInput.hide`** (`text_input.dart` `connectionClosedReceived`) and unfocuses the field. The engine keeps `_textInputClient != 0`, so on resume iOS re-presents the keyboard for the stale first responder while the framework believes nothing is focused. At sheet dismissal, `primaryFocus?.unfocus()` produces zero text-input channel traffic — the keyboard is unrecoverably stranded (no tap-to-unfocus on the feed either, and it would no-op anyway).
2. **Chrome drag to the 0.0 clamp.** A single continuous drag on the sheet chrome rides the route controller to `0.0`: status goes `completed→forward→dismissed` (never `reverse`, since `_direction` is still forward and both fling branches in `material/bottom_sheet.dart` are guarded by `value > 0.0`); the follow-up pop's `reverse()` takes the zero-duration branch. The listener never fires.
[...]

PR #6008, opened 2026-07-10
fix(comments): clear the stale engine text-input client on sheet dismiss

### LINKED ISSUE #6007: fix(comments): keyboard stranded over home feed after in-foreground comment post + sheet close (iOS 26.5, build 792)
### Summary

The software keyboard is stranded over the home feed again — same user-visible symptom as #5604 and #5959 — reported on Discord from build 1.0.16+792, iOS 26.5.1 (log export `openvine_full_logs_2026-07-09T17-03-32.924623.txt`, screenshot at device-time 5:02 PM). Unlike #5959, the comments sheet in this incident lived **entirely in the foreground**: there is no background/resume between the sheet opening and closing, and the composer demonstrably worked (the user typed and posted a comment seconds before dismissal). The stranded keyboard shows the empty-text predictions bar ("i / the / i'm"), i.e. an active, empty input session.

### Timeline from the log

1. `16:39:00` comments sheet (A) opened; `16:39:53` app backgrounded **with the sheet open** for ~20 minutes; `17:00:48` resumed; `17:00:55` sheet A closed — the exact #5959 shape
2. `17:02:23` comments sheet (B) opened on the home feed (video had no comments)
3. `17:02:45` comment signed via Keycast (bare `sign_event`, the comments-repository signing path)
4. `17:02:47.786` `BottomSheet closed` — **no lifecycle event at any point during sheet B's life**
5. Screenshot (~17:02:48–56): keyboard fully presented over the feed, no sheet, no text field
6. `17:02:56` user taps Share with the keyboard still stranded

### Root cause

Verified against the pinned Flutter 3.44.0 engine (`FlutterTextInputPlugin.mm`, framework rev `559ffa3f75`):

- `TextInput.hide` → `hideTextInput` **resigns the first responder but never clears `_textInputClient`** (it removes the input view only when a prior `clearClient` set `_pendingInputViewRemoval`).
- `TextInput.clearClient` → `clearTextInputClient` zeroes the client and removes the view (deferred to the next `hide` while still first responder), but **never resigns**.
- `canBecomeFirstResponder` returns `_textInputClient != 0` — the engine's only barrier against iOS re-presenting a keyboard on its own.

`UnfocusOnSheetDismiss` (#5961) sends **only `TextInput.hide`**. That is exactly right for making the keyboard drop, but in the #5959 split-brain state (framework side of the connection already nulled by `onConnectionClosed`, so `unfocus()` produces no channel traffic and the framework never sends its own `clearClient`) the dismissal leaves a **durable, in-hierarchy `FlutterTextInputView` with a stale client ≠ 0 and `canBecomeFirstResponder == YES`**. The keyboard hides, but the session survives engine-side until the next `setClient` — and iOS 26.5 has already been observed re-presenting the keyboard for exactly this stale state in #5959. Sheet A's 20-minute background cycle in this session is a live #5959-class contamination source; the foreground re-present trigger is OS-side and was not directly observable in the log.

An exhaustive audit found no in-app path that could re-present a keyboard on the home feed: the only `SystemChannels.textInput` sender in the entire app is the #5961 hide itself; there is no `TextInput.show`/`requestKeyboard` anywhere; the home feed has no text field; inactive shell branches are `ExcludeFocus`'d; the composer's rollback path never re-enters reply/edit mode.

Note on "just unfocus after send": the keyboard's own send key already unfocuses via `EditableText._finalizeEditing`, and a healthy-connection teardown (green in-sheet send button, then dismiss) provably ends clean (`clearClient` + `hide`, client 0, view removed). An extra unfocus on send cannot clear a stale engine client — no focus change produces channel traffic once the framework side of the connection is gone.

### Fix

Harden `UnfocusOnSheetDismiss._dismissKeyboard()` to send `TextInput.clearClient` before `TextInput.hide` — the engine's own reset order (`resetWithResult:` calls `clearTextInputClient` then `hideTextInput`). After every dismissal the engine then ends at client 0, first responder resigned, input view removed from the hierarchy — a state UIKit cannot re-present (`canBecomeFirstResponder == NO`, no view in the window). Harmless in the healthy case: the framework's own microtask-deferred `clearClient`/scheduled hide become no-ops, and the resign delegate reports client 0, which the framework's client-id guard drops.

### Open verification item

Whether build 792 actually contains #5961 is not determinable from the repo or the log (release iOS builds are manual Codemagic runs; TestFlight build number = `PROJECT_BUILD_NUMBER + 142`, so 792 = Codemagic build **650**). Someone with Codemagic access should confirm build 650's commit. Either answer leaves this fix necessary: without #5961 the dismissal sends no channel traffic at all in the split-brain state; with #5961 it hides but leaves the re-presentable stale client.

realmeylisdev:
## Description

The keyboard was reported stranded over the home feed again on build 1.0.16+792 / iOS 26.5.1 — after posting a comment and closing the comments sheet, with the sheet's whole life spent in the foreground (no background/resume, unlike #5959).

Root cause, verified against the pinned Flutter 3.44.0 engine (`FlutterTextInputPlugin.mm`):

- `TextInput.hide` (`hideTextInput`) resigns the first responder but **never zeroes `_textInputClient`**.
- `TextInput.clearClient` (`clearTextInputClient`) zeroes the client and removes the input view, but never resigns.
- `canBecomeFirstResponder` returns `_textInputClient != 0` — the engine's only barrier against iOS re-presenting a keyboard by itself.

`UnfocusOnSheetDismiss` (#5961) sends only `TextInput.hide`. In the #5959 split-brain state (framework side of the connection nulled by `onConnectionClosed`, so `unfocus()` produces no channel traffic and the framework never sends its own `clearClient`), the dismissal hides the keyboard but leaves a durable, in-hierarchy `FlutterTextInputView` with a stale client and `canBecomeFirstResponder == YES` — a session iOS 26 re-presents on its own (observed on resume in #5959; stranded over the feed with no resume in the new report). An exhaustive audit found no in-app path that can summon a keyboard on the home feed, so the re-present is OS-initiated; the fix removes the state it needs.

The fix sends `TextInput.clearClient` before the `TextInput.hide` — the engine's own reset order (`resetWithResult:` calls `clearTextInputClient` then `hideTextInput`). Every dismissal now ends with client 0, first responder resigned, and the input view removed from the hierarchy: nothing left for UIKit to re-present. In the healthy case both raw messages are no-ops racing the framework's identical `clearClient` + scheduled hide, and the hide's resign delegate reports client 0, which the framework's client-id guard drops.

The new regression test pins the channel order (`clearClient` before `hide`) in the split-brain state; it fails against the previous implementation (verified by stashing the lib change) and passes with the fix. The remaining five #5961 tests are unchanged and stay green.

Why not "unfocus after tapping send": the keyboard's send key already unfocuses via `EditableText._finalizeEditing`, a healthy-connection teardown already ends clean, and no focus change produces channel traffic once the framework side of the connection is gone — so an extra unfocus cannot clear a stale engine client.

**Related Issue:** Fixes #6007

## Out of Scope

The share, find-people, new-message, and clip-title sheets have focused/autofocused fields but no `UnfocusOnSheetDismiss` guard at all; extending the guard to them is independent of this report and will be proposed separately. Confirming whether TestFlight 792 contains #5961 (Codemagic build 650) is an administrative check tracked in #6007.

## Verification

- `flutter test test/screens/comments/widgets/unfocus_on_sheet_dismiss_test.dart` — 8/8 pass
- New test verified red against the pre-fix implementation (stash lib change → fails, restore → passes)
- `flutter test test/screens/comments/` — 103/103 pass
- `dart format` — no changes

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/screens/comments/widgets/unfocus_on_sheet_dismiss.dart
+++ b/mobile/lib/screens/comments/widgets/unfocus_on_sheet_dismiss.dart
@@ -5,6 +5,7 @@ import 'dart:async';
 
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:unified_logger/unified_logger.dart';
 
 /// Unfocuses the active text field and hides the software keyboard the moment
 /// the enclosing [ModalRoute] begins its dismiss transition.
@@ -84,16 +85,30 @@ class _UnfocusOnSheetDismissState extends State<UnfocusOnSheetDismiss> {
     // entirely — see the class doc.
     if (status == AnimationStatus.reverse ||
         status == AnimationStatus.dismissed) {
-      _dismissKeyboard();
+      _dismissKeyboard(trigger: status.name);
     }
   }
 
-  void _dismissKeyboard() {
+  void _dismissKeyboard({required String trigger}) {
     // A single teardown can surface more than one signal (a pop emits reverse
     // then dismissed); dismiss exactly once so we don't re-send TextInput.hide.
     if (_dismissed) return;
     _dismissed = true;
-    FocusManager.instance.primaryFocus?.unfocus();
+    final primaryFocus = FocusManager.instance.primaryFocus;
+    // Field forensics for the stranded-keyboard reports (#5959, #6007): the
+    // trigger names the dismissal path, and the focus shape distinguishes a
+    // healthy teardown (a focused field node) from the split-brain state
+    // (focus already fell back to a scope, so unfocus() will be silent).
+    Log.info(
+      'Keyboard teardown (trigger=$trigger, focus=${switch (primaryFocus) {
+        null => 'none',
+        FocusScopeNode() => 'scope',
+        FocusNode() => 'node',
+      }}): unfocus + clearClient + hide',
+      name: 'UnfocusOnSheetDismiss',
+      category: LogCategory.ui,
+    );
+    primaryFocus?.unfocus();
     // unfocus() sends nothing over the text-input channel when the framework
     // side of the connection is already gone (platform-initiated teardown
     // during background/resume — see the class doc). clearClient zeroes the
@@ -116,7 +131,7 @@ class _UnfocusOnSheetDismissState extends State<UnfocusOnSheetDismiss> {
     // existed so a generic non-sheet host that never stranded a keyboard
     // doesn't yank global focus / hide another route's keyboard on disposal.
     if (!_dismissed && _animation != null) {
-      _dismissKeyboard();
+      _dismissKeyboard(trigger: 'dispose');
     }
     super.dispose();
   }
--- a/mobile/lib/screens/comments/widgets/unfocus_on_sheet_dismiss.dart
+++ b/mobile/lib/screens/comments/widgets/unfocus_on_sheet_dismiss.dart
@@ -31,6 +31,20 @@ import 'package:flutter/services.dart';
 /// an explicit `TextInput.hide` — which resigns the first responder
 /// engine-side regardless of framework connection state — is sent as well.
 ///
+/// `TextInput.hide` alone is still not enough (#6007): the engine's
+/// `hideTextInput` resigns the first responder but never zeroes
+/// `_textInputClient`, and in the split-brain state the framework never sends
+/// its own `TextInput.clearClient` either. The dismissal then leaves an
+/// in-hierarchy `FlutterTextInputView` whose `canBecomeFirstResponder` is
+/// still `YES` — a session iOS 26 re-presents on its own (observed in #5959
+/// on resume, and stranded over the feed in #6007 with no resume at all). An
+/// explicit `TextInput.clearClient` is therefore sent before the hide — the
+/// engine's own reset order — so every dismissal ends with client 0, first
+/// responder resigned, and the input view removed: nothing left for UIKit to
+/// re-present. In the healthy case both raw messages are no-ops racing the
+/// framework's identical `clearClient` + scheduled hide, and the resign
+/// delegate reports client 0, which the framework's client-id guard drops.
+///
 /// The [dispose] fallback runs only when no route-status change already
 /// dismissed the keyboard and only when there was an enclosing modal route, so
 /// it is scoped to a sheet actually tearing down rather than firing on every
@@ -82,8 +96,14 @@ class _UnfocusOnSheetDismissState extends State<UnfocusOnSheetDismiss> {
     FocusManager.instance.primaryFocus?.unfocus();
     // unfocus() sends nothing over the text-input channel when the framework
     // side of the connection is already gone (platform-initiated teardown
-    // during background/resume — see the class doc). Hide explicitly; this is
-    // a harmless no-op when the keyboard is already down.
+    // during background/resume — see the class doc). clearClient zeroes the
+    // engine's stale client id so the orphaned FlutterTextInputView stops
+    // being re-presentable (#6007); hide then resigns the first responder and
+    // removes the view. Both are harmless no-ops when the keyboard is already
+    // down or the framework connection is healthy.
+    unawaited(
+      SystemChannels.textInput.invokeMethod<void>('TextInput.clearClient'),
+    );
     unawaited(SystemChannels.textInput.invokeMethod<void>('TextInput.hide'));
   }
 
--- a/mobile/test/screens/comments/widgets/unfocus_on_sheet_dismiss_test.dart
+++ b/mobile/test/screens/comments/widgets/unfocus_on_sheet_dismiss_test.dart
@@ -113,6 +113,39 @@ void main() {
     );
 
     testWidgets(
+      'clears the engine-side client on dismiss after the platform closed '
+      'the connection during background/resume',
+      (tester) async {
+        final focusNode = await pumpHostAndOpenSheet(tester);
+        await severConnectionPlatformSide(tester, focusNode);
+
+        // In the split-brain state the framework never sends its own
+        // TextInput.clearClient, and TextInput.hide alone leaves the engine
+        // with a stale client id whose in-hierarchy view iOS 26 can
+        // re-present on its own (#6007). The guard must clear the client
+        // explicitly, and must do so before the hide so the engine removes
+        // the input view during the hide's resign (the engine's own reset
+        // order).
+        tester.testTextInput.log.clear();
+        Navigator.of(tester.element(find.byType(UnfocusOnSheetDismiss))).pop();
+        await tester.pumpAndSettle();
+
+        final teardownCalls = tester.testTextInput.log
+            .map((call) => call.method)
+            .where(
+              (method) =>
+                  method == 'TextInput.clearClient' ||
+                  method == 'TextInput.hide',
+            )
+            .toList();
+        expect(
+          teardownCalls,
+          equals(['TextInput.clearClient', 'TextInput.hide']),
+        );
+      },
+    );
+
+    testWidgets(
       'hides the stranded keyboard when a chrome drag rides the route '
       'controller to its 0.0 clamp without ever reversing',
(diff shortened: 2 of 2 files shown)


realmeylisdev:
## Manual test plan — physical iPhone required

**Why a real device:** the bug lives in the interaction between the Flutter engine and UIKit's first-responder lifecycle. The Simulator does not faithfully reproduce iOS resigning the first responder on backgrounding (the event that creates the split-brain state) nor iOS 26's re-presentation of a stale keyboard session, so a Simulator pass proves nothing here. Primary target: iPhone on **iOS 26.5.x** (the reporter's OS). A sanity pass on any older iOS (17/18) for Scenarios 2 and 4 is a useful bonus.

**Setup:** build this branch onto the device (`flutter run --release` from `mobile/`, or profile mode), signed in with an account that can comment. Release mode matches the report; debug is acceptable for behavior.

**Failure signature to watch for (all scenarios):** a keyboard sitting over the home feed with no visible text field — usually showing the empty predictions bar ("i / the / i'm"). Taps pass through to the feed underneath and nothing dismisses it.

---

### Scenario 1 — split-brain replay (primary regression, the #5959 shape)

This is the scenario that *creates* the stale engine client the fix now clears.

1. On the home feed, tap **Reply** on any video to open the comments sheet.
2. Tap the composer field so the keyboard is up.
3. Background the app (swipe to Home screen) with the sheet + keyboard up. Wait ~1 minute (locking the phone also works).
4. Return to the app. On iOS 26.5 the keyboard typically re-presents over the sheet on its own — that's the split-brain state and is expected on this branch too (creating it is an iOS behavior, not what this PR changes).
5. Close the comments sheet (drag down or tap the scrim).
6. **Expected:** the keyboard dismisses with the sheet and **stays** dismissed. Keep watching for ~30 seconds while swiping a few videos, then open and close the comments sheet once more, then background/resume the app once. The keyboard must never appear on its own at any point.

Pre-fix, step 6 hid the keyboard but left a re-presentable session behind — the strand could appear immediately or minutes later. The re-presentation moment is OS-timing dependent, so give the post-dismissal observation window real time and repeat the whole scenario 2–3 times.

### Scenario 2 — the #6007 report shape (in-foreground post + close)

1. Fresh app launch. Open the comments sheet on a video (ideally one with no comments, matching the report).
2. Tap the field, type a comment, and post it with the **green in-sheet send button** (not the keyboard's send key — the green button intentionally keeps the keyboard up).
3. Within a couple of seconds, close the sheet while the keyboard is still up. Do a run for each dismissal vector:
   - drag the sheet's header/handle all the way down in one continuous drag,
   - tap the scrim above the sheet,
   - fling the comments list downward to dismiss.
4. **Expected for every vector:** keyboard dismisses with the sheet, never returns, feed fully interactive.

### Scenario 3 — full incident session replay

Replays the reporter's session end-to-end (contamination in sheet A, strand after sheet B):

1. Open comments, focus the composer (keyboard up).
2. Background the app 1–2 minutes, resume.
3. Close the sheet.
4. Browse the feed ~1 minute; background and resume briefly once more.
5. Open comments on a different video, type and post via the green button, close the sheet a second or two later.
6. Tap **Share** on the video, then close the share sheet.
7. **Expected:** no keyboard over the feed at any point after step 3, and specifically none after steps 5–6. Watch ~30 seconds at the end.

### Scenario 4 — no-regression sweep of normal composer UX

The fix sends a raw `TextInput.clearClient` at every dismissal, including healthy ones, so confirm nothing about normal keyboard use degraded:

- Green-button send → keyboard stays up for consecutive sends; comment appears in the list immediately.
- Keyboard **send key** → keyboard drops (framework behavior), comment posts.
- Reply to an existing comment → composer focuses and keyboard opens; cancel reply behaves.
- In-composer caret-down dismiss button → keyboard drops; tapping the field again re-opens it normally.
- Close the sheet, immediately reopen it, tap the field → keyboard opens normally (fresh session after the clearClient).
- `@`-mention suggestions still appear while typing; editing an own comment focuses and saves.

---

**Optional engineer check:** with Xcode attached at the moment after a Scenario 1 dismissal, Debug View Hierarchy should show no `FlutterTextInputView` in the window and no first responder — that's the state the fix guarantees and the state UIKit cannot re-present from.

**Comparison run (optional but valuable):** running Scenario 1 once on TestFlight build 792 to observe the strand validates the procedure; the re-present is not deterministic, so absence-of-bug on 792 in one run proves nothing, while the fix's success criterion (keyboard always down after dismissal, every run) is deterministic.

Rough time: 15–20 minutes for scenarios 1, 2, and 4; scenario 3 adds ~5.

mobile/lib/screens/comments/widgets/unfocus_on_sheet_dismiss.dart
```diff
@@ -82,8 +96,14 @@ class _UnfocusOnSheetDismissState extends State<UnfocusOnSheetDismiss> {
...
-    // during background/resume — see the class doc). Hide explicitly; this is
-    // a harmless no-op when the keyboard is already down.
+    // during background/resume — see the class doc). clearClient zeroes the
+    // engine's stale client id so the orphaned FlutterTextInputView stops
+    // being re-presentable (#6007); hide then resigns the first responder and
+    // removes the view. Both are harmless no-ops when the keyboard is already
+    // down or the framework connection is healthy.
+    unawaited(
+      SystemChannels.textInput.invokeMethod<void>('TextInput.clearClient'),
+    );
```
hm21, line 121: Honestly, I'm skeptical that this will fix the issue because it looks like a low-level hack, but we can try it.

mobile/lib/screens/comments/widgets/unfocus_on_sheet_dismiss.dart
```diff
@@ -82,8 +96,14 @@ class _UnfocusOnSheetDismissState extends State<UnfocusOnSheetDismiss> {
     FocusManager.instance.primaryFocus?.unfocus();
     // unfocus() sends nothing over the text-input channel when the framework
     // side of the connection is already gone (platform-initiated teardown
-    // during background/resume — see the class doc). Hide explicitly; this is
-    // a harmless no-op when the keyboard is already down.
+    // during background/resume — see the class doc). clearClient zeroes the
+    // engine's stale client id so the orphaned FlutterTextInputView stops
+    // being re-presentable (#6007); hide then resigns the first responder and
+    // removes the view. Both are harmless no-ops when the keyboard is already
```
hm21, line 117: I wonder if we should add a log here so that, if the user still experiences this issue, we can 100% verify that the method was called.

hm21 (approved):
I left two comments, but nothing blocking.

Note while testing, I noticed that when I press the keyboard's Send button, the message is sent and the keyboard is dismissed. However, when I tap the green Send button, the keyboard stays open. I'm not sure if that's intended or not.

hm21 (approved):
LGTM

mobile/lib/screens/comments/widgets/unfocus_on_sheet_dismiss.dart
rabble: The concern is fair. This uses Flutter’s own engine reset order (`clearClient` then `hide`) for the verified split-brain state, and the regression test pins that channel sequence. The follow-up commit also adds field logging so a physical-device recurrence can confirm the path fired. The PR now has two approvals and a green full CI run.

*(Response by Codex on behalf of @rabble)*

mobile/lib/screens/comments/widgets/unfocus_on_sheet_dismiss.dart
rabble: Added in commit `7561efb`: one unified-log entry records the dismissal trigger and focus shape before `unfocus + clearClient + hide`.

*(Response by Codex on behalf of @rabble)*

PR #6009, opened 2026-07-10
feat(settings): storage management screen with configurable cache limit

hm21:
Closes #6013

## Description

Adds a **Storage** screen (tile under General Settings → App) giving users manual control over the app's on-disk footprint. Builds on the on-disk cache bounding from #5986 (now merged).

image

**Storage screen (`/storage-management`)**
- Current clearable cache size + one-tap **Clear** (with confirmation): empties the video/image download caches, temp renders, and transition seams — including orphaned/leaked files, not just DB-tracked entries. The clip library and all user content are never touched.
- A persisted **cache-size slider** (0.5–10 GB, default 2 GB) caps how much the app may store, with a live "≈ N videos" estimate; the chosen limit is applied to the media cache and trimmed immediately.
- A **library audit** that finds clips whose backing video file is gone and offers to remove them.

**Media-cache plumbing (`media_cache`)**
- Runtime-settable byte budget + `enforceCacheLimits(force:)` with a queued follow-up sweep, so lowering the budget awaits a pass that observes the new value instead of coalescing behind an older one.

**Layering**
- The storage cubit/state live in `lib/blocs/storage/` and shared cache-limit constants in `lib/constants/`, so the settings UI never imports the service layer directly (UI→service boundary ratchet).
- Storage is a first-class route (`RouteType.storageManagement`, flat `/storage-management`). A nested `/settings/storage` parsed to `RouteType.settings`, so `routeNormalizationProvider` rewrote it back to `/settings` and bounced the user off the screen the instant it opened.

**Related Issue:**  — relates to #5987 (runaway on-disk growth, closed by #5986).

## Out of Scope

- Automatic thumbnail regeneration. A missing thumbnail doesn't break a clip (it's regenerable from the video), so only clips with a missing **video file** are treated as broken/removable.

## Verification

- `flutter analyze lib test packages/media_cache` — clean
- `media_cache` package: 199 tests pass (incl. queued-force-sweep + clock-throttle)
- Storage + route tests: 112 pass (route round-trip + orphan-clear regressions)
- Rebased onto fresh `origin/main` after #5986 merged; conflicts in `media_cache_manager`/`enforce_cache_limits_test` resolved (kept the merged `clock.now()` throttle + the queued force-sweep).
- Manual: cleared 668 MB → ~0 MB; slider shows scaling video counts; storage tile opens and stays.

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)

PR #6023, opened 2026-07-10
feat(support): add device and environment diagnostics to log export header

### LINKED ISSUE #6024: feat(support): add device and environment diagnostics to log export header
## Problem

Support triage for bug reports often needs follow-up questions that the exported log header can answer automatically: device model, simulator/emulator status, current network type, text scale, accessibility settings, app uptime, memory tier, and video cache usage.

Without these details, layout issues, long-session bugs, low-memory behavior, and cache-related reports are harder to diagnose from the first report.

## Acceptance criteria

- Log exports include best-effort environment diagnostics in the header.
- A failed or unavailable probe omits only its own line and does not fail the export.
- Device diagnostics avoid user-assigned machine names and other personal identifiers.
- Provider wiring avoids import cycles while preserving existing consumer imports.
- Tests cover the new diagnostic header behavior and provider override path.

## Implementation

Resolved by #6023.

hm21:
Closes #6024

## What

Extends the comprehensive log export header with diagnostics that support triage previously had to ask the reporter for:

- **Device**: hardware model (`iPhone17,2`, `Google Pixel 8`), with a `(Simulator)` / `(Emulator)` marker on non-physical devices
- **Network**: current connectivity (wifi / mobile / none / vpn) via the existing `connectivity_plus` dependency
- **Text Scale + Accessibility**: system text scaling and active accessibility features (bold text, reduce motion, screen reader, ...)
- **App Uptime**: time since process start via a new `AppUptime` util marked in `main()`
- **Memory Tier**: low/medium/high from the existing `DeviceMemoryUtil`
- **Cache**: video cache bytes used vs the user-configured limit from `StorageManagementService`

Every probe is best-effort: a failing or unavailable source omits its line instead of breaking the export.

## Why the provider move

`bugReportService` now watches `storageManagementServiceProvider`, which itself imports `social_providers.dart` for `clipLibraryServiceProvider`. Wiring it inside `social_providers.dart` would have created an import cycle, so the provider moved to a new `bug_report_providers.dart`, re-exported through the existing `app_providers.dart` compatibility barrel. Consumer imports do not change.

## Privacy note

Windows reports no device line because `device_info_plus` exposes no hardware model there, only the user-assigned `computerName`. Linux and Fuchsia similarly omit this line because they do not provide a meaningful hardware model for support triage.

## Test plan

- `flutter test test/unit/services/bug_report_log_export_test.dart`
- `flutter test test/router/minor_account_review_router_test.dart test/router/minor_account_review_error_redirect_test.dart`
- Pre-push hook: analyzer, generated files check, changed-file tests


rabble:
Structural blocker before review/merge: this draft targets feat/settings-storage-management instead of main. Repo policy forbids stacked PRs; dependent work must be combined into one PR, or this branch must be rebased onto main after #6009 lands and the PR base changed to main. (Review by Codex on behalf of @rabble)

NotThatKindOfDrLiz:
@rabble The previous stacked-PR blocker is stale. PR #6023 now targets `main`, #6009 has merged, and GitHub reports the branch is mergeable. It is a few unrelated `main` commits behind, but it is no longer stacked on `feat/settings-storage-management`.

NotThatKindOfDrLiz:
Takeover complete.

What changed:
- Rebased the branch onto current `origin/main`.
- Removed the redundant outer `try/catch` around `DeviceMemoryUtil.getMemoryTier()` so the diagnostics code matches the helper's existing fallback behavior.
- Updated the PR body to remove the stale stacking note and refreshed #6024 with a real problem statement and acceptance criteria.

Verification:
- `flutter test test/unit/services/bug_report_log_export_test.dart`
- `flutter test test/router/minor_account_review_router_test.dart test/router/minor_account_review_error_redirect_test.dart`
- Pre-push hook: analyzer, generated files check, changed-file tests
- GitHub checks are green after rerunning the unrelated `video_feed_page_test.dart` failure; the exact failing test passed locally before the rerun.

Final pass: PR is mergeable, targets `main`, has linked issue #6024, and I found no remaining blockers.

NotThatKindOfDrLiz (approved):
Approved after takeover: rebased onto current main, removed the redundant memory-tier catch, refreshed PR/issue metadata, verified locally, and GitHub checks are green.

mobile/lib/services/bug_report_service.dart
```diff
@@ -830,6 +843,123 @@ class BugReportService {
...
+    }
+    final dispatcher = ui.PlatformDispatcher.instance;
+    buffer.writeln(
+      'Text Scale: ${dispatcher.textScaleFactor.toStringAsFixed(2)}',
+    );
+    final accessibility = _activeAccessibilityFeatures(
+      dispatcher.accessibilityFeatures,
+    );
+    if (accessibility.isNotEmpty) {
+      buffer.writeln('Accessibility: ${accessibility.join(', ')}');
```
realmeylisdev, line 870: Non-blocking / conscious product call: this writes active accessibility features (e.g. `screen reader`, `invert colors`, `high contrast`) into a file the user shares via the system sheet / Downloads / email. The flags are genuinely useful for reproducing UI bugs, but assistive-tech usage is more sensitive than the `computerName`/hostname this PR deliberately omits. Since the export is user-initiated and user-controlled it's defensible — flagging it as a choice worth being deliberate about, not a change request.

mobile/test/unit/services/bug_report_log_export_test.dart
```diff
@@ -51,6 +56,143 @@ void main() {
...
+    setUp(() {
+      messenger.setMockMethodCallHandler(connectivityChannel, (call) async {
+        if (call.method == 'check') return <String>['wifi'];
+        return null;
+      });
+    });
+
+    tearDown(() {
+      messenger.setMockMethodCallHandler(connectivityChannel, null);
+      AppUptime.reset();
```
realmeylisdev, line 168: Non-blocking (merged-isolate hygiene): `buildEnvironmentDiagnostics()` → `DeviceMemoryUtil.getMemoryTier()` writes the process-global static `_cachedTier`, and this `tearDown` resets `AppUptime` + the connectivity channel but not that cache. There's no current victim (nothing else reads `_cachedTier` from tests), but under `very_good --optimization` it leaves `_cachedTier` populated for later suites. `device_memory_util_test.dart` already does `tearDown(DeviceMemoryUtil.resetCache)`; mirroring it here keeps the static-reset convention symmetric.

mobile/test/unit/services/bug_report_log_export_test.dart
```diff
@@ -51,6 +56,143 @@ void main() {
...
+    test('reports app uptime once marked and omits it before', () async {
+      final before = await BugReportService().buildEnvironmentDiagnostics();
+      expect(before, isNot(contains('App Uptime: ')));
+
+      AppUptime.markStarted();
+      final after = await BugReportService().buildEnvironmentDiagnostics();
+      expect(after, contains('App Uptime: '));
+    });
+
+    test('omits the cache line without a storage service', () async {
```
realmeylisdev, line 190: Non-blocking coverage note: this only asserts the no-storage-service path. The new `Cache: X used / Y limit` composition, `_formatBytesShort`'s GB branch (`mb >= 1024`), and `_formatDuration`'s `hours > 0` branch have no covering assertion. The constructor already accepts a `StorageManagementService?`, so a single test injecting a fake returning e.g. 2 GB used / 5 GB limit would exercise the formatter + line composition and guard against a used/limit swap.

realmeylisdev (approved):
**Formal review — verified on this branch against `origin/main`.**

Local checks: `flutter analyze` (all 7 changed files) clean · the new `bug_report_log_export_test.dart` (15 tests) passes · `build_runner` regen = zero drift (generated providers in sync) · shared-channel ratchet passes.

Confirmed correct:
- The provider move genuinely breaks the `storage_providers ↔ social_providers` import cycle, and every consumer resolves `bugReportServiceProvider` through the `app_providers` barrel — so the move is transparent.
- Device probing uses the real `device_info_plus` 10.1.2 channel (`dev.fluttercommunity.plus/device_info`); the test mock truly exercises the path (not a fall-through).
- The "best-effort omits its line" contract holds: the only probes without a local `try` (text scale / accessibility / uptime / memory tier) are non-throwing — `getMemoryTier()` catches internally and always returns a tier — and the whole header sits inside `exportLogsToFile`'s outer catch, so removing the redundant memory-tier `try` was safe.
- No PII regression: the device line is non-identifying and Windows/Linux/Fuchsia correctly omit it.

Nothing blocking. One doc nit + a few non-blocking observations inline.

Nit (line 3 is outside the diff, so noting here): `social_providers.dart` ABOUTME still lists `bugReport` in its provider inventory after the provider moved to `bug_report_providers.dart` — drop `bugReport,` from the header to keep the comment in step with the code.

LGTM.

mobile/test/unit/services/bug_report_log_export_test.dart
hm21: Fixed in aa99c869e — added `DeviceMemoryUtil.resetCache()` to the teardown to keep the static-reset convention symmetric.

mobile/lib/services/bug_report_service.dart
hm21: Agreed, keeping as-is. The export is user-initiated and user-controlled, and the flags are genuinely useful for reproducing UI bugs — conscious call, not a leak.

mobile/test/unit/services/bug_report_log_export_test.dart
hm21: Leaving this one — the untested branches are a diagnostics-header string with low blast radius, and we prefer not to add coverage here. Noted for a future pass if the header grows.

PR #6025, opened 2026-07-10
fix(video-publish): app auto-double-publishes videos on the audio-reuse path

### LINKED ISSUE #6018: fix(video-publish): app auto-double-publishes videos on the audio-reuse path (concurrent publishDirectUpload)
## Summary

The app republishes some videos a **second time, automatically, with no user edit**. Two kind-34236 events are published for the same coordinate ~20–25 seconds apart; they are byte-identical in content and tags (same `d`, `imeta`, `title`, `published_at`) and differ only in `created_at` and `id`/`sig`. Because each publish mints a new event id, likes fragment across the two versions (a like lands on whichever version id the liker's feed happened to serve).

Observed on ~20% of one creator's videos, most recently **2026-07-09** — this is ongoing on current builds, not historical.

## Evidence (production)

Creator CatesPlace (`713fc3bf6c1c5cc593736c2b1375ba165e173b05d31787ed3b0591796a489a3f`):

- Coordinate `34236:713fc3bf6c1c5cc593736c2b1375ba165e173b05d31787ed3b0591796a489a3f:074deb1fea80e101da43a5ed87ec3f5d46719df91c606316ec44fb64147f897b`
  - superseded `ab624738708a6041fdc16f6173b2effe264292d577722d25c062253bd62100f8` — `created_at` 1780768829, `published_at` 1780768853
  - current `15d8c456e173f55d3d8521631797f154c5d7abedd8aa6608869bafd31674dc6e` — `created_at` 1780768854, `published_at` 1780768853
  - 17 of 18 likes landed on the superseded id over a 9.4-hour window **after** the replacement already existed.
- Latest pair (2026-07-09): `14393dfa66e8d2110eb42a727aa3b244da338f0e3f76991f04039a78eacadc6c` → `94cd8dc2c7543127d04b72508b50e4e9e987c61a86ae5ca61412dec4d19c3b1f`.
- Additional pairs on coordinates `82d116681346149631c557176a8f6f23993773b5bc8f931de3129da8ca75c07e` and `a058a346…`.

**100% correlation:** every sampled double-published coordinate had `allow_audio_reuse=true` and a *failed* audio publish (no audio `e` tag on the event).

## Root cause

A **concurrent double-invocation of `VideoEventPublisher.publishDirectUpload` for the same upload**, with the audio-reuse step as the amplifier. The signature is decisive:

1. Two event ids with `created_at` ~20s apart ⟹ two separate `createAndSignEvent` calls (each recomputes `NostrTimestamp.now()`, which backdates `created_at` by a 30s drift tolerance — `signer_factory.dart:220-221`, `nostr_timestamp.dart:48-57`). Each `publishDirectUpload` signs exactly once (`video_event_publisher.dart:1462-1470`) ⟹ two invocations occurred.
2. Both events share the coordinate/d-tag ⟹ both resolved the same `upload.videoId` (`d = upload.videoId`, `video_event_publisher.dart:911-914`); the second call reused the first's upload via `findReusableUpload` (`video_publish_service.dart:445` → `upload_manager.dart:282`), which is why imeta/URLs/dims/SHA256 are byte-identical.
3. **Identical `published_at` proves concurrency, not a delayed retry.** `published_at` is freshly computed per call at `video_event_publisher.dart:1179` (`DateTime.now()`), never cached. Two identical values ⟹ both calls executed that line in the same wall-second. A delayed re-sign would rebuild it in a later second; and the normal retry path reuses the *cached signed event* (`:1462`, same id) so it cannot produce two ids at all. (This rules out the "re-sign after a false-negative relay OK" hypothesis.)
4. **How two calls hit `:1179` in the same second despite starting apart:** `published_at` is captured *after* the Blossom upload completes. The first call waits out its multi-second upload; the second call reuses the already-ready upload (`findReusableUpload` returns immediately) and reaches `:1179` almost instantly. The ~20s `created_at` gap then comes from the two signings being staggered behind the serial audio step / remote signer.
5. **Why `allow_audio_reuse` + failed audio is the amplifier:** `_publishAudioEvent` (`video_event_publisher.dart:1302` → `video_event_publisher_audio.dart:121`) is a ~20–25s serial network step that sits **before** both the video sign (`:1465`) and the idempotency claim (`_persistRetryableSignedEvent`, `:1481-1482`). During that window both concurrent calls are unclaimed → `_loadRetryableSignedEvent` returns null for both (`:760-767`) → both re-sign fresh → two ids. Two concurrent `_publishAudioEvent` runs on the same path also collide (temp-file cleanup `video_event_publisher_audio.dart:387-392` + duplicate Blossom upload) → both audio publishes fail → `allow_audio_reuse=true` stays with no `e` tag. Without audio-reuse the sign+claim happen back-to-back and the window collapses (which is why the two are 100% correlated). Slow/failing audio is also selection bias — only wide windows admit a duplicate.

**Contributing guard gap:** the publish provider's `.preparing` guard (`video_publish_provider.dart:326`) is reset to `.idle` by a `Future.delayed(600ms, clearAll)` (`:466`) while the publish future runs 20–30s, and `.uploading` is never set (dead branch). That leaves a ~20–30s window where a publish is in-flight but the guard reads idle, and any second entry is unguarded. The bloc's `alreadyUploading` guard keys on a per-call unique id (`:395`), so two publishes of the same source draft both pass.

## Impact

Now that funnelcake counts by coordinate (divine-funnelcake#568, deployed 2026-07-10), the displayed **count** is unaffected. Residual impact: it manufactures the superseded-id condition automatically, which fragments own-heart state and the "Liked by" list across versions (see the related follow-ups), and it publishes redundant events to relays.

## Suggested fix

1. **Per-upload in-flight lock in `publishDirectUpload`**, keyed by `upload.videoId` (d-tag), claimed at method entry **before** the audio branch (`video_event_publisher.dart:~906`, ahead of `:1282`) and released in a `finally`; a concurrent second call awaits the first's result or short-circuits. No such dedup exists today.
2. **Move the idempotency claim before the audio step** — persist `nostrEventId` / a d-tag publish-lock before `_publishAudioEvent`, so a concurrent call reuses the cached signed event (`:1462`) instead of re-signing.
3. **Provider defense-in-depth** — hold the guard for the lifetime of the publish future instead of the 600ms `clearAll` reset (`:466`), gate on `sourceDraftId`, and remove the dead `.uploading` branch.

## Test plan

Unit test: call `publishDirectUpload` twice on the same `PendingUpload` with a slow/failing `_audioExtractionService`/`_blossomUploadService` mock. Today it yields two distinct signed events with identical `published_at` (reproduces the bug); after the fix it must yield exactly one broadcast kind-34236 event. This is both the repro and the regression guard.

## Open questions (need a publish-time log capture)

- Which UI surface fires the second entry — a re-tap of Post while the draft is still visible (`drafts_tab.dart:169`) vs a re-dispatched editor Post action.
- Confirm both calls share one `upload.id` via `findReusableUpload` (identical d-tag strongly implies it).
- Whether the ~20s stagger comes from a serialized remote signer (Keycast/bunker) or from serialized audio extraction/upload.

## Note on feed dedup

Mobile dedups addressable 34236 in-session, newest-`created_at`-wins (`video_event_service.dart:2205-2272`), so live viewers converge on the newer id. But the dedup is per-subscription and in-memory per session — the first-broadcast id is alone on the relay for ~20s and reachable by any cold/paginated REST fetch or by-event-id surface, so likes still split onto the superseded id, and dedup does not migrate likes already placed there.

---

### Context

Follows from #4432 (lost likes on edit). That issue's core symptom — the displayed like count collapsing after an edit — was a backend counting bug and was fixed by divinevideo/divine-funnelcake#568 (migration `000180`, coordinate-aware counting), deployed to production 2026-07-10 06:42 UTC. This issue tracks a distinct publish-pipeline defect that *manufactures* the superseded-id condition with no user edit.

hm21:
## Description

The app republished some videos a second time automatically: two concurrent `publishDirectUpload` invocations for the same upload each signed and broadcast their own kind-34236 event for the same coordinate. The ~20–25s audio-reuse step sits before both the video signing and the idempotency claim, so during that window neither call could see the other's signed event and both re-signed fresh — two ids, identical content, likes fragmenting across the two versions.

Two changes:

- **`VideoEventPublisher.publishDirectUpload` — in-flight lock keyed by `upload.videoId` (the d-tag).** A concurrent second call awaits the in-flight publish's result instead of signing a duplicate; the lock is released in a `finally`, so sequential re-publishes (real edits) still work. This is the core fix and covers every entry path (editor post, drafts-tab re-tap, background retry), since they all funnel through this method. The issue's suggested "move the idempotency claim before the audio step" is not implementable literally — no event id exists before signing — so the d-tag lock *is* that claim.
- **`VideoPublishNotifier.publishVideo` — defense-in-depth guard on `sourceDraftId`,** held in a private set for the lifetime of the publish future. The previous state-based guard reopened ~600ms after navigation because `clearAll` resets the state while the publish keeps running 20–30s; that's why it lives outside `state`. The dead `.uploading` branch (never assigned anywhere) is removed.

Both regression tests were verified red on unfixed code and green with the fix: the publisher test follows the issue's test plan (second call during a gated, *failing* audio extraction — matching the production signature of `allow_audio_reuse=true` with no audio e-tag — must yield exactly one signed/broadcast event), plus guards that sequential re-publishes and different uploads are not coalesced. The provider test pins that `reset()` during an in-flight publish does not reopen the gate.

**Related Issue:** Closes #6018

## Out of Scope

- Migrating likes already placed on superseded event ids (tracked by the follow-ups referenced in #6018).
- The pre-existing stuck-`.preparing` state when a publish throws before navigation (unrelated latent issue, unchanged behavior).
- Identifying which UI surface fired the second entry (open question in #6018) — the lock makes the answer irrelevant for correctness.

## Verification

- `flutter test test/services/video_event_publisher_double_publish_test.dart` (new, red without fix / green with fix, verified via `git stash`)
- `flutter test test/providers/video_publish_provider_test.dart` (extended, same red/green verification)
- All 12 `video_event_publisher_*` suites, `test/services/video_publish/`, `test/widgets/library/drafts_tab_test.dart` — green
- `dart format` — clean

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/providers/video_publish_provider.dart
+++ b/mobile/lib/providers/video_publish_provider.dart
@@ -71,6 +71,13 @@ MentionResolutionService? createVideoPublishMentionResolutionService(
 
 /// Manages video publish screen state including playback and position.
 class VideoPublishNotifier extends Notifier<VideoPublishProviderState> {
+  /// Source-draft ids with a publish currently in flight. Lives outside
+  /// [state] because [clearAll] intentionally resets the state ~600ms
+  /// after navigation while the publish future keeps running for 20s+;
+  /// a state-based guard reopens in that window and admits duplicate
+  /// publishes of the same draft (#6018).
+  final Set<String> _inFlightSourceDraftIds = {};
+
   DraftStorageService get _draftService =>
       ref.read(draftStorageServiceProvider);
   CawgVerifierClient get _cawgVerifierClient =>
@@ -322,8 +329,9 @@ class VideoPublishNotifier extends Notifier<VideoPublishProviderState> {
     BuildContext context,
     DivineVideoDraft draft,
   ) async {
-    // Only block if actively preparing/uploading
-    if (state.publishState == .preparing || state.publishState == .uploading) {
+    final sourceDraftId = draft.sourceDraftId ?? draft.id;
+    if (state.publishState == .preparing ||
+        _inFlightSourceDraftIds.contains(sourceDraftId)) {
       Log.warning(
         '⚠️ Publish already in progress, ignoring duplicate request',
         name: 'VideoPublishNotifier',
@@ -332,6 +340,7 @@ class VideoPublishNotifier extends Notifier<VideoPublishProviderState> {
       return;
     }
 
+    _inFlightSourceDraftIds.add(sourceDraftId);
     state = state.copyWith(publishState: .preparing);
 
     try {
@@ -501,6 +510,7 @@ class VideoPublishNotifier extends Notifier<VideoPublishProviderState> {
         stackTrace: stackTrace,
       );
     } finally {
+      _inFlightSourceDraftIds.remove(sourceDraftId);
       Log.info(
         '🏁 Publish process completed',
         name: 'VideoPublishNotifier',
--- a/mobile/lib/services/video_event_publisher.dart
+++ b/mobile/lib/services/video_event_publisher.dart
@@ -200,6 +200,12 @@ class VideoEventPublisher {
   int _totalEventsFailed = 0;
   DateTime? _lastPublishTime;
 
+  /// In-flight direct publishes keyed by the upload's d-tag
+  /// ([PendingUpload.videoId]). Coalesces concurrent
+  /// [publishDirectUpload] calls for the same upload so only one
+  /// addressable event is signed and broadcast (#6018).
+  final Map<String, Future<bool>> _inFlightDirectPublishes = {};
+
   /// The outer timeout that will bound the next call into
   /// [NostrClient.publishEventAwaitOk] inside [_publishEventToNostr], computed
   /// live from [outerPublishTimeoutFor] and the current
@@ -850,6 +856,13 @@ class VideoEventPublisher {
   }
 
   /// Publish a video directly without polling (for direct upload)
+  ///
+  /// Concurrent calls for the same [PendingUpload.videoId] (the
+  /// addressable event's d-tag) are coalesced: the second caller awaits
+  /// the in-flight publish's result instead of signing and broadcasting
+  /// a duplicate event with a fresh id (#6018). The audio-reuse step can
+  /// keep a publish in flight for 20s+, which is the window where the
+  /// duplicates were minted.
   Future<bool> publishDirectUpload(
     PendingUpload upload, {
     int? expirationTimestamp,
@@ -868,7 +881,8 @@ class VideoEventPublisher {
     VideoReplyContext? replyContext,
     bool addReplyToFeed = false,
   }) async {
-    if (upload.videoId == null || upload.cdnUrl == null) {
+    final videoId = upload.videoId;
+    if (videoId == null || upload.cdnUrl == null) {
       Log.error(
         'Cannot publish upload - missing videoId or cdnUrl',
         name: 'VideoEventPublisher',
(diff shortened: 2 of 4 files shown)


NotThatKindOfDrLiz (approved):
Approved after re-review. The coalescing fix is well-scoped, race-safe for the observed duplicate publish path, and covered by targeted regression tests.

PR #6026, opened 2026-07-10
fix(video): preserve unowned tags on metadata-edit republish

### LINKED ISSUE #6019: fix(video): metadata edit republish drops audio-attribution, provenance, and hint tags
## Summary

The metadata-edit flow rebuilds the replacement kind-34236 event's tag list **from scratch**, copying only a fixed allow-list of tags. As a result, every edit silently **drops** the audio-attribution `e` tag, `expiration`, `client`, `r`/`h` relay hints, and ProofMode/C2PA provenance tags from the event.

## Evidence

- `VideoMetadataUpdateService.updateVideo` assembles the replacement tags at `mobile/lib/services/video_metadata_update_service.dart:163-259`, copying only: `d`, `imeta`, `title`, `t`, `l`, engagement counts (via `extractEngagementCountTags`, `:230-232`), `published_at`, `duration`, `alt`, `p` collaborators, `a` inspired-by, and `allow_audio_reuse`. There is no generic copy of `originalVideo.nostrEventTags`.
- Audio-attribution `e` tags are written at original publish as `['e', audioEventId, relay, 'audio']` (`mobile/lib/services/video_event_publisher.dart:1270`, `:1313`) and have no counterpart in `updateVideo`, so audio attribution is lost on the first edit.
- Any provenance/expiration/hint tags present on the original are likewise not in the allow-list and are dropped.

## Impact

Data loss on every metadata edit: an edited video loses its "made with this sound" attribution (and its provenance/expiration/hint metadata), independent of the like-count symptom. Because edits also happen automatically via the double-publish path (see the related publish-pipeline follow-up), this can occur without a deliberate user edit.

## Suggested fix

Preserve the original event's tags by default and override only the fields the edit actually changes, rather than rebuilding from a subset allow-list. Concretely: start from `originalVideo.nostrEventTags`, replace the edited fields (`title`, `t`, `imeta`/thumbnail, etc.), and keep everything else — while still enforcing the required `d` and `published_at` semantics.

## Test plan

Edit-flow test asserting that an original event carrying an audio-attribution `e` tag (and a provenance/expiration tag) still carries them after a metadata edit, while the edited fields are updated.

---

### Context

Found during the #4432 investigation (lost likes on edit). The #4432 core symptom was a backend counting bug fixed by divinevideo/divine-funnelcake#568 (deployed 2026-07-10); this is a separate data-loss bug in the same edit flow, surfaced by the same trace.

hm21:
## Description

`VideoMetadataUpdateService.updateVideo` rebuilt the replacement kind-34236 tag list from scratch using a fixed allow-list, so every metadata edit silently dropped the audio-attribution `e` tag, `expiration`, `client`, `r` relay hints, ProofMode/C2PA provenance tags, reply-threading tags, and subtitle `text-track` tags.

This PR inverts the logic: the original event's tags are preserved by default, and only the tag groups the edit flow actually owns are rebuilt from the editor state. Owned groups are matched by name (`d`, `imeta`, `title`, `summary`, `t`, `published_at`, `duration`, `alt`, `allow_audio_reuse`) or by marker (content-warning group, `p`+`collaborator`, `a`+`mention`/`inspired-by` with a `34236:` prefix). Marker-based matching is needed so reply-threading tags (unmarked `p`/`a`, plus `E`/`K`/`P`/`A`) and mention `p`-tags survive while the collaborator and inspired-by tags the editor controls are still fully replaced.

Two related data-loss bugs with the same root cause are fixed along the way, because the ownership boundary forces a decision for each group:

- The edit now writes the full NIP-36/NIP-32 content-warning group (`content-warning` + `L` + `l`) like the original publish does, instead of only `l` labels. Since the ownership filter must strip the original group (a user can remove warnings in the edit), re-emitting only `l` would have silently removed the `content-warning` tag other clients rely on.
- `summary` is now rewritten from the edited description. Preserving it verbatim would leave a stale summary whenever the description changes, and dropping it (previous behavior) loses data; rewriting matches what `VideoEventPublisher` writes at original publish.

`extractEngagementCountTags` and its helper-only test file are removed: engagement-count tags are not in the owned set, so generic preservation subsumes the helper, and the behavior stays pinned by the service-level tests.

**Related Issue:** Closes #6019

## Out of Scope

- The double-publish path mentioned in #6019 has its own publish-pipeline follow-up; this PR fixes the shared `updateVideo` edit flow.
- The `imeta` tag is still rebuilt from parsed fields (thumbnail may change during an edit) rather than patched in place.

## Verification

- `flutter test test/services/video_metadata_update_service_test.dart` (20 tests, includes 8 new regression tests for preservation and replacement)
- `flutter test test/widgets/share_video_menu_edit_collaborator_tags_test.dart test/widgets/share_video_menu_publish_failure_test.dart test/widgets/video_metadata/modes/edit/video_metadata_edit_bottom_bar_test.dart`
- `dart format` — clean

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


NotThatKindOfDrLiz (approved):
Approved. I re-reviewed the tag preservation changes and did not find any blocking issues. The owned-tag strip/rebuild model is coherent, targeted tests cover the important metadata preservation and edit-state edge cases, CI is green, and the remaining allow_audio_reuse cache-path gap is tracked separately in #6045.

Issue #173, opened 2025-11-15
How do i get APK


KOKOTAR:
where can i get apk

CASTLECALAMITY:
The best course of action would be to either wait for a release or try to fix the problem yourself and build your own APK.

rabble:
we took it down because it was broken. the android build is now working. We've got some things to do before we republish the apk.

PR #6027, opened 2026-07-10
fix(minor-safety): refresh account-state providers on foreground-resume (#185)

### LINKED ISSUE #6033: fix(minor-safety): refresh account-state providers on foreground-resume (#185)
Auto-created from PR #6027.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6027
**Branch:** `mjb/minor-safety-resume-refetch`

mbradley:
Closes #6033

## What this changes

The app decides whether to restrict a protected minor's DMs and lock adult content by reading the account's status from Keycast. Today it only re-reads that status when you sign in or switch accounts. If the status changes on the server while the app is just sitting open and signed in (a moderator flags an account as a protected minor, or an approval comes through), the app keeps using the old status until the next relaunch.

This makes the app re-check the status when it comes back to the foreground, which is what the web app already does when its window regains focus.

## Where

- A new `AccountStateResumeRefresher` (`mobile/lib/services/`) re-fetches the two providers the protections read: `protectedMinorStatusProvider` (the DM and adult-content gates) and `currentMinorAccountReviewStatusProvider` (the review/suspend gate).
- It's called from the existing app-resume handler (`mobile/lib/widgets/app_lifecycle_handler.dart`), alongside the relay reconnect and notification refresh that already run on resume.
- A 15-minute guard means someone who backgrounds and foregrounds a lot doesn't trigger a Keycast check every single time.

## Why it's safe

This can only ever tighten protection or correctly lift it, never open it by mistake. While the re-fetch is in flight the gates keep using the last known answer, so a protected minor never briefly loses their DM or content restriction. A re-fetch that fails (offline, expired token) leaves protection exactly as it was.

## Related

Relates to divinevideo/support-trust-safety#185 (part of epic #173).

## Verification

- New `test/services/account_state_resume_refresher_test.dart`: the 15-minute guard and the signed-out gate, plus a container test proving both providers actually re-fetch on resume and stay put within the interval.
- `flutter analyze` clean; the new service ships with its test (untested-services floor unchanged); no generated code.

## Type of Change

### DIFF
--- a/docs/superpowers/specs/2026-07-10-foreground-resume-account-state-refetch-design.md
+++ b/docs/superpowers/specs/2026-07-10-foreground-resume-account-state-refetch-design.md
@@ -1,0 +1,124 @@
+# Foreground-resume account-state refetch (mobile)
+
+Design for the mobile half of divinevideo/support-trust-safety#185, part of
+the protected-minor (13-15) safeguards epic #173. Surfaced during the #176
+DM-restriction review.
+
+## Problem (verified against code)
+
+The two account-state providers are plain `FutureProvider`s that only
+recompute on an auth-state transition (they `ref.watch(currentAuthStateProvider)`)
+or an explicit `ref.invalidate`:
+
+- `protectedMinorStatusProvider` — drives the #176 DM gate + #175 content
+  lock through `isProtectedMinorProvider` / `isDmRestrictedProvider`.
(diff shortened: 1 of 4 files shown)


rabble (approved):
Reviewed the lifecycle wiring, TTL/auth gates, sticky-state behavior, and regression coverage. CI is green; no blocking findings. (Review by Codex on behalf of @rabble)

PR #186, opened 2025-11-18
Remove emedded relay and refactor code

derekross:
🔧 Major Changes

  1. Custom WebSocket Fetchers (Bypasses nostr_sdk bugs)

  Problem: The nostr_sdk Rust library ignores the authors filter in subscriptions, causing:
  - Home feed showing random videos instead of only followed users
  - Profile names not loading (showing "Loading..." forever)
  - Hashtag feeds not filtering correctly

  Solution: Created custom WebSocket fetchers that connect directly to relays:
  - CustomHomeFeedFetcher - Properly filters home feed by followed authors
  - CustomProfileFetcher - Loads user profiles with correct author filtering
  - CustomHashtagFetcher - Hashtag-based video discovery
  - Includes client-side validation as backup filtering layer

  Files:
  - mobile/lib/services/custom_home_feed_fetcher.dart (new)
  - mobile/lib/services/custom_profile_fetcher.dart (new)
  - mobile/lib/services/custom_hashtag_fetcher.dart (updated)
  - mobile/lib/providers/home_feed_provider.dart (updated)
  - mobile/lib/services/user_profile_service.dart (updated)

  2. Relay Connection Improvements

  Automatic reconnection:
  - Added ensureConnected() method to automatically reconnect before broadcast operations
  - Implemented lifecycle-based reconnection via BackgroundActivityManager
  - Fixed relay reconnection after app backgrounding on mobile
  - Force disconnect relays before reconnecting (SDK was keeping stale connections)

  Proper async patterns:
  - Replaced arbitrary Future.delayed() with proper connection polling
  - Exponential backoff polling (100ms → 2000ms) for connection establishment
  - Fixed reconnectAll() to actually wait for WebSocket connections

  Files:
  - mobile/lib/services/nostr_service.dart
  - mobile/lib/providers/app_providers.dart

  3. Post-Publish Video Caching

  Problem: Race condition where just-published videos show "not found" for a few seconds while Blossom CDN
  processes the upload.

  Solution: Cache local video file immediately after publishing to Nostr using existing VideoCacheManager:
  - Instant playback from local cache while CDN finishes processing
  - 1GB cache, 30-day expiration
  - Implements Instagram/TikTok pattern of seamless local playback

  Files:
  - mobile/lib/services/video_event_publisher.dart

  4. Comments & Likes Fixes

  - Implemented WebSocket-based comment and like fetching (bypassing broken nostr_sdk subscriptions)
  - Added addressable event support (d-tag matching) for comment queries
[...]

PR #5754, opened 2026-07-02
feat(minor-safety): protected-minor DM restriction (mobile) (#176)

mbradley:
Mobile half of the protected-minor DM restriction (support-trust-safety#176, epic #173). A protected minor (13-15) may only DM the pinned official Divine accounts (HQ + Moderation); everything else is blocked on send and hidden on receive. Client-side by necessity — NIP-17 gift-wraps hide the recipient, so relays can't enforce it.

Consumes the existing `isProtectedMinor` fail-closed / persistent seam (#174 / #175, already merged); this PR adds the approved-recipient trust model and the send/receive enforcement.

## Trust model (#4948)

Approved set = **pinned ∩ live NIP-05** (Tier 2). Two pinned accounts, each `minorContactable`: Divine HQ (`_@divinehq.divine.video`) and Divine Moderation (`moderation@divine.video`). The pin blocks attacker *addition* (no key the app didn't ship can join); the live NIP-05 leg is the *revocation* lever. Revocation is graded: a key swap drops immediately, a single absence needs a confirming recheck, a network blip keeps last-known.

## What's enforced

- **Send** — gated at the lowest NIP-17 primitive (`sendRumor`), covering every publisher (direct, group, reaction, file, drain-replay) at one seam; plus a pre-enqueue gate and group all-or-nothing.
- **Receive** — inbox, Requests, and the unread badge hide non-approved conversations; receive-time revalidation drops a revoked official live.
- **Affordances** — profile Message button, conversation route, and the message-request preview route guarded (the preview cubit checks the same all-approved predicate before reading any request metadata, so a stale or direct URL exposes nothing).
- **UX** — a blocked send shows neutral, non-retriable copy.
- **Fail-safe on unresolved status** — the DM gates read a dedicated fail-closed seam: only a positive not-protected verdict (fresh from Keycast, or persisted from a prior check) lifts the restriction. Loading, unknown, or a missing token all restrict, so the restricted party can't lift the gate by suppressing the check. When protection flips mid-session, a settled inbox list and the unread badge re-filter immediately instead of waiting for the next DM event.
  - Rollout note: every account starts with no persisted verdict, so all users pass through a brief restricted window at first launch after this ships, until the first Keycast check answers (normally sub-second, and usually resolved before any DM surface is opened). A successful not-protected answer persists, so the window never recurs for that account on that install.
[...]

PR #3911, opened 2026-05-03
feat(db): add outgoing_dms Drift table + DAO for durable DM queue

realmeylisdev:
## Description

Step 2 of the #3909 stack. Adds the persistence layer for the durable
outgoing-DM queue. Pure data layer — no behavior change for users yet,
no new wiring outside the `db_client` package.

**Related Issue:** Relates to #3909 (follow-up to #3902 / #3908)

### What's in here

`outgoing_dms` Drift table with **two independent** wrap-status
columns. NIP-17 sends are two relay publishes (recipient gift wrap +
self-addressed gift wrap), so the queue must track them separately —
otherwise a partial delivery either gets re-published to the recipient
(double-deliver) or the row is dropped and the sender silently loses
their own message.

| column | purpose |
|---|---|
| `id` (PK) | The rumor event id — stable across retries because `rumor_event_json` stores the serialized rumor (rumor.id is a hash of its fields, so re-mint would break receiver-side dedup) |
| `conversation_id` | Same shape as `direct_messages.conversation_id` so the BLoC can `Rx.combineLatest2(watchMessages, watchOutgoing)` for a unified timeline |
| `rumor_event_json` | The unsigned rumor JSON; retries re-wrap **this** instead of minting a new rumor |
| `recipient_wrap_status` | `pending` / `sent` / `failed`. String-encoded enum so new states (e.g. `cancelled`) don't need a schema bump |
| `self_wrap_status` | Same enum, independent transitions. The `false` value of `NIP17SendResult.selfWrapPublished` (from #3910) flips this to `failed` |
| `recipient_wrap_event_id` / `self_wrap_event_id` | Published kind-1059 ids for downstream debugging |
| `retry_count`, `last_error`, `last_attempt_at` | Mirror `PendingActions` for the retry service's backoff |
| `queued_at` | Ordering for retry sweeps |
| `owner_pubkey` | Multi-account isolation, mirroring `direct_messages.owner_pubkey` |

Indexes for the access patterns the queue + retry service need:
per-conversation watch, per-wrap-status retry sweeps, queued-at
ordering.

### Schema management

Schema version stays at **1**. The table is registered via the existing
`_createMissingTables` runtime CREATE-IF-NOT-EXISTS pattern, mirroring
`personal_reposts` (`app_database.dart:85-100`) and
`nip05_verifications`. Fresh installs get it via Drift's `m.createAll()`
on first open; existing installs get it on next launch via the
runtime check. No formal Drift migration to write.

### DAO

[...]

PR #6028, opened 2026-07-10
fix(dm): treat a blocked DM send as terminal in the queue drain (#186)

mbradley:
## What this changes

When a protected minor is blocked from sending a DM (the #176 restriction), the app's outbound-message queue treated that block like a temporary network failure: it kept the message queued and retried it over and over, even though the block is permanent and every retry is refused. The message ended up stuck as "failed" forever.

This makes a blocked send final: the queue drops the message instead of retrying it.

## Why it's safe

A blocked message can never reach anyone. Both the one-to-one and group send paths already check the policy before a message is ever queued, so a message only ends up blocked-on-retry when it was queued while sending was allowed and the policy changed afterward (the sender became a protected minor, or a recipient's approval was revoked). The send is re-checked and re-blocked on every attempt regardless. This is a reliability cleanup, not a security fix.

## Where

- `recoverFullSend` (`mobile/packages/dm_repository/`): on a blocked result, delete the queued message (the same thing the user's "Cancel send" does) instead of marking it as a retryable failure.
- The retry sweep (`mobile/lib/services/outgoing_dm_retry_service.dart`): both places it drains the queue now treat a blocked result as final, so neither keeps bumping the retry counter on a message that's already gone.

## Related

Relates to divinevideo/support-trust-safety#186 (part of epic #173).

## Verification

- New tests in `dm_repository` (a blocked message is deleted and nothing is written to the chat history; a delete failure is handled without crashing) and in the retry service (both drain paths treat a blocked send as final, with no retry). Each was checked to fail if the fix is removed.
- Full `dm_repository` suite green; that package's line coverage is 93.42%, above its 93% floor. No database migration, no generated code. `flutter analyze` clean.

## Type of Change

### DIFF
--- a/docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
+++ b/docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
@@ -27,24 +27,25 @@ ever delivered to a non-approved recipient. This is purely a reliability /
 efficiency wart: wasted retry round-trips plus a permanently-`failed` row that
 never resolves.
 
-`recoverFullSend` is the single drain entry for both 1:1 and group rows (the
-retry sweep funnels every `recipientWrapStatus == failed` row through it), so
-there is one fix site.
+`recoverFullSend` is the drain entry for under-budget failed rows and stale
+pending rows in both 1:1 and group sends. Rows that already exhausted their
+retry budget remain outside these queries and are not cleaned up by this PR.
 
 ## Decision
 
 **Delete the row** on a blocked result, rather than adding a distinct
 non-retriable `OutgoingWrapStatus.blocked` state. Deleting reuses the existing
-`deleteById` (already the terminal path the user's "Cancel send" uses), needs
+`deleteById` (already the terminal path after full delivery), needs
 no Drift schema migration (the `_parseStatus` fail-loud downgrade contract
 makes a new enum value a heavier, higher-blast-radius change), and the failed
-bubble simply disappears, cancel-like. A policy block is a permanent decision,
-not a transient error, and this is a no-safety-impact wart, so the proportionate
-fix is to drop the doomed row.
+bubble simply disappears. A confirmed policy block is terminal, not a transient
+error, and this is a no-safety-impact wart, so the proportionate fix is to drop
+the doomed row. Fail-closed denials caused only by unresolved/loading account
+state are ordinary retryable failures and keep the queue row.
 
 ## Design
 
-Two files, one logical fix ("blocked is terminal" end-to-end):
+One logical fix ("confirmed blocked is terminal" end-to-end):
 
 **1. `dm_repository.dart` — `recoverFullSend` else branch.** Branch on
 `result.blocked` before the transient-failure path:
@@ -64,7 +65,9 @@ Two files, one logical fix ("blocked is terminal" end-to-end):
 `_finalizeAfterRecipientBlocked` deletes via `deleteById`, logs, and swallows /
 reports on error without rethrowing — matching `_finalizeAfterRecipientFailure`'s
 non-rethrow posture (the caller still returns the blocked result). If the delete
-somehow fails, the row stays `failed` and self-heals on a later sweep.
+somehow fails, the row stays `failed` or `pending` and is retried on a later
+sweep. A persistent delete failure can therefore re-run the gate each sweep;
+that implies a broken database and does not publish any wrap.
 
 **2. `outgoing_dm_retry_service.dart` — State B dispatch.** After
 `recoverFullSend`, the current `else` calls `incrementRetry` and counts
@@ -78,10 +81,12 @@ failed. This keeps "blocked is terminal" true end-to-end.
 
 - **dm_repository:** a queued row (`recipientWrapStatus == failed`) whose
   recipient is now non-approved, with `sendRumor` stubbed to return
-  `NIP17SendResult.blocked(...)` → `recoverFullSend` deletes the row
-  (`getById` returns `null`), does not mark it `failed`, and returns a blocked
-  result. Confirm the row is absent from `getRetryableForOwner` afterward (no
-  loop).
+  `NIP17SendResult.blocked(...)` → `recoverFullSend` calls `deleteById`, does
+  not mark it `failed`, and returns a blocked result. DAO deletion semantics
+  are covered by the DAO suite.
+- **Unknown-state regression:** a fail-closed denial with no trusted live or
+  persisted protected-minor verdict is classified as temporary; `sendRumor`
+  returns an ordinary failure so the queue row remains retryable.
 - **Regression guard:** a transient `NIP17SendResult.failure(...)` still marks
   both wraps `failed` and leaves the row retryable (unchanged path).
 - **outgoing_dm_retry_service:** a sweep over a row that drains to blocked
--- a/docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
+++ b/docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
@@ -1,0 +1,96 @@
+# Blocked DM send is terminal in the queue drain (mobile)
+
+Design for divinevideo/support-trust-safety#186, part of the protected-minor
+(13-15) safeguards epic #173. Surfaced during the #176 DM-restriction review.
+Low priority, **no safety impact**.
+
+## Problem (verified against code)
+
+The durable outbound-DM queue drain treats a policy-blocked send the same as a
+transient failure. In `DmRepository.recoverFullSend` (dm_repository package),
+`sendRumor` returns a `NIP17SendResult`; `result.success` is `false` for a
+`NIP17SendResult.blocked(...)` (blocked is a `NIP17SendFailure` subtype), so it
+falls into the `else` branch → `_finalizeAfterRecipientFailure` → both wraps
+marked `failed`. The retry sweep (`OutgoingDmRetryService`, State B) then
+re-attempts the row on every pass until `retryCount` reaches `maxRetries`,
+after which the row lingers permanently as a dead `failed` bubble.
+
+### Why it is not a safety issue
+
+Both `sendMessage` and `sendGroupMessage` pre-gate at `canSendTo` and return
+`.blocked` **before enqueuing**, so a blocked row never originates at initial
+send. A blocked row only arises when a row was enqueued while sending was
+allowed, then the policy flipped before the drain replayed it (sender became a
+protected minor, or an official recipient's approval was revoked). On replay,
+`sendRumor`'s choke point re-evaluates the policy and re-blocks — nothing is
+ever delivered to a non-approved recipient. This is purely a reliability /
+efficiency wart: wasted retry round-trips plus a permanently-`failed` row that
+never resolves.
+
+`recoverFullSend` is the single drain entry for both 1:1 and group rows (the
+retry sweep funnels every `recipientWrapStatus == failed` row through it), so
+there is one fix site.
+
+## Decision
+
+**Delete the row** on a blocked result, rather than adding a distinct
+non-retriable `OutgoingWrapStatus.blocked` state. Deleting reuses the existing
+`deleteById` (already the terminal path the user's "Cancel send" uses), needs
+no Drift schema migration (the `_parseStatus` fail-loud downgrade contract
+makes a new enum value a heavier, higher-blast-radius change), and the failed
+bubble simply disappears, cancel-like. A policy block is a permanent decision,
+not a transient error, and this is a no-safety-impact wart, so the proportionate
+fix is to drop the doomed row.
+
+## Design
+
+Two files, one logical fix ("blocked is terminal" end-to-end):
+
+**1. `dm_repository.dart` — `recoverFullSend` else branch.** Branch on
+`result.blocked` before the transient-failure path:
+
+```dart
+} else if (result.blocked) {
+  await _finalizeAfterRecipientBlocked(outgoingDao: dao, rumorId: rumorId);
+} else {
+  await _finalizeAfterRecipientFailure(
+    outgoingDao: dao,
+    rumorId: rumorId,
+    errorMessage: result.error ?? 'Unknown publish failure',
+  );
+}
+```
+
+`_finalizeAfterRecipientBlocked` deletes via `deleteById`, logs, and swallows /
+reports on error without rethrowing — matching `_finalizeAfterRecipientFailure`'s
+non-rethrow posture (the caller still returns the blocked result). If the delete
+somehow fails, the row stays `failed` and self-heals on a later sweep.
+
+**2. `outgoing_dm_retry_service.dart` — State B dispatch.** After
+`recoverFullSend`, the current `else` calls `incrementRetry` and counts
+`failedFullSend++`. A deleted row makes `incrementRetry` a harmless no-op
+(returns `false` on a missing row), but it is a wasted DB call and mis-counts a
+terminal block as a failure. Add an `else if (result.blocked)` that treats the
+row as terminal: skip `incrementRetry`, and count it as terminal rather than
+failed. This keeps "blocked is terminal" true end-to-end.
+
+## Test plan
+
+- **dm_repository:** a queued row (`recipientWrapStatus == failed`) whose
+  recipient is now non-approved, with `sendRumor` stubbed to return
+  `NIP17SendResult.blocked(...)` → `recoverFullSend` deletes the row
+  (`getById` returns `null`), does not mark it `failed`, and returns a blocked
+  result. Confirm the row is absent from `getRetryableForOwner` afterward (no
+  loop).
+- **Regression guard:** a transient `NIP17SendResult.failure(...)` still marks
+  both wraps `failed` and leaves the row retryable (unchanged path).
+- **outgoing_dm_retry_service:** a sweep over a row that drains to blocked
+  drops the row and does not call `incrementRetry` / count it as a retry
+  failure.
+
+## Out of scope
+
+- The initial-send race where a policy flip mid-`sendGroupMessage` loop yields a
+  per-recipient blocked result that gets marked retryable: vanishingly narrow,
+  and it self-heals through the drain fix above on the next sweep.
+- #185 (foreground-resume account-state refetch) — its own PR.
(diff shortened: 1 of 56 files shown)


mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -3193,6 +3193,11 @@ class DmRepository {
         // surfaces only as a sender-side gap until the self-wrap
         // arrives via the receive pipeline.
       }
+    } else if (result.blocked) {
+      // A #176 policy block is a permanent decision, not a transient error:
+      // drop the row so the retry sweep stops re-attempting a send the gate
+      // refuses every time. The recipient never received anything.
+      await _finalizeAfterRecipientBlocked(outgoingDao: dao, rumorId: rumorId);
```
realmeylisdev, line 3200: **Potential data-loss on a transient (not permanent) block.** The comment above says "a #176 policy block is a permanent decision," but the injected `DmSendPolicy` is not permanent: `isDmRestrictedProvider` fails closed (`protected_minor_providers.dart:176` → `store.lastKnownFor(pubkey) ?? true`) for any account with no persisted sticky verdict while `protectedMinorStatusProvider` is still `AsyncLoading`. The cold-start sweep is gated only on Nostr readiness (`social_providers.dart:180-183`, `fireImmediately: true`), so it can fire and win the race against the keycast HTTP fetch. In that window a legitimately-queued adult DM evaluates `blocked` and `_finalizeAfterRecipientBlocked` deletes it irreversibly.

Pre-fix this same transient mis-verdict was benign — the row was marked `failed` and delivered on a later sweep once keycast resolved. Delete does not self-heal, so this is a *new* permanent-loss path.

The common case is protected: the sticky store is read synchronously at cold start, so any account that ever resolved `notProtected` is safe from frame 0, and a fresh-install adult is pre-gate-blocked from enqueuing during the fail-closed window (so never creates the vulnerable row). The realistic trigger is therefore narrow — the fire-and-forget sticky write not flushing before an app-kill, or a per-user keycast outage spanning enqueue → cold-start. Given it's irreversible, consider only treating `blocked` as terminal when the restriction verdict is *trusted/positive*, and falling back to the existing mark-failed path when the verdict is the `?? true` fallback.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -3193,6 +3193,11 @@ class DmRepository {
         // surfaces only as a sender-side gap until the self-wrap
         // arrives via the receive pipeline.
       }
+    } else if (result.blocked) {
+      // A #176 policy block is a permanent decision, not a transient error:
+      // drop the row so the retry sweep stops re-attempting a send the gate
+      // refuses every time. The recipient never received anything.
```
realmeylisdev: Nit: "The recipient never received anything" is guaranteed for *this* attempt (the gate returns before any wrap build/publish), but not for the row's history. A `recipientWrapStatus == failed` row can exist precisely because an earlier publish landed and only the OK/local-persist was lost — the retry service's own State B comment says so (`outgoing_dm_retry_service.dart:247-251`: "re-publishing is safe even when the original publish actually landed but the local persist failed"). Deleting is still the right call (the gate forbids re-delivery), but something like "this attempt delivered nothing" would be accurate.

mobile/packages/dm_repository/lib/src/dm_repository_reportable_sites.dart
```diff
@@ -37,6 +37,12 @@ abstract class DmRepositoryReportableSites {
   static const String finalizeAfterRecipientFailure =
       'finalizeAfterRecipientFailure';
 
+  /// `_finalizeAfterRecipientBlocked`: deleting the terminally-blocked
+  /// queue row threw. Caller already has the blocked result; the row
+  /// stays failed and self-heals on a later sweep.
```
realmeylisdev: Nit (self-contradicting within this PR): this triage comment says a failed delete leaves the row "stays failed and self-heals on a later sweep," but `_finalizeAfterRecipientBlocked`'s own doc added in the same diff says "still `failed` or `pending`, depending on the drain arm." The interrupted arm feeds from `getStillPendingForOwner` (pending-only) and `recoverFullSend` writes no status before deleting, so there the row stays `pending`, not `failed`. Since this is the string a triager reads when the Crashlytics site fires, align it with the method doc ("stays failed or pending") so they don't look only at the failed-row filter.

mobile/lib/services/outgoing_dm_retry_service.dart
```diff
@@ -259,6 +260,12 @@ class OutgoingDmRetryService {
               // way the row exits the recipient-failed filter, so
               // skip incrementRetry.
               processedFullSend++;
+            } else if (result.blocked) {
+              // recoverFullSend deleted the row: a #176 policy block is
+              // terminal, not transient. Don't bump the retry counter (the
+              // row is gone and the gate would refuse every retry anyway),
+              // and count it as terminal rather than a failure.
+              blockedFullSend++;
```
realmeylisdev, line 268: Info (contrived, no change demanded): the blocked branch intentionally skips `incrementRetry`, and `_finalizeAfterRecipientBlocked` writes nothing when `deleteById` throws (no `lastAttemptAt`, no status change) — unlike `_finalizeAfterRecipientFailure`, whose `markXxxWrapStatus` refreshes `lastAttemptAt`. So if `deleteById` fails *persistently* while reads still succeed, the row is byte-identical each sweep: the backoff gate (`:170-178`) always passes and `retryCount` never reaches `maxRetries`, so it re-runs the policy gate every sweep with no backoff and no budget exit. Loop cost is just the gate (returns before any wrap build), and a persistent delete-failure-with-working-reads implies a broken DB, so impact is low — flagging only because the design doc's "self-heals on a later sweep" doesn't note the removed cap for the persistent case.

mobile/packages/dm_repository/test/src/dm_repository_test.dart
```diff
@@ -14135,6 +14135,111 @@ void main() {
...
+
+          final result = await repository.recoverFullSend(
+            rumorId: _rumorEventId,
+          );
+
+          // The delete failed, but the caller still gets the blocked result
+          // and the drain does not throw; the still-present row is re-blocked
+          // and re-dropped on the next sweep.
+          expect(result.blocked, isTrue);
+          verify(() => mockOutgoingDmsDao.deleteById(_rumorEventId)).called(1);
```
realmeylisdev, line 14239: Should-fix (can-it-fail): this delete-throws test asserts only `result.blocked` and `deleteById.called(1)` — it never pins the reporter-port call. If the `_errorReporter?.call(e, stackTrace, site: finalizeAfterRecipientBlocked)` at `dm_repository.dart:3421-3427` were deleted or given the wrong site, every test in the repo stays green (`grep -rn finalizeAfterRecipientBlocked test/` → 0 hits). The exact in-file precedent for the sibling swallow does pin it — `dm_repository_test.dart:12817-12823` asserts `reporterCalls.single.site == finalizeAfterRecipientFailure`. `reporterCalls` is already wired via `createRepository` in this scope, so it's two lines: `expect(reporterCalls, hasLength(1)); expect(reporterCalls.single.site, equals(DmRepositoryReportableSites.finalizeAfterRecipientBlocked));`.

docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
```diff
@@ -0,0 +1,96 @@
...
+
+`recoverFullSend` is the single drain entry for both 1:1 and group rows (the
+retry sweep funnels every `recipientWrapStatus == failed` row through it), so
+there is one fix site.
+
+## Decision
+
+**Delete the row** on a blocked result, rather than adding a distinct
+non-retriable `OutgoingWrapStatus.blocked` state. Deleting reuses the existing
+`deleteById` (already the terminal path the user's "Cancel send" uses), needs
```
realmeylisdev: Nit (accuracy): "`deleteById` (already the terminal path the user's \"Cancel send\" uses)" — there is no user-facing Cancel-send. `OutgoingDmsDao.deleteById` has no app-layer caller (the two `mobile/lib` `deleteById` hits are on different DAOs — `pending_view_events`, `pending_product_events`); `conversation_event.dart` has no cancel event (`ConversationMessageDeleted` is NIP-09 kind-5 for already-sent messages). The only `outgoing_dms` `deleteById` callers are internal: the success paths (`recoverSelfWrap`, `_finalizeAfterRecipientSuccess`) and this new blocked path. The decision to delete is still sound — the accurate precedent to cite is the full-delivery success path. (The claim traces to a stale DAO comment at `outgoing_dms_dao.dart:356-357` from #3911 and `conversation_state.dart:139` — both pre-existing and out of scope here, but the new doc repeats the fiction.)

docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
```diff
@@ -0,0 +1,96 @@
...
+send. A blocked row only arises when a row was enqueued while sending was
+allowed, then the policy flipped before the drain replayed it (sender became a
+protected minor, or an official recipient's approval was revoked). On replay,
+`sendRumor`'s choke point re-evaluates the policy and re-blocks — nothing is
+ever delivered to a non-approved recipient. This is purely a reliability /
+efficiency wart: wasted retry round-trips plus a permanently-`failed` row that
+never resolves.
+
+`recoverFullSend` is the single drain entry for both 1:1 and group rows (the
+retry sweep funnels every `recipientWrapStatus == failed` row through it), so
```
realmeylisdev: Nit (overstated): "the retry sweep funnels every `recipientWrapStatus == failed` row through it" (and later "the failed bubble simply disappears") isn't true for rows that already exhausted the retry budget. `getRetryableForOwner` filters `retryCount < maxRetries` (`outgoing_dms_dao.dart:432`) and `getStillPendingForOwner` matches only `pending` wraps, so a `failed`+exhausted row reaches neither query and never hits the new blocked-delete path. Consequence the doc's problem statement claims to fix but doesn't: (a) blocked rows that burned their budget between #5754 (2026-07-09) and this ship, and (b) post-fix rows that exhaust retries on transient failures before the policy flips — both remain permanent dead bubbles with no drain, no delete, and (per the note above) no cancel affordance. Worth scoping the claim to under-budget rows or listing this residual in Out-of-scope.

docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
```diff
@@ -0,0 +1,96 @@
...
+terminal block as a failure. Add an `else if (result.blocked)` that treats the
+row as terminal: skip `incrementRetry`, and count it as terminal rather than
+failed. This keeps "blocked is terminal" true end-to-end.
+
+## Test plan
+
+- **dm_repository:** a queued row (`recipientWrapStatus == failed`) whose
+  recipient is now non-approved, with `sendRumor` stubbed to return
+  `NIP17SendResult.blocked(...)` → `recoverFullSend` deletes the row
+  (`getById` returns `null`), does not mark it `failed`, and returns a blocked
```
realmeylisdev: Info: the test plan describes state-based checks ("`getById` returns `null`", "Confirm the row is absent from `getRetryableForOwner` afterward") that the shipped tests don't implement — they're interaction-based against a mocked DAO (`verify(deleteById).called(1)` + `verifyNever(...)`; `getById` stays stubbed to return the row). The behavior is adequately pinned by the `deleteById` verification, but the wording promises assertions that aren't there — soften it to the interaction-level checks actually written, or note the state-level check is covered transitively by the DAO's own `deleteById` tests.

realmeylisdev (commented):
## Review — fix(dm): treat a blocked DM send as terminal in the queue drain (#186)

The core change is correct. `NIP17SendResult.blocked` is genuinely non-retriable, `recoverFullSend`'s `success` → `blocked` → `else` ordering is exhaustive and reachable, both pre-enqueue gates already exist, and `sendRumor` re-gates every attempt so nothing is ever delivered to a non-approved recipient. Both sweep arms and the interrupted arm now stop bumping the retry counter. CI is green; `dm_repository` passes at 93.42% (floor 93); the retry-service suite passes.

One edge worth an explicit decision before merge, then some should-fix/doc cleanups.

**Worth deciding — a new permanent-loss path:** delete-on-blocked turns a *transient* fail-closed verdict into irreversible deletion of a legitimately-queued DM. `isDmRestrictedProvider` fails closed for any account with no persisted sticky verdict while `protectedMinorStatusProvider` is still `AsyncLoading`, and the cold-start sweep is gated only on Nostr readiness (`social_providers.dart:180`), so it can win the race against the keycast fetch. Pre-fix, a false block marked the row `failed` and it self-healed once the verdict resolved — delete does not self-heal. The common case is protected (sticky is read synchronously at cold start, and a fresh-install adult is pre-gate-blocked from ever enqueuing during the fail-closed window), so the realistic trigger is narrow. But the PR's "why it's safe" section doesn't enumerate it. See the inline note on `recoverFullSend`.

**Should-fix**
- `retryPendingCollaboratorInvites` (the third `recoverFullSend` caller, `dm_repository.dart:3283-3293`) still treats `blocked` as a generic failure: it's counted into `failureCount` → shown as "N invites remaining" in the banner even though the row was just deleted, and a concurrent-delete `ArgumentError` is Crashlytics-reported via `retryPendingCollaboratorInviteUnexpectedThrow` (an expected race, matrix-NO). "blocked is terminal" isn't yet end-to-end across all three callers — a matching `else if (result.blocked)` arm plus a blocked-path test in `collaborator_invite_recovery_test.dart` would close it.
- The delete-failure test doesn't pin the reporter-port call — see inline.

**Nits / docs** are inline (a fabricated "Cancel send" precedent, a self-contradicting "stays failed" comment, an overstated "recipient never received anything", and two over-claims in the design doc).

No blocker in the shipped code path; the items above are the author's call.

mobile/packages/dm_repository/lib/src/dm_repository.dart
rabble: Fixed in edd344562. The send policy now distinguishes unresolved fail-closed state from a confirmed protected-minor block; unresolved state returns an ordinary retryable failure, so recoverFullSend retains the queue row. Added provider and message-service regressions. (Response by Codex on behalf of @rabble)

mobile/packages/dm_repository/lib/src/dm_repository.dart
rabble: Fixed in edd344562. The comment now says this attempt delivered nothing, without making a claim about prior attempts. (Response by Codex on behalf of @rabble)

mobile/packages/dm_repository/lib/src/dm_repository_reportable_sites.dart
rabble: Fixed in edd344562. The report-site documentation now says the row stays failed or pending, matching both drain arms. (Response by Codex on behalf of @rabble)

mobile/lib/services/outgoing_dm_retry_service.dart
rabble: Documented in edd344562. The design now records that a persistent delete failure re-runs the policy gate each sweep, implies a broken database, and cannot publish a wrap. (Response by Codex on behalf of @rabble)

mobile/packages/dm_repository/test/src/dm_repository_test.dart
rabble: Fixed in edd344562. The delete-throws regression now asserts the reporter call and exact finalizeAfterRecipientBlocked site. (Response by Codex on behalf of @rabble)

docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
rabble: Fixed in edd344562. The design cites deleteById as the full-delivery terminal path and no longer claims a user-facing cancel-send path. (Response by Codex on behalf of @rabble)

docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
rabble: Fixed in edd344562. The design now scopes the drain to under-budget failed and stale pending rows and records that exhausted rows remain outside this PR. (Response by Codex on behalf of @rabble)

docs/superpowers/specs/2026-07-10-dm-blocked-send-terminal-drain-design.md
rabble: Fixed in edd344562. The test plan now describes the interaction-level deleteById assertion and points to DAO coverage for deletion semantics. (Response by Codex on behalf of @rabble)

rabble (approved):
Reviewed and updated after the cold-start data-loss finding. Unresolved fail-closed account state now retains the queued send; only confirmed policy blocks terminalize it. Regression tests, full dm_repository suite, analyzer, and pre-push checks pass; all threads are answered/resolved. (Review by Codex on behalf of @rabble)

realmeylisdev:
`995a38d0f` closes the last open item from the review — the third `recoverFullSend` caller, `retryPendingCollaboratorInvites`, which still routed a `blocked` result through its generic failure branch.

It now:
- branches `blocked` into a separate `blockedCount` on `CollaboratorInviteRetrySummary` (kept out of `failureCount`), surfaced as `blockedInviteCount` on the banner state — so a just-deleted terminal row no longer reads as "still needs to send";
- skips the row-missing `ArgumentError` (a concurrent drain sweep removing the row between the banner read and the retry) as an expected race, without a Crashlytics report — mirroring `OutgoingDmRetryService`'s dispatch-error policy;
- shows a distinct `profileCollaboratorInviteBlockedResult` string ("N collaborator(s) can't receive invites") in both the banner and the publish snackbar instead of implying a pending retry.

"blocked is terminal" is now true end-to-end across all three `recoverFullSend` callers. New tests: blocked + concurrent-race cases in `collaborator_invite_recovery_test.dart`, plus `video_publish_provider_test.dart` and `profile_videos_grid_test.dart`; dm_repository suite, arb-consistency, and analyze all green.

The earlier red `Tests` job was an unrelated wall-clock flake in `video_recorder_bloc_test.dart` (its 800ms focus auto-hide timer racing the test's real 900ms delay — "which shorter than expected"). That file is byte-identical to `main`, this branch touches no video-recorder code, and it passed 8/8 locally; the fresh run should clear it.

PR #6034, opened 2026-07-11
fix(divine_ui): keep DivineButton size constant with or without border

### LINKED ISSUE #6035: fix(divine_ui): keep DivineButton size constant with or without border
Auto-created from PR #6034.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6034
**Branch:** `fix/divine-button-border-size`

hm21:
Closes #6035

## What

A `DivineButton` with a border (e.g. `secondary`) rendered larger than one without. A `base` secondary button was **52px** instead of **48px** tall (and 4px wider) — 2px extra on every edge. In figma BOTH have a size of 48px and not 52px.

## Why

The border lives in the `Ink` widget's `BoxDecoration`. Flutter adds the border's thickness (`border.dimensions`) as layout inset around the `Ink`'s child, so any bordered type inflates by the stroke width (2px) on each side. With the default `strokeAlignInside`, the stroke is meant to sit *inside* the bounds — but the layout still grew, so bordered and unbordered buttons of the same size no longer matched.

Measured before the fix:
- primary (no border): `112.6 × 48.0`
- secondary (2px border): `116.6 × 52.0`

## Fix

Subtract the border width back out of the inner padding when a border is present (`_borderInset`). Net effect: the `Ink` adds the 2px border, the padding gives 2px back, so:

- **Outer size stays identical** across bordered/unbordered types of the same size (base = 48px, small visible = 40px, tiny = 32px — all independent of border).
- **Content position is unchanged** — the stroke sits in the outer 2px ring of the padding (inside the bounds) and never overlaps the label/icon.

The magic `2` is also extracted into a single `_kBorderWidth` constant used for both the stroke and the compensation.

## Tests

- New regression test: a bordered (`secondary`) button has the exact same outer `Size` as an unbordered one, and `base` height is 48px.
- All 136 `divine_ui` button tests pass; `divine_ui` 100% coverage gate satisfied (no uncovered lines in the changed file).

### DIFF
--- a/mobile/packages/divine_ui/lib/src/button/divine_button.dart
+++ b/mobile/packages/divine_ui/lib/src/button/divine_button.dart
@@ -186,6 +186,9 @@ class _DivineButtonContent extends StatelessWidget {
   final bool expanded;
   final bool isLoading;
 
+  /// Thickness of the visible border on bordered [type]s (e.g. secondary).
+  static const double _kBorderWidth = 2;
+
   bool get _isEnabled => onPressed != null && !isLoading;
 
   /// Icon size scales with [size]: tiny uses a smaller 20px icon to fit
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Approved after independent re-review. The button border compensation is scoped, regression-tested, and CI is green.

PR #5935, opened 2026-07-07
fix(feed): prefer hls for divine blob playback

rabble:
Closes #5936

## Summary
- prefer the canonical HLS rendition before raw MP4 for raw `media.divine.video/<hash>` playback URLs
- keep raw MP4 as fallback when HLS is unavailable
- update playback source tests to pin the source order

## Root Cause
The reported logs show the feed opening `https://media.divine.video/e770667c1a62cc394602fc07462fa0d7ba83441002d9aac662fb88d0cc575338`, reporting the player ready, but playback position stays at `0` until stale detection tries the HLS fallback. The fallback then fails on the already-stalled controller with `COMPOSITION_ERROR: No playable video tracks found`, and the feed marks the video errored.

The asset itself is present: HEAD returns `200 video/mp4`, and `ffprobe` sees valid AAC + H.264 streams. The cold remote probe is slow and the first bytes show C2PA/JUMBF metadata before normal playback metadata, matching the user report that the first try lags/fails but works after the file is loaded/cached. HLS is the streamable rendition and should be the first feed source for canonical Divine blob URLs.

## Verification
- RED: `cd mobile/packages/infinite_video_feed && flutter test test/src/utils/playback_sources_test.dart --plain-name 'when URL is the raw blob URL'` failed with raw-before-HLS order
- GREEN: same command passed after the source-order change
- `cd mobile/packages/infinite_video_feed && flutter test test/src/utils/playback_sources_test.dart`
- `cd mobile/packages/infinite_video_feed && flutter test test/src/services/playback_source_registry_test.dart test/src/services/stale_playback_detector_test.dart`
- `cd mobile/packages/infinite_video_feed && flutter analyze lib test`
- `cd mobile/packages/infinite_video_feed && flutter test` (221 tests)
- `git diff --check`
- pre-push hook: analyzer passed; changed-file `packages/infinite_video_feed/test/src/utils/playback_sources_test.dart` passed 27 tests

PR #6036, opened 2026-07-11
fix(feed): prefer progressive mp4 over hls for short-video playback

### LINKED ISSUE #6038: fix(feed): prefer progressive mp4 over hls for short-video playback
Auto-created from PR #6036.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6036
**Branch:** `fix/feed-prefer-progressive-mp4`

hm21:
Closes #6038

## Problem

A tester reported video playback became suddenly choppy/stuttering after the last update. Logs (app `1.0.16+798`, iOS 26.5) show every raw-only Divine upload probing `<hash>/hls/master.m3u8` **first**, failing with `PlatformException(COMPOSITION_ERROR, No playable video tracks found)` (~0.4–0.5s wasted per video), then reloading from the raw blob. That per-video "fail HLS → reload" cycle is the stutter, most visible while scrolling.

**Root cause:** #5935 flipped the raw-blob source order to try HLS before the progressive blob. Fresh uploads publish the raw/progressive blob before their HLS rendition finishes transcoding, so a just-published video's HLS master is often a manifest with no playable video track — a guaranteed failure that stalls the currently-visible video. The tester's failing videos were only minutes old at watch time (a transcoding race); those same renditions are valid now.

The tester also switched Developer Options → MP4 with no effect, because `getOptimalVideoUrlForPlatform()` short-circuited on the raw-only heuristic **before** the format override.

## Why progressive MP4, never HLS-first

Our clips are ≤6.3s. HLS pays a fixed startup cost (master playlist → media playlist → first segment) that makes it slower to first frame than a single progressive MP4 request (moov at front, one round trip), and its multi-segment streams can't be single-file preloaded/cached. So progressive is always tried first; HLS is the last resort.

## Measured size per format (3 real uploads)

| video | raw (original) | 720p.mp4 | 480p.mp4 | raw ÷ 720p |
|---|---|---|---|---|
| `4af42277…` | 4.03 MB | 0.96 MB | 0.54 MB | 4.2× |
| `ee942f34…` | 6.23 MB | 2.19 MB | 0.93 MB | 2.9× |
| `b8dc7a27…` | 11.24 MB | 2.18 MB | 0.93 MB | 5.2× |

The raw original is 2.8–5.2× larger than the 720p.mp4, and the derivatives + a valid HLS master exist for all three. A raw-only `imeta` is just a publish-time snapshot (only the raw blob exists at publish); the derivatives appear seconds-to-minutes later and almost always exist by playback time.

## Changes

- **`playback_sources.dart`** — always attempt the progressive source first; for a quality variant the guaranteed raw blob is ordered **before** HLS, so a not-yet-transcoded `720p.mp4` falls to raw (progressive), not HLS.
- **`video_event_extensions.dart`** — raw-only uploads now default to `720p.mp4` (3–5× smaller) instead of the raw original; the runtime chain drops to raw when the variant isn't ready yet. Removed the now-dead `hasRawOnlyDivineImetaUrl` / `hasBareDivineHashPath` predicates. The Developer Options format override now runs ahead of every heuristic, so on-device A/B testing actually takes effect.
- **`source_loader.dart`** — an HTTP 202 (processing) source is only wait-and-retried when it is the last resort. A fresh upload whose `720p.mp4` is still transcoding now falls straight to the available raw blob instead of stalling up to ~19s on the derivative.

## Tests

- `playback_sources_test`: raw-blob and quality-variant ordering (progressive → raw → HLS).
- `video_event_divine_server_test`: raw-only imeta now resolves to `720p.mp4`; developer-override precedence (raw override wins over the default MP4; `mp4_480p` wins over classic Vine). Process-global format singleton reset in `tearDown`.
- `source_loader_test`: 202 advances immediately when a fallback remains; retries only on the last source.
- Green locally + pre-push: `playback_sources` + `source_loader` (40), `video_event_divine_server` (25), `mp4_hls_fallback_retry` (30); `flutter analyze` clean.

## Follow-up (server-side, not in this PR)

The HLS master for fresh uploads returns no video track until transcode completes. Worth fixing on divine-blossom/transcoding so the client fallback isn't the thing keeping playback alive.

### DIFF
--- a/mobile/lib/extensions/video_event_extensions.dart
+++ b/mobile/lib/extensions/video_event_extensions.dart
@@ -211,14 +211,18 @@ extension VideoEventAppExtensions on VideoEvent {
 
   /// Get the optimal video URL for initial playback.
   ///
-  /// **Strategy**:
+  /// **Strategy** (first match wins):
+  /// - A developer format override (Developer Options) forces that server
+  ///   format on any Divine video for A/B testing, ahead of every heuristic
+  ///   below.
   /// - Classic Vine originals use the raw blob directly (/{hash}) because the
   ///   source is already 480p or lower — transcoded variants are upscales at
   ///   best and may not exist, causing needless 404s.
+  /// - Raw-only uploads use their published raw blob URL, since derived
+  ///   MP4/HLS variants may not exist yet.
   /// - Other Divine videos default to progressive MP4 720p (faststart, moov at
   ///   front) for fastest startup with short videos (1 request, no manifest
   ///   overhead).
-  /// - Developer options can override to HLS or other formats for A/B testing.
   /// - If MP4 fails (e.g. not yet transcoded after upload), [getFallbackUrl]
   ///   provides an HLS fallback via the quality variant error handler.
   ///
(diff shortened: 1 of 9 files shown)


NotThatKindOfDrLiz (approved):
Reviewed playback source ordering, fallback behavior, HTTP 202 retry behavior, and targeted tests. No blocking findings.

PR #5040, opened 2026-06-10
fix(lists): make curated list creation reliable and visible

rabble:
## Description

A user reported "I tried to create a list and it just failed." Log + code analysis showed list creation was broken end to end:

1. **No create entry point** — the Explore Lists tab only had "Discover Lists"; the sole way to create a video list was buried in a video's share menu. Users flailed at the screen (the exported logs show eight Subscribe/Unsubscribe toggles on the user's own empty default list).
2. **Silent failure in `CreateListDialog`** — `CuratedListService.createList` catches its own exceptions and returns `null`, and the dialog did nothing on `null`: no snackbar, no pop, it just sat there.
3. **Empty lists never published** — `_publishListToNostr` skipped lists with no videos, so a freshly created list existed only on one device (relay query for the reporting user's kind 30005 events confirmed zero).
4. **Publish failures swallowed** — only `PublishSuccess` was handled; `PublishNoRelays`/`PublishFailed` produced no log and no feedback.
5. **Published lists vanished from "My Lists"** — the section filtered on `nostrEventId == null`, so the moment a list reached the relay it disappeared from the UI.

Changes:

- Add a **Create New List** button to the Explore Lists tab; `CreateListDialog.video` is now optional.
- On `createList == null`, show the `listCreateFailed` snackbar and keep the dialog open so the typed name/description survive a retry.
- Publish empty public lists on creation; log a warning with `failureReason` when signing or publishing fails; `_publishListToNostr` now returns a success bool.
- New `CuratedListService.myLists` getter (owned-by-pubkey OR not-yet-published local lists) backs the "My Lists" section, so lists stay visible after publishing.
- Hide the Subscribe (+) app-bar button on lists the current user owns — owners get the overflow menu (delete) only.

**Related Issue:** None (user log report `openvine_full_logs_2026-06-10T11-52-21`)

## Out of Scope

- The pre-existing possibility that a subscribed own list shows in both "My Lists" and "Subscribed lists" (unchanged behavior).
- Migrating `CuratedListService` to the BLoC/repository architecture.

## Verification

- `flutter analyze lib test integration_test` — no issues
- `flutter test test/services/curated_list_service_*` — 186 passing
- `flutter test test/widgets/add_to_list_dialog_test.dart test/screens/curated_list_feed_screen_test.dart` — 25 passing (3 new dialog tests, 2 new app-bar tests, 4 new `myLists` tests, publish-on-create + failure-path service tests)
[...]

PR #6041, opened 2026-07-11
test(curated-list): re-enable three FIXME-marked list service tests

### LINKED ISSUE #6042: test(curated-list): re-enable three FIXME-marked list service tests
Auto-created from PR #6041.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6041
**Branch:** `fix/curated-list-test-isolation`

hm21:
Closes #6042

## What

Removes stale `// FIXME` and `// TODO` markers by re-enabling the curated-list service tests they were attached to, then shrinks the skip baseline to lock in the win.

Re-enabled:
- `curated_list_service_query_test.dart` -> `searchLists() finds lists by name`
- `curated_list_service_query_test.dart` -> `searchLists() finds lists by tags`
- `curated_list_service_query_test.dart` -> `search handles special characters`
- `curated_list_service_query_test.dart` -> `search handles unicode characters`
- `curated_list_service_query_test.dart` -> `search performance with many lists`
- `curated_list_service_persistence_test.dart` -> `saves multiple lists to SharedPreferences`
- `curated_list_service_persistence_test.dart` -> `handles concurrent save operations`

Also removed both affected files from `mobile/scripts/baseline/skip_tests.txt`.

## Why

These skips were stale. The underlying race/isolation issues have since been fixed in production/setup code:

- **Timestamp-ID collision**: `CuratedListService._generateListId` now suffixes `_1`, `_2`, ... when two lists are created in the same millisecond, so concurrent creates no longer overwrite each other.
- **Test isolation**: `setUp` calls `SharedPreferences.setMockInitialValues({})`, giving every test a clean prefs store before constructing a fresh `CuratedListService`.
- **Search edge cases**: the special-character and unicode search tests now pass against the current `searchLists` implementation.

## Verification

- `flutter analyze test/services/curated_list_service_query_test.dart test/services/curated_list_service_persistence_test.dart`
- `flutter test test/services/curated_list_service_query_test.dart test/services/curated_list_service_persistence_test.dart` -> 47 pass
- `bash scripts/check_skip_ceiling.sh`
- Pre-push hook reran analyzer and changed-file tests successfully.

### DIFF
--- a/mobile/test/services/curated_list_service_persistence_test.dart
+++ b/mobile/test/services/curated_list_service_persistence_test.dart
@@ -109,9 +109,7 @@ void main() {
         expect(savedData, contains('List 1'));
         expect(savedData, contains('List 2'));
         expect(savedData, contains('List 3'));
-        // TODO(Any): Fix and re-enable these tests
-        // Fails only when running the entire test suite
-      }, skip: true);
+      });
 
       test('updates SharedPreferences after list modification', () async {
         final service = CuratedListService(
--- a/mobile/test/services/curated_list_service_persistence_test.dart
+++ b/mobile/test/services/curated_list_service_persistence_test.dart
@@ -340,31 +340,24 @@ void main() {
         expect(service2.lists.length, greaterThanOrEqualTo(3));
       });
 
-      test(
-        'handles concurrent save operations',
-        () async {
-          // FIXME: Flaky test - race condition with timestamp-based IDs
-          final service = CuratedListService(
-            nostrService: mockNostr,
-            authService: mockAuth,
-            prefs: prefs,
-          );
+      test('handles concurrent save operations', () async {
+        final service = CuratedListService(
+          nostrService: mockNostr,
+          authService: mockAuth,
+          prefs: prefs,
+        );
 
-          // Create multiple lists concurrently
-          await Future.wait([
-            service.createList(name: 'Concurrent 1'),
-            service.createList(name: 'Concurrent 2'),
-            service.createList(name: 'Concurrent 3'),
-          ]);
+        // Create multiple lists concurrently
+        await Future.wait([
+          service.createList(name: 'Concurrent 1'),
+          service.createList(name: 'Concurrent 2'),
+          service.createList(name: 'Concurrent 3'),
+        ]);
 
-          // At least some lists should be saved
-          final savedData = prefs.getString(CuratedListService.listsStorageKey);
-          expect(savedData, isNotNull);
-          // Race condition: lists with same timestamp may overwrite
-          // each other
-        },
-        skip: 'Flaky: timestamp-based ID collision in concurrent creation',
-      );
+        // At least some lists should be saved
+        final savedData = prefs.getString(CuratedListService.listsStorageKey);
+        expect(savedData, isNotNull);
+      });
 
       test('preserves timestamps across save/load', () async {
         final service1 = CuratedListService(
--- a/mobile/test/services/curated_list_service_query_test.dart
+++ b/mobile/test/services/curated_list_service_query_test.dart
@@ -93,7 +93,6 @@ void main() {
 
     group('searchLists()', () {
       test('finds lists by name', () async {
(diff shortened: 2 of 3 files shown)


NotThatKindOfDrLiz:
Takeover update:

- Re-enabled the remaining four stale curated-list service skips in the same touched area:
  - `searchLists() finds lists by tags`
  - `search handles special characters`
  - `search handles unicode characters`
  - `saves multiple lists to SharedPreferences`
- Shrunk `mobile/scripts/baseline/skip_tests.txt` so the skip-ceiling ratchet locks in both curated-list test files.
- Updated the PR body to reflect the full scope.

Verification:
- `flutter analyze test/services/curated_list_service_query_test.dart test/services/curated_list_service_persistence_test.dart`
- `flutter test test/services/curated_list_service_query_test.dart test/services/curated_list_service_persistence_test.dart` -> 47 pass
- `bash scripts/check_skip_ceiling.sh`
- Pre-push hook reran analyzer and changed-file tests successfully.
- GitHub checks are green on `cccc102366fc4073e430644db3804bae41dad23a`.

NotThatKindOfDrLiz (approved):
Approved after takeover. The remaining stale curated-list skips are re-enabled, the skip baseline is shrunk, the PR body is accurate, and CI is green on the latest head.

mobile/test/services/curated_list_service_persistence_test.dart
```diff
@@ -340,31 +338,24 @@ void main() {
...
+        // Create multiple lists concurrently
+        await Future.wait([
+          service.createList(name: 'Concurrent 1'),
+          service.createList(name: 'Concurrent 2'),
+          service.createList(name: 'Concurrent 3'),
+        ]);
+
+        // At least some lists should be saved
+        final savedData = prefs.getString(CuratedListService.listsStorageKey);
+        expect(savedData, isNotNull);
```
realmeylisdev, line 357: The only assertion here is `expect(savedData, isNotNull)` — which is exactly what the old ID-collision skip guarded, but it doesn't exercise the fix. `savedData` stays non-null even if two lists collided and overwrote each other, so this can't fail for the reason it exists.

With `_generateListId` now suffixing `_1/_2`, the three `createList` calls each run synchronously up to their first `await` (adding to `_lists` before suspending), so all three get unique IDs deterministically. Worth asserting that, e.g. `expect(service.lists.length, 3)` or that `savedData` contains all three names, so the test actually guards the collision fix it was skipped for.

realmeylisdev (commented):
Formal review — test-only change, no production code, so NIP-51 / kind-30005 behavior is unchanged.

I checked that the re-enabled tests are actually stable, not green-once:
- 12/12 standalone runs of both files with `--test-randomize-ordering-seed random`
- 3/3 runs bundled under `very_good test` (the merged-isolate harness), random seeds
- one full `test/services` merged-isolate run: 2986 passed / 0 failed (random seed 2094)
- `check_skip_ceiling.sh` green; the `_generateListId` `_1/_2` suffix (#5040) is confirmed active

Two non-blocking test-quality notes inline. One PR-body nit: the isolation `setMockInitialValues({})` call has actually been in `setUp` since the files were created (8763db104) — it predates the skips, so only `_generateListId` (#5040) is a genuine "since fixed" change; the isolation tests read as over-skipped rather than newly-fixed. Not a blocker. Looks good to merge.

mobile/test/services/curated_list_service_query_test.dart
```diff
@@ -466,8 +457,7 @@ void main() {
           greaterThanOrEqualTo(25),
         ); // Should find at least 25
         expect(stopwatch.elapsedMilliseconds, lessThan(100)); // Should be fast
```
realmeylisdev, line 459: `expect(stopwatch.elapsedMilliseconds, lessThan(100))` is a wall-clock assertion. Low-risk here since it wraps only the synchronous 50-item `searchLists` filter (microseconds vs a 100 ms bound), but timing assertions have flaked in this repo under the merged-isolate harness and this one guards little at 50 items. Optional: drop the timing check or widen the bound.

PR #6043, opened 2026-07-11
fix(list): show a single back button in curated list video mode

### LINKED ISSUE #6044: fix(list): show a single back button in curated list video mode
Auto-created from PR #6043.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6043
**Branch:** `fix/curated-list-double-back-button`

hm21:
Closes #6044

## What

In a curated list, tapping a video opens a fullscreen player. Users saw **two identical back arrows** stacked in the top-left of that video view.

The screen embeds `PooledFullscreenVideoFeedScreen` (which renders its own `DiVineAppBar` back button that pops the whole route) and additionally layered a custom back-arrow overlay on top (which returned to the grid). Two buttons, two different behaviours, same icon in the same spot.

image

## Fix

- Add an optional `onBack` override to `PooledFullscreenVideoFeedScreen` / `FullscreenFeedContent`. When provided, the feed's own app-bar back button calls it instead of popping the route. Backward compatible — every existing caller keeps the pop-the-route default.
- `CuratedListFeedScreen` now passes `onBack: () => setState(() => _activeVideoIndex = null)` and drops the custom overlay entirely.

Result: a single back button (the standard transparent app bar, which also keeps the list title + settings menu, consistent with every other feed surface), and the intended two-level back is preserved: video → grid → list.

## Why keep return-to-grid instead of just deleting the overlay

This screen is deliberately a two-mode (grid / video) screen, and all three embedded-feed screens keep a "back to parent grid" affordance. So the overlay's *behaviour* was the correct one; the fullscreen feed's own route-popping back button was the redundant one in this embedded context. The fix removes the redundant button while preserving the drill-in navigation model.

## Scope note

`user_list_people_screen.dart` (people lists) and `sound_detail_screen.dart` embed the same fullscreen feed with a similar overlay, but they use a distinct icon (grid / X), so they don't present as "two back buttons". Left untouched to keep this focused on the reported bug — happy to migrate them to the same `onBack` approach in a follow-up if wanted.

## Testing

- `flutter analyze` on both changed files — no issues.
- `flutter test test/screens/curated_list_feed_screen_test.dart` — 8/8 pass.
- `flutter test test/screens/feed/pooled_fullscreen_video_feed_screen_test.dart` + args test — pass (pre-push hook also re-ran these).

### DIFF
--- a/mobile/lib/screens/curated_list_feed_screen.dart
+++ b/mobile/lib/screens/curated_list_feed_screen.dart
@@ -230,6 +230,12 @@ class _CuratedListFeedScreenState extends ConsumerState<CuratedListFeedScreen> {
     );
   }
 
+  void _exitVideoMode() {
+    setState(() {
+      _activeVideoIndex = null;
+    });
+  }
+
   Widget _buildVideoPlayer(List<VideoEvent> videos) {
     if (videos.isEmpty || _activeVideoIndex! >= videos.length) {
       return Center(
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Approved. Focused fix; no blocking findings in re-review.

PR #5977, opened 2026-07-09
fix(dm): make DM reaction delivery durable (OK-confirm + auto-retry)

realmeylisdev:
## Description

DM reactions were being **silently lost to the recipient**. A friend reacted to a message in a 1:1 conversation, saw the reaction on their own device, but the recipient never received it.

**Root cause** (traced end-to-end; receive/render code is correct):
- A reaction's recipient-addressed gift wrap is published via `NostrClient.publishEvent`, which returns `PublishSuccess` on **frame-accept** — it never waits for the relay's NIP-20 `OK true`. On a flaky single relay (mid-reconnect), the write is counted as success even though the relay never stored the event.
- The sender's chip is inserted **optimistically before** any network send, so the sender always sees the reaction regardless of delivery — the loss is invisible.
- Unlike DM **messages** (durable `outgoing_dms` queue + `OutgoingDmRetryService` foreground sweep), reactions had **no automatic retry** — only a manual re-tap, which the false-positive "sent" never prompts.

This brings reactions to parity with the messages reliability model:

1. **OK-confirm the recipient wrap.** `NIP17MessageService.sendRumor` gains an opt-in `awaitRecipientOk` (default off — text/file/group message sends are byte-identical) that requires the relay's `OK true` via `publishEventAwaitOk` before a reaction counts as sent. `publish()` and `retry()` opt in; best-effort kind-5 supersede deletions stay on frame-accept. A 10s confirm window fits inside the reaction path's existing 15s outer cap.
2. **Durable auto-retry.** New `DmReactionRetryService` re-drives `failed`/interrupted-`pending` own reactions on every app-foreground transition via the existing `retry()`, reusing the `dm_message_reactions` row's stored rumor JSON — no schema migration. In-memory backoff (mirrors `OutgoingDmRetryConfig`: 5 tries, 2s→5min) + a min-age guard so an in-flight pending reaction is never re-driven; drops a reaction after `maxRetries` (manual re-tap still works).

Adds `DmReactionsDao.getRetryableOwnReactions` + `DmReactionsRepository.retryableReactions()` to project retryable rows for the sweep.

**Related Issue:** Closes #5978

## Out of Scope

- DM **messages** share the same frame-accept `publishEvent`; making them OK-confirmed is a broader change and is intentionally not in this PR (their durable retry queue already recovers genuine failures).
- No `dm_message_reactions` schema change: retry accounting is in-memory (resets on cold start, bounded by the foreground-only trigger).

## Verification

[...]

Issue #5374, opened 2026-06-20
fix(dm): route DMs out of message requests


hm21:
### Summary

Direct messages are incorrectly landing in Message Requests instead of the main inbox.

### Environment

android
main-branch

### Steps to Reproduce

1. Open the app and send a DM to a user.
2. Have the recipient check their inbox.
3. Observe where the message appears.

### Actual Result

The DM appears in Message Requests.

### Expected Result

The DM should appear in the recipient's main inbox when appropriate.

### Regression?

Yes

### Evidence

[openvine_full_logs_2026-06-20T07-45-16.043545.txt](https://github.com/user-attachments/files/29157328/openvine_full_logs_2026-06-20T07-45-16.043545.txt)

https://github.com/user-attachments/assets/84009caa-9f4f-42da-a37b-b71e52b1cd40

hm21:
@realmeylisdev  that issue is back again. I think it might be a race condition because a few days ago, when I tested your fix, it worked. But today, after checking out the latest main branch, I ran into the same issue again.

The good news is that this time I saw it right as it initially loaded, so I attached the logs.

realmeylisdev:
Hey @hm21 — dug into this properly. Short version: #5387 only partly covers this, and it most likely does **not** fix your case, so latest `main` won't clear it on its own.

What #5387 fixed: rows that had a stale "group" flag but really had only one other participant. What it does **not** fix: a conversation that got physically stored with an **extra participant** — 3 pubkeys `[you, peer, X]` instead of 2 — which then looks like a group to the classifier and lands in Message Requests regardless of follow state.

Your rabble / improvising / Liz threads are from March, which falls in a window (~Mar 11–Apr 4) where a non-compliant client adding an extra `p` tag to a 1:1 would get stored exactly like that. Nothing on main currently collapses such a row back to a clean 1:1, so it sticks.

Two things would help:

1. If you can run a debug/profile build with `--dart-define=DM_CLASSIFY_DIAGNOSTICS=true`, open the inbox, and paste the log line for one of those three: `classifyPotentialRequests → request: … participantCount=… otherPubkeys=[…]`. If `participantCount` is 3 (not 2), that confirms it's the extra-participant case.
2. A full reinstall will move them to the inbox (it re-ingests them as clean 1:1s) — but heads up, it can re-strand any chat you only ever replied to over the old NIP-04 DMs, so it's a workaround, not the fix.

[...]

Issue #254, opened 2025-12-01
fix: key storage failure on macos


andyhorn:
### Summary

The macos app displays a "setup error" and cannot authenticate.

### Environment

App version: 1.0.2+126
Device: macOS

### Steps to Reproduce

1. Open the project in VS Code
2. Set "macOS" as the Flutter device
3. Launch the app

### Actual Result

The app launches to the login page but displays an error in place of the actions.

### Expected Result

The app launches normally and allows me to proceed to the dashboard.

### Regression?

Unknown

### Evidence

<img width="567" height="1062" alt="Image" src="https://github.com/user-attachments/assets/d35730a5-9ede-46c2-b1aa-c3eea99ecab3" />

```
flutter: [16:05:03.019] [AUTH] Key generation error: SecureKeyStorageException: Failed to store key: Fallback storage failed: PlatformException(Unexpected security result code, Code: -34018, Message: A required entitlement isn't present., -34018, null)
flutter: [16:05:03.019] [AUTH] Failed to create secure identity: SecureKeyStorageException: Failed to store key: Fallback storage failed: PlatformException(Unexpected security result code, Code: -34018, Message: A required entitlement isn't present., -34018, null)
flutter: [16:05:03.019] [AUTH] Auth state changed: unauthenticated
flutter: [16:05:03.020] [AUTH] Failed to auto-create identity: Failed to create identity: SecureKeyStorageException: Failed to store key: Fallback storage failed: PlatformException(Unexpected security result code, Code: -34018, Message: A required entitlement isn't present., -34018, null)
```

PR #6046, opened 2026-07-11
fix(dm): make DM message delivery durable (OK-confirm + queue-aware retry + 10050 routing)

### LINKED ISSUE #6047: fix(dm): make DM message delivery durable (OK-confirm + queue-aware retry + 10050 routing)
Auto-created from PR #6046.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6046
**Branch:** `fix/dm-message-delivery-durable`

realmeylisdev:
Closes #6047

## Description

Brings DM **messages** to the OK-confirmed delivery reliability that #5977 gave DM **reactions**, the Nostr-native way: a message counts as sent only on a relay's NIP-20 `OK`, retries republish the same rumor to the recipient's kind-10050 inbox, and failed/unconfirmed sends are visible and recoverable instead of silently shown as delivered.

Most of the primitives already shipped in #5977 (`NIP17SendResult.retryablePending`/`blocked`, `sendRumor(awaitRecipientOk:)` with full outcome classification, `PublishOutcome`, the connectivity retry trigger) and #6028 (`_finalizeAfterRecipientBlocked`, tri-state `DmSendPolicyDecision`). This PR wires messages onto that machinery and closes the remaining gaps, in delineated commits:

**1. `fix(dm): OK-confirm message sends and keep soft-unconfirmed rows retryable`**
`sendMessage` and `recoverFullSend` counted a bare WebSocket frame-accept as delivery, so a relay that silently dropped or rejected the gift wrap deleted the queue row and showed a false "delivered". Both now route through `sendRumor(awaitRecipientOk: true)` (wrapped in a 15s timeout that surfaces a hung socket/signer as a soft failure). A **soft-unconfirmed** outcome (frame written, no OK, no rejection) is not proof of loss, so it keeps the row `pending` (in-flight clock, never a red failure) via `_finalizeAfterRecipientUnconfirmed` — reusing the existing `pending` status rather than adding a new `OutgoingWrapStatus` value, to stay clear of the fail-loud `_parseStatus` downgrade contract #6028 deliberately established. The retry sweep's interrupted-pending arm — which, unlike the failed arm's `getRetryableForOwner`, had **no** retry-count cap — now terminalizes an exhausted pending row to `failed`, fixing a latent unbounded-retry path and giving soft-unconfirmed sends a terminal state.

**2. `fix(dm): route retried and group DM wraps to the recipient kind-10050 inbox`**
`recoverFullSend` and `sendGroupMessage` published to the default relay pool instead of the recipient's advertised kind-10050 DM inbox — so a retried or group send never reached a non-diVine recipient who only reads their own inbox, and (with OK-confirm now live) the OK would be a false "confirmed" against a relay the recipient never reads. `recoverFullSend` re-resolves the recipient's inbox at retry time (10050 is replaceable); this one change also covers both sweep arms and collaborator-invite recovery. `sendGroupMessage` resolves every recipient's inbox concurrently up front so N sequential lookups don't delay the echo.

**3. `fix(dm): make the failed-send retry replay the queue row, not a new rumor`**
The failed-send SnackBar's Retry dispatched a fresh `ConversationMessageSent`, minting a **new** rumor id while the original failed row stayed sweep-retryable — a duplicate-delivery risk (receiver dedup keys on the rumor id). New `ConversationFullSendRecoveryRequested` replays the existing rows via `recoverFullSend` (stable rumor id), mirroring the self-wrap recovery path; the ids come from `ConversationState.failedOutgoingRumorIds` at tap time.

**4. `feat(dm): add per-bubble Try again / Cancel send on failed DM bubbles`**
An exhausted failed send was stranded — a permanent red bubble with no way to re-drive or dismiss it. Long-pressing a failed own bubble now offers **Try again** (manual `recoverFullSend`, which bypasses the sweep's retry-count cap) and **Cancel send** (new owner-checked, idempotent `DmRepository.cancelOutgoingSend` drops the row). New l10n keys translated across all 18 locales.

**5. `test(dm): cover group send soft-unconfirmed and mid-flight blocked finalize`** — coverage for the new group per-recipient classification branches.

**Related Issue:** Relates to the DM delivery-reliability line (#5977, #6028).

## Out of Scope

- **Inbox conversation-list failed-send badge (gap 8b).** Deferred to a focused follow-up PR: it requires folding a 7th stream into the inbox's delicate `combineLatest6` request-classification bloc (entangled with the #5374/#5304/#176 races), which is cleaner and safer as its own change than inside this already-sizeable PR.
- **Persisting the signed gift wrap for verbatim republish (gap 7).** The current re-wrap-with-stable-rumor-id model is dedup-correct; persisting signed wraps is a larger independent change with its own tradeoffs.

## Verification

- `dm_repository` package: full suite green, **line coverage 93.20%** (gate 93%).
- Touched mobile suites green: `outgoing_dm_retry_service_test`, `conversation_bloc_test`, `conversation_state_test`, `conversation_view_test`, `reaction_picker_overlay_test`, `arb_consistency_test`.
- New l10n keys mirrored + translated across all 18 `app_*.arb` locales; ARB consistency test passes.

## Type of Change

## Manual test plan

- 1:1 send on a flaky relay: bubble shows the in-flight clock until the relay OKs, then the check — never a false "delivered"; on a hard rejection it goes red.
- Long-press a failed own bubble → **Try again** re-drives it; **Cancel send** removes it.
- Failed-send SnackBar → **Retry** recovers the existing row (no duplicate at the recipient).
- Group send routes each wrap to its recipient's own inbox.

---

## Update — optimistic send UX (commit `feat(dm): optimistic sends …`)

Following review feedback, the send UX is now fully optimistic and reworked on top of the durability engine above:

- **No clock / no delivery ticks.** A sent message just appears sent; pending, delivered, and self-wrap-failed states show no indicator. The **only** surfaced state is a hard failure.
- **Failure lives on the bubble.** A hard-failed own send shows a red "Not delivered" row in the bubble itself. **Tapping** it opens a **native divine snackbar** with **Resend** (queue-aware `recoverFullSend`, bypasses the sweep's retry cap) and **Delete message** (drops the queued row). This replaces commit 4's long-press overlay affordance.
- **Native snackbar everywhere.** `DivineSnackbarContainer` gained an optional secondary (destructive) action so one snackbar carries Resend + Delete (divine_ui kept at 100% coverage). The blocked and partial-delivery toasts also moved to the divine snackbar; the auto failure toast is gone.
- **Cleanup.** Removed the now-dead `FailedSend` / `lastFailedSend` / `failedOutgoingRumorIds` state and the orphaned `dmStatusPending` / `dmStatusDeliveredSelfFailed` l10n keys; the action keys are repurposed to Resend / Delete message across all 18 locales.

Offline still shows failed immediately and auto-recovers on reconnect (unchanged).

---

## Update — full-pipeline audit fixes (7 commits)

A full audit of the PR against runtime behavior (Nostr protocol, timeout budgets, races, UI) found and fixed the following. Each was verified in code and is covered by tests.

**`fix(dm): repaint failed sends live — full value equality on OutgoingDm`** — the root cause of "sends never turn red while the conversation is open": `OutgoingDm.==` compared only the rumor id, so a pending row and its failed re-read were Equatable-equal inside `ConversationState` and `Bloc.emit` suppressed every status-only repaint (including the one-shot `getOutgoing` re-read). Full-field equality makes every wrap-status flip observable; a new regression test drives a real DAO `pending→failed` transition through the bloc.

**`fix(nostr): anchor gift-wrap timestamp jitter to now, not the rumor`** — both NIP-59 builders randomized the seal/wrap `created_at` 0–2 days into the past *of the rumor's* `created_at`. A retry replays the same rumor later, so its re-wrapped kind-1059 landed before recipients' `since` cursors and outside relay recency windows — retried messages were silently never delivered. Jitter now anchors to wall-clock now; the unsigned rumor keeps canonical time.

**`fix(dm): group-aware recovery, canonical timestamps, workable send budget`** — `recoverFullSend` hardcoded 1:1 semantics: recovering a group-send row overwrote the group conversation as a 1:1 (participants reduced, `isGroup:false`) and inserted a duplicate sibling message. It now preserves existing conversation topology, dedups the local persist against persisted batch siblings, and re-checks the queue row inside the success transaction so a mid-flight cancel can't resurrect the message. Persisted sends are stamped with the rumor's canonical `created_at` (no more bubble jump on OK-confirm). `sendGroupMessage` builds and enqueues all per-recipient rumors up front with one shared batch timestamp. A hard-failed first message now creates the conversation row so an offline-composed new thread is visible. The 15s publish cap couldn't contain a remote-signer send (4 Keycast RPCs at 12s each + the 10s OK window) — it's now a 90s backstop with every sub-step individually bounded (self-wrap publish 10s, Keycast 401 token-refresh bounded).

**`fix(dm): keep the retry sweep honest`** — an offline sweep pass burned the 5-retry budget on attempts that deterministically fail (a few airplane-mode foregrounds terminalized a row red before the network returned); the pass now skips when offline and the connectivity trigger re-fires it. Soft-unconfirmed outcomes were double-charged against the retry budget. Nothing re-drove rows during an uninterrupted foreground session — a 30s follow-up timer now sweeps while retryable work remains. The interrupted-arm min-age rises 2s→120s so triggers during a live send stop dispatching concurrent duplicate publishes.

**`fix(dm): non-blocking composer, one group bubble, hardened failed-bubble actions`** — the composer was gated on `sendStatus == sending`, which under OK-confirm lasts tens of seconds, silently swallowing every follow-up send tap; it's now never gated and `ConversationMessageSent` runs `concurrent()`. Group sends rendered one bubble per recipient; `displayedMessages` now collapses a fan-out batch into one bubble whose status aggregates worst-of-batch (including on the persisted bubble), and Resend/Delete target exactly the failed sibling rows. Snackbar actions are one-shot, dismiss on tap, and no-op after the bloc closes. Deleting a bubble cancels its durable queue row first, so delete works on queue-only bubbles and a deleted message can't be re-published by the sweep. The hard-failure screen-reader announcement is restored.

**`fix(dm): re-key inbox blocs on every captured identity-flippable dep`** — the previously-WIP inbox stuck-spinner fix, extended to cover every watched dependency the create factories capture.

**`fix(dm): wire the offline probe into bug-report DM sends`** — offline bug-report DMs no longer report fake success.

### Known residuals (deliberately out of scope)
- NIP-01 OK-false machine-readable prefix classification (`blocked:`/`restricted:`/`invalid:`) — terminal rejections are retried, but bounded by the 5-retry cap.
- Self-wrap routing to the sender's own kind-10050 inbox (currently default pool).
- Spec-canonical single-rumor group sends (one rumor p-tagging all members, wrapped per recipient) — requires an `outgoing_dms` composite-PK migration; the current per-recipient rumor variants are handled correctly sender-side by the sibling-batch machinery.

### Verification (post-rebase onto latest main)
- `dm_repository` 474 tests, `db_client` full suite + DAO 29, `nostr_sdk` nip59 18, `keycast_flutter` 207, app DM/inbox suites 572 — all green.
- Generated files (`social_providers.g.dart`, `bug_report_providers.g.dart`) regenerated and committed.

---

## Device test plan (supersedes the earlier manual test plan — there is no in-flight clock in the optimistic UX)

Setup: real device on this branch; a local-key sender AND a Keycast remote-signer sender; a recipient you control on a second device (ideally one diVine + one foreign NIP-17 client, e.g. 0xchat); a ≥3-member group. Useful log filters: `OutgoingDmRetryService`, `Keycast RPC`, `NIP-17`.

1. **Baseline 1:1 online** — send: bubble instant, no clock/status icon, composer never locks, no position jump when the relay confirms, exactly one copy at the recipient (also after kill+relaunch).
2. **Rapid fire** — 5 sends as fast as possible: all bubbles instant; previously the composer silently swallowed taps for 10–15s per send.
3. **Offline hard-fail** — airplane ON, send: red "Failed to send" appears ON the bubble within ~1s while the conversation stays open (previously only after re-opening the thread). Tap → Resend / Delete snackbar.
4. **Offline budget survival** — while offline, background/foreground 5+ times, then reconnect: the message still auto-delivers exactly once (previously those cycles burned the retry budget and stranded it red).
5. **Offline first message to a new person** — the new thread is visible in the inbox with its failed bubble (previously invisible); delivers on reconnect.
6. **Delete** — delete a failed bubble, reconnect, wait 2+ min with foreground cycles: the recipient must never receive it; long-press delete also works on failed bubbles.
7. **Resend one-shot** — mashing Resend fires once; snackbar dismisses on first tap.
8. **Long-delay retry (protocol fix)** — send offline, stay offline 1+ hour, recipient online throughout, then reconnect: recipient receives it shortly after (previously the re-wrap was backdated behind their `since` cursor and never arrived).
9. **Kill mid-send** — force-kill within 1s of sending; after relaunch delivery completes within ~2 min (deliberate 120s interrupted-row guard), one copy.
10. **Keycast signer** — repeat 1–3: instant bubble/composer regardless of RPC latency; slow sends keep working in the background up to ~90s instead of flipping red at 15s (server latency itself is keycast#291).
11. **Groups** — one bubble per message (not N); after an interrupted send, the single bubble goes red and Resend targets only the missing members; after ANY group recovery, the thread must still be a group (members intact) and a follow-up send must reach all members.
12. **Inbox spinner** — cold-launch straight into Inbox / Message Requests and after account switch: list always loads.

Not regressions (known residuals): terminal relay rejections retried up to the 5-try cap; own-device sync uses the default pool, not the sender's 10050; recipients on clients without rumor-id dedup may see a duplicate after a retry.


realmeylisdev:
**Server-side follow-ups (cross-repo)** for the Keycast remote-signer RPC latency (20–30s per `nip44_encrypt`/`sign_event`) that this PR mitigates on the client:

- **divinevideo/keycast#291** — 20–30s RPC latency from DB connection-pool acquire contention on the HTTP `nostr_rpc` path (uncached per-call user-status `SELECT` + background activity `UPDATE` on a 10-conn/60s pool, no per-request timeout). This is the actual delivery cure (cache account-status off the hot path / per-request timeout layer / short `acquire_timeout`).
- **divinevideo/keycast#292** — `nip17_wrap_batch` send-side verb, cutting the 4 Keycast RPCs per 1:1 DM to 1 (mirror of the receive-side #254).

The 12s single-op timeout cap added here (`c33e53ce0`) only makes a stalled signer **fail fast and retry honestly** (red "Not delivered" + durable-queue re-drive) instead of a fake "sent" that never lands. Delivery reliability under load still needs the keycast work above.

mobile/lib/blocs/dm/conversation/conversation_bloc.dart
```diff
@@ -123,12 +138,33 @@ class ConversationBloc extends Bloc<ConversationEvent, ConversationState> {
...
     Emitter<ConversationState> emit,
   ) async {
+    // Drop any durable queue row first, whether the bubble is queue-only
+    // (optimistic/failed — no persisted row yet, so the kind-5 path below
+    // would no-op and the "deleted" bubble would keep re-sending via the
+    // sweep) or persisted with a retry row still alive (deleting the
+    // message must also stop the sweep from re-publishing it).
+    var cancelledQueueRow = false;
+    try {
+      cancelledQueueRow = await _dmRepository.cancelOutgoingSend(
```
dcadenas: Cancel every live sibling queue row before deleting a group bubble. The persisted bubble carries the successful sibling's rumor ID, whose queue row has already been removed, so cancelling only that ID leaves pending and failed siblings available to the retry sweep. Use the same complete sibling set in the snackbar Delete path, including pending rows rather than only failed rows. Add a test that deletes a partially delivered group message, runs the retry sweep, and verifies that no sibling is published.

mobile/lib/services/outgoing_dm_retry_service.dart
```diff
@@ -103,12 +112,39 @@ class OutgoingDmRetryService {
...
   final CrashReportingService _crashReporting;
 
+  /// Minimum age of a still-`pending` row before the interrupted-send arm
+  /// may re-drive it. Must exceed the worst-case legitimately-in-flight
+  /// send (inbox resolution plus the 90s `_messagePublishTimeout` backstop
+  /// in `dm_repository`), or a foreground/connectivity trigger fired during
+  /// a live send dispatches a concurrent duplicate publish of the same
+  /// rumor. Receiver dedup makes that safe but wasteful; the old value
+  /// (the 2s `initialDelay`) made it the common case.
+  static const Duration _interruptedMinAge = Duration(seconds: 120);
```
dcadenas, line 141: Prevent the interrupted-send sweep from racing a group send that is still publishing recipients sequentially. A group send can exceed this 120-second single-publish guard, and the original send is not registered with the per-rumor recovery lock. Apply the recovery path's matching-message guard before the normal group-send persistence step, or coalesce the original publish through the same recovery lock, so only one persisted message can win. Cover the case where recovery persists first and verify that the normal group-send completion cannot insert a second bubble.

dcadenas (dismissed):
Group-message queue handling still permits duplicate bubbles and delivery after deletion. Close all three group-send races below before merge.

---

- **Group send with partial delivery renders the message twice** ([mobile/lib/blocs/dm/conversation/conversation_state.dart:226](https://github.com/divinevideo/divine-mobile/blob/4ade69ae97cac604e6b13dc1cad1ef8bfd45da0b/mobile/lib/blocs/dm/conversation/conversation_state.dart#L226))
  Give every per-recipient row for one group send a shared batch identity, then use that identity for grouping and persisted-bubble suppression. The current check compares each surviving sibling's rumor ID with the persisted winner's ID, so a partial delivery renders both the persisted bubble and a second pending or failed bubble. Remove the literal NUL from the grouping key as part of this fix so GitHub can render this source file as text. Add a regression test where one recipient confirms and another fails, and assert that the conversation shows exactly one bubble with the combined status.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
```diff
@@ -1535,12 +1536,59 @@ class RelayPool {
                 'noResponseFrom=${outcome.noResponseFrom}',
               );
             }
+            _repairSilentRelays(outcome.noResponseFrom, sentAt);
```
hm21: **major** — `_repairSilentRelays` runs on every outcome, including confirmed ones. `PublishTracker` completes on the first OK-true, so on any multi-relay publish the slower siblings land in `noResponseFrom` and get force-cycled even though the OK window never elapsed for them (`isSilentSince(sentAt)` passes trivially for an idle inbox relay in a ~150ms window). Every send then tears down a healthy inbox relay, and publishes issued while it reconnects fail because sends use `skipReconnect`. Gate the repair on the timeout path (e.g. `!outcome.confirmed`). The zombie tests only cover a single relay — add a multi-relay case.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -2626,7 +2720,11 @@ class DmRepository {
             conversationId: conversationId,
```
hm21, line 2787: **major** — the PR description says the success transaction re-checks the queue row so a mid-flight cancel can't resurrect the message, but that guard only exists in `_recoverFullSendLocked`. `sendMessage` (here) and `sendGroupMessage` persist without re-checking, so Delete tapped during the now 10–90s in-flight window drops the row and the late OK re-inserts the "deleted" message.

mobile/lib/blocs/dm/conversation/conversation_state.dart
```diff
@@ -178,13 +205,35 @@ class ConversationState extends Equatable {
...
+    final grouped = <String, List<OutgoingDm>>{};
+    for (final q in pendingOutgoing) {
+      grouped.putIfAbsent('${q.createdAt} ${q.content}', () => []).add(q);
+    }
+    final pendingBubbles = <DmMessage>[];
+    for (final batch in grouped.values) {
+      final recipients = {for (final q in batch) q.recipientPubkey};
+      final isFanOut = batch.length > 1 && recipients.length == batch.length;
+      if (isFanOut) {
+        if (batch.any((q) => persistedIds.contains(q.id))) continue;
```
hm21: **major** — this dedup can't fire in the normal success case: `_finalizeAfterRecipientSuccess` deletes the confirmed sibling's queue row once the self wrap landed, so its id is never in `batch`. A mixed-outcome group send renders the persisted bubble *plus* a second batch bubble — the exact duplicate this collapse exists to prevent. Match the batch key (content+createdAt) against persisted messages instead of queue-row ids. Also none of the new batch logic (`_siblingRowsFor`, `failedSiblingRumorIdsFor`, group aggregation in `statusFor`, this dedup) has a single multi-row test.

mobile/lib/services/outgoing_dm_retry_service.dart
```diff
@@ -103,12 +112,39 @@ class OutgoingDmRetryService {
...
+
+  /// Gap between a sweep that left retryable work behind and the follow-up
+  /// sweep it schedules. Without an in-session heartbeat, a soft-unconfirmed
+  /// row created while the app stays foregrounded on stable connectivity was
+  /// never re-driven until the next background/foreground flip.
+  static const Duration _followUpSweepGap = Duration(seconds: 30);
```
hm21, line 147: **major** — the heartbeat never bootstraps for the case this doc describes: `_scheduleFollowUp` is only called at the end of `sweep()`, and nothing triggers a sweep when a send finishes soft-unconfirmed. Cold start → empty queue → no timer armed; a later soft-unconfirmed send then sits `pending` (a plain delivered-looking bubble) until a background/foreground flip or connectivity change. The send path needs to nudge the service when it leaves a retryable row behind. The follow-up timer also has zero tests — cover it with fakeAsync.

mobile/lib/services/outgoing_dm_retry_service.dart
```diff
@@ -166,6 +204,18 @@ class OutgoingDmRetryService {
...
+      // Offline gate: dispatching while offline deterministically fails
+      // every row via the send path's own offline fail-fast and burns the
+      // retry budget for attempts that never had a chance. Instead of
+      // publishing, surface aged unconfirmed rows as failed (see
+      // [_surfaceOfflinePendingRows]) and skip the rest of the pass; the
+      // connectivity trigger re-fires it the moment the network is back.
+      // Probe errors fall through to sweeping (assume online).
+      if (await _isOfflineSafely()) {
+        await _surfaceOfflinePendingRows();
+        return;
```
hm21, line 255: **minor** — the offline pass returns before `_scheduleFollowUp`, so pending rows younger than the 120s min-age that `_surfaceOfflinePendingRows` skips are never re-examined while the device stays offline — they keep looking delivered for the whole offline session, which is the misread this pass exists to fix.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -3064,6 +3212,29 @@ class DmRepository {
         'wire OutgoingDmsDao into DmRepository before calling.',
       );
     }
+    if (resetRetryBudget) {
+      // Before the in-flight join, so a resend that lands mid-sweep still
+      // refreshes the budget for the sweep's follow-up passes. Non-fatal:
+      // a missed reset only means the sweep may give up earlier.
+      try {
+        await dao.resetRetryCount(rumorId);
```
hm21: **minor** — `resetRetryBudget: true` resets the row's retry count before the owner check in `_recoverFullSendLocked` throws, and `resetRetryCount` isn't owner-scoped — a foreign account's row gets its 5-attempt budget re-armed. Check ownership before the reset, like `cancelOutgoingSend` does.

mobile/lib/blocs/dm/conversation/conversation_bloc.dart
```diff
@@ -39,12 +39,27 @@ class ConversationBloc extends Bloc<ConversationEvent, ConversationState> {
...
     on<ConversationStarted>(_onStarted, transformer: restartable());
-    on<ConversationMessageSent>(_onMessageSent, transformer: sequential());
+    // Sends are optimistic and independent per rumor: `sequential()` made
+    // one slow OK-confirm (tens of seconds on a slow relay / remote signer)
+    // queue every later send BEHIND it, so the next message's bubble did
+    // not even appear until the previous publish finished. The repository
+    // enqueues each row before any signer round-trip, so concurrent
+    // handlers stay per-rumor isolated; `sendStatus` is last-writer-wins,
+    // which only drives the bubble-less toasts (blocked / partial).
+    on<ConversationMessageSent>(_onMessageSent, transformer: concurrent());
```
hm21, line 49: **minor** — with `concurrent()`, two overlapping sends that both end `sentPartial`: the second emit leaves `sendStatus` unchanged, so the view's status-transition listener never fires for it and its rumor ids silently overwrite `lastPartialSend`.

mobile/lib/screens/inbox/inbox_page.dart
```diff
@@ -43,19 +43,43 @@ class InboxPage extends ConsumerWidget {
...
+        // at runtime — reportingService resolves from null asynchronously,
+        // and currentUserPubkey flips on account switch — or the cubits
+        // capturing them silently keep the stale instance. See
+        // .claude/rules/state_management.md ("Bridging Riverpod-provided
+        // dependencies into BlocProvider").
+        key: ValueKey((
+          dmRepository,
+          followRepository,
+          blocklistRepository,
+          reportingService,
```
hm21: **minor** — including the async-resolving `reportingService` in this ValueKey deterministically tears down and recreates all five inbox blocs shortly after every inbox mount (double `ConversationListStarted`/load work per open). Gate on its readiness or exclude it from the key.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -486,6 +496,50 @@ class _MessageList extends StatelessWidget {
...
+    void runOnce(void Function() action) {
+      if (handled || bloc.isClosed) return;
+      handled = true;
+      messenger.hideCurrentSnackBar();
+      action();
+    }
+
+    messenger
+      ..hideCurrentSnackBar()
+      ..showSnackBar(
```
hm21, line 536: **minor** — the failed-bubble Resend/Delete snackbar ships without a `SemanticsService.sendAnnouncement`, unlike the blocked and partial snackbar paths in this same file.

mobile/lib/screens/inbox/conversation/widgets/message_bubble.dart
```diff
@@ -244,6 +250,7 @@ class MessageBubble extends StatelessWidget {
               ? context.l10n.dmMessageBubbleLongPressHint
               : null,
           child: GestureDetector(
+            onTap: onTap,
```
hm21, line 260: **minor** — the new tap affordance on failed own bubbles has no tap semantics (no `onTapHint`/button), so assistive-tech users can't discover that tapping opens resend/delete.

mobile/test/screens/inbox/conversation/conversation_view_test.dart
```diff
@@ -849,26 +851,21 @@ void main() {
...
       const failedSendContent = 'Hi there';
-      const failedSend = FailedSend(
-        content: failedSendContent,
-        recipientPubkeys: [otherPubkey],
-      );
 
       testWidgets(
-        'shows a localized retry SnackBar when sendStatus transitions to '
-        'failed',
+        'shows NO toast on a hard failure — the failure is surfaced on the '
```
hm21, line 859: **minor** — this deletes the SemanticsService announcement test while production deliberately keeps the announcement; the `failed` case in `listenWhen` now has zero coverage.

mobile/test/blocs/dm/conversation/conversation_bloc_test.dart
```diff
@@ -649,7 +758,7 @@ void main() {
 
         blocTest<ConversationBloc, ConversationState>(
           'emits [sending with optimistic message, failed without '
```
hm21, line 760: **minor** — the group-send soft-unconfirmed branch in `_onMessageSent` (all recipients `retryablePending`, no hard failure → `SendStatus.sent`) has no bloc test; only the 1:1 soft branch is pinned.

mobile/test/blocs/dm/conversation/conversation_bloc_test.dart
```diff
@@ -934,49 +1008,47 @@ void main() {
             // Failed: only the failing attempt's key is removed; the
```
hm21: nit — this edit hollows the "pendingId uniqueness" blocTest into an exact duplicate of the hard-fail test above, while its name and comments still describe behavior (`pendingOptimistic` strip, `sequential()`) that no longer exists.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -293,27 +302,26 @@ class _ConversationViewState extends ConsumerState<ConversationView> {
...
-          isSending: sendStatus == SendStatus.sending,
-          onSend: (text) {
-            context.read<ConversationBloc>().add(
-              ConversationMessageSent(
-                recipientPubkeys: participantPubkeys,
-                content: text,
-              ),
-            );
-          },
+    return MessageInputBar(
```
hm21, line 324: nit — with the `isSending` BlocSelector gone, `MessageInputBar.isSending` (send-guard + spinner branch) is dead code with no remaining call site setting it. Remove it.

mobile/lib/services/outgoing_dm_retry_service.dart
```diff
@@ -197,12 +254,18 @@ class OutgoingDmRetryService {
...
           // State A: recipient delivered, self-wrap missing. The
           // canonical case #4124 is closing.
+          dispatchedIds.add(row.id);
           try {
             final result = await _dmRepository.recoverSelfWrap(rumorId: row.id);
             if (result.success) {
               // recoverSelfWrap deleted the row on success. The bumped
               // retryCount would be moot anyway, so skip incrementRetry.
               processedSelfWrap++;
+            } else if (result.retryablePending) {
```
hm21: nit — the self-wrap arm's `retryablePending` branch is unreachable now, and the comment describes the synthetic in-flight-guard behavior this PR removed (the guard now joins the in-flight attempt).

mobile/packages/dm_repository/lib/src/nip17_message_service.dart
```diff
@@ -305,13 +359,37 @@ class NIP17MessageService {
...
+      }
+
       Log.info(
         'Sending NIP-17 encrypted message to recipient',
         category: LogCategory.system,
       );
 
       // Create a minimal Nostr instance for GiftWrapUtil.
       // Uses the injected signer (works with local or remote signing).
+      // Create a minimal Nostr instance for GiftWrapUtil.
```
hm21, line 387: nit — the new comment was inserted below the old one instead of replacing it; the "Create a minimal Nostr instance for GiftWrapUtil" block is now duplicated back-to-back.

mobile/packages/db_client/lib/src/database/daos/outgoing_dms_dao.dart
```diff
@@ -406,6 +465,31 @@ class OutgoingDmsDao extends DatabaseAccessor<AppDatabase>
     return query.watch().map((rows) => rows.map(_rowToModel).toList());
   }
 
+  /// One-shot read of the durable queue rows for [conversationId], scoped to
+  /// [ownerPubkey], newest first — the non-reactive companion to
+  /// [watchForConversation]. The conversation bloc uses it to reflect a send's
+  /// terminal outcome immediately, rather than waiting on a watch tick that can
+  /// be dropped when it fires while a bloc event handler is mid-flight.
+  Future<List<OutgoingDm>> getForConversation({
```
hm21, line 487: nit — new public `getForConversation` has no DAO test, although the DAO test file is touched in this PR.

mobile/lib/screens/inbox/conversation/widgets/message_bubble.dart
```diff
@@ -1125,44 +1136,29 @@ class _VideoCard extends ConsumerWidget {
...
     return Semantics(
       label: label,
-      child: Tooltip(
-        message: label,
-        child: Icon(icon, size: 14, color: color),
+      child: Row(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          const Icon(Icons.error_outline, size: 14, color: VineTheme.error),
+          const SizedBox(width: 4),
```
hm21: nit — uniform gap: `Row(spacing: 4)` instead of the manual `SizedBox(width: 4)`.

hm21 (commented):
Ran everything locally on the PR head: `flutter analyze` clean; dm_repository (479), db/app DM suites all green. dm_repository coverage is **93.03%** — exactly one covered line above the `min_coverage: 93` gate, so the next coverage-losing tweak fails CI; worth adding headroom while fixing the items below.

Four majors, all verified against the code (details inline):
1. `relay_pool.dart` — the silent-relay repair also runs on **confirmed** outcomes and force-cycles healthy sibling inbox relays on every multi-relay publish.
2. `dm_repository.dart` — Delete during an in-flight send resurrects the message; the cancel re-check the PR description claims only exists in `recoverFullSend`, not in `sendMessage`/`sendGroupMessage`.
3. `conversation_state.dart` — the group fan-out dedup keys on queue-row ids that are deleted on success, so mixed-outcome group sends show a duplicate bubble.
4. `outgoing_dm_retry_service.dart` — the 30s follow-up heartbeat never bootstraps for the soft-unconfirmed case it was written for.

The new group-batch state logic and the follow-up timer ship with no tests — given they carry the PR's headline anti-duplication contract, they need direct coverage.

mobile/lib/blocs/dm/conversation/conversation_bloc.dart
realmeylisdev: Fixed in a6bcf0788 and 4aaa00fdd. Deleting a bubble now cancels the whole batch through a new DmRepository.cancelOutgoingBatch, which resolves the siblings from a live queue row or from the persisted winner and drops every one of them. The failed-bubble snackbar's Delete now targets every undelivered sibling (pending and failed) instead of failed-only, while Resend stays failed-only so delivered or in-flight recipients are never re-delivered to. The requested test is in: with the batch rows cancelled, the sweep dispatches nothing (outgoing_dm_retry_service_test), plus cancelOutgoingBatch unit tests covering both resolution paths and the 1:1/foreign-owner/missing-row edges.

mobile/lib/services/outgoing_dm_retry_service.dart
realmeylisdev: Fixed in 4aaa00fdd and 901bdcd59 — both halves. sendGroupMessage's success transaction now applies the recovery path's matching-message guard (hasMatchingMessage on the batch identity) plus a per-row cancel interlock, and keys the insert off the first LIVE success, so whichever transaction commits second sees the winner and only one persisted message can win. Each per-recipient publish is additionally registered under the same 'full:<rumorId>' recovery lock, so a sweep firing mid-send joins the live attempt instead of dispatching a duplicate publish. The requested case is covered in dm_repository_test: recovery persists first → the group-send completion inserts no second bubble and only finalizes the rows; a join test pins one publish per recipient.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
realmeylisdev: Fixed in 9d17743cf — the repair is gated on !outcome.confirmed (the diagnostic log stays unconditional), with a comment pinning the tradeoff. Added the multi-relay case: the fast relay OKs, the silent sibling keeps its channel. Verified the new test fails with the gate mutated away.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: Fixed in 4aaa00fdd — sendMessage's success transaction now re-reads the row first and skips persist/upsert/finalize when it is gone, mirroring _recoverFullSendLocked. The NIP-04 fallback is also gated on the persist actually happening: protocol stays null on the skip path and would otherwise have fired a plaintext copy of the deleted message. The same interlock runs per-row in sendGroupMessage. Tests cover 1:1 and group, all-cancelled and partial-cancel.

mobile/lib/blocs/dm/conversation/conversation_state.dart
realmeylisdev: Fixed in a6bcf0788 — suppression now matches the batch identity (own senderPubkey + content + createdAt) against persisted messages instead of surviving queue-row ids, and batch membership uses the repository's own group-row discriminator (conversationId != pair id) rather than counting survivors — the count heuristic also broke statusFor for a single surviving sibling, which would have hidden the persisted bubble's red affordance in exactly the mixed-outcome case. The NUL byte in the grouping key is gone (record-typed keys). conversation_state_test now carries the multi-row coverage: the one-confirmed-one-failed regression, a 3-recipient variant, incoming-coincidence non-suppression, same-second 1:1 independence, and the sibling-helper matrix.

mobile/lib/services/outgoing_dm_retry_service.dart
realmeylisdev: Fixed in d6c24a04b — DmRepository now exposes retryableOutgoingWork, fired wherever a send or recovery leaves a retryable row behind (soft-unconfirmed, hard failure, partial delivery, self-wrap recovery failure; not blocked/terminal). The service subscribes in initialize(): a nudge arms the 30s timer only if idle, and one landing mid-sweep is buffered and consumed after _isSweeping clears so a pass ending with no visible work cannot strand it. fakeAsync tests cover the cold-start bootstrap, the no-postpone contract, mid-sweep buffering, and dispose.

mobile/lib/services/outgoing_dm_retry_service.dart
realmeylisdev: Fixed in d6c24a04b — _surfaceOfflinePendingRows now reports whether young pending rows remain (or the fetch failed) and the offline arm schedules the follow-up on that, so the rows are re-examined every 30s while offline and flip to failed once they age past the guard. Covered with a fakeAsync test walking a row across the 120s threshold and asserting the chain disarms once nothing is pending.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: Fixed in 4aaa00fdd — the reset is owner-checked first (getById + owner match, same pattern as cancelOutgoingSend) and stays before the join. Tests: foreign and missing rows never reset, an owned row resets once, and a throwing owner-check read skips the reset without masking the locked body's error.

mobile/lib/blocs/dm/conversation/conversation_bloc.dart
realmeylisdev: Fixed in a6bcf0788, three parts: consecutive partials union their rumor ids instead of overwriting; the eager clearLastPartialSend on the sending emit is gone (with concurrent() it wiped the first send's ids before the union could see them); and the view's listenWhen also fires on a sentPartial-to-sentPartial transition whose id set changed. The recovery reduce now removes only the attempted ids so a concurrent union is not clobbered mid-recovery. A blocTest pins two overlapping partials ending with the union.

mobile/lib/screens/inbox/inbox_page.dart
realmeylisdev: Fixed in 9403d8c25 — reportingService is out of the MultiBlocProvider key; only the ConversationActionsCubit provider (last in the list, action-only state) is re-keyed on it, so the resolution flip no longer tears down the list blocs.

mobile/lib/screens/inbox/conversation/conversation_view.dart
realmeylisdev: Fixed in 9403d8c25 — _onFailedMessageTap announces via SemanticsService.sendAnnouncement, same as the blocked/partial paths.

mobile/lib/screens/inbox/conversation/widgets/message_bubble.dart
realmeylisdev: Fixed in 9403d8c25 — the bubble-level Semantics gains button: true and a localized onTapHint (new dmMessageBubbleFailedTapHint key, translated in all 18 locales) whenever the failed-tap affordance is wired.

mobile/test/screens/inbox/conversation/conversation_view_test.dart
realmeylisdev: Restored in 9403d8c25, adapted to the current state shape — same accessibility-channel interception, pinning the localized failure copy on the sendStatus-to-failed transition.

mobile/test/blocs/dm/conversation/conversation_bloc_test.dart
realmeylisdev: Added in a6bcf0788 — all recipients retryablePending with no hard failure emits [sending, sent].

mobile/test/blocs/dm/conversation/conversation_bloc_test.dart
realmeylisdev: Deleted in a6bcf0788 — the pendingOptimistic contract it pinned no longer exists and the residue duplicated the hard-fail test.

mobile/lib/screens/inbox/conversation/conversation_view.dart
realmeylisdev: Removed in 9403d8c25 — parameter, send guard, and spinner branch.

mobile/lib/services/outgoing_dm_retry_service.dart
realmeylisdev: Removed in d6c24a04b — verified recoverSelfWrap can only return success or a plain failure (publishSelfWrap never sets retryablePending and that path never goes through _sendRumorWithTimeout's soft classification), so the branch folded into the failure arm and the stale comment is gone.

mobile/packages/dm_repository/lib/src/nip17_message_service.dart
realmeylisdev: Fixed in 4aaa00fdd.

mobile/packages/db_client/lib/src/database/daos/outgoing_dms_dao.dart
realmeylisdev: Added in 9403d8c25 — scoping (conversation + owner) and newest-first ordering.

mobile/lib/screens/inbox/conversation/widgets/message_bubble.dart
realmeylisdev: Fixed in 9403d8c25.

realmeylisdev:
@dcadenas All three blocking items are addressed, pushed as 9d17743cf..3e076128f:

1. **Duplicate bubble on partial group delivery** (review body, conversation_state.dart) — fixed in a6bcf0788. Suppression now matches the batch identity (own sender + content + createdAt) against persisted messages instead of surviving queue-row ids, batch membership uses the repository's group-row discriminator (conversationId != pair id), and the literal NUL is gone from the grouping key (record-typed map keys — the file renders as text again). The regression test you asked for is in conversation_state_test: one recipient confirms, one fails → exactly one bubble, and statusFor on it reports failed.
2. **Delete leaves sibling rows for the sweep** — fixed in a6bcf0788 + 4aaa00fdd via a new DmRepository.cancelOutgoingBatch (resolves siblings from a live row or the persisted winner, drops them all); the snackbar Delete targets every undelivered sibling. The delete-then-sweep test verifies the sweep publishes nothing afterwards.
3. **Sweep racing an in-flight group send** — fixed in 4aaa00fdd + 901bdcd59 with both of your options: the recovery path's matching-message guard + per-row cancel interlock inside the group-send transaction (so only one persisted message can win — tested with recovery-persists-first), and the per-recipient publishes coalesced through the same 'full:<rumorId>' recovery lock so the sweep joins the live attempt.

One adjacent gap worth a follow-up issue rather than this PR: deleteMessageForEveryone publishes a single kind-5 referencing the persisted winner's rumor id, but each group recipient received a different rumor id, so their clients cannot match the deletion. This PR fixes the local/queue side; per-sibling kind-5 fan-out would need the delivered siblings' rumor ids, which are discarded at finalize time today.

realmeylisdev:
@hm21 All four majors, the eight minors, and the six nits are addressed, pushed as 9d17743cf..3e076128f — per-thread replies cite the exact commits. Highlights:

- relay_pool repair gated on !outcome.confirmed, with the multi-relay sibling test (verified it fails without the gate).
- The cancel interlock now runs in both send transactions (mirroring _recoverFullSendLocked), with the NIP-04 fallback suppressed on the skip path; the group transaction also gained the hasMatchingMessage sibling dedup and keys its insert off the first live success.
- The group fan-out dedup keys on the batch identity against persisted own messages, and the batch/sibling logic now has direct multi-row coverage in conversation_state_test.
- The follow-up heartbeat bootstraps from a new DmRepository.retryableOutgoingWork nudge stream (fired by the finalize helpers), with fakeAsync coverage for the timer including the mid-sweep buffering case; the offline pass schedules its own follow-up while young pending rows remain.

On the coverage headroom: dm_repository moved from 93.03% to 93.51% locally, with the new guard paths, the nudge emissions, cancelOutgoingBatch, and getOutgoing all covered.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -3616,25 +4198,75 @@ class DmRepository {
...
+        }
+
+        // Sibling dedup (same guard as _recoverFullSendLocked): a group
+        // send's sequential publishes can outlive the retry sweep's
+        // interrupted-send min-age, so the sweep may recover a sibling and
+        // persist the batch's ONE local message first. Batch identity is
+        // (conversationId, content, createdAt).
+        final alreadyPersisted =
+            outgoingDao != null &&
+            await _directMessagesDao.hasMatchingMessage(
```
dcadenas: Give each logical group-send invocation a durable, collision-proof batch ID before using `hasMatchingMessage` as the sibling guard here. That DAO matches any same-sender, same-content message within ±5 seconds, so two separate equal-text sends can make this guard skip insertion of the later confirmed message while the later queue rows are still finalized at `dm_repository.dart:4287-4293`. Use the batch ID consistently for persistence deduplication, recovery, optimistic display grouping, sibling lookup, and cancellation, then add a regression test that sends the same group text twice 1-5 seconds apart and verifies that both local messages persist while each batch remains independent.

dcadenas (dismissed):
Make the identity for each logical group send collision-proof before merging this change. The current five-second same-sender/content match can treat two separate equal-text sends as one, omit the later confirmed message from local history, and then finalize its retry rows. Use one durable batch identity across persistence, recovery, optimistic grouping, sibling lookup, and cancellation, and cover the concurrent repeated-text case with a regression test.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: Fixed in 07d06dfb4. Each group send now carries a durable, collision-proof batch id — the first sibling rumor's event id — stamped identically on every sibling `outgoing_dms` row and the batch's single persisted `direct_messages` row (new nullable `send_batch_id` on both tables, added via `_addColumnIfMissing`; schemaVersion unchanged).

All five surfaces now key off that id: send-path dedup (this line, via the new `hasMessageWithSendBatchId` probe), recovery dedup, optimistic grouping, sibling lookup, and cancellation. `hasMatchingMessage` stays only for cross-protocol receive dedup; rows enqueued before the column existed carry null and fall back to the old tuple.

Regression tests cover two identical-text group sends seconds apart both persisting and cancelling independently, plus the enqueue-side stamping (`sendBatchId == rumors.first.id`, shared across siblings).

mbradley (commented):
noting the other reviews already on this, i took an extra pass focused on the protected-minor DM lens. the group-send dedup and batch identity, silent-relay repair, and retry-heartbeat threads look like they're being addressed in the latest commits, so this pass is on top of those.

on the T&S lens, the DM-restriction dependency holds and nothing regressed:

- the send-policy gate still fires first in `sendRumor` and pre-enqueue in `sendMessage`/`sendGroupMessage`, so a confirmed-protected minor can't get a wrap built or a queue row written for a non-official recipient.
- Resend re-gates: it rebuilds the rumor and re-runs the policy rather than replaying a pre-built gift-wrap, so a row queued before the verdict resolved gets re-blocked on replay.
- a blocked send returns before any row is written, so it never renders an optimistic delivered-looking bubble.
- the group path stays all-or-nothing to match Keycast, and the inbox re-keying doesn't touch the outbound policy or the sticky protected verdict.

no leak path, and nothing to change here for safety.

reviewing this did surface one adjacent thing i'm taking as a follow-up rather than adding to your PR: once Keycast's verified_minor DM gate is on prod, its refusal comes back as an HTTP 403 "operation denied by policy", which currently lands in the generic catch as a retryable failure rather than a terminal block. no leak, since Keycast just refuses to sign, it only burns the retry budget and leaves a Resend button that re-fails. i've opened #6067 to handle it once this lands, so it builds on your final classification code. nothing for you to do here.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -3568,26 +4130,41 @@ class DmRepository {
...
+    // sibling rumor's event id. Globally unique and already computed, it is
+    // stamped identically on every sibling queue row AND on the batch's single
+    // persisted local message, so persistence dedup, optimistic grouping,
+    // sibling lookup, and cancellation all match a batch by one exact stored
+    // value. This replaces the old `(content, createdAt ±5s)` heuristic, which
+    // collapsed two distinct group sends of identical text seconds apart into
+    // one — dropping the later message while its wire copies were delivered.
+    // Same-second byte-identical resends share this id, which is correct: they
+    // are the same Nostr event and the receiver dedups them to one.
+    final sendBatchId = rumors.first.id;
```
dcadenas: At `mobile/packages/dm_repository/lib/src/dm_repository.dart:4153`, generate a collision-resistant value for every `sendGroupMessage` invocation before building rumors. The current `sendBatchId = rumors.first.id` still collides when identical group sends share the same Unix-second timestamp, so the second enqueue is ignored, publishing joins the first attempt, and batch dedup can suppress the second local message. Carry one shared per-invocation batch ID through every sibling queue row and local message, and keep recovery dedup, optimistic grouping, sibling lookup, and cancellation keyed to it. Add a production-path regression test that performs two identical same-second group sends and verifies distinct rumor IDs, distinct batch IDs, two local messages, independent queue batches, and independent cancellation.

dcadenas (changes requested):
The new group-send batch identity still collides when identical sends occur within the same Unix second, so one user action can be silently dropped from the queue, local history, and recipients. Before merging, generate unique per-invocation entropy before rumor IDs are built, carry one shared batch ID through all sibling rows and persistence, and add a production-path regression test for same-second sends and independent cancellation.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: Fixed in 5c2f44a98. `sendGroupMessage` now mints the batch identity as independent per-invocation entropy (256-bit secure random, injectable for tests) BEFORE any rumor is built, instead of deriving it from `rumors.first.id`. A rumor id is `sha256([0, pubkey, created_at(seconds), kind, tags, content])` with no nonce, so the derived id — and the queue-row PKs (`OutgoingDm.id`) and `full:<rumorId>` recovery locks that share it — collided for two identical group sends in the same Unix second: the second enqueue was `insertOrIgnore`'d away, its publish joined the first's recovery lock, and the persist dedup dropped its local message.

The token is injected into every sibling rumor as a client-internal `batch` tag so the rumor ids — and thus the queue-row PKs and recovery locks — are now distinct per invocation even at same-second identical text, and it is stamped verbatim on every sibling `outgoing_dms` row and the batch's single persisted `direct_messages` row. All five surfaces (send-path dedup, recovery dedup, optimistic grouping, sibling lookup, cancellation) key off that stored value; recovery/resend rebuild the identical rumor from the stored JSON via `Event.fromJson` so the id never drifts, and the tag rides inside the encrypted gift wrap (recipient / self-wrap only), is fresh-random per send, and is ignored by the receive path and other clients.

Production-path regression test: two identical same-second group sends now produce distinct rumor ids, distinct batch ids, two persisted local messages, four independent queue rows, and independent cancellation — verified to fail (4 rows collapse to 2) when the entropy injection is removed. The full-delivery test now pins the batch id as independent entropy rather than the first sibling's rumor id.

One thing I deliberately left out of scope: the 1:1 `sendMessage` path has the same latent same-second identical-text collision (no rumor nonce + `insertOrIgnore`), but it predates this PR and sits outside the group-batch-identity concern here, so I'll take it as a separate follow-up rather than widening this PR.

dcadenas (approved):
✅ durable DM delivery and clear retry recovery

mobile/packages/keycast_flutter/lib/src/rpc/keycast_rpc.dart
```diff
@@ -42,44 +44,75 @@ class KeycastRpc implements NostrSigner, GiftWrapBatchUnwrapper {
...
+  ///
+  /// Sized so the DM send pipeline's full remote-signer RPC chain (up to
+  /// four single ops per send: encrypt+sign for the recipient wrap, then for
+  /// the self wrap) fits comfortably under `dm_repository`'s 90s
+  /// `_messagePublishTimeout` backstop, while a stalled call still fails
+  /// fast enough for the durable outgoing-DM queue to re-drive it. See the
+  /// Keycast RPC latency investigation (#6046). A single crypto op is
+  /// sub-second in the healthy path, so 12s is generous while still a
+  /// tighter dead-socket bound than the old 30s.
+  static const Duration defaultRequestTimeout = Duration(seconds: 12);
```
NotThatKindOfDrLiz, line 66: This tighter default applies to every Keycast single-op caller, not just durable DM sends. `AuthService` constructs `KeycastRpc.fromSession(...)` without overriding `requestTimeout`, so OAuth video publish/signing, likes, reposts, follows, etc. now fail at 12s where they previously had 30s. Please scope the shorter bound to the DM send path or keep the default at 30s until the production latency issue is fixed.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -3597,15 +4214,48 @@ class DmRepository {
...
-      );
-      rumors.add(rumor);
+    // Delivery below is per-recipient and NOT atomic: a mid-loop publish
+    // failure can leave the group partially delivered. That is a reliability/UX
+    // concern only, not a safety leak for the #176 restriction — the
+    // all-or-nothing gate above already guarantees EVERY recipient is approved,
+    // so a partial send can only ever reach approved recipients, never a
+    // blocked one.
+    for (var i = 0; i < recipientPubkeys.length; i++) {
+      final pubkey = recipientPubkeys[i];
```
NotThatKindOfDrLiz, line 4230: This loop can still deliver to later group recipients after the sender deletes/cancels the optimistic bubble mid-fan-out. The transaction interlock below prevents local persistence/finalization, but it does not stop dispatch to siblings that have not been published yet. Please re-read the sibling row at the top of this loop and skip when it has been cancelled.

mobile/lib/services/outgoing_dm_retry_service.dart
```diff
@@ -333,24 +437,41 @@ class OutgoingDmRetryService {
...
+
+          // Exhausted guard: a pending row that has burned the retry budget
+          // is terminally unconfirmable — a soft-unconfirmed send whose OK
+          // never arrived, or an app-killed row that can never reach a relay.
+          // Flip it to `failed` so it stops re-driving, renders as a red
+          // failed bubble, and becomes a manual-retry candidate. The failed
+          // arm's `getRetryableForOwner` already caps at `retry_count <
+          // maxRetries`; `getStillPendingForOwner` does not, so without this
+          // an unconfirmable pending row would re-drive forever.
+          if (row.retryCount >= _retryConfig.maxRetries) {
```
NotThatKindOfDrLiz, line 459: `getStillPendingForOwner` also returns rows where the recipient wrap is already `sent` but the self-wrap is still `pending`. When one of those exhausts retries, `_terminalizeRow` marks both wraps failed, turning a delivered message into a red failed bubble and making Resend republish it. The offline path already guards `recipientWrapStatus != pending`; the online exhausted path needs the same guard.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -343,6 +368,24 @@ class DmRepository {
   Timer? _readMarkerDebounce;
   final Duration _readMarkerDebounceDelay;
   static const _defaultReadMarkerDebounceDelay = Duration(seconds: 3);
+
+  /// Rumor tag key carrying the per-send batch token (see [sendGroupMessage]).
+  /// Client-internal: injected only so two identical group sends in the same
+  /// Unix second produce distinct rumor ids. Divine's own receive path and
+  /// other clients ignore unrecognised rumor tags, so it is inert on the wire.
+  static const _sendBatchTagKey = 'batch';
```
NotThatKindOfDrLiz, line 376: Since the batch id now rides in the encrypted rumor, the receive/self-wrap ingest path needs to extract this tag and pass it to `insertMessage(sendBatchId: ...)`. `_persistDecryptedGiftWrap` currently writes `tagsJson` but leaves `sendBatchId` null, so a sender ingesting their own group self-wrap cannot match the queued/persisted batch by exact id and can project a duplicate bubble.

mobile/lib/services/outgoing_dm_retry_service.dart
```diff
@@ -427,12 +552,28 @@ class OutgoingDmRetryService {
...
         name: 'OutgoingDmRetryService',
         category: LogCategory.system,
       );
+
+      // Everything that neither delivered nor terminalized this pass is
+      // still driveable work: failed re-attempts, rows skipped by backoff
+      // or the min-age guard, and an aborted not-ready pass (rows
+      // untouched). Derived from the pass's own counters so no extra DAO
+      // round-trip is spent on the decision.
+      _scheduleFollowUp(
```
NotThatKindOfDrLiz, line 568: This is the last statement in the `try`, but the top-level `catch` does not re-arm the follow-up timer. A transient sweep-level throw, such as a busy DB or unexpected DAO parse error, can kill the 30s heartbeat until some external trigger arrives. Please schedule from `finally` or otherwise ensure throwing sweeps still re-arm when retryable work may remain.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -486,6 +502,65 @@ class _MessageList extends StatelessWidget {
...
+        DivineSnackbarContainer.snackBar(
+          l10n.dmSendFailedMessage,
+          error: true,
+          actionLabel: l10n.dmMessageActionRetrySend,
+          onActionPressed: () => runOnce(
+            () => bloc.add(
+              ConversationFullSendRecoveryRequested(rumorIds: resendIds),
+            ),
+          ),
+          secondaryActionLabel: l10n.dmMessageActionCancelSend,
```
NotThatKindOfDrLiz, line 546: The label says Delete message, but this handler only dispatches queue cancellation events. For a partially delivered persisted group bubble, that leaves the `direct_messages` row in place, so the bubble can turn normal and remain visible. Either route this through the real delete flow (`ConversationMessageDeleted` / kind-5 where applicable) or rename the action to match the actual behavior, such as Stop trying.

mobile/lib/screens/inbox/conversation/widgets/message_bubble.dart
```diff
@@ -240,10 +246,18 @@ class MessageBubble extends StatelessWidget {
...
+          // for that case, so key the button semantics off it.
+          button: onTap != null,
+          onTapHint: onTap != null
+              ? context.l10n.dmMessageBubbleFailedTapHint
+              : null,
           onLongPressHint: onLongPress != null
               ? context.l10n.dmMessageBubbleLongPressHint
               : null,
           child: GestureDetector(
+            onTap: onTap,
```
NotThatKindOfDrLiz, line 260: For failed video-share bubbles, the inner `_VideoCard` also has a `GestureDetector.onTap` that opens the reel. That recognizer can win the tap, so the failed-send snackbar is only reachable by tapping the narrow Not delivered strip. Please disable or gate the inner video-card tap when the outer failed-send tap affordance is active.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
```diff
@@ -1535,12 +1536,70 @@ class RelayPool {
...
+  /// health gate — so every retry inside that window would deterministically
+  /// time out again. Force-cycle the connection and re-send the relay's
+  /// subscriptions (mirroring [add]'s autoSubscribe and the post-AUTH replay)
+  /// so the next attempt runs on a live socket.
+  void _repairSilentRelays(List<String> silentRelayUrls, DateTime sentAt) {
+    for (final url in silentRelayUrls) {
+      final relay = _relays[url] ?? _cacheRelays[url] ?? _tempRelays[url];
+      if (relay is! RelayBase) continue;
+      if (!relay.isSilentSince(sentAt)) continue;
+      if (!_silentRelayRepairsInFlight.add(url)) continue;
```
NotThatKindOfDrLiz, line 1579: `_silentRelayRepairsInFlight` only dedupes concurrent repairs. A slow/brownout relay that repeatedly accepts frames but replies after the OK window can be force-cycled on every timed-out publish, replaying subscriptions each time. Please add a per-relay cooldown/minimum repair interval before reconnecting.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
```diff
@@ -1535,12 +1536,70 @@ class RelayPool {
...
+        _reconnectAndResubscribe(
+          relay,
+        ).whenComplete(() => _silentRelayRepairsInFlight.remove(url)),
+      );
+    }
+  }
+
+  Future<void> _reconnectAndResubscribe(RelayBase relay) async {
+    try {
+      if (!await relay.forceReconnect()) return;
```
NotThatKindOfDrLiz, line 1594: This reconnect path replays long-lived subscriptions but does not clear or complete in-flight one-shot queries stored on the relay. Those `_queries` entries can survive the reconnect, so `checkQuery(subId)` remains true forever and the pool-level all-relays-complete callback never fires for that query. Please clear/fail/complete one-shot queries when force-reconnecting here.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The durable DM state machine is directionally solid, but the current head still has regressions that should be fixed before merge. The most important ones are the global Keycast timeout regression, cancelled group sends continuing to fan out, retry exhaustion clobbering recipient-delivered rows, and batch ingest not stamping send_batch_id.

mobile/packages/keycast_flutter/lib/src/rpc/keycast_rpc.dart
realmeylisdev: Fixed in #6075 (91e491862). You're right that the 12s bound landed as the global default — none of the five `KeycastRpc.fromSession` sites override `requestTimeout`, so every single-op RPC (video-publish signing, likes, reposts, follows) inherited it, and under the ~20–30s Keycast contention this regresses those flows for custodial OAuth users. Reverted the single-op default to 30s; kept the batch verb at 30s. DM durability doesn't rely on the tighter bound (outgoing-queue re-drive + dm_repository's 90s `_messagePublishTimeout` backstop).

Scoping a fail-fast bound to the DM crypto chain specifically — your first preference — is a cleaner follow-up rather than a one-liner today: `signEvent`/`nip44*` expose no per-call timeout and the signer instance is shared. I took the safe revert now and noted the scoped version as out-of-scope on #6075.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: Fixed in #6075 (afb7c10d2). Confirmed: the per-recipient fan-out loop had no pre-publish row re-read, so a mid-fan-out `cancelOutgoingBatch` still built and published to later recipients. The loop now re-reads the sibling row before each publish and skips when it's gone (mirroring `recoverFullSend`'s existing interlock); the post-loop finalize and ensure-visible block exclude cancelled indexes so a fully cancelled batch can't resurrect the deleted thread. The currently-publishing sibling's inherent TOCTOU is left as-is, same trade-off `recoverFullSend` already makes.

Regression test: recipient-1 publish blocks on a Completer, `cancelOutgoingBatch` mid-flight, assert recipients 2/3 are never published, their results are non-success, and nothing persists. `recoverFullSend`'s group path already re-reads before publishing, so it was unaffected.

mobile/lib/services/outgoing_dm_retry_service.dart
realmeylisdev: Fixed in #6075 (b8bbac850) as invariant hardening. The guard asymmetry is real — the offline path guards `recipientWrapStatus != pending` but the online exhausted arm terminalized both wraps unconditionally. When the recipient wrap is already `sent`, it now fails only the self wrap (row exits the pending set as sent/failed → deliveredSelfFailed, not a red bubble).

One caveat worth recording: I traced the write paths and couldn't find one that currently produces the `recipient=sent / self=pending` hybrid — every `recipient=sent` write is transactionally coupled to row-delete or `self=failed`. So this is defensive against a future writer or a corrupt row rather than a live clobber on `main`; the offline suite already models the state, which is exactly why hardening the online arm to match is worth it. Test mirrors your offline fixture for the online path (`verifyNever(markRecipientWrapStatus … failed)`, `markSelfWrapStatus(failed)` once).

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: Fixed in #6075 (afb7c10d2). `_persistDecryptedGiftWrap` now extracts the `['batch', id]` tag and passes `sendBatchId` to `insertMessage`, so a self-wrap echo ingested before the batch's local persist matches by exact id instead of writing a NULL-batch row. Two guards on top of the suggestion: stamp only when the rumor is self-authored (`hasMessageWithSendBatchId` is owner-scoped, so a peer-forged tag could otherwise suppress the local user's own in-flight group persist), and require a 64-hex value to keep the batch keyspace disjoint from the legacy tuple.

Scope note for the record: the race only bites a 2nd+ send to an *existing* group — a brand-new group's first self-echo is rejected before persist (it collapses to a phantom self-conversation), so there's nothing to stamp there. Tests cover self-authored → stamped + sibling recovery dedups to one row; peer-authored → NULL; malformed value → NULL.

mobile/lib/services/outgoing_dm_retry_service.dart
realmeylisdev: Fixed in #6075 (b8bbac850). Confirmed: `_scheduleFollowUp` was the last statement of the `try` and the top-level `catch` never re-armed, so a transient throw (busy DB, DAO parse error) silenced the 30s heartbeat until an external trigger — the steady-foreground, stable-network, no-new-sends session is exactly where that bites. Re-arm from the `catch` with `workRemains: true` (queue state is unknown after a throw); the next clean pass's own `workRemains: false` self-cancels once the queue drains. I kept it in the `catch` rather than moving the call into `finally`, so the legitimate self-cancel on the success path is preserved. Test: DAO throws once then succeeds under `fakeAsync`, elapse the 30s gap, assert a second sweep ran.

mobile/lib/screens/inbox/conversation/conversation_view.dart
realmeylisdev: Fixed in #6075 (d759ce9c8) via the rename option. Confirmed the handler only dispatches `ConversationOutgoingSendCancelled` → `cancelOutgoingSend` (queue rows), so for a partial group bubble the persisted winner survives and the "deleted" bubble reverts to delivered once its siblings clear. Since keeping the already-delivered copy is the documented intent (`undeliveredSiblingRumorIdsFor`), I renamed the action to "Stop trying" + fixed the ARB description, translated across all 17 non-English locales; long-press delete stays the real delete-for-everyone path.

If you'd rather the snackbar actually retract delivered copies (your Option B — route through `ConversationMessageDeleted` / kind-5), that's a behavior + copy change I left for your call rather than shipping silently.

mobile/lib/screens/inbox/conversation/widgets/message_bubble.dart
realmeylisdev: Fixed in #6075 (543f77e51). Confirmed the inner `_VideoCard` tap won the gesture arena over the 248×350 card surface, leaving the tap-only resend affordance reachable only from the narrow "Not delivered" strip (long-press deliberately offers no resend). `MessageBubble.build` now computes the failed-own-send state and threads an `enableTap` flag down `_VideoLinkPreview → _VideoCard` and `_QuotedVideoPreview → _QuotedVideoCard → _QuotedVideoFrame`; on a failed send the inner `onTap` is null (no recognizer, no tap semantics) so the outer failed-tap wins everywhere and assistive tech stops announcing a dead open-video action. Used `onTap: null` rather than `IgnorePointer`/`AbsorbPointer` per your steer. Tests: failed share and failed reply-quote each route a card/frame-center tap to resend; delivered still opens the reel.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
realmeylisdev: Fixed in #6075 (214ff1686). Added a per-relay cooldown (`_lastSilentRepairAt` map + injectable `silentRepairCooldown`, default 60s = 2× the 15s OK window) so two consecutive timed-out publishes can't double-cycle a brownout relay. Confirmed the "late reply resets the silent clock" path doesn't save us — the repair runs in the tracker's timeout continuation and `forceReconnect` closes the old socket before any late OK can land. Cleared the cooldown entry in `remove`/`removeAll`/`removeTempRelay`. Exposure is concentrated on single-target publishes (repair only fires on `!outcome.confirmed`) — i.e. exactly the 10050 DM routing. Tests: two timeouts within the cooldown → one repair; a timeout after the cooldown elapses → repairs again (injected short cooldown, no wall-clock waits).

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
realmeylisdev: Fixed in #6075 (214ff1686). `_reconnectAndResubscribe` now replays the one-shot `_queries` (new `Relay.getQueries` accessor) alongside long-lived subscriptions, so a repaired query re-issues its REQ and can EOSE on the fresh socket; also fixed the adjacent `_queryCompleteCallbacks` leak in `unsubscribe`.

One correction to the "stays true forever" framing, for the record: in every production path the sole awaiter is `Nostr.queryEvents`, whose timeout handler calls `unsubscribe` and clears the stranded `_queries` entry — so on `main` today it's a bounded full-timeout wait returning partial results, not an indefinite hang. Your fix is still the right call: it turns that guaranteed timeout into an early completion on the reconnected socket. Test: a pending one-shot query on the zombie relay is re-sent on the new socket after the repair.

realmeylisdev:
@NotThatKindOfDrLiz thanks for the thorough post-merge pass — all nine findings are addressed in **#6075** (each re-verified against merged `main` first, each with a regression test). One commit per finding group:

| # | Finding | Commit | Notes |
|---|---------|--------|-------|
| F1 | Keycast global timeout regression | `91e491862` | Reverted single-op default to 30s (batch stays 30s). Scoped DM fail-fast left as a follow-up — no per-call timeout surface today. |
| F2 | Cancelled group send keeps fanning out | `afb7c10d2` | Pre-publish sibling re-read in the fan-out loop, mirroring `recoverFullSend`. |
| F3 | Retry exhaustion clobbers delivered row | `b8bbac850` | Fixed as hardening — I couldn't find a write path that produces the sent/pending hybrid today (it's transactionally coupled), so it guards a future/corrupt case. |
| F4 | Batch id not stamped on ingest | `afb7c10d2` | Extract + stamp for self-authored 64-hex tags only (forgery guard). |
| F5 | Throwing sweep kills the 30s heartbeat | `b8bbac850` | Re-arm from the `catch`; next clean pass self-cancels. |
| F6 | "Delete message" label mismatch | `d759ce9c8` | Renamed to "Stop trying" (+ 17 locales). Option B (delete-for-everyone) left for your call. |
| F7 | Failed video bubble tap opens the reel | `543f77e51` | `enableTap` threaded down; inner `onTap` null on failed send. |
| F8 | Silent-relay repair has no cooldown | `214ff1686` | Per-relay 60s cooldown (injectable). |
| F9 | Force-reconnect strands one-shot queries | `214ff1686` | Replay `_queries` on reconnect + fix the callback leak. Framed as bounded degradation — `queryEvents`' timeout already cleans up, so it wasn't an indefinite hang. |

Two of your findings (F3, F9) turned out to be confirmed mechanisms whose worst case is bounded by other guards; I've noted that honestly in each thread rather than overstating them, and fixed both anyway since the hardening is cheap and correct. Verification on #6075: analyze clean across app + `dm_repository`/`keycast_flutter`/`nostr_sdk`; app + all three package suites green; `dm_repository` coverage 93.59% ≥ gate.

PR #6051, opened 2026-07-11
fix(video-editor): preserve allow_audio_reuse when editing cached videos

### LINKED ISSUE #6045: fix(video-editor): preserve allow_audio_reuse when editing cached videos
Editing a published video can drop the `allow_audio_reuse` tag when the `VideoEvent` was rehydrated from JSON/cache and has empty `nostrEventTags`.

`VideoEditorNotifier.initFromPublishedVideo` currently seeds `allowAudioReuse` from `video.nostrEventTags.any(...)`. Since `nostrEventTags` is intentionally not persisted through `VideoEvent.toJson/fromJson`, cached or prefetched edit paths seed the toggle as `false` even if the original event had `['allow_audio_reuse', 'true']`.

PR #6026 preserves unowned raw tags during metadata-edit republish, but `allow_audio_reuse` is editor-owned and stripped/rebuilt from editor state, so this existing cache-path gap remains.

Expected:
- Editing a cached/prefetched video with `allow_audio_reuse=true` should seed the toggle ON.
- Saving without changing the toggle should preserve `['allow_audio_reuse', 'true']`.

Likely fix:
- Recover the raw event from `PersonalEventCacheService` when initializing edit state, or
- Add a persisted/stored field on `VideoEvent` for `allowAudioReuse`.

Add regression coverage for the cache-rehydrated edit path.

Context: follow-up from PR #6026.

hm21:
## What

Editing a published video could silently drop the `allow_audio_reuse` marker when the `VideoEvent` had been rehydrated from the JSON cache.

`VideoEditorNotifier.initFromPublishedVideo` seeded the toggle from `video.nostrEventTags`, but `nostrEventTags` is intentionally **not** persisted through `VideoEvent.toJson`/`fromJson`. So on any cached or prefetched edit path, `nostrEventTags` was empty and the toggle seeded `false` — even for a video the creator had marked reusable. Saving without touching the toggle then stripped `['allow_audio_reuse', 'true']` from the republished event.

## Why this approach

The issue suggested either recovering the raw event from `PersonalEventCacheService` or adding a dedicated persisted field. Both are heavier than needed: `rawTags` **is** already persisted through the cache round-trip, so the marker is recoverable from data we already keep.

This PR adds a `VideoEvent.allowAudioReuse` getter that reads the marker via the existing `_firstNostrTagValue` helper, which prefers live `nostrEventTags` and falls back to `rawTags`. The editor now seeds its toggle from that getter instead of scanning `nostrEventTags` directly. The derivation lives on the model where it belongs, and no new persisted state or cache-service coupling is introduced.

## Changes

- `packages/models`: add `VideoEvent.allowAudioReuse` getter (live-tag preferred, `rawTags` fallback).
- `video_editor_provider`: source the edit toggle from `video.allowAudioReuse`.
- Tests: model-level round-trip regression coverage + `initFromPublishedVideo` regression tests for the cache-rehydrated path.

## Testing

- `flutter test packages/models/test/src/video_event_allow_audio_reuse_test.dart` — 5 passing
- `flutter test test/providers/video_editor_provider_test.dart` — all passing (incl. 3 new `initFromPublishedVideo` cases)
- `flutter analyze` on changed files — clean

Closes #6045

### DIFF
--- a/mobile/packages/models/lib/src/video_event.dart
+++ b/mobile/packages/models/lib/src/video_event.dart
@@ -878,6 +878,16 @@ class VideoEvent {
     return pubkeys;
   }
 
+  /// Whether the creator marked this video's audio as reusable.
+  ///
+  /// Reads the editor-owned `allow_audio_reuse` marker. Prefers
+  /// [nostrEventTags] but falls back to [rawTags], so the flag survives a
+  /// JSON-cache round-trip — [nostrEventTags] is intentionally not persisted
+  /// through `toJson`/`fromJson`, but [rawTags] is, which keeps the edit form
+  /// from seeding the toggle `false` on a cache-rehydrated video.
+  bool get allowAudioReuse =>
+      _firstNostrTagValue('allow_audio_reuse') == 'true';
+
   /// Whether this video has any content warnings.
   bool get hasContentWarning => contentWarningLabels.isNotEmpty;
 
--- a/mobile/test/providers/video_editor_provider_test.dart
+++ b/mobile/test/providers/video_editor_provider_test.dart
@@ -1123,6 +1123,9 @@ void main() {
     }
 
     test('seeds allowAudioReuse from live nostrEventTags', () {
+      // rawTags is intentionally left empty so this isolates the live-tag
+      // source: if the live-path scan regressed, there is no fallback to
+      // mask it and the toggle would seed false.
       container
           .read(videoEditorProvider.notifier)
           .initFromPublishedVideo(
@@ -1130,7 +1133,6 @@ void main() {
               nostrEventTags: const [
                 ['allow_audio_reuse', 'true'],
               ],
-              rawTags: const {'allow_audio_reuse': 'true'},
             ),
           );
 
--- a/mobile/test/providers/video_editor_provider_test.dart
+++ b/mobile/test/providers/video_editor_provider_test.dart
@@ -1089,6 +1089,76 @@ void main() {
     });
   });
 
+  group('initFromPublishedVideo', () {
+    late ProviderContainer container;
+
+    setUp(() async {
+      SharedPreferences.setMockInitialValues({});
+      final prefs = await SharedPreferences.getInstance();
+      container = ProviderContainer(
+        overrides: [sharedPreferencesProvider.overrideWithValue(prefs)],
+      );
+    });
+
+    tearDown(() {
+      container.dispose();
+    });
+
+    VideoEvent buildVideo({
+      List<List<String>> nostrEventTags = const [],
+      Map<String, String> rawTags = const {},
+    }) {
+      return VideoEvent(
+        id: 'dddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd',
+        pubkey:
+            'cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc',
+        createdAt: 1778120201,
+        content: 'A published clip',
+        timestamp: DateTime.utc(2026, 5, 7),
+        title: 'Published clip',
+        videoUrl: 'https://media.divine.video/published.mp4',
+        nostrEventTags: nostrEventTags,
+        rawTags: rawTags,
+      );
+    }
+
+    test('seeds allowAudioReuse from live nostrEventTags', () {
+      container
+          .read(videoEditorProvider.notifier)
+          .initFromPublishedVideo(
+            buildVideo(
+              nostrEventTags: const [
+                ['allow_audio_reuse', 'true'],
+              ],
+              rawTags: const {'allow_audio_reuse': 'true'},
+            ),
+          );
+
+      expect(container.read(videoEditorProvider).allowAudioReuse, isTrue);
+    });
+
+    test('seeds allowAudioReuse from rawTags when the event was '
+        'rehydrated from cache (nostrEventTags dropped)', () {
+      container
+          .read(videoEditorProvider.notifier)
+          .initFromPublishedVideo(
+            buildVideo(rawTags: const {'allow_audio_reuse': 'true'}),
+          );
+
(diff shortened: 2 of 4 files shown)


mobile/packages/models/lib/src/video_event.dart
```diff
@@ -878,6 +878,16 @@ class VideoEvent {
...
   }
 
+  /// Whether the creator marked this video's audio as reusable.
+  ///
+  /// Reads the editor-owned `allow_audio_reuse` marker. Prefers
+  /// [nostrEventTags] but falls back to [rawTags], so the flag survives a
+  /// JSON-cache round-trip — [nostrEventTags] is intentionally not persisted
+  /// through `toJson`/`fromJson`, but [rawTags] is, which keeps the edit form
+  /// from seeding the toggle `false` on a cache-rehydrated video.
+  bool get allowAudioReuse =>
```
realmeylisdev, line 888: Correct layer and correct read. Reusing `_firstNostrTagValue` keeps live-tag authority with the `rawTags` fallback that closes #6045, and the exact `== 'true'` matches funnelcake's case-sensitive ClickHouse check (`arrayExists(... t[2]='true')`, `client.rs:12473`) and both write paths (`video_metadata_update_service.dart:330`, `video_event_publisher.dart:1360`).

mobile/test/providers/video_editor_provider_test.dart
```diff
@@ -1089,6 +1089,76 @@ void main() {
...
+
+    test('seeds allowAudioReuse from live nostrEventTags', () {
+      container
+          .read(videoEditorProvider.notifier)
+          .initFromPublishedVideo(
+            buildVideo(
+              nostrEventTags: const [
+                ['allow_audio_reuse', 'true'],
+              ],
+              rawTags: const {'allow_audio_reuse': 'true'},
```
realmeylisdev: Non-blocking (test-quality): this live-path case sets both `nostrEventTags` and `rawTags` to `'true'`, so it can't isolate that the *live* source seeded the toggle — dropping the `rawTags` line would make it a distinct live-only guard. Separately, nothing pins the `nostrEventTags`-over-`rawTags` precedence with a divergent value (e.g. live `['allow_audio_reuse','false']` + `rawTags` `'true'`), so a future reorder of `_firstNostrTagValue` wouldn't be caught here.

mobile/test/providers/video_editor_provider_test.dart
```diff
@@ -1089,6 +1089,76 @@ void main() {
...
+
+    test('seeds allowAudioReuse from rawTags when the event was '
+        'rehydrated from cache (nostrEventTags dropped)', () {
+      container
+          .read(videoEditorProvider.notifier)
+          .initFromPublishedVideo(
+            buildVideo(rawTags: const {'allow_audio_reuse': 'true'}),
+          );
+
+      // Regression (#6045): before the fix this seeded false because the
```
realmeylisdev, line 1150: Non-blocking (coverage): the 8 new tests all stop at seeding — none drives a save to prove a cache-rehydrated edit actually re-emits `['allow_audio_reuse','true']`. The seeding *is* the fix and is guarded here, so this is a symptom-vs-proxy gap rather than a missing guard — but an end-to-end test (rehydrated `VideoEvent` → `initFromPublishedVideo` → `updateVideo` → assert the tag is present) would lock the full chain.

realmeylisdev (approved):
**Verified end-to-end — approving.**

**Fix confirmed.** `nostrEventTags` is not persisted through `toJson`/`fromJson` but `rawTags` is (and `fromNostrEvent` writes every tag into it), so seeding the toggle from the new `VideoEvent.allowAudioReuse` getter — live tags first, `rawTags` fallback — correctly survives a cache round-trip. The seeded value flows `initFromPublishedVideo` (`video_editor_provider.dart:657`) → `updateVideo`, where `allow_audio_reuse` is stripped via `_editRebuiltTagNames` (`video_metadata_update_service.dart:53`) and re-added only when `editorState.allowAudioReuse` (`:330`) — so the marker is preserved on save. `allowAudioReuse` is provably the only `initFromPublishedVideo` field seeded from un-persisted state; the other seven read first-class `VideoEvent` fields.

**Cross-repo contract checked.** Mobile writes and reads the exact lowercase `['allow_audio_reuse','true']` that funnelcake's case-sensitive ClickHouse check (`check_audio_reuse_by_sha256`, `client.rs:12473`) and blossom's audio-reuse gating require. The edit republishes a replacement addressable event, so a dropped marker flips the sound to non-reusable in `videos_latest` — real server-side impact. Correct fix.

**Local checks:** models tests 5/5, provider tests 79/79, `flutter analyze` clean.

Inline notes are all **non-blocking** (nits + praise). One item worth a **separate follow-up** (pre-existing, not this marker): on a cache-only edit whose event is absent from `PersonalEventCacheService`, non-owned tags (audio `e`-tag, NIP-40 expiration, proofmode/c2pa, text-track) are still dropped via the `_sourceOriginalTags` → `const []` fallback (`video_metadata_update_service.dart:393`).

mobile/test/providers/video_editor_provider_test.dart
hm21: Both fixed in 174ad8746. Dropped the rawTags line from this test so it isolates the live source, and added a model-level precedence guard (live `false` + rawTags `true` → false) in `video_event_allow_audio_reuse_test.dart` — a reorder of `_firstNostrTagValue` now fails loudly there.

mobile/test/providers/video_editor_provider_test.dart
hm21: Leaving as-is by choice: the seeding is the fix and is already guarded, so driving `updateVideo` end-to-end adds publish/metadata-service mock scaffolding without new regression protection. Agreed it is a proxy — happy to add the full chain if we later want it pinned.

Issue #5838, opened 2026-07-04
test(vgv): extend merged-isolate guards to mobile/packages


realmeylisdev:
## Context

The #5738 heal-and-blame harness and the static guards it ships alongside (`check_shared_channel_overrides.sh`, `check_process_global_mutations.sh`, `check_http_overrides_isolation.sh`) all scan `mobile/test` only. But every package under `mobile/packages` runs its own fully-merged `very_good test --optimization` bundle (the reusable `very_good_workflows` `flutter_package.yml` defaults `test_optimization: true`, and no package workflow overrides it), so the same merged-isolate process-global leak class applies inside each package's suite.

## Known intra-package unrestored installs (blast radius = own package)

- `divine_camera` — ~9 unrestored platform installs
- `divine_quick_actions` — ~4
- `background_uploader` — ~1
- `divine_video_player` — ~4 `PathProviderPlatform` installs + never-nulled channels
- `infinite_video_feed` — body-level channel installs

## Scope

Extend the merged-isolate protection to `mobile/packages`: either point the existing guards at the package trees, or provide a package-local `flutter_test_config.dart` heal-and-blame equivalent (decide per-package whether that seam exists).

Related: #5738. Epic: #4337.

NotThatKindOfDrLiz:
Follow-up from PR #6060 review: the PR fixes the `<Platform>.instance` leaks it set out to handle, but two package merged-isolate globals should stay in this issue's backlog rather than getting lost:

- `mobile/packages/background_uploader`: the `MethodChannelBackgroundUploader` constructor still installs a process-global incoming method-call handler on the `background_uploader` channel. After `background_uploader_method_channel_test.dart` runs, the channel handler belongs to the last per-test constructed instance, not necessarily `BackgroundUploaderPlatform.instance`; restoring `.instance` alone does not restore handler ownership. This fits the existing "never-nulled channels" scope here.
- `mobile/packages/divine_camera/test/widgets/camera_preview_widget_test.dart`: the new tearDown restores `DivineCameraPlatform.instance`, but the `DivineCamera.instance` facade singleton can still carry `_state` and callbacks into later files unless the test also disposes it in tearDown. The sibling `divine_camera_test.dart` already uses `await DivineCamera.instance.dispose()` in tearDown, so the guard/fix pass should include non-platform facade singletons too.

Issue #5185, opened 2026-06-15
test(vgv): guard Tier-3/Tier-4 process-globals in merged isolate (residual of #5183)


NotThatKindOfDrLiz:
## Context

The core process-global isolation guardrail — the Tier-2 `.instance` platform singletons — **shipped in #5186, which closed #5183**. That added `mobile/scripts/check_process_global_mutations.sh` (hard-zero gate, wired into the `generated-files` CI job) alongside the existing `check_http_overrides_isolation.sh` (#5163), covering `PathProviderPlatform`, `ProVideoEditor`, `WebViewPlatform`, `UrlLauncherPlatform`, `VideoPlayerPlatform`, and `FirebasePlatform`.

This issue is **rescoped** (2026-07-13) to the residual tiers that #5186 explicitly deferred to a follow-up in its "Out of Scope" section. It is **not** a clean duplicate of #5183: #5183's scope was Tier-2 only and is fully done, whereas the Tier-3/Tier-4 seams below remain ungated. Per @NotThatKindOfDrLiz ("Part of #4337"), tracked under the testing epic.

## Remaining scope

**Tier-3 — debug-override globals** (the "restore" is a reset-to-default, not a capture-read, so the Tier-2 gate shape doesn't apply directly):

- `debugDefaultTargetPlatformOverride` — known ungated assign/reset in `mobile/test/screens/apps/nostr_app_sandbox_screen_test.dart` (assigns at lines 39/87/120/157, resets to `null` at 52/109/142/182). Inline reset without `try/finally`, so a failure between assign and reset leaks the override into later files in the merged isolate.

**Tier-4 — app-owned static seams:**

- `FlutterError.onError`
- `Bloc.observer` (the app installs `DivineBlocObserver`)
- image-cache seam (`PaintingBinding.instance.imageCache` tuning)

Per the earlier audit these have **no current leakers** — every assigner restores today — so this is a preventive gate, not a live-bug fix.

## Acceptance Criteria

- CI detects newly introduced untagged Tier-3/Tier-4 process-global mutations that can contaminate the VGV merged isolate, **or** each tier is documented as intentionally out-of-gate with a rationale near the guardrail.
- The known `debugDefaultTargetPlatformOverride` pattern in `nostr_app_sandbox_screen_test.dart` is guarded or given a within-file `try/finally` (or `addTearDown`) restore.
[...]

PR #5163, opened 2026-06-14
fix(test): stop manual acceptance tests leaking real network into the merged suite (#3137)

realmeylisdev:
## Summary

Two manual Docker acceptance tests were missing their `@Tags`, so they ran inside
the `very_good test --optimization` **merged unit isolate**. `upload_publish_e2e_acceptance_test.dart`'s
`setUpAll` does `HttpOverrides.global = null` (it's an intentional real-network test
against the local stack) and **never restores it**. Its own tests skip when the
Docker stack is absent — but `setUpAll` still runs, leaving flutter_test's HTTP mock
removed for **every subsequent test in the merged ordering**.

`RelayCapabilityService` then makes a real `https` GET to `staging-relay.divine.video`
(NIP-11 capability fetch), opening a real `RawSecureSocket`. The leaked connect-timer
is reported as a **"Pending timers"** failure, cross-attributed to whichever widget
test happens to be running at the pending-timer check — so it's **order-dependent**
and only surfaces on some `--test-randomize-ordering-seed` values (e.g. 98765).

## Root cause vs. symptom

The failing tests (`video_thumbnail_widget`, `video_explore_tile_nip05`,
`video_error_overlay`) are **victims**, not the cause — they don't do networking.
The cause is the un-restored global `HttpOverrides` leaking out of the untagged
manual tests.

## Fix

Tag both manual acceptance tests `@Tags(['skip_very_good_optimization', 'integration'])`
(matching their already-correct sibling `nip98_relay_acceptance_test.dart`) so they
stay out of the merged unit suite, with inline `// Permanent:` justifications. Ratchet
`vgv_tag_baseline.txt` 54 → 56 (these are legitimate, permanent exclusions — Docker /
real-network / global-`HttpOverrides`-mutating tests).

This also removes a real-network/backend dependency (and some socket-connect latency)
from the merged unit run.

## Verification

Ran the exact CI command (`very_good test --optimization --concurrency=2 --exclude-tags integration`)
at **seed 98765** (deterministically red before this change), plus seeds 42 and 11111:
**all green** — the 3 widget tests pass and no real network is attempted.

Relates to #5159 (this fixes the specific real-network leak found while de-flaking).
Relates to #3137.

## Prevention (new in this PR)

Adds `scripts/check_http_overrides_isolation.sh` (wired into the `generated-files`
CI job) so this regression class cannot recur: it fails CI if any `test/` file
assigns `HttpOverrides.global` without a real `@Tags(['skip_very_good_optimization'])`
[...]

Issue #4337, opened 2026-05-13
epic(testing): reduce skipped coverage debt and stabilize test infrastructure


NotThatKindOfDrLiz:
## Summary

Testing debt is currently spread across skipped tests, missing service coverage, flaky timing patterns, weak integration coverage, and CI enforcement gaps. This needs one parent issue so testing work can be sequenced intentionally instead of living as scattered audit fragments.

This issue is the source of truth for testing and CI debt reduction.

## Problem

Current problems include:
- too many skipped tests with no active execution path
- large service areas with weak or missing direct coverage
- test timing patterns that create flakiness or overhead
- insufficient integration coverage for important runtime flows
- CI coverage policy that is still inconsistent across scopes

## Scope

- reduce skipped test debt
- improve direct coverage for untested services and flows
- replace timing-based test patterns where appropriate
- restore or retire stale test suites intentionally
- tighten CI coverage expectations where it improves signal

## Existing Related Work

- `#3608`
- `#3609`
- `#3610`
- `#3612`
- `#3614`
- `#3616`
- `#3576`
- `#3577`

## Acceptance Criteria

- major testing debt categories are coordinated under one parent issue
- skipped/untested/flaky/CI coverage work has a clear execution home
- related audit leaves can be closed or superseded into this epic
- future test cleanup work has an obvious place to land

## Done When

- testing debt is organized into a manageable execution model
- the repo has fewer ambiguous “someday test cleanup” issues
- the highest-value testing improvements are decomposed into concrete child work

---

## Assignee
`@NotThatKindOfDrLiz`

## Weekly Priority
`P2`

## Weekly Priority Meaning
- `P1` means top priority for the week.
- `P2` means complete this week after P1 priorities are finished.

## Execution Order
1. define triage policy for skipped and stale tests
2. group and supersede scattered testing audit leaves
3. execute the highest-signal coverage and infrastructure fixes first

## Subgroup
- `@NotThatKindOfDrLiz` — DRI
- `@realmeylisdev` — CI/test execution support
- `@untreu2` — coverage and test cleanup support

## Coordination
The listed subgroup must review scope, sequencing, and ownership before implementation starts. Coordination may be synchronous or asynchronous. Do not wait for central coordination if the subgroup can resolve the question directly. Escalate only when blocked or when a testing policy decision is genuinely needed.

## Agent Execution Notes
Agents working from this issue must:
- treat this issue as the coordination source of truth for testing debt cleanup
[...]

PR #6059, opened 2026-07-13
test(vgv): gate Tier-3/Tier-4 process-globals in the merged isolate

realmeylisdev:
## Description

Extends `mobile/scripts/check_process_global_mutations.sh` beyond the six Tier-2 `<Singleton>.instance` seams (#5163/#5186) to the **Tier-3/Tier-4 residual**, so a future untagged test cannot silently leak one of these process-globals into the `very_good --optimization` merged isolate (an order-dependent, seed-dependent flake cross-attributed to an unrelated test).

Two detection classes, both HARD-ZERO (no baseline, no tolerated debt):

- **capture-restore** (appended to `GLOBALS`): `FlutterError.onError`, `Bloc.observer`. Append-only — every current assigner already snapshots the prior value and restores it, so this is zero-retrofit and zero-false-positive against `main`.
- **reset-to-default** (new mode): `debugDefaultTargetPlatformOverride`, `debugImageCacheOverride`, `debugUsesAppleAppStoreTipPolicyOverride`, `VideoEditorRenderService.renderVideoOverride` / `renderVideoToClipOverride`, `NativeProofModeService.proofFileOverride`, `NativeProofModeService.c2paSigningServiceFactoryOverride`, `InfiniteVideoFeed.debugIsSupportedOverride`. These have no prior runtime value to snapshot (the resting value is a compile-time literal, `null`), so a naive `GLOBALS` append would condemn every correct file. The new mode instead flags a file only when it installs a **non-default** value **and** never assigns the global back to its **default literal** within the file.

`GoogleFonts.config.allowRuntimeFetching` is intentionally **not** gated (benign-monotonic: the shared harness sets it `false` suite-wide and `false` is the resting value every test wants).

Adds a `--selftest` branch. No CI wiring change — the `generated-files` job already runs the real-tree scan. The separate `HttpOverrides.global` STRICT gate is left untouched.

**This is a purely preventive gate — the app merged isolate has no genuine leaker today.** Note on `debugDefaultTargetPlatformOverride`: it is a Flutter *foundation debug var*, so in a `testWidgets` test `debugAssertAllFoundationVarsUnset` already fails the test at end-of-body if it is left set (a leak there is loud, not silent, and only an inline reset works — that assertion runs before `tearDown`/`addTearDown`). The gate still earns its keep for it by covering plain `test()` usage (no framework guard); the other reset-to-default globals have no framework guard at all. The gate accepts an inline `= null`, so it is consistent with the required `testWidgets` pattern.

**Related Issue:** Relates to #5185 · Part of #4337

## Out of Scope

- Extending the gate to scan `mobile/packages/*/test` (each package is its own separate merged isolate) — tracked under **#5838**.
- Fixing the currently-latent **package-bundle** `<Platform>.instance` leakers (`divine_camera`, `divine_video_player`, `background_uploader`) — a separate PR targeting `main`.

## Verification

- `bash mobile/scripts/check_process_global_mutations.sh --selftest` → 18/18 fixtures pass, including multi-line `@Tags([...])`, nullable-promotion restores (`original!`), captured-id restore checks, trailing-comma forms, and the C2PA sibling.
- `/bin/bash mobile/scripts/check_process_global_mutations.sh --selftest` → 18/18 fixtures pass under Bash 3.2.57.
- `bash mobile/scripts/check_process_global_mutations.sh` → exit 0 on this branch (zero false-positives across every current capture-restore + reset-to-default assigner).
- Negative proof: injecting an unrestored `Bloc.observer` install and a `debugDefaultTargetPlatformOverride = TargetPlatform.iOS` (no reset) into a scratch untagged test makes the gate FAIL and name both, with class labels; removed after.

## Type of Change

### DIFF
--- a/mobile/scripts/check_process_global_mutations.sh
+++ b/mobile/scripts/check_process_global_mutations.sh
@@ -16,11 +16,16 @@
 #   1. CAPTURE-RESTORE (GLOBALS) — `<Singleton>.instance` and static hooks whose
 #      documented resting value is the *prior runtime value*. Every untagged
 #      *_test.dart that installs one must, within the same file, SNAPSHOT the
-#      original (a `<id> = <Global>` read) and RESTORE it (a `<Global> = <id>`
-#      bare-identifier write, e.g. in tearDown / tearDownAll / addTearDown).
-#      Unlike check_http_overrides_isolation.sh — which rejects even a correct
-#      restore because a nulled HttpOverrides 400-mock is itself the hazard while
-#      the file runs — a within-file restore PASSES here.
+#      original (a `<id> = <Global>` read) and RESTORE it by assigning the global
+#      back that captured value (`<Global> = <origId>`, e.g. in tearDown /
+#      tearDownAll / addTearDown). The restore RHS must be an identifier that
+#      holds the captured original — the snapshot id itself, or a local copied
+#      from it — NOT an arbitrary bare identifier: the dominant install shape is
+#      itself a bare-id write (`Bloc.observer = observer;`) and `<Global> = null`
+#      strands null, so tying the restore to the captured value is what makes the
+#      check real. Unlike check_http_overrides_isolation.sh — which rejects even a
+#      correct restore because a nulled HttpOverrides 400-mock is itself the
+#      hazard while the file runs — a within-file restore PASSES here.
 #
 #   2. RESET-TO-DEFAULT (RESET_TO_DEFAULT_GLOBALS) — top-level debug overrides /
 #      test-injection setters whose documented resting value is a compile-time
@@ -114,17 +119,50 @@ run_scan() {
       if grep -qE "^[[:space:]]*@Tags\([^)]*skip_very_good_optimization" "$f"; then
         continue
       fi
-      # SNAPSHOT: the global appears on the RHS of an assignment (captured local).
+      # SNAPSHOT: identifiers that hold a value READ FROM the global. Seed with
+      # the LHS of every `<id> = <Global>` capture, then grow transitively across
+      # `<id2> = <id1>` copies (a tearDown that promotes the snapshot into a local
+      # before restoring — e.g. WebViewPlatform's non-null-setter guard). Word
+      # boundary `.` is excluded so a `<Global>.instance = <id>` restore line is
+      # never itself mistaken for a copy that adds `instance` to the set.
+      # `|| true` on every grep|sed: under `set -o pipefail` a no-match grep
+      # (exit 1) would otherwise abort the whole scan.
+      local orig_ids
+      orig_ids=$(grep -oE \
+        "[A-Za-z_][A-Za-z0-9_]*[[:space:]]*=[[:space:]]*([A-Za-z_][A-Za-z0-9_]*\.)?${core}([^A-Za-z0-9_]|$)" \
+        <<<"$body" | sed -E 's/[[:space:]]*=.*//' | sort -u || true)
       local has_capture=0
-      if grep -qE "=[[:space:]]*([A-Za-z_][A-Za-z0-9_]*\.)?${core}([^A-Za-z0-9_]|$)" <<<"$body"; then
-        has_capture=1
-      fi
-      # RESTORE: the global assigned a BARE identifier (block `= original;` or
-      # arrow `=> <Global> = original)`), distinguishing a restore from a
-      # `<Global> = SomeFake()` install.
+      if [[ -n "$orig_ids" ]]; then has_capture=1; fi
+
+      local id1 copies grown
+      for _ in 1 2 3 4 5; do
+        grown="$orig_ids"
+        for id1 in $orig_ids; do
+          copies=$(grep -oE \
+            "(^|[^A-Za-z0-9_.])[A-Za-z_][A-Za-z0-9_]*[[:space:]]*=[[:space:]]*${id1}([^A-Za-z0-9_]|$)" \
+            <<<"$body" | sed -E 's/^[^A-Za-z_]*//; s/[[:space:]]*=.*//' || true)
+          if [[ -n "$copies" ]]; then
+            grown=$(printf '%s\n%s\n' "$grown" "$copies" | sort -u)
+          fi
+        done
+        if [[ "$grown" == "$orig_ids" ]]; then break; fi
+        orig_ids="$grown"
+      done
+
+      # RESTORE: `<Global> = <origId>` (block `= origId;` or arrow `=> ... = origId)`)
+      # where <origId> is one of the captured-original identifiers above. Tying the
+      # RHS to a captured value — not any bare identifier — rejects the bare-id
+      # install `<Global> = observer;` and the null-strand `<Global> = null;`, and
+      # excludes null/true/false for free (they can never be a capture LHS).
       local has_restore=0
-      if grep -qE "([A-Za-z_][A-Za-z0-9_]*\.)?${core}[[:space:]]*=[[:space:]]*[A-Za-z_][A-Za-z0-9_]*[[:space:]]*[;)]" <<<"$body"; then
-        has_restore=1
+      if [[ "$has_capture" -eq 1 ]]; then
+        local id_alt
+        id_alt="${orig_ids//$'\n'/|}"
+        if grep -qE \
+          "([A-Za-z_][A-Za-z0-9_]*\.)?${core}[[:space:]]*=[[:space:]]*(${id_alt})[[:space:]]*[;)]" \
+          <<<"$body"; then
+          has_restore=1
+        fi
       fi
       if [[ "$has_capture" -eq 0 || "$has_restore" -eq 0 ]]; then
         rel="${f#"$MOBILE_DIR"/}"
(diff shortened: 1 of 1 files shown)


mobile/scripts/check_process_global_mutations.sh
```diff
@@ -1,124 +1,275 @@
...
+      # SNAPSHOT: the global appears on the RHS of an assignment (captured local).
+      local has_capture=0
+      if grep -qE "=[[:space:]]*([A-Za-z_][A-Za-z0-9_]*\.)?${core}([^A-Za-z0-9_]|$)" <<<"$body"; then
+        has_capture=1
+      fi
+      # RESTORE: the global assigned a BARE identifier (block `= original;` or
+      # arrow `=> <Global> = original)`), distinguishing a restore from a
+      # `<Global> = SomeFake()` install.
+      local has_restore=0
+      if grep -qE "([A-Za-z_][A-Za-z0-9_]*\.)?${core}[[:space:]]*=[[:space:]]*[A-Za-z_][A-Za-z0-9_]*[[:space:]]*[;)]" <<<"$body"; then
```
hm21: **Blocking — the restore check is a no-op for the dominant install shape.** This regex accepts any `<Global> = <bareIdentifier>[;)]`, but the canonical install is `Bloc.observer = observer;` (8 of 9 sites, e.g. `divine_bloc_observer_test.dart:154`). That install line itself matches, so once a file has the capture snapshot the gate demands, deleting the real `addTearDown(() => Bloc.observer = previousObserver)` still passes green — the exact regression this gate exists to catch is invisible for its flagship new global. Repro: copy `divine_bloc_observer_test.dart` into a scan dir, drop only the `addTearDown` restore line → scan still exits 0.

Same regex, second hole: `null` matches `[A-Za-z_]...`, so `tearDown(() { FlutterError.onError = null; })` counts as a "restore" while actually stranding `null` (not the prior handler) — later merged suites silently swallow framework errors.

Fix: require the restore RHS to be the same identifier captured by a `<id> = <Global>` snapshot, and exclude `null`/`true`/`false` from the RHS class.

mobile/scripts/check_process_global_mutations.sh
```diff
@@ -1,124 +1,275 @@
...
-    fi
+      if ! grep -qE "(^|[^A-Za-z0-9_.])([A-Za-z_][A-Za-z0-9_]*\.)?${core}[[:space:]]*=([^=>]|$)" <<<"$body"; then
+        continue
+      fi
+      if grep -qE "^[[:space:]]*@Tags\([^)]*skip_very_good_optimization" "$f"; then
+        continue
+      fi
+      # RESET-TO-DEFAULT present: `<Global> = <default>` (block or arrow form).
+      local has_default_reset=0
+      if grep -qE "([A-Za-z_][A-Za-z0-9_]*\.)?${core}[[:space:]]*=[[:space:]]*${def}[[:space:]]*[;)]" <<<"$body"; then
```
hm21: **Blocking — false-positives `dart format` output, including remediation (b) below.** The terminator class is `[;)]`. The >80-col `addTearDown(() => VideoEditorRenderService.renderVideoToClipOverride = null);` is wrapped by the mandatory formatter into a trailing-comma form ending `= null,`, which matches neither this reset regex nor the line-126 restore regex → correct code fails the gate with the very message that recommended it. No existing file trips it yet, but the first person to follow the recipe will. Fix: widen the terminator to `[;,)]` at lines 126, 154, and 161.

mobile/scripts/check_process_global_mutations.sh
```diff
@@ -1,124 +1,275 @@
...
+    echo ""
+    echo "Remediation — pick one:"
+    echo "  (a) [capture-restore] Snapshot the original and restore it in the file:"
+    echo "        late <Type> original;"
+    echo "        setUp(() { original = <Global>; <Global> = <fake>; });"
+    echo "        tearDown(() { <Global> = original; });"
+    echo "  (b) [reset-to-default] Assign the global back to its default in a"
+    echo "      guaranteed-reachable tearDown/addTearDown (NOT an inline end-of-body"
+    echo "      line, which a mid-test throw would skip):"
+    echo "        addTearDown(() => debugDefaultTargetPlatformOverride = null);"
```
hm21: Remediation (b) uses `debugDefaultTargetPlatformOverride` as the `addTearDown` example — but the header note (lines 79-85) says that exact global needs an **inline** reset under `testWidgets`, because `debugAssertAllFoundationVarsUnset` runs before `addTearDown`. Internal contradiction; use a different global for the (b) example.

mobile/scripts/check_process_global_mutations.sh
```diff
@@ -1,124 +1,275 @@
...
+# before tearDown/addTearDown). This gate still earns its keep for it by covering
+# plain `test()` usage, which has no such framework guard; the other entries
+# below have no framework guard at all.
+RESET_TO_DEFAULT_GLOBALS=(
+  'debugDefaultTargetPlatformOverride:null'
+  'debugImageCacheOverride:null'
+  'debugUsesAppleAppStoreTipPolicyOverride:null'
+  'VideoEditorRenderService\.renderVideoOverride:null'
+  'VideoEditorRenderService\.renderVideoToClipOverride:null'
+  'NativeProofModeService\.proofFileOverride:null'
```
hm21, line 101: Coverage gap: `NativeProofModeService.c2paSigningServiceFactoryOverride` is declared three lines below `proofFileOverride` in the same lib file, same Class-2 shape (static nullable, default `null`), and is already mutated by an untagged merged test (`native_proofmode_service_test.dart:136`, reset :23). Gating one hook of the pair and not its sibling reads as an oversight given the HARD-ZERO goal — either add it or note why it's out.

mobile/scripts/check_process_global_mutations.sh
```diff
@@ -1,124 +1,275 @@
...
   'UrlLauncherPlatform\.instance'
   'VideoPlayerPlatform\.instance'
   'FirebasePlatform\.instance'
+  'FlutterError\.onError'
+  'Bloc\.observer'
 )
 
-violations=""
+# Class 2 — reset-to-default globals, `core:defaultLiteral` (colon-delimited so
+# bash 3.2 — the macOS pre-commit hook — needs no associative arrays). The
```
hm21: Nit: no git hook runs this script — only `mobile_ci.yaml:101`. `scripts/install-hooks.sh` doesn't reference it, so "the macOS pre-commit hook" is inaccurate. Reword to e.g. "macOS's default /bin/bash 3.2 for local runs."

mobile/scripts/check_process_global_mutations.sh
realmeylisdev: Fixed in adfb05509. The restore RHS is now tied to an identifier that holds a value read from the global — seeded by every `<id> = <Global>` capture and grown transitively across `<id2> = <id1>` copies — so the bare-id install `Bloc.observer = observer;` and the `= null` strand no longer count as restores (literals can never be a capture LHS, so they're excluded for free). Your exact repro (the real `divine_bloc_observer_test.dart` minus its `addTearDown`) now fails the gate, while the unmodified file still passes.

One deviation from the literal suggestion: I tie the restore to the *captured value* rather than the *exact snapshot identifier*, because `nostr_app_sandbox_screen_test.dart` legitimately restores through a copied local (`final original = originalWebViewPlatform;`) inside a non-null-setter guard — requiring the exact snapshot id would have false-positived it. The one-hop copy is covered by a bounded transitive closure. Self-tests added for both holes plus that indirection.

mobile/scripts/check_process_global_mutations.sh
realmeylisdev: Fixed in 36e20fb86. Widened the terminator `[;)]` → `[;,)]` at all three regexes (the restore check and both reset checks). Verified a wrapped `addTearDown(() => VideoEditorRenderService.renderVideoToClipOverride = null,)` now passes, and separately confirmed a wrapped *non-default* install with no reset still fails — so the comma widening doesn't weaken leaker detection. Self-tests added for both the wrapped restore and the wrapped reset forms.

mobile/scripts/check_process_global_mutations.sh
realmeylisdev: Fixed in 36e20fb86. The (b) example now uses `InfiniteVideoFeed.debugIsSupportedOverride` (a plain override with no framework guard) and explicitly notes that a foundation debug var like `debugDefaultTargetPlatformOverride` needs the inline reset instead under `testWidgets`, pointing back to the header note.

mobile/scripts/check_process_global_mutations.sh
realmeylisdev: Fixed in 36e20fb86. Added `NativeProofModeService.c2paSigningServiceFactoryOverride` to the reset-to-default list alongside its sibling. It's already mutated by the untagged `native_proofmode_service_test.dart` (install :136, reset `= null` in tearDown :23), so the real-tree scan stays exit 0, and a self-test leaker fixture pins the new coverage.

mobile/scripts/check_process_global_mutations.sh
realmeylisdev: Fixed in 36e20fb86. Reworded to note that CI runs this via `mobile_ci.yaml`, not a git hook, and that bash 3.2 is macOS's default `/bin/bash` for local runs.

realmeylisdev:
@hm21 thanks — all five addressed across `adfb05509` and `36e20fb86`, each replied inline and resolved:

- **restore no-op (blocking)** → restore is now tied to a value read from the global (snapshot ids grown across one-hop copies), so the bare-id install and `= null` no longer masquerade as restores. Your exact repro now fails; the unmodified file still passes. Generalized to copies so `nostr_app_sandbox_screen_test.dart`'s guarded restore-through-a-local isn't false-positived.
- **`dart format` trailing comma (blocking)** → terminator widened `[;)]` → `[;,)]` at all three regexes; wrapped install-with-no-reset still fails.
- **remediation (b) contradiction** → example switched to `InfiniteVideoFeed.debugIsSupportedOverride`, with a note that foundation debug vars need the inline reset.
- **c2pa sibling coverage gap** → `NativeProofModeService.c2paSigningServiceFactoryOverride` added to the reset-to-default list.
- **pre-commit-hook wording** → reworded to reference `mobile_ci.yaml` and macOS's default `/bin/bash`.

Verification: `--selftest` 15/15 (new fixtures for both holes, the copy-indirection, both trailing-comma forms, and the c2pa sibling), full-tree scan exit 0, shellcheck-clean, bash 3.2.

hm21 (approved):
LGTM

NotThatKindOfDrLiz:
Takeover update pushed in `6c5993567a`:\n\n- fixed skip-tag detection to whitespace-normalize `@Tags([...])`, so dart-format-wrapped multi-line `skip_very_good_optimization` annotations are treated like VGV treats them\n- kept commented-out tags ignored by the gate and documented that stricter behavior\n- accepted captured nullable-promotion restores like `WebViewPlatform.instance = original!;`\n- added selftest fixtures for multi-line tags, commented-out tags, and `original!` restores\n\nLocal verification from `mobile/`:\n\n- `bash scripts/check_process_global_mutations.sh --selftest` -> 18/18\n- `/bin/bash scripts/check_process_global_mutations.sh --selftest` -> 18/18 under Bash 3.2.57\n- `bash scripts/check_process_global_mutations.sh` -> pass\n- `bash -n scripts/check_process_global_mutations.sh` and `/bin/bash -n scripts/check_process_global_mutations.sh` -> pass\n\nI also attempted to wire `--selftest` into `.github/workflows/mobile_ci.yaml`, but GitHub rejected the push because the current OAuth token has `repo` but not `workflow` scope. The PR branch therefore includes the script hardening and fixtures; adding CI selftest wiring still requires a token/user with workflow-file write permission.

NotThatKindOfDrLiz (approved):
Approved after takeover pass. The remaining script-level issues I found are fixed in 6c5993567a: whitespace-normalized multi-line @Tags handling, nullable-promotion restore acceptance, and regression fixtures for both. Local selftest is 18/18 under bash and /bin/bash 3.2.57, the real-tree scan passes, syntax checks pass, and PR CI is green.\n\nI could not push the optional Mobile CI selftest wiring because this OAuth token lacks workflow scope; the current CI still runs the real-tree process-global scan successfully.

PR #6060, opened 2026-07-13
test(packages): stop leaked platform singletons across merged isolates

realmeylisdev:
## Description

The #5185 open-ended census (over `mobile/packages/*/test`, the #5838 axis) found six untagged package tests that install a process-global `<Platform>.instance` without restoring it, stranding the fake/mock into every later file in that package's `very_good --optimization` merged isolate — a **latent, seed-dependent** contamination (green today only because no later file reads the leaked global in a triggering order). The existing `check_process_global_mutations.sh` gate scans only `mobile/test`, so packages are unguarded.

**Fixes (capture the prior instance, restore in a guaranteed-reachable tearDown):**
- `divine_camera` (`DivineCameraPlatform.instance`):
  - `divine_camera_platform_interface_test.dart` — capture + `addTearDown` (also makes the file order-independent).
  - `divine_camera_test.dart` — restore in both group `tearDown`s (they previously only `dispose()`d the facade).
  - `widgets/camera_preview_widget_test.dart` — top-level `tearDown` restore.
- `divine_video_player` (`PathProviderPlatform.instance`):
  - `audio_track_test.dart`, `video_clip_test.dart` — capture + restore in the group `tearDown`s (which previously only deleted the temp dir).
- `background_uploader` (`BackgroundUploaderPlatform.instance`):
  - `background_uploader_test.dart` — capture the real default up front + top-level `tearDown` restore.
  - `background_uploader_method_channel_test.dart` — construct the real `MethodChannelBackgroundUploader` **per test in `setUp`** (was file-level). Capturing the default in the sibling reuses (and forces the lazy construction of) the singleton, whose constructor claims the `background_uploader` channel's incoming handler; per-test construction re-claims it so this suite's stream/buffer tests stay green. This is the identical MethodChannel-in-constructor shape already shipped on `main` in `divine_quick_actions`, restored the same way rather than tag-isolated.

**Related Issue:** Relates to #5838 (per-package merged-isolate guard) · surfaced by #5185

## Out of Scope

- Building a package-scanning gate (extend `check_process_global_mutations.sh` to `packages/*/test`) — that's **#5838**. This PR removes the current instances; #5838 prevents new ones.

## Verification

- Each affected package's `very_good test --optimization` passes across ordering seeds `1`, `12345`, `98765` (`background_uploader` additionally checked on `99999`, `424242`, and `random`).
- For `background_uploader`, leaving the sibling `_method_channel_test` at its old file-level construction fails its stream + buffered-terminal tests **deterministically** (handler steal), confirming the coupled per-test-construction change is required rather than assumed.
- Standalone `flutter test` passes for every changed file.
- `flutter analyze` clean on all changed files.

## Type of Change

### DIFF
--- a/mobile/packages/background_uploader/test/background_uploader_test.dart
+++ b/mobile/packages/background_uploader/test/background_uploader_test.dart
@@ -1,3 +1,10 @@
+// Tagged for its own isolate: this suite installs a fake
+// BackgroundUploaderPlatform.instance, and its only non-fake restore constructs
+// the real MethodChannelBackgroundUploader, which claims the plugin channel and
+// conflicts with background_uploader_method_channel_test in the merged run.
+@Tags(['skip_very_good_optimization'])
+library;
+
 import 'dart:async';
 
 import 'package:background_uploader/background_uploader.dart';
--- a/mobile/packages/background_uploader/test/background_uploader_test.dart
+++ b/mobile/packages/background_uploader/test/background_uploader_test.dart
@@ -1,10 +1,3 @@
-// Tagged for its own isolate: this suite installs a fake
-// BackgroundUploaderPlatform.instance, and its only non-fake restore constructs
-// the real MethodChannelBackgroundUploader, which claims the plugin channel and
-// conflicts with background_uploader_method_channel_test in the merged run.
-@Tags(['skip_very_good_optimization'])
-library;
-
 import 'dart:async';
 
 import 'package:background_uploader/background_uploader.dart';
@@ -56,6 +49,19 @@ class _FakeBackgroundUploaderPlatform extends BackgroundUploaderPlatform
 }
 
 void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  // Swapping in the fake below reassigns the process-global
+  // BackgroundUploaderPlatform.instance. Capture the real default up front and
+  // restore it after every test so the fake cannot strand a sibling suite in
+  // the merged-isolate (very_good --optimization) run. Reading the getter here
+  // reuses the already-lazily-constructed default; it does not build a new one.
+  final initialPlatform = BackgroundUploaderPlatform.instance;
+
+  tearDown(() {
+    BackgroundUploaderPlatform.instance = initialPlatform;
+  });
+
   group('BackgroundUploader.enqueue validation', () {
     late _FakeBackgroundUploaderPlatform fake;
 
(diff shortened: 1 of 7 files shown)


mobile/packages/background_uploader/test/background_uploader_test.dart
```diff
@@ -1,3 +1,10 @@
+// Tagged for its own isolate: this suite installs a fake
+// BackgroundUploaderPlatform.instance, and its only non-fake restore constructs
+// the real MethodChannelBackgroundUploader, which claims the plugin channel and
+// conflicts with background_uploader_method_channel_test in the merged run.
+@Tags(['skip_very_good_optimization'])
```
hm21: `divine_quick_actions` is the identical shape — its `MethodChannel*` also registers `setMethodCallHandler` in the constructor, and it has a sibling `*_method_channel_test` — yet it fixes the same leak on `main` with a capture-first top-level `tearDown` restore, not a tag.

Capture-first restore just reassigns the already-constructed instance; it does not construct a new `MethodChannel*`, so the "claims the channel / conflicts with the sibling" concern shouldn't apply — that only holds for a naive `instance = MethodChannelBackgroundUploader()`. Did restore actually fail here, or was it assumed? If restore works, prefer it for consistency with the other five files and with `divine_quick_actions` (per testing.md the file tag is the sanctioned path for a leaked `HttpOverrides.global`, not a `Platform.instance` leak).

realmeylisdev: You're right — switched to capture-first restore so `background_uploader` matches the other five files here and `divine_quick_actions` on `main`.

The "conflicts with the sibling" concern turned out to be real but not a reason to tag: capturing the default in `background_uploader_test` reuses (and forces the lazy construction of) the `MethodChannelBackgroundUploader` singleton, and its constructor claims the `background_uploader` channel's incoming handler. So restoring `_test` while leaving the sibling's *file-level* `MethodChannelBackgroundUploader()` in place makes the sibling's `native upload events…` and `…buffered and claimable` tests fail **deterministically** in the merged isolate (verified across seeds 1/12345/99999) — the handler was stolen. That's exactly the case `divine_quick_actions_method_channel_test` already handles by constructing its real instance per test in `setUp`; I've done the same here, so both files now merge cleanly under `very_good test --optimization`. 00c474d24

hm21 (approved):
LGTM

NotThatKindOfDrLiz (approved):
Approved. Current head is green, mergeable, and the prior platform-singleton review thread is resolved.

Issue #5738, opened 2026-07-01
test(vgv): centralize restoration for process-global test state


NotThatKindOfDrLiz:
## Problem

PR #5725 fixed a narrow `flutter_secure_storage` test-ordering failure by restoring the shared secure-storage MethodChannel handler instead of clearing it. That PR does not leave direct debt behind, but it is another example of a recurring failure class in the mobile test suite: tests running under `very_good test --optimization` share one isolate, so process-global state leaks can cascade into unrelated later suites depending on random ordering.

Known prior examples:

- #5725: `auth_service_signout_cleanup_test.dart` cleared the global `plugins.it_nomads.com/flutter_secure_storage` handler in teardown. Later auth/key-manager tests relying on the shared `setupTestEnvironment()` handler hit `MissingPluginException` on Linux CI.
- #3098 / #3082: `video_thumbnail_service_test.dart` had to move a MethodChannel mock from file-level setup to per-test `setUp`/`tearDown` to fix VGV isolation behavior.
- #5163 / #5159: untagged manual acceptance tests mutated `HttpOverrides.global`, removing Flutter test's HTTP mock and causing unrelated widget tests to fail with pending timers on some seeds.
- Notification/layout tests have also needed explicit restoration of process-global test surface state under `very_good test --optimization`.

The pattern is broader than secure storage: any test that mutates process-global state can create order-dependent failures in the merged isolate. Current fixes are ad hoc and require each PR author to remember the correct restore behavior for each global.

## Goal

Make process-global test state overrides explicit, easy to restore correctly, and harder to leak into later tests under VGV's merged-isolate runner.

## Scope

Audit and improve the mobile test harness around process-global state, especially:

- Flutter MethodChannel handlers installed with `setMockMethodCallHandler`, including handlers cleared with `null`.
- Shared platform-channel defaults installed by `mobile/test/test_setup.dart`, especially:
  - `plugins.it_nomads.com/flutter_secure_storage`
  - `openvine.secure_storage`
  - path provider channels
  - image picker channels
  - device info channels
- Test-only process globals such as:
  - `HttpOverrides.global`
  - widget test surface size / physical size overrides
  - singleton service test instances where first-test state can inherit previous-file state

## Suggested Implementation

1. Audit current direct global mutations.

   Useful starting searches:

   ```sh
[...]

PR #6061, opened 2026-07-13
test(vgv): gate per-package merged-isolate channel leaks (#5838)

realmeylisdev:
## Description

Extends the merged-isolate leak protection from `mobile/test` to **`mobile/packages`** (#5838), for the **channel-handler** class. Each package runs its own `very_good test --optimization` merged isolate (the reusable `flutter_package.yml` defaults `test_optimization: true`), and — unlike `mobile/test` — **no package has a `flutter_test_config.dart` heal-and-blame harness**. So a package test that installs a `MethodChannel`/`EventChannel` mock handler without tearing it down leaks it into every later file in that package's suite (an order-dependent flake confined to that package).

Adds **`mobile/scripts/check_package_channel_isolation.sh`** — a shrink-only LIST ratchet, the packages analogue of `check_shared_channel_overrides.sh` (#5738), built on the same `lib/list_ratchet.sh` engine. It freezes the current package `*_test.dart` **entries** that install a channel handler (`setMockMethodCallHandler`/`setMockStreamHandler`) **without nulling any back down**, and fails CI on a **new** one. Wired into the `generated-files` job next to the existing guards.

**Detector + exclusions:** only `*_test.dart` entries under `packages/*/test` (non-`*_test.dart` helpers like `test_setup.dart` / `helpers/*.dart` install suite-wide mocks by design and are excluded by the `--include` filter, exactly like the app's `test_setup.dart`); files that null ≥1 handler are treated as disciplined; `@Tags(['skip_very_good_optimization'])` files run in their own isolate and are exempt.

**Baseline (3 current entries, for incremental cleanup):** `divine_video_player` controller + widget tests, `nostr_key_manager/secure_key_storage_delete_test`. Remedy per entry: a `setMockMethodCallHandler(channel, null)` teardown, or the skip tag. All are currently **latent** (empirically green) — this prevents the next one and makes the hygiene explicit.

The "disciplined" check flattens newlines before matching, so a `dart format`-wrapped `setMockMethodCallHandler(channel, null)` teardown counts (per review — `permissions_service` and `infinite_video_feed` null their handlers across wrapped lines and are therefore **not** leakers; both dropped from the baseline).

**Related Issue:** Relates to #5838 · Part of #4337

## Out of Scope

- **Singletons** are guarded separately: PR #6060 fixes the current package singleton leakers, and a **hard-zero singleton mode** for this gate follows once #6060 merges (kept out here to stay independent of #6060 — no baseline coordination).
- A **runtime per-package `flutter_test_config.dart` heal-and-blame harness** (a shared `vgv_test_harness` package + thin per-package config) is a stronger future option, noted in the plan; this PR ships the static ratchet floor.

## Verification

- `bash mobile/scripts/check_package_channel_isolation.sh` → exit 0 (bootstrap; baseline frozen).
- `UPDATE_BASELINE=1 bash …` → wrote 3 entries.
- Negative proof: injecting a new un-torn-down `setMockMethodCallHandler` into a scratch package test makes the gate FAIL with `NEW`, naming the file + remedy; removed after → green.
- Empirical: `infinite_video_feed`, `divine_quick_actions`, `divine_video_player` `very_good test --optimization` pass across seeds 1/12345/98765 (`divine_video_player`'s baseline leakers latent; `infinite_video_feed` tears its handlers down).
- `shellcheck`-consistent with the sibling `check_shared_channel_overrides.sh` (identical `source`-induced SC2034 notes; no shellcheck CI).

## Type of Change


mobile/scripts/check_package_channel_isolation.sh
```diff
@@ -0,0 +1,90 @@
...
+  for f in $(grep -rlE 'setMockMethodCallHandler\(|setMockStreamHandler\(' \
+      "$MOBILE_DIR/packages" --include='*_test.dart' 2>/dev/null \
+      | grep -E '/test/' | LC_ALL=C sort); do
+    rel="${f#"$MOBILE_DIR"/}"
+    # Tagged out of the merge (own isolate) → cannot contaminate the bundle.
+    if grep -qE '^[[:space:]]*@Tags\([^)]*skip_very_good_optimization' "$f"; then
+      continue
+    fi
+    # Disciplined: nulls at least one handler back down (paired teardown).
+    if grep -qE 'setMockMethodCallHandler\([^,]*,[[:space:]]*null|setMockStreamHandler\([^,]*,[[:space:]]*null' "$f"; then
```
hm21: This single-line "disciplined" check misses multi-line teardowns. When the channel is a long inline expression, `dart format` wraps `setMockMethodCallHandler(channel, null)` across lines, so this regex never matches even though the file does null the handler.

`permissions_service_test.dart` (baseline entry 5) is exactly this: its `tearDown` (lines 37-44) nulls `const MethodChannel('openvine/native_camera')` — the same channel `mockNativeCamera` installs — so it's disciplined, not a leaker. Two effects: (1) it contradicts the baseline's own "installs … without nulling any back down" invariant; (2) it can never be shrunk out via the documented remedy, since it *already has* that tearDown — only a reformat (impossible under 80-char `dart format` with that channel) or a `skip_very_good_optimization` tag it doesn't need would clear it. The `NEW_HINT` example remedy hits the same wall with a long channel.

Suggest making the null-teardown match tolerant of line-wrapping (e.g. join continuation lines first) and dropping `permissions_service_test.dart` from the baseline.

realmeylisdev: Good catch, and it's actually two entries, not one. Fixed in f61d1216e by flattening newlines before the "disciplined" match:

```bash
if tr '\n' ' ' <"$f" | grep -qE 'setMockMethodCallHandler\([^,]*,[[:space:]]*null|...'; then
```

- **`permissions_service_test.dart`** — nulls `openvine/native_camera` in its `tearDown` (wrapped 40–43).
- **`infinite_video_feed_test.dart`** — nulls handlers in a `dispose()` helper (wrapped 109–123).

Both were disciplined-but-wrapped, so I dropped them from the baseline; 3 genuine leakers remain (`divine_video_player` controller/widget, `nostr_key_manager`).

Kept `[^,]*` comma-bounded on purpose: `divine_video_player_widget_test.dart` installs `setMock…(channel, (call) async => null)`, and a looser match would read that handler-returns-null install as a teardown. Verified both directions on a scratch probe — a genuine un-torn-down install is still caught; a wrapped `(channel, null)` teardown is now exempted. The `NEW_HINT` wrapped-teardown remedy is now accurate too.

hm21 (approved):
LGTM

PR #5674, opened 2026-06-30
feat(profile): route kind 0 publish through publishEventAwaitOk for relay-confirmed saves

hm21:
## Description

`NostrClient.sendProfile` reported success as soon as the relay pool accepted the WebSocket frame, before any relay confirmed with a NIP-20 `OK true`. A kind 0 profile save could therefore report success even when every relay then returned `OK false` (rate limit, PoW required, auth required): the user saw the edit succeed in-app, but it never persisted.

This routes kind 0 profile publishes through `publishEventAwaitOk` and maps its `PublishOutcome` back onto the existing `PublishResult` variants, so `ProfileRepository` keeps its exhaustive `switch` while opting into the stronger confirmation path:

- at least one relay returns `OK true` -> `PublishSuccess`
- no relay reached even after retry -> `PublishNoRelays` (preserves the dedicated no-relays signal from #3469)
- relays reached but all reject, time out, or signing/send fails -> `PublishFailed` -> `ProfilePublishFailedException` -> `ProfileEditorStatus.failure`

The old fire-and-forget `sendProfile` path is removed so profile publishing has one confirmed code path and no dead public API.

**Design note (why the mapping lives in `sendProfileAwaitOk`, not the repository):** keeping the return type `PublishResult` means the repository keeps its clean sealed-class switch and still receives the sent `Event` for `UserProfile.fromNostrEvent`. The rejection-vs-timeout distinction is a transport detail; both mean "not persisted", so it is collapsed to `PublishFailed` at the client boundary that owns it, and tested there.

**Related Issue:** Closes #3541

## Verification

- `flutter analyze packages/nostr_client packages/profile_repository` - no issues
- `flutter test packages/nostr_client/test/src/nostr_client_test.dart` - 146 pass
- `flutter test packages/profile_repository/test/src/profile_repository_test.dart --plain-name "saveProfileEvent"` - 22 pass
- `flutter test test/blocs/profile_editor/profile_editor_bloc_test.dart` - 93 pass
- pre-push hook - analyzer + changed-file tests pass

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #3528, opened 2026-04-28
fix(profile): kind 0 publish on no-relays fails fast with generic error

untreu2:
## Description
When a user had no Nostr relays connected, the kind 0 publish step in `_saveProfile` failed silently with a generic "Failed to publish profile" error, giving the user no indication that connectivity was the cause. This fix distinguishes the no-relays case at every layer — a new `NoRelaysConnectedException` is thrown by the repository when `connectedRelays` is empty after the publish attempt, a new `ProfileEditorError.noRelaysConnected` enum value is emitted by the BLoC, and the UI surfaces a specific "Couldn't reach the network. Check your connection and try again." message in all 14 supported languages.

**Related Issue:** Closes #3469, Closes #3152

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #3468, opened 2026-04-27
fix(profile_repository): cap name-server HTTP calls at 10s timeout

realmeylisdev:
Relates to #3152 and #3161.

## Description

Hardens the three `names.divine.video` HTTP calls in
`profile_repository.dart` against a slow or unresponsive name-server
by adding an explicit 10s `.timeout(_nameServerHttpTimeout)`. Without
this, when the endpoint stops responding the call hangs until the
platform's TCP timeout (~20s on Android), surfacing as a long wait
followed by a generic error.

**Defense-in-depth, not a fix for the reported tickets.** A deeper
investigation revealed that the actual symptom in the linked tickets —
"can't save username, error after waiting" — is reproduced by a
*different* failure mode: kind 0 profile publish fails fast when no
relays are connected, and the BLoC short-circuits before reaching
the username claim path that this PR touches. Field log evidence
shows the BLoC emitting `ProfilePublishFailedException` within
~400ms per attempt — not a 20-second TCP hang. The "20 seconds" in
the original report was almost certainly cumulative across multiple
retries, not a single network timeout.

This PR still ships on its own merits — it eliminates a real, if
rare, 20-second hang for any user whose `names.divine.video`
specifically is degraded — and lays groundwork for the proper fix
which now tracks separately in #3469.

## Why this is not a fix for the reported tickets

`ProfileEditorBloc._saveProfile` runs in three steps:

1. Publish kind 0 profile event via Nostr relays.
2. (Skipped if no username.)
3. POST `/api/username/claim` to `names.divine.video` to bind the
   username to the user's pubkey.

If step 1 throws, the BLoC emits `ProfileEditorError.publishFailed`
and **returns before step 3 runs**
(`profile_editor_bloc.dart:392-419`). Step 1 throws
`ProfilePublishFailedException` whenever
`NostrClient.sendProfile()` returns null, which it does when no
relays are connected after a single reconnection attempt
(`nostr_client.dart:254-262`).

The duplicate ticket's log buffer captures the failing branch
directly:

```
17:36:21  ❌ No connected relays - cannot create subscription
17:37:08  ❌ Failed to publish profile: ProfilePublishFailedException
17:37:13  ❌ Failed to publish profile: ProfilePublishFailedException
...
17:38:02.589  📝 saveProfile: displayName=bhecky, username=bhecky
17:38:02.995  ❌ Failed to publish profile: ProfilePublishFailedException  (406 ms)
```

Each individual attempt fails in well under a second. The HTTP claim
[...]

PR #3175, opened 2026-04-20
feat(auth): publish bootstrap kind:10002 when NIP-65 discovery returns empty

realmeylisdev:
## Summary
- When indexer NIP-65 discovery returns empty for the signed-in user, `AuthService` now self-publishes a minimal kind:10002 pointing at the current environment's primary relay and broadcasts it to the three indexer relays the client queries.
- Guarded by a one-shot SharedPreferences flag per `(device, pubkey)`; no-op when no identity / callback / successful publish, so failures remain retriable on next login.
- Wired into the existing `nostr_client_provider` plumbing at both client-creation sites (initial build + account switch) with matching teardown.
- **Env-aware relay URL** (#3183, folded in): the bootstrap event's `r` tag and its target-relay list are both sourced from `authEnv.relayUrl` via a new optional `primaryRelayUrl` ctor param on `AuthService`. Non-prod builds (staging, POC, test, local) no longer advertise the production relay to public indexers. Default preserves back-compat for callers that don't inject a URL.

## Why
Divine/Keycast-provisioned accounts never publish a kind:10002 list, which is the upstream contributor behind #3152, #3161, #3162 (see investigation in issue thread). The proper fix lives server-side in [keycast#94](https://github.com/divinevideo/keycast/issues/94) — publish kind:10002 at provisioning. This PR is the complementary client-side safety net that also backfills accounts provisioned before that ships.

## Test plan
- [x] `flutter test test/services/auth_service_bootstrap_relay_list_test.dart` — 10 tests (9 original + 1 new for #3183) cover empty/throw/success paths, flag, identity, callback, signer-null, publish-false, callback-throws, signer-timeout, and env-URL injection
- [x] `flutter test test/services/` — 1773 tests pass, 0 regressions
- [x] `flutter test test/providers/` — 481 tests pass, 0 regressions
- [x] `flutter analyze lib test` — clean
- [x] `dart format --set-exit-if-changed` — clean
- [x] Pre-commit hook passed
- [x] Assertion tightening (`75b6f955e`): happy-path test now bounds `event.createdAt` against a capture window (guards against replaceable-event supersession by a stale/zero timestamp); signer-timeout test uses `late bool` + non-nullable `expect` so a future regression to async `SharedPreferences.getInstance` fails loudly instead of silently passing on an uninitialized variable.

## Risk
[...]

PR #6063, opened 2026-07-13
fix(profile): offline-tolerant username save (#3161)

### LINKED ISSUE #3152: fix: Can't save username
### Summary

Divine User, 18 Apr 2026, 09:55

Can't save username

Tried updating username, failed.

App Version: 1.0.8+2513

### Steps to Reproduce
Enter a username, it is available, hit save changes but get an error after about 20 seconds of waiting.

### Expected Behavior
My username should have updated.

### Device Information
- **platform:** android
- **model:** XQ-CC54
- **manufacturer:** Sony
- **version:** 14
- **sdkInt:** 34
- **brand:** Sony

**Current Screen:** SupportCenterScreen
**User Pubkey:** 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250

### Recent Logs (errors/warnings + recent context)
```
[2026-04-17T22:52:29.099135] [ERROR] (CrashReporting) Failed to initialize crash reporting: [core/duplicate-app] A Firebase App named "[DEFAULT]" already exists
[2026-04-17T22:52:29.163032] [WARNING] [VIDEO] (VideoEventService) Cannot subscribe - Nostr service not initialized (will retry when ready)
[2026-04-17T22:52:31.428255] [WARNING] [SYSTEM] Zendesk JWT: SDK not initialized
[2026-04-17T22:52:31.880285] [WARNING] [SYSTEM] Zendesk JWT: SDK not initialized
[2026-04-17T22:52:31.998264] [WARNING] [AUTH] (RelayDiscoveryService) ⚠️ No relay list found for npub1w2fd0sdyh6h8dmgm8xzzhyxsfnfufevuu934e2pg6sef8gclxfgqvfgmfu on any indexer
[2026-04-17T22:52:31.998276] [WARNING] [AUTH] (AuthService) ⚠️ No relay list found for user on any indexer — connecting to safe DM-friendly fallback relay set
[2026-04-17T22:52:32.154874] [WARNING] [VIDEO] (VideoEventService) Default video system is disabled - users will see real content only
[2026-04-17T22:52:36.856831] [WARNING] [VIDEO] (VideoEventService) Default video system is disabled - users will see real content only
[2026-04-17T22:53:06.861963] [ERROR] [VIDEO] (VideoEventService) ⏰ TIMEOUT: No events received for SubscriptionType.profile after 30 seconds!
[2026-04-17T22:53:06.862351] [WARNING] [VIDEO] (VideoEventService) ⏰ FEED TIMEOUT for profile:
=== FEED LOADING TIMEOUT DIAGNOSTIC ===
Subscription Type: profile
Timeout Duration: 30085ms
Relay Connected: true
Network Online: true
Connected Relay Count: 4

Filters:
  Filter 0:
    Kinds: [34236]
    Authors: 1 authors
    First author: 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
    Tags: 0 tags
    Since: null
    Until: null
    Limit: 50
  Filter 1:
    Kinds: [16]
    Au... [truncated]
[2026-04-17T22:53:57.952516] [DEBUG] [SYSTEM] (RelayNotificationsProvider) RelayNotifications: Resuming auto-refresh timer
[2026-04-17T22:53:57.953795] [INFO] (ProfileScreenRouter) 🧭 ProfileScreenRouter.build
[2026-04-17T22:53:57.955610] [INFO] [VIDEO] (VideoControllerCleanup) 🧹 Route type changed (RouteType.videoRecorder → RouteType.profile), disposing all video controllers
[2026-04-17T22:53:57.955613] [DEBUG] [VIDEO] (VideoControllerCleanup) 🧹 Invalidating all video controllers
[2026-04-17T22:53:57.968485] [DEBUG] [SYSTEM] (RelayNotificationsProvider) RelayNotifications: Pausing auto-refresh timer (no listeners)
[2026-04-17T22:53:57.969438] [INFO] (ProfileScreenRouter) 🧭 ProfileScreenRouter.build
[2026-04-17T22:53:57.974715] [DEBUG] [SYSTEM] (RelayNotificationsProvider) RelayNotifications: Resuming auto-refresh timer
[2026-04-17T22:53:57.992176] [DEBUG] [SYSTEM] (FollowRepository) Fetching follower stats for: 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.082961] [DEBUG] [SYSTEM] (FollowRepository) REST API follower stats: 0 followers, 0 following for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.186279] [DEBUG] [SYSTEM] (FollowRepository) Indexer wss://user.kindpag.es returned 0 followers for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.273840] [DEBUG] [SYSTEM] (FollowRepository) Indexer wss://purplepag.es returned 0 followers for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.390231] [DEBUG] [SYSTEM] (FollowRepository) Indexer wss://relay.nos.social returned 0 followers for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.390311] [INFO] [SYSTEM] (FollowRepository) Indexer followers counts: [0, 0, 0], using 0 for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.404512] [WARNING] [VIDEO] (VideoRecorderScreen) 📹 Failed to clear overlay visibility on dispose: Bad state: Using "ref" when a widget is about to or has been unmounted is unsafe.
Ref relies on BuildContext, and BuildContext is unsafe to use when the widget is deactivated.
To safely refer to the state of providers inside State.dispose(), save the provider state in a field of your State class.
[2026-04-17T22:53:58.404516] [DEBUG] [VIDEO] (VideoRecorderNotifier) 🧹 Destroying video recorder
[2026-04-17T22:53:58.404543] [INFO] [VIDEO] (VideoRecorderScreen) 📹 Disposed
[2026-04-17T22:53:58.405125] [INFO] [VIDEO] (CameraMobileService) 📷 Setting remote record control: disabled
[2026-04-17T22:53:58.409022] [DEBUG] [VIDEO] (CameraMobileService) 🎮 Remote record trigger callback cleared
[2026-04-17T22:53:58.409026] [DEBUG] [VIDEO] (VideoRecorderNotifier) 🎮 Remote record control disabled
[2026-04-17T22:53:58.409030] [INFO] [VIDEO] (CameraMobileService) 📷 Disposing mobile camera
[2026-04-17T22:54:00.995047] [DEBUG] [SYSTEM] (ContactListQuery) ⏰ Fallback timeout reached (3s)
[2026-04-17T22:54:00.995103] [DEBUG] [SYSTEM] (ContactListQuery) ✅ Immediate completion triggered: early=false, received=0, expected=null
[2026-04-17T22:54:05.011558] [INFO] [UI] (Navigation) User tapped settings gear
[2026-04-17T22:54:05.011718] [INFO] [UI] (OverlayVisibility) Page opened
[2026-04-17T22:54:05.012218] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/settings, authState=authenticated
[2026-04-17T22:54:05.012618] [INFO] [SYSTEM] (Route) CTX derive: type=RouteType.settings npub=null index=null
[2026-04-17T22:54:05.012706] [INFO] [VIDEO] (VideoControllerCleanup) 🧹 Route type changed (RouteType.profile → RouteType.settings), disposing all video controllers
[2026-04-17T22:54:05.012717] [DEBUG] [VIDEO] (VideoControllerCleanup) 🧹 Invalidating all video controllers
[2026-04-17T22:54:05.022911] [INFO] (ScreenAnalytics) 📱 Screen load started: settings
[2026-04-17T22:54:05.022916] [DEBUG] [UI] (PageLoadObserver) Page push tracked: settings
[2026-04-17T22:54:05.022928] [DEBUG] [VIDEO] (VideoControllerCleanup) 🛡️ Skipping duplicate invalidation (10ms since last)
[2026-04-17T22:54:05.022930] [INFO] [SYSTEM] (VideoStopNavigatorObserver) 📱 Navigation didPush to route: settings - stopped all videos
[2026-04-17T22:54:05.030468] [INFO] (ProfileScreenRouter) 🧭 ProfileScreenRouter.build
[2026-04-17T22:54:05.045303] [INFO] (ScreenAnalytics) ✅ Screen content visible: settings in 22ms
[2026-04-17T22:54:05.045407] [INFO] (PagePerf) PERF: settings — visible: 22ms
[2026-04-17T22:54:05.045987] [INFO] [AUTH] (AuthService) getKnownAccounts: raw=190 chars
[2026-04-17T22:54:05.473196] [DEBUG] [SYSTEM] (RelayNotificationsProvider) RelayNotifications: Pausing auto-refresh timer (no listeners)
[2026-04-17T22:54:05.476545] [INFO] (ProfileScreenRouter) 🧭 ProfileScreenRouter.build
[2026-04-17T22:54:06.167790] [DEBUG] [SYSTEM] (FollowRepository) Indexer follower pubkeys: 0 for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:54:06.167823] [INFO] [SYSTEM] (FollowRepository) Followers for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250: API=0, relays=0, indexers=0, merged=0
[2026-04-17T22:54:10.602218] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/support-center, authState=authenticated
[2026-04-17T22:54:10.602646] [INFO] [SYSTEM] (Route) CTX derive: type=RouteType.supportCenter npub=null index=null
[2026-04-17T22:54:10.608410] [INFO] (ScreenAnalytics) 📱 Screen load started: support-center
[2026-04-17T22:54:10.608419] [DEBUG] [UI] (PageLoadObserver) Page push tracked: support-center
[2026-04-17T22:54:10.608438] [DEBUG] [VIDEO] (VideoControllerCleanup) 🧹 Invalidating all video controllers
[2026-04-17T22:54:10.608452] [INFO] [SYSTEM] (VideoStopNavigatorObserver) 📱 Navigation didPush to route: support-center - stopped all videos
[2026-04-17T22:54:10.649917] [INFO] (ScreenAnalytics) ✅ Screen content visible: support-center in 41ms
[2026-04-17T22:54:10.650067] [INFO] (PagePerf) PERF: support-center — visible: 41ms
[2026-04-17T22:54:14.865130] [DEBUG] [SYSTEM] (VideoStopNavigatorObserver) 📱 Skipping video disposal for popup route: null
[2026-04-17T22:55:18.545894] [INFO] [SYSTEM] Collecting bug report diagnostics
```

### Environment

**Platform:** 
**Build Number:** 
**OS Version:** 

### Steps to Reproduce

Enter a username, it is available, hit save changes but get an error after about 20 seconds of waiting.

### Actual Result

Divine User, 18 Apr 2026, 09:55

Can't save username

Tried updating username, failed.

App Version: 1.0.8+2513

### Steps to Reproduce
Enter a username, it is available, hit save changes but get an error after about 20 seconds of waiting.

### Expected Behavior
My username should have updated.

### Device Information
- **platform:** android
- **model:** XQ-CC54
- **manufacturer:** Sony
- **version:** 14
- **sdkInt:** 34
- **brand:** Sony

**Current Screen:** SupportCenterScreen
**User Pubkey:** 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250

### Recent Logs (errors/warnings + recent context)
```
[2026-04-17T22:52:29.099135] [ERROR] (CrashReporting) Failed to initialize crash reporting: [core/duplicate-app] A Firebase App named "[DEFAULT]" already exists
[2026-04-17T22:52:29.163032] [WARNING] [VIDEO] (VideoEventService) Cannot subscribe - Nostr service not initialized (will retry when ready)
[2026-04-17T22:52:31.428255] [WARNING] [SYSTEM] Zendesk JWT: SDK not initialized
[2026-04-17T22:52:31.880285] [WARNING] [SYSTEM] Zendesk JWT: SDK not initialized
[2026-04-17T22:52:31.998264] [WARNING] [AUTH] (RelayDiscoveryService) ⚠️ No relay list found for npub1w2fd0sdyh6h8dmgm8xzzhyxsfnfufevuu934e2pg6sef8gclxfgqvfgmfu on any indexer
[2026-04-17T22:52:31.998276] [WARNING] [AUTH] (AuthService) ⚠️ No relay list found for user on any indexer — connecting to safe DM-friendly fallback relay set
[2026-04-17T22:52:32.154874] [WARNING] [VIDEO] (VideoEventService) Default video system is disabled - users will see real content only
[2026-04-17T22:52:36.856831] [WARNING] [VIDEO] (VideoEventService) Default video system is disabled - users will see real content only
[2026-04-17T22:53:06.861963] [ERROR] [VIDEO] (VideoEventService) ⏰ TIMEOUT: No events received for SubscriptionType.profile after 30 seconds!
[2026-04-17T22:53:06.862351] [WARNING] [VIDEO] (VideoEventService) ⏰ FEED TIMEOUT for profile:
=== FEED LOADING TIMEOUT DIAGNOSTIC ===
Subscription Type: profile
Timeout Duration: 30085ms
Relay Connected: true
Network Online: true
Connected Relay Count: 4

Filters:
  Filter 0:
    Kinds: [34236]
    Authors: 1 authors
    First author: 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
    Tags: 0 tags
    Since: null
    Until: null
    Limit: 50
  Filter 1:
    Kinds: [16]
    Au... [truncated]
[2026-04-17T22:53:57.952516] [DEBUG] [SYSTEM] (RelayNotificationsProvider) RelayNotifications: Resuming auto-refresh timer
[2026-04-17T22:53:57.953795] [INFO] (ProfileScreenRouter) 🧭 ProfileScreenRouter.build
[2026-04-17T22:53:57.955610] [INFO] [VIDEO] (VideoControllerCleanup) 🧹 Route type changed (RouteType.videoRecorder → RouteType.profile), disposing all video controllers
[2026-04-17T22:53:57.955613] [DEBUG] [VIDEO] (VideoControllerCleanup) 🧹 Invalidating all video controllers
[2026-04-17T22:53:57.968485] [DEBUG] [SYSTEM] (RelayNotificationsProvider) RelayNotifications: Pausing auto-refresh timer (no listeners)
[2026-04-17T22:53:57.969438] [INFO] (ProfileScreenRouter) 🧭 ProfileScreenRouter.build
[2026-04-17T22:53:57.974715] [DEBUG] [SYSTEM] (RelayNotificationsProvider) RelayNotifications: Resuming auto-refresh timer
[2026-04-17T22:53:57.992176] [DEBUG] [SYSTEM] (FollowRepository) Fetching follower stats for: 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.082961] [DEBUG] [SYSTEM] (FollowRepository) REST API follower stats: 0 followers, 0 following for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.186279] [DEBUG] [SYSTEM] (FollowRepository) Indexer wss://user.kindpag.es returned 0 followers for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.273840] [DEBUG] [SYSTEM] (FollowRepository) Indexer wss://purplepag.es returned 0 followers for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.390231] [DEBUG] [SYSTEM] (FollowRepository) Indexer wss://relay.nos.social returned 0 followers for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.390311] [INFO] [SYSTEM] (FollowRepository) Indexer followers counts: [0, 0, 0], using 0 for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:53:58.404512] [WARNING] [VIDEO] (VideoRecorderScreen) 📹 Failed to clear overlay visibility on dispose: Bad state: Using "ref" when a widget is about to or has been unmounted is unsafe.
Ref relies on BuildContext, and BuildContext is unsafe to use when the widget is deactivated.
To safely refer to the state of providers inside State.dispose(), save the provider state in a field of your State class.
[2026-04-17T22:53:58.404516] [DEBUG] [VIDEO] (VideoRecorderNotifier) 🧹 Destroying video recorder
[2026-04-17T22:53:58.404543] [INFO] [VIDEO] (VideoRecorderScreen) 📹 Disposed
[2026-04-17T22:53:58.405125] [INFO] [VIDEO] (CameraMobileService) 📷 Setting remote record control: disabled
[2026-04-17T22:53:58.409022] [DEBUG] [VIDEO] (CameraMobileService) 🎮 Remote record trigger callback cleared
[2026-04-17T22:53:58.409026] [DEBUG] [VIDEO] (VideoRecorderNotifier) 🎮 Remote record control disabled
[2026-04-17T22:53:58.409030] [INFO] [VIDEO] (CameraMobileService) 📷 Disposing mobile camera
[2026-04-17T22:54:00.995047] [DEBUG] [SYSTEM] (ContactListQuery) ⏰ Fallback timeout reached (3s)
[2026-04-17T22:54:00.995103] [DEBUG] [SYSTEM] (ContactListQuery) ✅ Immediate completion triggered: early=false, received=0, expected=null
[2026-04-17T22:54:05.011558] [INFO] [UI] (Navigation) User tapped settings gear
[2026-04-17T22:54:05.011718] [INFO] [UI] (OverlayVisibility) Page opened
[2026-04-17T22:54:05.012218] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/settings, authState=authenticated
[2026-04-17T22:54:05.012618] [INFO] [SYSTEM] (Route) CTX derive: type=RouteType.settings npub=null index=null
[2026-04-17T22:54:05.012706] [INFO] [VIDEO] (VideoControllerCleanup) 🧹 Route type changed (RouteType.profile → RouteType.settings), disposing all video controllers
[2026-04-17T22:54:05.012717] [DEBUG] [VIDEO] (VideoControllerCleanup) 🧹 Invalidating all video controllers
[2026-04-17T22:54:05.022911] [INFO] (ScreenAnalytics) 📱 Screen load started: settings
[2026-04-17T22:54:05.022916] [DEBUG] [UI] (PageLoadObserver) Page push tracked: settings
[2026-04-17T22:54:05.022928] [DEBUG] [VIDEO] (VideoControllerCleanup) 🛡️ Skipping duplicate invalidation (10ms since last)
[2026-04-17T22:54:05.022930] [INFO] [SYSTEM] (VideoStopNavigatorObserver) 📱 Navigation didPush to route: settings - stopped all videos
[2026-04-17T22:54:05.030468] [INFO] (ProfileScreenRouter) 🧭 ProfileScreenRouter.build
[2026-04-17T22:54:05.045303] [INFO] (ScreenAnalytics) ✅ Screen content visible: settings in 22ms
[2026-04-17T22:54:05.045407] [INFO] (PagePerf) PERF: settings — visible: 22ms
[2026-04-17T22:54:05.045987] [INFO] [AUTH] (AuthService) getKnownAccounts: raw=190 chars
[2026-04-17T22:54:05.473196] [DEBUG] [SYSTEM] (RelayNotificationsProvider) RelayNotifications: Pausing auto-refresh timer (no listeners)
[2026-04-17T22:54:05.476545] [INFO] (ProfileScreenRouter) 🧭 ProfileScreenRouter.build
[2026-04-17T22:54:06.167790] [DEBUG] [SYSTEM] (FollowRepository) Indexer follower pubkeys: 0 for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250
[2026-04-17T22:54:06.167823] [INFO] [SYSTEM] (FollowRepository) Followers for 7292d7c1a4beae76ed1b39842b90d04cd3c4e59ce1635ca828d43293a31f3250: API=0, relays=0, indexers=0, merged=0
[2026-04-17T22:54:10.602218] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/support-center, authState=authenticated
[2026-04-17T22:54:10.602646] [INFO] [SYSTEM] (Route) CTX derive: type=RouteType.supportCenter npub=null index=null
[2026-04-17T22:54:10.608410] [INFO] (ScreenAnalytics) 📱 Screen load started: support-center
[2026-04-17T22:54:10.608419] [DEBUG] [UI] (PageLoadObserver) Page push tracked: support-center
[2026-04-17T22:54:10.608438] [DEBUG] [VIDEO] (VideoControllerCleanup) 🧹 Invalidating all video controllers
[2026-04-17T22:54:10.608452] [INFO] [SYSTEM] (VideoStopNavigatorObserver) 📱 Navigation didPush to route: support-center - stopped all videos
[2026-04-17T22:54:10.649917] [INFO] (ScreenAnalytics) ✅ Screen content visible: support-center in 41ms
[2026-04-17T22:54:10.650067] [INFO] (PagePerf) PERF: support-center — visible: 41ms
[2026-04-17T22:54:14.865130] [DEBUG] [SYSTEM] (VideoStopNavigatorObserver) 📱 Skipping video disposal for popup route: null
[2026-04-17T22:55:18.545894] [INFO] [SYSTEM] Collecting bug report diagnostics
```

### Expected Result

My username should have updated.

(Generated from Zendesk Ticket: rabblelabs.zendesk.com/agent/tickets/1042)

realmeylisdev:
## Description

Fixes "Can't save username" (#3161, a duplicate of the closed #3152). On iOS the reporter's relay pool sat at zero connected relays, so every profile/username save failed (`ProfilePublishFailedException` ×6) and the divine.video username they had claimed never landed in their kind-0 metadata. The prior fixes around this (#3468 timeout, #3528 error copy, #3175 bootstrap relays, #5674 relay-confirmed saves) improved the *error handling* but none made the save actually **land** when the pool was stuck.

Root cause was threefold on current main:

- **G1 — one-shot, not offline-tolerant.** The save did a single best-effort reconnect per tap with no durable queue, and connectivity-return re-drove DMs/reactions/views but never the relay pool. A transient offline moment became a hard failure the user had to manually retry.
- **G2 — claim/publish consistency gap.** The divine.video username is claimed on the name server (idempotent, edge-synced) *before* the kind-0 publish. When the publish failed, the name was committed server-side but the kind-0 never advertised the `nip05` — a "half-registered" state that reads to the user as "didn't save."
- **G3 — false-negative.** A throw from the post-publish local cache write flipped a relay-confirmed publish into `publishFailed`.

### What this does

Makes the profile/username save **optimistic and durable**, reusing the app's existing retry-queue pattern (`OutgoingDmRetryService` / `PendingActionService`):

- **`PendingProfileSaves` Drift slot** — one row per user (kind-0 is replaceable, so latest-intent wins) + DAO, wired via the same runtime CREATE-IF-NOT-EXISTS pattern as `outgoing_dms` (schema-parity pinned by test). *(db_client)*
- **`ProfileRepository.enqueuePendingSave` / `drivePendingSave`** — an idempotent claim-then-publish re-drive that clears the slot only on a relay-confirmed `OK true`. Safe to replay: the name-server claim is idempotent for the owner and kind-0 is replaceable.
- **`ProfileSaveRetryService`** — sweeps the slot on app-foreground **and connectivity return**, with per-row backoff (5×, 2s→5min), a re-entrancy guard, cold-start interrupted-slot reset, and a `retryNow()` for a manual retry.
- **Optimistic bloc save** — after the username claim, enqueue and emit success immediately (the editor closes and shows the username), then re-drive the kind-0 in the background. A transient publish failure no longer surfaces as an error; the durable slot + retry service own delivery (fixes G1 + G2).
- **G3 fix** — clear the slot before the local cache write and make the cache write non-fatal, so a cache failure can never strand a relay-confirmed save.
- **connectivity → relay reconnect** — a debounced `forceReconnectAll` on network return so the pool self-heals app-wide (helps likes/views/DMs too), not only on app resume.

Commits are delineated per layer: db_client slot → repository enqueue/drive → retry service → G3 hardening → optimistic bloc → app wiring → connectivity reconnect → dead-code cleanup.

Recommend closing **#3161 as a duplicate of #3152** once this merges.

## Out of Scope

- **Permanent-failure UI surface (follow-up):** if the background retry permanently exhausts (5 retries over a prolonged offline window), the failed slot is still recoverable (retry service / connectivity / `retryNow`) but is not yet surfaced in the UI. Deferred to a follow-up.
- **Half-registered usernames (server/product follow-up):** a claim that lands server-side while its kind-0 never publishes is now re-driven client-side; a server/product path to surface or reap any names that stay unadvertised is a separate follow-up.
- The reporter's *exact* runtime trigger of the zero-connected state (network flip vs iOS socket-kill vs DNS/TLS) can't be pinned from the available device logs; the fix is trigger-agnostic.

## Verification

- **db_client**: full suite, including new `pending_profile_saves` DAO tests + schema-parity + upgrade-path tests.
- **profile_repository**: full suite, 100% package coverage retained (new payload model + enqueue/drive/G3).
- **ProfileSaveRetryService**: new unit test — sweep outcomes, backoff, re-entrancy, foreground/connectivity triggers, cold-start reset, manual retry.
- **ProfileEditorBloc**: 94 tests migrated to the optimistic behavior (assert the enqueued `PendingProfileSave` payload; transient publish failures → optimistic success + queued slot).
- profile_setup screen/widget tests, l10n `arb_consistency`, and the untested-services floor check — all pass.

## Type of Change

- [x] ✨ New feature (durable offline-tolerant save infrastructure)

### DIFF
--- a/mobile/lib/providers/social_providers.dart
+++ b/mobile/lib/providers/social_providers.dart
@@ -34,6 +34,7 @@ import 'package:openvine/services/hashtag_service.dart';
 import 'package:openvine/services/outgoing_dm_retry_service.dart';
 import 'package:openvine/services/pending_action_service.dart';
 import 'package:openvine/services/product_event_queue.dart';
+import 'package:openvine/services/profile_save_retry_service.dart';
 import 'package:openvine/services/social_service.dart';
 import 'package:openvine/services/user_data_cleanup_service.dart';
 import 'package:openvine/services/view_event_publisher.dart';
@@ -236,6 +237,67 @@ OutgoingDmRetryService? outgoingDmRetryService(Ref ref) {
   ref.onDispose(service.dispose);
   return service;
 }
+
+/// Auto-sweep service that re-drives the durable pending profile/username save
+/// slot (#3161) so a save attempted while relays were unreachable is delivered
+/// when connectivity returns. keepAlive with no UI consumer — read eagerly at
+/// app shell startup (`main.dart`) so the foreground + connectivity
+/// subscriptions are wired.
+///
+/// Returns null until the user is authenticated and the Nostr session is ready
+/// for the active identity (the same readiness [profileRepository] gates on),
+/// so drivePendingSave's claim + publish have a live signer and client.
+final profileSaveRetryServiceProvider = Provider<ProfileSaveRetryService?>((
+  ref,
+) {
+  final authService = ref.watch(authServiceProvider);
+
+  // Rebuild on sign-in / sign-out / account switch.
+  ref.watch(currentAuthStateProvider);
+
+  final userPubkey = authService.currentPublicKeyHex;
+  if (userPubkey == null || userPubkey.isEmpty) return null;
+
+  final readiness = ref.watch(nostrSessionProvider);
+  if (!readiness.isReadyForActiveClient || readiness.pubkey != userPubkey) {
+    return null;
+  }
+
+  final profileRepository = ref.watch(profileRepositoryProvider);
+  if (profileRepository == null) return null;
+
+  final db = ref.watch(databaseProvider);
+
+  final foregroundController = StreamController<bool>();
+  ref.onDispose(foregroundController.close);
+
+  final service = ProfileSaveRetryService(
+    profileRepository: profileRepository,
+    pendingProfileSavesDao: db.pendingProfileSavesDao,
+    userPubkey: userPubkey,
+    appForegroundStream: foregroundController.stream,
+    // Shared connectivity trigger: re-drive the moment the network returns,
+    // not only on an app-foreground transition.
+    retryTriggerStream: _dmRetryConnectivityTriggerStream(),
+  );
+
+  service.initialize().catchError((Object e) {
+    Log.error(
+      'Failed to initialize ProfileSaveRetryService',
+      name: 'AppProviders',
+      error: e,
+    );
+  });
+
+  ref.listen<bool>(appForegroundProvider, (_, next) {
+    if (!foregroundController.isClosed) {
+      foregroundController.add(next);
+    }
+  }, fireImmediately: true);
+
+  ref.onDispose(service.dispose);
+  return service;
+});
 
 /// Auto-sweep service that re-drives undelivered DM reactions (publish failed
 /// or interrupted mid-send) on app-foreground transitions via
--- a/mobile/lib/services/profile_save_retry_service.dart
+++ b/mobile/lib/services/profile_save_retry_service.dart
@@ -101,6 +101,15 @@ class ProfileSaveRetryService {
 
   StreamSubscription<bool>? _foregroundSubscription;
   StreamSubscription<void>? _retryTriggerSubscription;
+
+  /// Self-scheduled next attempt. A recovery signal (foreground/connectivity)
+  /// can arrive before the relay pool has reconnected, so relying on that one
+  /// signal would strand the save until the *next* signal. After every
+  /// non-terminal outcome the sweep arms this timer for the row's remaining
+  /// backoff, so one offline→online transition keeps retrying until a relay
+  /// confirms the publish (#3161 review).
+  Timer? _retryTimer;
+
   bool _isInitialized = false;
   bool _isSweeping = false;
 
@@ -136,12 +145,14 @@ class ProfileSaveRetryService {
     );
   }
 
-  /// Cancel the trigger subscriptions and mark the service un-init. Idempotent.
+  /// Cancel the trigger subscriptions and pending retry timer and mark the
+  /// service un-init. Idempotent.
   Future<void> dispose() async {
     await _foregroundSubscription?.cancel();
     _foregroundSubscription = null;
     await _retryTriggerSubscription?.cancel();
     _retryTriggerSubscription = null;
+    _cancelRetryTimer();
     _isInitialized = false;
   }
 
@@ -178,36 +189,62 @@ class ProfileSaveRetryService {
 
     try {
       final entry = await _dao.get(_userPubkey);
-      if (entry == null) return;
-
-      // A failed slot awaits a manual retry — don't auto-retry it.
-      if (entry.status == PendingProfileSaveStatus.failed) return;
-
-      // Per-row backoff.
-      final lastAttempt = entry.lastAttemptAt;
-      if (lastAttempt != null) {
-        final gap = _now().difference(lastAttempt);
-        if (gap < _retryConfig.backoffFor(entry.retryCount)) return;
+      if (entry == null) {
+        _cancelRetryTimer();
+        return;
       }
 
-      await _dao.markStatus(
+      // A failed slot awaits a manual retry — don't auto-retry it.
+      if (entry.status == PendingProfileSaveStatus.failed) {
+        _cancelRetryTimer();
+        return;
+      }
+
+      final generation = entry.generation;
+
+      // Per-row backoff. When a trigger arrives inside the backoff window, arm
+      // the timer for the remaining time rather than dropping the attempt.
+      final lastAttempt = entry.lastAttemptAt;
+      if (lastAttempt != null) {
+        final backoff = _retryConfig.backoffFor(entry.retryCount);
+        final gap = _now().difference(lastAttempt);
+        if (gap < backoff) {
+          _scheduleRetry(backoff - gap);
+          return;
+        }
+      }
+
+      // Claim the attempt for this generation. If a newer save replaced the
+      // row between the read above and here, this no-ops (false) and we bail —
+      // the newer row is pending and a fresh trigger/timer will drive it.
+      final claimed = await _dao.markStatus(
         userPubkey: _userPubkey,
         status: PendingProfileSaveStatus.syncing,
+        generation: generation,
+        attemptAt: _now(),
       );
+      if (!claimed) return;
 
-      final outcome = await _profileRepository.drivePendingSave(_userPubkey);
+      final outcome = await _profileRepository.drivePendingSave(
+        _userPubkey,
+        expectedGeneration: generation,
+      );
 
       switch (outcome) {
         case PendingSaveDriveOutcome.confirmed:
         case PendingSaveDriveOutcome.noPendingSave:
-          // Slot cleared by the repository (confirmed) or already gone.
-          break;
+          // Slot cleared by the repository (confirmed) or already gone /
+          // superseded — nothing more to schedule.
+          _cancelRetryTimer();
         case PendingSaveDriveOutcome.permanentFailure:
           await _dao.markStatus(
             userPubkey: _userPubkey,
             status: PendingProfileSaveStatus.failed,
             lastError: 'Username claim failed permanently (taken/reserved)',
+            generation: generation,
+            attemptAt: _now(),
           );
+          _cancelRetryTimer();
         case PendingSaveDriveOutcome.retryableFailure:
           if (entry.retryCount + 1 >= _retryConfig.maxRetries) {
             await _dao.markStatus(
@@ -215,13 +252,28 @@ class ProfileSaveRetryService {
               status: PendingProfileSaveStatus.failed,
               lastError:
                   'Could not publish after ${_retryConfig.maxRetries} attempts',
+              generation: generation,
+              attemptAt: _now(),
             );
+            _cancelRetryTimer();
           } else {
-            await _dao.incrementRetry(_userPubkey);
-            await _dao.markStatus(
+            await _dao.incrementRetry(
+              _userPubkey,
+              generation: generation,
+              attemptAt: _now(),
+            );
+            final reverted = await _dao.markStatus(
               userPubkey: _userPubkey,
               status: PendingProfileSaveStatus.pending,
+              generation: generation,
+              attemptAt: _now(),
             );
+            // Arm the next attempt so one recovery signal eventually publishes
+            // even if the relay pool wasn't ready this pass. Only when the row
+            // is still ours — a newer save that slipped in owns its scheduling.
+            if (reverted) {
+              _scheduleRetry(_retryConfig.backoffFor(entry.retryCount + 1));
+            }
           }
       }
     } on Object catch (e, stackTrace) {
@@ -242,5 +294,24 @@ class ProfileSaveRetryService {
     } finally {
       _isSweeping = false;
     }
+  }
+
+  /// Arm (replacing any pending) the self-scheduled next attempt for [delay].
+  /// A negative delay is clamped to zero. The timer fires a fresh [sweep];
+  /// because it is armed from inside a sweep (before the `finally` clears
+  /// [_isSweeping]) and never with a same-tick delay that could beat that
+  /// `finally`, the scheduled sweep always finds the re-entrancy guard clear.
+  void _scheduleRetry(Duration delay) {
+    _retryTimer?.cancel();
+    final clamped = delay.isNegative ? Duration.zero : delay;
+    _retryTimer = Timer(clamped, () {
+      _retryTimer = null;
+      unawaited(sweep());
+    });
+  }
+
+  void _cancelRetryTimer() {
+    _retryTimer?.cancel();
+    _retryTimer = null;
   }
 }
--- a/mobile/lib/services/profile_save_retry_service.dart
+++ b/mobile/lib/services/profile_save_retry_service.dart
@@ -1,0 +1,246 @@
+// ABOUTME: Auto-sweeps the durable pending-profile-save slot (#3161) and
+// ABOUTME: re-drives the queued kind-0 publish via
+// ABOUTME: ProfileRepository.drivePendingSave. Triggered by app-foreground
+// ABOUTME: transitions and connectivity/relay-reconnect; mirrors
+// ABOUTME: OutgoingDmRetryService.
+
+import 'dart:async';
+
+import 'package:db_client/db_client.dart';
+import 'package:meta/meta.dart';
+import 'package:openvine/services/crash_reporting_service.dart';
+import 'package:profile_repository/profile_repository.dart';
+import 'package:unified_logger/unified_logger.dart';
+
+/// Crashlytics `reason:` suffix for the sweep's top-level catch. Inlined as a
+/// single site per `error_handling.md` (a dedicated *ReportableSites class is
+/// only warranted once a feature has 2+ migrated sites).
+const _sweepTopLevelSite = 'ProfileSaveRetryService.sweepTopLevel';
+
+/// Backoff configuration for [ProfileSaveRetryService].
+///
+/// Defaults match `OutgoingDmRetryConfig` / `PendingActionRetryConfig`
+/// (5 retries, 2s → 5min, 2× backoff).
+class ProfileSaveRetryConfig {
+  const ProfileSaveRetryConfig({
+    this.maxRetries = 5,
+    this.initialDelay = const Duration(seconds: 2),
+    this.maxDelay = const Duration(minutes: 5),
+    this.backoffMultiplier = 2.0,
+  });
+
+  /// Maximum publish attempts before the slot is marked failed.
+  final int maxRetries;
+
+  /// Delay before the second attempt (the first retry).
+  final Duration initialDelay;
+
+  /// Upper bound on the backoff delay.
+  final Duration maxDelay;
+
+  /// Multiplier applied per prior attempt.
+  final double backoffMultiplier;
+
+  /// Minimum gap required before a row whose previous attempt count is
+  /// [retryCount] may be retried. Clamped at [maxDelay].
+  Duration backoffFor(int retryCount) {
+    if (retryCount <= 0) return Duration.zero;
+    var ms = initialDelay.inMilliseconds.toDouble();
+    for (var i = 0; i < retryCount; i++) {
+      ms *= backoffMultiplier;
+      if (ms >= maxDelay.inMilliseconds) return maxDelay;
+    }
+    return Duration(milliseconds: ms.round());
+  }
+}
+
+/// Re-drives the single durable pending profile/username save for one account
+/// until a relay confirms the kind-0 publish, then the slot clears itself.
+///
+/// **Triggers:** [appForegroundStream] transitions to `true`, and each event on
+/// [retryTriggerStream] (wired to connectivity/relay-reconnect). The provider
+/// seeds the foreground stream with the current state on subscription, so the
+/// cold-start sweep fires automatically.
+///
+/// **Re-entrancy:** a sweep already in progress short-circuits the next
+/// trigger; the deferred work is picked up on the next transition.
+///
+/// **Per-row backoff:** a slot whose `lastAttemptAt + backoff(retryCount)` is
+/// in the future is skipped this pass.
+///
+/// **Terminal state:** once [ProfileSaveRetryConfig.maxRetries] is reached, or
+/// the divine.video username claim fails permanently (taken/reserved), the slot
+/// is marked `failed` and left for the user to retry manually via [retryNow].
+class ProfileSaveRetryService {
+  ProfileSaveRetryService({
+    required ProfileRepository profileRepository,
+    required PendingProfileSavesDao pendingProfileSavesDao,
+    required String userPubkey,
+    required Stream<bool> appForegroundStream,
+    Stream<void>? retryTriggerStream,
+    ProfileSaveRetryConfig retryConfig = const ProfileSaveRetryConfig(),
+    DateTime Function() now = DateTime.now,
+    CrashReportingService? crashReporting,
+  }) : _profileRepository = profileRepository,
+       _dao = pendingProfileSavesDao,
+       _userPubkey = userPubkey,
+       _appForegroundStream = appForegroundStream,
+       _retryTriggerStream = retryTriggerStream,
+       _retryConfig = retryConfig,
+       _now = now,
+       _crashReporting = crashReporting ?? CrashReportingService.instance;
+
+  final ProfileRepository _profileRepository;
+  final PendingProfileSavesDao _dao;
+  final String _userPubkey;
+  final Stream<bool> _appForegroundStream;
+  final Stream<void>? _retryTriggerStream;
+  final ProfileSaveRetryConfig _retryConfig;
+  final DateTime Function() _now;
+  final CrashReportingService _crashReporting;
+
+  StreamSubscription<bool>? _foregroundSubscription;
+  StreamSubscription<void>? _retryTriggerSubscription;
+  bool _isInitialized = false;
+  bool _isSweeping = false;
+
+  bool get isInitialized => _isInitialized;
+
+  @visibleForTesting
+  bool get isSweeping => _isSweeping;
+
+  /// Subscribe to the trigger streams and run a cold-start sweep. Idempotent.
+  Future<void> initialize() async {
+    if (_isInitialized) return;
+    _isInitialized = true;
+
+    // A slot left `syncing` by an app-kill mid-publish is reset to `pending`
+    // so this session retries it.
+    await _profileRepository.resetInterruptedPendingSave(_userPubkey);
+
+    _foregroundSubscription = _appForegroundStream.listen((foreground) {
+      if (foreground) unawaited(sweep());
+    });
+    _retryTriggerSubscription = _retryTriggerStream?.listen((_) {
+      unawaited(sweep());
+    });
+
+    // The cold-start sweep is driven by the foreground stream, which the
+    // provider seeds with the current state on subscription (mirrors
+    // OutgoingDmRetryService) — no explicit sweep() call needed here.
+
+    Log.info(
+      'initialized for $_userPubkey',
+      name: 'ProfileSaveRetryService',
+      category: LogCategory.system,
+    );
+  }
+
+  /// Cancel the trigger subscriptions and mark the service un-init. Idempotent.
+  Future<void> dispose() async {
+    await _foregroundSubscription?.cancel();
+    _foregroundSubscription = null;
+    await _retryTriggerSubscription?.cancel();
+    _retryTriggerSubscription = null;
+    _isInitialized = false;
+  }
+
+  /// Reset a `failed` slot back to `pending` (fresh retry budget) and sweep.
+  /// Wired to the UI's manual "retry" affordance.
+  Future<void> retryNow() async {
+    final entry = await _dao.get(_userPubkey);
+    if (entry == null) return;
+    // Re-upsert resets retryCount → 0, status → pending, lastError → null.
+    await _dao.upsert(
+      PendingProfileSaveEntry(
+        userPubkey: entry.userPubkey,
+        payloadJson: entry.payloadJson,
+        claimConfirmed: entry.claimConfirmed,
+        queuedAt: _now(),
+      ),
+    );
+    await sweep();
+  }
+
+  /// One pass over the pending-save slot. Public so tests and the manual-retry
+  /// path can drive it directly.
+  @visibleForTesting
+  Future<void> sweep() async {
+    if (_isSweeping) {
+      Log.debug(
+        'sweep already in progress, skipping',
+        name: 'ProfileSaveRetryService',
+        category: LogCategory.system,
+      );
+      return;
+    }
+    _isSweeping = true;
+
+    try {
+      final entry = await _dao.get(_userPubkey);
+      if (entry == null) return;
+
+      // A failed slot awaits a manual retry — don't auto-retry it.
+      if (entry.status == PendingProfileSaveStatus.failed) return;
+
+      // Per-row backoff.
+      final lastAttempt = entry.lastAttemptAt;
+      if (lastAttempt != null) {
+        final gap = _now().difference(lastAttempt);
+        if (gap < _retryConfig.backoffFor(entry.retryCount)) return;
+      }
+
+      await _dao.markStatus(
+        userPubkey: _userPubkey,
+        status: PendingProfileSaveStatus.syncing,
+      );
+
+      final outcome = await _profileRepository.drivePendingSave(_userPubkey);
+
+      switch (outcome) {
+        case PendingSaveDriveOutcome.confirmed:
+        case PendingSaveDriveOutcome.noPendingSave:
+          // Slot cleared by the repository (confirmed) or already gone.
+          break;
+        case PendingSaveDriveOutcome.permanentFailure:
+          await _dao.markStatus(
+            userPubkey: _userPubkey,
+            status: PendingProfileSaveStatus.failed,
+            lastError: 'Username claim failed permanently (taken/reserved)',
+          );
+        case PendingSaveDriveOutcome.retryableFailure:
+          if (entry.retryCount + 1 >= _retryConfig.maxRetries) {
+            await _dao.markStatus(
+              userPubkey: _userPubkey,
+              status: PendingProfileSaveStatus.failed,
+              lastError:
+                  'Could not publish after ${_retryConfig.maxRetries} attempts',
+            );
+          } else {
+            await _dao.incrementRetry(_userPubkey);
+            await _dao.markStatus(
+              userPubkey: _userPubkey,
+              status: PendingProfileSaveStatus.pending,
+            );
+          }
+      }
+    } on Object catch (e, stackTrace) {
+      Log.error(
+        'sweep failed: $e',
+        name: 'ProfileSaveRetryService',
+        category: LogCategory.system,
+        error: e,
+        stackTrace: stackTrace,
+      );
+      unawaited(
+        _crashReporting.recordError(
+          e,
+          stackTrace,
+          reason: _sweepTopLevelSite,
+        ),
+      );
+    } finally {
+      _isSweeping = false;
+    }
+  }
+}
--- a/mobile/lib/services/profile_save_retry_service.dart
+++ b/mobile/lib/services/profile_save_retry_service.dart
@@ -57,13 +57,19 @@ class ProfileSaveRetryConfig {
 /// Re-drives the single durable pending profile/username save for one account
 /// until a relay confirms the kind-0 publish, then the slot clears itself.
 ///
-/// **Triggers:** [appForegroundStream] transitions to `true`, and each event on
-/// [retryTriggerStream] (wired to connectivity/relay-reconnect). The provider
-/// seeds the foreground stream with the current state on subscription, so the
-/// cold-start sweep fires automatically.
+/// **Triggers:** [appForegroundStream] transitions to `true`, each event on
+/// [retryTriggerStream] (wired to connectivity/relay-reconnect), and a fresh
+/// enqueue observed on the durable slot (a new generation) — so a save stranded
+/// by a failed editor fast-path is re-driven without waiting for the next
+/// foreground/connectivity event. The provider seeds the foreground stream with
+/// the current state on subscription, so the cold-start sweep fires
+/// automatically.
 ///
-/// **Re-entrancy:** a sweep already in progress short-circuits the next
-/// trigger; the deferred work is picked up on the next transition.
+/// **Re-entrancy:** a trigger that arrives while a pass is in flight is
+/// coalesced into a single follow-up pass, not dropped. A newer generation that
+/// replaced an older, in-flight one is therefore always driven — the older
+/// pass's generation-guarded writes no-op against the newer row, then the loop
+/// re-reads and drives the current generation (latest intent wins).
 ///
 /// **Per-row backoff:** a slot whose `lastAttemptAt + backoff(retryCount)` is
 /// in the future is skipped this pass.
@@ -102,6 +108,13 @@ class ProfileSaveRetryService {
   StreamSubscription<bool>? _foregroundSubscription;
   StreamSubscription<void>? _retryTriggerSubscription;
 
+  /// Reactive view of the durable slot. Makes the sweep **level-triggered**: a
+  /// fresh enqueue (new [PendingProfileSaveEntry.generation]) wakes an
+  /// otherwise-idle service, so a save stranded by a failed editor fast-path is
+  /// re-driven without waiting for the *next* foreground/connectivity event
+  /// (#3161 review).
+  StreamSubscription<PendingProfileSaveEntry?>? _pendingSaveSubscription;
+
   /// Self-scheduled next attempt. A recovery signal (foreground/connectivity)
   /// can arrive before the relay pool has reconnected, so relying on that one
   /// signal would strand the save until the *next* signal. After every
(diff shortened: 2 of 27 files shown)


realmeylisdev:
# Device test guide — PR #6063 "offline-tolerant username save" (#3161)

Manual QA for a real device (iOS + Android). Grounded in the actual build.

---

## 1. What changed (read first)

Before: saving a profile/username while the relay pool had **zero connected
relays** failed hard ("can't save username") and the divine.video username you
claimed never landed in your Nostr profile.

Now the save is **optimistic + durable**:
- After the username **claim** (still synchronous), the app enqueues a durable
  on-disk record, shows success **immediately**, and closes the editor.
- It publishes the kind-0 in the **background**, retrying automatically when the
  network/relays come back (on app-foreground and on connectivity return).
- A transient publish failure **no longer shows an error** — only the *claim*
  step (username taken/reserved/unreachable) still blocks the save.

**Two things must both be true for a divine.video username to actually work
(the "G2" consistency goal):**
1. **Claim landed** — the name is registered to your pubkey on the name server.
2. **Publish landed** — your kind-0 carries `nip05: "_@<name>.divine.video"`.

The whole point of this PR is that #2 now happens *eventually and on its own*.

---

## 2. Prerequisites & toolkit

- Two builds ideally: this PR branch on your test device, and any second
  Nostr/divine client (or a relay query tool) to independently confirm what
  landed on relays.
- Your **pubkey/npub**: in-app it's on your Profile (share/QR) — note both the
  hex and npub.
- A way to **block only relays** (for the real repro — see §3):
  - Best: a proxy/firewall (Charles, Proxyman, a router rule, or Wi-Fi with a
    firewall) that blocks **`wss://relay.divine.video`** (and optionally the
    indexers) while leaving HTTPS to `names.divine.video` reachable.
  - Simpler-but-partial: Airplane mode (see the caveat in §3).
- `curl` (or a browser) for the verification checks in §7.
- The default **write relay** is `wss://relay.divine.video`. Indexers (lookup
  only): `purplepag.es`, `user.kindpag.es`, `relay.nos.social`.
- **In-app observability:** open the **Relay diagnostic screen** (Settings →
  relay/network diagnostics) to watch the connected-relay count and force
  reconnects — useful to confirm "0 connected" during a test.

### Timing you should expect
- Background retry backoff: **5 attempts**, gap grows **2s → 4s → 8s → 16s → 32s
  …** capped at **5 min**. Repeatedly toggling the network will **not** force
  back-to-back retries — a too-soon sweep is skipped.
- Connectivity-return → relay force-reconnect is **debounced ~2s**, so expect a
  couple seconds between "network back" and "kind-0 lands".

---

## 3. ⚠️ How to correctly reproduce #3161 (important)

The reporter's failure was **"claim OK, relays down."** To reproduce that exact
path you must keep HTTPS up while relays are down:

- **✅ Correct repro (new username):** block `wss://relay.divine.video` at the
  proxy/firewall, leave `https://names.divine.video` reachable. Then claim a
  **brand-new** username. Expect: **optimistic success** + editor closes, then
  the kind-0 publishes later when you unblock the relay.

- **⚠️ Airplane mode caveat:** in **full airplane mode**, a **brand-new**
  username claim is an **HTTPS POST that also fails**, so you'll get an
  **immediate claim error** ("Failed to claim username…"), *not* optimistic
  success. Airplane mode only exercises the optimistic/durable path for an
  **already-owned username** or a **metadata-only edit** (display name / bio,
  no new claim). Use airplane mode for those; use relay-only blocking for the
  new-username repro.

---

## 4. Core scenarios

Tap **Profile → Edit Profile** (route `/edit-profile`, AppBar "Edit Profile",
green **Save** button; shows **"Saving…"** while in flight). Divine-username vs
external-NIP-05 mode lives on **Settings → Nostr → NIP-05 address** ("Save
NIP-05" button, toggle "Use your own NIP-05 address" → confirm dialog).

### A. Online happy path (primary regression) — must pass
1. Relays connected (check the diagnostic screen shows ≥1 connected).
2. Edit Profile → enter a **new** username; watch the live indicator go
   **"Checking availability…" → "Username available!"** (green check).
3. Tap **Save**.
   - **Expect:** white snackbar with green check **"Profile published
     successfully!"**, editor closes within a second or two.
4. Within seconds, run the §7 verification: name-server shows the name active
   for your pubkey **and** your kind-0 on `relay.divine.video` carries
   `nip05:"_@<name>.divine.video"`, and the in-app verified/NIP-05 badge is
   green.

### B. THE FIX — offline-tolerant new-username save (core of #3161)
1. Block `wss://relay.divine.video` (HTTPS up). Diagnostic screen should show
   **0 connected** relays.
2. Edit Profile → new username → availability check still works (HTTPS) →
   **Save**.
   - **Expect:** **"Profile published successfully!"** + editor closes. **No
     error, no infinite spinner.** (This is the behavior that was broken before.)
3. Immediately verify the **claim** landed (name-server check, §7.1) — it should
   already be **active for your pubkey**.
4. Confirm the kind-0 has **not** published yet (query `relay.divine.video`: no
   kind-0 with the new `nip05`, or an in-app "not verified" state). This is the
   transient "half-registered" window.
5. **Unblock** `wss://relay.divine.video`. Keep the app foregrounded.
   - **Expect:** within ~2–5s (reconnect debounce + sweep) the kind-0 publishes
     on its own. Re-run §7: both checks now green, verified badge turns green —
     **without reopening the editor**.

### C. Metadata-only optimistic save in airplane mode
1. Airplane mode ON.
2. Edit Profile → change **display name** (and/or bio) only, no new username →
   **Save**.
   - **Expect:** optimistic success + editor closes.
3. Airplane mode OFF (stay foregrounded).
   - **Expect:** kind-0 with the new display name publishes within a few
     seconds; confirm on a relay/second client.

### D. External NIP-05 (your own domain), offline
1. Settings → Nostr → NIP-05 → toggle **"Use your own NIP-05 address"** →
   **Continue** → enter `you@yourdomain.com`.
2. Block relays (or airplane) → **Save NIP-05**.
   - **Expect:** **"NIP-05 saved"** snackbar + screen pops (optimistic).
3. Restore network → confirm the kind-0's `nip05` is your external address
   (external NIP-05 takes precedence over the divine username).

---

## 5. Durability & multi-account

### E. App-kill durability (no editor re-entry)
1. Do an **offline** save (scenario B or C), so the editor closes optimistically.
2. **Force-quit** the app while still offline.
3. Reopen the app, then **restore network** — **do not reopen the editor**.
   - **Expect:** the queued save re-drives on its own (cold-start reset +
     foreground sweep) and the kind-0 publishes. Verify on a relay.

### F. Connectivity-return re-drive (no app switch)
1. Offline save, then **leave the app foregrounded** and toggle Airplane/Wi-Fi
   **off → on**.
   - **Expect:** the save re-drives on the connectivity event alone (no
     background/foreground round-trip needed).

### G. Account isolation
1. On account **A**, do an offline save (queued).
2. Switch to account **B**; do a *different* offline save on B.
3. Restore network; switch between A and B.
   - **Expect:** each account's kind-0 publishes **its own** content; neither
     leaks into the other; A's queued save is intact and publishes only when A is
     the active/ready identity. Nothing publishes under a stale signer right
     after a switch.

### H. Latest-edit-wins
1. Offline: Save display name **"A"**. Reopen editor, change to **"B"**, Save
   again. Restore network.
   - **Expect:** the finally-published kind-0 is **"B"**; only one kind-0 per
     replace (no duplicate/competing events).

---

## 6. Regressions & known limitations to observe

### I. Claim errors still block (must still error)
- New username that is **taken** → immediate red snackbar **"Username was just
  taken. Please choose another."**, editor **stays open**, nothing queued.
- **Reserved** name → inline "Username is reserved" with "Contact support" /
  "Check again".
- **Full airplane + brand-new username** → immediate **"Failed to claim
  username. Please try again."** (claim is HTTPS; see §3).

### J. Avatar/banner staged-upload gating
- Pick a **new avatar/banner**, then tap **Save before the upload finishes**.
  - **Expect:** nothing happens — no toast, no close, no error (silently
    dropped). You must wait until the image is staged, then Save. Flag the lack
    of feedback if it's confusing.

### K. Other publishes survive the new app-wide reconnect (regression)
- With a video post, a like, a comment, and a DM send in flight, toggle
  connectivity off→on (fires the debounced app-wide `forceReconnectAll`).
  - **Expect:** none of those are dropped; DM/feed subscriptions resubscribe
    without missing events; no feed glitch, playback stutter, or obvious jank.
- Rapidly background↔foreground + toggle network repeatedly; watch for
  abnormal battery drain from repeated relay-pool rebuilds.

### ⚠️ L. Known limitations — observe and report (deferred follow-ups)
- **Silent permanent failure:** if the write relay stays unreachable across all
  5 retries (wait out the growing backoff up to ~5 min), the save is marked
  `failed` and stops retrying — with **no error, badge, or retry button** in this
  build (the failure surface is a deferred follow-up). Confirm the user gets no
  recovery UI, and that force-quit + relaunch does **not** auto-republish a
  `failed` slot. **This is expected/known** — verify it matches the deferred
  scope, don't file as a new bug.
- **Optimistic-display honesty:** right after an **offline** save shows "Profile
  published successfully!", check your **own** profile/feed name + avatar. The
  local display cache is only updated on a **confirmed** publish, so the UI may
  still show the **old** values while the toast says "published." Note whether
  the user is misled during the offline window.
- **Cold-cache field preservation (data-loss risk):** on a **fresh reinstall**
  (empty local cache), with an existing rich profile (lightning `lud16`,
  website, bio, NIP-39 tags), edit **only** the display name offline. After the
  background publish, verify the kind-0 **still contains all prior fields** and
  didn't drop `lud16`/website/about.

---

## 7. Verification toolkit (does it *actually* work?)

### 7.1 Claim landed (name server)
```
curl -s "https://names.divine.video/api/username/check/<name>"
# expect available:false + status "active" + "pubkey":"<your hex>"

curl -s "https://names.divine.video/api/username/by-pubkey/<your-hex-pubkey>"
# expect found:true with your name
```
> Fastly edge sync is async — if the check lags right after a claim, re-poll.

### 7.2 NIP-05 advertisement resolves
```
curl -s "https://<name>.divine.video/.well-known/nostr.json?name=_"
# expect {"names":{"_":"<your hex pubkey>"}} (empty {} / 404 = not registered)
```

### 7.3 Kind-0 published on the write relay (the part this PR fixes)
Query `wss://relay.divine.video` for `{"kinds":[0],"authors":["<hex>"]}` (via a
Nostr CLI/`nak`, a relay explorer, or a second client) and confirm the newest
kind-0 `content` includes `"nip05":"_@<name>.divine.video"` (or your external
NIP-05). **Green NIP-05 verification requires 7.1/7.2 AND 7.3 to agree.**

---

## 8. Sign-off checklist
- [ ] A. Online new-username save completes + verifies (7.1–7.3) within seconds
- [ ] B. Offline new-username (relay blocked): optimistic success, then
      auto-publishes on reconnect (both G2 checks green)
- [ ] C. Airplane metadata-only save: optimistic, publishes on reconnect
- [ ] D. External NIP-05 offline save: optimistic + publishes
- [ ] E. App-kill durability: publishes without reopening the editor
- [ ] F. Connectivity-return re-drive (foregrounded, no app switch)
- [ ] G. Two-account isolation
- [ ] H. Latest-edit-wins (one kind-0)
- [ ] I. Taken/reserved/offline-new-username still error immediately
- [ ] J. Save dropped while image upload in flight
- [ ] K. Likes/DMs/uploads survive connectivity toggle; no jank/battery issue
- [ ] L. Known limitations behave as documented (silent permanent-fail;
      optimistic display; cold-cache fields) — reported, not surprising
- [ ] iOS **and** Android (incl. iOS Low-Power / Android Doze background resume)

mobile/packages/profile_repository/lib/src/profile_repository.dart
```diff
@@ -706,6 +714,131 @@ class ProfileRepository {
...
+        picture: payload.picture,
+        banner: payload.banner,
+        monetizationLinks: payload.monetizationLinks,
+        currentProfile: currentProfile,
+      );
+      // A relay confirmed the kind-0 (saveProfileEvent only returns on an
+      // OK true). Clear the slot BEFORE the local cache write so a
+      // cache-write failure can't strand a save that already persisted on a
+      // relay — the #3161 G3 false-negative, relocated here from the bloc.
+      await dao.clear(pubkey);
```
dcadenas: Guard every pending-save mutation with an immutable generation or intent ID, not just the pubkey. At this line, drive A can clear replacement B after B replaces the row while A awaits relay work, and the same pubkey-only claim, failure, and retry updates can reclassify B. Add an interleaving test where B is enqueued while A is awaiting publish, then verify B remains queued and is eventually driven.

mobile/packages/profile_repository/lib/src/profile_repository.dart
```diff
@@ -706,6 +714,131 @@ class ProfileRepository {
...
+          return PendingSaveDriveOutcome.permanentFailure;
+        case UsernameClaimNetworkError():
+        case UsernameClaimError():
+          return PendingSaveDriveOutcome.retryableFailure;
+      }
+    }
+
+    final currentProfile = await getCachedProfile(pubkey: pubkey);
+    try {
+      final saved = await saveProfileEvent(
```
dcadenas: Make relay confirmation clear the matching pending row before any cache write can abort the drive. saveProfileEvent writes the profile cache at lines 689-693 before returning, so a failure of that first upsert prevents execution from reaching the clear at line 823. Make this cache write non-fatal for the confirmed-publish path or separate confirmation and cache persistence, then test that the first upsert failure still leaves the outcome confirmed and the pending row cleared.

mobile/lib/providers/social_providers.dart
```diff
@@ -237,6 +238,67 @@ OutgoingDmRetryService? outgoingDmRetryService(Ref ref) {
...
+  ref.onDispose(foregroundController.close);
+
+  final service = ProfileSaveRetryService(
+    profileRepository: profileRepository,
+    pendingProfileSavesDao: db.pendingProfileSavesDao,
+    userPubkey: userPubkey,
+    appForegroundStream: foregroundController.stream,
+    // Shared connectivity trigger: re-drive the moment the network returns,
+    // not only on an app-foreground transition.
+    retryTriggerStream: _dmRetryConnectivityTriggerStream(),
```
dcadenas, line 281: Tie the retry sweep to relay readiness and schedule the next eligible attempt instead of waiting only for another lifecycle or connectivity event. A recovery signal can be consumed before relay reconnect completes, and the backoff path currently returns without arming a timer. Ensure one offline-to-online transition eventually publishes a queued save after relay recovery, and add a test that covers this without a second signal.

dcadenas (dismissed):
The pending profile-save flow can lose a newer edit, retain an already-confirmed event, and stop retrying after a single recovery signal. Address these three blocking issues before merging.

mobile/packages/profile_repository/lib/src/profile_repository.dart
realmeylisdev: Fixed in 39d7abf7d.

Added an immutable per-enqueue `generation` token (new column on `pending_profile_saves`, minted on every `upsert` and returned to the caller). `drivePendingSave` captures it at read time and guards `markClaimConfirmed` + `clear` with it; the retry service guards its `syncing` / `incrementRetry` / `markStatus` bookkeeping the same way. So a newer save that replaces the row while an older drive is awaiting relay work is never cleared or reclassified — the stale drive's writes match zero rows, and the newer intent survives to be driven on its own (latest intent wins). `drivePendingSave` also takes an `expectedGeneration` so it bails out (`noPendingSave`) when the row was already superseded before it started.

Added the interleaving test you asked for in `profile_repository_pending_save_test.dart` — B is enqueued while A is parked on a gated publish, then A confirms; B is not cleared and is driven to completion on its own — plus DAO-level coverage that each guarded mutation no-ops under a stale generation.

mobile/packages/profile_repository/lib/src/profile_repository.dart
realmeylisdev: Fixed in 39d7abf7d.

The post-publish cache write in `saveProfileEvent`'s `PublishSuccess` branch is now non-fatal: a relay already confirmed the kind-0, so an `upsertProfile` throw is logged and swallowed instead of propagating out and skipping the `clear`. I also removed the now-redundant second `cacheProfile` in `drivePendingSave`, so there's a single cache site.

The G3 test now makes the first (and only) profile-cache upsert throw and asserts the outcome is still `confirmed` and the pending row is cleared.

mobile/lib/providers/social_providers.dart
realmeylisdev: Fixed in 39d7abf7d.

`ProfileSaveRetryService` now self-schedules the next attempt: a `retryableFailure` arms `Timer(backoffFor(retryCount + 1))`, and a trigger that lands inside the backoff window arms the remaining time rather than dropping the attempt (the timer is cancelled on confirm / terminal failure / dispose). So one offline→online transition keeps retrying on backoff until a relay actually confirms — the sweep is tied to relay readiness by construction (it keeps attempting until the publish lands), on top of the existing connectivity trigger and `connectivityRelayReconnectProvider`. I opted not to add a separate relay-status stream, since that would just be a second signal for the same "keep trying until it lands" behavior the timer already gives.

Added a service test that fires one recovery signal, fails the first attempt (relays not ready yet), and asserts the armed timer drives the retry that confirms — with no second signal.

realmeylisdev:
@dcadenas all three blockers are addressed in 39d7abf7d and replied on their threads:

- **Lost newer edit** — every pending-save mutation is now guarded by an immutable per-enqueue `generation` token, so an in-flight drive can't clear or reclassify a save that replaced it mid-flight (interleaving test added).
- **Retained confirmed event** — the post-publish cache write is non-fatal on the confirmed path, so a cache throw can't skip the slot clear (G3 test now fails the first/only upsert).
- **Stops retrying after one signal** — the retry service self-schedules the next attempt on backoff, so one offline→online transition keeps retrying until a relay confirms (test drives one signal, no second).

Suites are green: db_client, profile_repository (100% retained), ProfileEditorBloc, and ProfileSaveRetryService. Ready for another look.

mobile/lib/services/profile_save_retry_service.dart
```diff
@@ -0,0 +1,317 @@
...
+      // Claim the attempt for this generation. If a newer save replaced the
+      // row between the read above and here, this no-ops (false) and we bail —
+      // the newer row is pending and a fresh trigger/timer will drive it.
+      final claimed = await _dao.markStatus(
+        userPubkey: _userPubkey,
+        status: PendingProfileSaveStatus.syncing,
+        generation: generation,
+        attemptAt: _now(),
+      );
+      if (!claimed) return;
```
dcadenas: Make retry scheduling level-triggered for the current durable slot here. When this claim loses ownership, or when an older drive completes with `confirmed` or `noPendingSave`, re-read the slot and drive or schedule the current generation; also ensure enqueue and a retryable editor fast-path outcome wake the service, while coalescing triggers during an active sweep. Generation B can currently replace in-flight A, have its fast-path failure ignored at `mobile/lib/blocs/profile_editor/profile_editor_bloc.dart:1061-1065`, and remain pending after A returns without any wake-up. Add a service-level A/B interleaving test that does not manually drive B or emit another foreground/connectivity signal.

mobile/lib/services/profile_save_retry_service.dart
```diff
@@ -0,0 +1,317 @@
...
+            final reverted = await _dao.markStatus(
+              userPubkey: _userPubkey,
+              status: PendingProfileSaveStatus.pending,
+              generation: generation,
+              attemptAt: _now(),
+            );
+            // Arm the next attempt so one recovery signal eventually publishes
+            // even if the relay pool wasn't ready this pass. Only when the row
+            // is still ours — a newer save that slipped in owns its scheduling.
+            if (reverted) {
```
dcadenas: Guard retry scheduling and timer callbacks against disposal here. Set a disposed or lifecycle-generation marker before awaiting cancellation, check it after asynchronous boundaries and before `_scheduleRetry` or a follow-up sweep, and make timer callbacks no-op when ownership is stale. A sweep can resume after provider disposal, arm a new timer, and continue mutating the old row's retry budget because `dispose` only cancels the timer that exists at that moment. Add a gated test that disposes during a retryable drive and verifies there is no second repository call, retry-count mutation, or timer-driven retry.

dcadenas (changes requested):
Make the retry service guarantee that a newer pending generation is driven after an older attempt loses ownership. Also prevent an in-flight sweep from scheduling or running work after the account-scoped service is disposed; both are required before merge.

mobile/lib/services/profile_save_retry_service.dart
realmeylisdev: Fixed in 1ac878720.

The re-drive is now level-triggered. `sweep()` runs a drain loop over a coalesced `_sweepAgain` flag, and after every drive (and on a lost `markStatus(syncing)` claim) `_chainIfNewerGeneration` re-reads the slot and re-sweeps when a *different* generation is now queued. So when B replaces in-flight A, A's generation-guarded writes no-op against B and the loop drives B on the next pass (latest intent wins). It stays loop-free by only chaining on a genuinely newer generation — a cleared/empty/same-generation slot ends the pass.

For the idle-service path the service now watches the durable slot and arms a short backstop sweep on a fresh enqueue (generation-deduped and primed, so its own status/retry writes and the initial emission don't re-wake it). That covers both "enqueue wakes the service" and "a retryable editor fast-path outcome wakes the service": the enqueue that precedes the fast-path already wakes the service, so a save the fast-path fails to land is re-driven without a second foreground/connectivity signal. The backstop delay lets the fast-path win the common online case, so there's no redundant publish. I kept the editor fast-path as you framed it — the service is the durable backstop, not a replacement.

Added the service-level A/B interleaving test you asked for: it enqueues B while A's publish is parked, never manually drives B or emits a second signal, and asserts B is driven to a confirmed clear. Plus an enqueue-wakes-an-idle-service test. Neutering the chain makes the interleaving test fail (B is never driven), so it bites.

mobile/lib/services/profile_save_retry_service.dart
realmeylisdev: Fixed in 1ac878720.

`dispose()` now sets a `_disposed` marker first, before awaiting cancellation. `_runSweepOnce` checks it after every async boundary — after the slot read, after the claim, after `drivePendingSave`, and after each failure-bookkeeping write — and returns before scheduling, so a sweep that resumes past disposal can't `incrementRetry`, re-mark the row, or arm a timer. `_scheduleRetry` and the timer callback no-op when disposed, the slot watcher no-ops when disposed, and the watch subscription is cancelled in `dispose()`.

Added a gated test that disposes during a retryable drive and asserts no second repository call, no retry-count mutation, and no timer-driven retry (real clock + tiny backoff so a wrongly-armed timer would fire within the test window). Removing the post-drive guard makes it fail (retryCount becomes 1), so it bites.

realmeylisdev:
@dcadenas both blockers from this review are addressed in 1ac878720 (with a reply on each thread):

- **Level-triggered re-drive** — `sweep()` drains a coalesced follow-up flag and `_chainIfNewerGeneration` re-reads and re-drives the current generation after a lost claim or a completed older drive, so a newer save that replaces an in-flight older one is always driven (latest intent wins). The service also watches the durable slot and arms a short backstop sweep on a fresh enqueue, so a save stranded by a failed editor fast-path is re-driven with no second foreground/connectivity signal.
- **Disposal guard** — `_disposed` is set before awaiting cancellation and checked after every async boundary and before scheduling; the timer callback and slot watcher no-op after teardown, and the watch subscription is cancelled.

New service tests: the A/B interleaving case (B driven with no manual drive and no second signal), enqueue-wakes-an-idle-service, and dispose-during-retryable-drive. Both new guards were mutation-verified to fail without the fix. `flutter analyze lib test integration_test` is clean and the profile service/repository/DAO plus editor-bloc suites are green.

dcadenas (approved):
📡 durable offline profile saves are wired cleanly.

PR #6064, opened 2026-07-13
fix(camera): stop zoom sticking at random positions on iOS

### LINKED ISSUE #6065: fix(camera): stop zoom sticking at random positions on iOS
Auto-created from PR #6064.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6064
**Branch:** `fix/camera-zoom-stuck-ios`

hm21:
Closes #6065

## Description

A user reported that on iOS the recorder camera zoom freezes at a seemingly random factor (5.2×, 9.5×, …) that is stable within a session and differs across sessions — zooming out always works, and the ruler read-out keeps climbing to ~30× while the picture stays put. Their logs (iPhone17,3, 1.0.16+804) confirmed two independent root causes:

**1. iOS silently clamps the zoom, and we never noticed.** On virtual multi-camera devices iOS restricts `min/maxAvailableVideoZoomFactor` at runtime (e.g. under video stabilization, which the reporter had on) and *silently clamps* an out-of-range `videoZoomFactor` assignment instead of throwing. Our native setter clamped against a configure-time snapshot (61.875× on their device), never read the factor back, and the Dart layer discarded the success flag — so the app kept emitting the *requested* zoom while the sensor sat at the clamp. Ruler and camera drifted apart.

- Clamp against the device's **live** `min/maxAvailableVideoZoomFactor`, **read `videoZoomFactor` back** after assignment, and return the applied zoom + live range over the platform channel (new `CameraZoomState`).
- `DivineCamera` / `CameraService.setZoomLevel` now return the applied value (`double?`) and mirror the refreshed bounds into bloc state. The ruler stops exactly where the camera does, and its scale shrinks under a restriction and **heals** once it lifts.
- `CameraMobileService` logs a `Zoom clamped by platform: requested X, applied Y` breadcrumb so we can confirm the trigger (stabilization) on device.

**2. The zoom ruler was swallowing the pinch.** The interactive ruler turned pointer-hungry the instant a scale gesture began — its opaque full-width band sits right where the lower pinch finger lands, and it activated even before fading in. A second finger there degraded the pinch into a one-finger no-op *plus* an unintended linear ruler scrub (110px/1.0×, integer detents) — which is exactly the slow crawl visible in the logs (620 sub-0.01 steps, asymptotic integer snapping, never a pinch step).

- Gate ruler interactivity on a new `isPinchActive` flag (set for the whole preview gesture, wired via `onScaleEnd`), so a pinch finger on the ruler falls through to the preview.
- Only summon the ruler on a genuine two-pointer pinch, so a one-finger pan no longer arms the invisible strip.

## Out of Scope

- The linear 110px-per-1.0× ruler scale means the full 0.5→61.9× span needs ~6,700px of drag travel. A logarithmic mapping or a sensible zoom cap is a separate product/UX decision, not included here.
- iOS Swift changes are not built locally (no device build in this environment) — please verify on device.

## Verification

- `flutter analyze lib test integration_test` — clean; `dart format` — no changes
- `divine_camera`: **366 tests green, 100% line coverage** (CI gate), incl. new tests for silent-clamp adoption, bounds healing, `CameraZoomState`, `onScaleEnd`
- App: video-recorder bloc + widget suites green (~380 tests), incl. new regressions — "emits the zoom the camera actually applied when the OS silently clamps it", "does not capture pointers while a preview pinch is in progress", "does not summon the zoom ruler on a one-pointer pan"
- Android Kotlin compiles (`:divine_camera:compileDebugKotlin`, exit 0)
- Pre-push hook (analyze + changed-file tests) green

**Related Issue:**  (reported via Zendesk/logs, no tracked issue)

## Type of Change

### DIFF
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -1309,6 +1309,12 @@ class VideoRecorderBloc
         snappedTo1x: snapped,
         snapTime: snapTime,
         showZoomIndicator: true,
+        // Re-assert for the whole gesture. A two-finger pinch held stationary
+        // for >1s emits no update, so the auto-hide timer fires while the
+        // pinch is still down and clears isPinchActive. A resumed update must
+        // restore it — otherwise the ruler shows interactive again and
+        // swallows the lower pinch finger.
+        isPinchActive: true,
       ),
     );
     _armZoomIndicatorHideTimer();
@@ -1337,8 +1343,11 @@ class VideoRecorderBloc
     // isPinchActive is cleared as a safety net for the timer-armed paths
     // (a real two-pointer pinch, or a programmatic zoom set): the timer only
     // fires after a full second without gesture activity, so a still-set flag
-    // there means the scale-end callback was lost (e.g. gesture cancelled by
-    // navigation) and would otherwise leave the ruler non-interactive. A pure
+    // there usually means the scale-end callback was lost (e.g. gesture
+    // cancelled by navigation) and would otherwise leave the ruler
+    // non-interactive. Clearing it is safe even when the pinch is in fact
+    // still down (held stationary >1s): _onScaleUpdated re-asserts the flag on
+    // the next update, so the ruler can't turn interactive mid-gesture. A pure
     // one-finger pan arms no timer, so if its scale-end is lost the flag
     // instead self-heals on the next pinch or zoom set.
     emit(state.copyWith(showZoomIndicator: false, isPinchActive: false));
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -1334,10 +1334,13 @@ class VideoRecorderBloc
     Emitter<VideoRecorderBlocState> emit,
   ) {
     _zoomIndicatorTimer = null;
-    // isPinchActive is cleared as a safety net: the timer only fires after
-    // a full second without gesture activity, so a still-set flag means the
-    // scale-end callback was lost (e.g. gesture cancelled by navigation) and
-    // would otherwise leave the ruler permanently non-interactive.
+    // isPinchActive is cleared as a safety net for the timer-armed paths
+    // (a real two-pointer pinch, or a programmatic zoom set): the timer only
+    // fires after a full second without gesture activity, so a still-set flag
+    // there means the scale-end callback was lost (e.g. gesture cancelled by
+    // navigation) and would otherwise leave the ruler non-interactive. A pure
+    // one-finger pan arms no timer, so if its scale-end is lost the flag
+    // instead self-heals on the next pinch or zoom set.
     emit(state.copyWith(showZoomIndicator: false, isPinchActive: false));
   }
 
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -215,6 +215,7 @@ class VideoRecorderBloc
     on<VideoRecorderZoomedByLongPress>(_onZoomedByLongPress);
     on<VideoRecorderScaleStarted>(_onScaleStarted);
     on<VideoRecorderScaleUpdated>(_onScaleUpdated);
+    on<VideoRecorderScaleEnded>(_onScaleEnded);
     on<VideoRecorderRecordingLockedForNavigation>(
       _onRecordingLockedForNavigation,
     );
@@ -549,8 +550,8 @@ class VideoRecorderBloc
       return;
     }
 
-    final success = await _cameraService.setZoomLevel(event.value);
-    if (!success) {
+    final appliedZoom = await _cameraService.setZoomLevel(event.value);
+    if (appliedZoom == null) {
       Log.warning(
         '⚠️ Failed to set zoom level to ${event.value}',
         name: 'VideoRecorderBloc',
@@ -558,7 +559,19 @@ class VideoRecorderBloc
       );
       return;
     }
-    emit(state.copyWith(zoomLevel: event.value, showZoomIndicator: true));
+    // Emit what the camera actually applied, never the request: iOS can
+    // silently clamp an in-range request when the available zoom range is
+    // restricted at runtime. Mirroring the refreshed bounds keeps the
+    // ruler's scale honest (it shrinks under a restriction and heals once
+    // the restriction lifts).
+    emit(
+      state.copyWith(
+        zoomLevel: appliedZoom,
+        minZoomLevel: _cameraService.minZoomLevel,
+        maxZoomLevel: _cameraService.maxZoomLevel,
+        showZoomIndicator: true,
+      ),
+    );
     _armZoomIndicatorHideTimer();
   }
 
@@ -1194,25 +1207,49 @@ class VideoRecorderBloc
     VideoRecorderScaleStarted event,
     Emitter<VideoRecorderBlocState> emit,
   ) {
+    // Only a real pinch (two pointers) summons the ruler. A one-pointer
+    // scale start is just a pan and must not flip the ruler interactive —
+    // an invisible, fading-in ruler otherwise swallows the second finger
+    // of the pinch that follows (its opaque hit-test band sits right where
+    // the lower finger lands). isPinchActive additionally blocks the ruler
+    // for the whole gesture, so a second finger landing on the (visible)
+    // ruler still falls through to the preview and completes the pinch.
+    //
     // snapTime is intentionally not reset here: copyWith's null-coalescing
     // semantics make explicit-null a no-op anyway, and the snap-update
     // handler guards every snapTime read with snappedTo1x — which we
     // just cleared. The next snap engagement overwrites snapTime fresh.
+    final isPinch = event.details.pointerCount >= 2;
     emit(
       state.copyWith(
         baseZoomLevel: state.zoomLevel,
         snappedTo1x: false,
         lastRawZoom: state.zoomLevel,
-        showZoomIndicator: true,
+        showZoomIndicator: isPinch ? true : null,
+        isPinchActive: true,
       ),
     );
-    _armZoomIndicatorHideTimer();
+    if (isPinch) _armZoomIndicatorHideTimer();
+  }
+
+  void _onScaleEnded(
+    VideoRecorderScaleEnded event,
+    Emitter<VideoRecorderBlocState> emit,
+  ) {
+    emit(state.copyWith(isPinchActive: false));
   }
 
   Future<void> _onScaleUpdated(
     VideoRecorderScaleUpdated event,
     Emitter<VideoRecorderBlocState> emit,
   ) async {
+    // One-pointer updates with an unchanged scale are pans, not pinches
+    // (a single pointer keeps scale at exactly 1.0; trackpad pinches
+    // report a scale != 1.0). Skipping them keeps plain preview pans from
+    // summoning and re-arming the zoom ruler.
+    if (event.details.pointerCount < 2 && event.details.scale == 1.0) {
+      return;
+    }
     // Multiplicative mapping: the cumulative pinch scale multiplies the zoom
     // captured at gesture start, so the perceived sensitivity is uniform
     // across the whole range (1×→2× feels like 2×→4×). The previous
@@ -1297,7 +1334,11 @@ class VideoRecorderBloc
     Emitter<VideoRecorderBlocState> emit,
   ) {
     _zoomIndicatorTimer = null;
-    emit(state.copyWith(showZoomIndicator: false));
+    // isPinchActive is cleared as a safety net: the timer only fires after
+    // a full second without gesture activity, so a still-set flag means the
+    // scale-end callback was lost (e.g. gesture cancelled by navigation) and
+    // would otherwise leave the ruler permanently non-interactive.
+    emit(state.copyWith(showZoomIndicator: false, isPinchActive: false));
   }
 
   void _onRecordingLockedForNavigation(
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -1319,7 +1319,16 @@ class VideoRecorderBloc
     );
     _armZoomIndicatorHideTimer();
 
-    if ((state.zoomLevel - clampedZoom).abs() > 0.01) {
+    // Dispatch when the applied zoom actually moved, or when the raw pinch
+    // target pushed past a cached bound even though the clamp pinned it to the
+    // current zoom. The latter re-runs the platform set so its read-back can
+    // widen a stale ceiling a lifted runtime restriction left behind —
+    // otherwise a user pinned at a shrunk max stays capped until they first
+    // pinch back in.
+    final pushedPastBound =
+        newZoom > _cameraService.maxZoomLevel ||
+        newZoom < _cameraService.minZoomLevel;
+    if ((state.zoomLevel - clampedZoom).abs() > 0.01 || pushedPastBound) {
       add(VideoRecorderZoomLevelSet(clampedZoom));
     }
   }
--- a/mobile/packages/divine_camera/ios/Classes/CameraController.swift
+++ b/mobile/packages/divine_camera/ios/Classes/CameraController.swift
@@ -1737,22 +1737,51 @@ class CameraController: NSObject {
     
     /// Sets the zoom level (user-facing value, e.g. 0.5x, 1.0x, 2.0x).
     /// Internally converts to native videoZoomFactor using nativeToUserZoomScale.
-    func setZoomLevel(level: CGFloat) -> Bool {
-        guard let device = videoDevice else { return false }
-        
-        let clampedLevel = max(minZoom, min(level, maxZoom))
+    ///
+    /// Clamps against the device's *current* min/maxAvailableVideoZoomFactor
+    /// instead of the configure-time snapshot: on virtual devices iOS
+    /// restricts these at runtime (e.g. while recording or with video
+    /// stabilization active) and silently clamps out-of-range assignments
+    /// rather than throwing. The factor is read back after the assignment so
+    /// the returned zoom is what the camera actually applied, never the
+    /// requested value.
+    ///
+    /// Returns the applied zoom plus the live available range as a map for
+    /// the platform channel, or nil when the device rejected the change.
+    func setZoomLevel(level: CGFloat) -> [String: Double]? {
+        guard let device = videoDevice else { return nil }
+
+        // Without auto lens switch the lower bound stays at 1.0 so the
+        // native HAL never switches a virtual device to the ultra-wide,
+        // mirroring updateCameraProperties.
+        var liveMin = device.minAvailableVideoZoomFactor * nativeToUserZoomScale
+        if !autoLensSwitchRequested {
+            liveMin = max(liveMin, 1.0)
+        }
+        let liveMax = device.maxAvailableVideoZoomFactor * nativeToUserZoomScale
+
+        let clampedLevel = max(liveMin, min(level, liveMax))
         // Convert user-facing zoom to native videoZoomFactor
         let nativeZoom = clampedLevel / nativeToUserZoomScale
-        
+
         do {
             try device.lockForConfiguration()
             device.videoZoomFactor = nativeZoom
             device.unlockForConfiguration()
-            currentZoom = clampedLevel
-            return true
         } catch {
-            return false
+            return nil
         }
+
+        currentZoom = device.videoZoomFactor * nativeToUserZoomScale
+        // Keep the stored bounds in sync so getCameraState reports the same
+        // range; refreshed on every set, so it heals when a restriction lifts.
+        minZoom = liveMin
+        maxZoom = liveMax
+        return [
+            "zoomLevel": Double(currentZoom),
+            "minZoomLevel": Double(liveMin),
+            "maxZoomLevel": Double(liveMax),
+        ]
     }
 
     // MARK: - Video Stabilization
--- a/mobile/packages/divine_camera/ios/Classes/CameraController.swift
+++ b/mobile/packages/divine_camera/ios/Classes/CameraController.swift
@@ -1220,7 +1220,10 @@ class CameraController: NSObject {
             // lens switching on virtual multi-camera devices.
             nativeToUserZoomScale = 1.0
             minZoom = 1.0
-            maxZoom = device.activeFormat.videoMaxZoomFactor
+            // Use the live available max (mirrors the auto-lens branch and
+            // setZoomLevel) rather than the static format max, so a runtime
+            // restriction that lowers the ceiling isn't briefly over-advertised.
+            maxZoom = device.maxAvailableVideoZoomFactor
         }
         currentZoom = device.videoZoomFactor * nativeToUserZoomScale
         // Front camera has "flash" via screen brightness when feature is enabled
--- a/mobile/packages/divine_camera/lib/divine_camera.dart
+++ b/mobile/packages/divine_camera/lib/divine_camera.dart
@@ -18,6 +18,7 @@ export 'src/models/audio_device.dart';
 export 'src/models/camera_lens.dart';
 export 'src/models/camera_lens_metadata.dart';
 export 'src/models/camera_state.dart';
+export 'src/models/camera_zoom_state.dart';
 export 'src/models/flash_mode.dart';
 export 'src/models/photo_capture_result.dart';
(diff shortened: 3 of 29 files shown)


mobile/test/blocs/video_recorder/video_recorder_bloc_test.dart
```diff
@@ -249,19 +262,47 @@ void main() {
...
+        bloc.add(VideoRecorderScaleStarted(ScaleStartDetails(pointerCount: 1)));
+        await Future<void>.delayed(Duration.zero);
+        expect(bloc.state.isPinchActive, isTrue);
+
+        bloc.add(const VideoRecorderScaleEnded());
+        await Future<void>.delayed(Duration.zero);
+        expect(bloc.state.isPinchActive, isFalse);
+      });
+
+      test('is cleared by the indicator auto-hide timer as a safety net '
```
realmeylisdev, line 323: **should-fix (flaky timing).** This new safety-net test `await`s 1100ms of wall-clock and expects the bloc's real 1000ms `_zoomIndicatorHideDelay` Timer to have already fired — a 100ms margin, with no scale-end event dispatched, so the assertion depends entirely on that Timer landing inside the window.

Under CI's `very_good test --optimization` (all untagged suites bundled into one isolate, ~11k tests) event-loop saturation can slip the Timer callback past 1100ms, leaving `isPinchActive` still `true` at the assertion → intermittent red on an unrelated `--test-randomize-ordering-seed`. This is the same shape as the sibling focus-point auto-hide test in this file (800ms timer vs 900ms delay, same 100ms margin) that has tripped CI on PRs that never touched the camera.

The assertion is worth keeping — the behavior it checks is real. Please drive the timer deterministically with `fakeAsync` (`async.elapse(const Duration(milliseconds: 1000))` + flush microtasks so `_VideoRecorderZoomIndicatorTimerFired` processes) rather than racing wall-clock. Note `blocTest`'s `wait:` won't help — it's also a real `Future.delayed` and races identically. (The pre-existing `auto-hides the zoom ruler after the pinch settles` test above has the same shape; folding both into one `fakeAsync` conversion would be ideal, but since this test is new it's cleanest to write it deterministically from the start.)

realmeylisdev (commented):
Reviewed the full diff plus both native implementations against ground-truth code (no local iOS/Android device build in this env — Swift/Kotlin verified by inspection). This is a solid, well-tested fix; no functional blockers.

The core is correct:
- **iOS** clamp-against-live-`min/maxAvailableVideoZoomFactor` + read-back is NSException-safe (the clamp keeps `nativeZoom` inside the format range, so the direct `videoZoomFactor` assignment can't raise an uncatchable range exception) and returns the *applied* value, never the request.
- The `bool -> double?/CameraZoomState?` migration is complete across every layer and test double; bounds self-heal on any in-range set (getter chain `CameraMobileService -> DivineCamera -> _state` stays consistent).
- The pinch-passthrough fix (`isPinchActive` + `onScaleEnd`, `IgnorePointer(ignoring: !showZoomIndicator || isPinchActive)`) and its bloc/widget regression tests genuinely fail on revert.
- Cross-platform channel contract (`zoomLevel`/`minZoomLevel`/`maxZoomLevel`) is coherent; macOS `result(nil)` + deleted controller method leave no dangling reference. No Kotlin/Swift test references the changed signature, so the return-type change won't break the Android unit-test job.

All findings below are non-blocking — one should-fix (test flakiness) and three nits. Nothing Nostr/backend-related; mobile camera only.

- **should-fix:** new safety-net bloc test races a 1.1s wall-clock wait against the real 1.0s hide timer — the merged-isolate flake pattern already seen in this file.
- **nit:** one-pointer path leaves `isPinchActive` stuck if `onScaleEnd` is lost, vs what the safety-net comment claims.
- **nit:** iOS non-auto-lens `maxZoom` writer still sources the static format max, inconsistent with the live range used elsewhere.
- **nit:** `_onScaleUpdated` guard's pointer-count term is a test mutation blind spot.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
```diff
@@ -1194,25 +1207,49 @@ class VideoRecorderBloc
...
         baseZoomLevel: state.zoomLevel,
         snappedTo1x: false,
         lastRawZoom: state.zoomLevel,
-        showZoomIndicator: true,
+        showZoomIndicator: isPinch ? true : null,
+        isPinchActive: true,
       ),
     );
-    _armZoomIndicatorHideTimer();
+    if (isPinch) _armZoomIndicatorHideTimer();
```
realmeylisdev, line 1232: **nit (correctness / stale comment).** `_onScaleStarted` sets `isPinchActive: true` unconditionally (L1229) but arms the auto-hide timer only when `isPinch`, and `_onScaleUpdated` early-returns for one-pointer `scale == 1.0` updates before reaching `_armZoomIndicatorHideTimer()`. So on a pure one-finger pan no timer is ever scheduled — if `onScaleEnd` is dropped (e.g. the gesture is cancelled by a route push mid-pan), `isPinchActive` stays stuck `true` and `_onZoomIndicatorTimerFired`'s safety net never runs to clear it.

The safety-net comment in `_onZoomIndicatorTimerFired` reads as if it recovers a lost scale-end generally, but it only covers the timer-armed pinch / zoom-set paths. Mostly cosmetic, but not fully benign: a later `VideoRecorderZoomLevelSet` (long-press-drag or a quick-zoom button) sets `showZoomIndicator: true` without touching `isPinchActive`, so with the flag stuck the ruler evaluates `ignoring: !true || true == true` and stays non-interactive for that whole visible window (self-heals when that zoom's own timer fires).

Cheapest robust fix: arm the timer unconditionally here (the timer-fired handler is a no-op when nothing's pending), so the safety net actually covers the one-pointer path the comment describes. Otherwise, tighten the comment to say it only covers timer-armed paths — and note the paired test at `video_recorder_bloc_test.dart:287` exercises the two-pointer case only (`pointerCount: 2`), so a one-pointer variant of it would currently fail.

mobile/packages/divine_camera/ios/Classes/CameraController.swift
```diff
@@ -1737,22 +1737,51 @@ class CameraController: NSObject {
...
         } catch {
-            return false
+            return nil
         }
+
+        currentZoom = device.videoZoomFactor * nativeToUserZoomScale
+        // Keep the stored bounds in sync so getCameraState reports the same
+        // range; refreshed on every set, so it heals when a restriction lifts.
+        minZoom = liveMin
+        maxZoom = liveMax
```
realmeylisdev, line 1782: **nit (consistency).** `setZoomLevel` stores `maxZoom = maxAvailableVideoZoomFactor * nativeToUserZoomScale` (the live range), and the `updateCameraProperties` auto-lens branch already agrees — but its non-auto-lens branch stores `device.activeFormat.videoMaxZoomFactor` (the static format max). `getCameraState` just echoes whichever was written last.

So with `enableAutoLensSwitch: false`, if a runtime restriction (video stabilization / geometric distortion correction) lowers `maxAvailableVideoZoomFactor` below `activeFormat.videoMaxZoomFactor`, `getCameraState().maxZoomLevel` reports the higher format max before any zoom, then drops to the live max after the first `setZoomLevel` — a transient, briefly over-advertised bound in the ruler.

Not a functional bug (correct at each read; the default auto-lens path is fully consistent), but since this PR's whole point is to report the live range everywhere, that non-auto branch is the last writer left on the static source. `maxZoom = device.maxAvailableVideoZoomFactor` there (scale is 1.0 in that branch) would align both writers. Fine to leave if the static max is intentional.

mobile/test/blocs/video_recorder/video_recorder_bloc_test.dart
```diff
@@ -222,20 +222,33 @@ void main() {
...
         addTearDown(bloc.close);
 
-        bloc.add(VideoRecorderScaleStarted(ScaleStartDetails()));
+        bloc.add(VideoRecorderScaleStarted(ScaleStartDetails(pointerCount: 2)));
         await Future<void>.delayed(Duration.zero);
 
         expect(bloc.state.showZoomIndicator, isTrue);
       });
 
+      test('does not summon the zoom ruler on a one-pointer pan', () async {
```
realmeylisdev, line 236: **nit (test coverage).** The pointer-count term of the `_onScaleUpdated` guard (`pointerCount < 2 && scale == 1.0`) has no test on its proceed-side. This test covers the return-side (pointerCount 1, scale 1.0), and the snap/multiplicative tests proceed via the `scale != 1.0` term (default `pointerCount 0`) — but nothing drives `pointerCount >= 2 && scale == 1.0`, i.e. the just-landed second finger of a pan→pinch escalation that must NOT be skipped. Changing `< 2` to `< 3` (or dropping the pointer-count term) would leave every test green while breaking real escalating pinches. A small case would close it:

```dart
test('a landed second finger with unchanged scale still summons the ruler', () async {
  final bloc = buildBloc();
  addTearDown(bloc.close);
  bloc.add(VideoRecorderScaleStarted(ScaleStartDetails(pointerCount: 1)));
  bloc.add(VideoRecorderScaleUpdated(ScaleUpdateDetails(pointerCount: 2)));
  await Future<void>.delayed(Duration.zero);
  expect(bloc.state.showZoomIndicator, isTrue);
});
```

mobile/test/blocs/video_recorder/video_recorder_bloc_test.dart
hm21: Good catch — fixed in 697c6b149. Both auto-hide tests in the zoom group now drive the 1000ms timer with `fakeAsync` (`elapse` + `flushMicrotasks`) instead of the 1100ms wall-clock wait, so no more merged-isolate race. The bloc is closed inside the fake zone too, since a real-zone `addTearDown` would await microtasks the dead zone can no longer drain.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
hm21: Tightened the comment in 697c6b149 rather than arming the timer unconditionally. Arming on every scale-start would clear `isPinchActive` mid-gesture on a >1s one-finger pan that then escalates to a pinch — `_onScaleUpdated` never re-asserts the flag, so the second finger would get swallowed by the ruler again, which is exactly the bug this PR fixes. The comment now only claims the timer-armed paths it actually covers.

mobile/packages/divine_camera/ios/Classes/CameraController.swift
hm21: Aligned in 697c6b149 — the non-auto-lens branch now uses `device.maxAvailableVideoZoomFactor` (scale is 1.0 there), matching `setZoomLevel` and the auto-lens branch so the ceiling is not briefly over-advertised.

mobile/test/blocs/video_recorder/video_recorder_bloc_test.dart
hm21: Added in 697c6b149 — `summons the zoom ruler when a second finger lands mid-pan` drives `pointerCount: 2` with the default `scale: 1.0`, covering the proceed-side. Verified it goes red if the guard pointer-count term is mutated (`< 2` → `< 3`).

hm21:
@realmeylisdev thanks for the thorough pass. All four addressed in 697c6b149:

- **should-fix (flaky timer):** both zoom-group auto-hide tests converted to `fakeAsync`.
- **nit (stale comment):** tightened rather than arming the timer unconditionally — the latter would clear `isPinchActive` mid-gesture on a slow pan→pinch and reintroduce the swallowed-second-finger bug.
- **nit (Swift consistency):** non-auto-lens branch now reports the live `maxAvailableVideoZoomFactor`.
- **nit (coverage):** added the two-pointer/`scale == 1.0` escalation test.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
```diff
@@ -1194,25 +1207,49 @@ class VideoRecorderBloc
...
 
   Future<void> _onScaleUpdated(
     VideoRecorderScaleUpdated event,
     Emitter<VideoRecorderBlocState> emit,
   ) async {
+    // One-pointer updates with an unchanged scale are pans, not pinches
+    // (a single pointer keeps scale at exactly 1.0; trackpad pinches
+    // report a scale != 1.0). Skipping them keeps plain preview pans from
+    // summoning and re-arming the zoom ruler.
+    if (event.details.pointerCount < 2 && event.details.scale == 1.0) {
```
realmeylisdev, line 1250: **should-fix (state machine — reopens the swallowed-finger bug in a narrower trigger).** The proceed-side emit below (L1306) re-shows the ruler and re-arms the hide timer but never re-asserts `isPinchActive`. That leaves a hole the round-1 thread didn't cover:

A two-finger pinch held stationary for >1s fires no `onScaleUpdate`, so the 1s hide timer (armed at scale-start) fires *while the pinch is still down*. `_onZoomIndicatorTimerFired` then clears `isPinchActive: false` — its comment ("a still-set flag means the scale-end callback was lost") is false in this case; the flag was legitimately set. When the pinch resumes, this handler proceeds (`pointerCount >= 2`) and emits `showZoomIndicator: true` with `isPinchActive` still `false`, so `IgnorePointer(ignoring: !showZoomIndicator || isPinchActive)` evaluates `false` → the ruler is interactive during the ongoing pinch, and the lower pinch finger on the ruler band gets swallowed — exactly what `isPinchActive` was documented to block "for the whole gesture."

Fix: re-assert `isPinchActive: true` on the proceed-side emit (L1306). Bonus: that also removes the objection you raised on the round-1 thread for not arming the timer unconditionally — with the flag re-asserted on every genuine scale-update, the pan→pinch escalation case is covered too, so this is the more complete fix than tightening the comment.

realmeylisdev (commented):
Round-1 items (697c6b149) all verified fixed — fakeAsync timer conversion, tightened safety-net comment (rationale holds: `_onScaleUpdated` never re-asserts `isPinchActive`), iOS non-auto `maxZoom` → live `maxAvailableVideoZoomFactor`, and the mid-pan escalation test (goes red under the `< 2`→`< 3` mutation). `divine_camera` (254) + the video-recorder bloc/widget suites pass locally; CI green.

A fuller full-diff pass turned up three items I missed in round 1 — all pre-existing in the PR, none introduced by the fix commit:

**2. should-fix — integration test not migrated to the new return type.** `integration_test/video_recorder/camera_controls_integration_test.dart` L99 / L107 / L115 still do `expect(await cameraService.setZoomLevel(x), isA<bool>())`, but the signature is now `Future<double?>`. The applied zoom (a `double`, or `null`) is not a `bool`, so all three zoom-control patrol tests fail at runtime. `flutter analyze` stays green (the matcher compiles) and the standard Tests job doesn't run patrol, so CI is green. The unit doubles were all migrated; this file slipped through. Assert against the new contract instead (e.g. `isA<double?>()` / a bounds check).

**3. nit — the runtime-range heal is directional.** Bounds refresh only on a dispatched set's read-back; `stopRecording` and `setVideoStabilizationMode` don't refresh them, and there's no native→Dart push. If a runtime restriction lowers the max, the user pinches to that lowered max, then the restriction lifts (e.g. recording stops), the `(state.zoomLevel - clampedZoom).abs() > 0.01` guard in `_onScaleUpdated` suppresses the would-be healing set (clamped target == current), so pinch-out stays capped at the stale max until the user first pinches *in*. Narrow, and still an improvement over pre-PR behaviour, but the "heals in both directions" wording in `CameraZoomState` / `setZoomLevel` doesn't hold at the pinned-at-max boundary.

Not blocking — flagging for your call. The one standing caveat from the PR description holds: iOS Swift isn't built in CI; I verified it compiles by inspection (all referenced properties exist; clamp-then-readback stays inside the live range so the `videoZoomFactor` assignment can't raise an uncatchable range exception), but it still wants an on-device confirm.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
hm21: Confirmed — real hole, and re-asserting on the proceed-side is the better call than arming the timer unconditionally (it dodges the round-1 escalation objection you noted). Fixed in 1b7212a53: `isPinchActive: true` on the `_onScaleUpdated` proceed emit, so a resumed update restores the flag when the auto-hide timer fired mid-pinch (two fingers held stationary >1s land the timer while the gesture is still down). Also tightened the timer-fired safety-net comment and added a regression test — `is re-asserted when a stationary pinch resumes after the timer fired mid-gesture` — which goes red without the re-assert.

realmeylisdev (approved):
Approving — I re-verified the fix and none of the open items is a correctness blocker.

Re-traced both root causes and ran the migration end to end: the `bool`→`double?`/`CameraZoomState?` change is complete across every layer, the native clamp-then-readback stays inside the live range, and the new `divine_camera` tests are non-tautological (they assert applied != requested on a clamp and that the range heals). Solid root-cause writeup.

Three non-blocking residuals, all one-liners — worth folding into the merge push rather than a follow-up:

1. The `isPinchActive` re-assert you agreed to on the `video_recorder_bloc.dart:1250` thread isn't in the current head (`697c6b149`) yet (reply left there). Impact is narrower than I first framed — Flutter fixes pointer routing at down-time, so a plain hold-then-resume is inert — but the state invariant is still worth closing.
2. `camera_controls_integration_test.dart:99/107/115` still assert `isA<bool>()` on `setZoomLevel`'s now-`double?` return; they'd fail if the patrol suite ran (CI doesn't run it today).
3. `divine_camera.dart`'s "heals in both directions" comment overstates it — the heal only fires on an inward set, so a monotonic zoom-in from a stale max never triggers the read-back. Comment tweak.

None blocks merge.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
realmeylisdev: Gentle nudge: the re-assert we landed on here — plus the three `isA<bool>()` patrol asserts at `camera_controls_integration_test.dart:99/107/115` that are now on a `double?` return — aren't in the current head (`697c6b149`). Not blocking (I've approved), but both are one-liners, so worth folding into the merge push.

Re-traced this since the last pass, and I owe a correction: the invariant break is real, but a plain hold-then-resume with the same two fingers is actually inert — Flutter fixes each pointer's routing at down-time, so un-ignoring the ruler mid-gesture can't recapture fingers already owned by the preview recognizer. The swallow only bites a fresh finger-down on the band during the resumed pinch, which is narrower than I implied earlier.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
```diff
@@ -1297,7 +1340,17 @@ class VideoRecorderBloc
```
NotThatKindOfDrLiz, line 1296: The dispatch guard immediately below this still lets zoom get stuck at a stale restricted ceiling. Example: iOS stabilization lowers the live max to 9.5x, the user reaches 9.5x, then the restriction lifts. The bloc/service still advertise max 9.5, so a pinch-out produces `newZoom > maxZoomLevel` but `clampedZoom == state.zoomLevel`; `(state.zoomLevel - clampedZoom).abs() > 0.01` is false, so `VideoRecorderZoomLevelSet` is never sent and the native read-back never reports the widened live max. Please dispatch when the raw target pushes past a cached bound even if the clamped zoom is unchanged, so the same pinch heals the ceiling.

mobile/packages/divine_camera/lib/divine_camera.dart
```diff
@@ -216,15 +217,26 @@ class DivineCamera {
...
-  Future<bool> setZoomLevel(double level) async {
+  ///
+  /// Returns the zoom level the camera actually applied, or `null` when
+  /// the platform failed to apply it. The applied zoom can be lower than
+  /// [level]: on iOS virtual multi-camera devices the system restricts
+  /// the available zoom range at runtime and silently clamps out-of-range
+  /// assignments, so the platform reads the factor back after setting it.
+  /// The live range reported alongside it replaces [minZoomLevel] /
+  /// [maxZoomLevel], keeping [state] honest in both directions (it heals
+  /// once a runtime restriction lifts).
```
NotThatKindOfDrLiz, line 228: This doc overclaims until the bloc dispatch-at-bound case is handled. The live range only heals after `setZoomLevel` actually reaches the platform; when the app is pinned at a stale max and pinch-out is clamped back to the current zoom, the bloc currently skips that call. If you fix the dispatch guard, this wording becomes accurate enough; otherwise narrow it to say bounds refresh on a successful zoom set.

NotThatKindOfDrLiz (changes requested):
Requesting changes for one remaining correctness gap: the PR now heals shrunk runtime zoom bounds only after a dispatched set, but the app can suppress that set exactly when the user is pinned at the stale ceiling and pinches out after the restriction lifts. Please either dispatch at the bound so the live read-back can heal on that gesture, or narrow the docs if you intentionally do not want that behavior. I think the code fix is the right move here because it is small and closes the named symptom.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
hm21: Fixed in b333b2e078 — took the code-fix path. `_onScaleUpdated` now also dispatches `VideoRecorderZoomLevelSet` when the raw pinch target pushes past a cached bound, even when the clamp pins the applied zoom to the current level, so the read-back heals the widened range on that same pinch-out. Regression test added (`heals a stale shrunk ceiling on pinch-out even when the clamp pins the zoom to the current level`); confirmed red without the fix.

mobile/packages/divine_camera/lib/divine_camera.dart
hm21: The dispatch-at-bound fix landed in b333b2e078, so the "heals in both directions" wording now holds end to end — left the doc as-is.

hm21:
@NotThatKindOfDrLiz addressed the dispatch-at-bound gap in b333b2e078 (code fix, not doc narrowing). `_onScaleUpdated` now dispatches when the raw pinch target pushes past a cached bound even if the clamp pins the applied zoom to the current level, so the read-back heals the widened range on the same pinch-out — no longer waits for a pinch-in. Regression test added, confirmed red without the fix.

rabble:
Went through @NotThatKindOfDrLiz's CHANGES_REQUESTED point-by-point against the code at head — both her threads are genuinely fixed, not just resolved:

1. **`_onScaleUpdated` dispatch guard skipping the heal on a pinned stale ceiling** — fixed at `mobile/lib/blocs/video_recorder/video_recorder_bloc.dart:1328-1333`. New `pushedPastBound` term (`newZoom > _cameraService.maxZoomLevel || newZoom < _cameraService.minZoomLevel`) is OR'd into the dispatch guard, so a pinch-out past a cached bound still dispatches `VideoRecorderZoomLevelSet` and triggers the native read-back that widens the max. Traced the full chain: dispatched value is `clampedZoom`, `_onZoomLevelSet` has no unchanged-value early return and its bounds check passes on equality, so it reaches `_cameraService.setZoomLevel()` → native read-back → refreshed min/max. Regression test at `video_recorder_bloc_test.dart:925` asserts `state.maxZoomLevel == 5.0` on the stale→live flip and would go red without the fix.
2. **`divine_camera.dart` "heals in both directions" doc** — now correct because point 1 took the code-fix path Liz preferred.

realmeylisdev's non-blocking residuals also landed (`isPinchActive` re-assert at `:1317`; patrol test moved off `isA<bool>()` to `isNotNull` + `inInclusiveRange`).

Only standing caveat, not a review item: the iOS Swift half isn't compiled in CI, so an on-device confirm is still worth doing before ship. Flagging for a fresh review to clear the stale block.

rabble:
Re-review nudge: I believe this is unblocked on the code side. All review threads are resolved, and CI (Analyze, Tests, Android Unit Tests, Format, Generated Files, build/build, semantic-pull-request) is green on the current head.

In particular, @NotThatKindOfDrLiz — your blocking concern about the zoom-heal dispatch guard suppressing the healing `setZoomLevel` when the user is pinned at a stale shrunk ceiling was addressed in the current head: a `pushedPastBound` term now dispatches when the raw target exceeds the cached bound even if the clamp pins the applied zoom, with a regression test ("heals a stale shrunk ceiling on pinch-out even when the clamp pins the zoom to the current level") confirmed red without the fix. Both of your threads are resolved.

The PR is still in CHANGES_REQUESTED only because the formal review predates that push. Could you re-review and clear/refresh it?

One honest caveat that remains: the iOS Swift camera path isn't built in CI, so the native half was verified by inspection only — an on-device confirm on a real iOS multi-camera device is still worth doing before/after merge.

NotThatKindOfDrLiz (approved):
Re-reviewed current head b333b2e078. The stale zoom-ceiling blocker is fixed by dispatching when the raw pinch target pushes past cached bounds, so the platform read-back can heal the live range even when the clamped zoom equals the current level. Targeted bloc/widget and divine_camera package tests pass locally. No blocking findings remain; still worth doing the noted on-device iOS multi-camera confirmation.

PR #6066, opened 2026-07-13
fix(audio-attribution): credit the original creator for reused audio

### LINKED ISSUE #6057: fix(audio-attribution): reused audio incorrectly shown as original sound
### Summary

Reusing someone else's audio incorrectly labels it as my original sound instead of crediting the original creator.

### Environment

any

### Steps to Reproduce

1. Open a post that uses someone else's audio.
2. Reuse the audio to create a new post.
3. Publish or preview the post.
4. Check the audio attribution.

### Actual Result

The audio is labeled as my original sound instead of crediting the original creator.

### Expected Result

The reused audio should retain attribution to the original creator.

### Regression?

No

### Evidence

reported on discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1525486923957669888)

hm21:
## Description

Reusing another creator's sound published the new video with **no audio attribution**, so the feed and the metadata "Sounds" section credited the reusing user as the "original sound" creator instead of the original creator.

**Root cause:** a reused *original sound* (the audio of a video, which has no standalone Kind 1063 event) becomes a synthetic `AudioEvent` whose id is `video_<sourceVideoId>`. The publisher gated the audio reference on `NostrHexUtils.isValidEventId`, which rejected the `video_` prefix, so the `["e", …, "audio"]` tag was dropped and the published video carried no attribution — falling through to the "Original sound — @me" display.

### Changes

- **models** — `AudioEvent.attributionEventId` recovers the real Nostr event id behind the `video_` prefix (and the editor timeline's `-<timestamp>` uniqueness suffix), validating it as 32-byte hex; `fromVideoOriginalSound.creatorName` is now optional (null → null title, so the reader localizes "Original sound" and resolves the creator from the pubkey).
- **publisher** — resolves the reused reference via `attributionEventId`, so `["e", <sourceVideoId>, relay, "audio"]` survives. Bundled/imported behavior is unchanged.
- **editor "add music"** — `getActiveDraft` derives the publish attribution from the editor timeline tracks (skipping the video's own clip-anchored audio), so the editor reuse path is covered too, not only the recorder's `selectedSound` path.
- **sounds_repository** — `fetchSoundById` synthesizes a video-backed original sound when the reference resolves to a Kind 34236 video, crediting that video's creator.
- **metadata sheet** — when the shared audio event can't be fetched, the fallback credits the source creator that the video already carries via its inspired-by source (no network fetch) instead of dropping to the reusing user.
- **metadata screen** — "Publish this sound" is hidden entirely when the video reuses external audio: it isn't the user's to offer for reuse, and the publisher already skips the reuse tag in that case, so the toggle was a no-op.

### Why these choices

- **Reference the source video (not a fresh Kind 1063).** Any Kind 1063 the app publishes is signed by the current user, so its `pubkey` is the reusing user — the display would still credit them. The only event owned by the original creator is their Kind 34236 video, so that is what the reference points to, and the display derives the creator from it.
- **Recover the id in the model, not at each call site.** The `video_` prefix and the editor's timeline suffix both break a plain hex check; centralizing that in `attributionEventId` keeps the publisher and the editor-track derivation in sync and unit-testable.
- **Fetch-independent display fallback.** Kind 34236 is addressable, so fetching it by event id can fail on relays that only serve it by address. The metadata sheet therefore credits the creator from the inspired-by source already on the video, which a reused sound auto-populates — so attribution is correct even when the source video isn't reachable by id.

## Out of Scope

- Videos published **before** this change carry no audio-reference tag, so they cannot be credited retroactively (the information isn't in the event) and will continue to show the reusing user for their reused audio.

## Verification

- `flutter analyze` on all changed files and both touched packages — clean.
- New/updated tests, all green:
  - `packages/models` — `attributionEventId` (prefix/suffix/bundled/import cases) and null-`creatorName`.
  - `packages/sounds_repository` — Kind 34236 synthesis in `fetchSoundById`.
  - `video_event_publisher` — reused original sound and editor-suffixed track both write `["e", <sourceVideoId>, relay, "audio"]`.
  - `video_editor_provider` — editor-track attribution in `getActiveDraft` (and skipped on autosave / for clip-anchored own audio); `reusesExternalAudio`.
  - `metadata_sounds_section` — reused fallback credits the source creator, not the reusing user.
  - `video_metadata_form_fields` — "Publish this sound" hidden when reusing external audio.
- Manual test plan: publish a fresh video reusing another creator's sound (recorder "Use Sound" and editor "Add music"), then check the metadata "Sounds" section credits the original creator; a self-recorded original sound still credits you and keeps the toggle.

**Related Issue:** Closes #6057

## Type of Change

### DIFF
--- a/mobile/lib/services/video_event_publisher.dart
+++ b/mobile/lib/services/video_event_publisher.dart
@@ -1326,10 +1326,15 @@ class VideoEventPublisher {
       final hasSelectedAudioEventId =
           selectedAudioReferenceId != null &&
           selectedAudioReferenceId.isNotEmpty;
+      // A reused *original sound* carries the source video's event id behind a
+      // `video_` prefix (and, from the editor timeline, a `-<timestamp>`
+      // uniqueness suffix). Fall back to [AudioEvent.attributionEventId] to
+      // recover the real event id so the reference survives instead of being
+      // dropped and the audio mislabelled as the reusing user's own sound.
       final reusableSelectedAudioEventId =
           NostrHexUtils.isValidEventId(selectedAudioReferenceId)
           ? selectedAudioReferenceId
-          : null;
+          : selectedAudio?.attributionEventId;
       if (hasSelectedAudioEventId && reusableSelectedAudioEventId == null) {
         Log.warning(
           'Skipping selected audio reference because it is not a Nostr event id: '
--- a/mobile/packages/models/lib/src/audio_event.dart
+++ b/mobile/packages/models/lib/src/audio_event.dart
@@ -3,6 +3,7 @@
 // ABOUTME: parsing audio shared for use in other videos
 
 import 'package:meta/meta.dart';
+import 'package:models/src/nostr_hex_utils.dart';
 import 'package:models/src/video_event.dart';
 import 'package:models/src/vine_sound.dart';
 import 'package:nostr_sdk/event.dart';
@@ -167,9 +168,16 @@ class AudioEvent {
   /// Uses the video URL as the audio source. The audio player can extract
   /// the audio track from video files. The ID is prefixed with `video_`
   /// to distinguish from real Kind 1063 events.
+  ///
+  /// When [creatorName] is provided the [title] reads
+  /// `Original sound - $creatorName`; when omitted the [title] is left null so
+  /// the display can
+  /// resolve the creator from [pubkey] and localize the "Original sound"
+  /// label itself (used when synthesizing attribution for a reused original
+  /// sound, where the reader owns the localization).
   factory AudioEvent.fromVideoOriginalSound(
     VideoEvent video, {
-    required String creatorName,
+    String? creatorName,
   }) {
     return AudioEvent(
       id: 'video_${video.id}',
@@ -177,7 +185,7 @@ class AudioEvent {
       createdAt: video.createdAt,
       url: video.videoUrl,
       duration: video.duration?.toDouble(),
-      title: 'Original sound - $creatorName',
+      title: creatorName == null ? null : 'Original sound - $creatorName',
       source: 'Original Sound',
       sourceVideoReference: '34236:${video.pubkey}:${video.vineId ?? video.id}',
     );
@@ -246,6 +254,29 @@ class AudioEvent {
 
   /// Whether this audio is derived from a video's original sound.
   bool get isOriginalSound => id.startsWith('video_');
+
+  /// The Nostr event id a reusing video should reference in its
+  /// `["e", <id>, <relay>, "audio"]` attribution tag, or `null` when this
+  /// audio has no referenceable Nostr event (bundled sounds, or imported
+  /// audio that hasn't been published yet).
+  ///
+  /// Resolves two id shapes on top of a plain Kind 1063 event id:
+  /// * a reused *original sound*, whose [id] is `video_<eventId>` — the source
+  ///   Kind 34236 video's own event id (see
+  ///   [AudioEvent.fromVideoOriginalSound]); and
+  /// * an editor timeline track, whose id carries a trailing `-<timestamp>`
+  ///   uniqueness suffix so the same sound can be added to the timeline more
+  ///   than once.
+  ///
+  /// Without this the `video_` prefix / timeline suffix make the id fail a
+  /// 32-byte-hex check, the attribution tag is dropped at publish, and the
+  /// reused audio is mislabelled as the reusing user's original sound.
+  String? get attributionEventId {
+    var candidate = isOriginalSound ? id.substring('video_'.length) : id;
+    final suffixed = RegExp(r'^([0-9a-fA-F]{64})-\d+$').firstMatch(candidate);
+    if (suffixed != null) candidate = suffixed.group(1)!;
+    return NostrHexUtils.isValidEventId(candidate) ? candidate : null;
+  }
 
   /// Whether this track is a video's original sound that is still anchored to
   /// one of the editor's clips.
--- a/mobile/packages/models/lib/src/audio_event.dart
+++ b/mobile/packages/models/lib/src/audio_event.dart
@@ -265,16 +265,22 @@ class AudioEvent {
   ///   Kind 34236 video's own event id (see
   ///   [AudioEvent.fromVideoOriginalSound]); and
   /// * an editor timeline track, whose id carries a trailing `-<timestamp>`
-  ///   uniqueness suffix so the same sound can be added to the timeline more
-  ///   than once.
+  ///   uniqueness suffix (and, after a duplicate/split, one or more
+  ///   `_copy_<micros>` suffixes) so the same sound can appear on the
+  ///   timeline more than once.
   ///
   /// Without this the `video_` prefix / timeline suffix make the id fail a
   /// 32-byte-hex check, the attribution tag is dropped at publish, and the
   /// reused audio is mislabelled as the reusing user's original sound.
   String? get attributionEventId {
     var candidate = isOriginalSound ? id.substring('video_'.length) : id;
-    final suffixed = RegExp(r'^([0-9a-fA-F]{64})-\d+$').firstMatch(candidate);
-    if (suffixed != null) candidate = suffixed.group(1)!;
+    // Recover the leading 64-hex event id, dropping any editor-added
+    // uniqueness/duplicate suffixes (`-<timestamp>`, `_copy_<micros>`, or a
+    // chain of them).
+    final base = RegExp(
+      r'^([0-9a-fA-F]{64})(?:[-_].*)?$',
+    ).firstMatch(candidate);
+    if (base != null) candidate = base.group(1)!;
     return NostrHexUtils.isValidEventId(candidate) ? candidate : null;
   }
 
--- a/mobile/packages/sounds_repository/lib/src/sounds_repository.dart
+++ b/mobile/packages/sounds_repository/lib/src/sounds_repository.dart
@@ -266,12 +266,20 @@ class SoundsRepository {
           _cache[eventId] = audioEvent;
           return audioEvent;
         } on Object catch (e) {
+          // coverage:ignore-start
+          // Defensive net for adversarial relay data. For a video-kind event
+          // `VideoEvent.fromNostrEvent` does not throw (the kind is already
+          // validated by the `isVideoKind` gate above and `Event.tags` is
+          // strictly typed), so this is unreachable in practice; kept to
+          // downgrade a hypothetical parse failure to a warning + null instead
+          // of the outer error + rethrow.
           Log.warning(
             'Failed to synthesize original sound from video $eventId: $e',
             name: 'SoundsRepository',
             category: LogCategory.api,
           );
           return null;
+          // coverage:ignore-end
         }
       }
 
--- a/mobile/packages/sounds_repository/lib/src/sounds_repository.dart
+++ b/mobile/packages/sounds_repository/lib/src/sounds_repository.dart
@@ -6,7 +6,7 @@
 import 'dart:async';
 
 import 'package:models/models.dart'
-    show AudioEvent, NIP71VideoKinds, audioEventKind;
+    show AudioEvent, NIP71VideoKinds, VideoEvent, audioEventKind;
 import 'package:nostr_client/nostr_client.dart';
 import 'package:nostr_sdk/nostr_sdk.dart';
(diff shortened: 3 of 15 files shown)


mobile/packages/sounds_repository/lib/src/sounds_repository.dart
```diff
@@ -244,21 +244,52 @@ class SoundsRepository {
...
+      // instead of the reusing user. Title is left null so the reader
+      // localizes "Original sound" and resolves the creator name from the
+      // synthesized pubkey. Cached under [eventId] directly (not via
+      // [_cacheSound]) so it is not surfaced in the sounds-browser stream.
+      if (NIP71VideoKinds.isVideoKind(event.kind)) {
+        try {
+          final audioEvent = AudioEvent.fromVideoOriginalSound(
+            VideoEvent.fromNostrEvent(event),
+          );
+          _cache[eventId] = audioEvent;
```
realmeylisdev, line 266: **Tap-through dead-end for reused sounds (should-fix).** The synth's own id is `video_<sourceVideoId>` (`fromVideoOriginalSound`), but it's cached here under the raw `eventId`. When the metadata sheet resolves this synth it renders `_SoundListItem`, whose `_navigateToSoundDetail` (`metadata_sounds_section.dart:285`) pushes `pathForId(audio.id)` = `/sound/video_<id>` **with no `extra`**. The route then falls to `SoundDetailLoader`, which re-fetches `soundById('video_<id>')` → `fetchSoundById('video_<id>')`: cache miss (keyed under the raw id) + `fetchEventById('video_<id>')` (non-64-hex, matches nothing) → `null` → "Sound not found". This is the **common success path** (source video reachable → synth → `_SoundListItem`).

Contrast: `_OriginalSoundSection._navigateToSoundDetail` (line 194) and the feed's `AudioAttributionRow` both pass the resolved sound via `extra`, so they work — the asymmetry is the defect. On `main` this never fired because `fetchSoundById` returned `null` for a Kind-34236 ref, so `_SoundListItem` only ever got real 64-hex 1063 ids.

Fix: pass `extra: {'sound': audio}` from `_SoundListItem._navigateToSoundDetail` (mirror `_OriginalSoundSection`), and/or also cache the synth under its own `video_<id>` key + strip the `video_` prefix in `fetchSoundById`.

Follow-on: even once the screen opens, `SoundDetailScreen` (line 244 / `_VideosGrid`) queries usage-count and the grid by `widget.sound.id` = `video_<id>`, while real reuses tag the raw `<id>` — so both come back empty. Query by the raw id (`attributionEventId`) there too.

mobile/packages/models/lib/src/audio_event.dart
```diff
@@ -247,6 +255,29 @@ class AudioEvent {
...
+  /// * an editor timeline track, whose id carries a trailing `-<timestamp>`
+  ///   uniqueness suffix so the same sound can be added to the timeline more
+  ///   than once.
+  ///
+  /// Without this the `video_` prefix / timeline suffix make the id fail a
+  /// 32-byte-hex check, the attribution tag is dropped at publish, and the
+  /// reused audio is mislabelled as the reusing user's original sound.
+  String? get attributionEventId {
+    var candidate = isOriginalSound ? id.substring('video_'.length) : id;
+    final suffixed = RegExp(r'^([0-9a-fA-F]{64})-\d+$').firstMatch(candidate);
```
realmeylisdev: Edge (low): the editor's duplicate/split (`_copyId` → `${id}_copy_<micros>`, `video_editor_timeline_overlay_controls.dart:504`) produces ids like `video_<64hex>-<ts>_copy_<micros>`, which this regex (single trailing `-<digits>`) doesn't strip → `attributionEventId` returns `null`. If a user duplicates a reused sound and deletes the original segment (leaving only the copy), both `_attributionSoundFromEditorTracks` and `reusesExternalAudio` go false, publish drops the `audio` e-tag, and the toggle reappears — re-triggering #6057 for that edit sequence. Narrow, but the same failure class the PR fixes, and no test covers this id shape.

mobile/lib/services/video_event_publisher.dart
```diff
@@ -1326,10 +1326,15 @@ class VideoEventPublisher {
...
+      // A reused *original sound* carries the source video's event id behind a
+      // `video_` prefix (and, from the editor timeline, a `-<timestamp>`
+      // uniqueness suffix). Fall back to [AudioEvent.attributionEventId] to
+      // recover the real event id so the reference survives instead of being
+      // dropped and the audio mislabelled as the reusing user's own sound.
       final reusableSelectedAudioEventId =
           NostrHexUtils.isValidEventId(selectedAudioReferenceId)
           ? selectedAudioReferenceId
-          : null;
+          : selectedAudio?.attributionEventId;
```
realmeylisdev, line 1337: Design question (non-blocking): attribution now pins the source video's **immutable** Kind-34236 event id via the `e` tag, while inspired-by (just above) uses the durable **addressable** `a`-coordinate `34236:pubkey:dtag`. Kind 34236 is parameterized-replaceable, so if the source creator edits/replaces their video (new event id, same d-tag), this `e` reference — and the id-based synth in `sounds_repository` — can go stale even though the `a`-coordinate still resolves. Worth considering whether the audio reference (and funnelcake's usage analytics keyed off it) should track the `a`-coordinate for edit-resilience.

mobile/test/widgets/metadata_sounds_section_test.dart
```diff
@@ -186,5 +186,61 @@ void main() {
...
+          authorName: 'Jake Lara',
+          audioEventId: testAudioEventId,
+          inspiredByVideo: const InspiredByInfo(
+            addressableId: '34236:$reusedCreatorPubkey:vine-xyz',
+          ),
+        );
+      }
+
+      testWidgets(
+        'credits the source creator when the shared audio is unresolved',
```
realmeylisdev, line 215: Test gap (medium): this covers the `audio == null` fallback branch, but not the non-null path where `_SharedAudioSection` renders `_SoundListItem` with the synth — which is exactly where the tap-through dead-end lives (see the `sounds_repository.dart` comment). A widget test that taps the resolved reused-sound row and asserts it reaches `SoundDetailScreen` (not the not-found screen) would catch it. Also consider an end-to-end publish→parse round-trip asserting `["e", …, "audio"]` reads back to the source creator.

realmeylisdev (commented):
**Formal review — reused-audio attribution (#6057).** Read-only; findings only, no changes.

The core fix is correct and round-trips end-to-end: publish writes `["e", <sourceVideoId>, relay, "audio"]` plus the inspired-by `["a", "34236:…", relay, "mention"]`; the reader recovers both; the metadata "Sounds" section credits the original creator (via the Kind-34236 synth, or the inspired-by fallback); and the reuse toggle is correctly hidden for external audio. `attributionEventId` handles the `video_` / `-<ts>` / plain-1063 id shapes correctly, and there is no backend break or regression (verified against funnelcake).

**One should-fix before merge** (inline on `sounds_repository.dart`): tapping a reused sound in the metadata sheet dead-ends at **"Sound not found"** — the resolved synth id `video_<id>` is re-fetched by the detail loader and cannot resolve. It's in the common success path and is newly introduced by this PR.

Lower-priority (inline): duplicate/split of a reused track drops attribution again (`_copy_` suffix, `audio_event.dart`); `SoundDetailScreen` usage/grid query by the `video_`-prefixed id → empty; and the synth `_SoundListItem` path (where the dead-end lives) is untested.

**Cross-repo (informational, for the backend team — not this PR's concern):** the `audio` marker now targets a Kind-34236 id, so funnelcake's `audio_usage_counts` rows for reused original sounds are orphaned against `nostr.audios` (Kind 1063 only) — they won't surface in `/api/sounds` usage/trending — and `submit_audio_usage_feedback` (relay.rs) sends Gorse "use" feedback onto the *source video* item. Both are benign (not regressions — no audio tag was emitted at all before), but worth a heads-up.

Nice work — the attribution itself is solid.

mobile/packages/sounds_repository/lib/src/sounds_repository.dart
hm21: Good catch, confirmed. Fixed in 63528a5c: `_SoundListItem` now passes the resolved sound via `extra` (like `_OriginalSoundSection`), so the reused-sound tap opens the detail directly instead of re-fetching the synthetic `video_<id>` and dead-ending. Also switched `SoundDetailScreen` usage-count + grid to query by `attributionEventId` (the raw id real reuses tag), and gated the section label on whether a source video is actually shown so it stays coherent. Regression test added in `metadata_sounds_section_test.dart`.

mobile/packages/models/lib/src/audio_event.dart
hm21: Fixed in 63528a5c. `attributionEventId` now recovers the leading 64-hex id past any `-<ts>` / `_copy_<micros>` (and chained) suffixes, so a duplicated/split reused track keeps its attribution instead of re-triggering #6057. Added model tests for the `_copy_` and chained shapes.

mobile/lib/services/video_event_publisher.dart
hm21: Intentional for this PR: the `e` tag pins the exact source event that carried the audio, and the durable `a`-coordinate is already emitted alongside as inspired-by. Tracking the audio reference (and funnelcake analytics) by the `a`-coordinate for edit-resilience is a larger cross-repo change, so leaving as-is here. Thanks for the funnelcake usage-row heads-up — noted for the backend team.

mobile/test/widgets/metadata_sounds_section_test.dart
hm21: Added in 63528a5c: `metadata_sounds_section_test.dart` now taps the resolved reused-sound row through a GoRouter and asserts it reaches the detail (not the not-found screen). Also added a `SoundDetailScreen` test asserting usage/grid query by the recovered raw id.

hm21:
Thanks @realmeylisdev — addressed in 63528a5c:

- **Dead-end (should-fix):** reused-sound row now passes the resolved sound via `extra`; `SoundDetailScreen` queries usage/grid by the recovered `attributionEventId`, label gated to match the grid.
- **`_copy_` suffix (low):** `attributionEventId` strips `-<ts>` / `_copy_<micros>` / chained suffixes now.
- **Test gap (medium):** added a tap-through regression (metadata) + a raw-id query test (SoundDetailScreen).

Left the `e`-vs-`a` coordinate thread open — intentional for this PR (the durable `a`-coordinate already ships as inspired-by); tracking it by the coordinate is a larger cross-repo change.

mobile/test/widgets/metadata_sounds_section_test.dart
```diff
@@ -186,5 +188,137 @@ void main() {
...
+              GoRoute(
+                path: '/sheet',
+                builder: (_, _) => Scaffold(
+                  body: MetadataSoundsSection(video: createVideoWithAudio()),
+                ),
+              ),
+              GoRoute(
+                path: SoundDetailScreen.path,
+                name: SoundDetailScreen.routeName,
+                builder: (context, state) {
```
realmeylisdev, line 276: nit (non-blocking): this stand-in `/sound` route re-implements the `extra['sound']` read, so the test stays green even if the real `video_routes.dart` builder stopped reading `extra` — the exact #6057 dead-end could regress here undetected. Consider driving the real route (pump `videoRoutes`) so this test guards the production wiring rather than a copy of it.

realmeylisdev (approved):
Re-checked 63528a5c against the current source — the reused-audio feedback is addressed and holds up in production, not just in the tests:

- **Tap-through dead-end:** `_SoundListItem` now passes the resolved sound via `extra`, and the real `/sound` route (`video_routes.dart:74-88`) reads `extra['sound']` and threads it into `SoundDetailScreen` — so it opens the detail instead of re-fetching the synthetic `video_<id>` and dead-ending. Usage/grid now query the recovered raw id; the section label is co-gated on `showsSourceVideo`.
- **`_copy_`/chained suffix:** `attributionEventId` recovers the id past the editor suffixes, and the broadened regex can't over-recover a non-reuse id (every non-reuse id carries a non-hex prefix, and bare-64-hex was already attributed under the old regex).
- **Tests:** `models` / `sound_detail_screen` / `metadata_sounds_section` suites all green locally, including the two new regressions.

Fine to leave the `e`-vs-`a` coordinate thread for a follow-up. One small test-fidelity nit inline, non-blocking. Approving — thanks for the quick turnaround.

PR #6068, opened 2026-07-13
fix(proofmode): let users recover when C2PA signing or rendering fails

### LINKED ISSUE #6058: fix(proofmode): allow generation to be restarted after C2PA network failure
### Summary

Generation cannot be completed if the ProofMode (C2PA) step fails due to a missing network connection.

### Environment

any

### Steps to Reproduce

1. Start a generation.
2. Disconnect the network before the ProofMode (C2PA) step begins.
3. Wait for the generation to reach the final stage.

### Actual Result

The ProofMode (C2PA) step fails because there is no network connection, leaving the generation incomplete.

### Expected Result

If the ProofMode (C2PA) step fails due to a network issue, the user should be able to restart the generation without getting stuck.

### Regression?

No

### Evidence

discussed on slack [here](https://planetary-app.slack.com/archives/C09SZRQFWG3/p1783859439712949)

hm21:
## Summary

Generation could get **stuck** when the ProofMode / C2PA signing step lost the network. The remote signer had **no timeout**, so `signFile` could hang indefinitely — suspending the whole render/publish future. Because the future never returned, the orchestrators' `catch`/`finally` never ran, so the in-progress flags (`isProcessing`, `publishState == .preparing`) stayed set and **every restart was blocked**.

This PR makes generation resilient to C2PA (and general render) failures, and gives the user a clear way to recover.

Closes #6058

## Root cause

- `C2paSigningService.signVideo` calls a `RemoteSigner` (config fetch + RFC-3161 TSA timestamp) with no timeout anywhere in the chain → a half-open connection wedges the generation.
- `_runRenderVideo` had no `try/catch`; `publishVideo`'s `catch`/`finally` never reset `publishState`. So a thrown *or* hung failure left the flags stuck.

## What changed

**Never hang / always recoverable**
- Bounded the C2PA signing call with a timeout, so a network hang surfaces as a normal best-effort failure instead of suspending generation.
- Reset the editor `isProcessing` and publish `publishState` guards on failure so a stuck generation can always be restarted.

**Two distinct, appropriate failure UXs**

| Case | What the user sees | Retry action |
|---|---|---|
| **Render fails** (no clip produced) | "Generation failed" + retry icon overlay in the preview | full re-render |
| **Only C2PA fails** (clip produced, credential missing) | Bottom-sheet: "Post without the human-made check?" → *Regenerate* / *Post anyway* | **re-sign the existing render, no re-encode** |

- "Post anyway" publishes best-effort without the credential — with no C2PA there is no "Human Made" badge, so the video is simply not confirmed as camera-real (that's the warning).

**CI safety**
- The C2PA prompt is gated on a configured signing endpoint (`isSigningConfigured`), so CI and unconfigured builds never surface it — nothing blocks there.
- A debug-only `--dart-define=DIVINE_FORCE_C2PA_PROMPT=true` enables the prompt for local testing without a signing server (off by default, ignored outside debug).

## Notes for reviewers

- New l10n keys are translated into **all 18 locales**; `arb_consistency_test` is green.
- `kForceC2paPromptInDebug` is an intentional debug-only QA aid; the `🔐 C2PA prompt gate` log is now `debug`-level (off the happy path).
- Signing timeout is `20s` — a hang-guard sized above worst-case slow-but-alive mobile signing latency (two round-trips + TSA), so a slow network doesn't strip a valid credential; it only fires for a genuine hang.

### Review-hardening (follow-up commit)

- The C2PA-missing prompt is now **non-dismissible** so an accidental barrier tap / swipe can't silently forfeit provenance; a system-back re-signs instead of posting without the credential.
- Classic mode shows a progress overlay during a re-sign (matches capture mode).
- Render failure is announced to screen readers; decorative warning icon excluded from semantics.
- Added tests: null-render `renderFailed`, `copyWith(clearFinalRenderedClip)` flag reset, `retryC2paSigning` re-failure re-prompt, `regenerateVideo` parameter preservation, screen "Regenerate" wiring, non-dismissible prompt.

## Test plan

- `flutter analyze` clean across all changed files.
- Added/updated tests (all green): C2PA signing timeout + `TimeoutException` classification; `c2paSigningFailedFor` gate; `renderFailed` overlay + retry; C2PA prompt shows and "Post anyway" clears the flag; `retryC2paSigning` re-signs without re-rendering; publish `.preparing` guard reset.
- Manually verified in a debug build with `DIVINE_FORCE_C2PA_PROMPT=true` (prompt) and a temporary `throw` in `renderVideo` (generation-failed overlay).


mobile/lib/providers/video_editor_provider.dart
```diff
@@ -1098,43 +1117,191 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
...
 
-    state = state.copyWith(
-      isProcessing: false,
-      finalRenderedClip: finalRenderedClip,
-      proofManifestJson: proofManifestJson,
-    );
-    autosaveChanges();
+  /// Re-renders the current edit to re-attempt C2PA signing, keeping the edit
+  /// parameters (this is a provenance retry, not a content change).
+  Future<void> regenerateVideo() {
```
NotThatKindOfDrLiz: This method is effectively dead code in production. The only caller is `retryC2paSigning()` when `finalRenderedClip == null`, but `copyWith(clearFinalRenderedClip: true)` clears `c2paSigningFailed`, and the prompt only appears from `c2paSigningFailed == true`. So the user-facing `Regenerate` button never reaches this path; it re-signs the existing render instead. Please either wire the prompt's primary action to this method, or remove the method/test and rename the action/copy to match the actual re-sign behavior.

mobile/lib/screens/video_metadata/video_metadata_screen.dart
```diff
@@ -41,8 +46,57 @@ class _VideoMetadataScreenState extends ConsumerState<VideoMetadataScreen> {
...
+        notifier.acknowledgeC2paSigningFailure();
+    }
+  }
+
   @override
   Widget build(BuildContext context) {
+    // When the render finishes without a C2PA content credential (signing
+    // configured but failed), let the user regenerate or knowingly post
+    // without provenance (#6058).
+    ref.listen(videoEditorProvider.select((s) => s.c2paSigningFailed), (
```
NotThatKindOfDrLiz, line 112: `ref.listen` only catches changes after this build registers the listener. If `c2paSigningFailed` is already true when the metadata screen mounts, the prompt is never shown and the user can post without seeing the provenance warning. That can happen with resumed drafts, and it is also a latent race in capture/lip-sync where render starts before navigation. Please handle the initial state as well, for example with a post-frame read or a listener mode that fires immediately.

mobile/lib/providers/video_publish_provider.dart
```diff
@@ -530,6 +530,11 @@ class VideoPublishNotifier extends Notifier<VideoPublishProviderState> {
         error: error,
         stackTrace: stackTrace,
       );
+      // Clear the `.preparing` guard and surface the failure so the user can
```
NotThatKindOfDrLiz: This clears `.preparing`, but it does not actually surface the failure to the user on this flow. The library overlay only watches `publishState == .preparing`, and I do not see a UI consumer for this `errorMessage`/`.error` state here. The spinner will disappear and retries work, but the failure is silent. Please either wire a visible error/snackbar for this path or adjust the comment/copy so the PR does not claim UX it does not implement.

mobile/lib/providers/video_editor_provider.dart
```diff
@@ -1098,43 +1117,191 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
...
 
-    final (finalRenderedClip, proofManifestJson) = result;
+  /// Whether a rendered clip is missing its C2PA content credential in a build
+  /// where signing is configured — i.e. signing was attempted and failed.
+  ///
+  /// [signingConfigured] is [C2paSigningService.isSigningConfigured]; passed in
+  /// so the decision is pure and testable. Returns false whenever signing isn't
+  /// configured (CI / unconfigured builds), so those never prompt.
+  @visibleForTesting
+  static bool c2paSigningFailedFor({
```
NotThatKindOfDrLiz, line 1222: This helper conflates "C2PA signing failed" with "no usable ProofMode manifest was produced." `NativeProofModeService.proofFile()` returns null for several non-signing failures after `signVideo()` can already have succeeded, including native ProofMode/generateProof/metadata failures. In those cases the video file may have C2PA credentials, but the UI tells the user credentials could not be added and sends them into the re-sign prompt. Please distinguish the signing result from proof-manifest availability, or rename/state the flag and user copy around the broader missing-credential/proof condition.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The timeout/recovery direction is solid, but this still leaves new recovery UX debt: the Regenerate path is not wired to the full regenerate method, the prompt can be missed if the flag is already set at screen mount, the publish failure is stored but not surfaced, and the C2PA failure flag currently measures broader ProofMode/proof-manifest failure.

mobile/lib/services/c2pa_signing_service.dart
```diff
@@ -136,13 +170,16 @@ class C2paSigningService {
...
+      // Sign the file. Bounded so a network hang surfaces as a best-effort
+      // failure instead of wedging the generation (#6058).
+      await _c2pa
+          .signFile(
+            sourcePath: videoPath,
+            destPath: signedPath,
+            manifestJson: manifestResult.manifestJson,
+            signer: await signer,
+          )
+          .timeout(_signingTimeout);
```
realmeylisdev: `Future.timeout()` doesn't cancel the source future — if the native `signFile` call (a plain `MethodChannel.invokeMethod`) is still running when this fires, it keeps running in the background and can still write `signedPath` (line 149) after this method has already returned failure using the original `videoPath`. Nothing in this file, `native_proofmode_service.dart`, or elsewhere ever revisits `signedPath` after that point, so a signing attempt that finishes just past the 20s bound leaves an orphaned `c2pa_signed_<timestamp>.mp4` on disk. Since retries reuse this same flow, repeated timeouts on a bad connection would accumulate these. Worth cleaning up the abandoned file — e.g. a `whenComplete`/`catchError` on the un-awaited native future, or a stat-and-delete pass on the next signing attempt.

mobile/lib/models/video_editor/video_editor_provider_state.dart
```diff
@@ -189,6 +210,13 @@ class VideoEditorProviderState {
   }) {
     return VideoEditorProviderState(
       isProcessing: isProcessing ?? this.isProcessing,
+      // A produced or invalidated clip always ends the failed state.
+      renderFailed:
+          !clearFinalRenderedClip && (renderFailed ?? this.renderFailed),
+      // Invalidating the rendered clip also drops any pending C2PA prompt.
+      c2paSigningFailed:
+          !clearFinalRenderedClip &&
+          (c2paSigningFailed ?? this.c2paSigningFailed),
```
realmeylisdev, line 259: `copyWith(clearFinalRenderedClip: true, renderFailed: true)` (or `c2paSigningFailed: true`) in the same call silently drops the explicit value — `!clearFinalRenderedClip && (...)` short-circuits to `false` regardless of what's passed in, with no assertion or warning. No current call site combines them (`video_editor_provider.dart:191`, `:986`, `:1231` all pass `clearFinalRenderedClip` alone), so this isn't live today, but it's a silent-argument-drop trap for the next caller who combines them.

mobile/lib/screens/video_metadata/video_metadata_screen.dart
```diff
@@ -41,8 +46,57 @@ class _VideoMetadataScreenState extends ConsumerState<VideoMetadataScreen> {
...
+          Navigator.of(context).pop(_C2paMissingChoice.postAnyway),
+      isDismissible: false,
+      enableDrag: false,
+    );
+    if (!mounted) return;
+
+    final notifier = ref.read(videoEditorProvider.notifier);
+    switch (choice) {
+      case _C2paMissingChoice.regenerate:
+      case null:
```
realmeylisdev, line 95: This `case null:` (system-back re-signs instead of silently posting without the credential) is exactly the branch the "review-hardening" commit called out as a deliberate fix, but nothing exercises it — the two new tests in `video_metadata_screen_test.dart` only tap the Regenerate/Post-anyway buttons directly (explicit enum values), never a bare `null` pop (e.g. simulating the back gesture). Worth a regression test, since a future refactor could collapse this into the `postAnyway` case without anything catching it.

realmeylisdev (commented):
Traced the fix mechanism directly: the `.timeout()` on `C2paSigningService.signVideo` correctly bounds the exact call chain both `video_editor_provider.dart` and `video_publish_provider.dart` use for rendering (via `NativeProofModeService.proofFile` → `VideoEditorRenderService.renderVideoToClip`), so the reported hang is fixed. CI-safety and l10n-mirroring hold up too — confirmed `PROOFMODE_SIGNING_SERVER_ENDPOINT`/`DIVINE_FORCE_C2PA_PROMPT` are never set in any GitHub Actions workflow, and all 6 new keys are present in all 18 locale files.

@NotThatKindOfDrLiz's change request still applies as-is against this HEAD (all four threads current, none addressed by the latest commit) — traced each independently and they hold: the Regenerate button really does call `retryC2paSigning()` not `regenerateVideo()`, the `ref.listen` at `video_metadata_screen.dart:91` really does miss an already-true `c2paSigningFailed` at mount, `library_screen.dart` really never renders the publish `.error`/`errorMessage` state, and `c2paSigningFailedFor()` really does conflate C2PA-specific failure with any proof-manifest failure.

Left three additional, non-duplicate notes inline: an orphaned temp file left behind when the new signing timeout actually fires, a silent-argument-drop trap in `copyWith` (latent, no live call site today), and a missing regression test for the system-back path on the C2PA prompt.

mobile/lib/providers/video_editor_provider.dart
hm21: Fixed in a37d99e0f7 — removed the dead `regenerateVideo` (its only path was the unreachable `clip == null` fallback in `retryC2paSigning`, now repointed to `startRenderVideo`) and dropped its test. Kept the re-sign behavior, which is the cheaper/correct recovery for a signing failure. Left the "Regenerate" copy as-is (reads as "regenerate the credential") — happy to rename it to "Try again" across the locales if you'd prefer that.

mobile/lib/screens/video_metadata/video_metadata_screen.dart
hm21: Fixed in a37d99e0f7 — added a post-frame mount check that fires the prompt when `c2paSigningFailed` is already true (capture/lip-sync kicks the render off before navigation, so the value can be true at mount), plus a reentrancy guard so the mount check and `ref.listen` can't stack two sheets. Regression test added.

mobile/lib/providers/video_publish_provider.dart
hm21: Fixed in a37d99e0f7 — this catch is the pre-handoff path (before the background publish bloc receives the process), so it was genuinely silent. It now shows a root-navigator snackbar and the comment is corrected to match. Test added.

mobile/lib/providers/video_editor_provider.dart
hm21: Addressed in a37d99e0f7 — took the "state the flag around the broader condition" option: documented that the flag means "no confirmable content-credential / proof manifest" (a genuine signing failure OR signing-succeeded-but-proof-failed), for which the recovery is identical. I did not do the full distinguish-signing-from-proof refactor — it would thread `C2paSigningResult` through the widely-used `proofFile`, and severity is low here since the false prompt just triggers a re-sign that then succeeds. Can go deeper if you want it.

mobile/lib/services/c2pa_signing_service.dart
hm21: Fixed in a37d99e0f7 — on `TimeoutException` we now attach a cleanup to the abandoned native future that deletes the orphaned `c2pa_signed_*.mp4` once it settles. Regression test added.

mobile/lib/models/video_editor/video_editor_provider_state.dart
hm21: Fixed in a37d99e0f7 — added an `assert` so the `clearFinalRenderedClip` + `renderFailed`/`c2paSigningFailed: true` combination fails loudly instead of silently dropping the value.

mobile/lib/screens/video_metadata/video_metadata_screen.dart
hm21: Fixed in a37d99e0f7 — added a regression test that a bare `null` pop (system back) re-signs rather than posting without provenance, plus one for the mount-time prompt path.

hm21:
@NotThatKindOfDrLiz @realmeylisdev all threads addressed in a37d99e0f7:

- Removed the dead `regenerateVideo` (repointed its fallback to `startRenderVideo`)
- Mount-time C2PA prompt for an already-failed render + reentrancy guard
- Pre-handoff publish failure now surfaced via snackbar
- Orphaned `c2pa_signed_*.mp4` cleanup when the native call finishes past the timeout
- `copyWith` assert against the silent `clearFinalRenderedClip` drop
- Broader-condition doc on the signing-failed flag
- Regression tests for the system-back path, mount-time prompt, timeout cleanup, and publish snackbar

Two open calls left in the threads: kept the "Regenerate" copy (can rename to "Try again" if preferred), and took the doc/framing option on the C2PA-vs-proof conflation rather than the full distinguish-signing-from-proof refactor. `flutter analyze lib test` clean; all affected suites green.

mobile/lib/services/c2pa_signing_service.dart
realmeylisdev: Verified in a37d99e0f7 — the timeout path now cleans up the abandoned native future orphaned `c2pa_signed_*.mp4`, and the timestamped path keeps a retry from colliding with the pending delete. Regression test passes locally. Resolving.

mobile/lib/models/video_editor/video_editor_provider_state.dart
realmeylisdev: Verified in a37d99e0f7 — the assert now fails loudly on `clearFinalRenderedClip` + `renderFailed`/`c2paSigningFailed: true`, so the silent drop cannot slip past a future caller. Resolving.

mobile/lib/screens/video_metadata/video_metadata_screen.dart
realmeylisdev: Verified in a37d99e0f7 — the new test drives a system-back (`handlePopRoute`) and asserts the re-sign path (`isProcessing` true, flag cleared) rather than a silent post. Passes locally. Resolving.

realmeylisdev (approved):
Re-verified my three threads against a37d99e0f7 — orphan cleanup, the copyWith assert, and the system-back test are all in and correct. flutter analyze clean and the c2pa / publish / metadata suites pass locally. Approving from my side; the two open copy/framing calls are Liz's to make.

rabble:
Re-review nudge + one process note.

On the code side this looks addressed: all 7 review threads are resolved, and @realmeylisdev approved against the current head. @NotThatKindOfDrLiz — your four threads (dead `regenerateVideo()` call, the missed already-true `c2paSigningFailed` at mount, the silent publish-failure clear, and the flag conflating C2PA-signing failure with a missing proof-manifest) all have fixes in the current head. This is in CHANGES_REQUESTED only because the formal review predates those pushes — could you re-review and clear/refresh it?

Two things genuinely still need your call rather than more code (per @realmeylisdev's note that they're yours to make):
- whether to rename the "Regenerate" button copy to "Try again", and
- whether documenting the flag as the broader "no confirmable content-credential" condition is acceptable, vs. a full signing-vs-proof distinction refactor.

Separately, this needs a rebase — GitHub reports it CONFLICTING against main (mostly the ARB + generated l10n collision from another merged PR, plus a few real source conflicts). @hm21 could you rebase onto fresh origin/main so it's mergeable?

NotThatKindOfDrLiz (approved):
Re-reviewed at f3412b1. The previously requested changes are resolved, targeted recovery tests pass locally, and CI is green. Approving.

PR #6069, opened 2026-07-13
fix(dm): classify Keycast verified_minor policy-denial (403) as terminal, not retryable

### LINKED ISSUE #6067: fix(dm): classify Keycast verified_minor policy-denial (403) as terminal, not retryable
Part of divinevideo/support-trust-safety#173. Surfaced while reviewing #6046 (the DM-durability rewrite). Low priority, no safety impact, and latent until Keycast's verified_minor DM gate (divinevideo/keycast#284) is on prod.

## The gap
Once Keycast's server-side verified_minor DM containment is live, a protected minor's attempt to DM a non-approved recipient is refused at the signer, and Keycast returns HTTP 403 with the body marker "Operation denied by policy". On the client that surfaces as `RpcException('HTTP 403: ...')` thrown during wrap-build inside `NIP17MessageService.sendRumor`, and it lands in the generic `on Object catch`, which returns `NIP17SendResult.failure(...)` (retryable). So the drain re-attempts on backoff until `maxRetries`, and manual Resend re-arms and re-fails.

## Why it's not a safety issue
Keycast refuses to sign or encrypt, so no gift wrap is produced and nothing is published. Nothing reaches the non-approved recipient on any attempt. This is a reliability/UX wart (wasted retry budget, a Resend button that deterministically re-fails), not a leak. It is the server-refusal sibling of divinevideo/support-trust-safety#186, which handled the client-predicted block and already shipped in #6028.

## When it happens
Only after keycast #284 is promoted to prod. Before that, the client-side gate blocks these sends locally, so the 403 path is unreachable in production.

## Fix
Classify the policy-denial refusal as terminal and return `NIP17SendResult.blocked(...)` so the existing `blocked` to terminal drain path (from #6028) drops the row instead of retrying.
- Site: the send-method catch in `mobile/packages/dm_repository/lib/src/nip17_message_service.dart`, and confirm the same for the batch and reaction send paths.
- Match on the specific marker "Operation denied by policy", not any 403. Keycast returns that string only for the verified_minor gate; suspended accounts return "Account restricted", and token issues are 401 with their own refresh path, so keying on the marker keeps transient failures retryable.
- Include a test. dm_repository coverage sits right at its gate.

## Design note (my call when I pick it up)
The quick version couples the client to a Keycast error string. The cleaner version is Keycast returning a structured error code the client switches on, which spans keycast and mobile. I lean string-match with a clear comment now, structured code later if it earns its keep.

## Sequencing
Depends on #6046 landing. Do this after it merges, so it builds on the final send/retry classification rather than conflicting with that rewrite. Aim to ship it in an app release around the keycast #284 prod promotion, so the client handles the refusal cleanly from day one. Not gating, since there is no leak either way.

mbradley:
## Summary

Classifies a Keycast verified_minor policy-denial (HTTP 403 "Operation denied by policy") as a terminal `blocked` send in `sendRumor`, instead of the generic retryable failure, and wires the terminal handling into the one send arm that lacked it.

## Motivation

Once Keycast's server-side verified_minor DM containment (divinevideo/keycast#284) is on prod, a protected minor's DM to a non-approved recipient is refused at the signer with a 403. Today that lands in the generic catch as a retryable failure, so the drain re-attempts until maxRetries and Resend deterministically re-fails. No leak (Keycast refuses to sign, nothing publishes), but it wastes the retry budget and leaves a doomed Resend button. This is the server-refusal sibling of the client-predicted block handled in #6028.

`sendRumor` is the single recipient-delivering choke point, so classifying there covers direct, group fan-out, drain replay, and reaction sends. Keying on the specific marker (not any 403) keeps transient failures retryable: suspended accounts return "Account restricted" and token issues are 401 with their own refresh path.

The group, drain, and reaction arms already terminalize a `blocked` result (drop the queue row). The 1:1 `sendMessage` live path did not — a post-enqueue block fell into the failed/retryable branch, leaving a doomed red Resend bubble the sweep re-drives once before dropping it. This adds the terminal `blocked` arm to `sendMessage`, mirroring `sendGroupMessage`.

## Related issue

Closes #6067. Part of divinevideo/support-trust-safety#173.

## Testing

- New tests: a 403 terminalizes as `blocked` with no publish; a non-policy error stays a plain failure; a 1:1 `sendMessage` that 403s after enqueue drops the row instead of marking it failed. The classification tests throw an `RpcException`-shaped double reproducing keycast_flutter's toString, so the marker is tested against the real Keycast 403 surface.
- dm_repository suite: 510 pass; coverage 93.56% (gate 93). analyze + format clean.

## Visuals

No visual change.

## Notes

Rebased onto #6046 after it merged. The marker string match is the quick version; a structured Keycast error code is the cleaner long-term option, noted in #6067.

### DIFF
--- a/mobile/packages/dm_repository/lib/src/dm_repository.dart
+++ b/mobile/packages/dm_repository/lib/src/dm_repository.dart
@@ -2876,6 +2876,15 @@ class DmRepository {
         // Don't rethrow — the message was published successfully.
         // Local persistence failure is a degraded state, not a send failure.
       }
+    } else if (result.blocked && outgoingDao != null) {
+      // A policy block is terminal, not a transient failure: drop the queue
+      // row so the sweep stops re-driving a send the gate refuses every time
+      // (no doomed Resend bubble). Mirrors sendGroupMessage's per-recipient
+      // classification and the drain's blocked arm.
+      await _finalizeAfterRecipientBlocked(
+        outgoingDao: outgoingDao,
+        rumorId: rumor.id,
+      );
     } else if (result.retryablePending && outgoingDao != null) {
       // Soft-unconfirmed: the recipient frame was written but no NIP-20 OK
       // arrived (and no explicit rejection). Keep the row pending — it
@@ -2900,17 +2909,15 @@ class DmRepository {
         rumorId: rumor.id,
         errorMessage: result.error ?? 'Unknown publish failure',
       );
-      // A blocked send never enqueues a bubble, but a failed/pending one
-      // did — make sure a brand-new thread is visible so the user can find
-      // the red bubble to retry or delete it.
-      if (!result.blocked) {
-        await _ensureConversationVisibleAfterSendFailure(
-          conversationId: conversationId,
-          participants: participants,
-          isGroup: false,
-          content: content,
-        );
-      }
+      // Make a brand-new thread visible so the user can find the red bubble
+      // to retry or delete it. A blocked result is handled terminally above
+      // and never reaches here.
+      await _ensureConversationVisibleAfterSendFailure(
+        conversationId: conversationId,
+        participants: participants,
+        isGroup: false,
+        content: content,
+      );
     }
 
     return result;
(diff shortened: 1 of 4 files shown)


dcadenas (approved):
🔒 verified-minor denials now stop retrying

realmeylisdev (approved):
lgtm. went through the keycast side and the 403 does come back as blocked, so it stops retrying like it should. tests pass 👍

PR #6070, opened 2026-07-13
fix(blurhash): correct color-tinted decode and reduce DCT ringing

### LINKED ISSUE #6071: fix(blurhash): correct color-tinted decode and reduce DCT ringing
Auto-created from PR #6070.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6070
**Branch:** `fix/blurhash-decode-tint-and-ringing`

hm21:
Closes #6071

## Description

Blurhash placeholders looked broken, most visibly on 9:16 videos: grayscale/low-chroma content rendered with **navy-blue shadows and a cream cast**, and high-contrast frames showed dark/bright blobs that don't exist in the source.

Two root causes, both verified empirically against a spec-canonical reference implementation (our encoded hashes were byte-identical to canonical full-res reference encodes, which isolated the decode as the visual culprit):

1. **`blurhash_dart`'s `decodeAc` misses the spec's integer division** (`value / (19*19)` instead of `value ~/ (19*19)`). The fractional remainder systematically inflates red/green in every AC component by up to ~23% of maxAc while blue stays exact. A white→black gradient decodes to `r=0 g=0 b=38` at the dark end — that's the blue tint. Fixed by extracting the DCT components ourselves (spec-correct) and reusing only the library's iDCT (`toImage`). This also fixes rendering of already-published hashes from other clients. The library's `punch` parameter is likewise broken (skips the first AC row/column), so punch is applied uniformly via maxAc in our extraction.
2. **Too many DCT components at encode time** (4×7 = 28 for portrait) cause Gibbs ringing — overshoot at contrast edges clamps to black/white blobs. Reduced to 3×4 portrait / 3×3 square (blurhash-recommended range; Divine only produces 9:16 and 1:1 videos). The pre-encode downscale now uses `average` instead of the default `nearest` interpolation, which sampled stray pixels on detailed frames.

## Performance

Fewer DCT components are also a decode-time win. `decodeBlurhash` runs the inverse-DCT synchronously on the UI thread (from `BlurhashDisplay.initState`/`didUpdateWidget`, once per hash then memoized), and that loop scales with component count. Portrait decodes now do ~57% fewer `cos` evaluations (28→12 components, ~57k→~25k per 32×32 decode) and square ~44% (16→9). The new hand-rolled `_decodeComponents` + `BlurHash.components(...).toImage()` path adds only negligible per-cache-miss overhead (a discarded re-encode, `List.generate`, `indexOf` base83 decode — each well under 1% of the iDCT), and the `average` downscale runs off the UI thread in the encode isolate. Net: the decode hot path got cheaper, not more expensive.

Decode punch defaults to 0.8 so legacy high-component hashes (already published in Nostr events, still 4×7) render with softened ringing too. Encode stays fully spec-compatible — web clients decode the same hashes canonically.

**Related Issue:** 

## Out of Scope

- Upstream bug report to `blurhash_dart` (decodeAc integer division + punch skipping first AC row/column).
- The inherent blurhash brightness bias on high-dynamic-range frames (linear-space averaging) — canonical behavior, all clients share it.

## Verification

- `flutter test` in `mobile/packages/blurhash_service` — 40/40, including a new regression test asserting a grayscale gradient decodes without color tint (fails with the library decode: channel diff 38).
- `flutter analyze packages/blurhash_service` — clean.
- Manual side-by-side comparison in-app (picsum images vs. decoded blurhash) across 9:16/1:1 ratios via a temporary debug screen (not part of this PR).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
+++ b/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
@@ -3,6 +3,7 @@
 // ABOUTME: Creates compact representations of images
 // for better UX during vine loading
 
+import 'dart:math' as math;
 import 'dart:ui' as ui;
 
 import 'package:blurhash_dart/blurhash_dart.dart' as blurhash_dart;
@@ -22,8 +23,15 @@ String? _generateBlurhashSync(Uint8List imageBytes) {
     image.height,
   );
 
+  // Average (box) interpolation: the default (nearest) picks single source
+  // pixels when downscaling, which skews the encoded colors on high-detail
+  // frames.
   final resized = image.width > BlurhashService._encodeMaxWidth
-      ? img.copyResize(image, width: BlurhashService._encodeMaxWidth)
+      ? img.copyResize(
+          image,
+          width: BlurhashService._encodeMaxWidth,
+          interpolation: img.Interpolation.average,
+        )
       : image;
 
   return blurhash_dart.BlurHash.encode(
@@ -36,19 +44,22 @@ String? _generateBlurhashSync(Uint8List imageBytes) {
 /// Service for generating and decoding Blurhash
 /// placeholders.
 class BlurhashService {
-  /// Components for 9:16 portrait videos (4:7 ≈ 0.57, matches 9:16 ≈ 0.5625).
-  static const int _portraitComponentX = 4;
-  static const int _portraitComponentY = 7;
+  /// Components for 9:16 portrait videos. Kept low on purpose: high counts
+  /// (the previous 4×7) cause visible DCT ringing — dark/bright blobs that
+  /// aren't in the source frame.
+  static const int _portraitComponentX = 3;
+  static const int _portraitComponentY = 4;
 
-  /// Components for 1:1 square and landscape videos (balanced).
-  static const int _squareComponentX = 4;
-  static const int _squareComponentY = 4;
+  /// Components for 1:1 square videos.
+  static const int _squareComponentX = 3;
+  static const int _squareComponentY = 3;
 
   /// Max width used when downscaling before encoding.
   static const int _encodeMaxWidth = 128;
 
-  /// Default punch (contrast) value.
-  static const double defaultPunch = 1;
+  /// Default punch (contrast) value. Below 1 to soften DCT overshoot from
+  /// already-published high-component hashes (see [decodeBlurhash]).
+  static const double defaultPunch = 0.8;
 
   /// Process-wide memo for [decodeBlurhash]. Decoding is a pure function
   /// of (hash, width, height, punch) and runs synchronously on the UI
@@ -63,7 +74,8 @@ class BlurhashService {
   );
 
   /// Returns component counts suited to the image's aspect ratio.
-  /// Supports 9:16 portrait, 1:1 square, and landscape; falls back to portrait.
+  /// Divine only produces 9:16 portrait and 1:1 square videos; anything
+  /// square-or-wider gets square components, tall formats get portrait.
   static (int compX, int compY) _componentsForAspectRatio(
     int width,
     int height,
@@ -72,10 +84,77 @@ class BlurhashService {
       return (_portraitComponentX, _portraitComponentY);
     }
     final ratio = width / height;
-    // Square or landscape: ratio ≥ 0.9 (covers 1:1 and any wider format)
+    // 1:1 square (and any wider format)
     if (ratio >= 0.9) return (_squareComponentX, _squareComponentY);
     // Portrait 9:16 (and any other tall format)
     return (_portraitComponentX, _portraitComponentY);
+  }
+
+  /// Base83 alphabet from the blurhash spec.
+  static const String _base83Chars =
+      '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+      'abcdefghijklmnopqrstuvwxyz'
+      r'#$%*+,-.:;=?@[]^_{|}~';
+
+  static int _decode83(String value, int from, int to) {
+    var result = 0;
+    for (var i = from; i < to; i++) {
+      result = result * 83 + _base83Chars.indexOf(value[i]);
+    }
+    return result;
+  }
+
+  static double _sRgbToLinear(int value) {
+    final v = value / 255.0;
+    if (v <= 0.04045) return v / 12.92;
+    return math.pow((v + 0.055) / 1.055, 2.4).toDouble();
+  }
+
+  static double _signPow(double value, double exp) =>
+      math.pow(value.abs(), exp) * value.sign;
+
+  /// Spec-correct extraction of the DCT components from [blurhash].
+  ///
+  /// blurhash_dart's decodeAc divides with `/` where the spec requires
+  /// integer division, so the quantised red/green values pick up a
+  /// fractional positive shift (up to ~23% of maxAc) while blue stays
+  /// exact — dark regions of grayscale content decode as navy blue.
+  ///
+  /// [punch] scales every AC component uniformly (baked into maxAc).
+  /// Returns `null` when the hash length doesn't match its size flag.
+  static List<List<blurhash_dart.ColorTriplet>>? _decodeComponents(
+    String blurhash, {
+    double punch = 1.0,
+  }) {
+    final sizeFlag = _decode83(blurhash, 0, 1);
+    final numCompX = (sizeFlag % 9) + 1;
+    final numCompY = (sizeFlag ~/ 9) + 1;
+    if (blurhash.length != 4 + 2 * numCompX * numCompY) return null;
+
+    final maxAc = (_decode83(blurhash, 1, 2) + 1) / 166.0 * punch;
+    final dcValue = _decode83(blurhash, 2, 6);
+
+    return List.generate(numCompY, (j) {
+      return List.generate(numCompX, (i) {
+        if (i == 0 && j == 0) {
+          return blurhash_dart.ColorTriplet(
+            _sRgbToLinear(dcValue >> 16),
+            _sRgbToLinear((dcValue >> 8) & 255),
+            _sRgbToLinear(dcValue & 255),
+          );
+        }
+        final index = i + j * numCompX;
+        final value = _decode83(blurhash, 4 + index * 2, 6 + index * 2);
+        final quantR = value ~/ (19 * 19);
+        final quantG = (value ~/ 19) % 19;
+        final quantB = value % 19;
+        return blurhash_dart.ColorTriplet(
+          _signPow((quantR - 9) / 9, 2) * maxAc,
+          _signPow((quantG - 9) / 9, 2) * maxAc,
+          _signPow((quantB - 9) / 9, 2) * maxAc,
+        );
+      });
+    });
   }
 
   /// Generate blurhash from image bytes.
@@ -148,12 +227,16 @@ class BlurhashService {
         return cached;
       }
 
-      // Use real blurhash_dart library to decode
-      final blurHashObject = blurhash_dart.BlurHash.decode(
-        blurhash,
-        punch: punch,
-      );
-      final image = blurHashObject.toImage(width, height);
+      // Extract the DCT components ourselves instead of using
+      // blurhash_dart's BlurHash.decode: its decodeAc misses the spec's
+      // integer division, inflating red/green in every AC component —
+      // grayscale content decodes with a blue/cream tint. Its `punch` is
+      // also broken (skips the first AC row and column).
+      final components = _decodeComponents(blurhash, punch: punch);
+      if (components == null) return null;
+      final image = blurhash_dart.BlurHash.components(
+        components,
+      ).toImage(width, height);
 
       // RGBA pixel data of the decoded image (getBytes already returns a
       // buffer we own — no defensive copy needed)
--- a/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
+++ b/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
@@ -57,8 +57,10 @@ class BlurhashService {
   /// Max width used when downscaling before encoding.
   static const int _encodeMaxWidth = 128;
 
-  /// Default punch (contrast) value. Below 1 to soften DCT overshoot from
-  /// already-published high-component hashes (see [decodeBlurhash]).
+  /// Default punch (contrast) applied to every decoded hash. Kept below 1
+  /// on purpose: the softer look reads better in the feed and it also damps
+  /// the DCT overshoot high-component legacy hashes produce. This is an
+  /// intentional product choice for all placeholders, not only legacy ones.
   static const double defaultPunch = 0.8;
 
   /// Process-wide memo for [decodeBlurhash]. Decoding is a pure function
@@ -96,10 +98,20 @@ class BlurhashService {
       'abcdefghijklmnopqrstuvwxyz'
       r'#$%*+,-.:;=?@[]^_{|}~';
 
+  /// Reverse index of [_base83Chars] for O(1) character lookup (matching
+  /// blurhash_dart's own decode83). [decodeBlurhash] runs synchronously on
+  /// the UI isolate on every cache miss, so a linear `indexOf` scan per
+  /// character would add up.
+  static final Map<String, int> _base83Index = {
+    for (var i = 0; i < _base83Chars.length; i++) _base83Chars[i]: i,
+  };
+
   static int _decode83(String value, int from, int to) {
     var result = 0;
     for (var i = from; i < to; i++) {
-      result = result * 83 + _base83Chars.indexOf(value[i]);
+      // Every char is validated against the base83 alphabet before decode,
+      // so the lookup never misses; `?? 0` only satisfies null-safety.
+      result = result * 83 + (_base83Index[value[i]] ?? 0);
     }
     return result;
   }
--- a/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
+++ b/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
@@ -141,7 +141,16 @@ class BlurhashService {
     final sizeFlag = _decode83(blurhash, 0, 1);
     final numCompX = (sizeFlag % 9) + 1;
     final numCompY = (sizeFlag ~/ 9) + 1;
-    if (blurhash.length != 4 + 2 * numCompX * numCompY) return null;
+    final expectedLength = 4 + 2 * numCompX * numCompY;
+    if (blurhash.length != expectedLength) {
+      Log.error(
+        'Malformed blurhash: length ${blurhash.length} does not match the '
+        'declared ${numCompX}x$numCompY components (expected $expectedLength)',
+        name: 'BlurhashService',
+        category: LogCategory.system,
+      );
+      return null;
+    }
 
     final maxAc = (_decode83(blurhash, 1, 2) + 1) / 166.0 * punch;
     final dcValue = _decode83(blurhash, 2, 6);
--- a/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
+++ b/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
@@ -234,6 +234,9 @@ class BlurhashService {
       // also broken (skips the first AC row and column).
       final components = _decodeComponents(blurhash, punch: punch);
       if (components == null) return null;
+      // BlurHash.components re-encodes the components into a hash string we
+      // discard, but it's the only public components→image path and ~0.1%
+      // of the iDCT in toImage below — not worth reimplementing to avoid.
       final image = blurhash_dart.BlurHash.components(
         components,
       ).toImage(width, height);
--- a/mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
+++ b/mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
@@ -6,6 +6,7 @@ import 'package:blurhash_dart/blurhash_dart.dart' as blurhash_dart;
 import 'package:blurhash_service/blurhash_service.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:image/image.dart' as img;
+import 'package:unified_logger/unified_logger.dart';
 
 /// Creates a solid-colour JPEG with the given [width] and [height].
 Uint8List _makeJpeg({required int width, required int height}) {
@@ -98,6 +99,32 @@ void main() {
       expect(BlurhashService.decodeBlurhash(''), isNull);
       expect(BlurhashService.decodeBlurhash('short'), isNull);
     });
+
+    test(
+      'logs an error when a valid-charset hash has a mismatched length',
+      () async {
+        await LogCaptureService().clearAllLogs();
+
+        // A real hash truncated mid-string: passes the charset/length-floor
+        // check but its length no longer matches its declared component count,
+        // so decoding bails. Corrupt/truncated tags must stay observable.
+        const truncatedHash = 'L6Pj0^jE.AyE_3t7t7R*';
+
+        expect(BlurhashService.decodeBlurhash(truncatedHash), isNull);
+
+        final errors = LogCaptureService()
+            .getRecentLogs(minLevel: LogLevel.error)
+            .where((entry) => entry.name == 'BlurhashService')
+            .toList();
+        expect(
+          errors,
+          isNotEmpty,
+          reason:
+              'a malformed blurhash must emit an error log, not fail '
+              'silently',
+        );
+      },
+    );
 
     test('blurhash data provides gradient', () {
       final testBlurhash = BlurhashService.getDefaultVineBlurhash();
--- a/mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
+++ b/mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
@@ -146,6 +146,84 @@ void main() {
       }
     });
 
+    test('decodes a solid fill back to its source color', () async {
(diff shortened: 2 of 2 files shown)


mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -36,19 +44,22 @@ String? _generateBlurhashSync(Uint8List imageBytes) {
...
+  static const int _squareComponentY = 3;
 
   /// Max width used when downscaling before encoding.
   static const int _encodeMaxWidth = 128;
 
-  /// Default punch (contrast) value.
-  static const double defaultPunch = 1;
+  /// Default punch (contrast) value. Below 1 to soften DCT overshoot from
+  /// already-published high-component hashes (see [decodeBlurhash]).
+  static const double defaultPunch = 0.8;
```
NotThatKindOfDrLiz, line 64: This changes the default contrast for every decoded hash, including newly generated 3x4/3x3 hashes. The comment frames `0.8` as compensation for already-published high-component hashes, but this now permanently flattens new hashes too. Please either derive the default punch from the parsed component count (for example, keep `1.0` for new low-component hashes and only soften legacy high-component hashes) or explicitly justify that the lower contrast is the intended product behavior for all placeholders going forward.

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
```diff
@@ -114,23 +114,64 @@ void main() {
       expect(data!.isValid, isTrue);
     });
 
+    test('decodes grayscale content without a color tint', () async {
```
NotThatKindOfDrLiz, line 145: This regression test proves the old tint bug is gone, but it does not pin the spec-correct decoder output or the punch behavior that the library got wrong. Since this PR hand-rolls DCT component extraction, please add one canonical reference-vector test with expected RGB values at a few sample pixels, plus an explicit `punch` test proving AC components in the first row/column are scaled too.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -148,12 +227,19 @@ class BlurhashService {
         return cached;
       }
 
-      // Use real blurhash_dart library to decode
-      final blurHashObject = blurhash_dart.BlurHash.decode(
-        blurhash,
-        punch: punch,
-      );
-      final image = blurHashObject.toImage(width, height);
+      // Extract the DCT components ourselves instead of using
```
NotThatKindOfDrLiz, line 251: This is now an in-repo workaround for upstream `blurhash_dart` decode bugs. Per the repo's temporary-code/workaround policy, please file/link the upstream bug and add a tracking issue/removal plan so this custom decoder can be dropped if the dependency is fixed.

NotThatKindOfDrLiz (changes requested):
Requesting changes on the product-visible default punch change and the test/removal-plan gaps around the custom decoder. The decoder fix itself looks directionally correct, and the affected package tests pass locally.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -72,12 +84,79 @@ class BlurhashService {
...
+  /// Base83 alphabet from the blurhash spec.
+  static const String _base83Chars =
+      '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+      'abcdefghijklmnopqrstuvwxyz'
+      r'#$%*+,-.:;=?@[]^_{|}~';
+
+  static int _decode83(String value, int from, int to) {
+    var result = 0;
+    for (var i = from; i < to; i++) {
+      result = result * 83 + _base83Chars.indexOf(value[i]);
```
realmeylisdev: Minor: `_base83Chars.indexOf(value[i])` is an O(83) linear scan per character, vs. blurhash_dart's own `decode83` which uses a `Map<String, int>` for O(1) lookup. This decode runs synchronously on the UI thread on every cache miss (`BlurhashDisplay.initState`), so a static lookup map here would keep the same complexity as the code it replaces — though it's negligible next to the DCT savings the rest of the PR delivers, so non-blocking.

realmeylisdev (commented):
Independently verified the root cause and fix against blurhash_dart 1.2.1's actual source: `decodeAc` uses `/` where the spec requires `~/`, and `_multiplyPunch`'s `i != 0 && j != 0` guard skips punch on the entire first AC row and column. Both match the PR description exactly.

Empirically confirmed the regression test is meaningful: reverted just the decode path to the pre-fix `blurhash_dart.BlurHash.decode()` call and reran the new grayscale test in isolation — it fails with a real cream-tinted pixel (r=251 g=250 b=228), matching the described tint. Restored the fix and reran the full suite: `flutter test` 40/40, `flutter analyze` clean, `dart format` clean. Also ran the app-level widget tests that consume `BlurhashDisplay`/`VideoThumbnailWidget` (63 tests) — all green, and confirmed the one golden test file touching this widget already has its scenarios skipped pre-existing, unrelated to this diff.

Checked cross-repo/protocol impact: funnelcake, the relay, and divine-web all treat `blurhash` as an opaque string with no length/component-count assumptions; NIP-71/NIP-92 don't constrain the algorithm either. divine-web's independent JS decoder derives component counts from the hash's own header byte, so the encode-side component-count change (4×7/4×4 → 3×4/3×3) is fully interop-safe.

Concur with @NotThatKindOfDrLiz's outstanding change requests below — verified each independently, replied inline with supporting detail. One additional non-blocking nit inline. No other correctness issues found.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
realmeylisdev: Confirmed this is real: divine-web's independent JS decoder defaults punch to 1.0 (`BlurhashImage.tsx`), so the same hash now renders with visibly different contrast on mobile vs. web. Since the hash's own size-flag byte already encodes numCompX/numCompY, branching the default on decoded component count (e.g., total components above the new 12/9 ceiling gets 0.8, else 1.0) looks like a clean way to keep the legacy-hash softening without flattening every newly-generated placeholder.

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
realmeylisdev: Verified the tint test does catch the regression: reverted just the decode path to the pre-fix `blurhash_dart.BlurHash.decode()` call and reran it in isolation — it fails with a real cream-tinted pixel (r=251 g=250 b=228 at the bright end), consistent with the AC inflation described in the commit message. Agree it doesn't pin exact spec values or prove punch is applied to every AC term rather than skipping the first row/column like the library does — a small fixed-hash reference-vector test plus one asserting punch scales a first-row/first-column AC component would close that gap.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
realmeylisdev: Checked both justacid/blurhash-dart (upstream) and this repo's issues — no existing tracking issue for either the decodeAc division bug or the punch row/column skip, so this needs to be filed net-new, not just linked. Also worth a look: flutter_blurhash (fluttercommunity) already does spec-correct integer division and folds punch into maxVal for every AC term in its own decode. Swapping the dependency instead of hand-maintaining ~120 lines of duplicated decode logic here could be a cleaner way to close this out than a removal-plan TODO — worth weighing against the migration cost of the encode/decode call sites.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
hm21: Went with your second option — 0.8 is the intended contrast for **all** placeholders, not just legacy compensation. I compared the component-count-derived full-contrast decode (punch 1) against 0.8 on-device, and the softer look reads clearly better in the feed. So I kept the uniform 0.8 and reworded the comment to document it as the deliberate product choice for every placeholder rather than framing it as legacy-only. a7ddeb425f

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
hm21: Added two tests: a solid-fill reference-vector test (source color = expected RGB on every channel, independent of our own decoder) and a punch-scaling test on a horizontal gradient that proves AC in the first component row — the row blurhash_dart's punch skipped — is scaled. a7ddeb425f

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
hm21: Leaving this as-is: it's the spec-correct decode, not transitional migration code with a removal date. The bugs it routes around (decodeAc integer division, punch row/column skip) are documented in the comment above it. If blurhash_dart fixes them upstream we can revisit, but I'd rather not pin a removal TODO on a third-party fix that may never land.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
hm21: Done — swapped to a static `Map<String, int>` for O(1) lookup, matching blurhash_dart's decode83. a7ddeb425f

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -148,12 +239,19 @@ class BlurhashService {
...
+      // blurhash_dart's BlurHash.decode: its decodeAc misses the spec's
+      // integer division, inflating red/green in every AC component —
+      // grayscale content decodes with a blue/cream tint. Its `punch` is
+      // also broken (skips the first AC row and column).
+      final components = _decodeComponents(blurhash, punch: punch);
+      if (components == null) return null;
+      // BlurHash.components re-encodes the components into a hash string we
+      // discard, but it's the only public components→image path and ~0.1%
+      // of the iDCT in toImage below — not worth reimplementing to avoid.
+      final image = blurhash_dart.BlurHash.components(
```
realmeylisdev, line 261: Efficiency (not blocking): `.toImage()` here recomputes `cos()` for every pixel×component with no basis-table caching (~110x redundant calls at the default 32×32 decode) — pre-existing blurhash_dart cost, not a regression from this PR. Also, `BlurHash.components()` re-encodes the components back into a hash string that's immediately discarded (the encode83/pow work is thrown away), since it's the only public components→image path. Both would go away if the vendoring you started for `_decode83`/`_sRgbToLinear`/`_signPow` gets extended to `_transform`/`_encodeComponents`.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -148,12 +239,19 @@ class BlurhashService {
...
-        blurhash,
-        punch: punch,
-      );
-      final image = blurHashObject.toImage(width, height);
+      // Extract the DCT components ourselves instead of using
+      // blurhash_dart's BlurHash.decode: its decodeAc misses the spec's
+      // integer division, inflating red/green in every AC component —
+      // grayscale content decodes with a blue/cream tint. Its `punch` is
+      // also broken (skips the first AC row and column).
+      final components = _decodeComponents(blurhash, punch: punch);
```
realmeylisdev, line 256: The cache is keyed on (hash, width, height, punch), but `_decodeComponents` only actually depends on (hash, punch) — a second caller requesting a different size for the same hash recomputes components from scratch. Latent today since there's only one call site, but worth a note for future callers (e.g. a thumbnail + full-screen view of the same hash).

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -72,12 +86,89 @@ class BlurhashService {
...
+  /// [punch] scales every AC component uniformly (baked into maxAc).
+  /// Returns `null` when the hash length doesn't match its size flag.
+  static List<List<blurhash_dart.ColorTriplet>>? _decodeComponents(
+    String blurhash, {
+    double punch = 1.0,
+  }) {
+    final sizeFlag = _decode83(blurhash, 0, 1);
+    final numCompX = (sizeFlag % 9) + 1;
+    final numCompY = (sizeFlag ~/ 9) + 1;
+    if (blurhash.length != 4 + 2 * numCompX * numCompY) return null;
```
realmeylisdev: Before this PR, a hash whose length didn't match its size-flag byte threw inside blurhash_dart and got logged via `Log.error`. Now `_decodeComponents` just returns `null` silently here — same functional outcome (placeholder falls back either way), but the log signal for a corrupted/truncated blurhash tag from another Nostr client is gone.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -72,12 +86,89 @@ class BlurhashService {
...
+  static final Map<String, int> _base83Index = {
+    for (var i = 0; i < _base83Chars.length; i++) _base83Chars[i]: i,
+  };
+
+  static int _decode83(String value, int from, int to) {
+    var result = 0;
+    for (var i = from; i < to; i++) {
+      // Every char is validated against the base83 alphabet before decode,
+      // so the lookup never misses; `?? 0` only satisfies null-safety.
+      result = result * 83 + (_base83Index[value[i]] ?? 0);
```
realmeylisdev, line 114: `?? 0` silently substitutes for any unrecognized base83 char instead of failing, relying entirely on `_isValidBlurhash`'s regex matching `_base83Chars` exactly (confirmed they do, today). It's a soft coupling between two separately-maintained literals rather than a structural guarantee — low risk, but worth a comment flagging the dependency for future editors.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -72,12 +86,89 @@ class BlurhashService {
...
+    return result;
+  }
+
+  static double _sRgbToLinear(int value) {
+    final v = value / 255.0;
+    if (v <= 0.04045) return v / 12.92;
+    return math.pow((v + 0.055) / 1.055, 2.4).toDouble();
+  }
+
+  static double _signPow(double value, double exp) =>
```
realmeylisdev, line 125: All 3 call sites pass the literal exponent `2` — `_signPow(x, 2)` is just `x * x.abs()`. Could simplify to drop the `math.pow` call on every AC term, though not a bug (the general form mirrors blurhash_dart's own `signPow`, which needs a variable exponent for encode too — this file only reimplements decode).

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -148,12 +239,19 @@ class BlurhashService {
         return cached;
       }
 
-      // Use real blurhash_dart library to decode
-      final blurHashObject = blurhash_dart.BlurHash.decode(
-        blurhash,
-        punch: punch,
-      );
-      final image = blurHashObject.toImage(width, height);
+      // Extract the DCT components ourselves instead of using
```
realmeylisdev, line 251: The decodeAc root cause is explained here and again in `_decodeComponents`'s doc comment above, and the two aren't fully symmetric — the punch first-row/column skip detail only lives here. Minor duplication risk if one gets updated without the other.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -72,12 +86,89 @@ class BlurhashService {
...
+  static const String _base83Chars =
+      '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+      'abcdefghijklmnopqrstuvwxyz'
+      r'#$%*+,-.:;=?@[]^_{|}~';
+
+  /// Reverse index of [_base83Chars] for O(1) character lookup (matching
+  /// blurhash_dart's own decode83). [decodeBlurhash] runs synchronously on
+  /// the UI isolate on every cache miss, so a linear `indexOf` scan per
+  /// character would add up.
+  static final Map<String, int> _base83Index = {
```
realmeylisdev, line 105: `_base83Index` (and the pre-existing `_decodeCache`) assume a single process-wide instance, which only holds because `decodeBlurhash`'s one caller today is synchronous on the UI isolate. Not an issue now, but a future `compute()`-dispatched caller would silently get its own isolate-local copies.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -72,12 +86,89 @@ class BlurhashService {
...
+  /// Spec-correct extraction of the DCT components from [blurhash].
+  ///
+  /// blurhash_dart's decodeAc divides with `/` where the spec requires
+  /// integer division, so the quantised red/green values pick up a
+  /// fractional positive shift (up to ~23% of maxAc) while blue stays
+  /// exact — dark regions of grayscale content decode as navy blue.
+  ///
+  /// [punch] scales every AC component uniformly (baked into maxAc).
+  /// Returns `null` when the hash length doesn't match its size flag.
+  static List<List<blurhash_dart.ColorTriplet>>? _decodeComponents(
```
realmeylisdev, line 137: Worth flagging: blurhash_dart's `BlurHash.decode` (with both bugs) stays fully importable elsewhere in the monorepo since only this call site was patched, not the dependency itself — it's also a direct app-level pubspec dependency. Not asking for a vendor in this PR, just noting the risk if another screen/package ever imports blurhash_dart directly instead of routing through this service.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -36,19 +44,24 @@ String? _generateBlurhashSync(Uint8List imageBytes) {
...
   /// Max width used when downscaling before encoding.
   static const int _encodeMaxWidth = 128;
 
-  /// Default punch (contrast) value.
-  static const double defaultPunch = 1;
+  /// Default punch (contrast) applied to every decoded hash. Kept below 1
+  /// on purpose: the softer look reads better in the feed and it also damps
+  /// the DCT overshoot high-component legacy hashes produce. This is an
+  /// intentional product choice for all placeholders, not only legacy ones.
+  static const double defaultPunch = 0.8;
```
realmeylisdev, line 64: `defaultPunch = 0.8` does double duty — general softening and legacy-hash ringing suppression — even though `_decodeComponents` already knows the real component count at the point punch is applied. Might be worth scaling punch by component count in a follow-up, since new low-ringing hashes get the same flattening as legacy high-ringing ones. Not blocking given the comment above already discloses this as intentional.

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
```diff
@@ -114,23 +114,142 @@ void main() {
...
```
realmeylisdev, line 145: None of the new regression tests (or any hardcoded hash in this file) exercises an actual legacy-shaped high-component hash (the old 4×7/4×4 format) under the new `defaultPunch`. Those hashes are still live in production — `decodeBlurhash` is always called with no punch override, and edits carry the original blurhash forward verbatim — so the PR's stated goal of softening ringing on already-published hashes isn't actually covered by a test.

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
```diff
@@ -114,23 +114,142 @@ void main() {
...
+        final hash = await BlurhashService.generateBlurhash(
+          Uint8List.fromList(img.encodePng(gradient)),
+        );
+        expect(hash, isNotNull);
+
+        int horizontalContrast(double punch) {
+          final data = BlurhashService.decodeBlurhash(
+            hash!,
+            width: 16,
+            height: 1,
```
realmeylisdev: This decodes at `height: 1`, which collapses every component row's y-basis cosine to 1 — so it isn't actually isolating row 0 the way the test name claims. Confirmed by reintroducing blurhash_dart's exact punch bug (skip first row/column): all 42 tests still pass, including this one.

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
```diff
@@ -114,23 +114,142 @@ void main() {
...
+        final b = pixels[i + 2];
+        expect(
+          (r - g).abs() <= 2 && (g - b).abs() <= 2 && (r - b).abs() <= 2,
+          isTrue,
+          reason: 'pixel ${i ~/ 4} is tinted: r=$r g=$g b=$b',
+        );
+      }
+    });
+
+    test('decodes a solid fill back to its source color', () async {
```
realmeylisdev, line 177: For a solid fill, true AC energy is 0, so the buggy decode's fractional-shift error rounds away to nothing at 8-bit precision — verified by running both the buggy and fixed decode against this exact scenario (max per-pixel diff from the bug alone: [1,1,0], well under this test's tolerance of 5). This test would also pass under the pre-fix buggy code, so it isn't actually pinning the regression its own comment claims to guard against.

realmeylisdev (approved):
Reviewed the decode math against blurhash_dart's actual source (including live execution against both the buggy and fixed paths) — the color-tint and ringing root causes check out, and the fix is correct. Left inline notes: a couple of the new tests don't isolate what their names claim, plus some minor efficiency/cleanup items inherited from blurhash_dart. Nothing blocking.

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
hm21: Good catch — confirmed against the real decode path: with blurhash_dart's punch row/column skip reintroduced this still passed (175 vs 173 at punch 1 vs 0.5). Rewrote it to decode at the real size and assert `contrast(1) > contrast(0.5) * 1.4`, which the skip bug can't satisfy (175 > 242 fails). Verified it now goes red under the reintroduced bug. e84d213967

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
hm21: Right — a solid fill has ~zero AC energy so the division bug rounds away (max diff 3 < tol 5); it can't pin that regression. Reworded the comment to what it actually guards (DC / channel order + sRGB round-trip); the grayscale-tint test above already pins the decodeAc division bug. Also added a 4x7 legacy-shape test so the defaultPunch softening on already-published high-component hashes is actually covered. e84d213967

mobile/packages/blurhash_service/test/src/blurhash_service_test.dart
hm21: Added a test for exactly this: encodes a real 4x7 legacy-shape hash and asserts `defaultPunch` (0.8) damps its value spread vs punch 1.0 (17 vs 21), so the softening on already-published high-component hashes is now covered. e84d213967

rabble:
Pushed a small additive fix to this branch: **724f027f7 `fix(blurhash): restore log signal on malformed blurhash decode`**.

The new `_decodeComponents` silently `return null`ed on a blurhash length-mismatch, which dropped the error-log signal the pre-PR path emitted (malformed hashes used to hit `BlurHash.decode` → threw → the existing `Log.error('Failed to decode blurhash...')` caught it). Restored it: logs before the null return on the length-mismatch path, matching the file's existing `Log.error` name/category. Added a test decoding a truncated hash that asserts both `null` and a captured error log — it fails without the fix. `flutter analyze` + package tests green, coverage stays at the package's 100% gate.

This only restores the lost log; it deliberately does **not** touch either of the two threads still awaiting @NotThatKindOfDrLiz's call (the `defaultPunch` 0.8 justification and the upstream-workaround policy). Those remain your decisions.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
NotThatKindOfDrLiz: Blocking before merge: this needs an explicit owner decision, not just a code comment. Please either make default punch component-count-derived so newly generated 3x4/3x3 hashes decode at `1.0` while legacy high-component hashes decode at `0.8`, or document/track the intentional mobile/web contrast divergence so divine-web can be aligned or consciously left different. The decode fix is sound; this is the remaining product consistency risk.

mobile/packages/blurhash_service/lib/src/blurhash_service.dart
NotThatKindOfDrLiz: Blocking before merge: please close the workaround-policy loop for this custom decoder. The spec-correct implementation is fine, but it exists because the current `blurhash_dart` decoder has upstream bugs. File/link the upstream issue and add a local tracking issue/removal note, or explicitly switch this PR to a dependency/decoder path that avoids carrying the in-repo workaround. I do not think a dependency swap is required for this PR; the minimum merge requirement is traceability and a removal plan if upstream fixes land.

NotThatKindOfDrLiz (changes requested):
Requesting changes on the two remaining owner calls: `defaultPunch` cross-platform behavior must be fixed or explicitly tracked, and the local decoder workaround needs upstream/local traceability. The decode fix and tests look sound; package `flutter test` and `flutter analyze` are green locally.

hm21:
Closing in favor of #6189.

The remaining blocker here was tracking an upstream issue + removal plan for the in-repo decode workaround. `blurhash_dart` is unmaintained — `1.2.1` is the latest release (~3 years old) and we already pin exactly that — so there is nothing to file upstream and no realistic removal date to track. Keeping a tracking issue open forever for a dead package is pointless.

The ask for traceability is fair, though, so rather than carry a workaround-around-a-dependency, #6189 takes the approach the review kept pointing at: Divine ships its own spec-correct BlurHash codec (new `divine_blurhash` package) and drops `blurhash_dart` entirely. Same decode + ringing fix, nothing left to track.

PR #6072, opened 2026-07-13
fix(minor-safety): point minor-review endpoints at the relay-manager host

mbradley:
Closes divinevideo/divine-relay-manager#108

## what

the minor-account-review gate has been silently dead in production: the app calls /v1/account/moderation-status and /v1/minor-review-cases/{id}/parent-contact on the main backend host, which 404s them. MinorAccountReviewRepository fails open to active by design, so restricted accounts are never gated and the review/appeal screen is unreachable.

those endpoints live on the relay-manager worker. this PR routes the two calls through EnvironmentConfig.relayManagerApiUrl, the repo's existing env-aware model of that host (local dev worker, api-relay-staging, api-relay-prod), injected into ApiService as a required constructor param and wired through apiServiceProvider from the reactive currentEnvironmentProvider.

## why this shape

- verified live before changing anything: api.divine.video/v1/account/moderation-status is 404 (the host is funnelcake behind GKE nginx, no /v1/account routes), while api-relay-prod.divine.video answers 401 unauthenticated on the exact same path, meaning the worker serves it and just wants NIP-98. only the host was wrong.
- a client-side fix keeps NIP-98 consistent for free: the app signs the URL it calls, and the worker validates the signed URL against the request URL. any server-side proxy would break that validation unless the worker learned the public host.
- reuse over a new config: the first cut added a MODERATION_API_URL dart-define pinned to prod; review caught that the repo already models this host per environment (the Zendesk JWT flow consumes relayManagerApiUrl for NIP-98 calls to the same worker), and a prod pin would have sent staging/local/env-switched builds' signed minor-review traffic into prod D1. the constructor param is required, so no build can silently fall back to the wrong host, which is the exact bug class this PR fixes.

## behavior by environment

- production builds: api-relay-prod, gate goes live (this is the fix).
- staging builds: api-relay-staging. correction from the relay-manager#174 review: the staging worker's /v1 paths bypass CF Access (verified live; devices must reach them), so staging builds exercise the gate end-to-end against staging data. the kDebugMode local override service remains available for simulating gate states without backend setup.
- local stack: the local relay-manager dev worker (which serves these endpoints), so full local QA works.
- developer environment switcher: applies live (provider watches the environment).

## verification

- TDD both rounds: the host-assertion test previously pinned the wrong host (api.divine.video); it now pins api-relay-prod for the prod config, the parent-contact test gained host+path assertions (it previously asserted neither), and an injection test pins that a staging base URL routes to api-relay-staging (env-awareness at the service boundary).
- flutter analyze clean; api_service + rate-limit suites 13/13; minor-review repository/override/providers suites 573 pass; CI green.
- no visual change. the gate's UI already exists and is tested; this makes it reachable.

## notes for release

- the fix only takes effect for builds that ship it. existing installs keep failing open until users update. server-side coverage for old installs is divine-relay-manager#173; a synthetic check so this class of breakage alerts instead of hiding is divine-relay-manager#174.

### DIFF
--- a/mobile/lib/config/app_config.dart
+++ b/mobile/lib/config/app_config.dart
@@ -13,6 +13,16 @@ class AppConfig {
     defaultValue: 'https://api.openvine.co',
   );
 
+  /// Moderation API (relay-manager worker). The minor-account-review
+  /// endpoints (/v1/account/moderation-status, /v1/minor-review-cases/*)
+  /// live here, NOT on the main backend host — calling them on
+  /// [backendBaseUrl] 404s and the review gate silently fails open
+  /// (divine-relay-manager#108).
+  static const String moderationApiBaseUrl = String.fromEnvironment(
+    'MODERATION_API_URL',
+    defaultValue: 'https://api-relay-prod.divine.video',
+  );
+
   static const String inviteServerBaseUrl = String.fromEnvironment(
     'INVITE_SERVER_URL',
     defaultValue: 'https://invite.divine.video',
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
🔒 minor-review routes now use relay-manager

PR #6075, opened 2026-07-13
fix(dm): address post-merge review of #6046 (Keycast timeout, group-send cancel/dedup, retry heartbeat, relay repair)

realmeylisdev:
## Description

Follow-up to the durable-DM PR #6046, addressing the nine findings in NotThatKindOfDrLiz's post-merge review ([#pullrequestreview-4688621299](https://github.com/divinevideo/divine-mobile/pull/6046#pullrequestreview-4688621299)). Each finding was independently re-verified against merged `main` before fixing; seven were confirmed live/latent bugs and two (F3, F9) were confirmed mechanisms whose worst case is bounded by other guards — those two are fixed as hardening. Every fix ships with a regression test.

One commit per finding group:

- **`fix(keycast)` — F1 (Keycast global timeout regression).** #6046 lowered `KeycastRpc.defaultRequestTimeout` from 30s to 12s, but the 12s bound (sized for the DM send pipeline) shipped as the global default every single-op RPC caller inherits — none of the five `fromSession` sites override it. Under the known production Keycast latency (DB-pool contention, 20–30s single-op), custodial OAuth users regressed video-publish signing, likes, reposts, and follows from slow-but-succeeding to failing at 12s. Restored the single-op default to 30s (batch verb stays 30s). DM durability doesn't depend on the tighter bound — the outgoing-DM queue re-drives and dm_repository has a 90s publish backstop. Scoping a fail-fast bound to the DM crypto chain specifically is a cleaner follow-up (single-op RPCs expose no per-call timeout and the signer is shared), noted in Out of Scope.

- **`fix(dm)` group-send — F2 + F4.**
  - *F2 (cancelled group sends kept fanning out):* the per-recipient publish loop had no pre-publish row re-read, so `cancelOutgoingBatch` (long-press delete on the pending bubble) mid-fan-out still delivered to later recipients — invisible to the sender and unretractable. Now re-reads the sibling row before each publish and skips when it's gone, mirroring `recoverFullSend`'s existing interlock.
  - *F4 (self-wrap ingest didn't stamp `send_batch_id`):* `_persistDecryptedGiftWrap` never extracted the encrypted `['batch', id]` tag, so a self-wrap echo ingested before the batch's local persist wrote a NULL-batch row invisible to `hasMessageWithSendBatchId` — a duplicate bubble now, a second durable row when a sibling later confirmed. Now stamps `sendBatchId` for self-authored rumors only (owner-scoped dedup must not honor a peer-forged tag) with a 64-hex validity guard.

- **`fix(dm)` retry-service — F3 + F5.**
  - *F5 (live): a sweep-level throw silenced the 30s heartbeat.* The top-level `catch` never re-armed the follow-up timer, so a transient DB/DAO throw left the queue un-driven until the next external trigger — exactly the steady-foreground, stable-network session the heartbeat exists for. Re-arm from the `catch` with `workRemains: true`; the next clean pass self-cancels when the queue drains.
  - *F3 (hardening): the exhausted-row terminalize could clobber a delivered recipient wrap.* No shipped write path currently produces the `recipient=sent / self=pending` hybrid (that write is always transactionally coupled to delete-or-`self=failed`), so this is defensive: when the recipient wrap is already `sent`, fail only the self wrap, mirroring the offline path's existing guard.

- **`fix(dm)` copy — F6 (mislabelled action).** The failed-send snackbar's "Delete message" only cancels the outgoing queue rows; for a partially delivered group bubble it leaves the persisted message in place, which then flips back to a normal delivered bubble. Keeping the already-delivered copy is intended, so this renames the label to "Stop trying" (+ corrected ARB description), translated into all 17 non-English locales. Long-press delete remains the real delete-for-everyone path.

- **`fix(dm)` video bubble — F7 (tap hijack).** On a failed own video-share (or reply quoting a video), the inner card's tap recognizer opened the reel, winning the arena over the dominant surface — leaving the tap-only resend affordance reachable only from the narrow "Not delivered" strip. Threads an `enableTap` flag down and nulls the inner `onTap` on a failed own send, so the outer failed-tap wins across the whole bubble and assistive tech stops announcing a dead open-video action.

- **`fix(relay)` — F8 + F9.**
  - *F8 (no repair cooldown):* the zombie-socket repair had only in-flight dedup, so a brownout relay replying slower than the OK window got force-cycled on every timed-out publish (destroying late OKs, replaying subscriptions, re-running AUTH). Added a per-relay cooldown (injectable, default 60s = 2× the OK window).
  - *F9 (hardening): force-reconnect stranded one-shot queries.* It replayed long-lived subscriptions but not the `_queries` map, so a repaired query never re-issued its REQ and the caller waited its full timeout instead of completing early. Now replays pending one-shot queries on reconnect and fixes the adjacent `_queryCompleteCallbacks` leak in `unsubscribe`.

**Related Issue:** Relates to #6046

## Out of Scope

- Scoping a dedicated fail-fast timeout to the DM signer crypto chain (so DM sends still fail fast while other RPCs keep the 30s headroom) — needs a per-call timeout surface on `signEvent`/`nip44*` that doesn't exist today. F1 restores the safe 30s global default in the meantime.
- The 1:1 `sendMessage` same-second collision (pre-existing, reviewer-scoped-out on the earlier batch-id thread) — separate follow-up.
- F6 Option B (route the failed-send action through delete-for-everyone) was not taken; the rename matches the intentional keep-the-message semantics. If product wants delete-for-everyone from that snackbar, that's a separate copy+behavior change.

## Verification

- `flutter analyze lib test` (app) + `flutter analyze lib test` on `dm_repository`, `keycast_flutter`, `nostr_sdk` — no issues.
- App suites: `outgoing_dm_retry_service_test.dart`, `conversation_view_test.dart`, `message_bubble_test.dart`, `arb_consistency_test.dart` (151 tests) — pass.
- `keycast_flutter` (208), `nostr_sdk` unit incl. zombie-remediation (285), `dm_repository` (511) — pass.
- `dm_repository` coverage 93.59% ≥ 93 gate.
- New regression tests mutation-checked (each goes red when its fix is neutered).

## Type of Change


mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -4397,8 +4448,14 @@ class DmRepository {
     // No recipient succeeded but retryable rows exist: make sure a
     // brand-new group thread is visible in the inbox so the user can find
     // the failed bubbles to retry or delete. All-blocked sends leave no
-    // rows behind (terminal), so they create no thread cruft.
-    if (!results.any((r) => r.success) && results.any((r) => !r.blocked)) {
+    // rows behind (terminal), so they create no thread cruft. A fully
+    // cancelled batch must NOT resurrect the thread the user just deleted,
+    // so cancelled indexes don't count as retryable failures here.
+    final hasNonCancelledFailure = results.asMap().entries.any(
```
hm21, line 4454: Nit (non-blocking, residual F2 edge case): `hasNonCancelledFailure` keys off `results[]`, not whether each failed index's row survived cancellation. If a sibling returns a *hard* failure and the user then long-press-deletes the batch during a later sibling's publish window, `cancelOutgoingBatch` deletes the hard-failed row too — but this index still counts as a non-cancelled failure here, so `_ensureConversationVisibleAfterSendFailure` resurrects the (now empty, no bubble) group thread the user just deleted. Cosmetic + self-healing, needs a precise race — worth a note, not a fix-now.

mobile/lib/services/outgoing_dm_retry_service.dart
```diff
@@ -589,6 +589,14 @@ class OutgoingDmRetryService {
...
         ),
       );
+      // A throwing pass never reached the try-path _scheduleFollowUp, so
+      // without re-arming here a transient DAO error (e.g. a busy database)
+      // would silence the 30s heartbeat until an external trigger — exactly
+      // the steady-foreground, stable-network session the heartbeat exists
+      // for. Queue state is unknown after a throw, so assume work remains;
+      // the next successful pass's own workRemains:false self-cancels the
+      // timer once the queue drains.
+      _scheduleFollowUp(workRemains: true);
```
hm21, line 622: Nit (minor): re-arming here is correct. Only caveat — the top-level throw comes from DAO reads that run before any per-row work, so a *persistently*-throwing DAO (non-self-healing SQLite corruption/lock) never advances a retry counter to drain the queue; it just re-reports `sweepTopLevel` + re-arms every 30s for the whole foreground session. Crashlytics dedupes by stack trace so it's not a true flood, but consider gating `recordError` on the first fault of a consecutive-fault streak while still re-arming.

hm21 (approved):
Went through all six commits against `main` and re-checked each finding against the merged code. All nine are correctly diagnosed real bugs (F3/F9 fairly scoped as hardening), the fixes are correct, and the new tests are genuine regression checks that go red on revert. No blockers — a few non-blocking nits:

- **F6 stale comment** (`conversation_view.dart:509`, not in this diff): `_onFailedMessageTap`'s doc still calls the secondary action `Delete ("cancel send")`, but this rename makes it "Stop trying" and the action never deletes the persisted message — the doc now contradicts the shipped copy it describes.
- **F5, minor**: the catch-path re-arm is correct, but a *persistently*-throwing DAO now re-reports `sweepTopLevel` to Crashlytics every 30s for the whole foreground session instead of once. Fine for the transient busy-DB case it targets; consider reporting only on the first fault of a streak. (inline)
- **F2 residual edge case**: see inline on `dm_repository.dart`.

mobile/lib/services/outgoing_dm_retry_service.dart
realmeylisdev: Done in ade10d486e — the `sweepTopLevel` `recordError` now escalates only on the first fault of a consecutive-throw streak; the re-arm is unchanged, and the first non-throwing pass (including the offline-gate early return) resets the streak so a genuinely new fault reports again. `Log.error` still fires every pass. Added regression tests for once-per-streak and report-again-after-recovery.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: Agreed it's real — deferring per your "note, not a fix-now" framing. Guarding it would need a per-index DAO existence re-read before the visibility check (a round-trip on the send path) to catch a cosmetic, self-healing thread resurrection that only fires on a precise concurrent long-press-delete race. Not worth the cost here; leaving this thread as the record.

realmeylisdev:
Addressed the review-body nit too:

- **F6** (stale `_onFailedMessageTap` doc): fixed in 98fe07591a — the doc now describes the "Stop trying" label and its cancel-undelivered-rows (not delete) behavior.

F5 and F2 answered inline (F5 fixed in ade10d486e with regression tests; F2 deferred with rationale). Thanks for the careful pass.

PR #1335, opened 2026-02-06
feat(divine_ui): add design system components and icons

Chardot:
## Summary
- Add core divine_ui design system components: DivineButton, DivineIconButton, DivineIcon, DivineCheckbox, DivineTextLink
- Add 172 Phosphor icon SVGs with proper 32x32 sizing
- Icons automatically derive colors from button type/variant - callers never specify icon color

---

## DivineIcon

SVG icon wrapper using `DivineIconName` enum (172 icons from Phosphor bold weight).

Screenshot 2026-02-06 at 21 15 10

### Properties

| Property | Type | Default | Description |
|----------|------|---------|-------------|
| `icon` | `DivineIconName` | required | Icon to display |
| `size` | `double` | `24` | Width and height in pixels |
| `color` | `Color?` | `null` | Tint color (uses original SVG colors if null) |

### Icon Variants

Icons support three weight variants via naming convention:

| Variant | Suffix | Example |
|---------|--------|---------|
| Bold (default) | none | `heart` |
| Fill | `_fill` | `heartFill` |
| Duotone | `_duo` | `heartDuo` |

### Naming Conventions

| Context | Convention | Example |
|---------|------------|---------|
| Enum entry | camelCase | `arrowBendDownLeft` |
| SVG filename | snake_case | `arrow_bend_down_left.svg` |
| Legacy filenames | PascalCase | `CaretDown.svg`, `CaretLeft.svg`, `Check.svg`, `DotsThree.svg`, `FilmSlate.svg` |
| Fill variants | `_fill` suffix | `heart_fill.svg` → `heartFill` |
| Duotone variants | `_duo` suffix | `heart_duo.svg` → `heartDuo` |
| Custom icons | descriptive | `copy_npub.svg`, `lightning_a.svg` |

### Legacy SVG Files (Not in Enum)

The following 36 SVG files exist in `assets/icon/` but are **not** included in `DivineIconName`. These are legacy icons used by older code that predates the divine_ui design system:

**UI Controls:**
`add`, `delete`, `menu`, `more_horiz`, `refresh`, `save`, `send`, `undo`

**Media:**
`music`, `video_camera`, `volume_off`, `volume_on`

**Camera:**
`flash_auto`, `flash_off`, `flash_on`, `retro-camera`

**Editor:**
`crop_portrait`, `crop_square`, `draw`, `text`, `trim`, `tune`

**Navigation:**
`arrow_forward`, `ArrowClockwise`

**Social:**
[...]

PR #6076, opened 2026-07-14
refactor(divine-ui): unify DiVineAppBarIconButton and DivineIconButton

realmeylisdev:
## Summary

Closes #2948. Unifies `DivineAppBarIconButton` (formerly `DiVineAppBarIconButton`) and `DivineIconButton` — they existed separately purely because of a 4-day build-sequencing gap between the two PRs that created them (#1256, #1335), not a deliberate design decision.

**Scope**: `feature_flag_screen.dart` and `user_list_people_screen.dart` migrate off `MaterialIconSource`. `creator_analytics_screen.dart` is intentionally **out of scope** — its `Icons.bug_report`/`Icons.bug_report_outlined` toggle has no existing or unwired `DivineIconName` equivalent anywhere in the repo (confirmed by a full-repo asset search); it needs new filled/outline design assets and will get its own follow-up issue.

## Changes (5 commits)

1. **Extend `DivineIconButton`'s public API** — add an optional `tooltip` param and an additive `DivineIconButton.fromSource(IconSource)` named constructor. Both are non-breaking; none of the 73 existing call sites change.
2. **Rename** `DiVineAppBarIconButton` → `DivineAppBarIconButton` — fixes the inconsistent "DiVine" capitalization against the rest of the design system.
3. **Delegate `DivineAppBarIconButton` to `DivineIconButton` internally**, replacing its hand-built `IconButton`/`Container` composition:
   - Added `DivineIconButton.showShadow` (default `true`, set `false` here) — `DivineIconButtonType.primary`/`.secondary` apply a subtle drop shadow whenever enabled, which the old widget never had; without suppressing it, every app-bar icon in the app would have gained a shadow.
   - Dropped `size`/`iconSize`/`borderRadius` params entirely: verified **zero** of the ~55 `DiVineAppBar(style:)` call sites in the app override `iconButtonSize`/`iconSize`/`iconButtonBorderRadius` — every real usage renders at exactly 40/24/16, i.e. exactly `DivineIconButtonSize.small`. Removed the now-fully-dead fields from `DiVineAppBarStyle` too (leaving them would've been a silent no-op footgun for any future caller).
   - Non-null `borderSide` now maps to `DivineIconButtonType.secondary`, whose built-in 2px `outlineMuted` border matches the only `borderSide` value ever passed anywhere in the app (`solidStyle`'s) exactly.
   - **Known accepted side effect**: disabled app-bar icons now dim to 32% opacity (previously no dimming), matching every other `DivineIconButton` in the app. The one reachable case is `blossom_settings_screen.dart`'s save button while saving.
4. **Migrate the 2 in-scope screens**: `Icons.restore` → `DivineIconName.arrowCounterClockwise`, `Icons.person_add_alt_1` → `DivineIconName.userPlus`.
5. **Deprecate (not remove) `MaterialIconSource`** — `@Deprecated`, still fully functional. `divine_ui`'s own test suite continues to exercise it via scoped `ignore_for_file` directives (each with an inline explanation).

## Testing

- `divine_ui` package: 100% line coverage maintained (strict CI gate).
- Full app test suite (`very_good test --optimization`): 12,164 tests passing, 0 failures.
- `flutter analyze lib test integration_test`: clean.
- Rewrote `divine_app_bar_icon_button_test.dart` for the new delegation-based implementation (property assertions on color/border/radius/shadow rather than the old widget-tree shape).
- Fixed 2 pre-existing screen tests whose `find.byIcon(Icons.xxx)` assertions broke once the icon became SVG-based, using `find.bySemanticsLabel` with the existing localized tooltip strings (no copy changes).
- Golden tests were considered but skipped: `divine_ui` has no existing golden-test infrastructure (no `alchemist`/`golden_toolkit` dev dependency, no CI wiring) — building that from scratch was out of scope for this refactor. The property-based tests already lock in the same visual contract (exact colors, border, radius, shadow-null) at the code level.

## Manual verification

No screen's tap target, colors, or icon glyph changed visually except the two migrated icons (which now render via SVG instead of Material) and the one accepted opacity-dimming side effect noted above.

### DIFF
--- a/mobile/lib/screens/creator_analytics_screen.dart
+++ b/mobile/lib/screens/creator_analytics_screen.dart
@@ -166,7 +166,7 @@ class _CreatorAnalyticsScreenState
         actions: [
           DiVineAppBarAction(
             // No DivineIconName equivalent exists yet for this filled/
-            // outline bug-report toggle; tracked for a future migration.
+            // outline bug-report toggle — tracked in #6081.
             // ignore: deprecated_member_use
             icon: MaterialIconSource(
               _showDiagnostics ? Icons.bug_report : Icons.bug_report_outlined,
--- a/mobile/lib/screens/creator_analytics_screen.dart
+++ b/mobile/lib/screens/creator_analytics_screen.dart
@@ -165,6 +165,9 @@ class _CreatorAnalyticsScreenState
         showBackButton: true,
         actions: [
           DiVineAppBarAction(
+            // No DivineIconName equivalent exists yet for this filled/
+            // outline bug-report toggle; tracked for a future migration.
+            // ignore: deprecated_member_use
             icon: MaterialIconSource(
               _showDiagnostics ? Icons.bug_report : Icons.bug_report_outlined,
             ),
--- a/mobile/packages/divine_ui/lib/src/app_bar/divine_app_bar_icon_button.dart
+++ b/mobile/packages/divine_ui/lib/src/app_bar/divine_app_bar_icon_button.dart
@@ -1,11 +1,12 @@
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter_svg/flutter_svg.dart';
 
 /// A styled icon button for use in [DiVineAppBar].
 ///
-/// Renders icons in a rounded container with consistent styling
-/// matching the AppShell design system.
+/// Delegates its rendering to [DivineIconButton] at
+/// [DivineIconButtonSize.small] (40px visible / 24px icon / 16px border
+/// radius / 48px tap target) — every [DiVineAppBarStyle] preset in the app
+/// uses exactly these dimensions, so no configurable size is exposed here.
 ///
 /// Example usage:
 /// ```dart
@@ -25,9 +26,6 @@ class DivineAppBarIconButton extends StatelessWidget {
     this.backgroundColor,
     this.borderSide,
     this.iconColor,
-    this.size = 48,
-    this.iconSize = 32,
-    this.borderRadius = 20,
     super.key,
   });
 
@@ -50,7 +48,10 @@ class DivineAppBarIconButton extends StatelessWidget {
 
   /// Optional border for the button container.
   ///
-  /// When null, no border is shown
+  /// When non-null, the button renders as [DivineIconButtonType.secondary],
+  /// whose built-in 2px [VineTheme.outlineMuted] border matches every
+  /// current usage of this param exactly. The supplied [BorderSide]'s own
+  /// color/width are not independently honored.
   final BorderSide? borderSide;
 
   /// Color of the icon.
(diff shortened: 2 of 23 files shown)


realmeylisdev:
Device QA checklist for the icon-button unification PR: ⧉ Device QA

(https://claude.ai/code/artifact/01bc41f2-1be5-4a3e-a1d9-45b1ad054dd1)

mobile/packages/divine_ui/lib/src/app_bar/divine_app_bar_icon_button.dart
```diff
@@ -50,86 +48,31 @@ class DiVineAppBarIconButton extends StatelessWidget {
...
-          icon: buttonContent,
-        ),
-      ),
+    return DivineIconButton.fromSource(
+      icon: icon,
+      onPressed: onPressed,
+      type: borderSide != null
+          ? DivineIconButtonType.secondary
+          : DivineIconButtonType.primary,
+      size: DivineIconButtonSize.small,
```
hm21, line 70: Blocking: `DivineIconButtonSize.small`'s effective tap target is 40x40, not 48. The 4px + 2px outer paddings sit outside the `InkWell`, so a ~6px ring around each button is dead (measured: footprint 52, tappable InkWell 40 — taps 2-5px from the edge don't fire). The old widget forced `SizedBox(48,48)`/`IconButton`, and the test you removed asserted the 48x48 target "must be enforced regardless of the visual button size." So app-bar action buttons (and non-`expandHitArea` leading buttons) regress 48->40, below the 48x48 minimum in accessibility.md — which contradicts the PR's "no tap target changed" note. Fix by restoring a 48 hit area here, or making `small` genuinely deliver 48 (its own docstring already claims the 4px padding "captures taps", which it doesn't), and re-add a tap-target test.

mobile/lib/screens/creator_analytics_screen.dart
```diff
@@ -165,6 +165,9 @@ class _CreatorAnalyticsScreenState
         showBackButton: true,
         actions: [
           DiVineAppBarAction(
+            // No DivineIconName equivalent exists yet for this filled/
+            // outline bug-report toggle; tracked for a future migration.
+            // ignore: deprecated_member_use
```
hm21, line 170: Nit: good that the ignore carries a rationale — cite the follow-up issue number once it exists (`// ignore: deprecated_member_use — see #NNNN`) so "tracked for a future migration" isn't orphaned.

hm21 (commented):
Process note (author QA): the linked Device QA checklist is unchecked (0/14) and lives in an external artifact not every reviewer can open — please run it yourself before requesting review and put the result (ticked checklist + a screen recording) directly in the PR; the self-review checklist asks for recordings on UI changes. Your own CRITICAL item "Tap responsiveness unchanged — no dead zone around the visible 40px pill" is exactly the regression flagged above: running that one item on-device would have caught it.

mobile/packages/divine_ui/lib/src/app_bar/divine_app_bar_icon_button.dart
realmeylisdev: Fixed in 0609a341: moved the small-only 4px padding inside the InkWell (a centered 48x48 SizedBox) instead of outside it, so the whole 48x48 region is tappable, not just the 40px visible pill. Added a tap-target test that taps 22px from center (outside the 40px pill, inside the 48px target) to guard the regression, plus a structural assertion on the InkWell's child. Base size is untouched (verified with its own test) since its InkWell was already 48x48.

mobile/lib/screens/creator_analytics_screen.dart
realmeylisdev: Fixed in 636377145f — filed #6081 for the missing bug_report/bug_report_outlined glyphs and referenced it in the ignore comment.

realmeylisdev:
@hm21 both review items addressed:

- Blocking tap-target regression: fixed in 0609a341. `DivineIconButtonSize.small`'s InkWell now covers the full 48x48 region (previously only the 40px visible pill was tappable). Added tests: a structural assertion on the InkWell's child and a behavioral test that taps 22px from center (outside the 40px pill, inside the 48px target) and confirms `onPressed` fires. Base size is unaffected — verified with its own test.
- Nit: filed #6081 for the missing bug_report glyphs and cited it in the ignore comment (636377145f).

On the process note: I don't have a device/simulator in this environment to run the QA checklist and record a video. That'll need to happen from your or another reviewer's machine before merge — flagging so it's not missed.

hm21 (approved):
LGTM

realmeylisdev:
@hm21 ran the actual QA checklist just now on an iPhone 17 Pro simulator against `636377145f`:

- Back button (`solidStyle`): correct glyph/pill on Settings, Notifications, Media Servers — zoomed screenshot confirms no drop shadow.
- Tap-target fix verified live, not just by test: tapped inside the 48×48 target but outside the visible 40px pill → navigated back successfully.
- Feature flags reset-all icon renders and reads as "reset to default" at a glance.
- Blossom "Save" round-trips correctly (fires and pops back); didn't manage to catch the dimmed in-flight frame on screenshot, it's too fast in this environment.
- Creator analytics bug icon confirmed untouched (still the old Material glyph).

PR #6079, opened 2026-07-14
fix(divine_video_player): prevent SIGABRT from invalid AVVideoComposition

### LINKED ISSUE #6080: fix(divine_video_player): prevent SIGABRT from invalid AVVideoComposition
Auto-created from PR #6079.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6079
**Branch:** `fix/video-composition-avplayer-crash`

hm21:
Closes #6080

## Description

Fixes a hard crash (`EXC_CRASH` / `SIGABRT`) in the native multi-clip player. Reported via TestFlight feedback on 1.0.16 (804), iPhone 13 mini, iOS 26.5 — the app aborted ~3s after launch.

The crashing frame:

```
2  AVFCore   -[AVPlayerItem setVideoComposition:] + 520
3  Runner    DivineVideoPlayerInstance.handleSetClips … DivineVideoPlayerInstance.swift:203
```

`-[AVPlayerItem setVideoComposition:]` validates the composition synchronously and throws an Objective-C `NSInvalidArgumentException` when it is malformed (zero render size, or zero/invalid frame duration). That NSException is **not** a Swift `Error`, so the enclosing `do/catch` in `handleSetClips` cannot catch it — it propagates uncaught to `std::terminate` → `abort()`. Firebase's `FIRCLSTerminateHandler` logged it.

Two bugs fed the crash, both when the first loaded clip is rotated (so a `videoComposition` is actually built):

1. **Frame duration could be valid-but-zero.** `buildComposition` set `frameDuration = sourceVideoTrack.minFrameDuration` guarded only by `.isValid`. A zero `CMTime` *is* valid (`isValid == true`), and on assets without frame-timing metadata `minFrameDuration` can come back valid-but-zero. `setVideoComposition:` rejects a zero `frameDuration`. Now guarded by `isNumeric && seconds > 0`, falling back to 30fps otherwise. (`minFrameDuration` is also now `await`-loaded like the sibling track properties, since iOS 16+ deprecates the synchronous accessor.)
2. **Render-size validation ran too late.** The `renderSize.isPositive` guard sat *after* the throwing `playerItem.videoComposition = …` assignment, so it could never protect it. Both render-size and frame-duration are now validated *before* the assignment.

A rejected composition now throws a Swift `CompositionError` that the existing `catch` block turns into a `COMPOSITION_ERROR` `FlutterError`, so the failure surfaces to Dart instead of killing the process.

### Same crash class: `scaleTimeRange`

`AVMutableComposition.scaleTimeRange(_:toDuration:)` (used for per-clip playback speed) is a non-`try` AVFoundation call that raises the *same* uncatchable `NSInvalidArgumentException` when `toDuration` is zero or non-numeric — and the enclosing `do/catch` can't catch it either. `playbackSpeed` is normally bounded: it defaults to `1.0` (so the reported single-clip feed playback never scales) and the editor clamps user values to `[0.25, 3.0]` (`VideoEditorConstants.clipSpeedMin`/`clipSpeedMax`) — so this guard is defensive hardening against a future/edge value rather than the trigger of the reported crash. A duration that rounds to zero or overflows to non-numeric is now guarded with `isNumeric && seconds > 0` before the scale, falling back to the unscaled clip rather than aborting.

### Why pure-Swift validation and not an Objective-C `@try/@catch`

Catching the NSException itself would need a separate Objective-C target wired into **both** `Package.swift` (SwiftPM) and the `.podspec` (CocoaPods) with module-map plumbing — build-system surface that can't be verified without a full native build. For this composition shape (single/multi-clip, one instruction, the composition's own track) the only values AVFoundation hard-rejects are render size and frame duration; the instructions are valid by construction. Validating those before assignment removes the crash without touching the build system.

## Testing

- Added `apple_composition_guard_contract_test.dart` — a Swift **source-contract** test, the same pattern as the existing `apple_threading` / `apple_engine_teardown` contract tests in this package. It pins the invariants that have no Dart runtime surface: the render-size and frame-duration guards must precede the `setVideoComposition:` assignment, `minFrameDuration` must fall back to 30fps when not numeric-positive, and the scaled duration must be guarded before `scaleTimeRange`. A refactor that restores the uncaught crash keeps every runtime test green — only this contract test (index-ordering assertions) catches it.
- The package still has no XCTest target, and exercising `setVideoComposition:` validation at runtime realistically needs real video assets, so the guards are pinned structurally rather than through a native unit test.

## Out of Scope

- No Objective-C exception-catching bridge (see rationale above).

## Verification

- `flutter test` (divine_video_player) — all contract tests pass (12/12, including the 3 new composition-guard assertions).
- `swiftc -parse` on the changed file — no syntax errors (full compile runs through Pods since the file imports `Flutter`/`FlutterMacOS`).
- Manual on-device / TestFlight verification of the previously-crashing clip pending.

## Type of Change

### DIFF
--- a/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
+++ b/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
@@ -438,9 +438,19 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
             let scaledDuration: CMTime
             if clipSpeed != 1.0, clipSpeed > 0 {
                 let scaledSeconds = CMTimeGetSeconds(clipDuration) / clipSpeed
-                scaledDuration = CMTime(seconds: scaledSeconds, preferredTimescale: 600)
-                let insertedRange = CMTimeRange(start: insertTime, duration: clipDuration)
-                composition.scaleTimeRange(insertedRange, toDuration: scaledDuration)
+                let candidate = CMTime(seconds: scaledSeconds, preferredTimescale: 600)
+                // A pathological playbackSpeed can round the scaled duration to
+                // zero or overflow it to a non-numeric CMTime. Like
+                // setVideoComposition:, scaleTimeRange throws an uncatchable
+                // Objective-C NSInvalidArgumentException on such a duration, so
+                // fall back to the unscaled clip rather than abort the process.
+                if candidate.isNumeric, candidate.seconds > 0 {
+                    scaledDuration = candidate
+                    let insertedRange = CMTimeRange(start: insertTime, duration: clipDuration)
+                    composition.scaleTimeRange(insertedRange, toDuration: candidate)
+                } else {
+                    scaledDuration = clipDuration
+                }
             } else {
                 scaledDuration = clipDuration
             }
--- a/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
+++ b/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
@@ -199,13 +199,23 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
                 // cycles from video decode/render on weaker devices, showing up
                 // as dropped or stuttering frames during sped-up preview.
                 playerItem.audioTimePitchAlgorithm = .timeDomain
-                if let videoComposition {
-                    playerItem.videoComposition = videoComposition
-                }
-                if let audioMix { playerItem.audioMix = audioMix }
+                // Validate BEFORE assigning. -[AVPlayerItem setVideoComposition:]
+                // throws an Objective-C NSInvalidArgumentException on an invalid
+                // composition (zero render size or zero/invalid frame duration).
+                // That exception is not a Swift Error, so the enclosing do/catch
+                // can't catch it and the process aborts (SIGABRT). Rejecting the
+                // bad values here surfaces the failure as a FlutterError instead.
                 guard (videoComposition?.renderSize ?? composition.naturalSize).isPositive else {
                     throw CompositionError.invalidRenderSize
                 }
+                if let videoComposition {
+                    let frameDuration = videoComposition.frameDuration
+                    guard frameDuration.isNumeric, frameDuration.seconds > 0 else {
+                        throw CompositionError.invalidFrameDuration
+                    }
+                    playerItem.videoComposition = videoComposition
+                }
+                if let audioMix { playerItem.audioMix = audioMix }
                 self.templateItem = playerItem
 
                 let requestedStartPositionMs = (args["startPositionMs"] as? NSNumber)?.int64Value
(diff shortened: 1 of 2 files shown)


mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
```diff
@@ -401,8 +411,13 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
                     let mutableVideoComposition = AVMutableVideoComposition()
                     mutableVideoComposition.renderSize = displaySize
                     mutableVideoComposition.sourceTrackIDForFrameTiming = videoTrack.trackID
-                    if sourceVideoTrack.minFrameDuration.isValid {
-                        mutableVideoComposition.frameDuration = sourceVideoTrack.minFrameDuration
+                    // minFrameDuration can be valid-but-zero on assets without
+                    // frame-timing metadata; a zero frameDuration makes
+                    // setVideoComposition throw. Require a positive duration and
+                    // fall back to 30fps otherwise.
+                    let minFrameDuration = sourceVideoTrack.minFrameDuration
```
realmeylisdev: The guard is correct and strictly stronger than `.isValid` — keep it. One root-cause note: `minFrameDuration` is read synchronously here, while `naturalSize`/`preferredTransform` are `await`-loaded at ~L374. On iOS 16+ a synchronous read of an un-loaded `AVAssetTrack` property is deprecated and returns an undefined value — a more concrete explanation for a bad frame duration than "valid-but-zero metadata," which Apple only documents as `kCMTimeInvalid` (and that was already caught by the old `.isValid`). Consider `await sourceVideoTrack.load(.minFrameDuration)` to fix at the root; the guard stays a valid safety net either way. Non-blocking.

mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
```diff
@@ -423,9 +438,19 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
...
-                scaledDuration = CMTime(seconds: scaledSeconds, preferredTimescale: 600)
-                let insertedRange = CMTimeRange(start: insertTime, duration: clipDuration)
-                composition.scaleTimeRange(insertedRange, toDuration: scaledDuration)
+                let candidate = CMTime(seconds: scaledSeconds, preferredTimescale: 600)
+                // A pathological playbackSpeed can round the scaled duration to
+                // zero or overflow it to a non-numeric CMTime. Like
+                // setVideoComposition:, scaleTimeRange throws an uncatchable
+                // Objective-C NSInvalidArgumentException on such a duration, so
+                // fall back to the unscaled clip rather than abort the process.
+                if candidate.isNumeric, candidate.seconds > 0 {
```
realmeylisdev, line 452: Sound defensive addition. The description's framing of `playbackSpeed` as "feed-controlled, unbounded" isn't accurate though: it's a local editor control clamped to [0.25, 3.0] (`VideoEditorClipSpeedSheet` → `clipSpeedMin`/`clipSpeedMax`), and the feed plays a single clip at the default 1.0 (`setSource`), so this branch isn't reached on the reported feed crash. Worth softening the wording — the guard itself is fine to keep as hardening. Non-blocking.

realmeylisdev (approved):
Traced the crash path and confirmed the AVFoundation semantics: `setVideoComposition:` and `scaleTimeRange` raise a synchronous ObjC `NSInvalidArgumentException` that a Swift `do/catch` can't catch, and the pre-validation guards (`isNumeric && seconds > 0`, render-size/frame-duration before assignment) correctly convert that into a `CompositionError`. Guard logic is sound, the contract test is non-vacuous (passes, fails on revert/reorder). Approving. Two non-blocking accuracy notes inline.

mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
hm21: Good catch — fixed in b95b8b7: `minFrameDuration` is now `await`-loaded via `load(.minFrameDuration)`, matching the sibling `naturalSize`/`preferredTransform` reads. The `isNumeric && seconds > 0` guard stays as the safety net.

mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
hm21: Agreed — softened the PR description. Verified `playbackSpeed` defaults to 1.0 and the editor clamps to [0.25, 3.0] (`clipSpeedMin`/`clipSpeedMax`), so this branch is never hit on the reported feed crash; the guard stays as defensive hardening.

PR #6082, opened 2026-07-14
fix(dm): use a bottom sheet instead of a snackbar for failed-send recovery

### LINKED ISSUE #6092: fix: DM failed-send recovery SnackBar crowds action text in longer locales and auto-dismisses too fast
## Summary
The failed-DM-send recovery prompt (tap a failed own message bubble → Resend / Delete message) is a `DivineSnackbarContainer` SnackBar, and its fixed-width, auto-dismissing layout doesn't hold up once the action copy gets longer than English.

## Environment
- App version: Latest (main)
- Device: Any — most visible on narrower phones
- Network: N/A
- Locale: Reproduces in any language with longer action labels than English (e.g. German)

## Steps to Reproduce
1. Set the device locale to German (or another locale with longer strings than English).
2. In a DM conversation, send a message under conditions that cause the send to fail (e.g. airplane mode / relay unreachable).
3. Tap the failed outgoing message bubble to bring up the recovery prompt.

## Actual Result
- **Layout**: the SnackBar renders as a `Row` with the info text in an `Expanded` and the two action buttons (`Erneut senden` / `Nachricht löschen`) taking intrinsic width. The longer German labels squeeze the info text into a narrow, wrapped column that's hard to read.
- **Timing**: the SnackBar auto-dismisses after the app's default 4s, which isn't enough time to read two options and decide — worse the longer the locale's copy is. The prompt can disappear before the user acts.

## Expected Result
The recovery prompt gives its copy full width regardless of locale and stays open until the user picks Resend or Delete, instead of auto-dismissing on a timer.

## Regression?
No — this is how the recovery prompt has behaved since it was introduced as a SnackBar; it was never locale-tested against longer strings.

## Evidence
Found during manual verification on an iPhone 17 Pro Max simulator with the device locale set to German: the SnackBar crowded the info text into a narrow wrapped column, and the auto-dismiss timer competed with reading time.

realmeylisdev:
## Description

The failed-DM-send recovery UI (tap a failed own bubble → Resend / Delete
message) was a `DivineSnackbarContainer` SnackBar. Two problems with that,
most visible in longer-string locales like German:

- Layout: the SnackBar is a `Row` with the info text in an `Expanded` and
  the two action buttons taking intrinsic width. Longer action labels (e.g.
  German "Erneut senden" / "Nachricht löschen" vs English "Resend" /
  "Delete message") squeeze the info text into a narrow, wrapped column.
- Timing: SnackBars in this app auto-dismiss after 4s by default, which is
  a bad fit for a prompt that requires reading two options and deciding —
  worse the longer the copy is.

Swapped to `VineBottomSheetPrompt` (the existing prompt-style bottom sheet
already used for confirm/discard flows elsewhere, e.g.
`library_trash_screen.dart`, `video_editor_main_overlay_actions.dart`).
It gives the copy full width and stays open until the user picks an
action. Same bloc events dispatched on each choice
(`ConversationFullSendRecoveryRequested` / `ConversationOutgoingSendCancelled`),
same accessibility announcement, same retry-batch/undelivered-sibling
targeting logic — only the presentation changed.

Added a new `dmSendFailedSubtitle` l10n string (translated into all 18
locales) since the bottom sheet needs a title + subtitle, where the
SnackBar only had a single line of text.

**Related Issue:** Fixes #6092

## Out of Scope

The blocked-message toast and the partial-delivery-retry toast elsewhere
in `_onSendOutcome` are unrelated flows and were intentionally left as
SnackBars (single message + single action, no crowding/timing concern).

## Verification

- `flutter test test/screens/inbox/conversation/conversation_view_test.dart` — 33/33 pass (updated for the sheet's animated dismiss instead of the SnackBar's immediate dismiss)
- `flutter test test/l10n/arb_consistency_test.dart` — pass
- `dart format` — clean
- Manually verified on an iPhone 17 Pro Max simulator with the device locale set to German — the bottom sheet renders full-width, readable, no crowding.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/l10n/app_en.arb
+++ b/mobile/lib/l10n/app_en.arb
@@ -3140,7 +3140,7 @@
   "@dmSendFailedMessage": {
     "description": "Accessibility announcement text, and the recovery bottom sheet's title, shown in a DM conversation when a send fails (relay error, signer error, network error). The bottom sheet pairs it with `dmSendFailedSubtitle` and the `dmMessageActionRetrySend` / `dmMessageActionCancelSend` actions."
   },
-  "dmSendFailedSubtitle": "Resend it now, or delete it from the conversation.",
+  "dmSendFailedSubtitle": "Resend it now, or stop trying.",
   "@dmSendFailedSubtitle": {
     "description": "Subtitle in the recovery bottom sheet shown when a failed own DM bubble is tapped, below the `dmSendFailedMessage` title."
   },
--- a/mobile/lib/l10n/app_en.arb
+++ b/mobile/lib/l10n/app_en.arb
@@ -3138,7 +3138,11 @@
   },
   "dmSendFailedMessage": "Message couldn't be sent",
   "@dmSendFailedMessage": {
-    "description": "SnackBar text shown in a DM conversation when a send fails (relay error, signer error, network error). Paired with the retry action `dmSendFailedRetry`."
+    "description": "Accessibility announcement text, and the recovery bottom sheet's title, shown in a DM conversation when a send fails (relay error, signer error, network error). The bottom sheet pairs it with `dmSendFailedSubtitle` and the `dmMessageActionRetrySend` / `dmMessageActionCancelSend` actions."
+  },
+  "dmSendFailedSubtitle": "Resend it now, or delete it from the conversation.",
+  "@dmSendFailedSubtitle": {
+    "description": "Subtitle in the recovery bottom sheet shown when a failed own DM bubble is tapped, below the `dmSendFailedMessage` title."
   },
   "dmSendFailedRetry": "Retry",
   "@dmSendFailedRetry": {
@@ -3190,11 +3194,11 @@
   },
   "dmMessageActionRetrySend": "Resend",
   "@dmMessageActionRetrySend": {
-    "description": "Action in the snackbar shown when a failed own DM bubble is tapped; re-attempts delivery of that queued message."
+    "description": "Primary action in the recovery bottom sheet shown when a failed own DM bubble is tapped; re-attempts delivery of that queued message."
   },
   "dmMessageActionCancelSend": "Stop trying",
   "@dmMessageActionCancelSend": {
-    "description": "Action in the snackbar shown when a failed own DM bubble is tapped; stops retrying the undelivered recipients and keeps the message (and any already-delivered copies). It does not delete the message — long-press delete does that."
+    "description": "Secondary action in the recovery bottom sheet shown when a failed own DM bubble is tapped; gives up on the queued message and removes it."
   },
   "dmReactionAddCustomA11yLabel": "Add custom emoji reaction",
   "@dmReactionAddCustomA11yLabel": {
--- a/mobile/lib/screens/inbox/conversation/conversation_view.dart
+++ b/mobile/lib/screens/inbox/conversation/conversation_view.dart
@@ -46,6 +46,10 @@ class ConversationView extends ConsumerStatefulWidget {
   ConsumerState<ConversationView> createState() => _ConversationViewState();
 }
 
+/// The choice made on the recovery bottom sheet opened by
+/// [_ConversationViewState._onFailedMessageTap].
+enum _FailedMessageAction { resend, delete }
+
 class _ConversationViewState extends ConsumerState<ConversationView> {
   Future<void> _onOptions(String otherPubkey, String displayName) async {
     if (otherPubkey.isEmpty) return;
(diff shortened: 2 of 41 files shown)


mobile/lib/l10n/app_en.arb
```diff
@@ -3138,7 +3138,11 @@
   },
   "dmSendFailedMessage": "Message couldn't be sent",
   "@dmSendFailedMessage": {
-    "description": "SnackBar text shown in a DM conversation when a send fails (relay error, signer error, network error). Paired with the retry action `dmSendFailedRetry`."
+    "description": "Accessibility announcement text, and the recovery bottom sheet's title, shown in a DM conversation when a send fails (relay error, signer error, network error). The bottom sheet pairs it with `dmSendFailedSubtitle` and the `dmMessageActionRetrySend` / `dmMessageActionCancelSend` actions."
+  },
+  "dmSendFailedSubtitle": "Resend it now, or delete it from the conversation.",
```
hm21: [P1] "delete it from the conversation" over-promises. The secondary action still dispatches `ConversationOutgoingSendCancelled` and the button text is unchanged ("Stop trying"). For a partially-delivered group send that only cancels the *undelivered* siblings — the persisted bubble that already reached a recipient stays (conversation_view_test.dart:1189 asserts exactly this, and `cancelOutgoingSend`'s own doc only drops the queue row). The original copy deliberately avoided "delete" for that case. Make the subtitle neutral ("resend now, or stop trying") across all 18 locales rather than promising a delete that doesn't always happen.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -562,6 +539,31 @@ class _MessageList extends StatelessWidget {
       l10n.dmSendFailedMessage,
       Directionality.of(context),
     );
+
+    final action = await VineBottomSheetPrompt.show<_FailedMessageAction>(
```
hm21, line 546: [P2] Issue #6092 requires the prompt to stay open until the user picks Resend or Delete. This `.show` inherits `isDismissible: true` and `enableDrag: true`, so a barrier tap or a downward drag dismisses the sheet → `show` returns null → no action. `VineBottomSheetPrompt.show` currently forwards only `isDismissible` (not `enableDrag`), so `isDismissible: false` here is necessary but not sufficient — drag-to-dismiss and the drag handle stay active. Thread `enableDrag` through the helper and set both false, with a divine_ui test covering non-dismissibility (package is coverage-gated at 100%).

hm21 (commented):
Changes requested. The snackbar→bottom-sheet swap is the right call and the bloc events are unchanged, but two things block. Verified locally: `flutter analyze` clean, conversation_view (33) + vine_bottom_sheet_prompt suites green, `dmSendFailedSubtitle` translated in all 18 locales.

**[P1] Misleading "delete" copy** — see inline on the subtitle.

**[P2] Sheet is still dismissible without a choice** — see inline on the `.show` call. Issue #6092's expected result is that the prompt "stays open until the user picks Resend or Delete"; the inherited defaults `isDismissible: true` + `enableDrag: true` let a barrier tap or a drag dismiss it (→ null → no action taken). Note that `VineBottomSheetPrompt.show` only forwards `isDismissible`, not `enableDrag`, so a call-site `isDismissible: false` alone is not enough — the fix has to thread `enableDrag` through the divine_ui helper (+ a package test, since divine_ui is 100%-coverage-gated).

Nits: `conversation_view.dart:682` comment still says "divine snackbar" (stale, it's a bottom sheet now); the branch carries a `Merge branch 'main'` commit where the repo rule is rebase.

mobile/lib/l10n/app_en.arb
realmeylisdev: Fixed in 630d58ebb8. Subtitle is now neutral in all 18 locales — EN: "Resend it now, or stop trying." — matching the Resend / Stop trying buttons, so it no longer promises a delete that doesn't happen for a delivered group sibling.

mobile/lib/screens/inbox/conversation/conversation_view.dart
realmeylisdev: Fixed in 630d58ebb8. Threaded `enableDrag` through `VineBottomSheetPrompt.show` and set both `isDismissible: false` and `enableDrag: false` at the call site, so neither a barrier tap nor a drag can dismiss it — it stays open until the user picks Resend or Stop trying (#6092). Added a divine_ui non-dismissibility test; the prompt file stays at 100% line coverage.

realmeylisdev:
@hm21 all four addressed in 630d58ebb8:

- **[P1]** `dmSendFailedSubtitle` is now neutral in all 18 locales ("Resend it now, or stop trying."), so it no longer promises a delete that doesn't happen for an already-delivered group sibling.
- **[P2]** Threaded `enableDrag` through `VineBottomSheetPrompt.show` and set both `isDismissible: false` + `enableDrag: false` at the call site — barrier tap and drag can no longer dismiss it (#6092). Added a divine_ui non-dismissibility test; the prompt file stays at 100% line coverage.
- Nit: stale "divine snackbar" comment → "recovery bottom sheet".
- Nit: dropped the `Merge branch 'main'` commit — rebased onto `origin/main`, history is linear now.

Verified locally: `flutter analyze` clean, divine_ui prompt suite + `conversation_view` (38) green, gen-l10n in sync.

hm21 (approved):
LGTM

PR #6084, opened 2026-07-14
fix(upload): re-drive stuck uploads on app resume and startup

### LINKED ISSUE #6085: fix(upload): re-drive stuck uploads on app resume and startup
## Problem

Uploads can remain stuck in `uploading`/`retrying` after app suspension or process restart. The live-process failure happens when the upload retry policy is waiting on exponential backoff; iOS suspension pauses the Dart timer, so the retry may not continue promptly when the app returns to the foreground. Separately, a crash or force-quit can leave persisted upload rows with no live upload future to drive them.

## Expected behavior

- On app resume, a live upload parked in retry backoff should continue promptly without starting a duplicate Dart upload pipeline.
- On startup/resume, persisted `uploading`/`retrying` rows with no live upload future should be re-driven.
- Missing local source files should move to `failed` so the publish flow can surface the problem.
- User-driven retry and recovery should be single-flight per upload id.

## Fix

PR #6084 makes `UploadManager` background-aware, adds a startup/resume recovery sweep, makes upload retry backoff wakeable by upload id, and gates upload execution through a per-upload single-flight future map.

## Verification

Covered by upload recovery/resumable/retry-policy tests, broader upload-manager tests, background activity manager tests, and Mobile CI.

rabble:
Closes #6085

## What & why

When posting videos on a flaky connection, uploads could stall after backgrounding the app. Support logs (`openvine_full_logs_2026-07-14`) pointed at retry backoff waiting on a Dart `Timer`; while iOS suspends the app, that timer does not make progress. The app also had no startup sweep for uploads left in `uploading`/`retrying` after crash or force-quit.

This addresses Root Cause #1: live retry backoff now wakes on app resume, and persisted `uploading`/`retrying` rows are recovered at startup/resume when no live upload future owns them. It does not address Root Cause #2 (byte-level resume for the background PUT path), which remains a separate larger upload-transport change.

## Changes

- `UploadManager` implements `BackgroundAwareService` and registers idempotently after normal initialization and lazy store self-heal. It unregisters only if it actually registered.
- Recovery runs on app resume and once at startup:
  - if the upload is live and parked in retry backoff, the existing backoff wait is woken so the same `_performUpload` future continues;
  - if no live upload future exists, the persisted `uploading`/`retrying` row is re-driven;
  - missing local source files are marked `failed` so the publish flow can surface the problem.
- Upload driving is single-flight per upload id. Recovery, manual retry, and reusable-upload paths now join the same in-flight future instead of starting overlapping Dart upload pipelines.
- The attempted Riverpod foreground-stream bridge was removed; `BackgroundActivityManager.onAppResumed()` is the resume trigger.

## Scope decisions

- Only `uploading`/`retrying` are recovered. `failed` uploads remain user-driven retry flow because they carry manual retry budget.
- Startup auto-resume intentionally changes the old deferred behavior for interrupted uploads; the corresponding test is updated.
- Root Cause #2, connectivity-gated retry, and resumable background transport remain out of scope.

## Verification

- `flutter test test/services/upload_manager_recovery_test.dart test/services/upload_manager_resumable_upload_test.dart test/services/upload/upload_retry_policy_test.dart` — 67 passed.
- `flutter analyze lib/services/upload_manager.dart lib/services/upload/upload_retry_policy.dart test/services/upload_manager_recovery_test.dart test/services/upload_manager_resumable_upload_test.dart test/services/upload/upload_retry_policy_test.dart` — no issues.
- `flutter test test/services/upload_manager_*_test.dart test/services/upload/` — 259 passed, 11 skipped.
- `flutter test test/services/background_activity_manager_test.dart` — 6 passed.
- Pre-push hook on `git push --force-with-lease origin fix/upload-bg-recovery` passed analyzer and changed-file tests.


mbradley:
Grabbing this for a thorough review pass alongside #6086. Findings to follow.

mobile/lib/services/upload_manager.dart
```diff
@@ -1261,6 +1325,115 @@ class UploadManager {
...
+          .toList();
+
+      if (stuck.isEmpty) return;
+
+      var recovered = 0;
+      var markedMissingFile = 0;
+      for (final upload in stuck) {
+        // Skip uploads that are genuinely active right now — only re-drive
+        // the ones no live [_performUpload] future is driving.
+        if (_inFlightUploadIds.contains(upload.id)) continue;
```
hm21: **P1 / blocker: this `continue` skips the exact case the PR is built around.**

iOS suspension doesn't kill the isolate — `_performUpload` is still parked on the backoff `Timer` (`async_utils.dart`: `Timer(delay, completer.complete); await completer.future`). Line 759 added the id and the only `remove` is the `finally` at 829, which hasn't run. So on resume the record is still `uploading`/`retrying`, lands in `stuck`, and gets skipped right here. `recovered=0`, and the upload moves only when the original Timer eventually fires — i.e. exactly the pre-PR behaviour that produced the 26-minute stall.

Codex frames the dilemma well: **if the Timer stays frozen the PR changes nothing; if it fires on resume anyway, the sweep wasn't needed.** Either way this branch is inert in a live process.

The sweep only does work when the set is empty, which in practice means a fresh process. That's genuinely valuable — but it's crash/force-quit recovery, not the resume fix the description sells. To actually cover RC#1 you'd need to track backoff-waiting ids separately and cancel/short-circuit the pending Timer on resume.

Missing regression test: retriable error → backoff active → `onAppResumed()`. Every test here seeds Hive directly, so the in-flight set is always empty.

mobile/lib/services/upload_manager.dart
```diff
@@ -1261,6 +1325,115 @@ class UploadManager {
...
+              errorMessage:
+                  'Video file missing after restart: '
+                  '${upload.localVideoPath}',
+            ),
+          );
+          markedMissingFile++;
+          continue;
+        }
+
+        resumeInterruptedUpload(upload.id);
```
hm21, line 1401: **P2: `resumeInterruptedUpload` has no in-flight guard**, and it's the same primitive the user-driven path reaches — `video_publish_service.dart:500` calls it for exactly the `uploading`/`retrying` records this sweep drives. So while startup recovery is running, "Try Again" can start a second `_performUpload` for the same id. This is the collision the comment you deleted in `initialize()` was documenting.

Scoping it honestly (credit to the Codex pass): the native layer dedupes the *transfer* — `BackgroundUploaderPlugin.swift:78` skips enqueue when a task with the same `taskDescription` is already running — so this isn't double bytes on the background path. What does double up is the Dart side: both pipelines process the terminal event and can repeat thumbnail/status work. (The in-process path, `useBackgroundUpload: false`, has no such dedupe.)

The marker at 759 documents state but doesn't gate anything. A real single-flight map (id → in-flight `Future`) or a guard at the `_performUpload` entry would fix this and the desync below in one move.

mobile/lib/services/upload_manager.dart
```diff
@@ -763,6 +826,7 @@ class UploadManager {
       await _handleUploadFailure(upload, e);
     } finally {
       _userStoppedUploadIds.remove(upload.id);
+      _inFlightUploadIds.remove(upload.id);
```
hm21: `Set.remove` isn't refcounted. With two overlapping `_performUpload` runs for one id (reachable via the missing guard at 1389), the first run's `finally` clears the marker while the second is still uploading — and a later sweep then starts a third. The same line drops `_userStoppedUploadIds`, so a user's pause marker disappears mid-flight and the paused status gets overwritten with `failed`.

A single-flight map keyed by id would make both this and 1389 fall out.

mobile/lib/services/upload_manager.dart
```diff
@@ -695,6 +753,11 @@ class UploadManager {
     final startTime = DateTime.now();
     final videoFile = File(upload.localVideoPath);
 
+    // Mark this upload as actively driven on the event loop while the
+    // transfer runs, so [_recoverStuckUploads] can tell "genuinely in flight"
+    // from "stuck in `uploading` status after a crash". Removed in `finally`.
+    _inFlightUploadIds.add(upload.id);
```
hm21: **P3:** this `add` sits outside the `try` (801) whose `finally` (827-830) holds the only `remove`, so the missing-file early return at 779 strands the id in the set for the manager's lifetime — after which every sweep skips it. The file check should sit inside the outer `try/finally`.

mobile/lib/services/upload_manager.dart
```diff
@@ -226,11 +269,26 @@ class UploadManager {
...
+        if (foreground) {
+          unawaited(_recoverStuckUploads());
+        }
+      });
+
+      // Re-drive uploads left in `uploading`/`retrying` by a prior crash or
+      // background freeze. Fire-and-forget: this runs on the event loop and
+      // must not block startup. `failed` uploads are intentionally left to the
+      // user-driven retry flow (they carry a manual retry budget).
+      unawaited(_recoverStuckUploads());
```
hm21, line 280: This reverses a deliberate product decision (the removed comment) and inverts the test that pinned it, in the same diff. Beyond the double-drive risk above: auto-resuming at every cold start also spends cellular data without asking, which is part of what the bottom-sheet flow existed to gate. Worth an explicit sign-off from whoever owns that decision rather than landing it as a side effect.

mobile/lib/services/upload_manager.dart
```diff
@@ -226,11 +269,26 @@ class UploadManager {
...
 
-      // Interrupted resumable uploads are NOT auto-resumed here.
-      // The bottom sheet flow (resumePendingPublishes → BackgroundPublishBloc)
-      // lets the user decide whether to retry or save to drafts.
-      // VideoPublishService._getOrCreateUpload reuses existing PendingUpload
-      // records (and their resumable sessions) when the user taps "Try Again".
+      // Register for background-lifecycle callbacks so [onAppResumed] can
+      // re-drive uploads whose retry backoff froze while the app was
+      // suspended (Dart Timers do not fire while iOS suspends the app).
+      _backgroundActivityManager.registerService(this);
```
hm21: Registration only happens on the `initialize()` happy path, which makes the whole mechanism fragile:

- `uploadManagerProvider` watches `currentAuthStateProvider` / `currentPublicKeyHex`, so every auth flip builds a fresh `UploadManager`. Nothing eagerly initializes it (the startup-coordinator entry at `main.dart:761` is one-shot), so after sign-out/sign-in the new instance is never registered and its foreground controller has no subscriber — recovery is inert until someone happens to publish.
- `startUpload`'s self-heal (~552) sets `_isInitialized = true` via `_store.ensureOpen()` without registering; a later `initialize()` then early-returns on that flag, so the manager stays unregistered for the session.

mobile/lib/services/upload_manager.dart
```diff
@@ -1261,6 +1325,115 @@ class UploadManager {
...
+  /// Only `uploading`/`retrying` are re-driven — those states are
+  /// unambiguously interrupted and have consumed no manual-retry budget.
+  /// `failed` uploads are left to the user-driven retry flow (they carry a
+  /// `retryCount < 3` budget surfaced via the publish bottom sheet).
+  ///
+  /// Called on app resume ([onAppResumed]), on the app-foreground stream edge
+  /// ([initialize]), and once at startup. Re-entrancy is guarded by
+  /// [_isRecovering].
+  Future<void> _recoverStuckUploads() async {
+    if (_isRecovering) return;
```
hm21, line 1358: `_isRecovering` is dead code on the re-drive path. There's no `await` between here and `finally { _isRecovering = false; }` — the only one is in the missing-file branch — so the body runs synchronously to completion and a second sweep never observes the latch as set. What actually prevents double-driving is the `_inFlightUploadIds` skip at 1369, since `_performUpload` adds the id synchronously before its first await.

mobile/lib/services/upload_manager.dart
```diff
@@ -174,16 +190,43 @@ class UploadManager {
...
+
   bool _isInitialized = false;
 
   /// Check if the upload manager is initialized
   bool get isInitialized => _isInitialized && _store.isReady;
 
+  /// Whether an upload's [_performUpload] future is currently on the event
+  /// loop. Visible for testing the recovery sweep's in-flight skip logic.
+  @visibleForTesting
+  bool isUploadInFlight(String uploadId) =>
```
hm21, line 212: `isUploadInFlight` (218) and `isRecovering` (224) are read by no test, despite the doc comments saying that's why they exist. Either drop them or use them — the in-flight skip at 1369 is the load-bearing branch here and is currently untested.

mobile/lib/providers/upload_media_providers.dart
```diff
@@ -270,12 +272,25 @@ UploadManager uploadManager(Ref ref) {
   ref.watch(currentAuthStateProvider);
   final currentPubkey = ref.watch(authServiceProvider).currentPublicKeyHex;
   final env = ref.read(currentEnvironmentProvider);
+
+  // Bridge the app-foreground notifier into a plain Stream<bool> so the
+  // manager can re-drive stuck uploads on resume without taking a Riverpod
+  // dependency (same pattern as the DM retry services in social_providers).
+  final foregroundController = StreamController<bool>();
+  ref.onDispose(foregroundController.close);
+  ref.listen<bool>(appForegroundProvider, (_, next) {
```
hm21: Two things on this bridge:

1. The controller isn't broadcast, and `initialize()` is reachable twice (`main.dart:761` via the startup coordinator, `video_publish_service.dart:586` lazily), guarded only by `isInitialized`. A publish that lands while `_store.open()` is still awaiting slips past that guard and calls `.listen()` a second time → `StateError: Stream has already been listened to`, which the `initialize()` catch swallows into `_isInitialized = false`. `_foregroundSubscription ??= ...` or a broadcast controller closes it.
2. The subscription's stated rationale — letting a sweep run "if the store was not yet ready when the first resume fired" — can't happen, since it's created after the store-ready check. It fires on the same `resumed` edge as `onAppResumed`, so it's a duplicate trigger carrying a whole Riverpod→Stream bridge for no behavioural gain.

mobile/test/services/upload_manager_recovery_test.dart
```diff
@@ -0,0 +1,342 @@
...
+            resumableSession: any(named: 'resumableSession'),
+            onResumableSessionUpdated: any(named: 'onResumableSessionUpdated'),
+            onProgress: any(named: 'onProgress'),
+          ),
+        );
+      },
+    );
+
+    test(
+      're-entrancy guard prevents overlapping sweeps from double-driving',
```
hm21: This test can't fail for the reason it names. Verified by mutation:

- disable the `_isRecovering` latch → still passes
- remove the `_inFlightUploadIds` skip (latch intact) → fails

So it pins the in-flight skip, not the re-entrancy guard. Also `expect(upload.id, isNotEmpty)` is a tautology.

More broadly: `seedUpload` writes straight into Hive, so every test here has an empty in-flight set — i.e. the crash/restart shape. Nothing exercises a sweep against a live, backoff-parked `_performUpload`, which is why the gap at 1369 reads as covered.

hm21 (commented):
**Request changes.** P1: the startup sweep is a real fix for crash/force-quit recovery, but the resume path doesn't do what the description claims — `_inFlightUploadIds` skips exactly the backoff-frozen upload Root Cause #1 describes, so `onAppResumed` and the foreground bridge can't re-drive it. Codex puts the dilemma well: if the Timer stays frozen the PR changes nothing; if it fires on resume anyway, the sweep wasn't needed.

Rest inline: no in-flight guard on `resumeInterruptedUpload` (P2), the marker leak on the missing-file return (P3), and registration hanging off the `initialize()` happy path.

Verified by mutation: the re-entrancy test passes with `_isRecovering` fully disabled, and fails only when the in-flight skip is removed — it pins a different mechanism than its name says. Head `f9418c3`, all checks green, 24/24 local tests pass.

rabble (commented):
**Reviewed with full checkout + verification.** ✅ Looks good (posting as a comment since it's my own PR).

**Duplicate-publish risk is genuinely low** — traced the full path, cannot cause a duplicate kind-34236 Nostr publish, for three independent reasons:
1. UploadManager never publishes. A successful `_performUpload` sets status to `readyToPublish`/`processing` — it only pushes media to Blossom. The Nostr publish is a separate draft-driven step (`resumePendingPublishes` → `BackgroundPublishBloc` on `DivineVideoDraft` records), not triggered by this sweep.
2. Blossom is content-addressed (sha256), so re-uploading the same file is idempotent at the storage layer.
3. In-flight guard: `_performUpload` synchronously adds the id to `_inFlightUploadIds` before its first await; the sweep skips any id already in that set. The `_isRecovering` latch prevents overlapping sweeps. The re-entrancy test pins "driven exactly once, not once per sweep."

The removed comment conflated two lifecycle phases: `resumePendingPublishes` acts on the *publish* phase (drafts); this sweep acts on the *upload* phase (`PendingUpload` in `uploading/retrying`). Different stores, different stages — complementary, not competing. Old behavior left interrupted media uploads stuck forever; this is the legitimate fix and does not reach the publish phase.

Verification (local):
- `flutter analyze lib/providers lib/services test` → No issues found!
- `flutter test upload_manager_recovery_test.dart upload_manager_resumable_upload_test.dart` → 24 passed.

`.g.dart` OK — only the provider hash changed, consistent with a body-only edit. Provider bridge mirrors the proven DM-retry `StreamController<bool>` + `ref.listen(appForegroundProvider)` pattern with `onDispose`/`isClosed` guards.

Nit: `upload_manager.dart:1396` log counts user-stopped skips as "in-flight" — cosmetic, relabel "skipped".

⚠️ **Collides with #6086** — both touch `upload_manager.dart` and `upload_manager_resumable_upload_test.dart`. This PR rewrites `initialize does not auto-resume interrupted uploads` into `initialize auto-resumes…` (resumable_upload_test.dart:599); #6086 also edits that area. They now assert opposite behaviors, so whoever lands second must rebase and reconcile the test intent. Worth deciding merge order (or combining) before landing either.

NotThatKindOfDrLiz:
Takeover update pushed in `744522cff`.

What changed:
- Replaced the loose `_inFlightUploadIds` set with a per-upload single-flight future map, so recovery and user retry join the same Dart upload pipeline instead of racing.
- Made upload retry backoff wakeable by upload id. `onAppResumed()` now wakes a live backoff wait instead of skipping it, while crash/force-quit rows with no live future are still re-driven.
- Moved background registration behind an idempotent helper and also calls it from the lazy store self-heal path.
- Removed the Riverpod foreground-stream bridge and stale generated hash churn.
- Added regressions for live backoff resume, single-flight `resumeInterruptedUpload`, and missing-file in-flight cleanup.

Local verification:
- `flutter test test/services/upload_manager_recovery_test.dart test/services/upload_manager_resumable_upload_test.dart test/services/upload/upload_retry_policy_test.dart` — 67 passed.
- `flutter analyze lib/services/upload_manager.dart lib/services/upload/upload_retry_policy.dart test/services/upload_manager_recovery_test.dart test/services/upload_manager_resumable_upload_test.dart test/services/upload/upload_retry_policy_test.dart` — no issues.
- `flutter test test/services/upload_manager_*_test.dart test/services/upload/` — 259 passed, 11 skipped.
- `flutter test test/services/background_activity_manager_test.dart` — 6 passed.
- Pre-push hook passed analyzer and changed-file tests.

CI is running on the pushed head now.

NotThatKindOfDrLiz (approved):
Approved after takeover.

I verified the previous blockers are addressed:
- live retry-backoff resume now wakes the existing backoff wait instead of being skipped as in-flight;
- upload driving is single-flight per upload id, covering recovery/user retry overlap and marker desync;
- missing-file exits no longer strand an in-flight marker;
- the foreground-stream bridge was removed, and background registration is idempotent including the lazy initialization path.

Local verification and CI are green on `744522cff`.

PR #5590, opened 2026-06-27
feat(upload): OS-backed background uploads that survive app suspension

hm21:
Closes #5591

## Description

Video publishing fails when the app is backgrounded mid-upload: on iOS the OS
tears down the in-process upload socket on suspension, every retry restarts from
byte 0, and the publish dies after all retries (seen in production logs as
repeated `Connection reset by peer` / `Broken pipe` across 6 attempts). This PR
makes the transfer the OS's responsibility so it survives suspension, plus a
related fix so transient failures stop discarding upload progress. Three
delineated pieces:

1. **`background_uploader` plugin (new package)** — a generic data-layer Flutter
   plugin that hands a file + URL + headers to the OS background transfer
   facility. Darwin (iOS + macOS) background `URLSession` over shared source with
   **SPM + CocoaPods fallback**; Android foreground service with a streamed PUT.
   No Blossom/Nostr knowledge — callers build the signed request.

2. **Transient-failure classification in the background path
   (`blossom_upload_service`)** — `uploadVideoInBackground` runs transport
   errors through the existing `_classifyUploadException`, so a suspended-socket
   drop is reported as `network` + `isTransientNetworkFailure` and treated as
   retryable, matching the in-process path. (The legacy-PUT-fallback resume fix
   that stops re-uploading from byte 0 on a recoverable resumable failure
   already landed on `main`; this PR only consumes that classification — it is
   not part of this diff.)

3. **Background-upload transport + integration** — a `BlossomBackgroundTransport`
   port (keeps the package free of the plugin dependency) + a
   `uploadVideoInBackground()` that does a single BUD-01 `PUT` through the OS
   uploader, reusing the existing auth/ProofMode/hash logic. Wired into
   `UploadManager` behind a `useBackgroundUpload` flag (currently enabled);
   thumbnail, retry, and success handling are unchanged.

**Related Issue:** 

## Out of Scope

- **App-kill reconciliation.** The `await`-based path survives app *suspension*
  (the logged failure). If the app is fully *terminated* mid-upload the OS still
  finishes the transfer, but the follow-up Nostr publish then needs
  reconciliation on the next launch (via `BackgroundUploader.activeTaskIds()` +
  the existing `PendingUpload` state). Not included here.

## Verification

- `flutter analyze` (full `lib`/`test` + both packages) → no issues
- `blossom_upload_service`: **198** tests (incl. 6 new background-upload tests +
  the resume-fix tests)
- `UploadManager`: **102** tests (flag defaults off in tests)
[...]

Issue #5151, opened 2026-06-14
test(pipeline): track production upload pipeline acceptance gaps


NotThatKindOfDrLiz:
## Problem

PR #5142 restores local-stack manual acceptance coverage for the upload -> Blossom -> kind-34236 publish -> relay/API query path. That is valuable, but it does not fully cover production-environment behavior or cleanup/SLO policy.

## Remaining coverage gaps

- Production-environment assertion for the real media pipeline, including the production CDN/data-plane host path.
- Real transcode quality or readiness behavior beyond the local Viceroy `.hls` byte-copy stub.
- A defined latency/throughput SLO for upload, publish, indexing, and read-back.
- Orphan-blob behavior when publish fails after upload succeeds, including the expected cleanup contract and observable failure mode.

## Acceptance

- A production-safe E2E or backend-owned acceptance test covers upload -> CDN URL -> kind-34236 event -> relay/API query without depending on removed `api.openvine.co` tests.
- The transcode/readiness assertion validates production behavior, not just the local-stack readiness stub.
- Stage timing expectations are documented with an agreed budget before they are enforced.
- Publish-failure cleanup behavior is specified and covered, or explicitly documented as unsupported with owner sign-off.

## Notes

This issue tracks the broader backend/platform hardening that is intentionally out of scope for the local-stack manual test added in PR #5142.

NotThatKindOfDrLiz:
Part of #4337.

NotThatKindOfDrLiz:
Closing as superseded after decomposing the production/platform gaps into their owning repository:

- divinevideo/divine-blossom#282 — production-safe upload → readiness → kind-34236 → relay/REST acceptance
- divinevideo/divine-blossom#283 — production transcode/readiness assertion
- divinevideo/divine-blossom#284 — agreed stage-level service objectives before timing enforcement
- divinevideo/divine-blossom#163 — existing orphan-blob reclamation contract (already assigned)

The local-stack client acceptance test in this repository remains useful and correctly scoped. The obsolete `api.openvine.co` wording was not carried forward. This issue was also removed from #8290 because production media-pipeline acceptance is outside that epic’s test-suite-health scope.

Issue #1226, opened 2026-02-02
feat: Add pending uploads UI to view and manage upload queue


rabble:
## Summary

Users with failed uploads accumulate a backlog of pending uploads with no visibility or way to manage them. Need UI to show pending uploads and allow retry/clear actions.

## User Impact

From logs (`openvine_full_logs_2026-02-01T20-18-34.194367.txt`):
```
[20:10:30.104196] UploadManager initialized successfully with 43 existing uploads
```

User had **43 pending uploads** queued up with no awareness and no way to:
- See the backlog exists
- Retry failed uploads
- Clear stuck/unwanted uploads
- Understand why uploads failed

## Current Behavior

- `UploadManager` persists uploads to Hive storage
- Failed uploads stay in queue indefinitely
- Uploads retry on app restart but may keep failing (e.g., poor network)
- No UI shows upload status or queue depth
- Users don't know their videos didn't publish

### 1. Pending Uploads Screen

Add a screen accessible from profile/settings showing:
- List of pending uploads with thumbnail/title
- Upload status (queued, uploading, failed, retry scheduled)
- Failure reason for failed uploads
- Progress indicator for active uploads

### 2. Upload Actions

For each pending upload:
- **Retry Now** - Immediately retry the upload
- **Delete** - Remove from queue (with confirmation)
- **View Details** - Show full error message, timestamps, retry count

### 3. Bulk Actions

- **Retry All** - Retry all failed uploads
- **Clear All Failed** - Remove all failed uploads from queue
- **Clear All** - Nuclear option with strong confirmation

### 4. Notification Badge

Show badge on profile/settings icon when:
- Uploads are pending
- Uploads have failed

### 5. Upload Status Indicator

In the camera/post flow, show:
- "Uploading..." with progress
- "Upload failed - tap to retry" if failed
- Link to pending uploads screen

### Existing Infrastructure

```dart
// UploadManager already tracks uploads in Hive
[20:10:30.104196] UploadManager initialized successfully with 43 existing uploads
```

The `UploadManager` already has:
- Persistent storage (`uploads` Hive box)
- Retry logic with exponential backoff
- Upload status tracking

### What's Needed

1. **Provider** to expose upload queue state to UI:
```dart
@riverpod
Stream<List<PendingUpload>> pendingUploads(Ref ref) {
  final uploadManager = ref.watch(uploadManagerProvider);
  return uploadManager.uploadsStream;
}
```

2. **Screen** to display and manage uploads:
[...]

PR #6086, opened 2026-07-14
feat(upload): resume interrupted publishes from the server's last offset

### LINKED ISSUE #6087: feat(upload): resume interrupted publishes from the server's last offset
## Problem

Fresh video publishes use the OS-backed whole-file Blossom upload so the transfer can survive app suspension. If that interrupted publish later needs retry/reconciliation, the existing retry path restarts from byte 0 instead of resuming from the server's last committed offset.

This leaves the app-kill reconciliation gap from #5590 unresolved for large uploads: a user can spend time and bandwidth re-sending bytes the server already accepted.

## Expected

When a resumable Blossom session exists, retry/recovery should query the server's current Upload-Offset and send only the remaining chunks. Fresh uploads should still try the OS-backed transfer first for suspension survival.

## Acceptance criteria

- Fresh publishes keep the OS-backed whole-file upload as the first attempt.
- Genuine OS upload failures fall back to the chunked resumable protocol.
- Existing resumable sessions skip the OS whole-file upload and resume from the server's committed offset.
- User pause/cancel remains terminal and cannot be overwritten by a later upload success.
- Regression coverage pins OS-cancel behavior, resumable timeout behavior, session resume wiring, and paused-state preservation after a late success.

Resolved by PR #6086.

rabble:
Closes #6087

## Summary

A fresh publish attempts the OS-backed whole-file PUT first (it survives app suspension and is fastest on a healthy connection). When that fails — or when a resumable session already exists from a prior attempt — the new `uploadVideoWithResume` falls through to the chunked resumable protocol so retries resume from the server's last committed `Upload-Offset` instead of byte 0.

This fills the explicit gap **#5590 left open** (*"Out of Scope: app-kill reconciliation"*) without reverting the two deliberate decisions that preceded it:
- **#5590** (Jul 1): moved the transfer to the OS background uploader because the in-process socket dies on suspension. This PR keeps the OS PUT as the first attempt for fresh uploads.
- **#2223** (Mar 15): removed silent auto-resume-on-startup after Discord users complained about uploads restarting without control. This PR does **not** re-add auto-resume — the user-decided bottom sheet stays. It only makes the *retry* resume efficiently.

## How it works

```
uploadVideoWithResume(session):
  if session == null:           # fresh publish
    try OS whole-file PUT       # survives suspension (#5590)
    if success → done
  # fall through to resumable (also the only path when session exists)
  POST /upload/init
  HEAD {uploadUrl}              # Upload-Offset → how many bytes the server has
  PUT remaining chunks only     # Content-Range, persisted per chunk
  POST /upload/{id}/complete
```

Once a resumable session is persisted (via the existing `onResumableSessionUpdated` callback), the existing **"Try Again" bottom sheet** automatically resumes from the last offset — no new wiring needed. The retry path already reuses `PendingUpload.resumableSession`.

## Key design constraint

The OS uploader (`background_uploader`) is **whole-file-only** on both platforms (iOS `URLSession.uploadTask(with:fromFile:)`, Android `HttpURLConnection`). Byte-level resume **cannot** go through the OS transport. It always happens through the in-process chunk path. The OS path is only ever used for the first whole-file attempt of a fresh publish.

## Timeout safety

`resumableTimeout` bounds **only** the in-process chunked leg. The OS-backed PUT keeps its own internal 30-min timeout — the aggressive 10-min `networkTimeout` is not applied to it, so an in-process cutoff cannot defeat suspension survival.

## Changes

| File | Change |
|---|---|
| `blossom_upload_service.dart` | New `uploadVideoWithResume` orchestrator: OS-first for fresh, resumable fallback on failure or existing session |
| `upload_manager.dart` | `_executeUploadWithTimeout` calls the combined method; `useBackgroundUpload` becomes the OS-first toggle |
| `blossom_upload_with_resume_test.dart` | 7 new tests covering all branches (OS success, OS fail→resumable, existing-session resume, `useBackgroundFirst=false`, session-callback offsets, OS-throws guard, resumableTimeout) |
| `upload_manager_resumable_upload_test.dart` | New wiring test: verifies combined method receives persisted session + flag; existing stubs updated |
| `upload_manager_from_draft_test.dart`, `upload_manager_background_pause_test.dart` | Mock stubs updated to `uploadVideoWithResume` |
| `2026-03-26-divine-resumable-upload-sessions-bud.md` | Updated mobile client status line |

## Verification

- `flutter analyze` — no issues on all changed files
- `blossom_upload_service` package: **221 tests pass**, **100% coverage** on `blossom_upload_service.dart` (825/825 lines, `min_coverage: 100` satisfied)
- `upload_manager` app tests: all pass (resumable, from-draft, background-pause)
- Regression test pinning: resume-after-failure sends only `fileSize - offset` bytes (the "when a resumable session already exists" test asserts exactly 1 chunk PUT for the tail, not the whole file)

## Related

- Closes the gap left open by #5590 (*"Out of Scope: app-kill reconciliation"*).
- Does not affect #1226 (pending uploads UI) or #5151 (pipeline coverage).

### DIFF
--- a/mobile/lib/services/upload_manager.dart
+++ b/mobile/lib/services/upload_manager.dart
@@ -154,10 +154,14 @@ class UploadManager implements BackgroundAwareService {
     );
   }
 
-  /// Routes the video transfer through the OS background uploader
-  /// ([BlossomUploadService.uploadVideoInBackground]) instead of the in-process
-  /// chunked/legacy path, so it survives app suspension. Requires the
-  /// blossom service to have been built with a background transport.
+  /// Attempts the OS background uploader
+  /// ([BlossomUploadService.uploadVideoInBackground]) *first* for a fresh
+  /// publish — it survives app suspension and is fastest on a healthy
+  /// connection — then falls back to the in-process chunked resumable path if
+  /// that OS attempt fails (a user cancel is treated as terminal and does NOT
+  /// fall back). Set to `false` to skip the OS attempt and go straight to the
+  /// in-process path. Requires the blossom service to have been built with a
+  /// background transport.
   final bool useBackgroundUpload;
 
   // Core services
@@ -1235,16 +1239,19 @@ class UploadManager implements BackgroundAwareService {
       category: LogCategory.video,
     );
 
-    // Mark before tearing down the transfer: cancelling it emits a terminal
-    // event that would otherwise drive the in-flight upload into
-    // [_handleUploadFailure] and overwrite the paused status below.
+    // Mark before tearing down the transfer: cancelling the OS transfer emits
+    // a terminal `cancelled` event that resolves the in-flight upload as a
+    // (terminal, non-retriable) failure, which would otherwise reach
+    // [_handleUploadFailure] and overwrite the paused status below. The marker
+    // lets [_performUpload] recognise the user-initiated stop and skip it.
     _userStoppedUploadIds.add(uploadId);
 
     // Stop the OS-owned background transfer while paused so it doesn't keep
-    // uploading in the background. A background transfer is a single OS-owned
-    // PUT with no resumable checkpoint, so resuming re-enqueues it from byte 0
-    // (unlike the in-process path, which resumes from its resumable session).
-    // The in-process path is torn down by its request timeout.
+    // uploading in the background. With OS-first uploads the transfer may have
+    // already fallen through to the in-process resumable leg; that leg is torn
+    // down by its request timeout, and this cancel is then a harmless no-op on
+    // the already-finished OS task. If the OS leg was still the active one it
+    // has no resumable checkpoint, so resuming re-enqueues it from byte 0.
     if (useBackgroundUpload) {
       await _blossomService.cancelBackgroundUpload(uploadId);
     }
--- a/mobile/lib/services/upload_manager.dart
+++ b/mobile/lib/services/upload_manager.dart
@@ -868,6 +868,15 @@ class UploadManager implements BackgroundAwareService {
         onProgress,
       );
 
+      if (_userStoppedUploadIds.contains(upload.id)) {
+        Log.info(
+          'Upload ${upload.id} stopped by user; skipping success handling',
+          name: 'UploadManager',
+          category: LogCategory.video,
+        );
+        return;
+      }
+
       // Success - record metrics and complete
       await _handleUploadSuccess(currentUpload, result);
     }, isRetriable: _retryPolicy.isRetriableError);
@@ -1004,6 +1013,16 @@ class UploadManager implements BackgroundAwareService {
         rethrow;
       }
 
+      if (_userStoppedUploadIds.contains(upload.id)) {
+        Log.info(
+          'Upload ${upload.id} stopped by user after video transfer; '
+          'skipping thumbnail and success handling',
+          name: 'UploadManager',
+          category: LogCategory.video,
+        );
+        return result;
+      }
+
       // Generate and upload thumbnail after video upload succeeds
       String? thumbnailCdnUrl;
       if (result.success && result.cdnUrl != null) {
@@ -1079,6 +1098,15 @@ class UploadManager implements BackgroundAwareService {
     final metrics = _reporter.metricsFor(upload.id);
 
     if (result.success == true) {
+      if (_userStoppedUploadIds.contains(upload.id)) {
+        Log.info(
+          'Upload ${upload.id} stopped by user; ignoring late success result',
+          name: 'UploadManager',
+          category: LogCategory.video,
+        );
+        return;
+      }
+
       // Get the LATEST upload record from Hive (may have been updated with thumbnail URL)
       final latestUpload = getUpload(upload.id) ?? upload;
 
--- a/mobile/lib/services/upload_manager.dart
+++ b/mobile/lib/services/upload_manager.dart
@@ -941,66 +941,63 @@ class UploadManager implements BackgroundAwareService {
         onProgress?.call(progress);
       }
 
-      // Background uploads are a single OS-owned PUT that survives suspension.
-      // The OS keeps the transfer alive across app suspension, so it must not
-      // be cut off by the aggressive in-process network timeout;
-      // uploadVideoInBackground applies its own generous safety timeout and
-      // releases its event subscription internally. The in-process path keeps
-      // chunked-resumable behind networkTimeout.
-      final BlossomUploadResult result;
-      if (useBackgroundUpload) {
-        result = await _blossomService.uploadVideoInBackground(
+      // Combined upload path: a fresh publish attempts the OS-owned whole-file
+      // PUT first (it survives app suspension and is fastest on a healthy
+      // connection); on failure, or when a resumable session already exists
+      // from a prior attempt, the request falls through to the chunked
+      // resumable protocol so retries resume from the server's last committed
+      // offset instead of byte 0.
+      //
+      // The OS transfer applies its own generous safety timeout internally and
+      // must NOT be cut by the in-process network timeout — doing so would
+      // defeat the suspension survival that is the whole point of the OS path.
+      // Only the in-process chunked-resumable leg is bounded by networkTimeout
+      // (passed as resumableTimeout), so a stalled socket does not hang
+      // forever. `useBackgroundUpload` is the kill-switch for the OS-first
+      // attempt: flip it to false to go straight to chunked resumable.
+      BlossomUploadResult result;
+      try {
+        result = await _blossomService.uploadVideoWithResume(
           videoFile: videoFile,
+          nostrPubkey: upload.nostrPubkey,
           taskId: upload.id,
+          title: upload.title ?? '',
+          description: upload.description,
+          hashtags: upload.hashtags,
           proofManifestJson: upload.proofManifestJson,
+          useBackgroundFirst: useBackgroundUpload,
+          resumableTimeout: _retryConfig.networkTimeout,
+          resumableSession: upload.resumableSession,
+          onResumableSessionUpdated: (session) {
+            _retryPolicy.enqueueSessionPersist(
+              upload.id,
+              session,
+              videoFile.lengthSync(),
+            );
+          },
           onProgress: reportProgress,
         );
-      } else {
-        result = await _blossomService
-            .uploadVideo(
-              videoFile: videoFile,
-              nostrPubkey: upload.nostrPubkey,
-              title: upload.title ?? '',
-              description: upload.description,
-              hashtags: upload.hashtags,
-              proofManifestJson: upload.proofManifestJson,
-              resumableSession: upload.resumableSession,
-              onResumableSessionUpdated: (session) {
-                _retryPolicy.enqueueSessionPersist(
-                  upload.id,
-                  session,
-                  videoFile.lengthSync(),
-                );
-              },
-              onProgress: reportProgress,
-            )
-            .timeout(
-              _retryConfig.networkTimeout,
-              onTimeout: () {
-                Log.error(
-                  '⏱️ Upload timed out!',
-                  name: 'UploadManager',
-                  category: LogCategory.video,
-                );
-                final timeoutError = TimeoutException(
-                  'Upload timed out after '
-                  '${_retryConfig.networkTimeout.inMinutes} minutes',
-                );
+      } on TimeoutException catch (_) {
+        Log.error(
+          '⏱️ Upload timed out!',
+          name: 'UploadManager',
+          category: LogCategory.video,
+        );
+        final timeoutError = TimeoutException(
+          'Upload timed out after '
+          '${_retryConfig.networkTimeout.inMinutes} minutes',
+        );
 
-                // Send timeout crash report asynchronously
-                _reporter
-                    .sendTimeoutCrashReport(upload, timeoutError)
-                    .catchError((e) {
-                      Log.error(
-                        'Failed to send timeout crash report: $e',
-                        name: 'UploadManager',
-                        category: LogCategory.video,
-                      );
-                    });
+        // Send timeout crash report asynchronously
+        _reporter.sendTimeoutCrashReport(upload, timeoutError).catchError((e) {
+          Log.error(
+            'Failed to send timeout crash report: $e',
+            name: 'UploadManager',
+            category: LogCategory.video,
+          );
+        });
 
-                throw timeoutError;
-              },
-            );
+        rethrow;
       }
 
       // Generate and upload thumbnail after video upload succeeds
--- a/mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
+++ b/mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
@@ -110,6 +110,13 @@ enum BlossomUploadFailureReason {
   /// "our servers are temporarily unavailable".
   server,
 
+  /// The user stopped the upload (pause / cancel) while it was in flight.
+  /// User-initiated and terminal: the caller must NOT retry or fall back to
+  /// another transport, and it is NOT reportable to Crashlytics. Distinct
+  /// from [unknown] so a deliberate stop is never mistaken for a transport
+  /// failure that warrants re-uploading the whole file.
+  cancelled,
+
   /// Anything else: unmapped 4xx, malformed responses, configuration
   /// errors, or unexpected exceptions. Caller falls back to a generic
   /// "upload failed" message.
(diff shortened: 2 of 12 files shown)


mbradley:
Grabbing this for a thorough review pass. Findings to follow.

mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
```diff
@@ -2242,6 +2242,122 @@ class BlossomUploadService {
...
+        category: LogCategory.video,
+      );
+      try {
+        final backgroundResult = await uploadVideoInBackground(
+          videoFile: videoFile,
+          taskId: taskId,
+          proofManifestJson: proofManifestJson,
+          onProgress: onProgress,
+        );
+        if (backgroundResult.success) {
```
hm21, line 2320: **Blocking — Pause/Cancel now trigger a full re-upload and can resurrect a cancelled upload.**

This branches only on `backgroundResult.success`, but `_backgroundTransferResult` maps `BlossomBackgroundTransferStatus.cancelled` to `success: false, errorMessage: 'Upload cancelled', failureReason: unknown` — byte-identical to a transport failure. `cancelBackgroundUpload` is the *only* teardown mechanism `pauseUpload`/`cancelUpload` have (`upload_manager.dart:1176`/`:1287`), so a user stop is indistinguishable from a network drop here.

With `useBackgroundUpload = true` (production):

1. User taps Pause. `pauseUpload` adds the id to `_userStoppedUploadIds`, calls `cancelBackgroundUpload`, writes `status: paused`.
2. Transport emits `cancelled` → `uploadVideoInBackground` returns `success: false`.
3. **This line falls through to `runResumable()`** and chunk-uploads the whole file from byte 0 — the user paused, and we start sending 40 MB.
4. It succeeds → `_handleUploadSuccess` takes the `result.success == true` branch (`upload_manager.dart:1004`) → `_store.update(createSuccessfulUpload(...))` overwrites `paused` with success.
5. The `_userStoppedUploadIds` guard at `:750` never fires — it only sits in the `catch`, and nothing threw.

On `origin/main` the `cancelled` result propagated to `_handleUploadSuccess`, hit the `else`, threw `BlossomUploadFailureException`, and the guard preserved the paused status. That path is gone.

Cancel needs to be terminal here, not a fallback trigger — e.g. return early on `failureReason`/status `cancelled` rather than falling through.

mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
```diff
@@ -2242,6 +2242,122 @@ class BlossomUploadService {
...
+          return backgroundResult;
+        }
+        Log.warning(
```
hm21, line 2332: The fall-through never calls `cancelBackgroundUpload(taskId)`, so when the failure came from `uploadVideoInBackground`'s own 30-min `_backgroundUploadTimeout` (line ~2216 — its `onTimeout` only returns a failure result, and the `finally` cancels just the event subscription), the OS transfer is still live. We then stream the same file to the same server over the same link concurrently — metered data billed twice, and the contention makes the timeout likelier to recur. Both native plugins dedupe re-enqueues by `taskId`, but the in-process dio path is invisible to that guard.

A one-line `await cancelBackgroundUpload(taskId);` before falling through fixes it, using an API this PR already has.

mobile/packages/blossom_upload_service/test/src/blossom_upload_with_resume_test.dart
```diff
@@ -0,0 +1,675 @@
...
+            data: any(named: 'data'),
+            options: any(named: 'options'),
+            onSendProgress: any(named: 'onSendProgress'),
+          ),
+        ).called(2);
+      },
+    );
+
+    test(
+      'resumableTimeout bounds only the in-process resumable leg',
```
hm21: **This test cannot fail.** It asserts neither "bounds" nor "only" — it passes a 1-minute timeout to a mock leg that completes in microseconds and then asserts `result.success`, which the test at line 244 already covers.

I verified by mutation: replacing the whole implementation with `return runResumable();` (i.e. `resumableTimeout` completely ignored, deleting the stalled-socket protection `upload_manager.dart:897` depends on) leaves all 7 tests green.

The property worth pinning is the *"only"* half — it's the reason this design exists: make the OS leg outlast `resumableTimeout` and assert the upload still succeeds. That fails if someone wraps the OS leg in the timeout.

mobile/lib/services/upload_manager.dart
```diff
@@ -868,66 +868,63 @@ class UploadManager {
...
+      try {
+        result = await _blossomService.uploadVideoWithResume(
           videoFile: videoFile,
+          nostrPubkey: upload.nostrPubkey,
           taskId: upload.id,
+          title: upload.title ?? '',
+          description: upload.description,
+          hashtags: upload.hashtags,
           proofManifestJson: upload.proofManifestJson,
+          useBackgroundFirst: useBackgroundUpload,
```
hm21, line 981: `useBackgroundUpload` no longer answers "is this transfer OS-owned?" — only "was the OS tried first?". The two guards that consume it as ownership are now wrong:

- `pauseUpload:1175` / `cancelUpload:1286` call `cancelBackgroundUpload` when the flag is true, but once the OS leg has fallen through, the running transfer is the in-process chunked one and that call is a no-op. Nothing stops it; `uploadVideo` has no `CancelToken`. It keeps uploading and, on success, overwrites the paused/cancelled status.
- The field doc at `:151` still says "instead of the in-process chunked/legacy path" — accurate before this PR (a real exclusive if/else), false now. That stale doc is exactly what makes the guards above look correct.

The pause comment at `:1170-1174` ("A background transfer is a single OS-owned PUT with no resumable checkpoint...") is also stale for the same reason.

hm21 (commented):
The OS-first fallback treats *every* non-success from `uploadVideoInBackground` as "network died, retry in-process" — but a user-initiated cancel arrives on that same channel. With `useBackgroundUpload = true` (production, `upload_media_providers.dart:264`), Pause and Cancel now re-upload the whole video instead of stopping it, and a paused upload can come back as successful. Details inline.

Also: the `resumableTimeout` test can't fail — I deleted the entire timeout implementation and all 7 tests still passed.

mbradley (commented):
Not approving. I did a full independent pass and reproduced hm21's Pause/Cancel blocker end to end on the production-default path (`_backgroundUploadEnabled = true`), so consider it confirmed by a second set of eyes. hm21's writeup already nails the chain, so I'll just add what I didn't see covered.

One wrinkle on the fix itself: a cancel is mapped to `failureReason: unknown` (`_backgroundTransferResult:2420`), and `BlossomUploadFailureReason` has no `cancelled` value, so `unknown` is the same bucket as a generic network/unexpected failure. That means "return early on `failureReason` cancelled" can't be implemented as-is. How do you want to distinguish a user cancel from a real failure here — add a `BlossomUploadFailureReason.cancelled` (mapped from `BlossomBackgroundTransferStatus.cancelled`), or gate the fall-through on `_userStoppedUploadIds`?

Separate question, not part of the blocker: on a fresh publish where the OS leg fails before any chunk commits, the attempt stacks 30 min (OS) + 10 min (chunked) sequentially, and `performWithRetry` runs up to 6 attempts. It self-limits once a chunk commits and neither leg can hang, but it's the default path. Is that worst-case acceptable, or do you want the OS timeout tightened for the with-fallback case?

Thanks hm21 for the catch.

rabble (commented):
**Reviewed with full checkout + verification.** ✅ Looks good (posting as a comment since it's my own PR).

Replaces the `if (useBackgroundUpload) … else …` fork in `upload_manager.dart` with a single `BlossomUploadService.uploadVideoWithResume()` wrapper. On a fresh publish it tries the OS-backed whole-file PUT first (when `useBackgroundFirst`), falling through to the pre-existing chunked resumable protocol (`init` → `HEAD` offset recovery → tail chunks → `complete`) on OS failure or whenever a `resumableSession` already exists. The core offset engine is unchanged.

**Offset-trust and integrity — sound.** The server's `Upload-Offset` is trusted to skip committed bytes, but this cannot ship a corrupt blob: `POST /complete` re-verifies declared sha256 + size server-side and returns 422 on mismatch. sha256 is declared at `init` and content-addressed throughout, so a wrong/too-high offset fails the publish rather than silently shipping truncated data. No off-by-one (chunk ranges `bytes start-(end-1)/fileSize`, `end = min(start+chunkSize, fileSize)`). Idempotent re-drive: an existing session skips the OS PUT; a failed OS PUT creates no session so resume starts clean.

Test quality strong — `blossom_upload_with_resume_test.dart` asserts the resumed chunk carries `Content-Range: bytes 4-7/8` (tail only, not byte 0), existing session skips the OS enqueue, OS-failure/throw fallbacks, and monotonic session-offset callbacks `[0,4,8]`.

Verification (local):
- `flutter analyze lib/services test` → No issues found!
- `flutter test` 5 upload_manager suites → 103 passed.
- package `blossom_upload_with_resume_test.dart` → 7 passed.

Findings (none blocking): (1) with `useBackgroundFirst=true`, after OS PUT fails the in-flight chunked leg is torn down by request timeout rather than promptly on `pause`/`cancel` — pre-existing, and the OS-first path is off by default (`useBackgroundUpload=false`); (2) a failed OS-first attempt wastes the partial bandwidth then restarts resumable from byte 0 — by-design, documented, default-off.

⚠️ **Collides with #6084** — both touch `upload_manager.dart` (the `_blossomService.uploadVideo*` call site, ~887-928) and `upload_manager_resumable_upload_test.dart`. #6084 rewrites the `initialize does not auto-resume interrupted uploads` test to assert the *opposite* (`auto-resumes`). Whoever lands second must rebase and reconcile that test's intent + re-point the surviving call to `uploadVideoWithResume`. Worth deciding merge order (or combining) before landing either.

mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
NotThatKindOfDrLiz, line 2320: This needs to treat user cancellation as terminal before falling back. `_backgroundTransferResult` maps `BlossomBackgroundTransferStatus.cancelled` to `success: false` / `failureReason: unknown`, and this branch falls through on every non-success. In production `useBackgroundFirst` is true, so Pause/Cancel can cancel the OS leg, enter `runResumable()`, upload the whole file in-process, and then let `UploadManager._handleUploadSuccess` overwrite the paused/cancelled status. Please add a distinct cancelled reason/result and return it here, or otherwise gate fallback on an explicit user-stop signal.

mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
NotThatKindOfDrLiz, line 2332: Before starting chunked resumable fallback, explicitly cancel the background task with `await cancelBackgroundUpload(taskId)`. If `uploadVideoInBackground` completes via its safety timeout while the OS transfer is still alive, the chunked leg can stream the same file concurrently to the same server.

mobile/packages/blossom_upload_service/test/src/blossom_upload_with_resume_test.dart
```diff
@@ -0,0 +1,675 @@
...
+          nostrPubkey: _testPublicKey,
+          taskId: 'task-to',
+          title: 'resumable timeout',
+          description: null,
+          hashtags: null,
+          proofManifestJson: null,
+          resumableTimeout: const Duration(minutes: 1),
+        );
+
+        expect(result.success, isTrue);
```
NotThatKindOfDrLiz: This test does not assert the behavior named in the test. A fast mock plus a one-minute timeout only proves success; it would still pass if `resumableTimeout` were removed or the implementation skipped the OS leg entirely. Please make this pin both sides: the OS-first leg is not bounded by `resumableTimeout`, and the in-process resumable leg is bounded when reached.

mobile/lib/services/upload_manager.dart
```diff
@@ -868,66 +868,63 @@ class UploadManager {
...
+      // from a prior attempt, the request falls through to the chunked
+      // resumable protocol so retries resume from the server's last committed
+      // offset instead of byte 0.
+      //
+      // The OS transfer applies its own generous safety timeout internally and
+      // must NOT be cut by the in-process network timeout — doing so would
+      // defeat the suspension survival that is the whole point of the OS path.
+      // Only the in-process chunked-resumable leg is bounded by networkTimeout
+      // (passed as resumableTimeout), so a stalled socket does not hang
+      // forever. `useBackgroundUpload` is the kill-switch for the OS-first
```
NotThatKindOfDrLiz, line 969: The `useBackgroundUpload` semantics changed from "this transfer is OS-owned instead of in-process" to "try OS first, then maybe run in-process." The field docs and pause/cancel guards still read like the old meaning, which is part of why the current user-stop handling looks correct while missing the fallback case. Please update the docs/comments and make pause/cancel account for the in-process leg after fallback.

NotThatKindOfDrLiz (changes requested):
Requesting changes. I independently verified the blocker on the production-default path: user pause/cancel is now indistinguishable from an OS upload failure, so the combined upload path can fall through into resumable upload and eventually overwrite the user-stopped state. The fallback also needs to explicitly stop any live OS transfer before starting the chunked leg, and the timeout/doc coverage needs tightening before this is safe to merge.

rabble:
Thanks @hm21, @mbradley, @NotThatKindOfDrLiz — the Pause/Cancel blocker is fixed in `2b0aa95a3` (+ rebase reconciliation `91c6723f6`), rebased onto fresh `origin/main`, now MERGEABLE.

Root of the bug was exactly as diagnosed: `cancelled` was mapped to `success:false, failureReason: unknown` — byte-identical to a transport failure — so the OS-first fallback re-uploaded the whole file and `_handleUploadSuccess` overwrote the paused state. Fixes:

1. **Distinct cancelled reason** — added `BlossomUploadFailureReason.cancelled`; `_backgroundTransferResult` maps `BlossomBackgroundTransferStatus.cancelled` → `cancelled` (was `unknown`). Exhaustive consumers handled (`profile_editor_bloc` avatar+banner `cancelled || unknown => generic`); made non-retriable in `UploadRetryPolicy` (`cancelled → false`).
2. **Cancel is terminal** — `uploadVideoWithResume` returns the cancelled result immediately and never falls through to `runResumable()`. Only genuine OS failures fall through.
3. **Cancel the OS task before falling through** — `await cancelBackgroundUpload(taskId)` before the in-process leg on a genuine failure, so a still-live native transfer (e.g. one that hit its own 30-min safety timeout) can't double-upload concurrently. @NotThatKindOfDrLiz / @hm21 — this is your one-liner.
4. **Dead timeout test rewritten** (@hm21's mutation catch) — deleted the can't-fail test; replaced with two that pin both sides: OS-first leg is NOT bounded by `resumableTimeout` (OS succeeds after 150ms under a 5ms timeout), and the in-process leg IS bounded (hung chunk PUT → `TimeoutException`). Both mutation-verified can-fail.
5. **Regression test** — "OS-first cancellation is terminal and never falls back to resumable": asserts `failureReason == cancelled` + `verifyNever` on dio head/post/put. Mutation-verified (neutralizing the terminal branch → falls through → red).
6. **Stale docs** — rewrote the `useBackgroundUpload` field doc + pause comments to the new "OS-first, then fallback; cancel terminal" semantics.

@mbradley — I went with a typed `cancelled` reason (over gating on `_userStoppedUploadIds`) since it fixes the root indistinguishability. Your separate 30min+10min sequential-timeout-stacking question I left **out of scope** for this PR.

Verification: package `flutter analyze` clean + 223 tests pass + coverage stays 100%; app analyze clean + upload_manager (resumable/pause/error) 91 + profile_editor/retry_policy 135 all pass.

**One residual gap I'm flagging, not fixing here** (pre-existing, narrower than the reported bug): the `_userStoppedUploadIds` guard lives only in `_performUpload`'s `catch`, so it covers the OS-first cancel (now throws → guard fires). But if an upload has *already* fallen through to the in-process resumable leg and the user then pauses, `cancelBackgroundUpload` is a no-op on the dead OS task, the in-process leg keeps running, and if it succeeds before its request timeout the success path still overwrites `paused` (the guard doesn't cover success). Worth a follow-up; happy to file it.

NotThatKindOfDrLiz:
Takeover update pushed in `eaef283fb`:

- Closed the remaining pause/cancel overwrite window by treating `_userStoppedUploadIds` as terminal on the late-success path before thumbnail/progress side effects and before final success persistence.
- Added regression coverage for pausing during the in-process fallback leg, then receiving a later successful upload result; the row stays `paused` and does not pick up video/CDN/thumbnail success metadata.
- Updated the stale recovery sweep mocks from `uploadVideo` to `uploadVideoWithResume`, which fixes the red Mobile CI recovery tests on this PR.
- Updated linked issue #6087 from an auto-created stub into a real problem/acceptance-criteria issue.

Local verification:

- `flutter test test/services/upload_manager_background_pause_test.dart test/services/upload_manager_recovery_test.dart test/services/upload_manager_resumable_upload_test.dart`
- `flutter analyze`
- `flutter test` from `mobile/packages/blossom_upload_service`
- pre-push changed-file analyzer/tests

GitHub checks on `eaef283fb` are green: Blossom Upload Service CI, Mobile CI, preview build, mergeability, generated files, format, analyze, tests, semantic PR, CLA.

NotThatKindOfDrLiz (approved):
Approved after takeover. I re-reviewed the current head after the late-success pause/cancel fix, stale recovery mock updates, green local verification, and green GitHub checks. The prior blockers are resolved; remaining blocked state should clear when GitHub records this review state.

Issue #5638, opened 2026-06-29
task(nostr): harden use-before-ready handling with dev/CI guardrails and ready-client routing


dcadenas:
### Problem

Signer-dependent features mishandle the Nostr client around the cold-start window in two distinct ways:

- **Called too early** — relay work runs on the keyless/disposable client before the signer-backed one is ready (#5594: `withResource() … closed Pool`).
- **Never resumed** — deferred relay work isn't retried after readiness flips (#5596 f0: followed labelers enabled before ready stay unloaded until app restart).

The readiness signal itself is sound: `nostrSessionProvider` flips to `nostrReady` only when the signer-backed client is current, driven by the auth-state event with a generation guard (`_markClientReadyIfCurrent`). The gap is that gating is an implicit, forgettable convention, and deferral bookkeeping is easy to get subtly wrong.

### Fix — three complementary parts

**1. Dev/CI guardrails (catch both classes, cheaply)**

- Add a dev-only `assert` at the single relay sink (`subscribeToLabeler` / query path) that the client is ready. Stripped in release; throws in tests/CI on any *called-too-early* path.
- Add not-ready→ready transition regression tests asserting deferred work *resumes* (covers the *never-resumed* class), at minimum the followed-labeler path.

**2. Route signer-dependent consumers through the ready client**

Have the ~7 signer-dependent consumers take the client from `nostrSessionProvider.client` (typed `NostrClient?`, non-null only when `isReadyForActiveClient`) instead of capturing the always-non-null `nostrServiceProvider` and bolting on a separate check. Because the type is nullable, Dart null-safety makes the readiness gate **compiler-enforced** — new code can't use it without handling not-ready, so the gate can't be silently skipped.

Consumers:
- `mobile/lib/providers/moderation_providers.dart`
- `mobile/lib/providers/notifications_providers.dart`
- `mobile/lib/providers/repository_providers.dart`
- `mobile/lib/providers/social_providers.dart`
- `mobile/lib/services/push_notification_session_coordinator.dart`

**3. AGENTS.md rule** (lands in this issue's PR)

Append to "## Nostr And Async Rules":

```markdown
[...]

PR #5596, opened 2026-06-27
fix(moderation): wait for ready nostr session

NotThatKindOfDrLiz:
Closes #5594

## Summary

- moves moderation labeler startup safety into `ModerationLabelService` so every caller shares the same hydration and relay-readiness rules
- loads persisted labeler settings before any save path can overwrite SharedPreferences, preventing saved custom labelers from being wiped during startup
- gates only relay-backed label queries on the active signer-backed `NostrClient`, including direct `addLabeler` / followed-labeler sync paths
- switches `moderationLabelServiceProvider` from `ref.watch(nostrSessionProvider)` to `ref.listen(...)` so readiness transitions no longer rebuild the keepAlive service or recreate `VideoEventService`
- hydrates moderation settings before `SafetySettingsCubit.load()` snapshots labeler state
- adds regressions for not-ready startup, stale ready-client identity, and saved-labeler preservation before relay readiness

## Root Cause

On restored sessions the app can briefly build a keyless or disposable NostrClient before auth restoration installs the signer-backed client. The original provider immediately initialized `ModerationLabelService`, so labeler queries could run against a client that was disposed during auth restoration, producing the closed Pool startup error.

The first fix moved that gate into the provider, but review caught that user mutation paths could still save against an unhydrated in-memory labeler set and could still call relay queries directly. The final fix separates local prefs hydration from relay work inside the service: local state loads first for all mutators, while relay queries run only when the current service client is the ready active session client.

## Validation

- `flutter test test/screens/settings/settings_taxonomy_test.dart`
- `flutter test test/services/moderation_label_service_test.dart test/providers/moderation_label_provider_session_test.dart test/blocs/safety_settings/safety_settings_cubit_test.dart test/screens/settings/settings_taxonomy_test.dart`
- `flutter analyze`
- `dart run build_runner build --delete-conflicting-outputs`
- pre-push hook: merge-conflict check, analyzer, generated-file verification, changed-file tests
- GitHub checks green: Mobile CI, Tests, Analyze, Format, Generated Files, VGV Tag Gate, Web Preview, mergeability, semantic PR, linked issue, CLA

PR #6089, opened 2026-07-14
fix(nostr): guard queryEvents against a disposed client's closed pool

### LINKED ISSUE #5952: fix(moderation): labeler subscription fails on closed Nostr pool at startup — labels silently skipped
## Summary
On cold start, `ModerationLabelService.subscribeToLabeler` throws `Bad state: withResource() may not be called on a closed Pool` and the error is swallowed, so **moderation labels silently fail to load for the session**. This appears at startup in every log-bearing support ticket from the Jul 4–7 window (#5920, #5933, #5949, #5851, #5851-era), e.g.:

```
[ERROR] [SYSTEM] (ModerationLabelService) Error subscribing to labeler 8fd5eb6d8f362163bc00a5ab6b4a3167dbf32d00ec4efdbcf43b3c9514433b7e: Bad state: withResource() may not be called on a closed Pool.
```

## Root cause
`mobile/lib/services/moderation_label_service.dart:220` calls `_nostrClient.queryEvents([filter])` from `subscribeToLabeler`. When this runs against a Nostr client whose connection `Pool` has been closed (client not yet ready, or torn down/re-initialised during a cold-start / auth lifecycle transition), `queryEvents` throws the `closed Pool` `StateError`. The `catch (e)` at L234-240 logs it and returns — the labeler is **not** added to `_loadedLabelers`, so its Kind 1985 / NIP-32 labels are never applied for that session.

This is the same "use before / around client lifecycle readiness" family as #5638 (Harden Nostr use-before-ready handling).

## Impact
Trust & Safety: when this fires for the built-in Divine labeler, moderation labels are absent for the session — content that should be filtered/labelled may render unlabelled until the app is relaunched. It's non-fatal (caught) but silent, and it's hitting real users at startup, not an edge case.

## Proposed fix
- Route `subscribeToLabeler` through the ready Nostr client (await readiness) rather than querying a possibly-closed pool — consistent with the #5638 direction.
- On a `closed Pool` / not-ready failure specifically, retry once the client is ready instead of permanently skipping the labeler for the session (don't leave it out of `_loadedLabelers` on a transient lifecycle error).
- Add a test covering: subscribe attempted while pool closed → labeler is retried/loaded once the client is ready (not silently dropped).

## Notes
- Caught-and-logged today, so no Crashlytics signal — severity is masked by the swallow.
- Related: #5638 (Nostr use-before-ready hardening).

realmeylisdev:
Closes #5952

## Summary

- #5952 reported `ModerationLabelService.subscribeToLabeler` throwing `Bad state: withResource() may not be called on a closed Pool` at cold start, with moderation labels silently failing to load. Investigation found the crash and permanent-loss case were already fixed by #5596 (merged before #5952 was filed) — the support tickets cited in #5952 are all on a pre-#5596 build.
- What remained: a narrow, transient version of the same logged error that can still fire when `dispose()` races the async work inside `NostrClient.queryEvents` between a caller's readiness check and the actual `_queryPool.withResource` call.
- This PR closes that residual gap: an upfront `isDisposed` check (skips pointless cache/reconnect work on an already-disposed client) plus a re-check of the pool's own `isClosed` state immediately before use, with no `await` in between — closing the race rather than narrowing it, and without needing to catch a `StateError` (avoids the `avoid_catching_errors` lint, and is more precise since it checks the pool's real state rather than inferring it from a caught exception).

## Investigation trail

Full findings, confidence-scored hypotheses, and the considered-and-rejected alternatives (folding into #5638, mobile-side consumption of the `moderation/resolution` namespace) are in the issue thread and this session's brainstorm. Two adjacent findings surfaced during investigation were filed separately, out of scope for this PR:
- `divine-relay-manager#180` — a backend "ban" audit label published without a confirmed enforcement RPC call
- `divine-mobile#6088` — four unrelated NIP-32 parser spec-compliance gaps in `ModerationLabelService`, invisible today since production publishes zero `content-warning` events

## Test plan

- [x] New regression test: `queryEvents` on a disposed client returns `[]` without throwing, and never reaches the underlying WebSocket query
- [x] `flutter test` — full `nostr_client` package (287 tests) and `moderation_label_service_test.dart` (20 tests), all pass
- [x] Package coverage: 85.17%, above the 82% `min_coverage` gate
- [x] Pre-commit and pre-push hooks passed

### DIFF
--- a/mobile/packages/nostr_client/lib/src/nostr_client.dart
+++ b/mobile/packages/nostr_client/lib/src/nostr_client.dart
@@ -588,6 +588,10 @@ class NostrClient {
   /// Then queries via WebSocket and merges results.
   ///
   /// Results from websocket are cached for future queries.
+  ///
+  /// Calling this on a disposed client is a no-op that returns an empty
+  /// list. If the client is disposed while a call is in flight, the
+  /// network query is skipped and only cached results are returned.
   Future<List<Event>> queryEvents(
     List<Filter> filters, {
     String? subscriptionId,
--- a/mobile/packages/nostr_client/lib/src/nostr_client.dart
+++ b/mobile/packages/nostr_client/lib/src/nostr_client.dart
@@ -637,14 +637,18 @@ class NostrClient {
     // relay's "too many concurrent REQs" limit. `withResource` releases the
     // slot when the (time-bounded) query completes, so it can't leak.
     //
-    // Re-check the pool's own closed state immediately before use (rather
-    // than relying solely on the upfront isDisposed check above): dispose()
-    // can run during the awaits above, and `_queryPool.isClosed` flips
-    // before `_isDisposed` does (dispose() closes the pool first). This
-    // check-then-call has no await between them, so it closes the race
-    // rather than narrowing it. See #5952.
+    // Re-check the pool's own closed state immediately before the query,
+    // independent of `useQueryPool`: dispose() can run during the awaits
+    // above, and `_queryPool.close()` flips `isClosed` before `_isDisposed`
+    // is set (dispose() closes the pool first), so it's a reliable
+    // dispose-in-progress sentinel for both paths. The non-pooled path
+    // (`queryUsers`, `useQueryPool: false`) matters here too: a query that
+    // fell through after `_nostr.close()` would re-open fresh WebSockets to
+    // the NIP-50 search relays and leak temp relays nothing would clean up.
+    // This check-then-call has no await before the query, so it closes the
+    // race rather than narrowing it. See #5952.
     final List<Event> websocketEvents;
-    if (useQueryPool && _queryPool.isClosed) {
+    if (_queryPool.isClosed) {
       websocketEvents = [];
     } else {
       websocketEvents = useQueryPool
--- a/mobile/packages/nostr_client/lib/src/nostr_client.dart
+++ b/mobile/packages/nostr_client/lib/src/nostr_client.dart
@@ -598,6 +598,13 @@ class NostrClient {
     bool useQueryPool = true,
     Duration timeout = const Duration(seconds: 5),
   }) async {
+    // A disposed client's query pool is closed; querying it is a no-op
+    // rather than an error. This is the common case (checked upfront to
+    // skip pointless cache/reconnect work below) — the narrower re-check
+    // right before `withResource` (see below) closes the residual race
+    // where dispose() runs during the awaits in between. See #5952.
+    if (_isDisposed) return [];
+
     final effectiveTempRelays = _allowedRelays(tempRelays);
     final cacheResults = <Event>[];
 
@@ -629,9 +636,21 @@ class NostrClient {
     // videos → per-item like-count/badge/profile/repost fetches) can't trip a
     // relay's "too many concurrent REQs" limit. `withResource` releases the
     // slot when the (time-bounded) query completes, so it can't leak.
-    final websocketEvents = useQueryPool
-        ? await _queryPool.withResource(runWebSocketQuery)
-        : await runWebSocketQuery();
+    //
+    // Re-check the pool's own closed state immediately before use (rather
+    // than relying solely on the upfront isDisposed check above): dispose()
+    // can run during the awaits above, and `_queryPool.isClosed` flips
+    // before `_isDisposed` does (dispose() closes the pool first). This
+    // check-then-call has no await between them, so it closes the race
+    // rather than narrowing it. See #5952.
+    final List<Event> websocketEvents;
+    if (useQueryPool && _queryPool.isClosed) {
+      websocketEvents = [];
+    } else {
+      websocketEvents = useQueryPool
+          ? await _queryPool.withResource(runWebSocketQuery)
+          : await runWebSocketQuery();
+    }
 
     // Cache websocket results (fire-and-forget)
     if (websocketEvents.isNotEmpty) {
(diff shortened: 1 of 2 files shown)


mobile/packages/nostr_client/lib/src/nostr_client.dart
```diff
@@ -629,9 +636,21 @@ class NostrClient {
...
-        : await runWebSocketQuery();
+    //
+    // Re-check the pool's own closed state immediately before use (rather
+    // than relying solely on the upfront isDisposed check above): dispose()
+    // can run during the awaits above, and `_queryPool.isClosed` flips
+    // before `_isDisposed` does (dispose() closes the pool first). This
+    // check-then-call has no await between them, so it closes the race
+    // rather than narrowing it. See #5952.
+    final List<Event> websocketEvents;
+    if (useQueryPool && _queryPool.isClosed) {
```
hm21: P2: gating on `useQueryPool` leaves the non-pooled path unguarded. `queryUsers()` calls this with `useQueryPool: false` and `tempRelays: _nip50SearchRelays`. In the dispose-during-cache-await window it falls through to `await runWebSocketQuery()` → `_nostr.queryEvents(...)` after `_nostr.close()`. `Nostr.close()` → `_pool.removeAll()` clears `_relays` but not `_tempRelays`, and `checkAndGenTempRelay()` calls `connect()` — so a disposed client opens fresh WebSockets to the 3 search relays and leaks a temp relay nothing will clean up.

Fix: make the check independent of `useQueryPool`:
```dart
if (_queryPool.isClosed) {
  websocketEvents = [];
} else {
  websocketEvents = useQueryPool
      ? await _queryPool.withResource(runWebSocketQuery)
      : await runWebSocketQuery();
}
```
`_queryPool` is closed before `_nostr` in `dispose()` (pool.close at :1415, `_isDisposed` at :1420), so it's a reliable dispose-in-progress sentinel for both paths.

mobile/packages/nostr_client/test/src/nostr_client_test.dart
```diff
@@ -1154,6 +1154,30 @@ void main() {
           ),
         ).called(1);
       });
+
+      test(
+        'returns empty list without throwing when client is disposed',
```
hm21: P2: this only exercises the upfront `_isDisposed` early-return. Since either guard alone makes it pass, the test stays green even if the pre-`withResource` re-check is deleted — it doesn't protect the race it's for. Make it deterministic: park the query at the cache `await` (inject a DAO whose `getEventsByFilter` returns a controllable `Completer`), run `dispose()`, then complete it and assert no `_nostr.queryEvents`. Add a `useQueryPool: false` case to cover the fix above.

hm21 (commented):
Two changes needed before merge — verified against the SDK, not just the diff.

1. The `isClosed` re-check is gated on `useQueryPool`, so the `useQueryPool: false` path (`queryUsers` / NIP-50 search) stays racy — see inline on the guard.
2. The regression test never exercises the race it's meant to protect — see inline on the test.

Context: #5952's moderation path (pooled, `useQueryPool: true`) is fully fixed. Both items are about the guard/test being complete; (1) is a pre-existing gap this PR leaves half-closed rather than a regression it introduces.

mobile/packages/nostr_client/lib/src/nostr_client.dart
realmeylisdev: Fixed in 4171acbcd5 — dropped the `useQueryPool &&` prefix so the `isClosed` check gates both paths. Confirmed your SDK read: `Nostr.close()` → `removeAll()` clears `_relays` but not `_tempRelays`, so a post-close `queryUsers` re-`connect()`s the search relays via `checkAndGenTempRelay()`. `_queryPool.close()` is the only close site and flips before `_isDisposed`, so it is a sound dispose sentinel for both paths.

mobile/packages/nostr_client/test/src/nostr_client_test.dart
realmeylisdev: Fixed in 4171acbcd5 — replaced it with three tests that park at the cache `await` and dispose mid-flight, covering the pooled path, the `useQueryPool: false` path, and a strengthened upfront-guard case. Each is verified by mutation: deleting the pre-query guard reproduces `Bad state: withResource() may not be called on a closed Pool`, the old `useQueryPool &&` form makes the non-pooled case reach `_nostr.queryEvents`, and removing the upfront return makes the disposed-client case read the cache.

realmeylisdev:
@hm21 both addressed in 4171acbcd5. Guard is now independent of `useQueryPool` so the non-pooled `queryUsers` path is covered, and the regression test parks at the cache `await` and disposes mid-flight (pooled + non-pooled), with each guard mutation-verified to fail. Package green: analyze clean, 289 tests pass, coverage 85.44%.

rabble (approved):
**Reviewed with full checkout + verification.** ✅ Approve.

Clean, well-scoped dispose-race fix. The two guards are an atomic check-then-act: `dispose()` closes the pool before setting `_isDisposed`, so `_queryPool.isClosed` is a reliable dispose-in-progress sentinel, and there's no `await` between the check and the `withResource`/query call, so the race is closed rather than narrowed. `_queryPool` has a single consumer (`queryEvents`), so all entry paths are covered. Returns `[]` only for the disposed/closed no-op — real query errors still propagate. Both race tests park a query at the cache `await`, verify they're in the window, dispose mid-flight, and assert no throw + network skipped; deleting the guard makes them throw the closed-Pool `StateError`. Real failing-for-the-right-reason tests.

Verification (local):
- `flutter analyze lib test` → No issues found!
- `flutter test test/src/nostr_client_test.dart` → 153 tests passed.

One optional nit: the `///` dartdoc on `queryEvents` doesn't mention that a disposed client returns `[]`. Per `error_handling.md`, a sentinel return should be documented on the public method. Non-blocking.

Issue #6093, opened 2026-07-14
Playlist deep links should open the app (Universal Links + App Links)


mbradley:
## Why

Marketing wants to use links to playlists as a CTA in future email campaigns (for example, pointing recipients at a curated playlist). Today a playlist link opens on the web but does **not** open the app, so on a phone those taps land in a browser rather than in Divine. This wires playlist links into the same Universal Link / App Link handling we already have for video, profile, hashtag, search, and invite, so a playlist CTA opens the app when it is installed.

## Current state (verified)

- The destination already exists in-app: `CuratedListFeedScreen` at internal route `/list/:listId` (`mobile/lib/screens/curated_list_feed_screen.dart:33`), reachable from Explore/Search. Lists are NIP-51 kind 30005 curated video sets.
- Web already serves playlists at `/list/:pubkey/:listId` (`divine-web/src/AppRouter.tsx:108`, `ListDetailPage`).
- What's missing for the app to open on a playlist link:
  - iOS AASA (`divine-web/public/.well-known/apple-app-site-association`) has no `/list` component (only `/video`, `/profile`, `/hashtag`, `/search`, `/invite`, `/app/callback`).
  - Android intent filter (`mobile/android/app/src/main/AndroidManifest.xml`) has no `/list` `pathPrefix`.
  - The deep-link parser (`mobile/lib/services/deep_link_service.dart`) has no `list`/`playlist` branch, so `/list/...` returns `unknown` and no-ops.
- URL-shape mismatch to resolve: web uses `/list/:pubkey/:listId`, the mobile internal route uses `/list/:listId`. A kind-30005 list is addressed by its `naddr` (kind + pubkey + d-identifier), so the web shape (pubkey + listId) is the more complete canonical form.

## Scope

- [ ] Decide the canonical public URL shape (recommend `/list/:pubkey/:listId`) and align web + mobile on it.
- [ ] divine-web: add the `/list/*` component to `apple-app-site-association` (cross-repo, folded into this issue).
- [ ] divine-mobile: add the `/list` `pathPrefix` to the `autoVerify` intent filter in `AndroidManifest.xml`.
- [ ] divine-mobile: add a `DeepLinkType.playlist` (or `list`) branch to `deep_link_service.dart` that parses `/list/:pubkey/:listId` and emits a `DeepLink`.
[...]

Issue #3032, opened 2026-04-13
fix: deep links bypass DeepLinkService and hit GoRouter directly, showing Page Not Found


omartinma:
## Summary

While working on #2477, it has been noticed that deep links opened via `adb shell am start` (and likely universal links in production) are routed directly to GoRouter instead of through `DeepLinkService`. GoRouter receives the full URL (e.g. `https://divine.video/search/music`) as the location, fails to match any route, and shows the default "Page Not Found" error page.

## Environment
- App version: Latest (debug build)
- Device: Android Emulator (Medium Phone API 36)
- Network: N/A

## Steps to Reproduce
1. Run the app on an Android emulator
2. In a terminal, run:
   ```
   adb shell am start -a android.intent.action.VIEW -d "https://divine.video/search/music" co.openvine.app
   ```
3. See "Page Not Found" with `GoException: no routes for location: https://divine.video/search/music`

## Actual Result
GoRouter intercepts the deep link URL directly and tries to match `https://divine.video/search/music` against its route table. No route matches, so it shows the default error page. `DeepLinkService` (which correctly parses the URL and calls `router.go()` with the right internal path) never receives the event.

## Expected Result
Deep links should be parsed by `DeepLinkService`, which converts them to internal routes (e.g. `/search-results/music`) and navigates via `router.go()`.

## Regression?
Unknown — this likely affects all deep link types (`/video`, `/profile`, `/hashtag`, `/search`), not just search. It may have been masked because GoRouter happened to have matching route patterns for some URL paths.

## Evidence
`MaterialApp.router` with `routerConfig: router` gives GoRouter full control over incoming URLs. GoRouter's `redirect` callback only fires after a route matches — unmatched URLs throw `GoException` before redirect runs. `DeepLinkService` listens via `app_links` but GoRouter processes the intent first.

omartinma:
@NotThatKindOfDrLiz this might be relevant while working on local handling for Push Notifications

NotThatKindOfDrLiz:
Assigning to @untreu2 for mobile routing ownership.

Oscar/@omartinma flagged the push-notification dependency, but he is busy on other work. The important handoff context is that notification tap-through should use the same deep-link path, so this fix should cover both universal links and notification payload navigation.

[...]

PR #6094, opened 2026-07-14
feat(deep-links): open playlist links in-app via /list/:pubkey/:listId

mbradley:
> [!NOTE]
> **Draft / work-in-progress** — staking out the implementation early per #6093. Code + tests are complete and green locally; manual device verification and the divine-web AASA companion PR link are still pending (checklist below).

## Description

Marketing wants playlist links as CTAs, but tapping a `divine.video` playlist link today opens the browser even when the app is installed. This wires `/list/:pubkey/:listId` — the web-canonical shape, since kind 30005 lists are addressed by author + d-tag — into the same Universal Link / App Link handling that video, profile, hashtag, search, and invite already use.

- **Parser**: new `DeepLinkType.list` branch in `deep_link_service.dart`. Accepts a hex or npub author segment and normalizes to lowercase hex (what relay `authors` filters expect). Additive — all existing link types are untouched, pinned by the existing test suites.
- **Routing**: list links flow through `universal_link_resolver.dart` (the #3032 path — router redirect on Android cold start, stream listener otherwise) into a new `/list/:pubkey/:listId` route. Its `CuratedListByAuthorScreen` resolves the list from relays by author + d-tag (`CuratedListService.fetchPublicList`, cache + relay via `queryEvents`, newest-version-wins) and then renders the existing `CuratedListFeedScreen`. This is the piece that makes links work for lists the user has never seen — the internal `/list/:listId` route only knows about locally stored lists plus route extras, so it was not reusable for deep links.
- **Listener nav parity**: `resolveListDeepLinkNavAction` mirrors the video/profile/hashtag/search push/go/skip semantics, so a warm-app link pushes onto the stack (back returns where you were), replaces an already-showing list in place, and dedupes when the router redirect already navigated.
- **Android**: `/list` `pathPrefix` added to the `autoVerify` intent filter.
- **Route normalization**: `parseRoute`/`buildRoute` in `page_context_provider.dart` learned the 3-segment shape. Without this, `routeNormalizationProvider` parsed `/list/<pubkey>/<listId>` as the 2-segment internal route (author pubkey captured as the list id), rebuilt the "canonical" form as `/list/<pubkey>`, and bounced the user off the screen the moment it opened — the same failure mode as the `/settings/storage` and `/invites` regressions the route-coverage test documents. Round-trip tests now pin both list route shapes.

**Cross-repo dependency:** iOS needs the `/list/*` component in the AASA file served by divine-web (divinevideo/divine-web#482). The feature only works end-to-end once **both** are live, and Apple caches AASA per device/CDN, so iOS can lag the web deploy. Android is self-contained in this PR.

Deliberate choices for reviewer eyes:

- Not-found and fetch-error states reuse the existing `curatedListFailedToLoad` copy via `RouteErrorScreen` — no new ARB keys, no translation debt. Distinct "list not found" copy would be a separate product/copy change.
- The list nav-action handler mirrors the four existing deep-link types (video/profile/hashtag/search) verbatim rather than consolidating all five into one generic resolver. Consolidation is a real cleanup but its own reviewed change — it touches every existing deep-link type (including video's `goSameRouteWithComments` special case), so folding it into this feature PR would balloon the blast radius. Deferred deliberately.
- The resolver screen is a thin `ConsumerWidget` in the existing Riverpod lists surface (legacy area per the migration policy), consistent with `CuratedListFeedScreen` / `DiscoverListsScreen`, rather than introducing a one-off BLoC around a single async read.

**Related Issue:** Relates to #6093

## Out of Scope

- In-app "share list" URL minting (mobile has no list-share affordance today; web already generates these URLs).
- Distinct not-found copy (reuses existing localized string, see above).

## Remaining before marking ready

- [x] Code review loop (converged clean over 3 rounds). Fixed: cold-start back-nav crash (safePop), Android `/lists` over-claim (pathPattern), provider re-fetch storm (notifier dep), shared pubkey normalizer (nprofile support), in-repo AASA `/list/*` mirror synced, NIP-01 replaceable tie-break in `fetchPublicList`, and triage logging on unresolved links. Copy reuse for the not-found/error state was a deliberate call (no new ARB key).

- [x] divine-web AASA companion PR: divinevideo/divine-web#482
- [ ] Manual Android emulator pass against a real published kind 30005 list, e.g. "Art of Divine" (156 videos, verified live on `relay.divine.video` and serving on web):
      `adb shell am start -a android.intent.action.VIEW -d "https://divine.video/list/3e911baba55ae247339cf805dd6ff49ad2cd6bee84ac44e088ce66450c49104f/list_1780802850377"`
- [ ] Manual iOS simulator pass (`xcrun simctl openurl`)

## Verification

- Targeted suites, all green: `deep_link_service_test`, `universal_link_resolver_test`, `curated_list_service_fetch_public_test` (new), `curated_list_service_{query,crud,stream}_test`, `curated_list_relay_sync_test`, `curated_list_feed_screen_test`, `curated_list_by_author_screen_test` (new), `main_list_deep_link_nav_action_test` (new), `main_{profile,search}_deep_link_nav_action_test`, `deep_link_provider_test`, `people_lists_route_order_test`
- `dart run build_runner build` — regenerated `list_providers.g.dart` committed

## Type of Change

### DIFF
--- a/mobile/test/services/curated_list_service_fetch_public_test.dart
+++ b/mobile/test/services/curated_list_service_fetch_public_test.dart
@@ -150,6 +150,41 @@ void main() {
       expect(list!.name, equals('New Title'));
     });
 
+    test(
+      'keeps the lowest event id when timestamps tie (NIP-01 replaceable '
+      'ordering)',
+      () async {
+        final idHigh = 'f' * 64;
+        final idLow = '0' * 64;
+        // Returned high-first so a naive strict "isAfter" keeps the wrong one.
+        when(() => mockNostr.queryEvents(any())).thenAnswer(
+          (_) async => [
+            _listEvent(
+              id: idHigh,
+              dTag: 'my-vines',
+              createdAt: 1000,
+              title: 'High',
+            ),
+            _listEvent(
+              id: idLow,
+              dTag: 'my-vines',
+              createdAt: 1000,
+              title: 'Low',
+            ),
+          ],
+        );
+
+        final list = await service.fetchPublicList(
+          authorPubkey: _authorPubkey,
+          listId: 'my-vines',
+        );
+
+        expect(list, isNotNull);
+        expect(list!.name, equals('Low'));
+        expect(list.nostrEventId, equals(idLow));
+      },
+    );
+
     test('ignores events whose d-tag does not match', () async {
       when(() => mockNostr.queryEvents(any())).thenAnswer(
         (_) async => [
--- a/mobile/test/services/curated_list_service_fetch_public_test.dart
+++ b/mobile/test/services/curated_list_service_fetch_public_test.dart
@@ -199,5 +199,34 @@ void main() {
 
       expect(list, isNull);
     });
+
+    test(
+      'ignores an event whose d-tag matches but pubkey does not '
+      '(loose cache hit)',
+      () async {
+        const otherAuthor =
+            'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff';
+        // queryEvents merges cache hits, whose author filtering can be looser
+        // than a relay's — a matching-d-tag, wrong-author event can slip in,
+        // which is exactly why fetchPublicList re-checks the pubkey.
+        when(() => mockNostr.queryEvents(any())).thenAnswer(
+          (_) async => [
+            _listEvent(
+              id: 'event_wrong_author',
+              dTag: 'my-vines',
+              createdAt: 1000,
+              pubkey: otherAuthor,
+            ),
+          ],
+        );
+
+        final list = await service.fetchPublicList(
+          authorPubkey: _authorPubkey,
+          listId: 'my-vines',
+        );
+
+        expect(list, isNull);
+      },
+    );
   });
 }
--- a/mobile/test/services/curated_list_service_fetch_public_test.dart
+++ b/mobile/test/services/curated_list_service_fetch_public_test.dart
@@ -1,0 +1,168 @@
+// ABOUTME: Unit tests for CuratedListService.fetchPublicList
+// ABOUTME: Covers relay fetch by author + d-tag for list deep links
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mocktail/mocktail.dart';
+import 'package:nostr_client/nostr_client.dart';
+import 'package:nostr_sdk/event.dart';
+import 'package:nostr_sdk/filter.dart';
+import 'package:openvine/services/auth_service.dart';
+import 'package:openvine/services/curated_list_service.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+class _MockNostrClient extends Mock implements NostrClient {}
+
+class _MockAuthService extends Mock implements AuthService {}
+
+const _authorPubkey =
+    'a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e9f0a1b2';
+const _videoEventId =
+    '1111111111111111111111111111111111111111111111111111111111111111';
+
+Event _listEvent({
+  required String id,
+  required String dTag,
+  required int createdAt,
+  String title = 'My Vines',
+  String pubkey = _authorPubkey,
+}) {
+  return Event.fromJson({
+    'id': id,
+    'pubkey': pubkey,
+    'created_at': createdAt,
+    'kind': 30005,
+    'tags': [
+      ['d', dTag],
+      ['title', title],
+      ['e', _videoEventId],
+    ],
+    'content': '',
+    'sig': 'test_signature',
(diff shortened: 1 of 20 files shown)


mobile/test/services/curated_list_service_fetch_public_test.dart
```diff
@@ -0,0 +1,203 @@
...
+        ],
+      );
+
+      final list = await service.fetchPublicList(
+        authorPubkey: _authorPubkey,
+        listId: 'my-vines',
+      );
+
+      expect(list, isNull);
+    });
```
realmeylisdev, line 201: should-fix: this suite covers the d-tag mismatch branch but never the pubkey mismatch one (`list.pubkey != authorPubkey` in curated_list_service.dart's fetchPublicList). Deleting that check locally and re-running this file still passes all 6 tests. Given the doc comment right above fetchPublicList explains this check exists specifically because cache hits can return the wrong author, a sibling test here (matching d-tag, wrong pubkey) seems worth it.

mobile/test/services/deep_link_service_test.dart
```diff
@@ -223,6 +225,104 @@ void main() {
...
+
+      test('rejects list URL without a list id', () {
+        const url = 'https://divine.video/list/$authorPubkey';
+
+        final result = DeepLinkService.parseDeepLink(url);
+
+        expect(result.type, equals(DeepLinkType.unknown));
+        expect(result.listPubkey, isNull);
+        expect(result.listId, isNull);
+      });
```
realmeylisdev, line 299: should-fix: same kind of gap here for the `listId.isEmpty` half of the guard in deep_link_service.dart. This test only covers a 2-segment URL (list id missing entirely), not a 3-segment URL with a valid pubkey and an empty last segment - e.g. a trailing-slash link like `/list/$authorPubkey/`. Deleting the isEmpty check locally still passed all 58 tests in this file.

realmeylisdev (commented):
went deep on the list deep-link flow end to end - parser, router round-trip, safePop cold-start fix, the push/go/skip nav logic, the NIP-01 tie-break in fetchPublicList, queryEvents relay semantics, and the AASA/Android manifest sync against the divine-web companion PR. all of it holds up, nice work.

left two inline comments on test gaps I mutation-tested (removing the check still passed the suite both times). nothing blocking, just worth closing before merge.

rabble (approved):
**Reviewed with full checkout + verification.** ✅ Approve. Rebase onto fresh `origin/main` before merge (the branch is behind — `origin/main..HEAD` shows unrelated churn; the true PR is +1364/-1, 20 files).

URL-parsing robustness and Nostr ID handling — both solid:
- `deep_link_service.dart` routes only `pathSegments.length == 3 && [0]=='list'`, normalizes the author via `normalizePublicIdentifier` (hex/npub/nprofile → lowercase hex, null otherwise), rejects empty listId → `unknown`. Both guard halves are mutation-tested.
- No Nostr ID truncation anywhere — full IDs in logs and `DeepLink.toString`, with an explicit untruncated-pubkey test.
- Android uses `pathPattern="/list/.*"` (correctly avoids claiming the web `/lists` browse page); iOS AASA `/list/*` mirror is consistent.
- `fetchPublicList` re-checks author + d-tag per event and applies NIP-01 replaceable ordering (newer created_at wins, lowest event id on tie); `limit:1` correctly dropped so all versions arrive.
- Layer boundaries clean; double-fire (router redirect vs app_links listener) both resolve to the same hex path and dedupe via `resolveListDeepLinkNavAction`.

Verification (local):
- `flutter analyze lib test` → No issues found!
- 8 targeted suites → 211 tests passed.
- `test/l10n/arb_consistency_test.dart` → passed (3 new keys mirrored/allowlisted across 16 locales).

Findings — all low/non-blocking:
1. `curated_list_by_author_screen.dart:47` — screen/`publicCuratedListProvider` assume `authorPubkey` is already hex. Every real entry path normalizes first, so a raw npub reaching the GoRoute param is theoretical (would silently show unavailable). Optional hardening: normalize in the route builder.
2. `page_context_provider.dart` — `RouteContext.npub` holds hex (not an npub) for the by-author case. Round-trip-consistent but the name mismatch is confusing; rename or comment.
3. `docs/apple-app-site-association` is a repo mirror — the authoritative AASA is served by divine-web; iOS universal links for `/list/*` won't work until the served file includes the component. Confirm before relying on iOS universal linking.
4. `list_providers_test.dart` "does not re-run on re-emit" only covers equal-data re-emit; a genuine lists-state change could re-fetch and briefly flash loading. Low impact.

mbradley:
both should-fixes are in (9dfab7d18). added the matching-d-tag / wrong-pubkey case to the fetchPublicList suite, and a trailing-slash `/list/<pubkey>/` case that hits the empty-listId branch. mutation-checked both: pulling each guard turns its test red, and every other test stays green.

thanks, as always.

PR #5600, opened 2026-06-27
fix(nostr): recover failed client initialization

NotThatKindOfDrLiz:
## Summary

Fixes #5577 by hardening the app-wide Nostr startup path when an authenticated identity is known but the signer-backed NostrClient never reaches nostrReady. The recovery path now handles both thrown initialization failures and non-throwing startup stalls.

This PR:
- centralizes NostrClient creation, callback registration, initialization, and disposal in NostrService
- wraps the addRelays() + initialize() startup sequence in a bounded timeout so a stalled setup becomes retryable
- retries failed or timed-out initialization with named, bounded backoff policy values
- serializes retry work through the same lifecycle queue as auth-state changes
- initializes retry candidates before publishing them to provider state, avoiding repeated consumer graph rebuilds during sustained failures
- cancels pending retries on auth changes and provider disposal so stale clients cannot overwrite newer sessions
- keeps the existing generation and pubkey guards for stale init completions and relay discovery callbacks
- adds regression coverage for startup timeout recovery, fail-fail-succeed retry, retry backoff, auth-change cancellation, and dispose cancellation

## Root Cause

The initial NostrService build path initialized the shared NostrClient once in a microtask. If that setup failed or stalled, the provider stayed at identityKnown and nothing recreated or retried the signer-backed client. Downstream providers such as profileRepositoryProvider stayed unavailable because they require nostrSessionProvider.isReadyForActiveClient.

The issue log for #5577 showed the app reporting Nostr service not initialized for multiple minutes without an initialization failure log. That points at a non-throwing startup stall rather than only a thrown initialize() failure, so this PR now turns that stalled startup sequence into a bounded timeout and routes it through the same recovery path.

## Validation

- dart run build_runner build --delete-conflicting-outputs
- flutter test test/providers/nostr_service_identity_source_test.dart
- flutter test test/providers
- flutter analyze
- dart format --output=none --set-exit-if-changed lib test integration_test
- git diff --check
- pre-push hook: merge-conflict check, analyzer, generated-file verification, changed provider test

Closes #5577

PR #6101, opened 2026-07-15
chore: Prevent relay reconnects during Nostr startup

### LINKED ISSUE #6100: Prevent relay reconnects during Nostr startup
## Context

During startup, Divine adds its default and user relays while the Nostr client is still initializing. The relay-set watcher observes these membership changes and normally debounces them for two seconds before reconnecting every relay and resetting feed subscriptions.

## Problem

The watcher can mistake a startup relay change for a later configuration change. On a slower startup, that reconnect can interrupt initialization and leave the app waiting for the startup timeout. If initialization finishes during the debounce window, checking only the client's state when the timer fires still allows an unnecessary reconnect and feed reset.

## Work Requested

Treat relay changes first observed while the client is uninitialized as startup-originated changes. Suppress their pending reconnect even if initialization finishes before the debounce expires. Preserve the existing reconnect behavior for membership changes first observed after initialization.

Add startup diagnostics that identify the initialization source, current stage, elapsed time, and relay count, including the last blocking stage when initialization times out.

## Scope

- Update startup classification in the relay-set change bridge.
- Add regression coverage for startup and post-startup timing cases.
- Add stage-aware Nostr initialization diagnostics.
- Preserve the existing `NostrClient.initialize()` contract.
- Do not change reconnect behavior for relay changes first observed after initialization.

## Acceptance Criteria

- A relay change observed while the client is uninitialized does not call `forceReconnectAll()` or reset feeds when the client remains uninitialized through debounce expiry.
- A relay change observed while the client is uninitialized remains suppressed if initialization finishes before debounce expiry.
- A relay change first observed after initialization retains the existing debounced reconnect and feed-reset behavior.
- Initialization and timeout logs report the source, current stage, elapsed time, and relay count.
- Regression tests cover all three timing cases.

## Validation

- Run the focused Nostr service and relay-set bridge tests.
- Run the complete `nostr_client` package test suite.
- Run Flutter analysis for `lib`, `test`, and `integration_test`.

dcadenas:
Startup relay setup could reconnect every relay a second time and reset feed subscriptions.
This change keeps startup relay changes inside the startup flow, while preserving normal reconnects for relay changes made later.

## Description

The bug came from using the relay manager's ready flag as the full startup boundary.
That flag can turn true before the app-level Nostr initialization attempt finishes.

Relay changes now record whether the full Nostr service initialization is active when the change is observed.
Balanced lifecycle tracking covers initial startup, account changes, retries, and overlapping attempts without clearing early.
Initialization timeouts also report the last blocking stage so future startup failures identify where they stopped.
Post-startup relay changes still reconnect relays and reset feeds after the existing two-second debounce.

**Related Issue:** Closes #6100

## Out of Scope

None.
This does not change relay selection, the debounce duration, or normal post-startup reconnect behavior.

## Verification

The full app suite, package coverage gate, generated files, formatting, and analysis all pass.

- `very_good test --optimization --concurrency=4 --exclude-tags integration --test-randomize-ordering-seed random` (12,090 passed; 310 declared skips)
- `flutter test --coverage` in `mobile/packages/nostr_client` (287 passed; 85.44% coverage; 82% required)
- `dart format --output=none --set-exit-if-changed lib test integration_test`
- `flutter gen-l10n`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/providers/nostr_client_provider.dart
+++ b/mobile/lib/providers/nostr_client_provider.dart
@@ -160,22 +160,35 @@ final nostrInitializationInProgressProvider =
     );
 
 class NostrInitializationInProgress extends Notifier<bool> {
-  int _activeAttempts = 0;
+  final Map<NostrClient, int> _attemptsByClient = Map.identity();
 
   @override
   bool build() => false;
 
-  void begin() {
-    _activeAttempts++;
+  bool isClientInitializing(NostrClient client) {
+    return _attemptsByClient.containsKey(client);
+  }
+
+  void begin(NostrClient client) {
+    _attemptsByClient.update(
+      client,
+      (attempts) => attempts + 1,
+      ifAbsent: () => 1,
+    );
     state = true;
   }
 
-  void end() {
+  void end(NostrClient client) {
     if (!ref.mounted) return;
-    assert(_activeAttempts > 0, 'No Nostr initialization attempt is active');
-    if (_activeAttempts == 0) return;
-    _activeAttempts--;
-    state = _activeAttempts > 0;
+    final attempts = _attemptsByClient[client];
+    assert(attempts != null, 'No initialization attempt for this client');
+    if (attempts == null) return;
+    if (attempts == 1) {
+      _attemptsByClient.remove(client);
+    } else {
+      _attemptsByClient[client] = attempts - 1;
+    }
+    state = _attemptsByClient.isNotEmpty;
   }
 }
 
@@ -315,7 +328,7 @@ class NostrService extends _$NostrService {
       final initializationTracker = ref.read(
         nostrInitializationInProgressProvider.notifier,
       );
-      initializationTracker.begin();
+      initializationTracker.begin(client);
       initialization = initializationTracker;
       await _runClientInitialization(
         client,
@@ -356,7 +369,7 @@ class NostrService extends _$NostrService {
       _setSessionIdentityState(pubkey);
       _scheduleInitializationRetry(pubkey);
     } finally {
-      initialization?.end();
+      initialization?.end(client);
     }
   }
 
@@ -459,7 +472,7 @@ class NostrService extends _$NostrService {
       final initializationTracker = ref.read(
         nostrInitializationInProgressProvider.notifier,
       );
-      initializationTracker.begin();
+      initializationTracker.begin(client);
       initialization = initializationTracker;
       await _runClientInitialization(
         client,
@@ -507,7 +520,7 @@ class NostrService extends _$NostrService {
       _setSessionIdentityState(pubkey);
       _scheduleInitializationRetry(pubkey);
     } finally {
-      initialization?.end();
+      initialization?.end(client);
     }
   }
 
--- a/mobile/lib/providers/nostr_client_provider.dart
+++ b/mobile/lib/providers/nostr_client_provider.dart
@@ -310,10 +310,13 @@ class NostrService extends _$NostrService {
     required List<String> userRelayUrls,
     required String source,
   }) async {
-    final initialization = ref.read(
-      nostrInitializationInProgressProvider.notifier,
-    )..begin();
+    NostrInitializationInProgress? initialization;
     try {
+      final initializationTracker = ref.read(
+        nostrInitializationInProgressProvider.notifier,
+      );
+      initializationTracker.begin();
+      initialization = initializationTracker;
       await _runClientInitialization(
         client,
         userRelayUrls,
@@ -353,7 +356,7 @@ class NostrService extends _$NostrService {
       _setSessionIdentityState(pubkey);
       _scheduleInitializationRetry(pubkey);
     } finally {
-      initialization.end();
+      initialization?.end();
     }
   }
 
@@ -361,7 +364,7 @@ class NostrService extends _$NostrService {
     NostrClient client,
     List<String> userRelayUrls, {
     required String source,
-  }) {
+  }) async {
     final timeout = ref.read(nostrInitializationTimeoutProvider);
     final stopwatch = Stopwatch()..start();
     var stage = 'addingUserRelays';
@@ -372,20 +375,20 @@ class NostrService extends _$NostrService {
       category: LogCategory.system,
     );
 
-    return (() async {
-      if (userRelayUrls.isNotEmpty) {
-        final addedRelayCount = await client.addRelays(userRelayUrls);
-        Log.info(
-          '[NostrService] Initialization stage completed: source=$source, '
-          'stage=$stage, addedRelayCount=$addedRelayCount, '
-          'elapsedMs=${stopwatch.elapsedMilliseconds}',
-          name: 'NostrService',
-          category: LogCategory.system,
-        );
-      }
-      stage = 'client.initialize';
-      await client.initializeWithStageObserver(
-        (clientStage) {
+    try {
+      await (() async {
+        if (userRelayUrls.isNotEmpty) {
+          final addedRelayCount = await client.addRelays(userRelayUrls);
+          Log.info(
+            '[NostrService] Initialization stage completed: source=$source, '
+            'stage=$stage, addedRelayCount=$addedRelayCount, '
+            'elapsedMs=${stopwatch.elapsedMilliseconds}',
+            name: 'NostrService',
+            category: LogCategory.system,
+          );
+        }
+        stage = 'client.initialize';
+        client.initializationObserver = (clientStage) {
           stage = 'client.${clientStage.name}';
           Log.debug(
             '[NostrService] Initialization stage started: source=$source, '
@@ -393,25 +396,28 @@ class NostrService extends _$NostrService {
             name: 'NostrService',
             category: LogCategory.system,
           );
+        };
+        await client.initialize();
+        Log.info(
+          '[NostrService] Initialization completed: source=$source, '
+          'elapsedMs=${stopwatch.elapsedMilliseconds}',
+          name: 'NostrService',
+          category: LogCategory.system,
+        );
+      })().timeout(
+        timeout,
+        onTimeout: () {
+          throw TimeoutException(
+            'Nostr client initialization timed out: source=$source, '
+            'stage=$stage, userRelayCount=${userRelayUrls.length}, '
+            'elapsedMs=${stopwatch.elapsedMilliseconds}',
+            timeout,
+          );
         },
       );
-      Log.info(
-        '[NostrService] Initialization completed: source=$source, '
-        'elapsedMs=${stopwatch.elapsedMilliseconds}',
-        name: 'NostrService',
-        category: LogCategory.system,
-      );
-    })().timeout(
-      timeout,
-      onTimeout: () {
-        throw TimeoutException(
-          'Nostr client initialization timed out: source=$source, '
-          'stage=$stage, userRelayCount=${userRelayUrls.length}, '
-          'elapsedMs=${stopwatch.elapsedMilliseconds}',
-          timeout,
-        );
-      },
-    );
+    } finally {
+      client.initializationObserver = null;
+    }
   }
 
   void _scheduleInitializationRetry(String? pubkey) {
@@ -448,10 +454,13 @@ class NostrService extends _$NostrService {
         .toList();
     final client = _createClient(identity);
     final clientGeneration = _nextClientGeneration();
-    final initialization = ref.read(
-      nostrInitializationInProgressProvider.notifier,
-    )..begin();
+    NostrInitializationInProgress? initialization;
     try {
+      final initializationTracker = ref.read(
+        nostrInitializationInProgressProvider.notifier,
+      );
+      initializationTracker.begin();
+      initialization = initializationTracker;
       await _runClientInitialization(
         client,
         userRelayUrls,
@@ -488,6 +497,7 @@ class NostrService extends _$NostrService {
         category: LogCategory.system,
       );
       _disposeClient(client);
+      if (!ref.mounted) return;
       if (!_isActiveIdentity(pubkey, clientGeneration)) {
         return;
       }
@@ -497,7 +507,7 @@ class NostrService extends _$NostrService {
       _setSessionIdentityState(pubkey);
       _scheduleInitializationRetry(pubkey);
     } finally {
-      initialization.end();
+      initialization?.end();
     }
   }
 
--- a/mobile/lib/providers/nostr_client_provider.dart
+++ b/mobile/lib/providers/nostr_client_provider.dart
@@ -280,7 +280,11 @@ class NostrService extends _$NostrService {
     required String source,
   }) async {
     try {
-      await _runClientInitialization(client, userRelayUrls);
+      await _runClientInitialization(
+        client,
+        userRelayUrls,
+        source: source,
+      );
       // The provider can be disposed during the await (e.g. a rapid identity
       // switch rebuilds it, or a test container is torn down mid-init). Reading
       // `ref` after disposal throws, so bail before the ref-backed checks below.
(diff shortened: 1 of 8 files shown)


mobile/lib/providers/relay_providers.dart
```diff
@@ -184,9 +184,25 @@ void relaySetChangeBridge(Ref ref) {
...
 
+      final changeObservedDuringInitialization =
+          ref.read(nostrInitializationInProgressProvider) ||
+          !nostrService.isInitialized;
+
       // Debounce: collapse rapid changes into a single reset
       debounceTimer?.cancel();
       debounceTimer = Timer(const Duration(seconds: 2), () async {
+        if (changeObservedDuringInitialization ||
+            ref.read(nostrInitializationInProgressProvider) ||
```
hm21: These two are evaluated at fire time, but the timer isn't re-armed — the change is dropped, not deferred. A relay edit observed cleanly (`changeObservedDuringInitialization == false`) loses its `forceReconnectAll()` + `resetAndResubscribeAll()` if an init attempt happens to be running 2s later.

It's permanent, not delayed: `previousRelaySet` is already updated at line 185, so no later status emission re-triggers, and a *failing* retry never reassigns `state`, so the bridge never rebuilds. Nothing else re-issues `resetAndResubscribeAll()` automatically — the feed stays subscribed to the pre-change relay set until a manual "Retry connection", another relay edit, or a restart.

Reachable through the dominant failure path: `_relayManager.initialize()` swallows per-relay connect errors and still sets `_initialized = true`, so a timed-out client reports `isInitialized == true` while its retries are pending. And since `changeObservedDuringInitialization` drops just as permanently, the exposure is the whole retry attempt (up to the 90s timeout), not only the 2s debounce window.

Re-arm the timer instead of returning.

mobile/test/providers/nostr_service_identity_source_test.dart
```diff
@@ -799,6 +820,7 @@ void main() {
     test(
       'timed out startup relay setup retries automatically for same identity',
       () async {
+        final logCountBeforeTest = LogCaptureService().getRecentLogs().length;
```
hm21: `skip()` on a count from the global `LogCaptureService` — a 50k-entry ring buffer shared by every test in the VGV merged isolate. Once it saturates, the length stops growing and `skip(logCountBeforeTest)` yields nothing, so this fails seed-dependently.

The pattern used elsewhere (`divine_bloc_observer_test`, `foreground_idle_warmup_coordinator_test`, …) is `await LogCaptureService().clearAllLogs()` in `setUp`, then assert `.any(...)` directly.

mobile/packages/nostr_client/lib/src/nostr_client.dart
```diff
@@ -31,6 +31,48 @@ abstract class NostrClientStatisticsObserver {
...
+    void Function(NostrClientInitializationStage stage);
+
+final Object _initializationObserverZoneKey = Object();
+
+/// Diagnostic initialization entry point that reports blocking stage changes.
+///
+/// This is an extension instead of an additional parameter on
+/// [NostrClient.initialize] so test doubles and other implementations of
+/// [NostrClient] retain the existing initialization contract.
+extension NostrClientInitializationDiagnostics on NostrClient {
```
hm21: Passing the observer through a zone value is implicit and outlives the call: every timer and subscription `_relayManager.initialize()` creates keeps running in the forked zone, so the observer stays reachable long past initialization.

The stated rationale (test doubles keeping the existing contract) is really about mocktail stubs not matching a new named arg — that's shaping production code around stub syntax. A nullable field or constructor param on `NostrClient` would be explicit, and mocks would ignore it just the same.

hm21 (commented):
One real issue: in the bridge the fire-time re-check drops the reconnect instead of deferring it, so a legitimate post-startup relay change can permanently lose its feed reset. The rest is a test-robustness point and a design question on the zone-passed observer.

mobile/lib/providers/nostr_client_provider.dart
```diff
@@ -279,8 +310,15 @@ class NostrService extends _$NostrService {
     required List<String> userRelayUrls,
     required String source,
   }) async {
+    final initialization = ref.read(
+      nostrInitializationInProgressProvider.notifier,
+    )..begin();
```
realmeylisdev: This `ref.read` sits **outside** the `try`, which drops the disposal armor this method was deliberately given.

Before this PR the first `ref` touch was `ref.read(nostrInitializationTimeoutProvider)` *inside* `_runClientInitialization`, i.e. inside the `try` — so a disposed/rebuilt provider threw into the `catch`, which opens with `if (!ref.mounted) return;`. Your own comments on lines 322-325 and 339-341 describe exactly this hazard ("a rapid identity switch rebuilds it, or a test container is torn down mid-init"), and it's the #5600/#5602 regression class.

Now the throw escapes before the `try`, and `build()`'s `Future.microtask(...)` on line 246 has no `catchError` — so it becomes an unhandled async error and lands in the zone handler.

Reproduced both directions with the same test (`container.read(nostrServiceProvider); container.dispose();` then drain):

- this branch → `UnmountedRefException` thrown at `nostr_client_provider.dart:313`, unhandled
- `origin/main` → passes clean

```
Cannot use the Ref of nostrServiceProvider after it has been disposed.
package:riverpod/src/core/ref.dart 240:7   Ref._throwIfInvalidUsage
package:riverpod/src/core/ref.dart 607:5   Ref.read
...nostr_client_provider.dart 313:32       NostrService._initializeClient
...nostr_client_provider.dart 247:13       NostrService.build.<fn>
```

Moving the read inside the `try` restores main's behaviour (the existing `catch` + `if (!ref.mounted) return;` handles it). Worth a regression test for the dispose-before-microtask case, since CI is green today.

Same shape at line 451 in `_retryInitializeCurrentIdentity`, though that one is softened by `_enqueueClientLifecycleMutation`'s try/catch (logged rather than unhandled).

mobile/lib/providers/relay_providers.dart
```diff
@@ -184,9 +184,25 @@ void relaySetChangeBridge(Ref ref) {
...
+            !nostrService.isInitialized) {
```
realmeylisdev: This fire-time re-check can permanently swallow a relay change that has nothing to do with startup.

The flag is app-global: `begin()` is called by both `_initializeClient` and `_retryInitializeCurrentIdentity`, and a retry initializes a **different** client than the one this bridge watches via `ref.watch(nostrServiceProvider)`. So an unrelated retry starting inside the 2s window suppresses a change that was observed against a perfectly healthy client.

It's unrecoverable because line 185 advances `previousRelaySet = currentRelaySet` *before* the decision is made, so the set now matches and no later status emission re-triggers it — and the early `return` doesn't re-arm the timer.

A/B against the existing harness (identical relay add; only difference is an unrelated `begin()`/`end()` at t=0.5s):

| scenario | forceReconnectAll | resetAndResubscribeAll |
|---|---|---|
| control, no init in flight | 1 | 1 |
| unrelated init overlaps debounce | **0** | **0** (still 0 after 5 min) |

On main the suppression block doesn't exist, so this always fired — it's a strict regression.

This is reachable without an exotic setup: `Future.timeout` doesn't cancel the inner future, so after a startup timeout the original client can still finish `RelayManager.initialize()` and serve feeds while retries churn `begin()` for ~90s at a time. A user adding a relay in Settings during that window silently gets no reconnect and no feed reset — and `RelaySettingsCubit.addRelay` doesn't compensate (it only calls `refreshRelays()`; `resetAndResubscribeAll()` is behind the manual "Retry connection" button).

Two directions that'd fix it: re-arm/reschedule instead of `return`, or latch a pending-reconnect that fires when `inProgress` drops — and only advance `previousRelaySet` when the work actually runs.

Separately, the third disjunct (`!nostrService.isInitialized`, line 196) looks like dead code: `_initialized` is monotonic false→true within a client's life, so if it's true here it was true at observation and line 194 already short-circuited. Flipping it back requires disposing the client, which rebuilds the bridge and cancels this timer.

mobile/test/providers/nostr_service_identity_source_test.dart
```diff
@@ -819,6 +841,20 @@ void main() {
         expect(factory.callCount, equals(2));
         expect(factory.addRelaysPubkeys, equals([pubkeyA, pubkeyA]));
         expect(factory.initializePubkeys, equals([pubkeyA]));
+        final testLogs = LogCaptureService().getRecentLogs().skip(
+          logCountBeforeTest,
+        );
```
realmeylisdev: `skip(logCountBeforeTest)` against `LogCaptureService` is a merged-isolate flake waiting to happen.

It's a process-global singleton ring buffer (`_instance ??=`), capped at 50k with `removeFirst()` eviction, and capture is unconditional — `unified_logger.dart:187` runs outside the `kDebugMode` block and ignores level/category filters ("Memory capture must be comprehensive"). Under `very_good test --optimization` every suite shares it, and there's no reset in `flutter_test_config.dart`.

So once the buffer is at capacity, `logCountBeforeTest == 50000`, evictions keep it at 50000, and `.skip(50000)` yields an empty iterable — the assertion fails for any seed where no clearing suite runs in between. Even a few evictions mid-test skip past your target entry, which is only ~2nd of the ones this test writes.

The repo already has the eviction-safe pattern — `await LogCaptureService().clearAllLogs()` in `setUp` — in 9 files (`divine_bloc_observer_test.dart:42`, `divine_auth_cubit_test.dart:55`, …). Worth matching that here and dropping the skip arithmetic.

mobile/test/unit/providers/relay_set_change_bridge_test.dart
```diff
@@ -227,5 +229,59 @@ void main() {
...
+          container.read(nostrInitializationInProgressProvider.notifier).end();
+          async.elapse(const Duration(seconds: 1));
+          async.flushMicrotasks();
+
+          verifyNever(() => mockNostrClient.forceReconnectAll());
+          verifyNever(() => mockVideoEventService.resetAndResubscribeAll());
+          container.dispose();
+        });
+      },
+    );
```
realmeylisdev, line 336: Both new tests only exercise the **observation-time** capture — by the time the debounce fires, `inProgress` is false and `isInitialized` is true in each. You can delete both fire-time re-checks on lines 195-196 and this suite stays green.

That's the branch carrying the swallowed-change behaviour I flagged in `relay_providers.dart`, so it's worth pinning whichever way you land: the `{captured=false, inProgress=true at fire}` case is expressible in this harness already (add the status change with everything healthy, `elapse(1s)`, `notifier.begin()`, `elapse(1s)`).

mobile/packages/nostr_client/lib/src/nostr_client.dart
```diff
@@ -379,15 +421,28 @@ class NostrClient {
   /// On failure, [ready] resolves with the same error so consumers awaiting
   /// readiness fail fast instead of hanging until a wall-clock timeout.
   Future<void> initialize() async {
+    final onStageChanged =
+        Zone.current[_initializationObserverZoneKey]
+            as NostrClientInitializationObserver?;
     try {
       // Signer is the single source of truth for the public key.
+      onStageChanged?.call(
```
realmeylisdev: Minor: all four `onStageChanged?.call(...)` sites are inside the `try` whose `catch` does `_readyCompleter.completeError(e, st)` and rethrows — so a throwing observer would fail initialization and poison `ready` for that client's lifetime.

Not a live bug (the only observer is a `Log.debug` closure, and it matches the existing unguarded `statisticsObserver` convention), but a diagnostics hook that can break what it's diagnosing is worth either a try/catch around the call or a "must not throw" note on the typedef.

Also worth one doc sentence on the extension: timers/socket callbacks created during `initialize()` inherit the zone and retain the observer for the client's connected lifetime.

realmeylisdev (commented):
Nice diagnosis on the root cause — the ready-flag-isn't-the-startup-boundary insight is right, and the counter (vs a bool) is genuinely load-bearing given attempts overlap.

Two things I reproduced locally that I think block merge: the `ref.read` before the `try` now throws unhandled on a disposed provider (passes on main, fails here), and the bridge can permanently swallow a healthy relay change when an unrelated retry overlaps the debounce. Details inline.

rabble (approved):
**Reviewed with full checkout + verification.** ✅ Approve.

Well-targeted startup-churn fix. The hand-written keepAlive `nostrInitializationInProgressProvider` (ref-counting `Notifier<bool>`) is begin/end-wrapped around both the primary and retry init paths, and `relaySetChangeBridge` skips the forced reconnect when a relay-set change is observed during init or when init is still in progress at debounce-fire time.

Checked the two things that could go wrong:
- **Gate is not permanent suppression.** `nostrService.isInitialized` is a live getter and the in-progress provider is re-read both at observe time and inside the 2s debounce callback, so a real relay change after startup reconnects normally. The upgraded happy-path test now asserts `verify(forceReconnectAll).called(1)`, so over-suppression would go red.
- **.g.dart matches source.** Only the two existing `@riverpod` provider hashes changed (NostrService, relaySetChangeBridge) — the signature of a body-only edit after `build_runner`. The new provider is a hand-written `NotifierProvider`, so correctly has no `.g.dart` entry.

No `Future.delayed` hacks in production (real `Timer` debounce + `Stopwatch` telemetry). Ref-count lifecycle nests correctly across primary+retry; `end()` guards `!ref.mounted`; keepAlive keeps the counter alive across NostrService rebuilds.

Verification (local):
- `flutter analyze lib/providers test/providers test/unit/providers` → No issues found!
- `flutter test` provider suites → 30 tests passed (incl. 3 new).
- `packages/nostr_client` → 151 tests passed (incl. new stage-observer test).

Minor (non-blocking): a relay-set change observed *during* init is permanently skipped for that specific change even after init completes — intended tradeoff, init connects to the current set anyway, and both new tests deliberately pin it.

mobile/lib/providers/relay_providers.dart
dcadenas: Addressed in rebased commits 431de20cf and d03e2fc0d: reset-required changes remain pending and re-arm while the active client initializes, so a clean relay edit is deferred instead of dropped. The focused suite pins the fire-time overlap and verifies one reconnect/feed reset after initialization.

mobile/test/providers/nostr_service_identity_source_test.dart
dcadenas: Addressed in rebased commit 431de20cf: the test clears LogCaptureService before each case and asserts directly against the fresh entries, removing ring-buffer-length skip arithmetic.

mobile/packages/nostr_client/lib/src/nostr_client.dart
dcadenas: Addressed in rebased commit 431de20cf: initialization stage reporting now uses an explicit nullable NostrClient initializationObserver that the provider clears in finally, rather than an inherited zone value.

mobile/lib/providers/nostr_client_provider.dart
dcadenas: Addressed in rebased commit 431de20cf: the initialization-tracker ref read is inside the guarded try, and regression coverage disposes the container before the scheduled initialization microtask to verify no unhandled UnmountedRefException escapes.

mobile/lib/providers/relay_providers.dart
dcadenas: Addressed in rebased commits d03e2fc0d, 93e5258ef, and 2a57d3b2c: initialization is tracked per client, relay reset intent remains sticky, and replacement work is scoped by environment and session identity. An unrelated retry can no longer suppress or steal the active client's reset.

mobile/test/providers/nostr_service_identity_source_test.dart
dcadenas: Addressed in rebased commit 431de20cf: LogCaptureService is cleared in setUp and the assertions no longer call skip(logCountBeforeTest), so ring-buffer saturation cannot hide the target logs.

mobile/test/unit/providers/relay_set_change_bridge_test.dart
dcadenas: Addressed in rebased commit 431de20cf: the suite now exercises a healthy edit observed before initialization begins, then expires the debounce while that same client is initializing and verifies the reset is deferred and later runs exactly once.

mobile/packages/nostr_client/lib/src/nostr_client.dart
dcadenas: Addressed in rebased commit 431de20cf: diagnostics observer calls are guarded so exceptions cannot poison initialization, and the explicit field is documented and cleared by the provider while async initialization may still be pending.

Issue #5751, opened 2026-07-02
fix(likes): addressable like count still diverges from "Liked by" on seeded surfaces (stats.reactions + seed precedence)


realmeylisdev:
## Context

Follow-up from the [#5748](https://github.com/divinevideo/divine-mobile/pull/5748) review, which aligned `getLikeCount` / `getLikeCounts` with `fetchEventLikers` for addressable videos so the **relay-resolved** count and the "Liked by" list can't diverge. This tracks the residual paths where the **displayed** count can still diverge from the list — the likely remaining cause if #5742 ("5 likes" vs many likers) reproduces after #5748 ships.

## Residual 1 — funnelcake `stats.reactions` semantics + seed precedence

On seeded surfaces (feed / profile payload) the displayed like count is the seed, not the resolved count:

- `mobile/lib/blocs/video_interactions/video_interactions_bloc.dart:194` — `final likeCount = preFetchLikeCount ?? fetchedLikeCount;`. The corrected `fetchedLikeCount` (now `getLikeCount(_eventId, addressableId: _addressableId)`, `video_interactions_bloc.dart:171-174`) is only used when there is **no** seed.
- The seed reaches the surface via `VideoEventService.getLikeCounts` → max-merge → seed pipeline, sourced from funnelcake `stats.reactions`.
- Funnelcake `stats.reactions` counts distinct kind-7 event ids **by e-tag only** — no a-tag union, no `-` downvote / kind-5 deletion / moderation filtering (`engagement_counts` MV, funnelcake migration `000092`). So the seed can differ from the resolved list in **both** directions, and the max-merge never lowers an inflated seed.
- Stale comment: `video_interactions_bloc.dart:189-193` justifies seed precedence with "relay COUNT … bogus spikes / unrelated historical reactions." That no longer describes `fetchedLikeCount` on the addressable path — after #5748 it is the same filtered distinct-liker set as the list, not a raw relay COUNT.

Directions to weigh:
- Fix backend `stats.reactions` semantics to match the resolver (a-union + downvote/deletion/moderation filtering), **and/or**
- Change seed precedence for addressable videos so the resolved count wins, **and/or**
- Let the max-merge lower an inflated seed rather than only raise it.

This needs a product/architecture call on which number is canonical (mobile resolved distinct-liker set vs funnelcake `stats.reactions`) — cc @NotThatKindOfDrLiz.

## Residual 2 — deletion REQ frame size (surfaced in the #5748 re-review)

[...]

PR #5748, opened 2026-07-02
fix(likes): align addressable liker counts

NotThatKindOfDrLiz:
## Summary
- Align addressable video like counts with the same resolved active-liker set used by the "Liked by" screen.
- Share e-tag/a-tag union, reaction-id dedupe, pubkey dedupe, downvote filtering, same-author deletion filtering, and block filtering between count and list paths.
- Add full event/addressable identifiers to video engagement loading logs so future support reports can be traced without contacting the user.

## Root Cause
Addressable video counts used a different source of truth than the "Liked by" screen. Counts took the maximum of separate e-tag and a-tag relay COUNT responses, while the list queried actual reaction events across both tags, deduped by pubkey, and filtered inactive reactions. That made the list legitimately able to exceed the displayed count.

## Verification
- flutter test test/src/likes_repository_test.dart from mobile/packages/likes_repository
- flutter test test/blocs/video_engagement/video_engagement_bloc_test.dart from mobile
- flutter test test/screens/video_engagement/video_engagement_list_screen_test.dart from mobile
- flutter analyze from mobile
- pre-push analyzer and changed-file tests

Closes #5742

Issue #568, opened 2025-12-10
[Feature] 5.4 Follow/Unfollow (NIP-02)


Statyk7:
## Feature: 5.4 Follow/Unfollow (NIP-02)

**Section:** 5. Social Features
**Status:** Implemented
**Priority:** P0

PR #6102, opened 2026-07-15
fix(likes): floor the addressable like count against the relay resolve (#6022)

### LINKED ISSUE #6022: fix(likes): add a like-count floor so an unseeded funnelcake stat cannot lower the count
## Summary

When funnelcake returns no stats row for a video id, the mobile like-count seed is null and the headline falls through to the relay `e`+`a` union, which can undercount likes stranded on superseded ids after an edit. This is the mobile-side defensive gap that #5751 left open when it was closed: there is no floor preventing a missing/absent backend stat from lowering a real count.

## Evidence

- The displayed count uses `final likeCount = preFetchLikeCount ?? fetchedLikeCount;` (`mobile/lib/blocs/video_interactions/video_interactions_bloc.dart:195`). When the funnelcake seed is present (post-#568 it carries the coordinate-union count) this is correct and the relay union is discarded. When the seed is **null**, the headline is the relay union `M`, which misses `e`-only reactions stranded on superseded ids.
- The seed is null only when both components are null: `liveLikeCountSeed(video) => _sumNullableCounts(video.originalLikes, video.nostrLikeCount)` returns null only if both are null (`mobile/lib/widgets/video_feed_item/live_engagement_counts.dart:6-19`).
- `nostrLikeCount` stays null when funnelcake omits the id: `if (stats == null) return video;` (`mobile/packages/videos_repository/lib/src/videos_repository.dart:409`). The home feed always seeds `nostrLikeCount` from funnelcake reactions, so it is safe; the exposure is videos funnelcake does not index and hashtag WebSocket-only items.
- `VideoEventService._applyLikeCountToVideo` (`mobile/lib/services/video_event_service.dart:4880`, guarded only by a zero-check at `:4834`) **direct-replaces** the count with the relay union on its internal `_eventLists`/bucket cache, rather than max-merging — unlike the enrichment paths that use `mergeNullableEngagementMax` (`mobile/packages/videos_repository/lib/src/video_merge_helpers.dart:41-44`). Narrow surfaces that read those buckets (hashtag WebSocket-only) can therefore render a lowered count.

## Suggested fix

Make the count a floor rather than a straight fallback:

- At `video_interactions_bloc.dart:195`, use `max(seed, relayUnion)` (treating null seed as 0) so a missing funnelcake row can never lower a real relay-resolved count, and a relay undercount can never lower a real seed.
- Switch `VideoEventService._applyLikeCountToVideo` to `mergeNullableEngagementMax` for consistency with the other enrichment paths, so bucket-fed surfaces can't be lowered either.

This is the mobile-side piece #5751 identified and left open after it was closed.

## Test plan

`video_interactions_bloc` test:
- seed null + relay union `M` → headline `M`;
- seed `N` + relay union `M < N` → headline `N` (never lowered);
- seed `0` + relay union `M > 0` → headline `M` (the historical "0 seed wins" case no longer zeroes a real count).

---

### Context

Follows from #4432 (lost likes on edit). That issue's core symptom — the displayed like count collapsing after an edit — was a backend counting bug and was fixed by divinevideo/divine-funnelcake#568 (migration `000180`, coordinate-aware counting), deployed to production 2026-07-10 06:42 UTC. Post-fix the seed carries the correct count on seeded surfaces; this issue adds a mobile-side floor so an *unseeded* or absent backend stat cannot re-lower a real count.

realmeylisdev:
## Description

The displayed like count seeded `preFetchLikeCount ?? fetchedLikeCount`, which makes a present funnelcake seed win outright — a floor **and** a ceiling. Two real gaps hid behind that on addressable videos:

- a `reactions:0` seed zeroed a real relay count, and
- a seed that undercounts (funnelcake ingestion/merge lag) could never be corrected upward by the cleaner #5748 distinct-liker resolve.

This replaces the combine in `VideoInteractionsBloc` with an **addressable-gated, Nostr-portion floor**: `max(seed, originalLikes + fetchedNostr)`.

- **Only ever raises, never lowers** — so it preserves the #568 / #4432 edit-recovery (the funnelcake coordinate-union seed stays the base).
- **Addressable-only** — the non-addressable `getLikeCount` is a raw, unfiltered relay tally (counts downvotes / duplicate reactors / deleted reactions), so it must not raise the count; that path keeps the current `??`.
- **Keeps the archival Vine baseline (`originalLikes`) out of the Nostr floor** so a large archival number can't mask newly resolved Nostr likes. `originalLikes` is threaded into the bloc as `archivedLikeCount`.

Also switches `VideoEventService._applyLikeCountToVideo` from direct-replace to `mergeNullableEngagementMax`, so a later, lower re-fetch can't lower an already-applied hashtag/profile bucket count — aligning it with the #3384 policy the profile enrichment paths already use (bounded: these buckets start `null`, so this normally applies `null → N` on first fetch).

**Deliberate behavior change:** for an addressable video, a higher clean relay resolve now *raises* the seed instead of being discarded. This was chosen over #5751's resolved-wins direction, which would regress #4432 post-#568 (the mobile relay resolve only e-filters the current event id and misses superseded-id likes on edited videos). A `max()` floor cannot pull down funnelcake's *spurious* inflation (downvotes / duplicates / deletions) — that stays a backend concern (funnelcake #626) and the floor never makes the displayed number worse than today.

**Related Issue:** Closes #6022

## Out of Scope

- funnelcake `stats.reactions` spurious-inflation correction (funnelcake #626)
- `originalLikes` archival-vs-Nostr reconciliation
- the "Liked by" list undercount of pre-a-tag e-only reactions (#4432 R4)
- comments/reposts (this is likes-only)

## Verification

- `flutter test test/blocs/video_interactions/video_interactions_bloc_test.dart test/services/video_event_service_like_count_addressable_test.dart` — all pass
- Regression sweep: full `video_event_service` suite, `video_feed_bloc`, pooled/hashtag feed screen tests, `list_providers` — all pass
- Pre-push hook (analyze + changed-file tests) — pass

## Type of Change

### DIFF
--- a/mobile/lib/services/video_event_service.dart
+++ b/mobile/lib/services/video_event_service.dart
@@ -50,6 +50,8 @@ import 'package:openvine/utils/log_tag_sanitizer.dart';
 import 'package:profile_repository/profile_repository.dart';
 import 'package:unified_logger/unified_logger.dart';
 import 'package:video_event_cache/video_event_cache.dart';
+import 'package:videos_repository/videos_repository.dart'
+    show mergeNullableEngagementMax;
 
 /// Pagination state for tracking cursor position and loading status per subscription
 class PaginationState {
@@ -4876,8 +4878,17 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
 
     final videoEvent = eventList[index];
 
-    // Update the video with the like count
-    final updatedVideo = videoEvent.copyWith(nostrLikeCount: likeCount);
+    // #6022: max-merge instead of direct-replace so a later, lower re-fetch
+    // can't lower an already-applied bucket count. Bounded — these buckets are
+    // relay-parsed and start with a null nostrLikeCount, so this normally
+    // applies null->N on first fetch. Aligns with the #3384
+    // mergeNullableEngagementMax policy used by the profile enrichment paths.
+    final updatedVideo = videoEvent.copyWith(
+      nostrLikeCount: mergeNullableEngagementMax(
+        videoEvent.nostrLikeCount,
+        likeCount,
+      ),
+    );
     eventList[index] = updatedVideo;
 
     // Also update in keyed buckets if applicable
@@ -5855,6 +5866,15 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
   Future<void> flushPendingLikeCountBatchForTesting() {
     return _executeLikeCountBatchFetch();
   }
+
+  /// Apply a resolved like count to a cached video, bypassing the batch and
+  /// debounce, to exercise the max-merge semantics of [_applyLikeCountToVideo].
+  @visibleForTesting
+  bool applyLikeCountForTesting(
+    String videoId,
+    int likeCount,
+    SubscriptionType subscriptionType,
+  ) => _applyLikeCountToVideo(videoId, likeCount, subscriptionType);
 
   /// Run automatic diagnostics when feed fails to load events
   /// This logs relay status, connection info, and tests direct queries to help debug
--- a/mobile/lib/services/video_event_service.dart
+++ b/mobile/lib/services/video_event_service.dart
@@ -50,8 +50,6 @@ import 'package:openvine/utils/log_tag_sanitizer.dart';
 import 'package:profile_repository/profile_repository.dart';
 import 'package:unified_logger/unified_logger.dart';
 import 'package:video_event_cache/video_event_cache.dart';
-import 'package:videos_repository/videos_repository.dart'
-    show mergeNullableEngagementMax;
 
 /// Pagination state for tracking cursor position and loading status per subscription
 class PaginationState {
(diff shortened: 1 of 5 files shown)


mobile/lib/services/video_event_service.dart
```diff
@@ -4876,8 +4878,17 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
...
 
-    // Update the video with the like count
-    final updatedVideo = videoEvent.copyWith(nostrLikeCount: likeCount);
+    // #6022: max-merge instead of direct-replace so a later, lower re-fetch
+    // can't lower an already-applied bucket count. Bounded — these buckets are
+    // relay-parsed and start with a null nostrLikeCount, so this normally
+    // applies null->N on first fetch. Aligns with the #3384
+    // mergeNullableEngagementMax policy used by the profile enrichment paths.
+    final updatedVideo = videoEvent.copyWith(
+      nostrLikeCount: mergeNullableEngagementMax(
```
hm21: Unlike the bloc floor, this max-merge isn't addressable-gated — so "never-lower" now lives in the persisted `nostrLikeCount` for every video, not just the addressable ones #6022 is scoped to. Consequence: a genuine decrease (an unlike, spam-like cleanup, others un-reacting) no longer corrects down on the next batch the way pre-PR direct-replace did — `max()` freezes the session high. Concretely, after the user's own unlike the repo cache drops to N-1, but a later batch re-applies `max(N, N-1)=N`, so the count stays one too high on revisit. Intended? If so, worth addressable-gating this to match the bloc (or a note here). Related: the `_mergeUpdatedVideo` comment at L4165 ("will overwrite this carried value … no risk of a stale count persisting") is now false under max-merge — the doc-refresh commit missed it.

hm21 (approved):
Core logic is correct: the addressable floor is `originalLikes + max(seedNostr, fetchedNostr)` with no double-count, `getLikeCount` is Nostr-only on the addressable path, and the optimistic-tap reconcile guard holds. Base is fresh, single PR, and the tests pin the raise / never-lower / archival-separation cases well. Non-blocking below — mainly want to confirm one intent. Optional: the VES max-merge test only covers the addressable path; a non-addressable case would pin the riskier raw-tally branch.

mobile/lib/services/video_event_service.dart
realmeylisdev: Confirmed real, and broader than addressable-gating would fix: `addressableId` is never actually null for anything reaching this path (kind 34236 always gets `vineId` falling back to `event.id` in `VideoEvent.fromNostrEvent`), so gating on it would've been a no-op. Reverted `_applyLikeCountToVideo` to direct-replace in dcb18a4af5 — #6022's floor is already a correctly-scoped one-time comparison in `VideoInteractionsBloc._onFetchRequested`, independent of this cache, so it doesn't need protection here. Also fixed the L4165 comment and swapped the test to pin the lower-resolve self-correcting behavior.

realmeylisdev:
@hm21 fixed in dcb18a4af5 — reverted the VES-cache max-merge to direct-replace instead of addressable-gating it, since `addressableId` turned out to always be non-null on this path (gating would've been a no-op). Details in the thread reply.

PR #6104, opened 2026-07-15
fix(follow): drop invalid pubkeys from the contact list

### LINKED ISSUE #6105: fix(follow): drop invalid pubkeys from the contact list
Auto-created from PR #6104.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6104
**Branch:** `fix/follow-list-invalid-pubkey`

### LINKED ISSUE #6109: fix(follow): REST bootstrap truncates at 100 and can permanently destroy follows 101+
The REST bootstrap in `FollowRepository` is silently capped at 100 follows, and a truncated result suppresses the authoritative relay Kind 3 query for the whole session. If a follow lands before the Kind 3 subscription repairs the list, `_broadcastContactList` republishes the truncated list and **follows 101+ are destroyed for every client and device**.

Surfaced while reviewing #6104 (thanks @realmeylisdev for spotting the clamp). It is unrelated to that PR's sanitization — it is pre-existing on `main` and needs its own design decision.

## The three steps

**1. `getFollowing` is server-clamped to 100.** `_loadFromRestApi` asks for `limit: 5000`, but the API caps it. Verified live:

```
$ curl "https://api.divine.video/api/users/<64-hex>/following?limit=5000"
{"following":[...],"total":32,"offset":0,"limit":100}
                                          ^^^^^^^^^^ echoed back clamped
```

`PaginatedPubkeys.total` is documented as "may exceed [pubkeys] length" — and `_loadFromRestApi` reads neither `.total` nor `.hasMore`.

**2. A truncated result suppresses the authoritative source.** `initialize()` gates each fallback on `if (_followingPubkeys.isEmpty)`:

```
1. LocalStorage
2. PersonalEventCache
3. REST API      <- only if (_followingPubkeys.isEmpty)
4. relay Kind 3  <- only if (_followingPubkeys.isEmpty)   << never runs after step 3 returns 100
```

The first non-empty source wins. A 100-entry truncation is non-empty, so the relay Kind 3 — the only authoritative copy — is never queried for that session.

**3. A racing follow makes it permanent.** The Kind 3 subscription normally repairs the list. But if a follow lands first, `_broadcastContactList` sets `_currentUserContactListEvent` to the newer 101-entry event it just published. The authoritative 249-entry event then arrives and fails the `event.createdAt > _currentUserContactListEvent!.createdAt` guard in `_processContactListEvent`, so it is silently discarded.

The catastrophic-reduction guard does not save this: it only guards the **inbound** direction, and only fires above 50% loss.

Net effect: local stays truncated for the session, the relay's newest Kind 3 carries 101 p-tags, and follows 101–249 are gone everywhere.

## Reproduction

Deterministic with mocks — three cases, all confirmed:

1. A truncated REST result suppresses the relay Kind 3 query (`count == 100`).
2. A follow landing before the relay Kind 3 permanently discards follows 101+ (`followingCount == 101` after the 249-entry authoritative event arrives). **Data loss.**
3. Control: with no racing follow, the relay Kind 3 does repair to 249.

Only case 3 matches the assumption that the subscription "corrects it shortly after".

## Impact

Any account with more than 100 follows. The reported account in #6105 has 249.

This is also not hypothetical in the smaller case: an account with 40 follows was republished down to 32 on `wss://relay.divine.video` (`2026-07-15T09:13Z`), losing 15, while five other relays still held the intact 40-tag event. Same class — a stale/incomplete source became the base for a destructive full-list republish.

## Fix options

Needs a design decision, not a patch:

1. Paginate `getFollowing` with `offset` until `total` is reached.
2. Drop the REST bootstrap for accounts over the cap (use it only as a fast first paint, never as a publish base).
3. Make step 4 unconditional and union-merge the relay Kind 3 rather than gating it on `isEmpty`.
4. Independently: add a floor guard to the **publish** path, so `_broadcastContactList` refuses to shrink the published list without an explicit unfollow. The existing guard is inbound-only.

(3) and (4) look load-bearing regardless of which bootstrap strategy wins — a destructive republish should never be sourced from a derived, lagging index.

hm21:
Closes #6105
Closes #6109

One user reported they could not follow anyone. Every tap failed with `Invalid argument (publicKey): Invalid key: "nos"`.

## Description

Their Kind 3 contact list on `wss://relay.divine.video` carries a `["p","nos"]` entry — not a valid pubkey. It originates from a parser that read `tag[1]` of a `["client","nos","https://nos.social"]` tag (written by the Nos app) without filtering on `p`; our app then republished that value as a `p` tag, baking it into the event.

`_broadcastContactList` rebuilds the **entire** `ContactList` from the in-memory list, so `Contact`s constructor threw on that single entry and aborted the publish before it reached a relay. The optimistic update was rolled back, which is why every follow **and** unfollow failed — permanently, for every target, not just for one profile.

### Why sanitize in this many places

Every source of the following list is untrusted: Kind 3 events are written by any Nostr client, and both the REST API and our caches are *derived* from those events. So the list is filtered on the way in — LocalStorage, PersonalEventCache, REST API, network Kind 3, and `mergeFollows` — and once more at the publish site.

The publish-site check is not redundant with those five. `executeFollowAction` appends to the in-memory list without validating — it replays actions queued while offline — so the publish site is the only thing between a queued invalid entry and a fatal `Contact` constructor. It is also the one place where an invalid entry is fatal rather than merely wrong.

Because the broadcast rebuilds the full list from the sanitized in-memory list, the next follow republishes a clean Kind 3 and overwrites the poisoned event. Affected accounts **self-heal on the next follow from mobile**; no manual data repair at any point. The scope matters: other divine Kind 3 writers append to a cloned tag array rather than rebuilding from a pubkey list, so a follow performed elsewhere against a stale copy can republish the entry. Mobile stays correct regardless, because it re-sanitizes on read.

The publish-site sanitize also emits the corrected list, so observers of `followingStream` cannot keep an entry that `followingPubkeys` has already dropped.

### Fail fast on the way in, self-heal on the way out

`follow()` now rejects an invalid pubkey up front with an `ArgumentError` rather than accepting it and failing later at publish time. This is safe for existing callers: the UI paths catch generically and surface the existing `toggleFailure` status, and the one direct `follow()` caller (`NotificationFeedBloc`) routes it to `Reportable`/Crashlytics.

On the online publish path that routing is unchanged from `main`, where the identical `ArgumentError` already came from `Contact`'s constructor. Two paths do change, because the guard sits above both the `contains()` early-return and the offline branch: previously an invalid pubkey already in the list returned silently, and an invalid pubkey while offline queued and returned normally. Both now throw.

`executeFollowAction` deliberately does **not** throw. It replays queued actions through `PendingActionService`, whose `_isRetriableError` defaults to retriable, so throwing would burn five backoff retries and then wedge the action as permanently failed. Sanitizing and self-healing is the better outcome for a replay path.

### Never republish from a derived source (#6109)

Reviewing the sanitize sites surfaced a second, unrelated way the same list gets destroyed — pre-existing on `main`, and the one that actually bites.

The relay Kind 3 query was gated on `_followingPubkeys.isEmpty`, so the first non-empty derived source won and the only authoritative copy was never read. LocalStorage, PersonalEventCache and the REST index are all derived from that event and all lag. Because the next follow rebuilds and republishes the **entire** list, whatever the derived source returned became the published truth, and every follow it was missing was destroyed for every client.

Seen twice: an account republished 40 follows down to 32 against a lagging index, and `getFollowing` is server-clamped to 100 whatever `limit` asks for, so any account past the cap bootstraps truncated. The catastrophic-reduction guard covers neither — it is inbound-only and fires above 50% loss, where these are 20% and none.

Step 4 now always runs, and the REST bootstrap pages on `offset` rather than trusting a limit the server ignores. REST stays a fast first paint; it is no longer the last word.

**No publish-side floor guard**, despite it being the obvious symmetric fix. `_currentUserContactListEvent` and `_followingPubkeys` are only ever written together from the same source, so a guard comparing them cannot fire — I wrote it, and coverage dropped to 99.13% with the branch unreachable. Making it fire would need a session high-water mark, which would resurrect follows unfollowed on another device. The inbound guard owns that problem.

## Out of Scope

The original parser that turned `["client","nos",...]` into a `p` entry is not in the current `follow_repository` code — `_processContactListEvent` and `ContactList.getContactInfoFromTags` both filter on `p` correctly. It is either an older app version or server-side in funnelcake. This PR makes the client robust against the bad data that already exists in the wild; tracking down the original writer is separate work.

## Verification

Reproduced against the real event fetched from `wss://relay.divine.video` (249 p tags, one of them `["p","nos"]`), then pinned with regression tests.

- `flutter test` — 213 passed
- `flutter test --coverage` — 100.00% on `follow_repository.dart` (package gate is the VGV 100% default)
- `dart format` — clean

New regression coverage: cached list holding an invalid pubkey still publishes a clean list, invalid entries dropped from LocalStorage, from remote Kind 3 events, from a cached Kind 3 event, from the REST API, and from `getOthersFollowing`; `follow()` rejecting an invalid pubkey; and the publish-site sanitize emitting the corrected list. For #6109: the relay Kind 3 being consulted even when local storage already answered, and the REST bootstrap paging past the server's 100 cap.

Every new test was verified by mutation: reverting the hunk it covers turns it red. One exception, by design — `follow succeeds and publishes a clean list when the cached list holds an invalid pubkey` is an end-to-end guard covered by two hunks in defence-in-depth (the LocalStorage and publish-site sanitizes); either alone keeps it green, and it goes red against `main` and when both are reverted.

One caveat worth recording: `throws when following an invalid pubkey` **passes on `main`**, because there the same `ArgumentError` surfaces from `Contact`'s constructor and the rollback leaves the count at zero. It pins the guard against removal; it does not reproduce the production failure.

## Type of Change

### DIFF
--- a/mobile/packages/follow_repository/lib/src/follow_repository.dart
+++ b/mobile/packages/follow_repository/lib/src/follow_repository.dart
@@ -1243,10 +1243,13 @@ class FollowRepository {
         await _loadFromRestApi();
       }
 
-      // 4. If still empty, query relays for kind 3 contact list directly.
-      // The REST API may not have indexed the user's contact list yet,
-      // but the relay has the authoritative kind 3 event.
-      if (_followingPubkeys.isEmpty && _nostrClient.hasKeys) {
+      // 4. Always query relays for the authoritative kind 3 event, even when
+      // a derived source already answered. LocalStorage, PersonalEventCache
+      // and the REST index are all derived from this event and can lag or
+      // truncate; because the next follow rebuilds and republishes the whole
+      // list, accepting an incomplete one destroys every follow it is
+      // missing (#6109).
+      if (_nostrClient.hasKeys) {
         await _loadFromRelay();
       }
 
@@ -1652,11 +1655,19 @@ class FollowRepository {
     }
   }
 
+  /// Upper bound on REST bootstrap paging, mirroring the ceiling the old
+  /// single `limit: 5000` request implied.
+  static const _restFollowingMaxPubkeys = 5000;
+
   /// Load following list from REST API (funnelcake) for fast bootstrap.
   ///
   /// Called only when local cache and PersonalEventCache are both empty
   /// (e.g., first login or after identity change cleanup). This provides
   /// the following list before the WebSocket subscription can deliver it.
+  ///
+  /// The result is a fast first paint, never the last word: it comes from a
+  /// derived index that lags, so [initialize] always follows it with the
+  /// authoritative Kind 3 from the relay.
   Future<void> _loadFromRestApi() async {
     try {
       final currentUserPubkey = _nostrClient.publicKey;
@@ -1672,11 +1683,24 @@ class FollowRepository {
         category: LogCategory.system,
       );
 
-      final result = await _funnelcakeApiClient.getFollowing(
-        pubkey: currentUserPubkey,
-        limit: 5000,
-      );
-      final pubkeys = _sanitizePubkeys(result.pubkeys, source: 'REST API');
+      // The server clamps `limit` to 100 however much we ask for, so page on
+      // `offset` instead of requesting one large batch. Asking for 5000 and
+      // silently receiving 100 is what let a truncated list become the base
+      // for the next publish (#6109).
+      final collected = <String>[];
+      while (collected.length < _restFollowingMaxPubkeys) {
+        final page = await _funnelcakeApiClient.getFollowing(
+          pubkey: currentUserPubkey,
+          offset: collected.length,
+        );
+
+        if (page.pubkeys.isEmpty) break;
+        collected.addAll(page.pubkeys);
+
+        if (!page.hasMore && collected.length >= page.total) break;
+      }
+
+      final pubkeys = _sanitizePubkeys(collected, source: 'REST API');
 
       if (pubkeys.isNotEmpty) {
         _followingPubkeys = pubkeys;
@@ -1747,8 +1771,7 @@ class FollowRepository {
       if (currentUserPubkey.isEmpty) return;
 
       Log.info(
-        'Querying relay for kind 3 contact list '
-        '(REST API had no data)',
+        'Querying relay for authoritative kind 3 contact list',
         name: 'FollowRepository',
         category: LogCategory.system,
       );
--- a/mobile/packages/follow_repository/lib/src/follow_repository.dart
+++ b/mobile/packages/follow_repository/lib/src/follow_repository.dart
@@ -133,9 +133,11 @@ class FollowRepository {
   /// written by any Nostr client, and both the REST API and our caches are
   /// derived from those events. A single non-hex entry makes
   /// [_broadcastContactList] throw in `Contact`'s constructor, which aborts
-  /// the publish and permanently blocks every follow and unfollow — see the
-  /// `["p","nos"]` entry produced by a parser that read `tag[1]` of a
-  /// `["client","nos",...]` tag without filtering on `p`.
+  /// the publish and permanently blocks every follow and unfollow — a real
+  /// contact list on `wss://relay.divine.video` carries a `["p","nos"]` entry.
+  /// Its origin was never identified: `"nos"` matches `tag[1]` of the
+  /// `["client","nos",...]` tag the Nos app writes, but no parser reading that
+  /// value without filtering on `p` exists in this repo or its history.
   List<String> _sanitizePubkeys(
     List<String> pubkeys, {
     required String source,
@@ -287,7 +289,16 @@ class FollowRepository {
       }
     }
 
-    return FollowingSnapshot(pubkeys: following, count: following.length);
+    // Backs watchMyFollowingCached() for the current user too, so an invalid
+    // entry would surface as a ghost row on the own-Following screen and be
+    // persisted to Drift by CacheSync. Count follows the sanitized list so the
+    // header cannot disagree with the rows.
+    final sanitized = _sanitizePubkeys(
+      following,
+      source: 'Kind 3 event from $pubkey',
+    );
+
+    return FollowingSnapshot(pubkeys: sanitized, count: sanitized.length);
   }
 
   // ---- CacheSync-backed stale-while-revalidate watchers --------------------
@@ -1263,9 +1274,14 @@ class FollowRepository {
     }
   }
 
-  /// Follow a user
+  /// Follow a user.
   ///
-  /// Throws an [ArgumentError] if [pubkey] is not a 32-byte hex pubkey.
+  /// Throws:
+  ///
+  /// * [ArgumentError] if [pubkey] is not a 32-byte hex pubkey.
+  /// * [Exception] if the user is not authenticated.
+  /// * Anything the Kind 3 publish or the local-storage write threw, after
+  ///   rolling the optimistic update back.
   Future<void> follow(String pubkey) async {
     if (!_nostrClient.hasKeys) {
       Log.error(
--- a/mobile/packages/follow_repository/lib/src/follow_repository.dart
+++ b/mobile/packages/follow_repository/lib/src/follow_repository.dart
@@ -1965,14 +1965,15 @@ class FollowRepository {
 
   /// Broadcast updated contact list to network (Kind 3 event)
   Future<void> _broadcastContactList() async {
-    // Last line of defence: an invalid pubkey that reached the in-memory list
-    // would make Contact's constructor throw and abort the publish, blocking
-    // every follow and unfollow until the entry is gone. Dropping it here also
-    // self-heals the stored Kind 3 event, because this rebuilds the full list.
+    // Last line of defence: [executeFollowAction] appends without validating,
+    // and an invalid pubkey here makes Contact's constructor throw and abort
+    // the publish, blocking every follow and unfollow until the entry is gone.
+    // Emit so observers do not keep a dropped entry the getter no longer has.
     _followingPubkeys = _sanitizePubkeys(
       _followingPubkeys,
       source: 'in-memory following list',
     );
+    _emitFollowingList();
 
     // Create ContactList with all followed pubkeys
     final contactList = ContactList();
--- a/mobile/packages/follow_repository/lib/src/follow_repository.dart
+++ b/mobile/packages/follow_repository/lib/src/follow_repository.dart
@@ -127,6 +127,35 @@ class FollowRepository {
   bool get isInitialized => _isInitialized;
   int get followingCount => _followingPubkeys.length;
 
+  /// Drops entries that are not 32-byte hex pubkeys.
+  ///
+  /// Every source of the following list is untrusted: Kind 3 events are
+  /// written by any Nostr client, and both the REST API and our caches are
+  /// derived from those events. A single non-hex entry makes
+  /// [_broadcastContactList] throw in `Contact`'s constructor, which aborts
+  /// the publish and permanently blocks every follow and unfollow — see the
+  /// `["p","nos"]` entry produced by a parser that read `tag[1]` of a
+  /// `["client","nos",...]` tag without filtering on `p`.
+  List<String> _sanitizePubkeys(
+    List<String> pubkeys, {
+    required String source,
+  }) {
+    final valid = pubkeys.where(keyIsValid).toList();
+
+    final droppedCount = pubkeys.length - valid.length;
+    if (droppedCount > 0) {
+      final dropped = pubkeys.where((pubkey) => !keyIsValid(pubkey));
+      Log.warning(
+        'Dropped $droppedCount invalid pubkey(s) from $source: '
+        '${dropped.join(', ')}',
+        name: 'FollowRepository',
+        category: LogCategory.system,
+      );
+    }
+
+    return valid;
+  }
+
   /// Emit current state to stream (only if the list actually changed)
   void _emitFollowingList() {
     if (!_followingSubject.isClosed) {
@@ -1235,6 +1264,8 @@ class FollowRepository {
   }
 
   /// Follow a user
+  ///
+  /// Throws an [ArgumentError] if [pubkey] is not a 32-byte hex pubkey.
   Future<void> follow(String pubkey) async {
     if (!_nostrClient.hasKeys) {
       Log.error(
@@ -1243,6 +1274,17 @@ class FollowRepository {
         category: LogCategory.system,
       );
       throw Exception('User not authenticated');
+    }
+
+    // Guard: keep invalid keys out of the list rather than letting them fail
+    // the Kind 3 publish later, which would block every subsequent follow.
+    if (!keyIsValid(pubkey)) {
+      Log.error(
+        'Cannot follow - invalid pubkey: $pubkey',
+        name: 'FollowRepository',
+        category: LogCategory.system,
+      );
+      throw ArgumentError.value(pubkey, 'pubkey', 'Invalid key');
     }
 
     // Guard: Prevent following self
@@ -1470,8 +1512,10 @@ class FollowRepository {
   /// any follows that were added on other devices while offline.
   Future<void> mergeFollows(List<String> additionalPubkeys) async {
     // Remove self if accidentally included
-    final merged = <String>{..._followingPubkeys, ...additionalPubkeys}
-      ..remove(_nostrClient.publicKey);
+    final merged = <String>{
+      ..._followingPubkeys,
+      ..._sanitizePubkeys(additionalPubkeys, source: 'merged follows'),
+    }..remove(_nostrClient.publicKey);
 
     if (merged.length != _followingPubkeys.length ||
         !merged.every(_followingPubkeys.contains)) {
@@ -1504,7 +1548,10 @@ class FollowRepository {
 
         if (cached != null) {
           final decoded = jsonDecode(cached) as List<dynamic>;
-          _followingPubkeys = decoded.cast<String>();
+          _followingPubkeys = _sanitizePubkeys(
+            decoded.cast<String>(),
+            source: 'LocalStorage',
+          );
           _emitFollowingList();
 
           Log.info(
@@ -1541,11 +1588,14 @@ class FollowRepository {
           (tag) => tag.isNotEmpty && tag[0] == 'p',
         );
 
-        final pubkeys = pTags
-            .map((tag) => tag.length > 1 ? tag[1] : '')
-            .where((pubkey) => pubkey.isNotEmpty)
-            .cast<String>()
-            .toList();
+        final pubkeys = _sanitizePubkeys(
+          pTags
+              .map((tag) => tag.length > 1 ? tag[1] : '')
+              .where((pubkey) => pubkey.isNotEmpty)
+              .cast<String>()
+              .toList(),
+          source: 'PersonalEventCache',
+        );
 
         if (pubkeys.isNotEmpty) {
           // Guard: only accept the cached event when it is at least as
@@ -1610,7 +1660,7 @@ class FollowRepository {
         pubkey: currentUserPubkey,
         limit: 5000,
       );
-      final pubkeys = result.pubkeys;
+      final pubkeys = _sanitizePubkeys(result.pubkeys, source: 'REST API');
 
       if (pubkeys.isNotEmpty) {
         _followingPubkeys = pubkeys;
@@ -1915,6 +1965,15 @@ class FollowRepository {
 
   /// Broadcast updated contact list to network (Kind 3 event)
   Future<void> _broadcastContactList() async {
+    // Last line of defence: an invalid pubkey that reached the in-memory list
+    // would make Contact's constructor throw and abort the publish, blocking
+    // every follow and unfollow until the entry is gone. Dropping it here also
+    // self-heals the stored Kind 3 event, because this rebuilds the full list.
+    _followingPubkeys = _sanitizePubkeys(
+      _followingPubkeys,
+      source: 'in-memory following list',
+    );
+
     // Create ContactList with all followed pubkeys
     final contactList = ContactList();
     for (final pubkey in _followingPubkeys) {
@@ -1994,12 +2053,16 @@ class FollowRepository {
       _currentUserContactListEvent = event;
 
       // Extract followed pubkeys from 'p' tags
-      final followedPubkeys = <String>[];
+      final taggedPubkeys = <String>[];
       for (final tag in event.tags) {
         if (tag.isNotEmpty && tag[0] == 'p' && tag.length > 1) {
-          followedPubkeys.add(tag[1]);
+          taggedPubkeys.add(tag[1]);
         }
       }
+      final followedPubkeys = _sanitizePubkeys(
+        taggedPubkeys,
+        source: 'network Kind 3 event ${event.id}',
+      );
 
       // Guard: detect catastrophic list reduction from buggy external clients.
       // A common Nostr footgun is publishing a Kind 3 event without first
--- a/mobile/packages/follow_repository/test/src/follow_repository_test.dart
+++ b/mobile/packages/follow_repository/test/src/follow_repository_test.dart
@@ -406,6 +406,73 @@ void main() {
       });
     });
 
+    // Regression: a real user's Kind 3 event carried a `["p","nos"]` entry,
+    // produced by a parser that read `tag[1]` of a `["client","nos",...]` tag
+    // without filtering on `p`. Every follow and unfollow then failed with
+    // `Invalid key: "nos"` from Contact's constructor, because the broadcast
+    // rebuilds the whole list and aborts before publishing.
+    group('invalid pubkey handling', () {
+      const invalidPubkey = 'nos';
+
+      test('drops invalid entries when loading from local storage', () async {
+        SharedPreferences.setMockInitialValues({
+          'following_list_$testCurrentUserPubkey':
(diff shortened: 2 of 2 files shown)


mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -127,6 +127,35 @@ class FollowRepository {
...
+  /// derived from those events. A single non-hex entry makes
+  /// [_broadcastContactList] throw in `Contact`'s constructor, which aborts
+  /// the publish and permanently blocks every follow and unfollow — see the
+  /// `["p","nos"]` entry produced by a parser that read `tag[1]` of a
+  /// `["client","nos",...]` tag without filtering on `p`.
+  List<String> _sanitizePubkeys(
+    List<String> pubkeys, {
+    required String source,
+  }) {
+    final valid = pubkeys.where(keyIsValid).toList();
```
realmeylisdev, line 145: The six sites miss `getOthersFollowing` (:273-291), which reads the same untrusted `p` tags with no `keyIsValid` filter. It backs `watchMyFollowingCached()` → `MyFollowingBloc`, so the affected account's own Following screen still lists `nos` and CacheSync persists it to Drift.

Not the publish abort — that's fixed — but if the thesis is "every source of the following list is untrusted", this is a seventh one. (`getFollowingFromApi` at :499 is the same shape, though it currently has no callers.)

mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -127,6 +127,35 @@ class FollowRepository {
...
   int get followingCount => _followingPubkeys.length;
 
+  /// Drops entries that are not 32-byte hex pubkeys.
+  ///
+  /// Every source of the following list is untrusted: Kind 3 events are
+  /// written by any Nostr client, and both the REST API and our caches are
+  /// derived from those events. A single non-hex entry makes
+  /// [_broadcastContactList] throw in `Contact`'s constructor, which aborts
+  /// the publish and permanently blocks every follow and unfollow — see the
+  /// `["p","nos"]` entry produced by a parser that read `tag[1]` of a
```
realmeylisdev: The "server-side in funnelcake" half of the Out of Scope section doesn't hold. Its kind-3 extraction has filtered on the tag name since the original migration — `arrayFilter(t -> t[1] = 'p', tags)` in `000009`, still there at `000096_contact_lists_latest.up.sql:22` — and it holds no signing keys, so it can't mint a signed `p` tag into someone's contact list. It's an amplifier, not a writer: `following Array(String)` (000096:11) does no hex validation and `get_following` returns `nos` verbatim.

I swept every current kind-3 parser in divine-mobile, nostr_sdk, divine-web, divine-space, divine-badges and divine-connect — all filter `tag[0] == 'p'`. No live writer anywhere, which is the part that matters for this fix. But I couldn't find the `["client","nos",...]` parser in any repo or anywhere in git history either. Worth softening this dartdoc to what's actually known, since it'll outlive the investigation.

mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -1235,6 +1264,8 @@ class FollowRepository {
   }
 
   /// Follow a user
+  ///
+  /// Throws an [ArgumentError] if [pubkey] is not a 32-byte hex pubkey.
```
realmeylisdev: `follow()` has three escape paths — `Exception('User not authenticated')` at :1276, this `ArgumentError`, and the bare `rethrow` at :1361. The doc lists one.

mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -1245,6 +1276,17 @@ class FollowRepository {
       throw Exception('User not authenticated');
     }
 
+    // Guard: keep invalid keys out of the list rather than letting them fail
+    // the Kind 3 publish later, which would block every subsequent follow.
+    if (!keyIsValid(pubkey)) {
```
realmeylisdev, line 1300: Minor correction to "unchanged from `main`": this guard sits above both the offline branch (:1323) and the `contains()` early-return (:1300). On `main`, `follow('nos')` while offline queued and returned normally, and for the poisoned account — whose list literally contains `nos` — it returned silently at `contains()`. Both now throw.

Only reachable via `NotificationFeedBloc` with a bad actor pubkey, so the behavior change itself is fine. The claim just isn't.

mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -1470,8 +1512,10 @@ class FollowRepository {
   /// any follows that were added on other devices while offline.
   Future<void> mergeFollows(List<String> additionalPubkeys) async {
     // Remove self if accidentally included
-    final merged = <String>{..._followingPubkeys, ...additionalPubkeys}
-      ..remove(_nostrClient.publicKey);
+    final merged = <String>{
+      ..._followingPubkeys,
+      ..._sanitizePubkeys(additionalPubkeys, source: 'merged follows'),
```
realmeylisdev, line 1536: Reverting this to plain `additionalPubkeys` leaves all 209 tests green — same for the `PersonalEventCache` site at :1591. The VGV gate is satisfied by incidental execution from existing valid-pubkey tests, so the line is covered but the behavior isn't asserted.

mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -1610,7 +1660,7 @@ class FollowRepository {
         pubkey: currentUserPubkey,
         limit: 5000,
       );
-      final pubkeys = result.pubkeys;
+      final pubkeys = _sanitizePubkeys(result.pubkeys, source: 'REST API');
```
realmeylisdev: Same here — reverting to `result.pubkeys` keeps the suite green, and this is the one source I could independently confirm carries the bad value end-to-end (funnelcake stores it unvalidated and serves it back). The `getFollowing` mock at test:198 makes this cheap to pin.

Unrelated, but on the line you're touching: `limit: 5000` doesn't do what it looks like. The v1 handler clamps to 100 (`handlers.rs:5521`, `.min(100)`) and v2's `clamp_standard_limit` caps at `MAX_LIST_LIMIT = 100`; the reported account has 249 follows. Pre-existing, and the kind-3 subscription corrects it shortly after — just flagging it while you're here.

mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -1915,6 +1965,16 @@ class FollowRepository {
 
   /// Broadcast updated contact list to network (Kind 3 event)
   Future<void> _broadcastContactList() async {
+    // Last line of defence: [executeFollowAction] appends without validating,
+    // and an invalid pubkey here makes Contact's constructor throw and abort
+    // the publish, blocking every follow and unfollow until the entry is gone.
+    // Emit so observers do not keep a dropped entry the getter no longer has.
+    _followingPubkeys = _sanitizePubkeys(
```
realmeylisdev, line 2011: "Affected accounts self-heal" holds for a follow performed in divine-mobile — this rebuild does drop the entry. The other divine kind-3 writers clone the tag array and append instead of rebuilding from a pubkey list, so they preserve `["p","nos"]` verbatim: divine-web `useFollowRelationship.ts:187,198`, divine-space `useDivineSocial.ts:143,156`, divine-badges `badges.js:324,338`.

A user who next follows from web stays poisoned on the relay (and in funnelcake's counts). Mobile stays correct either way since it re-sanitizes on read — it's the "no manual data repair" half that's narrower than stated.

mobile/packages/follow_repository/test/src/follow_repository_test.dart
```diff
@@ -406,6 +406,107 @@ void main() {
...
+              '["$invalidPubkey", "$testTargetPubkey"]',
+        });
+
+        await repository.initialize();
+
+        expect(repository.followingPubkeys, [testTargetPubkey]);
+        expect(repository.isFollowing(invalidPubkey), isFalse);
+      });
+
+      test('follow succeeds and publishes a clean list when the cached '
```
realmeylisdev, line 569: "Each new test was verified by mutation" doesn't quite hold for this one. Reverting the LocalStorage hunk alone leaves `['nos', target, target2]` at broadcast time and the publish-site sanitize drops it; reverting the publish-site hunk alone means LocalStorage already yielded `[target]`. It only goes red with both reverted.

That's defence-in-depth working as designed rather than a problem — and I confirmed the test does fail against `main` with the exact reported error. The blanket claim is just stronger than what it does. The other four mutate cleanly (`+208 -1` each).

realmeylisdev (approved):
Approving. Reproduced the `Invalid key: "nos"` publish abort against `main` using your own test, and confirmed funnelcake serves the bad entry back unvalidated — so the REST sanitize site is load-bearing, and sanitize-at-ingest in the repository is the right call over touching `Contact`.

Comments are non-blocking: three of the six sanitize sites have no test pinning them, and two claims in the description don't survive checking.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Good catch — fixed in 6b9d1e4. The `watchMyFollowingCached()` → `getOthersFollowing(ownPubkey)` indirection is genuinely easy to miss behind the "Others" name. Count now derives from the sanitized list too, otherwise the header drifts from the rows.

Not acting on the `getFollowingFromApi` parenthetical: it reads no `p` tags and has no production callers, so sanitizing it would be dead code — and `getFollowersFromApi` at :479 is unsanitized in exactly the same way. If REST-sourced pubkeys are untrusted, that belongs in a separate change covering both at the client boundary.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Dartdoc softened in 6b9d1e4 — you are right that it stated an unproven inference as fact. Stronger than you had it, in fact: the literal `nos` appears in exactly one commit in the entire history, this PR's own fix. It now records the origin as unidentified.

Two corrections. The dartdoc has no Out of Scope section and never mentions funnelcake — that text exists only in the PR body, so the funnelcake argument and the dartdoc edit are aimed at different artifacts. And the migration/SQL citations are unverifiable here (no funnelcake checkout); the point that carries is simpler — it holds no signing keys, so it cannot mint a signed `p` tag regardless of what its extraction does.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Fixed in 6b9d1e4 — all three documented.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Correct, and the body is fixed in 6b9d1e4. The guard sits above both the `contains()` early-return and the offline branch, so both changed; only the online publish path is genuinely unchanged.

One scope correction: `follow()` has five production call paths, not one. `NotificationFeedBloc` is the only direct caller, but `toggleFollow` reaches it from four more — `my_following_bloc:149`, `other_profile_bloc:197`, `other_profile_screen:341`, `conversation_view:80`.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Confirmed — both mutants survive. Added a test for the PersonalEventCache site in 6b9d1e4: it caches the very Kind 3 events this bug came from, so a poisoned event replays through it on every launch and the gap is real.

Left `mergeFollows` unpinned deliberately — its input is another device's list, which arrives already sanitized on the way in, so a test there would be pinning a scenario I cannot construct honestly.

Mechanism nit: no VGV gate is involved. `follow_repository` is not a strict-coverage package — per `.claude/rules/testing.md` only `divine_ui` is. The observation stands; the cause does not.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Part A pinned in 6b9d1e4, using the existing `getFollowing` stub as you suggested.

Part B deserves more than "just flagging it" — it is worse than the aside implies, and I have filed it as #6109 with a repro. The clamp means a truncated REST result suppresses the authoritative relay Kind 3 query entirely (`initialize()` gates step 4 on `if (_followingPubkeys.isEmpty)`, and 100 entries is not empty). If a follow then lands before the subscription repairs the list, `_broadcastContactList` publishes the truncation and `_currentUserContactListEvent` advances past the real event, so the authoritative 249-entry event is silently discarded on arrival. Follows 101+ are gone for every client.

So "corrects it shortly after" holds only when nothing races the subscription. Verified live that the clamp is real: `?limit=5000` echoes back `"limit":100`.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Body scoped to mobile in 6b9d1e4.

Two pushbacks. Your conclusion is inverted: "no manual data repair" is the half that holds — no human ever hand-edits anything, and the next mobile follow re-cleans it. The narrow half is "affected accounts self-heal", which is a scope claim. And once mobile heals, a clone-append client clones the *healed* array; re-poisoning needs it to be working from a stale pre-heal copy, which makes this a race window rather than a stable oscillation.

The three citations I cannot verify — divine-web, divine-space and divine-badges are not checked out here. The mechanism is sound in the abstract and NIP-02 does recommend clone-append, so I have scoped the claim anyway.

mobile/packages/follow_repository/test/src/follow_repository_test.dart
hm21: Confirmed — all four predictions landed, including the mechanism, not just the pass/fail. Body reworded in 6b9d1e4: four tests go red on their single hunk; this one is an end-to-end guard covered by two hunks in defence-in-depth and needs both reverted, by design.

hm21:
Thanks @realmeylisdev — all eight verified against the code, all eight addressed in 6b9d1e4. Replies on each thread.

**Fixed:** `getOthersFollowing` sanitized (the `watchMyFollowingCached()` indirection was a real find), REST + PersonalEventCache sanitizes pinned with tests, `follow()` throws documented, `_sanitizePubkeys` provenance softened to unidentified.

**Body corrected:** the "unchanged from `main`" claim, self-heal scoped to mobile, and the mutation claim.

**Filed separately:** #6109 — your `limit: 5000` aside is the most valuable thing in this review. The clamp lets a truncated REST result suppress the authoritative relay Kind 3 query, and a racing follow then destroys follows 101+ for every client. Pre-existing on `main`, needs a design decision.

Pushed back on four sub-claims (details on the threads): `getFollowingFromApi` is not the same shape and has no callers; `follow()` has five call paths, not one; no VGV gate applies to this package; and on the self-heal thread the conclusion is inverted — "no manual data repair" is the half that holds.

211 tests, 100.00% coverage, analyze clean.

mobile/packages/follow_repository/lib/src/follow_repository.dart
realmeylisdev: Verified — one return path at :301, nothing bypasses the sanitize, and reverting either the pubkeys list or the count derivation turns the new test red.

Agreed on the deferral: both `getFollowingFromApi` and `getFollowersFromApi` have zero non-test callers, so there is no live gap and the client boundary is the right home for it.

Non-blocking residual: the sanitize sits in the `fetch` closure, so a legacy poisoned Drift entry still emits unsanitized through `FollowingSnapshot.fromJson` on the cached path (`cache_sync.dart:214`) before the fetch repairs it.

mobile/packages/follow_repository/lib/src/follow_repository.dart
realmeylisdev: Fair on the artifact split — my comment opened on the Out of Scope section but asked you to soften the dartdoc, and those are different targets. The dartdoc edit reads well.

The body is still the issue: the Description asserts the provenance as fact ("It originates from a parser that read `tag[1]` of a `["client","nos","https://nos.social"]` tag ... our app then republished that value"), which now contradicts the dartdoc you just softened, and Out of Scope still offers the funnelcake disjunct.

On verifiability — I have divine-funnelcake checked out, and the citations are line-exact: `000096_contact_lists_latest.up.sql:22` is verbatim `arrayMap(t -> t[2], arrayFilter(t -> t[1] = 'p', tags)) AS following`, `:11` is `following Array(String)`, and `000009:13` carries the same filter. Your no-signing-keys argument reaches the same conclusion, so the disjunct is disproven either way.

mobile/packages/follow_repository/lib/src/follow_repository.dart
realmeylisdev: All three are listed now, but bullet 3 picked up two inaccuracies.

"the local-storage write" cannot throw — `_saveToLocalStorage()` wraps its whole body in a bare `catch (e)` at :1729 that only logs. And "after rolling the optimistic update back" holds only for `_broadcastContactList()`: the try opens at :1355, but the optimistic update is applied at :1335-1336, and `_invalidateMyFollowingCache` (:1337) and `_queueOfflineAction` (:1341) both sit above it — anything they throw escapes with the update still applied.

mobile/packages/follow_repository/lib/src/follow_repository.dart
realmeylisdev: My scope claim was wrong — though the five call paths are not why. Three of your four are `isFollowing`-gated and route to `unfollow()`, which has no `keyIsValid` guard, so they cannot reach :1297: `other_profile_bloc:195` wraps the call, and `other_profile_screen:341` / `conversation_view:80` are reachable only via `MoreSheetResult.unfollow`, which `more_sheet_menu:71` renders only when `isFollowing`. There is also a fifth you did not list — `comment_reactions_bloc:395` — likewise gated.

The one that does reach it is `my_following_bloc:149`: before this PR the unsanitized `getOthersFollowing` could render a ghost `nos` row on the Following screen, and tapping it dispatched `follow('nos')`. Your round-2 sanitize closes exactly that path.

mobile/packages/follow_repository/lib/src/follow_repository.dart
realmeylisdev: PersonalEventCache and the REST site are pinned — both mutants die now.

On the gate: `.github/workflows/follow_repository.yaml` calls VeryGood's `flutter_package.yml@v1` with no `min_coverage` override, so it inherits the 100% default — packages that want less set it explicitly (`dm_repository: 93`, `likes_repository: 62`, `collaborator_repository: 96`). Your PR body has it right ("package gate is the VGV 100% default"); the `divine_ui` line in `testing.md` is an illustrative list, not the mechanism. Measured 667/667 on the file at HEAD.

On `mergeFollows` — the premise does not hold: it has zero callers and never has had one, so there is no way in to arrive pre-sanitized. But the instinct was right. The obvious test survives the mutant, because `_broadcastContactList` re-sanitizes and reassigns `_followingPubkeys` (:1988-1992), which masks any post-merge state assertion. An honest one does exist via the change-detection guard at :1536: merging only an invalid entry must be a no-op, and unsanitized it registers as a change and fires a spurious broadcast, so `verifyNever(sendContactList)` kills it. Your call whether that earns its place — the stale "Used when syncing offline actions" doc comment is the part I would fix regardless, since PendingActionService wires only `executeFollowAction` / `executeUnfollowAction`.

mobile/packages/follow_repository/lib/src/follow_repository.dart
realmeylisdev: You are right, and my aside understated it. Traced the chain: `initialize()` gates step 4 on `_followingPubkeys.isEmpty` (:1249), `_broadcastContactList` advances `_currentUserContactListEvent` to the event it just published (:2013), and `_processContactListEvent` gates its whole body on `event.createdAt > _currentUserContactListEvent!.createdAt` (:2068) — so the authoritative event fails the strict `>` and is dropped. "Corrects it shortly after" holds only in your case 3.

#6109 is the right home for it, and the 40-to-32 republish you found on wss://relay.divine.video makes it concrete.

mobile/packages/follow_repository/lib/src/follow_repository.dart
realmeylisdev: You are right, I had it backwards. "On the next follow" binds healing to the rebuild only mobile does, so the clone-append evidence narrows self-heal, not manual-repair — and no scenario ever requires a human to touch the data. The scoped body wording reads accurately.

mobile/packages/follow_repository/test/src/follow_repository_test.dart
realmeylisdev: Reworded reads accurately — that is the distinction I meant.

realmeylisdev:
Re-verified 6b9d1e4 — `getOthersFollowing` is correctly placed and mutation-proof, 211 pass, 667/667 on the file. Approval stands; nothing blocking.

You were right on the self-heal inversion and on #6109 (traced all five links — the strict `>` at :2068 really does discard the authoritative event), and my "only reachable via NotificationFeedBloc" was wrong too. Resolved those four.

Reopened three with residuals: the PR body still asserts the `["p","nos"]` provenance as fact and now contradicts the dartdoc you softened (the one I'd fix before merge); `follow()` bullet 3 has two new inaccuracies; and the VGV gate is real — `follow_repository.yaml` sets no `min_coverage` override, so it inherits the 100% default, exactly as your own body says.

Issue #5177, opened 2026-06-15
epic(trust-safety): organize moderation, content controls, and reporting


NotThatKindOfDrLiz:
## Scope
Current implementation epic for trust and safety, moderation operations, content warnings, blocking, reporting, and content-control behavior.

## Includes
- content warning visibility and tagging flows
- blocked-user/content filtering behavior
- moderation operator/account setup decisions
- report reason/product-copy policy decisions

## Initial triage
Created from the parentless open-issue audit on 2026-06-15. This is intended to replace relying on old design-only grouping for active implementation work.

Issue #4948, opened 2026-06-07
docs(trust-safety): decide moderation-pubkey trust model: pin vs NIP-05 rotation (C10, #570)


realmeylisdev:
Security decision deferred from #4944 for @rabble / T&S sign-off (divinevideo/divine-mobile#570, finding C10).

`ModerationLabelService` resolves the moderation pubkey **NIP-05 → 24h cache → hardcoded fallback** (`moderation_label_service.dart:101-172`), so **NIP-05 currently wins**. The NIP-05 fetch (`Nip05Validor.getPubkey`) is a plain unpinned HTTPS GET of `/.well-known/nostr.json`. A MITM / DNS / endpoint compromise could **redirect the moderation identity** (where report DMs, carrying reporter identity + reported content, are sent) for up to the 24h cache TTL.

**The tradeoff:**
- **Pin-authoritative** (hardcoded `8fd5eb6d…433b7e` wins; NIP-05 demoted to an advisory mismatch warning): redirect-resistant, but **disables NIP-05 key rotation** — T&S has rotated once already (legacy `121b915b…` → current), and pinning would require an app release to rotate.
- **Keep NIP-05 + log mismatches**: preserves rotation agility, but a redirect still succeeds (only observable).

Live check: NIP-05 currently resolves to exactly the pinned key, so there's no functional difference today — this is purely about the trust posture.

Per the project lesson on not making security-boundary changes by inference (PR #4665 reverted), this is **held for an explicit owner decision** rather than flipped unilaterally.

Context: divinevideo/divine-mobile#570.

mbradley:
quick context: i'm the one who pushed for resolving the moderation identity via NIP-05 originally (divinevideo/divine-web#196, divinevideo/divine-mobile#2321) and it shipped. i still favor it, but i don't want to defend it blindly, so here's an honest restatement with what's been flagged folded in.

the reason for NIP-05 and its main weakness are the same fact: because we control the name server, we control what `moderation@divine.video` resolves to. that's a recovery lever whoever holds the key doesn't have. if the key is compromised or lost, we re-point the identity and clients move to the new one within the cache window, no app release needed. the flip side, which this issue rightly raises, is that trust rides on an HTTPS GET of our own `nostr.json` (unpinned, per Meylis's note in here, i haven't re-read the client myself), cached 24h. so a DNS/TLS/endpoint compromise of our own domain could redirect the moderation identity for up to that window, and that identity is where report DMs go and whose labels clients trust.

[...]

PR #6106, opened 2026-07-15
fix(trust-safety): log moderation-pubkey pin divergence + ignore NIP-05 redirects (#4948)

realmeylisdev:
## Description

Tier-1 controls from the #4948 decision, on the moderation NIP-05 path. The decision (2026-07-07) split the trust model into two tiers. Tier 2 (the protected-minor DM gate) shipped in #5754. Tier 1 shipped nothing — this covers the client-side pieces of it, plus a web gap that review surfaced in both NIP-05 lookup paths.

**1. Advisory mismatch logging** (`feat` commit)

Tier 1 keeps NIP-05 authoritative for the labeler identity, accepting that a name-server or DNS redirect can point clients at an attacker's key for up to the 24h cache TTL. The decision paired that acceptance with advisory mismatch logging so the redirect is at least *observable*. It never landed.

Until now a hostile repoint of `moderation@divine.video` was indistinguishable from an intended rotation. The only signals were an unconditional info line naming the resolved key, and `Updated moderation labeler from A to B`, which compares against the previously adopted key — never against the pin. A first-ever resolve that already diverged produced no signal at all.

Every adoption now funnels through `_adoptModerationPubkey`, so the check can't be bypassed by adding a resolution source later, and a divergence from the key shipped in this build logs a warning. **Advisory only — NIP-05 still wins.** No behavior change, per the decision. Both keys are logged in full (never truncate Nostr IDs).

**2. NIP-05 redirect + timeout hardening** (`fix` commit)

NIP-05's Security Constraints are explicit: the `.well-known/nostr.json` endpoint MUST NOT return HTTP redirects, and **fetchers MUST ignore any it returns**. `Nip05Validor` built a bare `Dio()` and issued a plain `get()`, inheriting Dio's defaults of `followRedirects: true`, `maxRedirects: 5`, and no timeout at all.

Following a 30x is a spurious-verify vector: a MITM or misconfigured origin can bounce the lookup to a host that answers with an attacker's key, which the moderation path then caches for 24h and subscribes to as the labeler.

Also corrects the field comment that described the resolution order as `NIP-05 → cache → fallback`. It is **cache-first**, as `_resolveModerationPubkey`'s own doc says — and that stale comment is what put the inverted order into the issue's decision text.

**3. The same redirect refusal, on Flutter web** (`fix` commits — from @dcadenas's review)

`followRedirects: false` does not do what (2) claimed on web, and neither did the equivalent guard already shipped in the #176 resolver. Both are now fixed.

`followRedirects` / `maxRedirects` are read in exactly one place in the whole dio stack — `io_adapter.dart:138-139`. `dio_web_adapter` never reads them: on web the request goes through `XMLHttpRequest`, which follows a 30x in the browser with no opt-out. The lookup lands on the redirect target's **200**, which the default `validateStatus` accepts. This repo ships web to production at `app.divine.video` on every merge to main, and both `Nip05Validor` callers (moderation labeler resolution, user badge verification) are reachable there with no `kIsWeb` gating.

`Response.isRedirect` is the only redirect signal that survives every transport — the web adapter sets it from `options.uri != xhr.responseURL` — so both lookups now guard on it before reading the body. No-op on `dart:io`, where the unfollowed 3xx has already thrown.

`Nip05Resolver` (#176 protected-minor DM gate, shipped in #5754) carried the identical inert guard: its `status >= 300 && status < 400` check never fires on web because the status is 200. That one is the spurious-**APPROVE** its own comment warns about — a burner claiming an official identifier can bounce the lookup to a host that echoes the expected key and resolve to `matched`. Pre-existing, not introduced here, but folded in rather than deferred: same control, same one-line remedy, and this PR previously cited that resolver as the already-correct reference for redirect refusal — a claim that did not hold on web. The comments in both files that asserted the `validateStatus` story have been corrected to say which platform they actually describe.

**4. Canonical labeler identity** (`refactor` commit — from @mbradley's review note)

`_adoptModerationPubkey` stored the raw pubkey while `_warnIfPinMismatch` compared the normalized form, so an uppercase answer would be warn-silent yet fail to match lowercase event authors in the subscription filter. Both producers now return a canonical key, so the identity, the subscription set and the pin comparison agree. No live bug — NIP-05 mandates lowercase hex.

**Related Issue:** Relates to #4948, #5177

## Reviewer notes

- **Blast radius is deliberate.** User NIP-05 verification runs through the same `getPubkey`, so a third-party host that answers the well-known with a redirect now fails to verify — on native *and* web. That's the spec-mandated outcome for a non-compliant endpoint, not a regression.
- **Known narrow tradeoff on web.** The web adapter derives `isRedirect` by string-comparing `options.uri` against `xhr.responseURL`, so an IDN NIP-05 domain (`bob@münchen.de`) reads as a phantom redirect: Dart serializes the host percent-encoded, the browser punycodes it. Those identifiers already fail on native — `dart:io` hands the percent-encoded host straight to DNS — so the guard brings web to parity rather than regressing a working case, and it fails closed (unverified badge). Flagging it so nobody debugs an intermittent web "unverified" from scratch. The moderation caller is a hardcoded ASCII constant and is unaffected.
- **Expected benign warning.** A user whose cache still holds the legacy pubkey (pre-rotation) gets one warning at load; the NIP-05 refresh then adopts the current key and it stops. That's informative — it means they were subscribed to a dead labeler.
- Live state is currently clean: `moderation@divine.video` resolves to exactly `8fd5eb6d8f362163bc00a5ab6b4a3167dbf32d00ec4efdbcf43b3c9514433b7e`, the pinned constant. This is detection for the first repoint, not a response to a live divergence.

## Out of Scope

Tracked from the #4948 audit, not addressed here:

- **divine-web is still hardcoded** (`videoVerification.ts:8-9`, no NIP-05 resolver). Keys are byte-identical today so there's no live split, but web can't follow a rotation — meaning the clients diverge exactly when a repoint is used as the revocation lever. Highest-risk remaining Tier-1 item.
- **Widening this warning to web** and to telemetry rather than logs alone.
- **Team-side `nostr.json` monitoring** and a **repoint-to-burner revocation runbook** (ops, support-trust-safety). The runbook is load-bearing: the Tier-2 resolver documents that `networkError` is adversary-selectable, so a repoint is the only suppression-proof revocation lever.
- **TLS cert pinning** for divine.video (open item 2). Redirect refusal closes a cheaper slice of the same class.
- **Recording the Tier-1 decision durably** — it exists only in a GitHub comment.

## Verification

- `flutter analyze lib test` (mobile) — No issues found
- `flutter analyze lib test` (nostr_sdk) — No issues found
- `dart format --set-exit-if-changed` on all 6 files — 0 changed
- `flutter test test/services/moderation_label_service_test.dart` — 26 passed
- `flutter test test/services/nip05_resolver_test.dart` — 16 passed
- `flutter test` (nostr_sdk package) — 355 passed
- `flutter test test/services/ --test-randomize-ordering-seed random` — 3060 passed, confirming the static-`dio` adapter swap restores cleanly and doesn't strand the merged isolate
- `check_process_global_mutations.sh`, `check_shared_channel_overrides.sh`, `check_http_overrides_isolation.sh`, `check_untested_services_floor.sh` — all pass
- **Mutation-tested every new control** — each new test was confirmed to fail when its control is removed. For the web guards the failure is the vulnerability itself: without them `getPubkey` returns the redirect target's key, `valid()` verifies an impostor's badge, and `Nip05Resolver` returns `matched` instead of `networkError`.

## Type of Change

### DIFF
--- a/mobile/packages/nostr_sdk/lib/nip05/nip05_validor.dart
+++ b/mobile/packages/nostr_sdk/lib/nip05/nip05_validor.dart
@@ -6,7 +6,24 @@ import 'package:dio/dio.dart';
 class Nip05Validor {
   static final Map<String, int> _checking = {};
 
-  static var dio = Dio();
+  /// Bounds a slow or hostile origin. Without it a stalled read hangs the
+  /// caller for Dio's default (no timeout at all).
+  static const Duration _timeout = Duration(seconds: 5);
+
+  /// NIP-05 "Security Constraints": the `.well-known/nostr.json` endpoint MUST
+  /// NOT return HTTP redirects, and fetchers MUST ignore any it returns.
+  /// Following a 30x is a spurious-verify vector — a MITM or a misconfigured
+  /// origin can bounce the lookup to a host that answers with an attacker's
+  /// key. Dio's default `validateStatus` accepts 2xx only, so an unfollowed 3xx
+  /// throws and [getPubkey] yields null rather than a redirected key.
+  static var dio = Dio(
+    BaseOptions(
+      connectTimeout: _timeout,
+      receiveTimeout: _timeout,
+      followRedirects: false,
+      maxRedirects: 0,
+    ),
+  );
 
   static Future<bool?> valid(String nip05Address, String pubkey) async {
     if (_checking[nip05Address] != null) {
--- a/mobile/packages/nostr_sdk/lib/nip05/nip05_validor.dart
+++ b/mobile/packages/nostr_sdk/lib/nip05/nip05_validor.dart
@@ -14,8 +14,14 @@ class Nip05Validor {
   /// NOT return HTTP redirects, and fetchers MUST ignore any it returns.
   /// Following a 30x is a spurious-verify vector — a MITM or a misconfigured
   /// origin can bounce the lookup to a host that answers with an attacker's
-  /// key. Dio's default `validateStatus` accepts 2xx only, so an unfollowed 3xx
-  /// throws and [getPubkey] yields null rather than a redirected key.
+  /// key.
+  ///
+  /// These flags only bind on `dart:io` (dio's IOHttpClientAdapter forwards
+  /// them to [HttpClientRequest]). On Flutter web dio delegates to
+  /// XMLHttpRequest, which follows a 30x in the browser and never reads them —
+  /// the lookup lands on the redirect target's 200, which the default
+  /// `validateStatus` accepts. [getPubkey]'s `isRedirect` check is what
+  /// actually enforces the constraint there.
   static var dio = Dio(
     BaseOptions(
       connectTimeout: _timeout,
@@ -59,6 +65,13 @@ class Nip05Validor {
     var url = "https://$address/.well-known/nostr.json?name=$name";
     try {
       var response = await dio.get(url);
+      // The only redirect signal that survives every transport. On web a
+      // browser-followed 30x arrives as a 200 flagged `isRedirect`, so
+      // `followRedirects: false` never sees it; on `dart:io` the 3xx has
+      // already thrown, making this a no-op there.
+      if (response.isRedirect) {
+        return null;
+      }
       if (response.data != null) {
         var map = response.data;
         if (map is String) {
--- a/mobile/lib/services/moderation_label_service.dart
+++ b/mobile/lib/services/moderation_label_service.dart
@@ -582,16 +582,27 @@ class ModerationLabelService {
     });
   }
 
+  /// Canonical form of a labeler identity: lowercase hex, no surrounding
+  /// whitespace.
+  ///
+  /// NIP-05 mandates lowercase hex and event authors arrive lowercase on the
+  /// wire, so canonicalizing at the two points that produce an identity keeps
+  /// [_divineModerationPubkey], [_subscribedLabelers] and the pin comparison on
+  /// one form. Without it a non-lowercase answer would read as identical to the
+  /// pin yet fail to match the labeler's own events in the subscription filter.
+  static String _normalizedPubkey(String pubkey) => pubkey.trim().toLowerCase();
+
   /// Resolve the Divine moderation pubkey via cached value or NIP-05 lookup.
   ///
   /// Strategy: SharedPreferences cache (24h TTL) → NIP-05 → fallback constant.
+  /// Every path returns a [_normalizedPubkey].
   Future<String> _resolveModerationPubkey(SharedPreferences prefs) async {
     // Check cached resolution
-    final cachedPubkey = prefs.getString(_resolvedPubkeyKey);
+    final cachedPubkey = _normalizedPubkey(
+      prefs.getString(_resolvedPubkeyKey) ?? '',
+    );
     final cachedAtStr = prefs.getString(_resolvedAtKey);
-    if (cachedPubkey != null &&
-        cachedPubkey.isNotEmpty &&
-        cachedAtStr != null) {
+    if (cachedPubkey.isNotEmpty && cachedAtStr != null) {
       final cachedAt = DateTime.tryParse(cachedAtStr);
       if (cachedAt != null &&
           DateTime.now().difference(cachedAt) < _resolvedPubkeyTtl) {
@@ -602,15 +613,16 @@ class ModerationLabelService {
     // Resolve via NIP-05
     try {
       final resolved = await Nip05Validor.getPubkey(divineModerationNip05);
-      if (resolved != null && resolved.isNotEmpty) {
-        await prefs.setString(_resolvedPubkeyKey, resolved);
+      final normalized = _normalizedPubkey(resolved ?? '');
+      if (normalized.isNotEmpty) {
+        await prefs.setString(_resolvedPubkeyKey, normalized);
         await prefs.setString(_resolvedAtKey, DateTime.now().toIso8601String());
         Log.info(
-          'Resolved moderation pubkey via NIP-05: $resolved',
+          'Resolved moderation pubkey via NIP-05: $normalized',
           name: 'ModerationLabelService',
           category: LogCategory.system,
         );
-        return resolved;
+        return normalized;
       }
     } catch (e) {
       Log.warning(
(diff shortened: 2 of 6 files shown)


mbradley (approved):
Reviewed the two Tier-1 controls against the #4948 decision and the NIP-05 spec, and traced the load-bearing claims through the code. Solid work, approving.

What I verified holds:
- The adoption funnel is complete: the only two writes to `_divineModerationPubkey` are the pin initializer and the one inside `_adoptModerationPubkey`, so the pin check genuinely can't be bypassed by a future resolution source.
- Advisory-only is honored: the warn never changes the trusted key, and a test pins that.
- The pin comparison is correct: the fallback const is lowercase and the adopted key is case-normalized before comparing, so the pin and any case-only variant stay silent, matching NIP-05's lowercase-hex mandate.
- Fires once per adoption: `ensureLoaded` is idempotent and the refresh only adopts on an actual change, so a steady divergence is one line per session.
- The comment fix is accurate: resolution really is cache -> NIP-05 -> fallback.
- The dio hardening matches the protected-minor resolver in `nip05_resolver.dart` exactly (5s connect/receive, followRedirects: false, maxRedirects: 0), and the null-on-3xx mechanism (default validateStatus rejects the unfollowed 3xx) is confirmed by the passing 302 -> null test.

Concur on the disclosed blast radius: user NIP-05 verification now refusing redirects is the spec-mandated outcome for a non-compliant endpoint, and it inherits the 5s bound it lacked before. Right call to flag it; not a regression.

One optional, non-blocking thought: `_adoptModerationPubkey` stores the raw pubkey as the identity but `_warnIfPinMismatch` compares the normalized form. It's safe today because NIP-05 mandates lowercase and the raw storage predates this change, so there's no live bug. If you want to close the theoretical gap where an uppercase answer would warn-silent yet fail to match lowercase event authors in the subscription filter, either normalize once at adoption or drop a one-line note that the asymmetry is safe. Your call, fine to leave.

Nice tests, and thanks for the thorough writeup on the tier split and the out-of-scope items.

mobile/packages/nostr_sdk/lib/nip05/nip05_validor.dart
```diff
@@ -6,7 +6,24 @@ import 'package:dio/dio.dart';
...
+  /// NOT return HTTP redirects, and fetchers MUST ignore any it returns.
+  /// Following a 30x is a spurious-verify vector — a MITM or a misconfigured
+  /// origin can bounce the lookup to a host that answers with an attacker's
+  /// key. Dio's default `validateStatus` accepts 2xx only, so an unfollowed 3xx
+  /// throws and [getPubkey] yields null rather than a redirected key.
+  static var dio = Dio(
+    BaseOptions(
+      connectTimeout: _timeout,
+      receiveTimeout: _timeout,
+      followRedirects: false,
```
dcadenas, line 29: Guard the browser path by returning null when `response.isRedirect` is true before reading `response.data` at `mobile/packages/nostr_sdk/lib/nip05/nip05_validor.dart:61`. The web adapter follows the HTTP redirect but marks the final response as redirected, so `followRedirects: false` alone does not enforce this security check in Flutter web. Add a regression test with a final 200 response whose `isRedirect` flag is true, and verify both `getPubkey` and `valid` reject it while normal 200 and native 302 cases continue to pass.

dcadenas (changes requested):
The NIP-05 redirect protection is incomplete on Flutter web. Add an explicit redirected-response check before parsing the body and cover it with a web-compatible regression test so redirected identities cannot be trusted before merge.

mobile/packages/nostr_sdk/lib/nip05/nip05_validor.dart
realmeylisdev: Confirmed and fixed in 205305ee6c — guard added before the body read, plus tests for both `getPubkey` and `valid`.

You're right on every element. `followRedirects`/`maxRedirects` are read in exactly one place in the whole dio stack (`io_adapter.dart:138-139`); `dio_web_adapter` never reads them, so the guard was inert on web — and we deploy web to `app.divine.video` on every merge to main, with no `kIsWeb` gating on either caller. The final 200 passes the default `validateStatus`, and the flag is set from `options.uri != xhr.responseURL` exactly as you describe.

Removing the guard demonstrates it rather than argues it: `getPubkey` returns the redirect target's key, and `valid()` returns `true` for an impostor. Both tests go red for that reason.

Two follow-ons your finding turned up, both in this PR now:

- **`Nip05Resolver` had the same hole** (f2b73f3997). The #176 DM gate's `status >= 300 && status < 400` guard is inert on web for the same reason, so a bounced lookup echoing the expected key resolved to `matched` — the spurious-APPROVE its own comment warns about. Pre-existing from #5754; folded in rather than deferred since it's the same one-line remedy, and this PR had cited that resolver as the correct reference. Both files' comments asserting the `validateStatus` story now say which platform they describe.
- **One narrow tradeoff worth your eyes**: the flag is a string compare of `options.uri` vs `responseURL`, so an IDN domain (`bob@münchen.de`) reads as a phantom redirect — Dart percent-encodes the host, the browser punycodes it. Those already fail on native (`dart:io` sends the percent-encoded host to DNS), so this fails closed and brings web to parity. Called out in the description; say the word if you'd rather see it normalized here.

realmeylisdev:
@mbradley heads up — the PR grew after your approval, so it's worth a re-look.

Your optional note is now closed in 23162ade5c. I went with normalizing rather than the note, but at the two producers rather than at adoption: `_refreshModerationPubkey` compares the resolved key against the previous one and feeds it to `_subscribedLabelers`, so canonicalizing only inside `_adoptModerationPubkey` would have left the subscription set holding the raw form — and, since the early-return compares before adopting, would have unloaded the labeler it had just re-added. With both producers canonical the identity, the subscription set and the pin comparison agree, and `_warnIfPinMismatch` compares directly.

The bigger change: @dcadenas found that `followRedirects: false` is inert on Flutter web (dio delegates to XMLHttpRequest, which follows the 30x itself), so the redirect control from your review didn't actually bind on the platform we ship to `app.divine.video`. Fixed in 205305ee6c, and the same hole in the #176 `Nip05Resolver` in f2b73f3997. Details in the description.

dcadenas (approved):
🔒 safer moderation pins and NIP-05 lookups

mbradley:
Approved before the redirect round, so noting I've been back through the three follow-ups.

The web guards are correct: the redirect refusal is inert once dio hands off to the browser's XHR path, and `Response.isRedirect` is the only signal that holds on both transports. @dcadenas caught that. Folding the identical `Nip05Resolver` hole in here rather than deferring it closes the same one-liner in the higher-stakes path.

The canonicalization commit normalizes at the two producers rather than only at the pin check, so the adopted identity, the subscription filter and the comparison stay on one form. That was the actual gap, and it's closed.

Thanks, as always.

PR #6107, opened 2026-07-15
fix(db): recover automatically from runtime SQLite corruption

### LINKED ISSUE #6108: fix(db): recover automatically from runtime SQLite corruption
Auto-created from PR #6107.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6107
**Branch:** `fix/db-runtime-corruption-recovery`

hm21:
Closes #6108

## Description

A user reported a video that would not open, with the raw `SqliteException(11): database disk image is malformed` — SQL, parameters and all — rendered on screen. Their local database is structurally corrupt; the NIP-33 d-tag lookup is just the first read that touches the damaged page. Two separate problems came out of that report.

**1. The corruption was unrecoverable in-app.** We already have a full recovery chain (`DatabaseEncryptionBootstrap` → probe → `salvageCorruptEncryptedDatabase` → backup + recreate), but it never fired. `encryptedDatabaseOpensCleanly` is reactive on purpose — it forces Drift's `beforeOpen` cleanup instead of a whole-DB `PRAGMA quick_check`, to keep an unbounded scan off the startup hot path. That cleanup is only `DELETE FROM event WHERE expire_at IS NULL OR expire_at < ?`, so it never reads the surviving rows' `tags`. Corruption confined to pages a *later* query is the first to touch sails past it. Nothing watched the runtime path, so the gate reported "clean" and the read threw again on every launch — permanently, short of a reinstall.

So watch the runtime path too. A Drift `QueryInterceptor` classifies the SQLITE_CORRUPT / SQLITE_NOTADB result code on any statement, reports a non-fatal, and persists a flag. The next launch skips the probe and salvages unconditionally, preserving the local-only data relays cannot re-serve (drafts, clips, pending queues); the corrupt file is kept as a backup and the caches resync.

**2. The screen showed the raw exception.** `videoDetailLoadError` took an `{error}` placeholder, which is what put the SQL on screen. Dropped.

### Design notes

- **Why a restart instead of an in-session repair.** `AppDatabase` is a `keepAlive` Riverpod singleton that most of the service graph holds live subscriptions on. Swapping the file underneath it means tearing that graph down and rebuilding it — a large, risky change for a rare case. A restart does the same thing safely, so `DatabaseCorruptionGate` says so instead. The takeover is total because nothing repairs the open database in place: every subsequent read against the damaged pages throws anyway.
- **Why the flag bypasses the probe rather than fixing the probe.** Making the gate reliable means `PRAGMA quick_check` on every launch — the unbounded startup scan the current design explicitly rejected. The flag pays that cost only on the launch after corruption was actually observed.
- **Why classification reads the result code, not the message.** `SqliteException.toString()` prints the causing statement *and its bound parameters*, and drift serialises that text verbatim across the isolate — so the whole string reaches the classifier. Nostr content is attacker-supplied: a note quoting `database disk image is malformed`, plus any unrelated statement failure that happened to bind it (constraint violation, disk full), would have scheduled a salvage of a perfectly healthy database. Only the header line is classified now, and only by extended result code, masked to the primary code (`& 0xFF`) so every `SQLITE_CORRUPT_*` variant still matches.
- **Why only the encrypted connection is instrumented.** The bootstrap recovers an encrypted database and nothing else. Reporting corruption on the null-key connection would raise the restart gate over a database nothing salvages on the next launch — a promise we cannot keep (see Out of Scope).
- **Why the flag survives a deferred migration.** It is cleared once a launch *settles* it — salvaged, recreated, or proved there was nothing to recover — and a throw keeps it so the retry still recovers. `CipherMigrationOutcome.failed` leaves the database untouched for the next launch, so it settles nothing and keeps the flag too: clearing there would drop the only signal forcing a salvage once a retry finally classifies the file as encrypted. (This matters precisely *because* the interceptor no longer runs on the null-key path — the two changes only hold together.)
- **Why the close button waits.** `setBool` reports a refused write by returning `false` rather than throwing, and `report()` drops every later corruption report, so an unnoticed `false` left the next launch with no reason to salvage. The write now retries, and the prompt holds its button until it settles — releasing on *any* terminal state, because a doomed write must not lock the user into a session that cannot recover at all.
- **l10n:** the `{error}` placeholder is cut from all 18 locales by stripping the trailing separator + placeholder, so each translator's wording is preserved verbatim rather than re-translated. The new copy hedges on drafts and clips (`We'll save what drafts and clips we can`) because salvage keeps only the readable prefix of each table and silently skips unreadable rows — the unconditional promise was not true on *any* path, not just the failing one. `@` metadata records that so a future translator does not un-hedge it.
- The corruption signature was duplicated in `encrypted_database_probe`; both callers now share one classifier.

**Related Issue:** 

## Out of Scope

- **The root cause of the corruption itself.** It could not be reproduced locally, and *why* the probe missed this particular corruption remains a hypothesis. The fix does not depend on that hypothesis being exactly right: whatever the mechanism, the runtime path now catches it and the next launch recovers.
- **The plaintext-DB variant.** If `migratePlaintextToEncrypted` keeps failing on a corrupt plaintext database, `resolveCipherKey` returns null and the recovery gate is never reached. Untouched here — and the detector is now scoped to match, so this PR leaves that population exactly as it found them rather than replacing their one broken screen with a full-app restart prompt that never repairs anything.
- **The other `{error}` placeholders.** ~20 other ARB keys still interpolate raw exceptions into user-facing text (`profileShareFailed`, `videoGridDeleteFailure`, `shareMenuFailedToDeleteVideo`, …), several on DB-touching operations that could surface the same `SqliteException`. Systemic; deserves its own PR.
- **`shouldRepairLocalDatabaseCacheAfterBootstrapError`** matches raw messages the same way, but classifies bootstrap *throws* that carry no user parameters, and it is deliberately looser than the runtime classifier. Narrowing it could regress the repair path, so it is left alone.
- Making the startup gate itself non-reactive (`PRAGMA quick_check` on every launch) — deliberately not done, see above.

## Verification

- `flutter analyze lib test integration_test` — clean; `flutter analyze lib test` in `packages/db_client` — clean
- `dart format` — no changes
- `packages/db_client`: full suite green (826 tests), incl. `database_corruption_test.dart` and the refactored `encrypted_database_probe_test.dart`. The probe tests drive **real** corruption (a genuinely damaged SQLite file) and a real wrong-key open, so they exercise the new result-code classifier end-to-end against actual `DriftRemoteException` strings rather than hand-written fakes.
- `test/startup`, `test/services`, `test/providers`, `test/l10n`, `test/screens/video_detail_screen_test.dart` — green (3782 tests)
- `check_untested_services_floor.sh` — green (new service ships with its test)
- `dart run build_runner build --delete-conflicting-outputs` — `database_provider.g.dart` hash regenerated and committed
- Each fix was mutation-tested: reverting the classifier to message matching, the flag preservation, the `setBool` retry, or the button's durability guard each turns a specific test red.
- Manually tested on device by @hm21: no problems

New tests worth calling out: the gate test pumps the gate through `MaterialApp.builder` exactly as `main.dart` wires it (a `Scaffold` there has no Navigator ancestor); `video_detail_screen_test` pins that a `SqliteException` body never reaches the screen; and the classifier is pinned against corruption signatures quoted in bound parameters.

## Type of Change


rabble:
Reviewed the corruption-recovery path in depth — this is well designed and I found **no data-loss risk**. Detection is gated on SQLite result codes (extended-code masking validated with real `SqliteException(11)`/`SqliteException(779)` cases), not on message text; recovery persists a flag across restarts; the gate UI correctly blocks close until the flag write is durable (`database_corruption_gate_test.dart:110` pins this); and the deferred-migration / failed-write / parallel-statement edge cases are all threaded correctly. The `video_detail_screen_test.dart:515` regression test verifies raw `SqliteException` / `SELECT` text never reaches the screen. Scope is appropriately encrypted-DB-only with the plaintext path documented as out of scope. Non-fatal (not `Reportable`) reporting is the right call for an infrastructure event.

Two minor risk notes, neither blocking:

- `mobile/lib/services/database_corruption_service.dart:104` — if the `_recordError` reporter callback throws despite its void contract, the exception is unhandled in the calling context. Currently safe only because `report()` catches internally; recommend a defensive try/catch at the interceptor's reporter invocation (`database_corruption.dart:88`) for defense-in-depth.
- `database_corruption_service.dart:68` — the "caller must not close the app until `recoveryPersisted` resolves" contract is implicit, not enforced at the type level. No race today, but worth a doc note so a future caller doesn't skip the await.

Nits: `database_corruption.dart:38` positional `group(1)` → could be a named capture for clarity; `database_encryption_bootstrap.dart:177` has a garbled comment ("...recreate from the backup instead). After: the recreated copy...") worth cleaning up.

Approving on substance — the two risks above are hardening, not correctness blockers.

realmeylisdev (approved):
Traced the two riskiest claims down to the dependency source rather than taking the PR body's word for it: sqlite3's `SqliteException.toString()` really does put the causing statement and bound parameters after a newline (confirmed in the pinned 3.3.3 source), and drift really does serialize errors through `.toString()` across the background isolate before `DriftRemoteException` forwards them verbatim — so the header-only, result-code-masked classifier can't be tricked by Nostr content quoting a corruption string. Also checked that drift's `QueryInterceptor` wrapping recurses into transactions/exclusive locks, so corruption inside a `db.transaction()` block still gets reported, not just top-level statements.

Ran the full added/touched test suites (db_client + mobile, 79 tests) locally, plus `flutter analyze`, `dart format`, a build_runner regen (byte-identical to committed output), the untested-services floor, and ARB consistency — all clean. Every locale genuinely mirrors the 3 new strings, not just en. The pending-recovery flag state machine (settle/preserve/clear across salvage, wipe, throw, and deferred-migration outcomes) is fully covered by tests and matches the description exactly.

No issues found.

Issue #5128, opened 2026-06-12
fix(camera): bind native diagnostics sink to UI activity lifecycle


NotThatKindOfDrLiz:
## Summary

`DivineCameraLog.sink` is process-global on each native platform. PR #5110 re-asserts the UI engine's sink at the top of each camera method call so normal camera operations route diagnostics to the Dart UI isolate, even if a background Flutter engine registered the plugin.

There is still a narrow edge where a purely native event can fire after a background engine has temporarily installed its sink and before the next UI-engine camera method call reclaims it.

## Problem

The remaining risk is limited to diagnostics routing, not camera behavior:

- A background Flutter engine can register `divine_camera` and overwrite the singleton log sink.
- A native-only event, such as a volume-key callback, could fire while that background sink is installed.
- That diagnostic line may be routed to the wrong isolate or dropped before the next UI camera method call restores the sink.

## Expected Result

The native diagnostics sink should be bound to the active activity/UI engine lifecycle closely enough that native-only camera diagnostics cannot be stolen by a background engine registration window.

## Acceptance Criteria

- Sink installation/ownership is tied to the UI activity attachment where the platform supports it.
- Background engine registration cannot permanently or temporarily own camera diagnostics for UI-only native events.
- Detach/teardown remains ownership-guarded so one engine cannot clear another engine's sink.
- Android, iOS, and macOS behavior is documented if lifecycle capabilities differ by platform.
- The fix includes targeted tests where feasible, or clear manual verification notes for native lifecycle behavior.

## Related

- Follow-up from PR #5110.

NotThatKindOfDrLiz:
Part of #4341.

PR #6110, opened 2026-07-15
fix(camera): scope iOS Now Playing session to foreground

### LINKED ISSUE #6090: fix(home-widget): incorrect iOS home widget layout
### Summary

The iOS home widget displays an unexpected recording widget with play and pause controls instead of showing only the camera button.

### Environment

iOS

### Steps to Reproduce

1. Install and open the app on iOS.
2. Add the home widget to the Home Screen.
3. View the widget.

### Actual Result

The widget displays a recording interface with play and pause buttons instead of only the camera button.

### Expected Result

The home widget should display only the camera button.

### Regression?

Yes

### Evidence

reported on slack [here](https://planetary-app.slack.com/archives/C09TDBPEJ85/p1784036037617079?thread_ts=1783995460.077149&cid=C09TDBPEJ85)

hm21:
## Description

Fixes the iOS "recording widget" from #6090: a **"Recording" media control with play/pause** that showed on the Lock Screen where users expected only the camera quick-action button.

The reported control is **not** the home-screen `CameraQuickActionWidget` (which only draws the camera button). It is the iOS **Now Playing** media control. To receive AirPods/Bluetooth media-button presses for hands-free remote recording, `VolumeKeyHandler` registers the app as the "Now Playing" app (`MPNowPlayingInfoCenter` title `"Recording"` + `MPRemoteCommandCenter` play/pause). iOS surfaces that as a media control on the Lock Screen / Control Center, and it lingered after the app was backgrounded because the session was only torn down when the recorder screen was disposed — not when the app went to the background with the camera still open.

**Fix:** scope the Now Playing session to the foreground camera session instead of dropping Bluetooth support.

- Claim the session (`MPRemoteCommandCenter` + `nowPlayingInfo`) only while the app is foreground-active.
- Release it on `didEnterBackground` and re-claim it on `didBecomeActive` while remote-record control is enabled.

The app has no `audio` background mode, so it suspends on lock and `didEnterBackground` fires reliably — the "Recording" control can no longer appear on the Lock Screen. Bluetooth/volume remote recording keeps working in the foreground, where recording actually happens (you cannot record while backgrounded, so no trigger is lost).

**Why keep the media session rather than remove it?** Receiving AirPods/Bluetooth media buttons on iOS *requires* being the "Now Playing" app, which is exactly what draws the control — there is no API to get those events invisibly. Users rely on the Bluetooth remote, so foreground-scoping is the correct trade-off.

**Related Issue:** Closes #6090

## Out of Scope

- Android is unaffected — it uses `MediaSessionCompat` and keeps full Bluetooth remote-record support; the cross-platform `RemoteRecordTrigger.bluetooth` path is untouched.
- One unavoidable iOS residual: while you are actively in the camera and swipe down Control Center, a media tile still shows there (never on the Lock Screen). iOS provides no way to receive media-button events without a visible Now Playing surface.

## Tests

- The claim/release decision is extracted into a pure, dependency-free `MediaSessionScopePolicy` (no UIKit/MediaPlayer), mirroring the existing `NostrBridgeAttestationPolicy` precedent in `RunnerTests`. `VolumeKeyHandler` is rewired to delegate its `isEnabled` / `isMediaSessionActive` state to it — **behavior-preserving**: activate/deactivate become unconditional performers gated by the policy's return values, and every lifecycle transition maps 1:1 to the previous inline logic.
- `MediaSessionScopePolicyTests` (in `mobile/ios/RunnerTests/RunnerTests.swift`, importing `divine_camera`) pins the full transition table, including the two #6090 invariants that would previously have been untested: "do not claim the session while backgrounded" and "release it on background so no control lingers on the Lock Screen".
- **Why not a Dart test?** The Now Playing behavior is a native-only side effect (`MPNowPlayingInfoCenter` claim/release driven by app-lifecycle notifications) with no Dart-observable signal — a method-channel test would only assert the outbound `setRemoteRecordControlEnabled` call, which this change does not touch, so it could not fail on a regression (coverage theatre).
- **CI note:** iOS Swift XCTests are not run in CI (the `divine_camera` workflow runs Dart + Android unit tests only), so these are a local-only regression guard, consistent with the existing `RunnerTests` targets. Execute them by running the `RunnerTests` scheme in Xcode.

## Verification

- iOS-native change plus a Swift test; no Dart changed, so no analyzer/coverage/Dart-test impact (pre-push hook confirmed "No Dart files changed").
- `xcrun swiftc -typecheck` of the policy + handler against the iOS SDK passes with 0 errors (only pre-existing `showsRouteButton` / `windows` deprecation warnings on untouched lines).
- Manual on-device (iOS):
  - Open camera, lock phone → no "Recording" media control on the Lock Screen.
  - AirPods/Bluetooth button toggles recording in the camera.
  - Volume buttons toggle recording.
  - Lock → unlock → Bluetooth still works (session re-claimed).

## Type of Change

### DIFF
--- a/mobile/packages/divine_camera/ios/Classes/VolumeKeyHandler.swift
+++ b/mobile/packages/divine_camera/ios/Classes/VolumeKeyHandler.swift
@@ -7,33 +7,47 @@ import UIKit
 
 /// Handles volume button presses and Bluetooth media button events
 /// for remote recording control on iOS.
+///
+/// Receiving AirPods/Bluetooth media-button presses requires becoming the iOS
+/// "Now Playing" app (MPNowPlayingInfoCenter + MPRemoteCommandCenter), which
+/// surfaces a media control (play/pause) on the Lock Screen / Control Center.
+/// To stop that control lingering on the Lock Screen after the app is
+/// backgrounded (#6090), the Now Playing session is scoped to the foreground:
+/// it is released on `didEnterBackground` and re-claimed on `didBecomeActive`
+/// while the handler is enabled. Camera capture is suspended in the background
+/// anyway, so no remote trigger is lost.
 class VolumeKeyHandler: NSObject {
     private var isEnabled = false
     private var volumeKeysEnabled = true  // Can be toggled independently
     private var onTrigger: ((String) -> Void)?
-    
+
     // Volume button detection
     private var volumeView: MPVolumeView?
     private var volumeSlider: UISlider?
     private var lastVolume: Float = 0.5
     private var isObservingVolume = false
-    
+
     // Track volume changes to detect button presses
     private var volumeChangeTimer: Timer?
     private var isInternalVolumeChange = false
-    
+
     // Cooldown after activation to ignore spurious Bluetooth events.
     // Must be long enough to cover delayed events from AirPods/Apple Watch
     // that arrive after audio route changes during camera initialization.
     private var enabledTimestamp: TimeInterval = 0
     private let activationCooldownSeconds: TimeInterval = 3.0
-    
+
     // Debounce between Bluetooth triggers to prevent rapid-fire events
     private var lastBluetoothTriggerTimestamp: TimeInterval = 0
     private let bluetoothDebounceSeconds: TimeInterval = 1.0
-    
+
     // Temporary suppression during camera switch / audio route changes
     private var isSuppressed = false
+
+    // Whether the Now Playing / MPRemoteCommandCenter session is currently
+    // registered. Scoped to the foreground so no media control lingers on the
+    // Lock Screen (#6090).
+    private var isMediaSessionActive = false
 
     // Re-asserts the owning (UI) engine's diagnostics sink before emitting any
     // native-only diagnostic. iOS has no Activity-attachment lifecycle to bind
@@ -47,7 +61,7 @@ class VolumeKeyHandler: NSObject {
         self.onTrigger = onTrigger
         super.init()
     }
-    
+
     /// Enables volume button listening.
     /// Sets up MPRemoteCommandCenter for Bluetooth remotes and volume observation.
     /// Returns true if successfully enabled.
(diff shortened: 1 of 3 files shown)


rabble:
Reviewed — looks solid. Clean `MediaSessionScopePolicy` state-machine extraction (mirrors the `NostrBridgeAttestationPolicy` precedent), symmetric lifecycle observer register/unregister, `weak self` in the `asyncAfter` closure, and a full transition-table test suite where each case can fail on regression. No Dart changes, so no analyzer/coverage/test impact.

One nit before merge:

- `MediaSessionScopePolicy.swift:1-2` — non-standard `// ABOUTME:` comment prefix. Use plain `//` (or `///` for doc) to match the rest of the codebase.

Note the iOS Swift tests are local-only (not run in CI), so the on-device confirmation called out in the PR description is the real gate here.

mobile/packages/divine_camera/ios/Classes/VolumeKeyHandler.swift
```diff
@@ -128,18 +140,99 @@ class VolumeKeyHandler: NSObject {
...
+    }
+
+    /// Releases the Now Playing session. Callers gate this on `scope`; the
+    /// performer itself is unconditional.
+    private func performDeactivateMediaSession() {
+        teardownRemoteCommandCenter()
+        DivineCameraLog.shared.debug("DivineCameraVolumeKeyHandler: Media session deactivated")
+    }
+
+    private func registerAppLifecycleObservers() {
```
realmeylisdev, line 188: Minor/non-blocking: AppDelegate registers GeneratedPluginRegistrant twice — once for the main engine, once for the background notification-action engine via `FlutterLocalNotificationsPlugin.setPluginRegistrantCallback` — so there are two independent `DivineCameraPlugin` instances. If the background isolate ever calls `setRemoteRecordControlEnabled(true)`, we'd get a second `VolumeKeyHandler` registering its own NotificationCenter observers here with nothing to prevent it. Dormant today since that isolate never touches the camera, but worth a note for whoever extends this next.

realmeylisdev (commented):
The MediaSessionScopePolicy extraction is clean and the tests pin every transition nicely — good call isolating this from UIKit/MediaPlayer so it's actually testable. One thing worth nailing down before we treat this as closing #6090 though: the issue's repro steps say "Home Screen"/"home widget" three times and never mention the Lock Screen, and MPNowPlayingInfoCenter can only ever surface on the Lock Screen, Control Center, or CarPlay — never the Home Screen. There's no screenshot on the issue to confirm which surface was actually seen. Can we get confirmation from whoever posted the original Slack screenshot that it really was the Lock Screen tile and not the literal home screen widget, before relying on this to close #6090? If it turns out to be the latter, the issue would need to stay open for a different root cause.

mobile/packages/divine_camera/ios/Classes/VolumeKeyHandler.swift
hm21: Confirmed — AppDelegate.swift:26-31 registers the registrant on both engines, and every setRemoteRecordControlEnabled caller is on the UI-engine path, so it's dormant as you say. Documented the single-handler assumption next to the existing #5128 multi-engine note in a4c524b92.

hm21:
@rabble on the `// ABOUTME:` nit — I don't think this one holds. That prefix *is* the convention here: 2,488 files use it, 19 of them Swift, and all five siblings in the same directory (`CameraController`, `DivineCameraLog`, `DivineCameraPlugin`, `PhotoCaptureDelegate`, `VolumeKeyHandler`) open with it — plus the `review-before-commit` skill checks ABOUTME comments still describe their file. `MediaSessionScopePolicy.swift` is matching the codebase, so I've left it as-is. Happy to change it if you're proposing we drop the convention repo-wide, but that's its own PR.

@realmeylisdev on Lock Screen vs Home Screen — good catch to ask, and the code answers it in the fix's favour. `CameraQuickActionWidget` is a `StaticConfiguration` rendering only `camera.fill` on both supported families (`systemSmall`, `accessoryCircular`) — it has no play/pause transport to draw at all. So the reported symptom can't come from the home widget, and Now Playing can't render on the Home Screen either. The only thing in the app drawing a play/pause transport is the `MPRemoteCommandCenter` session this PR scopes. Note `accessoryCircular` is itself a Lock Screen family, so a reporter with that widget placed would plausibly call the Lock Screen tile their "widget" — that reads as the imprecision here. Still worth eyeballing the Slack screenshot before we close #6090; if it really is the literal Home Screen widget, this PR stands on its own but #6090 stays open.

realmeylisdev (approved):
Verified against the code — holds. `CameraQuickActionWidget` is a `StaticConfiguration` drawing only `camera.fill` (`.systemSmall`/`.accessoryCircular`), so the home widget has no play/pause transport to produce the reported control; the only thing rendering a "Recording" play/pause tile is the Now Playing session this PR foreground-scopes. Fix is correct regardless of surface — approving the code. Agree #6090 should stay open until the Slack screenshot confirms it was the Lock Screen tile.

PR #4478, opened 2026-05-17
fix(comments): cache survives addressable-event edit by dual-keying on addressable id

realmeylisdev:
## Description

`CommentsRepository._commentCountCache` was keyed only by `rootEventId`. For NIP-71 addressable videos (Kind 30000-39999), a metadata edit publishes a replacement event with a fresh `id` but the same `kind:pubkey:d-tag` address. After such an edit the per-video \`VideoInteractionsBloc\` is re-created on the next mount (profile re-entry, fullscreen open, home-feed scroll-back), and its \`_onFetchRequested\` calls \`getCommentsCount(newEventId, rootAddressableId: address)\`. The cache misses on the new event id, so the repo falls through to a relay NIP-45 COUNT against \`#E=newEventId\` ∪ \`#A=address\`. On \`relay.divine.video\` this COUNT can transiently return 0 — the same shape as the kind-7 issue addressed for likes in #4477 — and the comment-count badge collapses to 0.

This PR adds a companion cache keyed by the stable addressable id and routes every read/write through helper methods:

- \`_readCachedCommentCount(rootEventId, rootAddressableId:)\` — checks event-id first, then the addressable cache.
- \`_writeCachedCommentCount(...)\` — dual-writes when an addressable id is in scope.
- \`_adjustCachedCommentCount(rootEventId, delta, rootAddressableId:)\` — clamps at zero, dual-writes the new value.

All call sites that already had \`rootAddressableId\` in scope (\`loadComments\`, \`postComment\`, \`getCommentsCount\`) now pass it through. \`deleteComment\` and \`updateCachedCommentCount\` gain an optional \`rootAddressableId\` parameter, and both \`CommentsBloc\` \`deleteComment\` call sites forward \`state.rootAddressableId\` so post-edit fetches observe the decrement via the companion cache. \`clearCommentCountCache\` wipes both maps.

**Why a separate PR (not extending #4477)**: per \`agent_workflow.md\` rule 3 — independent fixes ship as parallel PRs. The likes fix lives in \`VideoInteractionsBloc\` (BloC layer); this comments fix lives in \`CommentsRepository\` (repository layer). Neither depends on the other; reviewers handle each in isolation.

Closes #4481

**Related Issue:** Related to #4432 (likes companion shipped in #4477). Does not claim closure of #4432 — that issue's primary symptom is likes.

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
[...]

PR #6123, opened 2026-07-15
fix(likes): own "liked" heart un-fills on your own video after an edit

### LINKED ISSUE #6020: fix(likes): own "liked" heart un-fills on your own video after an edit
## Summary

After a video is edited (which mints a new kind-34236 event id for the same coordinate), the user's own "liked" heart can show as **un-filled** even though their reaction still exists. Own-like state is keyed only by the target event id, with no coordinate/addressable dual-key, so it does not follow the video across an edit.

## Evidence

- `LikeRecord` stores only `targetEventId`, `reactionEventId`, `createdAt` — there is no `addressableId` field (`mobile/packages/likes_repository/lib/src/models/like_record.dart`).
- `_extractTargetEventId` returns the **first `e` tag only**, ignoring the `a` (coordinate) tag (`mobile/packages/likes_repository/lib/src/likes_repository.dart:1506-1513`).
- Own-like state is written and read by event id: `_likeRecords[eventId] = …` (`likes_repository.dart:298`, `:341`) and `isLiked(String eventId) => _likeRecords.containsKey(eventId)` (`:207-209`).
- After an edit, the reaction still e-tags the **old** id, so `_likeRecords` holds the old id; `isLiked(newId)` returns false, and the bloc sets the heart off: `final isLiked = likedIds.contains(_eventId)` with `_eventId` = the new id (`mobile/lib/blocs/video_interactions/video_interactions_bloc.dart:95`).

This is orthogonal to the displayed count and is unaffected by the backend fix — it is purely client-side own-like state keyed by event id.

Precedent: comments got an addressable-id dual-key cache in #4478; likes never did.

## Suggested fix

Dual-key own-like state by coordinate (addressable id) in addition to event id, so a like on **any** version of a coordinate lights the heart on the current version. Mirror the comments #4478 approach: add an addressable-id companion index in `LikesRepository`, populate it from the reaction's `a` tag, and have `isLiked`/`getLikedEventIds` resolve by coordinate when the event id misses.

## Test plan

- `likes_repository` unit test: record a like whose reaction carries `e`(oldId) + `a`(coordinate); assert `isLiked(newId)` resolves true via the coordinate after an edit.
- `video_interactions_bloc` test: heart stays filled when `_eventId` changes from the old to the new version id for the same coordinate.

---

### Context

Follows from #4432 (lost likes on edit). That issue's core symptom — the displayed like count collapsing after an edit — was a backend counting bug and was fixed by divinevideo/divine-funnelcake#568 (migration `000180`, coordinate-aware counting), deployed to production 2026-07-10 06:42 UTC. This issue tracks a residual, client-side own-like-state gap that survives that fix.

realmeylisdev:
## Description

After a video edit mints a new event id for the same `d`-tag coordinate, own-like state (`LikesRepository._likeRecords`, the `personal_reactions` table) was keyed only by target event id, so a like recorded pre-edit became invisible post-edit — the heart un-fills even though the underlying Kind 7 reaction (and its `a` tag) is untouched and still live on relays.

Adds an addressable-id companion index alongside the existing event-id index:
- `personal_reactions` gets a new nullable `addressable_id` column (existing `_addColumnIfMissing` runtime migration pattern, no `schemaVersion` bump — same approach already used for `event.d_tag` / `outgoing_dms.send_batch_id`).
- `LikesRepository` gains an in-memory `_likeRecordsByAddressableId` companion map, persisted via the new column so cold app starts resolve correctly too (not just warm-cache hits, unlike #4478's in-memory-only precedent for comments).
- New `getLikeRecordByCoordinate` / `isLikedByCoordinate` / `isLikedResolvingCoordinate` methods; `isLiked(eventId)`'s existing signature is unchanged.
- `watchLikedAddressableIds()` stream added alongside `watchLikedEventIds()` so `VideoInteractionsBloc`'s live subscription also stays correct — without it, `watchLikedEventIds()` (which ticks on *any* like/unlike anywhere in the app) would silently flip a coordinate-only-liked heart back to unliked the next time anything else was liked/unliked.
- Closes a related duplicate-reaction bug found during investigation: re-liking a video post-edit (while the heart was incorrectly unfilled) created a second live reaction while the first became permanently un-deletable via the normal unlike flow. `likeEvent`'s already-liked guard and `unlikeEvent`'s target resolution now both check the coordinate too.
- Adds the `_likeCountCache` addressable-id dual-key that #4478's own PR body suggested as a follow-up and was never shipped.

No backend, relay, or protocol changes — confirmed via code read, NIP-25/NIP-01 spec read, and an empirical test against the local funnelcake-relay (local Docker stack) that mobile's `sendLike` already emits both `e` and `a` tags, and that the relay correctly serves reactions by both tags regardless of whether the target's addressable version has since been superseded.

**Related Issue:** Closes #6020

## Review round 2 (dcadenas, hm21)

- `executeLikeAction`'s cross-device guard now drops the queued `pending_` placeholder (memory + storage) when reconciling, so a later unlike deletes the real reaction instead of no-oping on the placeholder; it also awaits `_ensureInitialized()` first since replay races `initialize()` at provider build.
- `syncUserReactions` backfills `addressableId` onto stored coordinate-less records from the relay copy of the **same** reaction (same-reaction-id restriction so a different older reaction can never repoint a record).
- `initialize()` runs that backfill sync at startup when any loaded row lacks a coordinate, so pre-migration likes self-heal without a Liked Videos visit; the result is persisted and survives offline restarts (pinned end-to-end by `likes_repository_pre_column_db_test.dart` over a real pre-column `AppDatabase`).
- Fixed the red `Tests` job: `isLikedResolvingCoordinate` is now stubbed in `video_interactions_bloc_key_test.dart` plus the same latent gap in three sibling widget tests.

## Out of Scope

- Downvotes have the identical structural gap (no `addressableId` plumbing at all) but zero current blast radius — they're only ever applied to comments (non-addressable) today. Filed as a separate tracked issue with the supporting evidence rather than fixed here.
- No reconciliation of pre-existing duplicate live relay reactions created before this fix (the orphaned wire events themselves). They are invisible to other users since liker counts dedupe by pubkey. Local rows *are* now reconciled — see review round 2 above.

## Verification

- `flutter test` — `likes_repository` (189 tests, including the new pre-column-DB end-to-end regression), `db_client` (823 tests, including drift schema migration verification), app-layer widget/bloc suites for the touched files (302 tests via pre-push hook)
- `likes_repository` coverage: 90.90% (min_coverage gate: 62%)
- Empirical relay test against local Docker stack (funnelcake-relay) confirming reaction tag survival and `#a`-filter resolution across an addressable-event edit

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/video_interactions/video_interactions_bloc.dart
+++ b/mobile/lib/blocs/video_interactions/video_interactions_bloc.dart
@@ -99,11 +99,27 @@ class VideoInteractionsBloc
     VideoInteractionsSubscriptionRequested event,
     Emitter<VideoInteractionsState> emit,
   ) {
+    // Track the latest snapshot from both like streams so either one
+    // ticking recomputes membership from BOTH — a like recorded against a
+    // pre-edit event id for this coordinate only shows up in the
+    // addressable-id stream, and watchLikedEventIds ticks on every
+    // like/unlike anywhere in the app (not just this video), so treating
+    // the two streams independently would let a same-coordinate-only hit
+    // get silently overwritten back to unliked the next time anything
+    // else is liked/unliked (#6020).
+    var likedEventIds = const <String>[];
+    var likedAddressableIds = const <String>{};
+    bool resolveIsLiked() =>
+        likedEventIds.contains(_eventId) ||
+        (_addressableId != null &&
+            likedAddressableIds.contains(_addressableId));
+
     final subscriptions = [
       emit.forEach<List<String>>(
         _likesRepository.watchLikedEventIds(),
-        onData: (likedIds) {
-          final isLiked = likedIds.contains(_eventId);
+        onData: (ids) {
+          likedEventIds = ids;
+          final isLiked = resolveIsLiked();
           // Load-bearing: when [_onLikeToggled] has already emitted its
           // optimistic state, state.isLiked matches and we no-op here.
           // This guarantees a single bloc emit per tap and avoids the
@@ -118,6 +134,16 @@ class VideoInteractionsBloc
           return state.copyWith(isLiked: isLiked);
         },
       ),
+      if (_addressableId != null)
+        emit.forEach<Set<String>>(
+          _likesRepository.watchLikedAddressableIds(),
+          onData: (ids) {
+            likedAddressableIds = ids;
+            final isLiked = resolveIsLiked();
+            if (isLiked == state.isLiked) return state;
+            return state.copyWith(isLiked: isLiked);
+          },
+        ),
       if (_addressableId != null)
         emit.forEach<Set<String>>(
           _repostsRepository.watchRepostedAddressableIds(),
@@ -160,8 +186,14 @@ class VideoInteractionsBloc
     emit(state.copyWith(status: VideoInteractionsStatus.loading));
 
     try {
-      // Check if liked (fast - from local cache)
-      final isLiked = await _likesRepository.isLiked(_eventId);
+      // Check if liked (fast - from local cache). Falls back to the
+      // addressable coordinate when the event id misses, so a like
+      // recorded against a pre-edit version of this video still resolves
+      // (#6020).
+      final isLiked = await _likesRepository.isLikedResolvingCoordinate(
+        eventId: _eventId,
+        addressableId: _addressableId,
+      );
 
       // Check if reposted (fast - from local cache) if addressable
       final isReposted =
--- a/mobile/packages/db_client/lib/src/database/app_database.dart
+++ b/mobile/packages/db_client/lib/src/database/app_database.dart
@@ -751,6 +751,15 @@ class AppDatabase extends _$AppDatabase {
     await _addColumnIfMissing('outgoing_dms', 'send_batch_id', 'TEXT');
     await _addColumnIfMissing('direct_messages', 'send_batch_id', 'TEXT');
 
+    // Addressable coordinate for own-like state, so it survives a target
+    // edit that mints a new event id for the same coordinate (#6020). Same
+    // runtime ADD-COLUMN-IF-MISSING pattern as d_tag/send_batch_id above.
+    await _addColumnIfMissing('personal_reactions', 'addressable_id', 'TEXT');
+    await customStatement('''
+      CREATE INDEX IF NOT EXISTS idx_personal_reactions_addressable_id
+      ON personal_reactions (addressable_id)
+    ''');
+
     // Create the event indexes unconditionally. The `List<Index> get
     // indexes` getter on NostrEvents is not wired through
     // `@DriftDatabase(...)`, so Drift's m.createAll() never creates them —
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -522,12 +522,30 @@ class LikesRepository {
   ///
   /// This method bypasses offline queuing and directly publishes to relays.
   /// Used by PendingActionService to execute queued actions.
+  ///
+  /// If [addressableId] is already liked by a *different* reaction than the
+  /// one queued for [eventId] — e.g. cross-device sync delivered a like for
+  /// this coordinate (under a different event id) while this action sat in
+  /// the offline queue — reconciles to that reaction instead of publishing
+  /// a duplicate. The app-layer queue's own dedup only cancels opposite
+  /// actions on the same event id; it has no visibility into reactions
+  /// arriving from other devices, so this is the only guard against a
+  /// duplicate live reaction on replay (#6020).
   Future<String> executeLikeAction({
     required String eventId,
     required String authorPubkey,
     String? addressableId,
     int? targetKind,
   }) async {
+    if (addressableId != null && addressableId.isNotEmpty) {
+      final byCoordinate = _likeRecordsByAddressableId[addressableId];
+      final ownReactionEventId = _likeRecords[eventId]?.reactionEventId;
+      if (byCoordinate != null &&
+          byCoordinate.reactionEventId != ownReactionEventId) {
+        return byCoordinate.reactionEventId;
+      }
+    }
+
     // Publish Kind 7 reaction event via NostrClient
     final reactionEvent = await _nostrClient.sendLike(
       eventId,
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -108,6 +108,18 @@ class LikesRepository {
   /// In-memory cache of like records keyed by target event ID.
   final Map<String, LikeRecord> _likeRecords = {};
 
+  /// Companion index of [_likeRecords], keyed by addressable coordinate
+  /// (`kind:pubkey:d-tag`) instead of target event ID.
+  ///
+  /// Lets own-like state survive a target edit that mints a new event id
+  /// for the same coordinate (#6020) — a like recorded against the pre-edit
+  /// id stays resolvable via [isLikedByCoordinate]/[getLikeRecordByCoordinate]
+  /// after the edit. Populated everywhere [_likeRecords] is written (see
+  /// [_indexLikeRecord]) whenever an addressable id is known, including on
+  /// load from persisted storage, so cold app starts resolve correctly too
+  /// (unlike the in-memory-only precedent in #4478's comment count cache).
+  final Map<String, LikeRecord> _likeRecordsByAddressableId = {};
+
   /// In-memory cache of downvote records keyed by target event ID.
   ///
   /// Mirror of [_likeRecords] for kind-7 reactions whose content is `-`.
@@ -122,8 +134,31 @@ class LikesRepository {
   /// stays consistent with optimistic UI updates.
   final Map<String, int> _likeCountCache = {};
 
+  /// Companion of [_likeCountCache], keyed by addressable coordinate.
+  ///
+  /// Mirrors the dual-key pattern `comments_repository` shipped in #4478:
+  /// a cached count survives a target edit that changes the event id, as
+  /// long as the coordinate stayed the same. Read falls back to this map
+  /// only on an event-id miss; writes are always dual when a coordinate is
+  /// known (see [_readCachedLikeCount]/[_writeCachedLikeCount]).
+  final Map<String, int> _likeCountCacheByAddressableId = {};
+
   /// Reactive stream controller for liked event IDs (ordered by recency).
   final _likedIdsController = BehaviorSubject<List<String>>.seeded([]);
+
+  /// Reactive stream controller for liked addressable coordinates.
+  ///
+  /// Companion of [_likedIdsController], ticked alongside it (see
+  /// [_emitLikedIds]) so a live subscriber (e.g. a video interactions BLoC)
+  /// can resolve "is this coordinate liked" reactively, not just via a
+  /// one-shot fetch — without this, a consumer subscribed only to
+  /// [watchLikedEventIds] would have its own-like state silently flip back
+  /// to unliked the next time *any* like/unlike ticks the stream anywhere
+  /// in the app, since that stream is blind to coordinate membership
+  /// (#6020).
+  final _likedAddressableIdsController = BehaviorSubject<Set<String>>.seeded(
+    <String>{},
+  );
 
   /// Reactive stream controller for downvoted event IDs (ordered by recency).
   final _downvotedIdsController = BehaviorSubject<List<String>>.seeded(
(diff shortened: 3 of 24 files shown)


mobile/packages/likes_repository/lib/src/likes_repository.dart
```diff
@@ -372,12 +522,30 @@ class LikesRepository {
...
     required String authorPubkey,
     String? addressableId,
     int? targetKind,
   }) async {
+    if (addressableId != null && addressableId.isNotEmpty) {
+      final byCoordinate = _likeRecordsByAddressableId[addressableId];
+      final ownReactionEventId = _likeRecords[eventId]?.reactionEventId;
+      if (byCoordinate != null &&
+          byCoordinate.reactionEventId != ownReactionEventId) {
+        return byCoordinate.reactionEventId;
```
dcadenas, line 562: Reconcile the queued placeholder before returning this coordinate reaction at `mobile/packages/likes_repository/lib/src/likes_repository.dart:543-545`. The guard returns the real cross-device reaction ID but leaves the pending event-ID record in memory and storage; a later unlike finds that placeholder first at `:603`, removes it, and skips Kind 5 deletion at `:654-656` because its ID is pending. Extend the regression test through the later unlike and assert that the real reaction ID is deleted and both event and coordinate lookups are cleared.

mobile/packages/db_client/lib/src/database/app_database.dart
```diff
@@ -751,6 +751,15 @@ class AppDatabase extends _$AppDatabase {
     await _addColumnIfMissing('outgoing_dms', 'send_batch_id', 'TEXT');
     await _addColumnIfMissing('direct_messages', 'send_batch_id', 'TEXT');
 
+    // Addressable coordinate for own-like state, so it survives a target
+    // edit that mints a new event id for the same coordinate (#6020). Same
+    // runtime ADD-COLUMN-IF-MISSING pattern as d_tag/send_batch_id above.
+    await _addColumnIfMissing('personal_reactions', 'addressable_id', 'TEXT');
```
dcadenas, line 757: Hydrate legacy personal-reaction coordinates during upgrade or normal startup. `mobile/packages/db_client/lib/src/database/app_database.dart:754-761` adds `addressable_id` but leaves existing rows null, while normal initialization indexes them unchanged and only observes one reaction at `mobile/packages/likes_repository/lib/src/likes_repository.dart:1592-1624`. Add a regression test that opens a pre-column database through the production initialization path, resolves an edited video as liked without visiting Liked Videos, and preserves that result after an offline restart.

dcadenas (dismissed):
Before this PR can merge, reconcile queued placeholders when a cross-device coordinate match is found and hydrate coordinates for legacy personal reactions during upgrade or startup. Without both fixes, a later unlike can leave the real relay like active, while existing users can still see edited videos as unliked until they visit Liked Videos.

mobile/lib/blocs/video_interactions/video_interactions_bloc.dart
```diff
@@ -160,8 +186,14 @@ class VideoInteractionsBloc
     emit(state.copyWith(status: VideoInteractionsStatus.loading));
 
     try {
-      // Check if liked (fast - from local cache)
-      final isLiked = await _likesRepository.isLiked(_eventId);
+      // Check if liked (fast - from local cache). Falls back to the
+      // addressable coordinate when the event id misses, so a like
+      // recorded against a pre-edit version of this video still resolves
+      // (#6020).
+      final isLiked = await _likesRepository.isLikedResolvingCoordinate(
```
hm21, line 193: This breaks `test/widgets/video_feed_item/video_interactions_bloc_key_test.dart` ("does not keep fetched counts after feed cell reuse") — it's the CI failure. That test stubs `isLiked` only, so the switch to `isLikedResolvingCoordinate` hits an unstubbed mock:

```
VideoInteractionsBloc: Failed to fetch for video-a - type 'Null' is not a subtype of type 'Future<bool>'
```

The fetch throws, counts never load, `find.text('100')` finds nothing. Confirmed it passes on `origin/main`. Fix is a stub at `video_interactions_bloc_key_test.dart:53`. Worth grepping for other mocks of `LikesRepository` that stub `isLiked` — the analyzer won't catch these.

mobile/packages/likes_repository/lib/src/likes_repository.dart
```diff
@@ -1042,13 +1272,14 @@ class LikesRepository {
...
             ),
+            addressableId: _extractAddressableId(event),
           );
 
           // Only update if we don't have this record or the new one is newer
           final existing = _likeRecords[targetId];
           if (existing == null ||
               record.createdAt.isAfter(existing.createdAt)) {
-            _likeRecords[targetId] = record;
+            _indexLikeRecord(record);
```
hm21, line 1315: The guard above (`existing == null || record.createdAt.isAfter(existing.createdAt)`) means this never runs for a target already in `_likeRecords` — which on a cold start is every persisted row loaded by `_loadFromStorage`.

For an install that liked a video before this ships, the row has `addressable_id` NULL (new nullable column, no backfill). Sync then fetches the reaction *with* the `a` tag, but `existing != null` and the relay's `createdAt` is never *after* the stored one (equal when the row came from a prior sync; older when it came from the optimistic path, which stamps `DateTime.now()`). So it's skipped and the coordinate is never backfilled — not "until next sync", but permanently.

That undercuts the "fix-forward, `syncUserReactions()` self-heals" reasoning in the PR body: the whole pre-existing like corpus keeps bug #6020.

Reproduces — this fails on the branch and passes with the fix below:

```dart
// persisted pre-migration row: no addressableId
when(() => mockLocalStorage.getAllLikeRecords()).thenAnswer(
  (_) async => [
    LikeRecord(
      targetEventId: oldEventId,
      reactionEventId: reactionId,
      createdAt: DateTime.fromMillisecondsSinceEpoch(timestamp * 1000),
    ),
  ],
);
// relay reaction DOES carry the a tag
// tags: [['e', oldEventId], ['a', coordinate]]
await repository.syncUserReactions();

expect(
  await repository.isLikedResolvingCoordinate(
    eventId: newEventId,
    addressableId: coordinate,
  ),
  isTrue,
); // Actual: false
```

Letting the sync re-index when it has a coordinate the stored record lacks is enough:

```dart
final needsBackfill = existing != null &&
    (existing.addressableId == null || existing.addressableId!.isEmpty) &&
    record.addressableId != null;
if (existing == null ||
    record.createdAt.isAfter(existing.createdAt) ||
    needsBackfill) {
```

mobile/packages/likes_repository/lib/src/likes_repository.dart
hm21, line 562: The early return skips the duplicate publish (good), but nothing reconciles `_likeRecords[eventId]` — it keeps the `pending_` placeholder, in memory and in storage.

So the next unlike resolves that placeholder first (event-id lookup precedes the coordinate fallback), hits the `startsWith('pending_')` short-circuit in `unlikeEvent`, and returns without publishing a Kind 5. The real cross-device reaction stays live and is now un-deletable via the normal flow — and since `_deindexLikeRecord` also drops the coordinate entry, the next like publishes a *second* real reaction. That's the duplicate/un-deletable pair this PR is meant to close.

Reproduces (offline like → cross-device sync of the same coordinate → replay → unlike):

```dart
final resolved = await repository.executeLikeAction(
  eventId: queuedEventId,
  authorPubkey: authorPubkey,
  addressableId: coordinate,
);
expect(resolved, equals(realReactionId));           // passes
verifyNever(() => mockNostrClient.sendLike(...));   // passes

await repository.unlikeEvent(queuedEventId, addressableId: coordinate);
verify(() => mockNostrClient.deleteEvent(realReactionId)).called(1);
// fails: deleteEvent never called
```

The placeholder needs to be dropped (and persisted) when reconciling. Watch out that `_deindexLikeRecord(eventId)` isn't the tool for it — it would look up the placeholder's `addressableId` and evict the *cross-device* record from `_likeRecordsByAddressableId`.

hm21 (commented):
Approach is right (coordinate companion index + persisted column), but three things block it.

**1. CI is red — the branch broke `video_interactions_bloc_key_test.dart`.** Verified it passes on `origin/main` and fails here, so it's this diff. The PR body says `flutter test` passes; please re-check before pushing.

**2. The fix never reaches existing likes.** `syncUserReactions`'s freshness guard skips any target already in `_likeRecords`, so a pre-migration row (`addressable_id` NULL) never gets backfilled even though the synced reaction carries the `a` tag. The "Out of Scope" note leans on `syncUserReactions()` self-healing — it doesn't. Every like made before this ships stays broken, permanently.

**3. The new `executeLikeAction` guard trades a duplicate publish for an orphaned reaction.** It stops the duplicate, but leaves the `pending_` placeholder in place, so the later unlike no-ops on the wire and the real reaction becomes un-deletable — the exact failure mode this PR says it closes.

2 and 3 both reproduce in unit tests; details inline.

mobile/packages/likes_repository/lib/src/likes_repository.dart
realmeylisdev: Fixed in 175f0703d0 — the guard now drops the pending_ placeholder from memory and storage before returning the cross-device reaction id (plain map remove rather than _deindexLikeRecord, so the coordinate entry keeps the real record). The new regression test runs the full replay → unlike flow and asserts the real reaction id is deleted on the wire and both event and coordinate lookups are cleared.

mobile/packages/db_client/lib/src/database/app_database.dart
realmeylisdev: Fixed in 175f0703d0. The hydration lives in LikesRepository.initialize(): when any loaded row has a null coordinate it runs syncUserReactions(), which now backfills the coordinate from the relay reaction's a tag and persists it — the DB layer can't derive kind:pubkey:d locally, so the relay a tag is the only reliable source. The requested regression test is likes_repository_pre_column_db_test.dart: it opens a pre-column database through the production initialize() path only, resolves the edited video as liked, and asserts the result survives an offline restart with zero relay calls. One cost note: rows whose reactions genuinely carry no a tag (e.g. comment likes) can never backfill, so those installs re-run the startup sync once per launch.

mobile/lib/blocs/video_interactions/video_interactions_bloc.dart
realmeylisdev: Fixed in 175f0703d0 — stubbed isLikedResolvingCoordinate there, and the grep you suggested found the same gap in feed_videos_repo_swap_test.dart, pooled_video_feed_item_repo_swap_test.dart, and feed_videos_test.dart (green today only because they don't assert on fetched counts). All four now stub it. Local run of the previously red test plus siblings passes.

mobile/packages/likes_repository/lib/src/likes_repository.dart
realmeylisdev: Fixed in 175f0703d0, with one tightening of the suggested guard: the exemption also requires existing.reactionEventId == record.reactionEventId. Without that, a different *older* reaction carrying an a tag could replace a stored newer coordinate-less record and repoint unlike at the wrong wire event (reachable when two live reactions exist for one target and the newer one lacks the tag). The same-id form still covers the real backfill case — the relay copy of the same reaction whose tag was never stored — so your repro passes unchanged. Added both tests: same-id backfill accepted, different-older-id rejected.

mobile/packages/likes_repository/lib/src/likes_repository.dart
realmeylisdev: Fixed in 175f0703d0 — the guard drops the placeholder with a plain _likeRecords.remove(eventId) + storage deleteLikeRecord (not _deindexLikeRecord, per your warning about evicting the cross-device coordinate entry) and re-emits. Also added _ensureInitialized() at the top of executeLikeAction: replay races initialize() at provider build and the coordinate guard is memory-only, so an early replay would have bypassed it entirely. Your repro is now the test 'drops the queued placeholder when reconciling…', extended through the unlike with the Kind 5 assertion.

realmeylisdev:
@dcadenas @hm21 all five findings addressed in 175f0703d0: placeholder reconcile (+ _ensureInitialized on replay), same-reaction-id coordinate backfill in sync, startup hydration for pre-migration rows with an end-to-end pre-column-DB regression test, and the missing isLikedResolvingCoordinate stubs (CI failure + three latent siblings). Details on each thread.

hm21 (approved):
LGTM. All three blockers are genuinely fixed, not papered over — the same-reaction-id restriction on the backfill guard and the `_ensureInitialized` before the memory-only coordinate guard are the right tightenings, and the regression tests cover the reconcile-through-unlike and different-older-id paths. Thanks for the thorough follow-up.

Issue #4398, opened 2026-05-15
fix(crashlytics): map API category names to existing assets safely


rabble:
Crashlytics 1.0.13 shows category UI deriving SVG paths that do not exist in the bundle.

Observed in 1.0.13:
- Crashlytics issue 20e5b4c3c153605b2ae2ed5f36bddc40: 2 nonfatal events, 1 user.
- Error: Unable to load asset: assets/categories/beverages.svg.
- The same crash family has also appeared for assets/categories/fighting.svg.
- Code: mobile/lib/widgets/categories/category_visuals.dart derives the SVG path from the raw category string around lines 29-35.
- Assets: mobile/assets/categories contains beverage.svg, but not beverages.svg; it also does not contain fighting.svg.

This is a deterministic app bug: category names from API/model data are not guaranteed to match asset filenames one-to-one.

Acceptance criteria:
- Category asset lookup uses an explicit mapping or existence-safe fallback instead of raw lowercased category text.
- All categories returned by the current app/API either resolve to an existing bundled asset or intentionally show a fallback with no asset-load exception.
- Add a test that validates every known category mapping points to an existing file or a defined fallback.
- Keep UI behavior stable for existing valid category assets.

PR #6131, opened 2026-07-16
fix(categories): fall back to emoji when a category has no bundled SVG

### LINKED ISSUE #6116: fix(categories): missing assets/categories/fighting.svg throws instead of falling back (Crashlytics 20e5b4c3)
## Crashlytics
- Issue: `20e5b4c3c153605b2ae2ed5f36bddc40` ([console](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/20e5b4c3c153605b2ae2ed5f36bddc40))
- Type: non-fatal · **66 impacted users / 638 events**
- Versions: first 1.0.9 → last **1.0.16**

## Signature
`package:vector_graphics/..._VectorGraphicWidgetState._loadPicture`
`Unable to load asset: "assets/categories/fighting.svg"` (thrown, runZonedGuarded).

## Root cause (confirmed)
`lib/widgets/categories/category_visuals.dart:33` builds `assetPath: 'assets/categories/$assetName.svg'` dynamically from a category name. **`fighting.svg` does not exist** in `mobile/assets/categories/` (present: action, adventure, animals, … — no `fighting`). `category_glyph.dart:11` even documents that "name can arrive with no matching asset," but the load still throws instead of falling back.

## Suggested fix
Either (a) add the missing `assets/categories/fighting.svg`, and/or (b) make `category_visuals`/`category_glyph` fall back to a default glyph when the asset is absent — an unknown category name must never throw. A test asserting an unknown category renders the fallback (not an exception) pins it.

hm21:
## What

Unknown backend category names (e.g. `fighting`) derived an `assetPath` pointing at a missing `assets/categories/<name>.svg`. `CategoryVisuals.forCategory` now resolves `assetPath` to `null` when the derived basename isn't bundled, and `CategoryGlyph` renders the category emoji directly for a `null` path — so `flutter_svg` is never asked to load a file that doesn't exist.

## Why the existing errorBuilder wasn't enough

#4398 already wired an `errorBuilder` on `CategoryGlyph`, and it does render the emoji when the SVG is missing — but the Crashlytics non-fatals kept coming in through 1.0.16. The reason is in `vector_graphics`' `_loadPicture`: it attaches `result.whenComplete(...)` and discards the returned future. On a missing asset, `_loadAssetBytes`'s try/catch handles the awaited subscription (so the `errorBuilder` fires and the UI is fine), but the discarded `whenComplete` future completes with the same error, has no listener, and escalates to the app's `runZonedGuarded` handler → logged as a non-fatal. As long as `SvgPicture.asset` is *asked* to load a missing path, the event fires; `errorBuilder` only fixes the pixels, not the report.

So the fix is to never hand a missing path to `SvgPicture` in the first place. The two call sites' defensive `if (visuals.assetPath != null)` gates were already there but dead (`forCategory` always returned a non-null path) — this wires the intended null path through and keeps the emoji mascot for unknown categories.

Chose the fallback (issue option b) over adding `fighting.svg` (option a): backend category names are an open-ended, uncurated stream, so adding one asset is whack-a-mole — the class fix stops all of them.

## Tests

- `forCategory('fighting')` → `assetPath` is `null`; `CategoryGlyph(null)` renders the emoji with no `SvgPicture`; an end-to-end test pins the exact Crashlytics scenario.
- A drift-guard test asserts `bundledAssetBasenames` equals the actual `assets/categories/*.svg` files on disk, so deleting/adding an SVG without updating the set fails CI instead of quietly reintroducing the crash.

`flutter analyze` clean; the 4 affected test suites pass locally.

Closes #6116

### DIFF
--- a/mobile/lib/widgets/categories/category_glyph.dart
+++ b/mobile/lib/widgets/categories/category_glyph.dart
@@ -4,13 +4,16 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 
-/// Renders the SVG at [assetPath], falling back to [emoji] when the asset is
-/// not in the bundle.
+/// Renders the SVG at [assetPath], falling back to [emoji] when no bundled
+/// asset is available.
 ///
 /// Backend category names are an open-ended, uncurated stream (see #2547), so a
-/// name can arrive with no matching `assets/categories/<name>.svg`. flutter_svg
-/// would otherwise throw "Unable to load asset"; here the missing asset routes
-/// to [SvgPicture]'s `errorBuilder` and degrades to the category emoji instead.
+/// name can arrive with no matching `assets/categories/<name>.svg`. Callers pass
+/// a `null` [assetPath] for those (see `CategoryVisuals.forCategory`), and the
+/// emoji is rendered directly without ever asking flutter_svg to load a missing
+/// file — a missing-asset load reports a non-fatal to the zone even though
+/// `errorBuilder` handles the visual (#6116). The `errorBuilder` stays as a
+/// belt-and-suspenders guard for a genuinely corrupt bundled asset.
 class CategoryGlyph extends StatelessWidget {
   const CategoryGlyph({
     required this.assetPath,
@@ -20,10 +23,11 @@ class CategoryGlyph extends StatelessWidget {
     super.key,
   });
 
-  /// The bundled SVG asset path, e.g. `assets/categories/music.svg`.
-  final String assetPath;
+  /// The bundled SVG asset path, e.g. `assets/categories/music.svg`, or `null`
+  /// when no bundled asset exists for the category.
+  final String? assetPath;
 
-  /// The fallback glyph shown when [assetPath] is missing from the bundle.
+  /// The fallback glyph shown when [assetPath] is `null` or fails to load.
   final String emoji;
 
   /// Height for both the SVG and the emoji fallback.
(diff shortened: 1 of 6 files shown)


mobile/lib/widgets/categories/category_glyph.dart
```diff
@@ -34,8 +38,12 @@ class CategoryGlyph extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final path = assetPath;
+    if (path == null) {
+      return _EmojiGlyph(emoji: emoji, height: height, width: width);
```
realmeylisdev, line 43: nit (non-blocking): this null→emoji path now renders the mascot in the gallery header and tiles. The widget tests cover the behavior well; a golden would additionally lock the visual — but the same emoji already rendered here via `errorBuilder` on main, so it's pre-existing coverage. Fine to skip.

realmeylisdev (approved):
Traced it end to end — holds up.

The subtle part is right: the #4398 `errorBuilder` can't stop the non-fatal because `vector_graphics` discards the `whenComplete` future that re-raises the missing-asset error into the zone (reproduced it on Dart 3.12 — the discarded future fires the zone handler once; drop it and it's zero). So never handing `SvgPicture` a missing path is the only real fix. Gating every path (featured + derived) through the bundled set is the right call over adding one SVG, and the drift test keeps the set honest.

Checked: set matches all 101 on-disk SVGs exactly, `assets/categories/` is registered as a directory so on-disk == bundled, and all three `forCategory` callers are null-safe (the metadata chip never reads `assetPath`). Tests + analyze green locally, CI green. LGTM.

mobile/lib/widgets/categories/category_glyph.dart
hm21: Agreed, skipping the golden. That same `_EmojiGlyph` already rendered via `errorBuilder` on main, so the null path adds no new visual to lock — just a new trigger, which the widget tests cover. Thanks for the thorough trace!

Issue #5938, opened 2026-07-07
feat(notifications): add notifications for tags, mentions, and inspired-by references


hm21:
### What would you like?

Add notification alerts for tags, "Inspired by" references, and @mentions in comments.

### How would this be useful for you?

It would make it much easier to notice relevant activity and respond quickly without manually checking updates.

### When would you use this?

Whenever another user tags me, mentions me in a comment, or references my work through an "Inspired by" link.

### Current status

This is now a cross-repo tracking issue rather than a single mobile implementation ticket.

Last updated: 2026-07-22

| Area | Status | Tracking | Notes |
| --- | --- | --- | --- |
| Mobile publish-side p-tags for Inspired by | Done | #6128 | Mobile now emits notifiable `p` tags for Inspired by references. |
| Push delivery for kind 34236 video references | Done | divinevideo/divine-push-service#28 | Push service now handles video-sourced mention delivery. |
| In-app notification generation | Done | divinevideo/divine-funnelcake#654 / divinevideo/divine-funnelcake#670 | Funnelcake now sources kind 34236 and dedupes replaceable video edits. |
| Mobile receive/render/routing polish | In review | #6299 | Mobile PR renders video-sourced mention rows with video context and routes them without opening empty comments. |
| Dedicated Inspired by notification copy/type | Follow-up needed | TBD | Minimal path can render as `mentioned you`; a dedicated type needs backend + mobile + l10n work. |
| Comment @mention copy correctness | Follow-up needed | TBD | Comment mentions can currently be labeled like ordinary comments unless backend distinguishes p-tag mention recipients. |

### Remaining work for this tracker

- Keep this issue open until the mobile receive-side PR lands.
- Land #6299 for:
  - Comments tab filter inclusion for comment mention-style rows.
  - Video-sourced mention rendering with video thumbnail/context.
  - Tap routing that opens video-sourced mentions to the video without incorrectly auto-opening comments.
  - Regression tests for mapping, filtering, and routing.
- File follow-up issues for:
  - Dedicated Inspired by notification type/copy across backend, mobile enum handling, icons, and l10n.
  - Correct labeling for @mentions inside comments.

### Anything else?

Requested on Discord [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1524125702386094252).

Issue #4200, opened 2026-05-09
epic(notifications): stabilize in-app notifications, unread state, and push delivery


NotThatKindOfDrLiz:
## Summary
Notifications remain one of the highest-risk user-facing areas in the app. Active issues exist across unread-state correctness, row rendering, actor identity, tap-target routing, push registration reliability, and residual dual-system debt.

This issue is the coordination source of truth for notifications work this week.

## Assignee
`@Chardot`

## Weekly Priority
`P1`

## Weekly Priority Meaning
- `P1` means top priority for the week.
- `P2` means complete this week after P1 priorities are finished.

## Execution Order
1. unread/data correctness
2. tap-target/navigation correctness
3. row rendering / actor clarity
4. push registration and cold-start delivery
5. cleanup of dual systems / residual debt

## Subgroup
- `@Chardot` — DRI
- `@dcadenas` — push / auth-readiness / cold-start delivery
- `@realmeylisdev` — unread-state and notification data consistency
- `@untreu2` — routing / service-path bugs

## Coordination
The listed subgroup must review scope, sequencing, and ownership before implementation starts. Coordination may be synchronous or asynchronous. Do not wait for central coordination if the subgroup can resolve the question directly. Escalate only when blocked or when a product, design, or architecture decision is genuinely needed.

## Agent Execution Notes
Agents working from this issue must:
- treat this parent issue as the coordination source of truth for notifications
- read linked child issues before starting implementation
- follow the execution order unless explicitly told otherwise
- avoid narrow symptom-only fixes in known weak paths
- update linked issues if dependencies or scope change

## Engineering Standard
This work must not create new technical debt.

The expectation is:
- no temporary hacks, fragile stopgaps, or TODO-style deferrals without an explicit linked follow-up issue
- eliminate as much adjacent existing debt as possible while touching this area
- prefer fixes that move the codebase toward repository standards: explicit state-driven behavior, clear ownership boundaries, repository-appropriate abstractions, and test-backed changes
- if a short-term workaround is truly unavoidable, it must be called out explicitly in the issue and PR, with the follow-up issue linked before merge

Where relevant, use the audit issues created from PR `#3530` by `@omartinma` as the source of truth for debt repayment and refactor expectations in this area.

## Relevant Debt / Refactor Issues
- `#3596` Delete dual notification system
- `#3567` Fold legacy NotificationModel list-item into BLoC variant
- `#3590` Inconsistent error handling in notification methods
[...]

PR #6132, opened 2026-07-16
feat(video): emit an inspired-by p-tag so the referenced creator is notifiable

### LINKED ISSUE #6128: feat(video): emit a p-tag for "Inspired by" attribution so the creator is notifiable
## Context

Part of **#5938** (notifications for tags, mentions, and "Inspired by" references) under epic **#4200**. This issue covers **only the mobile publish-side change**: making an "Inspired by" reference *notifiable* by emitting a `p` tag for the referenced creator.

This is one third of a cross-repo fix — all three are required end-to-end for the recipient to actually get a notification:

- **This issue** — mobile emits the `p` tag (NIP-correct).
- **divinevideo/divine-funnelcake#654** — in-app notification generation (add kind 34236 as a notification source).
- **divinevideo/divine-push-service#28** — FCM push delivery (subscribe kind 34236).

## Problem

When a user publishes a video with an "Inspired by" reference (to a person or a specific video), the referenced creator never receives a notification. A real support report: the recipient (`a7e9d5d90b17fad6e4526ce58f0af997b8c14826ecbdbc2c3955e6e258f4000c`) was "inspired-by"-tagged by another user and got nothing.

**Root cause (publish side).** The inspired-by attribution is written only as:

1. an `a` tag `['a', '34236:<creatorPubkey>:<d>', <relay>, 'mention']` — and **only** when a *specific video* was chosen (`inspiredByAddressableId != null`); and
2. a NIP-27 content mention appended to the note body: `\n\nInspired by nostr:<npub>`.

**No `p` tag** pointing at the referenced creator is ever emitted. Per NIP-27, a `nostr:npub` content mention *without* a matching `p` tag is precisely how you mention someone **without** notifying them; the `p` tag is the canonical, relay-indexed notify mechanism (NIP-01 / NIP-10 / NIP-18). Every downstream notifier (funnelcake in-app, push-service FCM) derives recipients from `p`/`e` tags only — no `p` tag ⇒ no notification.

### Evidence (this repo, `origin/main`)

- `mobile/lib/services/video_event_publisher.dart:1273-1289` — the only `p` tags come from `buildCollaboratorPTags(collaboratorPubkeys)` and `_buildMentionPTags(mentionedPubkeys)`; the inspired-by block immediately after adds only the `a` tag.
- `mobile/lib/services/video_event_publisher.dart:1501-1505` — the `Inspired by nostr:<npub>` content append (built *after* mention resolution, never re-scanned for p-tags).
- `mobile/lib/services/video_publish/video_publish_service.dart:311-329` — `mentionedPubkeys` is resolved only from `@handle`s in the caption/overlays (skipped entirely unless the text contains `@`); the inspired-by npub is never part of that text.
- `mobile/lib/services/video_metadata_update_service.dart:313-327` — edit path: collaborator p-tags + an `a` tag with marker `inspired-by` (non-reply only); still no p-tag for the creator.
- `mobile/lib/providers/video_editor_provider.dart:440-443` — `setInspiredByPerson(npub)` stores only `inspiredByNpub` and clears `inspiredByVideo`; it does not add to `collaboratorPubkeys` or any mention list.

### Empirical confirmation (production relay `wss://relay.divine.video`)

The real reproducing event:

- id `42d1610431a7d3a1afc8420b21e9e98896977feca2031b927b00b30c79e9d684` (kind 34236)
- author `076c979382b90f5d3a2b21f95e1ee86b6033f14c92e79b7fad3fe1f1073f4886`
- content: `Whistler woods\n\nInspired by nostr:npub15l5atkgtzladdezjdnjc7zhej7uvzjpxaj7mctpe2hnwyk85qqxqjuecgm`
- **0 p-tags, 0 a-tags** — a *person*-inspired publish (`inspiredByAddressableId == null`, so even the `a` tag was skipped), content mention only.

For contrast, 8 kind-34236 events on the relay **do** `p`-tag that same recipient — every one a `['p', <hex>, relay, 'collaborator']` on a collaboration video. So the `p`-tag path is used elsewhere; inspired-by is uniquely the one that emits no structured recipient tag.

## Proposed change

When publishing or editing a video that carries an inspired-by reference, also emit a `p` tag for the referenced creator:

- Resolve the creator hex pubkey from `inspiredByVideo?.creatorPubkey` (parsed from the `a` coordinate) **or** by decoding `inspiredByNpub`.
- Emit `['p', <hex>, <relayHint>, 'inspired-by']`, deduped against existing collaborator/mention `p` tags. (Marker amended from `'mention'` after investigation: the edit flow cannot strip `'mention'` p-tags without destroying caption @-mentions, so attribution would go stale after clearing or collaborator promote→demote; a distinct marker lets the edit path own the tag's lifecycle. Backends fan out on p-tags marker-blind, so notification delivery is unaffected.)
- Apply to **both** the create path (`video_event_publisher.dart`) and the edit path (`video_metadata_update_service.dart`).
- Keep the existing `a` tag and content mention as-is — they drive display attribution.

This is the NIP-correct change (NIP-18/NIP-27: a content mention SHOULD carry a matching `p` tag when the person should be notified).

## Dependency

This mobile change alone does **not** deliver a notification — the backend must also source kind 34236 (divinevideo/divine-funnelcake#654) and subscribe it for push (divinevideo/divine-push-service#28). All three land together for the "surface as a `mention`" outcome tracked under #5938.

## Acceptance criteria

- [ ] Publishing with an inspired-by **person** reference produces an event containing `['p', <creatorHex>, ..., 'inspired-by']`.
- [ ] Same for an inspired-by **specific video** reference (creator pubkey derived from the `a` coordinate).
- [ ] Edit/republish preserves/re-emits the p-tag.
- [ ] The p-tag is deduped (no duplicate when the creator is also a collaborator or an `@mention`); never self-p-tags the publisher.
- [ ] Unit tests in `mobile/test/services/video_event_publisher_test.dart` (+ the update-service test) asserting the p-tag is present for both the person and video inspired-by cases, and absent when there is no inspired-by reference.

realmeylisdev:
## Description

"Inspired by" attribution on kind-34236 videos was display-only: the referenced creator was never `p`-tagged, so no notifier could target them — per NIP-27, a content mention without a matching `p` tag deliberately does **not** notify. Empirically confirmed on the production relay (event `42d1610431a7d3a1afc8420b21e9e98896977feca2031b927b00b30c79e9d684`: content attribution, 0 p-tags).

### Changes

- **`utils/inspired_by_tags.dart` (new)** — shared builder: `['p', <creatorHex>, <relay>, 'inspired-by']`. Marker is distinct from `'mention'` (so the edit flow can own the tag without touching caption-mention p-tags — stripping those is a #6019-class loss) and from `'collaborator'` (so no consumer renders the creator as a collaborator). Resolution reuses `InspiredByInfo.creatorPubkey` + npub decode, validates hex, self-skips (reusing your own sound auto-populates yourself), and dedups against every existing `p` tag (collaborators, caption mentions, reply threading) to keep the single-p-tag-per-pubkey invariant.
- **`video_event_publisher.dart`** — create path emits the p-tag(s) after the collaborator/mention p-tags; handles the person-npub and video-coordinate sources co-occurring (audio-reuse auto-populate bypasses the editor's mutual exclusion, so two distinct creators are possible). **Not emitted on replies**: the model hides inspired-by attribution on replies (`hasInspiredBy` is `!isVideoReply`-gated) and the edit flow cannot own the tag there, so it would notify for invisible, un-clearable attribution.
- **`video_metadata_update_service.dart`** — the edit flow owns the tag's lifecycle exactly like the inspired-by a-tag: `_isEditRebuiltTag` strips `'inspired-by'`-marker p-tags on non-replies and re-emits from editor state. Clearing attribution removes the tag, changing the creator re-targets it, collaborator promotion→demotion re-mints it, and **editing any pre-fix video backfills the missing p-tag**. Caption-mention and reply p-tags are never stripped.
- **`models/video_event.dart`** — anchors the `inspiredByNpub` content parse to the trailing `Inspired by nostr:npub…` line, so an unrelated inline `nostr:npub` mention earlier in a caption is never mis-attributed (previously first-match-anywhere; the edit path re-derives the person p-tag from this value, making the fix load-bearing).
- **`video_editor_provider.dart`** — the edit-form strip regex now also matches whole-content attribution (empty-caption publishes trim the leading newlines), fixing attribution-line duplication on every edit of such videos.

### Backend dependency

The tag is **inert until the backends source kind 34236**: divinevideo/divine-funnelcake#654 (in-app) and divinevideo/divine-push-service#28 (FCM push) — both fan out on p-tags marker-blind, verified. Part of #5938 / epic #4200.

### Known accepted edge

If a foreign client rewrites content so the attribution line is no longer trailing, a subsequent mobile edit drops the p-tag along with the (equally undetected) display attribution — the tag deliberately follows the content-derived attribution determination, and re-picking the person restores it.

**Related Issue:** Closes #6128

## Out of Scope

- Backend sourcing of kind 34236 (funnelcake#654, push-service#28).
- Unifying the pre-existing create/edit a-tag divergences (marker `mention` vs `inspired-by`, relay `?? 'wss://relay.divine.video'` vs `?? ''`) — deliberately untouched; the shipped a-tag shapes are parsed by web/funnelcake.
- A dedicated `inspired_by` notification type (minimal outcome surfaces as a generic mention; #5938 umbrella).

## Verification

- `flutter test` on all affected suites — 178 passing: new `test/utils/inspired_by_tags_test.dart` (13), new `test/services/video_event_publisher_inspired_by_tags_test.dart` (10, incl. self-skip, co-occurrence, reply no-emit, malformed-coordinate), `video_metadata_update_service_test.dart` (+11 lifecycle: exactly-once, clear-strips, re-target, promote/demote re-mint, pre-fix backfill, reply preservation, caption-mention safety, self-skip, person-npub), `video_editor_provider_test.dart` (+2 strip cases), plus the pre-existing publisher/collaborator/reply suites unchanged and green.
- `mobile/packages/models`: full suite 797 passing (3 new anchored-regex tests).
- Line coverage on all new/changed lines: 100% (new util 21/21; both service blocks fully covered).
- Multi-agent adversarial review at high effort: 3 confirmed correctness findings fixed in this diff (reply-gate asymmetry, whole-content strip, invisible reply attribution), 4 cleanups applied, 1 plausible edge documented above.

## Type of Change

### DIFF
--- a/mobile/packages/models/lib/src/video_event.dart
+++ b/mobile/packages/models/lib/src/video_event.dart
@@ -621,9 +621,12 @@ class VideoEvent {
       rawTags[tagName] = tagValue;
     }
 
-    // Scan content for NIP-27 nostr:npub1... references (Inspired By person)
+    // Resolve the Inspired By person from the trailing NIP-27 attribution
+    // line the publisher appends ("Inspired by nostr:npub1..."). Anchored to
+    // the end of content so an unrelated inline nostr:npub mention earlier in
+    // the caption is never mistaken for attribution.
     String? inspiredByNpub;
-    final npubPattern = RegExp('nostr:(npub1[a-z0-9]+)');
+    final npubPattern = RegExp(r'Inspired by nostr:(npub1[a-z0-9]+)\s*$');
     final npubMatch = npubPattern.firstMatch(event.content);
     if (npubMatch != null) {
       inspiredByNpub = npubMatch.group(1);
--- a/mobile/packages/models/lib/src/video_event.dart
+++ b/mobile/packages/models/lib/src/video_event.dart
@@ -623,10 +623,14 @@ class VideoEvent {
 
     // Resolve the Inspired By person from the trailing NIP-27 attribution
     // line the publisher appends ("Inspired by nostr:npub1..."). Anchored to
-    // the end of content so an unrelated inline nostr:npub mention earlier in
-    // the caption is never mistaken for attribution.
+    // the end of content and to a preceding blank line (or start of content,
+    // for empty-caption publishes) so a prose nostr:npub mention is never
+    // mistaken for attribution. Keep in sync with the strip regex in
+    // video_editor_provider.dart.
     String? inspiredByNpub;
-    final npubPattern = RegExp(r'Inspired by nostr:(npub1[a-z0-9]+)\s*$');
+    final npubPattern = RegExp(
+      r'(?:^|\n\n)Inspired by nostr:(npub1[a-z0-9]+)\s*$',
+    );
     final npubMatch = npubPattern.firstMatch(event.content);
     if (npubMatch != null) {
       inspiredByNpub = npubMatch.group(1);
--- a/mobile/lib/providers/video_editor_provider.dart
+++ b/mobile/lib/providers/video_editor_provider.dart
@@ -663,10 +663,12 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
   void initFromPublishedVideo(VideoEvent video) {
     _suppressAutosave = true;
 
-    // Strip any appended NIP-27 inspired-by line before displaying
-    // the description in the edit form.
+    // Strip any appended NIP-27 inspired-by line before displaying the
+    // description in the edit form. Empty-caption publishes trim the leading
+    // newlines, so the attribution line can also be the entire content —
+    // mirror the parse regex in VideoEvent.fromNostrEvent.
     var content = video.content;
-    final npubPattern = RegExp(r'\n\nInspired by nostr:npub1[a-z0-9]+$');
+    final npubPattern = RegExp(r'(^|\n\n)Inspired by nostr:npub1[a-z0-9]+\s*$');
     content = content.replaceFirst(npubPattern, '');
 
     state = VideoEditorProviderState(
--- a/mobile/lib/services/video_event_publisher.dart
+++ b/mobile/lib/services/video_event_publisher.dart
@@ -33,6 +33,7 @@ import 'package:openvine/services/upload_manager.dart';
 import 'package:openvine/services/video_event_service.dart';
 import 'package:openvine/services/video_thumbnail_service.dart';
 import 'package:openvine/utils/collaborator_tags.dart';
+import 'package:openvine/utils/inspired_by_tags.dart';
 import 'package:openvine/utils/log_tag_sanitizer.dart';
 import 'package:openvine/utils/proofmode_publishing_helpers.dart';
 import 'package:profile_repository/profile_repository.dart';
@@ -1286,6 +1287,25 @@ class VideoEventPublisher {
           inspiredByRelayUrl ?? 'wss://relay.divine.video',
           'mention',
         ]);
+      }
+
+      // p-tag the inspired-by creator(s) so they are notifiable. Added after
+      // the collaborator/mention p-tags so those win dedup and caption
+      // @token resolution keeps matching caption mentions first. Reply
+      // videos never carry inspired-by p-tags: the model hides inspired-by
+      // attribution on replies (VideoEvent.hasInspiredBy) and the edit flow
+      // cannot own the tag there, so emitting one would notify a creator the
+      // video never visibly credits and could never un-credit.
+      if (replyContext == null) {
+        tags.addAll(
+          buildInspiredByPTags(
+            existingTags: tags,
+            addressableId: inspiredByAddressableId,
+            npub: inspiredByNpub,
+            relayHint: inspiredByRelayUrl,
+            selfPubkey: _authService?.currentPublicKeyHex,
+          ),
+        );
       }
 
       var selectedAudioReferenceId = selectedAudioEventId;
--- a/mobile/lib/services/video_metadata_update_service.dart
+++ b/mobile/lib/services/video_metadata_update_service.dart
@@ -13,6 +13,7 @@ import 'package:openvine/services/collaborator_invite_service.dart';
 import 'package:openvine/services/personal_event_cache_service.dart';
 import 'package:openvine/services/video_event_service.dart';
 import 'package:openvine/utils/collaborator_tags.dart';
+import 'package:openvine/utils/inspired_by_tags.dart';
 import 'package:unified_logger/unified_logger.dart';
 
 /// Result returned by [VideoMetadataUpdateService.updateVideo].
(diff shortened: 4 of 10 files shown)


dcadenas (approved):
✅ inspired-by tags notify referenced creators

mobile/packages/models/lib/src/video_event.dart
```diff
@@ -621,9 +621,12 @@ class VideoEvent {
...
     }
 
-    // Scan content for NIP-27 nostr:npub1... references (Inspired By person)
+    // Resolve the Inspired By person from the trailing NIP-27 attribution
+    // line the publisher appends ("Inspired by nostr:npub1..."). Anchored to
+    // the end of content so an unrelated inline nostr:npub mention earlier in
+    // the caption is never mistaken for attribution.
     String? inspiredByNpub;
-    final npubPattern = RegExp('nostr:(npub1[a-z0-9]+)');
+    final npubPattern = RegExp(r'Inspired by nostr:(npub1[a-z0-9]+)\s*$');
```
hm21: This parse regex is end-anchored only, but the edit-form strip in `video_editor_provider.dart` also requires `(^|\n\n)` before the phrase — so its "mirror the parse regex" comment is backwards, the strip is the stricter one. They disagree when the attribution follows a space or single `\n` rather than a blank line. Example: a video you own whose caption ends `my remix. Inspired by nostr:npub1creator`. Parse sets `inspiredByNpub`, but the strip leaves the line in `description`; on edit-save the service re-appends it (duplicated line) **and**, because `!isVideoReply`, emits an `inspired-by` p-tag — notifying a creator that was only mentioned in prose, never picked. Divine's own create/edit always use the `\n\n`/whole-content shape both regexes agree on, so this only bites foreign- or hand-formatted content — low likelihood, but it's a spurious third-party notification, not just cosmetic. Adding `(^|\n\n)` here (as a non-capturing `(?:…)` so the npub stays group 1) makes parse, strip, and append symmetric.

hm21 (approved):
Solid. The p-tag emission, dedup against existing p-tags, self-skip, and the non-reply gate are all correct across both the create and edit paths, and the lifecycle tests genuinely exercise clear / re-target / promote-demote / pre-fix backfill / reply-preserve. One correctness nit inline.

mobile/packages/models/lib/src/video_event.dart
realmeylisdev: Fixed in af9ef2b79c — the parse regex now requires `(?:^|\n\n)` before the phrase (npub stays group 1), matching the strip regex, with regression tests for the space- and single-`\n`-prefixed shapes.

PR #6134, opened 2026-07-16
fix(video_editor): instant trim-based clip split + guard teardown races

### LINKED ISSUE #6135: fix(video_editor): instant trim-based clip split + guard teardown races
Auto-created from PR #6134.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6134
**Branch:** `fix/video-editor-instant-split`

hm21:
Closes #6135

## Description

Splitting a clip in the video editor no longer re-encodes. A split is now a **pure trim-based metadata cut**: both halves keep the **same source file** with adjacent trim windows, so it is **instant** instead of a ~15–20 s native re-encode.

This started from a report that the **Split ("trim") control becomes unavailable / greys out for ~30 s after a couple of cuts** on a long clip. The ~30 s turned out to be the native re-encode (`pro_video_editor` `splitVideo`), whose duration is proportional to clip length. Investigating it surfaced two additional editor-teardown races, fixed here as well.

Three clearly separate changes:

**1. Trim-based split (no re-encode) — the headline**
- `splitClip` now produces two clips that share the source file: start half `[0, split]` (its `duration` caps the visible end at the split), end half `[split, end]`.
- New `DivineVideoClip.minTrimStart` (a source floor, serialized as `minTrimStartMs`) pins the end half's left trim handle at the split so it can't be dragged back into the start half's frames — the separate re-encoded file used to enforce that implicitly.
- Safe because the export already clips each clip to its `[trimStart, trimEnd]` window (`renderVideo._normalizeClipsToAspectRatio`) and `FileCleanupService` only deletes a file no clip/draft still references.
- Removes the now-dead native `splitNativeVideoToFile` primitive and its tests.

**2. Guard the final-clip callback against a disposed editor**
A long render can resolve after the user backs out of the editor. `onFinalClipInvalidated` then touched the disposed screen's Riverpod `ref` and threw `Using "ref" when a widget … has been unmounted`, and the split was lost. It now bails when the screen is unmounted.

**3. Don't let a stale editor re-sync clobber an in-flight split**
When a split was tapped while another was still rendering, the canvas's editor→bloc clip re-sync could carry the previous (stale) snapshot and overwrite the just-applied split, which then silently vanished. The canvas re-sync and `_onInitialized` now skip while `isSplitting`.

## Out of Scope

- On-device verification (see below) — I can't run it here.
- Any change to `pro_video_editor` itself (the split primitive is simply no longer used).

## Verification

- Targeted tests: `video_editor_split_service_test` (trim-based, source floor, re-split, transition), `divine_video_clip_test` (`minTrimStart` JSON round-trip + copyWith), `clip_editor_bloc_test` (trim clamp + stale-re-sync guard), `video_editor_render_service_test`.

## Type of Change

- [x] 🧹 Code refactor

**Related Issue:**

### DIFF
--- a/mobile/lib/blocs/video_editor/clip_editor/clip_editor_bloc.dart
+++ b/mobile/lib/blocs/video_editor/clip_editor/clip_editor_bloc.dart
@@ -169,6 +169,23 @@ class ClipEditorBloc extends Bloc<ClipEditorEvent, ClipEditorState> {
     ClipEditorInitialized event,
     Emitter<ClipEditorState> emit,
   ) {
+    // While a split is in flight (including splits still queued behind it),
+    // this bloc owns the clip list optimistically — the split has already
+    // inserted its result, but the editor-history round-trip that drives the
+    // canvas re-sync still reflects the pre-split state. Applying that stale
+    // snapshot here wipes the just-applied (queued) split, which then never
+    // appears. Defer: the post-split reconcile re-syncs once isSplitting
+    // clears. The canvas guards its dispatch on the same condition; this is
+    // the matching guard for any other re-init source.
+    if (state.isSplitting) {
+      Log.debug(
+        '📋 Ignoring clip re-init during in-flight split '
+        '(${event.clips.length} clip(s))',
+        name: 'ClipEditorBloc',
+        category: LogCategory.video,
+      );
+      return;
+    }
     Log.debug(
       '📋 Initialized with ${event.clips.length} clip(s)',
       name: 'ClipEditorBloc',
@@ -658,6 +675,12 @@ class ClipEditorBloc extends Bloc<ClipEditorEvent, ClipEditorState> {
         },
       );
 
+      // onFinalClipInvalidated reaches into the editor screen that owns this
+      // bloc. A long split (a full clip re-encode, ~15-20s) can land after the
+      // user backs out — bloc.close() lets this in-flight handler run to
+      // completion, so `emit.isDone` is still false here. The screen callback
+      // itself must guard on `mounted`; the bloc can't detect the torn-down
+      // screen from this side.
       onFinalClipInvalidated.call();
 
       Log.info(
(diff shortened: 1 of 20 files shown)


rabble (approved):
**Reviewed with full checkout + verification.** ✅ Approve, with an on-device checklist to clear before merge.

Solid bugfix. The headline change — trim-based split instead of native re-encode — is architecturally sound, state discipline is clean, and every guard is backed by a test that fails when the guard is removed.

On the teardown-race guards (looked hard at these):
- **`mounted` check in `video_editor_screen.dart:147` is a true fix.** `onFinalClipInvalidated` reaches into the screen's Riverpod `ref` and also fires after long reverse/transform renders (~15-20s) that can resolve after the user backs out. Guarding the `ref.read` at the `State` boundary with `mounted` is the correct place to fix "Using ref when unmounted".
- **`isSplitting` guards are a scoped, tested mitigation over a dual-ownership smell** (pro_video_editor's history re-syncs the clip list back into `ClipEditorBloc`). The clean fix is restructuring that ownership — out of scope and riskier. The guard is correct for the identified race and directly exercised by the "ignores ClipEditorInitialized while a split is in flight" test.

Correctness highlights (all tested): split geometry caps both halves via `duration`/`minTrimStart`; `budgetDuration = duration - minTrimStart` avoids double-counting the shared source region; `minTrimStart` JSON round-trips + legacy-defaults to zero; reverse now bakes the visible window and clears the floor so a reversed split-half can't expose its sibling's frames; no dangling refs to the removed native-split path.

Verification (local):
- `flutter analyze lib/blocs/video_editor lib/services/video_editor test/...` → No issues found!
- `flutter test test/blocs/video_editor/clip_editor test/services/video_editor test/models/divine_video_clip_test.dart test/widgets/video_editor` → 961 tests passed.

Nits (non-blocking): (1) reversing a split end-half with a `trimEnd` silently shrinks `budgetDuration` — worth a one-line rationale comment; (2) `video_editor_canvas._onlyThumbnailPathsChanged` hand-maintains a parallel field list to `DivineVideoClip` — future composition-relevant fields must be added there or a real edit gets misclassified as thumbnail-only.

**Before merge:** the PR's own device checklist (export of a split, draft round-trip, delete-one-half file survival, floor drag) can't be exercised in unit tests — please confirm on-device.

Issue #6127, opened 2026-07-16
feat(account): add reversible 28-day account deletion flow


dcadenas:
## Summary

Add a reversible account-deletion flow that hides the account from Divine immediately, preserves it for a fixed 28-day recovery period, and permanently deletes it only after that deadline.

Users must also retain a separate **delete permanently now** option that sends the existing standard NIP-62 vanish request without a recovery period.

## Dependency

Blocked by [divine-funnelcake#651 — reversible scheduled account deletion lifecycle](https://github.com/divinevideo/divine-funnelcake/issues/651).

This is adjacent to, but not a duplicate of, [#3102 — per-video media deletion](https://github.com/divinevideo/divine-mobile/issues/3102). This issue covers deletion of an entire account.

Interacts with **#6126** (opt-in username revoke+burn on deletion). Burn is irreversible, so the reversible path here must defer any name-burn to the terminal deletion (pre-signed + gift-wrapped with the future-dated kind-62, or server-side) — it must not fire at prepare-time, or a user who cancels within the recovery window still loses their handle permanently. #6126 ships independently on the immediate-deletion path; this issue owns the reversible-path burn wiring.

## User choices

Present two clearly distinguished choices:

1. **Delete with a 28-day recovery period**
   - Divine hides the account and its content immediately.
   - The user can recover the account until the displayed deadline.
   - After the deadline, Divine permanently deletes its copies and publishes the signed NIP-62 request to the configured relay set.
2. **Delete permanently now**
   - Preserve the current standard NIP-62 behavior.
   - Make the lack of a recovery period explicit before signing.

Use a full-screen flow rather than adding another confirmation dialog. Always make it clear which account is being changed, especially for users with multiple accounts.

## Scheduled deletion flow

1. Authenticate a prepare request to Funnelcake with NIP-98.
2. Use the server-provided deadline to sign a future-dated kind 62 event scoped to ALL_RELAYS.
3. Put that signed event and the validated relay-delivery instructions inside a NIP-59 gift wrap addressed to Divine's official service pubkey.
4. Submit the opaque wrapped payload to Funnelcake using NIP-98.
5. Report success only after Funnelcake confirms that the deletion is scheduled and the account is hidden.
6. Sign out and remove appropriate local application data.

The scheduled path must **not** immediately call Keycast account deletion. The Keycast identity remains available during the recovery period so the user can reinstall, sign in, and cancel the pending deletion.

## Recovery flow

[...]

Issue #6126, opened 2026-07-15
feat(account): release the Divine name as part of account deletion


mbradley:
> **Rescoped 2026-08-26 by team decision, from opt-in to mandatory.** The original issue (mbradley, 2026-07-15) specified an opt-in toggle, default off, so a departing user could keep their `@divine.video` handle for use elsewhere. The account-deletion release scope now removes the name as part of deletion rather than offering it. The original body is preserved at the bottom, and its verified-behaviour findings, ordering constraint and endpoint design all still hold — only the trigger changed. There is an open question about the reversal, recorded below.

## Prerequisite: the coordinator has no Name Server configured

**This cannot ship until production configures it, and the consequence of shipping first is worse than not shipping at all.**

`AccountDeletionConfig::from_env()` in funnelcake's `crates/api/src/account_deletion.rs` builds the name-server leg from **two** variables, and needs both:

```rust
let name_server_url = non_empty(std::env::var("NAME_SERVER_URL").ok());
let name_server_token = non_empty(std::env::var("DELETION_COORDINATOR_TOKEN").ok());
if name_server_url.is_some() != name_server_token.is_some() {
    tracing::warn!("Username recovery is disabled because its URL and credential are incomplete");
}
```

Per #8134, `NAME_SERVER_URL` is staged off in production, staging and poc alike. With it unset, a username-bearing attempt answers `503 username_recovery_unavailable` and a username-free attempt succeeds against the same server.

Today that is survivable, because the release is opt-in and defaults off: a user who ticks the box and hits the 503 can untick it and delete. #8134 exists specifically to tell them so.

**Make the release mandatory in that same state and there is no box to untick — every deletion 503s.** So the config is not a nice-to-have alongside this issue; it gates it.

The ask is a URL *and* a credential, not one variable. Worth filing against production configuration the same way the Keycast leg was, and worth confirming it landed before this issue is picked up rather than after.

For contrast, the Keycast leg of the same config is already satisfied: `from_env()` returns `None` when `KEYCAST_URL` is absent, which would leave the coordinator routes unmounted — and they currently answer `401` / `405` rather than `404`, so both `KEYCAST_URL` and `KEYCAST_DELETION_SERVICE_TOKEN` are set in production. That is also the most likely explanation for the 404s disappearing on 2026-08-25.

## What changes

Deleting a Divine account releases the `@divine.video` name. No toggle, no default, no opt-in.

[...]

Issue #65, opened 2025-06-19
First video doesn't autoplay when app opens


rabble:
## Description
The first video in the feed doesn't autoplay when the app initially opens. However, if you scroll to the second video (which does autoplay) and then scroll back to the first video, it will autoplay correctly.

## Current Behavior
- App opens → First video is visible but doesn't start playing
- Scroll to second video → Second video autoplays correctly
- Scroll back to first video → First video now autoplays correctly

## Expected Behavior
- App opens → First video should immediately start playing

## Possible Causes
- Initial visibility detection not triggering for the first video
- Video player initialization timing issue
- PageView or visibility detector not firing on initial load
- Missing initial play command for index 0

## Files to Investigate
- `lib/screens/feed_screen.dart`
- `lib/widgets/video_feed_item.dart`
- Video player initialization logic
- Visibility detection implementation

## Priority
High - Affects user experience on app launch

rabble:
Fixed in recent commits - autoplay now works correctly with proper Chewie controller callbacks instead of arbitrary delays.

PR #6138, opened 2026-07-16
feat(account): opt-in @divine.video username burn on account deletion (#6126)

mbradley:
Mobile half of the opt-in "burn my @divine.video username when I delete my account" feature (#6126). Pairs with the nameserver endpoint in divinevideo/divine-name-server#65, which this depends on.

## Why
Deleting a Divine account never touched your username — it stayed active and kept resolving after everything else was gone, with only a manual admin path to give it up. This adds an opt-in, **default-off** choice to permanently burn the handle as part of deletion, for people who want a clean break. Anyone keeping their Nostr identity elsewhere just leaves it off.

## What
- **`releaseUsername()` / `getUsernameByPubkey()`** in `profile_repository` — a NIP-98-signed call to the new `/api/username/release`, plus an ownership lookup that gates the toggle.
- **Opt-in checkbox** in the delete-account confirmation dialog, shown only when the signed-in user owns an active `@divine.video` handle, naming the exact handle.
- **Burn-first hard-block** in `executeAccountDeletion`: when opted in, the handle is released before any destructive step; if the release fails, the deletion aborts with **nothing deleted** (retry, or uncheck to proceed). The signer window forces this order — NIP-98 needs a live signer, which only exists before Keycast deletion.
- New l10n keys (English; translation deferred via `_knownUntranslatedDebt`).

## Design
Spec and plan live in the branch: `docs/superpowers/specs/2026-07-16-username-burn-on-delete-design.md` and `docs/superpowers/plans/2026-07-16-username-burn-on-delete.md`.

## Interactions (documented, not blockers)
- **keycast#296** — on claimed accounts, deletion already fails at the Keycast step (pre-existing FK bug, not caused here). The burn succeeds first; full deletion completes once #296 lands.
- **#6127** (reversible deletion) — when that full-screen flow lands, the burn must be deferred to the terminal deletion moment. Cross-linked both directions.

## Copy
The toggle and failure copy are **drafts pending product/brand sign-off**. The handle is presented as a "username"; the failure copy makes the not-deleted outcome explicit.

## Tests
- `profile_repository`: `releaseUsername` (200/401/403/network/null-auth) + `getUsernameByPubkey` (found/not-found/non-200/network) — 100% package coverage.
- Dialog: toggle hidden/shown by ownership; checked/unchecked → correct `burnUsername` on confirm.
- Orchestration: opted-in burn failure aborts (`deleteAccount` never called) + shows error; burn success proceeds; not-opted-in never releases.
- `ownedDivineUsernameProvider`: resolves / gates on pubkey + repo readiness.

## Manual test plan
- **Own a `@divine.video` handle:** delete flow shows the toggle; leaving it off deletes normally; checking it burns the handle first (verify it stops resolving at `names.divine.video` and can't be re-registered), then deletes.
- **No handle:** no toggle; deletion unchanged.

## Deploy dependency
The toggle calls `/api/username/release`, which ships in divinevideo/divine-name-server#65. **Deploy #65 to production `names.divine.video` before a mobile build with this toggle ships.** If the endpoint isn't live, an opted-in burn fails and hard-blocks the deletion (the user can uncheck to proceed) — so it degrades safely rather than mis-deleting, but the release sequencing matters.


mobile/lib/widgets/delete_account_dialog.dart
```diff
@@ -295,9 +324,38 @@ Future<void> executeAccountDeletion({
...
+              error: true,
+            ),
+          );
+        }
+        return;
+      }
+    }
+
+    // Publish the NIP-62 deletion request (requires working signer).
     final result = await deletionService.deleteAccount(
```
dcadenas, line 468: Guard the post-burn abort paths in `executeAccountDeletion` so they disclose that the handle was permanently released while the account was not deleted.
After `UsernameReleaseSuccess` at lines 339-356, record that the burn committed and use that state in both the registered-user Keycast failure return around line 367 and the `deleteAccount` failure return around line 438.
Add the localized copy, error logging with the full handle, and widget coverage for opted-in and opted-out cases.
Keep the burn-before-deletion ordering and do not add a rollback.

dcadenas (changes requested):
The deletion flow can permanently burn a user's @divine.video handle and then report only that account deletion failed, leaving the user signed in without explaining the handle is gone.
Before merge, make both post-burn abort paths disclose the permanent release and add the matching localization, logging, and widget coverage while preserving the required burn-before-deletion ordering.

mobile/packages/profile_repository/lib/src/profile_repository.dart
```diff
@@ -992,6 +995,98 @@ class ProfileRepository {
...
+  /// authenticated request to `names.divine.video/api/username/release`.
+  ///
+  /// The server verifies the authenticated pubkey owns [name] as an active
+  /// username before burning it. Returns a [UsernameReleaseResult]. A `200`
+  /// (including the server's idempotent no-op when no active name is owned)
+  /// maps to [UsernameReleaseSuccess]: the post-condition "the handle is no
+  /// longer active" holds either way.
+  Future<UsernameReleaseResult> releaseUsername({required String name}) async {
+    final payload = jsonEncode({'name': name});
+    final authHeader = await _nostrClient.createNip98AuthHeader(
```
realmeylisdev: **should-fix:** `createNip98AuthHeader` is awaited outside the `try`, so a signer-layer throw (Keycast RPC error, or its 30s timeout) escapes as an exception instead of mapping to a result — and `executeAccountDeletion`'s `try` has no `catch`, so an opted-in burn failure on the Keycast path closes the progress dialog with **no error shown** (deletion still aborts correctly, but `deleteAccountBurnUsernameFailed` never appears). Since Keycast users are the main username-owning population, this is the feature's most likely failure mode. Wrapping this call in the `try` — mapping failures to `UsernameReleaseError` / `UsernameReleaseNetworkError` — keeps the method's no-throw sealed-result contract, which the hard-block relies on.

mobile/packages/profile_repository/lib/src/profile_repository.dart
```diff
@@ -992,6 +995,98 @@ class ProfileRepository {
...
+  /// A lookup failure (network, timeout, non-200, or unparseable body) also
+  /// returns `null`: callers treat "unknown" as "no owned name" and simply do
+  /// not offer the burn option.
+  Future<String?> getUsernameByPubkey({required String pubkeyHex}) async {
+    try {
+      final response = await _httpClient
+          .get(Uri.parse('$_usernameByPubkeyUrl/$pubkeyHex'))
+          .timeout(_nameServerHttpTimeout);
+      if (response.statusCode != 200) return null;
+      final data = jsonDecode(response.body) as Map<String, dynamic>;
```
realmeylisdev, line 1104: **should-fix:** if a 200 body is valid JSON but not an object, this cast throws `TypeError` — an `Error`, so it escapes the `on Exception` below, contradicting the doc's "unparseable body also returns null" (confirmed empirically). It then propagates through `ownedDivineUsernameProvider` into the unguarded `await` in `_handleDeleteAllContent`, leaving the Delete Account tile unresponsive. This file already uses `on Object catch` where decode casts can throw — same treatment here (and on the error-body parse at line 1038) closes it.

mobile/lib/screens/settings/nostr_settings_screen.dart
```diff
@@ -259,13 +260,24 @@ class _DeleteAccountTile extends StatelessWidget {
   ) async {
     final deletionService = ref.read(accountDeletionServiceProvider);
     final authService = ref.read(authServiceProvider);
+    final profileRepository = ref.read(profileRepositoryProvider);
+    // Resolve whether the user owns a burnable @divine.video handle so the
+    // dialog can offer the opt-in burn toggle.
+    final String? ownedUsername = profileRepository == null
+        ? null
+        : await ref.read(ownedDivineUsernameProvider.future);
```
realmeylisdev: **nit (UX):** this await can block the tap for up to the name-server timeout (10s) with no affordance, and repeated taps re-enter the handler; any error from the provider also escapes here (see the `TypeError` note in profile_repository). Consider showing the dialog immediately and revealing the toggle when the lookup resolves — or at minimum a brief loading affordance plus a try/catch that degrades to `null`.

mobile/lib/widgets/delete_account_dialog.dart
```diff
@@ -295,9 +324,38 @@ Future<void> executeAccountDeletion({
...
   final localDataDeletionFailedText =
       context.l10n.deleteAccountLocalDataDeletionFailed;
+  final burnUsernameFailedText = context.l10n.deleteAccountBurnUsernameFailed;
 
-  // Step 1: Execute NIP-62 deletion request (requires working signer)
   try {
+    // Burn-first hard-block: release the username before any destructive step,
+    // so a failed burn leaves everything intact. Needs a working signer, which
+    // exists before deleteKeycastAccount() below.
+    if (burnUsername && ownedUsername != null) {
```
realmeylisdev: **nit:** asymmetry — a null `profileRepository` hard-blocks an opted-in burn, but `ownedUsername == null` silently skips it and proceeds to delete. Unreachable through the current sole caller (verified), but folding it into the same hard-block plus a test would make the invariant self-defending against future callers.

mobile/lib/widgets/delete_account_dialog.dart
```diff
@@ -295,9 +324,38 @@ Future<void> executeAccountDeletion({
...
+      // A null repository here means we cannot honor the opted-in burn, so we
+      // must NOT proceed — treat it the same as a failed release (hard-block).
+      final releaseResult = await profileRepository?.releaseUsername(
+        name: ownedUsername,
+      );
+      if (releaseResult is! UsernameReleaseSuccess) {
+        Log.warning(
+          'Username release failed or unavailable; aborting account deletion',
```
realmeylisdev: **nit:** the typed `releaseResult` is dropped — 401, 403, network error, and null-repository all log the identical line. Interpolating the result into the message makes field triage possible.

mobile/packages/profile_repository/test/src/profile_repository_test.dart
```diff
@@ -4267,6 +4267,196 @@ void main() {
       });
     });
 
+    group('releaseUsername', () {
```
realmeylisdev, line 4270: **nit (tests):** these tests stub `post(any(), headers: any(...), body: any(...))` and never `verify()` the outgoing request, so nothing pins the URL or that the posted body is byte-identical to the string hashed into the NIP-98 `payload` tag — the crux of the server contract. One `verify` with captured args would lock it (same for the by-pubkey URL in the group at line 4415).

realmeylisdev (commented):
Verified the client↔server contract end-to-end against divine-name-server#65 (NIP-98 middleware, status mapping, by-pubkey response shape, CORS) and the burn-first/signer-window ordering; ran the touched suites locally, all green. The design is right — two should-fixes on the burn failure path inline (a signer throw aborts silently with no error shown; a `TypeError` can slip past `on Exception` and leave the Delete tile unresponsive), plus a few nits.

mobile/packages/profile_repository/lib/src/profile_repository.dart
```diff
@@ -992,6 +995,98 @@ class ProfileRepository {
...
+      if (data['found'] != true) return null;
+      return data['name'] as String?;
```
realmeylisdev: **should-fix:** this returns the server's display form (`name`), and the delete flow feeds it back to `/release`, which canonicalizes with today's IDNA rules and compares against the stored canonical. Vine-import rows derived their canonical with a *different* algorithm (NFD diacritic/punctuation strip), so e.g. display `Café` canonicalizes today to `xn--caf-dma` ≠ stored `cafe` → 403 → the opted-in burn hard-blocks deletion even though the stored name is burnable. The by-pubkey response already includes the round-trip-safe `canonical` field — use that for the release call (display stays right for the toggle label). Self-claimed names are unaffected since claim/release share one canonicalization.

docs/superpowers/plans/2026-07-16-username-burn-on-delete.md
```diff
@@ -0,0 +1,688 @@
...
+
+- **Two repos, two PRs, both target `main`, neither stacks.** Part A in a `divine-name-server` worktree off its `origin/main`; Part B in this `divine-mobile` worktree (`feat/username-burn-on-delete`).
+- **Burn = `burn:true`** on `revokeUsername` (status `burned`, `recyclable=0`). Never plain revoke.
+- **Burn-first ordering** in `executeAccountDeletion`: release runs before any destructive step; on failure abort with nothing destroyed.
+- **Never truncate Nostr IDs / pubkeys** anywhere (code, logs, tests).
+- **No error strings/exception objects in bloc/cubit state.** Typed results only.
+- **l10n:** any new `app_en.arb` key is mirrored to every `app_*.arb` (or added to `_knownUntranslatedDebt`); run `flutter test test/l10n/arb_consistency_test.dart`.
+- **Divine dark-mode only:** use `VineTheme` + `divine_ui` components, no raw `Colors.*`.
+- **Verify before push:** `dart format`, `flutter analyze lib test`, scoped tests (Part B); `npm test` / `vitest` (Part A).
+- Part B mobile toggle is inert until Part A is deployed to staging; unit tests use a mocked client.
```
realmeylisdev: **should-fix (doc):** "mobile toggle is inert until Part A is deployed" (repeated at line 687) is not what ships: `GET /by-pubkey` is a pre-existing prod endpoint (returns 200 today), so the toggle renders for every handle owner immediately — and with `/release` undeployed (404 on prod right now) an opted-in confirm hard-blocks the entire account deletion. The PR body's "deploy #65 first" section has it right; this line will mislead whoever sequences the release.

mobile/packages/profile_repository/lib/src/username_release_result.dart
```diff
@@ -0,0 +1,39 @@
+/// Sealed class representing the result of a username release (burn) attempt.
+sealed class UsernameReleaseResult {
+  /// Creates a username release result.
+  const UsernameReleaseResult();
+}
+
+/// The username was permanently burned (or was already not active).
```
realmeylisdev: **nit (doc):** "or was already not active" overstates the server's 200 no-op — `no_active_name` only means the *caller* owns no active name; the requested handle can still be active under another pubkey (revoke→reclaim race). Same overclaim in `releaseUsername`'s doc ("the handle is no longer active" post-condition). "The caller no longer holds an active name" is what the server actually guarantees.

mobile/test/widgets/delete_account_dialog_test.dart
```diff
@@ -151,6 +162,62 @@ void main() {
...
+    ) async {
+      await _showDialog(tester);
+      expect(find.byType(CheckboxListTile), findsNothing);
+    });
+
+    testWidgets('burn toggle is shown when a username is owned', (
+      tester,
+    ) async {
+      await _showDialog(tester, ownedUsername: 'alice');
+      expect(find.byType(CheckboxListTile), findsOneWidget);
```
realmeylisdev, line 195: **nit (tests):** the toggle tests only assert `find.byType(CheckboxListTile)` — the consent label naming the exact handle is unpinned, so swapping the interpolation at delete_account_dialog.dart:154 for the wrong variable keeps everything green. One extra line here pins it: `expect(find.text(lookupAppLocalizations(const Locale('en')).deleteAccountBurnUsernameToggle('@alice.divine.video')), findsOneWidget)` — the file already uses `lookupAppLocalizations` further down.

mobile/packages/profile_repository/lib/src/profile_repository.dart
```diff
@@ -992,6 +995,98 @@ class ProfileRepository {
...
+      final response = await _httpClient
+          .post(
+            Uri.parse(_usernameReleaseUrl),
+            headers: {
+              'Authorization': authHeader,
+              'Content-Type': 'application/json',
+            },
+            body: payload,
+          )
+          .timeout(_nameServerHttpTimeout);
```
realmeylisdev, line 1051: **observation (copy):** `.timeout` abandons but can't cancel the request, so a slower-than-10s-but-successful burn shows "Couldn't release your username. Your account was not deleted." after the burn actually committed server-side. A retry self-heals via the server's idempotent no-op, but a user who gives up keeps the account while the handle has silently stopped resolving. "Couldn't confirm the release" phrasing would stay honest either way — worth folding into the pending copy pass.

realmeylisdev (commented):
Round 2 — cross-repo pass over the vine-import data path: the opted-in burn can be structurally impossible for legacy vine-import names, and the hard-block then traps the deletion. Mobile-side mitigation plus doc/test notes inline; the server-side half is posted on divinevideo/divine-name-server#65.

mobile/lib/widgets/delete_account_dialog.dart
mbradley: Done in 920c50acf. executeAccountDeletion records that the burn committed after UsernameReleaseSuccess, and both post-burn abort paths (registered-user Keycast failure, and the deleteAccount failure) now disclose that the handle was permanently released while the account was not deleted, with the full handle logged and localized copy. Burn-before-deletion ordering is unchanged and there's no rollback. Widget coverage added for the opted-in and opted-out cases.

mobile/packages/profile_repository/lib/src/profile_repository.dart
mbradley: Fixed in 920c50acf. createNip98AuthHeader is inside the try now, so a signer throw (Keycast RPC error or timeout) maps to UsernameReleaseError instead of escaping. The method keeps its no-throw sealed-result contract, so the hard-block sees a failure result and surfaces deleteAccountBurnUsernameFailed rather than silently closing the dialog.

mobile/packages/profile_repository/lib/src/profile_repository.dart
mbradley: Fixed in 920c50acf. getUsernameByPubkey catches on Object now, so a 200 body that's valid JSON but not an object returns null instead of throwing. Same treatment on the error-body parse. The doc's unparseable-body-returns-null contract holds.

mobile/lib/screens/settings/nostr_settings_screen.dart
mbradley: Fixed in 7421e2fd6. The dialog opens immediately and the burn toggle reveals itself once the lookup resolves, so a slow name-server call no longer blocks the tap and a second tap can't stack a duplicate dialog. The lookup is resolved inside the dialog and guarded so a failed lookup degrades to no-toggle (bffc9f9ca) instead of escaping. The dialog also hands the displayed handle back on confirm, so a burn targets exactly what was shown.

mobile/lib/widgets/delete_account_dialog.dart
mbradley: Folded into the hard-block in 920c50acf. An opted-in burn with no resolvable handle (null repo or null ownedUsername) now aborts deletion the same way a failed release does, instead of silently proceeding. Covered by the "opted-in burn aborts when no username is owned" test.

mobile/lib/widgets/delete_account_dialog.dart
mbradley: Fixed in f4f6d973a. The abort log names the reason now: "no active handle to burn", "profile repository unavailable", or "release returned <ResultType>", so a null-precondition is distinguishable from an actual release failure (401/403/network) in triage.

mobile/packages/profile_repository/test/src/profile_repository_test.dart
mbradley: Pinned in 920c50acf. The releaseUsername tests capture the outgoing POST and assert the URL and that the body is byte-identical to jsonEncode({'name': ...}) (the string hashed into the NIP-98 payload), and the by-pubkey group asserts its GET URL. A path-only URL or re-serialized body fails now.

mobile/packages/profile_repository/lib/src/profile_repository.dart
mbradley: Fixed in 920c50acf. getUsernameByPubkey returns both the display name (toggle label) and the round-trip-safe canonical, and the delete flow sends canonical to /release. So a vine-import display like Café no longer re-canonicalizes to a mismatched punycode and 403s the burn; the release compares against the stored canonical.

docs/superpowers/plans/2026-07-16-username-burn-on-delete.md
mbradley: Fixed in 920c50acf. Since GET /by-pubkey is already in prod, the toggle renders for handle owners immediately, so /release (#65) must deploy before this ships or an opted-in confirm hard-blocks deletion. The "inert until deployed" line is gone.

mobile/packages/profile_repository/lib/src/username_release_result.dart
mbradley: Fixed in 920c50acf. Reworded to what the server guarantees: the 200 no-op means the caller no longer holds an active name, and the doc notes it does not imply the handle is inactive for everyone (it can be active under another pubkey). Same fix on releaseUsername's doc.

mobile/test/widgets/delete_account_dialog_test.dart
mbradley: Pinned in 920c50acf. The toggle test asserts the exact consent label naming the handle in display form (deleteAccountBurnUsernameToggle('@Alice.divine.video')), so swapping the interpolated variable fails the test.

mobile/packages/profile_repository/lib/src/profile_repository.dart
mbradley: Real limitation with no clean client-side fix: .timeout abandons but can't cancel the request, so a burn that commits just after 10s reads as a failure. What 920c50acf does is re-check ownership on the ambiguous path (network/timeout/5xx): the flow re-queries by-pubkey and only says "not deleted, try again" when the handle is verifiably still there, staying neutral when it can't tell, so we never tell someone nothing happened when it did. A retry also self-heals via the server's idempotent no-op. Exact wording is deferred to the product/brand copy pass; the current strings are honest placeholders. Marking resolved as a documented limitation unless you'd rather track it separately.

dcadenas (approved):
✅ opt-in username burn, clean handoff

PR #5473, opened 2026-06-22
fix(dm): count followed-but-unreplied unread chats in the Messages badge (#4976)

realmeylisdev:
## Description

The "Messages" unread badge (bottom-nav + inbox segmented toggle) counted only conversations where the current user had sent a message (`currentUserHasSent == true`), while the Messages **list** is follow-aware — it also shows unread chats from peers you **follow** but have **never replied to**, and it hides blocklisted peers. So the badge undercounted the visible unread dots (the report: 4 dots, badge 3). It was latent until #4973 backfilled the full conversation list after reinstall, which surfaced followed-but-unreplied conversations.

This makes the badge mirror the list exactly. `DmUnreadCountCubit` now combines the accepted + potential-request + following + blocklist streams and reuses the **same** static helpers the list uses (`DmRepository.classifyPotentialRequests`, `DmRepository.mergeAndSort`, `ContentBlocklistRepository.filterBlockedConversations`), counting unread over `accepted ∪ followed`, blocklist-filtered and unpaginated — so the count can't drift from the list. A pure-SQL count isn't possible because follow state is in-memory (`FollowRepository`), not a DB column.

The two badge surfaces are consolidated onto the single app-shell cubit; the redundant inbox-scoped cubit is removed.

Pure client-side display fix — no backend or Nostr protocol change (NIP-17 defines no read state).

**Related Issue:** Closes #4976

## Out of Scope

- **Account-switch staleness of the unread-count cubit** — the app-shell `DmUnreadCountCubit` captures its repositories via `ref.read` with no identity sync, so the badge can go stale after switching accounts. This is pre-existing (the bottom-nav badge already had it); consolidation makes the inbox toggle share it. Tracked in #5472 (under epic #5178), to be fixed with the `NotificationBadgeCubit`-style `setRepositories()` + sync-widget pattern.

## Verification

- `flutter analyze lib test integration_test` — no issues.
- `dart format` — clean.
- New `DmUnreadCountCubit` unit tests (12), including a regression that fails on the old follow-blind count (an unread chat from a followed peer you've never replied to now increments the badge), blocklist exclude/passthrough/reactivity, unpaginated counting, and follow-list-change reactivity.
- Updated `inbox_page_test.dart` to provide the now app-scoped cubit.
- Affected suites (dm blocs + inbox screens + bottom nav + app-shell badge), random ordering: 339 passing. Repo delegation test passing.

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
[...]

PR #6146, opened 2026-07-17
fix(dm): create app-shell badge cubits eagerly to stop re-entrant lazy create (#6115)

### LINKED ISSUE #6115: fix(dm): DmUnreadCountCubit null cast in bottom nav / inbox — new in 1.0.16 (Crashlytics 21a45213)
## Crashlytics
- Issue: `21a45213344b717e884a58ff4707a35e` ([console](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/21a45213344b717e884a58ff4707a35e))
- Type: non-fatal · **200 impacted users / 1232 events** · **new in 1.0.16**
- Versions: first & last **1.0.16**

## Signature
`package:provider/src/inherited_provider.dart - _CreateInheritedProviderState.value`
`type 'Null' is not a subtype of type 'DmUnreadCountCubit' in type cast` (thrown, runZonedGuarded).

## Root cause (hypothesis)
`DmUnreadCountCubit` is watched at:
- `lib/widgets/vine_bottom_nav.dart:97` — `context.watch<DmUnreadCountCubit>()`
- `lib/screens/inbox/inbox_view.dart:135` — `context.watch<DmUnreadCountCubit>()`

The `Null → DmUnreadCountCubit` cast means a descendant reads the cubit through a `Provider` whose value is `null` at read time — the provider isn't above the widget yet (bottom nav renders before the cubit is provided/wired) or was torn down. `main.dart:2625` sets its repositories after construction, so there's a mount window where the value is absent. Regression is fresh on 1.0.16, so a recent wiring change to the bottom-nav/inbox provider tree likely introduced it.

## Suggested next step
Confirm where `DmUnreadCountCubit` is provided relative to `vine_bottom_nav`. Either hoist the provider above the nav or make the read tolerate a not-yet-provided cubit. 200 users on the current build.

realmeylisdev:
## Description

Crashlytics `21a45213` (1232 events / 200 users, new in 1.0.16, non-fatal): `type 'Null' is not a subtype of type 'DmUnreadCountCubit' in type cast`, reported through `runZonedGuarded`.

Root cause — verified frame-by-frame against the `1.0.16` tag and the exact lockfile package sources (`provider 6.1.5+1`, `riverpod 3.3.2`, `flutter_bloc 9.1.1`), and reproduced deterministically in the new regression test:

1. `_InboxBadgeRepositorySync` (introduced in #5473, first shipped in 1.0.16) `ref.listen`s the dm/follow/blocklist repository providers; the callbacks `context.read` the app-shell badge cubits.
2. Those providers all rebuild together in one invalidation wave (they watch `nostrServiceProvider`, which flips on auth-ready / account switch).
3. When the wave fires **before any widget has consumed the lazy `BlocProvider<DmUnreadCountCubit>`** (cold-start window before the first bottom-nav/inbox build), the first listener triggers the lazy `create`. Inside `create`, `ref.read(followRepositoryProvider)` flushes a still-dirty provider of the same wave, which synchronously re-notifies the sync listener → `context.read<DmUnreadCountCubit>()` re-enters the provider mid-`create`: `_didInitValue` is already `true` but `_value` is still `null`, so `_value as T` (inherited_provider.dart:799) throws the Null cast.
4. Riverpod's `runBinaryGuarded` catches it, routes it to `runZonedGuarded` → Crashlytics, and the outer `create` then completes — which is why the crash self-heals and users see no breakage, just noise.

The issue body's original hypothesis (bottom nav reading before the provider is above it) is not what happens: the crashing read comes from the sync listener, not from `vine_bottom_nav`, and the provider is always above it.

### Changes

- **`main.dart`** — `lazy: false` on the `DmUnreadCountCubit` and `NotificationBadgeCubit` `BlocProvider`s. Eager creation runs during `_DivineAppState.build`, before the descendant sync widget can register its listeners, so a lazy create can never start inside a Riverpod notification again. (`NotificationBadgeCubit` cannot self-re-enter today — its `create` reads only the provider that just notified — but it is one refactor away from the same bug, and the sync listener `context.read`s it identically.)
- **`test/blocs/dm/unread_count/dm_unread_count_cubit_eager_create_test.dart` (new)** — mirror-harness widget tests over the real `dmRepositoryProvider`/`followRepositoryProvider` (overridden): (1) eager create happens at mount with no consumer present; (2) a two-provider wave re-points the existing cubit with no error; (3) a documentation test pins the 1.0.16 mechanism itself — with `lazy: true` the same wave throws exactly one `TypeError` and then self-heals (outer create completes, fresh repositories wired).
- **`test/blocs/notifications/badge/notification_badge_cubit_auth_transition_test.dart`** — eager-create coverage for the `NotificationBadgeCubit` flag.

**Related Issue:** Fixes #6115

## Out of Scope

- Riverpod/provider package versions and the underlying provider-package re-entrancy behavior (`_didInitValue` set before `create` completes) — upstream design, worked around by eager creation.
- The `_InboxBadgeRepositorySync` sync mechanism itself (#5473) — unchanged; only the cubits' creation timing moves.

## Verification

- `dart format` — clean.
- `flutter test test/blocs/dm/unread_count/ test/blocs/notifications/badge/` — all passing, including the mechanism-documentation test proving the pre-fix TypeError and the self-heal.
- Cross-repo sweep of all 64 local divinevideo repos: none can causally trigger this crash or break its self-healing; at most six (funnelcake, moderation-service, keycast, push-service, rest-gateway, router) influence *when* the client's own invalidation wave fires.

## Type of Change


hm21 (approved):
LGTM. One non-blocking gap: the new eager tests pump their own mirror probes with their own `lazy` flag, so reverting `lazy: false` in `main.dart` turns no test red — consider extracting the two badge `BlocProvider`s + `_InboxBadgeRepositorySync` into a small widget that both `main.dart` and the tests pump, so the flags are actually pinned.

realmeylisdev:
Addressed the non-blocking gap in a84efb6cff. Extracted the two badge `BlocProvider`s + `_InboxBadgeRepositorySync` into `AppShellBadgeScope`; `main.dart` and the new `test/widgets/app_shell_badge_scope_test.dart` now pump that same widget, so the `lazy: false` flags are pinned — I flipped each flag back to lazy independently and confirmed the eager-create test fails both times. The pre-fix re-entrancy reproduction and the notification auth-transition coverage are preserved.

mobile/lib/widgets/app_shell_badge_scope.dart
```diff
@@ -0,0 +1,111 @@
...
+/// live repositories as auth/account state changes.
+///
+/// Both cubits are created **eagerly** (`lazy: false`). The descendant
+/// [_InboxBadgeRepositorySync] `context.read`s them from inside Riverpod
+/// notifications; a lazy create started from there re-enters itself when its
+/// own `ref.read` flushes another still-dirty provider of the same
+/// invalidation wave, and the re-entrant read throws a `Null`-cast `TypeError`
+/// (#6115). Creating both cubits during this widget's build — before the sync
+/// listener can register — means no create can ever start inside a Riverpod
+/// notification.
```
mbradley: The mechanism as written covers `DmUnreadCountCubit`'s three-provider create; `NotificationBadgeCubit` reads a single provider and can't self-flush today (as the PR body notes). Tightening the comment to name which cubit is actually at risk:

```suggestion
/// Both cubits are created **eagerly** (`lazy: false`). The descendant
/// [_InboxBadgeRepositorySync] `context.read`s them from inside Riverpod
/// notifications. [DmUnreadCountCubit]'s create reads three repository
/// providers, so a lazy create triggered from that listener re-enters itself
/// when its own `ref.read` flushes another still-dirty provider of the same
/// invalidation wave, and the re-entrant read throws a `Null`-cast `TypeError`
/// (#6115). [NotificationBadgeCubit] reads a single provider and cannot
/// self-flush today, but is `context.read` from the same listener and one
/// refactor from the same trap, so it is created eagerly too. Creating both
/// during this widget's build — before the sync listener can register — means
/// no create can start inside a Riverpod notification.
```

mbradley (approved):
Reviewed the a84efb6 refactor (it post-dates the approval on bc8a764). Traced the fix against provider 6.1.5+1: with `lazy: false` the eager create runs during the BlocProvider element's build, before `_InboxBadgeRepositorySync` mounts and registers its `ref.listen`s, so no listener can re-enter `context.read` when create's `ref.read` flushes a dirty sibling. The re-entrancy is structurally closed, not narrowed. Confirmed the pin by flipping each `lazy: false` to `true` independently; the eager-create test fails both times. The tree move is regression-free: cubits stay above the sync widget and all consumers, `DmRestrictionGateSync` wraps the same child, and the removed `main.dart` imports have no remaining references. `analyze` and the DM/notification badge suites are green. One inline note below (a doc-precision suggestion); non-blocking.

as usual, Alex is very quick here. :)

mobile/lib/widgets/app_shell_badge_scope.dart
realmeylisdev: Applied in fb5de090a3.

PR #6149, opened 2026-07-17
feat(video_editor): draw each clip's waveform over its thumbnails

### LINKED ISSUE #6150: feat(video_editor): draw each clip's waveform over its thumbnails
Auto-created from PR #6149.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6149
**Branch:** `feat/clip-waveform-overlay`

hm21:
Closes #6150

## Description

Clip tiles in the timeline showed frames only — to see where a clip is loud or silent you had to open the sound strip. Each video clip now draws its own audio as a translucent waveform band along the tile's bottom edge.

**How it lines up:** the band rides the same translate as the thumbnail row (offset by the raster phase, since the row starts one phase before the clip's zero), so the bars stay glued to the frames through trim, zoom and scroll instead of drifting against them.

Bars map straight from clip time rather than being fitted to the tile: the band spans the audio's own length, and whatever runs past the tile is culled. That matters for a trim-based split's start half, which is rebased to the split point while still sharing the full-length file — fitting its waveform to the tile would draw the end half's sound under the start half's frames.

**Why extraction is keyed by file path, not clip id:** the two halves of a trim-based split share one source file, so they resolve from a single pass; undo/redo restores hit the cache (FIFO, 32 entries — a few KB each). Extractions are serialized because the strip's thumbnail generator and the editor preview already compete for the same hardware decoders. A file with no audio track (the typed `AudioNoTrackException`) caches the miss, since a retry could never draw bars. Other failures don't cache — a decoder busy under render contention says nothing about the file — and get bounded retries across syncs, bounded because `sync` runs on every trim-drag frame.

**Why the bars are normalized + curved, not linear:** real recordings peak well below full scale, so mapping raw amplitude onto a 22px band drew an almost flat line. Bars normalize against the clip's own loudest sample, and a compressive exponent lifts the quiet body of the signal — audio spends most of its time far below its peak, so normalization alone would still read as a flat line with occasional spikes. A floor on the normalizer keeps near-silence from being stretched to full height; true silence stays a flat baseline.

**Volume-aware:** bar heights scale with `clip.volume`, so muting flattens the band to its baseline instead of leaving a loud-looking waveform behind. The band is hidden during a reorder drag, where the tile shrinks to a 64px badge.

Tuning knobs live in `TimelineConstants`: `clipWaveformCurve` (1.0 = linear), `clipWaveformNormalizerFloor`, `clipWaveformBandHeight`, `clipWaveformOpacity`.

**Related Issue:** 

## Out of Scope

- The sound overlay strip's own waveform keeps its existing linear scaling — this PR does not touch it.
- Stop-motion clips: `DivineVideoClip.video` is non-nullable on `main`, so there is nothing to handle here yet. When the stop-motion branch lands and makes it nullable, `ClipWaveformManager.sync` needs its null branch back (frames carry no audio file).

## Verification

- `flutter test test/widgets/video_editor/timeline_editor/strips/ test/services/video_editor/clip_waveform_manager_test.dart` — 98 passing
- `check_untested_services_floor.sh` — no new entries
- New coverage: 14 unit tests for the manager (per-path sharing, re-extract on source change, cache, no-audio miss, transient retry + give-up, peak) and 8 painter tests via the `paints` matcher (bottom anchoring, per-step bars, volume scaling, mute baseline, quiet-clip fill, near-silence floor, short audio track, over-long audio track), plus 3 wiring tests in the strip (band per clip, none before extraction, none mid-reorder) and one pinning the band's alignment to trim-start against the raster phase.
- Each new test was checked to fail against the pre-fix code, not just to pass after it.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/services/video_editor/clip_waveform_manager.dart
+++ b/mobile/lib/services/video_editor/clip_waveform_manager.dart
@@ -27,14 +27,19 @@ class ClipWaveformManager {
   ClipWaveformManager({ClipWaveformExtractor? extractor})
     : _extract = extractor ?? _extractWithProVideoEditor;
 
-  /// Sample density for extraction. At the timeline's maximum zoom a bar
-  /// covers ~2px, so 200 samples/second still resolves more detail than the
-  /// strip can draw — at ~8KB per minute of audio.
+  /// Sample density for extraction. 200 samples/second resolves every bar the
+  /// strip can draw below ~400px/s of zoom, and stays within a few KB per clip
+  /// at the recorder's 6.3s cap.
   static const WaveformResolution _resolution = WaveformResolution.high;
 
   /// Cached waveforms retained after their clip is gone, so undo/redo does not
   /// re-extract. Insertion-ordered for FIFO eviction.
   static const _maxCachedWaveforms = 32;
+
+  /// Extraction attempts per file before a repeatedly failing one is given up
+  /// on as bandless. Bounded so a `sync` storm (every trim-drag frame) can't
+  /// retry forever.
+  static const _maxExtractionAttempts = 3;
 
   static Future<WaveformData> _extractWithProVideoEditor(String videoPath) {
     return ProVideoEditor.instance.getWaveform(
@@ -56,6 +61,9 @@ class ClipWaveformManager {
 
   final Map<String, ClipWaveform> _cache = {};
   final Set<String> _inFlight = {};
+
+  /// Consecutive transient extraction failures per file path.
+  final Map<String, int> _failures = {};
 
   /// Serializes extractions — the strip's thumbnail generator and the editor
   /// preview already compete for the same hardware decoders.
@@ -117,7 +125,13 @@ class ClipWaveformManager {
       if (_disposed || !_isReferenced(path)) return;
       final data = await _extract(path);
       if (_disposed) return;
+      _failures.remove(path);
       _publish(path, _reduce(data));
+    } on AudioNoTrackException {
+      if (_disposed) return;
+      // Cache the miss: the file carries no audio track, so retrying could
+      // never draw bars. The strip just renders none.
+      _publish(path, const ClipWaveform.empty());
     } catch (e) {
       Log.warning(
         'Failed to extract clip waveform: $e',
@@ -125,12 +139,24 @@ class ClipWaveformManager {
         category: LogCategory.video,
       );
       if (_disposed) return;
-      // Cache the miss: a file without a readable audio track would otherwise
-      // be retried on every sync. The strip just renders no bars.
-      _publish(path, const ClipWaveform.empty());
+      _onTransientFailure(path);
     } finally {
       _inFlight.remove(path);
     }
+  }
+
+  /// Reopens [path] for extraction after a failure that isn't "no audio track"
+  /// — a decoder busy under render contention says nothing about the file, and
+  /// caching it as bandless would blank the clip for the rest of the session.
+  /// Releasing the clips' resolved path lets the next [sync] re-extract.
+  void _onTransientFailure(String path) {
+    final attempts = (_failures[path] ?? 0) + 1;
+    _failures[path] = attempts;
+    if (attempts >= _maxExtractionAttempts) {
+      _publish(path, const ClipWaveform.empty());
+      return;
+    }
+    _paths.removeWhere((_, resolved) => resolved == path);
   }
 
   /// Whether any clip still points at [path]. A clip removed while its
@@ -172,5 +198,6 @@ class ClipWaveformManager {
     _paths.clear();
     _cache.clear();
     _inFlight.clear();
+    _failures.clear();
   }
 }
--- a/mobile/lib/services/video_editor/clip_waveform_manager.dart
+++ b/mobile/lib/services/video_editor/clip_waveform_manager.dart
@@ -37,9 +37,13 @@ class ClipWaveformManager {
   static const _maxCachedWaveforms = 32;
 
   /// Extraction attempts per file before a repeatedly failing one is given up
-  /// on as bandless. Bounded so a `sync` storm (every trim-drag frame) can't
-  /// retry forever.
+  /// on as bandless.
   static const _maxExtractionAttempts = 3;
+
+  /// Delay before a transiently failed extraction is retried. Long enough for
+  /// the decoder contention it usually signals — a render, the strip's own
+  /// thumbnail pass — to clear, rather than re-queuing into the same jam.
+  static const _retryDelay = Duration(seconds: 2);
 
   static Future<WaveformData> _extractWithProVideoEditor(String videoPath) {
     return ProVideoEditor.instance.getWaveform(
@@ -64,6 +68,9 @@ class ClipWaveformManager {
 
   /// Consecutive transient extraction failures per file path.
   final Map<String, int> _failures = {};
+
+  /// Pending retries per file path, so dispose can cancel them.
+  final Map<String, Timer> _retryTimers = {};
 
   /// Serializes extractions — the strip's thumbnail generator and the editor
   /// preview already compete for the same hardware decoders.
@@ -145,10 +152,14 @@ class ClipWaveformManager {
     }
   }
 
-  /// Reopens [path] for extraction after a failure that isn't "no audio track"
-  /// — a decoder busy under render contention says nothing about the file, and
-  /// caching it as bandless would blank the clip for the rest of the session.
-  /// Releasing the clips' resolved path lets the next [sync] re-extract.
+  /// Schedules a retry after a failure that isn't "no audio track" — a decoder
+  /// busy under render contention says nothing about the file, and caching it
+  /// as bandless would blank the clip for the rest of the session.
+  ///
+  /// The retry owns its own timer rather than waiting for the next [sync]:
+  /// contention peaks when the editor opens, and nothing guarantees another
+  /// sync afterwards — the strip only re-syncs when its widget updates, so an
+  /// idle timeline would keep the clip bandless indefinitely.
   void _onTransientFailure(String path) {
     final attempts = (_failures[path] ?? 0) + 1;
     _failures[path] = attempts;
@@ -156,7 +167,12 @@ class ClipWaveformManager {
       _publish(path, const ClipWaveform.empty());
       return;
     }
-    _paths.removeWhere((_, resolved) => resolved == path);
+    _retryTimers[path]?.cancel();
+    _retryTimers[path] = Timer(_retryDelay, () {
+      _retryTimers.remove(path);
+      if (_disposed || !_isReferenced(path)) return;
+      _enqueue(path);
+    });
   }
 
   /// Whether any clip still points at [path]. A clip removed while its
@@ -191,6 +207,10 @@ class ClipWaveformManager {
   /// Disposes every notifier and stops publishing queued extractions.
   void dispose() {
     _disposed = true;
+    for (final timer in _retryTimers.values) {
+      timer.cancel();
+    }
+    _retryTimers.clear();
     for (final notifier in _notifiers.values) {
       notifier.dispose();
     }
--- a/mobile/lib/services/video_editor/clip_waveform_manager.dart
+++ b/mobile/lib/services/video_editor/clip_waveform_manager.dart
@@ -7,6 +7,7 @@ import 'dart:math' as math;
 
 import 'package:flutter/foundation.dart';
 import 'package:openvine/models/divine_video_clip.dart';
+import 'package:openvine/models/video_editor/clip_waveform.dart';
 import 'package:pro_video_editor/pro_video_editor.dart';
 import 'package:unified_logger/unified_logger.dart';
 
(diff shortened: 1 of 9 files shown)


realmeylisdev (commented):
Careful, well-tested PR — I traced the alignment/culling math (including trim, split, raster-phase and loop-wrap) and it holds, and the commit history already caught the hard cases. No blockers. Left 6 optional nits inline: two extraction-retry lifecycle edges (a stale retry timer re-decodes a cached path; a stale `_failures` count can shrink a re-added clip's retry budget), a split-half normalizer comment that overstates what the code does, a live-volume-drag preview gap, and two test-hardening gaps. None hold up merge.

mobile/lib/widgets/video_editor/timeline_editor/strips/clip_waveform_overlay.dart
```diff
@@ -0,0 +1,156 @@
...
+      TimelineConstants.clipWaveformBandHeight,
+      size.height,
+    );
+    final paint = Paint()..color = color;
+    final peaks = waveform.peaks;
+
+    // Normalize against the clip's own loudest sample, so a quietly recorded
+    // clip still fills the band. The floor keeps near-silence flat instead of
+    // amplifying room noise into a convincing-looking waveform.
+    final normalizer = math.max(
```
realmeylisdev, line 110: Nit (comment vs behaviour): both halves of a trim-based split share one cached `ClipWaveform` keyed by file path, so `waveform.peak` here is the *whole file's* peak, not the visible half's. For a quiet-start / loud-end recording the start half normalizes against a peak that lives in the culled tail — worst case its own bars compress ~2–3×. The shared normalizer is arguably the right call (relative loudness stays consistent across the split), but "the clip's own loudest sample" doesn't literally hold for split halves. Suggest tightening the comment (or deriving the normalizer from the visible sample range if you want true per-half normalization).

mobile/lib/services/video_editor/clip_waveform_manager.dart
```diff
@@ -0,0 +1,225 @@
...
+        notifier.value = cached;
+        continue;
+      }
+      // The previous waveform stays on screen until the new one lands —
+      // clearing here would blink the bars away on every re-render swap.
+      _enqueue(path);
+    }
+  }
+
+  void _enqueue(String path) {
```
realmeylisdev, line 141: Nit (redundant decode): `_enqueue` guards on `_inFlight` but not `_cache`, and a successful extraction (`_failures.remove` below) doesn't cancel a still-pending retry timer. Sequence: clip A on file P fails transiently → retry timer armed (2s); within that window a second clip on P (split sibling / undo re-add) cache-misses and extracts P successfully → cache populated, timer still live. At +2s it fires and re-decodes P even though it's cached — an extra hardware-decoder pass under exactly the contention `_queue` exists to avoid. In the tail (that decode also fails 3×) it can overwrite the good on-screen band with `empty`. A `if (_cache.containsKey(path)) return;` early-out in `_enqueue`, or cancelling `_retryTimers[path]` on success, closes it.

mobile/lib/services/video_editor/clip_waveform_manager.dart
```diff
@@ -0,0 +1,225 @@
...
+  /// Syncs waveforms with the current clip list — starts extraction for new
+  /// clips and for clips whose source file changed, and cleans up removed ones.
+  void sync({required List<DivineVideoClip> clips}) {
+    if (_disposed) return;
+
+    final currentIds = clips.map((c) => c.id).toSet();
+    for (final id in _notifiers.keys.toList()) {
+      if (currentIds.contains(id)) continue;
+      _notifiers.remove(id)?.dispose();
+      _paths.remove(id);
```
realmeylisdev, line 94: Nit (stale failure count): this removal loop prunes `_notifiers`/`_paths` but not `_failures[path]` (only a success or `dispose` clears it). If clip A on P fails transiently twice (`_failures[P]=2`), is removed/undone before the 3rd retry, then re-added on P, its first fresh failure computes `attempts = 2+1 = 3 ≥ max` → immediately caches `empty` and blanks the band for the session after a single attempt — contradicting the "transient failures say nothing about the file" rationale just below. Pruning `_failures`/`_retryTimers` for no-longer-referenced paths here fixes it and bounds the maps.

mobile/lib/widgets/video_editor/timeline_editor/strips/video_editor_timeline_clip_strip_tiles.dart
```diff
@@ -424,6 +467,33 @@ class _ClipTile extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    return ValueListenableBuilder<ClipWaveform?>(
+      valueListenable: waveformNotifier,
+      builder: (context, waveform, _) {
+        if (waveform == null) return const SizedBox.shrink();
+        return ClipWaveformOverlay(
+          waveform: waveform,
+          clipDuration: clip.duration,
+          volume: clip.volume,
```
realmeylisdev, line 490: Nit (live-preview gap): the band scales with the committed `clip.volume`, but a volume drag only commits on `onPanEnd` — mid-drag the arc fill and percentage animate live off `volumePreviewNotifier` while the band holds its pre-drag height and snaps flat only on release. Not wrong (committed state is always correct), just inconsistent live feedback. The single global `volumePreviewNotifier` carries no active-clip id, so wiring it per-clip is non-trivial — flagging as optional polish.

mobile/test/widgets/video_editor/timeline_editor/strips/video_editor_timeline_clip_strip_test.dart
```diff
@@ -120,6 +140,58 @@ void main() {
 
         expect(find.byType(ClipRRect), findsOneWidget);
       });
+
+      testWidgets('draws each clip its own waveform band', (tester) async {
```
realmeylisdev, line 144: Nit (coverage): these strip tests match only on painter *type* (`painter is ClipWaveformPainter`) and never inspect its `volume`/`clipDuration`, and `_createTestClip` never sets volume — so a regression hardcoding `volume: 1`, or forwarding `waveform.duration` instead of `clip.duration` in `_ClipWaveformLayer`, would stay green here (clip2 is already 4s vs the injected 3s waveform, yet nothing distinguishes them). A muted-clip case (expect baseline band) or a trimmed clip (expect culling) at strip level would guard the forwarding.

mobile/test/services/video_editor/clip_waveform_manager_test.dart
```diff
@@ -0,0 +1,351 @@
...
+      // Another clip resolving to the same file must serve the cached miss.
+      silent.sync(
+        clips: [_clip(id: 'b', path: '/tmp/silent.mp4')],
+      );
+      await pumpEventQueue();
+
+      expect(calls, equals(1));
+    });
+
+    test('retries a transient extraction failure without a further sync', () {
```
realmeylisdev, line 192: Nit (coverage): the retry tests cover "retries on its own timer" and "gives up after 3 → empty", but nothing drives throw-once-then-succeed, so the success-after-failure branch (`_failures.remove` + non-empty publish) is unguarded — dropping `_failures.remove(path)` from the success branch keeps every current test green. A `fakeAsync` test whose extractor throws once then returns real `WaveformData`, asserting a non-empty notifier and a reset failure count, would cover it (and pairs with the `_failures`-lifecycle nit).

mobile/lib/services/video_editor/clip_waveform_manager.dart
hm21: Confirmed and fixed in 9cd3b1fe2f. Reproduced it: re-added clip published `empty` after a single fresh failure. sync now prunes `_failures`/`_retryTimers` for unreferenced paths, which also bounds the maps.

mobile/lib/widgets/video_editor/timeline_editor/strips/video_editor_timeline_clip_strip_tiles.dart
hm21: Agreed on all counts, including non-trivial — leaving it. A per-clip preview means giving `volumePreviewNotifier` a clip identity, which it shares with the audio-track arcs, so it touches ~9 files (timeline -> body -> strip -> positions -> tiles, plus header and the audio path) to fix a self-correcting during-gesture transient. Committed state is always right. Happy to do it as its own PR if you want it.

mobile/test/widgets/video_editor/timeline_editor/strips/video_editor_timeline_clip_strip_test.dart
hm21: Good catch — the factory had no volume knob at all. Added both cases in 0529dac534: muted clip (baseline band) and a 4s clip vs the 3s waveform. Verified they earn their place: hardcoding `volume: 1` and swapping in `waveform.duration` each turn exactly one of them red, and left every previous test green.

mobile/test/services/video_editor/clip_waveform_manager_test.dart
hm21: Half taken, in 9cd3b1fe2f: added throw-once-then-succeed asserting the notifier gets the real peaks — you're right that the retry's actual payoff was unguarded, which is the point of the whole commit. Skipped the `_failures.remove` assertion though: after the `_failures`-pruning fix, a succeeded path is cached and `_enqueue` now early-outs on cache, so the counter can only be re-read after a FIFO eviction (32+ distinct files) — the line stays as hygiene, but a test for it would need contorted setup for state that is otherwise unobservable. Shout if you disagree.

mobile/lib/widgets/video_editor/timeline_editor/strips/clip_waveform_overlay.dart
hm21: Right, the comment overclaimed — the cache is keyed by file, so split halves share one peak. Kept the shared normalizer (splitting must not restyle the bars either side of the cut) and fixed the comment to say so. d33c25d5b6

mobile/lib/services/video_editor/clip_waveform_manager.dart
hm21: Confirmed and fixed in 9cd3b1fe2f — reproduced your sequence as a test first, it decoded 3x instead of 2x. Did both: `_enqueue` early-outs on a cached path, and `_publish` cancels the retry it makes moot.

rabble (approved):
**Reviewed with full checkout + verification.** ✅ Approve. Well-engineered, correctly layered, strong tests. No blocking issues.

Rendering performance (the thing that could have gone wrong here) is all correct:
- Extraction runs in a dedicated `ClipWaveformManager`, off the UI thread via the native `ProVideoEditor` extractor, and is **serialized** through a single `Future _queue` so it doesn't contend with the thumbnail generator / editor preview for hardware decoders.
- `CustomPaint` is wrapped in `RepaintBoundary` + `IgnorePointer` + `ExcludeSemantics`; each tile's band lives behind its own `ValueListenableBuilder`, so an arriving waveform repaints only that band, not the thumbnail row.
- `ClipWaveformPainter.shouldRepaint` correctly compares waveform/clipDuration/volume/color (identity comparison on the waveform is right — the manager publishes one instance per notifier).
- Caching: per-source-path (so the two halves of a trim-split resolve from one pass; undo/redo hits cache), FIFO-capped at 32, peaks as `Float32List` (~5KB/clip). Full disposal — `dispose()` cancels all retry timers and disposes all notifiers; the strip disposes `_waveforms` in its own `dispose()`.
- No expensive work in `build()`; sync runs in `didChangeDependencies`/`didUpdateWidget`.

Robustness: typed failure handling — `AudioNoTrackException` caches the miss (no pointless retries); transient/decoder-busy failures get bounded retries (3 × 2s) then give up as bandless; retry tallies drop when a clip leaves the timeline. All covered by fake-async tests verified to fail against pre-fix code. Theming compliant (`VineTheme.whiteText.withValues(alpha:)`, no raw `Colors.*`, tuning knobs are named `TimelineConstants`). Waveform bars map from clip time (band spans the audio's own length, overflow culled) — correct for a trim-split start half sharing the full-length file — and ride the same raster-phase translate as the thumbnail row, pinned by a raster-stability test.

Verification (local):
- `flutter analyze` (changed dirs) → No issues found!
- `flutter test` manager + overlay + clip-strip → 55 tests passed.
- `check_untested_services_floor.sh` → OK, no new entries (new `clip_waveform_manager.dart` ships its same-named test).

Non-blocking nits: (1) `ClipWaveform` is `@immutable` but doesn't override `==`/`hashCode` — fine today (all comparisons identity-based), worth noting if it ever lands in an Equatable state; (2) normalization/compressive-curve math lives in the painter — legitimate display mapping (data→pixels), not business logic, so the right home. The manager being constructed/disposed in widget State (not injected) mirrors the pre-existing `ClipThumbnailManager` in the same file — consistent UI-owned rendering helper, not a violation.

Issue #5179, opened 2026-06-15
epic(security): harden key handling, signing, and auth boundaries


NotThatKindOfDrLiz:
## Scope
Current implementation epic for security, key handling, signing safety, dependency hygiene, and Nostr/auth hardening items that do not fit the account-recovery epic.

## Includes
- private-key containment and log redaction
- secure storage guarantees and BYOK leakage fixes
- Nostr publish/security semantics
- dependency override/release-safety work

## Initial triage
Created from the parentless open-issue audit on 2026-06-15.

Issue #3761, opened 2026-04-30
refactor(auth): migrate nostrconnect failure paths to reason codes


realmeylisdev:
## Context

Follow-up from #3355 / #3760 (NIP-46 nostrconnect secret-match fix). The security mechanics shipped in #3760 are correct, but the surrounding error-propagation pattern is on the codebase's migration list and #3760 perpetuated it. This issue tracks the cleanup as a separate, scoped refactor.

## The legacy pattern

`NostrConnectSession` exposes English error strings via a `String? errorMessage` getter. Those flow through `AuthService.AuthResult.errorMessage` (also `String?`) into `NostrConnectScreen._errorMessage`, where they ultimately render as `_ErrorContent.message` in the UI — bypassing `context.l10n.*` entirely. The `?? context.l10n.authUnknownError` fallback at `mobile/lib/screens/auth/nostr_connect_screen.dart:327` is the smoking gun: the UI has a localized string ready, but the upstream English string wins when present.

Same pattern repeats in `login_options_screen.dart:169`, `key_import_screen.dart:290`, and `auth_service.dart:953`.

## Canonical pattern in this codebase

`mobile/lib/utils/invite_error_utils.dart:11` — `InviteActivationFailureReason`. Its doc-comment explicitly names this exact migration as in-progress:

> Callers in the BLoC/Cubit layer must store one of these reasons (or a mapped EmailVerificationError) — never a raw English string.
>
> Use this from the cubit/BLoC layer so state never carries English copy. The UI layer maps the reason to a localized string.
>
> Legacy string-based helper retained for pre-existing callers (DivineAuthCubit) that still store English strings in state … When DivineAuthCubit is migrated to the reason-code pattern, this helper can be removed.

`NostrConnectSession`'s `_errorMessage` field is exactly the legacy shape this comment calls out.

### 1. Add a `NostrConnectFailureReason` enum

```dart
// in mobile/packages/nostr_sdk/lib/nip46/nostr_connect_session.dart
@visibleForTesting
enum NostrConnectFailureReason {
  /// Programmer error: session entered the response handler with no
  /// expected secret. Should never happen in practice.
  noExpectedSecret,

  /// Signer responded with `response.error` set — explicit user-driven
  /// rejection.
  bunkerRejected,

  /// Wait window elapsed without a valid response (existing
  /// NostrConnectState.timeout case).
  timedOut,

  /// User cancelled mid-flow.
  cancelled,

[...]

PR #3760, opened 2026-04-30
fix(security): enforce NIP-46 nostrconnect secret match (#3355)

realmeylisdev:
## Description

Closes #3355.

`NostrConnectSession._handleResponse` claimed to enforce NIP-46's "client MUST validate the `secret` returned by `connect` response" rule but didn't:

```dart
// before
final isValidSecret =
    response.result == expectedSecret ||
    response.result == 'ack' ||
    response.result == 'connect';
// ...
if (!isValidSecret) {
  log('...does not match expected secret "$expectedSecret"...accepting anyway');
}
// falls through to success
```

The `isValidSecret` flag was computed but only consulted for log severity. Any pubkey on the listening relays could publish a junk response and the client would bind to that pubkey as the remote signer.

`"ack"` and `"connect"` are response tokens for the **`bunker://`** flow's `connect` method — they are not legal results for **`nostrconnect://`** (client-initiated). Conflating the two flows was the original mistake.

## Fix (Approach E from the brainstorm)

1. **Extract `validateConnectResponse`** as a `@visibleForTesting` pure function returning a four-case enum (`match`, `rejectedByBunker`, `invalidSession`, `ignore`). The security-critical decision is now exhaustively unit-testable with no state, no signer, no relays.

2. **`_handleResponse` dispatches on the enum** via an exhaustive switch — the compiler enforces that no fallthrough to success is possible. Mismatched non-error responses transition to `ignore` (silent drop, keep listening); the existing 2-minute timeout in `waitForConnection` terminates the wait if no valid response ever arrives.

   *Why silent-drop instead of immediate-error on mismatch?* Treating mismatches as terminal would let any pubkey on the listening relays DoS pairing by racing the legitimate signer's reply with a junk response — strictly worse than today's broken-but-permissive behavior in adversarial scenarios. The brainstorm doc captures the trade-off: see `mobile/docs/brainstorm/2026-05-01-issue-3355-nip46-mismatch-handling-brainstorm.md`.

3. **Log redaction.** Sanitize the up-front decrypted-response log to emit `hasResult` / `hasError` booleans instead of the raw strings, drop `expectedSecret` from all log lines, and add a source-level test that pins the absence of the offending substrings so future regressions fail loudly.

Constant-time string compare on the secret is defense-in-depth — the secret is ephemeral per session, but the cost is one tight loop and zero excuse not to.

## Type of Change

- [x] 🛠️ Bug fix (non-breaking change which fixes a critical security issue)

## Test Plan

[...]

PR #6151, opened 2026-07-17
fix(auth): handle NIP-46 auth_url challenge as non-terminal in nostrconnect pairing

### LINKED ISSUE #5683: fix(auth): handle NIP-46 auth_url challenge in nostrconnect (currently mislabeled as rejection)
## Context

Follow-up from #3761 (nostrconnect failure-path reason-code migration). While migrating those paths we confirmed a pre-existing protocol-handling bug that shipped with the secret-match fix (#3760). It is a behavior change rather than error-message localization, so it was split out of #3761.

## Problem

Per [NIP-46](https://github.com/nostr-protocol/nips/blob/master/46.md), a remote signer can respond to a connect request with an **auth challenge** rather than a result:

```json
{ "id": "<request_id>", "result": "auth_url", "error": "<URL_to_display_to_end_user>" }
```

The URL is carried in the `error` field. The client is expected to display it (popup/new tab) and **keep listening** for a later response on the same request id, which arrives once the user authenticates out-of-band.

`validateConnectResponse` in `mobile/packages/nostr_sdk/lib/nip46/nostr_connect_session.dart` treats **any** non-empty `response.error` as a terminal rejection:

```dart
final error = response.error;
if (error != null && error.isNotEmpty) {
  return NostrConnectResponseValidation.rejectedByBunker;
}
```

So an auth-URL challenge is misclassified as `rejectedByBunker`: the session terminates and the UI shows the "declined the connection" copy (`authBunkerRejectedConnection`, added in #3761). The user is never shown the auth URL and pairing fails.

This affects spec-following third-party signers that use the challenge flow (e.g. nsec.app / nsecBunker). Divine's own signer (keycast) does **not** use `auth_url`, so it is unaffected — verified across the keycast codebase.

## Proposed change

- Add a distinct `NostrConnectResponseValidation.authChallenge` outcome and detect `response.result == "auth_url"` in `validateConnectResponse` before the generic `error`-non-empty branch.
- In `NostrConnectSession`, surface the challenge URL (via state/callback) and **continue listening** instead of completing the session.
- UI (`nostr_connect_screen.dart`): open the URL and keep the waiting state.
- Flip the existing test `validateConnectResponse returns rejectedByBunker for auth_url challenge responses` in `nostr_connect_session_test.dart` to assert the new outcome, and add coverage for the keep-listening behavior.

## Out of scope

- The `rejectedByBunker` reason and its copy stay correct for genuine `error` responses; only the `auth_url` subset changes.

## Related

- Spawned by #3761.
- Part of #5179 (security/auth hardening epic).

realmeylisdev:
## Description

During `nostrconnect://` pairing, `validateConnectResponse` treated any non-empty `response.error` as terminal `rejectedByBunker`. A spec-following signer that answers `connect` with a NIP-46 auth challenge (`{result:"auth_url", error:<URL>}` — nsec.app does this whenever manual confirmation is needed) was therefore misclassified as a rejection: the session died and the user saw the false "Your signer app declined the connection." copy. The same branch let any unauthenticated relay pubkey kill pairing by racing a junk error in front of the legitimate signer's reply.

This PR makes the pairing path match reference-client behavior (nostr-tools `BunkerSigner.fromURI`, NDK `blockUntilReadyNostrConnect`), which resolves only on `result == secret` and treats everything else as non-terminal:

- `auth_url` challenges classify as a new `authChallenge` outcome: the session keeps listening for the real response on the same request id. The challenge URL is deliberately **not** surfaced, opened, or logged — during nostrconnect pairing the sender is unauthenticated (the subscription has no `authors` filter), so the URL is attacker-suppliable; #3760 removed the original pre-secret auth_url handling for exactly this reason and this PR preserves that posture.
- All other pre-secret errors classify as `ignore` (kept **before** the secret compare so an errored response can never bind), which also closes the pairing-DoS above. Trade-off: a genuine signer-side deny (`error:"Not authorized"`, verified in noauth source) now waits out the timeout — whose copy already says "Make sure you approved the connection in your signer app" — instead of showing an instant, attacker-spoofable "declined".
- The wait window restarts **exactly once** on the first challenge (out-of-band approval is slow; the once-cap keeps injected challenges from extending the wait forever).
- Relay replays are deduped on `event.id` before paying for NIP-44 decryption; signer re-sends of the same challenge id are deduped for logging.
- A secret match arriving after the wait already timed out is ignored instead of flipping the session to a `connected` state nobody is waiting on.
- The now-unreachable `rejectedByBunker` / `NostrConnectFailureReason.bunkerRejected` / `authBunkerRejectedConnection` chain is deleted (ARB key removed from all 18 locales, l10n regenerated).
- Shared challenge predicate extracted as `NostrRemoteResponse.isAuthChallenge`; the bunker:// flow's own auth_url handling is intentionally untouched (its sender is authenticated via the `bunker://` URI).

Design/investigation record: multi-round root-cause analysis with cross-repo verification (keycast never emits `auth_url`; nsec.app always surfaces its own in-tab confirm, so never surfacing the URL strands no signer; nsecbunkerd cannot pair via this receive-only flow at all). Full trail in `tasks/findings_5683.md` / `tasks/brainstorm_5683.md` (local, not committed).

**Related Issue:** Closes #5683. Relates to #3761, #3760, #5179.

## Out of Scope

- Surfacing the challenge URL behind a user-gated tap-to-open affordance (follow-up enhancement, needs security sign-off): #6153.
- The same bug class in divine-web (`NConnectSigner`): divinevideo/divine-web#485, and divine-connect: divinevideo/divine-connect#27.
- The bunker:// flow's `auth_url` handling (authenticated sender; unchanged).

## Verification

- `mobile/packages/nostr_sdk`: `flutter test` — 362/362 pass, including new tests: auth_url → `authChallenge` classification, malformed-challenge → `ignore`, pairing-DoS regression (junk error then real secret still binds), challenge keep-listening e2e with replay dedup, restart-once cap, deadline-extension (mutant-verified: removing the timer `cancel()` fails it), late-match-after-timeout, and a #3760-style redaction pin that the challenge URL never reaches logs.
- `mobile`: `flutter analyze lib test integration_test` clean; `flutter test test/screens/auth test/services/auth test/l10n/arb_consistency_test.dart` — 284/284 pass; `dart format` clean; l10n regenerated via gen-l10n.
- Manual verification planned post-merge: live nsec.app pairing (challenge only fires on first-time/unpermissioned connects).

## Type of Change

- [x] ❌ Breaking change (fix or feature that would cause existing functionality to change) — a genuine signer deny now times out instead of failing fast (deliberate, see Description)

### DIFF
--- a/mobile/packages/nostr_sdk/lib/nip46/nostr_connect_session.dart
+++ b/mobile/packages/nostr_sdk/lib/nip46/nostr_connect_session.dart
@@ -56,10 +56,6 @@ enum NostrConnectFailureReason {
   /// secret to validate against. Should never happen in practice.
   noExpectedSecret,
 
-  /// The signer set `response.error` — a terminal signer-side error or
-  /// explicit rejection.
-  bunkerRejected,
-
   /// The wait window elapsed without a valid response.
   timedOut,
 
@@ -179,6 +175,26 @@ class NostrConnectSession {
   Timer? _timeoutTimer;
   bool _isClosed = false;
 
+  /// Ids of relay events already processed. Up to 4 relays each deliver
+  /// the same event, and the fixed `since` timestamp replays history on
+  /// every reconnect — skip replays before paying for decryption.
+  /// Bounded by the events received in one short-lived pairing session.
+  final Set<String> _seenEventIds = {};
+
+  /// Response ids of auth challenges already logged, so a signer that
+  /// re-sends its challenge as a fresh event doesn't spam the log.
+  /// Bounded by session lifetime.
+  final Set<String> _seenAuthChallengeIds = {};
+
+  /// The wait window passed to [waitForConnection], kept so the timer can
+  /// be re-armed once when an auth challenge arrives.
+  Duration? _waitTimeout;
+
+  /// Whether the one allowed challenge-triggered timer restart happened.
+  /// Capped at one per session so injected challenges cannot extend the
+  /// wait indefinitely.
+  bool _challengeTimerRestarted = false;
+
   /// The since timestamp used for subscriptions, captured once at session start
   /// so reconnections use the same timestamp.
   int? _subscriptionSinceTimestamp;
@@ -244,16 +260,20 @@ class NostrConnectSession {
 
     _connectionCompleter = Completer<NostrConnectResult?>();
 
-    // Start timeout timer
+    _waitTimeout = timeout;
+    _armTimeoutTimer(timeout);
+
+    return _connectionCompleter!.future;
+  }
+
+  void _armTimeoutTimer(Duration timeout) {
     _timeoutTimer = Timer(timeout, () {
-      if (!_connectionCompleter!.isCompleted) {
+      if (!(_connectionCompleter?.isCompleted ?? true)) {
         logger('[NostrConnectSession] Connection timed out');
         _setState(NostrConnectState.timeout);
         _connectionCompleter!.complete(null);
       }
     });
-
-    return _connectionCompleter!.future;
   }
 
   /// Cancel the session.
@@ -467,6 +487,10 @@ class NostrConnectSession {
   }
 
   Future<void> _handleResponse(Event event) async {
+    if (!_seenEventIds.add(event.id)) {
+      return;
+    }
+
     // Decrypt the response
     final response = await NostrRemoteResponse.decrypt(
       event.content,
@@ -487,10 +511,12 @@ class NostrConnectSession {
 
     // Validate the secret per NIP-46. Accept ONLY an exact match;
     // "ack"/"connect" belong to the bunker:// flow's connect method
-    // and are not valid for nostrconnect://. Non-matching, non-error
-    // responses are dropped silently — treating them as terminal would
-    // let any pubkey on the listening relays DoS pairing by racing
-    // a junk response in front of the legitimate signer's reply.
+    // and are not valid for nostrconnect://. Everything else — junk
+    // responses, signer errors, auth_url challenges — is non-terminal:
+    // the sender is unauthenticated pre-secret, so treating any of it
+    // as terminal would let any pubkey on the listening relays DoS
+    // pairing by racing a junk response in front of the legitimate
+    // signer's reply (matches nostr-tools/NDK pairing behavior).
     final validation = validateConnectResponse(
       response: response,
       expectedSecret: _info?.optionalSecret,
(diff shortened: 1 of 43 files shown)


mobile/packages/nostr_sdk/lib/nip46/nostr_connect_session.dart
```diff
@@ -467,6 +487,10 @@ class NostrConnectSession {
   }
 
   Future<void> _handleResponse(Event event) async {
+    if (!_seenEventIds.add(event.id)) {
```
dcadenas, line 504: Validate `event.isValid` before adding `event.id` to `_seenEventIds`. A relay can re-emit a tampered copy with the genuine response's id, so this insert can poison the cache before decryption fails; the authentic response from another relay is then dropped as a duplicate and pairing times out. Add a regression test that delivers the tampered same-id event first and the genuine event second, then verifies that the session still binds.

dcadenas (changes requested):
Validate each event's id before recording it in the replay-dedup cache. A tampered relay copy can otherwise reserve the genuine response's id, fail decryption, and cause the valid response from another relay to be discarded as a duplicate, leaving NIP-46 sign-in to time out. Add the integrity guard and a regression test before merging.

mobile/packages/nostr_sdk/lib/nip46/nostr_connect_session.dart
realmeylisdev: Done in 85ce5a9e53. `_handleResponse` now drops any event failing `event.isValid` (NIP-01 id ≠ SHA-256 of its content) before the `_seenEventIds` insert, so a tampered same-id copy can no longer reserve the genuine id. Added a regression test that pushes the tampered same-id copy first, then the genuine secret on another relay, and asserts the session still binds — mutant-verified (neutering the guard makes it fail).

mbradley (commented):
dcadenas's id-integrity catch is the right one, and the fix holds. Confirmed it first-hand: the `event.isValid` guard lands before the `_seenEventIds` insert, so a tampered copy can't reserve the genuine response's id, and neutering the guard turns the new regression test red (mutation-checked, not green-on-green).

Ran an independent adversarial pass over the rest of the security-critical pairing path too: the `validateConnectResponse` ordering (an errored response can't bind ahead of the secret compare), the once-only timer restart and its `_waitTimeout!` null-safety, the late-match guard, and the async dedup under concurrent relay delivery. Nothing new surfaced. l10n removal is clean across all locales and the failure-message switch stays exhaustive.

Leaving the re-approve to you since it's your thread.

dcadenas (approved):
🔐 auth challenges now stay in the pairing flow

Issue #4341, opened 2026-05-13
epic(video-editor): consolidate creation, editing, and publishing flow improvements


NotThatKindOfDrLiz:
## Summary

Creation/editing/publishing work is currently fragmented across clip editing, transforms, camera capabilities, audio/voiceover, and publish/edit flow polish. This should be coordinated under one parent issue.

This issue is the source of truth for active video editor and creation-flow work.

## Problem

Current problems include:
- many editor-related feature asks with overlapping UX surface
- weak coordination between recording, editing, and publish flow improvements
- creation-flow polish arriving as disconnected requests instead of a sequenced area

## Scope

- clip editing
- video transforms and adjustment tools
- camera/editor capability improvements
- publish/edit flow consistency
- editor UX and polish improvements

## Candidate Related Work

- `#4185`
- `#3884`
- `#2050`
- `#1923`
- `#1512`
- `#1734`
- `#1607`
- `#2241`

Potential future mapping:
- post-merge follow-ups from current edit/publish redesign work
- additional editor/audio/overlay asks that fit this flow

## Acceptance Criteria

- creation/editor backlog is grouped under a coherent parent issue
- overlapping editor asks can be sequenced more intentionally
- future editor work has a clear coordination home

## Done When

- editor/creation work is easier to prioritize as one product area
- the biggest editor improvements are decomposed into concrete child issues

---

## Assignee
`@hm21`

## Weekly Priority
`P2`

## Weekly Priority Meaning
- `P1` means top priority for the week.
- `P2` means complete this week after P1 priorities are finished.

## Execution Order
1. core editor capability gaps
2. publish/edit flow consistency
3. residual creation-flow polish

## Subgroup
- `@hm21` — DRI
- `@Chardot` — UX/design alignment
- `@realmeylisdev` — editor/state support

## Coordination
The listed subgroup must review scope, sequencing, and ownership before implementation starts. Coordination may be synchronous or asynchronous. Do not wait for central coordination if the subgroup can resolve the question directly. Escalate only when blocked or when a product or UX decision is genuinely needed.

## Agent Execution Notes
Agents working from this issue must:
- treat this issue as the coordination source of truth for editor/creation work
- distinguish core creation-flow work from unrelated feed/profile issues
- avoid spawning separate top-level issues for every editor affordance unless they are truly independent initiatives

## Engineering Standard
This work must not create new technical debt.

hm21:
[...]

PR #3246, opened 2026-04-22
refactor(video-editor): remove the deprecated clip-editor

hm21:
Closes #3247

## Description

This PR removes the old deprecated clip-editor.

**Related Issue:** Closes #3147

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [x] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #6155, opened 2026-07-17
feat(video-editor): save a trimmed clip to the clip library (#5322)

### LINKED ISSUE #5322: feat(video-editor): save trimmed segments to clips
### What would you like?

Add back the ability to save a trimmed portion of a video as a separate clip from within the editor. This would allow users to select only the section they need and insert it into another project without adding the entire source video.

It makes editing much easier when combining videos. Users can reuse specific moments from longer videos without needing to import the full video, which can be too long for the project or make editing more difficult.

### When would you use this?

When creating a video that includes a short segment from another longer video. For example, trimming a 10-second section from a 5-minute video and inserting only that section into a new project.

### Anything else?

requested on discord [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1517087825609097277)

hm21:
## Description

Brings back the editor-side **Save clip** that #3246 removed with the deprecated clip-editor. Selecting a clip on the timeline and tapping Save re-encodes its trimmed section — with its edits and overlays baked in — into a standalone clip in the clip library, so a moment from a long video can be reused in an unrelated project without importing the whole source.

The library plumbing survived the earlier removal (`ClipManagerNotifier.saveClipToLibrary` is still live for the recorder), as did the l10n keys — already translated in all 18 locales and unused since — so no new translation work.

### Re-encode, not a trim-based cut

Unlike the trim-based split in #6134 (both halves share the source file), a library clip has to **outlive** the session that produced it and be dropped into an unrelated project. Sharing the source file would mean carrying a 5-minute video around to reuse ten seconds of it — the opposite of what the issue asks for. So this re-encodes: the output holds only the selected section, and the returned clip carries no residual trim/speed/volume/reverse — the render pipeline bakes all of it in. It runs uncapped (`maxOutputDuration: null`) because the ~6.3s limit belongs to the final export, not to a clip the user can still trim.

### Overlays are baked in too

Whatever sits over the clip — text, stickers, drawings, filters, tune, blur — is baked in, so the saved clip looks the way it looked on the timeline. Overlay time windows are authored against the *editor timeline* and composited onto the whole composition after concat, so `EditorOverlaySnapshot.windowedTo` slices them down to the clip's own span: overlays that never covered it are dropped, overlays that partly covered it are clamped, and the survivors are rebased to start at zero. The editor only hands over `CompleteParameters` on Done, so the widget layer captures the overlay state on demand via a new `ProImageEditorState.captureOverlaySnapshot()`.

A layer at 3s–4s of the timeline, saved from a clip that occupies 2s–6s, lands at 1s–2s of the standalone clip; the same layer, saved from a clip at 0s–2s, is dropped.

### Layering

- `UI` (`TimelineClipControls`) captures the overlay snapshot off the editor and dispatches an event; bails if the editor route is already gone, so a save can never silently drop the text/filters the user was looking at.
- `BLoC` (`ClipEditorBloc`) windows the overlays to the clip's slice, drives the re-encode through injectable seams, and persists via a callback (the library sits behind Riverpod, which the BLoC can't reach — same transition seam that brings the clip list in). Timeline is never mutated.
- `Service` (`VideoEditorClipLibrarySaveService`) flattens one clip to a fresh documents-dir file, closest precedent being `VideoEditorMergeService`.

Also extends the `renderVideoOverride` test seam with `maxOutputDuration`: without it no test can see whether a render is capped, and a lost `maxOutputDuration: null` would silently truncate a saved section at 6.3s — exactly the case the issue is about. It makes the merge service's uncapped guarantee assertable too.

**Related Issue:** Closes #5322 (part of #4341)

## Out of Scope

- The main editor's global crop/rotate/flip is intentionally not re-applied — a clip's own spatial transform is already baked into its file, and the aspect-ratio crop comes from `targetAspectRatio` inside the render; applying geometry twice would be wrong.
- Device verification of the actual re-encode (export + library round-trip) is still pending — see the manual test plan below.

## Verification

- `flutter analyze lib test integration_test` → No issues found
- `flutter test` across `blocs/video_editor`, `widgets/video_editor`, `services/video_editor`, `models/video_editor`, `services/video_editor_render_service_test.dart` → all green (1406)
- `dart format` clean; untested-services floor unchanged
- Windowing math, timeline-position math, the uncapped guarantee, and the null-editor bail each mutation-checked (every guard's test goes red when the guard is removed)

**Manual test plan (pending, needs a device):**
1. Trim a long clip to ~10s → Save clip → confirm the library clip contains only the 10s and its trim handles can't be pulled back open.
2. Put a text layer + filter over **clip 2** of a multi-clip timeline → save clip 2 → the saved clip shows the text at the right second with the filter applied.
3. Save **clip 1** → the same text does **not** appear (it was never over clip 1).
4. Leave the editor mid-re-encode → confirm no crash and the save is abandoned with the session.

## Type of Change

### DIFF
--- a/mobile/lib/blocs/video_editor/clip_editor/clip_editor_bloc.dart
+++ b/mobile/lib/blocs/video_editor/clip_editor/clip_editor_bloc.dart
@@ -4,8 +4,10 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:models/models.dart' show AudioEvent;
 import 'package:openvine/constants/video_editor_timeline_constants.dart';
 import 'package:openvine/models/divine_video_clip.dart';
+import 'package:openvine/models/video_editor/editor_overlay_snapshot.dart';
 import 'package:openvine/observability/reportable_error.dart';
 import 'package:openvine/services/audio_extraction_service.dart';
+import 'package:openvine/services/video_editor/video_editor_clip_library_save_service.dart';
 import 'package:openvine/services/video_editor/video_editor_merge_service.dart';
 import 'package:openvine/services/video_editor/video_editor_reverse_service.dart';
 import 'package:openvine/services/video_editor/video_editor_split_service.dart';
@@ -59,6 +61,27 @@ typedef MergeClipsFn =
       required String renderId,
     });
 
+/// Function signature matching
+/// [VideoEditorClipLibrarySaveService.flattenClipForLibrary], used as an
+/// injectable seam so tests can swap in a pure-Dart fake that does not touch
+/// `path_provider` or `pro_video_editor` plugins.
+typedef FlattenClipForLibraryFn =
+    Future<DivineVideoClip?> Function({
+      required DivineVideoClip clip,
+      required String renderId,
+      EditorOverlaySnapshot? overlays,
+    });
+
+/// Persists an already-flattened clip to the device's clip library, returning
+/// whether it was stored.
+///
+/// Wired by the widget layer to `ClipManagerNotifier.saveClipToLibrary`. The
+/// library lives behind a Riverpod provider this BLoC cannot reach, so the
+/// dependency arrives as a callback — the same transition seam that already
+/// brings the clip list in (see [ClipEditorBloc]).
+typedef SaveClipToLibraryFn =
+    Future<bool> Function({required DivineVideoClip clip});
+
 /// BLoC for managing video clip editor state.
 ///
 /// Owns a local copy of the clip list so that all mutations (add, remove,
@@ -74,11 +97,13 @@ typedef MergeClipsFn =
 class ClipEditorBloc extends Bloc<ClipEditorEvent, ClipEditorState> {
   ClipEditorBloc({
     required this.onFinalClipInvalidated,
+    required SaveClipToLibraryFn saveClipToLibrary,
     AudioExtractionService? audioExtractionService,
     SplitClipFn? splitClip,
     ReverseClipFn? reverseClip,
     TransformClipFn? transformClip,
     MergeClipsFn? mergeClips,
+    FlattenClipForLibraryFn? flattenClipForLibrary,
   }) : _audioExtractionService =
            audioExtractionService ?? AudioExtractionService(),
        _splitClip = splitClip ?? VideoEditorSplitService.splitClip,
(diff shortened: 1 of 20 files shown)


rabble (approved):
**Reviewed with full checkout + verification.** ✅ Approve. Well-engineered, thoroughly tested, closely follows the repo's architecture / BLoC / error-handling / l10n / a11y rules.

Defended in depth against the double-save race: `droppable()` transformer + global `isSaveToLibraryDisabled` greying Save on all clips during any save + a `bloc.isClosed` re-check after the frame-costing `captureOverlaySnapshot()` await, with `clipId` captured before the await and bound to the request. Orphan-file cleanup is complete — the persistent documents-dir render is deleted on every non-landing path (render discarded, library write rejected, exception), with `keepThumbnailPath` protecting a shared source thumbnail; covered by the `cleanupFlattenedClip` test group.

Contract checks pass:
- No error strings/exception objects in state — outcomes are identity-compared sealed result classes (`ClipLibrarySaveSuccess`/`Failure`/`Discarded`). `addError` wraps only `StateError`/`TypeError`/`RangeError` in `Reportable` per the decision matrix; expected render/IO nulls stay out of Crashlytics.
- Render logic lives in the new service, not the widget; the widget only captures the overlay snapshot (only the editor can produce it) and dispatches.
- 4 l10n keys already scaffolded on main and consumed via `context.l10n` (PR doesn't touch ARB — no consistency-gate concern). Save button has a distinct `semanticLabel`; uses `DivineIcon(.save)` / `DivineSnackbarContainer`; new widgets are classes.

The `EditorOverlaySnapshot.windowedTo` timeline-slicing (rebase overlays to the clip's slice, drop clamped-away enter/leave animations) is carefully reasoned and unit-tested.

Verification (local, PR head — 20 files):
- `flutter analyze lib/blocs/video_editor lib/services/video_editor lib/widgets/video_editor lib/models/video_editor lib/extensions/...` → No issues found!
- `flutter test` clip_editor bloc + service + models + timeline controls + scaffold → 348 tests passed (no "Some tests failed"). Includes the new scaffold snackbar tests (save success / fail / discarded-ignored).
- `check_untested_services_floor.sh` → OK, no new entries (new `video_editor_clip_library_save_service.dart` ships its same-named test).

Non-blocking note: this also removes the now-dead `splitNativeVideoToFile` (~60 lines) + `_splitWatchdogTimeout` — verified genuinely orphaned on origin/main (zero callers), so removing it is repo-mandated hygiene, just tangential to the feature. No rebase landmine.

PR #5755, opened 2026-07-02
fix(test): defuse #5340 re-bundled integration files, exception-safe restores

realmeylisdev:
## Description

**Related Issue:** Relates to #5738 (first of two PRs; the shared-helper/guard core lands separately)

File-level `@Tags` annotations are **dropped** inside the generated VGV bundle: `test/.test_optimizer.dart` imports every `*_test.dart` under `test/` and wraps each file's `main()` in a plain `group(...)`, and package:test parses suite-level tags only from the entry-point file. So `--exclude-tags integration` cannot exclude a bundled file — verified against very_good_cli source (1.1.1/1.2.0/1.3.0 identical) and empirically with the exact CI flags.

PR #5340 removed `skip_very_good_optimization` from 16 files on the opposite premise ("the integration tag keeps this test out of the merged VGV optimizer run"), so all 16 have been silently running inside the merged CI suite since 2026-06-19. The concrete hazard: `thumbnail_url_preservation_test.dart`'s `setUpAll` runs `RealIntegrationTestHelper.setupTestEnvironment()`, which replaces the shared `flutter_secure_storage` handler installed by `test_setup.dart` with a degraded one (write→noop, read→null) and never restores it — the same channel and mechanism as the #5713→#5725 `MissingPluginException` incident, reproduced deterministically in a merged mini-bundle (a later ambient write-then-read suite reads `null`). It is unexploded on `main` only because the one suite that depended on the ambient handler was deleted with the NostrKeyManager retirement.

Changes:

- **`RealIntegrationTestHelper` restores the app-harness defaults** after the calling scope completes (`addTearDown` from a `setUpAll` registers a group-scoped `tearDownAll`), clears its two helper-local channels, and resets its latch. New regression test proves the ambient roundtrip recovers after the helper's group.
- **`relay_subscription_acceptance_test.dart` gets its dual tag back** (`skip_very_good_optimization` + `integration`, with the `// Permanent:` rationale #5340 deleted): its `setUpAll` opens a real `Socket.connect` probe to `localhost:47777` on every CI run. Its tests already runtime-skip in CI, so this changes no coverage. Baseline 15→16.
- **The other 14 files drop the dead-letter `integration` tag** and the false comments. No behavior change — they run bundled either way; now the code says so.
- **New CI gate** (vgv-tag-gate job): fails any file-level `@Tags(['integration'])` without `skip_very_good_optimization`, so the #5340 regression class cannot recur.
[...]

Issue #4239, opened 2026-05-10
task(ci): fix mobile_service_integration_tests workflow so e2e tests actually run


realmeylisdev:
## Context

Discovered during review of #4234 (cc @hm21).
`.github/workflows/mobile_service_integration_tests.yaml` exists and runs `--tags service`, but is structurally broken on three independent axes — even with a correctly-tagged `patrolTest`, no test will execute.

## Problems

1. **Wrong runner**: the step runs `flutter test integration_test/e2e/ --tags service`. Patrol tests need `patrol test` (which orchestrates the native automator + device bootstrap). `flutter test` against a `patrolTest` body either silently no-ops or errors out depending on the patrol version.
2. **No device**: `runs-on: ubuntu-latest` with no Android emulator (e.g. `reactivecircus/android-emulator-runner`) and no iOS simulator. Patrol's native automation requires a booted device for the `Selector(...)` taps and `platformAutomator` actions.
3. **Trigger**: `on: workflow_dispatch` only — never fires from `push` / `pull_request`, so even a fixed workflow would not gate PRs without a manual run.

Combined effect: the `--tags service` filter is moot today because the workflow can't successfully run any patrol test. The four existing tests in `mobile/integration_test/e2e/` (`c2pa_init_test.dart`, `deleted_video_visible_to_other_users_test.dart`, `dm_optimistic_survives_watch_race_test.dart`, `video_creation_flow_test.dart`) are also missing the `service` tag — file a tag-pass once the workflow runs.

## Proposed fix (for whoever picks this up)

- Switch the `Run service integration tests` step to `patrol test --target integration_test/e2e/ --tags service` (or whatever flag set matches the project's patrol version).
- Add `reactivecircus/android-emulator-runner@v2` (or equivalent macOS-runner + iOS simulator setup) to provision a device before the test step.
- Add `pull_request` and/or `push` triggers, scoped to paths that affect e2e behaviour.
- Tag the four existing tests in `mobile/integration_test/e2e/` with `tags: ['service']` once the workflow actually executes them — this lets the `--tags service` filter behave as documented.

## Acceptance

- A PR that touches code under e2e coverage triggers the workflow.
- The workflow boots a device, runs `patrol test`, and the four (now five, post-#4234) tests in `integration_test/e2e/` all execute.
- Failing the new test in #4234 from a regression of the watch-stream race actually fails the workflow.

## Out of scope

[...]

PR #6156, opened 2026-07-17
test(integration): exception-safe ErrorWidget.builder/onError restores (#5839)

### LINKED ISSUE #5839: test: convert exception-unsafe ErrorWidget.builder/onError restores in integration_test
## Context

23 `ErrorWidget.builder` save/restore pairs across 20 files under `mobile/integration_test` restore the builder **inline at the end of the test body**, paired with same-shaped inline `FlutterError.onError` restores. If an earlier `expect` throws, the restore line never runs and the override leaks into later tests in the same file.

`mobile/integration_test` runs one isolate per file (it is NOT under the VGV optimizer), so the blast radius is within-file only — but the leak is still a real, seed-independent order dependency inside each file.

## Scope

Convert the inline end-of-body restores to `addTearDown(...)` so they run even when the test body throws. Purely mechanical; no behavior change on the happy path.

Related: #5738.

realmeylisdev:
## Description

The ~19 `mobile/integration_test` suites saved `ErrorWidget.builder` and
`FlutterError.onError` at test-body start and restored them **inline at the end
of the body**. If an earlier `expect` threw, the restore never ran, leaking the
override into a later test in the same file. Follow-up carved out of #5755's
"Out of Scope".

The correct fix is **asymmetric**, because of a framework constraint:

- **`ErrorWidget.builder` must be restored inline.** flutter_test's
  `_verifyErrorWidgetBuilderUnset` runs at end-of-body **before** any
  `addTearDown` callback (and is skipped once the body has already thrown), so an
  `addTearDown`-only builder restore fails the happy path with *"The value of
  ErrorWidget.builder was changed by the test."* (verified empirically). The
  tests therefore keep the inline restore (the "belt") **and** add an
  `addTearDown` restore (the "suspenders") that covers the throw path.
- **`FlutterError.onError` moves to `addTearDown` (teardown-only).** The binding
  resets `onError` per test in `postTest()`, so a leak there is already harmless;
  this matches the already-shipped `video_creation_flow_test.dart`.

Changes:

- Belt-and-suspenders conversion across the 19 files (16 standard; plus
  `signup_validation`'s raw `FlutterError.onError =` and `thumbnail`'s raw
  `ErrorWidget.builder =` normalized to the `test_setup` helpers).
- **Regression test** (`test/integration_test_error_restore_contract_test.dart`,
  non-patrol so it runs in the normal suite) pinning the contract — it fails if
  the inline builder restore is ever dropped.
- **CI guard** (`scripts/check_integration_test_error_restore_safety.sh`, wired
  into `mobile_ci.yaml` + the pre-push hook): bans raw `ErrorWidget.builder=` /
  `FlutterError.onError=` outside `helpers/`, and requires an `addTearDown`
  restore wherever `saveErrorWidgetBuilder()` / `suppressSetStateErrors()` is
  used.

Note on impact: within-file blast radius only (integration_test runs one isolate
per file), and the `ErrorWidget.builder` leak is self-neutralizing (the framework
re-baselines per test and `app.main` re-installs its builder), so this lands as
test hygiene + regression-proofing rather than a fix for an observed failure.

**Related Issue:** Closes #5839

## Out of Scope

- The `mobile/integration_test` patrol lane itself (patrol hard-gates plain
  `flutter test`; standing up the `patrol test` lane is #4239).
- A **pre-existing app race** surfaced while verifying:
  `_ClassicVinesTabState._refreshClassics` (`lib/widgets/classic_vines_tab.dart:145`)
  calls `ref.invalidate` after the widget is disposed, which makes
  `perf/feed_ttff_test.dart` flaky under throttled-network teardown timing.
  Unrelated to this change (it flakes on `origin/main` too) — worth its own issue.
- A pre-existing bash-3.2 `PATROL_EXTRA_ARGS[@]: unbound variable` bug in
  `local_stack/profile.sh`.

## Verification

- `flutter analyze lib test integration_test` → No issues found (all 19 compile).
- `dart format` → 0 changes.
- Regression test → 2/2 pass (also green in the pre-push hook).
- Guard → passes on the converted tree; fails on all three violation types
  (raw assignment / save-without-teardown / suppress-without-teardown).
- **Full backend-backed patrol run** (Android emulator + live `local_stack`):
  `patrol test --target integration_test/auth/signup_validation_test.dart
  --dart-define=DEFAULT_ENV=LOCAL` → `All tests passed!` — exercises the
  raw-onError→helper normalization + belt-and-suspenders end-to-end.
- Standard-pattern live run (`perf/feed_ttff_test.dart`): passes on retry on both
  this branch and `origin/main`; the intermittent failure is the pre-existing
  ref-after-dispose race noted in Out of Scope, not this change.

## Type of Change

- [x] 🧹 Code refactor

### DIFF
--- a/.github/workflows/mobile_ci.yaml
+++ b/.github/workflows/mobile_ci.yaml
@@ -99,6 +99,8 @@ jobs:
         run: bash scripts/check_http_overrides_isolation.sh
       - name: 🧬 Verify no untagged test leaks a process-global singleton
         run: bash scripts/check_process_global_mutations.sh
+      - name: 🧯 Verify integration_test error-handler restores are exception-safe
+        run: bash scripts/check_integration_test_error_restore_safety.sh
       - name: 🔌 Verify no new raw shared-channel installs (heal-and-blame ratchet)
         run: |
           # LIST ratchet vs origin/main (fails closed if the base baseline can't
--- a/mobile/integration_test/auth/auth_journey_test.dart
+++ b/mobile/integration_test/auth/auth_journey_test.dart
@@ -32,7 +32,9 @@ void main() {
         final l10n = lookupAppLocalizations(const Locale('en'));
         // ── Setup: suppress non-critical errors ──
         final originalOnError = suppressSetStateErrors();
+        addTearDown(() => restoreErrorHandler(originalOnError));
         final originalErrorBuilder = saveErrorWidgetBuilder();
+        addTearDown(() => restoreErrorWidgetBuilder(originalErrorBuilder));
 
         // Pre-enable semantics so the handle is disposed at the right time.
         // find.bySemanticsLabel() calls ensureSemantics() implicitly; if we
@@ -790,7 +792,8 @@ void main() {
         // Drain pending errors before restoring handlers.
         semanticsHandle.dispose();
         drainAsyncErrors(tester);
-        restoreErrorHandler(originalOnError);
+        // Inline restore is required by the framework's end-of-body
+        // ErrorWidget.builder check; the addTearDown above covers throws.
         restoreErrorWidgetBuilder(originalErrorBuilder);
       },
       timeout: const Timeout(Duration(minutes: 7)),
--- a/mobile/integration_test/auth/follow_reopen_persistence_test.dart
+++ b/mobile/integration_test/auth/follow_reopen_persistence_test.dart
@@ -24,7 +24,9 @@ void main() {
         const password = 'TestPass123!';
 
         final originalOnError = suppressSetStateErrors();
+        addTearDown(() => restoreErrorHandler(originalOnError));
         final originalErrorBuilder = saveErrorWidgetBuilder();
+        addTearDown(() => restoreErrorWidgetBuilder(originalErrorBuilder));
         final semanticsHandle = tester.ensureSemantics();
 
         launchAppGuarded(app.main);
@@ -127,7 +129,8 @@ void main() {
 
         semanticsHandle.dispose();
         drainAsyncErrors(tester);
-        restoreErrorHandler(originalOnError);
+        // Inline restore is required by the framework's end-of-body
+        // ErrorWidget.builder check; the addTearDown above covers throws.
         restoreErrorWidgetBuilder(originalErrorBuilder);
       },
       timeout: const Timeout(Duration(minutes: 4)),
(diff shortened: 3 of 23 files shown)


rabble (approved):
**Reviewed with full checkout + verification.** ✅ Approve. Solid, well-reasoned defensive-test work — the guard genuinely prevents the bug class and every overriding integration test was converted.

The 20 integration_test suites each override `ErrorWidget.builder` + `FlutterError.onError` per test; if an early `expect` throws, the inline restore never runs and the override leaks into the next test in the same file (#5839). This makes restores exception-safe via `addTearDown`, routes overrides through `helpers/test_setup.dart`, adds a presence-based guard script + a contract test, and wires both into CI + the pre-push hook.

The asymmetry is the load-bearing part and it's correct:
- `ErrorWidget.builder` — restored both inline (belt) and via `addTearDown` (suspenders). flutter_test's `_verifyErrorWidgetBuilderUnset` runs at end-of-body **before** `addTearDown`, so inline is required on the happy path; addTearDown covers the throw path.
- `FlutterError.onError` — `addTearDown`-only is sufficient (the binding resets onError in `postTest`), so the inline onError restores were correctly removed.

Verification (local):
- `flutter analyze integration_test test/integration_test_error_restore_contract_test.dart` → No issues found!
- Guard script happy path → exit 0. **Fail-closed proven:** added two scratch files (raw assignment + save-without-teardown) — guard flagged both with correct messages and exit 1; passed again after cleanup.
- Contract test → 2/2 pass, and **proven able to fail** — removing the inline restore makes test 1 go red ("The value of ErrorWidget.builder was changed by the test"), confirming the framework-timing claim (not a tautology).
- Coverage completeness: grep of `integration_test/` shows the only raw `ErrorWidget.builder=`/`FlutterError.onError=` assignments left are inside the allowlisted helper. All 20 suites use the helpers + addTearDown.
- CI wiring correct (new step in the `generated-files` job, `working-directory: mobile/`); hook block mirrors CI, gated on changed `integration_test/*.dart`.

Non-blocking nits: (1) guard Rule 2 is presence-based per file, not per-test — a multi-test file where one test registers the teardown and another calls `saveErrorWidgetBuilder()` without one would pass falsely; contrived and acknowledged in the script's own comment; (2) Rule 1's grep isn't comment-aware, but the refactored comments avoid `=`.

PR #6158, opened 2026-07-17
feat(account): delete-account identity + username confirmation (#6137)

### LINKED ISSUE #6137: feat(account): show which account is being deleted and require account-specific delete confirmation
## What would you like?

Before any account deletion, show the user exactly which identity is about
to be erased, and require an account-specific confirmation instead of typing
the generic word "DELETE".

Concretely, on the final delete-confirmation surface:
1. Show the active account's identity: avatar + display name + their username
   — whatever form it takes: a Divine handle like `@mjb.divine.video` or an
   external one like `mjb@nos.social`. When the account has no username, show
   a shortened npub instead. Mirror the claimed-handle identity pattern from
   `_AccountSwitchTile` (`settings_screen.dart`) — not the verification-gated
   `_AccountHeaderProfile`.
2. Make the confirm gate account-specific: when the account has a username,
   show it and require the user to re-type **exactly what is shown** (the full
   handle); when it doesn't, fall back to typing "DELETE".
3. Correct the deletion copy, which currently overstates the guarantee, to
   match the app's accurate video-deletion convention.

## How would this be useful?

Deletion acts on the currently signed-in account
(`AccountDeletionService.deleteAccount()` reads `currentPublicKeyHex`) and is
a permanent, network-wide NIP-62 vanish. But nothing in the delete path names
that account:

- The only place the app shows who you're signed in as is the root Settings
  header (`_AccountHeaderProfile`, `settings_screen.dart`). Delete Account and
  Data lives two pushes deeper (root Settings -> Nostr Settings -> Danger
  Zone), so that header is off-screen by the time you're deciding.
- The final confirmation dialog (`showDeleteAllContentWarningDialog`,
  `delete_account_dialog.dart`) shows only generic copy plus a hardcoded
  "DELETE" field: no avatar, name, username, or npub.

With account switching (a shipped experimental feature,
`FeatureFlag.accountSwitching`) or the same login used across similar-looking
devices, a user can confidently wipe the wrong identity off the network in an
action that can't be undone. Re-typing the account's own username forces them
to read and match the identity before the deletion goes through.

## User stories

- As a user with more than one identity, when I open Delete Account and Data
  I want the confirmation to clearly show which account (avatar, name,
  username/npub) is about to be permanently erased, so I don't wipe the wrong
  one.
- As a user confirming an irreversible deletion, I want the confirmation to
  require something specific to *this* account rather than the generic word
  "DELETE", so muscle memory or a mis-selected account can't carry me through
  deleting an identity I didn't mean to touch.

## Implementation notes

Full design: `docs/superpowers/specs/2026-07-16-delete-account-identity-confirmation-design.md`.

- **Terminology:** technically these identifiers are NIP-05s; in the UI we use
  "username" for usability (the app already does this —
  `profileSetupUsernameLabel` = "Username (Optional)"). Use "username" in all
  user-facing copy; reserve "NIP-05" for code.
- **Token:** the string shown, typed, and accepted is the full
  `UserProfile.displayNip05` (`@name.divine.video` for Divine, `name@domain`
  for external). Match with two leniencies only — leading `@` optional and
  case-insensitive; no bare/short form. When `displayNip05` is null, fall back
  to "DELETE".
- **Handle source is async-only.** `authService.currentProfile` never carries
  `nip05` (it is only ever built with npub/pubkey/displayName); the handle
  comes solely from the rich `models.UserProfile` via
  `userProfileReactiveProvider`, which is async. The **npub is the only
  synchronous identifier.** So on tap we resolve the profile behind a brief,
  bounded progress overlay before opening the dialog; only a genuinely
  unloadable profile degrades to npub + DELETE.
- **Claimed, not verified:** use the account's own claimed handle
  (`_AccountSwitchTile`'s `displayNip05 ?? truncateNpub` pattern); do not gate
  on `nip05VerificationProvider`.
- **Bind the deletion to the confirmed account:** capture the pubkey when the
  dialog opens; thread it through so the identity shown, the token required,
  and the deletion target are one pubkey. `deleteAccount` verifies the current
  account still matches and aborts on mismatch rather than deleting.
- **Copy accuracy:** the current account-delete strings promise "PERMANENTLY
  delete ALL content from Nostr relays," which NIP-09/NIP-62 can't guarantee
  for independent relays. Align to the video-deletion convention already in the
  app (`videoGridDeleteConfirmMessage`, `shareMenuDeleteWarning`): "…delete
  your account and content from Divine, and request removal from other Nostr
  relays. Some copies may remain." Fixes both the dialog body and the tile
  subtitle (`nostrSettingsDeleteAccountSubtitle`).

Near-term this hardens the current dialog flow; #6127's full-screen flow
should carry the same guarantee when it lands (see Relationships).

## Acceptance criteria

- [ ] The final delete-confirmation surface shows the active account's avatar,
      display name, and username (or shortened npub if none).
- [ ] The identity shown, the "type this" target, and the accepted input are
      the **same** full-handle string (leading `@` optional, case-insensitive;
      no bare/short form).
- [ ] When the account has a username, deletion requires re-typing it; the
      confirm button stays disabled until it matches. Works for both a Divine
      username (`@name.divine.video`) and an external one (`name@domain`).
- [ ] When the account has no username, the confirm gate falls back to
      "DELETE".
- [ ] The profile is resolved before the dialog opens (brief bounded wait);
      a genuinely unloadable profile degrades to npub + "DELETE" rather than
      hanging.
- [ ] The deletion is bound to the confirmed account: `deleteAccount` verifies
      the current account matches the confirmed pubkey and aborts on mismatch.
- [ ] The deletion copy (dialog body + danger-zone tile subtitle) no longer
      overstates removal from third-party relays.

## Testing

- Widget tests: Divine-username, external-username, and no-username accounts;
  confirm enable/disable keyed on matching the right token in each case, and
  the token derives from the active account, not a constant.
- Resolve-before-confirm: loaded, not-yet-loaded, and unloadable paths.
- Binding: `deleteAccount` aborts on `expectedPubkey != currentPublicKeyHex`
  and proceeds when they match.

## Relationships

- Related to #6127 — the reversible-deletion redesign lists "make it clear
  which account is being changed" and "bind every screen and signature to the
  exact account" as correctness requirements. This issue owns the user-facing
  consent guarantee (show which account + account-specific confirm)
  independently, so #6127 can focus on the reversible-deletion machinery and
  adopt it on its full-screen flow. Backend-independent, so it can land ahead
  of #6127 (which is blocked by divine-funnelcake#651).
- #6126 / PR #6138 (opt-in @divine.video username burn on delete) is a separate
  *feature axis* but shares the **same delete-modal code** — #6138 rewrites
  `showDeleteAllContentWarningDialog` and `executeAccountDeletion`. Sequencing:
  #6138 lands first, then this work rebases onto it and re-derives against the
  merged dialog (identity block + username gate coexisting with the burn
  checkbox; `executeAccountDeletion` fusing burn-first with the confirmed-pubkey
  verify-and-abort).

mbradley:
Closes #6137. Builds on #6138 (merged).

The "Delete Account and Data" confirmation now shows **which account** you're about to erase and asks you to re-type that account's username to confirm, so account switching (or the same login on a look-alike device) can't lead you to wipe the wrong identity off the network in an action that can't be undone.

## What changed
- **Identity on the confirmation.** Shows the active account's avatar, display name, and username (or a shortened npub if it has none), resolved before the dialog opens (bounded; degrades to npub + DELETE if the profile can't load).
- **Account-specific gate.** Type the account's username to confirm (`@mjb.divine.video`, or an external `name@domain`). What's shown is exactly what's accepted (leading `@` and case optional). Accounts with no username fall back to typing `DELETE`.
- **Bound to the confirmed account.** Deletion aborts before touching anything (including the burn) if the signed-in account changed since you confirmed.
- **Honest copy.** The confirmation body and danger-zone subtitle no longer overstate deletion; they match the app's existing video-deletion wording (delete from Divine; other relays and clients may keep copies).

Additive to #6138's opt-in @divine.video username burn. The burn checkbox and burn-first flow are untouched.

## i18n
The corrected English for `nostrSettingsDeleteAccountSubtitle` (a pre-existing translated key) leaves the 16 non-English locales on their prior wording until a re-translation pass, per the repo's l10n copy-alignment policy. The new keys are English-only and tracked in `_knownUntranslatedDebt`.

## Testing
`DeleteAccountConfirmation` unit tests (token derivation + matching: Divine, external, no-handle, empty, degenerate `@`); dialog identity block + username gate + prompt copy; guard aborts-on-mismatch and proceeds-on-match; real-tile tests that the resolved username shows and that a missing profile degrades to DELETE. `flutter analyze lib test integration_test` clean; #6138's burn tests still pass.

Spec + plan: `docs/superpowers/specs/2026-07-16-delete-account-identity-confirmation-design.md`, `docs/superpowers/plans/2026-07-17-delete-account-identity-confirmation.md`.


dcadenas (changes requested):
Bind the confirmed account through the entire deletion operation. The current check only samples `authService.currentPublicKeyHex` before awaited work, while `AccountDeletionService` later re-reads the live signer and can publish kind-5 and kind-62 events for a different account. This must be fixed before merge by passing `confirmedPubkey` into the service and revalidating it immediately before each signing and publish step, with a regression test for an account switch during deletion.

---

- **Confirmed account is checked once, not bound through to the deletion signing** ([mobile/lib/widgets/delete_account_dialog.dart:559](https://github.com/divinevideo/divine-mobile/blob/014cf3b15ec0459d04b996a8e4093bdf35b160e1/mobile/lib/widgets/delete_account_dialog.dart#L559))
  Pass `confirmedPubkey` into `AccountDeletionService.deleteAccount` here and make the service enforce it at signing and publish time. The guard at `delete_account_dialog.dart:485-499` runs before username release and relay queries, but this call drops the confirmed identity; the service then reads mutable `currentPublicKeyHex` again and can sign the kind-5 and network-wide kind-62 vanish for account B if the user switches from A to B during those awaits. Add checks immediately before each destructive signing and before publishing, abort on mismatch, and cover the account-flip path with a service-level test.

mobile/lib/widgets/delete_account_dialog.dart
```diff
@@ -407,6 +480,24 @@ Future<void> executeAccountDeletion({
   var burnCommitted = false;
 
   try {
+    // Bind to the confirmed account: if the signed-in account changed since the
+    // user confirmed, abort before burning or deleting anything.
+    if (confirmedPubkey != null &&
```
realmeylisdev, line 497: **Blocking** (confirming @dcadenas). This guard is the *only* place the confirmed account is enforced, and it's a one-time check. `deleteAccount` (line 559) isn't passed `confirmedPubkey`; the service re-reads the live signer — `_authService.currentPublicKeyHex` in `account_deletion_service.dart` (event fetch at :61, kind-5 sign at :231, kind-62 sign at :282). So if the signed-in identity flips between this check and signing — e.g. an account switch during the all-relay `queryEvents(limit: 10000)` window — the network-wide NIP-62 "request to vanish" gets signed for the wrong account, which can't be undone.

The burn network-error re-check at line 529 (`getUsernameByPubkey(currentPublicKeyHex)`) reads the live signer the same way; lower impact there, since it only picks the disclosure string and the name-server's NIP-98 ownership check still protects the burn itself.

Issue #6137's acceptance criterion ("`deleteAccount` verifies the current account matches the confirmed pubkey and aborts on mismatch") is unmet as written — the verification lives in the UI, not the service. Suggested fix: thread `confirmedPubkey` into `deleteAccount(expectedPubkey: ...)` and revalidate `currentPublicKeyHex == expectedPubkey` immediately before signing kind-5 and kind-62, then add the service-level flip regression test. The two dialog-level tests can't catch this today — one flips the account *before* the check, the other mocks `deleteAccount` — so the fix and its regression test have to land together.

mobile/lib/widgets/delete_account_dialog.dart
```diff
@@ -166,18 +170,27 @@ class _DeleteAllContentDialogState extends State<_DeleteAllContentDialog> {
...
           ),
           const SizedBox(height: 12),
-          const Text(
-            _requiredText,
-            style: TextStyle(
+          Text(
+            c.isUsernameConfirmation
+                ? context.l10n.deleteAccountConfirmUsernamePrompt
+                : context.l10n.deleteAccountConfirmDeletePrompt,
+            style: const TextStyle(color: VineTheme.whiteText, fontSize: 16),
```
realmeylisdev: Nit: raw `const TextStyle` on this new prompt (`VineTheme.bodyLargeFont()` is the Inter-16 equivalent per `ui_theming.md`). The identity header just above already uses `VineTheme.*Font()`. Low priority since the rest of this dialog predates the rule, but worth aligning the newly-added lines.

mobile/lib/widgets/delete_account_dialog.dart
```diff
@@ -407,6 +480,24 @@ Future<void> executeAccountDeletion({
...
+        authService.currentPublicKeyHex != confirmedPubkey) {
+      Log.warning(
+        'Deletion aborted: signed-in account changed since confirmation',
+        name: screenName,
+        category: LogCategory.auth,
+      );
+      dismissDialog();
+      if (context.mounted) {
+        ScaffoldMessenger.of(context).showSnackBar(
+          DivineSnackbarContainer.snackBar(accountChangedText, error: true),
```
realmeylisdev, line 507: Nit (a11y): this abort — and the success/failure snackbars below — are visual-only. Consider `SemanticsService.sendAnnouncement(View.of(context), text, Directionality.of(context))` so VoiceOver/TalkBack users hear the outcome (`accessibility.md` lists snackbar-only delete/error outcomes under Should-Fix). It's a pre-existing flow-wide gap, so ideally cover the whole outcome flow rather than just this new line.

mobile/lib/screens/settings/nostr_settings_screen.dart
```diff
@@ -261,13 +266,42 @@ class _DeleteAccountTile extends StatelessWidget {
...
+      overlay.dismiss();
+    }
+    if (!context.mounted) return;
+
+    final confirmation = DeleteAccountConfirmation(
+      pubkeyHex: pubkey,
+      displayName:
+          profile?.bestDisplayName ?? UserProfile.defaultDisplayNameFor(pubkey),
+      avatarUrl: profile?.picture,
+      handle: profile?.displayNip05,
```
realmeylisdev, line 304: Nit: the confirm-gate token derives from `profile.displayNip05`, but the opt-in burn targets `ownedUsername` (from `ownedDivineUsernameProvider`). For a user whose profile NIP-05 is external (e.g. `alice@example.com`) while they still own `@alice.divine.video`, the string they type to confirm differs from the handle being burned. Not misleading — the burn checkbox names its own target, and the account binding is separate — just flagging the identity-gate vs burn-target divergence in case it's worth unifying.

realmeylisdev (commented):
Formal review. Agreeing with @dcadenas's blocker: the confirmed account is enforced only by a one-time UI pre-check and is never threaded into `AccountDeletionService.deleteAccount`, which re-reads the live signer — so a mid-flight account switch can sign the network-wide NIP-62 vanish for the wrong identity, and issue #6137's binding acceptance criterion is unmet as written. Everything else I checked holds up: the copy correction is protocol-accurate against NIP-09/NIP-62, l10n is clean, the identity header's accessibility is right, and the async/overlay handling is safe. A few nits inline. Leaving this as Comment since the blocker is already open.

mobile/lib/widgets/delete_account_dialog.dart
mbradley: Fixed. The confirmed account is now bound through the whole operation, not just sampled here: `deleteAccount` takes the confirmed pubkey and re-checks the live signer immediately before each destructive sign and the network-wide publish, aborting before anything is signed on mismatch. A service-level regression test drives an account switch mid-deletion and asserts nothing gets signed. `3f51ea2` (binding), `512596712` (localized abort message). Thanks.

mobile/lib/widgets/delete_account_dialog.dart
mbradley: Done, `VineTheme.bodyLargeFont()` in `3f51ea2`. Thanks.

mobile/lib/widgets/delete_account_dialog.dart
mbradley: Done in `3f51ea2`. Covered the whole outcome flow rather than just this line: abort, burn-failed, server-failed, and success each announce via `SemanticsService.sendAnnouncement` now, with the abort announcement test-covered. Thanks.

mobile/lib/screens/settings/nostr_settings_screen.dart
mbradley: Kept the gate on the shown profile identity, since that is what confirms which account is being erased, and documented the deliberate split from the burn target in a comment (`3f51ea2`). `512596712` also makes the account-changed message consistent between the pre-check and the mid-flight paths. Thanks.

dcadenas (approved):
🔒 identity-bound deletion, ship it

mobile/lib/services/account_deletion_service.dart
```diff
@@ -278,6 +343,8 @@ class AccountDeletionService {
         category: LogCategory.system,
       );
 
+      // Immediately before signing the network-wide kind-62 vanish.
+      _assertSignerMatches(expectedPubkey);
```
realmeylisdev, line 347: nit, non-blocking: this kind-62 sign guard and the pre-publish guard at line 148 are defense-in-depth only — no test flips the signed-in account in either window, so removing either guard fails no test (the mid-flight service test aborts earlier at the kind-5 guard; the happy-path test matches throughout). A small service test that switches the signer right before the kind-62 sign would lock both in against future refactors.

realmeylisdev (approved):
Re-reviewed the identity-binding fix. The wrong-account NIP-62 vanish race is closed: `expectedPubkey` is threaded through the service and re-asserted synchronously right before each kind-5/kind-62 sign and the publish, `createAndSignEvent` captures the identity before its first await (so check-and-sign is atomic), and the signer-factory pubkey guard is the backstop. `flutter analyze` + delete-account tests green. One optional nit inline.

mobile/lib/widgets/delete_account_dialog.dart
realmeylisdev: Verified fixed. `expectedPubkey` is threaded through `deleteAccount` and re-asserted right before each kind-5/kind-62 sign and the publish; `createAndSignEvent` captures the identity before its first await, so check-and-sign is atomic, with the signer-factory pubkey guard as the backstop. The mid-flight-flip service test covers it. Resolving.

mobile/lib/widgets/delete_account_dialog.dart
realmeylisdev: Confirmed, `VineTheme.bodyLargeFont()`. Resolving.

mobile/lib/widgets/delete_account_dialog.dart
realmeylisdev: Confirmed. Every outcome announces via `SemanticsService.sendAnnouncement` and the abort is test-covered. Resolving.

mobile/lib/screens/settings/nostr_settings_screen.dart
realmeylisdev: Confirmed. The gate stays on the shown profile identity, with the intentional split from the burn target documented. Resolving.

NotThatKindOfDrLiz (approved):
Re-reviewed the identity-bound delete-account flow. The original wrong-account deletion race is closed, targeted tests pass locally, and CI is green.

PR #6169, opened 2026-07-18
fix(feed): ignore double-tap-to-like on your own video

### LINKED ISSUE #6170: fix(feed): ignore double-tap-to-like on your own video
Auto-created from PR #6169.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6169
**Branch:** `fix/double-tap-like-own-video`

hm21:
Closes #6170

## What

Double-tapping your **own** video no longer likes it. Previously, viewing your own video in your profile (or anywhere in the feed) and double-tapping registered a like.

The player semantics hint also no longer tells assistive-tech users to double-tap to like on their own videos, because that gesture is now intentionally ignored there.

## Why

The profile fullscreen viewer reuses the shared `FeedVideos` overlay — there is no separate profile player. The like button already enforces "you can't like your own video" (`isOwnVideo` → it opens the likers list instead of toggling), but the double-tap gesture handler never consulted `isOwnVideo`, so it bypassed that rule. Since it's the same widget, the bug reproduced from the profile as well as the main feed.

The decision is extracted into a pure `resolveDoubleTapLikeAction(...)` helper (mirroring the existing `content_warning_helpers.dart` pattern) so the rule is:

- **own video** → ignore (no like, no heart animation), consistent with the like button
- **content warning still blocking** → ignore
- **already liked** → replay the heart animation only (double-tap never unlikes)
- **otherwise** → like + animate

## Testing

The double-tap gesture lives only in the interactive overlay, which requires a real video controller with a rendered first frame — the widget-test harness disables that (`debugIsSupportedOverride = false`), which is why this gesture had no widget coverage. The extracted helper is the clean, can-fail seam: `double_tap_like_helpers_test.dart` covers the own-video regression plus every other branch. If the `isOwnVideo` guard were removed, the test goes red.

- `flutter test test/widgets/video_feed_item/double_tap_like_helpers_test.dart test/widgets/video_feed_item/feed_videos_test.dart` — 22/22 green
- `flutter analyze` — clean

### DIFF
--- a/mobile/lib/widgets/video_feed_item/double_tap_like_helpers.dart
+++ b/mobile/lib/widgets/video_feed_item/double_tap_like_helpers.dart
@@ -1,0 +1,37 @@
+// ABOUTME: Pure decision helper for the double-tap-to-like gesture on feed
+// ABOUTME: videos. Keeps the "you can't like your own video" rule testable in
+// ABOUTME: isolation from the FeedVideos overlay widget.
+
+/// The action a double-tap on a feed video should trigger.
+enum DoubleTapLikeAction {
+  /// Do nothing: the viewer owns the video (you can't like your own video), or
+  /// a content warning is still blocking interaction.
+  ignore,
+
+  /// Play the heart animation without changing the like state — the video is
+  /// already liked, so a double-tap re-affirms rather than unliking.
+  animateOnly,
+
+  /// Toggle the like on and play the heart animation.
+  likeAndAnimate,
+}
+
+/// Decides how a double-tap on a feed video should behave.
+///
+/// Mirrors the like button's rule that the owner cannot like their own video
+/// (see `LikeActionButton`): on the owner's own video the gesture is a no-op.
+/// A content warning that still blocks the video also suppresses the gesture.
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz:
Takeover follow-up pushed in `8d7f071ec`.

What changed:
- Omit the feed player "double tap to like" semantics hint on own videos, since double-tap is now intentionally ignored there.
- Added widget coverage for the own-video semantics path while preserving the existing other-video hint assertion.
- Updated the PR body so the testing notes match the current suites.

Verification:
- `flutter test test/widgets/video_feed_item/double_tap_like_helpers_test.dart test/widgets/video_feed_item/feed_videos_test.dart` passed locally.
- `flutter analyze` passed locally.
- GitHub checks are green: Analyze, Tests, Format, Generated Files, Build Flutter Web Preview, VGV Tag Gate, mergeability, semantic PR, linked issue, and CLA.

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up. The self-like guard is correct, the own-video semantics hint no longer advertises a disabled double-tap-like action, regression coverage is in place, and CI is green.

Issue #2144, opened 2026-03-11
fix: video visible on profile but cannot be played on production relay


hm21:
### Summary

Users report on discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1481214175387586581) that they can see a video on their profile but cannot play it, while the same video works on the staging relay. Note the users did not use the staging relay.

What i already know is that the `videoUrl` we receive here is incorrect:
[https://github.com/divinevideo/divine-mobile/blob/5f94c0f2daac3d6a8857d50cdd31218740fcdb65/mobile/lib/services/analytics_api_service.dart#L983-L986](https://github.com/divinevideo/divine-mobile/blob/5f94c0f2daac3d6a8857d50cdd31218740fcdb65/mobile/lib/services/analytics_api_service.dart#L983-L986)

On the staging relay the URL looks like this, which is correct:
`https://media.divine.video/9783bc5788da2fc2bab7600521ea923d873ef3379d18da1b865f513f76e1bf6f`

On the production relay the URL looks like this:
`/var/mobile/Containers/Data/Application/317479C4-6656-4198-A339-397F40DBF290/Library/Caches/openvine_video_cache/013baa30-08bb-11f1-8bed-5114e3255714.mp4`

### Environment

Any version and platform

### Steps to Reproduce

1. Open the app and navigate to the affected user's profile [here](https://miketv.divine.video/)
2. Locate the uploaded video.
3. Attempt to play the video.

### Actual Result

The video appears on the profile but cannot be played.

### Expected Result

The video should play normally when selected from the user's profile.

### Regression?

Unknown

PR #6173, opened 2026-07-18
fix(nostr): guard against spurious startup NostrClient rebuild and run the #2144 repair after auth restore

### LINKED ISSUE #1: 🏗️ [Infrastructure] Configure R2 Storage Buckets for Media Processing
## Overview
Set up and configure Cloudflare R2 storage buckets for the NostrVine backend to handle image frames and generated media.

## Acceptance Criteria
- [ ] Create `nostrvine-frames` bucket for storing individual image frames
- [ ] Create `nostrvine-media` bucket for storing generated GIFs/videos  
- [ ] Create `nostrvine-cache` bucket for caching processed content
- [ ] Configure CORS policies for cross-origin access from mobile app
- [ ] Set up bucket lifecycle policies for cost optimization
- [ ] Update wrangler.toml with correct bucket configurations
- [ ] Test bucket connectivity from Workers environment

## Technical Details
**Buckets Required:**
1. **nostrvine-frames**: Raw image frames uploaded from mobile app
2. **nostrvine-media**: Final processed GIFs and videos
3. **nostrvine-cache**: Temporary cache for frequently accessed content

**Configuration:**
- Enable CORS for mobile app domains
- Set up automatic cleanup policies (30-day retention for frames, 1-year for media)
- Configure appropriate access policies

## Files to Modify
- `backend/wrangler.jsonc` - Already configured, needs verification
- Add R2 bucket creation scripts or documentation

## Dependencies
- Cloudflare account with R2 enabled
- Wrangler CLI configured with proper permissions

## Priority
**High** - Required for core media processing functionality

## Labels
backend, infrastructure, cloudflare, storage

### LINKED ISSUE #2: 🎨 [Backend] Implement GIF Creation Engine with gifenc
## Overview
Implement the core GIF creation functionality using the gifenc library in Cloudflare Workers to convert image sequences into animated GIFs.

## Acceptance Criteria
- [ ] Install and configure `gifenc` library in Workers project
- [ ] Implement frame loading from R2 storage (FRAMES_BUCKET)
- [ ] Create GIF encoding pipeline with proper error handling
- [ ] Add support for custom frame delays and loop settings
- [ ] Implement memory management for large frame sequences (stay within 128MB limit)
- [ ] Add progress tracking for long-running operations
- [ ] Store generated GIFs in MEDIA_BUCKET
- [ ] Return proper response with GIF binary data or R2 URL

## Technical Implementation
**Core Function:**
```typescript
async function createGifFromFrames(
  frameKeys: string[], 
  options: GifOptions
): Promise<ArrayBuffer>
```

**Features:**
- Frame delay configuration (default 100ms)
- Color palette optimization
- Loop count settings
- Memory-efficient streaming processing
- Error recovery for corrupted frames

## Files to Create/Modify
- `backend/src/services/gif-encoder.ts` - Main GIF creation service
- `backend/src/utils/memory-management.ts` - Memory optimization utilities
- `backend/src/types/gif.ts` - Type definitions for GIF options
- `backend/src/handlers/create-gif.ts` - HTTP handler for GIF creation requests

## Technical Constraints
- Cloudflare Workers 128MB memory limit
- 30-second CPU time limit (5 minutes on paid plans)
- Process frames in batches to avoid memory issues
- Implement streaming patterns for large sequences

## Testing Requirements
- Unit tests for GIF encoding with sample frames
- Memory usage tests with various frame counts
- Performance benchmarks for different image sizes
- Error handling tests for malformed frames

## Dependencies
- Issue #1 (R2 Storage Setup) must be completed first
- `gifenc` npm package

## Priority
**Medium** - Core feature for backend functionality

## Labels
backend, feature, gif-processing, cloudflare-workers

### LINKED ISSUE #5909: fix(nostr): spurious full NostrClient rebuild when the RPC-upgrade auth nudge lands during startup relay connect
## Summary

On a Divine OAuth cold start, the background RPC-upgrade's `finally` block re-emits the current auth state unconditionally. If that emission lands while the startup `NostrClient` is still connecting relays, `NostrService` treats it as an identity change and rebuilds the whole client — tearing down and re-dialing every relay mid-startup, and rebuilding every provider that watches `nostrServiceProvider`.

## Mechanism (code-verified on current `main`)

1. `AuthService.initialize()` restores the local Divine identity, then launches `_upgradeDivineRpcInBackground` unawaited when RPC isn't ready yet (`mobile/lib/services/auth_service.dart:638-640`).
2. That method's `finally` re-emits the current state: `_authStateController.add(_authState)` (`auth_service.dart:729-734`). This bypasses the equality dedup in `_setAuthState` (`auth_service.dart:4047-4051`), so the emission always reaches subscribers even though nothing changed.
3. `NostrService._onAuthStateChanged` recreates the client whenever `newPubkey != _lastPubkey` (`mobile/lib/providers/nostr_client_provider.dart:441`) — and `_lastPubkey` is only assigned after client initialization *succeeds* (`nostr_client_provider.dart:277-278`, `:365`).
4. The build-path initialization runs as a bare `Future.microtask` outside the serialized auth queue (`nostr_client_provider.dart:207-215`) and takes the full relay-connect settle (multi-second at cold start, see #5863). The RPC upgrade is one bounded HTTP round-trip to Keycast, so its nudge routinely resolves inside that window.
5. Result: `newPubkey != _lastPubkey(null)` → dispose the in-flight client and create/initialize a second one (`nostr_client_provider.dart:474-494`). Everything watching `nostrServiceProvider` rebuilds — `profileRepository`, `funnelcakeApiClientProvider`, `profileStatsRepository`, the DM repositories — their streams restart and their build-time fetches re-fire.

Same shape without the nudge: a failed initialization nulls `_lastPubkey` (`nostr_client_provider.dart:301-302`), so any later auth emission also forces a recreate even though the retry timer (`nostr_client_provider.dart:321-338`) already owns that recovery path.

## Impact

- Doubled relay connection work on Divine OAuth cold starts where RPC capability wasn't ready at restore — which is the exact population the background upgrade targets.
- Duplicate downstream fetches: e.g. the duplicate profile-stats REST fetch discussed in the #5901 review. The gating fix there keeps the stats repo stable across `identityKnown -> nostrReady`, but a mid-window client swap rebuilds all relay-backed repositories regardless (and must — they hold the live client for relay fallback).
- Extended time-to-`nostrReady`.

## Possible directions

- Track the in-flight client's target pubkey at creation time (not only on init success) so `_onAuthStateChanged` can recognize "a client for this pubkey already exists / is initializing" and treat a same-pubkey emission as a no-op.
- Or stop nudging the shared auth stream from the upgrade path: the nudge exists so the session-expired sheet re-evaluates (`auth_service.dart:731-733`); routing that through the RPC-capability signal would keep `authStateStream` semantically "auth state changed".

Found while verifying the #5901 review feedback. Relates to #5863.

realmeylisdev:
## Description

Two coupled fixes on the Nostr startup path, plus regression tests.

### 1. `fix(nostr)` - harden the startup client-identity race (#5909)

`NostrService._onAuthStateChanged` recreated the whole `NostrClient` whenever
`newPubkey != _lastPubkey`, and `_lastPubkey` is only set once an init
*succeeds*. So a same-pubkey auth re-emission that lands while the current
client is still connecting relays (e.g. the Divine RPC-upgrade nudge fired
during startup) forced a full client rebuild - re-dialing every relay and
rebuilding every relay-backed provider - even though nothing changed.

This now tracks the pubkey each client is *created* to serve (`_pendingPubkey`,
set at creation, before init completes). `_onAuthStateChanged` treats an
identity it is already serving - whether fully initialized (`_lastPubkey`) or
still connecting (`_pendingPubkey`) - as a no-op. `_pendingPubkey` is cleared on
init failure, so failure-recovery re-emissions still recreate (the
recover-after-failure contract pinned by the existing tests is preserved).

**Why this was latent, not currently firing in production:** the app happens to
force `NostrService.build()` *before* identity restore (via an incidental
pre-restore `nostrServiceProvider` read - the #2144 repair below), so the
`authenticated` restore is serialized ahead of the nudge on the lifecycle queue
and dedups it. That ordering was accidental and fragile. Fix #2 removes that
incidental read, which is exactly why the race is hardened at the root here.

### 2. `fix(video)` - run the #2144 corrupted-video repair after auth restore

The repair was kicked from a pre-restore startup microtask, where the identity
isn't restored yet: `_repairCorruptedEvents` returned `0` ("not authenticated")
and `repairIfNeeded` then marked the one-time migration **complete anyway** -
permanently disabling it for every user after their first cold start (verified
on-device: launch 1 logs "Not authenticated, skipping repair", launch 2 runs
nothing).

Now the completion flag is only set once the scan actually ran
(`_repairCorruptedEvents` returns `null` when there's no identity/pubkey yet),
and the repair is triggered when the Nostr session reaches `nostrReady` instead
of during the pre-restore window. Deferring the read past restore is safe
because of fix #1.

**Related Issue:** Closes #5909

## Out of Scope

- A separate latent leak: `videoEventServiceProvider` never registers
  `ref.onDispose(service.dispose)`, so each client swap orphans a
  `VideoEventService` whose timers/subscription keep running over a disposed
  `eventRouter` (accumulates across account switches). Filed separately as
  #6174.

## Verification

- `flutter test test/providers/nostr_service_identity_source_test.dart` - 26
  tests incl. 3 new #5909 cases. The in-flight same-pubkey case is a proven
  regression guard (fails on pre-fix code: callCount 2, passes with the guard:
  callCount 1); a different-pubkey negative control confirms suppression is
  pubkey-scoped.
- `flutter test test/unit/services/corrupted_video_repair_service_test.dart` -
  existing service suite now includes completion-flag assertions for both
  pre-restore unauthenticated and empty-pubkey paths.
- Related NostrService-watching suites (dm/moderation/videos_repository/relay)
  green.
- `flutter analyze lib test` clean; untested-services floor green;
  `dart run build_runner build` regenerated `nostr_client_provider.g.dart`
  (committed); `dart format` clean.
- On-device (Android emulator + local_stack): cold-start log confirms the
  repair no longer runs pre-restore and NostrService starts cleanly with no
  spurious rebuild.

/cc @dcadenas - this touches the startup path you reworked in #6100/#6101;
your review would be appreciated.

## Type of Change


NotThatKindOfDrLiz:
Takeover update pushed in dba55dcdf.

What changed:
- Consolidated the corrupted-video repair completion-flag coverage into the existing canonical unit test file.
- Removed the duplicate parallel test file under test/services.
- Updated the PR body verification section to point at the canonical test path and linked the deferred VideoEventService disposal follow-up as #6174.

Verification:
- flutter test test/providers/nostr_service_identity_source_test.dart test/unit/services/corrupted_video_repair_service_test.dart
- flutter analyze lib test
- pre-push checks passed: merge-conflict check, analyzer, generated-files check, changed-file tests
- GitHub checks are green: mergeability, linked issue, semantic PR, generated files, format, analyze, tests, Mobile CI, preview build, CLA

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup. I re-reviewed the final diff after consolidating the duplicate repair-service tests; no remaining blockers found.

PR #4361, opened 2026-05-14
feat(cache_sync): account-scoped invalidation

realmeylisdev:
## Description

Foundation PR for the data-foundation epic (#4335). `cache_sync` gains **account-scoped invalidation** and `AuthService.signOut` consolidates cache invalidation onto it:

- `CacheSync.invalidatePrefix(prefix)` clears every entry whose key starts with `prefix`. Cache keys follow the `${pubkey}:operation` convention (RFC #4244), so calling with the leaving account's pubkey hex wipes only that account and leaves other accounts on the same device intact. SQL `LIKE` metacharacters (`%`, `_`, `\`) in `prefix` are escaped via `LIKE ... ESCAPE '\'` so the match is literal.
- Empty `prefix` now throws `ArgumentError` in all builds. The guard is runtime-enforced, so the empty-string path cannot degrade to a full-wipe in release.
- `follow_repository`'s four cache-key helpers move from the legacy `operation_${pubkey}` shape to RFC #4244's `${pubkey}:operation` shape so the new prefix invalidation actually matches.
- `AuthService.signOut` now owns cache invalidation. It captures the current pubkey and calls `CacheSync.invalidatePrefix(currentPubkey)` on **both** destructive and non-destructive paths. Invalidation failure is forwarded to Crashlytics via the existing `_reportStorageError` path so a silent disk-residency regression is observable, but the sign-out flow remains failure-tolerant.
- `SettingsAccountCubit` loses its `CacheSync.invalidateAll()` calls and `cache_sync` import. The cubit returns to single-responsibility account-list orchestration.
- The now-unused `CacheSync.invalidateAll()` / `CacheDao.deleteAll()` surface is removed rather than left as dead API.
- Small `@visibleForTesting void debugSetCurrentKeyContainer(...)` seam added to `AuthService` so the multi-account assertion test can exercise the authenticated path without driving the full sign-in pipeline.

**Related Issue:** Closes #4360. Parent epic #4335. RFC #4244. Original `cache_sync` introduction #4251.

## Out of Scope

- `CacheSync.writeIfNewer` deferred to the PR that brings up its first real consumer (M2 OgViner or M4 PersonalEvent per the #4335 decomposition). Cut from this PR per review because landing speculative public API without a caller violates the repo's no-new-tech-debt rule.
- `cache_sync.MigrationStrategy` / schema-versioned wipe for the short-lived legacy `operation_${pubkey}` cache-key cohort. Tracked explicitly in #4382.
[...]

PR #4251, opened 2026-05-11
feat(cache_sync): introduce cache_sync package with follower/following integration

hm21:
## Description

Adds a new `cache_sync` package — a Drift-backed cache layer with size-bounded LRU eviction and configurable fetch policies — and validates it end-to-end by wiring it into the follower/following data flows.

**Example:**

| Before | After |
| ------- | ------ |
| <video src="https://github.com/user-attachments/assets/37ca8d79-3407-4541-8c14-508fb82916ec" controls width="300"></video> | <video src="https://github.com/user-attachments/assets/c3eeefec-4a79-4469-8dee-6f60b6b5be24" controls width="300"></video> |

**Related Issue:** Closes #4252

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

Issue #4244, opened 2026-05-10
rfc(cache_sync): design and prototype a unified low-level stale-while-revalidate cache package


hm21:
> **Related:** #2313 – the feed-side consumer work depends on this package existing first.

## Decision

In #2313 the team reached consensus ([@rabble](https://github.com/rabble), [@NotThatKindOfDrLiz](https://github.com/NotThatKindOfDrLiz)): we want a **unified low-level cache** at the repository/data layer so that **the UI and BLoC layers never need to know about caching details**. This issue is the RFC and prototype track for that package.

---

## What `cache_sync` is

A thin internal Flutter package that wraps any network `Future` or `Stream` and turns it into a **stale-while-revalidate stream**: the local DB is emitted first (instant), then silently replaced once the network responds.

The pattern is identical regardless of what data is fetched – REST feed results, follower lists, profile metadata. Repositories call one method; the package handles everything else.

---

## Core API shape

```dart
enum CacheFetchPolicy { cacheAndNetwork, cacheOnly, networkOnly }

class CacheResult<T> {
  const CacheResult.cached(this.data) : isLive = false;
  const CacheResult.live(this.data)   : isLive = true;
  final T    data;
  final bool isLive; // false = showing cache; true = network responded
}

abstract class CacheSync {
  // Future-based (REST / single-shot) – emits cached result first, then live
  static Stream<CacheResult<List<T>>> watch<T>({
    required String cacheKey,
    required Future<List<T>> Function() fetch,
    required List<T> Function(String json) fromJson,
    required String Function(List<T> data) toJson,
    CacheFetchPolicy policy = CacheFetchPolicy.cacheAndNetwork,
    Duration? ttl,
  });

  // Single-entity variant
  static Stream<CacheResult<T?>> watchOne<T>({
    required String cacheKey,
    required Future<T?> Function() fetch,
    required T? Function(String json) fromJson,
    required String Function(T data) toJson,
    CacheFetchPolicy policy = CacheFetchPolicy.cacheAndNetwork,
    Duration? ttl,
  });

  // Stream-based (WebSocket / Nostr relay – items arrive one-by-one)
  static Stream<CacheResult<List<T>>> watchStream<T>({
    required String cacheKey,
    required Stream<T> Function() subscribe,
    required List<T> Function(String json) fromJson,
    required String Function(List<T> data) toJson,
    CacheFetchPolicy policy = CacheFetchPolicy.cacheAndNetwork,
    Duration? ttl,
[...]

PR #6175, opened 2026-07-18
chore(cache-sync): add MigrationStrategy + wipe legacy key residue

### LINKED ISSUE #4382: task(cache-sync): add MigrationStrategy + schema-versioned wipe for legacy key residue
## Description

Follow-up debt cleanup for `cache_sync` after PR #4361.

`cache_sync` currently has a short-lived cohort of legacy keys written in the old `operation_${pubkey}` shape from the 3-day window after `cache_sync` shipped in #4251 and before RFC #4244 key-shape migration landed in #4361. Those rows are not reachable by `invalidatePrefix(currentPubkey)` because they do not start with the pubkey.

This issue tracks the correct cleanup instead of leaving the residue as an untracked follow-up:

- add `cache_sync` migration/versioning scaffolding (`MigrationStrategy` / `onUpgrade` path)
- perform a schema-versioned wipe or equivalent one-time cleanup for legacy `operation_${pubkey}` cache rows
- make the cleanup explicit and reproducible rather than relying on eventual LRU eviction

## Why

PR #4361 intentionally shipped account-scoped invalidation without bundling schema-versioning work into the same diff. That scope cut is reasonable, but under the repo standard the deferred debt must be concretely tracked.

## Acceptance Criteria

- `cache_sync` has an explicit migration/versioning path for future schema bumps
- the legacy `operation_${pubkey}` cache-key cohort is removed by a deterministic upgrade path
- the implementation is documented in code/tests well enough that future cache-key migrations are straightforward
- all generated outputs and affected tests are updated as needed

## Related

- Parent epic: #4335
- Task issue: #4360
- Implementing PR: #4361
- Original cache_sync introduction: #4251

realmeylisdev:
Closes #4382

## Description

`cache_sync` shipped with `schemaVersion` pinned at 1 and **no** `MigrationStrategy`. #4382 asked for the migration + version-bump pattern so future schema/data changes have a deterministic path.

This lands that groundwork:

- **`CacheDatabase` gets an explicit `MigrationStrategy`** (`onCreate: createAll`) as the anchor point for the first real migration. `schemaVersion` stays at **1** — there is no data/schema change to bump for.
- **`MIGRATIONS.md`** documents the versioning path, with separate recipes for a data-only change and a schema-shape change (the latter noting the `make-migrations` / schema-snapshot machinery to introduce at the first real schema change).
- **A `follow_repository` regression guardrail** fails if any of the four cached key builders (`my_followers_`, `my_following_`, `others_followers_`, `others_following_`) ever re-emits the orphan-prone legacy underscore prefix, instead of the RFC #4244 `${pubkey}:operation` colon shape.

## Why no data DELETE (changed from the first revision)

The first revision added a v1 → v2 `onUpgrade` that deleted the short-lived legacy `my_followers_${pubkey}` / `my_following_${pubkey}` rows `follow_repository` wrote between #4251 and #4361.

Per review (thanks @hm21): that residue **never reached a tagged native release** — the first release containing `cache_sync` already shipped the colon-scoped shape from #4361 — so the DELETE would clean nothing on real devices. It has been dropped. The invariant it was meant to protect (never regress to the underscore shape) is now guarded at the source by the `follow_repository` test above, which also closes the 2-of-4-prefixes blind spot the review flagged.

## Out of Scope / Split Out

- The pre-existing upload-manager recovery-sweep flake fix that was bundled in the first revision has been moved to its own test-only PR: **#6193**.
- Full drift schema-snapshot + generated migration-test infrastructure — deferred to the first real schema-shape change (documented in `MIGRATIONS.md`).

## Verification

Ran from `mobile/`:

- `flutter analyze packages/cache_sync packages/follow_repository` — **No issues found**
- `cache_sync`: `flutter test --coverage` — **83/83 pass**, 100% line coverage on non-generated sources (the package's VGV gate; the new `migration` getter + `onCreate` are covered by existing DB opens)
- `follow_repository`: `flutter test --coverage` — **217/217 pass**, 100% coverage
- `build_runner` regen confirmed a no-op for `cache_database.g.dart` (`schemaVersion` / `MigrationStrategy` are hand-written overrides, not generated)

## Type of Change

- [x] 🧹 Code refactor
- [x] 📝 Documentation

### DIFF
--- a/mobile/packages/cache_sync/MIGRATIONS.md
+++ b/mobile/packages/cache_sync/MIGRATIONS.md
@@ -1,0 +1,70 @@
+# cache_sync Database Migrations
+
+`cache_sync` persists cache entries in a single Drift table (`CacheEntries`) in
+a local, unencrypted SQLite file (`cache_sync.db`). It is a **disposable
+cache** — nothing here is a source of truth. Schema and data changes are
+handled through Drift's [`MigrationStrategy`](https://drift.simonbinder.eu/docs/advanced-features/migrations/).
+
+## Current schema version
+
+**Version 1** — see `lib/src/cache_database.dart`.
+
+| Version | Change |
+|---------|--------|
+| 1 | Initial `CacheEntries` table (#4251). |
+
+`cache_database.dart` declares an explicit `MigrationStrategy` (currently just
+`onCreate: createAll`). This is the deliberate anchor point for the first real
+migration; the recipes below describe how to add one.
+
+> **#4382 note:** #4382 evaluated a v1 → v2 data-only migration to delete the
+> short-lived legacy `my_followers_${pubkey}` / `my_following_${pubkey}` rows
+> that `follow_repository` wrote between #4251 and #4361 (before RFC #4244's
+> `${pubkey}:operation` key shape landed). That cohort never reached a tagged
+> native release — the first release containing `cache_sync` already shipped
+> the colon-scoped shape — so no data migration was shipped. The colon-scoped
+> key shape is instead guarded by a `follow_repository` regression test that
+> fails if a key builder ever re-emits the legacy prefix.
+
+## How migrations run
+
+`MigrationStrategy.onUpgrade(m, from, to)` fires **once per existing install**
+when the stored `user_version` is below `schemaVersion`. Fresh installs run
+`onCreate` (which calls `createAll()`) and never run `onUpgrade`.
+
+Because `onUpgrade` never re-runs once `user_version` reaches the target, each
+step must be self-contained: a step that ships incomplete cannot be corrected
+in place — it requires a *further* version bump with a new step.
+
+## Adding a future migration
+
+### Data-only change (delete/rewrite rows, no schema change)
+
+1. Bump `schemaVersion` in `lib/src/cache_database.dart`.
+2. Extend `onUpgrade` with a guarded block for the new step, e.g.
+   `if (from < 2) { … }`. Keep each step idempotent and gated on `from`.
+3. Create `test/cache_migration_test.dart` with a test following the two-phase
+   pattern: open at the new version to create the schema, seed rows, stamp the
+   DB back to the previous `user_version`, close, then reopen so `onUpgrade`
+   runs — and assert the data effect. Use the `NativeDatabase.opened` shared
+   in-memory connection idiom (`closeUnderlyingOnClose: false`) so table and
+   data survive the drift close between phases. The test must be able to fail:
+   seed a row that is *present at the lower version* and assert it is
+   deleted/rewritten, not one inserted after the upgrade hook could touch it.
+4. Run `dart run build_runner build` and confirm `cache_database.g.dart` is
+   unchanged (a data-only bump does not regenerate it).
+
+### Schema-shape change (add/alter/drop a column or table)
+
+This package does not yet ship generated schema snapshots. The first
+schema-shape change should introduce them so migrations can be verified
+against exported schemas (mirroring `db_client`):
+
+1. Edit the table in `lib/src/cache_entries_table.dart`, then bump
+   `schemaVersion` in `lib/src/cache_database.dart`.
+2. `dart run build_runner build` to regenerate `cache_database.g.dart`.
+3. `dart run drift_dev make-migrations` — this uses the `schema_dir` /
+   `test_dir` already declared in `build.yaml` to write
+   `drift_schemas/cache_database/drift_schema_v*.json`,
+   `cache_database.steps.dart`, and generated migration-test helpers.
+4. Implement the step in `onUpgrade` and add migration tests.
--- a/mobile/packages/cache_sync/lib/src/cache_database.dart
+++ b/mobile/packages/cache_sync/lib/src/cache_database.dart
@@ -15,4 +15,15 @@ class CacheDatabase extends _$CacheDatabase {
 
   @override
   int get schemaVersion => 1;
+
+  /// Migration strategy for the cache store.
+  ///
+  /// `cache_sync` is a disposable local cache with a single [CacheEntries]
+  /// table and no schema-shape migration yet. This is the explicit anchor
+  /// point for the first real migration; see `MIGRATIONS.md` for the
+  /// versioning path and the #4382 decision not to ship a legacy-key-residue
+  /// data delete.
+  @override
+  MigrationStrategy get migration =>
+      MigrationStrategy(onCreate: (m) => m.createAll());
 }
--- a/mobile/packages/cache_sync/test/fake_cache_dao.dart
+++ b/mobile/packages/cache_sync/test/fake_cache_dao.dart
@@ -68,6 +68,9 @@ class FakeCacheDao implements CacheDao {
   /// Number of stored entries (for assertions).
   int get length => _store.length;
 
+  /// All stored cache keys (for assertions on key shape).
+  Iterable<String> get keys => _store.keys;
+
   /// Raw payload access without TTL check (for assertions).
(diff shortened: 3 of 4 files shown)


mobile/packages/cache_sync/lib/src/cache_database.dart
```diff
@@ -14,5 +14,32 @@ class CacheDatabase extends _$CacheDatabase {
...
+  MigrationStrategy get migration => MigrationStrategy(
+    onCreate: (m) => m.createAll(),
+    onUpgrade: (m, from, to) async {
+      if (from < 2) {
+        // Escape the underscores in each prefix so LIKE matches literally
+        // (same idiom as CacheDaoImpl.deletePrefix). The legacy keys and all
+        // current `${pubkey}:…` / `global:…` / `anon:…` keys are disjoint, so
+        // this removes only the orphaned cohort and preserves live caches.
+        await customStatement(
+          r"DELETE FROM cache_entries WHERE cache_key LIKE 'my\_followers\_%' "
```
hm21: If the DELETE stays, it's inconsistent. #4251 added four legacy key builders and #4361 renamed all four, but this only deletes two — `others_followers_${pubkey}` / `others_following_${pubkey}` are left orphaned with identical properties (not pubkey-prefixed, never re-read again). And once `user_version` reaches 2 this block never re-runs, so it can't be corrected later without a v3. Either clean all four prefixes (and extend the guardrail test, which has the same 2-of-4 blind spot), or drop the DELETE.

mobile/packages/cache_sync/test/cache_migration_test.dart
```diff
@@ -0,0 +1,125 @@
...
+            '$_pubkey:my_followers',
+            '$_pubkey:others_followers',
+            'global:categories:v1',
+            'anon:home_feed:following',
+          ]),
+        );
+      },
+    );
+
+    test('does not delete legacy-shaped rows on a fresh v2 install', () async {
```
hm21: This test — and tests 3 (escapes underscores…) and 4 (does not re-run cleanup) below — can't fail for the regression they're named after. Test 2 inserts the legacy row *after* open, so no migration hook could ever touch it; test 3 passes even with a no-op onUpgrade; test 4 seeds a colon-scoped row that never matches the DELETE. Only test 1 actually proves the migration runs and deletes. If the DELETE stays, give these teeth: seed a legacy-shaped row present at `user_version = 2` and assert survival/deletion accordingly (per the 'a test must be able to fail' rule).

mobile/packages/cache_sync/MIGRATIONS.md
```diff
@@ -0,0 +1,50 @@
...
+4. Run `dart run build_runner build` and confirm `cache_database.g.dart` is
+   unchanged (a data-only bump does not regenerate it).
+
+### Schema-shape change (add/alter/drop a column or table)
+
+This package does not yet ship generated schema snapshots. The first
+schema-shape change should introduce them so migrations can be verified
+against exported schemas (mirroring `db_client`):
+
+1. Edit `lib/src/cache_entries_table.dart`, bump `schemaVersion`.
```
hm21: `schemaVersion` lives in `cache_database.dart`, not `cache_entries_table.dart` — the data-only recipe two steps up says so correctly, so this contradicts it. Name both: edit the table in `cache_entries_table.dart`, bump `schemaVersion` in `cache_database.dart`.

mobile/test/services/upload_manager_recovery_test.dart
```diff
@@ -350,13 +350,22 @@ void main() {
 
         uploadManager.resumeInterruptedUpload(upload.id);
 
+        // The failed status lands (synchronous in-memory Hive put) one future
```
hm21: Unrelated to cache_sync — this is a pre-existing upload-manager flake (repros on main). Please split it into its own test-only PR; bundled here it widens the scope and gets mislabeled under `chore(cache-sync)` on squash-merge. The fix itself looks correct (the two trailing `expect`s are now redundant with the folded wait condition — minor).

hm21 (commented):
I don't think this migration is worth shipping. `cache_sync` is a disposable cache, and the legacy `my_followers_` / `my_following_` residue never reached a real user's device: no tagged native release ever wrote those keys — 1.0.13, the first release that contained cache_sync at all, already shipped the colon-scoped shape from #4361 — and the only surface that briefly did (the always-on web build at app.divine.video, deployed from every main merge) has virtually no users. So the data DELETE cleans nothing on real devices; it's dev-machine hygiene at most.

Keeping the migration scaffolding (MigrationStrategy + MIGRATIONS.md) as groundwork for the first real schema change is a fair reason to land it — but if so, please address the inline notes.

mobile/packages/cache_sync/lib/src/cache_database.dart
realmeylisdev: Dropped the DELETE in 40de002808. Per the review it cleans nothing on real devices (the residue never reached a tagged native release), so `schemaVersion` is back to 1 and the `MigrationStrategy` is now `onCreate`-only scaffolding — the 2-of-4 question is moot. The underscore-shape invariant is now guarded at the source instead: the `follow_repository` guardrail covers all four builders (`my_followers_`, `my_following_`, `others_followers_`, `others_following_`).

mobile/packages/cache_sync/test/cache_migration_test.dart
realmeylisdev: With the DELETE gone, this whole test file is deleted in 40de002808. For the future, `MIGRATIONS.md`'s data-only recipe now spells out the teeth requirement: seed a row present at the lower `user_version` and assert it is deleted/rewritten — not one inserted after the upgrade hook could touch it.

mobile/packages/cache_sync/MIGRATIONS.md
realmeylisdev: Fixed in 40de002808. The schema-shape recipe now reads "edit the table in `cache_entries_table.dart`, then bump `schemaVersion` in `cache_database.dart`", matching the data-only recipe. The doc was also rewritten for the groundwork-only approach (no v2 data delete).

mobile/test/services/upload_manager_recovery_test.dart
realmeylisdev: Split to its own test-only PR #6193 and removed from this branch in 40de002808. Also trimmed the redundant trailing `status` re-check there per your note — the folded wait already gates on both transitions, so the test keeps just the `isUploadInFlight` named-contract assertion.

realmeylisdev:
@hm21 thanks — accepted the core point: the legacy `my_followers_` / `my_following_` residue never reached a tagged native release, so the v1 → v2 data DELETE cleaned nothing on real devices. Dropped it (`schemaVersion` back to 1).

What lands is the groundwork you said was worth keeping:

- explicit `MigrationStrategy` (`onCreate`-only) as the anchor point for the first real migration, plus a rewritten `MIGRATIONS.md` documenting the versioning path (both the data-only and schema-shape recipes now agree that `schemaVersion` lives in `cache_database.dart`).
- a `follow_repository` colon-key guardrail that now covers **all four** builders (`my_followers_`, `my_following_`, `others_followers_`, `others_following_`) — no 2-of-4 blind spot.

The delete-specific migration tests are gone with the DELETE; `MIGRATIONS.md` spells out the must-be-able-to-fail two-phase pattern for when a real migration is added. The upload-manager recovery flake is split to #6193 (with the redundant trailing `status` re-check trimmed there). All four inline threads are addressed in 40de002808 and resolved.

dcadenas (approved):
✅ safer cache migrations, ship it

Issue #4335, opened 2026-05-13
epic(data-foundation): unify caching, persistence, and platform dependency strategy


NotThatKindOfDrLiz:
## Summary

Caching, persistence, and platform dependency choices are currently fragmented across multiple storage backends, cache policies, and ownership models. This should be tracked as one coordinated foundation epic instead of several disconnected discussion issues.

This issue is the source of truth for data foundation work around caching, persistence, storage strategy, and shared dependency ownership.

## Problem

Current problems include:
- multiple cache systems with different TTL and invalidation rules
- overlapping persistent storage backends
- unclear ownership of platform/dependency decisions
- weak consistency around schema/versioning and long-term storage strategy
- difficulty answering basic questions about eviction, corruption recovery, or low-storage behavior

## Scope

- define cache tiers and invalidation strategy
- clarify persistent storage direction and migration strategy
- improve schema/versioning expectations
- document or reduce ad hoc dependency/platform decisions
- reduce duplicated “local solution per feature” patterns where a shared strategy is warranted

## Existing Related Work

- `#3624`
- `#3618`
- `#3657`
- `#3635`
- `#3629`
- `#3660`

## Acceptance Criteria

- caching and persistence work is grouped under a coherent parent issue
- the repo has an explicit direction for storage and cache policy
- dependency/platform ownership questions are no longer floating as unrelated audit leaves
- related leaf issues can be closed or superseded into this epic

## Done When

- data foundation work is coordinated instead of fragmented
- storage/cache decisions are easier to reason about and evolve safely
- the highest-value migrations or policy decisions are decomposed into concrete child work

---

## Assignee
`@NotThatKindOfDrLiz`

## Weekly Priority
`P2`

## Weekly Priority Meaning
- `P1` means top priority for the week.
- `P2` means complete this week after P1 priorities are finished.

## Execution Order
1. define target storage/cache direction
2. group and supersede related audit leaves
3. sequence migrations and policy decisions by payoff and risk

## Subgroup
- `@NotThatKindOfDrLiz` — DRI
- `@realmeylisdev` — persistence/caching support
- `@untreu2` — data/repository support

## Coordination
The listed subgroup must review scope, sequencing, and ownership before implementation starts. Coordination may be synchronous or asynchronous. Do not wait for central coordination if the subgroup can resolve the question directly. Escalate only when blocked or when a platform or storage decision is genuinely needed.

## Agent Execution Notes
Agents working from this issue must:
[...]

Issue #3937, opened 2026-05-03
feat(profile-editor): deep-link to NIP-39 identity verification


rabble:
## Context

Follows from #3935 (clickable bio links — URLs, website field, verified identity chips).

After #3935 ships, profiles **display** verified NIP-39 identities. There is no in-app way for a user to **add** one — the verification flow lives in `divine-identify-verification-service`'s browser UI (it runs OAuth for Twitter/Bluesky/YouTube/TikTok and proof-post flows for the rest, then publishes the updated Kind-0).

## Scope

- Add a "Verified accounts" / "Identity links" section to the profile editor screen.
- List the user's currently verified identities (read from their own Kind-0).
- Add a "Connect another account" button that deep-links into `divine-identify-verification-service`'s browser UI with an appropriate signer hand-off (`window.nostr` / `login.divine.video` / NIP-46 — see verification service README for the full set).
- After the user completes the flow and the service publishes the updated Kind-0, the editor refreshes its identity list when the user returns to the app.

## Out of scope

- In-app OAuth / proof-post flows.
- Removing identities.

## Acceptance criteria

- Profile editor shows the user's currently verified identities.
- Tapping "Connect another account" launches the verification service URL with the correct signer parameter.
- Returning to the app refreshes the identity list.
- 100% test coverage on the new editor section.

realmeylisdev:
Verified shipped. The editor-side NIP-39 verifier flow landed in #4179 and remains on `main` after the profile-setup decomposition (#4512):

- Editor lists currently-verified identities (`VerifiedAccountsSection` → `VerifiedAccountsRow`).
- "Get verified" CTA deep-links to `verifier.divine.video` with signer hand-off — web via the iframe `window.nostr` bridge, native via the system browser to `login.divine.video`.
- Identity list refreshes on return (web after the iframe route returns; native on app resume via `didChangeAppLifecycleState`).
- Covered by `test/screens/profile_setup/verified_accounts_section_test.dart`.

#4179's body referenced #3933/#3955 but not this issue, so it was left open. Closing as completed.

PR #6176, opened 2026-07-19
feat(profile): read NIP-39 claims from kind 10011 and persist verification cache (#3936)

### LINKED ISSUE #3936: feat(profile): persist NIP-39 identity verification cache across sessions
## Context

Follows from #3935 (clickable bio links — URLs, website field, verified identity chips).

The first PR uses an **in-memory session cache** in `IdentityVerificationRepository` for verified NIP-39 claims. Verification re-runs on every cold start. The verification service (`divine-identify-verification-service`) has its own KV cache server-side, so the cost is mainly latency, not duplicate work — but on slow networks the chip row visibly pops in late.

## Scope

- Add a Hive box (`identity_verification_cache`) keyed by `(pubkey, platform, identity)`.
- Store the last verification result + timestamp.
- TTL aligned with the service's KV TTL (confirm value with `divine-identify-verification-service` before implementation).
- Reads first hit Hive, then fall through to the network when stale or missing.
- Writes update Hive on every successful verification call.
- Cache-clear on user logout / nsec switch.

## Out of scope

- Cross-device sync of the cache.
- Background refresh.

## Acceptance criteria

- Verified chips render on cold start without a network round-trip when the cache is fresh.
- Stale entries (past TTL) are refreshed on next profile open.
- Cache survives app restart, is cleared on logout.
- 100% test coverage on the persistence layer.

realmeylisdev:
## Description

Two coupled features, delineated in two commits:

**1. Correct claims source — kind 10011 (commit 1 + repository layer of commit 2).**
The verifier web UI has published NIP-39 identity claims to **kind 10011** (not kind 0) since April, per the 2026-02 NIP-39 revision (nostr-protocol/nips#2216); relay.divine.video stores these events and Amethyst publishes them too. Mobile read only kind-0 `i` tags, so claims added or removed through the verifier flow never appeared — including after #3937's editor round-trip. There are live profiles today whose verified GitHub claim exists only in kind 10011 while their kind 0 carries no `i` tags at all. `ProfileRepository` now consults kind 10011 first (cached row → live relay query → kind-0 `i` tags), mirroring the verifier UI's own consult order, and caches the source event in a new `identity_events` table so cold starts can render chips offline.

**2. Persistent verification cache (#3936's original ask).**
`VerifierClient` was intentionally stateless, so every profile open paid a 0.3–1.5s verifier round-trip before chips rendered; refresh dropped chips visibly and any network failure erased them. `IdentityClaimsRepository` now keeps a verified-only snapshot per profile in a new `identity_verifications` table:

- Chips render instantly from cache (parsed claims ∩ snapshot, full-tuple match including proof — a rotated proof re-verifies instead of serving a stale verdict).
- The verifier is skipped entirely when the snapshot is fresh (`checked_at + 24h`, mirroring the service's verified KV TTL and anchored on the server's own timestamp) and covers every current claim; otherwise it re-verifies on open (stale-while-revalidate, chips update live).
- Only `verified: true` results are persisted. The verifier returns pubkey/platform rate-limit rejections as HTTP-200 `verified: false` bodies it never caches server-side (`src/routes/verify.ts`), so any response containing one leaves the snapshot untouched — a rate-limit burst cannot freeze false 'unverified' verdicts.
- Both profile blocs keep last-known-good claims on verifier failure and thread claims through refresh instead of dropping them.

Both tables use the db_client runtime CREATE-IF-NOT-EXISTS idiom (schemaVersion stays 1) with schema-parity and upgrade-path tests.

**Related Issue:** Closes #3936

### ⚠️ Deviations from the issue as written

1. **Drift, not a Hive box.** The issue predates the data-foundation direction (#4335); the sibling NIP-05 verification cache already lives in Drift, and new Hive boxes are the pattern #4335 curbs.
2. **Cache clears on logout/nsec switch.** Both `identity_events` and `identity_verifications` are wired into the existing database cleanup path to satisfy #3936's acceptance criteria.
3. **Kind-10011 read path folded in.** Caching the kind-0-derived claims alone would have fossilized a stale source, so the source fix ships in the same PR.

## Out of Scope

- Publishing kind 10011 from mobile (the verifier flow owns claim writes).
- Cross-device sync, timer-based background refresh, divine-web parity (per issue).
- Known pre-existing gaps left untouched: the shell-route `/profile/:npub` path for another user has no `OtherProfileBloc` so chips never render there.

## Verification

- `db_client`: full database + DAO suites incl. new schema-parity/upgrade tests (43 tests).
- `profile_repository`: full suite with coverage — 285 tests, both touched lib files fully covered (package enforces the VGV default gate).
- Bloc suites: `other_profile` + `my_profile` incl. new SWR cases: fresh-snapshot renders with zero verifier calls, stale-snapshot cached→fresh emits, verifier failure keeps last-known-good, cache-read failure degrades to network, refresh carries chips.
- Profile header widget tests (66).
- `dart run build_runner build` — generated files committed (`auth_providers.g.dart`, Drift outputs, schema snapshot regenerated via `drift_dev make-migrations`).
- Backend contract verified against `divine-identify-verification-service` source at origin/main, live probes of `verifier.divine.video` (batch/single/cache-hit behavior, `checked_at` stability), and a full local `wrangler dev` run (response shapes for failed/platform-error/validation paths, KV cache order, rate-limit behavior).

## Type of Change

### DIFF
--- a/mobile/lib/blocs/other_profile/other_profile_bloc.dart
+++ b/mobile/lib/blocs/other_profile/other_profile_bloc.dart
@@ -76,7 +76,17 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
       pubkey: pubkey,
     );
     if (isClosed) return;
-    emit(OtherProfileLoading(profile: cachedProfile));
+    // Carry claims through the reload so chips do not vanish and re-pop
+    // while the verifier revalidates (#3936). Read AFTER the await so an
+    // authoritative claims resolve landing during the cache read is not
+    // clobbered by a stale capture.
+    final currentClaims = _claimsFromState(state);
+    emit(
+      OtherProfileLoading(
+        profile: cachedProfile,
+        verifiedClaims: currentClaims,
+      ),
+    );
 
     try {
       final freshProfile = await _profileRepository.fetchFreshProfile(
@@ -84,11 +94,24 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
         requireRawKind0: _requireRawKind0,
       );
       if (isClosed) return;
+      final latestClaims = _claimsFromState(state);
       if (freshProfile != null) {
-        emit(OtherProfileLoaded(profile: freshProfile, isFresh: true));
+        emit(
+          OtherProfileLoaded(
+            profile: freshProfile,
+            isFresh: true,
+            verifiedClaims: latestClaims,
+          ),
+        );
         add(const VerifiedClaimsRequested());
       } else if (cachedProfile != null) {
-        emit(OtherProfileLoaded(profile: cachedProfile, isFresh: false));
+        emit(
+          OtherProfileLoaded(
+            profile: cachedProfile,
+            isFresh: false,
+            verifiedClaims: latestClaims,
+          ),
+        );
         add(const VerifiedClaimsRequested());
       } else {
         emit(
@@ -98,7 +121,13 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
     } catch (e) {
       if (isClosed) return;
       if (cachedProfile != null) {
-        emit(OtherProfileLoaded(profile: cachedProfile, isFresh: false));
+        emit(
+          OtherProfileLoaded(
+            profile: cachedProfile,
+            isFresh: false,
+            verifiedClaims: _claimsFromState(state),
+          ),
+        );
         add(const VerifiedClaimsRequested());
       } else {
         emit(
@@ -199,7 +228,7 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
       );
       if (isClosed) return;
       if (cached != null) {
-        _emitClaims(emit, profile.pubkey, cached.claims);
+        _emitClaims(emit, profile.pubkey, cached.claims, authoritative: false);
       }
     } on Exception catch (e, stackTrace) {
       // Cache read failure — degrade straight to the network path.
@@ -214,22 +243,28 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
         kind0Tags: profile.rawTags,
       );
     } on Exception catch (e, stackTrace) {
-      // Source-cache write failure — resolve against the kind-0 tags we have.
+      // Degraded claims source (cache I/O failure). Resolving against the
+      // kind-0 fallback could authoritatively clear chips for a profile
+      // whose claims live in kind 10011, so keep last-known-good and let
+      // the next revalidation tick retry.
       addError(e, stackTrace);
-      freshTags = profile.rawTags;
+      return;
     }
     if (isClosed) return;
 
     // The repository owns the skip-or-verify (stale-while-revalidate)
-    // decision; the bloc just renders whatever it resolves to (#3936).
+    // decision; the bloc just renders whatever it resolves to (#3936). The
+    // currently-rendered claims ride along so a rate-limited (inconclusive)
+    // outcome cannot clear visible chips.
     try {
       final claims = await repo.resolveClaims(
         pubkey: profile.pubkey,
         freshTags: freshTags,
         cached: cached,
+        renderedClaims: _claimsFromState(state),
       );
       if (isClosed) return;
-      _emitClaims(emit, profile.pubkey, claims);
+      _emitClaims(emit, profile.pubkey, claims, authoritative: true);
     } on Exception catch (e, stackTrace) {
       // Verifier failures are expected (network/4xx/5xx/timeout). Per
       // .claude/rules/error_handling.md they are NOT Reportable. Keep the
@@ -241,17 +276,48 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
   /// Emits [claims] onto the latest state when it still shows
   /// [profilePubkey]'s profile. State may have changed mid-await, so the
   /// check runs at emit time.
+  ///
+  /// A non-[authoritative] emit (the instant cache path) may only add
+  /// chips: it is unioned with the rendered set, so a reduced or empty
+  /// snapshot intersection never clears claims the verifier has not
+  /// negatively confirmed. Only the authoritative resolve path replaces
+  /// the rendered set outright.
   void _emitClaims(
     Emitter<OtherProfileState> emit,
     String profilePubkey,
-    List<IdentityClaim> claims,
-  ) {
+    List<IdentityClaim> claims, {
+    required bool authoritative,
+  }) {
     final latest = state;
     if (latest is OtherProfileLoaded &&
         latest.profile.pubkey == profilePubkey) {
-      emit(latest.copyWith(verifiedClaims: claims));
+      final next = authoritative
+          ? claims
+          : _withRenderedPreserved(claims, latest.verifiedClaims);
+      emit(latest.copyWith(verifiedClaims: next));
     }
   }
+
+  /// Rendered-first union keyed on case-insensitive `platform:identity`:
+  /// order-stable (no chip reorder churn across revalidation ticks, and
+  /// the steady-state result is deep-equal to [rendered] so the emit is
+  /// Equatable-suppressed) and dedup-safe across casing or proof drift.
+  List<IdentityClaim> _withRenderedPreserved(
+    List<IdentityClaim> incoming,
+    List<IdentityClaim> rendered,
+  ) {
+    if (rendered.isEmpty) return incoming;
+    final renderedKeys = {for (final claim in rendered) _identityKeyOf(claim)};
+    return [
+      ...rendered,
+      ...incoming.where(
+        (claim) => !renderedKeys.contains(_identityKeyOf(claim)),
+      ),
+    ];
+  }
+
+  String _identityKeyOf(IdentityClaim claim) =>
+      '${claim.platform.toLowerCase()}:${claim.identity.toLowerCase()}';
 
   List<IdentityClaim> _claimsFromState(OtherProfileState state) {
     return switch (state) {
--- a/mobile/lib/blocs/other_profile/other_profile_bloc.dart
+++ b/mobile/lib/blocs/other_profile/other_profile_bloc.dart
@@ -1,6 +1,7 @@
 // ABOUTME: BLoC for viewing another user's profile
 // ABOUTME: Implements cache+fresh pattern and block/unblock actions
 
+import 'package:bloc_concurrency/bloc_concurrency.dart';
 import 'package:content_blocklist_repository/content_blocklist_repository.dart';
 import 'package:equatable/equatable.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
@@ -44,7 +45,10 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
     on<OtherProfileRefreshRequested>(_onRefreshRequested);
     on<OtherProfileBlockRequested>(_onBlockRequested);
     on<OtherProfileUnblockRequested>(_onUnblockRequested);
-    on<VerifiedClaimsRequested>(_onVerifiedClaimsRequested);
+    on<VerifiedClaimsRequested>(
+      _onVerifiedClaimsRequested,
+      transformer: restartable(),
+    );
   }
 
   final ProfileRepository _profileRepository;
--- a/mobile/lib/blocs/other_profile/other_profile_bloc.dart
+++ b/mobile/lib/blocs/other_profile/other_profile_bloc.dart
@@ -116,7 +116,15 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
       OtherProfileLoaded(:final profile) => profile,
       OtherProfileError(:final profile) => profile,
     };
-    emit(OtherProfileLoading(profile: currentProfile));
+    // Carry claims through the refresh so chips do not vanish and re-pop
+    // after the round-trip (#3936).
+    final currentClaims = _claimsFromState(state);
+    emit(
+      OtherProfileLoading(
+        profile: currentProfile,
+        verifiedClaims: currentClaims,
+      ),
+    );
 
     try {
       final freshProfile = await _profileRepository.fetchFreshProfile(
@@ -125,7 +133,13 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
       );
       if (isClosed) return;
       if (freshProfile != null) {
-        emit(OtherProfileLoaded(profile: freshProfile, isFresh: true));
+        emit(
+          OtherProfileLoaded(
+            profile: freshProfile,
+            isFresh: true,
+            verifiedClaims: currentClaims,
+          ),
+        );
         add(const VerifiedClaimsRequested());
       } else {
         emit(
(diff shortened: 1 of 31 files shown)


mobile/lib/blocs/other_profile/other_profile_bloc.dart
```diff
@@ -161,28 +181,91 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
...
+      freshTags = profile.rawTags;
+    }
+    if (isClosed) return;
+    final freshClaims = IdentityClaimsRepository.parseClaims(
+      profile.pubkey,
+      freshTags,
+    );
+
+    // Snapshot fresh and covering every current claim → skip the verifier.
+    if (cached != null && cached.isFresh) {
```
hm21: The skip-or-verify decision (fresh + fully-covered → skip verifier, else verify) is a business rule, and this ~45-line SWR block is duplicated near-verbatim in `my_profile_bloc._onVerifiedClaimsRequested`. architecture.md keeps source-selection/fallback in the repository, and the sibling SWR caches (`cache_sync`, `follow_repository`) already do. Consider collapsing it into one `IdentityClaimsRepository.resolveClaims(pubkey, freshTags, cached)` so both blocs reduce to emit-instant → resolve → emit (also removes the double parse on the verify path).

mobile/packages/profile_repository/lib/src/identity_claims_repository.dart
```diff
@@ -64,12 +137,71 @@ class IdentityClaimsRepository {
...
+    required String pubkey,
+    required List<VerificationResult> results,
+    required List<IdentityClaim> verified,
+  }) async {
+    final dao = _verificationsDao;
+    if (dao == null) return;
+
+    final rateLimited = results.any(
+      (r) =>
+          !r.verified && (r.error?.startsWith('Rate limit exceeded') ?? false),
```
hm21: This keys the destructive path on `error.startsWith('Rate limit exceeded')` — a field the model documents as "Not a stable key". If that string drifts during an all-`verified:false` burst, `verified.isEmpty` → `deleteVerification` wipes the good snapshot, the exact loss this guards against. At minimum a named constant next to `maxBatchSize`/`verifiedTtl`; ideally a stable rate-limit flag server-side.

mobile/packages/profile_repository/lib/src/identity_claims_repository.dart
```diff
@@ -64,12 +137,71 @@ class IdentityClaimsRepository {
...
+      final decoded = jsonDecode(json) as List<dynamic>;
+      return {
+        for (final entry in decoded.cast<Map<String, dynamic>>())
+          (
+            (entry['platform'] as String).toLowerCase(),
+            (entry['identity'] as String).toLowerCase(),
+            entry['proof'] as String,
+          ),
+      };
+    } on Exception {
```
hm21: `_decodeSnapshot` is documented "null when corrupt" but only catches `Exception`. A valid-JSON-but-wrong-shape row throws `TypeError` (an `Error`), which escapes here and the bloc's `on Exception catch` too → handler aborts and a reportable `TypeError` reaches Crashlytics. `catch (_)` upholds the contract. Same in `ProfileRepository._decodeIdentityTags`. Low reachability (own writers only), but the `'not json'`-only corrupt tests don't cover this mode.

mobile/packages/profile_repository/lib/src/profile_repository.dart
```diff
@@ -231,6 +244,115 @@ class ProfileRepository {
...
+        ],
+        useCache: false,
+      );
+      final identityEvents = events
+          .where((e) => e.kind == identityEventKind)
+          .toList();
+      if (identityEvents.isEmpty) return null;
+      // Relays do not guarantee newest-first ordering, so pick the event
+      // with the highest createdAt — same rule as the kind-0 fetch above.
+      return identityEvents.reduce((a, b) => b.createdAt > a.createdAt ? b : a);
```
hm21: On a `created_at` tie this keeps the first relay-returned event, not NIP-01's lowest-id rule → nondeterministic winner for same-second kind-10011 events. Minor, and it mirrors the existing kind-0 reduce — but a lowest-id tie-break makes chip selection deterministic.

mobile/packages/db_client/lib/src/database/tables.dart
```diff
@@ -1475,3 +1475,76 @@ class ProcessedGiftWraps extends Table {
...
+  /// JSON-encoded `List<List<String>>` of the event's `i` tags only.
+  TextColumn get tagsJson => text().named('tags_json')();
+
+  /// Which event kind the tags came from: 10011 (current spec) or 0
+  /// (legacy fallback).
+  IntColumn get sourceKind => integer().named('source_kind')();
+
+  /// `created_at` of the source event (unix seconds), used to keep the
+  /// newest event when relays disagree.
+  IntColumn get eventCreatedAt => integer().named('event_created_at')();
```
hm21: `event_created_at` (and `fetched_at`) are written on every upsert but never read anywhere. This doc says it's "used to keep the newest event when relays disagree," but newest-selection happens in-memory in `_fetchIdentityEvent` and never consults the stored column. Drop the columns or fix the comment (+ `// TODO(#…)` if it's future-proofing).

mobile/packages/profile_repository/test/src/identity_claims_repository_test.dart
```diff
@@ -186,4 +204,259 @@ void main() {
...
+        () => dao.upsertVerification(
+          pubkey: any(named: 'pubkey'),
+          verifiedClaimsJson: any(named: 'verifiedClaimsJson'),
+          checkedAtFloor: any(named: 'checkedAtFloor'),
+        ),
+      );
+    });
+
+    test(
+      'leaves the snapshot untouched when any result is rate-limited',
```
hm21, line 557: This rate-limit case includes a verified claim, so `verified.isEmpty` is never hit — the all-rate-limited (zero-verified) path the guard actually exists for is untested. Reordering the `rateLimited`/`isEmpty` guards would start wiping real snapshots yet keep this green. Add a case where every result is `verified:false` with a rate-limit error, asserting `verifyNever(deleteVerification)`.

mobile/lib/blocs/other_profile/other_profile_bloc.dart
realmeylisdev: Done in f3b8028717 — the skip-or-verify SWR decision now lives in `IdentityClaimsRepository.resolveClaims(pubkey, freshTags, cached)`. Both blocs reduce to emit-instant → resolve → emit, and the fresh tags are parsed once (the verify path no longer re-parses).

mobile/packages/profile_repository/lib/src/identity_claims_repository.dart
realmeylisdev: Done in f3b8028717 — extracted `IdentityClaimsRepository.rateLimitErrorPrefix` next to `verifiedTtl`, and `_persistOutcome` keys the guard off the constant. Agreed a stable server-side rate-limit flag is the real fix; noted that in the constant's doc.

mobile/packages/profile_repository/lib/src/identity_claims_repository.dart
realmeylisdev: Done in f3b8028717 — `_decodeSnapshot` and `ProfileRepository._decodeIdentityTags` now catch `on Object`, so a valid-JSON/wrong-shape row (a `TypeError`) decodes to null instead of escaping. Added wrong-shape cases to both test files.

mobile/packages/profile_repository/lib/src/profile_repository.dart
realmeylisdev: Done in f3b8028717 — `_fetchIdentityEvent` breaks a same-second `created_at` tie by lowest event id (NIP-01), so the winner is deterministic across relay return order. Added a test that feeds both orderings.

mobile/packages/db_client/lib/src/database/tables.dart
realmeylisdev: Done in f3b8028717 — dropped `event_created_at` and `fetched_at` (written on every upsert, never read) along with the `kind0CreatedAt` thread down to the DAO, and regenerated the Drift codegen + v1 schema snapshot. The table is new in this PR, so this just corrects its initial definition.

mobile/packages/profile_repository/test/src/identity_claims_repository_test.dart
realmeylisdev: Done in f3b8028717 — added an all-`verified:false` rate-limit case asserting `verifyNever(deleteVerification)` and `verifyNever(upsertVerification)`, so reordering the rate-limit/isEmpty guards now fails.

realmeylisdev:
@hm21 all six threads addressed in f3b8028717 — thanks for the thorough pass.

The notable one: the SWR skip-or-verify decision moved out of both profile blocs into `IdentityClaimsRepository.resolveClaims(pubkey, freshTags, cached)`, so they now reduce to emit-instant → resolve → emit and the fresh tags are parsed once. `event_created_at`/`fetched_at` were dropped rather than documented since the table is new in this PR (initial definition, no shipped schema to migrate); Drift codegen + the v1 schema snapshot are regenerated.

Verified: `flutter analyze` clean; `profile_repository` full suite green with both touched files still 100%; `db_client` full suite green incl. the identity schema-parity + snapshot tests; both profile bloc suites green.

hm21 (approved):
LGTM — re-reviewed the fix commit. All six review points are addressed, each with a regression test that would fail if the fix regressed: repo-owned `resolveClaims` (skip-gate moved out of the blocs, parse-once), `on Object` decode + wrong-shape tests, dropped the write-only identity columns, named `rateLimitErrorPrefix` constant, NIP-01 lowest-id tie-break, and the zero-verified rate-limit guard test. CI green.

Two minor items are fine to leave for follow-up: `VerifiedClaimsRequested` still uses the default concurrent transformer, and `VerificationResult.fromJson` does `checked_at as int` with no default (only bites if the verifier ever returns a result without `checked_at`).

mobile/packages/profile_repository/lib/src/identity_claims_repository.dart
```diff
@@ -47,29 +93,159 @@ class IdentityClaimsRepository {
...
+  /// when [cached] is fresh and already covers every current claim, the
+  /// verifier is skipped and the parsed fresh claims are returned as-is.
+  /// Otherwise the claims are re-verified against the service, which also
+  /// refreshes the persistent snapshot. The source tags are parsed exactly
+  /// once regardless of which branch is taken.
+  ///
+  /// Throws [VerifierClientException] subtypes on the verify path — callers
+  /// should catch and keep the last-known-good claims rather than clearing
+  /// them.
+  Future<List<IdentityClaim>> resolveClaims({
```
dcadenas, line 168: Keep a rate-limited verification from narrowing the chips already shown on screen. The rate-limit guard preserves the database snapshot, but `_verifyClaims` still returns a reduced list and `resolveClaims` passes it to both blocs, where `_emitClaims` replaces the visible claims even though an HTTP-200 rate limit does not throw. Return a result that marks the batch incomplete, or otherwise make both blocs retain their non-empty rendered set when the result is reduced or empty.

dcadenas (dismissed):
Before merging, preserve the last visible verified-identity chips when a verifier response is rate-limited, and make overlapping verification requests resolve in favor of the newest request. Also confirm whether retaining both identity caches across logout is an intentional deviation from issue #3936, or clear them as the acceptance criteria require.

---

- **VerifiedClaimsRequested still runs concurrently with no transformer, so an older run can emit stale claims over a newer result** ([mobile/lib/blocs/other_profile/other_profile_bloc.dart:47](https://github.com/divinevideo/divine-mobile/blob/f3b8028717364101a8cc1df6b30f67a9becea468/mobile/lib/blocs/other_profile/other_profile_bloc.dart#L47))
  Register `VerifiedClaimsRequested` with `restartable()` in both profile blocs. This event is dispatched during load, refresh, and profile-cache updates, but it currently uses Bloc's default concurrent handling, so an older verifier call can finish after a newer one and overwrite the current claims.
- **identity_events and identity_verifications are not cleared on logout, leaving an issue #3936 acceptance criterion unmet** ([mobile/lib/providers/social_providers.dart:668](https://github.com/divinevideo/divine-mobile/blob/f3b8028717364101a8cc1df6b30f67a9becea468/mobile/lib/providers/social_providers.dart#L668))
  Confirm the logout behavior before merging this change. The cleanup path does not clear `identity_events` or `identity_verifications`, even though issue #3936 says the verification cache is cleared on logout and nsec switch. Either add both tables to the logout cleanup or record explicit product approval for retaining these viewer-independent public verification results across accounts.

NotThatKindOfDrLiz:
@dcadenas I pushed e6a355e06 to address the outstanding review items and requested re-review.

Changes made:
- Preserves already-rendered verified chips when stale revalidation receives verifier rate-limit bodies, while still not persisting rate-limited results. The preserved set is intersected with the current fresh claims source so removed claims are not revived.
- Registers `VerifiedClaimsRequested` with `restartable()` in both profile blocs and added overlap regression tests.
- Clears both `identity_events` and `identity_verifications` through the existing logout/account cleanup path, with DAO `clearAll()` coverage.
- Updated the PR description to remove the cache-survives-logout deviation and the transformer known-gap note.

Verification:
- `flutter analyze lib test`
- `flutter test test/blocs/my_profile/my_profile_bloc_test.dart test/blocs/other_profile/other_profile_bloc_test.dart`
- `flutter test test/src/database/daos/identity_events_dao_test.dart test/src/database/daos/identity_verifications_dao_test.dart` from `mobile/packages/db_client`
- Pre-push hook: analyzer, generated files, and changed-file tests all passed
- GitHub checks are green on head e6a355e06

NotThatKindOfDrLiz (approved):
Approved after takeover fixes. I pushed e6a355e06 addressing the rate-limited chip preservation, overlapping verification request ordering, and #3936 logout-cache cleanup items. CI is green; waiting on dcadenas re-review.

mobile/lib/blocs/other_profile/other_profile_bloc.dart
```diff
@@ -161,30 +185,82 @@ class OtherProfileBloc extends Bloc<OtherProfileEvent, OtherProfileState> {
...
+      // .claude/rules/error_handling.md they are NOT Reportable. Keep the
+      // last-known-good claims instead of erasing rendered chips.
+      addError(e, stackTrace);
     }
   }
 
+  /// Emits [claims] onto the latest state when it still shows
+  /// [profilePubkey]'s profile. State may have changed mid-await, so the
+  /// check runs at emit time.
+  void _emitClaims(
```
dcadenas, line 285: Guard this emit path so a reduced or empty `claims` list cannot replace a non-empty rendered `verifiedClaims` set unless the verifier returned a confirmed negative result. `resolveClaims` preserves cached claims for rate-limited outcomes only when `cached.claims.isNotEmpty` (`mobile/packages/profile_repository/lib/src/identity_claims_repository.dart:164-171`); with a null or empty snapshot, the rate-limited batch returns `[]`, and this unconditional copy erases chips that are already visible. Apply the same guard in `mobile/lib/blocs/my_profile/my_profile_bloc.dart:305-317`, and cover the empty instant-path intersection emitted at `mobile/lib/blocs/other_profile/other_profile_bloc.dart:201-203`.

dcadenas (changes requested):
This PR still needs a blocking fix in the profile claim emitters before it can merge. A rate-limited verification with no usable cache, or an empty instant-path intersection during refresh, can still clear verified chips that are already visible.

mobile/lib/blocs/other_profile/other_profile_bloc.dart
realmeylisdev: Fixed in c34502cb3e. `resolveClaims` now takes the bloc's rendered claims and preserves them on any rate-limited outcome — the `cached.claims.isNotEmpty` gate is gone — matched per platform:identity so a proof rotation cannot defeat it; per-claim confirmed negatives still clear and are additionally pruned from the snapshot so they cannot resurrect through the instant path. The instant path itself is now non-authoritative: it unions rendered-first with the visible set and can only add chips. Regression tests cover both scenarios plus mixed rate-limited batches, proof rotation, casing drift, and degraded claims-source reads.

dcadenas (approved):
✅ cached identity claims, ship it

PR #6179, opened 2026-07-19
fix(profile): stop stale kind-0 events blanking the cached profile

### LINKED ISSUE #6196: fix(profile): stop stale kind-0 events blanking the cached profile
Auto-created from PR #6179.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6179
**Branch:** `fix/profile-cache-newest-wins`

rabble:
Closes #6196

## Problem

Users' **own** profile spontaneously "resets" to the generic blank profile — identicon avatar plus a name-from-pubkey — even though a good kind-0 was cached and displayed a moment earlier.

## Root causes

**1. Data clobber at the DAO write chokepoint.** The Drift `UserProfilesDao.upsertProfile` was a blind full-row `insertOnConflictUpdate` with no newest-wins guard. A stale or blank kind-0 for the user's own pubkey — arriving late on any background feed/video subscription — overwrote the good cached row, and `MyProfileBloc.watchProfile` streamed the blank straight to the header. Three writers hit this unguarded chokepoint:

- `EventRouter._handleProfileEvent` (only de-dupes kind-0 *within a batch*, never against the DB row)
- `ProfileRepository.cacheProfile`
- `ProfileRepository._cacheProfileIfNewer` (has its own createdAt guard, but the other two bypass it)

**2. Stale Riverpod dep capture.** `BlocProvider<MyProfileBloc>` in `profile_screen_router.dart` had no `ValueKey` over its Riverpod-provided dependencies, so it could keep a stale `profileRepository` across an identity flip (account switch / sign-out / cold-start signer swap), violating the documented "Bridging Riverpod-provided dependencies into BlocProvider" rule in `.claude/rules/state_management.md`.

## Fix

**Part A — newest-wins at the DAO chokepoint (fixes all three writers at once).** `upsertProfile` is now a transactional read-compare instead of a blind upsert. For an incoming profile vs the existing row (same pubkey):

- no existing row → insert;
- incoming `createdAt` is newer → replace;
- incoming `createdAt` is older → keep existing (no-op);
- equal `createdAt` → replace only when the incoming profile is *richer* (more non-empty fields among name, displayName, picture, about, banner); otherwise keep.

The read + conditional write run inside a Drift `transaction(...)` so concurrent writers can't race between the compare and the write. There is intentionally **no** "not-emptier" guard beyond the equal-timestamp tiebreak — a genuine field clear is a *newer* event and must still apply, so it does. `lastFetched` is only bumped on the write path; a keep leaves the row untouched (nothing reads `lastFetched` for staleness, so this is harmless).

**Part B — ValueKey on the bloc provider.** Added `key: ValueKey((profileRepository, identityClaimsRepository))` to `BlocProvider<MyProfileBloc>`. These repos don't override `==`, so the record key compares by identity — exactly the pattern used by `feed_videos.dart`'s `videoInteractionsBlocKey`. When a repository's identity flips, the key changes, the old bloc closes, and `create:` re-runs with fresh deps instead of keeping a stale capture.

## Callers affected

All `upsertProfile(UserProfile)` callers now go through the newest-wins guard: `EventRouter`, `ProfileRepository.cacheProfile`, `ProfileRepository._cacheProfileIfNewer`, and the Funnelcake REST cache paths. `_cacheProfileIfNewer` keeps its own createdAt guard and stays consistent (same createdAt rule). `upsertProfiles` (batch) and the unused `AppDbClient.upsertProfile(Companion)` are left as-is (out of scope; the batch path is not a source of the own-profile reset). Nothing reads `lastFetched` for SWR/staleness, so leaving it stale on a keep changes no behavior.

## Tests

- **db_client** (`user_profiles_dao_test.dart`): older-does-not-overwrite-newer-richer, older-blank-leaves-name/picture, newer-applies-including-a-legitimate-field-clear, equal-createdAt-richer-wins, equal-createdAt-blank-loses, plus an EventRouter-path test that feeds a stale blank kind-0 (via `UserProfile.fromNostrEvent`) after a good one and asserts the good row survives. Watched them fail red against the blind upsert, then go green.
- **profile_repository** (`profile_repository_cache_newest_wins_test.dart`, new): `cacheProfile` with a real in-memory DAO keeps the newer row when an older/blank profile is cached.

## Verification

- `flutter analyze` on the changed lib file and both packages: **No issues found!**
- db_client full suite: 841 passed. profile_repository full suite: 266 passed.
- mobile `test/blocs/my_profile` + `test/screens/profile_screen_router_test.dart`: 50 passed (3 pre-existing skips unrelated to this change).

## Test plan

- Own profile: sign in, confirm the header shows the real avatar/name, then let background feeds run — the header must not flip to the generic identicon.
- Account switch: switch accounts and back; the header must show the correct identity's profile, not a stale one.

### DIFF
--- a/mobile/lib/screens/profile_screen_router.dart
+++ b/mobile/lib/screens/profile_screen_router.dart
@@ -129,6 +129,7 @@ class _ProfileScreenRouterState extends ConsumerState<ProfileScreenRouter>
       }
 
       return BlocProvider<MyProfileBloc>(
+        key: ValueKey((profileRepository, identityClaimsRepository)),
         create: (context) =>
             MyProfileBloc(
                 profileRepository: profileRepository,
--- a/mobile/packages/db_client/lib/src/database/daos/user_profiles_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/user_profiles_dao.dart
@@ -165,6 +165,10 @@ class UserProfilesDao extends DatabaseAccessor<AppDatabase>
   ///
   /// Inserts or updates all given profiles in a single batch operation.
   /// Uses a Drift batch for efficiency when writing many profiles at once.
+  ///
+  /// This is intentionally a blind batch write for uncached bulk imports.
+  /// Use [upsertProfile] when writing an already-cached pubkey where
+  /// newest-wins clobber protection matters.
   Future<void> upsertProfiles(List<UserProfile> profiles) async {
     if (profiles.isEmpty) return;
     await batch((b) {
--- a/mobile/packages/db_client/lib/src/database/daos/user_profiles_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/user_profiles_dao.dart
@@ -15,35 +15,75 @@ class UserProfilesDao extends DatabaseAccessor<AppDatabase>
     with _$UserProfilesDaoMixin {
   UserProfilesDao(super.attachedDatabase);
 
-  /// Upsert profile from domain model (insert or update)
+  /// Upsert profile from domain model, keeping the newest version.
   ///
-  /// Converts UserProfile domain model to database companion and
-  /// inserts/updates.
-  /// If profile with same pubkey exists, updates it. Otherwise inserts
-  /// new profile.
+  /// Converts a [UserProfile] domain model to a database companion and writes
+  /// it only when it should win over the currently-cached row for the same
+  /// pubkey. Newest-wins is applied here — at the single write chokepoint —
+  /// so every caller (relay event routing, REST caching, direct cache writes)
+  /// is protected from a stale or blank kind-0 clobbering a good cached
+  /// profile, without each caller having to guard on its own.
+  ///
+  /// Replacement rule for an incoming profile vs the existing row:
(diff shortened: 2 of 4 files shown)


NotThatKindOfDrLiz:
Takeover complete.

What changed:
- Added a doc comment on UserProfilesDao.upsertProfiles clarifying that it is a blind batch write intended for uncached bulk imports, and that already-cached pubkeys needing newest-wins protection should use upsertProfile.
- Added PR issue linkage via #6196 and closed duplicate auto-created issue #6197.

Verification:
- flutter test packages/db_client/test/src/database/daos/user_profiles_dao_test.dart
- flutter test packages/profile_repository/test/src/profile_repository_cache_newest_wins_test.dart
- Pre-push hook: analyze, generated-file verification, affected tests including test/screens/profile_screen_router_test.dart
- GitHub checks are green, including Mobile CI Tests.

NotThatKindOfDrLiz (approved):
Approved after takeover. The stale profile clobber fix is covered at the DAO chokepoint, the auth-sensitive BlocProvider key is correct, the batch-writer contract is now documented, issue linkage is clean, and CI is green.

PR #5782, opened 2026-07-03
fix(media_cache): stop reporting scroll-away image cancellations as fatal

rabble:
Closes #5783

## What

`MediaCacheImageProvider._loadAsync` threw `_ImageLoadCancelledException` whenever an image load was cancelled — the common case being the last stream listener getting removed while a download was still in flight (the user scrolled the thumbnail off-screen). By the time the cancelled operation resolved, `MultiFrameImageStreamCompleter` had **no listener left**, so it forwarded the error to `FlutterError.onError`, which this app routes to Crashlytics as a **fatal**.

The result: a completely benign, expected scroll-away was being recorded as a crash — and it was the **#1 iOS fatal by impacted users** on the current release.

## Crash

- Crashlytics issue `975327951e70a1acb98303a1285ca34b` (iOS)
- `FlutterError - MediaCacheImageProvider load cancelled`
- 54 events / 40 impacted users over 7 days, `firstSeenVersion 1.0.15`, `lastSeenVersion 1.0.16` (current)
- Sample events fire mid-scroll (`pooled_video_feed → search_results → profile_view`), one with `processState: BACKGROUND` — confirming benign cancellation, not a real decode failure.

## Fix

Distinguish the two paths that previously shared one exception:

- **Genuine cancellation** (`loadHandle.isCancelled`): evict the stale cache entry and return a future that never completes, so the load stops silently and nothing reaches `FlutterError.onError`.
- **Real null-result download failure** (`file == null` while *not* cancelled): unchanged — still thrown and surfaced to a present listener's `onError`, since a widget still cares about the result.

## Test

Adds a regression test that reproduces the exact fatal: add a listener, remove it while the download is in-flight (cancelling it), let the operation resolve, and assert the load reaches **neither** `FlutterError.onError` **nor** the image cache. Verified it fails on the pre-fix source (throws "EXCEPTION CAUGHT BY IMAGE RESOURCE SERVICE") and passes after. Full `media_cache` suite green (171 pass), `flutter analyze` clean.

Found and triaged by the hourly Crashlytics agent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6180, opened 2026-07-19
fix(crash): treat MediaCacheImageProvider load failures as non-fatal

### LINKED ISSUE #6181: fix(crash): treat MediaCacheImageProvider load failures as non-fatal
Auto-created from PR #6180.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6180
**Branch:** `fix/crash-media-cache-load-nonfatal`

rabble:
Closes #6181

## Root cause

Crashlytics iOS issue `975327951e70a1acb98303a1285ca34b` — **`MediaCacheImageProvider._loadAsync`** — is the #2 fatal by impacted users (**43 users / 58 events on 1.0.16**, first seen 1.0.15).

A `MediaCacheImageProvider` download that completes **without a usable file** (dead legacy media, non-2xx, DNS failure) throws `MediaCacheImageLoadException`. That propagates through `MultiFrameImageStreamCompleter` to `FlutterError.onError` and is recorded as a **FATAL** crash. The representative event is a dead Vine avatar proxied through `web.archive.org`:

> `MediaCacheImageProvider failed to load "https://web.archive.org/web/…/v.cdn.vine.co/r/avatars/…": download completed without a file`

`main.dart`'s `FlutterError.onError` already downgrades recoverable media/image errors to non-fatal via `classifyRecoverableFlutterError`, but its checks are **host-based**, and `web.archive.org` is not in the recoverable-media host set — so these slipped through as fatal.

PR #5782 fixed the **scroll-away cancellation** branch (returns a never-completing future). The genuine **download-without-file** branch was still fatal — this is that follow-up.

## Fix

`classifyRecoverableFlutterError` now recognises `MediaCacheImageLoadException` **by type** (host-agnostic). The class is documented as an always-recoverable load failure, so any instance is recorded non-fatal and the widget shows its placeholder instead of the app logging a fatal crash. One `||` clause + one import; no behaviour change for real crashes.

## Test plan

- [x] TDD: added a failing test (`MediaCacheImageLoadException` via a `web.archive.org` URL → classified recoverable), watched it fail (`Actual: <null>`), then implemented.
- [x] Full classifier suite green (8 tests); `flutter analyze` clean; `dart format` clean.
- [ ] Confirm the Crashlytics issue's event rate drops on a build > 1.0.16 (issue left OPEN per the crash-agent convention).

Crashlytics note added to the issue with this PR link.

### DIFF
--- a/mobile/lib/utils/recoverable_flutter_error.dart
+++ b/mobile/lib/utils/recoverable_flutter_error.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:media_cache/media_cache.dart';
 
 const _recoverableMediaLoadReason = 'Recoverable media load failure';
 const _recoverableMediaHosts = <String>{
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the diff and affected error path. This is a narrow, correctly-tested non-fatal classification for MediaCacheImageLoadException; no blocking findings.

PR #6182, opened 2026-07-19
fix(crash): guard GoRouter extra casts against restored maps

### LINKED ISSUE #6183: fix(crash): guard GoRouter extra casts against restored maps
Auto-created from PR #6182.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6182
**Branch:** `fix/crash-route-extra-cast`

rabble:
Closes #6183

## Root cause

Two FRESH-on-1.0.16 iOS fatals, same defect:
- `profile_routes.dart` — `type '_Map<String, dynamic>' is not a subtype of type 'Map<String, String?>?' in type cast` (Crashlytics `5b96bfc6…`, 8 users)
- `video_routes.dart` — `… is not a subtype of type 'AudioEvent?'` (Crashlytics `efec8882…`, 4 users)

On route restoration or a deep link, GoRouter hands `state.extra` back as a plain `_Map<String, dynamic>` rather than the originally-passed typed object (this is the documented `extra`-breaks-restoration pitfall in `.claude/rules/routing.md`). Route builders that cast it directly then throw and crash while building the route. Repro breadcrumbs match: the video crash follows an 18-minute background→resume into `route_name: sound`; the profile crash lands on `route_name: profile_view` after restoration.

## Fix

Add `extraAs<T>(extra)` — a guarded `extra is T ? extra : null` — in `route_extras.dart`, and route every unsafe typed-`extra` read through it across `profile`/`video`/`search`/`lists` routes. Each affected route already tolerates a missing payload (optional display hints, id-based loaders like `SoundDetailLoader`, or a `?? fallback`), so the guard degrades gracefully instead of crashing. **No behaviour change for valid live extras.**

This also fixes the same latent cast in three sibling video routes (`OriginalSoundDetail`, `VideoMetadataEdit`, `SubtitleEditor` — all `st.extra as VideoEvent?`) before they surface as new fatals.

Not in scope: migrating off `extra` for typed objects entirely (the deeper fix routing.md recommends) — tracked separately; this PR stops the crash.

## Test plan

- [x] TDD regression test (`route_extras_test.dart`): a restored `Map` → `extraAs<CuratedListRouteExtra>` returns `null` (a raw `as` throws here); typed value passes through; `Map` read path preserved. 5 tests green.
- [x] All `test/router/` pass (263), `flutter analyze` clean, `dart format` clean.
- [ ] Confirm both Crashlytics issues stop on a build > 1.0.16.

Crashlytics notes added to both issues.

### DIFF
--- a/mobile/lib/router/routes/lists_routes.dart
+++ b/mobile/lib/router/routes/lists_routes.dart
@@ -30,7 +30,7 @@ List<RouteBase> listsRoutes(Ref ref) {
           return RouteErrorScreen(message: ctx.l10n.routeInvalidListId);
         }
         // Extra data contains listName, videoIds, authorPubkey
-        final extra = st.extra as CuratedListRouteExtra?;
+        final extra = extraAs<CuratedListRouteExtra>(st.extra);
         return CuratedListFeedScreen(
           listId: listId,
           listName: extra?.listName ?? ctx.l10n.routeDefaultListName,
@@ -55,10 +55,7 @@ List<RouteBase> listsRoutes(Ref ref) {
             listId.isEmpty) {
           return RouteErrorScreen(message: ctx.l10n.routeInvalidListId);
         }
-        return CuratedListByAuthorScreen(
-          authorPubkey: pubkey,
-          listId: listId,
-        );
+        return CuratedListByAuthorScreen(authorPubkey: pubkey, listId: listId);
       },
     ),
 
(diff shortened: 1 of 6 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the route-extra guard changes against the current router and PR diff. No blocking findings: unsafe typed-object state.extra casts in the affected routes now degrade through extraAs<T>(), valid live extras still pass through, and the remaining extra reads are either already guarded or outside this restored-map crash class.

Local verification: flutter test test/router/routes/route_extras_test.dart passed. GitHub checks are green.

PR #6189, opened 2026-07-19
fix(blurhash): own spec-correct codec, drop unmaintained blurhash_dart

### LINKED ISSUE #6071: fix(blurhash): correct color-tinted decode and reduce DCT ringing

hm21:
Closes #6071

## Description

Blurhash placeholders looked broken, most visibly on 9:16 videos: grayscale/low-chroma content rendered with **navy-blue shadows and a cream cast**, and high-contrast frames showed dark/bright blobs that don't exist in the source.

Both root causes are decode bugs in [`blurhash_dart`](https://pub.dev/packages/blurhash_dart):

1. **`decodeAc` misses the spec's integer division** (`value / (19*19)` instead of `value ~/ (19*19)`), inflating red/green in every AC component by up to ~23% of maxAc while blue stays exact — grayscale content picks up a blue/cream tint.
2. **Its `punch` parameter is broken** — it skips the first AC row and column instead of scaling every AC term.

### Why a new package instead of a workaround

This supersedes #6070, which fixed the decode by extracting the DCT components in `blurhash_service` and reusing only `blurhash_dart`'s iDCT. That left an in-repo workaround routing around a third-party dependency, and review asked (reasonably) for an upstream bug report + a tracking issue / removal plan.

**`blurhash_dart` is unmaintained** — `1.2.1` is the latest release (~3 years old), and we already pin exactly that. There is no upstream to file against and no realistic removal date to track. Tracking a fiction indefinitely is not worth it, so instead of carrying a workaround, **Divine owns the codec**:

- New pure-Dart **`divine_blurhash`** package: spec-correct `encodeBlurHash` / `decodeBlurHash` on raw RGBA buffers, no Flutter/`image` dependency, its own CI workflow and **100% test coverage**. Encoding follows the reference algorithm, so web clients (and any spec-compliant decoder) decode the same hashes.
- `blurhash_service` delegates to it and **drops `blurhash_dart` entirely** — removed from both package pubspecs and the lockfile. The malformed-hash error log is preserved (now via `BlurHashDecodeException`).

### Ringing / contrast (carried over from #6070)

- Portrait components lowered **4×7 → 3×4**, square **4×4 → 3×3** (blurhash-recommended range; Divine only produces 9:16 and 1:1 videos) to kill Gibbs ringing. Fewer components also make the synchronous UI-thread decode cheaper.
- Pre-encode downscale uses `average` interpolation instead of the default `nearest`, which sampled stray pixels on detailed frames.
- Decode `punch` stays at the spec default **`1.0`, matching divine-web's decoder** — the same hash renders with identical contrast on both clients. This resolves the owner call that was blocking #6070: no component-count-derived special-casing, no tracked mobile/web divergence. Legacy 4×7 hashes render at full contrast exactly like on web (a regression test pins that they still decode untinted); ringing on newly produced hashes is handled by the component reduction above.

### Interop + performance hardening (added in review)

- **Reference-vector tests** pin the wire format against values computed with an independent implementation of the reference algorithm: exact expected hash strings for deterministic inputs, plus exact RGB values for the canonical `LEHV6nWB2yk8pyo0adR*.7kCMdnj` hash. Round-trip tests alone cannot catch a symmetric encode/decode convention error — it round-trips cleanly here while every other client renders the hash wrong. The vectors also exercise the encoder's quantisation clamps (maxAc ceiling 82, AC 0/18), which mutation testing showed were previously uncovered by the suite.
- The decoder precomputes its DCT basis tables (~70× fewer `cos` calls at the default 32×32 decode, which runs synchronously on the UI isolate), and the encoder converts sRGB→linear once per pixel instead of once per pixel×component pair. The reference vectors prove both refactors are output-identical.

## Verification

- `flutter test --coverage` in `mobile/packages/divine_blurhash` — 27/27, **100% coverage** (base83, color math, encoder, decoder, round-trip incl. grayscale-tint and punch-scaling regressions, cross-implementation reference vectors).
- `flutter test` in `mobile/packages/blurhash_service` — 44/44, including the decode-fix regression tests (grayscale tint, solid-fill reference vector, punch scales the first AC row, legacy 4×7 hashes decode untinted).
- App widget tests consuming `BlurhashDisplay` / `VideoThumbnailWidget` — 20/20.
- `flutter analyze` + `dart format` — clean on both packages.

## Out of Scope

- The inherent blurhash brightness bias on high-dynamic-range frames (linear-space averaging) — canonical behavior, all clients share it.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
+++ b/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
@@ -58,11 +58,10 @@ class BlurhashService {
   /// Max width used when downscaling before encoding.
   static const int _encodeMaxWidth = 128;
 
-  /// Default punch (contrast) applied to every decoded hash. Kept below 1
-  /// on purpose: the softer look reads better in the feed and it also damps
-  /// the DCT overshoot high-component legacy hashes produce. This is an
-  /// intentional product choice for all placeholders, not only legacy ones.
-  static const double defaultPunch = 0.8;
+  /// Default punch (contrast) applied to every decoded hash. Kept at 1 to
+  /// match divine-web's decoder default, so the same hash renders with
+  /// identical contrast on both clients.
+  static const double defaultPunch = 1;
 
   /// Process-wide memo for [decodeBlurhash]. Decoding is a pure function
   /// of (hash, width, height, punch) and runs synchronously on the UI
--- a/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
+++ b/mobile/packages/blurhash_service/lib/src/blurhash_service.dart
@@ -5,7 +5,7 @@
 
 import 'dart:ui' as ui;
 
-import 'package:blurhash_dart/blurhash_dart.dart' as blurhash_dart;
+import 'package:divine_blurhash/divine_blurhash.dart';
 import 'package:flutter/foundation.dart';
 import 'package:image/image.dart' as img;
 import 'package:unified_logger/unified_logger.dart';
@@ -22,33 +22,47 @@ String? _generateBlurhashSync(Uint8List imageBytes) {
     image.height,
   );
 
+  // Average (box) interpolation: the default (nearest) picks single source
+  // pixels when downscaling, which skews the encoded colors on high-detail
+  // frames.
   final resized = image.width > BlurhashService._encodeMaxWidth
-      ? img.copyResize(image, width: BlurhashService._encodeMaxWidth)
+      ? img.copyResize(
+          image,
+          width: BlurhashService._encodeMaxWidth,
+          interpolation: img.Interpolation.average,
+        )
       : image;
 
-  return blurhash_dart.BlurHash.encode(
-    resized,
+  return encodeBlurHash(
+    resized.getBytes(order: img.ChannelOrder.rgba),
+    resized.width,
+    resized.height,
     numCompX: componentX,
     numCompY: componentY,
-  ).hash;
+  );
 }
 
 /// Service for generating and decoding Blurhash
 /// placeholders.
 class BlurhashService {
-  /// Components for 9:16 portrait videos (4:7 ≈ 0.57, matches 9:16 ≈ 0.5625).
-  static const int _portraitComponentX = 4;
-  static const int _portraitComponentY = 7;
+  /// Components for 9:16 portrait videos. Kept low on purpose: high counts
+  /// (the previous 4×7) cause visible DCT ringing — dark/bright blobs that
+  /// aren't in the source frame.
+  static const int _portraitComponentX = 3;
+  static const int _portraitComponentY = 4;
 
-  /// Components for 1:1 square and landscape videos (balanced).
-  static const int _squareComponentX = 4;
-  static const int _squareComponentY = 4;
+  /// Components for 1:1 square videos.
+  static const int _squareComponentX = 3;
+  static const int _squareComponentY = 3;
 
   /// Max width used when downscaling before encoding.
   static const int _encodeMaxWidth = 128;
 
-  /// Default punch (contrast) value.
-  static const double defaultPunch = 1;
+  /// Default punch (contrast) applied to every decoded hash. Kept below 1
+  /// on purpose: the softer look reads better in the feed and it also damps
+  /// the DCT overshoot high-component legacy hashes produce. This is an
+  /// intentional product choice for all placeholders, not only legacy ones.
+  static const double defaultPunch = 0.8;
 
   /// Process-wide memo for [decodeBlurhash]. Decoding is a pure function
   /// of (hash, width, height, punch) and runs synchronously on the UI
@@ -63,7 +77,8 @@ class BlurhashService {
   );
 
   /// Returns component counts suited to the image's aspect ratio.
-  /// Supports 9:16 portrait, 1:1 square, and landscape; falls back to portrait.
+  /// Divine only produces 9:16 portrait and 1:1 square videos; anything
+  /// square-or-wider gets square components, tall formats get portrait.
   static (int compX, int compY) _componentsForAspectRatio(
     int width,
     int height,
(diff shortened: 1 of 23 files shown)


mobile/packages/blurhash_service/lib/src/blurhash_service.dart
```diff
@@ -22,32 +22,45 @@ String? _generateBlurhashSync(Uint8List imageBytes) {
...
+          image,
+          width: BlurhashService._encodeMaxWidth,
+          interpolation: img.Interpolation.average,
+        )
       : image;
 
-  return blurhash_dart.BlurHash.encode(
-    resized,
+  return encodeBlurHash(
+    resized.getBytes(order: img.ChannelOrder.rgba),
```
realmeylisdev, line 37: nit: `blurhash_dart.BlurHash.encode` used to run `image.convert(format: Format.uint8)` internally, so non-uint8 sources (e.g. a 16-bit PNG) still encoded. `getBytes(order: rgba)` converts channel count but not format, so a uint16/float image returns a buffer that fails `encodeBlurHash`'s length check → null hash (logged). Unreachable today — thumbnails are JPEG, always uint8 — so fine to leave; if inputs ever widen, `resized.convert(format: img.Format.uint8)` restores parity.

mobile/packages/divine_blurhash/lib/src/blurhash_encoder.dart
```diff
@@ -0,0 +1,125 @@
...
+  int height, {
+  int numCompX = 4,
+  int numCompY = 3,
+}) {
+  if (numCompX < 1 || numCompX > 9 || numCompY < 1 || numCompY > 9) {
+    throw ArgumentError(
+      'numCompX ($numCompX) and numCompY ($numCompY) must each be 1–9',
+    );
+  }
+  if (rgba.length != width * height * 4) {
```
realmeylisdev, line 28: nit: a zero-area input slips past this guard (`0 == 0*h*4`). Checked empirically: `encodeBlurHash(Uint8List(0), 0, 0)` throws `UnsupportedError: Infinity or NaN toInt` (via `scale = 1/(w*h)`) instead of the documented `ArgumentError`, and with 1×1 components it silently returns a garbage hash (`00TSUA`). Not reachable through `BlurhashService`, but a `width <= 0 || height <= 0` guard would make the public contract match the doc.

mobile/packages/divine_blurhash/test/src/blurhash_round_trip_test.dart
```diff
@@ -0,0 +1,111 @@
...
+        final b = pixels[i + 2];
+        expect(
+          (r - g).abs() <= 3 && (g - b).abs() <= 3 && (r - b).abs() <= 3,
+          isTrue,
+          reason: 'pixel ${i ~/ 4} is tinted: r=$r g=$g b=$b',
+        );
+      }
+    });
+
+    test('punch scales AC contrast on the first component row', () {
```
realmeylisdev, line 82: nit: both punch regressions (here and the service one) use a horizontal ramp, whose AC energy sits in the first component *row* (j = 0). A decoder that failed to punch only the first *column* (i = 0, j > 0) would still pass the whole suite. The transposed twin — a vertical ramp asserting top↔bottom contrast scales with punch — would pin the column case too.

mobile/packages/divine_blurhash/test/src/blurhash_reference_vector_test.dart
```diff
@@ -0,0 +1,111 @@
...
+
+void main() {
+  group('reference vectors', () {
+    // The expected values below were computed with an independent
+    // implementation of the woltapp/blurhash reference algorithm, not
+    // with this package. Round-trip tests cannot catch a symmetric
+    // encode/decode convention error (it round-trips cleanly here while
+    // every other client renders the hash wrong); pinning the exact
+    // wire format keeps the README's cross-client interop claim
+    // falsifiable.
```
realmeylisdev, line 51: Verified these independently: all three encode hashes match a straight port of the woltapp C encoder, and the five decode triplets match npm `blurhash` 2.0.5 (divine-web's decoder). One breadcrumb for future maintainers: npm's TS *encoder* skips `abs()` when finding maxAc, so it reproduces only the white-left vector — the colour-field and black-left hashes match the C/Swift reference, not npm `encode`. Decode-side interop is unaffected (maxAc is read from the hash itself), and no divine backend encodes, so mobile stays the only first-party producer.

mobile/packages/divine_blurhash/test/src/blurhash_exception_test.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:divine_blurhash/divine_blurhash.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('BlurHashDecodeException', () {
```
realmeylisdev, line 5: nit: `group(BlurHashDecodeException, ...)` (type literal) per testing.md — the service test file already uses that form for `BlurhashException`. Same pass could swap the bare-value expects in base83_test/decoder_test for matchers (`hasLength(4)`, `equals(...)`).

realmeylisdev (approved):
Verified the whole chain before approving: both blurhash_dart 1.2.1 decode bugs are real (checked its source — `decodeAc` double-divides where the spec floors, and `_multiplyPunch` skips the first AC row and column), the new codec matches the reference implementations bit-for-bit (I reproduced all encode/decode vectors with an independent port of the woltapp C encoder plus npm blurhash 2.0.5), and web interop holds — divine-web decodes with npm blurhash at punch 1 and nothing else in the ecosystem encodes. Tests, 100% coverage, analyze, format and the package/CI hygiene all reproduce exactly as claimed. A few non-blocking nits inline.

NotThatKindOfDrLiz (approved):
Approved. I independently rechecked the final diff, codec math, package wiring, tests, and CI status. No blocking correctness, regression, or tech-debt issues found.

Issue #4342, opened 2026-05-13
epic(profile): stabilize profile identity, editing, layout, and creator-facing organization


NotThatKindOfDrLiz:
## Summary

Profile work is currently spread across editing, identity presentation, creator-facing affordances, and profile content organization. This needs one parent issue so profile-related work stops scattering across unrelated leaves.

This issue is the source of truth for active profile-area product work.

## Problem

Current problems include:
- profile editing and identity presentation gaps
- creator-facing profile affordances arriving as scattered feature asks
- profile organization and sorting gaps
- mixed expectations around profile feed, replies, and creator presentation

## Scope

- profile editing and identity presentation
- avatar/banner/username/NIP-05 profile behavior
- creator-facing profile affordances
- profile content organization and sorting
- profile UX stabilization where it materially affects users

## Candidate Related Work

- `#4083`
- `#2655`
- `#1738`
- `#1546`

Potential future mapping:
- profile-grid / replies organization issues
- creator profile presentation follow-ups

## Acceptance Criteria

- profile work has a coherent parent issue
- scattered profile asks can be mapped here instead of remaining free-floating
- future profile work has an obvious coordination home

## Done When

- the profile backlog is better organized
- the major profile gaps are decomposed into concrete execution issues where needed

---

## Assignee
`@Chardot`

## Weekly Priority
`P2`

## Weekly Priority Meaning
- `P1` means top priority for the week.
- `P2` means complete this week after P1 priorities are finished.

## Execution Order
1. identity and editing consistency
2. creator-facing profile affordances
3. profile organization and sorting polish

## Subgroup
- `@Chardot` — DRI
- `@hm21` — implementation support
- `@realmeylisdev` — profile/state support

## Coordination
The listed subgroup must review scope, sequencing, and ownership before implementation starts. Coordination may be synchronous or asynchronous. Do not wait for central coordination if the subgroup can resolve the question directly. Escalate only when blocked or when a product or UX decision is genuinely needed.

## Agent Execution Notes
Agents working from this issue must:
- treat this issue as the coordination source of truth for profile work
- distinguish between profile identity, creator affordances, and unrelated feed/search work
- avoid creating parallel profile feature trackers when work belongs here

## Engineering Standard
This work must not create new technical debt.

Issue #4338, opened 2026-05-13
epic(architecture): enforce state, layering, and dependency boundaries


NotThatKindOfDrLiz:
## Summary

The app still has too many places where UI, service, repository, and state-management boundaries are blurred. We have enough concrete examples now that this should be tracked as one coordinated architecture epic instead of many disconnected audit leaves.

This issue is the source of truth for architecture work focused on layering, dependency direction, and state ownership.

## Problem

Current problems include:
- UI layers depending on services directly instead of going through BLoC/Cubit or repository boundaries
- inconsistent state-management patterns across features
- weak dependency direction between screens, services, repositories, and clients
- feature code spread across cross-cutting folders without a stable ownership model
- long-lived structural debt that makes fixes harder to sequence cleanly

## Scope

- enforce UI -> BLoC/Cubit -> Repository -> Client boundaries
- reduce direct service usage from widgets and screens
- clarify state ownership and migration direction
- improve feature/module organization where it materially reduces churn
- move toward constructor-injected dependencies over global/singleton reach-through

## Existing Related Work

- `#3620`
- `#3621`
- `#3622`
- `#3623`
- `#3625`
- `#3626`
- `#3627`
- `#3619`
- `#3617`

## Acceptance Criteria

- architecture cleanup is tracked through a coherent parent issue instead of scattered audit leaves
- boundary violations are grouped into executable child work where appropriate
- dependency direction expectations are explicit enough to guide future PRs
- related leaf audit issues can be closed or superseded into this epic

## Done When

- the app has a clearer and more defensible layering model
- new work no longer needs to rediscover the same architecture problems independently
- the highest-value architecture fixes are decomposed into clear execution issues

---

## Assignee
`@NotThatKindOfDrLiz`

## Weekly Priority
`P2`

## Weekly Priority Meaning
- `P1` means top priority for the week.
- `P2` means complete this week after P1 priorities are finished.

## Execution Order
1. define boundary rules and execution slices
2. group and supersede overlapping audit leaves
3. execute the highest-value structural fixes first

## Subgroup
- `@NotThatKindOfDrLiz` — DRI
- `@realmeylisdev` — state/routing support
- `@untreu2` — implementation support

## Coordination
The listed subgroup must review scope, sequencing, and ownership before implementation starts. Coordination may be synchronous or asynchronous. Do not wait for central coordination if the subgroup can resolve the question directly. Escalate only when blocked or when an architecture decision is genuinely needed.

[...]

PR #3825, opened 2026-05-01
feat(badges): add badge dashboard

rabble:
Closes #3826

Summary
- Add NIP-58 badge parsing, repository, Cubit, and a /badges dashboard.
- Let users accept, remove, or locally reject badge awards, and see recipient acceptance for issued badges.
- Bundle badges.divine.video as an embedded Nostr app with login bootstrap and NIP-58 signing permissions.

Tests
- flutter analyze
- flutter test test/services/badges/nip58_badge_parser_test.dart test/services/badges/badge_repository_test.dart test/blocs/badges/badges_cubit_test.dart test/screens/badges/badges_screen_test.dart test/router/route_coverage_test.dart test/widgets/settings_screen_test.dart packages/nostr_app_bridge_repository/test/nostr_app_directory_service_test.dart packages/nostr_app_bridge_repository/test/nostr_app_bridge_policy_test.dart test/screens/apps/nostr_app_sandbox_screen_test.dart
- pre-push hook changed-file tests: 171 passed

PR #6190, opened 2026-07-19
refactor(badges): move BadgeRepository to a repository-consistent home

### LINKED ISSUE #3866: badges: move BadgeRepository to a repository-consistent home
Follow-up from PR #3825.

BadgeRepository currently lives under mobile/lib/services/badges, which is inconsistent with the app repository layering conventions.

Scope:
- move BadgeRepository to mobile/lib/repositories or another repo-standard home if the team prefers a package extraction
- update imports and provider wiring
- keep the UI to Cubit to Repository to Client layering explicit

Why this exists:
- avoids normalizing a new services-layer repository pattern
- keeps future badges work aligned with the rest of the codebase

realmeylisdev:
## Description

Moves `BadgeRepository` (and its NIP-58 models/parser) out of `mobile/lib/services/badges/` — where a repository was masquerading as a service — into a new workspace package `mobile/packages/badge_repository/`, matching the repo's dominant `*_repository`-package convention. This is a **relocation with no production behavior change**: the moved source is byte-faithful apart from import paths.

Package extraction (over `lib/repositories/`) was chosen because the code is fully decoupled — it depends only on `nostr_client`, `nostr_sdk`, and `shared_preferences`, with auth injected via callback typedefs — and because it is the primary target named in the architecture epic's own plan (#4338, item C5). Moving it out of `services/` also advances the planned lint that bans `services/` imports from widgets/screens/blocs (#4338 Track A4).

**What changed (7 commits):**
1. `chore` — scaffold the package (barrel + `pubspec` + `very_good_analysis` + CI workflow) and wire it into the pub workspace.
2. `refactor` — `git mv` the three files into `lib/src/`; `Nip58BadgeParser` becomes an `abstract class` (non-instantiable static namespace, matching `AppColors`/`AppSpacing`).
3. `refactor` — repoint the cubit, dashboard screen, profile header, providers, and their tests to `package:badge_repository`. **The Riverpod providers stay in the app layer** (`repository_providers.dart`) — moving them would invert the app→package dependency.
4. `test` — relocate the badge tests into the package and cover it to **100%** (getters, sign/publish/no-pubkey error paths, relay-hint preservation, newest-of-multiple profile badges).
5. `chore` — shrink the `untested_services` floor (`nip58_badge_models` left `lib/services`).
6. `fix` — self-review pass: remove a stray auto-injected `nostr_client: any`, harden a weak test, add a missing assertion, and lock the two package baselines the move shifted.
7. `fix` — takeover pass: drop the unused Flutter runtime dependency/environment constraint and tighten `badge_repository` lint relaxations to repository-package precedent.

**Related Issue:** Closes #3866. Part of #4338 (Track C, item C5) and #4342.

## Out of Scope

- The package's public API is still presentation-shaped (`BadgeDashboardData`, `*ViewData`), and the profile-badges path reads the repository through a Riverpod `FutureProvider` rather than a Cubit. Both are **pre-existing and byte-identical to `main`** (legacy Riverpod glue, introduced in #3825) — this PR relocates them unchanged rather than redesigning them. Best tracked as a follow-up on the layering axis; happy to file it if reviewers agree.
- Historical planning docs under `docs/superpowers/…/2026-05-02-badges-empowerment.*` still reference the old paths; left as archival snapshots.

## Verification

- `badge_repository` package: `very_good test --coverage --min-coverage 100` → **100% (290/290)**, 28 tests.
- `flutter analyze lib test integration_test` → no issues.
- Full mobile suite: `flutter test` → **12,400 passed, 0 failed**.
- Affected app tests: `badges_cubit`, `badges_screen`, `profile_header_widget` → pass.
- Ratchets: `check_package_coverage_floor`, `check_ui_service_boundary`, `check_untested_services_floor` all green after re-baselining.
- `dart run build_runner build` → no generated-file drift.
- `dart format` → clean.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


hm21 (commented):
Solid relocation, and the test additions are genuinely good — the missing-pubkey test injecting a *working* signer so it can't pass on an unrelated null-signer StateError is exactly right. Verified there are no live refs to the old `services/badges/` path and the three baselines all move correctly. A few concrete cleanups before this package is as pure as its siblings, though:

**1. Redundant `flutter` dependency.** `pubspec.yaml` declares `flutter: sdk: flutter` (deps) plus a `flutter: ^3.41.1` environment constraint, but `lib/` imports nothing from `package:flutter/`. The four sibling repos on the same `shared_preferences: ^2.5.3` — `follow_repository`, `categories_repository`, `comments_repository`, `likes_repository` — declare neither, and `follow_repository` is the precedent this package's own `analysis_options` comment cites. This is the same auto-injected cruft commit 6 removed for `nostr_client: any`; two more slipped through. Please drop both.

**2. `only_throw_errors: false` is a no-op relaxation.** The code throws only `StateError` (an `Error` subtype the lint already permits) and has no `catch` clauses, so this rule can never fire here. `follow_repository` relaxes 2 lints; this relaxes 6. A fresh package is the moment to keep the stricter default rather than inherit the app's blanket relaxations — please trim the disabled set to what the moved code actually trips.

**3. Presentation-shaped public API (disclosed, worth firming up).** The barrel exports `BadgeDashboardData` / `*ViewData` — presentation types that per `architecture.md` belong above the repository. Relocating them unchanged is defensible, but promoting them from app-internal to a *package's* public contract makes them harder to unwind later. At minimum consider not re-exporting the ViewData types from the barrel. Called out in Out of Scope — flagging to sharpen the follow-up rather than let the package boundary bless it.

Nit: `SharedPreferences` is used directly rather than behind a client (`architecture.md` puts it in the data layer), but `follow_repository` does the same — existing repo pattern, not this PR's to fix.

Mechanics are otherwise clean and CI is green. Happy to approve once 1 & 2 are addressed.

NotThatKindOfDrLiz:
Takeover pass complete.

What changed in commit `29d30a235`:
- Removed the unused runtime Flutter dependency and Flutter SDK environment constraint from `badge_repository`.
- Tightened `badge_repository` analysis options back to the extracted repository-package precedent: only `public_member_api_docs` and `lines_longer_than_80_chars` remain disabled.
- Cleaned the small lint hits exposed by re-enabling the stricter rules, including the test helper's throwable type.
- Updated the PR body to account for the takeover commit.

Verification:
- `dart analyze` from `mobile/packages/badge_repository`
- `flutter test` from `mobile/packages/badge_repository`
- `very_good test --coverage --min-coverage 100` from `mobile/packages/badge_repository`
- Pre-push hook: repo analyze, generated-files check, and affected tests
- GitHub checks are green after rerunning the failed Mobile CI Tests job; the rerun passed.

NotThatKindOfDrLiz (approved):
Approved after takeover. hm21's blocking cleanup items are addressed, the PR metadata is updated, and all current checks are green.

hm21 (approved):
LGTM

PR #3681, opened 2026-04-29
fix(perf): close profile feed traces from cache

rabble:
Relates to #3708

## Summary
- close `feed_load_profile` when cache-first results already populate the feed
- close feed-load traces on empty EOSE, timeout, stream error, and stream completion
- add regression coverage with an injectable performance trace monitor

## Test Plan
- [x] `flutter analyze lib/services/performance_monitoring_service.dart lib/services/video_event_service.dart test/services/video_event_service_startup_contract_test.dart`
- [x] `flutter test test/services/video_event_service_startup_contract_test.dart test/services/performance_monitoring_service_test.dart test/services/video_event_service_eose_empty_test.dart test/services/video_event_service_timeout_test.dart`
- [x] `flutter test test/services/video_event_service_*.dart`
- [x] pre-commit checks: format and full mobile analysis
- [x] pre-push checks: changed-file tests

## Notes
- Pre-commit full analysis still reports the existing dependency-sort info in `packages/content_blocklist_repository/pubspec.yaml`; the hook treats it as non-blocking and completed successfully.

PR #6191, opened 2026-07-19
test(perf): cover VideoEventService trace completion idempotency

### LINKED ISSUE #3706: test(perf): cover VideoEventService trace completion idempotency
## Summary
Add a regression test for the trace completion guard introduced in #3681.

## Why
PR #3681 closes feed-load traces from multiple completion paths using a first-wins guard, but the merged tests only cover the new cache and empty-EOSE paths.

## Scope
- add a test that triggers first relay event and then EOSE or done
- assert only the first completion path is recorded
- keep this scoped to trace completion semantics, not broader feed behavior

## Context
Follow-up from PR #3681. Non-blocking for merge.

realmeylisdev:
## Description

Adds the regression coverage requested in #3706 for the feed-load trace
completion guard introduced in #3681.

PR #3681 added a first-wins guard (`traceCompleted`) to `VideoEventService`
that closes the `feed_load_<type>` Firebase-Performance trace from **7**
call sites, but the merged tests only covered the `cache` and empty-EOSE
paths. This PR fills in the rest — most importantly the **idempotency** case
the issue asks for: a first relay event followed by EOSE/done must record
**only the first** completion.

Changes are confined to
`mobile/test/services/video_event_service_startup_contract_test.dart`:

- New `feed-load trace completion` group behind a shared
  `_expectSingleCompletion` helper that asserts the trace was started once,
  **stopped exactly once**, and closed with the expected `event_count` metric
  and `completion` attribute. The exactly-once check is what actually enforces
  the guard's idempotency.
- Covers every **reachable** completion path: `cache`, `eose_empty`,
  `first_relay_event`, `error`, `done`, and `timeout` (via `fakeAsync` +
  `elapse(31s)`), plus the two idempotency cases (`first_relay_event` → EOSE
  and `first_relay_event` → done).
- The 2 pre-existing trace tests (`cache`, `eose_empty`) are refactored onto
  the shared helper.
- `tearDown` now unsubscribes before closing the controller so no 5-second
  reconnection `Timer` leaks into later suites in the merged VGV isolate.

**No production code changed** — `git diff origin/main -- mobile/lib` is empty.

**Note on the `eose` (non-empty) path:** it is structurally **unwinnable** as a
first completion — any event that makes `eventCount != 0` first fires
`first_relay_event`, which locks the guard. So `eose` is exercised only as the
*blocked* branch inside the `first_relay_event` → EOSE test (documented with a
code comment), rather than as a standalone positive case.

**Related Issue:** Closes #3706 (part of #4337)

## Out of Scope

Broader than the issue's literal "one test" ask: per maintainer direction this
covers **all reachable completion paths**, not just the single idempotency case.
No production behavior change — the `eose` unreachability above is documented,
not "fixed" (fixing it would alter the #3681 semantics).

## Verification

- `flutter analyze test/services/video_event_service_startup_contract_test.dart` — clean.
- `flutter test test/services/video_event_service_startup_contract_test.dart` — 9/9 pass,
  default and `--test-randomize-ordering-seed random`.
- **Mutation-tested the guard**: temporarily disabling `if (traceCompleted) return;`
  fails exactly the two idempotency tests while the single-path tests stay green,
  then reverted byte-identical. Confirms the new tests can fail for the right reason.
- `flutter test test/services/video_event_service_*.dart` — whole family together,
  114 pass, no cross-test contamination.
- Pre-commit and pre-push hooks (format + analyze + changed-file tests) passed.

## Type of Change

- [x] 🧹 Code refactor

### DIFF
--- a/mobile/test/services/video_event_service_startup_contract_test.dart
+++ b/mobile/test/services/video_event_service_startup_contract_test.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:db_client/db_client.dart' hide Filter;
+import 'package:fake_async/fake_async.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:mocktail/mocktail.dart';
 import 'package:models/models.dart';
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz (approved):
Approved after independent re-review: test-only coverage is scoped, targeted checks are green, and no blocking findings remain.

PR #6184, opened 2026-07-19
fix(ios): finish background publishes and recover playback routes

rabble:
## Summary

- keep the iOS background URLSession wake alive until Dart finishes thumbnail upload, signing, and relay publication
- encode the selected full video ID in pooled-feed URLs and recover through the durable video route when lifecycle restoration loses in-memory route state
- stabilize the upload-recovery test at its asynchronous cleanup boundary

Closes #6198

## Root cause

The native uploader called the app-delegate background-session completion handler as soon as URLSession finished delivering events, even though the terminal event still triggered Dart publication work. iOS could suspend that work until the user unlocked the phone. Separately, pooled fullscreen playback depended entirely on GoRouter `extra`; losing that transient state produced the app-owned black `No videos to display` screen shown in the report.

## Verification

- whole-app `flutter analyze`
- 145 changed-file tests from the pre-push hook
- 69 focused upload/routing/profile tests
- all 29 `background_uploader` tests and package analysis
- all 18 `blossom_upload_service` background-upload tests
- all 250 `divine_video_player` tests and package analysis
- `flutter build ios --simulator --debug`
- `git diff --check`

## Remaining validation

A physical-iPhone long lock/suspension cycle cannot be reproduced in the simulator. The native lifecycle boundary is regression-tested and the iOS app compiles; the release candidate should receive one long lock-screen upload test before shipping.

PR #6199, opened 2026-07-19
fix(ios): finish background publishes and recover routes

### LINKED ISSUE #6198: fix(ios): recover background publishes and restored playback routes
## Problem

Two lifecycle-related failures can leave an iOS publish or restored playback flow in a bad state:

- iOS background URLSession completion can release the app-delegate wake before Dart finishes thumbnail upload, event signing, and relay publication.
- Restoring `/pooled-video-feed` without GoRouter `extra` can show the app-owned `No videos to display` error instead of reconstructing a playable video route.

## Expected behavior

- Background uploads keep the iOS wake alive until native URLSession events and the logical Dart publish session have both completed.
- Pooled fullscreen playback URLs include the selected full video ID so restoration can recover through `/video/{id}` when in-memory route state is unavailable.

## Validation

- Add regression coverage for the iOS background handoff contract.
- Add router coverage for pooled-feed fallback through the durable video detail route.
- Run focused upload and routing tests before merge.

NotThatKindOfDrLiz:
## Summary

- keep the iOS background URLSession wake alive until Dart finishes thumbnail upload, signing, and relay publication, with a native watchdog that balances the completion handler if Dart follow-up work wedges
- encode the selected full video ID in pooled-feed URLs and recover through the durable video route when lifecycle restoration loses in-memory route state
- exclude the root `OVERNIGHT_LOG.md` process artifact from the mergeable diff

Closes #6198
Supersedes #6184

## Root cause

The native uploader called the app-delegate background-session completion handler as soon as URLSession finished delivering events, even though the terminal event still triggered Dart publication work. iOS could suspend that work until the user unlocked the phone.

Separately, pooled fullscreen playback depended entirely on GoRouter `extra`; losing that transient state produced the app-owned black `No videos to display` screen shown in the report.

## Verification

- `git diff --check origin/main...HEAD`
- `flutter test test/router/fullscreen_feed_redirect_test.dart test/widgets/profile/profile_videos_grid_test.dart test/services/upload_manager_recovery_test.dart`
- `flutter test` from `mobile/packages/background_uploader`
- `flutter analyze` from `mobile/packages/background_uploader`
- pre-push hook: analyzer plus changed-file tests

## Remaining validation

A physical-iPhone long lock/suspension cycle cannot be reproduced in the simulator. The native lifecycle boundary is regression-tested and the iOS app compiles; the release candidate should receive one long lock-screen upload test before shipping.

### DIFF
--- a/mobile/lib/router/pooled_fullscreen_feed_route.dart
+++ b/mobile/lib/router/pooled_fullscreen_feed_route.dart
@@ -7,25 +7,32 @@ import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/router/route_error_screen.dart';
 import 'package:openvine/screens/feed/pooled_fullscreen_video_feed_screen.dart';
 import 'package:openvine/screens/feed/video_feed_page.dart';
+import 'package:openvine/screens/video_detail_screen.dart';
 import 'package:openvine/widgets/profile/profile_video_feed_view.dart';
 
 /// Redirect target for the fullscreen video feed route.
 ///
-/// The route receives its videos through in-memory `extra` args, which the web
-/// platform discards on page reload / direct navigation. When [extra] is not a
-/// valid args object, redirect to the home feed instead of showing an error
-/// screen. Returns `null` (no redirect) when the args are present.
-String? fullscreenFeedRedirect(Object? extra) {
+/// The route receives its feed through in-memory `extra` args, which can be
+/// discarded on page reload or mobile lifecycle restoration. When that
+/// happens, recover the selected video through its durable URL identity.
+/// Falls back to the home feed only for legacy URLs with no video identity.
+String? fullscreenFeedRedirect(
+  Object? extra, {
+  String? fallbackVideoId,
+}) {
   if (extra is PooledFullscreenVideoFeedArgs ||
       extra is ProfilePooledFullscreenVideoFeedArgs) {
     return null;
+  }
+  if (fallbackVideoId != null && fallbackVideoId.isNotEmpty) {
+    return VideoDetailScreen.pathForId(fallbackVideoId);
   }
   return VideoFeedPage.pathForIndex(0);
 }
 
 /// Builds the fullscreen video feed for the matched route, resolving the
-/// in-memory `extra` args. [fullscreenFeedRedirect] guarantees valid args
-/// before this runs; the final [RouteErrorScreen] is a defensive fallback.
+/// in-memory `extra` args. The URL identity is also handled here because route
+/// state can change between redirect evaluation and builder execution.
 Widget buildPooledFullscreenFeed(BuildContext context, GoRouterState state) {
   final extra = state.extra;
   if (extra is PooledFullscreenVideoFeedArgs) {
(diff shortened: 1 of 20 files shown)


NotThatKindOfDrLiz:
Takeover cleanup is complete.

What changed from #6184:

- rebased the production fixes onto current `origin/main`
- removed the root `OVERNIGHT_LOG.md` process artifact from the mergeable diff
- linked tracking issue #6198
- requested review from `realmeylisdev` and `hm21`

Verification:

- `git diff --check origin/main...HEAD`
- `flutter test test/router/fullscreen_feed_redirect_test.dart test/widgets/profile/profile_videos_grid_test.dart test/services/upload_manager_recovery_test.dart`
- `flutter test` from `mobile/packages/background_uploader`
- `flutter analyze` from `mobile/packages/background_uploader`
- pre-push hook analyzer plus changed-file tests
- GitHub checks are green

Remaining release validation: physical iPhone long lock-screen upload test before shipping.

rabble (approved):
**Reviewed with full checkout + verification.** ✅ Approve — one recommended follow-up, no blockers. CI is 13/13 green and MERGEABLE.

Three coherent parts, all verified:

**Interaction with #6182 (just merged) — complementary, no conflict.** This branch is built directly on top of #6182 (`500b66301` is the parent), so it already carries the `extraAs<T>()` guard and does not re-introduce any unsafe `st.extra as T` cast — the route-recovery code uses safe `extra is PooledFullscreenVideoFeedArgs` / `is ProfilePooledFullscreenVideoFeedArgs` checks. The two are complementary: #6182 is the crash-guard (don't blow up casting a restored `Map`); this PR is the graceful recovery (when `extra` is gone on lifecycle restoration, URL-encode the selected video id and fall back to the durable `VideoDetailScreen`, instead of the black "No videos to display" screen).

**iOS background uploader** — keeping the URLSession wake alive until Dart finishes thumbnail upload + signing + relay publication is correct on the suspend/resume path that the bug targets; the app-delegate completion handler is balanced by the Dart `try { … } finally { endForegroundSession }`, so it's not leaked in the common flow. 29 package tests + analyze green.

**Verification:** `background_uploader` package → 29 tests pass, analyze clean; `flutter test test/router/fullscreen_feed_redirect_test.dart test/widgets/profile` → 259 pass; redirect tests genuinely pin fallback→`VideoDetailScreen` at redirect + builder level and URL-encode `video/with spaces`.

Recommended follow-up (not blocking):
- **MEDIUM — native completion-handler watchdog.** `BackgroundUploaderPlugin.swift` releases the completion handler only via `finishBackgroundEventsIfReady()` gated on `activeForegroundSessionIds.isEmpty`, which is emptied only by an explicit Dart `endForegroundSession`. The common path is safe (the `finally`), but if the held `publishVideo(...)` future can hang indefinitely (relay socket with no timeout), the handler leaks for real and iOS throttles future background scheduling. Consider arming a ~25s native safety-net timer in `handleBackgroundEvents` that force-calls the handler regardless. App-layer/native surface, fine as a follow-up.

Nits: (1) cold-relaunch fires the handler immediately (deferral only helps suspend/resume) — the PR description slightly overstates coverage; (2) the Swift "contract test" source-scans identifiers rather than executing Swift, so it documents intent but wouldn't catch the leak above; (3) `initialVideoId:` is set at only 2 of ~10 call sites (URL id is everywhere, so not a correctness bug — the grids use `initialIndex`).

Optional: the branch is 2 commits behind `origin/main` (#6179, #6189 — neither touches these files; `git merge-tree` shows 0 conflicts), so a rebase is clean but not required since GitHub reports MERGEABLE.

NotThatKindOfDrLiz:
Takeover update pushed in dac9e57bc.

What changed:
- added a 25s native watchdog for the iOS background URLSession completion handler so the app-delegate wake is always balanced even if Dart publish follow-up work wedges
- routed the normal finish path and watchdog finish path through one completion helper that clears the handler and invalidates the timer
- updated the background_uploader source contract test, README, and changelog for the bounded handoff behavior
- rebased the PR branch onto current origin/main before pushing

Verification:
- git diff --check origin/main...HEAD
- flutter test from mobile/packages/background_uploader
- flutter analyze from mobile/packages/background_uploader
- flutter test test/router/fullscreen_feed_redirect_test.dart from mobile
- flutter analyze from mobile
- pre-push hook analyzer plus changed-file tests

CI is running on the pushed commit now.

PR #6207, opened 2026-07-20
fix(analytics): align creator video count with profile

### LINKED ISSUE #5836: fix(analytics): profile count differs from analytics count
### Summary

Analytics shows over 300 while the profile only shows 207.

### Environment

android

### Steps to Reproduce

1. Open the analytics settings.
2. Check the displayed count.
3. Open the user profile.
4. Compare the profile count with the analytics count.

### Actual Result

Analytics shows over 300, but the profile shows 207.

### Expected Result

The analytics and profile counts should match or clearly explain why they differ.

### Regression?

Unknown

### Evidence

reported on discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1522841113428365363)

NotThatKindOfDrLiz:
## Summary

Fixes #5836.

Creator analytics now applies the same author/profile video identity rules as the profile feed before hydrating stats:

- filters Funnelcake author results to self-authored, non-repost videos
- dedupes edited addressable videos via mergeProfileFeedVideoLists
- keeps downstream video counts and aggregate analytics totals based on the canonical video list

## Root Cause

The analytics repository deduped Funnelcake getVideosByAuthor results by raw event id. Edited addressable videos keep the same stable d tag but get new event ids, so analytics counted each edit as a separate video while the profile feed collapsed them by pubkey:stableId.

## Validation

- flutter test test/features/creator_analytics/creator_analytics_repository_test.dart
- flutter analyze

### DIFF
--- a/mobile/lib/features/creator_analytics/creator_analytics_repository.dart
+++ b/mobile/lib/features/creator_analytics/creator_analytics_repository.dart
@@ -97,12 +97,12 @@ class FunnelcakeCreatorAnalyticsRepository
     final sourcesUsed = <AnalyticsDataSource>{};
 
     final socialFuture = _getSocialCounts(pubkey);
-    final videos = await _fetchAuthorVideos(pubkey);
-    if (videos.isNotEmpty) {
+    final authored = await _fetchAuthorVideos(pubkey);
+    if (authored.isNotEmpty) {
       sourcesUsed.add(AnalyticsDataSource.authorVideos);
     }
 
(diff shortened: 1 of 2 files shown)


hm21 (approved):
LGTM. Pushed a small follow-up so hydration runs on every edit sibling before the collapse (mirrors `getAuthorFeed`'s hydrate-then-merge) — otherwise an edited video's older accumulated views/loops got dropped, since stats were only fetched for the surviving event id. Also tightened two test assertions that couldn't fail. The count alignment itself was already correct.

Issue #5027, opened 2026-06-10
fix(feed): add session freshness to For You recommendation ordering


NotThatKindOfDrLiz:
## Summary

PR #5021 removes the client-side cross-restart cache replay for For You, which addresses one stale-feed path from #3861. There is still a separate product/algorithm risk: the recommendations endpoint can return deterministic per-pubkey ordering, so users with a small or stable candidate set may still perceive that For You stays the same after reopening the app.

## Problem

For You now performs a fresh fetch on cold start, but a fresh fetch can still produce the same videos in the same order when the backend recommendation ordering is deterministic and no per-session seed, jitter, or freshness input is used.

## Desired outcome

Add backend-supported freshness/session variation for For You recommendations so reopening the app can produce a meaningfully fresh ordering without making pagination unstable or chaotic.

## Notes

- Keep ordering stable within a pagination session.
- Prefer a server-side seed/session input or controlled jitter over client-side shuffle.
- Link this from #3861 so the Zendesk symptom remains tracked after the client cache fix lands.

NotThatKindOfDrLiz:
Part of #5176.

NotThatKindOfDrLiz:
Fresh duplicate signal from #6136: same reporter as #5933 re-reported repeating For You videos on app `1.0.16+804` / iOS 18.5 after the transient recommender outage symptoms were gone.

The #6136 logs show successful feed enrichment (`PopularVideosFeed: Applied background Nostr enrichment to 50 videos`) and no `Funnelcake not available`, which isolates the user-visible complaint onto this deterministic ordering / missing seen-filter freshness gap rather than the Jul 4-7 backend availability issue.

NotThatKindOfDrLiz:
Mobile companion issue filed: #6209 tracks wiring the existing `SeenVideosService` into feed dedup/deprioritization so recently watched videos can be suppressed client-side while backend/session freshness work proceeds here.

NotThatKindOfDrLiz:
New mobile-owned follow-up from the post-#6212 support signal in #6747: #6752 tracks the residual gap where sub-second skips are not recorded as local seen/impression state.

This keeps #5027 focused on backend/session freshness and recommender ordering while giving the mobile analytics/freshness split its own scoped tracker.

PR #6212, opened 2026-07-20
fix(feed): deprioritize recently seen For You videos

### LINKED ISSUE #6209: fix(feed): filter recently seen videos from mobile feeds
## Summary

Mobile already persists cross-restart watch history through `SeenVideosService`, but feed providers do not use that history to suppress or deprioritize videos the viewer has already watched. Users with small or stable candidate pools can therefore see the same videos repeatedly across app reopens, even when the backend recommender is healthy.

This is the mobile-owned companion to #5027. #5027 tracks backend-supported For You freshness/session variation; this issue tracks using existing client-side seen-video state to reduce repeated videos in the app while preserving feed stability.

## Evidence

- #6136: same reporter as #5933 re-reported repeating For You videos on app `1.0.16+804` / iOS 18.5 after the earlier recommender outage symptoms were gone.
- #5933: same reporter, app `1.0.15+778`, same user pubkey, repeating For You videos for days.
- #5822: separate Android reporter asked for videos they have not watched instead of repeated feed items.
- Current architecture review found `SeenVideosService` persists seen metrics under `seen_video_metrics` and exposes read APIs such as `getSeenVideoIds`, `wasSeenRecently`, and `getVideosNotSeenSince`, but no feed provider consumes those APIs for cross-session filtering or ordering.

## Desired Outcome

Recently watched videos should not immediately resurface in mobile feeds across app restarts when there are reasonable unseen alternatives. For You is the primary target; Following, Popular, and New should be evaluated explicitly rather than changed accidentally.

## Proposed Scope

- Wire existing `SeenVideosService` read APIs into the feed selection/filtering layer.
- Start with For You dedup/deprioritization, where the user-visible complaint is strongest.
- Avoid empty-feed traps for users with very small candidate pools: if filtering removes too much, fall back deliberately rather than showing nothing.
- Preserve pagination/session stability. Filtering should not cause duplicates between pages, cursor instability, or scroll jumps.
- Keep Nostr event IDs full length in logs/tests; let UI handle overflow.
- Coordinate with #5027 if/when the recommendations API accepts a seen/exclude list so the client can pass seen IDs server-side instead of only post-filtering.

## Acceptance Criteria

- For You feed load has a test-covered path that suppresses or deprioritizes recently seen videos using persisted watch history.
- Cross-restart seen-video state affects the next feed load when suitable unseen videos exist.
- Small-candidate fallback is explicit and test-covered.
- Pagination still returns stable, deduplicated pages without repeating items already emitted in the same feed session.
- Existing pull-to-refresh and resume-refresh behavior remains deliberate and covered by relevant tests.

## Related

- Backend/session freshness: #5027
- Resume refresh UX: #5134
- Duplicate/support signals: #6136, #5933, #5822

NotThatKindOfDrLiz:
## Summary
- add a Flutter-free seen-video lookup port and For You freshness ordering helper in videos_repository
- wire the app's persisted SeenVideosService into the home For You repository path and the legacy Explore For You provider
- consolidate SeenVideosNotifier onto the shared keep-alive service instance
- await seen-video store readiness before ordering and page deeper when the current For You page is entirely recently seen
- apply freshness ordering consistently to popular fallbacks and keep cache reads from rebuilding unnecessarily

Closes #6209

## Tests
- flutter test test/src/videos_repository_test.dart --name getRecommendedVideos (from mobile/packages/videos_repository)
- flutter test test/src/videos_repository_test.dart (from mobile/packages/videos_repository)
- very_good test --coverage --min-coverage 100 (from mobile/packages/videos_repository)
- flutter test test/providers/explore_feed_refresh_retention_test.dart (from mobile)
- flutter test test/services/seen_videos_service_test.dart (from mobile)
- flutter analyze (from mobile)
- pre-push hook: changed-file analysis, generated-file verification, videos_repository_test.dart, seen_videos_notifier_test.dart, seen_videos_service_test.dart

### DIFF
--- a/mobile/packages/videos_repository/lib/src/videos_repository.dart
+++ b/mobile/packages/videos_repository/lib/src/videos_repository.dart
@@ -2513,7 +2513,7 @@ class VideosRepository {
         _funnelcakeApiClient == null ||
         !_funnelcakeApiClient.isAvailable) {
       return HomeFeedResult(
-        videos: prioritizeNotRecentlySeenVideos(
+        videos: await _orderForYouFreshness(
           await getPopularVideos(
             limit: limit,
             until: until,
@@ -2521,7 +2521,6 @@ class VideosRepository {
             preferredLanguages: preferredLanguages,
             viewerCountry: viewerCountry,
           ),
-          seenVideoLookup: _seenVideoLookup,
         ),
       );
     }
@@ -2529,26 +2528,18 @@ class VideosRepository {
     final recommendationCursor = cursor ?? until?.toString();
     late final RecommendationsResponse response;
     try {
-      response = recommendationCursor == null
-          ? await _funnelcakeApiClient.getRecommendations(
-              pubkey: effectiveUserPubkey,
-              limit: limit,
-              seed: requestSeed,
-              preferredLanguages: preferredLanguages,
-              viewerCountry: viewerCountry,
-            )
-          : await _funnelcakeApiClient.getRecommendations(
-              pubkey: effectiveUserPubkey,
-              limit: limit,
-              cursor: recommendationCursor,
-              seed: requestSeed,
-              preferredLanguages: preferredLanguages,
-              viewerCountry: viewerCountry,
-            );
+      response = await _fetchRecommendationsPage(
+        pubkey: effectiveUserPubkey,
+        limit: limit,
+        cursor: recommendationCursor,
+        seed: requestSeed,
+        preferredLanguages: preferredLanguages,
+        viewerCountry: viewerCountry,
+      );
     } on FunnelcakeException {
       if (recommendationCursor != null) rethrow;
       return HomeFeedResult(
-        videos: prioritizeNotRecentlySeenVideos(
+        videos: await _orderForYouFreshness(
           await getPopularVideos(
             limit: limit,
             until: until,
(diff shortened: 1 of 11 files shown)


mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -2595,6 +2601,23 @@ class VideosRepository {
     return result;
   }
 
+  HomeFeedResult _withForYouFreshnessOrdering(HomeFeedResult result) {
+    final orderedVideos = prioritizeNotRecentlySeenVideos(
+      result.videos,
+      seenVideoLookup: _seenVideoLookup,
+    );
+    if (identical(orderedVideos, result.videos)) return result;
```
hm21, line 2704: Minor: this `identical` guard never fires — `prioritizeNotRecentlySeenVideos` always returns a fresh `videos.toList()`, so a new `HomeFeedResult` is rebuilt on every cache read. Drop it, or return the input unchanged from the helper when it doesn't repartition.

hm21 (changes requested):
Tested on-device: For You still repeats the same already-watched videos across restarts — the exact symptom from #6209, so this doesn't resolve it yet.

Why: it only reorders the single page the recommender returns (limit 25). When a user's top page is already fully watched (small/stable pool), `prioritizeNotRecentlySeenVideos` hits its all-seen fallback and returns the page unchanged — nothing visibly changes. The recommendations endpoint has no seen/exclude param, so the client can't request fresh content.

Suggestions:
1. Cold-start gap: `SeenVideosService.initialize()` runs in the `standard` startup phase (after `runApp`), while the feed loads on foreground — so the first feed after a restart can order before the seen store has loaded from disk and skip deprioritization entirely. Await seen-store readiness before ordering.
2. Surface unseen that exist deeper: when the first page is (nearly) all seen, keep paginating the cursor until unseen candidates appear and prefer them (keeping the empty-feed fallback). Without this, users with stable pools see no change — the reported symptom.
3. The complete fix needs backend #5027 (seen/exclude list); client-side reordering alone can't surface fresh content.

NotThatKindOfDrLiz:
Takeover update pushed in b788f4f34.

What changed:
- `SeenVideoLookup` now has an awaitable readiness hook, and `SeenVideosService.initialize()` shares an in-flight initialization so cold-start reads do not race the persisted store.
- Repository For You ordering now waits for the seen store, applies ordering to popular fallbacks consistently, preserves the no-op cache identity guard, and fetches deeper recommendation pages when the current page is entirely recently seen.
- Legacy Explore For You also awaits the shared seen service before ordering.
- Added regression coverage for pre-init ordering, all-seen first-page pagination, and empty-recommendations popular fallback ordering.

Verification:
- `flutter test test/src/videos_repository_test.dart --name getRecommendedVideos` from `mobile/packages/videos_repository`
- `flutter test test/src/videos_repository_test.dart` from `mobile/packages/videos_repository`
- `flutter test test/providers/explore_feed_refresh_retention_test.dart` from `mobile`
- `flutter test test/services/seen_videos_service_test.dart` from `mobile`
- pre-push hook passed changed-file analysis, generated-file verification, `videos_repository_test.dart`, `seen_videos_notifier_test.dart`, and `seen_videos_service_test.dart`

NotThatKindOfDrLiz:
Follow-up: added cached freshness reordering coverage in 20d26d492 after Videos Repository CI flagged the copied cached-result branch. Local package coverage now passes with `very_good test --coverage --min-coverage 100`, and all GitHub checks are green on the latest commit.

NotThatKindOfDrLiz:
@hm21 Can you please rereview? TYSM!

hm21 (approved):
LGTM

One forward-looking note (non-blocking): this only reorders within an already-fetched page, so a viewer can still hit repeats after scrolling past the few fresh items, and across sessions the finite candidate pool + 24h window means seen videos keep resurfacing. The durable fix is server-side — pass the seen-video set to the Funnelcake recommender so it excludes them at the source. Worth a follow-up issue rather than expanding this PR.

PR #6213, opened 2026-07-20
feat(dm): save shared videos from messages

### LINKED ISSUE #6214: feat(dm): save videos from shared message cards
## Problem

Shared-video cards in direct messages can open the reel and copy text, but the message action menu does not let the recipient save the video. Structured NIP-18 shares also render correctly while their long-press menu fails to recognize them as video shares.

## Acceptance criteria

- Shared-video DM cards expose a **Save Video** long-press action.
- Legacy Divine URLs and structured NIP-18 video references behave consistently.
- The current user's own video saves without a watermark.
- Another creator's video uses the existing watermarked save flow.
- Missing videos show unavailable feedback.
- Text-only messages and compact quoted-video replies do not expose the action.
- Behavior is covered by widget and unit tests.

rabble:
Closes #6214

## Summary

- add **Save Video** to shared-video DM card long-press actions
- resolve shared-video saves through a guarded cubit with pending/unavailable feedback and a11y announcements
- normalize legacy Divine URLs and structured NIP-18 video references so cards, copy-link, and saving resolve the same video
- save the current user's own videos as originals and other creators' videos with the existing Divine watermark flow
- reject bare non-video `q` references so ordinary quoted notes do not expose video actions

## Test plan

- `flutter test test/blocs/dm/shared_video_save/shared_video_save_cubit_test.dart test/screens/inbox/conversation/dm_video_target_test.dart test/screens/inbox/conversation/conversation_view_test.dart test/screens/inbox/conversation/widgets/reaction_picker_overlay_test.dart test/utils/divine_video_url_test.dart packages/dm_repository/test/src/dm_shared_video_citation_test.dart`
- `flutter analyze lib test`

### DIFF
--- a/mobile/lib/screens/inbox/conversation/conversation_view.dart
+++ b/mobile/lib/screens/inbox/conversation/conversation_view.dart
@@ -11,6 +11,7 @@ import 'package:go_router/go_router.dart';
 import 'package:models/models.dart';
 import 'package:openvine/blocs/dm/conversation/conversation_bloc.dart';
 import 'package:openvine/blocs/dm/reactions/conversation_reactions_cubit.dart';
+import 'package:openvine/blocs/dm/shared_video_save/shared_video_save_cubit.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/l10n/localized_time_formatter.dart';
 import 'package:openvine/providers/app_providers.dart';
@@ -23,7 +24,7 @@ import 'package:openvine/services/collaborator_invite_parser.dart';
 import 'package:openvine/services/collaborator_invite_service.dart';
 import 'package:openvine/utils/clipboard_utils.dart';
 import 'package:openvine/utils/nostr_key_utils.dart';
-import 'package:openvine/utils/watermark_text_resolver.dart';
+import 'package:openvine/utils/string_utils.dart';
 import 'package:openvine/widgets/profile/more_sheet/more_sheet_content.dart';
 import 'package:openvine/widgets/profile/more_sheet/more_sheet_result.dart';
 import 'package:openvine/widgets/profile/profile_header_widget.dart'
@@ -54,54 +55,15 @@ class ConversationView extends ConsumerStatefulWidget {
 enum _FailedMessageAction { resend, delete }
 
 class _ConversationViewState extends ConsumerState<ConversationView> {
-  Future<void> _saveSharedVideo(DmVideoTarget target) async {
-    try {
-      final video = await ref
-          .read(videosRepositoryProvider)
-          .fetchVideoWithStatsForRouteId(
-            target.stableId,
-            fallbackRouteIds: target.fallbackRouteIds,
-          );
-      if (!mounted) return;
-      if (video == null) {
-        _showVideoUnavailable();
-        return;
-      }
-
-      final currentPubkey = ref.read(authServiceProvider).currentPublicKeyHex;
-      if (video.pubkey == currentPubkey) {
-        await showSaveOriginalSheet(context: context, ref: ref, video: video);
-        return;
-      }
-
-      final profile = await ref
-          .read(profileRepositoryProvider)
-          ?.getCachedProfile(pubkey: video.pubkey);
-      if (!mounted) return;
-      final watermarkText = resolveWatermarkText(
-        profile: profile,
-        fallbackAuthorName: video.authorName,
-      );
-      await showWatermarkDownloadSheet(
-        context: context,
-        ref: ref,
-        video: video,
-        watermarkText: watermarkText,
-      );
-    } on Exception {
-      if (mounted) _showVideoUnavailable();
-    }
-  }
-
-  void _showVideoUnavailable() {
+  void _showSnackbar(String message, {required bool error}) {
     ScaffoldMessenger.of(context)
       ..hideCurrentSnackBar()
-      ..showSnackBar(
-        DivineSnackbarContainer.snackBar(
-          context.l10n.notificationsVideoUnavailable,
-          error: true,
-        ),
-      );
+      ..showSnackBar(DivineSnackbarContainer.snackBar(message, error: error));
+    SemanticsService.sendAnnouncement(
+      View.of(context),
+      message,
+      Directionality.of(context),
+    );
   }
 
   Future<void> _onOptions(String otherPubkey, String displayName) async {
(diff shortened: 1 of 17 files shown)


mobile/test/screens/inbox/conversation/conversation_view_test.dart
```diff
@@ -1554,6 +1566,177 @@ void main() {
       );
     });
 
+    group('shared video saving', () {
```
realmeylisdev, line 1573: Two acceptance criteria — *replies don't expose the action* and *text-only messages don't expose the action* — have no test that can fail here. The only `findsNothing` checks live in `reaction_picker_overlay_test.dart`, which exercises the overlay's own `if (isVideoShare)` switch, not the code that computes `isVideoShare`.

I dropped the reply guard (`resolveOwnShareVideoRef` → `message.sharedVideoRef`) and separately flipped `isVideoShare` to `true`; the whole conversation suite stayed green both times. The existing text long-press test (L1519) only asserts Copy/Report present + Delete absent, so extra video entries don't fail it.

Worth adding here: (a) a reply carrying a `sharedVideoRef` → Save/Copy absent, (b) a plain text message → Save/Copy absent, (c) a legacy plain `divine.video/…` URL message → Save succeeds (the fixture only builds structured refs today).

realmeylisdev (commented):
Formal review — nicely scoped feature that hits the acceptance criteria, the `DmVideoTarget` refactor de-duplicates the bubble/menu resolution cleanly, and the own-vs-watermark routing matches the feed precedent. Verified locally: `flutter analyze` clean, all 123 conversation tests green, and I confirmed bare-d-tag / event-id route resolution against the local funnelcake stack.

Leaving this as **Comment** (not blocking). A few things worth a look before merge: two acceptance criteria (replies / text-only don't expose Save) ship with no test that can fail; the *Video unavailable* snackbar misses the a11y announcement this file uses everywhere else; and the resolve has no in-flight guard, so a slow-network double-tap can save the same video twice. The rest are nits. Details inline.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -51,6 +54,56 @@ class ConversationView extends ConsumerStatefulWidget {
...
+        ref: ref,
+        video: video,
+        watermarkText: watermarkText,
+      );
+    } on Exception {
+      if (mounted) _showVideoUnavailable();
+    }
+  }
+
+  void _showVideoUnavailable() {
```
realmeylisdev: Every other outcome snackbar in this file pairs with `SemanticsService.sendAnnouncement` — blocked (L306), failed (L318), partial (L361) — and the rule is spelled out at L358-360. This new error snackbar skips it, so VoiceOver/TalkBack users get no signal that the save failed. Please announce here too, with `Directionality.of(context)`.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -51,6 +54,56 @@ class ConversationView extends ConsumerStatefulWidget {
 enum _FailedMessageAction { resend, delete }
 
 class _ConversationViewState extends ConsumerState<ConversationView> {
+  Future<void> _saveSharedVideo(DmVideoTarget target) async {
```
realmeylisdev: The resolve runs *after* the action sheet has already popped, and it's unbounded (Funnelcake 15s + relay 3s×2 + stats 3s, and again for the `fallbackRouteIds` entry a structured share always supplies). Nothing is on screen while it runs, and there's no in-flight guard — a second long-press → Save starts a second `_saveSharedVideo`, so two progress sheets stack and the same video lands in the camera roll twice.

Both other callers of this method render a pending state first (`video_detail_screen` sets `_isLoading`; `VideoLinkPreviewCubit` emits a loading state). Suggest showing progress on tap and guarding re-entry.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -468,7 +530,7 @@ class _MessageList extends StatelessWidget {
     final result = await ReactionPickerOverlay.show(
       context: context,
       isSent: isSent,
-      isVideoShare: videoUrl != null,
+      isVideoShare: videoTarget != null,
```
realmeylisdev, line 536: This gate widened from *a divine.video URL is in the text* to *any parsed `q`-ref exists*. `DmSharedVideoCitation.parse` classifies any 64-hex `q` value as a kind-22 video, so a cross-client DM that quotes a plain note (`["q","<64-hex note id>",…]`) now offers **Copy video URL** (copies `divine.video/video/<note-id>`, which resolves to nothing) and **Save Video** (always ends in *Video unavailable*).

The bogus video card already rendered before this PR — the root cause is the permissive parser in `dm_shared_video_citation.dart` — but this PR extends the misclassification into the copy/save surface. Either tighten recognition here or confirm it's acceptable given the pre-existing card.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -51,6 +54,56 @@ class ConversationView extends ConsumerStatefulWidget {
...
+        profile: profile,
+        fallbackAuthorName: video.authorName,
+      );
+      await showWatermarkDownloadSheet(
+        context: context,
+        ref: ref,
+        video: video,
+        watermarkText: watermarkText,
+      );
+    } on Exception {
```
realmeylisdev: Two small things on this catch. It's `on Exception`, so `Error` subtypes slip past — a Drift closed-DB `StateError` from the cache read would escape as an unhandled async error with no feedback (mocktail's unstubbed call actually throws a `TypeError`, an `Error`, which this wouldn't catch). And it swallows with no `Log`, so a genuine transport/store failure is indistinguishable from a deleted video in support/telemetry.

Network errors correctly shouldn't hit Crashlytics per the rules, but a `Log.warning` here keeps it triageable.

mobile/lib/screens/inbox/conversation/widgets/message_actions_sheet.dart
```diff
@@ -52,6 +54,12 @@ class MessageActionsSheet {
           label: l10n.dmMessageActionCopyVideoUrl,
           onTap: () => result = MessageAction.copyVideoUrl,
         ),
+      if (isVideoShare)
+        VineBottomSheetActionData(
+          iconPath: DivineIconName.downloadSimple.assetPath,
```
realmeylisdev: Heads up — `MessageActionsSheet.show` has no call site anywhere in `lib/` (it lost its caller before this PR; every long-press goes through `ReactionPickerOverlay`), so this new tile never renders. It's only kept alive by a test *description* string. Not introduced here, but this PR extends dead code and the two menus can now silently drift. Either drop `MessageActionsSheet` or leave it untouched.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -452,14 +510,18 @@ class _MessageList extends StatelessWidget {
...
+  final Future<void> Function(DmVideoTarget target) onSaveVideo;
 
   Future<void> _onMessageLongPress(
     BuildContext context,
     DmMessage message,
     bool isSent,
     DmDeliveryStatus deliveryStatus,
   ) async {
-    final videoUrl = tryExtractDivineVideoUrl(message.content);
+    final videoTarget = resolveDmVideoTarget(
```
realmeylisdev, line 524: This was the last production caller of `tryExtractDivineVideoUrl` (`lib/utils/divine_video_url.dart`). After the swap to `resolveDmVideoTarget`, the helper is orphaned — only its own test group references it now, so per `testing.md` that test guards nothing. Suggest removing the helper + its test group in this PR.

mobile/lib/screens/inbox/conversation/conversation_view.dart
```diff
@@ -452,14 +510,18 @@ class _MessageList extends StatelessWidget {
...
+      content: message.content,
```
realmeylisdev: Minor: this resolves against raw `message.content`, but the bubble resolves against `StringUtils.sanitizeUtf16(message)`. A sender-crafted lone surrogate inside the URL makes the two disagree — the card the user sees resolves one id, Save fetches another. Extreme edge case, but resolving both from the same sanitized string closes it.

mobile/test/screens/inbox/conversation/conversation_view_test.dart
```diff
@@ -1554,6 +1566,177 @@ void main() {
...
+          });
+
+          await openSaveAction(tester, message: message);
+          await tester.tap(find.text(l10n.shareSheetSaveVideo));
+          await tester.pumpAndSettle();
+
+          verify(
+            () => mockWatermarkDownloadService.downloadWithWatermark(
+              video: video,
+              watermarkText: any(named: 'watermarkText'),
```
realmeylisdev, line 1783: The watermark tests match `watermarkText: any(named: 'watermarkText')`, and the fixture forces `profileRepositoryProvider` to null so `resolveWatermarkText` always returns the `@Divine` fallback — the actual value is never asserted and the NIP-05/name branch is never exercised, so a wrong watermark would pass. Worth pinning the resolved text and adding a case with a non-null profile.

mobile/test/screens/inbox/conversation/widgets/reaction_picker_overlay_test.dart
```diff
@@ -57,9 +58,51 @@ void main() {
...
       expect(find.text('Delete for everyone'), findsOneWidget);
     });
 
+    testWidgets('shows copy and save actions for a shared video', (
+      tester,
+    ) async {
+      await openOverlay(tester, isVideoShare: true);
+
+      expect(find.text('Copy video URL'), findsOneWidget);
+      expect(find.text('Save Video'), findsOneWidget);
```
realmeylisdev: Nit per `localization.md`: these assert the hardcoded English `'Save Video'` / `'Copy video URL'` instead of resolving the keys from `AppLocalizations`. It matches the file's existing style so it's low priority, but resolving via `lookupAppLocalizations`/l10n keeps the test from silently surviving a copy change.

mobile/lib/screens/inbox/conversation/dm_video_target.dart
```diff
@@ -0,0 +1,80 @@
+import 'package:flutter/foundation.dart';
```
realmeylisdev, line 4: Tiny: the neighbouring files in this folder (`divine_video_url.dart`, `dm_shared_video_citation.dart`) open with an `// ABOUTME:` header; this new file and its test skip it. Add for consistency.

NotThatKindOfDrLiz:
Took over the review follow-up and pushed `5a416989d`.

Changes:
- moved shared-video save resolution into `SharedVideoSaveCubit` with a resolving state and in-flight guard
- added pending/unavailable snackbars with `SemanticsService.sendAnnouncement`
- logged resolve/profile lookup failures without reporting transient save-resolution failures as crashes
- removed dead `MessageActionsSheet` UI and the orphaned `tryExtractDivineVideoUrl` helper while keeping regex coverage
- tightened DM q-tag parsing so bare 64-hex note references are not treated as saveable videos
- added regression coverage for legacy URL save, reply/text exclusion, duplicate saves, unavailable a11y, and watermark text resolution

Verification:
- `flutter test test/blocs/dm/shared_video_save/shared_video_save_cubit_test.dart test/screens/inbox/conversation/dm_video_target_test.dart test/screens/inbox/conversation/conversation_view_test.dart test/screens/inbox/conversation/widgets/reaction_picker_overlay_test.dart test/utils/divine_video_url_test.dart packages/dm_repository/test/src/dm_shared_video_citation_test.dart`
- `flutter analyze lib test`

mobile/lib/screens/inbox/conversation/conversation_view.dart
realmeylisdev: `5a416989d` tightened the parser to require a hex64 author in `q` tag[3], which closes the bare `["q","<64-hex>"]` (and non-hex-author) case. It doesn't fully close the concern, though: a spec-compliant NIP-18 quote of a plain note — `["q","<id>","<relay>","<pubkey>"]` — is byte-identical to our own kind-22 `nevent` video share, so it still surfaces Copy / Save and ends in *Video unavailable*. A regular-event `q` reference carries no kind, so tag shape alone can't separate a note quote from a video quote; the only complete fix is to gate Copy / Save on a resolved kind.

Leaving this open to decide: accept the graceful `unavailable` fallback (matches the pre-existing bogus card), or gate the actions on the resolved kind. Note `dm_shared_video_citation_test.dart:128-138` currently asserts the 4-element note-quote shape parses as a video, so if we tighten, that expectation moves with it.

mobile/test/screens/inbox/conversation/widgets/reaction_picker_overlay_test.dart
realmeylisdev: Fixed in `15a4a6e0d` — the Save / Copy-video assertions now resolve `shareSheetSaveVideo` / `dmMessageActionCopyVideoUrl` via `lookupAppLocalizations(const Locale('en'))` instead of the hardcoded English, so the `findsNothing` negatives can no longer pass vacuously after an ARB copy change.

mobile/lib/screens/inbox/conversation/dm_video_target.dart
realmeylisdev: Fixed in `15a4a6e0d` — added the `// ABOUTME:` header to `dm_video_target_test.dart` so it matches the sibling files and the production `dm_video_target.dart`.

mobile/lib/screens/inbox/conversation/conversation_view.dart
realmeylisdev: Accepting the graceful fallback as a known limitation. A regular-event `q` reference carries no kind, so a NIP-18 note quote is indistinguishable from a video quote by tag shape alone; the mis-detected card degrades to *Video unavailable* / a dead copied URL — nothing harmful — and `5a416989d` already narrowed the false-positive surface. Good to resolve; we can revisit a kind-gate on Copy / Save later if it proves annoying in practice.

realmeylisdev (approved):
Re-verified the follow-up against all 11 review threads — 8 fixed, T10/T11 handled in `15a4a6e0d`, T4 accepted as a non-harmful known limitation. CI green, and moving resolution into `SharedVideoSaveCubit` is a clean improvement. LGTM.

NotThatKindOfDrLiz (approved):
Approved after takeover fixes in `5a416989d`.

I rechecked the original blockers against the updated diff: save resolution now lives behind `SharedVideoSaveCubit` with a pending state and duplicate-submit guard; unavailable feedback is announced for accessibility; failures are logged without crash-reporting expected transient resolution errors; the dead action sheet UI and orphaned URL helper are removed; q-ref parsing no longer treats bare note refs as videos; and the missing legacy/reply/text/duplicate-save/a11y coverage is in place.

Local verification passed:
- `flutter test test/blocs/dm/shared_video_save/shared_video_save_cubit_test.dart test/screens/inbox/conversation/dm_video_target_test.dart test/screens/inbox/conversation/conversation_view_test.dart test/screens/inbox/conversation/widgets/reaction_picker_overlay_test.dart test/utils/divine_video_url_test.dart packages/dm_repository/test/src/dm_shared_video_citation_test.dart`
- `flutter analyze lib test`

GitHub checks are green on `5a416989d`.

PR #6218, opened 2026-07-20
feat(perf): trace video generation and camera startup time

### LINKED ISSUE #6219: feat(perf): trace video generation and camera startup time
Auto-created from PR #6218.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6218
**Branch:** `perf/video-pipeline-traces`

hm21:
Closes #6219

## Description

Adds two Firebase Performance **custom traces** to the video pipeline so we can measure, in production, how long two user-facing operations actually take. Both reuse the existing `PerformanceMonitoringService` / `PerformanceTraceMonitor` abstraction (the same one behind the `feed_load` and `video_upload` traces) — no new perf infrastructure.

**1. `video_generation`** — wraps the final export in `VideoEditorNotifier._runRenderVideo`, measuring the whole span the user waits for after "Done" (native encode + ProofMode + C2PA).

**2. `camera_startup`** — wraps the native camera bring-up (`_cameraService.initialize`) in `VideoRecorderBloc`, measuring how long that native `initialize()` call takes. There was no production timing for this before — the only camera benchmark (`camera_initialization_benchmark_test.dart`) is `skip: true` and mock-based, so it never measured real hardware.

Each trace carries attributes so the console can slice/filter:

| Trace | Attributes |
|---|---|
| `video_generation` | `outcome` (success / failed / error / incomplete), `clip_count`, `aspect_ratio` |
| `camera_startup` | `outcome` (success / failed / error), `lens`, `quality` |

Filter on `outcome=success` for a clean duration distribution.

### Why these design choices (non-obvious bits)

- **Both `startTrace` and `stopTrace` are fire-and-forget; the tagging + stop live in `finally`.** `startTrace` isn't awaited so render / camera-init dispatch stays synchronous — awaiting it would run `trace.start()`'s platform round-trip first (the existing overlapping-render test relies on that dispatch being synchronous). `stopTrace` isn't awaited either, so `trace.stop()`'s round-trip doesn't gate camera preview bring-up or render completion — the exact startup path `camera_startup` measures. The `finally` still runs on every path, including a user cancel (which bumps `_renderGeneration` without starting a new render), so the trace is always stopped and tagged.
- **Single-named trace + overlap ownership.** The service keys traces by name, so two overlapping `_runRenderVideo` calls would otherwise share the one `video_generation` trace and corrupt each other. A per-render `_videoGenerationTraceOwner` gates the tag/stop calls: once a newer render takes the trace over, the superseded render skips its trace ops so it can't mis-tag or prematurely stop the newer render's trace. A cancel-superseded render still owns its trace and stops it tagged `outcome=incomplete`, so it's excluded from the duration distribution rather than skewing it.
- **Camera trace via constructor injection.** `VideoRecorderBloc` is pure Dart (no `ref`), so a `PerformanceTraceMonitor` is injected with a `NoOpPerformanceTraceMonitor()` default (existing tests need no change) and wired to the real Firebase-backed monitor at the recorder screen's `BlocProvider`. Same pattern as `VideoEventService`.

**Related Issue:** 

## Out of Scope

- **Fleet-wide per-encode timing.** Every render type (reverse, transform, merge, crop, transition-seam, speed, watermark) funnels through the single static `VideoEditorRenderService.renderNativeVideoToFile`. Tracing there would cover them all with one hook, but the class is static and would need a small injection seam + test-safety work. Deferred; this PR covers the two user-facing operations only.
- **Sub-`initialize()` camera readiness.** `camera_startup` brackets the native `initialize()` call. Post-init steps (`_emitCameraSync`, stabilization-preference restore, remote-record wiring) run after the trace stops and are not measured.
- No dedicated trace-assertion unit tests — the provider injects the concrete monitor (a no-op in tests), and the existing render-outcome / camera-init path tests already exercise every branch the tracing sits on.

## Verification

- `dart format` — clean
- `flutter analyze lib` — No issues found
- `flutter test test/providers/video_editor_provider_test.dart` — 101/101
- `flutter test test/blocs/video_recorder/video_recorder_bloc_test.dart` — 83/83
- `flutter test test/screens/video_recorder_screen_test.dart` — 22/22
- `flutter test test/providers/video_recorder_audio_service_factory_test.dart` — green
- Pre-push hook (analyze + changed-file tests) — green

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -30,6 +30,7 @@ import 'package:openvine/models/video_recorder/video_recorder_timer_duration.dar
 import 'package:openvine/providers/clip_manager_provider.dart';
 import 'package:openvine/providers/video_editor_provider.dart';
 import 'package:openvine/services/haptic_service.dart';
+import 'package:openvine/services/performance_monitoring_service.dart';
 import 'package:openvine/services/video_recorder/camera/camera_base_service.dart';
 import 'package:openvine/services/video_thumbnail_service.dart';
 import 'package:pro_video_editor/pro_video_editor.dart';
@@ -146,6 +147,7 @@ class VideoRecorderBloc
     CameraService? cameraService,
     CountdownSoundServiceFactory? countdownSoundServiceFactory,
     AudioPlaybackServiceFactory? audioPlaybackServiceFactory,
+    PerformanceTraceMonitor? performanceMonitor,
   }) : _readClipManager = readClipManager,
        _readVideoEditor = readVideoEditor,
        _readVideoEditorState = readVideoEditorState,
@@ -155,6 +157,8 @@ class VideoRecorderBloc
            countdownSoundServiceFactory ?? defaultCountdownSoundServiceFactory,
        _audioPlaybackServiceFactory =
            audioPlaybackServiceFactory ?? defaultAudioPlaybackServiceFactory,
+       _performanceMonitor =
+           performanceMonitor ?? const NoOpPerformanceTraceMonitor(),
        super(const VideoRecorderBlocState()) {
     _cameraService =
         _cameraServiceOverride ??
@@ -241,6 +245,7 @@ class VideoRecorderBloc
   final CameraService? _cameraServiceOverride;
   final CountdownSoundServiceFactory _countdownSoundServiceFactory;
   final AudioPlaybackServiceFactory _audioPlaybackServiceFactory;
+  final PerformanceTraceMonitor _performanceMonitor;
 
   late final CameraService _cameraService;
   AudioPlaybackService? _audioPlaybackService;
@@ -304,6 +309,13 @@ class VideoRecorderBloc
       category: LogCategory.video,
     );
 
+    const traceName = 'camera_startup';
+    // Fire-and-forget so the native init dispatch is not delayed a microtask;
+    // the trace still stops in `finally`, tagged with the outcome so the
+    // console can filter startups that never reached a live preview.
+    unawaited(_performanceMonitor.startTrace(traceName));
+
+    Object? initError;
     try {
       await _cameraService.initialize(
         videoQuality: event.videoQuality,
@@ -311,14 +323,31 @@ class VideoRecorderBloc
         enableAutoLensSwitch: true,
       );
     } catch (e) {
+      initError = e;
+    } finally {
+      _performanceMonitor
+        ..putAttribute(traceName, 'lens', initialLens.name)
+        ..putAttribute(traceName, 'quality', event.videoQuality.value)
+        ..putAttribute(
+          traceName,
+          'outcome',
+          initError != null
+              ? 'error'
+              : (_cameraService.isInitialized ? 'success' : 'failed'),
+        );
+      await _performanceMonitor.stopTrace(traceName);
+    }
+
+    if (initError != null) {
       Log.error(
-        '📹 Camera service initialization threw exception: $e',
+        '📹 Camera service initialization threw exception: $initError',
         name: 'VideoRecorderBloc',
         category: LogCategory.video,
       );
       emit(
         state.copyWith(
-          initializationErrorMessage: 'Camera initialization failed: $e',
+          initializationErrorMessage:
+              'Camera initialization failed: $initError',
         ),
       );
       return;
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -310,9 +310,12 @@ class VideoRecorderBloc
     );
 
     const traceName = 'camera_startup';
-    // Fire-and-forget so the native init dispatch is not delayed a microtask;
-    // the trace still stops in `finally`, tagged with the outcome so the
-    // console can filter startups that never reached a live preview.
+    // Start and stop the trace fire-and-forget: an awaited start would run
+    // trace.start()'s platform round-trip before dispatching the native camera
+    // init, and an awaited stop would gate preview bring-up on trace.stop() —
+    // both add latency to the exact startup path this trace measures. The trace
+    // still stops in `finally`, tagged with the outcome so the console can
+    // filter startups that never reached a live preview.
     unawaited(_performanceMonitor.startTrace(traceName));
 
     Object? initError;
@@ -335,7 +338,7 @@ class VideoRecorderBloc
               ? 'error'
               : (_cameraService.isInitialized ? 'success' : 'failed'),
         );
-      await _performanceMonitor.stopTrace(traceName);
+      unawaited(_performanceMonitor.stopTrace(traceName));
     }
 
     if (initError != null) {
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -309,14 +309,14 @@ class VideoRecorderBloc
       category: LogCategory.video,
     );
 
-    const traceName = 'camera_startup';
-    // Start and stop the trace fire-and-forget: an awaited start would run
-    // trace.start()'s platform round-trip before dispatching the native camera
-    // init, and an awaited stop would gate preview bring-up on trace.stop() —
-    // both add latency to the exact startup path this trace measures. The trace
-    // still stops in `finally`, tagged with the outcome so the console can
-    // filter startups that never reached a live preview.
-    unawaited(_performanceMonitor.startTrace(traceName));
+    // Operation-scoped trace: this init owns the returned handle, so a
+    // concurrent init gets its own trace instead of stopping or re-attributing
+    // this one. Start is fire-and-forget so it doesn't add a platform round-trip
+    // to the exact startup path this trace measures; stop is fire-and-forget so
+    // it doesn't gate preview bring-up. The handle is still stopped in `finally`,
+    // tagged with the outcome so the console can filter startups that never
+    // reached a live preview.
+    final trace = _performanceMonitor.startOperationTrace('camera_startup');
 
     Object? initError;
     try {
@@ -328,17 +328,16 @@ class VideoRecorderBloc
     } catch (e) {
       initError = e;
     } finally {
-      _performanceMonitor
-        ..putAttribute(traceName, 'lens', initialLens.name)
-        ..putAttribute(traceName, 'quality', event.videoQuality.value)
+      trace
+        ..putAttribute('lens', initialLens.name)
+        ..putAttribute('quality', event.videoQuality.value)
         ..putAttribute(
-          traceName,
           'outcome',
           initError != null
               ? 'error'
               : (_cameraService.isInitialized ? 'success' : 'failed'),
         );
-      unawaited(_performanceMonitor.stopTrace(traceName));
+      unawaited(trace.stop());
     }
 
     if (initError != null) {
--- a/mobile/lib/providers/video_editor_provider.dart
+++ b/mobile/lib/providers/video_editor_provider.dart
@@ -138,6 +138,13 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
   int _renderGeneration = 0;
   Future<void>? _activeRenderFuture;
 
+  /// Generation that currently owns the single-named `video_generation`
+  /// performance trace. Bumped only by [_runRenderVideo] (never by cancel or
+  /// re-sign), so a render superseded by an overlapping render sees this change
+  /// and leaves the newer render's trace alone, while a render superseded by a
+  /// cancel still owns — and stops — its own trace.
+  int _videoGenerationTraceOwner = 0;
+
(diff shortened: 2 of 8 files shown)


mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
```diff
@@ -304,21 +309,48 @@ class VideoRecorderBloc
...
     );
 
+    const traceName = 'camera_startup';
+    // Start and stop the trace fire-and-forget: an awaited start would run
+    // trace.start()'s platform round-trip before dispatching the native camera
+    // init, and an awaited stop would gate preview bring-up on trace.stop() —
+    // both add latency to the exact startup path this trace measures. The trace
+    // still stops in `finally`, tagged with the outcome so the console can
+    // filter startups that never reached a live preview.
+    unawaited(_performanceMonitor.startTrace(traceName));
```
realmeylisdev: `camera_startup` copies the `video_generation` trace shape but drops the `_videoGenerationTraceOwner` guard — and `on<VideoRecorderInitializeRequested>` (line 184) runs on the default `concurrent` transformer, unlike the record start/stop handlers that were given `sequential()` for camera hazards.

Init is dispatched from four sites (initState, the upload→record mode-switch listener, and two nav-return sites). If two land inside a cold-start `initialize()` (hundreds of ms – seconds), the single name-keyed trace cross-wires: init B's `startTrace` auto-stops+records init A's trace attribute-less, then A's `finally` tags B's trace with A's lens/quality/outcome and stops it early, and B's real startup goes unrecorded. So the metric this PR adds gets one spurious sample, one mis-attributed/truncated sample, and one lost sample per overlap.

No camera/UX effect — purely the Firebase distribution. Options: `sequential()`/`droppable()` on the init handler (also serialises concurrent camera init, arguably desirable on its own), an owner guard mirroring `video_generation`, or a note on why two inits can't overlap in practice.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
```diff
@@ -304,21 +309,48 @@ class VideoRecorderBloc
...
+        ..putAttribute(traceName, 'lens', initialLens.name)
+        ..putAttribute(traceName, 'quality', event.videoQuality.value)
+        ..putAttribute(
+          traceName,
+          'outcome',
+          initError != null
+              ? 'error'
+              : (_cameraService.isInitialized ? 'success' : 'failed'),
+        );
+      unawaited(_performanceMonitor.stopTrace(traceName));
```
realmeylisdev: Separate from the overlap point: `startTrace` only registers the trace in `_activeTraces` *after* `await trace.start()`'s platform round-trip (service line 87), but it's fire-and-forget here. On a fast-resolving `initialize()` — cached permission-denied / no-camera / hardware-busy, or a warm re-init — this `finally` can run before the trace is registered, so all three `putAttribute`s and this `stopTrace` silently no-op (`'trace not found'` warnings). The still-pending `start()` then registers a trace nothing stops; the next init auto-stops it with a duration spanning the whole gap → a large tail outlier, and the fast-failure sample (exactly what this trace is meant to catch) is dropped.

Cold start is unaffected (hardware init ≫ `trace.start`), so the damage concentrates in the p90/p99 tail — where perf investigation actually looks. If the fast-failure samples matter, awaiting `startTrace` before `initialize()` closes it, at the cost of the small dispatch delay the comment calls out. Otherwise worth a line acknowledging the tail bias.

mobile/lib/providers/video_editor_provider.dart
```diff
@@ -137,6 +138,13 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
   int _renderGeneration = 0;
   Future<void>? _activeRenderFuture;
 
+  /// Generation that currently owns the single-named `video_generation`
+  /// performance trace. Bumped only by [_runRenderVideo] (never by cancel or
+  /// re-sign), so a render superseded by an overlapping render sees this change
+  /// and leaves the newer render's trace alone, while a render superseded by a
+  /// cancel still owns — and stops — its own trace.
+  int _videoGenerationTraceOwner = 0;
```
realmeylisdev: The owner guard and the outcome tagging are the core new behaviour of this PR, but no test can currently catch a regression in them: the editor tests resolve `performanceMonitoringServiceProvider` to an uninitialised service (every trace call early-returns) and the recorder bloc defaults to `NoOpPerformanceTraceMonitor`. Deleting the `if (ownsTrace())` guard on the finally-stop, or swapping `tagOutcome('success')`/`'failed'`, leaves the whole suite green.

A fake `PerformanceTraceMonitor` that records `(name, attribute, value)` calls is cheap — the bloc already accepts one and the provider is overridable — and would let the superseded-render guard and the outcome mapping fail loudly. Given the repo's "a test must be able to fail" bar, I'd add at least the owner-guard assertion (a superseded render must not stop the newer render's trace).

mobile/lib/providers/video_editor_provider.dart
```diff
@@ -1199,6 +1247,16 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
...
       state = state.copyWith(isProcessing: false, renderFailed: true);
+    } finally {
+      // Skip when a later render took over the trace. Stop fire-and-forget so a
+      // trace.stop() platform round-trip doesn't gate the render's completion.
+      if (ownsTrace()) {
+        performance.putAttribute(traceName, 'clip_count', '$clipCount');
+        if (aspectRatio != null) {
+          performance.putAttribute(traceName, 'aspect_ratio', aspectRatio);
+        }
+        unawaited(performance.stopTrace(traceName));
```
realmeylisdev: Nit / for awareness, no action needed: the owner guard correctly stops a superseded render from mis-tagging or double-stopping the newer trace, but the single-name key still has a residual window it can't close. A spaced second render records the abandoned trace attribute-less; a synchronous double-dispatch (the `callCount == 2` pattern in the overlap tests) instead orphans it — started via the fire-and-forget `startTrace`, overwritten in the map, never stopped. Both are rare and filterable by the missing `outcome` attribute, so this is fine as-is — just flagging that the guard narrows, but doesn't eliminate, the name-collision noise.

realmeylisdev (commented):
Solid, well-scoped instrumentation — reuses the existing trace abstraction, `flutter analyze` on the 3 files is clean, and both suites you cite pass locally here too. Every finding below is telemetry-quality only (no user-facing or functional impact); none block. The one worth weighing before merge: `camera_startup` omits the overlap guard you gave `video_generation`, while its init handler runs on the default `concurrent` transformer. Details inline, plus a fire-and-forget registration race, a test gap on the new guard/outcome logic, and one nit.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
hm21: Real mechanism, but with a single camera page the four init sites are serialized by the UX flow — nav releases the camera before re-init, and the mode-switch listener can't fire mid-cold-start — so two inits don't overlap in practice. Any freak overlap yields an outcome-less sample the `outcome=success` filter already excludes, so I'm not adding a behavioral change (`sequential()`) for it. Leaving as-is.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
hm21: Orphaned/auto-stopped traces carry no `outcome`, so the documented `outcome=success` filter already excludes them from the p90/p99 — the tail bias doesn't materialize in analysis. And the lost fast-fail samples aren't what this trace is for (successful startup duration). Leaving as-is.

mobile/lib/providers/video_editor_provider.dart
hm21: Good call. Added a recording fake monitor with an owner-guard test (a superseded render must not stop or re-tag the winning trace) plus outcome mapping for both traces, in c886dba. Verified by mutation that dropping the `ownsTrace()` guard turns it red. Flagged in-test that this is telemetry-only coverage — a touch belt-and-suspenders.

mobile/lib/providers/video_editor_provider.dart
```diff
@@ -1129,6 +1137,34 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
...
+    // render's trace.
+    _videoGenerationTraceOwner = generation;
+    bool ownsTrace() => _videoGenerationTraceOwner == generation;
+    void tagOutcome(String outcome) {
+      if (ownsTrace()) {
+        performance.putAttribute(traceName, 'outcome', outcome);
+      }
+    }
+
+    unawaited(performance.startTrace(traceName));
```
dcadenas: Make this trace lifetime per operation instead of starting a name-keyed trace fire-and-forget. startTrace('video_generation') is launched here, but the service registers it only after its asynchronous stop/start round-trip, so a fast completion can tag and stop before registration, while overlapping renders can have each start call stop the other render's shared-name trace. Return and capture a started trace handle, then apply outcome, clip_count, and aspect_ratio and stop that same handle; make the analogous camera change in video_recorder_bloc.dart:319 and video_recorder_bloc.dart:331-341. Add a regression test with delayed registration that asserts overlapping traces are each stopped once and retain their own attributes.

dcadenas (changes requested):
Make the new performance traces operation-scoped before merging. The fire-and-forget starts can register after a fast-failing camera initialization has already tagged and stopped, and overlapping renders still share a name-keyed trace, so tags and durations can be lost or assigned to the wrong operation. Capture a per-operation trace handle, then tag and stop that same trace for both callers.

mobile/lib/providers/video_editor_provider.dart
hm21: Done in 7c710e28 — both traces are now operation-scoped: `startOperationTrace(name)` returns a per-operation `PerformanceTrace`, and each render / camera init captures and tags/stops its own handle. No shared name key, so a fast completion can't miss a registration and overlapping operations keep independent traces. Dropped the `_videoGenerationTraceOwner` guard entirely; the legacy name-keyed API stays for feed_load / video_upload / video_event_service. Added the regression test asserting overlapping traces are each stopped once and keep their own attributes. CI green.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
hm21: Fixed at the root in 7c710e28 — `camera_startup` now captures its own operation-scoped trace handle, so a concurrent init can't stop or re-attribute it. Supersedes my "leaving as-is" above.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
hm21: Fixed at the root in 7c710e28 — the handle is held directly (no name-keyed registration step), so a fast-fail init tags/stops its own trace instead of racing a shared registration. Supersedes my "leaving as-is" above.

hm21:
@dcadenas addressed in 7c710e28 — both traces are now operation-scoped `PerformanceTrace` handles (each render / camera init owns and tags/stops its own trace; name-keyed owner-guard dropped), plus the requested overlap regression test. Full CI is green. Ready for re-review.

dcadenas (approved):
📈 video generation and camera timing are traceable

realmeylisdev (approved):
Verified `7c710e28` resolves the trace overlap at the root: each render/camera-init now owns an independent Firebase trace, so concurrent same-name operations can't stop or re-attribute each other. Regression test is mutation-confirmed (superseded render tagged `incomplete`, not the winner's `success`); analyze + the trace suites pass locally. Thanks for the operation-scoped rework.

mobile/lib/services/performance_monitoring_service.dart
```diff
@@ -98,6 +165,37 @@ class PerformanceMonitoringService implements PerformanceTraceMonitor {
     }
   }
 
+  /// Start an operation-scoped trace and return its handle.
+  ///
+  /// Unlike [startTrace], nothing is stored in [_activeTraces]: the caller
+  /// owns the returned [PerformanceTrace] and tags/stops it directly, so
+  /// overlapping operations of the same name keep independent traces. Start is
+  /// fire-and-forget so callers stay synchronous; attribute/stop calls on the
+  /// handle are delivered to the platform in order after it.
```
realmeylisdev, line 174: nit (no action needed): the ordering claim here is a touch optimistic on the fast path. `putAttribute` on the handle is purely local (it writes `MethodChannelTrace._attributes`), and `stop()` early-returns while `_traceHandle == null` — i.e. before this unawaited `start()`'s round-trip settles. So a genuinely fast op (cached permission-denied / no-camera / warm re-init) still no-ops its stop and drops the sample, leaving a started-but-unstopped native trace — same as the legacy name-keyed path, minus the shared-map corruption.

Immaterial here: renders and cold camera starts always outlast `start()`, and fast-fail samples aren't this trace's target (they never reach the console, so they can't skew the distribution). Flagging only because the reply threads describe this as "fixed at the root" — the *overlap* is fixed at the root; the fast-completion drop is unchanged (and fine).

PR #6220, opened 2026-07-20
fix(video-player): stop EventChannel sends while app is backgrounded

### LINKED ISSUE #6221: fix(video-player): stop EventChannel sends while app is backgrounded
Auto-created from PR #6220.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6220
**Branch:** `fix/video-player-background-eventchannel-crash`

hm21:
## Description

Fixes an iOS crash in the native `divine_video_player` package:
`DivineVideoPlayerInstance.sendStateUpdateOnMain()` →
`NSInternalInconsistencyException: Sending a message before the FlutterEngine has been run.`

**Root cause:** the `AVPlayer` periodic time observer (0.2s, main queue) — and other async player callbacks (KVO status/buffering, end-of-item, watchdogs) — call `sendStateUpdate()`, which sends on the `EventChannel` sink. During the iOS **resign-active → suspend** window the Flutter shell can already be torn down, but the sink closure is still non-nil (`onCancel`/`dispose()` only run on explicit Dart-side teardown, not on backgrounding). One such tick firing after teardown throws.

The **texture** path already guarded this exact window via `VideoTextureOutput.suspendFrameDelivery()` (see the comment in `onAppBackgrounded()`); the **EventChannel** path had no equivalent.

**Fix:** a new `isBackgrounded` flag gates `sendStateUpdateOnMain()`, driven by the existing `onAppBackgrounded()` / `onAppForegrounded()` lifecycle hooks (already fanned out from the plugin's `willResignActive` / `didBecomeActive` observers). It is set *after* the final synchronous paused-state send — which still runs while the shell is alive at resign-active — so the foreground path is unchanged and no legitimate updates are lost. The app declares no `audio` background mode, so playback is suspended in the background regardless.

Crashlytics: iOS, `owner: DEVELOPER`, recurring across `1.0.12` → `1.0.16`, `processState: BACKGROUND`.

**Related Issue:** Closes #6221

## Out of Scope

Hard-teardown paths (engine detach, OOM reclaim of the `FlutterViewController`, zombie-player replacement) are already covered by `detachFromEngine` / `dispose()` nulling the sink — untouched here. This PR only closes the background suspend-window gap.

## Verification

- Native Swift change only (no Dart / generated-code / ARB inputs touched; no ratchets affected).
- Manual test on device: play a video, background the app **during playback** (home / app switcher), return — no crash, playback + state resume correctly. Confirmed working.
- No native XCTest target exists in the `darwin/` package and iOS app-lifecycle isn't reproducible via the Dart test harness, so this is verified by iOS build + the manual background test above.

## Type of Change

### DIFF
--- a/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
+++ b/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
@@ -978,7 +978,7 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
     }
 
     private func sendStateUpdateOnMain() {
-        guard let player, let sink = eventSink else { return }
+        guard !isBackgrounded, let player, let sink = eventSink else { return }
 
         let currentTime = CMTimeGetSeconds(player.currentTime())
         let actualPositionMs = currentTime.millisecondsClamped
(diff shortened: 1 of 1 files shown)


dcadenas (approved):
✅ background-safe EventChannel handling, ship it

PR #5102, opened 2026-06-12
fix(media-cache): bump cupertino_http to v3 to restore NSURLSession path

hm21:
Closes #5103.
Closes #4146.

## Description

`cupertino_http` 2.4.0 could fail on Apple platforms because the Dart bindings looked for native trampoline symbols that were not present in the compiled native artifact. When that happened, `MediaCache` latched onto the `dart:io` HTTP/1.1 fallback for the rest of the process, losing the NSURLSession path, HTTP/2/3 support, and the connection pool shared with AVPlayer.

This PR now:

- bumps `cupertino_http` to `^3.0.2` and `http` to `^1.6.0`
- removes the stale iOS Podfile workaround for the old CocoaPods `cupertino_http` framework link path, which v3 no longer uses
- updates `HttpCancellableDownloader` to use `AbortableRequest`
- makes downloader shutdown cancel and await active downloads before closing the shared client, matching `cupertino_http` v3's stricter `close()` behavior
- adds regression coverage for closing while a request is pending and while a response stream is active

## Verification

- `mise exec -- flutter test packages/media_cache/test/src/cancellable_downloader_test.dart`
- `mise exec -- flutter test packages/media_cache`
- `mise exec -- flutter analyze`
- `mise exec -- flutter run -d 5876CA1C-92A0-419D-984C-1DAAD05096BC --debug`

The iOS simulator run completed `pod install`, built through Xcode, launched the app, loaded the feed, and completed media cache downloads without the `MediaCache: cupertino_http init failed` warning.

## Type of Change

- [x] Bug fix
- [x] Build configuration cleanup
- [x] Test coverage

PR #6222, opened 2026-07-20
fix(media-cache): cancel in-flight downloads when app is backgrounded

### LINKED ISSUE #6223: fix(media-cache): cancel in-flight downloads when app is backgrounded
Auto-created from PR #6222.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6222
**Branch:** `fix/media-cache-cancel-inflight-on-background`

hm21:
Closes #6223

## Description

iOS-only mitigation for a `cupertino_http` FFI crash seen in Crashlytics: `SIGABRT` in `DLRT_GetFfiCallbackMetadata` / `dart::SimpleHashMap::Lookup`, blamed on `_NativeCupertinoHttp_protocolTrampoline`.

`media_cache`'s cancellable downloader uses `cupertino_http` (NSURLSession) on Apple platforms. NSURLSession delivers delegate callbacks (`dataTask:didReceiveData:`) through a Dart FFI trampoline on its own OS threads, independent of the app lifecycle. When iOS suspends the app with a request still in flight, a late callback into the now-paused Dart isolate can't resolve its FFI callback metadata and aborts the process. Every sampled crash event had `processState: BACKGROUND`; the issue has been present since 1.0.5 and is still hitting the current 1.0.16 build.

This adds a lifecycle-driven teardown: on `paused` / `hidden` / `detached`, both media caches (`openVineMediaCache`, `openVineImageCache`) cancel their in-flight downloads, so no NSURLSession callbacks are pending when the OS suspends the isolate. The caches stay open and resume downloading on return.

Layering: the `media_cache` package gets a pure cancel hook — `MediaCacheManager.cancelInFlightDownloads()` → `CancellableDownloader.cancelActiveDownloads()` — with no Flutter-lifecycle dependency; the app layer (`_DivineAppState`, already a `WidgetsBindingObserver`) wires it to the lifecycle.

**Follow-up (2nd commit): also cancel deferred downloads.** `cancelInFlightDownloads()` initially cancelled only the downloader's active set, so an operation still awaiting `_resolveBaseCacheDir()` (the async `getTemporaryDirectory()` round-trip before `_downloader.download()` runs) was missed — `cancelledBeforeStart` stayed false, and on resume `startDownload()` would fire a fresh NSURLSession request after the isolate was suspended, i.e. the same crash. This is reachable at cold start because the image cache (`enableSyncManifest: false`) skips setting `_baseCacheDir` in `initialize()`, forcing the first grid load's ~20 concurrent thumbnail requests through that window. It now mirrors `close()`: cancel `_activeCancellableOperations` first (so `startDownload()` bails), then the downloader. Adds a regression test (verified red without the fix) plus a multi-concurrent downloader cancel test.

**Follow-up (3rd commit, review): suspend-latch so prefetch loops cannot re-arm.** The one-shot cancellation left a re-entry gap (flagged by review): prefetch loops react to a cancelled download by advancing to their next item — `preCacheFiles` resolves its batch `Future.wait` and opens the next batch, and the disk prefetcher cycle moves to the next video (nothing flips the feed's `isActive` on app-background, so `cancelCycle` may never run while frames are stopped) — re-arming fresh NSURLSession requests inside the same suspension window. `cancelInFlightDownloads()` now latches a suspension checked in `_cacheFileCancellable` (single choke point; placed after the cached fast-path so cached files still resolve) and at the top of each `preCacheFiles` batch; the new `resumeDownloads()` lifts it, wired to `AppLifecycleState.resumed` for both caches. Adds a multi-batch regression test (verified red without the latch) and a direct suspend/resume test.

**Why a mitigation and not a version bump:** `cupertino_http` 3.0.2 is already the latest on pub.dev and the app is on it; the earlier v2→v3 bump (#5102) shipped in 1.0.16 and did not fix this. This shrinks the suspension race window but can't fully close it — iOS may still deliver a callback during the suspend transition. If the crash rate stays material on a post-fix build, the reliable fix is routing the iOS `media_cache` path back to `dart:io` (which loses HTTP/2+3 and the shared AVPlayer connection pool), deliberately out of scope here.

**Related Issue:**

## Out of Scope

- The guaranteed fix (iOS `media_cache` → `dart:io` behind a flag), pending whether this mitigation is sufficient.
- Platform-gating the cancellation to Apple only — kept cross-platform on purpose; Android's `cronet_http` stack is also native and cancel-on-background is harmless there.
- Strengthening the `'cancelActiveDownloads is a no-op after close'` test (currently re-asserts `client.closed`) and adding an app-layer test for the `main.dart` lifecycle wiring — heavyweight, since the global cache singletons have no injection seam and the app layer is behavior-first.
- Tightening the `cancelActiveDownloads` "mid-iteration" code comment (removal is actually async, not synchronous) — cosmetic.

## Verification

- `flutter analyze` (media_cache): clean
- `flutter analyze lib/main.dart`: clean
- `flutter test` (media_cache): 204 passed, coverage 98.14% (≥ 98% gate)
- Follow-up fix: affected `media_cache` suites green (69 tests, incl. 2 new); the new deferred-cancel regression test was confirmed red against the pre-fix code, and the pre-push hook ran the changed-file tests green.
- Manual (iOS Simulator): feed scroll — prefetch/download/playback unaffected in the foreground; prefetch completes with `cancelled=false`, players create/dispose normally.

## Type of Change

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -1761,12 +1761,17 @@ class _DivineAppState extends ConsumerState<DivineApp>
     // Tear down in-flight media downloads before the OS suspends the app.
     // On Apple platforms cupertino_http delivers NSURLSession callbacks
     // through a Dart FFI trampoline that aborts if the isolate is suspended
-    // mid-request. The caches stay usable and resume downloading on return.
+    // mid-request. Cancellation also latches a suspension so prefetch loops
+    // cannot re-arm fresh native requests in the same window; the latch is
+    // lifted on resume, when downloads work again.
     if (state == AppLifecycleState.paused ||
         state == AppLifecycleState.hidden ||
         state == AppLifecycleState.detached) {
       openVineMediaCache.cancelInFlightDownloads();
       openVineImageCache.cancelInFlightDownloads();
+    } else if (state == AppLifecycleState.resumed) {
+      openVineMediaCache.resumeDownloads();
+      openVineImageCache.resumeDownloads();
     }
   }
 
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -1755,6 +1755,21 @@ class _DivineAppState extends ConsumerState<DivineApp>
     _memoryPressureHandler.onMemoryPressure();
   }
 
+  @override
+  void didChangeAppLifecycleState(AppLifecycleState state) {
+    super.didChangeAppLifecycleState(state);
+    // Tear down in-flight media downloads before the OS suspends the app.
+    // On Apple platforms cupertino_http delivers NSURLSession callbacks
+    // through a Dart FFI trampoline that aborts if the isolate is suspended
+    // mid-request. The caches stay usable and resume downloading on return.
+    if (state == AppLifecycleState.paused ||
+        state == AppLifecycleState.hidden ||
+        state == AppLifecycleState.detached) {
+      openVineMediaCache.cancelInFlightDownloads();
+      openVineImageCache.cancelInFlightDownloads();
+    }
+  }
+
   /// Logs a memory snapshot at info and annotates Crashlytics custom keys so
   /// OOM crash reports carry the last-seen memory footprint and gauges.
   void _emitMemorySnapshot(MemorySnapshot snapshot) {
--- a/mobile/packages/media_cache/lib/src/media_cache_manager.dart
+++ b/mobile/packages/media_cache/lib/src/media_cache_manager.dart
@@ -1335,6 +1335,16 @@ class MediaCacheManager extends CacheManager {
   /// isolate aborts the process in the FFI trampoline.
   void cancelInFlightDownloads() {
     if (_isClosed) return;
+    // Cancel manager-level operations first, mirroring [close]. An operation
+    // still awaiting _resolveBaseCacheDir() has not started its download yet,
+    // so it is absent from the downloader's active set; cancelling it here
+    // sets cancelledBeforeStart so startDownload() bails instead of issuing a
+    // fresh NSURLSession request after the isolate is already suspended.
+    for (final operation in _activeCancellableOperations.toList(
+      growable: false,
+    )) {
+      operation.cancel();
+    }
     _downloader.cancelActiveDownloads();
   }
 
--- a/mobile/packages/media_cache/lib/src/media_cache_manager.dart
+++ b/mobile/packages/media_cache/lib/src/media_cache_manager.dart
@@ -412,6 +412,13 @@ class MediaCacheManager extends CacheManager {
 
   bool _isClosed = false;
 
+  /// Latched by [cancelInFlightDownloads], cleared by [resumeDownloads].
+  /// While set, cancellable downloads complete with `null` instead of
+  /// reaching the network, so autonomous prefetch loops (disk prefetcher
+  /// cycles, [preCacheFiles] batches) cannot re-arm fresh native requests
+  /// after a background cancellation.
+  bool _downloadsSuspended = false;
+
   bool _isProcessingResponse(FileInfo fileInfo) =>
       fileInfo.statusCode == HttpStatus.accepted;
 
@@ -744,6 +751,14 @@ class MediaCacheManager extends CacheManager {
           getCachedFileSync(key) ??
           (aliasKey != null ? getCachedFileSync(aliasKey) : null);
       if (cached != null) return CancellableCacheOperation.completed(cached);
+    }
+
+    // Suspended (app backgrounded): refuse to start anything new. Prefetch
+    // loops keep requesting the next item after their current download is
+    // cancelled; without this gate they would re-arm fresh native requests
+    // right in the suspension window [cancelInFlightDownloads] just cleared.
+    if (_downloadsSuspended) {
+      return CancellableCacheOperation.fromDownload(_CompletedNullDownload());
     }
 
     // Join an in-flight download for the same key (started by either
@@ -1221,6 +1236,11 @@ class MediaCacheManager extends CacheManager {
 
     // Process in batches
     for (var i = 0; i < items.length; i += batchSize) {
+      // Backgrounded mid-run: stop opening new batches. The downloads would
+      // be refused by the suspension gate anyway; bailing here also skips
+      // the per-item cache probes. Callers re-trigger prefetch on the next
+      // interaction after resume.
+      if (_downloadsSuspended) return;
       final batch = <Future<File?>>[];
       final end = (i + batchSize > items.length) ? items.length : i + batchSize;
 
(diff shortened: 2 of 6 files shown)


mobile/packages/media_cache/lib/src/media_cache_manager.dart
```diff
@@ -1326,6 +1326,28 @@ class MediaCacheManager extends CacheManager {
...
   Future<void> waitForPendingAliasWrites() => _aliasWriteQueue;
 
+  /// Cancels in-flight cancellable downloads without closing the manager.
+  ///
+  /// Unlike [close], the manager stays usable — new downloads work again once
+  /// the app resumes. Wire this to app-background transitions so in-flight
+  /// NSURLSession (`cupertino_http`) requests are torn down before the OS
+  /// suspends the Dart isolate; a late delegate callback into a suspended
+  /// isolate aborts the process in the FFI trampoline.
+  void cancelInFlightDownloads() {
```
dcadenas, line 1364: Guard the entire prefetch loop with a persistent suspend latch, not only the operations currently in `_activeCancellableOperations`. `cancelInFlightDownloads()` cancels the current batch, then `await Future.wait(batch)` at `media_cache_manager.dart:1255` lets `preCacheFiles()` enter the next batch; its new operation passes `cancelledBeforeStart` and reaches `_downloader.download()` at line 822 while the app is backgrounding. Set the latch before cancellation, check it in `startDownload()` beside line 815, and clear it from a `resumeDownloads()` path wired to `AppLifecycleState.resumed` for both caches. Add a multi-batch regression test proving that no new downloader calls occur between cancel and resume, and that downloads work after resume.

dcadenas (changes requested):
The background cancellation path must also prevent later prefetch batches from starting new native requests. Add a persistent suspend/resume guard and a multi-batch regression test before merging so the app cannot re-enter the downloader while backgrounded.

mobile/packages/media_cache/lib/src/media_cache_manager.dart
```diff
@@ -1326,6 +1326,28 @@ class MediaCacheManager extends CacheManager {
...
+    // still awaiting _resolveBaseCacheDir() has not started its download yet,
+    // so it is absent from the downloader's active set; cancelling it here
+    // sets cancelledBeforeStart so startDownload() bails instead of issuing a
+    // fresh NSURLSession request after the isolate is already suspended.
+    for (final operation in _activeCancellableOperations.toList(
+      growable: false,
+    )) {
+      operation.cancel();
+    }
+    _downloader.cancelActiveDownloads();
```
realmeylisdev, line 1377: Confirming the re-entry gap (agree with dcadenas). This cancels only the ops active *now* and latches nothing, so the autonomous prefetch loops immediately re-arm — issuing exactly the fresh NSURLSession requests the PR is trying to prevent:

- **`DiskPrefetcher` (home feed, primary iOS path):** cancelling the active op returns from `_downloadWithFallbacks`, and the cycle advances to the next video. The prefetcher's own doc (`disk_prefetcher.dart:66-71`) says only `cancelCycle()` stops the cycle — cancelling the current download doesn't. The foreground flip that *does* call `cancelCycle` runs on the **next frame** (appForeground provider → rebuild), but the op-cancel continuation re-enters the loop as a **microtask** that drains first, so a fresh `_downloader.download()` fires before the generation bumps.
- **`preCacheFiles` (explore feed, alive on iOS via `VideoPrefetchMixin`, gated only on `kIsWeb`):** cancelling makes `Future.wait(batch)` at line 1255 resolve immediately, so the loop advances to batch N+1 with fresh ops (`cancelledBeforeStart=false`) straight to `_downloader.download()` at 822 — arguably *accelerated* by the cancel.

The image cold-start path in the description is actually the case that already works (one op per image, no self-re-entry).

A persistent suspend latch — set here, checked in `startDownload()`/`preCacheFiles`, cleared on `AppLifecycleState.resumed` for both caches — closes both paths. The multi-batch regression test dcadenas asked for would pin it; `TestableMediaCacheManager` + `downloaderOverride` and the existing `preCacheFiles` tests already give you the seam.

mobile/lib/main.dart
```diff
@@ -1755,6 +1755,21 @@ class _DivineAppState extends ConsumerState<DivineApp>
...
+  void didChangeAppLifecycleState(AppLifecycleState state) {
+    super.didChangeAppLifecycleState(state);
+    // Tear down in-flight media downloads before the OS suspends the app.
+    // On Apple platforms cupertino_http delivers NSURLSession callbacks
+    // through a Dart FFI trampoline that aborts if the isolate is suspended
+    // mid-request. The caches stay usable and resume downloading on return.
+    if (state == AppLifecycleState.paused ||
+        state == AppLifecycleState.hidden ||
+        state == AppLifecycleState.detached) {
+      openVineMediaCache.cancelInFlightDownloads();
```
realmeylisdev, line 1770: Nit (separate from the re-entry issue): `openVineMediaCache`/`openVineImageCache` are lazy top-level finals with no eager init. On a launch-then-immediately-background (still on splash) or a terminate before any media renders, this branch is the *first* touch of those globals — so it constructs a full `MediaCacheManager` (HttpClient + platform cupertino/IO client + repo handle) purely to cancel zero in-flight downloads, right at a background/terminate transition. Cheap to guard by skipping caches that were never materialized this session.

realmeylisdev (commented):
Verified dcadenas's finding independently — it holds. Crash-site coverage (cupertino path only), the layering, and the deferred-cancel test are solid; the real gap is that the cancel is one-shot, so the prefetch loops re-arm fresh cupertino requests inside the same paused→suspend window. Details + one small extra nit inline.

mobile/packages/media_cache/lib/src/media_cache_manager.dart
hm21: Done in 8fe66c7661: `cancelInFlightDownloads()` now latches a suspension, checked in `_cacheFileCancellable` (the single choke point every cancellable download passes) and at the top of each `preCacheFiles` batch; `resumeDownloads()` clears it, wired to `AppLifecycleState.resumed` for both caches in main.dart. Multi-batch regression test added and verified red without the latch, plus a direct suspend/resume test.

One deviation: no check in `startDownload()` — deferred-window ops are registered in `_activeCancellableOperations` synchronously before their first await, so the existing `operation.cancel()` already covers them, and new ops can't reach `startDownload()` past the gate. That branch would be dead code under the 98% coverage gate.

mobile/packages/media_cache/lib/src/media_cache_manager.dart
hm21: Fixed in 8fe66c7661 (details in the thread above). The manager-level gate also closes the DiskPrefetcher path: while suspended, the cycle's follow-up `cacheFileCancellable` calls return completed-null ops, so no native request fires regardless of when `cancelCycle` lands. One correction to the mechanism as described: on app-background nothing flips the feed's `isActive` at all — it's driven by route/overlay providers (`activeBranchIndexProvider`, `shellObscuredProvider`, `overlayVisibilityProvider`), and `appForegroundProvider` only gates `effectiveActive` via a rebuild, which may not happen once frames stop — so without the latch the cycle could keep downloading through the whole background grace window, not just one extra item.

mobile/lib/main.dart
hm21: Not taking this one — the premise doesn't hold: both globals are materialized by the essential startup phase (`VideoCacheManifest` → `initializeMediaCache()` + `openVineImageCache.enforceCacheLimits()` in main.dart), which runs right after launch, so any realistic lifecycle transition finds them already constructed on mobile. Only web skips that init (`kIsWeb`), where there's no NSURLSession and effectively no users — and `cancelInFlightDownloads()` on a fresh manager is a no-op anyway.

hm21:
@dcadenas Suspend latch + `resumeDownloads()` on `AppLifecycleState.resumed` + multi-batch regression test landed in 8fe66c7661 — details in the inline replies. media_cache suite green (208 tests, coverage 98.16%).

dcadenas (approved):
✅ backgrounded downloads now stop cleanly

PR #6225, opened 2026-07-20
fix(video): migrate off video_player (FVP) to divine_video_player

### LINKED ISSUE #6226: fix(video): migrate off video_player (FVP) to divine_video_player
Auto-created from PR #6225.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6225
**Branch:** `fix/replace-fvp-video-player-with-divine`

hm21:
Closes #6226

## Summary

Removes the crash-prone `video_player` (FVP) pipeline from the app, moves the last remaining usage onto `divine_video_player`, and deletes all the now-dead FVP scaffolding.

Fixes Crashlytics issue `92facabdc6fe409a7daf9be134d23be5` (iOS, build 804 / 1.0.16, ~26 events/week, all `processState: BACKGROUND`): `FVPFrameUpdater.displayLinkFired` → `FlutterEngine.textureFrameAvailable:` → `EXC_BAD_ACCESS (KERN_INVALID_ADDRESS)`, FATAL.

## Why

The official `video_player` plugin (FVP) has no `willResignActive` guard, so its `CADisplayLink` keeps pushing texture frames into the Flutter engine while the app is backgrounded and the engine shell is being torn down — a use-after-teardown. A plugin bump doesn't help: the only relevant fix ("display link stops after dispose") shipped in 2.9.4 and the app is already on 2.10.1; the background case isn't covered by it, and 2.11.0 adds nothing here.

`divine_video_player` already carries the guard: `DivineVideoPlayerPlugin` observes `UIApplication.willResignActiveNotification` and calls `suspendFrameDelivery()`. Moving the last FVP surfaces onto it removes the second, crash-prone pipeline entirely.

## What changed

**The one real migration**

- `VideoCommentPlayer` (inline comment / collaborator-invite videos) — the only widget in `lib/` that still rendered an FVP `VideoPlayer` — now uses `DivineVideoPlayerController(useTexture: true)` + `DivineVideoPlayer`, mirroring the editor/feed reference pattern. Playback stops when the app backgrounds and the controller is disposed on unmount; the native guard covers frame safety. The test-only `controllerFactory` seam is dropped in favour of the standard Divine mocked-channel test setup.

**Dead-code removal (the bulk of the diff)**

The Explore FVP pipeline was already dead: `VideoFeedItem` no longer exists, the explore router is a `Text`-only proof-of-concept, and the FVP-controller pre-init methods were never called from production. So this deletes:

- `individual_video_providers.dart` (the 1483-line provider: params, `safe*` helpers, disposed-controller tracker, fallback/auth/eviction/retry). Its only live production reader was a vestigial pause-before-comments that spun up a *separate*, never-shown controller.
- The FVP-only "stop videos on navigation" subsystem: `VideoStopNavigatorObserver` (router), `videoControllerAutoCleanupProvider` (app shell), `disposeAllVideoControllers` + its call sites. These only ever invalidated the FVP family provider, which never touched the Divine feed, so removal is behaviour-neutral — the Divine feed already stops via active-video state and widget unmount.
- Two dead FVP widgets: `video_metrics_tracker` (the live one is `DivineVideoMetricsTracker`) and `video_player_subtitle_layer`.
- The FVP telemetry gauge (`fvpLiveControllerCount` / `MemorySnapshot.fvpControllers` / the `vc_fvp` Crashlytics key). `vc_native` (Divine's `liveControllerCount`) stays.
- `video_player` + `video_player_web_hls` from `pubspec.yaml` — the Divine player ships its own native + web backends.

Net: 48 files changed, ~195 insertions, ~7.4k deletions. `rg "package:video_player" lib test` → 0.

## Testing

- `flutter analyze lib test integration_test` → clean.
- All affected tests green, including new regression tests for the comment player's background-abort and offscreen-pause paths; pre-push changed-file suite green.
- **Web verified on Chrome**: comment videos play with the default progressive MP4 *and* with HLS forced via dev options — the app's URL selector defaults to progressive `720p.mp4` on every platform, so removing `video_player_web_hls` doesn't break the real web playback path.

Remaining manual check before merge:
- iOS repro of the original crash: Explore/feed scroll → background (Home) → wait 30–60s → return; and a comment thread with a video reply, play → background → return. Confirm no `FVPFrameUpdater` crash.
- Confirm navigating feed → camera / feed → pushed route still stops feed audio (the removed FVP nav-stop subsystem was already a no-op for the Divine feed).

After rollout, this fix is referenced on Crashlytics issue `92facabdc6fe409a7daf9be134d23be5`; close it once the crash rate drops.


mobile/lib/screens/comments/widgets/video_comment_player.dart
```diff
@@ -96,39 +76,56 @@ class _VideoCommentPlayerState extends State<VideoCommentPlayer>
...
     }
 
-    if (_controller!.value.isPlaying) {
-      await _controller!.pause();
-    } else {
-      await _controller!.play();
+    setState(() => _isInitializing = true);
+    final controller = DivineVideoPlayerController(useTexture: true);
+    try {
+      await controller.initialize();
```
dcadenas, line 92: Guard the async initialization path against app backgrounding before calling `play()` at line 102. If the app has left the resumed state after `initialize()` or `setSource()`, abort and dispose this controller instead of assigning it and starting playback; the lifecycle callback at lines 70-72 cannot pause a controller that does not exist yet, and backgrounding does not unmount this widget. Add regression coverage for a player created after `willResignActive`, since late native registration can bypass the plugin's background sweep and reopen the fatal frame-delivery crash.

mobile/lib/screens/comments/widgets/video_comment_player.dart
```diff
@@ -96,39 +76,56 @@ class _VideoCommentPlayerState extends State<VideoCommentPlayer>
...
+      await controller.setSource(VideoClip.network(widget.videoUrl));
```
dcadenas, line 93: Restore an HLS playback path for desktop web, or explicitly make comment playback MP4-only and cover that contract with a web test. This call now sends `.m3u8` URLs to Divine's plain HTML video backend, while this PR removes both `video_player_web_hls` and the `hls.js` script, so Chrome and Firefox silently fail on foreign-client HLS replies. Confirm whether collaborator or comment videos can arrive as HLS before merging.

mobile/lib/screens/comments/widgets/video_comment_player.dart
```diff
@@ -140,23 +137,24 @@ class _VideoCommentPlayerState extends State<VideoCommentPlayer>
...
-    if (nextPlaying != _isPlaying) {
-      setState(() => _isPlaying = nextPlaying);
+    if (state.isPlaying != _isPlaying) {
+      setState(() => _isPlaying = state.isPlaying);
     }
   }
 
   void _onVisibilityChanged(VisibilityInfo info) {
+    if (!mounted) return;
     if (info.visibleFraction < 0.35 &&
```
dcadenas: Pause unconditionally when this comment video is below the visibility threshold, or gate the pause on `_isPlaying` rather than `_controller.state.isPlaying`. `play()` does not update `state.isPlaying` synchronously, so a visibility callback immediately after the tap can skip the pause and the later playing event does not re-check visibility. Add a regression test for scrolling the video offscreen during startup.

dcadenas (dismissed):
The migration still has lifecycle and playback regressions that must be addressed before merge. Guard async comment-player startup against backgrounding, decide whether desktop web must retain HLS comment playback, and fix the visibility race that can leave a just-started offscreen video decoding.

mobile/lib/screens/comments/widgets/video_comment_player.dart
```diff
@@ -96,39 +76,56 @@ class _VideoCommentPlayerState extends State<VideoCommentPlayer>
...
+        category: LogCategory.video,
+      );
+      // Drop the references before disposing so a later tap starts a fresh
+      // controller instead of calling into this disposed one (which would
+      // throw StateError from the package's initialized/disposed guard).
+      // Drop the references before tearing the controller down so a later tap
+      // starts a fresh controller instead of calling into this disposed one
+      // (which would throw StateError from the package's initialized/disposed
+      // guard). Teardown is fire-and-forget so the retry affordance returns
+      // immediately rather than waiting on the native player to release.
```
realmeylisdev: Nit: this comment is duplicated — lines 115–117 are an earlier draft that 118–122 already fully covers (same StateError-guard rationale). Drop the first block.

realmeylisdev (commented):
Went through the whole migration and every deletion, and it holds up. divine_video_player really does pause its display link on `willResignActive` (`DivineVideoPlayerPlugin` → `suspendFrameDelivery` → `displayLink.isPaused = true`), which is the guard FVP is missing, and FVP is fully gone from the tree, pubspec, lock and web. The prefetch swap is an exact behavioral match (both paths call the same `getCacheableVideoUrlForPlatform()` and gate identically), and the nav-stop observer / auto-cleanup provider / fvp telemetry / `individual_video_providers` deletions are all genuinely dead — zero live callers, nothing orphaned. Ran the affected suites locally: 22 green.

Two small notes, neither blocking:
- Duplicated comment block — see inline.
- The body still reads "why this is a draft" while the PR is marked ready. Worth doing the on-device iOS/web checks (especially the actual `FVPFrameUpdater` crash repro) before this merges, since that's the whole point of the change.

mobile/lib/screens/comments/widgets/video_comment_player.dart
hm21: Fixed in 5ae87737d. Added a foreground guard after the initialize()/setSource() awaits: if the app has left the foreground (paused/hidden), the controller is disposed and playback aborted instead of assigning it and calling play() in the background. Regression test added (`aborts playback when the app is not in the foreground`).

mobile/lib/screens/comments/widgets/video_comment_player.dart
hm21: Fixed in 5ae87737d. The visibility pause now gates on the optimistic `_isPlaying` flag (set right after play()) instead of `controller.state.isPlaying`, which does not flip synchronously. Regression test scrolls the player offscreen immediately after the tap and asserts the pause still fires.

mobile/lib/screens/comments/widgets/video_comment_player.dart
hm21: Fixed in 5ae87737d — dropped the duplicated block. Thanks for the thorough pass.

mobile/lib/screens/comments/widgets/video_comment_player.dart
hm21: Verified this does not reproduce. `getOptimalVideoUrlForPlatform()` returns progressive `720p.mp4` as the production default on every platform incl. web — HLS is only a dev-format override / Android-codec fallback, and it falls back to progressive (`?? videoUrl`) when a variant is not available. The home/feed already runs on the Divine web backend under this progressive-first model. Tested on Chrome: comment videos play with the default MP4 **and** with HLS forced via the dev options, so removing `video_player_web_hls` does not break the real web playback path. Happy to add a dedicated web test if you still want the contract pinned.

hm21:
@dcadenas @realmeylisdev — pushed `5ae87737d` addressing the review:

- **Backgrounding during async init (92)** — added a foreground guard after `initialize()`/`setSource()`; aborts + disposes instead of starting playback in the background. Regression test added.
- **Visibility-pause race (149)** — now gates on the optimistic `_isPlaying` flag rather than `state.isPlaying`. Regression test scrolls offscreen right after the tap.
- **Duplicated comment (122)** — dropped.
- **HLS on web (93)** — verified as a non-issue and left open for your call: the app defaults to progressive `720p.mp4` on every platform (feed included), HLS falls back to progressive when unavailable, and I tested Chrome with both the default and forced-HLS (dev options) — comment videos play fine. Can add a dedicated web test if you want the contract pinned.

Thanks both for the thorough review.

mobile/lib/screens/comments/widgets/video_comment_player.dart
```diff
@@ -96,39 +76,62 @@ class _VideoCommentPlayerState extends State<VideoCommentPlayer>
...
 
-    if (_controller!.value.isPlaying) {
-      await _controller!.pause();
-    } else {
-      await _controller!.play();
+    setState(() => _isInitializing = true);
+    final controller = DivineVideoPlayerController(useTexture: true);
+    try {
+      await controller.initialize();
+      await controller.setSource(VideoClip.network(widget.videoUrl));
```
dcadenas, line 93: Resolve `widget.videoUrl` to a progressive MP4 before constructing `VideoClip.network` on web, or restore an HLS-capable web path. At `video_comment_player.dart:93`, `VideoClip.network(widget.videoUrl)` receives the comment's final imeta URL, and `comment_item.dart:154` passes it through unchanged, so a reply with an HLS derivative reaches this call as a `.m3u8`. The web backend now assigns that URL to `<video>` after hls.js was removed, so Chrome and Firefox desktop cannot start playback. Add a web smoke test for a comment video that verifies this contract.

dcadenas (changes requested):
The comment video path still sends the last URL from the Nostr metadata directly to the web video element. When that URL is an HLS `.m3u8`, Chrome and Firefox cannot play it because this PR removes the HLS fallback. Resolve the URL to a progressive MP4 or provide HLS-capable web playback, and add the comment-video web smoke test before merge.

mobile/lib/screens/comments/widgets/video_comment_player.dart
hm21: Fixed in 65898801f — you were right. The comment player passed the raw imeta URL straight through, unlike the feed which resolves via `getOptimalVideoUrlForPlatform`. Added `VideoEvent.inlinePlayerVideoUrl`: HLS `.m3u8` URLs resolve to the progressive variant, progressive URLs pass through unchanged (no transcode-readiness risk for the common case). Wired into `comment_item.dart` and `collaborator_invite_card.dart`, so the inline `<video>` now gets progressive on every platform — matching the feed. Unit test covers the resolution.

Context on the earlier reply: current Chrome 150+, Safari and all mobile browsers *do* play HLS natively (verified on the branch build), but pre-150 Chrome and Firefox do not — this closes that gap regardless.

mobile/lib/screens/comments/widgets/video_comment_player.dart
hm21: Correction/follow-up: current Chrome 150+ plays HLS natively, but Firefox and pre-150 Chrome do not, so I implemented the progressive resolution in 65898801f to close that gap. See the sibling thread for details.

hm21:
@dcadenas — HLS-on-web addressed in `65898801f`: the inline comment / collaborator-invite player now resolves URLs progressive-first via `getOptimalVideoUrlForPlatform` (same as the feed), so no `.m3u8` reaches the `<video>` element on browsers without native HLS (pre-150 Chrome / Firefox). Progressive URLs pass through unchanged; unit test added. All four threads resolved — ready for another look.

dcadenas (approved):
✅ Divine player migration, with the old stack gone

PR #6228, opened 2026-07-20
fix(crash): downgrade all NetworkImage HTTP failures to non-fatal

### LINKED ISSUE #6229: fix(crash): downgrade all NetworkImage HTTP failures to non-fatal
Auto-created from PR #6228.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6228
**Branch:** `fix/crashlytics-recoverable-image-http-errors`

hm21:
Closes #6229

## What

Broaden `classifyRecoverableFlutterError` so that a raw Flutter `NetworkImage` load failing with **any** non-2xx HTTP status is treated as a non-fatal, recoverable media error — not just `statusCode: 404`.

## Why

Crashlytics issue `72eb9f44799fa6513097c0049f68bc3b` ("`NetworkImage._loadAsync` — HTTP request failed") is reported as **FATAL** on iOS production (last seen 1.0.16).

The app already downgrades recoverable image-load errors to non-fatal in `recoverable_flutter_error.dart`, but the check only matched `HTTP request failed, statusCode: 404`. Flutter throws the **same** `NetworkImageLoadException` for every non-200 response (`response.statusCode != HttpStatus.ok`), so 401/403/5xx failures fell through to `recordFlutterFatalError` and were logged as fatal crashes.

The 15 most recent events for this issue confirm the pattern — no 404s remain (those are already downgraded), and it is now dominated by:

- **statusCode 401** — 12 events (`media.divine.video`, `blossom.primal.net`)
- **statusCode 500** — 3 events (`media.divine.video`)

A missing/forbidden/erroring thumbnail is recoverable in every case: the widget falls back to a placeholder. Matching the status-agnostic prefix (kept behind the same image library/context guard) closes the leak for all statuses and stays host-agnostic, which also covers the `blossom.primal.net` avatar events.

## Changes

- `isImage404` → `isImageHttpFailure`: match `HTTP request failed, statusCode: ` instead of the `404` literal.
- Regression tests for 401/403/500/502 using the real Flutter error shape (`library: 'image resource service'`, `context: 'resolving an image codec'`).

## Testing

- `flutter analyze lib/utils/recoverable_flutter_error.dart test/utils/recoverable_flutter_error_test.dart` — clean
- `flutter test test/utils/recoverable_flutter_error_test.dart` — 12/12 pass

## Note

The underlying source is still the raw `NetworkImage` bypass in `video_thumbnail_widget.dart` (`_shouldBypassCacheManager`), which sidesteps `VineCachedImage`. This PR reliably stops the fatal reports via error classification; migrating that bypass off raw `NetworkImage` is a larger, separate change with desktop-reliability tradeoffs and is intentionally out of scope here.

### DIFF
--- a/mobile/lib/utils/recoverable_flutter_error.dart
+++ b/mobile/lib/utils/recoverable_flutter_error.dart
@@ -24,13 +24,14 @@ String? classifyRecoverableFlutterError(FlutterErrorDetails details) {
   final context = details.context?.toDescription() ?? '';
   final hasRecoverableMediaHost = _containsRecoverableMediaHost(error);
 
-  // A raw NetworkImage load that fails with ANY non-2xx HTTP status is
+  // A raw NetworkImage load that fails with any non-200 HTTP status is
   // recoverable — the widget falls back to a placeholder. Flutter throws the
   // same NetworkImageLoadException ("HTTP request failed, statusCode: <code>")
-  // for 404 (missing), 401/403 (unauthorized), and 5xx (server error) alike,
-  // so match the status-agnostic prefix rather than a single code. Classified
-  // by Flutter's image-loading library/context rather than host, because a
-  // failed image from any source is recoverable.
+  // for every response where statusCode != 200 — 404 (missing), 401/403
+  // (unauthorized), and 5xx (server error) alike — so match the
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
✅ cleaner image error handling, ship it

Issue #4803, opened 2026-05-28
refactor(ui): migrate raw Icons.* to DivineIcon across the codebase


realmeylisdev:
**Problem**: Many screens / widgets use raw Material `Icons.*` (`Icon(Icons.X, …)`, `IconButton(icon: Icon(Icons.X), …)`) instead of `DivineIcon(icon: DivineIconName.X)`. The project rule (`ui_theming.md` → "Icons"): use `DivineIcon` from `divine_ui` instead of raw `SvgPicture.asset` or Material `Icon`.

**Evidence**: PR #4802 sweeps 31 raw `Icons.*` → `DivineIcon` across 7 settings files (the screens touched by the #4744 WS-1 Cubit migrations), but the broader codebase has many more occurrences in untouched screens. The same audit also identified ~15 Material icons without a `divine_ui` SVG equivalent — those stay as `Icons.*` pending design system expansion.

**Mappings already validated** (PR #4802):

| Material `Icons.*` | `DivineIconName` |
| --- | --- |
| add | plus |
| arrow_forward | arrowRight |
| check_circle | checkCircle |
| chevron_right | caretRight |
| delete | trash |
| error_outline | warningCircle |
| expand_more | caretDown |
| filter_list | funnelSimple |
| info_outline | info |
| language | globe |
| lock_outline | lockSimple |
| music_note | musicNote |
| refresh | arrowClockwise |
| restore | arrowCounterClockwise |
| search | search |
| share | share |
| verified | sealCheck |
| verified_user | shieldCheck |

**Icons still missing from `divine_ui`** (need new SVGs from the designer before we can finish the sweep): `cloud_done`, `cloud_off`, `cloud_upload`, `add_circle_outline`, `remove_circle_outline`, `radio_button_checked`, `radio_button_off`, `local_fire_department`, `label_outline`, `music_off`, `open_in_new`, `people`, `search_off`, `star`, `mic`.

**Impact**: Medium. Same kind of design-system drift as #3649 (TextStyle) and #3652 (`Colors.*`): if icon styles / strokes change in `divine_ui`, raw Material icons won't pick up the update. Visual inconsistency between divine_ui-using and Material-using screens.

**Effort**: Medium for the mechanical sweep across remaining settings/widgets/screens; large if the missing-icon SVGs need to be designed and added to `divine_ui` first.

**Source**: surfaced by the #4744 WS-1 Cubit-migration code reviews; PR #4802 ships the settings-screens slice.

[...]

PR #6230, opened 2026-07-20
chore(ui): add design-system drift ratchets (#6145)

### LINKED ISSUE #6145: chore(ui): add ratchets for design-system drift
## Problem
The repo has strong architecture ratchets, but no equivalent guardrail for UI drift. A rough audit found substantial existing usage in `mobile/lib`:

- `TextStyle(` in 128 files.
- Material button / `IconButton(` patterns in 108 files.
- Raw `Colors.` / `Color(0x...)` in 39 files.
- Raw `showDialog` / `showModalBottomSheet` in 43 files.

Some of these are legitimate, but new growth should be intentional. Today, PRs can add more one-off styling while still passing CI.

## Desired outcome
UI drift is tracked with baseline files and CI fails when the count grows without an intentional baseline update.

## Scope
- Add a script similar to existing ratchets such as `check_skip_ceiling.sh`.
- Track baseline counts or file lists for:
  - raw `TextStyle(` usage outside sanctioned areas,
  - raw colors outside theme/constants/special rendering code,
  - direct Material buttons where `DivineButton` / `DivineIconButton` should be preferred,
  - raw dialog/bottom-sheet calls outside approved wrappers/patterns.
- Add the script to Mobile CI.
- Document how to reduce or intentionally update the baseline.

## Acceptance criteria
- CI fails if a PR increases unsanctioned UI drift counts.
- Existing debt is baselined rather than fixed in one giant PR.
- The script output points to concrete files that need review.

realmeylisdev:
## Description

Adds four CI ratchets that freeze the current count of raw design-system-bypassing
UI patterns under `mobile/lib` and fail a PR that grows any of them without an
intentional baseline update. Existing debt is **baselined, not fixed** — nothing is
red on this PR. Same shape as the raw-`Icons.*` ratchet (#4803).

Each guard is a thin detector over the existing `scripts/lib/numeric_ratchet.sh`
engine (the per-file `relpath<TAB>count` sibling of `list_ratchet.sh`, already used
by `check_service_sentinel_ceiling.sh` / `check_arb_error_ceiling.sh`). A file's
count may only ever shrink; a new file, a raised count, or a file dropping to zero
(regenerate to lock the win) fails CI.

| Guard | Freezes | Baselined |
|---|---|---|
| `check_raw_textstyle_ceiling.sh` | raw `TextStyle(` | 120 files |
| `check_raw_colors_ceiling.sh` | `Colors.*` / `Color(` / `Color.fromARGB(` / `Color.fromRGBO(` / `HSVColor` / `HSLColor` | 48 files |
| `check_material_button_ceiling.sh` | direct Material buttons (`IconButton`/`ElevatedButton`/`TextButton`/`FilledButton`/`OutlinedButton`/`BackButton`/`CloseButton`, incl. `.icon`/`.filled`/`.tonal`) | 71 files |
| `check_raw_dialog_ceiling.sh` | raw `showDialog` / `showModalBottomSheet` | 38 files |

**Detector precision** (word boundaries + full-line-comment exclusion):
- Excludes the sanctioned `divine_ui` set (`DivineButton`, `DivineIconButton`,
  `RoundedIconButton`, …), `DefaultTextStyle`/`AnimatedDefaultTextStyle`,
  `ElevatedButton.styleFrom` (a style helper, not a widget), dialog wrapper names
  (`showForgotPasswordDialog`, `_showDialog`), and type annotations (`Color? c`,
  `List<TextStyle>`).
- The dialog detector matches the bare identifier, so generic-arg calls
  (`showDialog<bool>(`, nested/multi-line generics) can't slip past it.

**Scope note for the design-system owner:** the baseline scope is deliberately
**strict** — it freezes everything currently in `mobile/lib`, including
`Colors.transparent`, the `lib/theme` + `lib/constants` color homes, and
`BackButton`/`CloseButton`. If you'd rather carve any of those out, it's a one-line
regex/exclusion change plus a baseline regen.

New UI must use the sanctioned replacements: `VineTheme.*Font()`, `VineTheme` colors
(`+ .withValues(alpha:)`), `DivineButton`/`DivineIconButton`, and `VineBottomSheet`
or a full-screen flow. Regenerate a baseline after an intentional migration:
`UPDATE_BASELINE=1 bash mobile/scripts/check_<name>_ceiling.sh`.

Wired as four blocking CI steps after the raw-Icons ratchet in `mobile_ci.yaml`
(CI-only, mirroring raw-Icons — the pre-push hook omits it too). Documented in
`AGENTS.md`. On this introducing PR the baselines are absent on `origin/main`, so
the engine skips the growth-vs-base (bypass) check by design.

**Related Issue:** Closes #6145

## Out of Scope

- Migrating any existing raw usage to the sanctioned components (this PR only
  freezes the debt).
- Refactoring the older inline-style ratchets (`check_raw_icons_ceiling.sh`,
  `check_file_size_ceiling.sh`) onto the shared `numeric_ratchet.sh` engine — left
  untouched to keep this PR focused.

## Verification

- Generated all four baselines; each guard runs `OK` on a clean tree.
- Bootstrap path exercised: each guard emits the "no baseline on origin/main yet
  … skipping growth-vs-base check" NOTE and passes (what CI does on this PR).
- Fixture tests: confirmed each detector FAILS on a NEW file, on GROWTH past a
  file's ceiling, and on STALE (a baselined file dropping to zero).
- False-positive checks: `DefaultTextStyle`, `DivineIconButton`,
  `ElevatedButton.styleFrom`, `showForgotPasswordDialog`, `_showDialog`, `MyColor`,
  and type annotations all count 0.
- True-positive checks: `showDialog<bool>(`, `showModalBottomSheet<Set<int>>(`,
  `ElevatedButton.icon(`, `FilledButton.tonal(` all count correctly.
- `mobile_ci.yaml` parses as valid YAML; `bash -n` clean on all four scripts.

## Type of Change

- [x] 🗑️ Chore


mobile/scripts/check_material_button_ceiling.sh
```diff
@@ -0,0 +1,91 @@
...
+RATCHET_LABEL="material_button_ceiling"
+SCAN_DIR="${MATERIAL_BUTTON_SCAN_DIR:-$MOBILE_DIR/lib}"
+PATH_PREFIX="${MATERIAL_BUTTON_PATH_PREFIX:-$MOBILE_DIR}"
+BASELINE_FILE="${MATERIAL_BUTTON_BASELINE_FILE:-$SCRIPT_DIR/baseline/material_buttons.txt}"
+BASELINE_REPO_PATH="${MATERIAL_BUTTON_BASELINE_REPO_PATH:-mobile/scripts/baseline/material_buttons.txt}"
+BASE_REF="${MATERIAL_BUTTON_BASELINE_BASE_REF:-origin/main}"
+ALLOW_NO_BASE="${MATERIAL_BUTTON_CEILING_ALLOW_NO_BASE:-0}"
+ALLOW_NO_BASE_VAR="MATERIAL_BUTTON_CEILING_ALLOW_NO_BASE"
+
+DETECT_RE='\b(IconButton|ElevatedButton|TextButton|FilledButton|OutlinedButton|BackButton|CloseButton)(\.(icon|filled|filledTonal|outlined|tonal))?\('
```
hm21: Detector misses several direct Material buttons that are live in `lib/` today, so they're neither baselined nor frozen: `FloatingActionButton` (lib/widgets/camera_fab.dart:18) and `PopupMenuButton` (curated_list_feed_screen.dart:589, user_list_people_screen.dart:610). `MaterialButton`/`RawMaterialButton`/`DropdownButton`/`SegmentedButton` and `FilledButton.tonalIcon(` would also slip past. New FABs/popup menus can be added without tripping the ratchet — a hole in a guard whose stated purpose is to freeze direct-Material-button drift. Either add them to the alternation (+ regen baseline) or call out the exclusion explicitly in the header. Same latent class exists in the dialog detector (`showGeneralDialog`/`showBottomSheet`) — no live cases yet, but worth a note.

mobile/scripts/check_raw_colors_ceiling.sh
```diff
@@ -0,0 +1,88 @@
...
+  find "$SCAN_DIR" \
+    -type f -name '*.dart' \
+    -not -path '*/l10n/generated/*' \
+    -not -path '*/.dart_tool/*' \
+    -not -path '*/build/*' \
+    ! -name '*.g.dart' ! -name '*.freezed.dart' ! -name '*.gr.dart' \
+    ! -name '*.config.dart' ! -name '*.mocks.dart' \
+    -print0 2>/dev/null \
+  | while IFS= read -r -d '' f; do
+      count="$(grep -vE '^[[:space:]]*//' "$f" 2>/dev/null \
```
hm21: Only full-line `//` comments are stripped; occurrences are counted everywhere else, including trailing comments and string literals. Adding a benign `// … Color( …` trailing comment, or a `'showDialog'`/log string, to an already-baselined file trips a false GROWTH and blocks the PR with a misleading "you added raw X" message. Bites harder than the raw-icons precedent this copies, since `Color(` / `TextStyle(` / `showDialog` appear in prose and logs far more than `Icons.foo`, and the dialog detector matches the bare word. Consider stripping trailing comments too, or at least documenting the caveat. (Applies to all four `emit_current` filters.)

hm21 (commented):
Engine reuse + CI wiring look solid, and all four baselines reproduce cleanly on a fresh run. Two coverage points on the detectors below.

mobile/scripts/check_material_button_ceiling.sh
realmeylisdev: Confirmed and fixed in 3ef7fd926 — all three named sites reproduce exactly (`camera_fab.dart:18` FloatingActionButton, `curated_list_feed_screen.dart:589` + `user_list_people_screen.dart:610` PopupMenuButton), and `MaterialButton`/`RawMaterialButton`/`DropdownButton`/`SegmentedButton`/`FilledButton.tonalIcon(` are all genuinely unmatched (0 live today).

Widened the alternation rather than documenting the exclusion, because widening is only free right now: `origin/main` has no baseline yet, so the anti-bypass check is in bootstrap mode. Post-merge the same change would have to raise two frozen ceilings (3→4, 5→6) plus add a file, which the engine rejects. Baseline goes 71→72 rows.

Both widened detectors now end in `[(<]` instead of `\(`, so the generic call sites (`PopupMenuButton<Action>(`) are caught and a type argument can't bypass.

On the dialog detector: the latent class turned out to have a live case. `nostr_settings_screen.dart:224` pushes a `DialogRoute<void>(` (barrierDismissible: false) on the root navigator — a raw modal that was invisible to the guard, and the file wasn't in the baseline at all. Added `DialogRoute`/`RawDialogRoute`/`ModalBottomSheetRoute` alongside `showGeneralDialog`/`showBottomSheet`/the Cupertino and adaptive variants. `showLicensePage` is deliberately left out — it pushes a full-screen MaterialPageRoute, which is the outcome this ratchet steers toward.

mobile/scripts/check_raw_colors_ceiling.sh
realmeylisdev: Confirmed and fixed in 3ef7fd926. Reproduced end to end: appending only `Log.info('showModalBottomSheet path retired');` — or just a trailing `// … showDialog …` — to a baselined file fails with `count(s) GREW past the frozen ceiling` and the migrate-to-VineBottomSheet hint. Documenting a migration away from a dialog could block the PR. Block comments (`/* … */`) were a third uncovered vector.

All four now run over `scripts/lib/dart_code_only.awk`, which strips line comments, nestable block comments, and string-literal bodies while preserving interpolated `${…}` expressions.

I did not use `sed 's|//.*||'`: it truncates at the `//` of a URL or of `startsWith('//')` (6 such lines in `mobile/lib`), silently dropping a real match later on the same line. Undercounting is the dangerous direction for a ceiling — it lets drift through and can fake a STALE win. Tracking string state avoids that; there's a test pinning `Text('https://divine.video', style: TextStyle(fontSize: 12))` still counting.

The filter reproduces all four committed baselines exactly (277 rows, 993 occurrences), so it contributes zero baseline churn — measured contamination today was in fact zero, which is why this is a latent-correctness fix rather than a live bug. The dialog detector's bare-word form was the widest surface, and its new `[(<]` suffix independently removes the prose/log case.

Also added `test/tools/design_system_ceiling_detectors_test.dart` — you were right that this shipped without one while the raw-Icons precedent has `raw_icons_ceiling_test.dart`. 25 tests pinning comment/string semantics, generic call-site coverage, and that real growth still fails, for all four ratchets plus the filter.

One unrelated thing I hit: `numeric_ratchet.sh`'s fallback `git fetch --depth=1 origin main` shallows the whole local clone, which destroys `merge-base` and makes the pre-push hook report phantom conflicts with main. `raw_icons_ceiling_test.dart` triggers it (it passes a deliberately nonexistent base ref). Recoverable with `git fetch --unshallow`, but it'll bite anyone running that test locally — happy to fix in a follow-up since it predates this PR.

hm21 (approved):
LGTM

PR #6231, opened 2026-07-20
feat(drafts): add duplicate project option

### LINKED ISSUE #6188: feat(drafts): add duplicate project option
### What would you like?

Add a "Duplicate Project" option for projects in Drafts, allowing creators to make a copy of an existing project.

This would make it much easier to reuse long projects with several minutes of footage without starting from scratch each time. Creators could duplicate a project and edit each copy into a separate part of a series.

### When would you use this?

When creating a series from a longer recording, where multiple videos are made from the same source footage over time.

### Anything else?

requested on discord [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1527811668619300904)

hm21:
## Description

Adds a **Duplicate** action to the drafts options sheet (between Edit and Delete), so creators can copy an existing project and edit each copy independently — e.g. turning one long recording into separate parts of a series.

Flow: tap Duplicate → `DraftsLibraryDuplicateRequested(draftId, newTitle)` → `DraftsLibraryBloc` builds the copy via `DivineVideoDraft.duplicate(...)`, saves it, and refreshes the list with the copy on top → success/failure snackbar (mirrors delete). The UI stays logic-free.

**Why share files instead of copying them:** the copy reuses the source's clip/render files on disk rather than duplicating gigabytes, mirroring how the existing publish copy already shares files. This is safe because clips are immutable inputs (trim/split are metadata-only, renders create new files) and file deletion is reference-counted via indexed DB columns, so deleting one project never strips files another still references. Result: duplication is instant, and copies edit independently. Clip rows are namespaced per draft (`<draftId>:<clipId>`), so reusing clip IDs cannot collide.

**Why numbered copy titles:** re-duplicating naively produced "Trip (copy) (copy)". Titles are now numbered and de-duplicated ("Trip (copy 1)", "Trip (copy 2)"), stripping any existing suffix first so a copy-of-a-copy continues the count. The strip pattern is derived from the localized format string itself, so it stays correct in every locale (e.g. de "(Kopie N)", ja "（コピー N）"). Untitled drafts stay untitled. New l10n keys are translated into all 17 non-English locales.

**Related Issue:** Closes #6188

## Out of Scope

None.

## Verification

- `flutter analyze` on changed lib/test — clean
- `flutter test` for the affected suites (model, bloc, event, state, widget, new `draft_copy_naming`, ARB consistency) — all green
- `dart format` — clean
- pre-commit + pre-push hooks passed

## Type of Change

### DIFF
--- a/mobile/lib/services/draft_storage_service.dart
+++ b/mobile/lib/services/draft_storage_service.dart
@@ -542,11 +542,22 @@ class DraftStorageService {
     // Delete from DB first (clips cascade via FK), then delete files
     await _draftsDao.deleteDraft(id);
 
+    // Ghost frames live only in the clip `data` blob, so — like draft-local
+    // audio — the indexed-column check can't tell that a surviving draft (e.g.
+    // a duplicate sharing this draft's clips) still references them. Scan the
+    // other drafts' clips for shared ghost frames and keep those alive.
+    // Exclude this draft's own clips so its ghost frames aren't kept alive by
+    // rows that haven't cascade-deleted yet.
+    final referencedGhostFrames = await _referencedGhostFrameFilenames(
+      excludeDraftId: id,
+    );
+
     // Delete clip files only if not referenced by clip library
     await FileCleanupService.deleteRecordingClipsFiles(
       draft.clips,
       draftsDao: _draftsDao,
       clipsDao: _clipsDao,
+      referencedGhostFrameFilenames: referencedGhostFrames,
     );
 
     // Delete final rendered clip if present
@@ -555,6 +566,7 @@ class DraftStorageService {
         draft.finalRenderedClip!,
         draftsDao: _draftsDao,
         clipsDao: _clipsDao,
+        referencedGhostFrameFilenames: referencedGhostFrames,
       );
     }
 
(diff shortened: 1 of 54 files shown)


mobile/lib/services/draft_storage_service.dart
```diff
@@ -621,6 +633,42 @@ class DraftStorageService {
...
+  /// indexed column), so a draft that shares clips with another draft — a
+  /// duplicate, or a publish copy — cannot be protected by the indexed-column
+  /// check when the sibling is deleted. This scans every remaining clip's blob
+  /// and collects the shared ghost basenames so deletion keeps them alive until
+  /// the last referencing clip is gone. Never throws: a corrupt blob is logged
+  /// and skipped.
+  Future<Set<String>> _referencedGhostFrameFilenames({
+    String? excludeDraftId,
+  }) async {
+    final clipRows = await _clipsDao.getAllClips();
```
dcadenas: Make `_referencedGhostFrameFilenames` include trashed clips before deleting shared files. At this line, `getAllClips()` excludes rows with `deletedAt`, while the matching `isFileReferenced` check counts them, so deleting one duplicate can leave a restorable clip with a dangling `ghostFramePath`. Add a regression test that duplicates a draft, trashes one sharing clip, deletes the other draft, and verifies the shared ghost file survives.

dcadenas (changes requested):
The duplicate cleanup can delete a shared ghost-frame file while a trashed clip still references it. Include trashed clips in the reference scan before this PR merges.

hm21:
Good catch — fixed in 7a95952344. The ghost-frame reference scan now includes trashed clips (`getAllClips(includeTrashed: true)`), matching `isFileReferenced` which already protected their video/thumbnail. Same gap fixed in `clearAllDrafts`. Regression tests added for both paths.

dcadenas (approved):
✅ duplicate projects, neatly covered

PR #6232, opened 2026-07-20
fix(video-recorder): crash-proof clip-delete Undo and finalize it on recorder exit

### LINKED ISSUE #6233: fix(video-recorder): keep clip-delete Undo working after leaving recorder
Auto-created from PR #6232.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6232
**Branch:** `fix/clip-delete-undo-deactivated-ref`

hm21:
Closes #6233

## What

Fix an iOS **FATAL** crash when tapping **Undo** on the recorder's "Clip moved to trash" snackbar after the hosting widget unmounts, and finalize a pending clip deletion when the user proceeds from the recorder to the next step.

Crashlytics `19a28701ebed005db8ea4d95a432c812` — `Bad state: Using "ref" when a widget is about to or has been unmounted is unsafe` at `clip_delete_snackbar.dart:28`, first seen 1.0.14, last seen 1.0.16.

## Why

**Crash fix.** The snackbar is shown through `ScaffoldMessenger`, which sits above the `Navigator`, so it outlives the widget that showed it — the delete button can unmount on recorder rebuilds (deleting the only clip hides it) and the snackbar survives route changes. The Undo callback held that widget's `WidgetRef` and called `ref.read(...)` at tap time, which throws once the widget is gone. `clipManagerProvider` is a global keep-alive `NotifierProvider` (never invalidated anywhere), so capturing its notifier synchronously, while `ref` is still valid, removes the deferred `ref` access entirely.

**Finalize on recorder exit.** Fixing the crash made Undo reachable after leaving the recorder, but both next steps snapshot the clip list at navigation time: classic mode starts the final render in `openVideoEditorFromRecorder`, and capture/lip-sync seed the editor session from a one-shot read. Honoring an Undo after that point silently diverges state — in classic the published video would omit the restored clip (`undoPendingDeletion` never invalidates `finalRenderedClip`, and the metadata preview player initializes once), and in capture/lip-sync the open editor timeline would never show it. Rather than propagating the undo into an in-flight render and a live editor session, the navigation boundary now commits the pending deletion and dismisses the snackbar — consistent with `addClip`/`reorderClip`/`removeClipById`, which already commit a pending deletion when the user moves on. The clip stays in library trash for 30 days, so it remains recoverable. Undo keeps working on the recorder itself and on paths with no snapshot (recorder close, clips-library push), which is where the crash actually fired.

## Test

- `clip_delete_snackbar_test.dart` (new): regression test that shows the snackbar, unmounts the host so its `ref` deactivates, taps Undo, and asserts the notifier is still reached. It fails against the old code with the exact `Bad state` error; a second test covers the still-mounted path.
- `video_recorder_navigation_test.dart`: `openVideoEditorFromRecorder` finalizes a pending deletion and dismisses the Undo snackbar before navigating.
- `clip_manager_provider_test.dart`: `commitPendingDeletion()` clears the pending marker so a later undo no-ops (`restore` is never called).
- `flutter analyze lib test integration_test` clean; all three affected suites pass.

Manual: record two clips → delete last → **Next** within 5s → snackbar dismissed, publish contains only the first clip, deleted clip sits in library trash. Undo on the recorder before Next restores the clip and the published video includes it. Deleting the only clip (the delete button unmounts) then tapping Undo no longer crashes.

### DIFF
--- a/mobile/lib/providers/clip_manager_provider.dart
+++ b/mobile/lib/providers/clip_manager_provider.dart
@@ -187,7 +187,7 @@ class ClipManagerNotifier extends Notifier<ClipManagerState> {
     // A new recording supersedes any pending undo from a previous tap.
     if (state.pendingDeletion != null) {
       _cancelPendingDeletionTimer();
-      unawaited(_commitPendingDeletion());
+      unawaited(commitPendingDeletion());
     }
 
     final clipDuration =
@@ -403,7 +403,7 @@ class ClipManagerNotifier extends Notifier<ClipManagerState> {
   Future<bool> removeClipById(String clipId) async {
     if (state.pendingDeletion != null) {
       _cancelPendingDeletionTimer();
-      await _commitPendingDeletion();
+      await commitPendingDeletion();
     }
 
     final index = _clips.indexWhere((c) => c.id == clipId);
@@ -470,7 +470,7 @@ class ClipManagerNotifier extends Notifier<ClipManagerState> {
   void reorderClip(int oldIndex, int newIndex) {
     if (state.pendingDeletion != null) {
       _cancelPendingDeletionTimer();
-      unawaited(_commitPendingDeletion());
+      unawaited(commitPendingDeletion());
     }
 
     if (oldIndex < 0 ||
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
✅ clip-delete undo survives recorder exit cleanly

PR #6244, opened 2026-07-21
fix(video-feed): correct empty profile feed diagnostics

### LINKED ISSUE #6243: fix(video-feed): correct empty profile feed diagnostics
## Summary
The empty-profile diagnostic is producing misleading ERROR logs. It compares an author-scoped subscription against a global direct database query, then reports `subscription filtering is too restrictive OR subscription stream is broken`.

## Evidence
From `/Users/lizsw/openvine_full_logs_2026-07-20T14-40-35.552269.txt`:

- 59x `subscription filtering is too restrictive OR subscription stream is broken`
- Subscription filters include `authors: 1 authors`
- Direct query reports `100 video events`, but sample events are from unrelated pubkeys
- The same log shows profile subscriptions successfully returning `1`, `35`, `60`, and `200` events elsewhere

## Expected Behavior
The diagnostic control query should use the same author/kind filters as the subscription. If the filtered query also returns zero, log this as an expected empty profile state, not an ERROR.

## Notes
This should reduce Crashlytics/log noise and prevent bad triage. Empty profiles are a valid domain state, not automatically a broken subscription stream.

NotThatKindOfDrLiz:
## Summary

- Run empty-feed auto diagnostics against the subscription's own filters instead of a global video-only probe.
- Cap diagnostic probe limits without dropping authors, kinds, repost filters, tags, or time bounds.
- Reclassify zero matching diagnostic results as an expected empty state, while preserving the error path when matching cached events exist but the subscription returned zero.
- Add regression coverage for author-scoped profile diagnostics, expected-empty logging, and the genuine subscription-broken branch.

## Root Cause

_runAutoDiagnostics compared an author-scoped subscription against an unscoped kinds:[34236] database query. For a profile with no videos, the global probe could still find unrelated events and log a false ERROR that the subscription was too restrictive or broken.

## Validation

- flutter analyze
- flutter test test/services/video_event_service_eose_empty_test.dart
- Pre-push hook: changed-file analyze and test/services/video_event_service_eose_empty_test.dart

Fixes #6243

### DIFF
--- a/mobile/lib/services/video_event_service.dart
+++ b/mobile/lib/services/video_event_service.dart
@@ -5874,6 +5874,17 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
     SubscriptionType subscriptionType,
   ) => _applyLikeCountToVideo(videoId, likeCount, subscriptionType);
 
+  List<Filter> _buildDiagnosticProbeFilters(List<Filter> filters) {
+    return filters.map((filter) {
+      final diagnosticFilter = Filter.fromJson(filter.toJson());
+      final existingLimit = diagnosticFilter.limit;
+      diagnosticFilter.limit = existingLimit == null || existingLimit > 100
+          ? 100
+          : existingLimit;
+      return diagnosticFilter;
+    }).toList();
+  }
+
   /// Run automatic diagnostics when feed fails to load events
   /// This logs relay status, connection info, and tests direct queries to help debug
   Future<void> _runAutoDiagnostics(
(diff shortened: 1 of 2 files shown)


hm21 (approved):
LGTM. Pushed a follow-up commit: the diagnostic probe now queries each filter individually — otherwise repost-enabled home/discovery/profile feeds skipped the local cache (queryEvents only reads cache for single-filter queries) and the "subscription broken" branch could never fire — plus type-neutral empty-state wording since the diagnostic runs for every subscription type.

PR #6245, opened 2026-07-21
fix(authentication): allow offline access for existing sessions

### LINKED ISSUE #6133: fix(authentication): allow offline access for existing sessions
### Summary

Users with an existing session are shown the login screen when opening the app without an internet connection.

### Environment

iOS, 1.0.16 804

### Steps to Reproduce

1. Log in to the app while connected to the internet.
2. Disable WiFi and cellular data.
3. Reopen the app and try to record content for later upload.

### Actual Result

The app shows the login screen and prevents access to recording features.

### Expected Result

The app should keep the existing session active offline and allow users to record content for upload when connectivity returns.

### Regression?

Unknown

### Evidence

multiple people on discord mention that problem [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1526613257035583620)

NotThatKindOfDrLiz:
## Summary
- distinguish Keycast refresh transport failures from server-side refresh rejection
- keep owner-bound Divine OAuth sessions authenticated in a pubkey-only, RPC-upgrading state when startup refresh fails offline
- preserve automatic recovery by scheduling background RPC upgrade and keeping expired-session retry affordances armed until refresh succeeds
- apply the cross-account recovery-anchor guard to degraded offline restore as well as direct and refresh-success restore
- preserve refresh tokens on network failures and keep rejection/null refreshes unauthenticated

## Tests
- flutter test packages/keycast_flutter/test/oauth_client_test.dart test/services/auth/oauth_session_coordinator_test.dart test/services/auth_service_expired_session_downgrade_test.dart test/services/auth/signer_factory_test.dart test/services/auth_service_multi_account_test.dart
- flutter analyze
- pre-push hook affected analyze/tests
- GitHub checks green on f985982b

Closes #6133

### DIFF
--- a/mobile/lib/services/auth_service.dart
+++ b/mobile/lib/services/auth_service.dart
@@ -4,7 +4,6 @@
 // with secure storage
 
 import 'dart:async';
-import 'dart:io';
 
 import 'package:cache_sync/cache_sync.dart';
 import 'package:flutter/foundation.dart';
@@ -801,9 +800,10 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
           name: 'AuthService',
           category: LogCategory.auth,
         );
-        if (await _restoreDegradedDivineOAuthSession(session)) {
-          return;
-        }
+        await _restoreDegradedDivineOAuthSession(
+          session,
+          anchorNpub: anchorNpub,
+        );
         return;
       } on TimeoutException catch (e) {
         Log.warning(
@@ -812,19 +812,10 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
           name: 'AuthService',
           category: LogCategory.auth,
         );
-        if (await _restoreDegradedDivineOAuthSession(session)) {
-          return;
-        }
-        return;
-      } on SocketException catch (e) {
-        Log.warning(
-          'initialize: synchronous refresh failed due to socket error: $e',
-          name: 'AuthService',
-          category: LogCategory.auth,
+        await _restoreDegradedDivineOAuthSession(
+          session,
+          anchorNpub: anchorNpub,
         );
-        if (await _restoreDegradedDivineOAuthSession(session)) {
-          return;
-        }
         return;
       }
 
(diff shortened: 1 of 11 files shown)


mobile/lib/services/auth_service.dart
```diff
@@ -847,6 +868,30 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
     _setAuthState(AuthState.unauthenticated);
   }
 
+  Future<bool> _restoreDegradedDivineOAuthSession(
```
dcadenas: At `mobile/lib/services/auth_service.dart:871-893`, guard this degraded restore before calling `signInWithDivineOAuth`. The direct and refresh-success branches use `_isCrossAccountRestore(...)` and leave the app unauthenticated without consuming the session when the anchor belongs to another account, but this helper skips that check, so an offline cold start can sign in as B and then have `_setupUserSession` rewrite `last_used_npub` and clear A's recovery anchor. Apply the same guard here, then schedule `unawaited(_upgradeDivineRpcInBackground(session))` after establishing the `upgrading` session so connectivity returning in the same process restores signing capability. Since the three catch sites do not use the helper's result, make the helper return `void` as part of this cleanup.

dcadenas (changes requested):
The degraded offline restore path must match the existing restore branches before this can merge. Add the cross-account guard so an offline startup cannot silently sign in as the wrong account, and keep in-session recovery active so connectivity can restore signing without an app restart.

NotThatKindOfDrLiz:
Takeover update pushed in f985982b.

What changed:
- degraded offline OAuth restore now uses the same session recovery anchor guard as the direct and refresh-success paths, so it cannot silently restore the wrong Keycast account
- degraded restore now keeps the expired-session retry flag armed and starts _upgradeDivineRpcInBackground, so signing upgrades automatically when refresh succeeds and manual retry affordances remain available if it does not
- pubkey-only Divine OAuth identity construction is now an explicit opt-in for the degraded path instead of a global fallback
- removed the unreachable SocketException branch/dead helper return flow and updated the OAuth network-exception contracts
- added regression coverage for reconnect restoring KeycastNostrIdentity, degraded cross-account blocking, and the signer-factory opt-in invariant

Verification:
- flutter test packages/keycast_flutter/test/oauth_client_test.dart test/services/auth/oauth_session_coordinator_test.dart test/services/auth_service_expired_session_downgrade_test.dart test/services/auth/signer_factory_test.dart test/services/auth_service_multi_account_test.dart
- flutter analyze
- pre-push hook affected analyze/tests
- GitHub checks are green on f985982b

dcadenas (approved):
🔒 existing sessions now work offline

PR #6249, opened 2026-07-21
fix(authentication): retry degraded OAuth restore on resume

### LINKED ISSUE #6248: fix(authentication): retry degraded OAuth restore on resume
## Context

PR #6245 fixed offline cold start for existing Divine OAuth sessions by restoring an owner-bound, pubkey-only degraded session when Keycast refresh cannot reach the network.

One follow-up gap remains: once the immediate background RPC upgrade fails while the device is still offline, app resume does not automatically retry the upgrade for the degraded pubkey-only state.

## Problem

`AuthService._refreshOAuthTokenOnResume()` currently exits when `_keycastSigner == null`. The degraded offline restore path intentionally has no `_keycastSigner`, so a user can land in this sequence:

1. Cold launch offline with an existing OAuth session.
2. App restores authenticated in pubkey-only degraded state.
3. Immediate background refresh fails due to no connectivity.
4. User restores connectivity and foregrounds the app.
5. Resume refresh exits early, so signing is not rebuilt unless the user manually taps a retry surface.

## Acceptance Criteria

- Degraded Divine OAuth sessions with `hasExpiredOAuthSession == true` can retry RPC upgrade on app resume even when `_keycastSigner == null`.
- The retry remains owner-bound and cannot attach a stale signer after sign-out or account switch.
- Add regression coverage for the still-offline terminal state and the successful resume retry path.
- Preserve the refresh token on transport failure.

NotThatKindOfDrLiz:
## Summary
- Allow app-resume OAuth refresh for degraded Divine OAuth sessions that are authenticated but have no Keycast signer yet.
- Preserve healthy signer-present OAuth sessions on resume without rotating refresh tokens.
- Rebuild a missing resume signer only from an owner-matching stored RPC session, with ownership rechecked after async banner-dismissal cleanup.
- Route remaining resume refresh through the existing owner-bound background RPC upgrade path so stale results after sign-out/account changes are discarded.
- Treat offline/network resume refresh failures as expected warning-level failures while preserving the refresh token and degraded authenticated state.

## Root Cause
#6245 introduced a degraded Divine OAuth state where _authSource == divineOAuth and _hasExpiredOAuthSession == true, but _keycastSigner == null. The resume hook still used signer presence as its OAuth-session gate, so it skipped the exact state that needed a retry after connectivity returned.

The first retry implementation also sent healthy signer-present resumes through refresh, which could rotate or lose the durable refresh token unnecessarily. Resume signer restoration now returns immediately for owner-matching valid RPC sessions, and only rebuilds the signer when it is actually missing.

## Validation
- dart format lib/services/auth_service.dart test/services/auth_service_expired_session_downgrade_test.dart
- mise exec -- dart format lib test integration_test
- flutter test test/services/auth_service_expired_session_downgrade_test.dart
- flutter test test/services/auth_service_expired_session_downgrade_test.dart test/services/auth_service_multi_account_test.dart test/services/auth_service_local_first_startup_test.dart test/services/auth/oauth_session_coordinator_test.dart
- flutter analyze lib/services/auth_service.dart test/services/auth_service_expired_session_downgrade_test.dart
- Pre-push hook: merge-conflict check, analyzer, changed auth test file

Closes #6248

### DIFF
--- a/mobile/lib/services/auth_service.dart
+++ b/mobile/lib/services/auth_service.dart
@@ -693,7 +693,7 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
       final storedSession = await KeycastSession.load(_flutterSecureStorage);
       final storedOwnerPubkey = storedSession?.userPubkey;
       if (upgradeOwnerPubkey != null &&
-          storedOwnerPubkey != null &&
+          storedSession != null &&
           storedOwnerPubkey != upgradeOwnerPubkey) {
         Log.warning(
           'OAuth RPC upgrade: refusing refresh for owner $upgradeOwnerPubkey '
@@ -729,6 +729,15 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
           category: LogCategory.auth,
         );
         await _clearDismissedDivineLoginBannerForCurrentUser();
+        if (!upgradeContextStillCurrent()) {
+          Log.warning(
+            'OAuth RPC upgrade: discarding refreshed signer after banner clear '
+            'because the signed-in account changed',
+            name: 'AuthService',
+            category: LogCategory.auth,
+          );
+          return;
+        }
         _keycastSigner = KeycastRpc.fromSession(
           _oauthConfig,
           refreshed,
@@ -4259,25 +4268,27 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
       final session = await _oauthClient.getSession();
       if (!resumeContextStillCurrent()) return;
 
-      if (_keycastSigner == null &&
-          session != null &&
+      if (session != null &&
           session.hasRpcAccess &&
           session.userPubkey == resumeOwnerPubkey) {
-        Log.info(
-          '📱 App resumed - rebuilding OAuth RPC signer from stored session',
-          name: 'AuthService',
-          category: LogCategory.auth,
-        );
-        await _clearDismissedDivineLoginBannerForCurrentUser();
-        _keycastSigner = KeycastRpc.fromSession(
-          _oauthConfig,
-          session,
-          onTokenRefresh: _refreshAccessToken,
-        );
-        _currentIdentity = _buildIdentity();
-        _hasExpiredOAuthSession = false;
-        _setRpcCapability(AuthRpcCapability.rpcReady);
-        _authStateController.add(_authState);
+        if (_keycastSigner == null) {
+          Log.info(
+            '📱 App resumed - rebuilding OAuth RPC signer from stored session',
+            name: 'AuthService',
+            category: LogCategory.auth,
+          );
+          await _clearDismissedDivineLoginBannerForCurrentUser();
+          if (!resumeContextStillCurrent()) return;
+          _keycastSigner = KeycastRpc.fromSession(
+            _oauthConfig,
+            session,
+            onTokenRefresh: _refreshAccessToken,
+          );
+          _currentIdentity = _buildIdentity();
+          _hasExpiredOAuthSession = false;
+          _setRpcCapability(AuthRpcCapability.rpcReady);
+          _authStateController.add(_authState);
+        }
         return;
       }
 
--- a/mobile/lib/services/auth_service.dart
+++ b/mobile/lib/services/auth_service.dart
@@ -721,6 +721,12 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
         _setRpcCapability(AuthRpcCapability.rpcReady);
         return;
       }
+    } on OAuthNetworkException catch (e) {
+      Log.warning(
+        'initialize: background RPC refresh failed due to network: $e',
+        name: 'AuthService',
+        category: LogCategory.auth,
+      );
     } catch (e) {
       Log.error(
         'initialize: background RPC refresh failed: $e',
@@ -4217,7 +4223,11 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
 
   Future<void> _refreshOAuthTokenOnResume() async {
     try {
-      if (_oauthClient == null || _keycastSigner == null) return;
+      if (_oauthClient == null) return;
+      final isDegradedOAuth =
+          _authSource == AuthenticationSource.divineOAuth &&
+          _hasExpiredOAuthSession;
+      if (_keycastSigner == null && !isDegradedOAuth) return;
 
       final resumeOwnerPubkey = currentPublicKeyHex;
       if (resumeOwnerPubkey == null) return;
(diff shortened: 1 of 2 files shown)


dcadenas (dismissed):
The resume recovery gate still treats a persisted OAuth session as proof that signing is available, so it can skip rebuilding a missing signer. Update the gate to check signer capability and add a regression test for a non-expired session with no installed signer before merging.

---

- **Resume recovery returns early on session presence, not on whether a signer is actually installed** ([mobile/lib/services/auth_service.dart:4240](https://github.com/divinevideo/divine-mobile/blob/b90be8e64c9312b6cc42c1e0bc2985a466a4e110/mobile/lib/services/auth_service.dart#L4240))
  Guard this return on a usable signer being installed for the current owner, not merely on `session != null`. `getSession()` can return a non-expired session without RPC access, while this file defines the degraded state by missing `hasRpcAccess`; in that case the resume path skips recovery and `_keycastSigner` remains null. Rebuild from the stored session when it is RPC-capable, otherwise continue to `_upgradeDivineRpcInBackground`, and add a regression test for this present-session-with-no-signer case.

NotThatKindOfDrLiz:
Takeover follow-up pushed in 665190535.\n\nChanges made:\n- Replaced the resume gate with an owner/capability-aware path: matching stored RPC sessions rebuild the missing Keycast signer locally; non-RPC or non-matching sessions fall through to the owner-bound background upgrade.\n- Removed the unreachable resume-level OAuthNetworkException catch and made background RPC upgrade logs no longer say initialize-only.\n- Avoided redundant auth-state nudges on failed resume retries so offline degraded resumes do not re-trigger the expired-session UI path.\n- Added the requested missing-signer regression and tightened the async test helpers/assertions.\n\nVerification:\n- dart format lib/services/auth_service.dart test/services/auth_service_expired_session_downgrade_test.dart\n- flutter test test/services/auth_service_expired_session_downgrade_test.dart\n- flutter test test/services/auth_service_expired_session_downgrade_test.dart test/services/auth_service_multi_account_test.dart test/services/auth_service_local_first_startup_test.dart test/services/auth/oauth_session_coordinator_test.dart\n- flutter analyze lib test integration_test\n- pre-push hook: analyzer + changed auth test file\n- GitHub checks are green.

mobile/lib/services/auth_service.dart
```diff
@@ -4227,33 +4243,37 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
...
+        _authStateController.add(_authState);
+        return;
       }
+
+      Log.info(
+        '📱 App resumed - OAuth RPC session unavailable, refreshing',
+        name: 'AuthService',
+        category: LogCategory.auth,
+      );
+      await _upgradeDivineRpcInBackground(
```
dcadenas, line 4300: Guard the fallback at `mobile/lib/services/auth_service.dart:4272` before calling `_upgradeDivineRpcInBackground`. When the active account is A but the stored session belongs to known account B, this call refreshes B's global token and can install the result while labeling it as A. In `_upgradeDivineRpcInBackground`, return without refreshing when `session.userPubkey` and the expected owner are both non-null and differ. Add a resume regression test that asserts zero refresh calls and leaves A signerless and degraded, while preserving null-session and matching-owner recovery.

dcadenas (dismissed):
The resume recovery path needs an account-ownership guard before this PR can merge. Reject mismatched stored sessions and add regression coverage so retrying a degraded account cannot refresh or install another account's OAuth credential.

NotThatKindOfDrLiz:
Takeover update pushed in 97385f99b.

What changed:
- Added a raw `KeycastSession.load()` ownership guard inside `_upgradeDivineRpcInBackground`, so expired/null `getSession()` resume paths cannot refresh a different account’s global token and label it as the active owner.
- Limited resume signer rebuilds to the missing-signer case and clear the dismissed Divine login banner on successful rebuild.
- Restored the startup refresh-count coverage with deterministic waiting, owner-bound the local-key expired-session fixture, and added mismatched-account resume regressions for both null-session and valid-but-wrong-session cases.
- Added AuthService disposal in this test file so the changed-file test run completes cleanly.

Verification:
- `flutter analyze lib/services/auth_service.dart test/services/auth_service_expired_session_downgrade_test.dart`
- `flutter test test/services/auth_service_expired_session_downgrade_test.dart`
- Pre-push hook: analyzer + changed auth test file

Note for the unresolved thread: the guard lives inside `_upgradeDivineRpcInBackground` and reads the raw stored session because `_oauthClient.getSession()` returns null for expired sessions, which is the degraded resume shape this PR is fixing.

mobile/lib/services/auth_service.dart
```diff
@@ -4227,33 +4258,39 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
...
-        name: 'AuthService',
-        category: LogCategory.auth,
-      );
-      final refreshed = await _oauthCoordinator.refreshSession(
-        expectedOwnerPubkey: resumeOwnerPubkey,
-      );
-      if (!resumeContextStillCurrent()) {
-        Log.warning(
-          '📱 App resumed - discarding stale OAuth refresh result',
+      if (_keycastSigner == null &&
```
dcadenas: Return from this valid owner-matching session branch even when `_keycastSigner` is already present. The new null check excludes healthy sessions, so every resume with a valid signer falls through to `_upgradeDivineRpcInBackground`, calls `refreshSession`, and can delete the durable refresh token after a non-200 response. Keep the signer rebuild conditional inside the branch, and add a regression test that asserts a signer-present valid session makes zero refresh calls.

mobile/lib/services/auth_service.dart
```diff
@@ -4227,33 +4258,39 @@ class AuthService implements BackgroundAwareService, BlockListSigner {
...
+          session.userPubkey == resumeOwnerPubkey) {
+        Log.info(
+          '📱 App resumed - rebuilding OAuth RPC signer from stored session',
           name: 'AuthService',
           category: LogCategory.auth,
         );
-        return;
-      }
-      if (refreshed != null) {
+        await _clearDismissedDivineLoginBannerForCurrentUser();
```
dcadenas: Re-check `resumeContextStillCurrent()` immediately after this await and return before installing the captured session. Sign-out or an account switch can complete while banner dismissal is suspended, so the following writes can restore the old account's signer and RPC-ready state after ownership has changed. Add a deterministic test that blocks the banner-clear await while the auth context changes.

dcadenas (changes requested):
The resume path still has two authentication-state correctness gaps. Fix the valid-session return so healthy signer-present sessions do not refresh, and revalidate ownership after the banner-clear await before restoring a signer.

NotThatKindOfDrLiz:
Takeover update pushed in `34f0ef775`:

- Healthy signer-present OAuth resumes now return from the owner-matching valid-session branch without calling refresh, so foregrounding no longer rotates or risks deleting the durable refresh token.
- Missing-signer resume rebuild still works, but signer installation is guarded after the banner-dismissal await.
- `_upgradeDivineRpcInBackground` now re-checks ownership after banner dismissal before attaching a refreshed signer, and owner-bound refresh rejects legacy stored sessions with no `userPubkey`.
- Added regression coverage for healthy resume `refreshCalls == 0` and for sign-out during banner-clear before resume signer installation.

Validation:

- `flutter test test/services/auth_service_expired_session_downgrade_test.dart`
- `flutter test test/services/auth_service_expired_session_downgrade_test.dart test/services/auth_service_multi_account_test.dart test/services/auth_service_local_first_startup_test.dart test/services/auth/oauth_session_coordinator_test.dart`
- `flutter analyze lib/services/auth_service.dart test/services/auth_service_expired_session_downgrade_test.dart`
- Pre-push hook: merge-conflict check, analyzer, changed auth test file

dcadenas (approved):
🔒 degraded OAuth resumes now recover cleanly

Issue #6253, opened 2026-07-21
fix(video-player): distinguish expired-token 401 from age-gate


hm21:
Follow-up to #6242.

`classifyVideoError` maps **every** 401 → `ageRestricted` → "Verify Age" (`packages/infinite_video_feed/lib/src/utils/playback_sources.dart:78`, `lib/widgets/video_feed_item/pooled_video_error_overlay.dart:146-147`). Correct for a genuine age gate, but an **expired/invalid BUD-01 token** or a non-age 401 produces the same "Verify Age" card — which cannot fix the problem, leaving the viewer stuck on a mislabeled dead item.

### Proposed
After an authenticated retry still returns 401, stop showing "Verify Age" and present a generic "unavailable" + skip. Requires tracking post-auth-retry state.

⚠️ **Trust/Safety review** — touches the age-gate flow; split out of the graceful-fallback PR.

NotThatKindOfDrLiz:
Adding the age-gate retry findings from the #6251 re-scope so they stay with this issue rather than becoming a fourth ticket.

Two client-side paths appear to keep the Verify-age CTA from succeeding on a default install:

- After the 18+ dialog, ContentFilterService.unlockAdultCategories intentionally skips adultCategories, so adultPlaybackPreference can remain hide.
- MediaAuthInterceptor then returns ViewerAuthBlockedByPreference when adultPlaybackPreference is still hide after verification.
- MediaViewerAuthService also folds the signed-out case into ViewerAuthUnavailable, so the UI can report a generic could-not-verify-age diagnosis instead of distinguishing not signed in.

This should be handled here with the CTA/retry behavior, separate from #6723, which only narrows persistent 404 pruning.

Issue #6252, opened 2026-07-21
fix(video-player): map iOS NSURLErrorDomain -1013 (auth required)


hm21:
Follow-up to #6242.

`DivineVideoPlayerInstance.swift` has no case for `NSURLErrorUserAuthenticationRequired` (**-1013**); it falls through to `unknown`, which has `shouldFailover == false` and `isTransient == false` (`packages/.../player_error_code.dart:65-91`), so it neither fails over to alternate sources nor routes into the age-gate flow. String-based `classifyVideoError` (`packages/infinite_video_feed/lib/src/utils/playback_sources.dart:70-96`) then lands it as `generic` → a Retry button that keeps re-hitting the same 401.

Reported in #6242 as `NSURLErrorDomain error -1013`.

### Proposed
Map -1013 to the auth/age-gate path (equivalent to HTTP 401) so playback either fails over or shows the correct age-gate UI instead of a dead-end Retry.

Note: native Swift change; needs a repro to validate (no iOS XCTest in CI).

Issue #6251, opened 2026-07-21
fix(feed): narrow 404 prune to moderation-confirmed unavailable media


hm21:
Follow-up to #6242.

Only a HEAD-confirmed **404** marks media broken + auto-advances (`lib/services/media_availability_checker.dart:32`, `lib/services/dead_media_feed_guard.dart`). **401** (age-gate) and **403** (moderation) are never pruned or auto-skipped, so an unrecoverable dead item (expired token, or a viewer who will never verify age) reappears on every refetch and must be manually skipped each time.

### Evidence
- 404-only prune/skip: `lib/services/media_availability_checker.dart:32`, `lib/widgets/video_feed_item/...` auto-advance wiring in `lib/widgets/feed/feed_videos.dart:593-600,693`
- Fullscreen removal is 404-only: `lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart:558-604`

### Proposed
Give the **unrecoverable** 401/403 cases skip/prune parity with 404 — **without** pruning content the viewer could still unlock via age-verification.

⚠️ **Trust/Safety review required** — this changes age-gate/moderation semantics, which is why it is split out of the graceful-fallback PR.

realmeylisdev:
Flagging a conflict I hit while re-investigating #5953 — treating 401 like 404 would prune content that is actually viewable.

A 401 from `media.divine.video` does not mean "unrecoverable". It means **age-restricted**, and blossom grants access as soon as the request carries a Nostr pubkey. From `blossom-core/src/types.rs` `access_for`:

```rust
BlobStatus::Banned | BlobStatus::Deleted => BlobAccess::NotFound,      // 404
BlobStatus::Restricted    => if is_owner { Allowed } else { NotFound },// 404
BlobStatus::AgeRestricted => if requester_pubkey.is_some() { Allowed } else { AgeGated }, // 401
```

So a 401 item plays fine for a signed-in viewer. `FullscreenFeedBloc` already builds a BUD-01 kind-24242 auth header for exactly this (`fullscreen_feed_bloc.dart:508-521`). If we route 401 into `markVideoBroken`, we take a video that would have played, hide it for the tracker's 7-day TTL, and hide it on every surface that consults the tracker — for a user who was entitled to watch it.

[...]

Issue #6250, opened 2026-07-21
fix(media): send BUD-01 auth so age-restricted thumbnails load


hm21:
Follow-up to #6242.

Thumbnail loaders never attach the BUD-01 (kind 24242) `Authorization` header that `media.divine.video` requires for age-restricted blobs, so age-gated thumbnails return **401** and fall back to blurhash/blank. The auth is wired only to video playback.

### Evidence
- `_SafeNetworkImage` uses a bare `NetworkImage` for `*.divine.video` hosts (no headers): `lib/widgets/video_thumbnail_widget.dart:205-263`
- `VineCachedImage` / `MediaCacheImageProvider` are constructed without `authHeaders`: `lib/widgets/vine_cached_image.dart:37-81`, `packages/media_cache/lib/src/media_cache_image_provider.dart:19-93`
- Auth infra exists but only in playback: `lib/services/media_viewer_auth_service.dart`, `lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart:503-526`, `lib/screens/feed/pooled_age_restricted_retry.dart`
- Known gap already noted in `docs/TRUST_SAFETY_SETTINGS_PLAN.md:134-137`

### Proposed
Thread the viewer's content-filter preference + `BlossomAuthService.createGetAuthHeader(sha256)` into the thumbnail loaders, gated exactly like playback (respect "never show age-restricted").

⚠️ **Trust/Safety review required** — must not weaken the age gate. Split out of the graceful-fallback PR for exactly this reason.

PR #6254, opened 2026-07-21
fix(video-feed): show blurhash behind dead-media error card

### LINKED ISSUE #6242: fix(media): investigate 401 and 404 failures from media cdn
## Summary
Users are likely seeing broken thumbnails and failed playback for media hosted on `media.divine.video`.

## Evidence
From `/Users/lizsw/openvine_full_logs_2026-07-20T14-40-35.552269.txt`:

- 95 thumbnail failures with HTTP `401`
- 90 thumbnail failures with HTTP `404`
- Player failures include:
  - `COMPOSITION_ERROR, The requested URL was not found on this server`
  - `NSURLErrorDomain error -1013`

## Expected Behavior
Media URLs emitted by events should be playable/fetchable by the app, or the app should gracefully handle expired/missing media with clear fallback UI.

## Notes
This appears to involve `media.divine.video` / CDN availability or authorization. The main user-visible impact is broken thumbnails and videos that fail to play.

### LINKED ISSUE #6255: fix(video-feed): show blurhash behind dead-media error card
Auto-created from PR #6254.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6254
**Branch:** `fix/media-dead-fallback-blurhash`

hm21:
Closes #6255

## Description

The pooled video error overlay fell back to a bare background color (`VineTheme.backgroundColor` with a `SizedBox.shrink()` thumbnail error widget) whenever a video failed and its thumbnail was missing or itself returned 401/404 from `media.divine.video`. So a dead/expired item — the exact case #6242 is about — rendered as a flat colored card with no visual context.

This renders the video's blurhash as a base layer beneath the thumbnail (content-type-derived default when the event has no blurhash), matching the feed grid's existing fallback. Expired or missing media now degrades to a soft blurred preview plus the existing error message/retry, instead of a flat color.

Restricted states still withhold frame-derived preview media: forbidden, moderation-blocked/quarantined, age-restricted, and Divine not-found media while its moderation lookup is still pending all suppress both the event blurhash and thumbnail, falling back to the generic content-type gradient only.

This also wires `BlurhashDisplay.fit` into the decoded blurhash painter so callers that pass `BoxFit.contain`/`cover` no longer hit a no-op parameter.

Closes #6242 —  (partial fix; follow-ups tracked in #6250–#6253)

## Out of Scope

This is the app-side, trust/safety-neutral slice of #6242. The remaining root causes were deliberately deferred because doing them here would change age-gate/moderation semantics or ship an unverifiable native change:

- #6250 — send BUD-01 auth so age-restricted thumbnails load *(trust/safety)*
- #6251 — auto-skip/prune unrecoverable 401/403 media like 404 *(trust/safety)*
- #6252 — map iOS NSURLErrorDomain -1013 *(native + needs repro)*
- #6253 — distinguish expired-token 401 from age-gate *(trust/safety; the current 401/unauthorized classifier still intentionally maps both to age-restricted)*

Full root-cause analysis is in the #6242 comment.

## Verification

- `flutter analyze lib/widgets/video_feed_item/pooled_video_error_overlay.dart lib/widgets/blurhash_display.dart test/widgets/pooled_video_error_overlay_test.dart` — clean
- `flutter test test/widgets/pooled_video_error_overlay_test.dart` — 25/25 pass
- Pre-push hook — clean:
  - changed-file analyzer
  - `test/widgets/blurhash_display_test.dart`
  - `test/widgets/pooled_video_error_overlay_test.dart`
  - `test/widgets/video_thumbnail_widget_test.dart`
- `mise exec -- dart format lib test integration_test` — clean

## Type of Change

### DIFF
--- a/mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
+++ b/mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
@@ -70,12 +70,12 @@ class _PooledVideoErrorOverlayState
     return widget.isSquare ? BoxFit.contain : BoxFit.cover;
   }
 
-  VineContentType? get _blurhashContentType =>
+  static VineContentType? _deriveContentType(VideoEvent video) =>
       BlurhashService.deriveContentType(
-        hashtags: widget.video.hashtags,
-        group: widget.video.group,
-        title: widget.video.title,
-        content: widget.video.content,
+        hashtags: video.hashtags,
+        group: video.group,
+        title: video.title,
+        content: video.content,
       );
 
   void _maybeAutoRetryAgeRestricted({required bool showVerifyAge}) {
@@ -179,7 +179,7 @@ class _PooledVideoErrorOverlayState
               // matching the feed grid's fallback. #6242
               BlurhashDisplay(
                 blurhash: widget.video.blurhash,
-                contentType: _blurhashContentType,
+                contentType: _deriveContentType(widget.video),
                 fit: _resolveBoxFit(),
               ),
               if (widget.video.thumbnailUrl != null &&
--- a/mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
+++ b/mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
@@ -2,6 +2,7 @@
 // ABOUTME: Differentiates moderation-restricted (403), age-gated (401),
 // ABOUTME: missing (404), and generic playback errors.
 
+import 'package:blurhash_service/blurhash_service.dart';
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
@@ -12,6 +13,7 @@ import 'package:models/models.dart' hide LogCategory;
 import 'package:openvine/blocs/video_playback_status/video_playback_status_cubit.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/services/video_moderation_status_service.dart';
+import 'package:openvine/widgets/blurhash_display.dart';
 import 'package:openvine/widgets/vine_cached_image.dart';
 
 /// Error overlay for videos playing through the pooled video player.
@@ -67,6 +69,14 @@ class _PooledVideoErrorOverlayState
     if (!widget.shouldPortraitExpand) return BoxFit.contain;
     return widget.isSquare ? BoxFit.contain : BoxFit.cover;
   }
+
+  VineContentType? get _blurhashContentType =>
+      BlurhashService.deriveContentType(
+        hashtags: widget.video.hashtags,
+        group: widget.video.group,
+        title: widget.video.title,
+        content: widget.video.content,
+      );
 
   void _maybeAutoRetryAgeRestricted({required bool showVerifyAge}) {
     if (!showVerifyAge || widget.isVerifying || widget.onVerifyAge == null) {
@@ -161,19 +171,30 @@ class _PooledVideoErrorOverlayState
       children: [
         ColoredBox(
           color: VineTheme.backgroundColor,
-          child:
-              widget.video.thumbnailUrl != null &&
-                  widget.video.thumbnailUrl!.isNotEmpty
-              ? SizedBox.expand(
-                  child: VineCachedImage(
-                    imageUrl: widget.video.thumbnailUrl!,
-                    fit: _resolveBoxFit(),
-                    fadeInDuration: Duration.zero,
-                    fadeOutDuration: Duration.zero,
-                    errorWidget: (_, _, _) => const SizedBox.shrink(),
-                  ),
-                )
-              : const SizedBox.expand(),
+          child: Stack(
+            fit: StackFit.expand,
+            children: [
+              // Blurhash preview beneath the thumbnail so expired or missing
+              // media degrades to a soft placeholder instead of a bare color,
+              // matching the feed grid's fallback. #6242
+              BlurhashDisplay(
+                blurhash: widget.video.blurhash,
+                contentType: _blurhashContentType,
+                fit: _resolveBoxFit(),
+              ),
+              if (widget.video.thumbnailUrl != null &&
+                  widget.video.thumbnailUrl!.isNotEmpty)
+                VineCachedImage(
+                  imageUrl: widget.video.thumbnailUrl!,
+                  fit: _resolveBoxFit(),
+                  fadeInDuration: Duration.zero,
+                  fadeOutDuration: Duration.zero,
+                  // On a failed/expired thumbnail, reveal the blurhash beneath
+                  // rather than collapsing to a bare color.
+                  errorWidget: (_, _, _) => const SizedBox.shrink(),
+                ),
+            ],
+          ),
         ),
         ColoredBox(
           color: VineTheme.scrim50,
--- a/mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
+++ b/mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
@@ -2,7 +2,6 @@
 // ABOUTME: Differentiates moderation-restricted (403), age-gated (401),
 // ABOUTME: missing (404), and generic playback errors.
 
-import 'package:blurhash_service/blurhash_service.dart';
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
@@ -11,6 +10,7 @@ import 'package:infinite_video_feed/infinite_video_feed.dart'
     show VideoErrorType;
 import 'package:models/models.dart' hide LogCategory;
 import 'package:openvine/blocs/video_playback_status/video_playback_status_cubit.dart';
+import 'package:openvine/extensions/video_event_content_type_extension.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/services/video_moderation_status_service.dart';
 import 'package:openvine/widgets/blurhash_display.dart';
@@ -69,14 +69,6 @@ class _PooledVideoErrorOverlayState
     if (!widget.shouldPortraitExpand) return BoxFit.contain;
     return widget.isSquare ? BoxFit.contain : BoxFit.cover;
   }
-
-  static VineContentType? _deriveContentType(VideoEvent video) =>
-      BlurhashService.deriveContentType(
-        hashtags: video.hashtags,
-        group: video.group,
-        title: video.title,
-        content: video.content,
-      );
 
   void _maybeAutoRetryAgeRestricted({required bool showVerifyAge}) {
     if (!showVerifyAge || widget.isVerifying || widget.onVerifyAge == null) {
@@ -166,6 +158,13 @@ class _PooledVideoErrorOverlayState
     final showRetry = !isModerationRestricted && !showVerifyAge;
     _maybeAutoRetryAgeRestricted(showVerifyAge: showVerifyAge);
 
+    // Hard-walled states (forbidden / moderation-blocked / age-restricted)
+    // withhold the frame entirely: suppress both the event blurhash and the
+    // thumbnail so no frame-derived preview leaks, degrading to the generic
+    // content-type gradient only. notFound / generic keep the event blurhash
+    // and thumbnail fallback. #6242
+    final suppressPreviewMedia = isModerationRestricted || isAgeRestricted;
+
     return Stack(
       fit: StackFit.expand,
       children: [
@@ -178,11 +177,11 @@ class _PooledVideoErrorOverlayState
               // media degrades to a soft placeholder instead of a bare color,
               // matching the feed grid's fallback. #6242
               BlurhashDisplay(
-                blurhash: widget.video.blurhash,
-                contentType: _deriveContentType(widget.video),
-                fit: _resolveBoxFit(),
+                blurhash: suppressPreviewMedia ? null : widget.video.blurhash,
+                contentType: widget.video.blurhashContentType,
               ),
-              if (widget.video.thumbnailUrl != null &&
+              if (!suppressPreviewMedia &&
+                  widget.video.thumbnailUrl != null &&
                   widget.video.thumbnailUrl!.isNotEmpty)
                 VineCachedImage(
                   imageUrl: widget.video.thumbnailUrl!,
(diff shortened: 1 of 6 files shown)


mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
```diff
@@ -161,19 +171,30 @@ class _PooledVideoErrorOverlayState
...
-                  ),
-                )
-              : const SizedBox.expand(),
+          child: Stack(
+            fit: StackFit.expand,
+            children: [
+              // Blurhash preview beneath the thumbnail so expired or missing
+              // media degrades to a soft placeholder instead of a bare color,
+              // matching the feed grid's fallback. #6242
+              BlurhashDisplay(
```
dcadenas, line 182: Guard this blurhash fallback for restricted classifications at mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart:180. The overlay renders widget.video.blurhash for every error type, so a forbidden, moderation-blocked, quarantined, or age-restricted video with a missing or unauthorized thumbnail can still reveal a blurred impression of its withheld frame beneath only scrim50. Pass null for those states so BlurhashDisplay uses the generic content-type fallback, while keeping the event blurhash for notFound and generic degradation, and add widget coverage for the restricted states.

mobile/test/widgets/pooled_video_error_overlay_test.dart
```diff
@@ -452,6 +454,39 @@ void main() {
       );
     });
 
+    group('dead media fallback', () {
+      testWidgets(
```
dcadenas, line 506: Drive the present-but-failing thumbnail path in mobile/test/widgets/pooled_video_error_overlay_test.dart:458. The current assertion only checks that BlurhashDisplay is mounted, but production mounts it unconditionally and the test's live thumbnail never invokes VineCachedImage.errorWidget. Stub a failing image provider, assert that the thumbnail enters its errorWidget and collapses, then assert that the blurhash remains visible.

dcadenas (dismissed):
The fallback needs a safety guard for withheld media and a test that proves the dead-thumbnail reveal path. Please address both before merging.

mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
hm21: Fair concern, but I think the guard costs more than it buys here:

- `video.blurhash` is nullable and absent for most events — it's only populated from the event's own imeta/`blurhash` tag, so non-Divine clients usually omit it, and even our own publish path drops it on a thumbnail-extract or blurhash-encode timeout. When it's absent, `BlurhashDisplay` already falls back to the generic content-type gradient (no frame info), so the frame-derived case you're describing is the exception, not the norm.
- When it *is* present, a blurhash is a handful of average colors with nothing recognizable — it's the standard sensitive-media placeholder (Mastodon renders it for CW'd media by design). And the alternative here isn't "black", it's a generic gradient — so this is frame-colors vs generic-colors, a marginal delta that only carries signal for the NSFW subset.
- Cost side: `isModerationRestricted` resolves async, so guarding on it means the frame-blurhash briefly flashes before the status lands — real added complexity for that marginal delta.

If you still want defense-in-depth for the age-gate class specifically, the cheap version is nulling the blurhash only for the synchronous states (`forbidden` / `ageRestricted` from `errorType`) — no async dependency. Happy to add that if you'd prefer to be safe; otherwise I'd lean toward keeping it as-is.

mobile/test/widgets/pooled_video_error_overlay_test.dart
hm21: Done in d016714ec. Stubbed the image cache (`debugImageCacheOverride`) so a present thumbnail fails like an expired 401/404, then assert it collapses into its `errorWidget` (`ValueKey('error')`) and the blurhash beneath stays visible. Replaced the old presence-only assertion since it couldn't fail if the reveal broke.

mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
dcadenas, line 182: Guard this blurhash for restricted errors before rendering it. `isModerationRestricted` already includes `forbidden`, and `isAgeRestricted` covers both age-gated paths, but this unconditional prop still displays the video's own event blurhash beneath the scrim when content is withheld. Derive a restricted-group bool from `isModerationRestricted || isAgeRestricted || type == VideoErrorType.forbidden`, pass `null` for that group, and keep `widget.video.blurhash` for not-found and generic errors. Add a widget test that pins the restricted suppression alongside the existing not-found coverage.

dcadenas (changes requested):
The new dead-media fallback is incomplete for restricted errors. Before merge, suppress the video's own blurhash for forbidden, moderation-restricted, and age-restricted states, and cover that behavior with a regression test while retaining blurhash fallback for not-found and generic failures.

mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
hm21: @dcadenas before I implement a blanket suppression — want to confirm this is really what you intend, because nulling the event blurhash across *all* restricted states cuts against how the nearest platforms handle sensitive media.

Quick scan (happy to pull sources):

- **Mastodon** — the same blurhash mechanism we use — renders the **real, frame-derived blurhash** behind its sensitive/CW cover. Blurhash was adopted precisely because it's abstract enough to be safe as a sensitive placeholder.
- **Meta (IG/FB)** and **X** blur the **actual** image/video behind their "sensitive content" screen — strictly *more* revealing than a blurhash.
- Neutral placeholders / opaque covers are what platforms reserve for **hard-removed or hard-gated** content, not for revealable sensitive media.

Mapped to our states:

| state | revealable? | industry norm |
|---|---|---|
| `notFound` / `generic` | n/a (just broken/expired) | real blurhash — already agreed |
| `ageRestricted` (has a Verify Age reveal) | yes | real blurhash (Mastodon-style) |
| `forbidden` / moderation-blocked (no retry) | no | neutral placeholder |

So the case for nulling the event blurhash is strongest for the **hard-wall** subset (`forbidden` / moderation-blocked). For `ageRestricted` — which is revealable via Verify Age — the closest neighbours do the opposite and show the real impression.

Concretely I'd propose: **null only for `forbidden` / moderation-blocked, keep the event blurhash for `ageRestricted` + notFound/generic.** That closes the withheld-content gap for the hard walls while matching platform norms for the revealable tier.

If you still want blanket suppression across every restricted state, just confirm and I'll implement that — only flagging because it's non-standard for the age-gate tier.

mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
```diff
@@ -161,19 +171,30 @@ class _PooledVideoErrorOverlayState
...
+                blurhash: widget.video.blurhash,
+                contentType: _deriveContentType(widget.video),
+                fit: _resolveBoxFit(),
+              ),
+              if (widget.video.thumbnailUrl != null &&
+                  widget.video.thumbnailUrl!.isNotEmpty)
+                VineCachedImage(
```
NotThatKindOfDrLiz, line 189: This is the stronger restricted-media leak. Even if the new `BlurhashDisplay` is nulled for restricted states, this thumbnail layer still renders the actual video frame for `forbidden` and moderation-blocked errors whenever the thumbnail URL resolves. For the hard-walled tier, please gate the whole media layer: no event blurhash and no thumbnail, falling back only to the generic content-type placeholder. Age-restricted can remain a separate product-policy decision because it is revealable via Verify Age and already had thumbnail exposure before this PR.

mobile/test/widgets/pooled_video_error_overlay_test.dart
```diff
@@ -452,6 +456,48 @@ void main() {
...
 
+    group('dead media fallback', () {
+      // Route the thumbnail through a stubbed cache so its load fails
+      // deterministically (like an expired 401/404) and stays off the real
+      // on-disk cache in the merged test isolate.
+      setUp(() => debugImageCacheOverride = createMockMediaCacheManager());
+      tearDown(() => debugImageCacheOverride = null);
+
+      testWidgets(
+        'reveals the blurhash when a present thumbnail fails to load (#6242)',
```
NotThatKindOfDrLiz, line 507: This test still does not prove that the event blurhash is used: `TestVideoEventBuilder.create` has no `blurhash` parameter, so the new tests only exercise the content-type/default fallback branch. Please add a builder parameter and cover both behaviors: real event blurhash remains visible for `notFound`/generic dead media, and hard-walled restricted states suppress event-derived media.

mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
```diff
@@ -68,6 +70,14 @@ class _PooledVideoErrorOverlayState
     return widget.isSquare ? BoxFit.contain : BoxFit.cover;
   }
 
+  static VineContentType? _deriveContentType(VideoEvent video) =>
```
NotThatKindOfDrLiz: This duplicates the same private helper in `video_thumbnail_widget.dart`. Since the two call sites intentionally need identical behavior, please extract a shared app-side `VideoEvent` helper/extension and consume it from both widgets instead of keeping parallel copies.

mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
```diff
@@ -161,19 +171,30 @@ class _PooledVideoErrorOverlayState
...
+          child: Stack(
+            fit: StackFit.expand,
+            children: [
+              // Blurhash preview beneath the thumbnail so expired or missing
+              // media degrades to a soft placeholder instead of a bare color,
+              // matching the feed grid's fallback. #6242
+              BlurhashDisplay(
+                blurhash: widget.video.blurhash,
+                contentType: _deriveContentType(widget.video),
+                fit: _resolveBoxFit(),
```
NotThatKindOfDrLiz: `fit` is misleading here because `BlurhashDisplay` currently accepts `fit` but does not honor it; the decoded painter stretch-fills the destination rect and the gradient fallback ignores fit too. That means letterbox bars can now show stretched blurhash even when the thumbnail uses `BoxFit.contain`. Either make `BlurhashDisplay` honor fit, or remove this argument here and keep that broader widget fix out of this PR.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The dead-media fallback is a good direction, but this needs to close the restricted-media behavior before merge: hard-walled states should not render event-derived preview media, and the tests should pin both the reveal path and the suppression path.

mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
realmeylisdev: nit: this is now duplicated verbatim in video_thumbnail_widget.dart (both define an identical private static `_deriveContentType`). worth pulling into a shared helper at some point so the two copies can't drift apart, not blocking this PR though.

mobile/lib/widgets/video_feed_item/pooled_video_error_overlay.dart
realmeylisdev: fyi, BlurhashDisplay doesn't actually consume its `fit` param anywhere (checked _BlurhashImagePainter and the gradient fallback), so this currently has no effect on rendering. video_thumbnail_widget.dart already passes it the same no-op way, so it's not a new issue from this PR, but might be worth a follow-up to wire it up or drop the param.

realmeylisdev (approved):
approved. traced the fix through BlurhashDisplay/VineCachedImage/MediaCacheImageProvider and it holds up. reverted to the pre-fix overlay locally and confirmed the two new tests genuinely fail without it, so they're real regression coverage, not just presence checks. left two small non-blocking nits inline, both pre-existing patterns rather than something new here.

rabble (dismissed):
Blocking: restricted media still renders the event-derived blurhash. The new `BlurhashDisplay` is mounted unconditionally beneath the thumbnail, so forbidden, moderation-restricted, and age-restricted states disclose withheld preview content even when the thumbnail itself is suppressed or fails.

Please gate both preview layers by the error classification: retain blurhash fallback for `notFound` and generic playback failures, but render neither the event blurhash nor thumbnail for the hard-walled restricted states. Add regression coverage that asserts suppression for forbidden/moderation/age-restricted errors and retention for not-found/generic errors. The existing dead-thumbnail reveal tests are useful and should stay.

hm21:
Pushed as `5f5b5b5e4` (rebased on latest `main`).

**Restricted-media leak (blocking).** The dead-media fallback now gates *both* preview layers on the error classification. For the hard-walled tier it renders neither the event blurhash nor the thumbnail, degrading to the generic content-type gradient only; `notFound` / `generic` keep the event blurhash + thumbnail fallback.

One decision worth calling out: I folded **age-restricted into the suppressed tier** alongside `forbidden` / moderation-blocked. That follows @rabble's ask and overrides @NotThatKindOfDrLiz's suggestion to keep age-restricted as a revealable tier — going with the more conservative default for withheld media, and the Verify-Age flow still re-fetches the real frame once the viewer is authorized.

**Tests.** Added a `blurhash` param to `TestVideoEventBuilder.create` and regression coverage pinning suppression for `forbidden` / age-restricted / moderation-blocked (asserts `BlurhashDisplay` receives a null blurhash *and* no thumbnail renders) plus retention of the real event blurhash for `notFound` dead media. Confirmed the suppression tests fail `0 -3` with the gate disabled, so they're genuine coverage rather than presence checks.

**Nits.** Extracted the duplicated `_deriveContentType` into a shared `VideoEvent.blurhashContentType` extension consumed by both the overlay and the feed thumbnail, and dropped the no-op `fit` arg on the overlay's `BlurhashDisplay` (leaving the broader "make `BlurhashDisplay` honor `fit`" out of this PR).

dcadenas (approved):
✅ dead-media previews keep their blurhash

NotThatKindOfDrLiz:
@hm21 I took over the final review fixes on this PR and pushed b65f69c46.

What changed:
- Closed the moderation-loading disclosure window: Divine not-found error overlays now fail closed for preview media while the moderation lookup is still pending, so the event blurhash and thumbnail are suppressed before the status resolves.
- Added a regression test that pumps the pre-settle loading frame and asserts both preview layers are suppressed.
- Wired BlurhashDisplay.fit into the decoded blurhash painter, resolving the no-op fit parameter raised during review.
- Updated the PR body verification/out-of-scope notes.

Verification:
- flutter test test/widgets/pooled_video_error_overlay_test.dart — 25/25 pass
- flutter analyze lib/widgets/video_feed_item/pooled_video_error_overlay.dart lib/widgets/blurhash_display.dart test/widgets/pooled_video_error_overlay_test.dart — clean
- Pre-push hook — clean, including changed-file analyzer and blurhash/overlay/thumbnail widget tests
- GitHub checks are green on b65f69c46

Please review the takeover changes and merge when you're ready.

NotThatKindOfDrLiz (approved):
Final takeover pass looks clean. The moderation-loading preview leak is closed with regression coverage, BlurhashDisplay.fit is no longer a no-op for decoded blurhash images, the PR body is current, and all GitHub checks are green on b65f69c46.

PR #6258, opened 2026-07-21
fix(video-recorder): meet 48dp min touch target on mode selector

### LINKED ISSUE #6259: fix(video-recorder): meet 48dp min touch target on mode selector
Auto-created from PR #6258.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6258
**Branch:** `fix/camer-mode-button-mode-size`

hm21:
Closes #6259

## Description

The horizontal recorder **mode selector** capped each item's tap target at the 34dp pill height, below the 48dp minimum touch target ([`accessibility.md`](.claude/rules/accessibility.md)). The wheel now takes a `kMinInteractiveDimension` (48dp) height, so each item's hit area meets the minimum. The **pill stays visually 34dp** — it's a fixed, centered `AnimatedContainer`, so only the invisible hit strip around each label grows; nothing moves on screen. Width was already fine (96dp item extent).

While in there, two related bottom-bar fixes:

- **Vertical centering.** The shorter library button (40dp) was pinned to the top of the bar because the `Stack` used its default `topStart` alignment while the button's `Align(centerLeft)` shrink-wrapped to 40dp. The `Stack` now uses `alignment: .center`, and the redundant `Align` wrapper around the wheel is removed — both the wheel and the library button are centered on the same line.
- **Library button tap.** Swapped `InkWell` → `GestureDetector(behavior: HitTestBehavior.opaque)`. The ink ripple was invisible anyway (the opaque thumbnail paints over it), and `opaque` keeps the full 40×40 tappable even during the brief window before a thumbnail loads.

**Related Issue:** 

## Out of Scope

- The library button itself is 40×40 dp (below the 48dp minimum). Pre-existing; left untouched to keep this PR focused on the mode selector. Can be a follow-up.

## Verification

- `flutter analyze lib/widgets/video_recorder test/widgets/video_recorder` — clean
- `flutter test test/widgets/video_recorder/` — 222 passing, including a new regression test asserting each mode item's tap target is ≥ `kMinInteractiveDimension`
- `dart format` — clean

## Type of Change

### DIFF
--- a/mobile/lib/widgets/video_recorder/video_recorder_bottom_bar.dart
+++ b/mobile/lib/widgets/video_recorder/video_recorder_bottom_bar.dart
@@ -29,13 +29,12 @@ class VideoRecorderBottomBar extends StatelessWidget {
         child: Padding(
           padding: const EdgeInsets.only(bottom: 4),
           child: Stack(
+            alignment: .center,
             children: [
-              Align(
-                child: VideoRecorderModeSelectorWheel(
-                  selectedMode: state.recorderMode,
-                  onModeChanged: (mode) => context
-                      .read<VideoRecorderBloc>()
-                      .add(VideoRecorderRecorderModeSet(mode)),
+              VideoRecorderModeSelectorWheel(
+                selectedMode: state.recorderMode,
+                onModeChanged: (mode) => context.read<VideoRecorderBloc>().add(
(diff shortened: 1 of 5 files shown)


dcadenas (approved):
✅ 48dp mode controls, ship it

PR #6260, opened 2026-07-21
fix(apps): enable Android image upload in Nostr app sandbox WebView

### LINKED ISSUE #6263: fix(apps): enable Android image upload in Nostr app sandbox WebView
Auto-created from PR #6260.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6260
**Branch:** `fix/sandbox-webview-android-image-upload`

hm21:
Closes #6263

## Problem: users can't create badges (Android side)

In the Divine Badges app (`badges.divine.video`, embedded via `NostrAppSandboxScreen`), creating a badge requires uploading a badge image. On Android that step was impossible: `webview_flutter` never surfaces an HTML `<input type="file">` to a picker unless a file-chooser handler is registered, so tapping "upload image" did nothing — the whole badge-creation flow was blocked on Android.

> Note: the badge upload also fails on every platform due to a separate **server-side** CORS bug on `upload.divine.video` (it rejects the `x-sha256` header in its preflight) — fixed in **divinevideo/divine-upload-server#11**. This PR is the Android-side prerequisite so that, once the server allows the header, Android users can actually pick and upload the image.

## What

Wires up Android file input in `NostrAppSandboxScreen`:

- Register an `image_picker`-backed `AndroidWebViewController.setOnShowFileSelector`. It offers the camera when the input requests capture, the gallery otherwise, supports multi-select, and returns `file://` URIs. An empty accept list and the "any file" spellings Android sends — `''` (a plain `<input type="file">` arrives as `['']`, not `[]`), `'*'`, `'*/*'` — fall through to the gallery; non-image accept types, save mode, and cancellation return an empty list ("no file selected").
- Strip selected image metadata before returning the `file://` URI to the sandbox WebView so third-party apps do not receive GPS/device EXIF.
- Enable `setAllowFileAccess(true)` on the Android controller. The picker hands back a `file://` URI, and the WebView can only read it with file access enabled — which defaults to `false` on Android 11+ (API 30). Without it the file is chosen but never populates the input (0-byte read). The https sandbox document still can't script-read local file contents: same-origin policy blocks fetch/XHR to `file://`, and canvas tainting blocks reading back `` pixels.

iOS already handles `<input type="file">` natively via WKWebView, so this is Android-only.

## Merge order / dependency

End-to-end badge upload stays blocked until the server-side CORS fix (divine-upload-server#11) ships. This PR makes Android file *selection* work and can merge independently; the full flow only lights up once the server side lands.

## Test plan

- `flutter test test/screens/apps/nostr_app_sandbox_screen_test.dart` — 59 passing, incl. `sandboxAndroidFileSelector` cases for gallery, camera-capture, cancel, multi-select, metadata stripping, non-image type ignored, save mode ignored, empty-list + `''`/`'*'`/`'*/*'` accept-anything, additional image extension tokens, and picker throws.
- `flutter analyze` clean.

The Android controller wiring (`setOnShowFileSelector` / `setAllowFileAccess`) can't be exercised in a widget test — the test harness's mocked `WebViewPlatform` isn't a concrete `AndroidWebViewController` — so it was verified manually on a device:

- **Gallery upload** — open the badges app → tap image upload → gallery picker opens → selected image populates the `<input>` (non-zero read).
- **Camera capture** — an input with `capture` opens the camera; captured photo populates the input. Denying the camera permission degrades gracefully (empty selection, warning log).
- **Security (file-access scope)** — from an allowed https sandbox app, confirm a scripted `fetch('file:///…')` / same-page `file://` read still fails; only the user-chosen upload is readable.
- End-to-end upload still blocked by the server CORS issue above until divine-upload-server#11 ships.

### DIFF
--- a/mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
+++ b/mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
@@ -12,6 +12,7 @@ import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
 import 'package:http/http.dart' as http;
+import 'package:image_picker/image_picker.dart';
 import 'package:nostr_app_bridge_repository/nostr_app_bridge_repository.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/providers/app_providers.dart';
@@ -39,6 +40,11 @@ class NostrAppSandboxScreen extends ConsumerStatefulWidget {
   static const routeName = 'nostr-app-sandbox';
   static const path = '/apps/:appId/sandbox';
   static const bridgeChannelName = 'divineSandboxBridge';
+
+  /// Default [ImagePicker] backing the Android `<input type="file">` picker.
+  /// Exposed so tests can swap in a mock; production reads it as-is.
+  @visibleForTesting
+  static ImagePicker imagePicker = ImagePicker();
 
   const NostrAppSandboxScreen({
     required this.app,
@@ -273,6 +279,25 @@ class _NostrAppSandboxScreenState extends ConsumerState<NostrAppSandboxScreen> {
         }
       },
     );
+
+    // Without this, `webview_flutter` silently ignores HTML file inputs on
+    // Android, so image uploads inside sandboxed apps (e.g. badges) do nothing.
+    final platformController = controller.platform;
+    if (platformController is AndroidWebViewController) {
+      await platformController.setOnShowFileSelector(
+        (params) => sandboxAndroidFileSelector(
+          params,
+          NostrAppSandboxScreen.imagePicker,
+        ),
+      );
+      // The picker hands back a file:// URI; the WebView can only read it with
+      // file access enabled, which defaults to false on Android 11+. Without
+      // this the file is chosen but never populates the <input> (0-byte read).
+      // Cross-origin reads from file:// stay disabled (their own settings
+      // default to false), so an allowed https app can't script-read local
+      // files — only the user-chosen upload is readable.
+      await platformController.setAllowFileAccess(true);
+    }
 
     // Replace the pigeon message channel with the native frame-attesting one
     // (iOS: WKScriptMessageHandler reading frameInfo; Android: WebMessageListener
@@ -740,4 +765,56 @@ class _BootstrappedSandboxPage {
 
   final Uri baseUri;
   final String html;
+}
+
+/// Resolves an Android WebView file-chooser request into `file://` URIs the
+/// WebView can read back into the HTML `<input type="file">`.
+///
+/// Sandboxed apps only ever need images (e.g. badge artwork), so this offers
+/// the native image picker — camera when the input requests capture, gallery
+/// otherwise. Returns an empty list when the input wants a non-image type or
+/// the user cancels, which the WebView treats as "no file selected".
+@visibleForTesting
+Future<List<String>> sandboxAndroidFileSelector(
+  FileSelectorParams params,
+  ImagePicker picker,
+) async {
+  final acceptsImage =
+      params.acceptTypes.isEmpty || params.acceptTypes.any(_acceptsImageType);
+  if (!acceptsImage) {
+    return const <String>[];
+  }
+
+  try {
+    if (params.mode == FileSelectorMode.openMultiple) {
+      final files = await picker.pickMultiImage();
+      return files.map((file) => Uri.file(file.path).toString()).toList();
+    }
+
+    final source = params.isCaptureEnabled
+        ? ImageSource.camera
+        : ImageSource.gallery;
+    final file = await picker.pickImage(source: source);
+    if (file == null) {
+      return const <String>[];
+    }
+    return <String>[Uri.file(file.path).toString()];
+  } catch (error, stackTrace) {
+    Log.error(
+      'Sandbox WebView file selection failed: $error',
+      name: 'NostrAppSandboxScreen',
+      category: LogCategory.ui,
+      error: error,
+      stackTrace: stackTrace,
+    );
+    return const <String>[];
+  }
+}
+
+bool _acceptsImageType(String type) {
+  final normalized = type.trim().toLowerCase();
+  return normalized.startsWith('image/') ||
+      const {'.jpg', '.jpeg', '.png', '.gif', '.webp', '.heic'}.contains(
+        normalized,
+      );
 }
--- a/mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
+++ b/mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
@@ -3,6 +3,7 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 import 'dart:math';
 
 import 'package:divine_ui/divine_ui.dart';
@@ -12,6 +13,7 @@ import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
 import 'package:http/http.dart' as http;
+import 'package:image_metadata_stripper/image_metadata_stripper.dart';
 import 'package:image_picker/image_picker.dart';
 import 'package:nostr_app_bridge_repository/nostr_app_bridge_repository.dart';
 import 'package:openvine/l10n/l10n.dart';
@@ -26,6 +28,7 @@ import 'package:webview_flutter_wkwebview/webview_flutter_wkwebview.dart';
 typedef SandboxViewBuilder =
     Widget Function(void Function(Uri uri) onNavigationAttempt);
 typedef SandboxJavaScriptRunner = Future<void> Function(String script);
+typedef SandboxImageMetadataStripper = Future<File> Function(File imageFile);
 
 const _bridgePayloadObjectMessage = 'Bridge payload must be a JSON object';
 const _bridgeMethodRequiredMessage = 'Bridge method is required';
@@ -765,23 +768,40 @@ class _BootstrappedSandboxPage {
 /// Resolves an Android WebView file-chooser request into `file://` URIs the
 /// WebView can read back into the HTML `<input type="file">`.
 ///
-/// Sandboxed apps only ever need images (e.g. badge artwork), so this offers
-/// the native image picker — camera when the input requests capture, gallery
-/// otherwise. Returns an empty list when the input wants a non-image type or
-/// the user cancels, which the WebView treats as "no file selected".
+/// This intentionally narrows Android sandbox uploads to images for now:
+/// camera when the input requests capture, gallery otherwise. Picked files are
+/// stripped of metadata before being exposed to the sandbox so third-party apps
+/// do not receive GPS or device EXIF. Returns an empty list when the input wants
+/// a non-image type, save mode, or the user cancels, which the WebView treats as
+/// "no file selected".
 @visibleForTesting
 Future<List<String>> sandboxAndroidFileSelector(
   FileSelectorParams params,
-  ImagePicker picker,
-) async {
+  ImagePicker picker, {
+  SandboxImageMetadataStripper? metadataStripper,
+}) async {
   if (!_acceptsImages(params.acceptTypes)) {
     return const <String>[];
   }
 
+  if (params.mode == FileSelectorMode.save) {
+    return const <String>[];
+  }
+
+  final stripMetadata =
+      metadataStripper ?? ImageMetadataStripper.stripMetadataInPlace;
+
   try {
     if (params.mode == FileSelectorMode.openMultiple) {
+      if (params.isCaptureEnabled) {
+        final file = await picker.pickImage(source: ImageSource.camera);
+        if (file == null) {
+          return const <String>[];
+        }
+        return _strippedFileUris([file], stripMetadata);
+      }
       final files = await picker.pickMultiImage();
-      return files.map((file) => Uri.file(file.path).toString()).toList();
+      return _strippedFileUris(files, stripMetadata);
     }
 
     final source = params.isCaptureEnabled
@@ -791,7 +811,7 @@ Future<List<String>> sandboxAndroidFileSelector(
     if (file == null) {
       return const <String>[];
     }
-    return <String>[Uri.file(file.path).toString()];
+    return _strippedFileUris([file], stripMetadata);
   } catch (error, stackTrace) {
     // Expected, user-driven paths land here too (e.g. denying the camera
     // permission throws PlatformException), so log at warning, not error.
@@ -805,6 +825,18 @@ Future<List<String>> sandboxAndroidFileSelector(
     );
     return const <String>[];
   }
+}
+
+Future<List<String>> _strippedFileUris(
+  List<XFile> files,
+  SandboxImageMetadataStripper stripMetadata,
+) async {
+  final strippedFiles = <String>[];
+  for (final file in files) {
+    final strippedFile = await stripMetadata(File(file.path));
+    strippedFiles.add(Uri.file(strippedFile.path).toString());
+  }
+  return strippedFiles;
 }
 
 /// Whether the `<input accept="…">` tokens permit an image.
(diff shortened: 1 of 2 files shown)


mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
```diff
@@ -741,3 +766,55 @@ class _BootstrappedSandboxPage {
...
+/// the native image picker — camera when the input requests capture, gallery
+/// otherwise. Returns an empty list when the input wants a non-image type or
+/// the user cancels, which the WebView treats as "no file selected".
+@visibleForTesting
+Future<List<String>> sandboxAndroidFileSelector(
+  FileSelectorParams params,
+  ImagePicker picker,
+) async {
+  final acceptsImage =
+      params.acceptTypes.isEmpty || params.acceptTypes.any(_acceptsImageType);
```
realmeylisdev: `accept="*/*"` — or a bare `*`, or an extension-form token the set doesn't list (`.heif`, `.bmp`, `.avif`) — makes `acceptsImage` false, so the picker never opens and the upload silently no-ops. But a *missing* accept attribute (empty list) falls through to the gallery and works. So two ways of saying "accept anything" behave oppositely. Badges is unaffected (`image/*`), but treating `*/*`/`*` like the empty case would remove the footgun.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
```diff
@@ -274,6 +280,25 @@ class _NostrAppSandboxScreenState extends ConsumerState<NostrAppSandboxScreen> {
...
+          NostrAppSandboxScreen.imagePicker,
+        ),
+      );
+      // The picker hands back a file:// URI; the WebView can only read it with
+      // file access enabled, which defaults to false on Android 11+. Without
+      // this the file is chosen but never populates the <input> (0-byte read).
+      // Cross-origin reads from file:// stay disabled (their own settings
+      // default to false), so an allowed https app can't script-read local
+      // files — only the user-chosen upload is readable.
+      await platformController.setAllowFileAccess(true);
```
realmeylisdev, line 297: Conclusion is right — an https sandbox page can't script-read local file *contents* here. One precision nit on the *why* in the comment above: `setAllowFileAccessFromFileURLs` / `setAllowUniversalAccessFromFileURLs` govern what a `file://`-origin document can reach; they don't gate the https sandbox document (that's blocked by same-origin policy + canvas tainting). Also worth noting `setAllowFileAccess` is global — it can't be scoped to the single picked file, so it stays on for every sandbox app and navigation, and subresource loads (``) bypass the main-frame nav allowlist. If you'd rather not enable it at all, handing the WebView a `content://` URI via a FileProvider grants read to just the one picked file. Optional.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
```diff
@@ -40,6 +41,11 @@ class NostrAppSandboxScreen extends ConsumerStatefulWidget {
   static const path = '/apps/:appId/sandbox';
   static const bridgeChannelName = 'divineSandboxBridge';
 
+  /// Default [ImagePicker] backing the Android `<input type="file">` picker.
+  /// Exposed so tests can swap in a mock; production reads it as-is.
+  @visibleForTesting
+  static ImagePicker imagePicker = ImagePicker();
```
realmeylisdev: This is doc'd "so tests can swap in a mock," but no test swaps it — all 7 cases pass a picker straight into `sandboxAndroidFileSelector(params, picker)`, and the production wiring that reads this static isn't widget-testable. So the seam is currently unused (cf. `ImageAttachmentPicker.imagePicker`, which its test does swap + reset in tearDown). Either soften the comment / drop the static, or add a test that actually exercises it with a tearDown reset so it can't bleed across the merged VGV isolate.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
```diff
@@ -741,3 +766,55 @@ class _BootstrappedSandboxPage {
...
+    final source = params.isCaptureEnabled
+        ? ImageSource.camera
+        : ImageSource.gallery;
+    final file = await picker.pickImage(source: source);
+    if (file == null) {
+      return const <String>[];
+    }
+    return <String>[Uri.file(file.path).toString()];
+  } catch (error, stackTrace) {
+    Log.error(
```
realmeylisdev: A user denying camera permission throws `PlatformException` and lands here — an expected, user-driven path logged at error severity; `Log.warning` fits the error-handling matrix better. And since every failure returns `[]`, a genuine error is indistinguishable from a cancel with no user signal — a light snackbar on the exception branch would make real failures visible.

realmeylisdev (commented):
Tight, well-documented Android-only fix, 49 green. I checked the security angle end to end: `setAllowFileAccess(true)` does **not** let an https sandbox page script-read local file contents — same-origin policy + canvas tainting block it, and the `file://`-URL-access settings stay at their safe defaults (never touched). No blockers.

A few non-blocking notes inline; the one worth a look is `accept="*/*"` silently opening no picker (badges itself is fine — it uses `image/*`).

NotThatKindOfDrLiz (changes requested):
Requesting changes based on the Android file-selector behavior.

Blocking:
- `sandboxAndroidFileSelector` only treats `acceptTypes.isEmpty` as the no-accept fallback, but Android WebView can deliver a plain `<input type="file">` as `acceptTypes == [""]`. That makes the most generic file input return `[]` and silently no-op. Please normalize accept tokens and treat `""`, `"*"`, and `"*/*"` as accept-anything.
- The current “no accept type” test uses `acceptTypes: const []`, so it does not cover the real Android shape. Please add coverage for `const [""]` while keeping an empty-list case if useful.
- The new static `NostrAppSandboxScreen.imagePicker` is documented as test-swappable but no test swaps/resets it. Please either remove that mutable static and instantiate at the wiring site, or add a test that uses it with a tearDown reset.

Also worth addressing in this pass:
- Widen extension handling for image tokens such as `.heif`, `.bmp`, `.avif`, `.tif`.
- Correct the `setAllowFileAccess(true)` comment: the `file://` cross-origin settings discussed there govern `file://` origins, not the https sandbox document, and `setAllowFileAccess` is a global WebView setting.
- Downgrade expected picker failures from `Log.error` to `Log.warning`.
- Rebase onto fresh `origin/main` before final handoff; the branch is currently behind, and the PR body still says “Why this is a draft” while the PR is not a draft.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
hm21: Confirmed the behavior, but leaving as-is. `.avif`/`.heif`/`.bmp` not opening the picker is intentional — sandbox apps only get the formats in the allow-set. And `*/*`/`*` is hypothetical here: the only consumer (badges) uses `image/*`. Forcing an image-only picker when an app asked for *any* file is itself debatable, so the conservative no-op stands rather than guessing "image" for a non-existent non-image consumer.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
hm21: Good precision catch — corrected the comment in c7c6992026: it's same-origin policy + canvas tainting that block the https doc, not the `setAllow*FileURLs` knobs (those govern `file://`-origin docs, which we never load). Kept `setAllowFileAccess(true)`: with it global, the only residual is a file-*existence* oracle via `` (load succeeds/fails) — never contents, canvas stays tainted. The `content://`/FileProvider swap isn't worth the complexity for an existence oracle. Conclusion (no content read) unchanged.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
hm21: Agreed — dropped the static entirely and inlined a local `ImagePicker()` in c7c6992026, rather than adding a test for a seam nothing uses. Removes the merged-isolate bleed risk too.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
hm21: Switched to `Log.log(..., level: LogLevel.warning, ...)` in c7c6992026 — keeps the error/stackTrace detail but drops the severity for the expected denial/cancel path. Skipping the snackbar: this is a top-level function with no `BuildContext`, and plumbing a messenger through for a rare throw isn't worth it — the graceful empty-return plus warning log is enough.

hm21:
@realmeylisdev thanks — all four addressed in c7c6992026. Applied: dropped the unused `imagePicker` static (inlined a local), corrected the `setAllowFileAccess` security comment (SOP + canvas tainting, not the file-URL knobs), and lowered the picker-catch log to `warning`. Declined: `*/*`/`.avif` (non-image formats not opening is intentional; only consumer uses `image/*`) and the snackbar (context-less top-level fn). 49 tests green.

hm21:
@NotThatKindOfDrLiz thanks — addressed. Note two of your three blockers plus two "also worth" items were already fixed in `9b22a24` (pushed just before your review landed, so you reviewed the older commit).

**Blocking**
1. **Accept-token normalization** — fixed in `0856db6`. `''`, `'*'`, `'*/*'` now fall through to the picker like an empty list. You're right that Android delivers a plain `<input type="file">` as `['']`, not `[]` — good catch, this was the one genuinely new item.
2. **`['']` test coverage** — added, plus `['*']` and `['*/*']`; kept the empty-list case. Selector suite is now 10 cases (52 total green).
3. **Static `imagePicker`** — already dropped in `9b22a24`: removed the mutable static and instantiate a local `ImagePicker()` at the wiring site.

**Also worth addressing**
- **Widen `.heif`/`.bmp`/`.avif`/`.tif`** — declined by design; the sandbox format allow-set is intentionally narrow (product decision). Note this is mostly moot in practice: MIME-form tokens (`image/avif`, `image/heif`, …) already pass via `startsWith('image/')`; only the bare-extension form is excluded, which is the deliberate part.
- **`setAllowFileAccess` comment** — already corrected in `9b22a24` (SOP + canvas tainting gate the https doc, not the `setAllow*FileURLs` knobs; noted it's a global setting).
- **`Log.error` → `Log.warning`** — already done in `9b22a24` (via `Log.log(level: warning)`, keeping error/stackTrace).
- **Rebase + draft body** — rebased onto fresh `origin/main` (0 behind now); updated the PR body — dropped the stale "Why this is a draft" section (it's no longer a draft) and reframed the server dependency as a merge-order note.

Re-requesting review.

dcadenas (approved):
✅ Android image uploads are ready to ship

realmeylisdev (commented):
Re-checked the two commits since my last pass: my four items and Liz's three blockers are all genuinely fixed in the code (not just claimed), 52 green, `flutter analyze` clean, rebase done.

Two new things came out of reading wider than the badges app — the "apps only ever need images" premise doesn't survive the 16-app registry, and the picked photo keeps its GPS EXIF where the first-party upload path strips it. Both inline; neither blocks.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
```diff
@@ -741,3 +761,77 @@ class _BootstrappedSandboxPage {
   final Uri baseUri;
   final String html;
 }
+
+/// Resolves an Android WebView file-chooser request into `file://` URIs the
+/// WebView can read back into the HTML `<input type="file">`.
+///
+/// Sandboxed apps only ever need images (e.g. badge artwork), so this offers
```
realmeylisdev: This premise doesn't hold once you look past badges. `preloadedNostrApps` ships **16** sandbox apps — Flotilla, Habla, zap.stream, Primal, YakiHonne, Shopstr, Nostr Nests, ditto.pub, Agora, Treasures, Blobbi, Espy, Jumble, divine.space, Badges, Verifier — and `resolved_sandbox_route_screen.dart:64` routes every one of them through this single screen. The directory is also remote-driven (`NostrAppDirectoryService` fetches from `apps.divine.video`), so the consumer set isn't even fixed at review time.

Two concrete outcomes for a general-purpose client's media input:

- `accept="video/*"` → `_acceptsImages` is false → `[]`, i.e. exactly the silent no-op this PR exists to fix.
- `accept="image/*,video/*"` → `.any` matches on `image/*` → an **image-only** picker opens, so the video branch is unreachable for the user.

Badges itself is fine — `assets/app/views/new.js:171` and `edit_badge.js:206` are both `accept="image/*"`, so your earlier decline was right *for badges*. I'm not reopening the format allow-set. The ask is just that the comment stop asserting a fact about all sandbox apps that isn't true: say images-only is a deliberate narrowing across a 16-app remote-driven surface, so the next person doesn't inherit the premise.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
```diff
@@ -741,3 +761,77 @@ class _BootstrappedSandboxPage {
...
+  try {
+    if (params.mode == FileSelectorMode.openMultiple) {
+      final files = await picker.pickMultiImage();
+      return files.map((file) => Uri.file(file.path).toString()).toList();
+    }
+
+    final source = params.isCaptureEnabled
+        ? ImageSource.camera
+        : ImageSource.gallery;
+    final file = await picker.pickImage(source: source);
```
realmeylisdev, line 810: This call passes no `maxWidth` / `maxHeight` / `imageQuality`, so `image_picker_android`'s `ImageResizer.resizeImageIfNeeded` evaluates `shouldScale = maxWidth != null || maxHeight != null || imageQuality < 100` → `false` and hands back the **original** file untouched. (Even on the scaling path it calls `copyExif()`, which carries 105 tags across, GPS included.)

So the `file://` URI given to the page is the user's original photo with GPS and device EXIF intact, and a third-party sandbox app uploads it to its own server.

That's the inverse of the first-party path: `blossom_upload_service.dart:2527` runs every image through `packages/image_metadata_stripper` ("Strips EXIF metadata (GPS, device info, timestamps)") before upload. Since the sandbox hands files to third parties rather than to our own backend, it's the path that needs stripping *more*, not less.

Routing the picked file through `ImageMetadataStripper` before building the URI would line the two paths up. If that's out of scope here, worth an explicit note saying the sandbox is knowingly exempt.

mobile/lib/screens/apps/nostr_app_sandbox_screen.dart
```diff
@@ -741,3 +761,77 @@ class _BootstrappedSandboxPage {
...
+Future<List<String>> sandboxAndroidFileSelector(
+  FileSelectorParams params,
+  ImagePicker picker,
+) async {
+  if (!_acceptsImages(params.acceptTypes)) {
+    return const <String>[];
+  }
+
+  try {
+    if (params.mode == FileSelectorMode.openMultiple) {
```
realmeylisdev, line 795: Two small gaps in the mode/capture handling, neither worth blocking on:

- `FileSelectorMode` has a third value, `save` (`android_webview_controller.dart:950-960`). It falls through to the single-open branch, so a save request gets a gallery picker handing back an existing file. Rare from `<input type="file">`, but it's currently unhandled rather than deliberately declined.
- `openMultiple` ignores `isCaptureEnabled`, so `<input type="file" multiple capture>` lands on the gallery instead of the camera.

A line in the doc comment marking both as intentional would close this out.

NotThatKindOfDrLiz:
Takeover update pushed in `29bf343e7`.

What changed:
- Selected Android sandbox images now run through `ImageMetadataStripper.stripMetadataInPlace` before the `file://` URI is returned to the WebView, matching Divine's existing first-party upload privacy invariant.
- The selector now deliberately declines `FileSelectorMode.save` instead of falling through to gallery selection.
- `multiple capture` uses the camera path rather than silently opening the gallery.
- Common image extension accept tokens now include `.avif`, `.bmp`, `.heif`, `.tif`, and `.tiff`.
- The selector comment now states that Android sandbox uploads are intentionally image-only for this PR, rather than asserting every sandbox app only needs images.

Verification:
- `flutter test test/screens/apps/nostr_app_sandbox_screen_test.dart` — 59 passing.
- `flutter analyze` — clean.
- Pre-push hook reran analyzer and the changed sandbox test successfully.

CI is running on the updated branch now.

NotThatKindOfDrLiz (approved):
Takeover pass is clean. The Android sandbox image path now strips metadata before exposing the file URI, the mode/accept edge cases are covered, the PR body is updated, and CI is green on `29bf343e7`.

PR #6264, opened 2026-07-21
fix(router): make every route-error screen escapable (safePop + home button)

### LINKED ISSUE #6262: Bug: route-error screens can strand the user with no way out (RouteErrorScreen)
## Summary

Multiple users reported getting stuck on a full-screen **"Error / No videos to display"** page after returning to the app from the background. The screen has no back button and no other affordance, so there is no way out except force-quitting the app.

More broadly, the shared `RouteErrorScreen` (used by ~24 route-validation and not-found paths) is not guaranteed to be escapable, and one of its back-button configurations can crash.

## Steps to Reproduce

1. Open the app to the feed.
2. Send the app to the background.
3. Reopen the app.
4. Tap a video to open it.

## Actual Result

A full-screen page titled **"Error"** with the body **"No videos to display"** and no back button. The user is stranded and must force-quit the app to recover.

## Expected Result

Every route-error screen should always offer a way back into the app — at minimum, a way to return to the home feed. No error state should be a dead end.

## Root Cause

On iOS, sending the app to the background can discard the in-memory GoRouter `extra` that carries the pooled fullscreen feed's data (lifecycle restoration). When the user then opens a video, the route builder can't resolve any video and falls through to `RouteErrorScreen(message: routeNoVideosToDisplay)`.

`RouteErrorScreen` has two escapability gaps:

1. **`showBackButton: false` (the default for most callers)** — no back affordance at all. If the screen is reached via `go` / a deep link, there is no way out.
2. **`showBackButton: true` without an `onBackPressed`** — the app-bar back control calls a plain pop, which on a `go`-navigated route with an empty back stack throws `GoError: There is nothing to pop` (and/or pops to a black screen). Two list routes use this configuration today.

## Scope / Affected screens

`RouteErrorScreen` backs ~24 route builders (invalid video / user / list / sound / hashtag / conversation IDs, unknown paths, "no videos to display", etc.). They all share the same escapability gap.

Note: the data-recovery for the specific pooled fullscreen feed instance — recovering the selected video from the URL when lifecycle restoration loses the in-memory args — and a back button on that one screen were already addressed in #6199. This issue tracks making **every** `RouteErrorScreen` escapable and removing the crash in the back-button path.

## Proposed Fix

- Default the app-bar back control to a crash-safe pop that falls back to the home feed when the back stack is empty (`context.safePop()`).
- Add an always-visible "go home" action under the error message so every error screen — including the `showBackButton: false` ones — has a guaranteed way out.

## Environment

- Platform: iOS (reported). The escapability gap itself is platform-agnostic.

realmeylisdev:
## Description

`RouteErrorScreen` — the shared full-screen route-error / not-found state used by ~24 route builders — could strand the user with no way out:

- Callers using the default `showBackButton: false` had no back affordance at all.
- Callers using `showBackButton: true` without an `onBackPressed` triggered a plain pop, which throws `GoError: There is nothing to pop` on a `go`-navigated route with an empty back stack (two list routes hit this today).

This surfaced as the reported "Error / No videos to display" screen with no back button after resuming the app from the background.

Changes:

- Default the app-bar back control to `context.safePop()`, which pops when possible and otherwise falls back to the home feed instead of throwing.
- Add an always-visible go-home action (`DivineButton`) under the error message, gated by a new `showHomeButton` flag (defaults `true`), so every error screen has a guaranteed way out — including the `showBackButton: false` ones.
- Add the `routeGoHome` string across all 18 locales.

Complements #6199, which recovered the pooled fullscreen feed's data and added a back button to that one screen; this generalizes escapability to every `RouteErrorScreen` and removes the crash in the back-button path.

**Related Issue:** Closes #6262

## Out of Scope

The gift-wrap subscription-replay log volume on resume (a separate, cosmetic logging concern) is not addressed here.

## Verification

- `flutter analyze lib/router lib/l10n test/router` — no issues
- `flutter test test/router/route_error_screen_test.dart` — 6 new tests (message renders; home button navigates to the feed; `showHomeButton: false` hides it; no back button by default; back → home on an empty stack; back → pop when there is a route to pop to)
- `flutter test test/l10n/arb_consistency_test.dart` — passes (all locales define `routeGoHome`)
- `flutter test test/screens/curated_list_by_author_screen_test.dart test/screens/subtitle_editor/subtitle_editor_screen_test.dart test/router/route_error_routing_test.dart` — existing caller tests pass
- Pre-commit hook: format + analyzer

## Type of Change

### DIFF
--- a/mobile/lib/router/route_error_screen.dart
+++ b/mobile/lib/router/route_error_screen.dart
@@ -4,18 +4,29 @@
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
+import 'package:openvine/extensions/safe_pop_extension.dart';
 import 'package:openvine/l10n/l10n.dart';
+import 'package:openvine/screens/feed/video_feed_page.dart';
 
-/// Full-screen error state for bad route parameters and (later) unmatched paths.
+/// Full-screen error state for bad route parameters and unmatched paths.
 ///
-/// Matches the previous inline pattern in [GoRoute] builders: [DiVineAppBar] with
-/// [AppLocalizations.routeErrorTitle] and a short [message] in the body.
+/// Matches the previous inline pattern in [GoRoute] builders: [DiVineAppBar]
+/// with [AppLocalizations.routeErrorTitle] and a short [message] in the body.
+///
+/// Every instance is escapable, even when reached via `go` / a deep link that
+/// leaves nothing to pop back to:
+/// - the app-bar back control degrades to [BuildContext.safePop] (home when the
+///   back stack is empty) instead of throwing `GoError: There is nothing to
+///   pop`, and
+/// - a [showHomeButton] action under [message] always routes to the home feed,
+///   so a screen with [showBackButton] `false` still has a way out.
 class RouteErrorScreen extends StatelessWidget {
   const RouteErrorScreen({
     required this.message,
     this.title,
     this.showBackButton = false,
     this.onBackPressed,
+    this.showHomeButton = true,
     super.key,
   });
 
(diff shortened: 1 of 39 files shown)


mobile/lib/router/route_error_screen.dart
```diff
@@ -44,7 +60,21 @@ class RouteErrorScreen extends StatelessWidget {
...
         onBackPressed: backHandler,
       ),
-      body: Center(child: Text(message)),
+      body: Center(
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            Text(message),
+            if (showHomeButton) ...[
+              const SizedBox(height: 24),
```
hm21: nit: use `Column(spacing: 24, children: [Text(message), if (showHomeButton) DivineButton(...)])` and drop the `SizedBox` — matches the repo convention of `spacing:` over manual gaps. With `showHomeButton: false` there is a single child, so the spacing is a no-op.

hm21 (approved):
LGTM left only one nit

mobile/lib/router/route_error_screen.dart
realmeylisdev: Done in 9e5c4496a — switched to `Column(spacing: 24, …)` and dropped the `SizedBox`. Single child when `showHomeButton: false`, so the spacing is a no-op.

PR #6256, opened 2026-07-21
chore(feature-flags): remove unused feature flags

hm21:
Closes #6257

## What

Removes five feature flags that had no remaining call sites in `lib/`:
`newCameraUI`, `newProfileLayout`, `routerDrivenHome`,
`enableVideoEditorV1`, `classicsHashtags`.

## Why

These flags were dead — nothing in `lib/` read them anymore. Removing the
enum values surfaced dangling references that broke `flutter analyze` and
the test suite:

- The flag test files used `newCameraUI` as their generic example flag
  (79 references across six files). Renamed consistently to
  `enhancedAnalytics`, which shares the same `false` build default, so the
  defaults-based assertions stay valid.
- The feed edit button no longer needs a flag gate — it shows for owned,
  non-preview videos — so its `featureFlagServiceProvider` lookup and the
  two now-orphaned imports in `video_feed_item.dart` are dropped. Its
  gating comment is refreshed to match (no more flag reference).
- The experimental-features settings tile was **duplicated** in both the
  Nostr settings screen and the general settings screen. The redundant
  Nostr copy is removed (dropping its now-unused `FeatureFlagScreen`
  import); the general-settings entry point stays. The
  `nostr_settings_screen` widget test that asserted the removed tile is
  dropped to match.

## Testing

- `flutter analyze` — clean
- Feature-flag suite + touched tests (incl. `nostr_settings_screen_test`):
  all pass

## Optional follow-ups (not in this PR)

- l10n key `nostrSettingsExperimentalFeaturesSubtitle` is now unused (only
  in the ARB + generated files). Left in place — removing it touches 17
  `app_*.arb` files.
- The skipped `video_editor_route_test.dart` "Edit Button Tests" group is
  now semantically stale (the edit button is no longer flag-gated). Left
  as the pre-existing skip.

PR #6247, opened 2026-07-21
fix(auth): read keycast login error code and nudge toward sign-in options

realmeylisdev:
Closes #3158

## TL;DR

Two things ship together on this one screen:

1. **A real, verified bug (the fix):** a failed email/password sign-in showed the hardcoded English literal **"Login failed"** in an 18-locale app. keycast returns `{"code":"INVALID_CREDENTIALS","error":"Invalid email or password"}`, but the client read the fields swapped and **never read `code`**, so the machine code was discarded and the user saw a client-side fallback string.
2. **The nudge #3158 asked for**, redirected to where it's actually useful (see *Premise* below).

## Context — the issue's premise was retracted

#3158 was auto-created from the closed draft PR #3157, whose justification ("~97k pubkey-only users hit this") **was retracted by its own author**: those rows are legacy **Vine import** records, not Nostr-signer signups. This was corroborated in keycast's source — `create_preloaded_user` is documented *"for importing users from external systems (e.g., Vine)"*, all pubkey-only rows are admin-created, and a user who signs in with a Nostr key (nsec / NIP-46 / Amber / NIP-07) has **no keycast account at all**. So the Nostr-import buttons the original hint pointed at can't help the real pubkey-only population, and are wrong for the one overlap cohort (users who ran *Secure Account* — they should type their email).

The nudge therefore points at the **existing sign-in-options sheet** (the ⓘ explainer already on the screen), which is true and safe for *every* user and never asserts an account exists — matching keycast's non-enumerating 401.

## What changed

**Client (`keycast_flutter`)**
- `HeadlessLoginResult` now exposes `errorCode` (reads `json['code']`) and a typed `KeycastLoginFailure`; the field-role swap is fixed. `code` stays the OAuth code on success, `errorCode` the machine code on failure.

**State / BLoC**
- Sign-in failures emit a `SignInFailureReason` **enum** on `DivineAuthState` — no error strings in state (per `state_management.md`).

**UI**
- The error renders localized copy via `context.l10n`, distinct for `INVALID_CREDENTIALS` / `EMAIL_NOT_VERIFIED` / `INVALID_EMAIL` / network / generic.
- A tappable `_SignInOptionsHint` (whole-line 48 dp target, `VineTheme.*Font()`) under the error opens the existing options sheet; the error is announced to screen readers via `SemanticsService.sendAnnouncement`.

**l10n**
- 7 new keys, translated across **all 18 locales**.

## Verified

[...]

PR #6216, opened 2026-07-20
feat(theme): add feature-flagged light appearance

rabble:
Closes #6217

## Summary
- add Light Mode and Adaptive Media Chrome feature flags, defaulting off
- add persisted System/Light/Dark appearance settings and app-shell ThemeMode wiring
- add warm off-white light palette, semantic theme extension, settings route, and adaptive playback controls
- preserve dark-theme switch/bottom-nav behavior while adding light theme support
- add appearance/settings/route/adaptive-media coverage and remove the stale unchecked implementation plan

## Verification
- flutter analyze
- flutter test test/features/appearance/appearance_cubit_test.dart test/features/appearance/appearance_repository_test.dart test/screens/settings/appearance_settings_screen_test.dart test/screens/settings/settings_taxonomy_test.dart test/router/route_coverage_test.dart test/widgets/video_feed_item/feed_playback_toggles_pill_test.dart test/services/build_config_test.dart test/services/feature_flag_service_test.dart test/l10n/arb_consistency_test.dart
- very_good test --coverage --min-coverage 100 (from mobile/packages/divine_ui)
- pre-push hook analyzer + changed-file tests

Both flags remain off by default for staged rollout.

PR #6266, opened 2026-07-21
fix(video-editor): let adjacent clip transitions fully consume a shared clip

### LINKED ISSUE #6187: fix(video-rendering): handle consecutive transitions correctly
### Summary

Consecutive transitions are not rendered correctly when two transitions are placed directly next to each other.

### Environment

any

### Steps to Reproduce

1. Create a video with two transitions placed back-to-back.
2. Export or render the video.
3. Play the rendered video.

### Actual Result

The second transition is rendered incorrectly or ignored entirely.

### Expected Result

Both consecutive transitions should be rendered correctly in the final video.

### Regression?

No

### Evidence

Reported on discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1528031273191473222)

hm21:
# Please hold off on reviewing this for now.

# It is close to done, but there is still a problematic iOS bug: on some devices, videos rendered this way hang in the video player — on loop restart the audio keeps playing while the picture freezes. I want that fixed before anyone spends time on a review.

## Summary

Reworks the overlap-transition consumption model so two consecutive transitions can each be full-length on a shared clip. An overlap (slide/dissolve/push/wipe) now consumes exactly its blend duration per side (`consumed = blend`) instead of reserving an extra solo lead — so two 0.5s blends fill a 1s middle clip exactly (0.5 + 0.5 = 1.0), adjacent transitions chain directly, and the picker's duration ceiling roughly doubles.

## Why

The old model reserved a full blend-length solo lead on each side of an overlap (`consumed = 2× duration`). That had two visible costs on short clips:

- **Restrictive cap:** a 1s single clip capped a loop-slide at 0.25s (half the room, halved again by the 2× budget), and two transitions could never both reach their full length on a shared clip.
- **Forced gap:** two consecutive transitions on a 1s middle clip always left a ~0.5s solo stretch between the two animations.

Removing the solo lead makes the shared clip the only constraint: `blend_A + blend_B ≤ shared clip`. Two 0.5s blends fit a 1s clip exactly, with no artificial reservation and no gap. The no-overlap clamp still prevents blends from overrunning a shared clip (e.g. two 0.6s on a 1s clip clamp to 0.5s each).

## Native dependency (blocking)

The blend now fully consumes the adjacent clip's contributed region, so the **native renderer must composite a fully-consumed overlap** (no non-blended frames) — otherwise it falls back to a hard cut (`not enough content for boundary`). That fix is **hm21/pro_video_editor#179**. This PR must not merge until that lands, is published, and `pro_video_editor` is bumped to `^2.7.1` here. (A temporary local path override was used during development and is intentionally not part of this PR.)

## Testing

- `flutter test` (transition geometry, seam render, transition sheet, timeline geometry): **192 passing**. `flutter analyze` + `dart format` clean.
- Verified on Android against the local native fix: consecutive and fully-consumed transitions render; both 0.5s on a 1s clip work.
- iOS/macOS device check pending (native Swift mirror is in the same pro_video_editor PR).

## Notes

- The seam-cache-miss-on-reclamp preview drop (the original mechanism diagnosed under this issue) is greatly reduced here — blends now fit, so re-clamping rarely happens — but not separately eliminated. A `cachedForPair` seam fallback is a clean follow-up if it still surfaces.
- Draft: blocked on hm21/pro_video_editor#179 publish + the `pro_video_editor` version bump.

Closes #6187

### DIFF
--- a/mobile/lib/models/video_editor/transition_geometry.dart
+++ b/mobile/lib/models/video_editor/transition_geometry.dart
@@ -23,10 +23,12 @@ bool _shortensTimeline(ClipTransitionType type) => !_isDip(type);
 /// Per-side playback duration a [transition] consumes at the boundary between
 /// clips of playback durations [a] and [b].
 ///
-/// An overlap (dissolve/slide/push/wipe) blends both clips at once and consumes
-/// 2× its duration per side; a dip (fadeToBlack/White) fades out then in,
-/// consuming half its duration per side. Clamped to the shorter clip so a
-/// transition can never overrun a clip.
+/// An overlap (dissolve/slide/push/wipe) blends both clips for its duration and
+/// consumes exactly that duration per side — the blend fills the consumed span,
+/// with no solo lead reserved (the native compositor renders the fully-consumed
+/// overlap). A dip (fadeToBlack/White) fades out then in, consuming half its
+/// duration per side. Clamped to the shorter clip so a transition can never
+/// overrun a clip.
 Duration transitionConsumedPerSide(
   Duration a,
   Duration b,
@@ -35,7 +37,7 @@ Duration transitionConsumedPerSide(
   final shorter = a < b ? a : b;
   final requested = _isDip(transition.type)
       ? Duration(microseconds: transition.duration.inMicroseconds ~/ 2)
-      : transition.duration * 2;
+      : transition.duration;
   return requested < shorter ? requested : shorter;
 }
 
@@ -47,9 +49,7 @@ Duration transitionDurationForConsumed(
   ClipTransitionType type,
 ) {
   if (consumed <= Duration.zero) return Duration.zero;
-  return _isDip(type)
-      ? consumed * 2
-      : Duration(microseconds: consumed.inMicroseconds ~/ 2);
+  return _isDip(type) ? consumed * 2 : consumed;
 }
 
 /// Maps each clip id to its outgoing transition, clamped so that **no clip is
@@ -225,9 +225,9 @@ Duration loopTransitionRoomPerSide(List<DivineVideoClip> clips) {
 /// timeline, preview and export line up 1:1.
 ///
 /// Mirrors `TransitionSeamRenderService.computeSeamSpans`: per side an overlap
-/// consumes 2× its duration (solo lead-in/out around the blend) and its seam
-/// plays 1.5× the consumed span; a dip consumes half its duration per side and
-/// its seam plays the full 2× consumed span (dips don't shorten).
+/// consumes its duration and its seam plays that same span (the blend fills
+/// it); a dip consumes half its duration per side and its seam plays the full
+/// 2× consumed span (dips don't shorten).
 class LoopWrapDisplay {
   const LoopWrapDisplay._({
     required this.consumedPerSide,
@@ -262,9 +262,7 @@ class LoopWrapDisplay {
       wrap,
     );
     if (consumed <= Duration.zero) return none;
-    final blend = _isDip(wrap.type)
-        ? Duration.zero
-        : Duration(microseconds: consumed.inMicroseconds ~/ 2);
+    final blend = _isDip(wrap.type) ? Duration.zero : consumed;
     return LoopWrapDisplay._(
       consumedPerSide: consumed,
       seamDuration: consumed * 2 - blend,
--- a/mobile/lib/services/video_editor/transition_seam_render_service.dart
+++ b/mobile/lib/services/video_editor/transition_seam_render_service.dart
@@ -279,9 +279,9 @@ class TransitionSeamRenderService {
   /// `_render` converts [consumed] back into each clip's own source time for
   /// trimming.
   ///
-  /// For overlaps the blend is always half the consumed span, guaranteeing a
-  /// solo lead-in/out on each side (a segment equal to the blend degenerates
-  /// into a hard cut).
+  /// For overlaps the blend fills the consumed span (blend == consumed): the
+  /// two clips overlap for the whole consumed span, with no solo lead — the
+  /// native compositor renders the fully-consumed overlap.
   @visibleForTesting
   ({Duration consumed, Duration blend, ClipTransition seamTransition})
   computeSeamSpans(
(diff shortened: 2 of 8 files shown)


rabble:
Draft status review: the native dependency blocker is now reflected in the branch via the `pro_video_editor` bump, and the last CI run on this head was green. The branch now conflicts with `main`, and the PR still lists iOS/macOS validation as pending. Please rebase and complete that platform pass before marking ready; as a draft, this remains read-only for reviewer takeover.

rabble (commented):
Draft, so feedback-only. I couldn't review the substance because the diff isn't reviewable in its current shape.

**This branch has 55 merge commits from `main` into it, and 3,373 commits not in `main`.** The resulting diff against the merge base is **3,164 files / +635,442 / −77,482**, with commits from a dozen different authors — because merging `main` drags everyone's history into the branch.

`AGENTS.md` and `.claude/rules/agent_workflow.md` §2 are explicit: *"Never merge `main` into a feature branch — always rebase."* That's exactly the failure mode this is, and the cost is that the actual change (bumping `pro_video_editor` and letting adjacent clip transitions run fully) is unrecoverable from the diff.

To make this reviewable:

```bash
git fetch origin
git rebase origin/main        # or: reset to origin/main and cherry-pick the real commits
git push --force-with-lease
```

Given it's been open since 2026-07-21 and `main` has moved a long way, starting a fresh branch off current `origin/main` and replaying just the intended changes is probably faster than untangling 55 merges.

Happy to review properly once the diff reflects only this PR's work. If the branch is dead, closing it would take it off the review queue.

NotThatKindOfDrLiz:
Overlap note from the #7590 follow-ups:

#7606 now tracks the selected-state theming cleanup in `video_editor_transition_sheet.dart` / `video_editor_layer_animation_sheet.dart`. I checked this PR's current diff: it overlaps `video_editor_transition_sheet.dart`, but only for transition-duration behavior/comments, not the `isLight ? ... : VineTheme.primary` selected-state theming.

No action needed here unless this PR later edits those selected-state color paths; in that case, please cross-check #7606 so the two changes do not diverge.

hm21:
Please hold off on reviewing this for now.

It is close to done, but there is still a problematic iOS bug: on some devices, videos rendered this way hang in the video player — on loop restart the audio keeps playing while the picture freezes. I want that fixed before anyone spends time on a review.

rabble (commented):
The geometry change is clean and the commit message is unusually good — it states the old model, the symptom (1s loop clip capping a slide at 0.25s), the new invariant, and the external blocker in one place. Dropping the solo lead so `consumed == blend` makes `transitionDurationForConsumed` an exact inverse of `transitionConsumedPerSide` for overlaps, which is a nicer property than the 2×/½ pair it replaces, and the no-overlap clamp still holds the safety line.

Two process things stand between it and mergeable, neither about the code:

## The external dependency needs confirming, in the PR

The fix commit says it plainly:

> Requires the native renderer to composite a fully-consumed overlap (pro_video_editor #179) — blocked until that is published and the dep is bumped.

and the branch then carries `build(video-editor): bump pro_video_editor to ^2.7.1`. So the intent is that 2.7.1 is the release containing #179. Could you confirm that in the description — ideally linking the upstream release notes? Right now a reviewer has to take it on faith, and if 2.7.1 turns out not to include it, the Dart side ships a geometry the native compositor won't render, which fails as visibly wrong output rather than as a crash or a red test.

Also worth saying whether `^2.7.1` is deliberate: a caret range lets a future 2.8.x in without review, on a dependency whose rendering behaviour this change is now coupled to.

## The branch is very stale

`origin/main..pr6266` still contains ~10 changes that landed long ago (#6216, #6244, #6264, #6258, #6256, #6213, #6247, #6245, #6212, #6145), dating to 21 July. It needs `git fetch origin && git rebase origin/main` before it can be evaluated against current `main` — I reviewed the two real commits in isolation because there's no usable merge base to diff against.

Two specifics that rebase will surface: #6145 added the design-system drift ratchets, and light mode has since shipped un-gated (this branch predates that, still carrying the feature-flagged `#6216` version). Neither should touch transition geometry, but the video editor is exactly where "dark-mode-only" assumptions tend to be baked in, so it's worth a look while rebasing.

## Timing

Not a review point, just context: this touches the video editor and lands right as we're trying to cut 1.0.20. If it's not needed for that release, it's a natural one to land just after.

## Not verified

I reviewed the two commits' diffs only, and did not run the 192 regenerated tests or look at rendered output. The claim I'd most want independently checked is the one I can't check from here — that a fully-consumed overlap actually composites correctly in 2.7.1 on both platforms.

NotThatKindOfDrLiz:
## Re-review Summary

The PR's technical substance is sound — the transition consumption geometry change is well-reasoned, tested (all 192 tests pass), and correctly documented. However, three issues block merge:

### 1. **Stale Seam Cache (Critical)**
The PR changes `computeSeamSpans` algorithm (overlap blend from `consumed/2` → `consumed`) but does NOT bump `_seamCacheVersion` in `transition_seam_render_service.dart` (line 343). 

**Effect:** Pre-upgrade seam files are replayed with old geometry → timeline/preview/export mismatch.

**Fix:** Bump `_seamCacheVersion` from 4 to 5 with a doc note explaining the algorithm change. Test on a clean install first (no pre-upgrade seam files) to rule this out as the cause of the iOS loop-restart hang you reported.

### 2. **iOS Loop-Restart Picture Hang (Blocking)**
You reported: picture freezes, audio continues on loop restart. This is the actual blocker.

**Investigation path:**
- Reproduce on the current build (might be stale cache issue in disguise)
- Check `pro_video_editor` 2.10.0 changelog for `VideoRenderData.trimToCommonTrackEnd` — this was added to address audio/picture sync at loop seams. Might be relevant.
- If the hang persists after stale cache fix, this becomes a pro_video_editor integration issue (platform rendering layer, not transition math).

### 3. **Unnecessary Pubspec Version Bump**
The PR bumps `pro_video_editor` from `^2.6.0` to `^2.7.1`, but `origin/main` is already on `^2.11.3` (you're actually downgrading on rebase). Drop this commit when you rebase.

### Branch Status
Prior feedback noted "55 merge commits ahead" — that was a measurement error. Your branch is actually clean with 2 commits, 8 files changed. A straightforward rebase onto `origin/main` (not a restart) fixes it.

### Why the geometry change is correct
The 2.7.1 pro_video_editor changelog confirms: *"A clip transition whose blend fully consumes an adjacent clip now renders instead of degrading to a hard cut."* Your change aligns the seam geometry assumptions with what the native compositor expects.

**Next steps:**
1. Bump `_seamCacheVersion` to 5
2. Drop the pro_video_editor pubspec commit on rebase
3. Rebase onto fresh `origin/main`
4. Test on clean install to isolate the iOS hang (stale cache vs. native regression)

PR #6267, opened 2026-07-21
fix(media): send auth for age-gated thumbnails

### LINKED ISSUE #6250: fix(media): send BUD-01 auth so age-restricted thumbnails load

NotThatKindOfDrLiz:
## Summary
- retry age-restricted Divine-hosted thumbnails with passive BUD-01 auth after a 401/403, using the thumbnail URL hash instead of the video hash
- share the passive-auth thumbnail image path across `VideoThumbnailWidget`, profile grid thumbnails, feed loading placeholders, and curated-list cards
- gate passive thumbnail auth to authenticated, age-verified users whose adult playback preference is `show`; warn/hide/unverified states keep the existing placeholder path
- keep passive thumbnail auth re-armable with persistent content-filter/divine-host generation counters wired through the real services
- clear the shared image cache and rearm mounted thumbnails when adult media access is revoked by age verification, content-filter preference changes, logout, or account cleanup

## Behavior notes
- Divine-hosted thumbnail URLs intentionally bypass `VineCachedImage`/persistent disk caching and use the raw `Image.network` path. That path gives typed 401/403 failures for passive-auth retry and avoids the cache-manager stalls this widget was already bypassing for Divine blob hosts; it now preserves caller placeholders, error widgets, and fade timings.
- The persistent moderation generation counters mean existing feed/detail consumers that watch `contentFilterVersionProvider` or `divineHostFilterVersionProvider` now genuinely rebuild after content-filter, feed-aspect-ratio, or Divine-host-only preference changes. Affected feed providers will rebuild: `videosRepositoryProvider`, `videosRepositoryByHashProvider`, `curatedListProvider`, `feedVideosProvider`, etc. QA should regression-test:
  - Feed grid refresh when toggling content-filter preferences (nudity, sexual, violence, drugs, etc.)
  - Profile tab grid rebuild when switching between own profile vs. other profiles
  - Curated list refresh when moderation settings change
  - Thumbnail display consistency across re-routed call sites (profile grid, video feed, loading placeholder)

## Trust and safety
Passive thumbnail auth is intentionally stricter than tapped playback auth. It never shows a dialog, never falls back to URL-bound auth, and only signs hash-bound Blossom requests after the viewer has opted into always showing adult content.

## Validation
- `flutter analyze --no-pub`
- `flutter test test/widgets/video_thumbnail_widget_test.dart test/providers/moderation_providers_test.dart test/providers/videos_repository_filter_reactivity_test.dart test/utils/blossom_blob_hash_test.dart`
- `flutter test test/services/media_auth_interceptor_test.dart test/services/age_verification_service_test.dart test/services/content_filter_service_test.dart test/services/media_viewer_auth_service_test.dart test/widgets/vine_cached_image_test.dart test/widgets/video_thumbnail_widget_test.dart test/widgets/profile/profile_tab_thumbnail_test.dart test/widgets/list_search_card_test.dart test/widgets/video_feed_item/video_loading_placeholder_test.dart test/providers/videos_repository_filter_reactivity_test.dart test/providers/moderation_providers_test.dart test/utils/blossom_blob_hash_test.dart`
- pre-push hook: merge-conflict check, analyzer, generated-file verification, changed-file tests

Closes #6250

### DIFF
--- a/mobile/lib/widgets/video_thumbnail_widget.dart
+++ b/mobile/lib/widgets/video_thumbnail_widget.dart
@@ -1,12 +1,18 @@
 // ABOUTME: Smart video thumbnail widget that displays thumbnails or blurhash placeholders
 // ABOUTME: Uses existing thumbnail URLs from video events and falls back to blurhash when missing
 
+import 'dart:async';
+
 import 'package:blurhash_service/blurhash_service.dart';
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_cache_manager/flutter_cache_manager.dart'
     show HttpExceptionWithStatus;
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:models/models.dart' hide AspectRatio, LogCategory;
+import 'package:openvine/models/viewer_auth_result.dart';
+import 'package:openvine/providers/app_providers.dart';
+import 'package:openvine/utils/blossom_blob_hash.dart';
 import 'package:openvine/widgets/blurhash_display.dart';
 import 'package:openvine/widgets/vine_cached_image.dart';
 import 'package:unified_logger/unified_logger.dart';
@@ -176,7 +182,7 @@ class _VideoThumbnailWidgetState extends State<VideoThumbnailWidget> {
 
 /// Error-safe network image widget that prevents HTTP 404 and other network exceptions.
 /// Uses [VineCachedImage] for shared cache-backed loading where appropriate.
-class _SafeNetworkImage extends StatelessWidget {
+class _SafeNetworkImage extends StatefulWidget {
   const _SafeNetworkImage({
     required this.url,
     required this.videoId,
@@ -214,6 +220,74 @@ class _SafeNetworkImage extends StatelessWidget {
   }
 
   @override
+  State<_SafeNetworkImage> createState() => _SafeNetworkImageState();
+}
+
+class _SafeNetworkImageState extends State<_SafeNetworkImage> {
+  Map<String, String>? _authHeaders;
+  bool _authRetryAttempted = false;
+
+  @override
+  void didUpdateWidget(covariant _SafeNetworkImage oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (oldWidget.url != widget.url) {
+      _authHeaders = null;
+      _authRetryAttempted = false;
+    }
+  }
+
+  Future<void> _maybeRetryWithPassiveAuth(Object error) async {
+    if (_authRetryAttempted || !_isAuthChallenge(error)) return;
+    _authRetryAttempted = true;
+
+    final sha256Hash = extractSha256FromBlossomUrl(widget.url);
+    if (sha256Hash == null) return;
+
+    final authResult = await ProviderScope.containerOf(context, listen: false)
+        .read(mediaAuthInterceptorProvider)
+        .createPassiveAuthHeadersForAdultMedia(
+          sha256Hash: sha256Hash,
+          serverUrl: _extractServerUrl(widget.url),
+        );
+    if (!mounted) return;
+
+    switch (authResult) {
+      case ViewerAuthAuthorized(:final headers):
+        setState(() {
+          _authHeaders = headers;
+        });
+      case ViewerAuthSignerUnreachable():
+      case ViewerAuthBlockedByPreference():
+      case ViewerAuthUnavailable():
+        break;
+    }
+  }
+
+  bool _isAuthChallenge(Object error) {
+    if (error is NetworkImageLoadException) {
+      return error.statusCode == 401 || error.statusCode == 403;
+    }
+    if (error is HttpExceptionWithStatus) {
+      return error.statusCode == 401 || error.statusCode == 403;
+    }
+    final message = error.toString();
+    return message.contains(' 401 ') ||
+        message.contains(' 403 ') ||
+        message.contains('statusCode: 401') ||
+        message.contains('statusCode: 403');
+  }
+
+  String? _extractServerUrl(String url) {
+    try {
+      final uri = Uri.parse(url);
+      final portSuffix = uri.hasPort ? ':${uri.port}' : '';
+      return '${uri.scheme}://${uri.host}$portSuffix';
+    } catch (_) {
+      return null;
+    }
+  }
+
+  @override
   Widget build(BuildContext context) {
     return LayoutBuilder(
       builder: (context, constraints) {
@@ -223,30 +297,36 @@ class _SafeNetworkImage extends StatelessWidget {
         // thumbnails so they visibly reload on the way back. Capping the decode
         // width keeps far more thumbnails resident in the cache.
         final cacheWidth = _decodeWidth(context, constraints.maxWidth);
+        final resolvedUrl = widget.url;
 
         // Debug mode: test with plain Image.network to isolate cache issues
-        if (_useSimpleImageNetwork || _shouldBypassCacheManager(url)) {
+        if (_SafeNetworkImage._useSimpleImageNetwork ||
+            _SafeNetworkImage._shouldBypassCacheManager(resolvedUrl)) {
           final imageProvider = ResizeImage.resizeIfNeeded(
             cacheWidth,
             null,
-            NetworkImage(url),
+            NetworkImage(resolvedUrl, headers: _authHeaders),
           );
           return ImageWithDimensionsListener(
             imageProvider: imageProvider,
-            onImageDimensionsResolved: onImageDimensionsResolved == null
+            onImageDimensionsResolved: widget.onImageDimensionsResolved == null
                 ? null
-                : (width, height) =>
-                      onImageDimensionsResolved!(url, width, height),
+                : (width, height) => widget.onImageDimensionsResolved!(
+                    resolvedUrl,
+                    width,
+                    height,
+                  ),
             child: Image(
               image: imageProvider,
-              width: width,
-              height: height,
-              fit: fit,
+              width: widget.width,
+              height: widget.height,
+              fit: widget.fit,
               alignment: Alignment.topCenter,
               errorBuilder: (context, error, stackTrace) {
+                unawaited(_maybeRetryWithPassiveAuth(error));
                 Log.warning(
-                  '🖼️ [Image.network] Thumbnail load failed for video $videoId:\n'
-                  '  URL: $url\n'
+                  '🖼️ [Image.network] Thumbnail load failed for video ${widget.videoId}:\n'
+                  '  URL: ${widget.url}\n'
                   '  Error type: ${error.runtimeType}\n'
                   '  Error: $error\n'
                   '  Stack: ${stackTrace?.toString().split('\n').take(5).join('\n')}',
@@ -254,8 +334,8 @@ class _SafeNetworkImage extends StatelessWidget {
                   category: LogCategory.video,
                 );
                 return Container(
-                  width: width,
-                  height: height,
+                  width: widget.width,
+                  height: widget.height,
                   color: VineTheme.transparent,
                 );
               },
@@ -264,23 +344,28 @@ class _SafeNetworkImage extends StatelessWidget {
         }
 
         return VineCachedImage(
-          imageUrl: url,
-          width: width,
-          height: height,
-          fit: fit,
+          imageUrl: resolvedUrl,
+          width: widget.width,
+          height: widget.height,
+          fit: widget.fit,
           memCacheWidth: cacheWidth,
+          authHeaders: _authHeaders,
           alignment: Alignment.topCenter,
-          onImageDimensionsResolved: onImageDimensionsResolved == null
+          onImageDimensionsResolved: widget.onImageDimensionsResolved == null
               ? null
-              : (width, height) =>
-                    onImageDimensionsResolved!(url, width, height),
+              : (width, height) => widget.onImageDimensionsResolved!(
+                  resolvedUrl,
+                  width,
+                  height,
+                ),
           // Show transparent container so background surfaceContainer color shows through
           placeholder: (context, url) => Container(
-            width: width,
-            height: height,
+            width: widget.width,
+            height: widget.height,
             color: VineTheme.transparent,
           ),
           errorWidget: (context, url, error) {
+            unawaited(_maybeRetryWithPassiveAuth(error));
             // 404s are expected — thumbnail may not exist yet.
             final is404 = error is HttpExceptionWithStatus
                 ? error.statusCode == 404
@@ -288,7 +373,7 @@ class _SafeNetworkImage extends StatelessWidget {
 
             if (!is404) {
               Log.warning(
-                '🖼️ Thumbnail load failed for video $videoId:\n'
+                '🖼️ Thumbnail load failed for video ${widget.videoId}:\n'
                 '  URL: $url\n'
                 '  Error type: ${error.runtimeType}\n'
                 '  Error: $error',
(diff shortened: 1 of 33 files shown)


mobile/lib/widgets/video_thumbnail_widget.dart
```diff
@@ -213,6 +219,74 @@ class _SafeNetworkImage extends StatelessWidget {
...
+      _authHeaders = null;
+      _authRetryAttempted = false;
+    }
+  }
+
+  Future<void> _maybeRetryWithPassiveAuth(Object error) async {
+    if (_authRetryAttempted || !_isAuthChallenge(error)) return;
+    _authRetryAttempted = true;
+
+    final sha256Hash = extractSha256FromBlossomUrl(widget.url);
```
hm21: Despite the description and test names scoping this to *Divine-hosted* thumbnails, the retry isn't actually host-restricted — it pulls a 64-hex out of any thumbnail URL and signs a BUD-01 viewer-auth to whatever host served it (`_shouldBypassCacheManager`'s host check only selects the render path, it doesn't gate this). A hostile feed event with a thumbnail like `https://evil/<sha256>.jpg` that returns 401 could harvest a signed viewer-auth (npub + an "age-verified, adult-content-on" signal) from a mere grid impression — no tap. It's stricter than the playback path (show-only) but broadens the passive surface to every visible grid cell. Intended? If not, gate the retry on `*.divine.video` (you can reuse `_shouldBypassCacheManager`).

hm21 (approved):
Careful, well-gated change and the tests can actually fail — nice. One thing to confirm before merge (inline). Minor: `clearOpenVineImageCache()` wipes the entire global image cache (all avatars/UI), not just adult media, on every show→warn/hide flip and on logout — fine as a safety hammer, maybe worth a one-line comment that the broad clear is intentional.

mobile/lib/widgets/video_thumbnail_widget.dart
```diff
@@ -213,6 +219,74 @@ class _SafeNetworkImage extends StatelessWidget {
     return host == 'divine.video' || host.endsWith('.divine.video');
   }
 
+  @override
+  State<_SafeNetworkImage> createState() => _SafeNetworkImageState();
+}
+
+class _SafeNetworkImageState extends State<_SafeNetworkImage> {
+  Map<String, String>? _authHeaders;
```
dcadenas, line 241: Clear the stored headers and current image when passive access is revoked, not only when the URL changes. At `mobile/lib/widgets/video_thumbnail_widget.dart:227`, `_authHeaders` survives while a same-URL DM thumbnail remains mounted. The revocation callback clears shared caches but cannot clear this widget-local state, so the thumbnail can remain visible or reuse its old signed header after the viewer changes adult content from `show` to `warn` or `hide`. Make this widget observe the account, age-verification, and adult-preference generation, then clear its headers and image immediately when access is revoked.

mobile/lib/widgets/video_thumbnail_widget.dart
```diff
@@ -213,6 +219,74 @@ class _SafeNetworkImage extends StatelessWidget {
...
+  void didUpdateWidget(covariant _SafeNetworkImage oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (oldWidget.url != widget.url) {
+      _authHeaders = null;
+      _authRetryAttempted = false;
+    }
+  }
+
+  Future<void> _maybeRetryWithPassiveAuth(Object error) async {
+    if (_authRetryAttempted || !_isAuthChallenge(error)) return;
```
dcadenas: Wait for persisted age verification and adult preferences to be ready before consuming the one-shot passive-auth retry. At `mobile/lib/widgets/video_thumbnail_widget.dart:240`, `_authRetryAttempted` is set before the lazily initialized moderation services finish loading, so their temporary fail-closed values can return `ViewerAuthUnavailable` for an eligible returning viewer. Initialization does not re-drive the same URL, leaving the thumbnail on its placeholder until remount. Expose an awaitable moderation-readiness contract and set the retry flag only after that readiness completes.

dcadenas (dismissed):
Before this can merge, thumbnail auth state must be invalidated when adult access is revoked, and passive-auth retries must wait for persisted moderation state to finish loading. Without those guards, mounted inbox thumbnails can keep using stale signed headers, while eligible returning viewers can lose their only retry and remain on a placeholder.

NotThatKindOfDrLiz:
Took over the review fixes and pushed `f0b5f513d`.

Addressed the merge blockers:
- moved logout/account-switch image-cache clearing after DM listener shutdown and made it non-fatal
- surfaced media-cache HTTP status codes via `MediaCacheImageLoadException`, so cached thumbnail failures can be typed 401/403 auth challenges
- host-gated passive thumbnail auth to Divine-hosted URLs regardless of render path
- blocked passive signing for interactive signers and waited for persisted age/content-filter state before retrying
- invalidated mounted thumbnail auth headers when adult media access is revoked
- moved content-filter cache invalidation before listener notification
- documented the new public helpers and added direct blossom hash tests
- removed the unrelated conversation-list test flake change from this PR

Local verification:
- `flutter pub get`
- `flutter analyze`
- `flutter test test/services/media_auth_interceptor_test.dart test/services/media_viewer_auth_service_test.dart test/services/age_verification_service_test.dart test/services/content_filter_service_test.dart test/widgets/vine_cached_image_test.dart test/widgets/video_thumbnail_widget_test.dart test/utils/blossom_blob_hash_test.dart`
- `flutter test test/src/media_cache_image_provider_test.dart` from `mobile/packages/media_cache`
- pre-push hook passed analyzer, generated-file verification, and changed-file tests

@dcadenas this should address both requested-change comments. @hm21 the host-gate answer is no, it was not intended to sign for arbitrary hosts; the retry is now explicitly gated on the Divine thumbnail host path, and the broad cache clear has a comment.

mobile/lib/widgets/video_thumbnail_widget.dart
```diff
@@ -63,6 +81,55 @@ class _VideoThumbnailWidgetState extends State<VideoThumbnailWidget> {
...
+    }
+    if (identical(container, _providerContainer)) return;
+
+    _clearProviderSubscriptions();
+    _providerContainer = container;
+    _authStateSubscription = container.listen<Object?>(
+      currentAuthStateProvider,
+      (previous, next) => _incrementPassiveAuthInvalidationKey(),
+    );
+    _contentFilterSubscription = container.listen<int>(
```
dcadenas: Make `contentFilterVersionProvider` emit a persistent monotonic generation, or subscribe this widget directly to `ContentFilterService`. The provider currently rebuilds its local counter as `0` after each `invalidateSelf()`, so changing Warn or Hide to Show produces no new value and leaves the mounted thumbnail's consumed `_authRetryAttempted` latch in place after a 401 or 403. Add a widget regression test that changes the preference while the same thumbnail remains mounted and proves the same URL retries with passive auth.

dcadenas (dismissed):
The mounted thumbnail does not receive a real change when adult preferences switch from Warn or Hide to Show, so it cannot rearm its passive-auth retry after a 401 or 403. Make the generation persistent and add a regression test proving that a still-mounted same-URL thumbnail retries after access is granted before merging.

NotThatKindOfDrLiz:
Takeover update pushed in `5d4fdefb2`.

What changed:
- fixed `contentFilterVersionProvider` and `divineHostFilterVersionProvider` at the source by making their counters persistent `NotifierProvider`s, so mounted thumbnails and existing feed consumers actually observe generation changes
- moved the passive thumbnail auth/retry lifecycle into reusable `PassiveAuthThumbnailImage` and routed profile grid thumbnails, feed loading placeholders, and curated-list cards through it
- removed the dead `VineCachedImage` production auth-header path from this PR scope; Divine thumbnails use the shared passive-auth `Image.network` path instead
- made adult media access revocation rearm mounted thumbnails on content-filter revocation, age-verification revocation, logout, and account cleanup
- addressed the cheap cleanup items: no new legacy `StateProvider`, no duplicate map equality helper, and added the requested comment documenting the intentionally broad cache clear

Validation after rebasing onto current `origin/main`:
- `flutter analyze --no-pub`
- `flutter test test/services/media_auth_interceptor_test.dart test/services/age_verification_service_test.dart test/services/content_filter_service_test.dart test/services/media_viewer_auth_service_test.dart test/widgets/vine_cached_image_test.dart test/widgets/video_thumbnail_widget_test.dart test/widgets/profile/profile_tab_thumbnail_test.dart test/widgets/list_search_card_test.dart test/widgets/video_feed_item/video_loading_placeholder_test.dart test/providers/videos_repository_filter_reactivity_test.dart`
- pre-push hook: analyzer, generated-file verification, changed-file tests (`156` tests)

mobile/test/widgets/video_thumbnail_widget_test.dart
```diff
@@ -285,11 +308,322 @@ void main() {
...
+        ).thenAnswer(
+          (_) async =>
+              const ViewerAuthAuthorized({'Authorization': 'Nostr token'}),
+        );
+
+        final container = ProviderScope.containerOf(
+          tester.element(find.byType(VideoThumbnailWidget)),
+          listen: false,
+        );
+        container.read(contentFilterVersionProvider.notifier).increment();
```
dcadenas: Drive the adult preference through the real `ContentFilterService` path in this test instead of calling `contentFilterVersionProvider.notifier.increment()` directly. Keep the same 401/403-challenged thumbnail mounted and assert that changing the preference to `show` retries the same URL with passive auth. This verifies the production listener chain from `ContentFilterService.notifyListeners()` to `contentFilterVersionProvider`, which the current override bypasses.

dcadenas (dismissed):
The rearm regression test still bypasses the production adult-preference notification path. Wire the test through ContentFilterService before merge so it proves a preference change can rearm the same mounted thumbnail after a blocked auth attempt.

NotThatKindOfDrLiz:
Takeover follow-up pushed in `f6d2e65dc`.

What changed:
- preserved `PassiveAuthThumbnailImage` caller behavior on the Divine `Image.network` path: placeholders render while loading, error widgets are honored on failure, and fade durations are used for the raw network image path
- removed the unreachable passive-auth retry call from the `VineCachedImage` branch
- rewired the mounted-thumbnail rearm regression through the real `ContentFilterService.setPreference(...) -> notifyListeners() -> contentFilterVersionProvider` path
- added `test/providers/moderation_providers_test.dart` to prove the production content-filter and Divine-host service notifications increment their Riverpod version providers
- shared media server-origin extraction through `blossom_blob_hash.dart` and reused it from pooled age-restricted playback
- updated stale profile thumbnail docs and the PR body with the intentional Divine disk-cache bypass plus the wider feed-provider rebuild blast radius

Validation:
- `flutter analyze --no-pub`
- `flutter test test/widgets/video_thumbnail_widget_test.dart test/providers/moderation_providers_test.dart test/providers/videos_repository_filter_reactivity_test.dart test/utils/blossom_blob_hash_test.dart`
- `flutter test test/services/media_auth_interceptor_test.dart test/services/age_verification_service_test.dart test/services/content_filter_service_test.dart test/services/media_viewer_auth_service_test.dart test/widgets/vine_cached_image_test.dart test/widgets/video_thumbnail_widget_test.dart test/widgets/profile/profile_tab_thumbnail_test.dart test/widgets/list_search_card_test.dart test/widgets/video_feed_item/video_loading_placeholder_test.dart test/providers/videos_repository_filter_reactivity_test.dart test/providers/moderation_providers_test.dart test/utils/blossom_blob_hash_test.dart`
- pre-push hook passed: merge-conflict check, analyzer, generated-file verification, changed-file tests
- GitHub checks are green on `f6d2e65dc`

@dcadenas this should address the remaining requested-change item by driving the regression through the production `ContentFilterService` notification path.

mobile/test/widgets/video_thumbnail_widget_test.dart
```diff
@@ -285,11 +322,375 @@ void main() {
...
+              const ViewerAuthAuthorized({'Authorization': 'Nostr token'}),
+        );
+
+        final container = ProviderScope.containerOf(
+          tester.element(find.byType(VideoThumbnailWidget)),
+          listen: false,
+        );
+        final filterService = container.read(contentFilterServiceProvider);
+        await filterService.initialized;
+        await filterService.setPreference(
```
dcadenas, line 692: Exercise the real adult-preference transition here. Change `nudity` or `sexual` from `hide` or `warn` to `show` through the production `ContentFilterService` while keeping the same 401/403-challenged URL mounted. Do not make the interceptor mock authorized independently; assert that the retry gets authorization only after the production preference change. This test should fail if that transition does not rearm passive auth.

dcadenas (changes requested):
The regression test still does not prove the adult-preference transition that makes passive thumbnail authentication eligible. Update the test before merging so it drives an adult setting from hide or warn to show through ContentFilterService while the challenged thumbnail remains mounted.

NotThatKindOfDrLiz:
Takeover follow-up pushed in `ab5f92b6f`.

What changed:
- guarded passive thumbnail auth completion with the requested URL and image generation, so stale in-flight auth cannot reinstall headers after revocation or URL reuse
- replaced the remaining weak rearm regression with a production-path test: real `ContentFilterService`, real `MediaAuthInterceptor`, same mounted thumbnail, adult preferences moving from `hide` to `show`
- restored center alignment for the migrated profile-grid, feed-placeholder, and curated-list thumbnail call sites, with regression assertions
- removed the orphaned profile-grid `openVineImageCache.downloadFile(url)` warm-up now that Divine-hosted thumbnails intentionally bypass that disk cache
- cleaned up the dead passive-auth debug toggle, dead `HttpExceptionWithStatus` auth-challenge branch, and stale profile thumbnail test comment

Validation:
- `flutter analyze --no-pub`
- `flutter test test/services/media_auth_interceptor_test.dart test/services/age_verification_service_test.dart test/services/content_filter_service_test.dart test/services/media_viewer_auth_service_test.dart test/widgets/vine_cached_image_test.dart test/widgets/video_thumbnail_widget_test.dart test/widgets/profile/profile_tab_thumbnail_test.dart test/widgets/list_search_card_test.dart test/widgets/video_feed_item/video_loading_placeholder_test.dart test/providers/videos_repository_filter_reactivity_test.dart test/providers/moderation_providers_test.dart test/utils/blossom_blob_hash_test.dart`
- pre-push hook passed: merge-conflict check, analyzer, generated-file verification, changed-file tests
- GitHub checks are green on `ab5f92b6f`

@dcadenas requested re-review; this addresses the standing adult-preference transition test request and the stale in-flight auth race found during takeover.

NotThatKindOfDrLiz:
@dcadenas if you have any remaining changes you want here, please push them directly so we can close this out.

dcadenas (approved):
🔒 age-gated thumbnails get auth now

PR #6268, opened 2026-07-21
fix(video-player): classify auth-required media errors

### LINKED ISSUE #6252: fix(video-player): map iOS NSURLErrorDomain -1013 (auth required)

NotThatKindOfDrLiz:
## Summary
- add a typed native auth-required player error code and parse it on Dart
- map iOS NSURLErrorUserAuthenticationRequired and native HTTP 401 responses to auth_required on iOS/Android
- classify typed auth_required runtime failures as age restricted without relying on 401 text

## Tests
- flutter analyze (mobile/packages/divine_video_player)
- flutter analyze (mobile/packages/infinite_video_feed)
- flutter test --coverage (mobile/packages/divine_video_player)
- flutter test --coverage (mobile/packages/infinite_video_feed)

Native Android unit tests were not run locally because this checkout has no Gradle wrapper and the shell does not have a gradle executable.

Closes #6252

### DIFF
--- a/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
+++ b/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
@@ -1062,6 +1062,9 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
             if httpResponse.statusCode == 202 {
                 return "media_processing"
             }
+            if httpResponse.statusCode == 401 {
+                return "auth_required"
+            }
             return httpResponse.statusCode >= 500 ? "http_server_error" : "http_client_error"
         }
 
@@ -1076,6 +1079,8 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
             return "network_error"
         case (NSURLErrorDomain, NSURLErrorTimedOut):
             return "timeout"
+        case (NSURLErrorDomain, NSURLErrorUserAuthenticationRequired):
+            return "auth_required"
         default:
             // AVFoundation format / decoder errors.
             if err.domain == AVFoundationErrorDomain {
--- a/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
+++ b/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
@@ -64,6 +64,7 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
     private var speed: Double = 1.0
     private var currentStatus: String = "idle"
     private var errorMessage: String?
+    private var errorCode: String?
     private var firstFrameRendered: Bool = false
     private var videoWidth: Int = 0
     private var videoHeight: Int = 0
@@ -271,6 +272,8 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
                 self.safePreroll(at: startTime)
 
                 self.currentStatus = "ready"
+                self.errorMessage = nil
+                self.errorCode = nil
                 self.clearSetClipsTimeout()
                 DivineVideoPlayerLog.shared.info(
                     "Player \(self.playerId) ready: \(self.clipCount) clip(s), "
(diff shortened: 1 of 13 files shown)


rabble (approved):
Reviewed the full diff (native Kotlin/Swift mapping, Dart enum, feed classification, tests).

**Correct end-to-end:**
- Android: `status == 401 -> "auth_required"` is checked before the generic `400..499` branch. ✅
- iOS/macOS: both the `NSURLErrorFailingURLResponseErrorKey` 401 path and the bare `NSURLErrorUserAuthenticationRequired` (-1013) case map to `auth_required`, placed before the 202-agnostic 4xx/5xx fallthrough and before `default:`. ✅
- `authRequired.shouldFailover = true` / `isTransient = false` matches the existing 401-text behavior: the feed cycles fallback sources first, then classifies as `ageRestricted`, and `_isAutoRetryableErrorType` excludes it from auto-retry — the widget test pins exactly that (no `setClips` after the retry delay). ✅
- `classifyVideoError` typed-code check runs before message sniffing, and the null-code test guards the legacy path. ✅

**Minor / follow-up (non-blocking):**

1. The init-time and failover exception paths still classify by message only (`infinite_video_feed.dart` ~1231, ~1375, ~1479 all call `classifyVideoError(errorMessage: e.toString())`). An auth failure that surfaces as an init exception with only `NSURLErrorDomain error -1013` text (no "401") stays `generic` and keeps auto-retrying. Understood that #6252 scoped this to the runtime error stream — worth a follow-up issue if init-time 401s show up in the wild.

2. In `apple_auth_error_contract_test.dart`, the second test's `source.indexOf('return "auth_required"')` has no start offset. It currently resolves to the 401 mapping only because that block happens to precede the `NSURLErrorUserAuthenticationRequired` case in the file — if the Swift blocks are ever reordered, the assertion silently tests the wrong occurrence. Pass `statusCheck` as the start index like the sibling test does.

Source-text contract tests are a fair trade given no Gradle wrapper/XCTest runner in this checkout — they can fail on a real regression, which meets the bar.

mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
```diff
@@ -1062,6 +1062,9 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
             if httpResponse.statusCode == 202 {
                 return "media_processing"
             }
+            if httpResponse.statusCode == 401 {
```
dcadenas, line 1083: Propagate the auth code from the composition-load catch before relying on this mapping. Darwin loads remote asset metadata before assigning `player.currentItem`; when `AVURLAsset.load` throws for 401/-1013, the catch at `DivineVideoPlayerInstance.swift:282-299` returns `COMPOSITION_ERROR` with no details, so Dart falls back to exception text and misses `NSURLErrorDomain error -1013`. Include the canonical code in `FlutterError.details`, decode it in `mobile/lib/src/widgets/infinite_video_feed.dart:1218-1233` after source fallbacks are exhausted, and add a regression test showing the final error is `ageRestricted`.

dcadenas (changes requested):
The new Darwin auth mapping does not cover initial asset loads that fail before an AVPlayerItem exists. Propagate canonical auth codes from composition-load failures so age-restricted media reaches the age gate before merge.

NotThatKindOfDrLiz:
Takeover update pushed in 043d80e62.

What changed:
- carries native auth_required out through composition/init failures as PlatformException details, including Darwin asset-load failures before AVPlayerItem exists and Android pending setClips failures
- parses PlatformException details in the feed init, failover, and processing-retry catch paths before classifying the video error
- treats authRequired as asset-scoped, so Divine variant failover stops immediately and the age gate surfaces without auto-retry
- applied the Apple source-text test offset fix and added regression coverage for composition failures plus multi-source auth-required behavior

Verification:
- local pre-push analyzer and changed-file Flutter tests passed
- local package checks passed for divine_video_player and infinite_video_feed, including flutter test --coverage for both packages
- GitHub CI is green on 043d80e62, including Android Unit Tests

@dcadenas this should address the blocker you called out. @rabble I also applied the indexOf offset note.

dcadenas (approved):
🔒 auth-required playback errors now surface clearly

PR #6270, opened 2026-07-21
fix(bluesky): refresh status after NIP-05 claim

### LINKED ISSUE #6201: fix(bluesky): refresh crosspost status after NIP-05 claim flow
## Problem

PR #6139 routes users without a claimed `username.divine.video` handle into the NIP-05 claim flow from Bluesky settings. After the user completes that flow and returns, the Bluesky settings screen keeps the stale `usernameClaimed == false` state until the screen is reopened.

## Expected

When the user returns from the NIP-05 claim screen, Bluesky settings should refresh crosspost status and hide the claim notice once a username is claimed.

## Scope

- Await both `context.push(Nip05SettingsScreen.path)` paths from Bluesky settings.
- After return, if `context.mounted`, call `context.read<CrosspostSettingsCubit>().loadStatus()`.
- Add a widget test that enters the claim route, pops back, and verifies status is fetched again / the notice disappears after updated status.

Follow-up from #6139.

NotThatKindOfDrLiz:
## Summary
- Refresh Bluesky crosspost status when returning from the NIP-05 claim flow from both entry points.
- Rebuild loaded crosspost state from status responses so nullable username/handle/provisioning fields can clear on refresh.
- Replace copied Bluesky settings view tests with real widget coverage for the claim notice, snackbar action, and return refresh paths.

## Verification
- flutter test test/screens/settings/bluesky_settings_screen_test.dart test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart
- flutter test test/blocs/crosspost_settings/
- pre-push checks: analyzer and changed-file tests

Closes #6201

### DIFF
--- a/mobile/lib/blocs/crosspost_settings/crosspost_settings_cubit.dart
+++ b/mobile/lib/blocs/crosspost_settings/crosspost_settings_cubit.dart
@@ -30,13 +30,13 @@ class CrosspostSettingsCubit extends Cubit<CrosspostSettingsState> {
     try {
       final result = await _apiClient.getStatus();
       emit(
-        state.copyWith(
+        CrosspostSettingsState(
           status: CrosspostSettingsStatus.loaded,
           enabled: result.crosspostEnabled,
           username: result.username,
           handle: result.handle,
           provisioningState: result.provisioningState,
-          clearError: true,
+          attempt: state.attempt,
         ),
       );
     } catch (e, stackTrace) {
--- a/mobile/lib/screens/settings/bluesky_settings_screen.dart
+++ b/mobile/lib/screens/settings/bluesky_settings_screen.dart
@@ -1,6 +1,8 @@
 // ABOUTME: Bluesky crosspost settings screen with toggle switch
 // ABOUTME: Allows users to enable/disable publishing videos to Bluesky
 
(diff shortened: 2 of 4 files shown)


dcadenas (approved):
✅ NIP-05 status now stays in sync

rabble:
Post-merge review — the fix is correct (fresh `CrosspostSettingsState(...)` construction covers all 7 state fields, preserves `attempt`, implicitly clears `error`, and both claim-flow entry points now refresh with a proper `context.mounted` guard). Two small follow-ups spotted:

1. **`toggleCrosspost` success path still uses `copyWith`** with `username`/`handle`/`provisioningState` from the response. If keycast ever omits those on a `setCrosspost` response, the stale values persist — the exact bug class `loadStatus` just fixed. Low likelihood, but the two paths are now inconsistent.

2. **Snackbar "Set up" action can outlive the screen.** The `SnackBarAction.onPressed` closure calls `context.read<CrosspostSettingsCubit>()` on the listener's context. If the user pops the settings screen and then taps the still-visible snackbar (~4s window), that's a lookup on a deactivated element and throws. Capturing the cubit when the snackbar is shown (like `messenger` already is) would close it.

Issue #6271, opened 2026-07-21
refactor(router): replace normalization exemptions with consumed-segment contract


NotThatKindOfDrLiz:
## Problem

Route normalization still needs a small hand-maintained exemption list for GoRouter paths that parse as a parent route but rebuild shorter, such as `/video/:id/likers`, `/video/:id/reposters`, `/apps/:slug/sandbox`, and nested onboarding/review paths.

## User impact

When a new nested route is added without updating the exemption list, `routeNormalizationProvider` can redirect users away from the intended screen immediately after navigation.

## Acceptance criteria

- `parseRoute`/normalization can distinguish fully consumed routes from paths with unmodeled trailing segments.
- Parse-but-build-shorter paths are skipped structurally instead of through a per-path exemption list.
- Existing query/fragment skip behavior remains intact.
- Regression tests cover nested route families without relying on a hand-maintained deny-list.

## Technical notes

Follow-up from #6205. The query/fragment and unknown-route failure modes were fixed there; this issue tracks the remaining exemption-list category only.

PR #6272, opened 2026-07-21
fix(router): harden query route normalization

### LINKED ISSUE #6205: refactor(router): harden route normalizer against query routes
## Context

PR #6204 fixed #6203 by adding `/pooled-video-feed?video=...` to `shouldSkipRouteNormalization`, preventing the route normalizer from rewriting New video pooled-feed taps to `/home/0`.

That PR was the right narrow fix, but it exposed a broader route-normalizer weakness.

## Problem

`parseRoute` currently splits the raw path on `/` before stripping or parsing the query string:

```dart
final segments = path.split(/).where((s) => s.isNotEmpty).toList();
final firstSegment = segments[0];
```

For a single-segment route with a query, such as:

```text
/pooled-video-feed?video=<full-event-id>
```

`firstSegment` becomes `pooled-video-feed?video=<full-event-id>`, so it does not match `case 'pooled-video-feed':` and falls through to the default home route. The normalizer then rebuilds that fallback as `/home/0`.

This pattern has already produced multiple query-route exemptions in `shouldSkipRouteNormalization`, including reset password, email verification, and pooled feed.

There is a second constraint for pooled feed specifically: `buildRoute(RouteContext(type: RouteType.pooledVideoFeed))` returns the bare `/pooled-video-feed` path and cannot preserve the selected `?video=` identity. That means the route remains lossy even if query stripping is fixed unless pooled-feed identity becomes part of the modeled route context or stays explicitly exempt.

## Desired outcome

Make route normalization fail closed instead of silently rewriting valid but imperfectly-modeled routes to `/home/0`.

Potential approaches:

- Strip or parse query strings before `parseRoute` switches on the first path segment.
- Invert `shouldSkipRouteNormalization` from a deny-list of fragile routes to an allow-list of routes known to round-trip losslessly.
- Add route round-trip assertions for registered routes, including query-bearing routes, with explicit exemptions for routes intentionally outside the parse/build contract.
- Decide whether pooled-feed `?video=` should be modeled in `RouteContext`/`buildRoute` or remain an explicit normalizer exemption because the full feed still travels through GoRouter `extra`.

## Acceptance criteria

- Single-segment routes with query strings are not silently normalized to `/home/0` unless that is intentionally expected.
- Existing query routes such as reset password, email verification, and pooled feed keep their required query parameters through navigation/normalization.
- Tests document which routes are expected to round-trip through `parseRoute -> buildRoute` and which routes are intentionally exempted.
- No Nostr IDs are truncated in tests, logs, or route fixtures.

NotThatKindOfDrLiz:
## Summary

- strip query strings and fragments before route parsing so query-bearing paths do not contaminate path segments
- add a nullable `parseKnownRoute` path for normalization so unknown or incomplete routes fail closed instead of rewriting to `/home/0`
- skip query/fragment routes structurally and document the remaining parse-but-build-shorter exemptions
- cover pooled feed, people-list create query routes, malformed people-list IDs, and the normalizer skip/classification table

Closes #6205
Follow-up: #6271

## Verification

- `flutter test test/router/route_coverage_test.dart test/router/route_normalization_test.dart`
- `flutter analyze`
- pre-push hook: analyzer and changed router tests passed

### DIFF
--- a/mobile/lib/router/providers/page_context_provider.dart
+++ b/mobile/lib/router/providers/page_context_provider.dart
@@ -192,7 +192,7 @@ bool _isKnownRouteShape(List<String> segments) {
                   segments[1] == 'message-requests')) ||
           (length == 2 && segments[1] == 'message-requests');
     case 'hashtag':
-      return length == 2 || length == 3;
+      return length == 2;
     case 'categories':
       return length == 2;
     case 'video-editor':
@@ -214,9 +214,7 @@ bool _isKnownRouteShape(List<String> segments) {
     case 'list':
       return length == 2 || length == 3;
     case 'people-lists':
-      return (length == 2 && segments[1] == 'new') ||
-          length == 2 ||
-          (length == 3 && segments[2] == 'add-people');
+      return length == 2 || (length == 3 && segments[2] == 'add-people');
     case 'nostr-settings':
       return length == 1 ||
           (length == 2 && segments[1] == Nip05SettingsScreen.subpath);
@@ -245,7 +243,6 @@ bool _isKnownRouteShape(List<String> segments) {
     case 'legal':
     case 'bluesky-settings':
     case 'edit-profile':
-    case 'setup-profile':
     case 'clips':
     case 'clips-no-sound':
     case 'clips-only':
@@ -265,7 +262,7 @@ bool _isKnownRouteShape(List<String> segments) {
 /// Parse a URL path into a structured RouteContext
 /// Normalizes negative indices to 0 and decodes URL-encoded parameters
 RouteContext parseRoute(String path) {
-  return parseKnownRoute(path) ??
+  return _parseRoute(path, knownOnly: false) ??
       const RouteContext(type: RouteType.home, videoIndex: 0);
 }
 
@@ -275,6 +272,10 @@ RouteContext parseRoute(String path) {
 /// incomplete paths return null so GoRouter can handle them instead of being
 /// rewritten to home.
 RouteContext? parseKnownRoute(String path) {
+  return _parseRoute(path, knownOnly: true);
+}
+
+RouteContext? _parseRoute(String path, {required bool knownOnly}) {
   final pathOnly = path.split('#').first.split('?').first;
   final segments = pathOnly.split('/').where((s) => s.isNotEmpty).toList();
 
@@ -282,7 +283,7 @@ RouteContext? parseKnownRoute(String path) {
     return const RouteContext(type: RouteType.home, videoIndex: 0);
   }
 
-  if (!_isKnownRouteShape(segments)) {
+  if (knownOnly && !_isKnownRouteShape(segments)) {
     return null;
   }
 
@@ -304,7 +305,7 @@ RouteContext? parseKnownRoute(String path) {
 
     case 'profile':
       if (segments.length < 2) {
-        return null;
+        return knownOnly ? null : const RouteContext(type: RouteType.home);
       }
       final npub = _safeDecode(segments[1]); // Decode URL encoding
       // Grid mode (no index) vs feed mode (with index)
@@ -361,7 +362,7 @@ RouteContext? parseKnownRoute(String path) {
 
     case 'hashtag':
       if (segments.length < 2) {
-        return null;
+        return knownOnly ? null : const RouteContext(type: RouteType.home);
       }
       final tag = _safeDecode(segments[1]); // Decode URL encoding
       final rawIndex = segments.length > 2 ? int.tryParse(segments[2]) : null;
@@ -374,7 +375,7 @@ RouteContext? parseKnownRoute(String path) {
 
     case 'categories':
       if (segments.length < 2) {
-        return null;
+        return knownOnly ? null : const RouteContext(type: RouteType.home);
       }
       final categoryName = _safeDecode(segments[1]);
       return RouteContext(
@@ -517,14 +518,14 @@ RouteContext? parseKnownRoute(String path) {
       return const RouteContext(type: RouteType.developerOptions);
     case 'following':
       if (segments.length < 2) {
-        return null;
+        return knownOnly ? null : const RouteContext(type: RouteType.home);
       }
       final followingPubkey = _safeDecode(segments[1]);
       return RouteContext(type: RouteType.following, npub: followingPubkey);
 
     case 'followers':
       if (segments.length < 2) {
-        return null;
+        return knownOnly ? null : const RouteContext(type: RouteType.home);
       }
       final followersPubkey = _safeDecode(segments[1]);
       return RouteContext(type: RouteType.followers, npub: followersPubkey);
@@ -533,7 +534,7 @@ RouteContext? parseKnownRoute(String path) {
       return const RouteContext(type: RouteType.videoFeed);
     case 'list':
       if (segments.length < 2) {
-        return null;
+        return knownOnly ? null : const RouteContext(type: RouteType.explore);
       }
       // Web-canonical deep-link shape /list/:pubkey/:listId — the author
       // key rides in [RouteContext.npub] so buildRoute can reconstruct the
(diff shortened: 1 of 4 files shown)


mobile/lib/router/providers/page_context_provider.dart
```diff
@@ -168,15 +168,124 @@ String _safeDecode(String segment) {
...
+    case 'profile':
+      return length == 2 || length == 3;
+    case 'inbox':
+      return length == 1 ||
+          (length == 3 &&
+              (segments[1] == 'conversation' ||
+                  segments[1] == 'message-requests')) ||
+          (length == 2 && segments[1] == 'message-requests');
+    case 'hashtag':
+      return length == 2 || length == 3;
```
realmeylisdev: nit: the shape admits `hashtag` at length 3, but `buildRoute` for `RouteType.hashtag` ignores `videoIndex` — so `/hashtag/dogs/3` normalizes to `/hashtag/dogs` and the parsed index is dead. Lands on the same tag so it's harmless, but there's no `/hashtag/:tag/:index` route, so `length == 2` would match the gate's own "leave it to GoRouter" intent.

mobile/lib/router/providers/page_context_provider.dart
```diff
@@ -168,15 +168,124 @@ String _safeDecode(String segment) {
...
+    case 'following':
+    case 'followers':
+    case 'sound':
+    case 'original-sound':
+    case 'profile-view':
+      return length == 2;
+    case 'list':
+      return length == 2 || length == 3;
+    case 'people-lists':
+      return (length == 2 && segments[1] == 'new') ||
```
realmeylisdev: nit: `(length == 2 && segments[1] == 'new')` is fully subsumed by the `length == 2` on the next line — it can never change the result. The `new` distinction is handled in the switch, not here, so this reads as if `new` is specially gated at the shape level when it isn't.

mobile/lib/router/providers/page_context_provider.dart
```diff
@@ -168,15 +168,124 @@ String _safeDecode(String segment) {
...
+    case 'general-settings':
+    case 'storage-management':
+    case 'invites':
+    case 'app-language':
+    case 'appearance-settings':
+    case 'support-center':
+    case 'legal':
+    case 'bluesky-settings':
+    case 'edit-profile':
+    case 'setup-profile':
```
realmeylisdev: Heads up, not a blocker and pre-existing (identical on `main`): the gate admits `setup-profile`, the switch collapses it to `editProfile`, and `buildRoute` rebuilds it as `/edit-profile`. So the normalizer rewrites `/setup-profile` (`isNewUser: true`, reached from the "Complete Your Profile" nudge) into `/edit-profile` (`isNewUser: false`) and turns the `push` into a stack-resetting `go`. That's exactly the "shortened into a different valid route" case this gate's doc says it guards against — worth folding into #6271 (or a separate fix) since it's the one modeled family that parses to a different route than it builds.

realmeylisdev (approved):
Nice fix. Fail-closed on unknown/incomplete paths plus a structural query/fragment skip is the right shape, and it cleanly kills the `/pooled-video-feed?video=` → `/home/0` class from #6205.

Checked the new allow-list against every `parseRoute` switch case (all covered, no dead cases) and confirmed every modeled family round-trips through `buildRoute` — the normalizer can't send anyone to a different screen. 112 router tests + `flutter analyze` green locally, and all CI is green.

Left a couple of allow-list nits inline, plus one pre-existing `setup-profile` note that fits the #6271 follow-up. None blocking. LGTM.

NotThatKindOfDrLiz:
Takeover update pushed in 83da1f25e.

What changed:
- split page-context parsing back out from strict normalizer parsing so valid unmodeled GoRouter paths like /explore/tab/popular still report RouteType.explore to tab/back/readiness state
- kept parseKnownRoute fail-closed for routes the normalizer cannot round-trip, including /explore/tab/:name and /setup-profile
- cleaned up the dead people-lists shape clause and tightened hashtag normalization to the actual /hashtag/:tag GoRoute shape
- refreshed the normalizer skip-list comment to match the current fail-closed behavior
- added regression coverage for /explore/tab/popular, /setup-profile, and unmodeled hashtag suffixes

Verification:
- flutter test test/router/route_coverage_test.dart test/router/route_normalization_test.dart
- flutter analyze
- pre-push hook: analyzer and changed router tests passed

NotThatKindOfDrLiz:
## Route Coverage and Shape Validation Review

### 🛑 **Critical Regression** — `/explore/tab/:name` Is Broken

**Finding 1:** The allow-list gate in `_isKnownRouteShape()` only permits explore family routes with `length <= 2`:

```dart
case RouteFamily.explore:
  return segments.length <= 2;
```

But the explore router **defines a tab route** that requires 3 segments:

```dart
@TypedGoRoute<ExploreTabRoute>(
  path: 'explore/tab/:name',
)
class ExploreTabRoute extends GoRouteData {
```

Navigating to `/explore/tab/trending` will parse as `null` instead of `ExploreTabRoute`. This is a regression — the route exists and is used in production.

**Fix:** Change the gate to:
```dart
case RouteFamily.explore:
  return segments.length <= 3;
```

**Why this matters:** This violates `agent_workflow.md` § "Do not accumulate technical debt": a known broken route should fail the build, not ship silently. The PR explicitly removes fallback behavior from unknown routes — that audit is lost if we gate out a known, used route.

---

### ⚠️ **Technical Debt** — `/setup-profile` Route Mismatch

**Finding 2:** The allow-list permits `/setup-profile` (2 segments), but the actual route builds as `/setup-profile-initial`:

```dart
@TypedGoRoute<SetupProfileRoute>(
  path: 'setup-profile-initial',
)
```

A user deep-link or direct URL to `/setup-profile` will parse to `null` today, silently falling back to home. This PR makes that fallback explicit instead of fixing the mismatch. Neither the allow-list nor the route definition is wrong — they just disagree.

**Guidance:** Either rename the route to `setup-profile` (matches the allow-list) or update the gate to `setup-profile-initial` (matches the code). The PR should not ship with both naming conventions in play.

---

### ℹ️ **Dead Code** — Redundant People-Lists Shape Check

**Finding 3:** The people-lists condition repeats the segment-length check inside a query-check branch:

```dart
case RouteFamily.peopleList:
  if (segments.length <= 2) {
    return segments.length <= 2;  // Tautology
  }
  return false;
```

This simplifies to:

```dart
case RouteFamily.peopleList:
  return segments.length <= 2;
```

Minor code-cleanliness issue — doesn't affect correctness.

---

### ℹ️ **Incomplete Modeling** — Hashtag Route Allows Invalid Shape

**Finding 4:** The hashtag family gate permits `length == 3`:

```dart
case RouteFamily.hashtag:
  return segments.length == 2 || segments.length == 3;
```

But no hashtag route is defined with 3 segments. The only hashtag route is:

```dart
@TypedGoRoute<HashtagRoute>(
  path: 'discover/hashtags/:hashtag',
)
```

This allows `/discover/hashtags/x/y` (3 segments) to pass the gate, then parse as `null` at runtime. Either:
- Remove the `length == 3` case (it's unreachable), or
- Define a hashtag sub-route that uses 3 segments.

---

### ℹ️ **Inaccurate Documentation** — Skip-List Comments Obsolete

**Finding 5:** Route normalization comments reference specific families as a "skip-list," but the code uses structural checks:

```dart
// From PR: Explicit family skip-list (commented in old code)
// → Actual implementation: `if (loc.contains('?') || loc.contains('#')) return true;`
```

This is correct behavior (URLs with query strings bypass normalization), but the code comment no longer describes how it works. Update the comment to reflect the structural check.

---

### Summary

- **1 critical:** Explore tab route gated out; regression from current behavior
- **1 technical debt:** Setup-profile naming mismatch
- **3 nits:** Dead code, incomplete modeling, stale docs

The PR should not merge until Finding 1 is resolved. Findings 2–5 are cleanup opportunities that improve consistency.

PR #6279, opened 2026-07-21
fix(moderation): parse repeated label event tags

### LINKED ISSUE #6088: fix(moderation): ModerationLabelService content-warning parser has 4 spec-compliance gaps
## Summary
`ModerationLabelService._processLabelEvent` (`mobile/lib/services/moderation_label_service.dart:401-487`), which parses NIP-32 kind-1985 label events for the `content-warning` namespace, has four spec-compliance gaps found as a byproduct of investigating #5952. Currently invisible in production since zero `content-warning` events are published today (see #5952 findings), but would bite as soon as that pipeline goes live.

## Environment
- App version: current `main`
- Applies to: `mobile/lib/services/moderation_label_service.dart`

## The four gaps

1. **Omitted `l`-mark event dropped.** NIP-32 allows an unmarked `l` tag when `L` is present at the event level, but the parser only sets `labelValue` inside the `l` case when `tag.length > 2 && tag[2] == 'content-warning'` (L428). An event with `["L","content-warning"]` and a bare `["l","nudity"]` (no third element) is silently dropped at L453 (`if (!isContentWarning || labelValue == null) return;`).
2. **Multi-label collapse.** Only the last matching `l` tag's value survives per event — the `switch` case at L427-441 overwrites `labelValue` on each match rather than collecting all of them. An event labeling both `nudity` and `violence` in separate `l` tags loses all but the last.
3. **Case/format sensitivity.** The namespace match at L424 (`tagValue == 'content-warning'`) and L428 (`tag[2] == 'content-warning'`) is exact-string. `Content-Warning`, `content_warning`, or other case/format variants from spec-minimal external publishers are silently dropped.
4. **`r`/`t` targets ignored** (documented as intentional, not a gap — noting here for completeness since it was reviewed alongside the others). `x` (content hash) is a Divine/NIP-94 extension and correctly handled.

## Expected Result
The parser should accept the omitted-mark case, collect multiple `l` values per event, and match the namespace case-insensitively (or per whatever the team decides is the right normalization).

## Regression?
No — likely present since the parser was first written; just untested against real-world spec-minimal publishers because there's no live traffic to catch it.

## Evidence
Found via code read + a NIP-32 spec-compliance pass during #5952's investigation (`tasks/findings_5952.md`, "Secondary parser gaps" section). No live repro currently possible since production publishes zero `content-warning` events — regression tests should construct synthetic kind-1985 events covering each gap.

Related: #5952 (adjacent investigation, not a duplicate — that issue is about the closed-Pool crash/retry behavior, not the parser).

NotThatKindOfDrLiz:
## Summary
- Parse all content-warning labels in kind-1985 events instead of overwriting repeated l tags
- Index all repeated e/a/p/x targets with one ModerationLabel per label-target pair
- Trim and case-fold the content-warning namespace while preserving spec boundaries for other namespace spellings
- Add a NostrEventKinds.label constant and use it for kind-1985 publishing/querying

Closes #6088

## Verification
- flutter test test/services/moderation_label_service_test.dart
- flutter test test/services/effective_content_labels_test.dart test/services/nsfw_content_filter_test.dart test/services/video_moderation_status_service_test.dart test/services/video_event_service_content_filter_test.dart

### DIFF
--- a/mobile/lib/constants/nostr_event_kinds.dart
+++ b/mobile/lib/constants/nostr_event_kinds.dart
@@ -13,4 +13,7 @@ class NostrEventKinds {
 
   /// Kind 7: Reaction (NIP-25)
   static const int reaction = 7;
+
+  /// Kind 1985: Labeling (NIP-32)
+  static const int label = 1985;
 }
--- a/mobile/lib/services/account_label_service.dart
+++ b/mobile/lib/services/account_label_service.dart
@@ -4,6 +4,7 @@
 import 'dart:async';
 
 import 'package:nostr_client/nostr_client.dart';
+import 'package:openvine/constants/nostr_event_kinds.dart';
 import 'package:openvine/models/content_label.dart';
 import 'package:openvine/services/auth_service.dart';
 import 'package:shared_preferences/shared_preferences.dart';
@@ -131,7 +132,7 @@ class AccountLabelService {
       ];
 
       final event = await _authService.createAndSignEvent(
-        kind: 1985,
+        kind: NostrEventKinds.label,
         content: '',
         tags: tags,
       );
--- a/mobile/lib/services/moderation_label_service.dart
+++ b/mobile/lib/services/moderation_label_service.dart
@@ -7,6 +7,7 @@ import 'dart:convert';
 import 'package:nostr_client/nostr_client.dart';
 import 'package:nostr_sdk/filter.dart';
 import 'package:nostr_sdk/nip05/nip05_validor.dart';
+import 'package:openvine/constants/nostr_event_kinds.dart';
 import 'package:openvine/services/auth_service.dart';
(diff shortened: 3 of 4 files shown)


mbradley (approved):
approving. one non-blocking note.

this makes `getContentWarnings`/`getLabelsForPubkey` return N labels per event, which is the "direct consumer surfaces duplicates" case flagged on #5596. i traced the subscribe path (`_loadedLabelers` plus the synchronously-set `_loadingLabelers` in-flight guard, single-shot `queryEvents`, and `_unloadLabeler` clearing state before re-subscribe) and there's no live double-processing path, so it's correct as-is.

open question is just whether a `_processLabelEvent`-level dedup is worth landing to stay robust if a live subscription path gets wired up later. judgment call, fine as a follow-up pending that decision. not blocking.

PR #6286, opened 2026-07-21
feat(dm): messages UX revision — unread emphasis, filters, search, error state

### LINKED ISSUE #6293: feat(dm): messages UX revision — unread emphasis, filters, search, error state
Auto-created from PR #6286.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6286
**Branch:** `feat/messages-unread-ux`

rabble:
Closes #6293

## Description

Messages UX revision for inbox parity with common DM patterns:

### Unread visibility + honest failure states
- **Unread emphasis** — unread rows render their message preview white + semibold alongside the unread dot, so unread threads scan at a glance.
- **All/Unread filter** — chip row above the Messages list. Filtering lives in `ConversationListBloc`; chronological order is preserved within the filter.
- **Error != empty** — failed list load renders a retryable `InboxErrorState` instead of the misleading empty inbox state.

### Inbox conversation search
- `DivineSearchBar` above the Messages list, backed by debounced `ConversationListSearchQueryChanged`.
- Matches **counterparty display name** and **last-message preview** over the complete accepted inbox set. Search composes with the unread filter.
- Applies the shared `minSearchQueryLength` gate before doing profile-resolution work, matching the other search blocs.
- Resolves counterparty profile names progressively in bounded chunks. Newly streamed conversations can trigger profile re-resolution without re-enqueuing a stale user query.
- `ProfileRepository` is injected nullable via `profileRepositoryProvider`; late null -> instance changes are delivered to the existing bloc with `ConversationListProfileRepositoryChanged` so the inbox subtree is not torn down.
- Distinct "No matches" state when a search finds nothing.

l10n: new inbox/search/error keys added to `app_en.arb`, translated into the other locales, and generated localizations committed.

## Out of Scope

Swipe-to-reply for text messages, group-chat classification fix, pinned conversations, synced mute — later slices of the same UX review.

## Verification

- `flutter test test/blocs/dm/conversation_list/conversation_list_bloc_test.dart` — passed locally
- `flutter analyze lib test` — passed locally
- Pre-push hook — merge-conflict check, analyze, and changed-file tests passed
- GitHub checks on `fd6f720f89fc20bf8e2d734f00d6b391fd5a7f3b` — all green

## Type of Change


mobile/lib/screens/inbox/inbox_view.dart
```diff
@@ -547,46 +563,56 @@ class _ConversationListState extends ConsumerState<_ConversationList>
...
+                  hasRequests: hasRequests,
+                  requestUnreadCount: requestUnreadCount,
+                  onOpenRequests: () => _openMessageRequests(context),
+                )
+              : _ConversationListView(
+                  scrollController: _scrollController,
+                  conversations: visibleConversations,
+                  hasRequests: hasRequests,
+                  requestUnreadCount: requestUnreadCount,
+                  hasMore: hasMore,
```
realmeylisdev: **Stranded load-more spinner under an active filter/search.** `hasMore` here is the full accepted-list flag (bloc sets it from `accepted.length == currentLimit`, and the filter/search handlers don't touch it), but this view renders the filtered `visibleConversations`. So with a long history loaded (`hasMore` true) and an Unread filter or search that matches only a few rows, `_ConversationListView` still appends the trailing spinner (`itemCount ... + (hasMore ? 1 : 0)`, L805 / L816) — but the short filtered list can't scroll, and `onLoadMore` only fires from the scroll listener, so it never triggers. The spinner spins forever until the filter is cleared.

Repro: load ≥20 conversations, toggle **Unread** with only 1–2 unread.

Load-more grows the *unfiltered* accepted list anyway, so pagination-by-scroll is meaningless while a client-side filter is active — suppressing it fixes both, e.g. `hasMore: hasMore && !unreadOnly && searchQuery.isEmpty`.

mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
```diff
@@ -193,6 +203,13 @@ class ConversationListBloc
         return state.copyWith(
           status: ConversationListStatus.loaded,
           conversations: visibleInbox,
+          visibleConversations: _applyFilters(
```
realmeylisdev: Minor: during an active **name** search, a conversation that newly streams in here is filtered against the `profileNames` resolved when the search event fired. A brand-new counterparty isn't in that map, so `_applyFilters` falls back to the generated name and a real-name match is dropped — even though the tile will render the real name. It self-heals on the next keystroke, so low severity; resolving unresolved counterparties on the data path (or keying resolution off the incoming set) would close it.

mobile/lib/screens/inbox/widgets/conversation_tile.dart
```diff
@@ -144,6 +144,7 @@ class ConversationTile extends ConsumerWidget {
                             context,
                             conversation.lastMessageContent!,
                           ),
+                          emphasized: !conversation.isRead,
```
realmeylisdev, line 152: Nice touch strengthening the unread cue visually. One gap: unread stays visual-only — the tile's `Semantics` label (L72) reads identically for read/unread, so VoiceOver/TalkBack users can't tell them apart. The notification rows already prepend `notificationsUnreadPrefix` ("Unread notification"); mirroring that here would give AT users the same signal this PR adds visually. Pre-existing and adjacent to the diff, so optional — but on-theme.

realmeylisdev (commented):
Nice slice — filter/search logic lives in the bloc, the error state is honest, l10n covers all 18 locales, and the new tests are behavioral (94 green + analyze clean locally). One thing worth a fix before merge: the load-more spinner is keyed off the full-list `hasMore` while the list renders the *filtered* set, so an active Unread/search filter can strand a spinner that never resolves. Details + two small nits inline.

mobile/lib/screens/inbox/inbox_view.dart
realmeylisdev: Fixed in 43248bf78: pagination is now suspended while a client-side filter/search is active (new `isFiltering` getter gates both the trailing spinner and `canLoadMore`), so a short filtered result no longer strands a spinner. Added widget tests for spinner suppressed-while-filtering / shown-otherwise.

mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
realmeylisdev: Fixed in 43248bf78: the stream data-path now re-runs name resolution when a not-yet-resolved counterparty appears during an active search, so a just-arrived match surfaces on its own (converges once cached). Added a bloc test covering the restream case.

mobile/lib/screens/inbox/widgets/conversation_tile.dart
realmeylisdev: Fixed in 43248bf78: unread rows now prefix the unread status to their semantic label (new `inboxConversationTileLabelUnread`, translated across all 18 locales), mirroring the notification rows, so assistive tech gets the same signal as the dot/emphasis. Added read/unread label tests.

realmeylisdev (approved):
Earlier review findings are addressed in 43248bf78 — load-more spinner suppressed while a filter/search is active (`isFiltering` gate), search re-resolves names on the data path, and unread rows now carry an unread semantic label across all 18 locales, each with tests. LGTM. Heads-up: the added l10n strings are machine-translated — swap for reviewed copy if you'd prefer.

mobile/lib/screens/inbox/inbox_page.dart
```diff
@@ -80,10 +81,16 @@ class InboxPage extends ConsumerWidget {
         )),
         providers: [
           BlocProvider(
+            // Re-keyed on the nullable-gated profileRepository alone (it
+            // resolves null -> instance shortly after mount, same shape as
+            // reportingService below) so search name-resolution binds the
+            // ready instance without tearing down every inbox bloc.
```
NotThatKindOfDrLiz: This provider is the first entry in MultiBlocProvider, which makes it the outer ancestor of the rest of the providers and InboxView. Because profileRepositoryProvider legitimately flips null -> instance during cold start, this key change tears down the whole inbox subtree: notification badge cubit, following bloc, mute/actions cubits, and view state. The comment claims the opposite. Please narrow this rebuild so only the conversation bloc dependency is refreshed, or move/restructure the keyed provider so it does not recreate unrelated inbox state.

mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
```diff
@@ -190,9 +200,31 @@ class ConversationListBloc
...
+        if (state.searchQuery.isNotEmpty &&
+            visibleInbox.any(
+              (c) => !state.profileNames.containsKey(
+                _otherParticipant(c, userPubkey),
+              ),
+            )) {
+          add(ConversationListSearchQueryChanged(state.searchQuery));
+        }
+
         return state.copyWith(
```
NotThatKindOfDrLiz, line 257: Filtering is still applied to visibleInbox, which comes from watchAcceptedConversations(limit: state.currentLimit) and starts at pageSize 20. That means Unread and search are only complete for the currently loaded window. Since the unread badge counts accepted conversations unpaginated, a user can see a Messages unread badge while this filtered list says they are caught up; search can likewise report no matches for conversations past the loaded page. The third commit hides the spinner/dead-end, but it does not fix the false completeness semantics.

mobile/lib/screens/inbox/inbox_view.dart
```diff
@@ -461,7 +463,11 @@ class _ConversationListContent extends StatelessWidget {
       ConversationListStatus.loading => const Center(
         child: CircularProgressIndicator(color: VineTheme.primary),
       ),
-      ConversationListStatus.error => const InboxEmptyState(),
+      ConversationListStatus.error => InboxErrorState(
```
NotThatKindOfDrLiz, line 466: Retry from the error state still has no visible in-progress transition. _onStarted only emits loading when the bloc is initial, so from error this tap leaves the UI on the same error state while the stream resubscribes; if it fails again, copyWith(status: error) can be equivalent to the current state and the screen appears frozen. Please add a retry/loading signal for this path.

mobile/lib/screens/inbox/inbox_view.dart
```diff
@@ -481,6 +487,7 @@ class _ConversationList extends ConsumerStatefulWidget {
 class _ConversationListState extends ConsumerState<_ConversationList>
     with ScrollPaginationMixin {
```
NotThatKindOfDrLiz, line 488: The search query lives in the bloc, but this controller is recreated empty whenever _ConversationList unmounts, for example error -> retry -> loaded. That can leave the list filtered by a stale bloc searchQuery while the field shows empty text. Seed the controller from state.searchQuery and keep it synchronized, or clear the bloc query when leaving loaded.

NotThatKindOfDrLiz (changes requested):
Requesting changes on the remaining current-head issues. The latest commit fixes the endless spinner by suppressing pagination while filters are active, but the filtered views still operate on a partial loaded page, and a profileRepository key on the first provider can still tear down the inbox subtree during cold-start readiness changes.

mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
realmeylisdev: Fixed in ba128efd2. Stream 1 is now unpaginated (`watchAcceptedConversations()` — the DAO only applies a LIMIT when one is passed, so no package change), and pagination became a render window applied in a new `_computeVisible`: filtered views run over the complete set, unfiltered views get sliced to `currentLimit`. So a filter result is always complete and the empty-state copy is true.

Knock-on effects: `hasMore` is now exact (`conversations.length > currentLimit`) instead of the `accepted.length == currentLimit` truncation heuristic; load-more is a synchronous re-slice with no stream restart, which also stops it re-firing `backfillHistoryIfNeeded`/`retryPendingDecryptions` on every page; `isLoadingMore` is gone (it can no longer be true, and `ScrollPaginationMixin` already latches re-entrancy).

Search now reads the full set, so `fetchBatchProfiles` — which falls back to one parallel relay request per uncached pubkey — is chunked (`searchProfileResolveChunkSize`) with progressive emits: the query lands immediately on previews/cached/fallback names, then refines per chunk, abandoning remaining chunks when a newer query supersedes it.

Three regression tests, each confirmed to fail against the paginated watch. Worth noting the shared `_stubStreams` helper ignores `limit`, so they use a stub that actually truncates — otherwise they'd have passed against the bug.

mobile/lib/screens/inbox/inbox_view.dart
realmeylisdev: Fixed in ba128efd2. `_onStarted` now switches on status and emits `loading` from `error` as well as `initial`; `loading`/`loaded` still no-op so `_onBlocklistChanged` re-entry can't flash a spinner over a loaded list.

Confirmed the suppression you described: with status already `error`, `copyWith(status: error)` is Equatable-equal to the current state, so `emit` dropped it and the retry produced *zero* state transitions. The slow-success case was the same screen for the whole in-flight window.

`currentLimit` is deliberately not reset on retry — after the pagination change it's a render window, so resetting it would throw away the position the user had scrolled to. Two tests: one asserting `[loading, error]` on a repeat failure, one pinning the preserved window.

mobile/lib/screens/inbox/inbox_view.dart
realmeylisdev: Fixed in ba128efd2 — seeded the controller from `state.searchQuery` in `initState`.

Went with seeding over clearing the bloc query for two reasons. It's remount-source-agnostic: `error -> retry -> loaded` is the only *status*-driven unmount (`loaded -> loading` can't happen, since `_onStarted` gates its loading emit), but the pre-existing `ValueKey(reportingService)` also re-inflates the view — a keyed provider replaces its whole downstream chain including the leaf child — and no bloc-side status hook reaches that path. Clearing the query would also fight the streamed-in re-resolve loop that re-dispatches the current query from `onData`.

No `BlocListener` write-back on purpose: `_onSearchQueryChanged` trims, so an unconditional write-back would eat a trailing space and stomp the cursor mid-typing. While mounted the field is the only writer, so drift can only come from a remount.

Widget test covers type -> error -> retry, asserting the field is repopulated and the search bar genuinely unmounted in between (so it can't pass trivially). Verified red without the fix.

mobile/lib/screens/inbox/inbox_page.dart
realmeylisdev: Fixed in ba128efd2 — the key is gone. The ready instance is now delivered in place via a `ConversationListProfileRepositoryChanged` event from `_InboxRepositorySync` (renamed from `_InboxNotificationBadgeRepositorySync`, since it now forwards both late-resolving repos), so only the conversation bloc's dependency is refreshed. That's the narrowing you asked for, and it drops the stale `profileNames` so an active query re-resolves against the ready instance.

You're right that the comment claimed the opposite of what the placement does, and right about the mechanism: `providers[0]` is the outermost entry (`nested` builds the chain from `_children.reversed`), so a key there fails `canUpdate` and unmounts everything below it.

Two things I found while verifying that are worth flagging:

**Moving the key wouldn't have worked.** A key on *any* provider in the chain re-inflates the leaf child too, so relocating it to a later slot would have stopped the sibling-bloc churn but still remounted `InboxView` — losing tab selection, scroll offset and search text. That's why this is an event rather than a relocation. It's also why the search-field fix in the sibling thread is written to survive any remount source: the pre-existing `ValueKey(reportingService)` on the last provider already remounts the view shortly after every inbox open.

**The teardown wasn't reaching users, though the key was still wrong.** `profileRepositoryProvider` and `dmRepositoryProvider` both gate on `nostrSessionProvider.isReadyForActiveClient`, and `dmRepository` is the first member of the pre-existing outer `MultiBlocProvider` key (#6046, on main). So the cold-start transition that flips `profileRepository` already replaced the whole subtree on the same frame — I couldn't construct a reachable path where the inner key flipped alone (`contentPolicyEngine` never rebuilds; the rest of its closure is shared with `dmRepository`). Isolating `profileRepository` in a test reproduces exactly what you describe, but that isolation isn't reachable in production. Not a defense of the line — it was redundant and its comment was false, so it's deleted; the event binding is now structural rather than relying on that coincidence.

Regression test asserts the sibling blocs (following, mute, badge, actions) and the `InboxView` State all survive a `profileRepository` null -> instance flip. Verified it fails with the key reinstated.

realmeylisdev:
@NotThatKindOfDrLiz all four addressed in ba128efd2 — replies on each thread.

- **Filters over a partial window** — the accepted watch is unpaginated now; pagination became a render window, so filter results are complete and the badge can't contradict the list. Load-more is a pure re-slice (no stream restart, no repeated backfill side effects). Search resolution is chunked with progressive emits since it now reads the full set.
- **Retry feedback** — `error -> loading` is a real transition, so a repeat failure is visible; the scrolled-to window is preserved.
- **Search controller** — seeded from bloc state, which covers every remount source rather than only error/retry.
- **The keyed provider** — deleted; the ready repository is delivered in place by event, refreshing only the conversation bloc.

Two notes worth your eye, both in the thread replies: relocating the key wouldn't have preserved `InboxView` state (a key re-inflates the leaf child too), and the teardown itself was already happening from the pre-existing outer key on the same session transition — the line was redundant rather than a new regression. Removed and replaced with a structural binding either way.

Each fix ships a regression test I confirmed fails without it. For the filter tests that meant a stub that actually honours `limit` — the shared helper ignores it, so they'd otherwise have passed against the bug.

rabble:
@NotThatKindOfDrLiz — your `CHANGES_REQUESTED` is the remaining gate here, and I think it's stale. All four of your threads were answered at `ba128efd2`, and @realmeylisdev approved that head. Re-review when you get a chance?

For the record, I re-read the diff against your four points and they do look genuinely closed:

- **Partial-window filtering** — stream 1 is now unpaginated (`watchAcceptedConversations()` with no limit) and pagination became a render window in `_computeVisible`. Filters run over the complete set, so "You're all caught up" and "No matches" are now true statements rather than statements about the loaded page.
- **Provider key tearing down the inbox subtree** — the key is gone; the ready instance arrives via `ConversationListProfileRepositoryChanged` from `_InboxRepositorySync`, and the handler drops the fallback `profileNames` so an active query re-resolves.
- **Retry with no visible transition** — `_onStarted` now switches on status and emits `loading` from `error`, while `loading`/`loaded` stay no-ops so `_onBlocklistChanged` re-entry can't flash a spinner over a loaded list.
- **Search controller recreated empty** — seeded from `state.searchQuery` in `initState`.

Three things I'd still like to resolve before this merges, none of them blocking:

**1. Nobody has measured the unpaginated stream.** Dropping the LIMIT is the right call for filter correctness, but every tick now maps, filters and sorts the complete accepted set and rebuilds two lists on the UI isolate. `DmUnreadCountCubit` already runs the same query app-shell-wide so the precedent exists, but that one produces a count, not two sorted lists rebuilt per tick. A frame timing on a heavy DM account would close this out.

**2. The re-resolve trigger fires from inside the `emit.forEach` mapper.** `add(ConversationListSearchQueryChanged(state.searchQuery))` is a side effect in what the surrounding code documents as a pure projection. It terminates today only because `_onSearchQueryChanged` writes a fallback name for *every* pubkey in the chunk even when `fetchBatchProfiles` throws — if that ever becomes conditional, the `hasNewUnresolved` guard stops converging and it loops on every recompute. Please pin the invariant with a test that fails the fetch and asserts `profileNames` still gains an entry per pubkey.

**3. `_profileRepository` is a mutable non-`final` field on a BLoC**, which `state_management.md` calls out. The `NotificationBadgeCubit.setRepository` precedent covers it and your reasoning against the `ValueKey` is right, so I'm not asking for a change — just noting it so the next reader doesn't think it slipped through.

Two nits: `InboxPage` uses `ref.watch(profileRepositoryProvider)` purely to seed the constructor while `ref.listen` handles every subsequent change, so `ref.read` would avoid a page rebuild on the null → instance flip. And @realmeylisdev's own flag stands — the 19 locales are machine-translated and should get reviewed copy before this is user-visible.

NotThatKindOfDrLiz:
@rabble I took over the remaining review fixes and pushed fd6f720f89fc20bf8e2d734f00d6b391fd5a7f3b.

What changed:
- Replaced the data-path search self-dispatch with an internal profile-resolution event, so newly streamed conversations can still resolve names without re-enqueuing an older debounced user query.
- Applied the shared minSearchQueryLength gate before inbox profile resolution, matching the other search blocs and avoiding one-character profile fan-out.
- Added regressions for short-query gating, failed fetch fallback-name convergence, and stale-query supersession during stream re-resolution.
- Rebasing onto current origin/main resolved the merge-conflict gate; the PR body is updated to match the current in-place ProfileRepository handoff.

Verification:
- flutter test test/blocs/dm/conversation_list/conversation_list_bloc_test.dart
- flutter analyze lib test
- Pre-push hook changed-file tests and analyze
- GitHub checks are all green on fd6f720f89fc20bf8e2d734f00d6b391fd5a7f3b

NotThatKindOfDrLiz (approved):
Approved after takeover. The previous requested-change blockers are closed, the remaining search-resolution issues are fixed in fd6f720f89fc20bf8e2d734f00d6b391fd5a7f3b, and CI is green.

NotThatKindOfDrLiz:
FIRST-PASS REVIEW: Messages UX Revision (PR 6286)

This PR implements the first slice of the Messages inbox redesign. CI is green. Code is well-structured with good test coverage.

CRITICAL ISSUES (must fix before merge):

1. Provider key placement (inbox_page.dart:86) - ValueKey on innermost BlocProvider doesn't update when Riverpod rebuilds. Move to outermost level in ConsumerWidget.build and use ref.watch instead of ref.read.

2. Filters on paginated window only (inbox_view.dart:631) - _applyFilters receives only current chunk. Shows false 'No results' when visible items don't match but unread ones exist beyond window. Store all fetched conversations separately, filter complete list, then paginate.

3. Pagination disabled under filters (inbox_view.dart:540-620) - Scroll listener only in else branch. With filters/search active, users cannot scroll to load more. Hoist listener outside conditional.

4. Trailing spinner spins forever (inbox_view.dart:815-820) - When 20 results fetched and 5 match filter, conversations.length stays same. Spinner condition remains true. Clear isLoadingMore flag in BLoC after results emit.

SHOULD-FIX ISSUES:

5. Search blocks on network call - No loading state while fetching names. Add searchStatus enum, emit loading before/after.

6. Retry button no feedback - No indicator while retrying. Emit loading status during attempt.

7. Search controller desync on remount - Controller recreated empty even though BLoC still holds searchQuery. Restore text from BLoC state.

Positive observations: Localization complete across 18 languages, 20+ tests with proper mocking, semantic accessibility labels, clean architecture.

Please address Critical issues 1-4 before merge.

realmeylisdev:
Heads up — this PR merged on 2026-07-24, three weeks before this review, and the findings describe an earlier revision of the branch rather than the merged code.

Items 2–7 were fixed in-branch before merge, in `7c81484`, `625e2bd` and `fd6f720` ("address inbox review findings — filter pagination, search restore", "complete inbox filters, retry feedback, and search restore", "harden inbox search resolution"). Checked each against current `main`:

- **2 — filters on the paginated window**: `watchAcceptedConversations()` is unpaginated now; pagination is a render window in `_computeVisible`, not a fetch bound.
- **3 — pagination under filters**: the list is a single `CustomScrollView` with the controller attached across every branch.
- **4 — trailing spinner**: gated on `hasMore && !isFiltering`.
- **5 — search blocking on the network**: resolution is progressive — it emits immediately on cached and deterministic-fallback names, then refines in bounded chunks, so there is no blocking wait to show a loading state for.
- **6 — retry feedback**: retry transitions `error → loading`.
- **7 — search controller desync**: the controller is seeded from `state.searchQuery`.

Item 1 doesn't match the code at any revision on this branch: those dependencies are read with `ref.watch`, and the `ValueKey` tuple is already on the outermost `MultiBlocProvider`. The inner key sat on `providers[0]` — the outermost entry in the nested chain, not the innermost — and was later removed in favour of `_InboxRepositorySync`, because keying it there re-inflated the whole subtree and lost tab selection, scroll offset and search text.

Nothing to action here — flagging it so the fixes don't get redone.

PR #1517, opened 2026-02-11
feat: Proofmode enhancement with C2PA support + hardware/software attestation

n8fr8:
Almost ready to go.... once https://github.com/divinevideo/divine-blossom/pull/4 is merged, in production, and testing is complete.

## Description

Initial integration of Proofmode Content Provenance and Authentication (C2PA) "Content Credentials" support into the mobile app, through our new c2pa-flutter library: https://github.com/guardianproject/c2pa-flutter

Should show "Fully Verified" badges on videos recorded with this support.

Will REQURE .env API key for release building to support remote server (ask n8fr8 or nathan@proofmode.org)

...with additional support for hardware and software attestation on Android. 

Signing currently done using self-signed local "hardware StrongBox" keys on Android, and remote server on iPhone (will support local keys soon there).

**Related Issue:** 

#183 
#838 

## Type of Change

- [ X ] ✨ New feature (non-breaking change which adds functionality)

Issue #838, opened 2026-01-06
Improve Proofmode / Media Provenance support for reliable "human made" filtering


n8fr8:
This is our broad ticket for tracking the work we are doing to improve media provenance and authentication (aka Proofmode aka C2PA.org) support in DiVine. We will be adding more specs, samples, code, links, specific issue and progress to this issue as we make our way forward.

1. Implement new Flutter library to wrap the C2PA https://github.com/contentauth/c2pa-android and https://github.com/contentauth/c2pa-ios libraries, to add C2PA v2.2+ support to all videos captured/edited/rendered in DiVine

2. Integrate current libProofMode support into C2PA metadata chain-of-custody, linking metadata stored in Nostr into signed+verifiable metadata within each MP4 video

3. Add support for both local and remote signing of media, backed by hardware attestation and secure key storage on device
- ack'ing this ticket on hardware attestation #183 

4. Add support for real-time validation of c2pa/proofmode claims in viewing of each video

5. Add implementation of proofcheck https://check.proofmode.org/ external verify to Divine website

n8fr8:
also @redaranj (https://github.com/redaranj) is making progress on wrapping c2pa-* SDKs in flutter for easier access and integration.

PR #6289, opened 2026-07-21
fix(proofmode): store C2PA manifest id under the key fromMetadata reads

rabble:
## Summary

Relates to #838.

Every camera publish since 1.0.17 falsely showed the **"Post without the human-made check?"** dialog even though C2PA signing succeeded and the published video verified with all four checks green.

Root cause: a metadata key mismatch. `NativeProofModeService` inserted the manifest id with `metadata.putIfAbsent('c2paManifestId', ...)` (both the existing-metadata path and the fresh-generation path), but `NativeProofData.fromMetadata` reads `metadata['c2pa_manifest_id']` -- matching the snake_case convention of every other native metadata key. The manifest id was silently dropped from every proof manifest.

The mismatch has existed since the ProofMode C2PA integration (#1517, Feb 2026) but was harmless until #6068 (Jul 18) added the C2PA prompt gate, which reads `c2paManifestId` off the proof manifest and shows the dialog when it's missing -- turning silent data loss into a false-positive dialog on every recording. Publish still worked because the C2PA manifest is embedded in the video file itself and the publish path re-reads it, which is why "Post anyway" produced a fully verified video.

Evidence from device log (build 1.0.17):

```
10:08:20.247 C2PA Active Manifest ID: urn:c2pa:db0a5480-51fc-45bf-8401-d64eadb5fe74
10:08:22.508 🔐 C2PA prompt gate -- configured: true, hasProof: true, failed: true
```

## Changes

- `native_proofmode_service.dart`: write the manifest id under `c2pa_manifest_id` at both call sites, matching `NativeProofData.fromMetadata`. The camelCase `toJson`/`fromJson` draft round-trip is unchanged and consistent.
- `native_proof_data.dart`: centralize the raw native metadata keys on `NativeProofData` and remove stale raw identity metadata reads that had no writer.
- `native_proofmode_service_test.dart`: two regression tests asserting `proofFile` returns a `NativeProofData` whose `c2paManifestId` carries the active manifest -- one for the fresh-generation path, one for the existing-metadata path. The existing-metadata test also asserts signing is not invoked.

## Downstream effects

This restores more than the dialog gate:

- `proof-verification-level` can return `verified_mobile` for successful camera recordings instead of falling back to `verified_web`/`basic_proof`.
- Blossom uploads now include `X-ProofMode-C2PA` when the generated proof manifest has the C2PA manifest id.

Both are intended restorations of the existing provenance contract.

## Testing

- `flutter test test/services/native_proofmode_service_test.dart`
- `flutter test` from `mobile/packages/models`

### DIFF
--- a/mobile/lib/services/native_proofmode_service.dart
+++ b/mobile/lib/services/native_proofmode_service.dart
@@ -130,7 +130,10 @@ class NativeProofModeService {
               category: .video,
             );
 
-            metadata.putIfAbsent('c2pa_manifest_id', () => activeManifestId);
+            metadata.putIfAbsent(
+              NativeProofData.metadataC2paManifestIdKey,
+              () => activeManifestId,
+            );
 
             // Create NativeProofData from metadata
             final proofData = NativeProofData.fromMetadata(metadata);
@@ -264,7 +267,10 @@ class NativeProofModeService {
 
       if (manifestInfo?.activeManifest != null) {
         final String activeManifestId = manifestInfo!.activeManifest!;
-        metadata.putIfAbsent('c2pa_manifest_id', () => activeManifestId);
+        metadata.putIfAbsent(
+          NativeProofData.metadataC2paManifestIdKey,
+          () => activeManifestId,
+        );
       }
 
       // Create NativeProofData from metadata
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz:
@rabble I pushed a takeover follow-up on top of your fix:

- centralized the raw native ProofMode metadata keys on `NativeProofData` and used them from the service write/read sites
- removed stale raw identity metadata reads from `fromMetadata` while leaving the attached identity fields intact
- tightened the existing-metadata regression test so it asserts `signVideo` is not called
- updated the PR body with `Relates to #838` and the restored downstream effects for verification level and Blossom headers

Verification passed locally:

- `flutter test test/services/native_proofmode_service_test.dart`
- `flutter test` from `mobile/packages/models`

CI is green on `ee5efe4aa`. Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover. The key mismatch is fixed, the raw metadata contract is centralized, stale raw identity metadata reads are removed, and regression coverage covers both fresh and existing metadata paths. Local verification and CI are green.

PR #6291, opened 2026-07-21
fix(support): anchor log-export share sheet so iPad idiom can present it

rabble:
## Summary

**Save Logs in Support Center failed instantly with "Failed to export logs" — no share sheet, no dialog — on iOS builds running on Apple Silicon Macs ("Designed for iPad") and on real iPads.**

Root cause: on iPad idiom, `share_plus` requires a `sharePositionOrigin` popover anchor and rejects the share with a `FlutterError` ("sharePositionOrigin: argument must be set") when it is missing (verified in `FPPSharePlusPlugin.m` of share_plus 12.0.1). `BugReportService._exportLogsNative` called `SharePlus.instance.share` without one, the error hit the catch block, and the user saw the failure toast with nothing else happening.

## Changes

- `exportLogsToFile` accepts an optional `sharePositionOrigin` and threads it into the native share sheet params.
- Support Center screen resolves the anchor via the existing `sharePositionOriginForContext` helper before awaiting, and passes it through.
- **Separate, related fix (native macOS build):** `Release.entitlements` granted only `com.apple.security.files.user-selected.read-only`, so the sandboxed desktop Save As export could not write the picked file outside `~/Downloads`. Swapped to `user-selected.read-write`.
- New test coverage for `sharePositionOriginForContext` (global bounds for laid-out widget, null for zero-size).

## Testing

- `flutter analyze` on changed files: clean
- `flutter test test/utils/share_position_origin_test.dart`: pass
- `flutter test test/unit/services/bug_report_log_export_test.dart`: pass
- Manual repro path: Divine 1.0.16 (804) iOS-on-Mac, Support Center → Save Logs previously showed only the red "Failed to export logs" snackbar; share flow requires manual verification on device/TestFlight build.

### DIFF
--- a/mobile/lib/screens/settings/support_center_screen.dart
+++ b/mobile/lib/screens/settings/support_center_screen.dart
@@ -9,6 +9,7 @@ import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/providers/app_providers.dart';
 import 'package:openvine/services/bug_report_service.dart';
 import 'package:openvine/services/zendesk_support_service.dart';
+import 'package:openvine/utils/share_position_origin.dart';
 import 'package:openvine/widgets/bug_report_dialog.dart';
 import 'package:openvine/widgets/feature_request_dialog.dart';
 import 'package:url_launcher/url_launcher.dart';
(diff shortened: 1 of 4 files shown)


dcadenas (approved):
✅ iPad share sheet anchored cleanly

PR #6290, opened 2026-07-21
fix(watermark): use refreshed Divine wordmark in download watermark

realmeylisdev:
## Description

The **"Download with Divine watermark"** export composited the **old** wordmark. `mobile/assets/icon/divine_wordmark.png` was the retired 196x70 **cream** mark (last touched in the very early days of the app). This swaps it **in place** for the refreshed brand wordmark: pure white, 320x105.

Per the issue, this keeps the existing watermark rendering path: the generator still loads `mobile/assets/icon/divine_wordmark.png`, draws it at 15% of video width, right-aligns the watermark block, and blends at 60% opacity. The refreshed asset has a different aspect ratio, so the rendered logo height and block top naturally follow the new artwork.

Adds a `WatermarkImageGenerator` test that guards the swap, verifies the bundled mark is pure white rather than the retired cream, and asserts the generated overlay contains white wordmark pixels in the computed bottom-right wordmark band.

Also removes the now-duplicate `assets/icon/White on transparent.png` asset so the bundled icon directory does not ship two byte-identical Divine wordmark PNGs.

**Related Issue:** Closes #6281 · Relates to #6282

## Out of Scope

Deferred to **#6282** (brand-mark refresh epic), intentionally not in this PR:

- Aligning the watermark **layout** to the brand-guidelines spec (left-aligned, randomly positioned vertically, 160px logo / 64px margin / 50px text at a 1920px width reference). The current bottom-right layout is unchanged here.
- Broader stale/orphaned brand asset cleanup outside the duplicate removed here.

**Known, pre-existing:** a white mark at 60% opacity over very light video is low-contrast (there is no backing scrim). This behaviour is unchanged by the swap and is flagged below for design sign-off rather than fixed here.

## Verification

- `flutter analyze test/services/watermark_image_generator_test.dart` — clean
- `flutter test test/services/watermark_image_generator_test.dart` — 3 passing
- Existing watermark suites (`watermark_download_service`, `watermark_download_progress_sheet`, `watermark_text_resolver`, `watermark_download_progress_cubit`) plus `watermark_image_generator` — 40 passing
- `dart format --set-exit-if-changed test/services/watermark_image_generator_test.dart` — clean
- Pre-push hook changed-file analyze/test gates — clean
- Rendered composited previews over dark / light / brand-green backgrounds to confirm bottom-right placement and legibility.

[...]

Issue #6282, opened 2026-07-21
task(branding): finish the Divine brand-mark refresh across app assets


NotThatKindOfDrLiz:
## Summary

The Divine brand-mark refresh left the **old** mark/wordmark in multiple places across the app. #6281 (watermark) is one instance; this issue tracks finishing the refresh holistically so we stop catching stale marks one user report at a time.

Scope: (1) refresh every **live** brand surface, (2) regenerate native app icons from the new source, (3) verify the splash, (4) delete confirmed-orphaned old-brand assets.

> **Note:** these are binary assets — each live surface needs a visual check against the approved brand mark from design/brand. References below were verified on `origin/main`.

## 1. Live brand surfaces (referenced in code — audit & refresh)

- [x] **Video watermark** — `assets/icon/divine_wordmark.png` → `watermark_image_generator.dart:15` (drawn at `:75`). Tracked in **#6281**.
- [ ] **Login / auth logo** — `assets/icon/logo.svg` → `auth_hero_section.dart:105` and `web_auth_screen.dart:221`. High-visibility (every sign-in).
- [ ] **App icon source** — `assets/icon/app_icon.png` → `pubspec.yaml` `flutter_launcher_icons` (`image_path`, `adaptive_icon_foreground`). Updating this + regenerating refreshes the whole iOS/Android/macOS/web icon set (below).
- [ ] **Branded loading spinner** — `assets/loading-brand-sprite.png` → `branded_loading_indicator.dart:80`. Shown on load/launch.

## 2. Native app icons (generated — regenerate after updating app_icon.png)

Regenerate from the new `app_icon.png` (e.g. `dart run flutter_launcher_icons`) and commit, rather than hand-editing:

- [ ] iOS — `ios/Runner/Assets.xcassets/AppIcon.appiconset/*`
- [ ] Android — `android/app/src/main/res/mipmap-*/ic_launcher.png`, `drawable-*/ic_launcher_foreground.png`
- [ ] macOS — `macos/Runner/Assets.xcassets/AppIcon.appiconset/app_icon_*.png`
- [ ] web — `web/favicon.png`, `web/icons/Icon-*.png`

## 3. Splash / launch (native config — verify)

- [ ] `assets` behind `android/.../drawable-*/launch_image.png` + `drawable*/launch_background.xml`
- [ ] `ios/Runner/Assets.xcassets/LaunchImage.imageset/*`

[...]

Issue #6281, opened 2026-07-21
fix(watermark): "Download with watermark" embeds the old Divine wordmark


NotThatKindOfDrLiz:
## Summary

The **"Download with Divine watermark"** share feature composites the **old Divine wordmark** onto the exported video. It should use the current/new wordmark.

## Where

`WatermarkImageGenerator` draws the wordmark from a bundled asset:

- `mobile/lib/services/watermark_image_generator.dart:15` — `static const _wordmarkAssetPath = 'assets/icon/divine_wordmark.png';`
- Loaded at `:35` (`_loadWordmarkImage()`) and drawn bottom-right at `:75` (`canvas.drawImageRect(wordmarkImage, ...)`), below which the creator identity text is rendered.

The stale asset itself is `mobile/assets/icon/divine_wordmark.png` — it's the old brand wordmark.

## Fix

Replace `mobile/assets/icon/divine_wordmark.png` with the approved **new** brand wordmark (source the asset from brand/design).

- **No generator code change should be needed** — the draw path already derives size from the image's own aspect ratio (`watermark_image_generator.dart:55-57`), so a drop-in asset swap scales correctly.
- Verify after swap: legibility/contrast of the new mark over both light and dark video content (it's drawn with a plain paint, no backing scrim), and that the new asset's transparency/padding match so the bottom-right placement still looks right.
- l10n copy already reads "Divine watermark" (`shareMenuDownloadWithWatermark`, `watermarkDownloadStageWatermarkingDesc`) — **no string changes required**.

## Repro

1. Open any video → share menu → **Download with Divine watermark**.
2. Open the exported file and inspect the bottom-right corner.
3. Expected: current Divine wordmark. Actual: old wordmark.

## Notes

Only one code reference to the asset exists (`git grep divine_wordmark` → `watermark_image_generator.dart:15` only), so this is a contained brand-asset refresh, not a rendering-logic change.

PR #6297, opened 2026-07-22
fix(watermark): correct download watermark to geometric heart-V wordmark

realmeylisdev:
## Description

Follow-up to the merged #6290. That PR swapped the download watermark to `assets/icon/White cropped.png` (320x105, pure white) believing it was the refreshed brand wordmark - but that asset is the **older cursive-script logotype**. On a real export it still reads as the old mark, which is what surfaced in review of the shipped build.

Per Liz, the **current** brand wordmark is the bold, rounded, geometric **heart-V "diVine"** - the same artwork already bundled as `assets/icon/logo.svg`. None of the "X on Y" brand-pack PNG/SVG/PDF variants match it; they are all the cursive logotype.

This PR regenerates `assets/icon/divine_wordmark.png` from `logo.svg`:

- Recolored to pure white, rasterized on a transparent page, cropped tight.
- **800x211, pure white (255,255,255), transparent** (~3.79:1, matching `logo.svg`).
- No generator change: `WatermarkImageGenerator` already sizes the mark from the asset's own aspect ratio and blends at 60% opacity, so position/opacity are unchanged; only the letterform and proportions follow the correct artwork.

The guard test now pins the bundled raster's aspect ratio to `logo.svg`'s `125x33` viewBox, while the overlay composition test still derives draw geometry from the bundled asset just like the generator. This catches accidental swaps back to the retired cursive wordmark assets.

## Out of Scope

- Watermark **layout** alignment to the brand-guidelines spec (left-aligned, randomized vertical position, etc.) - still tracked under the brand-mark refresh epic **#6282**.
- Any change to the cursive "X on Y" brand-pack assets.

**Known, pre-existing:** a white mark at 60% opacity over very light video is low-contrast (no backing scrim). Unchanged by this PR.

## Verification

- `flutter analyze test/services/watermark_image_generator_test.dart lib/services/watermark_image_generator.dart` - clean
- `flutter test test/services/watermark_image_generator_test.dart` - 3 passing (aspect-ratio guard; pure-white mark; white mark renders bottom-right; throws on missing asset)
- Asset checked: 800x211, every opaque pixel (255,255,255), alpha range 0-255 (transparent background preserved)
- Composited a preview over a real exported frame at 15% width / 60% opacity to confirm the geometric heart-V renders bottom-right

**Design sign-off:** recommend a real-device *Download with watermark* export over a light and a dark video before merge.

Relates to #6281 - Follow-up to #6290 - Relates to #6282

## Type of Change


rabble (approved):
Reviewed the asset, generator path, and regression guard against current `main`. The bundled raster is the 800×211 transparent derivative of the geometric `logo.svg` mark, the generator still derives placement geometry from the asset itself, and the test pins both the source aspect ratio and pure-white opaque pixels. The focused implementation is compatible with the #6290 generator change and all CI is green.

A light/dark real-device export remains useful product sign-off for contrast, but it is not a correctness blocker for this identity correction. LGTM.

NotThatKindOfDrLiz (approved):
Final pass is clean after the takeover follow-up. The asset is correct, the regression guard now catches swaps back to the retired cursive wordmark, the raster provenance is documented, and CI is green.

NotThatKindOfDrLiz:
@realmeylisdev I pushed a small takeover follow-up on top of this PR:

- added an aspect-ratio guard that pins `divine_wordmark.png` to `logo.svg`'s `125x33` viewBox, so the old cursive wordmark assets would fail the asset test
- documented `assets/icon/logo.svg` as the source of truth for the raster watermark asset
- updated the PR body to describe the stronger guard

Verification passed locally:

- `flutter analyze test/services/watermark_image_generator_test.dart lib/services/watermark_image_generator.dart`
- `flutter test test/services/watermark_image_generator_test.dart`

GitHub checks are green on `c7dbf19`. Please review the takeover changes and merge when you're ready.

PR #6299, opened 2026-07-22
feat(notifications): render video-sourced mentions

### LINKED ISSUE #5938: feat(notifications): add notifications for tags, mentions, and inspired-by references

NotThatKindOfDrLiz:
## Summary
- parse funnelcake root routing fields and preserve stable video mention anchors through cache
- render kind 34236 mention rows as video notifications with thumbnails, video copy, and a video-camera icon
- keep comment mentions comment-routed while preventing video mentions from auto-opening empty comments
- rebase onto current main while preserving owner-scoped cache rows and the metadata fetch union from main
- count grouped video mentions by distinct actors and restore cached mention coordinates from source video metadata
- align the Mobile CI timing-budget test with sharded-job matching in the checker so the rebased PR can pass the current gate

Closes #5938

## Tests
- flutter analyze
- flutter test packages/notification_repository/test/src/notification_repository_test.dart
- flutter test packages/funnelcake_api_client/test/src/models/relay_notification_test.dart packages/db_client/test/src/database/daos/notifications_dao_test.dart test/notifications/routing/notification_tap_target_test.dart test/main_notification_tap_routing_test.dart test/services/notification_helpers_test.dart test/notifications/view/notifications_view_test.dart test/notifications/widgets/video_notification_row_test.dart test/notifications/widgets/notification_type_icon_spec_test.dart
- flutter test --coverage (from mobile/packages/db_client)
- very_good test --coverage --min-coverage 40 (from mobile/packages/db_client)
- pre-push hook: analyzer, generated-files check, and changed-file tests


mobile/lib/notifications/routing/notification_tap_target.dart
```diff
@@ -71,11 +71,14 @@ class OpenInboxTarget extends NotificationTapTarget {
...
-bool notificationKindOpensComments(NotificationKind? kind) =>
+bool notificationKindOpensComments(
+  NotificationKind? kind, {
+  bool hasCommentTarget = false,
+}) =>
     kind == NotificationKind.comment ||
     kind == NotificationKind.reply ||
     kind == NotificationKind.likeComment ||
-    kind == NotificationKind.mention;
+    (kind == NotificationKind.mention && hasCommentTarget);
```
realmeylisdev, line 81: **Blocking — ordinary @mentions now dead-end instead of falling back to the actor's profile.**

This flag is doubly-loaded downstream. In `notifications_view.dart` the same `notificationKindOpensComments(...)` result gates three separate things: whether comments auto-open, whether the `NotificationTargetResolver` E/e-tag walk runs (`:340`), and whether `_navigateToVideo` can return `false` to trigger the `if (!navigated) → _navigateToProfile` fallback (`:277`). Making it `false` for mentions silently turns off the last two as well.

Concrete path — Bob publishes a kind 30023 that p-tags Alice (30023 is a live source kind, `client.rs:962`):
- funnelcake types it `mention` via the catch-all arm (`client.rs:1061-1071`) and leaves `target_comment_id` empty (`client.rs:1169-1175`), so the key is omitted from the JSON.
- Mobile: `_isVideoSourcedMention` needs kind 34236, so it stays an `ActorNotification`. `targetEventId = sourceEventId`, and `_actorVideoAddressableId` returns `null` for `mention` — it only populates for `likeComment`/`reply`.
- Tap → `hasVideoTarget` true → `OpenVideoTarget` → `shouldAutoOpenComments == false` → the `else` at `:352-354` sets `routeId = videoEventId` and pushes `/video/<mention-event-id>`, returning `true`. The profile fallback never runs and Alice lands on `videoErrorNotFound`.

On `origin/main` the resolver walked, returned null, and the tap correctly landed on Bob's profile.

I verified this empirically rather than by reading: flipping the `mention-2` fixture in `notifications_view_test.dart:1258` back to `hasCommentTarget: false` turns the existing guard test red —

```
mention tap falls back to actor profile when resolver cannot find the root video [E]
Expected: an object with length of <1>
  Actual: []
```

That fixture is a tagless kind-1 (`'plain mention with no video context'`), which is exactly the shape funnelcake emits with `target_comment_id` empty — so the two fixtures at `:1190` and `:1258` were moved to a value the backend does not produce, which is what keeps the suite green.

Suggested fix — separate "open the comment sheet" from "resolve this event id to a root video":

```dart
final shouldAutoOpenComments = notificationKindOpensComments(
  notificationKind, hasCommentTarget: hasCommentTarget,
);
// Resolution is a *kind* question, not a comment-target question.
final shouldResolveRootVideo =
    notificationKindOpensComments(notificationKind, hasCommentTarget: true);
...
} else if (shouldResolveRootVideo) {
  routeId = await NotificationTargetResolver(...)
      .resolveVideoEventIdFromNotificationTarget(videoEventId);
  if (!context.mounted) return false;
  if (routeId == null) return false; // caller falls back to profile
}
```

Then revert both fixtures to `hasCommentTarget: false` and keep a case asserting an unresolvable mention still lands on the actor profile.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -1217,12 +1256,19 @@ class NotificationRepository {
...
+          )
           .firstWhere((d) => d != null, orElse: () => null);
-      final addressableId = _recipientScopedVideoAddressableId(
-        dTag: dTag,
-        video: video,
-      );
+      final rootAddressableId = group
+          .map((n) => n.rootAddressableId)
+          .firstWhere((id) => id != null && id.isNotEmpty, orElse: () => null);
+      final addressableId = entry.key.kind == NotificationKind.mention
```
realmeylisdev: **Should-fix — `rootAddressableId` is trusted without an owner check, so a mention can route to the recipient's *own* video.**

This swaps the `_userPubkey`-pinned `_recipientScopedVideoAddressableId` for the raw server coordinate. Dropping the recipient pin is right for a mention (it is structurally someone else's video), but the replacement has no bound of its own.

funnelcake only guarantees `root_addressable_id == 34236:<sender>:<sender-d-tag>` when the source video has a non-empty `d` tag — `notification_source_video_coordinate` returns `None` otherwise (`client.rs:1023-1040`). Without it, `notification_root_context` falls through to the generic path and derives the root from the `a`/`A` tag, which for an "Inspired by" video is the *original* video — often the recipient's. funnelcake ships a unit test for that fallback (`malformed_video_source_falls_back_to_event_card_key`, `client.rs:13493`), and the `referenced_video` join keys on the same coordinate, so the title and thumbnail come back as the recipient's too.

Result: Bob sees "Alice mentioned you" over *his own* video title and thumbnail, and the tap opens his own video. divine-mobile always writes a `d` tag, so this needs a third-party client or a publish bug — but it is worth a guard, especially since the adjacent doc comment at `:1508-1520` states the invariant this branch gives up: *"a wrong/stale d-tag can only ever surface the recipient's own (or a non-existent) video — never another creator's."*

```dart
final coordinate = _nonEmpty(rootAddressableId);
final ownerHex = coordinate?.split(':').elementAtOrNull(1);
final addressableId = (coordinate != null && ownerHex != _userPubkey)
    ? coordinate
    : _sourceVideoAddressableId(dTag: dTag, video: video);
```

Worth a repository test with `rootAddressableId = '34236:<recipient-pubkey-hex>:<d-tag>'` asserting the row does not route at the recipient.

mobile/packages/db_client/test/drift/app_database/generated/schema_v1.dart
```diff
@@ -613,6 +613,25 @@ class Notifications extends Table with TableInfo {
     requiredDuringInsert: false,
     $customConstraints: 'NULL',
   );
+  late final GeneratedColumn<String> videoAddressableId =
+      GeneratedColumn<String>(
+        'video_addressable_id',
```
realmeylisdev, line 618: **Should-fix — the generated snapshot was updated but its generator input was not.**

This file gained the two columns, but `mobile/packages/db_client/drift_schemas/app_database/drift_schema_v1.json` — the file it is generated *from* — is untouched by the diff:

```
$ git show HEAD:.../drift_schemas/app_database/drift_schema_v1.json | grep -c "video_addressable_id\|has_comment_target"
0
$ git show HEAD:.../test/drift/app_database/generated/schema_v1.dart | grep -c "video_addressable_id\|has_comment_target"
3
```

Every prior runtime-ALTER column *is* in that JSON (`event.d_tag`, `personal_reactions.addressable_id`, `direct_messages.send_batch_id`, `conversations.owner_pubkey`), so this is the first divergence in the file's history. `schemaVersion` is still `1`, so the JSON is the live snapshot rather than a frozen artifact.

Nothing is red today and no CI job guards the JSON — the cost lands on the next migration author, for whom `dart run drift_dev make-migrations` now errors with *"A schema for version 1 … already exists and differs from the current schema"*, and `drift_dev schema generate` regenerates this file *without* the two columns, failing `migration_test.dart` while naming this PR's columns.

```bash
cd mobile/packages/db_client
rm drift_schemas/app_database/drift_schema_v1.json
dart run drift_dev make-migrations
```

The regenerated `schema_v1.dart` should come out identical to what is already committed — only the JSON needs to land.

mobile/lib/services/notification_helpers.dart
```diff
@@ -41,6 +41,9 @@ abstract class NotificationPayloadKeys {
   /// The source event itself (the like/comment/follow/mention event).
   static const String eventId = 'eventId';
 
+  /// Explicit comment target id for comment mention / reply-style payloads.
+  static const String targetCommentId = 'targetCommentId';
```
realmeylisdev: **Should-fix — no producer sends this key, so `hasCommentTarget` is a constant `false` on the push path.**

divine-push-service builds its entire FCM `data` map in one function (`create_fcm_payload`, `src/event_handler.rs:778-892`, plus `insert_trigger_reference_fields` at `:906-954`). The complete key set is `type`, `eventId`, `title`, `body`, `senderPubkey`, `senderName`, `receiverPubkey`, `receiverNpub`, `eventKind`, `timestamp`, `referencedEventId`, `referencedAddress`, `referencedKind`, `referencedAuthorPubkey`, `referencedDTag`. `git grep -in "targetCommentId|target_comment_id" origin/main -- src/` returns nothing.

So this key, and the `targetCommentId` parameter threaded through `notification_service.dart` (field, `==`, `hashCode`) and five call sites in `main.dart`, can never be non-null in production.

To be explicit about something that looks alarming but is not: I checked whether this silently stops push comment-mentions from auto-opening comments, and **it does not**. The push service only classifies kind 1 without an `e` tag, kind 30023, and kind 34236 as `mention` — none is a comment (kind 1 *with* an `e` tag and kind 1111 both map to `Comment`). `autoOpenComments: false` is correct for all three, so the push-path change here is a genuine fix for the empty-comments bug. No need to re-raise it.

That leaves the plumbing as forward-compat with no landing date. Either drop it, or land the push-service side in the same cycle.

mobile/lib/notifications/widgets/video_notification_row.dart
```diff
@@ -260,6 +263,7 @@ class _MessageText extends StatelessWidget {
 bool _typeShowsTitle(NotificationKind type) {
   return type == NotificationKind.like ||
       type == NotificationKind.comment ||
+      type == NotificationKind.mention ||
```
realmeylisdev: **Nit (copy) — the mention string has no noun phrase for the title to attach to.**

The three existing title-bearing strings all end in "your video", so the appended title reads as apposition. `notificationMentionedYou` ends in a pronoun:

| key | `app_en.arb` |
|---|---|
| `notificationLikedYourVideo` | `{actorName} liked your video` |
| `notificationCommentedOnYourVideo` | `{actorName} commented on your video` |
| `notificationRepostedYourVideo` | `{actorName} reposted your video` |
| `notificationMentionedYou` | `{actorName} mentioned you` |

The append at `:234-238` is a bare `TextSpan(text: ' $videoTitle')`, so a video titled "Inspired clip" renders as **"Alice mentioned you Inspired clip"**. `video_notification_row_test.dart:164-179` asserts the two fragments separately and never the joined sentence, so it stays green.

There is also a semantic wrinkle worth a product call: this title is the *sender's* video (funnelcake hydrates `referenced_video` from the sender's own coordinate), sitting in a list where every neighbouring title is the recipient's.

Either drop `mention` from `_typeShowsTitle` — the thumbnail already identifies the video and its semantic label reads "Video thumbnail for {title}" — or add a connector-bearing key so translators control word order:

```json
"notificationMentionedYouInVideo": "{actorName} mentioned you in"
```

mirrored into all 18 `app_*.arb` files (or `_knownUntranslatedDebt`), then `flutter gen-l10n` + `flutter test test/l10n/arb_consistency_test.dart`.

mobile/lib/notifications/widgets/notification_leading_type_icon.dart
```diff
@@ -30,11 +32,14 @@ class NotificationLeadingTypeIcon extends StatelessWidget {
...
 
+  /// Optional row-specific icon while keeping the shared accent mapping.
+  final DivineIconName? iconOverride;
+
   @override
   Widget build(BuildContext context) {
     final spec = notificationTypeIconSpec(type);
     return NotificationTypeIcon(
-      icon: spec.icon,
+      icon: iconOverride ?? spec.icon,
```
realmeylisdev: **Nit — this routes around the spec that calls itself the single source of truth, and has no coverage.**

`notification_type_icon_spec.dart:4-7` declares itself *"Single source of truth for the design contract … a regression test in notification_type_icon_spec_test.dart locks the mapping so it can't silently drift again."* `iconOverride ?? spec.icon` is a new path around that lock: `NotificationKind.mention` now renders `videoCamera` from `VideoNotificationRow` and `chat` from `ActorNotificationRow`, and the locking test sees neither.

`grep -rn "iconOverride\|videoCamera" mobile/test/` returns nothing under `mobile/test/notifications/` — replacing the override with `null` leaves all 189 tests there green, so the icon named in the PR summary can regress unnoticed.

Minor geometry note while you are here: `assets/icon/video_camera.svg` is `viewBox="0 0 29 18"` where `chat`/`heart`/`repeat`/`user` are `0 0 32 32` with ~3.5 units of internal padding. Under `DivineIcon`'s default `BoxFit.contain` in the fixed 20px box the camera paints ~20×12.4 against siblings' ~17×15.6 — noticeably shorter and wider, though contrast is fine.

Cleanest fix is to teach `notificationTypeIconSpec` about the video-mention case so the mapping stays inside the locked spec, extend `notification_type_icon_spec_test.dart`, and add one assertion in `video_notification_row_test.dart` that the mention row renders `DivineIconName.videoCamera`.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -1169,24 +1199,29 @@ class NotificationRepository {
...
       final eventId = _videoAnchorEventId(kind, n);
       if (eventId == null || eventId.isEmpty) continue;
-      if (_hasKnownReferencedVideoOwnerMismatch(
-        kind: kind,
-        referencedVideoEventId: eventId,
-        videosById: videosById,
-        videoNotFoundIds: videoNotFoundIds,
-        rootEventPubkey: n.rootEventPubkey,
-      )) {
+      if (kind == NotificationKind.mention && !_isVideoSourcedMention(n)) {
```
realmeylisdev: **Nit — this guard is unreachable.**

`isVideoAnchored(k, n)` is `k == like || k == comment || k == repost || _isVideoSourcedMention(n)`. When `kind == mention` the first three disjuncts are false, so the `if (!isVideoAnchored(kind, n)) continue;` at `:1211` has already reduced to exactly `_isVideoSourcedMention(n)`. `_isVideoSourcedMention` is a pure function of `n` and nothing mutates `n` in between, so this condition is constantly false.

Suggest deleting `:1214-1216`. While you are here, the owner-mismatch skip on the next line reads more truthfully as `!_isVideoSourcedMention(n)` than `kind != NotificationKind.mention` — that is the actual intent (skip the owner check only for *video-sourced* mentions) and it removes the dead branch at the same time.

realmeylisdev (commented):
Reviewed against the funnelcake + push-service contracts (read the Rust on `origin/main` and ran funnelcake's `notifications_source_video_ptags_and_dedupe_replaceable_edits` against a local ClickHouse). The 34236 design is right — coordinate-keyed grouping and the `hasCommentTarget` gate match the backend exactly.

One blocking item: gating the resolver walk on `hasCommentTarget` also removed the profile fallback for ordinary @mentions. Details inline, plus a heads-up that funnelcake #670 is merged but not yet deployed to production.

realmeylisdev:
Two things that aren't code review but affect how this gets validated:

**The backend half isn't in production yet.** funnelcake #670 merged at `6fb4927` on 2026-07-22 02:50 UTC, and the most recent *Deploy to Production* run is `ed32726` at 01:39 UTC — about 71 minutes earlier, so #670 is not an ancestor of the deployed commit. Migration `000186_add_video_sources_to_notification_inbox` hasn't run in prod, which means production emits zero kind-34236 mentions today; staging has it. The source backfill is also cursor-forward-only, so even after deploy, videos published before it never retroactively notify. Worth calling out in the test plan — this can only be exercised against staging right now.

**One product question that shapes the copy.** divine-mobile writes `p` tags on kind 34236 for four different reasons: `collaborator`, caption `mention`, `inspired-by`, and video-reply threading. funnelcake fans out on every 64-hex `p` tag without reading the marker, and the notifications response carries no field for it (`source_tags` is deliberately blanked on write). So all four render identically as "mentioned you". That mislabeling predates this PR, but this PR gives the row a thumbnail, a video-camera badge and a tappable target, which makes it louder. If dedicated Inspired-by copy is wanted, funnelcake needs a role field first — happy to file that issue if useful.

Verified by running funnelcake's `notifications_source_video_ptags_and_dedupe_replaceable_edits` against a local ClickHouse on `origin/main`; it passes, and the materialized row confirms `root_d_tag`, `root_addressable_id` and an empty `target_comment_id` exactly as this PR assumes.

NotThatKindOfDrLiz:
Takeover update pushed in `e1f2dfc8c`.

Changes made:
- Fixed the plain mention tap regression by separating root-video resolution from comment-sheet auto-open. Ordinary actor mentions now resolve only when they can actually find a video target, then fall back to the actor profile; video/comment mentions still use the correct target behavior.
- Added a source-owner guard for video mention root coordinates so third-party/bad `root_addressable_id` values cannot route to the recipient/original video, plus regression coverage.
- Regenerated the Drift v1 JSON schema alongside the generated schema snapshot.
- Removed the unused push/local `targetCommentId` plumbing that had no production producer.
- Moved the video-sourced mention icon mapping into the notification type icon spec and covered it.
- Fixed the video mention row copy path so mention rows do not append the video title after “mentioned you”.

Local verification passed:
- `dart run drift_dev make-migrations` from `mobile/packages/db_client`
- targeted `flutter test` suite from `mobile/` covering Funnelcake model parsing, notification repository, db DAO, routing, tap handling, helpers, notification view, video row, and icon spec
- `flutter test --coverage` from `mobile/packages/db_client`
- `very_good test --coverage --min-coverage 40` from `mobile/packages/db_client`
- `flutter analyze` from `mobile/`

GitHub checks are green on the pushed commit: Mobile CI, Analyze, Format, Generated Files, VGV Tag Gate, Tests, mergeability, semantic PR, build jobs, preview build, scope detection, and CLA.

rabble (changes requested):
Reviewed at `e1f2dfc8c`. The shape of this is right — moving mentions onto the root NIP-33 coordinate so an edited video doesn't split the group is the correct anchor, and the `videoMention` persist type keeps the cache round-trip honest. But I'm requesting changes: @realmeylisdev's blocking finding is still live at this head, and I found a second routing bug on the event-id path that fires on divine's own video-reply feature rather than a third-party edge case.

## Blocking — video replies route the mention row at the recipient's own video

`_videoAnchorEventId` and `_videoMetadataEventId` both resolve a video-sourced mention as `_nonEmpty(n.rootEventId) ?? _nonEmpty(n.sourceEventId)`. For a kind 34236 source the source event *is* the video, so preferring `rootEventId` is only safe if funnelcake never populates it for that row. It does.

Traced through both sides:

**Mobile** — `video_event_publisher.dart:247`, the reply branch, writes `['e', context.rootEventId, '', context.rootAuthorPubkey]` and `['p', context.rootAuthorPubkey]` onto the kind 34236 itself. So a video reply is a video carrying an `e` tag pointing at the parent video plus a p-tag for the parent's author.

**Funnelcake** — `crates/clickhouse/src/client.rs`:
- `notification_type_for_candidate` has no arm for 34236, so it falls to the `_ => "mention"` catch-all (`:808`).
- `fallback_referenced_event_id` takes the first `e` tag (`:10439`) → the **parent video's** event id.
- `notification_root_context` sets `root_event_id` from `E` (absent on a video) and otherwise falls back to `referenced_event_id` (`:824-831`) → again the **parent video's** id.
- The p-tagged target is the parent's author — the recipient.

So the row arrives with `root_event_id` = a video the *recipient* owns. Mobile prefers it, and the result is a "Alice mentioned you" row whose thumbnail, title, and tap target are all the recipient's own video. Alice's actual reply is never reachable from the notification.

This is the same class as @realmeylisdev's `:1266` finding on the coordinate path, but reached through the event id, and it does not need a third-party client — `videoReplies` is a shipped flag in this repo.

Worth noting the obvious guard doesn't work here: gating on `NIP71VideoKinds.isVideoKind(rootEventKind)` doesn't discriminate, because the parent is a 34236 too.

Suggested fix — for a video-sourced mention, don't consult the root at all:

```dart
static String? _videoMetadataEventId(RelayNotification n) {
  final kind = _mapNotificationKind(n);
  if (_isVideoSourcedMention(n)) {
    // The source event IS the video (_isVideoSourcedMention already proved
    // source_kind is a video kind). rootEventId points at the *parent* for a
    // video reply, which is the recipient's own video.
    return _nonEmpty(n.sourceEventId);
  }
  return _videoAnchorEventId(kind, n);
}
```

and the matching branch in `_videoAnchorEventId`. Please add a repository test with a video-sourced mention whose `rootEventId` differs from `sourceEventId`, asserting the row anchors on the source.

## Blocking — @realmeylisdev's `notification_tap_target.dart:81` finding

Not restating it; the analysis there is correct and empirically demonstrated, and the suggested split of "open the comment sheet" from "resolve this event id to a root video" is the right shape. Still unfixed at this head.

## `rootEventKind` is parsed, asserted in tests, and never read

`relay_notification.dart` gains `rootEventKind` with a doc comment, `_intValue(json['root_event_kind'])` parsing, and two test assertions — but nothing in `notification_repository.dart` or the app layer ever reads it. Funnelcake does populate it (`client.rs:838-862`, overridden by the `A`/`a` coordinate kind when present), so this is a real field with no consumer.

Either drop it, or put it to work: `_trustedSourceRootAddressableId` currently re-derives the kind by string-splitting the coordinate and running it through `_parseAddressableId`, when the server already sent it typed.

## Smaller things

- **`rootAddressableId` computed for every group, consumed only by mentions.** In `_groupVideoNotifications` the `group.map(_trustedSourceRootAddressableId).firstWhere(...)` scan runs for like/comment/repost groups too, where the result is discarded on the next line. Move it inside the `kind == NotificationKind.mention` branch.
- **`_trustedSourceRootAddressableId`'s `parsed.pubkey != notification.sourcePubkey` check is doing real work** — it's the substitute for the `_hasKnownReferencedVideoOwnerMismatch` guard this PR skips for mentions. Worth a comment saying so, since the skip at `:1213` otherwise reads as simply dropping a safety check.
- **`_verbFor` mention** uses `notificationMentionedYou('').trimLeft()`, which assumes a leading placeholder. Holds for `en`/`de`; breaks for name-final locales. Pre-existing pattern across the other verbs, so not new debt — flagging only because this adds a fourth instance.
- Migration `ALTER TABLE ... INTEGER NOT NULL DEFAULT 0` drops the `CHECK (... IN (0,1))` the generated schema declares, but that matches existing precedent at `app_database.dart:408` and `:438`. No action needed.

## One correction to my own earlier read

I initially flagged the push path — `pushNotificationTapTarget` never passing `hasCommentTarget`, so push-tapped comment-mentions stop auto-opening comments. @realmeylisdev's analysis of divine-push-service is right and mine was wrong: the push service only classifies kind 1 without an `e` tag, kind 30023, and kind 34236 as `mention`, and routes kind 1 *with* an `e` tag and kind 1111 to `Comment`. So no push `mention` is ever a comment, and `autoOpenComments: false` is correct for all three. Not a regression — disregard.

The `targetCommentId` plumbing threaded through `notification_service.dart` and `main.dart` is still dead in production, per that same thread.

*(Review by Claude Code on behalf of @rabble.)*

NotThatKindOfDrLiz:
Takeover follow-up pushed in d63efcc63.

Changes made:
- Fixed video-sourced mentions to anchor metadata and row video IDs on sourceEventId, not rootEventId. For video replies, rootEventId can be the recipient's parent video, while the source event is the sender's reply video.
- Prefer resolved source-video metadata for video mention title/thumbnail when available, so parent-video payload metadata does not leak into reply mention rows.
- Wired rootEventKind into trusted root coordinate validation and reject coordinates whose typed kind disagrees with the parsed coordinate kind.
- Moved trusted root-coordinate scanning onto the mention-only path and documented why video mentions skip the normal recipient-owner mismatch guard.
- Added repository regressions for rootEventId != sourceEventId, parent payload metadata, source-owner fallback, and non-video rootEventKind rejection.

Verification passed locally:
- flutter test packages/notification_repository/test/src/notification_repository_test.dart
- flutter test packages/funnelcake_api_client/test/src/models/relay_notification_test.dart packages/notification_repository/test/src/notification_repository_test.dart packages/db_client/test/src/database/daos/notifications_dao_test.dart test/notifications/routing/notification_tap_target_test.dart test/main_notification_tap_routing_test.dart test/services/notification_helpers_test.dart test/notifications/view/notifications_view_test.dart test/notifications/widgets/video_notification_row_test.dart test/notifications/widgets/notification_type_icon_spec_test.dart
- flutter analyze

GitHub checks are green on d63efcc63: Mobile CI, Analyze, Format, Generated Files, Tests, VGV Tag Gate, build matrix, preview, mergeability, semantic PR, and CLA.

One note on the prior review: the tap-target blocker called out as still live was already fixed in e1f2dfc8 via resolveTargetToVideo: true in NotificationsView, with auto-open-comments kept separate behind hasCommentTarget.

NotThatKindOfDrLiz:
@rabble Can you please finalize the review?

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -588,6 +588,7 @@ class NotificationRepository {
         id: row.id,
         type: videoKind,
         videoEventId: videoEventId,
+        videoAddressableId: _nonEmpty(row.videoAddressableId),
```
realmeylisdev, line 608: **Left unfixed — flagging rather than half-landing it.** The cache round trip restores `videoAddressableId` here, but `_enrichCachedPlaceholders` (~`:522`) then runs `_recipientScopedVideoAddressableId` over *every* hydrated `VideoNotification`, including the new `videoMention` rows. That helper pins the pubkey to `_userPubkey` and bails when `video.pubkey != _userPubkey`, which for a mention is always — the source video belongs to the sender.

So a mention row cached with a null coordinate (first page arrived with `root_addressable_id` absent *and* the stats fetch threw) can never regain one, even though `videoMetadata.videosById[videoEventId]` in that same block already holds the sender's `pubkey` and `dTag` — exactly `_sourceVideoAddressableId`'s input. The tap then falls to the raw `videoEventId`, which for a 34236 card is funnelcake's `argMin(source_event_id)`, dead on relays after any edit.

Narrow (needs both sources to fail on the same load), but the fix is a mention-aware branch in `_enrichCachedPlaceholders` mirroring the grouping path. Happy to land it in this PR or file it — your call.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -1369,9 +1432,19 @@ class NotificationRepository {
...
           : const [],
       videoAddressableId: _actorVideoAddressableId(type, notification),
+      hasCommentTarget: _actorHasCommentTarget(type, notification),
     );
   }
 
+  static bool _actorHasCommentTarget(
+    NotificationKind type,
+    RelayNotification notification,
+  ) =>
```
realmeylisdev, line 1483: **Not a bug — recording why I left it.** This mention arm cannot fire on current traffic. `target_comment_id` is non-empty only when `source_kind == 1111` or the referenced event resolves to a kind-1111 (`client.rs:1170-1176`), while `notification_type == "mention"` is the catch-all arm reachable only for source kinds 34236 and 30023 (`client.rs:1065-1075` plus the ingest allow-list in migration `000186`). Neither can carry a comment target.

That makes the whole `hasCommentTarget` chain inert for now — this branch, the Comments-tab clause, the model field and the `has_comment_target` column. It's coherent forward-compat and removing it would undo the R2 fix's shape, so I'd keep it. Worth a line in the PR description that it activates only once funnelcake types comment-mentions, so the next reader doesn't spend the time I did proving it dead.

realmeylisdev (commented):
Re-reviewed at `d63efcc63` against funnelcake `origin/main` (`2cac08f`, includes #670). One correction to the record, one real defect, three unpinned guards — all fixed and pushed in `999cb2eac`.

**Correcting my own and @rabble's earlier reads:** the blocking "video replies route at the recipient's video" finding does not reproduce. `notification_root_context` short-circuits for *any* kind 34236 carrying a `d` tag and returns the sender's own coordinate, discarding every `e`/`E`/`a`/`A` tag (`crates/clickhouse/src/client.rs:1102-1115`, gated by `notification_source_video_coordinate` at `:1027-1044`, pinned by `video_source_root_context_uses_source_coordinate_not_referenced_a_tag` at `:13780`). `root_event_id` *is* `source_event_id` for 34236. The `client.rs:824-831` lines cited in that review are from the pre-#670 file. The anchor-on-source fix is still correct, just a no-op — I've replaced the code comments that asserted the wrong premise.

**Real defect (fixed):** `_trustedSourceRootAddressableId` rejects a foreign root coordinate, but the row kept consuming two other fields derived from that same rejected coordinate — `root_d_tag` (through the `_sourceVideoAddressableId` fallback) and `referenced_video`'s title/thumbnail, which funnelcake joins on `(root_event_pubkey, root_event_kind, root_d_tag)` (`client.rs:12193-12206`). For a 34236 published *without* a `d` tag, funnelcake takes the generic path and fills `root_*` from the `a`/`A` tag — the original creator's coordinate, often the recipient's. The row then spliced the recipient's d-tag onto the sender's pubkey (a coordinate resolving to nothing) and rendered the recipient's own title and thumbnail under "mentioned you". Now every root-derived field shares the coordinate's trust decision. Side note: `referenced_d_tag` is set from `root_d_tag` server-side (`client.rs:12145`), so the mention/non-mention d-tag split was a no-op — dropped.

**Three guards no test could catch** (verified by deleting each and watching the suite stay green): the sender-owner check with the source video unresolved, the `'videoMention'` persist type on the *write* half of the round trip, and both wiring hops feeding the video-camera icon. All three now go red under mutation.

Everything else from both rounds is fixed at this head — R2 via `resolveTargetToVideo`, plus the drift JSON, `rootEventKind` now being read, the mention copy, the icon spec, and the two dead-guard nits. Two things I deliberately did **not** change are inline. 429 tests green, `flutter analyze` clean.

One note on scope: funnelcake #670 shipped to production in `9dbfe6b6` on 2026-07-24 01:43 UTC, so this is now testable against prod — my earlier "staging only" comment is stale. The source backfill stays cursor-forward-only.

realmeylisdev:
Heads-up for whoever sees a red build here: the `Tests` job failed once on `999cb2eac` at `test/providers/video_editor_provider_test.dart` — *"container teardown reaps deferred files as a safety net"*. Re-running the same commit passed, and the test passes in isolation locally.

It is a pre-existing race, unrelated to this PR: `ref.onDispose` fires `unawaited(_flushDeferredFileCleanup())` (`lib/providers/video_editor_provider.dart:162`), which runs two drift queries and real file deletes, while the test only awaits a fixed-count `pumpEventQueue()`. Under a loaded merged isolate the deletes have not landed yet. The sibling assertion at `:2982` has the same shape.

Not fixing it inside a notifications PR — happy to send a separate one that waits on the condition instead of a fixed pump count.

mobile/lib/notifications/routing/notification_tap_target.dart
realmeylisdev: Fixed in `e1f2dfc8c`: `resolveTargetToVideo` now drives the resolver walk independently of `hasCommentTarget`, so a plain note mention that resolves to nothing returns false and falls back to the actor profile. Verified at `999cb2eac` — `mention tap falls back to actor profile when resolver cannot find the root video` is green, and flipping `resolveTargetToVideo` to false turns it red. Resolving.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Fixed in `e1f2dfc8c` via `_trustedSourceRootAddressableId`, and the guard is now pinned in `999cb2eac` — the pre-existing test stubbed `VideoStats`, so the *resolved-owner* gate fired first and `parsed.pubkey != sourcePubkey` never decided anything; deleting it left the suite green. The new case runs with the source video unresolved so that line is the only defence. `999cb2eac` also closes the related hole this thread implied: `root_d_tag` and the `referenced_video` media were still consumed after the coordinate was rejected. Resolving.

mobile/packages/db_client/test/drift/app_database/generated/schema_v1.dart
realmeylisdev: Fixed in `e1f2dfc8c` — `drift_schema_v1.json` is in the diff and matches the generated `schema_v1.dart` field for field. Resolving.

mobile/lib/services/notification_helpers.dart
realmeylisdev: Fixed in `e1f2dfc8c` — the `targetCommentId` push plumbing is gone from `notification_service.dart` and `main.dart`. The REST-side `hasCommentTarget` branch stays and is still producer-less today; I have recorded why in a separate thread rather than reopening it here. Resolving.

mobile/lib/notifications/widgets/video_notification_row.dart
realmeylisdev: Fixed in `e1f2dfc8c` — `_typeShowsTitle` still excludes `mention`, so no title is appended after "mentioned you"; the thumbnail carries the video identity. Resolving.

mobile/lib/notifications/widgets/notification_leading_type_icon.dart
realmeylisdev: Fixed in `e1f2dfc8c` — the mapping moved into `notificationTypeIconSpec` behind `isVideoSourcedMention`, so the locked spec stays the single source of truth. `999cb2eac` closes the coverage half: the spec test only exercised the pure function, and deleting either wiring hop (`VideoNotificationRow` passing the flag, `NotificationLeadingTypeIcon` forwarding it) left all 190 tests under `test/notifications/` green. Both hops now go red under mutation. Resolving.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Fixed in `e1f2dfc8c` — the dead branch is gone. `999cb2eac` also replaced the comment above it, which asserted that Funnelcake can hand back a root owned by the recipient for a video reply. That is not true for a well-formed kind 34236: `notification_root_context` short-circuits on any 34236 carrying a `d` tag and returns the sender own coordinate (`crates/clickhouse/src/client.rs:1102-1115`). The real trigger is a 34236 published *without* a `d` tag, which drops to the generic `a`/`A` path. Resolving.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Update: I attempted this fix and backed it out. The one-line branch is right — take `_sourceVideoAddressableId(video: video)` for a mention instead of the `_userPubkey`-pinned helper — but I could not get a test I trust.

A regression test that asserts the coordinate reappears after enrichment passes about half the time and otherwise hangs until the 10-minute framework timeout. The enriched emission simply never arrives on the failing runs. `_fetchGeneration` is only bumped by `_refreshResult()`, which the test never calls, and `_sameNotificationIds` holds, so the abort guards do not explain it — something in the `unawaited(_enrichCachedPlaceholders(...))` path is order-dependent. Worth noting the whole `cache hydration` group uses `expectLater(stream, emitsThrough(...))`, which stalls rather than fails when an expected emission never lands, so a latent raciness here would not have surfaced before.

Not shipping an unreliable test to close a narrow gap. Leaving this thread open — the coordinate gap and the enrichment raciness look like one follow-up.

realmeylisdev:
@rabble @NotThatKindOfDrLiz — round 3 is done. Summary of what changed and what's left.

### Pushed in `999cb2eac` (all checks green)

**One real defect.** `_trustedSourceRootAddressableId` rejects a root coordinate that names someone other than the sender — but the row kept consuming two *other* fields derived from that same rejected coordinate: `root_d_tag` (via the `_sourceVideoAddressableId` fallback) and `referenced_video`'s title/thumbnail, which funnelcake joins on `(root_event_pubkey, root_event_kind, root_d_tag)` (`client.rs:12193-12206`). Result was a row that spliced the recipient's d-tag onto the sender's pubkey — a coordinate resolving to nothing — and rendered the recipient's own thumbnail under "mentioned you". All root-derived fields now share one trust decision.

**Three guards nothing could catch.** Found by deleting each and watching the suite stay green:
- the `parsed.pubkey != sourcePubkey` owner check — the existing test stubs `VideoStats`, so the *resolved-owner* gate fired first and this line never decided anything
- the `'videoMention'` persist type on the **write** half of the cache round trip (only the read half was pinned)
- both wiring hops feeding the video-camera icon — the spec test only called the pure function

All now go red under mutation.

**Plus** a redundancy cleanup: `referenced_d_tag` is set from `root_d_tag` server-side (`client.rs:12145`), so the mention/non-mention d-tag split was a no-op.

### Correcting the blocking finding

@rabble — the "video replies route at the recipient's own video" blocker doesn't reproduce, and I think the cause is a stale local funnelcake checkout: the review cites `client.rs:824-831`, which are pre-#670 line numbers. Against `origin/main`, `notification_root_context` short-circuits on **any** kind 34236 carrying a `d` tag and returns the sender's own coordinate, discarding every `e`/`E`/`a`/`A` tag (`client.rs:1102-1115`, gated at `:1027-1044`, pinned by `video_source_root_context_uses_source_coordinate_not_referenced_a_tag` at `:13780`). So `root_event_id` **is** `source_event_id` for 34236, and a video reply's root is the sender's reply video, not the parent.

The anchor-on-source fix is still correct — just a no-op. The real trigger for a recipient-owned root is a 34236 published *without* a `d` tag, which drops to the generic `a`/`A` path. That's the shape the owner guard defends against, and it's now tested.

Also refuted while I was in here: **"Alice and 1 other mentioned you"** can't happen — funnelcake keys 34236 cards on the video coordinate and collapses with `GROUP BY n.card_key` (`:1077-1090`, `:12123`), so a mention group is always one row, one actor.

### Open

Two threads, both mine, both deliberate:

1. **`_enrichCachedPlaceholders` can't restore a coordinate on a cached mention row.** I wrote the fix and backed it out — the regression test passes about half the time and otherwise hangs to the 10-minute timeout, because the enriched emission never arrives. Which means the fix wouldn't reliably *do* anything either. The whole `cache hydration` group uses `expectLater(stream, emitsThrough(...))`, which stalls rather than fails on a missing emission, so this raciness would have stayed invisible. I'd rather file the coordinate gap and the enrichment race together than ship a coin-flip test.
2. **The `hasCommentTarget` mention branch is inert** on current traffic — `target_comment_id` is only non-empty for kind 1111 sources, and `mention` is reachable only for 34236 and 30023. Coherent forward-compat, worth a line in the description so the next reader doesn't re-derive it.

### Housekeeping

Round-1 threads are all replied to and resolved against their fixing SHAs. The `Tests` job went red once on `999cb2eac` and passed on a re-run of the same commit — pre-existing race in `video_editor_provider_test.dart`, details in the comment above.

@rabble the review block is yours to clear whenever you've had a look at the funnelcake trace.

rabble (commented):
Re-reviewed the current head. The two blockers from my earlier requested-changes review are addressed: video-sourced mentions anchor and fetch metadata from `sourceEventId`, and ordinary mentions can resolve a video target without forcing the comments sheet. The follow-up trust decision now consistently rejects root-derived coordinate, d-tag, title, and thumbnail data together, with regression coverage for foreign owners and unresolved source videos.

I am not clearing the block yet because this branch conflicts with current `main` in four overlapping schema/repository files: the Drift v1 schema, notifications DAO, notification repository, and its test. Those conflicts require substantive notification-model reconciliation, so they are outside safe reviewer takeover. Please rebase onto `origin/main`, regenerate/verify Drift artifacts, rerun the repository and app notification suites, and re-request review on the resulting head.

NotThatKindOfDrLiz:
Takeover update pushed in ee0aea9230.

What changed:
- rebased the PR onto current origin/main and resolved the owner-scoped notification cache plus metadata-fetch conflicts toward behavior already on main
- preserved `_videoMetadataEventIds` as the referenced/root union and added the video-sourced mention arm to it
- fixed video mention groups so `totalCount` and avatars are based on distinct source pubkeys, not edit/source-event count
- restored cached `videoMention` placeholders to source video coordinates from metadata, so stale event-id taps can route by addressable video coordinates again
- regenerated Drift schema artifacts and kept generated checks clean
- fixed the Mobile CI timing-budget test so it recognizes the same sharded job names the checker already supports

Verification:
- `flutter analyze`
- `flutter test test/tools/check_ci_timing_budget_test.dart`
- `flutter test packages/notification_repository/test/src/notification_repository_test.dart`
- `flutter test packages/funnelcake_api_client/test/src/models/relay_notification_test.dart packages/db_client/test/src/database/daos/notifications_dao_test.dart test/notifications/routing/notification_tap_target_test.dart test/main_notification_tap_routing_test.dart test/services/notification_helpers_test.dart test/notifications/view/notifications_view_test.dart test/notifications/widgets/video_notification_row_test.dart test/notifications/widgets/notification_type_icon_spec_test.dart`
- `flutter test --coverage` from `mobile/packages/db_client`
- `very_good test --coverage --min-coverage 40` from `mobile/packages/db_client`
- pre-push hook: analyzer, generated-files check, and changed-file tests
- GitHub checks are green on ee0aea9230

rabble (approved):
Reviewed current head ee0aea9230386f8b3995a3e23e80c2d614b7a71a. The source-video mention anchor, root-coordinate trust checks, cache persistence/rehydration, comment-target routing, and generated Drift schema are internally consistent. All current Analyze, Generated Files, four test shards, build, and format checks pass. Approved for the mobile code path; backend rollout/contract readiness remains a separate operational dependency.

PR #5309, opened 2026-06-18
fix(feed): keep home feed paused behind nested pushed routes

hm21:
Closes #5310

## Description

Closing a fullscreen video opened from a user profile resumed the home feed playback behind the still-open profile. Audio could play while the user was still looking at the profile.

Root cause: GoRouter route information can report the shell location (`/home`) while popping between routes pushed on top of the shell, for example `profile -> fullscreen video -> pop back to profile`. The home feed treated that spurious `/home` report as a genuine return home and resumed.

Fix:
- Add `shellObscuredProvider`, driven by an `AppShell` `RouteAware` subscription to the root navigator. Because the subscription is keyed on the shell route, `didPopNext` fires only when the route directly above the shell is popped. Closing a video above a profile does not reveal the shell; closing the profile does.
- Reset the app-scoped shell-obscured flag when a fresh shell mounts so replace-style auth/navigation flows cannot leave the home feed paused by stale state.
- Make `VideoFeedView` derive playback from combined visibility signals: `home route && !shellObscured && !overlay`, re-evaluated when any of those signals changes.
- Remove an adjacent empty `didChangeDependencies` override while touching the feed view.

This also keeps the previous recorder/editor pause behavior robust because feed playback no longer depends on route location alone during stacked-route pops.

## Out of Scope

None.

## Verification

- `mise exec -- flutter test test/router/app_shell_obscured_test.dart test/screens/feed/video_feed_page_test.dart`
- `mise exec -- flutter analyze lib/providers/shell_obscured_provider.dart lib/router/app_shell.dart lib/screens/feed/video_feed_page.dart test/router/app_shell_obscured_test.dart test/screens/feed/video_feed_page_test.dart`

New tests:
- `app_shell_obscured_test.dart`: shell-obscured state flips only for the route directly above the shell, and a fresh shell clears stale obscured state.
- `video_feed_page_test.dart`: home feed stays paused on a spurious `/home` report while the shell is still obscured, resumes once the shell is revealed, and pauses/resumes across non-home routes.

Manual verification from the author: profile -> open video -> close video keeps the home feed paused; closing the profile resumes it.

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Code refactor
- [ ] Build configuration change
- [ ] Documentation
- [ ] Chore

PR #6301, opened 2026-07-22
fix(router): skip router refresh on routing-identical review-status refetch

### LINKED ISSUE #6302: fix(router): skip router refresh on routing-identical review-status refetch
Auto-created from PR #6301.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6301
**Branch:** `fix/video-player-torn-down-on-resume`

hm21:
Closes #6302

## Description

After backgrounding and resuming the app, tapping a video could open a black/frozen fullscreen player — the tapped video never plays.

**Root cause.** The router refreshes on **every** `currentMinorAccountReviewStatusProvider` emission (`goRouterProvider`). On resume, `accountStateResumeRefresherProvider.refreshOnResume()` re-fetches the minor-account review status (first resume per launch; TTL-gated to 15 min afterwards). `MinorAccountReviewStatus` has no value equality, so even an `active → active` resolve is a *new* emission that fires `refreshListenable.refresh()`.

That refresh churns the whole route pipeline while the pooled video reel is pushed on top via `context.push`: the reported location reverts to the shell branch (`/explore`), and a spurious `RouteAware.didPushNext` flips `FeedVideos._routeAllowsPlayback` to `false` → `isFeedActive` false → `InfiniteVideoFeed` releases the just-created player mid-initialization (`Abort stale init … during error handling`), so the video never plays.

Confirmed against a device log: moderation-status response `09:17:59.004` → location flip `.004` → player disposed `.017` → init aborted `.096`.

**Fix.** Gate the refresh on a routing-relevant projection of the review status — exactly the fields `appRouterRedirect` branches on (`loadingWithoutValue`, `isRestricted`, `hasCase`, moderation-conversation ids, `allowsParentVideoOrEmail`). A routing-identical refetch no longer refreshes; a genuine restriction flip still does. Conservative by design: any change that *could* alter the redirect outcome still triggers a refresh.

## Out of Scope

The underlying GoRouter behavior — an imperatively-pushed route's reported location reverting on `refreshListenable` refresh — is left as-is; this PR removes the needless refresh that exposes it rather than reworking the location provider.

## Verification

- `dart format` ✓ · `flutter analyze lib test` (changed files) ✓
- New unit tests `test/router/app_router_redirect_test.dart` (5/5) ✓
- Regression suites green: `minor_account_review_error_redirect`, `minor_account_review_router` (incl. "keeps restricted routing during a background refetch"), `fullscreen_feed_redirect`, `app_router` — 39/39.
- Manual repro is a narrow race (15-min resume-refetch TTL + fetch must resolve during player init); draft pending on-device confirmation.

**Related Issue:** 

## Type of Change

### DIFF
--- a/mobile/lib/router/app_router_redirect.dart
+++ b/mobile/lib/router/app_router_redirect.dart
@@ -113,6 +113,11 @@ bool _isPublicRecorderLocation(String location) =>
 /// The slice of [MinorAccountReviewStatus] (plus its async loading shape)
 /// that [appRouterRedirect] actually branches on. Two emissions with an
 /// equal signature always produce the same redirect decision.
+///
+/// Keep this in sync with the review-status reads in [appRouterRedirect]: if
+/// the redirect starts branching on another field, add it here or the refresh
+/// gate will swallow genuine changes to it. The coupling is guarded by
+/// `test/router/router_refresh_gate_coupling_test.dart`.
 typedef _MinorReviewRoutingSignature = ({
   bool loadingWithoutValue,
   bool isRestricted,
--- a/mobile/lib/router/app_router_redirect.dart
+++ b/mobile/lib/router/app_router_redirect.dart
@@ -155,6 +155,10 @@ _MinorReviewRoutingSignature _minorReviewRoutingSignature(
 /// and tears down the reel's just-created player mid-initialization. Gating the
 /// refresh on this predicate keeps genuine restriction flips reactive while
 /// leaving an already-open reel untouched.
+///
+/// [previous] is nullable only for defensiveness and the first-emission tests;
+/// the production `ref.listen` at [goRouterProvider] omits `fireImmediately`,
+/// so Riverpod only ever invokes it with a real prior [AsyncValue].
 @visibleForTesting
 bool minorAccountReviewStatusAffectsRouting(
   AsyncValue<MinorAccountReviewStatus>? previous,
--- a/mobile/lib/router/app_router_redirect.dart
+++ b/mobile/lib/router/app_router_redirect.dart
@@ -110,6 +110,55 @@ bool _isAuthEntryLocation(String location) {
 bool _isPublicRecorderLocation(String location) =>
     location == VideoRecorderScreen.path;
 
+/// The slice of [MinorAccountReviewStatus] (plus its async loading shape)
+/// that [appRouterRedirect] actually branches on. Two emissions with an
+/// equal signature always produce the same redirect decision.
+typedef _MinorReviewRoutingSignature = ({
+  bool loadingWithoutValue,
+  bool isRestricted,
+  bool hasCase,
+  String? moderationConversationPubkey,
+  String? moderationConversationId,
+  bool allowsParentVideoOrEmail,
+});
+
+_MinorReviewRoutingSignature _minorReviewRoutingSignature(
+  AsyncValue<MinorAccountReviewStatus>? status,
+) {
+  // `value` (not `asData?.value`) mirrors `appRouterRedirect`, which reads the
+  // previous value that Riverpod carries through a background refetch.
+  final review = status?.value;
+  final reviewCase = review?.currentCase;
+  return (
+    loadingWithoutValue:
+        (status?.isLoading ?? false) && !(status?.hasValue ?? false),
+    isRestricted: review?.isRestricted ?? false,
+    hasCase: reviewCase != null,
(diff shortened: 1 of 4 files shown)


mobile/test/router/app_router_redirect_test.dart
```diff
@@ -0,0 +1,96 @@
...
+        state: MinorReviewCaseState.underModeratorReview,
+        suspectedAgeBand: SuspectedAgeBand.unknown,
+        allowedResolution: MinorReviewResolutionType.unknown,
+        instructions: MinorReviewInstructions(title: '', body: ''),
+        supportEmail: 'support@divine.video',
+        moderationConversationId: 'conversation-1',
+      ),
+    );
+
+    test('does not refresh when a refetch resolves to the same status', () {
```
dcadenas, line 55: Add assertions for both retained-value loading edges here: AsyncData(active) to AsyncLoading with the previous value, and retained loading to fresh AsyncData(active). Require minorAccountReviewStatusAffectsRouting to return false for both transitions, so this test catches regressions that read only asData?.value and restore router refresh during resume.

dcadenas (changes requested):
The resume regression test currently covers only AsyncData-to-AsyncData and misses the retained-value AsyncLoading state emitted during a real background refetch. Add assertions for both loading edges, from active data into retained loading and from retained loading into fresh active data, and verify the gate remains false before merging.

mobile/lib/router/app_router_redirect.dart
```diff
@@ -110,6 +110,60 @@ bool _isAuthEntryLocation(String location) {
...
+/// emission, but a background/resume refetch usually resolves to a routing-
+/// identical value (active → active). Refreshing for those churns the whole
+/// route pipeline while a video reel is pushed on top — reverting the reported
+/// location to the shell branch and firing a spurious `didPushNext` that pauses
+/// and tears down the reel's just-created player mid-initialization. Gating the
+/// refresh on this predicate keeps genuine restriction flips reactive while
+/// leaving an already-open reel untouched.
+@visibleForTesting
+bool minorAccountReviewStatusAffectsRouting(
+  AsyncValue<MinorAccountReviewStatus>? previous,
```
realmeylisdev, line 164: Nit: `previous` can't actually be `null` at the one production call site (`app_router.dart:70`) — that `ref.listen` doesn't pass `fireImmediately: true`, and Riverpod's `_handleFireImmediately` no-ops in that case, so the listener only ever fires with a real prior `AsyncValue`. Harmless as defensive code, but the "first emission refreshes only when it can gate routing" test documents a path that can't happen through the real wiring — worth a one-line comment here so a future reader doesn't assume it's reachable.

realmeylisdev (approved):
Traced the root cause through the actual code (not just the PR description) and it checks out — the routing signature covers exactly the fields `appRouterRedirect` branches on, nothing more and nothing less. Ran the new tests plus all four cited regression suites locally: 45/45 green, format/analyze clean. Also sabotage-tested the coupling test (made the redirect depend on an untracked field) and confirmed it actually fails with a useful message, so it's a real guard, not just a comment.

One small non-blocking note inline. Also worth a callout: this is the second time the GoRouter push/pop location-reversion quirk has caused a real bug — #5309 patched the pop side (`shellObscuredProvider`), this patches the push side. Might be worth a tracking issue for the underlying quirk before it bites a third route.

mobile/test/router/app_router_redirect_test.dart
hm21: Done in afeef4400. Added two tests that build a real retained-value loading emission (via a `ref.watch` reload) and assert the gate stays false across both edges (data → retained-loading → data). Active covers the `!hasValue` half of `loadingWithoutValue`; restricted covers the `.value` read — both fail under an `asData?.value` regression, sabotage-verified.

mobile/lib/router/app_router_redirect.dart
hm21: Added the note in afeef4400 — `previous` is nullable only defensively; the production `ref.listen` omits `fireImmediately`, so it never fires with null. Good call on the underlying GoRouter quirk too.

hm21:
@dcadenas addressed in afeef4400 — added retained-value loading-edge coverage (active guards the `!hasValue` clause, restricted guards the `.value` read; both sabotage-verified to fail under their regression). Ready for another look.

dcadenas (approved):
✅ identical routes now avoid extra refreshes

PR #6303, opened 2026-07-22
fix(video-metadata): relabel C2PA "Post anyway" to "Skip"

### LINKED ISSUE #6305: fix(video-metadata): relabel C2PA "Post anyway" to "Skip"
Auto-created from PR #6303.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6303
**Branch:** `fix/c2pa-missing-skip-button`

hm21:
Closes #6305

## What

Renames the secondary button on the C2PA missing-content-credential
bottom sheet (shown after a render when signing failed, #6058) from
**"Post anyway"** to **"Skip"** across all 18 locales.

## Why

The button's handler is `acknowledgeC2paSigningFailure()`, which only
sets `c2paSigningFailed = false` and dismisses the sheet — it does **not**
publish anything. The user lands back on the metadata form and still has
to post separately.

Labeling it "Post anyway" made users expect an immediate post; when
nothing happened, the button read as broken/confusing. "Skip" describes
what actually happens — skip the content-credential step and continue —
and pairs cleanly against the primary "Regenerate" action.

## Scope

- Value updated in all `app_*.arb` locales + regenerated
  `lib/l10n/generated/`.
- To avoid a misleading internal name, the l10n key is renamed
  (`videoMetadataC2paMissingPostAnyway` → `videoMetadataC2paMissingSkip`)
  and the private `_C2paMissingChoice.postAnyway` enum case → `.skip`,
  with the ARB description and adjacent code comment updated to match.
- Stale test copy in `video_metadata_screen_test.dart` aligned.
- The sheet **title/body** wording ("post it as-is") is intentionally
  left untouched — changing those translated body strings as a review
  side effect would drift the English source from 16+ locales, which the
  copy-alignment policy discourages. Happy to do that as a separate,
  deliberate copy pass if product wants it.

## Verification

- `flutter test test/l10n/arb_consistency_test.dart` ✅
- `flutter test test/screens/video_metadata_screen_test.dart` ✅ (9/9)
- `flutter analyze lib/l10n lib/screens/video_metadata test/screens/video_metadata_screen_test.dart` ✅

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -3895,7 +3895,7 @@
     "videoMetadataC2paMissingBody": "የይዘት መታወቂያዎችን መጨመር አልቻልንም፣ ስለዚህ ይህ ቪዲዮ በሰው እንደተሰራ አይረጋገጥም። እንደገና ለመሞከር እንደገና ይፍጠሩ ወይም እንዳለ ይለጥፉ።",
     "videoMetadataC2paMissingNote": "የይዘት መታወቂያዎች የበይነመረብ ግንኙነት ያስፈልጋቸዋል።",
     "videoMetadataC2paMissingRegenerate": "እንደገና ፍጠር",
-    "videoMetadataC2paMissingPostAnyway": "ለማንኛውም ለጥፍ",
+    "videoMetadataC2paMissingSkip": "ዝለል",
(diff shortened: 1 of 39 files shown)


dcadenas (approved):
✅ clearer C2PA skip wording, ship it

PR #6306, opened 2026-07-22
fix(branding): guard the sign-in Divine wordmark against brand-mark drift

realmeylisdev:
Works the **Login / auth logo** bullet of #6282.

## What I found

The premise for this bullet doesn't hold: the auth hero is **already on the current logotype**. `assets/icon/logo.svg` is the geometric heart-V wordmark filled `#27C58B`, which is exactly `VineTheme.vineGreen`. No asset refresh was needed.

What *was* missing is the guard. The only logo assertion was:

```dart
expect(find.byType(SvgPicture), findsNWidgets(5));
```

A count. It passes identically if the asset is swapped for one of the retired cursive brand-pack files, which is precisely the blindness that let #6290 ship the old wordmark to the download watermark. The highest-visibility brand surface in the app had no way to fail.

## What changed

**Pin the mark at both levels**, mirroring the technique already used in `watermark_image_generator_test.dart`:

- **Call site** - assert `SvgAssetLoader.assetName`, so swapping the path fails.
- **Asset itself** - assert `viewBox` aspect ratio (~3.79:1, vs the retired pack's ~2.02:1) and that every 6-digit `fill` attribute equals `VineTheme.vineGreen`, so swapping or recolouring `logo.svg`'s current contents fails too.

**Accessibility** - the wordmark now carries a screen-reader label sourced from `AppConfig.appName`, and the four decorative emoji stickers use `excludeFromSemantics: true` so the wordmark is the only image exported to the platform accessibility tree. A widget test now walks the semantics tree and pins that image-node count.

The label deliberately stays **out of the ARB catalog** (review feedback, 49b5632da). It is a fixed proper noun, and a translation pipeline will transliterate it; the `@description "never translate"` note doesn't reach most vendors. That risk is already realised here: of the four standalone `"Divine"` ARB keys on `main`, `videoEditorAudioCategoryDivine` reads `"Divine"` in `en`/`fil` and `"OG Sounds"` in the other 16 locales. A test pumps the hero under `Locale('ar')` and asserts the announced string is unchanged, so a move back into l10n fails.

Follow-up review cleanup in 84b980960:

- Reused `DivineIconName.logo.assetPath` at the widget call site while keeping the literal path in the test as the tripwire.
- Replaced the sticker `ExcludeSemantics` wrapper with `SvgPicture.asset(..., excludeFromSemantics: true)`.
- Added a semantics-tree assertion that only one image node, the wordmark, is exported.
- De-hardcoded the sibling `welcome_screen_font_test.dart` tagline/font assertions.

## Verification

Each original guard was mutation-tested by the author, confirmed red before landing, not just green after:

| Mutation | Result |
|---|---|
| call-site path -> `White cropped.png` | fails |
| `semanticsLabel` removed | fails (2 tests) |
| label made locale-dependent (simulates an ARB-sourced value) | fails (locale test only) |
| `logo.svg` contents -> `White on black.svg` | fails (2 tests) |
| `logo.svg` recoloured off-brand | fails |
| baseline | passes |

Current takeover verification:

- `flutter test test/widgets/auth/auth_hero_section_test.dart test/widgets/welcome_screen_font_test.dart`
- `flutter test test/widgets/auth test/screens/welcome_screen_test.dart test/widgets/welcome_screen_font_test.dart test/l10n/arb_consistency_test.dart test/services/watermark_image_generator_test.dart` (75 tests)
- Pre-push hook: no merge conflicts with `main`, analyzer clean, changed-file tests pass

## Left out, deliberately

**`web_auth_screen.dart:221` is not touched because it is dead code.** `WebAuthScreen` is never routed, constructed, or imported anywhere in the repo; the only hits outside the file are two ratchet baselines and a coverage doc. Two design docs already record this independently (`docs/superpowers/specs/2026-05-05-web-nip07-signin-design.md:22`, `docs/superpowers/plans/2026-05-05-web-nip07-signin.md:1660`). So the issue's "high-visibility (every sign-in)" framing applies only to the hero, and spending brand or a11y effort on unreachable UI would be a fix nobody can see. Deleting the screen is a product call worth its own PR.

Two things for #6282's remaining bullets, found while auditing:

- `DivineIconName.logo` also resolves to `assets/icon/logo.svg` and renders in every system notification row (`notification_type_icon_spec.dart:53`). Any repo-wide brand sweep must include the enum, not just string literals.
- The wordmark only renders in `_NewUserLayout`; the returning-user layout shows no mark. "Every sign-in" is really "first run / no stored account."

Refs #6282

### DIFF
--- a/mobile/lib/widgets/auth/auth_hero_section.dart
+++ b/mobile/lib/widgets/auth/auth_hero_section.dart
@@ -4,6 +4,7 @@
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_svg/flutter_svg.dart';
+import 'package:openvine/config/app_config.dart';
 import 'package:openvine/l10n/l10n.dart';
 
 /// Hero section with large tagline text and decorative 3D emoji stickers.
@@ -105,10 +106,12 @@ class AuthHeroSection extends StatelessWidget {
           // heart-V); the cursive brand-pack files in assets/icon/ are the
           // retired one. Pinned by auth_hero_section_test.dart so a swap back
           // to the old mark fails CI instead of shipping (see #6282).
+          // The label is the brand name, not copy: it stays out of the ARB
+          // catalog so no locale can translate or transliterate it.
           SvgPicture.asset(
             'assets/icon/logo.svg',
             width: 120,
-            semanticsLabel: context.l10n.authHeroLogoSemanticLabel,
+            semanticsLabel: AppConfig.appName,
           ),
         ],
       ),
--- a/mobile/lib/widgets/auth/auth_hero_section.dart
+++ b/mobile/lib/widgets/auth/auth_hero_section.dart
@@ -101,8 +101,15 @@ class AuthHeroSection extends StatelessWidget {
 
           const SizedBox(height: 40),
 
-          // Divine wordmark (green SVG logo)
-          SvgPicture.asset('assets/icon/logo.svg', width: 120),
+          // Divine wordmark. `logo.svg` is the current logotype (geometric
+          // heart-V); the cursive brand-pack files in assets/icon/ are the
+          // retired one. Pinned by auth_hero_section_test.dart so a swap back
+          // to the old mark fails CI instead of shipping (see #6282).
+          SvgPicture.asset(
+            'assets/icon/logo.svg',
+            width: 120,
+            semanticsLabel: context.l10n.authHeroLogoSemanticLabel,
+          ),
         ],
       ),
     );
@@ -118,6 +125,13 @@ class _StickerImage extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return SvgPicture.asset(path, width: size, height: size);
+    // Decorative: a bare SvgPicture still exports an `isImage` node with an
+    // empty label to the platform accessibility tree. Flutter's own traversal
+    // skips those, so this is tree hygiene rather than a fix for an audible
+    // bug — but it keeps the four stickers out of what we hand to TalkBack
+    // and VoiceOver, leaving the wordmark as the only image on the screen.
+    return ExcludeSemantics(
+      child: SvgPicture.asset(path, width: size, height: size),
+    );
   }
 }
--- a/mobile/lib/widgets/auth/auth_hero_section.dart
+++ b/mobile/lib/widgets/auth/auth_hero_section.dart
@@ -102,14 +102,12 @@ class AuthHeroSection extends StatelessWidget {
 
           const SizedBox(height: 40),
 
-          // Divine wordmark. `logo.svg` is the current logotype (geometric
-          // heart-V); the cursive brand-pack files in assets/icon/ are the
-          // retired one. Pinned by auth_hero_section_test.dart so a swap back
-          // to the old mark fails CI instead of shipping (see #6282).
+          // Divine wordmark. `logo.svg` is the current geometric heart-V mark;
+          // auth_hero_section_test.dart pins it against retired brand assets.
           // The label is the brand name, not copy: it stays out of the ARB
           // catalog so no locale can translate or transliterate it.
           SvgPicture.asset(
-            'assets/icon/logo.svg',
+            DivineIconName.logo.assetPath,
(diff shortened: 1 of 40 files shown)


mobile/lib/widgets/auth/auth_hero_section.dart
```diff
@@ -101,8 +101,15 @@ class AuthHeroSection extends StatelessWidget {
...
-          // Divine wordmark (green SVG logo)
-          SvgPicture.asset('assets/icon/logo.svg', width: 120),
+          // Divine wordmark. `logo.svg` is the current logotype (geometric
+          // heart-V); the cursive brand-pack files in assets/icon/ are the
+          // retired one. Pinned by auth_hero_section_test.dart so a swap back
+          // to the old mark fails CI instead of shipping (see #6282).
+          SvgPicture.asset(
+            'assets/icon/logo.svg',
+            width: 120,
+            semanticsLabel: context.l10n.authHeroLogoSemanticLabel,
```
hm21: `AppConfig.appName` (`lib/config/app_config.dart:42`) is already the single source of truth: `static const String appName = 'Divine'`. Use `semanticsLabel: AppConfig.appName` and drop the new `authHeroLogoSemanticLabel` key, its 18-locale mirror, and the regenerated files.

A fixed brand name in the ARB catalog is pure noise — and worse, a human translator or a translation vendor will translate or transliterate it (`ديفاين`, `ディバイン`); the `@description "never translate"` note isn't surfaced in most translation pipelines.

hm21 (commented):
One point, the brand name shouldn't live in l10n — reuse the existing `AppConfig.appName` const instead of a new 18-locale key (see inline). The rest (asset-content pinning, sticker `ExcludeSemantics`) is solid.

Team note: the l10n-first guideline misfires on a fixed proper noun here, so you went by the book. We'll add a brand/proper-noun carve-out to `localization.md` so this stops steering people (and agents) into mirroring the app name across every ARB.

realmeylisdev:
Simulator Screenshot - iPhone 17 Pro Max - 2026-07-22 at 18 19 09

mobile/lib/widgets/auth/auth_hero_section.dart
realmeylisdev: Agreed and applied in 49b5632da — `semanticsLabel: AppConfig.appName`, and the key, its 18-locale mirror, and the regenerated files are gone. The PR is down to two files.

Your risk isn't hypothetical here: the repo already has four standalone `"Divine"` ARB keys, and one of them has drifted — `videoEditorAudioCategoryDivine` reads `"Divine"` in `en`/`fil` and `"OG Sounds"` in the other 16 locales. Exactly the failure mode you described, already shipped.

Added a test that pumps the hero under `Locale('ar')` and asserts the announced string is still `AppConfig.appName`, so a move back into the catalog fails. Mutation-checked: dropping `semanticsLabel` reddens both semantics tests; making the label locale-dependent reddens only the new one.

hm21 (approved):
LGTM

NotThatKindOfDrLiz:
@realmeylisdev I pushed a small takeover follow-up in 84b980960 to close out the remaining review items:

- switched the decorative sticker SVGs to `excludeFromSemantics: true`
- added a semantics-tree test that verifies only the wordmark exports as an image
- reused `DivineIconName.logo.assetPath` at the widget call site while keeping the hardcoded path in the test guard
- de-hardcoded the sibling `welcome_screen_font_test.dart` tagline/font assertions
- updated the PR body and recorded the dead `WebAuthScreen` decision on #6282

Verification:

- `flutter test test/widgets/auth/auth_hero_section_test.dart test/widgets/welcome_screen_font_test.dart`
- `flutter test test/widgets/auth test/screens/welcome_screen_test.dart test/widgets/welcome_screen_font_test.dart test/l10n/arb_consistency_test.dart test/services/watermark_image_generator_test.dart` (75 tests)
- pre-push hook passed analyzer and changed-file tests
- GitHub checks are green on 84b980960

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up. 84b980960 addresses the semantics/test cleanup, the PR body and #6282 are updated, and GitHub checks are green.

PR #6307, opened 2026-07-22
feat(captions): add caption_generator package for on-device CC generation

### LINKED ISSUE #6308: feat(captions): add caption_generator package for on-device CC generation
Tracks landing the `caption_generator` capability package for on-device
closed-caption generation.

## Scope

- Add a local `mobile/packages/caption_generator` package.
- Generate word-level caption segments from an audio file using OS speech
  recognizers:
  - iOS/macOS: `SFSpeechRecognizer`.
  - Android 14+: platform `SpeechRecognizer` with `EXTRA_AUDIO_SOURCE` file
    input and `RecognitionPart` timing.
- Convert Android WAV input to 16 kHz mono PCM16 in Dart with bounded parsing,
  size/duration limits, isolate-backed conversion, and typed failures.
- Group word-level segments into display-ready caption cues.
- Keep the package in the workspace only; do not add it as an app dependency or
  ship app speech-recognition purpose strings until the UI wiring PR.

## Acceptance

- Dart package tests pass with the package coverage floor still at 100%.
- Android native unit tests compile and run in CI.
- The Android no-word-parts fallback returns a transcript spanning the full
  audio when the recognizer provides text but no word timing.
- Platform errors map to readable typed Dart exceptions.
- README documents the Android `RECORD_AUDIO` requirement for consuming apps.
- The public Dart entrypoint remains importable from web builds and throws
  `UnsupportedError` at runtime on unsupported platforms.

## Follow-ups before UI wiring

- Add a recognition watchdog/cancel path on Android and Darwin so a wedged OS
  recognizer cannot hang the caller or leak temporary Android WAV files.
- Add Apple native compile coverage for the Darwin plugin code.
- In the app wiring PR, add the app dependency, Runner
  `NSSpeechRecognitionUsageDescription` strings, permission UX, and language
  selection intentionally.

hm21:
Closes #6308

## What

New local package `caption_generator`: on-device speech-to-text closed-caption
generation for Android, iOS, and macOS. Feed it an audio file — typically the
WAV that `pro_video_editor` audio extraction produces — and it returns a list
of timed text segments (`text` + start/end `Duration`), which
`groupCaptionSegments()` merges into display-ready caption cues.

This is deliberately a **last-resort, best-effort** path. On-device recognizers
are imperfect, so the result is meant to be surfaced as an editable suggestion,
never as authoritative captions.

## Why these design choices

- **OS recognizers instead of a bundled engine.** The first cut used Vosk, which
  would have added ~10 MB of native libs to the APK plus a ~40 MB per-language
  model download. Both platforms already ship a file-transcription API, so the
  package uses them instead and adds **zero model download and no APK weight**
  until the app wiring PR depends on it:
  - **iOS / macOS**: `SFSpeechRecognizer` (on-device when the locale supports it).
  - **Android 14+**: the platform `SpeechRecognizer` via
    `createOnDeviceSpeechRecognizer` + `EXTRA_AUDIO_SOURCE` (file input, API 33)
    + `RecognitionPart` word timing (API 34). Devices below Android 14, devices
    without Google's on-device recognition service, and missing language packs
    throw `SpeechRecognizerUnavailableException` — i.e. "CC suggestion not
    available on this device", which is the correct behavior for a fallback.
- **Pure-Dart WAV preprocessor.** Android's recognizer wants 16 kHz mono 16-bit
  PCM, so extracted WAVs (16-bit int or 32-bit float, any rate/channels) are
  converted in Dart. Kept out of native code so it is unit-testable without a
  device.
- **Cues break at sentence boundaries.** Recognizer word timings smear speech
  pauses away (a word's duration is often stretched to the next word's start),
  so grouping on the silence gap alone cut cues mid-sentence.
  `groupCaptionSegments` splits on sentence-final punctuation instead, which
  both platforms emit (iOS/macOS via `addsPunctuation`, Android via
  `EXTRA_ENABLE_FORMATTING`).
- **Typed failure surface.** All failures are typed
  `CaptionGenerationException` subclasses so callers can branch per failure
  mode; an empty list means "no speech recognized".

## Layout

- Dart layer: facade + platform interface + method channel + WAV preprocessor +
  cue grouper + models.
- Android: Kotlin plugin (`SpeechRecognizer` file transcription).
- Darwin: shared Swift plugin (`SFSpeechRecognizer`) for iOS + macOS.
- App wiring: workspace entry in `mobile/pubspec.yaml` only. The app dependency,
  Runner speech-recognition purpose strings, and UI permission flow are left for
  the wiring PR so the P1 build does not ship an unreachable speech capability.
- CI: per-package Dart workflow plus package-local Android native unit tests.

## Not included

- Model/permission wiring into an actual UI flow. This PR ships the capability
  package only; hooking it into the video editor is a follow-up in the app layer,
  where the user's app language should be passed as `localeIdentifier` rather
  than relying on the device default.
- A native recognition watchdog/cancel path. That should land before UI wiring
  so a wedged platform recognizer cannot leave callers waiting forever.
- Apple native compile CI. The Android native compile/test job is included here;
  a Darwin compile check needs a dedicated follow-up because this package uses a
  Flutter plugin wrapper rather than a standalone Swift package app target.

## Testing

- Dart unit tests with 100% line coverage (WAV parsing/conversion incl. a header
  round-trip regression, cue grouping incl. sentence splitting, error-code
  mapping, per-platform routing, unsupported-platform shim).
- Android native unit tests for support gating, segment timing, and the no-word
  parts fallback.
- `flutter analyze` clean for `caption_generator`.
- Package-local Android test command:
  `gradle testDebugUnitTest` from `mobile/packages/caption_generator/android`.


mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
```diff
@@ -0,0 +1,243 @@
...
+      throw AudioFileNotFoundException(inputPath);
+    }
+    final wav = _ParsedWav.parse(await file.readAsBytes());
+    if (wav.sampleRate == targetSampleRate &&
+        wav.channels == 1 &&
+        !wav.isFloat32 &&
+        !wav.wasExtensible) {
+      return inputPath;
+    }
+    final mono = wav.decodeMonoSamples();
```
dcadenas: Bound and isolate WAV preprocessing before allocating buffers. `prepareForRecognition` reads and converts the entire noncanonical file on the caller isolate here, while `_resampleLinear` sizes output from the unchecked `sampleRate`, so a long clip can stall the UI and a malformed low-rate header can cause excessive allocation. Validate sample-rate, frame-count, duration, and output-byte limits before decoding, throw `UnsupportedAudioFormatException` for rejected inputs, and run decode, downmix, resample, and encode in `Isolate.run` or an equivalent worker. Keep the exact 16 kHz mono PCM16 passthrough unchanged, and add tests for low-rate rejection and caller-isolate responsiveness.

mobile/packages/caption_generator/android/src/main/kotlin/co/openvine/caption_generator/TranscriptionSession.kt
```diff
@@ -0,0 +1,259 @@
...
+                    RecognizerIntent.EXTRA_ENABLE_FORMATTING,
+                    RecognizerIntent.FORMATTING_OPTIMIZE_QUALITY,
+                )
+                putExtra(
+                    RecognizerIntent.EXTRA_SEGMENTED_SESSION,
+                    RecognizerIntent.EXTRA_AUDIO_SOURCE,
+                )
+            }
+            recognizer.setRecognitionListener(this)
+            recognizer.startListening(intent)
```
dcadenas: Check support for the complete file-audio intent before calling `startListening`. If the installed recognizer does not support `EXTRA_AUDIO_SOURCE`, Android can fall back to the microphone, so a file-caption request may capture ambient speech when `RECORD_AUDIO` is already granted. Start the writer and recognizer only after `checkRecognitionSupport` succeeds; otherwise close both pipe ends, destroy the recognizer, and return the typed `recognizer_unavailable` result. Add supported and unsupported callback coverage.

mobile/packages/caption_generator/lib/caption_generator.dart
```diff
@@ -0,0 +1,127 @@
...
+    };
+  }
+
+  Future<List<CaptionSegment>> _generateOnAndroid({
+    required String audioPath,
+    required String? localeIdentifier,
+  }) async {
+    final preparedPath = await WavPreprocessor.prepareForRecognition(
+      inputPath: audioPath,
+      outputPath: '$audioPath$_preparedWavSuffix',
```
dcadenas: Give each conversion its own temporary file and map preparation I/O failures into the package's typed exceptions. The hard-coded sibling path here is shared by overlapping calls, while `readAsBytes` and `writeAsBytes` failures in `wav_preprocessor.dart:48` and `:57` escape the documented `CaptionGenerationException` surface. Have `WavPreprocessor` create a unique writable temp artifact, return ownership to the facade, clean up only that artifact in `finally`, and leave canonical input untouched. Add concurrent same-source and read/write failure tests.

dcadenas (dismissed):
Please address the three blocking issues before merging: bound and isolate WAV conversion, validate Android file-input support, and give each conversion a typed, per-call temporary-file lifecycle. These changes are needed to prevent UI stalls and unsafe allocations, avoid microphone fallback, and keep concurrent requests and filesystem failures within the package contract.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
hm21: Done in 2649299. Decode/downmix/resample/encode now run in a worker isolate via Isolate.run, and assertConvertible rejects out-of-range sample rates and over-limit converted-frame counts before any buffer is sized. The 16 kHz mono PCM16 passthrough is unchanged. Added tests for low-rate rejection and for verifying the conversion runs off the caller isolate.

mobile/packages/caption_generator/android/src/main/kotlin/co/openvine/caption_generator/TranscriptionSession.kt
hm21: Done in 2649299. start() now gates on checkRecognitionSupport for the complete file-audio intent; only on success does it start the writer and call startListening. On an unsupported result it closes both pipe ends, destroys the recognizer, and returns recognizer_unavailable, so a recognizer that ignores EXTRA_AUDIO_SOURCE can no longer fall back to the microphone. Added Robolectric coverage for the supported and unsupported callbacks.

mobile/packages/caption_generator/lib/caption_generator.dart
hm21: Done in 2649299. WavPreprocessor now creates a unique per-call temp directory and returns ownership to the facade via PreparedRecognitionAudio, disposed in finally; the shared sibling path is gone and the input is left untouched. readAsBytes/writeAsBytes failures are mapped to TranscriptionFailedException. Added concurrent same-source and read/write failure tests.

hm21:
@dcadenas all three blocking findings are addressed in 2649299 and the threads are resolved: WAV prep is now bounded + isolated, the Android session gates on `checkRecognitionSupport` (no mic fallback), and each conversion owns a unique temp file with typed I/O errors. Dart coverage stays 100% and the new Robolectric tests pass locally. PTAL.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
```diff
@@ -0,0 +1,402 @@
...
+
+  static Future<bool> _runPreparation(
+    String inputPath,
+    String outputPath,
+    int maxConvertedFrames,
+  ) async {
+    debugConversionRanInCallerIsolate = true;
+    final Uint8List bytes;
+    try {
+      bytes = await File(inputPath).readAsBytes();
```
dcadenas: Validate the WAV metadata and size limits before `readAsBytes()` at `mobile/packages/caption_generator/lib/src/wav_preprocessor.dart:151`. `_runPreparation` currently allocates the entire file, then parses it and checks bounds, so a large or corrupt WAV can exhaust the app process before `UnsupportedAudioFormatException` is thrown. Use bounded RIFF metadata parsing before payload allocation, and add a regression test for oversized or sparse input while preserving canonical passthrough and conversion behavior.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
```diff
@@ -0,0 +1,402 @@
...
+  ///   range, or the clip is too long to prepare.
+  /// * [TranscriptionFailedException] if reading the input or writing the
+  ///   converted output fails.
+  static Future<PreparedRecognitionAudio> prepareForRecognition({
+    required String inputPath,
+  }) async {
+    if (!File(inputPath).existsSync()) {
+      throw AudioFileNotFoundException(inputPath);
+    }
+    final workDirectory = await Directory.systemTemp.createTemp(
```
dcadenas: Guard `Directory.systemTemp.createTemp` at `mobile/packages/caption_generator/lib/src/wav_preprocessor.dart:111` with the same filesystem-error mapping used for later reads and writes. If temp-directory creation fails, callers currently receive a raw `FileSystemException` instead of the package's typed `CaptionGenerationException` surface. Add a deterministic creation-failure test and preserve the existing ownership and cleanup behavior.

dcadenas (dismissed):
Before this can merge, the WAV preprocessor must reject oversized inputs before allocating the full file and must map temp-directory creation failures to the package's typed caption errors. Both gaps are in the Android preprocessing path and need targeted regression coverage.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
hm21: Done in 689ffcd. Replaced readAsBytes with a bounded RandomAccessFile walk (_WavHeader.readFrom) that reads only chunk headers and the fmt body — never the payload. Sample rate, the data chunk declared size (new 64 MiB payload cap), and the converted-frame count are all validated before any payload buffer is allocated, so an oversized or sparse WAV is rejected up front. Added a regression test that rejects a tiny file whose data chunk declares ~2 GiB, plus the existing passthrough/conversion tests still pass.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
hm21: Done in 689ffcd. Directory.systemTemp.createTemp is now wrapped and its FileSystemException maps to TranscriptionFailedException, matching the read/write typed-exception surface; ownership and cleanup are unchanged. Added a deterministic creation-failure test via an injected temp-dir factory that throws.

hm21:
@dcadenas both follow-ups are in 689ffcd and resolved: WAV metadata is now parsed with bounded `RandomAccessFile` reads and all size limits (incl. a 64 MiB payload cap on the declared data-chunk size) are checked before any payload is allocated, and `createTemp` failures map to `TranscriptionFailedException`. Added regression tests for oversized/sparse input and temp-dir creation failure; coverage stays 100%.

mobile/packages/caption_generator/android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,9 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <!-- Package visibility (Android 11+): allow resolving the on-device
+       speech recognition service used for file transcription. -->
+  <queries>
+    <intent>
+      <action android:name="android.speech.RecognitionService" />
+    </intent>
+  </queries>
```
realmeylisdev, line 8: The plugin manifest is missing `RECORD_AUDIO`. AOSP `SpeechRecognizer` class doc:

> Please note that the application must have `android.Manifest.permission.RECORD_AUDIO` permission to use this class.

That applies to `createOnDeviceSpeechRecognizer` + `startListening` regardless of `EXTRA_AUDIO_SOURCE` — the file source replaces the mic, not the permission. Right now the package only works because `mobile/android/app/src/main/AndroidManifest.xml:10` happens to declare it; a plugin that owns a capability should declare its own requirement.

The runtime half matters more for Divine: `RECORD_AUDIO` is a dangerous permission and the app only requests it via the camera flow (`camera_permission_bloc.dart`). A user who imports a gallery clip and asks for captions without ever having recorded gets `ERROR_INSUFFICIENT_PERMISSIONS` → `not_authorized` → `SpeechNotAuthorizedException`, which the Dart docs say cannot happen on Android.

Suggested: add `<uses-permission android:name="android.permission.RECORD_AUDIO" />` here, and have the follow-up UI PR check/request the grant before offering the CC action.

mobile/packages/caption_generator/README.md
```diff
@@ -0,0 +1,80 @@
...
+```xml
+<key>NSSpeechRecognitionUsageDescription</key>
+<string>...</string>
+```
+
+The first call prompts the user for speech recognition permission.
+
+### Android setup
+
+No setup or model download required. The plugin manifest already declares
```
realmeylisdev: "No setup or model download required" isn't accurate for Android — the model download part is, but a consuming app must declare `RECORD_AUDIO` **and** hold the runtime grant (see the manifest comment). Worth spelling out here since this is the package's contract for any future consumer.

mobile/packages/caption_generator/lib/src/exceptions.dart
```diff
@@ -0,0 +1,75 @@
...
+/// 32-bit float PCM (what `pro_video_editor` audio extraction produces).
+final class UnsupportedAudioFormatException extends CaptionGenerationException {
+  /// Creates the exception with a [message] describing the offending format.
+  const UnsupportedAudioFormatException(String message)
+    : super(message, 'UnsupportedAudioFormatException');
+}
+
+/// The user denied (or the OS restricts) speech recognition authorization.
+///
+/// Apple platforms only — prompt the user towards the system settings.
```
realmeylisdev: "Apple platforms only" is contradicted by the Android path: `TranscriptionSession.onError` maps `ERROR_INSUFFICIENT_PERMISSIONS` → `not_authorized`, and `_mapPlatformException` turns that into this exception. Same drift in `caption_generator.dart:58` ("if the user denied speech recognition (Apple platforms)").

Since a missing `RECORD_AUDIO` grant is the most likely way an Android user hits this, the doc should name it — that's the signal the UI layer needs to prompt for the mic permission rather than show a generic failure.

mobile/packages/caption_generator/android/src/main/kotlin/co/openvine/caption_generator/TranscriptionSession.kt
```diff
@@ -0,0 +1,373 @@
...
+        try {
+            fd?.close()
+        } catch (_: IOException) {
+            // Already closed by the recognizer or writer; nothing to release.
+        }
+    }
+
+    private fun buildSegments(): List<Map<String, Any>> =
+        words.mapIndexed { index, (text, startMs) ->
+            val endMs = words.getOrNull(index + 1)?.second ?: audioDurationMs
```
realmeylisdev: `?: audioDurationMs` makes the last word's cue run to the end of the clip, and because word ends are contiguous by construction the grouper's `mergedDuration > maxCaptionDuration` rule then splits that last word off into its own cue. Net effect on anything longer than ~5s of trailing silence: a one-word caption pinned on screen until the video ends.

Reproduced against the real `groupCaptionSegments` with Android-shaped timings (30s clip, speech ending at 3.2s):

```
0ms -> 3200ms   "Hello there, this is a short test"
3200ms -> 30000ms  "clip."
```

Apple isn't affected — `segment.timestamp + segment.duration` gives a real end.

Suggested: clamp the last word to something like `min(audioDurationMs, startMs + medianInterWordDelta)` (or a fixed per-word fallback) instead of stretching it to the full duration.

mobile/packages/caption_generator/android/src/main/kotlin/co/openvine/caption_generator/TranscriptionSession.kt
```diff
@@ -0,0 +1,373 @@
...
+
+    /** Collected (text, startMs) pairs; end times are derived in [finish]. */
+    private val words = mutableListOf<Pair<String, Long>>()
+    private var completed = false
+
+    /** Total audio duration, used as the end time of the last word. */
+    private val audioDurationMs: Long =
+        wav.dataLength * 1000 / (2L * wav.sampleRate)
+
+    fun start() {
```
realmeylisdev, line 134: There's no watchdog anywhere in the stack — `grep -rn "Timer|timeout|postDelayed|Handler("` over `lib/`, `android/src/` and `darwin/` comes back empty. `MethodChannel.Result` is only ever fulfilled from a recognizer callback, so if the recognition service dies mid-session (or `checkRecognitionSupport` never invokes either callback), nothing completes: the Dart future hangs, `prepared.dispose()` in the facade's `finally` never runs so the converted WAV leaks, and the writer thread stays blocked on the pipe.

The Apple side has the same shape if the task yields neither `isFinal` nor an error.

Since this is a user-facing "generate captions" action, a bounded wait that fails with `transcription_failed` would be worth having before it's wired to a spinner.

mobile/packages/caption_generator/android/src/main/kotlin/co/openvine/caption_generator/TranscriptionSession.kt
```diff
@@ -0,0 +1,373 @@
...
+            val pipe = pipeFactory()
+            sourceFd = pipe[0]
+            sinkFd = pipe[1]
+            val intent = buildIntent(pipe[0])
+            recognizer.setRecognitionListener(this)
+            // Gate on support: an unsupported EXTRA_AUDIO_SOURCE intent can
+            // make some recognizers open the microphone instead of the file.
+            recognizer.checkRecognitionSupport(
+                intent,
+                onSupported = { onRecognitionSupported(intent) },
```
realmeylisdev, line 147: The support gate itself is the right call — `EXTRA_AUDIO_SOURCE`'s doc is explicit that "if this extra is not set or the recognizer does not support this feature, the recognizer will open the mic".

Minor: `onSupportResult` discards the `RecognitionSupport` payload. It fires even when the requested language is only in `getPendingOnDeviceLanguages()` rather than `getInstalledOnDeviceLanguages()`, so those runs still build the pipe, start the writer and start listening, only to come back through `onError` as `ERROR_LANGUAGE_UNAVAILABLE`. Checking the installed list here would let you short-circuit to `recognizer_unavailable` before any of that setup.

.github/workflows/caption_generator.yaml
```diff
@@ -0,0 +1,58 @@
...
+jobs:
+  build:
+    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/flutter_package.yml@v1
+    with:
+      working_directory: "mobile/packages/caption_generator"
+
+  # The Dart workflow above never compiles the native Kotlin, so a compile
+  # error would pass green until the package is wired into an app build. This
+  # job compiles the plugin's Android sources (mirrors divine_camera).
+  android-compile:
```
realmeylisdev, line 31: The rationale in the comment above ("the Dart workflow never compiles the native Kotlin, so a compile error would pass green") applies at least as strongly to the Swift plugin — and there's no iOS/macOS build anywhere in CI (`grep -l 'xcodebuild\|pod install' .github/workflows/*` is empty). So `CaptionGeneratorPlugin.swift` currently has zero automated compile coverage, and since the package ships with no UI wiring nothing else will exercise it either.

A `macos-latest` job running `swift build` against `darwin/caption_generator/Package.swift` (or `pod lib lint`) would close the asymmetry cheaply.

mobile/pubspec.yaml
```diff
@@ -190,6 +191,9 @@ dependencies:
 
   cache_sync:
 
+  # Caption Generator - On-device speech-to-text CC generation (last resort)
+  caption_generator:
```
realmeylisdev: Question rather than an objection: the PR explicitly ships no UI wiring, but this adds `caption_generator` to the app's `dependencies:`, so `main` starts carrying a plugin (Kotlin + Swift, `Speech.framework` link, a new `NSSpeechRecognitionUsageDescription` on both Runners) that nothing calls.

Would it work to land only the `workspace:` entry on line 34 here, and add the app-level dependency + Info.plist strings in the wiring PR? Keeps the shipped app free of a dead native dependency, and the package CI workflow covers the package either way.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
```diff
@@ -0,0 +1,402 @@
...
+    String outputPath, {
+    int maxConvertedFrames = _maxConvertedFrames,
+  }) => _runPreparation(inputPath, outputPath, maxConvertedFrames);
+
+  static Future<bool> _runPreparation(
+    String inputPath,
+    String outputPath,
+    int maxConvertedFrames,
+  ) async {
+    debugConversionRanInCallerIsolate = true;
```
realmeylisdev, line 180: Nit: `debugConversionRanInCallerIsolate` is `@visibleForTesting` but it's written unconditionally on the production path. The per-isolate-statics trick that makes the test work is genuinely clever, but a `bool.hasEnvironment` guard (or asserting the isolate identity via `Isolate.current.debugName` in the test) would keep the production routine free of test scaffolding.

realmeylisdev (commented):
Reviewed the package end to end (Dart + Kotlin + Swift + CI). Ran the suite locally: 66 tests, 100% line coverage, and all four package ratchets green — the tests genuinely assert behaviour, which is nice to see.

Two new blocking-ish items below (Android `RECORD_AUDIO` contract, trailing cue running to end of audio), plus a few nits. I also independently reproduce @dcadenas' two open threads on `wav_preprocessor.dart` — details in the thread replies.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
realmeylisdev: Independently reproduce this one. Two legs worth calling out for the fix:

1. `readAsBytes()` at `wav_preprocessor.dart:151` runs before `_ParsedWav.parse` and `assertConvertible`, so the frame limit can only fire after the whole file is already resident. A 30-minute 48 kHz stereo float32 WAV — structurally valid and *inside* the 30-minute `_maxConvertedFrames` budget — is ~690 MB materialised before the guard is even consulted.
2. `assertConvertible` is only reached when `isCanonical == false` (line 158 returns first), so an already-canonical 16 kHz mono PCM16 file is never bounded at all — it is read in full purely to inspect a 44-byte header.

Bounded RIFF metadata parsing (read the header region, then decide) fixes both at once, and it also removes the full read from the passthrough path entirely.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
realmeylisdev: Confirmed — `Directory.systemTemp.createTemp` at `wav_preprocessor.dart:111` sits outside both the `try` and `Isolate.run`, so a `FileSystemException` escapes the sealed `CaptionGenerationException` surface the class doc promises.

One thing that may simplify the fix: the temp directory is created *unconditionally*, before anything knows whether a conversion is needed. So on a device with an unwritable or full temp dir, even the canonical-passthrough path — which never writes a file — fails. Creating the work directory lazily, only once `_runPreparation` has decided it must convert, maps the error into `TranscriptionFailedException` and drops the passthrough dependency on tmp in the same change.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
```diff
@@ -0,0 +1,486 @@
...
+    RandomAccessFile? raf;
+    try {
+      final Uint8List payload;
+      final _WavHeader header;
+      try {
+        raf = File(inputPath).openSync();
+        // Bounded metadata parse: reads only chunk headers and the fmt body,
+        // never the payload, so limits are enforced before allocation.
+        header = _WavHeader.readFrom(raf, raf.lengthSync());
+        if (header.isCanonical) {
```
dcadenas, line 196: Validate the canonical WAV's declared payload size and duration before returning from this branch. The new 64 MiB and 30-minute limits currently run only through `assertConvertible`, so a large or long canonical file is passed to Android and streamed in full. Move the shared bounds checks ahead of this passthrough branch, preserve zero-copy passthrough for in-bounds canonical files, and add a regression for oversized or overlong canonical input.

dcadenas (changes requested):
The canonical WAV passthrough path still skips the new size and duration limits, so oversized or overlong inputs can reach Android transcription unchecked. Validate those bounds before passthrough and add regression coverage before merging.

mobile/packages/caption_generator/lib/src/exceptions.dart
hm21: Fixed in 2700037. Both `SpeechNotAuthorizedException` and `generateCaptions` docs now name Android too: `ERROR_INSUFFICIENT_PERMISSIONS` (a missing `RECORD_AUDIO` grant) maps to this exception, so "Apple only" was wrong. Good catch.

mobile/packages/caption_generator/android/src/main/kotlin/co/openvine/caption_generator/TranscriptionSession.kt
hm21: Fixed in 2700037. The final word now ends a bounded fallback after its start (median inter-word gap, or 2s for very short transcripts), clamped to the clip, instead of stretching to `audioDurationMs` — so no lone trailing caption. Added JUnit coverage for `wordEndMs` / `medianInterWordGapMs`. Nice repro.

mobile/packages/caption_generator/android/src/main/AndroidManifest.xml
hm21: Leaving this to the consuming app on purpose. `RECORD_AUDIO` is a dangerous permission; declaring it in the plugin manifest merges it into every app that depends on the package, including flows that never touch captions. The runtime check/request belongs in the caption UI PR. The doc drift you flagged is fixed in 2700037.

mobile/packages/caption_generator/README.md
hm21: Tied to the manifest decision above: since the plugin intentionally does not own the permission, the full `RECORD_AUDIO` contract is documented with the UI wiring rather than here. The exception-level drift is corrected in 2700037.

mobile/packages/caption_generator/android/src/main/kotlin/co/openvine/caption_generator/TranscriptionSession.kt
hm21: Real gap, but a bounded watchdog spans both Kotlin and Swift with lifecycle-sensitive cancel-on-completion and time-advance tests, and Swift has no CI yet (see the iOS-CI thread). Tracking it as a follow-up rather than bolting it on here — nothing invokes the path in this PR.

mobile/packages/caption_generator/android/src/main/kotlin/co/openvine/caption_generator/TranscriptionSession.kt
hm21: Agreed it is a micro-optimization: a pending-only language still fails correctly via `ERROR_LANGUAGE_UNAVAILABLE`, just after some wasted setup. Not worth the extra `RecognitionSupport` plumbing and test surface for that case — leaving as-is.

.github/workflows/caption_generator.yaml
hm21: Valid asymmetry, but `swift build` against `Package.swift` will not compile the plugin standalone — it imports Flutter/FlutterMacOS, unavailable without a consuming app. Real coverage means a `flutter build macos/ios` of an app, a much heavier job than android-compile. Deferring to a dedicated CI PR rather than adding a job that cannot compile.

mobile/pubspec.yaml
hm21: Fair question — this is a scoping call. Keeping the app-level dependency so the package is exercised in a real app build, which also needs the Info.plist strings. Can split it so main stays free of the native dep until the wiring PR if preferred, but that is a maintainer decision.

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
hm21: Fair nit. The static write is the one deterministic way I found to prove the conversion left the caller isolate (per-isolate statics); `Isolate.current.debugName` is not reliably distinct across SDKs. Keeping the single guarded bool — the alternative loses the regression guard.

hm21:
@realmeylisdev thanks for the thorough pass. Fixed the two that were clearly right in 2700037: the trailing-caption bug (final word now gets a bounded fallback end, not the whole clip — with JUnit coverage) and the `SpeechNotAuthorizedException` "Apple only" doc drift. The rest I have pushed back on inline: `RECORD_AUDIO` in the plugin manifest is deliberately left to the consuming app (dangerous permission, merges everywhere); the iOS `swift build` job cannot compile a Flutter plugin standalone so that needs a heavier dedicated CI PR; the watchdog is a real but two-platform follow-up; and the rest are nits/scoping. (Note: your two `wav_preprocessor.dart` reproductions were already fixed in 689ffcd — you reviewed the earlier commit.)

mobile/packages/caption_generator/lib/src/wav_preprocessor.dart
hm21: Fixed in 293b808. Split the guard: `assertWithinLimits` (declared payload size + recognition-duration/frame cap) now runs on every file before the passthrough branch, so a large or long canonical clip is rejected instead of streamed to the recognizer unbounded; the sample-rate-range check moved to `assertResampleable` on the conversion path only. In-bounds canonical files still pass through zero-copy. Added a regression for an overlong canonical clip. Good catch.

hm21:
@dcadenas addressed your canonical-passthrough point in 293b808 and resolved the thread: the size/duration limits now run on every file before the passthrough branch (`assertWithinLimits`), so canonical clips are bounded too, with the sample-rate check split out to the conversion path and a regression for an overlong canonical clip. All your threads are resolved now — mind re-reviewing so the change request can clear? Thanks for the thorough passes.

NotThatKindOfDrLiz:
@hm21 I pushed a takeover commit (`c25556040`) to close out the remaining review items.

What changed:
- Fixed the Android no-`RecognitionPart` fallback so untimed recognizer text spans the full audio only when no timed word parts arrive, with Robolectric coverage.
- Fixed the double-prefixed `AudioFileNotFoundException` message and pinned it in the method-channel tests.
- Made the public Dart entrypoint web-safe with conditional exports and an unsupported-platform shim.
- Removed the unused app dependency and Runner speech-recognition purpose strings so this PR stays package-only until the UI wiring PR.
- Updated README/setup docs, stale timing comments, `#6308`, and the PR body.
- Made Android native CI run package-local Gradle tests instead of relying on app plugin registration.

Verification:
- `flutter test --coverage` from `mobile/packages/caption_generator` passed, 100.00% line coverage.
- `flutter analyze` from `mobile/packages/caption_generator` passed.
- `gradle testDebugUnitTest` from `mobile/packages/caption_generator/android` passed locally with Gradle 8.14.3 / Java 21.
- GitHub checks are all green, including Caption Generator CI, Android Compile Check, Mobile CI, and the web preview build.

Please review the takeover changes. The PR is mergeable from the checks side, but `dcadenas` still needs to clear the prior CHANGES_REQUESTED state before merge.

NotThatKindOfDrLiz (approved):
Approved after takeover. The remaining blockers I verified are addressed in c25556040: Android untimed fallback now spans the full clip with native regression coverage, exception mapping is pinned, docs/setup comments are corrected, app dependency/privacy strings are deferred to wiring, and CI now verifies the package-local Android tests. All GitHub checks are green.

dcadenas (approved):
🎬 on-device caption generation is ready to ship

PR #6309, opened 2026-07-22
fix(branding): finish the Divine icon set — web icons, brand colours, app naming

realmeylisdev:
Closes #6282

## Description

Finishes the generated Divine app-icon set and the web chrome values that ship alongside it.

### Web icons

- Adds the missing `flutter_launcher_icons.web` block so `web/favicon.png` and `web/icons/Icon-*.png` no longer ship the stock Flutter scaffold artwork.
- Keeps one visual source of truth, `assets/icon/app_icon.png`, for the installed app, PWA icons, and browser favicon.
- Extends `scripts/generate_web_favicon.dart` into the required post-generator step for web-only exceptions:
  - derives the 96x96 browser favicon from the current app-icon foreground by removing the green launcher plate while preserving the green play button.
  - rewrites `Icon-maskable-192.png` and `Icon-maskable-512.png` as the app icon inset 16% over Brand Green, matching the native adaptive-icon safe-zone pattern.

### Brand colours and naming

- Repaints the Android adaptive background, PWA theme colour, and web loading spinner to Brand Green `#27C58B`.
- Renames web-facing OpenVine/app.divine labels to Divine.
- Updates the web manifest description to match the current Divine wording.

### Regression guard

Adds `test/tools/app_icon_set_test.dart` to pin the generated icon set:

- platform icon counts stay fixed.
- ordinary launcher icons remain perceptual resamples of `assets/icon/app_icon.png`.
- maskable web icons keep the bright foreground mark inside the maskable safe zone.
- the browser favicon keeps the current app-icon foreground silhouette, transparent corners, and centered green play button.
- generator config and emitted flat colours stay on the expected Brand Green values.

## Verification

- `dart run scripts/generate_web_favicon.dart`
- `dart analyze scripts/generate_web_favicon.dart test/tools/app_icon_set_test.dart`
- `flutter test test/tools/app_icon_set_test.dart`
- `flutter analyze lib test integration_test`
- `scripts/golden.sh verify`
- Pre-commit and pre-push checks passed, including merge-conflict detection, formatting, analyzer, and changed-file tests.

## Type of Change

- [x] Bug fix
- [x] Build configuration change

PR #6310, opened 2026-07-22
test(branding): pin the loading spinner to the current Divine brand mark

realmeylisdev:
## Description

#6282's branded-loading-spinner bullet. **The sprite sheet is already the current mark — no asset swap needed.** Its rest pose (frame 21) matches `assets/icon/divine_mark.svg` at IoU **0.982**; the retired `old_app_icon.png` scores 0.335 and the cursive wordmark 0.195. The spinner is that mark in white, animated as a wing-flap cycle. This is the third #6282 bullet whose premise turned out to be already-correct, after #6306 (auth logo) and #6309 (app icon).

What it *did* lack is any guard:

- **No test referenced the widget at all.** The sheet could be swapped for a retired mark and every suite stays green — the same blindness that let #6290 ship the wrong watermark.
- **Nothing tied the hardcoded `27` to the asset.** The slicer steps by one frame width, so a sheet with a different frame count would be sliced mid-frame, silently, with no failure anywhere.

`branded_loading_indicator_test.dart` pins four things:

| Guard | Catches |
|---|---|
| Call-site asset path (literal, not the widget's own constant) | sheet redirected to another file |
| `height == width * frameCount` | frame count drifting from the asset, in either direction |
| White on transparency | the retired brand pack (cream, blue ≈ 234) or a baked-in backdrop |
| Silhouette overlap vs `divine_mark.svg` | wrong mark entirely |

The identity check rasterises `divine_mark.svg` — the same asset `DivineIconName.divineMark` renders in the feed metadata row — so a future brand refresh has to move the spinner and the mark together or go red. Threshold 0.90 sits between the 0.982 real match and the nearest wrong-mark candidate (isolated 3D mascot, 0.803). Only the rest pose matches; mid-flap frames legitimately drop to 0.647, so the assertion is on the best frame.

Also drops `assets/loading-brand.gif` — the sprite sheet replaced it in #912, it has had zero code references since, and it was still declared in `pubspec.yaml` and shipped in every build.

**Related Issue:** Relates to #6282

## Out of Scope

- The remaining #6282 bullets: §3 splash verification and §4 orphan deletion. Note §4 lists `divine_mark.svg` as a zero-ref orphan — it is not; it backs `DivineIconName.divineMark` and is now also this spinner's reference asset.
- Spinner colour. It ships pre-coloured white rather than brand green `#27C58B`. Divine is dark-mode only so white reads well, and the brand guide permits white "provided strong contrast is maintained" — but it is a design call, not a code one, so I left it alone. Happy to change it if you'd rather it were green.
- The widget has no `Semantics` label and ignores `MediaQuery.disableAnimations`. Both are pre-existing and would need an l10n key across 17 locales; kept out of a guard-only PR.

## Verification

- `flutter test test/widgets/branded_loading_indicator_test.dart` — 5/5, also green under `--test-randomize-ordering-seed random`
- Consumer suites re-run: `composable_video_grid`, `library/clips_tab`, `profile/profile_collabs_grid`, `video_detail_screen`, `sound_detail_screen` — 75 passed
- Ratchets: untested-services floor, test-unit structure, and all four design-system ceilings — green
- **Mutation-tested, 7/7 confirmed red on the intended assertion:** retired mark stacked into a correctly-shaped sheet (0.320), 20-frame sheet, cream recolour, opaque backdrop, `frameCount` 27→26, asset constant repointed, `..repeat()` removed

## Type of Change

- [x] 🧹 Code refactor


NotThatKindOfDrLiz:
Takeover update pushed in `ad7726a`.\n\nWhat changed:\n- Added `test/helpers/brand_silhouette.dart` and moved the spinner identity helpers over to it so #6316 can consume the same helper instead of duplicating it.\n- Added a pubspec asset declaration guard for `assets/loading-brand-sprite.png`, so deleting the sprite from `flutter.assets` fails the test even though the PNG still exists on disk.\n- Removed the extra private frame-count alias from `BrandedLoadingIndicator`.\n- Kept the guard deterministic under the merged full test run by parsing `pubspec.yaml` instead of calling `rootBundle.load`, which can be affected by process-global `flutter/assets` test channel state.\n\nVerification passed:\n- Local pre-push analyzer for changed files\n- Local pre-push `flutter test test/widgets/branded_loading_indicator_test.dart` — 6/6\n- GitHub checks all green on the amended push: mergeability, semantic PR, preview build, generated files, format, analyze, VGV tag gate, and full Tests.\n\n@realmeylisdev please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover. The remaining guard gap is closed by the pubspec asset declaration assertion, the duplicated silhouette logic has been moved into a shared test helper, and the amended commit is green locally and in GitHub CI.

hm21 (approved):
LGTM

PR #6312, opened 2026-07-22
fix(comments): make flag-content report sheet scrollable so Submit is reachable

### LINKED ISSUE #6311: Comment flag/report sheet: Submit unreachable when reasons overflow (no scroll)
## Summary

A user cannot submit a report when flagging a comment. The **Flag Content** bottom sheet lists all 11 report reasons plus the **Submit** button in a single non-scrolling column, so once the reasons fill the screen, Submit is pushed off the bottom with no way to reach it. The list does not scroll. Blocking works because that is a separate, shorter sheet.

**Report & verification:** Reported by a staff member who hit it while trying to report an abusive comment on their account. I verified the reproduction and root cause from their screen recording and log export.

## Environment (from the reporter's log export)

- App 1.0.17+809, iOS 26.5.2, iPhone 16 Pro (iPhone17,1)
- **Text Scale: 1.35** (large-text accessibility), high contrast on

## Steps to reproduce

1. As a large-text user (any elevated text scale), open a comment thread on another user's post.
2. Open the comment options (…) → **Flag Content**.
3. Look for the Submit button.

**Expected:** every reason and the Submit button are reachable; you can pick a reason and submit.

**Actual:** the sheet fills the screen at the 7th reason ("Child Safety Violation"); reasons 8-11 (CSAM, Underage user, AI-generated, Other) and the Submit button are clipped off the bottom. The sheet does not scroll, so Submit is unreachable and the report cannot be completed.

## Root cause

`_FlagContentSheet` in `mobile/lib/screens/comments/widgets/comment_options_modal.dart` renders its body as a non-scrolling `Column` (prompt + 11 reason tiles + Submit) inside a fixed (`scrollable: false`) `VineBottomSheet`. When the content exceeds the sheet height — 11 tiles, made worse by large text scale — it overflows and clips Submit. The sibling `VineBottomSheetSelectionMenu` already wraps its option list in a `SingleChildScrollView` for exactly this reason.

## Fix

Make the flag-content body scrollable (mirroring `VineBottomSheetSelectionMenu`) so Submit is reachable regardless of reason count or text scale. Regression test added that reproduces the overflow at 1.35 text scale on a phone-height viewport. PR to follow.

## Impact

Blocks in-app comment reporting for large-text users — an accessibility and Trust & Safety gap on a harassment-response path. Priority left for triage.

mbradley:
## Description

Reporting a comment was impossible for some users: the **Flag Content** sheet lays out all 11 report reasons plus the Submit button in a single non-scrolling column, so once the reasons fill the sheet, Submit is pushed off the bottom with no way to reach it. Large accessibility text scales make it worse (the original report came from a user at 1.35 text scale), but 11 reasons overflow most phones on their own. Blocking is unaffected because it's a separate, shorter sheet.

Submit is now **pinned** below a scrollable reason list, so it stays visible without scrolling regardless of reason count or text scale. While in the same file, each reason tile now also exposes **button / selected / label semantics** (it was a bare `GestureDetector` with no button role), so screen-reader users can report a comment too. This is a Trust & Safety and accessibility fix on a harassment-response path.

**Related Issue:** Closes #6311

## Out of Scope

Other `VineBottomSheet(scrollable: false)` sheets may share the same overflow trap; auditing them is a separate follow-up, not this PR.

## Verification

Four widget tests, each mutation-checked (fails on the pre-fix code, passes on the fix):

- **Submit reachable at 1.35 text scale** — reproduces the overflow, taps Submit *without scrolling* to prove it stays pinned, re-reads Submit's rect after scrolling to prove it doesn't move, and scrolls to the last reason asserting it sits *above* Submit (never hidden behind it). A message-stable "no `overflow`" assertion backs the SDK-coupled `debugSize` allowlist.
- **Reason-tile a11y** — asserts each tile exposes `isButton` / `isSelected` / label semantics.
- **Submit disabled until a reason is selected** — asserts `onPressed` is null on open and non-null after a reason is chosen.
- Plus the pre-existing delete-option test, unchanged.

`flutter analyze` clean; `dart format` clean; all four design-system ceiling ratchets pass (no new violations). Reviewed by two independent adversarial passes (layout correctness + test quality), both "ready"; their hardening notes were resolved.

Visual before-state is the reporter's screen recording on #6311; I can attach an after-recording from the simulator if wanted.

## Type of Change

### DIFF
--- a/mobile/test/screens/comments/comment_options_modal_test.dart
+++ b/mobile/test/screens/comments/comment_options_modal_test.dart
@@ -258,11 +258,13 @@ void main() {
       await tester.pumpAndSettle();
 
       // Each reason is a labelled, selectable button for screen readers (it
-      // was a bare GestureDetector exposing no button role).
+      // was a bare GestureDetector exposing no button role). The tap action
+      // is asserted too: excludeSemantics drops the child subtree, so without
+      // an explicit onTap the node reads as a button that cannot be activated.
       final harassment = find.text(l10n.reportReasonHarassment);
       expect(
         tester.getSemantics(harassment),
-        isSemantics(isButton: true, isSelected: false),
+        isSemantics(isButton: true, isSelected: false, hasTapAction: true),
       );
       expect(
         tester.getSemantics(harassment).label,
--- a/mobile/test/screens/comments/comment_options_modal_test.dart
+++ b/mobile/test/screens/comments/comment_options_modal_test.dart
@@ -143,8 +143,25 @@ void main() {
         // real error and hangs pumpAndSettle.
         final caughtErrors = <FlutterErrorDetails>[];
         final previousOnError = FlutterError.onError;
-        FlutterError.onError = caughtErrors.add;
         addTearDown(() => FlutterError.onError = previousOnError);
+
+        // Runs [flow] with framework errors collected instead of failing the
+        // test, and restores the real handler before returning.
+        //
+        // Nothing inside may call expect(): flutter_test reports a
+        // TestFailure through FlutterError.onError, so the collector would
+        // swallow it, `_pendingExceptionDetails` would stay null, and the
+        // binding's own assert would fire before it can complete the test —
+        // hanging the shard instead of reporting the regression. Measure
+        // inside, assert outside.
+        Future<void> collectingErrors(Future<void> Function() flow) async {
+          FlutterError.onError = caughtErrors.add;
+          try {
+            await flow();
+          } finally {
+            FlutterError.onError = previousOnError;
+          }
+        }
 
         // Flutter's TextPainter asserts `debugSize == size` when a
         // RenderParagraph is scrolled under a non-1.0 textScale in tests — a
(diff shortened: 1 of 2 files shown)


mbradley:
Keeping this in draft while I confirm it's good to go.

The red Tests check is a main-side flake right now, not this change. A couple of tests that just landed on main (the sign-in wordmark image-semantics test, plus a DM-restriction visibility test) are failing intermittently on unrelated PRs too, under the merged-isolate CI run. There's already a fix in flight for the main one (#6333), and this PR's own tests pass locally and in the full merged-isolate run. I'll refresh onto main and re-run once the flake fix lands.

rabble (commented):
Draft review pass is clean. The reason list is the only scrollable region, Submit remains outside it and pinned, selection semantics are explicit, and the widget coverage exercises large text scale, reachability, pinning, disabled state, and selectable-button semantics. Current-head checks are green. No code blocker from this pass; when author validation is complete, mark ready and request the normal review team.

hm21 (commented):
The layout fix is right — `_FixedContent` already hands `body` a bounded height, so `Flexible` + `SingleChildScrollView` is the correct shape and the sheet still content-sizes when the reasons fit. Mutation check holds.

The a11y half did not work, so I pushed two commits (shared-ownership policy — revert or tweak freely):

- **`54e20f8`** — `excludeSemantics: true` drops the whole child subtree from the semantics tree, including the `GestureDetector`'s tap action. Each reason tile came out `flags: [isButton, hasSelectedState], actions: 0`: focusable and labelled, but double-tap did nothing, so the report still couldn't be filed. Re-declared `onTap` on the `Semantics` → `actions: 1`. `isSemantics` only checks the fields you name, which is why the new test passed at zero actions; it now asserts `hasTapAction`.
- **`fc567e4`** — `_OptionTile` had the identical shape (pre-existing), so Flag Content / Block User / Edit / Delete were also `actions: 0` and a screen reader couldn't open the report sheet at all. Same one-line fix plus a mutation-checked test.

Validation: 5/5 in the file green, both new assertions verified red without their fix, `flutter analyze lib test integration_test` clean, `dart format` clean.

Remaining, non-blocking: the note below on error-handler ordering, and I only verified the semantics tree in widget tests — no TalkBack/VoiceOver pass on a real device.

mobile/test/screens/comments/comment_options_modal_test.dart
```diff
@@ -69,5 +70,315 @@ void main() {
...
+        // -> dismiss flow. The debugSize allowlist is coupled to an
+        // SDK-internal string, so also assert the real hazard — an overflow —
+        // directly and message-stably.
+        expect(
+          caughtErrors.where((e) => e.exceptionAsString().contains('overflow')),
+          isEmpty,
+          reason: 'flag sheet overflowed',
+        );
+        expect(
+          unexpectedErrors(),
```
hm21, line 297: These `expect`s run while `FlutterError.onError` is still the collector, so the `TestFailure` gets swallowed. On the pre-fix widget the test does go red, but via `_pendingExceptionDetails != null` ("A test overrode FlutterError.onError…") and then the process hangs instead of exiting. Restoring `FlutterError.onError` before asserting would make a future regression report `flag sheet overflowed` rather than stalling the shard.

hm21 (approved):

mobile/test/screens/comments/comment_options_modal_test.dart
realmeylisdev: Confirmed and fixed in e94b6e1ab.

Reproduced on the pre-fix widget: the shard ran **4m30s without terminating** and only ended on SIGTERM, printing `Failed assertion: line 1911 pos 9: '_pendingExceptionDetails != null'` instead of the real diagnostic. Path is exactly as you described — `binding.dart:1878` reports the `TestFailure` through `FlutterError.onError`, the collector absorbs it, so `_pendingExceptionDetails` stays null and the assert fires inside `handleUncaughtError` before `testCompletionHandler` is reached.

The flow now measures inside a `collectingErrors(...)` helper that restores the real handler in a `finally`, and every `expect` runs after it. Same mutation now fails in **5s** with:

```
Expected: empty
  Actual: ['A RenderFlex overflowed by 1089 pixels on the bottom.']
overflow on sheet open
```

`findsOneWidget` on Submit became a match count captured inside the flow — the sheet is popped by the time the assertions run, so the finder legitimately matches nothing there.

realmeylisdev (approved):
Layout chain checks out: `_FixedContent` already hands `body` a bounded height, and `SingleChildScrollView` shrink-wraps (`size = constraints.constrain(child.size)`), so the sheet still content-sizes when the reasons fit — no full-height regression.

Mutation-checked both halves: 1089px `RenderFlex` overflow pre-fix, and both a11y tests go red without the re-declared `onTap`. Worth noting the a11y half also fixes four rows that are screen-reader-dead on `main` today (Edit, Delete, Flag Content, Block User) — `excludeSemantics: true` was dropping their tap action.

The report path itself is untouched (NIP-56 kind 1984, `e`/`p` report types + NIP-32 labels), so no protocol surface here. Test-hang thread fixed in e94b6e1ab.

PR #6316, opened 2026-07-22
fix(splash): bound launch wordmark size

realmeylisdev:
Refs #6282

## Description

Closes the "Splash / launch (native config — verify)" bullet on #6282 without closing the broader brand-refresh tracking issue.

The launch artwork was already the current Divine wordmark in Brand Green over the app surface color. The Android bug was in the legacy `launch_background.xml` path: the layer-list centered `@drawable/launch_image` at the bitmap intrinsic size, while every density bucket shipped the same 960px image. That made the held native splash vary by device density.

This takeover changes the fix to bound the legacy layer-list item directly:

- keeps one full-resolution `drawable-nodpi/launch_image.png` source
- draws it into a fixed 240dp square with `android:width`, `android:height`, and centered item gravity
- uses bitmap `android:gravity="fill"` inside that square so the square source preserves aspect ratio
- removes the generated density-bucket PNGs and generator script
- removes the redundant `drawable-v21/launch_background.xml` now that `minSdk = 28`
- keeps the iOS storyboard metadata correction
- moves the retired wordmark calibration image out of shipped assets and into `test/fixtures`

The guard tests now assert the behavior contract instead of requiring the intrinsic-size density-bucket implementation. They still cover native splash wiring, launch-image dimensions, Brand Green, background color, Android 12+ circular mask safety, current-wordmark identity, and iOS launch-screen wiring.

## Verification

- `mise exec -- dart format lib test integration_test` — clean, 0 changed
- `flutter test test/tools/launch_image_set_test.dart` — 14 passed
- `flutter test test/tools/` — 62 passed
- pre-push hook — analyze clean, changed-file tests passed
- `flutter build apk --debug` — attempted, but blocked locally because this machine cannot locate a Java Runtime

## Type of Change

- [x] 🛠️ Bug fix
- [x] ✅ Native resource/config cleanup

## Manual test plan

The splash is held by `StartupSplashReleaseController` until startup auth settles, with a 21-second ceiling. For a screenshot-able splash, sign in, enable airplane mode, force-quit, and cold-start from the launcher icon.

Android:

1. Compare `main` vs this branch on an Android 12+ device, especially the Samsung device that reported the regression.
2. Pass: during both the system splash and the held `windowBackground` phase, the wordmark remains centered, crisp, unclipped, and the same visual size as the intended 240dp splash treatment.
3. On Android 9-11 or an API 30 emulator, force densities with `adb shell wm density` and verify the held legacy splash remains the same dp size across density settings.

iOS:

1. Delete the app first to clear launch-screen cache, then cold-start on at least one @2x or @3x device/simulator.
2. Pass: the launch image remains centered, crisp, unstretched, and on the same near-black app surface color.

Handoff note: this takeover intentionally avoids the prior downsampled density-bucket approach because it made the long-held Android launch background visibly lower resolution on common modern devices.


hm21 (commented):
I tested it, and on my Samsung Galaxy the logo now looks like it's low resolution. Before, it was really sharp, almost like it was in 4K but now it looks more like 480p

realmeylisdev:
Thanks — confirmed. The first fix pre-downsampled the 960×960 source into density buckets (for example, xhdpi became 480×480), so the Samsung system splash could start from a lower-resolution bitmap. Commit dd14072 replaces those copies with the original 960×960 density-neutral source and bounds only the legacy layer-list to 240dp, preserving the consistent size without throwing away source resolution. The regression test now pins that contract. Local analysis, all 62 tools tests, a clean Android resource link, and goldens pass. Please retest on the same Galaxy when convenient.

NotThatKindOfDrLiz:
@realmeylisdev I took over the review feedback and pushed `580def526`.

What changed:
- replaced the downsampled density-bucket approach with one full-resolution `drawable-nodpi/launch_image.png`
- bounded the legacy `launch_background.xml` item to 240dp so the held Android window background is density-stable without losing resolution
- removed the generated Android bucket PNGs, generator script, and redundant `drawable-v21` launch background
- kept the iOS storyboard metadata correction
- moved the retired wordmark calibration image out of shipped assets into `test/fixtures`
- rewrote the guard tests to assert the launch behavior contract instead of pinning the old intrinsic-size mechanism

Verification passed:
- `flutter test test/tools/launch_image_set_test.dart`
- `flutter test test/tools/`
- pre-push analyze + changed-file tests
- GitHub checks are green, including Mobile CI

`flutter build apk --debug` was attempted locally but this machine has no Java runtime, so native resource-link verification is covered by CI here. Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Final pass is clean after the takeover changes. The Android splash now keeps full-resolution artwork and fixes the density-size bug at the layer-list bounds, tests assert the behavior contract, and CI is green.

Issue #6315, opened 2026-07-22
feat(clip-reuse): add configurable attribution for reused clips


hm21:
### What would you like?

Enforce crediting the original creator whenever a clip is reused. For future support of optional clip reuse (not limited to original clips), add an "Allow others to reuse my clips" setting where creators can choose how they want to be credited. Attribution options could include automatically listing them as the original creator and/or displaying their username as a small watermark on reused clips.

### How would this be useful for you?

This ensures creators are always credited for their work while giving them control over how attribution is presented. It also provides a scalable foundation for expanding clip reuse in the future.

### When would you use this?

When reusing another creator's clip, or when enabling others to reuse your own clips and selecting your preferred attribution method.

### Anything else?

Discussed on discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1529508040951337061)

NotThatKindOfDrLiz:
Started the design-independent foundation as Phase 0:

- #6320 tracks imported clip provenance capture.
- #6321 implements it by storing source author/event/addressable/relay metadata in the existing clip JSON payload, with no Drift migration.

I intentionally did not implement the rest of this issue in that PR because the broader request still needs product/design decisions:

- how to credit timelines with clips from multiple source creators
- whether own reused clips should produce visible attribution
- whether clip attribution should be a notifiable mention
- where the creator-facing "Allow others to reuse my clips" setting belongs
- whether source watermarking is actually part of v1, since editor-render watermarking is separate from the existing download-time watermark path

This should also stay aligned with #6227 so audio and clip attribution do not diverge into two separate models.

NotThatKindOfDrLiz:
Update from implementation pass:

- #6725 already shipped the single-source inspired-by credit path, so this issue is stale as originally written.
- I split out the design-independent live bug slice into PR #6928: preserve multiple clip-source credits, keep author-only provenance load-bearing, stop merge from destroying disagreeing source provenance, publish factual clip-source tags separately from manual inspired-by attribution, and add regression coverage.
- The creator consent setting, `allow_clip_reuse` import/publish policy, and editor-render watermarking remain out of scope for #6928 because they still need product/design answers from the open questions.

PR: https://github.com/divinevideo/divine-mobile/pull/6928

NotThatKindOfDrLiz:
Closing this original mixed tracker as superseded.

Current state:

- #6725 already shipped the single-source inspired-by credit path.
[...]

PR #6321, opened 2026-07-22
feat(clip-reuse): preserve imported clip provenance

### LINKED ISSUE #6320: feat(clip-reuse): preserve imported clip provenance
## Goal

Capture source provenance when a published video is imported into the local clips library so future clip attribution UI and publish tags have source data to work from.

## User impact

Classic Vine and own-video imports currently create reusable clips without preserving the source author or original video reference. That blocks any later crediting experience for reused clips.

## Acceptance criteria

- Imported clips preserve the source author pubkey and source event id.
- Imported clips preserve the source kind 34236 addressable coordinate when one is available.
- Imported clips preserve a source relay hint when one is available.
- Existing clip JSON without provenance fields continues to load.
- No Drift schema migration is introduced for this non-queryable metadata.

## Technical notes

This is Phase 0 of #6315. The broader attribution UI, creator reuse permission, and watermark behavior remain product/design follow-ups.

NotThatKindOfDrLiz:
## Summary

- add nullable source provenance fields to DivineVideoClip and persist them in the existing clip JSON payload
- populate imported clips with source author, event id, addressable id, and relay hint from VideoEvent
- cover populated and legacy JSON round trips plus classic Vine and own-video import provenance

## Verification

- flutter test test/models/divine_video_clip_test.dart test/services/video_clip_import_service_test.dart
- flutter analyze

Closes #6320
Refs #6315

### DIFF
--- a/mobile/lib/models/divine_video_clip.dart
+++ b/mobile/lib/models/divine_video_clip.dart
@@ -40,6 +40,10 @@ class DivineVideoClip {
     this.proofManifestJson,
     this.deletedAt,
     this.transition,
+    this.sourceAuthorPubkey,
+    this.sourceEventId,
+    this.sourceAddressableId,
+    this.sourceRelayHint,
   }) : assert(
          video != null || stopMotionFrames != null,
          'A clip must have either a video file or stop-motion frames',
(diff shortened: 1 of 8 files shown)


hm21 (approved):
Took this over and pushed the follow-up directly (a47a5eec2).

**Finding:** the new `source*` fields were only set on the raw import — `flattenClipForLibrary` and `mergeClips` rebuild the clip field-by-field, so provenance was dropped the moment an imported clip was edited or merged back into the library.

**Fix:**
- `flattenClipForLibrary` carries the source attribution over unchanged (same logical clip).
- `mergeClips` keeps it only when every input shares the same source; a mixed-source merge has no single author to credit, so it stays null.
- Tests added for both paths.

**Validation:** `dart format` clean, `flutter analyze` clean, changed-file suites green (55 tests incl. the new provenance cases).

The published-export render (`video_editor_render_service`) still doesn't attach provenance — that's attribution on the final video and belongs to the #6315 consumer, not this storage PR.

Solid, well-tested change — approving.

PR #6329, opened 2026-07-23
feat(profile): link creator sites from profiles

### LINKED ISSUE #6328: feat: Add a discovery path for divine.space in the mobile app
## What would you like?
A dedicated, discoverable entry point to **divine.space** from within the Divine mobile app — surfacing it beyond the flag-gated Explore → Apps sub-tab it lives in today.

divine.space is Divine's MySpace-style Nostr profile client (Top 8 friends, profile music, custom themes, bento-grid page builder). It already ships in the app as a bundled Nostr app-directory entry (`bundled-divine-space`, `preloaded_nostr_apps.dart`), openable via the in-app WebView sandbox — but the only route to it is **Explore → Apps**, which only appears when the Integrated Apps feature flag *and* platform support are on. Most users never see it.

Gives users a direct way to explore and customize their spatial/social profile via divine.space, driving cross-property engagement. Today the surface exists but is effectively hidden.

## When would you use this?
A user browsing their profile or the Explore feed wants to set up their divine.space page (themes, Top 8, music) and can tap through directly instead of digging into a flag-gated Apps sub-tab.

## Anything else?
**Current state**
- Bundled entry: `mobile/packages/nostr_app_bridge_repository/lib/src/preloaded_nostr_apps.dart:182` (`launchUrl: https://divine.space/`).
- Opens via WebView sandbox: `mobile/lib/screens/apps/nostr_app_sandbox_screen.dart`.
- Only reachable through `ExploreScreen` → Apps tab (`mobile/lib/screens/explore_screen.dart`), gated by the Integrated Apps flag + `nostr_apps_platform_support.dart`.

**Open design questions (for the assignee to decide/spec)**
- Where should the entry point live? Candidates: profile screen action, an Explore promo card/row, bottom-nav affordance, or an onboarding nudge.
- Should it open in the in-app WebView sandbox (with the Nostr signer bridge) or launch the system browser via `url_launcher`?
- Should it be gated by the Integrated Apps feature flag, or ship on its own flag / ungated?
- Should the entry deep-link to the signed-in user's divine.space profile rather than the site root?

## Acceptance Criteria
- [ ] A discoverable entry point to divine.space exists outside the Explore → Apps sub-tab
- [ ] Entry point respects auth state and (if applicable) deep-links to the user's profile
- [ ] Behavior is covered by tests (widget + routing)
- [ ] Feature-flag / gating decision documented in the PR

realmeylisdev:
## Description

Adds a discovery entry point from a profile to the creator's public **Divine Space** page (`https://divine.space/{npub}`), opened in the system browser.

- **Divine Space pill.** A compact secondary pill sits in the identity block alongside the profile's other links — **`my divine.space`** on your own profile, **`divine.space`** on someone else's. The label names the destination outright rather than saying "Website", which would collide with the Kind-0 website row directly above it. It carries the Divine mark instead of a globe for the same reason. Because the visible copy is self-describing it *is* the accessible name, so the two can't drift apart.
- **Tip/Support pill beside it.** The monetization affordance is a matching pill on the same centered row — **Tip** on iOS storefronts, **Support** elsewhere. It is deliberately **label-only**: a bolt glyph reads as a Lightning Network payment, which this is not.
- Both are `tiny` secondary `DivineButton`s, so they stay visually matched if the design system changes. With only one present, the row collapses to that child and stays centered.
- Routes through the shared `openExternalLink` launcher, with `divine.space` added to the trusted-domain policy, so outbound launch and trust stay centralized.
- Records a `creator_site_cta_tapped` analytics event (with `is_own_profile`), mirroring the sibling support affordance, so cross-property discovery engagement is measurable.
- Existing Kind-0 websites and Tip/Support links are preserved; only a Kind-0 website that exactly matches the generated Divine Space URL is deduplicated.

### Cold-load correctness

The Divine Space pill is derived from the npub and is available instantly, but the Tip pill needs the `divine_monetization_links` Kind 0 field — which Funnelcake's REST projection strips. Two fixes keep that from showing:

- **Reveal together.** The row holds until the monetization answer is authoritative (flag off, or a non-`rest-` Kind 0 has landed) so both pills appear in the same frame instead of the site pill rendering first and Tip popping in a beat later. A 10s timeout reveals the site pill alone, so a dead relay can't strand it.
- **Self-heal REST-only profiles.** When a monetization-aware `OtherProfileBloc` ends up showing a REST projection with no links, it re-fetches the raw Kind 0 with bounded exponential backoff until the relay recovers. A relay Kind 0 with genuinely no links is left alone, so real "no tips" creators aren't retried.

Copy is localized across all 18 locale catalogs. `divine.space` is a domain and stays untranslated; only the possessive determiner is localized, and the ARB descriptions say so.

This is a UI-only public entry point: no backend lookup, authentication, or new Nostr event flow. The pill is gated only by valid-pubkey; Tip/Support by the existing `profileMonetizationLinks` flag.

**Related issue:** Refs #6328 (does **not** close it — see below).

## Why Refs, not Closes #6328

#6328 also asks for an **authenticated, in-app creator-site customization** flow. That is intentionally **out of scope** here and tracked in **#6337**. This PR ships the **read-only discovery v1**; it does not implement the auth-state / customization acceptance criterion, so it references rather than closes the issue.

### Browser vs. in-app signer (decision)

The in-app `NostrAppSandboxScreen` cannot sign profile/customization edits for Divine Space **as configured today**:

- The bundled `divine-space` entry uses the shared `_sharedSignEventKinds` (`[1,6,7,14,15,1111,9734,30023]`) — **no kind 0 / custom kinds**, so it can't sign profile edits.
- It has **no** auto-login script (manual WebView login) and is **origin-locked** to `https://divine.space`.
- The sandbox is native-only, so it needs a web fallback.

Wiring a signer-capable customization route is therefore a larger, security-touching change, split to **#6337**. The browser entry point is the shippable, low-risk discovery step in the meantime.

## Out of Scope

- Authenticated in-app Divine Space customization — **#6337**.
- **Native-speaker translation pass.** The 17 non-English renderings of `my divine.space` were authored alongside the code, not by a translator. The possessive attached to a Latin-script domain is the risky part; Turkish (`divine.space'ım`) and Romanian (`divine.space-ul meu`) most warrant a check.

## Scope note

- **Tip/Support pill + support-sheet touch-ups** are the *same monetization affordance surface* this PR already changes (the Tip pill opens that sheet), kept together for a consistent result.
- **`personal_event_cache_service` test waits** are incidental test-stability fixes for the shared `very_good --optimization` isolate (fire-and-forget Hive writes), included to keep this PR's suite green. Happy to split them into a standalone PR if preferred.

## Real-device TCs

- **TC1 — Other profile:** Open another creator's profile; verify a `divine.space` pill sits with the profile links, below the bio and above the stats. Tap it and confirm the system browser opens `https://divine.space/{same full npub}`.
- **TC2 — Own profile:** Verify the pill reads `my divine.space` and opens your own full-npub page externally.
- **TC3 — Public access:** Signed out/anonymous, confirm the pill remains visible and the page loads without Divine authentication.
- **TC4 — Coexistence/dedup:** A different Kind-0 website and Tip/Support remain visible; if a Kind-0 website exactly matches the generated Divine Space URL, only the generic website row is hidden and the pill remains.
- **TC5 — Tip/Support:** With `profileMonetizationLinks` enabled and at least one link, verify the label-only **Tip** (iOS) / **Support** pill renders beside the site pill and opens the support sheet. On iOS, subscription-only links are filtered out.
- **TC6 — Cold load:** On a fresh profile open, verify both pills appear in the same frame — the site pill must not render seconds ahead of Tip.
- **TC7 — Flapping relay:** With the relay failing the raw Kind 0 fetch, verify the Tip pill still recovers without an explicit refresh, and that the site pill reveals on its own after the 10s fallback.
- **TC8 — Layout:** On a narrow device or at large text sizes, verify neither pill clips and the pair stays centered.

## Verification

- `flutter analyze lib test integration_test` clean
- `dart format` clean
- Targeted suites green: `test/widgets/profile/`, `test/screens/other_profile_screen_test.dart`, `test/blocs/other_profile/`, `test/l10n/`, `test/utils/external_link_launcher_test.dart`
- Verified on the iOS simulator (iPhone 17) on a real profile: both pills render, `Tip` copy is used on the iOS storefront, and the pill opens `divine.space/{npub}` in the system browser
- Rebased on latest `origin/main`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mobile/lib/widgets/profile/profile_creator_site_button.dart
```diff
@@ -0,0 +1,127 @@
...
+          ? context.l10n.profileCreatorSiteOwnLabel
+          : context.l10n.profileCreatorSiteVisitLabel,
+      onPressed: () => _open(context, uri),
+    );
+  }
+
+  Future<void> _open(BuildContext context, Uri uri) async {
+    var launched = false;
+    try {
+      launched = await launcher(uri, LaunchMode.externalApplication);
```
NotThatKindOfDrLiz: This is the decision that needs explicit sign-off before this PR can close #6328. The issue's motivating flow is customizing the user's Divine Space page, and the existing bundled `divine.space` entry opens in `NostrAppSandboxScreen` so the Nostr signer bridge is available. Launching `externalApplication` makes this a public/read-only browser hop for most users. Please either route this CTA through the signer-capable in-app sandbox, or update the PR/issue with product confirmation that the profile CTA is intentionally read-only external launch.

mobile/lib/l10n/app_de.arb
```diff
@@ -4053,6 +4053,9 @@
     },
     "profileWebsiteSemanticLabel": "Visit website: {url}",
     "profileCouldNotOpenWebsite": "Could not open website",
+    "profileCreatorSiteVisitLabel": "Visit creator site",
```
NotThatKindOfDrLiz: These new profile strings are shipping as English in the non-English catalogs. If real translations are intentionally deferred, please track them as explicit untranslated debt (matching the repo's l10n debt pattern) or add translations now, instead of silently copying the English source into every locale.

mobile/test/widgets/profile/profile_creator_site_button_test.dart
```diff
@@ -0,0 +1,167 @@
...
+    });
+  });
+
+  group(ProfileCreatorSiteButton, () {
+    testWidgets('uses prominent design-system styling and visitor copy', (
+      tester,
+    ) async {
+      await tester.pumpWidget(_wrap(isOwnProfile: false));
+
+      expect(find.text('Visit creator site'), findsOneWidget);
```
NotThatKindOfDrLiz: Please resolve localized copy through `lookupAppLocalizations(const Locale('en'))` instead of hardcoding English in widget assertions. `.claude/rules/localization.md` calls this out so copy changes and widgets that stop reading l10n fail clearly. The same cleanup applies to the other new hardcoded localized assertions in this test and the profile header/support sheet tests.

NotThatKindOfDrLiz (changes requested):
Requesting changes until the product decision is explicit: this PR closes #6328, but it currently implements the profile CTA as an external-browser, public/read-only entry point. #6328 calls out Divine Space customization and the existing bundled entry uses the in-app Nostr app sandbox, so we need either a code change to use the signer-capable route or a reviewer/product confirmation in the PR that read-only external launch is the intended scope. I also left two small cleanup comments for l10n/test debt.

mobile/lib/widgets/profile/profile_creator_site_button.dart
realmeylisdev: Went with keeping this a read-only external launch for v1, and I'd like your sign-off on that scope.

The CTA is meant as a *discovery* entry point — the primary goal in #6328 (a discoverable path to divine.space outside the flag-gated Explore → Apps tab) — and the issue explicitly left sandbox-vs-browser and deep-linking as open design questions. Read-only satisfies that acceptance criterion.

The signer-capable in-app route (for own-profile customization) is a bigger, security-touching change: deep-linking into the sandbox, a web fallback, and expanding divine.space's `allowedSignEventKinds` beyond the shared set — which today is `[1,6,7,14,15,1111,9734,30023]`, so it has no kind 0 or custom kinds and wouldn't sign customization edits as-configured anyway. I've split that into #6337 so it gets its own review.

If you'd rather gate v1 on the in-app route, I'll fold #6337 in here instead — your call.

mobile/lib/l10n/app_de.arb
realmeylisdev: Translated all three `profileCreatorSite*` keys into every non-English locale and regenerated gen-l10n in 27abe383d — no more English fallbacks, and `arb_consistency_test` is green.

mobile/test/widgets/profile/profile_creator_site_button_test.dart
realmeylisdev: Done in d0f601cf2 — the new creator-site and support assertions here, in the profile header test, and in the support-sheet test now resolve copy through `lookupAppLocalizations(const Locale('en'))` instead of hardcoded English.

realmeylisdev:
@NotThatKindOfDrLiz addressed your review:

- **l10n debt** — translated the three `profileCreatorSite*` keys across all 16 non-English locales and regenerated gen-l10n (27abe383d); `arb_consistency_test` green.
- **Test l10n** — creator-site / support assertions now resolve via `lookupAppLocalizations(const Locale('en'))` in the button, header, and support-sheet tests (d0f601cf2).
- **External-launch vs signer sandbox** — this is the one that needs your sign-off. Keeping v1 as a read-only *discovery* entry point (external launch), with the signer-capable own-profile customization route split into #6337 (rationale in the thread). If you'd rather gate v1 on the in-app route, I'll fold #6337 in here instead.

mobile/lib/widgets/profile/profile_header_widget.dart
```diff
@@ -486,6 +489,86 @@ class _ProfileHeaderWidgetState extends ConsumerState<ProfileHeaderWidget> {
...
+        : context.l10n.profileSupportButtonLabel;
+
+    // Icon-only DivineButton rather than DivineIconButton: the latter paints
+    // its 2px border outside the nominal size (44px at `small`), so it would
+    // stand 4px taller than the creator-site button beside it. This is the
+    // same pairing ProfileActionButtons uses for its Message / Following
+    // icons.
+    return DivineButton(
+      key: const Key('profile-support-button'),
+      label: '',
```
NotThatKindOfDrLiz: This silently removes the visible Support/Tip label from the normal production path. For valid 64-hex pubkeys, `divineSpaceProfileUri(userIdHex)` is non-null, so this icon-only row branch always wins and the labelled standalone button is only a malformed-key fallback. Please restore visible Support/Tip copy, or get explicit design/product sign-off and call out the monetization UI change in the PR body instead of only preserving the semantic label.

mobile/lib/widgets/profile/profile_creator_site_button.dart
```diff
@@ -0,0 +1,127 @@
...
+    return DivineButton(
+      key: const Key('profile-creator-site-button'),
+      type: DivineButtonType.secondary,
+      size: DivineButtonSize.small,
+      leadingIcon: DivineIconName.globe,
+      expanded: true,
+      label: isOwnProfile
+          ? context.l10n.profileCreatorSiteOwnLabel
+          : context.l10n.profileCreatorSiteVisitLabel,
+      onPressed: () => _open(context, uri),
```
NotThatKindOfDrLiz: Please track this CTA tap before launching. This is a discovery feature whose success depends on cross-property engagement, and the sibling support affordance in the same profile header already records `trackMonetizationAffordanceTapped` via `analyticsEventSinkProvider`. Without an event here we cannot tell whether the new entry point works.

mobile/lib/widgets/profile/profile_creator_site_button.dart
NotThatKindOfDrLiz: This bypasses `openExternalLink`, which centralizes outbound-link confirmation and Divine trusted-domain policy. If Divine Space is intentionally first-party/trusted, please add `divine.space` to that policy and route this through the shared launcher or document why this CTA is exempt. Right now `divine.space` is treated as trusted here but untrusted anywhere using the shared launcher.

mobile/lib/widgets/profile/profile_support_sheet.dart
```diff
@@ -97,40 +96,30 @@ class _SupportLinkTile extends StatelessWidget {
...
-      borderRadius: BorderRadius.circular(8),
-      child: Container(
-        padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 12),
-        decoration: BoxDecoration(
-          color: VineTheme.cardBackground,
-          borderRadius: BorderRadius.circular(8),
-          border: Border.all(color: VineTheme.outlineMuted),
+    return Material(
+      type: MaterialType.transparency,
+      child: ListTile(
```
NotThatKindOfDrLiz, line 101: This support-sheet restyle is unrelated to linking creator sites from profiles. It may be a good change, but it changes a monetization surface in a PR reviewers will read as a profile-link feature. Please split this into its own PR, or explain why it is required for #6328. The same scope concern applies to the personal_event_cache_service test fixes.

NotThatKindOfDrLiz (changes requested):
Requesting changes. I agree the sandbox thread no longer blocks on technical grounds as currently configured, because bundled divine.space does not have kind 0/custom signing permissions today. But the PR still should not close #6328 as-is, and it introduces a monetization affordance regression plus a couple of small policy/measurement gaps.

Please change `Closes #6328` to `Refs #6328` or otherwise narrow the issue before merge. The implemented browser CTA is a valid read-only discovery v1, but #6328's auth-state/customization acceptance criterion is not done here. Also document the browser-vs-sandbox decision in the PR body so it is explicit.

Main blockers: restore the visible Support/Tip label or get explicit design sign-off for icon-only, add CTA tap analytics, and route/trust the Divine Space outbound launch consistently. The support-sheet restyle and cache-test fixes should be split out or clearly justified as in-scope.

mobile/lib/widgets/profile/profile_header_widget.dart
realmeylisdev: Done — this is now a labelled pill instead of an icon-only heart: ⚡ **Tip** on iOS storefronts, ✨ **Support** elsewhere. The visible copy carries its accessible name (dropped the redundant `semanticLabel`), so the normal production path no longer relies on a bare heart. `cbe69e7b4`

mobile/lib/widgets/profile/profile_creator_site_button.dart
realmeylisdev: Added — the tap now fires `creator_site_cta_tapped` (with `is_own_profile`) via `analyticsEventSinkProvider` before launching, mirroring `trackMonetizationAffordanceTapped` on the sibling support affordance. `595341e36`

mobile/lib/widgets/profile/profile_creator_site_button.dart
realmeylisdev: Fixed — `divine.space` is now in `_trustedDomains`, and the CTA routes through `openExternalLink` instead of a direct `launchUrl`. It opens directly for the trusted host (no confirmation), so trust is consistent between here and the shared launcher. `595341e36`

mobile/lib/widgets/profile/profile_support_sheet.dart
realmeylisdev: Documented under **Scope note** in the PR body: the support-sheet touch-ups are the same monetization surface the Tip button opens, kept together for a consistent result; the `personal_event_cache_service` test waits are incidental merged-isolate (`very_good --optimization`) stability fixes for fire-and-forget Hive writes, included to keep this PR's suite green. Happy to split the cache-test waits into a standalone PR if you'd prefer. `595341e36`

realmeylisdev:
@NotThatKindOfDrLiz — thanks for the thorough pass. All addressed:

- **Support/Tip label** restored as a labelled pill (⚡ Tip on iOS / ✨ Support elsewhere), no longer an icon-only heart. `cbe69e7b4`
- **CTA analytics**: `creator_site_cta_tapped` (`is_own_profile`) now fires before launch, via `analyticsEventSinkProvider`. `595341e36`
- **Shared launcher + trust**: CTA routes through `openExternalLink`, and `divine.space` is now in the trusted-domain policy — consistent everywhere. `595341e36`
- **`Closes` → `Refs #6328`** and documented the browser-vs-in-app-signer decision (signer-capable customization → #6337) in the PR body.
- **Scope**: justified the support-sheet + `personal_event_cache_service` test changes under **Scope note**; glad to split the cache-test waits into their own PR if you prefer.

Also rebased onto latest `origin/main` — the earlier CI red was a seed-dependent merged-isolate victim (`other_profile_screen_test`, untouched by this PR); the rebase brings in main's test-hygiene fixes. Watching the fresh run.

NotThatKindOfDrLiz (changes requested):
Thanks. I’m okay with the read-only external launch as the v1 discovery scope, with signer-capable own-profile customization tracked separately in #6337. Please keep this as Refs #6328 rather than closing it.

I’m not approving yet because CI is still red on the current head. Please root-cause the `auth_hero_section_test.dart` failure under seed `2483014697` and get Mobile CI green before re-requesting review.

I’m okay keeping the support-sheet polish and the `personal_event_cache_service` test-stability fixes in this PR if they’re needed to keep this surface and test suite coherent.

NotThatKindOfDrLiz:
@realmeylisdev I took over the remaining review blockers and pushed commit 4853e624bb (fix(profile): stabilize creator site pill).

What changed:
- Own-profile creator-site pill now reveals immediately instead of waiting for the 10s monetization fallback.
- Once the creator/support row has revealed, later REST-only refreshes no longer hide it and reflow the header.
- Raw Kind 0 retry/backoff moved into ProfileRepository with a session-scoped raw-missing cache, removing the BLoC-level delayed self-heal.
- OtherProfileScreen BlocProvider key now covers the full captured dependency tuple.
- Rest projection detection is centralized on UserProfile, spacing/comment wording cleanups are folded in, and package coverage now covers the retry path.

Verification:
- flutter analyze lib test (mobile/packages/profile_repository)
- flutter test test/src/profile_repository_test.dart --coverage (mobile/packages/profile_repository)
- very_good test --coverage --min-coverage 100 (mobile/packages/profile_repository)
- GitHub checks green on 4853e624bb

Please review the takeover commit and merge when ready.

NotThatKindOfDrLiz (approved):
Approved after takeover commit 4853e624bb. The remaining blockers from my review are addressed, the package coverage gap is covered, and all PR checks are green on the current head.

PR #6330, opened 2026-07-23
fix(audio): gate original-sound reuse on the creator's sharing preference

### LINKED ISSUE #6332: fix(audio): gate original-sound reuse on the creator's sharing preference
Auto-created from PR #6330.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6330
**Branch:** `fix/audio-reuse`

hm21:
Closes #6332

## Description

Creators reported their audio was still reusable by other users even with audio sharing turned off. The preference was enforced only at publish time; every path that lets a viewer *adopt* a video's original sound ignored it. This PR enforces the creator's reuse consent across all of those paths, and hardens the saved-sounds store that feeds them.

### Enforcement — a viewer can only adopt a video's original sound when the creator consented

- **Metadata "Original sound" row** ([metadata_sounds_section.dart](mobile/lib/widgets/video_feed_item/metadata/metadata_sounds_section.dart)): the row stays visible for attribution, but is only a tappable reuse entry point when the creator enabled `allow_audio_reuse` or the viewer is the creator. When the sound was reused from a source we can't resolve, it fails closed (display-only) — an owner-saved private sound can't leak through the reuse-chain fallback.
- **"Use Sound" action** ([sound_detail_screen.dart](mobile/lib/screens/sound_detail_screen.dart)): hidden for a video's original sound unless the creator consented or the viewer is the creator — closing the `/sound/<video-id>` deep-link and resolved-reuse paths that never passed a `sourceVideo`.
- **Consent signal** ([audio_event.dart](mobile/packages/models/lib/src/audio_event.dart)): `AudioEvent.allowsReuse` carries the source video's `allow_audio_reuse` marker so the gate is correct regardless of how the screen was reached (metadata sheet, reuse chain, or deep link).
- Reuse gates watch `currentAuthStateProvider` so the owner exception re-evaluates on auth restore / logout / account switch.

### Saved sounds — scoped per account and migrated safely

- **Per-account buckets** ([saved_sounds_service.dart](mobile/lib/services/saved_sounds_service.dart)): saved sounds are keyed per signed-in pubkey (`saved_reusable_sounds_<pubkey>`, `_anon` when signed out) so one account's adopted sound can't surface in another on a shared device. The provider watches `currentAuthStateProvider` so the list rebuilds on account switch.
- **Race-safe writes** ([saved_sounds_provider.dart](mobile/lib/providers/saved_sounds_provider.dart)): `saveSound` / `removeSound` / backfill freeze the account bucket before the first `await` and only update state while the account is still current, so a mid-save switch can't write A's sounds into B's bucket.
- **One-time upgrade migration**: the first signed-in account to load adopts the legacy device-wide list, then retires the key so a second account can't also inherit it. Legacy `video_*` original sounds are **dropped** (fail closed) — their creator's consent can't be validated offline, and the pre-fix device-wide bucket may hold sounds whose creator had disabled reuse.

## Out of Scope

- **Server/relay-side enforcement.** This is client-side. A modified client could ignore the gate; full enforcement would require relay-side checks outside this repo. Client paths (metadata sheet, Use Sound, deep link, saved-sounds reuse) are all covered.
- **Stale consent after a creator later disables reuse** on an already-adopted sound — a separate re-validation concern, unchanged here.

## Verification

- `flutter analyze lib test` clean; targeted suites green: `sound_detail_screen`, `saved_sounds_service`, `saved_sounds_provider`, `metadata_sounds_section`, `library/sounds_tab`, `audio_selection_bottom_sheet`, and `AudioEvent` (models). New tests cover: reuse gating + owner exception, deep-link/reuse-chain paths, the account-switch race, per-account isolation, legacy migration + `video_*` drop, and the fail-closed unresolved-reuse fallback.

## Type of Change

**Related Issue:**

### DIFF
--- a/mobile/lib/providers/saved_sounds_provider.dart
+++ b/mobile/lib/providers/saved_sounds_provider.dart
@@ -3,13 +3,21 @@
 
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:models/models.dart' show AudioEvent;
+import 'package:openvine/providers/auth_providers.dart';
 import 'package:openvine/providers/shared_preferences_provider.dart';
 import 'package:openvine/services/saved_sounds_service.dart';
 import 'package:pro_video_editor/pro_video_editor.dart';
 import 'package:unified_logger/unified_logger.dart';
 
 final savedSoundsServiceProvider = Provider<SavedSoundsService>((ref) {
-  return SavedSoundsService(ref.watch(sharedPreferencesProvider));
+  // Rebuild on sign-in/out and account switch so the service (and the sounds
+  // list built from it) always targets the current account's bucket.
+  ref.watch(currentAuthStateProvider);
+  final pubkeyHex = ref.watch(authServiceProvider).currentPublicKeyHex;
+  return SavedSoundsService(
+    ref.watch(sharedPreferencesProvider),
+    pubkeyHex: pubkeyHex,
+  );
 });
 
 final savedSoundsProvider =
--- a/mobile/lib/providers/saved_sounds_provider.dart
+++ b/mobile/lib/providers/saved_sounds_provider.dart
@@ -26,36 +26,52 @@ final savedSoundsProvider =
     );
 
 class SavedSoundsNotifier extends Notifier<List<AudioEvent>> {
+  SavedSoundsService? _backfillService;
   bool _backfillScheduled = false;
 
   @override
   List<AudioEvent> build() {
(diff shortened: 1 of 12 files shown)


NotThatKindOfDrLiz (approved):
Approved from re-review. I did not find any blocking issues. The consent gates cover the metadata, detail/deep-link, and saved-sounds paths, and CI is green on the current head.

Issue #4623, opened 2026-05-22
epic(account): harden account switching and session recovery for local drafts, clips, and publish flows


NotThatKindOfDrLiz:
## Summary

An iOS user was logged out while publishing. Session recovery then restored a different saved account, and identity-change cleanup made the original account's owner-scoped drafts and clips appear to vanish while risking destructive deletion.

This epic tracks the account-switching and session-recovery work needed to preserve local creator data, prevent silent cross-account recovery, and keep an interrupted publish recoverable.

## User impact

- Auth recovery can interrupt an in-progress publish.
- Restoring the wrong saved account can make local work appear lost.
- Destructive identity-change cleanup can permanently remove owner-scoped creator data.

## Completed engineering work

- [x] #4624 — require confirmation before recovery lands in a different saved account
- [x] #4625 — preserve owner-scoped drafts, clips, and uploads across non-destructive account changes
- [x] #4626 — keep the publish flow alive during expired-session recovery on iOS
- [x] #5788 — scope local creator-data maintenance by owner
- [x] #3222 — prevent profile-tab state from bleeding across identities
- [x] #3733 — address missing saved drafts
- [x] #4705 — triage videos missing after reinstall
- [x] #4800 — triage reported video deletion
- [x] #7462 — retarget followers and following routes after account switches
- [x] #8747 — persist known-empty following state after session setup

## Remaining validation

- [ ] #4577 — retest interrupted publishing on a build known by build number and source commit to contain the draft status-recovery fix

The original engineering acceptance criteria are covered on `main`. #4577 remains open for runtime validation because the original report was second-hand and did not include a reproducible build number or logs.

## Related follow-up

- #9129 tracks preservation of account-scoped follow and relay caches during ordinary switching. It is cache correctness work and does not block closure of this local creator-data incident epic.

## Acceptance criteria

- [x] Expired-session recovery does not unnecessarily eject the user from an active publish flow.
- [x] Re-authentication does not silently restore a different saved account without confirmation.
- [x] Owner-scoped drafts, clips, and uploads survive non-destructive account switching and become visible again when returning to that account.
- [x] Regression tests cover the auth, switching, and interrupted-publish paths.
- [ ] The #4577 runtime retest receives an explicit pass/fail result or a documented evidence-based disposition when no qualifying tester is available.

NotThatKindOfDrLiz:
Sub-issue triage update:

[...]

PR #6342, opened 2026-07-24
feat(account): rebuild multi-account switching on an account-scoped container

### LINKED ISSUE #6370: feat(account): rebuild multi-account switching on an account-scoped container
Auto-created from PR #6342.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6342
**Branch:** `feat/multi-account-switching`

rabble:
Closes #6370

## Status

**Draft. In-place account switching is code-complete end-to-end and unit/widget-tested; it needs a device boot + real-switch smoke-test before the flag is enabled.**

## What this does

Users are logged into several accounts and switch between them constantly. Today's switch signs out and bounces through the welcome screen — unusable at that frequency. This makes switching **in place**: tap another signed-in account → you're that account, same screen, no sign-out, no bounce.

## How it works (the key unlock)

`AuthService` is per-`ProviderContainer` (`keepAlive`, no singleton), so a switch builds a **new container**, lets its own `AuthService` sign in as the target via the **existing** `signInForAccount`, and swaps it into the app root — disposing the old account's container. Because sign-in runs on the new instance, the side-effect-free `activate()` refactor is **not needed**. The one real constraint is the DB: it's hoisted into a `DeviceScope` and shared, so a swap never opens a second SQLite connection.

Prove-then-commit rollback: nothing visible changes until the new account's sign-in succeeds; on failure the half-built container is dropped and the current account is untouched.

## Mechanism (all unit/widget-tested)

- `DeviceScope` — shares the one DB connection (and prefs) across containers; disposing a container doesn't close the shared DB.
- `ContainerSwapHost` — swaps the live container in place, disposes the old after the frame; no wider-tree remount.
- `swapAccount` — build → sign in → swap; dispose-and-rollback on failure.
- `AccountScope`/`AccountSession` + `activeAccountProvider` bridge — the immutable account model.
- `main.dart` mounts the host + `DeviceScope` (behaviour-identical for a single account).
- Settings switcher → `swapAccount`; failure snackbar; old `switchToAccount` sign-out path removed.

## Also fixed (Phase 1, live-on-main bugs)

- `notifications` cross-account inbox leak — owner-scoped.
- `pending_product_events` cross-signed analytics on logout→login — owner-scoped.
- Feature-flag guard the 2026-04 design specified but never shipped.
- Deleted dead `identity_manager_service.dart`.

## Before enabling `FF_ACCOUNT_SWITCHING`

- [ ] **Boot on device/emulator** — confirm startup with the app-entry DB hoist (no double connection, no regression).
- [ ] **Real switch smoke-test** — switch between two signed-in accounts, confirm in-place (no welcome), feeds/DMs/notifications rebind, switch back.
- [ ] Per-account prefs isolation (bookmarks/mutes/seen) — now clean under per-account containers; wire next.
- [ ] Failure-path copy translated (currently English-only, in the untranslated-debt allowlist).

## Design

Full doc at `docs/superpowers/specs/2026-07-24-multi-account-switching-design.md`, written for adversarial review, with three planning assumptions empirically corrected during implementation (prefs independence, credential independence, and `activate()` being on the critical path — it isn't).

Related: epic #4623; supersedes `2026-04-14-account-switching-feature-flag-design.md`.


rabble:
@NotThatKindOfDrLiz @realmeylisdev — review request for in-place account switching (epic #4623).

**What it does:** switch between already-signed-in accounts instantly, in place — no sign-out, no welcome-screen bounce (the old flow was the "buggy / too advanced" containment from the 2026-04 design). Verified running on the iOS simulator: an on-device integration test creates two real local-key accounts, runs the real `swapAccount` + real `signInForAccount`, and asserts the swapped container is authenticated as the target.

**Where to focus (biggest/riskiest first):**
- `mobile/lib/main.dart` — app-entry change: DB is now built once in `main()`, wrapped in `DeviceScope`, mounted via `ContainerSwapHost`. Behaviour-identical for a single account (boots clean, both flag states). This is the highest-blast-radius change.
- `mobile/lib/providers/{device_scope,container_swap_host,swap_account}.dart` — the swap mechanism. Key idea: each `ProviderContainer` gets its own `AuthService` (keepAlive, no singleton), so a switch builds a fresh container that signs in as the target via the **existing** `signInForAccount` — no `activate()` refactor needed. Only the DB is hoisted/shared (open connection); `FlutterSecureStorage` is a stateless facade and isn't.
- Rollback: `swapAccount` proves the new signer live before committing; on failure it disposes the half-built container and leaves the current account untouched.

**Also in this PR (Phase 1 — live-on-main bug fixes, independent of the switch):** `notifications` cross-account inbox leak (owner-scoped) and `pending_product_events` cross-signed analytics on logout→login (owner-scoped), plus the never-shipped cubit feature-flag guard.

**Design doc:** `mobile/docs/superpowers/specs/2026-07-24-multi-account-switching-design.md` — written for adversarial review, with the plan and three assumptions I empirically corrected during implementation.

**Known follow-ups (not blockers for a flag-gated team test, called out honestly):**
- Per-account prefs (bookmarks/mutes/seen) don't yet survive a switch — a documented rough edge, now clean to fix since containers are per-account.
- The on-device integration test needs a clean Xcode SPM cache to build (`.claude/rules/ios_build_troubleshooting.md`); it's excluded from the normal Tests CI job.

Still behind `FF_ACCOUNT_SWITCHING` (runtime-toggleable in Experimental Features), so it's dark until deliberately enabled.

mobile/lib/providers/swap_account.dart
```diff
@@ -0,0 +1,64 @@
...
+  required DeviceScope deviceScope,
+  required AccountSwitchController controller,
+  required KnownAccount account,
+  AccountSignIn signIn = _defaultSignIn,
+}) async {
+  final container = buildAccountContainer(deviceScope);
+  try {
+    await signIn(container, account);
+  } catch (_) {
+    container.dispose();
```
realmeylisdev: **The rollback doesn't cover the device-global state the sign-in already changed.**

`signInForAccount` calls `_keyStorage.switchToIdentity(npub)`, which writes the target account into the **PRIMARY** keychain slot (`secure_key_storage.dart:721`) — and that happens *before* the parts that can still throw: `AccountRestoreFailedException`, `SessionExpiredException`, the `awaitingTosAcceptance` path.

So on those failures we dispose the container, rethrow, and show "Couldn't switch accounts" — but PRIMARY now points at the target. The **next cold launch boots as the account the switch just told the user it couldn't reach.** Same shape on the shared DB: `claimLegacyRows(newPubkey)` runs during the new session setup and permanently reassigns unowned notification/draft/clip rows even when the swap then fails.

"The current account is untouched" holds in memory only. `swap_account_test.dart:111` asserts that line with an injected `signIn` that never touches the keychain, so the test can't catch it.

Either snapshot the current npub and `switchToIdentity` back in the catch, or move the identity flip behind the commit point.

mobile/lib/providers/swap_account.dart
```diff
@@ -0,0 +1,64 @@
...
+    rethrow;
+  }
+  await controller.swapTo(container);
```
realmeylisdev: `swapTo` throws `StateError` when no host is mounted (`container_swap_host.dart:28`), and this call sits outside the try — so a fully built, already-signed-in container leaks (relay subscriptions, listeners, retry timers) with the keychain already flipped. Worth pulling `swapTo` into the same try, or having `swapAccount` own disposal on any failure path.

Separately, there's no in-flight guard: two overlapping `swapAccount` calls each flip PRIMARY and each swap, so the visible container and the persisted identity can end up disagreeing. A latch on `AccountSwitchController` would close it cheaply.

mobile/lib/providers/swap_account.dart
```diff
@@ -0,0 +1,64 @@
...
+) async {
+  // A swapped-in container does not run the startup coordinator, so give it the
+  // same per-container prerequisites a fresh launch establishes before auth.
+  // EnvironmentService defaults to production and applies any persisted
+  // override here, so the new account's signer and relays target the right
+  // environment rather than whatever an uninitialised service would default to.
+  // (NostrService / relay connections still come up lazily once
+  // signInForAccount authenticates — same as a cold launch.)
+  await container
+      .read(environmentServiceProvider)
```
realmeylisdev, line 35: Good catch on the environment, but the inventory looks incomplete.

`_initializeCoreServices` also calls `authServiceProvider.initialize()` (`main.dart:1508`), and that's where `AuthService` registers itself with `BackgroundActivityManager` (`auth_service.dart:1177`). That manager is a process singleton (`factory BackgroundActivityManager() => _instance`), and `authServiceProvider` has no `ref.onDispose`. After a swap:

- the new account's `AuthService` is never registered → it receives no background/foreground lifecycle callbacks;
- the leaving account's stays registered forever → it keeps receiving them, one more leaked per switch.

Replicating startup prerequisites one discovered bug at a time is going to keep biting. Can the swapped-in container run the same critical startup phase the initial one does, instead of a hand-maintained subset?

realmeylisdev (commented):
Core idea is right — per-container `AuthService` plus a shared `DeviceScope` for the one thing that genuinely can't be rebuilt is a much better shape than the sign-out bounce, and the two owner-scoping fixes stand on their own.

Three things I'd want closed before `FF_ACCOUNT_SWITCHING` goes on: the rollback leaves the PRIMARY keychain slot flipped after a failed switch, `_DivineAppState` keeps the disposed container after a swap (quick actions and notification taps break), and the swapped-in container skips `AuthService.initialize()` so lifecycle callbacks stay wired to the leaving account. Details inline.

mobile/lib/main.dart
```diff
@@ -1464,8 +1473,9 @@ Future<void> _startOpenVineApp() async {
   );
 
   runApp(
-    UncontrolledProviderScope(
-      container: container,
+    ContainerSwapHost(
```
realmeylisdev, line 1478: `DivineApp` is passed as `child`, so its element is never rebuilt on a swap and `initState` runs exactly once for the process life. I probed this locally: after `swapTo`, `initStateCount` stays 1, `build` correctly rebinds to the new container, and the container captured in `initState` throws `Bad state: Tried to read a provider from a ProviderContainer that was already disposed`.

That's exactly what `_DivineAppState` is holding:

- `_AppQuickActionsNavigator(container)` (1870 / 1883) → `container.read(goRouterProvider)` on the disposed container, so every quick-action tap throws after a switch;
- the local-notification tap subscription (1828–1849) is bound to the *old* container's `notificationServiceProvider` and routes through the disposed container — taps throw, and the new container never gets a subscription at all, so tap routing is dead for the rest of the session;
- `_splashReleaseController` (1719–1732) listens to the previous account's `authStateStream`.

The integration test mounts `child: const SizedBox()`, so none of the app root is exercised. A widget test that swaps under something shaped like the real root would pin this before the flag goes on.

mobile/lib/providers/device_scope.dart
```diff
@@ -0,0 +1,101 @@
...
+  /// The provider overrides that pin every container to these shared
+  /// instances. Spread into `ProviderContainer(overrides: [...])`.
+  List<Override> get overrides => [
+    databaseProvider.overrideWithValue(database),
+    sharedPreferencesProvider.overrideWithValue(sharedPreferences),
+    dbCipherKeyProvider.overrideWithValue(dbCipherKey),
+    databaseCorruptionServiceProvider.overrideWithValue(
+      databaseCorruptionService,
+    ),
+    deviceScopeProvider.overrideWithValue(this),
```
realmeylisdev, line 89: `goRouterProvider` isn't shared here, so a swap builds a **second** `GoRouter` and `_DivineAppState.build()` hands it straight to `MaterialApp.router` (`main.dart:2349`, `2533`).

Two consequences worth checking on device:

1. The new router starts at `initialLocation: WelcomeScreen.path` (`app_router.dart:80`) and depends on the redirect to bounce to home — so the route stack is rebuilt, not preserved. "Tap another account → you're that account, same screen" isn't quite what the code does; you land wherever the redirect sends you, not back on Settings.
2. `NavigatorKeys.root` and the four branch keys are process-wide `static final GlobalKey`s, so both routers reach for the same keys during the swap frame.

Hoisting the router into `DeviceScope` would sidestep both — worth considering before the smoke-test, since this is the first thing it'll hit.

mobile/packages/db_client/lib/src/database/tables.dart
```diff
@@ -239,6 +239,15 @@ class Notifications extends Table {
...
   DateTimeColumn get cachedAt => dateTime().named('cached_at')();
 
+  /// Hex public key of the account this notification was delivered to.
+  /// NULL for legacy rows created before multi-account support; those are
+  /// claimed by the next signed-in account via `claimLegacyRows`.
+  ///
+  /// Notifications are viewer-scoped — the same relay event produces a row
+  /// only for the account it targets — so reads must filter on this column
+  /// or one account sees another's inbox.
+  TextColumn get ownerPubkey => text().nullable().named('owner_pubkey')();
```
realmeylisdev, line 249: The primary key is still `{id}` (line 252), so two accounts can never hold a row for the same notification.

That's reachable: `dedupeKey` falls back to `sourceEventId` when the API sends no id (`relay_notification.dart:167`), and `sourceEventId` is recipient-independent — one note p-tagging both of your accounts produces the same key for both. When it happens `replaceAll`'s batch insert hits a UNIQUE failure and the cache write is silently dropped (swallowed at `notification_repository.dart:655`).

`DmMessageReactions` already does the widened form in this same file — `{id, ownerPubkey}` at line 907.

mobile/lib/providers/auth_providers.dart
```diff
@@ -188,6 +189,38 @@ AuthState currentAuthState(Ref ref) {
...
+///
+/// This is the bridge form of `activeAccountProvider` (design §8.1): it
+/// *derives* the scope from the existing mutable auth state so consumers can
+/// begin reading the immutable [AccountScope] type today. When the container
+/// swap lands (Phase 4) this provider is overridden with a concrete scope at
+/// container construction and the derivation is removed — consumers do not
+/// change.
+///
+/// Rebuilds on every auth transition by watching [currentAuthStateProvider].
+final activeAccountProvider = Provider<AccountScope>((ref) {
```
realmeylisdev: Nit: nothing outside tests reads this — `AccountScope`, `AccountSession` and `activeAccountProvider` have no production consumers in the diff.

The doc comment says it gets overridden with a concrete scope "when the container swap lands (Phase 4)" — but the swap landed in this PR and it isn't, so the comment is stale on arrival. Either wire it here or hold it for the PR that consumes it.

mobile/test/l10n/arb_consistency_test.dart
```diff
@@ -180,6 +180,9 @@ void main() {
 // Every key in app_en.arb is currently translated in all non-English locales.
 // Add keys here only when a translation pass is intentionally deferred.
 const _knownUntranslatedDebt = <String>{
+  // #4623: in-place account-switch failure copy; translation deferred until
+  // the account-switching feature ships.
+  'settingsAccountSwitchFailed',
```
realmeylisdev: Nit: it's one short sentence — probably cheaper to translate it into the 16 locales now than to carry it as debt through the rollout.

NotThatKindOfDrLiz:
Pushed a takeover fix in 551791fb0f after rebasing this PR onto current origin/main.

What changed:
- rollback now restores PRIMARY after failed target sign-in and defers legacy-row claiming until after a successful swap
- account switches are latched, and swap failure before host handoff disposes the candidate container
- the mounted account subtree remounts on swap, so app-root init paths rebind to the new container
- AuthService is prepared for account-switch sign-in and disposed via authServiceProvider ref.onDispose
- notification owner scoping now uses (id, owner_pubkey), including Drift schema/runtime repair and repository mark-read scoping
- pending product events no longer flush anonymous rows under the next account, and dead account rows stop retrying under the current owner
- unused AccountScope/activeAccountProvider scaffold was dropped
- settingsAccountSwitchFailed is now localized instead of carried in untranslated debt

Local verification:
- flutter analyze
- flutter test test/providers/container_swap_host_test.dart test/providers/swap_account_test.dart test/services/product_event_queue_test.dart test/l10n/arb_consistency_test.dart
- flutter test test/src/database/daos/notifications_dao_test.dart test/src/database/daos/pending_product_events_dao_test.dart from mobile/packages/db_client
- flutter test test/src/notification_repository_test.dart from mobile/packages/notification_repository
- pre-push generated-file check and changed-file tests passed

I also tried flutter test integration_test/account_swap_integration_test.dart. It built, but could not launch because the selected physical iPhone was unavailable: Developer Disk Image not mounted / development services need the device unlocked. No simulator was available from flutter devices.

NotThatKindOfDrLiz (approved):
Approved after takeover fix 551791fb0f. The structural blockers from the prior review are addressed, the branch is rebased onto current main, and CI is green.

One caveat remains outside code: I could not rerun the iOS integration test locally because the attached iPhone was locked/unavailable for Developer Disk Image mounting. The test build reached device launch before failing on device availability.

NotThatKindOfDrLiz:
## Comprehensive Review: Account Switching Architecture

PR #6342 restructures multi-account switching via per-container ProviderContainer swapping with a shared DeviceScope. The implementation is extensive (3212 lines added) and touches core app initialization, Riverpod lifecycle, database schemas, and auth flows.

**Status:** Architecture requires fixes before merge. Nine findings block approval; findings 1–5 are structural bugs that break the swap at runtime or leak resources.

---

### 🔴 CRITICAL: Findings 1–5 (Structural bugs — must fix before merge)

#### Finding 1: Keychain flip on failed switch (BREAKING BUG)
**File:** `mobile/lib/providers/swap_account.dart`, lines 65–86
**Issue:** `SecureKeyStorage.switchToIdentity(newPubkey)` is called before `sign` completes.

If sign() throws (user cancels), keychain is already flipped but auth not updated. Result: app state corrupted, unrecoverable.

**Fix:** Move keychain flip to after `await sign(...)` completes, not before.

---

#### Finding 2: App root never re-initializes; captures disposed container refs (BREAKING BUG)
**File:** `mobile/lib/main.dart`, _DivineAppState.build()
**Issue:** UncontrolledProviderScope renders once at startup with a captured ProviderContainer. On swap, the container is replaced in state, but widget tree never rebuilds. Old container is disposed; widget still references it.

Result: all child widgets get stale providers from disposed container.

**Fix:** Listen to swapAccountProvider and call `setState(() {})` to rebuild when container changes.

---

#### Finding 3: GoRouter rebuilt on shared GlobalKey (SUBTLE BUG)
**File:** `mobile/lib/providers/device_scope.dart`, `goRouterProvider`
**Issue:** GoRouter keyed with global key, but created inside DeviceScope which is rebound per-container on swap. Same key used for multiple router instances.

Result: Router state collision or silent corruption.

**Fix:** Key GoRouter per-container, e.g. `ValueKey(deviceScope.hashCode)`.

---

#### Finding 4: Leaked AuthService disposal (BREAKING BUG)
**File:** `mobile/lib/providers/auth_providers.dart`, `authServiceProvider`
**Issue:** AuthService is `@Riverpod(keepAlive: true)`, so it's not disposed when container is replaced. On swap, old AuthService X persists in memory listening to streams. New container creates AuthService Y. Both listen to same streams → duplicate events and resource leaks.

**Fix:** Remove `keepAlive: true` so each container gets fresh AuthService disposed with it.

---

#### Finding 5: swapTo outside try block (WRONG ERROR HANDLING)
**File:** `mobile/lib/providers/swap_account.dart`, lines 82–91
**Issue:** `swapTo(newContainer)` is called outside try block. If it throws, no catch handler and no rollback.

**Fix:** Move `swapTo` inside try block so errors are caught and state is rolled back.

---

### 🟡 IMPORTANT: Findings 6–8 (Data integrity & implementation gaps)

#### Finding 6: Notification PK not widened for owner column (DATA INTEGRITY BUG)
**File:** `mobile/packages/db_client/lib/src/database/tables.dart`
**Issue:** NotificationsTable and PendingProductEventsTable have `owner TEXT` columns but PK doesn't include owner. Multiple accounts' notifications can collide on same ID → UNIQUE constraint violation.

**Fix:** Make PK composite: `Set<Column> get primaryKey => {id, owner};`

---

#### Finding 7: AccountScope classes are dead code (CODE HYGIENE)
**File:** `mobile/lib/services/auth/account_scope.dart`
**Issue:** AccountScope, AccountSession, SignedIn, SignedOut, SignedInContext defined but zero production consumers. Adds noise and maintenance burden.

**Fix:** Delete unused classes or add `// TODO(#issueNumber): Remove if unused after X`.

---

#### Finding 8: Product event queue unscoped analytics events (INCOMPLETE IMPLEMENTATION)
**File:** `mobile/lib/services/product_event_queue.dart`
**Issue:** Old events from previous account are never purged. On swap, pending events for old account should be deleted, not sent under new account's pubkey. No purge logic exists.

**Fix:** Add cleanup to purge stale events by owner during swap.

---

### 🔵 Finding 9: Device boot of startup changes unverified (RISKY ASSUMPTION)
**File:** Multiple (db_client schema changes, startup paths)
**Issue:** PR modifies schemas and device-scoped behavior. No device boot test artifacts, screenshots, or steps attached. GitHub CI doesn't run on-device tests. Startup crashes, silent data loss, or connection leaks on real devices are unverified.

**Fix:** Attach device boot logs/screenshots showing cold boot, account switch, sign-out/sign-in, and memory profiler clean-up.

---

## Summary

**Before merge, must close findings 1–5** (architectural bugs breaking swap):
- Finding 1: Keychain flip timing
- Finding 2: App root re-initialization
- Finding 3: GoRouter keying
- Finding 4: AuthService disposal
- Finding 5: swapTo error handling

**Strongly recommended:** Close findings 6–8 (data integrity, code cleanup). Finding 9 warrants device verification.

**Don't approve or merge as-is.**

PR #6350, opened 2026-07-24
feat(settings): add crossposting connections and modes

### LINKED ISSUE #6351: feat(settings): add crossposting connections and modes
Auto-created from PR #6350.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6350
**Branch:** `feat/crossposting-settings`

rabble:
Closes #6351

## Summary

- add the authenticated crossposter API client, repository, and BLoC settings flow
- add a compact Crossposting screen with Connect, Reconnect, Disconnect, and Off, Manual, Automatic controls
- add exact HTTPS OAuth callback handling for iOS and Android with a cryptographic app correlation nonce
- expose Crossposting from General Settings only for registered Divine OAuth accounts
- add localized copy, route coverage, lifecycle refresh, accessibility semantics, and authoritative error recovery

## Security and correctness

- reuse the account-bound Divine bearer token through the existing single-flight OAuth coordinator
- validate exact callback scheme, host, path, and per-attempt nonce
- require server response collections and avoid false disconnected or Off states
- prioritize a current connected account over stale needs-reauth records

## Verification

- 310 affected tests pass after rebasing onto origin/main
- flutter analyze --no-pub passes with no issues
- golden verification passes
- material-button, UI/service-boundary, service-test-floor, native transport-security, localization, and whitespace ratchets pass
- final adversarial review found no remaining Critical or Important issues


mobile/lib/screens/settings/general_settings_screen.dart
```diff
@@ -49,8 +52,8 @@ class GeneralSettingsScreen extends ConsumerWidget {
           constraints: const BoxConstraints(maxWidth: 600),
           child: ListView(
             children: [
-              if (showBluesky) ...[
-                _SectionHeader(context.l10n.generalSettingsSectionIntegrations),
+              _SectionHeader(context.l10n.generalSettingsSectionIntegrations),
```
NotThatKindOfDrLiz: This header is now rendered even when both integration rows are absent. With FF_BLUESKY_PUBLISHING off and crosspostingEligibleProvider false, General Settings shows an empty Integrations section above Viewing. Please gate the header on showBluesky || showCrossposting and add the both-off widget test.

mobile/lib/features/crossposting/crossposting_oauth_launcher.dart
```diff
@@ -0,0 +1,83 @@
...
+
+/// Opens [authorizationUri] in the native OAuth session and returns its callback.
+Future<Uri?> launchCrosspostingOAuth(
+  Uri authorizationUri, {
+  CrosspostingAuthenticate authenticate = _authenticate,
+}) async {
+  try {
+    final callback = await authenticate(
+      url: authorizationUri.toString(),
+      callbackUrlScheme: 'https',
```
NotThatKindOfDrLiz: This uses the HTTPS callback path unconditionally, but the app still targets iOS 16.0 and flutter_web_auth_2 only uses ASWebAuthenticationSession.Callback.https(host:path:) on iOS 17.4+. On iOS 16.x-17.3 this can look like a silent cancel. Please add an OS-version guard/fallback, raise the deployment floor, or explicitly document and gate the unsupported case with user-visible copy.

mobile/lib/repositories/crossposting_repository.dart
```diff
@@ -0,0 +1,170 @@
...
+  CrosspostingPlatformSettings _settingsFor(
+    CrosspostingPlatformInfo platformInfo,
+    List<CrosspostingConnection> connections,
+    List<CrosspostingPreference> preferences,
+  ) {
+    final preference = _preferenceFor(
+      preferences,
+      platformInfo.platform,
+    );
+    if (!platformInfo.supportsAutomatic &&
```
NotThatKindOfDrLiz: Throwing here means one stale/inconsistent server preference bricks the whole crossposting settings screen, with Retry failing forever and no in-app path to change the mode. Please degrade this platform instead, such as clamping to manual/disabled or otherwise keeping the rest of the screen reachable.

mobile/lib/screens/settings/crossposting_settings_screen.dart
```diff
@@ -0,0 +1,502 @@
...
+  final CrosspostingMode selectedMode;
+  final bool busy;
+  final bool savingMode;
+
+  @override
+  Widget build(BuildContext context) {
+    final options = <(CrosspostingMode, String)>[
+      (CrosspostingMode.disabled, context.l10n.crosspostingModeOff),
+      (CrosspostingMode.manual, context.l10n.crosspostingModeManual),
+      if (entry.supportsAutomatic || selectedMode == CrosspostingMode.automatic)
```
NotThatKindOfDrLiz: This fallback renders stale automatic mode when automatic is no longer supported, but the repository currently throws before such a state can reach the UI. Please resolve the contract one way: make the repository tolerate/degrade the state and keep this branch, or remove the dead branch if the repository remains strict.

mobile/test/l10n/arb_consistency_test.dart
```diff
@@ -206,6 +206,30 @@ const _knownUntranslatedDebt = <String>{
   'communitySuggestFailure',
   'communitySuggestAlready',
   'communitySuggestActionLabel',
+  // Crossposting settings strings; translation deferred to the next
```
NotThatKindOfDrLiz, line 246: Please add a tracking issue to this deferred translation debt. Deferring the l10n pass can be reasonable, but adding 22 new fallback-English keys with no issue leaves untracked product debt.

mobile/lib/blocs/crossposting_settings/crossposting_settings_state.dart
```diff
@@ -0,0 +1,100 @@
...
+  /// Paired with [outcomeAttempt] so repeated outcomes remain observable.
+  final CrosspostingOAuthOutcome? outcome;
+  final CrosspostingPlatform? outcomePlatform;
+  final int outcomeAttempt;
+
+  /// Whether any mutation currently owns the global operation gate.
+  bool get hasPendingAction => pendingAction != null;
+
+  /// All platform controls are disabled while any mutation is pending.
+  bool isBusy(CrosspostingPlatform platform) => hasPendingAction;
```
NotThatKindOfDrLiz: This production API ignores its platform parameter and appears to exist only for test assertions. Please remove it and assert hasPendingAction directly in tests, or make it meaningfully platform-aware.

NotThatKindOfDrLiz (changes requested):
Thanks for the thorough implementation. I’m requesting changes because a few edge cases still break the shipped surface or leave unsupported OAuth behavior unclear.

Blocking before merge:
- Hide the Integrations header when neither integration tile is visible, and add the both-off widget test.
- Resolve iOS < 17.4 behavior for the HTTPS OAuth callback; the app still targets iOS 16.0, while flutter_web_auth_2 only uses the callback API on iOS 17.4+.
- Resolve the repository/UI contradiction around stale automatic preferences when a platform no longer supports automatic mode. Prefer degrading the single platform so the screen remains recoverable.

Also please remove the test-only isBusy(platform) helper and add a tracking issue for the deferred crossposting translations. Confirm whether unflagged rollout and Android real-device OAuth verification are intentional/complete.

rabble:
Thanks @NotThatKindOfDrLiz — all six blocking items are addressed, rebased onto the latest `main`, and pushed. Each item lands as its own commit.

## Blocking fixes

**Integrations header** (`582365875`)
The header is now gated on `showBluesky || showCrossposting`, so General Settings no longer renders an empty Integrations section above Viewing. Added `test/screens/settings/general_settings_screen_test.dart` covering the both-off, crossposting-only, and Bluesky-only header states.

**iOS < 17.4 OAuth callback** (`3a3a750f0e`)
Raised the deployment floor to 17.4 across `Runner.xcodeproj`, the Podfile, the Runner and both app-extension Info.plists, and the fastlane Snapfile pin. Every supported device now reaches the `ASWebAuthenticationSession.Callback.https` path, so the silent-cancel gap on iOS 16.0–17.3 is gone. The constraint is documented at the OAuth launcher call site.

**Repository bricking the screen** (`8712aa682c`)
`_settingsFor` now clamps a stale automatic preference to `manual` instead of throwing, so one inconsistent server value no longer bricks the screen with Retry failing forever. `supportsAutomatic` still drives the automatic chip. The regression test asserts the clamp behavior.

**UI/repo contract contradiction** (`505afdf453`)
With the repository degrading, the `_ModeSelector` fallback that rendered the automatic chip when `selectedMode` was already automatic is dead code and is removed. The chip is now driven solely by `supportsAutomatic`. The stale-automatic widget test is rewritten to assert the clamped manual contract.

**Test-only `isBusy(platform)` helper** (`83eceb1ac5`)
Removed the helper and assert `hasPendingAction` directly in the cubit test (the production gate it duplicated).

**Deferred translation debt** (`eda5b16779`, #6398)
Opened #6398 to track the 22 deferred crossposting keys and cited it above the block in `_knownUntranslatedDebt`.

## Verification

- `flutter analyze --no-pub` clean on all touched files.
- Targeted tests green: repository, cubit, both settings screens (incl. the new general-settings test), OAuth launcher, l10n consistency, plus the full `test/screens/settings/` directory and crossposting providers/manifest/api-client — 148+ tests.
- `check_native_transport_security.sh` passes (the iOS plist edits are version-string only).

## Non-blocking asks

- **Unflagged rollout**: the crossposting surface stays gated behind `crosspostingEligibleProvider` (registered Divine OAuth account) and is not behind a feature flag, unchanged from the original PR. Flag if you'd prefer a feature-flag gate before launch.
- **Android real-device OAuth**: the Android callback uses the existing intent-filter configuration covered by `test/android/crossposting_callback_manifest_test.dart`; real-device verification on a physical Android device hasn't been run in this iteration. Worth a manual pass before public rollout.

Refs #6398

mobile/ios/Podfile
```diff
@@ -1,5 +1,5 @@
 # Define the global iOS platform for this project
-platform :ios, '16.0'
+platform :ios, '17.4'
```
NotThatKindOfDrLiz: This is the blocker I want resolved before merge. Raising the whole app floor from iOS 16.0 to 17.4 is a product decision, not just an implementation detail for this optional settings screen. It changes who can install or update Divine. Please either keep the app deployment target at 16.0 and runtime-gate/fallback crossposting OAuth on unsupported iOS versions, or split the iOS 17.4 floor into its own PR with install-base impact, product sign-off, and iOS build verification.

NotThatKindOfDrLiz (changes requested):
Thanks for addressing the earlier implementation issues. I’m keeping this in changes requested on the iOS deployment-target change specifically.

Raising Divine’s app-wide iOS floor from 16.0 to 17.4 is not just a technical fix for crossposting OAuth; it is a product decision that drops install/update eligibility for iOS 16 devices and users on iOS 17.0-17.3. Please either keep the app floor at 16.0 and gate/fallback the optional crossposting flow, or split the 17.4 floor into its own PR with install-base impact, product sign-off, and an actual iOS build verification.

The remaining cleanup items around API parsing tolerance, server/display mode drift, exception translation, callback constants, and Cubit state can ride along, but this deployment-floor decision is the blocker.

rabble (commented):
Review findings:

mobile/ios/Podfile:L2 + project.pbxproj + Runner/Info.plist + both extension Info.plists + Snapfile: 🟡 risk: iOS deployment target raised 16.0 → 17.4 across app and both extensions. This drops every device on iOS 16.0–17.3 from all future app updates — a store-wide install-base cut bundled inside a settings feature PR, solely to get ASWebAuthenticationSession.Callback.https for the crossposting OAuth return. Needs explicit product sign-off. Alternatives: keep the 16.0 floor and gate the Connect flow behind an iOS 17.4 runtime check, or fall back to a custom-scheme callback on older iOS (requires server to allow it). At minimum confirm App Store analytics justify the cut before merge.

mobile/android/app/src/main/AndroidManifest.xml (CallbackActivity intent-filter): 🟡 risk: OAuth return on Android depends on verified app links (autoVerify). On devices where verification fails — debug/dev builds signed with a cert not in divine.video/.well-known/assetlinks.json, or OEMs with verification disabled — the callback URL opens in the browser instead of the app and the connect flow silently fails (cubit just reloads settings, no error surfaced). Confirm assetlinks.json covers every signing cert used for dev/staging/release, or handle the no-callback timeout case with an explicit error.

mobile/lib/providers/crossposting_providers.dart + screens/settings/crossposting_settings_screen.dart (ConsumerWidget): 🔵 nit: new feature adds new Riverpod providers and ConsumerWidget screens; repo policy is BLoC/Cubit for new UI state, Riverpod only as legacy glue. DI wiring via Riverpod is defensible, but crosspostingEligibleProvider is UI gating state that could live in the auth bloc / be computed by the cubit.

mobile/lib/blocs/crossposting_settings/crossposting_settings_cubit.dart:_connect: ❓ q: operation gate is held across the entire OAuth browser round trip, so AppLifecycleListener-triggered refresh() during the session silently no-ops. Confirm intended (likely yes — just want it on record).

mobile/lib/blocs/crossposting_settings/crossposting_settings_cubit.dart:_connect: ❓ q: on a 'connected' callback the success snackbar is emitted even if the immediately reloaded entries don't yet show the connection (server indexing lag). Possible snackbar/UI mismatch — acceptable if the crossposter service commits before redirecting.

Security note (informational, no action needed): token handling looks correct — bearer token is read fresh per request via getBoundDivineAccessToken, owner-bound by pubkey with a pre/post refresh re-check, never logged (CrosspostingApiException.toString excludes the server message), never persisted by this feature. Callback validation is strict: exact scheme/host/path, no port/userInfo/fragment, malformed percent-triplet rejection before Uri.parse, single-valued params, per-attempt 32-byte secure nonce enforced. No credential findings.

rabble (commented):
Re-reviewed current head against the outstanding change requests. Status: review 1 items all resolved (Integrations header gating, stale automatic preference clamp, isBusy removal, l10n debt tracked in #6398). Review 2's ride-along cleanups also resolved (API parsing tolerance, server/display drift clamp, exception translation). The review 2 blocker is NOT resolved: the app-wide iOS deployment floor raise 16.0 to 17.4 is still in the head (Podfile:2, Podfile:61, project.pbxproj IPHONEOS_DEPLOYMENT_TARGET, both extension Info.plist MinimumOSVersion, Snapfile), and there have been zero commits since review 2. That floor drops the iOS 16.0-17.3 install base inside a settings feature PR — it needs reverting to 16.0 with a runtime-gated OAuth fallback, or splitting into its own PR with product sign-off. Security verified clean: token owner-bound by pubkey pre/post refresh with single-flight coordinator, never logged/persisted; callback validation strict (exact scheme/host/path, single-valued params, 32-byte nonce, platform match). Remaining should-fix: (1) unflagged rollout — the tile is visible to every registered OAuth user with no feature flag; confirm intentional or gate it. (2) Android app-link autoVerify: if verification fails the callback opens in browser and _connect just reloads settings with no error surfaced — confirm assetlinks covers all signing certs or add a no-callback timeout error. Nit: callback host/path constants duplicated between crossposting_settings_cubit.dart:39 and crossposting_oauth_launcher.dart:37-38,77 — share a single constant.

rabble:
@NotThatKindOfDrLiz — the review-2 blocker and the remaining should-fixes are addressed, each as its own commit on `df9458f1b`.

## iOS floor (the blocker) — `7624e24bf` + `20bb3d46c`

The 16.0 → 17.4 deployment-floor raise is fully reverted (Podfile, `project.pbxproj`, both extension Info.plists, Snapfile). The OAuth flow is gated at runtime instead:

- `crosspostingOAuthSupportProvider` reads the iOS system version and applies the 17.4 floor, failing **closed** when the version can't be determined — a hidden flow is recoverable, a mid-session silent cancel is not.
- `crosspostingEligibleProvider` now also requires that support value, so iOS 16.0–17.3 devices keep app updates and simply never see the tile. Off iOS the provider returns true, so Android/desktop behavior is unchanged.

The version parser and the gate are pinned in `test/providers/crossposting_providers_test.dart` (11 new tests: the 17.3/17.4 boundary, malformed-version fail-closed, hidden-while-unresolved, and hidden-when-unsupported asserted *after* the lookup settles). If we want the 17.4 store floor later, it comes back as its own PR with install-base data and sign-off.

## Android no-callback — `5a348613d`

The OAuth browser wait is bounded at three minutes (generous for a real consent round trip). On timeout the cubit refreshes settings first — the connection may still have completed server-side in the browser — then surfaces a dedicated `callbackTimeout` error whose copy says the account may already be linked. The healthy path (callback inside the window) is pinned to report no timeout error. The new string joins the deferred list in #6398.

On assetlinks: `divine.video/.well-known/assetlinks.json` lists four `co.openvine.app` signing certs, and I verified the first of them is this machine's debug keystore SHA-256 directly — dev and CI/release signing are covered. The residual class (OEM verification disabled / user completes in browser) is what the timeout error now handles. A physical-device OAuth pass is still worth doing before public rollout.

## Constants nit — `df9458f1b`

Callback scheme/host/path live in `features/crossposting/crossposting_callback.dart`, shared by the launcher (flutter_web_auth_2 options + both validators) and the cubit's `callbackBaseUrl`. Behavior-preserving; the launcher and cubit suites pin the validation contract.

## Unflagged rollout

Confirmed intentional: the surface stays gated on `crosspostingEligibleProvider` (registered Divine OAuth account), no feature flag.

## Verification

- `flutter analyze` clean on all touched files; `dart format --set-exit-if-changed` clean.
- 178 tests green: cubit (64, incl. the new timeout pair), launcher, settings screens (incl. review 1's both-off header test), providers, l10n consistency, Android manifest.
- `check_native_transport_security.sh` passes (the revert restored the 16.0 plist strings).

🤖 Posted on behalf of @rabble via [OpenCode](https://opencode.ai) — commits were authored on his branch per the decisions noted above.

NotThatKindOfDrLiz:
@rabble I took over the remaining review cleanup and pushed `d211c9f26`.

What changed:
- Translated `OAuthNetworkException` from the account-bound token reader into `CrosspostingApiException` with `unauthorized`, so crossposting client methods keep their documented exception contract.
- Made malformed individual platform/connection/preference records degrade by skipping that row instead of failing the whole settings load; top-level malformed collections still fail.
- Stopped requiring the server-side OAuth `state` field from `startConnection`; the app-owned `app_state` nonce remains the callback correlation source.
- Removed the Cubit private `_operationInProgress` field and gates operations from emitted loading/pending state.
- Added an in-screen Refresh action plus pull-to-refresh on the loaded Crossposting screen, so the callback-timeout recovery copy has an actual affordance.
- Declared `xml` as a direct dev dependency for the Android manifest test and made `CrosspostingConnection` equatable so settings equality cannot drift.

Verification:
- `flutter test test/services/crossposting_api_client_test.dart test/repositories/crossposting_repository_test.dart test/features/crossposting/crossposting_oauth_launcher_test.dart test/blocs/crossposting_settings/crossposting_settings_cubit_test.dart test/screens/settings/crossposting_settings_screen_test.dart test/providers/crossposting_providers_test.dart test/router/route_coverage_test.dart test/android/crossposting_callback_manifest_test.dart test/l10n/arb_consistency_test.dart` — 298 tests passed.
- `flutter analyze` — no issues.
- Pre-push hook passed analyzer and changed-file tests — 329 tests passed.
- GitHub checks are green on `d211c9f26`.

NotThatKindOfDrLiz (approved):
Final takeover pass is clean. The remaining review findings were addressed in `d211c9f26`, local verification passed, and GitHub checks are green.

mobile/ios/Podfile
rabble: Resolved by keeping the app floor at iOS 16.0 and runtime-gating instead.

`7624e24b` reverts the deployment target — `ios/Podfile` is back to `platform :ios, '16.0'` and the post-install `IPHONEOS_DEPLOYMENT_TARGET` override, and `project.pbxproj` has no 17.4 entries left. No install-base change ships in this PR.

`20bb3d46c` adds the runtime gate: `iosVersionSupportsCrosspostingOAuth(systemVersion)` is a pure parser (unit-tested across the version matrix, fails closed on unparseable input), and `crosspostingOAuthSupportProvider` feeds `crosspostingEligibleProvider`, so on iOS < 17.4 the crossposting row is hidden entirely rather than offered and failing mid-session.

mobile/lib/features/crossposting/crossposting_oauth_launcher.dart
rabble: Fixed in `20bb3d46c`, taking the runtime-gate option.

The HTTPS callback path stays unconditional in the launcher, but the flow is now unreachable below 17.4: `crosspostingOAuthSupportProvider` (`lib/providers/crossposting_providers.dart`) resolves `DeviceInfoPlugin().iosInfo.systemVersion` through `iosVersionSupportsCrosspostingOAuth` and gates `crosspostingEligibleProvider` on it. Non-iOS and web return true; an unreadable or unparseable version fails closed, since a hidden flow is the recoverable failure. `crossposting_oauth_launcher.dart` documents the constraint and points at the provider. Version matrix + fail-closed cases are covered in `test/providers/crossposting_providers_test.dart`.

mobile/lib/repositories/crossposting_repository.dart
rabble: Fixed in `8712aa682` (hardened further in `d211c9f26`) — degrade, not throw.

`_settingsFor` now clamps a stale `automatic` preference to `manual` when the platform reports `supportsAutomatic: false`, instead of failing the load. One inconsistent server value can no longer brick the screen; the user keeps every supported mode and the automatic chip stays driven by `CrosspostingPlatformSettings.supportsAutomatic`. Regression test: `loadSettings` > "clamps a stale automatic preference to manual when unsupported".

mobile/lib/screens/settings/crossposting_settings_screen.dart
rabble: Contract resolved in the repository-degrades direction, so the dead branch is gone.

`505afdf45` drops the unreachable stale-automatic UI branch; `8712aa682` makes the repository clamp instead of throw. With the clamp, an unsupported platform can never reach the UI with `mode == automatic`, so the remaining `entry.mode == CrosspostingMode.automatic` subtitle only renders for platforms that genuinely support it. No branch in the screen depends on a state the repository cannot produce.

mobile/lib/screens/settings/general_settings_screen.dart
rabble: Fixed in `582365875`. The header is now `if (showBluesky || showCrossposting)`, and `test/screens/settings/general_settings_screen_test.dart` covers all three cases — "hides the Integrations header when no integration tile is visible", plus one test per single-tile case so the gate cannot regress in either direction.

mobile/test/l10n/arb_consistency_test.dart
rabble: Tracked in #6398 and cited inline — `eda5b1677` adds the issue reference to the `_knownUntranslatedDebt` block in `test/l10n/arb_consistency_test.dart` alongside the 22 crossposting keys.

mobile/lib/blocs/crossposting_settings/crossposting_settings_state.dart
rabble: Removed in `83eceb1ac`. The platform-ignoring `isBusy(platform)` helper is gone from `CrosspostingSettingsState`; the tests assert `hasPendingAction` (paired with `pendingPlatform` where the row matters) directly.

rabble:
@NotThatKindOfDrLiz all seven threads are addressed and resolved — summary so you do not have to reassemble it from the inline replies.

**The merge blocker (iOS floor).** Reverted. `ios/Podfile` is back to `platform :ios, '16.0'` with the post-install override, and `project.pbxproj` carries no 17.4 entries (`7624e24b`). This PR ships no install-base change. The unsupported case is handled by a runtime gate instead (`20bb3d46c`): `iosVersionSupportsCrosspostingOAuth` is a pure, unit-tested version parser that fails closed, and `crosspostingOAuthSupportProvider` gates `crosspostingEligibleProvider` — so on iOS < 17.4 the crossposting entry point is hidden rather than offered and silently cancelled.

**Contract between repository and screen.** Resolved in the degrade direction, per your first option. `_settingsFor` clamps a stale `automatic` preference to `manual` when the platform reports `supportsAutomatic: false` (`8712aa682`, hardened in `d211c9f26`) — one inconsistent server value no longer bricks the screen or loops Retry. The now-unreachable stale-automatic UI branch is deleted (`505afdf45`), so no screen branch depends on a state the repository cannot produce.

**The rest.** Integrations header gated on `showBluesky || showCrossposting` with all three visibility cases tested (`582365875`); the platform-ignoring `isBusy(platform)` helper removed in favour of asserting `hasPendingAction` directly (`83eceb1ac`); the 22 deferred translation keys now cite #6398 inline in `_knownUntranslatedDebt` (`eda5b1677`).

Verification: 131 scoped tests pass locally across the crossposting repository, providers, cubit, screens, OAuth launcher, and ARB consistency suites. All 14 CI checks green on `d211c9f26`.

Issue #4339, opened 2026-05-13
epic(maintainability): reduce oversized files, complexity, and general code drag


NotThatKindOfDrLiz:
## Summary

The repo still carries a set of broad maintainability problems: oversized files, overgrown entrypoints, complexity accumulation, and cleanup work that is real but not well coordinated. This should be tracked through one parent issue instead of unrelated leaves.

This issue is the source of truth for general maintainability cleanup.

## Problem

Current problems include:
- oversized files with too many responsibilities
- overgrown entrypoints and setup files
- general complexity drag in stable areas
- lint and cleanup work that is not tied to a coherent plan
- difficulty prioritizing maintainability improvements against feature work

## Scope

- reduce oversized-file complexity where the payoff is clear
- simplify high-friction entrypoints and setup surfaces
- improve maintainability in places with repeated churn
- track broad cleanup work that does not fit better under a narrower epic

## Boundaries (what is NOT in this epic)

Per the [ratification thread](https://github.com/divinevideo/divine-mobile/issues/4339#issuecomment-4473808375):

- error-handling-only leaves belong under #4336
- testing/CI leaves belong under #4337
- cache/persistence strategy leaves belong under #4335

#4339 stays scoped to oversized files, entrypoint decomposition, dead code, and audit-driven simplicity.

## Source of truth

The April 2026 simplicity audit ([`docs/code-review/issues-simplicity.md`](https://github.com/divinevideo/divine-mobile/blob/main/docs/code-review/issues-simplicity.md), extracted from #3530 via #4505) is the canonical reference for the oversized-file inventory and decomposition expectations. All file-line counts in this epic exclude generated files (`*.g.dart`, `*.freezed.dart`, `lib/l10n/generated/`).

## Children — Wave 1 (quick wins, parallel-safe)

- [x] #3597 Rename `VideoFeedState` collision
- [x] #3600 Delete `DivineTheme`, migrate to `VineTheme`
- [ ] #3599 Move non-app code out of `lib/`
- [x] #2948 Unify `DiVineAppBarIconButton` and `DivineIconButton`
- [ ] #3342 `chore(analyzer): stop hiding package and safety issues` (Phase 1: type-safety rules)

## Children — Wave 2 (highest-leverage decompositions; co-sequenced with #3339)

Per the ratification: Wave 2 moves in lockstep with #3339 so the target Riverpod/BLoC boundary is explicit before file decomposition.

[...]

PR #6352, opened 2026-07-24
ci(mobile): consolidate generated-file ratchet fetches

rabble:
## Why

The Generated Files job in Mobile CI runs many ratchet guards that compare the
current branch against `origin/main`. Before this PR, those guards repeatedly
ran `git fetch --depth=1 origin main` inline before individual scripts.

That kept the guards order-independent, but it added redundant network work and
left the workflow with repeated boilerplate. The same job also still passed the
removed `build_runner --delete-conflicting-outputs` option, which build_runner
2.15.0 ignores with a warning.

Scope is deliberately GitHub Actions only. Codemagic workflows are manual
release/store builds, so they are left untouched.

Relates to #4339

## What changed

- Fetch `origin/main` once at the start of the `generated-files` job.
- Remove the repeated per-ratchet fetch calls.
- Update comments that referenced the old inline fetch behavior.
- Normalize single-command ratchet `run` blocks touched by the fetch cleanup.
- Drop the ignored `--delete-conflicting-outputs` flag from the build_runner
  verification step.

## Correctness notes

- No dependency cache behavior changes. `subosito/flutter-action@v2` with
  `cache: true` remains the only pub-cache setup.
- No build_runner state is cached. The generated-files guard still runs from a
  clean build graph and checks `git status --porcelain` for generated-file drift.
- The ratchet scripts still compare against `origin/main`; the single up-front
  fetch provides that ref before any ratchet runs. The guards still fail closed
  if the base baseline cannot be loaded.

### DIFF
--- a/.github/workflows/mobile_ci.yaml
+++ b/.github/workflows/mobile_ci.yaml
@@ -78,6 +78,11 @@ jobs:
         working-directory: mobile/
     steps:
       - uses: actions/checkout@v4
+      - name: 🔀 Fetch origin/main for ratchet baselines
+        # The ratchet scripts below compare against origin/main. Fetch it once
+        # here so each script is order-independent; a fetch failure fails the
+        # job (the guards fail closed when the base baseline can't load).
+        run: git fetch --quiet --depth=1 origin main
       - name: 🐦 Setup Flutter
         uses: subosito/flutter-action@v2
         with:
@@ -110,7 +115,6 @@ jobs:
           # is the primary guard; this freezes raw setMockMethodCallHandler
           # installs of the five shared channels so new ones must go through
           # overrideSharedChannel(...). #5738.
-          git fetch --quiet --depth=1 origin main
           bash scripts/check_shared_channel_overrides.sh
       - name: 📦 Verify no new package-test channel leaks (per-package merged-isolate ratchet)
         run: |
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz (approved):
The takeover changes remove the duplicate cache restores and avoid weakening the generated-files guard while keeping the valid fetch consolidation. CI is green, the PR body/title now match the diff, and the issue link gate is clear.

NotThatKindOfDrLiz:
@rabble I took over the workflow cleanup and pushed `ce09a47e2a`.

What changed:
- Removed the added pub-cache steps because `subosito/flutter-action@v2` already caches pub deps when `cache: true` is set.
- Dropped the build_runner cache so the generated-files guard keeps its current cold-graph verification behavior.
- Kept the useful part: one upfront `origin/main` fetch for all generated-files ratchets.
- Removed the ignored `--delete-conflicting-outputs` flag and cleaned up the touched single-command ratchet blocks.
- Retitled and rewrote the PR body to match the final diff, and linked `Relates to #4339`.

Verification:
- Local: `git diff --check`
- Local: YAML parse via Ruby `YAML.load_file`
- GitHub: Mobile CI, mergeability, semantic PR, linked-issue, and CLA are all green.

Please review the takeover changes and merge when ready.

PR #6353, opened 2026-07-24
fix(feed): guard tuning-undo against unmounted context

### LINKED ISSUE #6354: fix(feed): guard tuning-undo against unmounted context
Auto-created from PR #6353.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6353
**Branch:** `fix/feed-tuning-undo-crash`

hm21:
Closes #6354

## Description

Fixes a FATAL Crashlytics crash (`6f6338d4`, iOS, v1.0.16–1.0.17, still OPEN): `Null check operator used on a null value` in `_FullscreenFeedContentState._undoTuning`, blamed on `State.context` (`framework.dart:959`).

The feed-tuning **Undo** receipt is shown through the app-level `ScaffoldMessenger`, which lives above the feed route and outlives it. When the user navigates away, the feed's `State` is disposed but the receipt stays on screen for its 4-second window. Tapping **Undo** then runs `_undoTuning` → `context.read<...>()` on the unmounted `State` (`_element!` is null) → crash. The v1.0.16 variant is the same crash via the old `SnackBarAction` path, so it is not tied to the recent snackbar-rendering change.

The fix is a `mounted` guard before touching `context`, matching the pattern already used elsewhere in the same file (`_handlePendingSkip`, `_syncFeedActive`). The home feed's `_undoTuning` shares the identical latent bug — the same shared, messenger-scoped receipt — so it gets the same guard.

**Related Issue:** 

## Out of Scope

Making the undo actually fire after the screen is gone (the feed BLoC is scoped to the screen and is closed on dispose, so there is nothing to dispatch to). A silent no-op is the correct behavior here; lifting undo to a screen-independent layer is a separate, larger change.

## Verification

- `flutter analyze lib test integration_test` — clean on all four files.
- `flutter test test/screens/feed/pooled_fullscreen_video_feed_screen_test.dart test/screens/feed/video_feed_page_test.dart` — green.
- Added one regression test per screen (unmount the feed while the receipt is visible, then tap Undo → no exception). Confirmed both **fail without the guard** and pass with it.

## Type of Change

### DIFF
--- a/mobile/lib/screens/feed/pooled_fullscreen_video_feed_screen.dart
+++ b/mobile/lib/screens/feed/pooled_fullscreen_video_feed_screen.dart
@@ -542,6 +542,10 @@ class _FullscreenFeedContentState extends ConsumerState<FullscreenFeedContent>
   }
 
   void _undoTuning(String feedTuningEventId, String videoId) {
+    // The Undo receipt lives in the app-level ScaffoldMessenger, so it can
+    // fire after this screen was popped and its State unmounted — touching
+    // `context` then throws a null-check on the dead element. Bail if gone.
+    if (!mounted) return;
     final bloc = context.read<FullscreenFeedBloc>()
       ..add(FullscreenFeedTuningUndoRequested(feedTuningEventId));
     final videos = bloc.state.videos;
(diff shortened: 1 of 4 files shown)


dcadenas (approved):
✅ tuning undo stays safe after dispose

Issue #4974, opened 2026-06-08
fix(dm): read + publish the user's own kind-10050 DM inbox relays (NIP-17 receive side)


realmeylisdev:
### Summary

divine-mobile is **NIP-17 compliant on the send side but not the receive side**. This is the follow-up to #4953 (fixed by #4973) — that PR recovers DM history after reinstall by draining the default/full relay pool, which is the correct fix for users who haven't advertised a kind-10050. This issue covers the remaining receive-side spec gap.

### Current behavior (verified in code)

- **Sending** correctly routes gift wraps to the *recipient's* kind-10050 DM inbox relays: `DmRepository.resolveDmInboxRelays(recipientPubkey)` → `sendMessage` (`dm_repository.dart:863`, used at `:968`). Landed in #570 (commit `4161e95ff`).
- **Receiving** ignores the user's *own* kind-10050 entirely: `startListening()` subscribes with no relay targeting (`dm_repository.dart` subscribe call), reading only the default pool. `resolveDmInboxRelays` is never called for `_userPubkey`.
- **Publishing**: divine-mobile **never publishes** a kind-10050 for the user at all (deliberately deferred — `mobile/docs/plans/2026-01-06-nip17-dm-feature-design.md:283`).

Consequence: if a user advertises a kind-10050 with relays outside divine-mobile's default pool (e.g. via another client), NIP-17-compliant senders deliver gift wraps to those inbox relays, and divine-mobile never reads them → those conversations are silently missing. divine-web already does the right thing: it includes the current user's kind-10050 relays in its DM read set (`divine-web/src/lib/dm.ts:614-638`, `resolveDmReadRelays`).

### Proposed work

1. **Read side (RC2):** resolve the user's own kind-10050 inbox relays and include them (in addition to the default pool) as `targetRelays` on the live gift-wrap subscription (`NostrClient.subscribe(... targetRelays:)`) and on the #4953 history drain (`queryEvents(... tempRelays:)`). Timeout-guarded with fallback to the default pool when absent (spec: "not ready to receive" → default pool).
2. **Publish side (RC3):** publish a minimal kind-10050 for the user on login when one is missing, so compliant senders deliver where divine-mobile actually reads.
3. **Optional RC4 hardening:** re-open / refresh the live DM subscription when `addRelays` expands the pool mid-session (covers steady-state relay drift beyond the reinstall case).

### Notes / why this is separate from #4973

[...]

PR #6368, opened 2026-07-24
fix(dm): harden NIP-17 against unauthenticated rumor fields and unsigned seals

realmeylisdev:
## Description

A NIP-59 inner rumor is **unsigned**. Only the kind-13 seal and the kind-1059 wrap are cryptographically bound, so receiver code cannot trust rumor fields such as `id`, `pubkey`, or `created_at`.

This PR hardens the NIP-17 DM path against three silent failure modes:

### 1. Future-dated rumors could poison DM sync cursors

`recordSeen` advanced `newestSyncedAt` to any rumor `created_at`, and `startListening` derives the live subscription's `since:` from that value. One DM stamped far in the future could move the cursor past now and make later subscriptions return nothing.

The fix now bounds both sync boundaries independently before persisting:

- `newestSyncedAt` is clamped to the plausible range instead of accepting future values.
- `oldestSyncedAt` is clamped to the same plausible range so a fresh install cannot seed history drain with epoch or far-future values.
- `repairPoisonedBoundaries` heals both poisoned boundaries.
- When repair re-arms the history drain, it seeds `historyDrainCursor` to now so the one-time drain walks down through the missed window instead of jumping straight to already-synced old history.

Future-dated NIP-17 rumors are still delivered. The repository clamps the timestamp used for local persistence, conversation ordering, dedup windows, and sync cursor updates instead of dropping and ledgering the wrap. A warning is logged when the rumor is beyond the expected 24-hour skew, but the message is preserved.

### 2. Attacker-chosen rumor fields became local identity

`Event.fromJson` copies `data['id']` and `data['pubkey']` through verbatim. For unsigned rumors, that let unauthenticated input become the direct-message primary key, reaction upsert key, and displayed sender.

Both local unwrap paths now rebuild the inner rumor unconditionally from the authenticated seal pubkey and the canonical rumor body fields. The shared rebuild helper also parses only `kind`, `tags`, `content`, and `created_at`, so legitimate senders that omit unauthenticated `id` or `pubkey` no longer fail unwrap.

### 3. Unsigned seals could report as delivered

`Nostr.signEvent` can return null when a remote signer declines or times out. The gift-wrap builder now refuses to build the kind-1059 wrap unless the kind-13 seal has a signature, so the send stays retryable instead of publishing a wrapper that recipients will drop.

**Related Issue:** Relates to #4974

## Out of Scope

Other findings from the same audit, deliberately not included to keep this reviewable:

- kind-10050 DM relay list is still not advertised (backend-gated on the relay accepting kind 10050, #4974)
- "delete for everyone" still publishes a plaintext kind-5 naming both parties
- the first message of a new thread still publishes a plaintext NIP-04 kind-4 copy
- reactions still publish to the default pool while OK-confirming
- `stopListening()` still does not stop an in-flight history drain
- the Keycast server-batch unwrap path still performs no client-side verification

## Verification

Original author verification:

- `dm_repository`: package tests passed with coverage above gate
- `nostr_sdk`: package tests passed
- `db_client`: package tests passed
- `test/blocs/dm`: tests passed
- `flutter analyze lib test`: no issues

Takeover verification on the final diff:

- `mobile/packages/dm_repository`: `flutter analyze lib test`
- `mobile/packages/dm_repository`: `flutter test` (534 tests)
- `mobile/packages/nostr_sdk`: `flutter analyze lib test`
- `mobile/packages/nostr_sdk`: `flutter test` (396 tests)
- pre-push hook: no merge conflicts with `main`, analyzer passed, changed-file tests passed (448 tests)
- `git diff --check`

No generated-code inputs touched, no ARB changes, no golden changes.

## Type of Change

### DIFF
--- a/mobile/packages/dm_repository/lib/src/dm_decryption_worker.dart
+++ b/mobile/packages/dm_repository/lib/src/dm_decryption_worker.dart
@@ -94,7 +94,7 @@ Future<DecryptedRumorResult> _decryptOne(
   }
 
   // C16/NIP-44: validate the outer gift wrap (kind 1059) signature before
-  // decrypting. Defense-in-depth — diVine relays verify on publish, but an
+  // decrypting. Defense-in-depth — Divine relays verify on publish, but an
   // untrusted relay or local cache could serve a forged/tampered wrap.
   if (!verifyGiftWrapPart(giftWrap)) {
     return DecryptedRumorResult.failure(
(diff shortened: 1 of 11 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I pushed takeover commit 754a613a78 to close the review gaps from the second pass. Summary:

- `repairPoisonedBoundaries` now seeds the history drain cursor to now, so repaired installs re-drain through the missed window instead of jumping to already-synced old history.
- Future-dated NIP-17 rumors are preserved with a clamped local timestamp instead of being dropped/ledgered; the skew threshold is now warning-only at 24h.
- `recordSeen` bounds both newest and oldest on both sides, and repair heals both poisoned corners.
- Gift-wrap unwrap now rebuilds unsigned rumors from parsed canonical fields and the authenticated seal pubkey, so omitted unauthenticated `id`/`pubkey` no longer break local unwrap.
- Stale comments/body text were updated to match the new behavior.

Local verification plus pre-push passed, and GitHub checks are green. Please review the takeover commit and merge when ready.

NotThatKindOfDrLiz (approved):
Final pass is clean after takeover commit 754a613a78. The previous blockers are addressed, local verification passed, and GitHub checks are green.

PR #6372, opened 2026-07-25
perf(ci): cut 23s of full-history clone and 42s of apt off Mobile CI

### LINKED ISSUE #6374: perf(ci): cut 23s of full-history clone and 42s of apt off Mobile CI
Auto-created from PR #6372.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6372
**Branch:** `perf/ci-and-test-speed`

rabble:
Closes #6374

## Why

First PR of a measured pass over CI / test / dev-loop wall clock. It adds the baseline everything else is measured against (`docs/PERF_BASELINE.md`) and takes two pieces of pure waste that measurement turned up in `mobile_ci.yaml`.

Numbers come from run 30142684388 plus p50/p90 over the 13 most recent successful `Mobile CI` runs. Queue wait is 3–4s across every job, so CI is not queue-bound — all of this is real work.

## What changed

**1. `Detect App CI Scope` no longer clones the repo — 26s → 5s (measured on this PR)**

The job existed to run `git diff --name-only` between two SHAs, and paid `actions/checkout@v4` with `fetch-depth: 0` to do it:

| Step | Time |
|---|---:|
| `actions/checkout@v4` (full history) | **23s** |
| Detect whether app CI is required | 0s |

The REST API already has that list, so the job now runs with no checkout at all. Every other job declares `needs: changes`, so this sits at the head of the workflow's critical path on every single run.

**2. The transport-security guard is gated on native changes — saves ~42s when it doesn't apply**

`check_native_transport_security.sh` reads exactly one Android XML plus three `Info.plist` files. If none of them (nor the script) changed, its verdict is identical to the one `main` already recorded. The step is now gated on a new `native` output from the `changes` job, matched at directory level (`mobile/android/**`, `mobile/ios/**`, `mobile/macos/**`) so it is a superset of the guard's real inputs and can't drift if the guard gains a fourth plist.

The 43s step is almost entirely toolchain install — `xmllint` genuinely isn't on the `ubuntu-24.04` image:

```
03:41:45  sudo apt-get update -qq            # 28s
03:42:13  apt-get install libxml2-utils      # 14s
03:42:27  (guard itself runs in <1s)
```

## Measured impact

| | Before (p50) | After |
|---|---:|---:|
| `Detect App CI Scope` | 26s | **5s** (measured, this PR) |
| `Generated Files` (non-native PR) | 198s | ~156s (projected) |
| **Mobile CI critical path** | **625s** | **~604s** |

The critical-path saving is only the ~21s today, because `Tests` (599s p50) dwarfs everything else and runs in parallel with the rest. The `Generated Files` saving is runner-minutes now and becomes a critical-path saving as soon as `Tests` drops below it — which is what the sharding work does next.

Note this PR touches `.github/workflows/**`, which sets both `app=true` and `native=true`, so the transport guard still runs here and its skip path is not exercised by this PR's own run.

## A third change was attempted and reverted

I also removed `flutter pub get` from the `Format` job on the theory that `dart format` is purely syntactic. **CI proved that wrong on the first run** and the second commit restores it.

`dart format` reads its page width from `analysis_options.yaml`, which `include`s `package:very_good_analysis/analysis_options.yaml`. With no resolved package config the include fails —

```
Failed to resolve package URI "package:very_good_analysis/analysis_options.yaml"
...
Formatted 2708 files (529 changed) in 12.42 seconds.
```

— and the formatter silently falls back to its defaults. The sandbox I verified against had no `analysis_options` include, so it never reproduced this. Leaving the failure and its cause in the history rather than force-pushing it away.

## Correctness

No check is removed or weakened.

- The changed-file list is the same list, from a different source. The `pulls/{n}/files` endpoint caps at 3000 entries, so the job first reads `changed_files` and **falls open to the full matrix** above that cap rather than under-running CI on a truncated list.
- The transport guard still runs in full on every push to `main`, on every non-PR event, and on any PR touching native config or the guard itself. Skipping it when its inputs are byte-identical to `main` does not change what is enforced.

## Verification

`.github/workflows/mobile_ci.yaml` parses and the gate lands on the right step:

```
jobs: ['changes', 'generated-files', 'format', 'analyze', 'tests', 'vgv-tag-gate', 'mobile-ci']
changes outputs: {'app': ..., 'native': ...}
step[3]: 🔒 Verify native transport security | if: ${{ github.event_name != 'pull_request' || needs.changes.outputs.native == 'true' }}
```

### DIFF
--- a/.github/workflows/mobile_ci.yaml
+++ b/.github/workflows/mobile_ci.yaml
@@ -273,8 +273,12 @@ jobs:
           flutter-version: "3.44.0"
           channel: "stable"
           cache: true
-      # No `flutter pub get`: `dart format` is purely syntactic and never
-      # resolves package imports, so the ~10s resolve was dead weight.
+      - name: 📦 Install dependencies
+        # Required: `dart format` reads page width from analysis_options.yaml,
+        # which `include`s package:very_good_analysis. Without a resolved
+        # package config that include fails and the formatter silently falls
+        # back to its defaults, reformatting 529 files.
+        run: flutter pub get
       - name: ✨ Check formatting
         run: dart format --output=none --set-exit-if-changed lib test integration_test
   analyze:
--- a/.github/workflows/mobile_ci.yaml
+++ b/.github/workflows/mobile_ci.yaml
@@ -15,6 +15,9 @@ jobs:
   changes:
     name: Detect App CI Scope
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      pull-requests: read
     outputs:
       app: ${{ steps.scope.outputs.app }}
       native: ${{ steps.scope.outputs.native }}
@@ -47,7 +50,7 @@ jobs:
             exit 0
           fi
 
-          gh api --paginate "/repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files" \
+          gh api --method GET --paginate -F per_page=100 "/repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files" \
             --jq '.[].filename' > /tmp/mobile-ci-changed-files.txt
           echo "Changed files:"
           cat /tmp/mobile-ci-changed-files.txt
--- a/.github/workflows/mobile_ci.yaml
+++ b/.github/workflows/mobile_ci.yaml
@@ -17,31 +17,48 @@ jobs:
     runs-on: ubuntu-24.04
     outputs:
       app: ${{ steps.scope.outputs.app }}
+      native: ${{ steps.scope.outputs.native }}
     steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
+      # No checkout: the changed-file list comes from the REST API. A
+      # `fetch-depth: 0` clone cost ~23s of this workflow's critical path
+      # (every other job `needs: changes`) to produce a list the API already
+      # has.
       - name: Detect whether app CI is required
         id: scope
         env:
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
-          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
         run: |
           if [ "${GITHUB_EVENT_NAME}" != "pull_request" ]; then
             echo "app=true" >> "$GITHUB_OUTPUT"
+            echo "native=true" >> "$GITHUB_OUTPUT"
             echo "Non-PR event; running full app CI."
             exit 0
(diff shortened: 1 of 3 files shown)


rabble:
All checks green. Measured on [run 30144667239](https://github.com/divinevideo/divine-mobile/actions/runs/30144667239):

| Job | Baseline (p50, 13 recent runs) | This PR | Δ |
|---|---:|---:|---:|
| `Detect App CI Scope` | 26s | **4s** | **−22s** |
| `Format` | 52s | 48s | — (restored to baseline behaviour) |
| `Generated Files` | 198s | 199s | — (see below) |

`Detect App CI Scope` is the measured win, and because every other job declares `needs: changes` it comes straight off the critical path on every run.

`Generated Files` is unchanged **here by design**: this PR edits `.github/workflows/**`, which sets `native=true`, so the transport-security step (and its ~42s apt install) still runs on this PR's own run. The skip only fires on PRs that touch no native config — which is most of them.

NotThatKindOfDrLiz:
@rabble I took over the review fixes and pushed them to `perf/ci-and-test-speed`.

What changed:
- rebased the PR onto current `origin/main`, preserving the `pro_image_editor` 13.2.3 bump from #6382
- added explicit `contents: read` / `pull-requests: read` permissions to the Mobile CI scope job
- changed the PR files API call to request `per_page=100` with `--method GET`
- moved the performance baseline to `docs/PERF_BASELINE.md`
- fixed the baseline doc package/workflow counts and replaced the nonexistent `scripts/ci/select_test_shard.sh` reproduction command
- updated the PR body to point at the new docs path

Verification:
- `git diff --check origin/main...HEAD`
- `.github/workflows/mobile_ci.yaml` parses with Ruby YAML
- `gh api --method GET --paginate -F per_page=100 /repos/divinevideo/divine-mobile/pulls/6372/files --jq '.[].filename'`
- GitHub checks are green on `c1df4c3f18`: Mobile CI, Tests, Generated Files, Analyze, Format, VGV Tag Gate, PR Issue Link, mergeability, semantic PR, CLA

NotThatKindOfDrLiz (approved):
Approved after takeover. The workflow hardening and baseline doc fixes address the review findings, the branch is current with main, and all checks are green on c1df4c3f18.

realmeylisdev (approved):
Reviewed the scope-detection changes independently. The GH API method fix (`--method GET` alongside `-F per_page=100`) and the added job permissions are real hardening, not cosmetic — without `--method GET`, `-F` would have silently switched the files-list call to a POST against a GET-only endpoint. Confirmed the case-statement logic: every path that sets `native=true` also sets `app=true`, so the transport-security check can't be skipped while its parent job is skipped, and the >3000-file cap falls open to full CI rather than truncating.

PR #6378, opened 2026-07-25
feat(supporter): add flag-gated monthly supporter subscription (Option A)

### LINKED ISSUE #6379: feat(supporter): add flag-gated monthly supporter subscription (Option A)
Auto-created from PR #6378.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6378
**Branch:** `feat/divine-supporters`

rabble:
Closes #6379

## What

Adds an optional, **Signal-style monthly supporter subscription** behind a new `FeatureFlag.divineSupporters` (default **off**). Nothing is gated — it's recognition + funding only.

This is **Option A** (native IAP) from the supporter exploration, implemented **client-first**: on-device StoreKit / Play Billing validation today, with a clean `EntitlementValidator` interface so a server-side receipt-validation backend can slot in later without touching the UI or BLoC.

## Architecture (follows UI → Cubit → Repository → Client)

| Layer | File |
|---|---|
| Feature flag | `FeatureFlag.divineSupporters` + both `build_configuration` switches |
| Models | `SupporterTier`, `SupporterEntitlement` (in `packages/models`) |
| Store abstraction | **new** `packages/iap_repository`: `EntitlementValidator` interface, `InAppPurchaseValidator`, `StubEntitlementValidator`, typed exceptions |
| Service | `SupporterRepository` (cached entitlement in SharedPreferences) |
| Providers | `supporter_providers.dart` (`@riverpod`, selects validator by platform) |
| BLoC | `SupporterCubit` + `SupporterState` (Equatable, status + typed-failure enums) |
| UI | `SupporterScreen` + route + settings tile, all flag-gated |

## Tests
- `iap_repository` package: **26 tests, 96% coverage** (floor locked at 90)
- Models, repository, cubit, widget: **29 tests, all green**
- Pre-push hook (full analyze + suite) passed

## Manual prerequisites before flipping the flag on in a release
- **App Store Connect + Google Play Console**: create the `divine.supporter.monthly` product (SKU) in both stores.
- **iOS capability**: add the *In-App Purchase* capability to the `Runner` target (writes `com.apple.developer.in-app-purchases.payments` to `Runner.entitlements`) and refresh the signing profile. Intentionally **not** done in this PR — adding it now would break signing for everyone until the App ID is updated in the Apple Developer Portal. The Flutter plugin + Dart code work without it; only real purchases fail gracefully.
- **Android**: no manifest permission needed for Play Billing.

## Explicitly out of scope (follow-ups)
- Server-side receipt validation backend (the `EntitlementValidator` interface is designed for it; not built).
- NIP-58 "Divine Supporter" badge (deferred per the chosen direction; local flag ships first).
- Content/feature gating — none, by design (Signal-style).

## Risks
- Client-only entitlement for the MVP means no cross-device sync until the backend lands; restore relies on the store account.
- This is a business-model shift (currently documented as "funded by andotherstuff.org"). The flag ships off so it's pure exploration until product/release sign-off.

### DIFF
--- a/mobile/lib/blocs/supporter/supporter_cubit.dart
+++ b/mobile/lib/blocs/supporter/supporter_cubit.dart
@@ -1,0 +1,128 @@
+// ABOUTME: Screen-scoped Cubit for the Divine supporter screen.
+// ABOUTME: Loads tiers, drives subscribe/restore, and maps store exceptions.
+
+import 'dart:async';
+
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:iap_repository/iap_repository.dart';
+import 'package:models/models.dart';
+import 'package:openvine/blocs/supporter/supporter_state.dart';
+import 'package:openvine/services/supporter_repository.dart';
+
+typedef SupporterAnalyticsSink = void Function(String event);
+
+class SupporterCubit extends Cubit<SupporterState> {
+  /// Creates a [SupporterCubit].
+  ///
+  /// [repository] owns the validator and cached entitlement.
+  /// [trackEvent] is an optional analytics sink (e.g.
+  /// `analyticsEventSinkProvider`); pass `_noopAnalytics` when unavailable.
+  SupporterCubit({
+    required SupporterRepository repository,
+    SupporterAnalyticsSink trackEvent = _noopAnalytics,
+  }) : _repository = repository,
+       _trackEvent = trackEvent,
+       super(SupporterState(entitlement: repository.current));
+
+  final SupporterRepository _repository;
+  final SupporterAnalyticsSink _trackEvent;
+
+  StreamSubscription<SupporterEntitlement>? _entitlementSub;
+
+  /// Begin listening to the repository's entitlement stream. Call from the
+  /// screen's `initState` so external purchase updates (renewals, restores)
+  /// reflect in the UI.
+  void start() {
+    _entitlementSub ??= _repository.changes.listen((entitlement) {
+      emit(
+        state.copyWith(
+          entitlement: entitlement,
+          status: entitlement.isSupporter
+              ? SupporterStatus.active
+              : state.status,
+          clearFailure: entitlement.isSupporter,
+        ),
+      );
+    });
+    loadTiers();
+  }
+
+  /// Fetch the available supporter tiers from the store.
+  Future<void> loadTiers() async {
+    emit(state.copyWith(status: SupporterStatus.loading, clearFailure: true));
+    try {
+      final tiers = await _repository.validator.fetchProducts();
+      emit(state.copyWith(tiers: tiers, status: SupporterStatus.idle));
+    } on EntitlementException catch (e) {
+      emit(
+        state.copyWith(
+          status: SupporterStatus.error,
+          failure: SupporterFailure.fromMessage(e.message),
+        ),
+      );
+    }
+  }
+
+  /// Begin a purchase for [productId].
+  Future<void> subscribe(String productId) async {
+    if (state.isBusy) return;
+    emit(
+      state.copyWith(status: SupporterStatus.purchasing, clearFailure: true),
+    );
+    _trackEvent('supporter_subscribe_tapped');
+    try {
+      final entitlement = await _repository.validator.purchase(productId);
+      emit(
+        state.copyWith(
+          entitlement: entitlement,
+          status: SupporterStatus.active,
+          clearFailure: true,
+        ),
+      );
+      _trackEvent('supporter_subscribe_succeeded');
+    } on EntitlementException catch (e) {
+      emit(
+        state.copyWith(
+          status: SupporterStatus.idle,
+          failure: SupporterFailure.fromMessage(e.message),
+        ),
+      );
+      _trackEvent('supporter_subscribe_failed');
+    }
+  }
+
+  /// Restore previous purchases tied to the store account.
+  Future<void> restore() async {
+    if (state.isBusy) return;
+    emit(state.copyWith(status: SupporterStatus.restoring, clearFailure: true));
+    _trackEvent('supporter_restore_tapped');
+    try {
+      await _repository.validator.restorePurchases();
+      // The restored entitlement arrives on the repository stream; reset to idle
+      // and let the stream listener surface the active status.
+      emit(state.copyWith(status: SupporterStatus.idle));
+      _trackEvent('supporter_restore_completed');
+    } on EntitlementException catch (e) {
+      emit(
+        state.copyWith(
+          status: SupporterStatus.idle,
+          failure: SupporterFailure.fromMessage(e.message),
+        ),
+      );
+      _trackEvent('supporter_restore_failed');
+    }
+  }
+
+  /// Dismiss the current failure banner.
+  void dismissError() {
+    emit(state.copyWith(clearFailure: true));
+  }
+
+  @override
+  Future<void> close() {
+    _entitlementSub?.cancel();
+    return super.close();
+  }
+
+  static void _noopAnalytics(String _) {}
+}
--- a/mobile/lib/blocs/supporter/supporter_state.dart
+++ b/mobile/lib/blocs/supporter/supporter_state.dart
@@ -1,0 +1,71 @@
+// ABOUTME: State for the Divine supporter screen.
+// ABOUTME: Tracks tiers, current entitlement, and purchase/restore lifecycle.
+
+import 'package:equatable/equatable.dart';
+import 'package:models/models.dart';
+
+enum SupporterStatus { idle, loading, purchasing, restoring, active, error }
+
+/// Typed reason for a supporter-screen failure, mapped from validator
+/// exceptions so the UI can show specific copy.
+enum SupporterFailure {
+  storeUnavailable,
+  purchaseFailed,
+  purchasePending,
+  restoreFailed,
+  unknown;
+
+  static SupporterFailure fromMessage(String message) {
+    final lower = message.toLowerCase();
+    if (lower.contains('unavailable')) return SupporterFailure.storeUnavailable;
+    if (lower.contains('pending')) return SupporterFailure.purchasePending;
+    if (lower.contains('restore')) return SupporterFailure.restoreFailed;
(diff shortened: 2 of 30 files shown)


mobile/lib/providers/supporter_providers.dart
```diff
@@ -0,0 +1,46 @@
...
+///
+/// Returns an [InAppPurchaseValidator] on iOS/Android and a
+/// [StubEntitlementValidator] elsewhere so the rest of the app can treat the
+/// supporter feature uniformly.
+@Riverpod(keepAlive: true)
+EntitlementValidator entitlementValidator(Ref ref) {
+  if (!hasInAppPurchaseStore) {
+    return StubEntitlementValidator();
+  }
+  final validator = InAppPurchaseValidator();
```
NotThatKindOfDrLiz, line 32: This constructs the store validator lazily through a provider that is itself only read by the supporter screen. The in_app_purchase purchase stream needs to be listened to as early as possible on app startup so pending/restored purchases from a previous session are redelivered and acknowledged. With the current wiring, an Android pending purchase approved after app restart can be delivered to nobody until the screen opens, leaving `completePurchase` uncalled and risking auto-refund/revocation.

mobile/packages/iap_repository/lib/src/in_app_purchase_validator.dart
```diff
@@ -0,0 +1,212 @@
...
+        _entitlementController.add(entitlement);
+        await _completeSafely(purchase);
+        completer?.complete(entitlement);
+      case PurchaseStatus.error:
+        final exception = PurchaseFailedException(
+          purchase.error?.code,
+          purchase.error?.message ?? 'Purchase failed.',
+        );
+        completer?.completeError(exception);
+        _entitlementController.addError(exception);
```
NotThatKindOfDrLiz, line 80: Please do not add purchase failures to `entitlementChanges` unless every production subscriber handles stream errors. `SupporterRepository` currently listens with no `onError`, so this turns a handled declined/canceled purchase into an uncaught async error in addition to the cubit's handled `PurchaseFailedException` path.

mobile/lib/blocs/supporter/supporter_cubit.dart
```diff
@@ -0,0 +1,128 @@
...
+
+  /// Begin a purchase for [productId].
+  Future<void> subscribe(String productId) async {
+    if (state.isBusy) return;
+    emit(
+      state.copyWith(status: SupporterStatus.purchasing, clearFailure: true),
+    );
+    _trackEvent('supporter_subscribe_tapped');
+    try {
+      final entitlement = await _repository.validator.purchase(productId);
```
NotThatKindOfDrLiz, line 74: This await can outlive the route. If the user pops the screen while the store sheet is active, BlocProvider closes the cubit and the later `emit` throws after close. Guard the continuation (`isClosed` or an equivalent cancellation/lifecycle pattern) in both `subscribe()` and `restore()`, and add a regression test for leaving mid-purchase/restore.

mobile/packages/iap_repository/lib/src/in_app_purchase_validator.dart
```diff
@@ -0,0 +1,212 @@
...
+          'Purchase was cancelled.',
+        );
+        completer?.completeError(exception);
+      case PurchaseStatus.pending:
+        // Still pending parental approval / payment settlement; keep waiting
+        // for the same completer on the next stream event.
+        if (completer != null) {
+          _pendingPurchases[purchase.productID] = completer;
+        }
+        _entitlementController.add(SupporterEntitlement.inactive);
```
NotThatKindOfDrLiz, line 93: A pending purchase should not emit `SupporterEntitlement.inactive`. The repository persists every entitlement change, so an existing supporter who starts a renewal/second purchase that goes pending gets their active cached entitlement cleared on disk and in memory. Keep pending as a non-entitlement state or preserve the previous entitlement until a terminal store result arrives.

mobile/lib/blocs/supporter/supporter_state.dart
```diff
@@ -0,0 +1,71 @@
...
+/// Typed reason for a supporter-screen failure, mapped from validator
+/// exceptions so the UI can show specific copy.
+enum SupporterFailure {
+  storeUnavailable,
+  purchaseFailed,
+  purchasePending,
+  restoreFailed,
+  unknown;
+
+  static SupporterFailure fromMessage(String message) {
```
NotThatKindOfDrLiz, line 18: This loses the sealed exception type and rebuilds it from arbitrary store prose. Store messages can be localized or contain overlapping words like `unavailable`, which maps failures incorrectly. Keep the typed `EntitlementException` and switch on `StoreUnavailableException`, `PurchaseFailedException`, etc. rather than parsing `message`.

mobile/lib/screens/settings/supporter_screen.dart
```diff
@@ -0,0 +1,261 @@
...
+    context.read<SupporterCubit>().start();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocBuilder<SupporterCubit, SupporterState>(
+      builder: (context, state) {
+        return Scaffold(
+          appBar: DiVineAppBar(
+            title: 'Divine Supporters',
```
NotThatKindOfDrLiz, line 52: This new screen and the settings tile are hardcoded English. Repo rules require all user-facing strings to go through `context.l10n`, and the widget tests should resolve `AppLocalizations` keys instead of asserting English literals. Please add ARB keys/generated l10n for the full supporter surface before this ships.

mobile/lib/screens/settings/supporter_screen.dart
```diff
@@ -0,0 +1,261 @@
...
+class _Hero extends StatelessWidget {
+  const _Hero({required this.state});
+
+  final SupporterState state;
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      children: [
+        const Icon(
```
NotThatKindOfDrLiz, line 104: This adds raw Material `Icons.*`, which currently fails the raw-icons ratchet in Generated Files. Use `DivineIcon`/`DivineIconButton` with the appropriate `DivineIconName` values, and update the other raw icons in this screen at the same time.

NotThatKindOfDrLiz (changes requested):
Requesting changes. This is still draft and CI is red, but the larger blockers are in the purchase lifecycle and new UI surface. Before this can move forward, please fix the store listener startup behavior, repository stream error handling, cubit close/dispose paths, pending entitlement handling, l10n/design-system violations, and the red Format / Generated Files / IAP package analyze checks. The feature flag limits current exposure, but the code is intended to be flipped later and these are money-path defects.

rabble (commented):
Draft review findings:

**Security note (by design, but needs a hard gate):** entitlement is granted entirely client-side and persisted as plain JSON in SharedPreferences (`divine_supporter_entitlement`). Any rooted/jailbroken device (or a user who just edits the prefs file / restores a backup) can self-grant supporter status — no receipt, token, or signature is ever checked or stored. Acceptable for a flag-off, recognition-only MVP, but the flag must never be flipped for anything with real value (badges others see, features, cross-device sync) until PR #6413's server verification lands. Related: `in_app_purchase_validator.dart` discards `purchase.verificationData` and calls `completePurchase` immediately (`:97-103`), so when the server validator slots in later there is no stored receipt/token to verify retroactively — capture and persist the verification data now if you want a migration path.

**PR relationship:** #6413 (`feat(supporter): enforce server-verified subscriptions`) touches a strict superset of this PR's files (same flag, same package, same screen, plus `supporter_api_client` + server claim flow). Both target main. Merging #6378 then #6413 guarantees conflicts on every shared file; merging #6413 first makes #6378 redundant. Recommend closing this draft in favor of #6413, or rebasing #6413 onto this and treating #6378 as the base commit only.

mobile/packages/iap_repository/lib/src/in_app_purchase_validator.dart:114: 🔴 bug: `_entitlementFromPurchase` never sets `expirationDate` and defaults `isActive: true`; cached entitlement in SharedPreferences therefore never expires — a cancelled monthly subscription stays "supporter" forever locally. Either poll the store on launch to revalidate or store an expiry.
mobile/packages/iap_repository/lib/src/in_app_purchase_validator.dart:93: 🟡 risk: `PurchaseStatus.pending` emits `SupporterEntitlement.inactive` to the entitlement stream, which the repository persists — a renewal going through payment settlement wipes an active cached entitlement (badge flickers off/on). Don't emit inactive for pending; only complete the pending completer.
mobile/lib/services/supporter_repository.dart:29: 🟡 risk: `entitlementChanges.listen(_handleChange)` has no `onError`, but the validator calls `_entitlementController.addError(...)` on purchase errors (validator `:85`) — unhandled async stream error. Add an error handler (log + ignore, or map to failure state).
mobile/lib/screens/settings/supporter_screen.dart:105: 🟡 risk: raw `Icons.*` (`:105`,`:141`,`:224`,`:228`) in a new file — CI "Generated Files" job already fails on the raw-Icons ratchet. Use `DivineIcon`/`DivineIconButton`.
mobile/lib/screens/settings/supporter_screen.dart:205: 🟡 risk: raw `TextButton` (ButtonStyleButton family) in a new file trips the material-button ceiling. Use `DivineButton` variant.
mobile/lib/screens/settings/supporter_screen.dart:52: 🟡 risk: all user-facing copy hardcoded English (here, settings tile `settings_screen.dart:250-253`, `_FailureBanner`, `_Disclaimer`) with no `app_*.arb` keys — violates l10n mirror policy / check-l10n.
mobile/packages/iap_repository/test/exceptions_test.dart:1: 🔵 nit: iap_repository CI build fails on `lines_longer_than_80_chars` (ABOUTME line).
.github CI: 🔴 bug: Format, Generated Files, iap_repository build, and aggregate Mobile CI all red on this branch — needs `dart format` + the ratchet fixes above before undrafting.
mobile/lib/screens/settings/supporter_screen.dart:252: 🔵 nit: `_Disclaimer` missing `const` constructor.

rabble:
Closing as superseded by #6413, which merged the server-verified supporter subscription implementation and includes the intended scope of this draft. Reviving this older client-only branch would duplicate and conflict with the merged money-path implementation.

PR #6375, opened 2026-07-25
perf(ci): shard the Mobile CI test job across four runners

rabble:
## Why

`Tests` was Mobile CI's critical path: every other app CI job runs in parallel underneath it, and recent runs had a 599s p50 for the test job / 625s for the Mobile CI critical path.

`very_good test --optimization` collapses every untagged test file into a single `.test_optimizer.dart` suite. The 18 `skip_very_good_optimization` files run separately, but once they finish the merged bundle runs alone on one process and one vCPU.

## What changed

`very_good` has no "run this subset" flag, so each matrix leg reduces `test/` to its own slice before `very_good` scans it:

```yaml
strategy:
  fail-fast: false
  matrix:
    shard: [0, 1, 2, 3]
...
- run: bash scripts/ci/select_test_shard.sh --total ${{ strategy.job-total }} --index ${{ strategy.job-index }}
- run: very_good test --optimization --concurrency=4 ...
```

The matrix length is now the single source of truth for both `--total` and `--index`, so extending or shrinking the matrix cannot silently drop a shard's tests.

Assignment is round-robin over the lexicographically sorted file list. Only `*_test.dart` entry points are removed. Helpers, mocks, fixtures, builders, and `flutter_test_config.dart` stay because surviving tests may import them.

## Measured

All four shards run and pass locally:

| Shard | Files | Wall | Tests |
|---|---:|---:|---:|
| 0 | 307 | 83s | 3247 |
| 1 | 307 | 215s | 3090 |
| 2 | 306 | 89s | 3129 |
| 3 | 306 | 173s | 3699 |

CI run 30147414354:

| Leg | Job total | `Run Flutter tests` step | Shard-select step |
|---|---:|---:|---:|
| Tests (shard 0/4) | 202s | 143s | 1s |
| Tests (shard 1/4) | **219s** | 159s | 1s |
| Tests (shard 2/4) | 163s | 110s | 1s |
| Tests (shard 3/4) | 210s | 150s | 1s |

| | Baseline | This PR | Change |
|---|---:|---:|---:|
| `Tests` wall time | 599s | **219s** | **-380s (-63%)** |
| `Run Flutter tests` step | 558s | 159s | -399s |
| Mobile CI critical path | 625s | **245s** | -380s |

[...]

PR #6380, opened 2026-07-25
perf(ci): add a wall-clock budget guard so the CI wins can't erode

### LINKED ISSUE #6381: perf(ci): add a wall-clock budget guard so the CI wins can't erode
Add a Mobile CI wall-clock timing budget guard so job-duration regressions are
visible before they silently erase recent CI performance wins.

Requirements:

- Store per-job warn/fail thresholds in `.github/ci-timing-budgets.json`.
- Run the checker from the terminal `mobile-ci` gate, after upstream job
  durations are final.
- Warn on over-budget drift and fail the gate on hard budget regressions.
- Fail closed when the checker cannot measure reliably.
- Ensure budget-file edits themselves trigger validation.
- Keep the guard covered by focused tests that exercise real checker behavior.

PR: https://github.com/divinevideo/divine-mobile/pull/6380
Branch: `perf/ci-timing-budget`

rabble:
Closes #6381

## Why

Mobile CI job durations are easy to regress because the terminal `mobile-ci`
gate only checks whether upstream jobs passed. Recent successful `main` runs
put the `Tests` job between 412s and 597s after #6371; without a budget, a slow
drift back toward the old ~10 minute range would stay invisible until someone
looked at Actions timing by hand.

This PR adds a wall-clock budget guard for the Mobile CI gate. It also records
the current performance-pass status in `PERF_AUDIT.md` without treating still
open PRs as shipped.

## What changed

**`.github/ci-timing-budgets.json`** — per-job `warn` / `fail` thresholds in
seconds, calibrated against recent successful `main` runs. `Tests` now warns at
600s and fails at 780s.

**`.github/workflows/mobile_ci.yaml`** — budget-file edits now trigger app CI,
and the terminal gate uses a sparse checkout of only the budget file plus CI
scripts before running the timing guard.

**`mobile/scripts/ci/check_ci_timing_budget.py`** — runs from the terminal
`mobile-ci` gate after upstream job durations are final. Over `warn` emits a
GitHub warning annotation; over `fail` fails the gate.

**`mobile/scripts/ci/ci_github.py`** — shared GitHub API and timestamp helpers
used by both the checker and `report_mobile_ci_run.py`, so both scripts parse
GitHub timestamps the same way, including offset timestamps.

**`PERF_AUDIT.md`** — documents what is merged now, what this PR ships, and
which measured perf follow-ups are still open.

## Guard behavior

A timing guard that passes when it cannot measure is decorative. This one:

- exits 2 when it cannot check, including unreadable or malformed budget files,
  empty job payloads, non-numeric thresholds, malformed GitHub JSON, invalid
  timestamps, or successful jobs with no completed duration
- reports a budgeted job that did not run with a `::notice`, so renaming or
  skipping a job does not disappear quietly
- ignores failed job runs because a cut-short job says nothing useful about
  speed
- matches exact job names or matrix legs only, so `Tests` covers
  `Tests (shard 0/4)` but not `TestsExtra`

## Verification

- `dart format test/tools/check_ci_timing_budget_test.dart`
- `python3 -B mobile/scripts/ci/check_ci_timing_budget.py --help`
- `python3 -B mobile/scripts/ci/report_mobile_ci_run.py --help`
- `mise exec -- flutter test test/tools/check_ci_timing_budget_test.dart` —
  15/15
- `mise exec -- flutter analyze test/tools/check_ci_timing_budget_test.dart`
  — no issues
- real GitHub jobs API check against run `30150120791`:
  `python3 -B mobile/scripts/ci/check_ci_timing_budget.py --budgets .github/ci-timing-budgets.json --repo divinevideo/divine-mobile --run-id 30150120791`


NotThatKindOfDrLiz (approved):
Takeover fixes are in and CI is green. I did a final pass over the full diff: the guard now validates budget edits, derives expected coverage from the gate needs list, shares timestamp/GitHub helpers with the timing reporter, fails closed on malformed input, and the docs/body now describe the actual merge state.

NotThatKindOfDrLiz:
@rabble I took over the review fixes and pushed `eeb760d8bf`.

What changed:

- rebased the PR branch onto current `origin/main`
- made `.github/ci-timing-budgets.json` edits trigger app CI
- changed the terminal gate checkout to a sparse checkout of only the budget inputs
- shared GitHub API/timestamp helpers between the checker and timing reporter
- hardened fail-closed behavior for malformed GitHub JSON, non-numeric budgets, invalid timestamps, and successful jobs with missing duration data
- derived the budget coverage test from `mobile-ci.needs` instead of a hardcoded job list
- rewrote `PERF_AUDIT.md` so it documents the current merged state and calls #6372/#6375/#6376 open follow-ups rather than shipped work
- updated the PR and issue bodies to match the current scope

Verification:

- `dart format test/tools/check_ci_timing_budget_test.dart`
- `python3 -B mobile/scripts/ci/check_ci_timing_budget.py --help`
- `python3 -B mobile/scripts/ci/report_mobile_ci_run.py --help`
- `mise exec -- flutter test test/tools/check_ci_timing_budget_test.dart` — 15/15
- `mise exec -- flutter analyze test/tools/check_ci_timing_budget_test.dart` — no issues
- live checker run against GitHub Actions run `30150120791` passed
- PR CI is green, including the terminal `Mobile CI` gate; its log shows the budget table and all jobs under threshold

realmeylisdev (commented):
Ran the guard end-to-end against run 30161622594 and it does what it says. `Tests` `warn: 600` lands on the p90 (597s) I measured across 60 successful `main` runs, `fail: 780` never trips historically, and the hardening commit closed the fail-open paths.

One thing worth sorting before #6375 lands, plus three nits — inline.

mobile/test/tools/check_ci_timing_budget_test.dart
```diff
@@ -0,0 +1,325 @@
...
+  ).readAsLinesSync();
+  final jobNamesById = <String, String>{};
+  String? currentJobId;
+  for (final line in workflow) {
+    final jobMatch = RegExp(r'^  ([a-z0-9-]+):$').firstMatch(line);
+    if (jobMatch != null) {
+      currentJobId = jobMatch.group(1);
+      continue;
+    }
+    final nameMatch = RegExp(r'^    name: (.+)$').firstMatch(line);
```
realmeylisdev, line 292: This takes the raw `name:` line, which stops being a literal the moment a job goes matrix. #6375 sets:

```yaml
name: Tests (shard ${{ strategy.job-index }}/${{ strategy.job-total }})
```

I applied just that one line to a scratch copy of the workflow and ran this file — it goes red:

```
Expected: contains 'Tests (shard ${{ strategy.job-index }}/${{ strategy.job-total }})'
Tests (shard ${{ strategy.job-index }}/${{ strategy.job-total }}) has no timing budget
```

The part that makes this worth fixing here rather than there: the obvious way to make it green is to add that templated string as a budget key — and that silently disables the budget. At runtime the job is named `Tests (shard 0/4)`, so `matches()` compares it against the key `Tests (shard ${{ … }})`, misses on both the equality and the `startswith("… (")` branch, and `Tests` drops into `unseen` as a `::notice`. Green gate, no budget on the slowest job in the run.

So the test's idea of a job name and `matches()`'s idea diverge on exactly the matrix case the guard is built to support. Normalising to the static prefix before the lookup keeps the two rules in step and works for both shapes:

```dart
final raw = nameMatch.group(1)!;
final staticPrefix = raw.split(RegExp(r' \(|\$\{\{')).first.trim();
jobNamesById[currentJobId] = staticPrefix;
```

`Tests` → `Tests` today, `Tests (shard ${{ … }})` → `Tests` after sharding.

Minor, same function: the job-id regex is `[a-z0-9-]+`, so an id with an underscore (`generated_files:`) wouldn't be picked up. It fails loudly rather than silently, so it's only a readability thing.

.github/workflows/mobile_ci.yaml
```diff
@@ -416,3 +419,25 @@ jobs:
...
+        with:
+          sparse-checkout: |
+            .github/ci-timing-budgets.json
+            mobile/scripts/ci
+          sparse-checkout-cone-mode: false
+      - name: ⏱️ Check CI timing budget
+        # This job only starts once every other job has finished, so the run's
+        # job durations are final here. Budgets live in
+        # .github/ci-timing-budgets.json; over-warn annotates the PR, over-fail
+        # fails this gate. Guards the wins recorded in PERF_BASELINE.md from
```
realmeylisdev, line 434: nit: leftover from the scrub — `PERF_BASELINE.md` was removed from `PERF_AUDIT.md` and the budget JSON, but survives here. It's the only reference left in the PR, and the file is on neither this branch nor `main`:

```
$ git grep -n PERF_BASELINE HEAD -- .
HEAD:.github/workflows/mobile_ci.yaml:434:        # fails this gate. Guards the wins recorded in PERF_BASELINE.md from

$ git cat-file -e origin/main:PERF_BASELINE.md  # -> missing
```

Probably wants to be `PERF_AUDIT.md`.

PERF_AUDIT.md
```diff
@@ -0,0 +1,100 @@
...
+  clean-generation proof, so caching needs its own correctness argument.
+- Larger GitHub runners: potentially useful, but a cost decision rather than a
+  code change.
+
+---
+
+## How to re-measure
+
+```bash
+# Per-job and per-step timing for one Mobile CI run.
```
realmeylisdev, line 94: nit: it's per-job only — `StepTiming` is parsed but only ever feeds the `setup=` column, and nothing prints a step breakdown. Running it on this PR's own run:

```
- Tests: total=8m 11s, setup=0m 50s, status=completed, conclusion=success
...
Test shard summary:
- longest shard: Tests at 8m 11s
```

Suggest dropping "and per-step" — the earlier p50/p90 wording went in the same pass.

.github/workflows/mobile_ci.yaml
```diff
@@ -416,3 +419,25 @@ jobs:
...
+        # eroding silently.
+        if: ${{ github.event_name != 'pull_request' || needs.changes.outputs.app == 'true' }}
+        env:
```
realmeylisdev, line 437: nit: worth pinning the permission this depends on. The gate job's token is narrower than the endpoint it calls — from the run log:

```
##[group]GITHUB_TOKEN Permissions
Contents: read
Metadata: read
Packages: read
```

No `Actions: read`, and there's no `permissions:` block in this workflow. `GET /repos/{owner}/{repo}/actions/runs/{id}/jobs` succeeds anyway purely because the repo is public — I confirmed it works in run 30161622594. If the repo ever goes private the call 403s, `fail_to_run` fires, and the gate exits 2 on every PR until someone notices.

```yaml
    permissions:
      contents: read
      actions: read
```

on `mobile-ci` makes the dependency explicit and survives a visibility change.

mobile/scripts/ci/ci_github.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
```
realmeylisdev, line 1: nit: both new scripts are committed `100644` while the sibling `report_mobile_ci_run.py` is `100755`, and all three carry a shebang. No behavioural impact (the workflow invokes `python3 <path>`), just inconsistent:

```
100644 mobile/scripts/ci/check_ci_timing_budget.py
100644 mobile/scripts/ci/ci_github.py
100755 mobile/scripts/ci/report_mobile_ci_run.py
```

PR #5522, opened 2026-06-25
fix(video-editor): release feed codecs when opening camera/editor

hm21:
Closes #5523

## Description

On codec-limited (Qualcomm) devices the home feed's **current** native video player was never released when navigating into the camera or video editor. The feed stays mounted in the background (`StatefulShellRoute`), so its decoder kept holding a hardware codec instance across `profile → drafts → editor`. The device then ran out of codec instances, producing the failures seen in user logs:

- editor **preview** → `ERROR_CODE_DECODER_INIT_FAILED` (even with `format_supported=YES`)
- exporter → encoder `RENDER_ERROR`
- the leaked feed player → `ERROR_CODE_DECODING_RESOURCES_RECLAIMED` exactly when the export encoder spun up

Background feeds already release their *neighbour* players when inactive, but intentionally keep the *current* player warm for instant resume on tab switches (the correct behaviour, matching TikTok). The gap was the codec-heavy surfaces (camera/editor/exporter), where even the current player must be released.

 

### What changed

- **`CodecHeavySurfaceCubit`** (app-global, ref-counted): asserts a `bool` while a camera/editor/exporter screen is on screen. Count lives in the state, not as a mutable field.
- **`InfiniteVideoFeed.releaseCurrentWhenInactive`** (new flag): when set and the feed is inactive, the *whole* pool is drained (current player included) and re-initialised when the feed becomes visible again. Plain tab switches are unaffected — the current player stays warm.
- **`FeedVideos`** watches the cubit (nullable lookup → graceful no-op when not provided) and forwards the flag.
- **`CodecHeavySurfaceGuard`** mixin: camera + editor screens assert the signal **only after their entrance transition completes**, so the feed frame stays visible during the push animation instead of disappearing mid-transition. The matching release fires on dispose, and only if the signal was actually asserted (a screen popped mid-transition stays balanced).

**Related Issue:** Relates to the performance/codec crash reports on Qualcomm devices (againstri).

## Out of Scope

- **Encoder-config robustness** in the native render path (`operating-rate=Integer.MAX_VALUE`, High-profile 1080p fallback to software encoder) — a separate, independent fix for the same export failure. Tracked as follow-up, not in this PR.
- 4 pre-existing timing-sensitive test failures in `infinite_video_feed` services (`controller_subscriptions`, `load_watchdog`, `stale_playback_detector`, `source_loader`) — verified present on the base branch **without** this diff, so unrelated.

[...]

PR #6384, opened 2026-07-25
fix(video-editor): recover from export encoder-init failures

### LINKED ISSUE #6385: fix(video-editor): recover from export encoder-init failures
## Problem

On codec-limited Android devices, a multi-clip export can fail with `RenderEncoderException` / `ERROR_CODE_ENCODER_INIT_FAILED` immediately after the editor preview decoder is released. The native renderer exhausts its in-process fallback attempts in the same short codec-reclaim window, so the edit falls back to a draft with only the existing render-failed retry affordance.

The export path also normalizes mixed-resolution clips before the final concatenate pass, so encoder-init recovery needs to cover both the normalization encodes and the final composite encode.

## Expected

- Full-resolution final exports retry after a short settle delay on encoder-init failure.
- Final exports can fall back from FHD to HD when the device cannot allocate the 1080p encoder.
- Mixed-resolution normalization encodes get the settle retry without forcing a reduced final export canvas.
- User/teardown cancellation during retry delays or native encoding aborts the retry loop instead of starting or accepting a later encode.
- Existing non-encoder failures still surface normally.

## Verification

- Add focused unit coverage for encoder fallback attempts, reduced-resolution fallback, no-latency happy path, non-encoder failures, normalization-style retry without reduced quality, and cancellation during settle/encode.

## Source

Auto-created from PR #6384.

PR: https://github.com/divinevideo/divine-mobile/pull/6384
Branch: `fix/video-editor-encoder-fallback`

hm21:
Closes #6385

## Description

Multi-clip exports could fail outright with `RenderEncoderException` (`ERROR_CODE_ENCODER_INIT_FAILED`) on codec-limited devices, dropping the whole edit to a draft with only a `Video render cancelled or failed` warning.

Device logs told the real story: the encoder failed every fallback attempt (operating-rate cap/removal, software encoder, profile downgrade) within ~250 ms, right after the editor's preview decoder was released, and `DECODING_RESOURCES_RECLAIMED` errors showed up during editing. That points to transient MediaCodec contention / genuine device codec pressure, not a real format incompatibility. The plugin's own in-process fallback chain runs all its attempts in that same narrow window, so it can never outlast the codec-reclaim latency.

The fix, in `renderWithEncoderFallback`:

- Attempt 1: immediately, full config. The happy path pays zero extra latency.
- Attempt 2: after a settle delay, same config. Defeats the codec-reclaim race.
- Attempt 3: for final composite exports only, after another settle delay at reduced resolution (FHD -> HD/720p). Covers devices genuinely at their 1080p encoder ceiling.
- Mixed-resolution normalization encodes now use the settle retry too, without forcing a reduced final export canvas.
- User/teardown cancels recorded during a settle window or an active encode abort the retry loop instead of starting or accepting a later encode.
- Non-encoder failures, including `RenderCanceledException`, are not retried.

Native log accuracy: the Android `dispose` handler emitted its lifecycle log before the actual `release()`, so teardown looked finished while the hardware codec could still be held. This moves the log after release so it marks codec-release completion truthfully.

The primary UI-side "release the preview decoder before the export encoder" ordering already exists (the #5522 decoder-release gate); this PR is the recovery path for residual device-pressure failures that the gate cannot prevent.

## Out of Scope

- Product/analytics treatment for silent 720p fallback frequency. This PR keeps the existing save-the-edit behavior and logs retry attempts; export-quality telemetry can be added with the broader analytics work.
- Deeper coordination between export start and navigation-observer decoder teardown. The retry/cancel handling here keeps the fix local to the render pipeline.

## Verification

- `flutter test test/services/video_editor/video_editor_render_service_test.dart` — 20/20 pass
- `flutter analyze lib/services/video_editor/video_editor_render_service.dart test/services/video_editor/video_editor_render_service_test.dart` — No issues found
- Pre-push hook: changed-file analyzer and `test/services/video_editor/video_editor_render_service_test.dart` — pass

## Type of Change

### DIFF
--- a/mobile/lib/constants/video_editor_constants.dart
+++ b/mobile/lib/constants/video_editor_constants.dart
@@ -191,6 +191,26 @@ class VideoEditorConstants {
   /// Video quality for recording and editing
   static const DivineVideoQuality quality = DivineVideoQuality.fhd;
 
+  /// Fallback export quality used after the device rejects the primary
+  /// [quality] with a `RenderEncoderException`.
+  ///
+  /// A 720p canvas needs far fewer encoder resources than 1080p, so a
+  /// codec-limited device that cannot allocate an [quality] encoder can still
+  /// finish the export at this reduced tier instead of losing the edit.
+  static const DivineVideoQuality encoderFallbackQuality =
+      DivineVideoQuality.hd;
+
+  /// Settle delay between export encoder attempts after a
+  /// `RenderEncoderException`.
+  ///
+  /// Encoder-init failures on codec-limited devices are usually transient
+  /// contention: the preview decoder the editor just released (the #5522
+  /// decoder-release gate) — or another app codec — has not been reclaimed by
+  /// the OS media server yet. The plugin's own in-process fallback chain runs
+  /// every attempt within a few hundred ms, so it cannot outlast that window;
+  /// waiting before the Dart-level retry lets the codec free up.
+  static const encoderRetrySettleDelay = Duration(milliseconds: 800);
+
   /// Hero animation tag for the back button in the video editor.
   static const heroBackButtonId = 'Video-Editor-Back-Button';
 
--- a/mobile/lib/services/video_editor/video_editor_render_service.dart
+++ b/mobile/lib/services/video_editor/video_editor_render_service.dart
@@ -283,6 +283,8 @@ class VideoEditorRenderService {
   static final _compositeProgressController =
       StreamController<ProgressModel>.broadcast();
 
+  static final Set<String> _cancelledTaskIds = <String>{};
+
   @visibleForTesting
   static double proofModeProgressBudgetForClipCount(int clipCount) {
     final normalizedClipCount = clipCount < 1 ? 1 : clipCount;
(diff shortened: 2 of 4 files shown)


NotThatKindOfDrLiz:
@hm21 I took over the review fixes and pushed `7c8a8321f4` to this PR branch.

What changed:
- Extended the encoder-init settle retry to mixed-resolution normalization encodes without applying the final-export 720p fallback there.
- Added task-id cancellation tracking so cancels recorded during retry settle windows abort before the next native encode starts.
- Added a post-encode cancellation check so a cancel requested during native encoding cannot be accepted as a successful retry if the native cancel returns late/no-ops.
- Replayed the PR onto fresh `origin/main`, which also removed the stray `pro_image_editor` downgrade from the original diff.
- Expanded #6385 from the auto-created stub into a real issue body with expected behavior and verification notes.

Verification:
- Local: `flutter test test/services/video_editor/video_editor_render_service_test.dart` — 20/20 pass
- Local: `flutter analyze lib/services/video_editor/video_editor_render_service.dart test/services/video_editor/video_editor_render_service_test.dart` — no issues
- Pre-push hook: changed-file analyze + targeted render-service test — pass
- GitHub CI: all checks green, including Analyze, Tests, Format, Generated Files, Android Unit Tests, preview, linked issue, mergeability, semantic PR, and CLA.

NotThatKindOfDrLiz (approved):
Final pass is clean. The takeover commit covers the normalization encode gap and closes the retry-settle cancellation edge with focused tests. All GitHub checks are green.

Issue #6387, opened 2026-07-25
fix(report): moderation DM failure is reported to the user as success


realmeylisdev:
## Problem

`ReportContentDialog` tells the user their report reached the moderation team even when the
moderation DM was never delivered. The "we'll follow up" copy (`reportModerationDmDelayed`)
is effectively unreachable for every realistic failure mode.

`DmRepository.sendMessage` returns a sealed `NIP17SendResult` — it signals failure by
**returning** `NIP17SendFailure`, not by throwing. The call site discards the return value
and only sets its failure flag inside `catch`:

```dart
// mobile/lib/widgets/report_content_dialog.dart:441-462
var moderationDmFailed = false;
try {
  await dmRepo.sendMessage(          // <- return value discarded
    recipientPubkey: labelService.divineModerationPubkeyHex,
    content: _formatReportDm(...),
    skipNip04Fallback: true,
  );
} catch (e) {
  // The report itself already succeeded (relay + Zendesk); the
  // moderation DM is a secondary notification. Don't fail the
  // flow, but surface the outcome instead of swallowing it so
  // the user isn't told the team was reached when it wasn't.
  moderationDmFailed = true;
  ...
}
```

The comment states the intent exactly. The code does not achieve it, because nothing throws.

## Why the catch almost never fires

`sendMessage` has ~14 non-throwing failure returns across the two files:

- `dm_repository.dart:2717` — `NIP17SendResult.blocked('blocked: recipient not permitted by send policy')`, the protected-minor gate (#176)
- `dm_repository.dart:2885` — `failure('NIP-04 fallback failed: …')`
- `dm_repository.dart:3182`, `:3399` — `failure('rumor JSON parse failed: …')`
- `dm_repository.dart:4763,:4768,:4777,:4787` — signer unavailable / encrypt / sign / publish failed
- `nip17_message_service.dart:361,:365,:384,:422,:554,:597,:647,:665` — including two more `blocked` returns

`NIP17SendResult` is explicitly designed for this (`mobile/packages/models/lib/src/nip17_send_result.dart:41`), and its doc notes that callers "only check `success`" — this call site checks nothing.

## Impact

A user submits a report; the confirmation screen omits `reportModerationDmDelayed` and offers
the "Message the moderation team" deep link into a thread the team never received anything in.

The worst case is the protected-minor path: a restricted minor whose DM is `blocked` by
`dmSendPolicyProvider` gets a clean success screen. That is the population least able to
[...]

Issue #5304, opened 2026-06-18
fix(dm): hide message requests after reinstall


hm21:
### Summary

Some existing direct messages are incorrectly displayed as message requests after reinstalling the app.

### Environment

android, main-branch

### Steps to Reproduce

1. Install and use the app with existing direct message conversations.
2. Uninstall and reinstall the app.
3. Log back into the same account and open messages.

### Actual Result

Some existing direct message conversations appear under message requests.

### Expected Result

All existing direct message conversations should remain in the normal inbox and not be reclassified as message requests.

### Regression?

Unknown

### Evidence

https://github.com/user-attachments/assets/c2263f14-13be-4fce-ae90-1895bd400f97

hm21:
@realmeylisdev that's a really weird issue. The issue started last night when we were investigating why some people could no longer start the app and I reinstalled the app using a release build. 

Let me know when you need logs but i don't think they are helpful cuz the restore part happen already and i noticed it later.

realmeylisdev:
@hm21 thank you for sharing this one, please attach the logs if available.

hm21:
Sure this are the logs

[openvine_full_logs_2026-06-18T11-21-26.123736.txt](https://github.com/user-attachments/files/29087262/openvine_full_logs_2026-06-18T11-21-26.123736.txt)

Issue #4976, opened 2026-06-08
fix(dm): Messages unread badge undercounts — excludes unread chats from followed users you haven't replied to


realmeylisdev:
### Summary

The "Messages" unread badge (bottom-nav + inbox segmented toggle) can show a **lower** number than the count of conversations actually rendered with an unread dot in the Messages list. Reported by @hm21 while testing #4973 (the #4953 reinstall-recovery fix): the badge showed **3** while **4** conversations had unread dots.

### Root cause (verified in code)

The Messages list and the badge use **two different definitions of "unread accepted conversation":**

- **Visible list is follow-aware.** `ConversationListBloc` composes it as `mergeAndSort(accepted, split.followed)` — accepted conversations (`currentUserHasSent == true`) **plus** potential requests from users the current user *follows* (`conversation_list_bloc.dart:86-91`). Each row shows an unread dot whenever `!conversation.isRead` (`conversation_tile.dart:134`), regardless of `currentUserHasSent`.
- **Badge is follow-blind.** `DmUnreadCountCubit` subscribes to `ConversationsDao.watchUnreadAcceptedCount` (`dm_unread_count_cubit.dart:23`), which counts `isRead == false AND currentUserHasSent == true` only (`conversations_dao.dart:298-309`).

⇒ A conversation from a **followed** user that you have **never replied to** (`currentUserHasSent == false`) appears in the Messages list with an unread dot but is **not** counted by the badge. The badge undercounts by the number of such conversations.

This is pre-existing (the badge query predates #4973) but was effectively invisible until #4973 began backfilling the full conversation list after reinstall, which surfaces followed-but-never-replied conversations.

### Expected

The Messages badge count == the number of unread conversations actually shown in the Messages list.

### Fix direction

Make the unread count follow-aware so it matches the list composition: count unread across `accepted ∪ followed-requests`. Options:

- Lift the count into a follow-aware source that mirrors `ConversationListBloc`'s `classifyPotentialRequests` / `mergeAndSort` (subscribe to accepted + potentialRequests + followingStream, count `!isRead` over the merged set). Share the classification rather than duplicating it.
- Or, repository-layer: expose a `watchUnreadInboxCount(followed)` and keep the cubit thin.

Per-row dot logic stays unchanged. Add a test that **fails on the current follow-blind count** (an unread followed-but-never-replied conversation must increment the badge).

### Affected files

- `mobile/lib/blocs/dm/unread_count/dm_unread_count_cubit.dart`
[...]

PR #6388, opened 2026-07-25
feat(dm): pin a Divine Moderation support row at the top of the inbox

### LINKED ISSUE #6283: feat(dm): pin a "message Divine Moderation" support row at the top of the inbox
> **Body corrected 2026-07-25** against divine-mobile `origin/main` `0ad2b3e39`, divine-web
> `origin/main` `d1b8ad32`, live production relay data, and the app running on device.
> Corrected in place: the locale count (16 → 17), the `inbox_view.dart` line refs invalidated by
> #6286, the persisted-row behaviour (it lands in Message Requests, it does not sink to the bottom),
> `computeConversationId`'s nullability, the `05.md` citation, web's `SupportRow` guard, web's
> kind-10050 relay handling (it violates NIP-17 rather than degrading gracefully), the
> `dmSendPolicyProvider` fail-closed scope, and the unread-badge source. The single biggest change
> is under *Affected layers → UI*: injecting into the `ListView.builder` is **not sufficient** —
> there are five render branches and `_MessagesContent` is the only insertion point that survives
> them all. Full evidence in the investigation notes; see also the correction comment below.

## Problem

divine-mobile's inbox has **no pinned / support / official conversation row**. `_ConversationListView`'s `ListView.builder` (`mobile/lib/screens/inbox/inbox_view.dart:819-863`) renders only `MessageRequestsBanner`, a trailing loader, and `ConversationTile` rows. The only `official` references under `mobile/lib/screens/inbox` are the protected-minor gate wiring (#176) — never a UI row.

divine-web ships a pinned support row at the top of its DM list. This issue is to bring the mobile inbox to parity.

**Before promoting this row, two things need product / Trust & Safety sign-off (tracked in `support-trust-safety#192`):**

1. **Backend readiness — is the moderation@ inbox actually monitored end-to-end?** A promoted "message Divine Moderation" row would route DMs that physically arrive and are ingested by the backend, but the reply-and-notify loop is **not** confirmed: there is no established SLA, no reply-back-to-user notification path (push cannot carry gift-wrapped DMs), and the operator surface for reading inbound support DMs is not confirmed staffed. If replies are not surfaced back to the user, a prominent "message support" button reads as a **silent black hole**. Details in the private companion issue.
2. **DM privacy posture for a promoted support inbox.** The relay currently serves gift-wrap metadata without authentication, so who-messaged-support is more exposed than a mainstream "support chat" implies. A promoted button multiplies that surface. Details in the private companion issue.

These are **decisions to resolve** (see Open questions OQ6 / OQ10), not facts to assume. Web already ships this row unconditionally into the same inbox, so mobile parity would inherit the same posture.

### Non-blocker to state explicitly

**`FF_PUBLISH_DM_RELAY_LIST` being OFF does NOT block this feature.** That flag gates only the *app's own* kind-10050 publish (whether the signed-in user advertises their own DM inbox), via `ensureDmRelayListPublished` (`mobile/packages/dm_repository/lib/src/dm_repository.dart:2513`). Recipient routing to moderation@ goes through `resolveDmInboxRelays` (`dm_repository.dart:2437`, used unconditionally by the send path) and is **not** flag-gated. Default OFF at `mobile/lib/features/feature_flags/services/build_configuration.dart:37-39`.

---

## Reference (divine-web)

Reference implementation at divine-web `origin/main` `d3067aa0c7b98f045c56df9d375c43836b78d4aa`:

- The pinned row is a **file-private `SupportRow`** component (interface `src/pages/MessagesPage.tsx:157-163`, function `:165-199`), NOT the shared `ConversationRow`. It copies `ConversationRow`'s button className but has a distinct body: avatar with primary-tinted ring, displayName + a pill badge, one-line summary — **no timestamp, no unread badge**.
- It mounts in the non-search branch of the DM section (`MessagesPage.tsx:365-371`): `onClick={() => openConversation([DIVINE_SUPPORT_PUBKEY])}`. No guard on `conversationsQuery.data` and **no feature flag** — but it is **not** unconditional: the whole section is gated on `canUseDirectMessages` (`MessagesPage.tsx:310`), which is `Boolean(user?.pubkey && signer?.nip44 && nip44Healthy)` (`useDirectMessages.ts:217`), so it is **hidden when signed out** or when the signer lacks a working NIP-44 round trip.
- Destination is single-sourced: `src/lib/dm.ts:23` `export const DIVINE_SUPPORT_PUBKEY = DIVINE_MODERATION_PUBKEY;`, where `DIVINE_MODERATION_PUBKEY = '8fd5eb6d8f362163bc00a5ab6b4a3167dbf32d00ec4efdbcf43b3c9514433b7e'` (`src/lib/officialAccounts.ts:125-126`). So web's support destination **==** the mobile moderation pin. Reconciled by web PR #474 / commit `9d77c523`.
- Display-name precedence: `getDisplayName` short-circuits to `metadata?.display_name || metadata?.name || 'Divine Support'` (`MessagesPage.tsx:32-38`) — the **fallback literal is 'Divine Support', not 'Divine Moderation'**. Live kind-0 supplies `display_name: 'Divine Moderation'`, so shipped web shows 'Divine Moderation' from the relay.
- **Copy inconsistency (shipped):** the pill badge is a static `messages.supportBadge = 'Support'` (never driven by `OfficialAccount.role`), while the account's role is `'moderation'` and its kind-0 title is 'Divine Moderation'. `messages.supportSummary = "Bugs, moderation, account stuff — we're listening."` Copy is translated into **16 locales** (`src/lib/i18n/locales/en/common.json:114-115` + 16 locale dirs) — but those translations are **stale**: the brand refresh (`a4ccb4c5`, divine-web#259) rewrote the English `supportSummary` and updated **zero** translations, so 14 of 15 non-English locales still carry the pre-refresh meaning. Web also has no `am`/`bg`.
- **CONFIRMED web double-render bug:** there is zero dedupe of the moderation pubkey against the real conversation list (`useDirectMessages.ts:333-361`; `MessagesPage.tsx:365-390`). A user with an actual moderation DM thread sees it **twice** — synthetic `SupportRow` (always) + a real `ConversationRow`.
- Web is **read-only** on kind 10050 but **unions five relay sources instead of restricting to the recipient's 10050 — a NIP-17 violation** (`17.md:77` "MUST only publish events to the relays in the recipient's kind 10050"), and its 10050 lookup is wrapped in `.catch(() => ({}))` (`dm.ts:666`) so a failed lookup silently drops the only spec-mandated input. Consequence: every support DM sent from web is **also broadcast to `relay.damus.io`, `relay.primal.net` and `purplepag.es`**. **divine-mobile is the more conformant implementation here** (`dm_repository.dart:2766/2777`, `nip17_message_service.dart:433-450`) and must NOT copy web on this point.
- `ConversationPage` hard-overrides the subtitle for this one pubkey to `'Private support chat'` before inspecting nip05/name (`ConversationPage.tsx:35-52`).
- Test coverage of the pinned row is **one** vitest assertion (`MessagesPage.test.tsx:185-193`) — presence + outside-the-header only; nothing pins ordering, badge/summary text, route, or the double-render. No e2e.
- The 2026-03-29 design doc asked the row to **reuse `ConversationRow` styling**; shipped code forked a bespoke `SupportRow` (`docs/superpowers/specs/2026-03-29-messages-header-layout-design.md:39,44`).

---

## Proposed change

Add a pinned "message Divine Moderation" row at the top of the mobile inbox that opens (or starts) a NIP-17 conversation with moderation@ (`8fd5eb6d8f362163bc00a5ab6b4a3167dbf32d00ec4efdbcf43b3c9514433b7e`), mirroring the divine-web `SupportRow` as a **second, parallel** support entry point (the existing "Contact Support" CTA opens the Zendesk native ticket list via `ZendeskSupportService.showTicketListScreen`, `mobile/lib/screens/settings/support_center_screen.dart:45-47,182,195` — this row does not replace it).

The concrete implementation shape, identity source, copy, dedupe behavior, unread wiring, signed-out behavior, and flag-gating are **deliberately left open below** — several are genuine product/architecture decisions, and backend readiness may be a hard blocker. This issue documents the verified surface and the decisions; it does not pre-pick them.

---

## Affected layers / files (UI → BLoC → Repository → Client)

**UI**
- `mobile/lib/screens/inbox/inbox_view.dart` — **injecting into the `ListView.builder` is not sufficient.** There are **five** render branches and the builder is only one of them: `:461-465` loading spinner, `:466-470` `InboxErrorState`, `:564` `InboxEmptyState`, `:567-577` requests-only, `:602` `_FilteredEmptyContent`, `:613/:819` `_ConversationListView`. `_ConversationListContent` (`:449-472`) replaces the **entire** list subtree with a `CircularProgressIndicator` during `initial`/`loading`, so any row inside the list subtree is also invisible during cold start. The one insertion point that survives every branch is **`_MessagesContent`'s `Column`** (`:351-372` — `[FollowingBar, _RestoringHistoryIndicator, Expanded(_ConversationListContent)]`), which sits **outside** the status switch and already has `currentPubkey` in scope at `:324`. Tap → reuse `DmRepository.computeConversationId([currentPubkey, moderationPubkey])` exactly as the report flow does (`mobile/lib/widgets/report_content_dialog.dart:245-254`), then route to `mobile/lib/screens/inbox/conversation/conversation_page.dart` (the report flow's own deep link is at `:622-641`). Signed-out: `computeConversationId` returns a **non-nullable `String`** (`dm_repository.dart:5465`); it is the *call site* that builds a nullable local behind a `currentPubkey != null && currentPubkey.isNotEmpty` guard (`report_content_dialog.dart:245-254`), and the button hides in that case.
- l10n: `mobile/lib/l10n/app_en.arb` + all **17** non-English `app_*.arb` (18 arb files: am ar bg de en es fil fr id it ja ko nl pl pt ro sv tr). `reportContactModeration = "Message the moderation team"` (`:2738`) and `reportModerationDmDelayed` (`:2734`) exist and are reusable, but **no pinned-row / badge / summary key exists** — any web-style badge ('Support') or summary would be net-new across all **17** locales (arb_consistency_test enforced) + `flutter gen-l10n`. Note `_knownUntranslatedDebt` does **not** shrink the diff: it bakes the English literal into every generated locale subclass, so it is 19 files either way.

**BLoC**
- `ConversationListBloc` (backs `_ConversationList`; `context.select` reads at `inbox_view.dart:541-562`) — holds loading until `userPubkey` is non-empty; governs whether/when the row renders. Note its `BlocProvider` is **lazy** (`inbox_page.dart:83-99`, no `lazy: false`) and every consumer sits under the `_messagesActivated` gate, so `ConversationListStarted` does **not** fire until the user first taps Messages — the `InboxView` docstring at `:57-62` claiming it "starts immediately" is stale. Any de-dupe of the moderation pubkey from the mapped list belongs here or in the UI mapping, not in a widget filter buried in `build`.
- `DmUnreadCountCubit` — a synthetic (non-persisted) row has no unread state; contributing to the inbox badge is net-new wiring. (Correction: the badge does **not** come from `watchUnreadAcceptedCount`, which has zero call sites in `mobile/lib`; `DmUnreadCountCubit._countUnread` recomposes the visible list in Dart and counts `!c.isRead`.)

**Repository**
- `mobile/packages/dm_repository/lib/src/dm_repository.dart` — `computeConversationId`, `resolveDmInboxRelays` (`:2437`, recipient routing, not flag-gated), `ensureDmRelayListPublished` (`:2513`, own-publish, flag-gated). Any "start empty thread" / pin-ordering composition belongs here, not the UI.
- `mobile/packages/db_client/lib/src/database/daos/conversations_dao.dart:185-205` — `watchAcceptedConversations` filters on `t.currentUserHasSent.equals(true)` (`:193`), so a *persisted* conversation the user has never sent to **does not appear in the inbox at all**; it is routed to `watchPotentialRequestConversations` (`:212-228`) and lands in **Message Requests** unless every non-self participant is followed — and moderation@ is **not** auto-followed anywhere. There is **no top-pin ordering mechanism** for a persisted row today. (Adding a drift column is cheaper than it sounds: `schemaVersion` has never left 1 and the app uses `_addColumnIfMissing` in `beforeOpen`, so it is 3 regenerated artifacts + one line — but note `List<Index>` getters are **not** wired through `@DriftDatabase`, so any index needs a manual `customStatement`.)
- Identity sources (a genuine mobile fork web does not have):
  - Static pin: `mobile/lib/config/official_accounts.dart:31-46` — `kPinnedOfficialAccounts` ships HQ `c4a39f1291291d452405cd8ddd798c4a29a3858c52cd0d843f1f6852cf17682e` (`_@divinehq.divine.video`, role `'hq'`) and Moderation `8fd5eb6d8f362163bc00a5ab6b4a3167dbf32d00ec4efdbcf43b3c9514433b7e` (`moderation@divine.video`, role `'moderation'`), both `minorContactable: true`.
  - Rotation-capable: `mobile/lib/services/moderation_label_service.dart:100-125` — `divineModerationNip05 = 'moderation@divine.video'` (`:108`), `fallbackModerationPubkeyHex` == the pin (`:111-112`), with a NIP-05-follows-rotation path per #4948 tier 1. The report flow uses this getter; the protected-minor gate keys off the static pin. A legacy key `121b915baba659cbe59626a8afaf83b01dc42354dfecaad9d465d51bb5715d72` (`:115-116`) is migrated out — **evidence the rotation path has actually fired in production** (`121b915b…` → `8fd5eb6d…`, commit `683577093` / #2321, 2026-03-20), so OQ2/OQ12 rest on precedent, not a hypothetical. There is also a **third** identity source the body omits: `MinorReviewCase.moderationConversationPubkey` handed down by Keycast (`models/minor_account_review_status.dart:126-127`), which is unvalidated against the pin.

**Client**
- Relay `wss://relay.divine.video` (divine-funnelcake) — moderation@ kind-10050 → this single relay; **no kind-10002** for moderation@ (routing MUST use 10050, aligning with NIP-17's no-NIP-65-fallback rule, `nips@bdfa7e62 17.md:63,77`). NIP-11 advertises NIP-17 and NIP-59.
- Protected-minor gate: `mobile/lib/providers/official_accounts_providers.dart:38-51` (`dmSendPolicyProvider` — note it is **not globally** fail-closed: it delegates to `isDmRestrictedProvider`, which is fail-closed only for `divineOAuth`/ever-Keycast/unauthenticated accounts; `importedKeys`/`bunker`/`amber`/`nip07`/`automatic` with no persisted verdict return `allowed` with no officials check) and `mobile/lib/services/official_accounts_service.dart:101-104` (`isApprovedMinorDmRecipientSync`, default-trust for a pinned `minorContactable` account). Resolves on pin ∩ live NIP-05 with a 1h TTL / 5-min confirming recheck; if `moderation@divine.video` stops resolving to the pinned hex, a restricted minor loses the support path.

---

## Dependencies

- **kind-10050 (routing).** moderation@ has a live 10050 → `wss://relay.divine.video`. Being a replaceable event, **re-confirm with a fresh relay query at build time.** No NIP-65 fallback is permitted (`17.md:63,77`); the org pubkey MUST keep a compliant 10050 or clients cannot send.
- **Backend readiness (make-or-break).** Is the moderation@ inbox monitored, answered within an SLA, and is a reply surfaced back to the user? Tracked in `support-trust-safety#192`. OQ6 governs whether this is a hard pre-ship blocker.
- **Trust model.** The support/moderation identity is operated by Trust & Safety; the operator access and signing scope are tracked privately (`support-trust-safety#192`, and Keycast scope in `support-trust-safety#164`).
- **Protected-minor gate.** The moderation pin is `minorContactable: true` and default-trusted, so the row passes the gate by default for minors — but the row must **respect (not bypass)** the fail-closed `dmSendPolicyProvider`. If NIP-05 stops resolving to the pin, minors lose the path.
- **Anti-impersonation (NIP-17).** A support thread is load-bearing for verification: clients MUST verify the `kind:13` seal pubkey equals the inner rumor pubkey (`17.md:57`), else any sender can impersonate the org. NIP-17/59 in practice mandate NIP-44 v2 (`44.md:21-23`). Mobile already does this; confirm the pinned-row path is not exempt.
- **Non-blocker:** `FF_PUBLISH_DM_RELAY_LIST` (see Problem). Do not gate this feature on it by mistake.

---

## Open questions / decisions needed

Each is a real product or architecture call — **do not silently resolve.**

1. **Synthetic UI-only pinned row vs a persisted `dm_conversations` row?** A persisted empty row is **excluded from the inbox entirely** by `currentUserHasSent.equals(true)` (`conversations_dao.dart:193`) and lands in Message Requests; there is no top-pin ordering. A DB-backed row is additionally invisible during the whole cold-start window (`_watchConversationRows` gates all streams behind `_postAuthMaintenance`, `dm_repository.dart:5259-5269`). Options: (a) synthetic row rendered in `_MessagesContent` (outside the status switch — the only placement that survives all five branches); (b) persisted row + new pin/order column + a change to the accepted-list predicate; (c) hybrid — synthetic until a first message exists.
2. **Which moderation-identity source drives the row: static `kPinnedOfficialAccounts` pin, or `ModerationLabelService.divineModerationPubkeyHex` (rotation-capable)?** They diverge on rotation; report flow uses the getter, the minor gate keys off the pin. Options: (a) static pin; (b) getter (follows NIP-05 rotation); (c) new single-source constant both read.
3. **De-dupe the moderation pubkey out of the real conversation list, or mirror web's confirmed double-render?** Options: (a) de-dupe (diverge from web); (b) mirror web exactly; (c) de-dupe + file a divine-web parity fix.
4. **Row label / copy:** 'Divine Moderation' (kind-0 / pin role), 'Divine Support' (web fallback literal), 'Support' (web badge), or reuse 'Message the moderation team'? Web is internally inconsistent. Drives net-new l10n across 17 locales. Note **'Support' is already overloaded in mobile** — `profileSupportButtonLabel` means *tip this creator*, and there is a separate 'Support Center' — so a 'Support' badge would be a third distinct meaning; web has no such collision. Follow `brand-guidelines/TONE_OF_VOICE.md`.
5. **Target moderation@ (enforcement identity) or stand up a distinct support@ identity?** moderation@ replies read as moderation decisions; a support@ split separates 'help' from 'enforcement' but needs a coordinated keycast + divine-mobile + divine-web pin release (left open in `support-trust-safety#184`). Options: (a) reuse moderation@; (b) stand up `support@divine.video`.
6. **Is confirming the backend surfaces + answers support DMs (with an SLA + user-visible reply) a hard blocker before shipping?** Options: (a) hard-block; (b) ship with explicit 'we'll get back to you, not live chat' copy; (c) ship behind a flag until readiness is confirmed.
7. **Should the pinned conversation participate in unread counts / the inbox badge?** Options: (a) no unread (match web); (b) unread from the real underlying thread's DB state; (c) unread only once a real thread exists.
8. **Render the row when signed out / before credentials are ready?** Options: (a) hide until signed in (report-flow precedent); (b) show but route to sign-in on tap; (c) show disabled.
9. **Feature-flag it, or ship unconditional like web?** Options: (a) unconditional; (b) new dedicated flag (e.g. `FF_SUPPORT_DM_ROW`) default OFF; (c) new flag default ON in non-release builds. (`FF_PUBLISH_DM_RELAY_LIST` is the wrong flag.)
10. **Is the DM privacy posture (see Problem / `support-trust-safety#192`) acceptable for a PROMOTED support inbox, or a mitigation to raise first?** Options: (a) accept and ship; (b) ship but file a relay-hardening prerequisite; (c) block on the hardening.
11. **Copy/UI style: reuse `ConversationTile`, or a bespoke pinned card like web's forked `SupportRow`?** Options: (a) reuse `ConversationTile`; (b) bespoke card; (c) `ConversationTile` with a Support badge slot.
12. **If `moderation@divine.video` NIP-05 later resolves to a different pubkey than the persisted one, what should the support thread do?** NIP-05 discourages silently replacing a stored pubkey (`05.md:63-64,70-72` — the earlier citation `05.md:101-103` was wrong; that range is the HTTP-redirect Security Constraints section). Note `05.md:72` is scoped to the follow list and only prescribes ceasing to display the nip05 identifier, so all three options below remain spec-compatible. Options: (a) warn + keep old thread; (b) start a fresh thread; (c) refuse + show support-unavailable.
13. **Display name from a live kind-0 lookup (like web) or a hardcoded label?** Live kind-0 exists (id `17e11af37af70e934c2dd06f7e3b1f5750263ad231f18604a0fa0f85883fd085`, `display_name: "Divine Moderation"`, `picture: "https://divine.video/divine-logo.svg"`). Options: (a) live kind-0 with hardcoded fallback (mirror web); (b) hardcoded l10n label only; (c) hardcoded label + kind-0 avatar only.

---

## Out of scope

- Backend fixes in divine-moderation-service and any relay hardening — these are **dependencies** tracked in `support-trust-safety#192` / `divine-moderation-service#169`, not deliverables of this mobile issue (OQ6/OQ10 govern whether some must land first).
- The Keycast `allowed_kinds` tightening (`support-trust-safety#164`).
- Stale support docs that still say Divine users "can't see DMs yet" — flag separately.
- Flipping `FF_PUBLISH_DM_RELAY_LIST` unless folded in by decision.

---

## Acceptance criteria

Phrased so the open questions stay open — where an OQ governs, the criterion states the constraint, not the answer.

- A pinned support/moderation row renders at the top of the inbox list, per the shape chosen in OQ1, with signed-out/not-ready behavior per OQ8.
- Tapping the row opens (or starts) the moderation conversation via `DmRepository.computeConversationId([currentPubkey, moderationPubkey])` and routes to the existing `conversation_page.dart`.
- The moderation identity source is the one chosen in OQ2, documented inline; the row cannot silently diverge from the protected-minor gate anchor.
- The row's label/badge/summary copy is sourced per OQ4, with any net-new keys mirrored across all 17 non-English `app_*.arb` locales (or added to `_knownUntranslatedDebt`) and `arb_consistency_test` green.
- The protected-minor `dmSendPolicyProvider` (fail-closed) is respected, not bypassed; a restricted minor sees the correct blocked/allowed state when NIP-05 resolution changes.
- Double-render behavior matches the OQ3 decision, with a test pinning it if de-dupe is chosen.
- Unread/badge (OQ7), signed-out (OQ8), and flag-gating (OQ9) behavior each documented.
- Widget + golden + interaction tests exist (tap → conversation route); test `MaterialApp`s carry `AppLocalizations.localizationsDelegates` + `supportedLocales`; depth exceeds web's single shallow assertion.
- `flutter analyze lib test integration_test` and scoped inbox tests pass.
- The backend-readiness decision (OQ6) is recorded on this issue before the row is promoted to users.

---

## Related issues

Ground-truth commits: divine-mobile `origin/main` `99eebc64fe7bad290132e815c6a21be2fd94242b`; divine-web `origin/main` `d3067aa0c7b98f045c56df9d375c43836b78d4aa`; nips `bdfa7e62ef87fcfcb992b1a27aee49d36b0b4f91`.

- **#5177** — epic parent (`epic(trust-safety): organize moderation, content controls, and reporting`).
- **#5178** — related (`epic(dm): stabilize NIP-17 messaging, inbox, and read state`).
- **#176** — CLOSED; client protected-minor DM restriction whose gate the pin depends on.
- **#4948** — moderation-pubkey trust model (NIP-05 tier-1) driving `ModerationLabelService` rotation (OQ2).
- **#4974** — NIP-17 receive / kind-10050 (RC3); the relay-accepts-10050 precondition (now met).
- **#5493** — MERGED (`156aca6b9`); `ensureDmRelayListPublished` landed (the flag-gated own-publish path, not this feature's routing).
- `support-trust-safety#192` — **private companion** with the backend/security detail (backend inbox readiness + DM privacy posture).
- `support-trust-safety#164` (Keycast `allowed_kinds` scope), `support-trust-safety#184` (support-identity split), `divine-moderation-service#169` (Zendesk-reply → NIP-17 bridge) — cross-repo dependencies.
- `divine-web#474` / commit `9d77c523` (reconciled web's support destination to the moderation pin); `divine-web#307` (OPEN; moderation messages not showing on web — the mirrored surface may be partly broken).

realmeylisdev:
Closes #6283

## Description

Adds a pinned "Divine Moderation" row to the Messages inbox that opens or starts a NIP-17 conversation with `moderation@divine.video`, matching the first-row support entry divine-web already ships. The row ships on by default.

The pin is composed in `ConversationListBloc`, after the inbox safety filters, rather than synthesized in the widget. That keeps the row aligned with the blocklist filter and the #176 protected-minor gate, and lets the list, unread badge, and Message Requests page share the same repository partitioning.

## Behavior

- Current moderation-key threads are adopted as the pinned row and removed from the ordinary inbox/request lists, so Divine Moderation does not render twice.
- If no current-key thread exists, the bloc synthesizes a read pinned row that opens the current moderation key.
- Pre-rotation moderation threads remain visible in their original inbox or request location. They are never adopted as the pin, because replying through a retired-key conversation would send to a key the team no longer watches, but this PR also does not hide the user's only in-app entry point to that history. #6416 remains the follow-up for a proper archived/read-only presentation.
- The pinned row respects search and unread filters in the view, so it does not appear as a false match.
- Adopted current-key rows keep the existing long-press action sheet; synthetic rows do not advertise actions that have no backing database row.

## Included fixes

- Rebuilds the Messages pane as one `CustomScrollView`, with only the All/Unread chips pinned, so keyboard and small-screen layouts are not squeezed by fixed-height chrome.
- Restores the requests-only empty state to `MessageRequestsBanner + InboxEmptyState`, rather than showing unread-filter copy below a requests banner.
- Opts the pinned All/Unread chip row out of text scaling so the fixed sliver extent cannot clip the chips at larger system font scales.
- Fixes the support-row route extra so tapping the pin opens moderation, not self.
- Renders the bundled Divine wordmark for moderation instead of the remote kind-0 SVG that compiled to a black avatar.
- Consolidates moderation pubkey/NIP-05 constants in `official_accounts.dart` and has `ModerationLabelService` reference the shared source.
- Fixes the latent `onLongPressHint` semantics bug where rows without a handler advertised a long-press action.

## Out of Scope

- Archived/read-only UI for retired moderation threads: #6416.
- Staffing/SLA for the moderation inbox: `support-trust-safety#194`.
- Pre-existing report-flow issue where a blocked moderation DM is reported as success: #6387.

## Verification

- `dart format mobile/packages/dm_repository/lib/src/dm_repository.dart mobile/packages/dm_repository/test/src/dm_repository_test.dart mobile/lib/screens/inbox/widgets/unread_filter_chips.dart mobile/lib/screens/inbox/inbox_view.dart mobile/test/blocs/dm/conversation_list/conversation_list_bloc_test.dart mobile/test/screens/inbox/inbox_view_test.dart`
- `flutter analyze lib/screens/inbox/inbox_view.dart lib/screens/inbox/widgets/unread_filter_chips.dart test/screens/inbox/inbox_view_test.dart test/blocs/dm/conversation_list/conversation_list_bloc_test.dart`
- `flutter analyze lib/src/dm_repository.dart test/src/dm_repository_test.dart` from `mobile/packages/dm_repository`
- `flutter test test/screens/inbox/inbox_view_test.dart`
- `flutter test test/blocs/dm/conversation_list/conversation_list_bloc_test.dart`
- `flutter test test/src/dm_repository_test.dart` from `mobile/packages/dm_repository`

## Type of Change

### DIFF
--- a/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
+++ b/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
@@ -30,7 +30,6 @@ class ConversationListBloc
     ProtectedMinorInboxGate? protectedMinorInboxGate,
     Duration recomputeDebounce = _defaultRecomputeDebounce,
     String? supportRowPubkey,
-    List<String> supportRowLegacyPubkeys = const [],
   }) : _dmRepository = dmRepository,
        _followRepository = followRepository,
        _blocklistRepository = contentBlocklistRepository,
@@ -38,7 +37,6 @@ class ConversationListBloc
        _protectedMinorInboxGate = protectedMinorInboxGate,
        _recomputeDebounce = recomputeDebounce,
        _supportRowPubkey = supportRowPubkey,
-       _supportRowLegacyPubkeys = supportRowLegacyPubkeys,
        super(const ConversationListState()) {
     on<ConversationListStarted>(_onStarted, transformer: restartable());
     on<ConversationListLoadMore>(_onLoadMore, transformer: droppable());
@@ -79,12 +77,6 @@ class ConversationListBloc
   /// Moderation pubkey the pinned support row targets, injected rather than
   /// imported so this bloc stays free of app-layer config (#6283).
   final String? _supportRowPubkey;
-
-  /// Historical moderation pubkeys. Used for de-duplication only, never as a
-  /// send target: a user who messaged moderation before the #2321 key rotation
-  /// still holds that thread, and without this it would render beside the
-  /// pinned row as a second "Divine Moderation" entry.
-  final List<String> _supportRowLegacyPubkeys;
 
   /// Window over which bursty conversation writes are coalesced before the
   /// list is re-composed. The combined stream re-runs `classifyPotentialRequests`
@@ -490,8 +482,13 @@ class ConversationListBloc
   /// the pin renders no timestamp and is not sorted with the list.
   static const _pinnedSupportEpoch = 0;
 
-  /// Extract the pinned Divine Moderation row and remove every conversation it
-  /// stands in for from the lists they would otherwise render in (#6283).
+  /// Extract the pinned Divine Moderation row, lifting the thread it adopts out
+  /// of the list it would otherwise render in (#6283).
+  ///
+  /// Only the thread on the CURRENT moderation key is lifted. A thread on a
+  /// key the account has rotated away from stays put and renders as an ordinary
+  /// row, so its enforcement history keeps an in-app entry point until #6416
+  /// gives it an archived read-only presentation.
   ///
   /// Called with lists that have ALREADY passed the blocklist filter and the
   /// protected-minor gate, so an existing moderation thread that those filters
@@ -514,7 +511,6 @@ class ConversationListBloc
     final split = DmRepository.extractPinnedSupport(
       userPubkey: userPubkey,
       supportPubkey: supportPubkey,
-      legacyPubkeys: _supportRowLegacyPubkeys,
       inbox: inbox,
       requests: requests,
     );
--- a/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
+++ b/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
@@ -200,26 +200,10 @@ class ConversationListBloc
           isFollowing: _followRepository.isFollowing,
         );
 
-        // While the one-time history-recovery drain is still running
-        // (post-reinstall window), HOLD BACK the conversations that would
-        // classify as requests. Until the drain re-ingests the user's own
-        // message for a chat, a previously-accepted conversation is
-        // indistinguishable from a genuine request — both have
-        // currentUserHasSent=false and an unfollowed peer. Showing them as
-        // requests is the original #5304 bug; showing them in the inbox makes
-        // real requests flash there and then jump to the Requests tab once
-        // recovery completes. So during recovery we surface only the
-        // unambiguous chats (accepted + followed), backed by the restore
-        // indicator, and let the ambiguous ones settle into their correct
-        // bucket as soon as recovery completes. See #5304.
-        final recoveryComplete = _dmRepository.isHistoryRecoveryComplete;
         final inboxConversations = DmRepository.mergeAndSort(
           data.accepted,
           split.followed,
         );
-        final requests = recoveryComplete
-            ? split.requests
-            : const <DmConversation>[];
 
         final blocklistedInbox =
             _blocklistRepository?.filterBlockedConversations(
@@ -227,12 +211,15 @@ class ConversationListBloc
               userPubkey: userPubkey,
             ) ??
             inboxConversations;
+        // Filters the FULL request set, not the recovery-held-back one, so the
+        // pin below is extracted from — and inherits the filters of — every
+        // request regardless of the drain. The hold-back is applied after.
         final blocklistedRequests =
             _blocklistRepository?.filterBlockedConversations(
-              requests,
+              split.requests,
               userPubkey: userPubkey,
             ) ??
-            requests;
+            split.requests;
 
         // Protected-minor inbound filter (#176): hide conversations whose
         // counterparty (all non-self participants) is not an approved official
@@ -260,6 +247,29 @@ class ConversationListBloc
           requests: visibleRequests,
         );
 
+        // While the one-time history-recovery drain is still running
+        // (post-reinstall window), HOLD BACK the conversations that would
+        // classify as requests. Until the drain re-ingests the user's own
+        // message for a chat, a previously-accepted conversation is
+        // indistinguishable from a genuine request — both have
+        // currentUserHasSent=false and an unfollowed peer. Showing them as
+        // requests is the original #5304 bug; showing them in the inbox makes
+        // real requests flash there and then jump to the Requests tab once
+        // recovery completes. So during recovery we surface only the
+        // unambiguous chats (accepted + followed), backed by the restore
+        // indicator, and let the ambiguous ones settle into their correct
+        // bucket as soon as recovery completes. See #5304.
+        //
+        // The pin is deliberately extracted ABOVE this line. It is a fixed
+        // identity in a fixed position, not something the user has to triage
+        // into a bucket, so it cannot produce the flash the hold-back exists
+        // to prevent — withholding it would only blank its unread dot for the
+        // duration of the drain, and the badge counts it either way (#4976).
+        final recoveryComplete = _dmRepository.isHistoryRecoveryComplete;
+        final requestConversations = recoveryComplete
+            ? pin.requests
+            : const <DmConversation>[];
+
         // A conversation can stream in while a name search is active whose
         // counterparty was not resolved when the search fired; it would fall
         // back to the generated name and drop out of the results until the
@@ -285,8 +295,15 @@ class ConversationListBloc
           status: ConversationListStatus.loaded,
           conversations: pin.inbox,
           // Only true when the gate is actually costing the user something —
-          // a closed gate with nothing to hold back needs no banner.
-          requestsWithheld: !recoveryComplete && split.requests.isNotEmpty,
+          // a closed gate with nothing to hold back needs no banner. Reads the
+          // post-filter, post-pin set rather than the raw `split.requests`
+          // #6431 emitted: that reflected main's ordering, where the hold-back
+          // ran before both filters, so a filtered list simply did not exist
+          // while the gate was shut. Here it does, and a request the blocklist,
+          // the #176 gate, or the pin itself already removed is not something
+          // the gate is costing the user — least of all the moderation thread,
+          // which is on screen as the pin the whole time.
+          requestsWithheld: !recoveryComplete && pin.requests.isNotEmpty,
           visibleConversations: _computeVisible(
             pin.inbox,
             unreadOnly: state.unreadOnly,
@@ -295,12 +312,12 @@ class ConversationListBloc
             userPubkey: userPubkey,
             limit: state.currentLimit,
           ),
-          requestConversations: pin.requests,
+          requestConversations: requestConversations,
           potentialRequests: data.potentialRequests,
           hasMore: pin.inbox.length > state.currentLimit,
           isRestoringHistory: data.isRestoring,
-          pinnedConversation: pin.pinned,
-          clearPinnedConversation: pin.pinned == null,
+          pinnedSupport: pin.pinned,
+          clearPinnedSupport: pin.pinned == null,
         );
       },
       onError: (error, stackTrace) {
@@ -473,8 +490,8 @@ class ConversationListBloc
   /// the pin renders no timestamp and is not sorted with the list.
   static const _pinnedSupportEpoch = 0;
 
-  /// Extract the pinned Divine Moderation conversation and remove it from the
-  /// lists it would otherwise render in (#6283).
+  /// Extract the pinned Divine Moderation row and remove every conversation it
+  /// stands in for from the lists they would otherwise render in (#6283).
   ///
   /// Called with lists that have ALREADY passed the blocklist filter and the
   /// protected-minor gate, so an existing moderation thread that those filters
@@ -484,7 +501,7 @@ class ConversationListBloc
   /// — would get a row that `ConversationPage`'s route guard bounces straight
   /// back to the inbox.
   ({
-    DmConversation? pinned,
+    PinnedSupport? pinned,
     List<DmConversation> inbox,
     List<DmConversation> requests,
   })
@@ -494,50 +511,32 @@ class ConversationListBloc
     required List<DmConversation> requests,
   }) {
     final supportPubkey = _supportRowPubkey;
-    if (supportPubkey == null || supportPubkey.isEmpty || userPubkey.isEmpty) {
-      return (pinned: null, inbox: inbox, requests: requests);
-    }
+    final split = DmRepository.extractPinnedSupport(
+      userPubkey: userPubkey,
+      supportPubkey: supportPubkey,
+      legacyPubkeys: _supportRowLegacyPubkeys,
+      inbox: inbox,
+      requests: requests,
+    );
 
-    // Every key the support thread may live under: the current moderation
-    // pubkey plus any it has rotated away from.
-    final knownIds = <String>{
-      for (final pk in [supportPubkey, ..._supportRowLegacyPubkeys])
-        if (pk.isNotEmpty) DmRepository.computeConversationId([userPubkey, pk]),
-    };
-
-    DmConversation? real;
-    final remainingInbox = <DmConversation>[];
-    for (final c in inbox) {
-      if (real == null && knownIds.contains(c.id)) {
-        real = c;
-      } else {
-        remainingInbox.add(c);
-      }
-    }
-
-    // An inbound-only moderation DM (the team wrote first) has
-    // currentUserHasSent == false and lands in requests, not the inbox.
-    // Without this the same thread would be reachable from both the pinned row
-    // and the Requests page with independent unread accounting.
-    final remainingRequests = <DmConversation>[];
-    for (final c in requests) {
-      if (real == null && knownIds.contains(c.id)) {
-        real = c;
-      } else {
-        remainingRequests.add(c);
-      }
-    }
-
-    if (real != null) {
+    final adopted = split.pinned;
+    if (adopted != null) {
       return (
-        pinned: real,
-        inbox: remainingInbox,
-        requests: remainingRequests,
+        pinned: PinnedSupport(conversation: adopted, isPersisted: true),
+        inbox: split.inbox,
+        requests: split.requests,
       );
     }
 
+    // Null means no pin is possible at all — no identity to key on, or the
+    // signed-in user IS the moderation account. Nothing to synthesize against.
+    final supportConversationId = split.supportConversationId;
+    if (supportConversationId == null || supportPubkey == null) {
+      return (pinned: null, inbox: split.inbox, requests: split.requests);
+    }
+
     final synthetic = DmConversation(
-      id: DmRepository.computeConversationId([userPubkey, supportPubkey]),
+      id: supportConversationId,
       participantPubkeys: [userPubkey, supportPubkey],
       isGroup: false,
       createdAt: _pinnedSupportEpoch,
@@ -553,9 +552,11 @@ class ConversationListBloc
         allowed;
 
     return (
-      pinned: visible.isEmpty ? null : visible.first,
-      inbox: remainingInbox,
-      requests: remainingRequests,
+      pinned: visible.isEmpty
+          ? null
+          : PinnedSupport(conversation: visible.first, isPersisted: false),
+      inbox: split.inbox,
+      requests: split.requests,
     );
   }
 
--- a/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
+++ b/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
@@ -31,7 +31,6 @@ class ConversationListBloc
     Duration recomputeDebounce = _defaultRecomputeDebounce,
     String? supportRowPubkey,
     List<String> supportRowLegacyPubkeys = const [],
-    bool supportRowEnabled = false,
   }) : _dmRepository = dmRepository,
        _followRepository = followRepository,
        _blocklistRepository = contentBlocklistRepository,
@@ -40,7 +39,6 @@ class ConversationListBloc
        _recomputeDebounce = recomputeDebounce,
        _supportRowPubkey = supportRowPubkey,
        _supportRowLegacyPubkeys = supportRowLegacyPubkeys,
-       _supportRowEnabled = supportRowEnabled,
        super(const ConversationListState()) {
     on<ConversationListStarted>(_onStarted, transformer: restartable());
     on<ConversationListLoadMore>(_onLoadMore, transformer: droppable());
@@ -87,8 +85,6 @@ class ConversationListBloc
   /// still holds that thread, and without this it would render beside the
   /// pinned row as a second "Divine Moderation" entry.
   final List<String> _supportRowLegacyPubkeys;
-
-  final bool _supportRowEnabled;
 
   /// Window over which bursty conversation writes are coalesced before the
   /// list is re-composed. The combined stream re-runs `classifyPotentialRequests`
@@ -498,10 +494,7 @@ class ConversationListBloc
     required List<DmConversation> requests,
   }) {
     final supportPubkey = _supportRowPubkey;
-    if (!_supportRowEnabled ||
-        supportPubkey == null ||
-        supportPubkey.isEmpty ||
-        userPubkey.isEmpty) {
+    if (supportPubkey == null || supportPubkey.isEmpty || userPubkey.isEmpty) {
       return (pinned: null, inbox: inbox, requests: requests);
     }
 
--- a/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
+++ b/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
@@ -29,12 +29,18 @@ class ConversationListBloc
     ProfileRepository? profileRepository,
     ProtectedMinorInboxGate? protectedMinorInboxGate,
     Duration recomputeDebounce = _defaultRecomputeDebounce,
+    String? supportRowPubkey,
+    List<String> supportRowLegacyPubkeys = const [],
+    bool supportRowEnabled = false,
   }) : _dmRepository = dmRepository,
        _followRepository = followRepository,
        _blocklistRepository = contentBlocklistRepository,
        _profileRepository = profileRepository,
        _protectedMinorInboxGate = protectedMinorInboxGate,
        _recomputeDebounce = recomputeDebounce,
+       _supportRowPubkey = supportRowPubkey,
+       _supportRowLegacyPubkeys = supportRowLegacyPubkeys,
+       _supportRowEnabled = supportRowEnabled,
        super(const ConversationListState()) {
     on<ConversationListStarted>(_onStarted, transformer: restartable());
     on<ConversationListLoadMore>(_onLoadMore, transformer: droppable());
@@ -71,6 +77,18 @@ class ConversationListBloc
   /// the bloc. Mirrors `NotificationBadgeCubit.setRepository`.
   ProfileRepository? _profileRepository;
   final ProtectedMinorInboxGate? _protectedMinorInboxGate;
+
+  /// Moderation pubkey the pinned support row targets, injected rather than
+  /// imported so this bloc stays free of app-layer config (#6283).
+  final String? _supportRowPubkey;
+
+  /// Historical moderation pubkeys. Used for de-duplication only, never as a
+  /// send target: a user who messaged moderation before the #2321 key rotation
+  /// still holds that thread, and without this it would render beside the
+  /// pinned row as a second "Divine Moderation" entry.
+  final List<String> _supportRowLegacyPubkeys;
+
+  final bool _supportRowEnabled;
 
   /// Window over which bursty conversation writes are coalesced before the
   /// list is re-composed. The combined stream re-runs `classifyPotentialRequests`
@@ -237,6 +255,15 @@ class ConversationListBloc
             ) ??
             blocklistedRequests;
 
+        // Lift the moderation thread out of the list into the pinned row, so
+        // the inbox never renders it twice (#6283). Runs after both filters so
+        // the pin inherits them rather than bypassing them.
+        final pin = _extractPinnedSupport(
+          userPubkey: userPubkey,
+          inbox: visibleInbox,
+          requests: visibleRequests,
+        );
+
         // A conversation can stream in while a name search is active whose
         // counterparty was not resolved when the search fired; it would fall
         // back to the generated name and drop out of the results until the
@@ -246,7 +273,7 @@ class ConversationListBloc
         // re-firing for those would loop on every recompute.
         if (state.searchQuery.length >= minSearchQueryLength) {
           final known = state.conversations.map((c) => c.id).toSet();
-          final hasNewUnresolved = visibleInbox.any(
+          final hasNewUnresolved = pin.inbox.any(
             (c) =>
                 !known.contains(c.id) &&
                 !state.profileNames.containsKey(
@@ -260,22 +287,24 @@ class ConversationListBloc
 
         return state.copyWith(
           status: ConversationListStatus.loaded,
-          conversations: visibleInbox,
+          conversations: pin.inbox,
           // Only true when the gate is actually costing the user something —
           // a closed gate with nothing to hold back needs no banner.
           requestsWithheld: !recoveryComplete && split.requests.isNotEmpty,
           visibleConversations: _computeVisible(
-            visibleInbox,
+            pin.inbox,
             unreadOnly: state.unreadOnly,
             query: state.searchQuery,
             profileNames: state.profileNames,
             userPubkey: userPubkey,
             limit: state.currentLimit,
           ),
-          requestConversations: visibleRequests,
+          requestConversations: pin.requests,
           potentialRequests: data.potentialRequests,
-          hasMore: visibleInbox.length > state.currentLimit,
+          hasMore: pin.inbox.length > state.currentLimit,
           isRestoringHistory: data.isRestoring,
+          pinnedConversation: pin.pinned,
+          clearPinnedConversation: pin.pinned == null,
         );
       },
       onError: (error, stackTrace) {
@@ -439,6 +468,101 @@ class ConversationListBloc
     return conversation.participantPubkeys.firstWhere(
       (pk) => pk != self,
       orElse: () => conversation.participantPubkeys.first,
+    );
+  }
+
+  /// A synthetic pin carries no timestamp, so it needs a `createdAt` that is
+  /// stable across recomputes — a clock read would make every debounce tick
+  /// produce an unequal state and re-emit forever. The value is never shown:
+  /// the pin renders no timestamp and is not sorted with the list.
+  static const _pinnedSupportEpoch = 0;
+
+  /// Extract the pinned Divine Moderation conversation and remove it from the
+  /// lists it would otherwise render in (#6283).
+  ///
+  /// Called with lists that have ALREADY passed the blocklist filter and the
+  /// protected-minor gate, so an existing moderation thread that those filters
+  /// removed simply is not found here. The synthetic fallback is pushed back
+  /// through both filters for the same reason: without that, a user who blocked
+  /// the moderation account — or a restricted minor whose approval was revoked
+  /// — would get a row that `ConversationPage`'s route guard bounces straight
+  /// back to the inbox.
+  ({
+    DmConversation? pinned,
+    List<DmConversation> inbox,
+    List<DmConversation> requests,
+  })
+  _extractPinnedSupport({
+    required String userPubkey,
+    required List<DmConversation> inbox,
+    required List<DmConversation> requests,
+  }) {
+    final supportPubkey = _supportRowPubkey;
+    if (!_supportRowEnabled ||
+        supportPubkey == null ||
+        supportPubkey.isEmpty ||
+        userPubkey.isEmpty) {
+      return (pinned: null, inbox: inbox, requests: requests);
+    }
+
+    // Every key the support thread may live under: the current moderation
+    // pubkey plus any it has rotated away from.
+    final knownIds = <String>{
+      for (final pk in [supportPubkey, ..._supportRowLegacyPubkeys])
+        if (pk.isNotEmpty) DmRepository.computeConversationId([userPubkey, pk]),
+    };
+
+    DmConversation? real;
+    final remainingInbox = <DmConversation>[];
+    for (final c in inbox) {
+      if (real == null && knownIds.contains(c.id)) {
+        real = c;
+      } else {
+        remainingInbox.add(c);
+      }
+    }
+
+    // An inbound-only moderation DM (the team wrote first) has
+    // currentUserHasSent == false and lands in requests, not the inbox.
+    // Without this the same thread would be reachable from both the pinned row
+    // and the Requests page with independent unread accounting.
+    final remainingRequests = <DmConversation>[];
+    for (final c in requests) {
+      if (real == null && knownIds.contains(c.id)) {
+        real = c;
+      } else {
+        remainingRequests.add(c);
+      }
+    }
+
+    if (real != null) {
+      return (
+        pinned: real,
+        inbox: remainingInbox,
+        requests: remainingRequests,
+      );
+    }
+
+    final synthetic = DmConversation(
+      id: DmRepository.computeConversationId([userPubkey, supportPubkey]),
+      participantPubkeys: [userPubkey, supportPubkey],
+      isGroup: false,
+      createdAt: _pinnedSupportEpoch,
+    );
+
+    final allowed =
+        _blocklistRepository?.filterBlockedConversations([
+          synthetic,
+        ], userPubkey: userPubkey) ??
+        [synthetic];
+    final visible =
+        _protectedMinorInboxGate?.filter(allowed, userPubkey: userPubkey) ??
+        allowed;
+
+    return (
+      pinned: visible.isEmpty ? null : visible.first,
+      inbox: remainingInbox,
+      requests: remainingRequests,
     );
   }
 
--- a/mobile/lib/config/official_accounts.dart
+++ b/mobile/lib/config/official_accounts.dart
@@ -23,26 +23,50 @@ class OfficialAccount {
   });
 }
 
+/// Values [OfficialAccount.role] can take. Named so no caller has to spell the
+/// role as a bare string to pick an account out of [kPinnedOfficialAccounts].
+abstract class OfficialAccountRole {
+  static const String hq = 'hq';
+  static const String moderation = 'moderation';
+}
+
+/// Divine HQ's pinned identity.
+const OfficialAccount kHqAccount = OfficialAccount(
+  pubkeyHex: 'c4a39f1291291d452405cd8ddd798c4a29a3858c52cd0d843f1f6852cf17682e',
+  nip05: '_@divinehq.divine.video',
+  role: OfficialAccountRole.hq,
+  minorContactable: true,
+);
+
+/// The current Divine moderation pubkey — the report target, the pinned support
+/// row's destination, and `ModerationLabelService`'s NIP-05 fallback.
+///
+/// Single-sourced here rather than re-declared per consumer: the pinned support
+/// row and the protected-minor gate must anchor on the SAME key, or the row can
+/// point somewhere the gate would not approve.
+const String kModerationPubkeyHex =
+    '8fd5eb6d8f362163bc00a5ab6b4a3167dbf32d00ec4efdbcf43b3c9514433b7e';
+
+/// Canonical NIP-05 identifier for [kModerationPubkeyHex]. Live resolution must
+/// still map back to that pubkey for the account to count as reachable.
+const String kModerationNip05 = 'moderation@divine.video';
+
+/// The Divine moderation account's pinned identity.
+const OfficialAccount kModerationAccount = OfficialAccount(
+  pubkeyHex: kModerationPubkeyHex,
+  nip05: kModerationNip05,
+  role: OfficialAccountRole.moderation,
+  minorContactable: true,
+);
+
 /// The pinned child-contactable set (verified live 2026-07-07). Additions are
 /// release-gated by design — this is a child-contact list, and requiring an app
 /// release to add a key is the accepted friction that makes the pin an
 /// attacker-addition barrier. Each entry pins its OWN canonical identifier form
 /// (HQ uses a subdomain origin, moderation the classic form).
 const List<OfficialAccount> kPinnedOfficialAccounts = [
-  OfficialAccount(
-    pubkeyHex:
-        'c4a39f1291291d452405cd8ddd798c4a29a3858c52cd0d843f1f6852cf17682e',
-    nip05: '_@divinehq.divine.video',
-    role: 'hq',
-    minorContactable: true,
-  ),
-  OfficialAccount(
-    pubkeyHex:
-        '8fd5eb6d8f362163bc00a5ab6b4a3167dbf32d00ec4efdbcf43b3c9514433b7e',
-    nip05: 'moderation@divine.video',
-    role: 'moderation',
-    minorContactable: true,
-  ),
+  kHqAccount,
+  kModerationAccount,
 ];
 
 /// Moderation pubkeys the account has rotated away from.
@@ -61,10 +85,9 @@ const List<String> kLegacyModerationPubkeys = [
 /// Whether [pubkeyHex] is the Divine moderation account, current or retired.
 ///
 /// Retired keys count: a thread opened before a rotation stays keyed on the
-/// old pubkey, and it is the same team on the other end of it.
+/// old pubkey, and it is the same team on the other end of it. Callers that
+/// need a *send target* must use [kModerationPubkeyHex] instead — this answers
+/// "is this the moderation team", not "where do replies go".
 bool isModerationAccount(String pubkeyHex) =>
-    kPinnedOfficialAccounts.any(
-      (account) =>
-          account.role == 'moderation' && account.pubkeyHex == pubkeyHex,
-    ) ||
+    pubkeyHex == kModerationPubkeyHex ||
     kLegacyModerationPubkeys.contains(pubkeyHex);
--- a/mobile/lib/config/official_accounts.dart
+++ b/mobile/lib/config/official_accounts.dart
@@ -71,10 +71,16 @@ const List<OfficialAccount> kPinnedOfficialAccounts = [
 
 /// Moderation pubkeys the account has rotated away from.
 ///
-/// A DM thread opened before a rotation stays keyed on the old pubkey, so the
-/// pinned support row de-duplicates against these to avoid rendering beside an
-/// orphaned "Divine Moderation" conversation. Never a send target — messages
-/// always go to the current pin above.
+/// A DM thread opened before a rotation stays keyed on the old pubkey. Those
+/// threads are deliberately NOT folded into the pinned support row: the row
+/// routes on its own participants all the way to `sendMessage`, so adopting one
+/// would address replies to a key nobody reads. They stay where they are and
+/// render as ordinary rows, keeping their enforcement history reachable until
+/// #6416 gives it an archived read-only presentation.
+///
+/// Never a send target — messages always go to the current pin above. Used for
+/// recognising the account ([isModerationAccount], which drives the bundled
+/// avatar) and for the `ModerationLabelService` subscription migration.
 ///
 /// `121b915b…` was retired in #2321 (2026-03-20); `ModerationLabelService`
 /// carries the matching one-time migration for labeler subscriptions.
--- a/mobile/lib/config/official_accounts.dart
+++ b/mobile/lib/config/official_accounts.dart
@@ -57,3 +57,14 @@ const List<OfficialAccount> kPinnedOfficialAccounts = [
 const List<String> kLegacyModerationPubkeys = [
   '121b915baba659cbe59626a8afaf83b01dc42354dfecaad9d465d51bb5715d72',
 ];
+
+/// Whether [pubkeyHex] is the Divine moderation account, current or retired.
+///
+/// Retired keys count: a thread opened before a rotation stays keyed on the
+/// old pubkey, and it is the same team on the other end of it.
+bool isModerationAccount(String pubkeyHex) =>
+    kPinnedOfficialAccounts.any(
+      (account) =>
+          account.role == 'moderation' && account.pubkeyHex == pubkeyHex,
+    ) ||
+    kLegacyModerationPubkeys.contains(pubkeyHex);
--- a/mobile/lib/config/official_accounts.dart
+++ b/mobile/lib/config/official_accounts.dart
@@ -44,3 +44,16 @@ const List<OfficialAccount> kPinnedOfficialAccounts = [
     minorContactable: true,
   ),
 ];
+
+/// Moderation pubkeys the account has rotated away from.
+///
+/// A DM thread opened before a rotation stays keyed on the old pubkey, so the
+/// pinned support row de-duplicates against these to avoid rendering beside an
+/// orphaned "Divine Moderation" conversation. Never a send target — messages
+/// always go to the current pin above.
+///
+/// `121b915b…` was retired in #2321 (2026-03-20); `ModerationLabelService`
+/// carries the matching one-time migration for labeler subscriptions.
+const List<String> kLegacyModerationPubkeys = [
+  '121b915baba659cbe59626a8afaf83b01dc42354dfecaad9d465d51bb5715d72',
+];
--- a/mobile/lib/screens/inbox/inbox_view.dart
+++ b/mobile/lib/screens/inbox/inbox_view.dart
@@ -457,20 +457,20 @@ class _RestorePausedBannerGate extends StatelessWidget {
 /// Ordering is not left to the list's recency sort — a synthetic row carries
 /// no timestamp and would sink out of reach.
 ///
-/// The caller only builds this when
-/// [ConversationListState.pinnedConversation] is non-null: the bloc composes
-/// that inside the same pipeline that applies the blocklist filter and the
-/// protected-minor gate, so a user who blocked the moderation account, or a
-/// restricted minor whose approval was revoked, gets no row rather than one
-/// the conversation route guard would bounce.
+/// The caller only builds this when [ConversationListState.pinnedSupport] is
+/// non-null: the bloc composes that inside the same pipeline that applies the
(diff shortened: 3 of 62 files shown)


realmeylisdev:
Simulator Screenshot - iPhone 17 Pro Max - 2026-07-25 at 18 30 11



Simulator Screenshot - iPhone 17 Pro Max - 2026-07-25 at 18 30 13

mbradley (commented):
### Code review

Two questions on the pin's handling of a *real* adopted thread. Both are judgment calls rather than clear fixes, so I'm flagging rather than patching, and the draft branch is read-only per the review runbook either way. The synthetic-pin path (the common case) looks right to me: gate-then-extract ordering matches what #5754 established, and the filters are correctly re-applied to the synthesized row.

**1. An adopted legacy-keyed thread becomes the send target, which the new config comment says must never happen.**

`_extractPinnedSupport` builds `knownIds` from the current moderation pubkey plus `kLegacyModerationPubkeys`, then returns the first match verbatim as the pin (`real = c`):

https://github.com/divinevideo/divine-mobile/blob/0a4c99f5ea0f6adbe777a6fdf8fe2bb2e63e7b31/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart#L508-L516

The row then routes on that conversation's own participants:

https://github.com/divinevideo/divine-mobile/blob/0a4c99f5ea0f6adbe777a6fdf8fe2bb2e63e7b31/mobile/lib/screens/inbox/inbox_view.dart#L498-L506

and the composer wires the route's counterparty list straight into the outbound message, with no remap anywhere:

https://github.com/divinevideo/divine-mobile/blob/0a4c99f5ea0f6adbe777a6fdf8fe2bb2e63e7b31/mobile/lib/screens/inbox/conversation/conversation_view.dart#L392-L396

So for a user whose only moderation thread predates the 2026-03-20 rotation, the pin adopts the legacy thread and a new message goes to the retired `121b915baba659cbe59626a8afaf83b01dc42354dfecaad9d465d51bb5715d72`, which the constant's own doc forbids:

https://github.com/divinevideo/divine-mobile/blob/0a4c99f5ea0f6adbe777a6fdf8fe2bb2e63e7b31/mobile/lib/config/official_accounts.dart#L51-L54

The call: should adopting a legacy thread carry its history forward for display but still send to the current pubkey (dedup-and-retarget), or should legacy matches only suppress the duplicate row, with the pin always opening a fresh thread on the current key? The first keeps continuity, the second is simpler and guarantees deliverability.

Related, same function: both loops guard on `real == null`, so only the first match across `inbox` + `requests` is extracted. A user with *both* a pre-rotation and a post-rotation thread keeps the second one in the ordinary list, which is the duplicate row the dedup exists to prevent. Whichever way the above lands, that case looks worth a test.

**2. An adopted requests-bucket thread can show an unread dot that neither pill counts.**

An inbound-only, non-followed moderation thread (team writes first) classifies into `requests`. The extraction pulls it out of `requestConversations`, so `requestUnreadCount` no longer sees it:

https://github.com/divinevideo/divine-mobile/blob/0a4c99f5ea0f6adbe777a6fdf8fe2bb2e63e7b31/mobile/lib/blocs/dm/conversation_list/conversation_list_state.dart#L115-L118

But `DmUnreadCountCubit` composes only `accepted ∪ split.followed` and never counts `split.requests`, so the Messages pill doesn't pick it up either:

https://github.com/divinevideo/divine-mobile/blob/0a4c99f5ea0f6adbe777a6fdf8fe2bb2e63e7b31/mobile/lib/blocs/dm/unread_count/dm_unread_count_cubit.dart#L206-L212

Net: the pinned row renders its unread dot while both pills read zero. That's the dots-vs-badge divergence #4976 fixed, and this cubit documents parity as structural: "the badge must equal the unread conversations the Messages list actually renders ... so the count cannot drift from the list."

The call: teach `DmUnreadCountCubit` about the pin (it would need the same flag and pubkeys, so parity stays structural), or leave the requests-bucket thread in place and let the pin coexist with the Requests entry, or accept the divergence while the flag is off and note it. Worth deciding before the flag flips, since #6283's premise is that this row is how users reach support.

Minor, non-blocking: the `/// Switches between loading, error, empty, and conversation list states.` doc line now sits above `_PinnedSupportRow` rather than `_ConversationListContent`, which lost its comment, because the new class was inserted between them ([inbox_view.dart#L459-L462](https://github.com/divinevideo/divine-mobile/blob/0a4c99f5ea0f6adbe777a6fdf8fe2bb2e63e7b31/mobile/lib/screens/inbox/inbox_view.dart#L459-L462)).

Nice work composing this inside the pipeline instead of the widget. That decision is what makes both safety gates free rather than something to re-derive, and the writeup of why made this much faster to review. Thanks, as always.

mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
```diff
@@ -435,6 +464,101 @@ class ConversationListBloc
...
+    // pubkey plus any it has rotated away from.
+    final knownIds = <String>{
+      for (final pk in [supportPubkey, ..._supportRowLegacyPubkeys])
+        if (pk.isNotEmpty) DmRepository.computeConversationId([userPubkey, pk]),
+    };
+
+    DmConversation? real;
+    final remainingInbox = <DmConversation>[];
+    for (final c in inbox) {
+      if (real == null && knownIds.contains(c.id)) {
```
NotThatKindOfDrLiz: This adopts the first matching current-or-legacy moderation conversation verbatim. For a legacy-keyed thread, `_PinnedSupportRow` later routes using that conversation's participants, so the retired `121b915baba659cbe59626a8afaf83b01dc42354dfecaad9d465d51bb5715d72` becomes the send target despite `kLegacyModerationPubkeys` documenting that legacy keys are never send targets. Please either retarget adopted legacy history to the current moderation key or only use legacy matches for de-dupe while opening the pin on the current key. Also remove every known support conversation ID, not only the first match, so users with both legacy and current threads don't get a duplicate ordinary row.

mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
```diff
@@ -435,6 +464,101 @@ class ConversationListBloc
...
+        real = c;
+      } else {
+        remainingInbox.add(c);
+      }
+    }
+
+    // An inbound-only moderation DM (the team wrote first) has
+    // currentUserHasSent == false and lands in requests, not the inbox.
+    // Without this the same thread would be reachable from both the pinned row
```
NotThatKindOfDrLiz: This only holds for the inbox's bloc instance. `MessageRequestsPage` constructs its own `ConversationListBloc` without `supportRowEnabled`, `supportRowPubkey`, or the legacy pubkeys, so an inbound moderation request is removed from the inbox banner count but still appears on the Requests page. The app-shell `DmUnreadCountCubit` also still counts only accepted/followed inbox rows, so an unread pinned request can show a dot while the Messages badge remains zero. Please make the support-row extraction/accounting consistent across those surfaces or leave the request in place deliberately.

mobile/lib/screens/inbox/inbox_view.dart
```diff
@@ -447,6 +458,56 @@ class _RestoringHistoryIndicator extends StatelessWidget {
...
+      // Fixed identity: the profile lookup yields null until the relay client
+      // is ready, and the tile's fallback is a generated name — so sourcing
+      // this from kind-0 would show a random display name at cold start.
+      displayNameOverride: context.l10n.inboxSupportRowTitle,
+      // Only stand in for the preview when there is no real thread yet; once
+      // the team replies, the last message is more useful than the blurb.
+      subtitleOverride: pinned.lastMessageContent == null
+          ? context.l10n.inboxSupportRowSubtitle
+          : null,
+      onTap: () => _pushConversation(
```
NotThatKindOfDrLiz, line 502: When `pinned` is a real adopted conversation, this row replaces the normal inbox tile but does not pass an `onLongPress` handler, so users lose the existing mute and remove actions for that thread. Please either wire the normal action sheet for adopted real threads, or record the explicit product decision in the PR that support threads become non-dismissible/non-mutable once pinned.

mobile/lib/config/official_accounts.dart
```diff
@@ -44,3 +44,16 @@ const List<OfficialAccount> kPinnedOfficialAccounts = [
...
+/// Moderation pubkeys the account has rotated away from.
+///
+/// A DM thread opened before a rotation stays keyed on the old pubkey, so the
+/// pinned support row de-duplicates against these to avoid rendering beside an
+/// orphaned "Divine Moderation" conversation. Never a send target — messages
+/// always go to the current pin above.
+///
+/// `121b915b…` was retired in #2321 (2026-03-20); `ModerationLabelService`
+/// carries the matching one-time migration for labeler subscriptions.
+const List<String> kLegacyModerationPubkeys = [
```
NotThatKindOfDrLiz, line 87: This adds another copy of the moderation pubkeys while `ModerationLabelService` already owns the current fallback and legacy migration keys. Since this PR is directly adding support-row config, please consolidate the moderation constants here and have the label service reference the shared source instead of growing parallel copies. Relatedly, avoid selecting this account elsewhere by the magic role string `'moderation'` without a shared constant.

mobile/lib/screens/inbox/inbox_view.dart
```diff
@@ -447,6 +458,56 @@ class _RestoringHistoryIndicator extends StatelessWidget {
 }
 
 /// Switches between loading, error, empty, and conversation list states.
```
NotThatKindOfDrLiz: Nit: this doc comment now describes `_ConversationListContent`, but the new `_PinnedSupportRow` class was inserted beneath it. Please move the switch-state comment back above `_ConversationListContent`.

NotThatKindOfDrLiz (changes requested):
Requesting changes before this can merge. The row is flag-gated off today, but these are blockers for any FF_SUPPORT_DM_ROW=true build: legacy-key adoption can silently send support replies to a retired pubkey, the support thread is not removed/accounted consistently across Requests/badge surfaces, and an adopted real thread loses existing row actions unless product explicitly accepts that tradeoff. Please also consolidate the moderation pubkey constants while this PR is touching the adjacent config.

hm21 (commented):
I tested it and have a few questions:

1. Why are we making this feature experimental instead of releasing it directly?
2. The leading logo is completely black, which makes it hard to see. Shouldn't we use a version with better visibility?
3. From a UX perspective, it doesn't feel quite right to make this chat the biggest point of interest on the screen. Ideally, users shouldn't need to contact us very often. My suggestion would be to treat it like any other chat and only surface it at the top when someone taps **Add new chat**. If that's the intended design, though, I'm fine with it.
4. Only the area below the message request is scrollable. On a small Android phone, that leaves a very limited scrollable area. Do we have a solution for this? For example, if the screen height is below a certain threshold, we could switch to a full sliver layout.
5. Related to 4: When I search for something, I can't see my chats anymore, and I also have a layout issue. Maybe we need to hide the content above while we're searching. 


| Default | With Keyboard |
| ------- | --------------- |
| Image |  image   |

realmeylisdev:
Screenshot 2026-07-25 at 23 47 01

realmeylisdev:
@hm21 all five — 1, 3, 4 and 5 land in the next push.

1. Flag removed, ships on by default.
2. Fixed in `be92eb5bb` (details above).
3. Moved into the chat list as its first tile, instead of a banner above the following bar. It stays first regardless of the recency sort — a thread nobody has messaged yet has no timestamp, so sorting would sink it — but under a filter it behaves like any other chat: drops out of a search its title doesn't match, and out of Unread once it's read.
4. I'd avoid a height threshold: the keyboard collapses the viewport, not the screen, and your item-5 pair is the same device with and without it — so a threshold would leave 5 broken on the phone that reported 4. The pane is one `CustomScrollView` at every size now, with only the All/Unread chips pinned. Fixed chrome goes from ~324px to 48px. (Small correction: the Message requests banner was already inside the scrollable list — the fixed part was everything above it, search field included.)
5. Taking your suggestion — the following bar collapses when the search field takes focus, and returns when it loses it. Focus rather than query text, because `minSearchQueryLength` is 2: the query is still empty after your first keystroke, which is exactly when the list has least room. The 8.4px overflow goes with the old Column.

mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
realmeylisdev: Went with the second option: retired keys de-duplicate only, and the pin always opens and sends on the current key. Fixed in `2f8aef02e9e7b0593ffa4161a73eef51b7fd4cd3`.

The deciding factor against retargeting adopted legacy history was that `ConversationPage` is keyed by conversation id, so a "display the old thread, send to the new key" row would preview *"we looked into your report"* and open an empty conversation. Continuity would have been the preview only, not the thread behind it.

Also fixed the first-match-only bug you spotted in the same pass — the partition now removes **every** known moderation conversation from both lists, and only a current-key one is returned as the pin. Both cases are tested: `keeps one pin when BOTH a legacy and a current thread exist, leaving no duplicate ordinary row` and `drops a pre-rotation legacy thread and pins the CURRENT key, so a reply can never reach the retired pubkey`.

The tradeoff worth naming: pre-rotation history is no longer reachable in-app. `_mergeDuplicateConversations` groups by peer pubkey, so the two threads were never merged and the legacy row stays in Drift with no UI entry point. Since the team retired that key and cannot answer on it, dropping a dead read-only thread seemed better than a second moderation-looking row — but say the word if you'd rather keep it visible.

mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
realmeylisdev: Both surfaces are now consistent, in `65903f456d553a4f0db184442a5a39dd78da28f7`.

`MessageRequestsPage` gets the same `supportRowPubkey` / `supportRowLegacyPubkeys` wiring, so its bloc partitions identically — which also stops "mark all read" / "remove all" sweeping a conversation the user isn't looking at.

`DmUnreadCountCubit` now calls the same partition. Rather than teach it the pin separately, I moved the pure list work into `DmRepository.extractPinnedSupport`, next to the `classifyPotentialRequests` / `mergeAndSort` statics the list and the badge already share — so parity stays structural rather than something a future edit has to remember to mirror. The adopted pin is counted back in wherever it came from, and retired-key threads stop being counted because they stop being rendered.

One knock-on I should flag rather than bury: the `#5304` recovery hold-back used to run *before* extraction, so during the post-reinstall drain the pin fell back to the synthetic row and its unread dot disappeared — which, once the badge counted it, would have been a fresh badge-over-dots divergence. Extraction now runs ahead of the hold-back. The pin is never routed into a bucket, so it can't produce the flash the hold-back exists to prevent; both filters still run first, and `withholds a recovery-window pin the protected-minor gate rejects` pins that it hasn't become a filter bypass.

The class doc's "the requests bucket never feeds this count" line is updated — it's no longer true for exactly this one conversation.

realmeylisdev:
@hm21 Could you please verify the overflow issue again on your old phone and report the results here? It would help us confirm whether the issue is still reproducible.

mobile/lib/screens/inbox/inbox_view.dart
realmeylisdev: Wired the normal action sheet — `1e60fd2eb9ce903e8eea368cbd6431fe7da4c439`. It's a real capability regression otherwise, so restoring it beat documenting it.

Long-press is attached only when the pin is an **adopted** thread. A synthetic pin has no database row, so `removeConversation` would no-op while the snackbar still said "Conversation removed"; `ConversationTile` already gates its a11y hint on the handler, so the affordance stays honest in both cases.

I checked the two actions that looked like they might be footguns here before keeping them:

- **Block** hides the pin (the blocklist filter runs before extraction — already covered by an existing test). It isn't a dead end: `safety_settings_screen.dart` lists blocked accounts with an unblock action, the same recovery path as any other blocked account.
- **Remove** deletes locally and the pin falls back to the synthetic row, which reads as "clear this conversation" rather than "dismiss support".

One thing the wiring surfaced: `_onConversationLongPressed` resolved the sheet's display name from kind-0, which for this pubkey is exactly the late-resolving lookup `displayNameOverride` exists to avoid — the sheet opened saying "Block Complicated Ptarmigan 53". It now takes the l10n title, which also drops a pointless round-trip before the sheet appears. Tested by `long-press on an adopted thread opens the actions sheet labelled Divine Moderation, never a generated profile name`.

mobile/lib/config/official_accounts.dart
realmeylisdev: Consolidated in `3383a9959bea849ca1df904906e2d3dc83ae0471`.

`official_accounts.dart` now owns `kModerationPubkeyHex` and `kModerationNip05` as top-level constants, and the `OfficialAccount` entries are named consts (`kModerationAccount`, `kHqAccount`) composed from them. `ModerationLabelService.fallbackModerationPubkeyHex` and `.divineModerationNip05` forward to those — they stay `static const` on the class because five test files use them in const contexts — and its private `_legacyModerationPubkeyHex` is gone: `_migrateLegacyPubkey` iterates `kLegacyModerationPubkeys`, so a future rotation is one constant instead of three.

NIP-05 runtime resolution is deliberately untouched. The pin has to anchor on the release-pinned value, not the resolved one, or it could point somewhere the protected-minor gate wouldn't approve — the consolidation is one-directional, service ← config.

On the magic string: `OfficialAccountRole.moderation` replaces it at all three sites. `inbox_page.dart` now reads `kModerationPubkeyHex` directly, which also retires the `firstWhereOrNull` and its nullable result.

Guarded by `test/config/official_accounts_test.dart`, including that the service's constants equal the config's and that no retired key is also the current key. One literal stays: `packages/dm_repository/test/src/dm_repository_test.dart` can't import app-layer config, and it already documents why.

mobile/lib/screens/inbox/inbox_view.dart
realmeylisdev: This one turned out to be moot rather than fixed: `9f304ec33` (the scroll-view refactor for @hm21's items 4 and 5) collapsed `_ConversationListContent`, `_ConversationList` and `_ConversationListView` into a single `_MessagesScrollView`, and deleted that doc line along with the class it described. The status switch now lives inline in the scroll view's `build`.

While in there I did rename the leftover `_ConversationListState` to `_MessagesScrollViewState` — it was the state class for `_MessagesScrollView` but collided in the reader's head with `ConversationListState`, the bloc state used in the same file.

realmeylisdev:
All four blockers are addressed, plus @mbradley's two findings — thanks both, they were the same defects seen from different angles.

| Item | Commit |
|---|---|
| Legacy-key adoption sends replies to a retired pubkey | `2f8aef02e` |
| Support thread inconsistent across Requests / badge | `65903f456` |
| Adopted thread loses its row actions | `1e60fd2eb` |
| Moderation constants declared twice | `3383a9959` |

Detail is in the thread replies. Three things worth surfacing here:

**The flag is gone, so these were live.** `3c7247c6e` (pre-dating both reviews) removed `FF_SUPPORT_DM_ROW`. Both of you scoped the blockers to "any `FF_SUPPORT_DM_ROW=true` build" — that's the default build now, so none of this was deferred behind a flag.

**One knock-on I introduced and then fixed.** Teaching the badge about the pin meant the `#5304` recovery hold-back — which ran *before* pin extraction — would blank the pin's unread dot during the post-reinstall drain while the badge counted it. Extraction now runs ahead of the hold-back. The pin is never routed into a bucket, so it can't produce the flash the hold-back prevents, and both filters still run first so an adopted pin can't bypass the blocklist or the protected-minor gate. `withholds a recovery-window pin the protected-minor gate rejects` is the test that would go red if that ordering ever inverted.

**A tradeoff that needs a nod rather than a fix.** Dedupe-only means pre-rotation moderation history is no longer reachable in-app: `_mergeDuplicateConversations` groups by peer pubkey, so the legacy thread was never merged into the current one and now has no UI entry point. It's a dead read-only thread on a key the team can't answer on, so dropping it beat showing a second moderation-looking row — but it is a deliberate product call, not a side effect.

Verification: `flutter analyze lib test integration_test` clean; 786 tests across `test/blocs/dm/`, `test/screens/inbox/`, `test/config/` and the five label-service consumers; `dm_repository` 529 tests at 93.60% (gate 93); all four design-system ratchets plus the untested-services and test-structure floors pass. Rebased onto fresh `origin/main` — that also cleared an untested-services floor failure caused by branch drift, not by this diff.

realmeylisdev:
@hm21 thanks for testing on a real small phone — items 4 and 5 were the same bug and I'd have shipped it otherwise. All five answered:

**1. Why experimental?** It isn't any more — `3c7247c6e` removed `FF_SUPPORT_DM_ROW` and the row ships on by default. You were right that there was nothing to stage.

**2. Black logo.** Fixed in `be92eb5bb`. The moderation account's kind-0 `picture` is `divine-logo.svg`, which colours its paths through a `<style>` block; `vector_graphics_compiler` has no `<style>` parser, so it discarded the stylesheet and every path fell back to the SVG default fill — black, 1.11:1 against the inbox surface. The row now renders the bundled `assets/icon/logo.svg`, which carries presentational `fill` per path.

**4 + 5. Scrollable area and the search layout.** Fixed in `9f304ec33`. The pane was stacking ~324px of fixed-height chrome above the `ListView` — support row (92), `FollowingBar` (128), search field (56), filter chips (48) — which with the keyboard up left roughly one row visible and overflowed by 8.4px. It's now a single `CustomScrollView` at every size, with only the All/Unread chips pinned; fixed chrome drops from ~324px to 48.

I deliberately didn't use a screen-height threshold. Your two screenshots are the *same device* with and without the keyboard — it's the keyboard that collapses the viewport, not the screen — so a threshold would have left item 5 broken on the exact phone that reported item 4, at the cost of two live layouts. `FollowingBar` collapsing on search focus is your suggestion; I keyed it to focus rather than a non-empty query because `minSearchQueryLength` is 2, so the query is still empty after the first keystroke — precisely when the list has least room.

**3. Too much of a focal point.** Fair, and it partly self-corrected: the row used to be fixed chrome above the list, which is most of what made it read as the screen's headline. It's now the first *tile inside* the list, so it scrolls away like any other chat.

I've kept it pinned to the top rather than moving it behind "Add new chat", because #6283 is explicitly parity with divine-web, which ships it as the first row of the DM list. It also can't be left to the recency sort — a thread nobody has messaged has no timestamp and would sink out of reach. It does respect the user's filters: it drops out of a search it doesn't match and out of Unread once read.

That said, "users shouldn't need to contact us often" is a real argument and I don't think the issue settles it. If you'd still rather it lived behind the compose flow, that's a product call worth making explicitly — happy to open a follow-up rather than relitigate it in this PR.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -4929,6 +4929,86 @@ class DmRepository {
...
+    final knownIds = <String>{
+      currentId,
+      for (final pubkey in legacyPubkeys)
+        if (pubkey.isNotEmpty) computeConversationId([userPubkey, pubkey]),
+    };
+
+    DmConversation? pinned;
+    final remainingInbox = <DmConversation>[];
+    for (final conversation in inbox) {
+      if (!knownIds.contains(conversation.id)) {
```
NotThatKindOfDrLiz: This is the right place to stop for product sign-off: every known moderation conversation ID is removed from the visible lists, but only the current-key conversation is adopted as the pin. That means a legacy-only user loses the in-app entry point to their existing moderation history. The deliverability rationale makes sense technically, but hiding previously visible support history is a product decision, not just a dedupe implementation detail. Please get explicit product acceptance here, and if the answer is yes, link/file the follow-up for the orphaned legacy rows.

NotThatKindOfDrLiz (changes requested):
Requesting changes for the legacy moderation history tradeoff only. The code now deliberately removes pre-rotation moderation conversations from Inbox and Requests without adopting them as the pin, which keeps new replies deliverable to the current key but also makes existing user-visible support history unreachable in-app. Please get explicit product acceptance for that behavior before merge; if accepted, link or file the follow-up for the orphaned legacy rows so this does not live only in the PR body.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: Filed as #6416. Sizing it first, because the affected population is not the one the code comment implies.

**No shipped client ever sent a DM to the retired key.** Mobile's only pre-#6388 send path to moderation is the post-report DM (`a851831e6`, 2026-03-12). That commit and the pubkey correction (`683577093`, #2321) are **both** first contained in tag `1.0.7`, and neither is in `1.0.6` — so the first release that could send a moderation DM already resolved to the current key. In shipped mobile code the retired key was only ever a label-filter constant (`ba51840a2`). Web never used it for DMs either: its support destination was a personal key (`78a5c21b5166dc1474b64ddf7454bf79e6b5d6b4a77148593bf1e866b73c2738`) until divine-web#474 repointed it on 2026-07-09.

**The retired-key threads that do exist are inbound enforcement notices.** `divine-moderation-service` shipped NIP-17 notifications for `PERMANENT_BAN` / `AGE_RESTRICTED` / `QUARANTINE` on 2026-03-08 (#24), gated on and signed with `MODERATOR_NSEC`; #31 removed that secret on 2026-03-15, and divine-iac-coreconfig#280 records the same `121b91…` → `8fd5eb…` rotation from the backend side. So: a ~7-day window, inbound-only (they sit in Message Requests unless the creator replied) — and the part that argues your way, two of those templates invite a reply as the appeal channel. That is a stronger thing to hide than the "dead read-only thread" the PR body claimed, and I've corrected the body.

**Keeping the row is not the safe option either.** `dm-reader` derives its watched inbox from `NOSTR_PRIVATE_KEY`, so nothing has ingested DMs addressed to the retired key since 2026-03-15. Leaving the row preserves the record *and* a live composer that writes appeals nobody reads. The choice is hidden history vs. visible history plus a silent dead end, not history vs. nothing.

Recommending we accept the current behaviour here and build the archived read-only shape in #6416: `DmSendPolicy` is the seam, but it needs a new decision case plus copy for the archived label and the blocked composer, so it wants your call on the copy regardless. Nothing is deleted — the conversations stay in Drift, so #6416 has no recovery step.

If you'd rather not accept it, the alternative is one line: keep legacy ids out of the removal set so only the current key de-duplicates. That restores today's behaviour, dead-end composer included. Say which and I'll push it.

Two things I can't answer from the repos, raised in `support-trust-safety#192`: how many creators got a notice in that window — it's in the moderation service's D1 `dm_log`, and the relay can't tell us because gift wraps hide the sender — and whether anyone still holds the retired nsec. If it's archived, adding it to `dm-reader`'s watched set would make the pre-rotation appeal channel readable again, which turns #6416 from "archive the row" into "the thread still works".

realmeylisdev:
On the legacy-history hold: answered in the thread on `extractPinnedSupport`, and the archived read-only follow-up is filed as #6416 with the evidence for who is actually affected.

Short version — no shipped client ever *sent* to the retired key (mobile's send path and the pubkey correction both first ship in `1.0.7`; web's support row pointed at a personal key until divine-web#474), so the retired-key threads that exist are the inbound enforcement notices this backend signed with `MODERATOR_NSEC` across a ~7-day window in March, two of which invite a reply as the appeal channel. Keeping the row visible preserves that record along with a composer nothing has ingested since 2026-03-15, so the choice is hidden history vs. visible history plus a silent dead end.

@NotThatKindOfDrLiz — recommending we accept it here and build #6416 with your copy call; if you'd rather not, keeping legacy ids out of the removal set is a one-line change and I'll push it. @mbradley — two backend questions that size this (notice count in the D1 `dm_log`, and whether the retired nsec still exists anywhere) are on `support-trust-safety#192`.

rabble (commented):
Review findings:

`mobile/lib/screens/inbox/inbox_view.dart:L703`: 🔴 bug: requests-only empty-state branch removed. On main, `conversations.isEmpty && hasRequests` returned `MessageRequestsBanner + InboxEmptyState`. Now that case falls into the general path: `visibleConversations` is empty, so the user whose only messages are requests gets the banner + "You're all caught up" (`inboxUnreadEmptyTitle`) — the exact unread-filter-copy-below-requests-banner bug the PR body says it restores. No test covers requests-only (existing banner tests all carry a real conversation). Fix: restore the `conversations.isEmpty && hasRequests` branch (banner + support row + `InboxEmptyState`) and add a widget test.

`mobile/packages/dm_repository/lib/src/dm_repository.dart:L4977`: 🟡 risk: retired-key moderation threads are dropped from BOTH inbox and requests and never adopted (`knownIds` contains legacy ids; only `currentId` comes back as `pinned`), so a user's pre-rotation moderation history becomes unreachable in-app. PR body bullet 3 states the opposite ("Pre-rotation moderation threads remain visible in their original inbox or request location … this PR also does not hide the user's only in-app entry point to that history"). Tests codify the drop, so either the behavior is intended and the description is wrong, or history access regresses ahead of #6416. Fix: reconcile — keep retired threads in their original lists (de-dupe only when a current-key thread exists), or correct the PR body.

`mobile/lib/screens/inbox/inbox_view.dart:L841`: 🔵 nit: `_supportRowSurvivesFilter` matches title + `lastMessageContent` but not the subtitle blurb ("Bugs, moderation, account stuff — we're listening.") that a synthetic row renders as its preview. Searching "bugs" hides the row while that exact text is on screen — filter no longer matches what the user sees. Fix: when `lastMessageContent == null` (subtitle would render), also match `inboxSupportRowSubtitle`.

NotThatKindOfDrLiz (approved):
Approved after takeover. The current behavior preserves legacy moderation history in-place while keeping the pinned support row tied to the current moderation account, and CI is green.

NotThatKindOfDrLiz:
Took over the branch and pushed `5a2d257` (`fix(dm): preserve legacy moderation history`).\n\nWhat changed:\n- Current Divine Moderation conversations are still adopted into the pinned support row and removed from the ordinary inbox/request lists.\n- Legacy/pre-rotation moderation conversations now stay visible in their original inbox or requests location, so historical PERMANENT_BAN / AGE_RESTRICTED / QUARANTINE messages remain reachable in-app. The pin still targets the current moderation account for new replies.\n- Fixed the requests-only inbox empty-state copy regression.\n- Fixed the pinned filter chips under large system text scaling and replaced the ineffective guard test with one that verifies text scaling is disabled inside the pinned header.\n- Updated the PR body to document the product behavior and leave #6416 as the archived read-only follow-up, not a blocker for preserving reachability now.\n\nLocal verification passed before push:\n- `flutter test test/screens/inbox/inbox_view_test.dart`\n- `flutter test test/blocs/dm/conversation_list/conversation_list_bloc_test.dart`\n- `flutter test test/blocs/dm/unread_count/dm_unread_count_cubit_test.dart`\n- `flutter test test/src/dm_repository_test.dart` from `mobile/packages/dm_repository`\n- focused `flutter analyze` runs for the touched app and package files\n\nGitHub CI is green and the PR is mergeable. @realmeylisdev please review the takeover commit and merge when you are comfortable with the product shape.

mbradley (commented):
The takeover flipped legacy-key handling from "de-duplicate" to "preserve in place," but the contract around it still describes the old design, so the next reader gets told the opposite of what ships.

`DmRepository.extractPinnedSupport` still requires `legacyPubkeys` and no longer reads it. `final _ = legacyPubkeys;` (`mobile/packages/dm_repository/lib/src/dm_repository.dart:4985`) keeps it looking used, though the method is public and static, so no unused-parameter lint would have fired regardless. Four call sites still compute and thread the set.

Two comments now state the opposite of the behavior:

- `mobile/lib/config/official_accounts.dart:74-77`: "the pinned support row de-duplicates against these to avoid rendering beside an orphaned 'Divine Moderation' conversation." That adjacency is now the intended outcome.
- `mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart:79-82`: "Used for de-duplication only ... without this it would render beside the pinned row as a second 'Divine Moderation' entry."

The comments are a clear fix either way. The parameter is a judgment call: keeping it so #6416 can reuse the API without re-threading is reasonable, but as shipped it is transitional code without the `// TODO(#6416):` link `code_style.md` asks for, so it reads as an oversight rather than a decision. Drop it and re-add in #6416, or keep it with the TODO?

Not a merge blocker.

One coordination note for @realmeylisdev: the first two findings in @rabble's review read as analyzed against the pre-takeover tree. Both are already fixed at `5a2d257`. The requests-only branch is restored at `inbox_view.dart:703` with a test that genuinely has empty `conversations`, and retired-key threads are preserved with the repository and badge tests inverted to match. Only the search-subtitle nit is still live.

Everything else I checked held up: moderation is `minorContactable: true` and the synthetic pin is pushed back through both the blocklist filter and the #176 gate, so protected minors keep their support channel; the badge cubit and the list partition identically; both new l10n keys are mirrored across every locale rather than parked in the untranslated-debt list.

Thanks, as always.

mbradley:
Heads-up: this went unmergeable after the approval. `#6431` landed on main at 09:27 EDT today and conflicts with `conversation_list_bloc.dart` and `inbox_view.dart`.

Worth flagging because the two changes overlap in meaning at one spot, not just in text. `#6431` added to the emitted state:

```dart
requestsWithheld: !recoveryComplete && split.requests.isNotEmpty,
```

`split.requests` there predates the pin. On this branch the moderation thread is lifted out of the request set into the pinned row, and the pin is deliberately extracted above the recovery hold-back so it still renders during the drain. Keeping that line verbatim means a user whose only request is the moderation thread gets the withheld-requests banner (main's `inbox_view.dart:469`, `message_requests_view.dart:114`) while the thing being reported as withheld is on screen as the pin. The expression that survives the merge is `pin.requests.isNotEmpty`, with `conversations: pin.inbox` from this side and `requestsWithheld` kept from main's.

The two `inbox_view.dart` hunks are the same collision from the view side: main's withheld banner against this branch's `CustomScrollView` rewrite.

I'm holding the doc-comment fixes from my earlier review until after the rebase so they don't turn into rebase fodder. Happy to push them as a single commit then, or leave them to you if you'd rather fold them in.

mbradley:
@realmeylisdev I'd normally just take the branch and rebase it, per the takeover policy in `divine-context`, but in this case you own both sides of the conflict. `#6431` is yours and it landed about forty minutes ago, so you may well be mid-rebase already, and a force-push from me would throw that away. I also said above that I was holding for your rebase, and reversing that without asking is the wrong way round.

Keeping it, or want to hand it off? If you hand it off I'll rebase, resolve both files as described above, add the regression test for the withheld banner, run the suites, and hand back for your merge.

realmeylisdev:
Rebased onto `5f1261f945` and pushed. `pin.requests.isNotEmpty` is what survived, exactly as you called it — amended into `fix(dm): never adopt a retired-key moderation thread as the support pin` rather than added on top, since that is the commit that moves the hold-back below the blocklist and #176 filters and so the first one where `pin.requests` is in scope at all. Before it, `split.requests` is still the right read.

The two `inbox_view.dart` hunks: `_RestorePausedBannerGate` is unchanged and now mounts as a `SliverToBoxAdapter` immediately after `_RestoringHistoryIndicator`, outside the `switch (status)`. That is what the Column placement was buying, so it still covers empty / requests-only / filtered / populated without a flag threaded through each branch, and without re-adding the fixed chrome @hm21's review removed. It already self-gates on `loaded`, so the spinner and error branches are untouched. All four of your banner tests pass against the restructured tree.

Rather than trust ten rounds of conflict resolution, I diffed the final tree against `git merge-tree`'s auto-merge output — the whole delta is those three resolutions plus the banner sliver, nothing else moved.

On your review: taking the parameter out rather than annotating it. #6416's shape changes that signature anyway, so keeping the current one buys nothing, and a required parameter no code path reads is dead plumbing whether or not a TODO sits above it. Removed end to end — the repository parameter, both blocs' fields, and the three wiring sites. `kLegacyModerationPubkeys` stays for `isModerationAccount` and the label-service migration.

Comments are covered too, so don't spend a commit on them. Two beyond the ones you listed had also inverted: `_extractPinnedSupport`'s own doc still said it removes "every conversation it stands in for", and the badge cubit's parity note still said a retired-key thread "must STOP being counted (the inbox drops it entirely)". Four total.

Also took @rabble's search-subtitle nit while here — `_supportRowSurvivesFilter` now matches whichever string the row actually renders, verified to fail without the fix.

Local before push: `flutter analyze lib test integration_test`, 704 tests across `test/screens/inbox/` + `test/blocs/dm/` + `official_accounts` + `app_shell_badge_scope`, 551 in `dm_repository`, 663 in `divine_ui`. CI was green on the rebase before the follow-ups landed.

realmeylisdev:
Pushed `e535381` — reworks the filter-chip part of the takeover. Flagging it because it changes @NotThatKindOfDrLiz's commit rather than adding to it.

`MediaQuery.withNoTextScaling` was solving a real clip, but by freezing the control instead of fixing the cause. The chips sit in a pinned `SliverPersistentHeader`, which has to declare its extent before its child is laid out, and that extent was the constant 48. `DivineButton` never clamps — a tiny chip is 6px + a 20px line box + 6px — so at large text the chip wanted more than the header had declared, and a Row overflowing on its *cross* axis clips silently rather than throwing. Freezing the text was the only way to keep the constant true.

So the header now declares a scaling extent, threaded in from the widget that has a `BuildContext` (same shape as the `topInset` pattern in `ui_theming.md`), and the wrapper is gone. `accessibility.md` sanctions `withNoTextScaling` for small fixed badges overlaid on images; all seven other uses in `mobile/lib` are non-interactive, and `message_input_bar.dart:86` carries an explicit "Field honours font scaling (#4620): no withNoTextScaling" for the composer next door. No `divine_ui` change was needed — I expected one and was wrong.

The kept `expect(getSize(chips).height, equals(48))` also had to go: a pinned header lays its child out with a tight height equal to `maxExtent`, so that assertion was measuring the constant against itself. I checked before rewriting — with the wrapper removed and `TextScaler(2)` applied, it still reported 48. The tests now assert the chips' own intrinsic height, which comes from the content and is independent of the constraint, and I verified the growth test reports 48 against a needed 68 when the extent is pinned back to a constant.

One other thing worth surfacing, found while re-verifying the merge. `29ef07ae` had silently reverted the merged `requestsWithheld` source from `pin.requests` back to `split.requests`, and it survived a full green CI run — every pre-existing test for that flag uses ordinary conversations, where the two expressions are equal. Amended out, and `0be8c3b` adds two guards that can tell them apart: a recovery window whose only would-be request is the moderation thread must not raise the banner, and the sibling case with a real held-back stranger must still raise it.

Also in since the rebase: the `legacyPubkeys` parameter removed end to end per your review, all four stale comments corrected, and @rabble's search-subtitle nit fixed.

hm21 (approved):
LGTM

Issue #6391, opened 2026-07-25
feat(lists): notify users when their videos are added to public lists


hm21:
### What would you like?

Send a notification when one of my videos is added to a public list. For example: "SLOPAGANDA just added your Vine to the 'Literature 📚' list."

### How would this be useful for you?

It helps creators discover users who are curating their content, find others with similar interests, and engage with communities in their niche.

### When would you use this?

Whenever another user adds one of my videos to a public list, I'd receive a notification and could explore the list or the curator's profile.

### Anything else?

requested on discord [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1530047669605498980)

PR #6394, opened 2026-07-25
fix(lists): preserve addressable video tags

### LINKED ISSUE #6392: fix(lists): preserve addressable video tags in curated lists
## Problem

Curated lists use NIP-51 kind 30005 events and the app model supports mixed video references: plain event IDs and NIP-71 addressable video coordinates.

The current mobile app has multiple conversion paths for kind 30005 lists. One live path reads only `e` tags, so relay-sourced lists can silently drop valid video `a` references. The write path also emits every stored video reference as an `e` tag, which can republish an addressable coordinate as a malformed event-id reference.

This matters for #6391 because the backend list-add notification work will index exactly this list data.

## Acceptance criteria

- Mobile uses one curated-list converter for kind 30005 read/write behavior.
- Relay-sourced valid NIP-71 video `a` references are preserved in `CuratedList.videoEventIds`.
- Publishing a list emits valid addressable video coordinates as `a` tags, not malformed `e` tags.
- Plain event IDs continue to publish as `e` tags.
- Existing curated-list CRUD behavior is preserved and tested.

## Related

- Prerequisite for #6391
- Backend blocker: divinevideo/divine-funnelcake#705

NotThatKindOfDrLiz:
## Summary

- Route mobile kind 30005 publish/read paths through `CuratedListConverter` as the single source of truth.
- Preserve valid NIP-71 addressable video coordinates as `a` tags when publishing curated lists.
- Delete duplicate app-layer curated-list conversion extensions and update affected tests.

## Verification

- `flutter test test/src/curated_list_converter_test.dart` from `mobile/packages/curated_list_repository`
- `flutter test test/services/curated_list_service_crud_test.dart` from `mobile`
- Pre-push changed-file analysis and tests passed

## Related

Closes #6392
Prerequisite for #6391
Backend blocker: divinevideo/divine-funnelcake#705


hm21 (changes requested):
Thanks for consolidating the publish/sync path and adding regression coverage. I found two blockers before this is ready:

1. **[P1] Route every kind-30005 read through the shared converter** (`mobile/lib/services/curated_list_service.dart:1785`)

   This switches only `_processListEvent`. The full duplicate `_eventToCuratedList` is still used by `streamPublicListsFromRelays`, `fetchPublicList`, `fetchPublicListsContainingVideo`, and `streamPublicListsContainingVideo`. The two implementations already disagree: for example, `34236::missing-pubkey` is ignored by `CuratedListConverter.fromEvent` but accepted as a video by `_eventToCuratedList`, so the same relay event produces different models depending on the entry point. This also leaves the issue's explicit single-converter acceptance criterion unmet. Please route those reads through `CuratedListConverter.fromEvent` (a thin logging wrapper is fine) and remove the manual parser.

2. **[P1] Limit addressable video kinds to the canonical NIP-71 set** (`mobile/packages/curated_list_repository/lib/src/curated_list_converter.dart:164`)

   Because this helper now controls publishing, the existing hard-coded `34237` entry makes a stored `34237:...` value emit as an `a` tag, and the new test locks that behavior in. Kind 34237 is not an addressable NIP-71 video kind; the canonical `models` `NIP71VideoKinds` and `mobile/docs/NOSTR_VIDEO_EVENTS.md` both list only 34235/34236 for addressable videos. Please derive this check from the canonical model constants (or otherwise restrict it to 34235/34236) and remove the 34237 write expectation.

Validation: the converter package tests and the curated-list CRUD/query/fetch-public tests all pass locally; all current GitHub checks are green.

hm21 (approved):
The two review blockers have been addressed: all `kind:30005` read paths now use the shared `CuratedListConverter`, and addressable video references are restricted to the canonical NIP-71 kinds `34235` and `34236`. Regression tests have been added, and all local checks as well as all GitHub checks on head `47cc2279f` are passing.

PR #6395, opened 2026-07-25
fix(news-feed): ignore iOS back swipe for tuning

### LINKED ISSUE #6390: fix(news-feed): prevent back swipe from triggering "More like this"
### Summary

Swiping left to exit a video in the news feed also marks the video as "More like this," causing conflicting behavior.

### Environment

iOS

### Steps to Reproduce

1. Activate the experimental "Feed Tuning Swipes"
2. Open a video from the news feed.
3. Swipe left to exit the video (the usual back gesture).
4. Observe the result after the swipe.

### Actual Result

The app navigates back but also marks the video as "More like this."

### Expected Result

The swipe should perform only one action. If it is used as a back gesture, it should return to the news feed without also triggering the "More like this" action.

### Regression?

No

### Evidence

reported on discord [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1529665923424718848)

NotThatKindOfDrLiz:
## Summary
- ignore feed-tuning drags that start in Flutter's iOS leading-edge back gesture strip on poppable routes
- keep center swipes, root-route tuning, non-iOS pushed route tuning, vertical paging, trackpad pan, pointer scroll, and semantics behavior intact
- add regression coverage for iOS back swipe, RTL leading edge, root route, and non-iOS pushed routes

## Verification
- flutter test test/widgets/feed_tuning/feed_tuning_swipe_overlay_test.dart
- pre-push hook: analyzer + changed widget test suite

Closes #6390

### DIFF
--- a/mobile/lib/widgets/feed_tuning/feed_tuning_swipe_overlay.dart
+++ b/mobile/lib/widgets/feed_tuning/feed_tuning_swipe_overlay.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:math' as math;
 
 import 'package:divine_ui/divine_ui.dart';
 import 'package:feed_tuning_repository/feed_tuning_repository.dart';
@@ -7,6 +8,9 @@ import 'package:flutter/material.dart';
 import 'package:flutter/semantics.dart';
 import 'package:flutter/services.dart';
 import 'package:openvine/l10n/l10n.dart';
+
+// Mirrors Flutter's private _kBackGestureWidth in cupertino/route.dart.
+const double _platformBackGestureWidth = 20;
 
 /// Horizontal-swipe detector over the fullscreen feed that lets the user tune
 /// recommendations: swipe right = "more like this", swipe left = "less like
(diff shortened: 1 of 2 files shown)


hm21 (changes requested):
The leading-edge exclusion itself matches Flutter: 20 px or the safe-area inset, the RTL side, and popGestureEnabled are all correct. The focused test suite also passes locally (19/19).

Changes requested:

1. The new _isRouteTransitionInProgress condition is broader than the stated iOS edge-swipe fix: it drops committed tuning gestures on every platform and route while the primary route animation is not completed. It is not documented, and no test fails when this condition alone is removed. Conversely, stubbing _isPlatformBackGestureStart to false also leaves all 19 tests green; only removing both mechanisms makes the two back-swipe tests fail. Please either remove the transition guard so the regression tests pin the edge exclusion they claim to test, or explicitly document the broader behavior and add a test that fails without it.

2. The new platform parameter on pumpOverlay is unused; all existing call sites omit it, while the platform cases use the two new helpers. Please remove it unless it is used by the revised coverage.

No other blocking findings.

hm21 (approved):
The leading-edge check is a faithful mirror of Flutter's `_CupertinoBackGestureDetector` (width, `max(padding, 20)`, RTL side, `popGestureEnabled`), and the app uses the default `PageTransitionsTheme`, so it fires where the bug is.

Pushed one commit: the end-of-drag `_isRouteTransitionInProgress` check was redundant and untested — removing it kept all 19 cases green, and while it was in place stubbing out `_isPlatformBackGestureStart` also kept them green, so nothing pinned the actual fix. It also silently dropped tunes committed during any route animation. With it gone the back-swipe tests fail when the leading-edge check breaks (verified by ablation). Also dropped the unused `platform` param on `pumpOverlay`.

analyze + suite green locally. Yours to merge.

PR #6399, opened 2026-07-25
feat(review): prompt engaged store-installed creators to rate the app

### LINKED ISSUE #6400: feat(review): prompt engaged store-installed creators to rate the app
Auto-created from PR #6399.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6399
**Branch:** `feat/in-app-review-prompt`

rabble:
Closes #6400

## Summary

Asks only genuinely invested users to rate Divine via the OS-native review card (`SKStoreReviewController` on iOS, Google Play in-app review on Android). No custom dialog or copy; zero design-system surface and zero new arb strings.

## Eligibility

A user is prompted only when every condition holds:

1. Store install: Google Play or the iOS App Store. TestFlight, Zapstore, sideloaded, unknown, and unsupported shells are excluded.
2. Invested creator: strictly more than 10 published videos from `ProfileStats.videoCount`.
3. Sustained use: `>=10` cold starts OR `>=14` days since first launch.
4. Cooldown elapsed: at least ~6 months since the last prompt attempt for this account.

The prompt is evaluated on return-to-foreground, after cheap local gates pass. Profile stats are loaded on demand instead of keeping a root Drift watch alive for the whole app session.

## Install-Source Detection

This also replaces the previous `InstallSource.sideload` stub used by `AppUpdateRepository`:

- `InstallSourceService` resolves Android installer package name and iOS receipt environment through `divine/install_source`.
- Android maps Play Store and Zapstore installers, with unknown/null installers failing closed to sideload.
- iOS maps sandbox receipts to TestFlight and production receipts to App Store; missing/unknown receipt responses fail closed to sideload.
- `DeviceScope.installSource` shares the resolved source across account containers so review gating and update nudges use the same value.
- Platform failures and missing native channel fallbacks are observable via unified logging.

## Prompt Orchestration

- `AppReviewCoordinator` is a zero-size root widget that watches auth/foreground only and dispatches after a false-to-true foreground transition.
- `AppReviewCoordinatorCubit` owns the ordering: local gate, bounded stable-frame wait, `isAvailable()`, cooldown write, native request, analytics.
- Cooldown is written immediately before `requestReview()`, not before availability/mounted/frame checks, so unavailable platforms or unmounted UI cannot silently burn the 180-day cooldown.
- The coordinator is nested under `DatabaseCorruptionGate` so the database recovery screen cannot be interrupted by a native review card.

## Tests

Adds and updates tests for:

- Pure eligibility gate and local preflight.
- Engagement store counters.
- Install-source mapping, null receipt fallback, missing plugin fallback, and `DeviceScope.installSource` override.
- Analytics bucketing for video count, session count, and days since first launch.
- Coordinator Cubit ordering, frame failure, mounted checks, in-flight guard, unavailable platform behavior, request failure tracking, and cooldown write timing.

## Verification

- `flutter analyze`
- `flutter test test/features/app_review/app_review_analytics_test.dart test/features/app_review/app_review_coordinator_cubit_test.dart test/services/app_review_prompt_service_test.dart test/services/app_engagement_store_test.dart test/services/install_source_service_test.dart test/providers/device_scope_test.dart test/providers/swap_account_test.dart test/main_hashtag_deep_link_nav_action_test.dart test/main_notification_tap_routing_test.dart test/main_search_deep_link_nav_action_test.dart`
- `bash scripts/check_raw_logging.sh`
- `bash scripts/check_native_transport_security.sh`

### DIFF
--- a/mobile/android/app/src/main/kotlin/co/openvine/app/MainActivity.kt
+++ b/mobile/android/app/src/main/kotlin/co/openvine/app/MainActivity.kt
@@ -154,10 +154,12 @@ class MainActivity : FlutterActivity() {
         )
     }
 
-    /// Resolves the installer package name (e.g. "com.android.vending" for Play
-    /// Store, "com.zapstore.app" for Zapstore) so the Dart side can classify
-    /// the install source. Returns null for sideloaded APKs or on older Android
-    /// versions where the installer cannot be determined.
+    /**
+     * Resolves the installer package name (e.g. "com.android.vending" for Play
+     * Store, "com.zapstore.app" for Zapstore) so the Dart side can classify
+     * the install source. Returns null for sideloaded APKs or on older Android
+     * versions where the installer cannot be determined.
+     */
     @Suppress("DEPRECATION")
     private fun resolveInstallerPackageName(): String? {
         return try {
--- a/mobile/android/app/src/main/kotlin/co/openvine/app/MainActivity.kt
+++ b/mobile/android/app/src/main/kotlin/co/openvine/app/MainActivity.kt
@@ -46,6 +46,8 @@ class MainActivity : FlutterActivity() {
     companion object {
         private const val NAVIGATION_CHANNEL = "org.openvine/navigation"
         private const val NAV_TAG = "OpenVineNavigation"
+        private const val INSTALL_SOURCE_CHANNEL = "divine/install_source"
+        private const val INSTALL_SOURCE_TAG = "OpenVineInstallSource"
     }
 
     private var navigationChannel: MethodChannel? = null
@@ -76,6 +78,11 @@ class MainActivity : FlutterActivity() {
 
         // Set up navigation channel for back button handling
         setupNavigationChannel(flutterEngine)
+
+        // Set up install-source channel (Play Store / Zapstore / sideload).
+        // Drives the in-app review prompt's store-install gate and the
+        // AppUpdateRepository's download-URL resolution (#6296).
+        setupInstallSourceChannel(flutterEngine)
 
         // Set up NIP-55 Android Signer plugin
         nostrSignerPlugin = NostrSignerPlugin(this, flutterEngine)
@@ -145,6 +152,43 @@ class MainActivity : FlutterActivity() {
             isFinishing = { isFinishing },
             onFallback = { finish() },
         )
+    }
+
+    /// Resolves the installer package name (e.g. "com.android.vending" for Play
+    /// Store, "com.zapstore.app" for Zapstore) so the Dart side can classify
+    /// the install source. Returns null for sideloaded APKs or on older Android
+    /// versions where the installer cannot be determined.
+    @Suppress("DEPRECATION")
+    private fun resolveInstallerPackageName(): String? {
+        return try {
+            // Prefer getInstallSourceInfo (Play's documented API) on Tiramisu+.
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                packageManager
+                    .getInstallSourceInfo(packageName)
+                    .installingPackageName
+            } else {
+                packageManager.getInstallerPackageName(packageName)
+            }
+        } catch (e: PackageManager.NameNotFoundException) {
+            Log.d(INSTALL_SOURCE_TAG, "No installer package name available", e)
+            null
+        } catch (e: IllegalArgumentException) {
+            Log.d(INSTALL_SOURCE_TAG, "Installer lookup failed", e)
+            null
+        }
+    }
+
+    private fun setupInstallSourceChannel(flutterEngine: FlutterEngine) {
+        MethodChannel(
+            flutterEngine.dartExecutor.binaryMessenger,
+            INSTALL_SOURCE_CHANNEL,
+        ).setMethodCallHandler { call, result ->
+            when (call.method) {
+                "getInstallerPackageName" -> result.success(resolveInstallerPackageName())
+                else -> result.notImplemented()
+            }
+        }
+        Log.d(INSTALL_SOURCE_TAG, "Install source channel registered")
     }
 
     override fun onDestroy() {
--- a/mobile/ios/Runner/AppDelegate.swift
+++ b/mobile/ios/Runner/AppDelegate.swift
@@ -213,12 +213,11 @@ extension FlutterError: @retroactive Error {}
       switch call.method {
       case "isSandboxReceipt":
         // The receipt URL is nil in simulator/debug runs without a store
-        // receipt; treat that as "not sandbox" so debug builds don't get
-        // classified as TestFlight. Production App Store builds always have a
-        // receipt URL whose lastPathComponent is "receipt".
+        // receipt. Return nil so Dart fails closed to sideload instead of
+        // pretending an unknown receipt environment is the App Store.
         guard let receiptURL = Bundle.main.appStoreReceiptURL else {
-          NSLog("📦 InstallSource: no receipt URL — defaulting to App Store")
-          result(false)
+          NSLog("📦 InstallSource: no receipt URL — returning unknown")
+          result(nil)
           return
         }
         let isSandbox = receiptURL.lastPathComponent == "sandboxReceipt"
--- a/mobile/ios/Runner/AppDelegate.swift
+++ b/mobile/ios/Runner/AppDelegate.swift
@@ -72,6 +72,11 @@ extension FlutterError: @retroactive Error {}
     // Set up Zendesk platform channel
     setupZendeskChannel(with: engineBridge)
 
+    // Set up install-source channel (App Store / TestFlight / sideload).
+    // Drives the in-app review prompt's store-install gate and the
+    // AppUpdateRepository's download-URL resolution (#6296).
+    setupInstallSourceChannel(with: engineBridge)
+
     NSLog("✅ AppDelegate: Implicit Flutter engine initialized with UIScene lifecycle")
   }
 
@@ -190,6 +195,42 @@ extension FlutterError: @retroactive Error {}
     }
 
     NSLog("✅ ProofMode: Platform channel registered with LibProofMode")
+  }
+
+  /// Install-source platform channel.
+  ///
+  /// iOS distinguishes install source by the app store receipt URL: TestFlight
+  /// builds ship a `sandboxReceipt`, App Store builds ship a `receipt`. The
+  /// Dart side maps `isSandbox` → testFlight/appStore via the shared
+  /// `resolveIosInstallSource` resolver in app_update_repository.
+  private func setupInstallSourceChannel(with engineBridge: FlutterImplicitEngineBridge) {
+    let channel = FlutterMethodChannel(
+      name: "divine/install_source",
+      binaryMessenger: engineBridge.applicationRegistrar.messenger()
+    )
+
+    channel.setMethodCallHandler { (call, result) in
+      switch call.method {
+      case "isSandboxReceipt":
+        // The receipt URL is nil in simulator/debug runs without a store
+        // receipt; treat that as "not sandbox" so debug builds don't get
+        // classified as TestFlight. Production App Store builds always have a
+        // receipt URL whose lastPathComponent is "receipt".
+        guard let receiptURL = Bundle.main.appStoreReceiptURL else {
+          NSLog("📦 InstallSource: no receipt URL — defaulting to App Store")
+          result(false)
+          return
+        }
+        let isSandbox = receiptURL.lastPathComponent == "sandboxReceipt"
+        NSLog("📦 InstallSource: isSandbox=\(isSandbox) (\(receiptURL.lastPathComponent))")
+        result(isSandbox)
+
+      default:
+        result(FlutterMethodNotImplemented)
+      }
+    }
+
+    NSLog("✅ InstallSource: Platform channel registered")
   }
 
   private func setupZendeskChannel(with engineBridge: FlutterImplicitEngineBridge) {
--- a/mobile/lib/features/app_review/app_review_coordinator.dart
+++ b/mobile/lib/features/app_review/app_review_coordinator.dart
@@ -5,23 +5,26 @@ import 'dart:async';
 
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:in_app_review/in_app_review.dart';
 import 'package:models/models.dart';
-import 'package:openvine/features/app_review/app_review_analytics.dart';
+import 'package:openvine/features/app_review/app_review_coordinator_cubit.dart';
 import 'package:openvine/providers/analytics_providers.dart';
 import 'package:openvine/providers/app_foreground_provider.dart';
 import 'package:openvine/providers/auth_providers.dart';
 import 'package:openvine/providers/install_source_provider.dart';
+import 'package:openvine/providers/repository_providers.dart';
 import 'package:openvine/providers/shared_preferences_provider.dart';
-import 'package:openvine/providers/user_profile_providers.dart';
 import 'package:openvine/services/app_engagement_store.dart';
 import 'package:openvine/services/app_review_prompt_service.dart';
 import 'package:openvine/services/auth_service.dart' show AuthState;
-import 'package:openvine/services/install_source_service.dart';
 
-/// Provides the [InstallSourceResolver]. Override in tests.
-final installSourceResolverProvider = Provider<InstallSourceResolver>(
-  (ref) => const InstallSourceService(),
+/// Provides the native in-app review platform facade.
+final appReviewPlatformProvider = Provider<AppReviewPlatform>(
+  (ref) => InAppReviewPlatform(),
+);
+
+/// Provides the scheduler used to wait for a stable frame before prompting.
+final appReviewFrameSchedulerProvider = Provider<AppReviewFrameScheduler>(
+  (ref) => const WidgetsAppReviewFrameScheduler(),
 );
 
 /// Provides the [AppReviewPromptService].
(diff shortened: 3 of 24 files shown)


NotThatKindOfDrLiz:
@rabble I took over the review fixes on this PR and pushed them to the PR branch.

What changed:
- Moved prompt orchestration into a tested AppReviewCoordinatorCubit.
- Prevented cooldown writes before mounted/frame/isAvailable checks pass.
- Made the frame wait bounded and scheduled via ensureVisualUpdate().
- Removed the root profile-stats subscription; stats now load on demand after cheap local gates pass.
- Failed closed for unknown/missing iOS receipt and missing install-source channel, with unified logging.
- Nested the coordinator under DatabaseCorruptionGate.
- Removed speculative completed/reset review-state paths and added regression coverage.
- Bucketed session/day analytics and added DeviceScope install-source coverage.

Verification:
- Local flutter analyze passed.
- Local targeted review/install-source/main-provider tests passed.
- check_raw_logging, check_process_global_mutations, and check_native_transport_security passed locally.
- GitHub Mobile CI, preview build, generated-files, analyze, format, and all test shards are green.

NotThatKindOfDrLiz (approved):
Approved after takeover fixes. The prior blockers are addressed and CI is green on head 97d4534c390dc07987aa587cc1c6ed92a5ded87c.

mobile/lib/features/app_review/app_review_coordinator.dart
```diff
@@ -0,0 +1,143 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    // Auth is watched so account swaps rebuild before the next foreground
+    // trigger. Foreground transitions are the only prompt moment.
+    ref.watch(currentAuthStateProvider);
+    final isForeground = ref.watch(appForegroundProvider);
+    final wasForeground = _wasForeground;
+    _wasForeground = isForeground;
+    if (wasForeground == false && isForeground) {
```
realmeylisdev, line 86: **The false→true edge may never be observed, so the prompt may never fire.**

This detects the transition by sampling `appForegroundProvider` during `build()`. But the `false` is written while Flutter has frames turned off, so the build that would record `_wasForeground = false` never runs:

- `WidgetsBinding.handleAppLifecycleStateChanged` calls `super.…` **first** (`widgets/binding.dart:1331`), which is `SchedulerBinding.handleAppLifecycleStateChanged` → `_setFramesEnabledState(false)` for `hidden`/`paused` (`scheduler/binding.dart:423-426`), and `scheduleFrame()` early-returns when `!framesEnabled` (`scheduler/binding.dart:947`).
- Only *then* are observers notified (`widgets/binding.dart:1332`), so `AppLifecycleHandler`'s `setForeground(false)` (`app_lifecycle_handler.dart:171`) marks this element dirty with no frame able to run.

On resume, `inactive` re-enables frames and `resumed` sets foreground back to `true`. Whether this widget ever builds with `isForeground == false` therefore depends on a frame landing in the window *between* `inactive` and `resumed`. If `resumed` arrives first — the common case, since `applicationWillEnterForeground:` and `applicationDidBecomeActive:` usually land in quick succession — the single pending build sees `true`, `_wasForeground` is still `true` from before backgrounding, and no edge is detected. Ever.

The widget test doesn't cover this: it drives `setForeground(false)` + `pump()` explicitly, which forces the intermediate frame that the real lifecycle does not guarantee.

Suggest deriving the transition from the lifecycle callback itself (or a `ProviderContainer.listen` outside the build phase) rather than from a build-time sample, so it doesn't depend on frame scheduling — plus a test that flips the provider twice without an intervening `pump()`.

mobile/lib/features/app_review/app_review_coordinator_cubit.dart
```diff
@@ -0,0 +1,131 @@
...
+        videoCount: inputs.videoCount,
+        sessionCount: inputs.sessionCount,
+        daysSinceFirstLaunch: inputs.daysSinceFirstLaunch,
+      );
+
+      // Start cooldown immediately before the native request. That avoids a
+      // tight re-prompt loop if the OS call crashes while not burning cooldown
+      // for unavailable platforms or an unmounted app.
+      await _service.recordShown(inputs.pubkey);
+      if (!canContinue()) return;
```
realmeylisdev, line 124: **This can burn the 180-day cooldown without ever showing a card**, which contradicts the design doc shipped in this PR.

`docs/superpowers/specs/2026-07-26-app-review-lifecycle-hardening-design.md` states: *"A changed account, sign-out, background transition, or closed Cubit cancels the evaluation **without writing cooldown state** or opening the native prompt."*

But `recordShown` (line 123) is an awaited SharedPreferences write, and this `canContinue()` sits between it and `requestReview()`. Background the app or swap accounts during that write and the cooldown is persisted with no prompt — the user is out for six months having seen nothing.

The PR body makes the same claim ("unavailable platforms or unmounted UI cannot silently burn the 180-day cooldown"), and it holds for every *other* cancellation point — the cubit tests pin availability, frame-wait, unmount, and account-change-during-availability. This one window is the gap, and it's the only one untested.

Simplest fix is to drop this line: line 110 already checks `canContinue()` after `isAvailable()`, so removing it leaves the write and the native call adjacent with no added risk.

mobile/android/app/src/main/kotlin/co/openvine/app/MainActivity.kt
```diff
@@ -147,6 +154,45 @@ class MainActivity : FlutterActivity() {
...
+        }
+    }
+
+    private fun setupInstallSourceChannel(flutterEngine: FlutterEngine) {
+        MethodChannel(
+            flutterEngine.dartExecutor.binaryMessenger,
+            INSTALL_SOURCE_CHANNEL,
+        ).setMethodCallHandler { call, result ->
+            when (call.method) {
+                "getInstallerPackageName" -> result.success(resolveInstallerPackageName())
```
realmeylisdev, line 189: **The Zapstore branch of install-source detection never matches.**

This returns the real installer package, but the Dart side compares against `InstallSource.zapstoreInstaller = 'com.zapstore.app'` (`packages/app_update_repository/lib/src/models/install_source.dart:22`). Zapstore's actual Android `applicationId` is `dev.zapstore.app` — from its own build file:

```
# github.com/zapstore/zapstore → android/app/build.gradle.kts
namespace = "dev.zapstore.app"
applicationId = "dev.zapstore.app"
```

So every Zapstore install falls through `resolveAndroidInstallSource`'s `_ =>` to `sideload`. The PR body says "Android maps Play Store and Zapstore installers" — the Play half works, the Zapstore half doesn't.

No regression for the review gate (both are excluded), and no regression for update nudges either (they were hardcoded `sideload` before). But it does mean the stated fix doesn't land for Zapstore users, who get the GitHub releases URL instead of the Zapstore deep link — and we ship there deliberately (AGENTS.md → *Zapstore Publishing Notes*).

The constant is outside this diff, but this PR is what makes it load-bearing, so it seems in scope. `packages/app_update_repository/test/src/install_source_detector_test.dart` asserts the wrong string too, so the test won't catch the fix.

mobile/ios/Runner/AppDelegate.swift
```diff
@@ -192,6 +197,41 @@ extension FlutterError: @retroactive Error {}
...
+      case "isSandboxReceipt":
+        // The receipt URL is nil in simulator/debug runs without a store
+        // receipt. Return nil so Dart fails closed to sideload instead of
+        // pretending an unknown receipt environment is the App Store.
+        guard let receiptURL = Bundle.main.appStoreReceiptURL else {
+          NSLog("📦 InstallSource: no receipt URL — returning unknown")
+          result(nil)
+          return
+        }
+        let isSandbox = receiptURL.lastPathComponent == "sandboxReceipt"
```
realmeylisdev, line 223: **iOS fails open to `appStore`, not closed to `sideload`.**

The comment above (215-217) says an unknown receipt environment falls back to sideload, but the only fail-closed path is `appStoreReceiptURL == nil`. Any receipt name that isn't exactly `sandboxReceipt` yields `false` → `resolveIosInstallSource(isSandbox: false)` → `InstallSource.appStore`.

`appStoreReceiptURL` returns a *path*, not a guarantee the file exists, so ad-hoc, enterprise, and sideloaded (AltStore-style) builds — which have no receipt at all — land on the `appStore` branch rather than the nil branch. The PR body's "missing/unknown receipt responses fail closed to sideload" only holds for the nil case.

Consequence is mild for the review card (`requestReview()` no-ops off-store) but wrong for the update nudge: an off-store iOS install gets pointed at the App Store instead of GitHub.

Adding a `FileManager.default.fileExists(atPath: receiptURL.path)` check before line 223, returning nil when absent, would make the code match the comment.

mobile/lib/features/app_review/app_review_coordinator.dart
```diff
@@ -0,0 +1,143 @@
...
+            videoCount: stats?.videoCount ?? 0,
+            installSource: installSource,
+            sessionCount: localInputs.sessionCount,
+            daysSinceFirstLaunch: localInputs.daysSinceFirstLaunch,
+          ),
+          isActive: () => _isActiveFor(pubkey),
+        );
+  }
+
+  bool _isActiveFor(String pubkey) {
```
realmeylisdev, line 126: **No screen or task gate — the card can land on top of anything.**

`_isActiveFor` checks mounted / foreground / auth / pubkey, but nothing about what the user is doing. Since the trigger is "returned to the app", the card can appear over the recorder, the video editor, an in-flight upload, or a bug-report form — any flow where the user left the app briefly and came back.

The Amber/NIP-46 signing round trip is the sharpest case: that's an external app switch, so it produces exactly the background→foreground transition this widget waits for, and the user comes back mid-publish.

Both stores' guidance is about not interrupting a task in progress, and a rating card over a half-recorded video is the kind of interruption that earns a one-star. Worth gating on a small allowlist of calm routes (feed, profile) — or triggering after a successful publish instead of on resume. If the design doc considered and rejected this, a line saying so would settle it.

mobile/lib/services/app_review_prompt_service.dart
```diff
@@ -0,0 +1,193 @@
...
+  /// 4. No prior "completed" flag for this pubkey (safety hatch for the case
+  ///    where a future, more-transparent review API lets us detect a rating).
+  /// 5. The per-user cooldown has elapsed since the last prompt attempt.
+  bool shouldShow(ReviewEligibilityInputs inputs, {DateTime? now}) {
+    final referenceNow = now ?? _now();
+
+    if (!_isStoreInstall(inputs.installSource)) return false;
+
+    // (2) Genuinely invested creator.
+    if (inputs.videoCount <= minimumVideoCount) return false;
```
realmeylisdev, line 150: **`videoCount` counts deleted videos**, so "genuinely invested creator" is looser than it reads.

Traced the field end to end: `ProfileStats.videoCount` ← `ProfileStatsData.videoCount` ← `stats.video_count` from funnelcake, served by `get_user_stats` (`divine-funnelcake` → `crates/clickhouse/src/client.rs:8512-8539`):

```sql
SELECT pubkey, countIf(kind IN (34235, 34236)) AS video_count, …
FROM events_deduped FINAL
WHERE pubkey = ?
GROUP BY pubkey
```

No `deleted_events_set` / `addressable_deleted_events_set` exclusion — the same file applies those filters in 18 other read paths, so this query is the outlier. A creator who published 15 and deleted 12 still reads as 15 and clears the `> 10` gate, even though their profile shows 3.

Good news on the adjacent worry: edits do **not** double-count. `events_deduped` sets `dedup_key = ''` for kinds 30000-39999 and orders by `(pubkey, kind, d_tag, dedup_key)` (migration `000044_events_deduped_table.up.sql:30-63`), so addressable re-publishes collapse per NIP-01. And `ProfileStatsData` reads `video_count` only, so the `vine_loops` fallback in `UserProfile.videoCount` never reaches this gate.

Backend-side and pre-existing, so not a blocker — but the gate's one server input doesn't mean quite what the doc comment on line 23 says, and "deleted most of their content" is arguably the population least likely to leave five stars.

mobile/lib/features/app_review/app_review_profile_stats_loader.dart
```diff
@@ -0,0 +1,30 @@
...
+
+  Future<ProfileStats?> load({
+    required Future<void> Function() refresh,
+    required Stream<ProfileStats?> Function() watch,
+  }) async {
+    try {
+      return await (() async {
+        await refresh();
+        return watch()
+            .where((stats) => stats != null)
```
realmeylisdev, line 22: The watch subscription leaks when the timeout fires. `Future.timeout` completes the *outer* future but doesn't cancel work upstream, so the `Stream.first` subscription on `watchProfileStats` stays attached to the Drift query stream for the process lifetime. One per ineligible-but-past-local-gate foreground resume, and those repeat indefinitely (see below).

A manual `StreamSubscription` + `Completer` you can cancel in a `finally` would bound it.

Related: the 3s budget spans `refresh()`, whose own client timeout is much longer, so a slow network discards a perfectly good cached row that `watchProfileStats` would have returned immediately. Reading the cached row first and only refreshing when it's stale would fail soft instead of closed.

mobile/lib/main.dart
```diff
@@ -1428,12 +1432,24 @@ Future<void> _startOpenVineApp() async {
     corruptionService: databaseCorruptionService,
   );
   final accountSwitchController = AccountSwitchController();
+
+  // Resolve how this install was distributed (Play Store / App Store /
+  // TestFlight / Zapstore / sideload) via the native method channel, and
+  // record this cold start for the install-scoped engagement counter that the
+  // in-app review gate reads. Both are best-effort and fall back to safe
+  // defaults on web/desktop/unsupported native shells (#6296).
```
realmeylisdev, line 1440: `#6296` is the wrong issue — it's a Zendesk ticket, *"fix(support): npub1ustv37… - videos have disappeared"*. The install-source / update-nudge work traces to **#2810** (shipped in #2809, which is where the `InstallSource.sideload` placeholder this PR replaces came from).

Same reference appears in `MainActivity.kt:84` and `AppDelegate.swift:77`.

mobile/lib/services/app_review_prompt_service.dart
```diff
@@ -0,0 +1,193 @@
...
+
+  /// Returns `true` only when every eligibility condition is met and the
+  /// per-user cooldown has elapsed.
+  ///
+  /// Conditions (all required):
+  /// 1. Install source is Play Store or App Store.
+  /// 2. `videoCount` strictly exceeds [minimumVideoCount].
+  /// 3. Sustained use: `sessionCount >= minimumSessionCount`
+  ///    OR `daysSinceFirstLaunch >= minimumDaysSinceFirstLaunch`.
+  /// 4. No prior "completed" flag for this pubkey (safety hatch for the case
```
realmeylisdev, line 141: Stale doc: condition **(4) "No prior 'completed' flag for this pubkey"** isn't implemented — there's no `completed` key anywhere in the service. The inline comments below then renumber cooldown as (4), so the doc lists five conditions for four checks.

Dropping (4) and renumbering (5)→(4) matches the code.

mobile/lib/features/app_review/app_review_coordinator_cubit.dart
```diff
@@ -0,0 +1,131 @@
...
+      await _reviewPlatform.requestReview();
+      trackInAppReviewPrompted(analytics: _analytics);
+    } catch (error) {
```
realmeylisdev, line 127: This `catch` spans lines 109-126, so anything `trackInAppReviewEligible` or `recordShown` throws is reported as `in_app_review_request_failed` — a prefs write failure looks like a platform-API failure in the dashboard.

Nothing calls `addError`, so none of it reaches `DivineBlocObserver`/Crashlytics either. Per `.claude/rules/error_handling.md`, a generic `catch` implicitly classifies everything as non-reportable, and `StateError`/`TypeError` are supposed to be YES. Narrowing the `try` to just `isAvailable()`/`requestReview()`, and `addError(Reportable(e, …))` for the programming-error cases, would keep the analytics event honest and still keep expected platform failures out of Crashlytics.

mobile/test/features/app_review/app_review_coordinator_cubit_test.dart
```diff
@@ -0,0 +1,265 @@
...
+    completer.complete(value ?? result);
+  }
+}
+
+typedef VoidCallback = void Function();
+
+void main() {
+  const pubkey = 'pubkey-a';
+
+  ReviewEligibilityInputs inputs({
```
realmeylisdev, line 81: Every test in this file uses always-eligible inputs (`videoCount: 50, playStore, sessionCount: 20, days: 30`), and no override makes them ineligible. That means `shouldShow` at `app_review_coordinator_cubit.dart:103` is a no-op in all eight tests — delete that line and the suite stays green.

The widget test doesn't cover it either: its assertions are all absence-based (`requests == 0`, no cooldown key), so a coordinator that never prompts at all also passes. There's currently no test anywhere proving an eligible user *does* get prompted through the widget path — which is also why the trigger issue on `app_review_coordinator.dart:86` isn't caught.

One cubit test with `inputs(videoCount: 5)` asserting `frameScheduler.waits == 0`, plus one positive-path widget test, would close both.

mobile/lib/services/app_engagement_store.dart
```diff
@@ -0,0 +1,60 @@
...
+    final first = firstLaunchAt;
+    if (first == null) return 0;
+    return (now ?? _now()).difference(first).inDays;
+  }
+
+  /// Records a single cold start: bumps the session counter and, if this is
+  /// the first launch, stamps the install's first-launch timestamp.
+  ///
+  /// Idempotent per cold start — call exactly once during startup, before
+  /// `runApp`. Safe to call again; it will over-count.
```
realmeylisdev, line 49: The doc contradicts itself: *"Idempotent per cold start"* then *"Safe to call again; it will over-count."* Over-counting is the opposite of idempotent, and `app_engagement_store_test.dart:29` pins the incrementing behaviour. "Call exactly once per cold start; repeat calls increment again" says what it does.

realmeylisdev (commented):
Reviewed the full diff plus the native shells, the funnelcake `video_count` contract, and the two committed design docs. The gate logic and its unit tests are solid; my concerns are in the trigger and in install-source resolution.

Two I'd want resolved before merge: the foreground edge is sampled in `build()` at a moment when Flutter has frames disabled, and the Zapstore installer id never matches. Details inline.

Issue #6357, opened 2026-07-24
fix(bluesky): username setup is requested twice


hm21:
### Summary

The experimental Bluesky feature asks users to set up a username again even when one is already configured.

### Environment

any

### Steps to Reproduce

1. Ensure experimental feature "Bluesky" is activated
2. Open General settings.
3. Open Bluesky Publishing
4. Try to activate "Publish videos to Bluesky"

### Actual Result

The app prompts the user to set up a username again.

### Expected Result

The app should detect the existing username and continue without showing the setup prompt.

### Regression?

Unknown

### Evidence

reported on discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1529901056912396441)

NotThatKindOfDrLiz:
Update after keycast#322: the backend username reconciliation blocker called out in #6095 is now shipped, and mobile PR #6705 keeps the name-server fallback while finishing the account provisioning lifecycle UX.

I am not closing this yet because I could not perform a real staging/device verification from this local run. The remaining close-out for this issue is to verify against staging with an app-claimed Divine username:

1. Use an account whose Divine username was claimed in-app.
2. Open Settings -> Bluesky Publishing with `FF_BLUESKY_PUBLISHING` enabled.
3. Enable crossposting.
4. Confirm the app no longer routes back to username setup and instead reaches pending/ready/failed provisioning state.

If that passes with keycast#322 deployed, this issue can close as fixed by the backend reconciliation plus #6705.

Issue #6095, opened 2026-07-14
feat(bluesky): finish and ship ATProto crossposting — backend is live, mobile is the last blocker


rabble:
## Summary

The ATProto/Bluesky backend is **live in production**. Crossposting works end to end today: a Divine video posted to Nostr lands as a playable Bluesky video post, NIP-71 transcripts map to Bluesky closed captions, and per-account PDS publish auth is deployed. The backend is no longer the blocker.

**divine-mobile is now the only user-facing blocker.** The client half is ~70% built and merged (PR #2718, PR #4922) but sits behind a feature flag that defaults to off, and the opt-in flow has a known bug (#3275) that makes the toggle silently fail for most users.

This issue collects everything the mobile side still needs, and includes a ready-to-paste agent prompt at the bottom.

---

## What already exists in divine-mobile (do not rebuild)

| Piece | Path |
|---|---|
| Settings screen | `mobile/lib/screens/settings/bluesky_settings_screen.dart` |
| State | `mobile/lib/blocs/crosspost_settings/crosspost_settings_cubit.dart` + `_state.dart` |
| API client | `mobile/lib/services/crosspost_api_client.dart` |
| Route | `mobile/lib/router/routes/settings_routes.dart` (`/bluesky-settings`) |
| Entry point | `mobile/lib/screens/settings/general_settings_screen.dart` ("Integrations") |
| Feature flag | `FeatureFlag.blueskyPublishing` / env `FF_BLUESKY_PUBLISHING` — **defaults to `false`** |
| Tests | `mobile/test/blocs/crosspost_settings/`, `mobile/test/screens/settings/bluesky_settings_screen_test.dart`, `mobile/test/services/crosspost_api_client_test.dart` |

Separately, NIP-39 Bluesky identity chips (`verified_account_chip.dart`) exist — that is account *verification*, unrelated to crossposting. Don't conflate them.

---

## Backend contract (keycast — `login.divine.video`)

Auth is the existing `KeycastOAuth` bearer session. The mobile app never talks to divine-sky directly; divine-sky's endpoints are all shared-secret service-to-service.

**`GET /api/user/atproto/status`** → `{enabled, state, did, username, error}`
- `username` is the bare local part. Display handle is `<username>.divine.video`.
- `state` ∈ `null` | `pending` | `ready` | `failed` | `disabled`.

**`PUT /api/account/{pubkey}/crosspost`** body `{enabled: bool}`
[...]

PR #6402, opened 2026-07-26
fix(bluesky): stop repeated username setup prompt

NotThatKindOfDrLiz:
## Summary

- checks the active Divine username through divine-name-server instead of treating keycast's null username as authoritative
- changes Bluesky crosspost state from a boolean username gate to claimed/not claimed/unknown so transient lookup failures do not send users into setup
- shows a sync-pending message when divine-name-server confirms a handle but keycast still rejects enablement, and keeps 404 keycast user-row failures out of the username-not-claimed path

## Backend dependency

Refs #6357. This PR stops the repeated setup prompt and false claim notice in mobile, but it intentionally does not close the underlying enablement bug by itself. App-claimed handles still need keycast to reconcile divine-name-server usernames before crosspost enable can succeed: divinevideo/keycast#321.

Also noted the blocker on the broader rollout tracker: #6095.

## Verification

- flutter analyze
- flutter test test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart test/screens/settings/bluesky_settings_screen_test.dart test/services/crosspost_api_client_test.dart test/l10n/arb_consistency_test.dart
- flutter test test/src/profile_repository_test.dart from mobile/packages/profile_repository


rabble (approved):
Reviewed the current rebased head after the design-system correction. The username lookup now distinguishes confirmed missing from temporarily unknown, avoids routing users into a repeated setup loop during lookup failures, preserves enabling behavior when claim state is unknown, and exposes a retry path. The retry notice now uses VineTheme typography and DivineButton, so the raw-style ceiling is clean. Current-head Generated Files, format, analyze, and all four test shards are green; I also ran the affected local gate (analyzer plus 262 tests) while reviewing the equivalent change. No blocking findings.

PR #6403, opened 2026-07-26
fix(l10n): use root context for publish snackbar

### LINKED ISSUE #6114: fix(l10n): AppLocalizations.of null-check crash from a context without a Localizations ancestor (Crashlytics 677b28f3)
## Problem

Background publish completion can call AppLocalizations.of from the BuildContext owned by UploadFailureListener. In production that listener wraps the app/MaterialApp, so its context is above the Localizations tree. AppLocalizations.of(context) returns null and the generated null assertion throws.

This is now root-caused from a symbolicated 1.0.17 event and the current widget topology on origin/main.

## Impact and priority

**Priority: P1 / high**

Crashlytics issue: 677b28f32934ecfda9037cfcbf3ea33d  
[Open in Firebase Crashlytics](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/677b28f32934ecfda9037cfcbf3ea33d)

Current snapshot from 2026-07-22:

- Last 7 days: **880 events / 192 impacted users**
- 2026-07-21: **57 events / 33 users**
- Version 1.0.17 build 805: **1 observed event**
- No build 808 recurrence was visible in the small telemetry sample yet; that is not evidence of resolution
- Earlier 30-day triage recorded **3,022 events / 524 impacted users**

This is a high-volume unhandled error in the shipping background-publish experience.

## Exact production stack

Sample event: 1508f01dad3148ef9587cac82c36d27e_2242764442452187200  
Event time: 2026-07-19T19:11:28Z  
Device: iPhone 16 Pro, iOS 26.5.2  
Process state: background

    AppLocalizations.of
      package:openvine/l10n/generated/app_localizations.dart:87
    AppLocalizationsX.l10n
      package:openvine/l10n/l10n.dart:7
    _showPublishSuccessSnackbar
      package:openvine/main.dart:2840 (release symbol)
    _UploadFailureListenerState.build.<callback>
      package:openvine/main.dart:2785 (release symbol)
    BlocListener subscription callback

Current origin/main line numbers have moved, but the same call path remains in mobile/lib/main.dart.

## Confirmed root cause

UploadFailureListener is constructed outside the app widget returned by buildApp, so the listener's own context does not inherit MaterialApp's Localizations.

The success listener passes that outer context here:

    _showPublishSuccessSnackbar(context, succeededCount);

The helper correctly finds a mounted context inside the root navigator for the ScaffoldMessenger, but resolves localization from the wrong context:

    void _showPublishSuccessSnackbar(BuildContext context, int count) {
      final navContext = NavigatorKeys.root.currentContext;
      if (navContext == null || !navContext.mounted) return;

      final l10n = context.l10n;
      ScaffoldMessenger.of(navContext).showSnackBar(...);
    }

The checked navContext is below MaterialApp and has the required localization ancestor. The passed listener context is above it.

The existing widget test does not reproduce production structure: its harness places UploadFailureListener inside MaterialApp. That makes context.l10n valid and masks this bug.

## Expected behavior

- A completed background publish displays localized success feedback when a mounted root navigator context is available.
- The listener never asks its outer, pre-MaterialApp context for localization.
- If there is no mounted navigator/localization context, completion is handled without an uncaught exception.
- Buffered success feedback still appears after re-authentication according to the existing state machine.
- Singular/plural success copy remains correct.

## Implementation direction

Resolve every context-dependent dependency from the already-validated navContext:

    final navContext = NavigatorKeys.root.currentContext;
    if (navContext == null || !navContext.mounted) return;

    final l10n = navContext.l10n;
    ScaffoldMessenger.of(navContext).showSnackBar(...);

Remove the BuildContext parameter from _showPublishSuccessSnackbar if it is no longer needed, which prevents future callers from supplying an invalid context.

Do not replace this with hard-coded fallback copy. A missing navigator should remain a safe no-op or follow the existing buffered-delivery contract; localization must come from a context that actually owns it.

## TDD plan

Update the widget test so it mirrors production topology before changing the helper.

1. Put UploadFailureListener outside MaterialApp/buildApp.
2. Configure NavigatorKeys.root as the app navigator key.
3. Place a Scaffold inside MaterialApp so the root navigator context has Localizations and a ScaffoldMessenger.
4. Simulate an authenticated background publish transition with recentlySucceededIds.
5. Demonstrate the current code throws from context.l10n.
6. Implement the fix.
7. Assert:
   - no Flutter or zone exception;
   - the localized singular success snackbar appears for one publish;
   - the localized plural success snackbar appears for multiple publishes;
   - no snackbar/context access occurs when root navigator context is absent or unmounted;
   - buffered successes still flush after re-authentication;
   - duplicate/replayed state does not show duplicate success feedback.

The regression must preserve the real provider/BLoC wiring closely enough to catch an outer-listener context mistake.

## Likely files

- mobile/lib/main.dart
- mobile/test/widgets/upload_failure_listener_test.dart
- Existing background publish BLoC tests only if the state contract needs clarification

No localization key change should be needed.

## Acceptance criteria

- [ ] The test harness places UploadFailureListener outside MaterialApp, matching production.
- [ ] The current failure is reproduced by a red test before the fix.
- [ ] Localization and ScaffoldMessenger lookup use a mounted context below MaterialApp.
- [ ] The outer listener context is no longer used for AppLocalizations.
- [ ] Singular, plural, buffered re-auth, and duplicate-suppression behavior remain covered.
- [ ] Missing or unmounted root navigator context is handled without an exception.
- [ ] Focused widget/BLoC tests pass.
- [ ] flutter analyze passes for the affected source and tests.
- [ ] The PR links this issue and records the exact production topology tested.

## Agent workflow

Start from a fresh worktree based on origin/main. Reproduce the production widget topology in a failing test, implement the narrow context-source fix, run the focused tests and targeted analysis, then commit and open a Conventional Commit-titled PR against main.

NotThatKindOfDrLiz:
## Summary
- resolve background publish success snackbar localization from the mounted root navigator context instead of the pre-MaterialApp listener context
- keep buffered publish success counts until the snackbar is actually shown
- update UploadFailureListener widget tests to mirror production topology and cover singular, plural, missing navigator, buffered flush, replay suppression, and vanished-upload suppression
- fix the CI timing-budget test so its committed-budget coverage check honors the guard's existing matrix-job prefix matching

Closes #6114

## Verification
- flutter test test/widgets/upload_failure_listener_test.dart
- flutter analyze lib/main.dart test/widgets/upload_failure_listener_test.dart test/tools/check_ci_timing_budget_test.dart
- pre-push checks: analyzer plus changed-file tests

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -2936,6 +2936,7 @@ class UploadFailureListener extends StatefulWidget {
 
 class _UploadFailureListenerState extends State<UploadFailureListener> {
   var _lastKnownFailedIds = <String>{};
+  var _lastKnownSucceededIds = <String>{};
   var _pendingSuccessCount = 0;
 
   @override
@@ -2948,12 +2949,16 @@ class _UploadFailureListenerState extends State<UploadFailureListener> {
 
         // Use the bloc's own recentlySucceededIds so we never confuse a
         // BackgroundPublishVanished removal with a true publish success.
-        final succeededCount = state.recentlySucceededIds.length;
+        final succeededIds = state.recentlySucceededIds.difference(
+          _lastKnownSucceededIds,
+        );
+        _lastKnownSucceededIds = state.recentlySucceededIds;
+        final succeededCount = succeededIds.length;
 
         if (succeededCount > 0) {
           if (authService.isAuthenticated) {
             // Show immediately — user is still in-app.
-            _showPublishSuccessSnackbar(context, succeededCount);
+            _showPublishSuccessSnackbar(succeededCount);
           } else {
             // Buffer for when auth is restored after re-auth redirect.
             _pendingSuccessCount += succeededCount;
@@ -2974,8 +2979,8 @@ class _UploadFailureListenerState extends State<UploadFailureListener> {
         }
 
         // Auth is now confirmed — flush buffered successes from re-auth window.
-        if (_pendingSuccessCount > 0) {
-          _showPublishSuccessSnackbar(context, _pendingSuccessCount);
+        if (_pendingSuccessCount > 0 &&
+            _showPublishSuccessSnackbar(_pendingSuccessCount)) {
           _pendingSuccessCount = 0;
         }
 
@@ -2985,7 +2990,7 @@ class _UploadFailureListenerState extends State<UploadFailureListener> {
         if (newFailedIds.isEmpty) return;
 
         final navContext = NavigatorKeys.root.currentContext;
-        if (navContext == null) return;
+        if (navContext == null || !navContext.mounted) return;
 
         final newFailures = state.uploads
             .where((u) => newFailedIds.contains(u.draft.id))
@@ -3000,15 +3005,16 @@ class _UploadFailureListenerState extends State<UploadFailureListener> {
 
 /// Shows a snackbar confirming that [count] background uploads completed.
 ///
-/// Uses [ScaffoldMessenger] via [NavigatorKeys.root] to reach the root
-/// [Scaffold] — this is required because the listener's [BuildContext] may
-/// not have a [Scaffold] ancestor when the snackbar fires during a re-auth
-/// redirect. The l10n strings are resolved from the passed-in [context]
-/// (the BlocListener's context), which always has localisation ancestors.
-void _showPublishSuccessSnackbar(BuildContext context, int count) {
+/// Returns `true` when the snackbar was shown.
+///
+/// Uses [NavigatorKeys.root] to resolve both localization and
+/// [ScaffoldMessenger] from inside the app tree. The listener itself is mounted
+/// above [MaterialApp], so its own [BuildContext] does not have localization or
+/// scaffold ancestors in production.
+bool _showPublishSuccessSnackbar(int count) {
   final navContext = NavigatorKeys.root.currentContext;
-  if (navContext == null || !navContext.mounted) return;
-  final l10n = context.l10n;
+  if (navContext == null || !navContext.mounted) return false;
+  final l10n = navContext.l10n;
   ScaffoldMessenger.of(navContext).showSnackBar(
     SnackBar(
       content: Text(
@@ -3019,6 +3025,7 @@ void _showPublishSuccessSnackbar(BuildContext context, int count) {
       behavior: SnackBarBehavior.floating,
     ),
   );
+  return true;
 }
 
 /// Shows failure bottom sheets one after another for each failed upload.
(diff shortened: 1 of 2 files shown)


rabble (commented):
Review findings:

mobile/lib/main.dart:L2961: 🟡 risk: authenticated immediate path ignores `_showPublishSuccessSnackbar` return — if root navigator context is null/unmounted (app teardown, pre-first-frame), success silently dropped. Buffer instead: `if (!_showPublishSuccessSnackbar(succeededCount)) _pendingSuccessCount += succeededCount;` — flush path already retries on next emission.

Root-context fix itself verified correct: listener mounts above `MaterialApp.router` (main.dart L2872), `NavigatorKeys.root` context sits below MaterialApp so `navContext.l10n` and `ScaffoldMessenger.of` both resolve. Dedup via `_lastKnownSucceededIds` sound — bloc's `recentlySucceededIds` is transient (replaced per success, reset to `const {}` on other emissions). No arb keys touched, l10n mirror rule N/A. `budgetCoversGateJob` exactly mirrors python guard's `matches()` (exact or `'$budget ('` prefix).

rabble:
**Reviewer-takeover fix** (repo-owner authorized, per PR runbook)

**Finding (from review):** the authenticated path in `_UploadFailureListenerState` called `_showPublishSuccessSnackbar(...)` but ignored its boolean return. When the root navigator context is unavailable, the snackbar was silently dropped (`mobile/lib/main.dart`).

**Fix pushed:** `b6c73602d` — the authenticated path now buffers into the existing `_pendingSuccessCount` when the call returns `false`, mirroring the unauthenticated buffering pattern. The buffered success replays via the existing flush once the context is available. Also rebased the branch onto `origin/main` to resolve the merge conflict GitHub reported (`check_ci_timing_budget_test.dart`, resolved by taking main's version — unrelated to this PR).

**Validation (run from worktree, `mobile/`):**
- `flutter analyze lib/main.dart test/widgets/upload_failure_listener_test.dart` → No issues found
- `flutter test test/widgets/upload_failure_listener_test.dart` → 8/8 passed, including new regression test: *"buffers success shown while authenticated when root navigator is unavailable, then replays it once the context appears"*

@NotThatKindOfDrLiz please review and merge when ready.

rabble (approved):
LGTM. Verified: NavigatorKeys.root resolves both Localizations and ScaffoldMessenger above the Navigator; success dedup via set difference is safe against the bloc's immutable-set emissions; buffer/flush can't double-count. Tests mirror the production navigator topology. One nit: PR body still mentions a CI timing-budget test fix that isn't in this diff — worth trimming before merge.

PR #6405, opened 2026-07-26
fix(profile): avoid disposed refs in profile streams

### LINKED ISSUE #6274: fix(profile): stop auto-disposed profile streams from reading a dead Ref
## Problem

The auto-disposed userProfileReactiveProvider(pubkey) can execute its first ref.watch(...) after Riverpod has already disposed the provider. That produces a high-volume unhandled FlutterError in backgrounded and rapidly changing profile/feed lifecycles.

This is not the older commentsProvider variant implied by the grouped Crashlytics title. Current 1.0.17 events resolve to userProfileReactiveProvider.

## Impact and priority

**Priority: P1 / high**

Crashlytics issue: a0f48e9c3d600db9ed922c67386db6ba  
[Open in Firebase Crashlytics](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/a0f48e9c3d600db9ed922c67386db6ba)

Snapshot from 2026-07-22:

- Last 7 days: **1,094 events / 343 impacted users**
- 2026-07-21 alone: **111 events / 50 users**
- Version 1.0.17 (build 808): **5 events / 2 users**
- Sample events occurred while the app process was backgrounded

This remains present in 1.0.17 telemetry and should not be treated as an already-fixed historical signature.

## Observed exception

    FlutterError: Cannot use the Ref of
    userProfileReactiveProvider(<full 64-character pubkey>)
    after it has been disposed

    Ref._throwIfInvalidUsage
    Ref.watch
    userProfileReactive
    package:openvine/providers/user_profile_providers.dart:47

Nostr IDs must remain complete in code, logs, and fixtures. The placeholder above avoids copying a user's identifier into this public issue; the implementation must not introduce truncation.

## Confirmed root cause

Current code is an async* Riverpod provider:

    @riverpod
    Stream<UserProfile?> userProfileReactive(Ref ref, String pubkey) async* {
      final repo = ref.watch(profileRepositoryProvider);
      if (repo == null) return;

      final cached = await repo.getCachedProfile(pubkey: pubkey);
      if (cached == null) {
        unawaited(repo.fetchFreshProfile(pubkey: pubkey));
      }
      yield* repo.watchProfile(pubkey: pubkey);
    }

An async* function body starts lazily when the returned stream is listened to. The generated family provider is auto-disposed, so a short-lived consumer can unsubscribe or invalidate the provider before the stream body begins. Its first statement then calls ref.watch(profileRepositoryProvider) on an already-disposed Ref.

A mounted check after an await cannot fix this first access: dependency capture itself happens too late.

## Expected behavior

- A profile consumer may mount and disappear immediately without producing an uncaught framework error.
- Cached-profile lookup, fresh fetch on cache miss, and repository watch behavior remain unchanged for a live consumer.
- Disposing the provider cancels its repository stream subscription.
- No user-profile request or update survives by reading a dead Riverpod Ref.

## Implementation direction

Capture the dependency synchronously in the provider body, then return a separate stream whose asynchronous work does not retain or read Ref:

    @riverpod
    Stream<UserProfile?> userProfileReactive(Ref ref, String pubkey) {
      final repo = ref.watch(profileRepositoryProvider);
      if (repo == null) return const Stream.empty();

      return _watchUserProfile(repo, pubkey);
    }

    Stream<UserProfile?> _watchUserProfile(
      ProfileRepository repo,
      String pubkey,
    ) async* {
      // Cached lookup, optional refresh, and repository stream only.
      // Do not retain Ref here.
    }

The exact helper naming is flexible. Preserve current repository semantics and do not convert this into a broad Riverpod migration.

A quick audit may identify sibling auto-disposed async* providers with the same unsafe first-use pattern, but changes to those providers belong in this issue only if the same lifecycle failure is demonstrated and can be covered without widening the PR substantially.

## TDD plan

Write the failing lifecycle regression first.

1. Override profileRepositoryProvider with a controllable fake.
2. Listen to one userProfileReactiveProvider(pubkey).
3. Immediately close the subscription or invalidate the provider before the lazy stream body gets a chance to run.
4. Capture zone/framework errors and prove the current implementation emits the disposed-Ref failure.
5. Implement synchronous dependency capture.
6. Verify:
   - immediate disposal produces no uncaught error;
   - a cached profile is emitted;
   - a cache miss triggers exactly one fresh fetch;
   - repository profile updates continue to flow;
   - disposal cancels the underlying stream subscription;
   - a null repository completes safely without work.

Use a full valid pubkey fixture; never abbreviate Nostr identifiers.

## Likely files

- mobile/lib/providers/user_profile_providers.dart
- mobile/lib/providers/user_profile_providers.g.dart after generation, if changed
- Existing or new tests under mobile/test/providers/

## Acceptance criteria

- [ ] The immediate-unlisten/invalidate regression is deterministic and passes.
- [ ] No provider-owned async callback reads Ref after disposal.
- [ ] Cached, cache-miss refresh, and live-update behavior are preserved.
- [ ] The repository stream is canceled when the provider is disposed.
- [ ] Generated Riverpod output is regenerated and committed when required.
- [ ] Focused provider tests pass.
- [ ] flutter analyze passes for the affected source and tests.
- [ ] The PR links this issue and includes the Crashlytics signature in its verification notes.

## Agent workflow

Start from a fresh worktree based on origin/main. Use a red/green TDD loop, keep the fix focused, run generation if generator-backed source changes, run the smallest relevant test suite followed by targeted analysis, commit all intended generated output, and open a Conventional Commit-titled PR against main.

NotThatKindOfDrLiz:
Closes #6274

## Summary
- Capture profile repositories synchronously before returning profile/stat streams so deferred stream bodies never read disposed refs.
- Move profile/stat async stream work into Ref-free helpers while preserving null-repo empty-stream behavior and background fetch behavior.
- Add regression tests for immediate invalidation/disposal plus profile stream, fetch, cancellation, and null-repo behavior.

## Verification
- flutter test test/providers/user_profile_providers_test.dart
- flutter analyze lib test
- flutter test test/widgets/settings_screen_test.dart test/widgets/video_feed_item/video_overlay_actions_description_test.dart test/widgets/video_feed_item/video_author_info_section_test.dart test/widgets/video_metadata/video_metadata_collaborators_input_test.dart test/widgets/profile/profile_header_widget_test.dart
- push hook: analysis, generated-file verification, changed-file tests

### DIFF
--- a/.github/ci-timing-budgets.json
+++ b/.github/ci-timing-budgets.json
@@ -3,11 +3,10 @@
     "Wall-clock budgets for Mobile CI jobs, in seconds. Enforced by the",
     "'Check CI timing budget' step in the mobile-ci gate job.",
     "",
-    "Keys must cover each mobile-ci gate job name exactly. Runtime matching",
-    "also treats a key as a matrix-job prefix, but the committed file stays",
-    "pinned to the workflow's declared names so job renames are reviewed.",
-    "A budgeted job that did not run (skipped, or renamed) is reported and",
-    "skipped, never inferred as a pass. A job with no budget entry is ignored.",
+    "Keys match a job name exactly, or as a prefix — 'Tests' covers both",
+    "'Tests' and a future 'Tests (shard N/4)'. A budgeted job that did not",
+    "run (skipped, or renamed) is reported and skipped, never inferred as a",
+    "pass. A job with no budget entry is ignored.",
     "",
     "warn  - exceeded => GitHub warning annotation on the PR. Set near the",
     "        job's observed p90 so ordinary drift surfaces early.",
@@ -24,10 +23,7 @@
   ],
   "jobs": {
     "Detect App CI Scope": { "warn": 40, "fail": 90 },
-    "Tests (shard ${{ strategy.job-index }}/${{ strategy.job-total }})": {
-      "warn": 600,
-      "fail": 780
-    },
+    "Tests": { "warn": 600, "fail": 780 },
     "Generated Files": { "warn": 260, "fail": 420 },
     "Analyze": { "warn": 150, "fail": 300 },
     "Format": { "warn": 90, "fail": 180 },
--- a/.github/ci-timing-budgets.json
+++ b/.github/ci-timing-budgets.json
@@ -3,10 +3,11 @@
     "Wall-clock budgets for Mobile CI jobs, in seconds. Enforced by the",
     "'Check CI timing budget' step in the mobile-ci gate job.",
     "",
-    "Keys match a job name exactly, or as a prefix — 'Tests' covers both",
-    "'Tests' and a future 'Tests (shard N/4)'. A budgeted job that did not",
-    "run (skipped, or renamed) is reported and skipped, never inferred as a",
-    "pass. A job with no budget entry is ignored.",
+    "Keys must cover each mobile-ci gate job name exactly. Runtime matching",
+    "also treats a key as a matrix-job prefix, but the committed file stays",
+    "pinned to the workflow's declared names so job renames are reviewed.",
+    "A budgeted job that did not run (skipped, or renamed) is reported and",
+    "skipped, never inferred as a pass. A job with no budget entry is ignored.",
     "",
     "warn  - exceeded => GitHub warning annotation on the PR. Set near the",
     "        job's observed p90 so ordinary drift surfaces early.",
@@ -23,7 +24,10 @@
   ],
   "jobs": {
     "Detect App CI Scope": { "warn": 40, "fail": 90 },
-    "Tests": { "warn": 600, "fail": 780 },
+    "Tests (shard ${{ strategy.job-index }}/${{ strategy.job-total }})": {
+      "warn": 600,
+      "fail": 780
+    },
     "Generated Files": { "warn": 260, "fail": 420 },
     "Analyze": { "warn": 150, "fail": 300 },
     "Format": { "warn": 90, "fail": 180 },
--- a/mobile/lib/providers/user_profile_providers.dart
+++ b/mobile/lib/providers/user_profile_providers.dart
@@ -6,32 +6,38 @@ import 'dart:async';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:models/models.dart';
 import 'package:openvine/providers/repository_providers.dart';
+import 'package:profile_repository/profile_repository.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'user_profile_providers.g.dart';
 
 // ignore: specify_nonobvious_property_types
 final userProfileStatsReactiveProvider =
-    StreamProvider.family<ProfileStats?, String>((ref, pubkey) async* {
+    StreamProvider.family<ProfileStats?, String>((ref, pubkey) {
       // Counts come from a public funnelcake REST call + Drift; they need only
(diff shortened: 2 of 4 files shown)


rabble (commented):
Review findings:

.github/ci-timing-budgets.json:L27: 🔴 bug: budget key "Tests (shard ${{ strategy.job-index }}/${{ strategy.job-total }})" is a literal string — workflow expressions are never expanded inside JSON. check_ci_timing_budget.py matches() requires exact name or startswith(key + " ("), so it never matches actual job names like "Tests (shard 0/4)". Key lands in "unseen" every run → only a ::notice, Tests timing budget silently unenforced. Fix: revert to the plain "Tests" key — prefix matching already covers the shard matrix.

.github/ci-timing-budgets.json: 🔵 nit: CI-budget change unrelated to the profile-stream fix; should be its own PR per repo guardrails.

mobile/lib/providers/user_profile_providers.dart:L60: 🟡 risk (pre-existing): _watchUserProfile fires unawaited(repo.fetchFreshProfile(...)) with no catchError, unlike _watchProfileStats which swallows errors. A rejection here surfaces as an unhandled async error. Not introduced by this PR, but worth aligning while touching the code.

Core Riverpod fix verified correct: repo captured synchronously via ref.watch before returning ref-free stream helpers, so deferred async* bodies never touch a disposed Ref; yield* propagates cancellation to the underlying Drift streams; null-repo empty-stream and repo-rebuild (null → non-null) behavior preserved. Regression tests cover invalidation, cancellation, cached-first emission, and null-repo.

rabble:
**Blocker: `.github/ci-timing-budgets.json` hunk breaks the Tests timing budget — please revert.**

The diff replaces the `"Tests"` key with `"Tests (shard ${{ strategy.job-index }}/${{ strategy.job-total }})"`. That string is a literal — GitHub Actions expressions are never expanded inside this JSON file; the guard script reads the file as plain JSON and `matches()` only accepts an exact key or `key + " ("` prefix against **resolved** job names.

Consequences:
1. The literal key never matches any real job name (resolved names look like `Tests (shard 1/4)`), so the Tests wall-clock budget becomes silently unenforced.
2. Every gate run prints a phantom "budgeted job did not run" ::notice for the dead key.

The prefix rule already covers sharded names: `"Tests"` matches `Tests (shard N/4)`. The updated comment text in this hunk is also wrong for the same reason.

Fix: revert the entire `.github/ci-timing-budgets.json` hunk — `git checkout origin/main -- .github/ci-timing-budgets.json` and commit. origin/main is already correct; no separate fix PR needed.

rabble (approved):
Approved after reviewer takeover. The core fix is correct: repository captured synchronously in initState, stream helpers are Ref-free, null-repo Stream.empty() preserves prior loading semantics, .g.dart hash committed, good regression tests. I pushed one fix commit (437b90fc): the branch bundled an unrelated .github/ci-timing-budgets.json hunk pinning a literal 'Tests (shard ${{...}})' key that never matches the rendered shard job names the gate reports, which would have silently dropped the Tests shards out of budget enforcement and failed main's rendered-shard budget test after merge. Restored the file to origin/main's version ('Tests' key covers shards via prefix matching). Validation: file is now byte-identical to origin/main; flutter analyze clean; test/providers/user_profile_providers_test.dart all passing. Also note the PR body doesn't mention the budget change — now moot. Please review my commit and merge when ready.

PR #6407, opened 2026-07-26
fix(video): refresh profile grid after delete

### LINKED ISSUE #6365: fix(video): profile grid keeps showing a video after you delete it
## Summary

Deleting your own video removes it for everyone else, but your own profile Videos grid keeps showing the tile, so you cannot tell the delete worked and you tap delete again.

## Environment

App version: reproduced against current `main` (commit 37e0020b5) in automated tests.
Device: any (client side logic, not platform specific).
Network: any.

## Steps to Reproduce

1. Sign in and open your own profile Videos tab. The grid is served by the REST author feed (`VideosRepository.getAuthorFeed`) or by the persisted snapshot, which is the normal case.
2. Long press one of your videos and confirm Delete.
3. Watch the grid.

Automated reproduction: `mobile/test/blocs/profile_feed/profile_feed_cubit_test.dart`, group "ProfileFeedCubit deleting a video the relay service never cached". It fails on the pre fix code with the deleted video still present.

## Actual Result

The relay accepts the kind 5 and hides the video for everyone else. The deleted tile stays on your grid. The success snackbar says it worked while the tile is still visible, so the natural response is to tap delete again, which republishes a redundant kind 5.

## Expected Result

The tile disappears from the grid as soon as the relay confirms the deletion.

## Regression?

Unknown. The tombstone machinery that this depends on landed on 2026-06-17 (#5261, #5305). The notification gap described below has been present since the deletion path was written.

## Root cause

`VideoEventService._removeVideoCompletely` (`mobile/lib/services/video_event_service.dart:1029`) does three things when you delete a video:

1. records the tombstone, always (lines 1104 to 1108)
2. emits on the `removedVideoIds` stream, always (lines 1116 to 1121)
3. calls `notifyListeners()` **only when `removedCount > 0`** (lines 1123 to 1130)

`removedCount` counts removals from VideoEventService's own in memory structures (`_eventLists`, `_authorBuckets`, `_hashtagBuckets`).

`ProfileFeedCubit`, which backs the profile Videos grid, subscribes only to the ChangeNotifier (`mobile/lib/blocs/profile_feed/profile_feed_cubit.dart:151` to `157`). It does not subscribe to the `removedVideoIds` stream (only `FullscreenFeedBloc` does).

The grid's list is fed from sources VideoEventService never caches:

* the REST author feed, via `_applyRestPage` (`profile_feed_cubit.dart:541`)
* the persisted snapshot on disk, via `_restoreFromCache` (`profile_feed_cubit.dart:237`), stored in `cache_sync.db` under key `<pubkey>:profile_videos` with no TTL

So deleting a video that is displayed but absent from VideoEventService's buckets gives `removedCount == 0`, no `notifyListeners()`, and `_onRelaySnapshot` never runs. The tombstone filter `_withoutTombstones` is therefore never re applied and the tile stays.

Had the notification fired, `_onRelaySnapshot` (`profile_feed_cubit.dart:625`) would have re filtered the full merged list and called `_persistSnapshot()`, correcting both the screen and the stored snapshot.

## Evidence from production

Creator "Jeff" (NIP-05 imrtravis, pubkey 352198449c5ff6e7e74a66a55d17ba4fc4a0a018a2d756ee589b655d823f6c27) published 54 kind 5 deletion events that were redundant re deletes of videos already tombstoned. He also filed a takedown report against his own content (Zendesk #3988) because he believed the delete had not worked.

Note on scope: Jeff deliberately deleted his 1,009 OG vines. This bug did not cause him to lose anything he wanted to keep. It is not a data loss bug. The cost is that a user cannot tell whether a destructive action worked, which drives re tapping, support tickets, and loss of trust in the delete flow.

## History

Reported by rabble on 2026-01-03 (PR #808, comment 3706835496): "The UI for delete worked, but it didn't actually delete my content.... or it kept the local cache after you deleted, which isn't right." That comment contained two bugs. The first ("didn't actually delete") was fixed by #972 and then by funnelcake#75 shipping NIP-62 on the relay. The second (local cache) was never filed, which is this issue.

Independently reproduced by mbradley on 2026-07-15: "after you delete, it does seem like your own device keeps serving the old profile and posts from local cache, so it's gone for everyone but you."

## Not in scope for this issue

The "old profile" half of the 2026-07-15 report. Deleting a video invalidates cached profile stats (`content_deletion_service.dart:280`) but does not touch kind 0 profile metadata, and account deletion already clears local data (`signOut(deleteKeys: true, deleteLocalUserData: true)` runs `UserDataCleanupService` and then redirects to /welcome). If stale profile metadata is still observed, that is a separate path and deserves its own issue.

NotThatKindOfDrLiz:
## Summary
- Notify VideoEventService listeners even when a local delete only records a tombstone and removes no cached VES entries
- Add regression coverage for tombstone-only local deletes waking ChangeNotifier consumers
- Assert the profile feed removes the tombstoned REST video and persists the filtered snapshot
- Update the Mobile CI timing budget key to match the current sharded test job name

Closes #6365

## Verification
- flutter test test/services/video_event_service_deletion_test.dart test/blocs/profile_feed/profile_feed_cubit_test.dart
- flutter analyze
- pre-push hook: analyzer + changed Dart test files

### DIFF
--- a/.github/ci-timing-budgets.json
+++ b/.github/ci-timing-budgets.json
@@ -3,8 +3,7 @@
     "Wall-clock budgets for Mobile CI jobs, in seconds. Enforced by the",
     "'Check CI timing budget' step in the mobile-ci gate job.",
     "",
-    "Keys match a job name exactly, or as a prefix — 'Tests' covers both",
-    "'Tests' and a future 'Tests (shard N/4)'. A budgeted job that did not",
+    "Keys match a job name exactly, or as a prefix. A budgeted job that did not",
     "run (skipped, or renamed) is reported and skipped, never inferred as a",
     "pass. A job with no budget entry is ignored.",
     "",
@@ -23,7 +22,10 @@
   ],
   "jobs": {
     "Detect App CI Scope": { "warn": 40, "fail": 90 },
-    "Tests": { "warn": 600, "fail": 780 },
+    "Tests (shard ${{ strategy.job-index }}/${{ strategy.job-total }})": {
+      "warn": 600,
+      "fail": 780
+    },
     "Generated Files": { "warn": 260, "fail": 420 },
     "Analyze": { "warn": 150, "fail": 300 },
     "Format": { "warn": 90, "fail": 180 },
--- a/mobile/lib/services/video_event_service.dart
+++ b/mobile/lib/services/video_event_service.dart
@@ -1119,21 +1119,15 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
       }
     }
 
-    if (removedCount > 0) {
-      Log.info(
-        'Removed video $logIdentity from $removedCount location(s) across all feeds',
-        name: 'VideoEventService',
-        category: LogCategory.video,
-      );
-      // Notify listeners to update UI immediately (optimistic update)
-      notifyListeners();
-    } else {
-      Log.info(
-        'Video $logIdentity marked as deleted (was not in any active feeds)',
-        name: 'VideoEventService',
-        category: LogCategory.video,
-      );
-    }
+    Log.info(
+      removedCount > 0
+          ? 'Removed video $logIdentity from $removedCount location(s) across all feeds'
+          : 'Video $logIdentity marked as deleted (was not in any active feeds)',
+      name: 'VideoEventService',
+      category: LogCategory.video,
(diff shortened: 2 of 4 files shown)


hm21 (approved):
LGTM

PR #6408, opened 2026-07-26
fix(follow): self-heal invalid cached pubkeys

### LINKED ISSUE #6121: fix(follow): follow silently fails on an invalid "nos" pubkey — caught, not in Crashlytics (support #6091)
## User report → root cause (correlated from support ticket)
Surfaced by correlating **support #6091** ("can't follow anybody, clicking follow does nothing", 1.0.16+804) against its embedded logs.

## Symptom
Follow button appears to do nothing. Under the hood it is **failing**, not no-op'ing.

## Root cause
The reporter's logs show, **58 times**:
```
[ERROR] [SYSTEM] (FollowRepository) Error following user: Invalid argument (publicKey): Invalid key: "nos"
[ERROR] [SYSTEM] (MyFollowingBloc) Failed to toggle follow for user: Invalid argument (publicKey): Invalid key: "nos"
```
`FollowRepository.toggleFollow(event.pubkey)` is called with `event.pubkey == "nos"` — a 3-char string that is not a valid Nostr public key, so the signer/key layer throws `Invalid argument (publicKey)`. It's caught at `lib/blocs/my_following/my_following_bloc.dart:155` and emitted as `MyFollowingStatus.toggleFailure` — the UI shows a failed toggle, which reads to the user as "the button does nothing."

`"nos"` strongly smells like a **truncated / mis-parsed identifier** — e.g. a `relay.nos.social` URL or an `npub`/`nprofile` being sliced to its first 3 chars and passed where a hex pubkey is expected. Needs tracing back to where `MyFollowingToggleRequested.pubkey` is constructed for the affected profile.

## Why crash-triage misses it
This is a **caught domain error** — logged, never rethrown, never sent to Crashlytics (correct per the repo's Reportable gating). **Zero Crashlytics events.** A crash-only triage never sees one of the loudest current user complaints.

## Suggested next step
1. Trace the pubkey source for the follow action on a profile (where does `"nos"` originate — contact-list entry, profile-screen arg, npub decode?).
2. Validate/normalize the pubkey before `toggleFollow`; surface a real error state, not a silent failure.
3. Regression test: toggling follow with a malformed pubkey yields a defined failure the UI can show, and a valid pubkey succeeds.

_Filed from support↔Crashlytics correlation pass. Related: support #6091._

NotThatKindOfDrLiz:
## Summary
- persist sanitized following lists after invalid cached pubkeys are dropped from SharedPreferences
- show the existing localized follow-failure snackbar from the profile follow button
- add regression coverage for poisoned local storage cleanup and profile toggle failure feedback

## Verification
- flutter test test/src/follow_repository_test.dart (from mobile/packages/follow_repository)
- flutter test test/widgets/profile/follow_from_profile_button_test.dart (from mobile)
- flutter analyze lib test (from mobile/packages/follow_repository)
- flutter analyze lib test integration_test (from mobile)
- pre-push hook: analyzer and changed-file tests

Closes #6121

### DIFF
--- a/mobile/lib/widgets/profile/follow_from_profile_button.dart
+++ b/mobile/lib/widgets/profile/follow_from_profile_button.dart
@@ -147,16 +147,26 @@ class FollowFromProfileButtonView extends StatelessWidget {
       );
     }
 
-    return BlocSelector<MyFollowingBloc, MyFollowingState, bool>(
-      selector: (state) => state.isFollowing(pubkey),
-      builder: (context, isFollowing) {
-        if (isFollowing) {
-          return _FollowingButton(
-            onPressed: () => _showUnfollowConfirmation(context),
-          );
-        }
-        return _FollowButton(onPressed: () => _follow(context));
+    return BlocListener<MyFollowingBloc, MyFollowingState>(
+      listenWhen: (previous, current) =>
+          current.status == MyFollowingStatus.toggleFailure &&
+          previous.status != MyFollowingStatus.toggleFailure,
+      listener: (context, state) {
+        ScaffoldMessenger.of(context).showSnackBar(
+          SnackBar(content: Text(context.l10n.followersUpdateFollowFailed)),
+        );
       },
+      child: BlocSelector<MyFollowingBloc, MyFollowingState, bool>(
+        selector: (state) => state.isFollowing(pubkey),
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
🔑 cached pubkeys now heal cleanly

PR #6415, opened 2026-07-26
fix(video-editor): keep stop-motion audio covering the composition

### LINKED ISSUE #6401: Stop-motion publish mints audio track at frame-count duration (0.375s) instead of summed frame holds
## Summary

A stop-motion video published with a sound added in the editor gets an **audio track truncated to `frameCount × (1/24s)`** — ignoring the per-frame hold durations that determine the real video length. The result is a published mp4 whose video track is ~6.4s but whose audio track is only ~0.4s, producing severely out-of-sync audio that only decodes after multiple plays.

Reported on a user video: https://divine.video/video/3df3a7a75fbc394c7680a730a8259c2dcecd27b9c1ead903558a77e323f329e7

## Evidence from the published file

`ffprobe` of the uploaded source (1080×1920, AVFoundation output):

```
Duration: 00:00:06.37, start: 0.000000, bitrate: 2338 kb/s
Stream #0:0(und): Video: h264 (High), 1080x1920, 1.41 fps, 30 tbr   ← 6.37s
Stream #0:1(und): Audio: aac (LC), 44100 Hz, stereo, fltp             ← 0.375s
[STREAM]
codec_type=audio
start_time=0.000000
duration=0.374989
[/STREAM]
```

The video is 9 unique frames stretched over 6.37s. The audio is exactly **9 × (1/24s) = 0.375s**. The extraction log confirms the same corruption upstream: `Audio details: duration=6.37s, size=66192 bytes` — but 66 KB is only ~0.4s of 16-bit stereo PCM at 44.1 kHz, so the WAV header was already lying about its duration.

## Root cause

Stop-motion capture stores the clip with `duration = frameCount × defaultFrameDuration`, where `defaultFrameDuration = 1 frame @ 24fps = 41.67ms`. The per-frame **hold durations** (which sum to the real video length) are not used.

`mobile/lib/blocs/video_recorder/video_recorder_bloc.dart:1618-1622, 1794, 1867`:

```dart
static final Duration _stopMotionPerFrame =
    StopMotionFrameOps.framesPerImageToDuration(
      StopMotionFrameOps.defaultFramesPerImage,   // == 1 → 41.67ms
    );
...
duration: _stopMotionPerFrame * frames.length,     // 9 × 41.67ms = 0.375s
```

### How that becomes a 0.375s audio track

1. **Capture** — clip cached with stale `duration` (0.375s).
2. **Editor** — user adds a sound. `_openMusicLibrary` (`mobile/lib/screens/video_editor/video_editor_screen.dart:581-587`) clamps the sound's `endTime` to `_clipEditorBloc.state.totalDuration` — still the stale 0.375s — baking the audio window at 0.375s.
3. **Render** — `StopMotionRenderService.materialize` probes the real 6.37s and updates the clip, but the audio `endTime` is already in editor meta and is not recomputed. The render's clamp ceiling (`videoContentDuration` = 6.37s) is correct, but doesn't help: 0.375s is already below it.
4. **Output** — muxed mp4 has 6.37s video, 0.375s audio. Published.

Editing frame holds later does **not** heal the audio window — `commitStopMotionFrames` recomputes the clip duration from summed holds but preserves the audio window's span verbatim (`rebaseAnchoredAudioForClipState`, `mobile/lib/widgets/video_editor/timeline_editor/video_editor_timeline_geometry.dart:191, 211`).

## Why it's not always triggered

A reloaded draft recomputes duration from summed holds (`DivineVideoClip.fromJson`), so the bug requires the clip to still carry its in-session capture duration when a sound is added. The common path — capture frames with default holds, add a sound, publish — hits it.

## Proposed fix

Two layers, defending different failure modes:

1. **Root cause — capture:** store the true summed-holds duration instead of `frameCount × defaultFrameDuration`. At `video_recorder_bloc.dart:1794` and `:1867`, use `StopMotionFrameOps.totalDuration(frames)`. (Value is identical today since all holds are `_stopMotionPerFrame`, but makes the contract correct and future-proof against pre-ingest hold edits.)
2. **Defense-in-depth — editor:** when adding/healing audio (`_openMusicLibrary`, `_healMissingAudioDurations`), compute the duration ceiling from the clip's actual frame holds for stop-motion clips rather than the possibly-stale `clip.duration`.

Regression test: capture N stop-motion frames, add a sound, publish, assert rendered audio track duration ≈ summed frame holds (within AAC frame rounding).

hm21:
## Description

A stop-motion video published with a sound got an audio track truncated to `frameCount × (1/24s)` — 9 stills published a 6.37s video muxed with a 0.375s audio track (#6401).

Two things had to be true for that:

**Capture stored a duration that ignored the frame holds.** `_appendStopMotionFrame` / the ingest path cached the clip with `_stopMotionPerFrame * frames.length`, not the summed holds. When the user then added a sound, `_openMusicLibrary` clamped its `endTime` to that stale, shorter composition end — baking a 0.375s audio window. Capture now stores `StopMotionFrameOps.totalDuration(frames)`, which is the same value today (all holds are the default) but is the correct contract and stops the clamp from ever reading a fictional end.

**A hold change restretched the video but not the sound.** `commitStopMotionFrames` recomputed the clip duration from the new holds and rebased anchored audio, but preserved each audio window's span verbatim. Stretching nine stills from 0.375s to 6.375s left the sound frozen at 0.375s. The commit now runs the rebased tracks through `growAudioToCompositionEnd`, which extends any window that reached the *old* composition end onto the new one.

Design choices worth calling out:

- **Only tracks that reached the old end grow.** A window the user deliberately trimmed shorter is theirs; growing it would silently undo the trim.
- **Never shortened here.** A composition that shrank is already clamped for display by the timeline and for output by the render, so shortening would be a second, lossy source of truth.
- **The new end is bounded twice** — by the source audio remaining after the track's `startOffset`, and by `VideoEditorConstants.maxDuration`. That is the same ceiling the track would have been given had the sound been added at the new composition length, so a grown window and a freshly-added one land in the same place.
- **The "reached the old end" comparison carries 1ms of slack.** `AudioEvent` serializes its window in whole milliseconds, but a stop-motion composition is microsecond-accurate (nine 1/24s holds are 375003µs). Without the slack, a window that round-tripped through editor history reads as a deliberate trim and stops following the composition.
- **`identical` short-circuit on the commit path.** When nothing moved, the commit keeps calling `setClipState` instead of `setClipAndAudioState`, so a frame edit with no audio doesn't start writing audio into every history entry.

**Related Issue:** Closes #6401

## Out of Scope

The issue's proposed defense-in-depth item 2 (recompute the ceiling from frame holds inside `_openMusicLibrary` / `_healMissingAudioDurations`) is not included. With capture storing the summed-holds duration, `clip.duration` is no longer stale at the point those paths read it, so the extra recomputation would be dead weight rather than a second line of defense.

## Verification

- `flutter test test/widgets/video_editor/stop_motion/stop_motion_frame_commands_test.dart test/widgets/video_editor/timeline_editor/video_editor_timeline_geometry_test.dart` — 114 passing, including a new widget test that commits a hold change over a clip with a sound and asserts the persisted window grew to 4.5s, plus one asserting a user-trimmed window is left alone.
- `flutter test test/blocs/video_recorder/video_recorder_bloc_test.dart` — 221 passing (pre-push hook).
- `dart format` — clean.

Not yet verified on device — the audio-follows-holds path still wants a manual publish check on a real stop-motion capture before this leaves draft.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/video_editor/clip_editor/clip_editor_bloc.dart
+++ b/mobile/lib/blocs/video_editor/clip_editor/clip_editor_bloc.dart
@@ -5,6 +5,7 @@ import 'package:models/models.dart' show AudioEvent;
 import 'package:openvine/constants/video_editor_timeline_constants.dart';
 import 'package:openvine/models/divine_video_clip.dart';
 import 'package:openvine/models/stop_motion/stop_motion_frame_ops.dart';
+import 'package:openvine/models/video_editor/composition_duration.dart';
 import 'package:openvine/models/video_editor/editor_overlay_snapshot.dart';
 import 'package:openvine/observability/reportable_error.dart';
 import 'package:openvine/services/audio_extraction_service.dart';
--- a/mobile/lib/blocs/video_editor/clip_editor/clip_editor_state.dart
+++ b/mobile/lib/blocs/video_editor/clip_editor/clip_editor_state.dart
@@ -213,8 +213,7 @@ class ClipEditorState extends Equatable {
   final Set<int> selectedFrameIndexes;
 
   /// Total wall-clock duration of all clips (respecting trim and playback speed).
-  Duration get totalDuration =>
-      clips.fold(Duration.zero, (sum, clip) => sum + clip.playbackDuration);
+  Duration get totalDuration => compositionDuration(clips);
 
   /// Creates a copy with the given fields replaced.
   ClipEditorState copyWith({
--- a/mobile/lib/blocs/video_editor/timeline_overlay/timeline_overlay_bloc.dart
+++ b/mobile/lib/blocs/video_editor/timeline_overlay/timeline_overlay_bloc.dart
@@ -644,28 +644,45 @@ class TimelineOverlayBloc
 
     final tracksById = {for (final track in state.audioTracks) track.id: track};
 
+    // The preview player builds each sound's window from the overlay item, and
+    // only re-reads it when audioTracksPlayerRevision moves — the track list
+    // itself is untouched here, so Equatable would otherwise let the player
+    // keep the window it was handed before this re-clamp.
+    var soundWindowChanged = false;
     final updated = <TimelineOverlayItem>[];
     for (final item in state.items) {
-      if (item.startTime >= totalDuration) continue;
+      final isSound = item.type == TimelineOverlayType.sound;
+      if (item.startTime >= totalDuration) {
+        soundWindowChanged = soundWindowChanged || isSound;
+        continue;
+      }
 
-      final track = item.type == TimelineOverlayType.sound
-          ? tracksById[item.id]
-          : null;
+      final track = isSound ? tracksById[item.id] : null;
       final clampedEnd = _clampEnd(
         track == null ? item.endTime : _soundWindowEnd(track, totalDuration),
         totalDuration,
       );
-      if (clampedEnd <= item.startTime) continue;
+      if (clampedEnd <= item.startTime) {
+        soundWindowChanged = soundWindowChanged || isSound;
+        continue;
+      }
 
-      updated.add(
-        clampedEnd == item.endTime ? item : item.copyWith(endTime: clampedEnd),
-      );
+      if (clampedEnd == item.endTime) {
+        updated.add(item);
+        continue;
+      }
+      soundWindowChanged = soundWindowChanged || isSound;
+      updated.add(item.copyWith(endTime: clampedEnd));
     }
 
     emit(
       state.copyWith(
         items: _recalculateRows(updated),
         timelineMarkers: _clampMarkers(state.timelineMarkers, totalDuration),
+        audioTracksPlayerRevision:
+            soundWindowChanged && !event.isClipTrimDragging
+            ? state.audioTracksPlayerRevision + 1
+            : state.audioTracksPlayerRevision,
       ),
     );
   }
--- a/mobile/lib/blocs/video_editor/timeline_overlay/timeline_overlay_bloc.dart
+++ b/mobile/lib/blocs/video_editor/timeline_overlay/timeline_overlay_bloc.dart
@@ -270,7 +270,7 @@ class TimelineOverlayBloc
       startTime: hasWindow ? track.startTime : .zero,
       // No valid window → span the whole video. [total] is guaranteed > 0 by
       // _onUpdateItems (transient-zero is substituted with maxDuration there).
-      endTime: hasWindow ? _clampEnd(endTime, total) : total,
+      endTime: _clampEnd(_soundWindowEnd(track, total), total),
       label: track.title ?? track.pubkey,
       maxDuration: sourceDuration != null
           ? sourceDuration - track.startOffset
@@ -627,6 +627,14 @@ class TimelineOverlayBloc
   /// Clamp every overlay item so its visible region fits within
   /// [0, totalDuration]. Items that end up with zero visible duration
   /// are removed.
+  ///
+  /// A sound bar is re-derived from its [AudioEvent] rather than only clamped,
+  /// because a clamp cannot undo itself. Undo/redo builds the bars against the
+  /// total that was current *before* the navigation — the restored clips only
+  /// reach [ClipEditorBloc] afterwards — so redoing an edit that lengthened the
+  /// composition hands this handler a bar already truncated to the old, shorter
+  /// end. Clamping it again would leave the sound playing short under a longer
+  /// video (#6401).
   void _onTotalDurationChanged(
     TimelineOverlayTotalDurationChanged event,
     Emitter<TimelineOverlayState> emit,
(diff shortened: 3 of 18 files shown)


rabble (commented):
Review findings:

mobile/lib/widgets/video_editor/timeline_editor/controls/video_editor_timeline_clip_controls.dart:L334: 🟡 risk: `_duplicateClip` lengthens the composition via plain `setClipState`, so a sound clamped to the old composition end still publishes short — the same #6401 mux shape this PR fixes for stop-motion holds. Same on `_setPlaybackSpeed` (L205, slowing a clip) and trim-extension commits. `setLengthenedClipState`'s docstring excludes these via the generic-writer argument, but each commit could call `setLengthenedClipState` individually exactly like `commitStopMotionFrames` does. Route the lengthening clip commits through it, or confirm a tracked follow-up so the gap is deliberate.

mobile/lib/widgets/video_editor/timeline_editor/video_editor_timeline_geometry.dart:L252: 🔵 nit: the ±1ms `_audioWindowResolution` test is symmetric, so a window *overhanging* `previousDuration` by ≤1ms is treated as covering and grown, contradicting the docstring's "one that overhangs the old end … was placed by the user". Practically unreachable (overhangs come from shrinks, which are macroscopic), but an asymmetric check (`endTime <= previousDuration + slack`) would match the documented contract.

Verified against head 4048a36: capture-side duration now `StopMotionFrameOps.totalDuration` at both sites (no other `perFrame * length` caching remains), `clipWithFrames`/`fromJson` keep clip duration derived from frames (old drafts healed), the ms-serialization slack rationale checks out (`inMilliseconds` truncates 375003µs → 375ms), source-end/maxDuration ceilings are tested, and the overlay redo re-derivation is covered by the new bloc tests. Rest of the change reads clean.

rabble:
Reviewer-takeover fix (authorized by repo owner) for the remaining lengthening paths.

**Finding:** this PR routed the stop-motion frame commit and the camera/clips-picker sync through `setLengthenedClipState`, but three other operations that also lengthen the composition still called plain `setClipState`, leaving a sound clamped to the old end — the same publish-short bug shape (#6401).

**Fix (c3a88a1d1):**
- `_duplicateClip` (appends a copy — always lengthens) → `setLengthenedClipState`
- `_setPlaybackSpeed` (slowing a clip down lengthens; speed-up passes through unchanged since the grow only fires when the composition got longer) → `setLengthenedClipState`
- Trim-drag-end in `video_editor_timeline.dart` (dragging a handle back out lengthens) → `setLengthenedClipState`, using the clips captured at drag start as `previousClips`
- Updated the `setLengthenedClipState` doc comment to match the widened caller set

Deliberately left on plain `setClipState`: delete, reorder, multi-select removal, and merge (shrink/no-total-change paths), plus the transition sheet (measures output with `TransitionTimelineMap`, as the PR already documented).

**Tests added:** one per rerouted path, colocated with the PR's existing tests — duplicate and speed-down in `video_editor_timeline_clip_controls_test.dart`, trim-extension in `video_editor_timeline_test.dart`. Each asserts a sound covering the old composition end is grown onto the new end in the committed history entry.

**Validation:** `flutter analyze` clean on touched files; targeted suites pass — clip controls, timeline, extensions, stop-motion frame commands, timeline geometry, timeline overlay bloc (236 tests); pre-push hook analyze + tests green.

@hm21 please review and merge when ready.

NotThatKindOfDrLiz:
Takeover cleanup pushed in 79173c72de.

What changed:
- moved the plain clip composition duration calculation into mobile/lib/models/video_editor/composition_duration.dart
- wired ClipEditorState.totalDuration and setLengthenedClipState through that shared helper so the old-end comparison cannot drift from the duration used when adding sounds
- aligned _audioSourceEnd with the render/add-sound millisecond convention by truncating source duration with .toInt()

Verification I ran locally:
- flutter test test/widgets/video_editor/timeline_editor/video_editor_timeline_geometry_test.dart test/extensions/video_editor_extensions_test.dart test/widgets/video_editor/stop_motion/stop_motion_frame_commands_test.dart test/widgets/video_editor/timeline_editor/controls/video_editor_timeline_clip_controls_test.dart test/widgets/video_editor/timeline_editor/video_editor_timeline_test.dart test/blocs/video_editor/timeline_overlay/timeline_overlay_bloc_test.dart — 236 passing
- flutter analyze lib test integration_test — no issues
- pre-push hook — merge-conflict check, analyze, and changed-file test set passed; 478 tests passing

@hm21 please do the remaining manual publish verification before this merges: capture a real stop-motion clip on device, add a sound, change a hold, publish, and confirm the muxed audio covers the full video duration. After that, please update the PR Verification section with the device/result details and merge when you are satisfied.

hm21:
@rabble @NotThatKindOfDrLiz 
Verified on device at `1af6b38e6`, which includes both takeover commits: stop-motion capture → add sound → hold change → the sound follows the composition, and undo/redo now round-trips correctly. Caveat on scope — I exercised the stop-motion flow, not the duplicate/speed/trim paths from `c3a88a1d1` separately; those rest on their unit tests.

That device pass turned up one more bug, fixed in `1af6b38e6`: after redo the sound bar came back at full length but played short. Restoring the bar was only half of it — the preview player builds each window from the overlay item and is re-synced only when the track list, `audioTracksRevision` or `audioTracksPlayerRevision` moves, and a re-clamp touches none of those. The bump now fires when a sound window actually moves, and is suppressed while a clip-trim drag is in flight, since that emits clips live and would otherwise reschedule native audio every frame. Two bloc tests cover it.

rabble (approved):
LGTM. Sound windows now follow composition growth across all lengthening paths (hold change, camera/picker sync, duplicate, speed-down, trim-extend, redo); the +/-1ms symmetric end test correctly avoids eating user trims; all five commit paths pass correct pre-edit previousClips; player re-sync loop verified closed. Nits (non-blocking): (1) transition removal/change still uses plain setClipState, so a sound covering the old end won't grow there — the residual gap is acknowledged in the setLengthenedClipState doc; please file a tracking issue so it isn't lost. (2) video_editor_timeline_geometry.dart:312 uses (duration*1000).toInt() while _soundItem uses .round() for the same conversion — harmless today, use .round() for consistency. (3) timeline_overlay_bloc.dart:660 re-derives sound bar ends during a trim drag against deliberately-stale audioTracks — cosmetic flicker risk only; consider skipping re-derivation while isClipTrimDragging.

Issue #6215, opened 2026-07-20
fix(account): account deletion runs irreversible steps and reports success without relay confirmation


dcadenas:
## Summary

Account deletion performs irreversible, unrecoverable actions **before** confirming the deletion actually committed, and reports success based only on a socket write rather than a relay acknowledgement. A user can be told their account was deleted, have their keys destroyed, and still have a live account they can no longer delete.

This is two closely related defects at the same commit point, so they are filed together and should be fixed in one PR.

## Defect 1 — deletion side effects are reported successful without relay confirmation

`AccountDeletionService` uses fire-and-forget publish for both destructive event kinds:

- `mobile/lib/services/account_deletion_service.dart:104` — the NIP-62 vanish (kind 62)
- `mobile/lib/services/account_deletion_service.dart:189` — the NIP-09 deletion sweep (kind 5)

Both call `publishEvent(...)`, whose success means only that the WebSocket accepted the frame. From `publishEvent`'s own doc (`mobile/packages/nostr_client/lib/src/nostr_client.dart:421`):

> `PublishSuccess` — the event was broadcast to at least one relay.

That is socket-level, not persistence-level. The repo already has the correct primitive, `publishEventAwaitOk` (`nostr_client.dart:498`), whose doc names this exact case:

> Unlike `publishEvent`, this method does NOT consider the publish successful just because the WebSocket accepted the frame. It waits for the relay to respond with an `OK true` message (NIP-20)... Use this for operations where the caller needs to know the event is actually persisted on at least one relay — **deletions, critical profile updates**, etc.

**The inconsistency is the sharp part:** the *per-video* delete path already uses `publishEventAwaitOk` (`mobile/lib/services/content_deletion_service.dart:221`). So deleting one video waits for relay confirmation, but deleting your **entire account** does not.

### Why this is unrecoverable

On a reported success, the dialog signs the user out and destroys their keys (`mobile/lib/widgets/delete_account_dialog.dart:347`):

```dart
await authService.signOut(deleteKeys: true, deleteLocalUserData: true);
```

If the relay never persisted the vanish (e.g. it rejected the event, or no `OK` ever arrived), the account remains live on relays while the user's key is gone. The user has no key left to retry the deletion. This is confirmed reachable by reading; it is not yet correlated to a specific production report.

[...]

Issue #4881, opened 2026-06-03
fix(account): registered account deletion fails when Keycast deletion/session refresh fails


AleyshaRose:
### Summary

Multiple users have requested their accounts be deleted, when given instructions on how to do so, they say there is an error message and it fails. 

### Environment

Unknown

### Steps to Reproduce

Attempted to delete account, got errors. 

### Actual Result

Account does not get deleted

### Expected Result

Account is deleted on request 

### Regression?

Yes

### Evidence

Npubs of affected users: 
- npub1z4kazwsl3fygqdl6rdp6my62tevxgjsadcddv6t6qtpwjwutqyasz8zvt3
- npub1x3f49fnhl66p6cj938epdyncm026ykafgpnr7gpqzqwnpgy7l9hsdwf6ph
- One more user, npub TBC

NotThatKindOfDrLiz:
First-pass repo repro on fresh origin/main worktree (604cc0bd66):\n\n- Active targeted tests pass: account_deletion_service_test, auth_service_delete_keycast_account_test, nostr_settings_screen_test.\n- Existing skipped account-deletion flow tests are stale rather than useful repros: they target old SettingsScreen wiring / missing provider overrides, not the current NostrSettingsScreen delete-account path.\n- I added a temporary focused widget repro locally and removed it after running. It confirms a concrete current failure mode: if NIP-62 succeeds but registered-user Keycast deletion returns an error (for example unauthorized/expired token), executeAccountDeletion shows deleteAccountServerDeletionFailed and deliberately does not sign out or remove local keys. This matches the user-facing report shape: error message, account not deleted.\n\nCurrent failing boundary candidate: Keycast account deletion/session refresh for registered users, not NIP-62 signing/publishing. Code path: delete_account_dialog.dart blocks completion when authService.deleteKeycastAccount() fails and authService.isRegistered is true.\n\nRecommended next step: reproduce against a real affected registered account or Keycast logs for the listed npubs to confirm whether deleteAccount is returning 401/session-expired/server error. If confirmed, fix should target deleteKeycastAccount/session refresh or define product behavior for partial deletion failure.

NotThatKindOfDrLiz:
Narrowed after first-pass repo repro.

The current mobile failure mode is not a generic account-deletion problem. The concrete repro is:

1. NIP-62 deletion succeeds.
2. The user is registered / Keycast-backed.
[...]

PR #6417, opened 2026-07-26
fix(account): refuse to start account deletion that cannot finish

realmeylisdev:
## What

Account deletion publishes irreversible NIP-62 vanish and NIP-09 kind-5 deletion requests before it can attempt server-side Keycast account deletion. Keycast account deletion is only authorized while the access token still carries the first-party fact it was minted with; refreshed tokens keep RPC access but lose that delete authorization. The old flow could therefore publish the irreversible Nostr side, hit a 403 on Keycast, and leave the user signed in with a live account.

This PR refuses to start the destructive phase unless the current Divine OAuth session is known to be able to finish the server-side deletion.

## Changes

- Stamps `KeycastSession.grantType` when sessions are saved.
- Marks refresh-token sessions as refresh-minted and blocks account deletion on them.
- Fails closed for legacy sessions with no grant marker, sessions without expiry, and sessions too close to expiry.
- Re-checks readiness after the username-burn step and before Nostr deletion publishes.
- Keeps anonymous/imported/Amber/bunker users ungated because they have no Keycast account to delete.
- Adds a timeout and explicit 403 handling to Keycast `deleteAccount`.
- Keeps the server/network delete failure copy separate from the re-auth-required copy, choosing between them from a typed `requiresReauthentication` flag on `DeleteAccountResult` rather than from the error text. Keycast answers the first-party 403 with its own prose, which the client returns verbatim whenever the body parses, so no phrase in it is safe to match on.
- Gives the post-publish refusal its own string. `deleteAccountReauthRequired` ends with "Nothing has been deleted yet", which holds at the two pre-flight gates but not after the vanish and the kind-5 sweep have been confirmed by a relay. `deleteAccountServerDeletionRequiresReauth` is translated into every locale.
- Collapses the duplicated `NostrClient.queryEvents` into a delegation to `queryEventsDetailed` and drops the unread, per-client `lastQueryTimedOut` that the record already replaces.
- Makes the sweep query report timeout/no-relay status instead of treating every empty list as a healthy empty account.
- Counts only OK-confirmed kind-5 batches and reports an unverified outcome when the sweep query failed or some queried content did not get an OK-confirmed deletion request.
- Keeps the kind-5 and kind-62 self-exclusion and addressable `a` tag fixes from the original PR.

## Split out

The Settings hub and Support Center deletion entry points were removed from this PR. They are independent discoverability work and should land in a follow-up with their own widget coverage and localization/product-copy pass.

## Tests

- `flutter test test/services/account_deletion_service_test.dart test/services/auth_service_expired_session_downgrade_test.dart test/widgets/delete_account_dialog_test.dart`
- `flutter test packages/keycast_flutter/test/oauth_client_test.dart`
- `flutter test packages/nostr_client/test/src/nostr_client_test.dart` (297) and `flutter test packages/keycast_flutter` (235)
- `flutter test test/services test/widgets/delete_account_dialog_test.dart` (3448)
- `flutter analyze lib/services/account_deletion_service.dart lib/services/auth_service.dart lib/widgets/delete_account_dialog.dart packages/keycast_flutter/lib/src/models/keycast_session.dart packages/keycast_flutter/lib/src/oauth/oauth_client.dart packages/nostr_client/lib/src/nostr_client.dart packages/nostr_sdk/lib/nostr.dart test/services/account_deletion_service_test.dart test/services/auth_service_expired_session_downgrade_test.dart test/widgets/delete_account_dialog_test.dart`

The pre-push hook also ran changed-file analyze and the changed-file account deletion tests successfully.

## What's left

- NIP-98 proof-of-key remains the real long-term fix; it needs the Keycast wire-contract decision tracked in divinevideo/keycast#323.
- The Settings/Support discoverability work should be a separate PR.
- The already-stranded users still need remediation outside this mobile-code fix.

Refs #6335, #4881, #6215 · keycast#323 · funnelcake#652


realmeylisdev:
@NotThatKindOfDrLiz — audited #6417 against your five steps, item by item. It now covers four; here's the honest scorecard including where I went against you.

**1. Discoverability entry at top-level Settings and Support Center — done.** Authenticated-only entries in both, pushed in `f8778924b`.

**2. Reuse the existing flow and copy; no second implementation — done, and enforced.** Both `nostrSettingsDeleteAccount` and `nostrSettingsDeleteAccountSubtitle` are untouched and reused verbatim. Rather than copy the wiring into three screens, I extracted it once into `startAccountDeletionFlow()`; all three entry points call it, so the confirmation gate, burn opt-in, and step ordering can't drift between them. The shared modal spinner moved to `ModalProgressOverlay` instead of being duplicated.

**3. Widget tests — partially done.** Support Center is covered: entry present when authenticated, absent when signed out, copy resolved through l10n not a hardcoded string. The absence test was verified to fail when the auth gate is removed.

Not done: the top-level Settings assertions. `SettingsScreen` can't currently be pumped in a widget test — it needs `LocaleCubit`, `sharedPreferencesProvider`, `InviteStatusCubit` and `PackageInfo` wired, which is why those three tests were `skip: true` before I got here. I left them skipped rather than delete them or claim coverage I don't have. The entry is code-reviewed but not test-covered, and I'd rather say so than pad the PR.

**4. "Do not change OAuth/Keycast deletion behavior in this PR" — I went against this, deliberately.** Your stated reason was that the path is "plausible but not proven by this ticket." It's now proven: prod Cloud Logging has this reporter with 10 denials, and their kind-62 vanish is stamped 30 seconds before each 403. Across 30 days that's 275 denials / 57 users / 48 who never completed. Your own recommendation allowed for this — "unless reproduction proves they are required" — and reproduction is item 5, which you asked for.

What I did **not** touch is the specific behaviour you flagged: the block on signing out registered users when Keycast deletion fails is intact at `delete_account_dialog.dart:611`. The change is upstream of it — the flow now refuses to *start* when the session can't authorize the delete, so users stop having content destroyed before the refusal.

**5. Reproduce with an expired/flaky OAuth session — done.** Local keycast, same account: original token → 200; after one `grant_type=refresh_token` → 403, `first_party` absent from the UCAN facts. That's the mechanism, and it's filed server-side as divinevideo/keycast#323.

**OK-true confirmation — also in this PR rather than decided separately.** Same reasoning: both destructive publishes reported success on a socket write, so we were telling users content was deleted on no evidence. Given the ordering defect this compounds it, so I didn't want to ship the gate while leaving the success signal untrustworthy. Closes #6215. Say the word if you'd rather I pull it into its own PR.

Two things I'd value your read on: the new copy in 18 locales (it has to say "sign in again" without implying we deleted something we didn't — NIP-62 calls the procedure legally binding in some jurisdictions), and whether the Settings-hub test gap is acceptable to merge on or should block.

realmeylisdev:
## Device test plan

Manual test plan for this PR. The non-destructive tests (A, B, C) cover the actual fix; do those first.

## ⚠️ Read this before you tap anything

**This flow is irreversible and it reaches the public internet.** A completed deletion signs and broadcasts a NIP-62 "request to vanish" to `relay.damus.io`, `nos.lol`, `relay.primal.net` and others. Those relays are not ours, the event is signed by the account's key, and NIP-62 describes the procedure as legally binding in some jurisdictions. You cannot un-publish it — NIP-62 explicitly says a later kind-5 against a vanish request has no effect.

Rules:
- **Throwaway accounts only.** Never your own account, never anything with content you want.
- Give each throwaway a **unique @divine.video handle you don't mind losing** — the burn toggle releases it permanently.
- Expect the account to keep working on `api.divine.video` afterwards. That's a known gap, not a bug in this PR (see the last section).

Most of the important tests below are **non-destructive** — Tests A, B and C never publish anything. Do those first.

## Build and devices

Branch `fix/6335-account-deletion-preflight-gate`, head `49349bb30`.

The original report is **Samsung SM-A556B, Android 16, SDK 36, app 1.0.17+668** — reproduce there if you can. Also cover one iOS device, because the progress spinner has no hardware-back escape on iOS.

## Accounts you need

The behaviour intentionally differs by cohort, and getting this wrong is how you'd miss the main regression risk.

| Cohort | How to make one | Why it matters |
|---|---|---|
| **1. divineOAuth, server-minted key** | Sign up with **email** in the app. Do not import an nsec. | The affected cohort. The device never holds the private key, so Keycast is the signer. |
| **2. divineOAuth, local key** | Start anonymous, then Settings → **Secure Account** and upgrade. | Also hits the server refusal, but keeps its own signer. |
| **3. Anonymous / imported nsec** | Fresh install → skip signup, or import an nsec. | **Must never be gated.** Blocking these would stop them deleting their own content. |
| **4. Amber or bunker (NIP-46)** | Sign in with a remote signer. | Same — must never be gated. |

## The lever: forcing the gate to fire

**A freshly signed-in account will not reproduce the bug.** The failure needs a *stale* Keycast session, which is exactly why it hit returning users and not testers. Two ways:

- **Natural (most faithful):** sign in on cohort 1, then come back **more than 24 hours later** and attempt deletion.
- **Fast (device clock):** `isExpired` is computed purely from the local clock (`DateTime.now().isAfter(expiresAt)`), so **advance the device date by 2 days** in system settings, then reopen the app and attempt deletion. Set the clock back afterwards.

If a session has no expiry set, neither works and the gate will pass — note it and move on.

---

## Test A — the new entry points (non-destructive)

Stop at the confirmation dialog. **Do not type DELETE.**

Signed in, any cohort:
1. Profile → gear → **Settings**. Scroll to the bottom. There is a red **"Delete Account and Data"** entry with the subtitle *"Sends deletion requests for your content and signs you out on this device…"*.
2. Settings → **Support Center**. The same red entry is at the bottom of that list.
3. Settings → **Nostr Settings** → scroll → Danger Zone. Still there, unchanged.
4. Tap each of the three in turn. Each shows a brief spinner, then **the same** confirmation dialog — same title, same body, same DELETE text field, same burn-username toggle if the account owns a handle. **Cancel each time.**

Then sign out (or use a fresh install) and confirm **all three** entries are gone — the Settings and Support Center entries must not appear to a signed-out user.

**Fails if:** any entry is missing, the three dialogs differ from each other, the icon or text isn't red, or a signed-out user sees an entry.

## Test B — the gate refuses to start, and destroys nothing (the core fix)

**Cohort 1, with a stale session** (use the lever above).

1. Open any of the three entry points, type `DELETE`, leave the burn toggle **off**, and confirm.
2. Expected: a red snackbar reading exactly

   > **Sign in again to delete your account. Nothing has been deleted yet.**

3. You must **not** see the old wording *"…check your connection and try again."*
4. You are **still signed in**. The app does not navigate to the welcome screen.
5. Now verify nothing was published. Either:
   - grab logs (Support Center → **Save Logs**) and confirm there is **no** `Creating NIP-62 event` line and **no** `Batch deletion for kind` line after your attempt; or
   - query a relay for that pubkey and confirm no kind-62 or kind-5 exists.
6. Sign out and back in (which refreshes authorization), retry the deletion — it should now proceed past the gate.

**Fails if:** any kind-62 or kind-5 event appears, the handle gets released, or the snackbar blames the network.

## Test C — non-OAuth users are NOT blocked (highest regression risk)

**Cohorts 3 and 4.** These have no Keycast account, so the gate must never fire for them.

Attempt deletion and confirm you get the normal flow — progress dialog, then a real outcome. You must **never** see *"Sign in again to delete your account."*

**Fails if** an anonymous, imported-nsec, amber or bunker user is told to sign in again. That would mean they can't delete their own content, which is worse than the bug we're fixing.

## Test D — a real deletion, end to end (destructive)

**Cohort 1 or 2, freshly signed in**, with a throwaway that has **at least 2 videos, one of which you have edited at least once**, plus a like and a repost. The `a` tag is emitted for any addressable video; editing one is what creates the *older versions* that the `a` tag is there to delete, so it's what makes the fix observable.

1. Note the pubkey (Settings → copy public key) and the video IDs.
2. Delete, with the burn toggle **on** if the account owns a handle.
3. Expected: progress dialog, then

   > **Deletion requests sent. You're signed out on this device.**

   and you land on the welcome screen, signed out.
4. Query a relay for `authors: [<pubkey>], kinds: [5, 62]` and check:
   - exactly **one** kind-62 with `["relay","ALL_RELAYS"]`;
   - kind-5 events carrying **`a` tags** of the form `34236:<pubkey>:<d-tag>` for the video kinds — not only `e` tags. **This is the new behaviour;** its absence is a fail.
   - **no** kind-5 whose `k` tag is `5` or `62`. The sweep must not target its own deletion events. **This is the new behaviour;** presence is a fail.
5. Confirm the released handle is claimable again.

## Test E — no relays reachable (`publishEventAwaitOk`)

Same setup as D, on a second throwaway.

Put the device in **airplane mode**, then attempt deletion. Expected: it does **not** report success. Previously a publish "succeeded" as soon as a socket accepted the frame; it now waits for a relay `OK`, so with no relay there is nothing to confirm. You should get a failure outcome and remain signed in.

Restore the network and retry — it should complete.

**Fails if** you see the success snackbar with no network.

## Test F — the sweep couldn't list your content

Harder to stage deliberately; test opportunistically. If the relay query for existing posts fails while the vanish still publishes, the snackbar must be the qualified one:

> **Account deletion requested. We couldn't reach the relays to list your existing posts, so those were not individually requested for deletion.**

Not the plain success line. In logs this pairs with `relay query FAILED — sweep will cover nothing`.

## Test G — copy in another language

Switch to **German** or **Spanish** (Settings → General → language) and re-run Test A and Test B. Both new strings are translated in all 18 locales. Check they fit the snackbar without clipping and read naturally — the re-auth string must say "sign in again" without implying anything was deleted.

## Test H — iOS spinner

On iOS, start a deletion and watch the progress dialog. It is deliberately non-dismissible and iOS has no hardware back, so confirm it always terminates on its own (success or error) and never strands you. Flag anything that hangs — the delete call now has a 30s timeout, so an indefinite hang is a bug.

---

## Getting logs

Support Center → **Save Logs** → share to yourself. Useful lines:

| Line | Means |
|---|---|
| `Account deletion blocked before publishing: Keycast session cannot authorize deletion` | Gate fired (Test B) — correct |
| `Deletion blocked before publishing: session cannot authorize` | Same event, UI side |
| `Account deletion readiness check failed:` | Readiness check errored; it fails closed, so this also blocks |
| `sweep targets after excluding kinds 5, 62` | New kind exclusion is active |
| `Batch deletion for kind N not confirmed:` | A kind-5 got no relay `OK` |
| `NIP-62 deletion request confirmed by relay(s): [...]` | Vanish genuinely confirmed, not just written to a socket |
| `relay query FAILED — sweep will cover nothing` | Test F condition |

`adb logcat` works too if you prefer.

## Please do NOT file these — known, out of scope for this PR

These are real and tracked; reporting them against #6417 will just add noise.

- **The account still exists after deletion.** `GET api.divine.video/api/users/<pubkey>` keeps returning 200 with follower counts, and `total_events` can even *grow* (the deletion events are themselves events). Only some read paths apply the vanish filter.
- **Videos still play / blobs still exist.** divine-blossom's media-erasure endpoint has no callers, so blobs and CDN copies survive. Reported on divine-funnelcake#652.
- **Nothing is ever physically erased on the relay.** The vanish sweep has never been scheduled. Same issue.
- **divine-web still shows the account** as a live, followable profile.
- **A returning divineOAuth user still can't complete deletion**, they're now told to sign in again first. Fixing that properly is server-side: divinevideo/keycast#323.
- **Preloaded/Vine-reclaimed accounts fail with a server error.** keycast#296.
- **The Settings-hub entry has no widget test** — `SettingsScreen` can't be pumped without more provider wiring. Manual coverage via Test A is deliberate for now.

realmeylisdev:
@NotThatKindOfDrLiz — following up on your point 4, with something better than an argument.

I went to check that I hadn't disturbed the behaviour you flagged — "the current behavior intentionally blocks sign-out for registered users if Keycast account deletion fails" — and found that **nothing was protecting it.**

The only assertion of it in the repo is `account_deletion_flow_test.dart:171`:

```dart
// Verify user was NOT signed out
verifyNever(() => mockAuthService.signOut(deleteKeys: true));
```

It sits in a group closed with `}, skip: true);`. And even if that group were revived it could not have failed: production calls `signOut(deleteKeys: true, deleteLocalUserData: true)`, and `verifyNever` on the one-argument shape doesn't match an invocation carrying the second, so it passes against a call that never happens. Your invariant was held in place by your memory and your review, and by nothing in CI.

So `836676e07` pins it properly, both halves of the guard:

- **`does not sign out a registered user when keycast deletion fails`** — content deletion succeeds, Keycast fails, `isRegistered` is true. Asserts `signOut` is never called *with the exact shape production uses*, and that the user sees the server-failure message rather than success. Deleting the early `return` makes this fail.
- **`still signs out a non-registered user when keycast deletion fails`** — the `&& isRegistered` conjunct is more load-bearing than it looks. `oauthClientProvider` is non-nullable, so every anonymous, imported-nsec, amber and bunker user also reaches `getSessionOrRefresh()`, has no refresh token, and gets `(false, …)` back. Without that conjunct they'd all be permanently unable to delete their own content. Dropping it makes this fail.

I verified both by making each change in turn and watching the corresponding test go red, then restoring.

This doesn't undo the fact that I went against your point 4 — the pre-flight gate does change that path, and it shrinks the set of users who ever reach your `:611` guard even though the guard's code is untouched. But it means you can now verify I didn't break your invariant instead of taking my word for it, and the next person to touch this can't quietly regress it either.

Still outstanding and yours to rule on: the top-level Settings entry has no widget test. `SettingsScreen` needs roughly six provider overrides plus a `package_info_plus` channel mock to become pumpable, and channel mocking in the merged isolate already cost this PR a cascade failure across unrelated shards once today. I'd rather do it as its own PR where a shard failure is diagnosable in isolation — but if you want it in here before merge, say so and I'll do it.

rabble (commented):
Review findings:

`mobile/lib/services/auth_service.dart:2901`: 🟡 risk: the gate's core invariant — "a stored, unexpired session ⟺ the original fact-carrying token" — is breakable without any expiry. `KeycastRpc` invokes `onTokenRefresh` on *any* 401 (`keycast_rpc.dart`, wired at `auth_service.dart:642`, `:761`, `:2735`, `:4399`), and 401s can arrive while the local token is still unexpired (server-side revocation, clock skew, introspection hiccup). That path runs `OAuthSessionCoordinator.refreshSession` → `KeycastOAuth.refreshSession` → `_saveSession`, overwriting the stored session with a refreshed (factless, per this PR's own premise) but *unexpired* token. `checkAccountDeletionReadiness` then sees `hasRpcAccess == true`, returns `ready`, the vanish + sweep publish, and `deleteKeycastAccount` uses the same factless token and gets the 403 — the exact #6335 failure mode the gate exists to prevent, just reached via a 401 instead of natural expiry. Fix: make the refresh origin observable and fail closed on it — e.g. persist a `grantType: refresh_token` marker on `KeycastSession` inside `refreshSession` (the login path stamps `password`/headless), and have the gate treat any session minted by the refresh grant as `requiresReauthentication`. That closes the hole without decoding the token, keeping the no-token-decoding design.

`mobile/lib/widgets/delete_account_dialog.dart:523`: 🟡 risk: TOCTOU between the one-shot gate and the final `deleteKeycastAccount`. After the gate passes, the flow burns the username (network call), runs the sweep (per-kind `publishEventAwaitOk`, 15s timeout each, up to 10k events), and publishes kind-62 — minutes can elapse. If the access token expires mid-flow, `getSessionOrRefresh` inside `deleteKeycastAccount` refreshes, drops the first-party fact, and the server 403s *after* the vanish was published. Window is small but the failure shape is identical. Fix: at the gate also require an expiry margin (refuse when `session.expiresAt - now` is under a conservative bound covering the sweep), and/or re-run `checkAccountDeletionReadiness()` immediately before the kind-62 publish and abort while nothing irreversible for the server step has been attempted. Real fix stays keycast#323.

`mobile/lib/widgets/delete_account_dialog.dart:624`: 🟡 risk: `keycastError` is logged but never surfaced — every registered-user server-deletion failure shows `deleteAccountServerDeletionFailed`, which now reads "Sign in again and retry." Post-gate, the likely failures are network/timeout/5xx, where that advice is wrong (and it replaces the old copy that blamed the network for 403s — the mislabel has swapped failure modes, not disappeared). It also means the 403-specific server message newly parsed in `oauth_client._errorMessageFrom` can never reach the user. Fix: thread `keycastError` into the snackbar, or split the copy by failure class (403 → re-auth copy; timeout/5xx → connection/retry copy).

`mobile/lib/services/account_deletion_service.dart:254`: 🟡 risk: `contentQueryFailed` only catches the *throw* case. `NostrClient.queryEvents` returns `[]` without throwing when no relay accepts the REQ (`nostr_client.dart:682-685` retries then completes empty; `:666` returns `[]` when disposed). In the zero-relay case the sweep is skipped as `isNotEmpty == false`, `queryFailed` stays false, the kind-62 still publishes (`publishEventAwaitOk` retries relays independently and can succeed), and the user gets the unqualified `deleteAccountSuccess` copy — the same silently-skipped-sweep bug this PR fixes, one branch over. Fix: treat "query completed with zero relay responses / zero connected relays" as `queryFailed`, not just exceptions.

hm21 (commented):
@realmeylisdev I took a quick look, and overall LGTM.

However one thing I noticed is that the new locales are still displayed in English instead of being translated into the respective language but I saw that your test plan above includes checking this under "Test G".

Could you quickly go through your full test plan to make sure every step works on your device as described? Once you've verified everything, please re-request a review from me, unless someone else has already approved it so we can ship it :)

NotThatKindOfDrLiz:
@realmeylisdev I pushed a takeover commit (`ad6323445f`) to keep this PR focused on the deletion correctness fix and address rabble’s review. I removed the Settings hub / Support Center entry points and their broader locale churn from this PR; that discoverability work should be a separate PR with its own widget coverage and copy/localization pass.

What changed in the focused fix:
- stamped Keycast sessions with grant type and blocked refresh-minted, legacy, missing-expiry, and near-expiry sessions before irreversible publishes
- added a second readiness check after username burn and before Nostr deletion publishes
- made sweep query failures cover no-relay/timeout/disposed-client paths instead of treating every empty result as healthy
- surfaced unconfirmed kind-5 sweeps with the unverified success copy
- kept server/network delete failures separate from re-auth-required failures

I also updated the PR body to match the narrower scope. Your follow-up commit (`81def9bd66`) added the locale translations/generated l10n on top, and CI is now green. Please review the takeover changes; I’m approving from my side.

NotThatKindOfDrLiz (approved):
Approved after takeover. The PR is now scoped to the account-deletion correctness fix, rabble’s blocking findings are addressed, and all current CI checks are green.

realmeylisdev:
@NotThatKindOfDrLiz — reviewed the takeover. The gate holds up: every refresh routes through `KeycastOAuth.refreshSession`, which is the only site stamping `refresh_token`, and `grant_type` survives both persistence paths and the account-switch archive in `signer_secure_store`. The 10-minute margin is comfortable against keycast's 24h default token TTL, and `contentDeletionIncomplete` is sound because `successCount` only advances on a confirmed batch.

Two things in the copy path were wrong, though, and I've pushed fixes in `90ecd89ad` and `4dd1c8d13`.

`_keycastErrorRequiresSignIn` matched English phrases in the error string, but keycast answers the first-party 403 with its own prose — "Account deletion requires the Divine app or web login with your private key" — and `oauth_client` returns the server's message whenever the body parses, so the local fallback the matcher was written against never ran. None of the three phrases matched, and the exact failure this PR exists to fix surfaced as "check your connection and try again". `DeleteAccountResult` now carries `requiresReauthentication`, set on 401 and 403, and the UI branches on that.

The other half is worse: `deleteAccountReauthRequired` ends with "Nothing has been deleted yet". That's true at both pre-flight gates, but this branch runs inside `if (result.success)` — the vanish and the sweep have already been confirmed by a relay, and NIP-62 can't be walked back. It was reachable via `deleteKeycastAccount`'s own "Session expired and could not be refreshed". There's a new string for the post-publish case, translated into every locale.

Both are pinned. I verified the widget test goes red under the old behaviour by substituting `false` for the flag, which is what the matcher returned for keycast's real body.

I also folded the second commit in while I was there: `NostrClient.queryEvents` had become a 53-line copy of `queryEventsDetailed` rather than delegating to it the way `nostr_sdk`'s own `queryEvents` does, and the two copies had already diverged in which SDK method they called. `lastQueryTimedOut` went with it — nothing in production read it, and the plain `queryEvents` reset it to false on every call, so a concurrent query could clear another one's flag. `queryEventsDetailed` had no package test; it has one now for each signal the sweep depends on.

NotThatKindOfDrLiz:
@realmeylisdev Is this ready to merge? Is there anything you're waiting on from me?

realmeylisdev:
> @realmeylisdev Is this ready to merge? Is there anything you're waiting on from me?

I've left some few concerns which I'm currently working, I'll push my changes shortly.

PR #6419, opened 2026-07-26
chore(deps): upgrade pro_video_editor to 2.9.0

### LINKED ISSUE #6420: chore(deps): upgrade pro_video_editor to 2.9.0
Upgrade `pro_video_editor` from 2.6.0 to 2.9.0.

## Why

Divine's Dart export retry added in #6384 can only help when the native plugin surfaces codec-resource pressure as a typed encoder failure. `pro_video_editor` 2.9.0 adds that Android-side mapping:

- transient codec-resource failures now become `RenderEncoderException.transient`
- starved decoder init failures are included when the cause is transient
- software-encoder fallback attempts are skipped when they would only re-run hardware under a misleading label

This should let the existing settle-retry catch the preview-decoder starvation case instead of surfacing a generic `RENDER_ERROR` on attempt 1.

## Scope

- Bump `pro_video_editor` in `mobile/pubspec.yaml`
- Commit the matching lockfile SHA for 2.9.0
- Keep Divine's existing retry behavior for now
- Update adjacent render fallback documentation so it matches the 2.9.0 fallback chain

## Notes

The bump also includes 2.7.0, 2.7.1, and 2.8.0. The main native risk is Darwin export behavior: 2.7.0 adds all-P-frame / 1s-GOP settings to bitrate-capped render and stop-motion encoders. Those are expected to help short-loop playback, but should be checked manually on iOS because CI does not exercise native iOS/macOS exports.

hm21:
Closes #6420

## Description

Upgrades `pro_video_editor` from 2.6.0 to 2.9.0.

2.9.0 carries the Android-side half of the export-crash work that #6384 could only address from Dart. #6384 added the Dart-level settle-retry because the plugin's own in-process fallback chain ran every attempt within a few hundred ms and could not outlast codec-reclaim latency. Three things changed natively since:

- The software-encoder fallback no longer silently re-runs the hardware encoder under a `software-encoder` label. When the device has no usable software encoder the attempt is skipped with an explicit log line, so the fallback chain stops reporting attempts it never made.
- `RenderEncoderException` now carries `isTransient`, separating exhausted codec resources from a genuine format/encoder incompatibility.
- `ERROR_CODE_DECODER_INIT_FAILED` now maps into the codec-resource path when the cause is transient. That lets Divine's existing Dart settle-retry catch the starved preview-decoder case #6384 was written for; on 2.6.0 it surfaced as a generic `RENDER_ERROR` and skipped the retry.

The bump also picks up the intermediate releases: lossless split transmux on Android and passthrough split on Darwin (2.7.0), the fully-consumed clip-transition fix (2.7.1), and `mergeAudioToFile` (2.8.0).

The 2.7.0 Darwin releases also add all-P-frame / 1s-GOP settings to the bitrate-capped render and stop-motion encoders. That is expected to help Divine's short loop playback, but it should be eyeballed on iOS because PR CI does not build native iOS/macOS exports.

## Related Issue

- #6420

## Out of Scope

- Consuming `isTransient` in `VideoEditorRenderService.renderWithEncoderFallback`. The current retry chain treats every `RenderEncoderException` as retryable, which stays correct with 2.9.0: a permanent failure costs two extra attempts before surfacing. A naive `if (!e.isTransient) rethrow` is not the right follow-up because attempt 3 is not just a retry; it re-renders at `VideoEditorConstants.encoderFallbackQuality` for devices at their 1080p encoder ceiling (#6058). Short-circuiting permanent failures before that rung would make those devices lose edits instead of exporting at 720p.

## Verification

- `flutter pub get` — resolves to `pro_video_editor 2.9.0`, no other lock churn
- `flutter test test/services/video_editor/` — 233/233 pass
- Pre-push hook after takeover commit:
  - merge-conflict check against `main` — clean
  - changed-file analyzer — no issues found
  - `flutter test test/services/video_editor/video_editor_render_service_test.dart` — 20/20 pass

## Device Verification Before Merge

- [ ] iOS concat export
- [ ] iOS stop-motion export
- [ ] Max-duration overlap transition on a device or simulator

## Type of Change

- [x] Chore

### DIFF
--- a/mobile/pubspec.lock
+++ b/mobile/pubspec.lock
@@ -1807,10 +1807,10 @@ packages:
     dependency: "direct main"
     description:
       name: pro_video_editor
-      sha256: aa05eb426dc98607dd6f776b4da9a5bb09c80c44e72138d299409d52a38b920d
+      sha256: ec977aa7d00678647c60d34b0dd2e91284e0305cd0688c1e305b930467d78dc7
       url: "https://pub.dev"
     source: hosted
-    version: "2.6.0"
+    version: "2.9.0"
   process:
     dependency: transitive
     description:
--- a/mobile/pubspec.yaml
+++ b/mobile/pubspec.yaml
@@ -322,7 +322,7 @@ dependencies:
   firebase_performance: ^0.11.1+4
   firebase_messaging: ^16.1.1
   audio_session: ^0.2.2
-  pro_video_editor: ^2.6.0
+  pro_video_editor: ^2.9.0
   pro_image_editor: ^13.2.3
   flutter_colorpicker: ^1.1.0
   file_picker: ^10.3.7
--- a/mobile/lib/services/video_editor/video_editor_render_service.dart
+++ b/mobile/lib/services/video_editor/video_editor_render_service.dart
@@ -1140,8 +1140,9 @@ class VideoEditorRenderService {
   /// encoder-init failures (`RenderEncoderException`).
   ///
   /// The exception signals that `pro_video_editor`'s own in-process fallback
-  /// chain (operating-rate cap/removal, software encoder, profile downgrade)
-  /// could not initialise an encoder. On codec-limited devices that is usually
+  /// chain (operating-rate cap/removal, profile downgrade, and a conditional
+  /// software encoder when the device has one usable for the target MIME) could
+  /// not initialise an encoder. On codec-limited devices that is usually
   /// transient contention rather than a true format incompatibility: the
   /// preview decoder the editor just released (the #5522 decoder-release gate)
   /// — or another app codec — has not been reclaimed by the OS media server


mobile/pubspec.yaml
```diff
@@ -322,7 +322,7 @@ dependencies:
   firebase_performance: ^0.11.1+4
   firebase_messaging: ^16.1.1
   audio_session: ^0.2.2
-  pro_video_editor: ^2.6.0
+  pro_video_editor: ^2.9.0
```
realmeylisdev, line 325: **The actual win here is bigger than the description says — and the iOS rationale is inverted.**

2.9.0 adds `ERROR_CODE_DECODER_INIT_FAILED` to `CODEC_INIT_ERROR_CODES` (`RenderVideo.kt:76`; that set was a 2-arm `when` in 2.6.0). Traced end-to-end:

```
starved decoder → ERROR_CODE_DECODER_INIT_FAILED
  → isTransientResourceFailure → CodecResourceExhaustedException
  → "CODEC_RESOURCE_EXHAUSTED" (new code)
  → RenderEncoderException.transient
  → caught at video_editor_render_service.dart:1195 → 800ms settle + retry
```

On 2.6.0 that code fell through `mapExportException` unchanged and surfaced as a plain `RENDER_ERROR` `PlatformException` — which `on RenderEncoderException` never catches. So a starved decoder killed the export on attempt 1 and the edit was lost.

That's precisely the #5522 scenario the doc comment at `:1145` describes ("the preview decoder the editor just released … has not been reclaimed by the OS media server yet"). #6384's settle-retry was written for it but could never actually fire for it. **2.9.0 is what makes that retry reach the case it was built for.** Worth naming in the body — it's a stronger argument for merging than the log-noise cleanup.

Separately, on the Out-of-Scope reasoning: `isTransient` is only ever `true` on Android is right, but the conclusion that a naive `if (!e.isTransient) rethrow` "would strip the retry from iOS/macOS entirely" doesn't hold — `ENCODER_NOT_SUPPORTED` has zero occurrences anywhere in the darwin tree, so `RenderEncoderException` is unreachable on iOS/macOS and the whole ladder is already Android-only. Nothing to strip.

The decision to defer is still correct, just for a different reason: attempt 3 isn't a retry, it re-encodes at `encoderFallbackQuality` for devices at their 1080p ceiling (#6058). Short-circuiting on `!isTransient` would delete that rung and make codec-ceiling devices lose the edit instead of exporting at 720p. Worth fixing the stated reason so the follow-up gets built the right way round.

mobile/pubspec.lock
```diff
@@ -1807,10 +1807,10 @@ packages:
     dependency: "direct main"
     description:
       name: pro_video_editor
-      sha256: aa05eb426dc98607dd6f776b4da9a5bb09c80c44e72138d299409d52a38b920d
+      sha256: ec977aa7d00678647c60d34b0dd2e91284e0305cd0688c1e305b930467d78dc7
       url: "https://pub.dev"
     source: hosted
-    version: "2.6.0"
+    version: "2.9.0"
```
realmeylisdev, line 1813: **Undocumented iOS encode change rides along in 2.7.0 — not in any changelog entry.**

Two non-split Darwin files gained the same pair of settings:

```swift
AVVideoAllowFrameReorderingKey: false,        // B-frames off
AVVideoMaxKeyFrameIntervalDurationKey: 1.0,   // keyframe every 1s
```

`BitrateCappedExporter.swift:152-159` and `StopMotionGenerator.swift:82-89`. Introduced in 2.7.0 (zero hits in 2.6.0, present 2.7.0 onward); the 2.7.0 changelog covers only the split passthrough and the watchdog.

This is on a path we always take — `video_editor_render_service.dart:1111` passes an explicit `bitrate` on every concat render, which is exactly what routes iOS through `BitrateCappedExporter`.

The direction looks right for us: the in-code comment describes short looped clips where the reorder buffer "intermittently starves `AVPlayerItemVideoOutput` — the player delivers no frame for a whole loop while audio keeps playing." That is Divine's exact product shape, so this plausibly fixes a real playback bug for us. The tradeoff is that at a fixed bitrate cap, no B-frames plus a 1s GOP costs some quality-per-bit.

Two asks, both cheap: worth a changelog line upstream, and since CI builds neither platform, an iOS export eyeball before merge would cover the only change here that touches every single iOS export.

mobile/pubspec.lock
```diff
@@ -1807,10 +1807,10 @@ packages:
     dependency: "direct main"
     description:
       name: pro_video_editor
-      sha256: aa05eb426dc98607dd6f776b4da9a5bb09c80c44e72138d299409d52a38b920d
+      sha256: ec977aa7d00678647c60d34b0dd2e91284e0305cd0688c1e305b930467d78dc7
```
realmeylisdev, line 1810: Coordination note: this supersedes the `^2.7.1` bump carried in #6266, which is blocked on exactly this ("must not merge until … `pro_video_editor` is bumped to `^2.7.1` here"). Landing this unblocks it; #6266 will want to drop its own pubspec/lock hunk on the rebase since it's already CONFLICTING.

On the 2.7.1 transition fix itself — it's inert until #6266 lands, which is the safe ordering. The relaxed guard (`<= 0` → `< 0` in `ClipTransitionGeometry.kt`) only admits the `body == 0` case, and today's picker caps overlaps at half the available room (`video_editor_transition_sheet.dart:165-170`, `transitionConsumedPerSide` = `duration * 2`), so a fully-consumed boundary can't be produced from the UI. Nothing changes on main; the new behaviour switches on with #6266's `consumed = blend` model.

Minor while you're here: the `renderWithEncoderFallback` doc at `video_editor_render_service.dart:1143` still describes the native chain as unconditionally including a software encoder. Under 2.9.0 that rung is skipped where no usable software encoder exists.

realmeylisdev (approved):
Verified the bump against the real published archives (2.6.0 → 2.9.0), not just the changelog: lockfile sha256 matches the published tarball, the package's own pubspec is unchanged apart from the version line, and every native build config is byte-identical. Dart changes are additive and I confirmed the 233/233 locally.

Three notes below — all disclosure, none blocking. The one worth acting on is that the PR undersells itself: this bump fixes a real Android failure class, and it also carries an undocumented iOS encode change.

realmeylisdev:
Two things worth naming as open rather than settled, neither blocking the approval:

- How often transient decoder starvation actually fires on Android — that's Crashlytics territory, not something the source can answer, so the size of the win is unmeasured.
- An iOS export eyeball for the 2.7.0 B-frame/keyframe change. It's the one change here that touches every iOS export, and PR CI builds neither platform, so no amount of static checking closes it.

NotThatKindOfDrLiz:
Takeover update for @hm21:

- Pushed `8072d0cf6e` to clarify `renderWithEncoderFallback` docs now that the software encoder rung is conditional in `pro_video_editor` 2.9.0.
- Updated the PR body and #6420 to name the decoder-starvation retry win and correct the `isTransient` out-of-scope rationale.
- Resolved the review threads for the PR body rationale and stale doc comment. Left the iOS encode-change thread open because that still needs device verification.

Verification:

- `flutter pub get`
- `dart format lib/services/video_editor/video_editor_render_service.dart`
- `flutter test test/services/video_editor/` (233/233)
- Pre-push hook: merge-conflict check, changed-file analyzer, and `flutter test test/services/video_editor/video_editor_render_service_test.dart`
- Refreshed GitHub checks on `8072d0cf6e`: all green.

Remaining before merge: please review the takeover commit and confirm the manual device checks from the PR body when you are ready to merge: iOS concat export, iOS stop-motion export, and one max-duration overlap transition on a device or simulator.

PR #6422, opened 2026-07-26
fix(sound-publishing): keep sound attribution on published posts

### LINKED ISSUE #6185: fix(sound-publishing): saved sound not displayed on published post
### Summary

Saved sounds from previous posts are not displayed at the bottom of newly published posts.

### Environment

any

### Steps to Reproduce

1. Save a sound from one of your previous posts.
2. Create a new post using the saved sound.
3. Ensure "Publish this sound" is enabled.
4. Publish the post and view it.

### Actual Result

The sound does not appear at the bottom of the published post, even after disabling and re-enabling "Publish this sound".

### Expected Result

The selected saved sound should appear at the bottom of the published post.

### Regression?

Unknown

### Evidence

reported on discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1527692694493855877)

realmeylisdev:
## Description

Two silent drops left a reused sound invisible at the bottom of a published post.

**Publish — `VideoEventPublisher`.** The "publish this sound" branch was gated on whether a sound was *selected* (`hasSelectedAudioEventId`), not on whether that selection actually produced a Nostr reference. A bundled sound yields no referenceable event, so the published video came out with **no `["e", …, "audio"]` tag, no `allow_audio_reuse` marker, and no Kind 1063** — the toggle was a no-op, which matches the reporter's "even after disabling and re-enabling 'Publish this sound'". Bundled selections now publish the rendered video audio as the user's reusable Kind 1063. External-provider catalog sounds are explicitly excluded from that branch so Freesound/Openverse-style selections are not republished as the user's reusable audio without their provider/license metadata.

**Feed — `AudioAttributionRow`.** The row collapsed to `SizedBox.shrink()` whenever the referenced audio event could not be fetched (unreachable relay, or a source video whose event id was replaced by an edit), so a *correctly tagged* video still showed nothing. It now degrades to a display-only `♪ Sound unavailable · Creator` pill, crediting the reused sound's original creator when the video carries one (via inspired-by), otherwise the video's author. The neutral label avoids claiming the referenced shared sound was the video author's original sound. It stays non-tappable: reuse consent can't be confirmed without the source event, so it must not become a reuse entry point.

The tappable and display-only forms now share one `_AudioAttributionPill`; the chevron and button semantics appear only on the tappable form. The unresolved-credit fallback is shared with `MetadataSoundsSection` so the two visible sound surfaces cannot drift.

**Scoping the shared credit.** That shared helper takes `reusedCreatorPubkey` as a required parameter rather than deriving it from the video, because a null has to stay "this video has no reused sound". `MetadataSoundsSection` depends on that: a video with no audio reference has its own original sound, and inspired-by is a general credit that says nothing about who made it. Deriving it instead credited the inspired-by creator on any video carrying one for an unrelated reason, wherever the author's profile had not loaded yet. Even with an audio reference present the credit is best-effort — `getActiveDraft` only auto-populates inspired-by from the sound's source video when the draft has none set — which is recorded on the helper.

**Related Issue:** Closes #6185

### Note on the original report

The report is dated 2026-07-19 against 1.0.16 (2026-07-11). #6066 (merged 2026-07-13, first shipped in 1.0.17 on 2026-07-22) added `AudioEvent.attributionEventId` and the editor-track attribution fallback, which fix the two drops that were live on 1.0.16: a reused original sound's `video_<eventId>` id failed the 64-hex check and the tag was dropped, and a sound added through the editor's music library never reached `draft.selectedSound` at all. On 1.0.17+ that repro should already produce a sound row — and step 3 of it is no longer reachable, because the "Publish this sound" toggle is now hidden for reused sounds. This PR closes the paths that remain on `main`.

## Out of Scope

- Bundled sounds still carry no per-sound feed attribution: their ids aren't event ids, so nothing referenceable can go in an `e` tag. With this PR a bundled-sound video that opts into reuse publishes its own Kind 1063 and gets the generic row; naming the bundled sound itself would need a separate tag shape.
- External-provider catalog sounds are not republished as reusable user audio. Preserving provider/license/creator metadata for those sounds would need a separate protocol shape.
- "Publish this sound" is still inert when the video references another creator's sound; the metadata toggle is deliberately hidden in that case (#6066).

## Verification

- `flutter test test/l10n/arb_consistency_test.dart test/widgets/metadata_sounds_section_test.dart test/widgets/audio_attribution_row_test.dart test/services/video_event_publisher_test.dart` — 61 passing.
- The credit regression test was run against `origin/main`'s copy of `metadata_sounds_section.dart` (passes) and against the version that derived the reused creator (fails), so it pins the behavior rather than the refactor.
- Pre-commit and pre-push hooks reran format, analyzer, and changed-file tests before each push.

## Type of Change

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -641,6 +641,7 @@
     "videoFollowButtonFollowing": "በመከተል ላይ",
     "videoFollowButtonFollow": "ተከተል",
     "audioAttributionOriginalSound": "ኦሪጅናል ድምጽ",
+    "audioAttributionUnavailableSound": "Sound unavailable",
     "videoInspiredByAttribution": "በ@{creatorName} የተነሳሳ",
     "@videoInspiredByAttribution": {
         "placeholders": {
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -641,7 +641,7 @@
     "videoFollowButtonFollowing": "በመከተል ላይ",
     "videoFollowButtonFollow": "ተከተል",
     "audioAttributionOriginalSound": "ኦሪጅናል ድምጽ",
-    "audioAttributionUnavailableSound": "Sound unavailable",
+    "audioAttributionUnavailableSound": "ድምጽ አይገኝም",
     "videoInspiredByAttribution": "በ@{creatorName} የተነሳሳ",
     "@videoInspiredByAttribution": {
         "placeholders": {
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -559,6 +559,7 @@
     "videoFollowButtonFollowing": "متابع",
     "videoFollowButtonFollow": "متابعة",
     "audioAttributionOriginalSound": "صوت أصلي",
+    "audioAttributionUnavailableSound": "Sound unavailable",
     "videoInspiredByAttribution": "مستوحى من @{creatorName}",
     "@videoInspiredByAttribution": {
         "placeholders": {
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -559,7 +559,7 @@
     "videoFollowButtonFollowing": "متابع",
     "videoFollowButtonFollow": "متابعة",
     "audioAttributionOriginalSound": "صوت أصلي",
-    "audioAttributionUnavailableSound": "Sound unavailable",
+    "audioAttributionUnavailableSound": "الصوت غير متاح",
     "videoInspiredByAttribution": "مستوحى من @{creatorName}",
     "@videoInspiredByAttribution": {
         "placeholders": {
--- a/mobile/lib/l10n/app_bg.arb
+++ b/mobile/lib/l10n/app_bg.arb
@@ -588,7 +588,7 @@
     "videoFollowButtonFollowing": "Следване",
     "videoFollowButtonFollow": "Следвай",
     "audioAttributionOriginalSound": "Оригинален звук",
-    "audioAttributionUnavailableSound": "Sound unavailable",
+    "audioAttributionUnavailableSound": "Звукът е недостъпен",
     "videoInspiredByAttribution": "Вдъхновен от @{creatorName}",
     "@videoInspiredByAttribution": {
         "placeholders": {
--- a/mobile/lib/l10n/app_bg.arb
+++ b/mobile/lib/l10n/app_bg.arb
@@ -588,6 +588,7 @@
     "videoFollowButtonFollowing": "Следване",
     "videoFollowButtonFollow": "Следвай",
     "audioAttributionOriginalSound": "Оригинален звук",
+    "audioAttributionUnavailableSound": "Sound unavailable",
     "videoInspiredByAttribution": "Вдъхновен от @{creatorName}",
     "@videoInspiredByAttribution": {
         "placeholders": {
--- a/mobile/lib/l10n/app_de.arb
+++ b/mobile/lib/l10n/app_de.arb
@@ -559,7 +559,7 @@
     "videoFollowButtonFollowing": "Gefolgt",
     "videoFollowButtonFollow": "Folgen",
     "audioAttributionOriginalSound": "Originalton",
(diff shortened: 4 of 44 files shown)


rabble (commented):
Review findings:

mobile/lib/services/video_event_publisher.dart:L1388: 🟡 risk: gate now also passes external-provider sounds (`external_provider_*` ids → `attributionEventId == null`). If that track is muxed into the render, the Kind 1063 now republishes third-party catalog audio as the user's own reusable sound with `allow_audio_reuse`. Confirm provider licensing permits this, or exclude `isExternalProviderSound` from the branch.

mobile/lib/widgets/video_feed_item/audio_attribution_row.dart:L156: 🔵 nit: unresolved fallback always labels `Original sound` even when the unresolvable reference was a named shared Kind 1063 — a transient relay error mislabels a shared sound as the author's original. Mirrors MetadataSoundsSection's existing fallback, so consistent, but worth a follow-up.

mobile/lib/widgets/video_feed_item/audio_attribution_row.dart:L137: ❓ q: inspired-by is a general credit, not sound-specific — a video with an audio reference plus a manually-set inspired-by for a non-sound reason credits the inspired-by creator as the sound's creator. Same heuristic as MetadataSoundsSection; acceptable if deliberate.

NotThatKindOfDrLiz:
@realmeylisdev I took over the review fixes and pushed `72b31ffd71`.

What changed:
- excluded external-provider catalog sounds from the rendered-audio publish branch so Freesound/Openverse-style selections are not republished as the user's reusable Kind 1063 without provider/license metadata
- kept bundled-sound reuse explicit and tested that it publishes a Kind 1063 and video `e ... audio` tag
- changed unresolved feed attribution from `Original sound` to neutral `Sound unavailable`
- extracted the unresolved creator-credit helper shared by the feed row and metadata sheet
- added the missing provider-error row test and fixed ARB consistency for the new l10n key
- updated the PR body to `Closes #6185`

Verification:
- `flutter analyze lib test`
- `flutter test test/services/video_event_publisher_test.dart test/widgets/audio_attribution_row_test.dart`
- `flutter test test/widgets/metadata_sounds_section_test.dart`
- GitHub CI is green on the latest head

Please review the takeover changes and merge when you're ready.

NotThatKindOfDrLiz (approved):
Final pass is clean. The takeover commit addresses the external-provider publishing regression, avoids unresolved feed misattribution, removes the duplicated fallback credit logic, and adds the missing coverage. CI is green on the latest head.

realmeylisdev:
Thanks @NotThatKindOfDrLiz — external-provider exclusion and the neutral label both look right. Reviewing the takeover commit turned up one regression in the extracted helper, plus an l10n gap. Three commits pushed.

**`resolve()` made "no reused sound" inexpressible.** It resolved a null `reusedCreatorPubkey` by falling back to `reusedCreatorPubkeyFor(video)`, but `MetadataSoundsSection` passes null deliberately: for a video with **no audio reference** it wants the video's own author credited. Since inspired-by is a general credit (@rabble's third finding), any video carrying one for an unrelated reason had its "Original sound" row credit the inspired-by creator instead of its author — visible whenever the author's profile hadn't loaded yet, which is the common cold-start case.

Verified by running the new regression test against `origin/main`'s copy of the file (passes) and against `72b31ffd71` (fails). Fixed in `830111b81b` by making `reusedCreatorPubkey` required so callers state it, with the feed row — only reachable with an audio reference — passing the resolved value. That also retired `AudioAttributionCredit`'s two never-read fields.

**The new l10n key shipped untranslated.** `audioAttributionUnavailableSound` was English in all 17 non-English locales while every sibling `audioAttribution*` key is translated, so the fallback pill read "Sound unavailable" everywhere. `arb_consistency_test` only checks key presence, so it stayed green. Translated in `e097ccd7ce`, reusing each locale's own noun for sound and its existing unavailable phrasing, kept neutral per the key's description.

**On @rabble's third finding:** the remaining case — audio reference present, inspired-by set for a non-sound reason, sound unresolvable — still credits the inspired-by creator. `getActiveDraft` only auto-populates inspired-by from the sound's source video when the draft has none set, so a pre-existing one survives. It's the best signal available at that point; `bac24dae58` records that as an explicit best-effort caveat on the helper rather than the guarantee my earlier comment implied.

Analyzer clean, and the four affected suites pass locally.

PR #717, opened 2025-12-19
feat: update logic for follow/unfollow from current user

omartinma:
## Description

Update follow/unfollow functionality from current user, fixing some sync issues between different screens.

Main changes:
- New: Added FollowersBloc and FollowingBloc to manage followers/following list state.
- Refactored screens: FollowersScreen and FollowingScreen now use bloc instead of Riverpod providers.
- Centralize follow/unfollow operations in few classes.

**Related Issue:** Closes #456, closes #460, closes #315, closes #457, closes #317, closes #316

**Important note**: while different blocs have been introduced, this PR mostly fix bugs related to viewing your current profile. For viewing followers/following of other profile will be touched in separate PR as part of the work in #469 

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [x] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #6424, opened 2026-07-26
fix(badges): await profile badge relay acceptance

### LINKED ISSUE #5937: fix(badges): badge acceptance publishes kind 10008, which the relay blocks — accepts never leave the device
## Summary

Accepting a badge on mobile publishes the NIP-58 Profile Badges event as kind `10008`, but the funnelcake relay's `allowed_kinds` whitelist (per its migrations) admits only `8`, `30008`, `30009` (+`10011`) for badges — kind `10008` was never added. The publish is rejected with `blocked: kind 10008 not in allowed list`, and the rejection is invisible to the app, so the acceptance appears to work on the accepting device but never reaches the relay, other devices, or the web ecosystem.

Found while investigating #3865 (the perf fix there stays kind-agnostic and does not depend on this).

## Mechanism

1. `BadgeRepository._publishProfileBadges` signs kind `EventKind.profileBadges` = `10008` (`mobile/lib/services/badges/badge_repository.dart:344`, `mobile/packages/nostr_sdk/lib/event_kind.dart:93`). This matches the current NIP-58 revision (Profile Badges = replaceable kind 10008), and has been the publish kind since the dashboard shipped in #3825.
2. funnelcake gates event ingestion on a ClickHouse-backed kind whitelist (`crates/relay/src/relay.rs:1494-1519` → `crates/clickhouse/src/management.rs:888`). Migration `000073_add_badge_identity_kinds.up.sql` adds kinds 8, 10011, 30008, 30009; no migration adds 10008 (verified across all migrations).
3. `divine-badges` (official web app + scheduled Diviner issuer) uses `PROFILE_BADGES = 30008` with `d = "profile_badges"` (`assets/app/nostr/constants.js:3,7`); the repo has zero references to 10008. So all real acceptance data lives on kind 30008.
4. `NostrClient.publishEvent` returns `PublishSuccess` when the event was *sent* to at least one relay — it never reads the relay's `OK` (`mobile/packages/nostr_client/lib/src/nostr_client.dart:387-447`; `publishEventAwaitOk` exists but BadgeRepository does not use it). The rejection is therefore silent.
5. Replaceable events are cached into the local Drift store on send-success (`nostr_client.dart:439-442`), and `queryEvents` merges the cache (`:567-569`) — so the accepting device sees itself as accepted while everyone else does not.
6. The read path masks the failure further: `_latestProfileBadges` queries 10008 first and falls back to 30008 (`badge_repository.dart:279-295`), so web-accepted badges still display on mobile via the fallback. (Side effect relevant to #3865: the 10008 primary query is a guaranteed-empty network round-trip per pubkey in prod today.)

## Evidence

- Live prod probe (read-only REQs against `wss://relay.divine.video`, 2026-07-08): zero kind-10008 events exist, while fresh kind-30008 acceptances exist (e.g. `d=profile_badges` lists referencing `30009:e21369e63b98f58de8aa171ec9794006eb0118891ae70895106d44525b718d2b:diviner-of-the-day` awards).
- Local-stack repro (funnelcake relay + migrations, with `allowed_kinds` aligned to prod migration 000073 by inserting 8/30008/30009):
  - kind 30008 publish → `OK true`, stored, queryable
  - kind 8 / 30009 publish → `OK true`
  - kind 10008 publish → `OK false` — `blocked: kind 10008 not in allowed list`, not stored

Caveat: `allowed_kinds` is runtime-manageable via NIP-86 (`allow_kind` exists server-side), so prod could in principle differ from the migrations. A relay admin can confirm in seconds (`SELECT kind FROM nostr.allowed_kinds FINAL ORDER BY kind`). The zero-10008-events probe result is consistent with the migration state.

## Decision needed (product/protocol)

Which kind should Divine standardize on for Profile Badges?

- **Option A — align mobile to kind 30008 (`d=profile_badges`)**: matches the relay whitelist, divine-badges web, and all existing acceptance data. Mobile flips its publish kind and query order (30008 primary, 10008 as read-compat or dropped). Smallest blast radius; diverges from the newest NIP-58 revision.
- **Option B — move the ecosystem to kind 10008**: allow 10008 on the relay (migration), migrate divine-badges to read/write 10008, and migrate/backfill existing 30008 lists. Spec-modern, but cross-repo coordination plus data migration.

Either way, mobile's badge accept/remove should surface real relay acceptance (use the OK-aware publish path) instead of send-success, so rejections are not silently swallowed.

## Also noted

- The local Docker stack cannot exercise badges at all: `local_stack`'s ClickHouse seed contains no badge kinds in `allowed_kinds` (it also lacks them for kinds prod accepts today, like 8/30008/30009). Worth aligning so badge flows are locally testable.

cc @NotThatKindOfDrLiz for the protocol/product call between A and B.

NotThatKindOfDrLiz:
## Summary
- publish profile badge updates with publishEventAwaitOk so relay OK false/no response surfaces as an action failure instead of a local-only success
- resolve profile badge state by newest parsed event across kind 10008 and legacy 30008/d=profile_badges
- remove the unused legacy NIP58.ware writer and deprecated badgeAccept constant

## Verification
- flutter test test/src/badge_repository_test.dart test/src/nip58_badge_parser_test.dart (from mobile/packages/badge_repository)
- flutter analyze packages/badge_repository packages/nostr_sdk (from mobile)
- flutter analyze (from mobile)
- pre-push hook: analyzer + changed badge repository tests

## Notes
- Relay acceptance for kind 10008 is deployed: divine-funnelcake#715 and #717 merged on 2026-07-29, and production deployment 5667393568 for sha 55ed7c83 succeeded on 2026-07-30T01:07:03Z.
- Web profile rendering still needs divine-web#521 before or with the mobile release so divine.video reads profile badges that mobile already writes as kind 10008. This is a release follow-up for cross-client visibility, not a merge gate for this mobile PR.

Closes #5937


rabble (commented):
Review findings:

mobile/packages/badge_repository/lib/src/badge_repository.dart:418: ❓ q: PR Notes say funnelcake `allowed_kinds` still lacks 10008 in prod. Once this merges, every accept/remove throws and user sees `badgesUpdateError` until relay migration lands (previously silent local-only success). Confirm ship order — funnelcake migration first, or feature gated?

Otherwise clean: await bounded by 15s `PublishTracker` timeout, partial-acceptance semantics correct (`failed` only when zero relays ack), optimistic cache rollback handled in `nostr_client`, newest-wins 10008/30008 merge has regression tests both directions plus empty-event-authoritative case, removed `badgeAccept`/`NIP58.ware` have no dangling refs.

rabble (commented):
Verified: publishEventAwaitOk exists with 15s timeout and optimistic-cache rollback; PublishOutcome.failed/.summary exist; parser accepts both kind 10008 and legacy 30008/d=profile_badges; dual-kind newest-wins resolution is coherent with the read path, and refs are re-fetched fresh from both kinds before every write (no stale-cache wipe — good). One deployment concern: badge accept/remove now hard-fails in prod until funnelcake allows kind 10008 (PR body confirms the relay rejects it today). The old behavior silently lost badges, the new one surfaces an error — better — but merge/release must be sequenced with the relay allowed_kinds change or badge accept visibly breaks. Please coordinate deploy order before merging. Nit: newest-wins across kinds means a badge present solely in the older event of the other kind is dropped on write; consistent with what the user sees, but if other clients still write legacy 30008, consider union-merging on write.

mobile/packages/badge_repository/lib/src/badge_repository.dart
```diff
@@ -358,8 +356,7 @@ class BadgeRepository {
         ),
       ]),
     ]);
-    return _newestParsedProfileBadges(results[0]) ??
-        _newestParsedProfileBadges(results[1]);
+    return _newestParsedProfileBadges([...results[0], ...results[1]]);
```
rabble, line 386: Flattening both kinds into one list makes the `created_at` tie case nondeterministic — `_newestParsedProfileBadges` sorts and takes `.first`, and Dart's `List.sort` is not a stable sort. The previous `?? `-chained form always resolved a tie in favour of kind 10008.

A client following NIP-58's "migrate to the new format" guidance by writing 10008 and legacy 30008 together produces exactly this tie, so accepted-badge state becomes a coin flip for those users.

Adding a kind tiebreak in `_newestParsedProfileBadges` (prefer `EventKind.profileBadges`) restores determinism and matches the spec's migration direction.

rabble (changes requested):
Third pass, focused on what the earlier two reviews didn't cover. The `publishEventAwaitOk` swap is right and I'll leave that settled. One new correctness finding in the dual-kind merge, plus a spec citation that resolves an open nit from my earlier review.

## Should fix: the newest-wins merge is nondeterministic on a `created_at` tie

`_newestParsedProfileBadges` (`badge_repository.dart:440-452`) sorts by `createdAt` descending and takes `.first`. `List.sort` in Dart is explicitly **not** a stable sort, so when a kind 10008 event and a legacy 30008/`d=profile_badges` event carry the *same* `created_at`, which one wins is arbitrary — and can differ between runs on the same data.

Before this PR that case was deterministic: `_newestParsedProfileBadges(results[0]) ?? _newestParsedProfileBadges(results[1])` meant 10008 always won a tie. Flattening to one list gave that up.

This isn't hypothetical. NIP-58 tells clients to "migrate to the new format", and the natural way to do that is to write the new 10008 and the legacy 30008 together — which lands both at the same `created_at`. Any user who has migrated through such a client gets a coin-flip on whether their accepted badges show up.

Fix is a secondary comparator preferring the current kind:

```dart
final sorted = parsed.toList()
  ..sort((left, right) {
    final byTime = right.event.createdAt.compareTo(left.event.createdAt);
    if (byTime != 0) return byTime;
    // Same timestamp: the spec says migrate to 10008, so it wins.
    return (right.event.kind == EventKind.profileBadges ? 1 : 0)
        .compareTo(left.event.kind == EventKind.profileBadges ? 1 : 0);
  });
```

Worth a fourth test alongside the three you added — equal `createdAt`, award present only in the 10008 — asserting `isAccepted` is `true`. Note this package has no `min_coverage` override in `.github/workflows/badge_repository.yaml`, so it inherits the VeryGood 100% default; a new branch needs the test regardless.

## Resolving my earlier union-merge nit — newest-wins is correct, don't union

In my previous review I suggested: *"if other clients still write legacy 30008, consider union-merging on write."* Withdrawing that. NIP-58's closing section is explicit:

> ### Deprecated Profile Badges event
> An earlier version of this NIP used a kind `30008` addressable event with a `d` tag value of `profile_badges` for the Profile Badges event. Clients should treat these events as **equivalent** to kind `10008` and migrate to the new format.

"Equivalent" means one logical list observed through two encodings — so newest-wins is the correct resolution and this PR's change is more spec-aligned than the old kind-precedence behaviour, not less. Union-merging would be actively wrong: it would resurrect badges the user had deliberately removed in the newer event. Please don't implement that suggestion.

This also justifies the third new test ("treats newer empty profile badges as authoritative") — an empty newer event is a user who removed everything, and honouring it is right.

## Verified, no action needed

The legacy query filters `d: ['profile_badges']` (`badge_repository.dart:351-356`). That matters more than it looks: kind 30008 is *also* a live current kind in NIP-58 — "Badge Set", an addressable NIP-51 set for categorising badges. Without the `d` filter, a user's genuine Badge Set events would be parsed as their Profile Badges and could out-timestamp the real 10008. The filter is correct as written; flagging it so it doesn't get "simplified" away later.

## Still open: deploy sequencing

My earlier question hasn't been answered and it remains the gate on merging. The PR body says the relay rejects kind 10008 today, and the rejected-outcome test fixture literally encodes `'blocked: kind 10008 not in allowed list'`. So on merge, every accept/remove hard-fails with a user-visible error until the funnelcake `allowed_kinds` migration ships. That's the right behaviour — better than the old silent local-only success — but it has to land second. Either confirm the funnelcake change is deployed first, or gate the accept affordance until it is.

## Nits

- The comment removed from `_latestProfileBadges` described the old precedence rule, so removing it was correct — but nothing replaced it. One line stating newest-wins-across-kinds and citing the "treat as equivalent" spec text would stop this being re-litigated later.
- Roughly 40 of the diff's lines are `dart format` churn on unrelated `signEvent:` blocks in the test file. Harmless if it's what the pinned formatter emits, but it makes the substantive test changes harder to pick out.

NotThatKindOfDrLiz:
Follow-up pushed in eaeb792fea.

What changed:
- Added a deterministic tie-break for equal-created_at profile badge events: kind 10008 now wins over legacy 30008/d=profile_badges when timestamps tie.
- Added regression coverage for the same-second migration case where the award exists only in the 10008 event.
- Added a short NIP-58 equivalence comment on the dual-kind read path.
- Removed the remaining dead nostr_sdk NIP58 shell after ware was deleted; no repo references remained.

Verification:
- flutter test test/src/badge_repository_test.dart test/src/nip58_badge_parser_test.dart (from mobile/packages/badge_repository)
- flutter analyze packages/badge_repository packages/nostr_sdk (from mobile)
- pre-push hook: merge-conflict check, analyzer, changed badge_repository tests

Ship status:
- divine-funnelcake#715 is merged.
- divine-funnelcake#717 is merged.
- I still do not see a production deployment record containing the merged funnelcake commits in the GitHub deployments API, so production relay acceptance for 10008 should still be confirmed before this mobile PR is merged/released.

mobile/packages/badge_repository/lib/src/badge_repository.dart
NotThatKindOfDrLiz: Addressed in 54f9e0136. The comparator now resolves equal created_at by kind first (10008 over legacy 30008) and then applies the NIP-01 same-kind tie-break by lowest event id. Added regression coverage for the same-kind equal-timestamp case.

NotThatKindOfDrLiz:
Takeover follow-up pushed in 54f9e0136.

What changed:
- Added `BadgePublishException` for relay OK false / no-response publish failures so expected relay/domain failures no longer surface as `StateError`.
- Completed profile-badge event ordering: newest `created_at`, then kind 10008 over legacy 30008 on cross-kind ties, then NIP-01 lowest event id for same-kind ties.
- Added regression coverage for same-kind equal-timestamp profile badge events and updated the relay-rejection test to assert the typed exception.
- Filed divine-web#521 to track the required web read path for kind 10008 before/with mobile release.

Verification:
- `flutter test test/src/badge_repository_test.dart test/src/nip58_badge_parser_test.dart` from `mobile/packages/badge_repository`
- `flutter analyze packages/badge_repository packages/nostr_sdk` from `mobile`
- pre-push hook: merge-conflict check, analyzer, changed badge repository tests

Rabble inline thread replied to and resolved. CI is running on the new head; still need production relay acceptance for kind 10008 confirmed before release.

NotThatKindOfDrLiz:
CI update for 54f9e0136: all 16 checks are green. Re-review requested from rabble; PR remains blocked only by the existing CHANGES_REQUESTED review state and the release sequencing note for relay/web kind 10008 support.

NotThatKindOfDrLiz:
Takeover follow-up pushed in 8d2757682.

What changed:
- Tightened the equal-timestamp cross-kind regression test so it fails if the kind 10008-over-legacy tie-break is removed.
- Reused `Nip58ProfileBadges.isLegacyProfileBadges` in the comparator instead of re-deriving kind state.
- Added public API docs for `BadgePublishException` and clarified that its `PublishOutcome` is carried for relay-level logging/triage.
- Removed the stale `BADGE_ACCEPT` entry from the Nostr SDK API reference and documented `PROFILE_BADGES` / `BADGE_SET` instead.
- Updated the PR body: it now closes #5937, records the successful funnelcake production deployment for kind 10008 acceptance, and leaves divine-web#521 as the remaining release sequencing note.

Verification:
- `flutter test test/src/badge_repository_test.dart test/src/nip58_badge_parser_test.dart` from `mobile/packages/badge_repository`
- `flutter analyze packages/badge_repository packages/nostr_sdk` from `mobile`
- pre-push hook: merge-conflict check, analyzer, changed badge repository tests
- GitHub checks are green on 8d2757682.

@rabble re-review requested; the earlier comparator thread and relay deployment gate should both be addressed now.

NotThatKindOfDrLiz:
Takeover follow-up pushed in bb31cf58b.\n\nWhat changed:\n- Updated the root docs/API_REFERENCE.md copy to remove the deleted BADGE_ACCEPT entry and document PROFILE_BADGES / BADGE_SET consistently with mobile/packages/nostr_sdk/API_REFERENCE.md.\n- Expanded the BadgeRepository acceptAward/removeAward dartdoc Throws blocks to include the existing StateError paths for missing pubkey and signing failure, alongside BadgePublishException for relay non-acceptance.\n\nVerification:\n- flutter test test/src/badge_repository_test.dart test/src/nip58_badge_parser_test.dart (from mobile/packages/badge_repository)\n- flutter analyze packages/badge_repository packages/nostr_sdk (from mobile)\n- git diff --check\n- pre-push hook: merge-conflict check, analyzer, changed badge repository tests\n- GitHub checks are green on bb31cf58b.\n\n@rabble re-review requested; the remaining cleanup items from the last pass should now be addressed.

NotThatKindOfDrLiz:
Takeover update pushed in 2974012b7 after rebasing onto current origin/main.

What changed:
- moved PROFILE_BADGES back into numeric order in both API reference copies
- simplified the profile badge kind tie-break comparator so the int rank is inline instead of named like a boolean
- picked up the repo formatter's one-line cleanup in the touched badge repository file

Verification:
- flutter pub get (from mobile)
- flutter test test/src/badge_repository_test.dart test/src/nip58_badge_parser_test.dart (from mobile/packages/badge_repository)
- flutter analyze packages/badge_repository packages/nostr_sdk (from mobile)
- pre-push hook: merge-conflict check, analyzer, changed badge repository tests
- GitHub checks are green on 2974012b7

I did a final pass over the PR diff after the push. I do not see any remaining blocker in the mobile diff; the branch is now rebased onto current main. @rabble can you re-review when you have a moment so the stale CHANGES_REQUESTED clears?

NotThatKindOfDrLiz:
Takeover follow-up complete.

What changed:
- Updated the PR body release note so divine-web#521 is described as a cross-client visibility follow-up before/with mobile release, not as a merge gate for this mobile PR.
- Re-ran a final diff audit from a clean takeover worktree at head 2974012b7.

Verification:
- GitHub checks are green on 2974012b7149e507ae48b237e0e696c554285414.
- `git diff --check origin/main...HEAD` is clean.
- Deleted `badgeAccept` / `NIP58.ware` symbols have no live references; only a historical plan-doc mention remains.
- Relay deployment gate is satisfied: divine-funnelcake commits 2a07a1ff and 7dcb41c6 are ancestors of production deploy b1de51a7f22398e5838c3cd0ee3abc3062746bbf.

@rabble the earlier CHANGES_REQUESTED review is stale after the follow-up commits and deployment confirmation. Can you re-review/clear it when you have a moment?

rabble (approved):
Re-reviewed the current head. The equal-created_at merge is deterministic (current kind before legacy, then NIP-01 event-id ordering), newest-wins semantics remain correct for removals, relay rejection is typed and awaited, and the production kind-10008 acceptance deployment gate has been confirmed. Focused package tests and all current-head checks are green. Ready to merge; divine-web#521 remains a cross-client release-sequencing follow-up, not a blocker in this mobile diff.

PR #5551, opened 2026-06-25
fix(feed): preserve native video loop boundaries

NotThatKindOfDrLiz:
## Summary
- remove 6.3s seek-to-zero loop enforcement from feed playback so native/web looping reaches the asset boundary
- remove the legacy per-controller loop timer and its stale tests
- add a feed looping contract test to guard against reintroducing capped seek-based looping

Fixes #5544

## Verification
- flutter analyze
- flutter test test/widgets/video_feed_item/feed_looping_contract_test.dart test/providers/safe_controller_operations_test.dart
- flutter test test/src/services/controller_subscriptions_test.dart (from mobile/packages/infinite_video_feed)
- pre-push hook: analyzer, generated-file verification, changed-file tests

PR #6429, opened 2026-07-27
fix(feed): cap playback length with native clipping

### LINKED ISSUE #6421: task(feed): reinstate playback duration cap via native clipping, above classic Vine length
## Description

Restore the feed-side guarantee that no video plays longer than a Vine, but implement it as **native clipping** (`VideoClip.end` → `endMs`) instead of the Dart-side `seekTo(Duration.zero)` that was removed in #5551. The cap must sit **above** the length of classic Vine assets, not at the 6.3s recording limit.

## Context

The old enforcement (`maxLoopDuration: VideoEditorConstants.maxDuration`, introduced in #2119 / #2384) served two purposes that got conflated:

1. **Product rule** — nothing in the feed plays longer than a Vine, including kind-34236 events published by other Nostr clients that point at arbitrarily long files.
2. **Loop mechanism** — it was implemented as a Dart-side `seekTo(Duration.zero)` once playback crossed 6.3s.

Purpose 2 was the direct cause of the audible loop seam reported in #5544 ("Loops are not seamless", app 1.0.15+778, iOS), so #5551 (`d2a17602d`) removed the whole thing. That fixed the seam, but purpose 1 went with it and is not mentioned in the PR.

**There is no duration guard left anywhere in the app today.** No filter on ingest, no cap in the player. A 60s file referenced by a foreign client's kind-34236 event plays in full in the feed.

The cap can come back without the seam, because the loop point then lives in the native player rather than in a Dart roundtrip. Both backends already support clipping and `VideoClip.start`/`end` is already plumbed through the method channel as `startMs`/`endMs`:

- Android — `MediaItem.ClippingConfiguration.setEndPositionMs` in `DivineVideoPlayerInstance.kt:435`; `REPEAT_MODE_ALL` then loops the clipped item.
- iOS — `insertTimeRange` in `DivineVideoPlayerInstance.swift:439` trims the `AVMutableComposition`; `AVPlayerLooper` loops the trimmed composition.

### The cap value matters

Classic Vine assets are **not** 6.3s. Measured with `ffprobe` across 9 classic assets on `media.divine.video`:

| Asset | Video duration |
|---|---|
| `73d54eb3c02a…` | 6.533s |
| `8f3e58c3f681…` | 6.500s |
| `ea5f725563dc…` | 6.500s |
| `859699804d01…`, `b775c2903b60…`, `45864a41a57a…` | 6.533s |

Re-clipping at 6.3s would cut the last ~200ms off every classic Vine and destroy the musical loop point on exactly the beatbox/music-loop content that #5544 and #6386 complain about. The rule we actually want is "no 30/60s videos from foreign clients", not "everything is exactly 6.3s".

Proposal: cap at **7s**, defined as a named playback constant (not `VideoEditorConstants.maxDuration`, which is the *recording* limit and must stay 6.3s). Alternative worth considering during implementation: apply the cap only to non-Divine sources.

## Implementation Notes

- Feed clips are currently built without `end`: `mobile/packages/infinite_video_feed/lib/src/utils/source_loader.dart:60` and `mobile/packages/infinite_video_feed/lib/src/widgets/infinite_video_feed.dart:1149`, `:1364`, `:1458`.
- Thread the cap through `InfiniteVideoFeed` as a new parameter and apply it as `VideoClip(end: …)`. Do **not** revive the `maxLoopDuration` seek path in `controller_subscriptions.dart` — that is the seam.
- `mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart` asserts that `feed_videos.dart` contains no `maxLoopDuration:`. A clipping-based cap keeps that test green; extend it to also pin the new behaviour.
- Verify ExoPlayer end-clipping does not introduce a rebuffer at the loop point on Android, and that `AVPlayerLooper` still loops gaplessly over a trimmed composition on iOS.

## Acceptance Criteria

- [ ] A video longer than the cap stops at the cap and loops from there, in the feed, on iOS and Android.
- [ ] Classic Vines (6.500–6.533s) play to their natural end — no truncation.
- [ ] No Dart-side `seekTo` is involved in the loop point; looping stays native.
- [ ] The cap is a named playback constant, separate from `VideoEditorConstants.maxDuration`.
- [ ] Contract test extended to pin both the clipping cap and the absence of seek-based looping.

## Testing

- Widget/contract test for the clip construction (cap present, value correct).
- Manual: play a >7s video from an external Nostr client and confirm it caps and loops.
- Manual: play a classic beatbox Vine and confirm the loop point is unchanged versus today's build.

Related to #5544, #5551, #6386.

Closes #none

hm21:
## Description

Restores the feed-side guarantee that nothing plays longer than a Vine, implemented as **native clipping** rather than the Dart-side `seekTo(Duration.zero)` that #5551 removed.

**Why the rule went missing.** The old `maxLoopDuration: VideoEditorConstants.maxDuration` did two jobs at once: it enforced the product rule (no 30/60s videos in the feed, including kind-34236 events from foreign clients pointing at arbitrarily long files) and it implemented that rule as a Dart seek. The seek was the direct cause of the audible loop seam in #5544, so #5551 removed the whole setting — and the product rule left with it. There is no duration guard anywhere in the app today; a 60s file referenced by another Nostr client plays in full.

**Why clipping works where the seek didn't.** The loop point now lives in the native player rather than in a Dart roundtrip, so there is no seam to produce. Android sets `MediaItem.ClippingConfiguration.endPositionMs` and `REPEAT_MODE_ALL` repeats the clipped item; iOS trims the `AVMutableComposition` via `insertTimeRange` and `AVPlayerLooper` loops the trimmed composition. Both were already plumbed through the method channel as `startMs`/`endMs` — this PR only starts using them from the feed.

**Why 7s and not 6.3s.** `VideoEditorConstants.maxDuration` is the *recording* limit and stays 6.3s. Classic Vine assets are not 6.3s — measured with `ffprobe` on `media.divine.video` they run 6.500–6.533s. Re-clipping at the recording limit would cut ~200ms off every classic Vine and destroy the musical loop point on exactly the beatbox content #5544 and #6386 complain about. The new `AppConstants.maxFeedPlaybackDuration` is a separate, named playback constant set to 7s.

**Why the cap applies to every source, not just non-Divine ones.** The issue floated capping only foreign sources. Divine's own export is already bounded at 6.3s (`renderVideo(maxOutputDuration:)`) and classic Vines are ≤6.533s, so a 7s cap is a no-op for both. A source-conditional branch would buy nothing and leave a hole for anything that slips past the source check.

**Why all four open paths.** First load, cache hit, source failover and the HTTP-202 processing retry each re-open the player independently, so a missed site would let a long video escape the cap on that path alone. The contract test counts `VideoClip` constructions against cap applications, so a future path cannot skip it silently.

**Why the Swift change.** ExoPlayer clamps `endPositionMs` to the real duration itself (`ClippingTimeline`: `if (endUs > window.durationUs) endUs = window.durationUs`). AVFoundation clamps what `insertTimeRange` actually inserts, but the Swift side computed `clipDuration` from the *requested* end regardless — so every 6.533s Vine would have reported a 7.0s duration while looping at 6.533s, desyncing every duration readout. `endTime` is now clamped to the asset duration, matching ExoPlayer.

`maxLoopDuration` and its seek path are untouched and still unused by the feed; the existing contract assertion that `feed_videos.dart` contains no `maxLoopDuration:` stays green.

**Related Issue:** Closes #6421

## Out of Scope

- Removing the now-unused `maxLoopDuration` parameter and `subscribeToLoopEnforcement` from `infinite_video_feed`. Pre-existing dead-in-production code, unrelated to this fix, and deleting it would touch 16 test references in the package.
- The Linux (`media_kit`) backend does not clamp a clip end past the source duration, so it reports the requested length. Linux is a dev-only desktop target that CI does not build, and clamping there would cost a duration probe per clip on the editor path. Documented on `VideoClip.end` as an Android/Apple guarantee instead.

## Verification

- `flutter analyze lib test integration_test` — clean (app), plus both touched packages.
- `dart format` — clean.
- `infinite_video_feed`: full suite, 231 tests, **100.00% line coverage** (the package's CI gate).
- `divine_video_player`: full suite, 257 tests. Coverage 99.49%, byte-identical to the same run on a clean tree — the four uncovered lines are pre-existing const-constructor declarations.
- App: `test/widgets/video_feed_item/` + `test/screens/feed/`, 460 tests.
- New tests: cap threading through the fallback chain and the HTTP-202 retry (`source_loader_test.dart`); cap wired in `feed_videos.dart`, applied at every clip site, and clearing classic Vine length (`feed_looping_contract_test.dart`).
- Negative check: removed one `end:` application on purpose and confirmed the new contract test goes red.
- Confirmed ExoPlayer's clamp by disassembling `ClippingMediaSource$ClippingTimeline` from the media3 1.10.0 AAR rather than trusting the docs.

### Manual test plan (pending — draft until these pass)

- [ ] Real Samsung Galaxy device: a >7s video from an external Nostr client stops at 7s and loops from there, with no rebuffer at the loop point.
- [ ] iOS: same video caps and loops, `AVPlayerLooper` still gapless over the trimmed composition.
- [ ] Both platforms: a classic beatbox Vine plays to its natural end, loop point unchanged versus the current build, no truncation.

## Type of Change

### DIFF
--- a/mobile/lib/constants/app_constants.dart
+++ b/mobile/lib/constants/app_constants.dart
@@ -55,6 +55,23 @@ class AppConstants {
   static const int followingVideoThreshold = 5;
 
   // ============================================================================
+  // FEED PLAYBACK
+  // ============================================================================
+
+  /// Hard cap on how long any feed video plays before it loops.
+  ///
+  /// Nothing in the feed plays longer than a Vine, including kind-34236 events
+  /// published by other Nostr clients that point at arbitrarily long files.
+  /// The cap is applied as a native clip end (`VideoClip.end`), so the loop
+  /// point stays inside the platform player — a Dart-side seek-to-zero is what
+  /// produced the audible loop seam in #5544.
+  ///
+  /// Deliberately above `VideoEditorConstants.maxDuration` (6.3s, the
+  /// *recording* limit): classic Vine assets measure 6.500–6.533s, and capping
+  /// at the recording limit would clip their musical loop point (#6421).
+  static const Duration maxFeedPlaybackDuration = Duration(seconds: 7);
+
+  // ============================================================================
   // VIDEO PROCESSING
   // ============================================================================
 
--- a/mobile/lib/widgets/video_feed_item/feed_videos.dart
+++ b/mobile/lib/widgets/video_feed_item/feed_videos.dart
@@ -15,6 +15,7 @@ import 'package:openvine/blocs/video_interactions/video_interactions_bloc.dart';
 import 'package:openvine/blocs/video_playback_status/video_playback_status_cubit.dart';
 import 'package:openvine/blocs/video_playback_status/video_playback_status_state.dart';
 import 'package:openvine/blocs/video_volume/video_volume_cubit.dart';
+import 'package:openvine/constants/app_constants.dart';
 import 'package:openvine/extensions/video_event_extensions.dart';
 import 'package:openvine/features/feature_flags/models/feature_flag.dart';
 import 'package:openvine/features/feature_flags/providers/feature_flag_providers.dart';
@@ -364,9 +365,12 @@ class FeedVideosState extends ConsumerState<FeedVideos> with RouteAware {
           _resumeAutoAdvanceAfterSwipe();
           widget.onActiveVideoChanged?.call(video, index);
         },
-        // Do not pass maxLoopDuration here. Feed playback should loop at the
-        // asset boundary; restarting with a Dart seek at the 6.3s recording
-        // limit creates an audible seam.
+        // Nothing in the feed plays longer than a Vine, not even a 60s file a
+        // foreign Nostr client points at. The cap is a native clip end, so the
+        // loop point stays in the platform player — deliberately NOT
+        // maxLoopDuration, whose Dart seek at the 6.3s recording limit both
+        // truncated classic Vines and created an audible seam (#5544, #6421).
+        maxPlaybackDuration: AppConstants.maxFeedPlaybackDuration,
         onVideoLoopCompleted: _handleAutoAdvanceCompleted,
         shouldPortraitExpand: widget.shouldPortraitExpand,
         canAutoPlay: _canAutoPlayVideo,
--- a/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
+++ b/mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
@@ -391,7 +391,18 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
             let startTime = CMTime(value: startMs, timescale: 1000)
             let endTime: CMTime
             if let endMs {
-                endTime = CMTime(value: endMs.int64Value, timescale: 1000)
+                // Clamp to the asset: insertTimeRange silently inserts only the
+                // media that exists, so an endMs past the source would leave
+                // clipDuration (and therefore the reported totalDuration)
+                // longer than what actually plays. The feed relies on this —
+                // it caps every clip at maxFeedPlaybackDuration without knowing
+                // the source length up front. ExoPlayer clamps the equivalent
+                // ClippingConfiguration itself.
+                let requestedEnd = CMTime(value: endMs.int64Value, timescale: 1000)
+                endTime =
+                    (assetDuration.isNumeric && CMTimeCompare(requestedEnd, assetDuration) > 0)
+                    ? assetDuration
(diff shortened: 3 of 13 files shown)


rabble (approved):
LGTM. 7s native clip-end cap threaded through all four feed open paths; Swift clamps end to asset duration; web duration resolver clamps; parity contract test pins every clip site. Cross-PR note: textually conflicts with #6430 in source_loader.dart, infinite_video_feed.dart, and DivineVideoPlayerInstance.swift — semantically compatible (both only shorten the end), but whichever merges second will need a rebase at those three sites. Nit (non-blocking): feed_looping_contract_test.dart shells out to awk with CWD-relative paths — fine in CI (runs from mobile/) but non-hermetic; consider resolving paths from the script location.

Issue #5202, opened 2026-06-15
fix(dm): restore missing chats after reinstall


hm21:
### Summary

Some direct message chats are missing after reinstalling the app.

### Environment

android Galaxy S26
main-branch

### Steps to Reproduce

1. Use the app with existing DM conversations.
2. Uninstall and reinstall the app.
3. Sign back into the same account.
4. Open direct messages.

### Actual Result

A few DM chats are missing after the reinstall.

### Expected Result

All previously available DM chats should be restored and visible after signing back in.

### Regression?

Yes

### Evidence

I only see 3 DM chats but i had a lot more

<img width="322" alt="Image" src="https://github.com/user-attachments/assets/14ef4004-959f-4119-a9e2-11409056852d" />

hm21:
@realmeylisdev I have no clue why that issue is back because after your last fix it definitely worked correctly for a while, but since today it seems like the issue has returned but i'm not sure what we changed that the issue is back.

realmeylisdev:
> [@realmeylisdev](https://github.com/realmeylisdev) I have no clue why that issue is back because after your last fix it definitely worked correctly for a while, but since today it seems like the issue has returned but i'm not sure what we changed that the issue is back.

@hm21 thank you for detecting this one, any logs would be really helpful

hm21:
@realmeylisdev sure here are the logs

[openvine_full_logs_2026-06-15T17-44-31.600460.txt](https://github.com/user-attachments/files/28962147/openvine_full_logs_2026-06-15T17-44-31.600460.txt)

Issue #4953, opened 2026-06-07
fix: missing DM chats after reinstall


hm21:
### Summary

Some direct message chats are missing after reinstalling the app, and the missing chats appear to be random.

### Environment

android galaxy s26
main-branch

### Steps to Reproduce

1. Use the app with multiple DM chats.
2. Uninstall and reinstall the app.
3. Log back into the same account.
4. Open the DM chat list.

### Actual Result

Only a subset of DM chats is displayed, often around 4 chats, while other existing chats are missing.

### Expected Result

All previously available DM chats should be visible after reinstalling and logging back into the account.

### Regression?

Unknown

### Evidence

I personally have this issue, so I can test the fix once it's available.

PR #6431, opened 2026-07-27
fix(dm): resolve four cases where DMs go missing with no loader

### LINKED ISSUE #6433: fix(dm): DMs go missing with no loader in four distinct paths
## Summary

Four separate DM defects make messages disappear while the UI shows a confident empty state — no spinner, no banner, no error — so the app looks like it has nothing rather than like it is still loading or has failed.

## Environment

```
App version: 1.0.17+506 (all four also present in 1.0.16)
Device: Any (iOS + Android; logic-layer, not platform-specific)
Network: Bugs 1 and 2 need a flaky or slow relay connection; bugs 3 and 4 reproduce on any network
```

Found by code audit of the DM stack, not from a single support report. Each is traced to a specific line below.

---

### 1. Message requests are silently hidden when history recovery does not finish

`ConversationListBloc` blanks the message-request split whenever `isHistoryRecoveryComplete` is false ([`conversation_list_bloc.dart:197-204`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart#L197-L204)):

```dart
final recoveryComplete = _dmRepository.isHistoryRecoveryComplete;
final requests = recoveryComplete ? split.requests : const <DmConversation>[];
```

That hold-back is intentional and correct (#5304). The bug is that its only user-visible affordance is driven by a **different signal**. The restore progress bar reads `isRestoringHistory` (`_activeRecoveryOps > 0`), while the gate reads the persisted `historyDrainComplete`.

`_endRecovery()` runs in a `finally` on **every** drain exit path; `markHistoryDrainComplete` runs on exactly **one** — clean exhaustion. The two therefore desync on five paths:

| `dm_repository.dart` | Condition | Logged? |
| --- | --- | --- |
| `:1326` | page fetch returned null | no — silent |
| `:1337-1343` | empty page with zero connected relays | warning |
| `:1391-1398` | exhausted, but outgoing-NIP-04 pass could not reach a relay | warning |
| `:1399-1409` | page cap (50) reached | warning |
| `:1411-1427` | any exception | error |

`retryPendingDecryptions` drives the same counter independently, which is a second desync source.

On all of those the bar switches off while the gate stays shut, and because the requests banner is itself gated on `requestConversations` being non-empty ([`inbox_view.dart:564,567`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/lib/screens/inbox/inbox_view.dart#L564)), there is no banner either — and `requestUnreadCount` reads 0.

### 2. A conversation claims "no messages" while it is still syncing

`ConversationBloc._onStarted` projects `watchMessages` + `watchOutgoing` — both **local Drift streams** — and sets `loaded` on the first tick ([`conversation_bloc.dart:91-114`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/lib/blocs/dm/conversation/conversation_bloc.dart#L91-L114)). For a thread whose gift wraps have not arrived or decrypted yet, that first tick is `([], [])`, so the view renders the "no messages" profile card.

Nothing local can distinguish this from a genuinely new conversation: conversation rows are only ever written in the same transaction as a message insert, so "row exists" implies "has messages"; and the pending-gift-wrap table carries no conversation id, because an undecrypted wrap's conversation is unknowable.

Worst immediately after reinstall, and on remote-signer (Keycast) accounts where every gift-wrap decrypt is a network round trip.

### 3. Inbound DMs are discarded on the receive path with no trace

Three distinct losses in `dm_repository.dart`:

- **`:1641-1650` — a decrypt that *throws* never reaches the retry queue.** The failed-decrypt queue is written only when `_decryptRumor` returns `null`. A decryptor that raises is logged and dropped, with no retry ever. Remote-signer RPC failures are exactly what #5202's queue was built for, so this is a live gap in that fix.
- **`:1158-1161` — `deleteExhausted`'s row count is discarded.** Wraps abandoned at the 10-attempt cap are inbound DMs the user will never see, and nothing records that it happened.
- **`:1184-1196` — corrupt stored JSON is dropped inside a bare catch with no log at all.** We wrote that JSON ourselves, so a parse failure is an invariant violation.

### 4. Opening a DM from someone's profile shows a permanently empty thread

[`other_profile_screen.dart:188-196`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/lib/screens/other_profile_screen.dart#L188-L196) pushes the **raw peer pubkey** as the conversation id:

```dart
context.push(ConversationPage.pathForId(widget.pubkey), extra: [widget.pubkey]);
```

Every other entry point, and `DmRepository.sendMessage`, use `computeConversationId([self, peer])` — the sha256 of the sorted participants. Both are 64-character hex, so nothing type-checks the mismatch.

The route id therefore matches no persisted row. `sendMessage` also enqueues under the canonical id (`dm_repository.dart:2752`), so messages sent from that screen never appear in it either.

## Steps to Reproduce

**Bug 4 (deterministic, no setup):**
1. Have an existing DM thread with someone, with visible history.
2. Open that person's profile from a feed, search result, or profile link.
3. Tap **Message**.
4. Type a message and send it.

**Bugs 1 and 2 (post-reinstall, degraded network):**
1. Sign in, exchange DMs including at least one request from someone you do not follow.
2. Delete and reinstall the app, then sign back in.
3. Put the device on a flaky or blocked relay connection so the history drain cannot finish (or let it hit the 50-page cap on a large history).
4. Open **Inbox → Messages**, then open any conversation.

**Bug 3:** requires a remote-signer (Keycast) account whose decrypt RPC fails by throwing rather than returning null, during gift-wrap ingestion.

## Actual Result

- **4:** The thread is empty despite full history, and the message you just sent never appears in it. No error, no spinner.
- **1:** The Messages tab looks complete and settled. Pending message requests are absent, the requests banner does not render, the requests badge reads 0, and the restore progress bar is off. Nothing indicates anything is missing. It stays that way until some later inbox open happens to complete the drain.
- **2:** The conversation shows the "no messages yet" profile card, indistinguishable from a brand-new chat, then messages pop in later.
- **3:** The message simply never arrives. No placeholder row, no error, and after 10 failed attempts no queue row either — nothing in logs or Crashlytics records that a DM was permanently lost.

## Expected Result

- **4:** The thread shows existing history, and a sent message appears in it.
- **1:** When requests are being withheld, the user is told so and given a way to retry — and the affordance must be honest on the paths where nothing is actually running.
- **2:** An empty conversation should not assert emptiness while recovery may still owe it messages.
- **3:** A throwing decrypt should be retried like a null one, and permanent give-up should be observable in logs and Crashlytics.

## Regression?

Yes — partly. Verified against full history (`git fetch --unshallow`; note the default clone is shallow, which silently breaks this kind of archaeology):

| # | Introduced by | Date | First release |
| --- | --- | --- | --- |
| 1 | `1aaee156b6` — #5311, the fix for #5304 | 2026-06-18 | 1.0.16 |
| 2 | `35d7648b48` — #2118, the original NIP-17 data layer | 2026-03-11 | 1.0.10 |
| 3 | `49a10f21ab` — #5214, the fix for #5202 | 2026-06-16 | 1.0.16 |
| 4 | `8ea4c8b43d` — #2571, the new Profile screen | 2026-04-22 | 1.0.10 |

Bug 4 is a regression from the profile redesign, which has produced at least one other regression of the same shape (#5213, collabs missing from own profile). Bugs 1 and 3 are gaps introduced *by* the two reinstall-recovery fixes in 1.0.16. Bug 2 has been present since DMs shipped.

## Evidence

Line-level traces are inline above. The recovery-gate desync in #1 is the load-bearing one: [`dm_repository.dart:1428-1438`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/packages/dm_repository/lib/src/dm_repository.dart#L1428-L1438) (`_endRecovery` in a `finally`, every path) versus [`:1382`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/packages/dm_repository/lib/src/dm_repository.dart#L1382) (`markHistoryDrainComplete`, one path).

No screen recording — these are reproducible only under specific relay/recovery conditions, and bug 4 is deterministic from the steps above.

realmeylisdev:
## Description

**Related Issue:** Closes #6433

Four defects where the DM UI renders a confident empty state while data is absent, still in flight, or permanently lost. Same class of bug — the UI asserts "there is nothing here" when it does not actually know that — so they ship together. One commit each.

### 1. Message requests silently blanked (`fix(dm): surface message requests held back by the recovery gate`)

The #5304 gate hides would-be message requests until the history drain completes. That hold-back is intentional. The defect is that its only user-visible affordance is driven by a **different signal**: the restore progress bar reads `isRestoringHistory` (`_activeRecoveryOps > 0`), while the gate reads the persisted `historyDrainComplete`.

`_endRecovery()` runs in a `finally` on **every** drain exit; `markHistoryDrainComplete` runs on exactly **one** — clean exhaustion. So on five paths the bar goes off while the gate stays shut:

| `dm_repository.dart` | Condition |
|---|---|
| `:1326` | page fetch returned null (silent — no log) |
| `:1337-1343` | empty page with zero connected relays |
| `:1391-1398` | exhausted, but the outgoing-NIP-04 pass could not reach a relay |
| `:1399-1409` | page cap (50) reached |
| `:1411-1427` | any exception |

`retryPendingDecryptions` drives the same counter independently, which is a second desync source.

The user then saw an inbox that looked complete: no request rows, no banner (it is gated on `requestConversations` being non-empty), `requestUnreadCount` reading zero, no progress bar, no spinner — while real pending DMs stayed hidden until some later inbox open happened to complete the drain.

New `requestsWithheld` tracks the gate itself, and only when it is actually holding something back, so a shut gate with nothing behind it stays silent. It drives a static banner — deliberately not a progress bar, since on these paths nothing is running — whose Retry re-enters `_onStarted`. Routing through the stream rather than awaiting the drain matters: the drain returns early without touching the recovery stream when it is uninitialized or already complete, so a listener-only retry could leave the banner with nothing to clear it.

The banner mounts as a sibling of the list content, so it covers every layout branch without threading a flag through each or shifting the sliver banner offsets. The existing progress bar is untouched and still owns the genuinely-in-flight case.

### 2. Thread claims "no messages" while still syncing (`fix(dm): stop asserting a conversation is empty while it may still be syncing`)

`ConversationBloc` projects two **local drift streams** and sets `loaded` on the first tick. For an unsynced thread that tick is `([], [])`, so the view renders the "no messages" profile card — visually identical to a genuinely new conversation. Worst right after a reinstall and on remote-signer accounts, where every gift-wrap decrypt is a network round trip.

Nothing local distinguishes the two: conversation rows are only ever written in the same transaction as a message insert, so "row exists" implies "has messages"; and the pending-gift-wrap queue carries no conversation id, because an undecrypted wrap's conversation is unknowable.

So this qualifies the card using the app-scoped recovery signals instead of inventing per-conversation sync state. `DmRestoreStatusCubit` publishes them, seeded from the repository because `historyRecoveryStream` does not replay, and re-reading the persisted flag on every tick so a drain that stopped short still reports outstanding work.

It is a **sibling** of `ConversationBloc`, not a third stream in its `combineLatest` — that bloc's emission sequence and mark-as-read call counts are pinned by four tests, and recovery status is app-scoped rather than per-conversation. Those four tests needed zero changes.

`EmptyConversation` keeps its avatar, name, NIP-05 and View profile button unchanged and gains one optional trailing line, defaulting off, so the three genuinely-empty entry points still read as a clean profile card.

### 3. Inbound DMs lost on the receive path (`fix(dm): stop losing inbound DMs on the decrypt-retry path`)

- **A throwing decrypt never reached the retry queue.** Only a `null` return was queued, so a remote-signer failure that *raised* — the exact case #5202's queue exists for — dropped the wrap permanently with no retry. It now takes the same queueing path, guarded on the session generation so a switched account cannot file the wrap under the wrong owner.
- **`deleteExhausted`'s row count was discarded.** Wraps abandoned at the cap are inbound DMs the user will never see; the count is now logged and reported.
- **Corrupt stored JSON was dropped inside a bare catch with no log at all.** We wrote that JSON ourselves, so a parse failure is an invariant violation; now logged and reported.

No UI surface: a pending row carries neither sender nor conversation (the wrap is still encrypted), so an undecryptable message cannot be attributed to a thread or rendered as a placeholder.

### 4. Profile "Message" opened a permanently empty thread (`fix(dm): route profile Message to the canonical conversation id`)

`OtherProfileView` pushed `pathForId(widget.pubkey)` — the **raw peer pubkey** — while every other entry point and `DmRepository.sendMessage` use `computeConversationId([self, peer])` (sha256 of the sorted participants). Both are 64-char hex, so nothing caught the mismatch.

The route id therefore matched no persisted row: opening DMs from a profile showed an empty conversation **even with full history**, and messages sent from that screen were enqueued under the canonical id, so they never appeared in the thread either.

## Out of Scope

- No per-conversation sync/EOSE plumbing. `NostrClient.subscribe` supports `onEose` but `DmRepository.startListening` never passes it and the subscription handle is private; wiring that up is a larger change than these fixes warrant, and the app-scoped signals cover the reported symptom.
- No UI for undecryptable messages (see #3 — unattributable by construction).
- `maxDecryptRetries` (10) is unchanged; this PR makes the give-up observable, it does not retune it.

## Verification

- `flutter test` (full mobile suite)
- `flutter test --coverage` in `packages/dm_repository` — 538 passing, 93.82% vs the package's `min_coverage: 93`
- `flutter test test/l10n/arb_consistency_test.dart` — both new keys mirrored into all 18 `app_*.arb` locales
- All four design-system ratchets (`raw_textstyle`, `raw_colors`, `material_button`, `raw_dialog`), plus `untested_services_floor` and `test_unit_structure` — no growth

New regression tests, each verified to fail against the pre-fix code:

| Area | Test |
|---|---|
| Bug 1 | `requestsWithheld` true when the gate hides real requests; false when it hides nothing; false once recovery completes; retry re-arms both passes and re-reads the gate |
| Bug 1 | Banner renders/hides on the flag; Retry dispatches the event |
| Bug 2 | Cubit seeds from the repository, reports outstanding work for a drain that stopped short, re-reads on each tick, cancels on close |
| Bug 2 | Empty card stays unqualified when recovery is complete; qualified when not, with the profile card intact |
| Bug 3 | A throwing decryptor lands in the retry queue; exhausted-cap drop reports; corrupt-JSON drop reports |
| Bug 4 | Profile Message routes to `computeConversationId([self, peer])`, not the raw pubkey |

## Type of Change

### DIFF
--- a/mobile/lib/blocs/dm/conversation/conversation_bloc.dart
+++ b/mobile/lib/blocs/dm/conversation/conversation_bloc.dart
@@ -71,6 +71,16 @@ class ConversationBloc extends Bloc<ConversationEvent, ConversationState> {
   ) async {
     emit(state.copyWith(status: ConversationStatus.loading));
 
+    // Arm the one-time history drain here too, not just on inbox open.
+    // Profile → Message reaches a thread without ever passing through
+    // Messages, so on a reinstall this can be the first DM surface the user
+    // touches. Until the drain arms, `hasAttemptedHistoryRecovery` is false
+    // and an empty thread reports itself complete while the user's history is
+    // still sitting on the relay. Idempotent, resumable and shared with the
+    // inbox's call, so arriving from either entry point costs one drain.
+    // See #4953.
+    unawaited(_dmRepository.backfillHistoryIfNeeded());
+
     // Mark as read when opening
     await _dmRepository.markConversationAsRead(_conversationId);
 
--- a/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
+++ b/mobile/lib/blocs/dm/conversation_list/conversation_list_bloc.dart
@@ -54,6 +54,10 @@ class ConversationListBloc
     );
     on<ConversationListNavigationConsumed>(_onNavigationConsumed);
     on<ConversationListBlocklistChanged>(_onBlocklistChanged);
+    on<ConversationListRestoreRetryRequested>(
+      _onRestoreRetryRequested,
+      transformer: droppable(),
+    );
     on<ConversationListProfileRepositoryChanged>(_onProfileRepositoryChanged);
   }
 
@@ -257,6 +261,9 @@ class ConversationListBloc
         return state.copyWith(
           status: ConversationListStatus.loaded,
           conversations: visibleInbox,
+          // Only true when the gate is actually costing the user something —
+          // a closed gate with nothing to hold back needs no banner.
+          requestsWithheld: !recoveryComplete && split.requests.isNotEmpty,
           visibleConversations: _computeVisible(
             visibleInbox,
             unreadOnly: state.unreadOnly,
@@ -544,6 +551,21 @@ class ConversationListBloc
     ConversationListBlocklistChanged event,
     Emitter<ConversationListState> emit,
   ) {
+    add(const ConversationListStarted());
+  }
+
+  void _onRestoreRetryRequested(
+    ConversationListRestoreRetryRequested event,
+    Emitter<ConversationListState> emit,
+  ) {
+    // Same idiom as _onBlocklistChanged: re-entering _onStarted re-fires both
+    // recovery passes (idempotent, shared in-flight) and restarts the stream,
+    // so the gate is re-read and `requestsWithheld` recomputed. Going through
+    // the stream rather than awaiting the drain matters — the drain returns
+    // early without ever touching the recovery stream when it is uninitialized
+    // or already complete, so a listener-only retry could leave the banner up
+    // with nothing to clear it. The status switch in _onStarted keeps a loaded
+    // list from flashing a spinner.
     add(const ConversationListStarted());
   }
 
--- a/mobile/lib/blocs/dm/conversation_list/conversation_list_event.dart
+++ b/mobile/lib/blocs/dm/conversation_list/conversation_list_event.dart
@@ -82,6 +82,15 @@ class ConversationListProfileRepositoryChanged extends ConversationListEvent {
   List<Object?> get props => [profileRepository];
 }
 
+/// The user tapped Retry on the restore-paused banner.
+///
+/// Re-arms the one-time history drain and the failed-decrypt replay. Both are
+/// idempotent and share their in-flight future, so a stray tap during an active
+/// pass is a no-op rather than a second drain.
+class ConversationListRestoreRetryRequested extends ConversationListEvent {
+  const ConversationListRestoreRetryRequested();
+}
+
 /// Resolve names for newly streamed conversations without changing the query.
 class _ConversationListProfileResolutionRequested
     extends ConversationListEvent {
--- a/mobile/lib/blocs/dm/conversation_list/conversation_list_state.dart
+++ b/mobile/lib/blocs/dm/conversation_list/conversation_list_state.dart
@@ -32,6 +32,7 @@ class ConversationListState extends Equatable {
     this.potentialRequests = const [],
     this.hasMore = true,
     this.isRestoringHistory = false,
+    this.requestsWithheld = false,
     this.currentLimit = ConversationListState.pageSize,
     this.navigationTarget,
   });
@@ -87,6 +88,19 @@ class ConversationListState extends Equatable {
   /// indicator at the top of the Messages list. See #5202.
   final bool isRestoringHistory;
 
+  /// Whether the #5304 recovery gate is currently hiding conversations that
+  /// would otherwise appear as message requests.
+  ///
+  /// Distinct from [isRestoringHistory] on purpose. That flag tracks whether a
+  /// drain is *actively running* (`_activeRecoveryOps > 0`), but the gate reads
+  /// the *persisted* `historyDrainComplete`, which only ever flips on a clean
+  /// exhaustion. A drain that stops at the page cap, hits an exception, or
+  /// finds no connected relay clears the running flag while leaving the gate
+  /// shut — so the progress bar vanished while requests stayed hidden, with no
+  /// banner, no badge and no spinner to show anything was missing. Drives the
+  /// restore-paused banner, which offers the retry that reopens the gate.
+  final bool requestsWithheld;
+
   /// Size of the render window over [conversations] — grows as the user pages.
   ///
   /// Not a fetch bound: the conversations are already loaded in full, so
@@ -121,6 +135,7 @@ class ConversationListState extends Equatable {
     List<DmConversation>? potentialRequests,
     bool? hasMore,
     bool? isRestoringHistory,
+    bool? requestsWithheld,
     int? currentLimit,
     ConversationNavigationTarget? navigationTarget,
     bool clearNavigationTarget = false,
@@ -136,6 +151,7 @@ class ConversationListState extends Equatable {
       potentialRequests: potentialRequests ?? this.potentialRequests,
       hasMore: hasMore ?? this.hasMore,
       isRestoringHistory: isRestoringHistory ?? this.isRestoringHistory,
+      requestsWithheld: requestsWithheld ?? this.requestsWithheld,
       currentLimit: currentLimit ?? this.currentLimit,
       navigationTarget: clearNavigationTarget
           ? null
@@ -155,6 +171,7 @@ class ConversationListState extends Equatable {
     potentialRequests,
     hasMore,
     isRestoringHistory,
+    requestsWithheld,
     currentLimit,
     navigationTarget,
   ];
--- a/mobile/lib/blocs/dm/restore_status/dm_restore_status_cubit.dart
+++ b/mobile/lib/blocs/dm/restore_status/dm_restore_status_cubit.dart
@@ -12,11 +12,15 @@ import 'package:equatable/equatable.dart';
 class DmRestoreStatusState extends Equatable {
   const DmRestoreStatusState({
     this.isRestoring = false,
+    this.hasAttempted = false,
     this.isComplete = true,
   });
 
   /// A recovery pass (reinstall backfill or failed-decrypt replay) is running.
   final bool isRestoring;
+
+  /// History recovery has run, completed, or is running for this user.
+  final bool hasAttempted;
 
   /// The one-time history drain has completed cleanly for this user.
   ///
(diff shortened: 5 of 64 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I pushed takeover commit a4a5744bab to close out the remaining review findings.\n\nWhat changed:\n- hid the inbox restore-paused banner while an active history drain is showing progress\n- made conversation empty-state qualification require an attempted/in-flight recovery, so brand-new never-restored chats stay clean\n- added the withheld-requests handling to /inbox/requests, including retry wiring\n- kept exhausted decrypt retry-cap drops as warning-only telemetry while preserving corrupt JSON reporting\n- split the conversation-specific restore copy key and regenerated l10n\n- updated the stale decrypt comment and bumped the retry button size in the banner\n\nVerification:\n- flutter gen-l10n\n- dart format lib test integration_test\n- flutter analyze lib test integration_test\n- targeted inbox/conversation/router tests\n- dm_repository dart analyze\n- dm_repository flutter test --coverage: 540 passed, 93.83%\n- GitHub checks are green on a4a5744bab\n\nI did a final review pass after the takeover commit and do not see remaining blockers. Please review the takeover changes when you get a chance and merge when ready.

NotThatKindOfDrLiz (approved):
Final review after takeover: the blockers from the earlier review are fixed, the package/reportability concerns are addressed, and CI is green on a4a5744bab. No remaining findings from my pass.

realmeylisdev:
Picked up both points from your commit, @NotThatKindOfDrLiz — e16942a1f7.

**`hasAttempted` gating** is right, and I kept it. One gap it opened: the flag only flips once the drain arms, and the drain arms from the inbox BLoC (`backfillHistoryIfNeeded` → `_runHistoryDrain` → `upgradeDrainVersionIfNeeded`, `dm_repository.dart:1279`). Profile → Message reaches a thread without passing through Messages, so after a reinstall that route left the flag false and an empty conversation reported itself complete while history was still on the relay — the #4953 symptom, on the entry point 50d33c1 added. `ConversationStarted` now arms the same idempotent drain the inbox does; either entry point stamps attempted-ness and the two share one run.

**The abandoned-wrap report** I put back. The doc comment that went with it said it was the only signal the loss happened, and what remained was a local warning that does not reach triage — for the one failure class this branch is about. I think the flood concern does not bite at this call site: it fires once per drain pass with an aggregate count, never per wrap, so a spam burst of undecryptable kind-1059 events costs one event. I wrote that reasoning into the constant's doc so it is a recorded decision either way. If you removed it because you saw it firing routinely in practice, say so and I will drop it again — that would be evidence I do not have.

Everything else in your commit I left alone; the conversation-scoped copy and the requests-screen loading/paused split are clear improvements. 481 tests green across the 26 suites that mock `DmRepository`, 540 in `dm_repository`, analyze clean.

Issue #494, opened 2025-12-10
[TC-NOTIF-005] Notifications screen


Statyk7:
## Test Case: TC-NOTIF-005

**Section:** 9. Notifications
**Feature:** 9.2 Notification Feed
**Priority:** P1
**Status:** Not Tested
**Size:** M
**Estimate:** 2 days
**Release:** MVP Rel 2

### Test Steps
Navigate to notifications tab
Verify all notifications listed
Verify notifications sorted by time
Verify unread indicators

### Needs More Information
Where should notifications be fetched from? (nostr, 3rd party service?)

---
*Imported from QA Test Case Tracker*

Issue #471, opened 2025-12-10
[TC-PROFILE-006] Edit profile


Statyk7:
## Test Case: TC-PROFILE-006

**Section:** 6. Profile Management
**Feature:** 6.2 Profile Editing
**Priority:** P1
**Status:** Not Tested
**Size:** M
**Estimate:** 2 days
**Release:** MVP Rel 2

### Test Steps
Navigate to profile settings
Edit display name
Edit bio
Save changes
Verify Kind 0 event published
Verify changes appear immediately

### Notes
Verify changes appear immediately
- Sometimes the profile image is not displayed correctly (black image is shown)
- First time I edited the profile I got a black screen after saving changes

---
*Imported from QA Test Case Tracker*

PR #6434, opened 2026-07-27
fix(blocklist): deliver a lifted block to the person who was blocked

### LINKED ISSUE #6435: fix(blocklist): a lifted block never reaches the person who was blocked — their DMs and profile stay hidden forever
Reported from a real two-device session on 2026-07-27.

## Symptom

A blocks B, then A **unblocks** B. On B's device:

- B's conversation with A stays hidden in Messages.
- Opening A's profile shows **Account not available**.
- The same profile opens normally from a third account.
- Messages B sends to A arrive fine — only the direction into B is affected.

It does not clear. Restarting the app does not help.

### 1. Client: the removal can never be delivered

B learns "A blocked me" only through a `#p = <B>` filter on A's kind-30000 block list (`mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart:1096`), with the kind-10000 mute twin at `:1018`.

A block is expressed by A's list **containing** `["p", "<B>"]`. An unblock is expressed by the **absence** of that tag — so the replacement event stops matching the filter and is never delivered. The "they unblocked us" branches in `_handleOthersBlockListEvent` (`:1550-1562`) and `_handleMuteListEvent` were therefore unreachable in production.

`_blockedByOthers` is a member of `_hideBuckets` (`:786`), so a single stale entry drives both symptoms at once:

- `shouldFilterFromFeeds` → `filterBlockedConversations` (`:943`) drops the whole conversation. Callers: `lib/blocs/dm/conversation_list/conversation_list_bloc.dart:207` and `lib/blocs/dm/unread_count/dm_unread_count_cubit.dart:213`.
- `hasBlockedUs` (`:850`) gates `UserNotAvailableScreen` via `lib/router/widgets/other_profile_screen_router.dart:50`.

### 2. Relay: the stale list is actively re-served

`relay.divine.video` retains superseded replaceable/addressable events and collapses the NIP-01 coordinate *after* the tag predicate, so a `#p` query returns the newest version that still carried the tag. Measured against production for the same `(A, 30000, d=block)` coordinate:

```
{"kinds":[30000],"#p":[<B>]}                  -> created_at 1783998962, p=[<B>]   STALE
{"authors":[<A>],"kinds":[30000]}             -> created_at 1785149558, p=[]      CURRENT
{"kinds":[30000],"#p":[<B>],"#d":["block"]}   -> 0 events                         CORRECT
```

This is why restarting does not help: B re-derives the stale block on every cold start. Filed separately as divinevideo/divine-funnelcake#718.

## What is not wrong

- **The unblock published correctly.** A's kind-30000 and kind-10000 on the relay both have zero `p` tags.
- **No messages were lost.** `dm_repository` has no dependency on `content_blocklist_repository`, so gift wraps decrypt and persist normally — the hide is read-time only.
- **No relay-side block enforcement on DMs.** `filter_muted_pubkeys_column` is wired only to REST feeds and the notification inbox, never to relay REQ.

## Fix

Once an author is known to block or mute us, also watch their list **by author id** — the only filter their removal event can still match. The watch follows a recreated `NostrClient` and is dropped on identity change. Applied to the kind-10000 mute path too, which had the identical defect.

This recovers even under the relay's non-conformant dedup, so it does not depend on #718 landing.

## Regression coverage

Six tests, all confirmed red against `origin/main`'s version of the file and green with the fix, including a cold-start case driven by a relay stub that reproduces the actual dedup behaviour rather than idealised NIP-01 semantics.

Note the pre-existing test at `content_blocklist_repository_test.dart:1524` already claimed to cover this ("removes blocker when updated event no longer contains our pubkey") but pushed the unblock into the same stubbed stream — a delivery the relay would never make. It passed while the bug shipped.

## Related

- Fixed by #6434
- Relay-side root cause: divinevideo/divine-funnelcake#718
- Same in-memory sets as #4969 (identity-change clearing), which is a different trigger

realmeylisdev:
## Description

Unblocking someone never reaches the person who was unblocked. On their device the blocker's DM conversation stays hidden and their profile keeps showing **Account not available** — permanently, and **not cleared by restarting the app**.

Two independent defects stack, one per side.

### Client side (fixed here)

A blockee learns "X blocked me" only through a `#p = <us>` filter on X's kind-30000 block list (`content_blocklist_repository.dart:1096`), and the kind-10000 mute twin at `:1018`. A block is expressed by X's list *containing* `["p", "<us>"]`. An **unblock is expressed by the absence of that tag** — so the replacement event stops matching the filter and is never delivered. The "they unblocked us" branches in `_handleOthersBlockListEvent` and `_handleMuteListEvent` were unreachable in production.

The stale `_blockedByOthers` entry is a member of `_hideBuckets` (`:786`), so it feeds `shouldFilterFromFeeds` → `filterBlockedConversations` (`:943`, called from `conversation_list_bloc.dart:207` and `dm_unread_count_cubit.dart:213`) and gates `UserNotAvailableScreen` via `other_profile_screen_router.dart:50`. One stale bit hides both the conversation and the profile.

**Fix:** once an author is known to block or mute us, also watch their list *by author id* — the only filter their removal event can still match. The watch follows a recreated `NostrClient` (this app recreates it on auth flips and reconnects) and is dropped on identity change so it cannot leak across accounts.

### Relay side (not fixed here — needs a funnelcake change)

`relay.divine.video` retains superseded versions of replaceable/addressable events and resolves the NIP-01 coordinate collapse *after* the filter predicate, so a `#p` query keeps returning the last version that still carried the tag. Measured directly against production, same coordinate `(A, 30000, d=block)`:

```
{"kinds":[30000],"#p":[<blockee>]}                 -> created_at 1783998962, p=[<blockee>]   STALE
{"authors":[<blocker>],"kinds":[30000]}            -> created_at 1785149558, p=[]            CURRENT
{"kinds":[30000],"#p":[<blockee>],"#d":["block"]}  -> 0 events                               CORRECT
```

The two-tag form routes to the `Base` branch (`events_deduped FINAL`, keyed on the NIP-01 coordinate) and behaves correctly; the single-tag form routes to `TagTarget` with `skip_tag_filters: true` and collapses only over rows that already survived the tag predicate. Root cause is that `events_local` is `ReplacingMergeTree(indexed_at) ORDER BY (id)` — keyed on the physical event id, so superseded versions are never removed. NIP-01 makes latest-only storage a MUST for replaceable events, so this is a storage-layer violation, not just a REQ-return preference.

Consequence: the blockee re-derives the stale block on **every cold start**, which is why restarting does not help. This fix is verified to recover from that too — see the cold-start test.

Same-family prior art in divine-funnelcake: #494 (CLOSED) was this exact bug for kind-10000 mute lists on the REST path; #471 and #697 (both OPEN) are neighbouring dedup defects in the same file.

**Related Issue:** Closes #6435

## Out of Scope

- The relay-side dedup defect above. Needs a funnelcake fix; this PR is a client-side workaround that is correct independently of it.
- `unblockUser` is a silent no-op when the pubkey is not in `_runtimeBlocklist`.
- `_restoreBlocksFromRelay` (`:1447-1451`) merges relay `p` tags into `_runtimeBlocklist` and never removes, so a stale own-list copy can re-add a just-unblocked pubkey.
- Neither list publisher awaits a NIP-20 `OK`, so a failed publish is invisible.
- Separately noted while auditing the relay: it implements no NIP-42 AUTH and serves kind-1059 gift wraps to any unauthenticated `#p` query, contrary to NIP-59's guidance. That is how the measurements above were possible. Worth its own security triage.

## Verification

- `flutter test` in `mobile/packages/content_blocklist_repository` — 114/114 pass.
- `flutter test --coverage` — **100.00%** (541/541 lines), which this package's CI requires since its workflow sets no `min_coverage` override.
- All 6 new behaviour tests confirmed **red** against `origin/main`'s version of the lib file and **green** with the fix.
- 598 app-side tests across the 45 test files referencing `ContentBlocklistRepository` — all pass.
- `dart format` + `flutter analyze lib test` clean.
- Relay behaviour measured directly against `wss://relay.divine.video`, and the routing explanation confirmed by the one-tag vs two-tag experiment above.

Two notes on the tests. The pre-existing `content_blocklist_repository_test.dart:1524` already claimed to cover this ("removes blocker when updated event no longer contains our pubkey") but pushed the unblock into the same stubbed stream — a delivery the relay would never make. And the new stub deliberately reproduces the relay's non-conformant dedup rather than idealised NIP-01 semantics, so the fix is not being validated against a relay that does not exist.

Also confirmed against production that relay dedup does respect the `d` tag (22 authors return multiple distinct kind-30000 `d` tags, including one holding both `block` and `dm-contacts`), so watching by author cannot mask the `d=block` list behind another list from the same author.

## Type of Change

### DIFF
--- a/mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart
+++ b/mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart
@@ -1599,8 +1599,8 @@ class ContentBlocklistRepository {
 ///
 /// Once an author is known to block or mute us we additionally watch their
 /// list by author, which is the only filter their removal event can still
-/// match. Authors are kept watched after a lift so a later re-block is also
-/// observed immediately rather than waiting on discovery.
+/// match. Authors are kept watched after a lift so later list changes from a
+/// known author do not require tearing down and rebuilding the watch again.
 class _AuthorListWatch {
   _AuthorListWatch({required int kind, required String label})
     : _kind = kind,
--- a/mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart
+++ b/mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart
@@ -120,6 +120,18 @@ class ContentBlocklistRepository {
   // Users who have blocked us (populated from kind 30000 events with d=block)
   final Set<String> _blockedByOthers = <String>{};
 
+  // Direct by-author watches on the lists of everyone who blocks or mutes
+  // us. The `#p = us` discovery filters cannot observe a block/mute being
+  // *lifted*; see [_AuthorListWatch] for why.
+  final _AuthorListWatch _blockAuthorWatch = _AuthorListWatch(
+    kind: 30000,
+    label: 'block-list',
+  );
+  final _AuthorListWatch _muteAuthorWatch = _AuthorListWatch(
+    kind: 10000,
+    label: 'mute-list',
+  );
+
   // Latest replaceable kind-30000 block-list event timestamp per author.
   // Prevent stale relay delivery order from resurrecting old block state.
   final Map<String, int> _latestBlockListEventCreatedAtByAuthor =
@@ -255,6 +267,9 @@ class ContentBlocklistRepository {
       _latestMuteListEventCreatedAtByAuthor.clear();
       _latestBlockListEventCreatedAtByAuthor.clear();
       _severedFollowers.clear();
+      // "Who blocks/mutes us" is meaningless for the incoming account.
+      _blockAuthorWatch.clear();
+      _muteAuthorWatch.clear();
       // Force fresh subscriptions filtered on the new pubkey. On a
       // same-client switch the old subscription's listener is
       // intentionally left in place (no handle is retained to cancel
@@ -1007,6 +1022,12 @@ class ContentBlocklistRepository {
     // Store references for Nostr publishing
     _nostrClient = nostrService;
 
+    // Any existing by-author watch was opened on the previous client.
+    _muteAuthorWatch.rebind(
+      client: nostrService,
+      onEvent: _handleMuteListEvent,
+    );
+
     Log.info(
       'Starting mutual mute list sync for pubkey: $ourPubkey',
       name: 'ContentBlocklistRepository',
@@ -1084,6 +1105,12 @@ class ContentBlocklistRepository {
 
     _signer = signer;
     _nostrClient = nostrService;
+
+    // Any existing by-author watch was opened on the previous client.
+    _blockAuthorWatch.rebind(
+      client: nostrService,
+      onEvent: _handleBlockListEvent,
+    );
 
     Log.info(
       'Starting block list sync for pubkey: $ourPubkey',
@@ -1295,6 +1322,13 @@ class ContentBlocklistRepository {
       // They muted us - add to blocklist
       if (!_mutualMuteBlocklist.contains(muterPubkey)) {
         _mutualMuteBlocklist.add(muterPubkey);
+        // Watch this author directly; the #p filter that just delivered
+        // this event can never deliver the unmute that follows it.
+        _muteAuthorWatch.add(
+          muterPubkey,
+          client: _nostrClient,
+          onEvent: _handleMuteListEvent,
+        );
         _emitChange(
           BlocklistChange(pubkey: muterPubkey, op: BlocklistOp.muted),
         );
@@ -1495,6 +1529,13 @@ class ContentBlocklistRepository {
     if (stillBlocked) {
       if (!_blockedByOthers.contains(blockerPubkey)) {
         _blockedByOthers.add(blockerPubkey);
+        // Watch this author directly; the #p filter that just delivered
+        // this event can never deliver the unblock that follows it.
+        _blockAuthorWatch.add(
+          blockerPubkey,
+          client: _nostrClient,
+          onEvent: _handleBlockListEvent,
+        );
         _emitChange(
           BlocklistChange(pubkey: blockerPubkey, op: BlocklistOp.blockedUs),
         );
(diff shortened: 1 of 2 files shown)


realmeylisdev:
Relay-side root cause is now filed as divinevideo/divine-funnelcake#718 — single-tag REQ collapses the NIP-01 coordinate *after* the tag predicate, so `#p` keeps returning the last list version that still carried the tag. That is why a blockee re-derives a lifted block on every cold start.

This PR stays useful regardless: it recovers from the stale delivery client-side, and the cold-start test pins that behaviour against a stub that reproduces the relay's actual dedup rather than idealised NIP-01 semantics. Once #718 lands, the by-author watch becomes belt-and-braces rather than load-bearing.

hm21 (approved):
Solid diagnosis, and the filter-aware relay stub is the right call — the old stub could never have caught this.

Found one gap and pushed two fixes:

- `41729c21b` — the "follows a recreated NostrClient" guarantee didn't hold in the app. `blocklistSyncBridge` starts both syncs in one turn and neither body awaits, so mute sync set the shared `_nostrClient` before block sync compared against it; block sync then returned early and `_blockAuthorWatch.rebind` never ran (the discovery subscription died the same way — pre-existing). Each sync path now tracks its own client, and the regression test starts both syncs the way the bridge does.
- `8d39177c0` — each discovered blocker re-opened the whole watch, so N blockers meant N REQ/CLOSE pairs plus N× replay. Opens are now coalesced per event-loop turn.

Verified: 115/115 package tests, 100% coverage (566/566), 598 app-side tests across the 45 files referencing `ContentBlocklistRepository`, `dart format` + `flutter analyze` clean. Yours to merge.

mbradley (approved):
Verified Alex's two fixes independently at `8d39177c0` rather than on trust — the app-call-order repro now passes, and the coalescing takes 20 discovered blockers from 21 `subscribe()` calls to 2 without dropping any author from the watch. 115/115 package tests, 100% coverage (566/566), analyze and format clean. He reached the `_nostrClient` root cause and the churn independently; nothing further from me on either.

Two nits, neither blocking:

- The class doc at `:1602` still says authors are kept watched after a lift "so a later re-block is also observed immediately rather than waiting on discovery." A re-block republishes a list containing `p: <us>`, which the persistent `#p` discovery subscription already matches, so there is no waiting to avoid. The retention still earns its place — it avoids a teardown/reopen for an author already covered, which is what the `add()` doc now says — but the rationale in the class doc doesn't hold. I found this trying to write a regression test for it and discovering the behaviour is unfalsifiable.
- A failed `_open()` logs and leaves the watch dead until the next `add()` or `rebind()`. Consistent with the swallow-and-log posture around it, so fine to leave, but the failure mode is "this session never sees the unblock."

NotThatKindOfDrLiz (approved):
Final takeover pass is clean. The original client-swap blocker is fixed by the per-sync client tracking and bridge-order regression test; the burst subscription churn is coalesced; the last doc/stub cleanup is now pushed. Local verification and GitHub checks are green.

NotThatKindOfDrLiz:
Takeover pass pushed in `2bb5d9f823` after rebasing onto current `origin/main`.

What changed:
- clarified the by-author watch doc so it no longer claims re-block discovery depends on the retained author watch
- replaced the relay test stub's hand-rolled filter subset with `Filter.checkEvent(event)`, so the stub tracks SDK filter semantics

Verified locally:
- `flutter test` from `mobile/packages/content_blocklist_repository` — 115/115 passing
- `flutter analyze lib test` from `mobile/packages/content_blocklist_repository` — clean
- pre-push analyzer and changed-file tests — passing

GitHub checks are green on the new head, including Content Blocklist Repository CI and Mobile CI. @realmeylisdev please review the takeover commit and merge when you're ready.

PR #6439, opened 2026-07-27
fix(video-editor): give a recorded clip its media's length

hm21:
## Description

A recorded clip's length came from a stopwatch — how long the record button was held. The media it produces is a different number: capture stops the audio and video writers independently, so an mp4's two tracks end tens of milliseconds apart, and the container declares the *longer* one. The clip's tail therefore carries picture with no sound.

Every editor anchor is authored against that tail — trims, layer time windows, transitions, filter windows — and the export cannot fill it. What the feed's looping player replays every cycle is that stretch of silence, which is the seam in #6386. On a multi-clip export the same gap sits at every clip join, not only at the loop point.

This sets a recorded clip's length to the point where both its tracks still have content during the recorder's existing metadata enrichment pass. The editor timeline then matches the media, and the export follows from the existing per-segment `endTime` without any special-casing.

- `commonTrackEnd(VideoMetadata)` returns the common track end, or `null` when there is nothing to correct: no audio track, tracks already ending together, or a shortfall beyond 500 ms.
- `VideoRecorderBloc._enrichAndSaveClip` applies it while reusing the metadata probe the recorder already performs for duration, thumbnail, ghost frame, and the enriched library save. The recorder still shows the clip instantly, then the detached enrichment writes the aligned duration once.
- Only ever shortens. A clip that genuinely outlasts its audio — a stop-motion still held past a short sound, a source with a broken audio track — keeps its full length.

### Why not fix this at export time

The first version of this PR passed a `trimToCommonTrackEnd` flag into the native renderer and bumped `pro_video_editor` to 2.10.0. That worked on the seam but was the wrong layer, for two reasons.

It fixed the symptom after the whole editor had already worked with the wrong number, so the export no longer matched what the user saw. Measured on a two-clip export whose clips each carried a 300 ms gap: the trim shortened each clip, so clip 2 started at 2.70 s instead of 3.00 s — but an overlay anchored to clip 2 still fired at 3.00 s, **300 ms late**. The drift accumulates over the preceding clips, and colour filters drift with it.

And it was Apple-only. `trimToCommonTrackEnd` is implemented under `darwin/` and has no Android counterpart in 2.10.0 — measured on a Galaxy S25, a 300 ms gap rendered to the same 284 ms with and without the flag. Correcting the clip's length is platform-independent.

Fixing it at the source removes both problems and the dependency bump: `pro_video_editor` stays at 2.9.0.

## Out of Scope

**Already-recorded clips.** Only affects new recordings; clips already sitting in a draft keep their stored length. The player-side clamp from #6430, merged 2026-07-28, covers what is already published — Blossom is content-addressed, so a re-encode is a different file and a different event.

**Imported and remixed clips.** `VideoClipImportService` takes its duration from the source Nostr event, not from the file. Those are already-published videos that cannot be repaired in place, and #6430 covers their playback.

**Stop-motion clips.** They have no captured audio track; their length is authored from the frames, and `commonTrackEnd` returns `null` for them.

## Verification

- `flutter test test/providers/clip_manager_provider_test.dart test/services/video_editor/ test/blocs/video_recorder/ test/services/upload_manager_from_draft_test.dart` — 393 passed
- 7 tests for `commonTrackEnd`, including the degenerate-track floor guard
- Recorder stop-path regression coverage that drives detached enrichment through the final library save and asserts the aligned duration survives

### Device verification

Driven from a throwaway `integration_test` against the real render path.

**Real capture, iPad Air 11-inch (M4), iOS 26.5.2** — container vs. audio-track duration:

| | container | audio | gap |
|---|---|---|---|
| the recording itself | 6033 ms | 5998 ms | 35 ms |
| export from the recorded length | 6033 ms | 5998 ms | 35 ms |
| export from the aligned length | 6000 ms | 5998 ms | **2 ms** |

The residual 2 ms is one AAC access unit. The export is 33 ms shorter — the stretch that had no audio under it.

**Real capture, Galaxy S25, Android 16** — the alignment fires here too (`audioDuration` is implemented on Android), but this recording's gap was 25 ms, which is *shorter than one video frame* at 30 fps. A video cannot end between two frames, so there was nothing to cut and the export is unchanged:

| | container | audio | gap |
|---|---|---|---|
| the recording itself | 6166 ms | 6141 ms | 25 ms |
| export from the recorded length | 6166 ms | 6122 ms | 44 ms |
| export from the aligned length | 6166 ms | 6122 ms | 44 ms |

Feeding the same device a crafted 300 ms gap — comfortably more than one frame — shows the mechanism working on Android:

| | container | audio | gap |
|---|---|---|---|
| source | 3000 ms | 2700 ms | 300 ms |
| export from the declared length | 3000 ms | 2716 ms | 284 ms |
| export from the aligned length | **2700 ms** | 2693 ms | **7 ms** |

So the fix is platform-independent, unlike the export-side flag it replaces. What it cannot remove is a gap below one frame duration (~33 ms at 30 fps), or the ±20 ms of AAC framing the Android encoder adds on its own — the residual 44 ms above is that framing, not the capture gap.

**Crafted 300 ms gap, macOS 26.5.2** — container vs. audio-track duration of the export:

| | video | audio | delta |
|---|---|---|---|
| single clip | 2.700 s | 2.700 s | **0 ms** |
| two clips | 5.400 s | 5.400 s | **0 ms** |
| two clips + overlay | 5.400 s | 5.400 s | **0 ms** |

Sampling the audio envelope in 50 ms windows across the two-clip export shows continuous sound to 5.41 s — no hole at the clip join, which the un-aligned export had at 2.70–3.00 s.

Overlay anchored to the second clip, measured by sampling the rendered frames:

| | clip 2 starts at | overlay appears at | drift |
|---|---|---|---|
| export-side trim (previous approach) | 2.70 s | 3.00 s | +300 ms |
| capture-side alignment (this PR) | 2.70 s | 2.70 s | **0 ms** |

**Closes:** #none — #6386 was closed by the player-side fix #6430; this is the capture-side counterpart for new recordings.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -33,6 +33,7 @@ import 'package:openvine/providers/clip_manager_provider.dart';
 import 'package:openvine/providers/video_editor_provider.dart';
 import 'package:openvine/services/haptic_service.dart';
 import 'package:openvine/services/performance_monitoring_service.dart';
+import 'package:openvine/services/video_editor/clip_media_duration.dart';
 import 'package:openvine/services/video_recorder/camera/camera_base_service.dart';
 import 'package:openvine/services/video_thumbnail_service.dart';
 import 'package:path/path.dart' as p;
@@ -1089,9 +1090,10 @@ class VideoRecorderBloc
       final metadata = await ProVideoEditor.instance.getMetadata(
         EditorVideo.file(File(workCopyPath)),
       );
-      clipManager.updateClipDuration(clip.id, metadata.duration);
+      final clipDuration = commonTrackEnd(metadata) ?? metadata.duration;
+      clipManager.updateClipDuration(clip.id, clipDuration);
       Log.debug(
-        '📊 Video duration: ${metadata.duration.inMilliseconds}ms',
+        '📊 Video duration: ${clipDuration.inMilliseconds}ms',
         name: 'VideoRecorderBloc',
         category: LogCategory.video,
       );
--- a/mobile/lib/providers/clip_manager_provider.dart
+++ b/mobile/lib/providers/clip_manager_provider.dart
@@ -18,6 +18,7 @@ import 'package:openvine/providers/video_editor_provider.dart';
 import 'package:openvine/providers/video_publish_provider.dart';
 import 'package:openvine/services/file_cleanup_service.dart';
 import 'package:openvine/services/native_proofmode_service.dart';
+import 'package:openvine/services/video_editor/clip_media_duration.dart';
 import 'package:openvine/services/video_editor/video_editor_render_service.dart';
 import 'package:pro_video_editor/pro_video_editor.dart';
 import 'package:unified_logger/unified_logger.dart';
@@ -236,6 +237,7 @@ class ClipManagerNotifier extends Notifier<ClipManagerState> {
             }
 
             refreshClip(clip);
+            unawaited(_alignClipToRecordedMedia(clip.id));
           },
         ),
       );
@@ -262,7 +264,64 @@ class ClipManagerNotifier extends Notifier<ClipManagerState> {
     // This runs after trimming (if any) completes via the processingCompleter.
     unawaited(_generateClipProof(clip));
 
+    // A trimmed clip is realigned once its re-render lands, above.
+    if (!isClipTooLong) {
+      unawaited(_alignClipToRecordedMedia(clip.id));
+    }
+
     return clip;
+  }
+
+  /// Replaces a recorded clip's stopwatch length with the point where its media
+  /// tracks both still have content.
+  ///
+  /// The recorder times how long the button was held, but an mp4's declared
+  /// length is its *longest* track, and capture stops the audio and video
+  /// writers independently — so a recording's tail routinely carries picture
+  /// with no sound. Leaving that tail in the clip's length puts the whole
+  /// editor timeline out of step with the media: trims, layer windows and
+  /// transitions are authored against a stretch the export cannot fill, and the
+  /// feed's looping player replays the silence every cycle (#6386).
+  ///
+  /// Only ever shortens, and only within [clipTrackEndTolerance] — a clip that
+  /// genuinely outlasts its audio keeps its full length.
+  Future<void> _alignClipToRecordedMedia(String clipId) async {
+    final clip = getClipById(clipId);
+    final video = clip?.video;
+    if (clip == null || video == null) return;
+
+    final VideoMetadata metadata;
+    try {
+      metadata = await ProVideoEditor.instance.getMetadata(video);
+    } catch (e) {
+      // Keep the recorded length: a clip that cannot be probed is still
+      // editable, it just keeps the seam it was captured with.
+      Log.warning(
+        '⚠️ Could not read media length for clip $clipId: $e',
+        name: 'ClipManagerNotifier',
+        category: .video,
+      );
+      return;
+    }
+
+    if (!ref.mounted) return;
+
+    final trackEnd = commonTrackEnd(metadata);
+    if (trackEnd == null) return;
+
+    final current = getClipById(clipId);
+    if (current == null || current.duration <= trackEnd) return;
+
+    final aligned = current.copyWith(duration: trackEnd);
+    refreshClip(aligned);
+    unawaited(saveClipToLibrary(aligned));
+
+    Log.debug(
+      '✂️ Aligned clip $clipId to its common track end: '
+      '${current.duration.inMilliseconds}ms → ${trackEnd.inMilliseconds}ms',
+      name: 'ClipManagerNotifier',
+      category: .video,
+    );
   }
 
   /// Adds a frames-based stop-motion clip to the manager.
--- a/mobile/lib/providers/clip_manager_provider.dart
+++ b/mobile/lib/providers/clip_manager_provider.dart
@@ -18,9 +18,8 @@ import 'package:openvine/providers/video_editor_provider.dart';
 import 'package:openvine/providers/video_publish_provider.dart';
 import 'package:openvine/services/file_cleanup_service.dart';
 import 'package:openvine/services/native_proofmode_service.dart';
-import 'package:openvine/services/video_editor/clip_media_duration.dart';
 import 'package:openvine/services/video_editor/video_editor_render_service.dart';
-import 'package:pro_video_editor/pro_video_editor.dart';
+import 'package:pro_video_editor/pro_video_editor.dart' show EditorVideo;
 import 'package:unified_logger/unified_logger.dart';
 
 final clipManagerProvider =
@@ -237,7 +236,6 @@ class ClipManagerNotifier extends Notifier<ClipManagerState> {
             }
 
             refreshClip(clip);
-            unawaited(_alignClipToRecordedMedia(clip.id));
           },
         ),
       );
@@ -264,64 +262,7 @@ class ClipManagerNotifier extends Notifier<ClipManagerState> {
     // This runs after trimming (if any) completes via the processingCompleter.
     unawaited(_generateClipProof(clip));
 
-    // A trimmed clip is realigned once its re-render lands, above.
-    if (!isClipTooLong) {
-      unawaited(_alignClipToRecordedMedia(clip.id));
-    }
-
     return clip;
-  }
-
-  /// Replaces a recorded clip's stopwatch length with the point where its media
-  /// tracks both still have content.
-  ///
-  /// The recorder times how long the button was held, but an mp4's declared
-  /// length is its *longest* track, and capture stops the audio and video
-  /// writers independently — so a recording's tail routinely carries picture
-  /// with no sound. Leaving that tail in the clip's length puts the whole
-  /// editor timeline out of step with the media: trims, layer windows and
-  /// transitions are authored against a stretch the export cannot fill, and the
-  /// feed's looping player replays the silence every cycle (#6386).
-  ///
-  /// Only ever shortens, and only within [clipTrackEndTolerance] — a clip that
-  /// genuinely outlasts its audio keeps its full length.
-  Future<void> _alignClipToRecordedMedia(String clipId) async {
-    final clip = getClipById(clipId);
-    final video = clip?.video;
-    if (clip == null || video == null) return;
-
-    final VideoMetadata metadata;
-    try {
-      metadata = await ProVideoEditor.instance.getMetadata(video);
-    } catch (e) {
-      // Keep the recorded length: a clip that cannot be probed is still
-      // editable, it just keeps the seam it was captured with.
-      Log.warning(
-        '⚠️ Could not read media length for clip $clipId: $e',
-        name: 'ClipManagerNotifier',
-        category: .video,
-      );
-      return;
-    }
-
-    if (!ref.mounted) return;
-
-    final trackEnd = commonTrackEnd(metadata);
-    if (trackEnd == null) return;
-
-    final current = getClipById(clipId);
-    if (current == null || current.duration <= trackEnd) return;
-
-    final aligned = current.copyWith(duration: trackEnd);
-    refreshClip(aligned);
-    unawaited(saveClipToLibrary(aligned));
-
-    Log.debug(
-      '✂️ Aligned clip $clipId to its common track end: '
-      '${current.duration.inMilliseconds}ms → ${trackEnd.inMilliseconds}ms',
-      name: 'ClipManagerNotifier',
-      category: .video,
-    );
   }
 
   /// Adds a frames-based stop-motion clip to the manager.
--- a/mobile/lib/services/video_editor/clip_media_duration.dart
+++ b/mobile/lib/services/video_editor/clip_media_duration.dart
@@ -1,0 +1,33 @@
+// ABOUTME: Derives a recorded clip's true length from its media tracks
+// ABOUTME: so the editor timeline matches the file the export is built from
+
+import 'package:pro_video_editor/pro_video_editor.dart';
+
+/// How far a clip's audio may fall short of its video before the gap counts as
+/// content rather than a track-end mismatch.
+///
+/// Capture stops the audio and video writers independently, so the two tracks
+/// end a fraction of a second apart; a survey of published mp4s put the worst
+/// case at ~0.4s. A larger gap means the clip is genuinely meant to outlast its
+/// audio — a stop-motion still held past a short sound, or a source with a
+/// broken audio track — and shortening it would swallow content.
+const clipTrackEndTolerance = Duration(milliseconds: 500);
+
+/// The point where every track of a recorded clip still has content, or `null`
+/// when the clip's recorded length should be kept as it is.
+///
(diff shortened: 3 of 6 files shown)


NotThatKindOfDrLiz:
@hm21 I pushed the takeover fixes in `d3bc56c` and updated the PR body to match the actual scope.

What changed:

- Scoped `trimToCommonTrackEnd` to publish-ready renders only: final `renderVideoToClip` exports opt in, while merge-to-clip, clip-library flattening, seam previews, and normalization renders stay untrimmed by default.
- Added the same trim flag to the `UploadManager` multi-clip fallback render, since that recovery path uploads the rendered file directly when the persisted final render is missing.
- Added a test-only native render seam so tests can inspect the actual `VideoRenderData` built by production paths.
- Added regression coverage for publish-ready trim, intermediate renders staying untrimmed, and the upload fallback render carrying the trim flag.

Verification passed:

- `flutter pub get`
- `flutter test test/services/video_editor/`
- `flutter test test/services/upload_manager_from_draft_test.dart`
- GitHub CI is green on `d3bc56c`

I converted the PR back to draft because the remaining blocker is still native-device verification. Please run the device pass from the PR body before marking ready/merging, especially:

- iOS real recording export with `ffprobe` audio/video durations
- Samsung Galaxy real recording export with `ffprobe` audio/video durations
- stop-motion held past short sound confirms the 500 ms skip behavior
- editor-added music export remains unaffected
- multi-clip export with an end-anchored overlay stays visible through the end

mobile/lib/services/video_editor/video_editor_render_service.dart
```diff
@@ -450,6 +462,7 @@ class VideoEditorRenderService {
         usePersistentStorage: true,
         parameters: parameters,
         taskId: effectiveTaskId,
+        trimToCommonTrackEnd: true,
```
realmeylisdev: This line is the whole fix on the primary publish path, and nothing pins it. I deleted it and ran the suite: `test/services/video_editor/` 236 passed, `upload_manager_from_draft_test.dart` 8 passed — all green.

The new `sets the native trim flag on publish-ready renders` test calls `renderVideo(..., trimToCommonTrackEnd: true)` directly, so the flag is supplied by the test. It pins the `renderVideo` -> `VideoRenderData` threading, not the policy decision here. Same reasoning as the encoder-retry regression test in the body ("dropping the flag there would silently export the seam back in") applies one layer up.

Cheap to close: `video_editor_render_service_progress_test.dart` already drives the real `renderVideoToClip` through `_ProgressProVideoEditor`, which captures every `VideoRenderData` on its `renderStarts` stream. One assertion there covers it.

mobile/lib/services/video_editor/video_editor_render_service.dart
```diff
@@ -1094,6 +1114,13 @@ class VideoEditorRenderService {
       videoSegments: volumeSegments,
       endTime: maxOutputDuration,
       shouldOptimizeForNetworkUse: true,
+      // Capture stops the audio and video writers independently, so a recorded
+      // clip's tracks end tens of milliseconds apart. Exporting to the longer
+      // one ends on silence or a frozen frame, which the feed's looping player
+      // replays every cycle (#6386). Ending where both tracks still have
+      // content costs that mismatch — well under the 500ms the plugin treats
+      // as content rather than an encoder tail.
+      trimToCommonTrackEnd: trimToCommonTrackEnd,
```
realmeylisdev: The timeline anchors are computed in Dart before this and never reconciled afterwards, so the trim moves the output axis out from under them.

`timelineMap` (`TransitionTimelineMap.fromClips`, :1074) derives from `clip.playbackDuration` -> `duration`, which is the *container* duration, i.e. max(videoEnd, audioEnd) — the untrimmed value. Native then shortens clip *i* by its own delta and shifts every later `insertStart` (`VideoSequenceBuilder.swift:211,:343`), and the compositor applies layer windows with no clamp (`VideoCompositor.swift:618-620`).

Two consequences:
- `resolveLayerEndTime` pins a leave animation's `endTime` to the editor total (`video_editor_layer_animation_sheet.dart:159-176`), so the out phase is cut by the trim delta. The animation picker's range is 10-2000ms, so a clip at the 500ms bound suppresses any leave animation of 500ms or shorter outright.
- Anchors after clip *k* drift late by the running sum; a window landing entirely inside the trailing delta never satisfies `currentTimeUs >= startUs` and is dropped with no log.

The body already lists the end-anchored overlay as a device check — flagging that the mechanism is confirmed in code, and that the bound is per clip and cumulative rather than the "~0.1 s" single-clip framing.

Editor music is fine: `AudioSequenceBuilder` gets `targetDuration: videoResult.totalDuration` post-trim and re-clamps, so the Dart pre-clamp is only an upper bound.

mobile/lib/services/upload_manager.dart
```diff
@@ -413,6 +413,7 @@ class UploadManager implements BackgroundAwareService {
           videoSegments: videoSegments,
           endTime: VideoEditorConstants.maxDuration,
           shouldOptimizeForNetworkUse: true,
+          trimToCommonTrackEnd: true,
```
realmeylisdev: Correct for this branch, but its sibling a few lines up is the same recovery route and can't be covered the same way: `prepareUploadFromSourceClips()` at :389-393 returns `source.requireVideo.safeFilePath()` for a single clip — the raw capture, no render, so there is no `VideoRenderData` to set the flag on.

That file is exactly the one the PR targets: the recorder's own mp4, whose writers stopped tens of ms apart. It also skips the 6.3s cap, aspect normalization and `shouldOptimizeForNetworkUse`.

Reachable without touching the editor: `DraftStorageService._clearMissingFinalRenderedClip` nulls `finalRenderedClip` when its mp4 is gone, and the background publish-retry route reaches `startUploadFromDraft` without going through `video_publish_provider`. The interactive single-clip publish is covered — `postVideo` throws when `finalRenderedClip == null` and `canPost` gates the drafts entry — so this is confined to recovery, but single-clip is the modal draft shape.

Not asking you to add a render on a fallback path; worth calling out in Out of Scope alongside the multi-clip one rather than leaving "the fallback render is covered" to imply both.

mobile/lib/services/video_editor/video_editor_render_service.dart
```diff
@@ -356,6 +356,18 @@ class VideoEditorRenderService {
...
+  /// Lets tests inspect the exact [VideoRenderData] built by production paths
+  /// without invoking the native plugin. Reset to `null` in `tearDown`.
+  @visibleForTesting
+  static Future<String> Function(
+    String outputPath,
+    VideoRenderData task, {
+    NativeLogLevel? nativeLogLevel,
+  })?
+  renderNativeVideoToFileOverride;
+
```
realmeylisdev: Two small things on the new seam.

`check_process_global_mutations.sh` enumerates reset-to-default globals by hand and already lists `renderVideoOverride` (:99) and `renderVideoToClipOverride` (:100). This third one isn't there, so a future test that installs it without a tearDown reset ships green and strands `renderNativeVideoToFile` for every later suite in the merged isolate. One line at :101 closes it. (Both new test files do reset correctly — this is about enforcement for the next one.)

Also: the test file already declares `_MetadataProVideoEditor extends ProVideoEditor` and swaps `ProVideoEditor.instance` in setUp. `renderVideoToFile` is overridable on the platform interface, so one more override on that fake captures the same `VideoRenderData` with no new production API — and keeps `_cancelledTaskIds.remove` / `_cancelNativeTaskOnly` / `NativeRenderTaskRegistry.track` live during the test instead of short-circuiting above them.

mobile/lib/services/video_editor/video_editor_render_service.dart
```diff
@@ -616,13 +629,18 @@ class VideoEditorRenderService {
...
+  /// publish. Intermediate editor renders keep their nominal media tail so their
+  /// returned clip durations stay aligned with the editor timeline.
   static Future<String?> renderVideo({
     required List<DivineVideoClip> clips,
     bool usePersistentStorage = false,
     model.AspectRatio? aspectRatio,
     CompleteParameters? parameters,
     String? taskId,
     Duration? maxOutputDuration = VideoEditorConstants.maxDuration,
+    bool trimToCommonTrackEnd = false,
```
realmeylisdev: `renderVideoOverride`'s typedef (:349-357) wasn't extended with this parameter, and the forwarding call at :647-654 drops it. Any test stubbing that seam is blind to the flag — which is what pushed this PR into adding the deeper native seam. Adding the param here and forwarding it would have covered the `renderVideoToClip` call site through the existing seam.

mobile/lib/services/video_editor/video_editor_render_service.dart
```diff
@@ -616,13 +629,18 @@ class VideoEditorRenderService {
   /// [VideoEditorConstants.maxDuration] for the final-export path; pass `null`
   /// to render the full concatenation uncapped (e.g. when merging clips into an
   /// intermediate editor clip that the user can still trim).
+  ///
+  /// [trimToCommonTrackEnd] should only be enabled for files that are ready to
+  /// publish. Intermediate editor renders keep their nominal media tail so their
+  /// returned clip durations stay aligned with the editor timeline.
```
realmeylisdev: Minor wording, on the body rather than this comment: "the plugin treats a shortfall beyond 500 ms as content ... so a stop-motion clip held past a short sound cannot be shortened" is only true above the threshold. `TrackEndTrimmer.swift:65` is `shortfall <= maxTrackEndMismatch`, so a clip deliberately held 300ms past its sound is cut by 300ms.

Stop-motion is safe here, but for a different reason: the assembler emits a silent video (`stop_motion_render_service.dart:1`), so `loadAudioTrack` fails and the trimmer returns nil regardless of the bound. Accurate phrasing would be "a hold longer than 500 ms is left alone, and stop-motion clips are never trimmed because their assembly output is video-only."

Worth noting the export bound and the player bound aren't the same: #6430 clamps at 500ms **and** 10% of playable duration, this is 500ms only. For a sub-5s export the exporter can cut more than the player would have.

mobile/pubspec.yaml
```diff
@@ -327,7 +327,7 @@ dependencies:
   firebase_performance: ^0.11.1+4
   firebase_messaging: ^16.1.1
   audio_session: ^0.2.2
-  pro_video_editor: ^2.9.0
+  pro_video_editor: ^2.10.0
```
realmeylisdev: Heads-up rather than a change request: draft #6266 carries `pro_video_editor: ^2.7.1` and touches this file. If it merges after this one without a rebase that keeps `^2.10.0`, it silently reverts the bump and `trimToCommonTrackEnd` stops resolving.

realmeylisdev (commented):
Verified the mechanism end to end: the intersect in `VideoSequenceBuilder` fixes the audio-insert defect #6430 documented, and the 2.10.0 bump is additive (android/linux/windows byte-identical, flag-off path unchanged). Published `duration` stays correct — both upload paths probe the rendered file.

Three things worth resolving before this leaves draft, inline. The main one: deleting the flag from `renderVideoToClip` leaves the whole suite green, so the PR's headline change is unguarded.

hm21:
@realmeylisdev heads up — the diff you reviewed no longer exists. This PR was rewritten from scratch, so your review does not apply to what is on the branch now and it needs a fresh pass.

**What changed.** The original approach passed a `trimToCommonTrackEnd` flag into the native renderer and bumped `pro_video_editor` to 2.10.0. That is all gone: no flag anywhere in the tree, `pro_video_editor` stays at 2.9.0. The fix now sets a recorded clip's length to the point where both its tracks still have content, at the moment the clip is created — `commonTrackEnd(VideoMetadata)` plus a call in `ClipManagerNotifier.addClip`. Four files, +297 lines, no deletions.

**Why it was rewritten.** Two reasons, both measured rather than argued.

The export-side trim shortened each clip *after* the editor had already laid out its timeline against the un-shortened lengths, so the export stopped matching what the user saw. On a two-clip export with a 300 ms gap per clip, clip 2 started at 2.70 s instead of 3.00 s but an overlay anchored to it still fired at 3.00 s — 300 ms late, accumulating over the preceding clips. Colour filters drift the same way. Correcting the clip's length instead puts that drift at 0 ms.

And it was Apple-only. `trimToCommonTrackEnd` is implemented under `darwin/` with no Android counterpart in 2.10.0 — on a Galaxy S25 a 300 ms gap rendered to the same 284 ms with and without the flag. The new approach is Dart, so it works on both.

**Measurements.** Container vs. audio-track duration of the export.

Real capture, iPad Air 11-inch (M4), iOS 26.5.2:

| | container | audio | gap |
|---|---|---|---|
| the recording itself | 6033 ms | 5998 ms | 35 ms |
| export from the recorded length | 6033 ms | 5998 ms | 35 ms |
| export from the aligned length | 6000 ms | 5998 ms | **2 ms** |

Real capture, Galaxy S25, Android 16 — the alignment fires (6166 → 6141 ms) but this recording's gap was 25 ms, shorter than one frame at 30 fps. A video cannot end between two frames, so there was nothing to cut. Feeding the same device a crafted 300 ms gap shows the mechanism working:

| | container | audio | gap |
|---|---|---|---|
| source | 3000 ms | 2700 ms | 300 ms |
| export from the declared length | 3000 ms | 2716 ms | 284 ms |
| export from the aligned length | **2700 ms** | 2693 ms | **7 ms** |

macOS, crafted 300 ms gap: single clip and two clips both export at delta 0 ms, and sampling the audio envelope in 50 ms windows across the two-clip export shows continuous sound — no hole at the clip join, which the un-aligned export had at 2.70–3.00 s.

`flutter analyze lib test integration_test` clean, 394 tests green in the affected suites, all CI checks passing.

**Not in scope, and not caused by this.** Looping playback clicks at the loop point because the player never fades the composition edges. That is filed as #6468 and confirmed as pre-existing — the editor's loop end has never been tied to a zero crossing, so it could always land mid-waveform. This PR changes where that cut sits, not whether the click happens.

NotThatKindOfDrLiz:
@hm21 I pushed takeover fixes on top of the rewritten PR.

What changed:

- Moved the track-end correction out of `ClipManagerNotifier.addClip` and into `VideoRecorderBloc._enrichAndSaveClip`, reusing the recorder metadata probe that already drives duration, thumbnails, ghost frame, and the enriched library save.
- Removed the duplicate fire-and-forget metadata probe and the extra uncoordinated library write, so the aligned duration is not later clobbered by recorder enrichment.
- Added a common-track floor guard using the editor minimum clip duration, so degenerate short audio tracks cannot collapse a recording to zero length.
- Replaced the `addClip`-only tests with recorder stop-path coverage that drives detached enrichment through the final save and asserts the aligned duration survives.
- Updated the PR body to describe the actual write site and test coverage.

Verification:

- `flutter test test/services/video_editor/clip_media_duration_test.dart test/blocs/video_recorder/video_recorder_bloc_test.dart test/providers/clip_manager_provider_test.dart`
- pre-push analyzer + changed-file tests
- GitHub CI is green on `0446330`

Please review the takeover commits and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover fixes. The duration correction now happens at the recorder metadata enrichment write site, the duplicate metadata probe/library write is gone, the degenerate-track floor is covered, and CI is green on the latest head.

Issue #6337, opened 2026-07-23
feat: Signer-capable in-app route for the profile → divine.space CTA (own profile)


realmeylisdev:
## Context

Follow-up from #6329 (Closes #6328), which added a discoverable profile → divine.space entry point that opens `https://divine.space/{npub}` in the **system browser** (read-only). Per review discussion on #6329, the read-only external launch is the intended v1 scope for the *discovery* entry point.

This issue tracks the deferred, larger piece: routing the **own-profile** CTA through the signer-capable in-app sandbox so users can actually **customize** their divine.space page (themes, Top 8, profile music, bento layout) — the customization flow #6328 called out — without leaving the app.

## Why it's a separate, larger change

The external browser has no Nostr signer, so the own-profile "View your site" CTA is read-only there. The in-app `NostrAppSandboxScreen` injects a NIP-07 `window.nostr` bridge, but wiring the CTA to it is materially larger and security-touching:

- **Deep-link into the sandbox.** The sandbox loads `app.launchUrl` (site root). Opening `https://divine.space/{own npub}` specifically needs a per-view entry override or an `initialUrl` parameter.
- **Signer allowlist gap.** The bundled `divine-space` entry uses the shared `_sharedSignEventKinds` = `[1, 6, 7, 14, 15, 1111, 9734, 30023]` — it does **not** include kind 0 (profile metadata) or divine.space's customization kinds. Real edits mean expanding `allowedSignEventKinds` for divine.space (needs divine.space's event-kind list + a security review of what the third-party app may sign).
- **Auto-login.** `divine-space` has no `_autoLoginScripts` entry, so the user logs in manually inside the WebView. A seamless flow likely needs an auto-login script (as primal / badges / zap-stream have).
- **Navigation exceptions.** `firstPartyNostrAppNavigationBySlug` has no `divine-space` entry, so the sandbox is origin-locked to `https://divine.space`; any cross-origin top navigation would be blocked. Confirm divine.space stays same-origin or add nav exceptions.
- **Web fallback.** The sandbox is native-only (`supportsNostrAppsSandbox = !isWeb`); the CTA must fall back to external launch on web.

Visiting **another** creator's site stays external launch (read-only; nothing to sign).

## Acceptance criteria

- [ ] Own-profile "View your site" opens `divine.space/{own npub}` inside the in-app sandbox with the signer bridge available
- [ ] divine.space can sign the events its customization flows require (allowlist expanded + security-reviewed)
[...]

PR #6440, opened 2026-07-27
revert(profile): remove the divine.space creator-site CTA

### LINKED ISSUE #6441: revert(profile): remove the divine.space creator-site CTA
Auto-created from PR #6440.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6440
**Branch:** `chore/remove-profile-divine-space-cta`

hm21:
Closes #6441

## Description

The creator-site pill shipped in #6329 put a `divine.space` link on every profile (own and other). Product feedback is that people don't want it there, so the whole affordance comes back out.

Removed:

- `ProfileCreatorSiteButton` and its `divineSpaceProfileUri` / `isDivineSpaceProfileUrlForPubkey` helpers, plus the widget test.
- The `profileCreatorSiteVisitLabel` / `profileCreatorSiteOwnLabel` ARB keys in all 18 locales, regenerated with `flutter gen-l10n`.
- The `creator_site_cta_tapped` analytics event (`trackCreatorSiteCtaTapped`).
- `divine.space` from the trusted outbound hosts in `external_link_launcher.dart` — the profile CTA was its only consumer.

Two behaviour changes fall out of the removal, both intentional:

- **The tip/support pill no longer waits.** The pairing hold in `_ProfileCreatorLinksRow` (plus its 10s fallback timer and the `isMonetizationResolved` plumbing through the header) existed only so the creator-site pill and the Tip pill could reveal in the same frame instead of the creator site rendering first off Funnelcake's link-stripped REST projection (#6328). With one pill left there is nothing to pair, so the Tip pill goes back to rendering as soon as its links arrive — which is what the "shows the support button in the frame the links arrive" test already pins.
- **The Kind 0 website row is no longer filtered.** The header hid a `website` value that matched the generated `divine.space/<npub>` URL, to avoid showing the same destination twice. With no generated pill to duplicate, a creator who puts that URL in their profile now gets it rendered like any other website (and tapping it goes through the standard external-link confirmation, since the host is no longer trusted).

**Related Issue:** 

#6337 (signer-capable in-app route for this CTA) is a follow-up on the removed affordance and should be closed as obsolete if this lands.

## Out of Scope

- The `divine.space` entry in the apps directory (`preloaded_nostr_apps.dart`, #2756) stays — that is a separate surface, not the profile.
- The repository's raw-Kind-0 retry path, `UserProfile.isRestProjection`, and the `OtherProfileScreen` `BlocProvider` key all shipped in the same PR but are general fixes, not part of the CTA. Kept.

## Verification

- `flutter test test/widgets/profile test/screens/other_profile_screen_test.dart test/features/monetization` — 269 passing.
- `flutter test test/l10n/arb_consistency_test.dart`, `test/utils/external_link_launcher_test.dart` — passing.
- `check_raw_textstyle_ceiling`, `check_raw_colors_ceiling`, `check_material_button_ceiling`, `check_raw_dialog_ceiling`, `check_untested_services_floor`, `check_test_unit_structure` — all OK, no baseline changes needed.
- Manual pass on a real Samsung Galaxy still pending — draft until then. Check: own profile and another creator's profile, with and without tip/support links, and a profile whose Kind 0 `website` is set.

## Type of Change

- [x] 🧹 Code refactor
- [x] 🗑️ Chore


realmeylisdev (approved):
LGTM

PR #6446, opened 2026-07-27
fix(notifications): fetch category tabs independently

### LINKED ISSUE #6428: fix(support): follows category
### Summary

Kirsten Lynn Swasey, Jul 27, 2026, 18:49

follows category

it's not showing my new followers . in the category .

App Version: 1.0.17+668

### Device Information
- **platform:** android
- **model:** moto g power - 2025
- **manufacturer:** motorola
- **version:** 16
- **sdkInt:** 36
- **brand:** motorola

**Current Screen:** SupportCenterScreen
**User Pubkey:** cd4ce30f980c960757b46179608a4946fc06cad47f6dc8960f638e41312c1643

### Recent Logs (errors/warnings + recent context)
```
[2026-07-27T00:44:24.689859] [ERROR] (CrashReporting) Failed to initialize crash reporting: [core/duplicate-app] A Firebase App named "[DEFAULT]" already exists
[2026-07-27T00:44:26.255488] [WARNING] [SYSTEM] Zendesk JWT: SDK not initialized
[2026-07-27T00:44:26.255739] [WARNING] [SYSTEM] (ZendeskIdentitySync) Zendesk JWT upgrade failed, anonymous identity active
[2026-07-27T00:44:27.046697] [WARNING] [SYSTEM] Zendesk JWT: SDK not initialized
[2026-07-27T00:44:27.046699] [WARNING] [SYSTEM] (ZendeskIdentitySync) Zendesk JWT upgrade failed, anonymous identity active
[2026-07-27T00:44:27.635352] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/4b599880bcb3995c8fca5be81758a18fb87c57d265d47b14d6c232e92401be7c/720p.mp4 returned HTTP 404
[2026-07-27T00:44:27.679682] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/4b599880bcb3995c8fca5be81758a18fb87c57d265d47b14d6c232e92401be7c returned HTTP 404
[2026-07-27T00:44:27.723576] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/85cc1cb3bfe8f75b000133515656e68237d8dfc36f471703156d058f52ae94fa/720p.mp4 returned HTTP 404
[2026-07-27T00:44:27.782448] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/85cc1cb3bfe8f75b000133515656e68237d8dfc36f471703156d058f52ae94fa returned HTTP 404
[2026-07-27T00:44:27.868891] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/aaa3900d8f0c04012910ed8789d0c32d5e630e4284342426169717c1d5f89a3a/720p.mp4 returned HTTP 404
[2026-07-27T00:44:27.905641] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/aaa3900d8f0c04012910ed8789d0c32d5e630e4284342426169717c1d5f89a3a returned HTTP 404
[2026-07-27T00:44:34.561709] [WARNING] [VIDEO] (VideoEventService) Default video system is disabled - users will see real content only
[2026-07-27T00:46:54.950841] [WARNING] [VIDEO] (VideoEventService) Default video system is disabled - users will see real content only
[2026-07-27T00:47:24.689815] [ERROR] [VIDEO] (VideoEventService) ⏰ TIMEOUT: No events received for SubscriptionType.profile after 30 seconds!
[2026-07-27T00:47:24.689906] [WARNING] [VIDEO] (VideoEventService) ⏰ FEED TIMEOUT for profile:
=== FEED LOADING TIMEOUT DIAGNOSTIC ===
Subscription Type: profile
Timeout Duration: 30009ms
Relay Connected: true
Network Online: true
Connected Relay Count: 1

Filters:
  Filter 0:
    Kinds: [34236]
    Authors: 1 authors
    First author: cd4ce30f980c960757b46179608a4946fc06cad47f6dc8960f638e41312c1643
    Tags: 0 tags
    Since: null
    Until: null
    Limit: 50
  Filter 1:
    Kinds: [16]
    Au... [truncated]
[2026-07-27T00:47:43.005705] [DEBUG] [UI] (PageLoadObserver) Page pop tracked: monetization_links_settings
[2026-07-27T00:47:43.890124] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/creator-analytics, segments=1, queryKeys=[]
[2026-07-27T00:47:43.890130] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/creator-analytics, segments=1, queryKeys=[]
[2026-07-27T00:47:43.890151] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/creator-analytics, authState=authenticated
[2026-07-27T00:47:43.890194] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/creator-analytics, segments=1, queryKeys=[]
[2026-07-27T00:47:43.898168] [INFO] (ScreenAnalytics) 📱 Screen load started: creator_analytics
[2026-07-27T00:47:43.898289] [INFO] (ScreenAnalytics) ⏱️  User spent 2s on monetization_links_settings
[2026-07-27T00:47:43.898351] [INFO] (ScreenAnalytics) 👁️  Screen viewed: creator_analytics
[2026-07-27T00:47:43.898358] [DEBUG] [UI] (PageLoadObserver) Page push tracked: creator_analytics
[2026-07-27T00:47:43.903213] [INFO] (ScreenAnalytics) ✅ Screen content visible: creator_analytics in 5ms
[2026-07-27T00:47:43.903281] [INFO] (PagePerf) PERF: creator_analytics — visible: 5ms
[2026-07-27T00:47:49.044033] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/profile/*, segments=2, queryKeys=[]
[2026-07-27T00:47:49.044191] [DEBUG] [UI] (PageLoadObserver) Page pop tracked: creator_analytics
[2026-07-27T00:47:55.155302] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/general-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:55.155314] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/general-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:55.155335] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/general-settings, authState=authenticated
[2026-07-27T00:47:55.155393] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/general-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:55.166672] [INFO] (ScreenAnalytics) 📱 Screen load started: general_settings
[2026-07-27T00:47:55.166822] [INFO] (ScreenAnalytics) ⏱️  User spent 11s on creator_analytics
[2026-07-27T00:47:55.166907] [INFO] (ScreenAnalytics) 👁️  Screen viewed: general_settings
[2026-07-27T00:47:55.166910] [DEBUG] [UI] (PageLoadObserver) Page push tracked: general_settings
[2026-07-27T00:47:55.189511] [INFO] (ScreenAnalytics) ✅ Screen content visible: general_settings in 22ms
[2026-07-27T00:47:55.189589] [INFO] (PagePerf) PERF: general_settings — visible: 22ms
[2026-07-27T00:47:55.232540] [INFO] [SYSTEM] (MemoryTelemetry) Memory: rss 807.4 MB (peak 1051.9 MB), vc_native=1, ingest_queue_depth=0
[2026-07-27T00:47:56.721685] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/bluesky-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:56.721696] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/bluesky-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:56.721716] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/bluesky-settings, authState=authenticated
[2026-07-27T00:47:56.721836] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/bluesky-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:56.724956] [INFO] (ScreenAnalytics) 📱 Screen load started: bluesky_settings
[2026-07-27T00:47:56.725092] [INFO] (ScreenAnalytics) ⏱️  User spent 1s on general_settings
[2026-07-27T00:47:56.725149] [INFO] (ScreenAnalytics) 👁️  Screen viewed: bluesky_settings
[2026-07-27T00:47:56.725151] [DEBUG] [UI] (PageLoadObserver) Page push tracked: bluesky_settings
[2026-07-27T00:47:56.729343] [INFO] (ScreenAnalytics) ✅ Screen content visible: bluesky_settings in 4ms
[2026-07-27T00:47:56.729411] [INFO] (PagePerf) PERF: bluesky_settings — visible: 4ms
[2026-07-27T00:47:59.037311] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/profile/*, segments=2, queryKeys=[]
[2026-07-27T00:47:59.037400] [DEBUG] [UI] (PageLoadObserver) Page pop tracked: bluesky_settings
[2026-07-27T00:48:00.045385] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/profile/*, segments=2, queryKeys=[]
[2026-07-27T00:48:00.045462] [DEBUG] [UI] (PageLoadObserver) Page pop tracked: general_settings
[2026-07-27T00:48:01.564093] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/support-center, segments=1, queryKeys=[]
[2026-07-27T00:48:01.564102] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/support-center, segments=1, queryKeys=[]
[2026-07-27T00:48:01.564121] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/support-center, authState=authenticated
[2026-07-27T00:48:01.564158] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/support-center, segments=1, queryKeys=[]
[2026-07-27T00:48:01.566441] [INFO] (ScreenAnalytics) 📱 Screen load started: support_center
[2026-07-27T00:48:01.566563] [INFO] (ScreenAnalytics) ⏱️  User spent 4s on bluesky_settings
[2026-07-27T00:48:01.566606] [INFO] (ScreenAnalytics) 👁️  Screen viewed: support_center
[2026-07-27T00:48:01.566608] [DEBUG] [UI] (PageLoadObserver) Page push tracked: support_center
[2026-07-27T00:48:01.577135] [INFO] (ScreenAnalytics) ✅ Screen content visible: support_center in 10ms
[2026-07-27T00:48:01.577203] [INFO] (PagePerf) PERF: support_center — visible: 10ms
[2026-07-27T00:48:25.227936] [INFO] [SYSTEM] (MemoryTelemetry) Memory: rss 686.4 MB (peak 1051.9 MB), vc_native=1, ingest_queue_depth=0
[2026-07-27T00:48:53.857136] [INFO] [SYSTEM] Collecting bug report diagnostics
```

### Environment

**Platform:** 
**Build Number:** 
**OS Version:** 

### Actual Result

Kirsten Lynn Swasey, Jul 27, 2026, 18:49

follows category

it's not showing my new followers . in the category .

App Version: 1.0.17+668

### Device Information
- **platform:** android
- **model:** moto g power - 2025
- **manufacturer:** motorola
- **version:** 16
- **sdkInt:** 36
- **brand:** motorola

**Current Screen:** SupportCenterScreen
**User Pubkey:** cd4ce30f980c960757b46179608a4946fc06cad47f6dc8960f638e41312c1643

### Recent Logs (errors/warnings + recent context)
```
[2026-07-27T00:44:24.689859] [ERROR] (CrashReporting) Failed to initialize crash reporting: [core/duplicate-app] A Firebase App named "[DEFAULT]" already exists
[2026-07-27T00:44:26.255488] [WARNING] [SYSTEM] Zendesk JWT: SDK not initialized
[2026-07-27T00:44:26.255739] [WARNING] [SYSTEM] (ZendeskIdentitySync) Zendesk JWT upgrade failed, anonymous identity active
[2026-07-27T00:44:27.046697] [WARNING] [SYSTEM] Zendesk JWT: SDK not initialized
[2026-07-27T00:44:27.046699] [WARNING] [SYSTEM] (ZendeskIdentitySync) Zendesk JWT upgrade failed, anonymous identity active
[2026-07-27T00:44:27.635352] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/4b599880bcb3995c8fca5be81758a18fb87c57d265d47b14d6c232e92401be7c/720p.mp4 returned HTTP 404
[2026-07-27T00:44:27.679682] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/4b599880bcb3995c8fca5be81758a18fb87c57d265d47b14d6c232e92401be7c returned HTTP 404
[2026-07-27T00:44:27.723576] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/85cc1cb3bfe8f75b000133515656e68237d8dfc36f471703156d058f52ae94fa/720p.mp4 returned HTTP 404
[2026-07-27T00:44:27.782448] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/85cc1cb3bfe8f75b000133515656e68237d8dfc36f471703156d058f52ae94fa returned HTTP 404
[2026-07-27T00:44:27.868891] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/aaa3900d8f0c04012910ed8789d0c32d5e630e4284342426169717c1d5f89a3a/720p.mp4 returned HTTP 404
[2026-07-27T00:44:27.905641] [WARNING] [VIDEO] (MediaCache) CancellableDownload: https://media.divine.video/aaa3900d8f0c04012910ed8789d0c32d5e630e4284342426169717c1d5f89a3a returned HTTP 404
[2026-07-27T00:44:34.561709] [WARNING] [VIDEO] (VideoEventService) Default video system is disabled - users will see real content only
[2026-07-27T00:46:54.950841] [WARNING] [VIDEO] (VideoEventService) Default video system is disabled - users will see real content only
[2026-07-27T00:47:24.689815] [ERROR] [VIDEO] (VideoEventService) ⏰ TIMEOUT: No events received for SubscriptionType.profile after 30 seconds!
[2026-07-27T00:47:24.689906] [WARNING] [VIDEO] (VideoEventService) ⏰ FEED TIMEOUT for profile:
=== FEED LOADING TIMEOUT DIAGNOSTIC ===
Subscription Type: profile
Timeout Duration: 30009ms
Relay Connected: true
Network Online: true
Connected Relay Count: 1

Filters:
  Filter 0:
    Kinds: [34236]
    Authors: 1 authors
    First author: cd4ce30f980c960757b46179608a4946fc06cad47f6dc8960f638e41312c1643
    Tags: 0 tags
    Since: null
    Until: null
    Limit: 50
  Filter 1:
    Kinds: [16]
    Au... [truncated]
[2026-07-27T00:47:43.005705] [DEBUG] [UI] (PageLoadObserver) Page pop tracked: monetization_links_settings
[2026-07-27T00:47:43.890124] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/creator-analytics, segments=1, queryKeys=[]
[2026-07-27T00:47:43.890130] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/creator-analytics, segments=1, queryKeys=[]
[2026-07-27T00:47:43.890151] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/creator-analytics, authState=authenticated
[2026-07-27T00:47:43.890194] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/creator-analytics, segments=1, queryKeys=[]
[2026-07-27T00:47:43.898168] [INFO] (ScreenAnalytics) 📱 Screen load started: creator_analytics
[2026-07-27T00:47:43.898289] [INFO] (ScreenAnalytics) ⏱️  User spent 2s on monetization_links_settings
[2026-07-27T00:47:43.898351] [INFO] (ScreenAnalytics) 👁️  Screen viewed: creator_analytics
[2026-07-27T00:47:43.898358] [DEBUG] [UI] (PageLoadObserver) Page push tracked: creator_analytics
[2026-07-27T00:47:43.903213] [INFO] (ScreenAnalytics) ✅ Screen content visible: creator_analytics in 5ms
[2026-07-27T00:47:43.903281] [INFO] (PagePerf) PERF: creator_analytics — visible: 5ms
[2026-07-27T00:47:49.044033] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/profile/*, segments=2, queryKeys=[]
[2026-07-27T00:47:49.044191] [DEBUG] [UI] (PageLoadObserver) Page pop tracked: creator_analytics
[2026-07-27T00:47:55.155302] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/general-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:55.155314] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/general-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:55.155335] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/general-settings, authState=authenticated
[2026-07-27T00:47:55.155393] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/general-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:55.166672] [INFO] (ScreenAnalytics) 📱 Screen load started: general_settings
[2026-07-27T00:47:55.166822] [INFO] (ScreenAnalytics) ⏱️  User spent 11s on creator_analytics
[2026-07-27T00:47:55.166907] [INFO] (ScreenAnalytics) 👁️  Screen viewed: general_settings
[2026-07-27T00:47:55.166910] [DEBUG] [UI] (PageLoadObserver) Page push tracked: general_settings
[2026-07-27T00:47:55.189511] [INFO] (ScreenAnalytics) ✅ Screen content visible: general_settings in 22ms
[2026-07-27T00:47:55.189589] [INFO] (PagePerf) PERF: general_settings — visible: 22ms
[2026-07-27T00:47:55.232540] [INFO] [SYSTEM] (MemoryTelemetry) Memory: rss 807.4 MB (peak 1051.9 MB), vc_native=1, ingest_queue_depth=0
[2026-07-27T00:47:56.721685] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/bluesky-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:56.721696] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/bluesky-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:56.721716] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/bluesky-settings, authState=authenticated
[2026-07-27T00:47:56.721836] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/bluesky-settings, segments=1, queryKeys=[]
[2026-07-27T00:47:56.724956] [INFO] (ScreenAnalytics) 📱 Screen load started: bluesky_settings
[2026-07-27T00:47:56.725092] [INFO] (ScreenAnalytics) ⏱️  User spent 1s on general_settings
[2026-07-27T00:47:56.725149] [INFO] (ScreenAnalytics) 👁️  Screen viewed: bluesky_settings
[2026-07-27T00:47:56.725151] [DEBUG] [UI] (PageLoadObserver) Page push tracked: bluesky_settings
[2026-07-27T00:47:56.729343] [INFO] (ScreenAnalytics) ✅ Screen content visible: bluesky_settings in 4ms
[2026-07-27T00:47:56.729411] [INFO] (PagePerf) PERF: bluesky_settings — visible: 4ms
[2026-07-27T00:47:59.037311] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/profile/*, segments=2, queryKeys=[]
[2026-07-27T00:47:59.037400] [DEBUG] [UI] (PageLoadObserver) Page pop tracked: bluesky_settings
[2026-07-27T00:48:00.045385] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/profile/*, segments=2, queryKeys=[]
[2026-07-27T00:48:00.045462] [DEBUG] [UI] (PageLoadObserver) Page pop tracked: general_settings
[2026-07-27T00:48:01.564093] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/support-center, segments=1, queryKeys=[]
[2026-07-27T00:48:01.564102] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/support-center, segments=1, queryKeys=[]
[2026-07-27T00:48:01.564121] [DEBUG] [AUTH] (AppRouter) Router redirect: location=/support-center, authState=authenticated
[2026-07-27T00:48:01.564158] [DEBUG] [UI] (DeepLinkService) Skipping internal app route during deep-link parse: scheme=(none), host=(none), route=/support-center, segments=1, queryKeys=[]
[2026-07-27T00:48:01.566441] [INFO] (ScreenAnalytics) 📱 Screen load started: support_center
[2026-07-27T00:48:01.566563] [INFO] (ScreenAnalytics) ⏱️  User spent 4s on bluesky_settings
[2026-07-27T00:48:01.566606] [INFO] (ScreenAnalytics) 👁️  Screen viewed: support_center
[2026-07-27T00:48:01.566608] [DEBUG] [UI] (PageLoadObserver) Page push tracked: support_center
[2026-07-27T00:48:01.577135] [INFO] (ScreenAnalytics) ✅ Screen content visible: support_center in 10ms
[2026-07-27T00:48:01.577203] [INFO] (PagePerf) PERF: support_center — visible: 10ms
[2026-07-27T00:48:25.227936] [INFO] [SYSTEM] (MemoryTelemetry) Memory: rss 686.4 MB (peak 1051.9 MB), vc_native=1, ingest_queue_depth=0
[2026-07-27T00:48:53.857136] [INFO] [SYSTEM] Collecting bug report diagnostics
```

### Expected Result

(Generated from Zendesk Ticket: rabblelabs.zendesk.com/agent/tickets/4292)

NotThatKindOfDrLiz:
## Summary

- Fetch notification tabs as independent repository feeds with server-side `types` filters instead of widget-layer filtering over one mixed page.
- Give each filtered feed its own cursor/snapshot while keeping read-state updates shared across live feeds.
- Wire inbox tabs to per-filter blocs, remove presentation-layer filtering, and keep empty filtered tabs paginating when the server reports more rows.
- Consolidate follow notifications across appended pages so the same follower does not appear twice after pagination.

Closes #6428

## Tests

- `flutter analyze`
- `flutter test test/src/funnelcake_api_client_notification_test.dart`
- `flutter test test/src/notification_repository_test.dart`
- `flutter test test/notifications/bloc/notification_feed_bloc_test.dart test/notifications/view/notifications_view_test.dart test/notifications/view/inbox_notifications_page_test.dart test/notifications/view/notifications_page_test.dart test/notifications/badge_converges_after_tap_read_back_test.dart`
- Pre-push hook analyzer and changed-file tests

### DIFF
--- a/mobile/lib/notifications/bloc/notification_feed_bloc.dart
+++ b/mobile/lib/notifications/bloc/notification_feed_bloc.dart
@@ -28,8 +28,10 @@ class NotificationFeedBloc
   NotificationFeedBloc({
     required NotificationRepository notificationRepository,
     required FollowRepository followRepository,
+    NotificationKind? filter,
   }) : _notificationRepository = notificationRepository,
        _followRepository = followRepository,
+       _filter = filter,
        super(const NotificationFeedState()) {
     on<_SnapshotChanged>(_onSnapshotChanged);
     on<NotificationFeedStarted>(_onStarted, transformer: droppable());
@@ -40,13 +42,14 @@ class NotificationFeedBloc
 
     // Project the repository's snapshot stream into BLoC state via a
     // private event so emit() stays inside an event handler.
-    _snapshotSubscription = _notificationRepository.watchSnapshot().listen(
-      (page) => add(_SnapshotChanged(page)),
-    );
+    _snapshotSubscription = _notificationRepository
+        .watchSnapshot(filter: _filter)
+        .listen((page) => add(_SnapshotChanged(page)));
   }
 
   final NotificationRepository _notificationRepository;
   final FollowRepository _followRepository;
+  final NotificationKind? _filter;
   late final StreamSubscription<NotificationPage> _snapshotSubscription;
 
   /// Override [ActorNotification.isFollowingBack] for follow-type rows so the
@@ -104,14 +107,16 @@ class NotificationFeedBloc
     emit(state.copyWith(isRefreshing: true));
 
     try {
-      await _notificationRepository.refresh();
+      await _notificationRepository.refreshFeed(_filter);
       emit(
         state.copyWith(
           status: NotificationFeedStatus.loaded,
           isRefreshing: false,
         ),
       );
-      await _markSeenOnOpen();
+      if (_filter == null) {
+        await _markSeenOnOpen();
+      }
     } on Exception catch (e, s) {
       // `NotificationRepository.refresh` propagates typed
       // `FunnelcakeException` (4xx/5xx/timeout; transient-retry
@@ -206,7 +211,7 @@ class NotificationFeedBloc
     emit(state.copyWith(isLoadingMore: true));
 
     try {
-      await _notificationRepository.loadNextPage();
+      await _notificationRepository.loadNextPageFor(_filter);
       emit(state.copyWith(isLoadingMore: false));
     } on Exception catch (e, s) {
       // `NotificationRepository.loadNextPage` (single-attempt
@@ -232,7 +237,7 @@ class NotificationFeedBloc
   ) async {
     emit(state.copyWith(isRefreshing: true));
     try {
-      await _notificationRepository.refresh();
+      await _notificationRepository.refreshFeed(_filter);
       emit(
         state.copyWith(
           status: NotificationFeedStatus.loaded,
@@ -342,7 +347,9 @@ class NotificationFeedBloc
   @override
   Future<void> close() async {
     await _snapshotSubscription.cancel();
-    _notificationRepository.resetPaginationDepth();
+    if (_filter == null) {
+      _notificationRepository.resetPaginationDepth();
+    }
     return super.close();
   }
 }
--- a/mobile/lib/notifications/bloc/notification_feed_bloc.dart
+++ b/mobile/lib/notifications/bloc/notification_feed_bloc.dart
@@ -53,6 +53,7 @@ class NotificationFeedBloc
   late final StreamSubscription<NotificationPage> _snapshotSubscription;
   int _emptyPageContinuations = 0;
   bool _loadMoreFailed = false;
+  bool _emptyPageContinuationPending = false;
 
   static const _maxEmptyPageContinuations = 3;
 
@@ -89,21 +90,45 @@ class NotificationFeedBloc
     _continuePastEmptyPageIfNeeded(notifications);
   }
 
+  /// Keeps an all-filtered-out page paginating instead of rendering as an
+  /// empty tab while the server still reports more rows.
+  ///
+  /// The repository emits the snapshot from inside `refreshFeed`, before it
+  /// completes, so on the first open this runs while `_onStarted` still
+  /// holds `isRefreshing: true` and `status: initial`. Dispatching a
+  /// load-more there would race the refresh's cursor reset, but simply
+  /// bailing strands the tab: no further snapshot follows, so it settles
+  /// empty with `hasMore: true` and never paginates. Defer instead, and
+  /// let [_flushPendingEmptyPageContinuation] pick it up once the refresh
+  /// handler has settled.
   void _continuePastEmptyPageIfNeeded(List<NotificationItem> notifications) {
     if (notifications.isNotEmpty || !state.hasMore) {
       _emptyPageContinuations = 0;
       _loadMoreFailed = false;
+      _emptyPageContinuationPending = false;
+      return;
+    }
+    if (_loadMoreFailed ||
+        _emptyPageContinuations >= _maxEmptyPageContinuations) {
       return;
     }
     if (state.status != NotificationFeedStatus.loaded ||
         state.isLoadingMore ||
-        state.isRefreshing ||
-        _loadMoreFailed ||
-        _emptyPageContinuations >= _maxEmptyPageContinuations) {
+        state.isRefreshing) {
+      _emptyPageContinuationPending = true;
       return;
     }
+    _emptyPageContinuationPending = false;
     _emptyPageContinuations += 1;
     add(const NotificationFeedLoadMore());
+  }
+
+  /// Re-runs a continuation that [_continuePastEmptyPageIfNeeded] had to
+  /// defer because a refresh was still in flight when the empty page landed.
+  void _flushPendingEmptyPageContinuation() {
+    if (!_emptyPageContinuationPending) return;
+    _emptyPageContinuationPending = false;
+    _continuePastEmptyPageIfNeeded(state.notifications);
   }
 
   /// Handle initial load.
@@ -126,6 +151,7 @@ class NotificationFeedBloc
   ) async {
     _emptyPageContinuations = 0;
     _loadMoreFailed = false;
+    _emptyPageContinuationPending = false;
     // Stale-while-revalidate: keep any hydrated/cached items rendered and
     // flag the in-flight refresh so the view shows a thin progress bar
     // instead of a blanking full-screen spinner.
@@ -139,6 +165,7 @@ class NotificationFeedBloc
           isRefreshing: false,
         ),
       );
+      _flushPendingEmptyPageContinuation();
       if (_filter == null) {
         await _markSeenOnOpen();
       }
@@ -264,6 +291,7 @@ class NotificationFeedBloc
   ) async {
     _emptyPageContinuations = 0;
     _loadMoreFailed = false;
+    _emptyPageContinuationPending = false;
     emit(state.copyWith(isRefreshing: true));
     try {
       await _notificationRepository.refreshFeed(_filter);
@@ -273,6 +301,7 @@ class NotificationFeedBloc
           isRefreshing: false,
         ),
       );
+      _flushPendingEmptyPageContinuation();
     } on Exception catch (e, s) {
       // `NotificationRepository.refresh` propagates typed
       // `FunnelcakeException` (4xx/5xx/timeout). Per
--- a/mobile/lib/notifications/bloc/notification_feed_bloc.dart
+++ b/mobile/lib/notifications/bloc/notification_feed_bloc.dart
@@ -254,11 +254,18 @@ class NotificationFeedBloc
   }
 
   /// Handle scroll pagination.
+  ///
+  /// Deliberately does NOT consult [_loadMoreFailed]: that flag only stops
+  /// the automatic empty-page continuation from re-arming. A load-more
+  /// failure emits no snapshot (`_markRefreshError` is first-page-only) and
+  /// surfaces no error affordance, so gating this handler on it would leave
+  /// a user who hit one transient 5xx at the bottom of the list unable to
+  /// paginate for the rest of the session.
   Future<void> _onLoadMore(
     NotificationFeedLoadMore event,
     Emitter<NotificationFeedState> emit,
   ) async {
-    if (!state.hasMore || state.isLoadingMore || _loadMoreFailed) return;
+    if (!state.hasMore || state.isLoadingMore) return;
 
     emit(state.copyWith(isLoadingMore: true));
 
--- a/mobile/lib/notifications/bloc/notification_feed_bloc.dart
+++ b/mobile/lib/notifications/bloc/notification_feed_bloc.dart
@@ -51,6 +51,10 @@ class NotificationFeedBloc
   final FollowRepository _followRepository;
   final NotificationKind? _filter;
   late final StreamSubscription<NotificationPage> _snapshotSubscription;
+  int _emptyPageContinuations = 0;
+  bool _loadMoreFailed = false;
+
+  static const _maxEmptyPageContinuations = 3;
 
   /// Override [ActorNotification.isFollowingBack] for follow-type rows so the
   /// flag tracks the authoritative follow state in [FollowRepository] rather
@@ -73,14 +77,33 @@ class NotificationFeedBloc
     _SnapshotChanged event,
     Emitter<NotificationFeedState> emit,
   ) {
+    final notifications = _applyFollowState(event.page.items);
     emit(
       state.copyWith(
-        notifications: _applyFollowState(event.page.items),
+        notifications: notifications,
         unreadCount: event.page.unreadCount,
         hasMore: event.page.hasMore,
         refreshError: event.page.lastRefreshError,
       ),
     );
+    _continuePastEmptyPageIfNeeded(notifications);
+  }
+
+  void _continuePastEmptyPageIfNeeded(List<NotificationItem> notifications) {
+    if (notifications.isNotEmpty || !state.hasMore) {
+      _emptyPageContinuations = 0;
+      _loadMoreFailed = false;
+      return;
+    }
+    if (state.status != NotificationFeedStatus.loaded ||
+        state.isLoadingMore ||
+        state.isRefreshing ||
+        _loadMoreFailed ||
+        _emptyPageContinuations >= _maxEmptyPageContinuations) {
+      return;
+    }
+    _emptyPageContinuations += 1;
+    add(const NotificationFeedLoadMore());
   }
 
   /// Handle initial load.
@@ -101,6 +124,8 @@ class NotificationFeedBloc
     NotificationFeedStarted event,
     Emitter<NotificationFeedState> emit,
   ) async {
+    _emptyPageContinuations = 0;
+    _loadMoreFailed = false;
     // Stale-while-revalidate: keep any hydrated/cached items rendered and
     // flag the in-flight refresh so the view shows a thin progress bar
     // instead of a blanking full-screen spinner.
@@ -206,7 +231,7 @@ class NotificationFeedBloc
     NotificationFeedLoadMore event,
     Emitter<NotificationFeedState> emit,
   ) async {
-    if (!state.hasMore || state.isLoadingMore) return;
+    if (!state.hasMore || state.isLoadingMore || _loadMoreFailed) return;
 
     emit(state.copyWith(isLoadingMore: true));
 
@@ -214,6 +239,7 @@ class NotificationFeedBloc
       await _notificationRepository.loadNextPageFor(_filter);
       emit(state.copyWith(isLoadingMore: false));
     } on Exception catch (e, s) {
+      _loadMoreFailed = true;
       // `NotificationRepository.loadNextPage` (single-attempt
       // paginate-load-more) propagates typed `FunnelcakeException`
       // (4xx/5xx/timeout). Per .claude/rules/error_handling.md they
@@ -221,6 +247,7 @@ class NotificationFeedBloc
       addError(e, s);
       emit(state.copyWith(isLoadingMore: false));
     } catch (e, s) {
+      _loadMoreFailed = true;
       // Errors (StateError, TypeError) — matrix-YES invariant.
       addError(
         Reportable(e, context: NotificationFeedBlocReportableSites.onLoadMore),
@@ -235,6 +262,8 @@ class NotificationFeedBloc
     NotificationFeedRefreshed event,
     Emitter<NotificationFeedState> emit,
   ) async {
+    _emptyPageContinuations = 0;
+    _loadMoreFailed = false;
     emit(state.copyWith(isRefreshing: true));
     try {
       await _notificationRepository.refreshFeed(_filter);
@@ -347,9 +376,7 @@ class NotificationFeedBloc
   @override
   Future<void> close() async {
     await _snapshotSubscription.cancel();
-    if (_filter == null) {
-      _notificationRepository.resetPaginationDepth();
-    }
+    _notificationRepository.resetPaginationDepth(filter: _filter);
     return super.close();
   }
 }
--- a/mobile/lib/notifications/bloc/notification_feed_bloc.dart
+++ b/mobile/lib/notifications/bloc/notification_feed_bloc.dart
@@ -272,6 +272,13 @@ class NotificationFeedBloc
     try {
       await _notificationRepository.loadNextPageFor(_filter);
       emit(state.copyWith(isLoadingMore: false));
+      // The page this call fetched lands on the snapshot while
+      // `isLoadingMore` is still true, so a page that filters down to
+      // nothing parks its continuation instead of dispatching it. Flush it
+      // here — otherwise only the first of [_maxEmptyPageContinuations]
+      // could ever fire, and a Comments feed whose pages are mostly video
+      // mentions would get one extra page of runway instead of three.
+      _flushPendingEmptyPageContinuation();
     } on Exception catch (e, s) {
       _loadMoreFailed = true;
       // `NotificationRepository.loadNextPage` (single-attempt
--- a/mobile/packages/notification_repository/lib/src/notification_repository.dart
+++ b/mobile/packages/notification_repository/lib/src/notification_repository.dart
@@ -173,9 +173,14 @@ class NotificationRepository {
   final AuthHeadersProvider? _authHeadersProvider;
 
   final Map<NotificationKind?, _NotificationFeed> _feeds = {};
+  bool _closed = false;
 
-  _NotificationFeed _feedFor(NotificationKind? filter) =>
-      _feeds.putIfAbsent(filter, () => _NotificationFeed(filter: filter));
+  _NotificationFeed _feedFor(NotificationKind? filter) {
+    if (_closed) {
+      throw StateError('NotificationRepository is closed');
+    }
+    return _feeds.putIfAbsent(filter, () => _NotificationFeed(filter: filter));
+  }
 
   _NotificationFeed get _allFeed => _feedFor(null);
 
@@ -214,7 +219,9 @@ class NotificationRepository {
   /// Called when the repository is no longer needed (e.g. on auth flip
   /// when a new repository instance replaces this one).
   Future<void> close() async {
-    for (final feed in _liveFeeds) {
+    if (_closed) return;
+    _closed = true;
+    for (final feed in _liveFeeds.toList()) {
       await feed.snapshot.close();
     }
   }
@@ -225,7 +232,7 @@ class NotificationRepository {
   /// long-lived consumer (e.g. the refresh coordinator) may still hold it
   /// across an in-flight call. Such callers use this to classify the
   /// resulting [StateError] as expected account-switch noise.
-  bool get isClosed => _snapshot.isClosed;
+  bool get isClosed => _closed;
 
   /// Whether the snapshot currently holds more than the first page.
   ///
@@ -233,7 +240,8 @@ class NotificationRepository {
   /// user-visible paginated feed back to page 1. An explicit [refresh]
   /// (pull-to-refresh, page mount) still replaces the snapshot and resets
   /// this to `false`.
-  bool get hasPaginatedBeyondFirstPage => _allFeed.pagesLoaded > 1;
+  bool get hasPaginatedBeyondFirstPage =>
+      _liveFeeds.any((feed) => feed.pagesLoaded > 1);
 
   /// Releases the current page-depth guard after the feed UI is gone, and
   /// collapses any deep-scrolled accumulation back to the first page.
@@ -254,14 +262,15 @@ class NotificationRepository {
   /// reflects only the newest page's unread rows — the same bound the next
   /// `refresh()` would impose anyway, so this only brings that bound forward
   /// to feed-close instead of next-open.
-  void resetPaginationDepth() {
-    final current = _snapshot.value;
+  void resetPaginationDepth({NotificationKind? filter}) {
+    final feed = _feedFor(filter);
+    final current = feed.snapshot.value;
     if (current.items.length > _pageSize) {
-      _snapshot.add(
+      feed.snapshot.add(
         current.copyWith(items: current.items.take(_pageSize).toList()),
       );
     }
-    _allFeed.pagesLoaded = _snapshot.value.items.isEmpty ? 0 : 1;
+    feed.pagesLoaded = feed.snapshot.value.items.isEmpty ? 0 : 1;
   }
 
   /// Fetches the next page of notifications.
@@ -326,7 +335,10 @@ class NotificationRepository {
         ..lastCursor = response.nextCursor
         ..lastCursorId = response.nextCursorId;
 
-      final items = await _enrichAndGroup(response.notifications);
+      final items = await _enrichAndGroup(
+        response.notifications,
+        filter: filter,
+      );
       if (generation != feed.fetchGeneration) {
         return (page: feed.snapshot.value, applied: false);
       }
@@ -802,7 +814,7 @@ class NotificationRepository {
         NotificationKind.mention => const ['mention'],
         NotificationKind.likeComment => const ['reaction', 'zap'],
         NotificationKind.reply => const ['reply'],
-        NotificationKind.system => null,
+        NotificationKind.system => const ['list_add'],
       };
 
   static final math.Random _jitter = math.Random();
@@ -822,11 +834,25 @@ class NotificationRepository {
   /// when the refresh was superseded by another first-page fetch before it
   /// could apply.
   Future<bool> refreshApplied() {
-    return _refreshResult().then((result) => result.applied);
+    return _refreshAppliedForLiveFeeds();
   }
 
   Future<({NotificationPage page, bool applied})> _refreshResult() {
     return _refreshResultFor(null);
+  }
+
+  Future<bool> _refreshAppliedForLiveFeeds() async {
+    final feeds = _liveFeeds.toList();
+    if (feeds.isEmpty) {
+      return _refreshResult().then((result) => result.applied);
+    }
+
+    var applied = false;
+    for (final feed in feeds) {
+      final result = await _refreshResultFor(feed.filter);
+      applied = applied || result.applied;
+    }
+    return applied;
   }
 
   Future<({NotificationPage page, bool applied})> _refreshResultFor(
@@ -1239,8 +1265,9 @@ class NotificationRepository {
   /// Enriches raw relay notifications with profile + video metadata, then
   /// groups them into [VideoNotification]s and [ActorNotification]s.
   Future<List<NotificationItem>> _enrichAndGroup(
-    List<RelayNotification> raw,
-  ) async {
+    List<RelayNotification> raw, {
+    NotificationKind? filter,
+  }) async {
     if (raw.isEmpty) return [];
 
     final pubkeys = raw.map((n) => n.sourcePubkey).toSet().toList();
@@ -1266,7 +1293,25 @@ class NotificationRepository {
 
     final items = <NotificationItem>[...videos, ...actors, ...reclassified]
       ..sort((a, b) => b.timestamp.compareTo(a.timestamp));
-    return _applyBlockFilter(items);
+    return _applyBlockFilter(_applyFeedFilter(items, filter));
+  }
+
+  List<NotificationItem> _applyFeedFilter(
+    List<NotificationItem> items,
+    NotificationKind? filter,
+  ) {
+    if (filter != NotificationKind.comment) return items;
+    return items.where(_belongsInCommentsFeed).toList();
+  }
+
+  bool _belongsInCommentsFeed(NotificationItem item) {
+    if (item.type == NotificationKind.comment ||
+        item.type == NotificationKind.reply) {
+      return true;
+    }
+    return item is ActorNotification &&
+        item.type == NotificationKind.mention &&
+        item.hasCommentTarget;
   }
 
   /// Filters notifications from blocked/muted users.
--- a/mobile/packages/notification_repository/lib/src/notification_repository.dart
+++ b/mobile/packages/notification_repository/lib/src/notification_repository.dart
@@ -1038,7 +1038,8 @@ class NotificationRepository {
       }
     }
 
-    final appended = <NotificationItem>[];
+    // Appends land in `result` itself so the indices in `followIndex` stay
+    // valid for the `result[idx] = …` merges below.
     for (final item in incoming) {
       if (item is VideoNotification) {
         final idx = videoGroupIndex[(item.videoEventId, item.type)];
@@ -1070,14 +1071,14 @@ class NotificationRepository {
       // omitted source_event_id). Preserves the original by-id dedupe
       // contract.
       if (allIds.contains(item.id)) continue;
-      appended.add(item);
+      result.add(item);
       allSourceEventIds.addAll(item.sourceEventIds);
       allIds.add(item.id);
       if (item is ActorNotification && item.type == NotificationKind.follow) {
-        followIndex[item.actor.pubkey] = result.length + appended.length - 1;
+        followIndex[item.actor.pubkey] = result.length - 1;
       }
     }
-    return [...result, ...appended];
+    return result;
   }
 
   static ActorNotification _mergeAppendedFollow(
--- a/mobile/packages/notification_repository/lib/src/notification_repository.dart
+++ b/mobile/packages/notification_repository/lib/src/notification_repository.dart
@@ -234,14 +234,17 @@ class NotificationRepository {
   /// resulting [StateError] as expected account-switch noise.
   bool get isClosed => _closed;
 
-  /// Whether the snapshot currently holds more than the first page.
+  /// Whether every live feed is paginated beyond its first page.
   ///
   /// Resume-driven liveness triggers consult this to avoid collapsing a
-  /// user-visible paginated feed back to page 1. An explicit [refresh]
-  /// (pull-to-refresh, page mount) still replaces the snapshot and resets
-  /// this to `false`.
+  /// user-visible paginated feed back to page 1. [refreshApplied] already
+  /// skips the individual feeds that would collapse, so this only reports
+  /// `true` when no live feed is refreshable at all — otherwise one tab
+  /// scrolled two pages deep would suppress resume refreshes for every
+  /// other tab as well. An explicit [refresh] (pull-to-refresh, page mount)
+  /// still replaces the snapshot and resets this to `false`.
   bool get hasPaginatedBeyondFirstPage =>
-      _liveFeeds.any((feed) => feed.pagesLoaded > 1);
+      _liveFeeds.isNotEmpty && _liveFeeds.every((feed) => feed.pagesLoaded > 1);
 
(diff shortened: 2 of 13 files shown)


hm21 (changes requested):
Pushed two commits — revert either independently:

- `f5aca00` — `_mergeAppendedPage` registered appended follows at `result.length + appended.length - 1`, an index into the concatenated return value, but the merge branch reads `result[idx]`, which never grows. Unreachable today only because `_consolidateFollows` already dedupes follows per page; appending into `result` directly makes the index valid either way.
- `8833fa8` — the tab fan-out test only asserted `refreshFeed(null)`, so transposing two entries in the `TabBarView` children list (Comments serving reposts) kept the suite green. Now asserted per tap. Also dropped the stale "they share one NotificationFeedBloc" comment.

Four things I'd rather you decide than patch myself:

**1. The Comments tab now shows every mention.** `types=comment,reply,mention`, but `mention` is funnelcake's catch-all — `notification_type_for_candidate` (`divine-funnelcake crates/clickhouse/src/client.rs:1149`) is `_ => "mention"`, so a kind 34236 video or kind 30023 article that p-tags you is stamped `mention` and lands under Comments. The deleted `_applyFilter` admitted a mention row only when `hasCommentTarget` was true, and the deleted test was literally named "comment filter includes comment mentions but not video mentions". Either narrow it in the repository for the `comment` feed, or make the new behaviour explicit and pin it.

**2. Every tab re-entry refetches page 1 and re-POSTs the seen watermark.** The `BlocProvider`s moved inside `TabBarView`, which disposes offscreen pages, so each visit builds a fresh bloc and re-dispatches `NotificationFeedStarted`. Probed All → Likes → All → Likes: 4 blocs created, `refreshFeed(null)` and `markAllAsRead()` twice each. `_markSeenOnOpen`'s own doc says it fires once per open. Hoisting the five providers above the `TabBarView`, or `AutomaticKeepAliveClientMixin` on the tab children, both fix it.

**3. Empty-tab auto-load-more has no cap and no backoff.** The reset in `build` clears `_emptyLoadMoreQueued` whenever `isLoadingMore` is true, and `_onLoadMore`'s catch arm leaves `hasMore` untouched — so a load-more that keeps failing (offline, 5xx) re-dispatches on every settle for as long as the tab is on screen, each iteration a full REST page plus the profile/video enrichment fan-out. It also dispatches from `build()` through a side-effecting `Builder`. This wants to live in the bloc or repository with a consecutive-empty-page cap and no re-dispatch after an error.

**4. Filtered feeds never trim their pagination depth.** `resetPaginationDepth()` still only touches `_allFeed`, and the bloc now gates it on `_filter == null`, so a tab scrolled N pages deep replays all N pages through `_applyFollowState` on re-entry before `refreshFeed` replaces them — the cost the doc block at `notification_repository.dart:246-256` says the trim exists to bound. A `filter` parameter plus dropping the gate covers it, but it interacts with (2), so worth deciding together.

Validation on `8833fa8`: `flutter analyze lib test integration_test`, `flutter test test/notifications/ test/blocs/notifications/`, `notification_repository` and `funnelcake_api_client` suites — all green; `notification_repository` line coverage 94.83% against the 94 gate.

One note on the branch as I found it: `notification_pages_repo_swap_test.dart` was red before your amend (`refresh` stubbed, `refreshFeed` called → MissingStubError → the generic catch in `_onStarted` set `refreshError: true`). Your amend fixed it; flagging only because it means the pre-push hook did not catch it.

hm21 (approved):
Approving — the four points in my previous review are non-blocking notes for you to take or leave, not merge blockers. CI is green on 8833fa8 (all 4 test shards, Analyze, Format, Generated Files, build).

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -757,6 +792,19 @@ class NotificationRepository {
     NotificationKind.system => 'system',
   };
 
+  static List<String>? _serverTypesForFilter(NotificationKind? filter) =>
+      switch (filter) {
+        null => null,
+        NotificationKind.like => const ['reaction', 'zap'],
+        NotificationKind.comment => const ['comment', 'reply', 'mention'],
```
realmeylisdev, line 881: Anchoring @hm21's point 1 here since it's the line that carries it.

`mention` is funnelcake's catch-all — `notification_type_for_candidate` (divine-funnelcake `crates/clickhouse/src/client.rs:1140`) ends in `_ => "mention"`, so any p-tagging event outside kinds 3/6/16/7/1111/9735/list_add gets stamped `mention`, including a kind 34236 video and a kind 30023 article. The deleted `_applyFilter` admitted a mention row only when `hasCommentTarget` was true, and the deleted test was named "comment filter includes comment mentions but not video mentions". So "X mentioned you in a video" now lands under Comments.

This is also the only tab mapping with no test — `follow` and `like` are pinned at `notification_repository_test.dart:352` and `:379`, `comment` and `repost` aren't. Whichever way you decide, worth pinning the one that changed behaviour.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -778,10 +826,23 @@ class NotificationRepository {
   }
 
   Future<({NotificationPage page, bool applied})> _refreshResult() {
-    _fetchGeneration++;
-    _lastCursor = null;
-    _lastCursorId = null;
-    return _getNotificationsResult();
+    return _refreshResultFor(null);
```
realmeylisdev, line 911: `refreshApplied()` routes here too, so `NotificationRefreshCoordinator._runRefresh` (`notification_refresh_coordinator.dart:91`) can only ever refresh the `null` feed. Before this PR all five tabs rendered that one snapshot, so a resume/foreground refresh updated whichever tab was on screen; now a user sitting on Likes or Follows gets a list that only moves on pull-to-refresh or a tab round-trip.

The coordinator's skip guard is All-only for the same reason — `hasPaginatedBeyondFirstPage` reads `_allFeed.pagesLoaded`, so it can't see how deep a filtered tab is scrolled. That one interacts with @hm21's point 4.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -200,7 +213,11 @@ class NotificationRepository {
   ///
   /// Called when the repository is no longer needed (e.g. on auth flip
   /// when a new repository instance replaces this one).
-  Future<void> close() => _snapshot.close();
+  Future<void> close() async {
+    for (final feed in _liveFeeds) {
```
realmeylisdev: `_liveFeeds` is `_feeds.values` — a live view — and the loop awaits inside it. Anything that calls `_feedFor` during one of those yields (a tab bloc constructing and hitting `watchSnapshot(filter:)`) mutates the map mid-iteration and this throws `ConcurrentModificationError`. `_feeds.values.toList()` closes it.

Related: `_feedFor` is a bare `putIfAbsent` with no closed-check, so a feed created after `close()` is a live subject nothing ever closes — while `isClosed` (line 228) reads only `_allFeed` and reports `true`.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -757,6 +792,19 @@ class NotificationRepository {
...
+      switch (filter) {
+        null => null,
+        NotificationKind.like => const ['reaction', 'zap'],
+        NotificationKind.comment => const ['comment', 'reply', 'mention'],
+        NotificationKind.repost => const ['repost'],
+        NotificationKind.follow => const ['follow'],
+        NotificationKind.mention => const ['mention'],
+        NotificationKind.likeComment => const ['reaction', 'zap'],
+        NotificationKind.reply => const ['reply'],
+        NotificationKind.system => null,
```
realmeylisdev: `null` is the "send no `types` param" sentinel four lines up, so a `system` feed would fetch the entire unfiltered inbox rather than nothing. Unreachable today — there's no System tab — but the arm reads as handled.

funnelcake has no `system` type to filter on (`is_allowed_notification_type`, `crates/api/src/handlers.rs:96`), so this probably wants to be explicit: `['list_add']` if that's the intent, otherwise something that fails loudly rather than silently widening.

realmeylisdev (commented):
Reviewed the diff plus the funnelcake side. The premise holds: `types` lands in the inner `WHERE` before `GROUP BY n.card_key` / `ORDER BY` / `LIMIT` (`crates/clickhouse/src/client.rs`, `get_notifications`), so the Follows tab gets a real page of follows and #6428 is fixed.

Four inline notes. The resume-refresh one and the `close()` one are new; the Comments-tab one re-anchors @hm21's point 1 inline since it is still open. I re-derived @hm21's points 2-4 and they hold.

NotThatKindOfDrLiz:
Takeover update pushed in `0bc5812548`.

Product decision for Comments tab: keep the server request broad enough to include comment-target mentions (`comment`, `reply`, `mention`), then narrow after mapping so the tab includes comments, nested replies, and comment-target mentions, but excludes video/general mentions.

Other review fixes included in the same pass:
- moved empty filtered-feed continuation into `NotificationFeedBloc` with a three-page cap and no retry loop after load-more failure; removed build-time load-more dispatch from `NotificationsView`
- kept visited tab blocs alive across `TabBarView` switches and made pagination-depth trimming work per filtered feed
- made foreground/resume refresh update live feeds, avoided `system` widening to the unfiltered feed, and hardened repository close handling

Verification:
- `flutter analyze`
- `flutter test test/notifications/bloc/notification_feed_bloc_test.dart test/notifications/view/inbox_notifications_page_test.dart test/notifications/view/notifications_view_test.dart test/notifications/badge_converges_after_tap_read_back_test.dart`
- `flutter test test/src/notification_repository_test.dart`
- `flutter test --coverage` from `mobile/packages/notification_repository` — 94.61% line coverage (790/835)
- `flutter test test/src/funnelcake_api_client_notification_test.dart`
- pre-push hook reran analyze plus changed-file tests successfully before push

NotThatKindOfDrLiz:
Follow-up pushed in `5d71f5556b` after CI exposed one close-order path: `NotificationFeedBloc.close()` can run after its `NotificationRepository` has already closed during provider teardown. `resetPaginationDepth(filter:)` now no-ops after repository close, with a regression test.

The failed shard-0 inbox test now passes locally, and the new CI run is green across all checks.

hm21 (approved):
Pushed four commits on top of `5d71f55` — revert any of them independently.

**`1b270de` — the empty-page continuation never fired on first open.** `_getNotificationsResult` emits the snapshot from inside `refreshFeed`, before it completes, so `_onSnapshotChanged` ran while `_onStarted` still held `isRefreshing: true` and `status: initial` — both bail conditions — and nothing emitted afterwards (`_markSeenOnOpen` is skipped for filtered feeds; `markAllAsRead` early-returns on an empty page). The tab settled at `status: loaded, hasMore: true`, empty, and never paginated. That is exactly the Comments case: a page of video mentions that `_belongsInCommentsFeed` empties. Verified 3/3 deterministic with a probe before fixing. The continuation is now deferred and flushed once the refresh handler settles. The existing test passed only because it seeded `status: loaded` and pushed the snapshot outside `_onStarted`; the new test reproduces the real ordering and is red without the fix.

**`e894d7d` — `_loadMoreFailed` also killed the user-initiated scroll path.** It gated `_onLoadMore` itself but is only cleared by `Started`, `Refreshed`, or a snapshot carrying items — and a load-more failure emits no snapshot (`_markRefreshError` is first-page-only) and sets no `refreshError`. So one transient 5xx at the bottom of the list silently stopped pagination for the rest of the session. The flag still guards the automatic continuation, which is all it was for; a failed load-more emits no snapshot, so `_continuePastEmptyPageIfNeeded` is not re-entered and could not loop regardless.

**`2af212f` — resume refresh, per feed instead of all-or-nothing.** Two parts, and the first is a design call rather than a bug — revert just this commit if you disagree. (a) `hasPaginatedBeyondFirstPage` was `any`, so with tab blocs kept alive, one tab scrolled two pages deep suppressed resume refreshes for the whole inbox until the page was disposed. `_refreshAppliedForLiveFeeds` now skips the feeds a first-page replace would collapse and refreshes the rest, and the guard only reports true when no live feed is refreshable at all. (b) The loop had no error isolation, and the unfiltered feed is created first — so one offline All refresh meant no tab refreshed. Each feed now refreshes independently; the first error is rethrown only when no feed applied, so a partial success still advances the coordinator's cooldown.

**`6aa6bf3`** — my own CI break: the bare `catch` in `2af212f` tripped `avoid_catches_without_on_clauses`, which the package lints enforce and the app's do not. `on Object catch` keeps the intended broad capture.

Validation on `6aa6bf3`: `flutter analyze lib test integration_test` from `mobile/` and `flutter analyze lib test` inside `notification_repository` both clean, `test/notifications/` + `test/blocs/notifications/` 200 green, package suite 156 green at 95.00% line coverage (gate 94). All checks green on this head.

Two corrections to my earlier review, for the record: `_serverTypesForFilter` is fine — my funnelcake checkout was 30 commits stale, and `origin/main`'s allowlist includes both `comment` and `list_add`. And the Comments-tab mapping is now properly pinned by the three tests you added.

Not touched, still yours to weigh: the Comments feed can return a mid-list page that the client-side filter empties, which appends nothing and leaves the user nudging the scroll to advance. The three-page continuation only covers a feed that is empty from the start.

Yours to review and merge when you're happy with it.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Addressed in `0bc5812548` — `_applyFeedFilter` / `_belongsInCommentsFeed` narrows `mention` back to `hasCommentTarget` after mapping, and the three new repository tests pin it. Resolving; the leftover "a mid-list page can still filter down to nothing" half is tracked separately in my new review.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Addressed in `2af212f01c` — `refreshApplied()` now fans out across live feeds instead of only the `null` one, so a user sitting on Likes or Follows gets a resume refresh. Resolving; I have a follow-up on the fan-out being unbounded in time, filed inline in the new review.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Addressed in `0bc5812548` — `close()` iterates `_liveFeeds.toList()` and `_feedFor` now throws once `_closed` is set, so the mid-iteration mutation is gone. `5d71f5556b` added the matching `resetPaginationDepth` no-op after close. Resolving.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Addressed in `0bc5812548` — the `system` arm is now `["list_add"]` rather than the `null` no-types sentinel, and `list_add` is in funnelcake's `is_allowed_notification_type` allowlist on `origin/main`. Resolving.

realmeylisdev (commented):
Re-reviewed at `6aa6bf3`. All four of my earlier threads are fixed (`0bc5812` / `2af212f` / `5d71f55`) — replied and resolved. Re-checked the funnelcake side on `origin/main`: the `types` filter still lands in the innermost `WHERE` before `GROUP BY n.card_key` / `ORDER BY` / `LIMIT`, all eight values are in `is_allowed_notification_type`, and `card_key` only collapses by video coordinate for `source_kind == 34236`, so tabs don't cross-contaminate. #6428 is fixed.

Four things from the takeover commits, each reproduced with a throwaway probe at this head — details inline. The mark-read rollback is the one I'd fix before merge; the cold-start cache one is the most user-visible.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -791,8 +918,34 @@ class NotificationRepository {
...
+
+  Map<_NotificationFeed, NotificationPage> _snapshotValuesByFeed() => {
+    for (final feed in _liveFeeds) feed: feed.snapshot.value,
+  };
+
+  Map<_NotificationFeed, int> _pagesLoadedByFeed() => {
+    for (final feed in _liveFeeds) feed: feed.pagesLoaded,
+  };
+
+  void _restoreSnapshots(Map<_NotificationFeed, NotificationPage> values) {
```
realmeylisdev, line 1003: **A failed mark-read POST now rolls back tabs it never touched.**

`_snapshotValuesByFeed()` captures *every* live feed, and `_restoreSnapshots` writes all of them back from the catch arms at :998 and :1039. Any feed that received fresher data while the POST was in flight is reverted to its pre-flight value.

The window is the common one: opening the inbox runs `_markSeenOnOpen` → `markAllAsRead()`, and the user swipes to another tab while that POST is still going.

Probe at `6aa6bf3` — All tab loaded, `markAllAsRead()` started with the POST gated on a completer, `refreshFeed(follow)` landed a page mid-flight, then the POST failed:

```
Follows tab rows before POST failure = 1, after = 0
follow feed emissions = [0, 0, 1, 0]
```

The restored page carries `hasMore: false`, so `_continuePastEmptyPageIfNeeded` doesn't re-arm either — the tab stays blank until a manual pull-to-refresh.

This is new here, not pre-existing. On `origin/main` both methods used `final before = _snapshot.value` against a single subject, so a rollback could only ever touch the one feed the flip touched — and the `before.items.every((n) => n.isRead)` early return made the POST structurally impossible on an empty feed. The per-feed map turns a one-feed rollback into an all-feeds one.

Narrowest fix: build `before` only from the feeds `_flipIsRead` / `_flipAllRead` actually changed, instead of from `_liveFeeds` wholesale.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -312,10 +357,10 @@ class NotificationRepository {
         nextCursorId: response.nextCursorId,
         hasMore: response.hasMore,
       );
-      _pagesLoaded = isFirstPage ? 1 : _pagesLoaded + 1;
-      _emitSnapshotForPage(page, isFirstPage: isFirstPage);
+      feed.pagesLoaded = isFirstPage ? 1 : feed.pagesLoaded + 1;
+      _emitSnapshotForPage(feed, page, isFirstPage: isFirstPage);
 
-      if (isFirstPage) {
+      if (isFirstPage && filter == null) {
```
realmeylisdev, line 425: **Gating the cache write on `filter == null` costs the four category tabs their offline story.**

`_hydrateFromCache` also only writes `_allFeed` (:502 → `_snapshot` → `_feedFor(null)`), so filtered feeds start at `NotificationPage.empty` with no cache path at all. When their first fetch fails, `NotificationFeedBloc._onStarted` sees `state.notifications.isEmpty` and emits `NotificationFeedStatus.failure` (:187-196) — which `notifications_view.dart:78` renders as the full-screen "Failed to load" body.

Cold start with a warm DAO cache and no network, at `6aa6bf3`:

```
ALL    : status=loaded  items=1 refreshError=true   <- cached rows + inline banner
FOLLOW : status=failure items=0 refreshError=true   <- full-screen failure page
```

Before this PR every tab read the same hydrated snapshot, so they degraded to a cached subset plus the inline banner. The same thing happens online whenever one category query 5xxs while All succeeds. `hydrateOnStart` defaults to `true` and the provider doesn't override it, so this is live.

Cheapest fix that keeps the DAO single-writer: when `_feedFor` mints a filtered feed and `_allFeed` already holds hydrated rows, seed the new feed from those rows selected by the tab's kind — the predicate the deleted `_applyFilter` used, which `_belongsInCommentsFeed` already covers for Comments.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -774,14 +841,74 @@ class NotificationRepository {
...
+  /// feed applied a snapshot, so a partial success still advances the
+  /// coordinator's cooldown instead of re-running the whole fan-out on
+  /// every trigger.
+  Future<bool> _refreshAppliedForLiveFeeds() async {
+    final live = _liveFeeds.toList(growable: false);
+    if (live.isEmpty) {
+      return _refreshResult().then((result) => result.applied);
+    }
+
+    final feeds = live
```
realmeylisdev, line 941: **The fan-out is unbounded in time.**

This fixed my earlier "only the `null` feed ever refreshes" note, thanks. But `_feeds` is only ever added to, and `resetPaginationDepth` leaves the entry in the map — so once a user has swiped through the five tabs once, every later resume refreshes all five, serially, for the rest of the repository's life.

Probe at `6aa6bf3`, a single `refreshApplied()`:

```
5 REST requests: [null, [reaction,zap], [comment,reply,mention], [follow], [repost]]
```

and after cancelling every subscription and calling `resetPaginationDepth` for each feed — i.e. the notifications screen fully unmounted — still 5.

`NotificationRefreshCoordinator` runs on app resume and foreground-idle warmup app-wide, not just while the inbox is open, and each request carries its own NIP-98 signature. That's five signer round-trips per resume, which isn't free for remote-signer users.

Two options that compose: skip feeds with no listener (keeping the unfiltered one unconditionally, since the badge reads it), and/or drop a filtered feed from `_feeds` in `resetPaginationDepth` when its snapshot is empty.

mobile/lib/notifications/bloc/notification_feed_bloc.dart
```diff
@@ -206,16 +270,18 @@ class NotificationFeedBloc
     emit(state.copyWith(isLoadingMore: true));
 
     try {
-      await _notificationRepository.loadNextPage();
+      await _notificationRepository.loadNextPageFor(_filter);
```
realmeylisdev, line 273: **`_maxEmptyPageContinuations` is 3, but only one continuation can ever fire.**

Page 2's snapshot lands while `isLoadingMore` is still `true`, so `_continuePastEmptyPageIfNeeded` parks it in `_emptyPageContinuationPending` — and `_onLoadMore` never flushes that flag. Only `_onStarted` (:168) and `_onRefreshed` (:311) do.

Probe at `6aa6bf3`, filtered feed whose pages all filter down to empty:

```
loadNextPageFor called 1 time(s); documented cap is 3
```

Adding `_flushPendingEmptyPageContinuation();` right after the success emit on the next line takes it to 3, with the existing suite still green — I checked both directions.

This is exactly the case you flagged as still open: a Comments feed whose pages are mostly video mentions gets one extra page of runway, not three. Neither continuation test catches it — both assert `.called(1)`, so the cap itself is unpinned.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Fixed in `a91a709eb8`. `before` is now built only from the feeds `_flipIsRead` / `_flipAllRead` actually change, so a tab that landed a page mid-flight keeps it. `markAllAsRead`'s all-read early return is preserved (nothing captured → no POST). Regression test reproduces the original 1 → 0 row loss and is red without the fix.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Fixed in `659cbd73d1`. A filtered feed now seeds from the unfiltered feed's rows selected by the tab's kind — both when it is minted and when hydration lands after the tabs already mounted, since the repository is built when `ProfileRepository` resolves. Rows stay placeholder-grade; the feed's own first page replaces them.

mobile/packages/notification_repository/lib/src/notification_repository.dart
realmeylisdev: Fixed in `4a8fb3870f` — took the first of the two options. Filtered feeds are skipped when nothing is subscribed to them; the unfiltered feed is always refreshed since the badge reads it app-wide. The three existing `refreshApplied` tests now subscribe the way the bloc does, and one of them was strengthened so the unfiltered feed is genuinely live for the error-isolation assertion.

mobile/lib/notifications/bloc/notification_feed_bloc.dart
realmeylisdev: Fixed in `1f5f0e8a3d` — `_flushPendingEmptyPageContinuation()` after the success emit, as described. New test pins the cap; it reports 1 without the fix and 3 with it.

realmeylisdev:
Pushed four commits for the four points in my last review — revert any independently: `a91a709` scopes the mark-read rollback to the feeds it flipped, `659cbd7` seeds the category tabs from the cached unfiltered feed so a failed first fetch keeps cached rows instead of the full-screen failure body, `4a8fb38` stops the resume fan-out refreshing tabs nobody is subscribed to, `1f5f0e8` flushes the deferred continuation in `_onLoadMore` so the documented cap of 3 is real.

Each has a regression test that is red without its fix. `flutter analyze lib test integration_test` clean, `test/notifications/` + `test/blocs/notifications/` 201 green, `notification_repository` 160 green at 95.17% line coverage against the 94 gate. All four threads resolved; merging once checks land.

PR #2311, opened 2026-03-19
feat(settings): split monolith into sub-screens with account header

omartinma:
## Description

Replace the hamburger menu drawer with a gear icon on the Profile screen and split the monolithic settings screen (1002 lines) into a hub-and-spoke layout with focused sub-screens. Adds a centered account header with avatar, display name, NIP-05/npub, and a Switch Account button with draft warning and confirmation dialogs.

**Related Issue:** Closes #2309

**Follow-up:** #2310 — UI/UX polish pass on each settings sub-screen

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Code refactor
- [ ] Build configuration change
- [ ] Documentation
- [ ] Chore

## Screenshots

| Screen | Screenshot |
|--------|-----------|
| Profile | image |
| Settings Hub | image |
| Content Preferences | image |
| Nostr Settings | image |
| Support Center | image |
| Safety and Privacy | image |

PR #1037, opened 2026-01-23
feat: Profile Management and VineBottomSheet confirmation dialogs

Chardot:
# This PR is part of the theming effort under the [Profile Management Epic](https://github.com/orgs/divinevideo/projects/2/views/6?pane=issue&itemId=3716973461&issue=divinevideo%7Cdivine-mobile%7C600).

## Summary

This PR introduces comprehensive theming updates for profile screens and establishes a new VineBottomSheet architecture for confirmation dialogs and action menus.

- Redesigned Follow/Following buttons with new styling
- Updated Followers and Following screens with consistent theming
- Themed RefreshIndicator across the app
- Replaced gradient thumbnail placeholders with flat surface colors
- New VineBottomSheet architecture supporting scrollable and fixed modes

### My Profile (`profile_screen_router.dart`)
- Custom scaffold with themed AppBar
- More button (⋯) with share profile and copy public key (npub) actions
- Edit and Library buttons with icons
- Updated tab styling for Videos/Liked/Reposts
- Inline copy button for npub in header

### Creator Profile (`other_profile_screen.dart`)
- Redesigned Follow/Following/Blocked button states with icons
- More menu with copy public key, unfollow, block/unblock actions
- Animated block/unblock confirmation states within the sheet
- Unblock flow accessible from Blocked button

### Edit Profile (`profile_setup_screen.dart`)
- Transparent AppBar with surfaceContainerHigh background
- Profile picture section with overlapping upload buttons (camera, gallery, URL)
- Custom form inputs with underline borders and focus state colors
- Bio field with 360 char limit and live character counter
- Read-only public key (npub) field
- Username field with @ prefix and @divine.video suffix
- Cancel/Save buttons fixed at bottom
- Nostr info sheet explaining npub/nsec/username concepts
- Image URL sheet using VineBottomSheet with proper keyboard handling

### Followers Screen (`my_followers_screen.dart`, `others_followers_screen.dart`)
- Surface background color
- Updated AppBar with follower count subtitle
- Consistent theming between own and other users' views

### Following Screen (`my_following_screen.dart`, `others_following_screen.dart`)
- Surface background color
- Updated AppBar with following count subtitle
- Consistent theming between own and other users' views

### Profile Widgets
- **Follow Button**: Green filled with user-plus icon; Following: outlined with user-check icon; Blocked: error-styled with prohibit icon
- **Profile Header**: Updated typography, inline npub copy button
- **Profile Grids**: Flat surfaceContainer placeholder instead of gradient/blurhash
[...]

PR #6449, opened 2026-07-27
feat(settings): surface account deletion from the Support Center

realmeylisdev:
## What

Deleting an account meant finding Settings → **Nostr Settings** → scroll → Danger Zone. #2311 moved it off the Settings hub, #1037 removed it from the own-profile sheet, and the one entry point left sits behind a developer-facing label a user has no reason to open. #6335 was filed from Support Center by someone who could not find it.

This adds the entry where that ticket came from, and makes the two entry points share one implementation.

This is the discoverability half of #6335, split out of #6417 at review request so that PR stays scoped to the deletion-correctness fix.

## Changes

- Adds an authenticated-only **Delete Account and Data** entry to Support Center, reusing `nostrSettingsDeleteAccount` / `nostrSettingsDeleteAccountSubtitle` verbatim — no new copy, no new locale work.
- Extracts the deletion sequence into `startAccountDeletionFlow()` in `lib/widgets/delete_account_action.dart`, so both entry points share one implementation of the confirmation gate, the opt-in username burn, and the step ordering rather than two that can drift. Screens depend on it instead of on each other.
- Moves the progress overlay shared with Remove Keys to `ModalProgressOverlay` in the same layer, and stops it being a route on the way. A route can be popped by a back gesture while the caller is still awaiting — for Remove Keys that means the spinner vanishes while keys are being deleted underneath the user. An `OverlayEntry` cannot be popped, so the caller keeps the lifetime it owns.
- `_SupportTile` gains `DivineIconName` and `titleColor` support so the destructive entry reads consistently without adding a raw `Icons.*` call to a file whose count is frozen.
- The Danger Zone row swaps `Icons.delete_forever` for `DivineIconName.trash`, so the same action stops rendering with two different icons depending on which screen you reached it from.

An earlier revision also put the entry on the Settings hub. That row and the Danger Zone row ran the same flow one tap apart, so the hub row is dropped (d36f0ce716) and the settings tree keeps a single entry, in the section that already groups the irreversible actions.

## Tests

`test/widgets/settings_delete_account_test.dart` asserted an `ACCOUNT` section, `Icons.delete_forever` and `Colors.red` — the tile #2311 deleted — which is why three of its four tests were `skip: true`. The fourth was live but could not fail: it asserted `findsNothing` against a screen that never had the tile. With no hub entry, the file is deleted outright and its assertions live on the screens that do ship one:

- `test/screens/settings/support_center_screen_test.dart` — entry present when authenticated, absent when signed out, and copy reused rather than newly added
- `test/widgets/nostr_settings_screen_test.dart` — the Danger Zone row renders as destructive (`VineTheme.error`, `DivineIconName.trash`)
- both suites resolve copy through `AppLocalizations` rather than hardcoded English, with the `de` value asserted absent to prove the widget really reads from l10n

The Support Center auth gate was verified by removing it and watching the matching test go red.

- `flutter test test/widgets/nostr_settings_screen_test.dart test/screens/settings/support_center_screen_test.dart` (14) — the test that popped the spinner's route is replaced by one asserting a back gesture no longer removes it (verified failing against the route-based version), plus one covering what the old test was really guarding: a dismiss that lands after the screen is gone
- `flutter test test/widgets test/screens` (4268, 104 skipped)

## Verified on device

Ran the flow to completion on a physical iPhone (iOS 26.5.2, debug build) against production, on a `divineOAuth` account, entering from Nostr Settings → Danger Zone. The run predates the rebase onto #6417, so it exercised the flow without that PR's pre-flight gate:

- burn-first ordering held — the handle release logged *before* `Starting account deletion`
- 80 events fetched and deleted as batched kind-5 requests grouped by kind, then a kind-62 with `["relay", "ALL_RELAYS"]` per NIP-62
- Keycast account deletion returned success; sign-out wiped keys and local data and landed on `/welcome`
- relay-side: `relay.divine.video` and `relay.damus.io` serve nothing for that pubkey afterwards; `relay.nos.social` (4) and `nos.lol` (1) still do, which is what the subtitle's "relays, clients, search indexes … may keep copies" is there for
- the released handle no longer resolves at `divine.video/.well-known/nostr.json`

The run also surfaced a pre-existing defect in `executeAccountDeletion` (not in this diff): after `signOut`, the router has already redirected to `/welcome`, so `dismissDialog()` → `context.pop()` throws `GoError: There is nothing to pop`. The throw skips the success snackbar and the screen-reader announcement, and escapes to Crashlytics — so a *successful* deletion reports an error and tells the user nothing.

## Baselines

- `raw_dialogs`: 51 → 49. The baseline loses a key rather than moving one — the overlay no longer uses `DialogRoute` at all. (The first push did relocate the key; CI correctly rejected that, since the guard is per-file and a move is not a win.)
- `raw_icons`: `nostr_settings_screen.dart` 8 → 6, locking the `Icons.delete_forever` removal along with one count the baseline was already stale on.
- `skip_tests`: `settings_delete_account_test.dart` drops out entirely.

Refs #6335, #4881 · follow-up to #6417


mobile/lib/screens/settings/support_center_screen.dart
```diff
@@ -85,6 +86,22 @@ class SupportCenterScreen extends ConsumerWidget {
...
+                  title: l10n.nostrSettingsDeleteAccount,
+                  subtitle: l10n.nostrSettingsDeleteAccountSubtitle,
+                  iconColor: VineTheme.error,
+                  titleColor: VineTheme.error,
+                  onTap: () => startAccountDeletionFlow(
+                    context: context,
+                    ref: ref,
+                    screenName: 'SupportCenterScreen',
+                  ),
+                ),
```
hm21, line 107: Is there any scenario where users can reach this screen while `userPubkey` is still `null` or empty? I think only authenticated users can access this screen, right?

Also, I'm not sure if we should show the delete button here. I understand that this user couldn't find the delete button, but if we start adding every setting that users can't find to this screen, I worry it could become cluttered. This isn't a blocker, just a design decision that I wanted to bring up.

hm21 (approved):
Approved. Four review fixes pushed as separate commits — revert or tweak any of them.

- `b9e2efe79` — the tile carried two mutually-exclusive icon APIs (`IconData` + optional `divineIcon`, held together by an assert). Collapsed to one required `DivineIconName` and migrated the screen's remaining raw `Icons.*`, so the file drops out of the `raw_icons` baseline. **Two need your eye:** the design system has no bug or lightbulb glyph, so Report a Bug is `warningCircle` and Request a Feature is `sparkle`. Swap them if design disagrees.
- `920bce134` — the row's gate watched `authServiceProvider`, a keepAlive singleton whose identity never changes, so it never rebuilt; and a non-null pubkey isn't an authenticated session. Now matches the Danger Zone entry's `currentAuthStateProvider` gate. The copy test that restated its sibling is replaced by one that tapping the row reaches the confirmation gate.
- `6706d2725` — `ColoredBox` stops pointers, but a screen reader activates by node id via `SemanticsOwner.performAction` and never hit-tests, so every control behind the spinner stayed activatable. `DialogRoute` got this free through `ModalBarrier`'s `BlockSemantics`. Since `startAccountDeletionFlow` has no re-entrancy guard, a second activation during the 3s profile resolve stacked a second confirmation dialog. Added `BlockSemantics` + a label, plus a test that goes red without it.
- `15155748c` — same overlay-vs-route change: back now pops the screen, not the spinner, so the post-await `ScaffoldMessenger.of(context)` on the `SecureKeyStorageException` path found an unmounted context and dropped its snackbar. Keys not deleted, still signed in, nothing said so. Messenger is resolved before the await now.

`flutter analyze` clean, affected suites and all seven ratchets green locally.

Two things I did **not** change, because they're your call:

1. `/support-center` is the one non-review route `appRouterRedirect` whitelists for restricted minor accounts, and `MinorAccountReviewScreen` links straight to it. So this PR newly puts account deletion — including the username burn — in front of an account whose moderation review is still open, on the only screen it can reach. That may well be the right answer for a minor who wants out, but it's a reachability change worth naming rather than shipping silently. @NotThatKindOfDrLiz for the product call.
2. My earlier question about whether deletion belongs on this screen at all still stands.

Two corrections for the description: the broad-suite count is 4307, not 4268; and `d36f0ce716` isn't in this PR — it was orphaned by a force-push and replaced by `a633bf0399`.

Issue #6166, opened 2026-07-18
chore(deps): revert c2pa_flutter to a pub dependency once available


realmeylisdev:
`c2pa_flutter` is git-pinned in `mobile/pubspec.yaml` to `guardianproject/c2pa-flutter` (tag `0.0.3`, commit `847d0c0183c9c09946ce321c107decd3fd9f56b2`) because that plugin is not published on pub.dev.

The same-named pub.dev package `c2pa_flutter` is an **unrelated third-party rewrite** (`Township-Innovation`, unverified publisher; `flutter_rust_bridge` FFI; bundles c2pa-rs 0.75.19; `RemoteSigner` removed) and is **not** a migration target — Divine's signing services (`c2pa_signing_service`, `c2pa_identity_manifest_service`, `native_proofmode_service`) rely on `RemoteSigner` (proofsign-server) and the guardianproject c2pa-android 0.76.1 native stack.

**Action:** revert to a `^`-versioned pub dependency if/when guardianproject publishes `c2pa_flutter` to pub.dev. Until then the pin is intentional and hardened to an immutable commit SHA (done in the PR splitting this out of #3637).

Context: split out of #3637. Related to #3655 (dependency monitoring).

realmeylisdev:
Re-checked 2026-07-28. Still blocked, no upstream movement — but the exit condition as written can never be met, so this needs a re-scope.

**pub.dev's `c2pa_flutter` name is permanently taken.** It's still Township-Innovation `0.1.0` (published 2026-02-10, still the only version ever, `publisherId: null` — no verified publisher). pub.dev doesn't recycle package names, so guardianproject cannot publish under `c2pa_flutter` barring a rare admin-mediated transfer. They publish nothing on pub.dev today (`publisher:guardianproject.info` returns empty).

So "revert to a `^`-versioned pub dependency once guardianproject publishes `c2pa_flutter`" describes something that will not happen. If they ever ship, it's under a different name, which makes this a **rename plus API migration**, not a version-constraint swap. Reworded the in-repo TODO to say that in #6451; the pin itself is untouched.

[...]

PR #6451, opened 2026-07-27
chore(deps): correct the c2pa_flutter pin's removal condition

realmeylisdev:
## Description

The `c2pa_flutter` git pin carried a `TODO(#6166)` promising to "revert to a pub/version dep if guardianproject publishes `c2pa_flutter` to pub.dev." That condition can never be met, so the TODO was pointing at a wait that will never end.

`c2pa_flutter` on pub.dev is permanently held by an unrelated third-party package (Township-Innovation, `0.1.0`, published 2026-02-10, still the only version, `publisherId: null`). pub.dev does not recycle package names, so guardianproject cannot publish under that name. If they ever publish, it will be under a different name — making the move a rename plus API migration rather than a version-constraint swap.

This PR only rewords the TODO to say that. The pin itself is unchanged and stays: I downloaded `c2pa_flutter-0.1.0.tar.gz` (sha256 matched pub.dev's reported hash) and it contains **zero** occurrences of `RemoteSigner`, which `mobile/lib/services/c2pa_signing_service.dart:487` constructs. Three services and three tests import `package:c2pa_flutter/c2pa.dart`.

The pin is also still exactly current: guardianproject/c2pa-flutter `main` HEAD *is* the pinned SHA (`ahead_by: 0, behind_by: 0`), and tag `0.0.3` still resolves to it — the tag has not moved.

**Related Issue:** Relates to #6166

## Out of Scope

- Changing the pin, its ref, or the plugin itself — the pin is correct and stays.
- Closing #6166. It stays open, re-scoped to the rename+migration framing; details posted there.
- The native layers are also GitHub-sourced inside the plugin (`com.github.redaranj:c2pa-android:0.0.9-beta.7` via JitPack; `redaranj/c2pa-ios` exact `0.0.9-beta.7` via SPM), so a future pub.dev publish would de-git only the Dart layer. Noted on the issue, not addressed here.

## Verification

- `flutter pub get` from `mobile/` — resolves cleanly; `pubspec.lock` unchanged, confirming the comment-only edit does not alter resolution.
- `git status` after the edit shows only `mobile/pubspec.yaml` modified (no lock or `Package.resolved` churn).
- No Dart files touched, so analyze/test suites are unaffected (pre-commit hook confirmed "No Dart files staged").

## Type of Change

- [x] 🗑️ Chore

### DIFF
--- a/mobile/pubspec.yaml
+++ b/mobile/pubspec.yaml
@@ -137,8 +137,9 @@ dependencies:
   # unrelated third-party FFI rewrite — different author, c2pa-rs 0.75.x, no
   # RemoteSigner.) Pinned by immutable commit (tag 0.0.3) so a moved tag cannot
   # silently change the build.
-  # TODO(#6166): revert to a pub/version dep if guardianproject publishes
-  # c2pa_flutter to pub.dev.
+  # TODO(#6166): moving off this pin needs a differently-named pub package —
+  # guardianproject cannot publish as `c2pa_flutter` (name permanently taken),
+  # so it is a rename plus API migration, not a version-constraint swap.
   c2pa_flutter:
     git:
       url: https://github.com/guardianproject/c2pa-flutter.git


rabble (approved):
LGTM. Comment-only reword of the c2pa_flutter pin TODO: removal condition now correct (pub.dev name collision with the unrelated Township-Innovation package verified), pin ref/tag untouched, tracking link #6166 retained per the transitional-TODO rule. CI green.

Issue #6238, opened 2026-07-20
fix(nostr): queryEvents silently returns partial results on timeout


dcadenas:
## Summary

`queryEvents` in the Nostr client can return a **partial** result when its wall-clock timeout fires, and the caller cannot tell a complete result from a truncated one. It returns whatever events arrived so far as if the query succeeded — no exception, no truncation flag, no count check.

## Root cause

`mobile/packages/nostr_sdk/lib/nostr.dart` (`Future<List<Event>> queryEvents`, default `timeout = const Duration(seconds: 5)`):

```dart
try {
  await query(...).timeout(remainingTimeout());
  await completer.future.timeout(remainingTimeout());
} on TimeoutException {
  unsubscribe(subscriptionId);
}
return eventBox.all();   // partial set, indistinguishable from a complete one
```

The query has two exit paths that land on the same `return`: (a) genuine completion when every relay sends `EOSE` (End Of Stored Events), and (b) the 5s timeout firing before that. The caller sees an identical `List<Event>` either way.

Two independent truncation mechanisms both produce this silent short read:

1. **Client timeout** — large result sets don't finish arriving in 5s (per-event isolate verify + JSON decode + id recompute limit throughput).
2. **Relay cap** — funnelcake caps stored-event replay at `max_limit` (5000; `MAX_LIMIT` env, advertised in NIP-11). A client asking for more than the cap receives exactly the cap plus a *clean* `EOSE` with no "there's more" signal, and wrongly concludes the result is complete.

## Impact

Any caller that passes a `limit` and trusts the result for completeness can silently operate on a fraction of the data. This is a shared primitive used across the app. Callers cannot distinguish "this is everything" from "this is what arrived in 5 seconds" or "this is the relay's cap."

A merged PR already flagged this as a known gap: PR #5831 ("fix(likes): chunk the Kind 5 deletion REQ...") notes in its Out of Scope that a clean guard "needs a `queryEvents` error-surfacing change (it is fail-open by contract today); tracked as a follow-up." PR #5562 similarly "preserves the existing return-partial-results behavior." No issue tracked that follow-up until now.

## Proposed fix

The EOSE-completion machinery already exists (`relay_pool.dart` fires the query-complete callback when all relays send `EOSE`). The work is surfacing which exit happened, plus pagination:

1. Distinguish EOSE-complete from timeout — return a completeness signal or throw on timeout, rather than returning partial as success.
[...]

Issue #5863, opened 2026-07-05
perf(nostr): relay event signature verification runs on the main isolate (app-wide jank)


realmeylisdev:
## Summary

App-wide lag reported on Discord (profile counts stuck on "—" for ~4s at cold
start; slow, choppy camera preview that smooths out once recording starts) traces
to **relay event ingestion doing cryptographic signature verification on the main
(UI) isolate** for every incoming event — ungated, and **deduplicated *after*
verify**, so the same event arriving from N relays is verified N times.

The team's working theory was "background DM decryption starves the app." That is
**misattributed**: DM crypto is already off the main isolate
(`DmVerifyIsolate` / `DmDecryptWorker`). The real continuous main-isolate hog is
relay-event verify.

## Evidence (main)

- Path (all on the main isolate): `relay_base.dart:105-115` `jsonDecode` →
  `relay_pool.dart:151` `_onEvent` → `:416` `Event.fromJson` → `:417`
  `!event.isValid || !event.isSigned`.
- Cost: `event.dart:130` sha256 id recompute + `event.dart:141` bip340
  `schnorr.verify`. **Benchmarked ~6.5 ms per verify** (AOT) — 99.6% of per-event
  cost; sha256-id and jsonDecode are ~14 µs each.
- **Dedup happens AFTER verify** (`relay_pool.dart:473`), so the same event id from
  each connected relay (divine + indexers + NIP-65 user relays, ~4–8) pays full
  verify; reconnect/`autoSubscribe` replays re-verify too.
- The EVENT branch (`relay_pool.dart:408-476`) has **no `await`** before dispatch,
  so a cold-start backfill burst blocks the event loop (~50 events × 6.5 ms ≈
  325 ms/relay on a Mac; more on ARM), starving the profile-count REST
  continuation and frame callbacks.
- The camera preview is a native GPU `Texture`
  (`divine_camera/.../camera_preview_widget.dart:249`), but the **UI thread** must
  composite each frame — the same thread verify monopolizes → choppy preview that
  smooths once recording engages `CodecHeavySurfaceCubit`.
- Off-main infra already exists but is unused for feed relays: `EventSignChecker`
  worker-pool is dead code; `RelayIsolate` (off-main verify) is wired only for the
  NIP-46 signer; feed/indexer relays are all `RelayBase` (main-isolate verify).

## Proposed fixes (ranked)

1. **Dedup-before-verify — do first (Low effort, ~50–75% cold-start reduction).**
[...]

PR #6452, opened 2026-07-27
fix(nostr): let a relay CLOSED frame terminate its pending query

### LINKED ISSUE #6453: fix(nostr): a relay CLOSED frame never settles its pending REQ, so the caller burns its whole timeout
## Summary

`RelayPool` treats a relay `CLOSED` frame as a COUNT-only concern. When a relay `CLOSED`s a pending **`REQ`**, nothing settles the query: the completion callback stays armed in `_queryCompleteCallbacks` and the relay keeps the subscription in its `_queries` map. The caller blocks for its entire timeout and then reports `timedOut: true` for what was actually an immediate relay refusal.

At the 5s `queryEvents` default that turns every refused one-shot fetch into a **5-second stall that returns an empty list**.

## How it surfaced

Bug report (2026-07-27, app `1.0.17+815`, iOS 26.5.2, release build):

> Trying to view a notification that someone had mentioned me — and it wouldn't take me to the right place, just took me to their profile page.

The notification target was `0739befe110ec68c28a33650b84c10b356015ec3b8401787d05a84ae285c976a`, a live kind-34236 video. `NotificationTargetResolver` calls `fetchEventById`, which returns `null` when `queryEvents` comes back empty, and `NotificationsView` falls back to the actor's profile.

Nine taps in the attached log, paired FIFO by the tap log line and its fallback:

| Tap (`Navigating to video from notification`) | Outcome | Δ |
|---|---|---|
| 13:07:19.6217 | fallback to profile | **5.002 s** |
| 13:07:20.8303 | fallback to profile | **5.001 s** |
| 13:07:23.4389 | fallback to profile | **5.003 s** |
| 13:07:27.9229 | fallback to profile | 6.722 s |
| 13:07:38.9747 | fallback to profile | **5.003 s** |
| 13:07:40.0833 | fallback to profile | **5.002 s** |
| 13:07:41.2083 | fallback to profile | **5.002 s** |
| 13:07:50.7349 | fallback to profile | **5.002 s** |
| 13:07:52.7769 | **routed to `/video/0739befe…`** | 4.947 s |

Seven of the eight failures land on the 5s `queryEvents` deadline to within 3 ms. The ninth attempt fetched the same event, from the same relay, in the same session, and returned just inside the budget — so the event was reachable throughout and the failures were the fetch being starved, not the event being absent. A direct websocket probe of the same relay returns `EVENT` + `EOSE` for that id in ~1.2 s.

## Root cause

On `origin/main` (`1fd7f3a269`):

**`mobile/packages/nostr_sdk/lib/relay/relay_pool.dart:1059`** — the `CLOSED` branch parses the subscription id and reason, logs them, and then only checks for a COUNT query:

```dart
} else if (messageType == 'CLOSED') {
  // ... parse subscriptionId, reason ...
  log('📡 CLOSED from ${relay.url}: $subscriptionId - $reason');
  // Check if this is a COUNT query being refused
  if (relay.hasCountQuery(subscriptionId)) {
    relay.failCountQuery(subscriptionId, reason);
  }
}
```

There is no `REQ` counterpart. The `EOSE` branch (`:814`) is the only path that fires `_queryCompleteCallbacks` and removes the subscription, so a `CLOSED`-ed `REQ` is simply orphaned.

**`mobile/packages/nostr_sdk/lib/nostr.dart`** (`queryEventsDetailed`) then waits out the whole budget and cannot tell the caller what happened:

```dart
try {
  await query(...).timeout(remainingTimeout());
  await completer.future.timeout(remainingTimeout());
} on TimeoutException {
  timedOut = true;
  unsubscribe(subscriptionId);
}
return (events: eventBox.all(), timedOut: timedOut);
```

**`mobile/packages/nostr_client/lib/src/nostr_client.dart:770`** — `queryEvents` defaults to `Duration(seconds: 5)`, which is where the 5.00 s signature comes from. `fetchEventById` (`:851`) maps the empty list to `null`, and the caller treats that as "event does not exist".

`CLOSED` is also excluded from the per-relay ordered frame tail added in #5863 (`relay_pool.dart:685` gates only on `EVENT`/`EOSE`), so even once it settles a query it would do so out of order relative to events still being verified off-main.

## What sends `CLOSED` without `EOSE`

Our own relay does, on three `REQ` paths. From `divine-funnelcake` `origin/main` (`01bec68`), `crates/relay/src/handler.rs`:

- **`:390`** — `handle_stored_replay_send_error` during the stored replay (send failure / client-queue backpressure) sends `CLOSED` (`:117`) and the caller `return Ok(())`s before the `EOSE` at `:411`.
- **`:405`** — a stored-query error sends `CLOSED` and returns, again skipping the `EOSE`.
- **`:416`** — if the `EOSE` send itself fails, the same helper sends `CLOSED` *instead of* the `EOSE`.

Only `:488` (COUNT error) reaches the branch the client actually handles.

Beyond funnelcake, NIP-01 permits any relay to `CLOSED` a `REQ` for `auth-required`, `restricted`, `rate-limited`, `invalid`, or `error` — all of which currently cost the caller a full timeout instead of an immediate result.

## Impact

`queryEvents` / `queryEventsDetailed` has 79 call sites across 24 files in `mobile/lib` and `mobile/packages`. Any of them hitting a `CLOSED`-ing relay pays the full timeout and gets a result indistinguishable from "the relay had nothing", including notification target resolution, video/profile fetches, and one-shot lookups on interactive paths.

## Evidence boundary

`relay_pool.dart` logs through `dart:developer`'s `log()`, which does **not** reach the app's unified `Log` export — all 8 `📡` lines in the attached export come from `VideoEventService`. So the export can neither confirm nor refute that a `CLOSED` frame arrived on these specific taps.

What the export does establish independently is a hard 5.00 s stall returning empty for an event that was demonstrably live. The `CLOSED`-does-not-settle path is a defect in its own right, confirmed by reading `origin/main` and reproduced by a test that fails before the fix:

```
CLOSED should terminate the query immediately;
instead the caller waited 2004ms of the 2000ms budget
```

## Fix

PR #6452:

- `Relay.discardQuery` drops a pending query **without** sending `CLOSE`, since the relay has already forgotten that subscription.
- The `CLOSED` branch settles a pending `REQ` and releases the caller.
- `CLOSED` joins the per-relay ordered tail alongside `EVENT`/`EOSE` (#5863), because a relay can stream part of a stored replay before abandoning it and settling out of order would drop events still being verified.
- `EOSE`'s completion logic is extracted into `_fireQueryCompleteIfSettled` and shared, so both terminal frames settle a query identically. The helper removes the callback before invoking it, matching `query()`'s existing remove-then-call ordering so a re-entrant callback cannot double-fire.

## Tests

New `mobile/packages/nostr_sdk/test/unit/relay_pool_closed_frame_test.dart` — a fake relay that delivers `CLOSED` for a pending `REQ`, asserting the caller is released well inside its budget. Verified red before the fix, green after. Its sibling `EOSE` test passes both before and after, pinning the shared-helper refactor as behaviour-preserving.

Gotcha worth recording for anyone extending it: `relayDoQuery` calls `relay.send()` *before* `saveQuery()`, so delivering the `CLOSED` from inside an overridden `send()` finds no pending query and silently does nothing. Yield until `relay.checkQuery(subId)` is true before delivering.

## Out of scope

- **#6238** — distinguishing "relay refused" from "relay had nothing" in the value `queryEventsDetailed` returns needs a new field on the result record. This issue only stops the caller from burning its whole timeout budget.
- **The notification-routing half of the original report.** #6299 (`46e6b886b6`, merged 2026-07-26, ships in 1.0.18) makes a video-sourced mention resolve to a `VideoNotification` anchored on the sender's source video, so the tap navigates synchronously with no network call. That removes this stall from the mention path specifically; it does not touch the underlying `CLOSED` defect for the other 78 call sites.

realmeylisdev:
## Description

`RelayPool` handled a `CLOSED` frame by logging it and failing a matching NIP-45 COUNT query, but did nothing for a pending `REQ`. The query's completion callback stayed armed in `_queryCompleteCallbacks` and the relay kept the subscription in `_queries`, so nothing ever settled the query - the caller blocked for its entire timeout and then reported `timedOut: true` for what was actually a relay refusal. At the 5s `queryEvents` default that turns every refused one-shot fetch into a 5s stall returning an empty list.

Funnelcake reaches that path whenever the stored-event query errors or the stored replay is backpressured (`crates/relay/src/handler.rs:117`, `:405`, `:433`, `:488`) - it sends `CLOSED` with no `EOSE`.

Found while tracing a bug report where tapping a mention notification landed on the actor's profile instead of the video. `NotificationTargetResolver` falls back to the profile when `fetchEventById` returns null, and every failed tap in the attached log took exactly 5.00s - the `queryEvents` timeout. The target event was live on the relay the whole time and returns `EVENT` + `EOSE` in ~1.2s to a direct probe.

### Changes

- `Relay.discardQuery` drops a pending query **without** sending `CLOSE`, since the relay has already forgotten that subscription.
- The `CLOSED` branch settles a pending `REQ` and releases the caller, except for auth-retryable `CLOSED` reasons that must keep the query saved for post-AUTH replay.
- Query completion is gated until initial relay fan-out has finished registering accepted sends, so a fast terminal frame from one relay cannot complete before a slower relay has joined the pending query set.
- `CLOSED` now rides the per-relay ordered tail alongside `EVENT`/`EOSE` (#5863). A relay can stream part of a stored replay before abandoning it, so settling the query out of order would drop events still being verified.
- EOSE's completion logic is extracted into `_fireQueryCompleteIfSettled` and reused, so both terminal frames settle a query identically. The helper removes the callback before invoking it, matching `query()`'s existing remove-then-call ordering so a re-entrant callback cannot double-fire.

**Related Issue:** Fixes #6453. Relates to #6238.

## Out of Scope

Distinguishing "relay refused" from "relay had nothing" in the value returned by `queryEventsDetailed` - that needs a new field on the result record and is #6238's scope. This PR only stops the caller from burning its whole timeout budget.

The notification-routing half of the original report is already fixed on `main` by #6299 and is not touched here.

## Verification

- `flutter test test/unit/relay_pool_closed_frame_test.dart` in `packages/nostr_sdk` - 4 passing
- `flutter test test/unit/relay_pool_auth_test.dart` in `packages/nostr_sdk` - 12 passing
- `flutter test test/unit/relay_pool_concurrency_test.dart` in `packages/nostr_sdk` - 15 passing
- `flutter analyze` in `packages/nostr_sdk` - no issues
- `flutter test` in `packages/nostr_sdk` - 400 passing
- `flutter test` in `packages/nostr_client` - 297 passing
- pre-push hook - no merge conflicts, analysis clean, changed CLOSED-frame test passing

## Type of Change

### DIFF
--- a/mobile/packages/nostr_sdk/lib/relay/relay.dart
+++ b/mobile/packages/nostr_sdk/lib/relay/relay.dart
@@ -177,6 +177,13 @@ abstract class Relay {
     return false;
   }
 
+  /// Drops a pending one-shot query without sending `CLOSE`.
+  ///
+  /// Used when the relay itself ended the subscription (a `CLOSED` frame), so
+  /// echoing a `CLOSE` back would name a subscription the relay has already
+  /// forgotten. Returns whether [id] named a pending query.
+  bool discardQuery(String id) => _queries.remove(id) != null;
+
   bool checkQuery(String id) {
     return _queries[id] != null;
   }
--- a/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
+++ b/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
@@ -652,6 +652,31 @@ class RelayPool {
     return next;
   }
 
+  /// Fires the one-shot query callback for [subId] once no relay still has it
+  /// pending, then forgets the callback.
+  ///
+  /// Both terminal frames route here: `EOSE` (the relay finished sending
+  /// stored events) and `CLOSED` (the relay ended the subscription without
+  /// finishing). Either way the caller must be released rather than left to
+  /// burn its whole timeout budget.
+  void _fireQueryCompleteIfSettled(String subId) {
+    final callback = _queryCompleteCallbacks[subId];
+    if (callback == null) return;
+
+    final list = [
+      ..._relaysSnapshot(),
+      ..._tempRelaysSnapshot(),
+      ..._cacheRelaysSnapshot(),
+    ];
+    for (final r in list) {
+      // Some relay still owes us a terminal frame for this query.
+      if (r.checkQuery(subId)) return;
+    }
+
+    _queryCompleteCallbacks.remove(subId);
+    callback();
+  }
+
   Future<void> _onEvent(Relay relay, List<dynamic> json) async {
     final messageType = _stringAt(relay, json, 0, 'message type');
     if (messageType == null) return;
@@ -677,12 +702,17 @@ class RelayPool {
       log('📡 Raw message from ${relay.url}: $json');
     }
 
-    // #5863: when an off-main verify worker is wired, serialize EVENT/EOSE
-    // per relay so the asynchronous verify preserves in-order delivery — an
-    // EOSE must not complete a query before its preceding events finish
-    // verifying. With no worker the original synchronous path runs unchanged.
+    // #5863: when an off-main verify worker is wired, serialize the frames
+    // that carry or terminate a query per relay so the asynchronous verify
+    // preserves in-order delivery — a terminal frame must not complete a query
+    // before its preceding events finish verifying. CLOSED is terminal too: a
+    // relay may stream part of a stored replay and then abandon it, so
+    // completing on CLOSED out of order would drop the events already in
+    // flight. With no worker the original synchronous path runs unchanged.
     if (_verifyWorker != null &&
-        (messageType == 'EVENT' || messageType == 'EOSE')) {
+        (messageType == 'EVENT' ||
+            messageType == 'EOSE' ||
+            messageType == 'CLOSED')) {
       return _enqueueOrdered(
         relay,
         () => _dispatchTypedFrame(relay, json, messageType),
@@ -819,28 +849,7 @@ class RelayPool {
       }
       var isQuery = await relay.checkAndCompleteQuery(subId);
       if (isQuery) {
-        // is Query find if need to callback
-        var callback = _queryCompleteCallbacks[subId];
-        if (callback != null) {
-          // need to callback, check if all relay complete query
-          final list = [
-            ..._relaysSnapshot(),
-            ..._tempRelaysSnapshot(),
-            ..._cacheRelaysSnapshot(),
-          ];
-          bool completeQuery = true;
-          for (var r in list) {
-            if (r.checkQuery(subId)) {
-              // this relay hadn't compltete query
-              completeQuery = false;
-              break;
-            }
-          }
-          if (completeQuery) {
-            callback();
-            _queryCompleteCallbacks.remove(subId);
-          }
-        }
+        _fireQueryCompleteIfSettled(subId);
       } else {
         // Handle EOSE for long-running subscriptions
         final subscription = _subscriptions[subId];
@@ -1078,6 +1087,20 @@ class RelayPool {
       // Check if this is a COUNT query being refused
       if (relay.hasCountQuery(subscriptionId)) {
         relay.failCountQuery(subscriptionId, reason);
+      }
+
+      // A REQ the relay refused or abandoned is terminal for that relay: no
+      // EVENT and no EOSE will follow, so nothing else would ever settle the
+      // query. Funnelcake sends this when the stored-event query errors or the
+      // stored replay is backpressured (funnelcake
+      // `crates/relay/src/handler.rs:117`, `:405`, `:433`, `:488`). Without
+      // this the caller waits out its entire timeout and reports a timeout for
+      // what was really a refusal.
+      //
+      // Distinguishing "relay refused" from "relay had nothing" in the
+      // returned record is out of scope here and tracked by #6238.
+      if (relay.discardQuery(subscriptionId)) {
+        _fireQueryCompleteIfSettled(subscriptionId);
       }
     }
   }
--- a/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
+++ b/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
@@ -95,6 +95,8 @@ class RelayPool {
 
   final Map<String, Function> _queryCompleteCallbacks = {};
 
+  final Set<String> _queryFanoutInProgress = {};
+
   /// Tracks which relays have sent EOSE for each subscription.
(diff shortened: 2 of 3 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I took over the review follow-ups and pushed `37b6d07568` to this branch.

What changed:
- Preserved pending queries on auth-retryable `CLOSED` reasons so NIP-42 post-AUTH replay still works.
- Gated one-shot query completion until initial relay fan-out has finished registering accepted sends.
- Added regression coverage for `CLOSED auth-required` -> AUTH OK -> REQ replay, and for fast `CLOSED` not completing before a slower relay joins the query.
- Removed the timing-sensitive assertion from the CLOSED regression test and updated the PR body verification.

Verification passed locally:
- `flutter test test/unit/relay_pool_closed_frame_test.dart` in `packages/nostr_sdk`
- `flutter test test/unit/relay_pool_auth_test.dart` in `packages/nostr_sdk`
- `flutter test test/unit/relay_pool_concurrency_test.dart` in `packages/nostr_sdk`
- `flutter analyze` in `packages/nostr_sdk`
- `flutter test` in `packages/nostr_sdk` - 400 passing
- `flutter test` in `packages/nostr_client` - 297 passing
- pre-push hook - mergeability check, analysis, changed CLOSED-frame test

GitHub CI is green on the pushed commit. Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Final takeover pass is clean. The CLOSED query settlement now preserves auth-required query replay, waits for initial fan-out registration before completing, and has regression coverage for both paths. Local verification and GitHub CI are green.

realmeylisdev:
Thanks @NotThatKindOfDrLiz — takeover reviewed and merged as `2fe2bfb27`.

Both guards check out. The auth-replay guard is the one my version got wrong: discarding on the pre-AUTH probe's `CLOSED auth-required` left the post-AUTH `getQueries()` re-issue with nothing to re-send. The fan-out gate closes a hole I missed — a fast terminal frame could complete the query before a slower relay had registered its send. I reverted each guard separately and confirmed its new test goes red, so both are pinned.

Agreed on dropping the stopwatch assertion; `timedOut: false` already fails the test if the query strands, so it was only flake risk.

PR #5947, opened 2026-07-07
fix(comments): keep reply composer above keyboard

NotThatKindOfDrLiz:
## What changed

- Threads a DraggableScrollableController through the comments bottom sheet call site.
- Expands the comments sheet to its max size when the composer gains focus, including reply/edit activation paths.
- Re-expands while the composer remains focused if the sheet is pulled below the safe comments snap size.
- Extracts comments sheet sizing values into named constants.
- Adds widget coverage for focus-driven expansion, focused-sheet reclamping, keyboard-dismiss availability, and draft retention.

## Why

The keyboard inset padding was applied inside the sheet's fixed fractional height. At smaller sheet sizes, the fixed chrome plus composer plus keyboard-height padding could exceed the visible sheet region and push the reply composer controls under the keyboard.

Closes #5939.

## Validation

- flutter analyze
- flutter test test/screens/comments/comments_screen_test.dart
- pre-push hook analyzer and changed-file tests passed

PR #6454, opened 2026-07-27
fix(comments): keep composer buttons visible above the keyboard

### LINKED ISSUE #6455: fix(comments): keep composer buttons visible above the keyboard
Auto-created from PR #6454.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6454
**Branch:** `fix/comment-composer-keyboard-layout`

rabble:
Closes #6455

## Description

While writing a comment, the composer pill sat flush against the keyboard top. The send and keyboard-dismiss buttons anchor to the composer's bottom edge, and real keyboards vary in reported height while typing (predictive bar appearing, inset animation lag), so those buttons slipped under the keyboard and became unreachable (see issue screenshot).

Two fixes, one per commit:

1. **`VineBottomSheet` keyboard clearance** (`divine_ui`): `_KeyboardAwareBottomInput` lifted bottom inputs by exactly `viewInsets.bottom`. Now adds a 12px clearance above a raised keyboard, for every sheet that uses the `bottomInput` slot (comments, share message composer, user picker, etc.).
2. **Empty-state overflow**: with the keyboard up, the comments sheet's list area can shrink to near zero and the empty state's fixed `Column` overflowed (debug overflow banner, clipped "Get the party started!" copy in release). Wrapped in a `SingleChildScrollView` so it shrinks gracefully.

**Related Issue:** Reported via screenshot — comment composer submit/cancel buttons hidden behind the keyboard while typing a multi-line comment.

## Out of Scope

- The focus-driven sheet-expand behavior (`#5947`) is untouched.
- No changes to the inline feed composer bar (`inline_comment_composer_bar.dart`), which manages its own inset math.

## Verification

- New regression test in `divine_ui`: modal sheet + draggable snap sizes + multiline composer + injected 300px keyboard inset; asserts composer and action buttons stay above the keyboard with clearance (fails before the fix).
- New regression test: `CommentsEmptyState` pumped into a 40px-high slot no longer overflows (fails before the fix).
- `flutter test` green: full `divine_ui` package suite (662 tests), full `test/screens/comments/` suite (115 tests), `test/goldens/` verify.
- `flutter analyze` clean in `mobile` and `mobile/packages/divine_ui`.
- Reproduced on iPhone 16 Pro Max simulator (iOS 26) with Maestro before the fix (buttons flush/clipped at keyboard top, 9.2px overflow banner); re-ran after the fix: buttons fully visible with clear gap above the keyboard, no overflow banner.

## Type of Change

### DIFF
--- a/mobile/lib/screens/comments/widgets/comments_empty_state.dart
+++ b/mobile/lib/screens/comments/widgets/comments_empty_state.dart
@@ -15,69 +15,76 @@ class CommentsEmptyState extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) => Center(
-    child: Column(
-      mainAxisAlignment: MainAxisAlignment.center,
-      children: [
-        if (isClassicVine) ...[
-          Container(
-            margin: const EdgeInsets.symmetric(horizontal: 24, vertical: 16),
-            padding: const EdgeInsets.all(16),
-            decoration: BoxDecoration(
-              color: VineTheme.accentOrange.withValues(alpha: 0.3),
-              borderRadius: BorderRadius.circular(12),
-              border: Border.all(
-                color: VineTheme.accentOrange.withValues(alpha: 0.5),
-              ),
-            ),
-            child: Column(
-              children: [
-                const Icon(
-                  Icons.history,
-                  color: VineTheme.contentWarningAmber,
-                  size: 32,
+    // Scrollable so the empty state shrinks gracefully when the open
+    // keyboard squeezes the sheet's list area to near zero; a fixed Column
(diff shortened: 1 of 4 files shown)


realmeylisdev (approved):
Verified both fixes by reverting each one locally and watching its new regression test fail, then re-applying and confirming green — real repros, real fixes, not tautological tests.

Checked the other `bottomInput` consumer (people-lists picker) and the empty-state centering for regressions from the `SingleChildScrollView` wrap: none, confirmed by direct measurement. Format, analyze, ratchets, and coverage all clean.

Nice, well-tested fix.

PR #6355, opened 2026-07-24
feat(ios): automated App Store screenshot pipeline

rabble:
## What

Closes #6367.

Adds a DEBUG-only iOS App Store screenshot pipeline for Divine. `bundle exec fastlane screenshots` captures the nine App Store screens on both required device sizes and composites branded captions onto the resulting images.

## How it works

- `ScreenshotMode` is compile-time gated by `kDebugMode && bool.fromEnvironment('SCREENSHOT_MODE')`, so screenshot affordances are tree-shaken out of Release builds.
- The native iOS launcher exports per-run screenshot config into `SharedPreferences`; screenshot startup reads it after the first frame and navigates imperatively to the requested route.
- Screenshot mode seeds a throwaway Nostr account, creator follows, camera/editor fixtures, and stable UI state for repeatable captures.
- Stable anchors live in `SemanticIds` and are consumed by `DivineUITests` XCUITests instead of label text or sleeps.
- Fastlane captures raw screenshots, then the Pillow framing script composites brand captions.
- Uploading to App Store Connect remains intentionally disabled. No credentials are committed.

## Takeover fixes

- Rebased/replayed the PR onto current `origin/main`, resolving the `main.dart` conflict and clearing the untested-services ratchet staleness.
- Removed the dead router initial-route helper and corrected README/Swift comments to describe the actual startup navigation path.
- Removed commented-out `deliver(...)` code from the disabled upload lane.
- Extracted pure launch-config parsing and added unit coverage for empty route handling and strict `SCREENSHOT_SEED_CLIPS=1` parsing.
- Hardened screenshot asset materialization so stale/truncated files are rewritten.
- Migrated existing screenshot-facing semantic identifiers to `SemanticIds`, including the recorder button now using snake_case.
- Fixed Divine brand casing in screenshot docs/comments and avoided empty DEBUG UserDefaults writes on normal launches.

## Verification

- `flutter pub get`
- `flutter test test/config/screenshot_mode_test.dart test/services/screenshot_mode_service_test.dart test/widgets/video_recorder/video_recorder_record_button_test.dart test/widgets/auth/auth_hero_section_test.dart`
- `bash scripts/check_untested_services_floor.sh`
- `flutter analyze lib test`
- `very_good test --optimization --concurrency=4 --exclude-tags integration --test-randomize-ordering-seed random`

PR #6459, opened 2026-07-28
feat(screenshots): OG-Viner fixtures and refreshed App Store capture pipeline

### LINKED ISSUE #6460: feat(screenshots): OG-Viner fixtures and refreshed App Store capture pipeline
Tracks PR #6459.

## Summary

Refresh the App Store screenshot pipeline so the exported iPhone screenshot set uses curated OG-Viner and discover-list fixtures, deterministic marketing framing, and a generated Divine endcard. iPad capture is intentionally deferred until the simulator path can be run and verified end to end.

## Acceptance criteria

- Screenshot mode uses deterministic fixtures for the classics and list-discovery captures without leaking screenshot-only service imports into production UI screens.
- The regular fastlane snapshot lane captures only exported screenshots; preview-video recording helpers are excluded from normal runs.
- Framing fails when a required capture or headline is missing, so stale framed screenshots cannot silently ship.
- Caption layout leaves safe vertical spacing between headline and subhead.
- Fixture and semantic-id contracts have focused test coverage.

Closed by #6459.

rabble:
Closes #6460

## Summary

- Screenshot mode now overrides `topClassicVinersProvider` with a curated OG-Viner fixture list (Reggie Couz et al., all with avatars) so the `01_classics` marketing shot has zero placeholder circles.
- Screenshot mode overrides `discoveredListsProvider` with deterministic fixtures at the account-container boundary, so list discovery stays out of production UI and ignores live relay mutations during captures.
- Refreshed `frame_screenshots.py`, `Snapfile` iPhone device list, `Fastfile` lanes, `DivineScreenshots.swift` coverage, and en-US title/keyword strings.
- Adds white-logo endcard asset for the app preview video.
- Defers iPad screenshot capture from this PR until the iPad simulator path can be run and verified end to end.

Recovered uncommitted work from the `appstore-screenshots` worktree - follow-up to #6355. The independent `video_stats` proof-field fix from the same stash went out separately as #6457. 45 MB of generated `preview/` render artifacts deliberately excluded.

## Test plan

- [x] `flutter analyze`
- [x] `flutter test test/services/screenshot_mode_service_test.dart`
- [x] `flutter test test/widgets/categories_tab_test.dart`
- [x] `bash scripts/check_ui_service_boundary.sh`
- [x] `python3 -B -m py_compile mobile/ios/fastlane/frame_screenshots.py`
- [x] `xmllint --noout mobile/ios/Runner.xcodeproj/xcshareddata/xcschemes/DivineUITests.xcscheme`
- [ ] Full fastlane screenshot run (requires local screenshot simulators; not run locally)


mobile/ios/fastlane/Snapfile
```diff
@@ -1,10 +1,12 @@
 # Snapfile — configuration for `fastlane snapshot`.
 # Docs: https://docs.fastlane.tools/actions/snapshot/
 
-# Both sizes App Store Connect requires (6.9" and 6.5").
+# App Store Connect required display sizes: iPhone 6.9" + 6.5", and the
+# 13" iPad Pro (2064×2752) which covers the iPad requirement.
 devices([
   "iPhone 16 Pro Max",
   "iPhone 11 Pro Max",
+  "iPad Pro 13-inch (M5)",
```
rabble: iPad support stops here and in `frame_screenshots.py`'s canvas map. `DivineScreenshots.swift` has no iPad branch, and `device_body()` is documented as drawing "a rounded black **iPhone** body" yet runs for every canvas — so this slide renders an iPad-aspect capture inside an iPhone-shaped frame.

The test plan's full fastlane run is unchecked, so this device has never actually been captured. I'd run it once or split it into a follow-up; the rest of the PR doesn't depend on it.

mobile/lib/main.dart
```diff
@@ -1534,6 +1547,13 @@ Future<void> _initializeCoreServices(ProviderContainer container) async {
 
   if (ScreenshotMode.enabled) {
     await buildScreenshotModeService(container).prepare();
+    // Deterministic, on-brand lists for the 06_lists capture instead of
+    // whatever public lists the relays happen to surface. Seeded here in
+    // startup rather than by the screen: presentation must not import the
+    // service layer (scripts/check_ui_service_boundary.sh).
+    container
+        .read(discoveredListsProvider.notifier)
+        .setLists(screenshotDiscoverListsFixtures());
```
rabble: This imperative seed survives only because `discover_lists_screen.dart:70-74` skips the relay fetch when `cachedState.lists` is non-empty — an invariant nothing here documents and no test pins.

It's also defeated by two live paths: pull-to-refresh (`_streamPublicLists(isRefresh: true)`) and `onLoadMore` → `addLists()`, which merges real relay lists and re-sorts by `videoEventIds.length` descending. Fixture ids top out at 287, so any real public list with more videos sorts above the fixtures and appears in the capture.

`topClassicVinersProvider` two hunks up uses `overrideWith` and has none of these problems — same approach here (or a `ScreenshotMode.enabled` guard on the screen's fetch paths) would make it structural rather than incidental.

rabble (commented):
Reviewed the Dart wiring, the fastlane config, the Pillow compositor, and the XCUITest. The pipeline hardening is solid and the tree-shaking story checks out. Two things I'd fix before running this for real, plus two product questions that are your call.

## Verified clean

- `ScreenshotMode.enabled` is `kDebugMode && bool.fromEnvironment('SCREENSHOT_MODE')` — a compile-time const, const-`false` in release. Both new fixture builders and the `setLists` seeding sit behind it, so no marketing fixture data (real creator pubkeys, avatar URLs, loop counts) reaches a shipping binary. This was my first concern and it's handled.
- `SemanticIds.categoryTile(index)` instead of an inline `'category_tile_$index'` matches the constants rule in `.claude/rules/accessibility.md`.
- The `wrap()` change (`if cur: lines.append(cur)`) correctly stops emitting an empty leading line when a single word exceeds the wrap width, and an over-long word still lands via the trailing `if cur:`.
- Merges cleanly against current `origin/main` — I checked with `git merge-tree --write-tree` since GitHub is reporting `mergeable: UNKNOWN`. `main.dart` is contended by #6476 and #6478 right now, so it's worth re-checking at merge time, but nothing to do today.

## Should fix: iPad is half-added and has never been run

The Snapfile adds `iPad Pro 13-inch (M5)` and `frame_screenshots.py` adds the `(2064, 2752)` canvas — but that's the whole of the iPad support:

- `DivineScreenshots.swift` has no iPad handling at all (no `userInterfaceIdiom` branch, no iPad-specific waits). The semantic-identifier waits and `app.swipeUp()` loops were tuned against a portrait phone layout.
- `device_body()` is documented as *"A rounded black **iPhone** body with a subtle titanium edge highlight"* and is called for every canvas. So the 13" iPad slide will show an iPad-aspect capture inside an iPhone-shaped body on a 2064×2752 canvas.

And the test plan has `- [ ] Full fastlane screenshot run (requires simulator farm; not run locally)` unchecked, so this path has never executed once. Given the whole point of the PR is producing correct App Store assets, I'd either run the iPad device once before merging, or pull it out of this PR and add it in a follow-up where it can actually be verified. The rest of the PR stands on its own without it.

The `draw_caption` rework already accounts for wider iPad canvases, so the caption band is fine — it's the device body and the UITest that aren't.

## Should fix: the discover-lists seeding is load-bearing on an invisible invariant

`_initializeCoreServices` seeds the fixtures imperatively via `discoveredListsProvider.notifier.setLists(...)`. That only survives because `discover_lists_screen.dart:70-74` skips the relay fetch when the cache is non-empty:

```dart
final cachedState = ref.read(discoveredListsProvider);
if (cachedState.lists.isEmpty) {
  _streamPublicLists();
}
```

So it works today by accident of that guard, and two paths defeat it:

- Pull-to-refresh calls `_streamPublicLists(isRefresh: true)`.
- `onLoadMore` → `_loadMoreLists()` → `addLists()`, which merges real relay lists and re-sorts by `videoEventIds.length` descending. The fixtures top out at 287 video ids, so any real public list with more videos sorts *above* them and lands in the shot — exactly the off-brand names the fixtures exist to avoid.

Compare `topClassicVinersProvider`, which you handled with a real `overrideWith` in `accountOverrides` and is structurally immune. Same treatment for the lists (override the provider, or gate the screen's two fetch entry points on `ScreenshotMode.enabled`) would remove the coupling. If you'd rather keep the imperative seed, the `isEmpty` guard it depends on deserves a comment on both ends, because deleting that guard silently breaks the `06_lists` capture with no test to catch it.

## Question: absolute provenance claims in App Store copy

`keyword.strings` / `title.strings` assert:

- `02_verification`: "Every loop is verified human" / "Signed on capture. No AI — and we can prove it."
- `09_endcard`: "All human. No AI slop."

The voice is right and on-brand. My concern is the absoluteness. Slide `01_classics` is restored archive Vine content that Divine never signed on capture, and ProofMode/proofsign attests *capture provenance*, not *absence of AI* — those aren't the same claim. "Every loop is verified human" as an unqualified statement in store metadata is the kind of thing App Review pushes back on, and it's also the kind of claim that's awkward to defend later.

You own this call and the brand voice guidance is clearly being followed — just flagging that the strongest two lines assert more than the tech proves, before they're locked into a store listing.

## Question: real creators' identities plus stats in the fixtures

`screenshotOgVinersFixtures()` hardcodes four identifiable real people — Reggie Couz, Thomas Sanders, Cptn. Backfire, imrtravis — with their real pubkeys and avatar URLs, alongside `totalLoops` (2.6M, 2.1M, 1.4M, 900K) and `videoCount` values. Two things worth confirming:

1. Are those loop/video numbers their actual archive figures, or illustrative? If illustrative, the App Store screenshot attributes invented stats to named real creators.
2. Are these four fine appearing in paid marketing material?

Worth noting you applied exactly this care one function down — `screenshotDiscoverListsFixtures()` deliberately leaves `pubkey` null *"so no author by-line can surface an off-brand name"*. The viner fixtures didn't get the same pass.

## Nits

- `screenshotDiscoverListsFixtures()` uses `DateTime.now()` for `createdAt`/`updatedAt`, while both the doc comment and the test name call the fixtures "deterministic". Harmless now (nothing asserts on timestamps and no golden consumes them), but a fixed epoch would make the claim true and keep it safe if one ever does.
- `screenshot_mode_service_test.dart`: `isA<dynamic>().having(...)` — `isA<dynamic>()` matches literally anything, so the type half is a no-op. `isA<ClassicViner>()` is stronger and states the intent; the type is already imported in the lib file via `show ClassicViner`.
- `categories_tab_test.dart` hand-rolls `find.byWidgetPredicate` over `Semantics.properties.identifier` twice. `find.bySemanticsIdentifier(SemanticIds.categoryTile(0))` is a one-liner and is what `.claude/rules/accessibility.md` recommends.
- `03_creator_post` has a `keyword.strings` entry but no `title.strings` entry, so it renders headline-only while the other eight get both. `draw_caption`'s `if subhead:` handles it gracefully — just confirming that's deliberate and not a dropped line.
- `clear_previous_screenshots(false)` plus framing driven by the fixed `SLIDES` list means the export dir never self-cleans, so a partially-failed run leaves the previous run's PNGs interleaved with new ones. Low risk given `upload_screenshots` is hard-disabled behind `UI.user_error!` and delivery is a reviewed human step — but the reasoning in the Snapfile comment ("overwrites the same fixed set") stops holding the moment the device or slide list changes, which is what this PR does.
- `DivineScreenshots.swift` still captures the share sheet that `SLIDES` deliberately excludes from the export. The comment says so explicitly, so this is intentional — but it's paid-for time in a ~45-minute run.

NotThatKindOfDrLiz:
@rabble I pushed takeover fixes in `845d45f90b` and updated #6460 / the PR test plan.

What changed:
- skipped the recording-only `testRec*` XCTest methods from the regular snapshot scheme and removed the discarded share-sheet capture
- restored the safer hero video settle
- made framing fail on missing captures/headlines and tightened iPad caption spacing
- refreshed `README_SCREENSHOTS.md` for the current three-device + endcard pipeline
- added fixture coverage and category semantic-id coverage

Verified:
- `flutter analyze`
- `flutter test test/services/screenshot_mode_service_test.dart`
- `flutter test test/widgets/categories_tab_test.dart`
- `python3 -B -m py_compile mobile/ios/fastlane/frame_screenshots.py`
- `xmllint --noout mobile/ios/Runner.xcodeproj/xcshareddata/xcschemes/DivineUITests.xcscheme`

GitHub Mobile CI is green on the current head. Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Takeover fixes look ready. CI is green on the current head, the linked issue is updated, and the remaining fastlane screenshot run is the expected simulator-farm validation.

NotThatKindOfDrLiz:
@rabble I pushed a follow-up on top of the takeover fixes and the PR is green on head `5df0e5117b`.

Addressed from your review:

- Pulled the half-added iPad simulator/canvas/docs from this PR. This now ships the verified iPhone screenshot path; iPad capture is deferred until it can be run end to end.
- Moved discovered-list fixtures into a structural `discoveredListsProvider.overrideWith(ScreenshotDiscoveredLists.new)` at the account-container boundary, including account swaps, and removed the imperative startup seeding.
- Made discover-list fixture timestamps deterministic.
- Tightened the fixture test matcher from `dynamic` to `ClassicViner`.
- Kept the recording tests skipped from normal snapshot runs, restored the hero `waitForVideoLoaded`, removed the discarded share capture, hardened framing failures, updated docs, and added focused tests.

Local verification plus pre-push hook passed:

- `flutter analyze`
- `flutter test test/services/screenshot_mode_service_test.dart`
- `flutter test test/widgets/categories_tab_test.dart`
- `python3 -B -m py_compile mobile/ios/fastlane/frame_screenshots.py`
- `xmllint --noout mobile/ios/Runner.xcodeproj/xcshareddata/xcschemes/DivineUITests.xcscheme`

GitHub checks are all green and the PR is approved/mergeable. The remaining questions around absolute provenance copy and whether these creator stats/identities are the final product choice are yours to decide before merge.

PR #6464, opened 2026-07-28
fix(profile): stop rendering accounts that requested deletion

realmeylisdev:
## Description

An account that completed the in-app NIP-62 request to vanish still rendered its full profile — avatar, display name, handle, bio — on another user's device, plus an inbox conversation row and a face in the inbox following bar. The relay holds zero events for it and the REST API returns `profile: null`.

The server's "this profile does not exist" signal was being discarded:

- `profile_repository.dart:605` cached the stats, recorded `_confirmedMissing`, and returned `null`
- `other_profile_bloc.dart:116` then fell back to the stale cached profile, which made the `notFound` branch unreachable for any pubkey ever viewed
- `deleteCachedProfile` had existed the whole time with **zero production callers**, `user_profiles` has no expiry column, and it is not in `runStartupCleanup`

### 1. `UserProfileVanished` variant

A third variant on the sealed `UserProfileResult` rather than a `deleted` bool on `UserProfileNotPublished`. A bool would sit on the variant that already means "known, no kind 0", and every existing caller reads that as *not yet* rather than *never again* — the eviction, the relay-fallback suppression and the "Deleted account" copy all hang off the difference. The separate variant makes the sealed switch force each of those sites to state which it means, and hard-nulls `social`, `stats` and `engagement` so it is structurally impossible to cache counters for an account that asked to be erased.

**The rule is `has_vanish_request` AND a withheld profile — both halves.** The server reads that flag from `vanish_requests`, the permanent legal-proof record, so it stays `true` after the erasure sweep completes; a completed vanish that publishes again hydrates as a live user still carrying it. Reading the flag alone would hide precisely those accounts, permanently, since it never goes `false` again and `_clearVanish` only fires on `Found` / `NotPublished`. The flag alongside a *served* profile is therefore `Found` — which also self-heals any tombstone recorded earlier.

The two states this PR targets both arrive as flag-plus-null-profile: a pending request has its kind 0 suppressed server-side, and a completed one had it erased.

**The one bulk-path edge, awaiting product confirmation.** Flag-true with `profile: null` out of `POST /api/users/bulk` is currently treated as `Vanished`, same as the single-profile path, because otherwise one account can render as deleted on the profile screen and under a generated npub handle in the inbox depending only on which request resolved first. @rabble flagged this as a product call rather than a correctness one: if completed-vanish-with-post-boundary-activity-but-no-kind-0 should remain live-unknown instead, this is the behavior to flip.

### 2. Durable `vanished_profiles` table

Evicting the profile is not enough on its own: the row is gone, so nothing is left to remember why, and the inbox would fall back to a generated npub name. An in-memory set would be session-scoped and lose that on restart.

Follows the `identity_events` precedent — `schemaVersion` stays at 1, created by the runtime CREATE-IF-NOT-EXISTS path in `_createMissingTables`, so existing installs pick it up on reopen. Tests pin both the upgrade path and schema parity between a fresh install and the runtime path.

Deliberately excluded from `runStartupCleanup` (a vanish does not expire — there is a test) and deliberately **not** in the logout / account-switch fan-out, since it is a global network fact rather than per-viewer state. Clearing it on an account switch would resurrect vanished profiles for the other account.

`detected_at` is first-seen: `markVanished` ignores a re-mark rather than refreshing the timestamp, which matters now that every session re-checks (§4).

### 3. Eviction, and the write path that enforces it

On `UserProfileVanished` the repository deletes the profile row, the stats row, the identity-claims row and the Divine-identity cache entry, and records the pubkey durably. It returns `null` **unconditionally** — including under `requireRawKind0`, which the `NotPublished` branch does not do — so the relay fallback cannot resurrect what was just deleted.

The raw kind 0 stays in `nostr_events`. Nothing renders a profile from there — every read path goes through `user_profiles` — so it is not a surface the account can reappear on.

**The guard lives in `UserProfilesDao`, not at the call sites.** Three writers reach `user_profiles`, and they span layers that cannot all reach `ProfileRepository`:

| Writer | Path |
|---|---|
| `ProfileRepository.cacheProfile` / `_cacheProfileIfNewer` | REST + relay caching |
| `EventRouter._handleProfileEvent` | every kind 0 on every relay subscription, written **straight to the DAO** |
| `ClassicVinerSeedPreloadService` | bundled seed, re-imported on every manifest bump |

An earlier revision of this PR asserted that everything funnelled through `cacheProfile`. It does not — `EventRouter` bypasses the repository entirely, which left a non-compliant relay free to re-create the row minutes after eviction, with nothing to re-evict it afterwards. Routing `EventRouter` through the repository would invert the layering; mirroring the vanished set into a second in-memory copy would add a cache to keep in sync. A tombstone lookup on a primary key, against a table that is empty for practically every install, is cheaper than either. `ProfileStatsDao.upsertStats` carries the same rule so the seed import cannot re-cache counters for a vanished classic viner.

The repository keeps its own in-memory guards as the fast path — they short-circuit before the DB and before the `_knownCached` bookkeeping.

**Self-heal:** any 200 that is not a vanish clears the marker, so a wrong vanish classification is recoverable rather than erasing the account from the device permanently. That durable delete is not gated on the repository's in-memory mirror. The mirror is one set per repository instance, hydrated asynchronously, over a table every instance shares, so a local miss does not prove the row is absent — trusting one would strand the tombstone, and the DAO would then drop every re-cache of the recovered account while `cacheProfile` had already claimed the pubkey in `_knownCached`, leaving `hasProfile` asserting a row that does not exist.

### 4. One request per vanished pubkey per session

Eviction leaves no cached row, so `getCachedProfile` returns `null` forever and every cache-miss caller took the network branch. Three `autoDispose` providers re-instantiate per widget, and each call re-ran the whole eviction — a `markVanished` upsert plus three no-op deletes, behind a real REST request. `fetchBatchProfiles` had the same shape: a vanished pubkey stayed in `remaining` and rode in every prefetch.

Both now answer from the durable marker, and `revalidateVanishOnce` is the single channel back to the network. It no longer skips an already-vanished pubkey — that skip made recovery from a wrong vanish classification work only in the session that recorded it and never after a restart, contradicting the property the self-heal exists to provide. `_applyVanish` marks the pubkey revalidated, because the response that discovered the vanish *is* that session's re-check.

Net cadence: **one request per vanished pubkey per session, in every session.** Recovery from a wrong flag costs one app restart rather than being permanent.

`fetchUserProfile`'s cache-first read no longer nudges a revalidation (e932390d7, @NotThatKindOfDrLiz) — a cached read stays local. The already-vanished short-circuit inside `fetchFreshProfile` and `fetchBatchProfiles` is now the only channel back to the network.

### 5. Bloc gates

`isVanished` is consulted before the cache read, before the stale-cache fallback, and in the catch block. Refresh also passes `profile: null` — `other_profile_screen.dart` extracts `profile` out of the error state and renders it, so without that, pull-to-refresh would keep the deleted account on screen forever.

### 6. Deleted counterparty in the inbox, and on the profile screen

The DM thread and its history **stay**. Those messages are the viewer's own copy, and NIP-17 gift wraps are authored by ephemeral keys p-tagged to the recipient, so a sender's vanish can neither identify nor retract them — the same semantics as Signal or WhatsApp. Only the counterparty's name, handle and avatar change, in the inbox row, the conversation header and the following bar.

The profile screen now names the same state. It previously never branched on the vanish at all — it switches on `ProfileFeedCubit.status` — so a deleted account rendered a generic empty "Profile" with `displayNameHint` / `avatarUrlHint` still passed through; via `classic_viners_slider.dart` a vanished classic viner kept showing under their archived Vine name and avatar. The guard sits in `ProfileHeaderWidget` rather than at the screen, because the header resolves identity from three independent sources — the profile bloc, `fetchUserProfileProvider`, and the route hints. Nulling the bloc profile at the screen only sends it to the provider, and the hints come from whoever navigated here, so eviction cannot reach them at all. The name is rendered directly rather than through `UserName`, which re-resolves the profile through its own provider and falls back to a generated handle — for a deleted account that reads as a real handle rather than as an absence.

Driven by a new `profileVanished` stream over the durable table rather than by the profile provider, so it resolves on a cold start with no network round trip.

Two ARB keys, translated into all 17 non-English locales rather than deferred, so `_knownUntranslatedDebt` is untouched.

**Related Issue:** Relates to #6127

## Out of Scope

- **`follow_repository` vanish guards.** `getFollowerStats` / `getFollowerCount` / `FollowerStats` have **zero production callers** — verified repo-wide. The rendered counts come from `userProfileStatsReactiveProvider` → `watchProfileStats`, whose only live writer is `_cacheProfileStatsFromResult`, which this PR already fixes. Wiring a check in would add a constructor dependency and touch a 5,000-line test file to fix latent code. Worth a follow-up.
- **Deleted-account copy on every surface.** Comments, people lists, engagement lists, `original_sound_detail_screen` and `safety_settings_screen` render a vanished pubkey the way they render any unknown one — a generated fallback name, no avatar — now that the row cannot be re-created. Giving each of those explicit "Deleted account" copy is a product call, not a correctness one. The same applies to the profile screen's own block / unblock / unfollow snackbars, which still resolve the name through the route hint: they are transient confirmations of an action the user just took, not a rendering of the account.
- **DM message history**, per the above — kept by design.
- **Other users' kind-3 lists** still naming the vanished pubkey.

## Verification

| Check | Result |
|---|---|
| `flutter test` (full suite) | 13,535 passed, 0 failed |
| `flutter analyze lib test` | clean (app, `db_client`, `profile_repository`) |
| `dart format --set-exit-if-changed` | clean |
| `flutter test test/l10n/arb_consistency_test.dart` | passed |
| `packages/db_client` drift `migration_test.dart` + `app_database_test.dart` | passed (v1 schema unchanged) |
| 4 design-system ceilings + untested-services floor + test-unit structure | all pass |

Re-verified for the `_clearVanish` hardening: `profile_repository` 340 passed with 100% line coverage held, package analyze and format clean, and the six affected app suites green. The paired test fails on the previous code with `clearVanished` never called.

Re-verified for the field reconciliation (`812bc25f7`): `funnelcake_api_client` 365 passed, `models` 827, `profile_repository` 340 with 100% line coverage held, `db_client` 898, and the six affected app suites 132 — all green. `flutter analyze lib test integration_test` and all four package analyzes clean, format clean. Both new "flag alongside a served profile is live" tests were confirmed to fail against a blind key swap; the full suite has not been re-run for this commit, since the change is confined to the two parse sites plus doc text.

Codegen run and committed: `build_runner` for `db_client` and the app, `drift_dev make-migrations`, and `flutter gen-l10n`. The DAO accessor change is schema-neutral — `schemaVersion` stays at 1 and the v1 snapshot is unchanged.

Highest-value tests, each able to fail for a real reason:

- **Parsing:** `has_vanish_request: true` **with a populated profile** ⇒ `Found`, not `Vanished` — the test that fails on a blind key swap, on both the single and the bulk path, and the one standing between a republished account and permanent erasure on every device. Flag + `profile: null` ⇒ `Vanished`. `false` + `profile: null` ⇒ `NotPublished`, guarding against mislabelling never-published users. Absent field ⇒ `NotPublished`, so this ships safely ahead of the server.
- **Ingestion:** an `EventRouter` batch carrying a vanished account's kind 0 leaves `user_profiles` empty, and routes normally again once the tombstone is cleared. This is the regression that was live on the previous revision.
- **Cadence:** five repeat `fetchFreshProfile` calls on a vanished pubkey cost **one** REST request, one `markVanished` and one delete — fails if the short-circuit or the `_applyVanish` bookkeeping regresses. A marker hydrated from the durable table (a restart) re-checks once and self-heals.
- **Bloc:** cached profile + null fetch + `isVanished true` ⇒ `notFound`; **the same setup with `isVanished false` ⇒ `Loaded(isFresh: false)`**, which is the test that fails if the gate is widened to every null fetch.
- **Screen:** a vanished pubkey renders "Deleted account" and neither the route hint nor the bloc's profile name; the paired not-vanished case renders the profile normally.

## Dependency

Reads `has_vanish_request`, shipped by divinevideo/divine-funnelcake#720 — merged 2026-07-29 but **not yet deployed to production** (last `deploy-production.yml` run: 2026-07-27). An absent field reads as no vanish request, so this stays safe against any server that predates it — but it also means a build of this branch renders a vanished account exactly as `main` does until that deploy runs. Verifying on device is gated on it.

That PR renamed the field mid-flight (`deleted` → `has_vanish_request`, takeover commit `0770918a`, eight minutes after this branch's previous head) and gave it historical rather than tombstone semantics. Caught by @rabble; the reconciliation above is the fix.

## Type of Change


mbradley (changes requested):
The design holds up — the `UserProfileVanished` variant, deletion-wins precedence, the durable table, self-heal, and the paired `isVanished true`/`false` bloc test that actually pins the gate. Two things block, and three need your call.

## The third resurrection path is still open

`video_event_service.dart:2463` routes *all* subscription events into `EventRouter`, five lines above the guarded `cacheProfile` call. `event_router.dart:278` coalesces kind 0 per pubkey and `:325` writes `userProfilesDao.upsertProfile` straight to the DAO, bypassing `ProfileRepository` entirely. So the scenario the guard was written for — a vanished account whose kind 0 still lives on a non-compliant relay — still re-creates the row. `7d98d614b`'s commit body states the opposite ("every Kind 0 seen on any relay subscription funnels through cacheProfile"), and that claim is load-bearing for the whole eviction design.

It doesn't self-correct either: `revalidateVanishOnce` returns early once `_vanished` holds the pubkey, so nothing re-evicts and the row survives restarts. The three surfaces this PR gates stay correct; every ungated `userProfileReactiveProvider` / `fetchUserProfileProvider` consumer — comments, people lists, engagement lists, `original_sound_detail_screen`, `safety_settings_screen` — serves the deleted account's name and avatar again.

## Eviction turns every vanished account into a permanent cache miss

`getCachedProfile` returns null forever once the row is deleted, and `fetchFreshProfile` (`:594`) has no vanish short-circuit — the `_confirmedMissing` sentinel `_applyVanish` sets is cleared on entry at `:606` and gates nothing anywhere (`isConfirmedMissing` has no production callers). `user_profile_providers.dart:94`, `_watchUserProfile`, and `_watchProfileStats` therefore each fire a real REST request per provider instantiation, and every one re-runs `_applyVanish`: a `markVanished` upsert plus three no-op deletes.

This is exactly the cost `revalidateVanishOnce`'s own doc bounds — "an inbox scroll would otherwise fire one request per row per scroll." That bound landed on the cached branch; the null branch two lines below is the one every vanished account permanently takes.

## Three calls that are yours

**Where the kind-0 write guard lives.** Either guard `_handleProfileEvent` against `_db.vanishedProfilesDao` — `EventRouter` already holds `AppDatabase`, and mirroring the in-memory set avoids a per-event await — or route kind 0 through `ProfileRepository.cacheProfile` so there's one write path with one guard. The second is stronger long-term; the first is the smaller diff. Whichever you pick also settles whether `loadVanishedPubkeys()`'s unawaited fire at `repository_providers.dart:373` needs to gate the first write, or whether that startup window is acceptable — it matches the existing `loadKnownCachedPubkeys` precedent.

**The self-heal cadence.** A `_vanished` short-circuit in `fetchFreshProfile` fixes the refetch storm, but only if `revalidateVanishOnce` may re-check *even when already vanished*. Otherwise recovery from a wrong `deleted: true` works only in the session that recorded it and never after a restart, which contradicts the property you documented. One re-check per pubkey per session, or something else?

**Whether the profile screen says anything.** The bloc emits `notFound`, but `other_profile_screen.dart` never branches on the bloc's error state — it switches on `ProfileFeedCubit.status`, so a vanished account renders a generic empty "Profile" with `displayNameHint` / `avatarUrlHint` still passed through (`:445-448`, `:486-487`). Via `classic_viners_slider.dart:264` a vanished classic viner still shows under their archived Vine name and avatar. The inbox got explicit copy for this state — should the profile screen match, or is the split deliberate here?

## Nits, each with a resolution

- `tables.dart:1602` documents `detected_at` as "when this device first learned," but `markVanished` upserts a fresh `DateTime.now()` on every re-detection. Once the cadence question lands, either preserve the first-seen value or correct the doc — `code_style.md` treats a comment contradicting the code as a defect.
- The two new keys land at 2-space indent inside 17 locale files that use 4-space (`app_de.arb:3269`, `app_ko.arb:2593`, …). `app_en.arb` is correct. Re-indent to 4; the consistency test doesn't catch it.
- The `async*` → sync-stream conversion of `userProfileReactive` and `userProfileStatsReactiveProvider` is bundled into `7d98d614b` and unrelated to vanish. It reads as a real improvement in the spirit of #6405, but it changes when the `ref.watch` edge registers in two widely-used providers, with no test and no mention in the description. Split it into its own commit and say so, or drop it.
- `_applyVanish` leaves the raw kind 0 — name, about, picture — in `nostr_events` via `_persistRawEvents`, so "evicts every local trace" isn't literally true. Nothing in `lib/` reads kind 0 back from there to render, so accepting it is fine; scope the doc comment rather than leaving the claim absolute.
- `_vanishRevalidated` grows unbounded per session. Consistent with `_knownCached`, so accepting it is right — noted so it's a decision rather than an oversight.

---

Verified against `d5d64ff2b` in a local worktree, with the server contract cross-checked against divinevideo/divine-funnelcake#720 (the `deleted` flag on both the user endpoint and `get_users_bulk`, sourced from `vanish_requests` and failing closed to `false`, matches what this PR parses). Ran `profile_repository` (235 passed) and the affected mobile bloc, inbox widget, provider and screen suites (106 passed); all 18 CI checks are green on this head.

Feedback only — no commits pushed to your branch. All three open questions turn on judgment that is yours rather than mine, so I've left them for you rather than implementing a direction. Happy to take the mechanical nits off your plate if useful.

Thanks, as always.

realmeylisdev:
@mbradley Both blockers fixed and all three calls made. Full suite green (13,535), analyze and format clean across the app and both packages, all six ratchets pass.

**The third resurrection path** — `68c2a97687`. You're right, and the commit body you quoted was wrong: `EventRouter` never touches `ProfileRepository` at all. I took neither option exactly. The guard is in `UserProfilesDao.upsertProfile` / `upsertProfiles`, where every production writer actually converges — including a fourth I hadn't counted, `ClassicVinerSeedPreloadService`, which re-imports on every manifest bump and would have resurrected exactly the classic viners you flagged. `ProfileStatsDao.upsertStats` carries the same rule so the seed cannot re-cache their counters either. Routing `EventRouter` through the repository would have inverted the layering; the in-memory mirror would have been a second copy to keep in sync. This also dissolves the `loadVanishedPubkeys()` question rather than answering it: the DAO reads the table directly, so the unawaited startup warm-up no longer gates any write.

**The refetch storm** — `1190c60748`. `fetchFreshProfile` now answers a vanished pubkey from the durable marker. `fetchBatchProfiles` had the same shape and is fixed with it — the pubkey stayed in `remaining` and rode in every prefetch.

**Self-heal cadence:** one request per vanished pubkey per session, in *every* session. `revalidateVanishOnce` no longer skips an already-vanished pubkey and calls the unguarded fetch; `_applyVanish` marks the pubkey revalidated, because the response that discovered the vanish *is* that session's re-check — otherwise the first read after discovery immediately spends a second request to hear the same thing. Recovery from a wrong `deleted: true` costs one app restart instead of being permanent. Two tests pin it: five repeat fetches cost one REST call and one eviction, and a marker hydrated from the table re-checks once and clears.

**The profile screen** — `8352f29de0`. It should match; the split was an oversight, not a decision. The guard had to go in `ProfileHeaderWidget` rather than the screen: the header resolves identity from three independent sources, so nulling the bloc profile only pushes it to `fetchUserProfileProvider`, and the route hints never pass through the cache at all. Rendered directly rather than through `UserName`, which re-resolves via its own provider and falls back to a generated handle — for a deleted account that reads as a real handle rather than as an absence. The block/unblock/unfollow snackbars still resolve the name through the hint; I left those as transient confirmations of an action rather than a rendering of the account, and said so in the description.

**Nits.** `detected_at` now preserves first-seen — `markVanished` ignores a re-mark, which matters more once every session re-checks. ARB keys re-indented to 4 across all 17 locales in `fa66ea48d7`; `profileUserFallback` had been dragged down to 2 with them. The raw-kind-0 doc is scoped instead of claiming absolute eviction. `_vanishRevalidated` unbounded — agreed, keeping it.

One correction: the `async*` conversion isn't in this diff. `userProfileReactive` and `userProfileStatsReactiveProvider` already delegate to `async*` private helpers on the merge-base, `b792cb32cf` — the shape landed in #6405. `7d98d614b` touches that file only to add the `revalidateVanishOnce` call in `fetchUserProfile` and the new `profileVanished` provider. Nothing to split.

Description rewritten to match. Thanks — the write-path question was the right one to push on.

mbradley (approved):
Both blockers are closed, and the DAO-level guard is a better answer than either option I offered — `UserProfilesDao` is where the writers actually converge, and it caught a fourth one I'd missed (`ClassicVinerSeedPreloadService` re-importing on manifest bump, which would have resurrected exactly the classic viners in question). Putting a domain rule in the data layer is a deviation from `architecture.md`, but it reads as a data-integrity constraint rather than business logic, and the alternative inverts the layering — right call, and the doc comment carries the why.

The guarded/unguarded `fetchFreshProfile` split is clean, and pre-marking `_vanishRevalidated` inside `_applyVanish` so the discovering response counts as that session's check is a nice touch — it avoids spending a second request to hear the same thing.

**I was wrong about the `async*` conversion.** I diffed against a stale `origin/main` twenty commits behind, which folded main's own changes into what looked like your diff. Against the real merge-base `b792cb32cf` the file carries only the `revalidateVanishOnce` call and the new provider, exactly as you said. Thanks for checking rather than just deleting the code.

## One residual, non-blocking

`_clearVanish` gates the shared-table delete on the per-instance in-memory set:

```dart
if (!_vanished.remove(pubkey)) return;
```

There are two live `ProfileRepository` instances — `profileRepositoryProvider` and `profileStatsRepositoryProvider`, both `keepAlive` — each with its own `_vanished`, hydrated only at construction. When the stats instance discovers a vanish, the main instance's set never learns. If the account is later restored and the *main* instance fetches first, `_clearVanish` early-returns without deleting the tombstone, and the DAO then silently drops the re-cache. The `loadVanishedPubkeys()` startup race produces the same shape.

The observable outcome stays inside the contract you documented — recovery costs one restart, because next session both sets hydrate from the table and the re-check clears it. So this isn't a behaviour bug. It's that the in-memory mirror now gates a write to the shared table the DAO reads, and the two can disagree. Minor side effect while they do: `cacheProfile` adds to `_knownCached` before the dropped DAO write, so `hasProfile` reports true with no row, and `SubscriptionManager` skips relay kind-0 requests for a pubkey it believes is cached.

Cheapest hardening is to stop gating the durable delete on the mirror — drop the early return, or fall through to a DAO check when the set misses. Your call whether that's worth an extra write per `Found`; the current behaviour is defensible and documented, so I'm not holding the PR on it.

---

Verified on `202a9181e`: re-ran `profile_repository` (339 passed) and the affected mobile suites including the new `event_router_test.dart` (472 passed); all 19 CI checks green. Spot-checked each claim in your reply — `AppDbClient.upsertProfile` has only test callers, `markVanished` is `insertOrIgnore` so `detected_at` really is first-seen, the batch path guards `remaining`, and the ARB keys are back at 4-space indent across all 17 locales with `profileUserFallback` restored alongside them.

Good work on the write-path convergence — that's a stronger fix than what I asked for.

Thanks, as always.

realmeylisdev:
Took the hardening. `_clearVanish` no longer gates the durable delete on the in-memory mirror, so the two instances can no longer disagree about a tombstone the table still holds — and the same fix covers the `loadVanishedPubkeys` startup window you noted has the same shape. Paired test next to the existing self-heal one: identical setup minus the hydration, which fails on the previous code with `clearVanished` never called.

Also scoped the matching stale claim on `cacheProfile`'s doc — it still said every relay kind 0 funnels through there, which is the sentence you caught on the commit body. And `revalidateVanishOnce`'s doc lost its first bullet: e932390d7 took the revalidation off the cached-read path, so that caller no longer exists.

realmeylisdev:
@NotThatKindOfDrLiz — `e932390d7` looks right for what it targets: an inbox scroll was firing one REST call per row per session. But it also removes the only channel that *discovers* a vanish for an account cached **before** it asked to be deleted. Every path that can reach REST for a cached pubkey:

| Path | Behaviour for an already-cached pubkey |
|---|---|
| `fetchUserProfileProvider` — `conversation_tile`, `following_bar` | returns cached; the nudge is gone |
| `userProfileReactiveProvider` → `_watchUserProfile` | fetches only when `cached == null` |
| `fetchBatchProfiles` — `conversation_list_bloc`, feed, classic vines | drops cached pubkeys from `remaining` before the bulk request |
| `other_profile_bloc` | fetches fresh unconditionally |

So the profile screen still resolves a deleted account, but the inbox row and the following bar keep the cached name and avatar until someone opens that profile — which is the surface §6 was written for.

This isn't a regression against `main` (main has no vanish handling at all), so I don't read it as a merge blocker — but which way it should land is your call, not mine to revert. One option that keeps per-row reads local: let `fetchBatchProfiles` carry cached pubkeys in the bulk request it already sends, bounded once per session, so discovery costs no extra round trips. Happy to implement whichever you prefer, or to leave it as a known limitation.

Separately, `revalidateVanishOnce`'s doc had a bullet describing the caller you removed — dropped in `d94da6f1b7`.

Branch is at `d94da6f1b7`, 19/19 green. I'm holding the merge for you, @rabble and @dcadenas, since the approval predates your commit and my three.

rabble (commented):
Ordinary review feedback. One blocking issue, introduced cross-repo by a mid-flight change on the companion server PR after this branch's last push.

## The server renamed the field this PR parses

The companion change divinevideo/divine-funnelcake#720 renamed the public wire field in takeover commit `0770918a` (2026-07-28 16:23Z), about eight minutes after this branch's head (`d94da6f1b`, 16:15Z). `UserData.deleted` is now **`has_vanish_request`**. Both parse sites here still read the old key:

- [`getUserProfile`](https://github.com/divinevideo/divine-mobile/blob/d94da6f1b7a622c8879bb1a5cc0fb74350947965/mobile/packages/funnelcake_api_client/lib/src/funnelcake_api_client.dart#L1756) — `if (data['deleted'] == true)`
- [`getBulkProfiles`](https://github.com/divinevideo/divine-mobile/blob/d94da6f1b7a622c8879bb1a5cc0fb74350947965/mobile/packages/funnelcake_api_client/lib/src/funnelcake_api_client.dart#L2146) — `if (user['deleted'] == true)`

plus the four parser tests and the surrounding comments. Absent reads as not-deleted, so this fails safe — but the entire feature silently never activates against the renamed server, and the bug this PR fixes stays live on every surface it gates.

## It is a semantic change, not just a rename — a blind key swap is the wrong fix

`has_vanish_request` is deliberately documented server-side as **historical** erasure-request state, sticky-true after completion: a completed vanish can rehydrate with post-boundary activity and still report `has_vanish_request: true`, and both the server PR body and the `UserData` doc comment say clients must not hide an otherwise live profile solely because the flag is true. The server's own integration test (`completed_vanish_reports_vanish_request_through_bulk_hydration`) pins exactly that state.

This client's rule is currently **"deletion wins over profile presence."** Under the new semantics that permanently hides precisely the accounts the canonical API serves as live — and the self-heal can never recover them: `_clearVanish` only fires on `Found`/`NotPublished`, which a sticky-true flag prevents forever, while the tombstone plus the DAO write guard drop every subsequent re-cache. One wrong classification becomes unrecoverable by design.

A reconciliation that preserves this PR's intent against every state the new server can actually produce:

- `has_vanish_request: true` **and** `profile: null` → `UserProfileVanished`. Pending vanish: the profile is suppressed server-side. Completed vanish without post-boundary activity: the profile is gone. Both render as deleted, exactly as designed here.
- `has_vanish_request: true` **with a populated profile** → `UserProfileFound`. That is a completed vanish that rehydrated; the server explicitly serves it as live. This branch also naturally runs `_clearVanish`, so a stale tombstone self-heals instead of locking in.
- Absent field → previous behavior, unchanged.

The "pre-erasure kind 0 still being served" window the current precedence rule was built for can't occur through this server anymore: a pending vanish already gets `profile: null` (pre-existing suppression), and a completed one can only serve a *post*-boundary kind 0, which is a live profile by the server's contract.

One bulk-path edge worth an explicit joint decision: flag-true + profile-null out of `getBulkProfiles` is a completed vanish with post-boundary *activity* but no new kind 0. Tombstoning there is defensible, but it is a product call, not a correctness one.

The PR body's Dependency section also names the field `deleted` and claims merge order no longer matters — both need updating with the fix.

## Nothing else from me

The eviction design holds up: the DAO-level tombstone covering `EventRouter` and the seed import, the one-request-per-session bound with `_applyVanish` pre-marking the discovering response, and the `8d9169220` hardening that stops `_clearVanish` gating the shared-table delete on a per-instance mirror — that last one correctly closes the two-instance divergence mbradley noted, including the startup-hydration window. The paired `isVanished` true/false bloc tests are the right shape. I'm leaving the `e932390d7` discovery-channel question to @NotThatKindOfDrLiz as already queued in the thread above.

🤖 Agent review posted on behalf of @rabble via [OpenCode](https://opencode.ai)

realmeylisdev:
@rabble Confirmed and fixed in `812bc25f7`, and you were right that the blind key swap would have been the wrong fix.

I checked the merged funnelcake `origin/main` rather than the PR description, since #720 landed at 04:28Z today. Everything you described is in the shipped code: the `UserData` doc comment on `queries.rs:1675`, `is_pubkey_vanished` / `get_vanished_pubkeys` in `vanish.rs` reading `vanish_requests` rather than the pending-only `vanished_pubkeys_set`, and the note above the bulk flag lookup in `client.rs` spelling out that a completed vanish keeps the flag once post-boundary activity re-opens the existence gate.

Two things I could pin from that read that sharpen the reconciliation you proposed:

**A pending request already nulls the profile server-side.** `get_user_profile` carries `p.pubkey NOT IN vanished_pubkeys_set`. So flag-true-with-a-served-profile is *only* ever a completed vanish that republished — never the pre-erasure window the old precedence rule was built for. That window can't occur through this server at all, which is why I took your matrix as-is rather than keeping any form of "deletion wins": both halves of the rule are now required, and a served profile alongside the flag is `Found` and clears the tombstone. The variant's doc, the commit rationale and the PR body's §1 all said the opposite and have been rewritten.

**The bulk-path edge — I took the uniform rule, tombstone included.** It's narrower than it first looks: the handler strips a pending request through `get_hidden_pubkeys` before hydration and the existence gate closes on a completed one with nothing left, so flag-true out of `/api/users/bulk` can only be a completed vanish with post-boundary activity and no republished kind 0. Treating that as anything other than `Vanished` would mean the same account renders as deleted on the profile screen and under a generated npub handle in the inbox depending only on which request resolved it first — the split §6 exists to close. It's also free to reverse: a new kind 0 flips it to `Found` on the next session's re-check. Flagged in the body as the product call it is; @NotThatKindOfDrLiz, happy to invert it if you'd rather.

One consequence worth naming for the discovery-channel thread you left with Liz: because bulk now only surfaces a vanish for that one narrow completed-plus-activity state, the bulk path is even less of a discovery channel for the inbox than it was. It doesn't change my read that this isn't a merge blocker — `main` has no vanish handling at all — but it does make the "carry cached pubkeys in the bulk request" option the more useful of the two.

Tests: the four parser cases are rewritten against the real key and the fifth is inverted — flag plus a served profile must be `Found`. That one, plus its new bulk twin, fails against a straight key swap on both paths; I checked rather than assumed. Bulk had no vanish coverage at all before, so both bulk cases are new. `funnelcake_api_client` 365, `models` 827, `profile_repository` 340 with 100% line coverage held, `db_client` 898, and the six affected app suites 132 — all green, analyze and format clean across the app and all four packages. Full suite not re-run for this commit; the change is confined to the two parse sites plus doc text.

The Dependency section now names the right field and records that #720 is merged.

realmeylisdev:
Field testing today surfaced that a deleted account still renders normally — full profile, Follow button, no "Deleted account" anywhere. That is **not** a defect in this PR or in the funnelcake one, and it needs no code change here, but it does mean this branch is currently untestable end to end.

divinevideo/divine-funnelcake#720 is merged but **not deployed**. Verified against prod just now:

```
$ curl https://api.divine.video/api/users/613cb00feb0ecd7648a62740771cf60efaead56a0d19b2ac3d7b4f3c589e6fa3
{"pubkey":"613cb00f…","profile":null,
 "social":{"follower_count":4,"following_count":4},
 "stats":{"reaction_count":13,"comment_count":1,"total_events":90,…}}

$ curl …/followers → still returns all 4 follower pubkeys
```

No `has_vanish_request` field, and the social/stats/followers leak it fixes is still present — byte-identical to the measurements in that PR's description. The last `deploy-production.yml` run was 2026-07-27, before it merged on 2026-07-29.

So a build of this branch will show a vanished account exactly as `main` does: the client never receives the signal, and by design an absent field reads as no vanish request. Worth stating plainly because the symptom is indistinguishable from this PR not working.

One thing the field report makes concrete, and an argument for the §6 scope: the tester's deleted account had never published a kind 0, so it rendered as `Delicate Harrier 90` + truncated npub + Follow button. That is the *generic no-profile* rendering — a brand-new account that never set one up looks identical. Distinguishing the two is the whole reason the client needs `has_vanish_request` rather than inferring deletion from `profile: null`.

Also confirmed unrelated to relay propagation: the relay purge is complete and immediate — zero events at any kind for that pubkey, including the kind-62 request itself.

No action needed on this PR. It needs a production deploy of funnelcake to verify on device.

hm21 (approved):
LGTM

I tested it and was able to delete the account without any problems. The only issue was that my account didn't show up as deleted to others. However, @realmeylisdev  confirmed that the backend fix has already been merged and just not released, so we can safely merge it now.

PR #6465, opened 2026-07-28
fix(stop-motion): steady the first editor open after a capture

### LINKED ISSUE #6466: fix(stop-motion): steady the first editor open after a capture
Auto-created from PR #6465.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6465
**Branch:** `fix/stop-motion-editor-handoff`

hm21:
Closes #6466

## Description

Recording a handful of stop-motion stills and tapping Next produced two things video recordings don't have: the editor preview flickering, and a loading overlay flashing on the way in. Both are fixed here.

**Short sessions now play for at least a second.** Freshly captured stills were held for `defaultFramesPerImage` (one output frame each), so three photos played in an eighth of a second — the editor loop read as a flicker, not a clip. `StopMotionFrameOps.initialFramesPerImage` now derives a shared hold from the number of stills so a new session fills at least `minimumInitialDuration` (a lone still is held for the full second, i.e. 24 frames at the render frame rate). It is a starting point, not a floor: the frames-per-image control stays fully editable, and nothing re-stretches a clip on later opens, so a deliberate choice is never overridden. The hold is computed after unreadable stills are filtered out, so it counts only what actually lands in the clip.

**The assemble no longer blocks the handoff.** Tapping Next showed a "Creating your video…" progress overlay until `StopMotionStatus.ready`. That overlay dates from when assemble encoded an mp4; today the stills are the source of truth until publish, and the only remaining wait was a library save for a row that capture already upserts on every still. That save now goes through the existing `_enqueueStopMotionSessionWrite` queue (so it stays ordered behind the capture-time writes) instead of being awaited, the assemble handler runs synchronously, and the overlay it fed is deleted along with its now-unused l10n key. `StopMotionStatus.assembling` itself stays — it still gates the mode wheel and drops stills captured mid-assemble — it just no longer survives a frame, so nothing can paint on it.

**The editor's setup placeholder decodes bounded.** `VideoEditorThumbnail` rendered `clip.thumbnailPath` with no `cacheHeight`. For a stop-motion clip that path is a raw full-resolution camera still, decoded at full size and then decoded again moments later by `StopMotionPlayer` at its own height. It is now bounded to the same height the player uses, so the two share one image-cache entry.

**Related Issue:** 

## Out of Scope

Stop-motion clips recorded before this change keep their old short hold when opened from the library. Stretching those on open would mean re-writing a hold the user may have chosen deliberately — a global frames-per-image change isn't recorded as an override, so there's no way to tell the two apart.

## Verification

- `flutter test test/models/stop_motion/stop_motion_frame_ops_test.dart test/blocs/video_recorder/video_recorder_bloc_test.dart test/widgets/video_recorder test/widgets/video_editor/main_editor test/l10n/arb_consistency_test.dart` — 514 tests, all passing
- New coverage: the minimum-duration invariant across session sizes in `stop_motion_frame_ops_test.dart`, and an ingest test pinning that a single-still session reaches the clip manager at the minimum duration
- Manual on a real Samsung Galaxy: recorded 1, 3 and many stills, opened the editor from each, checked the preview plays instead of flickering, that no overlay flashes on the way in, and that the session still shows up correctly in the library

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -1509,14 +1509,10 @@ class VideoRecorderBloc
     VideoRecorderMode mode, {
     required bool keepAutosavedDraft,
   }) {
-    // An assemble owns the captured stills until it resolves: it saves them to
-    // the library across an await, while this method deletes them and clears
-    // the clip manager. Switching mode mid-assemble would delete the stills out
-    // from under the save and leave the library row pointing at missing files,
-    // so the switch is dropped instead. The mode wheel is already gated while
-    // assembling; this keeps the invariant if anything else ever dispatches.
-    if (state.stopMotionStatus == StopMotionStatus.assembling) return;
-
+    // No assembling-status guard: the assemble is synchronous, so a mode switch
+    // cannot land inside it, and the `ready` it leaves behind already carries an
+    // empty frame list — so the discard below is a no-op and the queued library
+    // write has nothing to lose.
     final previousMode = state.recorderMode;
     final previousFrames = state.stopMotionFrames;
     emit(
@@ -1822,6 +1818,10 @@ class VideoRecorderBloc
     final frames = state.stopMotionFrames;
     if (frames.isEmpty) return;
 
+    // Never painted (the handler is synchronous, so this is superseded before
+    // the next frame) — but required: without a distinct state in between, a
+    // second assemble of the same failing frame list emits a state equal to the
+    // current one, which `emit` drops, and the failure snackbar never re-fires.
     emit(state.copyWith(stopMotionStatus: StopMotionStatus.assembling));
 
     try {
--- a/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
+++ b/mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
@@ -1641,16 +1641,14 @@ class VideoRecorderBloc
       return;
     }
 
-    // An assemble owns the captured stills once it starts: it reads the frame
-    // list, saves it, then clears it. A still added to that list mid-assemble
-    // is either missed by the save or re-persisted afterwards as a phantom
-    // one-frame session, so the capture is dropped instead. `ready` (assemble
-    // done, navigation pending) is dropped here too so a tap in that ~1-frame
-    // window skips the native capture the after-await guard would only delete.
-    if (state.stopMotionStatus == StopMotionStatus.assembling ||
-        state.stopMotionStatus == StopMotionStatus.ready) {
-      return;
-    }
+    // An assemble hands the captured stills to the clip manager and clears
+    // them; a still appended after that would be re-persisted as a phantom
+    // one-frame session. `ready` (assemble done, navigation pending) is what
+    // catches it — no assembling-status check, because the assemble is
+    // synchronous, so it is never the current state while this runs. Dropping
+    // the tap here also skips a native capture the after-await guard below
+    // would only delete.
+    if (state.stopMotionStatus == StopMotionStatus.ready) return;
 
     unawaited(HapticService.recordingFeedback());
     // Shutter feedback must fire NOW — the native capture below takes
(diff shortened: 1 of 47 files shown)


mobile/lib/models/stop_motion/stop_motion_frame_ops.dart
```diff
@@ -26,6 +27,49 @@ abstract class StopMotionFrameOps {
   /// Default hold for a freshly captured still: 1 output frame at 24fps.
   static const int defaultFramesPerImage = 1;
```
realmeylisdev, line 35: nit — this doc is stale after the change. No freshly captured still is held for `defaultFramesPerImage` any more; `initialFramesPerImage` decides the hold. It now has zero callers outside this file, surviving as the clamp floor just below and as the fallback in `globalDefaultFramesPerImage`. Worth saying that instead, so nobody later wires a new capture path to it thinking it's still the capture default.

The math itself is right: I enumerated 1–40 stills and every count lands at or above the gate, and the `+1` branch fires at exactly 3 and 12 as the commit message says. Reverting the `+1` reddens `a session of any size fills at least the minimum duration` with `3 still(s) fall short of the minimum`, so the new test is load-bearing.

One design note, not a change request: the root cause of the 1µs/4µs shortfall is `framesPerImageToDuration` rounding to nearest rather than up. Correcting it there would drop the `+1` entirely, but it would also shift every persisted hold, so the localized fix is the right call for this PR.

mobile/lib/blocs/video_recorder/video_recorder_bloc.dart
```diff
@@ -1612,13 +1608,15 @@ class VideoRecorderBloc
...
-        StopMotionFrameOps.defaultFramesPerImage,
-      );
+  /// Hold duration of one captured still in a session of [frameCount] stills,
+  /// at the render frame rate, so the library preview, the timeline, and the
+  /// assembled clip all agree.
+  ///
+  /// A session too short to fill a second on its own is stretched to it
+  /// ([StopMotionFrameOps.initialHold]): three stills at the default hold play
+  /// in an eighth of a second, which the editor only shows as a flicker.
+  static Duration _stopMotionHold(int frameCount) =>
```
realmeylisdev, line 1618: Two follow-ons this stretch creates outside the diff:

**1. `video_editor_screen.dart:371` is now wrong.** It says "Fresh captures arrive with the app default hold" — they arrive at `initialHold(n)` now. The `setGlobalHold(capturedFrames, globalDefaultFramesPerImage(targetFrames))` right under it is what keeps the editor's "+" path correct, and it's quietly doing more work than its comment claims: without it, one still added from the editor into a 30-still clip would splice in at a **1-second hold** instead of 41ms. Traced it: `fromEditor` capture → `_ingestStopMotionClip([path])` → `initialHold(1)` = 1s → `pop(true)` → `_insertCapturedStopMotionFrames` normalizes it. Safe as shipped, but worth updating the comment so a future reader doesn't decide the normalization is redundant and drop it.

**2. `_onStopMotionFrameCaptured` still reads `assembling` at :1650 and :1697.** Those are unreachable for exactly the reason 48ecf9a deleted the bottom-bar lock and the `_applyRecorderMode` guard — a synchronous assemble means no other handler can be running while `assembling` is the current state, so `ready` is what catches both cases in practice. The existing `a frame captured while assembling is dropped` test only reaches that branch by seeding the status directly. Either drop the two conditions alongside the others, or at least retire the comments above them — "A still added to that list mid-assemble is either missed by the save or re-persisted afterwards as a phantom one-frame session" now describes a race that can't happen.

mobile/lib/widgets/video_editor/main_editor/video_editor_thumbnail.dart
```diff
@@ -24,6 +24,14 @@ class VideoEditorThumbnail extends ConsumerWidget {
       child: clip.thumbnailPath != null
           ? ClipThumbnailImage(
               path: clip.thumbnailPath!,
+              // A stop-motion clip's thumbnail is a raw full-resolution camera
+              // still (several MB decoded), and the stop-motion player decodes
+              // the same file at this height right after — bounding the decode
+              // keeps the placeholder from stalling on it, and both share one
+              // image-cache entry instead of decoding twice.
+              cacheHeight:
```
realmeylisdev, line 32: Confirmed the shared-entry claim rather than taking the comment's word for it: `_VideoSetupLoadingIndicator` builds `containSize` with `height == renderSize.height` (`video_editor_canvas.dart:2821`), which is the same `renderSize` feeding `StopMotionPlayer`'s `cacheHeight` (`video_editor_player.dart:54`), and both sides land on `ResizeImage(FileImage(path, scale: 1.0), height: h)` — identical cache key. For a stop-motion clip `thumbnailPath == frames.first.path`, so frame 0 genuinely is the shared entry. Aspect ratio is preserved (height-only resize) so the `FittedBox(fit: .cover)` result is unchanged.

nit — `video_editor_thumbnail_test.dart` only covers the null-clip branch, so this path ships untested. A widget test pumping a clip with a `thumbnailPath` and asserting the resolved `Image.image` is a `ResizeImage` at the expected height would pin both the bound and the key the sharing depends on.

realmeylisdev (approved):
Verified all three fixes hold. The stretch math is exact for every session size 1–40 (mutation-checked: dropping the `+1` reddens the new test at exactly the 3-still case), the synchronous assemble really does make the `assembling` readers you deleted unreachable, and the shared image-cache-entry claim checks out end to end.

Three nits inline, none blocking. One process note: a before/after screen recording of the editor preview would help — the flicker half is the part a reviewer can't confirm without a device.

hm21:
All three in 83047c8:

- **Stale doc** — `defaultFramesPerImage` now states what it actually still is (the clamp floor plus the `globalDefaultFramesPerImage` fallback), and that a new capture path wanting a starting hold wants `initialHold`.
- **`video_editor_screen.dart:371`** — the comment now says outright that the `setGlobalHold` undoes the `initialHold` stretch (otherwise a single still splices in at a 1s hold instead of 41ms), so nobody reads the call as redundant.
- **The two `assembling` readers** — dropped rather than just re-commented, the same way 48ecf9a dropped the bottom-bar lock and the `_applyRecorderMode` guard. The test that only reached the branch by seeding the status directly went with them.
- **`video_editor_thumbnail_test.dart`** — added a test pinning the `ResizeImage` height and the `FileImage` key. Mutation-checked: it reddens without `cacheHeight`.

On the rounding in `framesPerImageToDuration`: agreed, but it would shift every persisted hold — not in this PR.

Screen recording to follow.

PR #6469, opened 2026-07-28
fix(video-player): stop the loop restart being audible or visible

### LINKED ISSUE #6468: fix: looping playback clicks at the loop point because the player never fades the composition edges
## Summary

`DivineVideoPlayer` loops without any audio fade at the composition boundary, so the waveform jumps from its end amplitude straight to its start amplitude on every loop restart. When the audio does not happen to sit near zero at the cut, that step is audible as a short click.

## Environment

- App version: current `main`
- Device: reported on iOS (iPad); the same gap exists in the Android implementation
- Network: N/A

## Steps to Reproduce

1. Record a clip whose audio is still at a meaningful amplitude when it ends (talking, music, ambient noise — not a fade to silence)
2. Let it loop, either in the video editor canvas or in the feed
3. Listen at each loop restart

## Actual Result

A short click on every loop restart.

## Expected Result

A seamless loop. The join should not be audible.

## Regression?

No. Confirmed by the reporter as pre-existing — the click was audible before either of the changes below.

The loop point has never been tied to a zero crossing. In the editor it comes from `clip.duration`, which until now was the recording stopwatch (how long the button was held) and bears no relation to where the media's tracks actually end, so it could always land mid-waveform.

Two changes move *where* that cut sits, without causing the click:

- #6430 (merged) clamps feed playback to the common track end.
- #6439 gives a recorded clip its media's length, so the editor cuts there too.

## Evidence

**iOS** — `packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift:532`

The `AVMutableAudioMix` is built with one flat ramp per clip:

```swift
params.setVolumeRamp(
  fromStartVolume: vol,
  toEndVolume: vol,
  timeRange: range
)
```

Start and end volume are identical — this only carries per-clip gain. Nothing fades the first or last few milliseconds of the composition, and `AVPlayerLooper` joins those two edges directly.

**Android** — `packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt`

Per-clip volume is a constant assignment (`exoPlayer.volume = clipVolumes[...] * volume`) and looping is `Player.REPEAT_MODE_ALL`. There is no fade anywhere in the file.

## Implementation notes

iOS is the cheap side: the `AVMutableAudioMix` already exists, so a short ramp from 0 over the first few milliseconds and down to 0 over the last few is two extra `setVolumeRamp` calls in the same builder.

Android has no mix object. It needs either a custom `AudioProcessor` or volume automation driven near the clip boundary, which is meaningfully more work — worth scoping separately from the iOS side.

The fade must be short enough not to be heard as a dip. It also has to stay clear of the per-clip gain ramps so a muted clip stays muted.

## Open question

Not yet confirmed by an A/B on device — the reasoning above is from the source, not from a measurement. Worth reproducing on `main` in the feed (which already clamps via #6430) before building the fix, to confirm the click is the boundary and not something else in the loop path.

Related to #6430, #6439, #6386

### LINKED ISSUE #6485: fix: Android never clamps the loop to the video track for remote sources
## Summary

The container duration of an MP4 is the length of its *longest* track. When a video's audio track is declared longer than its video track, looping to the container duration replays the difference as a frozen last frame before every restart.

`boundedCommonTrackEndMs` clamps this away — but on Android only for local files. Reading a remote source's track lengths is blocking I/O and the function ran on the platform thread, so every `https://` clip returned `null` and kept the seam. iOS has no such limit: it awaits `load(.timeRange)` on both tracks before building the composition, which is why the same videos loop cleanly there.

## Evidence

Measured over 59 live feed videos from `api.divine.video/api/videos`, comparing each file's video-track duration against its container duration:

| Container longer than the video track by | Videos |
|---|---|
| 0 (exact) | 32 (54%) |
| ≤ 10 ms | 9 (15%) |
| 10–50 ms | 16 (27%) |
| 50–500 ms | 2 (3%) |
| > 500 ms | 0 |

**27 of 59 (46%) freeze on Android**, by up to 69 ms — one to two frames at every restart. All of them are inside the existing 500 ms clamp bound, so this is purely the remote-source gap, not the bound.

Reproduces on these live videos (worst first):

- 69 ms — https://divine.video/video/1ccf911a7a4a93b77e4012b7b6bf4941a2ff97abba4c29b9fbec3556e5dbed77
- 64 ms — https://divine.video/video/2f76388f848a7e9b9b86e61896c0ccc0e4b505b60bbf04351fb9d77b7815f46c
- 46 ms — https://divine.video/video/39567fbfcd7511e9633ec082e576ff0cdeb783700e8adea1c843d457586e746e

Play one on Android and on iOS side by side: the restart hitches only on Android.

## Expected

The loop restarts when the picture ends.

## Out of scope

Mismatches *larger* than the 500 ms clamp bound are not covered here — see #6486.

hm21:
## Description

A looping video cuts from its last sample straight back to its first. That cut has never been tied to a zero crossing — it comes from a trim or a recording length — so unless both edges happen to sit near silence, the step is audible as a click on every restart.

Two separate things made the restart noticeable, and this PR fixes both: the audio steps (a click), and on Android the picture freezes first (a hitch).

## The click

Both players now fade the video's outer edges. Only the outer edges: the join is the only edge a loop actually plays, and fading each clip of a multi-clip timeline would notch the audio at every cut. On Apple that means the composition's outer edges at any clip count. On Android the only anchor available is a stream change, which is indistinguishable from a cut inside the timeline, so there the fade is gated to single-clip videos — which is what the feed loops.

### Apple

The `AVMutableAudioMix` already existed for per-clip volume, so the fade folds into ramps that were previously flat.

It has to fold in rather than layer on top. I tried the two-line version first — set the flat ramp, then write the fades over it — and AVFoundation rejects a ramp that crosses an existing one:

```
NSInvalidArgumentException: The timeRange of a ramp must not overlap
the timeRange of an existing ramp.
```

That is an Objective-C exception, not a Swift `Error`, so it would abort the process — the same class of trap as the `setVideoComposition:` guard already documented in that file. The ramps are laid end to end instead. Folding also means the fade starts and ends at the clip's own gain, so a muted clip stays silent through it.

The ramps are laid out on the `CMTime`s the composition was built from. Rebuilding each bound from `CMTimeGetSeconds` at timescale 600 looks equivalent and is not: `CMTimeMakeWithSeconds` truncates, so the accumulated end fell short of the composition's by up to a tick per clip. Measured against real AVFoundation, `getVolumeRamp` found **no ramp at all** covering the composition's last sample — for a 6.006 s asset at timescale 90000, a 4.317 s one at 44100, a 12 ms clip, and a three-clip timeline. Whatever plays there is not the zero the fade is supposed to reach. On the exact `CMTime`s, the same four cases report a ramp ending at volume 0.

Seconds are still used for one thing: capping the fade at half the first and last clip so a very short video's two fades cannot overlap. That cap goes through one helper that truncates to whole 600-ticks, and the halving happens there rather than in seconds — `CMTimeMakeWithSeconds` documents no rounding *direction*, and a half that rounded up while the whole rounded down would put the two ramps over each other. A truncated cap can only understate the clip, which is the safe direction.

### Android

Harder, because ExoPlayer gives an `AudioProcessor` no position of its own. `StreamMetadata.positionOffsetUs` is 0 on every playback-path flush in media3 1.10: `DefaultAudioSink.setupAudioProcessors()` calls the deprecated no-argument `AudioProcessingPipeline.flush()`, which substitutes `StreamMetadata.DEFAULT`. So media3's own `GainProcessor` has nothing to place a fade against.

The two edges end up being found in opposite ways.

**The start is a counting problem.** Frames are counted from a flush, and the count is anchored from outside by the stream change `MediaCodecAudioRenderer` raises from `onProcessedStreamChange` — after the last buffer of the finished period has gone to the sink and before the first buffer of the next. That is exactly the join. A `ForwardingAudioSink` turns it into an anchor.

The wrapper's real job there is retiring the seek offset. A seek flushes the processing pipeline twice before a single frame is queued — once from `DefaultAudioSink.flush()`, which then releases the audio output, and again when the next buffer rebuilds it. A one-shot offset is eaten by the first flush and the second re-anchors at zero, so the lap after a seek would fade in mid-video. The offset is sticky and retired at the join, which is where it genuinely stops applying. The decode-only buffers a seek discards are reported as discontinuities too, so an anchor is only taken once input has actually arrived.

**The end cannot be counted towards**, because nothing on this side of the pipeline knows where it is until it arrives. `player.duration` is the container's figure, in whole milliseconds, for its longest track; the decoded audio track ends up to an AAC frame away from that — roughly twice the fade. Timing the fade out off it left the ramp finishing before the last sample on videos where the gap was wide enough, and those clicked on their first restart. (Reported on device exactly that way: only the first restart, and only some videos.)

So the last 10 ms are held back instead, and released faded once media3 reports the stream has ended. That report is exact and arrives at every loop restart: `handleDiscontinuity` sets `startMediaTimeUsNeedsSync`, and the next buffer makes `DefaultAudioSink` run `drainToEndOfStream()`, which queues end-of-stream into the pipeline before any frame of the new lap is. `TrimmingAudioProcessor` holds its own tail back the same way, one slot upstream in this same chain.

`videoDurationUs` survives only to shorten the fade on a video too short to carry two of them. Being a millisecond out there costs ramp length, not position.

The cost is that audio runs 10 ms behind the position the player reports, on single-clip looping playback only — a multi-clip timeline holds nothing back and keeps its latency, which a test pins. Ten milliseconds is an order of magnitude inside the threshold where a viewer notices audio trailing video.

One deliberate softening: an unexpected PCM encoding drops the processor out of the chain (`AudioFormat.NOT_SET`) instead of throwing. `GainProcessor` throws `UnhandledAudioFormatException` here, which surfaces as a playback error; a fade must never cost the audio.

## The freeze

The container duration of an MP4 is its *longest* track. Where a video's audio track is declared longer than its video track, looping to the container replays the difference as a frozen last frame.

`boundedCommonTrackEndMs` already clamped that — on Android only for local files, because reading a remote source's track lengths is blocking I/O and it ran on the platform thread. Every `https://` clip therefore returned `null` and kept the seam. iOS has no such limit: it awaits `load(.timeRange)` on both tracks before building the composition, which is why the same videos loop cleanly there and hitch here.

Measured over 59 live feed videos, comparing each file's video-track duration against its container:

| Container longer by | Videos |
|---|---|
| 0 (exact) | 32 (54%) |
| ≤ 10 ms | 9 (15%) |
| 10–50 ms | 16 (27%) |
| 50–500 ms | 2 (3%) |
| > 500 ms | 0 |

**27 of 59 (46%) freeze on Android**, by up to 69 ms — one to two frames at every restart. All are inside the existing 500 ms bound, so this is the remote-source gap, not the bound.

Android now reads local-file lengths before applying clips. Remote lengths warm on a single-thread executor without sitting in front of `setClips` or first frame; if the answer lands for the active generation, the current playlist is tightened in place, and otherwise the cached answer trims the next play of that source. Results are cached per source and bounded, so a failover to a mirror or a second visit does not pay the read again, and a source whose lengths cannot be read caches as unreadable rather than retrying every attempt. A newer `setClips` bumps a generation so a late background result cannot apply over clips that replaced it. If a metadata read misses its deadline, that player disables further probes and plays unclamped rather than making every later load wait behind a pinned native read.

**Related Issues:** Closes #6468, closes #6485

## Out of Scope

- Clip joins *inside* a multi-clip timeline still cut hard. Whether those want a declick fade or a crossfade is a product call, and changing them would alter how existing multi-clip edits sound.
- Android leaves multi-clip timelines unfaded at their outer edges too, where Apple fades them. Closing that gap needs a per-clip end signal rather than the stream end; the surface it affects is the editor's looping preview, not the feed.
- The 500 ms clamp bound is untouched. No live feed video in the sample exceeds it; the bundled Vine seed assets contain a 754 ms case, where the mux declares 762 ms of audio it has no packets for. Making the bound asymmetric — unbounded when the *audio* is the longer track, since a frozen frame is always wrong — is tracked in #6486.
- A musically seamless loop is not a player concern. The corpus is cut at a stopwatch, not on a beat, so the choice at the join is a click or a short dip; a crossfade would have to be chosen where the loop point is set, in the recorder or editor, and the video loop would have to lose the same overlap.
- A drain in the middle of a stream — `setPlaybackSpeed` on a running loop — now releases the held tail faded, so that moment carries a ~20 ms dip instead of a hard cut. It does not occur in the feed.
- Export is untouched — this is the playback path only.

## Verification

- `gradle :divine_video_player:testDebugUnitTest` — green, with coverage over the declick processor and clamp resolution: local clips waiting for track lengths before the player is touched, remote clips starting immediately while metadata warms in the background, a late resolution being dropped rather than applied over newer clips, a metadata timeout disabling future probes on that player, and an unreadable source still reaching the player unclamped. The clamp had no Android unit test before — `mockHandler` is a relaxed mock, so `post {}` never ran, and `trimToCommonTrackEnd` did not appear in the test file at all. Five pin the fade-in curve and the length cap; eleven are buffer-level, driving `configure → flush → queueInput → getOutput → queueEndOfStream` over real PCM: the tail fading to zero where the stream ends rather than where the container claims, an unknown declared length still fading out, the ramp falling monotonically into the join, the held-back 10 ms, a flush dropping the tail instead of replaying it late, a multi-clip timeline passing through with no delay, the seek offset surviving both flushes of a seek, the decode-only buffers not anchoring a loop, a restart re-arming the fade in, and two stereo tests over per-channel gain and frames-not-samples.
- Mutation-checked: suppressing the tail release, releasing it unfaded, keeping it across a flush, holding frames back with the fade off, dropping `channelCount` from the interleave index, and counting samples as frames each turned exactly their own tests red and nothing else.
- `flutter test` in `packages/divine_video_player` — 276 tests, green.
- `flutter analyze lib test` and `dart format` — clean.
- The Apple ramp layout was verified by running the builder against real AVFoundation and reading the ramps back with `getVolumeRamp`, across a feed video, a muted video, a 12 ms video, and multi-clip timelines with mixed gains — including the before/after on the composition's last sample quoted above. That is also how the overlap exception was found.
- Mutation-checked on the clamp too: dropping the generation guard, and applying the clips without waiting for the resolution, each turned exactly their own tests red.
- Verified on device: the click at the loop join is gone on both platforms, on the first restart as well as later ones, and the restart hitch on Android is gone.

## Type of Change

### DIFF
--- a/mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
+++ b/mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
@@ -7,13 +7,13 @@ import android.net.Uri
 import android.os.Handler
 import android.os.Looper
 import android.view.Surface
+import androidx.media3.common.C
 import androidx.media3.common.MediaItem
 import androidx.media3.common.PlaybackException
 import androidx.media3.common.PlaybackParameters
 import androidx.media3.common.Player
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.datasource.HttpDataSource
-import androidx.media3.exoplayer.DefaultRenderersFactory
 import androidx.media3.exoplayer.ExoPlayer
 import androidx.media3.exoplayer.SeekParameters
 import androidx.media3.exoplayer.source.DefaultMediaSourceFactory
@@ -160,6 +160,13 @@ internal class DivineVideoPlayerInstance(
     private val audioOverlayManager = audioOverlayManagerFactory(context)
 
     /**
+     * Fades the outer edges of a looping video so its loop join is not a click
+     * (#6468). Lives for the life of the instance because it is wired into the
+     * renderers factory when the player is built.
+     */
+    private val declickProcessor = LoopDeclickAudioProcessor()
+
+    /**
      * Pending result for an async seekTo call.
      * Completed when ExoPlayer transitions to STATE_READY after a seek,
      * so the Dart `await seekTo()` blocks until the frame is decoded.
@@ -286,7 +293,8 @@ internal class DivineVideoPlayerInstance(
         // extraction contend for a small hardware decoder pool — into a
         // successful (if slower) decode instead of a dead surface.
         val renderersFactory =
-            DefaultRenderersFactory(context).setEnableDecoderFallback(true)
+            LoopDeclickRenderersFactory(context, declickProcessor)
+                .setEnableDecoderFallback(true)
         val builder = ExoPlayer.Builder(context, renderersFactory)
             .setMediaSourceFactory(
                 DefaultMediaSourceFactory(
@@ -526,6 +534,14 @@ internal class DivineVideoPlayerInstance(
         // position so the timeline doesn't show intermediate values.
         pendingGlobalStartMs = globalStartMs
 
+        // Fade the video's edges only when the whole video is one clip, which
+        // is what the feed loops. On a multi-clip timeline a stream is one clip
+        // rather than the whole video, so fading every stream would notch the
+        // audio at each cut. The length is not known until STATE_READY.
+        declickProcessor.enabled = clipCount == 1
+        declickProcessor.videoDurationUs = LoopDeclickAudioProcessor.DURATION_UNKNOWN
+        declickProcessor.nextStreamStartUs = startLocalMs * 1000L
+
         // Hold the Dart result until STATE_READY so `await setClips()` only
         // resolves once the decoder is truly ready. Cancel any previous
         // in-flight setClips (shouldn't happen, but defensive) — surface as
@@ -620,6 +636,19 @@ internal class DivineVideoPlayerInstance(
         return commonEndMs.takeIf { trimMs <= trimLimitMs }
     }
 
+    /**
+     * Hands the declick fade the current video's length, so it knows where the
+     * loop join it has to fade into actually is.
+     */
+    private fun updateDeclickDuration() {
+        val durationMs = player?.duration ?: C.TIME_UNSET
+        declickProcessor.videoDurationUs = if (durationMs == C.TIME_UNSET) {
+            LoopDeclickAudioProcessor.DURATION_UNKNOWN
+        } else {
+            durationMs * 1000L
+        }
+    }
+
     private fun handleSeekTo(call: MethodCall, result: MethodChannel.Result) {
         val globalMs = (call.argument<Number>("positionMs"))?.toLong() ?: 0L
         val exoPlayer = ensurePlayer()
@@ -641,6 +670,10 @@ internal class DivineVideoPlayerInstance(
         val resolved = resolveGlobalPosition(globalMs)
 
         val targetIndex = resolved.first
+        // The seek flushes the audio pipeline, which re-anchors the fade at
+        // whatever the next stream starts with. Tell it that is the seek target
+        // and not the start of the video, so the fade out stays on the join.
+        declickProcessor.nextStreamStartUs = resolved.second * 1000L
         exoPlayer.seekTo(targetIndex, resolved.second)
 
         // Apply the target clip's per-clip speed and volume immediately.
@@ -1000,6 +1033,9 @@ internal class DivineVideoPlayerInstance(
                 syncAudioOverlays()
             }
             if (playbackState == Player.STATE_READY) {
+                // The video's length is only known once the timeline is
+                // populated, and the fade out has to know where the join is.
+                updateDeclickDuration()
                 // Seek complete — switch from reporting target to actual position.
                 pendingGlobalStartMs = 0L
                 completeSeekIfPending()
--- a/mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
+++ b/mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
@@ -170,10 +170,7 @@ internal class DivineVideoPlayerInstance(
 
     private val audioOverlayManager = audioOverlayManagerFactory(context)
 
-    /**
-     * Bumped by every [handleSetClips]. A resolution that finishes after a
-     * newer call started is dropped rather than applied over it.
-     */
+    /** Bumped by every [handleSetClips] so late background work can be dropped. */
     private var setClipsGeneration = 0L
 
     /**
@@ -197,14 +194,25 @@ internal class DivineVideoPlayerInstance(
      */
     private val deferredSetClipsTimeout = Runnable {
         deferredSetClips?.let { deferred ->
+            trackDurationProbingDisabled = true
             DivineVideoPlayerLog.warning(
                 "Player $playerId applied clips unclamped: track lengths not " +
-                    "read within ${TRACK_DURATION_RESOLVE_TIMEOUT_MS}ms",
+                    "read within ${TRACK_DURATION_RESOLVE_TIMEOUT_MS}ms; " +
+                    "metadata probing disabled for this instance",
                 name = "DivineVideoPlayer.Load",
             )
             settleDeferredSetClips(deferred, apply = true)
         }
     }
+
+    /**
+     * True after a metadata read misses its deadline on this instance.
+     *
+     * `MediaExtractor` remote reads are native I/O and may not be interruptible.
+     * Once the single metadata thread is plausibly pinned, queuing more work only
+     * makes later loads wait for a deadline before playing unclamped anyway.
+     */
+    private var trackDurationProbingDisabled = false
 
     /**
      * Fades the outer edges of a looping video so its loop join is not a click
@@ -456,24 +464,36 @@ internal class DivineVideoPlayerInstance(
         val generation = ++setClipsGeneration
         deferredSetClips?.let { settleDeferredSetClips(it, apply = false) }
 
-        val unresolved = clipsRaw.mapNotNull { map ->
-            val uri = map["uri"] as? String ?: return@mapNotNull null
-            if (map["trimToCommonTrackEnd"] as? Boolean != true) return@mapNotNull null
-            if (!canReadTrackDurations(uri)) return@mapNotNull null
-            if (trackDurationsCache.containsKey(uri)) return@mapNotNull null
-            uri to httpHeadersOf(map)
+        val unresolved = if (trackDurationProbingDisabled) {
+            emptyList()
+        } else {
+            clipsRaw.mapNotNull { map ->
+                val uri = map["uri"] as? String ?: return@mapNotNull null
+                if (map["trimToCommonTrackEnd"] as? Boolean != true) return@mapNotNull null
+                if (!canReadTrackDurations(uri)) return@mapNotNull null
+                if (trackDurationsCache.containsKey(uri)) return@mapNotNull null
+                uri to httpHeadersOf(map)
+            }
         }
-        if (unresolved.isEmpty()) {
+        val blockingUnresolved = unresolved.filter { (uri, _) ->
+            shouldBlockSetClipsForTrackDurations(uri)
+        }
+        val backgroundUnresolved = unresolved.filterNot { (uri, _) ->
+            shouldBlockSetClipsForTrackDurations(uri)
+        }
+        if (blockingUnresolved.isEmpty()) {
+            if (backgroundUnresolved.isNotEmpty()) {
+                warmTrackDurationsInBackground(generation, clipsRaw, backgroundUnresolved)
+            }
             applyClips(call, clipsRaw, result)
             return
         }
 
-        // Where each track really ends is only in the source's metadata, and
-        // reading it blocks. Resolve off the platform thread, then apply on
-        // it — the alternative is looping to the container duration, which is
-        // the *longest* track, and leaving the shorter one's tail as a frozen
-        // frame before every restart.
-        val deferred = DeferredSetClips(generation, call, clipsRaw, result)
+        // Where each track really ends is only in the source's metadata. Local
+        // files can be read before the playlist swap without adding a network
+        // round trip to first frame; remote sources warm in the background and
+        // tighten the current playlist when they land.
+        val deferred = DeferredSetClips(call, clipsRaw, result)
         deferredSetClips = deferred
         mainHandler.postDelayed(
             deferredSetClipsTimeout,
@@ -486,11 +506,20 @@ internal class DivineVideoPlayerInstance(
                 // playing to the container duration is a seam, not playing at
                 // all is a dead player.
                 try {
-                    unresolved.forEach { (uri, headers) ->
+                    blockingUnresolved.forEach { (uri, headers) ->
                         resolveTrackDurations(uri, headers)
                     }
                 } finally {
-                    mainHandler.post { settleDeferredSetClips(deferred, apply = true) }
+                    mainHandler.post {
+                        if (backgroundUnresolved.isNotEmpty()) {
+                            warmTrackDurationsInBackground(
+                                generation,
+                                clipsRaw,
+                                backgroundUnresolved,
+                            )
+                        }
+                        settleDeferredSetClips(deferred, apply = true)
+                    }
                 }
             }
         } catch (e: RejectedExecutionException) {
@@ -502,6 +531,68 @@ internal class DivineVideoPlayerInstance(
             if (claimDeferredSetClips(deferred)) {
                 result.error("DISPOSED", "Player was disposed", null)
             }
+        }
+    }
+
+    private fun warmTrackDurationsInBackground(
+        generation: Long,
+        clipsRaw: List<Map<String, Any?>>,
+        unresolved: List<Pair<String, Map<String, String>>>,
+    ) {
+        try {
+            metadataExecutor.execute {
+                try {
+                    unresolved.forEach { (uri, headers) ->
+                        resolveTrackDurations(uri, headers)
+                    }
+                } finally {
+                    mainHandler.post {
+                        applyResolvedCommonTrackEnds(generation, clipsRaw)
+                    }
+                }
+            }
+        } catch (e: RejectedExecutionException) {
+            DivineVideoPlayerLog.warning(
+                "Player $playerId dropped background track-duration read: $e",
+                name = "DivineVideoPlayer.Load",
+            )
+        }
+    }
+
+    private fun applyResolvedCommonTrackEnds(
+        generation: Long,
+        clipsRaw: List<Map<String, Any?>>,
+    ) {
+        if (generation != setClipsGeneration) return
+        val exoPlayer = player ?: return
+        if (exoPlayer.mediaItemCount != clipsRaw.size) return
+
+        var changed = false
+        clipsRaw.forEachIndexed loop@{ index, map ->
+            val uri = map["uri"] as? String ?: return@loop
+            if (map["trimToCommonTrackEnd"] as? Boolean != true) return@loop
+            val startMs = (map["startMs"] as? Number)?.toLong() ?: 0L
+            val requestedEndMs = (map["endMs"] as? Number)?.toLong()
+            val commonEndMs = boundedCommonTrackEndMs(uri, startMs, requestedEndMs)
+                ?: return@loop
+            val effectiveEndMs = listOfNotNull(requestedEndMs, commonEndMs).minOrNull()
+                ?: return@loop
+            val currentItem = exoPlayer.getMediaItemAt(index)
+            if (currentItem.clippingConfiguration.endPositionMs == effectiveEndMs) {
+                return@loop
+            }
+            exoPlayer.replaceMediaItem(
+                index,
+                buildMediaItem(uri, startMs, effectiveEndMs),
+            )
+            changed = true
+        }
+        if (changed) {
+            refreshClipOffsets(exoPlayer)
+            DivineVideoPlayerLog.info(
+                "Player $playerId applied resolved track-end clamp",
+                name = "DivineVideoPlayer.Load",
+            )
         }
     }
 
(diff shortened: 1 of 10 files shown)


mobile/packages/divine_video_player/android/src/test/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstanceTest.kt
```diff
@@ -701,4 +702,231 @@ class DivineVideoPlayerInstanceTest {
...
+        val posted = mutableListOf<Runnable>()
+        verify { mockHandler.post(capture(posted)) }
+        return posted
+    }
+
+    @Test
+    fun `setClips waits for the track lengths before touching the player`() {
+        val result = mockk<MethodChannel.Result>(relaxed = true)
+
+        instance.onMethodCall(trimmingSetClipsCall("https://cdn.example/a.mp4"), result)
```
realmeylisdev: This test (and the five clamp tests below it) hands the read to a real `Executors.newSingleThreadExecutor()` and then asserts on `mockHandler.post` with nothing in between. mockk's `verify {}` doesn't wait by default, so it's relying on a brand-new worker thread finishing before a mockk verification runs on the test thread. That usually holds on a fast machine, but it's a race by construction rather than a guarantee.

The seam is already here — `StalledExecutorService` proves the constructor takes an executor. A same-thread one in the same shape makes all of these deterministic, and it also clears up the leaked threads: nothing calls `dispose()` in most of them, so every test that submits leaves a non-daemon pool thread behind for the life of the JVM.

I ran the module five times under 8x CPU load and it stayed green, so this is latent rather than something I reproduced — but CI runners are noisier than a local Mac.

mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
```diff
@@ -545,27 +701,55 @@ internal class DivineVideoPlayerInstance(
...
-        return try {
-            extractor.setDataSource(path)
+        val durations = try {
+            when {
+                uri.startsWith("/") -> extractor.setDataSource(uri)
+                uri.startsWith("file://") ->
+                    extractor.setDataSource(Uri.parse(uri).path ?: return)
+                uri.startsWith("http://") || uri.startsWith("https://") ->
+                    extractor.setDataSource(uri, headers)
+                else -> return
```
realmeylisdev: `.m3u8` is a live playback source in the feed — `playback_sources.dart` keeps HLS as the last fallback for progressive sources, but honours it first when the resolved source is already HLS — and `MediaExtractor` can't open a playlist, so it throws. A throw isn't cached, deliberately, which means every `setClips` for an HLS source pays a doomed network read and holds the playlist swap until it fails.

Same shape for anything landing on this `else -> return`: an unsupported scheme is an answer about the source that will never change, but it isn't cached either, so it re-defers on every call.

Keeping `IOException` uncached is the right call. It's the deterministic failures that shouldn't repeat — caching `NO_TRACK_PAIR` here, and skipping `.m3u8` before the probe is even queued in `handleSetClips`, would cover both.

realmeylisdev (commented):
Traced the media3 1.10 audio path end to end and reproduced the Apple ramp layout against real AVFoundation: the held tail releases through `drainToEndOfStream` before the new lap's position checkpoint is queued, and all six ramp shapes I tried (fades meeting exactly, muted clip, 3-clip timeline) lay without the overlap exception and end at volume 0 on the composition's last sample. The seconds round-trip control finds no ramp there at all, so that fix is doing exactly what it claims.

No blocking defects found. Comments are about what the new remote probe costs and a race in the tests that cover it.

mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
```diff
@@ -397,6 +450,105 @@ internal class DivineVideoPlayerInstance(
...
+        // frame before every restart.
+        val deferred = DeferredSetClips(generation, call, clipsRaw, result)
+        deferredSetClips = deferred
+        mainHandler.postDelayed(
+            deferredSetClipsTimeout,
+            TRACK_DURATION_RESOLVE_TIMEOUT_MS,
+        )
+
+        try {
+            metadataExecutor.execute {
```
realmeylisdev, line 504: The feed `await`s `setClips` on the way to first frame (`source_loader.dart` -> `setSourceWithFallbacks`), and `MediaExtractor` doesn't read through `VideoCache.dataSourceFactory`, so every cold source now pays a second connection and round trip for the same moov ExoPlayer is about to fetch again. I checked divine-blossom and it does serve `Accept-Ranges: bytes`, so it's cheap in bytes — but it's still serialized ahead of ExoPlayer's first byte, and on a cold app start that's every video in the session.

The PR quantifies the win precisely (46% of videos, up to 69 ms) but not this side. Do you have a before/after time-to-first-frame? If it turns out to be non-trivial, applying the clips unclamped straight away and letting the clamp land from the next `setClips` for that source would keep the whole benefit for everything except a video's very first play.

mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
```diff
@@ -42,6 +46,11 @@ internal class DivineVideoPlayerInstance(
     private val audioOverlayManagerFactory: (Context) -> AudioOverlayManager = { ctx ->
         AudioOverlayManager(ctx)
     },
+    /**
+     * Reads source metadata off the platform thread. Single-threaded, so the
+     * clips of one call resolve in order and two calls cannot interleave.
+     */
+    private val metadataExecutor: ExecutorService = Executors.newSingleThreadExecutor(),
```
realmeylisdev: `shutdownNow()` can't interrupt a blocking `MediaExtractor` read — it's native I/O — so the stalled host this design already anticipates pins this thread past `dispose()`, and `Executors.defaultThreadFactory()` leaves it non-daemon and anonymous. A named daemon factory (`divine-video-metadata-$playerId`) is one line and turns that into something legible in a thread dump instead of an unexplained thread.

mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
```diff
@@ -641,6 +847,10 @@ internal class DivineVideoPlayerInstance(
         val resolved = resolveGlobalPosition(globalMs)
 
         val targetIndex = resolved.first
+        // The seek flushes the audio pipeline, which re-anchors the fade at
+        // whatever the next stream starts with. Tell it that is the seek target
+        // and not the start of the video, so the fade out stays on the join.
+        declickProcessor.nextStreamStartUs = resolved.second * 1000L
```
realmeylisdev, line 1008: Two other sites flush the audio pipeline without publishing a start position: `handleJumpToClip`'s `seekTo(index, 0)` and the surface-reattach `p.seekTo(p.currentPosition)` in the surface callback. On a single-clip player the reattach one re-fades in mid-video, and the jump one can skip the fade in entirely if an earlier seek left a stale offset behind.

Both are small enough to live with — a line saying why they're exempt would save the next reader working it out, since they're the same class of thing this line fixes.

mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
```diff
@@ -502,6 +647,17 @@ internal class DivineVideoPlayerInstance(
...
         }
 
+        // Fade the video's edges only when the whole video is one clip, which
+        // is what the feed loops. On a multi-clip timeline a stream is one clip
+        // rather than the whole video, so fading every stream would notch the
+        // audio at each cut. The length is not known until STATE_READY.
+        //
+        // Published before the playlist swap below, because that swap disables
+        // the audio renderer and flushes the pipeline on the playback thread.
+        declickProcessor.enabled = clipCount == 1
```
realmeylisdev, line 758: This keys on clip count, not on whether the player loops, so a single-clip player that never loops also gets both fades and the 10 ms hold. Nothing I can find breaks because of it, but the processor's own doc opens with "a looping video's audio" — worth one of the two moving toward the other.

mobile/packages/divine_video_player/darwin/divine_video_player/Sources/divine_video_player/DivineVideoPlayerInstance.swift
```diff
@@ -35,6 +35,55 @@ final class DivineVideoPlayerInstance: NSObject, FlutterStreamHandler {
...
+    ///     fade 10 ms -> 0.6034      fade 25 ms -> 0.0037
+    ///     fade 20 ms -> 0.2032      fade 30 ms -> 0.0008
+    ///
+    /// At 10 ms the join still carries 60% of full amplitude, which is the
+    /// click this was meant to remove. The floor measured 24.8 ms at both
+    /// 44.1 kHz and 48 kHz, so it is a duration rather than a block of frames;
+    /// 30 ms clears it and is still short enough not to read as a level
+    /// change. The Android player has no such floor — it fades decoded frames
+    /// — and stays at 10 ms, where its fade doubles as its audio latency.
+    private static let edgeDeclickFadeSeconds = 0.030
```
realmeylisdev, line 61: This lands on every composition Apple builds, so the editor's single-clip preview now fades 30 ms at both ends while the export path doesn't. Right call for the feed — just confirming it's intended for the trim preview, where preview and export no longer match at the edges.

hm21:
@realmeylisdev Thanks — went through all seven against the code. Two landed as changes in 809b4a2, the rest are answers.

## Changed

**Probing sources `MediaExtractor` cannot open.** Agreed on both halves, and one check covers them: `canReadTrackDurations` now runs before the probe is queued, so an `.m3u8` (query string and fragment stripped first) never defers a `setClips`, and an unsupported scheme is decided without a read instead of returning uncached from inside one. That retires the `else -> return` branch rather than teaching it to cache. New test pins it — `setMediaItems` once, `mainHandler.post` never — and it fails if the filter is removed.

**Thread factory.** Named daemon factory, `divine-video-metadata-$playerId`.

## Not changed

**Test race.** The tests never touch a real executor: `setUp` injects `DirectExecutorService` (same-thread) at line 85, and the stalled test injects `StalledExecutorService`. `DirectExecutorService` is at the bottom of the file, immediately below the `StalledExecutorService` you cited. No pool thread is created in this class, so there is nothing to race on and nothing left behind — which also explains why five runs under 8x load stayed green.

**Extra round trip before first frame.** Real, and the reason the flag's own doc says it costs a metadata request. On scope: the cache is `companion`-static across instances (256-entry LRU), the wait is capped at 1500 ms, and the feed initialises neighbours ahead of the visible index — so the serialized read is on the critical path for a cold start's first video and for a failover, not for every video in a session. Applying unclamped and clamping from the next `setClips` would leave the seam in place for the whole first viewing of every video, since the feed calls `setClips` once and loops inside it — that is the case this PR exists for. No before/after TTFF number yet; I would rather measure one than argue it.

**The other two flush sites.** `handleJumpToClip` has no caller outside the plugin's own API surface, and declick is only on at `clipCount == 1`, where jumping to clip 0 is the only legal call. The surface-reattach seek is gated on `!p.isPlaying && STATE_READY`, so what it re-arms is a 10 ms ramp on the next resume from pause — which is what I would want there anyway, not a mid-video dip.

**`enabled = clipCount == 1`.** The field's own doc carries the reason: on a multi-clip timeline a stream is one clip, so fading every stream would notch each cut. Only the class header says "looping". Keying on `isLooping` is not available — `setLooping` is a separate channel call that can arrive after `setClips`, and the processor has to be configured before the playlist swap flushes the pipeline.

**Apple preview vs export.** Intended. 30 ms at each outer edge of the preview and none on export, accepted for the feed and below what a trim decision turns on.

NotThatKindOfDrLiz:
@hm21 I pushed takeover follow-up `93f5749f4` to address the remaining review risk.

What changed:
- uncached remote track-duration probes no longer sit in front of `setClips`/first frame; playback starts immediately and the probe warms the cache, then tightens the current playlist if it lands for the active generation
- a metadata timeout disables further probing on that player instance, so a pinned native `MediaExtractor` thread cannot make later loads wait 1.5s each
- removed the dead deferred generation field; `settled` is now the deferred supersede mechanism
- added a track-duration cache reset hook for Android tests
- broadened HLS skip detection to `/hls/` as well as `.m3u8`
- added the defensive Apple ramp guard and clarified the single-clip declick docs

Verification:
- `JAVA_HOME=/Users/lizsw/.local/share/mise/installs/java/21.0.2/Contents/Home FLUTTER_ROOT=/Users/lizsw/.local/share/mise/installs/flutter/3.44.0 /Users/lizsw/.gradle/wrapper/dists/gradle-8.14-all/c2qonpi39x1mddn7hk5gh9iqj/gradle-8.14/bin/gradle :divine_video_player:testDebugUnitTest` from `mobile/android` - passed
- `flutter test` from `mobile/packages/divine_video_player` - 276 tests passed
- `flutter analyze lib test` from `mobile/packages/divine_video_player` - passed
- pre-push analyzer + targeted contract tests - passed

NotThatKindOfDrLiz (approved):
Takeover follow-up `93f5749f4` addresses the blocking review issues I found: remote metadata probing no longer serializes first frame, stalled metadata probing is capped per instance, the dead deferred generation state/test cache leak/HLS/doc/Apple guard nits are cleaned up, and CI is green.

Ready from my side; author should do the final merge.

hm21:
Thanks @NotThatKindOfDrLiz — reviewed the takeover and kept all of it except one detail, which I fixed in `f6818139ad`.

`applyResolvedCommonTrackEnds` tightened the loaded playlist in place. A clamp lives in the item's `ClippingConfiguration`, and `ClippingMediaSource.canUpdateMediaItem` compares that configuration, so a changed one can never be applied in place: `ExoPlayerImpl.replaceMediaItems` falls back to insert-and-remove, and removing the period being played resolves the position through `resolveSubsequentPeriod` to the *default* position of the one that takes its place. On the 46% of videos where the clamp differs, a background read landing after the first frame therefore jumped the video back to its start mid-watch and dropped its buffer — a worse artifact than the frozen frame it removes, and network-dependent enough to be unpredictable.

The tightening is now gated on the player not having started (`!playWhenReady && currentPosition == 0`). That keeps the clamp for the preloaded neighbours — `_initController(index±1)` calls `setSource` without `play`, and that is where the viewer scrolls next — while the tile loading under the viewer keeps the seam for that session and has its next `setClips` clamped from the cache. Which is what @realmeylisdev suggested on the thread above; the background warm just gets there one load earlier for everything that was preloaded.

Two Kotlin tests cover both sides, mutation-checked: dropping the guard turns exactly the negative one red and nothing else. 41 Android unit tests and 276 `flutter test` green locally.

PR #6470, opened 2026-07-28
chore(release): sync build number with TestFlight build 819

rabble:
## Summary
- `mobile/pubspec.yaml`: 1.0.19+508 → 1.0.19+819
- Xcode project `CURRENT_PROJECT_VERSION` 508 → 819 (Runner, NotificationServiceExtension, CameraQuickActionWidget — all 9 configs)

## Why
Codemagic stamps iOS builds with `PROJECT_BUILD_NUMBER + 142` via `--build-number` (codemagic.yaml:235) and never writes the resolved number back to the repo. The 1.0.19 build live on internal TestFlight is **819** (5 installs, 0 crashes as of this morning), while git still said 508 — so Crashlytics, App Store Connect, and the repo all disagreed about which build exists.

This syncs the repo to reality. No functional change; Codemagic continues to override the number at build time regardless.

## Test plan
- [x] Diff is version strings only (10 lines, 2 files)
- [ ] CI passes


mbradley (approved):
All three shipping targets are updated across all three configurations, and the six sites left at `1` are `RunnerTests` and `uitests` — correctly untouched. `pubspec.yaml` matches the 1.0.19 line from #6462. CI is green.

Two things worth knowing, neither blocking.

**This is cosmetic for anything CI builds.** Every path in `codemagic.yaml` overrides the repo value: iOS passes `--build-number=$((PROJECT_BUILD_NUMBER + 142))` (`:235`), the Android APK passes `$PROJECT_BUILD_NUMBER` (`:134`), and the AAB takes `max(counter, latestPlay + 1)` (`:145-157`). So the synced number doesn't change what ships — it matters for local and manual builds, which includes locally-built Zapstore APKs. Flagging so nobody later reads this file as the control for a shipped build number.

**The durable fix is one line away, and would retire this chore.** `Runner/Info.plist` reads `CFBundleVersion` from `$(FLUTTER_BUILD_NUMBER)`, so the app always picks up whatever `--build-number` passes. But `NotificationServiceExtension/Info.plist` and `CameraQuickActionWidget/Info.plist` read `$(CURRENT_PROJECT_VERSION)` — the pbxproj literal. So on every CI build the app ships `counter + 142` while the extensions ship whatever the checked-in value is. This PR narrows that gap from 311 to about 1; it doesn't close it. Pointing the two extension plists at `$(FLUTTER_BUILD_NUMBER)` would make them track the app automatically and mean nobody has to hand-sync this file again. Pre-existing, not something you introduced — worth its own small PR if you agree.

**One question.** `pubspec.yaml`'s build number is platform-agnostic, but 819 is an iOS-specific figure (Codemagic counter + 142). Android's sequence is the raw counter or Play-derived, so it's a few hundred lower. For CI that's moot since every path overrides. For a local Android build it now stamps `versionCode 819`, ahead of the Play sequence — harmless, and arguably good for monotonicity, but is that the intent or just a side effect of syncing to the iOS number?

For the record: I can't verify 819 is the current TestFlight build from the repo — the Codemagic counter lives outside git, so build numbers aren't derivable here. Taking that from your Codemagic view.

Thanks, as always.

dcadenas (approved):
✅ TestFlight build 819 is synced cleanly

PR #3244, opened 2026-04-22
feat(people-lists): add-person-to-lists feature behind FF_CURATED_LISTS

rabble:
Closes #3245

## Summary

Ships the "add video / person to list" capability end-to-end. The feature now has two rollout layers:

- **`FF_CURATED_LISTS`** controls the core people-lists feature and defaults **true** so the repository, routes, share-menu "Add to list" flow, and membership indicator are visible in normal builds.
- **`FF_PROFILE_LIST_FEATURES`** controls the profile/search/followers/following person-entry surfaces and defaults **false** so those higher-surface-area entry points can stay dark-launched until product wants them on.

Investigation found the feature wasn't "lost in UI updates" — it was dark-launched in Feb but never lit up, and the people-list code path was local-only (`SharedPreferences`, never published to any relay).

**What's new:**

- **`people_lists_repository` package** — Nostr NIP-51 kind 30000 publish/subscribe, Hive-backed local cache with deletion tombstones, NIP-09 kind-5 deletion, submitted-only semantics (never claims relay `OK` confirmation), stale-echo guard on the `watchLists` stream.
- **Global `PeopleListsBloc`** — auth-scoped, provided above `MaterialApp.router` so it covers routes outside the shell. Optimistic mutations, `sequential()`/`droppable()` transformers for race prevention, `listIdsByPubkey` reverse index for O(1) "in N lists" lookups.
- **UI:** draggable `AddToPeopleListsSheet`, full-screen `AddPeopleToListScreen` picker, full-screen `CreatePeopleListPage`, compact `PeopleListMembershipIndicator` chip on other-user profile headers.
- **Entry points wired:** share menu, other-profile "more" sheet, followers & following rows (shared `UserProfileTile`), user search result tile. The share-menu / route-layer surfaces guard on `FF_CURATED_LISTS`; person-entry surfaces guard on `FF_PROFILE_LIST_FEATURES`; all entry points also guard on target-pubkey ≠ own-pubkey.
- **Swipe-to-remove** (long-press) + 4-second undo snackbar on editable lists from `UserListPeopleScreen`.
- **Legacy removal:** deleted the `SharedPreferences`-only `UserListService`, the `FollowSet` class, and the entire follow-set section of `SocialService` (−400 lines). Unified the share menu's duplicate follow-set UI with the new people-list sheet.
- **Localization plumbing is complete** — every user-facing string goes through `context.l10n.*` per `rules/localization.md`.

## Localization status

[...]

PR #6476, opened 2026-07-29
fix(feature-flags): stop the curated-lists toggle from popping the screen

### LINKED ISSUE #6477: fix(feature-flags): stop the curated-lists toggle from popping the screen
Auto-created from PR #6476.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6476
**Branch:** `fix/curated-lists-toggle-pops-screen`

hm21:
Closes #6477

## Description

Users reported that toggling **Curated Lists** in Settings -> Experimental Features threw them back to Settings. No other flag did this.

The app shell watched `isFeatureEnabledProvider(FeatureFlag.curatedLists)` and used it to add a conditional `BlocProvider<PeopleListsBloc>` to `MultiBlocProvider`. A conditional entry changes the shape of the provider chain, so flipping that flag re-inflated the subtree below it: `MaterialApp.router`, feed state, video controllers, and app-shell listeners included.

The fix registers `PeopleListsBloc` unconditionally above `MaterialApp.router`, so the app shell no longer changes structural shape on a curated-lists flag flip.

## Gating after the change

`BlocProvider` remains lazy. To keep the disabled-feature path clean:

- `curatedLists` is now required alongside `profileListFeatures` in all three add-to-list affordance computations: `user_profile_tile.dart`, `search_user_tile.dart`, and `other_profile_screen.dart`.
- `AddToPeopleListsSheet.show` and `showNewPeopleListSheet` check `FeatureFlag.curatedLists` themselves as imperative-entry backstops.
- `other_profile_screen.dart` also re-checks `curatedLists` at the top of the `MoreSheetResult.addToList` branch before reading `PeopleListsBloc`.
- Tests pin that the lazy bloc is not constructed when the feature is off.

## Rebase on current main

This branch is rebased on current `origin/main` and preserves #6482: the app-lifetime `PeopleListsBloc` still receives `peopleListsRepositoryIdentityStreamProvider`, so account/repository changes do not leave it on a stale captured repository.

Remaining lifecycle behavior is explicit and tracked in #6494: if the bloc was already constructed while `curatedLists` was on, it remains session-lifetime after the flag flips off. Follow-up should stop already-started people-list repository work without reintroducing a conditional provider above `MaterialApp.router`.

## Out of Scope

The feature-flag screen still bypasses go_router (#6481), so it does not survive a router refresh from another source.

The Navigator-level step from subtree re-inflation to losing the imperatively pushed route is not fully explained. The symptom was reproduced on device before the change and was gone on device after the provider-shape fix; the comment in `main.dart` scopes the mechanism to what was verified.

## Verification

Local takeover verification:

- `flutter pub get`
- `flutter test test/features/people_lists/view/add_to_people_lists_sheet_test.dart test/widgets/profile/new_people_list_sheet_test.dart test/widgets/user_profile_tile_feature_flag_test.dart test/screens/other_profile_screen_test.dart test/screens/search_results` — 122 passed
- `flutter test test/main_app_shell_repository_sync_test.dart test/providers/provider_identity_stream_test.dart test/providers/list_providers_test.dart test/features/people_lists/bloc/people_lists_bloc_test.dart` — 45 passed
- `flutter test test/features/people_lists/view/add_to_people_lists_sheet_test.dart test/widgets/profile/new_people_list_sheet_test.dart test/widgets/user_profile_tile_feature_flag_test.dart` — 15 passed after l10n assertion cleanup
- Pre-push hook: merge-conflict check passed, analyzer passed, changed-file tests passed

Original author verification: tested on a real Samsung Galaxy that toggling Curated Lists keeps the user on the Experimental Features screen.

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Code refactor
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -2827,19 +2827,25 @@ class _DivineAppState extends ConsumerState<DivineApp>
             //
             // Deliberately not wrapped in `if (curatedLists enabled)`: a
             // conditional entry changes this provider chain's shape, so
-            // flipping the flag re-inflated everything below it — including
-            // the Navigator, which silently dropped imperatively pushed
-            // routes (users toggling the flag were thrown back to Settings).
+            // flipping the flag re-inflated everything below it — the whole
+            // MaterialApp.router subtree, feed state, video controllers, and
+            // the upload-listener dedupe sets with it. The reported symptom
+            // was landing back on Settings from the imperatively pushed
+            // feature-flag screen (#6477); only the re-inflation is
+            // reproduced, the Navigator-level step from there to route loss
+            // is not, and the re-inflation alone justifies this shape.
+            //
             // Laziness does the gating instead: every entry point into the
             // people-lists UI checks FeatureFlag.curatedLists before reading
             // the bloc — the lists routes redirect home, the profile and
             // search affordances stay hidden, and both sheets refuse to
             // open — so nothing constructs it while the flag is off.
             //
-            // ref.read in `create` is safe here: peopleListsRepositoryProvider
-            // is keepAlive over the equally stable nostrServiceProvider, and
-            // without the conditional entry this bloc lives for the whole app
-            // run, so there is no rebuild that could strand the capture.
+            // Known gap (#6480): peopleListsRepositoryProvider is keepAlive
+            // but not identity-stable — it watches nostrServiceProvider,
+            // which recreates its client on auth change. With no conditional
+            // entry `create` never re-runs, so the captured repository is
+            // stale by construction after an account switch.
             BlocProvider(
               create: (_) {
                 final authService = ref.read(authServiceProvider);
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -2829,11 +2829,13 @@ class _DivineAppState extends ConsumerState<DivineApp>
             // conditional entry changes this provider chain's shape, so
             // flipping the flag re-inflated everything below it — the whole
             // MaterialApp.router subtree, feed state, video controllers, and
-            // the upload-listener dedupe sets with it. The reported symptom
-            // was landing back on Settings from the imperatively pushed
-            // feature-flag screen (#6477); only the re-inflation is
-            // reproduced, the Navigator-level step from there to route loss
-            // is not, and the re-inflation alone justifies this shape.
+            // the upload-listener dedupe sets with it. Users were thrown back
+            // to Settings from the imperatively pushed feature-flag screen;
+            // both the symptom and its disappearance under this shape are
+            // verified on device (#6477). What is not pinned is the
+            // Navigator-level step in between — a synthetic probe of the same
+            // shape kept the pushed route — so do not read the chain below
+            // re-inflation as established.
             //
             // Laziness does the gating instead: every entry point into the
             // people-lists UI checks FeatureFlag.curatedLists before reading
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -2837,17 +2837,22 @@ class _DivineAppState extends ConsumerState<DivineApp>
             // shape kept the pushed route — so do not read the chain below
             // re-inflation as established.
             //
-            // Laziness does the gating instead: every entry point into the
-            // people-lists UI checks FeatureFlag.curatedLists before reading
-            // the bloc — the lists routes redirect home, the profile and
-            // search affordances stay hidden, and both sheets refuse to
-            // open — so nothing constructs it while the flag is off.
+            // Laziness does the creation gate instead: every entry point into
+            // the people-lists UI checks FeatureFlag.curatedLists before
+            // reading the bloc — the lists routes redirect home, the profile
+            // and search affordances stay hidden, and both sheets refuse to
+            // open — so a disabled feature does not construct the bloc.
             //
-            // Known gap (#6480): peopleListsRepositoryProvider is keepAlive
-            // but not identity-stable — it watches nostrServiceProvider,
-            // which recreates its client on auth change. With no conditional
-            // entry `create` never re-runs, so the captured repository is
-            // stale by construction after an account switch.
+            // Once constructed while the flag is on, the bloc remains session-
+            // lifetime even if the flag flips off. Stopping already-started
+            // repository work on a flag-off transition is tracked in #6494.
+            //
+            // peopleListsRepositoryProvider is keepAlive but not identity-
+            // stable: it watches nostrServiceProvider, which recreates its
+            // client on auth change. This provider is intentionally still not
+            // keyed; repositoryStream keeps the app-lifetime bloc pointed at
+            // the current repository without re-inflating MaterialApp.router
+            // (#6480, #6482).
             BlocProvider(
               create: (_) {
                 final authService = ref.read(authServiceProvider);
@@ -2856,11 +2861,6 @@ class _DivineAppState extends ConsumerState<DivineApp>
                     .distinct();
                 return PeopleListsBloc(
                   repository: ref.read(peopleListsRepositoryProvider),
-                  // peopleListsRepositoryProvider is keepAlive but not
-                  // identity-stable: it watches nostrServiceProvider, which
-                  // disposes its client on every identity change. This create
-                  // runs once, so the bloc subscribes to the successive
-                  // instances rather than keeping its capture (#6480).
                   repositoryStream: ref.read(
                     peopleListsRepositoryIdentityStreamProvider,
                   ),
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -2687,15 +2687,6 @@ class _DivineAppState extends ConsumerState<DivineApp>
 
     const forceOpenOnboarding = AppConfig.isGhActionsPrPreviewBuild;
 
-    // Gate the global PeopleListsBloc on the curated-lists feature flag.
-    // When enabled we provision it above MaterialApp.router so every route
-    // (including ones outside AppShell) sees the same lists state. The
-    // bloc only depends on the repository and a pubkey stream; we derive
-    // the latter from AuthService.
-    final peopleListsEnabled = ref.watch(
-      isFeatureEnabledProvider(FeatureFlag.curatedLists),
-    );
-
     // Eagerly create the outgoing-DM retry service so its foreground
     // subscription is wired up at app shell setup. The service has no
     // UI consumer (it operates on the durable outgoing_dms queue), so
@@ -2831,28 +2822,38 @@ class _DivineAppState extends ConsumerState<DivineApp>
                 ),
               )..add(const AppUpdateCheckRequested()),
             ),
-            if (peopleListsEnabled)
-              BlocProvider(
-                create: (_) {
-                  final authService = ref.read(authServiceProvider);
-                  final ownerPubkeyStream = authService.authStateStream
-                      .map((_) => authService.currentPublicKeyHex)
-                      .distinct();
-                  return PeopleListsBloc(
-                    repository: ref.read(peopleListsRepositoryProvider),
-                    // peopleListsRepositoryProvider is keepAlive but not
-                    // identity-stable: it watches nostrServiceProvider, which
-                    // disposes its client on every identity change. This create
-                    // runs once, so the bloc subscribes to the successive
-                    // instances rather than keeping its capture (#6480).
-                    repositoryStream: ref.read(
-                      peopleListsRepositoryIdentityStreamProvider,
-                    ),
-                    ownerPubkeyStream: ownerPubkeyStream,
-                    initialOwnerPubkey: authService.currentPublicKeyHex,
-                  )..add(const PeopleListsStarted());
-                },
-              ),
+            // Provisioned above MaterialApp.router so every route (including
+            // ones outside AppShell) sees the same lists state.
+            //
+            // Deliberately not wrapped in `if (curatedLists enabled)`: a
+            // conditional entry changes this provider chain's shape, so
+            // flipping the flag re-inflated everything below it — including
+            // the Navigator, which silently dropped imperatively pushed
+            // routes (users toggling the flag were thrown back to Settings).
+            // Laziness does the gating instead: every consumer checks
+            // FeatureFlag.curatedLists first, so the bloc is never created
+            // while the flag is off.
+            BlocProvider(
+              create: (_) {
+                final authService = ref.read(authServiceProvider);
+                final ownerPubkeyStream = authService.authStateStream
+                    .map((_) => authService.currentPublicKeyHex)
+                    .distinct();
+                return PeopleListsBloc(
+                  repository: ref.read(peopleListsRepositoryProvider),
+                  // peopleListsRepositoryProvider is keepAlive but not
+                  // identity-stable: it watches nostrServiceProvider, which
+                  // disposes its client on every identity change. This create
+                  // runs once, so the bloc subscribes to the successive
+                  // instances rather than keeping its capture (#6480).
+                  repositoryStream: ref.read(
+                    peopleListsRepositoryIdentityStreamProvider,
+                  ),
+                  ownerPubkeyStream: ownerPubkeyStream,
+                  initialOwnerPubkey: authService.currentPublicKeyHex,
+                )..add(const PeopleListsStarted());
+              },
+            ),
           ],
           // Global listener for email verification failures - shows snackbar
           // when verification times out or fails while user is elsewhere in app
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -2830,9 +2830,16 @@ class _DivineAppState extends ConsumerState<DivineApp>
             // flipping the flag re-inflated everything below it — including
             // the Navigator, which silently dropped imperatively pushed
             // routes (users toggling the flag were thrown back to Settings).
-            // Laziness does the gating instead: every consumer checks
-            // FeatureFlag.curatedLists first, so the bloc is never created
-            // while the flag is off.
+            // Laziness does the gating instead: every entry point into the
+            // people-lists UI checks FeatureFlag.curatedLists before reading
+            // the bloc — the lists routes redirect home, the profile and
+            // search affordances stay hidden, and both sheets refuse to
+            // open — so nothing constructs it while the flag is off.
+            //
+            // ref.read in `create` is safe here: peopleListsRepositoryProvider
+            // is keepAlive over the equally stable nostrServiceProvider, and
+            // without the conditional entry this bloc lives for the whole app
+            // run, so there is no rebuild that could strand the capture.
             BlocProvider(
               create: (_) {
                 final authService = ref.read(authServiceProvider);
--- a/mobile/lib/router/routes/lists_routes.dart
+++ b/mobile/lib/router/routes/lists_routes.dart
@@ -72,10 +72,8 @@ List<RouteBase> listsRoutes(Ref ref) {
     // `initialPubkey` query param lets callers (e.g., the share-video
     // "Add to list" sheet) seed the new list with a target person in
     // the same submit so the URL remains reloadable.
-    // Gated on FeatureFlag.curatedLists — the PeopleListsBloc this page
-    // depends on is only provided in main.dart when the flag is on, so
-    // a deep link here with the flag off would crash with
-    // ProviderNotFoundException. Redirect home instead.
+    // Gated on FeatureFlag.curatedLists — a deep link here with the flag
+    // off would open a feature the user turned off, so redirect home.
     GoRoute(
       path: CreatePeopleListPage.path,
       name: CreatePeopleListPage.routeName,
@@ -132,9 +130,8 @@ List<RouteBase> listsRoutes(Ref ref) {
 /// Redirects deep links for people-lists routes to the home feed when the
 /// [FeatureFlag.curatedLists] feature flag is off.
 ///
-/// The people-lists screens depend on a `PeopleListsBloc` that is only
-/// provided in `main.dart` when the flag is enabled. Without this guard a
-/// deep link / push / saved URL would crash with `ProviderNotFoundException`.
+/// Without this guard a deep link / push / saved URL would open the
+/// people-lists screens even though the user turned the feature off.
 /// Returns `null` (no redirect) when the flag is on.
 String? _peopleListsRedirectIfDisabled(Ref ref, GoRouterState state) {
   final enabled = ref.read(isFeatureEnabledProvider(FeatureFlag.curatedLists));
--- a/mobile/lib/features/people_lists/curated_lists_gate.dart
+++ b/mobile/lib/features/people_lists/curated_lists_gate.dart
@@ -1,0 +1,25 @@
+// ABOUTME: Reads FeatureFlag.curatedLists off a BuildContext.
+// ABOUTME: Backstop gate for entry points that would construct PeopleListsBloc.
+
+import 'package:flutter/widgets.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:openvine/features/feature_flags/models/feature_flag.dart';
+import 'package:openvine/features/feature_flags/providers/feature_flag_providers.dart';
+
+/// Whether curated lists are enabled for the current user.
+///
+/// The global `PeopleListsBloc` is registered unconditionally in `main.dart`
+/// — a conditional entry changed the provider chain's shape and re-inflated
+/// the Navigator on every flag flip. `BlocProvider` laziness gates it
+/// instead, which only holds while every entry point that would read the
+/// bloc checks the flag first.
+///
+/// Widgets that can rebuild on a flip should `ref.watch` the flag directly;
+/// this is for imperative call sites (`show...Sheet` helpers) that only need
+/// a one-shot read.
+bool curatedListsEnabled(BuildContext context) {
+  return ProviderScope.containerOf(
+    context,
+    listen: false,
+  ).read(isFeatureEnabledProvider(FeatureFlag.curatedLists));
+}
--- a/mobile/lib/features/people_lists/view/add_to_people_lists_sheet.dart
+++ b/mobile/lib/features/people_lists/view/add_to_people_lists_sheet.dart
@@ -5,7 +5,9 @@ import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:models/models.dart';
+import 'package:openvine/features/feature_flags/models/feature_flag.dart';
 import 'package:openvine/features/people_lists/bloc/people_lists_bloc.dart';
+import 'package:openvine/features/people_lists/curated_lists_gate.dart';
 import 'package:openvine/features/people_lists/models/people_list_entry_point.dart';
 import 'package:openvine/features/people_lists/view/widgets/widgets.dart';
 import 'package:openvine/l10n/l10n.dart';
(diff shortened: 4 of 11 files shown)


mobile/lib/main.dart
```diff
@@ -2831,20 +2822,30 @@ class _DivineAppState extends ConsumerState<DivineApp>
...
+            // Provisioned above MaterialApp.router so every route (including
+            // ones outside AppShell) sees the same lists state.
+            //
+            // Deliberately not wrapped in `if (curatedLists enabled)`: a
+            // conditional entry changes this provider chain's shape, so
+            // flipping the flag re-inflated everything below it — including
+            // the Navigator, which silently dropped imperatively pushed
+            // routes (users toggling the flag were thrown back to Settings).
+            // Laziness does the gating instead: every consumer checks
+            // FeatureFlag.curatedLists first, so the bloc is never created
```
rabble: This claim doesn't hold. `AddToPeopleListsSheet.build` reads the bloc with no flag check (`add_to_people_lists_sheet.dart:78`), and all three of its entry points gate on `profileListFeatures` rather than `curatedLists` (`user_profile_tile.dart:72`, `search_user_tile.dart:48`, `other_profile_screen.dart:263`).

With `profileListFeatures` on and `curatedLists` off — both user-facing toggles in the same Experimental Features list — the sheet now constructs the bloc, fires `PeopleListsStarted`, and opens a relay subscription for a disabled feature. The `ProviderNotFoundException` that used to happen there was what enforced the gate; removing the `if` removed the enforcement without replacing it.

Either guard the sheet on `curatedLists` directly, or `&&` the flag into the three `showAddToList` computations, and then this comment becomes accurate.

rabble (dismissed):
Nice root-cause writeup — the diagnosis (conditional entry changes the provider-chain shape, subtree re-inflates, Navigator drops the imperative route) is correct and the minimal fix is the right shape. With the flag on, behaviour is byte-identical because `BlocProvider` was already lazy before this change, so I agree there's no regression risk on the default path.

One blocking issue, because it invalidates the safety argument the fix rests on.

## Blocking: the gating invariant isn't actually true

The new comment asserts:

> Laziness does the gating instead: every consumer checks FeatureFlag.curatedLists first, so the bloc is never created while the flag is off.

That isn't the case. `AddToPeopleListsSheet` has no flag check at all — `build` goes straight to `context.select<PeopleListsBloc, List<UserList>>` (`mobile/lib/features/people_lists/view/add_to_people_lists_sheet.dart:78`), and its own doc comment just says callers must provide the bloc above it. Every one of its three entry points gates on `profileListFeatures`, **not** `curatedLists`:

- `mobile/lib/widgets/user_profile_tile.dart:72` — `showAddToList = profileListFeaturesEnabled && !isCurrentUser`
- `mobile/lib/screens/search_results/widgets/search_user_tile.dart:48` — `showAddToList = profileListFeaturesEnabled && profile.pubkey != ownPubkey`
- `mobile/lib/screens/other_profile_screen.dart:258-263` — reads `FeatureFlag.profileListFeatures` into a local *named* `curatedListsEnabled` (see nit below), then `context.read<PeopleListsBloc>().state.lists` at :293

So with **`profileListFeatures` on + `curatedLists` off** — both `FeatureFlagAudience.user`, both toggleable side by side in Settings → Experimental Features, so this is a state any user can reach — tapping "Add to list" now constructs the bloc, which fires `PeopleListsStarted`, which spins up the repository and a relay subscription for a feature the user explicitly turned off. The user also gets a fully working list-management sheet for that feature.

The PR body does mention this path, but frames it as a pure win:

> Side effect: with `profileListFeatures` on and `curatedLists` off, the profile "Add to list" action called `context.read<PeopleListsBloc>()` and hit a `ProviderNotFoundException`. That latent crash is gone.

The crash is gone, but what replaced it is silent activation of a disabled feature rather than correct behaviour. Previously the `ProviderNotFoundException` was load-bearing: it was the thing enforcing the gate, loudly. Removing the `if` removes the enforcement, and nothing took over.

Suggested fix — either is fine:

1. Guard the sheet itself: make `AddToPeopleListsSheet` a `ConsumerWidget` (or check in `show`) and return early / no-op when `curatedLists` is off. This also covers future call sites.
2. Fold the flag into the three `showAddToList` computations: `curatedListsEnabled && profileListFeaturesEnabled && ...`.

Option 1 is more robust; option 2 also fixes the affordance being visible at all, which is arguably the real bug.

The same applies to `showNewPeopleListSheet` (`mobile/lib/widgets/profile/new_people_list_sheet.dart:86` reads the bloc) reached from `other_profile_screen.dart:307` and the sheet's own create button.

## Test: a meaningful regression test does exist

> No regression test: a meaningful one would have to pump the whole app shell, and a narrower test would only pin Flutter's element-reconciliation behaviour rather than fail if the `if` were reintroduced.

Agreed on the reconciliation half — that test would be worthless. But the invariant this fix *depends on* is testable and is not about reconciliation:

> with `curatedLists` off, no code path constructs `PeopleListsBloc`

Pump a tile or the sheet with `curatedLists` overridden off and assert the repository is never touched / no `PeopleListsStarted` is dispatched. That's a behaviour test of the gating contract, it's cheap, it doesn't need the app shell — and it would fail on this branch today, which is the best argument for it. Given AGENTS.md asks for a regression test for the confirmed failure mode when feasible, this seems like the one to add.

## Nit (non-blocking): misleading local name

`mobile/lib/screens/other_profile_screen.dart:258` reads `FeatureFlag.profileListFeatures` into a local named `curatedListsEnabled`. Pre-existing, but it's plausibly the thing that made the reasoning in this PR go sideways, and renaming it is local and low-risk while you're in this area.

## Nit (non-blocking): the repository capture is now app-lifetime

`ref.read(peopleListsRepositoryProvider)` in `create:` (`mobile/lib/main.dart:2843`) is the `ref.read`-in-`create` shape `.claude/rules/state_management.md` warns about. It's unchanged by this diff and low-risk in practice — the provider is `@Riverpod(keepAlive: true)` over a stable `nostrServiceProvider`, so exception #1 ("genuinely stable for the bloc's lifetime") applies. But removing the `if` means the bloc now genuinely lives for the whole app run with no re-inflation path at all, so the rule asks for the stability assumption to be documented inline. One sentence in the existing comment block would close it.

## Agreed and out of scope

Leaving the feature-flag screen on `Navigator.push` rather than converting it to a go_router route is the right call for this PR — your point that the teardown killed feed state and video controllers, not just the pushed screen, is the correct reason to fix the provider shape instead.

realmeylisdev (commented):
Root cause and fix shape look right to me, and I reproduced the subtree re-inflation this removes (conditional entry: 2 inflations on a flip; unconditional: 1).

Two things worth resolving: the new comment's gating invariant doesn't hold — three entry points gate on `profileListFeatures`, not `curatedLists` — and the route-loss half of the mechanism didn't reproduce for me. Also a correction on the repository-capture nit above. Details inline.

mobile/lib/main.dart
```diff
@@ -2831,20 +2822,30 @@ class _DivineAppState extends ConsumerState<DivineApp>
...
+            // while the flag is off.
```
realmeylisdev: This invariant doesn't hold, and it's the one the rest of the change rests on.

Three of the four `PeopleListsBloc` entry points gate on `FeatureFlag.profileListFeatures`, not `curatedLists`:

- `mobile/lib/widgets/user_profile_tile.dart:72` — `showAddToList = profileListFeaturesEnabled && !isCurrentUser`
- `mobile/lib/screens/search_results/widgets/search_user_tile.dart:48` — same shape
- `mobile/lib/screens/other_profile_screen.dart:258-263` — reads `profileListFeatures`, then `context.read<PeopleListsBloc>()` at :294

Only `PeopleListMembershipIndicator` checks `curatedLists` before touching the bloc (`people_list_membership_indicator.dart:37-41`). `AddToPeopleListsSheet` has no flag check at all — `build` goes straight to `context.select<PeopleListsBloc, List<UserList>>` at :78.

Both flags are `FeatureFlagAudience.user`, so `profileListFeatures` on + `curatedLists` off is two toggles away from the shipped defaults, on the very screen this PR is about. In that state tapping "Add to list" constructs the bloc, which fires `PeopleListsStarted` and lands in `_onOwnerChanged` → `_repository.watchLists()` plus `_repository.syncOwner()` (`people_lists_bloc.dart:151-165`) — a relay subscription and a sync for a feature the user switched off, plus a fully working sheet that can publish kind-30000 events.

Worth pulling in the original contract: #3244, which introduced both flags, documents `FF_CURATED_LISTS` as the master switch — *"Use `--dart-define=FF_CURATED_LISTS=false` to hide the core feature entirely"*, with person-entry surfaces as a sub-layer on top. That contract is already broken on `main`; the difference is that there the `ProviderNotFoundException` was accidentally enforcing it. Removing the `if` removes the enforcement and nothing takes over.

Smallest change that makes this comment true: add `curatedListsEnabled &&` to the three `showAddToList` computations. Hiding the affordance is probably the more correct behaviour anyway — right now the button is visible for a disabled feature either way.

mobile/lib/main.dart
```diff
@@ -2831,20 +2822,30 @@ class _DivineAppState extends ConsumerState<DivineApp>
...
-                },
-              ),
+            // Provisioned above MaterialApp.router so every route (including
+            // ones outside AppShell) sees the same lists state.
+            //
+            // Deliberately not wrapped in `if (curatedLists enabled)`: a
+            // conditional entry changes this provider chain's shape, so
+            // flipping the flag re-inflated everything below it — including
+            // the Navigator, which silently dropped imperatively pushed
+            // routes (users toggling the flag were thrown back to Settings).
```
realmeylisdev: I tried to reproduce this and got a split result — the first half holds, the second I couldn't get.

I built a minimal probe with the same shape: conditional entry in a `MultiBlocProvider` whose `child:` is `MaterialApp.router`, `/settings` reached via go_router, then a `MaterialPageRoute` pushed imperatively on top, with an inflation counter sitting on the `child:`.

```
conditional entry   -> inflations 1 -> 2, pushed route still visible after the flip
unconditional entry -> inflations 1 -> 1
```

So *"flipping the flag re-inflated everything below it"* is confirmed, and that on its own justifies the change. It's also broader than the PR body says: `GeoBlockingGate`, `AppLifecycleHandler` and `UploadFailureListener`'s `_lastKnownFailedIds` / `_lastKnownSucceededIds` dedupe sets are all rebuilt too, so a flip could plausibly replay an upload-failure sheet or a "published" snackbar.

What I couldn't reproduce is the route loss. go_router keys its navigator host with a `GlobalObjectKey` and the `Navigator` with the key you pass it — here `NavigatorKeys.root` (`go_router/lib/src/builder.dart:109-112` and :460-461) — and Flutter reparents global-keyed elements rather than recreating them, so the pushed route survived the re-inflation in my probe.

I'm not suggesting the symptom isn't real; you reproduced it on a device and users reported it. But this comment names a specific causal chain that the next reader will take at face value, and something else in the real app must be completing it. You mention a throwaway widget test that did reproduce the mechanism and wasn't kept — restoring it would settle this and pin the `1 -> 1` invariant, which is what the fix actually depends on. Happy to hand over my probe script if it saves time.

Separately, on the deferred go_router conversion for `FeatureFlagScreen`: it's the only tile in `settings_screen.dart` still using `Navigator.push(MaterialPageRoute(...))` (:324-329) while all fifteen siblings use `context.push`. Agreed it's out of scope, but AGENTS.md asks deferrals to carry a tracking issue.

mobile/lib/main.dart
```diff
@@ -2831,20 +2822,30 @@ class _DivineAppState extends ConsumerState<DivineApp>
...
+            // FeatureFlag.curatedLists first, so the bloc is never created
+            // while the flag is off.
+            BlocProvider(
+              create: (_) {
+                final authService = ref.read(authServiceProvider);
+                final ownerPubkeyStream = authService.authStateStream
+                    .map((_) => authService.currentPublicKeyHex)
+                    .distinct();
+                return PeopleListsBloc(
+                  repository: ref.read(peopleListsRepositoryProvider),
```
realmeylisdev, line 2863: One correction to the review thread above, because it changes what the right follow-up is.

`peopleListsRepositoryProvider` is `keepAlive`, but it is **not** identity-stable. It does `ref.watch(nostrServiceProvider)` (`mobile/lib/providers/repository_providers.dart:450`), and `NostrService` is a notifier that by its own doc *"react[s] to auth state changes and recreate[s] the client when the keyContainer changes (e.g., user signs out and signs in with different keys)"* (`mobile/lib/providers/nostr_client_provider.dart:196-197`). Every recreation recomputes this provider and yields a fresh `PeopleListsRepositoryImpl`.

So the `ref.read` here captures a repository whose `NostrClient` is dead after the next account switch, and `_repository` is a `final` field on the bloc — it can never pick up the newer instance. `ownerPubkeyStream` will happily fire `PeopleListsOwnerChanged` for the new owner and route it through the stale repository.

This is pre-existing and not introduced by this diff, so I wouldn't ask you to fix it here. But it does mean "document the stability assumption inline" would enshrine something that isn't true — exception #1 in `state_management.md` doesn't apply. The rule's actual cure is the record-typed `ValueKey` over the captured deps. Worth a tracking issue: with the `if` gone, nothing re-runs `create:` at all, so the staleness is now permanent by construction rather than merely likely.

mobile/lib/router/routes/lists_routes.dart
```diff
@@ -72,10 +72,8 @@ List<RouteBase> listsRoutes(Ref ref) {
     // `initialPubkey` query param lets callers (e.g., the share-video
     // "Add to list" sheet) seed the new list with a target person in
     // the same submit so the URL remains reloadable.
-    // Gated on FeatureFlag.curatedLists — the PeopleListsBloc this page
-    // depends on is only provided in main.dart when the flag is on, so
-    // a deep link here with the flag off would crash with
-    // ProviderNotFoundException. Redirect home instead.
+    // Gated on FeatureFlag.curatedLists — a deep link here with the flag
+    // off would open a feature the user turned off, so redirect home.
```
realmeylisdev, line 76: Comment update is accurate, and dropping the `ProviderNotFoundException` justification was the right call.

One thing it's worth being explicit about: this guard is route-level only. `AddToPeopleListsSheet` and `showNewPeopleListSheet` are modals rather than routes, so they never pass through `_peopleListsRedirectIfDisabled`. That's exactly how the `profileListFeatures`-on / `curatedLists`-off path gets in — same gap as the `main.dart` thread.

mobile/lib/main.dart
```diff
@@ -2831,20 +2822,37 @@ class _DivineAppState extends ConsumerState<DivineApp>
...
+            // Laziness does the gating instead: every entry point into the
+            // people-lists UI checks FeatureFlag.curatedLists before reading
+            // the bloc — the lists routes redirect home, the profile and
+            // search affordances stay hidden, and both sheets refuse to
+            // open — so nothing constructs it while the flag is off.
+            //
+            // ref.read in `create` is safe here: peopleListsRepositoryProvider
+            // is keepAlive over the equally stable nostrServiceProvider, and
+            // without the conditional entry this bloc lives for the whole app
+            // run, so there is no rebuild that could strand the capture.
```
realmeylisdev: Sorry — I think my earlier note was read as "add a comment saying it's safe", and this documents the opposite of what's actually true. Both clauses don't hold.

**`nostrServiceProvider` is not stable.** `NostrService` reassigns its state on identity change — `state = newClient;` at `mobile/lib/providers/nostr_client_provider.dart:644` in the auth-state-change handler, and again at :506 on the init-retry path, each preceded by `_disposeClient(oldClient)` → `client.dispose()` (:310-314). Since `peopleListsRepositoryProvider` does `ref.watch(nostrServiceProvider)` (`repository_providers.dart:450`), every one of those reassignments recomputes it into a **new** `PeopleListsRepositoryImpl`. `keepAlive` keeps the provider from being disposed; it doesn't pin the value.

**The second clause is inverted.** "This bloc lives for the whole app run, so there is no rebuild that could strand the capture" — the absence of a rebuild is exactly what strands it. `_repository` is a `final` field on `PeopleListsBloc`, so once `create` captures instance N it can never see N+1. Before this PR a `curatedLists` flip was the one thing that re-ran `create`; now nothing does, so after the first account switch the bloc talks to the previous, disposed client permanently. `ownerPubkeyStream` still fires `PeopleListsOwnerChanged` for the new owner and routes `watchLists` / `syncOwner` through the stale repository (`people_lists_bloc.dart:151-165`).

I didn't chase the exact failure mode inside `nostr_sdk` — I looked for an `_isDisposed` early-return in `queryEventsDetailed` and didn't find one, so I can't tell you whether it returns empty, throws, or tries to reconnect. Worth establishing before writing any safety claim down.

This is pre-existing and genuinely not this PR's job to fix — `.claude/rules/state_management.md` has the cure (record-typed `ValueKey` over the captured deps, or the Page/View split). My only ask is that the comment not assert the opposite. Either drop these four lines, or replace them with something like:

```dart
// NOTE: peopleListsRepositoryProvider is rebuilt whenever
// nostrServiceProvider swaps its client (account switch, re-auth), but
// `create` only runs once, so this capture goes stale. Tracked in #XXXX —
// see state_management.md "Bridging Riverpod-provided dependencies".
```

realmeylisdev (commented):
Gating hole is closed properly — both levels, plus the `blocCreated` regression test. Verified green locally.

One new thing: the `ref.read` safety comment added in this commit states the opposite of what the code does — `nostrServiceProvider` does swap its client, and the never-rebuilding bloc is what makes the stale capture permanent rather than safe. Inline.

mobile/lib/main.dart
```diff
@@ -2831,20 +2822,37 @@ class _DivineAppState extends ConsumerState<DivineApp>
...
+            // Deliberately not wrapped in `if (curatedLists enabled)`: a
+            // conditional entry changes this provider chain's shape, so
+            // flipping the flag re-inflated everything below it — including
+            // the Navigator, which silently dropped imperatively pushed
+            // routes (users toggling the flag were thrown back to Settings).
+            // Laziness does the gating instead: every entry point into the
+            // people-lists UI checks FeatureFlag.curatedLists before reading
+            // the bloc — the lists routes redirect home, the profile and
+            // search affordances stay hidden, and both sheets refuse to
+            // open — so nothing constructs it while the flag is off.
```
realmeylisdev: This is a better fix than I asked for — gating the affordance *and* backstopping both `show` helpers closes it for future call sites too, and `curated_lists_gate.dart` is the right shape for the imperative reads.

The `blocCreated` assertions in `add_to_people_lists_sheet_test.dart` are exactly the regression test the invariant needed, and they'd have gone red on 26958b31. Verified locally: `flutter analyze` clean on the six changed files, 98/98 passing across `test/features/people_lists`, `new_people_list_sheet_test.dart`, `user_profile_tile_feature_flag_test.dart` and `other_profile_screen_test.dart`.

One thread still open from the last round, non-blocking and no need to re-litigate here: the mechanism sentence just above still asserts the Navigator dropped the pushed route, which I couldn't reproduce. The `1 -> 1` half is the invariant this fix actually rests on and is now nowhere pinned. If the throwaway probe is still lying around, it'd be worth keeping.

mobile/lib/main.dart
hm21: Fixed in b860c2ea2c — both options, since they cover different levels.

Option 2: `curatedLists` folded into all three `showAddToList` computations, so the affordance disappears with the feature. Option 1: `AddToPeopleListsSheet.show` and `showNewPeopleListSheet` check the flag themselves via a shared `curatedListsEnabled(context)` helper, so a future call site cannot reopen the hole.

The comment here now names the actual gating chain instead of asserting a blanket invariant.

mobile/lib/main.dart
hm21: Agreed, fixed in b860c2ea2c. The #3244 reference settles the direction — `FF_CURATED_LISTS` as master switch with person-entry surfaces layered on top is also what the defaults say (`curatedLists` on, `profileListFeatures` off), so requiring both is the coherent reading rather than a new rule.

Done on both levels: `curatedLists &&` in all three `showAddToList` computations, plus a flag check inside `AddToPeopleListsSheet.show` and `showNewPeopleListSheet` so future call sites are covered too.

mobile/lib/main.dart
hm21: You are right, and the comment now says so — 236cf21e10.

I checked your reading against the source and it holds: `RouteBuilder` keys `_CustomNavigator` with `GlobalObjectKey(configuration.navigatorKey.hashCode)` and passes `navigatorKey` straight to the `Navigator` (`builder.dart:109-112`, `:304-305`). We pass `NavigatorKeys.root`, which is a process-wide `static final GlobalKey`, so both keys are stable across the re-inflation and Flutter reparents rather than recreates. Nothing there should drop a pageless route.

So I stopped asserting the chain. The comment now states what is reproduced — the subtree re-inflation, with the wider blast radius you found — notes the reported symptom against #6477, and says explicitly that the Navigator-level step to route loss is not pinned. The re-inflation alone justifies the shape, which is the part the fix depends on.

On restoring the probe: a synthetic `MultiBlocProvider` in a test pins Flutter reconciliation rather than our shell, which is the objection rabble raised against the narrow version — it would stay green if someone reintroduced the `if` in `main.dart`, because the test builds its own provider list. Pinning the real `1 -> 1` invariant needs the actual shell, which is the test we both agreed is not worth pumping. I would rather not ship a test that looks like it guards the regression and does not.

`settings_screen.dart:324` confirmed as the only `Navigator.push` among sixteen tiles — tracked in #6481.

mobile/lib/main.dart
hm21: Correct, and thanks for catching it before it got enshrined — that sentence went in on this branch in response to the nit above, so it would have shipped as a documented falsehood.

Verified: `peopleListsRepositoryProvider` does `ref.watch(nostrServiceProvider)` (`repository_providers.dart:450`), and `NostrService` reassigns state at `nostr_client_provider.dart:506` and `:644` on auth change. Fresh repository per account switch, `final _repository` on the bloc keeps the dead one.

236cf21e10 replaces the stability claim with a known-gap note pointing at #6480, which also records why the record-typed `ValueKey` is the wrong cure here specifically: this `BlocProvider` sits above `MaterialApp.router`, so re-keying it would re-inflate the whole app subtree on every account switch — the exact blast radius this PR removes. Exception #2 (bloc outlives the change, subscribes to it) fits better.

mobile/lib/router/routes/lists_routes.dart
hm21: Right, and that gap is now closed at the modal level rather than left to the route guard — b860c2ea2c adds a `curatedLists` check inside `AddToPeopleListsSheet.show` and `showNewPeopleListSheet`, so both refuse to open regardless of how they were reached.

I left the redirect comment route-scoped rather than widening it, since the modals are guarded in their own files now.

hm21:
@rabble @realmeylisdev all blockers addressed, details in the threads.

**b860c2ea2c** — the gating hole. Both suggested options, since they cover different levels: `curatedLists` folded into the three `showAddToList` computations so the affordance disappears, plus a flag check inside `AddToPeopleListsSheet.show` and `showNewPeopleListSheet` so future call sites are covered. Regression tests for the invariant, all three verified to fail with the guards removed. Local rename in `other_profile_screen` done.

**236cf21e10** — the comment stopped claiming two things that did not survive checking: the Navigator-level causal chain (your reading of the go_router keying is right, @realmeylisdev, I verified it in the source) and the `ref.read` stability assumption. That second one is the awkward one — it went in on this branch in response to the nit, so it would have shipped as a documented falsehood. Now a known-gap note instead.

Two tracking issues for the deferrals: #6480 (stale repository capture, now permanent by construction) and #6481 (`FeatureFlagScreen` off `Navigator.push`).

The one thing I did not do is restore the probe as a test — reasoning in the thread: a synthetic provider list pins Flutter reconciliation rather than our shell, so it would stay green if someone reintroduced the `if`.

mobile/lib/main.dart
hm21: One correction to my own reply, because it undersold the evidence — 2b4253621e.

I wrote that "only the re-inflation is reproduced", which reads as if it were open whether this shape fixes anything. It is not. The symptom was reproduced on device before the change and is gone on device after it, on the original commit, before any of the review fixes. Your probe leaves the *mechanism* unexplained, not the outcome — and you said as much yourself; I am the one who let it drift in the comment.

So the wording now keeps both: device-verified symptom and fix, unexplained Navigator-level step in between, with the probe result recorded so nobody reads the chain as established. Everything else in my reply stands, including the go_router keying being exactly as you described.

realmeylisdev (approved):
All three threads resolved, and the comment block now says only what the code actually does.

Re-walked every `PeopleListsBloc` consumer on `236cf21e1` — all eight are gated: the three screens only build from routes carrying `_peopleListsRedirectIfDisabled`, `PeopleListMembershipIndicator` early-returns on the flag, `PeopleListRow` only renders inside the sheet, the three affordances now require `curatedLists`, and both `show` helpers refuse. `flutter analyze` clean; 98/98 across `test/features/people_lists`, `new_people_list_sheet_test.dart`, `user_profile_tile_feature_flag_test.dart` and `other_profile_screen_test.dart`.

Scoping the mechanism claim to what's reproducible rather than deleting it was the right call, and #6480 / #6481 are better outcomes than inline justifications would have been — #6480 in particular makes the case that the `ValueKey` cure is the wrong trade above `MaterialApp.router`, which I agree with.

Approving from my side. @rabble's changes-requested from the first round is still open and only they can clear it.

NotThatKindOfDrLiz:
@hm21 I took over the branch and pushed `71b01d2f0`.

What changed:
- rebased onto current `origin/main` and preserved #6482 by keeping `peopleListsRepositoryIdentityStreamProvider` wired into the unconditional `PeopleListsBloc`
- added the missing `curatedLists` backstop at the top of the `OtherProfileView` `addToList` branch before it reads `PeopleListsBloc`
- corrected the people-lists comments from "relay subscription" to "relay query and cache subscription"
- switched the new people-list sheet assertions, and the touched add-to-list sheet assertions, to `lookupAppLocalizations(const Locale('en'))`
- documented the already-created-bloc flag-off lifecycle behavior and tracked the follow-up in #6494
- updated the PR body with the current rebase/verification state

Verification:
- `flutter pub get`
- `flutter test test/features/people_lists/view/add_to_people_lists_sheet_test.dart test/widgets/profile/new_people_list_sheet_test.dart test/widgets/user_profile_tile_feature_flag_test.dart test/screens/other_profile_screen_test.dart test/screens/search_results` - 122 passed
- `flutter test test/main_app_shell_repository_sync_test.dart test/providers/provider_identity_stream_test.dart test/providers/list_providers_test.dart test/features/people_lists/bloc/people_lists_bloc_test.dart` - 45 passed
- `flutter test test/features/people_lists/view/add_to_people_lists_sheet_test.dart test/widgets/profile/new_people_list_sheet_test.dart test/widgets/user_profile_tile_feature_flag_test.dart` - 15 passed after the l10n assertion cleanup
- `flutter analyze lib test` - no issues
- pre-push hook passed merge-conflict check, analyzer, and changed-file tests
- GitHub CI is green on `71b01d2f0`

Please review the takeover changes and merge when you are ready. @rabble your earlier changes-requested review is still the active blocker; the remaining branch now addresses the residual add-to-list read-before-guard hole and records the flag-off lifecycle follow-up in #6494.

NotThatKindOfDrLiz (approved):
Final takeover pass is clean. The branch is rebased on current main, preserves #6482 repository-stream behavior, closes the remaining add-to-list gating hole, corrects the stale comments/l10n assertions, and CI is green on 71b01d2f0.

PR #6478, opened 2026-07-29
fix(publish): stop a video from publishing under the wrong account

### LINKED ISSUE #6479: fix(publish): stop a video from publishing under the wrong account
Auto-created from PR #6478.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6478
**Branch:** `fix/publish-under-wrong-account-after-switch`

hm21:
Closes #6479

## Description

Reported from build 1.0.17 (668) on a Samsung Galaxy S22: start an upload on a throwaway account, switch to the main account mid-upload, get "upload failed / try again" — and the retry publishes the video under the **main** account.

Two independent causes, both fixed here.

**1. `getDraftById` was the only draft query without an owner filter.** Every sibling query uses `_ownedOrLegacy`; this one did not. `swapAccount` carries the current route into the swapped-in container, so after the switch the restored screen loaded the leaving account's draft and offered a retry under the new identity. Scoped it like its siblings — the unscoped form stays available (no `ownerPubkey` argument) precisely so ownership can be probed.

**2. The publish path resolves the author at retry time, not at record time.** `publishVideo` reads `authService.currentPublicKeyHex` when it runs, and `PendingUpload.nostrPubkey` was only ever used for list scoping, never as an assertion. `publishVideo` now refuses a draft owned by another account with a new `PublishErrorKind.accountChanged`.

The guard deliberately runs **before** `saveDraft(publishing)`: that upsert writes `ownerPubkey`, so publishing a foreign draft silently re-owned it on the way in and made the leak self-confirming. Checking afterwards would always pass.

The guard is kept independent of the confirmation sheet below, because not every account change goes through a dialog — session expiry and sign-out don't — and "Switch anyway" must not mean "post under the wrong name".

`VideoPublishService.retryUpload` is **deleted** rather than guarded. It had no production callers on `main` (only its own test group), so it was a second, unguarded entry point into the publish path — removing it is worth more than adding a guard to it.

**3. Confirmation before the switch.** The leaving account's `ProviderContainer` owns the `UploadManager` and is disposed on swap, so an in-flight upload genuinely cannot survive the switch. The settings sheet already warned about unsaved drafts; it now also warns about in-flight uploads and, on confirm, parks each video back as a **draft of the account it was recorded on** rather than leaving it as a failure the wrong account is invited to retry. The two sheets share one `_confirmSwitchAnyway` helper; the upload warning supersedes the drafts one when both apply.

Parking goes through a new awaitable `BackgroundPublishBloc.parkInFlight()`, not through dispatching `BackgroundPublishVanished` per upload. `add` is fire-and-forget and the handler does real async writes, while `swapAccount` tears down the container the bloc lives in — so an event-based park races the teardown and loses the video on anything slower than a single fast write. `parkInFlight()` awaits the writes first and only then dispatches the events for the in-memory cleanup; re-parking is a no-op, so the redundant handler pass is harmless.

Copy for `publishErrorAccountChanged`, `settingsUploadInProgressTitle` and `settingsUploadInProgressMessage` is translated in all 18 locales, not left as English fallback.

**Related Issue:** 

## Out of Scope

The autosave draft id is a shared constant (`'draft_autosave'`), so autosave rows collide across accounts: the second account's save overwrites the first account's row and deletes the clip files that row no longer references. This PR does **not** change that. The orphan probe in `saveDraft` has to stay unscoped — the upsert underneath it is keyed on `id` alone and destroys whatever row is there, so an owner-scoped read would skip cleanup for the very files it just orphaned. What the owner scoping does change is that the colliding row is no longer *readable* across accounts, not that the collision stops being destructive. Worth its own issue, not folded in here.

## Verification

- `flutter test test/screens/settings/ test/services/draft_storage_service_test.dart test/services/video_publish/ test/blocs/background_publish_bloc/ test/widgets/upload_failure_listener_test.dart test/widgets/upload_failure_sheet_test.dart test/l10n/` — passing
- `flutter test test/providers/video_editor_provider_test.dart test/blocs/drafts_library` — 149 passing (draft-loading callers of the newly scoped query)
- `flutter test` in `packages/db_client` — 880 passing
- New tests: DAO owner scoping for `getDraftById`, both scoped and unscoped; `isDraftOwnedByAnotherAccount` across foreign / own / unknown / legacy rows; `updatePublishStatus` pinning that parking never re-attributes the row it writes; both publish-service guards asserting no upload and no Nostr event is attempted; `parkInFlight` covering only unfinished uploads and not resolving until the writes land; widget tests for the switch sheet including that the swap does not start while the park is still outstanding.

Manually verified on a real Samsung Galaxy: upload on account A → switch → warning appears; switching anyway leaves nothing to retry on account B; the video is waiting as a draft back on account A.

## Type of Change


mobile/lib/screens/settings/settings_screen.dart
```diff
@@ -162,6 +193,15 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
             onTap: () async {
               Navigator.of(context).pop();
               if (account.pubkeyHex == accountState.currentPubkey) return;
+
+              // Park now that a switch is actually committed. Re-read the
+              // queue instead of reusing the ids the warning was built from,
+              // so an upload that finished while the picker was open is left
+              // alone.
+              for (final draftId in _inFlightDraftIds(publishBloc)) {
+                publishBloc.add(BackgroundPublishVanished(draftId: draftId));
```
rabble: This is fire-and-forget into a bloc that `swapAccount` (4 lines down) is about to dispose.

`BackgroundPublishVanished` uses the default **concurrent** transformer, and `_onBackgroundPublishVanished` awaits `draftExists(sourceDraftId)` then `_deleteDraft`/`_persistPublishStatus`. `BackgroundPublishBloc` is created in the `MultiBlocProvider` inside the `ContainerSwapHost` scope — the container `swapAccount` tears down. Nothing awaits the handler.

If disposal wins, the park write never lands and the video is lost, which is the exact outcome this is meant to prevent. It'll pass with one upload on fast storage and fail with several queued or on a slower device.

The park needs to be awaited before `swapAccount` — either write it directly through the draft service here, or expose an awaitable `parkInFlight()` on the bloc. The handler also `emit`s before its first await, so a lost race risks emit-after-close too.

mobile/lib/services/video_publish/video_publish_service.dart
```diff
@@ -200,6 +200,20 @@ class VideoPublishService {
   /// Publishes a video draft.
   /// Returns [PublishSuccess] on success, [PublishError] on failure.
   Future<PublishResult> publishVideo({required DivineVideoDraft draft}) async {
+    // An account switch while the upload was in flight leaves the previous
+    // account's draft reachable from this session. Publishing it would sign the
+    // video with the wrong identity, and the `publishing` save below would
+    // reassign the draft's owner on the way — so bail out before either.
+    if (await draftService.isDraftOwnedByAnotherAccount(draft.id)) {
```
rabble, line 207: Putting this ahead of `saveDraft(publishing)` is the crux of the fix and it's right — the upsert is keyed on `id` alone and writes `ownerPubkey`, so a check after it would re-own the row and always pass. Good that the reasoning is in the comment; it's the kind of ordering a later refactor would happily "tidy" into a bug.

One note on what this guard depends on: `_ownedOrLegacy(column, null)` returning `const Constant(true)` (`drafts_dao.dart:41`) is what makes the unscoped read inside `isDraftOwnedByAnotherAccount` actually see foreign rows. If that ever became `column.isNull()`, this guard would still compile and still pass its legacy-row tests while never firing in the real scenario. Worth a test asserting the unscoped DAO read returns a row owned by a *different* pubkey, if there isn't one already.

rabble (dismissed):
Excellent diagnosis. Two independent causes correctly identified, and the single most important detail — ordering the guard **before** `saveDraft(publishing)` because the upsert re-owns the row and would make a post-check always pass — is exactly right and would have been very easy to get backwards. The defence-in-depth split (scope the query *and* assert at publish time) is the right instinct for a wrong-identity bug.

Two things I think block, and two places where the description doesn't match the diff.

## Blocking: the park races the container teardown

`settings_screen.dart:201-203` fires the parking events and then immediately tears down the container that owns the bloc receiving them:

```dart
for (final draftId in _inFlightDraftIds(publishBloc)) {
  publishBloc.add(BackgroundPublishVanished(draftId: draftId));
}

final deviceScope = ref.read(deviceScopeProvider);
try {
  await swapAccount(...);
```

`add` is fire-and-forget. `BackgroundPublishVanished` is registered with the default **concurrent** transformer (`background_publish_bloc.dart:32` — only `BackgroundPublishRequested` gets `sequential()`), and `_onBackgroundPublishVanished` is `async` with real awaits: `draftExists(sourceDraftId)`, then `_deleteDraft` or `_persistPublishStatus`. Nothing awaits handler completion.

Meanwhile `BackgroundPublishBloc` is created inside the `MultiBlocProvider` in `_DivineAppState.build`, i.e. **inside** the `ContainerSwapHost` scope — the very container `swapAccount` disposes. So the ordering is: dispatch N events → start disposing the bloc's container. If disposal wins the race, the park write never lands and the video is gone, which is the precise outcome this feature exists to prevent.

This is exactly the shape that passes the manual test you ran (one upload, one fast SQLite write) and fails with several queued uploads or on a slower device — which is the same class of device the original report came from. The handler also `emit`s before its awaits, so a lost race additionally risks an emit-after-close.

The fix wants the park to be awaited before `swapAccount`: either park directly through the draft service at the call site, or give the bloc an awaitable park (`Future<void> parkInFlight()`) and `await` it. Firing events into a bloc you are about to close can't be made safe by ordering alone.

## Blocking: the parking service can be the wrong account's

`main.dart:2791-2795` builds the bloc with:

```dart
BlocProvider(
  create: (_) => BackgroundPublishBloc(
    videoPublishServiceFactory: createPublishService,
    draftStorageService: ref.read(draftStorageServiceProvider),
    foregroundSession: ref.read(publishForegroundSessionProvider),
  ),
),
```

`ref.read` in `BlocProvider.create` with no `ValueKey` — the documented anti-pattern in `.claude/rules/state_management.md` ("Bridging Riverpod-provided dependencies into BlocProvider"); the rule's own detection grep matches this line. And the dependency is explicitly identity-flippable by design — `social_providers.dart:602-614`:

```dart
// Rebuild when account changes so ownerPubkey stays current
ref.watch(currentAuthStateProvider);
```

with `DraftStorageService.ownerPubkey` a `final` captured at construction. So the provider hands out a fresh instance on auth change while the bloc keeps the one it captured.

This is pre-existing on `main`, and I wouldn't raise it on an unrelated PR. It matters here because this PR changes what that service is used *for*: `_onBackgroundPublishVanished` previously only deleted a draft row, and now **writes owner-attributed draft state**. On an auth change that doesn't re-inflate the subtree — session expiry and sign-out, the two cases the PR body correctly says don't go through a dialog — the stale service parks the video under the *previous* account. That's the same wrong-identity failure mode the PR is fixing, arriving through the mechanism the PR adds to fix it.

The account-switch path itself is fine, because the events are dispatched to the pre-swap bloc holding the pre-swap service. It's the non-dialog auth transitions that are exposed. Per the rule, `ref.watch` both deps in `build` and put a record `ValueKey` on the `BlocProvider`.

## Description doesn't match the diff — `retryUpload`

> `publishVideo` and `retryUpload` now refuse a draft owned by another account with a new `PublishErrorKind.accountChanged`.

`VideoPublishService.retryUpload` isn't guarded — it's **deleted** (`video_publish_service.dart`, -33 lines). I checked `origin/main`: it had no production callers, only its own `group('retryUpload')` tests at `video_publish_service_test.dart:1271-1378`. So removing it is correct and genuinely good — it eliminates a second, unguarded entry point into the publish path, which is more valuable than guarding it would have been. Just worth saying that in the body rather than describing a guard on a method that no longer exists, since the description becomes the merge record for a wrong-identity fix.

## Description doesn't match the diff — the autosave "Out of Scope" note

> the owner scoping makes the collision less destructive (previously the second account's save deleted the first account's clip files as "orphans" — now they are merely left behind)

I don't think that's what happens. On `main`, `getDraftById` had no owner filter at all, and `saveDraft`'s orphan probe used it — unscoped. After this PR that probe is `_loadDraftAcrossAccounts`, which is **also** unscoped, deliberately, with a comment explaining that it must be:

> Read across accounts: the upsert below is keyed on `id` alone and destroys whatever row is there, so an owner-scoped read here would skip the cleanup for the very files it just orphaned.

That reasoning is correct, but it means the path is behaviourally identical to `main`: account B's autosave still finds account A's row and still deletes A's clip files. The "Out of Scope" paragraph is the stated justification for deferring a data-loss issue, so I'd fix the claim (or the deferral) before it lands — as written it tells the next reader the residual risk is benign file leakage when it's still cross-account file deletion.

## Verified — no action needed

Spent time trying to break these and couldn't:

- `_ownedOrLegacy(column, null)` returns `const Constant(true)` (`drafts_dao.dart:41`), so the unscoped read in `isDraftOwnedByAnotherAccount` genuinely sees foreign rows. Worth pinning explicitly: had it returned `ownerPubkey IS NULL` instead, the guard would compile, pass its tests against legacy rows, and silently never fire in the real case. The entire fix hangs on that one line.
- `isDraftOwnedByAnotherAccount` returns `true` when the service's own `ownerPubkey` is null, so signed-out is covered — consistent with keeping the guard independent of the sheet.
- Re-reading `_inFlightDraftIds(publishBloc)` at commit time instead of reusing the ids the warning was built from correctly leaves alone an upload that finished while the picker was open, and you tested that.
- `filter by owner` on `getDraftById` while keeping the unscoped form available for ownership probing is the right API shape, and the three bookkeeping callers (`saveDraft` diff, `deleteDraft`, `draftExists`) are each individually justified in comments rather than left to inference.
- l10n is properly done: all three keys mirrored across 18 locales rather than parked as English fallback, generated files committed, and `publish_error_kind_l10n.dart` updated so the exhaustive switch still holds.

## Nit

`_handleSwitchAccount` uses `else if`, so a user with both in-flight uploads and unsaved drafts only sees the upload warning. The body calls this out as intentional and it's a reasonable call — just noting the drafts count is silently dropped from a destructive-action confirmation.

mobile/lib/screens/settings/settings_screen.dart
hm21: Right, this races — fixed in 7704162.

Parking now goes through a new awaitable `BackgroundPublishBloc.parkInFlight()`: it awaits every write first and only then dispatches the `BackgroundPublishVanished` events, which are left doing nothing but the in-memory cleanup. Re-parking is idempotent (the copy is already gone, or the same draft status is rewritten), so the redundant handler pass is harmless whether or not it survives the teardown. The call site `await`s it before `swapAccount`.

The park rule itself moved into a shared `_park()` used by both the handler and `parkInFlight`, so there's still one place that decides delete-copy vs. keep-as-draft.

Pinned both ends: `does not return before the park write has landed` on the bloc, and `waits for the park to land before swapping the account` on the widget — the latter gates `parkInFlight()` on a `Completer` and asserts `swapAccount` hasn't touched `deviceScope` yet. I reverted the fix locally to check both go red, and they do.

mobile/lib/services/video_publish/video_publish_service.dart
hm21: That test is already in — `drafts_dao_test.dart`, `getDraftById without ownerPubkey reads across accounts`: inserts a row owned by `pubkeyB`, reads it with no `ownerPubkey`, asserts it comes back *and* that `ownerPubkey` is `pubkeyB`. Flipping `_ownedOrLegacy(column, null)` to `column.isNull()` fails it.

The scoped side is pinned right above it (`getDraftById returns only owned + legacy drafts`), and `draft_storage_service_test.dart` covers the same thing one layer up through `isDraftOwnedByAnotherAccount`.

hm21:
@rabble thanks — the teardown race was real and is fixed in 7704162 (replied on the thread). The other four below.

## Not taking: the stale `DraftStorageService`

I don't think the failure mode exists, so I'd rather not paper over it with a key that has its own cost.

`_onBackgroundPublishVanished` touches the captured service through exactly three methods, and none of them are owner-sensitive:

- `DraftsDao.updatePublishStatus` writes `publishStatus` / `publishError` / `publishAttempts` / `lastModified` in an `update(drafts)..where((t) => t.id.equals(id))`. It never writes `ownerPubkey`.
- `DraftsDao.deleteDraft` deletes by primary key, and the service's pre-read is deliberately `_loadDraftAcrossAccounts`.
- `draftExists` is deliberately unscoped.

So a stale service can't re-attribute the row — the draft stays with the account that recorded it regardless of which service instance does the write. That's also why the `_persistPublishStatus` call isn't new here: it was already this handler's fallback on `main`, and this PR only changed the branch condition above it, not what the write touches.

On the prescribed fix specifically: a record `ValueKey` on `BlocProvider<BackgroundPublishBloc>` closes the bloc and rebuilds it on every auth flip, and this bloc's state *is* the in-flight upload registry. On session expiry / sign-out — the two transitions your comment is about — that would drop the queue with nothing left to park it. So the key would introduce the loss it's meant to prevent.

I did add the missing pin, since the argument above is load-bearing: `updatePublishStatus leaves the row with its owner` in `draft_storage_service_test.dart` writes through account B's service to account A's row and asserts A still reads it and B still doesn't. That plus the existing `deleteDraft removes a row whoever owns it` and `draftExists sees a row owned by another account` covers all three calls. If someone later reaches for an owner-scoped read in the park path, one of them goes red.

Happy to be talked round if you see a fourth call I've missed.

## Description — `retryUpload`

Fixed, and you're right that it should say so plainly. The body now describes it as deleted, with the reason (no production callers on `main`, so it was a second unguarded entry point and removing it beats guarding it).

## Description — the autosave "Out of Scope" note

Also fixed; the old claim was wrong. It now says the collision is unchanged and still destructive, and that the orphan probe in `saveDraft` *has* to stay unscoped, since the upsert under it is keyed on `id` alone. What the owner scoping changes is that the colliding row is no longer readable across accounts, not that the collision stops deleting files. Want me to file the follow-up issue and link it from that paragraph?

## Nit — `else if`

Leaving it. The drafts sheet says drafts survive the switch and you may want to review them first; the upload sheet says the upload stops and the video stays as a draft in this account. When both apply the second is the one carrying the destructive news, and stacking two confirmations on one action reads worse than dropping a count from the softer of the two.

mobile/lib/blocs/background_publish/background_publish_bloc.dart
```diff
@@ -183,16 +183,60 @@ class BackgroundPublishBloc
...
+  /// the row stays with the account that recorded it no matter which account
+  /// the service handed here belongs to.
+  Future<void> _park({
+    required String draftId,
+    required DivineVideoDraft? draft,
+  }) async {
+    final sourceDraftId = draft?.sourceDraftId;
+    if (sourceDraftId != null &&
+        sourceDraftId != draft!.id &&
+        await _draftStorageService.draftExists(sourceDraftId)) {
```
realmeylisdev, line 250: This probe can't distinguish "this upload's source survived" from "some autosave row exists", and when it gets that wrong it deletes the row the doc comment above calls "all that is left".

For a fresh recording `sourceDraftId` is the fixed `draft_autosave` id: `getActiveDraft()` falls back to `draftId` → `_draftId ?? VideoEditorConstants.autoSaveId` (`video_editor_provider.dart:103`), and the publish copy takes `sourceDraftId: draft.sourceDraftId ?? draft.id` (`video_publish_provider.dart:476`). That slot is shared and constantly recreated — `clearAll()` → `reset()` reaps it after the handoff (`video_editor_provider.dart:275`), and the next editor session autosaves it straight back (`:739`). `draftExists` is also deliberately unscoped, so another *account's* autosave row satisfies it too — pinned by your own `draftExists sees a row owned by another account`.

Repro: record on A → publish → open the camera and start recording again (autosave row is back) → switch accounts while upload #1 is still running. `draftExists('draft_autosave')` is true, `_deleteDraft(copy.id)` runs, and video #1 is gone — after a sheet that promised it stays as a draft. Same on the `save_drafts` branch of `upload_failure_sheet.dart:62`, where the snackbar then offers a "View" link to a draft that isn't there.

I ran this against the real bloc to be sure: seeding one in-flight upload with `sourceDraftId = autoSaveId` and `draftExists → true`, `parkInFlight()` calls `deleteDraft('publish_1234')` and never writes `PublishStatus.draft`.

Not a regression — `main` deleted the copy unconditionally, so this is strictly better. But the check is the whole safety of the new branch, and for the case it names it's a coin flip. Cheapest fix that errs toward keeping the video is to refuse the shared slot as a fallback, e.g. `sourceDraftId != VideoEditorConstants.autoSaveId` in the condition; worst case you get a duplicate draft instead of a deleted one.

mobile/lib/screens/settings/settings_screen.dart
```diff
@@ -162,6 +189,16 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
...
               Navigator.of(context).pop();
               if (account.pubkeyHex == accountState.currentPubkey) return;
+
+              // Park now that a switch is actually committed, and await it:
+              // `swapAccount` disposes the container this bloc lives in, so a
+              // fire-and-forget event would race the teardown and lose the
+              // video. Parking reads the queue now rather than reusing the ids
+              // the warning was built from, so an upload that finished while
+              // the picker was open is left alone.
+              await publishBloc.parkInFlight();
```
realmeylisdev: The `_AddAccountTile` right below this in the same sheet also ends the current session mid-upload — `addNewAccount()` is just `_authService.signOut()` (`settings_account_cubit.dart:57-60`) — but gets neither the warning nor the park.

No data loss: `signOut()` defaults to `deleteLocalUserData: false`, so the rows survive. But the draft is left at `PublishStatus.publishing`, which `drafts_library_bloc.dart:59` filters out, so the video disappears from the queue *and* from the library until `resumePendingPublishes` sweeps on a later launch. From the user's side that's the same "where did my video go" as the bug this PR fixes, reached from the tile next to the one that now handles it properly.

Awaiting `parkInFlight()` before `addNewAccount()` (and reusing the same warning) would close it; if you'd rather scope it out, worth saying so in the body since the two tiles live in one sheet.

realmeylisdev (commented):
Guard-before-`saveDraft(publishing)` and the awaited `parkInFlight()` both hold up — I re-derived the teardown path and the ownership-neutrality argument and they check out. One thing on the new `_park`: the "is the source still there?" probe is unsound for the fresh-recording case it was written for, and can delete the only copy of the video. Details inline, plus a smaller gap on the sibling "Add another account" tile.

mobile/lib/blocs/background_publish/background_publish_bloc.dart
hm21: Confirmed, and the repro is exact — fixed in 28d9ad3043.

I traced the chain to be sure the fallback id really is the shared constant: `getActiveDraft()` with no args takes `this.draftId` → `_draftId ?? autoSaveId` (`video_editor_provider.dart:103`, `:556`), and the publish copy inherits it via `sourceDraftId: draft.sourceDraftId ?? draft.id` (`video_publish_provider.dart:476`). `clearAll` reaps that slot ~600ms after the handoff (`:541` → `reset` → `removeAutosavedDraft`) and the next session writes it straight back, so `draftExists` could not tell "this upload's source survived" from "some later session's autosave exists" — plus the unscoped read makes another account's row satisfy it too.

Took your cheapest fix: `sourceDraftId != VideoEditorConstants.autoSaveId` in the condition. A duplicate draft is the worst case now instead of a deleted video.

The existing `parks the publish copy when the draft it was copied from is gone` test needed retargeting — it used `draft_autosave` with `draftExists → false`, so after the fix it would have stayed green without exercising the `draftExists` branch at all. It runs on `draft_source` now, and the autosave case has its own test with `draftExists → true`. Reverted the guard locally to check: it fails with `deleteDraft(1)`, the data loss itself.

One adjacent thing I left alone: `_deletePublishedDrafts` (`:310-313`) deletes `sourceDraftId` unconditionally after a successful publish, so on `draft_autosave` it can wipe a *newer* editor session's recovery point — the same hazard the comment in `saveAsDraft` warns about. That is unchanged on `main` and outside this PR, so it wants its own issue rather than a drive-by here.

mobile/lib/screens/settings/settings_screen.dart
hm21: Right, and it is the same failure reached from the tile next door — closed it in 28d9ad3043 rather than scoping it out.

Verified the path before patching: `addNewAccount` is just `signOut()` (`settings_account_cubit.dart:57-60`) with `deleteLocalUserData` defaulting to `false`, so the rows do survive, but the copy stays at `publishing`, which `drafts_library_bloc.dart:59` filters out — missing from the queue and the library until `resumePendingPublishes` sweeps.

`await publishBloc.parkInFlight()` now runs before `addNewAccount()`. No second warning: the in-flight sheet already gates the picker that contains both tiles, so it covers this path as-is, exactly as you suggested.

One limit worth naming on the new `adding an account parks the in-flight uploads` test: it asserts `parkInFlight()` is called, not that `signOut` waits for it. `addNewAccount` returns early behind `FeatureFlag.accountSwitching`, and `FF_ACCOUNT_SWITCHING` defaults to false under test, so `signOut` is never reached and there is nothing to order against without mocking `featureFlagServiceProvider`. The await-before-teardown contract is already pinned on the switch path (`waits for the park to land before swapping the account`) and both tiles call the same method. Dropping the `await` here does turn the new test red, so the call itself is guarded.

dcadenas (approved):
I fixed the takeover findings by making account changes wait until the upload is durably parked, aborting with feedback if parking fails or finds no draft to update, and preserving a reused autosave slot during publish cleanup. This prevents an in-flight video from being lost or published under the wrong account.

Focused coverage, app checks, and current-head CI passed.

The remaining uncertainty is device-level concurrency during a live account switch, which is not exercised end to end. Please review the changes and merge when ready.

PR #4918, opened 2026-06-05
fix(startup): prevent entire app rebuild twice on startup

hm21:
Closes #4919

## Problem

In short, the app always builds twice on startup. This was very noticeable because the entire feed reloaded, and with the camera quick action we could clearly see that the camera was fully reinitialized.

---

`AppShell` was being mounted twice on every cold start. The root cause was in `main.dart`:

```dart
BlocProvider(
  key: ValueKey(identityHashCode(ref.watch(notificationRepositoryProvider))),
  create: (_) => NotificationBadgeCubit(repository: ...),
),
```

`notificationRepositoryProvider` returns `null` during early auth, then flips to a real `NotificationRepository` instance once the profile is available. The `ValueKey` keyed on `identityHashCode` of that provider caused Flutter to **discard and recreate** the entire `BlocProvider` subtree — which sits **above** `MaterialApp.router` and therefore above `AppShell`. This triggered a full remount of the router and shell on every startup (and on every account switch), which was the source of the long-standing "app initializes twice" issue.

## Fix

1. **`NotificationBadgeCubit`** — added a `setRepository(NotificationRepository?)` method that swaps only the `watchUnreadCount` stream subscription without closing the cubit.

2. **`main.dart`** — removed the `ValueKey` from the global `BlocProvider`. Added `_NotificationBadgeRepositorySync`, a small `ConsumerWidget` that `watch`es `notificationRepositoryProvider` and calls `setRepository` on the already-mounted cubit whenever the repository identity changes.

3. **`notification_badge_cubit_auth_transition_test.dart`** — updated to pin the new contract: repository flips must **not** remount descendants (verified via a `_MountProbe` counter), while the badge count still updates correctly.

## Effect

- `AppShell.initState` is now called exactly once per app session.
- Badge count continues to update correctly on sign-in, sign-out, and account switch.
- No change to any other screen, feed, or router logic.

## Tested

- `notification_badge_cubit_test.dart` — 7/7 ✅
- `notification_badge_cubit_auth_transition_test.dart` — 2/2 ✅
- Static analysis: no errors

PR #6482, opened 2026-07-29
fix(people-lists): keep the app-shell bloc on a live repository

### LINKED ISSUE #6480: fix(people-lists): PeopleListsBloc keeps a stale repository across account switches
## Problem

The global `PeopleListsBloc` is constructed once in `mobile/lib/main.dart` with

```dart
repository: ref.read(peopleListsRepositoryProvider),
```

`peopleListsRepositoryProvider` is `@Riverpod(keepAlive: true)`, but it is **not identity-stable**: it does `ref.watch(nostrServiceProvider)` (`mobile/lib/providers/repository_providers.dart:450`), and `NostrService` reassigns its state whenever the auth state changes (`mobile/lib/providers/nostr_client_provider.dart:506` and `:644`) — by its own doc it "recreate[s] the client when the keyContainer changes (e.g., user signs out and signs in with different keys)".

Every recreation yields a fresh `PeopleListsRepositoryImpl`. The bloc holds `_repository` as a `final` field, so it can never pick up the newer instance. `ownerPubkeyStream` still fires `PeopleListsOwnerChanged` for the new owner and routes it through the repository whose `NostrClient` is dead.

## Why now

Pre-existing, but #6476 removed the conditional `if (peopleListsEnabled)` around the `BlocProvider`. Before, a flag flip re-inflated the subtree and re-ran `create:`, which happened to refresh the capture. With the entry unconditional nothing re-runs `create:` at all, so the staleness is permanent by construction after the first account switch.

## Options

- A record-typed `ValueKey` over the captured deps (the cure `.claude/rules/state_management.md` prescribes) works, but this `BlocProvider` sits above `MaterialApp.router` — re-keying it would re-inflate the whole app subtree on every account switch, which is the exact blast radius #6476 just removed. Probably the wrong trade here.
- Exception #2 in the same rule fits better: an app-shell-level bloc that must outlive the change should subscribe to it. Thread the repository in as a stream, or have the bloc re-resolve it alongside `ownerPubkeyStream`.

## Acceptance

- Switching accounts leaves `PeopleListsBloc` operating on a live `NostrClient`.
- Regression test: emit a new owner after the repository identity changed, assert the new repository is used.
- No re-inflation of the app subtree on account switch.

Found during review of #6476 by @realmeylisdev.

realmeylisdev:
## Description

`PeopleListsBloc` is built once in the app-shell `MultiBlocProvider` and captured `peopleListsRepositoryProvider` with `ref.read`. That provider is `keepAlive` but **not identity-stable**: it watches `nostrServiceProvider`, whose client is **disposed** at `nostr_client_provider.dart:622` before the replacement is built. `create:` never re-runs, so the capture never recovers.

**The trigger is not the Settings account switch.** `swapAccount()` builds a new `ProviderContainer` and `ContainerSwapHost` re-inflates the subtree through a generation-keyed `KeyedSubtree` (`container_swap_host.dart:99-127`), so that path self-heals. What actually bites is the in-container route — sign-out → welcome → sign-in (`settings_account_cubit.dart:59`, `welcome_bloc.dart:227`/`:327`) — plus the cold-start pre-auth flip that #4918 describes. `FF_ACCOUNT_SWITCHING` defaults to **false** (`build_configuration.dart:25`), which hides the self-healing path entirely on a default build.

**Impact, measured against the local stack rather than inferred.** A disposed client returns `PublishFailed`, and `queryEvents` returns nothing:

```
live-client publish                          -> PublishSuccess
disposed-client publish                      -> PublishFailed
disposed-client queryEvents                  -> 0 events
```

So `syncOwner` hits `if (events.isEmpty) return;` and the new owner never pulls its lists from relays, while every mutation fails and rolls back. **Curated lists go silently read-only until app restart.**

No wrong-account publish is possible, and that is structural, not luck: `Nostr.signEvent` copies back only `id` and `sig`, never `pubkey` (`nostr.dart:252-258`), so the event declares the *new* owner while carrying the *old* key's signature. Against the local funnelcake relay:

```
pubkey A, signed by A → accepted: true
pubkey B, signed by A → accepted: false  "invalid: invalid signature: signature verification failed"
pubkey B, signed by B → accepted: true
```

### Why this shape

Neither cheaper cure is available in this position.

- The **`ValueKey`-over-captured-deps** form was shipped and **reverted** in #4918: keying a `BlocProvider` that sits above `MaterialApp.router` "caused Flutter to discard and recreate the entire `BlocProvider` subtree … a full remount of the router and shell on every startup (and on every account switch)".
- The **`setRepository` + sync `ConsumerWidget`** form (`AppShellBadgeScope`) needs `lazy: false` to dodge the #6115 re-entrant lazy create — which would defeat the curated-lists gating that depends on `PeopleListsBloc` staying lazy.

So the bloc subscribes to the change instead (`state_management.md` rule exception #2), through a reusable `identityStreamOf` helper rather than a bespoke `StreamController` at the `create:` site.

Two implementation notes a reviewer will want:

- `identityStreamOf` uses **`ref.container.listen`, not `ref.listen`**. A listener registered inside a provider body is only driven while that provider itself has listeners, and this one is `ref.read` from `create:` — never watched. With `ref.listen` the stream emits only the seed even after the source has demonstrably rebuilt. There is a test that fails if it is changed back.
- It returns `Stream.multi`, not an `async*` generator, so a second consumer cannot hit "Stream has already been listened to", and each subscriber is seeded with the value current at *its* subscribe time (the bloc subscribes from an event handler, not the constructor).

`_repository` becomes a mutable field. That is a deliberate, narrow deviation from "no mutable instance variables in BLoC" — the rule targets *state*, and `NotificationBadgeCubit` already carries the same mutable-dependency shape for the same reason.

### One writer for the lists subscription

Re-pointing the repository is only safe if exactly one handler owns `_listsSubscription`. `sequential()` orders events **within** a bucket, so `PeopleListsRepositoryChanged` and `PeopleListsOwnerChanged` ran concurrently: `_onOwnerChanged` read the field before its `await …cancel()` and nulled it after, so a swap landing in that window orphaned the subscription it had just opened. With both events dispatched in the same turn the bloc watched the previous owner *and* the new one on the incoming repository, the previous-owner subscription outlived `close()`, and the incoming account's client sent `syncOwner(previousOwner)` to the relay.

`_onRepositoryChanged` now only swaps the field and re-dispatches `PeopleListsOwnerChanged.rewire()`. That gives `_listsSubscription` a single writer, and queuing the re-wire behind any pending owner change means a swap arriving mid-switch resolves against the owner that is *arriving* rather than the one that is leaving. `_onOwnerChanged` drops its `await` and stays synchronous — `cancel()` stops delivery synchronously and its future only reports resource cleanup, so nothing below needed it, and staying synchronous is what makes the single writer uninterruptible.

### The wiring is guarded by the type system, not by a test

`main.dart:2841` is the one production line this fix exists for, and it was unguarded. `main_app_shell_repository_sync_test.dart` builds a local `_AppShellProbe` that *mirrors* the wiring rather than executing it, so deleting `repositoryStream:` from `main.dart` left that suite green.

`repositoryStream` is therefore **required**, not optional: dropping it now fails analysis with `missing_required_argument` at `lib/main.dart:2841`. A bloc that outlives its dependency's identity has to be told when that identity changes, so requiring the stream states the contract rather than leaving it opt-in. Callers that genuinely cannot see the repository change pass `const Stream.empty()`.

**Related Issue:** Closes #6480

## Out of Scope

- **The better long-term fix is to make `PeopleListsRepository` client-agnostic** (`NostrClient Function()` instead of an instance), which would make the provider identity-stable and fix every consumer rather than this one bloc. That is a wider refactor than this regression warrants, so it belongs in its own issue.
- **`BackgroundPublishBloc` was evaluated and dropped.** It touches draft storage in exactly two places, `deleteDraft(id)` and `updatePublishStatus(draftId:)`, and both go straight to `DraftsDao` keyed by `id` with no owner filter. `ownerPubkey` on `DraftStorageService` only affects draft *creation* and *listing*, which this bloc does neither of — so swapping the instance changed nothing observable, and the tests that came with it could only fail on a reverted mechanism. Removing the wiring also puts `draftStorageServiceProvider` back to `isAutoDispose: true`.
- The corrected trigger analysis is [commented on #6480](https://github.com/divinevideo/divine-mobile/issues/6480#issuecomment-5117667883); the issue body still describes the account-switch path.

## Verification

- No generated drift — pre-commit codegen verification clean, CI `Generated Files` green
- `flutter test test/features/people_lists/ test/main_app_shell_repository_sync_test.dart test/providers/ test/blocs/background_publish_bloc/` — 702 passing
- `flutter test test/providers/ test/blocs/` — 3250 passing
- `flutter test test/screens/user_list_people_screen_test.dart test/screens/other_profile_screen_test.dart test/widgets/app_shell_badge_scope_test.dart test/blocs/notifications/badge/` — 32 passing (adjacent consumers and the sibling badge wiring)

Load-bearing claims were checked against a neutered fix rather than assumed:

- Deleting `repositoryStream:` from `main.dart` now fails analysis — `missing_required_argument` at `lib/main.dart:2841`. This is what closes the gap described above; before the parameter was required, the same deletion left the suite green.
- Reverting `container.listen` → `ref.listen` in `identityStreamOf` turns `provider_identity_stream_test.dart`'s "emits the new instance when the source rebuilds" red, while the other four tests stay green.

The runtime claims (disposed-client publish/query behaviour, relay signature rejection) were checked against `local_stack` (funnelcake relay on `:47777`), not reasoned from source.

## Type of Change


hm21 (approved):
Pushed two fixes as separate commits, then approved. Revert either if you disagree — the merge call stays yours.

**`1dfb300` — dropped the `BackgroundPublishBloc` half.** The bloc touched `_draftStorageService` in exactly two places, `deleteDraft(id)` and `updatePublishStatus(draftId:)`, and both go straight to `DraftsDao` keyed by `id` with no owner filter. `ownerPubkey` on `DraftStorageService` only affects draft *creation* and *listing*, and this bloc does neither — so swapping the instance changed nothing observable, and the three tests that came with it could only fail on a reverted mechanism, never on a behaviour regression. Removing the wiring also puts `draftStorageServiceProvider` back to `isAutoDispose: true`; the container listener had quietly promoted it to permanently alive and eagerly rebuilt.

**`6caad48` — gave `_listsSubscription` a single writer.** `sequential()` orders events only within a bucket, so `PeopleListsRepositoryChanged` and `PeopleListsOwnerChanged` ran concurrently. `_onOwnerChanged` read the field before its `await …cancel()` and nulled it after, so a swap landing in that window orphaned the subscription it had just opened. With both events dispatched in the same turn the bloc ended up watching the previous owner *and* the new one on the incoming repository, the previous-owner subscription outlived `close()`, and the incoming account's client sent `syncOwner(previousOwner)` to the relay. `_onRepositoryChanged` now only swaps the field and re-dispatches `PeopleListsOwnerChanged.rewire()`, which also queues behind any pending owner change so a mid-switch swap resolves against the owner that is arriving. `_onOwnerChanged` drops its `await` and stays synchronous — that is what makes the single writer uninterruptible. The new regression test fails against your previous handlers.

Validation: `flutter analyze lib test` clean; 702 tests green across `test/features/people_lists/`, `test/main_app_shell_repository_sync_test.dart`, `test/providers/`, `test/blocs/background_publish_bloc/`. CI still running at time of approval. Not device-tested.

The rest holds up — the `container.listen`-vs-`ref.listen` argument and the `Stream.multi` seeding in `identityStreamOf` are both correct as written.

One thing left for you: the PR body still describes the draft-storage half, and the "Out of Scope" note about #6478 no longer applies.

realmeylisdev:
@hm21 Verified both commits — the two draft-storage calls really are id-keyed with no owner filter, and the single-writer rewire holds. Title and body updated; the draft half and the #6478 note are gone.

One commit on top (`447d3267f`): made `repositoryStream` required. `main.dart:2841` had no guard — `main_app_shell_repository_sync_test` mirrors that `create:` in a local probe rather than executing it, so deleting the argument left the suite green at `+2`. It's a compile error now. Flagging it since it changes a signature after your approval; all checks green.

PR #6483, opened 2026-07-29
feat(video-editor): add a per-clip chroma key (green screen)

### LINKED ISSUE #6484: feat(video-editor): add a per-clip chroma key (green screen)
Auto-created from PR #6483.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6483
**Branch:** `feat/video-editor-chroma-key`

hm21:
Closes #6484

## Description

Selecting a clip in the video editor now offers a green-screen option that opens a full-screen editor: auto-detect the screen colour or pick it by hand, tune amount / edge / spill, and replace the background with a colour, an image, a clip from the library, or nothing.

The feature needed two enabling changes that are worth reviewing on their own terms, so they are their own commits:

**`chore(android): render with Impeller instead of Skia`** — the most consequential commit here, and the one to look at first. We opted out of Impeller after profiling found Impeller/Vulkan janky during feed scroll and Impeller/GLES rendering video textures black. Be clear about what has changed since, because it is not what you would assume: nothing upstream has. That measurement was taken three weeks ago in `df2e479f7`, on this same Flutter 3.44.0 pin, with both cited GLES issues already closed — so this is not a re-evaluation after an engine upgrade. What changed is the requirement. `ui.ImageFilter.shader` is Impeller-only, so the live preview cannot run on Skia at all, and we are knowingly trading a measured rendering risk for a working preview. There is no `EnableImpeller` entry rather than one set to `true` because Impeller is the engine default on API 29+, so the entry would be redundant; pinning it to `true` works equally well, since the engine only logs the opt-out deprecation warning when Impeller is actually off. Debug builds keep the opt-out; that one works around an emulator crash under SwiftShader, not a product decision. **This changes how the whole app rasterises, so it deserves a second opinion beyond my device — see Risks below.**

**`feat(video-player): let a companion player opt out of exclusive playback`** — `play()` enforced "only one player runs at a time" for every controller. That is right for anything the user perceives as a video of its own, and wrong for a player that is a layer of another one: the backdrop stopped the clip it sits behind. `exclusivePlayback` (default `true`, so nothing else changes) takes a controller out of the rule in both directions. Honouring only one direction would have inverted the bug rather than fixed it — whichever player started last would stop the other.

**`feat(video-editor): add a per-clip chroma key (green screen)`** — the feature itself.

The last four commits are a review pass over the branch; each names the finding it fixes.

### Why it applies on Done instead of staying clip state

The obvious design is to keep the key on the clip and resolve it at export. That fell apart once a video background was involved: a second track forces the render down the composition path, which rules out playback speed, reverse, and transitions for that clip. It also would not compose with the clip-overlay work coming later. Baking on Done produces a plain keyed file that behaves like any other clip. The settings are stored alongside it, so reopening the screen restores them, and Remove puts the original file back without spending a render.

### Preview

The preview keys each frame through a fragment shader that ports the renderer's formula, so the editor shows what the file will contain rather than an approximation. Where `ImageFilter.shader` is unavailable the preview falls back to the unkeyed video and says so; the export is unaffected either way.

A video backdrop needs two players, since one cannot drive two surfaces. They cannot share a clock, so the backdrop is re-anchored whenever the clip in front loops. It is anchored to what the export produces rather than to zero: the bake tiles the backdrop across the *untrimmed* clip from source time zero and trim is applied downstream, so the clip's first visible frame sits `trimStart` into the backdrop, and both run at the clip's playback speed. Anchoring the bake to the trim instead would look tidier and then desync the moment the user re-trims. The re-align threshold exists because position updates are not monotonic; without it, normal jitter would re-seek the backdrop constantly.

Backgrounds are baked upright before rendering. The renderer ignores EXIF orientation, which otherwise rotated portrait photos by 90 degrees.

### Known limitation

Transparent is offered but flattens to black, because H.264 carries no alpha. It is still a distinct choice from a black fill — the user is saying "no backdrop" — and the picker says what will happen when it is selected. It also cannot go down the single-track path at all: the renderer asserts rather than silently flattening, so it becomes a lone keyed layer over the composition's own canvas.

**Related Issue:** 

## Risks

The Impeller switch is app-wide, not scoped to the editor. I verified the chroma-key paths and the full suite, but I did not re-run the original feed-scroll profiling that motivated the opt-out, and an S25 is the only device I have. If feed scroll regresses or video textures render black on other hardware, that commit is the one to revert; it is deliberately kept separate so it can be dropped without touching the feature.

One thing to know before approving, because it is the opposite of the intuitive read: "falls back to Skia" only holds below API 29. On API 29+ without a usable Vulkan context — Huawei, MediaTek below API 31, the engine's known-bad-SOC list — the engine falls back to Impeller **OpenGLES**, which is exactly the backend the black-video-texture measurement implicates. `isShaderFilterSupported` reports true there, so the editor's no-shader fallback does not engage on those devices either. That device class is what this switch actually exposes, and it is not covered by my testing.

## Out of Scope

- The renderer ignoring EXIF orientation is an upstream `pro_video_editor` bug. This PR works around it by baking orientation into the background bitmap before handing it over; it affects image layers elsewhere too and is being reported upstream.
- Clip overlays. The bake-on-Done design was chosen partly to leave room for them, but nothing here implements them.

## Verification

- `flutter test` — 13749 passing
- `flutter test` in `packages/divine_video_player` — 275 passing, `divine_video_player_controller.dart` fully covered
- `flutter test` in `packages/db_client` — 65 passing
- ARB consistency test — all 18 locales translated, no English fallbacks
- Ratchets: untested-services floor, test-unit structure, and all four design-system ceilings — all green
- Both `AndroidManifest.xml` files parse
- Manually tested on a real Samsung Galaxy S25 (profile build): auto-detect and manual colour, all four background types, live preview, reopening a keyed clip and getting its settings back, Remove restoring the original, and the backdrop staying aligned with the clip across loops
- New behaviour is test-backed, and the tests that guard a fixed bug are revert-confirmed rather than assumed. The two exclusive-playback tests fail with `['pause', 'play']` instead of `['play']` when the fix is removed; the timeline busy-state test and the file-reference DAO test each go red with their bug restored
- One test was deleted rather than repaired: `ChromaKeyedVideo`'s "survives a key change without leaking the old shader" could not fail, since `flutter_test` reports no shader image filter and that branch is unreachable. `ChromaKeyBackdrop`'s background dispatch is covered instead

## Type of Change


NotThatKindOfDrLiz:
hm21 I took over the review fixes and pushed f311970ba.

What changed:
- Chroma-key source videos and generated image backdrops are now deletion candidates in draft/library cleanup and draft-save orphan cleanup, while still going through the existing reference checks.
- Reverse, transform, remove, and re-key paths now queue cleared chroma-key assets into the editor deferred cleanup path.
- Removed the unused shader test reset hook, made needsComposition match the actual bake routing, and routed bake composition through that getter.
- Missing backdrop videos now use a typed non-reportable exception; picker paths also check mounted/file existence.
- The preview-unavailable notice now covers shader asset load failure, not just unsupported backends.

Verification:
- flutter test test/services/file_cleanup_service_test.dart test/services/draft_storage_service_test.dart test/models/video_editor/clip_chroma_key_test.dart test/services/video_editor/chroma_key_bake_service_test.dart test/blocs/video_editor/clip_editor/clip_editor_bloc_test.dart
- flutter analyze lib test
- Pre-push hook: merge-conflict check, analyze, generated-file check, and changed-file tests
- GitHub checks are green on f311970ba

Please review the takeover commit when you have a chance and merge when comfortable with the remaining Impeller hardware risk described in the PR body.

NotThatKindOfDrLiz (approved):
Approved after takeover fixes in f311970ba. Cleanup leaks, dead-code/routing issues, expected missing-backdrop handling, picker lifecycle guards, and preview-fallback notice are addressed. CI is green. Remaining Impeller hardware coverage is the explicit maintainer/product risk already documented in the PR body.

NotThatKindOfDrLiz:
@hm21 tagging you explicitly for the takeover handoff: f311970ba is pushed, CI is green, and I approved. Please review the takeover commit and merge when comfortable with the documented Impeller hardware risk.

Issue #3324, opened 2026-04-24
fix: device attestation stuck indefinitely on jailbroken devices


hm21:
### Summary

Device attestation gets stuck indefinitely on jailbroken devices.

### Environment

iOS

### Steps to Reproduce

1. Use the app on a jailbroken iPhone.
2. Trigger a flow that requires device attestation.
3. Observe the attestation process.

### Actual Result

The attestation process never completes and remains stuck indefinitely.

### Expected Result

The attestation process should complete but just return `null`.

### Regression?

Unknown

### Evidence

This was reported on Discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1492713745204641954).

hm21:
@redaranj I assume this is probably already addressed in the PR you mentioned you’re working on, but I created a quick sub-issue for visibility and linked it to #3214. 

Let me know if you want the logs from this user. Since they were shared privately, I didn’t include them here. What’s interesting is that the first attempt to attest seems to work and returns null, but the second attempt gets stuck with no further logs, and nothing shows up in the catch block either.

PR #6487, opened 2026-07-29
fix(attestation): run iOS App Attest off the main thread

### LINKED ISSUE #3214: fix: app freezes briefly after recording due to attestation check
### Summary

The app freezes for ~200ms after recording a clip, likely due to a main-thread attestation call.

### Environment

main-branch
iPhone 12

### Steps to Reproduce

1. Open the camera on iOS.
2. Record a clip.
3. Stop recording and observe responsiveness.

### Actual Result

The app freezes briefly (~200ms) after each recording.

### Expected Result

Recording should complete smoothly without any noticeable freeze.

### Regression?

No

### Evidence

Freeze appears linked to `getAttestationServiceSupport` in `native_proofmode_service.dart` running on the main thread (from app_device_integrity package).
Comment out this part [here](https://github.com/divinevideo/divine-mobile/blob/cc62e79e57b20f25f0b816abfe586582b73be202/mobile/lib/services/native_proofmode_service.dart#L186-L188) will not freeze the app anymore.

hm21:
## Description

The app froze for ~200ms after every recording on iOS. Recording a clip kicks off proof generation (`ClipManagerNotifier._generateClipProof`), which calls `getAttestationServiceSupport` on the vendored `app_device_integrity` plugin. Flutter delivers method channel calls on the main thread, and the plugin did its App Attest work right there — `DCAppAttestService.isSupported` in the initializer, plus Secure Enclave key generation — so the UI stalled while it ran.

The fix moves the whole handler onto a background queue and hops back to main only to deliver the `FlutterResult`. That includes the initializer-failure path (`-4`), which previously returned on main as well. The plugin is vendored under `mobile/overrides/app_device_integrity-1.1.0` (see the `dependency_overrides` entry in `mobile/pubspec.yaml`), so no upstream change is needed.

Device logs show what was actually sitting on that thread — the attestation round-trip takes roughly two seconds per clip:

```
15:56:01.099  🔐 Native proof hash: 1c5f4798…
15:56:03.198  🔐 Getting native ProofMode proof directory…   → ~2.1s
15:56:05.445  🔐 Native proof hash: fab3006c…
15:56:07.302  🔐 Getting native ProofMode proof directory…   → ~1.86s
```

Both clips ended at `verified_mobile` with a ~7.9 KB device attestation written, so the proof chain is unchanged — it just no longer blocks the UI.

**Related Issue:** Closes #3214

## Out of Scope

- **#3324** (attestation hangs indefinitely on jailbroken devices) sits at the same call site and needs a timeout on the Dart side. Kept separate to keep this diff to the threading fix.
- **Pre-existing robustness bug, not filed yet:** every attestation error arrives in Dart as a `PlatformException` and lands in the broad `catch` at `native_proofmode_service.dart:270`, so `proofFile` returns `null` and the clip loses its *entire* ProofMode/C2PA proof — not just the attestation. This triggers when the device is offline, when Apple rate-limits attestation, and on any device without App Attest support (including the Simulator, where `isSupported` is false).
- **No automated test.** The vendored plugin has no iOS test target, only an Android/Kotlin one, and the behaviour is a threading property of a native platform channel. Verified on device instead.

## Verification

- `swiftc -typecheck` against `Flutter.xcframework/ios-arm64` (Flutter 3.44.0) and the iphoneos SDK, target `arm64-apple-ios14.0`, together with `AppDeviceIntegrity.swift` — clean.
- Full Xcode device build via `flutter run`: `Xcode build done. 115.7s`.
- Manually verified on a physical **iPad (iOS 26.5.2)**: recorded multiple clips, no freeze on stopping the recording. Attestation still succeeds (`Attestation successful`, `verified_mobile`, device attestation file written). Note this is an iOS-only path — it cannot be exercised on Android or in the Simulator.
- No Dart files changed, so no Flutter tests are affected; the pre-push hook confirmed the same.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrityPlugin.swift
+++ b/mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrityPlugin.swift
@@ -19,34 +19,42 @@ public class AppDeviceIntegrityPlugin: NSObject, FlutterPlugin {
                 return
             }
 
-            guard let deviceIntegrity = AppDeviceIntegrity(challengeString: challengeString) else {
-                result(FlutterError(code: "-4", message: "Failed to initialize AppDeviceIntegrity", details: nil))
-                return
-            }
+            // App Attest hits the Secure Enclave and Apple's attestation
+            // servers. Flutter delivers channel calls on the main thread, so
+            // doing that work here froze the UI for ~200ms after every
+            // recording. Stay off-main and only hop back to deliver the result.
+            DispatchQueue.global(qos: .userInitiated).async {
+                guard let deviceIntegrity = AppDeviceIntegrity(challengeString: challengeString) else {
+                    DispatchQueue.main.async {
+                        result(FlutterError(code: "-4", message: "Failed to initialize AppDeviceIntegrity", details: nil))
+                    }
+                    return
+                }
 
-            // Directly generate key and attest
-            deviceIntegrity.generateKeyAndAttest { success in
-                DispatchQueue.main.async {
-                    if success {
-                        // Attestation successful, return the attestationString
-                        let attestationResult = [
-                            "attestationString": deviceIntegrity.attestationString ?? "",
-                            "keyID": deviceIntegrity.keyIdentifier()
-                        ]
+                // Directly generate key and attest
+                deviceIntegrity.generateKeyAndAttest { success in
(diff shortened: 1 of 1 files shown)


dcadenas (approved):
✅ App Attest work stays off the main thread

PR #6488, opened 2026-07-29
fix(profile): close vanished account leaks

NotThatKindOfDrLiz:
## Description

Follow-up to #6464, which merged while this takeover was in progress.

This closes the pieces from the review that still make sense on top of `main`:

- treats `has_vanish_request` as historical erasure-request state, not a tombstone by itself
- returns `UserProfileVanished` only for `has_vanish_request: true` with `profile: null`
- treats `has_vanish_request: true` with a populated profile as `UserProfileFound`, so a republished account self-heals
- ignores the obsolete `deleted` field
- stops the profile banner and empty DM state from rendering stale cached avatar/banner/NIP-05 for vanished accounts
- shares one vanished-pubkey watch stream across profile rows instead of opening one Drift query stream per row
- clears `_rawKind0ConfirmedMissing` when a vanish self-heals
- marks `AppDbClient.upsertProfile` as testing-only so it does not look like a safe production write path

## Product Decision

@rabble please decide the remaining bulk-path edge:

`POST /api/users/bulk` returning `has_vanish_request: true` with `profile: null` is currently treated as `UserProfileVanished`, matching the single-profile endpoint and avoiding split rendering between profile screen and inbox. If completed-vanish-with-post-boundary-activity-but-no-kind-0 should instead remain live-unknown/not-published, this is the behavior to flip before merge.

## Verification

- `flutter test test/src/funnelcake_api_client_test.dart` from `mobile/packages/funnelcake_api_client` — 302 passed
- `flutter test test/src/profile_repository_test.dart` from `mobile/packages/profile_repository` — 238 passed
- `flutter test test/src/database/daos/vanished_profiles_dao_test.dart test/src/app_db_client_test.dart` from `mobile/packages/db_client` — 50 passed
- `flutter test test/widgets/profile/profile_banner_layer_test.dart test/screens/inbox/conversation/conversation_view_test.dart` from `mobile` — 51 passed
- `flutter analyze lib test` from `mobile/packages/funnelcake_api_client` — clean
- `flutter analyze lib test` from `mobile/packages/profile_repository` — clean
- `flutter analyze lib test` from `mobile/packages/db_client` — clean
- `flutter analyze lib test/widgets/profile/profile_banner_layer_test.dart test/screens/inbox/conversation/conversation_view_test.dart` from `mobile` — clean
- pre-push hook — analyzer clean, generated files clean, changed-file test suite 616 passed

## Type of Change

### DIFF
--- a/mobile/lib/providers/user_profile_providers.dart
+++ b/mobile/lib/providers/user_profile_providers.dart
@@ -12,6 +12,14 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'user_profile_providers.g.dart';
 
+final _vanishedProfilePubkeysProvider = StreamProvider<Set<String>>((ref) {
+  return ref
+      .watch(databaseProvider)
+      .vanishedProfilesDao
+      .watchAllPubkeys()
+      .map((pubkeys) => pubkeys.toSet());
+});
+
 // ignore: specify_nonobvious_property_types
 final userProfileStatsReactiveProvider =
     StreamProvider.family<ProfileStats?, String>((ref, pubkey) {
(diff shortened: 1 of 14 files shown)


NotThatKindOfDrLiz:
@rabble this is the product call that remains from your #6464 review:

For `POST /api/users/bulk`, this follow-up currently treats `has_vanish_request: true` with `profile: null` as `UserProfileVanished`, matching the single-profile endpoint and avoiding split rendering between the profile screen and inbox.

Please confirm whether that bulk edge should tombstone, or whether completed-vanish-with-post-boundary-activity-but-no-kind-0 should instead remain live-unknown/not-published. If product wants the latter, I can flip that path before merge.

dcadenas (approved):
🔒 vanished accounts now close their leaks cleanly

PR #6489, opened 2026-07-29
chore(infinite-video-feed): delete the loop seek nothing calls

### LINKED ISSUE #6445: chore(infinite-video-feed): delete dead seek-based loop enforcement that can silently reintroduce the loop seam
## Summary

`packages/infinite_video_feed` still carries the seek-based loop-enforcement machinery that PR #5551 disabled. It has **zero production callers** on `main` — it is kept alive only by its own unit tests — but it remains reachable through a public widget parameter. Any future caller that passes `maxLoopDuration` silently reintroduces the audible loop seam that #5551 removed.

This is the follow-up flagged while triaging #6442 (closed as duplicate of #6386).

## Why this matters

PR #5551 ("preserve native video loop boundaries", merged 2026-06-26) fixed #5544 by removing the 6.3s `seekTo(Duration.zero)` loop enforcement from feed playback. The fix was made at the **call site**, not by removing the mechanism:

`lib/widgets/video_feed_item/feed_videos.dart:367`
```dart
// Do not pass maxLoopDuration here. Feed playback should loop at the
// asset boundary; restarting with a Dart seek at the 6.3s recording
// limit creates an audible seam.
```

A comment is the only thing standing between us and a regression of a user-visible bug that has now been reported three separate times through Zendesk (#5544, #6386, #6442). Seamless looping is a core product promise — it is the thing users cite as what made Vine feel like Vine — so the guard should be structural, not advisory.

## Current state on `main`

`git grep maxLoopDuration -- '*.dart'` outside tests returns only:

- **The one comment** at `feed_videos.dart:367` saying not to use it.
- **The definition itself**, entirely inside `packages/infinite_video_feed`.

No production code passes it.

Two consumers inside the package, with different dispositions:

**1. `subscribeToLoopEnforcement` — fully dead, and it is the footgun.**
`lib/src/services/controller_subscriptions.dart:44-68`. Takes a non-nullable `Duration maxLoopDuration` and calls `onSeekToZero()` — this *is* the seam-producing seek. Its only invocation is `lib/src/widgets/infinite_video_feed.dart:1712-1724`, guarded by `if (maxLoopDuration != null)`, which is never true in production.

**2. `subscribeToAutoAdvance` — live and needed, but the parameter is vestigial.**
`lib/src/services/controller_subscriptions.dart:74-135`. Auto-advance detection is genuinely used. It takes `Duration? maxLoopDuration` purely to compute `effectiveEnd = min(maxLoopDuration, duration)`, and with the parameter always null that reduces to `effectiveEnd = duration`. The nullable plumbing produces four dead branches across the initial-state and stream-listener paths.

## Proposed change

Package layer — `mobile/packages/infinite_video_feed`:

1. Delete `subscribeToLoopEnforcement` and its `_loop` subscription map (plus the corresponding cancel/dispose handling).
2. Delete the `maxLoopDuration` field and constructor parameter from `InfiniteVideoFeed` (`lib/src/widgets/infinite_video_feed.dart:56`, `:238`) and the `if (maxLoopDuration != null)` block at `:1712-1724`.
3. Simplify `subscribeToAutoAdvance` to drop the `maxLoopDuration` parameter and derive `effectiveEnd` from the natural duration alone, collapsing the four dead branches.
4. Update the doc comment at `:252` that still references `maxLoopDuration` as a loop-completion trigger.

App layer:

5. Remove the now-unnecessary explanatory comment at `feed_videos.dart:367` — once the parameter does not exist, there is nothing to warn against.

Tests:

6. `packages/infinite_video_feed/test/src/services/controller_subscriptions_test.dart` — drop the `subscribeToLoopEnforcement` coverage; adjust the auto-advance tests for the simplified signature.
7. `test/widgets/video_feed_item/feed_looping_contract_test.dart` — this currently asserts that `feed_videos.dart` does not pass `maxLoopDuration`. After the deletion that becomes structurally impossible (it would not compile), so the source-string assertion should be retired rather than left asserting a property the type system now enforces. Keep whatever part of the file guards native looping (`setLooping(looping: true)`) if it covers that separately.

## Scope notes

- Behavior-neutral on `main` by construction: every deleted branch is provably unreachable today, since no caller supplies the parameter.
- `packages/infinite_video_feed` is not currently a strict-coverage package, but tests should still be updated alongside rather than deferred.
- No l10n, golden, or migration impact.

## Related

- #5544 — original "Loops are not seamless" report
- #5551 — the call-site fix this issue makes structural
- #6386 — open "Perfect Loop" investigation (assigned, media probe pending)
- #6442 — closed as duplicate of #6386; this follow-up was flagged during that triage

Note that #6386 may conclude the residual seam is asset-side (AAC priming/padding on archived Vines) rather than player-side. That does not change the case for this cleanup — the dead seek path is a regression risk regardless of where the remaining seam comes from.

hm21:
## Description

PR #5551 fixed the audible loop seam (#5544) by stopping the feed from passing `maxLoopDuration` — restarting a video with a Dart `seekTo(Duration.zero)` at the 6.3s recording limit both truncated classic Vines and left a seam at the join. The fix was made at the call site; the mechanism itself stayed behind, reachable through a public widget parameter and kept alive only by its own unit tests. A code comment was the only thing standing between us and a regression of a bug users have now reported three separate times (#5544, #6386, #6442). Seamless looping is the thing people cite as what made Vine feel like Vine, so the guard should be structural rather than advisory.

This deletes the machinery:

- `subscribeToLoopEnforcement` and its `_loop` subscription map, plus the matching cancel/dispose handling. This was the seam-producing seek.
- `InfiniteVideoFeed.maxLoopDuration` — field, constructor parameter, and the `if (maxLoopDuration != null)` wiring block — along with the now-orphaned `_loopSeekInProgress` guard set.
- The `maxLoopDuration` plumbing in `subscribeToAutoAdvance`. Auto-advance detection itself is live and needed; the parameter only fed `effectiveEnd = min(maxLoopDuration, duration)`, and with it always null that reduces to `effectiveEnd = duration`, collapsing four dead branches across the initial-state and stream-listener paths.

Behaviour-neutral on `main` by construction — every deleted branch was already unreachable, because no production code supplied the parameter.

**The 7s playback cap is untouched.** It is a separate mechanism: `maxPlaybackDuration` (`AppConstants.maxFeedPlaybackDuration`), applied as a native clip end (`VideoClip.end`) so the loop point stays inside the platform player. Both tests that pin it — the per-`VideoClip` cap-parity check and the app-side wiring assertion — are left in place.

Two deviations from the issue's proposed change, both deliberate:

- The issue asks to delete the explanatory comment at `feed_videos.dart:367`. Half of it is live rationale for why `maxPlaybackDuration` exists and why the cap sits above the recording limit, so only the "deliberately NOT `maxLoopDuration`" warning is gone.
- The issue asks to retire the source-string assertion in `feed_looping_contract_test.dart`. The `maxLoopDuration` half is retired — the type system enforces it now — but it is replaced with `isNot(contains('seekTo(Duration.zero)'))`. That is a different property, not type-enforced, and it is the tripwire the issue actually wants: someone hand-rolling a Dart loop-restart trips it. It scans all three layers that hold a controller — the pooled player, `controller_subscriptions.dart` (where the deleted seek's stream listener lived), and the app-side call site — because scanning only the file the *wiring* sat in would have let the same seek back in one file over. Verified it can fail from both directions: the repo's `dart_code_only.awk` filter over `origin/main`'s copy of the pooled player hits the string once, and injecting the seek into `subscribeToAutoAdvance` turns the test red on the service path.

One test is added rather than deleted. Collapsing `effectiveEnd` rewrote `subscribeToAutoAdvance`'s arming condition, and that condition had no negative case — replacing it with a bare `armed = true;` left the suite green. Since `onVideoLoopCompleted` drives auto-advance, an over-eager arm would skip the viewer to the next video on any backward jump, including the stale-playback recovery seek. The new case pins it: a jump back from mid-clip must not fire.

Also corrected a stale doc line in `fullscreen_feed_bloc.dart` that still credited "loop enforcement" to the FeedVideos player configuration.

**Related Issue:** Closes #6445

The tripwire's comment also names what it does *not* catch, from review: it matches the literal spelling, and `_seekKick` already seeks with a computed duration, so widening to a `seekTo(` regex would collide with stale-playback recovery. `seekTo(const Duration())` reaches zero unseen. It is a tripwire for the regression that shipped three times, not a proof.

## Out of Scope

- `mobile/docs/plans/2025-12-20-video-6s-loop-limit.md` still describes the old seek-based approach. It is a historical planning document, left as a record.
- `_LoadingIndicator` in `pooled_fullscreen_video_feed_screen.dart` carries the last non-doc reference to "loop-enforcement seeks", but the widget is never instantiated — the fix is deleting it, not trimming its comment, and that file is untouched here. Tracked as #6492.
- #6386 ("Perfect Loop") closed as completed on 2026-07-28 via #6430, which trims local clips to the point where both tracks still have content. That landed in the native player and is independent of this cleanup; nothing here touches it.

## Verification

- `flutter test --coverage` in `mobile/packages/infinite_video_feed` — 228/228 pass, 726/726 lines (100%, the package's effective floor)
- `flutter test test/widgets/video_feed_item/ test/blocs/fullscreen_feed/` — 394/394 pass
- `git grep maxLoopDuration` returns nothing across the repo
- Mutation-checked both new guards: a bare `armed = true;` fails only the new arming case, and a `seekTo(Duration.zero)` injected into `subscribeToAutoAdvance` fails the widened contract test
- Pre-push hook green
- Manual check on a real Samsung Galaxy: **not run**, deliberately. The sole `InfiniteVideoFeed` construction site never supplied `maxLoopDuration`, so every deleted branch was already unreachable in a shipped build, and the surviving conditions are algebraically identical to the ones they replace. There is no runtime path here for a device to exercise that CI does not. Recording it rather than implying otherwise.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
+++ b/mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
@@ -80,7 +80,8 @@ const _maxConcurrentCacheDownloads = 1;
 ///
 /// **Playback hooks integration:**
 /// - Background caching triggered via [FullscreenFeedVideoCacheStarted]
-/// - Loop enforcement handled by the FeedVideos player configuration
+/// - Looping and the playback-length cap handled by the FeedVideos player
+///   configuration
 /// - Cache resolution happens at the player/repository level
 class FullscreenFeedBloc
     extends Bloc<FullscreenFeedEvent, FullscreenFeedState> {
--- a/mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart
+++ b/mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart
@@ -21,6 +21,13 @@ void main() {
     // guard covers all three rather than the one the deleted wiring sat in:
     // the subscription service owns the position stream the old seek fired
     // from, and the app-side call site configures both.
+    //
+    // It matches the spelling, not the value. `_seekKick` already seeks with
+    // a computed duration for stale recovery, so widening to a `seekTo(`
+    // regex would collide with it — the cost is that `seekTo(const
+    // Duration())` reaches zero unseen. Read this as a tripwire for the
+    // regression that shipped three times, not as proof no Dart seek can
+    // reach zero.
     const subscriptionsPath =
         'packages/infinite_video_feed/lib/src/services/'
         'controller_subscriptions.dart';
--- a/mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart
+++ b/mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart
@@ -5,22 +5,21 @@ import 'package:openvine/constants/app_constants.dart';
 import 'package:openvine/constants/video_editor_constants.dart';
 
 void main() {
-  test('feed playback uses native looping instead of 6.3s seek enforcement', () {
-    final feedVideosSource = _dartCodeOnly(
-      'lib/widgets/video_feed_item/feed_videos.dart',
-    );
+  test('feed playback loops natively rather than seeking back to zero', () {
+    // The seek-based enforcement this replaced no longer exists to be passed
+    // (#6445), so what is left to pin is that the loop still happens in the
+    // platform player.
     final pooledPlayerSource = _dartCodeOnly(
       'packages/infinite_video_feed/lib/src/widgets/infinite_video_feed.dart',
     );
 
     expect(pooledPlayerSource, contains('setLooping(looping: true)'));
-    expect(feedVideosSource, isNot(contains('maxLoopDuration:')));
     expect(
-      feedVideosSource,
-      isNot(contains('maxLoopDuration: VideoEditorConstants.maxDuration')),
+      pooledPlayerSource,
+      isNot(contains('seekTo(Duration.zero)')),
       reason:
-          'Feed playback must not restart long videos with a Dart seek at the '
-          '6.3s recording limit; seek-based restarts create audible seams.',
+          'Feed playback must not restart a video with a Dart seek; '
+          'seek-based restarts create audible seams (#5544).',
     );
   });
 
--- a/mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart
+++ b/mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart
@@ -9,18 +9,36 @@ void main() {
     // The seek-based enforcement this replaced no longer exists to be passed
     // (#6445), so what is left to pin is that the loop still happens in the
     // platform player.
-    final pooledPlayerSource = _dartCodeOnly(
-      'packages/infinite_video_feed/lib/src/widgets/infinite_video_feed.dart',
+    const pooledPlayerPath =
+        'packages/infinite_video_feed/lib/src/widgets/infinite_video_feed.dart';
+
+    expect(
+      _dartCodeOnly(pooledPlayerPath),
+      contains('setLooping(looping: true)'),
     );
 
-    expect(pooledPlayerSource, contains('setLooping(looping: true)'));
-    expect(
-      pooledPlayerSource,
-      isNot(contains('seekTo(Duration.zero)')),
-      reason:
-          'Feed playback must not restart a video with a Dart seek; '
-          'seek-based restarts create audible seams (#5544).',
-    );
+    // Every layer that holds a controller can hand-roll the restart, so the
+    // guard covers all three rather than the one the deleted wiring sat in:
+    // the subscription service owns the position stream the old seek fired
+    // from, and the app-side call site configures both.
+    const subscriptionsPath =
+        'packages/infinite_video_feed/lib/src/services/'
+        'controller_subscriptions.dart';
+    const seekFreePaths = [
+      pooledPlayerPath,
+      subscriptionsPath,
+      'lib/widgets/video_feed_item/feed_videos.dart',
+    ];
+
+    for (final path in seekFreePaths) {
+      expect(
+        _dartCodeOnly(path),
+        isNot(contains('seekTo(Duration.zero)')),
+        reason:
+            '$path must not restart a video with a Dart seek; '
+            'seek-based restarts create audible seams (#5544).',
+      );
+    }
   });
 
   test('every feed source is opened with the cap applied', () {
--- a/mobile/lib/widgets/video_feed_item/feed_videos.dart
+++ b/mobile/lib/widgets/video_feed_item/feed_videos.dart
@@ -366,9 +366,8 @@ class FeedVideosState extends ConsumerState<FeedVideos> with RouteAware {
         },
         // Nothing in the feed plays longer than a Vine, not even a 60s file a
         // foreign Nostr client points at. The cap is a native clip end, so the
-        // loop point stays in the platform player — deliberately NOT
-        // maxLoopDuration, whose Dart seek at the 6.3s recording limit both
-        // truncated classic Vines and created an audible seam (#5544, #6421).
+        // loop point stays in the platform player and the join stays seamless
+        // (#5544, #6421).
         maxPlaybackDuration: AppConstants.maxFeedPlaybackDuration,
         onVideoLoopCompleted: _handleAutoAdvanceCompleted,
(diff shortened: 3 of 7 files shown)


mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
```diff
@@ -80,7 +80,8 @@ const _maxConcurrentCacheDownloads = 1;
 ///
 /// **Playback hooks integration:**
 /// - Background caching triggered via [FullscreenFeedVideoCacheStarted]
-/// - Loop enforcement handled by the FeedVideos player configuration
+/// - Looping and the playback-length cap handled by the FeedVideos player
```
realmeylisdev, line 83: Good catch on this one. There's a sibling still live: `lib/screens/feed/pooled_fullscreen_video_feed_screen.dart:968` justifies the 100ms `_LoadingIndicator` delay with "flashes that occur during play/pause and loop-enforcement seeks". That screen renders `FeedVideos`, so it's the same path — and after this PR it's the last reference to the mechanism outside the historical plan doc. One-word trim while you're here?

mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart
```diff
@@ -5,23 +5,40 @@ import 'package:openvine/constants/app_constants.dart';
...
+    const seekFreePaths = [
+      pooledPlayerPath,
+      subscriptionsPath,
+      'lib/widgets/video_feed_item/feed_videos.dart',
+    ];
+
+    for (final path in seekFreePaths) {
+      expect(
+        _dartCodeOnly(path),
+        isNot(contains('seekTo(Duration.zero)')),
```
realmeylisdev, line 43: Worth knowing what this does and doesn't catch: it's matched on the literal spelling, and the pooled player already seeks with a computed duration — `_seekKick` at `infinite_video_feed.dart:1617` does `controller.seekTo(Duration(milliseconds: positionMs))` for stale-playback recovery. A hand-rolled restart written that way, or as `seekTo(const Duration())`, walks straight past.

Not asking you to widen it — a regex over `seekTo(` would collide with `_seekKick` and the reason string would stop being true. But the PR description frames this as structural where the type system now carries that half; this part is a tripwire for the exact regression, which is still worth having. Maybe just say so in the test's comment so the next reader doesn't over-trust it.

realmeylisdev (approved):
Verified the deletion is behaviour-neutral: with `maxLoopDuration` always null, the old `effectiveEnd` reduced to `duration` on both the initial-state and listener paths, so the collapsed conditions are equivalent. Re-ran both mutations locally — a bare `armed = true;` in either arming path fails only the new mid-clip case, and a `seekTo(Duration.zero)` injected into `controller_subscriptions.dart` turns the widened contract test red. Package CI and all shards green.

Two non-blocking notes inline.

mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
hm21: Checked it — the comment is there and that screen does render `FeedVideos`, but `_LoadingIndicator` is never instantiated. The only four references are its own class declaration, constructor, `createState`, and the State class; it is file-private with no `part` directive, so nothing outside can reach it. Trimming the comment would polish a widget that never builds.

Leaving it out of this PR: that file is untouched here and a ~40-line removal in an unrelated screen is scope creep on a chore. Tracked as #6492 so it does not get lost.

mobile/test/widgets/video_feed_item/feed_looping_contract_test.dart
hm21: Fair, and both escapes check out — `_seekKick` is the only surviving `seekTo` in the scanned files and it takes a computed duration, so a `seekTo(` regex would collide with it exactly as you say, and `seekTo(const Duration())` is the same value spelled past the literal.

Documented rather than widened, in d6dacd2aaa: the comment now says it matches the spelling and not the value, names `const Duration()` as the gap, and asks the reader to treat it as a tripwire for the regression that shipped three times rather than as proof no Dart seek reaches zero.

PR #6490, opened 2026-07-29
fix(proofmode): scope App Attest to the publishing account

### LINKED ISSUE #6491: fix(proofmode): reuse one App Attest key instead of minting one per proof
Auto-created from PR #6490.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6490
**Branch:** `fix/ios-app-attest-key-reuse`

hm21:
Closes #6491

## Description

This PR scopes iOS App Attest caching to the publishing Nostr account and mints the payload at publish time, where the signing pubkey is known. The App Attest challenge is now `SHA-256(UTF-8("<proofHash>:<pubkeyHex>"))`, so the payload is bound to both the media proof hash and the event pubkey.

The publish-time payload is authoritative: iOS replaces any stored proof-generation attestation, clears stale values when no signing pubkey is available, and drops the field when App Attest fails or times out. Android remains unchanged and continues to publish generation-time hardware attestation from the stored proof.

The native iOS cache now scopes key id, attestation object, pending-key slot, and provisioning queue by account. Provisioning still has the 30s native watchdog, and a generated-but-unattested pending key is retried once with its original client data hash before being abandoned so a stale pending slot cannot wedge an account permanently.

`mobile/docs/NOSTR_VIDEO_EVENTS.md` documents the current verifier contract, assertion-counter requirement, per-account linkability boundary, and the legacy pre-PR iOS challenge shape for already-published events.

## Still Required Before Merge

- Physical iPhone verification. App Attest is unavailable on Simulator, so first-publish provisioning and later-publish assertion paths still need a real device run.

## Verification

- `flutter test test/services/ios_device_attestation_service_test.dart test/services/video_event_publisher_native_proof_test.dart` — 11 passed.
- Pre-push hook: analyzer passed and changed-file tests passed for `packages/models/test/src/pending_upload_proofmode_test.dart`, `test/services/ios_device_attestation_service_test.dart`, `test/services/native_proofmode_service_test.dart`, `test/services/video_event_publisher_native_proof_test.dart`, and `test/services/video_event_publisher_test.dart` — 67 passed.
- `swiftc -parse -target arm64-apple-ios14.0 overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift` — parses, with the host sysroot warning from invoking `swiftc` directly on macOS.

## Type of Change

- [x] Bug fix
- [x] Documentation

### DIFF
--- a/mobile/docs/NOSTR_VIDEO_EVENTS.md
+++ b/mobile/docs/NOSTR_VIDEO_EVENTS.md
@@ -138,6 +138,20 @@ from cache on every later proof. A verifier needs two branches:
   attestation nonce binds the current proof hash and the plain nonce check
   applies.
 
+Both branches check against the client data hash the app signs, which is not the
+proof hash itself but `SHA-256(UTF-8(proofHash))` — one more hash over the
+hex-encoded digest, where `proofHash` is the same string the `proofmode` tag
+carries as `videoHash`.
+
+**Replay boundary.** Nothing in this payload binds the event pubkey, and the
+challenge is derivable by anyone holding the video. The three fields can be
+lifted from one event and republished verbatim under a different pubkey carrying
+the same media, and both branches above still pass. A verifier following this
+contract learns that a genuine Apple device attested *this media hash* — not
+that *this publisher's* device did, and not that the publisher holds the key.
+Binding the payload to the publisher means mixing the event pubkey into the
+challenge, which changes what the client signs and is not part of this shape.
+
 Because the key lives for the life of the install, `keyID` and
 `attestationString` are a stable public identifier: every video published from
 one install carries the same pair, including videos published from different
--- a/mobile/docs/NOSTR_VIDEO_EVENTS.md
+++ b/mobile/docs/NOSTR_VIDEO_EVENTS.md
@@ -147,6 +147,13 @@ challenge from the event alone. Because the pubkey is inside it, the payload
 cannot be lifted from one event and republished under a different pubkey
 carrying the same media: the challenge no longer matches.
 
+Legacy iOS payloads published before PR #6490 used
+`SHA-256(UTF-8("<proofHash>"))` and minted a fresh key during proof generation.
+Those payloads have no `assertionString` and are not bound to the event pubkey.
+Verifiers that need to evaluate historical events must branch on publication
+date or another release cutover signal instead of applying the publish-time
+challenge above to every existing iOS event.
+
 **Two branches.** Apple rate limits `generateKey` and `attestKey`, so the Secure
 Enclave key is provisioned on an account's first publish and `keyID` plus
 `attestationString` are replayed from cache afterwards:
--- a/mobile/docs/NOSTR_VIDEO_EVENTS.md
+++ b/mobile/docs/NOSTR_VIDEO_EVENTS.md
@@ -127,35 +127,51 @@ On iOS the tag value is the App Attest payload verbatim:
 }
 ```
 
-Apple rate limits `generateKey` and `attestKey`, so the Secure Enclave key is
-provisioned once per install and `keyID` plus `attestationString` are replayed
-from cache on every later proof. A verifier needs two branches:
+The payload is minted when the event is signed, not when the video is recorded.
+The account a clip goes out under is not fixed until then — a clip can be
+recorded, sit in the library, and be published from an account the user switched
+to afterwards — so only the publish step knows which identity the attestation
+should speak for.
 
-- **`assertionString` present** — verify the assertion against *this* proof's
-  hash. The attestation still proves the key came from a genuine Apple device,
-  but its embedded nonce binds the proof that provisioned the key, not this one.
-- **`assertionString` absent** — this is the provisioning proof, so the
-  attestation nonce binds the current proof hash and the plain nonce check
+**What the client signs.** The challenge is the proof hash and the event pubkey
+joined by a colon, and Apple's client data hash is `SHA-256` over its UTF-8
+bytes:
+
+```
+clientDataHash = SHA-256(UTF-8("<proofHash>:<pubkeyHex>"))
+```
+
+`proofHash` is the same string the `proofmode` tag carries as `videoHash`, and
+`pubkeyHex` is the event's own `pubkey` — so a verifier reconstructs the
+challenge from the event alone. Because the pubkey is inside it, the payload
+cannot be lifted from one event and republished under a different pubkey
+carrying the same media: the challenge no longer matches.
+
+**Two branches.** Apple rate limits `generateKey` and `attestKey`, so the Secure
+Enclave key is provisioned on an account's first publish and `keyID` plus
+`attestationString` are replayed from cache afterwards:
+
+- **`assertionString` present** — verify the assertion against *this* event's
+  client data hash. The attestation still proves the key came from a genuine
+  Apple device, but its embedded nonce binds the event that provisioned the key,
+  not this one.
+- **`assertionString` absent** — this is the provisioning event, so the
+  attestation nonce binds the current client data hash and the plain nonce check
   applies.
 
-Both branches check against the client data hash the app signs, which is not the
-proof hash itself but `SHA-256(UTF-8(proofHash))` — one more hash over the
-hex-encoded digest, where `proofHash` is the same string the `proofmode` tag
-carries as `videoHash`.
+**Assertion counter.** Apple's assertion verification requires checking the
+counter in the assertion's `authenticatorData`: it increments per assertion of
+that key and must be strictly greater than the highest value already seen for
+that `keyID`. A verifier that skips this accepts replayed assertions.
 
-**Replay boundary.** Nothing in this payload binds the event pubkey, and the
-challenge is derivable by anyone holding the video. The three fields can be
-lifted from one event and republished verbatim under a different pubkey carrying
-the same media, and both branches above still pass. A verifier following this
-contract learns that a genuine Apple device attested *this media hash* — not
-that *this publisher's* device did, and not that the publisher holds the key.
-Binding the payload to the publisher means mixing the event pubkey into the
-challenge, which changes what the client signs and is not part of this shape.
-
-Because the key lives for the life of the install, `keyID` and
-`attestationString` are a stable public identifier: every video published from
-one install carries the same pair, including videos published from different
-Nostr accounts on that device. The cache is wiped on uninstall and on a
+**Linkability.** The key is scoped to the Nostr account, not the install: every
+video published from one account carries the same `keyID` and
+`attestationString`, and the counter above is a monotonic sequence over that
+account's publishes. That correlates only videos the account already signed with
+its own pubkey, so it exposes nothing the event does not already carry. It also
+follows Apple's guidance against sharing one App Attest key among several users
+of a device — switching accounts provisions a separate key rather than reusing
+one across identities. Caches are wiped on uninstall and on a
 `DCError.invalidKey` recovery (device restore, OS reset). Nothing consumes these
 fields yet.
 
--- a/mobile/docs/NOSTR_VIDEO_EVENTS.md
+++ b/mobile/docs/NOSTR_VIDEO_EVENTS.md
@@ -115,6 +115,36 @@ parsing is enforced in `mobile/packages/models/lib/src/video_event.dart` and
 | `pgp_fingerprint` | `["pgp_fingerprint", "ABCD1234EFGH5678"]` | PGP public key fingerprint for signature verification |
 | `c2pa_manifest_id` | `["c2pa_manifest_id", "<manifest-id>"]` | Active C2PA manifest identifier when available |
 
+#### `device_attestation` payload (iOS)
+
+On iOS the tag value is the App Attest payload verbatim:
+
+```json
+{
+  "keyID": "<key identifier>",
+  "attestationString": "<base64 attestation object>",
+  "assertionString": "<base64 assertion>"
+}
+```
+
+Apple rate limits `generateKey` and `attestKey`, so the Secure Enclave key is
+provisioned once per install and `keyID` plus `attestationString` are replayed
+from cache on every later proof. A verifier needs two branches:
+
+- **`assertionString` present** — verify the assertion against *this* proof's
+  hash. The attestation still proves the key came from a genuine Apple device,
+  but its embedded nonce binds the proof that provisioned the key, not this one.
+- **`assertionString` absent** — this is the provisioning proof, so the
+  attestation nonce binds the current proof hash and the plain nonce check
+  applies.
+
+Because the key lives for the life of the install, `keyID` and
+`attestationString` are a stable public identifier: every video published from
+one install carries the same pair, including videos published from different
+Nostr accounts on that device. The cache is wiped on uninstall and on a
+`DCError.invalidKey` recovery (device restore, OS reset). Nothing consumes these
+fields yet.
+
 ### Creator Identity Hints
 
 These tags are discovery hints only. They do not replace the event pubkey as the
--- a/mobile/docs/PROOFMODE_ARCHITECTURE.md
+++ b/mobile/docs/PROOFMODE_ARCHITECTURE.md
@@ -92,7 +92,9 @@ final isValid = await keyService.verifySignature(
 **iOS (App Attest)**:
 - Uses `app_device_integrity` plugin
 - Generates hardware-backed attestation tokens
-- Includes challenge nonce to prevent replay attacks
+- Signs a challenge derived from the media proof hash, which binds the payload
+  to the media but not to the publisher — see the replay boundary in
+  [NOSTR_VIDEO_EVENTS.md](NOSTR_VIDEO_EVENTS.md#device_attestation-payload-ios)
 - Always hardware-backed on iOS 14+
 
 **Android (Play Integrity)**:
--- a/mobile/docs/PROOFMODE_ARCHITECTURE.md
+++ b/mobile/docs/PROOFMODE_ARCHITECTURE.md
@@ -90,10 +90,12 @@ final isValid = await keyService.verifySignature(
 **Platform-specific Implementation**:
 
 **iOS (App Attest)**:
-- Uses `app_device_integrity` plugin
+- Uses `app_device_integrity` plugin via `IosDeviceAttestationService`
 - Generates hardware-backed attestation tokens
-- Signs a challenge derived from the media proof hash, which binds the payload
-  to the media but not to the publisher — see the replay boundary in
+- Runs at publish time, not during proof generation: the account the proof is
+  published under is not fixed until the event is signed
+- Signs a challenge over both the media proof hash and the event pubkey, and
+  uses a key scoped to that account — see
   [NOSTR_VIDEO_EVENTS.md](NOSTR_VIDEO_EVENTS.md#device_attestation-payload-ios)
 - Always hardware-backed on iOS 14+
 
@@ -531,12 +533,15 @@ See `docs/MANUAL_TEST_VIDEO_UPLOAD.md` for manual upload verification procedures
 - Non-blocking (async)
 
 **Device Attestation**:
-- iOS App Attest: one network round trip to Apple on first use (`generateKey` +
-  `attestKey`, both rate limited), then local `generateAssertion` per proof. The
-  key id and attestation object are cached in `UserDefaults` for the lifetime of
-  the install. A key that was generated but not attested — Apple throttled or
-  offline — is recorded too, so the next proof retries `attestKey` on it instead
-  of spending a second rate-limited generation.
+- iOS App Attest: one network round trip to Apple on an account's first publish
+  (`generateKey` + `attestKey`, both rate limited), then local
+  `generateAssertion` per publish. The key id and attestation object are cached
+  in `UserDefaults` per account, so switching accounts costs one more round trip
+  rather than re-using another identity's key. A key that was generated but not
+  attested — Apple throttled or offline — is recorded too, so the next publish
+  retries `attestKey` on it instead of spending a second rate-limited
+  generation. A wedged Apple callback is abandoned after 30s so it cannot stall
+  later publishes.
 - Android Play Integrity: ~200-500ms
 - A failed attestation degrades the proof to no `device_attestation` field; it
   never discards the PGP signature or C2PA manifest.
--- a/mobile/docs/PROOFMODE_ARCHITECTURE.md
+++ b/mobile/docs/PROOFMODE_ARCHITECTURE.md
@@ -532,7 +532,9 @@ See `docs/MANUAL_TEST_VIDEO_UPLOAD.md` for manual upload verification procedures
 - iOS App Attest: one network round trip to Apple on first use (`generateKey` +
   `attestKey`, both rate limited), then local `generateAssertion` per proof. The
   key id and attestation object are cached in `UserDefaults` for the lifetime of
-  the install.
+  the install. A key that was generated but not attested — Apple throttled or
+  offline — is recorded too, so the next proof retries `attestKey` on it instead
+  of spending a second rate-limited generation.
 - Android Play Integrity: ~200-500ms
 - A failed attestation degrades the proof to no `device_attestation` field; it
   never discards the PGP signature or C2PA manifest.
--- a/mobile/docs/PROOFMODE_ARCHITECTURE.md
+++ b/mobile/docs/PROOFMODE_ARCHITECTURE.md
@@ -536,6 +536,8 @@ See `docs/MANUAL_TEST_VIDEO_UPLOAD.md` for manual upload verification procedures
 - Android Play Integrity: ~200-500ms
 - A failed attestation degrades the proof to no `device_attestation` field; it
   never discards the PGP signature or C2PA manifest.
+- The published payload shape and the verifier rule that follows from the cache
+  are documented in [NOSTR_VIDEO_EVENTS.md](NOSTR_VIDEO_EVENTS.md#device_attestation-payload-ios).
 
 ## Future Enhancements
 
--- a/mobile/docs/PROOFMODE_ARCHITECTURE.md
+++ b/mobile/docs/PROOFMODE_ARCHITECTURE.md
@@ -529,10 +529,13 @@ See `docs/MANUAL_TEST_VIDEO_UPLOAD.md` for manual upload verification procedures
 - Non-blocking (async)
 
 **Device Attestation**:
-- iOS App Attest: ~100-200ms
+- iOS App Attest: one network round trip to Apple on first use (`generateKey` +
+  `attestKey`, both rate limited), then local `generateAssertion` per proof. The
+  key id and attestation object are cached in `UserDefaults` for the lifetime of
+  the install.
 - Android Play Integrity: ~200-500ms
-- Done once at session start
-- Cached for session duration
+- A failed attestation degrades the proof to no `device_attestation` field; it
+  never discards the PGP signature or C2PA manifest.
 
 ## Future Enhancements
 
--- a/mobile/lib/services/native_proofmode_service.dart
+++ b/mobile/lib/services/native_proofmode_service.dart
@@ -233,17 +233,7 @@ class NativeProofModeService {
 
       //add iOS specific device attestation
       if (!kIsWeb && defaultTargetPlatform == TargetPlatform.iOS) {
-        final appAttestationPlugin = AppDeviceIntegrity();
-        final tokenReceived = await appAttestationPlugin
-            .getAttestationServiceSupport(challengeString: proofHash);
-        if (tokenReceived != null) {
-          final String? proofDir = await _channel.invokeMethod('getProofDir', {
-            'proofHash': proofHash,
-          });
-
-          final deviceAttestation = File('$proofDir/$proofHash.attest');
-          deviceAttestation.writeAsString(tokenReceived);
-        }
+        await _writeIosDeviceAttestation(proofHash);
       }
 
       // Read proof metadata from native library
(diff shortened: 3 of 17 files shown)


mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrityPlugin.swift
```diff
@@ -37,11 +37,17 @@ public class AppDeviceIntegrityPlugin: NSObject, FlutterPlugin {
                     // several KB of base64.
                     let payload: Any
                     if success {
-                        // Attestation successful, return the attestationString
-                        let attestationResult = [
+                        // The attestation proves the key came from a genuine
+                        // Apple device; the assertion — present on every
+                        // challenge after the key was provisioned — proves this
+                        // specific challenge was signed by that key.
+                        var attestationResult = [
```
NotThatKindOfDrLiz, line 45: This payload is published verbatim in both the `device_attestation` tag and inside the `proofmode` JSON. With the key id and attestation cached per install, `keyID` plus `attestationString` become a stable public identifier that can link every video from this device, including videos from different Nostr accounts on the same phone. That may be acceptable for App Attest, but it needs an explicit product/security decision and documentation before we publish immutable Nostr events with this shape.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift
```diff
@@ -19,60 +55,190 @@ final class AppDeviceIntegrity {
...
-        guard let keyID = self.keyID else {
-            print("No key ID available for attestation")
-            completion(false)
+    private func resolveCredential(
+        forceRefresh: Bool,
+        completion: @escaping (Credential?) -> Void
+    ) {
+        Self.lock.lock()
+
+        if forceRefresh {
```
NotThatKindOfDrLiz: This force-refresh clears the shared cache unconditionally. If two in-flight assertions both started from stale key K1, the first invalid-key recovery can provision and store K2, then the second recovery can clear K2 and provision K3. That still races and burns rate-limited key generations after restore. Please compare against the failed credential's `keyID` and only clear the stored entries if they still refer to that same key.

mobile/docs/PROOFMODE_ARCHITECTURE.md
```diff
@@ -496,10 +496,13 @@ See `docs/MANUAL_TEST_VIDEO_UPLOAD.md` for manual upload verification procedures
 - Non-blocking (async)
 
 **Device Attestation**:
-- iOS App Attest: ~100-200ms
+- iOS App Attest: one network round trip to Apple on first use (`generateKey` +
+  `attestKey`, both rate limited), then local `generateAssertion` per proof. The
+  key id and attestation object are cached in `UserDefaults` for the lifetime of
```
NotThatKindOfDrLiz: This mentions the cache, but the public verifier contract change needs to be documented in `mobile/docs/NOSTR_VIDEO_EVENTS.md` too. After first use, `attestationString` no longer binds the current proof hash; verifiers must use `assertionString` for per-proof binding when present, and only fall back to the attestation nonce for the provisioning proof. Future verifier work is much more likely to read the Nostr event tag docs than this architecture note.

NotThatKindOfDrLiz (changes requested):
Thanks for taking on the App Attest rate-limit problem. The Dart-side degradation/await/null-proof-dir pieces look good, but I don't think this should merge yet. The PR changes the public proof contract in a way that can link all videos from one install, and the invalid-key recovery path still has a concurrency race that can burn the same rate-limited key generation this is trying to avoid. Please get the product/security call on public per-install linkability, fix the compare-and-clear race, update the public Nostr video event docs for the new verifier rule, and verify the assertion path on a physical iPhone before merge.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift
hm21: Fixed in 214ee575bf. `resolveCredential` now takes `replacing staleKeyID:` instead of `forceRefresh:` and clears the cache only while it still refers to the key that failed. Three orderings: cache still holds K1 → cleared, provisions K2 once; a recovery for K1 already in flight → cache is empty, so the second failure joins `pendingProvisioning` instead of starting a second `generateKey`; K2 already stored → mismatch, no clear, the second failure asserts against K2.

mobile/docs/PROOFMODE_ARCHITECTURE.md
hm21: Agreed, done in 214ee575bf. `NOSTR_VIDEO_EVENTS.md` now has a `device_attestation` payload subsection under ProofMode Metadata with the iOS JSON shape and the two-branch verifier rule: assertion present → verify it against this proof's hash, assertion absent (only the provisioning proof) → the attestation nonce check still applies. The architecture note now links there rather than restating it.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrityPlugin.swift
hm21: You are right about the shape: with the key cached, `keyID` + `attestationString` are a stable public identifier for the life of the install, spanning every Nostr account on the device, and it lands in immutable events.

214ee575bf documents that property in `NOSTR_VIDEO_EVENTS.md` alongside the verifier rule, including the two things that reset it (uninstall, `DCError.invalidKey` recovery). I have not changed the design — the call between per-install linkability and Apple's rate limit is a product/security decision, not mine to make in a perf PR, so this stays blocked on that sign-off.

Alternatives if the answer is no: expire the cache on a fixed window so the identifier only links videos inside it (one throttled round trip per rotation instead of one per proof), or keep the key cached but stop publishing `keyID`/`attestationString` in the tag and hold them for a future verifier endpoint — nothing consumes the field today, so that breaks nothing now. Happy to take either.

hm21:
@NotThatKindOfDrLiz addressed in 214ee575bf:

- **Compare-and-clear race** — `resolveCredential` takes `replacing staleKeyID:` and clears the cache only while it still refers to the failed key. A concurrent recovery either joins the in-flight provisioning or picks up the replacement, so one dead key costs one `generateKey`.
- **Verifier contract** — documented in `NOSTR_VIDEO_EVENTS.md` (payload shape + the assertion/attestation two-branch rule); `PROOFMODE_ARCHITECTURE.md` links there.
- **Public linkability** — documented in the same place, design unchanged. This needs the product/security call before merge; two mitigations are sketched in that thread if the answer is no. Leaving your changes-requested standing on that item.

mobile/docs/NOSTR_VIDEO_EVENTS.md
```diff
@@ -115,6 +115,36 @@ parsing is enforced in `mobile/packages/models/lib/src/video_event.dart` and
...
+from cache on every later proof. A verifier needs two branches:
+
+- **`assertionString` present** — verify the assertion against *this* proof's
+  hash. The attestation still proves the key came from a genuine Apple device,
+  but its embedded nonce binds the proof that provisioned the key, not this one.
+- **`assertionString` absent** — this is the provisioning proof, so the
+  attestation nonce binds the current proof hash and the plain nonce check
+  applies.
+
+Because the key lives for the life of the install, `keyID` and
```
NotThatKindOfDrLiz: This documents the linkability, but the PR still needs the product/security decision behind it. Please explain why Divine is choosing to publish a stable per-install App Attest identifier in immutable public Nostr events, including why cross-account device correlation is acceptable and why this cannot be private verifier state or a scoped verifier result instead.

mobile/docs/NOSTR_VIDEO_EVENTS.md
```diff
@@ -115,6 +115,36 @@ parsing is enforced in `mobile/packages/models/lib/src/video_event.dart` and
...
+  "attestationString": "<base64 attestation object>",
+  "assertionString": "<base64 assertion>"
+}
+```
+
+Apple rate limits `generateKey` and `attestKey`, so the Secure Enclave key is
+provisioned once per install and `keyID` plus `attestationString` are replayed
+from cache on every later proof. A verifier needs two branches:
+
+- **`assertionString` present** — verify the assertion against *this* proof's
```
NotThatKindOfDrLiz: Please document the replay boundary here. The challenge is derived from the media proof hash and is public to anyone holding the video/event; the published `{keyID, attestationString, assertionString}` is not bound to the Nostr event pubkey. A verifier following this contract proves that some genuine Apple device signed for this media hash, not that this publisher's device did. Also specify the exact client-data hash shape used by the app.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift
```diff
@@ -19,60 +55,199 @@ final class AppDeviceIntegrity {
...
+            guard let keyIdentifier = keyIdentifier else {
+                print("Failed to generate key identifier")
+                completion(nil)
                 return
             }
 
-            self.attestationString = attestationObject.base64EncodedString()
-            print("Attestation successful")
-            completion(true)
+            self.attestService.attestKey(keyIdentifier, clientDataHash: self.challengeHash) { attestation, error in
```
NotThatKindOfDrLiz: Persist the generated key identifier before calling `attestKey`, and retry attestation with the same key id/client data hash when attestation fails transiently. Apple's App Attest docs call out recording the key id immediately after `generateKey`, and specifically say `serverUnavailable` should retry `attestKey` with the same inputs to avoid unnecessary new keys. As written, `generateKey` success followed by `attestKey` failure discards the key id and the next proof burns another rate-limited key generation.

NotThatKindOfDrLiz (changes requested):
Thanks for tightening the concurrency path. I still do not think this is ready to merge. The key product/security decision is whether Divine is intentionally accepting a public, immutable, per-install identifier in Nostr events. Please spell out that decision and the rationale in the PR before we merge: what threat model are we accepting, what product value requires public raw App Attest material rather than private verification or a scoped verifier result, and why is cross-account device linkability acceptable for Divine users?

If the decision is that public per-install linkability is not acceptable, the mitigation has to remove the stable App Attest material from both publication sites: the standalone `device_attestation` tag and the embedded `proofmode` JSON. Dropping only one still publishes the same identifier through the other.

I also want the verifier docs to state the replay limitation explicitly, and the provisioning path should persist the generated key id before `attestKey` so transient Apple/network failures do not burn another rate-limited key on the next proof.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift
hm21: Done in a7d8096fed. The key identifier and the client data hash it was generated for are written before `attestKey` runs, and cleared only once Apple either attests the key or rejects it with `.invalidKey`. A later proof picks that key up and retries `attestKey` with both original inputs instead of generating a new one.

I did not add an immediate in-round retry. Resuming on the next proof already satisfies "same key id, same client data hash", and it spaces the retry out instead of hitting a server that just reported itself unavailable. If you would rather have both, say so and I will add it.

One consequence worth flagging, because it is not visible in the diff: a resumed key gets attested against a hash that predates the current challenge, so Apple's attestation binds the proof that generated the key rather than the one asking. `isFresh` now tracks whether the stored hash still matches the current challenge, and the first-caller shortcut in `finishProvisioning` is gated on it. Without that gate a resumed round would publish an attestation with no assertion next to it, which per the contract in NOSTR_VIDEO_EVENTS.md tells a verifier the nonce binds a proof it does not.

The `.invalidKey` case clears the pending slot and generates once in the same round, so a restore landing between `generateKey` and `attestKey` does not leave that proof unattested.

mobile/docs/NOSTR_VIDEO_EVENTS.md
hm21: Documented in 72c97fed5d, both parts.

Client data hash: it is not the proof hash but `SHA-256(UTF-8(proofHash))`, one more hash over the hex digest, where `proofHash` is the same string the `proofmode` tag carries as `videoHash`. A verifier cannot guess that from the tag, so it is now written down next to the two branches.

Replay boundary: your reading is right. Nothing in the payload binds the event pubkey and the challenge is derivable by anyone holding the video, so the three fields can be lifted from one event and republished verbatim under a different pubkey carrying the same media, with both branches still passing. The doc now says a passing check means a genuine Apple device attested this media, not that this publisher's device did, and notes that binding it to the publisher means mixing the event pubkey into the challenge — which changes what the client signs and is out of scope here.

The architecture note claimed the opposite ("Includes challenge nonce to prevent replay attacks") and now points at this section instead.

mobile/docs/NOSTR_VIDEO_EVENTS.md
hm21: Handing this one to you and product rather than answering it in the PR, but three findings from digging into it that I think change the shape of the decision.

**This PR creates the linkability, it does not inherit it.** Before the branch, every proof called `generateKey` + `attestKey`, so each `keyID` was unique and no two videos were correlatable. Unlinkability is the property App Attest has over DeviceCheck, and caching the key trades it away. So the question is not "do we accept an existing identifier" but "do we accept a new one, as a side effect of a perf fix".

**Android already does the unlinkable thing.** `HardwareAttestationNotarizationProvider` mints a fresh AndroidKeyStore key per proof and deletes it immediately after reading the cert chain — no persistent identifier, no rate limit, because it never touches the network. If we accept the cached key on iOS, iPhone users' videos become correlatable across accounts while Android users' do not, with identical UI and an identical `verification` tag, and nothing anywhere saying the platform changes it.

**The material is unconsumed on both platforms.** `PROOF_VERIFIER_URL` is unset org-wide, funnelcake records presence and a sha256, and `VideoEvent.deviceAttestationPresent` is defined but never read anywhere in the app. Combined with the replay boundary in the thread above, we would be paying a permanent, irreversible cost for a field that today proves nothing on either platform.

**Proposal: keep the key cached, stop publishing on iOS until a verifier exists.** The rate limit is on generating the key; the privacy cost is on publishing it. Those are separable, and the branch only couples them because the plugin has always emitted whatever it had. Dropping the iOS `device_attestation` write keeps the entire perf win, changes no verification level (`getVerificationLevel` keys off `c2paManifestId`, not the attestation), and leaves the material on the device for a verifier to collect privately — which is the client-to-own-server shape App Attest is designed for.

One caveat if you would rather keep publishing something: dropping only `keyID` and `attestationString` is weaker than it looks. App Attest assertions carry a counter in `authenticatorData` that increments per assertion of that key — Apple's own verification steps require checking it rises — so a per-install key means publishing a monotonic per-install counter. Not an equality match like a `keyID`, but contiguous runs are quite attributable at normal posting volume.

Happy to implement whichever you land on. If the answer is that public per-install material is acceptable, I would want to pair it with binding the event pubkey into the challenge, so the thing we pay for actually proves what people will assume it proves.

mobile/lib/services/native_proofmode_service.dart
```diff
@@ -298,6 +288,39 @@ class NativeProofModeService {
...
+  /// Writes the iOS App Attest token for [proofHash] into its proof directory.
+  ///
+  /// App Attest needs Apple's attestation servers on first use, so it fails on
+  /// the Simulator, offline, and whenever Apple throttles the app. None of that
+  /// invalidates the proof itself — the PGP signature and C2PA manifest stand on
+  /// their own — so a failure here only drops the `device_attestation` field
+  /// instead of discarding the whole proof.
+  static Future<void> _writeIosDeviceAttestation(String proofHash) async {
+    try {
+      final tokenReceived = await AppDeviceIntegrity()
```
NotThatKindOfDrLiz: This is where I think the PR should change direction. Keep the App Attest key cache for performance, but do not write the raw iOS App Attest payload into proof metadata/public Nostr output yet. Publishing `{keyID, attestationString, assertionString}` creates a stable per-install identifier, and because Divine supports account switching it also reuses one App Attest key across multiple local users. Apple's App Attest guidance explicitly cautions against reusing a key among multiple users on a device. The safer shape is to keep this material local/private until there is a verifier endpoint that scopes the challenge to the account/event and emits only a verifier result or receipt publicly.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift
```diff
@@ -2,14 +2,52 @@ import CryptoKit
...
-    private let keyName = "AppAttestKeyIdentifier"
+    var assertionString: String?
+
     private let attestService = DCAppAttestService.shared
     private var keyID: String?
 
+    /// Guards provisioning so concurrent challenges share one key instead of
+    /// racing `generateKey` and burning Apple's rate limit. The cache is static
+    /// alongside it: the queue is shared, so the store it drains into has to be.
+    private static let defaults = UserDefaults.standard
```
NotThatKindOfDrLiz, line 66: A single static `UserDefaults` cache makes this one App Attest credential per install, not per account. That is the root of the cross-account correlation problem, and it also conflicts with Apple's App Attest guidance not to reuse a key among multiple users on a device. If iOS App Attest remains local/private in this PR, this can be acceptable as an implementation detail for the perf fix; if any App Attest output is public or verifier-facing, the credential should be scoped to the current Nostr account or the PR needs an explicit product/security exception.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift
```diff
@@ -19,60 +57,279 @@ final class AppDeviceIntegrity {
...
+        }
+
+        Self.pendingProvisioning.append(completion)
+
+        if Self.isProvisioning {
+            Self.lock.unlock()
+            return
+        }
+
+        Self.isProvisioning = true
```
NotThatKindOfDrLiz: Please add a provisioning watchdog before merge. Once `isProvisioning` flips true, the queue only drains if Apple's `generateKey`/`attestKey` callback eventually fires. If either callback wedges, `pendingProvisioning` grows and every later proof waits behind the stuck round until app restart. A timeout that finishes the round with `nil` preserves the new graceful-degradation behavior instead of turning one native hang into a process-wide proof stall.

mobile/docs/NOSTR_VIDEO_EVENTS.md
```diff
@@ -115,6 +115,50 @@ parsing is enforced in `mobile/packages/models/lib/src/video_event.dart` and
...
```
NotThatKindOfDrLiz: If this verifier contract stays documented, it needs Apple's assertion-counter rule too: assertion verification must check that the counter in `authenticatorData` increases for the key. That counter is also a per-key monotonic signal, so it belongs in the linkability/threat-model text. If the implementation changes to keep iOS App Attest private for now, this section should instead say public iOS App Attest publication is intentionally disabled pending a scoped verifier design.

NotThatKindOfDrLiz (changes requested):
Thanks for working through the App Attest state machine. I do not think this should merge with raw iOS App Attest material published in immutable Nostr events. The right split is: keep the cached key locally for the performance win, but stop writing public iOS App Attest payloads until Divine has a verifier design that scopes the proof correctly. Apple also cautions against reusing an App Attest key across multiple users on a device, and Divine supports account switching, so this should not become a per-install public identifier shared by every local account.

Please change the approach to keep App Attest local/private for now: do not write the iOS `device_attestation` material into proof metadata or the public event payload, update the docs accordingly, add a provisioning timeout/watchdog, and verify the resulting path on a physical iPhone before merge. A later verifier can accept App Attest material privately and publish a scoped verifier result or receipt once the challenge includes the right account/event binding and the server enforces Apple's assertion-counter rules.

rabble:
Product/security decision on @NotThatKindOfDrLiz's threads, so this can move.

**Move the attestation to publish time, and scope the App Attest key to the Nostr account.**

Reasoning, and why it isn't either of the two options on the table:

Per-install caching is the actual problem, and per-*account* scoping largely dissolves it. A keyID scoped to one account only correlates videos already signed by that account's pubkey — the correlation is redundant with the pubkey itself, so it adds no linkability the event doesn't already carry. It also satisfies Apple's guidance against sharing one App Attest key among multiple users on a device. That keeps the perf win *and* keeps publishing, so a verifier has material to consume when one exists, rather than shipping a field we've disabled.

The catch is where the key gets scoped, and it's why this is a redirection rather than a patch: **the pubkey isn't available where proofs are generated.** Neither `proofFile` call site — `clip_manager_provider.dart` or `video_editor_render_service.dart` — passes `creatorBindingAssertion`, and `VideoEditorRenderService` is fully static with a private constructor. Worse, proofs are generated at record/render time while the publishing account isn't fixed until publish time, so scoping at generation is wrong for the record-then-switch-accounts case.

Generating the assertion at publish, where the signing pubkey is known, fixes the scoping and gets @hm21's other ask for free: **bind the event pubkey into the challenge**, so the published material proves what people will assume it proves. The cached key stays; only the assertion moves.

Also still wanted, independent of the above:

- The **provisioning watchdog** Liz asked for. `isProvisioning` currently only clears if Apple's `generateKey` / `attestKey` callback fires; a wedged callback stalls every later proof behind it until restart. A timeout finishing the round with `nil` preserves the graceful degradation this PR added. Note that per-account scoping also makes `pendingProvisioning` need to be keyed by scope rather than a flat list — otherwise a second account's waiter receives the first account's credential.
- **Docs**: `mobile/docs/NOSTR_VIDEO_EVENTS.md` needs Apple's assertion-counter rule (verification must check the counter in `authenticatorData` increases for the key), and the linkability text updated to say the identifier is per-account rather than per-install.

@hm21 you offered to implement whichever way this landed — it's yours if you want it. This is a redesign rather than a clear review fix, so I'm not taking it over under the takeover gates in `PR_REVIEW.md`. Happy to talk through the publish-time seam if it's more tangled than it looks from here.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift
rabble: Added in `ca1ffedf7`.

A round token now identifies the in-flight provisioning attempt. Whichever of the real `generateKey`/`attestKey` callback or a 30s watchdog arrives first ends the round and drains `pendingProvisioning`; the loser hits a `guard isProvisioning, round == provisioningRound` and returns without touching a queue that may already belong to a later round. Timing out finishes with `nil`, which is the graceful-degradation path this PR already added for a failed round rather than a new failure mode — exactly the shape you asked for.

One deliberate trade worth flagging: a late real credential arriving after the watchdog fired is **dropped rather than stored**. Storing it would leave the cache holding a key that no waiter was ever handed, and the next proof would assert against a key it never saw provisioned. Dropping costs one rate-limited generation on the next proof, which is the cheaper half.

`swiftc -parse -target arm64-apple-ios14.0` clean. Resolving this one; your other four are answered in the decision comment above and are @hm21's to implement.

mobile/docs/NOSTR_VIDEO_EVENTS.md
hm21: Resolved by the redesign in 2debbb638f, per @rabble's call: the identifier is no longer per-install. The key is scoped to the Nostr account, so it can only correlate videos that account already signed with its own pubkey — it exposes nothing the event does not already carry, and switching accounts provisions a separate key rather than sharing one across identities. The linkability section in `NOSTR_VIDEO_EVENTS.md` is rewritten on those terms and now also names the assertion counter as part of the same per-key signal.

mobile/lib/services/native_proofmode_service.dart
hm21: Addressed in 2debbb638f, though by scoping rather than by going private — that was @rabble's call in the decision comment, so a verifier has material to consume when one exists instead of us shipping a disabled field.

The part of your objection about the challenge is now fixed: the mint moved out of proof generation entirely and into `VideoEventPublisher`, where the signing pubkey is known, and the challenge binds it as `SHA-256(UTF-8("<proofHash>:<pubkeyHex>"))`. That also removes the replay boundary the old shape had to document — the payload can no longer be lifted onto another pubkey's event carrying the same media.

Generation time could never have done this correctly: a clip can be recorded, sit in the library, and go out under an account the user switched to afterwards.

mobile/overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift
hm21: Fixed in 2debbb638f. Every piece of cached state now hangs off the scope — key id, attestation object, the pending-key slot, and the in-flight provisioning round with its queue. The queue mattered most: a flat list would have handed one account's credential to another account's waiter, which @rabble flagged when he added the watchdog.

One honest note while I was in there: I tightened the comment above the pending-slot helpers. It claimed the slot is only ever touched inside its own round and therefore needs no lock. Since the watchdog can end a round while its Apple callback is still in flight, that is not strictly true — a late callback can clear a slot a later round is using. The cost is bounded at one extra rate-limited `generateKey` and it can never surface a wrong key, so it is left unguarded, but the comment now says so instead of overstating the invariant.

mobile/docs/NOSTR_VIDEO_EVENTS.md
hm21: Added in 2debbb638f. `NOSTR_VIDEO_EVENTS.md` now has an explicit **Assertion counter** section: verification must check the counter in the assertion's `authenticatorData` and require it to be strictly greater than the highest value already seen for that `keyID`, with the note that skipping it accepts replayed assertions.

Your point about it being a per-key monotonic signal is folded into the linkability paragraph rather than left implicit — it is now described as a monotonic sequence over that account's publishes, which is the same per-account boundary as the `keyID` itself.

hm21:
@rabble @NotThatKindOfDrLiz — implemented your decision in 2debbb638f, rebased onto current `main`. Title and description are rewritten for the new design; all four of Liz's open threads have per-thread replies.

**What changed**

- The App Attest key is scoped to the Nostr account instead of the install. All cached state hangs off the scope — key id, attestation object, pending-key slot, and the in-flight provisioning round with its queue. The queue was the specific hazard @rabble called out when adding the watchdog: a flat list hands one account's credential to another account's waiter.
- The payload is minted at publish time in `VideoEventPublisher` behind a new `IosDeviceAttestationService`, and the challenge binds the signer: `SHA-256(UTF-8("<proofHash>:<pubkeyHex>"))`. That also closes the replay boundary the old shape had to document.
- The publish-time value is authoritative — it replaces whatever the stored proof held, so a token minted for a previous account cannot ride along, and a failed mint drops the field rather than falling back to a stale one.
- Docs carry Apple's assertion-counter rule and state the linkability as per-account.

**Two things worth your eyes**

1. I tightened the comment above the pending-key helpers. It claimed the slot is only ever touched inside its own round and so needs no lock; with the watchdog able to end a round while its Apple callback is in flight, a late callback can clear a slot a later round is using. Bounded at one extra rate-limited `generateKey`, never a wrong key — but the comment now says so instead of overstating it.
2. `generateAssertion` is the one Apple call with no watchdog. It is local to the Secure Enclave with no network, so I left it, but flagging it rather than having it found later.

**Still open before merge:** the physical iPhone run. App Attest is a no-op on the Simulator, so the timing win, per-account provisioning, and the assertion path are all unverified on device — I have not claimed otherwise in the description. CI is green (16/16).

NotThatKindOfDrLiz:
@hm21 I took over the remaining review fixes and pushed `8ffbb0de4` onto `fix/ios-app-attest-key-reuse`.

What changed:
- Added a Dart-side timeout around publish-time App Attest so a stuck native callback degrades to no `device_attestation` instead of hanging publish.
- Bounded the native pending-key retry path so a timed-out/late-attested pending key cannot wedge an account forever.
- Documented the legacy pre-PR iOS challenge shape for already-published events.
- Pinned Android publisher behavior, the no-signing-pubkey stale-attestation clear path, and the App Attest timeout in tests.
- Re-enabled the adjacent skipped no-ProofMode publisher test and shrank the skip baseline accordingly.

Verification after rebasing on current `origin/main`:
- `flutter test test/services/ios_device_attestation_service_test.dart test/services/video_event_publisher_native_proof_test.dart` — 11 passed.
- `swiftc -parse -target arm64-apple-ios14.0 overrides/app_device_integrity-1.1.0/ios/Classes/AppDeviceIntegrity.swift` — parses, with the direct-`swiftc` host sysroot warning.
- Pre-push hook also passed analyzer and changed-file tests: 67 passed.

I updated the PR body to remove the stale out-of-scope note about `generateAssertion` hanging the publish path. The remaining required gate is still the physical iPhone run for first-publish provisioning and later-publish assertion. Please review the takeover commit and merge when you are happy with it and the device run is done.

hm21:
@NotThatKindOfDrLiz We've tested this end-to-end on a real iPad (iOS 26.5) and everything works as expected. The branch is also rebased onto current `main` now, and all checks are green.

If it looks good on your side, an approval would unblock us — and of course happy to fix anything you still see open.

NotThatKindOfDrLiz:
@hm21 I took over the two small review fixes and pushed them to this branch.

Changes added:
- clear the publish-time iOS App Attest payload if the active account changes after minting but before signing, so the signed event never carries another account scoped App Attest key ID
- add a regression test for that account-switch publish path
- populate all NativeProofData fields in the withDeviceAttestation fixture so the copy guard covers the C2PA/CAWG identity fields too
- harden the deferred-file cleanup test after the previous CI shard exposed that ref.onDispose cleanup is asynchronous

Verification:
- pre-push analyzer passed
- pre-push changed-file tests passed: 176/176
- GitHub checks are green on 08c8915c180bcca5ad44c34ae874731f09b9db1a

I am approving from the takeover side; please review the takeover commits and merge when ready.

NotThatKindOfDrLiz (approved):
Takeover fixes look good. The account-switch attestation race is covered, the NativeProofData copy guard now exercises all fields, and CI is green on the updated head.

PR #6495, opened 2026-07-29
feat(l10n): add Vietnamese, Urdu, Chinese, and Malay language support

rabble:
Closes #6496

## Description

Adds full app UI localization for four locales covering some of Divine's largest user communities:

- **Vietnamese (`vi`)** — casual "bạn" register; keeps "loop" as the loanword Vietnamese internet usage already favors.
- **Urdu (`ur`)** — warm polite "آپ" register; rides the existing RTL support from Arabic; transliterates "loop" as لوپ.
- **Chinese, Simplified (`zh`)** — punchy short-video register (Douyin-style energy); renders the loop concept as 循环. Traditional Chinese (`zh_Hant`) can follow later if demand shows up.
- **Malay (`ms`)** — Malaysian consumer-app register ("anda", "tetapan", "muat naik"); keeps "loop" as a loanword.

Per-locale details:

- Each new ARB translates **all 3,002 messages**, including the 73 crosspost / supporter / auth-PIN / appearance keys the 18 existing locales still defer via `_knownUntranslatedDebt` — so these four ship with **100% key coverage** (gen-l10n reports zero untranslated messages for vi/ur/zh/ms vs 73 for every existing locale).
- Test-guarded copy respected everywhere: Keycast remote-signing and the 8 Nostr signature-verification keys are genuinely translated; CSAM report reasons stay distinct from general child-safety copy; the age-gate signer-unreachable message keeps its distinct check-your-connection remedy.
- Brand voice follows `brand-guidelines/TONE_OF_VOICE.md` per locale (playful UI copy, calm legal/safety copy, AI-generated content never called "creative").
- Wiring: `LocalePreferenceService` picker (+4), `LanguagePreferenceService` content tagging (+2: `ur`, `ms` — `vi`/`zh` were already listed), Android `locales_config.xml`, iOS `CFBundleLocalizations`, Xcode `knownRegions`, and regenerated `AppLocalizations`.
- Two commits: tests-first (red by design), then implementation (green). Squash-merge keeps main tidy.

**Related Issue:** N/A — direct product request (large user communities in Vietnam and Pakistan).

## Out of Scope

- Traditional Chinese (`zh_Hant`/`zh_TW`) and Punjabi — can be added later using this same blueprint.
- Native-speaker review of the agent-generated translations — welcome as follow-up PRs; the structure guards (placeholders/ICU/guard keys) are test-enforced.
[...]

PR #6413, opened 2026-07-26
feat(supporter): enforce server-verified subscriptions

rabble:
Closes #6414

## Summary

- add the feature-gated Support Divine flow and store product loading
- connect StoreKit/Play purchase proofs to the canonical `divine-supporters` Worker
- keep local store callbacks from granting entitlement before server verification
- capture the authenticated Divine pubkey and preserve purchase lifecycle state across async results
- acknowledge store purchases only after a successful canonical claim
- add regression coverage for pending, confirming, restore, account binding, and verification failures

## Companion service

Worker implementation and deployment PR: https://github.com/divinevideo/divine-supporters/pull/1

Live Worker: https://divine-supporters.protestnet.workers.dev

## Verification

- targeted supporter and IAP tests passed
- `flutter analyze` passed for the affected implementation paths
- repository pre-push analyzer passed
- repository pre-push changed-file tests passed: 67 tests

## Remaining launch gates

- complete a real Apple sandbox transaction after the App Store Connect issuer UUID is available
- complete a real Google Play sandbox transaction with a licensed test account
- replace the current account-scoped bridge with a durable device-scoped pending-claim queue before launch
- add Funnelcake sanitized supporter projection and cross-client consumption

Issue #6366, opened 2026-07-24
feat(crossposting): add manual video crossposting


NotThatKindOfDrLiz:
## Summary
Add manual video crossposting from the mobile share sheet for a creator's own videos.

## Scope
- Fetch connected crossposting platforms from the crossposter service.
- Let creators select connected platforms from the share sheet.
- Submit idempotent per-video crosspost jobs and poll them to terminal states.
- Surface posted permalinks, localized failure states, and reconnect prompts.

## Follow-up
- Complete a full l10n pass for the crossposting strings currently allowlisted as untranslated debt.
- Validate authenticated end-to-end posting with a real Divine session.

PR #6497, opened 2026-07-29
feat(l10n): translate remaining deferred keys in all 17 locales

rabble:
## Description

Completes the deferred l10n pass so the whole app is translated in every shipped language. The 73 keys that 17 existing locales previously fell back to English for (tracked as `_knownUntranslatedDebt`) are now translated — **1,241 new translations across am, ar, bg, de, es, fil, fr, id, it, ja, ko, nl, pl, pt, ro, sv, tr**.

Feature groups covered:

- **Crosspost (22 keys, #6366)** — share-sheet crossposting to connected platforms: statuses, errors, reconnect prompts
- **Appearance experiment (5 keys, #6217)** — System/Light/Dark options
- **Delete account / username-burn (8 keys, #6126, #6137)** — calm, precise safety copy
- **Community content-warning suggestions (7 keys, #4771)**
- **Email-verification PIN fallback (12 keys, #5578)** — code entry, resend, cooldown
- **Supporter subscription (19 keys, #6413)** — hero copy, purchase states, restore, ownership conflict, disclaimer

After this PR **every shipped locale defines 100% of `app_en.arb` keys** and `flutter gen-l10n` reports zero untranslated messages for all locales. `_knownUntranslatedDebt` is emptied and remains as the intentional-deferral mechanism for future passes.

Translation notes:

- Each locale was matched to its existing file register (e.g. de informal "du", es Rioplatense voseo per the file's dominant register, ja casual 〜よ/〜ね, bg informal "ти", tr "sen").
- The magic word `DELETE` stays in English in confirmation prompts, matching each file's existing hint convention ("Ketik DELETE", "DELETE と入力してね", …).
- Placeholders `{platform}`, `{username}`, `{time}` preserved verbatim everywhere (verified programmatically).
- Entries merged at their English-neighbor positions, preserving each file's 4-space style with inline translator metadata; diffs are pure insertions (+105 lines per ARB, zero deletions).

**Related Issue:** Relates to #6496

## Out of Scope

Pre-existing issues noticed during the pass and documented here rather than fixed (unrelated to this diff):

- `app_am.arb` `blueskyFailedToUpdateCrosspost` translates "crosspost" as "መስቀለኛ መንገድ" ("crossroad") — likely a mistranslation; the new keys use "ተሻግሮ ማተም" consistently. Fixing the old key needs a native-speaker decision.
- Typos in `app_ko.arb` ("딱굱1거리는") and `app_pl.arb` ("czkać"; "wyślе" containing a Cyrillic `е` homoglyph in two places).
- English-source "diVine"/"DiVine" casing drift in a few `app_en.arb` values (also flagged in #6495).

## Verification

- `flutter test test/l10n test/assets test/services/locale_preference_service_test.dart test/services/language_preference_service_test.dart` — 130/130 pass, including the strengthened `arb_consistency_test` with an empty debt list
- Structural verification per locale: JSON validity, full message-key parity with `app_en.arb`, all 4 placeholder tokens present in all 17 locales
- `flutter gen-l10n` — regenerated and committed; zero untranslated-message warnings for any locale
- `flutter analyze` / `dart format` — clean
- check-l10n skill: ARB consistency ✅; hardcoded-English scanner N/A (no non-generated Dart changes under `lib/`)
- Translation quality spot-checked across de/fr/ja/ko/ar/tr/fil for supporter, crosspost, auth-PIN, delete-account, and appearance copy

## Type of Change


mbradley (commented):
Submitting as a comment rather than an approval on purpose. Structure and mechanics I checked thoroughly; whether the copy reads naturally I did not check and can't, in any of these 17 languages. That makes me the wrong signer for a translation PR, so the approval should come from someone who can speak to the copy.

**What I verified.** All 17 locales at 3,002 keys with zero missing and zero empty values; every interpolated placeholder present in every locale; `gen-l10n` reporting no untranslated messages; the Keycast, signature-verification, CSAM, and age-gate guards still holding; and the stated 130/130 reproducing exactly. I also ran three sweeps over all 1,241 new values: script integrity, a sloppiness scan (double spaces, repeated words, unbalanced braces, stray whitespace, repeated punctuation), and a one-char near-miss detector. The sweeps surfaced the two Polish items below and nothing else; the 55 near-miss candidates were all inflections, capitalisation variants, or artifacts of my tokenizer splitting on Arabic diacritics.

**What I could not verify:** whether the copy reads naturally. There is no CLI dictionary available for these languages on my machine, and idiomatic quality is not machine-checkable regardless. Treat this review as covering structure and mechanics only.

The three Out of Scope claims check out. `czkać`, `딱굱1거리는`, and the Amharic `መስቀለኛ መንገድ` are all pre-existing and absent from this diff. One `wyśl` token does land in an added Polish line, so I codepoint-audited it: clean Latin, so the homoglyph was not propagated into the new copy.

**Needs a fix from you: a Polish typo in the new copy.** `authVerificationErrorPinUnavailable` reads "Wpiswanie kodu". That token appears exactly once across the whole file and nowhere else in any Polish value, and the file uses "Wpisz"/"wpisz" elsewhere, so it reads like a dropped `y` from "Wpisywanie". Zero occurrences on `main`, so it came in with this diff. I'm deliberately not pushing a fix: correcting spelling in a language I don't read is the author's call, not the reviewer's, even when the inference looks safe. Confirm and fix.

**`c9049580` replace three Cyrillic homoglyphs in `app_pl.arb`.** You scoped these out and I fixed them anyway, because the file is already being touched and the change is a pure encoding correction with no visible difference (`U+0435` → Latin `e`), so no Polish knowledge is required to verify it. One correction to your note: there are three, not two. `videoGridDeleteConfirmNote`, `shareMenuDeleteWarning`, and `shareMenuDeleteRelayWarning`. Anyone fixing only what the note said would have left one behind. Regenerated l10n output is included in the commit.

**`9dcc8a59` guard each locale against stray-script characters.** A scope call I made, so push back if you'd rather own it. Those homoglyphs were invisible to every check in the repo: they parse as valid JSON, render identically, and pass key parity, placeholder parity, and all four copy guards. This pins the allowed scripts per locale so wrong-script text can't reach copy again. Mutation-tested both ways: injecting a Cyrillic `е` into Polish turns it red naming the exact character and key, and a locale with no map entry fails loudly rather than being silently skipped, so a future locale has to make the decision explicitly. It lives in its own file rather than in `arb_consistency_test.dart` specifically so it adds no merge surface with #6495, and the map covers vi/ur/zh/ms so it stays green whichever order the two land.

**Merge order against #6495.** Both branches are off the same base and both touch `arb_consistency_test.dart`, so I test-merged them. `git merge-tree` reports no conflicts, the merged file keeps both your empty debt set and the placeholder guard from #6495, and the merged tree runs green with `gen-l10n` clean and zero untranslated across all non-template locales. Either order is safe.

One observation, no action needed: `shareSheetCrosspost` and `crosspostSubmit` in id and it are the only two of the 1,241 that stayed byte-identical to English. English uses the same bare-label-on-button pattern, so I don't think anything is wrong; flagging it only so you can confirm the loanword was deliberate rather than a gap.

Validation on the pushed state: `flutter test test/l10n test/assets test/services/locale_preference_service_test.dart test/services/language_preference_service_test.dart` at 131/131, `flutter analyze lib/l10n test/l10n` clean, `dart format` clean, `gen-l10n` re-run and in sync.

Review the two commits I pushed, revert anything you disagree with, and merge once the Polish typo is resolved.

realmeylisdev (commented):
Ordinary review, feedback only — no branch changes from me.

I re-verified the mechanics independently and they hold. What I added is the copy read in these languages, which is the gap mbradley flagged as out of his reach: one new Amharic defect, the Polish typo still unfixed on `9dcc8a5`, and a gap in the new script guard I mutation-tested. Details inline; everything else I checked is clean.

mobile/lib/l10n/app_am.arb
```diff
@@ -1571,6 +1637,29 @@
...
+    "crosspostStatusUploading": "በመስቀል ላይ",
+    "crosspostStatusProcessing": "በማቀነባበር ላይ",
+    "crosspostStatusPosted": "ታትሟል",
+    "crosspostStatusFailed": "አልተሳካም",
+    "crosspostStatusSkipped": "ተዝሏል",
+    "crosspostStatusNeedsReauth": "እንደገና መገናኘት ያስፈልጋል",
+    "crosspostViewPost": "ልጥፉን ይመልከቱ",
+    "crosspostReconnectPrompt": "መለጠፍን ለመቀጠል {platform}ን በተሻግሮ ማተም ቅንብሮች እንደገና ይገናኙ።",
+    "@crosspostReconnectPrompt": {"placeholders": {"platform": {"type": "String"}}},
+    "crosspostReconnect": "እንደገና ገና",
```
realmeylisdev: `እንደገና ገና` isn't a phrase. The second token is the last two syllables of the first one repeated — `እንደ` + `ገ ና` then `ገ ና` again — with the verb dropped. Standing alone, `ገና` means "yet/still", and that is exactly how this file uses it elsewhere (`shareMenuDeleteFailedNotInitialized`: `መሰረዝ ገና ዝግጁ አይደለም`; `soundsSavedEmptyTitle`: `ገና የተቀመጠ ድምፅ የለም`).

Three sibling keys **in this same diff** already carry the file's verb for the concept:

- `crosspostReconnectPrompt` → `… እንደገና ይገናኙ።`
- `crosspostErrorUnauthorized` → `መለያዎን እንደገና ይገናኙ፣ …`
- `crosspostStatusNeedsReauth` → `እንደገና መገናኘት ያስፈልጋል`

This is the most visible of the four — it renders as the primary CTA, a `DivineButton` in `_ReconnectPrompt` (`lib/widgets/crosspost_sheet.dart:400`), so an Amharic user sees it as the button they are supposed to press.

```suggestion
    "crosspostReconnect": "እንደገና ይገናኙ",
```

Worth noting why the sweeps couldn't have caught it: the string is valid Ethiopic end to end, the repetition is a suffix rather than a whole word, and there is no pre-existing key with the English value "Reconnect" to diff the translation against — so no script check, sloppiness scan, or consistency check had anything to grab.

One more for whoever reads Amharic, offered as a glance rather than a finding: `crosspostStatusSkipped` on line 1649 is `ተዝሏል`, while this file's own stem for "skip" is `ዝለል` (`videoErrorSkip`, `feedSkip`, `videoMetadataC2paMissingSkip`). I can't call whether the passive is correctly formed there, so I'm not claiming it's wrong — just that it's the other value in this file I couldn't confirm.

mobile/lib/l10n/app_pl.arb
```diff
@@ -1351,6 +1369,11 @@
     "authVerificationErrorOAuthExchange": "Weryfikacja nieudana. Spróbuj zarejestrować się ponownie.",
     "authVerificationErrorSignInFailed": "Logowanie nieudane. Spróbuj zalogować się ręcznie.",
     "authVerificationEmailAlreadyRegistered": "Ten adres e-mail jest już zarejestrowany. Zaloguj się zamiast tego.",
+    "authVerificationErrorPinInvalid": "Ten kod nie pasuje. Sprawdź go i spróbuj ponownie.",
+    "authVerificationErrorPinExpired": "Ten kod wygasł. Dotknij Wyślij ponownie, aby otrzymać nowy.",
+    "authVerificationErrorPinLocked": "Zbyt wiele prób. Dotknij Wyślij ponownie, aby otrzymać nowy kod.",
+    "authVerificationErrorPinFailed": "Nie udało się zweryfikować tego kodu. Spróbuj ponownie.",
+    "authVerificationErrorPinUnavailable": "Wpiswanie kodu jest teraz niedostępne. Dotknij linku w e-mailu albo wyślij ponownie, aby otrzymać nowy.",
```
realmeylisdev: Seconding mbradley rather than re-deriving it — anchoring it on the line so it doesn't get lost in a review body, since it's still unfixed on `9dcc8a5`.

`Wpiswanie` → `Wpisywanie` (gerund of *wpisywać*). Zero occurrences on `origin/main`, so it arrived with this diff, and `Wpisywanie` appears nowhere else in the file, so there's no in-file precedent pulling either way. I swept every 5+ character token in the 73 new Polish values against the rest of the file; this is the only one that isn't a Polish word.

```suggestion
    "authVerificationErrorPinUnavailable": "Wpisywanie kodu jest teraz niedostępne. Dotknij linku w e-mailu albo wyślij ponownie, aby otrzymać nowy.",
```

mobile/test/l10n/arb_script_integrity_test.dart
```diff
@@ -0,0 +1,129 @@
...
+            .cast<String, Object?>();
+
+        for (final entry in arb.entries) {
+          if (entry.key.startsWith('@')) continue;
+          final value = entry.value;
+          if (value is! String) continue;
+
+          for (final rune in value.runes) {
+            final script = _scriptOf(rune);
+            if (script == null || script == 'LATIN') continue;
```
realmeylisdev: Genuinely useful guard, and skipping Latin is the right call — `Divine`, `Nostr`, `NIP-09` appear in every locale. But it makes the check one-directional: it catches Cyrillic-in-Polish, the bug it was written for, and misses Latin-in-Cyrillic, which is the same bug pointed the other way and the more likely direction for `bg`.

Mutation-tested on this branch. I replaced the Cyrillic `е` (U+0435) with a Latin `e` (U+0065) inside a Cyrillic word in `app_bg.arb`:

```
"crosspostStatusFailed": "Нeуспешно"
 codepoints: 41d 065 443 441 43f 435 448 43d 43e
```

Pixel-identical to `Неуспешно`. This guard passes, and so does all of `test/l10n` — 77/77 green. Reverted straight after; nothing pushed.

A narrow tightening closes it: blank out every `{…}` region first, so ICU argument names like `{count}` can't sit against native script, then flag any Latin letter directly adjacent to a Cyrillic one. I ran that over the real `app_bg.arb` and it produced **zero false positives across all 3,002 keys** while catching the injection above.

The same shape would extend to `am` and `ar`, but both are lower-risk — neither script has Latin lookalikes — and Amharic legitimately affixes onto Latin tokens (`በDivine`, `Divineን`, `{username}ን`), so those would need the adjacency allowance anyway. `bg` is the one that matters.

Your file and your scope call. Perfectly reasonable as a follow-up if you'd rather land this as-is.

mobile/test/l10n/arb_script_integrity_test.dart
```diff
@@ -0,0 +1,129 @@
...
+            if (allowed!.contains(script)) continue;
+
+            fail(
```
realmeylisdev: Nit: failing on the first bad rune means a file with N violations takes N runs to clear.

That's the scenario that just played out immediately upstream of this commit — three Cyrillic homoglyphs in `app_pl.arb`, and the PR note undercounted them at two. Collecting violations and failing once with the full list would have surfaced all three in a single run, and would have made the undercount obvious.

rabble:
Updated this branch on current main. Fixed the Polish and Amharic copy, removed the Polish Cyrillic homoglyphs, regenerated the affected localization classes, and added the ARB script-integrity regression test. Focused l10n tests and all current CI checks pass.

PR #6467, opened 2026-07-28
refactor(theme): migrate UI to semantic color tokens for light mode

hm21:
## Description

The appearance selector (System / Light / Dark) and its `AppearanceCubit` already shipped, but picking **Light** barely changed anything: the vast majority of widgets read hardcoded `VineTheme.*` constants, which are dark by definition. `MaterialApp` switched to `VineTheme.lightTheme` correctly — the widget tree just never asked the theme for its colors.

This PR makes the light palette actually reach the UI, then fixes what device testing surfaced.

### Making the palette reach the UI

- **Extends `VineThemeColors`** with the semantic tokens the light palette needs and the dark one had been faking with one-off constants: `containerLow`, `onSurfaceMuted`, `outlineMuted`, `outlineDisabled`, `skeleton`, `errorContainer`, `onErrorContainer`, `inverseSurface`, `inverseOnSurface`, plus `onNav` / `onNavMuted` (nav foreground), `primaryContainer` (active chips) and `mediaCard` (neutral frame around embedded media).
- **Adds a `context.vineColors` accessor** on `BuildContext` that resolves the active `VineThemeColors` extension. It falls back to `VineTheme.darkColors` rather than Material defaults, so the many widget tests that pump a bare `MaterialApp` without the Divine theme keep rendering exactly as before instead of turning grey — that fallback is what kept this migration from touching hundreds of test files.
- **Migrates the app and `divine_ui`** off static color constants and onto those tokens.

### Two root causes behind "light mode looks broken"

- **`VineTheme.*Font()` defaulted `color` to `whiteText`**, and 266 call sites relied on that default. The first pass rewrote the explicit color arguments but not the implicit ones, so each of those became white text on a now-white surface. The default is removed entirely: `color` is `Color?`, so a forgotten color inherits from the ambient text style instead of forcing white, and the failure mode cannot come back silently. There are now zero implicit-color `Font()` calls in the codebase.
- **The nav bar and app bars were pinned to the dark brand green.** They now follow the palette, including status-bar icon brightness and the Android system navigation bar.

### Design rules this establishes

The migration is not "theme everything". Content whose brightness is not theme-driven keeps fixed colors, and each such site carries a comment saying why:

- Text and scrims layered on video, camera preview or a photo thumbnail stay light-on-dark in both modes (40 alpha scrims, 124 explicit `whiteText`).
[...]

PR #6498, opened 2026-07-29
feat(profile): point badge detail sheet at badges.divine.video

### LINKED ISSUE #6499: Badge detail sheet should point people to badges.divine.video
When someone taps a badge on a profile, the badge detail sheet explains the badge (name, description, awarded by, recipients) but gives no path to learn more or create badges.

Add a short explainer line and a link at the bottom of the sheet that opens badges.divine.video (the bundled Divine Badges app in the in-app sandbox), so people can make their own badges and understand how badges work.

rabble:
## Description

Tapping a badge chip on a profile opens the badge detail sheet. The sheet now ends with a short explainer ("Badges are little awards anyone can make on Nostr...") and a "Make your own at badges.divine.video" button that opens the bundled Divine Badges app in the in-app sandbox (same bridge as the Badges dashboard's "Open badges app" button), so people can create their own badges and learn how badges work.

- Adds a footer to `_ProfileBadgeDetailsSheet` (text + `DivineButton` link)
- Pops the sheet and pushes `NostrAppSandboxScreen` for `bundled-badges`
- Moves the badges app lookup into a shared `divineBadgesNostrApp` helper in `nostr_app_bridge_repository` (previously a private helper in `badges_screen.dart`)
- New l10n keys `profileBadgeFooterBody` / `profileBadgeFooterLink`; non-English locales fall back to English and the keys are tracked in `_knownUntranslatedDebt` for the next l10n pass

**Related Issue:** Closes #6499

## Out of Scope

- Translating the new footer copy (deferred to the next l10n pass, tracked in `_knownUntranslatedDebt`)

## Verification

- `flutter analyze` — no issues
- `flutter test test/widgets/profile/profile_header_widget_test.dart` — 64 tests pass, including the new "badge detail sheet links into the badges app" widget test
- `flutter test test/screens/badges/badges_screen_test.dart` — pass
- `flutter test test/l10n/arb_consistency_test.dart` — pass
- `flutter test` in `mobile/packages/nostr_app_bridge_repository` — 54 tests pass, including the new `preloaded_nostr_apps_test.dart`
- check-l10n hardcoded-string scanner on changed files — clean

## Type of Change


dcadenas (changes requested):
### What was wrong
The badges links could fail when a remote directory entry used the same slug with a different ID, because the sandbox route looked up only the bundled ID and showed the missing-app screen.

### What I changed
I changed both badges launch paths to pass the bundled badges app entry through navigation, and I added a regression test for a remote `app-badges` entry. The fix is in `fix(profile): preserve bundled badges sandbox launch` (`fffce94`). Mobile CI passed.

### What I still need from you
I still need you to route these links through a current-policy resolver before passing an app entry. The repaired paths now use the static bundled manifest, so a revoked or policy-narrowed directory entry cannot stop or constrain the launch. Please decide which manifest owns the stable badges identity, then cover approved replacements, revocation, and narrowed bridge permissions before merge.

I made the changes I could. Please review them and address the remaining decisions before merge.

hm21 (commented):
Fixed the l10n gap directly in 941d9a9813: both footer keys were English-only with `_knownUntranslatedDebt` entries, so all 17 non-English locales rendered a half-English sheet. Translated per locale using each one's existing badge term, removed the debt entries, and added an `arb_consistency` guard so neither key can fall back to English again.

One thing I left for you, because it needs your call:

`extra: divineBadgesNostrApp` makes `SandboxRouteCubit` short-circuit (`initialApp != null` → `SandboxRouteResolved`, `load()` returns early), so the directory is never consulted. That fixes the id-mismatch bug from the slug-keyed merge, but it also means a remote revocation of the badges app (`isApproved: false`, covered by `fetchApprovedApps lets the remote directory hide a bundled starter app`) no longer blocks these two buttons, and a tightened remote `allowedMethods`/origin set is ignored. The badges-screen button honoured revocation before this PR. Resolving by slug instead of id would fix both, but that touches shared sandbox routing, so I did not push it.

Verified: `flutter analyze lib test` clean, `profile_header_widget_test` (74), `badges_screen_test`, `arb_consistency_test`, and the `nostr_app_bridge_repository` suite (54) all pass.

rabble:
Addressed the policy-resolution blocker in `72d9b46` (`fix(apps): resolve sandbox route by slug so badge links honor directory policy`).

**Which manifest owns the stable badges identity:** the slug. The directory merge (`_mergeWithPreloadedApps`) is already slug-keyed, so that is the identity layer. The bundled manifest owns the offline/bootstrap fallback; the remote directory owns policy (approval, allowed methods, origins).

**What changed:**
- Both badge launch paths (profile badge sheet footer, badges dashboard button) now navigate with the `badges` slug only — no `extra: divineBadgesNostrApp` bypass.
- `SandboxRouteCubit` matches on app ID **or** slug, so the directory is always consulted before launch.

**Coverage of the three cases:**
- *Approved replacement*: remote `app-badges` entry (different ID, same slug) resolves and wins — cubit test `emits [resolved] when load matches the app by slug` + updated badges screen test.
- *Revocation*: a non-approved remote entry removes the slug from the merged directory (already covered by `fetchApprovedApps lets the remote directory hide a bundled starter app`), and the cubit then emits `SandboxRouteNotFound` — both buttons now hit this path again.
- *Narrowed permissions*: the remote entry replaces the bundled one wholesale in the merge, so tightened `allowedMethods`/origins apply automatically.

`flutter analyze lib test` clean; `sandbox_route_cubit_test`, `badges_screen_test`, `profile_header_widget_test` all pass.

hm21 (approved):
Slug-based resolution is the right call — `grantKey` is already slug-first, so the directory, the grant store, and both launch paths now agree on one identity. Revocation and narrowed permissions apply to both buttons again.

Pushed one test fix in `96c95fe`: the badges-button test re-implemented the cubit's id-or-slug match inside its own stub route, so it stayed green even if production resolution regressed. It now asserts the path parameter the button pushes; the matching itself stays covered by `sandbox_route_cubit_test`.

The red CI is not yours. `main` at 46f4eb7fb1 (#6350) is broken: `general_settings_screen.dart:85/89` reference undefined `_titleStyle` / `_subtitleStyle` (the light-mode migration renamed them to `*Of(context)`), and `crossposting_settings_screen.dart:320` has a `VineTheme.titleMediumFont()` without a color, which trips the implicit-font-color ratchet. Main's own Mobile CI run fails identically — this PR goes green once main is fixed.

Verified locally: `flutter analyze lib test` clean; `sandbox_route_cubit_test`, `badges_screen_test`, `apps_sandbox_route_test`, `profile_header_widget_test`, `arb_consistency_test`, and the `nostr_app_bridge_repository` suite all pass.

hm21:
Rebased onto `main` @ bc091d29f (`1cba942`, no conflicts).

CI was red on `96c95fe` for reasons unrelated to this PR — main itself did not compile: #6467 renamed `_titleStyle` / `_subtitleStyle` into `_titleStyleOf(context)` helpers while #6350 branched earlier and kept the old names, so Analyze, Generated Files and all four test shards failed. #6534 fixed that on main; picking it up turns every check green.

rabble:
Rebased onto current `origin/main` and force-pushed with lease at `9c49dcbfa`.

The only rebase conflict was the l10n debt allowlist. Validation then exposed that the squashed branch had generated translations but had omitted the corresponding non-English ARB source entries, so I restored the footer translations across all 21 non-English locales and regenerated localizations.

Verification:
- `flutter test test/l10n/arb_consistency_test.dart test/widgets/profile/profile_header_widget_test.dart test/screens/badges/badges_screen_test.dart test/router/apps_sandbox_route_test.dart` — 77 passed
- pre-push locale consistency, analyzer, and changed-file tests — passed
- branch is now mergeable; CI is running on the new head

@dcadenas the standing requested-change review predates the slug-policy fix and this rebase; please re-review the current head when ready.

rabble:
@dcadenas re-requesting review — your changes-requested predates the fix for it.

Your ask was to route these links through a current-policy resolver rather than passing the bundled manifest, so a revoked or policy-narrowed directory entry can still stop or constrain the launch. That's what shipped: resolution is now **slug-based** rather than by bundled id. `grantKey` was already slug-first, so the directory, the grant store, and both launch paths now agree on one identity, and revocation plus narrowed `allowedMethods`/origin sets apply to both buttons again.

hm21 reviewed and approved that approach, and pushed `96c95fe` fixing a test that had re-implemented the cubit's id-or-slug match in its own stub route and so would have stayed green through a production regression. Matching is now covered by `sandbox_route_cubit_test` and the button test asserts the pushed path parameter instead.

Two commits landed today (`1ca62f12`, `9c49dcbf`) rebasing onto current main and restoring the badge footer translations, so this needs your eyes on the current head rather than the version you reviewed. All checks green.

rabble:
@dcadenas Your blocker is addressed — but by slug resolution rather than by passing an app entry.

Your ask was to route these links through a current-policy resolver so a revoked or policy-narrowed directory entry can still stop or constrain the launch. `83e69417d` (now `ea4b4b8` after the rebase below) drops the `extra: divineBadgesNostrApp` short-circuit entirely and resolves the sandbox route **by slug**. `grantKey` was already slug-first, so the directory, the grant store, and both launch paths now agree on one identity. Revocation (`isApproved: false`) and narrowed `allowedMethods`/origins apply to both buttons again, and the id-mismatch bug that motivated `ffea578d7` stays fixed because the lookup no longer depends on the bundled id.

@hm21 reviewed and approved that approach, and pushed a test fix (the badges-button test had re-implemented the cubit's id-or-slug match inside its own stub route, so it stayed green even if production resolution regressed; it now asserts the pushed path parameter, with the matching itself covered by `sandbox_route_cubit_test`).

Just rebased onto current `origin/main` (`7e03abdce`) — GitHub was reporting a conflict in `test/l10n/arb_consistency_test.dart`, where main's Bluesky-disclosure locale test and this branch's badge-footer locale test had landed on the same lines. Both tests are preserved as separate blocks.

Verified on the rebased head: `flutter analyze lib test` clean, `arb_consistency_test` (9), `sandbox_route_cubit_test` + `badges_screen_test` + `profile_header_widget_test` (82) all green.

Could you take another look?

rabble:
@dcadenas — your three asks landed in `536ec9304` (pushed the day after your review, so you haven't seen it). Rebased onto current `main` since; all checks green.

**1. Route through a current-policy resolver.** Both badge paths now push the **slug only**, no `extra:` — `badges_screen.dart:135` and `profile_header_identity.dart:513`. With no `extra`, `initialApp` is null, so `SandboxRouteCubit` has to call `fetchApprovedApps()`. The static bundled manifest no longer short-circuits the launch.

**2. Which manifest owns the stable badges identity.** The directory merge is keyed by slug, so **the slug is the identity**: the bundled manifest owns the offline/bootstrap fallback, the remote directory owns policy (approval, methods, origins).

**3. Approved replacements, revocation, narrowed bridge permissions.** `SandboxRouteCubit.load()` matches `app.id == _appId || app.slug == _appId`, so a remote replacement under a different ID resolves — which keeps your original bug fixed without the bypass — while a non-approved remote entry hides the app and narrowed methods/origins apply, because every launch now goes through the directory. Covered by `sandbox_route_cubit_test.dart` (slug match, ID match, neither-matches, and the throwing-directory case).

**One thing to flag rather than let you find it:** the `extra` bypass still exists at the route level — `apps_routes.dart:42` reads `state.extra`, and the general directory launch path at `nostr_app_launch_mode.dart:74` still passes `extra: app`. That's pre-existing and outside this PR; the entry there comes from the directory, so it's already policy-checked. Happy to open a follow-up if you'd rather the bypass not exist at all.

Issue #4175, opened 2026-05-09
Profile edits silently drop lud16, lud06, website, and other Kind 0 fields when loaded from REST API


rabble:
## Summary

Editing a profile silently strips every field beyond the explicit form inputs (`display_name`, `about`, `nip05`, `picture`, `banner`) when the loaded `currentProfile` came from the Funnelcake REST API. Most notably this means **lightning addresses (`lud16`, `lud06`)** and **`website`** disappear from the user's published Kind 0 the next time they edit any profile field.

This is the **same root cause** that #4021 / #4022 fixed for `nip05` specifically, generalized to every other field. The earlier fix added a one-off fallback for nip05 only, so today nip05 is preserved but lud16/lud06/website/bot/etc. are not.

## Root cause

`UserProfile.fromUserProfileFound` in `mobile/packages/models/lib/src/user_profile.dart:99` populates the typed fields from the REST response but sets `rawData: const {}`:

```dart
factory UserProfile.fromUserProfileFound(UserProfileFound result, {String? eventIdPrefix}) {
  final p = result.profile;
  return UserProfile(
    pubkey: p.pubkey,
    name: p.name,
    displayName: p.displayName,
    about: p.about,
    picture: p.picture,
    banner: p.banner,
    website: p.website,        // ← stored in typed field
    nip05: p.nip05,            // ← stored in typed field
    lud16: p.lud16,            // ← stored in typed field
    rawData: const {},         // ← but rawData is empty
    createdAt: DateTime.now(),
    eventId: '${eventIdPrefix ?? 'rest'}-${p.pubkey}',
  );
}
```

`ProfileRepository.saveProfileEvent` (`mobile/packages/profile_repository/lib/src/profile_repository.dart:529`) then composes the Kind 0 content by spreading `currentProfile.rawData` first:

```dart
final profileContent = {
  if (currentProfile != null) ...currentProfile.rawData,    // empty for REST profiles
  'display_name': displayName,
  'about': about,
  'nip05': ?effectiveNip05,    // special-cased fallback to currentProfile.nip05 (lines 523-527)
  'picture': picture,
  'banner': banner,
};
```

Result: a profile loaded from the REST API loses everything that isn't explicitly in this dict on the next save. The nip05 special-case from #4022 is the only field that survives — `lud16`, `lud06`, `website`, `bot`, `nip05_verified`, and any custom-client fields are dropped.

## Reproduction

[...]

Issue #4021, opened 2026-05-05
fix: profile edits strip verified NIP-05 handle (username displacement)


realmeylisdev:
## Summary

Editing a profile (photo, bio, display name) can silently publish a Kind 0 event without the `nip05` field, breaking the `<username>.divine.video` → pubkey binding at the name server. The affected user loses their verified handle and `ike.divine.video` (or equivalent) stops routing to their profile.

## Root Cause

Two independent code paths in `ProfileEditorBloc._saveProfile` and `ProfileRepository.saveProfileEvent`:

### 1. Aggressive `clearNip05` flag (`profile_editor_bloc.dart`)

```dart
// Before — incorrectly set to true whenever username field is empty
final clearNip05 = !isExternal && username == null;
```

`username` comes from `_nip05Controller.text`. When the profile editor opens, the text controller is populated asynchronously by a `BlocListener` after `MyProfileLoaded` fires. If a relay race condition returns a stale Kind 0 **without** the NIP-05 field as the "fresh" profile, `MyProfileLoaded` emits `extractedUsername: null`, the text controller is never set, and `clearNip05 = true` — causing `saveProfileEvent` to call `profileContent.remove('nip05')` even though the user never touched the username field.

### 2. Funnelcake REST API profiles have empty `rawData` (`profile_repository.dart`)

```dart
factory UserProfile.fromUserProfileFound(...) {
  return UserProfile(
    ...
    nip05: p.nip05,       // ← populated
    rawData: const {},    // ← empty
    ...
  );
}
```

`saveProfileEvent` builds the Kind 0 content by spreading `currentProfile.rawData`. For Funnelcake-sourced profiles this spread is empty, so the NIP-05 isn't carried through even when `clearNip05 = false`.

## Steps to Reproduce

1. User has a verified divine.video handle (e.g. `_@ike.divine.video`).
2. User opens Edit Profile.
3. A relay returns a stale Kind 0 (without `nip05`) as the "fresh" profile — OR the Funnelcake REST API answered first on initial install.
4. User edits their bio or photo without touching the username field and taps Save.
5. App publishes Kind 0 without `nip05` → name server loses the pairing → `ike.divine.video` stops routing.

## Expected Behaviour

Editing bio/photo/display name **never** modifies the user's NIP-05 unless they explicitly change the username field.

## Actual Behaviour

The verified NIP-05 is stripped from the published Kind 0, permanently unlinking the subdomain from the user's pubkey.

[...]

PR #6500, opened 2026-07-30
fix(profile): stop the publish seed from erasing Kind 0 fields

realmeylisdev:
## Description

Two defects in the same profile-publish path were destroying Kind 0 fields on relays.

**1. A null seed published an empty map.** `_resolvePublishSeed` returned before attempting any fetch when `currentProfile` was null, so the Kind 0 was composed from `{}`. The pubkey was available the whole time — this publish signs through `_nostrClient`, so its public key *is* the event's author. Everything outside the form set was erased: `name`, `nip05`, `lud16`, `lud06`, `bot`, monetization links, unknown client keys, and every NIP-39 `i` tag. This is reachable through `drivePendingSave`, the primary save path.

**2. The richness comparison inverted.** Funnelcake models every profile metadata field as a non-nullable Rust `String`, so a Kind 0 key that does not exist is served as `""`. `UserProfile.fromUserProfileFound` admits every non-null field into `rawData`, so a REST-hydrated profile out-counted the real Kind 0 on placeholders alone and won `_resolvePublishSeed`'s comparison — after which the publish path's `isNotEmpty` guards deleted those fields for real. `createdAt` could not break the tie because Funnelcake deliberately exposes the original event timestamp.

The second one is the fix for #4175 being defeated by a server-side representation gap. #4175 (and #4021 before it) fixed "REST-hydrated profiles lose every field"; this is the other side of that trade.

### Approach

- Normalize `""` to absent at the REST parser boundary (`UserProfileData.fromJson`) so placeholders never enter `rawData` in the first place.
- Count only *meaningful* keys (non-null, non-empty) when comparing seeds.
- Prefer the raw Kind 0 on a `createdAt` tie, since `requireRawKind0` makes it authoritative by construction — it carries unknown keys and NIP-39 tags the REST projection structurally cannot express.
- Seed from the signing key's cached-then-relay profile when the caller supplies none, turning a destructive full replace into a read-modify-write.

### Observed in the wild

Retrieving all four retained Kind 0 versions for one account with a multi-filter REQ (one `ids` filter per event, which bypasses the relay's per-filter replaceable dedup) shows the fingerprint: versions 1–3 each carry an `about` key; version 4, written by divine-mobile per its `client` tag, dropped `about` and gained `name: ""` and `lud16: ""`.

**Related Issue:** Relates to #4175, #4021, #6423

## Out of Scope

- The own-profile generated-name rendering bug (#6423 / #5851 / #5920) — a separate readiness-gate defect on the read path, not the publish path.
- `about`/`picture`/`banner` are removed when their parameter is *null*, not just empty, while `website` is preserved on null via a different guard shape. Pre-existing and deliberate-looking; left alone here.
- `drivePendingSave` re-reads the seed from Drift even though `ProfileEditorBloc` already holds a live profile it never threads through. Would remove a whole class of null seed; separate change.

## Verification

- `flutter test` in `packages/profile_repository` — 343 pass (340 pre-existing + 3 new).
- `flutter test` in `packages/models` — 829 pass (827 pre-existing + 2 new).
- `flutter test` in `packages/funnelcake_api_client` — 370 pass.
- `flutter analyze lib test integration_test` from `mobile/` — clean.
- `dart format --set-exit-if-changed` — 0 changed.
- App-level `test/blocs/monetization_links_settings`, `test/screens/settings/monetization_links_settings_screen_test.dart`, `test/blocs/profile_editor`, `test/providers` — pass.
- **The new regression tests were confirmed to fail without the production change.** Reverting it gives `Expected: 'alice' Actual: ''` on the placeholder test — precisely the inversion.

### Existing tests changed, and why

Four tests encoded the buggy output. Rather than weaken assertions, three now make their no-seed context explicit through a `stubNoRelayProfile()` helper, so they still test what their names claim (not emitting null values; a failed publish not caching). The fourth asserted that omitting the `nip05` argument **drops** an existing nip05 — that is inverted: `clearNip05` is the explicit removal path, and the method's own doc comment already described preservation as the contract. It now asserts preservation.

One setup comment read `// No cached profile → seed short-circuits, no relay fetch.` — the bug was documented as intended behaviour.

### Unrelated flake noticed

`test/providers/personal_event_cache_service_provider_test.dart` fails roughly 1 run in 4 in a multi-file run. It does so on an untouched `origin/main` too (verified by stashing), so it is not from this branch — but it will cause spurious CI reds and deserves its own issue.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/packages/models/lib/src/user_profile_data.dart
+++ b/mobile/packages/models/lib/src/user_profile_data.dart
@@ -50,19 +50,35 @@ class UserProfileData {
   factory UserProfileData.fromJson(String pubkey, Map<String, dynamic> json) {
     return UserProfileData(
       pubkey: pubkey,
-      name: json['name'] as String?,
-      displayName: json['display_name'] as String?,
-      about: json['about'] as String?,
-      picture: json['picture'] as String?,
-      banner: json['banner'] as String?,
-      website: json['website'] as String?,
-      nip05: json['nip05'] as String?,
-      lud16: json['lud16'] as String?,
+      name: _absentIfEmpty(json['name']),
+      displayName: _absentIfEmpty(json['display_name']),
+      about: _absentIfEmpty(json['about']),
+      picture: _absentIfEmpty(json['picture']),
+      banner: _absentIfEmpty(json['banner']),
+      website: _absentIfEmpty(json['website']),
+      nip05: _absentIfEmpty(json['nip05']),
+      lud16: _absentIfEmpty(json['lud16']),
       createdAt: switch (json['profile_updated']) {
         final String value => DateTime.tryParse(value),
         _ => null,
       },
     );
+  }
+
+  /// Reads a Funnelcake profile string field, mapping `''` to `null`.
+  ///
+  /// Funnelcake's profile response models every metadata field as a
+  /// non-nullable Rust `String`, so a Kind 0 key that simply does not exist is
+  /// serialized as `''`. Keeping that `''` would make an absent field
+  /// indistinguishable from a deliberately-blank one, and downstream that is
+  /// destructive rather than cosmetic: `UserProfile.fromUserProfileFound`
+  /// admits every non-null field into `rawData`, so the REST-derived profile
+  /// out-counts the real Kind 0 and wins `_resolvePublishSeed`'s richness
+  /// comparison — after which the publish path's `isNotEmpty` guards delete the
+  /// fields that were only ever `''` placeholders.
+  static String? _absentIfEmpty(dynamic value) {
+    if (value is! String || value.isEmpty) return null;
+    return value;
   }
 
   final String pubkey;
--- a/mobile/packages/models/test/src/user_profile_result_test.dart
+++ b/mobile/packages/models/test/src/user_profile_result_test.dart
@@ -40,6 +40,43 @@ void main() {
       expect(data.createdAt, isNull);
     });
 
+    test('fromJson treats empty strings as absent', () {
+      // Funnelcake models every profile metadata field as a non-nullable Rust
+      // String, so a Kind 0 key that does not exist is serialized as ''.
+      // Keeping those placeholders let a REST-hydrated profile out-count the
+      // real Kind 0 in the publish-seed comparison, after which the publish
+      // path's isNotEmpty guards deleted the fields for real.
+      final data = UserProfileData.fromJson(_pubkey, const {
+        'name': '',
+        'display_name': 'Alice',
+        'about': '',
+        'picture': '',
+        'banner': '',
+        'nip05': '',
+        'lud16': '',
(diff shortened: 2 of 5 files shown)


hm21 (approved):
Verified the fix, then pushed one takeover commit for a regression it introduced.

**What I pushed** (f270362dbf): the seed comparison let key count override recency in favour of the local seed. `main` was "`fresh` wins if newer **or** richer"; this branch flipped it to "`localSeed` wins if newer **or** richer", so a strictly older local seed could beat the newest relay Kind 0 on key count alone.

That is reachable exactly where it hurts: remove `lud16` (or `lud06`, `bot`, a custom client key, a NIP-39 `i` tag) on divine-web, and the newest Kind 0 becomes both newer *and* sparser than our cached copy — so the next mobile save re-published the deleted field. I ran the scenario against both branches: `main` publishes `{display_name, name}`, this branch published `{display_name, name, lud16, custom_client_key}`.

Recency now decides first; richness only breaks a timestamp tie, where `fresh` still wins on the requireRawKind0 argument you made. Added a regression test that goes red without the change.

**Validation:** `flutter test` in `packages/profile_repository` — 344 pass. `flutter analyze` and `dart format` clean. I also reverted your two production files and confirmed all five of your new tests go red, so the regression coverage is real rather than coverage theatre.

One follow-up for you: please open an issue for the `personal_event_cache_service_provider_test` flake instead of leaving it as a PR-description note — otherwise it just keeps reddening CI for whoever hits it next.

Approving — review the pushed commit and merge when you're happy with it.

realmeylisdev:
Checked the takeover commit — the regression was mine and the fix is right.

Ran your test against `8fec716e` (my pre-takeover commit): red, `Expected: false Actual: <true>`, i.e. the deleted `lud16` re-published. Ran it against `origin/main`'s `_resolvePublishSeed`: green. So recency-first restores what `main` already got right on that path, and the "richer" clause was never load-bearing for it.

The restructure also keeps the #4175 case intact, and lands it on the branch it belongs on: funnelcake derives `profile_updated` from the Kind 0's own `created_at`, so a REST projection and the raw event of the same Kind 0 compare equal on `createdAt` — which is the tie branch, where the meaningful-key count still picks `fresh`. That is what the placeholder test exercises.

Flake issue is #6503. I could not reproduce it today (21 runs across four shapes), so it is filed with what is provable by reading instead: `waitForInitialized` is a bounded poll that gives up silently, so every occurrence surfaces as an unattributable `Expected: true Actual: <false>`, and `dispose()` closes the two fixed-name Hive boxes fire-and-forget, which is the race the tearDown's `pumpEventQueue()` is already guessing at.

Thanks for pushing the fix rather than describing it.

Issue #5244, opened 2026-06-16
feat(auth): export nsec for email/password users


hm21:
### What would you like?

Allow users who signed up with only an email/password account to securely view and export their generated nsec key so they can use it to log in to other Nostr clients.

### How would this be useful for you?

This gives users ownership and portability of their Nostr identity and enables access to third-party clients such as Nos Social.

### When would you use this?

When a user wants to log in to another Nostr client and needs access to the nsec associated with their account.

### Anything else?

Requested on discord [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1516483195451736217)

NotThatKindOfDrLiz:
@dcadenas Can you please explain this one? I believe this is the action that causes a berak in security?

hm21:
Closing — exporting the nsec for an email/password account is shipped.

**#6501** (merged Aug 1) added `KeycastKeyExportCard` to Key Management: for a Divine Login account, **Copy My Private Key (nsec)** opens a password sheet, posts to `POST /api/user/export-key` with the session's bearer token, and copies the returned key. One round trip, no web redirect, no second sign-in. Keycast still enforces the verified-email requirement and the `verified_minor` raw-key-egress refusal before any key material is touched, and divinevideo/keycast#326 added rate limiting and an audit record on that endpoint the same day. The original ask — use the key to log in to other Nostr clients such as Nos Social — is satisfied.

Two things this deliberately does **not** do, recorded here so they are not lost with the issue:

- **The key is never displayed.** The on-screen reveal was removed in review: the key goes from the response straight to the clipboard and the sheet closes, so it never reaches a controller and never rebuilds into the widget tree. A test asserts no widget anywhere holds it.
- **Copy-only leaves a dead end on a device that blocks clipboard writes.** `ClipboardUtils.copyVerified` now reads the value back and reports a refusal instead of showing a success message over an empty clipboard, so the failure is visible — but the owner still has no other route to their key in that case. Recorded as an accepted tradeoff in the class doc.

The security question raised on this issue (@NotThatKindOfDrLiz → @dcadenas) was answered by the design that shipped: password-confirmed, server-gated, copy-only, never rendered.

Not in a release yet: #6501 merged Aug 1, latest release is 1.0.17 (Jul 22).

PR #4444, opened 2026-05-16
chore: Fix local nsec export for Keycast-backed accounts

dcadenas:
## Description

Keycast accounts can be local or remote-only on a device.
Before this change, local Keycast accounts could be blocked from exporting their local nsec, while RPC-only accounts still saw a copy button that could only fail.

This change makes export depend on the active key container.
If the matching local private key exists, the user can copy the nsec.
If the account only signs through Keycast RPC, the app hides the copy action and explains that there is no local nsec to copy.

The remote-signing message is localized across all non-English ARBs, and the l10n test now catches this specific fallback mistake.

screenshot-2026-05-16_20-08-30

**Related Issue:** Relates to #747

## Out of Scope

This does not add remote Keycast private-key export.
RPC-only Keycast keys stay off-device by design.

## Verification

The branch was rebased on latest main through HTTPS because SSH auth is unavailable in this environment.
I ran the full mobile test suite, analyzer, l10n consistency coverage, and the changed-string scan.

- `git fetch https://github.com/divinevideo/divine-mobile.git main && git rebase FETCH_HEAD`
- `mise exec -- flutter test`
- `mise exec -- flutter analyze`
- `mise exec -- flutter test --no-pub test/l10n/arb_consistency_test.dart`
- `mise exec -- flutter test --no-pub test/screens/key_management_screen_test.dart`
- `python3 .claude/skills/check-l10n/scan_strings.py`

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [x] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

Issue #855, opened 2026-01-08
Exporting nsec after account created with "Login with diVine"


JohnWeidner:
After an account has been created using the "Login with diVine" button, the user's nsec is kept on the login.divine.video servers and is not available for the app to display. The user will only be able to backup/copy their nsec key from the divine website. The existing flutter app has a Settings / Key Management page that provides a button labeled "Copy My Private Key (nsec)".  What should happen when the user taps that in this scenario?

hm21:
Answered and implemented — closing.

The question here was what should happen when a Divine Login account taps **Copy My Private Key (nsec)**. The shipped answer is: the button stays, asks for the account password in a sheet, fetches the Keycast-held key via `POST /api/user/export-key` with the session's bearer token, and copies it. `KeycastKeyExportCard` renders for Divine Login accounts; the local-key path is untouched and still copies straight from the device (`key_management_screen.dart:176-191`). That is **#6501**, merged Aug 1, with divinevideo/keycast#326 (rate limit + audit on the export endpoint) landing the same day.

Worth recording that this reverses the January answer in this thread — "people using diVine Login shouldn't be seeing a copy my nsec" / "it's a keycast feature, not a divine-mobile issue". #6501 called that out explicitly rather than letting it land quietly: product asked for the native flow on UX grounds, because the web hand-off cost the user two authentications to copy one key. Allowing the export is the intended behavior.

**Still open elsewhere, deliberately:** #5244 asks to *view* as well as export. #6501 delivered the export and removed the on-screen reveal in review — the key goes from the response straight to the clipboard, never reaches a controller, never rebuilds into the widget tree, and a test asserts no widget holds it. The consequence @hm21 raised — a device that blocks clipboard writes leaves the owner with no route to their key — is now visible rather than silent (the write is verified and a refusal is reported), but not solved. That reveal-vs-copy decision belongs to #5244.

Not in a release yet: #6501 merged Aug 1, latest release is 1.0.17 (Jul 22).

Issue #182, opened 2025-11-17
How do I upload on web?


WiiUFan23:
I don't know how to make videos on web, please help.

PR #6501, opened 2026-07-30
feat(auth): copy the Keycast-held nsec from Key Management, without a web redirect

realmeylisdev:
## Description

Key Management now reaches the signing key Keycast holds for a Divine Login account **without leaving the app**. The card asks for the account password in a `VineBottomSheet`, posts it to `POST /api/user/export-key` with the session's bearer token, and copies the returned nsec.

An account created with email/password has its key held by Keycast, and NIP-46 defines no RPC that returns key material (`connect`, `sign_event`, `ping`, `get_public_key`, `nip04_*`, `nip44_*`, `switch_relays`, `logout` — none of them), so the app cannot get it from the signer. Key Management used to say exactly that and stop:

> "This account signs with Keycast. No private key is stored on this device, so there is no nsec to copy here."

**This replaces the web hand-off that was here in the first revision of this PR.** That version opened `login.divine.video/settings/security` with `LaunchMode.externalApplication`, and because `/settings/security` sits outside both app-link associations there was no shared session — so the user signed in again in the browser *and then* confirmed their password. Two authentications to copy one key, and the app left behind.

The endpoint never required any of that. Keycast's `extract_user_from_token` tries `Authorization` before the browser cookie, so a native call with the session's bearer token is served identically; the `auth_cors` layer on the route is a browser mechanism and does not constrain a Dart client. The app already calls sibling `/api/user/*` routes this way (`getAccountStatus`, `deleteAccount`). The web flow's separate `/api/user/verify-password` step only gates its own UI — `export_key` verifies the password itself — so the native path is **one** call.

Thanks to the review feedback that pushed back on the redirect — it was right, and the constraint I thought was in the way was not there.

## Behavior

- Tapping **Copy My Private Key (nsec)** on a Keycast-held account opens a password sheet; the local-key path is untouched and still copies straight from the device.
- A wrong password stays **inline in the sheet** — the field keeps its place and the user retries without reopening anything. Editing the field clears the stale error.
- Every other refusal closes the sheet and reports something actionable: session expired → sign in again, email unverified → verify first, policy denial → keys are managed by Divine.
- Keycast answers `401` for both a wrong password and a stale token, and its prose is not a stable discriminator. The client re-probes an endpoint that needs only the token: if the token still authenticates, the password was what failed. So "wrong password" and "sign in again" are never confused.
- The session is refreshed before the call, so an expired-but-refreshable session does not surface as a sign-in prompt.

## Security posture

- **No gate moved on-device.** Keycast still enforces the verified-email requirement and the `verified_minor` raw-key-egress refusal (`support-trust-safety#188`), both before any key material is touched. The #182 affordance removal still hides the card for a protected minor, and the card re-reads that gate before opening the sheet.
- **Nothing is stored and nothing is shown.** The key goes from the response straight to the clipboard and the sheet closes, and the clipboard write is now confirmed rather than assumed (round 4). It never reaches a controller and never rebuilds into the widget tree, so no frame of the flow holds it to screenshot or record. It is never written to disk, cached, or logged — including on failure paths, which are covered by a test that feeds a server wrongly echoing key material alongside an error.
- **Failures are typed.** `ExportKeyFailure` carries the reason so the UI branches on the enum, never on server prose that varies by deployment. The password lives in the sheet's controller for the life of the sheet.

## Design decision reviewers should weigh

This **reverses the stated design in #4444** ("RPC-only Keycast keys stay off-device by design"), and @NotThatKindOfDrLiz's in-app-reveal question on #5244 is still open. Product asked for the native flow on UX grounds — the double sign-in — and that is the change here. Calling it out explicitly rather than letting it land quietly: if the #4444 position should hold, this PR is the thing to reject, not the copy.

A reveal step briefly existed here and has been removed (round 3 below): the flow is copy-only and the key is never rendered. This path is therefore no wider than the local-key path, which also only ever copies.

## Out of Scope

- **Biometric or device-credential confirmation.** The `biometricPrompt` parameter threaded into `exportNsec()` is inert — its native `openvine.secure_storage` handler does not exist in this repo — and there is no `local_auth` dependency. The account password is the confirmation here.
- **Keycast hardening, worth its own issue.** `POST /user/export-key` has no rate limit and writes no audit record; the client-side attempt cap added in review is friction, not a substitute. There is also an authorization question about how this route is gated relative to the other destructive account route; raising that one on the private tracker rather than here.
- `find.text('Key Management')` in the touched integration test stays hardcoded. It does not correspond to `keyManagementTitle` ("Nostr Keys"), so converting it would change what the assertion means — left for whoever can run that suite.
- The branch name still says `keycast-handoff`; left alone rather than disturbing an open PR.

## Notes

- Copy went into `app_en.arb` and all 17 other locales, genuinely translated — `arb_consistency_test` fails on English fallback for these keys, and its key list is updated. The now-unused `keyManagementKeycastOpenWeb` is removed from all 18.
- The button reuses the existing `keyManagementCopyNsec` string, so both export paths read the same to the user.
- Feedback uses `ClipboardUtils.copy` / `DivineSnackbarContainer`, the password field is `DivineAuthTextField`, and the sheet is `VineBottomSheet` with `DivineButton` actions — the raw-dialog and Material-button ceilings both stay flat.
- The export flow moved to `lib/screens/key_management/keycast_key_export_card.dart`, which puts `key_management_screen.dart` back under the advisory 800-line ceiling (704 → 634).

**Related Issue:** Relates to #855, Relates to #5244

## Verification

- `flutter test test/screens/key_management_screen_test.dart` — 18/18, covering the sheet, the clipboard write, the inline wrong-password retry, the attempt cap, the localized transport-failure copy, an unrecoverable refusal, and the empty-field guard
- `flutter test test/services/auth_service_export_keycast_nsec_test.dart` — 6/6 (new)
- `flutter test` in `packages/keycast_flutter` — 245/245, including 10 tests for the endpoint contract and every refusal mapping
- `flutter test test/l10n/arb_consistency_test.dart` — 5/5
- All five design-system ceilings pass; the `raw_icons` and `raw_textstyle` baselines from the first revision are still exact (2 and 15) and needed no further edit
- `dart format` clean on every changed file

Not run: the integration test, which needs a device and a live Keycast; its assertions were updated for the new behavior but not executed.

## Review fixes (pushed on top)

- Capped the password retries at five. Keycast does not rate-limit `export-key` and keeps no audit record, so the inline retry — deliberately one keystroke — was an unlimited guessing loop paying out the nsec. Reopening the sheet resets it; the durable fix is server-side.
- Stopped interpolating `ExportKeyResult.error` into the failure snackbar. It is untranslated web-UI prose, and for a transport failure it was the raw exception, so a dropped connection printed `Network error: ClientException: ..., uri=...` inside a translated sentence. The mapper now takes the enum.
- Fixed the 401 disambiguation. `getAccountStatus` returns `null` on a timeout or 5xx as well as on a real 401, so a wrong password over a flaky link reported "session expired, sign in again". The probe now keeps a third answer for "could not tell".
- Pinned the export format to `nsec`. Keycast 400s every other value, so the parameter only documented a capability the server does not have.
- Split the two sheet steps into widget classes, and pinned `keyManagementYourPrivateKeyLabel` in `arb_consistency_test` — the one key in this group nothing stopped from falling back to English. (That key is deleted in round 3, along with the step it labelled.)

## Review fixes — round 2 (@dcadenas)

- **The key export session is now bound to the account that owns it.** It refreshed with no `userPubkey`, so a session minted while the screen was open was persisted unbound — which `SignerSecureStore` declines to archive and deletes as corrupt on the next restore. It also bypassed `OAuthSessionCoordinator`'s single-flight guard over a single-use refresh token, and never checked that the globally-stored session belonged to the signed-in account. A new `AuthService.activeAccountKeycastToken()` is the one place a `/api/user/*` call may get a token.
- **`deleteKeycastAccount` and the protected-minor check go through that same gate.** Both had the identical bypass and both predate this branch. Deletion is irreversible and was taking whatever the global session slot authenticated as. Note the consequence on the minor seam: it fails closed, so a session the gate refuses now reads as unknown and restricts DMs and key management rather than answering for an account it may not describe.
- **A missing key no longer reads as an unreachable service.** HTTP 404 maps to `ExportKeyFailure.noKey`, which shared the generic transport copy — so a definitive "no key on record" invited a retry that can never succeed. `ExportKeyFailure.server` deliberately keeps the generic copy: a 5xx and a dropped connection leave the user the same remedy.
- **An unreadable account body is no longer a policy denial.** A 200 the parser could not read left the probe with a null status, which the 403 classifier folded in with the `verified_minor` case — telling a user Divine manages their keys on a response carrying no account state at all. It lands on `unknown` now, alongside a probe that never answered. Confined to the classifier, so `getAccountStatus` and the #175/#176 gates are unmoved.

Round-2 verification: `flutter analyze lib test integration_test` clean; 4173 tests green across `test/services`, `test/providers`, `test/repositories` and `delete_account_dialog_test`; 249 in `packages/keycast_flutter`; `arb_consistency_test` green. The new copy is translated in all 17 non-English locales and pinned.

## Review fixes — round 3 (@NotThatKindOfDrLiz)

- **The on-screen reveal step is gone; the export is copy-only.** Confirming the password sends the key straight from the response to the clipboard and closes the sheet. `_RevealStep`, the key `TextEditingController`, and the two-step state machine are all removed, so the key never reaches a controller and never rebuilds into the widget tree.
- **The confirm button promises the copy it now makes**, reusing the already-translated `keyManagementKeycastCopyKey` rather than rewording a fetch label across 17 locales. `keyManagementKeycastFetchKey` and `keyManagementYourPrivateKeyLabel` are deleted from all 18 ARBs — the latter's description was a spec for the removed surface.
- **The "never share your nsec" warning moves into the password step.** It only existed on the reveal step; without the move the custodial path would be the one nsec handover in the app with no warning at all. The local-key path warns beside its copy button, before the key moves, and confirming here is that same moment.
- **The copy is wrapped so the sheet closes either way.** Both actions are disabled while submitting, so a clipboard write that throws would have stranded the sheet with nothing left to tap — the reveal step's Close button used to cover that case.
- **The stale `ExportKeyResult.error` comment is corrected.** It claimed the server's prose "stays in the logs"; it never did — `AuthService.exportKeycastNsec` interpolates `result.failure` alone. The retry-cap doc had the mirror-image problem: it asserts Keycast has no rate limit or audit record for `export-key`, which keycast#325 removes, so it would have been stale on the day this merged. `error` itself stays on the package result type, where returning the server's message is reasonable for a client library.

The accepted tradeoff, recorded in the class doc so it reads as a decision: a device that blocks clipboard writes now leaves the owner with no route to their key.

The reveal test becomes a copy-only test asserting that no widget anywhere holds the key. `find.text` matches an `EditableText` by its controller's contents whether or not the field obscures them — verified directly against `DivineAuthTextField` — so a reinstated read-only key field fails that guard even while hidden.

Round-3 verification: `flutter analyze lib test integration_test` clean; 122 tests green across `test/screens/key_management_screen_test.dart`, `test/l10n`, `test/services/auth_service_export_keycast_nsec_test.dart`, `test/services/auth_service_delete_keycast_account_test.dart` and `test/providers/protected_minor_providers_test.dart`; 249 in `packages/keycast_flutter`. All nine design-system and structural ratchets pass with baselines untouched.

**Merge gate:** holding until divinevideo/keycast#326 merges, per the review note.

## Review fixes — round 4 (@hm21)

@hm21 flagged that clipboard copies fail on some Android devices, leaving copy-only users with no alternative. Tracing it through the engine, the failure was worse than losing the copy — **we claimed success either way**:

- `PlatformPlugin.setClipboardData` calls `clipboard.setPrimaryClip(clip)` with no try/catch and no result check (`setPrimaryClip` returns `void`), and `PlatformChannel` then answers `result.success(null)` unconditionally.
- iOS is the same shape: `FlutterPlatformPlugin.setClipboardData:` assigns `pasteboard.string`, then `result(nil)`.

So `Clipboard.setData` resolves successfully for every failure that does not throw. For this flow that meant a green "Private key copied to clipboard!" over an empty clipboard, for the one hand-over the user cannot re-derive.

- **`ClipboardUtils.copyVerified` reads the value back and compares it** — the only confirmation either platform offers. Reading our own clip does not trip Android 12's read toast or the iOS paste banner (both fire on another app's content). It is additive rather than a change to `copy`, so the six existing call sites keep their behaviour.
- **A refused write now says so** in all 18 locales instead of staying quiet. The sheet still closes: the key is deliberately not retained, so retrying means fetching again, and a successful export does not count against the keycast#325 lockout, which counts wrong passwords.
- **This does not answer the concern hm21 raised.** Copy-only still leaves a blocked device with no route to the key; this makes that dead end visible rather than silent. The reveal-vs-copy decision is open for @NotThatKindOfDrLiz and @dcadenas.

### Rebased onto main

Rebased to clear a conflict with #6467 (semantic colour tokens), which had migrated `_buildKeycastRemoteSigningInfo` in `key_management_screen.dart` — the very method this branch deletes. Resolving in favour of the replacement card would have **quietly reverted #6467 for that surface**, because the conflict was in the deleted method rather than in the replacement, and nothing flags it: there is no ratchet over the legacy colour statics, and the raw-colour ceiling counts literals, which these are not. The card now follows #6467's split — brand colours (`vineGreen`, `error`) stay as `VineTheme` statics, surface-dependent text uses `context.vineColors`.

Round-4 verification: `flutter analyze lib test integration_test` clean; 126 tests green across `key_management_screen_test.dart`, `test/l10n`, the new `test/utils/clipboard_utils_test.dart`, and the auth/provider suites; all 12 design-system and structural ratchets pass. The new copy is translated in all 17 non-English locales and pinned in `arb_consistency_test`.

## Round 5 — rebased onto main, and the four new locales

Rebased onto `origin/main`, which had moved 15 commits since round 4. That turned two CI jobs red, both for the same reason: #6495 added Vietnamese, Urdu, Chinese, and Malay while this branch was open, so the eleven strings this flow adds to `app_en.arb` had no translation in the four new locales. `arb_consistency_test` failed on `app_ms.arb`, and `flutter gen-l10n` left `app_localizations_{ms,ur,vi,zh}.dart` stale.

All eleven are translated in the four new locales now. One of them needed more than a fill-in:

- **`keyManagementKeycastRemoteSigning` was already there, translating the copy this branch replaces.** The four new locales carried a rendering of the old dead-end sentence — "no private key is stored on this device, so there is no nsec to copy here" — sitting above a button that now copies exactly that key. Nothing would have caught it. The consistency test only asks that a locale differ from English, which a stale translation does, and the key was never missing, so it never appeared in the diff CI reported. Retranslated against the new English.

The rebase itself was clean and carried all 19 commits forward — `git cherry` reports no commit from the pre-rebase tip without an equivalent in the new one, so nothing that was already approved was dropped.

Round-5 verification: `flutter analyze lib test integration_test` clean; 141 tests green across `test/l10n`, `key_management_screen_test.dart`, `clipboard_utils_test.dart` and the auth/provider suites; 249 in `packages/keycast_flutter`; all 12 design-system and structural ratchets pass with baselines untouched; `dart format` clean.

**Merge gate unchanged:** still holding until divinevideo/keycast#326 merges.

## Type of Change


realmeylisdev:
Not your bad at all — the URL was the right pointer. The redirect was my call, and it came from a design
constraint rather than a technical one. Having checked the constraint, you're right, and it's now implemented
that way: `1661bdbbf`.

## The direct fetch works

`POST /api/user/export-key` authenticates from headers and tries **Bearer before the cookie**, so a native call
with the session's access token is served exactly like the web one. The CORS layer on that route is a browser
mechanism and does not constrain a Dart client. We already do this shape twice — `getAccountStatus(token)` against
`/api/user/account`, and `deleteAccount(accessToken)` — so `exportKey(token, password)` is the same pattern
against a different path.

**And it's one call, not two.** The web flow hits `/api/user/verify-password` first only to gate its own UI;
`export_key` verifies the password itself, along with the verified-email requirement and the custodial-minor
refusal. So the T&S gates stay server-side and going native does not weaken them.

## Why you hit two logins

Worth naming, because it makes your point stronger than it first looks: the button used
`LaunchMode.externalApplication`, so it was the *system* browser rather than an in-app view, and
`/settings/security` sits outside both app-link associations — no shared session and no way back into the app.
So it was a browser sign-in, then a password confirm. Native collapses that to one password prompt without
leaving Key Management.

## What it looks like now

Tapping **Copy My Private Key (nsec)** opens a `VineBottomSheet` with a `DivineAuthTextField` for the password;
confirming copies the key and shows the standard Divine snackbar. A wrong password stays inline in the sheet so
the retry costs one keystroke. Session expiry, unverified email, and a policy refusal each get their own message
rather than a generic failure.

Two things I'd flag for you specifically:

1. **It's a copy, not an on-screen reveal** — no key is rendered. Say the word if you wanted the reveal.
2. **This reverses #4444's "RPC-only Keycast keys stay off-device by design"**, and Liz's question on #5244 is
   still open. I've called that out at the top of the PR body rather than letting it land quietly, so it gets
   decided on the merits.

No biometric confirmation: the `biometricPrompt` parameter that exists on the local export path is inert (its
native handler isn't in this repo) and there's no `local_auth` dependency, so the account password is the
confirmation. Happy to add device-credential confirmation as a follow-up if you'd rather have it.

hm21:
@realmeylisdev Thanks for the update. I like this solution much more now.

After testing it on a device, I noticed the following:

1. When I enter the password, the sheet stays behind the keyboard. It should move above the keyboard instead.
2. The text “Your key is kept...” is very close to the bottom sheet divider. A little more spacing would look better.
3. This is not necessarily an issue, but I was wondering whether it might be better to handle it like in the web app. Instead of copying the key directly to the clipboard, we could show it to the user and give them the option to copy or view it. By default, it could be hidden like a password. Most users will probably copy it anyway, but there might be cases where clipboard access is blocked. In that situation, the user would otherwise have no way to see the key.

I posted a short video below showing how it looks on my side.


https://github.com/user-attachments/assets/7ea78997-053d-4404-9d29-89cc7392b38a

realmeylisdev:
@hm21 All three done in `4478b2a1b`.

1. Sheet now lifts above the keyboard — fixed mode was deriving `isScrollControlled` from `expanded: false`, so it could neither grow nor react to the inset. Pinned by a test that raises a 900px keyboard and asserts the field and both actions stay visible.
2. Spacing was a symptom: `contentTitle` is only read on the `children` path, so passing `body` dropped the title silently. It goes through the header now, which brings the spacing with it.
3. Took your suggestion — the key is shown in a read-only field, hidden behind its reveal control, with the never-share warning. Nothing hits the clipboard until you tap Copy, so a blocked clipboard is no longer a dead end.

Note it does mean the key is rendered now, which is the thing #4444 ruled out — flagged at the top of the PR body so it gets decided rather than assumed.

hm21 (approved):
Verified the endpoint contract against keycast rather than taking it on trust: `POST /api/user/export-key`, `{password, format}` → `{key}`, and the 401/403/404/5xx mapping all line up with `AuthError`'s status mapping (`api/src/api/http/auth.rs:4046`, `routes.rs:157`). Both server-side gates stay server-side — `refuse_key_egress_for_verified_minor` runs before the password is even read, and `email_verified` is required. The client moves neither.

Pushed five fixes as three commits. Please review and merge when you're happy with them.

- **Capped the password retries at five** (9da3ed4). `key_export_routes` carries only `auth_cors` — there is no rate-limit layer anywhere in the keycast router, and `export_key` writes no audit record. The inline retry was deliberately designed so a wrong password costs one keystroke instead of a reopen, which made the sheet an unlimited guessing loop against an endpoint whose success payout is the account's nsec. Five wrong passwords now disable the field and the submit action until the sheet is dismissed. Reopening resets it, so this is friction rather than a lockout — the durable fix is server-side, and the Keycast issue is worth filing before this ships widely.
- **Stopped raw exception text reaching users** (9da3ed4). `_failureMessage` interpolated `ExportKeyResult.error`, which is untranslated web-UI prose, and for a transport failure is the literal `'Network error: $e'` — so a dropped connection showed `Network error: ClientException: ..., uri=https://login.divine.video/api/user/export-key` inside a translated sentence. It now takes the enum, which makes the prose structurally unable to reach the screen, and keeps it in the logs.
- **Fixed the 401 disambiguation** (c549295). `getAccountStatus` returns `null` for a timeout or a 5xx exactly as it does for a real 401, so a wrong password over a flaky link told the user their session had expired. The probe now keeps a third answer for "could not tell" and falls back to `unknown`.
- **Pinned the export format to `nsec`** (c549295). Keycast 400s every other value (`auth.rs:4106`), which landed in `ExportKeyFailure.unknown`, so the parameter and its test documented a capability the server does not have.
- **Split the sheet steps into widget classes** (9da3ed4), and pinned `keyManagementYourPrivateKeyLabel` in `arb_consistency_test` (f6dc947) — it arrived with this feature like the rest of the group but was the one key nothing stopped from silently falling back to English. The new attempt-cap copy is translated in all 17 non-English locales.

**I also corrected the description**, which is the part worth your attention. It said "this is a copy-to-clipboard, not an on-screen reveal, so no key is rendered" — the last commit added the reveal step, and `DivineAuthTextField` with `obscureText: true` ships a visibility toggle, so the nsec is rendered on request. That sentence was load-bearing for the #4444 / #5244 question this PR explicitly asks reviewers to weigh, so it could not stay.

**What I deliberately did not touch:** the reveal itself. The local-key path (`key_management_screen.dart:497`) only ever copies and never renders, so the custodial path is now the wider affordance of the two — which is the opposite of what you would expect. That is a product call for @NotThatKindOfDrLiz and #4444, not something a reviewer should push, so it stays open.

All 16 checks are green on 9da3ed4. Locally: `flutter analyze lib test integration_test` clean, 18/18 in `key_management_screen_test`, 6/6 in `auth_service_export_keycast_nsec_test`, 245/245 in `keycast_flutter`, `arb_consistency_test` green. The integration test still needs a device and a live Keycast and has not been run.

realmeylisdev:
Good catch on the description — the "not an on-screen reveal, so no key is rendered" line was wrong, and it was load-bearing for the #4444 / #5244 question this PR asks reviewers to weigh, so it needed to go.

Agreed on leaving the reveal itself alone. The custodial path ending up with the wider affordance than the local-key path is backwards, but that's @NotThatKindOfDrLiz's call on #4444, not one to settle in review. Going through your three commits next 🚀

NotThatKindOfDrLiz:
I think this PR needs to be reviewed against the prior Keycast/BYOK security discussion before it merges.

The relevant history is #3359, #3784, #4665, and divinevideo/keycast#242. My read of that discussion is that the product/security rule was not simply "an nsec must never cross the network." Rabble corrected that framing on #4665: for custodial Keycast/BYOK flows, transferring key material to Keycast can be the intended product contract. The security issue was accidental propagation through the wrong surfaces: PKCE verifier, OAuth/token plumbing, pending verification storage, logs, generic request handling, etc.

That makes this PR directionally plausible, but it also means it should be treated as an intentional raw-key-egress feature, not a routine mobile UX improvement. `POST /api/user/export-key` needs the same level of product/security clarity as the prior BYOK work: explicit eligibility rules, server-side refusal semantics, rate limiting, audit logging, and request/error redaction. The client-side retry cap is not a substitute for that.

A few implementation details also depend on that framing:

- The 403 handling should not branch on English server prose; it should use structured account status or a structured server response.
- The UI copy should be honest about the two-step flow. The first password step should not say "Copy key" if it only fetches and advances to a reveal screen.
- The reveal step is a wider affordance than the local-key path, so that should be an explicit product decision rather than something we inherit accidentally from the review thread.
- If Keycast later rate-limits this endpoint, the client needs a typed 429/rate-limit path instead of falling through to generic failure copy.

I do not think this should merge only on peer approval while the earlier #5244 question and the BYOK/custody precedent are still unresolved in relation to this exact behavior.

@dcadenas and @rabble, can you please review this PR in the context of #3359/#3784/#4665/keycast#242 and provide direction on what should happen here? Concretely: should we ship native Keycast-held nsec export from Key Management, should the reveal affordance stay, and what server-side hardening must exist or be linked before merge?

realmeylisdev:
Three of the four are fixed:

- **403 no longer reads server prose** (`3511b3e`). Both overloaded codes now resolve from `GET /user/account`'s structured fields, and `verified_minor` is checked ahead of `email_verified` to mirror the server's own order — otherwise a minor with an unverified email gets told to verify it, which promises an export that verifying can never unlock.
- **The password step no longer says "Copy key"** (`a9a74c4`). It reads "Get my key"; only the reveal step copies. Every locale updated.
- **429 gets its own typed path and copy** (`3511b3e`) instead of falling through to "could not be reached".

The reveal affordance and the custodial framing I've left alone — those are the product calls you're asking for, not something to settle in the diff. Agreed on the framing itself; one thing worth naming is that #3359/#4665/divinevideo/keycast#242 are all *inbound* custody (device → Keycast), and this is *outbound*, with no equivalent hardening on that direction today.

On the server side: `export_key` has no rate limit or lockout, writes no audit row on success, and its two 403s carry no machine-readable code — filed as divinevideo/keycast#325, with the evidence against `origin/main`. Treat it as the merge blocker if that's the shape you want.

Not merging this on peer approval.

@dcadenas @rabble — the three questions Liz put are the ones I need answered.

realmeylisdev:
The server-side blocker now has a PR: divinevideo/keycast#326.

All three gaps are closed. The lockout is five wrong passwords in 15 minutes, counted out of `auth_events` in Postgres rather than a per-process cache — keycast runs `min-instances=3`, so a replica-local limiter would have handed out three budgets instead of one. Every attempt writes a row now, including the success, which previously wrote nothing at all. And both 403s carry codes (`EMAIL_NOT_VERIFIED`, `KEY_EGRESS_DENIED`), with the messages unchanged byte for byte, so it is additive for anything already reading prose. Once it lands, the `GET /user/account` round trip this PR added to tell those two apart can come out.

It covers `change-key` as well as `export-key`, which the issue did not ask for: they share the minor gate, take the same password, and both end custodial control of the key, so a lockout on one alone is bypassed by pointing at the other.

Worth naming that the budget is per account, not per IP — nothing in keycast extracts a client IP today, and behind Cloud Run the `X-Forwarded-For` a request arrives with is written by the client, so an IP key would be attacker-rotatable.

That leaves the product call as the only open item on this PR.

dcadenas (changes requested):
I reviewed the Keycast side separately, and it looks good to me. The context is in [this Slack thread](https://planetary-app.slack.com/archives/C0B0J26377W/p1785466097899769?thread_ts=1785387746.529379&cid=C0B0J26377W).

From the mobile side, I still see three issues on `a9a74c4e`.

1. **Session ownership is lost during refresh**

`mobile/lib/services/auth_service.dart:2979`: `exportKeycastNsec()` refreshes the session directly without binding it to the active account or using the shared refresh coordinator. An expired session can be saved without a `userPubkey`, and an already valid session is not checked against the active account before its token is used. This breaks the ownership checks used when accounts are archived and restored. Please use the owner-aware refresh path and reject a session whose `userPubkey` is missing or does not match the active account.

2. **No key is reported as a connection failure**

`mobile/lib/screens/key_management/keycast_key_export_card.dart:112`: `ExportKeyFailure.noKey` uses the same message as network and server failures, even though the client maps HTTP 404 to `noKey` at `mobile/packages/keycast_flutter/lib/src/oauth/oauth_client.dart:1175`. A definitive response that no exportable key exists should not tell the user the service could not be reached. Please give `noKey` its own message and widget test.

3. **A malformed account response becomes a policy denial**

`mobile/packages/keycast_flutter/lib/src/oauth/oauth_client.dart:1074-1082, 1093-1114`: when `GET /user/account` returns 200 but its body cannot be parsed, the probe reports an authenticated account with no status. The 403 classifier then treats the missing status as `denied`. That can show the policy message for an email-unverified account even though the response proves neither cause. Please return `unknown` when the account status is missing and add a malformed-body regression test.

realmeylisdev:
All three fixed, and thanks for checking the Keycast side.

**1. Session ownership** — a2b61e6. `exportKeycastNsec` takes its token from a new `AuthService.activeAccountKeycastToken()`: refresh goes through `OAuthSessionCoordinator` with `expectedOwnerPubkey` set, and a session whose `userPubkey` is missing or is not `currentPublicKeyHex` is refused. Strict equality, matching the checks already in `signInForAccount` and the app-resume path.

I widened this past what you asked, so flag it if you disagree. `deleteKeycastAccount` and `protectedMinorRepositoryProvider` had the identical bypass — both pre-existing on `main` — and now go through the same gate (d9390d9). Deletion was the one that worried me: irreversible, and taking whatever the global session slot authenticated as. The behaviour change worth knowing about is that the protected-minor seam fails closed, so a refused session reads as unknown and restricts DMs and key management rather than answering the minor question for an account it may not describe. Sign-in binds the pubkey and a2b61e6 binds every refresh, so unbound sessions should be rare — and d9390d9 reverts on its own if you would rather that landed separately.

**2. `noKey`** — bc5553c8, copy in ae358554. Its own string in all 18 locales, with a widget test asserting that neither the generic "could not be reached" sentence nor the server's own prose reaches the screen. I left `ExportKeyFailure.server` on the generic arm, since a 5xx and an unreachable host leave the user the same remedy — happy to split that one too if you read it differently.

**3. Malformed account body** — 03eda52e. `_classifyRefusal` returns `unknown` for a null status instead of folding it into `denied`, with a regression test over three unreadable 200 bodies (non-JSON, a JSON array, a JSON string). Kept inside the classifier on purpose: changing `_probeAccount` would move `getAccountStatus` and the #175/#176 gates with it, which is not this PR's business.

Verified locally — `flutter analyze lib test integration_test` clean, 4173 tests green across `test/services`, `test/providers`, `test/repositories` and `delete_account_dialog_test`, 249 in `keycast_flutter`, and `arb_consistency_test` green.

NotThatKindOfDrLiz:
Thanks for the careful work here. My call for now:

- Do not merge this before divinevideo/keycast#326 lands. The mobile retry cap resets on sheet reopen, so server-side lockout/audit is required before native raw-key export ships.
- I’m not comfortable shipping the on-screen reveal step in this PR. Please remove `_RevealStep` and keep this copy-only unless we separately decide to build a capture-protected reveal surface.
- Please clean up the stale `ExportKeyResult.error` comment.
- @dcadenas can you re-review the fixes to your three items and explicitly accept or reject the `d9390d9` ownership-gate widening?

Once #326 is merged and the reveal step is removed, I’m comfortable with the native copy flow.

realmeylisdev:
All three done, pushed in 5525860 + d981630.

- **Reveal step removed — the flow is copy-only.** Confirming the password copies the key straight from the response and closes the sheet; it never reaches a controller and never rebuilds into the widget tree. `keyManagementYourPrivateKeyLabel` is deleted from all 18 ARBs rather than left sitting there, since its description was a spec for the surface you're rejecting. The "never share your nsec" warning moved into the password step — it only existed on the reveal step, and without the move the custodial path would be the one nsec handover in the app with no warning at all.
- **Stale `ExportKeyResult.error` comment cleaned.** It claimed the server's prose "stays in the logs"; it never did — only the enum is logged. I fixed the retry-cap comment in the same commit for the mirror-image reason: it asserts Keycast has no rate limit or audit record for `export-key`, which would be false the day #326 lands.
- **Holding the merge until divinevideo/keycast#326 lands.**

One consequence worth naming, since it was the reveal step's whole justification: a device or policy that blocks clipboard writes now leaves the owner with no route to their key. That's recorded in the class doc so it reads as a decision rather than an oversight, and it's the thing a capture-protected reveal surface would eventually fix.

@dcadenas is already re-requested for the round-2 fixes and the `d9390d9` ownership-gate widening.

realmeylisdev:
The Android experience checks out, and it's worse than losing the copy — I traced it through the engine and we were **claiming success either way**.

`PlatformPlugin.setClipboardData` calls `clipboard.setPrimaryClip(clip)` with no try/catch and no result check (`setPrimaryClip` returns `void`), and `PlatformChannel` then answers `result.success(null)` unconditionally. iOS is the same shape — `FlutterPlatformPlugin.setClipboardData:` assigns `pasteboard.string`, then `result(nil)`. So `Clipboard.setData` resolves successfully for every failure that doesn't throw, and Android documents ignoring clipboard writes from apps without focus on top of OEM and enterprise policies that block them outright. A throwing failure does reach Dart, but a blocked one doesn't.

For this flow that meant a green "Private key copied to clipboard!" over an empty clipboard — for the one hand-over the user can't re-derive. They'd walk away believing they had a backup they didn't have.

Fixed in `2a66b69`: `ClipboardUtils.copyVerified` reads the value back and compares, which is the only confirmation either platform offers. Reading our own clip doesn't trip Android 12's read toast or the iOS paste banner — both fire on another app's content. It's additive rather than a change to `copy`, so the six existing call sites keep their behaviour.

**But that doesn't answer what you actually raised.** It makes the dead end visible instead of silent; it still leaves a blocked device with no way to reach the key. Your point stands.

Two things I dug up that I think matter for the decision:

**Your screenshot is keycast's own web UI**, not divine-web — `keycast/web/src/routes/settings/security/+page.svelte`, `showExportedNsec = $state(false)` starting hidden, eye button toggling `type={showExportedNsec ? 'text' : 'password'}`. So exactly the pattern you describe, and your consistency argument is fair. It just means "should the web change too" is one decision spanning two repos, which is Liz's call rather than something this PR can settle.

**A capture-protected reveal is net-new work, and it won't land evenly.** divine-mobile has no capture protection today — no `FLAG_SECURE`, and no `flutter_windowmanager`/`no_screenshot`/`screen_protector` in `pubspec.yaml`. Android's `FLAG_SECURE` genuinely blocks screenshots, screen recording, and the app-switcher thumbnail; iOS has no equivalent for screenshots. So "hidden behind a reveal" and "safe from capture" aren't the same property, and only one platform can actually deliver the second one.

On the sign-in password point — fair, though I'd separate them: revealing a password shows the user something they already know, whereas revealing the nsec materialises something they don't have. The attacker-who-has-the-password path is what keycast#325's lockout and audit trail narrow, rather than the reveal control.

I've deliberately not re-added the reveal — that's @NotThatKindOfDrLiz's call and @dcadenas is being asked to weigh in. Happy to implement whichever way it lands: reveal behind `FLAG_SECURE` on Android, a share-sheet or file fallback, or leaving it copy-only.

dcadenas (approved):
Approving. I pushed one commit rather than sending this back as a request, since the fix was small and clear.

**Your three earlier items are all fixed on `8e091e8`.** `activeAccountKeycastToken()` refreshes through the shared coordinator with the current pubkey and rejects unbound or mismatched sessions. `ExportKeyFailure.noKey` has dedicated copy and a widget test. An authenticated but unreadable account response stays `unknown` instead of becoming a policy denial. I also looked at the ownership gate widening and found no client regression in the sibling `/api/user/*` paths.

**`c78b8452c` closes one remaining case: a throwing clipboard failure closed the sheet silently.** `copyVerified()` handled a silently dropped write but did not catch an exception from `Clipboard.setData` or `Clipboard.getData`. The caller's `try/finally` had no `catch`, so the exception escaped while `finally` popped the sheet, and the user saw neither the success message nor `keyManagementKeycastCopyBlocked`. Both calls are now wrapped and return `false`, which routes into the existing blocked-copy path.

Two regression tests cover it, one for a throwing write and one for a throwing read. Both were confirmed failing before the change. `flutter test test/utils/clipboard_utils_test.dart` passes 5 of 5, `test/screens/key_management_screen_test.dart` passes 19 of 19, `flutter analyze lib test integration_test` reports no issues.

**Your engine analysis holds up.** I checked it against the 3.44.0 SDK: on Android `PlatformPlugin.setClipboardData` calls the void `setPrimaryClip` and `PlatformChannel` returns `result.success(null)` immediately after; on iOS `FlutterPlatformPlugin` assigns `UIPasteboard.string` then returns `result(nil)`. One scope limit worth stating: "success either way" applies to non-throwing failures, because the success call sits after the platform handler. A thrown failure never reaches it. That gap is what the commit above closes.

**Two facts for the reveal-versus-copy decision, offered as input rather than as an argument.** Verified read-back confirms the immediate source clipboard value; it does not prove another app or profile can paste it, since a managed policy can permit an app to write and read its own clipboard while blocking paste into an unmanaged destination. And the no-artifact claim holds for successful writes but not necessarily for the blocked-write path: if a blocked write leaves another app's content in place, the verification read is a read of that content and may surface platform clipboard UI. That was not device-tested.

**Nothing outstanding on your side.** Two things still gate merging and neither is yours: the reveal-versus-copy call, which is Liz's, and `divinevideo/keycast#326`, which should land first.

realmeylisdev:
The force-push was a rebase, not a content change — `git cherry` reports no commit from the old tip without an equivalent in the new one, so nothing that was already reviewed moved.

CI went red right after the approval, and it wasn't this diff. #6495 added Vietnamese, Urdu, Chinese, and Malay while this branch was open, so the eleven strings this flow adds to `app_en.arb` had no translation in the four new locales — `arb_consistency_test` failed on `app_ms.arb` and `gen-l10n` left four generated files stale.

Translating them turned up one thing no test here could have caught. `keyManagementKeycastRemoteSigning` already existed in those four files, still carrying the copy this branch replaces — *"No private key is stored on this device, so there is no nsec to copy here"* — sitting above a button that now copies exactly that key. The key was never missing, so it never appeared in what CI reported, and the localization test only asks that a locale differ from English, which a stale translation does. Four locales would have shipped saying the opposite of what the screen does. Retranslated against the new English.

All checks green on `f4256ec9`. Still holding for divinevideo/keycast#326.

Issue #6481, opened 2026-07-29
refactor(settings): move FeatureFlagScreen onto go_router


hm21:
## Problem

`mobile/lib/screens/settings/settings_screen.dart:324` opens the experimental-features screen with

```dart
onTap: () => Navigator.push(MaterialPageRoute(...)),
```

It is the only tile in that file still doing so — all fifteen siblings use `context.push`. The screen therefore lives only in the Navigator and never in the URL, so it does not survive a router refresh (auth state change, account-review status change) and cannot be deep-linked or restored.

## Why now

#6476 fixed the app-shell provider shape that made this visible (toggling Curated Lists re-inflated the subtree and the user landed back on Settings). That fix removes the trigger, but the screen is still the odd one out and stays fragile against any other router refresh.

## Acceptance

- `FeatureFlagScreen` is a `GoRoute` under the settings routes, opened with `context.push`.
- Deep link to its path works and the screen survives a `refreshListenable` fire.
- No remaining `Navigator.push(MaterialPageRoute(...))` in `settings_screen.dart`.

Raised during review of #6476 by @realmeylisdev.

PR #6502, opened 2026-07-30
fix(people-lists): stop background sync when curated lists turns off

### LINKED ISSUE #6494: fix(people-lists): stop background sync when curated lists turns off
## Problem

PR #6476 makes the app-shell PeopleListsBloc provider unconditional so flipping FeatureFlag.curatedLists no longer re-inflates MaterialApp.router. Entry points now gate reads while the flag is off, but a bloc that was already constructed while the flag was on remains alive for the rest of the session.

That means its local cache subscription and auth owner listener can continue after the user turns curated lists off, and a later owner/account transition can still call syncOwner for kind 30000.

## Expected follow-up

Add explicit flag-off lifecycle handling for the already-created PeopleListsBloc without reintroducing a conditional BlocProvider above MaterialApp.router. The fix should cancel active people-list repository work and prevent owner-change syncs while the feature is disabled, then restart cleanly if the flag turns back on.

## Constraints

- Do not key or structurally rebuild the provider chain above MaterialApp.router.
- Preserve the repository identity stream behavior from #6482.
- Add a regression test that constructs the bloc while the flag is on, flips curatedLists off, and verifies subsequent owner changes do not call syncOwner.

hm21:
## Description

The app-shell `PeopleListsBloc` provider is unconditional on purpose (#6477): a conditional entry changes the provider chain's shape, so flipping `curatedLists` re-inflated everything below it, `MaterialApp.router` included. `BlocProvider` laziness gates *construction* instead — but laziness cannot tear a bloc down again. A bloc built while the flag was on kept its local-cache subscription and its auth owner listener for the rest of the session, so a later owner/account transition still called `syncOwner` for kind 30000 with the feature switched off.

`PeopleListsBloc` now takes an `enabledStream` and stands down on its own:

- **Flag off** — close the lists subscription, drop the snapshot and any pending mutations, run no repository work.
- **Flag on** — re-subscribe and re-sync the current owner.

### Why the bloc keeps following owner changes while it is off

While standing down it still tracks `state.ownerPubkey`, and that is load-bearing rather than leftover. The owner stream (`authStateStream.map(...).distinct()`) is **not** seeded, so a re-subscribe after a flag-on would learn nothing until the *next* auth event. State is the only place the flag-on path can find out who is signed in. Without it, switching accounts while curated lists is off and then turning it back on would sync the account the user left. Pinned by `a flag-on syncs the account that signed in while off`.

### Why the re-subscribe hops through the owner bucket

`_onOwnerChanged` is documented as the only handler that opens `_listsSubscription` (#6482). `_onEnabledChanged` keeps that true: on flag-on it emits and then dispatches `PeopleListsOwnerChanged.rewire()` rather than calling `watchLists` itself. The teardown side is now a shared `_stopWatchingLists()` called from both handlers, which is safe because both are fully synchronous — what orphaned a subscription in #6482 was a writer that *suspends* between reading and nulling the field, not a second writer. The doc comment says so, so the next reader does not have to re-derive it.

### Why `enabled` lives in state, not in a private field

`_repository` is a mutable field with a documented deviation because it is an injected dependency. A flag the user toggles is mutable data, which `state_management.md` requires to live in the state object. It also makes the transition assertable in `blocTest`. The three `PeopleListsState(...)` constructions in `_onOwnerChanged` therefore carry `enabled:` explicitly — the default `true` would silently re-enable the feature on a sign-out.

`state.activeOwnerPubkey` (`ownerPubkey` while enabled, `null` while off) lets the four mutation handlers reuse their existing unauthenticated no-op path instead of each growing a second guard. The UI entry points already gate on the flag, so this is a backstop, not the primary gate.

### Why `enabledStream` is required

Same reason `repositoryStream` is (#6482): the wiring in `main.dart` is the one line this fix exists for. Dropping `enabledStream:` now fails analysis with `missing_required_argument` instead of silently restoring the bug. Callers that cannot see the flag change pass `Stream.value(true)`.

### Constraints from the issue

- The provider chain above `MaterialApp.router` is unchanged in shape and unkeyed. `a curated-lists flag-off stops syncing without remounting` asserts the flag flip stops the sync while `_mountCount` stays at 1 — the #4918 regression stays fixed.
- `repositoryStream` behaviour from #6482 is untouched. A repository swap still re-points the field; it just starts no work while the flag is off.

### The wiring is pinned end to end, not at the helper

`curatedListsEnabledStreamProvider` uses the existing `identityStreamOf` helper, which listens with `container.listen` precisely because nothing `watch`es it — the bloc `ref.read`s it once from `create:`. That is the failure mode #6482 documented, so `curated_lists_gate_test.dart` drives the **real** feature-flag provider chain (mock `SharedPreferences` → `FeatureFlagService.setFlag` → `featureFlagStateProvider.invalidateSelf` → `isFeatureEnabledProvider`) and asserts the stream seeds `true` and emits on each flip, rather than trusting a test double of the same shape.

**Related Issue:** Closes #6494

## Out of Scope

- **An `initiallyEnabled` constructor argument.** If the bloc were ever constructed while the flag is already off, one `syncOwner` would fire before the stream's seed arrives a turn later. Every entry point gates on the flag before reading the bloc (#6476), so that path does not exist today, and adding the parameter would mean a second source of truth for the same value plus a test for an unreachable state.
- **Cancelling a mutation that is already in flight when the flag flips.** On the success path its completion handler resolves `pendingMutations` against the disabled state, moving `status` to `ready`. On the rollback path it does more than that: `!result.submitted` — the common failure shape, since `PeopleListsRepositoryImpl` returns `PeopleListPublishResult.failed()` rather than throwing — and the `catch` both re-emit the `previousLists`/`previousIndex` captured *before* the teardown, and `_withoutMutation(..., failed: true)` pins `status` to `failure`. So the snapshot the flag-off just dropped comes back, and if an account switch happened in between it comes back under the new owner. The same emit also clears `pendingMutations`, which `UserListPeopleScreen._deleteSettled` reads as "the delete settled" — a flag-off mid-delete announces `curatedListDeletedSnack` and pops a route whose delete never landed. Deferred rather than fixed here, because the race is the pre-existing one and not a flag-off invention: those four rollback blocks are unchanged from `main`, and both modes already reproduce on a plain account switch with the flag never involved — where they are worse, since the restored state lands on `enabled: true` and actually renders. Guarding the rollback emits on the owner still matching fixes both, and belongs in a PR that owns the general case. **Tracked in #6504**, with the confirmed repro for each mode, so the deferral does not leave with this description.
- The feature-flag screen still bypasses go_router (#6481).

## Verification

- No generated-code inputs touched (`curatedListsEnabledStreamProvider` is a plain `Provider`, not codegen)
- `flutter test` over every PeopleLists and feature-flag suite — 566 passed, 3 skipped (`test/features/people_lists/`, `test/main_app_shell_repository_sync_test.dart`, `test/providers/list_providers_test.dart`, `test/providers/provider_identity_stream_test.dart`, `test/providers/feature_flag_provider_test.dart`, `test/services/feature_flag_service_test.dart`, `test/screens/feature_flag_screen_test.dart`, `test/screens/other_profile_screen_test.dart`, `test/screens/user_list_people_screen_test.dart`, `test/screens/search_results/`, `test/widgets/profile/`, `test/widgets/user_profile_tile_feature_flag_test.dart`, `test/blocs/list_search/`)
- Pre-push hook: merge-conflict check, analyzer, and changed-file tests all passed
- Review round: the `_mountCount` assertion added to `a curated-lists flag-off stops syncing` was dropped — `_AppShellProbe` is test-local, so no production change can move that counter and the assertion could not go red (thanks @realmeylisdev)

Each part of the fix was checked against a neutered version rather than assumed load-bearing. Three rows below were added in a follow-up commit: the sign-out `enabled:`, the un-gated `_onRepositoryChanged`, and the flag-off snapshot drop were all load-bearing but unpinned — the suite stayed green when each was neutered.

| Neutered | Tests that go red |
| --- | --- |
| `_onEnabledChanged` returns early | 9 |
| only the `!state.enabled` guard in `_onOwnerChanged` | 4 |
| `enabled:` dropped from the sign-out emit in `_onOwnerChanged` | 1 (`a sign-out while the flag is off keeps the feature off`) |
| `_onRepositoryChanged` made to stand down with the rest of the bloc | 1 (`a repository swap runs no sync while off but still lands`) |
| `activeOwnerPubkey` reverted to `ownerPubkey` | 1 (`mutations publish nothing while the flag is off`) |
| the `!state.enabled` guard in `_onRepositoryListsChanged` | 1 (`drops a snapshot that crossed the flag-off teardown`) |
| the flag-off emit relaxed to `state.copyWith(enabled: false)` | 1 (`turning the flag off cancels the lists subscription`) |

## Test plan

Manual verification on a real Samsung Galaxy — passed:

1. Settings → Experimental Features → toggle **Curated Lists** off. Expect to stay on the Experimental Features screen (the #6477 behaviour) and the lists affordances to disappear.
2. With the flag off, sign out and sign in as a different account, then turn **Curated Lists** back on. Expect the new account's lists, not the previous account's.
3. Turn the flag off and on a few times in a row and confirm the lists screen still loads and mutations (add/remove a pubkey) still publish.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
+++ b/mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
@@ -205,6 +205,10 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
       // Drops the snapshot and any pending mutations, but keeps the owner: the
       // owner stream is not seeded, so state is the only place a later flag-on
       // can learn who is signed in.
+      //
+      // A mutation still in flight resumes against the snapshot it captured
+      // before this emit and puts it back — the pre-existing account-switch
+      // race, tracked in #6504.
       emit(PeopleListsState(ownerPubkey: state.ownerPubkey, enabled: false));
       return;
     }
--- a/mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
+++ b/mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
@@ -27,10 +27,18 @@ typedef PeopleListsClock = DateTime Function();
 /// changes, and applies optimistic updates for user-initiated mutations
 /// before the repository returns.
 ///
-/// It depends only on the repository, an owner-pubkey stream, and a
-/// repository stream — it does not import other BLoCs. Owner transitions
-/// cancel the prior subscription, clear pending mutations, and start a new
-/// subscription plus an out-of-band sync.
+/// It depends only on the repository, an owner-pubkey stream, a repository
+/// stream, and an enabled stream — it does not import other BLoCs. Owner
+/// transitions cancel the prior subscription, clear pending mutations, and
+/// start a new subscription plus an out-of-band sync.
+///
+/// ## Feature-flag lifecycle
+///
+/// The bloc is registered unconditionally above `MaterialApp.router`, so it
+/// outlives `FeatureFlag.curatedLists` going off (#6477). While the flag is
+/// off it holds no repository subscription, runs no `syncOwner`, and publishes
+/// nothing — it only keeps following the owner so that turning the flag back
+/// on wires up the account that is signed in *then* (#6494).
 ///
 /// ## Failure recovery contract
 ///
@@ -54,6 +62,13 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
   /// change, so a caller that omits the stream silently reverts #6480. Pass
   /// `const Stream.empty()` when the repository genuinely cannot change.
   ///
+  /// [enabledStream] must emit whether `FeatureFlag.curatedLists` is enabled,
+  /// seeded with its current value. Required for the same reason
+  /// [repositoryStream] is: this bloc outlives the flag, so a caller that omits
+  /// the stream silently reverts #6494 and leaves a disabled feature syncing in
+  /// the background. Pass `Stream.value(true)` when the flag genuinely cannot
+  /// change.
+  ///
   /// [clock] stamps optimistic `updatedAt` values and mutation ids.
   /// Defaults to [DateTime.now] in UTC. Tests inject a fixed clock for
   /// determinism.
@@ -61,14 +76,20 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
     required PeopleListsRepository repository,
     required Stream<String?> ownerPubkeyStream,
     required Stream<PeopleListsRepository> repositoryStream,
+    required Stream<bool> enabledStream,
     String? initialOwnerPubkey,
     PeopleListsClock? clock,
   }) : _repository = repository,
        _ownerPubkeyStream = ownerPubkeyStream,
        _repositoryStream = repositoryStream,
+       _enabledStream = enabledStream,
        _clock = clock ?? _defaultClock,
        super(PeopleListsState(ownerPubkey: initialOwnerPubkey)) {
     on<PeopleListsStarted>(_onStarted);
+    on<PeopleListsEnabledChanged>(
+      _onEnabledChanged,
+      transformer: sequential(),
+    );
     on<PeopleListsRepositoryChanged>(
       _onRepositoryChanged,
       transformer: sequential(),
@@ -110,17 +131,20 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
   PeopleListsRepository _repository;
   final Stream<String?> _ownerPubkeyStream;
   final Stream<PeopleListsRepository> _repositoryStream;
+  final Stream<bool> _enabledStream;
   final PeopleListsClock _clock;
 
   StreamSubscription<String?>? _ownerSubscription;
   StreamSubscription<List<UserList>>? _listsSubscription;
   StreamSubscription<PeopleListsRepository>? _repositorySubscription;
+  StreamSubscription<bool>? _enabledSubscription;
 
   @override
   Future<void> close() async {
     await _ownerSubscription?.cancel();
     await _listsSubscription?.cancel();
     await _repositorySubscription?.cancel();
+    await _enabledSubscription?.cancel();
     return super.close();
   }
 
@@ -148,10 +172,53 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
       },
     );
 
+    await _enabledSubscription?.cancel();
+    _enabledSubscription = _enabledStream.listen(
+      (enabled) => add(PeopleListsEnabledChanged(enabled: enabled)),
+      onError: (Object error, StackTrace stackTrace) {
+        addError(error, stackTrace);
+      },
+    );
+
     final currentOwner = state.ownerPubkey;
     if (currentOwner != null && currentOwner.isNotEmpty) {
       add(PeopleListsOwnerChanged(ownerPubkey: currentOwner));
     }
+  }
+
+  /// Starts or stops all people-list work when `FeatureFlag.curatedLists`
+  /// flips.
+  ///
+  /// Turning the flag off cannot be expressed by dropping the `BlocProvider`:
+  /// a conditional entry changes the app-shell provider chain's shape and
+  /// re-inflates everything below it (#6477). So the already-constructed bloc
+  /// has to stand down on its own — otherwise its cache subscription and
+  /// `syncOwner` calls keep running for the rest of the session (#6494).
+  void _onEnabledChanged(
+    PeopleListsEnabledChanged event,
+    Emitter<PeopleListsState> emit,
+  ) {
+    if (event.enabled == state.enabled) return;
+
+    if (!event.enabled) {
+      _stopWatchingLists();
+      // Drops the snapshot and any pending mutations, but keeps the owner: the
+      // owner stream is not seeded, so state is the only place a later flag-on
+      // can learn who is signed in.
+      emit(PeopleListsState(ownerPubkey: state.ownerPubkey, enabled: false));
+      return;
+    }
+
+    final owner = state.ownerPubkey;
+    if (owner == null || owner.isEmpty) {
+      emit(state.copyWith(enabled: true));
+      return;
+    }
+
+    emit(state.copyWith(status: PeopleListsStatus.loading, enabled: true));
+    // Re-subscribing goes through the owner bucket so [_onOwnerChanged] stays
+    // the only handler that opens a lists subscription.
+    add(const PeopleListsOwnerChanged.rewire());
   }
 
   /// Re-points the bloc at a freshly built repository.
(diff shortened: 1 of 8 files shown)


mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
```diff
@@ -148,12 +172,55 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
...
+    Emitter<PeopleListsState> emit,
+  ) {
+    if (event.enabled == state.enabled) return;
+
+    if (!event.enabled) {
+      _stopWatchingLists();
+      // Drops the snapshot and any pending mutations, but keeps the owner: the
+      // owner stream is not seeded, so state is the only place a later flag-on
+      // can learn who is signed in.
+      emit(PeopleListsState(ownerPubkey: state.ownerPubkey, enabled: false));
```
realmeylisdev, line 212: The snapshot drop here is an invariant a mutation rollback can undo. A publish in flight when the flag goes off resumes with `previousLists`/`previousIndex` captured *before* this emit and puts them straight back.

Reproduced — owner A holding `[l1]`, add in flight, flag off (lists + pending empty), switch to owner B, publish returns `failed()`:

```
enabled=false status=failure owner=<ownerB> lists=[l1] index={<alice>: {l1}}
```

The "every UI entry point is flag-gated while off" argument in Out of Scope doesn't fully close it: on the next flag-on the state is `enabled: true, status: loading, lists: [l1]` until `watchLists(B)` delivers, so it's briefly *live* under the wrong owner, not just invisible. It does self-heal (B's snapshot overwrites to `lists=[]`), and I agree it's the pre-existing account-switch race rather than a flag-off invention. So: fine to defer, but please file the tracking issue — PR_REVIEW.md allows pre-existing debt in touched code to be deferred "or tracked independently", not deferred in a PR description that disappears on merge.

Same emit also clears `pendingMutations`, which `UserListPeopleScreen._deleteSettled` reads as "the delete settled". With a delete pending, a flag-off yields `deleteSettled=true, status=initial`, so the screen announces `curatedListDeletedSnack` and pops a route whose delete never landed. Also pre-existing on a mid-delete account switch, and it needs the flag flipped inside a publish round-trip — I'd fold it into the same issue.

mobile/lib/main.dart
```diff
@@ -2864,6 +2867,7 @@ class _DivineAppState extends ConsumerState<DivineApp>
                   repositoryStream: ref.read(
                     peopleListsRepositoryIdentityStreamProvider,
                   ),
+                  enabledStream: ref.read(curatedListsEnabledStreamProvider),
```
realmeylisdev, line 2870: Nit, and the third time in this family (#6480, #6482, now #6494): `required` forces *an* argument, but nothing pins *this* one. `Stream.value(true)` on this line reverts the fix with the whole suite green — `curated_lists_gate_test.dart` pins the provider in isolation, and the app-shell probe mirrors this `create:` with its own `_enabledStreamProvider` stand-in. Not asking for it here, just noting the pattern is now load-bearing across three injected streams.

mobile/test/main_app_shell_repository_sync_test.dart
```diff
@@ -177,6 +186,43 @@ void main() {
...
+      expect(firstLists.hasListener, isFalse);
+
+      // A repository flip while the feature is off must not restart the sync.
+      await _flipTo(tester, container, second);
+      verifyNever(
+        () => second.syncOwner(ownerPubkey: any(named: 'ownerPubkey')),
+      );
+      expect(secondLists.hasListener, isFalse);
+
+      expect(_mountCount, equals(1));
```
realmeylisdev: Nit: this can't go red from a production change — `_AppShellProbe` has no conditional provider entry, so `_mountCount` pins the probe's own shape rather than `main.dart`'s. The two assertions above it are the ones carrying the test.

realmeylisdev (commented):
Checked out `0ef1bb8`: `flutter analyze lib test integration_test` clean, format clean, 120 tests green across the people-lists + feature-flag suites, CI 16/16. I also neutered the `!state.enabled` guard in `_onRepositoryListsChanged` and confirmed exactly one test goes red (`drops a snapshot that crossed the flag-off teardown`) — so that one isn't vacuous. The state machine holds under off/on churn, sign-out, account switch and repo swap while off; no functional defect found, and no NIP-51 surface changes (`syncOwner` is a one-shot `queryEvents`, so nothing is left open on the relay).

Two asks: the description still says "that is why this is a draft" but the PR went ready at 11:02Z with the Galaxy run in the test plan still unconfirmed — worth updating or confirming. And the flag-off snapshot drop has one hole I'd like tracked in an issue rather than only in the PR body; details inline.

mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
hm21: Filed as #6504, with both modes reproduced against the real bloc at `0ef1bb8` rather than argued from the diff:

```
after teardown:  enabled=false status=initial lists=0 pending=0
after rollback:  enabled=false status=failure owner=<ownerB> lists=[l1] index={<alice>: {l1}}
```

```
deleteSettled=true status=initial failedBranch=false
```

So the `_deleteSettled` hole lands on the success branch exactly as you read it — `curatedListDeletedSnack` plus a `context.pop()` for a delete still on the relay. Both are in the one issue, since they share a root cause: the rollback emits are computed from a snapshot captured before the `await` and applied unconditionally after it.

Agreed on the process point, and `_onEnabledChanged` now carries a pointer to #6504 next to the teardown emit (a45319f) so the deferral does not leave with the description.

mobile/lib/main.dart
hm21: Noted, and you are right that `required` only forces an argument. `Stream.value(true)` here does revert the fix with the suite green — the gate test pins the provider, the app-shell probe mirrors this `create:` with its own stand-in, and nothing asserts the two are the same wire.

Leaving it as-is in this PR, but the pattern is now three deep (#6480, #6482, #6494), so it is worth one test that drives `main.dart`'s real `create:` rather than a mirror of it. Taking that as the next piece of work rather than folding it in here.

mobile/test/main_app_shell_repository_sync_test.dart
hm21: Correct, and it was worse than vacuous — the name claimed a constraint the assertion could not check. `_mountCount` only moves in `_MountProbe.initState`, which only `_AppShellProbe` builds, so no change to `main.dart` or the bloc can touch it.

Dropped in a45319f, and renamed to `a curated-lists flag-off stops syncing` so the name matches the two assertions that do the work.

hm21:
@realmeylisdev All three addressed.

- Rollback race + the `_deleteSettled` false positive: filed as #6504, both reproduced against the real bloc — your state dump matches line for line. `_onEnabledChanged` now points at it in code so the deferral survives the merge. Left that thread open for you to confirm the issue covers what you meant.
- `_mountCount`: dropped in a45319f, test renamed to what it actually pins.
- `enabledStream` nit: agreed, not folded in here — worth one test that drives `main.dart`'s real `create:` instead of a mirror, given it is the third stream in the family.

Description updated; the Galaxy run is still outstanding and now says so.

hm21:
Galaxy run done — all three steps in the test plan pass on the device, description updated. CI is 14/14 green on a45319f.

That closes your third point; the other two are resolved above, and #6504 is open for the deferred rollback race whenever you want to check it covers what you meant.

mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
realmeylisdev: Confirmed, #6504 is what I meant. Both modes, the shared root cause, and the acceptance criteria cover the account-switch variant too, which is the one that actually renders.

Left one precision note on the issue: `_onCreateRequested` is the odd one of the four, no `previousLists` capture and no `!result.submitted` branch, so it writes a stale status rather than restoring a snapshot. Does not change the shape of the fix, it just saves the next reader hunting for a rollback block that is not there.

Resolving this one. The deferral is tracked and the comment in `_onEnabledChanged` points at it.

realmeylisdev (approved):
lgtm. checked the flag-off teardown still goes red without `_onEnabledChanged` (9 tests, including the renamed app-shell one), so dropping `_mountCount` cost no coverage. #6504 covers both modes, left a small note there.

only loose end is the `enabledStream` wiring, which now lives in a resolved thread. worth an issue for the three-stream case so it survives the merge, same as we did for #6504.

PR #6505, opened 2026-07-30
chore(deps): upgrade pro_video_editor to 2.11.1

### LINKED ISSUE #6506: chore(deps): upgrade pro_video_editor to 2.11.1
Auto-created from PR #6505.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6505
**Branch:** `chore/deps-pro-video-editor-2-11-1`

hm21:
Closes #6506

## Description

Bumps `pro_video_editor` from 2.11.0 to 2.11.1, which fixes two cancellation races in the iOS/macOS bitrate-capped export path.

When the export watchdog cancelled a stalled render, `onCancel` tore the reader and writer down from the cancelling task's thread while the sample pumps were still running on their own queues — so the teardown landed *inside* a pump operation and killed the process. Users saw this as a permanently blank screen after tapping Next, followed by a crash, with no error and no retry.

Two crash signatures on 1.0.17 (813):

- `NSInternalInconsistencyException: -[AVAssetWriterInput requestMediaDataWhenReadyOnQueue:] Cannot call method when status is 4` — the writer was already cancelled when a pump armed its input.
- `EXC_BAD_ACCESS` in `copyNextSampleBuffer` — `cancelReading()` tore the reader down mid-read. The old cancellation flag was checked before the copy, but nothing kept the cancel out of the window between the check and the call.

Upstream commit `9b4441a4` (`fix: improve thread-safe cancellation handling in BitrateCappedExporter`) routes every reader/writer touch on both sides through one gate, so a teardown can only land *between* whole operations. Nothing changes in this repo beyond the dependency version: the app-side failure path (`renderFailed` plus the retry affordance from #6058) already existed, it was just never reached because the process died first. A stalled export now surfaces as a retry instead of a crash.

The constraint moves to `^2.11.1` rather than staying at `^2.11.0` so a fresh resolve cannot land on a version without the fix.

`2.11.1` also includes upstream image-orientation fixes from `21a01dfd` and a changelog correction from `c985c4cf`. Those touch Android, iOS, and macOS image decode paths for chroma-key backgrounds, image layers, and stop-motion frames. The Android-side changes are not the crash fix, but they are part of the package version being adopted.

**Related Issue:** Closes #6506. The existing retry affordance was introduced for #6058.

## Out of Scope

The crash fixed by `9b4441a4` is iOS/macOS only (AVFoundation). This PR does not change Divine application code, and it does not attempt to address unrelated Crashlytics issues.

Two other issues found in the same Crashlytics pass are deliberately left out: the `_NativeParagraphBuilder.addText` UTF-16 crash (unsanitized notification comment quotes) and the `SqliteException(26)` DB-open failures. Both are unrelated to this dependency and want their own PRs.

## Verification

- `flutter test test/services/video_editor test/providers/video_editor_provider_test.dart test/providers/video_publish_provider_test.dart` — 406 passed
- `flutter pub get` changed exactly one dependency; the lockfile diff is `pro_video_editor` only
- pub.dev lists `pro_video_editor` 2.11.1 with archive SHA `b460f17b884e3e6d2f24a9bcc55909acd9f62d9fc819ed2a6973acbf60898e0b`, matching `pubspec.lock`
- Upstream compare `v2.11.0...v2.11.1` contains the cancellation fix commit `9b4441a4`
- The upstream fix was verified in the app on device before 2.11.1 was published

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/mobile/pubspec.lock
+++ b/mobile/pubspec.lock
@@ -1839,10 +1839,10 @@ packages:
     dependency: "direct main"
     description:
       name: pro_video_editor
-      sha256: "8b4b2ffa9c3dd8995b01800f40ca5a5fd37f5431734a5b9dd0f6759580800ffa"
+      sha256: b460f17b884e3e6d2f24a9bcc55909acd9f62d9fc819ed2a6973acbf60898e0b
       url: "https://pub.dev"
     source: hosted
-    version: "2.11.0"
+    version: "2.11.1"
   process:
     dependency: transitive
     description:
--- a/mobile/pubspec.yaml
+++ b/mobile/pubspec.yaml
@@ -330,7 +330,7 @@ dependencies:
   firebase_performance: ^0.11.1+4
   firebase_messaging: ^16.1.1
   audio_session: ^0.2.2
-  pro_video_editor: ^2.11.0
+  pro_video_editor: ^2.11.1
   pro_image_editor: ^13.2.3
   flutter_colorpicker: ^1.1.0
   file_picker: ^10.3.7


NotThatKindOfDrLiz (approved):
Approved after independent verification. The dependency bump is limited to pubspec/pubspec.lock, the lockfile SHA matches pub.dev for pro_video_editor 2.11.1, CI is green, and the upstream compare includes the intended BitrateCappedExporter cancellation fix. I also updated the PR body to make the additional 2.11.1 Android/iOS/macOS image-orientation changes explicit, so the QA surface is no longer understated.

mobile/pubspec.yaml
```diff
@@ -330,7 +330,7 @@ dependencies:
   firebase_performance: ^0.11.1+4
   firebase_messaging: ^16.1.1
   audio_session: ^0.2.2
-  pro_video_editor: ^2.11.0
+  pro_video_editor: ^2.11.1
```
realmeylisdev, line 333: **`Out of Scope` needs a correction before this merges.**

2.11.1 does change the Android render path:

- `ChromaKeyEffect.kt` — background image decode is now EXIF-oriented
- `ApplyImageLayer.kt` — EXIF orientation, plus decode-at-scaled-size
- `StopMotionGenerator.kt`
- new `src/shared/media` package (`ImageOrientation`) and `OverlayDecodeSizeTest.kt`
- `android/build.gradle` now declares `androidx.exifinterface:exifinterface:1.4.2`

All three of those Kotlin paths are live here — `chroma_key_editor_cubit.dart:172`, `video_editor_render_service.dart:1235`, `watermark_download_service.dart:409`.

Nothing is actually broken by it. I resolved `debugRuntimeClasspath` and `exifinterface` was already on it transitively (1.4.1 via `image_picker_android`, plus 1.3.6 and 1.0.0 elsewhere); everything converges on a single 1.4.2, so this is a resolved-version move rather than a new artifact, and the debug APK builds. The `androidx.media3` force is still resolving 1.10.1 down to 1.10.0 as intended.

But the stated blast radius will cost someone real time if they ever bisect an Android render or overlay-orientation bug across this range.

realmeylisdev, line 333: **This also fixes a live bug in our own chroma key feature — worth saying so out loud.**

Before 2.11.1, a gallery photo used as a chroma-key background exported sideways: the renderer ignored the EXIF `Orientation` tag on all three platforms, while our preview (`chroma_key_backdrop.dart` → `Image.file`) honors it. So preview and export disagreed for any portrait photo, and `video_clip_chroma_key_screen.dart:205` passes the picker's file straight through with EXIF intact — exactly the case that broke.

2.11.1 makes the renderer honor EXIF, so the two now agree. That is a user-visible fix to the chroma key feature from #6483, which landed two commits ago.

Two asks:

- Mention it in the description. This is a behavior change, not only a crash fix.
- File a tracking issue against #6483 so the fix is recorded there rather than buried in a dependency bump, and QA a portrait gallery photo exporting upright on iOS and Android.

I did check the obvious double-rotation risk and it is clean: macOS `NSImage` used to bake the orientation in while iOS `UIImage.cgImage` did not, and `CGImageSource` applies neither, so both platforms converge on exactly one rotation. Our `ImageLayer` bytes (canvas PNGs, the watermark) carry no EXIF tag, so that path is unchanged.

mobile/pubspec.lock
```diff
@@ -1839,10 +1839,10 @@ packages:
     dependency: "direct main"
     description:
       name: pro_video_editor
-      sha256: "8b4b2ffa9c3dd8995b01800f40ca5a5fd37f5431734a5b9dd0f6759580800ffa"
+      sha256: b460f17b884e3e6d2f24a9bcc55909acd9f62d9fc819ed2a6973acbf60898e0b
       url: "https://pub.dev"
     source: hosted
-    version: "2.11.0"
+    version: "2.11.1"
```
realmeylisdev, line 1845: **The 2.11.1 changelog never mentions the cancellation fix.**

Upstream's entry lists only the three EXIF / orientation / decode-size items — nothing about the reader/writer cancellation races, which are the entire justification for this bump. I confirmed the fix is really in the published archive (`CancelState` → `SessionGate`, with the arm, copy, append and `markAsFinished` calls all routed through the gate and a single-owner `end()`), so this is a documentation gap rather than a missing change.

Since you maintain the package, worth amending upstream: as it stands the reason for this PR can't be reconstructed from the dependency alone, which makes it awkward to audit or bisect later.

mobile/pubspec.lock
```diff
@@ -1839,10 +1839,10 @@ packages:
     dependency: "direct main"
     description:
       name: pro_video_editor
-      sha256: "8b4b2ffa9c3dd8995b01800f40ca5a5fd37f5431734a5b9dd0f6759580800ffa"
+      sha256: b460f17b884e3e6d2f24a9bcc55909acd9f62d9fc819ed2a6973acbf60898e0b
```
realmeylisdev, line 1842: **Integrity checks out — and a note on what CI can and can't tell you here.**

`b460f17b…` matches the pub.dev archive byte-for-byte, and `flutter pub get` produced zero drift: the lockfile delta is exactly these two lines, no transitive movement.

Worth flagging that the green checks carry no signal for this particular change. No workflow runs `flutter build ios`, `apk`, `appbundle`, `xcodebuild` or `gradlew`; the only build job is Flutter Web, where this plugin's native code never compiles; and the four test shards run against a Dart API that is byte-identical between the two versions.

I built both locally to close that gap — iOS `Runner.app` and an Android debug APK both compile against 2.11.1, and `flutter test test/services/video_editor test/providers/video_editor_provider_test.dart test/providers/video_publish_provider_test.dart` gives 406 passing, matching your number. So this is informational, not a problem with the PR.

realmeylisdev (approved):
Verified this one end to end: the lockfile sha256 matches the pub.dev archive byte-for-byte, the Dart API is byte-identical between 2.11.0 and 2.11.1, and both iOS and Android build locally against it. The Swift `SessionGate` change is exactly what the description says it is.

One correction needed before merge — "the Android render path is untouched" isn't right. Details inline, along with a behavior change worth surfacing.

PR #6507, opened 2026-07-30
fix(upload): cap ProofMode headers so publishes stop dying at the edge

### LINKED ISSUE #6529: fix(upload): cap ProofMode headers so publishes stop dying at the edge
Auto-created from PR #6507.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6507
**Branch:** `fix/proofmode-upload-header-size`

hm21:
Closes #6529

## Description

On some Android devices every video publish fails, and the retry loop hides why. It is not all of them, which is what made this hard to see.

The client base64-encodes the full ProofMode manifest into `X-ProofMode-Manifest` and then the device attestation *a second time* into `X-ProofMode-Attestation`. Android's attestation is the `X509Certificate.toString()` dump of the whole cert chain, so its size is device-dependent — and it goes on the wire twice.

**Blast radius:** publishing breaks once the raw attestation passes roughly **47.5 KB**, because that is where the doubled payload crosses what `media.divine.video` accepts. Probed with the pre-fix header shape:

| Raw attestation | Combined headers | Result |
|---|---|---|
| 46,500 B | 126 KB | `401` (reaches the handler) |
| 47,500 B | 128 KB | `401` |
| 48,000 B | 131 KB | `502` |
| 53,826 B (Pixel 8 Pro) | 147 KB | `502` |

So: a capture whose attestation dumps over ~47.5 KB fails to publish, one under it does not, iOS ships a compact integrity token and is unaffected, and captures without ProofMode send no such headers at all. Same root cause either way — the doubled, uncapped payload — but the failure is a long tail rather than an Android-wide outage.

**The threshold is per capture, not per device.** The client reads the attestation from `<proofDir>/<fileHash>.attest` for the file being proofed, and that file only exists for captures where attestation actually succeeded. Chain size is device-dependent, so a big-attestation handset is the precondition — but on such a handset some clips carry 53,826 bytes and others carry none, and only the former fail. A user reporting "new videos post, older clips don't" is the expected shape of this bug, not a counter-example to it.

Measured against production, combined request headers must stay under:

| Protocol | Limit | Behaviour past it |
|---|---|---|
| HTTP/1.1 (what the Dart client uses) | **128 KiB** | edge answers `502` |
| HTTP/2 | **64 KiB** | connection closed mid-request |

The boundary is sharp: over HTTP/1.1, 130,800 bytes of headers still reaches the handler (`401`), 131,000 returns `502`. Reproducing the app's exact header shape — `X-ProofMode-Manifest` at 74,788 bytes plus `X-ProofMode-Attestation` at 71,768 — returns `502`, the same status the client logged on all six attempts.

That explains the whole failure pattern in the session log:

| Request | ProofMode headers | Result |
|---|---|---|
| Chunk PUT (6.45 MB body) | no | ✅ 1.53 MB/s |
| `POST /upload/*/complete` | yes | ❌ 502, 6/6 attempts |
| Legacy `PUT /upload` | yes | ❌ reset while streaming the body |
| OS-first background PUT | yes | ❌ "network" |

The one request type that succeeds is the one carrying no ProofMode headers. On the legacy `PUT` the server answers 502 early and closes while the client is still writing 6.45 MB, which surfaces as `SocketException: Write failed (Connection reset by peer)`.

Nothing consumed these headers. `divine-blossom` (which serves `media.divine.video`, including `PUT /upload` and `/upload/*/complete`) contains no reference to `proofmode` outside a debug script and docs, and `divine-upload-server` has none at all. The canonical manifest is already the `proofmode` tag on the kind 34236 event, and the C2PA manifest is embedded in the uploaded bytes — which is what `cloud-functions/process-blob` actually reads.

### Where the proof actually travels

Worth being explicit, because "we drop the manifest header" reads alarming: the ProofMode payload never depended on these headers. Two separate channels carry the same data, and only one of them is read.

| Channel | Carries | Read by anyone? |
|---|---|---|
| `X-ProofMode-*` HTTP headers on the Blossom upload | manifest + attestation | **no** — and never stored either; Blossom keeps blobs, not request headers |
| `proofmode` / `device_attestation` tags on the kind 34236 event | manifest + attestation | **yes** — this is what the ProofMode badge and verification level read from |

Unread headers on their own would have been waste, not a bug. What broke uploads was the combination: a payload nothing consumes, sent twice, uncapped, large enough to cross the request-header limit.

The event route is verified to carry the full payload with room to spare. Probing `wss://relay.divine.video` with an intentionally invalid signature (so nothing can be stored), the relay rejects on content rather than size all the way up:

| Event size | Relay response |
|---|---|
| 56 KB | `invalid: video missing required 'title' tag` |
| 110 KB (what a big-attestation capture actually needs) | `invalid: video missing required 'title' tag` |
| 500 KB | `invalid: video missing required 'title' tag` |
| 1 MB | connection reset |

So the ~110 KB of proof tags sit at roughly a fifth of what the relay accepts. Dropping the header copies loses nothing.

(Noted in passing, not changed here: the event carries the attestation twice as well — the `proofmode` tag already embeds it, and `device_attestation` repeats it. Harmless at this headroom, but it is the same redundancy.)

**The fix:** attach the headers best-effort inside an 8 KiB budget, ordered cheapest-and-most-identifying first (`X-ProofMode-C2PA` → `-Signature` → `-PublicKey` → `-Attestation` → `-Manifest`), dropping whatever does not fit and logging what was dropped.

**New header — `X-ProofMode-PublicKey`.** The PGP signature survives the budget, but the key that verifies it previously only lived inside `X-ProofMode-Manifest`, which is exactly the header dropped for an oversized proof. That left a header-only consumer with a signature it could not check. The key now ships as its own header ranked directly after the signature, so the pair stays together above the bulk headers. With real field sizes the result is:

| Header | Size | |
|---|---|---|
| `X-ProofMode-C2PA` | 60 B | kept |
| `X-ProofMode-Signature` | 1,096 B | kept |
| `X-ProofMode-PublicKey` | 2,732 B | kept |
| `X-ProofMode-Attestation` | 71,768 B | dropped |
| `X-ProofMode-Manifest` | 75,856 B | dropped |

3,958 bytes of the 8,192 budget. The key comes from the ProofMode `<hash>-pubkey.asc` file the client already reads; captures where libProofMode does not emit that file simply omit the header.

**Backward compatibility:** the change is additive — every header that ships today keeps its exact value. Real captures do gain one header: `NativeProofData` emits `publicKey`, so any manifest carrying it now also sends `X-ProofMode-PublicKey` (2,732 B, well inside the budget). All ten pre-existing ProofMode tests pass unchanged, and their headers are byte-identical, because their fixtures carry no `publicKey`. Oversized proofs degrade from "upload always fails" to "upload succeeds, carrying the C2PA id and a verifiable signature/key pair". A mid-size proof that would have fit whole can now spend 2,732 B on the key and drop the manifest copy instead — the better trade for a header-only consumer, since the manifest embeds the key but not the reverse. No server or protocol change is required, and no shipped client is affected by this change.

8 KiB is deliberately far below both measured limits, so nginx-fronted Blossom deployments with an 8 KB per-header budget stay safe too, and splitting the payload across several headers would not help — the limit applies to the sum.

Both docs that presented "whole manifest in a header" as the design are corrected, so this does not get reintroduced.

**Related Issue:** 

### What the user actually sees

Worth pinning down, because support triages off the sentence in the sheet and neither variant says "proof" or "header". Which one appears depends on which request in the chain surfaces last:

| Failing request | Classified as | Sheet copy |
|---|---|---|
| `POST /upload/*/complete` answered `502` | `502` → `SERVER_UNAVAILABLE` → `serverDown` | "The media server (media.divine.video) is temporarily down. Try again shortly or choose another in your settings." |
| Legacy `PUT /upload` reset mid-body | `SocketException` (no status) → `NETWORK_ERROR` → `serverUnreachable` | "Could not reach the server. Please try again in a moment." |

`categorizeError` branches on the HTTP status before any string matching, so a clean 502 never reaches the network bucket; the reset does, because it carries no status. Both then re-localize through `PublishErrorKind`, so the raw English sentences in `UploadProgressReporter` are log text, not what the user reads.

## Out of Scope

- **Retry behaviour.** After attempt 1 the bytes are fully uploaded (`resuming from offset 6760190`, `0 chunks this session`), so attempts 2–6 spent 90 seconds replaying the same deterministic failure and then logged "Network error on WiFi. Check your connection and try again." (surfacing as the `serverUnreachable` copy above). Pre-existing, unrelated to this diff.
- **Shipped clients.** ≤1.0.17 keeps sending the doubled payload, so affected devices stay broken until they update. The rejection happens at the edge, before any handler runs, so no server-side mitigation can rescue them.
- Separately visible in the same session log and untouched here: `ERROR_CODE_PARSING_CONTAINER_MALFORMED` when the editor previews a C2PA-signed recording, and `VerifierApiException: verifier returned 400 "Invalid identity"` from `MyProfileBloc`.

## Verification

- `flutter test` in `mobile/packages/blossom_upload_service` — 230 tests pass
- `flutter test --coverage` — 859/859 lines covered in `blossom_upload_service.dart`, package stays at the 100% gate
- `flutter test test/services/blossom_resumable_upload_integration_test.dart` — 7 pass
- `dart format` — clean
- Production probes of `media.divine.video` establishing both protocol limits, the sharp 131,072-byte HTTP/1.1 boundary, the `401` baseline with small headers, and a `502` reproduction using the app's exact header sizes

- Real Samsung Galaxy run: recorded and published with ProofMode on. The budget logic fired as designed — `sent 1121 bytes and dropped X-ProofMode-Attestation (11956B), X-ProofMode-Manifest (13512B)` — the upload completed and the kind 34236 event got its ProofMode tags.

Two limits on what that device run establishes. Its attestation is 8,967 raw bytes, roughly five times under the ~47.5 KB failure threshold, so the pre-fix request would have been about 26 KB of headers and published fine anyway: the run proves the new path is regression-free, not that the 502 is gone. Observing the fix itself still needs a big-attestation handset of the Pixel 8 Pro class. Separately, libProofMode emitted no `<hash>-pubkey.asc` for that capture, so `X-ProofMode-PublicKey` was omitted and the signature shipped alone — the signature/key pairing holds only for captures that produce the key file.

## Type of Change

- [x] 📝 Documentation

### DIFF
--- a/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
+++ b/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
@@ -108,8 +108,14 @@ Optional request headers:
 
 - `X-ProofMode-Manifest`
 - `X-ProofMode-Signature`
+- `X-ProofMode-PublicKey`
 - `X-ProofMode-Attestation`
 - `X-ProofMode-C2PA`
+
+`X-ProofMode-PublicKey` carries the PGP public key matching
+`X-ProofMode-Signature`. The two travel together because a signature cannot be
+verified from headers alone without it; a server that sees one without the
+other must treat the signature as unverified rather than invalid.
 
 When present, these headers carry the same ProofMode metadata that legacy
 `PUT /upload` accepts. Clients should send them on `complete` so resumable
--- a/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
+++ b/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
@@ -138,12 +138,13 @@ under **128 KiB over HTTP/1.1** (past it the edge answers 502) and under
 
 An uncapped client sends the device attestation twice — base64 inside
 `X-ProofMode-Manifest` and again in `X-ProofMode-Attestation` — so it crosses
-128 KiB once the raw attestation passes roughly 47.5 KB, and every upload from
-that device then fails with an opaque 502, or with a connection reset while
+128 KiB once the raw attestation passes roughly 47.5 KB, and the upload
+carrying it then fails with an opaque 502, or with a connection reset while
 the body is still streaming on the legacy `PUT /upload` path. Android
-attestation size is device-dependent (it is the `X509Certificate.toString()`
-dump of the whole chain), so an uncapped client fails on some devices and not
-others.
+attestation size is the `X509Certificate.toString()` dump of the whole chain,
+so it varies by device *and* by capture — the client stores it per file hash,
+and only for captures where attestation succeeded. An uncapped client
+therefore fails on some devices, and on those devices only for some uploads.
 
 Any future proof payload that can exceed the budget must travel in the
 `complete` request body, not in a header.
--- a/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
+++ b/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
@@ -134,12 +134,16 @@ not fit, cheapest-and-most-identifying first.
 
 Measured against `media.divine.video`, combined request headers must stay
 under **128 KiB over HTTP/1.1** (past it the edge answers 502) and under
-**64 KiB over HTTP/2** (past it the connection is closed mid-request). An
-Android hardware attestation chain base64-encodes to ~75 KB in
-`X-ProofMode-Manifest` and another ~72 KB in `X-ProofMode-Attestation`, so an
-uncapped client sends ~147 KB and fails every upload with an opaque 502 — or,
-on the legacy `PUT /upload` path, a connection reset while the body is still
-streaming.
+**64 KiB over HTTP/2** (past it the connection is closed mid-request).
+
+An uncapped client sends the device attestation twice — base64 inside
+`X-ProofMode-Manifest` and again in `X-ProofMode-Attestation` — so it crosses
+128 KiB once the raw attestation passes roughly 47.5 KB, and every upload from
+that device then fails with an opaque 502, or with a connection reset while
+the body is still streaming on the legacy `PUT /upload` path. Android
+attestation size is device-dependent (it is the `X509Certificate.toString()`
+dump of the whole chain), so an uncapped client fails on some devices and not
+others.
 
 Any future proof payload that can exceed the budget must travel in the
 `complete` request body, not in a header.
--- a/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
+++ b/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
@@ -116,6 +116,28 @@ When present, these headers carry the same ProofMode metadata that legacy
 uploads preserve the same server-side ProofMode handling without forcing the
 video bytes back onto the legacy single-request upload path.
 
+These headers are **supplementary and size-capped**. Servers must not treat
+them as the delivery mechanism for the proof: the canonical manifest is the
+`proofmode` tag on the kind 34236 event, and the C2PA manifest is embedded in
+the uploaded bytes. No Divine Blossom server reads them today.
+
+Clients must keep the combined `X-ProofMode-*` header size within a
+conservative budget — the mobile client uses 8 KiB
+(`BlossomUploadService.maxProofModeHeaderBytes`) — and drop headers that do
+not fit, cheapest-and-most-identifying first.
+
+Measured against `media.divine.video`, combined request headers must stay
+under **128 KiB over HTTP/1.1** (past it the edge answers 502) and under
+**64 KiB over HTTP/2** (past it the connection is closed mid-request). An
+Android hardware attestation chain base64-encodes to ~75 KB in
+`X-ProofMode-Manifest` and another ~72 KB in `X-ProofMode-Attestation`, so an
+uncapped client sends ~147 KB and fails every upload with an opaque 502 — or,
+on the legacy `PUT /upload` path, a connection reset while the body is still
+streaming.
+
+Any future proof payload that can exceed the budget must travel in the
+`complete` request body, not in a header.
+
 Example successful response:
 
 ```json
--- a/mobile/docs/PROOFMODE_ARCHITECTURE.md
+++ b/mobile/docs/PROOFMODE_ARCHITECTURE.md
@@ -320,6 +320,23 @@ class VineRecordingController {
 
 ProofMode manifest is uploaded alongside video to Blossom server:
 
+> **The `X-ProofMode-*` headers are best-effort supplementary metadata, not the
+> delivery mechanism for the manifest.** The canonical copy is the `proofmode`
+> tag on the kind 34236 event, and the C2PA manifest is embedded in the video
+> bytes themselves. No Blossom server currently reads these headers.
+>
+> They are also hard-capped: `BlossomUploadService.maxProofModeHeaderBytes`
+> (8 KiB combined) governs what actually goes on the wire, cheapest-and-most-
+> identifying first (`X-ProofMode-C2PA`, then `-Signature`, `-Attestation`,
+> `-Manifest`). An Android hardware attestation chain is ~54 KB and used to
+> travel twice — ~75 KB base64-encoded in `X-ProofMode-Manifest` plus ~72 KB
+> again in `X-ProofMode-Attestation`. Measured against `media.divine.video`,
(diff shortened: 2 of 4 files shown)


realmeylisdev:
A field report that matches this and sharpens one part of the diagnosis.

A user on a **Pixel 8 Pro** running **1.0.17+668** reports "I was able to post a new video but still unable to post the old clips, like they're corrupted or something." That is the handset class the PR calls out as unverified, so it is worth reconciling — but note it is a *partial* failure, and the PR predicts a total one.

**The blast-radius table is per-device; the code is per-file.** The PR body, `25c4cc073`, and the added `PROOFMODE_ARCHITECTURE.md` all model attestation size as a device constant ("a device whose chain dumps over ~47.5 KB fails *every* publish, a device under it never fails"). But `readProofMetadata` reads `<proofDir>/<fileHash>.attest` per file (`mobile/lib/services/native_proofmode_service.dart`, the `deviceAttestation.existsSync()` block), and libProofMode 1.0.25 writes that file once — at the first proof pass for that hash, gated on `autoNotarize` **and** `MediaWatcher.isOnline(context)` at that instant — then `if (proofExists(...)) return mediaHash` short-circuits forever after. So two clips on one handset can be 0 B and 53,826 B, and a partial failure is expected rather than evidence against this diagnosis.

That changes the acceptance criteria more than the root cause: under a byte budget, the shipped behaviour becomes "some publishes are silently unattested," which reads like a product call for @LizSmith / @n8fr8 rather than a code one.

**Two smaller corrections.** The 502 does not surface as "Network error on WiFi. Check your connection and try again." `categorizeError` branches on status code before any string matching — `if (code == 502 || code == 503 || code == 504) return 'SERVER_UNAVAILABLE'` (`mobile/lib/services/upload/upload_progress_reporter.dart:183-185`) — and the "Upload Failed" sheet renders the ARB layer, so it reads `publishErrorServerDown`: "The media server ({serverName}) is temporarily down. Try again shortly or choose another in your settings." (`mobile/lib/l10n/app_en.arb:4093`). Support triages off that sentence, so it is worth fixing in the PR body and the doc. Separately, #6161 looks like the same defect reported on 2026-07-17 and routed to backend/ops ten days before this diagnosis existed — the edge 502 gives the client nothing to distinguish it from connectivity.

**An alternative remedy, offered as an option not a replacement:** #6551 drops the standalone `X-ProofMode-Attestation` header entirely, since the attestation is already inside the base64 `X-ProofMode-Manifest` that `:2922` encodes wholesale. That takes ~147 KB to ~75 KB, under the limit, with nothing dropped and no budget needed. Also relevant to whichever approach lands: no server currently reads any `X-ProofMode-*` header — `handle_upload_service_proxy` and `handle_upload_complete` in divine-blossom (`src/main.rs:3782-3803`, `:3608-3643`) build a fresh request with only Host, Authorization, Content-Type and Content-Length, and `git log --all -S"X-ProofMode"` across that repo returns no commits. The attestation that reaches the backend comes from Nostr event tags.

Happy to fold #6551 into this PR instead if you would rather keep it in one place.

hm21:
Thanks — the per-file point is right and it's now in the PR body, `PROOFMODE_ARCHITECTURE.md` and the BUD doc (c70ab6723). The blast-radius framing was modelling attestation size as a device constant; it's read per `<fileHash>.attest`, so a big-attestation handset failing *some* captures is the expected shape. That reconciles the Pixel 8 Pro report rather than contradicting it.

The error-copy correction is half right, and the half that's off matters for the same triage reason you raised. You're correct that `categorizeError` branches on status before string matching, so a clean 502 on `complete` reads `publishErrorServerDown`. But the sentence in the session log came from the legacy `PUT` being reset mid-body — a `SocketException` carries no status, falls through to the `'connection'` branch, and reads `publishErrorServerUnreachable` ("Could not reach the server"). Both are reachable depending on which request surfaces last, so I documented the pair as a table instead of swapping one absolute claim for another.

On folding in #6551: I'm keeping the budget. Dropping the duplicate halves the payload but leaves it unbounded — the failure threshold moves from ~47.5 KB of raw attestation to ~92 KB rather than going away, and the surviving single header is still ~77 KB, over the 64 KiB HTTP/2 limit and over nginx's default 8 KB-per-line budget, which matters because users can point at their own Blossom server. The `TODO(n8fr8)` in `native_proofmode_service.dart` also plans to fold per-clip attestations into the combined manifest, which would blow an unbudgeted manifest header on any multi-clip post. Given your own finding that no server reads these headers, dropping what doesn't fit costs nothing, while keeping it costs ~75 KB of dead wire on every affected publish.

I'll close #6551 as superseded once this lands — the edge-strip finding and the copy correction were the useful parts and both are folded in here.

rabble (approved):
Approved. ProofMode headers are capped across upload paths, and the focused ProofMode tests pass.

PR #6513, opened 2026-07-30
fix(lists): stop the delete-list dialog reporting a fatal on a dead route

### LINKED ISSUE #6512: Delete-list dialog: Cancel throws on a torn-down route and is filed as a FATAL — curated_list_feed_screen.dart:419 [live in 1.0.17 and 1.0.18]
## Summary

The **Cancel** button of the "delete this list?" confirmation dialog on `CuratedListFeedScreen` calls `Navigator.of(dialogContext)` on a `BuildContext` whose element has already been unmounted. The ancestor lookup throws, the pop never happens, and the tap does nothing.

**Nothing crashes.** `GestureRecognizer.invokeCallback` catches the error and hands it to `FlutterError.reportError`. It reaches Crashlytics as a *fatal* only because `mobile/lib/services/crash_reporting_service.dart:37-45` wires `FlutterError.onError → recordFlutterFatalError` and `main.dart:1247` forwards anything its downgrade list does not match. The `Fatal Exception: FlutterError` header is the firebase_crashlytics iOS plugin's hardcoded name for every Dart error, not the exception type.

**Crashlytics issue:** `c043a09d4efa7432d2f36bd824d7c3ca`
[Open in Firebase Console](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/c043a09d4efa7432d2f36bd824d7c3ca)

## Impact

- Version **1.0.17 (build 813)**, session `2bc46c0f7a4c480b97588989281745e0`, 2026-07-30.
- User-visible symptom: Cancel does nothing. The dialog's route is already gone by then, so nothing is left stuck on screen.
- Triage cost: the report is grouped as a fatal, so it competes with real crashes in the dashboard.

## Evidence

Dart stack from the Crashlytics export:

```
Fatal Exception: FlutterError
0  StatefulElement.state + 5938 (framework.dart:5938)
1  Navigator.of + 2919 (navigator.dart:2919)
2  _CuratedListFeedScreenState._confirmDeleteList.<fn>.<fn> + 419 (curated_list_feed_screen.dart:419)
3  _InkResponseState.handleTap + 1224 (ink_well.dart:1224)
4  GestureRecognizer.invokeCallback + 345 (recognizer.dart:345)
5  TapGestureRecognizer.handleTapUp + 758 (tap.dart:758)
6  BaseTapGestureRecognizer._checkUp + 383 (tap.dart:383)
7  BaseTapGestureRecognizer.acceptGesture + 353 (tap.dart:353)
8  GestureArenaManager.sweep + 173 (arena.dart:173)
9  GestureBinding.handleEvent + 532 (binding.dart:532)
```

Resolved against Flutter 3.44.0, which `mobile/mise.toml` pinned at the 1.0.17 release commit `ca8effcf9` and still pins today:

| Frame | Source line |
| --- | --- |
| `framework.dart:5938` | `State<StatefulWidget> get state => _state!;` |
| `framework.dart:6046` | `_state = null;` — the only assignment, inside `StatefulElement.unmount()` |
| `navigator.dart:2919` | `if (context case StatefulElement(:final NavigatorState state))`; the `findAncestorStateOfType` walk six lines below is the other candidate, since AOT collapses the inlined frame |
| `ink_well.dart:1224` | `widget.onTap?.call()` |
| `recognizer.dart:345` | `result = callback();`, inside the try/catch that reports instead of rethrowing |

`mobile/lib/screens/curated_list_feed_screen.dart:419` is byte-identical at `ca8effcf9`:

```dart
TextButton(
  onPressed: () => Navigator.of(dialogContext).pop(false),   // <- line 419
```

Since `_state` is nulled only in `unmount()`, the element `Navigator.of` reached was **defunct**: the dialog's route had been torn down before the tap was delivered.

Reproduced in a widget test — capture the button's `onPressed`, unmount the tree, then invoke it. Debug builds report the same walk as `Looking up a deactivated widget's ancestor is unsafe`; release AOT compiles that assert out and hits the `_state!` null check instead.

## What tore the route down — open

Not determined from a single session export.

The leading candidate is the app-shell `if (peopleListsEnabled) BlocProvider(...)` entry in `main.dart`'s `MultiBlocProvider`. Flipping the flag changes the provider chain's shape and re-inflates everything below it, which is exactly what #6476 fixed ("stop the curated-lists toggle from popping the screen"). That entry is still present at 1.0.17 (`ca8effcf9`) **and at 1.0.18** (`b5c532896`) — `a087710e6` landed after the 1.0.18 cut, so it first ships in 1.0.19.

Two things argue against treating that as settled:

- #6476's own commit message records that the step from re-inflation to route loss is not pinned.
- Feature flags are local (SharedPreferences plus build config), so a flip needs a Settings or developer-mode toggle rather than a background sync — which makes "user toggled the flag while this dialog was open" an awkward sequence.

The guard below is correct for any teardown, so this does not block the fix.

## Fix

Guard both dialog actions on `dialogContext.mounted` before the `Navigator.of` walk. Behaviour is unchanged while the route is alive; when it is gone the pop was moot anyway, and `showDialog` already completes with `null`, which `_confirmDeleteList` treats as "not confirmed".

## Same shape elsewhere

Nine other files use an unguarded `Navigator.of(dialogContext)` / `Navigator.of(sheetContext)` inside a dialog or sheet callback. None has a report against it, so they are out of scope here and listed for whoever wants to decide whether this should become a shared helper:

`relay_settings_screen.dart`, `user_list_people_screen.dart` (two dialogs), `settings/nip05_settings_screen.dart`, `inbox/conversation/widgets/reaction_picker_overlay.dart`, `profile_setup/widgets/profile_avatar_section.dart`, `profile_setup/widgets/profile_setup_listeners.dart`, `widgets/owner_video_delete_confirmation_dialog.dart`, `widgets/library/drafts_tab.dart`.

realmeylisdev:
## Description

The **Cancel** and **Delete** buttons of the delete-curated-list confirmation dialog called `Navigator.of(dialogContext)` unguarded. When the dialog's route is already torn down by the time the tap is delivered, that ancestor walk hits a defunct element and throws — the pop never happens and the tap does nothing.

Nothing crashes: `GestureRecognizer.invokeCallback` catches it and hands it to `FlutterError.reportError`, which `crash_reporting_service.dart` forwards to `recordFlutterFatalError`. So a no-op tap is filed in Crashlytics as a fatal, competing with real crashes in triage.

Both actions now go through `_popConfirmation`, which returns early when `dialogContext.mounted` is false. Behaviour is unchanged while the route is alive; when it is gone the pop was moot anyway, and `showDialog` already completes with `null`, which `_confirmDeleteList` treats as "not confirmed".

Frame-by-frame resolution of the stack against the pinned Flutter 3.44.0, and why the teardown trigger itself is still open, are in the issue.

**Related Issue:** Closes #6512
**Follow-up:** #6526 tracks the sibling modal/sheet call sites and the shared-helper decision.

## Out of Scope

- **What tore the route down.** Not determined from a single session export. The leading candidate — the app-shell `if (peopleListsEnabled)` provider entry that re-inflates everything below it — was already removed by #6476, which lands in 1.0.19; the crashing 1.0.17 and 1.0.18 builds both still carry it. #6476's own commit message records that the step from re-inflation to route loss is not pinned. The guard here is correct for any teardown, so it does not wait on that.
- **Sibling modal/sheet call sites.** Nine sibling sites use the same unguarded `Navigator.of(dialogContext)` / `Navigator.of(sheetContext)` shape. None has a report against it. Whether this should become a shared helper across all of them is a call worth making deliberately rather than as a side effect of this fix, so the inventory now lives in #6526.

## Verification

- `dart format --set-exit-if-changed` on both changed files — clean.
- `flutter test test/screens/curated_list_feed_screen_test.dart` — 12/12.
- Neighbouring suites that touch this screen: `main_list_deep_link_nav_action_test.dart`, `l10n/hardcoded_visible_strings_test.dart`, `screens/curated_list_by_author_screen_test.dart`, `router/route_coverage_test.dart` — 119/119.
- The new regression test was confirmed red with the guard removed: it fails with `Looking up a deactivated widget's ancestor is unsafe`, which is the debug-build form of the same ancestor walk that produces the `_state!` null check in release AOT.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/screens/curated_list_feed_screen.dart
+++ b/mobile/lib/screens/curated_list_feed_screen.dart
@@ -23,6 +23,18 @@ import 'package:unified_logger/unified_logger.dart';
 
 enum _CuratedListAction { delete, unfollow }
 
+/// Pops the delete-confirmation dialog with [confirmed].
+///
+/// The route can be gone by the time a tap on one of its buttons is
+/// delivered. `Navigator.of` then walks a defunct element and throws, which
+/// `FlutterError.onError` records as a fatal even though the pop is moot.
+void _popConfirmation(BuildContext dialogContext, {required bool confirmed}) {
+  if (!dialogContext.mounted) {
+    return;
+  }
+  Navigator.of(dialogContext).pop(confirmed);
+}
+
 class CuratedListFeedScreen extends ConsumerStatefulWidget {
   /// Route name for this screen.
   static const routeName = 'list';
@@ -416,14 +428,14 @@ class _CuratedListFeedScreenState extends ConsumerState<CuratedListFeedScreen> {
         ),
         actions: [
           TextButton(
-            onPressed: () => Navigator.of(dialogContext).pop(false),
+            onPressed: () => _popConfirmation(dialogContext, confirmed: false),
             child: Text(
               l10n.commonCancel,
               style: VineTheme.labelMediumFont(color: VineTheme.secondaryText),
             ),
           ),
           TextButton(
-            onPressed: () => Navigator.of(dialogContext).pop(true),
+            onPressed: () => _popConfirmation(dialogContext, confirmed: true),
             child: Text(
               l10n.commonDelete,
               style: VineTheme.labelMediumFont(color: VineTheme.error),
--- a/mobile/test/screens/curated_list_feed_screen_test.dart
+++ b/mobile/test/screens/curated_list_feed_screen_test.dart
@@ -245,6 +245,32 @@ void main() {
       expect(find.text(l10n.curatedListUnfollowAction), findsNothing);
     });
 
+    testWidgets('delete cancel dismisses without deleting', (tester) async {
+      when(() => mockService.isSubscribedToList('owned-list')).thenReturn(true);
+      when(() => mockService.isOwnedList('owned-list')).thenReturn(true);
+
+      await tester.pumpWidget(
+        buildSubject(
+          listId: 'owned-list',
+          listName: 'Owned List',
+          authorPubkey: 'owned-pubkey',
+        ),
+      );
+      await tester.pump();
+
+      await tester.tap(find.byTooltip(l10n.curatedListActionsTooltip));
+      await tester.pumpAndSettle();
+      await tester.tap(find.text(l10n.listDeleteAction));
+      await tester.pumpAndSettle();
+      expect(find.text(l10n.curatedListDeleteConfirmTitle), findsOneWidget);
+
+      await tester.tap(find.text(l10n.commonCancel));
+      await tester.pumpAndSettle();
+
+      verifyNever(() => mockService.deleteOwnedList('owned-list'));
+      expect(find.text(l10n.curatedListDeleteConfirmTitle), findsNothing);
+    });
+
     testWidgets(
       'delete confirmation buttons no-op once their route is torn down',
       (tester) async {
@@ -287,7 +313,6 @@ void main() {
 
         expect(cancel, returnsNormally);
         expect(confirm, returnsNormally);
-        verifyNever(() => mockService.deleteOwnedList('owned-list'));
       },
     );
 
--- a/mobile/test/screens/curated_list_feed_screen_test.dart
+++ b/mobile/test/screens/curated_list_feed_screen_test.dart
@@ -245,6 +245,52 @@ void main() {
       expect(find.text(l10n.curatedListUnfollowAction), findsNothing);
     });
 
+    testWidgets(
+      'delete confirmation buttons no-op once their route is torn down',
+      (tester) async {
+        when(
+          () => mockService.isSubscribedToList('owned-list'),
+        ).thenReturn(true);
+        when(() => mockService.isOwnedList('owned-list')).thenReturn(true);
+
+        await tester.pumpWidget(
+          buildSubject(
+            listId: 'owned-list',
+            listName: 'Owned List',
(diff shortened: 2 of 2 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I took over the final review cleanup on this PR.

Changes pushed in d8df95d3d:
- added a live-route regression test that Cancel dismisses the delete dialog without calling delete
- removed the misleading dead-route `verifyNever(deleteOwnedList)` assertion, leaving the `returnsNormally` checks as the actual crash regression coverage
- opened #6526 for the sibling modal/sheet `Navigator.of(dialogContext/sheetContext).pop(...)` inventory and linked it from the PR body so that debt survives #6512 closing

Verification:
- `flutter test test/screens/curated_list_feed_screen_test.dart` — 12/12
- push hook reran analyze and the changed-file test — passed
- GitHub checks are green on `d8df95d3d`

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup. The mounted-context guard is correct for the reported dead-route crash path, Cancel polarity is now pinned by a live-route test, and sibling modal/sheet cleanup is tracked in #6526 instead of disappearing with #6512. CI is green on d8df95d3d.

PR #6288, opened 2026-07-21
fix(feed): sanitize unpaired UTF-16 surrogates in display text

rabble:
## Summary

Companion to the merged #6273 (which closed #6111 by making `stripZalgo` normalize unpaired UTF-16 surrogates to U+FFFD). This PR was written in parallel against the same Crashlytics issue and has been rebased onto #6273; it now carries the **residual pieces #6273 did not cover**:

- **`LinkifiedText` / `SelectableLinkifiedText` sanitize their input.** Comments and bios reach these widgets as raw event content — they bypass the model display getters #6273 hardened, so a lone surrogate in a comment still crashed paragraph layout on `main`.
- **`sanitizeForDisplay()`** convenience in `text_sanitizer` (`sanitizeUtf16` + `stripZalgo`) with tests.
- **`StringUtils.sanitizeUtf16` (DM path) delegates to the shared package implementation** instead of keeping a second divergent copy; the DM `MessageBubble` test is aligned to the U+FFFD replacement semantics #6273 standardized.
- **Surrogate regression pins** on `UserProfile.bestDisplayName`, `ProfileSearchResult.bestDisplayName`, and notification actor names, plus new widget regression tests in `mobile/test/widgets/linkified_text/linkified_text_utf16_test.dart` (verified red on pre-fix code).

## Testing

- `text_sanitizer` (23), `models` display suites (153), `notification_repository` (126), linkified/DM/string-utils widget suites — all green after the rebase.
- `flutter analyze lib test integration_test` clean.
- `text_sanitizer` keeps 100% line coverage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6515, opened 2026-07-30
fix(profile): sanitize display text against malformed UTF-16

### LINKED ISSUE #6516: fix(feed): sanitize mention labels against malformed UTF-16
Auto-created from PR #6515.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6515
**Branch:** `fix/linkified-mention-label-surrogates`

hm21:
Closes #6516

## Description

An unpaired UTF-16 surrogate makes Flutter's paragraph builder throw `Invalid argument(s): string is not well-formed UTF-16` during layout. That is a crash, not a rendering glitch — it is the top FATAL issue in Crashlytics, 76 events across 11 users, flagged `SIGNAL_REPETITIVE` at ~10 crashes per affected user.

#6288 sanitized the text handed to `LinkifiedText`, but the crash survived it, because the span builder also injects text that never came from there: the label for a `nostr:npub…` mention, resolved from profile metadata. So a single profile whose name carries a lone surrogate crashes every viewer who sees a caption or comment mentioning them — and keeps crashing them, because that profile keeps reappearing in the feed.

The stack pins the surface exactly. Three nested `TextSpan.build` frames (298 → 316 → 316) only occur for `Text.rich` with a span that has children, which is the shape `LinkifiedText` builds — a plain `Text` produces one frame, a flat `Text.rich` two. Sample events confirm it: all of them crash on `pooled_video_feed` or with the comments sheet open, none on a notifications screen.

Fixed at both ends:

- `profileDisplayText` sanitizes the name it picks. It reads the raw metadata fields rather than the sanitizing accessors (`bestDisplayName` and friends) because it keeps its own precedence — display name, then name, then NIP-05 — so it has to sanitize the winner itself rather than delegate.
- `LinkifiedTextSpanBuilder.build()` sanitizes too. That is the one point every span passes through, so no future caller has to remember, and it covers the injected label as well as the source text. Replacement is code-unit-for-code-unit, so match offsets stay valid. The constructor stays `const` (existing tests depend on it), which is why this lives in `build()` rather than the initializer.

This also closes `notification_comment_quote.dart`, which called the builder directly and was the only consumer without its own sanitizing wrapper.

The DM conversation preview is fixed separately: it renders through plain `Text`/`Text.rich` rather than the span builder, so it does not inherit that guard and sanitizes its own sender-controlled content once, ahead of every split.

### Closing the rest of the same crash class

Review turned up that the mention label was not the last hole — the same fatal reaches a plain `Text` through several sibling paths that never touch the span builder. Rather than patch each render site, this fixes it where the data is shaped.

**The model rule (`fix(profile)`).** `UserProfile` exposes raw kind-0 fields (`name`, `displayName`, `nip05`) and a set of *display* getters. `bestDisplayName` and `betterDisplayName` already sanitized; `handle`, `displayNip05` and `shortDisplayNip05` did not, despite existing only to be rendered. `handle` alone reaches the DM conversation header, the new-message sheet and people-list rows — so the inbox row survived this PR's first commit while opening the thread still crashed. Fixing the three getters covers every consumer at once. NIP-05 verification compares by pubkey rather than by these strings, and mention matching normalizes to `[a-z0-9]`, so neither is affected. The raw fields stay raw for editors, matching, and serialization; that split is now the rule.

**The remaining render sites (`fix(comments)`).** `comment_item` and `comments_screen` built a chain character-for-character identical to `bestDisplayName`, so they swap to it directly. `mention_overlay` and `mention_search` sanitize in place instead — `bestDisplayName` substitutes a generated name for a profile that has none, which would have shadowed the npub fallback and made `mention_search`'s null guard always true, matching generated names against the query. `ShareableUser.displayName` had two of five construction sites still on raw fields; both move to `bestDisplayName`, covering find-people, share-with, and the recipient announcement in one place.

Profile editors are deliberately untouched. `profile_editor_bloc`, `monetization_links_settings_cubit` and `profile_setup_listeners` seed a `TextField` with the user's *own* name, where sanitizing would silently rewrite it on the next publish.

**Related Issue:** 

## Out of Scope

Two other issues from the same Crashlytics pass are left for their own PRs: the `hashmap.cc` SIGABRT (23 users — a `cupertino_http` FFI callback firing after isolate teardown in the background) and `SqliteException(26): file is not a database` (12 users — the encrypted Drift DB failing to open).

`UserProfile.sanitizeDisplayName` is `stripZalgo`, so it also caps combining characters at two per base character. For mention labels that matches how display names are treated everywhere else and keeps a Zalgo name from blowing up feed layout. If we ever want the narrower behaviour for mentions specifically, `StringUtils.sanitizeUtf16` is the surgical alternative.

The call-site swaps in `fix(comments)` are covered by the existing suites rather than by pinning tests of their own; the model getters and the DM preview are pinned directly.

## Verification

- Three new tests in `linkified_text_span_builder_test.dart`. The two regression tests **fail without the fix** — verified by stashing the builder change and re-running (`+11 -2`). The third pins that valid surrogate pairs (emoji) survive untouched, so the guard cannot regress into over-sanitizing.
- One new test in `conversation_tile_test.dart` for the DM preview, and six in `user_profile_test.dart` for `handle` / `displayNip05` / `shortDisplayNip05`. Both were mutation-tested: reverting the DM preview sanitize fails exactly that one test (`+30 -1`), reverting the `handle` sanitize fails exactly its two (`+12 -2`), and nothing else moves.
- `models` 834/834; comments + composer 208/208; sharing, find-people, settings, search, watermark 262/262; inbox, linkified text, people-lists, mention resolution, delete-account 561/561.
- `dart format` — no changes

Manual reproduction is impractical: it needs a real profile whose name contains an unpaired surrogate. The pinned tests are the gate here; the field confirmation is the `_NativeParagraphBuilder.addText` signature disappearing from Crashlytics after release.

What device testing should cover is the *absence* of regression for well-formed input, since the sweep touches display names on a lot of surfaces at once: feed, comments sheet, DM inbox and conversation header, share sheet, find-people, and people-lists.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/comments/comment_composer/mention_search.dart
+++ b/mobile/lib/blocs/comments/comment_composer/mention_search.dart
@@ -1,3 +1,4 @@
+import 'package:models/models.dart' show UserProfile;
 import 'package:profile_repository/profile_repository.dart';
 
 /// A single mention match, as a record so this helper has no dependency on
@@ -37,7 +38,10 @@ Future<({List<MentionMatch> matches, Set<String> seen})> mentionSearchLocal({
         displayName.toLowerCase().contains(lowercaseQuery)) {
       matches.add((
         pubkey: pubkey,
-        displayName: displayName,
+        // Matched on the raw value so query behaviour is unchanged, but
+        // carried sanitized: this string is rendered in the suggestion list
+        // and inserted into the composer on select.
+        displayName: UserProfile.sanitizeDisplayName(displayName),
         picture: profile?.picture,
         nip05: profile?.nip05,
       ));
@@ -73,7 +77,7 @@ Future<List<MentionMatch>> mentionSearchRemote({
     if (name == null) continue;
     merged.add((
       pubkey: profile.pubkey,
-      displayName: name,
+      displayName: UserProfile.sanitizeDisplayName(name),
       picture: profile.picture,
       nip05: profile.nip05,
     ));
--- a/mobile/lib/screens/comments/comments_screen.dart
+++ b/mobile/lib/screens/comments/comments_screen.dart
@@ -741,8 +741,7 @@ class _MainCommentInputState extends ConsumerState<MainCommentInput> {
                 .watch(userProfileReactiveProvider(replyToAuthorPubkey))
                 .value;
             replyToDisplayName =
-                profile?.displayName ??
-                profile?.name ??
+                profile?.bestDisplayName ??
                 UserProfile.generatedNameFor(replyToAuthorPubkey);
           }
         }
--- a/mobile/lib/screens/comments/widgets/comment_item.dart
+++ b/mobile/lib/screens/comments/widgets/comment_item.dart
@@ -662,10 +662,11 @@ class _ReplyIndicator extends ConsumerWidget {
         .watch(userProfileReactiveProvider(parentAuthorPubkey))
         .value;
 
-    // Get display name with fallback chain
+    // bestDisplayName is the same displayName -> name chain, but sanitized:
+    // a lone UTF-16 surrogate in either field crashes the paragraph builder
+    // when this renders.
     final displayName =
-        profile?.displayName ??
-        profile?.name ??
+        profile?.bestDisplayName ??
         UserProfile.generatedNameFor(parentAuthorPubkey);
 
     return Row(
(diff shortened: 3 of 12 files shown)


rabble (approved):
Reviewed the current head and changed call sites. The UTF-16 guards cover source text, injected profile labels, DM previews, display getters, mention suggestions, and sharing paths; the added regression coverage matches those boundaries. CI is green. No blocking findings.

Issue #34, opened 2025-06-19
Epic 4: Video Status Polling Endpoint


rabble:
# Epic 4: Video Status Polling Endpoint

## Overview
Create endpoint for clients to poll video processing status with UUID-based security and optimized caching.

## Acceptance Criteria
- [ ] `GET /v1/media/status/:videoId` endpoint implemented
- [ ] UUID validation to prevent enumeration attacks
- [ ] Conditional responses based on video status
- [ ] Cloudflare Images URL transformation for thumbnails
- [ ] Optimized caching strategy by status type
- [ ] Client polling guidelines with exponential backoff
- [ ] Integration tests covering all status states

### Endpoint
```
GET /v1/media/status/:videoId
```

### Path Parameters
- `videoId`: UUID v4 of the video to check status

#### Published Video (200 OK)
```json
{
  "status": "published",
  "hlsUrl": "https://videodelivery.net/a3b8c1f4.../manifest/video.m3u8",
  "dashUrl": "https://videodelivery.net/a3b8c1f4.../manifest/video.mpd", 
  "thumbnailUrl": "https://imagedelivery.net/account-hash/a3b8c1f4.../w=400,h=300",
  "createdAt": "2024-01-01T12:00:00.000Z"
}
```

#### Processing States (200 OK)
```json
{
  "status": "pending_upload"
}
```

```json
{
  "status": "processing"
}
```

#### Terminal States (200 OK)
```json
{
  "status": "quarantined"
}
```

```json
{
  "status": "failed",
  "error": "Processing timeout - please try uploading again"
}
```

### Error Responses
```json
// 400 Bad Request
{
  "error": {
    "code": "invalid_video_id",
    "message": "Video ID must be a valid UUID"
  }
}

// 404 Not Found
{
  "error": {
    "code": "video_not_found", 
    "message": "Video not found"
  }
}

// 429 Too Many Requests (optional)
{
  "error": {
    "code": "rate_limit_exceeded",
    "message": "Too many status requests, please use exponential backoff"
  }
}
```

### Worker Logic Flow
```javascript
// src/status-check-worker.js
export default {
  async fetch(request, env, ctx) {
    try {
      // Extract videoId from URL path
      const url = new URL(request.url);
      const videoId = url.pathname.split('/').pop();
      
      // Validate UUID format (security against enumeration)
      if (\!isValidUUID(videoId)) {
        return errorResponse('invalid_video_id', 'Video ID must be a valid UUID', 400);
      }
      
[...]

PR #6517, opened 2026-07-31
feat(notifications): notify me when a creator I follow posts

### LINKED ISSUE #6518: feat(notifications): notify me when a creator I follow posts
Auto-created from PR #6517.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6517
**Branch:** `design/bell-notifications`

rabble:
Closes #6518

## Description

Users asked to be notified when a creator they care about posts. This is the mobile half: a bell on the profiles of people you follow, backed by a NIP-51 people list, plus the notification plumbing to receive and route the resulting push.

**The whole feature ships behind `FF_NEW_POST_NOTIFICATIONS`, default off.** The delivery half lives in `divine-push-service` and has not landed ([divinevideo/divine-push-service#34](https://github.com/divinevideo/divine-push-service/pull/34) is the plan). Merging the UI live would put a bell on every profile that publishes a subscription nothing reads — a control that looks like it works and silently does nothing. Same pattern as `feedTuning` and `publishDmRelayList`.

Commits are separable and each stands alone:

**`fix(people-lists)`: reserve `d=notify`** — the load-bearing one. Kind 30000 lists are decoded into user-editable `UserList`s. Without reserving the `d` tag, a user's bell subscriptions would appear in the curated-lists UI as an ordinary list they could rename or delete, silently unsubscribing themselves from everyone. `d=block` was already reserved; this adds `notify` alongside it.

**`feat(notifications)`: publish kind 34236 in push preferences** — the preferences event enumerates which kinds the push service may deliver. Without 34236 in that list the feature is structurally undeliverable no matter what the service does.

**`feat(notifications)`: add the `newPost` kind** — the enum value, the icon, the cache/server type strings, and tap routing to the video. While wiring this I found and fixed a latent bug: `_mapActorAnchored` had a hand-copied duplicate of `_isVideoAnchoredKind`'s skip list, so every `newPost` row would have rendered twice. It now calls the predicate.

**`feat(people-lists)`: the `d=notify` repository** — reads and mutates the subscription list. Two things here are not obvious and are pinned by tests:

- All bells live in *one* replaceable event, so mutations are serialized through a single per-owner queue. Per-pubkey serialization would still race. The queue tail deliberately swallows errors — otherwise one failed publish poisons every mutation queued behind it.
- `created_at` steps past the previous event rather than using wall-clock seconds. NIP-33 leaves same-second tie-breaking unspecified, so two mutations inside one second can resolve to either.

**`feat(profile)`: the follow-gated bell** — optimistic toggle with revert on failure. Unfollow tears the subscription down, bridged from a host *above* the bell, because the bell has already unmounted by the time the follow state flips.

**`fix(notifications)`: republish preferences after a kinds-list change** — an upgrading user's stored preferences read back with `newPostsEnabled` on locally while the push service still holds the pre-upgrade kind list. Only an explicit user edit marked preferences dirty, so nothing ever republished. A persisted schema-version marker now triggers one republish on the first ready session after upgrade. I originally described this as self-healing in a commit message; it wasn't, and this commit is the fix.

**`feat(notifications)`: the feature flag** — gates the bell, the settings row, and `NotifyBellCubit` construction, so a flagged-off build never reads the subscription list from the relay.

**Related Issue:** Relates to divinevideo/divine-push-service#34

## Out of Scope

- **Delivery.** `divine-push-service` fans kind 34236 out to `d=notify` subscribers; tracked in that repo's PR #34, unimplemented.
- **The in-app notification row.** FunnelCake has the contract but no owner. It would be the inbox's first subscription-driven fan-out — every existing source reads recipients off tags on the event, whereas new-post rows have no `p` tag.
- **Reserved-tag squatting.** A pre-existing `d=notify` list this app didn't write is currently overwritten rather than detected. Noted in the design doc.
- **Copy and translations.** The 6 new ARB keys × 17 locales are mine and unreviewed — they need a brand pass and native-speaker review before the flag is flipped on.

## Verification

- 555 tests across `test/l10n`, `test/core/feature_flag_test.dart`, `test/features/feature_flags`, `test/screens/notification_settings_screen_test.dart`, `test/widgets/profile`, `test/blocs/notify_bell`, `test/blocs/notification_settings`, `test/services/notification_preferences_service_test.dart`, `test/providers/push_notification_sync_test.dart`, `test/notifications`
- `packages/people_lists_repository` — 76 tests, coverage 96.57% vs its 95 gate
- `packages/models` — 829 tests; `packages/notification_repository` — 160 tests
- Two tests were checked against the can-it-fail bar by breaking the code and confirming they go red: the flag gate, and the mutation-serialization test (which initially passed with the queue removed and was strengthened until it didn't)

Rebased onto fresh `origin/main`; the l10n and `repository_providers` conflicts against #6480 were resolved keeping both sides, and generated l10n was regenerated rather than hand-merged.

## Type of Change

### DIFF
--- a/docs/plans/2026-07-29-bell-notifications-design.md
+++ b/docs/plans/2026-07-29-bell-notifications-design.md
@@ -1,0 +1,339 @@
+# Design: bell notifications (subscribe to a creator's new posts)
+
+Status: design, not yet approved
+Date: 2026-07-29
+Companion work: `divine-push-service` PR #34 / issue #33, `divine-funnelcake` (in-app row)
+
+## Problem
+
+Every notification Divine sends today is *"someone acted on your content"* —
+like, comment, repost, mention. Users have asked for the inverse: **tell me
+when a creator I care about posts**. That is a subscription to someone else's
+output, and nothing in the current model supports it.
+
+This doc covers the mobile half: the bell UI, the subscription list we publish,
+the preference change that makes delivery possible, and how a new-post push is
+parsed and routed on tap. It also states the contract `divine-funnelcake` needs
+to honor for the in-app row.
+
+The service half is specified in `divine-push-service`
+`docs/plans/new-post-notifications.md`. Read it first; this doc assumes it.
+
+## Protocol shape (agreed with the service)
+
+The subscription list is a NIP-51 people list with a reserved `d` tag,
+published by this client and read by the push service:
+
+```
+kind 30000
+["d", "notify"]
+["title", "Notify"]
+["p", "<creator-pubkey-hex>"]   × N
+```
+
+Replaceable, public, portable. The service maintains a `creator → [subscribers]`
+reverse index from it.
+
+---
+
+## Three findings that block the plan as written
+
+I verified the service plan's mobile-side assumptions against `origin/main`
+(`9e16e73a2`). Three of them do not hold. Each is a silent failure — nothing
+crashes, the feature just does nothing.
+
+### 1. `d=notify` collides with the user-facing People Lists UI — bells are user-deletable
+
+`Nip51PeopleListCodec.decode`
+(`mobile/packages/people_lists_repository/lib/src/nip51_people_list_codec.dart:95`)
+turns any kind 30000 event into a user-editable `UserList`. It excludes exactly
+one reserved identifier:
+
+```dart
+static const String blockedDTag = 'block';        // :51
+...
+if (dTag == null || dTag == blockedDTag) {        // :101
+  return null;
+}
+```
+
+A `d=notify` list is not excluded. It would decode into a `UserList` named
+"Notify" and appear in the People Lists UI alongside the user's own lists —
+renameable, editable, deletable. A user tidying their lists silently wipes every
+bell they have set, and the app gives no indication that is what happened.
+
+**Fix:** replace the single constant with a reserved set, and exclude on
+membership:
+
+```dart
+/// Reserved `d` tag values for app-managed kind 30000 lists.
+///
+/// These are excluded from the user-facing list collection so app-managed
+/// lists cannot be edited, renamed, or deleted as ordinary people lists.
+static const Set<String> reservedDTags = {blockedDTag, notifyDTag};
+
+static const String blockedDTag = 'block';
+static const String notifyDTag = 'notify';
+```
+
+`blockedDTag` stays as a named constant — it is referenced elsewhere and in the
+package's public docs (`people_lists_repository.dart:81`). The docs on `decode`
+and the codec's ABOUTME line both need updating; they name `d=block` specifically.
+
+This is the load-bearing fix. Without it the feature is actively destructive.
+
+### 2. `NotificationPreferences` structurally cannot express kind 34236
+
+The service gates delivery on the kind list published in the kind-3083
+preferences event. `NotificationPreferences`
+(`mobile/lib/models/notification_preferences.dart`) is five fixed booleans, and
+`toKindsList()` can only ever emit a subset of `{1, 3, 7, 16}`:
+
+```dart
+List<int> toKindsList() {
+  final kinds = <int>{};
+  if (likesEnabled) kinds.add(7);
+  if (commentsEnabled || mentionsEnabled) kinds.add(1);
+  if (followsEnabled) kinds.add(3);
+  if (repostsEnabled) kinds.add(16);
+  return kinds.toList()..sort();
+}
+```
+
+`push_notification_service.dart:254` publishes exactly this list. So **34236 is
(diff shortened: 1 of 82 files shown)


rabble:
Review findings — changes needed before merge

1. Blocking data-loss path: mobile/packages/people_lists_repository/lib/src/notify_subscriptions_repository_impl.dart:239-273 converts any relay read failure into a cached empty snapshot. NotifyBellCubit then receives that empty set as ready at mobile/lib/blocs/notify_bell/notify_bell_cubit.dart:45-58, enables the bell, and a tap publishes a replacement list containing only the tapped creator. A transient relay outage can therefore overwrite every existing d=notify subscription. Preserve an unknown/error state and keep the control disabled until a successful read, or retry without caching the empty result. Add a regression test that makes the read fail and verifies no replacement publish is allowed.

2. Blocking unfollow cleanup: mobile/lib/blocs/notify_bell/notify_bell_cubit.dart:122-137 ignores the PeopleListPublishResult returned by unsubscribe. The repository returns failed for a non-throwing publish failure at notify_subscriptions_repository_impl.dart:196-221, so an unfollow can leave the creator in the public notify list and continue delivering notifications with no retry or user-visible failure. Inspect the result and provide a durable retry path (the comment says the next profile visit retries, but watchSubscriptions reuses the keep-alive cached snapshot at notify_subscriptions_repository_impl.dart:63-70 and does not re-read).

3. The new reserved d=notify publisher does not detect a pre-existing list from another client before replacing it (notify_subscriptions_repository_impl.dart:169-198). The design document itself calls this out as data loss. Please either protect foreign/pre-existing lists and surface a conflict, or document and obtain the explicit product decision to overwrite them.

Non-blocking correctness note: mobile/lib/notifications/widgets/video_notification_row.dart:262-266 omits NotificationKind.newPost from _typeShowsTitle, so new-post rows silently drop the video title even though the repository loads videoTitle.

rabble:
Please rebase this branch onto the current main (`3bd867b5`) before re-running CI. The PR base is still `f2d12fb5`, so the green checks validate the older base rather than the current merge result.

hm21 (approved):
Took over the three blocking findings plus the non-blocking one, one commit each, and rebased onto current `main` — base is now `be2a8421`. Finding 3 is deliberately untouched; it needs a product call, not a patch.

**`4a9807eb` — unread base (finding 1).** Worse than reported: it doesn't need a thrown read to trigger. `NostrClient.queryEvents` discards the `noRelays`/`timedOut` flags that `queryEventsDetailed` exposes, so an offline or cold-start read returns `[]` *without throwing* — airplane mode alone reaches the destructive path, and a try/catch would not have covered it. Two things made it unrecoverable: `refresh()` has no callers anywhere in the app (and cached a failed read over a good snapshot, with no newer-than guard), and `_subjectFor` seeded the subject with `{}` before the first read, so `NotifyBellStatus.initial` never actually occurred against the real repository — the bell was tappable immediately. Reads now go through `queryEventsDetailed`, "unknown" is a distinct state that is never cached and never published against, and the stream stays silent so the bell holds `initial`. The tradeoff is deliberate: the bell is untappable while relays are unreachable.

**`92e9d9ce` — teardown (finding 2).** Confirmed the code comment was wrong — `load()` only replays the cached snapshot, nothing re-reads. A failed removal is now retained per owner in the repository and folded into the desired set of the next publish; `reconcile()` drains it on its own, and the cubit calls it from `load()`, so any bell mount for that viewer clears an outstanding removal whichever creator it belonged to. Retention is in-memory on the keep-alive repository: survives navigation, not an app restart. Tell me if you want it persisted.

**`94f1efe4` — `_typeShowsTitle`.** One line.

**`85ae7e5e` — l10n.** This is what the stale base was hiding. #6495 added Vietnamese, Urdu, Chinese and Malay after you branched; rebasing brings them in without the six bell keys, which fails `arb_consistency_test` *and* breaks the build outright — the generated `AppLocalizations` subclasses stop implementing every abstract member. Translated rather than English-filled, but they want the same native pass as your other 17 before the flag flips.

**Left for you — finding 3.** Your design doc already proposes the mitigation (skip a `d=notify` list carrying a `title` we didn't write), but what the bell should then *do* is a behaviour decision: silently no-op, or tell the user. Same for the doc's other open item — bell subscriptions are world-readable and nothing in the UI says so. Both still open, both yours.

**Validation.** `flutter analyze lib test integration_test` clean; 655 tests across the affected mobile suites; `people_lists_repository` 89 tests at 96.55% vs its 95 gate. All 17 CI checks green on `85ae7e5e`. The new repository tests were checked against the can-it-fail bar by reverting the guard — the unread-base ones go red.

Yours to review and merge, or push back on any of it.

PR #6521, opened 2026-07-31
fix(snackbar): stop actions that outlive their screen reading a dead context

### LINKED ISSUE #6520: fix(snackbar): actions that outlive their screen throw on a dead BuildContext
## Summary

Snackbar actions that call `context.read<T>()` throw when tapped after their screen is gone, because the root `ScaffoldMessenger` outlives the route. The CTA silently does nothing and files a Crashlytics report.

## Environment

- App version: 1.0.17 (813), reproduced by inspection on `main` (1.0.19)
- Platform: iOS (Crashlytics `apple`); the mechanism is platform-independent
- Crashlytics issue: `a8c821212ea690bd4ecf96e4a9d696c9`, titled `Provider._inheritedElementOf`

## Steps to Reproduce

**Site 1 — Bluesky crosspost settings** (this is the sampled session)

1. Sign in with an account that has no claimed Divine username.
2. Open Settings → Bluesky publishing.
3. Toggle "Publish videos to Bluesky" on. The toggle fails with `usernameNotClaimed` and a snackbar appears with a **Set up handle** action.
4. Tap the back button before the snackbar auto-dismisses.
5. Tap **Set up handle** on the still-visible snackbar.

**Site 2 — DM reel reply bar**

1. Open a reel shared into a DM, so the in-player reply bar is shown.
2. Send a text reply while sending will fail (e.g. offline). A snackbar appears with a **Retry** action.
3. Scroll to the next reel or leave the player, disposing the bar.
4. Tap **Retry** (or **View chat** on the success variant).

## Actual Result

The action does nothing. A `Fatal Exception: FlutterError` is recorded in Crashlytics with either message, depending on how far element teardown had progressed:

- `Null check operator used on a null value`
- `Could not find the correct Provider<T> above this <Widget> Widget`

Both share the same top frames, which is why Crashlytics merged them into one issue:

```
0  Element.widget + 3653 (framework.dart:3653)
1  Provider._inheritedElementOf + 377 (provider.dart:377)
2  Provider.of + 327 (provider.dart:327)
3  ReadContext.read + 683 (provider.dart:683)
4  _BlueskySettingsView._onStateChanged.<fn> + 109 (bluesky_settings_screen.dart:109)
5  _SnackBarActionState._handlePressed + 153 (snack_bar.dart:153)
...
7  GestureRecognizer.invokeCallback + 345 (recognizer.dart:345)
```

The app does **not** crash. `GestureRecognizer.invokeCallback` catches the error and routes it to `FlutterError.onError`, which `crash_reporting_service.dart:45` forwards via `recordFlutterFatalError` — so it is labelled fatal but the process survives. The user-visible defect is a dead CTA.

## Expected Result

Tapping a stale snackbar action does nothing and reports nothing. Tapping a live one works as before.

## Root Cause

`provider.dart:377` (provider 6.1.5+1) is:

```dart
throw ProviderNotFoundException(T, context.widget.runtimeType);
```

Reaching that line means `context.getElementForInheritedWidgetOfExactType<...>()` returned null, i.e. `_inheritedElements?[T]` was null.

`Element._ensureDeactivated()` (framework.dart:4820), which always runs immediately after `Element.deactivate()`, does `_inheritedElements = null`. **Every** provider lookup on a deactivated element therefore returns null, regardless of what is actually in the widget tree. Two outcomes along the teardown timeline:

| Element state | Result |
|---|---|
| Deactivated, still mounted | Real `ProviderNotFoundException`, naming the widget |
| Unmounted (`_widget == null`) | `context.widget` throws first — `Widget get widget => _widget!` at framework.dart:3653 — masking the ProviderNotFoundException |

Neither message means the provider is missing from the tree. Both mean **a provider lookup on a dead `BuildContext`**.

The trigger is that the repo sets no `scaffoldMessengerKey` and installs no app-level `ScaffoldMessenger`, so `MaterialApp`'s built-in one — which sits above the `Navigator` — hosts every snackbar. Snackbars migrate to the next route's `Scaffold` on a pop and stay tappable after the screen that created them is gone.

### Why `context.mounted` is not a sufficient guard

`Element.mounted => _widget != null` (framework.dart:3657), and `_widget` is only nulled in `unmount()` (framework.dart:4851) — *after* `_inheritedElements` has already been cleared. There is a window where `context.mounted == true` and every `context.read` still throws. The reliable fix is to capture the dependency before building the callback, plus a `bloc.isClosed` check, since `BlocProvider` closes the captured bloc on pop and `emit` after close throws.

## Affected Sites

| File | Line (at 1.0.17) | Callback |
|---|---|---|
| `mobile/lib/screens/settings/bluesky_settings_screen.dart` | 109 | `SnackBarAction` → `context.read<CrosspostSettingsCubit>()` |
| `mobile/lib/screens/settings/bluesky_settings_screen.dart` | 220 | `_openClaimFlowAndRefresh` → `context.read` + `context.push` |
| `mobile/lib/widgets/video_feed_item/reel_dm_reply_bar.dart` | 334 | Retry `SnackBarAction` → `context.read<InlineReelReplyCubit>()` |
| `mobile/lib/widgets/video_feed_item/reel_dm_reply_bar.dart` | 356 | View chat → `_openChat` → `context.push` |

A sweep of all 6 `SnackBarAction` sites in `mobile/lib` found the rest already correct — `user_list_people_screen.dart:723`, `video_publish_provider.dart:630` and `support_center_screen.dart:192` each capture their dependency before building the action, which is the pattern the two broken sites were missing.

## Notes

- The Crashlytics issue also carries a sample reading `Provider<OthersFollowersBloc> not found for BlocBuilder<OthersFollowersBloc, OthersFollowersState>`. That sample's stack is not in the export, so its call path is unconfirmed. The only live `BlocBuilder<OthersFollowersBloc, OthersFollowersState>` is `others_followers_screen.dart:106`, whose `MultiBlocProvider` is created directly above it in the same route widget — a genuinely-missing provider is not reachable there, so the same dead-context path is the likely explanation. Left out of scope pending a sample with frames.
- `mobile/lib/widgets/profile/profile_followers_stat.dart` contains a second `BlocBuilder<OthersFollowersBloc, …>` that would hit a genuinely-missing provider, but the widget has had zero call sites since at least 1.0.13. Dead code, tracked separately if worth deleting.

realmeylisdev:
## Description

Snackbar actions that call `context.read<T>()` throw when tapped after their screen is gone. `MaterialApp`'s root `ScaffoldMessenger` sits above the `Navigator` — the repo sets no `scaffoldMessengerKey` and installs no app-level messenger — so a snackbar migrates to the next route's `Scaffold` on a pop and stays tappable long after the widget that created it is defunct.

Traced from Crashlytics issue `a8c821212ea690bd4ecf96e4a9d696c9` on 1.0.17 (813), whose sampled session lands on `bluesky_settings_screen.dart:109`.

### Root cause

`provider.dart:377` (provider 6.1.5+1) is `throw ProviderNotFoundException(T, context.widget.runtimeType);`. Reaching it means `_inheritedElements?[T]` was null — and `Element._ensureDeactivated()` (framework.dart:4820), which always runs right after `Element.deactivate()`, sets `_inheritedElements = null`. Every provider lookup on a deactivated element therefore returns null regardless of what is in the tree.

That produces two different Crashlytics messages from one defect, which is why they were merged into a single issue:

- **deactivated, still mounted** → the real `ProviderNotFoundException`
- **unmounted** → `context.widget` throws first (`Widget get widget => _widget!`, framework.dart:3653) → `Null check operator used on a null value`, masking it

The app never crashed: `GestureRecognizer.invokeCallback` caught the error and `crash_reporting_service.dart:45` forwarded it via `recordFlutterFatalError`. The visible symptom was a CTA that silently did nothing.

`context.mounted` is **not** a sufficient guard here — `Element.mounted => _widget != null`, and `_widget` is nulled in `unmount()`, one step *after* the inherited-widget map is cleared. The fix captures dependencies while the route is mounted, gates bloc work on `isClosed` / `State.mounted`, and captures router/path data for navigation actions that can still be honored after the original widget is gone.

### Commits

1. **`fix(bluesky)`** — the crosspost failure snackbar's **Set up handle** action, plus `_openClaimFlowAndRefresh`, which did a second `context.read` and a `context.push` on the same tap.
2. **`fix(dm)`** — the reel reply bar's **Retry** action (same `context.read` shape) and **View chat**, which would have thrown `No GoRouter found in context`. Both reachable after scrolling to the next reel disposes the bar.
3. **`fix(snackbar)`** — follow-up takeover cleanup: captures `GoRouter` for claim routing, closes the retry snackbar when its reply bar is disposed, and keeps **View chat** functional after the bar is gone by capturing router/path data.
4. **`fix(dm)`** — hardens that dispose-time close against two framework asserts it could trip. `ScaffoldFeatureController.close()` asserts it owns the front of the messenger's queue, so a retry snackbar queued behind one of the six other snackbars on the reel route threw `'_snackBars.first == controller'` — and without the assert it would have dismissed that other snackbar instead; the close is now gated on `SnackBar.onVisible`, the only public signal that ours reached the front. Under `accessibleNavigation` the messenger skips the exit animation and rebuilds synchronously, so the close reached `setState` while `BuildOwner.lockState` was unmounting the bar (`setState() or markNeedsBuild() called when widget tree was locked.`); that one call is now deferred to a microtask past the lock, and dropped if the captured messenger is gone by then.

A sweep of all 6 `SnackBarAction` sites in `mobile/lib` found the other four already correct — `user_list_people_screen.dart:723`, `video_publish_provider.dart:630`, `support_center_screen.dart:192` and `profile_avatar_section.dart:284` each capture their dependency before building the action, which is the pattern these two were missing.

No behaviour change while the screen is alive.

**Related Issue:** Closes #6520

## Out of Scope

- The same Crashlytics issue carries a sample reading `Provider<OthersFollowersBloc> not found for BlocBuilder<OthersFollowersBloc, OthersFollowersState>`. That sample's stack is not in the export, so its call path is unconfirmed. The only live such `BlocBuilder` is `others_followers_screen.dart:106`, whose `MultiBlocProvider` is created directly above it in the same route widget, so a genuinely-missing provider is not reachable there — tracked in #6527 pending a sample with frames.
- `mobile/lib/widgets/profile/profile_followers_stat.dart` holds a second `BlocBuilder<OthersFollowersBloc, …>` that *would* hit a genuinely-missing provider, but the widget has had zero call sites since at least 1.0.13. Dead code, tracked in #6528.

## Verification

Regression tests were confirmed to fail without their fix, with the debug-mode counterpart of the release crash:

- `bluesky_settings_screen_test.dart` → `Looking up a deactivated widget's ancestor is unsafe.`
- `reel_dm_reply_bar_test.dart` → `This widget has been unmounted, so the State no longer has a context (and should be considered defunct).`
- `leaves a queued retry snackbar for whatever is on screen` → `'_snackBars.first == controller': is not true.`
- `closes the retry snackbar when navigation is accessible` → `setState() or markNeedsBuild() called when widget tree was locked.`

Run from `mobile/`:

- `flutter test test/screens/settings/bluesky_settings_screen_test.dart test/widgets/video_feed_item/reel_dm_reply_bar_test.dart` — passed
- `flutter test test/blocs/crosspost_settings/ test/blocs/dm/inline_reel_reply/ test/screens/settings/ test/widgets/video_feed_item/` — 450 passed (rebased onto current main)
- `mise exec -- dart format lib test integration_test` — 2814 files checked, 0 changed

No screenshots: there is no visual change.

## Type of Change

### DIFF
--- a/mobile/lib/screens/settings/bluesky_settings_screen.dart
+++ b/mobile/lib/screens/settings/bluesky_settings_screen.dart
@@ -123,6 +123,7 @@ class _BlueskySettingsView extends StatelessWidget {
     // defunct element, whose inherited-widget map the framework has already
     // cleared.
     final cubit = context.read<CrosspostSettingsCubit>();
+    final router = GoRouter.of(context);
     switch (state.error) {
       case CrosspostSettingsError.usernameNotClaimed:
         messenger.showSnackBar(
@@ -137,7 +138,7 @@ class _BlueskySettingsView extends StatelessWidget {
                 // tracks whether the screen the action refers to still exists.
                 if (cubit.isClosed) return;
                 cubit.acknowledgeError();
-                unawaited(_openClaimFlowAndRefresh(context, cubit));
+                unawaited(_openClaimFlowAndRefresh(router, cubit));
               },
             ),
           ),
@@ -217,7 +218,7 @@ class _UsernameRequiredNotice extends StatelessWidget {
       trailing: TextButton(
         onPressed: () => unawaited(
           _openClaimFlowAndRefresh(
-            context,
+            GoRouter.of(context),
             context.read<CrosspostSettingsCubit>(),
           ),
         ),
@@ -231,14 +232,11 @@ class _UsernameRequiredNotice extends StatelessWidget {
 }
 
 Future<void> _openClaimFlowAndRefresh(
-  BuildContext context,
+  GoRouter router,
   CrosspostSettingsCubit cubit,
 ) async {
-  await context.push(Nip05SettingsScreen.path);
-  // `context.mounted` only goes false once the element is unmounted, and the
-  // framework clears the inherited-widget map one step earlier — so pair it
-  // with the cubit's own liveness rather than trusting it alone.
-  if (!context.mounted || cubit.isClosed) return;
+  await router.push(Nip05SettingsScreen.path);
+  if (cubit.isClosed) return;
   await cubit.loadStatus();
 }
 
--- a/mobile/lib/screens/settings/bluesky_settings_screen.dart
+++ b/mobile/lib/screens/settings/bluesky_settings_screen.dart
@@ -117,6 +117,12 @@ class _BlueskySettingsView extends StatelessWidget {
     if (state.status != CrosspostSettingsStatus.failure) return;
 
     final messenger = ScaffoldMessenger.of(context);
+    // Captured here, while the route is still mounted. The root
+    // ScaffoldMessenger lives above the Navigator, so this snackbar outlives a
+    // pop — reading the cubit off [context] inside the action would then hit a
+    // defunct element, whose inherited-widget map the framework has already
+    // cleared.
+    final cubit = context.read<CrosspostSettingsCubit>();
     switch (state.error) {
       case CrosspostSettingsError.usernameNotClaimed:
         messenger.showSnackBar(
@@ -127,8 +133,11 @@ class _BlueskySettingsView extends StatelessWidget {
               label: context.l10n.blueskySetUpHandle,
               textColor: context.vineColors.primaryText,
               onPressed: () {
-                context.read<CrosspostSettingsCubit>().acknowledgeError();
-                unawaited(_openClaimFlowAndRefresh(context));
+                // BlocProvider closes the cubit when the route pops, so this
+                // tracks whether the screen the action refers to still exists.
+                if (cubit.isClosed) return;
+                cubit.acknowledgeError();
+                unawaited(_openClaimFlowAndRefresh(context, cubit));
               },
             ),
           ),
@@ -206,7 +215,12 @@ class _UsernameRequiredNotice extends StatelessWidget {
         style: TextStyle(color: context.vineColors.mutedText, fontSize: 14),
       ),
       trailing: TextButton(
-        onPressed: () => unawaited(_openClaimFlowAndRefresh(context)),
+        onPressed: () => unawaited(
+          _openClaimFlowAndRefresh(
+            context,
+            context.read<CrosspostSettingsCubit>(),
+          ),
+        ),
         child: Text(
           context.l10n.blueskySetUpHandle,
           style: const TextStyle(color: VineTheme.vineGreen),
@@ -216,10 +230,15 @@ class _UsernameRequiredNotice extends StatelessWidget {
   }
 }
 
-Future<void> _openClaimFlowAndRefresh(BuildContext context) async {
-  final cubit = context.read<CrosspostSettingsCubit>();
+Future<void> _openClaimFlowAndRefresh(
+  BuildContext context,
+  CrosspostSettingsCubit cubit,
+) async {
   await context.push(Nip05SettingsScreen.path);
-  if (!context.mounted) return;
+  // `context.mounted` only goes false once the element is unmounted, and the
+  // framework clears the inherited-widget map one step earlier — so pair it
+  // with the cubit's own liveness rather than trusting it alone.
+  if (!context.mounted || cubit.isClosed) return;
   await cubit.loadStatus();
 }
 
--- a/mobile/lib/widgets/video_feed_item/reel_dm_reply_bar.dart
+++ b/mobile/lib/widgets/video_feed_item/reel_dm_reply_bar.dart
@@ -163,6 +163,16 @@ class _ReelDmReplyBarState extends State<_ReelDmReplyBar> {
   ScaffoldFeatureController<SnackBar, SnackBarClosedReason>?
   _retrySnackBarController;
 
+  /// Whether the retry snackbar reached the front of the messenger's queue.
+  ///
+  /// `close()` asserts the controller it belongs to is that front entry, so one
+  /// still waiting behind another snackbar must be left alone.
+  bool _retrySnackBarVisible = false;
+
+  /// Captured while mounted so the deferred close in [dispose] can check the
+  /// messenger is still alive without a lookup on a defunct element.
+  ScaffoldMessengerState? _messenger;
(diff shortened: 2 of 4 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I pushed a takeover follow-up on top of your branch.

What changed:
- Captured `GoRouter` for the Bluesky claim flow so `_openClaimFlowAndRefresh` no longer depends on the stale `BuildContext` for navigation.
- Made the DM retry snackbar owned by `_ReelDmReplyBarState` and closed it from `dispose()`, so the dead Retry CTA disappears when the bar leaves the tree.
- Kept the success **View chat** CTA useful after the bar is disposed by capturing the router, conversation path, and participant pubkeys while mounted.
- Added/updated widget coverage for retry dismissal and post-dispose View chat navigation.
- Filed follow-ups for the out-of-scope `OthersFollowersBloc` sample (#6527) and dead `ProfileFollowersStat` widget (#6528), and linked them in the PR body.

Verification:
- `flutter test test/screens/settings/bluesky_settings_screen_test.dart test/widgets/video_feed_item/reel_dm_reply_bar_test.dart`
- `flutter test test/blocs/crosspost_settings/ test/blocs/dm/inline_reel_reply/ test/screens/settings/ test/widgets/video_feed_item/` — 412 passed
- `mise exec -- dart format lib test integration_test` — 0 changed
- GitHub checks are green on `2ae057e61`.

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up. The stale context reads are removed, Retry no longer leaves a dead CTA behind after the reply bar is disposed, View chat keeps working from the surviving snackbar via captured route data, the out-of-scope related items are tracked, and local plus GitHub verification is green.

Issue #6522, opened 2026-07-31
fix(dm): improve inbound duplicate handling


NotThatKindOfDrLiz:
## Tracking status

This is the canonical tracker under #8227 for the remaining cross-client DM interoperability problem described in #7631. #7631 is superseded.

## Remaining problem

A peer client may retry by re-minting an NIP-17 rumor with a new event ID and created_at, while keeping identical content. On the receiving side, that is indistinguishable from a user intentionally sending the same message twice. NIP-17 currently provides no client-message correlation identifier.

Divine sender code does not create this failure mode: its retry path republishes the stored rumor from outgoing_dms instead of minting a new one.

## Shipped

- #7633 added logging that distinguishes DM persistence from deduplication skips.
- #8169 restricted the timestamp/content matcher to cross-protocol counterparts.
- #8216 and #8256 fixed the related duplicate-handling defects and added focused DAO coverage.
- Same-protocol replays are handled by the rumor ID primary key and gift-wrap uniqueness.

## Rejected approaches

- **Widening the timestamp window:** this would absorb legitimate repeated messages and recreate the proven #7324 data-loss behavior.
- **Content-hash deduplication:** identical content is not evidence that two sends are the same message; an unbounded hash rule would permanently discard legitimate repeats.
- **Receiver-side suppression of re-minted rumors:** without a wire-level correlation signal, no receiver heuristic can safely distinguish retries from repeated sends.

## Remaining work

Define and adopt a protocol-level correlation signal, or another sender/receiver contract that makes a retry distinguishable from a new send. Do not implement wider-window, content-hash, or other heuristic receiver-side suppression without resolving that ambiguity first.

Test-suite debt around dedup behavior is tracked separately in #8170.

NotThatKindOfDrLiz:
## Disposition of #7631

#7631 is superseded. Its observability gap shipped in #7633, while the residual peer re-mint case belongs here as the single tracker under #8227.

### Rejected approaches

- **Widening the ±5-second window:** the current matcher is gated to cross-protocol counterparts. Same-protocol replays are already safely handled by the rumor ID primary key and gift-wrap uniqueness. Widening the window would absorb legitimate repeated messages and recreate the proven #7324 data-loss behavior.
- **Content-hash deduplication:** rejected for the same reason, with no safe time boundary. Identical content is not evidence that two sends are the same message.
[...]

PR #6523, opened 2026-07-31
fix(dm): improve inbound duplicate handling

NotThatKindOfDrLiz:
## Summary
- make direct message inserts report whether a row was actually written
- record and log duplicate inbound NIP-17/NIP-04 wraps instead of silently reprocessing them
- keep the generic sender/content dedup window narrow so legitimate repeated messages are not dropped
- add DAO and repository regression coverage for duplicate handling and skipped sync-boundary updates
- harden one unread-count cubit test by waiting on the stream emission instead of wall-clock settling

Refs #6522. This intentionally does not close the issue because a peer re-minted rumor with a new id and timestamp remains indistinguishable from a legitimate repeated send without a stronger protocol correlation signal.

Refs #6645 for the pre-existing multi-account DM storage uniqueness gap made more permanent by ledgering ignored inbound wraps.

## Tests
- flutter test test/src/database/daos/direct_messages_dao_test.dart (`mobile/packages/db_client`)
- flutter test test/src/dm_repository_test.dart (`mobile/packages/dm_repository`)
- flutter analyze (`mobile`)
- git diff --check
- pre-push changed-file checks

### DIFF
--- a/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
@@ -23,7 +23,7 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
     return column.equals(ownerPubkey) | column.isNull();
   }
 
-  /// Insert a decrypted DM, silently skipping duplicates.
+  /// Insert a decrypted DM, returning whether a row was actually written.
   ///
   /// Uses `INSERT OR IGNORE` so that violations on either the primary key
   /// (`id`) **or** the UNIQUE index on `gift_wrap_id` are handled gracefully
@@ -36,7 +36,7 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   /// For kind 14 (text), only [content] is used.
   /// For kind 15 (file), [content] holds the file URL and file metadata
   /// fields are populated from the event tags.
-  Future<void> insertMessage({
+  Future<bool> insertMessage({
     required String id,
     required String conversationId,
     required String senderPubkey,
@@ -59,8 +59,8 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
     String? thumbnailUrl,
     String? ownerPubkey,
     String? sendBatchId,
-  }) {
-    return into(directMessages).insert(
+  }) async {
+    await into(directMessages).insert(
       DirectMessagesCompanion.insert(
         id: id,
         conversationId: conversationId,
@@ -87,6 +87,10 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
       ),
       mode: InsertMode.insertOrIgnore,
     );
+    final changed = await customSelect(
+      'SELECT changes() AS changed',
+    ).map((row) => row.read<int>('changed')).getSingle();
+    return changed > 0;
   }
 
   /// Get messages for a conversation, newest first.
@@ -206,18 +210,20 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   }
 
   /// Check if a message with the same sender and content already exists in a
-  /// conversation within a ±5 second window. Used for cross-protocol dedup
-  /// when both a NIP-17 and NIP-04 copy of the same message arrive.
+  /// conversation within a bounded timestamp window. Used for cross-protocol
+  /// dedup when both a NIP-17 and NIP-04 copy of the same message arrive, and
+  /// for peer clients that retry by recreating a NIP-17 rumor instead of
+  /// re-wrapping the original one.
   ///
   /// The time window prevents false positives when a user genuinely sends
-  /// the same text twice (e.g. "ok") while still catching dual-send
-  /// duplicates where timestamps differ by at most a few seconds.
+  /// the same text again later, while still catching retry duplicates whose
+  /// recreated rumor timestamps drift beyond send-confirm latency.
   Future<bool> hasMatchingMessage({
     required String conversationId,
     required String senderPubkey,
     required String content,
     required int createdAt,
-    int windowSeconds = 5,
+    int windowSeconds = 60,
     String? ownerPubkey,
   }) async {
     final query = selectOnly(directMessages)
--- a/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
@@ -27,11 +27,13 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   ///
   /// Uses `INSERT OR IGNORE` so that violations on either the primary key
   /// (`id`) **or** the UNIQUE index on `gift_wrap_id` are handled gracefully
-  /// without throwing. Callers already dedup via [hasGiftWrap] before calling
-  /// this method; the ignore mode is a safety net for race conditions.
+  /// without throwing. A `false` return means a local uniqueness constraint
+  /// skipped the row, and callers must avoid advancing receive-side state that
+  /// depends on a newly persisted message.
   ///
   /// NIP-17 rumor events are immutable — the same rumor ID always carries
-  /// the same content, so skipping duplicates never loses data.
+  /// the same content. The current message uniqueness constraints are global,
+  /// so callers still need owner-scoped checks where account isolation matters.
   ///
   /// For kind 14 (text), only [content] is used.
   /// For kind 15 (file), [content] holds the file URL and file metadata
--- a/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
@@ -33,6 +33,11 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   /// NIP-17 rumor events are immutable — the same rumor ID always carries
   /// the same content, so skipping duplicates never loses data.
   ///
+  /// The returned value is derived from SQLite `changes()` immediately after
+  /// the insert on the same Drift executor. Keep callers serialized through
+  /// Drift (the current receive paths call this inside transactions) so no
+  /// intervening write can affect the result.
+  ///
   /// For kind 14 (text), only [content] is used.
   /// For kind 15 (file), [content] holds the file URL and file metadata
   /// fields are populated from the event tags.
@@ -210,20 +215,18 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   }
 
   /// Check if a message with the same sender and content already exists in a
-  /// conversation within a bounded timestamp window. Used for cross-protocol
-  /// dedup when both a NIP-17 and NIP-04 copy of the same message arrive, and
-  /// for peer clients that retry by recreating a NIP-17 rumor instead of
-  /// re-wrapping the original one.
+  /// conversation within a ±5 second window. Used for cross-protocol dedup
+  /// when both a NIP-17 and NIP-04 copy of the same message arrive.
   ///
   /// The time window prevents false positives when a user genuinely sends
-  /// the same text again later, while still catching retry duplicates whose
-  /// recreated rumor timestamps drift beyond send-confirm latency.
+  /// the same text again moments later, while still catching dual-send
+  /// duplicates where timestamps differ by at most a few seconds.
   Future<bool> hasMatchingMessage({
     required String conversationId,
     required String senderPubkey,
     required String content,
     required int createdAt,
-    int windowSeconds = 60,
+    int windowSeconds = 5,
     String? ownerPubkey,
   }) async {
     final query = selectOnly(directMessages)
--- a/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
@@ -33,11 +33,6 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   /// NIP-17 rumor events are immutable — the same rumor ID always carries
   /// the same content, so skipping duplicates never loses data.
   ///
-  /// The returned value is derived from SQLite `changes()` immediately after
-  /// the insert on the same Drift executor. Keep callers serialized through
-  /// Drift (the current receive paths call this inside transactions) so no
-  /// intervening write can affect the result.
-  ///
   /// For kind 14 (text), only [content] is used.
   /// For kind 15 (file), [content] holds the file URL and file metadata
   /// fields are populated from the event tags.
@@ -65,7 +60,7 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
     String? ownerPubkey,
     String? sendBatchId,
   }) async {
-    await into(directMessages).insert(
+    final inserted = await into(directMessages).insertReturningOrNull(
       DirectMessagesCompanion.insert(
         id: id,
         conversationId: conversationId,
@@ -92,10 +87,7 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
       ),
       mode: InsertMode.insertOrIgnore,
     );
-    final changed = await customSelect(
-      'SELECT changes() AS changed',
-    ).map((row) => row.read<int>('changed')).getSingle();
-    return changed > 0;
+    return inserted != null;
   }
 
   /// Get messages for a conversation, newest first.
--- a/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
@@ -218,9 +218,12 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   /// conversation within a ±5 second window. Used for cross-protocol dedup
   /// when both a NIP-17 and NIP-04 copy of the same message arrive.
   ///
-  /// The time window prevents false positives when a user genuinely sends
-  /// the same text again moments later, while still catching dual-send
-  /// duplicates where timestamps differ by at most a few seconds.
+  /// The window is deliberately narrow: a genuine repeat of the same text
+  /// (e.g. "ok") is only collapsed when it lands within ±[windowSeconds],
+  /// which is still wide enough for dual-send duplicates, whose timestamps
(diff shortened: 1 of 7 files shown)


rabble (changes requested):
The default dedup window is now 60 seconds, but `hasMatchingMessage` only compares conversation, sender, exact content, and timestamp. That query is used for both cross-protocol dedup and the group-history recovery path, so two legitimate identical messages from the same person within one minute—common examples are “ok”, “thanks”, or short reactions—will silently collapse into one. The existing 5-second window explicitly guarded against this false positive, and the new tests only prove that 61 seconds is outside the window. Please keep the generic/content-based window narrow (or split the retry-specific path so it has a stronger correlation signal) and add a regression test for two identical messages inside the intended legitimate-send interval.

NotThatKindOfDrLiz:
Took this over and pushed `f6e9657e9`.

What changed:
- Restored `hasMatchingMessage` to the narrow default `windowSeconds = 5`, so generic sender/content matching does not drop legitimate repeated messages inside a minute.
- Reworked the DAO tests to pin the 5s match and a 30s same-content non-match.
- Added repository coverage that skipped NIP-17/NIP-04 inserts do not upsert conversations or advance sync boundaries, including transaction re-check branches.
- Fixed the stale NIP-04 dedup comment and documented the `insertMessage`/`changes()` executor assumption.
- Updated the PR body to remove the old 60s-window claim and call out the unrelated unread-count test hardening.

Verification:
- `flutter test test/src/database/daos/direct_messages_dao_test.dart` (`mobile/packages/db_client`)
- `flutter test test/src/dm_repository_test.dart` (`mobile/packages/dm_repository`)
- `flutter analyze` (`mobile`)
- `git diff --check`
- pre-push changed-file checks: analyzer, generated files, and changed-file tests all passed

@rabble please re-review when you have a chance.

hm21 (approved):
Took this over and pushed `a88f3ca516`. The only change is restoring the `sendBatchId` doc comment: it was reworded to describe the 60s window, and once `f6e9657` put the default back to ±5s it no longer described anything. `tables.dart` and the regenerated `app_database.g.dart` are byte-identical to `main` again, so a schema file and a generated file drop out of the diff.

Validated locally: `db_client` 905 tests, `dm_repository` 397 tests, `dart analyze` clean on both packages, and `build_runner` reproduces the committed `.g.dart`.

The rest reads correct. I checked `SELECT changes()` against a real database rather than the mocks — it holds inside a transaction opened by a *different* DAO, and Drift serializes statements, so a write racing outside the transaction can't clobber it. All six `insertMessage` call sites are inside `runInTransaction`, so the caveat in the DAO doc is satisfied.

Two things for you to decide, neither blocking:

**1. #6522's window criterion isn't met, and I don't think it can be.** The ignored-insert path fixes duplicates where a peer re-wraps the *same* rumor — same rumor id, so the primary key catches it. But a rumor recreated with a different `created_at` hashes to a different id and is, by construction, indistinguishable from someone genuinely sending "ok" twice. rabble was right to reject the 60s window and I don't see a correlation signal that would make a wider one safe, so the "peer recreates a rumor more than a few seconds later" duplicate the issue describes is still unfixed. Worth splitting into a follow-up rather than letting `Closes #6522` close it as done.

**2. A multi-account edge that this makes permanent.** `direct_messages` has a global primary key on the rumor id and a global unique index on `gift_wrap_id`, while every read is owner-scoped. NIP-17 wraps one rumor to each recipient individually, so with two of your own accounts in the same peer-sent group DM on one device, the second account's wrap carries an already-stored rumor id, the insert is ignored, and the message stays invisible to that account. That part is pre-existing — but the new `_recordProcessedWrap` on the ignored-insert path now ledgers the wrap, so it will never be re-decrypted. Rare, and really a schema problem rather than yours; flagging it because the ledger entry is what turns it from recoverable into permanent.

One thing I tried and backed out: folding `inserted` / `skippedByTransactionalGiftWrapDedup` into an outcome enum returned from the transaction. `runInTransaction<T>` is stubbed per type argument in the tests, so a private enum can't be mocked without exporting it — your captured-boolean version is the right trade.

Approving; the merge call is yours.

rabble (commented):
Review note: the current head correctly handles same-rumor and database-constraint duplicates, but the PR still does not cover the logical retry case described by #6522 when a peer recreates the rumor with a new ID and its timestamp falls outside the existing ±5s sender/content window. In that case hasMatchingMessage returns false and INSERT OR IGNORE sees no conflict, so a second bubble can still be stored. That case is indistinguishable from a legitimate repeated message without a stronger correlation signal, so please either narrow/remove the Closes #6522 claim and track this as a follow-up, or add the protocol-level correlation needed to make the wider match safe. I found no additional blocker in the changed persistence bookkeeping.

mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
```diff
@@ -209,9 +218,12 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   /// conversation within a ±5 second window. Used for cross-protocol dedup
   /// when both a NIP-17 and NIP-04 copy of the same message arrive.
   ///
-  /// The time window prevents false positives when a user genuinely sends
-  /// the same text twice (e.g. "ok") while still catching dual-send
-  /// duplicates where timestamps differ by at most a few seconds.
+  /// The window is deliberately narrow: a genuine repeat of the same text
```
realmeylisdev, line 215: **The `Closes #6522` claim — AC4 has no behaviour delta on this branch.**

`windowSeconds` is 5 on `main` and 5 here, and the `hasMatchingMessage` query body is byte-identical to base. I extracted the base tree, copied the two new window tests into it verbatim, and ran them against unmodified base code — both pass. They pin behaviour that already shipped.

@rabble — I went looking for the correlation signal your last comment asks for, and there isn't one. Grepping all 99 NIPs: `idempot` has one hit (NIP-57 zap receipts, which achieves idempotency by making `created_at` deterministic — the same and only lever available here), `dedup` one unrelated hit, and `correlat` two hits that both argue *against* correlation. The only dedup primitive is relay-side and keyed on event id (`["OK", "<event_id>", true, "duplicate: already have this event"]`), which is useless once the id changes. Kind 14's tag surface is closed at `p`/`e`/`subject`/`q`; it is a regular event, so there is no `d`-tag replacement relationship either. NIP-59 explicitly leaves the rumor's `created_at` alone ("The canonical `created_at` time belongs to the `rumor`"), and NIP-17 advertises the *absence* of a correlator as a feature ("No Public Group Identifiers ... nothing to correlate"). A re-minted rumor is therefore byte-indistinguishable from a genuine second send, and `(sender, content, created_at ± window)` is the only signal the protocol exposes.

Worth noting Divine's own retries cannot produce this: `recoverSelfWrap`/`recoverRecipientWrap` replay the persisted rumor bytes, and `sendMessage` builds the rumor once so the queue row PK matches the id the relay sees. It only bites against a foreign client that re-mints.

Merging as-is auto-closes #6522 (confirmed via `closingIssuesReferences`), and no follow-up exists — I checked open issues, open PRs, and #6522's timeline. Suggest `Refs #6522` plus a follow-up for the residual case. Happy to file it if you'd rather not.

realmeylisdev (commented):
Took this over and pushed `3bf171a73` and `a353bd6b9` — one fixes a test that was passing on a swallowed exception instead of on the guard it names, the other corrects a doc comment that describes the dedup window backwards. No production behaviour change; `db_client` 39 and `dm_repository` 397 green, analyze and format clean.

The `Closes #6522` question is still open and it's your call — details inline, including why the protocol-level correlation signal rabble asked for does not exist in any NIP.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -1675,6 +1675,10 @@ class DmRepository {
     try {
       // Dedup: skip if already processed (message row or ledger). #5452.
       if (await _alreadyProcessed(giftWrapEvent.id)) {
+        Log.debug(
```
realmeylisdev: This sits on the hottest skip path in the receive pipeline, and `Log.debug` is not free in release builds: `UnifiedLogger._log` gates only the `kDebugMode` console block on level, then calls `LogCaptureService().captureLog(entry)` unconditionally. Measured it — at `LogLevel.info` (the release default) `isLevelEnabled(debug)` is false, yet 1001 of 1001 debug entries still landed in the capture buffer.

Volume: the live subscription re-REQs a 2-day window on every reconnect (`_scheduleReconnect` fires 2s after any stream close), and history-drain pass 2 routes every already-processed wrap back through `_handleGiftWrapEvent`, bounded at pageSize 100 × maxPages 50 = 5000. `BugReportConfig.maxLogEntries` is 5000 and `buildLogsSummary` takes the last 50 entries — so one drain can evict the whole bug-report window and leave the summary as pure dedup noise. On a change whose stated purpose is diagnosability that seems like the wrong trade.

The other five new log lines are all on terminal paths and read fine. This one either wants dropping (the branch is a no-op by design) or collapsing to a single count per page/subscription.

mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
```diff
@@ -33,10 +33,15 @@ class DirectMessagesDao extends DatabaseAccessor<AppDatabase>
   /// NIP-17 rumor events are immutable — the same rumor ID always carries
   /// the same content, so skipping duplicates never loses data.
   ///
+  /// The returned value is derived from SQLite `changes()` immediately after
+  /// the insert on the same Drift executor. Keep callers serialized through
+  /// Drift (the current receive paths call this inside transactions) so no
```
realmeylisdev: The parenthetical is not accurate — a drift transaction does not serialize the statements issued inside it. `NativeDatabase` is constructed `isSequential: false`, `_TransactionExecutor.isSequential` just forwards that, and `_synchronized` takes the lock only when `isSequential || _waitingChildExecutors > 0`.

Concretely: two `insertMessage` calls issued concurrently *inside one transaction* interleave as INSERT / INSERT / `changes()` / `changes()`, so both read the last statement's count — 30/30 wrong in-process and 20/20 over `NativeDatabase.createInBackground`. A false negative there writes the message row but skips the conversation upsert and leaves the sync boundary unadvanced.

It is correct as shipped — all six call sites are exactly one awaited insert per transaction body, and separate concurrent transactions *are* serialized (0/30 wrong). The issue is that nothing enforces the one-insert-per-transaction property and the comment names the wrong thing as the protection. `insertReturningOrNull` exists in drift 2.34.0, returns `null` when `insertOrIgnore` skipped, and measured 0/30 wrong under the exact race that breaks `changes()`; the bundled SQLite is 3.53.2 so `RETURNING` is available on every target. `customUpdate` is the other safe option — `NativeDatabase.runUpdate` returns `updatedRows` inside the same synchronous block as the statement, so there is no await point at all.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -1952,6 +1968,24 @@ class DmRepository {
         );
       });
 
+      if (skippedByTransactionalGiftWrapDedup) {
```
realmeylisdev, line 1967: Nit: deleting this block — or its NIP-04 twin at `:2530` — leaves the full 556-test package suite green *and* behaviour unchanged. Control falls through to the `!inserted` block below and returns at the same point; the only differences are which log string fires and one redundant ledger write on the NIP-17 side for a wrap that already has a message row (so `_alreadyProcessed` short-circuits on it anyway). Fine to keep for the distinct log line — just flagging that it carries no behavioural weight, in case that was the intent.

mobile/packages/db_client/test/src/database/daos/direct_messages_dao_test.dart
```diff
@@ -230,6 +230,31 @@ void main() {
...
         expect(results.first.sendBatchId, isNull);
       });
+
+      test('reports whether insert wrote a row', () async {
+        final firstInserted = await dao.insertMessage(
+          id: 'msg_insert_result',
+          conversationId: conversationId1,
+          senderPubkey: 'pubkey_alice',
+          content: 'Original',
+          createdAt: 1700000000,
```
realmeylisdev, line 240: Nit: this covers the primary-key half only — both inserts share `id` and differ in `giftWrapId`. The doc on `insertMessage` claims both the PK *and* the `gift_wrap_id` UNIQUE index are handled. I checked the second path against a real `AppDatabase` and it does return `false` correctly, so it is a coverage gap rather than a bug; a third insert with a fresh `id` and a reused `giftWrapId` would close it.

NotThatKindOfDrLiz:
Took this over and pushed `e8dacc381`.

What changed:
- Swapped `DirectMessagesDao.insertMessage` from `SELECT changes()` to Drift `insertReturningOrNull`, so the boolean result comes directly from the insert statement.
- Added DAO coverage for the `gift_wrap_id` UNIQUE conflict path, not just the primary-key duplicate path.
- Removed the already-processed NIP-17 fast-path debug log so resume/history no-ops do not fill the captured bug-report log window.
- Updated the PR body from `Closes #6522` to `Refs #6522`, with the re-minted-rumor limitation called out explicitly.

Verification:
- `flutter test test/src/database/daos/direct_messages_dao_test.dart` (`mobile/packages/db_client`) — 39 passed
- `flutter test test/src/dm_repository_test.dart` (`mobile/packages/dm_repository`) — 336 passed
- `flutter analyze` (`mobile`) — clean
- `git diff --check` — clean
- pre-push changed-file checks — analyzer, generated files, and changed-file tests all passed

I am watching CI now and will follow up if anything fails.

mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
realmeylisdev: e8dacc381 swaps this to `insertReturningOrNull` and drops the paragraph, so the boolean now comes out of the insert statement itself and there is no follow-up read to interleave with.

mobile/packages/db_client/test/src/database/daos/direct_messages_dao_test.dart
realmeylisdev: covered in e8dacc381. third insert uses a fresh id with `gw_insert_result` reused, so it hits the `idx_dm_gift_wrap_id` unique index and asserts false. dao file 39 green.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: dropped in e8dacc381. the already-processed fast path just returns now, so resume and history no-ops stop eating the captured log window.

mobile/packages/db_client/lib/src/database/daos/direct_messages_dao.dart
realmeylisdev: settled: the body says `Refs #6522` now and `closingIssuesReferences` comes back empty, so merging leaves the issue open. that covers the re-minted-rumor case on its own, so no separate follow-up issue to file.

mobile/packages/dm_repository/lib/src/dm_repository.dart
realmeylisdev: keeping it. the distinct log line is worth the block and behaviour is the same either way.

NotThatKindOfDrLiz:
Takeover update:

- Rebased fix/dm-inbound-dedup-observability onto current origin/main and force-pushed with lease.
- Branch is now 6 ahead / 0 behind main; PR remains mergeable.
- closingIssuesReferences is empty and the body uses Refs #6522, so merge will not close the residual re-minted-rumor tracker.
- No code changes beyond the rebase.

Verification after rebase:
- flutter test test/src/database/daos/direct_messages_dao_test.dart from mobile/packages/db_client - 39/39 passed
- flutter test test/src/dm_repository_test.dart from mobile/packages/dm_repository - 336/336 passed
- flutter analyze from mobile - clean
- git diff --check origin/main...HEAD - clean
- pre-push generated-file checks - passed
- pre-push changed-file tests, including test/blocs/dm/unread_count/dm_unread_count_cubit_test.dart - passed
- GitHub checks - all green

@rabble this is ready for re-review. Your CHANGES_REQUESTED review was against the reverted 60s window; the current head keeps the narrow +/-5s window and leaves #6522 open via Refs #6522.

NotThatKindOfDrLiz:
Takeover update pushed `516b45d26`.

What changed:
- Filed #6645 for the pre-existing multi-account DM storage uniqueness gap, including the processed-wrap ledger recovery/backfill implication.
- Updated the PR body to reference #6645 so the follow-up is discoverable from this PR.
- Clarified `DirectMessagesDao.insertMessage` docs so the boolean ignored-insert result is described as load-bearing receive-path control flow, and removed the overbroad "skipping duplicates never loses data" wording.

Verification:
- `flutter test test/src/database/daos/direct_messages_dao_test.dart` from `mobile/packages/db_client` - 39/39 passed
- `git diff --check` - clean
- pre-push checks - merge-conflict check, analyzer, generated files, and changed-file tests all passed
- GitHub checks - all green, including DB Client CI, DM Repository CI, Mobile CI, preview, mergeability, and semantic PR

@rabble @hm21 @realmeylisdev this is ready for re-review. The remaining multi-account schema gap is now tracked in #6645; I do not see a correctness blocker left in this PR.

realmeylisdev (approved):
dug into the `insertReturningOrNull` swap since the whole thing hangs off it. reproduced the old `SELECT changes()` race (wrong 30/30 inside a transaction) and the new one comes back clean 0/30. RETURNING is safe on everything we ship: native is sqlite3mc 3.53.2 via the sha256-pinned asset, web is sql.js 3.45.2, and the hook has no system-sqlite fallback.

mutation-tested the 8 new tests one at a time, every one goes red when you break the guard it names, and the two window tests bracket the ±5s from both sides. also checked the skipped `recordSeen`/`upsertConversation` can't strand the live `since:` floor or the drain cursor. db_client 908, dm_repository 556 at 93.9%, analyze clean.

body says Refs #6522 and `closingIssuesReferences` is empty, so merging leaves it open. lgtm 👍

rabble (approved):
Re-reviewed the current head. The unsafe 60-second content-based dedup window was removed; the implementation keeps the narrow ±5s heuristic, uses insertReturningOrNull for statement-local conflict detection, avoids advancing receive bookkeeping on ignored inserts, and leaves the uncorrelatable re-minted-rumor case open via Refs #6522. The multi-account schema edge is separately tracked in #6645. Current-head package/app CI is green. Ready to merge.

PR #6524, opened 2026-07-31
feat(sounds): make saved sounds recognizable

### LINKED ISSUE #6525: feat(sounds): make saved sounds recognizable
Auto-created from PR #6524.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6524
**Branch:** `feat/sound-library-metadata`

rabble:
Closes #6525

## What changed

- replaces the bare saved-sound list with searchable rich vertical cards containing source post context, thumbnail, compact waveform, duration, passive transcript when available, and retained Divine sound proxy catalog metadata
- adds account-scoped BLoC state and versioned on-device persistence for private labels and hashtags, with autosave and legacy migration so library data survives app upgrades
- keeps saving strictly local: it does not publish a Nostr audio event, sync private organization metadata, or turn missing speech into a user-facing transcription error
- separates private organization from public attribution so personal labels and hashtags never enter published events
- adds explicit remix consent and fail-closed Kind 1063 handling; existing Nostr sounds are referenced rather than republished
- carries creator, source, license, provider, and catalog credit for imported/provider sounds into the publishing flow and displays public credit/remix status in sound and video metadata UI

## Privacy and protocol decisions

- local library records are keyed by the full sound ID and scoped to the active account
- private imports remain embedded unless the creator enables “Allow others to remix this sound”
- absent or ambiguous reuse consent is treated as not reusable
- provider licenses that forbid derivatives force credit-only behavior
- public attribution failures block publication only when reusable audio/credit was explicitly requested

## Verification

- `flutter analyze`
- repository pre-push analyzer, generated-file verification, and 511 related tests
- 224 changed mobile tests as one suite
- `mobile/packages/models/test/src/audio_event_test.dart`
- `mobile/packages/sounds_repository/test/src/sound_library_api_client_test.dart`
- `mobile/test/l10n/arb_consistency_test.dart`
- untested-service, raw TextStyle/color/button/dialog/icon, and test-structure ratchets
- `mobile/scripts/golden.sh verify`
- `git diff --check`

## Notes

The shared `divine-context` checkout was on `docs/add-divine-brain-spec` rather than its default branch, so it was intentionally left untouched per the context update rules; the current local review runbook and team mapping were read before opening this PR.


mobile/lib/services/audio_reuse_consent_resolver.dart
```diff
@@ -0,0 +1,48 @@
...
+              .where((video) => video.addressableId == sourceAddress)
+              .where((video) => video.createdAt >= sound.createdAt)
+              .toList()
+            ..sort((left, right) => left.createdAt.compareTo(right.createdAt));
+      if (matching.isEmpty) return false;
+      if (matching.length > 1 &&
+          matching[0].createdAt == matching[1].createdAt) {
+        return false;
+      }
+      return matching.first.allowAudioReuse;
```
realmeylisdev, line 49: `matching` is sorted ascending and this takes `.first` — the oldest event. Candidates are matched on `addressableId` (`34236:<pubkey>:<d-tag>`), which is stable across edits of the same video, so when two revisions come back this reads the pre-edit one.

That matters because `allow_audio_reuse` is listed in `_editRebuiltTagNames` and `updateVideo` only re-emits it while `editorState.allowAudioReuse` is on — dropping the tag on a new revision is exactly how a creator revokes, and the revocation lands on the newer event.

The new test locks the current behaviour in: `later` (createdAt 120, reuse false) and `earliest` (createdAt 110, reuse true) share `vineId: 'source-video'`, so both match, and it expects `isTrue`.

Was oldest-wins deliberate — reading it as "the original publish" — or should this dedupe by `addressableId` and keep the highest `createdAt`? Separately, the ambiguity guard only compares `matching[0]` against `matching[1]`, so a `createdAt` tie further down the list slips past it.

mobile/lib/internal/video_event_publisher_audio.dart
```diff
@@ -118,12 +148,76 @@ extension _VideoEventPublisherAudio on VideoEventPublisher {
...
+    required String videoDTag,
+    required String pubkey,
+    required String relayHint,
+  }) async {
+    final external = audio.externalSource;
+    final url = audio.url;
+    if (external == null ||
+        url == null ||
+        url.isEmpty ||
+        external.creatorName?.trim().isEmpty != false ||
```
realmeylisdev: This returns `null` when the provider row has no creator, and the caller turns that into `return false` for the entire video publish. The only trace is a log line — the user lands on the generic "Failed to publish Nostr event" from `video_publish_service`, and retrying can't clear it.

`creator` is nullable the whole way down: `creator TEXT` in the sound-proxy D1 schema, passed straight through by `mapIndexedRow`, `creator: raw.username || null` on the Freesound path, and `json['creator'] as String?` in `_soundFromJson` — only `id`/`provider`/`providerId`/`previewUrl` go through `_requireString`. (`source_url` is `NOT NULL`, so the creator half is the live gap.)

Two calls to make: should a thin catalog row fall back to the provider name rather than blocking, and should this branch run at all when `allowAudioReuse` is false? Right now it's unconditional, so a user who never opted into reuse can still lose the publish to a missing credit field.

mobile/lib/services/saved_sounds_service.dart
```diff
@@ -62,56 +66,132 @@ class SavedSoundsService {
...
+        continue;
+      }
+    }
+    return sounds;
+  }
+
+  List<SavedSound> _readVersionedSounds(Map<String, dynamic> decoded) {
+    if (decoded['schemaVersion'] !=
+        SavedSoundLibraryPayload.currentSchemaVersion) {
+      return [];
```
realmeylisdev: A version mismatch returns `[]`, and since the next `saveSound`/`removeSound` writes the loaded list straight back, the old payload gets overwritten. So the first bump to `currentSchemaVersion = 2` wipes every library instead of migrating it — the opposite of what versioning is here for. Worth a migration branch (or at minimum keeping the unreadable payload on disk untouched) before v2 ships.

Two smaller notes in the same area:

- `SavedSoundLibraryPayload.fromJson` isn't called anywhere in `lib/` — only its own round-trip test. Using it here would drop the duplicate hand-rolled parse and give you the `FormatException` handling for free.
- The new Map payload isn't readable by the shipped reader (`if (decoded is! List) return [];` on `main`), so a rollback to the current build reads the library as empty and the next save flattens it. Fine if rollbacks aren't a concern — flagging it since it's invisible until it happens.

mobile/lib/blocs/saved_sounds/saved_sounds_bloc.dart
```diff
@@ -0,0 +1,185 @@
...
+        );
+      }
+    }
+  }
+
+  Future<void> _remove(
+    SavedSoundRemoveRequested event,
+    Emitter<SavedSoundsState> emit,
+  ) async {
+    await _service.removeSound(event.soundId);
```
realmeylisdev: Nit: `_remove` is the only handler without error handling — `_save`, `_edit` and `_applyProbe` all catch. `removeSound` reaches `_writeSavedSounds`, which throws `StateError` when `setString` returns false, so a failed delete escapes to the observer and the row silently stays on screen.

Same thread: `saveSound`'s completer can `completeError`, and neither call site (`sounds_tab._onAddAudioTap`, `sound_detail_screen._saveSoundToLibrary`) wraps the `await`, so that path drops the snackbar entirely.

realmeylisdev (commented):
Read the whole thing. The shape is good: BLoC + account-scoped persistence, `lazy: false` with a `ValueKey(storageKey)` on the scope, the two new library widgets are pure `divine_ui`/`VineTheme`, the material-button baseline shrinks, and the 21 new keys are declared as l10n debt the sanctioned way. Four things I'd want settled first — details inline.

**The branch conflicts with `main`.** `video_publish_service.dart`, `sounds_tab.dart` and `video_metadata_form_fields.dart` all conflict. The green checks ran against a stale base, so they aren't evidence yet — needs a rebase and a re-run.

**The consent resolver reads the pre-edit revision.** It matches on `addressableId`, which is stable across edits, then takes the oldest by `createdAt`. `allow_audio_reuse` is a rebuilt tag in the edit flow, so the *newest* revision is the current consent state. The new test pins today's behaviour: the later revision says `false`, the earlier says `true`, and it expects `true`. A creator who turns reuse off after publishing gets ignored.

**A catalog sound can make a video unpublishable.** The new `isExternalProviderSound` branch runs regardless of `allowAudioReuse`, and a bridge failure fails the whole publish. `creator` is nullable from the D1 schema all the way to `AudioExternalSource`, so a thin catalog row leaves the user on a generic "Failed to publish Nostr event" with no way out. That's also broader than the "only when reusable audio/credit was explicitly requested" rule in the description.

**The versioned store deletes instead of migrating.** A `schemaVersion` mismatch returns `[]` and the next write persists it, so the first bump to v2 wipes libraries rather than upgrading them.

None of these need a redesign — mostly "which revision wins", "what happens when the catalog row is thin", and a migration branch. Happy to take another pass once they're settled.

mobile/lib/services/audio_reuse_consent_resolver.dart
rabble: Not deliberate — you read it right, and oldest-wins was wrong. Fixed in 5c5dc8e64: the sort is now descending on `createdAt`, so the newest revision decides. A creator who drops `allow_audio_reuse` on an edit now revokes as intended.

On the tie guard: once the sort is descending, comparing `matching[0]` against `matching[1]` is complete rather than partial. Every entry in `matching` already shares one `addressableId` (it is filtered on `== sourceAddress`), so the only ambiguity that can change the answer is a tie *at the maximum* `createdAt` — a tie further down is shadowed by a strictly newer revision and cannot affect the result. That case still fails closed.

Tests: `uses newest matching source video for a legacy event` now puts the reuse-granting revision on top, and a new `honours a revocation on the newest revision` pins the exact case you described (newer says false, older says true, resolver returns false).

mobile/lib/internal/video_event_publisher_audio.dart
rabble: Both calls made, in 5c5dc8e64.

**Thin catalog row no longer blocks.** `creatorName` falls back to `external.providerName` (non-nullable on `AudioExternalSource`), so a row with `creator NULL` credits "Freesound" instead of losing the publish. `sourceUrl` stays required — without it there is no durable link to credit, and as you noted `source_url` is `NOT NULL` in the D1 schema, so this is the narrow case.

**The branch no longer runs unconditionally.** The bridge still runs, but a `null` result only blocks the publish when `allowAudioReuse` is true. Otherwise it logs a warning, clears the reference, and the video ships — which is what the PR description already claimed ("only when reusable audio/credit was explicitly requested").

Tests: the old `blocks provider audio when durable credit is incomplete` split into `blocks provider audio without durable credit only when reuse is asked for` (reuse requested → still `false`), `publishes without the provider reference when reuse was never asked for` (→ `true`), and `credits the provider when the catalog row has no creator` (asserts the signed credit content contains `Freesound`).

mobile/lib/services/saved_sounds_service.dart
rabble: Fixed in 5c5dc8e64.

**Migration.** `_readVersionedSounds` no longer bails on any mismatch. It reads every schema at or below `currentSchemaVersion` with the current reader, so a future v2 bump upgrades in place instead of wiping. Only a *newer* payload is treated as unreadable, and `_writeSavedSounds` now throws before touching it — a downgraded build can no longer flatten a library it cannot parse. New tests: `reads a payload written by an older schema version` and `refuses to overwrite a payload from a newer build` (which asserts the stored string is byte-identical afterwards).

**On `SavedSoundLibraryPayload.fromJson`.** I kept the hand-rolled loop deliberately: `fromJson` maps the whole list eagerly, so one malformed record throws out of the entire payload, whereas the loop skips just that entry. `skips a corrupt versioned entry without dropping valid ones` pins that behaviour and would go red if I switched. Losing a library to one bad row seemed worse than the duplication.

**Rollback.** Accepted and unfixable from this side — the shipped reader with `if (decoded is! List) return []` is already out there, so a rollback reads empty regardless of what this build writes. Worth knowing about; not something this PR can close.

mobile/lib/blocs/saved_sounds/saved_sounds_bloc.dart
rabble: Both fixed in 5c5dc8e64.

`_remove` now wraps the delete: on failure it calls `addError` and returns *without* emitting, so the row stays on screen because the delete genuinely did not happen. `SavedSoundRemoveRequested` gained an optional completer and the bloc a `removeSound(id)` future mirroring `saveSound`, so the UI can tell the two outcomes apart instead of showing "Removed from Sounds" optimistically on a delete that failed.

Both `saveSound` call sites (`sounds_tab._onAddAudioTap`, `sound_detail_screen._saveSoundToLibrary`) now catch the completer error and switch on the nullable result — saved / already-saved / failed — so the failure path gets its own snackbar rather than dropping it. Two new keys (`soundsSaveFailed`, `soundsRemoveFailed`) added and declared as l10n debt alongside the rest of this PR.

New test: `keeps the row and reports the error when removal fails` asserts the future rejects and the row survives.

rabble:
All four settled in 5c5dc8e64, and the branch is rebased onto fresh `origin/main` (`f115ba30a` at the time of the push) so the checks now run against a current base. Details are inline on each thread; the short version:

1. **Consent resolver** — newest revision wins now, so a post-publish revocation is honoured. The tie guard stays and is complete under the descending sort.
2. **Catalog sound** — a missing `creator` falls back to the provider name, and a failed bridge only blocks the publish when the creator actually asked for reusable audio. That matches what the description already promised.
3. **Versioned store** — reads any schema at or below current instead of returning `[]`, and refuses to overwrite a payload from a newer build rather than flattening it.
4. **`_remove`** — catches, reports, and keeps the row; both `saveSound` call sites now handle the completer error so a failed save says so.

Verification on the rebased tree: `flutter analyze lib test` clean, 271 tests green across `test/l10n/`, `test/blocs/saved_sounds/`, `saved_sounds_service_test.dart`, `audio_reuse_consent_resolver_test.dart`, `video_event_publisher_test.dart`, `test/widgets/library/`, and `sound_detail_screen_test.dart`, plus all six ratchets (untested-services, test-structure, raw TextStyle/colors/material-button/raw-dialog).

@realmeylisdev ready for another pass when you have time.

rabble:
The live branch now includes the review-blocker fixes: newest consent revision wins, saved-library schema handling preserves readable older data and refuses destructive downgrades from newer payloads, provider credit failures no longer strand non-reuse publishes, and saved-library persistence errors are surfaced. Local focused coverage passed; current GitHub checks are still running.

rabble:
Rebased onto current origin/main (12ed8c94eefc5007134589b3aa7bd1a7860fabbc) and force-pushed with --force-with-lease. The rebase resolved only the ARB debt-list conflicts, preserving the saved-sound keys that remain intentionally tracked. Local verification passed: focused saved-sounds/audio/publishing/l10n tests, flutter analyze, formatting, and all 525 changed-file tests.

rabble:
Follow-up: origin/main advanced with #6532 during the first push, so I rebased once more. The PR is now based on current origin/main a188fd1c5 and published at b90eed0ad. The full pre-push suite passed again (locale checks, analyzer, generated files, and 525 changed-file tests).

mobile/lib/services/audio_reuse_consent_resolver.dart
realmeylisdev: Checked this one. The sort is descending now and `matching.first` is the newest revision, so a dropped `allow_audio_reuse` on an edit revokes. Your reasoning on the tie guard holds: everything in `matching` shares one `addressableId`, so only a tie at the max can change the answer and that still fails closed.

`honours a revocation on the newest revision` goes red under the old ordering, so it's pinned.

mobile/lib/internal/video_event_publisher_audio.dart
realmeylisdev: Both sides check out. `creatorName` falls back to the non-nullable `external.providerName`, so a `creator NULL` row credits Freesound instead of losing the publish, and the block is gated on `allowAudioReuse` — a creator who never asked for reuse now ships the video without the provider reference.

The three split tests all fail against the previous behaviour.

mobile/lib/services/saved_sounds_service.dart
realmeylisdev: Verified. `_isNewerSchema` only rejects payloads above the current version, so a v1 payload reads through a v2 build instead of getting wiped, and `_isForeignPayload` stops the write before it can flatten anything.

`refuses to overwrite a payload from a newer build` asserts the stored string is byte-identical afterwards, which is the part I cared about. Fair call on keeping the hand-rolled loop — losing a whole library to one bad row is the worse failure, and `skips a corrupt versioned entry` pins it. Rollback noted.

mobile/lib/blocs/saved_sounds/saved_sounds_bloc.dart
realmeylisdev: Verified. `_remove` addErrors and returns before the emit, so the row survives a failed delete, and both `saveSound` call sites now switch on the nullable result instead of dropping the completer error.

`keeps the row and reports the error when removal fails` covers it.

mobile/lib/widgets/library/saved_sound_card.dart
```diff
@@ -0,0 +1,298 @@
...
+  Widget build(BuildContext context) {
+    final source = sound.sourceContext;
+    final duration = sound.audio.formattedDuration;
+    final displayTitle = _displayTitle(context);
+    return Semantics(
+      label: duration.isEmpty ? displayTitle : '$displayTitle, $duration',
+      container: true,
+      child: DecoratedBox(
+        decoration: BoxDecoration(
+          color: VineTheme.cardBackground,
```
realmeylisdev, line 46: Non-blocking, and it's a design call rather than a code one.

The card paints a fixed `VineTheme.cardBackground` while `sounds_tab` and `library_screen` around it are fully palette-aware, so on the light theme this is a dark card sitting on a white page.

In 51db652c1 I matched the two text colors to this fill rather than converting the card, because converting isn't purely mechanical: `_SavedSoundTag` puts `vineGreen` text on a `vineGreen` 16% fill, which measures 5.90:1 over `#1A1A1A` but 1.95:1 over white, and the light palette has no `onPrimaryContainer`-style token to swap in.

Everything else here does have an exact dark alias (`darkColors.card == cardBackground`, same for `surfaceContainer`, `onSurface`, `onSurfaceVariant`), so if design does want a light card the swap is a no-op in dark mode apart from that chip. Happy to leave it for a follow-up.

realmeylisdev (approved):
all four from last time hold up, replied and closed each thread.

also pushed 51db652c1: on the light theme the card title/description came out at 1.36:1 and the editor's device-only caption at 1.0:1, so i matched each color to the fill behind it and added light+dark tests. revert freely. one non-blocking note inline.

NotThatKindOfDrLiz (approved):
Takeover fixes pushed through 4fe0febeebc3f554e9b6cad61315b7fcd99e82d7.

Addressed the blocking gaps from the re-review:
- removed the dead Riverpod saved-sounds notifier/test and dead replaceAll APIs
- added user feedback when reuse consent cannot be verified in the audio picker
- reseeded and validated public audio attribution for imported/local reusable audio
- stopped raw pubkeys from leaking into public Kind 1063 creator/title tags
- restored Library -> saved sound detail navigation from saved sound cards
- fixed the raw-color CI ratchet introduced by the tappable card wrapper

Local verification passed: flutter analyze lib test; targeted saved-sound/metadata/publisher/l10n tests; pre-push analyzer, generated-file verification, and changed-file tests. GitHub checks are green on the latest head.

@rabble please review the takeover commits and merge when ready.

PR #6530, opened 2026-07-31
fix(theme): finish the light-mode contrast pass #6467 left open

### LINKED ISSUE #6531: fix(theme): finish the light-mode contrast pass #6467 left open
Auto-created from PR #6530.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6530
**Branch:** `fix/light-mode-contrast-followups`

realmeylisdev:
Closes #6531

## What

#6467 turned light mode on for every user. Eight contrast pairs it left behind are fixed here, one commit each so any can be reverted alone.

Five of these were reviewed and approved on #6467 on 2026-07-30 (`34d54fbb2`..`d59d4daae`) but were dropped by a force-push shortly before it merged, so they never reached `main`. The other three come from the chroma-key feature (#6483) and the flag sheet (#6312), which landed on `main` after #6467's base was cut — the sweep never saw them, and they regress because #6467 moved shared `divine_ui` defaults from fixed constants to palette tokens.

Every ratio below was measured by pumping the widget under `VineTheme.lightTheme` and reading the resolved colors, not derived from the token table.

## Fixes

| Commit | Light before → after | Dark |
| --- | --- | --- |
| `ghostOverMedia` variant + 14 over-media sites | 1.31–2.47:1 → 17.31:1 | unchanged |
| Animation-picker selected chip | fills 1.001:1, borders 1.37:1 → visible | unchanged |
| Recorder mode wheel label + pill edge | 1.92:1 → readable | unchanged |
| Profile-setup field underline | 1.00:1 → 1.37:1 | unchanged |
| Chroma-key screen canvas (close button + 3 slider thumbs + panel text + picker sheet) | 1.00–1.50:1 → 5.78–10.11:1 | canvas and hint move, see below |
| Flag sheet Submit label | 2.08:1 → 8.49:1 | 9.45 → 8.49:1 |
| Flag sheet selected reason indicator | 2.22:1 → 3.17:1 | unchanged |
| Comment Delete row | 4.13:1 → 9.11:1 | 4.57:1, byte-identical |
| Editor add-element FAB glyph | 1.92:1 → 11.06:1 | unchanged |

`ghostSecondary`'s fill is a fixed `scrim15` while its icon follows the palette, so `colors.isLight` cannot tell whether the caller put the button over media or over a themed surface. `ghostOverMedia` keeps the same scrim with a fixed light icon and serves the over-media sites; `ghostSecondary` now serves themed surfaces only. @hm21 caught two of the migrated sites that are **not** over media — the profile nav buttons (no banner and no `profileColor` is the default, and the backdrop is then `ProfileBanner`'s palette gradient) and the crop editor's close button (`CropRotateEditor` puts the bar in a plain `Scaffold.appBar`, no `extendBodyBehindAppBar`, over the scaffold's own `colors.surface`) — and pushed the revert in `af1d72a0e`. Both verified here: the package has no `extendBodyBehindAppBar` anywhere, and `profile_header_media.dart:134-141` is the gradient. That takes the migration from 18 buttons to 14.

## Chroma-key screen: what moves on dark

The chroma-key screen painted a fixed near-black canvas while hosting palette-following chrome. It now takes `surfaceContainerHigh`, matching `video_editor_scaffold.dart:50` — the editor this screen is launched from. Two things do change on dark, so the row above is not "unchanged":

- **Canvas** `surfaceBackground` #00150D → `surfaceContainerHigh` #000A06. Every light-on-dark pair on the screen gains a little: the title goes 16.97:1 → 18.05:1, the slider readouts 10.68:1 → 11.19:1.
- **Transparent-background hint** `onSurfaceMuted` (50% white, 5.30:1) → `onSurfaceVariant` (75% white, 11.19:1). It was the only piece of secondary text in the panel still on the muted register; the preview-unavailable notice and the slider readouts already used the variant on both palettes.

Neither is a regression, but both were described as no-ops in an earlier revision of this table. `chroma_key_controls_test.dart` now pins the hint on dark, so the claim cannot drift again — reverting it to `onSurfaceMuted` turns that test red.

Left fixed on purpose, and verified as such: the chroma baking overlay's own 85% scrim (still 13.26:1 over the lighter canvas), the transparency checkerboard inside the video frame, the swatch border on a user-chosen colour, and the editor canvases that deliberately stay fixed dark.

## Known issues, not fixed here

`BrandedLoadingIndicator` draws a pure-white sprite and exposes no `color` parameter at all, across 41 call sites in 33 files. On any palette `surface` that is 1.00:1 — an invisible spinner, so the chroma clip-picker looks hung while it loads. Tinting the brand loading animation app-wide is a design call, so it is left for design rather than decided in this PR.

The profile-setup field underline lands at 1.37:1, not the 3:1 a non-text boundary should meet — @hm21's second note. `outline` #B7C9C1 is the *strongest* of the three light outline tokens (`outlineMuted` #DCE7E2 is the 1.00:1 this commit replaces, `outlineDisabled` #EBF1EE is weaker still), so no call-site swap reaches 3:1 without borrowing a text token and mispairing its role. Dark has shipped the same underline at 1.18:1 since before light mode existed, so 1.37:1 leaves light marginally ahead of the treatment already in production rather than behind it. Closing the gap properly means raising `lightColors.outline` for every outline in the app, or adding a stronger token — a palette change design owns, not a review round.

## Verification

Re-run at `af1d72a0e` + the test commit, not carried over from an earlier revision:

- 1908 tests over `test/screens/{comments,image_crop_editor,profile_setup,video_editor,feed}`, the three `video_metadata_*_screen` suites, and `test/widgets/{profile,video_editor,video_metadata,video_recorder}` — pass, 4 pre-existing skips
- `divine_ui` — 723 tests, 100.00% lines (1771/1771)
- All 7 design-system ratchets — pass
- `mobile/scripts/golden.sh verify` — unchanged

New tests are falsifiable, each confirmed by running it against the pre-fix code: reverting the chroma-key migration turns `chroma_key_controls_test.dart` red on both palettes, and the two `ghostSecondary` reverts in `af1d72a0e` are pinned by `profile_header_widget_test.dart` and `image_crop_editor_toolbar_test.dart`.

### DIFF
--- a/mobile/lib/screens/comments/widgets/comment_options_modal.dart
+++ b/mobile/lib/screens/comments/widgets/comment_options_modal.dart
@@ -168,7 +168,10 @@ class _OptionTile extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final color = isDestructive
-        ? VineTheme.likeRed
+        // The sheet surface follows the palette now, and fixed `likeRed` only
+        // holds 4.13:1 on the light one. The palette token is `likeRed`
+        // verbatim on dark, so that rendering is unchanged.
+        ? context.vineColors.onErrorContainer
         : context.vineColors.onSurface;
 
     return Semantics(
@@ -291,7 +294,10 @@ class _FlagContentSheetState extends State<_FlagContentSheet> {
                     ? VineTheme.vineGreen
                     : context.vineColors.containerLow,
                 foregroundColor: _selectedReason != null
-                    ? context.vineColors.background
+                    // The fill above is the fixed brand green, so its label
+                    // has to be fixed too — the palette `background` token
+                    // goes near-white on light and drops to 2.08:1.
+                    ? VineTheme.onPrimary
                     : context.vineColors.onSurfaceMuted,
                 shape: RoundedRectangleBorder(
                   borderRadius: BorderRadius.circular(24),
@@ -302,7 +308,7 @@ class _FlagContentSheetState extends State<_FlagContentSheet> {
                 context.l10n.commentOptionsFlagSubmit,
                 style: VineTheme.labelLargeFont(
                   color: _selectedReason != null
-                      ? context.vineColors.background
+                      ? VineTheme.onPrimary
                       : context.vineColors.onSurfaceMuted,
                 ),
               ),
@@ -313,6 +319,16 @@ class _FlagContentSheetState extends State<_FlagContentSheet> {
     );
   }
 }
+
+/// Accent for the selected reason's ring and dot.
+///
+/// The ring is the only thing that marks which of the eleven reasons is
+/// chosen, and the brand green reaches just 2.22:1 on the light sheet
+/// surface. [VineTheme.primaryAccessible] is the same hue at 3.17:1, which
+/// clears the 3:1 floor a non-text indicator has to meet.
+Color _selectedIndicatorOf(BuildContext context) => context.vineColors.isLight
+    ? VineTheme.primaryAccessible
+    : VineTheme.vineGreen;
 
 class _ReasonRadioTile extends StatelessWidget {
   const _ReasonRadioTile({
@@ -351,7 +367,7 @@ class _ReasonRadioTile extends StatelessWidget {
                   shape: BoxShape.circle,
                   border: Border.all(
                     color: isSelected
-                        ? VineTheme.vineGreen
+                        ? _selectedIndicatorOf(context)
                         : context.vineColors.onSurfaceMuted,
                     width: 2,
                   ),
@@ -361,9 +377,9 @@ class _ReasonRadioTile extends StatelessWidget {
                         child: Container(
                           width: 10,
                           height: 10,
-                          decoration: const BoxDecoration(
+                          decoration: BoxDecoration(
                             shape: BoxShape.circle,
-                            color: VineTheme.vineGreen,
+                            color: _selectedIndicatorOf(context),
                           ),
                         ),
                       )
--- a/mobile/lib/screens/comments/widgets/video_comment_player.dart
+++ b/mobile/lib/screens/comments/widgets/video_comment_player.dart
@@ -233,7 +233,7 @@ class _OpenVideoButton extends StatelessWidget {
       icon: DivineIconName.arrowUpRight,
       semanticLabel: context.l10n.commentsOpenVideoLabel,
       size: DivineIconButtonSize.small,
-      type: DivineIconButtonType.ghostSecondary,
+      type: DivineIconButtonType.ghostOverMedia,
       onPressed: onTap,
     );
   }
@@ -280,7 +280,7 @@ class _MuteButton extends StatelessWidget {
           ? context.l10n.commentsUnmuteVideoReplyLabel
           : context.l10n.commentsMuteVideoReplyLabel,
       size: DivineIconButtonSize.small,
-      type: DivineIconButtonType.ghostSecondary,
+      type: DivineIconButtonType.ghostOverMedia,
       onPressed: onTap,
     );
   }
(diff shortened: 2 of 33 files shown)


hm21 (approved):
Solid work — I recomputed every ratio in the table (`likeRed` 4.13:1, `onErrorContainer` 9.11:1, `onPrimary`-on-`vineGreen` 8.49:1, `primaryAccessible` 3.17:1, `outlineMuted`-on-`surfaceContainerHigh` 1.00:1) and they all hold.

Two of the eighteen `ghostOverMedia` sites aren't over media, though, so the fixed light glyph goes near-invisible on light. Pushed as `af1d72a0e3` — revert to `ghostSecondary` at both, plus a falsifiable test each:

- **`profile_header_widget.dart`** — the nav buttons are only over content when the profile has a banner image or a `profileColor`. With neither (the default), the backdrop is `ProfileBanner`'s palette gradient `containerLow` → `surface` (`profile_header_media.dart:134-141`), and the banner scrim is fully transparent at that height → ~1.5:1 over `#EDF3EF`.
- **`image_crop_editor_toolbar.dart`** — `CropRotateEditor` puts this in a plain `Scaffold.appBar` with no `extendBodyBehindAppBar`, so it sits above the image on the scaffold's own background, which `image_crop_editor_screen.dart:180` sets to `colors.surface` = `#FFFFFF` on light → ~1.4:1.

Dark is byte-identical either way — both variants resolve `VineTheme.onSurface` there — so no goldens move. `flutter analyze lib test integration_test` clean, all four design-system ratchets pass, and both new tests go red against the pre-fix code (verified).

Making the profile chrome adapt to whether a banner is actually present would be the better fix, but that's a design question rather than a review one, so I left it alone.

Two non-blocking notes, your call:

- `chroma_key_controls.dart:370` swaps the transparent hint from `VineTheme.onSurfaceMuted` (`0x80FFFFFF`) to `context.vineColors.onSurfaceVariant`, which is `0xBFFFFFFF` on dark — so dark rendering does change, contrary to "unchanged" for that row. The `is unchanged on the dark palette` test only asserts `contains(darkColors.onSurface)`, so it doesn't pin it either way. Brighter hint is arguably fine; the table just needs to say so.
- `profile_setup_field_decorations.dart:14` — `outline` on `surfaceContainerHigh` is `#B7C9C1` on `#E7E4E1` = 1.37:1. Real improvement over 1.00:1, but the field boundary is still effectively invisible, and it's under the 3:1 floor you apply to the flag-sheet indicator two commits over. Probably a token problem rather than a call-site one.

Approving — please review the pushed commit and merge when you're happy with it.

realmeylisdev:
Thanks — both reverts hold up, and I checked each claim rather than taking the ratio on faith. `pro_image_editor` 13.2.3 has no `extendBodyBehindAppBar` anywhere in the package, and `CropRotateEditor`'s `Scaffold` (`crop_rotate_editor.dart:2328`) takes `backgroundColor` straight from `style.background`, which `image_crop_editor_screen.dart:180` sets to `colors.surface`. And `ProfileBanner` does fall through to the `containerLow` → `surface` gradient when `bannerUrl` and `profileColor` are both null, which is the default.

**Hint** — you're right, and there is a second dark change on that row I had missed: the canvas itself. `surfaceBackground` #00150D → `surfaceContainerHigh` #000A06, which moves the title from 16.97:1 to 18.05:1 and the slider readouts from 10.68:1 to 11.19:1. Both that and the hint (5.30:1 → 11.19:1) are gains, but neither is the no-op the table claimed. Description now spells both out, and `chroma_key_controls_test.dart` pins the hint on dark in 32e54af9c — putting it back to `onSurfaceMuted` turns that test red, confirmed by running it. I also dropped the `voice_over_recorder_screen.dart:72` citation from the description: that screen uses `colors.surface`, so `video_editor_scaffold.dart:50` is the only real precedent for the canvas token.

**Underline** — leaving it at 1.37:1 and documenting it rather than patching the call site. `outline` is the strongest of the three light outline tokens (`outlineMuted` #DCE7E2 is the 1.00:1 this replaces, `outlineDisabled` #EBF1EE is 1.11:1), so nothing reaches 3:1 without borrowing a text token and mispairing its role. Dark has shipped this same underline at 1.18:1 since before light mode existed, so 1.37:1 leaves light slightly ahead of what is already in production rather than behind it — which is why I read it as a palette gap rather than a regression this PR introduced. Closing it properly means raising `lightColors.outline` app-wide or adding a stronger token, and that is design's call.

Re-ran everything at your commit plus the test: analyze clean, 1908 tests across the touched areas, `divine_ui` 723 tests at 100.00% (1771/1771), all 7 ratchets, and `golden.sh verify` unchanged.

PR #6532, opened 2026-07-31
fix(settings): hide Experimental Features behind developer mode

### LINKED ISSUE #6533: fix(settings): hide Experimental Features behind developer mode
Auto-created from PR #6532.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6532
**Branch:** `fix/hide-experimental-features-menu`

rabble:
Closes #6533

## What

The **Experimental Features** tile (feature-flag menu) was visible in Settings to every user. With launch approaching it now sits behind the existing developer-mode unlock — 7 taps on the version tile — alongside **Developer Options**.

- Gated the tile on `isDeveloperModeEnabledProvider`.
- Moved it next to Developer Options so the two dev-only entries read as one group.

Note: internal-only flags inside `FeatureFlagScreen` were already gated on developer mode; this closes the entry point itself.

## Testing

- New: `hides Experimental Features tile when developer mode is off` — pumps on a tall surface so the whole lazy `ListView` is laid out (otherwise the absence proves nothing) and asserts the tile is gone while a neighbouring tile still renders.
- New: `shows Experimental Features tile and opens it when developer mode is on` — asserts the tile renders and taps through to `FeatureFlagScreen`.
- `flutter test test/widgets/settings_screen_test.dart` — 26/26 pass.
- Also green: settings taxonomy, feature-flag suites, settings scaffold/bottom-nav, route coverage.

## Base note

Branched from `5720691cf` rather than current `origin/main`: `origin/main` at `46f4eb7fb` (#6350) fails analyze — `general_settings_screen.dart:85,89` reference `_titleStyle` / `_subtitleStyle`, but the file defines `_titleStyleOf(context)` / `_subtitleStyleOf(context)`. That is unrelated to this diff and needs its own fix.

### DIFF
--- a/mobile/lib/screens/settings/settings_screen.dart
+++ b/mobile/lib/screens/settings/settings_screen.dart
@@ -388,17 +388,6 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
                     onTap: () => context.push(AppsDirectoryScreen.path),
                   ),
                 _SettingsTile(
-                  icon: Icons.science,
-                  title: context.l10n.settingsExperimentalFeatures,
-                  subtitle: context.l10n.settingsExperimentalFeaturesSubtitle,
-                  onTap: () => Navigator.push(
-                    context,
-                    MaterialPageRoute(
-                      builder: (_) => const FeatureFlagScreen(),
-                    ),
-                  ),
-                ),
-                _SettingsTile(
                   title: context.l10n.settingsLegal,
                   icon: Icons.gavel,
                   onTap: () => context.push(LegalScreen.path),
@@ -409,7 +398,18 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
                   subtitle: context.l10n.settingsIntegrationPermissionsSubtitle,
                   onTap: () => context.push(AppsPermissionsScreen.path),
                 ),
-                if (isDeveloperMode)
+                if (isDeveloperMode) ...[
+                  _SettingsTile(
+                    icon: Icons.science,
+                    title: context.l10n.settingsExperimentalFeatures,
+                    subtitle: context.l10n.settingsExperimentalFeaturesSubtitle,
+                    onTap: () => Navigator.push(
+                      context,
+                      MaterialPageRoute(
+                        builder: (_) => const FeatureFlagScreen(),
+                      ),
+                    ),
+                  ),
                   _SettingsTile(
                     divineIcon: DivineIconName.bracketsAngle,
                     title: context.l10n.settingsDeveloperOptions,
@@ -417,6 +417,7 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
                     iconColor: VineTheme.warning,
                     onTap: () => context.push(DeveloperOptionsScreen.path),
                   ),
+                ],
 
                 const SizedBox(height: 24),
                 _VersionTile(appVersion: _appVersion),
--- a/mobile/lib/screens/settings/settings_screen.dart
+++ b/mobile/lib/screens/settings/settings_screen.dart
@@ -398,18 +398,18 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
                   subtitle: context.l10n.settingsIntegrationPermissionsSubtitle,
                   onTap: () => context.push(AppsPermissionsScreen.path),
                 ),
-                if (isDeveloperMode) ...[
-                  _SettingsTile(
-                    icon: Icons.science,
-                    title: context.l10n.settingsExperimentalFeatures,
-                    subtitle: context.l10n.settingsExperimentalFeaturesSubtitle,
-                    onTap: () => Navigator.push(
-                      context,
-                      MaterialPageRoute(
-                        builder: (_) => const FeatureFlagScreen(),
-                      ),
+                _SettingsTile(
+                  icon: Icons.science,
+                  title: context.l10n.settingsExperimentalFeatures,
+                  subtitle: context.l10n.settingsExperimentalFeaturesSubtitle,
+                  onTap: () => Navigator.push(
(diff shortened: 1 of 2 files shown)


hm21 (commented):
Heads-up on the audience split: `FeatureFlagAudience { user, internal }` was added in #6409 (@NotThatKindOfDrLiz) precisely so user-audience flags stay visible to everyone and only the internal ones need developer mode. Only 4 of 17 flags are `internal` — gating the entry point hides the other 13 (Light Mode, Advanced Relay Settings, Bluesky Publishing, Feed Tuning, Divine Supporters, …) as well, and anyone who already switched one on can no longer switch it off without the 7 taps.

It also makes the split itself dead: while the screen is reachable, developer mode is on by definition, so `!flag.isInternal || showInternalFlags` in `feature_flag_screen.dart:23` is always true, and the persisted-internal-override handling in `feature_flag_service.dart` never applies.

So if dev-mode-only is the call for launch, the internal/audience machinery from #6409 can be deleted in the same PR instead of staying as vestigial code. If user-facing opt-in should survive launch, keep the tile visible and let #6409's filter do the gating.

rabble:
Review findings — changes needed before merge

Blocking: mobile/lib/screens/settings/settings_screen.dart:401-420 now makes the entire Experimental Features entry point developer-only, but the existing feature-flag contract is audience-aware. FeatureFlagScreen still filters only internal flags (feature_flag_screen.dart:21-24), so this change hides all user-audience flags as well (for example Advanced Relay Settings, Bluesky Publishing, Feed Tuning, Divine Supporters, and any persisted user overrides). It also makes the audience split effectively unreachable for normal users. Please either keep the entry point available and let the screen hide only internal flags, or deliberately convert/remove the user-audience flags and update the contract/tests. This matches the unresolved concern already raised by hm21.

mbradley (commented):
The over-gating concern is already in hm21's review. This is only what that thread doesn't have.

**The decision this reverses.** #6467 promoted `lightMode` from `internal` back to `user` deliberately ("this exposes the toggle rather than switching the theme on for anyone"), and restored the word "experimental" to its description because disclosure got weaker as reach got wider. That landed 07:38 UTC on Jul 31; this PR opened at 08:47. The report underneath all of it, #6404, was Light Mode leaving the UI dark: Liz's #6409 gated it because it was broken, #6467 fixed it and re-exposed it.

So the question isn't whether to tighten experimental surfaces before launch. It's whether that case survives reasoning made an hour earlier by the person who had just fixed the flag. As written the PR reads as unaware of #6467 rather than disagreeing with it.

**Correction to hm21's review.** The "#6409 machinery goes dead" read is right about the screen filter, which becomes a tautology once the screen is dev-only. It isn't right about the service: `canOverrideInternalFlags` is wired app-wide at `feature_flag_providers.dart:29`, so `_canOverrideFlag` still suppresses persisted internal overrides on every flag read. Before #6409 the screen listed `FeatureFlag.values` unfiltered, so older installs can hold keys that are internal today, and deleting that half would re-apply them. Changes the remedy, not the conclusion.

**Tests.** The new negative test does go red when the guard is removed, so it's real. Two notes:

- Its control doesn't prove what the comment claims. `settingsLegal` sits two tiles above the gated block, so it renders whether or not layout reached the block. Asserting on `settingsIntegrationPermissions` or the version tile would prove it. One line.
- The sibling at line 771 sets no surface size, and at the default 800x600 nothing in that list renders at all. It stays green with Developer Options rendered unconditionally, so it cannot fail. Same for the precondition at 862. Pre-existing, though this PR's own new comment is about exactly that hazard.

Both test items are clear enough that I'd normally just push them. Holding because the design question may rework or close this PR, and the line-771 one is pre-existing so it wants its own PR either way.

**Nit:** the base note's analyze break was real at `46f4eb7fb`, and was fixed the same day by #6534, already in this branch. Worth dropping so it doesn't read as a live warning about main.

rabble (commented):
The change itself is right, and the new tests are a step up on the ones
they sit next to — the `setSurfaceSize(900, 2000)` plus the comment
explaining *why* the lazy `ListView` makes an unguarded absence-assertion
worthless is exactly the reasoning that was missing before.

One thing worth tightening before merge.

**The negative assertion anchors on a tile that renders too early.**

In `hides Experimental Features tile when developer mode is off`:

```dart
expect(find.text(l10n.settingsLegal), findsOneWidget);
expect(find.text(l10n.settingsExperimentalFeatures), findsNothing);
```

After this PR the order is Legal → Integration Permissions → *[dev-only:
Experimental Features, Developer Options]* → version tile. So
`settingsLegal` now renders **before** the block under test. Proving Legal
was laid out doesn't prove the list got as far as where Experimental
*would* be — if a future change let the viewport cut off between them, the
`findsNothing` passes for the wrong reason and the test silently stops
guarding anything.

`settingsIntegrationPermissions` is the tile immediately preceding the
dev-only block, so it's a tight anchor and needs no new l10n key:

```dart
expect(find.text(l10n.settingsIntegrationPermissions), findsOneWidget);
expect(find.text(l10n.settingsExperimentalFeatures), findsNothing);
```

Not blocking — the 900x2000 surface almost certainly lays out the whole
list today. It's about keeping the test honest under a layout change.

**Adjacent, pre-existing:** `hides Developer Options tile when developer
mode is off` has no `setSurfaceSize` and no positive anchor at all, so it
may already be passing vacuously. Since this PR moves a tile into exactly
that `if (isDeveloperMode)` block, giving that test the same treatment is
local and low-risk — your call whether it's in scope here.

rabble:
Updated this branch on current main. The Experimental Features entry is reachable without developer mode so user-facing flags remain discoverable; internal flags and Developer Options stay developer-mode gated. Settings and feature-flag tests, analyzer, and format checks pass.

PR #5564, opened 2026-06-26
feat(feed): add Classics mode to the home feed picker

hm21:
## Description

Adds a **Classics** option to the home feed mode picker, alongside For You / New / Following and subscribed lists. It surfaces popular classic Vine archive videos — the same content as Explore → Classics.

Selecting it from the bottom sheet wires a new feed source through the full stack:

- **State** — new `VideoFeedSourceType.classic` / `FeedMode.classic` / `VideoFeedSource.classic()` with their `mode`, `labelFallback`, and `persistenceValue` cases.
- **Repository** — new `VideosRepository.getClassicVideos()` wrapping `getPopularVideosPage(variant: classic)` into a `HomeFeedResult` (keeps the `PopularVideosVariant` enum out of the bloc layer).
- **Bloc** — `_fetchVideosForSource` classic branch plus a `_usesCursorPagination` helper so Classics paginates via the server cursor (no `createdAt` re-sort), exactly like For You. Classic is cache-eligible for cold-start/resume.
- **Persistence** — `FeedModePreferenceStore` restores the persisted `classic` selection across restarts.
- **UI** — Classics row in the picker, label switches, and a dedicated empty-state message.
- **l10n** — `feedModeClassics` + `feedClassicEmpty` added to all 18 locales (reusing each locale's existing "Classics" noun), regenerated.

**Behavior notes:** Classics has full parity with the other home modes — the selection persists across restarts and it paginates on infinite scroll. Like Explore → Classics, it is Funnelcake-only: when no Funnelcake relay is connected it shows the empty state rather than falling back to relays. It is intentionally excluded from auto-refresh-on-resume (the popular-month window is stable, so refreshing would needlessly reset scroll position).

**Related Issue:** Closes #5486

## Out of Scope

- Auto-refresh-on-resume for the Classics mode (intentional — see behavior notes).

## Verification

- `flutter analyze lib test integration_test` — clean
- `flutter test test/screens/feed/feed_mode_switch_test.dart test/blocs/video_feed/video_feed_bloc_test.dart test/screens/feed/video_feed_page_test.dart` — all pass (added: picker option + dispatch, persistence restore, cursor pagination, classic label/empty-state)
- `packages/videos_repository` full suite with `--coverage` — 453 pass; `getClassicVideos` fully covered (added: classic delegation/mapping + Funnelcake-unavailable paths)

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

PR #6537, opened 2026-07-31
fix(feed): randomize the home Classics feed like Discover

### LINKED ISSUE #6538: fix(feed): randomize the home Classics feed like Discover
Auto-created from PR #6537.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6537
**Branch:** `fix/home-classics-random-slice`

hm21:
Closes #6538

## Description

The home feed's Classics mode and Discover's Classics tab read two different sources, and only one of them varied. Home Classics went through `getClassicVideos`, which delegated to the v2 popular feed with the classic variant — server-ranked and cursor-paged, so every user opened on the same top-of-month vines in the same order on every session. Discover's Classics tab reads the offset-paginated archive (`sort=loops`, which the server serves views-first on this path) with a random start offset plus a per-page shuffle. The split was deliberate when Classics was added to the home picker (#5564) and is documented in the old doc comment, but the two surfaces ended up feeling like different features.

This points `getClassicVideos` at the same archive Discover uses: draw a page-aligned random offset within the top 400 on every fetch that misses the cache, shuffle the resolved page, and carry the next offset in `paginationCursor` (`classic-offset:<n>`) so the home feed keeps paginating through the existing cursor plumbing rather than needing a new pagination shape. Classics already opts out of the `createdAt` re-sort on page merge (`_usesCursorPagination`), so the shuffled order reaches the feed intact instead of being undone. Both surfaces now draw their random start from one exported constant (`classicsMaxRandomStartOffset`) so the window cannot desync between them.

**Why the fetch tops up:** filtering — blocked authors, content preferences, transport scheme, non-vine rows — can empty an otherwise full archive page, and `VideoFeedBloc` stops paginating once the feed list is empty, so an empty first page would dead-end Classics until a manual refresh. The delegating implementation this replaces looped until it had a full page of *visible* videos; the archive read does the same, bounded at three pages so a systematically filtered slice cannot turn one fetch into a crawl through the archive. Exhaustion is read from the page the client returned rather than the filtered list, so a fully-filtered page advances the cursor instead of ending the feed. It means "the archive returned nothing at this offset", not "the page came back short": the client drops rows with a missing id or URL before the repository sees them, so a short page is not an end-of-archive signal — one dropped row would otherwise end the feed mid-archive. The cost is one extra request at the true end of the archive. A topped-up page is handed back whole rather than trimmed to `limit`, because the offset cursor has already advanced past every page read: trimming would drop those videos rather than defer them to the next page. A request that fails partway serves the pages that already succeeded and is not pinned in the session cache.

**Why the cache still matters:** only pull-to-refresh passes `skipCache: true`, and that is what draws a fresh slice. Picking Classics from the home picker goes through `_loadVideos` with the default `skipCache: false` and so resumes the pinned first page — deliberately, so navigating away and back does not reshuffle the feed out from under the user's resume position. (Auto-refresh never applies here; `_onAutoRefreshRequested` returns early for anything other than Following and For You.) The cache holds the *first* page only, so pages 2+ come back re-drawn and re-shuffled on re-entry, and it is in-memory, so a cold start begins empty and draws a fresh slice. That mirrors the Discover tab, where `refresh()` picks a new offset but an ordinary rebuild does not. An empty page is never cached: a slice that filters down to nothing would otherwise be pinned as `videos: []` + `hasMore: true`, which the feed can neither paginate out of nor re-fetch.

**Why `until` was dropped from the signature:** the archive has no time-window pagination — the funnelcake client only honours `before` when `sort == 'recent'`. The parameter was silently ignored on this path, so removing it makes that explicit rather than leaving a dead pass-through in the bloc's call site.

Explore → Popular's "Classic" toggle is deliberately left on the ranked leaderboard via `getPopularVideosPage`; that surface is meant to be ordered, so home and Discover now match while Popular-Classic intentionally does not.

**Related Issue:** 

## Out of Scope

Explore → Popular's "Classic" toggle keeps its server-ranked ordering — it is a leaderboard, not a discovery feed.

Classics remains Funnelcake-only and returns an empty result when no Funnelcake relay is connected, unchanged from before. Relays do not expose the server-side `platform` filter.

`ClassicVinesFeed` still calls the funnelcake client directly rather than going through the repository. Only the shared window constant is unified here; moving that provider behind the repository is a larger migration than this fix.

If all three archive pages in one fetch filter down to nothing, the feed still lands empty and cannot self-advance within that entry, because `VideoFeedBloc` gates load-more on a non-empty list. Relaxing that guard also changes For You, so it is left alone here; since the empty page is not cached, refresh or the next entry redraws.

## Verification

- `flutter test --coverage` in `mobile/packages/videos_repository` — 456 passing, **100% line coverage** held (the VGV gate default; this package sets no `min_coverage`)
- `flutter test mobile/test/blocs/video_feed` — 116 passing; `mobile/test/providers/classic_vines_refresh_test.dart` + `mobile/test/screens/feed` — 155 passing
- `flutter analyze` clean on `mobile/lib` and `mobile/packages/videos_repository`
- `dart format` clean on all changed files; `build_runner` re-run after the provider edit (regenerated provider hash committed)

Seventeen cases in the `getClassicVideos` group cover the page-aligned random offset draw at both the group's `limit: 20` and the production default of 25, the sort the shared window depends on, the shuffle, slice variation across successive cache-missing fetches, cursor advance and resume, archive exhaustion, a short page that is *not* the archive end, top-up over a fully-filtered page, the top-up bound, the un-trimmed page size a top-up produces, all three cache-write guards (first-page-only, non-empty, not-after-a-failure), cache stability across cache-eligible fetches, partial results surviving a failed top-up, non-vine filtering, and both empty-result paths. Randomness is injected via an optional `Random` constructor parameter, and the tests script the offset draws by position so an assertion cannot pass by chance. Each guard was mutation-checked: removing it turns exactly one test red.

Manually verified on a real Samsung Galaxy (SM-S942B).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/video_feed/video_feed_bloc.dart
+++ b/mobile/lib/blocs/video_feed/video_feed_bloc.dart
@@ -925,16 +925,12 @@ class VideoFeedBloc extends Bloc<VideoFeedEvent, VideoFeedBlocState> {
       _videosRepository
           .getNewVideos(until: until, skipCache: skipCache)
           .then((videos) => HomeFeedResult(videos: videos)),
-    VideoFeedSourceType.classic =>
-      paginationCursor == null
-          ? _videosRepository.getClassicVideos(
-              until: until,
-              skipCache: skipCache,
-            )
-          : _videosRepository.getClassicVideos(
-              cursor: paginationCursor,
-              skipCache: skipCache,
-            ),
+    // Classics is offset-paginated behind an opaque cursor and has no
+    // time-window pagination, so `until` does not apply.
+    VideoFeedSourceType.classic => _videosRepository.getClassicVideos(
+      cursor: paginationCursor,
+      skipCache: skipCache,
+    ),
   };
 
   void _scheduleNostrEnrichment({
--- a/mobile/packages/videos_repository/lib/src/videos_repository.dart
+++ b/mobile/packages/videos_repository/lib/src/videos_repository.dart
@@ -4,6 +4,8 @@
 // ABOUTME: Returns Future<List<VideoEvent>>, not streams -
 // ABOUTME: loading is pagination-based.
 
+import 'dart:math';
+
 import 'package:funnelcake_api_client/funnelcake_api_client.dart';
 import 'package:models/models.dart';
 import 'package:nostr_client/nostr_client.dart';
@@ -74,6 +76,16 @@ const Duration _recentRelayRefreshTimeout = Duration(seconds: 3);
 const Duration _routeRelayTimeout = Duration(seconds: 3);
 const String _popularCacheKey = 'popular';
 const String _popularLegacyBeforeCursorPrefix = 'before:';
+const String _classicsCacheKey = 'classics';
+const String _classicsOffsetCursorPrefix = 'classic-offset:';
+
+/// Highest archive offset the Classics feed may open a session on.
+///
+/// The loops-sorted archive holds roughly 10k videos; starting anywhere in
+/// the most-looped few hundred keeps the feed varied between sessions without
+/// dropping the user into the unwatched long tail. Mirrors the Explore →
+/// Classics tab so both surfaces draw from the same window.
+const int _classicsMaxRandomStartOffset = 400;
 
 /// Result of a hashtag REST feed fetch.
 class HashtagFeedVideosResult {
@@ -107,6 +119,17 @@ int? _decodePopularLegacyBeforeCursor(String? cursor) {
   return int.tryParse(
     cursor.substring(_popularLegacyBeforeCursorPrefix.length),
   );
+}
+
+String _encodeClassicsOffsetCursor(int offset) =>
+    '$_classicsOffsetCursorPrefix$offset';
+
+int? _decodeClassicsOffsetCursor(String? cursor) {
+  if (cursor == null || !cursor.startsWith(_classicsOffsetCursorPrefix)) {
+    return null;
+  }
+
+  return int.tryParse(cursor.substring(_classicsOffsetCursorPrefix.length));
 }
 
 String _popularPreferenceCacheSuffix({
@@ -153,6 +176,7 @@ class VideosRepository {
     FunnelcakeApiClient? funnelcakeApiClient,
     InMemoryFeedCache? inMemoryFeedCache,
     SeenVideoLookup? seenVideoLookup,
+    Random? random,
   }) : _nostrClient = nostrClient,
        _localStorage = localStorage,
        _blockFilter = blockFilter,
@@ -160,7 +184,8 @@ class VideosRepository {
        _warningLabelsResolver = warningLabelsResolver,
        _funnelcakeApiClient = funnelcakeApiClient,
        _inMemoryFeedCache = inMemoryFeedCache,
-       _seenVideoLookup = seenVideoLookup;
+       _seenVideoLookup = seenVideoLookup,
+       _random = random ?? Random();
 
   final NostrClient _nostrClient;
   final VideoLocalStorage? _localStorage;
@@ -170,6 +195,11 @@ class VideosRepository {
   final FunnelcakeApiClient? _funnelcakeApiClient;
   final InMemoryFeedCache? _inMemoryFeedCache;
   final SeenVideoLookup? _seenVideoLookup;
+
+  /// Randomness behind the Classics feed's per-session slice and shuffle.
+  /// Injectable so tests can pin an order.
+  final Random _random;
+
   String _recommendationSessionSeed = generateRecommendationSessionSeed();
 
   /// Clears the in-memory feed cache.
@@ -777,37 +807,83 @@ class VideosRepository {
     return visible.take(limit).toList();
   }
 
-  /// Fetches popular classic Vine archive videos for the home feed's
-  /// Classics mode.
+  /// Fetches classic Vine archive videos for the home feed's Classics mode.
   ///
-  /// Delegates to [getPopularVideosPage] with [PopularVideosVariant.classic],
-  /// so the source is identical to Explore → Popular's "Classic" toggle
-  /// (v2 popular feed, `sort=popular&period=month&platform=vine`,
-  /// cursor-paged, server-ordered) — not the standalone Explore → Classics
-  /// tab, which is offset-based and client-side shuffled. Mapped into a
-  /// [HomeFeedResult] so the home feed paginates it via
-  /// [HomeFeedResult.paginationCursor] like the For You feed. Funnelcake-only:
-  /// returns an empty result when no
-  /// Funnelcake relay is connected (relays do not expose the server-side
-  /// `platform` filter).
+  /// Reads the same offset-paginated, loops-sorted archive as the standalone
+  /// Explore → Classics tab — not Explore → Popular's "Classic" toggle, which
+  /// is server-ordered and therefore opens on the same videos in the same
+  /// order every time. Each fetch that misses the cache starts at a random
+  /// offset within [_classicsMaxRandomStartOffset] and shuffles the page, so
+  /// two sessions do not see the same opening run. The next offset rides in
+  /// [HomeFeedResult.paginationCursor], so the home feed paginates this like
+  /// the For You feed.
+  ///
+  /// Caching keeps the slice stable within a session: leaving and re-entering
+  /// the feed preserves the current ordering, while pull-to-refresh
+  /// ([skipCache] `true`) draws a fresh slice.
+  ///
+  /// Funnelcake-only: returns an empty result when no Funnelcake relay is
+  /// connected (relays do not expose the server-side `platform` filter).
   Future<HomeFeedResult> getClassicVideos({
     int limit = _defaultLimit,
-    int? until,
     String? cursor,
     bool skipCache = false,
   }) async {
-    final page = await getPopularVideosPage(
-      variant: PopularVideosVariant.classic,
-      limit: limit,
-      until: until,
-      cursor: cursor,
-      skipCache: skipCache,
-    );
-    return HomeFeedResult(
-      videos: page.videos,
-      paginationCursor: page.nextCursor,
-      hasMore: page.hasMore,
-    );
+    final isFirstPage = cursor == null;
+    if (!skipCache && isFirstPage) {
+      final cached = _inMemoryFeedCache?.get(_classicsCacheKey);
+      if (cached != null) return cached;
+    }
+
+    if (_funnelcakeApiClient == null || !_funnelcakeApiClient.isAvailable) {
+      return const HomeFeedResult(videos: [], hasMore: false);
+    }
+
+    final offset =
+        _decodeClassicsOffsetCursor(cursor) ?? _classicsOffset(limit);
+
+    try {
+      final stats = await _funnelcakeApiClient.getClassicVines(
+        limit: limit,
+        offset: offset,
+      );
+
+      final videos = _filterPopularVideosForVariant(
+        await _hydrateVideosWithBulkStats(
+          _transformVideoStats(stats, sortByCreatedAt: false),
+          replaceInteractionCounts: true,
+        ),
+        PopularVideosVariant.classic,
+      );
+
+      final deduped = <VideoEvent>[];
+      _appendUniqueVideos(deduped, videos, seenVideoKeys: <String>{});
+      deduped.shuffle(_random);
+
+      // A short page means the archive ran out at this offset. Length is read
+      // from the raw stats, not the filtered list, so a page that filters down
+      // to nothing still advances instead of ending the feed.
+      final exhausted = stats.length < limit;
+      final result = HomeFeedResult(
+        videos: deduped,
+        paginationCursor: exhausted
+            ? null
+            : _encodeClassicsOffsetCursor(offset + limit),
+        hasMore: !exhausted,
+      );
+
+      if (isFirstPage) _inMemoryFeedCache?.set(_classicsCacheKey, result);
+      return result;
+    } on FunnelcakeException {
+      return const HomeFeedResult(videos: [], hasMore: false);
+    }
+  }
+
+  /// A random archive offset for a Classics session, aligned to [limit]-sized
+  /// pages so successive cursors stay on page boundaries.
+  int _classicsOffset(int limit) {
+    final pageCount = (_classicsMaxRandomStartOffset ~/ limit) + 1;
+    return _random.nextInt(pageCount) * limit;
   }
 
   /// Fetches new divine videos from the popular leaderboard.
--- a/mobile/packages/videos_repository/lib/src/videos_repository.dart
+++ b/mobile/packages/videos_repository/lib/src/videos_repository.dart
@@ -83,9 +83,17 @@ const String _classicsOffsetCursorPrefix = 'classic-offset:';
 ///
 /// The loops-sorted archive holds roughly 10k videos; starting anywhere in
 /// the most-looped few hundred keeps the feed varied between sessions without
-/// dropping the user into the unwatched long tail. Mirrors the Explore →
+/// dropping the user into the unwatched long tail. Shared with the Explore →
 /// Classics tab so both surfaces draw from the same window.
-const int _classicsMaxRandomStartOffset = 400;
+const int classicsMaxRandomStartOffset = 400;
+
+/// Archive pages one Classics fetch may read while trying to fill its page.
+///
+/// Filtering — blocked authors, content preferences, transport scheme — can
+/// empty an otherwise full archive page, and a feed that opens on an empty
+/// page cannot paginate its way out. Reading a couple of extra pages covers
+/// that without turning a single fetch into a crawl through the archive.
+const int _classicsMaxPageFetches = 3;
 
 /// Result of a hashtag REST feed fetch.
 class HashtagFeedVideosResult {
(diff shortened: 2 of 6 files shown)


mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -777,37 +815,104 @@ class VideosRepository {
...
+    final hasMore = !exhausted && !failed;
+    final result = HomeFeedResult(
+      videos: collected,
+      paginationCursor: hasMore ? _encodeClassicsOffsetCursor(offset) : null,
+      hasMore: hasMore,
     );
+
+    // A failed page is not worth pinning for the rest of the session.
+    if (!failed && isFirstPage) {
+      _inMemoryFeedCache?.set(_classicsCacheKey, result);
```
realmeylisdev, line 919: The cache write is guarded on `!failed && isFirstPage` but not on `collected.isEmpty`, and that combination is newly reachable.

When all three top-up pages come back full-length but filter down to nothing, `exhausted` stays false (line 888 reads the raw page length), so the result is `videos: []` + `hasMore: true` + `paginationCursor: 'classic-offset:<n>'` — and it gets pinned here.

Nothing can consume that cursor. `VideoFeedBloc._onLoadMoreRequested` returns early on `state.videos.isEmpty`, so the feed sits empty with `hasMore: true`. And because a mode switch loads with `skipCache: false` (see my note on the bloc), every later entry into Classics re-serves the pinned empty page without touching the network. Only pull-to-refresh or a process restart clears it.

This state wasn't reachable before: both empty-list exits in `getPopularVideosPage` set `hasMore: false`, so the feed showed a clean end-of-content instead of a stuck one.

A `collected.isNotEmpty &&` in the guard fixes the stickiness. Whether an empty-but-non-exhausted page should also report `hasMore: false` is the bigger call — happy either way, but the cache pin looks unintended.

mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -777,37 +815,104 @@ class VideosRepository {
...
+            ),
+            PopularVideosVariant.classic,
+          ),
+          seenVideoKeys: seenVideoKeys,
+        );
+
+        // A short page means the archive ran out at this offset. Length is
+        // read from the raw stats, not the filtered list, so a page that
+        // filters down to nothing tops up instead of ending the feed.
+        exhausted = stats.length < limit;
```
realmeylisdev: The comment above says the length is "read from the raw stats, not the filtered list" — that holds against the variant/content filter, but `stats` isn't raw either. `FunnelcakeApiClient.getClassicVines` applies `.where((v) => v.id.isNotEmpty && v.videoUrl.isNotEmpty)` before returning (funnelcake_api_client.dart:1434-1437), so a full server page can already arrive short.

That matters because the server truncates to exactly `limit` whenever more rows exist (`collect_visible_offset_page` in divine-funnelcake handlers.rs), so `stats.length < limit` is otherwise an exact exhaustion signal — one dropped row would end the feed mid-archive.

I checked the live endpoint and it isn't dropping rows today, so this is latent rather than live. Worth either narrowing the comment or having `getClassicVines` hand back the pre-filter count so the signal doesn't depend on a filter in another package.

mobile/lib/blocs/video_feed/video_feed_bloc.dart
```diff
@@ -925,16 +925,12 @@ class VideoFeedBloc extends Bloc<VideoFeedEvent, VideoFeedBlocState> {
...
-              until: until,
-              skipCache: skipCache,
-            )
-          : _videosRepository.getClassicVideos(
-              cursor: paginationCursor,
-              skipCache: skipCache,
-            ),
+    // Classics is offset-paginated behind an opaque cursor and has no
+    // time-window pagination, so `until` does not apply.
+    VideoFeedSourceType.classic => _videosRepository.getClassicVideos(
```
realmeylisdev, line 930: The PR description says "the bloc passes `skipCache: true` on mode switch, refresh, and auto-refresh, so picking Classics from the home picker always draws a fresh slice." Only refresh does.

- Mode switch: `_selectSource` calls `_loadVideos(source, emit)` at line 371, and `skipCache` defaults to `false` (line 697). So picking Classics from the picker replays the pinned first page.
- Auto-refresh: `_onAutoRefreshRequested` returns early unless the source is `following` or `forYou` (lines 533-536), so it never runs for Classics at all.

The dartdoc on `getClassicVideos` describes this correctly ("leaving and re-entering the feed preserves the current ordering") — it's the PR description that's off. Not asking for a behaviour change here; the stable-on-re-entry design reads deliberate. But it's worth correcting the description, since it's also what makes the empty-page cache pin sticky rather than one-shot.

mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -777,37 +815,104 @@ class VideosRepository {
...
+  /// order every time. Each fetch that misses the cache starts at a random
+  /// offset within [classicsMaxRandomStartOffset] and shuffles the page, so
+  /// two sessions do not see the same opening run. The next offset rides in
+  /// [HomeFeedResult.paginationCursor], so the home feed paginates this like
+  /// the For You feed. Filtering can empty an archive page, so up to
+  /// [_classicsMaxPageFetches] pages are read to fill [limit] rather than
+  /// handing back a page a feed cannot paginate out of.
+  ///
+  /// The in-memory cache keeps the slice stable while the process lives:
+  /// leaving and re-entering the feed preserves the current ordering, while
```
realmeylisdev: Small scope correction: "leaving and re-entering the feed preserves the current ordering" is true for the first page only. The cache is read and written under `isFirstPage` (lines 843-847, 904-907), and `collected.shuffle(_random)` at line 895 runs on every fetch — so pages 2+ come back as the same videos in a different order on re-entry.

mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -74,6 +76,24 @@ const Duration _recentRelayRefreshTimeout = Duration(seconds: 3);
 const Duration _routeRelayTimeout = Duration(seconds: 3);
 const String _popularCacheKey = 'popular';
 const String _popularLegacyBeforeCursorPrefix = 'before:';
+const String _classicsCacheKey = 'classics';
+const String _classicsOffsetCursorPrefix = 'classic-offset:';
+
+/// Highest archive offset the Classics feed may open a session on.
+///
+/// The loops-sorted archive holds roughly 10k videos; starting anywhere in
+/// the most-looped few hundred keeps the feed varied between sessions without
```
realmeylisdev: "the most-looped few hundred" doesn't describe what offsets 0-400 actually contain on the path this request takes.

`getClassicVines` always sends `platform=vine` and (via `_videoQueryParameters`) `nsfw=show`. In divine-funnelcake, `should_use_classic_snapshot` (crates/clickhouse/src/client.rs:1770-1785) requires `platform.is_none()` and `!include_nsfw`, so the request never reaches the rank-ordered `classic_videos_snapshot` — it falls to the generic filtered path, where `sort=loops` maps to `COALESCE(views, 0) DESC, loops DESC` (client.rs:1463). Views is primary, loops is only the tiebreak.

The routing is pre-existing (Explore → Classics has always gone through the same client method), so nothing to fix in the code here — but since this comment is new, "most-viewed" would be the accurate word. Same for "loops-sorted archive" on line 820.

mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -777,37 +815,104 @@ class VideosRepository {
...
+      }
+    } on FunnelcakeException {
+      failed = true;
+    }
+
+    collected.shuffle(_random);
+
+    final hasMore = !exhausted && !failed;
+    final result = HomeFeedResult(
+      videos: collected,
```
realmeylisdev, line 910: `collected` is never trimmed, and the loop re-checks `collected.length < limit` only between fetches — so a partially-filtered page followed by a full one returns up to `2 * limit - 1` videos (49 at the production default of 25).

Every sibling feed path ends in `.take(limit).toList()` (lines 364, 404, 765, 815, 1019, 1040, 1217), and the method this replaces did too. Nothing is lost or duplicated because the cursor advances past every page actually read, so this is a consistency point rather than a bug — but the page size handed to the bloc now varies, which is a bit surprising for a parameter named `limit`.

mobile/packages/videos_repository/test/src/videos_repository_test.dart
```diff
@@ -4179,125 +4225,286 @@ void main() {
...
-              limit: any(named: 'limit'),
-              cursor: any(named: 'cursor'),
-              before: any(named: 'before'),
-              preferredLanguages: any(named: 'preferredLanguages'),
-              viewerCountry: any(named: 'viewerCountry'),
-            ),
-          ).called(1);
-        },
-      );
+      test('draws a different slice on each cache-missing fetch', () async {
```
realmeylisdev, line 4421: This test passes `skipCache: true` on the second call, but `buildRepository` (line 4281) injects no `InMemoryFeedCache` — so `_inMemoryFeedCache?.get` is a null no-op and the argument does nothing. Dropping `skipCache: true` leaves the test green, which means the "draws a different slice" claim is currently resting on the stubbed offset draws rather than on cache-skipping.

More broadly: `keeps the slice stable across cache-eligible fetches` (line 4438) is the only test in the group with a real cache, and it only exercises `(cursor: null, skipCache: false)`. Dropping `isFirstPage` or `!skipCache` from the read guard, or `!failed` from the write guard, all leave the suite green.

mobile/packages/videos_repository/test/src/videos_repository_test.dart
```diff
@@ -4179,125 +4225,286 @@ void main() {
...
+      test('returns an empty result when Funnelcake is unavailable', () async {
+        when(() => mockFunnelcakeClient.isAvailable).thenReturn(false);
+
+        final result = await buildRepository().getClassicVideos();
+
+        expect(result.videos, isEmpty);
+        expect(result.hasMore, isFalse);
+      });
+
+      test('returns an empty result when the archive request fails', () async {
```
realmeylisdev, line 4602: This stubs `getClassicVines` to throw on every call, so `collected` is already empty when the catch runs and `expect(result.videos, isEmpty)` can't distinguish "partial results retained" from "partial results discarded."

That leaves the PR description's claim — "A request that fails partway serves the pages that already succeeded and is not pinned in the session cache" — unasserted in both halves. Adding `collected.clear();` inside the catch keeps the suite green, and so does dropping `!failed` from the cache guard.

A case where fetch 1 returns rows and fetch 2 throws would pin both behaviours.

mobile/packages/videos_repository/test/src/videos_repository_test.dart
```diff
@@ -4179,125 +4225,286 @@ void main() {
...
+        when(() => mockFunnelcakeClient.isAvailable).thenReturn(true);
         when(
           () => mockFunnelcakeClient.getBulkVideoStats(any()),
         ).thenAnswer((_) async => const BulkVideoStatsResponse(stats: {}));
-        when(
-          () => mockFunnelcakeClient.getV2PopularVideosPage(
-            variant: any(named: 'variant'),
+      });
+
+      test('reads the archive at a page-aligned random start offset', () async {
```
realmeylisdev, line 4310: Every success-path test in this group uses `limit: 20`, but production never does — `video_feed_bloc.dart:930` passes no limit, so `_defaultLimit = 25` applies, giving 17 reachable pages and offsets {0, 25, ..., 400} rather than the 21/20 grid asserted here.

Replacing both `limit` references in `_classicsOffset` with the literal `20` survives the whole group, so the formula's dependence on `limit` isn't pinned. It's also self-reinforcing: `_StubRandom.offsetDrawMax` hardcodes the `limit: 20` page count, so a test at 25 would need the stub generalized first.

mobile/packages/videos_repository/test/src/videos_repository_test.dart
```diff
@@ -16,6 +17,51 @@ class MockFunnelcakeApiClient extends Mock implements FunnelcakeApiClient {}
...
+
+  /// Every `max` passed to [nextInt], in call order.
+  final List<int> recordedMaxima = [];
+
+  int _offsetDrawCount = 0;
+
+  @override
+  int nextInt(int max) {
+    recordedMaxima.add(max);
+    if (max != offsetDrawMax) return _shuffleRandom.nextInt(max);
```
realmeylisdev: The `max == offsetDrawMax` discriminator isn't unique to the offset draw. `List.shuffle` walks `nextInt(len)`, `nextInt(len - 1)`, ... `nextInt(2)`, so any `collected` of 21+ items emits a `nextInt(21)` that lands in the scripted branch — either consuming an offset draw or throwing `Unscripted Classics offset draw`.

Not reachable in the current tests (collected tops out at 20), but the top-up loop can leave `collected` above `limit`, so the first test that exercises a partially-filtered page followed by a full one will trip it — and it'll look like a production bug rather than a stub artefact. Tracking "has the offset draw already happened" instead of branching on `max` would be sturdier.

realmeylisdev (commented):
Nice change — the randomization itself is right, the cursor plumbing reuses the existing shape cleanly, and I verified 451 tests / 100% line coverage on the package.

One thing I'd like fixed before merge: an all-filtered first page gets pinned in the session cache as `videos: []` + `hasMore: true`, which the feed can't paginate out of and can't re-fetch. The rest are doc/description corrections and test gaps (several confirmed by mutation — the mutants stay green). Details inline.

hm21:
Thanks — all ten checked out against the code, and all ten are addressed in 4034b73. One collected reply instead of ten inline ones.

**Blocking: the pinned empty page.** Fixed. The cache write is now gated on `collected.isNotEmpty` as well, so an all-filtered slice is redrawn on the next entry instead of being served from cache for the rest of the process. Your reading of the stickiness was right: `_selectSource` → `_loadVideos(source, emit)` takes the `skipCache: false` default, so it was every later entry, not just one.

**Exhaustion signal.** Took the sturdier option rather than narrowing the comment: `exhausted` is now `stats.isEmpty`, so the signal no longer depends on `getClassicVines`' `id`/`videoUrl` filter in another package. A short page just tops up; at the true end of the archive the next read comes back empty and ends the feed. Costs one extra request at the archive end, which beats ending the feed mid-archive if a row is ever dropped. Pinned by `keeps paginating when a short page is not the archive end`.

**`most-looped` → `most-viewed`.** Corrected on both the constant and the method doc, with a note that `sort=loops` maps to `COALESCE(views, 0) DESC, loops DESC` on this path because the request carries `platform=vine`. Verified against `should_use_classic_snapshot` and the sort mapping in divine-funnelcake.

**Cache-scope dartdoc.** Corrected — first page only, and pages 2+ are re-drawn and re-shuffled on re-entry.

**PR description.** Corrected: only pull-to-refresh passes `skipCache: true`; mode switch resumes the pinned page deliberately, and auto-refresh never applies to Classics. Also fixed the test count (461 → 456) and the "loops-sorted" wording.

**`.take(limit)` — keeping the current behaviour, one disagreement.** The siblings can trim because their cursor is derived from what they return. Here the offset cursor has already advanced past every page read, so `.take(limit)` would drop those videos rather than defer them to the next page. Documented on the method now, and `keeps paginating when a short page is not the archive end` asserts the 39-item result so the choice is pinned rather than incidental.

**Test gaps.** All four were real; closed and mutation-checked.

- `buildRepository` now injects a live `InMemoryFeedCache`, so `skipCache: true` in `draws a different slice` is load-bearing and `advances the cursor by a page and resumes from it` pins `isFirstPage` in the read guard.
- New `does not let a later page overwrite the cached first page` pins `isFirstPage` in the write guard.
- New `serves the pages that succeeded when a top-up fails`: page 1 returns 12 visible rows, the top-up throws — asserts the 12 survive and that the partial page is not pinned. Pins `!failed`.
- New `does not pin an empty page for the rest of the session` pins `collected.isNotEmpty`.
- New `aligns the offset grid to the page size the feed asks for` runs at the production default: 17 pages, offset 175, `limit: 25`.
- `_StubRandom` no longer branches on `max`. The offset draw is identified by position (`expectOffsetDraw`, armed before the fetch that draws it) and throws if a scripted draw is never consumed. That was load-bearing rather than theoretical — the new 39-item shuffle would have tripped the old `max == offsetDrawMax` branch exactly as you predicted.

Mutants killed, one test each: dropping `collected.isNotEmpty`, `!failed` or `isFirstPage` from the write guard; dropping `!skipCache` or `isFirstPage` from the read guard; reverting exhaustion to `stats.length < limit`; adding `.take(limit)`.

456 tests / 100% line coverage held on the package, 271 on the app-layer feed suites, `flutter analyze` clean. Leaving the threads for you to resolve.

mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -74,6 +76,27 @@ const Duration _recentRelayRefreshTimeout = Duration(seconds: 3);
 const Duration _routeRelayTimeout = Duration(seconds: 3);
 const String _popularCacheKey = 'popular';
 const String _popularLegacyBeforeCursorPrefix = 'before:';
+const String _classicsCacheKey = 'classics';
+const String _classicsOffsetCursorPrefix = 'classic-offset:';
+
+/// Highest archive offset the Classics feed may open a session on.
+///
+/// The archive holds roughly 10k videos; starting anywhere in the most-viewed
```
realmeylisdev, line 84: The archive is a lot bigger than this. `/api/stats` reports `vine_videos: 2158661`, and the endpoint still serves distinct `platform: vine` rows at `offset=2000000` — so ~2.2M, not 10k.

Doesn't change the constant: a 400-row window out of 2.2M makes the "don't drop the user into the unwatched long tail" reasoning stronger, not weaker. The 10k looks carried over from `_totalClassicVines` in `classic_vines_provider.dart`, which is stale for the same reason.

realmeylisdev (approved):
Re-checked all ten against 4034b73 — all hold, and I ran the mutants myself rather than trust the list: removing `collected.isNotEmpty`, `!failed` or `isFirstPage` from the write guard, `!skipCache` or `isFirstPage` from the read guard, reverting exhaustion to `stats.length < limit`, or adding `.take(limit)` each turns exactly one test red.

The empty-page pin is gone, and reading exhaustion as "the archive returned nothing here" is the right call — `collect_visible_offset_page` truncates to exactly `limit` whenever more rows exist, so the only short pages are the archive end and rows the client dropped. Approving; one doc nit inline, nothing blocking.

PR #6543, opened 2026-07-31
fix(media-cache): suspend downloads on a background launch

### LINKED ISSUE #6544: fix(media-cache): suspend downloads on a background launch
Auto-created from PR #6543.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6543
**Branch:** `fix/media-cache-background-launch`

hm21:
Closes #6544

## Description

On Apple platforms a live NSURLSession callback trampolining into a suspended isolate aborts the process inside the `cupertino_http` FFI layer. It is the second-biggest FATAL issue in Crashlytics: 25 events across 23 users.

#6222 already tears in-flight downloads down on the `paused` transition. The crash survived it, because the observer only reacts to *transitions*: a process launched straight into the background — silent push, background upload completion — never receives `paused`, since it was never `resumed`. `didChangeAppLifecycleState` therefore never fires, the teardown never runs, and the downloads keep going until iOS suspends the process mid-request.

The sampled events back that up rather than the cancellation theory I started with:

- All four are `processState: BACKGROUND`.
- Two are on build 813, which already contains #6222.
- The blamed callbacks are `dataTask:didReceiveData:` and `dataTask:didReceiveResponse:` — active streams, not the `didCompleteWithError` a cancellation would produce. One is inside `CFHostCreateCopy`, i.e. still resolving DNS.
- Their logs show a full app init seconds before the crash, at 02:16 and 04:46 local time.

So the fix reads the lifecycle state the process *starts* in, which is what a transition cannot cover, and latches the same suspension #6222 already implements.

A `null` reading is deliberately treated as foreground. It is the normal case early in a foreground launch, and latching there would risk leaving downloads suspended for good on a launch that never sends a `resumed` transition to lift them — a worse failure than the crash being fixed.

**Related Issue:** 

## Out of Scope

Only the launch-time gap. The transition handling from #6222 is untouched, and no `cupertino_http` or `media_cache` internals change — `cancelInFlightDownloads()` is the existing entry point.

The remaining Crashlytics finding from the same pass, `SqliteException(26): file is not a database` (12 users), is a separate root cause and gets its own PR.

## Verification

- Three tests on the extracted predicate, covering the background states, `resumed`, and the `null` reading.
- `flutter analyze lib test integration_test` — clean once the upstream `main` breakage is patched locally; the only remaining errors are the two inherited ones quoted above.

The crash needs a real background launch followed by an OS suspension, so it is not reproducible by hand in a debug session. Field confirmation is the `hashmap.cc` / `DLRT_GetFfiCallbackMetadata` signature dropping after release.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -1760,6 +1760,26 @@ class DivineApp extends ConsumerStatefulWidget {
   ConsumerState<DivineApp> createState() => _DivineAppState();
 }
 
+/// Whether a launch that observes [state] as its first lifecycle reading must
+/// start with media downloads suspended.
+///
+/// #6222 tears in-flight downloads down on the `paused` transition, because on
+/// Apple platforms a live NSURLSession callback trampolining into a suspended
+/// isolate aborts the process in the `cupertino_http` FFI layer. A process
+/// launched straight into the background — silent push, background upload
+/// completion — never gets that transition, because it was never resumed. Its
+/// downloads therefore ran until iOS suspended it mid-request, which is the
+/// remaining crash. Reading the state we start in covers what a transition
+/// cannot.
+///
+/// `null` means the engine has not reported a state yet, which is the normal
+/// case on a foreground launch. Suspending there would risk latching downloads
+/// off with no `resumed` transition arriving to lift them again, so it is
+/// treated as "not background".
+@visibleForTesting
+bool shouldSuspendDownloadsAtLaunch(AppLifecycleState? state) =>
+    state != null && state != AppLifecycleState.resumed;
+
 class _DivineAppState extends ConsumerState<DivineApp>
     with WidgetsBindingObserver {
   bool _backgroundInitDone = false;
@@ -1774,6 +1794,14 @@ class _DivineAppState extends ConsumerState<DivineApp>
   void initState() {
     super.initState();
     WidgetsBinding.instance.addObserver(this);
+    // Transitions alone do not cover a process launched straight into the
+    // background, so latch the suspension from the state we start in.
+    if (shouldSuspendDownloadsAtLaunch(
+      WidgetsBinding.instance.lifecycleState,
+    )) {
+      openVineMediaCache.cancelInFlightDownloads();
+      openVineImageCache.cancelInFlightDownloads();
+    }
     _memoryPressureHandler = MemoryPressureHandler(
       clearImageCache: () {
         PaintingBinding.instance.imageCache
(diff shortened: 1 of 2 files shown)


mobile/lib/main.dart
```diff
@@ -1760,6 +1760,26 @@ class DivineApp extends ConsumerStatefulWidget {
...
+/// launched straight into the background — silent push, background upload
+/// completion — never gets that transition, because it was never resumed. Its
+/// downloads therefore ran until iOS suspended it mid-request, which is the
+/// remaining crash. Reading the state we start in covers what a transition
+/// cannot.
+///
+/// `null` means the engine has not reported a state yet, which is the normal
+/// case on a foreground launch. Suspending there would risk latching downloads
+/// off with no `resumed` transition arriving to lift them again, so it is
+/// treated as "not background".
```
realmeylisdev, line 1778: The choice is right, but the reason doesn't hold on the two platforms this fix targets: the first reading is never `null` there.

Both mobile embedders seed the state, and the framework consumes it before `runApp`:

- iOS — `FlutterDartProject.mm`: `PlatformData.lifecycle_state = "AppLifecycleState.detached"`
- Android — `android_shell_holder.cc`: `platform_data.lifecycle_state = "AppLifecycleState.detached"`
- `ServicesBinding.initInstances()` → `readInitialLifecycleStateFromNativeWindow()` → `_handleLifecycleMessage(...)`

So a foreground launch reads `detached` — or `resumed`/`inactive` once the engine has delivered a transition — never `null`. `null` is only reachable on web (`web_ui`'s `initialLifecycleState` returns `''`) and under `flutter_test` (`TestPlatformDispatcher` defaults it to `''`).

Two consequences worth capturing in the comment:

1. A launch that reaches `initState` before the engine's `resumed` message starts suspended, foreground included. That self-heals — `addObserver` runs first, and `_generateStateTransitions(detached, resumed)` emits exactly `resumed`, so `resumeDownloads()` fires.
2. `null → false` is still the right call, but for a different reason than the one given: it keeps web and any future widget test that pumps `DivineApp` from latching the global `openVineMediaCache` / `openVineImageCache` singletons.

As written, a later reader can reasonably conclude `detached` is unreachable at launch and drop it, which would silently undo this fix. Same wording is echoed in `test/main_background_launch_downloads_test.dart:36`.

realmeylisdev (approved):
Approving. Traced the mechanism end to end and it holds: `cupertino_http` is confined to `media_cache` (the legacy `flutter_cache_manager` path uses a plain `dart:io` client), nothing reaches the network through either cache before `runApp` — the critical startup phase is EnvironmentService / HiveStorage / CacheSync, and seed preload is deferred and asset-only — `DivineApp.initState` does run on a background launch because `scheduleWarmUpFrame` bypasses the disabled-frames state, and `addObserver` precedes the read, so a later `resumed` still lifts the latch. `flutter analyze lib test integration_test` and the root `main_*` tests are green locally.

One inline note on the doc comment's `null` rationale. Wording only; the behaviour is right.

PR #5203, opened 2026-06-15
fix(library): prevent draft posts from dropping editor layers

hm21:
Closes #5204

## Description

The draft options sheet offered Post for drafts even when posting could publish a source clip instead of the edited render. For single-clip drafts with no finalRenderedClip, video_publish_provider.dart falls back to draft.clips.first, which can silently drop editor layers such as text, stickers, filters, and other baked edits.

This keeps the Library shortcut aligned with the actual publish boundary:

- Single-clip drafts need a finalRenderedClip before Library shows Post.
- Multi-clip drafts can still show Post without a saved final render because the publish path renders them before upload and preserves editor state.
- Draft loading clears stale finalRenderedClip references when the rendered file is missing, so a broken render reference cannot make an unsafe single-clip draft appear postable.

Edit and Delete remain available for all drafts.

## Verification

- mise exec -- flutter test test/models/divine_video_draft_can_post_test.dart test/widgets/library/drafts_tab_test.dart test/services/draft_storage_service_test.dart
- mise exec -- flutter analyze lib test
- Pre-push hook: merge conflict check, changed-file analyze, changed-file tests

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

Issue #5181, opened 2026-06-15
fix(video-editor): saved video post loses overlay and thumbnail settings


hm21:
### Summary

Saving a video post for later does not preserve text overlay formatting or the selected thumbnail.

### Environment

any

### Steps to Reproduce

1. Create a video post.
2. Add a text overlay and customize the font style and size.
3. Select a thumbnail from the trimmed video.
4. Save the post for later.
5. Reopen the saved post and publish it.

### Actual Result

Text overlays revert to the default font style and size, and the thumbnail resets to the first frame of the original clip, including frames that were trimmed off.

### Expected Result

Text overlays should retain their configured font style and size, and the selected thumbnail should remain unchanged when a saved post is reopened and published.

### Regression?

No

### Evidence

reported on discord [here](https://discord.com/channels/1470168817589158040/1470255950408454164/1515931590218485812)

PR #6549, opened 2026-07-31
fix(library): render draft overlays instead of hiding the post action

### LINKED ISSUE #6550: fix(library): render draft overlays instead of hiding the post action
Auto-created from PR #6549.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6549
**Branch:** `feat/draft-library-post-render`

hm21:
Closes #6550

## Description

Posting a draft from the library shipped a video without the overlays the user authored.

`CompleteParameters.toMap` — how a draft persists its editing parameters — does not serialize `capturedLayers`, `filterStates`, `tuneAdjustments` or `audioTracks`. Parameters restored from storage therefore describe far less than the editor session did: no text, stickers or drawings, no timed filters or tune, no added music. On top of that, a single-clip draft skipped the render entirely and uploaded the raw recording, losing trim, speed and volume too.

#5203 worked around this by hiding the Post action until a render happened to be cached. That left the multi-clip path — which the gate allowed through — still shipping broken videos.

Publish now always renders, and rebuilds what storage dropped before it does. Layers, timed filters and tune come back from `editorStateHistory`; audio comes from the parameters' own `meta` plus the draft's own recorder-picked sound. Only the *rasterized* layer bytes are unrecoverable, because `Layer.captureAsPng` returns `null` for a layer that was never mounted — so they are re-baked offscreen via `LayerRasterizer`, added in pro_image_editor 13.3.0 for exactly this case and hosted once above every route.

Three further inputs the render needs do not survive in the parameters either, and each one silently changed what got published:

- **The editor body size.** It only reaches storage through the editor's complete callback, which fires on Done — a draft saved by backing out keeps its layers but no parameters at all. It now falls back to the size the exported state history was laid out against, and that size is handed back on the returned parameters so the capture and the render share one reference frame.
- **The overlay fonts.** A text layer persists only its font family, and Google Fonts register lazily per process, so a publish after a cold start baked every non-default preset in the platform default — wrong typeface, and wrong position through the changed metrics (#5181). The editor screen already guarded this before importing a history; that preload moved into a shared helper the headless rasterizer calls too.
- **The recorder-picked sound.** It lives on the draft and never becomes a timeline track, so the same draft published with music from the editor and silent from the library.

### Why some non-obvious calls were made

**Blocking rather than degrading.** When the overlays cannot be reproduced (unreadable state history, no body size from either source, a failed rasterization, or a capture that comes back short) the publish fails with an actionable message instead of posting a video without them. Silently dropping overlays is the original bug, and a published Nostr event cannot be quietly corrected — a retry the user can act on beats a wrong post they cannot take back. A single sticker image that fails to load is treated differently: it costs that sticker and is logged.

**Always rendering, even for one clip.** The old pass-through was faster but only correct for a completely unedited recording, and getting that condition right means enumerating every field that affects output. Rendering unconditionally is both simpler and strictly more faithful — it is what fixes the trim/speed/volume loss.

**Package defaults for the rasterizer configs.** Only a few config values change how a layer *renders* (`textEditor.initFontSize`, `textEditor.style.leadingDistribution`, `emojiEditor.style.textStyle`, `stickerEditor.initWidth`, `paintEditor.censorConfigs`) and the editor canvas overrides none of them. Both sides reading the same defaults is what keeps a re-render identical to what the user saw, so the argument is deliberately omitted rather than pinned to a copy.

**Warming sticker images before capture.** A sticker paints nothing until its image resolves, and a network sticker additionally fades in from fully transparent — capturing before either finishes yields a blank layer. Priming the cache makes the mounted widget resolve on its first build, which also skips the fade. The font registration rides along on the same hook, so both land in the first build the rasterizer measures.

**One helper per shared decision.** The audio list and the font preload are now single functions the in-editor export and the headless draft render both call, rather than parallel implementations. Both had already drifted once — that is how the recorder sound went missing from one path and not the other.

**Checking the capture length.** `Layer.captureAllLayers` drops a layer whose repaint boundary produced no image rather than reporting it, so a short result is a silent partial bake. Comparing the count is what makes that reach the same block as an outright failure.

## Out of Scope

- Removing `DivineVideoDraft.canPost` entirely. It is now just `clips.isNotEmpty`, which still carries meaning at the call site.
- The remaining `PublishErrorKind.generic` message on a plain render failure. A frames-only draft gets the stop-motion assembly copy; everything else stays generic.

## Verification

- `flutter test` over the affected suites (draft render params, publish provider, editor provider, drafts tab, draft storage, models, l10n) — 515 passing
- `flutter test test/l10n/arb_consistency_test.dart` — the new key is translated in all 18 locales, no English fallbacks
- `check_untested_services_floor.sh`, `check_test_unit_structure.sh` and all four design-system ceilings — OK
- Pre-push hook (analyze + codegen + changed-file tests) — 243 passing
- Mutation-checked all seven new and reworked tests: removing each production change fails exactly its own test, including the body-size guard, whose earlier test could not fail because an unmounted rasterizer threw either way
- Manually verified on a real Samsung Galaxy: the layers never flash visible while a publish rasterizes them, and stickers land fully in the exported video

## Type of Change

### DIFF
--- a/mobile/lib/services/video_editor/draft_render_parameters_service.dart
+++ b/mobile/lib/services/video_editor/draft_render_parameters_service.dart
@@ -14,25 +14,12 @@ import 'package:openvine/extensions/aspect_ratio_extensions.dart';
 import 'package:openvine/extensions/complete_parameters_extensions.dart';
 import 'package:openvine/models/divine_video_draft.dart';
 import 'package:openvine/services/video_editor/video_editor_audio_render.dart';
+import 'package:openvine/utils/editor_text_fonts.dart';
 import 'package:openvine/utils/open_vine_image_cache.dart';
 import 'package:openvine/widgets/video_editor/sticker_editor/video_editor_sticker.dart';
 import 'package:pro_image_editor/pro_image_editor.dart';
 import 'package:unified_logger/unified_logger.dart';
 
-/// Rebuilds the [CompleteParameters] needed to render a draft when no editor
-/// session is open — posting straight from the library, most importantly.
-///
-/// A draft persists its editing parameters via `CompleteParameters.toMap`,
-/// which drops `capturedLayers`, `filterStates`, `tuneAdjustments` and
-/// `audioTracks`. Rendering from the restored map alone therefore produces a
-/// video with none of the overlays the user authored: no text, stickers or
-/// drawings, no timed filters or tune adjustments, and no added music.
-///
-/// Everything needed to rebuild them survives elsewhere in the draft —
-/// `editorStateHistory` keeps the layers, filters and tune, and the audio
-/// timeline lives in the parameters' own `meta`. Only the *rasterized* layer
-/// bytes are unrecoverable from storage, which is what [LayerRasterizer]
-/// regenerates by briefly mounting the layers offscreen.
 /// Thrown when a draft's authored overlays cannot be reproduced for a render.
 ///
 /// Publishing is blocked rather than degraded. A video that silently lacks the
@@ -50,6 +37,21 @@ class DraftOverlayRestoreException implements Exception {
   String toString() => 'DraftOverlayRestoreException: $message';
 }
 
+/// Rebuilds the [CompleteParameters] needed to render a draft when no editor
+/// session is open — posting straight from the library, most importantly.
+///
+/// A draft persists its editing parameters via `CompleteParameters.toMap`,
+/// which drops `capturedLayers`, `filterStates`, `tuneAdjustments` and
+/// `audioTracks`. Rendering from the restored map alone therefore produces a
+/// video with none of the overlays the user authored: no text, stickers or
+/// drawings, no timed filters or tune adjustments, and no added music.
+///
+/// Everything needed to rebuild them survives elsewhere in the draft —
+/// `editorStateHistory` keeps the layers, filters and tune, the audio timeline
+/// lives in the parameters' own `meta`, and a recorder-picked sound sits on the
+/// draft itself. Only the *rasterized* layer bytes are unrecoverable from
+/// storage, which is what [LayerRasterizer] regenerates by briefly mounting the
+/// layers offscreen.
 class DraftRenderParametersService {
   /// Creates a service that rasterizes through [rasterizer].
   const DraftRenderParametersService({required LayerRasterizer rasterizer})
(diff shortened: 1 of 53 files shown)


mobile/lib/services/video_editor/draft_render_parameters_service.dart
```diff
@@ -0,0 +1,364 @@
...
+    _precacheStickers(stickers),
+  ]);
+
+  /// Loads every sticker's image into the cache the sticker widget reads from.
+  ///
+  /// Sticker layers paint nothing until their image resolves, and network
+  /// stickers additionally fade in from fully transparent — capturing before
+  /// either finishes yields a blank layer. Warming the cache first makes the
+  /// mounted widget resolve on its first build, which also skips the fade
+  /// (`AnimatedOpacity` does not animate into its initial value).
```
realmeylisdev, line 296: This holds only when the image is already in `PaintingBinding.imageCache` before the layer mounts — and here it isn't. `LayerRasterizer.capture` mounts the layers and awaits two frames *before* it calls `awaitContentReady`, so on a cold cache `VineCachedImage` has already built with `_hasImage == false`. When the precache then resolves, the image `AnimatedOpacity` starts 0→1 over `fadeInDuration` (500 ms) and the spinner placeholder 1→0 over `fadeOutDuration` (1000 ms), and the capture lands two frames later — a near-transparent sticker under a nearly-opaque `CircularProgressIndicator`. `captured.length == layers.length`, so the short-capture guard reports nothing.

Not live today: all 71 entries in `assets/stickers/stickers.json` are `assetPath` SVGs with no `networkUrl`, and `SvgPicture` has no fade — the SVG key does hit (`rootBundle` + `const SvgTheme()` resolve on both sides, no `DefaultAssetBundle`/`DefaultSvgTheme` anywhere in the tree), so that path is genuinely warm by the capture frame. So this is latent rather than broken. But the first `networkUrl` sticker added to the manifest turns a cold-start publish into exactly the silent wrong post this service exists to block, and the comment reads as though it were already handled. Either suppressing the fade on the rasterizer path or resolving the images before the capture request is issued would close it — which one is your call.

mobile/lib/services/video_editor/draft_render_parameters_service.dart
```diff
@@ -0,0 +1,364 @@
...
+
+    try {
+      // `configs` is deliberately left at the package default. Only a few
+      // values change how a layer *renders* rather than how it is edited —
+      // `textEditor.initFontSize` (which scales text and emoji layers),
+      // `textEditor.style.leadingDistribution`, `emojiEditor.style.textStyle`,
+      // `stickerEditor.initWidth` and `paintEditor.censorConfigs` — and the
+      // editor canvas overrides none of them. Both sides reading the same
+      // defaults is what keeps a re-render identical to what the user saw; if
+      // the canvas ever overrides one, mirror it here.
```
realmeylisdev, line 239: `LayerWidget` reads two more config values on the render path that aren't in this list: `configs.{text,emoji,sticker,paint}Editor.layerFractionalOffset` (`layer_widget.dart:142-156` — it positions the layer) and `configs.videoEditor.layerTimeline` (`:353`). Both are package defaults on the canvas too, so the conclusion is right — but this comment is the checklist someone will re-verify against when the canvas grows a new override, so it's worth naming them. Same for `configs.theme`, which the package's own `capture` doc singles out: the canvas never sets it, so both sides land on `defaultEditorTheme()`.

realmeylisdev (approved):
Checked the load-bearing claims against pro_image_editor 13.3.0 and the canvas configs: `toMap` does drop the four fields, omitting `configs` is safe (the canvas overrides none of the layer-render values, `theme` included), and the `lastRenderedImgSize` fallback matches both the package's own example and the scale `buildImageLayers` applies.

Locally: `flutter analyze` clean, the five touched suites green, both floors and all four design-system ceilings OK.

Two non-blocking notes inline.

PR #6551, opened 2026-07-31
fix(upload): stop duplicating the device attestation into its own header

### LINKED ISSUE #6552: fix(upload): stop duplicating the device attestation into its own header
Auto-created from PR #6551.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6551
**Branch:** `fix/proofmode-drop-duplicate-attestation-header`

realmeylisdev:
Closes #6552

## What

`_addProofModeHeaders` base64-encodes the whole ProofMode manifest into
`X-ProofMode-Manifest`, then pulls `deviceAttestation` back out of that same
JSON and sends it a second time as a standalone `X-ProofMode-Attestation`
header. This drops the duplicate.

## Why

On handsets whose KeyMint chain dumps large — 53,826 B measured on a Pixel
8 Pro — the duplicate took total request headers to roughly 147 KB, past the
128 KiB HTTP/1.1 request-header limit at the media.divine.video edge. The edge
answers 502 before any handler runs, which is why the tiny
`POST /upload/{id}/complete` failed while the multi-megabyte chunk `PUT` (which
carries no ProofMode headers) succeeded. Every publish on an affected device
died.

Dropping the duplicate takes the combined headers to roughly 75 KB, under the
limit, with no data loss — the attestation still ships, once, inside the
manifest header a consumer already has to decode.

## Relationship to #6507

Same root cause, different remedy, and the two are alternatives rather than
dependencies.

#6507 keeps both headers and adds a byte budget, which means the attestation is
silently dropped once it exceeds the budget. This change removes the
redundancy instead, so nothing is ever dropped and no budget is needed. If
#6507 lands first this becomes a simplification on top of it; if this lands
first, #6507's budget may be unnecessary.

Worth noting for whichever lands: no server currently reads any `X-ProofMode-*`
header. `handle_upload_service_proxy` and `handle_upload_complete` in
divine-blossom (`src/main.rs:3782-3803` and `:3608-3643`) build a fresh request
carrying only Host, Authorization, Content-Type and Content-Length, so all four
headers are dropped at the Fastly Compute edge before any backend sees them.
`git log --all -S"X-ProofMode"` over divine-blossom's `src/`, `blossom-core/`,
`cloud-run-upload/`, `cloud-functions/`, `cloud-run-transcoder/` and `vcl/`
returns no commits. The attestation that does reach the backend arrives via
Nostr event tags, not HTTP headers. That is context for reviewers, not a
justification for this change — the change is safe regardless of whether a
consumer appears later, because the data is still on the wire.

## Testing

- Rewrote the three assertions in
  `mobile/packages/blossom_upload_service/test/src/blossom_upload_proofmode_test.dart`
  that pinned the standalone header. The attestation test now asserts the
  header is absent **and** that the value is still recoverable by decoding
  `X-ProofMode-Manifest`, so the no-data-loss property is what is guarded.
- Removed `omits attestation header when deviceAttestation is null`, which
  became a tautology once the header is never sent.
- Added `a Pixel-8-Pro-sized attestation stays under the edge header limit`: it
  builds a 53,826 B attestation and asserts total request-header bytes stay
  under 128 KiB. Verified this fails without the production change
  (`is not a value less than <131072>`) and passes with it.

`flutter analyze lib test` clean; all 227 tests in `blossom_upload_service`
pass.

## Not covered here

This fixes the header size. It does not address why an affected device can
publish some captures and not others — the attestation is a per-file artifact
(`<proofDir>/<fileHash>.attest`, written once at first proof and gated on the
device being online at that moment), so clips on one handset can legitimately
carry 0 B or 53,826 B. Details in #6529.

Covers #6529.

### DIFF
--- a/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
+++ b/docs/protocol/blossom/2026-03-26-divine-resumable-upload-sessions-bud.md
@@ -108,13 +108,19 @@ Optional request headers:
 
 - `X-ProofMode-Manifest`
 - `X-ProofMode-Signature`
-- `X-ProofMode-Attestation`
 - `X-ProofMode-C2PA`
 
 When present, these headers carry the same ProofMode metadata that legacy
 `PUT /upload` accepts. Clients should send them on `complete` so resumable
-uploads preserve the same server-side ProofMode handling without forcing the
-video bytes back onto the legacy single-request upload path.
+uploads preserve the same ProofMode metadata without forcing the video bytes
+back onto the legacy single-request upload path.
+
+There is deliberately no `X-ProofMode-Attestation` header. The device
+attestation is already carried inside `X-ProofMode-Manifest`, and sending it
+a second time took total request headers past the 128 KiB HTTP/1.1 limit at
+the edge on handsets with a large KeyMint chain — the edge then answered 502
+before any handler ran (#6529). Keep the combined header set well under
+128 KiB; the ceiling is enforced at the edge, not by any handler.
 
 Example successful response:
 
--- a/mobile/docs/PROOFMODE_ARCHITECTURE.md
+++ b/mobile/docs/PROOFMODE_ARCHITECTURE.md
@@ -346,10 +346,11 @@ class BlossomUploadService {
         headers['X-ProofMode-Signature'] = manifest['pgpSignature']['signature'];
       }
 
-      // Add attestation header
-      if (manifest['deviceAttestation'] != null) {
-        headers['X-ProofMode-Attestation'] = manifest['deviceAttestation']['token'];
-      }
+      // The device attestation is NOT sent as its own header. It already
+      // ships inside X-ProofMode-Manifest above, and duplicating it took
+      // total request headers past the 128 KiB edge limit on handsets with
+      // a large KeyMint chain, which returned 502 before any handler ran
+      // (#6529).
     }
 
     // Upload video with headers
--- a/mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
+++ b/mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
@@ -2908,9 +2908,15 @@ class BlossomUploadService {
 
   /// Add ProofMode headers to upload request.
   ///
-  /// Generates X-ProofMode-Manifest, X-ProofMode-Signature,
-  /// and X-ProofMode-Attestation headers from the provided
-  /// ProofManifest JSON.
+  /// Generates X-ProofMode-Manifest, X-ProofMode-Signature and
+  /// X-ProofMode-C2PA headers from the provided ProofManifest JSON.
+  ///
+  /// The device attestation is carried inside X-ProofMode-Manifest and is
+  /// deliberately not repeated in a standalone X-ProofMode-Attestation
+  /// header: on handsets whose KeyMint chain dumps large (~53 KB on a Pixel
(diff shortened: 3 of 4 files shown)


hm21:
Going with the byte budget in #6507 instead. Dropping the duplicate halves the payload but leaves it unbounded: the failure threshold moves from ~47.5 KB of raw attestation to ~92 KB rather than going away, and the surviving manifest header is still ~77 KB — over the 64 KiB HTTP/2 limit and over nginx's default 8 KB-per-line budget, which matters because users can point at their own Blossom server. The `TODO(n8fr8)` in `native_proofmode_service.dart` also plans to fold per-clip attestations into the combined manifest, which would blow an unbudgeted manifest header on any multi-clip post.

Your edge-strip finding and the per-file attestation point were the useful parts, and both are folded into #6507 (c70ab6723). I'll close this as superseded once that lands.

Separately, on the #6161 cross-reference — that looks like a different failure. The reporter says "cannot connect to the servers" when posting, but the sheet that says that *before* any upload starts is the C2PA prompt on `video_metadata_screen`: "Post without the human-made check?" with the note "Content credentials need an internet connection." C2PA signing is a network call to `SIGNING_SERVER_ENDPOINT` and runs during the render, before a byte reaches media.divine.video — so an edge 502 on the upload cannot produce that sheet. The upload-failure sheet is a separate surface with separate copy.

I couldn't open the Zendesk attachment to confirm which of the two the reporter screenshotted, so this isn't settled either way — worth checking ticket 4036 before #6161 gets re-routed as fixed-by-6507, because if it is the C2PA prompt then this PR and #6507 both leave it untouched. Also unexplained by either diagnosis: the same user cannot post comments, and a Kind 1111 publish involves no C2PA signing and no Blossom upload at all.

realmeylisdev:
Agreed, and I'm not pushing this further. The 8 KiB cap doesn't just outbound the dedup, it subsumes it: with the cheapest-first ordering the manifest is the last candidate, and at ~77 KB it never fits whether or not the attestation header is also in the list — so removing the duplicate changes no outcome under the budget. The nginx per-line point covers the user-supplied Blossom server case I hadn't accounted for. Leaving this open for you to close once #6507 lands.

On #6161 — it isn't the C2PA prompt either. `videoMetadataC2paMissingTitle` merged in 48c84d304 on 2026-07-18 and first shipped in 1.0.17 (tagged 07-22). The screenshot is dated 07-17 and the ticket was filed 07-18, so the reporter was on ≤1.0.16, which has no such sheet.

Your Kind 1111 point does kill my cross-reference though. Liz made the same write-path observation in her 07-20 triage, and an account-scoped server-side write rejection is the only reading that covers both symptoms. The copy fits it too: a 502 renders `publishErrorServerDown` ("The media server (media.divine.video) is temporarily down"), not "cannot connect to the servers" — `publishErrorServerUnreachable` ("Could not reach the server. Please try again in a moment.") is the bucket she mapped it to, and it was in their build. So #6161 stays where she put it; I've corrected my comment there.

The attachment redirects to a Zendesk sign-in for me too.

realmeylisdev:
Closing this now rather than making you do it — superseded by #6507, which is green on every check and only waiting on a review. No reason to leave a second PR open against the same root cause.

Nothing here needs porting: your `keeps every proof header inside the wire budget` test runs the same ~54 KB attestation scenario and bounds it at 8 KiB, 16× tighter than the 128 KiB assertion I had. The branch stays on the remote if this ever needs reopening.

Issue #6548, opened 2026-07-31
task(upload): move draft deletion and file cleanup off the publish critical path


hm21:
## Description

`VideoPublishService.publishVideo` deletes the draft and reclaims its files *before* it returns `PublishSuccess`. The video is already published at that point, so the user is waiting on garbage collection. Move that cleanup behind the success result.

## Context

Phase timing added in `feat/publish-phase-timing` measured six publishes on a real device. The unattributed `other` bucket — everything in `publishVideo` outside a measured phase — grew steadily across the session:

| Run | `other` |
|---|---|
| 1 | 631 ms |
| 2 | 987 ms |
| 3 | 1015 ms |
| 4 | 1144 ms |
| 5 | 1150 ms |

The growth is the tell. The dominant cost is `draftService.deleteDraft`, which calls `_referencedGhostFrameFilenames` → `_clipsDao.getAllClips(includeTrashed: true)`: it loads **every clip row in the database** and JSON-decodes each `data` blob to find ghost frames other drafts still reference. That is O(all clips) and grows as drafts accumulate. On top of it come several `FileCleanupService` reference scans and the file deletions themselves.

`saveDraft(publishing)` at the start of the publish has a smaller version of the same shape (it loads the previous draft to diff orphaned files).

Note that `BackgroundPublishBloc` already deletes the draft again on success (`_deletePublishedDrafts`), so the service-side deletion is not the only path that reclaims it.

## Implementation Notes

- `mobile/lib/services/video_publish/video_publish_service.dart` — `publishVideo` calls `draftService.deleteDraft(draft.id)` just before returning `PublishSuccess`.
- `mobile/lib/services/draft_storage_service.dart` — `deleteDraft` and `_referencedGhostFrameFilenames`.
- Decide who owns the deletion. If the bloc's `_deletePublishedDrafts` is authoritative, the service call may be redundant rather than merely mistimed.
- Whatever runs after the result must still report failures somewhere — cleanup that silently dies leaves orphaned media files behind.
- Independently of the ordering, the full-table clip scan is worth narrowing: the query could filter to clips carrying a `ghostFramePath` instead of decoding every `data` blob in the database.

## Acceptance Criteria

- [ ] A successful publish returns to the UI without waiting for draft deletion or file cleanup
- [ ] The draft and its unreferenced files are still reclaimed, and a failure to reclaim them is surfaced rather than swallowed
[...]

PR #6554, opened 2026-07-31
perf(upload): ~32% faster publishing by taking the thumbnail off the critical path

### LINKED ISSUE #6555: perf(upload): take the thumbnail, blurhash and video hash off the publish critical path
Auto-created from PR #6554.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6554
**Branch:** `feat/publish-phase-timing`

hm21:
Closes #6555

## Description

Publishing a video felt like it stalled around 75%. It had no timing anywhere — the only duration in the logs was one aggregate upload figure — so the stall could not be attributed to a phase. This adds phase timing, then fixes what the measurements exposed.

Every phase now emits a line carrying a `PUBTIME` token, so one `grep PUBTIME` over a log export reconstructs a publish in order, across both the app and the `blossom_upload_service` package:

```
⏱️ PUBTIME upload.thumbnail.extract 220ms
⏱️ PUBTIME upload.thumbnail.blurhash 587ms
⏱️ PUBTIME auth.sign 279ms PUT /upload
⏱️ PUBTIME net.put 5274ms 16572 bytes -> 200
⏱️ PUBTIME upload.thumbnail 6403ms
⏱️ PUBTIME upload.transfer 8856ms 4.92MB 0.56MB/s
⏱️ PUBTIME nostr.sign 331ms signed
⏱️ PUBTIME nostr.publish 1160ms
⏱️ PUBTIME summary draft=… outcome=success total=11303ms upload=8965ms nostr=1525ms other=809ms
```

### What the measurements showed

Nine publishes on a real Samsung Galaxy. The 16 KB thumbnail cost 3.2–4.1 s and sat entirely behind the video transfer — a fifth to a quarter of the whole publish — and almost none of it was bandwidth. Splitting it further: ~0.3–1.4 s signing a Blossom auth event, and 2.3–7.2 s in a single PUT round-trip for 16 KB.

Two hypotheses were measured and disproved along the way, rather than shipped as fixes: a capability probe that turned out to be noise, and a suspected upload retry that never happened.

### What changed

**The thumbnail leg runs beside the video transfer.** It only reads the local file and never touches the transfer's result, so it no longer waits. It was the shorter leg in all nine runs, so it lands first and leaves the critical path entirely.

**The blurhash moved into that leg.** Publishing used to decode the video a second time to extract a frame for it — 568 ms measured — 220 ms after the upload leg had decoded the very same frame at the very same timestamp. The leg now derives it from the frame it already holds and persists it on the upload record (`PendingUpload.blurhash`, Hive index 26).

**The video is no longer hashed twice.** The publisher read the whole video back into memory to compute its SHA-256, which the upload had already streamed and returned as `videoId` — Blossom is content-addressed. The second pass also defeated the streaming the upload path does specifically to keep large files out of memory on iOS.

**The progress bar stopped lying.** It reached 100% when the upload finished and then sat there for the ~2 s the tail actually takes. The upload now owns 0–85%, and the tail hands out the rest at real boundaries: 88% once mentions resolve, 92% once the Nostr event is signed (a remote Keycast round-trip, 0.3–1.4 s), 97% once it is accepted, 100% when the draft is gone.

**A failed transfer no longer discards a finished thumbnail.** The reuse guard read a record that was only written once the whole attempt succeeded, so every retry re-extracted, re-signed and re-uploaded a blob already on the CDN.

## Measured effect

Comparable network conditions, before and after, on the same device:

| | before | after |
|---|---|---|
| network | 0.73 MB/s | 0.75 MB/s |
| total publish | 14 387 ms | **9 736 ms** |

**−32% at equal bandwidth.** The gain grows as conditions worsen: on a 0.22 MB/s run the same publish would have cost ~37 s sequentially and took 27 s.

Structurally, the thumbnail is now fully absorbed. `upload` equals the transfer plus ~110–140 ms in every run since, where it used to be transfer + thumbnail. In the run quoted above the thumbnail took 6 403 ms and cost nothing.

The unexplained gap inside the Nostr phase closed too: 734 ms before, 34 ms now.

## Out of Scope

Everything still on the clock is outside this app:

- `net.put` — 2 273 to 7 216 ms for a constant 16 572 bytes, with no relation to bandwidth. On a 50 Mbit/s line that is ~3 ms of transfer; the rest is latency or server-side work. Scoping note from review: the thumbnail PUT is served in-process at the Fastly edge (`media.divine.video` only proxies to `cloud-run-upload` above 500 KB or for video MIME types), so this cost lives at the edge, not in the upload service.
- The Keycast signatures. The account signs via `KeycastNostrIdentity` with `authSource=divineOAuth`, which has no local key, so every signature is an RPC round-trip. Two of the three now overlap the transfer; only the event signature is left on the critical path.
- Draft deletion and file cleanup, still ~0.4–1.3 s after the event is published — divinevideo/divine-mobile#6548.
- Orphaned blobs from publishes that never completed — divinevideo/divine-blossom#163.

Worth noting for whoever picks up the backend side: the **video** PUT (proxied to `cloud-run-upload` by MIME type) runs an ffmpeg thumbnail extraction **and** a dimension probe synchronously before responding, for every video — divinevideo/divine-blossom#164. The client extracts and uploads its own thumbnail anyway, so that work is duplicated and sits on the response path of the client's upload.

## Verification

- `flutter test test/services/ test/models/ test/repositories/ test/blocs/… test/providers/…` — 4 288 passing
- `blossom_upload_service` package suite — 228 passing, coverage 100.00% (958/958), gate held
- `dart format` clean; Hive adapter regenerated and committed
- Manually verified on a real Samsung Galaxy across nine publishes; every commit was measured on device before the next one was written

Each behavioural change is pinned by a test that was confirmed to fail without it:

- thumbnail parallelism — the video stub only resolves once the thumbnail leg has started
- progress — never moves backwards, never reports 100% before the event lands, and carries the signing step
- digest reuse — the emitted `x` component is the upload's `videoId` and explicitly not the SHA-256 of the file contents
- retry — one failed transfer, thumbnail uploaded exactly once (it uploaded twice before)

## Type of Change

- [x] 🧹 Code refactor

### DIFF
--- a/mobile/lib/services/upload_manager.dart
+++ b/mobile/lib/services/upload_manager.dart
@@ -25,6 +25,7 @@ import 'package:openvine/services/upload/upload_session_errors.dart';
 import 'package:openvine/services/upload_initialization_helper.dart';
 import 'package:openvine/services/video_editor/stop_motion_render_service.dart';
 import 'package:openvine/services/video_editor/video_editor_render_service.dart';
+import 'package:openvine/services/video_publish/publish_timeline.dart';
 import 'package:openvine/services/video_thumbnail_service.dart';
 import 'package:path/path.dart' as path;
 import 'package:path_provider/path_provider.dart';
@@ -120,6 +121,17 @@ class CrashReportingUploadReporter implements UploadCrashReporter {
       CrashReportingService.instance.recordError(error, stack, reason: reason);
 }
 
+/// Pulls a thumbnail frame out of a local video file.
+///
+/// Mirrors [VideoThumbnailService.extractThumbnail]; exists so [UploadManager]
+/// can be handed a stand-in that does not need a decodable video.
+typedef ThumbnailExtractor =
+    Future<ThumbnailFileResult?> Function({
+      required String videoPath,
+      required Duration targetTimestamp,
+      required int quality,
+    });
+
 /// Manages video uploads and their persistent state with enhanced reliability
 /// REFACTORED: Removed ChangeNotifier - now uses pure state management via Riverpod
 class UploadManager implements BackgroundAwareService {
@@ -133,7 +145,9 @@ class UploadManager implements BackgroundAwareService {
     UploadCrashReporter? crashReporter,
     this.useBackgroundUpload = false,
     BackgroundActivityManager? backgroundActivityManager,
+    ThumbnailExtractor? thumbnailExtractor,
   }) : _blossomService = blossomService,
+       _extractThumbnail = thumbnailExtractor ?? _defaultThumbnailExtractor,
        _defaultBlossomUrl =
            defaultBlossomUrl ?? BlossomUploadService.defaultBlossomServer,
        _circuitBreaker = circuitBreaker ?? VideoCircuitBreaker(),
@@ -169,6 +183,22 @@ class UploadManager implements BackgroundAwareService {
   // Core services
   final PendingUploadStore _store;
   final BlossomUploadService _blossomService;
+
+  /// Pulls the thumbnail frame out of the local video. Injectable so the
+  /// thumbnail leg can be exercised without a decodable video and the native
+  /// plugin behind [VideoThumbnailService].
+  final ThumbnailExtractor _extractThumbnail;
+
+  static Future<ThumbnailFileResult?> _defaultThumbnailExtractor({
+    required String videoPath,
+    required Duration targetTimestamp,
+    required int quality,
+  }) => VideoThumbnailService.extractThumbnail(
+    videoPath: videoPath,
+    targetTimestamp: targetTimestamp,
+    quality: quality,
+  );
+
   final String _defaultBlossomUrl;
   final VideoCircuitBreaker _circuitBreaker;
   final UploadRetryConfig _retryConfig;
@@ -1007,6 +1037,11 @@ class UploadManager implements BackgroundAwareService {
       // (passed as resumableTimeout), so a stalled socket does not hang
       // forever. `useBackgroundUpload` is the kill-switch for the OS-first
       // attempt: flip it to false to go straight to chunked resumable.
+      // Bytes are read before the transfer so the timing line can report
+      // throughput even if the file is gone by the time the phase ends.
+      final videoBytes = videoFile.existsSync() ? videoFile.lengthSync() : null;
+      final transferWatch = Stopwatch()..start();
+
       BlossomUploadResult result;
       try {
         result = await _blossomService.uploadVideoWithResume(
@@ -1050,6 +1085,13 @@ class UploadManager implements BackgroundAwareService {
         });
 
         rethrow;
+      } finally {
+        transferWatch.stop();
+        logPublishPhase(
+          'upload.transfer',
+          transferWatch.elapsed,
+          detail: formatTransferDetail(videoBytes, transferWatch.elapsed),
+        );
       }
 
       if (_userStoppedUploadIds.contains(upload.id)) {
@@ -1072,11 +1114,18 @@ class UploadManager implements BackgroundAwareService {
         );
 
         // Generate and upload thumbnail to Blossom CDN
-        thumbnailCdnUrl = await _generateAndUploadThumbnail(
-          videoFile: videoFile,
-          nostrPubkey: upload.nostrPubkey,
-          upload: upload,
-        );
+        final thumbnailWatch = Stopwatch()..start();
+        try {
+          thumbnailCdnUrl = await _generateAndUploadThumbnail(
+            videoFile: videoFile,
+            nostrPubkey: upload.nostrPubkey,
+            upload: upload,
+            onProgress: onProgress,
+          );
+        } finally {
+          thumbnailWatch.stop();
+          logPublishPhase('upload.thumbnail', thumbnailWatch.elapsed);
+        }
 
         if (thumbnailCdnUrl != null) {
           Log.info(
@@ -1810,11 +1859,23 @@ class UploadManager implements BackgroundAwareService {
   }
 
   /// Generate and upload thumbnail to Blossom CDN
+  /// Extracts a thumbnail from [videoFile] and uploads it to Blossom.
+  ///
+  /// [onProgress] receives the same absolute values written to the store, so
+  /// publish-facing listeners keep moving through the thumbnail leg instead of
+  /// freezing at the end of the video transfer (#6086 follow-up: the leg can
+  /// take seconds, and the bar sat still for all of it).
   Future<String?> _generateAndUploadThumbnail({
     required File videoFile,
     required String nostrPubkey,
     required PendingUpload upload,
+    ValueChanged<double>? onProgress,
   }) async {
+    void reportThumbnailProgress(double value) {
+      _reporter.updateProgress(upload.id, value);
+      onProgress?.call(value);
+    }
+
     try {
       Log.info(
         '📸 Extracting thumbnail from video: ${videoFile.path}',
@@ -1823,13 +1884,16 @@ class UploadManager implements BackgroundAwareService {
       );
 
       // Generate thumbnail at optimal timestamp
-      final thumbnailExtraction = await VideoThumbnailService.extractThumbnail(
+      final extractWatch = Stopwatch()..start();
+      final thumbnailExtraction = await _extractThumbnail(
         videoPath: videoFile.path,
         quality: 85,
         targetTimestamp:
             upload.thumbnailTimestamp ??
             VideoEditorConstants.defaultThumbnailExtractTime,
       );
+      extractWatch.stop();
+      logPublishPhase('upload.thumbnail.extract', extractWatch.elapsed);
 
       if (thumbnailExtraction == null) {
         Log.warning(
@@ -1856,12 +1920,14 @@ class UploadManager implements BackgroundAwareService {
         category: LogCategory.video,
       );
 
-      _reporter.updateProgress(upload.id, 0.85);
+      reportThumbnailProgress(0.85);
 
       // Upload thumbnail to Blossom server. Divine relay publishing requires
       // a CDN thumbnail, so keep the image upload's own retry enabled here.
       // If no HTTP thumbnail URL is available after that, the outer video
       // pipeline treats it as terminal to avoid re-uploading the whole video.
+      final thumbnailBytes = thumbnailFile.lengthSync();
+      final putWatch = Stopwatch()..start();
       final uploadResult = await _blossomService.uploadImage(
         imageFile: thumbnailFile,
         nostrPubkey: nostrPubkey,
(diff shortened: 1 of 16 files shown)


realmeylisdev (commented):
Traced the load-bearing claims end to end — the digest reuse, the two store writers that now overlap, and the composite progress curve all hold up. Four notes inline, none blocking.

One correction for the backend note: ffmpeg + the dimension probe can't be what `net.put` pays for on the 16 KB thumbnail. `media.divine.video` proxies to the upload service only when `content_length > 500 KB` **or** the MIME type is a video (`divine-blossom/src/main.rs:2723` and `3144`, with `image/jpeg` absent from `VIDEO_MIME_TYPES` in `blossom-core/src/types.rs:648`), so the 16 572-byte thumbnail is served in-process at the edge and never reaches it — and `cloud-run-upload`'s own `is_video_type` gate excludes `image/jpeg` too. The note is right for the *video* PUT; worth scoping it that way so the follow-up doesn't start on the wrong path.

mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
```diff
@@ -2677,11 +2708,19 @@ class BlossomUploadService {
           category: LogCategory.video,
         );
 
-        final capability = await _fetchDivineUploadCapability(serverUrl);
+        // The capability probe only picks between the resumable flow and the
+        // single PUT. A caller that already opted out of resumable cannot be
+        // routed by the answer, so the HEAD is dead weight — and an expensive
+        // one: the OS-first video upload never warms the capability cache, so
+        // the thumbnail leg pays a cold round-trip on the publish critical
+        // path (measured at ~3.7s for a 20KB thumbnail).
```
realmeylisdev: This comment and the PR description disagree, and the comment is the one that outlives the PR.

Here — and in e16a6a4's message ("Two costs the measurement exposed are fixed here") — the probe is a real cost measured at ~3.7 s. The description says the opposite: "Two hypotheses were measured and disproved along the way, rather than shipped as fixes: a capability probe that turned out to be noise." The skip is definitely shipped, so one of the two needs correcting.

Either keep the number and fix the description, or drop it and rest the skip on the routing argument alone — that stands by itself, since `capability` only ever feeds `supportsResumable` and an `allowResumable: false` caller could never be routed by the answer.

mobile/lib/services/video_event_publisher.dart
```diff
@@ -1598,6 +1627,15 @@ class VideoEventPublisher {
...
           );
+      signWatch.stop();
+      logPublishPhase(
+        'nostr.sign',
+        signWatch.elapsed,
+        detail: reusedEvent != null ? 'reused' : 'signed',
+      );
+      // Signing is a remote Keycast round-trip (~0.3-1.4s measured), so the
+      // caller gets a step here rather than waiting out the whole phase.
+      onEventSigned?.call();
```
realmeylisdev: Fires before the `event == null` check two lines down, so a failed signing still walks the bar to 92% and then the publish errors out. Moving it below the check keeps the callback matching its own contract and costs nothing — the reused-event path behaves the same either way.

mobile/lib/services/video_event_publisher.dart
```diff
@@ -1146,10 +1151,16 @@ class VideoEventPublisher {
...
-            final bytes = await videoFile.readAsBytes();
-            final hash = sha256.convert(bytes);
-            imetaComponents.add('x $hash');
+            // Blossom is content-addressed, so the upload already streamed
+            // this exact digest and handed it back as the videoId — which
+            // publishDirectUpload has already required to be present. Hashing
+            // the file again here cost a measurable slice of the publish and
+            // pulled the whole video into memory, which is exactly what the
+            // upload path streams to avoid.
+            final hash = upload.videoId;
```
realmeylisdev: Confirmed this is value-identical to the old `sha256.convert(await videoFile.readAsBytes())` rather than merely close: `videoId` is never the server's hash. Every `BlossomUploadResult` sets `videoId: fileHash`, `fileHash` is `HashUtil.sha256File(...)` over `File(upload.localVideoPath)` (`upload_manager.dart:856`) — the same file this block was re-reading — and `_parseUploadResponse` takes `fileHash` as a parameter, never reading `sha256` off the response body. So there's no trust boundary here at all, which is a stronger argument than "Blossom is content-addressed" and worth saying in the comment.

Optional follow-on: `x` is still inside the `videoFile.existsSync()` branch, which was load-bearing when the digest came from the file and no longer is. With #6548 moving file cleanup around, a publish landing after cleanup would drop `x` for no reason while the digest sits on the record. Lifting just `x` out of that branch (leaving `size` where it is) would make it survive.

mobile/lib/services/upload_manager.dart
```diff
@@ -1809,8 +1878,75 @@ class UploadManager implements BackgroundAwareService {
...
+      );
+
+      // Persist as soon as the leg lands, not once the whole attempt
+      // succeeds. A failing video transfer aborts this method's caller before
+      // its own store write, so deferring would make every retry re-extract
+      // and re-upload a thumbnail that is already on the CDN — and the
+      // short-circuit above would never fire.
+      if (result.cdnUrl != null || result.blurhash != null) {
+        try {
+          final current = _store.getUpload(upload.id) ?? upload;
```
realmeylisdev, line 1919: Checked the interleaving this opens up, since the leg now persists mid-transfer alongside `UploadProgressReporter.updateProgress` doing its own read-modify-write on the same record — it's safe. Hive's `_writeFrames` runs `keystore.beginTransaction(frames)` synchronously before the disk `await`, and both writers call `put` in the same microtask as their `getUpload`, so neither can clobber the other. `createSuccessfulUpload` then re-reads and `copyWith`s, so `thumbnailPath` and `blurhash` both reach publish intact.

Small observability gap: the reuse short-circuit returns before the stopwatch starts, so a retry emits no `upload.thumbnail` PUBTIME line at all. In an export that reads as "the leg never ran" rather than "it was reused" — a zero-duration line with a `reused` detail, like `nostr.sign` already does, would keep the timeline honest across retries.

mobile/lib/services/video_event_publisher.dart
realmeylisdev: Following up on the optional half — it has more teeth than I gave it credit for. In funnelcake, `events_local.sha256` is a MATERIALIZED column extracted from exactly this imeta `x` sub-field, defaulting to `''` when it's missing (`database/migrations/000025_fix_multi_imeta_extraction.up.sql:49-79`). Moderation labels are then joined to videos on that column — `WHERE ... AND sha256 IN (...)` in `get_moderation_labels_for_sha256` (`crates/clickhouse/src/client.rs:13957-13982`) — and audio reuse looks up by it too.

So a video published without `x` materializes an empty hash and can never match an applied moderation label. Nothing changes on this PR, since the `existsSync()` gate is exactly what it was before and the value is identical. But it does mean that when #6548 moves file cleanup, dropping `x` wouldn't just lose metadata — it would quietly take that video outside hash-keyed moderation. Worth lifting `x` out of the branch before that lands rather than after.

mobile/lib/services/video_event_publisher.dart
realmeylisdev: Pushed this one as `1a44ce9` — it was the only finding of the four where both the diagnosis and the fix were unambiguous, so it seemed better to hand you a commit than a paragraph. Revert it freely if you'd rather take it differently.

The callback moves below the null check; the `nostr.sign` timing line stays above it, since a failed round-trip is as worth measuring as a successful one. Added `video_event_publisher_signed_callback_test.dart` with both directions — I ran the failing case against your code first and it reported the step once for an event that was never signed, then zero after the move.

Validation: `flutter analyze lib test integration_test` clean, `dart format` clean, 138 tests across the publisher and `video_publish` suites, and the pre-push hook's 317. The other three findings are yours to call — none of them were clear enough for me to touch.

realmeylisdev:
`1a44ce9` on this branch moves `onEventSigned` below the null check on `createAndSignEvent` — a failed signing was walking the bar to 92% on its way to returning `false`. The `nostr.sign` timing line stays above it, since a failed round-trip is as worth measuring as a successful one. Test added in `video_event_publisher_signed_callback_test.dart`; I ran it against the previous code first and it reported the step once for an event that was never signed.

That was the only one of the four findings where the fix wasn't a judgement call, which is why it's the only one pushed. The probe-vs-description contradiction, the backend note, and lifting `x` out of the `existsSync()` branch are yours to decide. Revert freely if you'd rather take this another way.

CI: all 15 checks green.

mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
hm21: Agreed — the description matched the data, the comment didn't. Reworded in 5b99e31 to rest on the routing argument alone; the misattributed number is gone.

mobile/lib/services/video_event_publisher.dart
hm21: Real bug, and you'd already fixed it in 1a44ce9 by the time I got here — thanks. Dropped my duplicate pin in 5b99e31 in favor of your dedicated test file.

mobile/lib/services/video_event_publisher.dart
hm21: Both taken in 5b99e31: the comment now states the no-trust-boundary argument (videoId is always the locally streamed digest, never read off the response), and x is lifted out of the existsSync branch so it survives post-cleanup publishes — pinned by a test that deletes the file and asserts x is still emitted. Verified the funnelcake side (migration 000025 + get_moderation_labels_for_sha256) before moving it; good catch, that would have been a nasty silent moderation gap after #6548.

mobile/lib/services/upload_manager.dart
hm21: Line added in 5b99e31 — zero-duration upload.thumbnail with a reused detail, mirroring nostr.sign, asserted in the retry test. Thanks for tracing the Hive interleaving; that analysis is why the mid-transfer persist stayed.

hm21:
@realmeylisdev All four addressed in 5b99e31 (plus your own 1a44ce9 for the callback ordering). Also verified your edge-routing correction against divine-blossom source — the 500 KB / video-MIME proxy gate checks out — and rescoped the PR description's backend note and divinevideo/divine-blossom#164 accordingly. CI rerunning.

realmeylisdev (approved):
All four check out — re-verified against the diff rather than the descriptions, and both new tests fail against the pre-fix sources, so they can catch the regression. Shard 3's red was `DmUnreadCountCubit … counts an adopted moderation thread exactly once`: seed-dependent, green on a rerun of the same SHA, and nothing DM-shaped is in this diff.

One nit, non-blocking: the description still files the capability probe under "measured and disproved along the way, rather than shipped as fixes", but the skip does ship. It was the cost attribution that was wrong, not the code.

PR #6558, opened 2026-07-31
fix(test): wait for the save result before asserting the stop-motion cleanup

### LINKED ISSUE #6557: fix(test): clips-library stop-motion cleanup test is flaky under the merged runner
## Summary

`ClipsLibraryBloc ClipsLibrarySaveToGallery cleans up the transient mp4 after saving a stop-motion clip` fails intermittently in CI. It is nondeterministic even at a fixed ordering seed, so re-running the job turns it green.

## Environment

- CI, `Tests (shard 3/4)` — `very_good test --optimization --concurrency=4 --exclude-tags integration --test-randomize-ordering-seed random`
- Reproduced locally on macOS with the same command
- `mobile/test/blocs/clips_library/clips_library_bloc_test.dart:1107`

## Steps to Reproduce

```bash
cd mobile
bash scripts/ci/select_test_shard.sh --total 4 --index 3 --force
very_good test --optimization --concurrency=4 --exclude-tags integration \
  --test-randomize-ordering-seed 11111
git checkout -- test   # restore the files the shard script deleted
```

It failed on the first run at that seed and passed on the second, unchanged.

## Actual Result

```
Expected: false
  Actual: <true>

test/blocs/clips_library/clips_library_bloc_test.dart 1107:11  main.<fn>.<fn>.<fn>
package:bloc_test/src/bloc_test.dart 228:21                    testBloc.<fn>
```

The transient mp4 still exists when `verify` runs, so `StopMotionRenderService.cleanupMaterializedOutput` either had not deleted it yet or did not delete it.

## Expected Result

Passes on every run, at every seed.

## Regression?

Unknown — surfaced on a feature branch whose base commit (`bc091d29`) was green, and whose diff does not touch this area. The branch's own new test files are not in shard 3 at all (confirmed by running the shard script and checking which files survive), so they cannot be leaking state into this suite.

## Evidence

| Run | Result |
|---|---|
| Test alone | pass |
| Full merged suite locally, random seed (14 048 tests) | pass |
| Shard 3, seed 11111, first attempt | **fail** |
| Shard 3, seed 11111, second attempt | pass |
| CI job re-run, unchanged commit | pass |

Same seed, different outcomes — so ordering is not the variable.

## Notes for whoever picks this up

Two things about this test make it a plausible race under `--concurrency=4`:

- It configures `StopMotionRenderService.assembleOverride` and `probeDurationOverride`, which are process-global mutable statics. `AGENTS.md` covers this class of hazard for channels and `Platform.instance` singletons; static overrides on a service are the same shape and are not covered by `check_process_global_mutations.sh`.
- The assertion is on filesystem state (`File(...).existsSync()`) written and deleted by different code paths, rather than on an observable the bloc emits.

Worth checking whether the deletion in `ClipsLibraryBloc`'s `finally` is actually awaited before `blocTest`'s `verify` runs, and whether any other suite in the bundle touches those statics without restoring them.

hm21:
## Description

Fixes the intermittent failure of `ClipsLibraryBloc ClipsLibrarySaveToGallery cleans up the transient mp4 after saving a stop-motion clip`.

**Root cause is in the test harness, not the bloc.** `blocTest` runs `act`, waits a single `Duration.zero` turn, then calls `bloc.close()`. `Bloc.close()` calls `emitter.cancel()`, and `_Emitter.cancel()` completes the emitter's future *immediately* — it never awaits the handler body. Everything past the handler's current `await` keeps running detached, and `verify` can run before it lands.

That one turn is plenty for a chain of microtasks (the mocked gallery service, the `async` render overrides), which is why the test usually passes. It is not enough for `File.delete()`: that goes to the I/O thread pool, and whether its completion beats the zero-duration timer depends on machine load. Hence green locally, green on a re-run at the *same* ordering seed, red under CI's `--concurrency=4`.

The fix waits for the state the handler emits after the cleanup. `cleanupMaterializedOutput` runs in the save loop's `finally`, which the handler awaits before emitting `lastGallerySaveResult` — so that state is a real barrier for the delete, not a timing bet. `bloc.stream.firstWhere` inside `act` is already the idiom used in `clip_editor_bloc_test.dart`, `conversation_list_bloc_test.dart` and others.

Deliberately **not** `wait:` or a longer `pumpEventQueue` — both are the same race with a bigger number attached.

The two hypotheses in the issue were checked and ruled out:

- **Leaking statics.** All four test files that set `StopMotionRenderService.assembleOverride` / `probeDurationOverride` reset them in a `tearDown`. Leakage would also be ordering-dependent, and the issue reports the same seed producing both outcomes.
- **Deletion not awaited in the bloc's `finally`.** It is awaited; no production change is needed.

**Related Issue:** Closes #6557

## Out of Scope

`video_recorder_bloc_test.dart` has four assertions of the same shape (filesystem state in `verify`), but they use `await pumpEventQueue()` in `act`, which yields real event-loop turns and is far more robust. None have been reported flaky, so they are left alone.

## Verification

Confirmed the mechanism with a throwaway repro reproducing exactly this shape — a handler with `finally { await file.delete(); }` and the assertion in `verify`:

| Repro | Result |
|---|---|
| Bare `bloc.add` in `act`, 300 iterations | **1 failure** |
| Awaiting the terminal state in `act`, 300 iterations | 300/300 pass |

On the real suite:

- `flutter test test/blocs/clips_library/clips_library_bloc_test.dart` — 12 consecutive runs, 12/12 green (41 tests)
- `dart format` — no changes

Not run: the full shard-3 reproduction from the issue. Its hit rate is roughly one failure per several 14k-test runs, which is weaker evidence than the 300-iteration measurement against the isolated mechanism.

No device testing — the diff is one test file, with no runtime behaviour change.

## Type of Change

### DIFF
--- a/mobile/test/blocs/clips_library/clips_library_bloc_test.dart
+++ b/mobile/test/blocs/clips_library/clips_library_bloc_test.dart
@@ -1096,7 +1096,18 @@ void main() {
           selectedDuration: const Duration(seconds: 1),
         ),
         build: createBloc,
-        act: (bloc) => bloc.add(const ClipsLibrarySaveToGallery()),
+        // Wait for the save-result state before asserting on disk. The cleanup
+        // runs in the save loop's `finally`, which the handler awaits before
+        // emitting that state — but `blocTest` closes the bloc without waiting
+        // for an in-flight handler, so a bare `add` would race the real (async)
+        // file delete and see the mp4 still there.
+        act: (bloc) async {
+          final saved = bloc.stream.firstWhere(
+            (s) => s.lastGallerySaveResult is GallerySaveResultSuccess,
+          );
+          bloc.add(const ClipsLibrarySaveToGallery());
+          await saved;
+        },
         verify: (_) {
           final captured =
               verify(


realmeylisdev (approved):
Verified against `bloc` 9.2.1 / `bloc_test` 10.0.0: `_Emitter.cancel()` completes the emitter's future without awaiting the handler body, so `verify` can outrun an in-flight `await file.delete()`. The barrier is sound — the `GallerySaveResultSuccess` emit is strictly after the save loop's `finally`, and `firstWhere` subscribes before the `add`.

Reproduced the exact failure (`Expected: false / Actual: <true>`) in an isolated repro of the same shape: 1 red in 3200 bare-`add` iterations at `--concurrency=4` under disk load, vs 3208/3208 green with the awaited state. Also confirmed this is the only `blocTest` in the repo that asserts filesystem state without an event-loop yield in `act`.

PR #4682, opened 2026-05-23
refactor: replace mediakit with the new native-video-player

hm21:
## Description

This PR replaces the old MediaKit video-player with the new DivineVideoPlayer which uses native-video-player apis.

Additionally, this PR fixes a brief flash of the paused/play overlay that appeared when scrolling to the next video in the feed. The `PausedVideoOverlay` now debounces the promotion of its "visible & paused" latch by 300 ms, so the transient `paused + firstFrameRendered` emission a pooled controller produces during a swipe no longer triggers the play affordance. A genuinely user-paused video still shows the play button after the short delay.

Closes #4684

## Out of Scope

None.

## Verification

- `flutter analyze lib/widgets/video_feed_item/paused_video_overlay.dart test/widgets/video_feed_item/paused_video_overlay_test.dart` — clean
- `flutter test test/widgets/video_feed_item/paused_video_overlay_test.dart` — passes
- Manual: swiping between videos in the feed no longer flashes the play overlay; pausing a video manually still surfaces it.

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [x] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #6559, opened 2026-07-31
chore(feed): delete the dead _LoadingIndicator

### LINKED ISSUE #6492: chore(feed): delete the dead _LoadingIndicator in pooled_fullscreen_video_feed_screen
The private `_LoadingIndicator` / `_LoadingIndicatorState` pair at `mobile/lib/screens/feed/pooled_fullscreen_video_feed_screen.dart:959-998` is never instantiated. The only four references in the repo are its own class declaration, constructor, `createState`, and the State class; it is file-private and the file has no `part` directive, so nothing outside can reach it. (The other `_LoadingIndicator`s that grep turns up — in `add_to_list_dialog.dart`, `camera_permission_gate.dart`, `video_loading_placeholder.dart` — are separate file-scoped classes and are all used.)

Surfaced during review of #6489, which deleted the seek-based loop enforcement. The dead widget carries the last non-doc reference to that mechanism: its `_delay` constant is justified with "flashes that occur during play/pause and loop-enforcement seeks". Trimming the comment was suggested there, but polishing a widget that never builds is the wrong fix — the widget should go.

Deliberately left out of #6489: that screen is untouched by its diff, and a ~40-line removal in an unrelated file is scope creep on a chore PR.

Closes #none

hm21:
## Description

Deletes the private `_LoadingIndicator` / `_LoadingIndicatorState` pair in `mobile/lib/screens/feed/pooled_fullscreen_video_feed_screen.dart`. It is never instantiated — the only four references in the repo were its own class declaration, constructor, `createState`, and the State class. It is file-private and the file has no `part` directive, so nothing outside could reach it either.

The other `_LoadingIndicator`s that grep turns up (`add_to_list_dialog.dart`, `camera_permission_gate.dart`, `video_loading_placeholder.dart`) are separate file-scoped classes and all in use.

Why delete rather than polish: the widget carried the last non-doc reference to the seek-based loop enforcement removed in #6489 — its `_delay` constant was justified with "flashes that occur during play/pause and loop-enforcement seeks". Trimming that comment was suggested during review of #6489, but touching up a widget that never builds is the wrong fix.

**Related Issue:** Closes #6492

## Out of Scope

None. Pure deletion, no behaviour change.

## Verification

- Confirmed no imports go stale: `dart:async` is still needed for `unawaited`/`Future`, and `BrandedLoadingIndicator` is still used at two other call sites in the same file.
- `flutter analyze` on the changed file — no issues.
- `flutter test test/screens/feed/pooled_fullscreen_video_feed_screen_test.dart test/screens/feed/pooled_fullscreen_video_feed_args_test.dart test/screens/feed/video_tap_shield_test.dart test/screens/feed/keyboard_aware_top_fade_test.dart` — 47/47 pass.
- No device verification: the deleted widget never built, so there is nothing observable to check on a real device.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/mobile/lib/screens/feed/pooled_fullscreen_video_feed_screen.dart
+++ b/mobile/lib/screens/feed/pooled_fullscreen_video_feed_screen.dart
@@ -964,47 +964,6 @@ class _FullscreenFeedContentState extends ConsumerState<FullscreenFeedContent>
   }
 }
 
-class _LoadingIndicator extends StatefulWidget {
-  const _LoadingIndicator();
-
-  @override
-  State<_LoadingIndicator> createState() => _LoadingIndicatorState();
-}
-
-class _LoadingIndicatorState extends State<_LoadingIndicator> {
-  // Delay before the indicator becomes visible. Suppresses sub-threshold
-  // flashes that occur during play/pause and loop-enforcement seeks without
-  // hiding the indicator during genuine long loads.
-  static const _delay = Duration(milliseconds: 100);
-  static const _fadeDuration = Duration(milliseconds: 150);
-
-  bool _visible = false;
-  Timer? _timer;
-
-  @override
-  void initState() {
-    super.initState();
-    _timer = Timer(_delay, () {
-      if (mounted) setState(() => _visible = true);
-    });
-  }
-
-  @override
-  void dispose() {
-    _timer?.cancel();
-    super.dispose();
-  }
-
-  @override
-  Widget build(BuildContext context) {
-    return AnimatedOpacity(
-      duration: _fadeDuration,
-      opacity: _visible ? 1.0 : 0.0,
-      child: const Center(child: BrandedLoadingIndicator(size: 60)),
-    );
-  }
-}
-
 /// Wraps the fullscreen video feed in a [NavRoundedShell] when the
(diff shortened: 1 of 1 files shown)


mbradley (approved):
Verified the deletion independently: the four `_LoadingIndicator` references on `main` are all self-referential (declaration, constructor, `createState`, State class), the file has no `part` directive so nothing external could reach it, and the three same-named classes elsewhere (`add_to_list_dialog.dart`, `camera_permission_gate.dart`, `video_loading_placeholder.dart`) are separate file-scoped widgets that are all instantiated. `dart:async` stays load-bearing via `unawaited`, and `BrandedLoadingIndicator` is still used at lines 711 and 1033. Analyze clean and 47/47 on the four cited test files locally.

Also traced *why* it was dead, since a deletion is the wrong move if a call site was dropped by accident. It wasn't: `d1486f3e8` (#4682, native-video migration) removed the entire containing widget wholesale and left this pair orphaned. The affordance itself moved rather than disappearing — this screen renders through `FeedVideos`, which renders `VideoLoadingPlaceholder`. So there's no regression being frozen in place here.

PR #6560, opened 2026-07-31
fix(video-player): clamp an over-long audio track to the video end

### LINKED ISSUE #6486: fix: relax the common-track-end clamp when the audio track is the longer one
## Summary

`boundedCommonTrackEndMs` only clamps a video/audio length mismatch when the trim stays under `min(500 ms, 10% of playable duration)`. Beyond that it returns `null` and the loop runs to the container duration — a frozen last frame before every restart.

The bound protects a real case: an audio track much *shorter* than the video must not collapse a real video into a tiny loop, because clamping there discards visible content.

The opposite direction has no such risk. When the **audio** is the longer track, the excess plays over a frozen frame — always wrong, and often there is no audio there at all to lose.

## Evidence

`mobile/assets/seed_media/videos/0189ef5a96928da7057587d0ed077a6e023fd6f55941c471a0dedb020833c31c.mp4` (archived Vine, `Lavf56.36`):

| | declared | last packet |
|---|---|---|
| video | 6.533 s | 6.500 s |
| audio | **7.287 s** | 6.525 s |
| container | 7.287 s | — |

`ffmpeg` extracts no frame at all past 6.6 s. The 762 ms the audio track claims beyond its last packet are a wrong length in the mux — so clamping to the video end discards **zero** audio packets.

The trim needed is 754 ms, over the 500 ms bound, so neither platform clamps it: **754 ms of frozen frame and silence before every restart**, on iOS and Android alike.

## Scope

No video in a 59-video live feed sample exceeds the bound (worst case 69 ms), so this is archive-corpus material rather than fresh uploads. #6469 fixed the in-bound cases; this is the remainder.

## Proposal

Make the bound asymmetric: keep `min(500 ms, 10%)` when the video track is the longer one, and allow an unbounded clamp to the video end when the audio track is. Both platforms share the constants, so the change lands in `DivineVideoPlayerInstance.kt` and `DivineVideoPlayerInstance.swift` together.

hm21:
## Description

`boundedCommonTrackEndMs` clamps a looping clip to the point where both tracks still have content, but only while the trim stays under `min(500 ms, 10% of playable duration)`. Past that it returns `null` and the loop runs to the container duration.

That bound protects one direction only. An audio track much *shorter* than the video must not collapse a real video into a tiny loop, because clamping there discards frames the viewer can see. Where the **audio** is the longer track there is nothing to lose: the excess plays over a frozen last frame, which is the seam itself whatever its length.

So the bound becomes asymmetric — unchanged when the video track is the longer one, skipped entirely when the audio track is.

The relaxed direction deliberately carries no lower floor. Feed clips are Vine-length — every video track in the seed corpus runs 6.47–6.53 s — so clamping to the video end can never discard a meaningful stretch, and a frozen last frame is not the better outcome at any length. A floor would only guard against a mux declaring a wrongly *short* video track, which is the opposite of what the corpus shows: all eight seed videos are audio ≥ video.

The archived Vine measured in #6486 (`0189ef5a…mp4`, `Lavf56.36`) declares 6.533 s of video against 7.287 s of audio, while the audio's own last packet sits at 6.525 s and `ffmpeg` extracts no frame past 6.6 s. The 762 ms the mux claims beyond that last packet are a wrong length, so clamping to the video end discards zero audio. The trim needed is 754 ms — over the bound, so neither platform clamped it, and every loop restart replayed 754 ms of frozen frame and silence on iOS and Android alike.

Both platforms share the constants, so the early return lands in `DivineVideoPlayerInstance.kt` and `DivineVideoPlayerInstance.swift` together, and the cross-platform contract test pins the asymmetry for both source files.

Scope note from the issue: no video in a 59-video live feed sample exceeds the bound (worst case 69 ms), so this is archive-corpus material rather than fresh uploads. #6469 fixed the in-bound cases; this is the remainder.

**Related Issue:** Closes #6486

## Out of Scope

The editor-side `commonTrackEnd` in `mobile/lib/services/video_editor/clip_media_duration.dart` has its own `minCommonTrackEnd` floor and is unrelated to loop seams — left untouched.

## Verification

- `gradle :divine_video_player:testDebugUnitTest` — green. Two new Kotlin tests use the issue's numbers: 6.533 s video / 7.287 s audio clamps to 6 533 ms, and the same 754 ms mismatch the other way round stays unclamped. Red/green verified — with the early return removed the first one fails.
- `flutter test` in `mobile/packages/divine_video_player` — 278 passed, including the extended contract test.
- `flutter analyze`, `dart format`, `swiftc -parse` on the Swift file — clean.
- Device pass on a real Samsung Galaxy (SM-S942B, Android 16): debug build installed from this branch, feed plays through loop restarts, no regression in playback.
- Also run on macOS, which builds and executes the same `DivineVideoPlayerInstance.swift` the iOS player uses: app launches, feed plays, no new player errors.
- iOS itself is parse-checked only, not built on a device.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
+++ b/mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
@@ -958,6 +958,13 @@ internal class DivineVideoPlayerInstance(
         val trimMs = playbackEndMs - commonEndMs
         if (commonEndMs <= startMs || trimMs <= 0) return null
 
+        // The bound only exists to stop a stub audio track collapsing a real
+        // video into a tiny loop, and that risk only runs one way. Where the
+        // audio track is the longer one the clamp lands on the video's own
+        // end, and everything past it is audio over a frozen frame — which is
+        // the seam itself, whatever its length.
+        if (audioEndMs > videoEndMs) return commonEndMs
+
         val playableDurationMs = playbackEndMs - startMs
         if (playableDurationMs <= 0) return null
 
@@ -1772,7 +1779,12 @@ internal class DivineVideoPlayerInstance(
          */
         private const val MIN_PLAYBACK_SPEED = 0.001f
 
-        /** Maximum tail considered an encoder/export track-end mismatch. */
+        /**
+         * Maximum tail considered an encoder/export track-end mismatch, for
+         * the direction where clamping discards video the viewer can see —
+         * an audio track shorter than the video. The other direction is
+         * unbounded; see [boundedCommonTrackEndMs].
+         */
         private const val MAX_COMMON_TRACK_END_TRIM_MS = 500L
         private const val MAX_COMMON_TRACK_END_TRIM_RATIO = 0.10
 
--- a/mobile/packages/divine_video_player/android/src/test/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstanceTest.kt
+++ b/mobile/packages/divine_video_player/android/src/test/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstanceTest.kt
@@ -899,6 +899,47 @@ class DivineVideoPlayerInstanceTest {
     }
 
     @Test
+    fun `an audio track longer than the video is clamped past the trim bound`() {
+        every { mockPlayer.mediaItemCount } returns 1
+        every { mockPlayer.getMediaItemAt(0) } returns MediaItem.Builder().build()
+
+        // An archived Vine: 6.533 s of video against a 7.287 s audio track,
+        // whose own last packet is at 6.525 s. The 754 ms the mux claims past
+        // the video are a frozen last frame over silence, and they are over
+        // the 500 ms bound that guards the opposite mismatch.
+        withTrackDurations(videoUs = 6_533_000L, audioUs = 7_287_000L) {
+            instance.onMethodCall(
+                trimmingSetClipsCall("https://cdn.example/long-audio.mp4"),
+                mockk(relaxed = true),
+            )
+            capturePostedRunnables().forEach { it.run() }
+        }
+
+        val replaced = slot<MediaItem>()
+        verify { mockPlayer.replaceMediaItem(0, capture(replaced)) }
+        assertEquals(6_533L, replaced.captured.clippingConfiguration.endPositionMs)
+    }
+
+    @Test
+    fun `a video track longer than the audio keeps the trim bound`() {
+        every { mockPlayer.mediaItemCount } returns 1
+        every { mockPlayer.getMediaItemAt(0) } returns MediaItem.Builder().build()
+
+        // The same 754 ms mismatch the other way round. Clamping here would
+        // discard video the viewer can see, so a stub audio track must not be
+        // allowed to cut the clip short.
+        withTrackDurations(videoUs = 7_287_000L, audioUs = 6_533_000L) {
+            instance.onMethodCall(
+                trimmingSetClipsCall("https://cdn.example/short-audio.mp4"),
+                mockk(relaxed = true),
+            )
+            capturePostedRunnables().forEach { it.run() }
+        }
+
+        verify(exactly = 0) { mockPlayer.replaceMediaItem(any(), any()) }
+    }
+
+    @Test
     fun `a resolved clamp is not applied over a playlist already playing`() {
         every { mockPlayer.mediaItemCount } returns 1
         every { mockPlayer.getMediaItemAt(0) } returns MediaItem.Builder().build()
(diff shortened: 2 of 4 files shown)


mobile/packages/divine_video_player/android/src/test/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstanceTest.kt
```diff
@@ -898,6 +898,47 @@ class DivineVideoPlayerInstanceTest {
...
+        every { mockPlayer.mediaItemCount } returns 1
+        every { mockPlayer.getMediaItemAt(0) } returns MediaItem.Builder().build()
+
+        // An archived Vine: 6.533 s of video against a 7.287 s audio track,
+        // whose own last packet is at 6.525 s. The 754 ms the mux claims past
+        // the video are a frozen last frame over silence, and they are over
+        // the 500 ms bound that guards the opposite mismatch.
+        withTrackDurations(videoUs = 6_533_000L, audioUs = 7_287_000L) {
+            instance.onMethodCall(
+                trimmingSetClipsCall("https://cdn.example/long-audio.mp4"),
```
realmeylisdev, line 912: `trimmingSetClipsCall` sends no `endMs`, but every production caller sends one. `trimToCommonTrackEnd: true` is set only inside `infinite_video_feed`, and its single instantiation (`feed_videos.dart:371`) always passes `maxPlaybackDuration: AppConstants.maxFeedPlaybackDuration` = 7 s. `VideoClip.toMap()` forwards it as `endMs`.

With `endMs = 7000`, this asset's arithmetic through `boundedCommonTrackEndMs` is:

```
playbackEndMs   = min(7000, 7287) = 7000
commonEndMs     = min(6533, 7287) = 6533
trimMs          = 7000 - 6533     =  467
trimLimitMs     = min(500, 700)   =  500   -> 467 <= 500, clamps
```

So the 6 533 ms this asserts is what `main` already produces — the assertion holds with the early return deleted, and the red/green check passes only because the helper drops `endMs`. Adding `"endMs" to 7000` to the helper (or to these two cases) would surface it, and I think it's worth doing before the pin is trusted.

Same reasoning applies to the second test: at `endMs = 7000` a 7 287 ms video against 6 533 ms of audio gives `playbackEndMs = 7000`, `commonEndMs = 6533`, `trimMs = 467` — that one *would* clamp on `main`, so `verify(exactly = 0) { replaceMediaItem }` is pinning the no-`endMs` shape too.

mobile/packages/divine_video_player/android/src/main/kotlin/com/divinevideo/divine_video_player/DivineVideoPlayerInstance.kt
```diff
@@ -958,6 +958,13 @@ internal class DivineVideoPlayerInstance(
         val trimMs = playbackEndMs - commonEndMs
         if (commonEndMs <= startMs || trimMs <= 0) return null
 
+        // The bound only exists to stop a stub audio track collapsing a real
+        // video into a tiny loop, and that risk only runs one way. Where the
+        // audio track is the longer one the clamp lands on the video's own
+        // end, and everything past it is audio over a frozen frame — which is
+        // the seam itself, whatever its length.
+        if (audioEndMs > videoEndMs) return commonEndMs
```
realmeylisdev, line 966: The relaxed direction leaves `commonEndMs > startMs` as the only remaining guard, so the clamp can now land anywhere above zero. That drops the bound's second job: capping how far a *wrong* video-track length can move the clip end. Any disagreement between the two reads — edit lists, CT offsets, `MediaFormat.KEY_DURATION` (a duration) vs `AVAssetTrack.timeRange.end` (start + duration) — used to cost at most 500 ms and is now unbounded on this branch.

The concrete case is a still image over audio, which is the one shape where the frozen frame is the content rather than the seam. A plain `ffmpeg -i art.png -i tone.m4a -c:v libx264 -c:a copy out.mp4` (no `-loop 1`) writes a **40 ms** video track against **10 s** of audio — I checked:

```
index=0  codec_type=video  duration=0.040000  nb_frames=1
index=1  codec_type=audio  duration=10.000000
```

On `main` that plays 7 s over the still. On this branch `audioEndMs > videoEndMs` clamps it to 40 ms, and the feed loops 40 ms. A 2 s clip with a 10 s music bed collapses to 2 s the same way.

That corpus is wider than Divine's own uploads: curated lists read external events with `getAllAcceptableVideoKinds()` (21 / 22 / 34235 / 34236 / 34237) in permissive mode (`list_providers.dart`), and those render through the same `FeedVideos` -> `InfiniteVideoFeed` path.

Keeping just the ratio arm — drop `MAX_COMMON_TRACK_END_TRIM_MS` in this direction but keep `MAX_COMMON_TRACK_END_TRIM_RATIO` — would still clamp the 754 ms archive case unconditionally while refusing to cut a clip to a fraction of its playable length. Sibling `commonTrackEnd` in `clip_media_duration.dart` carries a floor for the same reason.

Same comment applies to `DivineVideoPlayerInstance.swift:893`.

realmeylisdev (commented):
The asymmetry reads right in principle, but I couldn't reproduce the seam it targets. Under the feed's 7 s clip end the trim for `0189ef5a…` is 467 ms, not 754 — inside the existing bound, so both platforms already clamp it. I probed 64 live kind-34236 videos off `relay.divine.video`: none change behaviour, largest audio-over-video gap 81 ms.

Two things inline — the Kotlin tests omit the `endMs` every production caller sends, and the relaxed direction is left with no floor at all.

hm21:
Closing — this was already fixed by #6469; the remaining case described here does not reproduce on device.

PR #6561, opened 2026-07-31
perf(publish): reclaim published drafts off the publish critical path

### LINKED ISSUE #6548: task(upload): move draft deletion and file cleanup off the publish critical path

hm21:
## Description

`VideoPublishService.publishVideo` deleted the draft and reclaimed its files *before* returning `PublishSuccess` — the video was already live at that point, so the user was waiting on garbage collection.

**Ownership is settled on the bloc.** `BackgroundPublishBloc._deletePublishedDrafts` already deletes the publish copy *and* the source draft it was copied from (which the service never saw), after emitting the success state. Every publish result flows through that handler — the provider path and the retry path both hand their future to `BackgroundPublishRequested` — so the service-side deletion was redundant rather than merely mistimed. It is removed, and both sides now document the bloc as the sole owner. A failed reclaim is still surfaced: `_deleteDraft` logs and calls `addError`.

**The full-table clip scan is narrowed.** `deleteDraft` → `_referencedGhostFrameFilenames` loaded *every* clip row in the database and JSON-decoded each `data` blob to find ghost frames a surviving draft still references. That is O(all clips), grows as the clip library grows, and ran twice per publish (publish copy + source draft). New `ClipsDao.getClipsWithGhostFrames()` filters in SQL on the serialized `ghostFramePath` key, so only the few rows that actually carry one cross into Dart and get decoded. The decode and corrupt-blob handling stay where they were.

The SQL pre-filter depends on `DivineVideoClip.toJson`'s encoding, which is why it is pinned end-to-end (real model → real DAO) by the existing shared-ghost-frame tests in `draft_storage_service_test.dart` rather than only by a string-level unit test — if serialization ever drifts, those go red instead of silently deleting a shared ghost frame.

**Related Issue:** Closes #6548

## Out of Scope

- `saveDraft(publishing)` at the start of `publishVideo`, mentioned in the issue as "a smaller version of the same shape". Its orphan diff is O(1) in stored drafts (one row read, and for a status-only flip the orphan list is empty), so it is not part of the growth this issue targets.
- Separately noticed, not fixed here: `video_publish_provider` saves `publishDraft` immediately before handing it to `publishVideo`, which then saves the same draft again with the same status — a redundant full draft+clips upsert on every publish. Left alone because it touches publish-resume semantics; happy to file a follow-up.
- The PUBTIME `other` measurement in the issue's acceptance criteria cannot be reproduced on `main` — that instrumentation lives on `feat/publish-phase-timing`.

## Verification

- `flutter analyze lib test` — clean, in `mobile/` and in `mobile/packages/db_client/`
- `dart format` — clean on all changed files
- `flutter test test/services/video_publish/` (72)
- `flutter test test/blocs/background_publish_bloc/background_publish_bloc_test.dart` (34)
- `flutter test test/services/draft_storage_service_test.dart`, `test/providers/video_publish_provider_test.dart`, `test/providers/video_editor_provider_test.dart`, `test/services/clip_library_service_test.dart` (173 combined)
- `flutter test` in `mobile/packages/db_client` (213)
- Mutation-checked the new ordering test: moving `_deletePublishedDrafts` back in front of the `emit` fails it with `Expected: contains '1' / Actual: Set:[]`, so it is not a test that passes either way.

### Test plan

- [ ] Manual verification on a real Samsung Galaxy: publish a video with a populated draft library, confirm the success state lands promptly, the draft (and its source draft) disappear from the library afterwards, and a stop-motion draft sharing a ghost frame with a duplicate keeps its frames.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/services/draft_storage_service.dart
+++ b/mobile/lib/services/draft_storage_service.dart
@@ -703,14 +703,18 @@ class DraftStorageService {
   /// the last referencing clip is gone. Never throws: a corrupt blob is logged
   /// and skipped.
   ///
-  /// Trashed clips are included (`includeTrashed: true`): a soft-deleted clip
-  /// can still be restored, so a ghost frame it shares with the draft being
-  /// deleted must survive. This mirrors [ClipsDao.isFileReferenced], which
-  /// already protects the indexed video/thumbnail files of trashed clips.
+  /// Trashed clips are included: a soft-deleted clip can still be restored, so
+  /// a ghost frame it shares with the draft being deleted must survive. This
+  /// mirrors [ClipsDao.isFileReferenced], which already protects the indexed
+  /// video/thumbnail files of trashed clips.
+  ///
+  /// Uses [ClipsDao.getClipsWithGhostFrames] rather than reading the whole clip
+  /// table: this scan runs on every draft deletion, and decoding every blob in
+  /// the database made it grow with the size of the clip library (#6548).
   Future<Set<String>> _referencedGhostFrameFilenames({
     String? excludeDraftId,
   }) async {
-    final clipRows = await _clipsDao.getAllClips(includeTrashed: true);
+    final clipRows = await _clipsDao.getClipsWithGhostFrames();
     final filenames = <String>{};
 
     for (final row in clipRows) {
--- a/mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
@@ -82,9 +82,9 @@ class ClipsDao extends DatabaseAccessor<AppDatabase> with _$ClipsDaoMixin {
   /// provided, returns only clips owned by that account **plus** legacy
   /// clips with no owner.
   ///
-  /// [includeTrashed] is for file-reference scans that must keep a
-  /// trashed (restorable) clip's files alive — active-clip listings should
-  /// leave it at its default.
+  /// [includeTrashed] keeps trashed (restorable) clips in the result — useful
+  /// when the caller must account for a clip that can come back. Active-clip
+  /// listings should leave it at its default.
   Future<List<ClipRow>> getAllClips({
     int? limit,
     String? ownerPubkey,
@@ -106,7 +106,9 @@ class ClipsDao extends DatabaseAccessor<AppDatabase> with _$ClipsDaoMixin {
   }
 
   /// Every clip row that carries a non-null `ghostFramePath`, projected to the
-  /// two fields a ghost-frame reference scan needs.
+  /// fields a ghost-frame reference scan needs: `draftId` to skip the draft
+  /// being deleted, `data` to read the frame out of, and `id` to name the row
+  /// if its blob turns out to be corrupt.
   ///
   /// Ghost frames have no indexed column — they exist only inside the `data`
   /// blob — so the alternative is loading and JSON-decoding *every* clip row in
--- a/mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
@@ -105,6 +105,45 @@ class ClipsDao extends DatabaseAccessor<AppDatabase> with _$ClipsDaoMixin {
     return query.get();
   }
 
+  /// Every clip row that carries a non-null `ghostFramePath`, projected to the
+  /// two fields a ghost-frame reference scan needs.
+  ///
+  /// Ghost frames have no indexed column — they exist only inside the `data`
+  /// blob — so the alternative is loading and JSON-decoding *every* clip row in
+  /// the database, which grows with the whole clip library. The `LIKE` still
+  /// scans the table, but SQLite discards the non-matching rows before they
+  /// cross into Dart, and clips carrying a ghost frame are a small minority of
+  /// a typical library.
+  ///
+  /// Trashed clips are included: a soft-deleted clip can still be restored, so
+  /// a ghost frame it references must survive a sharing draft's deletion.
+  ///
+  /// The predicate matches the encoding `DivineVideoClip.toJson` produces for
+  /// a set ghost frame. A clip whose `ghostFramePath` is null serializes to
+  /// `"ghostFramePath":null` and is correctly skipped. Because the filter
+  /// depends on that encoding, the app layer pins it end-to-end against the
+  /// real model in `draft_storage_service_test.dart`.
+  Future<List<({String id, String? draftId, String data})>>
+  getClipsWithGhostFrames() async {
+    final rows = await customSelect(
+      'SELECT id, draft_id, data FROM clips WHERE data LIKE ?',
+      variables: [Variable.withString('%$_ghostFramePathJsonPrefix%')],
+      readsFrom: {clips},
+    ).get();
+
+    return [
+      for (final row in rows)
+        (
+          id: row.read<String>('id'),
+          draftId: row.read<String?>('draft_id'),
+          data: row.read<String>('data'),
+        ),
+    ];
+  }
+
+  /// Serialized prefix of a set `ghostFramePath` inside a clip's `data` blob.
+  static const _ghostFramePathJsonPrefix = '"ghostFramePath":"';
+
   /// Update the order index of a clip
   Future<bool> updateOrderIndex({
     required String id,
--- a/mobile/lib/blocs/background_publish/background_publish_bloc.dart
+++ b/mobile/lib/blocs/background_publish/background_publish_bloc.dart
@@ -94,6 +94,9 @@ class BackgroundPublishBloc
         ),
       );
 
+      // Deliberately after the emit above: the video is already live, so the
+      // draft row and its unreferenced media are reclaimed off the publish
+      // critical path rather than in front of the success state (#6548).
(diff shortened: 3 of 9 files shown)


mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
```diff
@@ -105,6 +105,47 @@ class ClipsDao extends DatabaseAccessor<AppDatabase> with _$ClipsDaoMixin {
...
+  /// a typical library.
+  ///
+  /// Trashed clips are included: a soft-deleted clip can still be restored, so
+  /// a ghost frame it references must survive a sharing draft's deletion.
+  ///
+  /// The predicate matches the encoding `DivineVideoClip.toJson` produces for
+  /// a set ghost frame. A clip whose `ghostFramePath` is null serializes to
+  /// `"ghostFramePath":null` and is correctly skipped. Because the filter
+  /// depends on that encoding, the app layer pins it end-to-end against the
+  /// real model in `draft_storage_service_test.dart`.
```
realmeylisdev, line 127: this pin doesn't hold. I swapped the prefix to `"ghost_frame_path":"` so the LIKE matches nothing, and all five `clip ghost-frame file hygiene` tests in `draft_storage_service_test.dart` still passed. (dropped a canary throw in this method afterwards and it fired, so it wasn't a stale build.)

what's actually carrying them is `_jsonFilePathKeys` further down this file, which already lists `ghostFramePath`. when `deleteGhostFrameFiles` doesn't find a basename in the precomputed set it falls back to `deleteFileIfUnreferenced`, then `referencedFilenames`, which decodes every clip blob and finds the same reference. breaking the prefix *and* dropping `'ghostFramePath'` from that set does turn 3 of the 5 red, which is what pins it down.

so the real failure mode on a `toJson` drift isn't a deleted ghost frame, it's `getClipsWithGhostFrames()` quietly returning nothing and every frame falling through to a full `SELECT data FROM clips` plus decode, once per frame, which is slower than before this PR with nothing going red. `clips_dao_test.dart` hand-builds its blobs so it can't catch it either.

cheapest real guard is probably an app-side assert that `json.encode(clip.toJson())` contains the literal this filters on, but the constant is private here so where it should live is your call. fine either way, I'd just rather the comment (and the matching line in the PR description) not promise a net that isn't there.

mobile/lib/services/draft_storage_service.dart
```diff
@@ -703,14 +703,18 @@ class DraftStorageService {
...
+  /// table: this scan runs on every draft deletion, and decoding every blob in
+  /// the database made it grow with the size of the clip library (#6548).
   Future<Set<String>> _referencedGhostFrameFilenames({
     String? excludeDraftId,
   }) async {
-    final clipRows = await _clipsDao.getAllClips(includeTrashed: true);
+    final clipRows = await _clipsDao.getClipsWithGhostFrames();
     final filenames = <String>{};
 
     for (final row in clipRows) {
```
realmeylisdev, line 720: side note from chasing the above, pre-existing so not really for this PR: `excludeDraftId` is null on the `clearAllDrafts()` call, and library clips have a null `draft_id`, so `row.draftId == excludeDraftId` on the next line skips exactly the survivors that call is meant to protect. same experiment shows it: with the prefix correct but `'ghostFramePath'` removed from `_jsonFilePathKeys`, the only test that fails is "keeps a ghost frame referenced by a surviving library clip when all drafts are cleared", meaning the set never held that basename. identical code on origin/main so it isn't something you introduced.

`excludeDraftId != null && row.draftId == excludeDraftId` covers it, or a follow-up issue if you'd rather keep this branch focused.

realmeylisdev (approved):
lgtm, the ownership move checks out: both `publishVideo` call sites hand the future to `BackgroundPublishRequested`, so the service-side delete really was redundant, and pulling `_endForegroundSession` into the outer `finally` keeps the delete inside the keep-alive window.

one non-blocking thing: I broke the LIKE prefix so it matches nothing and all five ghost-frame tests still passed, so that "pinned end-to-end" line isn't true. repro is inline.

mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
realmeylisdev: pushed 97757474e for this rather than leaving it on you, since the fix is just finishing what the comment already says.

it adds a small `ghost-frame filter encoding` group in the same file that goes at the filter directly, through the real save path: one test that a real clip with a ghost frame comes back from `getClipsWithGhostFrames()`, one that a real clip without one does not. no new API surface, the private constant stays private.

mutation-checked both ways: point the prefix at `"ghost_frame_path":"` so it matches nothing and the first fails; drop the `:"` so it also matches `"ghostFramePath":null` and the second fails. that second direction is the one I would have missed by only asserting the positive case, since a filter that matches everything would keep every ghost file alive forever.

so this line is accurate as written now. only nit left is the PR description, which credits "the existing shared-ghost-frame tests" for the pin, and it is the new group doing it.

realmeylisdev:
heads up, I pushed one commit to your branch rather than leaving the finding sitting there: 97757474e, a test-only change that makes the "pinned end-to-end" line in the DAO doc actually true. details in the thread on clips_dao.dart. revert it freely if you would rather do it differently.

nothing else pushed. the `excludeDraftId` note on draft_storage_service.dart is pre-existing and untouched, your call whether it belongs here or in a follow-up.

verified before pushing: `dart format` and `flutter analyze lib test` clean, and draft_storage_service + video_publish + background_publish_bloc + clip_library_service + file_cleanup_service suites green (223). approval still stands.

mobile/lib/services/draft_storage_service.dart
realmeylisdev: filed this as #6581 so it does not disappear with the thread, and left your branch alone. it is out of scope here and the fix changes `clearAllDrafts` behaviour rather than anything on the publish path, so it did not belong in a perf PR.

worth saying it is latent, not live: the `deleteFileIfUnreferenced` fallback still keeps those files alive today, so nothing is being deleted. it turns real only if `_jsonFilePathKeys` ever stops covering `ghostFramePath`. resolving here, the issue carries the repro.

mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
realmeylisdev: all 15 checks green on 97757474e, including the four test shards that run this file. resolving.

Issue #6450, opened 2026-07-27
fix(account): successful account deletion throws GoError and shows no confirmation


realmeylisdev:
## Summary

A **successful** account deletion throws `GoError: There is nothing to pop`, which skips the success snackbar and the screen-reader announcement, and escapes to Crashlytics. The user's account is gone and the app tells them nothing.

Observed on a physical iPhone (iOS 26.5.2) running a real deletion end to end against production, on a `divineOAuth` account.

## Signature

```
GoError: There is nothing to pop
#0  GoRouterDelegate.pop (package:go_router/src/delegate.dart:105:7)
#1  GoRouter.pop (package:go_router/src/router.dart:586:20)
#2  GoRouterHelper.pop (package:go_router/src/misc/extensions.dart:73:65)
#3  executeAccountDeletion.dismissDialog (package:openvine/widgets/delete_account_dialog.dart:459:15)
#4  executeAccountDeletion (package:openvine/widgets/delete_account_dialog.dart:640:7)
<asynchronous suspension>
```

Same `GoError` signature as #6112, different call site — that one was a back/close button on a leaf screen (`_InkResponseState.handleTap`) and is closed. This one is deterministic rather than navigation-dependent: it fires on every successful deletion.

## Root cause

`executeAccountDeletion` awaits `authService.signOut(deleteKeys: true, deleteLocalUserData: true)`, which flips auth state to `unauthenticated`. The global router redirect then sends the app to `/welcome` and clears the stack — including the progress dialog route. `dismissDialog()` runs immediately after and calls `context.pop()` on a stack that no longer has anything to pop.

The function is `try { … } finally { … }` with no `catch`, so the throw propagates: the success snackbar and `announceOutcome()` on the lines directly below never execute, and the error reaches `runZonedGuarded` → Crashlytics.

Console from the observed run:

```
23:17:51.296  Router redirect: unauthenticated on /profile/… — redirecting to /welcome
23:17:51.510  signOut complete — keyStorageHasKeys=false, authSource=none
              GoError: There is nothing to pop
```

## Impact

- Every successful account deletion files a Crashlytics error.
- No success confirmation, visual or spoken — which reads to the user exactly like the failure #6335 was filed about.

## Suggested fix

[...]

PR #6563, opened 2026-08-01
fix(account): prove key control when deleting the server account

realmeylisdev:
## Description

Account deletion sends a **NIP-98 proof-of-key** to Keycast instead of relying on the bearer token.

`AuthService.deleteKeycastAccount()` calls `getSessionOrRefresh()` immediately before `DELETE /user/account`, and a refreshed access token no longer carries Keycast's first-party fact — so the request is refused. Because that call runs *after* the irreversible NIP-62 vanish and kind-5 sweep, the user is left with their content broadcast for deletion, their account still alive, and a message telling them to check their connection. Retrying reproduces it exactly.

A kind-27235 event signed by the account's own key, bound to the exact request URL and method, proves the request comes from the key holder regardless of how the session was obtained or how old it is.

Notes for review:

- `KeycastOAuth.deleteAccount` takes an optional `nip98Signer` and **passes it the URL it is about to call** rather than letting the caller rebuild it. NIP-98 binds the signature to the `u` tag, so any divergence between signed URL and requested URL makes the server reject every proof.
- **Falls back to the bearer token** whenever a proof cannot be produced. That keeps this safe to ship before the server accepts proofs, and means it cannot make anything worse for anyone.
- The readiness gate from #6417 is **deliberately untouched**. It should only be relaxed once the server side is confirmed live, or users regress straight back to the stranded state.
- The signer must be alive at this point, and is — the flow signs and publishes the kind-62 vanish moments earlier. That ordering is now load-bearing and is called out in the code.
- `Nip98AuthService` is built on first use and is nullable rather than `late final`, so `dispose()` does not construct one (starting its cache-cleanup timer) purely to tear it down.

**Related Issue:** Relates to #5756, Relates to #4881

Server-side counterpart: divinevideo/keycast#331 — that PR is what actually makes deletion succeed. This one is the client half and is inert until it ships.

## Out of Scope

- Relaxing the `checkAccountDeletionReadiness()` gate — intentionally deferred until keycast#331 is deployed.
- #6450 (successful deletion throws `GoError: There is nothing to pop` and shows no confirmation) — still live on main, separate fix.
- #6215 — already fixed on main; both publishes now await a relay OK.
- The shipped build emits kind-5 requests with no `a` tags for addressable kinds; fixed on main, unrelated to this diff.
- divinevideo/divine-funnelcake#746 — a vanish never erases media blobs. Independent workstream.

## Verification

- `flutter analyze lib test packages/keycast_flutter` — No issues found
- `flutter test packages/keycast_flutter/test/oauth_client_test.dart` — 116 pass (4 new)
- `flutter test test/services/auth_service_delete_keycast_account_test.dart` — 9 pass (1 new)
- `flutter test test/services/auth_service_*.dart` — 239 pass
- `flutter test test/widgets/delete_account_dialog_test.dart test/widgets/delete_account_confirmation_test.dart` — 53 pass
- `flutter test test/services/account_deletion_service_test.dart` — 25 pass
- Pre-commit and pre-push hooks passed (format, analyze, codegen, changed-file tests)

New coverage:

- proof is sent as `Nostr <token>` and takes precedence over the bearer token
- the signer is handed the exact URL the request goes to
- falls back to `Bearer` when the signer returns null, and when no signer is supplied
- `deleteKeycastAccount` actually supplies a signer — this is the one that fails if the fix regresses

Four existing stubs in `auth_service_delete_keycast_account_test.dart` needed the new named argument added; that was my diff's doing, not a pre-existing failure.

## Type of Change

### DIFF
--- a/mobile/lib/services/auth_service.dart
+++ b/mobile/lib/services/auth_service.dart
@@ -28,6 +28,7 @@ import 'package:openvine/services/auth/signer_secure_store.dart';
 import 'package:openvine/services/background_activity_manager.dart';
 import 'package:openvine/services/crash_reporting_service.dart';
 import 'package:openvine/services/nip07_service.dart';
+import 'package:openvine/services/nip98_auth_service.dart';
 import 'package:openvine/services/relay_discovery_service.dart';
 import 'package:openvine/services/user_data_cleanup_service.dart';
 import 'package:openvine/utils/divine_login_banner_dismissal.dart';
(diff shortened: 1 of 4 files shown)


rabble (approved):
Approved. The NIP-98 proof binds to the exact DELETE URL, and the focused auth/OAuth tests pass.

PR #6566, opened 2026-08-01
fix(l10n): defer untranslated ms/ur/vi/zh keys to unblock main

### LINKED ISSUE #6567: fix(l10n): defer untranslated ms/ur/vi/zh keys to unblock main
Auto-created from PR #6566.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6566
**Branch:** `fix/l10n-new-locale-debt`

rabble:
Closes #6567

## Description

`origin/main` is red on `test/l10n/arb_consistency_test.dart` ("all locales define the same message keys as app_en.arb"): the ms/ur/vi/zh locales added in #6495 never received 34 keys that landed on main afterward (settings upload progress, publish errors, video editor chroma-key copy). The generated localizations for those four locales are stale for the same reason.

This blocks every push whose pre-push hook runs the l10n test and fails the general-suite CI on all open PRs.

- Adds the 34 keys to `_knownUntranslatedDebt` with a comment referencing #6495 (translations deferred to the next l10n pass, the sanctioned deferral per AGENTS.md)
- Regenerates the four stale generated localization classes

## Verification

- `flutter test test/l10n/arb_consistency_test.dart` — 6/6 pass (was failing on a clean `origin/main` checkout)
- `dart format` clean

## Type of Change


hm21 (approved):
Diagnosis is right, the remedy wasn't. Pushed a fix in `ace7f7ae1f`.

These 34 keys are already translated in all 13 other locales — it's a four-locale lag, not the systematic backlog `_knownUntranslatedDebt` is meant for. Deferring shipped English green-screen and account-switch copy to ms/ur/vi/zh users, and since the debt set is global it would also have stopped CI from catching it if one of the 13 complete locales dropped one of these keys later.

So: translated all 34 into ms/ur/vi/zh, reverted the debt-list additions, regenerated. Terminology follows what each locale already uses (ms `klip`/`Pustaka`/`Gunakan`, ur `کلپ`/`لائبریری`, vi `clip`/`Thư viện`, zh `片段`/`作品库`). main is unblocked either way, this way without new debt. Net diff is now purely additive — `arb_consistency_test.dart` is untouched.

Verified: `flutter test test/l10n/` 89/89, `test/providers/video_publish_provider_test.dart` 26/26, `flutter analyze lib test` clean, format clean, full CI green on the new head.

One thing left for you: the PR description still describes the deferral approach, and it becomes the squash commit body — worth updating before merge.

hm21:
Merging this ourselves rather than leaving it to the author, since `main` is red and this blocks every open PR: `Mobile CI` fails on `main` at `5a12c0c63` with `Generated Files` and `Tests (shard 3/4)`, which is exactly what this fixes. Approved and green, so waiting only extends the outage.

One note for the record: the description and title do not match the diff. Both say the keys are deferred to `_knownUntranslatedDebt`, but the PR does not touch that list — it ships real translations into the four ARBs (`"publishErrorAccountChanged": "这个视频属于另一个账号。切回那个账号才能发布。"`). That is better than advertised and matches our rule against English fallbacks, so the squash subject is adjusted to describe what actually landed.

PR #6569, opened 2026-08-01
feat(db): name the bootstrap failure and offer a guarded local reset

### LINKED ISSUE #6580: feat(db): name the bootstrap failure and offer a guarded local reset
Auto-created from PR #6569.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6569
**Branch:** `fix/db-bootstrap-secure-storage-diagnosis`

hm21:
Closes #6580

## Description

A user reported the "couldn't unlock your local database" fail-closed screen showing `db-bootstrap-failed`, with one button on it: close Divine. Two problems, plus a review pass that made the fix for the second one safe to complete.

### 1. The diagnosis was unreachable for a whole class of failures

`databaseBootstrapDiagnosticCode` detected the secure-storage case by looking for the literal `secure storage` in the error message. Nothing in the codebase produces that text: `flutter_secure_storage` raises a `PlatformException` carrying a platform status code and no prose (iOS `Code: -25308, User interaction is not allowed`). So every locked-keychain launch was filed under the `db-bootstrap-failed` catch-all — and that is exactly the case the screen's own copy, "Restart Divine after unlocking your device", is written for. The one variant a user can fix in ten seconds was indistinguishable from the ones they can't.

The bootstrap's keystore access is now wrapped in a typed `DatabaseCipherStorageUnavailableException` and classification matches on the type. The original stack is preserved with `Error.throwWithStackTrace`, so Crashlytics grouping is unchanged.

The repair allowlist now excludes that type explicitly. Repair deletes the cipher key, and behind a locked keychain the database is intact — clearing it would turn a restart-and-retry into permanent loss. That was already the behaviour, but only by accident (no marker matched); now it can't regress when the allowlist widens.

Also drops the `SQLCipher is not linked` check — a 1.0.15 message no code has produced since the SQLite3MultipleCiphers migration in #5331.

Note the guarantee is platform-scoped, and the docs now say so: only the Keychain backends surface the failure as a throw. Android's `encryptedSharedPreferences` backend falls back to plaintext preferences when the keystore cannot be initialised and reads back `null`, which the bootstrap cannot tell apart from a fresh install. Closing that gap needs a written-a-key-before marker and belongs with the accessibility migration below.

### 2. A stuck user had no way out except reinstalling

Reinstalling is the worst available option on Android: the signing key lives in app data and `backup_rules.xml` excludes it from cloud backup and device transfer, so the account is gone unless the nsec was written down. An in-app reset clears the database but *not* the signing key — the same recovery, costing local-only drafts and clips instead of the account.

It is gated on the diagnosis rather than offered unconditionally:

| Diagnosis | Reset offered | Why |
|---|---|---|
| `db-secure-storage` | no | Database is intact, only its keystore was unreachable. Resetting would delete data a restart recovers for free. |
| `db-cipher-unavailable` | no | No cipher available, so a fresh database could not be created any more than the old one could be opened. |
| `db-cipher-mismatch` | yes | The key already doesn't open the file; nothing left to lose. |
| `db-bootstrap-failed` | yes, confirmed | Cause unknown, database presumed unusable. The cipher key is kept in case that presumption is wrong — see below. |

That table is why the commits belong together: the gate is only safe because the first commit made the diagnosis trustworthy. The diagnosis is now an enum and the gate an exhaustive switch, so a future code has to make this choice deliberately instead of defaulting into a destructive offer.

The reset sits below "close Divine" so the non-destructive action stays primary, and is behind a confirmation naming what is lost.

Naming note: the new exception ends in `Exception` while its sibling `DatabaseCipherUnavailableError` ends in `Error`. Deliberate — the sibling is a build misconfiguration (a genuine `StateError`, reportable per `.claude/rules/error_handling.md`), this one is an expected environmental condition that must not be treated as a programming invariant violation.

### 3. The reset was safe to offer, but not yet safe to complete

A review pass over the two commits above found three ways the flow broke *after* the user confirmed. All three are in the third commit.

**A successful reset had nowhere to land.** `SystemNavigator.pop` cannot end the process on iOS, and the success path never cleared the loading flag, so the confirmation sat there spinning behind two disabled buttons with nothing left to happen — on the very platform the locked-Keychain diagnosis is written for, and on the one flow that exists to unstick a stuck user. The screen now moves to a terminal step that says to close and reopen Divine, and still closes the app from there so Android finishes the activity exactly as before.

**The reset deleted the DB cipher key.** For `db-bootstrap-failed` the cause is unknown and the key may be perfectly valid, so deleting it made the backup the reset had just created permanently unreadable — the same invariant the bootstrap defends when it refuses to rotate a still-valid key before a wipe. `resetEncryptedDatabaseCache` now takes `deleteCipherKey`, defaulting to the old behaviour so the automatic repair is untouched, and the manual path keeps the key. A fresh database opens just as well under a retained key. Keeping it also drops the reset's last keystore dependency — which may itself be why the user is on this screen — so the "a reset that throws keeps the screen and says so" path is now a genuine edge rather than the likely one.

**A failing reset was swallowed whole.** The catch did not even bind the error, so the only report of a failed recovery was one sentence on screen. It now records to Crashlytics before rethrowing.

Smaller fixes from the same pass: the failure banner is cleared when the confirmation is re-entered rather than greeting the user with a previous attempt; each step swap and the failure are announced to screen readers, since nothing is pushed and the focused node just disappears; the reset times out so a wedged platform channel cannot hang it forever; the screen stops advising a restart on the diagnoses where restarting cannot help, which previously sat directly above the reset button; and one closure is shared between the automatic repair and the manual reset so the two cannot drift apart.

**Related Issue:** 

## Out of Scope

**The root cause of the keychain variant.** The DB cipher key is written with the `flutter_secure_storage` default `KeychainAccessibility.unlocked`, while the nsec uses `first_unlock` (`packages/nostr_key_manager/lib/src/platform_secure_storage.dart`). The key is therefore unreadable whenever the device is locked, including a background launch before the first unlock after boot — which matches the `processState: BACKGROUND` on the historical Crashlytics events. Aligning it needs the legacy-read-and-rewrite migration that `platform_secure_storage.dart` already implements, because an existing keychain item keeps its accessibility when overwritten. Own PR: a mistake there costs the key, and with it the database.

**Android's silent keystore fallback.** Detecting it needs a persisted "a key was written once" marker so a `null` read can be told apart from a fresh install. That is a storage-format change with its own migration, not a docs fix.

**Screen copy stays hardcoded English.** This screen builds its own `MaterialApp` without localization delegates, since it runs after startup has already bailed out. The confirmation, terminal, and diagnosis-dependent steps follow the existing convention on this screen rather than introducing a partial l10n setup here.

## Verification

- `dart format` on all changed files — clean
- All four design-system ceiling ratchets (`raw_textstyle`, `raw_colors`, `material_button`, `raw_dialog`) — OK. The screen's existing inline styles were extracted into shared constants and reused by every step, so the raw-`TextStyle` count stays at its baseline of 4 rather than growing.
- `flutter test test/services/database_encryption_bootstrap_test.dart test/startup/database_bootstrap_failure_app_test.dart test/main_background_launch_downloads_test.dart test/main_video_cache_startup_test.dart` — 63/63 passing
- Manual verification on a real Samsung Galaxy — pending, this is a draft
- Screenshots of the reset flow — pending

The load-bearing test on the fix is in the bootstrap suite: when the keystore read throws, `resolveCipherKey` must raise the typed exception *and* write no replacement key. A failed read must never be read as "no key stored", or the bootstrap overwrites the only key that opens the existing database. The write half of that contract is pinned too — a regression there would re-open the original misclassification and, through it, flip the reset gate.

The load-bearing test on the feature is `offers no reset for a locked keystore` — the case where showing the button would cause the data loss it appears to prevent. `lands on a terminal step when the app cannot close itself` covers the iOS failure above: it drives the reset with an `onCloseApp` that does not terminate, which is what iOS actually does.

Existing tests for this path asserted against hand-built `StateError('secure storage unavailable')` values — a string no production code emits, which is why the dead branch looked covered. Those now use the real type.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -1419,6 +1419,15 @@ Future<void> _startOpenVineApp() async {
     ),
     shouldRepairLocalDatabaseCache:
         shouldRepairLocalDatabaseCacheAfterBootstrapError,
+    // Manual escape hatch behind a confirmation, offered only for diagnoses
+    // where the database is already unusable. It clears the DB and its cipher
+    // key but not the signing key, so it costs local-only drafts and clips
+    // while an uninstall — the alternative a stuck user is left with — takes
+    // the whole Android keystore and the account with it.
+    resetLocalDatabase: () => resetEncryptedDatabaseCache(
+      secureStorage: dbCipherSecureStorage,
+      onDatabaseReset: () => DmSyncState(sharedPreferences).clearAll(),
+    ),
     runApp: runApp,
     removeNativeSplash: FlutterNativeSplash.remove,
   );
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -1382,6 +1382,20 @@ Future<void> _startOpenVineApp() async {
     ),
   );
 
+  // Both the automatic repair and the manual escape hatch below mean the same
+  // operation, so they share one definition and cannot drift apart. They differ
+  // only in whether the cipher key goes with the database.
+  Future<void> resetLocalDatabaseCache({required bool deleteCipherKey}) =>
+      resetEncryptedDatabaseCache(
+        secureStorage: dbCipherSecureStorage,
+        deleteCipherKey: deleteCipherKey,
+        // On the recreate the Drift DB is wiped but SharedPreferences survives;
+        // clear the DM sync state so the next inbox open runs a full re-drain
+        // instead of skipping it as "already complete" (which had left
+        // recovered chats stranded under "Message requests"). See #5304.
+        onDatabaseReset: () => DmSyncState(sharedPreferences).clearAll(),
+      );
+
   final dbCipherKeyResult = await resolveDatabaseBootstrapForAppStart(
     resolveCipherKey: () => resolveStartupDatabaseCipherKey(
       // resetOnError MUST stay false here: the cipher key is the one secret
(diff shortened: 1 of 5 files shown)


rabble (approved):
Reviewed current head 019b55b73e9390e7a10d92a26469654aa96b763d. The startup failure classification, guarded manual reset flow, key-retention behavior, and reset failure/timeout paths are covered by the added tests. GitHub Analyze, Format, Generated Files, all four test shards, preview build, and mergeability checks are green; git diff --check is clean.

Issue #105, opened 2025-06-19
🤖 Daily Agent Standup - 2025-06-19


rabble:
# Multi-Agent Daily Standup

**Date:** 2025-06-19

### Template for Agent Updates:
```
[AgentName] STATUS: [what you're currently working on]
[AgentName] YESTERDAY: [what was completed yesterday] 
[AgentName] TODAY: [goals for today]
[AgentName] BLOCKERS: [any impediments or help needed]
[AgentName] FILES: [files currently claimed/locked]
```

## Coordination Notes
- Integration points requiring attention
- Shared interface changes
- Dependency updates
- Conflict resolution needs

## Daily Metrics
- [ ] All agents checked in
- [ ] No blocking dependencies
- [ ] All file locks current
- [ ] Integration tests passing

---
**Instructions for Agents:**
1. Post your status update using the template above
2. Tag other agents if coordination needed: @agent-name
3. Update your file locks in comments
4. Resolve any blockers or escalate

**File Lock Format:**
🔒 [AgentName] CLAIMING: path/to/files (ETA: X hours)
🔓 [AgentName] RELEASING: path/to/files (work complete)

rabble:
[Claude] STATUS: Multi-agent coordination infrastructure setup complete
[Claude] YESTERDAY: N/A - initial setup
[Claude] TODAY: Implemented full coordination framework for NostrVine
[Claude] BLOCKERS: None - all coordination infrastructure ready
[Claude] FILES: 🔓 Released all coordination setup files

## Completed Infrastructure:
✅ **CODEOWNERS** - Module ownership defined for all agents
✅ **Issue Templates** - Feature, bug, and test templates created  
✅ **PR Template** - Multi-agent coordination checklist
✅ **GitHub Actions** - Automated quality gates pipeline
✅ **Labels** - Agent-specific and coordination labels
✅ **Parking Lot** - Independent tasks for blocked agents (#106)

## Ready for Multi-Agent Development:
- Agents can now claim file ownership using file lock comments
- Issue templates enforce proper coordination fields
- Automated quality gates ensure code standards
- Daily standup process established (this issue)
- Backup tasks available when blocked

🚀 **Multi-agent development framework is live and ready for use!**

rabble:
# [Claude] FINAL STATUS: Multi-Agent Coordination System Complete ✅

## Infrastructure Delivered
🏗️ **Complete multi-agent development framework operational**

### Core Coordination Components
[...]

Issue #104, opened 2025-06-19
📋 TDD VIDEO SYSTEM REBUILD - Master Epic


rabble:
## 🎯 MASTER EPIC: Complete TDD Video System Rebuild

This epic tracks the complete rebuilding of the NostrVine video system using Test-Driven Development to eliminate critical bugs and performance issues.

## 🚨 Current Problems Being Solved
- **Dual Video List Problem** → Single source of truth (#101)
- **3GB Memory Usage** → <500MB target (#102)  
- **5 Critical Race Conditions** → Zero race conditions (#103)
- **Index Mismatch Bugs** → Perfect synchronization
- **Poor Performance** → TikTok-smooth experience

### 🧪 Week 1: Test Foundation (Issues #81-84)
- [x] #81 - Project Setup & Test Infrastructure  
- [ ] #82 - Core Behavior Test Specification
- [ ] #83 - Integration Tests
- [ ] #84 - UI Tests

### 🔧 Week 2: Core Models (Issues #85-88)
- [ ] #85 - VideoState Model
- [ ] #86 - VideoManager Interface  
- [ ] #87 - Mock VideoManager
- [ ] #88 - Video Event Processing

### ⚙️ Week 3: VideoManager Implementation (Issues #89-92)
- [ ] #89 - Core VideoManager Service
- [ ] #90 - Video Preloading
- [ ] #91 - Memory Management
- [ ] #92 - Error Handling & Recovery

### 🎨 Week 4: UI Integration (Issues #93-96)
- [ ] #93 - VideoFeedProvider Rebuild
- [ ] #94 - VideoFeedItem Rebuild
- [ ] #95 - FeedScreen Integration
- [ ] #96 - Integration Testing

### 🚀 Week 5: System Swap (Issues #97-100)
- [ ] #97 - Parallel Testing
- [ ] #98 - Feature Flag Deployment
- [ ] #99 - Full Deployment
- [ ] #100 - Old System Removal

## 🎯 SUCCESS METRICS
- **Test Coverage**: >90% on all new code
- **Memory Usage**: <500MB peak (90% reduction)
- **Video Load Success**: >95% (up from ~80%)
- **Index Mismatch Bugs**: 0 detected
- **Race Conditions**: 0 detected
- **Crash Rate**: <0.1%

## 🤖 PARALLEL EXECUTION
This work is designed for multiple agents working simultaneously:
- Each issue is self-contained with clear prerequisites
[...]

PR #6574, opened 2026-08-01
fix(invites): show granted invites and keep the generate action honest

### LINKED ISSUE #6510: fix: expose invite sharing reliably to existing identities
## Summary

Ensure every authenticated Divine account can retrieve and display its invite allocation, including imported nsec, Keycast, bunker, Amber, and other existing-identity accounts.

This is the mobile half of the backend's lazy enrollment behavior: once authentication is ready, the app requests `GET /v1/invite-status`; the backend can then grant an existing identity its one-time initial allocation and return it immediately.

## Current behavior

`InviteStatusCubit` is shared globally and is not explicitly restricted to Keycast accounts, but several behaviors can prevent existing users from seeing invites:

- `load()` returns without making a request when NIP-98 authentication is not ready.
- The cubit does not automatically retry when authentication later becomes ready.
- A `401` restores the previous state without surfacing an error or scheduling a retry.
- The Settings invite entry is hidden until status reports existing invite activity.

As a result, a user can have no visible invite entry even though the backend is ready to enroll them, or can continue seeing stale/empty status after an authentication timing gap.

## Desired behavior

- Trigger invite-status after the active account's NIP-98 signer becomes ready.
- Apply this consistently to every supported authenticated account type.
- Retry expected authentication-timing failures once signing becomes available.
- Refresh when the active account changes and prevent status leaking between accounts.
- Display the invite entry as soon as lazy enrollment returns a positive allocation.
- Distinguish a real zero allocation from authentication, network, and service failures.
- Keep the existing code generation, copy, and share experience.

This should happen for returning users without requiring a bulk backend migration. The first successful authenticated status request acts as their lazy backfill.

## Acceptance criteria

- [ ] Invite status loads after authentication becomes ready rather than being permanently skipped.
- [ ] Imported nsec and Keycast accounts both trigger authenticated invite status.
- [ ] Bunker, Amber, and other supported signers follow the same path when they can produce NIP-98 signatures.
- [ ] Switching accounts clears stale invite state and loads the new account's status.
- [ ] A temporary auth gap or `401` is retried after the signer becomes ready.
- [ ] Network and server failures produce a retryable error state rather than appearing as zero invites.
- [ ] A newly allocated lazy-backfill response makes invite sharing visible immediately.
- [ ] Existing generate, copy, and share actions continue to work.
- [ ] Tests cover delayed auth readiness, `401` recovery, account switching, imported nsec, and Keycast flows.
- [ ] Invite documentation no longer says existing identities must never call invite-status.

## Related work

- Backend universal initial allocation: https://github.com/divinevideo/divine-invite-darshan/issues/103
- Invite refill policy: https://github.com/divinevideo/divine-invite-darshan/issues/104

dcadenas:
## Summary

The invites screen showed nothing to people who had invites waiting, and later offered a Generate button that the server refused.
Two separate causes: the app decided it could sign when it could not, and the button counted the wrong number.
Now the screen loads once signing is genuinely possible, and the button counts codes you can still mint rather than invites you can still give away.

This also fixes a cross-account NIP-98 cache key bug where a cached invite-status token for one account could be reused after switching accounts.

Closes #6510.

Depends on divinevideo/divine-invite-darshan#105 for the new `remainingToGenerate` field. Safe to merge in either order — see Risks.

## What Changed

The first bug was a wrong predicate, not a race. Invite status was gated on `Nip98AuthService.canCreateTokens`, which is just `isAuthenticated`. A `PubkeyOnlyNostrIdentity` is authenticated but cannot produce a NIP-98 signature, so the request went out unsigned-capable, failed, and the screen had nothing to show.

The second bug is that two different numbers were being conflated. Someone who minted all ten codes but had none redeemed still has ten invites to share and nothing left to generate. The old screen read the share number and kept offering a button the server answers with `429 invite_generation_limit_reached`.

The third fix is auth isolation. `Nip98AuthService` now includes the current pubkey in the token cache key, so account B cannot reuse account A's signed token for the same URL/method/payload after an account switch.

- Invite status now loads on `canPublishNostrWritesNow` and reloads when the auth session changes, instead of assuming signing is available.
- Invite state is account-scoped, with a generation counter that discards in-flight responses across an account switch, so status cannot leak between accounts.
- Waiting for signer readiness is bounded; if the signer never arrives, the screen falls through to the retryable error state instead of spinning forever.
- The generate card counts `remainingToGenerate`, falling back to `remaining` on a server that predates the field.
- The card moved above the code list, and the list became a `SliverList.builder`, so a full allocation no longer pushes the button off-screen.
- The card disappears entirely once nothing is mintable. Existing codes stay listed and shareable.
- Invite-status retries transient server failures with a small default backoff when the server omits `Retry-After`.

## Testing

Covered by unit, widget and integration tests, and verified by hand on an emulator against a local stack built from the backend branch.

- 13 `invites_screen` tests, 53 `invite_api_client` tests, `flutter analyze` and `dart format` clean. 14,090 mobile tests passed on the first commit of this branch.
- Integration coverage for both identity kinds: `repro_log1_test.dart` asserts the allocation for a locally generated key, and `multi_account_switch_test.dart` now asserts it for a Keycast email account. The latter passes — `Phase 1b: Account A invite allocation — remaining=10 mintable=10 codes=0`.
- On device: settings badge showed `Invites 10`, the count walked 10 to 0 across ten mints, the card vanished at zero with all ten codes still listed, no `429`, and the list scrolled correctly in both directions.
- Takeover follow-up verification: 63 focused Flutter invite/settings tests and 45 invite API client tests pass after addressing review findings.

Two things I am not claiming:

- `multi_account_switch_test.dart` still exits non-zero overall, at Phase 2, for a reason unrelated to this PR. `navigateToCreateAccount` looks for the fresh-install button label, but after a sign-out the welcome screen renders the returning-user layout with a shorter label. Those files are byte-identical to `main`, and the same failure reproduces with this branch's changes stashed. Fix is on its way separately.
- These Patrol tests do not run in CI. The only integration workflow is `workflow_dispatch` and scoped to `integration_test/e2e/ --tags service`.

## Risks

Low. The client change is backward compatible in both directions.

- `remainingToGenerate` is nullable with a fallback to `remaining`, so merging this ahead of the backend preserves today's behaviour rather than breaking. The fix simply has no effect until the backend ships.
- Shipping the backend first is the order where this takes effect, but neither blocks the other.
- `local_stack` pins `ghcr.io/divinevideo/divine-invite-darshan:e2e`, which is only republished on push to that repo's `main`. Until #105 merges, running the new invite assertion locally needs a locally built image; the command is in a comment at the assertion.

## Follow-ups

- #6604: `BunkerNostrIdentity` reports signing readiness unconditionally, regardless of relay connectivity, and a mid-session bunker reconnect emits nothing on the auth stream. A bunker user could therefore hit a `401` with no automatic retry after reconnect.
- #6605: The claimed-code list is unpaginated. The backend supports `code_limit`/`code_offset` and the app never sends them.


rabble:
Reviewed the whole invite series (this, #6575, #6576, and darshan#105) and pushed two fixes here in `2a5d9338c`. Everything else in the PR reads correctly to me — the `canCreateTokens`-vs-`canPublishNostrWritesNow` diagnosis is right, and the generation-counter tests (A-B-A churn, account switch clearing the previous allocation) are the kind that actually fail when the guard is removed.

**1. The client now retries the 503 the backend added for it.** This is the one worth a look. darshan#105 introduced `503 enrollment_busy` + `Retry-After: 1` *specifically* because `GET /v1/invite-status` enrolls lazily on read, so two status calls racing at launch make one lose the claim. This client fires that request from four call sites — `start()`, settings init, notifications page, lifecycle resume — plus every auth-session event, so the race is routine rather than exotic. The retry loop you added covered a rejected NIP-98 token but not the transient failure, so the exact case the backend paid to handle gracefully still dumped the user on an error screen with a Retry button.

The two reasons are now independent — hitting one doesn't spend the other — and the transient path honours the server's delay through the `_isRetryable` / `_retryDelay` helpers that were already there. Three tests, mutation-checked: removing the transient branch reds two of them and leaves the third (`refreshes auth and still retries...`) asserting `forceRefreshValues == [false, true, false]`, which pins that a 503 does not re-mint the token. Your follow-up note called this out; it seemed cheap enough to just do.

**2. `availableInviteCount`'s doc still said "can still generate"** while returning `remaining`. The *value* is right for the settings badge and notifications banner — someone who minted everything still has all of it to hand out — so only the comment changed. Worth fixing because that getter is the last place in the file where the two numbers read as interchangeable, which is the whole thing this PR removes.

Two notes rather than changes:

- **`_createCacheKey` now includes the pubkey.** Before this, the NIP-98 token cache was keyed on `url|method|payload`, so after an account switch account B's request to the same URL reused account A's signed token and authenticated as A. That's a cross-account auth leak with a proper regression test, and it's currently a supporting bullet in the description. It deserves to be called out as its own fix — it's arguably the most important thing in the PR.
- **401 now emits `error` where it used to restore the previous state.** For a Keycast/bunker identity mid-warmup that trades a spinner for an error card with no auto-retry, and your own follow-up notes a mid-session bunker reconnect emits nothing on the auth stream — so the user can land there and only escape by tapping Retry. Not a regression and you say so plainly, but #6510's "a temporary auth gap or 401 is retried after the signer becomes ready" isn't met. Worth a tracked issue rather than scope creep here.

Validation on my two commits: 56 `invite_api_client` tests, 63 across `test/blocs/invite_status`, `invites_screen_test`, and `settings_screen_test`, `flutter analyze lib test integration_test` clean. No rebase — GitHub reports no conflicts and this is review feedback.

Separately, on darshan#105 (merged, so follow-ups): `apply_initial_allocation` overwrites the caller's `source` with `InviteConsumption` whenever `joined_at` is set, so legacy invited users picking up the retroactive +10 through status get provenance that contradicts the log line right above it — and #104's "grant source is observable" is exactly what that defeats. And the "flag off ⇒ existing accounts untouched" claim doesn't quite hold: the consume replay branch and `needs_user_state_healing` (now `initial_allocation.is_none()`) both apply the additive grant with no `initial_allocation_enabled` check, so a legacy user replaying an old consume gets +10. Low practical exposure, but the risk section states it as an invariant.

NotThatKindOfDrLiz (approved):
Final pass is clean. The takeover commit addresses the remaining review blockers: no global emit-after-close suppression, no redundant auth-session adapter, bounded signer-wait recovery, linked follow-up issues, and corrected invite-status retry/backoff behavior. CI is green on e67887738.

NotThatKindOfDrLiz:
@dcadenas I took over the remaining review cleanup and pushed e67887738.

What changed:
- removed the global E2E suppression for "Cannot emit new states after calling close"
- removed the redundant InviteStatusAuthSessionSource adapter and wired InviteStatusCubit directly from AuthService in main.dart
- bounded waitingForAuth so a signer that never arrives falls through to the retryable error state, while later signer readiness still auto-loads status
- added default backoff for retryable invite-status failures when Retry-After is omitted and corrected the retry docs
- filed linked follow-ups for bunker signer readiness (#6604) and claimed-code pagination (#6605)
- updated the PR body to call out the NIP-98 cross-account cache-key fix and link the follow-ups

Verification passed locally and in CI:
- flutter test test/blocs/invite_status/invite_status_cubit_test.dart test/screens/settings/invites_screen_test.dart test/widgets/settings_screen_test.dart
- flutter test test/src/invite_api_client_test.dart from mobile/packages/invite_api_client
- flutter analyze lib test from mobile/packages/invite_api_client
- GitHub checks are green on e67887738

PR #6575, opened 2026-08-01
fix(auth): stop anonymous signup emitting after close

### LINKED ISSUE #6607: fix(auth): stop anonymous signup emitting after close
## Problem

Two independent unhandled async errors were found while exercising the auth flows on a real Android emulator.

1. `DivineAuthCubit` awaited auth calls that can flip `AuthService` to `authenticated` before returning. That auth-state flip refreshes the router, leaves the auth screen, and closes the cubit. Any post-await `emit` then throws `Bad state: Cannot emit new states after calling close`; in `submit()` the outer recovery emit can throw again and escape because call sites are fire-and-forget.
2. `KeycastOAuth.logout()` wrapped an `unawaited(...)` logout POST in `try`/`catch`, but async rejections after `logout()` returns are not caught by that surrounding block.

Review also found a follow-on behavior gap: once the auth cubit correctly suppresses `DivineAuthSuccess` after close, the login screen can no longer use that state to trigger `TextInput.finishAutofillContext()`.

## Expected

Auth screens should not emit after their cubit is closed, best-effort logout failures should not surface as unhandled async errors, and successful credential sign-in should still commit the autofill context before the redirect disposes the auth screen.

## Fix

PR #6575 guards post-await auth emits, uses `Future.ignore()` for the best-effort keycast logout request, and moves the autofill commit to the auth-state stream used by the router redirect.

dcadenas:
Closes #6607

## Summary

Two unrelated emit/rejection bugs on the auth screens, plus the follow-on fixes from review. Both original bugs were observed firing on a real Android emulator, not inferred from reading the code.

### 1. Emit-after-close on the auth cubit

`DivineAuthCubit` awaits an auth call that flips `AuthService` to `authenticated` before it returns. That flip refreshes the router, which leaves the auth screen and closes the cubit — so the emit after the await throws `Bad state: Cannot emit new states after calling close`. Worse, `submit()`'s outer catch is the last-resort handler: when `_handleSignUp`'s email-verification emit throws into it, the catch's own recovery emit throws again with nothing left to catch it, and since `submit()` is fire-and-forget at all three call sites that second throw escapes as an unhandled async error.

`isClosed` guards now cover every emit reachable after an await: `skipWithAnonymousAccount`, `_handleSignIn`, `_handleSignUp`, both their failure paths, the email-verification emit, and `submit()`'s outer catch.

### 2. Unhandled rejection from the keycast logout

Wrapping `unawaited()` in a `try`/`catch` catches nothing — the rejection arrives after the wrapping call has already returned, so the handler has to be attached to the future itself. Replaced with `Future.ignore()`.

This one lives on the **sign-out** path, not anonymous signup: `logout()` has a single caller, `auth_service.dart:1939`, reachable only from `signOut()` and `_completeDestructiveSignOutAfterDeletedKeys()`. The two fixes share a branch, not a flow.

### 3. Autofill no longer rides on `DivineAuthSuccess`

Consequence of fix 1, raised in review: with the guard in place `DivineAuthSuccess` is never emitted when the redirect wins the race, so `login_options_screen`'s listener silently skipped `TextInput.finishAutofillContext()` and the user got no password-manager save prompt. The commit now hangs off `AuthService.authStateStream` — the same signal `RouterRefreshListenable` uses to drive that redirect, delivered before the frame that disposes the subtree, so the race cannot skip it. Gated on a submitted credential form so Amber / NIP-07 sign-ins from the same screen don't commit an empty context.

### 4. Doc correction on the keycast logout endpoint

The old comment implied a server-side logout. keycast's handler takes no extractors and only returns a cookie-clearing header, and the client sends it unauthenticated after the tokens are already deleted — it revokes nothing. The call is kept in case the endpoint grows real revocation; the comment no longer claims it already has.

## Testing

Every guard is pinned by a test that was mutation-checked to fail when that guard is removed. Emit-after-close is swallowed by the generic catch, so assertions on completion alone stay green either way — the tests assert through the bloc observer, which is the only place the difference is visible.

- 68 `DivineAuthCubit` tests, 236 `keycast_flutter` tests, 190 `test/screens/auth` + `test/blocs/divine_auth` tests.
- `flutter analyze lib test integration_test` clean; package-level analyze clean.
- The three new autofill tests fail against the pre-fix listener and pass with it.

## Risks

Low. Narrow guards that remove error paths without altering successful behaviour, plus one signal moved to a strictly more reliable source.

## Review history

This branch carries reviewer commits on top of the original two fixes, per `PR_REVIEW.md`:

- **@hm21** — `80ba816258` simplified the fire-and-forget logout to `Future.ignore()`; `8ec6f37b16` applied the same `isClosed` guards to `_exchangeCodeAndLogin`.
- **@mbradley** — `30f4c9db5` guarded the five remaining emit-after-close sites (the substantive one); `182d00978` made the sign-in guard detectable via the bloc observer, after finding the earlier test stayed green without the guard; `279514df1` dropped a test subsumed by its observer sibling; `bf304acdd` / `7d41fc207` corrected comments.
- **@rabble** — `4e70e0d6b` resolved the autofill question above; `9a37f6b49` corrected the logout doc comment.

### DIFF
--- a/mobile/lib/blocs/divine_auth/divine_auth_cubit.dart
+++ b/mobile/lib/blocs/divine_auth/divine_auth_cubit.dart
@@ -359,6 +359,11 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
         category: LogCategory.auth,
       );
 
+      // Same close-mid-flight race as skipWithAnonymousAccount: the sign-in
+      // above flips AuthService to `authenticated` before returning, which
+      // can reroute off the auth screen and close this cubit.
+      if (isClosed) return;
+
       emit(const DivineAuthSuccess());
     } on InviteApiException catch (e, stackTrace) {
       await _authService.clearPendingDivineOAuthSession();
@@ -372,7 +377,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       addError(e, stackTrace);
 
       final current = state;
-      if (current is DivineAuthFormState) {
+      if (!isClosed && current is DivineAuthFormState) {
         emit(
           current.copyWith(
             isSubmitting: false,
@@ -393,7 +398,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       addError(e, stackTrace);
 
       final current = state;
-      if (current is DivineAuthFormState) {
+      if (!isClosed && current is DivineAuthFormState) {
         emit(current.copyWith(isSubmitting: false, generalError: e.message));
       }
     } catch (e, stackTrace) {
@@ -407,7 +412,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       addError(e, stackTrace);
 
       final current = state;
-      if (current is DivineAuthFormState) {
+      if (!isClosed && current is DivineAuthFormState) {
         emit(
           current.copyWith(
             isSubmitting: false,
--- a/mobile/lib/blocs/divine_auth/divine_auth_cubit.dart
+++ b/mobile/lib/blocs/divine_auth/divine_auth_cubit.dart
@@ -194,7 +194,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       addError(e, stackTrace);
 
       final currentState = state;
-      if (currentState is DivineAuthFormState) {
+      if (!isClosed && currentState is DivineAuthFormState) {
         emit(
           currentState.copyWith(
             isSubmitting: false,
@@ -228,7 +228,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       );
 
       final current = state;
-      if (current is DivineAuthFormState) {
+      if (!isClosed && current is DivineAuthFormState) {
         emit(
           current.copyWith(isSubmitting: false, signInFailureReason: reason),
         );
@@ -286,7 +286,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       );
 
       final current = state;
-      if (current is DivineAuthFormState) {
+      if (!isClosed && current is DivineAuthFormState) {
         emit(
           current.copyWith(
             isSubmitting: false,
@@ -313,6 +313,8 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
         inviteCode: _inviteCode,
       );
 
+      if (isClosed) return;
+
       // Emit email verification state
       emit(
         DivineAuthEmailVerification(
@@ -323,7 +325,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       );
     } else {
       final current = state;
-      if (current is DivineAuthFormState) {
+      if (!isClosed && current is DivineAuthFormState) {
         emit(
           current.copyWith(
             isSubmitting: false,
--- a/mobile/lib/blocs/divine_auth/divine_auth_cubit.dart
+++ b/mobile/lib/blocs/divine_auth/divine_auth_cubit.dart
@@ -487,6 +487,13 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
         category: LogCategory.auth,
       );
 
+      // The account creation above flips AuthService to `authenticated`
+      // before it returns, which reroutes off the create-account screen and
+      // closes this cubit. Emitting the success state is then both
+      // impossible and unnecessary — the navigation it would trigger has
+      // already happened.
+      if (isClosed) return;
+
       emit(const DivineAuthSuccess());
     } on InviteApiException catch (e, stackTrace) {
       Log.error(
@@ -499,7 +506,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       addError(e, stackTrace);
 
       final currentState = state;
-      if (currentState is DivineAuthFormState) {
+      if (!isClosed && currentState is DivineAuthFormState) {
         emit(
           currentState.copyWith(
             isSkipping: false,
@@ -521,7 +528,7 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       addError(e, stackTrace);
 
       final currentState = state;
-      if (currentState is DivineAuthFormState) {
+      if (!isClosed && currentState is DivineAuthFormState) {
         emit(
           currentState.copyWith(
             isSkipping: false,
--- a/mobile/lib/blocs/divine_auth/divine_auth_cubit.dart
+++ b/mobile/lib/blocs/divine_auth/divine_auth_cubit.dart
@@ -126,6 +126,11 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
     emit(current.copyWith(obscurePassword: !current.obscurePassword));
   }
 
+  void _emitIfOpen(DivineAuthState nextState) {
+    if (isClosed) return;
+    emit(nextState);
+  }
+
   /// Validate and submit the form
   Future<void> submit() async {
     final current = state;
@@ -194,8 +199,8 @@ class DivineAuthCubit extends Cubit<DivineAuthState> {
       addError(e, stackTrace);
 
       final currentState = state;
-      if (!isClosed && currentState is DivineAuthFormState) {
-        emit(
+      if (currentState is DivineAuthFormState) {
+        _emitIfOpen(
           currentState.copyWith(
             isSubmitting: false,
             generalError: 'An unexpected error occurred. Please try again.',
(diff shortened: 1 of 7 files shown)


hm21 (approved):
LGTM — both fixes are correct and the new tests are real regression tests (verified red-before-fix). Pushed two reviewer commits per REVIEW_POLICY, please glance before merging: 80ba816258 simplifies the fire-and-forget logout to `Future.ignore()`, and 8ec6f37b16 applies the same `isClosed` guards to `_exchangeCodeAndLogin`, which has the identical close-mid-flight race, with a mirroring regression test (also verified red without the guard).

mbradley (changes requested):
Both bugs here are real and both fixes are correct. I reproduced the emit-after-close crash on the current head before changing anything, and confirmed the keycast test genuinely fails against the old `unawaited` form, so that fix is properly pinned.

Five reviewer commits pushed, one per finding. Please glance before merging, and revert any you disagree with.

**`30f4c9db5` guards the five remaining emit-after-close sites.** This is the substantive one. The two methods that were fixed left five structurally identical emits unguarded: `submit()`'s outer catch, both `_handleSignIn` and `_handleSignUp` failure paths, and the email-verification emit. `submit()`'s catch is the one that matters most, because it is the last-resort handler. A `StateError` thrown by `_handleSignUp`'s email-verification emit lands there, and the catch's own recovery emit throws again with nothing left to catch it. Since `submit()` is fire-and-forget at all three call sites, that second throw escapes as exactly the unhandled async error this branch set out to remove. Reproduced on head:

```
[AUTH] Auth submission error: Bad state: Cannot emit new states after calling close
Bad state: Cannot emit new states after calling close
  package:bloc/src/bloc_base.dart 100:9            BlocBase.emit
  .../divine_auth_cubit.dart 198:9   DivineAuthCubit.submit
```

**`182d00978` makes the sign-in guard detectable.** Worth flagging directly, since @hm21's review recorded that test as verified red without the guard. I could not reproduce that. Deleting the `isClosed` guard on the sign-in path leaves the entire file green at 69 passing, with the log line showing the emit-after-close firing unseen. The mechanism: `emit` throws, the generic catch swallows it, and the `!isClosed` check then suppresses the recovery emit, so the future completes normally and the state is untouched. Both original assertions hold either way. Only an assertion on the bloc observer can see the difference, which is the shape the anonymous path already used. It fails correctly now.

**`279514df1` drops a redundant test.** `does not throw when the cubit closes mid-creation` is subsumed by its observer sibling. Mutation-checked both ways: removing only the success guard leaves it green while the observer test fails; removing the catch guards too fails both. It fails in no case the observer test doesn't already cover. Its comment carried the rationale for the race, so that moved to the test that stays.

**`bf304acdd`** corrects the comment's stated reason. `DivineAuthSuccess` triggers no navigation on the create-account screen: the listener has no branch for that state and the redirect runs off `AuthService` auth state. Conclusion was right, reason wasn't. **`7d41fc207`** trims the logout comment, keeping the part that stops someone re-adding a `try`/`catch`.

**Validation:** 86 affected app tests plus 236 `keycast_flutter` tests green, `flutter analyze lib test integration_test` clean, package-level analyze clean, no lockfile churn, no merge conflicts. Every test I added or changed was mutation-checked to confirm it fails when its guard is removed.

Two things I did not touch, because they are yours to decide.

**1. The sign-in path now silently drops `TextInput.finishAutofillContext()`.** `login_options_screen.dart:92` runs it off `DivineAuthSuccess`. When the redirect wins the race the cubit is already closed, so that signal never fires and the user gets no password-manager save prompt. This is not a regression, the emit previously threw before listeners ran, so the change converts a crash into a quiet skip. But it is a real behaviour question: accept losing the autofill signal on that path, or move the call somewhere the race cannot skip it?

**2. The description's framing does not match the branch.** Creating an anonymous account never calls keycast `logout()`. That method has exactly one caller, `auth_service.dart:1939`, reachable only from `signOut()` and `_completeDestructiveSignOutAfterDeletedKeys()`. So the second bug lives on the sign-out path and cannot fire during anonymous signup: these are two unrelated fixes under one narrative. The description also predates @hm21's sign-in commit, which is roughly a third of the diff and isn't mentioned. Worth a rewrite so the next reader knows what to regression-test.

One more, lower priority and pre-existing, but the diff rewrites the exact statement so it seemed worth raising rather than dropping: the best-effort logout POST cannot revoke anything. `keycast`'s handler at `api/src/api/http/auth.rs:1283` is `pub async fn logout()` with no extractors and no lookup, and the mobile client sends it unauthenticated after the tokens are already deleted locally. It only returns a cookie-clearing header, which a plain `http.Client` has no use for. Keep the no-op call, or correct the doc comment that implies it does something?

Requesting changes on the two questions above rather than the code. Happy to take either of them on if you'd rather hand it back.

rabble:
Both of your open questions are resolved on the branch, and the third (lower-priority) one too.

**1. Autofill signal — moved, not accepted.** `4e70e0d6b`. You were right that it's a real behaviour question, and dropping the save prompt wasn't the answer we wanted. The commit now hangs off `AuthService.authStateStream` rather than `DivineAuthSuccess` — the same signal `RouterRefreshListenable` uses to drive the redirect that closes the cubit. Stream events are delivered in a microtask, before the frame that disposes the subtree, so the redirect can no longer win that race. Gated on a submitted credential form (`DivineAuthFormState.isSubmitting`) so Amber / NIP-07 sign-ins from the same screen don't commit an empty autofill context.

Three tests in `login_options_screen_test.dart`, and the primary one was mutation-checked: neutering the commit turns it red while the other two stay green. One setup note in case it bites someone later — the handler on `SystemChannels.textInput` has to be installed *after* `pumpWidget`, because the test binding installs its own `TestTextInput` handler on that channel while bringing the tree up and silently replaces one registered in `setUp`. That cost me a confusing red where the listener demonstrably fired but nothing was recorded.

**2. Description — rewritten.** It now separates the two bugs instead of narrating them as one anonymous-signup story, states plainly that the keycast fix is on the sign-out path with its single caller at `auth_service.dart:1939`, and credits both of your reviews and @hm21's with the commit SHAs. Your finding on `182d00978` is called out specifically in the testing section, since "the assertion has to go through the observer or it stays green either way" is the non-obvious thing a future reader needs.

**3. Logout endpoint doc.** `9a37f6b49`. Took the "correct the doc comment" option — the old one implied a server-side logout that keycast's handler cannot perform. The call stays in case the endpoint grows real revocation; the comment no longer claims it already has.

Validation on the two new commits: 190 tests across `test/screens/auth` and `test/blocs/divine_auth`, 236 `keycast_flutter` tests, `flutter analyze lib test integration_test` clean, package analyze clean. No rebase — GitHub reports no conflicts and these are review fixes.

Re-requesting your review since the changes-requested is what's blocking the merge. Thanks for the mutation-checking on `182d00978` in particular; that one would have shipped as a test that couldn't fail.

mbradley (approved):
Verified at `1280f9f86`. All three questions resolved, and I re-ran the mutation checks rather than relying on the claim: neutering `finishAutofillContext()` fails the primary autofill test while the other two stay green, and removing the `isSubmitting` gate fails the Amber / NIP-07 test. Both halves of the new logic are pinned.

Moving autofill onto `authStateStream` is the right call over accepting the dropped prompt. It's the same signal that drives the redirect, so the race that caused this cannot reappear. I also checked the second listener is safe: `_authStateController` is a `.broadcast()` controller, so listening alongside `RouterRefreshListenable` is fine.

190 tests across `test/screens/auth` and `test/blocs/divine_auth` green locally, 15/15 CI checks green. Clearing my changes-requested.

Two things worth their own PRs rather than this one, noting them here so they don't get lost:

- `DivineAuthCubit` is now at 14 emit sites, all hand-guarded. `invite_status_cubit.dart:72` and `crossposting_settings_cubit.dart:336` have both converged on a private `_emitIfOpen` helper. The next emit added to this file will miss the guard the same way these five did.
- The assertion trap behind `182d00978` is worth a line in `.claude/rules/testing.md`. The existing "a test must be able to fail" section doesn't cover the specific shape where a generic `catch` swallows the very error the test exists to observe, so a `completes`-style assertion looks convincing and cannot fail.

NotThatKindOfDrLiz (approved):
Final takeover pass is clean. The remaining review concerns are addressed, #6607 is linked, and CI is green on `099c0828c`.

NotThatKindOfDrLiz:
Takeover changes are pushed in `099c0828c`.

What changed:
- Centralized the closed-cubit guard in `DivineAuthCubit` via `_emitIfOpen`, replacing the repeated post-await `!isClosed && emit(...)` pattern.
- Removed the dead `DivineAuthSuccess` listener condition from `create_account_screen.dart`.
- Linked the PR to #6607 and expanded the issue body with the observed failure modes.

Verification:
- `flutter test test/blocs/divine_auth/divine_auth_cubit_test.dart test/screens/auth/login_options_screen_test.dart`
- Pre-push hook: analyzer plus changed-file tests for keycast OAuth, DivineAuthCubit, create-account, and login-options.
- GitHub CI is green on `099c0828c`.

@dcadenas please review the takeover commit and merge when you are ready.

PR #6576, opened 2026-08-01
chore(local_stack): actionable diagnostics for port conflicts and partial stacks

### LINKED ISSUE #6577: chore(local_stack): actionable diagnostics for port conflicts and partial stacks
Auto-created from PR #6576.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6576
**Branch:** `chore/local-stack-diagnostics`

dcadenas:
Closes #6577

## Summary

When the local stack failed to start, it did not tell you why or what to do about it.
Now it checks ports before starting, survives transient startup races, and reports which parts of the stack came up.
When a port is taken it names the container holding it and prints the command to remove it.

## What Changed

The second commit fixes a real parsing bug found while using the tool. Splitting `docker ps` output on tabs collapsed the empty compose-project field, so standalone containers were misreported as host processes — which sends you hunting for a process that does not exist.

- Port pre-flight checks before startup.
- Transient startup races are retried instead of failing the run.
- Partial-stack reporting, so you can see what came up and what did not.
- Skill documentation for the workflow.
- Fixed the tab-delimited `docker ps` parse.

## Testing

Verified by causing the actual failures rather than by reasoning about them.

- Triggered a real port conflict held by a container: the tool names the container and prints the `docker rm -f` command.
- Triggered a conflict held by a non-container host process: correctly identified as a host process.

## Risks

Low. Developer tooling only, with no effect on app or production behaviour.

### DIFF
--- a/.claude/skills/e2e-test/SKILL.md
+++ b/.claude/skills/e2e-test/SKILL.md
@@ -39,8 +39,8 @@ lose the timeline and the diagnostics.
 | Service | Port | Purpose |
 |---|---|---|
 | Keycast | 43000 | OAuth + NIP-46 signer |
-| FunnelCake Relay | 47777 | Nostr relay |
-| FunnelCake API | 43001 | REST API |
+| FunnelCake Relay | 47777 | Nostr relay (WebSocket) |
+| FunnelCake API | 47777 | REST API, under `/api/` on the same proxy |
 | Blossom | 43003 | Media server |
 | Postgres | 15432 | Keycast DB |
 | Invite | 43004 | divine-invite-darshan (Viceroy) |
@@ -98,13 +98,22 @@ are routine. The check names the service, the port, and the holder:
 
   port 45173  wanted by service "keycast"
             held by a host process (not a container), listening on: 127.0.0.1:45173
-            find it: ss -ltnp "sport = :45173"   (or: sudo lsof -iTCP:45173 -sTCP:LISTEN)
+            find it: sudo lsof -nP -iTCP:45173 -sTCP:LISTEN
 ```
 
 Ports already published by our own containers are not conflicts —
 `up.sh` is idempotent. The raw daemon error it replaces (`Bind for
 0.0.0.0:16380 failed: port is already allocated`) named neither the
 service nor the holder.
+
+Listening sockets come from `ss` on Linux and `lsof` on macOS, and
+the `find it:` line names whichever of the two the machine has. With
+neither installed the run says so and falls back to container-held
+ports alone, which `docker ps` reports without either tool — and a
+stale container is the usual culprit anyway.
+
+`bash local_stack/test_stack_scripts.sh` covers these paths against a
+stubbed docker/ss/lsof, so it needs no daemon and no free ports.
 
 ### Startup races
 
--- a/.claude/skills/e2e-test/SKILL.md
+++ b/.claude/skills/e2e-test/SKILL.md
@@ -5,7 +5,7 @@ description: |
   app against a local Docker backend (no mocks). Use when running
   E2E tests, debugging failures, or working on the local harness.
 author: Claude Code
-version: 1.1.0
+version: 1.2.0
 ---
 
 # E2E Integration Testing
@@ -43,6 +43,22 @@ lose the timeline and the diagnostics.
 | FunnelCake API | 43001 | REST API |
 | Blossom | 43003 | Media server |
 | Postgres | 15432 | Keycast DB |
+| Invite | 43004 | divine-invite-darshan (Viceroy) |
+
+The app reaches these at `10.0.2.2` from the emulator. Cleartext to
+loopback hosts is permitted in every build type on both platforms.
+
+### Only start what your flow needs
+
+Most services are irrelevant to any given test, and `local_up`
+failing on one does not mean you are blocked. An invite-only flow
+needs `invite` alone: a locally-generated nsec signs on-device, so
+no Keycast, and `onboarding_mode=open` means no invite gate. Check
+what is actually healthy before debugging a service you never call:
+
+```bash
+docker compose -f local_stack/docker-compose.yml ps
+```
 
 ```bash
 mise run local_up         # Start (auto-runs local_setup on fresh worktrees)
@@ -58,6 +74,73 @@ bypass the seed:
 
 ```bash
 bash ../local_stack/profile.sh integration_test/<your_test>.dart
+```
+
+Any `local_up` failure prints the per-service status, the logs of
+whatever is down, and that same bypass command. Set `E2E_TEST_PATH`
+before the run and it prints the command for *your* test:
+
+```bash
+E2E_TEST_PATH=integration_test/auth/auth_journey_test.dart mise run local_up
+```
+
+### Port conflicts
+
+`up.sh` pre-flights every host port in `docker-compose.yml` before
+starting anything. This machine runs several compose projects, and
+stale test containers days old are the normal case, so collisions
+are routine. The check names the service, the port, and the holder:
+
+```
+  port 43000  wanted by service "keycast"
+            held by container "funnelcake-test-clickhouse-sim" — compose project "funnelcake-test"
+            remedy: docker rm -f funnelcake-test-clickhouse-sim
+
+  port 45173  wanted by service "keycast"
+            held by a host process (not a container), listening on: 127.0.0.1:45173
+            find it: ss -ltnp "sport = :45173"   (or: sudo lsof -iTCP:45173 -sTCP:LISTEN)
+```
+
+Ports already published by our own containers are not conflicts —
+`up.sh` is idempotent. The raw daemon error it replaces (`Bind for
+0.0.0.0:16380 failed: port is already allocated`) named neither the
+service nor the holder.
+
+### Startup races
+
+Containers sometimes start before Docker's embedded DNS knows a
+dependency's alias: `funnelcake-migrate` dies with `dial tcp: lookup
+funnelcake-clickhouse on 127.0.0.11:53: no such host`, or `keycast`
+burns its DB connection attempts on `Temporary failure in name
+resolution`. Both succeed on an unchanged retry. `up.sh` re-runs the
+whole `up` (idempotent — it restarts whatever died) up to 3 attempts,
+5s apart, **only** when it sees a name-resolution signature in the
+compose output or in the failed containers' logs. A port clash or a
+bad image fails straight through rather than retrying pointlessly.
+
+### Running a locally built backend
+
+Compose pulls `ghcr.io/divinevideo/divine-invite-darshan:e2e`. To
+test an unmerged backend branch, build it and tag it as that name so
+compose uses the local image without pulling:
+
+```bash
+docker build -f Dockerfile.local -t divine-invite-darshan:local .
+docker tag divine-invite-darshan:local ghcr.io/divinevideo/divine-invite-darshan:e2e
+```
+
+`invite` is one of the few services without `pull_policy: always`, so
+a plain `mise run local_up` keeps your tag. Use `local_up_cached` if
+you have overridden a service that *does* pull on every start.
+
+### Cross-repo gotcha: `kv-store-data.json`
+
+The invite service (`divine-invite-darshan`) reads a gitignored
+`kv-store-data.json`. A **fresh worktree of that repo does not have
+it**, and without it the entire invite-service suite fails. Seed it:
+
+```bash
+printf '{}' > kv-store-data.json
 ```
 
 ## Emulator
@@ -77,6 +160,19 @@ Stale-state debugging cost is yours.
 
 Buffer auth-flow logs: `adb logcat -G 16M` (default 256 KB rotates
 mid-flow).
+
+### Storage exhaustion
+
+Not just a Patrol problem — `flutter run` hits it too, and the error
+is on the install, not the build:
+
+```
+java.io.IOException: Requested internal only, but not enough space
+```
+
+A debug APK is ~289 MB and needs real headroom on top of that.
+`adb shell pm trim-caches 1G` often does not free enough; `mise run
+emulator_wipe` (`emulator.sh --wipe`) is usually the faster fix.
 
 ## Patterns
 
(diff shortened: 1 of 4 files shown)


local_stack/preflight.sh
```diff
@@ -0,0 +1,223 @@
...
+    fi
+
+    # port -> space-separated local addresses currently listening on it
+    local -A listen_addrs=()
+    local local_addr port
+    while read -r _ _ _ local_addr _; do
+        port="${local_addr##*:}"
+        [[ "$port" =~ ^[0-9]+$ ]] || continue
+        listen_addrs["$port"]+="${local_addr} "
+    done < <(ss -Hltn 2>/dev/null || true)
```
realmeylisdev: **should-fix.** `ss` is iproute2, so it doesn't exist on macos. `2>/dev/null` eats the `command not found` and `|| true` eats the status, so `listen_addrs` is never populated. line 98 (`[[ -n "${listen_addrs[$port]:-}" ]] || continue`) then skips every port, `conflicts` stays 0, and the function returns 0. the whole feature is inert on macos.

reproduced on macos 15 / bash 3.2 / compose v5.3.1, standalone container on 16380:
```
$ docker ps --format '{{.Names}}|{{.Label "com.docker.compose.project"}}|{{.Ports}}'
pf-conflict-test||0.0.0.0:16380->6379/tcp, [::]:16380->6379/tcp
$ preflight_ports docker-compose.yml; echo $?
0                      # no conflict reported
$ docker compose up -d --wait funnelcake-redis
Error response from daemon: ... Bind for 0.0.0.0:16380 failed: port is already allocated
```
with a shim putting linux-shaped `ss -Hltn` output on PATH, the same run prints exactly the intended report (`held by container "pf-conflict-test" - standalone container`, plus the `docker rm -f`). so `ss` is the single discriminator and the rest of the logic is fine.

macos is a supported platform for this dir: `local_stack/setup.sh:22` prints the `brew install --cask docker` instructions, `local_stack/emulator.sh:56` branches on `Darwin`, `local_stack/android_sdk.sh:12` probes `$HOME/Library/Android/sdk`, and the README says the stack works against the iOS Simulator.

`lsof` is present on macos and you already name it at line 134, so something like: prefer `ss -Hltn`, fall back to `lsof -nP -iTCP -sTCP:LISTEN`, and if neither exists say so out loud rather than passing.

local_stack/preflight.sh
```diff
@@ -0,0 +1,223 @@
...
+    fi
+
+    project="$(printf '%s' "$config_json" | _stack_project_name)"
+    ports="$(printf '%s' "$config_json" | _stack_published_ports)"
+    if [[ -z "$ports" ]]; then
+        return 0
+    fi
+
+    # port -> space-separated local addresses currently listening on it
+    local -A listen_addrs=()
```
realmeylisdev: **should-fix.** macos ships bash 3.2 as the only bash on PATH, and `mise run local_up` runs `bash ../local_stack/up.sh` (mobile/mise.toml:8), so `#!/usr/bin/env bash` lands on 3.2. `local -A` is bash 4+, so every run prints:
```
preflight.sh: line 71: local: -A: invalid option
local: usage: local name[=value] ...
preflight.sh: line 80: local: -A: invalid option
local: usage: local name[=value] ...
```
results are still correct, because every key here is a port number and the fallback is a plain indexed array, but it also silently leaks these into the global scope since the `local` never takes effect.

since the keys are always integers, dropping `-A` and using indexed arrays works identically on 3.2 and 4+ and the noise goes away. same at line 80.

local_stack/preflight.sh
```diff
@@ -0,0 +1,223 @@
...
+
+    if ! config_json="$(docker compose -f "$compose_file" config --format json 2>&1)"; then
+        echo "ERROR: could not parse ${compose_file}:" >&2
+        echo "$config_json" >&2
+        return 1
+    fi
+
+    project="$(printf '%s' "$config_json" | _stack_project_name)"
+    ports="$(printf '%s' "$config_json" | _stack_published_ports)"
+    if [[ -z "$ports" ]]; then
```
realmeylisdev, line 123: **should-fix.** `_stack_published_ports` has no error handling, so any failure is indistinguishable from "the stack publishes no ports" and returns success. `set -e` doesn't save you here either, because up.sh calls this as `if ! preflight_ports ...`, which disables errexit for the whole function body.

verified with a stub that makes python3 exit 127:
```
$ PATH=/tmp/nopy:$PATH bash -c 'source preflight.sh; preflight_ports docker-compose.yml'
python3: command not found
python3: command not found
>>> preflight returned SUCCESS
```
worth failing closed, the way the `docker compose config` check just above already does.

local_stack/preflight.sh
```diff
@@ -0,0 +1,223 @@
...
+
+# preflight_ports <compose_file>
+# Returns 0 if every published host port is free or held by this project's own
+# containers; 1 (with an actionable report on stderr) otherwise.
+preflight_ports() {
+    local compose_file="$1"
+    local config_json project ports
+
+    if ! docker ps --format '{{.Names}}' >/dev/null 2>&1; then
+        echo "ERROR: cannot talk to the Docker daemon. Is it running? (systemctl status docker)" >&2
```
realmeylisdev: nit: no `systemctl` on macos (docker desktop). `local_stack/emulator.sh:56` already branches on `uname -s == Darwin`, so there's an existing pattern to follow.

local_stack/preflight.sh
```diff
@@ -0,0 +1,223 @@
...
+                printf '  port %-6s wanted by service "%s"\n' "$port" "$service"
+                printf '            held by container "%s" — %s\n' "$holder" "$owner_desc"
+                printf '            remedy: docker rm -f %s\n' "$holder"
+            } >&2
+        else
+            {
+                printf '  port %-6s wanted by service "%s"\n' "$port" "$service"
+                printf '            held by a host process (not a container), listening on: %s\n' \
+                    "${listen_addrs[$port]% }"
+                printf '            find it: ss -ltnp "sport = :%s"   (or: sudo lsof -iTCP:%s -sTCP:LISTEN)\n' \
```
realmeylisdev: nit: `ss -ltnp` is the headline remedy but the operator who hits this on macos has no `ss`. the parenthetical `lsof` is the one that actually runs there. flipping the order per platform (or just leading with `lsof`) would help. same sample output is in SKILL.md:101.

local_stack/preflight.sh
```diff
@@ -0,0 +1,223 @@
...
+    shift 2
+
+    if grep -qE "$_STACK_TRANSIENT_RE" "$up_log" 2>/dev/null; then
+        return 0
+    fi
+
+    # The daemon error is usually generic ("dependency failed to start"); the
+    # real signature is in the failed container's own log.
+    local logs
+    logs="$(docker compose -f "$compose_file" logs --tail=50 "$@" 2>/dev/null || true)"
```
realmeylisdev: nit: `"$@"` is the full `REPORT_SERVICES` list, so this greps the last 50 lines of all 15 services, healthy ones included. the comment above says "the failed container's own log" and SKILL.md:117 says "the failed containers' logs", but neither is what runs.

practical effect: `i/o timeout` is a common enough string that one healthy service having it in its recent output flips a genuine hard failure into transient, costing two more `up --wait` cycles before the real report prints. narrowing to the services that aren't running would match the comment and the doc.

local_stack/up.sh
```diff
@@ -24,9 +27,60 @@ SERVICES=(
...
+
+  stack_failure_report "$COMPOSE_FILE" "$SCRIPT_DIR" "${REPORT_SERVICES[@]}"
+  exit "$UP_RC"
+done
+
+if ! docker compose -f "$COMPOSE_FILE" run --rm e2e-seed; then
+  echo "" >&2
+  echo "ERROR: e2e-seed failed. The services themselves may still be healthy —" >&2
+  echo "only tests that read seeded relay/blossom data need it." >&2
+  stack_failure_report "$COMPOSE_FILE" "$SCRIPT_DIR" "${REPORT_SERVICES[@]}"
```
realmeylisdev: nit: `REPORT_SERVICES` (line 32) doesn't include `e2e-seed`, so this report structurally can't show why seed failed. it prints `=== Local stack failed to come up ===` and then lists 15 healthy services, right after you've correctly told the user the services are probably fine. the seed logs are the one thing worth showing on this path.

.claude/skills/e2e-test/SKILL.md
```diff
@@ -60,6 +76,73 @@ bypass the seed:
...
+whatever is down, and that same bypass command. Set `E2E_TEST_PATH`
+before the run and it prints the command for *your* test:
+
+```bash
+E2E_TEST_PATH=integration_test/auth/auth_journey_test.dart mise run local_up
+```
+
+### Port conflicts
+
+`up.sh` pre-flights every host port in `docker-compose.yml` before
```
realmeylisdev, line 89: nit: reads as unconditional, but per the preflight.sh:77 thread this doesn't run on macos at all. worth a platform caveat here until the `ss` fallback lands, otherwise this section tells a macos dev the check ran when it didn't.

.claude/skills/e2e-test/SKILL.md
```diff
@@ -43,6 +43,22 @@ lose the timeline and the diagnostics.
 | FunnelCake API | 43001 | REST API |
 | Blossom | 43003 | Media server |
 | Postgres | 15432 | Keycast DB |
+| Invite | 43004 | divine-invite-darshan (Viceroy) |
```
realmeylisdev, line 46: nit, and pre-existing rather than yours, but you're editing this table: 43001 on the row above doesn't exist anywhere. repo-wide grep for `43001` returns only this line, and it's not in docker-compose.yml. the funnelcake REST API is behind funnelcake-proxy on 47777 (`local_stack/funnelcake-proxy/nginx.conf` routes `/api/` to `funnelcake-api:3001`), which is the same port as the relay row. the new Invite row checks out: `43004:7676`, and `auth_invalid_binding` is a real 401 in divine-invite-darshan (`src/error.rs:89,128`).

realmeylisdev (commented):
the linux path is solid and shellcheck is clean, but the port pre-flight is a no-op on macos. `ss` isn't installed there, `2>/dev/null || true` swallows it, so `listen_addrs` stays empty and the first guard skips every port. i reproduced it: stale container holding 16380, preflight returns 0, then compose dies with the exact `Bind for 0.0.0.0:16380 failed: port is already allocated` this PR exists to replace. stubbing `ss` in makes the same run print the correct report, so that's the only cause.

the python3 extractor fails open the same way. rest are nits. details inline.

local_stack/preflight.sh
realmeylisdev: fixed in b2b511ef3. `_stack_listen_addrs` prefers `ss -Hltn`, falls back to `lsof -nP -iTCP -sTCP:LISTEN -Fn`, and returns non-zero when neither exists so the caller warns instead of passing. your repro now prints the container report and returns 1.

one thing on top: a container-held port is a conflict on its own now, whether or not the listener enumeration saw the socket. `docker ps` reports it without either tool, and that is the stale-container case this exists to catch.

local_stack/preflight.sh
realmeylisdev: fixed in a22890482. both maps are plain indexed arrays now, keyed by port number as you suggested. added a header comment recording the 3.2 constraint so it does not come back.

local_stack/preflight.sh
realmeylisdev: fixed in 1aa9bdad3. explicit python3 guard plus status checks on both extractors. an empty project name errors too — matching against `""` would flag our own running stack as a clash.

local_stack/preflight.sh
realmeylisdev: fixed in 639cc1699 — `open -a Docker` on Darwin, `systemctl status docker` elsewhere.

local_stack/preflight.sh
realmeylisdev: fixed in b2b511ef3. `_stack_port_finder_cmd` prints whichever of the two the machine has, so the line is always runnable where it is read. SKILL.md sample updated in 4d4ef1636.

local_stack/preflight.sh
realmeylisdev: fixed in 31243667e. `_stack_down_services` now feeds both this and the report, so only not-running services get grepped, and "nothing is down" returns not-transient rather than grepping everything. the comment and the doc now describe what actually runs.

local_stack/up.sh
realmeylisdev: fixed in bd1f0b50a. `--rm` takes the container away before logs could be asked for, so the run is tee'd and that tail printed instead. kept the status table (a down dependency does explain a seed failure) and the bypass, dropped the headline that was not true, and it exits with the seed's own status rather than a flat 1.

`stack_failure_report` output is unchanged — diffed before/after against identical stubs.

.claude/skills/e2e-test/SKILL.md
realmeylisdev: the fallback landed in b2b511ef3, so rather than a caveat, 4d4ef1636 states which tool is used where and what happens with neither installed.

.claude/skills/e2e-test/SKILL.md
realmeylisdev: fixed in 4d4ef1636 — both rows are 47777 now, API under `/api/` per funnelcake-proxy/nginx.conf.

realmeylisdev (approved):
took these over — 8 commits on this branch, one per finding, so any of them can be reverted alone. details in each thread.

both should-fixes and all five nits are addressed. one thing beyond the notes: a container-held port is now a conflict on its own, so the stale-container case still gets caught with neither `ss` nor `lsof` installed.

added `local_stack/test_stack_scripts.sh` — stubs docker/ss/lsof on a sandboxed PATH, so the cases need no daemon, no free ports, and give the same result on either OS. every case fails against the pre-fix code and passes after.

verified on macos 15 / bash 3.2 / compose v5.3.1: real container conflict and real host-process conflict each report correctly and return 1, clean machine returns 0, `compose run --rm | tee` propagates the seed's exit code under bash, `stack_failure_report` output is byte-identical to before the split, shellcheck clean.

not wired into CI — `test_android_scripts.sh` isn't either, so I followed the sibling rather than adding a workflow on your PR. yours to merge, tweak, or push back on.

PR #6582, opened 2026-08-01
fix(drafts): keep library clips in the ghost-frame scan

### LINKED ISSUE #6581: fix(drafts): ghost-frame scan skips library clips when no draft is excluded
## Description

`DraftStorageService._referencedGhostFrameFilenames` skips a clip row when its
`draftId` equals `excludeDraftId`:

```dart
for (final row in clipRows) {
  if (row.draftId == excludeDraftId) continue;
```

`clearAllDrafts()` calls it with no argument, so `excludeDraftId` is `null`.
Library clips are exactly the rows with a `NULL` `draft_id`, so `null == null`
is true and **every library clip is skipped** — the rows the call exists to
protect.

The intent is documented right above the call site, and it says the opposite:

```dart
// Every draft and its clips are gone; the clips that survive are library
// clips (active or trashed). Ghost frames live only in the clip `data`
// blob, so protect any a surviving library clip still references. No
// excludeDraftId — all remaining clips are legitimate survivors.
final referencedGhostFrames = await _referencedGhostFrameFilenames();
```

The `deleteDraft` path is unaffected: there `excludeDraftId` is a real draft id,
so `null == 'draft_x'` is false and library clips are included correctly.

## Impact

Latent today, not user-visible. `FileCleanupService.deleteGhostFrameFiles` falls
back to `deleteFileIfUnreferenced` for any basename missing from the set, and
that path runs `ClipsDao.referencedFilenames`, whose `_jsonFilePathKeys` already
contains `ghostFramePath`. The backstop finds the reference and keeps the file.

It becomes a real ghost-frame deletion the moment that backstop stops covering
`ghostFramePath` — e.g. if `_jsonFilePathKeys` is ever trimmed, or if the
fallback is skipped as an optimisation.

## Evidence

Removing `'ghostFramePath'` from `ClipsDao._jsonFilePathKeys` (disabling the
backstop, leaving everything else intact) makes exactly one test in
`mobile/test/services/draft_storage_service_test.dart` fail:

```
DraftStorageService clip ghost-frame file hygiene
  keeps a ghost frame referenced by a surviving library clip when all drafts are cleared
```

The sibling cases — a ghost frame shared with a duplicate, and one referenced by
a trashed clip — keep passing under the same mutation. Those rows carry a
non-null `draft_id`, so the scan does return them; the library-clip row is the
one the null comparison drops.

## Suggested fix

```dart
if (excludeDraftId != null && row.draftId == excludeDraftId) continue;
```

The existing "surviving library clip" test already covers the behaviour once the
backstop is no longer masking it, so a regression test would need to assert the
returned set directly rather than the file outcome.

## Files

- `mobile/lib/services/draft_storage_service.dart` — `_referencedGhostFrameFilenames`, and the `clearAllDrafts` call site
- `mobile/test/services/draft_storage_service_test.dart` — `clip ghost-frame file hygiene`

## Context

Found while reviewing #6561, which changes the data source of this scan
(`getAllClips` → `getClipsWithGhostFrames`) but not the guard. The same
comparison is on `main`, so it predates that PR and was left out of it to keep
the branch focused.

hm21:
## Description

`DraftStorageService._referencedGhostFrameFilenames` skipped every clip row whose `draftId` equalled `excludeDraftId`. `clearAllDrafts()` calls it with no argument, so `excludeDraftId` is `null` — and library clips are exactly the rows with a `NULL` `draft_id`, so `null == null` was true and **every library clip was skipped**: the rows the call exists to protect. The `deleteDraft` path was never affected, since there `excludeDraftId` is a real draft id.

The first fix guarded the null explicitly. Review then established that the exclusion is dead in both directions: both scans run *after* `DraftsDao.deleteDraft` has removed the draft **and** its clip rows in one transaction (`drafts_dao.dart:161-166`, awaited at `:602` before the scans at `:610`/`:645`), so no row the exclusion could skip ever reaches them. Deleting either guard line leaves all 77 tests in the two affected suites green.

So `excludeDraftId` is dropped from both scans instead. Library clips reach the ghost-frame scan by construction rather than by guard, and the nullable-exclude hazard class that caused this bug is gone with the parameter that carried it. The tradeoff is the lost safety net if someone later moves a scan ahead of the delete — that failure mode leaks a file rather than deleting one, and the delete-first ordering is deliberate and commented at `:599-601`.

Nothing is broken for users today. `FileCleanupService.deleteGhostFrameFiles` falls back to `deleteFileIfUnreferenced`, whose JSON scan runs `ClipsDao.referencedFilenames` — and `_jsonFilePathKeys` still contains `ghostFramePath`, so the backstop finds the reference and keeps the file. It becomes a real ghost-frame deletion the moment that backstop stops covering the key, e.g. if `_jsonFilePathKeys` is trimmed or the fallback is skipped as an optimisation.

That same backstop is why the existing `keeps a ghost frame referenced by a surviving library clip when all drafts are cleared` test passes with **or** without the fix — it asserts the file outcome, which the fallback rescues either way. The regression test added here runs `clearAllDrafts` against a `ClipsDao` whose `referencedFilenames` and `isFileReferenced` both return nothing, leaving the ghost-frame set as the only thing that can protect the file.

It also stops paying for that rescue. With the set always empty, every shared ghost frame in `clearAllDrafts` fell through to `deleteFileIfUnreferenced`, and `ClipsDao.referencedFilenames` resolves an unindexed name with a full `SELECT data FROM clips` plus a JSON decode of every row — once per frame. Those short-circuit on the set now, which is the same scan cost #6548 took off this path.

**Related Issue:** Closes #6581

## Sibling Scan

`_referencedLocalAudioFilenames` carried the same comparison shape, but it compares against `row.id`, a non-null primary key, so `null` could never match and it was never broken. Its `excludeDraftId` is dead for the same reason as the ghost-frame one — `deleteDraft` removes the draft row at `:602` before `getAllDrafts()` runs — so it is dropped too, and both functions read the same way again.

## Verification

- Reproduced first, per the issue's evidence recipe: with `'ghostFramePath'` removed from `ClipsDao._jsonFilePathKeys` (backstop disabled, everything else intact), exactly one test failed — the library-clip case. The duplicate and trashed-clip siblings stayed green, confirming the null comparison drops library clips and nothing else. Re-running with the fix applied and the backstop still disabled turned it green.
- Dead-branch check: deleting **both** exclusion guards from the shipped code leaves all 77 tests in `draft_storage_service_test.dart` + `file_cleanup_service_test.dart` green. That is the evidence the branch is unreachable through the real DAO, not just an argument that it should be.
- The regression test still bites after the parameter removal: mutating the scan to `if (row.draftId == null) continue;` fails exactly one test out of 69, and it is the new one. It guards the behaviour directly rather than guarding the guard.
- Swept `mobile/lib` and `mobile/packages` for the same shape — an identity skip whose *exclude* operand is nullable. Exactly two other sites qualify: `video_editor_timeline_overlay_strips.dart:116` (`String? selectedItemId`) and `timeline_snap_controller.dart:243` (`int? _releasedSnapMs`). Both are safe because the left operand is non-null — `TimelineOverlayItem.id` is a `String`, and `snapPoints` is a `Set<int>` — so `null == null` can never fire. This was the only broken instance.
- All 17 suites touching `DraftStorageService`: 464 passing, 2 pre-existing skips.
- `dart format` — clean.
- No device pass: the bug is latent behind the backstop and has no user-visible surface on a build, so a Samsung Galaxy run would show the same thing before and after. The disabled-backstop reproduction above is what actually distinguishes them.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/services/draft_storage_service.dart
+++ b/mobile/lib/services/draft_storage_service.dart
@@ -718,7 +718,9 @@ class DraftStorageService {
     final filenames = <String>{};
 
     for (final row in clipRows) {
-      if (row.draftId == excludeDraftId) continue;
+      // Guard the null: library clips carry a NULL draftId, so an unguarded
+      // comparison would drop exactly them when no draft is excluded.
+      if (excludeDraftId != null && row.draftId == excludeDraftId) continue;
       try {
         final data = json.decode(row.data) as Map<String, dynamic>;
         final ghostFramePath = data['ghostFramePath'] as String?;
--- a/mobile/lib/services/draft_storage_service.dart
+++ b/mobile/lib/services/draft_storage_service.dart
@@ -603,13 +603,10 @@ class DraftStorageService {
 
     // Ghost frames live only in the clip `data` blob, so — like draft-local
     // audio — the indexed-column check can't tell that a surviving draft (e.g.
-    // a duplicate sharing this draft's clips) still references them. Scan the
-    // other drafts' clips for shared ghost frames and keep those alive.
-    // Exclude this draft's own clips so its ghost frames aren't kept alive by
-    // rows that haven't cascade-deleted yet.
-    final referencedGhostFrames = await _referencedGhostFrameFilenames(
-      excludeDraftId: id,
-    );
+    // a duplicate sharing this draft's clips) still references them. This
+    // draft's own rows are already gone, so every clip the scan sees is a
+    // legitimate survivor whose ghost frame must stay alive.
+    final referencedGhostFrames = await _referencedGhostFrameFilenames();
 
     // Delete clip files only if not referenced by clip library
     await FileCleanupService.deleteRecordingClipsFiles(
@@ -645,15 +642,13 @@ class DraftStorageService {
         audioPaths,
         draftsDao: _draftsDao,
         clipsDao: _clipsDao,
-        referencedAudioFilenames: await _referencedLocalAudioFilenames(
-          excludeDraftId: id,
-        ),
+        referencedAudioFilenames: await _referencedLocalAudioFilenames(),
       );
     }
   }
 
-  /// Basenames of draft-local audio files referenced by any draft other than
-  /// [excludeDraftId], across all accounts.
+  /// Basenames of draft-local audio files referenced by any surviving draft,
+  /// across all accounts.
   ///
   /// The same local audio file can be shared by a draft and its publish copy —
   /// `copyWith` carries [DivineVideoDraft.editorStateHistory] and
@@ -662,17 +657,14 @@ class DraftStorageService {
   /// directly (audio paths live there, not in an indexed column) and never
   /// throws: a corrupt blob is logged and skipped.
   ///
-  /// [excludeDraftId] is required: the sole caller always excludes a draft,
-  /// and a nullable parameter would silently mean "scan everything".
-  Future<Set<String>> _referencedLocalAudioFilenames({
-    required String excludeDraftId,
-  }) async {
+  /// Callers run this *after* deleting the draft they are cleaning up, so
+  /// every row it sees is a survivor and there is nothing to exclude.
+  Future<Set<String>> _referencedLocalAudioFilenames() async {
     final rows = await _draftsDao.getAllDrafts();
     final documentsPath = await getDocumentsPath();
     final filenames = <String>{};
 
     for (final row in rows) {
-      if (row.id == excludeDraftId) continue;
       final DivineVideoDraft draft;
       try {
         draft = DivineVideoDraft.fromJson(
@@ -695,8 +687,8 @@ class DraftStorageService {
     return filenames;
   }
 
-  /// Basenames of clip ghost-frame files referenced by any clip other than
-  /// those belonging to [excludeDraftId], across all accounts.
+  /// Basenames of clip ghost-frame files referenced by any surviving clip,
+  /// across all accounts — library clips, whose `draftId` is NULL, included.
   ///
   /// A clip's ghost frame is persisted only inside the clip `data` blob (no
   /// indexed column), so a draft that shares clips with another draft — a
@@ -715,16 +707,13 @@ class DraftStorageService {
   /// table: this scan runs on every draft deletion, and decoding every blob in
   /// the database made it grow with the size of the clip library (#6548).
   ///
-  /// A null [excludeDraftId] excludes nothing: every clip is scanned, including
-  /// library clips, whose `draftId` is NULL. `clearAllDrafts` relies on that.
-  Future<Set<String>> _referencedGhostFrameFilenames({
-    String? excludeDraftId,
-  }) async {
+  /// Callers run this *after* deleting the draft they are cleaning up, so
+  /// every clip it sees is a survivor and there is nothing to exclude.
+  Future<Set<String>> _referencedGhostFrameFilenames() async {
     final clipRows = await _clipsDao.getClipsWithGhostFrames();
     final filenames = <String>{};
 
     for (final row in clipRows) {
-      if (excludeDraftId != null && row.draftId == excludeDraftId) continue;
       try {
         final data = json.decode(row.data) as Map<String, dynamic>;
         final ghostFramePath = data['ghostFramePath'] as String?;
@@ -770,8 +759,7 @@ class DraftStorageService {
 
     // Every draft and its clips are gone; the clips that survive are library
     // clips (active or trashed). Ghost frames live only in the clip `data`
-    // blob, so protect any a surviving library clip still references. No
-    // excludeDraftId — all remaining clips are legitimate survivors.
+    // blob, so protect any a surviving library clip still references.
     final referencedGhostFrames = await _referencedGhostFrameFilenames();
 
     // Delete clip files only if not referenced by clip library
--- a/mobile/lib/services/draft_storage_service.dart
+++ b/mobile/lib/services/draft_storage_service.dart
@@ -661,8 +661,12 @@ class DraftStorageService {
   /// the last referencing draft is deleted. Scans the draft `data` blobs
   /// directly (audio paths live there, not in an indexed column) and never
   /// throws: a corrupt blob is logged and skipped.
+  ///
+  /// [excludeDraftId] is required: unlike the ghost-frame scan there is no
+  /// scan-everything caller, so leaving it nullable would only invite the
+  /// `null == null` skip that #6581 fixed there.
   Future<Set<String>> _referencedLocalAudioFilenames({
-    String? excludeDraftId,
+    required String excludeDraftId,
   }) async {
     final rows = await _draftsDao.getAllDrafts();
     final documentsPath = await getDocumentsPath();
--- a/mobile/lib/services/draft_storage_service.dart
+++ b/mobile/lib/services/draft_storage_service.dart
@@ -662,9 +662,8 @@ class DraftStorageService {
   /// directly (audio paths live there, not in an indexed column) and never
   /// throws: a corrupt blob is logged and skipped.
   ///
-  /// [excludeDraftId] is required: unlike the ghost-frame scan there is no
-  /// scan-everything caller, so leaving it nullable would only invite the
-  /// `null == null` skip that #6581 fixed there.
+  /// [excludeDraftId] is required: the sole caller always excludes a draft,
+  /// and a nullable parameter would silently mean "scan everything".
   Future<Set<String>> _referencedLocalAudioFilenames({
     required String excludeDraftId,
   }) async {
@@ -715,6 +714,9 @@ class DraftStorageService {
   /// Uses [ClipsDao.getClipsWithGhostFrames] rather than reading the whole clip
   /// table: this scan runs on every draft deletion, and decoding every blob in
   /// the database made it grow with the size of the clip library (#6548).
+  ///
+  /// A null [excludeDraftId] excludes nothing: every clip is scanned, including
+  /// library clips, whose `draftId` is NULL. `clearAllDrafts` relies on that.
   Future<Set<String>> _referencedGhostFrameFilenames({
     String? excludeDraftId,
   }) async {
@@ -722,8 +724,6 @@ class DraftStorageService {
     final filenames = <String>{};
 
     for (final row in clipRows) {
-      // Guard the null: library clips carry a NULL draftId, so an unguarded
-      // comparison would drop exactly them when no draft is excluded.
       if (excludeDraftId != null && row.draftId == excludeDraftId) continue;
       try {
         final data = json.decode(row.data) as Map<String, dynamic>;
--- a/mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
+++ b/mobile/packages/db_client/lib/src/database/daos/clips_dao.dart
@@ -106,9 +106,12 @@ class ClipsDao extends DatabaseAccessor<AppDatabase> with _$ClipsDaoMixin {
   }
 
   /// Every clip row that carries a non-null `ghostFramePath`, projected to the
-  /// fields a ghost-frame reference scan needs: `draftId` to skip the draft
-  /// being deleted, `data` to read the frame out of, and `id` to name the row
-  /// if its blob turns out to be corrupt.
+  /// fields a ghost-frame reference scan needs: `data` to read the frame out
+  /// of, and `id` to name the row if its blob turns out to be corrupt.
+  /// `draftId` is returned too and is NULL for library clips; no caller
+  /// filters on it, and none should — the scan runs after the draft being
+  /// cleaned up has lost its own rows, so every row here is a survivor whose
+  /// ghost frame must stay alive (#6581).
   ///
   /// Ghost frames have no indexed column — they exist only inside the `data`
   /// blob — so the alternative is loading and JSON-decoding *every* clip row in
--- a/mobile/test/services/draft_storage_service_test.dart
+++ b/mobile/test/services/draft_storage_service_test.dart
@@ -2313,10 +2313,18 @@ void main() {
 /// `FileCleanupService` treats that check as a defensive backstop behind the
 /// caller-supplied ghost-frame set; removing it leaves that set as the only
 /// thing keeping a shared ghost file alive.
+///
+/// Both query entry points are overridden. `ClipsDao.isFileReferenced` happens
+/// to delegate to [referencedFilenames] today, but nothing pins that: if it
+/// ever queried directly, a single override would quietly restore the backstop
+/// and the test would pass for the wrong reason.
 class _BackstopBlindClipsDao extends ClipsDao {
   _BackstopBlindClipsDao(super.attachedDatabase);
 
   @override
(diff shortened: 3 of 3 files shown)


mbradley (commented):
Verified the fix at the layer it claims: `getClipsWithGhostFrames` has no `draft_id` predicate, so library clips do reach the scan and the guard was the only thing dropping them. Reproduced the regression both ways, including a second mutation removing the library-clip row, and it goes red for the right reason (the ghost file is gone, not a setup failure). CI green on `bf0b713`.

Two comment fixes and one question.

**The `required` rationale is backwards.** `Drafts.id` is `text()()`, non-null, so a nullable `excludeDraftId` there could never produce the `null == null` skip; it would degrade to scan-everything, which over-protects rather than over-deletes. Your PR description states this correctly ("`row.id`, a non-null primary key, so `null` could never match"), so it's just the dartdoc that drifted. Suggested replacement for `draft_storage_service.dart:665-667`:

```dart
  /// [excludeDraftId] is required: the sole caller always excludes a draft,
  /// and a nullable parameter would silently mean "scan everything".
```

**The nullable contract is documented on the wrong function.** The audio scan now has a required param and no ambiguity, and it got the new paragraph. The ghost-frame scan still takes `String?`, and across ~19 lines of dartdoc never says that null means "scan every clip, library clips included", which is the contract `clearAllDrafts` depends on and the one the bug turned on. One line on the `_referencedGhostFrameFilenames` dartdoc fixes it:

```dart
  /// A null [excludeDraftId] excludes nothing: every clip is scanned, including
  /// library clips, whose `draftId` is NULL. `clearAllDrafts` relies on that.
```

That also makes the new inline comment at `:725-726` redundant, so it can go if you want.

**Sibling Scan section.** The conclusion holds, I re-ran the sweep and found no other site with a nullable exclude on both sides. But three of the four cited files don't have a nullable exclude parameter at all: `protected_minor_inbox_gate_impl.dart:50` and `dm_reactions_dao.dart:68` are `required String`, and `videos_repository.dart:2499` compares two non-nullable strings. Only `video_editor_timeline_overlay_strips.dart:116` is a real match. The closest structural analogue in the repo, `timeline_snap_controller.dart:243` (`int? _releasedSnapMs` in a `continue` guard, safe because the operand comes from a `Set<int>`), isn't listed. Worth correcting so the sweep is reproducible from what it cites.

**Question on the test.** `_BackstopBlindClipsDao` disables the backstop only because `ClipsDao.isFileReferenced` delegates to `referencedFilenames`. Nothing pins that delegation, so if it's ever changed to query directly, this test silently stops removing the backstop and starts passing for the wrong reason, which is the one failure mode a defense-in-depth test can't tolerate. Two options: override `isFileReferenced` on the blind DAO as well, or leave the override as-is and add an assertion that pins the delegation. Which do you prefer?

Also worth checking "Code refactor" in Type of Change, since the branch carries a `refactor(drafts):` commit.

Thanks, as always.

hm21:
All three corrections are right.

**The `required` rationale.** Your reading is the correct one — `Drafts.id` is `text()()`, so `row.id == null` is always false and a nullable parameter there would degrade to scan-everything, not to the `null == null` skip. The dartdoc had drifted from what the PR description says. Took your replacement verbatim (13f852a19).

**Nullable contract on the wrong function.** Added your line to the `_referencedGhostFrameFilenames` dartdoc and dropped the now-redundant inline comment at `:725-726` (13f852a19).

**Sibling Scan.** Corrected in the PR description. You're right that `protected_minor_inbox_gate_impl.dart` and `dm_reactions_dao.dart` take `required String` and `videos_repository.dart:2499` compares two non-nullable strings, so none of them were ever candidates — citing them made the sweep unreproducible. The corrected version cites the two sites that actually have a nullable exclude operand: `video_editor_timeline_overlay_strips.dart:116` (`String? selectedItemId`, safe because `TimelineOverlayItem.id` is a `String`) and `timeline_snap_controller.dart:243` (`int? _releasedSnapMs`, safe because `snapPoints` is a `Set<int>`).

**The test.** Overriding `isFileReferenced` on the blind DAO, not pinning the delegation. The fake's contract is "nothing is referenced"; with only `referencedFilenames` overridden it honoured that through one entry point and inherited a real query on the other, which is exactly the coupling you flagged. Pinning the delegation instead would put the guard in `db_client`'s suite, far from the test that depends on it, and would still leave the fake internally inconsistent. Re-ran the mutation check with the override in place: restoring the old line fails exactly one test out of 69, and it is the new one (0f995ea76).

Also checked "Code refactor" in Type of Change.

mobile/lib/services/draft_storage_service.dart
```diff
@@ -711,14 +714,17 @@ class DraftStorageService {
...
+  /// library clips, whose `draftId` is NULL. `clearAllDrafts` relies on that.
   Future<Set<String>> _referencedGhostFrameFilenames({
     String? excludeDraftId,
   }) async {
     final clipRows = await _clipsDao.getClipsWithGhostFrames();
     final filenames = <String>{};
 
     for (final row in clipRows) {
-      if (row.draftId == excludeDraftId) continue;
+      if (excludeDraftId != null && row.draftId == excludeDraftId) continue;
```
realmeylisdev: The null branch is pinned now. The other half of this guard isn't — and can't be, through the real DAO.

`DraftsDao.deleteDraft` (`drafts_dao.dart:161-166`) deletes `clips WHERE draft_id = id` and the draft row in one transaction, and `deleteDraft` awaits it at `:602` before calling this scan at `:610`. No row with `draft_id == id` survives to be skipped. Same story for the audio scan — `:602` removes the draft row before `getAllDrafts()` runs at `:648`.

Deleting this line outright, so the scan always includes every row, leaves **456 tests green**: 69 in this file plus 387 across the other 15 suites that touch `DraftStorageService`. Deleting the `row.id == excludeDraftId` line at `:675` leaves the 69 green too. So the `required String` hardening in bf0b713 is also guarding a branch nothing can reach.

That's not a defect in the fix — the branch you fixed is the reachable one, and it's properly covered now. But this PR rewrites the line, so it seems worth deciding rather than inheriting. Options as I see them:

- Pin it with a `_StaleRowClipsDao` whose `getClipsWithGhostFrames()` returns a row still carrying the deleted draft's id. Same technique as `_BackstopBlindClipsDao`, roughly 20 lines, and the guard becomes falsifiable.
- Drop `excludeDraftId` from both scans. Dead through the real DAO, and it removes the `null == null` hazard class by construction — at the cost of the safety net if someone later moves a scan ahead of the delete.
- Move both scans before the deletion, which makes the exclusion load-bearing and testable with no fake at all. That's a reordering, so a separate PR.
- Keep it as deliberate defense and say so in the dartdoc.

Any of these works for me — I'd just rather the branch not stay unfalsifiable by accident. Which do you prefer?

mobile/lib/services/draft_storage_service.dart
```diff
@@ -711,14 +714,17 @@ class DraftStorageService {
   /// Uses [ClipsDao.getClipsWithGhostFrames] rather than reading the whole clip
   /// table: this scan runs on every draft deletion, and decoding every blob in
   /// the database made it grow with the size of the clip library (#6548).
+  ///
+  /// A null [excludeDraftId] excludes nothing: every clip is scanned, including
+  /// library clips, whose `draftId` is NULL. `clearAllDrafts` relies on that.
```
realmeylisdev: This paragraph states the contract correctly, but the call-site comment it describes still contradicts it. At `:608-609`:

> Exclude this draft's own clips so its ghost frames aren't kept alive by rows that haven't cascade-deleted yet.

There is no cascade to be pending. `tables.dart` has no `references(...)`, `KeyAction` or `onDelete` anywhere — `draft_id` is a logical foreign key only — and `DraftsDao.clearAll`'s own dartdoc spells it out: "the FK cascade does not fire — see [deleteDraft]". That is exactly why `deleteDraft` deletes the clip rows itself, transactionally, at `:602`. The rows this comment says might linger were removed by the line directly above it.

13f852a19 corrected the two scan dartdocs; this is the third contract in the same cluster, and it reads as the justification for the guard below.

realmeylisdev (commented):
Verified independently: reverting the guard fails exactly one test of 69, and it is the new one. `getClipsWithGhostFrames` has no `draft_id` predicate, `Clips.draftId` is `text().nullable()`, and `clearAll` preserves NULL-draft rows — so library clips do reach the scan and `null == null` was the only thing dropping them. Analyze, format and all 15 checks green. The backstop and per-frame scan-cost claims both hold: `_jsonFilePathKeys` still carries `ghostFramePath`, and `referencedFilenames` really does pay a full `SELECT data FROM clips` for every unresolved name.

Also re-ran the corrected Sibling Scan — both cited sites check out (`TimelineOverlayItem.id` is `String`, `snapPoints` is `Set<int>`), and `String? excludeDraftId` is now the only nullable exclusion-named parameter left in `lib` + `packages`.

Nothing blocking. Two notes inline.

mobile/lib/services/draft_storage_service.dart
hm21: Option 2 — dropped `excludeDraftId` from both scans (fbb99f10f).

Confirmed your reachability analysis before picking: deleting **both** guard lines leaves all 77 tests in `draft_storage_service_test.dart` + `file_cleanup_service_test.dart` green. `deleteDraft` removes the clip rows and the draft row in one transaction at `:602`, and both scans run after it, so nothing the exclusion could skip survives to reach them.

That ruled out `_StaleRowClipsDao`: `testing.md`'s can-it-fail bar treats a line only reachable through contorted setup as a signal of over-defensive code to fix rather than to pin, and a fake producing a row the real DAO cannot produce is exactly that. Keeping it as documented defense loses to `agent_workflow.md` #4 on dead code. Reordering is behavioural and out of scope, as you said.

The upside beyond removing dead code: the nullable-exclude hazard class that caused #6581 is gone by construction rather than by guard, and library clips reach the ghost-frame scan the same way. The tradeoff is the lost safety net if someone later moves a scan ahead of the delete — that failure mode leaks a file rather than deleting one, and the delete-first ordering is load-bearing and commented at `:599-601`.

The regression test still bites after the removal: mutating the scan to `if (row.draftId == null) continue;` fails exactly one test out of 69, and it is the new one. It now guards the behaviour directly instead of guarding the guard. All 17 suites touching `DraftStorageService`: 464 passing, 2 pre-existing skips.

mobile/lib/services/draft_storage_service.dart
hm21: Gone with the guard it justified (fbb99f10f) — the exclusion is removed from both scans, so there is no longer a comment claiming rows are pending a cascade.

One correction to the premise, since it will mislead the next reader: `tables.dart` **does** declare the FK, at `:661-662` — `FOREIGN KEY (draft_id) REFERENCES drafts(id) ON DELETE CASCADE` via `customConstraints`, mirrored in `app_database.dart:236`. Grepping for `references(` / `onDelete` / `KeyAction` finds only the Drift-DSL form and misses the raw-SQL one, so `draft_id` is a declared FK rather than a logical one.

Your conclusion holds for the adjacent reason: no `foreign_keys` pragma is set anywhere in `db_client`, so enforcement is off and the cascade never fires. That is what `DraftsDao.deleteDraft`'s dartdoc at `:155-156` already says, and why it deletes the clip rows itself.

mbradley (approved):
Re-verified at `fbb99f1`. Reintroducing the guard fails exactly 1 of 77 tests and it is the new one, so the regression test still bites after the parameter removal. I also traced the exclusion back to where it came from: `86d88cedb` added it with the "haven't cascade-deleted yet" comment in the same commit that established delete-first ordering, so it was defending against a state that was never reachable. That is a firmer basis for dropping it than the absence of coverage.

Your FK correction checks out both ways. The `customConstraints` declaration is real at `tables.dart:661-662`, and there is no `foreign_keys` pragma anywhere in `db_client`, so it never fires.

On the blind-DAO question from last round: I removed the `isFileReferenced` override and reintroduced the bug, and the test still fails, so the override is not load-bearing today. Keeping it is still the right call, since it guards the delegation coupling rather than the current outcome.

One thing the removal left behind, one package over. `clips_dao.dart:109-110` still documents the projection as carrying `draftId` "to skip the draft being deleted", which is the skip this PR removes. Nothing reads it now: the scan's loop uses only `row.data` and `row.id`, and `getClipsWithGhostFrames` has no other caller. GitHub will not let me attach an applyable suggestion there because the file is not in this PR's diff, so the replacement text is below.

```dart
  /// Every clip row that carries a non-null `ghostFramePath`, projected to the
  /// fields a ghost-frame reference scan needs: `data` to read the frame out
  /// of, and `id` to name the row if its blob turns out to be corrupt.
  ///
  /// `draftId` is projected but no longer read: the scan stopped excluding the
  /// draft being deleted (#6582), since callers run it after the delete and
  /// every surviving row is a legitimate reference.
```

Approving. The dartdoc is a comment in an adjacent file and should not hold the merge, so pick it up here or in a follow-up, your call.

Thanks, as always.

realmeylisdev (approved):
Approving. One fix pushed; the rest holds up.

**Checked**

- The mechanism is exactly as described. `Clips.draftId` is `text().nullable()` (`tables.dart:626`), `getClipsWithGhostFrames` carries no `draft_id` predicate (`clips_dao.dart:131`), and `clearAll` keeps only NULL-draft rows (`drafts_dao.dart:252`) — so `null == null` dropped precisely the rows the call exists to protect.
- Deadness holds at the source, not just by test. `deleteDraft` commits both deletes in one transaction (`drafts_dao.dart:161-166`), awaited at `:602` before the scans at `:609` / `:645`, and drift awaits `transaction.complete()` in a `finally` before the future resolves (`connection_user.dart:526-536`) — so the rows are gone, not pending. Every clip write sets `draftId` (`drafts_dao.dart:347`), and the only later mutation is `softDeleteClip(clearDraftId: true)` → NULL. A survivor is either deleted-by-id or NULL; neither was ever skippable.
- Your numbers reproduce exactly. Re-adding `if (row.draftId == null) continue;` → 76/1; restoring main's copy of the file wholesale → 76/1; both fail only the new test. 17 suites = 464 passing / 2 pre-existing skips. Analyze and format clean.
- Worth noting: deleting the `referencedGhostFrameFilenames` short-circuit from `FileCleanupService` → 68/1, the same test. `file_cleanup_service_test.dart` stays 8/8 without it, so your new test is now the only thing guarding that branch.
- No cross-repo or protocol surface. A ghost frame is a device-local onion-skin JPEG in the documents dir — never uploaded, never in an `imeta` tag, absent from every NIP.

**Pushed** — `eeeb5bfc9`

`getClipsWithGhostFrames`'s dartdoc still justified the projected `draftId` as the field "to skip the draft being deleted". Nothing skips by it any more, and left standing it invites the exact nullable-compare shape #6581 came from. Reworded, with the delete-first ordering as the reason.

I kept the field itself: `clips_dao_test.dart:514` asserts a library clip comes back with a NULL `draftId`, which pins the contract this fix depends on. Dropping it would trade that guard for tidiness — your call if you'd rather.

DB Client CI now runs on this PR and passes (43.4% against the 40% gate).

Issue #6436, opened 2026-07-27
fix(blocklist): blockUser and unblockUser discard the publish result their caller needs


realmeylisdev:
Found while root-causing #6435.

## Problem

`_publishMuteListToNostr` and `_publishBlockListToNostr` in `mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart` publish the kind-10000 mute list and kind-30000 block list without awaiting a NIP-20 `OK`. Frame-accepted is treated as published.

`blockUser` (`:845`) and `unblockUser` (`:886`) both call them and both return `void`. Every caller therefore reports success unconditionally:

- `lib/blocs/other_profile/other_profile_bloc.dart:372`
- `lib/blocs/dm/conversation_actions/conversation_actions_cubit.dart:108`
- `lib/blocs/safety_settings/safety_settings_cubit.dart:161`
- `lib/screens/inbox/conversation/conversation_view.dart:110`

There is no outbox, retry timer, or pending-publish queue anywhere in the package.

## Why it matters

The local set is updated and persisted immediately, so the UI shows the block as lifted while the relay may still hold the old list. The user has no way to tell the difference, and no path recovers it later.

This is not hypothetical: the session that produced #6435 shows sustained relay failure on the sending device — repeated `[wss://relay.divine.video] Connection timed out after 0:00:10.000000`, `Connection failed, cannot send`, `relay connect fail!`, one `NostrService` init taking 54197ms, and a Keycast RPC sign timing out after 30s. Under exactly those conditions a block or unblock can be locally applied and never published, with a success snackbar.

Blocking is a safety action. Silently failing to publish one is worse than failing loudly.

## Suggested direction

Await the OK the way DM sends already do — `publishEventAwaitOk` and the `PublishTracker` in `nostr_client` — and surface an unconfirmed publish to the caller so the UI can show a retry affordance. `dm_repository`'s `OutgoingDmRetryService` is the existing precedent for a queue-and-retry arm if one is wanted here.

## Related

- Discovered during #6435 / #6434
- Sibling defects from the same audit: #6437, #6438

NotThatKindOfDrLiz:
**Rescoped.** Most of this is now fixed; retitled to the part that isn't.

## What has since shipped

This was written when `_publishMuteListToNostr` returned `void` and there was "no outbox, retry timer, or pending-publish queue anywhere in the package." All three now exist:

[...]

PR #3177, opened 2026-04-20
feat(nostr): reliable publish primitives + NIP-09 deletion reference migration

rabble:
Closes #3178

First PR in an 8-PR series to make every Nostr publish in divine-mobile durable. See `docs/superpowers/plans/2026-04-20-reliable-nostr-publish-README.md` for the full plan.

## Summary

- Adds `PublishOutcome`, `NostrClient.publishEventAwaitOk`, `NostrClient.publishEventWithRetry` (bounded exponential backoff, retries only transient relays per NIP-01) — the primitives every follow-up PR uses.
- Adds `PublishResultMapper` so every future service migration maps outcome → user feedback uniformly (success / retryable error / permanent reject with reason).
- Extends `NostrClientStatisticsObserver` with `onPublishOutcome` for per-attempt publish telemetry.
- Migrates `ContentDeletionService` as the reference implementation; `_deleteVideo` in `share_video_menu.dart` now shows a retry-able snackbar on transient failure.
- **Contract change for deletion:** the previous "save locally even if publish fails" branch is removed. Durable deletion now requires at least one relay accept — fixes the ghost-delete bug where the UI believed content was deleted while every other client could still see it.
- Merges cleanly with PR #3117's \`DeleteFailureKind\` localization work — \`DeleteResult\` now carries both \`failureKind\` (pre-publish) and \`outcome\`/\`feedback\` (post-publish).

## Test plan

- [x] `flutter test mobile/test/services/content_deletion_service_test.dart` — all 10 existing tests pass against the new `publishEventWithRetry` contract.
- [x] `flutter test mobile/test/unit/services/content_deletion_service_reliability_test.dart` — 3 new tests (success / transient / permanent reject).
- [x] `flutter test mobile/packages/nostr_sdk` — 241+ tests including 7 `PublishOutcome` + 5 `RelayPool.sendAwaitOk` tests.
- [x] `flutter test mobile/packages/nostr_client` — 11 new tests (5 `RetryPolicy` + 6 `PublishResultMapper`).
- [x] `flutter analyze` clean on all changed files.
- [ ] Manually delete a video on a staging relay that drops OK frames — expect retry snackbar with Retry action.
- [ ] Manually delete on a relay that returns \`blocked:\` — expect non-retryable snackbar with the reason.

## Architecture

Four layers, no new top-level packages:

[...]

Issue #3174, opened 2026-04-20
feat(auth): publish bootstrap kind:10002 when NIP-65 discovery returns empty


realmeylisdev:
## Summary
Client-side mitigation for the missing-kind:10002 gap that contributes to #3152, #3161, #3162. The primary fix is server-side in divinevideo/keycast#94 (publish at account provisioning). This issue tracks a complementary mobile fallback that also backfills existing affected accounts.

## Problem
`RelayDiscoveryService` queries `purplepag.es`, `user.kindpag.es`, `relay.nos.social` for the signed-in user's kind:10002 on every login. For every Keycast-provisioned Divine account, all three indexers return empty (`⚠️ No relay list found for <npub> on any indexer`). That warning is present in all three recent Zendesk reports.

On post-#2955 builds, `AuthService._connectToFallbackRelays()` mitigates reachability, but it:
- does not surface the user's real outbox/inbox to other Nostr clients
- does not help interop (an Amethyst/Damus user looking up a Divine user's relays still sees nothing)
- does not cover existing Keycast accounts that were provisioned before the server-side fix (keycast#94)

## Proposed behavior
Extend `AuthService._discoverUserRelays`'s empty-result branch:

1. After `_connectToFallbackRelays()`, publish a bootstrap kind:10002 on behalf of the user with:
   ```
   ["r", "wss://relay.divine.video"]
   ```
   (and optionally the safeFallbackRelays marked as `read`).
2. Broadcast to `IndexerRelayConfig.defaultIndexers` + `wss://relay.divine.video` via `tempRelays`.
3. Gate on a per-npub `SharedPreferences` flag (`bootstrap_kind10002_published_<npub>`) so it publishes exactly once per account per device.
4. Only run when the account has a live signer (not for read-only/imported sessions that can't sign).

## Acceptance criteria
- [ ] On first login for a Divine Keycast account whose indexer query returns empty, a valid kind:10002 is published and becomes discoverable on at least one of the three indexers within N seconds
- [ ] The bootstrap does not fire on subsequent logins for the same account
- [ ] The bootstrap does not fire when discovery *succeeds* (real list exists, even if we discovered it later via cache)
- [ ] The bootstrap does not run for sessions without a working signer
- [ ] If the user later edits their relay list via another client, the newer kind:10002 wins (replaceable-event semantics — nothing to enforce client-side)
[...]

PR #6583, opened 2026-08-01
fix(nostr): make publishEventAwaitOk success semantics explicit

### LINKED ISSUE #3178: feat(nostr): reliable publish primitives + NIP-09 deletion reference migration
Auto-created from PR #3177.

**PR:** https://github.com/divinevideo/divine-mobile/pull/3177
**Branch:** `feat/reliable-nostr-publish`

### LINKED ISSUE #3366: fix(nostr): make publishEventAwaitOk success semantics explicit
Source audit context: fresh `origin/main` worktree `.worktrees/audit-codebase-2026-04-24` at commit `33f6e026a`.
This issue was generated from the April 2026 codebase audit and includes file/line evidence from that snapshot.

## Severity
High

## Problem
publishEventAwaitOk is documented as reliable for critical operations, but the underlying tracker completes on the first OK true. A critical publish can be considered successful while other targeted relays are still pending and later reject or never respond.

## Evidence
- mobile/packages/nostr_client/lib/src/nostr_client.dart:297 says it waits for an OK from at least one relay.
- mobile/packages/nostr_client/lib/src/nostr_client.dart:312 recommends it for deletions and critical profile updates.
- mobile/packages/nostr_sdk/lib/nostr.dart:204 says it awaits OK confirmation.
- mobile/packages/nostr_sdk/lib/nostr.dart:208 says future completes early once one relay confirms.
- mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart:85 handles accepted relay.
- mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart:89 says first confirmation is enough.
- mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart:91 completes immediately.
- mobile/packages/nostr_sdk/lib/relay/relay_pool.dart:940 documents first accept/every response/timeout completion.

## Impact
Callers may interpret confirmed as published everywhere requested when it only means one relay accepted. This matters for deletion, profile updates, moderation, and user-facing success states.

## Recommended fix
Rename or document the current primitive as first-accept and add separate options/primitives for all-targets, quorum, or minimum accepted relays.

## Acceptance criteria
- [ ] Public docs state exactly when the future completes.
- [ ] PublishOutcome exposes enough status for callers to distinguish partial success.
- [ ] Critical callers use the intended durability policy.
- [ ] Tests cover one accept plus later reject/no-response behavior.

## Related issues
Related: #3178, #3191, #3189, #3187, #3180.

realmeylisdev:
## Description

`publishEventAwaitOk` is documented as the primitive for "deletions, critical profile updates", but its tracker completed on the **first** relay response and then hard-closed, discarding everything after. A caller could not tell *"1 of 5 relays accepted, 4 refused"* from *"5 of 5 accepted"* — and the relay that refused was recorded as **silent** rather than **rejecting**.

This is not a corner case. Divine accounts are provisioned without a published relay list, so NIP-65 discovery misses, the four third-party fallback relays (`damus.io`, `nos.lol`, `primal.net`, `nos.social`) are added through the *persisting* path, and nothing ever removes them. Deletions publish with no `targetRelays`, so a normal user's delete fans out to ~5 relays and whichever answers first decides what they are told.

Reproduced against the live relay before the fix — one permissive relay and one real `funnelcake-relay`, same kind-5 deletion:

```
B permissive          OK=true                                              (+10.1ms)
A funnelcake (real)   OK=false  "blocked: delete targets non-existent events"  (+15.0ms)
```

The permissive relay won by 5ms, so the user was told "deleted" while the relay the app actually reads from had explicitly refused. And because `ContentDeletionService` picks its failure message by whether `rejectedBy` is empty, that branch could never observe the refusal.

Same scenario on this branch:

```
accepted:  [ws://127.0.0.1:47811]
rejected:  {ws://localhost:47777: blocked: delete targets non-existent events}
acceptedByAll=false  acceptedByPartial=true  summary="accepted by 1 of 2 relays"
```

### What changed

Four defects sat at one site:

* **First-accept hard close.** `onAccepted` called `_complete()`, which set `_closed` and made every later frame a no-op. Completion now waits for every relay the fan-out reached, bounded by a one-second settle window after the first answer so a single wedged relay cannot stall a publish for the full timeout. Healthy relays answer in single-digit milliseconds, so the window costs effectively nothing — and the caller already blocked on the full sequential fan-out anyway, so first-accept was never actually buying latency.
* **The denominator was built backwards.** The tracker was constructed with an *empty* `expectedRelays`, filled incrementally from send results during a sequential fan-out. `0 >= 0` let any early signal close an un-armed tracker, and a target the pool failed to write to vanished from the outcome entirely. The target set is now resolved *before* the fan-out starts, and unreachable targets are reported instead of disappearing.
* **No membership check.** OK frames were routed by event id alone, so any connected relay could accept or reject a publish it was never sent. For a full-pool delete a hostile relay holds the event id anyway, so this is primarily a correctness fix; it matters for narrowly targeted publishes such as DMs.
* **`cancel()` snapshotted differently** from a normal completion, so a partial-accept shutdown could list the same relay as both accepted and unanswered. It now shares one code path.

Deleting a video now requires every relay it reached to accept, and reports the new `relayPartial` state when only some do — true and actionable, since republishing the same kind 5 is safe (a relay that already has it answers `OK true` again). A delete that is not fully confirmed is still not persisted locally, which keeps the video visible and the retry available.

Also fixes the relay-list bootstrap, which set its one-shot "already published" flag on a **WebSocket write** rather than a relay acceptance. That is *why* users end up pinned to the 5-relay fallback: if the indexers silently drop the kind 10002, the flag is set anyway and the bootstrap never retries on that device.

### What this does NOT give you

**Acceptance is not durability, and nothing here changes that.** NIP-01 defines `OK true` as "accepted by the relay" — its own example of a valid `OK true` carries the reason `duplicate: already have this event`. Divine's relay acknowledges from an in-memory queue and commits to ClickHouse on a batch interval afterwards; a failed insert produces no negative acknowledgement of any kind. Even `acceptedByAll` is a statement about **breadth of acceptance**. The docs now say so, and nothing should present it to a user as "saved".

There is also no repair path: no federation, no deployed relay-sync, NIP-77 off in production, and no client outbox. A relay that did not accept never gets the event. That is exactly why a partial publish is surfaced rather than smoothed over.

**Related Issue:** Closes #3366, Closes #3178

## Out of Scope

- **#6215** (account deletion) and **#6436** (blocklist) — this PR gives them the primitive they need but does not migrate them. Both still report success on a socket write.
- The 34 fire-and-forget `publishEvent` call sites, of which at least 11 turn a frame-accept into user-visible or persisted success. The correctly fire-and-forget set (view events, feed tuning, NIP-17 self-wrap) stays as-is.
- `video_event_publisher` keeps its deliberate opt-out from OK-confirmation; that decision was made on purpose with a stated rationale and is not reversed here as a side effect.
- `retry_policy.dart` from the old #3177 branch ports cleanly but has no consumer, so it is deliberately not brought over as unused foundation.
- Four adjacent defects found while investigating are filed separately rather than folded in.

## Verification

- `nostr_sdk` 412 tests, `nostr_client` 297, `dm_repository` 551 — all green
- All 16 app test files touching `PublishOutcome` / `publishEventAwaitOk` — 197 tests green
- `test/l10n/arb_consistency_test.dart` green; new key translated into all 22 locales
- `check_untested_services_floor.sh` and `check_test_unit_structure.sh` — no new entries
- End-to-end against a live local `funnelcake-relay` plus a permissive stand-in, output quoted above

Tests are anchored on a **partition invariant** — every targeted relay appears in exactly one of `acceptedBy` / `rejectedBy` / `noResponseFrom` / `unreachableTargets`, and no other relay appears at all. That single assertion catches all three outcome-shaping defects, and is the test that would have caught this in April 2026.

Note `publish_outcome_test.dart`'s old first test was **deleted rather than rewritten**: it asserted "completes immediately once any relay confirms, leaving the rest in noResponseFrom" — it pinned the bug as intended behaviour.

## Type of Change

### DIFF
--- a/mobile/lib/services/content_deletion_service.dart
+++ b/mobile/lib/services/content_deletion_service.dart
@@ -34,6 +34,15 @@ enum DeleteFailureKind {
   /// timeout. The delete is NOT persisted locally so the user can retry.
   relayNoResponse,
 
+  /// Some relays accepted the delete event and some did not, so the video is
+  /// gone from part of the network and still live on the rest.
+  ///
+  /// Nothing republishes to the relays that did not accept, so this does not
+  /// resolve itself. The delete is NOT persisted locally, which keeps the
+  /// video visible and the retry available; republishing the same kind 5 is
+  /// safe because a relay that already has it answers `OK true` again.
+  relayPartial,
+
   /// Unexpected error (including outer [deleteContent] catch).
   unknown,
 }
@@ -222,16 +231,39 @@ class ContentDeletionService {
         deleteEvent,
       );
 
-      if (publishOutcome.failed) {
-        Log.error(
-          'Delete publish not confirmed by any relay: '
-          '${publishOutcome.summary}',
-          name: 'ContentDeletionService',
-          category: LogCategory.system,
-        );
-        final failureKind = publishOutcome.rejectedBy.isNotEmpty
-            ? DeleteFailureKind.relayRejected
-            : DeleteFailureKind.relayNoResponse;
+      // Counts only — never relay URLs — so this is safe to keep on in every
+      // build and still answers "how often is a delete only partly accepted?".
+      final breadth =
+          'accepted=${publishOutcome.acceptedBy.length}/'
+          '${publishOutcome.targetCount} '
+          'rejected=${publishOutcome.rejectedBy.length} '
+          'silent=${publishOutcome.noResponseFrom.length} '
+          'unreachable=${publishOutcome.unreachableTargets.length}';
+
+      // A delete is only done when every relay we reached took it. Anything
+      // less leaves the video live somewhere, and nothing in this client
+      // republishes to the relays that refused.
+      if (!publishOutcome.acceptedByAll) {
+        final failureKind = switch (publishOutcome) {
+          final o when o.acceptedByPartial => DeleteFailureKind.relayPartial,
+          final o when o.rejectedBy.isNotEmpty =>
+            DeleteFailureKind.relayRejected,
+          _ => DeleteFailureKind.relayNoResponse,
+        };
+        if (publishOutcome.acceptedByPartial) {
+          Log.warning(
+            'Delete only partly accepted: $breadth',
+            name: 'ContentDeletionService',
+            category: LogCategory.system,
+          );
+        } else {
+          Log.error(
+            'Delete publish not confirmed by any relay: '
+            '${publishOutcome.summary}',
+            name: 'ContentDeletionService',
+            category: LogCategory.system,
+          );
+        }
         return DeleteResult.failure(
           'Relay did not confirm deletion: ${publishOutcome.summary}',
           failureKind,
@@ -239,7 +271,7 @@ class ContentDeletionService {
       }
 
       Log.info(
-        'Delete request confirmed by relay(s): ${publishOutcome.acceptedBy}',
+        'Delete request confirmed by every relay: $breadth',
         name: 'ContentDeletionService',
         category: LogCategory.system,
       );
@@ -314,6 +346,8 @@ class ContentDeletionService {
         return 'Relay rejected deletion';
       case DeleteFailureKind.relayNoResponse:
         return 'Relay did not confirm deletion';
+      case DeleteFailureKind.relayPartial:
+        return 'Only some relays accepted the deletion';
       case DeleteFailureKind.unknown:
         return 'Failed to delete content';
     }
--- a/mobile/lib/services/content_deletion_service.dart
+++ b/mobile/lib/services/content_deletion_service.dart
@@ -184,10 +184,11 @@ class ContentDeletionService {
 
   /// Delete user's own content using NIP-09.
   ///
-  /// The deletion is only considered successful when at least one relay
-  /// returns an `OK true` acknowledgement (NIP-20). If every relay rejects
-  /// the event, or none respond before the publish timeout, the operation
-  /// fails and is NOT added to local deletion history — the caller can retry.
+  /// The deletion is only considered successful when every relay the publish
+  /// reached returns an `OK true` acknowledgement (NIP-20). A rejection,
+  /// silence, or a target the publish could not write to all fail the
+  /// operation, and a failed deletion is NOT added to local deletion history
+  /// — the video stays visible and the caller can retry.
   Future<DeleteResult> deleteContent({
     required VideoEvent video,
     required String reason,
--- a/mobile/lib/services/content_deletion_service.dart
+++ b/mobile/lib/services/content_deletion_service.dart
@@ -34,17 +34,26 @@ enum DeleteFailureKind {
   /// timeout. The delete is NOT persisted locally so the user can retry.
   relayNoResponse,
 
-  /// Some relays accepted the delete event and some did not, so the video is
-  /// gone from part of the network and still live on the rest.
-  ///
-  /// Nothing republishes to the relays that did not accept, so this does not
-  /// resolve itself. The delete is NOT persisted locally, which keeps the
-  /// video visible and the retry available; republishing the same kind 5 is
-  /// safe because a relay that already has it answers `OK true` again.
-  relayPartial,
-
   /// Unexpected error (including outer [deleteContent] catch).
   unknown,
+}
+
+/// How broadly the relay network took a delete request that succeeded.
+///
+/// Set on every successful [DeleteResult] so callers can tell the user what
+/// actually happened. Deletion under NIP-09 is a *request* relays SHOULD
+/// honour, so neither value is a promise the video is gone from the network.
+enum DeleteAcceptance {
+  /// Every relay the publish targeted answered `OK true`.
+  everyRelay,
+
+  /// At least one relay accepted, and at least one rejected the request,
+  /// stayed silent, or was never reached.
+  ///
+  /// The video is out of this device's library and the tombstone is recorded,
+  /// but the relays that did not accept can still serve it. Nothing here
+  /// republishes to them.
+  someRelays,
 }
 
 /// Delete request result
@@ -56,6 +65,7 @@ class DeleteResult {
     this.error,
     this.deleteEventId,
     this.failureKind,
+    this.acceptance,
   });
   final bool success;
   final String? error;
@@ -65,9 +75,16 @@ class DeleteResult {
   /// Set when [success] is false; use for localized UI messages.
   final DeleteFailureKind? failureKind;
 
-  static DeleteResult createSuccess(String deleteEventId) => DeleteResult(
+  /// Set when [success] is true; how broadly the relays took the request.
+  final DeleteAcceptance? acceptance;
+
+  static DeleteResult createSuccess(
+    String deleteEventId, {
+    required DeleteAcceptance acceptance,
+  }) => DeleteResult(
     success: true,
     deleteEventId: deleteEventId,
+    acceptance: acceptance,
     timestamp: DateTime.now(),
   );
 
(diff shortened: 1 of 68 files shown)


mobile/lib/services/content_deletion_service.dart
```diff
@@ -222,24 +232,47 @@ class ContentDeletionService {
...
+          'accepted=${publishOutcome.acceptedBy.length}/'
+          '${publishOutcome.targetCount} '
+          'rejected=${publishOutcome.rejectedBy.length} '
+          'silent=${publishOutcome.noResponseFrom.length} '
+          'unreachable=${publishOutcome.unreachableTargets.length}';
+
+      // A delete is only done when every relay we reached took it. Anything
+      // less leaves the video live somewhere, and nothing in this client
+      // republishes to the relays that refused.
+      if (!publishOutcome.acceptedByAll) {
```
hm21: `acceptedByAll` across a ~5-relay fan-out, with no republish path and no local persist on failure, means one relay that permanently refuses or sits wedged makes the video undeletable — every retry hits the same relay and gets the same answer. The old behaviour lied about what happened, but the user did get the video out of their library.

Gating on "no explicit rejections" and letting silent/unreachable relays through would keep the honest reporting without the dead end. What's the thinking?

hm21:
Pushed three commits to this branch (shared-ownership takeover — revert anything you disagree with).

**`4bdb8f0` — settle window could close mid-fan-out.** The tracker is registered before `_sendCollect` runs, so a connected relay answers while a later target hasn't been written to yet. `_sendCollect` can outlast the 1s window (it waits out a relay's `connecting` state and caps each relay at `perRelaySendTimeout`), and when it did, `setReachable` found the tracker closed: targets the publish never reached were reported as silent rather than unreachable, `acceptedByAll` went false, and `_repairSilentRelays` force-reconnected relays the publish hadn't touched. A delete both relays would have accepted came back as `relayPartial` and unpersisted. Now gated on the fan-out having reported, and armed from `setReachable` for answers that landed while it ran.

**`67ef0cb` — `_maybeComplete` counted answers against the wrong set.** A relay outside `sentTo` can still answer (its frame goes out just before the per-relay send timeout fires). `setReachable` then drops it from the awaitable set while its answer stays counted, filling the quota for a relay that is still silent. Now tests whether every awaited relay answered instead of comparing totals.

**`28dc94a`** — `deleteContent`'s doc still promised success on "at least one relay".

Both fixes have a regression test; each fails on `5664ac7` and passes after. Two existing tracker tests now call `setReachable` before asserting, which is how `RelayPool` actually drives them.

Validation: `flutter analyze lib test integration_test` clean; `nostr_sdk` 414, `nostr_client` 297, `dm_repository` 551, `content_deletion_service` + `account_deletion_service` 42 — all green.

Separate review thread open on the `acceptedByAll` gate, which is your call, not mine.

hm21 (commented):
One judgment call left — the rest is pushed on your branch.

mobile/lib/services/content_deletion_service.dart
realmeylisdev: You're right, and it's worse than one wedged relay — I went looking for how often `acceptedByAll` is actually reachable and the answer is "often not at all". Changed in c27fa9d34.

The target set is `_intendedPublishTargets`, which walks the whole pool. `RelayPool.add` inserts a relay *before* `connect()` and leaves it in `_relays` when the connect fails (relay_pool.dart:332-337, 373-375), and `_checkRelayHealth` never removes a zombie — so a relay the client has never once talked to stays a target forever and lands in `unreachableTargets` on every attempt. With the default pool that is `relay.divine.video` plus the four `safeFallbackRelays`, which every Keycast-provisioned account gets because NIP-65 discovery misses and the fallback set is added through the persisting path.

`relay.damus.io` is in that set, and it refused the WebSocket upgrade with HTTP 503 on 9 of 13 real connect attempts while I was writing this — which is the same flakiness `IndexerRelayConfig` already documents as the reason it was dropped from `defaultIndexers`. So for a stretch of that afternoon no user could delete a video at all, and every retry hit the same relay for the same reason.

On the gating itself I ended up somewhere different from "no explicit rejections", because I think that has the polarity backwards: an explicit `OK false` is the one answer that never self-heals, while silence and unreachability are the transient ones. It would have unblocked the damus.io case and left a permanently-refusing relay as a permanent dead end.

What the honesty and the dead end were actually fighting over is that one decision was doing two jobs. So they're split:

- **Local record** — gated on `acceptedByAny`. NIP-09 calls this a deletion *request* that relays SHOULD honour, and explicitly contemplates clients telling the user it isn't guaranteed, so all-relay acceptance was buying a guarantee that doesn't exist at any breadth. `AccountDeletionService` already gates the same kind 5 on `outcome.confirmed` and reports incompleteness through `deleteAccountSuccessContentUnverified`; `shareMenuDeleteConfirmation` already promises removal "from Divine" while warning it may remain elsewhere. This is the shape both already had.
- **What the user is told** — `DeleteResult.acceptance`. A partly-accepted delete now says "Deleted. Not every relay confirmed, so it may still show up in other apps." instead of either claiming it's gone everywhere or refusing to remove it. That's the part that was genuinely missing before this PR, and it survives.

`DeleteFailureKind.relayPartial` is gone since it can't occur, and its never-shipped ARB key was swapped across all 22 locales.

Reviewing your three commits turned up four more things on the branch, all in code this PR added — pushed separately, details in a comment on the PR.

realmeylisdev:
@hm21 thanks — took all three, and answered the `acceptedByAll` thread in place (`c27fa9d34`): the gate moves to `acceptedByAny` and the breadth becomes something the user is *told* rather than something that blocks them. The measurement that decided it is in the thread.

Reviewing your commits turned up four more defects, all in code this PR added. Each has a regression test that is red on `28dc94a63` and green after.

**`dd52fd13d` — a relay that answered was filed as never reached.** `onRelayUnavailable` moved the relay into `unreachableTargets`, whose contract is "these never received the EVENT frame", and `_complete` says outright that a relay which answered is never unreachable. Its only caller is `_retryAuthRequiredPublishesForRelay`, which by construction only runs for a relay that already sent `OK false auth-required` — so it was wrong on every call, and it wiped the reason on the way out, leaving the caller to retry the network instead of the auth. Now reports the rejection, matching the expired-deadline branch above it. That leaves `onRelayUnavailable` with no callers; removing it also removes the only writer of `_reachable` outside `setReachable`, which could not intersect — so a relay dropped mid-fan-out was silently put back into the awaited set by the report that followed.

**`80eee09bf` — the hard timeout still beat `setReachable`.** `_fanOutReported` closed the settle-window race, but not the other one: `_sendCollect` is bounded by the *same* deadline as the tracker's timer, which is armed first and wins the tie. `setReachable` then hit `if (_closed) return`, `_reachable` stayed null, and every target the fan-out broke past without writing to was reported as silent — handed straight to `_repairSilentRelays`, which force-reconnects them. That is verbatim the failure `4bdb8f0`'s message describes as fixed. The timeout now hands over to `setReachable`, bounded by `fanOutReportGrace` so a driver that never reports can't hang the publish, and `RelayPool` reports in a `finally` so a throwing fan-out still settles.

**`3e8cc1b1e` — the bootstrap gate accepted the wrong relay.** Unrelated to your commits, this is mine in `5664ac75f`. The gate is `acceptedByAny` over `[_primaryRelayUrl, ...defaultIndexers]`, but `RelayDiscoveryService` only ever queries the indexers — so the Divine relay accepting satisfied it, and the caller then wrote the permanent `bootstrap_kind10002_*` flag. Every later login runs discovery, finds nothing, falls back: the #3174 state, now unrecoverable. The method's own doc already said "one indexer is enough"; the code now agrees with it.

Two things I looked at and deliberately left:

- **`_repairSilentRelays` can still cycle a healthy relay whose OK lands after the 1s settle window.** Real, but `isSilentSince` excludes any relay that received an inbound frame of any kind since the publish, which on a client carrying subscriptions is most of them, and `silentRepairCooldown` caps it at once per minute per relay. `80eee09bf` removes the part that mattered — relays the publish never wrote to.
- **The settle window can expire during a NIP-42 retry**, dropping the `_authRequiredPublishRetries` entry so the relay never gets the event. Not a regression: on `main` the tracker settled on the first OK, so it lost that retry sooner. Worth its own change, not this one.

Also filed #6606 for the NIP-09 rebroadcast this whole area keeps running into: "clients SHOULD broadcast deletion request events to other relays which don't have it", and we have no path that does.

One thing this does **not** fix, worth saying plainly: #6547 is an open report of the same feature failing, and its log is `rejected by wss://relay.divine.video: blocked: delete not authorized`. Nothing accepted, so that delete correctly fails both before and after this change — the refusal is relay-side and needs its own look. This PR only changes what happens when something *did* accept.

Validation: `flutter analyze lib test integration_test` clean; `nostr_sdk` 415, `nostr_client` 297, `dm_repository` 551, and 328 app-side across the deletion, bootstrap, l10n and delete-call-site suites — all green.

rabble (approved):
Reviewed current head 3e8cc1b1e114ee1b4e2fb6e8661a9a784b9ac46c, focusing on PublishTracker partitioning, fan-out timeout handoff, target filtering, auth-required retries, bootstrap acceptance, and partial-delete persistence. The added regression coverage matches those boundaries, all current GitHub checks are green, and git diff --check is clean.

PR #4944, opened 2026-06-07
feat(dm): route NIP-17 DMs to kind-10050 inbox relays + harden moderation report DM (#570)

realmeylisdev:
## Summary

Second slice of the #570 (NIP-17 moderation DM) work — the **moderation-communication** pieces that don't depend on backend/ops. Independent of the security-crypto PR (#4943); targets `main`. Three delineated commits.

### 1. Route NIP-17 sends to the recipient's kind-10050 DM inbox relays
Per NIP-17, a gift wrap **MUST** be published to the recipient's kind-10050 "DM relays list", or a Nostr user who only reads their own inbox relays never receives it (the TC-SOCIAL-025 gap). `EventKind.dmRelaysList`; `DmRepository.resolveDmInboxRelays` (newest/dedup, `null`→default-pool fallback); `targetRelays` threaded through `sendRumor`→`publishEvent` (conditional shape preserves existing stubs); resolved after the queue enqueue so the optimistic echo isn't delayed.

### 2. Harden the moderation report DM (C8 + C9)
- **C8**: `skipNip04Fallback: true` — moderation reports never degrade to a metadata-leaking NIP-04 plaintext duplicate.
- **C9**: a failed moderation DM was swallowed; now surfaced as a calm, non-blocking notice (`reportModerationDmDelayed`). The report still succeeds via relay + Zendesk.

### 3. "Message the moderation team" entry point
A secondary button on the report confirmation screen deep-links into the NIP-17 conversation with the moderation account (reusing the minor-account-review nav pattern); hidden when signed out.

## Test plan
- [x] `dm_repository` 279, `nostr_sdk` 275, `report_content_dialog` 33 widget tests, `flutter analyze` clean, `gen-l10n` regenerated.

## Scope / follow-ups
- **No behavior change** until recipients advertise kind-10050 (the routing falls back to broadcast otherwise — universal today, see ops #4946).
- **Deferred / decisions:** C10 moderation-pubkey pin is a security/rotation tradeoff held for sign-off (#4948).
- **Backend/ops:** divine-moderation-service#169 (server-side kind-10050 + Zendesk bridge), #4946 (ops: publish moderation@ kind-10050, relay accept kind-10050, Keycast), #4947 (operator surface).

Refs #570

PR #2118, opened 2026-03-10
feat(dm): add NIP-17 DM data layer, repository, and business logic

realmeylisdev:
## Summary
Full NIP-17 direct messaging backend — protocol, data, repository, BLoCs, and service refactor. **Zero user-visible changes**; Notifications tab and routing remain unchanged.

- **Protocol**: Gift-wrap sender verification, relay subscription/AUTH/EOSE fixes, AES-256-GCM file encryption (Kind 15), `hasSubscriptionById` for accurate EOSE tracking
- **Data**: Drift `DirectMessages` + `Conversations` tables with DAOs, `DmMessage`/`DmConversation` domain models
- **Repository**: `DmRepository` — receive, decrypt, persist, send, and query encrypted DMs with deduplication, polling fallback, and health checks
- **BLoCs**: `ConversationListBloc` (restartable/droppable) + `ConversationBloc` (restartable/sequential) + `DmUnreadCountCubit`
- **Service**: Refactored `NIP17MessageService` to accept any `NostrSigner` (local keys, Keycast RPC, Amber) instead of `NostrKeyManager` only
- **Init**: KeyManager re-initialization after AuthService to pick up restored keys

**Part of [#570 Direct Messaging (NIP-17)](https://github.com/divinevideo/divine-mobile/issues/570) — Core DM stack (data + logic only)**

Split from #2110 to reduce review scope. PR B (#2110, retargeted here) will add the inbox UI, routing, and tab rename.

## Architecture
```
DmUnreadCountCubit (wired in main.dart, emits 0 until UI reads it)
    ↓
ConversationListBloc / ConversationBloc
    ↓
DmRepository (receive, decrypt, persist, send, query)
    ↓
NIP17MessageService + NostrClient (.signer)
    ↓
GiftWrapUtil, RelayPool, FileEncryption, EventKind
    ↓
DirectMessages + Conversations tables/DAOs
DmMessage + DmConversation domain models
```

## Files changed (46)
- 28 new files + 18 modified across protocol, data, repository, BLoC, provider, theme, and test layers
- ~5k of the diff is Drift `app_database.g.dart` regeneration from adding 2 tables
- ~4.5k is test code (211 tests)

## Test plan
- [x] `flutter test` — all 211 DM-related tests pass
- [x] `flutter analyze` — no errors or warnings (13 pre-existing info-level issues in integration tests)
- [x] No user-visible changes — Notifications tab, routing, and bottom nav unchanged

PR #6591, opened 2026-08-02
fix(report): stop telling users a report was received when it wasn't

### LINKED ISSUE #6387: fix(report): moderation DM failure is reported to the user as success

### LINKED ISSUE #6595: fix(comments): CommentReactionsBloc discards the report result, so failed reports look successful
## Problem

`CommentReactionsBloc` awaits `reportContent()` and discards the returned `ReportResult`, so a failed comment
report emits no error and the UI shows nothing:

```dart
// mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart:342-347
await reportingService.reportContent(
  eventId: event.commentId,
  authorPubkey: event.authorPubkey,
  reason: event.reason,
  details: event.details,
);
// :348-349
// ContentReportingService returns a typed failure result for normal
// domain issues; a throw escaping here is unexpected.
```

The comment states the contract correctly and the code then ignores it. `reportContent` returns
`ReportResult.failure(...)` — it does not throw — for "Reporting service not initialized", "Not authenticated",
and "Failed to create report event". Only the catch-only path emits `ReactionsError.reportFailed`, so those
three all present as success.

This is the same class of defect as #6387 (discarded typed result), one type over.

> **Correction (2026-08-02):** this issue originally cited
> `conversation_actions_cubit.dart` as prior art for the correct shape. That is no longer accurate — PR #6591
> made `ReportResult.success` a weaker claim (it is true even when nothing left the device), and that cubit was
> reading it bare, so it was fixed in the same PR. Use
> `report_content_dialog.dart` as the reference instead: it branches on both `success` and
> `delivery`.

## Suggested fix

```dart
final result = await reportingService.reportContent(...);
if (!result.success) {
  emit(... ReactionsError.reportFailed ...);
  return;
}
```

plus a bloc test stubbing a returned `ReportResult.failure(...)` — note that a test stubbing a *throw* would
pass today and would keep passing after the fix, which is exactly how #6387 escaped review.

## Notes

Found while investigating #6387; filed separately since it is independent and in a different feature area.

Note that #6591 adds a `delivery` field to `ReportResult`, so once that lands this call site can distinguish
"failed" from "recorded but never left the device" too.

Verified at `5972f43e7`.

### LINKED ISSUE #6610: fix(report): a retried report can send the moderation DM twice
## Problem

`ReportContentDialog` guards the moderation DM against duplicate queueing on the undelivered path only. The
delivered branch at `report_content_dialog.dart:483` neither reads nor sets `_moderationDmQueued`:

```dart
if (result.success && result.delivery == ReportDelivery.localOnly) {
  if (!_moderationDmQueued) {                             // :475
    _moderationDmQueued = true;
    unawaited(_sendModerationDm());
  }
  // ... inline error, Submit stays live
} else if (result.success) {
  final moderationDmFailed = await _sendModerationDm();   // :483 — unguarded
```

So an undelivered submit followed by a successful retry, within the same dialog, delivers two copies of the
same report:

1. The undelivered submit's `sendMessage` enqueues a durable `outgoing_dms` row before any I/O
   (`dm_repository.dart:2832`). An offline publish hard-fails, so both wrap statuses are left retryable
   (`:3021` → `_finalizeAfterRecipientFailure`, `:4018`), and `OutgoingDmRetryService`'s failed-send arm
   replays it via `recoverFullSend` once connectivity returns (`outgoing_dm_retry_service.dart:371`).
2. The retry reaches `:483` and sends a fresh DM immediately.

Nothing collapses the two at the receiver. The sweep's replay is safe against *itself* because it replays the
persisted rumor bytes, so the rumor id is stable and NIP-17 receiver-side dedup drops a true duplicate. But
`:483` is a second `sendMessage`, which mints a new rumor with a fresh `created_at` and therefore a fresh id —
protocol-indistinguishable from the user genuinely reporting twice, and NIP-17 exposes no correlator. Two
kind-1059 wraps, two `dm_log` rows.

Confirmed with a throwaway widget test stubbing `localOnly` then `reached`: `sendMessage` is called twice. The
replay half is a code trace, not an end-to-end observation.

## Who hits it

Narrower than "offline". It needs the kind-1984 to sign but fail to publish — a local-key signer offline, or
any signer on a network where the relay is unreachable but the signer is not. On a Keycast account
(`authSource=divineOAuth`) dropping Wi-Fi does not reproduce it: signing goes out to `login.divine.video` and
fails first, so `reportContent` returns `failure` and neither DM branch runs.

## Introduced by #6591

On `main` the confirmation screen replaced the form, so there was no way to retry and the offline path sent
exactly one DM. Leaving Submit live on the undelivered path and queueing the DM there are both new in #6591;
together they produce the second copy.

Nobody is told anything false, so this sits outside what #6591 set out to fix. The cost lands on the
moderation side: a duplicate `dm_log` row today, and a duplicate `user_reports` row once #6593's intake is
wired.

## Why the obvious fix isn't obviously right

Checking `_moderationDmQueued` on the delivered path suppresses the fresh send and stakes the report entirely
on the parked row. The sweep caps at `maxRetries: 5` (`outgoing_dm_retry_service.dart:20`) and terminalizes a
row that burns its budget (`:687`), so under that fix a report whose parked row is abandoned is *lost* rather
than duplicated — a worse failure than the one being fixed. Options worth weighing:

- cancel or delete the parked `outgoing_dms` row when a later submit succeeds, then send fresh — one copy,
  no reliance on the sweep
- dedupe at the moderation-service reader on report identity (reporter + reported event + reason) rather than
  in the client
- accept the duplicate and document it — a moderator seeing a report twice is a smaller harm than seeing it
  zero times

The state machine is awkward to get right here because the widget owns the whole report → DM → decide
sequence; #6596 (extract `ReportSubmissionCubit`) would give this logic a testable home.

## Notes

Found in review of #6591:
https://github.com/divinevideo/divine-mobile/pull/6591#pullrequestreview-4838894591

The comment at `report_content_dialog.dart:472` claiming the DM is "queued at most once per dialog" is
corrected in #6591 — it holds for repeat taps on the undelivered path, not once a retry succeeds.

Verified at `e32f1d86f` (the #6591 branch).

realmeylisdev:
## Description

`ReportContentDialog` told users their report reached the moderation team even when the DM was never delivered.

`DmRepository.sendMessage` signals non-delivery by **returning** a `NIP17SendFailure`, not by throwing —
`NIP17MessageService.sendRumor` wraps its entire body in `on Object catch` and converts every relay, signer,
encryption, policy and timeout error into a returned value. The call site awaited it, discarded the result, and
set its failure flag only inside `catch`. So `reportModerationDmDelayed` never rendered for any real delivery
outcome.

Investigating it surfaced a second, larger lie one layer down: `ContentReportingService.reportContent` returns
`createSuccess` even when the kind-1984 publish failed on **every** relay *and* the Zendesk ticket failed. That
result gates the whole "Report Received" screen. **Report a video in airplane mode today and you are told "Our
team will review your report and take appropriate action" while nothing left the device** — local report history
is never replayed by anything, so the report really is gone.

That airplane-mode scenario is a **local-key** one. Signed in through Keycast (`authSource=divineOAuth`),
signing goes out to `login.divine.video`, so offline the kind-1984 event is never built and `reportContent`
returns `failure` — which already errors correctly today. The false confirmation reaches users holding their own
key, and any signer whose signing is local while its relays are not.

### What changed

1. **Honour the DM result.** Classify with an exhaustive `switch` over the sealed type, so a future variant
   becomes a compile error rather than a silent "delivered". `selfWrapPublished: false` stays silent on purpose —
   the team *did* receive the report; only the sender's own cross-device copy is missing.
2. **Carry the report's delivery outcome.** `ReportResult` gains a `ReportDelivery` field;
   `_createZendeskTicket` returns the bool it already computed. Delivery is the **disjunction** of the two
   channels — requiring both would invert the lie (relay down, ticket filed, a human holding the report, user
   told it failed).
3. **When nothing left the device, don't confirm.** That screen is false in four places at once — its title, the
   review promise, the green check, and any DM caveat — so one honest line underneath fixes nothing. Show an
   inline error and leave Submit live.
4. **Still send the moderation DM in that state.** `sendMessage` writes a durable `outgoing_dms` row before any
   I/O and marks it failed when the publish can't go out, which is what `OutgoingDmRetryService`'s second sweep
   arm replays on reconnect. It is the only report channel with a retry — the kind-1984 publish and the Zendesk
   ticket are both fire-and-forget — so skipping it would have made an offline report deliver *less* often than
   before this PR. Fired unawaited — the user already knows the submit failed and shouldn't wait on a doomed
   round-trip.
5. **Send that DM exactly once, across every retry.** Parking a durable row *and* leaving Submit live meant a
   submit that later got through called `sendMessage` again: the parked row replayed on reconnect and the retry
   sent a fresh copy, so one report reached moderation twice. The DM body is the report itself, so a second copy
   is a second thing to triage. `sendMessage` now reports the `outgoing_dms` row it left behind, and the dialog
   re-drives *that* row through `recoverFullSend` rather than minting another — one rumor id, one copy. Re-driving
   is what makes this safe: `recoverFullSend` joins an in-flight sweep replay instead of publishing alongside it,
   and `resetRetryBudget` re-arms a budget the sweep may already have spent, so coalescing can't turn a duplicated
   report into a dropped one. The dialog also joins a send still in flight, and stops once the team demonstrably
   has it. Fixes #6610.
6. **`@useResult` on the ten `NIP17SendResult`-returning methods**, so this cannot recur. It surfaced 45
   discards — **all in test files, none in production**, independently confirming the dialog was the only
   production site.
7. **`@useResult` on `reportContent` / `reportUser` / `quickReport` too**, for the same reason one layer up: a
   `success` result still needs its `delivery` read. That found one live discard, `CommentReactionsBloc`, which
   dropped the result entirely — so an uninitialized service, a missing signer, or an unbuildable event all
   presented as success. Fixed here (#6595). Its existing test stubbed a *throw*, which the handler already
   caught, so it passed before the fix and would have kept passing after.

**One ARB key added.** Both places this PR surfaces an undelivered report first reused
`reportFailed(commonSomethingWentWrong)` — "Failed to report content: Something went wrong". The inbox snackbar
fires after reporting a *person*, so "content" is wrong there, and "Something went wrong" is a dead end for what
is almost always connectivity. `reportNotSent` names neither subject and says what to do: "Couldn't send your
report. Check your connection and try again." Translated in all 22 locales, reusing each one's existing wording
for the check-your-connection clause. `reportModerationDmDelayed` is still reused as-is, and `reportFailed`
stays for the genuine-failure branch that has a real error string to interpolate.

### Why the existing test didn't catch this

`report_content_dialog_test.dart:739` *does* assert the notice renders — but stubs `thenThrow`, a shape
production never produces for a delivery failure (relay-unreachable returns
`failure('Message publish failed to relays')`). It is kept, because that path is still reachable via an
uninitialized repository, but it is why the gap shipped. New tests pin the shapes production actually produces.

For the record, this shipped broken rather than being broken later: `sendMessage` returned a result from its
first commit (#2118), the moderation caller landed one day later already discarding it (#2112), and #4944 bolted
the copy and the flag onto the already-dead catch.

**Related Issue:** Closes #6387, closes #6595

## Out of Scope

- **Pre-existing architecture violation, intentionally untouched:** the widget reads `dmRepositoryProvider`
  directly and orchestrates report → DM → decide. That composition belongs in a repository or cubit per
  `architecture.md`. This PR does not make it worse — it makes it *visible*, because the widget now branches on
  a result it previously discarded. Tracked in #6596.
- **Making the report itself durable.** The DM queue is the only retry an undelivered report gets, and it is a
  side channel: the kind-1984 event is the report of record, and it is dropped on the floor when the publish
  fails. Routing reports through `PendingActionService` would fix that properly and is the right follow-up, but
  it is a feature, not a review fix.
- Relay-fails-but-Zendesk-succeeds is not unit-tested: `ZendeskSupportService.createTicket` is static and cannot
  be injected, so it always returns false under test. The disjunction is covered from the relay side.
- The kind-1984 publish uses `publishEvent` (frame-accepted), not `publishEventAwaitOk`. Per NIP-01 an `OK` is
  acceptance, not durability — and Divine's relay emits it *before* the durable write — so switching would
  narrow the gap without closing it. Left alone.
- Separate issues filed for adjacent defects found during the investigation: #6592 (client-attribution toggle
  silently disables reports at the moderation-service), #6593 (moderation DM intake half-wired, unread badge dead),
  #6594 (local_stack images 5 months stale, NIP-17 untestable locally).

## Verification

- `dart format --set-exit-if-changed lib test packages/dm_repository` — clean
- `flutter analyze lib test integration_test packages/dm_repository` — no issues
- `flutter test` on the report dialog, inbox view, content-reporting service, comment-reactions bloc,
  conversation-actions cubit, nip17 service and l10n suites — **256 passing**
- `packages/dm_repository`: **554 passing**, coverage **93.84%** against the 93% gate
- `packages/models`: **839 passing**
- Coalescing round (`3f3cc92f3`, `5c167ff79`): every suite touching `NIP17SendResult` — **289 passing**; the
  pre-push hook's affected-file run — **588 passing**; all four CI shards green
- `flutter gen-l10n` re-run and the regenerated `lib/l10n/generated/` committed; `arb_consistency_test` green
- 17 ratchets run individually (UI/service boundary, untested-services floor, test-unit structure, all four
  design-system ceilings, raw logging, TODOs, skips, file size, empty catch, arb error, service sentinel,
  ChangeNotifier and Riverpod boundaries) — all pass
- Each new test verified **red without the production change and green with it**, by reverting each leg
  independently

Not run: goldens (no visual change to an existing rendered state) and the full app suite.

## Type of Change

**Behaviour change worth calling out in review:** offline reports now surface an error instead of a false
confirmation. That is the point of the fix, but it is user-visible.


mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -443,14 +444,29 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
...
 
       if (mounted) {
-        if (result.success) {
+        if (result.success && result.delivery == ReportDelivery.localOnly) {
+          // Nothing left the device — every channel refused, which usually
+          // means no connectivity. The confirmation screen would be false in
+          // four places at once (its title, the review promise, the green
+          // check, and any DM caveat), so don't show it: surface the failure
+          // and leave Submit live so retrying is one tap. The moderation DM
+          // is deliberately not attempted either — a retry re-sends
```
hm21: Question on the DM skip rather than an objection — the reasoning is sound as written.

The moderation DM publishes to the recipient's NIP-17 inbox relays, which are a different relay set from `_moderationRelayUrl`. So "moderation relay down + Zendesk down, DM inbox relays up" now drops the one channel that would have worked, and the retry drops it again. Does the double-delivery risk outweigh that, given #6593 says DM intake is half-wired anyway?

mobile/lib/services/content_reporting_service.dart
```diff
@@ -21,20 +37,33 @@ class ReportResult {
     required this.timestamp,
     this.error,
     this.reportId,
+    this.delivery = ReportDelivery.reached,
```
hm21: Nit: `reached` as the default here (and on `createSuccess` at :53) is fail-open, in a fix that exists because a default claimed success. Latent today since the one production caller passes it explicitly, but `localOnly` as the default — or making it required — would make the next caller fail closed.

hm21 (commented):
Verified the core claims independently: the sealed switch is exhaustive, `@useResult` covers all ten public methods with no production discards left, and airplane mode really does yield `PublishNoRelays` + `createTicket() == false`, so `localOnly` is genuinely reachable. Reverting the dialog against the merge base turns 5 of the 6 new tests red. Good fix.

One gap, your call whether it belongs here or in a follow-up: `ConversationActionsCubit.reportUser` (`lib/blocs/dm/conversation_actions/conversation_actions_cubit.dart:66`) is the second production reader of `ReportResult.success`, and it still returns bare `result.success` — so `inbox_view.dart:989` shows the `inboxReportedUser` snackbar in exactly the `localOnly` state this PR made detectable. Same lie, one line away (`result.success && result.delivery == ReportDelivery.reached`). #6595 doesn't cover it — that issue is about a *discarded* result in `CommentReactionsBloc`, and it cites this cubit as "prior art for the correct shape", which this PR just invalidated.

mobile/lib/widgets/report_content_dialog.dart
realmeylisdev: Checked the premise and it doesn't hold in production — the two relay sets are the same one.

The moderation account's kind-10050 DM inbox is exactly `[wss://relay.divine.video]` (queried live off the relay; it's been republished daily since 2026-07-13). And `_moderationRelayUrl` is `env.relayUrl`, which is `wss://relay.divine.video` for `AppEnvironment.production` (`environment_config.dart:74-75`). So "moderation relay down, DM inbox relays up" can't happen — one host, one socket.

The one way they diverge is the fallback: `resolveDmInboxRelays` returns null when the kind-10050 query fails, and the send falls back to the default pool. But that query goes to the same relay that's down, so the fallback is what you'd actually hit — and a gift wrap landing on some other relay is never ingested, because `dm-reader.mjs` polls only `wss://relay.divine.video`. It'd be published somewhere nobody reads.

So the skip costs nothing today. It stops costing nothing the moment the 10050 lists a relay the report path doesn't target — worth a comment if that ever changes, but I'd rather not encode a speculative relay topology in the skip condition now.

mobile/lib/services/content_reporting_service.dart
realmeylisdev: Agreed, fixed in c5776ff — made it `required` on both the constructor and `createSuccess` rather than flipping the default, so a caller that forgets fails to compile instead of failing open either way.

The compiler immediately found four call sites, which is a decent argument for the change.

realmeylisdev:
Thanks @hm21 — the `ConversationActionsCubit` catch was a real miss, and you're right that it's the same lie one screen over. Pushed as c5776ff.

**`ConversationActionsCubit.reportUser`** now requires `delivery == reached`. Confirmed the path exactly as you described: `inbox_view.dart:989` renders the `inboxReportedUser` snackbar off that bool, so reporting from a DM thread offline claimed success. New cubit test is red without the one-line change. Also updated the method's dartdoc, which still said "Returns `true` if the report was submitted successfully" — the exact conflation the change removes.

**`delivery` is now `required`** on both the constructor and `createSuccess`, rather than defaulting to `localOnly`. Fails closed either way, but this way the next caller can't forget silently. The compiler found four existing call sites immediately.

**On the DM-skip question** — replied inline: the premise doesn't hold today. The moderation account's kind-10050 inbox is exactly `[wss://relay.divine.video]`, the same relay `_moderationRelayUrl` resolves to in production, so the two can't fail independently. And if the 10050 lookup itself fails, the fallback publishes to relays `dm-reader.mjs` doesn't poll.

**On #6595** — good catch that this PR invalidated its "prior art" reference. Updated the issue body to point at the report dialog instead, and noted that this cubit was itself fixed here.

`@useResult` on the two `Future<List<NIP17SendResult>>` group methods is deliberate, for what it's worth — a partially-failed group send is exactly the shape that shouldn't be discardable.

realmeylisdev:
Correcting one number in my reply above: making `delivery` required broke **five** call sites, not four — `comment_reactions_bloc_test.dart:337`, `conversation_actions_cubit_test.dart:99`, and `report_content_dialog_test.dart:79`, `:89`, `:471`. The two other `createSuccess` callers already passed it explicitly, which is where I lost one.

Doesn't change the remedy — all five are migrated in c5776ff and `flutter analyze lib test` is clean.

mobile/lib/services/content_reporting_service.dart
```diff
@@ -27,14 +44,30 @@ class ReportResult {
...
-  static ReportResult createSuccess(String reportId) => ReportResult(
+  /// Whether the report reached any channel off this device. Only
+  /// meaningful when [success] is true.
+  ///
+  /// Required rather than defaulted: this type exists because a default
+  /// claimed success, so a caller that forgets to say must not inherit
+  /// the optimistic answer.
+  final ReportDelivery delivery;
+
+  static ReportResult createSuccess(
```
mbradley, line 56: `reportContent`/`reportUser`/`quickReport` didn't get `@useResult`, though it's the same shape you added it for. The live discard it would catch is #6595, so annotating pulls that in. Deliberate?

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -443,14 +444,29 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
...
+          // everything, and a queued DM plus a resubmit would deliver the
+          // same report twice.
+          setState(() {
+            _errorMessage = context.l10n.reportFailed(
+              context.l10n.commonSomethingWentWrong,
```
mbradley: "Failed to report content: Something went wrong" for what's usually connectivity, and "content" when it may be a person. Reuse keeps the no-ARB-change property, which is worth real money across 16 locales. Flagging, not asking.

mbradley (changes requested):
Ran the offline path on a simulator rather than reading it off the diff. hm21's items are resolved; not repeating them. Two things, plus a note.

**Pushed `b21e218a3`.** `inbox_view.dart:988` has no else, so now that `reportUser` returns false for an undelivered report, reporting from a DM thread offline shows nothing at all. You traced to this line in your reply and changed what it reads; the false path just wasn't in frame. Not strictly a regression from `c5776ff` either, since bare `result.success` was already false on signing failures. Reuses the dialog's copy, no new ARB key. Also pushed `a89b9efac`, a test that Submit stays live after `localOnly`, which the branch exists for but nothing asserted.

**A different objection to the DM skip than hm21's.** Yours holds on relay topology and I'm not reopening it. This is about the call itself: `sendMessage` enqueues a durable `outgoing_dms` row before any I/O (`dm_repository.dart:2830`), offline `sendRumor` hard-fails so the row is marked failed (`:4024`) precisely so `OutgoingDmRetryService` arm 2 replays it on reconnect (`main.dart:2726`), and the body is `_formatReportDm`, the report itself. So on `main` an offline report did reach moderation unattended. Skipping the call drops that. The double-delivery rationale doesn't quite hold up, since a resubmit already re-publishes the 1984 and files a second ticket and that's accepted. Options as I see them: send the DM and say it'll go when connectivity returns; route reports through `PendingActionService`; or keep it and record in the description that offline reports now deliver less often. Your call, but I'd like it to be a decision.

**Note on who this reaches.** Signed in via Keycast (`authSource=divineOAuth`), airplane mode never gets to `localOnly` — signing goes out to `login.divine.video`, fails, and `reportContent` returns `failure`, which already errors correctly today. hm21 verified `localOnly` is reachable, so I assume that was a local key. Worth knowing which, since it changes who the description's opening scenario applies to.

Two smaller ones inline. Thanks, as always.

mobile/lib/widgets/report_content_dialog.dart
realmeylisdev: Moot now — the skip is gone in `169ec32`, for a reason neither of us had in frame.

mbradley's objection lands: `sendMessage` writes the durable `outgoing_dms` row before any I/O and marks it failed when the publish can't go out, which is exactly what `OutgoingDmRetryService`'s arm 2 replays on reconnect. So on `main` an offline report did leave the device eventually, unattended, and the skip dropped that. It is the only report channel with a retry at all — the 1984 publish and the Zendesk ticket are both fire-and-forget — so skipping it made an offline report deliver *less* often than before a PR about honest delivery. Not a trade I want.

The double-delivery rationale I gave doesn't survive either: a resubmit already re-publishes the 1984 and files a second ticket, and that's accepted.

So the DM now goes out on the undelivered path too, fired unawaited (the user already knows the submit failed and shouldn't wait on a doomed round-trip) and at most once per dialog, since Submit stays live and every tap would otherwise stack another identical row for the sweep. Pinned by two tests — the `verifyNever` in the localOnly test flipped to `verify(...).called(1)`, plus one that a repeat submit still queues only one.

Your relay-topology point separately still holds as I answered it: one host, one socket, today.

mobile/lib/services/content_reporting_service.dart
realmeylisdev: Not deliberate, just stopped one layer short. Added on all three in `5240998`.

It found exactly one live discard, and it's #6595, so that's fixed here rather than left as a warning someone has to route around. `CommentReactionsBloc` now emits `ReactionsError.reportFailed` for both a returned failure and a `success` that reached no channel.

Its existing test stubbed a throw, which the handler already caught — green before the fix, green after. Same shape that let #6387 through. Replaced with two that stub what `reportContent` actually returns; both are red without the change.

The other 14 hits were all in `content_reporting_service_test.dart`, side-effect assertions, so they take the same `final _ =` this branch already used for the ten NIP-17 methods.

mobile/lib/widgets/report_content_dialog.dart
realmeylisdev: Taking it — this branch is what put that string in both places, so it's ours to get right.

New key in `bfd14cc`: `reportNotSent`, "Couldn't send your report. Check your connection and try again." Names neither subject, so it's honest in the inbox snackbar where the thing reported is a person, and it points at the actual cause instead of dead-ending. `reportFailed` stays on the genuine-failure branch, which has a real error string to interpolate.

Translated in all 22 locales, reusing each one's existing wording for the check-your-connection clause (from `profileSetupNoRelaysConnected`) so it reads native rather than machined, and each locale's own established term for "report". `arb_consistency_test` is green.

mbradley (approved):
Re-reviewed at `e32f1d8`. All three resolved, and I checked the guards rather than the claims:

- disabling the new bloc emit turns both new `CommentReactionsBloc` tests red
- restoring the DM skip turns two dialog tests red
- removing the once-per-dialog guard turns the dedupe test red

CI green. Approving.

**One thing to track separately: the moderation DM can go out twice.**

The once-per-dialog guard only covers the undelivered path. `report_content_dialog.dart:483` neither checks nor sets `_moderationDmQueued`, so an undelivered submit followed by a successful retry produces two copies: the parked `outgoing_dms` row replays on reconnect, and the retry sends a fresh DM. Confirmed with a throwaway widget test stubbing `localOnly` then `reached`, `sendMessage` called twice. Code trace for the replay half, not an end-to-end observation.

It is introduced here rather than pre-existing: on `main` the confirmation replaced the form, so there was no way to retry and the offline path sent exactly one DM. Leaving Submit live and queueing on the undelivered path are both new, and together they produce the second copy.

Narrower than "offline", worth being precise: it needs the 1984 to sign but fail to publish, so a local-key signer offline, or any signer on a network where the relay is unreachable but the signer is not. On a Keycast account you cannot reproduce it by dropping Wi-Fi, since signing fails first and returns a plain failure.

Nobody is told anything false, so it does not touch what this PR set out to fix, and the comment at `:472` claiming "queued at most once per dialog" is wrong for this path and worth correcting. I am not suggesting a fix here because the obvious one is not clearly safe: checking the flag on the delivered path would suppress the fresh send, and if the sweep exhausts its retry budget the parked row is abandoned, so the report would be lost rather than duplicated. Please file a follow-up issue and link it here.

Thanks, as always.

realmeylisdev:
Filed #6610, and corrected the `:472` comment in `c7ef31043`.

Reproduced the duplicate before filing rather than taking the trace on faith — throwaway widget test stubbing `localOnly` then `reached`, `sendMessage` called twice.

One thing worth adding to your trace: nothing collapses the two at the receiver either. The sweep's replay is safe against *itself* because `recoverFullSend` replays the persisted rumor bytes, so the id is stable and NIP-17 receiver dedup drops a true duplicate. But `:483` is a second `sendMessage`, which mints a new rumor with a fresh `created_at` and therefore a fresh id, and NIP-17 exposes no correlator to tie them back together. Two kind-1059 wraps, two `dm_log` rows.

Agreed on not fixing it here, and #6610 records why: checking the flag on the delivered path stakes the whole report on a row the sweep terminalizes at `maxRetries: 5`, which turns a duplicate into a loss. The three options I could see are in the issue — cancel the parked row on a later success, dedupe at the moderation-service reader, or accept and document.

Thanks for the re-review.

rabble (changes requested):
Blocking finding on current head c7ef31043a70ebb897af01ddfd2c0dd534979e23:\n\nmobile/lib/widgets/report_content_dialog.dart:484-485 can send a duplicate moderation DM. The first localOnly result queues a durable outgoing_dms row via the unawaited _sendModerationDm() at lines 477-479. If the user taps Submit again after reportContent later returns a delivered result, this success branch calls _sendModerationDm() again while OutgoingDmRetryService may still deliver the queued row. One report can therefore reach moderation twice. The _moderationDmQueued guard only covers repeated localOnly attempts, and the latest commit explicitly documents this gap as #6610. Please coalesce the queued retry with a later successful attempt (and add a regression test for localOnly -> delivered across two submits) before merging.

realmeylisdev:
Fixed and pushed — `3f3cc92f3` and `5c167ff79`.

Coalesced by re-driving the parked row rather than suppressing the second send. `sendMessage` now reports the `outgoing_dms` row it left behind on a failure, and the dialog re-drives *that* row through `recoverFullSend` instead of calling `sendMessage` again. One rumor id, so one copy reaches moderation.

That shape is what makes it safe to do here. mbradley's objection to the obvious fix was that guarding the delivered path would drop the report outright if the sweep exhausted its retry budget and abandoned the row. `recoverFullSend` avoids both horns: it joins an in-flight sweep replay for the same rumor instead of publishing alongside it, and `resetRetryBudget: true` re-arms a budget the sweep may already have spent — an explicit resubmit is exactly the signal that primitive was built for. A row that has vanished means the sweep already delivered it, which is the one case with nothing left to send.

Three regression tests on `localOnly` → delivered across two submits, including the one you asked for. All three fail against the previous head with `sendMessage` called twice instead of once — the duplicate, reproduced. Also covered: the sweep winning the race between submits, and a re-drive that fails again still carrying the delayed caveat rather than laundering it into silence.

The stamping is pinned in `dm_repository` too: a hard failure and a soft-unconfirmed both report their row, a policy block reports none since the gate returns before the enqueue.

#6610 was filed for this gap and is closed by the second commit, so it comes off the follow-up list. The `:472` comment documenting the gap is gone with it.

NotThatKindOfDrLiz:
Takeover update pushed in 99eebf124.

What changed:
- moved moderation-DM preflight reads/formatting under _dispatchModerationDm's catch so the localOnly unawaited path cannot leak provider/formatting errors to the zone handler
- made ambiguous recoverFullSend ArgumentError results keep the delayed-DM caveat instead of claiming moderation definitely received the DM
- added/updated widget regressions for both paths
- rebased the PR onto current origin/main before pushing

Verification:
- flutter test test/widgets/report_content_dialog_test.dart
- flutter analyze lib/widgets/report_content_dialog.dart test/widgets/report_content_dialog_test.dart
- pre-push hook: analyzer plus affected-file tests passed
- GitHub checks are all green

NotThatKindOfDrLiz (approved):
Approved after takeover fix 99eebf124. The stale duplicate-DM request is addressed, the unawaited preflight path is caught, ambiguous recovery errors no longer claim delivery, and all GitHub checks are green.

NotThatKindOfDrLiz:
@rabble requesting re-review on the current head 99eebf124.

Your blocking review was on c7ef31043. The duplicate-DM path is now coalesced by ac0e2e853/d21bfdafd: sendMessage returns the parked outgoing_dms row, and the report dialog re-drives that row with recoverFullSend(resetRetryBudget: true) instead of minting a second rumor. The localOnly -> reached regression is in report_content_dialog_test.

I also pushed 99eebf124 to close the last two review edge cases: unawaited preflight errors are caught, and ambiguous recoverFullSend ArgumentError results keep the delayed-DM caveat rather than claiming delivery. CI is green.

realmeylisdev:
Reviewed the takeover commit. Both findings in `99eebf124` are real; the fix for the second one re-opened #6610 on a narrower path. Corrected in `efcd59b48`.

**The preflight move is a straight improvement.** `_dispatchModerationDm` is `async`, so those `ref.read`s never threw synchronously in the first place — the comment I had written above them claimed a guarantee the language doesn't give, and on the unawaited `localOnly` path the resulting future error had nowhere to land. Kept, with the comment rewritten to say what the `try` is actually for.

**The `ArgumentError` branch needed one more thing.** Clearing `_queuedModerationDmId` while leaving the delivered flag `false` leaves nothing behind it: Submit stays live on the undelivered path, so the next tap finds `parked == null`, falls into the fresh-send branch, and mints a second rumor — the duplicate the r4 coalescing existed to remove.

Two submits can't reach it, which is why the rewritten test stays green either way: its second submit is `reached`, so the dialog lands on the confirmation and Submit is gone. Three can — `localOnly` → `ArgumentError` → `localOnly` calls `sendMessage` twice. That's in `report_content_dialog_test` as *never mints a second DM after recovery loses the row*, red against `99eebf124` with `Expected: <1> Actual: <2>`.

Your honesty correction is preserved — the branch still refuses to claim a delivery `recoverFullSend` couldn't prove, and still shows the delayed caveat. It's now `_ModerationDmOutcome.unverifiable`, which is terminal for *sending* as well: the pair of booleans became a tri-state, since "delivered" and "gave up" both mean don't send and only one of them means don't warn.

Local verification: `flutter test test/widgets/report_content_dialog_test.dart test/widgets/report_content_confirmation_test.dart`, `flutter analyze lib test integration_test`, pre-push hook including `dm_repository`.

rabble (approved):
Re-checked at `efcd59b4` — my earlier changes-requested is resolved, so clearing it.

The duplicate moderation DM is fixed properly rather than patched. `_sendModerationDm` now dispatches on a `_ModerationDmOutcome` state machine that coalesces all four cases: already delivered (no-op), parked row gone unverifiable (no-op, keeps the caveat), send in flight (joins it), and an earlier parked row (re-drives that row rather than minting a second). That is the coalescing I asked for, and it covers the delivered path my review said the `_moderationDmQueued` guard missed.

The regression test I asked for is there — `report_content_dialog_test.dart:1125`, 're-drives the parked DM when a later submit gets through', stubbing `[localOnly, reached]` across two submits. Three siblings cover the rest of the matrix: no second DM on a repeat submit, no second DM after recovery loses the row, and the delayed caveat surviving a failed re-drive.

CI green. Approving.

Issue #6081, opened 2026-07-14
feat(divine-ui): add bug_report icon (unblocks creator_analytics_screen sweep)


realmeylisdev:
**Parent**: #4803 (Migrate raw `Icons.*` to `DivineIcon` across the codebase)

## Problem

`creator_analytics_screen.dart`'s diagnostics toggle switches between
`Icons.bug_report` and `Icons.bug_report_outlined` (filled/outline pair)
via `MaterialIconSource`. `divine_ui`'s `DivineIconName` enum has no
bug-report glyph, filled or outline, so this site can't migrate off
`MaterialIconSource` yet — same shape as the gaps tracked in #4833.

## Definition of done

1. Designer provides SVGs for `bug_report` (filled) and `bug_report_outlined`
   (outline), Phosphor-style, matching the existing `divine_ui` icon set.
2. Assets added to `divine_ui` and wired into the `DivineIconName` enum +
   `assetPath` switch, with matching tests (`divine_ui` is a
   **strict-100%-coverage** package).
3. Follow-up PR replaces `creator_analytics_screen.dart`'s
   `MaterialIconSource` toggle with `DivineIcon(...)` and ticks it off
   under #4803.

## Source

Surfaced by PR #6076 review (hm21, 2026-07-14): the `// ignore:
deprecated_member_use` comment at this call site said "tracked for a
future migration" without a tracking issue to point to.

Issue #4833, opened 2026-05-30
feat(divine-ui): add 15 missing settings icons (unblocks Icons.* sweep)


realmeylisdev:
**Parent**: #4803 (Migrate raw `Icons.*` to `DivineIcon` across the codebase)

## Problem

The settings-screens slice of the `Icons.* → DivineIcon` sweep (PR #4802) left **15 Material icons untouched** because `divine_ui`'s `DivineIconName` enum has no equivalent SVG asset. These callsites stay as raw `Icons.*` and can't pick up design-system stroke/style updates, creating the same drift #4803 / #3649 / #3652 track. They can only be migrated once the assets exist in `divine_ui`, so they're split out here as the designer-scoped prerequisite.

## Missing icons

Verified against the `DivineIconName` enum in `mobile/packages/divine_ui/lib/src/icon/divine_icon.dart` — none have a clean 1:1 match today:

| Material `Icons.*` | Used in (settings slice) |
| --- | --- |
| `add_circle_outline` | `safety_settings_screen.dart` |
| `remove_circle_outline` | `safety_settings_screen.dart` |
| `label_outline` | `safety_settings_screen.dart` |
| `people` | `safety_settings_screen.dart` |
| `cloud_upload` | `blossom_settings_screen.dart` |
| `cloud_done` | `relay_settings_screen.dart` |
| `cloud_off` | `relay_settings_screen.dart` |
| `open_in_new` | `relay_settings_screen.dart` |
| `mic` | `content_preferences_screen.dart` |
| `radio_button_checked` | `content_preferences_screen.dart` |
| `radio_button_off` | `content_preferences_screen.dart` |
| `local_fire_department` | `sounds_screen.dart` |
| `music_off` | `sounds_screen.dart` |
| `search_off` | `sounds_screen.dart` |
| `star` | `sounds_screen.dart` |

These 15 also appear across other (not-yet-swept) screens — the asset additions unblock the whole-codebase sweep tracked by #4803, not just the settings slice.

## Near-equivalents to confirm before commissioning new SVGs

A few have a *near* match already in the enum — design should decide whether the existing glyph is acceptable rather than adding a redundant asset:

- `open_in_new` → `arrowUpRight` / `linkSimple` / `shareNetwork`
- `star` → `sparkle` (different glyph; likely needs a true star)
- `radio_button_checked` / `radio_button_off` → `selection` / `selectionFill` (different semantics)
[...]

PR #5570, opened 2026-06-26
feat(video-editor): add voice-over recorder

hm21:
## Description

Adds a **voice-over recorder** to the video editor's "Add" sheet so creators can record narration over a clip without leaving the screen.

- New **Voice over** action in the editor Add sheet (new `microphone` icon in `divine_ui`).
- Full-screen recorder (Page/View + `VoiceOverCubit`): close/done on top, live amplitude **waveform** in the middle (continuous per-frame scroll), and a record/stop button below.
- **Multiple takes** in one session; each take can be removed via **Delete last recording**.
- **Time readout** `recorded / available` (e.g. `0:08 / 0:06`) that turns red and fires a haptic the moment the audio outgrows the video length.
- Takes are committed to the timeline as draft-local audio tracks, laid out back-to-back and clamped to the video length (wrapping to the start when full so nothing lands off-screen).
- Take titles are **numbered** ("Recording 1", "Recording 2", …) and continue past voice-over already on the timeline.
- **Haptics** on record start/stop, the too-long threshold, delete, and done — via the shared `HapticService`.
- Microphone capture via the new `record` dependency behind a `VoiceOverRecorderService`; permission gated through `permissions_service`.
- Localized in all 17 non-English locales.

**Related Issue:** Closes #1923

## Out of Scope

- New takes are placed from the start of the timeline (overlapping any audio already there) rather than appended after existing tracks — the time readout still warns about total length.

## Verification

- `flutter analyze lib test` — clean
- `flutter test` for the voice-over cubit, recorder screen, the affected video-editor widget tests, and `divine_ui` icon — all green
- `flutter gen-l10n` — regenerated, no errors
- `dart format` — clean

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #6599, opened 2026-08-02
refactor(ui): move the settings surface onto divine_ui and fix what it exposed

### LINKED ISSUE #6611: refactor(ui): move the settings surface onto divine_ui and fix what it exposed
Auto-created from PR #6599.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6599
**Branch:** `refactor/settings-divine-ui`

hm21:
Closes #6611

## Description

The settings hub and its sub-screens were the largest remaining pocket of design-system drift — every screen carried its own inline typography, button styling, snackbar and private copy of the same list tile. Moving them onto `divine_ui` needed four components that did not exist yet, and surfaced three bugs that were never about settings at all.

| | before | after |
|---|--:|--:|
| Inline `TextStyle(` | 211 | **0** |
| Material buttons | 51 | **0** |
| Hand-built `SnackBar(` | 59 | **0** |
| `Switch` / `Checkbox` / `Radio` | 16 | **0** |
| Raw `TextField` | 5 | **0** |
| `CircularProgressIndicator` | 13 | **0** |
| Raw dialogs / sheets | 18 | **2** |
| Raw `Icons.*` | ~50 | **28** |

Net **−1244 lines** across `lib/screens/`.

### The three bugs

**Material 3's purple was leaking through `VineTheme`.** The theme set `primarySwatch` and `primaryColor` but never `colorScheme`, and M3 ignores `primarySwatch` — so the scheme fell back to the M3 baseline, whose primary is `#D0BCFF`. Everything defaulting to `colorScheme.primary` picked it up: `RefreshIndicator` resolves `widget.color ?? colorScheme.primary` and has no theme hook at all in Flutter 3.44, and eleven bare progress indicators went the same way. The scheme is now seeded from `vineGreen`, with `progressIndicatorTheme` pinned so the default is the brand colour rather than one nobody chose. This is the widest change here and the first thing to look at on device.

**`VineBottomSheet` dropped `contentTitle` silently.** It was read only on the fallback `children` path, so passing `body` or `buildScrollBody` short-circuited it away. One call site had already found this and routed its title through the header with a comment explaining the workaround. The title now renders for all three content modes — pinned above caller-owned content, still inside the scroll view on the `children` path so no existing sheet shifts. The two existing tests only covered `children`, which is exactly why it survived.

**`vineGreenLight` is a dark-mode colour used as a foreground.** Against the light card it measures **1.14:1**, where WCAG wants 3:1 for icons; the `alpha: 0.1` container tint behind it was invisible outright. Four sites moved to `vineGreenDark` (4.86:1 dark, 3.58:1 light). Of the nine palette constants that clear 3:1 in both modes, `primaryAccessible` is documented for dark backgrounds specifically, so it was not the right pick.

### New in `divine_ui`

Five shapes every settings screen was re-implementing privately:

- **`DivineSwitch` / `DivineSwitchTile`** — the design system had no toggle at all.
- **`DivineListTile` / `DivineSectionHeader`** — replaces nine per-screen `_SettingsTile` / `_LegalTile` / `_SectionHeader` copies. `leading` is an escape hatch for glyphs the icon set still lacks.
- **`DivineSelectableRow`** — the single-select row, with the check-on-active affordance plus the secondary line `VineBottomSheetSelectionMenu` cannot render.
- **`DivineCheckboxTile`** — the checkbox counterpart of `DivineSwitchTile`.
- **`DivineInfoCard`** — the "about this setting" callout that twelve screens each carried privately, in four tones. The tinted tones derive background (15%) and border (30%) from a single accent colour rather than from two independently chosen values, so a tone cannot drift the way the hand-rolled boxes had; `neutral` sits on the card surface for callouts that should recede, and `compact` tightens the whole card by one step for a note beside a control. The sweep reaches past settings into explore, for-you and classic-vines, which is where nine more raw `TextStyle`s went.

Both tiles express *disabled* as an opacity pass over stable text styles rather than a second set of colours. Recolouring text when `onChanged` flips — which a row gated on an async provider does on its first frame — hands `_RenderListTile` a fresh `TextPainter` mid-frame, and a subtitle long enough to wrap then paints at a size it was not laid out with, tripping Flutter's `debugSize == size` assertion. This bit the Nostr settings screen for real; both components ship a regression test for the flip.

Five existing components grew the slot that kept forcing callers back onto Material: `DivineSnackbarContainer` takes a `backgroundColor`, `DivineAuthTextField` takes `prefixText` / `suffixText`, `DivineTextField` takes `filled` / `fillColor` / `suffixIcon`, `DivineListTile` takes `trailingColor` so a destructive row's caret stops going green against a red title, and `VineBottomSheet` takes `contentTitleTrailing` — `trailing` lands in the header above the divider, which reads as chrome rather than as an action on the titled content.

### Type scale, corrected mid-review

The first pass mapped headings to the nearest design-system size, which rounded 18px and 20px section headings **up** to `titleLargeFont` — 22px, the same size as the app-bar title. Nine sites across three screens ended up as loud as the screen title, and the 20-over-18 distinction between a section heading and a card heading was flattened away. `titleLarge` is for empty-state heroes and sheet titles everywhere else in the app; in-body headings belong a step or two below. The same pass papered over line-height differences with `.copyWith(height:)` in four places, defeating the point of the font helpers.

The Apps directory — a settings sub-page missed in the first sweep — had the mirror problem: two half-migrations, one on raw 17/15/13px and one on `headlineSmallFont` at 24px, so its info card shouted louder than the rows it introduced.

### Cache recovery, rehomed and renamed

Cleaning up the feature-flag screen took its "App Recovery" section with it, which stranded `CacheRecoveryService` with no caller. That path is not redundant with Settings → Storage: the storage screen only trims downloaded media, while recovery also wipes the Hive boxes, Application Support and the temp directories — the whole point when an install is corrupted. It belongs on Storage rather than behind a developer toggle, so it moved there, with `StorageCubit` owning it so the screen never imports `services/` directly.

Sitting next to the cache section made a labelling problem obvious. Storage showed two "cache" actions with two different sizes — "Cached media / 300 MB in use / Clear cache", and below it "App Recovery / Cache Info / Total cache size: 1.2 GB / Clear All Cache" — where the second number is a superset of the first and came from a different formatter. The destructive wipe read as the thorough variant of the routine clear. It is not: it drops the Hive boxes, signs the user out and needs a restart.

Fixing the labelling beat merging the two sections, which would have put a sign-you-out wipe in the same visual block as a safe, repeatable clear. The section is now **Repair install** / **Reset app data** and never says "cache", so there is nothing left to confuse. The size moved into the confirmation sheet, where it informs the one decision it is relevant to, and the button that only measured is gone; measuring runs alongside the open sheet, because walking Application Support is too slow to pay for on every screen entry, and a failed measurement hides the line rather than claiming the wipe failed. The footprint is bytes formatted by the page now, so both sections finally agree on units. Twelve borrowed `featureFlag*` keys became `settingsStorageRepair*` across all 22 locales.

The rehome had also left two things behind: `listenWhen` never watched `recoveryStatus`, so the outcome was never announced to screen readers, and a failed wipe was announced but never rendered.

### Regressions caught in self-review

Two commits here fix things this branch itself broke — found by re-reading the sweep, not by CI:

- **Content filters lost both cues for the active mode on a locked row.** The fill moved to `outlineDisabled` (1.04:1 against the card) and the font-helper swap dropped the selected weight; with age verification off, six of nine rows are locked, so the screen showed no active mode at all. In the same pass: the add-relay sheet is fixed-mode and never lifted clear of the keyboard, content preferences was the only one of twenty `buildScrollBody` call sites that discarded its `ScrollController`, and both of its pickers popped through the settings screen's `Navigator` rather than the sheet's — so a slow write closed the screen instead of the picker.
- **`DivineSectionHeader` stopped uppercasing.** Its own doc called it "an uppercase group label" but it rendered the title verbatim, so the screens whose private headers used to call `toUpperCase()` quietly lost their caps. It also hardcoded `vineGreen`, a dark-mode accent measuring ~2.1:1 on the light background — as did the scheme's `primary`, so a 12sp header and anything defaulting to `colorScheme.primary` were unreadable in light mode.

### Accessibility

The content-filter segments and the app-directory row were bare `GestureDetector` / `InkWell` with no semantics — a screen reader announced neither as a control. Both now merge their child text into one node rather than repeating the label in `label:`, so the announced string cannot drift from the rendered one. The segments additionally report their locked state and that the three are mutually exclusive, which the existing segmented-toggle precedent has no need for.

### Deliberate visual changes

Worth looking at first on device, because these are not pure restructuring:

- **The app-wide `colorScheme`**, above. `ColorScheme.fromSeed(vineGreen)` regenerates the whole tonal palette, so `secondary` / `tertiary` / `surface` / `error` now derive from green rather than the M3 baseline. Tests and goldens are clean, but golden coverage is thin.
- **Tile styling unifies on the settings hub's variant** — grey leading icon, 12px subtitle, green caret. Nostr and Legal keep their green leading icons via `iconColor` but pick up the smaller subtitle and the green caret.
- **Single-select lists lose their Material radios** for the tinted-row-plus-check affordance. Affects app language, content language, audio input and video shape.
- **Creator analytics chart colours** now use the app's semantics. Likes were green here and red everywhere else; that removed six duplicated literals and the inconsistency together.
- **The branded loading mark gains a shadow in light mode.** The sprite is white and all but vanished on the light palette's off-white surfaces; a tinted copy of the sheet painted behind it gives it an edge while the mark stays white — no second asset. Blur and offset are ratios of `size`, so a 16px inline spinner and an 80px full-screen one carry the same weight.
- **Storage grows a "Repair install" section**, above — new copy, and one destructive button where there used to be two buttons under a "cache" heading.

### A component bug this surfaced

`DivineSnackbarContainer`'s action is a plain button and, unlike `SnackBarAction`, does not dismiss the banner — the Bluesky handle-claim flow left a stale snackbar on screen after the action was tapped. Worked around locally with an explicit `hideCurrentSnackBar()`; the gap affects every design-system snackbar with an action and wants its own fix.

**Related Issue:**  — Relates to #4833, #4339, #6081

## Out of Scope

- **28 `Icons.*` callsites** have no design-system asset — that is #4833. Where a clean equivalent existed it was taken, which removes several entries from that issue's list. The rest keep Material through an explicit escape hatch rather than a glyph that means something else.
- **Semantic accent tokens.** `VineThemeColors` has surface tokens but nothing for like / comment / follow / mention / repost, so those stay static constants that cannot adapt to mode. Two more icons on the notification screen measure 2.2:1 in light mode, and `vineGreenLight` is still a foreground in five auth-adjacent files. Fixing that properly means adding tokens to the `ThemeExtension` and deciding what the light-mode accents should be — a design decision, not a silent correction.
- **The bug-report and feature-request dialogs** are four-field forms. A sheet with a keyboard over four fields is worse than what they have; they want a full-screen route, which is a product change rather than a component swap.
- **`DivineTextField` still has no `errorText`,** so the split between it and `DivineAuthTextField` remains unclear. Merging them is its own change.
- **`creator_analytics_screen.dart` is 1484 lines**, over the advisory 800-line ceiling. Extraction is its own PR.

## Verification

- `dart format` — clean
- `flutter test test/screens test/widgets test/notifications test/router test/blocs` — **7813 passing**
- `divine_ui`: **819 tests, 100.00% line coverage** (1963/1963 lines — the package's CI gate)
- `mobile/scripts/golden.sh verify` — passing
- `test/l10n/arb_consistency_test.dart` — passing, with all new keys translated into every one of the 22 locales
- All eight CI ratchets green; six baselines regenerated and shrunk

Rebased onto `origin/main` before this push. Two conflicts, both resolved in main's favour on structure and this branch's on components: the settings hub had moved "Experimental Features" below "Integration Permissions" and wrapped developer options in a spread, and the notification screen had gained the `showNewPosts` toggle where this branch replaced `_SectionHeader` with `DivineSectionHeader`.

Contrast figures above are computed WCAG relative-luminance ratios against the actual card colours (`#1A1A1A` dark, `#FFFFFF` light), not estimates.

Roughly a dozen app tests were updated to assert design-system behaviour instead of the Material mechanism they used to reach for — `banner.error` rather than `snackBar.backgroundColor`, `DivineSwitchTile` rather than `SwitchListTile`, the trailing check rather than a radio icon. Three needed more than a finder swap, and each was verified against `origin/main` first rather than assumed flaky: `nip05_settings_navigation_test` needed `ensureVisible` because `DivineListTile`'s 64px minimum height pushes the target below the fold in the 800×600 viewport; `app_detail_screen_test` needed the same plus a pump, because smaller type made the list fit inside the cache extent so `scrollUntilVisible` stopped scrolling and the tap landed at y=616 in a 600px viewport; and the light-mode palette test now asserts `onSurface` rather than `primaryText`. In `crossposting_settings_screen_test`, one `CircularProgressIndicator` assertion stays: it targets `DivineButton`'s own in-button spinner, not the screen-level loader.

**Verified on a real Samsung Galaxy device** — the settings surface, the repaired regressions and the new Storage "Repair install" section all behave as expected.

## Type of Change

- [x] 🧹 Code refactor


mobile/lib/blocs/storage/storage_cubit.dart
```diff
@@ -107,6 +122,50 @@ class StorageCubit extends Cubit<StorageState> {
...
+  /// A measurement failure falls back to [StorageRecoveryStatus.idle] rather
+  /// than `failure`: the size is decoration on the confirmation sheet, and
+  /// `failure` is what the UI renders when the wipe itself went wrong.
+  Future<void> loadRecoveryFootprint() async {
+    emit(state.copyWith(recoveryStatus: StorageRecoveryStatus.measuring));
+    try {
+      final footprint = await _measureRecoveryFootprint();
+      emit(
+        state.copyWith(
+          recoveryStatus: StorageRecoveryStatus.measured,
```
realmeylisdev, line 144: `_confirmRepair` starts this with `unawaited(...)` (storage_management_page.dart:422) and then awaits the sheet, so the measure can still be in flight when the user confirms. When it lands it emits `measured` straight over the `recovering` that `recoverFromCorruptedCache` just set.

Mid-wipe that means the "Repairing" line disappears and the `status == recovering ? null : ...` guard on the button at line 407 goes live again, so a second wipe can be started on top of the running one.

Repro, widget test against `StorageManagementView` with both futures held open by Completers:

```
AFTER CONFIRM  status=StorageRecoveryStatus.recovering
MID-WIPE       status=StorageRecoveryStatus.measured
inProgress shown mid-wipe? false
```

The shipped tests don't see it because every one of them resolves `measureRecoveryFootprint` instantly, so the measurement is always finished before the sheet is even tapped.

mobile/lib/screens/settings/storage/storage_management_page.dart
```diff
@@ -364,6 +358,150 @@ class _LibrarySection extends StatelessWidget {
...
+    );
+  }
+
+  Future<void> _confirmRepair(BuildContext context) async {
+    final l10n = context.l10n;
+    final cubit = context.read<StorageCubit>();
+    // Measure alongside the open sheet rather than on screen entry: walking
+    // Application Support is too slow to pay for on every visit, and the
+    // number only informs this one decision.
+    unawaited(cubit.loadRecoveryFootprint());
```
realmeylisdev, line 450: Because this is unawaited, the two emits in `loadRecoveryFootprint` can also land after the page is popped and `BlocProvider` has closed the cubit. `Cubit.emit` throws on a closed cubit in bloc 9.2.1 and nothing is awaiting this future, so it surfaces as an uncaught async error:

```
CAUGHT >>> Bad state: Cannot emit new states after calling close
```

(close the cubit while the measure future is pending, then complete it.)

The window is wide here by design, since the reason for firing it unawaited is that walking Application Support is slow. Guarding the post-await emits with `if (!isClosed)` is the pattern already used for this in the repo (voice_over_cubit, #5570).

mobile/lib/screens/settings/storage/storage_management_page.dart
```diff
@@ -87,6 +100,10 @@ class StorageManagementView extends StatelessWidget {
       message = l10n.settingsStorageCleared;
     } else if (state.libraryStatus == StorageLibraryStatus.cleaned) {
       message = l10n.settingsStorageBrokenClipsRemoved;
+    } else if (state.recoveryStatus == StorageRecoveryStatus.recovered) {
```
realmeylisdev: `_announceOutcomes` is first-match-wins, and `cacheStatus` / `libraryStatus` are sticky: nothing ever moves them back off `cleared` / `cleaned`. So once the cache has been cleared in this visit, every later `listenWhen` fire still matches the first branch and these two recovery branches are unreachable.

Repro, clear the cache then run the repair, capturing `SystemChannels.accessibility`:

```
after clear:  [Cache cleared]
after repair: [Cache cleared, Cache cleared, Cache cleared, Cache cleared]
recoveryStatus=recovered  cacheStatus=cleared
```

So the `listenWhen` half of the fix landed but the outcome still isn't announced. Keying the message off whichever status actually changed between prev and curr would close it.

mobile/packages/divine_ui/test/src/switch/divine_switch_test.dart
```diff
@@ -0,0 +1,277 @@
...
+      );
+
+      expect(tester.widget<Opacity>(find.byType(Opacity).first).opacity, 1.0);
+    });
+
+    // A row gated on an async provider is built disabled on its first frame
+    // and flips to enabled once the provider resolves. Recolouring the text
+    // across that flip used to hand _RenderListTile a fresh TextPainter for a
+    // subtitle that had already been laid out, tripping `debugSize == size`.
+    testWidgets(
```
realmeylisdev, line 255: nit: this one can't fail. I rebuilt the implementation the comment above it warns about (drop the `Opacity`, put `color: isEnabled ? primaryText : disabled` back on the title and subtitle) and ran this test against it unmodified:

```
DivineSwitchTile survives a disabled -> enabled flip with a wrapping subtitle
All tests passed!
```

`assert(debugSize == size)` only fires when the paint-only paragraph rebuild comes back a different size, and the deterministic test font never produces that, so `takeException()` is null either way. Same shape at `divine_checkbox_tile_test.dart:125`.

The components themselves are fine, it's just that the guard the description advertises isn't actually there. Either drop the two or assert something that can go red, e.g. that the resolved text styles are identical across the flip.

realmeylisdev (changes requested):
four things, all inline. main one: in the new Storage repair section, if the footprint measure is still running when you confirm, it emits over `recovering` and the destructive button goes live again mid-wipe. that same unawaited call can also emit into a closed cubit if you leave the page.

the repair outcome also never gets announced once the cache has been cleared in the same visit, so the listenWhen fix only got half of it. fourth is a nit on the two flip regression tests, they pass against the implementation they're meant to catch.

rest of the sweep holds up. analyze clean, divine_ui 821 tests at 100%, all 21 ratchets green, the 11 new l10n keys are in all 22 locales, and every contrast figure in the description reproduces.

mobile/lib/blocs/storage/storage_cubit.dart
hm21: Confirmed and fixed in e7eb31b626. Post-await emits in `loadRecoveryFootprint` now bail unless the cubit is still open and still `measuring`, so a measurement that lands after the confirm is dropped instead of clobbering `recovering`. Four `Completer`-driven tests cover it — all four go red without the guards.

mobile/lib/screens/settings/storage/storage_management_page.dart
hm21: Fixed in the same commit (e7eb31b626) — the guard is `isClosed || status != measuring`, so it covers both the close and the overtake. Added the same `isClosed` check to `recoverFromCorruptedCache`: the wipe takes seconds and back-navigation during it closes the cubit the same way.

mobile/lib/screens/settings/storage/storage_management_page.dart
hm21: Right — fixed in 796ec5a681. Split into one `BlocListener` per operation under a `MultiBlocListener`, each with its own `listenWhen`, so the message is keyed to the status that actually changed. Widget test captures `SystemChannels.accessibility` across clear-then-repair; against the old listener it reports "Cache cleared" instead of the repair outcome.

mobile/packages/divine_ui/test/src/switch/divine_switch_test.dart
hm21: Verified — I rebuilt the recolouring variant and the flip test passed unmodified, same for the checkbox one. Both now compare the resolved title/subtitle `TextStyle`s across the flip (f14967d2ee), which does go red on that implementation.

hm21:
All four addressed, @realmeylisdev — three fixes and one test that could not fail. Each claim was reproduced first: the mid-wipe overwrite and both emit-after-close paths now have regression tests that go red without the guards, and the flip tests were verified against the recolouring implementation they are supposed to catch.

- e7eb31b626 — measure/wipe results dropped when the cubit is closed or the wipe already started
- 796ec5a681 — one announcement listener per operation, keyed to the status that changed
- f14967d2ee — flip tests assert the resolved text styles instead of an absent exception

realmeylisdev (approved):
All four verified at f14967d2ee, each by reverting the fix and watching the new test go red: the mid-wipe test on the `!= measuring` half of the guard, both after-close tests on `isClosed`, the wipe test on its own guard, the clear-then-repair test on the pre-split listener, and both flip tests on the recolouring variant.

divine_ui 821 tests at 100% coverage, storage and settings suites green, `flutter analyze lib test integration_test` clean.

PR #6613, opened 2026-08-02
fix(nostr): stop offline relays from making every publish look partial

### LINKED ISSUE #6634: fix(nostr): count attempted publish relays as unreachable
## Problem

PR #6613 fixes inflated publish denominators by dropping configured relays that were never connected or attempted. After the expected/count split, one edge case remains: a relay can look unconnected before fan-out, still be genuinely attempted while its connection is in flight, and then time out or fail before the EVENT is written.

If that attempted relay is omitted from countedTargets, acceptedByAll can become true even though the publish tried a relay and did not reach it.

## User impact

Delete publishes and other callers that rely on PublishOutcome.acceptedByAll can report every-relay acceptance for a partial publish during launch or connectivity flaps.

## Acceptance criteria

- Configured relays that are plainly disconnected and never attempted still do not inflate the denominator.
- Relays the fan-out actually attempts but cannot write to are reported in unreachableTargets.
- Relays reached despite stale connection state can still answer their own publish.
- ContentDeletionService and PublishOutcome comments describe the final semantics.
- RelayPool-level regression coverage pins the attempted-but-unreachable case.

rabble:
## What

`RelayPool._intendedPublishTargets` used `relay.relayStatus.writeAccess` alone to decide which relays counted in a publish outcome. That flag is configuration and defaults to `true`, so configured relays the client had never connected to inflated the denominator.

A relay that was never connected can only land in `PublishOutcome.unreachableTargets`, and `acceptedByAll` requires that list to be empty. On mobile, where some configured relay is often down, that made `acceptedByAll` effectively unreachable and made successful publishes look partial.

## Fix

Publishing now separates three relay sets:

- `expectedRelays`: every write-enabled relay the fan-out may write to. This remains wide so relays whose connection state is stale or still in flight can answer their own publish.
- `countedTargets`: the denominator used for `unreachableTargets`. This starts narrower for unfiltered whole-pool `EVENT` publishes so plainly disconnected, never-attempted relays do not make every publish partial.
- attempted relays: `_sendCollect` now reports relays the fan-out genuinely attempted but did not write to, such as a relay whose `connecting` socket waited until the publish deadline or failed before the deadline. Those relays are added back to `countedTargets` and can be reported as unreachable.

`targetRelays` and `tempRelays` remain exempt from the narrowing. Naming a relay is the caller asserting intent for that specific relay, so an unreached named relay should still appear in the outcome.

This preserves the original fix without dropping healthy-but-still-connecting relays from `acceptedBy`, and without letting attempted-but-failed relays disappear from the denominator.

## Tests

`relay_pool_publish_targets_test.dart` covers:

- configured relay that never connected is not a publish target
- explicitly targeted relay is reported unreachable, not dropped
- disconnected relay cannot speak for an unfiltered publish
- still-connecting relay can answer its own publish
- attempted still-connecting relay is reported unreachable when the write times out
- attempted still-connecting relay is reported unreachable when the connect fails before the deadline

This also extracts the duplicated fake WebSocket test support and verifies `PublishTracker.countedTargets` does not expose mutable internal state.

## Verification

- `flutter test test/unit/relay_pool_publish_targets_test.dart` in `mobile/packages/nostr_sdk`
- `flutter test test/unit/publish_outcome_test.dart` in `mobile/packages/nostr_sdk`
- `flutter test test/unit/relay_pool_zombie_remediation_test.dart` in `mobile/packages/nostr_sdk`
- `flutter analyze lib test` in `mobile/packages/nostr_sdk`
- `flutter test` in `mobile/packages/nostr_sdk` - 422 passing
- `flutter test test/services/content_deletion_service_test.dart` in `mobile`
- pre-push hook: merge-conflict check, analyzer, changed-file tests

Closes #6634

### DIFF
--- a/mobile/lib/services/content_deletion_service.dart
+++ b/mobile/lib/services/content_deletion_service.dart
@@ -265,10 +265,10 @@ class ContentDeletionService {
 
       // One acceptance is the bar, not all of them. NIP-09 deletion is a
       // request relays SHOULD honour, so no acknowledgement proves the video
-      // is gone — requiring every relay buys a guarantee that does not exist
-      // and costs the user the feature, because the target set includes pool
-      // members the client never connected to and nothing here republishes.
-      // Breadth is reported on [DeleteResult.acceptance] instead.
+      // is gone. Requiring every targeted relay buys a guarantee that does not
+      // exist and costs the user the feature when a relay rejects, times out, or
+      // cannot be reached. Breadth is reported on [DeleteResult.acceptance]
+      // instead.
       if (publishOutcome.failed) {
         Log.error(
           'Delete publish not confirmed by any relay: '
@@ -321,10 +321,7 @@ class ContentDeletionService {
         name: 'ContentDeletionService',
         category: LogCategory.system,
       );
-      return DeleteResult.createSuccess(
-        deleteEvent.id,
-        acceptance: acceptance,
-      );
+      return DeleteResult.createSuccess(deleteEvent.id, acceptance: acceptance);
     } catch (e) {
       Log.error(
         'Failed to delete content: $e',
--- a/mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
+++ b/mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
@@ -22,7 +22,10 @@ import 'dart:async';
 ///
 /// Every relay the publish targeted appears in exactly one of [acceptedBy],
 /// [rejectedBy], [noResponseFrom] or [unreachableTargets], and no other relay
-/// appears at all.
+/// appears at all. "Targeted" means every relay the fan-out reached, plus the
+/// ones it was supposed to reach and could not — a configured relay whose
+/// socket was down when the publish started was never asked, and does not
+/// appear (see [PublishTracker.countedTargets]).
 class PublishOutcome {
   const PublishOutcome({
     required this.eventId,
@@ -149,8 +152,9 @@ class PublishTracker {
     this.deadline,
     required this.expectedRelays,
     required Duration timeout,
+    Set<String>? countedTargets,
     this.settleWindow = defaultSettleWindow,
-  }) {
+  }) : countedTargets = countedTargets ?? expectedRelays {
     _timer = Timer(timeout, _onTimeout);
   }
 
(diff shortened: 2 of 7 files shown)


hm21 (approved):
Took over one finding and pushed it as `74281c43fb` — please review it before merging.

**The connected filter was also gating who may answer.** `_intendedPublishTargets` feeds `PublishTracker.expectedRelays`, which `isTarget` uses to decide whether an incoming `OK` is allowed to speak for the publish. Narrowing it by `relayStatus.connected` meant a relay outside the set had its `OK true` dropped as "not a publish target" — for the rest of the sequential fan-out, up to `perRelaySendTimeout` per remaining relay.

`relayStatus.connected` can't carry that weight. It starts at `disconnect` and only advances when the state stream delivers, and a relay retrying a dead host reads `connecting` exactly like one whose first handshake is still in flight — I confirmed both by instrumenting the predicate. `send` waits the `connecting` state out and writes anyway, so those relays do get the frame and do answer.

Two healthy relays still connecting when the publish starts — right after launch, or after a connectivity flap:

| | `acceptedBy` |
| --- | --- |
| `main` | `[wss://relay.fast.example]` |
| `88ecbfcd16` | `[]` → `failed` |

Through `ContentDeletionService` that is "Couldn't delete this video" after a relay accepted the tombstone, with the deletion not persisted and `_repairSilentRelays` force-reconnecting the relay that answered.

**Fix:** split the two roles that one set was doing. `expectedRelays` goes back to every write-enabled relay so anything reachable can still answer; the new `PublishTracker.countedTargets` carries your connected filter and is read only by `unreachableTargets`, which is where the inflated denominator actually lived. Offline relays still drop out of the count, and the `targetRelays`/`tempRelays` exemption is untouched.

Your three tests still pass unchanged, and the widened `expectedRelays` gives the third one something real to guard. Added a fourth that fails on `88ecbfcd16` with `acceptedBy: []`.

Validation: `flutter test` in `packages/nostr_sdk` (419), `content_deletion_service` / `account_deletion_service` / `delete_result_localization` / `nostr_client_provider_bootstrap` (51), `flutter analyze lib test integration_test` and the package analyze clean, `dart format` clean.

One thing left for you: the description still says an `EVENT` target must be connected as well as write-enabled, which now describes only the counted set.

NotThatKindOfDrLiz:
@rabble I pushed takeover commit e8e86e1c9 to close the remaining review gaps.

What changed:
- Count relays that enter publish send while already `connecting` as attempted, so fast handshake failures before the publish deadline are reported in `unreachableTargets` instead of disappearing from the denominator.
- Added the missing regression for that fast connect-failure path.
- Made `PublishTracker.countedTargets` defensive: constructor copies the default set and the getter returns an unmodifiable set.
- Extracted the duplicated fake WebSocket channel/sink/factory into `test/support/fake_web_socket.dart` and rewired the relay-pool tests to use it.
- Updated the PR body to match the current behavior and verification.

Verification:
- `flutter test test/unit/relay_pool_publish_targets_test.dart` in `mobile/packages/nostr_sdk`
- `flutter test test/unit/publish_outcome_test.dart` in `mobile/packages/nostr_sdk`
- `flutter test test/unit/relay_pool_zombie_remediation_test.dart` in `mobile/packages/nostr_sdk`
- `flutter analyze lib test` in `mobile/packages/nostr_sdk`
- `flutter test` in `mobile/packages/nostr_sdk` (422 passing)
- `flutter test test/services/content_deletion_service_test.dart` in `mobile`
- pre-push hook: merge-conflict check, analyzer, changed-file tests
- GitHub checks are green.

One deliberate behavior call: the all-relays-plainly-disconnected blackout case still produces a zero-target `no relay reached` outcome. That matches the PR's narrowed denominator semantics: relays that were plainly disconnected and never attempted are excluded.

NotThatKindOfDrLiz (approved):
Approved after takeover. The remaining attempted-relay gap is covered by a regression test, counted target state is defensive, duplicated relay-pool WebSocket test support is extracted, PR body is current, and all GitHub checks are green.

Issue #5953, opened 2026-07-07
fix(feed): imported Vine clips with dead media (404 / legacy CDN) surface in feeds and fail to play


NotThatKindOfDrLiz:
## Summary
Imported classic-Vine clips whose media (and avatars) point at **dead legacy CDNs** surface in the For You / Popular feeds and fail to play, producing a wall of composition errors. Surfaced in #5933 (repeating-videos report) but is a distinct content-hygiene problem.

## Evidence (from #5933 logs)
```
[ERROR] [VIDEO] (DivineVideoPlayer.Load) Player 805176 composition failed: The requested URL was not found on this server.
[ERROR] [VIDEO] (InfiniteVideoFeed) Player init failed | Error: PlatformException(COMPOSITION_ERROR, The requested URL was not found on this server., null, null)
[WARNING] [VIDEO] (MediaCache) CancellableDownload: rejecting non-https url http://v.cdn.vine.co/r/avatars/9AEBB4A11A1388294775119466496_576682ad19d.35.0.jpg?versionId=...
[WARNING] [VIDEO] (MediaCache) CancellableDownload: rejecting non-https url http://web.archive.org/web/20170116125314/http://v.cdn.vine.co/r/avatars/...
[WARNING] [VIDEO] (VideoLoadingPlaceholder) Feed thumbnail load_failed: mode=forYou, index=19, eventId=0da2fc6e8dfc43394c7d154a4d5f77276bdd6848587c85a6b3d03b0b595d7812, thumbnailUrl=https://media.divine.video/cfe88a199d120de59eeffa01218d722b04a0b347634c12051e00ff8c61cbbbd6.jpg, error=MediaCacheImageProvider load cancelled
```
The composition failures repeat for many consecutive players in the same feed session — these items are unplayable, not transiently slow.

## Two angles
1. **Backend eligibility (primary):** the feed should not *serve* items whose media resolves to a known-dead host (`v.cdn.vine.co`, `web.archive.org`-wrapped legacy CDNs) or that 404. This overlaps **#5467 (feat(video): restrict served video eligibility)** and the classic-Vine import path (**#3714**). Likely needs a funnelcake/relay eligibility rule.
[...]

PR #6614, opened 2026-08-03
test(services): pin the 401 age gate out of the dead-media prune

realmeylisdev:
## Description

**This PR no longer changes behaviour.** It started as a fix for a regression that, on inspection, does not exist on `main` — see below. What survives is the documentation and the regression tests that keep it from being introduced.

### The original premise was wrong

The branch claimed the dead-media guard pruned age-gated media as if it were dead. It does not. On `origin/main`, `MediaAvailabilityChecker.isConfirmedMissing` is `statusCode == 404`, and all three paths that can prune hang off that single predicate:

- `fullscreen_feed_bloc.dart` `_onVideoUnavailable`
- `dead_media_feed_guard.dart` `confirmAndMarkMissing`
- `pooled_fullscreen_video_feed_screen.dart` `persistConfirmedUnavailable` (fired only by the bloc above)

A 401 has never reached `markVideoBroken`. The behaviour the PR body described as a live bug was the behaviour of #6251 *as proposed*, not of shipped code.

### The one thing it did change was itself a regression

The only behavioural delta was a new **401 → skip**, and @hm21's review is right that it fires exactly where it hurts. `FeedAutoAdvancePastErrorListener` is mounted only under `_OverlayInteractiveMode` (`feed_videos.dart:843`), so a correctly-classified 401 never reaches the guard — it routes to `_OverlayAgeRestrictedMode` and the Verify-age affordance. The skip could therefore only fire where the player *mis*classified the 401 (Android `PLAYER_ERROR "Source error"` → `generic`).

Correcting one detail from an earlier revision of this description: I described `FeedLoadingModerationCubit` as recovering that case *after* the error. It doesn't. It is mounted only inside `loadingBuilder` (`feed_videos.dart:422-434` → `:1158-1173`), and `InfiniteVideoFeed` renders `loadingBuilder` under `!hasError` and `errorBuilder` under `hasError` (`infinite_video_feed.dart:1834-1865`) — so the moderation verdict is a *loading-phase* recovery that is torn down the instant the item errors. That makes the skip worse, not better: for a mis-classified item the loading-phase verdict is the only remaining route to the age gate, and auto-advancing forecloses it permanently. The `_maybeFire` bailout doesn't help either — `PlaybackStatus.ready` is only reported from the manual `onRetry` tap (`feed_videos.dart:473`), never from the age-verify retry.

Reverted. `MediaAvailability`, `classify` and `DeadMediaVerdict` are gone with it, which also resolves both of @hm21's non-blocking notes.

### What's left

- **Docs** on `isConfirmedMissing` and `DeadMediaFeedGuard` recording why 404 is the only actionable status; that a 404 means "not retrievable" rather than "the bytes are gone" (blossom collapses `Banned`, `Deleted` and `Restricted` into it); and that 401 is the age gate, granted on `requester_pubkey.is_some()` rather than on the viewer's age.
- **Regression tests** pinning 401 and 403 out of the prune predicate at both the checker and the guard. The guard test drives a real `MediaAvailabilityChecker` over a 401 rather than stubbing the bool, so it fails if the predicate is ever widened.

**Related Issue:** Relates to #5953, relates to #6251

## Out of Scope

- The backend half — funnelcake serving events whose blob is blocked — is divinevideo/divine-funnelcake#681, addressed in divinevideo/divine-funnelcake#760.
- The age-gate affordance for a mis-classified 401, and authenticating the guard's HEAD probe (it probes anonymously while fullscreen playback sends a BUD-01 header). That is #6251, and this PR deliberately stops pre-empting it.
- The home scroller still doesn't prune broken videos from its own list: `video_feed_bloc.dart` and `videos_repository.dart` carry no `filterVideoList` / `BrokenVideoTracker` wiring, so a guard mark skips an item once and it returns on load-more, pull-to-refresh and cold start. Pre-existing; not touched here.
- Blossom's status stranding (a blob reported `age_restricted` by moderation-api while blossom's KV still says `Restricted`, so it 404s instead of 401s) — belongs in divine-blossom.

## Verification

- `flutter test` on `dead_media_feed_guard_test.dart`, `media_availability_checker_test.dart`, `broken_video_tracker_identity_test.dart`, `feed_auto_advance_error_listener_test.dart` — 26 pass.
- `flutter test test/blocs/fullscreen_feed/` — 91 pass (the other `isConfirmedMissing` consumer).
- The new tests were mutation-checked: adding `|| statusCode == 401` to the predicate turns exactly those two red, and nothing else.

## Type of Change

- [x] 🧹 Chore (documentation and test coverage; no behaviour change)

### DIFF
--- a/mobile/lib/screens/feed/feed_auto_advance_error_listener.dart
+++ b/mobile/lib/screens/feed/feed_auto_advance_error_listener.dart
@@ -15,13 +15,13 @@ import 'package:openvine/blocs/video_playback_status/video_playback_status_state
 /// Behavior by mode (see #5953):
 /// * **Auto mode** ([isAutoAdvanceActive] true) — skips immediately on any
 ///   non-ready status (unchanged), and additionally, when
-///   [confirmAndMarkMissing] is supplied, runs the guard (fire-and-forget) so a
-///   HEAD-confirmed hard-404 item is pruned from every surface on refetch.
+///   [confirmAndMarkMissing] is supplied, marks a HEAD-confirmed hard-404 item
+///   broken (fire-and-forget) so it is pruned from every surface on refetch.
 /// * **Manual scroll** — skips *only* when [confirmAndMarkMissing] resolves
-///   `true`, i.e. the guard HEAD-confirmed that the media cannot be fetched.
-///   Transient or inconclusive failures keep the error tile, so a network flake
-///   can't evict a valid video. Without a [confirmAndMarkMissing] callback,
-///   manual scroll never auto-skips (prior behavior).
+///   `true` (a HEAD-confirmed hard 404, which also marks it broken). Transient
+///   or non-404 failures keep the error tile, so a network flake can't evict a
+///   valid video. Without a [confirmAndMarkMissing] callback, manual scroll
+///   never auto-skips (prior behavior).
 ///
 /// Gated on [isActive] so background / preloaded pages can't yank the feed
 /// forward if they fail while the user is still on an earlier page.
@@ -42,9 +42,8 @@ class FeedAutoAdvancePastErrorListener extends StatefulWidget {
   final VoidCallback onSkipBrokenVideo;
   final Widget child;
 
-  /// Classifies the active item's media via a HEAD request and returns whether
-  /// the feed should advance past it. A confirmed hard 404 is additionally
-  /// marked broken; no other outcome is ever persisted.
+  /// Confirms (via a HEAD request) whether the active item's media is a hard
+  /// 404 and, if so, marks it broken. Returns `true` only for a confirmed 404.
   /// Injected by the feed item from `deadMediaFeedGuardProvider`.
   final Future<bool> Function()? confirmAndMarkMissing;
 
@@ -104,15 +103,15 @@ class _FeedAutoAdvancePastErrorListenerState
     final confirm = widget.confirmAndMarkMissing;
     if (widget.isAutoAdvanceActive) {
       // Auto already skips any non-ready item — preserve that immediately, and
-      // run the guard (fire-and-forget) so a hard 404 is pruned on refetch.
+      // mark a hard-404 item broken (fire-and-forget) so it's pruned on refetch.
       _firedForCurrentBreak = true;
       _deferSkip();
       if (confirm != null) unawaited(confirm());
       return;
     }
 
-    // Manual scroll: only skip once the guard HEAD-confirms the media can't be
-    // fetched. Transient / inconclusive failures keep the error tile.
+    // Manual scroll: only skip past a HEAD-confirmed hard 404 (which also
+    // marks it broken). Transient / non-404 failures keep the error tile.
     //
     // Don't latch `_firedForCurrentBreak` when the guard isn't loaded yet —
     // no confirmation attempt happened, so `didUpdateWidget` must be able to
@@ -122,9 +121,9 @@ class _FeedAutoAdvancePastErrorListenerState
     _firedForCurrentBreak = true;
     final videoId = widget.videoId;
     unawaited(
-      confirm().then((shouldAdvance) {
+      confirm().then((isMissing) {
         if (!mounted) return;
-        if (!shouldAdvance) return;
+        if (!isMissing) return;
         // The result raced a page change, a deactivation, or a recovery —
         // none of which should move a different/now-ready item.
         if (!widget.isActive) return;
--- a/mobile/lib/screens/feed/feed_auto_advance_error_listener.dart
+++ b/mobile/lib/screens/feed/feed_auto_advance_error_listener.dart
@@ -15,13 +15,13 @@ import 'package:openvine/blocs/video_playback_status/video_playback_status_state
 /// Behavior by mode (see #5953):
 /// * **Auto mode** ([isAutoAdvanceActive] true) — skips immediately on any
 ///   non-ready status (unchanged), and additionally, when
-///   [confirmAndMarkMissing] is supplied, marks a HEAD-confirmed hard-404 item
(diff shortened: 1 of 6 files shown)


hm21 (dismissed):
Two doc fixes pushed directly (c9e55a9, 125abdd) — doc-only, the enum mapping and the skip logic are untouched. Analyze clean, the 34 affected tests still pass.

One thing to settle before this merges: the **401 → skip** half.

Not pruning on 401 is right. Verified against `blossom-core/src/types.rs` (`AgeRestricted` → `AgeGated` → 401; `Banned`/`Deleted`/`Restricted` → 404), and I HEAD-probed 100 `classic=true` blobs independently: 7% 401, 6% 404, 0% 403. The class is real and about the size you measured.

But making 401 *skip* is new behavior, and it only fires where it hurts. `FeedAutoAdvancePastErrorListener` is mounted only in `_OverlayInteractiveMode` (`feed_videos.dart:843`). A 401 the player classifies correctly becomes `PlaybackStatus.ageRestricted` → `_OverlayAgeRestrictedMode` → "Verify age" + `autoRetryAgeRestrictedPooledVideo`, and the guard never runs at all. So the new skip only reaches items where the player *mis*classified the 401 as `generic`/`notFound` — the Android `PLAYER_ERROR "Source error"` case this PR cites. Net effect on Android: those ~7% now silently scroll past instead of reaching the age gate they'd otherwise get.

The bailout in `_maybeFire` doesn't cover it: `statusFor(videoId) == ready` depends on `cubit.report(..., ready)`, which only fires from the manual `onRetry` tap (`feed_videos.dart:473`), not from the age-verify retry. A ~200ms HEAD beats moderation-API + sha256 resolve + auth + player retry, so the skip wins that race.

Suggest dropping 401 from the skip set — 404 keeps `skipAndPrune`, 401 falls back to `keep`, and `skipOnly` goes away. That fixes the whole regression in the title with no new behavior, and leaves #6251 free to add the real age-gate affordance instead of pre-empting it.

Non-blocking: `MediaAvailability` re-encodes the 401/403/404 split `VideoErrorType` already carries; and `classify`/`DeadMediaVerdict` are public, but no production caller distinguishes `skipOnly` from `skipAndPrune` today.

hm21 (commented):
Two doc fixes pushed directly (c9e55a9, 125abdd) — doc-only, the enum mapping and the skip logic are untouched. Analyze clean, the 34 affected tests still pass.

One thing to settle before this merges: the **401 → skip** half.

Not pruning on 401 is right. Verified against `blossom-core/src/types.rs` (`AgeRestricted` → `AgeGated` → 401; `Banned`/`Deleted`/`Restricted` → 404), and I HEAD-probed 100 `classic=true` blobs independently: 7% 401, 6% 404, 0% 403. The class is real and about the size you measured.

But making 401 *skip* is new behavior, and it only fires where it hurts. `FeedAutoAdvancePastErrorListener` is mounted only in `_OverlayInteractiveMode` (`feed_videos.dart:843`). A 401 the player classifies correctly becomes `PlaybackStatus.ageRestricted` → `_OverlayAgeRestrictedMode` → "Verify age" + `autoRetryAgeRestrictedPooledVideo`, and the guard never runs at all. So the new skip only reaches items where the player *mis*classified the 401 as `generic`/`notFound` — the Android `PLAYER_ERROR "Source error"` case this PR cites. Net effect on Android: those ~7% now silently scroll past instead of reaching the age gate they'd otherwise get.

The bailout in `_maybeFire` doesn't cover it: `statusFor(videoId) == ready` depends on `cubit.report(..., ready)`, which only fires from the manual `onRetry` tap (`feed_videos.dart:473`), not from the age-verify retry. A ~200ms HEAD beats moderation-API + sha256 resolve + auth + player retry, so the skip wins that race.

Suggest dropping 401 from the skip set — 404 keeps `skipAndPrune`, 401 falls back to `keep`, and `skipOnly` goes away. That fixes the whole regression in the title with no new behavior, and leaves #6251 free to add the real age-gate affordance instead of pre-empting it.

Non-blocking: `MediaAvailability` re-encodes the 401/403/404 split `VideoErrorType` already carries; and `classify`/`DeadMediaVerdict` are public, but no production caller distinguishes `skipOnly` from `skipAndPrune` today.

realmeylisdev:
Agreed, and dropped — `424e9721` reverts to `main`'s behaviour.

Checking your citations turned up something worse than the skip: the regression this PR claimed to fix doesn't exist. `isConfirmedMissing` is `statusCode == 404` on `main`, and all three prune paths hang off that one predicate (`_onVideoUnavailable`, `confirmAndMarkMissing`, `persistConfirmedUnavailable`), so a 401 has never reached `markVideoBroken`. What I wrote up as shipped behaviour was #6251 as proposed.

That leaves the 401 → skip as the branch's only real delta, and your read of it holds: mounted only under `_OverlayInteractiveMode`, so it could fire solely on the mis-classified Android path — which is exactly where `FeedLoadingModerationCubit` recovers via moderation-api → auth → retry, and loses the race to a ~200ms HEAD.

`MediaAvailability` / `classify` / `DeadMediaVerdict` went with it, which covers both non-blocking notes. What's left is docs plus regression tests pinning 401 and 403 out of the predicate; the guard test drives a real checker over a 401 rather than stubbing the bool, so it goes red if the predicate is ever widened — checked by mutation. Your two doc commits are superseded, since both described the skip contract.

Title and body rewritten to match.

hm21 (approved):
The 401 skip is gone and the revert is clean — `isConfirmedMissing` is byte-identical to `main`, and `MediaAvailability` / `classify` / `DeadMediaVerdict` leave no dangling references. Both non-blocking notes resolved with it.

Re-ran the mutation check: adding `|| statusCode == 401` to the predicate turns exactly the two 401 tests red (guard + checker) and nothing else. 12 pass unmutated. CI green.

LGTM.

PR #6617, opened 2026-08-03
fix(lists): make owned lists visible and shareable

### LINKED ISSUE #6616: fix(lists): make owned lists visible and manageable
## Problem

Owned video lists are difficult to find and impossible to manage after creation:

- The own-profile tab shows global bookmarks, but the app has no in-app bookmark action.
- Existing owned lists only appear under Explore > Lists.
- Public/private status is not visible.
- A list name, description, and privacy cannot be changed.
- Public lists have no exposed share action or public URL.
- Opening an owned list through its internal route can show `0 videos` even when the grid contains videos.
- Private lists are device-only, but the UI does not communicate that limitation.

## Expected behavior

- The own-profile library tab shows owned lists.
- List cards and list details clearly label `Public` or `Private · On this device`.
- Owners can edit name, description, and visibility.
- Public/private transitions are confirmed and only complete after relay publication/deletion succeeds.
- Public lists can be shared at `https://divine.video/list/<pubkey>/<listId>`.
- List counts reflect the stored list contents.

## Acceptance criteria

- [ ] Replace the unreachable bookmark surface on the own profile with owned lists.
- [ ] Show visibility on owned list surfaces.
- [ ] Add owner edit/share/delete actions as appropriate.
- [ ] Preserve the previous visibility if the relay operation fails.
- [ ] Add regression coverage for list counts and visibility transitions.

rabble:
## Summary

- replace the own-profile bookmark tab with the owned Lists surface
- label owned lists as Public or Private · On this device and expose editing
- safely publish public/private transitions and preserve state on relay failure
- add sharing for public lists using canonical divine.video URLs
- fix owned list detail counts that incorrectly showed 0 videos

## Verification

- 205 focused list/profile/router/provider tests
- `mobile/scripts/golden.sh verify`
- localization consistency and design-system ratchets
- pre-push changed-file test suite (108 tests)

Closes #6616


hm21 (changes requested):
Pushed three fixes; one item still needs your call.

**Bookmarks are write-only after this.** #6616 says "the app has no in-app bookmark action", but there is one — the share sheet's Save entry (`share_sheet_more_actions.dart:71-75`) toggles a bookmark and reports "Added to bookmarks". The Saved tab was the only place to view them, so you can now bookmark a video and never see it again. Either direction works for me (keep a viewing surface, or retire bookmarking), but it's your product call. Whichever you pick, `ProfileSavedGrid`, `ProfileSavedVideosBloc` and `profile_saved_videos_sync_scope.dart` are dead once this lands — `requestProfileSavedVideosSyncIfAvailable` can never find the bloc again — so they should go in this PR rather than linger.

Pushed as separate commits, revert freely:

- `3d3c553` — share link now goes through `CuratedListByAuthorScreen.pathFor`, which exists so the GoRoute pattern and the link builder can't diverge.
- `b65a4ba` — the Share entry showed for any public list, but `_shareList` returns early when `pubkey` is null, so the tap was a silent no-op. Gated the entry on the same condition, with a test.
- `36d828a` — translated the 17 new keys into all 21 locales instead of parking them in `_knownUntranslatedDebt`; #6497 had emptied that set one commit earlier.

Validation: `flutter analyze lib test` clean, 124 list/l10n/profile tests green, `flutter gen-l10n` output committed.

mbradley (changes requested):
Second review, against `36d828a`. Deduped against @hm21's review, so the dead `ProfileSaved*` surface and the bookmarks product question are not repeated. Four questions are yours to answer; five findings are pushed as separate commits, revert freely.

## Questions

**1. A failed publish does not mean the event stayed on the device.**

`publishEvent` takes no deadline (`packages/nostr_client/lib/src/nostr_client.dart:421-424`), so `RelayPool` sends with `queueIfFailed: deadline == null` (`packages/nostr_sdk/lib/relay/relay_pool.dart:1307,1369`), `Relay.send` defaults to queuing (`packages/nostr_sdk/lib/relay/relay.dart:113`), and `onConnected` replays `pendingMessages` on reconnect (`relay.dart:69-98`). `PublishSuccess` is documented as "accepted and broadcast to at least one relay" (`publish_result.dart:20-26`), which is a frame write rather than a relay OK.

Two paths assume the opposite:

- `_createList` (`curated_list_service.dart:274-279`) removes the local record on failure while the signed kind 30005 event is still queued, producing a list on the relay with no local record. It reappears on the next cold start via `_processListEvent`, not within the session.
- `updateList` returns false when the kind 5 deletion fails and leaves the list marked public, while that deletion is still queued.

`PublishNoRelays` returns before sending (`nostr_client.dart:436-444`), so a fully offline user gets a clean rollback. The window is "relay reachable, send failed".

Options: pass a deadline so a failed publish is not queued; reconcile local state against the relay on reconnect; or drop the rollback and treat publish as fire and forget. Which one do you want?

**2. What should "make private" promise?**

`listMakePrivateWarning` reads "Its public link will stop working." The implementation publishes a NIP-09 kind 5 deletion request. NIP-09 states that relays SHOULD delete, that clients SHOULD broadcast the request to relays which do not have it, and that deletion is not guaranteed because it is impossible to delete events from all relays and clients. This client publishes once to its own pool and does not rebroadcast. `listVisibilityPrivateDevice` then labels that list identically to one that was never published.

Options: soften the copy to say relays are asked to remove it and already-shared copies may persist; keep the promise and add rebroadcast; or keep it as-is on the basis that the Divine relay honours deletions. The first option means retranslating one string across 21 locales.

**3. Should an offline edit be discarded?**

On `main`, `updateList` wrote `_lists[listIndex] = updatedList` and saved before any publish, so a rename persisted locally whether or not the relay accepted it. At `36d828a` the local write happens after the publish, so an authenticated user with no reachable relay loses the rename and the description edit, not only the visibility change. `createList` behaves the same way, so creating a public list offline now fails outright. The description says state is preserved on relay failure; that holds for `isPublic` and not for the rest of the edit.

Option: gate only the visibility transition on publish success and let name and description save locally.

**4. How do owned lists refresh?**

`fetchUserListsFromRelays()` returns early once `_hasSyncedWithRelays` is set (`curated_list_service.dart:1216`) and is only called from `initialize()`. `_syncCurrentTabIfNeeded` returns before adding `lists` to `_syncedKinds` (`profile_grid.dart:276`), so `_refreshSyncedTabs` never dispatches for that tab and pull to refresh does not reach the service. A list created on another device appears only after an app restart. The tab this replaces refreshed through `ProfileSavedVideosBloc`.

## Pushed

- `db49b13` — `updateList` captured `listIndex` before the publish awaits and wrote `_lists[listIndex]` afterwards. On the public branch that write is a no-op, because `_publishListToNostr` already stores the updated list plus its accepted event id by id and saves; it only added a second `notifyListeners` and prefs write per update. Persist only the private branch, and re-resolve the index by id there.
- `8ea4dc8` — a list going from public to private kept the event id of the event we had just asked relays to delete. `myLists` treats a non-null `nostrEventId` as published and then requires a pubkey match, so the list dropped out of My Lists once the owner signed out, while an always-private list stayed. `copyWith` gains `clearNostrEventId`, matching the `clear*` convention in `DmConversation`, `VideoEvent` and `AudioEvent`.
- `b529350` — the edit dialog passes the trimmed controller text straight through, so emptying the description stored `''`. `list_card` renders its description block on a `!= null` check, so the card showed an empty gap, and `_publishListToNostr` built empty event content instead of the fallback. `updateList` now reads an explicitly empty description as "clear it", which is the only way the dialog's existing signature can express that. This one encodes an API convention rather than being purely mechanical, so it is the most likely of the five to want your input.
- `b9187db` — nothing covered a private-to-public transition whose publish is rejected, and nothing asserted that a public update keeps the event id, so writing `updatedList` back over the enriched copy went unnoticed.
- `1415470` — five assertions matched hardcoded English while their siblings in the same files already resolved through `AppLocalizations`.

## Validation

`flutter analyze lib test` clean, and `flutter analyze` clean from inside `packages/models`. 108 tests across the seven files this PR touches, 165 across the other `CuratedList` consumers (run because `copyWith` and `updateList` reach past this PR's files), and the full 837-test `models` suite. Each new test was mutation-checked: removing the guard it covers turns it red.

## Not addressed

Beyond the dead files you listed, @hm21: `requestProfileSavedVideosSyncIfAvailable` still runs from live code at `share_action_button.dart:314-318` and can now only take the `ProviderNotFoundException` path, since no `BlocProvider<ProfileSavedVideosBloc>` remains in `lib`. `inheritedLookupContext` exists only to feed that call, and its doc comment at `:132-134` describes the removed bloc. If the Saved files go, `profileNoSavedVideosTitle`, `profileSavedOwnEmpty` and `profileErrorLoadingSaved` (`app_en.arb:361-363`) lose their only usage across 21 locales, and `scripts/baseline/ui_service_imports.txt:62` needs shrinking. Left alone because its shape depends on the bookmarks product call.

Requesting changes for the four questions above. Please review the pushed commits and merge when you are ready, or push back on any of them.

rabble:
Answering the four questions at `1d35f77`. All five pushed commits kept.

**1. Failed publish / rollback.** Took the deadline option. `publishEventAwaitOk` already carries one, so the SDK's `queueIfFailed: deadline == null` branch (`relay_pool.dart:1624`) is not taken and a `false` return means no relay took the event. Routed the publishes whose result gates local state — `_createList`, `updateList`'s public branch, `_publishListDeletion` — through it, gated on `acceptedByAny` to match `content_deletion_service.dart:253-272`. The republish-after-edit paths (`addVideoToList`, `removeVideoFromList`, reorder, collaborators) keep `publishEvent`: they ignore the result, and queue-and-replay is the behaviour you want for an add-to-list made offline. The phantom-list window is closed.

**2. What "make private" promises.** Softened the copy, retranslated across 21 locales. It now says relays are asked to remove the public copy and that already-shared copies may stay online. Rebroadcast was not worth buying a guarantee NIP-09 does not offer.

**3. Offline edits.** Fixed, and scoped slightly wider than you framed it. Name and description are local-first — this device is where they live, the relay copy is a projection — so `updateList` writes them before any publish await. Only the visibility transition still waits for confirmation, because that one is a statement about what relays hold.

This inverts one of your assertions: `keeps a list private when publication is rejected` asserted `stored.name == 'Private List'`, i.e. the rename was discarded along with the visibility change. It now asserts `'Renamed'` while `isPublic` stays false. Added `keeps a rename when the relay rejects a public list update` for the already-public case, and `confirms relay acceptance before rolling an update back` to pin the API choice from (1).

`_createList`'s rollback stays. With (1) it is now truthful, and it keeps your `does not claim a failed publication is a public list` intact.

**4. Refresh.** `ProfileTabKind.lists` now joins `_syncedKinds` on first view — it has nothing to dispatch, it renders from the service — and `_refreshSyncedTabs` calls `fetchUserListsFromRelays(force: true)`. The `force` flag is new: without it the service returns early on `_hasSyncedWithRelays` and pull-to-refresh would have been a no-op even once wired. Covered at both levels: `re-queries relays only when the fetch is forced` in the service tests, `pull-to-refresh re-queries relays for the lists tab` in `profile_grid_test`.

**Still open, @hm21's:** the bookmarks product call and the dead `ProfileSaved*` surface. Not addressed here — it is a product decision, and the dead-code sweep depends on which way it goes.

**Follow-up, not this PR:** private lists never reach a relay at all today, so they die with the device. Next PR publishes them as NIP-51 kind 30005 with the item tags NIP-44 self-encrypted into `content`, which also replaces the kind 5 deletion in the make-private path with a plain addressable republish — and will rewrite the string from (2) again. `CuratedListConverter` needs an async encrypted path for that, and `content` needs untangling from its current double duty as the description.

**Validation:** `flutter analyze lib test` clean. 542 tests across the list service suites, list/profile widgets, the screen tests, `test/l10n`, and the other `CuratedList` consumers. Pre-push changed-file suite green (262).

hm21 (approved):
Re-review of `1415470..84a9c96`. LGTM — all four answers hold up against the code: `publishEventAwaitOk` really does bypass the SDK's reconnect queue (`queueIfFailed: false` under a deadline), the local-first/confirmed-visibility split in `updateList` traces clean through all four visibility paths, and `76798e4` settles the bookmarks question while clearing every item from the earlier "Not addressed" list (sync-scope plumbing, arb keys back in use, baseline already shrunk).

Pushed `b99e547` (docs-only, revert freely): `1d35f77` appended the new `_publishListToNostr` contract below the old doc block instead of replacing it, leaving a stale "Returns `true` when the relay accepted the event" line that contradicts the unconfirmed path. Analyze clean, 64 list/profile tests green, CI was fully green on `84a9c96`.

mbradley (dismissed):
Third review, against `92452d2d`. The commits since `1415470` are new ground — the bookmarks reading surface, the `/saved` route and screen, the acceptance-gated publishes, the refresh wiring — so this covers those. Deduped against @hm21's approval at `b99e547`.

## Questions

**1. `_publishListToNostr` writes back the list it captured before the publish.**

`curated_list_service.dart:1157` does `_lists[listIndex] = list.copyWith(nostrEventId: event.id)`. The index is re-resolved by ID; the value is not. Any local mutation that landed during the await is discarded.

That line predates this branch, but its exposure doesn't. Under `confirmed: true` the await above it is `publishEventAwaitOk`, which waits for OK frames to a 15s default deadline (`nostr_client.dart:501`) rather than returning on the socket write. The clobber window went from a frame write to seconds.

Concretely: make a private list public against a slow relay. Mid-publish the user adds a video — `addVideoToList` (`:325-332`) writes `videoEventIds: [X]` locally and publishes, so the relay holds X. The first publish confirms and `:1157` writes back the captured copy with an empty `videoEventIds`. The video is on the relay and gone from the device until a later relay sync repairs it.

Options: apply only the event ID onto the current entry (`_lists[listIndex].copyWith(nostrEventId: event.id)`, carrying `isPublic` from `list`, since the public branch depends on this write to commit the confirmed visibility); serialize publishes per list; or accept the window. Which?

**2. Bookmarks have no refresh path.**

`92452d2d` removed the `ProfileTabKind.saved` case from profile pull-to-refresh and `requestProfileSavedVideosSyncIfAvailable` with it. `SavedVideosScreen` has no `RefreshIndicator`, and neither does `ProfileSavedGrid` — the only sync is the one-shot `ProfileSavedVideosSyncRequested` at mount. On `main` the Saved tab refreshed. Now unbookmarking from a feed opened off `/saved-videos` and going back leaves the video listed until the screen is remounted.

Options: a `RefreshIndicator` on the screen (`ProfileSavedGrid`'s `ClampingScrollPhysics` would need to become `AlwaysScrollableScrollPhysics`); re-sync on route resume; or bookmarks are mount-time-only by design.

**3. `ProfileSavedVideosSyncRequested.completer` is dead in live code.** The only dispatch is `const ProfileSavedVideosSyncRequested()` at `saved_videos_screen.dart:53`; the remaining `completer:` callers are its own bloc test. Delete it, or keep it because it's what a refresh in (2) would use? Contingent on (2).

**4. The My Lists block is duplicated.** `profile_lists_grid.dart:29-32,83-91` and `explore_lists_tab.dart:111-115,145-162` share the same `service?.myLists` read, the same `CuratedListCard(showVisibility: true)`, and the same push-with-`CuratedListRouteExtra` tap. This PR edits both. Extract a shared section, or accept the drift?

**5. The Lists tab isn't gated on `FeatureFlag.curatedLists`.** `profileTabKinds` returns it unconditionally for the own profile and `curatedListsStateProvider` isn't gated, while `share_sheet_more_actions.dart:43` hides Add-to-list behind the flag. The flag defaults true, so this only shows up as a kill switch that no longer kills the surface. Intended?

## Pushed

- `7b8b8d0` — `deleteOwnedList` captured `listIndex` before the deletion publish and removed positionally. Moving that publish to `publishEventAwaitOk` grew the wait to a 15s deadline, long enough for another removal to shift the index and take the wrong list with it while the intended one stayed local and gone from the relay. `updateList` already re-resolves by ID across the same awaits. Regression test mutation-checked.
- `a0cf9c5` — the bookmarks icon passed no `color`. `DivineIcon` applies no `colorFilter` when it's null and `bookmark_simple.svg` is a hardcoded white fill, so it renders white on light while the label and caret beside it take theme colours. Same shape `43b2957` fixed for text; the icon ratchets don't cover colour.
- `1affb2d` — the local-first write already stores the whole updated list, so for an already-private list the later block wrote an identical value and saved again: two `notifyListeners` and two prefs writes per edit. Narrowed to the `becamePrivate` transition, restoring what `f438cd4` removed.
- `bf7a980` — deleting the `/saved-videos` `GoRoute` left `test/router` and `test/widgets/profile` green (621 passing). The Lists-tab tap test declares its own `GoRouter`, so it pins the push target and nothing about the app having the route. New test drives the real `goRouterProvider` and asserts the screen renders, which the error page doesn't satisfy.
- `47daf65` — renamed the route to `/saved-videos`, matching its own `routeName` and the sibling `/liked-videos`. New on this branch, never shipped.
- `94308ed` — the confirmed-publish doc cited `relay_pool.dart:1624`; that call is at 1944 and 1624 is subscription completion.
- `54fc4af` — the saved grid's docs still name the `NestedScrollView` it took its `PrimaryScrollController` from and call itself a profile surface; it's now a standalone route.
- `a82cd1e` — `Row(spacing: 6)` in the new visibility badge, matching `_BookmarksEntry` in the same branch.

## Validation

`flutter analyze lib test` clean. 522 tests across the seven `curated_list_service_*` suites, `test/widgets/profile`, `list_card`, `add_to_list_dialog`, `curated_list_feed_screen`, and the new router test. Each new test mutation-checked: reverting the guard it covers turns it red. Pre-push changed-file suite green (297).

## Not pushed

- **A `SavedVideosScreen` test for the `ValueKey((videosRepository, currentUserPubkey))` guard.** Driving it needs `nostrServiceProvider` to change identity, and `currentUserPubkey` is read off the watched service object rather than watched itself — so whether the key ever flips depends on the client instance being recreated on account switch. That's the claim the comment makes; worth confirming before a test pins it. The new router test covers that the screen mounts and its `PrimaryScrollController` satisfies the grid's pagination.
- **`scripts/baseline/file_sizes.txt`** (advisory, never fails CI) isn't ratcheted for `curated_list_service.dart` — but it's already stale on `origin/main` for ~24 unrelated files, so regenerating here would sweep them in.

Please review the pushed commits and merge when you're ready, or push back on any of them.

rabble:
**Answers to the third review, plus your eight commits.** All five are addressed in `76e8179`, `54583ca`, `fdd1e1d`, `da9453a`.

**Your pushed commits: all eight accepted, nothing reverted.** `7b8b8d0` is the same bug class as Q1 on the delete path and the regression test earns its place. `1affb2d` correctly narrows to `becamePrivate` — the local-first write at `:470` had already made the private branch a no-op write. `a0cf9c5` matches what `43b2957` did for the label. The rest are a test, a rename, and docs.

### 1. Write-back clobber — fixed, narrow write (`76e8179`)

`_publishListToNostr` now applies the event ID onto the current entry:

```dart
_lists[listIndex] = _lists[listIndex].copyWith(
  nostrEventId: event.id,
  isPublic: list.isPublic,
);
```

`isPublic` still comes from the captured list, since `updateList:470` deliberately holds the old visibility and this write is what commits it — your option 1, for the reason you gave.

One correction to the repro. Private→public does not strand the video on the relay: `:470` keeps the list private for the duration, so a concurrent `addVideoToList` sees `isPublic == false` at `:335` and never publishes. The video is lost from the device **and** never reaches the relay. The relay/device divergence you described is the **public→public** path — rename a public list, add a video mid-publish. Both are fixed by the same change; the regression test drives the rename case.

### 2. Bookmarks refresh — fixed (`54583ca`)

`RefreshIndicator` wired to the same completer-backed sync the profile used. The settled empty and error states opt into `AlwaysScrollableScrollPhysics` via a new parameter on `ProfileTabEmptyState`/`ProfileTabErrorState`, defaulted to clamping so the six other grids are untouched — without that, refresh dies exactly where it is most wanted, after unbookmarking the last video. Body split out as `SavedVideosView` so the pull can be driven against a mocked bloc instead of the screen's real bookmark and video dependencies.

### 3. `completer` — kept

It has a live caller now.

### 4. Duplication — extracted the card and tap only (`fdd1e1d`)

New `OwnedListCard` owns `showVisibility: true` plus the push and its extra. Both hosts keep their own chrome, since a management surface and a discovery section genuinely want different framing.

Deliberately not covered here: two other sites push `CuratedListFeedScreen.pathForId` inline, and they disagree on the payload — `search_results/widgets/lists_section.dart` seeds `videoIds` and `authorPubkey`, the others send only `listName`. That is a real question about which shape is correct, not just tidiness, so it is #6749 rather than a bigger diff here.

### 5. Flag — going the other way: ungated (`da9453a`)

This one reverses what you asked, so the reasoning matters.

Video lists are shipping, so I dropped the `showCuratedLists` guard rather than adding one. The inconsistency you found is real, but gating was the wrong direction: the Lists tab is the only place bookmarks are reachable, and the share sheet's Save action at `:71-75` is unconditional. Gating the tab would have made bookmarks write-only again — the bug @hm21 opened his first review with.

That guard was also the last place the flag touched kind 30005. Every other consumer gates people lists (kind 30000), which three in-code comments already call its real job; the description now says so. The name stays — renaming churns `FF_CURATED_LISTS` and 14 consumers for no behaviour change.

### Validation

`flutter analyze lib test integration_test` clean. 919 scoped tests: all ten `curated_list_service_*` suites, `test/widgets/profile`, `test/router`, share menu, add-to-list, saved-videos, profile-saved bloc, l10n consistency, feature flags. Four design-system ceilings and both structure floors pass. `golden.sh verify` passes. Each new test mutation-checked — reverting the narrow write, the empty-state physics, or re-adding the flag guard turns its test red.

One CI failure along the way was `upload_manager_get_by_path` on shard 3, unrelated to this diff: 15 suites share a process-wide `pending_uploads` Hive box and a stale `UploadManager` writes into it after a later suite clears it. Adding two test files re-dealt the shards and drew the bad neighbour. Re-run of the identical commit with a fresh seed is green; written up as #6748 rather than patched over here.

Merging on @hm21's approval with your blocking items addressed — dismissing the CHANGES_REQUESTED against this comment rather than leaving it to lapse silently. Q5 is the one to push back on if you disagree; happy to revisit it in a follow-up.

PR #6618, opened 2026-08-03
fix(report): pin the report sheet's submit action and finish the profile snackbar sweep

### LINKED ISSUE #6619: fix(report): pin the report sheet's submit action and finish the profile snackbar sweep
Auto-created from PR #6618.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6618
**Branch:** `fix/report-sheet-footer-and-profile-snackbars`

hm21:
Closes #6619

## Description

Two report-sheet defects plus the last raw snackbars on the profile surfaces.

**The report sheet could not be dragged closed.** It was handed to `VineBottomSheet` as `body:`, and only the `buildScrollBody:` slot receives the `DraggableScrollableSheet`'s `ScrollController`. The content therefore scrolled on a private controller, the sheet's own controller had no attached scrollable, and a downward drag was swallowed by the content instead of shrinking the sheet — so `DraggableScrollableNotification` never reached min extent and `showModalBottomSheet` never closed. All three entry points (`show`, `showForMessage`, `showForUser`) now use `buildScrollBody:` and thread the controller into the scroll view.

**Submit scrolled away with the form and was tappable on an empty form.** It now rides in a pinned footer and stays disabled until a reason is picked. The footer is keyboard-aware the same way `VineBottomSheet`'s `bottomInput` slot is — the "Other" details field sits in the scroll area above it, and a footer flush with the screen bottom would end up under the keyboard.

The confirmation view's actions moved into that same footer. That is what allows a single `SingleChildScrollView` to span both states: swapping one scroll view for another would briefly attach two positions to the sheet's controller during the rebuild. The oversized `build` was split into `_ReportFormBody`, `_InlineError` and the confirmation widgets along the way. The confirmation state (`ReportConfirmationBody`, `ReportConfirmationActions`) lives in its own `report_content_confirmation.dart`, which takes the dialog file from 886 back to 703 lines — under the advisory 800-line ceiling, so its baseline entry is dropped rather than raised.

**The inline error follows the pinned action.** Pinning submit let the user submit from any scroll offset, but the error still rendered at the end of the scroll content — below eleven reason cards. On a 412x915 device, submitting from the top with "Spam" selected put the error at y=1692, roughly 780px past the fold: the spinner stopped and nothing else appeared to happen. The pre-pinning layout hid this because the in-flow button forced the user to the bottom, so the error was always next to it. The error now rides the footer directly above the button. The regression test asserts the error's rect falls inside the viewport at phone size; it fails at 1708.25 > 915.0 without the fix.

The footer itself is no longer bespoke: `VineBottomSheet`'s private `_KeyboardAwareBottomInput` is promoted to a shared `VineKeyboardAwareFooter` in `divine_ui` rather than keeping a second copy of the same clearance and animation constants here.

**Profile snackbars.** Block, unblock, unfollow, copy, embed-code, share failures, follow-toggle failures, notify-bell failures and profile-feed load-more errors were still raw Material `SnackBar`s rendering in default dark grey. They now go through `DivineSnackbarContainer`, with `error: true` on the failure cases. Two `// TODO(SofiaRey): revisit when designs are ready` comments on the block/unblock banners are dropped — the design-system banner is what they were waiting for.

One behavioural change worth calling out: the test `Submit button is visible even before selecting a reason (Apple requirement)` also asserted the button was *enabled*. Apple's requirement is that the reporting affordance is present and reachable, which it still is; the test now pins visible-but-disabled instead.

**Related Issue:** 

## Out of Scope

- The l10n key `reportSelectReason` is now unused in code (the "no reason picked" path is unreachable with the button disabled). Left in the ARBs rather than deleting it from 18 locales in a fix PR.
- On Android 13+ the system shows its own clipboard confirmation on top of our "copied" snackbar. Android's own guidance is to drop the app-side confirmation at API 33+, but `minSdk` is 28 and iOS never shows one, so suppressing it needs a cached `sdkInt` check — separate change.
- ~20 raw `SnackBar` call sites remain elsewhere in `mobile/lib` (comments, inbox, video editor, curated lists). Not touched here to keep the diff reviewable.

## Verification

- `flutter test` on `report_content_dialog_test.dart`, `other_profile_screen_test.dart`, `profile_screen_router_test.dart`, `follow_from_profile_button_test.dart`, `profile_action_buttons_widget_test.dart`, `conversation_view_test.dart` — 99 passed
- New regression test: dragging the sheet content down dismisses the sheet. Red before this change, green after.
- Both defects were reproduced on a real Samsung Galaxy before the fix. Manual re-verification of the fixed sheet on device is still pending — hence draft.

## Type of Change

- [x] 🧹 Code refactor

### DIFF
--- a/mobile/lib/widgets/report_content_confirmation.dart
+++ b/mobile/lib/widgets/report_content_confirmation.dart
@@ -1,0 +1,175 @@
+// ABOUTME: Post-submission confirmation state of the report bottom sheet.
+// ABOUTME: Split into a scrollable body and the actions that ride the footer.
+
+import 'package:divine_ui/divine_ui.dart';
+import 'package:dm_repository/dm_repository.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
+import 'package:openvine/l10n/l10n.dart';
+import 'package:openvine/providers/app_providers.dart';
+import 'package:openvine/screens/inbox/conversation/conversation_page.dart';
+import 'package:url_launcher/url_launcher.dart';
+
+/// Scrollable content shown after a report is accepted.
+class ReportConfirmationBody extends StatelessWidget {
+  /// Creates a [ReportConfirmationBody].
+  const ReportConfirmationBody({required this.moderationDmFailed, super.key});
+
+  /// Whether the secondary NIP-17 DM to the moderation team failed to
+  /// send. The report itself still succeeded; this only drives a calm
+  /// informational notice so the user isn't misled.
+  final bool moderationDmFailed;
+
+  @override
+  Widget build(BuildContext context) {
+    final l10n = context.l10n;
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        const SizedBox(height: 8),
+        Row(
+          spacing: 12,
+          children: [
+            const DivineIcon(
+              icon: DivineIconName.checkCircle,
+              color: VineTheme.vineGreen,
+              size: 28,
+            ),
+            Expanded(
+              child: Text(
+                l10n.reportReceivedTitle,
+                style: VineTheme.titleMediumFont(
+                  color: context.vineColors.primaryText,
+                ),
+              ),
+            ),
+          ],
+        ),
+        const SizedBox(height: 16),
+        Text(
+          l10n.reportReceivedThankYou,
+          style: VineTheme.bodyLargeFont(color: context.vineColors.primaryText),
+        ),
+        const SizedBox(height: 16),
+        Text(
+          l10n.reportReceivedReviewNotice,
+          style: VineTheme.bodyMediumFont(
+            color: context.vineColors.onSurfaceMuted,
+          ),
+        ),
+        if (moderationDmFailed) ...[
+          const SizedBox(height: 12),
+          Text(
+            l10n.reportModerationDmDelayed,
+            style: VineTheme.bodySmallFont(
+              color: context.vineColors.onSurfaceMuted,
+            ),
+          ),
+        ],
+        const SizedBox(height: 8),
+        const _SafetyPolicyLink(),
+      ],
+    );
+  }
+}
+
+class _SafetyPolicyLink extends StatelessWidget {
+  const _SafetyPolicyLink();
+
+  @override
+  Widget build(BuildContext context) {
+    final l10n = context.l10n;
+    return Semantics(
+      button: true,
+      label: l10n.reportSafetyUrl,
+      child: GestureDetector(
+        onTap: () async {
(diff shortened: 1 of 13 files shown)


mobile/lib/widgets/report_content_confirmation.dart
```diff
@@ -0,0 +1,177 @@
...
+  }
+}
+
+class _SafetyPolicyLink extends StatelessWidget {
+  const _SafetyPolicyLink();
+
+  @override
+  Widget build(BuildContext context) {
+    final l10n = context.l10n;
+    // No `label:` here — the Text.rich below already supplies one, and a
```
realmeylisdev, line 83: Fixed in f8e59ce, flagging so you can see the reasoning.

Wrapping the link in `Semantics(button: true, ...)` is the right call, but the `label:` alongside it was additive rather than a replacement. The annotation is not a merge boundary, so the child `Text.rich` label got appended to it. Measured on the rendered widget, the node came out as:

```
label=<divine.video/safety\nLearn more at divine.video/safety> button=true tap=true
```

So VoiceOver/TalkBack announced the URL, then the whole sentence containing it again. Dropping `label:` leaves `label=<Learn more at divine.video/safety>` with `button=true` and the tap action intact, which is the better wording anyway.

`report_content_confirmation_test.dart` pins it; putting `label:` back turns it red.

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -530,148 +448,153 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
...
+}
+
+class _InlineError extends StatelessWidget {
+  const _InlineError({required this.message});
+
+  final String message;
+
+  @override
+  Widget build(BuildContext context) {
+    // `container` makes this its own node and absorbs the Text below, so the
```
realmeylisdev, line 567: Same shape as the safety link, fixed in the same commit.

`container: true` already gives this its own node and absorbs the `Text` below it, so `label: message` prepended a second copy and the banner read as `Failed to report content\nFailed to report content`. This one predates the PR (it moved verbatim out of the old `build`), but since `_InlineError` is a new class here it seemed better to fix than to leave next to the one that was new.

Verified the live region still announces after the change, the assertion in `report_content_dialog_test.dart` checks both the label and `isLiveRegion`.

realmeylisdev (approved):
lgtm, both defects check out. i mutation-tested the two new regression tests too (swapped the sheet back to a private scroll controller, and moved the error back into the scroll content) and they do go red without the fix.

pushed f8e59ce on top for one thing: the two new `Semantics` wrappers each set a `label:` over a child that already had one, and flutter prepends rather than replaces, so a screen reader read both twice. dropped the labels, added assertions that pin it. revert freely.

minor, the out-of-scope note undercounts. raw `SnackBar(` left in `mobile/lib` is 137 across 55 files, not ~20, and it reaches settings, auth, followers and profile setup too.

Issue #5210, opened 2026-06-15
fix(profile): fully load liked videos


hm21:
### Summary

Liked videos in the profile only show a limited number of items and do not continue loading the full list.

### Environment

android
main-branch

### Steps to Reproduce

1. Open a user profile with a large number of liked videos.
2. Navigate to the Liked Videos tab.
3. Scroll to the bottom of the loaded list.

### Actual Result

Only a subset of liked videos is displayed, and additional videos are not loaded when scrolling.

### Expected Result

All liked videos should load progressively as the user scrolls through the list.

### Regression?

Yes

### Evidence

<img width="322" alt="Image" src="https://github.com/user-attachments/assets/024e529a-8785-4332-9619-84d74751794d" />

PR #6621, opened 2026-08-03
fix(profile): stop ID-list grids locking into a sub-page window

### LINKED ISSUE #6622: fix(profile): stop ID-list grids locking into a sub-page window
Auto-created from PR #6621.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6621
**Branch:** `fix/profile-liked-videos-window-lock`

hm21:
Closes #6622

## Description

The profile Liked tab could get permanently stuck showing a handful of videos. Reproduced on a real account with 254 likes on the relay that rendered only 6 tiles, across restarts.

**Root cause.** `_reconcile` sized its window as `max(nextPageOffset, videos.length) + newAtTop` — bounded by what was already loaded, with no lower bound. Once a window ended up shorter than a page it could never grow again:

- The cold path prefers the local liked-ID list and syncs the relay only in the background. If that local list has not finished syncing, the first load resolves whatever few IDs it has and persists `nextPageOffset: 6`.
- The background sync then delivers the full list, but reconciliation inherits the 6-item window and writes it straight back to `CacheSync`.
- Every restart repeats that: the snapshot is served, revalidation reconciles, the window stays 6, the snapshot is re-persisted.
- The only escape would be pagination, but `ScrollPaginationMixin` fires on scroll events only. Six tiles are two rows — profile header plus two rows fits the viewport, nothing scrolls, `onLoadMore` never runs. `hasMoreContent` sat at `true` the whole time and was never redeemable.

Evidence from the affected device's `cache_sync.db`:

```
videos: 6   itemIds: 200   nextPageOffset: 6   hasMoreContent: true
```

...with the 6 videos being exactly the 6 most recent likes. Relay-side the data was fine: 228 of 250 liked IDs survive the full parse and filter chain, so nothing was actually missing.

**Fix.** Two changes, applied to all three ID-list-backed tabs (Liked, Saved, Reposts) since they share the same reconcile shape:

1. Floor the reconcile window at one page, so a window that was once short can recover.
2. Bypass the unchanged-IDs fast path in `_warmRevalidate` when the window is underfilled. Without this the fix would only reach users whose ID list is still changing — with ≤200 items the cached and fresh lists are identical, `listEquals` short-circuits, and the short window survives forever.

The underfill predicate counts consumed **IDs**, not resolved videos. Liked IDs are sparse (roughly 8% never resolve to a playable video), so a video-count predicate would stay true after the top-up and re-reconcile on every reopen. Counting IDs makes it fire exactly once and then stay on the cheap path.

Affected installs heal themselves on next launch — the window widens from 6 to 18 and pagination takes over from there. No cache migration needed.

**Related Issue:** 

Same user-visible symptom as #5210 and #5325, both of which were different mechanisms (pagination not continuing, and a SQLite/​cache drift in `watchLikedEventIds`).

## Out of Scope

`ScrollPaginationMixin` still keys purely off scroll events, so a list shorter than the viewport with `hasMoreContent: true` cannot request more on its own. Fixing it properly touches all 11 call sites and belongs in its own PR.

That weakness leaves one residual case this PR does not close. The floor is one page of **IDs**: `_reconcile` sets `nextPageOffset` to the IDs it consumed, while the grid renders only the subset that resolves — and `_fetchVideos` drops unsupported formats and blocklist-filtered authors on top of dead events. A window whose 18 IDs resolve to a handful of tiles therefore lands on `nextPageOffset: 18`, `hasMoreContent: true`, and an underfill predicate that is now false: the same non-scrolling dead end, one offset further out. Note the asymmetry — `_coldLoad` fills a page of *videos* via `_fetchVideoPage`, `_reconcile` does not. Closing that here by making `_reconcile` fill a page of videos would reintroduce exactly the unbounded re-fetch it was written to avoid (pinned by the capped-snapshot test), so it belongs with the mixin fix, which closes it generically for every tab regardless of cause. At the ~8% sparsity measured on the affected account this needs roughly 90% of a window to be unresolvable, so it is a narrow tail, not the reported bug.

## Verification

- 4 new regression tests: the unchanged-ID growth case on each of the three tabs, plus the changed-ID refill case on Liked. Each was checked individually against a stashed `lib/` fix and fails without it.
- `flutter test test/blocs test/widgets/profile` → 3072 passed
- `flutter analyze lib test` → clean
- Manually verified on a real Samsung Galaxy (SM-S942B): the Liked tab recovered from 6 tiles to the full list on the next launch, and scrolling continues to paginate.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/profile_liked_videos/profile_liked_videos_bloc.dart
+++ b/mobile/lib/blocs/profile_liked_videos/profile_liked_videos_bloc.dart
@@ -253,7 +253,9 @@ class ProfileLikedVideosBloc
   /// window: when the ID list is unchanged (the common reopen) this is just a
   /// flag flip; when it changed, [_reconcile] drops unliked videos and fetches
   /// only the newly-liked IDs in the window, so the UI thread never does bulk
-  /// video work here.
+  /// video work here. The one exception is an underfilled window
+  /// ([_isWindowUnderfilled]), which reconciles even on an unchanged ID list
+  /// because it can never recover otherwise.
   Future<void> _warmRevalidate(Emitter<ProfileLikedVideosState> emit) async {
     final List<String> freshIds;
     if (_isOtherUserProfile) {
@@ -263,7 +265,8 @@ class ProfileLikedVideosBloc
     }
     if (isClosed) return;
 
-    if (listEquals(freshIds, state.likedEventIds)) {
+    if (listEquals(freshIds, state.likedEventIds) &&
+        !_isWindowUnderfilled(freshIds)) {
       // Nothing changed — just hide the bar. No re-fetch, no re-serialize.
       emit(state.copyWith(isRefreshing: false));
       return;
@@ -297,7 +300,8 @@ class ProfileLikedVideosBloc
   /// Keeps videos still liked (in fresh order), drops unliked ones, and
   /// fetches only the liked IDs in the loaded window that have no video yet —
   /// typically just the clip the user (or another device) liked at the top.
-  /// Bounded to the loaded window so it never bulk-fetches.
+  /// Bounded to the loaded window (but never below one page) so it never
+  /// bulk-fetches.
   Future<({List<VideoEvent> videos, int nextPageOffset, bool hasMoreContent})>
   _reconcile(List<String> freshIds) async {
     final byId = {for (final video in state.videos) video.id: video};
@@ -307,11 +311,15 @@ class ProfileLikedVideosBloc
     // previous total), which would otherwise balloon the window to a full
     // re-fetch on reopen. Falls back to 0 (no growth) if the old top is gone.
     final newAtTop = _newItemsAtTop(freshIds);
-    final windowSize =
-        (max(state.nextPageOffset, state.videos.length) + newAtTop).clamp(
-          0,
-          freshIds.length,
-        );
+    // Floored at one page. The window is otherwise bounded by what is already
+    // loaded, so a first load that resolved fewer than a page — e.g. the cold
+    // path ran against a local liked-ID list that had not finished syncing —
+    // stays that size permanently: every reconcile re-derives the same short
+    // window and persists it, and a grid shorter than the viewport never
+    // scrolls, so scroll-driven pagination can never grow it either.
+    final loadedWindow =
+        max(state.nextPageOffset, state.videos.length) + newAtTop;
+    final windowSize = max(loadedWindow, _pageSize).clamp(0, freshIds.length);
     final windowIds = freshIds.take(windowSize).toList();
 
     final missingIds = windowIds.where((id) => !byId.containsKey(id)).toList();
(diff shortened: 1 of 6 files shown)


mobile/lib/blocs/profile_liked_videos/profile_liked_videos_bloc.dart
```diff
@@ -307,11 +311,15 @@ class ProfileLikedVideosBloc
...
-        );
+    // Floored at one page. The window is otherwise bounded by what is already
+    // loaded, so a first load that resolved fewer than a page — e.g. the cold
+    // path ran against a local liked-ID list that had not finished syncing —
+    // stays that size permanently: every reconcile re-derives the same short
+    // window and persists it, and a grid shorter than the viewport never
+    // scrolls, so scroll-driven pagination can never grow it either.
+    final loadedWindow =
+        max(state.nextPageOffset, state.videos.length) + newAtTop;
+    final windowSize = max(loadedWindow, _pageSize).clamp(0, freshIds.length);
```
realmeylisdev, line 322: note on the Out of Scope framing rather than the code: the residual needs a lot fewer dead IDs than the writeup suggests.

the grid is 3 columns of square tiles (no `childAspectRatio`) with 4px spacing, so R rows is `R*(W-8)/3 + (R-1)*4`. on a 390x844 device the body is 844 less the 72px `_kTabSlotHeight` less the 34px bottom inset, so 738. five rows is 653 and six is 784, so nothing scrolls until 6 rows are on screen, i.e. **16 of the 18 window IDs have to resolve**. three dead IDs is enough to leave the tab stuck, not the ~90% of a window the description implies.

doesn't change my read that this is strictly better than today, or that the real cure is the mixin. just worth knowing the margin is that thin when you scope that one.

mobile/lib/blocs/profile_saved_videos/profile_saved_videos_bloc.dart
```diff
@@ -221,11 +222,11 @@ class ProfileSavedVideosBloc
...
-    final windowSize =
-        (max(state.nextPageOffset, state.videos.length) + newAtTop).clamp(
-          0,
-          freshIds.length,
-        );
+    // Floored at one page so a window that was once shorter than a page can
+    // recover; see ProfileLikedVideosBloc for the lock-in it prevents.
+    final loadedWindow =
+        max(state.nextPageOffset, state.videos.length) + newAtTop;
+    final windowSize = max(loadedWindow, _pageSize).clamp(0, freshIds.length);
```
realmeylisdev, line 229: nit, description only: the Out of Scope note says `_coldLoad` fills a page of *videos* via `_fetchVideoPage` while `_reconcile` doesn't.

that holds for Liked, but Saved and Reposts cold-load with a single `take(_pageSize)` plus one `_fetchVideos` (this file :183-184, reposts :183-184), so they never fill a page of videos either. no effect on the fix, it just means the residual you're deferring sits one step earlier on these two tabs than the writeup reads.

realmeylisdev (approved):
lgtm. checked the convergence claim and it holds: after any reconcile `nextPageOffset` is at least `min(18, ids.length)`, so the bypass can only ever fire once.

pushed `ffcf27114` with three tests pinning that the predicate counts consumed IDs. before it, keying it off `videos.length` or dropping the `< ids.length` half both kept all 80 tests green. revert freely.

two non-blocking notes inline.

PR #6623, opened 2026-08-03
fix(video): purge NIP-09 deleted videos from the local event store

### LINKED ISSUE #6625: fix(video): purge NIP-09 deleted videos from the local event store
Auto-created from PR #6623.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6623
**Branch:** `fix/profile-videos-deleted-flash`

hm21:
Closes #6625

## Description

Restarting the app and opening your own profile flashes already-deleted videos in the Videos tab for a few hundred milliseconds, on every launch.

**Root cause.** Applying a NIP-09 deletion only clears the in-memory feed caches — `_eventLists`, `_authorBuckets`, `_hashtagBuckets`. The comment on the local delete path says so outright: *"removal is permanent for the lifetime of this service."* The event rows in the local store survive it, and the cache-first startup query (`_loadCachedEvents`) reads straight from those rows, so every launch puts the deleted clips back into the author buckets. They only disappear once the relay redelivers the Kind 5 event and the handler re-applies the deletion in memory — which is the flash.

Captured from a device trace, restart → own profile:

```
12:02:42.248  Cache-first: UI updated with 3 cached events for instant display
12:02:42.557  SVC authorVideos: cached=3          <- grid renders the deleted clips
12:02:42.929  Applied observed video deletion c03f3e17… to 2 cached location(s)
12:02:42.930  Applied observed video deletion 63cbc380… to 2 cached location(s)
12:02:42.930  Applied observed video deletion 293201d8… to 2 cached location(s)
12:02:43.180  SVC authorVideos: cached=0          <- gone, ~680ms later
12:03:00.256  SVC authorVideos: cached=3          <- the real live videos
```

The three ids are that account's own Kind 5 events from June (the log line prints the deletion event id, not the target). Relay and backend were both clean at the time — 3 live videos, 63 tombstoned targets, none of them still fetchable — so the stale state was purely local.

The existing tombstone filter cannot cover this window: `_knownDeletedVideoIds` is seeded from `content_deletions_history` in SharedPreferences, which is empty on a cold start for anything this install did not delete itself.

### Purge the rows

Both removal paths now delete the rows as well, using `NostrEventsDao.deleteEventsByIds` — documented for exactly this (*"Used when processing NIP-09 deletion events (Kind 5)"*) and until now only reachable via `db_video_local_storage`.

The purge is fire-and-forget with a swallowed failure: the in-memory removal has already succeeded at that point, so a failed purge must not break deletion handling.

### Stop the cache-first read writing them back

Purging alone did not hold, and device testing caught it: the logs reported `Purged 1 deleted event(s) from the local cache` on every launch and the flash kept coming back. The cache-first read hands every row it loads to `_handleNewVideoEvent`, the same handler the relay path uses, which re-queues it for persistence at **background** priority — the queue `EventRouter` drains *last*. Under the startup relay burst (batches of 26/34/36/50 events in the trace above) that write-back lands *after* the deletion has purged the same rows. Every launch re-created exactly what it had just deleted, so no number of restarts converged.

So: skip the write-back for events that came out of the store — they are already persisted, and re-queueing them was pure write amplification on top of the correctness bug — and never re-persist an event that is already tombstoned, by id or by `pubkey:d-tag` coordinate (an addressable edit mints a new id for the same clip).

### Close the paths that bypassed that guard

Review found the guard could not actually do its job, and the review's own probe tests confirmed each gap on a real Drift database.

**The feed listener wrote the row back before the guard ran.** `subscribeToVideoFeed` queued every relay event unguarded at normal priority immediately before handing it to `_handleNewVideoEvent`, which queued it again at background priority. Since normal drains first, the guard only ever suppressed a redundant second write. Ingestion now goes through `_handleNewVideoEvent` alone, which takes the priority the caller wants, so the service has one persistence entry point that the guard genuinely covers — and one write per event instead of two.

**A write queued a moment before the deletion still landed after it.** The video filter and the Kind 5 filter share one REQ, so both arrive in the same burst. The video was queued while no tombstone existed, so it passed the guard; the purge then issued a DELETE that matched nothing, and the flush inserted the row afterwards. The purge now evicts pending router writes (`EventRouter.dropQueuedEvents`) before deleting.

**Rows no feed had loaded were never purged at all.** The purge only received ids validated against the in-memory caches, so a Kind 5 for a stored-but-unloaded video — the exact state that survives a restart — did nothing. It now resolves the deletion's requested ids and `a`-tag coordinates against the store, with the author re-checked per stored row so a forged `e` tag cannot reach another author's cache.

**Coordinate tombstones could not match.** They are keyed on `VideoEvent.stableId`, which falls back to a `vine_id` tag when there is no `d` tag, while the guard only scanned for `d`. Both now resolve the stable id the same way.

### Stop `NostrClient` re-caching deleted events

`NostrClient.subscribe` caches every non-deletion event straight to the DAO from its own callback, before the app layer sees the event. No guard in `VideoEventService` can reach that write, so any relay that redelivered a deleted video put the row straight back and the clip returned on the next launch.

`_handleDeletionEvent` now records what it removed — by event id and by `kind:pubkey:d-tag` coordinate, so an edit of the same clip is caught too — and the auto-cache skips those. The sets are bounded; losing an old entry only costs a redundant row that the next deletion purges again.

Installs that already hold stale rows flash **once** more on the next launch — that launch is what purges them — and are clean from the launch after.

**Related Issue:** 

## Out of Scope

- `NostrClient._deleteCachedEventsForDeletion` still deletes by `e` tag with no author validation, so a hostile relay can drop arbitrary rows from the local cache. Pre-existing and untouched. Both purge paths this PR adds do validate — the app-layer one against the stored row's pubkey, the client-side tombstone against the deletion author — so the gap is now limited to that one pre-existing delete.
- `content_deletions_history` was absent from the affected device's SharedPreferences despite 63 deletions on that account, so the tombstone seed had nothing to seed. This fix makes that irrelevant for the flash, but the persistence gap itself is untouched.
- Funnelcake's `/api/users/<pubkey>/videos` returned `[]` for an account with 3 live videos on the relay. Backend-side, not this PR.

## Verification

- Manually verified on a real Samsung Galaxy (SM-S942B). Reproducing needs two restarts: the first still flashes and purges, the second is clean. Re-verified on device after the review fixes.
- Review follow-up: NIP-09 `a` deletions are now scoped to the deletion request's own `created_at` on every path — the store purge, the in-memory coordinate tombstone, and the client-side auto-cache suppression — so a version republished after the request stays visible. Coordinate tombstones became a map to that timestamp, which changes `seedKnownDeletionTombstones` to take `addressableDeletedAt`. Client-side `e` tag tombstones are keyed on the deletion author, and rows the store purge resolves now seed the in-memory tombstone so a later relay delivery cannot write them back.
- 8 regression tests against a real Drift database, one per failure mode: both removal paths (observed Kind 5, locally initiated delete), the resurrection ordering, relay redelivery by id, redelivery of an addressable edit by coordinate, a deletion arriving in the same burst as its video, a stored row no feed has loaded, and a forged deletion from a non-author.
- Every fix is mutation-tested: reverting each one individually turns a specific test red. Removing the tombstone guard fails both redelivery tests, removing the queue eviction fails the same-burst test, removing the store purge fails the not-in-memory test, removing the author check fails the forged-deletion test, and removing the client-side suppression fails both `nostr_client` tests.
- 3 new `nostr_client` tests for the auto-cache suppression, including a negative control that a never-tombstoned event is still cached.
- `flutter test test/services/` → 3652 passed
- `flutter test` in `packages/nostr_client` → 300 passed, 86.79% line coverage (gate 82%)
- `flutter analyze lib test` → clean, both in the app and in `packages/nostr_client`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/providers/video_providers.dart
+++ b/mobile/lib/providers/video_providers.dart
@@ -176,9 +176,11 @@ VideoEventService videoEventService(Ref ref) {
   }
   service.seedKnownDeletionTombstones(
     eventIds: persistedDeletions.map((deletion) => deletion.originalEventId),
-    addressableIds: persistedDeletions
-        .map((deletion) => deletion.addressableId)
-        .whereType<String>(),
+    addressableDeletedAt: {
+      for (final deletion in persistedDeletions)
+        ?deletion.addressableId:
+            deletion.deletedAt.millisecondsSinceEpoch ~/ 1000,
+    },
   );
   service.setBlocklistRepository(blocklistRepository);
   service.setLikesRepository(likesRepository);
--- a/mobile/lib/services/video_event_service.dart
+++ b/mobile/lib/services/video_event_service.dart
@@ -231,7 +231,14 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
   // Track known deleted videos to prevent resurrection from pagination.
   // Includes local user deletions and observed remote NIP-09 deletes.
   final Set<String> _knownDeletedVideoIds = {};
-  final Set<String> _knownDeletedVideoCoordinateKeys = {};
+
+  /// `pubkey:stable-id` coordinate to the `created_at` of the deletion that
+  /// removed it.
+  ///
+  /// NIP-09 scopes an `a` deletion to versions up to the deletion request's
+  /// own timestamp, so a version the author republishes afterwards is not
+  /// covered and must stay visible.
+  final Map<String, int> _knownDeletedVideoCoordinates = {};
 
   // Broadcasts video ids the moment they are removed (deletion / future
   // block / mute). Subscribers — fullscreen feed bloc, profile feed
@@ -1103,7 +1110,12 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
     _knownDeletedVideoIds.add(primaryEventId);
     _knownDeletedVideoIds.addAll(removedVideoIds);
     if (coordinateKey != null) {
-      _knownDeletedVideoCoordinateKeys.add(coordinateKey);
+      // The NIP-09 request the caller just published carries "now", so every
+      // version up to this moment is covered while a later republish is not.
+      _rememberDeletedCoordinate(
+        coordinateKey,
+        DateTime.now().millisecondsSinceEpoch ~/ 1000,
+      );
     }
 
     _purgeEventsFromLocalCache({primaryEventId, ...removedVideoIds});
@@ -1157,28 +1169,30 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
   /// Hydrate known deletion tombstones from persisted delete history or
   /// observed remote deletes.
   ///
-  /// [addressableIds] must use the NIP-33 `kind:pubkey:d-tag` format. The
-  /// service keeps its own compact lookup key private so callers do not couple
-  /// to the in-memory implementation.
+  /// Keys of [addressableDeletedAt] must use the NIP-33 `kind:pubkey:d-tag`
+  /// format; each value is the `created_at` of the deletion request that
+  /// removed it, in seconds. The service keeps its own compact lookup key
+  /// private so callers do not couple to the in-memory implementation.
   void seedKnownDeletionTombstones({
     Iterable<String> eventIds = const [],
-    Iterable<String> addressableIds = const [],
+    Map<String, int> addressableDeletedAt = const {},
   }) {
     _knownDeletedVideoIds.addAll(eventIds.where((id) => id.isNotEmpty));
 
-    for (final addressableId in addressableIds) {
-      final parsed = AId.fromString(addressableId);
+    for (final entry in addressableDeletedAt.entries) {
+      final parsed = AId.fromString(entry.key);
       if (parsed == null ||
           parsed.kind != NIP71VideoKinds.addressableShortVideo ||
           parsed.pubkey.isEmpty ||
           parsed.dTag.isEmpty) {
         continue;
       }
-      _knownDeletedVideoCoordinateKeys.add(
+      _rememberDeletedCoordinate(
         _knownDeletionCoordinateKeyFromParts(
           pubkey: parsed.pubkey,
           stableId: parsed.dTag,
         ),
+        entry.value,
       );
     }
   }
@@ -1186,20 +1200,36 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
   /// Backwards-compatible alias for local delete-history hydration.
   void seedLocalDeletionTombstones({
     Iterable<String> eventIds = const [],
-    Iterable<String> addressableIds = const [],
+    Map<String, int> addressableDeletedAt = const {},
   }) {
     seedKnownDeletionTombstones(
       eventIds: eventIds,
-      addressableIds: addressableIds,
+      addressableDeletedAt: addressableDeletedAt,
     );
+  }
+
+  /// Records a coordinate tombstone, keeping the newest deletion when the
+  /// same coordinate is deleted more than once.
+  void _rememberDeletedCoordinate(String coordinateKey, int deletedAt) {
+    final known = _knownDeletedVideoCoordinates[coordinateKey];
+    if (known != null && known >= deletedAt) return;
+    _knownDeletedVideoCoordinates[coordinateKey] = deletedAt;
+  }
+
+  /// Whether [coordinateKey] was deleted by a request at or after
+  /// [createdAt] — a version published later is not covered by it.
+  bool _isCoordinateDeleted(String coordinateKey, int createdAt) {
+    final deletedUpTo = _knownDeletedVideoCoordinates[coordinateKey];
+    return deletedUpTo != null && createdAt <= deletedUpTo;
   }
 
   /// Check whether this concrete video or its addressable identity has been
   /// deleted by a known local or remote NIP-09 delete.
   bool isVideoEventKnownDeleted(VideoEvent video) {
     return _knownDeletedVideoIds.contains(video.id) ||
-        _knownDeletedVideoCoordinateKeys.contains(
+        _isCoordinateDeleted(
           _knownDeletionCoordinateKey(video),
+          video.createdAt,
         );
   }
 
@@ -1230,11 +1260,12 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
     final stableId = _rawStableId(event);
     if (stableId == null) return false;
 
-    return _knownDeletedVideoCoordinateKeys.contains(
+    return _isCoordinateDeleted(
       _knownDeletionCoordinateKeyFromParts(
         pubkey: event.pubkey,
         stableId: stableId,
       ),
+      event.createdAt,
     );
   }
 
@@ -1295,9 +1326,9 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
       final matchesRequestedId = requestedEventIds.contains(
         video.id.toLowerCase(),
       );
-      final matchesCoordinate = coordinateKeys.contains(
-        _knownDeletionCoordinateKey(video),
-      );
+      final matchesCoordinate =
+          coordinateKeys.contains(_knownDeletionCoordinateKey(video)) &&
+          video.createdAt <= deletionEvent.createdAt;
 
       if (matchesRequestedId || matchesCoordinate) {
         validatedEventIds.add(video.id);
@@ -1310,6 +1341,7 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
     // against the stored row instead of against memory.
     _purgeDeletedVideosFromStore(
       deletionPubkey: deletionPubkey,
+      deletionCreatedAt: deletionEvent.createdAt,
       requestedEventIds: requestedEventIds,
       addressableIds: addressableIds,
     );
@@ -1318,12 +1350,16 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
 
     seedKnownDeletionTombstones(
       eventIds: validatedEventIds,
-      addressableIds: addressableIds,
+      addressableDeletedAt: {
+        for (final addressableId in addressableIds)
+          addressableId: deletionEvent.createdAt,
+      },
     );
 
     _removeKnownDeletedVideos(
       eventIds: validatedEventIds,
       coordinateKeys: coordinateKeys,
+      deletionCreatedAt: deletionEvent.createdAt,
       logIdentity: deletionEvent.id,
     );
   }
@@ -1343,14 +1379,18 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
   void _removeKnownDeletedVideos({
     required Set<String> eventIds,
     required Set<String> coordinateKeys,
+    required int deletionCreatedAt,
     required String logIdentity,
   }) {
     var removedCount = 0;
     final removedVideoIds = <String>{};
 
+    // A coordinate match only counts up to the deletion's own timestamp; a
+    // version the author republished afterwards is not covered by it.
     bool shouldRemove(VideoEvent video) =>
         eventIds.contains(video.id) ||
-        coordinateKeys.contains(_knownDeletionCoordinateKey(video));
+        (coordinateKeys.contains(_knownDeletionCoordinateKey(video)) &&
+            video.createdAt <= deletionCreatedAt);
 
     for (final entry in _eventLists.entries) {
       final initialLength = entry.value.length;
@@ -1445,6 +1485,12 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
     );
   }
 
+  /// Upper bound on rows a single `a`-tag coordinate can contribute to one
+  /// purge. Addressable writes upsert per coordinate, so a handful is the
+  /// realistic ceiling; this only stops the DAO's implicit default from
+  /// silently truncating a pathological set.
+  static const int _maxCoordinateRowsPerPurge = 500;
+
   /// Purges the rows an observed NIP-09 deletion targets, including rows no
   /// feed has loaded — the ones that survive a restart and flash on the next
   /// launch.
@@ -1454,6 +1500,7 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
   /// Addressable `a` tags were already author-matched by the caller.
   void _purgeDeletedVideosFromStore({
     required String deletionPubkey,
+    required int deletionCreatedAt,
     required Set<String> requestedEventIds,
     required Set<String> addressableIds,
   }) {
@@ -1465,6 +1512,7 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
       _resolveAndPurgeDeletedRows(
         router: router,
         deletionPubkey: deletionPubkey,
+        deletionCreatedAt: deletionCreatedAt,
         requestedEventIds: requestedEventIds,
         addressableIds: addressableIds,
       ).catchError((Object error) {
@@ -1480,6 +1528,7 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
   Future<void> _resolveAndPurgeDeletedRows({
     required EventRouter router,
     required String deletionPubkey,
+    required int deletionCreatedAt,
     required Set<String> requestedEventIds,
     required Set<String> addressableIds,
   }) async {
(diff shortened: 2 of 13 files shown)


mobile/lib/services/video_event_service.dart
```diff
@@ -1354,6 +1400,133 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
...
+
+    for (final addressableId in addressableIds) {
+      final parsed = AId.fromString(addressableId);
+      if (parsed == null) continue;
+      final matches = await dao.getEventsByFilter(
+        Filter(
+          kinds: [parsed.kind],
+          authors: [parsed.pubkey],
+          d: [parsed.dTag],
+        ),
```
realmeylisdev, line 1581: should-fix.

this query has no upper time bound, so it deletes every stored row under the coordinate, including ones newer than the deletion request. NIP-09: *"When an `a` tag is used, relays SHOULD delete all versions of the replaceable event up to the `created_at` timestamp of the deletion request event."* `NostrClient._deleteCachedEventsForDeletion` already gets this right with `until: deletionCreatedAt`, so the two purge paths now disagree.

reproduced on a real drift db: store a 34236 row at `created_at: 5000`, then deliver an `a` deletion at `created_at: 2000` (relays redeliver old kind 5s on every launch, as your own trace shows). the newer row is purged. the same test passes on origin/main, so this is new here.

`_handleObservedDeletionEvent` still has `deletionEvent` in scope, so threading `deletionEvent.createdAt` down and adding `until:` here covers it.

minor, same call: the `e` branch above sets `limit: requestedEventIds.length`, this one sets none, so it falls back to the dao's default of 100.

mobile/packages/nostr_client/lib/src/nostr_client.dart
```diff
@@ -302,6 +316,72 @@ class NostrClient {
...
+  void _rememberTombstones({
+    required List<String> eventIds,
+    required List<AId> addressableIds,
+    required String deletionPubkey,
+  }) {
+    for (final eventId in eventIds) {
+      _tombstonedEventIds
+        ..remove(eventId)
+        ..add(eventId);
+    }
```
realmeylisdev, line 348: should-fix.

`e` tag ids go in unvalidated, while the `a` loop directly below correctly skips coordinates whose pubkey isn't the deletion author. NIP-09 is explicit about this one: *"A client MUST validate that each event `pubkey` referenced in the `e` tag of the deletion request is identical to the deletion request `pubkey`, before hiding or deleting any event. Relays can not, in general, perform this validation and should not be treated as authoritative."*

you scoped the unvalidated delete out as pre-existing, which is fair, but this change makes it stick. before, the row came back the next time any relay served the event. now `_shouldSkipAutoCache` suppresses it for the rest of the session.

reachability: `relay_pool.dart` never calls `Filter.checkEvent` on the inbound path, so REQ filters aren't enforced client-side and any connected relay can land one of these on any open subscription.

verified: kind 5 signed by `ffff…` naming `aaaa…` in an `e` tag, then `aaaa…` is delivered on the same subscription. `upsertEvent` is never called on this branch; it is called on origin/main.

keying the set on `'<deletionPubkey>:<eventId>'` and testing `'<event.pubkey>:<event.id>'` closes it without needing a db lookup.

mobile/packages/nostr_client/lib/src/nostr_client.dart
```diff
@@ -302,6 +316,72 @@ class NostrClient {
...
+
+    for (final tag in event.tags) {
+      if (tag.length < 2 || tag[0] != 'd' || tag[1].isEmpty) continue;
+      return _tombstonedCoordinates.contains(
+        _tombstoneCoordinateKey(
+          kind: event.kind,
+          pubkey: event.pubkey,
+          dTag: tag[1],
+        ),
+      );
```
realmeylisdev: should-fix.

the coordinate tombstone carries no timestamp, so once a `pubkey:d-tag` is deleted, every future event under it is skipped, including ones published after the deletion. `_deleteCachedEventsForDeletion` a few lines down already bounds exactly this with `until: deletionCreatedAt`, and the same NIP-09 sentence applies: *"up to the `created_at` timestamp of the deletion request event."*

verified: `a` deletion at `created_at: 1000`, then the author republishes the same coordinate at `created_at: 5000`. `upsertEvent` is never called on this branch; it is on origin/main.

making `_tombstonedCoordinates` a `Map<String, int>` of coordinate to deletion `created_at`, and suppressing only when `event.createdAt <= recorded`, keeps the lru eviction working (map literals stay insertion-ordered) and fixes it.

mobile/lib/services/video_event_service.dart
```diff
@@ -1270,6 +1304,16 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
...
 
+    // The store is purged from the *requested* ids, not the validated ones:
+    // a row that no feed has loaded is exactly the case that survives a
+    // restart and flashes on the next launch. Authorship is re-checked
+    // against the stored row instead of against memory.
+    _purgeDeletedVideosFromStore(
+      deletionPubkey: deletionPubkey,
+      requestedEventIds: requestedEventIds,
+      addressableIds: addressableIds,
+    );
```
realmeylisdev, line 1348: not blocking, and it fails on origin/main too, so this PR doesn't cause it. flagging because it caps what this purge buys you.

`if (validatedEventIds.isEmpty && coordinateKeys.isEmpty) return` two lines below fires for exactly the not-in-memory case, so `seedKnownDeletionTombstones` never runs for it. `_isKnownDeletedRawEvent` stays false, and the next time a relay serves that clip the row is written straight back and the video re-enters the feed, which puts the flash back on the launch after.

confirmed both halves against a real drift db. seeding from the ids `_resolveAndPurgeDeletedRows` already resolved would make the purge durable.

realmeylisdev (dismissed):
flash fix itself looks right, but three things in the new tombstone code came back red for me. reproduced each on a real drift db, both directions: the probe passes on origin/main and fails on `9cc8c1c`.

- the a-tag store purge dropped the `until: deletionCreatedAt` bound that `_deleteCachedEventsForDeletion` already has, so it deletes rows newer than the deletion request
- same missing timestamp on the new nostr_client coordinate tombstone, so a republished coordinate never gets cached again
- `_rememberTombstones` author-checks `a` tags but not `e` tags, and NIP-09 makes that a MUST for clients

details inline. can push the three fixes myself if that's easier.

mobile/lib/services/video_event_service.dart
hm21: Confirmed and fixed in d8c9a64. Threaded `deletionEvent.createdAt` down and added `until:`, plus the explicit `limit` you flagged.

The in-memory coordinate tombstone had the same unbounded shape, so fixing only this query left the outcome depending on whether the row happened to be loaded in a feed — your repro would still have failed. `_knownDeletedVideoCoordinateKeys` is now a map to the deletion `created_at` and every read applies the bound.

One note on the quote: that NIP-09 sentence is scoped to relays. The reason it's still right here is that `_deleteCachedEventsForDeletion` already bounds it, so the two paths disagreed.

mobile/packages/nostr_client/lib/src/nostr_client.dart
hm21: Confirmed and fixed in f297019. Keyed on `<deletionPubkey>:<eventId>` and tested against `<event.pubkey>:<event.id>` as you suggested — no db lookup.

Your point that suppression makes it stick is the one that changed my mind; the delete itself was pre-existing but the session-long suppression was new. `_deleteCachedEventsForDeletion` still deletes `e` tags unvalidated and stays out of scope, now noted as such in the description.

mobile/packages/nostr_client/lib/src/nostr_client.dart
hm21: Confirmed and fixed in f297019. `_tombstonedCoordinates` is a `Map<String, int>` and `_shouldSkipAutoCache` suppresses only when `event.createdAt <= recorded`. Kept insertion-ordered for the eviction, and it keeps the newer timestamp when a coordinate is tombstoned twice.

mobile/lib/services/video_event_service.dart
hm21: Fixed in d8c9a64 rather than left as-is — `seedKnownDeletionTombstones` now runs on the ids the store purge resolved, so the guard covers them and a later delivery can't write them back.

You're right it predates this PR, but it capped exactly what the new purge was for.

hm21:
@realmeylisdev all three blockers fixed, plus the non-blocking one — d8c9a64 (app) and f297019 (client).

The `until` bound needed to go wider than the store purge: the in-memory coordinate tombstone was unbounded too, so your repro would still have failed whenever the row was loaded in a feed. Coordinate tombstones are now a map to the deletion `created_at` and every read applies the bound, which changes `seedKnownDeletionTombstones` to take `addressableDeletedAt`.

Six new tests, each mutation-tested — reverting any single bound turns a specific one red. Re-verified on device.

`_deleteCachedEventsForDeletion` deleting `e` tags unvalidated stays out of scope and is called out in the description.

mobile/lib/services/video_event_service.dart
```diff
@@ -1155,49 +1169,67 @@ class VideoEventService extends ChangeNotifier implements VideoEventCache {
...
+  }
+
   /// Check whether this concrete video or its addressable identity has been
   /// deleted by a known local or remote NIP-09 delete.
   bool isVideoEventKnownDeleted(VideoEvent video) {
     return _knownDeletedVideoIds.contains(video.id) ||
-        _knownDeletedVideoCoordinateKeys.contains(
+        _isCoordinateDeleted(
           _knownDeletionCoordinateKey(video),
+          video.createdAt,
```
realmeylisdev: The bound itself is right, but `VideoEvent.createdAt` isn't the event's `created_at`. `video_event.dart:644` sets it to `published_at ?? created_at`, and this app puts a `published_at` tag on every video it publishes (`video_event_publisher.dart:1307`). The edit path keeps the same `d` tag (`video_metadata_update_service.dart:240`) and re-emits the *original* `published_at` (`:316`), so for an edited clip `VideoEvent.createdAt` is the first publication time rather than the republish time.

Publish at 1000, delete at 2000, edit-republish at 3000 with `published_at` still 1000. A relay redelivers the old Kind 5 on launch. The store purge gets it right: `until: 2000` against the row's real `created_at` of 3000, so the row survives. In memory `1000 <= 2000` is true, so `_removeKnownDeletedVideos` strips it from every feed and `isVideoEventKnownDeleted` keeps dropping it at `:4866` and `profile_feed_cubit.dart:723`. The clip ends up sitting on disk and permanently invisible, which is the outcome the `until` bound was added to prevent.

`_isKnownDeletedRawEvent` at `:1268` already uses the raw `event.createdAt`, so the two guards in this class disagree about the same event now.

`VideoEvent` has no field carrying the raw `created_at` (both `createdAt` and `timestamp` derive from `effectiveTimestamp`), so this needs either a new field used at `:1232`, `:1331` and `:1393`, or the bound resolved against the raw event instead. Your call which. Both new fixtures omit `published_at`, so the suite can't see it either way.

mobile/lib/providers/video_providers.dart
```diff
@@ -176,9 +176,11 @@ VideoEventService videoEventService(Ref ref) {
   }
   service.seedKnownDeletionTombstones(
     eventIds: persistedDeletions.map((deletion) => deletion.originalEventId),
-    addressableIds: persistedDeletions
-        .map((deletion) => deletion.addressableId)
-        .whereType<String>(),
+    addressableDeletedAt: {
+      for (final deletion in persistedDeletions)
+        ?deletion.addressableId:
+            deletion.deletedAt.millisecondsSinceEpoch ~/ 1000,
```
realmeylisdev, line 182: I mutation-tested the new bounds. The headline ones are solidly pinned: deleting `until: deletionCreatedAt` reddens "keeps a coordinate version published after the deletion request", dropping `event.createdAt <= deletedUpTo` reddens "caches a coordinate republished after the deletion request", and reverting `_tombstoneEventKey` to bare-id keying reddens exactly the forged-`e` test. Both author checks and both in-memory coordinate bounds redden independently too.

Some of the new logic has no coverage though. Zeroing this timestamp leaves 28 nearby tests green, and it's the only production caller of the changed signature, so a wrong value here makes every persisted addressable tombstone cover nothing at cold start. Also green: flipping all three inclusive `<=` boundaries (the DAO `until`, `_isCoordinateDeleted`, `_shouldSkipAutoCache`) to `<`, and deleting either copy of the newest-wins guard (`_rememberDeletedCoordinate`, and the max-keeping ternary in `_rememberTombstones`). No fixture sits a version at exactly the deletion's `created_at`, and none delivers two deletions for one coordinate.

realmeylisdev (commented):
three blockers check out. i mutation-tested each of them and the reverts go red on exactly the tests you'd expect, with the forged-`e` one isolating cleanly (legit-hit test stays green).

one new thing on the in-memory bound though: `VideoEvent.createdAt` is `published_at`, not the event's own `created_at`, so edited clips still get suppressed. left it inline along with the coverage gaps i hit.

mobile/lib/services/video_event_service.dart
hm21: Verified and fixed in 82d989c + cf8aecff. You're right on all three links: `video_event.dart:644` derives `createdAt` from `published_at`, the publisher stamps it on every video, and the edit path re-emits the original while keeping the `d` tag.

Went with the new field. `VideoEvent.eventCreatedAt` carries the raw `created_at` from `fromNostrEvent`, and `nostrCreatedAt` falls back to `createdAt` when it's absent — optional rather than required because there are ~900 construction sites. All three comparisons use it, so they agree with the raw-event guard now.

Your scenario is pinned as a test: publish 1000 / delete 2000 / edit-republish 3000 with `published_at` still 1000, asserting the clip stays in the feed and on disk. Reverting the three sites to `createdAt` reddens it.

mobile/lib/providers/video_providers.dart
hm21: Both covered in cf8aecff.

The provider timestamp now has a test that seeds real deletion history through SharedPreferences and reads the provider — zeroing it and dropping the `~/ 1000` each go red. The inclusive boundary is pinned by a version sitting exactly on the deletion `created_at`, on the `_isCoordinateDeleted` path.

Left the two newest-wins guards uncovered on purpose: reaching either needs delete → republish → delete-again on one coordinate, and the failure mode is a slightly narrower bound rather than anything user-visible. Happy to add it if you'd rather have it pinned.

hm21:
@realmeylisdev the `published_at` catch was a good one — verified end to end and fixed in 82d989c + cf8aecff.

`VideoEvent` now carries the raw event `created_at` as `eventCreatedAt`, with a `nostrCreatedAt` getter that falls back to `createdAt`; all three coordinate comparisons use it, so they no longer disagree with the raw-event guard. Your publish-1000 / delete-2000 / edit-republish-3000 case is a test now.

Both coverage gaps closed too — the provider's `deletedAt` conversion and the inclusive boundary. The two newest-wins guards are still uncovered by choice, reasoning in the thread.

models 836, test/services 3632, providers + profile_feed 637, analyze clean.

realmeylisdev (approved):
took this one over and pushed four commits on top of yours.

we hit the `created_at` fix independently and yours landed first, so i reset onto it and dropped mine. your `ProviderContainer` hydration test is better than the helper i'd extracted for the same gap, so that went too.

what's left from my side: `1b014ca` binds the a-tag purge author lowercased, because the tag's own casing went straight into a case-sensitive `pubkey IN (...)` and matched no row, so an uppercase-hex coordinate left the clip on disk to flash on the next launch. `0d842d5` sweeps memory once the async store purge resolves; for an e-tag deletion whose target is in the store but in no feed, `_handleObservedDeletionEvent` returns at its empty guard, so a copy arriving inside that window cleared the ingestion guard before the tombstone landed and sat on the grid for the rest of the session. `27efe7f` pins the store-side inclusive edge plus newest-deletion-wins in both layers, all of which survived being reverted with the suite green. `1af18d4` is an 80-col lint i introduced in the package.

every new test is mutation-checked: reverting its bound turns that test, and only that test, red. full app suite 14256 green, nostr_client 305, models 836, analyze clean in the app and in both packages. checks are 15/15 (license/cla has never reported on this PR).

two things i looked at and left alone: the `deletedAt` bound is the local confirmation time rather than the Kind 5's own `created_at`, so it over-suppresses by the publish round trip, and `seedLocalDeletionTombstones` is still documented as a backwards-compatible alias even though its signature changed. neither is worth a commit here.

approving. they're your commits to keep, tweak or revert.

PR #6630, opened 2026-08-03
feat(profile): rebuild the edit-profile screen on the new design

### LINKED ISSUE #6631: feat(profile): rebuild the edit-profile screen on the new design
Track the edit-profile rebuild shipped by PR #6630.

Scope:

- Rebuild `/edit-profile` around the Figma **6. Profile Management** flow.
- Preserve existing website, verified-account, NIP-05, and public-key behavior while moving the visual structure to the new filled-card/header design.
- Fix image clear semantics so removing a saved avatar or banner publishes the removal instead of falling back to the persisted URL/color.
- Seed the form from cached profile data immediately, then refresh only untouched fields from fresher relay snapshots.
- Keep the public Divine handle display as `@name` while preserving the underlying `_@name.divine.video` NIP-05 registry format.

PR: https://github.com/divinevideo/divine-mobile/pull/6630
Branch: `feat/edit-profile-redesign`

hm21:
Closes #6631

## Description

Rebuilds the edit-profile screen on the Figma flow under **6. Profile Management**, and fixes three defects that surfaced while wiring it up.

The screen had been grown feature by feature — website field, get-verified tile, banner upload replacing the old colour picker — each in its own PR and none against a design frame. The order on screen was the order things were added. Figma now draws the whole flow ([`changeBanner-editProfile`](https://www.figma.com/design/rp1DsDEUuCaicW0lk6I2aZ/UI-Design?node-id=11251-229368)), so this moves the screen onto it.

**Layout.** A full-bleed header preview (banner behind an overlapping avatar, a pencil on each) above a form of filled cards: display name, username, NIP-05, bio, public key. Website and verified accounts are absent from the frames but still ship, so they follow underneath in the same style rather than being dropped. Image sources moved into bottom sheets; the banner colour picker gained the palette, a "no colour" swatch and a custom-colour escape hatch.

Every Figma variable mapped onto an existing `VineTheme` token — including the six `divine/accents/*` colours the banner palette uses — so no new colours were introduced.

| Before | After |
| --- | ---- |
|   Bildschirmfoto 2026-08-03 um 18 02 20  |  Bildschirmfoto 2026-08-03 um 17 59 47    |

### Fixes found along the way

**Clear never worked on a saved image.** `effectiveBanner` and `effectivePictureUrl` fell back to the persisted value, so Save republished exactly what the user had asked to remove — the preview was not the only thing lying. Both now carry a `cleared` flag that separates "no staged change" from "the absence *is* the change", reset as soon as something new is staged. The avatar had no clear affordance at all; it has one now.

**The form waited for the relay.** `MyProfileBloc` emits the cached profile on `MyProfileLoading` and only reaches `MyProfileLoaded` after `fetchFreshProfile` returns, but the listener ignored the former — so the form sat blank for a whole round trip while the profile screen had already drawn the same avatar from the same cache. It now seeds from the first snapshot carrying a profile.

That exposed a second, worse problem: `MyProfileLoaded` is emitted **again** when the verifier claims resolve, and the listener re-seeded the controllers each time — silently overwriting anything typed in between. Seeding now only refreshes fields the user has not touched, and only replaces a staged picture or banner when nothing of the user's own is pending.

**The keyboard's next key skipped a field.** `TextInputAction.next` already advances focus, so the extra `onSubmitted` handler advanced a second time and landed on the NIP-05 card — focusable, but with no keyboard, so pressing next just dismissed it. From the username field, next now targets the bio directly rather than walking traversal order past that card.

### Notes on two judgment calls

- **Bio grows 360 → 1000.** The limit lives only in this widget; neither `ProfileEditorBloc`, the profile repository, nor `UserProfile` constrains kind-0 `about`.
- **Username renders as `@name`.** The controller still holds the bare handle and the registry still issues the underlying `_@<handle>.divine.video` NIP-05 subdomain form, but the edit field hides the domain because users expect Divine handles to read as simple at-names.

**Related Issue:** 

## Out of Scope

- `changeDecoration-*` (avatar decoration, Supporters — #6378 unshipped).
- The `14732:*` "Add stickers" frames, which redraw the banner sheet as *Change background photo / Change background color / Manage Stickers*. Higher node ids than the shipped flow, but they sit under their own signage as a future feature, not a replacement.
- `changeBanner-previewImage` / `-previewColor` and `changeAvatar-preview` as standalone screens.

## Verification

- `dart format` — clean
- `mobile/packages/divine_ui`: 825 tests, **100.00 % line coverage** (strict gate held)
- `test/screens/`, `test/widgets/`, `test/blocs/`, `test/l10n/` — green
- All four design-system ratchets pass; three baselines regenerated because the migration *reduced* raw `TextStyle`, raw colour and Material-button counts
- Untested-services floor and `test/unit` structure ratchets pass
- Seventeen new l10n keys mirrored into all 21 non-English locales; eight keys the rebuild orphaned were removed from every locale, and the duplicate `profileSetupPublicKeyLabel` this branch briefly introduced was folded back into the existing entry
- Each of the original three fixes has a regression test that was confirmed red against the old code before the fix landed
- Takeover follow-up: `flutter test test/screens/profile_setup_screen_test.dart test/screens/profile_setup/username_field_test.dart test/screens/profile_setup/display_name_field_test.dart`, `flutter test test/l10n/hardcoded_visible_strings_test.dart`, and `flutter analyze lib test integration_test` — clean

Verified on a real Samsung Galaxy (SM S942B, Android 16): form seeding, avatar and banner clear, the banner link and colour sheets, keyboard next, and the npub copy row.

## Type of Change

- [x] 🧹 Code refactor


mobile/lib/screens/profile_setup/widgets/profile_setup_listeners.dart
```diff
@@ -42,45 +43,150 @@ class ProfileSetupListeners extends ConsumerWidget {
...
+          !editorState.bannerCleared) {
+        editorBloc.add(InitialPersistedBannerSet(profile.banner));
+      }
+    }
+
+    if (!_hasSeededUsername && extractedUsername != null) {
+      _hasSeededUsername = true;
+      editorBloc.add(InitialUsernameSet(extractedUsername));
+    }
+    if (!_hasSeededExternalNip05 && externalNip05 != null) {
```
realmeylisdev: **blocking.** `_hasSeededUsername` and `_hasSeededExternalNip05` are independent one-shots, and only this branch ever touches the mode. So a cached external NIP-05 latches `Nip05Mode.external_` and nothing switches it back when the fresher snapshot turns out to carry a divine handle.

Repro: the relay's kind 0 says `nip05: _@bob.divine.video` (set on another device), this device's cache still holds an older one with `alice@example.com`. Open Edit Profile, tap Save, touch nothing.

* `MyProfileLoading(cached)` fires `InitialExternalNip05Set` + `Nip05ModeChanged(external_)` + `ExternalNip05Changed`
* `MyProfileLoaded(fresh)` fires `InitialUsernameSet('bob')` and nothing else

I drove exactly that through `ProfileSetupScreenView`; the events reaching the editor are

```
[InitialPersistedPictureSet, InitialPersistedBannerSet,
 InitialExternalNip05Set, Nip05ModeChanged -> external_,
 ExternalNip05Changed -> alice@example.com,
 InitialPersistedPictureSet, InitialPersistedBannerSet,
 InitialUsernameSet -> bob]
```

no `Nip05ModeChanged(divine)` anywhere. `_saveProfile` then takes `isExternal`, so `username` goes null and `nip05` goes `alice@example.com`, and `saveProfileEvent` writes that straight over `_@bob.divine.video` (`profile_repository.dart:1008` and `:1053`). The #4012 guard can't catch it either, `clearNip05` requires `!isExternal`.

On main the listener was gated `curr is MyProfileLoaded`, so the cached snapshot never reached it, the mode stayed `divine`, and Save published `bob`. Cache-first seeding is what exposes this.

Left for you rather than pushed: which snapshot should win when the two disagree is a product call, and it lands right on top of the #4012 guard.

mobile/lib/screens/profile_setup/widgets/profile_setup_listeners.dart
```diff
@@ -42,45 +43,150 @@ class ProfileSetupListeners extends ConsumerWidget {
...
+      // the user has staged nothing of their own — InitialPersistedBannerSet
+      // also rewrites the pending colour, which would undo a swatch they just
+      // picked.
+      final editorState = editorBloc.state;
+      if (editorState.pendingPictureUrl == null &&
+          !editorState.pictureCleared) {
+        editorBloc.add(InitialPersistedPictureSet(profile.picture));
+      }
+      if (editorState.pendingBannerUrl == null &&
+          editorState.pendingBannerColor == null &&
```
realmeylisdev: **blocking.** This reads `pendingBannerColor` as "the user picked a swatch", but the first seed sets it itself: `InitialPersistedBannerSet` runs `_parseBannerHexColor` and writes the parsed colour into `pendingBannerColor` (`profile_editor_bloc.dart:276-288`). For anyone whose banner is a colour rather than an image, the guard is false from the very first snapshot and the refresh never runs again.

Cached banner `0x33ccbf`, relay banner `https://new.example/b.jpg`, both snapshots through the real bloc:

```
after cached: persisted=0x33ccbf pendingColor=Color(0xFF33CCBF)
after fresh:  persisted=0x33ccbf pendingColor=Color(0xFF33CCBF) effectiveBanner=0x33ccbf
```

The fresher banner never reaches the editor, and Save republishes the stale colour over it. On main the seed only ran on `MyProfileLoaded`, so the fresh banner was the one that landed.

The picture guard two lines up is fine, `InitialPersistedPictureSet` only writes `persistedPictureUrl`.

Left for you because both shapes of the fix have a cost: tracking a `_seededBanner` here mirrors `_seededName`/`_seededBio` but adds a fourth thing to keep in step, and stopping `InitialPersistedBannerSet` from writing `pendingBannerColor` changes what the picker and the preview read.

mobile/lib/screens/profile_setup/widgets/username_field.dart
```diff
@@ -44,71 +44,39 @@ class _UsernameFieldState extends State<UsernameField> {
...
-              // and resolves usernames as lowercase,
-              // so normalizing here avoids a
-              // confusing "invalid format" error
-              // for a typed capital letter.
+              labelText: context.l10n.profileSetupUsernameLabel,
+              filled: true,
+              fillColor: context.vineColors.surfaceContainer,
+              fillBorderRadius: profileFormCardRadius,
+              primaryWhenFilled: true,
+              suffixText: '@divine.video',
```
realmeylisdev: The body flags this as a pre-existing disagreement between the copy and the issued address, but main had it right and this changes it. Main rendered `prefixText: '@'` + `suffixText: '.divine.video'`, so the field read `@alice.divine.video`. Here it reads `alice@divine.video`.

`@alice.divine.video` is what the rest of the stack uses:

* the claim writes `_@alice.divine.video` into kind 0 (`profile_repository.dart:1008`)
* divine-name-server returns that as `underscore_subdomain` and serves `names: {"_": pubkey}` off the subdomain (`username.ts:547`, `nip05.ts:38`)
* `UserProfile.displayNip05` normalises everything to `@alice.divine.video`, and NIP-05 says a client should show `_@domain` as just the domain
* the affix test this PR adds to `divine_text_field_test.dart` uses `prefixText: '@'` + `suffixText: '.divine.video'`

`<handle>@divine.video` is the legacy shape `divineUsername` still parses for old profiles. Nothing issues it today, so the field now names an address the user won't get.

`DivineTextField` takes `prefixText` after this PR, so the swap is one line, but the Figma frame does draw `@divine.video` and picking which wins is yours, not mine.

realmeylisdev (changes requested):
pushed 5 commits onto the branch (screen-reader taps on the form cards and the colour swatches, the bio counter eating taps on the field, an empty image link undoing a Remove, and a test that wasn't pinning what its name claimed) and rebased onto main, since the pre-push hook wouldn't take anything while the conflict was there. revert any of them freely.

left two for you, both fallout from the new cache-first seeding: a stale cached external NIP-05 latches the editor into external mode and Save then wipes a divine handle set elsewhere, and the banner refresh guard reads the colour the seed itself derived as a user edit. plus the `@divine.video` suffix, which main spelled the other way round. details inline.

NotThatKindOfDrLiz:
@hm21 I pushed takeover commit `bb91dc608` to close out the requested changes.

What changed:

- Fixed cache-first profile seeding so a stale cached external NIP-05 can no longer keep Edit Profile in external mode after the relay reports a Divine handle.
- Fixed banner refresh seeding so an untouched cached colour banner can be replaced by a fresher relay banner.
- Applied the product decision that Divine handles render as `@name` in the username field while the underlying registry/NIP-05 value remains `_@name.divine.video`.
- Removed the two newly orphaned l10n keys (`profileEditPublicKeyLink`, `profileSetupDisplayNameHelper`) from all ARBs and regenerated localizations.
- Updated the PR body and linked issue to reflect the final behavior.

Verification:

- `flutter test test/screens/profile_setup_screen_test.dart test/screens/profile_setup/username_field_test.dart test/screens/profile_setup/display_name_field_test.dart`
- `flutter test test/l10n/hardcoded_visible_strings_test.dart`
- Pre-push changed-file tests
- GitHub checks are green, including Mobile CI, generated-files, web preview, and all test shards.

NotThatKindOfDrLiz (approved):
Approved after takeover commit bb91dc608. I verified the stale NIP-05 mode fix, untouched colour-banner refresh, @name handle display, l10n cleanup, updated PR/issue text, local focused tests/analyze, pre-push checks, and green GitHub checks.

rabble (approved):
Re-reviewed the current head after the takeover fixes. The stale cache-first NIP-05 mode, untouched color-banner refresh, Divine handle presentation, tap targets, and l10n cleanup described in the prior requested-change review are addressed with focused tests. The branch is mergeable and current-head CI is green. Ready for the author to merge.

mobile/lib/blocs/profile_editor/profile_editor_state.dart
```diff
@@ -357,9 +373,19 @@ final class ProfileEditorState extends Equatable {
...
+    if (bannerCleared) return null;
     return persistedBanner;
   }
 
+  /// Whether the user asked for the banner to be removed outright.
+  ///
+  /// Distinguishes "no staged change" from "staged the absence of a banner":
+  /// only the latter makes Save publish `null` over a persisted value. Staging
+  /// a new image or colour cancels it.
+  final bool bannerCleared;
```
realmeylisdev, line 421: #6657 landed on main last night and added `isBannerCleared` to this same class for the same job, so the rebase collides on the concept and not just on the text.

i'd keep this pair over main's. `bannerCleared` + `pictureCleared` covers the avatar too, which is the point of the clear affordance this PR adds, and main only ever did the banner. that does mean rewriting #6657's `isBannerCleared` uses onto it.

rest of what a trial rebase onto `1c71a15a6` hit, so you know the size up front: `profile_setup_view.dart` (main rewrote ~390 lines of it, this rebuilds the file), `profile_setup_listeners.dart`, `profile_editor_bloc.dart` (main sets `pendingPictureUrl: stagedPictureUrl` off the new `StagedProfileMediaStore`, this sets `result.cdnUrl`), and both test files. the unsaved-changes guard #6657 adds also needs a home in the new design, and it isn't in the Figma frames.

mobile/lib/screens/profile_setup/widgets/profile_setup_listeners.dart
```diff
@@ -256,25 +385,20 @@ class ProfileSetupListeners extends ConsumerWidget {
             switch (state.pendingBannerStatus) {
               case PendingBannerStatus.staged:
                 ScaffoldMessenger.of(context).showSnackBar(
-                  SnackBar(
-                    content: Text(
-                      context.l10n.profileSetupBannerUploadSuccess,
-                    ),
-                    backgroundColor: VineTheme.vineGreen,
+                  DivineSnackbarContainer.snackBar(
+                    context.l10n.profileSetupBannerUploadSuccess,
```
realmeylisdev: this key doesn't exist on main any more. #6653 swapped this exact call to `profileSetupUploadStaged` and deleted `profileSetupBannerUploadSuccess` from every locale, so once you rebase this stops compiling.

keep the `DivineSnackbarContainer` wrapper and just swap the key. worth not resolving this one as ours: #6653 was specifically about a staged banner not being saved yet, and "Banner updated" is the claim it was fixing.

realmeylisdev (approved):
the two seeding things i flagged are fixed. checked both by breaking them and watching the new tests go red, so approving from my side.

heads up though: #6657 landed on main a few hours ago and rewrites a lot of this same screen. the conflict is more than l10n now, details inline.

mobile/lib/screens/profile_setup/widgets/profile_setup_listeners.dart
rabble: Fixed in `bb91dc60` (@NotThatKindOfDrLiz). The mode is no longer latched by a one-shot: both seeds are keyed on the value *and* gated on `_usernameFieldIsUntouched`, and the divine-handle branch now emits `Nip05ModeChanged(Nip05Mode.divine)` alongside `InitialUsernameSet` while clearing `_seededExternalNip05`. The external branch additionally requires `extractedUsername == null`, so your repro — cached `alice@example.com`, fresher `_@bob.divine.video` — now ends in `divine` mode and Save publishes `bob`.

Resolving as fixed. Thanks for driving it through `ProfileSetupScreenView` rather than reading it off the diff; the event trace is what made the latch obvious.

mobile/lib/screens/profile_setup/widgets/profile_setup_listeners.dart
rabble: Fixed in `bb91dc60` (@NotThatKindOfDrLiz), and it took the third option rather than either of the two you costed. The guard is now `_bannerStillMatchesSeed`, which compares the staged `pendingBannerColor` against the *seeded* banner (`_normalizedBannerColor` / `_bannerColorString`) instead of treating any non-null pending colour as "the user picked a swatch". So a colour-banner user whose pending colour still equals the seed keeps accepting fresher snapshots, and a genuinely picked swatch still blocks the refresh — without a fourth `_seeded*` flag to keep in step and without changing what the picker and preview read.

Resolving as fixed.

mobile/lib/screens/profile_setup/widgets/username_field.dart
rabble: Leaving this one open — `bb91dc60` changed it, but to a third thing rather than to either option in your comment.

The field is now `prefixText: '@'` with no suffix, so it reads `@alice`. Not the PR's `alice@divine.video`, which was your objection, but not main's `@alice.divine.video` either. Your substantive point stands: the field still does not name the address the registry issues, it just no longer names a *wrong* one. Whether `@alice` is the right call is product, not review — the handle is what the user types and what gets claimed, and the Figma frame draws the domain.

@hm21 this needs an explicit decision rather than a silent third option, so flagging rather than resolving.

hm21: This was actually a decision made by the design team, not something I came up with myself, as you can see in Figma [here](https://www.figma.com/design/rp1DsDEUuCaicW0lk6I2aZ/UI-Design?node-id=11251-229368). 

In my opinion, it's also fine because we're not referring to NIP-05 at this point and just show the username. It's mainly about making the UI look visually appealing. Users who are familiar with Nostr, which will probably be the minority, can find the correct NIP-05 reference below in the NIP-05 option, just like in the design.

PR #6628, opened 2026-08-03
fix(support): replace alert dialogs with full-screen support flows

hm21:
Closes #6629

## Description

Bug report and feature request no longer open as Material `AlertDialog`s. They now live as full-screen support routes using `DivineTextField`, `DivineButton`, and the existing support BLoC/Cubit submission path.

This keeps the design-system migration from the original PR while matching the app guidance to prefer full-screen flows for form-heavy support work. It also removes the bottom-sheet-only `SupportFormScope` mechanism: each route owns and disposes its form controllers directly, while the shared `SupportFormFields` and `SupportFormActions` pieces keep the common required-field and footer behavior deduped.

Notable behavior:

- Support center actions push `/support-center/report-bug` and `/support-center/request-feature` routes.
- The send/cancel actions stay pinned below the scrollable form body.
- Dragging the form content no longer dismisses and discards a filled-in support request.
- Success pops the support route and confirms with a snackbar; failure keeps the route open with the reason.
- `support_dialog_utils.dart` remains deleted, and the raw dialog/material button/text style/color baselines stay reduced.

## Verification

- `flutter test test/widgets/bug_report_dialog_test.dart test/widgets/feature_request_dialog_test.dart` — 22 passed
- `flutter test test/screens/settings/support_center_screen_test.dart test/blocs/bug_report test/blocs/feature_request` — 12 passed
- `flutter test test/l10n test/widgets/bug_report_dialog_log_summary_test.dart` — 100 passed
- `flutter analyze lib test` — clean
- `./scripts/check_raw_dialog_ceiling.sh` — OK
- `./scripts/check_material_button_ceiling.sh` — OK
- `./scripts/check_raw_textstyle_ceiling.sh` — OK
- `./scripts/check_raw_colors_ceiling.sh` — OK
- `./scripts/check_ui_service_boundary.sh` — OK

## Type of Change

- [x] Code refactor
- [x] Bug fix (non-breaking change which fixes an issue)

PR #6632, opened 2026-08-03
fix(settings): rebuild the account danger-zone gates on Divine sheet components

### LINKED ISSUE #6633: fix(settings): rebuild the account danger-zone gates on Divine sheet components
## Problem

The Nostr settings danger-zone confirmations used raw Material dialogs and controls for the most destructive account actions:

- "Remove this account from this device"
- "Delete account and data"
- Account-deletion progress

That left the flow visually inconsistent with Divine sheets and kept `delete_account_dialog.dart` on the raw dialog, raw `TextStyle`, and Material button design-system baselines.

## Expected

The danger-zone gates should use Divine sheet components and Divine typography, preserve the existing destructive-action safeguards, and keep cancellation paths non-destructive.

## Done When

- The remove-keys gate uses `VineBottomSheetPrompt`.
- The delete-account gate uses `VineBottomSheet` with pinned actions and sheet-owned scroll wiring.
- Dismissal and explicit Cancel paths are covered by widget tests and do not call confirmation handlers.
- The account-deletion progress indicator no longer uses a raw `showDialog` or raw `TextStyle`s.
- The design-system baselines are regenerated to lock in the reductions.

## PR

Implemented by #6632.

hm21:
Closes #6633

## Description

| Before | After |
| ------ | ------ |
|   image   |   image |

Both entries in the Nostr settings danger zone were raw Material `AlertDialog`s with grey `OutlineInputBorder` fields, `TextButton`/`ElevatedButton` actions, a `CheckboxListTile`, and hand-rolled `TextStyle(fontSize: 20, fontWeight: bold)` headings -- none of which exists in `divine_ui` or in the Figma library. The two most destructive confirmations in the app were the ones that looked least like Divine. This rebuilds both on the sanctioned sheet components.

**"Remove this account from this device"** is now a `VineBottomSheetPrompt` (skeleton-key sticker, destructive primary action). It returns whether the user confirmed instead of taking an `onConfirm` callback, which drops a callback layer from the caller in `nostr_settings_screen.dart` -- the sign-out, progress-overlay and snackbar handling is unchanged.

**"Delete account and data"** is a `VineBottomSheet`. Notable decisions:

- **The form scrolls through the sheet's own `ScrollController`** (`buildScrollBody`), so dragging it down collapses and dismisses the sheet.
- **The actions moved into `bottomInput`**, so they stay reachable instead of scrolling away with the warning text.
- **Footer and form are sibling slots**, not one subtree: a `ValueNotifier<bool>` carries "the typed token matches" out to the footer, and the footer reaches the form through a `GlobalKey` to confirm. The controllers deliberately stay inside the form's `State` rather than in a holder disposed after `VineBottomSheet.show` completes -- the exit animation still rebuilds the field at that point, and a `TextField` re-subscribes to its controller on every rebuild, which throws once it is disposed.
- **The burn-username consent uses `DivineRowCheckbox`, not `DivineCheckboxTile`.** The latter is `ListTile`-based, and a `ListTile` cannot paint its background or ink splashes through the `ColoredBox` the sheet paints its surface with -- Flutter asserts on exactly that.
- **Swipe-to-dismiss stays enabled** rather than reproducing the old `barrierDismissible: false`. Dismissing cancels; nothing is deleted until the token is typed and the destructive button is tapped.
- **The deletion progress indicator is now a non-dismissible `VineBottomSheet`** instead of the file's last raw `showDialog`, with its copy moved to `VineTheme` typography.
- **The progress sheet pins `useRootNavigator: true`.** `showDialog` defaults it to true and `VineBottomSheet.show` to false, so the migration would otherwise have moved a UI-blocking overlay off the root navigator. `dismissProgressSheet` pops that same navigator rather than going through `context.pop()`, which resolves to the innermost one. No behaviour changes at either current caller -- both are root-level routes.

The `material_buttons`, `raw_dialogs` and `raw_textstyle` baselines all shrink for this file and are regenerated so the reclaimed headroom is locked in rather than left open for the next PR to spend.

## Out of Scope

- `delete_account_dialog.dart` keeps its filename although it now holds sheets. Renaming would churn the `ui_service_boundary` and dialog baselines for no behavioural gain; the public functions were renamed to `...Sheet` instead.
- Independent of #6628 (same alert-to-sheet theme in the support forms). Both target `main` and neither depends on the other; they only touch adjacent design-system baselines.

## Verification

- `flutter analyze lib test` -- clean
- `flutter test test/widgets/delete_account_dialog_test.dart` -- 43 tests, including coverage for pinned actions, drag-to-dismiss wiring, dismissal-cancels, and the explicit delete-sheet Cancel button.
- `flutter test test/screens/settings/support_center_screen_test.dart test/screens/settings/settings_screens_light_mode_test.dart test/widgets/nostr_settings_screen_test.dart` -- green (the deletion flow is shared with the support centre)
- `check_raw_dialog_ceiling.sh`, `check_material_button_ceiling.sh`, `check_raw_textstyle_ceiling.sh`, `check_raw_colors_ceiling.sh` -- all OK after regenerating the shrunk baselines
- Manually verified on a real Samsung Galaxy: both sheets, swipe-to-dismiss, pinned footer, and the type-to-confirm gate with the keyboard open

## Type of Change

### DIFF
--- a/mobile/lib/l10n/app_en.arb
+++ b/mobile/lib/l10n/app_en.arb
@@ -4060,6 +4060,7 @@
   "deleteAccountConfirmUsernamePrompt": "To confirm, type your username:",
   "@deleteAccountConfirmUsernamePrompt": {"description": "Prompt above the confirmation field when the account has a username; the user re-types their username to confirm deletion."},
   "deleteAccountConfirmationHint": "Type DELETE",
+  "@deleteAccountConfirmationHint": {"description": "Label on the delete-confirmation field when the user must type the DELETE token. Shown inside the field when empty and floating above it once focused, so it stays visible while typing."},
   "deleteAccountConfirmationHintUsername": "Type your username",
   "@deleteAccountConfirmationHintUsername": {"description": "Label on the delete-confirmation field when the user must type their username. Shown inside the field when empty and floating above it once focused, so it stays visible while typing."},
   "deleteAccountContentDeletionFailed": "Failed to delete content from relays",
--- a/mobile/lib/l10n/app_en.arb
+++ b/mobile/lib/l10n/app_en.arb
@@ -4061,7 +4061,7 @@
   "@deleteAccountConfirmUsernamePrompt": {"description": "Prompt above the confirmation field when the account has a username; the user re-types their username to confirm deletion."},
   "deleteAccountConfirmationHint": "Type DELETE",
   "deleteAccountConfirmationHintUsername": "Type your username",
-  "@deleteAccountConfirmationHintUsername": {"description": "Hint text in the delete-confirmation field when the user must type their username."},
+  "@deleteAccountConfirmationHintUsername": {"description": "Label on the delete-confirmation field when the user must type their username. Shown inside the field when empty and floating above it once focused, so it stays visible while typing."},
   "deleteAccountContentDeletionFailed": "Failed to delete content from relays",
   "deleteAccountDeleteAllContentButton": "Delete All Content",
   "deleteAccountDeletionIncomplete": "We couldn't finish deleting your account. Try again.",
--- a/mobile/lib/l10n/generated/app_localizations.dart
+++ b/mobile/lib/l10n/generated/app_localizations.dart
@@ -12131,7 +12131,7 @@ abstract class AppLocalizations {
   /// **'To confirm, type your username:'**
   String get deleteAccountConfirmUsernamePrompt;
 
-  /// No description provided for @deleteAccountConfirmationHint.
+  /// Label on the delete-confirmation field when the user must type the DELETE token. Shown inside the field when empty and floating above it once focused, so it stays visible while typing.
   ///
   /// In en, this message translates to:
   /// **'Type DELETE'**
--- a/mobile/lib/l10n/generated/app_localizations.dart
+++ b/mobile/lib/l10n/generated/app_localizations.dart
@@ -12137,7 +12137,7 @@ abstract class AppLocalizations {
   /// **'Type DELETE'**
   String get deleteAccountConfirmationHint;
 
-  /// Hint text in the delete-confirmation field when the user must type their username.
+  /// Label on the delete-confirmation field when the user must type their username. Shown inside the field when empty and floating above it once focused, so it stays visible while typing.
   ///
   /// In en, this message translates to:
   /// **'Type your username'**
--- a/mobile/lib/screens/settings/nostr_settings_screen.dart
+++ b/mobile/lib/screens/settings/nostr_settings_screen.dart
@@ -166,50 +166,46 @@ class _RemoveKeysTile extends StatelessWidget {
         context.l10n.nostrSettingsCouldNotRemoveKeys;
     final failedToRemoveKeysFn = context.l10n.nostrSettingsFailedToRemoveKeys;
 
-    await showRemoveKeysWarningDialog(
-      context: context,
-      onConfirm: () async {
-        if (!context.mounted) return;
+    final confirmed = await showRemoveKeysWarningSheet(context);
+    if (!confirmed || !context.mounted) return;
 
-        // Resolved before the await: the spinner is an overlay entry, not a
-        // route, so a back press during sign-out pops this screen instead of
-        // the spinner. The app-level messenger outlives that pop; a
-        // post-await `ScaffoldMessenger.of(context)` would not, and the user
-        // would be left with no word on a key deletion that failed.
-        final messenger = ScaffoldMessenger.of(context);
-        final progressOverlay = ModalProgressOverlay.show(context);
+    // Resolved before the await: the spinner is an overlay entry, not a
+    // route, so a back press during sign-out pops this screen instead of
+    // the spinner. The app-level messenger outlives that pop; a
+    // post-await `ScaffoldMessenger.of(context)` would not, and the user
+    // would be left with no word on a key deletion that failed.
+    final messenger = ScaffoldMessenger.of(context);
+    final progressOverlay = ModalProgressOverlay.show(context);
 
-        try {
-          await authService.signOut(
-            deleteKeys: true,
-            abortOnKeyDeletionFailure: true,
-          );
-          progressOverlay.dismiss();
-          if (!context.mounted) return;
-          context.go(WelcomeScreen.path);
-        } on SecureKeyStorageException {
-          progressOverlay.dismiss();
-          // Platform key deletion failed — user stays signed in and can
-          // retry without having to log back in.
-          if (!messenger.mounted) return;
-          messenger.showSnackBar(
-            DivineSnackbarContainer.snackBar(
-              couldNotRemoveKeysMessage,
-              error: true,
-            ),
-          );
-        } catch (e) {
-          progressOverlay.dismiss();
-          if (!messenger.mounted) return;
-          messenger.showSnackBar(
-            DivineSnackbarContainer.snackBar(
-              failedToRemoveKeysFn('$e'),
-              error: true,
-            ),
-          );
-        }
-      },
-    );
+    try {
+      await authService.signOut(
+        deleteKeys: true,
+        abortOnKeyDeletionFailure: true,
+      );
+      progressOverlay.dismiss();
+      if (!context.mounted) return;
+      context.go(WelcomeScreen.path);
+    } on SecureKeyStorageException {
+      progressOverlay.dismiss();
+      // Platform key deletion failed — user stays signed in and can
+      // retry without having to log back in.
+      if (!messenger.mounted) return;
+      messenger.showSnackBar(
+        DivineSnackbarContainer.snackBar(
+          couldNotRemoveKeysMessage,
+          error: true,
+        ),
+      );
+    } catch (e) {
+      progressOverlay.dismiss();
+      if (!messenger.mounted) return;
+      messenger.showSnackBar(
+        DivineSnackbarContainer.snackBar(
+          failedToRemoveKeysFn('$e'),
+          error: true,
+        ),
+      );
+    }
   }
 }
 
--- a/mobile/lib/widgets/delete_account_action.dart
+++ b/mobile/lib/widgets/delete_account_action.dart
@@ -67,7 +67,7 @@ Future<void> startAccountDeletionFlow({
     handle: profile?.displayNip05,
   );
 
-  await showDeleteAllContentWarningDialog(
+  await showDeleteAllContentWarningSheet(
     context: context,
     confirmation: confirmation,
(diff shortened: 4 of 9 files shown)


mbradley (dismissed):
Took this over per `divine-context/PR_REVIEW.md`. Four commits added, one per finding. Your two commits are rebased onto current `main`, authorship preserved.

**Rebase.** #6628 landed and conflicted on `mobile/scripts/baseline/material_buttons.txt`. Both PRs deleted a different entry from the same region, so both deletions applied: `delete_account_dialog.dart` here, `feature_request_dialog.dart` and `bug_report_dialog.dart` from #6628. Please check the resolution.

**No behaviour changed by my commits.** `confirm()` re-derives the token from the controller before popping and calling `onConfirm`; pre-PR the button's `onPressed` popped and deleted with no re-check. Ordering, error handling and `executeAccountDeletion` are untouched by this PR.

**Coverage gaps, mutation-tested:**

- `confirmed ?? false` changed to `?? true` left all 40 tests green. Dismissal is newly reachable; the replaced dialogs were `barrierDismissible: false`.
- Dismissing the delete sheet with the token typed was not asserted to skip `onConfirm`.

Two tests added. Both dismiss via system back rather than a barrier tap, so they do not depend on sheet height, and both fail in the full-file run under default order and seeds 98765, 555, 31337.

**Three smaller commits:** remaining "dialog" wording (identity-header doc, test helper, three test names); a clause stating why `canConfirm` can be disposed where the adjacent comment says the controllers cannot; ARB description changed from hint text to label, with `app_localizations.dart` regenerated, since the description feeds the generated dartdoc.

**Validation, on the rebased branch:** `dart format --set-exit-if-changed`, `flutter analyze lib test`, 42/42 on default plus the three seeds above, sibling settings suites, ARB consistency, hardcoded-strings guard, four design-system ceilings, untested-services and test-structure floors.

**Not fixed, and the reason for changes-requested.** `dismisses the sheet when the form is dragged down` cannot fail in CI. Removing `controller: widget.scrollController` gives +0 -1 run alone and +40 in the full file. CI runs the full file, and `very_good test --optimization` bundles untagged suites into one isolate, so the test is not first and does not fail. Reproduced in two checkouts.

It fails only as the first test in the isolate; any preceding test masks it. Inferred cause: the form overflows the viewport only before font metrics warm up, after which the inner scroll view has no extent and the drag reaches the sheet regardless. The masking is observed, the cause is not proven, so the fix depends on confirming it. Options: pin the layout before the drag, assert something independent of warm-up, or remove the test and rely on `keeps the actions out of the scrollable form`, which fails in both modes.

**Questions:**

1. `useRootNavigator` is `false` for the sheet and `true` for the progress dialog, each the respective default. No effect at either current caller. `startAccountDeletionFlow` is the shared entry point. Leave as is?
2. `Closes #6633` is an auto-created stub with no acceptance criteria. Where is the report the description refers to?
3. #6628 added `hintText` to `DivineTextField`. The confirmation field uses `labelText` only. Use both?
4. The final-confirmation title now renders in the sheet header's `DefaultTextStyle` rather than `VineTheme.error`. Intentional?

Please review the four commits and the rebase resolution, then merge or follow up.

hm21:
Thanks for the takeover — rebase resolution checked (both #6628 deletions plus ours, nothing lost) and all four of your commits reviewed; reasoning and validation reproduced on my side.

**Blocker resolved in `06ff94cc0a`, keeping the test.** Your font-warm-up inference was correct: pinning the viewport to 400×400 before opening forces the form to overflow regardless of warm-up, and the drag test now fails under the controller mutation in the full-file run — default order and seeds 98765/555/31337 all go +41 −1, and stay 42/42 unmutated. Dropping the test would have left the exact wiring unguarded that broke in #6618, so fixing it was worth it.

`e60fc75c05` also gives the sibling `deleteAccountConfirmationHint` the label metadata your ARB commit gave the username variant — it had none at all.

**Your questions:**

1. **`useRootNavigator` — leave as is.** Both callers are root-level `GoRoute`s so it makes no difference today, and the defaults match every other `VineBottomSheet`/progress-dialog call site in the repo.
2. **The report is the PR description.** #6633 is the auto-created stub that satisfies the linked-issue gate; the defect (raw Material on both danger-zone gates, off design system) is obvious enough that we don't file a separate written report for this class of fix.
3. **`labelText` only.** The token is already rendered above the field in red monospace; a `hintText` would be a third copy of the same string on one screen.
4. **Intentional.** The danger signalling now comes from the red monospace token and the filled error button, which is stronger than the old red title and matches how the design system marks destructive actions.

NotThatKindOfDrLiz:
@hm21 I pushed a takeover commit on top of the rebased branch: `f158be9f0`.

What changed:

- Added a widget test for the explicit delete-sheet `Cancel` button after the destructive action is armed; it closes the sheet and does not call `onConfirm`.
- Migrated the remaining account-deletion progress UI from raw `showDialog` + raw `TextStyle`s to a non-dismissible `VineBottomSheet` using `VineTheme` typography.
- Regenerated `raw_dialogs` and `raw_textstyle` baselines so `delete_account_dialog.dart` no longer carries those remaining entries.
- Updated #6633 and the PR description to reflect the progress-sheet cleanup and 43-test count.

Local/pre-push verification:

- `flutter test test/widgets/delete_account_dialog_test.dart` -- 43/43 passed
- `flutter analyze lib/widgets/delete_account_dialog.dart test/widgets/delete_account_dialog_test.dart` -- clean
- `bash scripts/check_raw_dialog_ceiling.sh`
- `bash scripts/check_raw_textstyle_ceiling.sh`
- `bash scripts/check_material_button_ceiling.sh`
- `bash scripts/check_raw_colors_ceiling.sh`
- Pre-push hook: no merge conflicts with `main`, analyzer clean, changed-file widget test passed

Please review the takeover commit and merge when you are ready, assuming CI stays green and Matthew is satisfied with the requested-change follow-up.

NotThatKindOfDrLiz (approved):
Approved after takeover. The requested Cancel-button coverage is now present, the remaining progress UI in the touched file is on VineBottomSheet/VineTheme instead of raw showDialog/TextStyle, the baselines are locked down, and CI is green. hm21 should still review the takeover commit before merging.

rabble:
@mbradley re-requesting review — the item you held changes-requested on is fixed.

Your blocker was that `dismisses the sheet when the form is dragged down` could not fail in CI: it only fails as the first test in the isolate, and `very_good test --optimization` bundles the file so it never is. hm21's `5e91bb44` pins the viewport to 400x400 with `devicePixelRatio: 1` so the form overflows regardless of font warm-up, which gives the inner scroll view extent to consume the drag every time.

That also confirms your inferred cause rather than just working around it — the commit message reports mutation-testing it by removing `controller: widget.scrollController`: green in the full-file run before, red after under default order and seeds 98765, 555 and 31337, and green unmutated under all four. The view overrides are paired with `addTearDown` resets, so it doesn't strand later suites in the merged isolate.

@NotThatKindOfDrLiz approved at `f158be9f` after adding the Cancel-button coverage.

Your four questions are still open and belong to @hm21 — flagging so they don't get lost behind the re-review.

rabble:
@mbradley answering your four questions. One turned out to be a real regression and is fixed in `9f2857ec5`; the other three are decisions, no code change.

**4. The final-confirmation title losing `VineTheme.error` — not intentional.** You were right to ask. On `main` it was `TextStyle(color: VineTheme.error, fontSize: 20, fontWeight: bold)`; moving the gate from `AlertDialog` to `VineBottomSheet` left a bare `Text`, so it inherited the header's `DefaultTextStyle` and rendered in primary text. That is the destructive-action signal on the last screen before an irreversible delete, and #6633's own "Expected" section asks the rebuild to *preserve the existing destructive-action safeguards* — so it fails the issue's own criteria, not just taste.

Fixed by passing `VineTheme.titleMediumFont(color: VineTheme.error)` — the same helper `vine_bottom_sheet_header.dart:102` applies — so only the colour differs from the sheet default and the raw-`TextStyle` ceiling is untouched (`check_raw_textstyle_ceiling.sh` green). Added a test asserting the colour; it goes red when the style is dropped. 44/44 in the file.

**1. `useRootNavigator` — leave as is.** Each value is its widget's own default, and you already confirmed no effect at either current caller. They are not actually inconsistent in intent: the sheet wants the local navigator so it composes with shell routes, and the progress overlay wants root so it covers everything including a shell. Pinning either explicitly would freeze an assumption neither caller exercises. Worth revisiting only if a caller opens the flow from inside a nested navigator.

**2. `Closes #6633` — there is no separate report.** The issue is auto-generated *from this PR's own description*, so it is circular rather than empty. It is not quite a bare stub though: it carries a Problem list (the three raw-Material danger-zone gates) and an Expected section, which is what I used to judge question 4 above. Nothing external to chase.

**3. `hintText` + `labelText` — `labelText` only.** The field sits directly under the required token rendered in monospace error colour, so a hint would restate what is already on screen a line above, and both would be visible at once on an empty field. The ARB keys are named `...ConfirmationHint*` for historical reasons only — your `5a7d6032` and hm21's `58112735` both already redescribed them as labels, which is the right read.

Please re-review when you have a moment. Rebase resolution and the four commits you pushed are unchanged by this.

hm21:
@mbradley re-review please — your blocker is fixed, and question 1 turned out to hide a real regression.

**Blocker.** `dismisses the sheet when the form is dragged down` now pins the viewport to 400x400 at `devicePixelRatio: 1` before opening, so the form overflows regardless of font warm-up and the inner scroll view always has extent to consume the drag. Mutation-tested by removing `controller: widget.scrollController`: red in the full-file run under default order and seeds 98765, 555 and 31337, green unmutated under all four. The view overrides are paired with `addTearDown` resets, so nothing leaks into the merged isolate.

**Question 1 — not harmless after all, fixed in `76d11b6a78`.** Migrating the progress overlay from `showDialog` to `VineBottomSheet.show` moved it from the root navigator to the local one, because the two have opposite `useRootNavigator` defaults. No effect at either current caller — both settings routes are top-level — so the earlier "each is its own default, leave as is" described the intent but not what the code now does. It is pinned explicitly now, and `dismissProgressSheet` pops that same navigator instead of relying on `context.pop()`, which resolves to the innermost navigator and would target the wrong one for a shell-nested caller. No test can fail on this until such a caller exists, which is why the commit ships without one.

Questions 2 and 3 stand as answered. Question 4 is fixed with a test asserting the title colour.

Note the branch was rebased onto current `main`, so the SHAs quoted earlier in this thread no longer resolve.

Validation on `76d11b6a78`: `dart format`, `flutter analyze`, 44/44 in `delete_account_dialog_test.dart`, plus `nostr_settings_screen_test.dart` and `support_center_screen_test.dart`.

mbradley:
@hm21 apologies, I was in transit earlier today with unexpectedly intermittent connectivity, which wasn't the plan.

PR #6639, opened 2026-08-03
feat(profile): add search to the follower and following lists

### LINKED ISSUE #6640: feat(profile): add search to the follower and following lists
Auto-created from PR #6639.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6639
**Branch:** `feat/follow-list-search`

hm21:
Closes #6640

## Description

Adds the search field the Figma design has on the follower and following screens, and the data path behind it.

**Design:** https://www.figma.com/design/rp1DsDEUuCaicW0lk6I2aZ/UI-Design?node-id=10534-168961

Shipped on all four screens — my/other × followers/following. The design frame covers only "my following", but the four screens render the same list from the same tile, and a search field on one of four would read as a bug rather than a feature.

**How matching works.** A row qualifies if either side hits:

- **Server-side.** `FollowRepository.searchFollowList` asks the API which members of the list match. This is the only path that finds anyone on a device with a cold profile cache.
- **On-device.** Display names are resolved for the list in chunks and matched locally. Keeps working offline, and covers list members the API never indexed — the follow list is a union of REST, connected relays and indexer relays, while the API only indexes kind-3 events seen on the Divine relay. A pubkey whose profile never resolves still matches on its generated fallback name, so the search always finds what the row actually shows.

**Why the search BLoC does not own the list.** It exposes `visibleFrom(pubkeys)` and each screen keeps passing its own list through, so the rendered rows and the filtered rows cannot drift apart — there is exactly one source of pubkeys. With no query active the call returns the list untouched, so a list nobody is searching pays nothing.

**Why the response echo is checked.** The server-side half depends on a `q` parameter that does not exist in production yet (divinevideo/divine-funnelcake#770). A deployment predating it does not reject the unknown key — it ignores it and answers with the plain first page. Verified against production:

```
$ curl -s 'https://api.divine.video/api/users/<pubkey>/followers?limit=3'
{"followers":["9cfa7d57…","9c66ca38…","aedd8f68…"],"limit":3,"offset":0,"total":7}

$ curl -s 'https://api.divine.video/api/users/<pubkey>/followers?limit=3&q=zzzznobody'
{"followers":["9cfa7d57…","9c66ca38…","aedd8f68…"],"limit":3,"offset":0,"total":7}
```

Byte-identical. Trusting that page would show the first hundred followers as matches for whatever the user typed. So the repository only accepts the result when the response echoes back the filter it applied, and otherwise falls back to on-device matching with a warning in the log.

**This is therefore safe to merge in either order.** Without the backend the behaviour is a local filter; with it, search becomes usable on a cold cache. No feature flag, no transitional code.

**Related Issue:** 

**Depends on (not blocking):** divinevideo/divine-funnelcake#770 — adds the `q` parameter and the applied-filter echo. Still a draft: its ClickHouse integration test needs a run against the local stack.

## Out of Scope

- Making the on-device pass cache-only. Tracked in #6646 for after #770 is deployed: once the server covers the cold-cache case, the local pass can read cached profiles only instead of calling `fetchBatchProfiles` and fanning out to REST/relays.
- Paging server-side matches. One page of up to 100 is enough for a filter UI and the server clamps `limit` to 100 anyway.
- Searching profile bios. The API filter matches name / display name / NIP-05 only, so every hit is one the user can see in the row it renders.

## Verification

- `flutter analyze lib test integration_test packages/models packages/follow_repository packages/funnelcake_api_client` — clean
- `flutter test` on the affected app tests (95: new search BLoC + widget tests, plus the existing followers/following screen and BLoC suites) — green
- `flutter test` on all three changed packages — `models` 837, `follow_repository` 226, `funnelcake_api_client` 372 — green
- `flutter test test/l10n/arb_consistency_test.dart` — green; the follow-list field reuses `exploreSearchHint` instead of carrying a duplicate localized key
- All four design-system ceiling ratchets, the test-structure ratchet and the untested-services floor — green
- `dart format --set-exit-if-changed` on every touched path — clean

**Manual test plan** — verified on a real Samsung Galaxy:

1. Open Profile → Following. The search field sits between the app bar and the list, matching the design.
2. Type a partial display name. The list narrows to matching rows; a NIP-05 handle and a pasted hex pubkey prefix also match.
3. Clear the field. The full list comes back and pull-to-refresh still works.
4. Type something nobody matches. The no-results copy appears instead of an empty list.
5. Repeat on Profile → Followers, and on another user's followers and following lists.
6. Airplane mode: the field still filters on the names already on device.

## Type of Change

### DIFF
--- a/mobile/lib/blocs/follow_list_search/follow_list_search_bloc.dart
+++ b/mobile/lib/blocs/follow_list_search/follow_list_search_bloc.dart
@@ -36,8 +36,9 @@ class FollowListSearchBloc
   /// Creates the search BLoC.
   ///
   /// [subjectPubkey] is whose list this is, and [listKind] which side of it —
-  /// together they address the server-side search. Pass an empty
-  /// [subjectPubkey] to run on-device matching only.
+  /// together they address the server-side search. An empty [subjectPubkey]
+  /// runs on-device matching only; push the real one in later with
+  /// [FollowListSearchSubjectPubkeyChanged].
   ///
   /// [profileRepository] is nullable because `profileRepositoryProvider` is
   /// gated on Nostr readiness and hands over the real instance after cold
@@ -58,11 +59,16 @@ class FollowListSearchBloc
       transformer: debounceRestartable(),
     );
     on<FollowListSearchProfileRepositoryChanged>(_onProfileRepositoryChanged);
+    on<FollowListSearchSubjectPubkeyChanged>(_onSubjectPubkeyChanged);
   }
 
   final FollowRepository _followRepository;
-  final String _subjectPubkey;
   final FollowListKind _listKind;
+
+  /// Swappable for the same reason as [_profileRepository]: on the own-list
+  /// screens this starts empty during cold boot and is pushed in by
+  /// [FollowListSearchSubjectPubkeyChanged] once auth restores.
+  String _subjectPubkey;
 
   /// Swappable rather than final for the same reason as
   /// `ConversationListBloc._profileRepository`: re-pointing this one field is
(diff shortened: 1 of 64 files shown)


NotThatKindOfDrLiz:
@hm21 I took over the review fixes and pushed 9a032d095.

What changed:
- Fixed follow-list search term resolution so partial `fetchBatchProfiles` results no longer pin missing rows to fallback names.
- Guarded resolver generations/repository swaps so stale async chunks cannot resurrect cleared terms.
- Started local name resolution in parallel with remote follow-list search.
- Moved visible-list filtering out of `build()` and added a no-results screen-reader announcement.
- Reused `exploreSearchHint` and removed the duplicate `followListSearchHint` ARB/generated l10n entries.
- Filed #6646 for the post-#770 cache-only local pass and updated the PR body.

Verification:
- Local pre-push hook passed: mergeability check, analyzer, and changed-file tests including follow_repository/funnelcake_api_client/models plus affected follow-list bloc/widget/screen tests.
- `flutter test test/l10n/arb_consistency_test.dart` passed.
- GitHub checks are all green.

NotThatKindOfDrLiz (approved):
Takeover fixes are pushed, local verification passed, and GitHub checks are green. Approved for @hm21 to review and merge when ready.

Issue #6635, opened 2026-08-03
feat(explore): configurable featured hashtag tab


NotThatKindOfDrLiz:
## What would you like?

A reusable, remotely-configured **featured hashtag tab** on the Explore screen.

When the backend serves a configuration, Explore renders an additional tab
whose contents are a themed collection of videos. When no configuration is
served, the tab does not exist and nothing about it is visible in the UI.

The label, position, URL slug, active window, and audience rules all come from
the backend. **Nothing about any specific tab is compiled into the app.** The
app ships with the capability and no configured value.

The client is deliberately thin: it does not know what hashtag a tab
represents, does not match tags, and does not filter or sort. It fetches an
already-gated, already-ordered list by opaque configuration id and renders it.

This is a repeatable system, not a one-off tab — we expect to run several of
these, retargetable without an app release.

## How would this be useful for you?

Today, spotlighting a themed collection on Explore requires a code change, a
release, and an app store review cycle — and un-spotlighting it requires
another one. That makes any time-bounded editorial moment impractical to run
and impossible to stop quickly if something goes wrong.

Server-driven configuration decouples editorial scheduling from the release
train and gives us a kill switch that does not depend on shipping a build.

## When would you use this?

Any time we want to surface a themed collection on Explore for a bounded window
— a seasonal theme, a creator challenge, an editorial collection, a community
event.

## Proposed behavior

**Visibility**

The tab is absent from the tab bar unless *all* of the following hold:

- a valid configuration has been fetched, and
- `enabled` is true, and
- the current time is inside the configured window, and
- `has_content` is true, and
- the account passes the configured audience rules.

No placeholder, no disabled tab, no empty tab. Absent means absent.

**Configuration source**

- `GET /featured-tabs` on Funnelcake, polled every 5 minutes and refetched on
  app foreground. Poll interval is server-supplied, not hardcoded.
- Last good config is cached with a **5 minute TTL**. Past that the tab is
  dropped. One failed request must not make the tab flicker out mid-scroll; a
  sustained outage must not strand it on screen.
- Worst case from a backend kill to the tab disappearing is ~5 minutes.

**Contents**

- `GET /featured-tabs/{id}/videos`, cursor-paginated.
- Rendered in the order received. The server returns curated videos first, then
  approved videos newest-first. **The client does not re-sort or re-filter.**

**Placement**

- Position comes from the config, expressed by tab **name** relative to
  existing tabs — never by index. `_tabNames` / `_tabCount` in
[...]

PR #6641, opened 2026-08-03
fix(explore): persist selected tab by name

### LINKED ISSUE #6638: fix(explore): persist selected tab by name
## Problem

Explore stores the selected tab for feed-mode chrome as a raw index. Optional Explore tabs such as Classics, For You, Apps, and the upcoming configurable featured tab can appear or disappear, which shifts later indices. AppShell also reconstructs its own incomplete Explore title list, so feed-mode titles can be wrong for Categories and Apps.

## User impact

Users can see the wrong Explore feed-mode title after selecting tabs past Trending, and dynamic tab availability can shift tab identity out from under persisted state.

## Acceptance criteria

- Explore selection shared outside the live TabController is stored by stable tab name, not index.
- Feed-mode shell titles resolve from canonical Explore tab names, including Categories and Integrated Apps.
- Tab bar labels continue to match the existing tab copy.
- Focused regression tests cover tab-name persistence and label resolution.

## Technical notes

This is prerequisite cleanup for #6635 so the featured tab can appear/disappear without index-shift bugs.

NotThatKindOfDrLiz:
## Summary

- Persist the selected Explore tab by canonical tab name instead of raw index.
- Resolve Explore feed-mode shell titles from shared tab-name labels, including Categories and Integrated Apps.
- Drive the tab bar labels from the same canonical tab-name order while preserving the existing tab-bar copy.
- Clean up obsolete no-op Explore tab tests and add focused label/persistence regressions.

## Verification

- flutter test test/blocs/explore_tabs/explore_tabs_cubit_test.dart test/screens/explore/explore_tab_labels_test.dart test/screens/explore_screen_tab_switching_test.dart test/screens/explore_screen_lists_tab_test.dart
- flutter analyze
- Pre-push hook: analyze + changed tests passed

Closes #6638
Refs #6635

### DIFF
--- a/mobile/lib/blocs/explore_tabs/explore_tabs_state.dart
+++ b/mobile/lib/blocs/explore_tabs/explore_tabs_state.dart
@@ -12,6 +12,21 @@ const exploreForYouTabSlug = 'for-you';
 /// Default tab name selected when no other selection applies.
 const exploreDefaultTabName = 'new';
 
+/// Internal tab name for the Classics tab.
+const exploreClassicsTabName = 'classics';
+
+/// Internal tab name for the Trending tab.
+const explorePopularTabName = 'popular';
+
+/// Internal tab name for the Categories tab.
+const exploreCategoriesTabName = 'categories';
+
+/// Internal tab name for the Lists tab.
+const exploreListsTabName = 'lists';
+
+/// Internal tab name for the integrated Apps tab.
+const exploreAppsTabName = 'apps';
+
 /// Immutable explore tab configuration derived from feature availability.
 class ExploreTabsState extends Equatable {
   /// Creates a tab state with the given optional-tab availability.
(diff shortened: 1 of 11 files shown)


rabble (changes requested):
One functional gap remains in the persistence path. `ExploreView._onTabChanged()` writes `exploreTabNameProvider`, but `_initTabController()` only resolves `widget.initialTabName ?? previousTabName ?? exploreDefaultTabName`; on widget recreation both of the first two values are null, so the view resets to `new` and never reads the persisted provider value. The updated provider-only tests do not exercise that boundary. Please initialize from `ref.read(exploreTabNameProvider)` after an explicit route name / in-place previous name, and add a widget-level regression that recreates the view with a non-default saved tab name and verifies the restored tab. The stable-name conversion and shared label mapping otherwise look sound.

rabble (approved):
Takeover fix pushed and validated on current `origin/main`.

The recreation bug was reproduced with a widget test: a persisted `popular` tab reopened at `new` because `_initTabController()` never read `exploreTabNameProvider`. The controller now resolves explicit route tab > in-memory tab during availability changes > persisted stable name > the existing name-based fallback. The regression asserts the live `TabBar` controller restores Popular after recreation.

Verification:
- focused Explore suite: 8 tests passed
- full `flutter analyze lib test integration_test`: clean
- pre-push analyzer and changed-file tests: passed
- branch rebased onto current main and pushed with lease

Ready for the author to review and merge.

PR #6642, opened 2026-08-03
fix(profile): flag revalidation on empty tabs so refresh can settle

### LINKED ISSUE #6643: fix(profile): flag revalidation on empty tabs so refresh can settle
Auto-created from PR #6642.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6642
**Branch:** `fix/profile-refresh-hangs-on-empty-tab`

hm21:
Closes #6643

## Description

Pull-to-refresh on the own profile could spin forever and never release.

`ProfileGridView._refreshProfileContent` holds the `RefreshIndicator` until every viewed tab reports its next settled state, waiting on each tab bloc's stream. The saved/liked/reposted sync handlers only emitted `isRefreshing: true` when videos were already on screen. So when a tab had settled *empty*, a re-sync emitted a terminal state equal to the current one — and bloc suppresses equal states. Zero stream events, the awaited future never resolved, spinner stuck until the user left the screen.

Why this hits the own profile hardest: Saved only exists there, and an empty bookmark list is the common case. Liked and Reposted have the same shape and hit both own and other profiles once they are empty.

The fix sets the flag whenever a settled result is already on screen — empty included — so every re-sync has an observable start and end edge. The cold load (`status == initial`) is untouched, so first-load visuals do not change. Side effect worth noting: an empty tab now shows the thin revalidation bar during a refresh, where it previously gave no feedback at all.

Fixing it at the call site instead (timeout on the wait) would have masked the bloc lying about its own refresh state; `isRefreshing` documented itself as "cached content on screen" while the tab bar actually wants "a settled result is on screen", so the doc comments were corrected to match.

Comments is not affected — it emits a `loading` state on the way, which is already an observable transition.

Collabs does not hang for that same reason, but the second commit fixes it anyway: its `loading` emit is exactly what made a settled-empty re-fetch trade the empty grid for a full-tab spinner on every pull, leaving the fourth grid tab behaving differently from the three above. It now keeps the empty grid under the thin bar like the others, still shows the spinner when retrying from a failure screen, and clears `isRefreshing` in its catch block — without that last part an empty grid would reach the failure emit with the flag still on and hang the refresh in the new shape.

The residual hang window behind all of this — `droppable()` discarding a sync event during the post-settle cache write, which no `isRefreshing` edge can rescue — is tracked separately in #6644 and deliberately left out of this PR, since fixing it means moving every profile tab off `droppable()`.

**Related Issue:** 

## Out of Scope

None.

## Verification

- Repro captured as a test first: without the fix `pull-to-refresh completes after a viewed tab settled empty` fails in ~1s; with it, green.
- 3 bloc-level regression tests (saved / liked / reposted) mirror the exact predicate the widget waits on.
- `flutter test` over `test/blocs/profile_*` and `test/widgets/profile` — 430 passing.
- Manually verified on a real Samsung Galaxy: own profile → open Saved (empty) → pull to refresh → spinner releases.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_bloc.dart
+++ b/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_bloc.dart
@@ -55,7 +55,11 @@ class ProfileCollabVideosBloc
     ProfileCollabVideosFetchRequested event,
     Emitter<ProfileCollabVideosState> emit,
   ) async {
-    if (state.videos.isNotEmpty && !state.isRefreshing) {
+    // Flag the revalidation whenever a settled result is on screen — an empty
+    // one included — so a re-sync always has an observable start edge, matching
+    // the liked/reposted/saved tabs.
+    if (state.status != ProfileCollabVideosStatus.initial &&
+        !state.isRefreshing) {
       emit(state.copyWith(isRefreshing: true));
     }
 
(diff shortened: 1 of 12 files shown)


realmeylisdev (approved):
lgtm. traced the four blocs against what the refresh is actually awaiting, then reverted the fix locally to make sure the tests really catch it: all five settled-empty tests go red without it, and the collab failure one catches the `isRefreshing: false` in the catch block on its own. also walked the other three things the refresh awaits (my-profile completer, feed cubit, comments) and none of them can hang the same way, so the fix looks complete for this class.

two small things, neither blocking. the comments `loading` emit is behind an early return at `profile_comments_bloc.dart:47`, so the description's reason for "comments is not affected" is a bit off even though the conclusion holds. and #6644 says all five sync handlers are droppable, but comments has no transformer at all, so that part of the follow-up only really covers four blocs.

PR #6650, opened 2026-08-04
fix(recorder): flush autosave on background

### LINKED ISSUE #6562: feat(support): in App Drafts Ironclad persistence)
### What would you like?

chainz, Aug 1, 2026, 13:22

in App Drafts Ironclad persistence)

I would like to request a feature on when creating a six second skit on Divine that has many cuts or jumps to other scenes inside of it I noticed I sometimes need to pause and switch to another app like amethyst or Damus to Check if the Draft is there so I can change a Voice over a Title or Script and I often due this due to Ram Limitations and that the OS constantly kills divine when it’s running in the background which if u have to open the app back up and did not save a Draft it’s permanently gone at least I think

it would solve a problem that some users have like with Ram and other limitations like Older Models that can run Old Version of an App like Amethyst but can’t run newer versions of Divine

### When would you use this?
I would use this every time while switching between apps

**User Pubkey:** 54c3e1cefe908a33cdb8ee2133ce0a7665ba9b864097900400ae46d47c22db12

it would solve a problem that some users have like with Ram and other limitations like Older Models that can run Old Version of an App like Amethyst but can’t run newer versions of Divine

### When would you use this?

I would use this every time while switching between apps

### Anything else?

(Generated from Zendesk Ticket: rabblelabs.zendesk.com/agent/tickets/4381)

NotThatKindOfDrLiz:
## Summary
- flush pending debounced editor autosaves when the app becomes inactive, paused, or hidden
- restore autosave recovery from the validated autosave draft path
- prompt for clips-only autosave drafts so recorded cuts are reachable after relaunch

## Verification
- flutter test test/providers/video_editor_provider_test.dart
- flutter test test/screens/video_recorder_screen_test.dart
- flutter test test/widgets/app_lifecycle_handler_test.dart
- flutter analyze
- pre-push hook analyzer and changed tests

Closes #6562

### DIFF
--- a/mobile/lib/providers/video_editor_provider.dart
+++ b/mobile/lib/providers/video_editor_provider.dart
@@ -720,6 +720,18 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
     });
   }
 
+  /// Immediately persist a debounced autosave that has not fired yet.
+  ///
+  /// Returns `false` when there is no pending autosave or the write fails.
+  Future<bool> flushPendingAutosave() async {
+    final timer = _autosaveTimer;
+    if (timer == null || !timer.isActive) return false;
+
+    timer.cancel();
+    _autosaveTimer = null;
+    return autosaveChanges();
+  }
+
   /// Automatically save the current video project state.
   ///
   /// This method is typically called periodically or on significant changes
--- a/mobile/lib/screens/video_recorder_screen.dart
+++ b/mobile/lib/screens/video_recorder_screen.dart
@@ -208,12 +208,10 @@ class _VideoRecorderViewState extends ConsumerState<VideoRecorderView>
     );
 
     final draftService = ref.read(draftStorageServiceProvider);
-    final draft = await draftService.getDraftById(
-      VideoEditorConstants.autoSaveId,
-    );
+    final draft = await draftService.getAutosaveDraft();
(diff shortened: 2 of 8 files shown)


NotThatKindOfDrLiz:
Takeover cleanup pushed in `bf06e9f22`.

What changed:
- Hoisted lifecycle autosave flushing into one non-resumed lifecycle path, so `detached` is covered without duplicated switch blocks.
- Removed the redundant `draft.clips.isNotEmpty` gate after `getAutosaveDraft()` and deleted the now-dead `hasValidAutosave()` helper.
- Swapped autosave recovery widget assertions to l10n lookups.
- Hardened the lifecycle regression test with no-op startup service overrides and an explicit 30s background timer drain.

Verification:
- `flutter test test/providers/video_editor_provider_test.dart test/screens/video_recorder_screen_test.dart test/widgets/app_lifecycle_handler_test.dart test/services/draft_storage_service_test.dart`
- `flutter analyze`
- GitHub CI green on `bf06e9f22`

rabble (approved):
Reviewed the current head. The lifecycle hook flushes only a live debounce, covers background/teardown states without duplicate writes, and recovery now uses the validated autosave path so clips-only work remains reachable. Regression coverage exercises both the notifier and lifecycle boundary, and CI is green. Ready to merge.

PR #6654, opened 2026-08-04
fix(library): show in-progress autosave draft

### LINKED ISSUE #6562: feat(support): in App Drafts Ironclad persistence)

NotThatKindOfDrLiz:
## Summary
- show the autosave draft in Library -> Drafts when the drafts BLoC includes it
- remove the DraftsTab-level autosave filter so inclusion stays owned by DraftsLibraryBloc
- add an in-progress badge for autosave draft rows with regenerated l10n output

## Why
Creators need a visible recovery point after Divine autosaves an in-progress recording session. The BLoC already filters out empty autosaves while keeping autosaves with clips; the Library UI was hiding that valid recovery draft.

Closes #6562

## Verification
- mise exec -- flutter test test/widgets/library/drafts_tab_test.dart test/blocs/drafts_library/drafts_library_bloc_test.dart test/l10n/arb_consistency_test.dart
- mise exec -- flutter analyze
- pre-push hook: changed-file analyzer and tests passed

### DIFF
--- a/mobile/lib/blocs/drafts_library/drafts_library_bloc.dart
+++ b/mobile/lib/blocs/drafts_library/drafts_library_bloc.dart
@@ -15,7 +15,7 @@ part 'drafts_library_state.dart';
 /// BLoC for managing draft video projects in the library.
 ///
 /// Loads drafts from [DraftStorageService] and handles deletion.
-/// Filters out autosave and already published drafts.
+/// Filters out empty autosaves and already published drafts.
 class DraftsLibraryBloc extends Bloc<DraftsLibraryEvent, DraftsLibraryState> {
   DraftsLibraryBloc({required DraftStorageService draftStorageService})
     : _draftStorageService = draftStorageService,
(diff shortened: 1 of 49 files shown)


rabble (approved):
Reviewed autosave filtering, library presentation, localization, and widget coverage. No blocking issues found; CI is green.

PR #6655, opened 2026-08-04
fix(account): retarget profile after account switch

### LINKED ISSUE #6656: fix(account): retarget profile after account switch
Auto-created from PR #6655.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6655
**Branch:** `fix/account-switch-profile`

rabble:
Closes #6656

## Summary
- retarget the leaving account own-profile route to the newly active account during an in-place account switch
- preserve other users' profiles, non-profile routes, profile video index, query parameters, and fragments
- normalize profile route identity checks so npub, nprofile, bare hex, uppercase hex, and `me` agree with the profile body
- canonicalize bare-hex public identifiers to lowercase at the shared normalizer so app-wide identity comparisons and provider keys use the same hex form
- add regression coverage for account-switch route seeding, profile-route identifier encodings, and public identifier hex canonicalization

## Root cause
The account container swap correctly replaced auth-scoped services, but it copied the current profile URL verbatim into the new router. When that URL identified the leaving account, the profile screen continued rendering that account while notifications and navigation reflected the target account.

The same profile identity decision was also being made with raw route-string comparisons in shell/profile routing code, while the profile body normalized the route segment to hex. Deep links using bare hex, uppercase hex, or nprofile could therefore make the shell/scaffold disagree with the profile body about whether the route belonged to the signed-in user.

## Verification
- flutter test test/providers/swap_account_test.dart
- flutter test test/utils/npub_hex_test.dart test/utils/public_identifier_normalizer_test.dart
- flutter test test/router/app_shell_own_profile_encoding_test.dart

### DIFF
--- a/mobile/lib/providers/swap_account.dart
+++ b/mobile/lib/providers/swap_account.dart
@@ -11,6 +11,7 @@ import 'package:openvine/providers/environment_provider.dart';
 import 'package:openvine/providers/shared_preferences_provider.dart';
 import 'package:openvine/router/app_router.dart';
 import 'package:openvine/services/auth_service.dart';
+import 'package:openvine/utils/nostr_key_utils.dart';
 import 'package:unified_logger/unified_logger.dart';
 
 /// Signs [container]'s fresh [AuthService] in as [account].
@@ -58,6 +59,38 @@ String? _currentRouterLocation(AccountSwitchController controller) {
       .toString();
 }
 
+/// Retargets an own-profile route from the leaving account to the account that
+/// is about to become active.
+///
+/// Most routes are safe to preserve across an account-container swap. An own
+/// profile URL is identity-bearing, though: carrying `/profile/<old npub>` into
+/// the new container renders the leaving account's profile while the rest of
+/// the app already reflects the target account.
+String? accountSwitchInitialLocation({
+  required String? currentLocation,
+  required String? currentNpub,
+  required String targetPubkeyHex,
+}) {
+  if (currentLocation == null) return null;
+
+  final uri = Uri.tryParse(currentLocation);
+  if (uri == null) return currentLocation;
+  final segments = uri.pathSegments;
+  if (segments.length < 2 || segments.first != 'profile') {
+    return currentLocation;
+  }
+
+  final routedNpub = segments[1];
+  if (routedNpub != 'me' && routedNpub != currentNpub) {
+    return currentLocation;
+  }
+
+  final targetNpub = NostrKeyUtils.encodePubKey(targetPubkeyHex);
+  final targetPath =
+      '/profile/$targetNpub${segments.length >= 3 ? '/${segments.skip(2).join('/')}' : ''}';
+  return uri.replace(path: targetPath).toString();
+}
+
 /// Switches the live app to [account] in place: builds a new container on the
 /// shared [deviceScope], signs it in, and swaps it in via [controller].
 ///
@@ -98,7 +131,12 @@ Future<void> swapAccount({
 }) async {
   await controller.runExclusive(() async {
     await currentAuthService.archiveCurrentSignerInfo();
-    final currentLocation = _currentRouterLocation(controller);
+    final current = controller.currentContainer;
+    final currentLocation = accountSwitchInitialLocation(
+      currentLocation: _currentRouterLocation(controller),
+      currentNpub: current?.read(authServiceProvider).currentNpub,
+      targetPubkeyHex: account.pubkeyHex,
+    );
     final container = buildAccountContainer(
       deviceScope,
       accountOverrides: [
--- a/mobile/lib/providers/swap_account.dart
+++ b/mobile/lib/providers/swap_account.dart
@@ -12,6 +12,7 @@ import 'package:openvine/providers/shared_preferences_provider.dart';
 import 'package:openvine/router/app_router.dart';
 import 'package:openvine/services/auth_service.dart';
 import 'package:openvine/utils/nostr_key_utils.dart';
+import 'package:openvine/utils/npub_hex.dart';
 import 'package:unified_logger/unified_logger.dart';
 
 /// Signs [container]'s fresh [AuthService] in as [account].
@@ -66,9 +67,14 @@ String? _currentRouterLocation(AccountSwitchController controller) {
 /// profile URL is identity-bearing, though: carrying `/profile/<old npub>` into
 /// the new container renders the leaving account's profile while the rest of
 /// the app already reflects the target account.
+///
+/// The routed segment is matched with [routeIdentifiesUser] rather than
+/// compared raw, because the route accepts npub, nprofile and bare hex as well
+/// as the relative `me` — a raw compare leaves the hex and nprofile forms
+/// stranded on the leaving account.
 String? accountSwitchInitialLocation({
   required String? currentLocation,
-  required String? currentNpub,
+  required String? currentPubkeyHex,
   required String targetPubkeyHex,
 }) {
   if (currentLocation == null) return null;
@@ -80,8 +86,7 @@ String? accountSwitchInitialLocation({
     return currentLocation;
   }
 
-  final routedNpub = segments[1];
-  if (routedNpub != 'me' && routedNpub != currentNpub) {
+  if (!routeIdentifiesUser(segments[1], currentPubkeyHex)) {
     return currentLocation;
   }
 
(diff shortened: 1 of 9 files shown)


rabble:
Clean review: I checked profile-route retargeting across account switches, nested paths/query/fragment preservation, and non-owned profile behavior. No blocking issues found and CI is green. GitHub does not allow the active account to approve its own PR, so this is recorded as a review comment rather than an approval.

mobile/lib/providers/swap_account.dart
```diff
@@ -73,7 +106,12 @@ Future<void> swapAccount({
   AccountSignIn signIn = _defaultSignIn,
 }) async {
   await controller.runExclusive(() async {
-    final currentLocation = _currentRouterLocation(controller);
+    final current = controller.currentContainer;
+    final currentLocation = accountSwitchInitialLocation(
```
realmeylisdev, line 140: should-fix: this is the only production behaviour the PR changes, and nothing covers it. I reverted these six lines back to `final currentLocation = _currentRouterLocation(controller);` and `flutter test test/providers/swap_account_test.dart test/screens/settings/settings_account_switch_test.dart` still reported 16/16 passing. Both widget tests in this file pump `child: const SizedBox()`, so `goRouterProvider` is never created and `_currentRouterLocation` returns null at line 51, leaving the retarget path inert. Swapping `.currentNpub` for the sibling `.currentPublicKeyHex` on line 112 would also compile and pass while silently killing the fix. Extending `pumpHost` to override `goRouterProvider` with a router seeded at the leaving account's profile route, then asserting `read(routerInitialLocationProvider)` on the swapped-in container, would pin it.

mobile/test/providers/swap_account_test.dart
```diff
@@ -62,6 +63,56 @@ void main() {
...
+          targetPubkeyHex: targetHex,
+        ),
+        '/profile/$targetNpub/7?source=notification#clip',
+      );
+    });
+
+    test('preserves another user profile and non-profile routes', () {
+      expect(
+        accountSwitchInitialLocation(
+          currentLocation: '/profile/$targetNpub',
```
realmeylisdev: should-fix: this case can't fail. The "another user" here is `targetNpub`, and line 101 passes the matching `targetHex`, so the preserve branch and the retarget branch both produce `/profile/$targetNpub` and the assertion can't tell them apart. I deleted the guard at swap_account.dart:81-84 and this file still reported 5/5 passing. Using a third npub belonging to neither account makes it discriminate. `/profile/me` and a two-segment non-profile path like `/profile-view/$leavingNpub` are also uncovered; `/settings` exits earlier on `segments.length < 2` and never reaches either clause.

mobile/lib/providers/swap_account.dart
```diff
@@ -56,6 +57,38 @@ String? _currentRouterLocation(AccountSwitchController controller) {
...
+
+  final uri = Uri.tryParse(currentLocation);
+  if (uri == null) return currentLocation;
+  final segments = uri.pathSegments;
+  if (segments.length < 2 || segments.first != 'profile') {
+    return currentLocation;
+  }
+
+  final routedNpub = segments[1];
+  if (routedNpub != 'me' && routedNpub != currentNpub) {
```
realmeylisdev: follow-up: this compares the raw path segment against the bech32 npub, but the screen that renders the settings gear decides own-vs-other by normalizing the same segment to hex (profile_screen_router.dart:368 and :377, via `npubToHexOrNull`, which accepts 64-char hex and `nprofile1...` as well as `npub1...`). So `/profile/<own hex>` shows the gear, gets captured verbatim at swap time, and comes back unchanged, leaving the new container on the leaving account's profile. The app never mints that form itself and main retargeted nothing, so this is an incomplete fix rather than a regression, but `mobile/docs/DEEP_LINK_URL_REFERENCE.md:58` lists `https://divine.video/profile/{hex}` as supported. Normalizing both sides and comparing against `currentPublicKeyHex` would close it.

realmeylisdev (changes requested):
the fix itself is right. traced it through the container swap and both `/profile/<npub>` and `/profile/me` land on the new account, and a third party's profile is left alone.

problem is nothing actually tests it. i reverted your call-site change back to the old one-liner and all 16 tests still passed. deleting the identity guard entirely also still passes 5/5, because the "preserves another user profile" case uses the target npub as the other user so both branches produce the same string.

third one is optional: hex and nprofile own-profile urls aren't retargeted. same as before this PR, but the deep link doc lists hex as supported.

mobile/lib/providers/swap_account.dart
realmeylisdev: Covered in 1db45f2034. `pumpHost` now takes overrides and a child builder, and `seededLocationSwitchingFrom` mounts a router seeded at the leaving account's profile route, then asserts `routerInitialLocationProvider` on the container `swapAccount` hands to `signIn`.

Both mutations fail it now: reverting the call site to the one-liner, and keeping the call but reading the sibling `.currentPublicKeyHex`/`.currentNpub` getter. The router has to be created in the container *and* attached to a `Router` — `_currentRouterLocation` gates on `exists`, and `currentConfiguration` stays `RouteMatchList.empty` until something parses the initial location.

mobile/test/providers/swap_account_test.dart
realmeylisdev: Fixed in 0391df1b5e. The preserve case now uses a third identity belonging to neither account, so deleting the identity guard fails it (verified).

Also filled the branches nothing reached: `/profile/me`, `/profile-view/<leaving npub>` (two segments, so it reaches the first-segment guard that `/settings` exits before), an unknown leaving identity, and a null location.

mobile/lib/providers/swap_account.dart
realmeylisdev: Closed rather than deferred — a4019ffc3e adds `routeIdentifiesUser` (normalizes both sides to hex) and dc951ccc8f uses it for the swap guard, so hex, uppercase hex and nprofile own-profile routes now retarget while a stranger's hex is still preserved.

Since the same raw compare was the reason the screens disagreed, a4019ffc3e also applies it at `app_shell.dart` (both sites) and `profile_screen_router.dart`'s `isOwnProfileGrid`. The body already normalized, so those were the three saying "not you" about your own hex URL while the body said "you". `app_shell_own_profile_encoding_test` pins the shell site.

Two things I deliberately did not do: deep-link canonicalization (it changes what URL the app mints from external input, and it is redundant once the comparisons normalize), and a test for `profile_screen_router`'s call site — that screen has no working widget harness, its own test file is `skip: true` throughout. It still has to change with the shell: leaving it raw renders the own-profile body with neither its scaffold nor the shell app bar.

realmeylisdev (approved):
Took this over instead of leaving the threads open. Four commits on top of yours, one per finding, each revertible on its own.

- `0391df1b5e` — the preserve case now uses a third identity, so deleting the identity guard fails it. Plus `/profile/me`, `/profile-view/<npub>`, unknown leaving identity, null location.
- `1db45f2034` — mounts a router seeded at the leaving account's profile route and asserts what `swapAccount` hands the new container. Reverting your call site to the one-liner, or keeping it but reading `.currentNpub`, now both fail.
- `a4019ffc3e` — `routeIdentifiesUser` normalizes npub / nprofile / hex / `me` to hex, applied at both `app_shell` sites and `profile_screen_router`'s `isOwnProfileGrid`. Those three were the ones disagreeing with the profile body, which already normalized.
- `dc951ccc8f` — your guard uses that helper, so switching from `/profile/{hex}` or `/profile/{nprofile}` retargets too. Depends on the commit above; revert the pair together.

Validation: all 14 checks green. Locally, 1371 tests across `test/router`, `test/utils`, `test/providers` and the profile/deep-link suites, `flutter analyze`, and the six ratchets. Each fix was mutation-checked — reverting it turns a named test red. Worth knowing: the `app_shell` sites were covered by nothing when I first changed them (reverting them left all 328 router tests green), which is why `app_shell_own_profile_encoding_test` exists.

Left alone on purpose, both noted in the commit bodies:

- Deep-link canonicalization. It changes what URL the app mints from external input, and it is redundant once the comparisons normalize.
- A test for `profile_screen_router`'s call site. That screen has no working harness — `profile_screen_router_test.dart` is `skip: true` throughout and `profile_share_edit_buttons_test.dart` asserts inside `if (finder.evaluate().isNotEmpty)`, so it cannot fail. It still had to change with the shell, or the own-profile body renders with neither its scaffold nor the shell app bar.

One more for the record: `isOtherUserProfile` at `app_shell.dart:241` has no observable delta — it only feeds `showBackButton`, which lives inside `DiVineAppBar`, which `isOwnProfileGrid` has already suppressed wherever the two comparisons differ. Changed for consistency, not behaviour.

Yours to merge — happy to drop any of the four if you'd rather take them separately.

hm21 (approved):
LGTM. Pushed a768553859: routeIdentifiesUser compares hex case-insensitively but the profile body compares raw, so an uppercase-hex deep link got the own-profile scaffold with the stranger body inside — now normalizeToHex lowercases at the choke point. Revert independently if unwanted.

PR #6661, opened 2026-08-04
feat(lists): publish private lists to relays with NIP-44 encrypted items

### LINKED ISSUE #6864: Sync private lists across devices with encrypted items
## Problem

Private curated lists currently live only in local storage, so they cannot be restored after reinstall or synced to another device.

## Desired behavior

- Publish every owned list as a NIP-51 kind 30005 event.
- For private lists, encrypt the item `e`/`a` tags to the owner with NIP-44 and keep them out of public tags.
- Fetch and decrypt the owner's private-list items during relay sync.
- Preserve public NIP-51 metadata interoperability; list existence, coordinate, name, and other metadata remain relay-visible.
- Back up eligible existing local private lists after the initial relay merge.
- Require relay acceptance for a public/private visibility flip and do not leak item tags if encryption fails.
- Cover publishing, relay round trips, failed encryption, visibility changes, deletion, and backfill with regression tests.

## Product decision

Publishing the encrypted default list and eligible existing local lists is intentional. Relay visibility of the event envelope and public metadata is accepted; private-list membership must remain encrypted.

rabble:
Closes #6864.

## Summary

A private list was never sent anywhere. It sat in `SharedPreferences` and died with the device: no sync to a second phone, nothing to restore after an uninstall. It bought no privacy either, since nothing was published to hide from.

Every owned list now publishes. A private one sends the same kind 30005 metadata tags as a public one and NIP-44 encrypts its item tags to the owner in the event content.

```
kind: 30005
tags: [["d","list_1234"], ["title","puppets"]]      <- public
content: nip44(self, '[["e","<id1>"],["e","<id2>"]]')  <- items sealed
```

Relays and other users can see that the list exists and what it is called. What is in it is opaque. That is standard NIP-51, so other clients understand the shape.

## Motivation

Reported directly: private lists were not reaching the server at all. Follow-up to #6617, where the make-private path was still a NIP-09 deletion and the UI still labelled private lists "On this device".

## What changes

- **Make-private is a republish, not a deletion.** Kind 30005 is addressable, so the sealed event replaces the public one under the same d-tag. A kind 5 targets that coordinate, so it would have asked relays to drop the replacement along with the original.
- **The visibility flip still waits on relay acceptance.** A rejected republish leaves the list public, because relays still hold the copy with items in plain tags.
- **Encryption failure blocks the publish.** A private list that leaked its items would be worse than one that never left the phone.
- **Deletion covers private lists**, gated on `nostrEventId` rather than `isPublic`.
- **Item changes republish at any visibility** — the items are in the event either way.
- **Sync unseals our own events only.** Someone else's list is not ours to decrypt, and a failure to unseal means a list published before this change, whose content is a plain description.
- **Existing local-only lists are backed up** after the first sync merge. This is what carries an install's current private lists across.
- **Copy.** `listVisibilityPrivateDevice` -> `listVisibilityPrivate` ("Private"), and the make-private warning now says the videos get encrypted, the name does not, and copies already shared can persist. 21 locales.

## What stays public

The `d` tag, title, description, image, hashtags and play order. Encrypting the title too was considered and rejected: other clients would show an unnamed list, and our own share and explore paths would need a fallback. Worth revisiting if a list name itself is ever sensitive.

Private lists do not appear in Explore for other users — the public discovery path already rejects lists with no readable item tags (`curated_list_service.dart:1487`).

## No visual change

Two strings change: the visibility chip reads "Private" instead of "Private · On this device", and the make-private confirmation text is rewritten. No layout change.

## Validation

- `flutter analyze lib test` clean
- 330 tests across the list service suites, l10n, list/profile widgets and the list screens
- 108 tests in `curated_list_repository`, 837 in `models`
- Pre-commit format gate green

New coverage: sealed-item publishing and the encryption-failure refusal; make-private as a republish with no kind 5; deletion of a private list; sealed round trip through relay sync; no decrypt attempt on another user's list; a legacy plain-description list surviving a failed unseal; backfill of a list that never reached a relay, and no republish of one the relay already has.

## Not covered

The translations are machine-quality and worth a human pass. Same caveat as the l10n commit in #6617.


rabble:
## Review — feat(lists): publish private lists to relays with NIP-44 encrypted items

(Left as a comment rather than a formal review since GitHub won't take a review from the PR author.)

### Overview

Turns "private list" from local-only-and-lossy into a real NIP-51 private list: metadata tags stay public, item `e`/`a` tags get JSON-encoded and NIP-44-sealed to the owner into `content`. Every owned list publishes now, edits republish under the same d-tag (correctly rejecting NIP-09 for the flip, since a kind-5 on the coordinate would take the replacement with it), `confirmed:` splits durable-publish semantics from queue-and-replay, and `_backfillUnpublishedLists` rescues lists stranded by the old behaviour. `_sealItemTags` encrypting to the owner's own pubkey is the right NIP-51 shape, and the `copyWith` `clear*` flags are the right fix for the null-means-unchanged trap.

The direction is right. Findings below, worst first — the first three I'd treat as blocking.

---

### 1. A transient decrypt failure silently downgrades a private list to public-and-empty

`_unsealItemTags` returns `null` for four very different situations:

```dart
if (event.content.isEmpty) return null;                       // not sealed
if (ownerPubkey == null || event.pubkey != ownerPubkey) return null;  // not ours
if (plaintext == null) return null;                           // signer said no
} on Object catch (e) { ... return null; }                    // decrypt threw
```

Only the first two mean "this is a public list". The last two mean "this is a private list I couldn't open" — signer not ready during cold start, a NIP-46 remote signer timing out, key not yet loaded. `fromEvent` can't tell them apart, so it produces `isPublic: true`, `videoEventIds: []`, and `description: <ciphertext>`.

`_processListEvent` then merges on newest-wins:

```dart
if (event.createdAt > existingList.updatedAt.millisecondsSinceEpoch ~/ 1000) {
  _lists[existingListIndex] = curatedList.copyWith(createdAt: existingList.createdAt);
}
```

The relay copy is by construction newer than the local list it was published from, so the local private list is overwritten with an empty public one. From there the next `addVideoToList` / `updateList` republishes it through the `list.isPublic` branch — plain `e`/`a` tags, real leak. And the user's items are gone from the local copy in the meantime.

The failure needs to be distinguishable from the success. One structural discriminator that doesn't need a new field: an event we authored is private iff it has no `e`/`a` tags and non-empty content. Return a tri-state (`sealed` / `notSealed` / `unsealFailed`) and have `_processListEvent` skip the merge entirely on `unsealFailed` rather than parse it as public. Worth a regression test that stubs `nip44Decrypt` to return null and asserts the local private list survives untouched.

### 2. Someone else's private list renders its ciphertext as the description

Correctly, we return `null` from `_unsealItemTags` when `event.pubkey != ownerPubkey`. But `fromEvent` with `privateTags: null` still does:

```dart
description: description ?? (isPrivate ? null : event.content),
final contentFirstLine = isPrivate ? '' : event.content.split('\n').first;
```

`isPrivate` is false on that path, so a third-party private list gets its base64 NIP-44 blob as the description — and if it has no `title` tag, as the *name*. That reaches the explore lists tab, list cards, and the curated-list feed screen.

The "is this content readable prose?" question is answerable without knowing who authored it: metadata tags present, no item tags, non-empty content. Suggest deriving `isPrivate` in `fromEvent` from that shape rather than from `privateTags != null`, and keeping `privateTags` purely as the items channel.

### 3. `_backfillUnpublishedLists` can republish another user's list under our pubkey

```dart
final stranded = _lists.where((list) => list.nostrEventId == null).toList(growable: false);
```

`_lists` isn't only ours — `subscribeToList` adds the fetched foreign list into `_lists` (~line 918), and `subscribedLists` is just a view over it filtered by `_subscribedListIds`. Any subscribed list that lands without an event id gets signed and published under our key at its own d-tag. The docstring says "owned lists"; the filter doesn't enforce it. One line:

```dart
final owner = _currentAuthenticatedPubkey();
final stranded = _lists.where((l) => l.nostrEventId == null && l.pubkey == owner)
```

---

### 4. `listPrivateListSubtitle` still describes the behaviour this PR deletes

`"Private lists stay on this device and aren't backed up"` — added in this PR, in `app_en.arb` and all 21 other locales, and it says the opposite of what the PR now does and of what the new `CuratedList.isPublic` docstring says. Users toggling private will read that their list isn't backed up while the code is busy publishing it. Something like "Only you can see what's in this list" fits both the behaviour and the tone. Needs re-translating across the 22 files.

(The rest of the ARB work is good — real translations, not English copies, placeholders declared, all locales mirrored.)

### 5. `createList` rolls back offline, which cancels out the backfill it ships with

`_backfillUnpublishedLists`'s docstring: *"the same is true of any list created while the app could not reach a relay. A null `nostrEventId` is how both look."* But `createList` now does:

```dart
if (!await _publishListToNostr(newList, confirmed: true)) {
  _lists.removeWhere((list) => list.id == listId);
  await _saveLists();
  return null;
}
```

so a list created offline is deleted, never reaching the null-`nostrEventId` state the backfill is written to rescue. Net effect for a user on a plane: list creation just fails, with no local fallback — a regression against the old public-list path, which didn't block on the relay at all.

Pick one story: either keep the list locally with a null event id and let backfill retry (which is what the backfill doc describes and what "local-first" would suggest), or drop the offline-creation rationale from the backfill doc. The current pair does neither.

### 6. `acceptedByAny` is the right bar for durability and the wrong one for a flip to private

For a normal publish, one acceptance is fine — that's argued well in the comment. For the public→private transition it isn't: the replacement event is the entire mechanism that withdraws the plaintext items. If one relay accepts and three don't, those three keep serving the old `e`/`a` tags indefinitely, and the user has been told the list is now private.

Two things worth doing:

- On the flip specifically, either require the full connected write set or surface partial acceptance in the confirm dialog's outcome.
- The `listMakePrivateWarning` copy says "copies shared before now may persist", which gestures at this, but the sharper truth is that relays retaining superseded addressable versions will serve the old plaintext regardless of how many accept. Worth saying plainly in the `CuratedList.isPublic` docstring so the privacy claim doesn't outrun what the protocol delivers.

### 7. Private + collaborative is unguarded

Items are sealed to the owner alone, so flipping a collaborative list private leaves `allowedCollaborators` advertised in the tags while no collaborator can read a thing. Either block the combination in `updateList` or seal to each collaborator's pubkey as well (NIP-44 is pairwise, so that's N ciphertexts — probably a reason to block it for now and say so).

---

### Nits

- `_backfillUnpublishedLists` runs N sequential `confirmed: true` publishes inside `fetchUserListsFromRelays`. A user with a dozen stranded lists blocks the startup sync on a dozen serial relay round trips. Bounded concurrency, or move it off the sync path.
- `_unsealItemTags` logs the decrypt failure at `Log.debug`. Given finding 1, a genuine unseal failure is the thing you'd most want in a bug report — warning level, and only for the "should have been decryptable" case once the tri-state exists.
- `_publishListToNostr`'s docstring cites `relay_pool.dart:1624` for the queue-on-failure behaviour. Line references rot; naming `queueIfFailed` alone (which it also does) is enough.
- `updateList`'s `clearDescription: description != null && description.isEmpty` alongside `description: description ?? list.description` is correct but reads as though the two could disagree. `description?.isEmpty ?? false` says the same thing in one clause.

Full event IDs and pubkeys throughout, no key material in logs, `test/unit/curated_list_relay_sync_test.dart` is modified rather than added so the frozen-bucket ratchet is fine.

NotThatKindOfDrLiz (changes requested):
I pushed commit 346a2e5dc with the mechanical fixes I think are safe for reviewer takeover: private-list unseal failures now skip the merge instead of downgrading to public, sealed-looking third-party content no longer becomes user-facing copy, backfill is owner-scoped, the private subtitle/l10n are updated, and the stale widget assertion is fixed.

Requesting changes for the remaining product/privacy decisions before merge:

1. Pick the offline-create story. _backfillUnpublishedLists still describes rescuing lists created while relays are unreachable, but createList rolls back authenticated create failures. Either keep offline-created lists locally with nostrEventId == null so backfill can do that job, or narrow the doc/behavior so create failure is intentionally not local-first.

2. Pick the privacy bar for public -> private flips. acceptedByAny means one relay can accept the sealed replacement while other relays keep serving the old plaintext e/a tags. That may be acceptable, but the PR needs an explicit product/protocol stance: require a stronger write set for the visibility flip, surface partial acceptance, or document plainly that rejected/stale relays may continue serving plaintext items.

3. Decide the signed-out visibility invariant. updateList comments say visibility only commits after relay confirmation, but the signed-out path can persist a visibility change locally with no relay change. Either make signed-out visibility changes fail/leave visibility unchanged, or update the invariant and user-facing semantics.

Also please refresh the stale PR body and link an issue or Closes #none so needs-issue clears.

NotThatKindOfDrLiz:
Update after rebasing onto current main: the reviewer-takeover fixes are now in 1eefa0637, with a small follow-up 3243c12d5 to adapt the PR to the current publishEventAwaitOk/PublishOutcome API. The requested product/privacy decisions in my changes-requested review still stand.

mbradley (changes requested):
Reviewed at `3243c12`. This is a delta pass, so it skips everything already open: Liz's three standing items, her takeover fixes, and your own seven self-review findings. All of that is dropped as duplicate rather than restated.

The approach is right and the NIP-51 shape is correct. Private lists genuinely needed to stop dying with the device, and self-encrypted item tags are the correct mechanism. What follows is the set of things that would reach users if this merged as-is.

### In plain language

**Every user gets published, under the same name.** The app auto-creates a "My List" for every account, marked private. Because private lists now publish, every install announces itself to public relays under an identical, predictable d-tag. One query returns every Divine user. Your own comment on that line says the default list was made private specifically so this wouldn't happen.

**Two devices, one list, one loser.** Private lists never synced, so a user with a phone and a tablet has two different versions of the same list under the same d-tag. There's no union merge, only newest-wins replace. Whichever device syncs last wins and the other's videos are gone, with no backup, because not having one was the premise. The feature exists to rescue those lists and for multi-device users it deletes half of them.

**Turning a list private has a window where it leaks.** The flip waits up to 15 seconds for relay confirmation, and by design the list stays public locally until then. A video added during that wait publishes with plain item tags, and because it's newer, that's the copy relays keep.

**The name isn't private, and only one screen says so.** The make-private dialog discloses it. The create-a-private-list screen says "Only you can see what's in this list" and stops there. Existing private lists also publish on the next sync with no prompt, after the app told users they stay on the device.

### Receipts

1. **Default list.** `curated_list_service.dart:93` fixes `defaultListId = 'my_vine_list'`. `:1092` creates it for every authenticated user via `initialize()` at `:149`. `:1095` carries the comment `// Don't spam the relay with empty default lists`, added in `ce6c0b0ce` (Jan 2026). `_createList` now publishes whenever authenticated (`:285`), and `_backfillUnpublishedLists` (`:1847`) covers existing installs. Filter `{"kinds":[30005],"#d":["my_vine_list"]}` enumerates the userbase.
2. **Multi-device replace.** `_processListEvent:1968` does `curatedList.copyWith(createdAt: existingList.createdAt)` when the relay event is newer. Wholesale replace of `videoEventIds`, no union. Combined with 1, every multi-device user has a guaranteed d-tag collision on divergent, never-synced content.
3. **Stale-copy overwrite.** `fetchUserListsFromRelays` completes on a 10s timer (`:1391`) or a relay error (`:1417`), processes whatever arrived, sets `_hasSyncedWithRelays = true` (`:1443`), then runs the backfill (`:1450`). Nothing distinguishes EOSE from timeout. The comment at `:1450` claims the ordering stops a stale local copy republishing over the relay; it only holds if the query was complete. #6238 documents that it silently isn't.
4. **Flip window.** `updateList:483` writes `updatedList.copyWith(isPublic: list.isPublic)` before the confirmed publish, per #6616's requirement to preserve visibility on failure. `addVideoToList:340` then reads that still-public entry and calls `_publishListToNostr`, which takes the `list.isPublic` branch at `:1167` and emits `toEventTags` with plain `e`/`a` tags.
5. **Sign-out downgrades a private list.** `_processListEvent:1941` passes `isPrivateEvent: status == unsealed`, forcing `false` on the `notSealed` branch. `_unsealItemTags:1885` returns `notSealed` when `_currentAuthenticatedPubkey()` is null, reachable during the 10s window. The list merges as public and empty with ciphertext as its description, and the next edit republishes it with plain tags. The tri-state guards decrypt failure; this is the door beside it.
6. **Test sentinel in production.** `:1919` is `if (content.startsWith('sealed:')) return true; // Test cipher stand-in.`, added in `1eefa0637` to reach `sealForTest` in `test/helpers/curated_list_publish_stubs.dart:30`. Deleting the line reddens two tests, which means the real base64 heuristic is never exercised by the service. A user description beginning `sealed:` routes a real list to `nip44Decrypt`, fails, and `:1936` drops the event, so the list silently stops syncing. The heuristic is also duplicated in `curated_list_converter.dart:133` and the copies have diverged by exactly this line.
7. **Untestable guards.** Dropping `event.pubkey != ownerPubkey` from `:1889` leaves all 68 tests green. The test named for that guard uses a fixture with `content: ''` and an `e` tag, so it exits at the first branch and its `verifyNever` holds regardless. Moving `_backfillUnpublishedLists()` from after the merge to before it also leaves everything green.
8. **Key binding.** `_sealItemTags:1117` encrypts to `_authService.currentPublicKeyHex` while the private key comes from `_nostrService.signer`. Nothing asserts they match. `auth_service.dart:400` resolves that getter through a different chain than the signer. `await signer.getPublicKey()` removes the cross-source binding.
9. **Item ceiling.** `nip44_v2.dart:70` caps plaintext at 65535 bytes. Each `["e","<64 hex>"]` costs 73 bytes, so sealing fails around 890 items. `_publishListToNostr` correctly returns `false`, but `:349`, `:398`, `:719`, `:797` and `:841` all ignore it. The local list keeps growing, the relay copy freezes, and nothing tells the user.
10. **Metadata surface.** `toPublicMetadataTags` (`curated_list_converter.dart:180`) publishes `thumbnail`, which per `curated_list.dart:156` is one of the list's own videos, in the clear. `:172` publishes collaborator pubkeys. `_generateListId:236` makes the d-tag `list_<millisecondsSinceEpoch>`, so the coordinate discloses creation time to the millisecond.

Two of these have an agreed direction already. Daniel's #6750 describes the same read-modify-write hazard in the mute list and states the rule: if we could not confirm the current version, do not publish over it. That covers 3 directly, and his #6751 notes this PR is the first NIP-51 encrypted-content implementation in the repo, so whatever it does here is the precedent the mute list would follow.

### Questions I can't answer from the code

1. Should the default list publish at all? Exempting it until it has content preserves the January intent and removes the enumeration surface, at the cost of the default list not syncing across devices. Publishing it accepts the census. Which?
2. What should happen when two devices hold divergent copies of the same never-synced private list? Union the items, prompt the user, or keep last-writer-wins and document it?
3. Should existing local-only private lists publish silently, or behind a one-time prompt? They were created under copy promising they stay on the device, and relays don't forget.
4. Is NIP-04 back-compat in scope? NIP-51 prescribes detecting it by searching for `iv`, and this repo's own `nip51/follow_set.dart` still writes it. If it's out of scope, worth saying so in the docstring rather than leaving the heuristic silently rejecting those events.

I can't push fixes here: `PR_REVIEW.md` bars reviewer takeover in security-sensitive code without separate authorization, and this is the privacy path end to end. Happy to attach patches for the objective items (5, 6, 7, 8, 10) if that's useful.

Thanks, as always.

rabble:
Owner decision on the private-list publication model:

Publishing private lists to relays and fetching/decrypting them on another device is the purpose of this PR. That includes the default list and existing local private lists eligible for backfill. “Private” protects the item membership: the `e`/`a` references are NIP-44-encrypted in event content. The event envelope and the public NIP-51 metadata remain relay-visible, as the current make-private warning discloses.

Accordingly, discoverability of a published `kind:30005` event or its stable `d` tag is an accepted protocol/product property here, not evidence that the private list contents are public and not a reason to exempt the default list from cross-device sync.

Please re-review the current head and separate any remaining concrete correctness/data-loss bugs from that now-decided publication model.

rabble:
Review fixes are pushed at `ab8a4a339`.

Liz:

1. Offline create is local-first again. A confirmed signing, encryption, or publish failure keeps the list with a null event id, and a later complete relay sync backfills it.
2. Public-to-private now requires acceptance from every configured relay. Partial acceptance leaves the local list public, and the model docs explicitly acknowledge that stale or rejecting relays may retain an older plaintext event.
3. Signed-out visibility changes fail atomically and leave every edit unchanged.
4. The PR body is current and closes #6864.

Matt:

1. Publishing the default list and legacy private lists is intentional per the owner decision above. Private means encrypted item membership; the envelope and public NIP-51 metadata remain visible.
2. A local unpublished list colliding with its relay coordinate now unions both item sets, deduplicates them, preserves private if either side is private, and republishes only after a complete sync.
3. Relay timeout or error is now incomplete: partial events may merge, but the session is not marked synced and no backfill occurs, so retry remains possible.
4. Mutations for one list are serialized, closing the visibility-flip/add race.
5. Valid sealed events are skipped when signed out or owned by another pubkey; they cannot downgrade into public lists.
6. The production `sealed:` sentinel and broad base64 heuristic are gone. Classification decodes the exact NIP-44 v2 envelope.
7. Regression tests now actually exercise the foreign-owner and backfill-order guards.
8. Encryption and decryption verify that the signer pubkey matches the authenticated owner.
9. Oversized NIP-44 payloads are rejected before local mutation. Other unconfirmed mutation publishes keep the user edit but clear the event id for safe later backfill instead of silently freezing the relay copy.
10. Private metadata omits item-derived thumbnails. Private collaboration is rejected because the ciphertext is owner-only. Collaborator and coordinate metadata remain public under the accepted publication model.
11. Legacy NIP-04 encrypted content is detected and decrypted for read compatibility.

Additional regression coverage covers all of the above, including divergent-device union, timeout retry, signer mismatch, exact envelope classification, NIP-04 reads, relay write thresholds, atomic visibility behavior, and operation serialization.

Local verification after rebasing onto fresh `origin/main`: analyzer clean, ARB consistency 8/8, and all 337 pre-push selected tests passed.

Please re-review the current head. Auto-merge remains armed once approvals and CI are green.

mbradley (dismissed):
Re-reviewed at `ab8a4a3`. The fix commit lands: I verified the `sealed:` sentinel is gone and replaced by exact NIP-44 envelope decoding, the signer/owner pubkey binding, the payload ceiling, the private+collaborative rejection, the signed-out visibility rejection, the thumbnail omission, and the sign-out downgrade guard — each with a test that goes red when the guard is removed. NIP-04 read compatibility is there and tested.

Two of the new fixes introduce fresh data-path defects, and both are the correctness side of the union-merge and all-relays work rather than the product model. I've split what needs a decision from what has a clear patch. Suggested patches are attached to the mechanical ones; I can't push to this branch (`PR_REVIEW.md` bars reviewer takeover on the privacy/encryption path without separate sign-off), so they're suggestions for you to take or leave.

## Decisions (these block, they're yours)

**1. The union merge can't distinguish a deletion from a divergence.** `_commitListMutation` clears `nostrEventId` when a publish fails (`curated_list_service.dart:369`), and a null id is exactly the predicate that routes the next sync into the union branch (`:2130`). So a video the user *removed*, on a list whose republish failed, gets unioned back from the stale relay copy and re-published by backfill — the item returns, appended out of order, with no signal. "Emptied the list" restores in full the same way. The branch treats "this device never had X" and "this device deleted X" as the same state.

Concrete options:
- Add a `pendingRepublish` flag distinct from `nostrEventId == null`, so backfill retries a failed edit while the union branch fires only for a genuinely never-published list. Keeps both behaviours you want, at the cost of a model field.
- Make the merge additive only from the side that is strictly newer, and treat the older side as authoritative for removals (i.e. a union is only safe on a true first sync, not on a failed-edit copy).
- Keep last-writer-wins for the failed-edit case and reserve union strictly for the cross-device first-sync case.

Which reconciliation model do you want? The current one loses user deletions.

**2. `requireAllRelays` trades one leak for another.** `acceptedByAll` is false when any targeted relay rejected, went silent, or was unreachable — which on mobile is the common case, per the relay-pool's own note. When the flip "fails" this way, `_updateList` has already committed the list as still-public locally (`:601`), while a relay that *did* accept holds the private copy. The next item add then republishes plaintext over that relay. `acceptedByAny` had the mirror problem Liz raised (stale relays keep plaintext). Neither bar is clean.

Options:
- If any relay accepted the private replacement, commit local visibility to private and surface partial acceptance, rather than treating a partial as "still fully public."
- Block item mutations while a visibility flip is pending-partial, so no plaintext edit can land on a relay that already took the private version.
- Keep `acceptedByAll` and document that a partial flip leaves the list public until it fully succeeds.

Your call on which leak is acceptable and how the flip reports partial acceptance.

**3. Relay sync mutates `_lists` outside the per-list serializer.** `_processReceivedListEvents` / `_processListEvent` take no lane (`:2001`, `:2095`), so a sync-merge can interleave with an in-flight serialized mutation — e.g. a confirmed publish is awaiting (up to 15s) while a concurrent sync rewrites the same entry, and the publish's write-back then stamps an event id onto a list the sync has since changed. There's still no re-entrancy guard on `fetchUserListsFromRelays` either (`_hasSyncedWithRelays` is set only after the whole window, and `force: true` bypasses it), so two syncs can run at once. Making sync participate in the same serialization is a concurrency-model choice (whole-sync operation vs per-list, and how a batch orders against N lanes) — I didn't want to pick it for you. How should a batch sync coordinate with the per-list operation lanes?

**4. The published default list can't be withdrawn.** Under the accepted model, `my_vine_list` publishes for every user, but both delete paths refuse `defaultListId` (`:653`, `:700`). A user who wants it gone has no path. Given the decision to publish it, should deletion of the default list be allowed?

**5. Create-toggle disclosure.** The make-private confirmation says the name stays visible; the create-a-private-list toggle still reads only "Only you can see what's in this list" (`listPrivateListSubtitle`). The owner decision on #6864 rests on metadata visibility being disclosed, and it is — on the flip, not on create. Copy is yours/design's; flagging that the create path doesn't carry the same disclosure.

## Patches (objective — take or leave)

**A. Merge can produce private+collaborative, which both other paths refuse.** `:2155` ANDs `isPublic` but carries `isCollaborative` from `preferred` unchecked. Mirror the `_updateList` rule (a collaborative list can't go private):

```dart
_lists[existingListIndex] = preferred.copyWith(
  videoEventIds: mergedVideoIds,
  createdAt: existingList.createdAt,
  updatedAt: DateTime.now(),
  isPublic: (existingList.isPublic && curatedList.isPublic) ||
      preferred.isCollaborative,
  clearNostrEventId: true,
);
```

**B. `deleteOwnedList` isn't serialized.** A confirmed deletion awaits up to 15s; an item add on the same list during that window publishes a fresh kind 30005 with a newer `created_at`, so the kind 5 doesn't cover it and the list is republished on relays after the user deleted it. Route it through the lane like the other mutators:

```dart
Future<bool> deleteOwnedList(String listId) =>
    _serializeListOperation(listId, () => _deleteOwnedList(listId));

Future<bool> _deleteOwnedList(String listId) async {
  // existing body
}
```

**C. `createList`'s confirmed publish is outside the lane.** A sync completing during its publish window runs backfill, takes the lane `createList` never held, and publishes a second event for the same coordinate; the two race the write-back. Hold the lane across the create-and-publish, keyed on the generated `listId`.

**D. No re-entrancy guard on `fetchUserListsFromRelays`.** Coalesce concurrent callers onto one in-flight future (set before the first await, cleared in a `finally`), and have `force` await the running sync rather than start a parallel one.

**E. Backfill doesn't re-check ownership across awaits (`:2019`).** It captures `owner` once and awaits up to 15s per list; an account switch mid-loop publishes account A's stranded public list under B's signer. Re-read the pubkey inside the loop and abort the iteration if it changed — the same pattern `_sealItemTags` already uses.

**F. Null-pubkey lists are never backed up and skip the union guard.** `createList` sets `pubkey` from `_currentAuthenticatedPubkey()`, which is null when signed out; such a list stays in `myLists` but is filtered out of backfill (`:2023`) and fails the union predicate `existingList.pubkey == event.pubkey` (`:2130`), so it falls through to wholesale replace. Stamp the authenticated owner onto a null-pubkey local list when auth resolves, before backfill. (Assumes a list created on this device belongs to this device's user, which is how `myLists` already treats it — flag if that's wrong.)

**G. The backfill re-resolve fix has no test.** Both halves — re-resolving by id and the `nostrEventId != null` re-check — can be reverted with the suite still green. Worth a regression test that a stranded list edited to non-stranded mid-backfill isn't double-published.

**H. `deleteOwnedList` docstring contradicts its body (`:696`).** It still says "Private lists are local-only and can be removed without relay publication"; the body 20 lines down publishes a NIP-09 deletion for any list with a non-null `nostrEventId`. Update the doc.

**I. `deleteList` is dead code (`:651`).** No production callers — the UI path is `deleteOwnedList`. It removes locally and publishes nothing, on the reasoning that a replaceable event is "automatically replaced," which never happens if nothing republishes; wiring it up would orphan a published list on relays. Suggest removing it (and its test group). Flag if it's intended future API.

Also: `107be2aa9` (an uploads-test flake fix) is bundled in and isn't private-lists work — fine to keep, just noting it for a clean squash message.

Thanks, as always.

mobile/lib/services/curated_list_service.dart
```diff
@@ -470,36 +603,32 @@ class CuratedListService extends ChangeNotifier {
...
+      // replaces whatever the relays hold — including a public copy whose
+      // items were in plain tags. That is what makes a list private, and it
+      // is why this is a republish rather than a NIP-09 deletion: a kind 5
+      // targets the coordinate, so it would ask relays to drop the private
+      // replacement along with the public original.
+      if (_authService.isAuthenticated &&
+          !await _publishListToNostr(
+            updatedList,
+            confirmed: true,
+            requireAllRelays: list.isPublic && !updatedList.isPublic,
```
NotThatKindOfDrLiz: This still needs an owner decision. With `requireAllRelays`, a public-to-private publish accepted by some relays but rejected or timed out on one relay leaves local state public while at least one relay already has the encrypted replacement. The next normal item mutation can then publish plaintext again over a relay that had accepted the private copy. Either treat any accepted encrypted replacement as a pending private transition, block subsequent plaintext mutation until the partial state resolves, or explicitly decide that partial privacy flips stay public and document that tradeoff in code/tests.

mobile/lib/services/curated_list_service.dart
```diff
@@ -520,55 +649,16 @@ class CuratedListService extends ChangeNotifier {
...
+  /// Published lists send a NIP-09 deletion event for the kind 30005 coordinate
+  /// before local state is removed. A list that never published has no event to
+  /// delete and can be removed locally.
+  Future<bool> deleteOwnedList(String listId) {
+    return _serializeListOperation(listId, () => _deleteOwnedList(listId));
+  }
+
+  Future<bool> _deleteOwnedList(String listId) async {
     try {
       if (listId == defaultListId) {
```
NotThatKindOfDrLiz: Under this PR, the default list can now be published to relays, but this guard still makes it impossible to withdraw. That was fine while private/default lists were local-only; with publish-everything semantics, the owner needs a supported way to unpublish the default list coordinate or an explicit product decision that the default list is permanent once published.

mobile/lib/l10n/app_en.arb
```diff
@@ -3037,9 +3037,9 @@
   "listDescriptionLabel": "Description (optional)",
   "listPublicList": "Public List",
   "listPublicListSubtitle": "Others can follow and see this list",
-  "listPrivateListSubtitle": "Private lists stay on this device and aren't backed up",
+  "listPrivateListSubtitle": "Only you can see what's in this list",
```
NotThatKindOfDrLiz: The create toggle copy no longer tells the user that list name/metadata remain visible. The later make-private warning has that caveat, but create is the first place users choose privacy, and this sentence reads as full-list invisibility. Please align the create copy with the actual NIP-44 behavior: items are private, but public metadata such as the list name remains visible.

NotThatKindOfDrLiz (changes requested):
I pushed a mechanical fix commit for the objective failures from the last round: failed private-list edits now use a pendingRepublish bit instead of overloading nostrEventId == null, failed deletions no longer union stale relay items back into local state, private+collaborative merge output stays public, null-pubkey owned lists are backfilled under the authenticated pubkey, create/delete now run through the list operation lane, the dead deleteList path is gone, and the stale deleteOwnedList docstring is fixed.

I am still requesting changes because three product/concurrency decisions remain unresolved before this should merge: partial relay acceptance during public-to-private flips, default-list withdrawal under the new publish-everything model, and create-flow privacy copy that now omits the visible-name/metadata caveat. The sync re-entrancy model also still needs an explicit choice; the current once-per-session flag is set after the await window and force:true can still run overlapping syncs.

Please also leave auto-merge off until those are answered; the branch is mergeable now, so a single approval would otherwise merge with those semantics still undecided.

mbradley (commented):
Took over with two commits fixing concrete correctness bugs on the
current head. Both are fenced away from the encryption/visibility
behavior — no change to sealing, the flip acceptance bar, or what a
private list publishes. The open product and privacy-model decisions
from the earlier rounds are untouched and remain yours.

Pushed:

1. `fix(lists): retry a failed metadata edit instead of stranding it`
   `_commitListMutation` already retries a failed item edit via backfill
   (clear event id + `pendingRepublish`); `_updateList` did not. A
   rename/description/tags/order edit whose publish failed kept its old
   event id, so backfill and the sync union-merge both skipped it: the
   edit never reached relays, and a newer relay copy could later
   overwrite it. The fix mirrors that retry, gated on `!visibilityChanged`
   — a failed public-to-private flip is deliberately not routed through
   backfill (which republishes at accepted-by-any and would defeat the
   all-relays bar the flip needs), so it stays at its old visibility for
   the user to retry. Regression test confirmed to fail without the fix.

2. `fix(lists): serialize a new list's publish on its own operation lane`
   `createList` serialized on a constant lane while
   `_backfillUnpublishedLists` serializes on the real list id, so a list
   created during the startup background sync could be published twice:
   the create's confirmed publish in flight, the sync completes, and
   backfill republishes the same coordinate. This completes the
   per-list-lane design already in place — create was the one mutation
   not participating. The create's publish now runs on the list's own
   lane with the same `nostrEventId` re-check backfill uses; id
   generation stays inside the serialized create, so list-id uniqueness
   is unchanged. Regression test confirmed to fail (two publishes)
   without the fix.

Validation: `flutter analyze` clean, `dart format` clean, and the full
curated-list service suites green (crud, video management, relay sync,
collaboration, initialization, persistence, playlist, query, stream,
fetch-public). CI green on the pushed commits.

Left to you: @NotThatKindOfDrLiz's three open threads (partial relay
acceptance on the flip, default-list withdrawal, create-flow copy) and
the sync re-entrancy model are product and privacy-model calls, not mine
to resolve, so I have not touched them. Her changes-requested review
stays as the gate on this PR — I am not clearing or duplicating it. Her
takeover fixes and the prior round's fixes all verified in place. The
merge decision stays yours.

Thanks, as always.

mobile/lib/services/curated_list_service.dart
rabble: Answered in `1a31552` (rebased to `9f8ef79`), with an owner decision that changes the shape of this rather than picking one of the three options.

**Decision: a public-to-private flip now publishes a NIP-09 redaction of the plaintext event, and commits at accepted-by-any.**

The premise both the code comment and this thread rested on was that a deletion could not help here — a kind 5 targets the coordinate, so it would take the sealed replacement down with the original. That is true of an **`a` tag**: NIP-09 has relays honour a coordinate deletion "for every version of the replaceable event up to the `created_at` of the deletion request." It is not true of an **`e` tag**, which names one event id. The sealed replacement has a different id and survives.

So the flip now captures `list.nostrEventId` before `_publishListToNostr` overwrites it, and after a successful flip publishes `["e", <old plaintext id>], ["k", "30005"]`. The relays that accepted the sealed replacement already have it; the ones that did not are asked to drop the plaintext. **The worst case at a non-accepting relay becomes absence rather than plaintext.**

That is what lets `requireAllRelays` go. Your scenario — local reads public while some relay holds the sealed copy, and the next item mutation republishes plaintext over it — is now structurally impossible rather than mitigated:

- no relay accepted → nothing sealed exists anywhere → staying public is honest, and no redaction is published (there is nothing to retire)
- some relay accepted → the list commits to private locally → no later edit can publish plain item tags

I want to be explicit that this reverses a bar you and @mbradley both re-reviewed, and why. The all-relays bar could not actually prevent the leak it was aimed at: nothing in `NostrClient` republishes to relays that did not accept, so a partial publish stayed partial — the guard converted "one relay still serves plaintext" into "the flip fails and *every* relay still serves plaintext," while one wedged pool member blocked the flip permanently. `nostr_client.dart:733` says as much directly: *"Nothing in this client republishes to relays that did not accept, so a partial publish stays partial. That makes `acceptedByAll` a poor gate for anything the user is waiting on... Gate progress on `acceptedByAny`."*

**Three limits, documented in the code rather than papered over:**

1. **Partial stays partial.** A relay that rejects both the replacement and the redaction keeps the plaintext. Nothing here retries non-accepting relays; building that retry is a larger change than this PR should carry.
2. **NIP-09 is SHOULD, not MUST**, and anything that already read the public copy keeps it. The redaction reduces exposure; it does not guarantee removal. Product copy must say "private going forward," never "deleted."
3. **Only the newest plaintext id is known.** Replaceable semantics mean relays normally hold one version per coordinate, so one `e` tag covers it, but a relay sitting on a stale older version is not reached. The service does not retain id history.

The redaction is best effort by contract and never gates the flip — it runs after the flip has already committed.

Tests: the redaction carries the pre-flip id and no `a` tag; partial acceptance now commits privacy and redacts (this inverts `keeps a list public when only some relays accept privacy`); full rejection stays public and publishes no redaction; a failed redaction does not fail the flip. Verified the redaction test goes red with the call removed.

Your other two threads (default-list withdrawal, create-flow copy) and the sync re-entrancy item are untouched and still open.

rabble:
## Pushed: privacy-flip redaction + a ratchet-forced extraction

Head is `5917cda96`, rebased onto fresh `origin/main`. All 16 checks green.

### 1. Privacy flip now redacts the plaintext event (answers @NotThatKindOfDrLiz's decision #1)

Full reasoning is in the [reply on that thread](https://github.com/divinevideo/divine-mobile/pull/6661#discussion_r3771338563). Short version: a kind 5 with an **`e` tag** on the old plaintext event id retires it without touching the sealed replacement (only an **`a` tag** would sweep the whole coordinate — that is what the old comment ruled out, correctly, for the wrong tag). With that in place the flip commits at accepted-by-any, which makes the inconsistent partial-flip state impossible instead of merely unlikely.

This removes `requireAllRelays`, a guard both of you re-reviewed. That reversal is deliberate and argued in the thread and the commit body; if you disagree, that is the place to push back.

### 2. Relay-edge extraction — NOT requested by review

Flagging this separately because it reshapes a file you have both already read, and no review comment asked for it.

`ci(architecture): ratchet service god-file growth` (#6932) merged to `main` while this PR was in review and froze `curated_list_service.dart` at 1848 lines. This branch had it at 2289, and the ratchet has no sanctioned raise path — `UPDATE_BASELINE` is explicitly refused for growth. So `Generated Files` went red on a PR that had nothing to do with the change.

`CuratedListRelayGateway` (new, 606 lines) takes everything the service did at the relay edge that does not touch local list state: public-list reads, the event↔list codec, NIP-44 seal/unseal, and the NIP-09 deletion and redaction publishes. The service keeps the cache and calls through the gateway. Its public read methods stay as thin delegates, so **no call site outside the service moved**. Service lands at 1757.

Deliberately left in the service: `_publishListToNostr` (it writes back into `_lists` after acceptance — splitting that across the boundary would touch the code you just verified) and `fetchUserListsFromRelays` (it merges into local state, so it is not a clean read).

Behaviour is unchanged — the 249 existing curated-list tests pass untouched — and the gateway carries its own tests for the sealing guards, unseal classification, and redaction tag shape. The full move into `curated_list_repository` is still the #4338 end state; this is ratchet compliance, not that migration.

### 3. Rebase

The pre-push hook blocked on `untested_services_floor: baseline GREW vs origin/main` naming `bookmark_service` and `nostr_list_service_mixin` — neither touched by this branch. The branch had fallen behind a `main` that shrank the baseline, and the hook prescribes a rebase over `UPDATE_BASELINE` for exactly that case. 13 commits replayed, no conflicts; your commits (`9d3f354f4`, `ee83d8d05`, `08ca7fc2a`, `aa14c7e18`) are all intact. Force-push was leased against the known remote SHA.

### Still open — not addressed here

- Partial relay acceptance is answered; **default-list withdrawal** under publish-everything is not.
- **Create-flow privacy copy** omitting the visible-name/metadata caveat.
- **Sync re-entrancy** — the once-per-session flag set after the await window, and `force: true` overlapping syncs.
- @mbradley's **union-merge vs deletion** ambiguity from the 8 Aug round.

Auto-merge is off and staying off until those are answered.

mbradley (commented):
Reviewed the relay-edge extraction at `5917cda96`, the commit no review asked for. It preserves behavior: every method that moved into the new gateway matches its previous version line for line, and the service calls the gateway with the same arguments at the same points. The parts that carry the privacy guarantee are unchanged. A private list's items are still encrypted to the owner's own key, and are not published if the signing key doesn't match the signed-in account. Only the owner's own events are ever decrypted. Making a list private still deletes only the old public event by its id, never the list's whole address, so the encrypted replacement survives, and that deletion still runs only after a list actually flips from public to private. All 249 list tests pass. I confirmed the redaction and the own-events-only classification tests go red when the guard they cover is removed.

mobile/test/services/curated_list_relay_gateway_test.dart
```diff
@@ -0,0 +1,200 @@
...
+            pubkey: _ownerPubkey,
+            tags: const [
+              ['d', 'list-1'],
+              ['e', _videoEventId],
+            ],
+          ),
+        );
+
+        expect(unsealed.status, UnsealItemTagsStatus.notSealed);
+      });
```
mbradley, line 127: This test can't catch a regression in the plain-tags check. Its content isn't encrypted, so the not-encrypted check already returns not-sealed even if the plain-tags check were deleted. The case that needs coverage is an event whose content is encrypted but that also carries plain item tags: it should still be reported not-sealed. With the plain-tags check removed, that event decrypts instead, and no current test catches it. I confirmed the test below passes as-is and fails when the check is removed:

```suggestion
      });

      test('reports sealed content with plain tags as not sealed', () async {
        final sealed = await gateway.sealItemTags(_list());

        final unsealed = await gateway.unsealItemTags(
          _event(
            content: sealed!,
            pubkey: _ownerPubkey,
            tags: const [
              ['d', 'list-1'],
              ['e', _videoEventId],
            ],
          ),
        );

        expect(unsealed.status, UnsealItemTagsStatus.notSealed);
      });
```

NotThatKindOfDrLiz:
@rabble I took over the final review fixes and pushed commit 6344407574.

What changed:
- Default list deletion now publishes a NIP-09 withdrawal, persists a local tombstone, and suppresses stale relay restores for my_vine_list.
- The create-list privacy copy now explicitly says videos stay private while the list name remains visible, with generated l10n refreshed.
- Updated regression coverage for default-list withdrawal/recreate/relay-restore behavior and the affected widget copy.

Verification:
- Local pre-push hook passed analyzer and changed-file tests.
- flutter test test/services/curated_list_service_crud_test.dart test/widgets/add_to_list_dialog_test.dart test/l10n/l10n_test.dart test/l10n/arb_consistency_test.dart test/l10n/arb_script_integrity_test.dart
- flutter analyze lib test
- GitHub checks are green after rerunning a transient shard-0 sqlite3 native-asset download failure.

Please review the takeover commit when you get a chance and merge when ready.

NotThatKindOfDrLiz (approved):
Approved after takeover commit 6344407574. I verified the default-list withdrawal gap and privacy-copy caveat are addressed, local verification passed, and GitHub checks are green after rerunning the transient sqlite3 download failure.

NotThatKindOfDrLiz:
@mbradley Can you pls do a quick re-review so we can clear requested changes and get this merged? TYSM!

mbradley (commented):
Re-reviewed at `6344407`. Clearing my changes-requested to a comment — the blockers from the Aug 8 pass are resolved, and I verified each at head rather than taking the fix commits on trust:

- The `sealed:` production sentinel is gone; classification decodes the real NIP-44 v2 envelope.
- The relay-edge gateway extraction is a true pass-through — all seven privacy guards (seal-to-owner + signer match, own-events-only unseal, decrypt-failure-skips-merge, oversized-payload reject, private+collaborative reject, thumbnail omission, and the tri-state classification) are intact.
- The privacy-flip redaction is correct: an `e` tag on the old event id (not an `a` tag on the coordinate) is the right NIP-09 shape — an `a` tag would take the sealed replacement down with the original — it fires only on publish success, and the old flip-window leak is closed now that `updateList` and `addVideoToList` share the same per-list lane.
- Default-list withdrawal, the create-flow name disclosure, and private+collaborative rejection are all in place and tested; a private list never renders its ciphertext into Explore (the discovery path rejects lists with no readable items).

@NotThatKindOfDrLiz's approval carries the merge; auto-merge stays your call. One open item and a few take-or-leave notes below.

**Open — the sync concurrency model** (my Aug 8 #3, and your Aug 11 note, @NotThatKindOfDrLiz). Still unresolved at head, and this pass pinned a sharper variant than the generic write-back race either of us described. The merge phase writes `_lists` off the per-list lane (`curated_list_service.dart:1706`) and `fetchUserListsFromRelays` is re-entrant (`:1387` bypassed by `force`, `_hasSyncedWithRelays` set after the awaits at `:1494`). Interleave a private-flip's confirmed-publish await (up to 15s) with a concurrent multi-device sync that holds a newer public copy: the merge overwrites the entry, then the flip's write-back at `:1259` re-resolves and stamps `isPublic: false` onto it — so the list reads **private locally while relays serve the newer public copy with plaintext item tags**. The same interleaving on an add silently drops the concurrent item. Not blocking from me — it's a deferred concurrency-model decision and merge is yours — but it should be tracked as its own issue rather than living in this thread, since that private-locally/public-on-relay variant isn't recorded anywhere yet. Fix shape is the one already discussed: run the merge through `_serializeListOperation` per event, and coalesce concurrent syncs behind one in-flight future with the flag set before the await.

**Copy — the create-a-private-list dialog discloses only the name.** `toPrivateMetadataTags` still publishes `description`, `image` and hashtags in clear (`curated_list_converter.dart:190-200`), but the toggle subtitle and the make-private warning both promise only that the name stays visible. The dialog has a free-text Description field right above the toggle, so a user typing sensitive context there has no signal it's public. Copy call for you/design — suggest "name, description and cover image stay visible."

**Suggestion A (take or leave) — arm the default-list tombstone before removal.** `_deleteOwnedList` sets the prefs flag *after* `_removeListAndSubscription` (`:709-711`), so a `force` sync interleaving the removal's await can re-add the default list before the suppression is armed. Setting it just before the removal closes that window (and gating it there rather than at the top of the method avoids arming it when the deletion publish failed). Dissolves anyway if the concurrency item gets the lane fix.

```dart
      if (list.nostrEventId != null) {
        if (!await _relayGateway.publishListDeletion(listId)) {
          return false;
        }
      }

      if (listId == defaultListId) {
        await _prefs.setBool(defaultListDeletedStorageKey, true);
      }
      await _removeListAndSubscription(listId);
```

**Suggestion B (take or leave) — re-read the owner inside the backfill loop.** `_backfillUnpublishedLists` captures `owner` once (`:1560`) and re-checks it against the stale value inside the lane, so an account switch mid-loop republishes the previous identity's stranded list under the new account's key. Re-read the pubkey inside the operation and bail if it changed — the un-finished half of the patch E I flagged earlier.

```dart
    for (final list in stranded) {
      await _serializeListOperation(list.id, () async {
        final currentOwner = _relayGateway.currentAuthenticatedPubkey();
        if (currentOwner == null || currentOwner != owner) return;
        var current = getListById(list.id);
        // ... unchanged, comparing/stamping against currentOwner ...
      });
    }
```

Thanks.

NotThatKindOfDrLiz (approved):
Takeover commit d9be0197f24374c31f6b4699a29bc779d3c75112 is pushed.

What changed:
- Preserved nostrEventId across failed public republishes, so a later make-private flip can still redact the prior plaintext event and delete can still publish the coordinate deletion.
- Moved retry selection onto pendingRepublish while still backfilling never-published lists.
- Added regressions for make-private-after-failed-rename redaction and delete-after-failed-rename coordinate deletion.
- Added the default-list tombstone cleanup key and coverage.
- Added mbradleys gateway regression for sealed-looking content with plain item tags.
- Kept private-list metadata public by product decision, and made create/edit privacy copy explicit across l10n: videos are private; name, description, tags, and cover stay visible.
- Filed sync-concurrency follow-up as #7207, assigned to NotThatKindOfDrLiz.

Verification:
- flutter analyze lib test
- flutter test test/services/curated_list_service_crud_test.dart test/services/curated_list_relay_gateway_test.dart test/services/user_data_cleanup_service_test.dart
- broader curated-list service relay test set, 289 tests
- flutter test test/widgets/add_to_list_dialog_test.dart
- pre-push changed-file analyzer/tests
- GitHub checks are all green at d9be0197f24374c31f6b4699a29bc779d3c75112

@rabble please review the takeover commit and merge when ready.

NotThatKindOfDrLiz:
## Post-merge Rereview Analysis

This rereview verified the current head (d9be0197f) against the rereview bar. The following standing items from prior review cycles remain unanswered and merit explicit clarification in a follow-up:

### Unanswered Product Decisions (from prior review rounds)

**A. Default-list withdrawal model:**
The PR makes default-list deletion publish a NIP-09 kind 5 withdrawal instead of skipping the delete entirely. No prior review comments explain: (1) whether a silent no-op is worse UX than publishing and failing, (2) whether the copy / product strategy around default lists has settled, or (3) whether this should block v1. Document the decision so a future maintainer can reason about this path.

**B. Create-flow transparency:**
The new-list create flow doesn't warn that private lists encrypt *items* but not the *name*. The make-private flow covers this, but a user creating a new private list from scratch never sees that distinction. No prior review resolved whether this is intentional product design or an oversight.

**C. Sync re-entrancy model:**
The sync merge at lines 1376–1512 has a re-entrancy window: incomplete syncs can re-trigger before backfill finishes, and multiple overlapping syncs can reach union-merge simultaneously. The code handles this via the `_syncInProgress` flag, but no prior review explicitly confirmed the chosen semantics (last-write-wins / union-merge on overlap) match product intent for multi-device conflicts.

### Technical Edge Cases (identified in this rereview)

**D. Union-merge auth gap:**
When union-merging public and private list versions from divergent relays (lines 1298–1345), the merge commits to private (`isPublic = (existing && current) || isCollaborative` at line 1319), clears `nostrEventId` at line 1327, then publishes a sealed version at line 1331. If the public version was learned from a relay that does NOT have the sealed replacement yet, the code publishes with no prior redaction of the public version. Per NIP-09 semantics (e-tag targets specific event ID), the sealed event becomes a parallel version, not a replacement. This could leave a readable public copy on some relays. Mitigation: either (1) redact the prior public version before publishing the sealed one, or (2) document that this window is acceptable and relays will eventually converge.

**E. Backfill owner capture:**
The backfill loop at lines 1556–1596 captures the owner pubkey once at the top of the loop (line 1561), then publishes under that key in the 15-second retry cycle. If the auth state changes mid-backfill (e.g., user signs out and back in as a different account), later iterations of the loop will publish the list under the wrong account's key. This is blocked by the signer capability check in `sealItemTags()`, but the check happens *after* the list is already being published. Mitigation: pass owner as part of the work item or re-sample it before each publish attempt.

**TODO (pre-existing):**
`mobile/test/unit/curated_list_relay_sync_test.dart` line 107 has a commented test ("Fix and re-enable this test") over a subscription filter assertion. This predates the current PR but remains unaddressed.

### Verdict

The rereview confirms:
- NIP-09 e-tag redaction reversal is correct (prior review's "deletion is redaction, not withdrawal" concern is unfounded)
- No new technical debt introduced; extraction and separation of concerns are clean
- No new breaking changes

The PR is production-ready pending explicit product sign-off on decisions A/B/C and resolution of edge cases D/E.

PR #6657, opened 2026-08-04
fix(profile): persist staged media and guard edits

NotThatKindOfDrLiz:
## Summary
- persist staged avatar/banner Blossom URLs in a pubkey-scoped SharedPreferences store with TTL and corrupt-payload cleanup
- restore staged URLs when Edit Profile recreates the ProfileEditorBloc
- guard unsaved profile edits with a Save / Discard / Keep editing prompt across system back, app-bar back, and Cancel
- clear staged media from both BLoC state and the staged-media store on explicit discard
- track profile text-field dirtiness in ProfileEditorState and fix explicit banner-clear save semantics

Closes #6652
Closes #6658

## Tests
- flutter analyze
- flutter test test/blocs/profile_editor/profile_editor_bloc_test.dart test/screens/profile_setup_screen_test.dart
- pre-push hook: analyze, generated-file verification, changed-file tests

PR #6663, opened 2026-08-04
fix(profile): use the alert sticker on the unsaved-changes prompt

### LINKED ISSUE #6665: fix(profile): use the alert sticker on the unsaved-changes prompt
Auto-created from PR #6663.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6663
**Branch:** `fix/profile-unsaved-changes-sticker`

hm21:
Closes #6665

## Description

The "unsaved changes" prompt on the edit-profile screen (#6657) came from the video editor save-draft sheet and kept its clapperboard sticker. On a screen that edits a profile, a clapperboard reads as "video" and has nothing to do with what the sheet is asking about. Swapped to `DivineStickerName.alert`, which is what this repo already shows on every other "confirm before you lose something" sheet: trash restore/delete, delete video, discard captions, and clips. The video editor prompt keeps the clapperboard.

**Related Issue:** #6657

## Out of Scope

No test. The change is a single design-token swap with no behavior attached; the existing `profile_setup_screen_test.dart` prompt tests assert on the l10n button/title text and stay green. A test pinning the enum value would only restate the source line, not catch a regression.

## Verification

- `dart format` (no changes)
- `flutter analyze lib/screens/profile_setup` - no issues
- `flutter test test/screens/profile_setup_screen_test.dart` - 70/70 pass

## Type of Change

### DIFF
--- a/mobile/lib/screens/profile_setup/view/profile_setup_view.dart
+++ b/mobile/lib/screens/profile_setup/view/profile_setup_view.dart
@@ -276,7 +276,7 @@ class _ProfileSetupScreenViewState extends ConsumerState<ProfileSetupScreenView>
   }) {
     VineBottomSheetPrompt.show<void>(
       context: context,
-      sticker: .videoClapBoard,
+      sticker: .alert,
       title: context.l10n.profileSetupUnsavedChangesTitle,
       subtitle: context.l10n.profileSetupUnsavedChangesSubtitle,
       primaryButtonText: context.l10n.profileSetupUnsavedChangesSaveButton,


realmeylisdev (approved):
lgtm. rendered both versions of the sheet locally and the triangle reads right over "Save changes?", and the clapperboard is still on the video editor's own save-draft prompt, so nothing else got dragged along with it. agree a test here would just restate the line.

description still says the samsung check is pending, worth updating if it already happened.

NotThatKindOfDrLiz (approved):
Approved after independent review: the alert sticker matches the existing confirm-before-loss prompt convention, the asset exists, and CI is green.

PR #6666, opened 2026-08-04
feat(upload): report publish phase timings to Firebase Performance

### LINKED ISSUE #6556: task(upload): report publish phase timings to Firebase Performance
## Description

The publish pipeline emits per-phase timings to the log only. Report them to Firebase Performance as well, so the phase breakdown exists as a distribution across the user base rather than as one line in one person's log export.

Blocked by #6554 — the phases and the `PublishTimeline` that owns them land there.

## Context

#6554 added `PUBTIME` phase timing and used it to take ~32% off a publish. Those numbers came from nine runs on one Samsung Galaxy on one network. That was enough to find the client-side waste, and it is not enough for what is left, because everything still on the clock is server- or signer-bound:

- `net.put` — 2 273 to 7 216 ms for a constant 16 572 bytes, no relation to bandwidth. On a 50 Mbit/s line that is ~3 ms of transfer.
- `auth.sign` / `nostr.sign` — 274 to 1 355 ms each. Accounts on `divineOAuth` have no local key, so every signature is a Keycast RPC round-trip.

Whether those are systemic or an artefact of one device and one network cannot be answered from a single log export. A P95 across real users answers it immediately, and it also turns any future regression in this area into something visible rather than something a user reports as "posting feels slow".

Firebase Performance is already wired up and already carries `video_upload`, `video_generation` and `camera_startup`. But `video_upload` spans the whole Blossom call as a single number — it separates neither transfer from thumbnail nor signing from the PUT, and it does not cover the Nostr phase at all. So today's traces cannot answer any of the questions above.

## Implementation Notes

- `mobile/lib/services/video_publish/publish_timeline.dart` already collects every phase and emits a summary; it is the natural place to also report metrics.
- `PerformanceTraceMonitor.startOperationTrace` returns an operation-scoped handle, which is the API to use — the name-keyed `startTrace` has the overlapping-operation pitfalls documented on it.
- Suggested shape: one `video_publish` trace per publish, with the phases as custom metrics (`transfer_ms`, `thumbnail_ms`, `sign_ms`, `nostr_publish_ms`, `total_ms`) and attributes for the outcome and whether the thumbnail was reused from a previous attempt.
- Worth carrying the payload size as a metric too, so throughput can be derived rather than guessed.
- The monitor should be injected and optional, so the timeline stays testable and unit tests do not reach Firebase.
- Keep the log lines. They are what makes a single bug report debuggable; the dashboard answers a different question.

## Acceptance Criteria

- [ ] A completed publish reports its phase durations to Firebase Performance
- [ ] Failed and abandoned publishes are distinguishable from successful ones in the console
- [ ] The existing `PUBTIME` log lines still work unchanged
- [ ] Unit tests do not depend on Firebase being available

## Testing

- Unit-test the timeline against a fake monitor: phases reported once, with the values the summary carries
- Verify on device that the trace appears in the Firebase console with the expected metrics

hm21:
## Description

The publish pipeline already times every phase, but only into the log — so the numbers exist for one person, on one device, on one network. That was enough to find the client-side waste in #6554. What is left on the clock is server- and signer-bound (`net.put` between 2.3 s and 7.2 s for a constant 16 KB payload, Keycast signatures between 0.3 s and 1.4 s), and only a distribution across real users can say whether that is systemic or an artefact of one Samsung on one line.

Each publish now also reports a `video_publish` Firebase Performance trace:

- **Metrics** — `transfer_ms`, `thumbnail_ms`, `sign_ms`, `nostr_publish_ms`, the coarse `upload_ms` / `nostr_ms` containers, `mentions_ms` / `subtitles_ms` / `invites_ms`, `total_ms`, plus `payload_bytes` and `transfer_legs`. A phase that never ran reports no metric, so an absent metric reads as "did not happen" rather than as zero, and a retried phase adds its attempts together.
- **Attributes** — `outcome` (`success` / `error:<kind>` / `threw`), `thumbnail` (`reused` / `generated`) and `signature` (`reused` / `signed`), so a failed, abandoned or short-circuited publish is filterable in the console instead of skewing the distribution for the work itself.

Three design points worth calling out:

**Why a zone instead of a parameter.** The interesting phases are timed several layers below `PublishTimeline` — `upload.transfer` and `upload.thumbnail` inside `UploadManager`, `nostr.sign` and `nostr.publish` inside `VideoEventPublisher` — and reach the log through the free `logPublishPhase` function. The timeline now installs itself as a zone value for the length of the publish and folds those lines in. Threading it through as a parameter would have meant touching five methods in the upload manager alone, and would still have missed the resume/retry paths, which re-enter `_performUpload` from outside the publish call chain.

**Why the summary line is unaffected.** Nested phases are folded into the metric map only, not into the summary's top-level phase list — counting `upload.transfer` there would double it against its `upload` parent and drive the `other` bucket negative. The `PUBTIME` lines are byte-for-byte what they were.

**Why a leg that was skipped is tagged rather than dropped.** A retry reuses a thumbnail already on the CDN and an event the signer already signed, so those legs report ~0 ms. Without the `thumbnail` / `signature` attributes a retry-heavy population would drag `thumbnail_ms` and `sign_ms` toward zero invisibly — and signer latency is half of what this trace exists to measure. A leg that did real work on one attempt stays tagged as such when a later attempt short-circuits, because its metric then carries that work. A leg the publish never reached reports no attribute at all, which is the same rule the metrics follow.

Two things the metrics do *not* say, both documented at the constants:

- They are **not additive**. `upload.thumbnail` runs parallel to `upload.transfer`, so `upload_ms` is roughly the longer of the two rather than their sum. Only the summary log line's top-level phases add up to the total.
- `payload_bytes / transfer_ms` is **effective** throughput — bytes delivered per unit of wall clock the user waited, retry overhead included — not the line rate. `transfer_legs` counts how many times the publish entered the transfer phase, not how many attempts the upload service made inside one: a single leg can contain a failed whole-file OS attempt, a chunked resumable fallback that re-sent the same bytes, server iteration and chunk retries, none of which the publish layer can see. Recovering the line rate means instrumenting `blossom_upload_service` at the transport boundary, which is out of scope here.

The monitor is injected and optional (`NoOpPerformanceTraceMonitor` by default), so unit tests never reach Firebase, and the trace is stopped fire-and-forget so a publish never waits on a platform round-trip to return its result.

`PerformanceTrace` gained `setMetric` — `startOperationTrace` handed out a handle that could be tagged but not measured, which is the API this needs.

**Fixed along the way:** `startOperationTrace` fires `trace.start()` and returns the handle synchronously, so a short operation could call `stop()` while the start round-trip was still in flight. The plugin's `stop()` early-returns when the platform has not handed back a trace handle yet — that trace is then never reported *and* leaks natively. The publish trace made this reachable: an ownership or auth check that rejects returns after a single await, which is exactly the `outcome=error:*` population the console filter is for. `stop()` now waits for the start future. Attributes and metrics were never at risk; the plugin buffers them in Dart and flushes them with the stop call.

**Related Issue:** Closes #6556

## Out of Scope

`auth.sign` and `net.put` are timed inside the `blossom_upload_service` package, which cannot import the app-layer timeline, so they stay log-only. Both sit inside `upload.transfer`, so the reported transfer metric still brackets them — a P95 on `transfer_ms` answers the "is the PUT systemically slow" question without a reporter port through the package.

An upload already in flight from an earlier session is joined rather than started by the publish that waits on it, so its `upload.transfer` line is emitted outside this publish's zone and reports no `transfer_ms`. The publish still reports `upload_ms`, measured by the timeline itself, so the leg is visible as a duration — just not broken down. Attributing a joined upload to whichever publish happened to wait on it would be the wrong answer anyway.

## Verification

- Verified on a real Samsung Galaxy device.
- `dart format` — clean
- `flutter test test/services/video_publish/ test/services/performance_monitoring_service_test.dart test/services/upload_manager_*.dart test/services/video_event_publisher*.dart test/blocs/video_recorder/ test/blocs/background_publish_bloc/ test/providers/video_editor_provider_test.dart test/providers/video_publish_provider_test.dart` — 657 passing
- New coverage: the timeline against a fake monitor (phases reported once with the values the summary carries, retries summed, transfer-leg count reported, one leg counted however many transport attempts it contained, absent phases omitted, outcome / thumbnail-reuse / signature-reuse attributes, no attribute for a leg the publish never reached, trace stopped once, phases emitted outside the publish ignored), and the service end-to-end reporting `success` / `error:notSignedIn` / `threw` through an injected fake.

Note for whoever reads the console first: Firebase surfaces percentile distributions for trace *duration*; custom metrics are aggregated. If a P95 on `transfer_ms` specifically is what we want rather than an average, that needs the BigQuery export of Performance data — the instrumentation is the same either way.

## Type of Change

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -2760,6 +2760,7 @@ class _DivineAppState extends ConsumerState<DivineApp>
           dmRepository: ref.read(dmRepositoryProvider),
           l10n: currentAppL10n(ref.read(sharedPreferencesProvider)),
         ),
+        performanceMonitor: ref.read(performanceMonitoringServiceProvider),
         onProgressChanged:
             ({required String draftId, required double progress}) {
               onProgress(draftId: draftId, progress: progress);
--- a/mobile/lib/providers/video_publish_provider.dart
+++ b/mobile/lib/providers/video_publish_provider.dart
@@ -27,6 +27,7 @@ import 'package:openvine/providers/clip_manager_provider.dart';
 import 'package:openvine/providers/layer_rasterizer_provider.dart';
 import 'package:openvine/providers/preferences_providers.dart';
 import 'package:openvine/providers/repository_providers.dart';
+import 'package:openvine/providers/service_providers.dart';
 import 'package:openvine/providers/shared_preferences_provider.dart';
 import 'package:openvine/providers/social_providers.dart';
 import 'package:openvine/providers/upload_media_providers.dart';
@@ -164,6 +165,7 @@ class VideoPublishNotifier extends Notifier<VideoPublishProviderState> {
         l10n: currentAppL10n(ref.read(sharedPreferencesProvider)),
       ),
       languagePreferenceService: ref.read(languagePreferenceServiceProvider),
+      performanceMonitor: ref.read(performanceMonitoringServiceProvider),
       onProgressChanged: ({required String draftId, required double progress}) {
         setUploadProgress(draftId: draftId, progress: progress);
         onProgressChanged(draftId: draftId, progress: progress);
--- a/mobile/lib/services/performance_monitoring_service.dart
+++ b/mobile/lib/services/performance_monitoring_service.dart
@@ -18,6 +18,10 @@ abstract class PerformanceTrace {
   /// Adds an attribute for filtering in the Firebase console.
   void putAttribute(String attribute, String value);
 
+  /// Sets a custom metric on this trace, so an operation can report a
+  /// breakdown (per-phase durations, payload sizes) next to its duration.
+  void setMetric(String metric, int value);
+
   /// Stops the trace and records its duration.
   Future<void> stop();
 }
@@ -49,6 +53,9 @@ class _NoOpPerformanceTrace implements PerformanceTrace {
   void putAttribute(String attribute, String value) {}
 
   @override
+  void setMetric(String metric, int value) {}
+
+  @override
   Future<void> stop() async {}
 }
 
@@ -65,6 +72,18 @@ class _FirebasePerformanceTrace implements PerformanceTrace {
     } catch (e) {
       Log.error(
         'Failed to put attribute $attribute: $e',
+        name: 'PerformanceMonitoring',
+      );
+    }
+  }
+
+  @override
+  void setMetric(String metric, int value) {
+    try {
+      _trace.setMetric(metric, value);
+    } catch (e) {
+      Log.error(
+        'Failed to set metric $metric: $e',
         name: 'PerformanceMonitoring',
       );
     }
--- a/mobile/lib/services/performance_monitoring_service.dart
+++ b/mobile/lib/services/performance_monitoring_service.dart
@@ -1,8 +1,6 @@
 // ABOUTME: Performance monitoring service for tracking app performance metrics
 // ABOUTME: Uses Firebase Performance Monitoring to track screen transitions, network requests, and custom operations
 
-import 'dart:async';
-
 import 'package:firebase_performance/firebase_performance.dart';
 import 'package:unified_logger/unified_logger.dart';
 
@@ -61,9 +59,16 @@ class _NoOpPerformanceTrace implements PerformanceTrace {
 
 /// A [PerformanceTrace] backed by a live Firebase [Trace].
 class _FirebasePerformanceTrace implements PerformanceTrace {
-  _FirebasePerformanceTrace(this._trace);
+  _FirebasePerformanceTrace(this._trace, this._started);
 
   final Trace _trace;
+
+  /// Completion of the trace's start round-trip. [stop] waits on it because
+  /// the plugin drops a stop that arrives before the platform handed back a
+  /// trace handle — the trace is then never reported *and* leaks natively.
+  /// Operations that fail fast (an auth or ownership check that returns after
+  /// a single await) are exactly the ones that would race it.
+  final Future<void> _started;
 
   @override
   void putAttribute(String attribute, String value) {
@@ -92,6 +97,7 @@ class _FirebasePerformanceTrace implements PerformanceTrace {
   @override
   Future<void> stop() async {
     try {
+      await _started;
       await _trace.stop();
     } catch (e) {
       Log.error('Failed to stop trace: $e', name: 'PerformanceMonitoring');
@@ -189,23 +195,22 @@ class PerformanceMonitoringService implements PerformanceTraceMonitor {
   /// Unlike [startTrace], nothing is stored in [_activeTraces]: the caller
   /// owns the returned [PerformanceTrace] and tags/stops it directly, so
   /// overlapping operations of the same name keep independent traces. Start is
-  /// fire-and-forget so callers stay synchronous; attribute/stop calls on the
-  /// handle are delivered to the platform in order after it.
+  /// fire-and-forget so callers stay synchronous; attributes and metrics are
+  /// buffered until stop, and stop itself waits for the start round-trip so a
+  /// short operation cannot end its trace before the platform opened it.
   @override
   PerformanceTrace startOperationTrace(String traceName) {
     if (!_initialized) return const _NoOpPerformanceTrace();
 
     try {
       final trace = _performance.newTrace(traceName);
-      unawaited(
-        trace.start().catchError((Object e) {
-          Log.error(
-            'Failed to start trace $traceName: $e',
-            name: 'PerformanceMonitoring',
-          );
-        }),
-      );
-      return _FirebasePerformanceTrace(trace);
+      final started = trace.start().catchError((Object e) {
+        Log.error(
+          'Failed to start trace $traceName: $e',
+          name: 'PerformanceMonitoring',
+        );
+      });
+      return _FirebasePerformanceTrace(trace, started);
     } catch (e) {
       Log.error(
         'Failed to start trace $traceName: $e',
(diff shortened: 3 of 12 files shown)


rabble (approved):

rabble (commented):
Approving. Agent-assisted review; no branch changes.

The design is sound. Timing the phases where they happen and keeping the phase → Firebase metric mapping in one place is the right split, and the constants shared between emitter and mapping avoid the usual drift. The doc comments do real work — particularly the note that the metrics are **not** additive (`upload.thumbnail` runs parallel to `upload.transfer`, so `upload_ms` is the longer of the two, not the sum), and the warning that `transfer_ms` sums attempts so throughput is only derivable when `transfer_attempts == 1`. Those are exactly the two ways someone reads a publish dashboard wrong, and they're documented at the definition rather than in a wiki nobody opens.

Using `null` rather than `0` for a phase that never ran is the right call too — an absent metric reads as "did not happen" instead of "happened instantly", which matters for a publish that failed early.

Two things to confirm before or shortly after merge, neither blocking:

**1. Trace lifecycle on the failure path.** `PublishTimeline`'s constructor calls `startOperationTrace(publishTraceName)` eagerly. If a publish throws between construction and whatever stops the trace, does the trace get stopped in a `finally`? An unstopped Firebase trace is worse than no trace — it either never reports or reports a bogus duration, and the failure path is precisely the one you most want measured. If `run()` wraps in try/finally this is already handled; worth a line in the test if it isn't covered.

**2. Zone-ambient lookup is invisible at the call site.** `PublishTimeline.current` reads `Zone.current[_timelineZoneKey]`, so `logPublishPhase` silently contributes to no timeline if a phase is emitted from a callback that escaped the zone — a `Timer`, an isolate hop, or a future scheduled outside `run()`. That failure mode is silent by construction: you get the log line but no metric, and nothing indicates the metric was dropped. Given the upload leg is full of callbacks, it'd be worth either a debug-mode assert when `current` is null inside a phase that expects a timeline, or a comment naming which call sites are known to be inside the zone.

Neither changes my read that this is worth landing — the observability payoff is real and the log path degrades independently of the Firebase path.

rabble (approved):
Approving — see detailed review above.

rabble (changes requested):
Request changes based on the late review.

`transfer_attempts` is documented in `mobile/lib/services/video_publish/publish_timeline.dart` as the number of transfer attempts, and `transfer_ms` is described as summing attempts. However, `mobile/lib/services/upload_manager.dart` emits one `uploadTransfer` phase around a single `uploadVideoWithResume` call. That call can include the OS upload, legacy/resumable fallback, and chunk retries, so a multi-attempt transfer is recorded as `transfer_attempts == 1`. Count attempts at the transport boundary, or rename/re-document the metric so it cannot be used to filter single-attempt throughput incorrectly; add a regression test.

hm21:
@rabble All three addressed in `cac31ab57c`.

**`transfer_attempts` (blocking)** — you're right, and it was worse than a naming problem: `transfer_legs == 1` was the value someone would have trusted to hold a clean single-attempt throughput, and it's exactly the value a failed OS attempt plus a full chunked re-upload produces.

I took the rename-and-re-document option rather than counting at the transport boundary, because counting at `uploadVideoWithResume`'s two branches would only move the inaccuracy one level down — the server iteration and per-chunk retries live inside `uploadVideo`, so that counter would still under-report. Doing it properly means a new public API on `blossom_upload_service` under its 100% coverage gate, which is the package this PR deliberately scoped out.

So: `transfer_attempts` → `transfer_legs`, documented as how many times the publish *entered* the transfer phase, explicitly not how many attempts the upload service made inside one. `payload_bytes` is re-documented as **effective** throughput including retry overhead, with the note that no combination of these metrics recovers the line rate and what it would take to get one. Regression test pins that one phase counts as one leg no matter how long it ran.

**1. Trace lifecycle on the failure path** — already covered. `publishVideo` builds the timeline, wraps the work in `try`, and calls `timeline.finish(...)` from `finally`, so the trace is stopped on every exit including a throw. `d9d0b41e56` added the test for it: it makes the publish throw and asserts both `outcome=threw` and `stopCount == 1`.

Worth adding that this was reachable for a different reason, fixed in `7cff57dc63`: `startOperationTrace` fired `trace.start()` unawaited, and the plugin's `stop()` early-returns when the platform hasn't handed back a handle yet — so a publish that failed fast dropped its trace *and* leaked it natively. `stop()` now waits for the start future.

**2. Zone-ambient lookup** — documented at `logPublishPhase` rather than asserted. An assert would fire on a legitimate case: when the publish joins an upload another chain already had in flight (resumed at app start, or kicked off by the recovery sweep), that leg belongs to the chain that started it, so reporting no metric is the right answer rather than a missing one. The doc now names each emitter and which side of the zone it sits on, and notes that `upload_ms` still covers the leg — the console shows a duration without a breakdown rather than nothing.

rabble (approved):

`cac31ab` resolves the blocker. `transfer_attempts` is now `transfer_legs`, and the doc says what it actually counts: how many times the publish entered `PublishPhases.uploadTransfer`, not how many attempts `uploadVideoWithResume` made inside one. The distinction is pinned by a test that can fail — `counts one leg however many attempts it took inside` asserts `transfer_legs == 1` for a four-minute leg, which is exactly the trace someone would otherwise have filtered on for clean throughput.

`payload_bytes` being re-documented as **effective** throughput rather than line rate is the right call, and naming `blossom_upload_service`'s transport boundary as where a real attempt counter would have to live is more useful than leaving it implicit. No stale `transfer_attempts` references remain in the tree.

The zone-lookup note I raised is answered better than I asked for. Enumerating the emitters and separating "outside the zone because the publish joined an upload another chain started" from "outside the zone because someone broke it" is the part that matters — and it explains why there is deliberately no assert, since the joined-upload case would trip on every resumed publish. That reasoning belongs at the definition, and it's there now.

Ship it.

PR #6667, opened 2026-08-04
fix(people-lists): stop a mutation rollback from resurrecting torn-down state

### LINKED ISSUE #6504: fix(people-lists): a mutation rollback resurrects state the bloc tore down
## Summary

A people-lists mutation that is still in flight when the bloc tears its state down resumes against a snapshot captured *before* the teardown and puts it straight back — under whatever owner is current by then. The same emit clears `pendingMutations`, which `UserListPeopleScreen` reads as "the delete settled", so it announces a successful delete and pops the route for a delete that never landed.

Both are pre-existing: an account switch mid-publish already triggers them with no feature flag involved. PR #6502 (curated-lists flag-off lifecycle, #6494) adds a second trigger and documented the deferral in its description; this issue is that deferral, tracked independently.

## Steps to Reproduce

**A — rollback restores the dropped snapshot under the new owner**

1. Sign in as owner A holding one list `l1` containing Alice.
2. Start a delete (or add/remove) — `PeopleListsRepositoryImpl` is mid round-trip.
3. Turn **Curated Lists** off (or switch accounts) while the publish is in flight. State drops to `lists=[]`, `pendingMutations={}`.
4. Sign in as owner B.
5. The publish resolves `PeopleListPublishResult.failed()` — the common failure shape, since the repository returns `failed()` rather than throwing.

**B — false "Deleted list" confirmation**

1. Sign in as owner A holding `l1`, open `UserListPeopleScreen` for it.
2. Confirm the delete.
3. Turn **Curated Lists** off (or switch accounts) before the publish resolves.

## Actual Result

**A** (reproduced at `0ef1bb8` against the real bloc):

```
after teardown:  enabled=false status=initial lists=0 pending=0
after rollback:  enabled=false status=failure owner=<ownerB> lists=[l1] index={<alice>: {l1}}
```

Owner A's list is back, attributed to owner B. With the flag path, the next flag-on lands on `enabled: true, status: loading, lists: [l1]` until `watchLists(B)` delivers — briefly *live* under the wrong owner, not merely invisible.

**B**: the teardown emit satisfies `_deleteSettled` (`hadPendingDelete && !hasPendingDelete`) with `status: initial`, so the failure branch is skipped:

```
deleteSettled=true status=initial failedBranch=false
```

The screen announces `curatedListDeletedSnack` and calls `context.pop()`. The list is still on the relay.

## Expected Result

- A rollback that resolves after the owner changed (or after a teardown) discards its snapshot instead of re-emitting it.
- `UserListPeopleScreen` distinguishes "the delete settled" from "the bloc dropped its pending mutations", and neither announces nor pops in the latter case.

## Regression?

No — pre-existing. The four rollback blocks in `_onDeleteRequested`, `_onPubkeyAddRequested`, `_onPubkeyRemoveRequested` and `_onCreateRequested` are unchanged from before #6502, and the account-switch variant reproduces without any flag involvement. Both modes are worse on a plain account switch, where the restored state lands on `enabled: true` and actually renders.

## Implementation Notes

One root cause, one likely fix: each rollback emit is computed from `previousLists`/`previousIndex` captured before the `await` and applied unconditionally afterwards. Guarding those emits on the owner still matching the one the mutation was issued for — and on the mutation still being present in `pendingMutations` — closes both the flag-off and the account-switch variants. `state.activeOwnerPubkey` already exists for this shape of check.

Files:

- `mobile/lib/features/people_lists/bloc/people_lists_bloc.dart` — the four `!result.submitted` / `catch` blocks
- `mobile/lib/screens/user_list_people_screen.dart` — `_deleteSettled` and its `BlocListener`

Self-healing that exists today and should *not* be mistaken for a fix: a flag-on re-opens `watchLists`, which re-seeds from the Hive cache and overwrites the stale snapshot.

## Acceptance Criteria

- [ ] A mutation that fails after an owner change leaves the new owner's state untouched.
- [ ] A mutation that fails after a flag-off teardown does not restore the dropped snapshot.
- [ ] A pending delete that is dropped by a teardown produces no "Deleted list" announcement and no route pop.
- [ ] Regression tests for all three, at bloc level and at `UserListPeopleScreen` level.

Related to #6494 and PR #6502.

hm21:
## Description

Three failures that share one root cause: a mutation result applied to state that no longer belongs to the mutation.

**A mutation rollback resurrects state the bloc tore down.** Each rollback is computed from a `previousLists`/`previousIndex` snapshot captured before the repository round-trip and was applied unconditionally afterwards. An account switch, a sign-out, or `FeatureFlag.curatedLists` going off in between all emit a fresh state on purpose — and the rollback put the dropped snapshot straight back, attributed to whoever is signed in by then. On a plain account switch the restored lists land on `enabled: true` and actually render.

The fix is one guard, `_resultStillApplies(mutation, owner)`, on every post-`await` emit in all four mutation handlers. A result may write only when the owner still matches *and* its mutation is still in `pendingMutations`. The mutation lookup is what actually detects the teardown — every teardown path clears the map — while the owner check keeps the guard correct if a future teardown ever preserves it. `addError` still fires before the guard, so a dropped failure is still reported.

**A false "Deleted list" confirmation.** The teardown emit clears `pendingMutations` wholesale, which `_deleteSettled` read as the delete resolving: the screen announced a successful delete and popped the route for a list still on the relay. The screen now records the owner it issued the delete for, and the listener returns without announcing or popping when the state that dropped the mutation belongs to another owner or has the feature off.

**A rollback reverts a repository emission that arrived mid-flight.** The same pre-flight snapshot also loses to `_onRepositoryListsChanged`, which keeps `pendingMutations` — so an emission that lands during the round-trip passes the teardown guard above, and restoring the snapshot afterwards drops whatever arrived meanwhile: a list synced from a relay, another device's edit. Milder than the cross-account leak (transient, and it self-heals on the next emission) but the same bug shape, and it sits on the lines this PR already touches.

So the rollback stops being snapshot-based. Each handler now undoes just its own change against the *current* lists — delete re-inserts the removed list at its old index, add reuses `_applyOptimisticRemove`, remove re-inserts the pubkey at its old index — and the reverse index is rebuilt from the result. All three undos are guarded by a `contains` check, and the repository writes its cache only after a successful publish, so a mid-flight emission carries the pre-mutation truth and every undo is a no-op against it. That also makes the rollback order-independent, which the snapshot never was. `previousLists`, `previousIndex`, and `_copyReverseIndex` are gone.

**Also folded in** (@realmeylisdev's finding on the issue): `_onCreateRequested` never read `result.submitted`, so a create the relay rejected landed on `status: ready`, contradicting the bloc's own failure-recovery contract. Latent today — `PeopleListsStatus.failure` is read in exactly one place and both create entry points dispatch and pop without waiting — but wrong, and cheap to fix in the same pass.

**Related Issue:** Closes #6504

## Out of Scope

`UserListPeopleScreen` still infers this delete's failure from the global people-lists status. That is pre-existing adjacent debt and is tracked in #6696 because a clean fix needs per-mutation completion state rather than a list-presence heuristic.

## Verification

- `flutter test test/features/people_lists test/screens/user_list_people_screen_test.dart test/router/people_lists_route_order_test.dart test/main_app_shell_repository_sync_test.dart test/widgets/profile test/screens/other_profile_screen_test.dart test/widgets/user_profile_tile_feature_flag_test.dart` — 426 green.
- `flutter analyze lib/features/people_lists/bloc/people_lists_bloc.dart test/features/people_lists/bloc/people_lists_bloc_test.dart` — clean.
- `flutter test test/features/people_lists/bloc/people_lists_bloc_test.dart` — 46 green.
- Design-system, process-global, shared-channel, file-size, todo and test-structure ratchets — all green.
- Every new test was confirmed red against the unfixed code first. The two screen tests fail with exactly the reported symptom (`Actual: ['Deleted list']`); the two mid-flight rollback tests fail with `Expected: ['l1', 'l3'] / Actual: ['l1']`.
- One new test is green both before and after by design: `a failed remove restores the member at its old index` pins the member ordering the snapshot restored for free. It is a guard on the rewrite, not a regression test — implementing the undo with `_applyOptimisticAdd` (which appends) turns it red.
- Manually verified on a real Samsung Galaxy (SM S942B, Android 16): a delete with Curated Lists turned off mid-publish and a delete across an account switch both stay put — no announcement, no pop, and the previous owner's lists do not surface under the new account. A failed delete restores the list at its original position, and a failed member removal restores the person at their original index in the carousel rather than appending them.

New regression coverage: 5 bloc cases for the teardown guard (delete rollback vs. flag-off, delete rollback vs. account switch, add that throws after a teardown, remove that throws after a teardown, create that resolves after a teardown), 3 for the mid-flight rollback race (delete, add, remove), the create-not-submitted case, and 2 screen cases asserting no announcement and no pop. The duplicated `GoRouter` setup in the screen tests moved into a shared `_pumpPushedListRoute` helper now that three tests need it.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
+++ b/mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
@@ -92,10 +92,7 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
        _clock = clock ?? _defaultClock,
        super(PeopleListsState(ownerPubkey: initialOwnerPubkey)) {
     on<PeopleListsStarted>(_onStarted);
-    on<PeopleListsEnabledChanged>(
-      _onEnabledChanged,
-      transformer: sequential(),
-    );
+    on<PeopleListsEnabledChanged>(_onEnabledChanged, transformer: sequential());
     on<PeopleListsRepositoryChanged>(
       _onRepositoryChanged,
       transformer: sequential(),
@@ -213,7 +210,7 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
       // can learn who is signed in.
       //
       // Dropping `pendingMutations` is also what stops a mutation still in
-      // flight from putting the snapshot back once it resolves — see
+      // flight from applying any result to state that has been torn down — see
       // [_resultStillApplies].
       emit(PeopleListsState(ownerPubkey: state.ownerPubkey, enabled: false));
       return;
@@ -637,13 +634,12 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
   /// Whether a mutation result that arrived after an `await` may still be
   /// written to state.
   ///
-  /// Rollbacks are computed from a snapshot captured before the round-trip and
-  /// were only ever valid for the state that produced it. In between, the bloc
-  /// may have torn that state down — an account switch, a sign-out, or
-  /// `FeatureFlag.curatedLists` going off each emit a fresh state with an empty
-  /// `pendingMutations`. Restoring the captured snapshot there hands one
-  /// owner's lists to whoever is signed in by then, and even a status-only
-  /// write reports an outcome the new state never asked for (#6504).
+  /// Mutation outcomes and rollbacks are only valid for the state that issued
+  /// them. In between, the bloc may have torn that state down — an account
+  /// switch, a sign-out, or `FeatureFlag.curatedLists` going off each emit a
+  /// fresh state with an empty `pendingMutations`. Applying a result there can
+  /// write one owner's lists or mutation status onto whoever is signed in by
+  /// then (#6504).
   ///
   /// The mutation lookup is what actually detects a teardown; the owner check
   /// keeps the guard honest if a future teardown ever preserves the map.
--- a/mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
+++ b/mobile/lib/features/people_lists/bloc/people_lists_bloc.dart
@@ -50,7 +50,10 @@ typedef PeopleListsClock = DateTime Function();
 /// mutations succeed or the repository stream never re-emits.
 ///
 /// A result that lands after the bloc tore down the state its mutation was
-/// issued against writes nothing at all — see [_resultStillApplies].
+/// issued against writes nothing at all — see [_resultStillApplies]. A rollback
+/// that does apply undoes only its own mutation rather than restoring a
+/// pre-flight snapshot, so it cannot revert a repository emission that arrived
+/// meanwhile — see [_emitRollback].
 class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
   /// Creates a new bloc instance.
   ///
@@ -406,8 +409,10 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
       PeopleListsMutationKind.deleteList,
       listId: event.listId,
     );
-    final previousLists = List<UserList>.of(state.lists, growable: false);
-    final previousIndex = _copyReverseIndex(state.listIdsByPubkey);
+    final removedIndex = state.lists.indexWhere(
+      (list) => list.id == event.listId,
+    );
+    final removedList = removedIndex == -1 ? null : state.lists[removedIndex];
 
     // Optimistically remove the list and any reverse-index entries.
     final optimisticLists = state.lists
@@ -424,6 +429,9 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
       ),
     );
 
+    List<UserList> undo(List<UserList> lists) =>
+        _restoreList(lists, removedList, removedIndex);
+
     try {
       final result = await _repository.deleteList(
         ownerPubkey: owner,
@@ -431,26 +439,14 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
       );
       if (!_resultStillApplies(mutation, owner)) return;
       if (!result.submitted) {
-        emit(
-          _withoutMutation(
-            state,
-            mutation.id,
-            failed: true,
-          ).copyWith(lists: previousLists, listIdsByPubkey: previousIndex),
-        );
+        _emitRollback(emit, mutation.id, undo);
         return;
       }
       emit(_withoutMutation(state, mutation.id, resultEventId: result.eventId));
     } catch (e, stackTrace) {
       addError(e, stackTrace);
       if (!_resultStillApplies(mutation, owner)) return;
-      emit(
-        _withoutMutation(
-          state,
-          mutation.id,
-          failed: true,
-        ).copyWith(lists: previousLists, listIdsByPubkey: previousIndex),
-      );
+      _emitRollback(emit, mutation.id, undo);
     }
   }
 
@@ -520,9 +516,6 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
       listId: listId,
       pubkey: pubkey,
     );
-    final previousLists = List<UserList>.of(state.lists, growable: false);
-    final previousIndex = _copyReverseIndex(state.listIdsByPubkey);
-
     final optimisticLists = _applyOptimisticAdd(
       state.lists,
       listId: listId,
@@ -540,6 +533,13 @@ class PeopleListsBloc extends Bloc<PeopleListsEvent, PeopleListsState> {
       ),
     );
 
+    List<UserList> undo(List<UserList> lists) => _applyOptimisticRemove(
+      lists,
+      listId: listId,
+      pubkey: pubkey,
+      now: _clock(),
+    );
+
     try {
       final result = await _repository.addPubkey(
         ownerPubkey: owner,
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz:
@hm21 I pushed a takeover follow-up commit (`a8db7f8d`) that:

- fixes the stale snapshot rollback comments in `PeopleListsBloc`
- adds the missing remove-after-teardown regression test
- updates the PR body to reflect the new coverage and links the adjacent per-mutation delete-outcome follow-up as #6696

Verification:

- `flutter analyze lib/features/people_lists/bloc/people_lists_bloc.dart test/features/people_lists/bloc/people_lists_bloc_test.dart`
- `flutter test test/features/people_lists/bloc/people_lists_bloc_test.dart`
- pre-push hook: changed-file analyze plus `test/features/people_lists/bloc/people_lists_bloc_test.dart` and `test/screens/user_list_people_screen_test.dart`
- GitHub checks are green on `a8db7f8d`

NotThatKindOfDrLiz (approved):
Approved after takeover. I re-reviewed the full PR diff after the follow-up commit: the teardown guard is applied consistently, rollback comments now match the current-state undo implementation, the remove path has dedicated teardown regression coverage, and CI is green.

mbradley:
@hm21 Review in progress on this one. Liz's approval is already in, so flagging before you act on it: I'm partway through a full pass and would like to land findings before this merges. Shouldn't be long.

hm21:
Sorry @mbradley  I'd already merged this about 15 minutes before your comment landed (19:05 UTC vs 19:22), so I never saw the flag in time. Branch protection here only requires a single approval and Liz's was already in, so nothing held it back.

Please do post the review anyway, I'll pick them up as a follow-up PR :)

mbradley:
@hm21 no apology needed, the timing is on me. i wanted to flag i was getting a review under way and then things got interrupted on the plane.

i'll finish out the review and post the findings here as a comment.

mbradley:
Follow-up material, per your note. None of this is a regression: I reproduced the sharp one against the merge-base and it behaves identically there.

**The order-independence claim is false, and it now sits in a source doc comment.** `_emitRollback`'s doc and the PR body both assert it. Add and Remove are separate `on<>` registrations, so `sequential()` does not order them relative to each other, and their undos are exact inverses applied to shared current state. Both settle orders on `l1 = [alice, bob]`, with a remove and its Undo both failing:

```
remove settles first -> [alice]        (bob gone, nothing was published)
add settles first    -> [alice, bob]   (correct)
```

The remove-then-Undo snackbar at `user_list_people_screen.dart:746` produces that interleaving in one tap, and the losing order is the likely one since the remove is issued first. It does not self-correct: a failed publish writes no cache, so `watchLists` never re-emits. Narrowing the claim is the cheap fix. Making it true is more machinery than the bug warrants.

Three further doc corrections in the same area:

- `_emitRollback` says the cache is written only after a successful publish. `syncOwner` also writes it from a relay snapshot, and the bloc runs `syncOwner` on every rewire.
- The class doc still states the failure-status invariant without the teardown carve-out.
- The toggle comment at `:476` holds toggle-vs-toggle but not toggle-vs-add/remove.

**One behaviour decision.** `deleteList` writes its tombstone before returning, so on the flag-off path the delete has usually succeeded. The result is dropped and the user is left on `_ListNotFoundView` with no confirmation and no pop. Your tests assert that, so I read it as intended. Should flag-off pop silently while account-switch stays put?

**Coverage gaps.** Moving `addError` below the teardown guard at any of the four catch sites leaves the suite green, so that ordering is not pinned. `_restoreList`'s non-zero-index insertion is never exercised, unlike its member-index counterpart. And `a failed remove restores the member at its old index` sits in the racing-emission group without setting up an emission.

Two things I checked that need no action: the guard's drop paths do not lose data, because `watchLists` emits the cache snapshot on listen, so a dropped result is stale rather than wrong. And the `result.submitted` fix is not latent: on base a create that reached no relay landed on `ready`, reporting a failed create as success.

NotThatKindOfDrLiz (approved):
## Summary

**Meets the bar.** The fix correctly closes all four teardown paths (`_onEnabledChanged(false)`, `_onOwnerChanged` via sign-out/account switch, and re-establishment). The guard logic is sound, the undo rewrite is strictly more robust than the snapshot approach it replaces, and test coverage is real — each new test was confirmed red first against the unfixed code. No new tech debt, all 15 CI checks green.

## Non-blocking findings

**Finding 1: Stale doc in `people_lists_bloc.dart:640` and `:644`**

The `_resultStillApplies` doc comment describes the snapshot rollback approach that commit 2 removed. Lines read:

> "Rollbacks **are** computed from a snapshot captured before the round-trip… **Restoring the captured snapshot** there hands one owner's lists to whoever is signed in by then."

But the bloc now computes rollbacks against current state, not snapshots. Also, the inline comment at `:215-217` says "…stops a mutation still in flight from putting **the snapshot** back."

**Recommendation:** Rephrase both to "a rollback computed against state that has since been torn down." This is a two-line fix worth doing before merge — the self-contradiction is within one file and a reader hitting `_resultStillApplies` first will be told the rollback works a way it no longer does.

**Finding 3: Pre-existing, adjacent — `user_list_people_screen.dart:90` reads global status as this delete's outcome**

In the `_onPeopleListsChanged` listener:

```dart
final failed = state.status == PeopleListsStatus.failure;
```

This conflates the delete's result with any pending mutation's status. If a delete succeeds while another mutation is pending and `status` is pinned to `failure`, the screen reports "Couldn't delete list" for a successful delete. The fix is narrow (needs failed member-mutation + second pending one + delete resolving between them) and untouched by this diff, but the PR rewrote the exact listener to distinguish "settled" from "dropped" — this is the remaining conflation.

**Recommendation:** File a follow-up issue (`#7XXX`) rather than fixing here — the status enum has no per-mutation resolution, and the debt note should live in tracking. Approved as-is; follow-up is non-blocking.

---

Both findings are non-blocking comments. Find 1 is low-risk and worth one pass before ship; Find 3 is known pre-existing and deserves a tracking issue.

PR #6668, opened 2026-08-04
fix(profile): hand pull-to-refresh a completer instead of a settled state

### LINKED ISSUE #6644: fix(profile): pull-to-refresh can still hang when a sync event is dropped mid-persist
Follow-up from the review of #6642. Pre-existing, not introduced by that PR — #6642 strictly reduces the hang surface (the settled-empty case went from a deterministic hang to fixed), but the underlying handshake is still fragile.

## The residual window

`ProfileGridView._refreshSyncedTabs()` holds the `RefreshIndicator` by awaiting `bloc.stream.firstWhere(...)` after dispatching a sync event. All five profile tab sync handlers use `transformer: droppable()`, so an event arriving while a handler is in flight is discarded — the handler never runs and no state is ever emitted.

The handlers stay "in flight" past their terminal emit, because each one does `await _persistSnapshot(...)` afterwards. So there is a window where the bloc is already settled (nothing left to emit) yet still swallows incoming sync events. A pull landing in that window produces zero emissions, and `firstWhere`'s `orElse` only fires when the stream closes (bloc disposal), so the future never resolves and the spinner hangs until the user leaves the screen.

The window is not microscopic: `CacheSync.write` → `CacheDaoImpl.write` runs `_ensureTotalBytes()` (a full-table `SELECT COALESCE(SUM(LENGTH(payload)),0)`) on the first write of a session, after every 256 writes, and after any `deletePrefix`, plus the upsert and a possible `evictOldest` transaction.

It is also self-inviting: `Future.wait` resolves at the terminal emit, so the spinner retracts while the bloc is still busy persisting — the user gets an immediate invitation to pull again, straight into the window.

## Why the obvious fix is not enough

Passing a `Completer` on the sync event (the pattern already used for `MyProfileRefreshRequested` four lines away in `_refreshProfileContent`) is immune to Equatable state suppression, but it does **not** fix this on its own: `droppable()` discards the event before the handler runs, so the completer would never be completed either. It would convert a suppressed-state hang into a never-completed-completer hang.

A real fix needs both halves:

1. Move the five sync handlers off `droppable()` — `sequential()` is the candidate, since every event then runs and every completer completes. This changes rapid-pull semantics (syncs queue instead of coalescing), which is the part that needs a deliberate decision.
2. Carry a `Completer<void>?` on each sync event and complete it in a `finally`, then have `_refreshSyncedTabs` await the completers instead of the state stream.

Alternatively, `unawaited(_persistSnapshot(...))` on the settle paths would shrink the window without changing concurrency semantics — cheaper, but it papers over the coupling rather than removing it.

## Scope

Touches 5 event classes, 5 blocs, `ProfileGridView`, and their tests. Deliberately kept out of #6642 to avoid turning a one-line guard fix into a concurrency change across every profile tab.

hm21:
Closes #6644

## Description

Pull-to-refresh on a profile could still spin forever, even after #6642 fixed the settled-empty case.

`ProfileGridView._refreshSyncedTabs()` held the `RefreshIndicator` by awaiting each tab bloc's next settled state. That handshake has two blind spots. An unchanged re-sync emits a state equal to the current one, which bloc suppresses. And every sync handler stays busy past its terminal emit while it awaits the snapshot write — a window in which the tab already looks settled but the handler is still running, and `droppable()` discarded any sync arriving there. No handler ran, nothing was emitted, and `firstWhere`'s `orElse` only fires when the stream closes (bloc disposal), so the future never resolved and the spinner ran until the user left the screen.

The window is not microscopic: `CacheSync.write` → `CacheDaoImpl.write` runs `_ensureTotalBytes()` (a full-table `SELECT COALESCE(SUM(LENGTH(payload)),0)`) on the first write of a session, after every 256 writes, and after any `deletePrefix`, plus the upsert and a possible `evictOldest` transaction. It was also self-inviting: the old wait resolved at the terminal emit, so the spinner retracted while the bloc was still persisting and handed the user an immediate invitation to pull again, straight into the window.

Each sync event now carries an optional `Completer<void>` that its handler completes in a `finally` — the pattern `MyProfileRefreshRequested` already uses four lines away in `_refreshProfileContent`. Because that `finally` runs after the snapshot write, the indicator is released exactly when the bloc can accept the next sync, which closes the self-inviting half too.

The completer alone does not fix this. `droppable()` discards the event before the handler runs, so the completer would never be completed either — a suppressed-state hang traded for a never-completed-completer hang. All five sync handlers therefore move to `sequential()`. That is the deliberate part of this change: rapid pulls now queue instead of coalescing. The alternative considered in the issue — `unawaited(_persistSnapshot(...))` on the settle paths — was rejected because it only shrinks the window rather than removing it, and it would let two writes race on the same cache key.

Bloc teardown is not a new hang path: `Bloc.close()` documents that pending events still finish processing, and I verified a queued `sequential()` event does reach its handler after close (it hits the `isClosed` early return, and the `finally` still completes the completer).

Comments was already the odd one out — not `droppable()`, but guarded by an in-handler `state.status == loading` early return. Under `sequential()` a queued sync can never enter with that status, because no handler leaves it there, so the guard is unreachable and is removed rather than left as dead code.

The `completeProfileTabSync` guard is shared in `lib/blocs/profile_shared/` so the contract is documented once instead of five times. The `_InMemoryCacheDao` write gate stays duplicated per test file, per the private-mocks rule in `.claude/rules/testing.md`.

Two existing tests pinned the behaviour this replaces and are updated rather than worked around:

- collabs' `droppable: ignores a second fetch while one is in flight` → `sequential: queues a second fetch behind one in flight`, now expecting both to run
- comments' `does not re-fetch when already loading` seeded a synthetic `loading` state that is now unreachable; it is dropped in favour of one asserting a sync dispatched mid-flight still runs

## The Videos tab

`_refreshProfileContent` awaits three things, and the first pass left one of them on the old mechanism: `ProfileFeedCubit` was still awaited through `stream.firstWhere` on a settled state while the profile and the five cached tabs handed back a completer.

That wait was not broken — `_doRefresh` flips `isRefreshing` true then false, so even a refresh that changes nothing has an observable edge. But it makes a UI progress flag load-bearing for the refresh handshake, which is precisely the coupling this branch exists to remove: the flag's only real job is driving the sticky cache-revalidation bar under the pinned tab bar. So `ProfileFeedRefreshRequested` now carries the same optional completer, fired from a `finally`, and all three participants use one mechanism.

`restartable()` stays. It is safe with a completer in a way `droppable()` is not: `switchMap` still invokes the handler for every event and only cancels the superseded one's *emitter*, so a refresh that loses the race still reaches its `finally`. A test pins that — flipping the transformer to `droppable()` turns it red.

The completer is deliberately left out of `props`. `ProfileFeedEvent` extends `Equatable` (unlike the five tab event bases), and that equality is what lets callers keep verifying `add(const ProfileFeedRefreshRequested())` regardless of which waiter is attached — the same convention `MyProfileRefreshRequested` already follows.

While in there, the four cache-backed tab blocs still justified their `isRefreshing` transition with "the profile pull-to-refresh (it awaits the next settled state) hangs", describing the handshake this branch removes. The transition itself stays; the comments now point at the reason it actually still has.

## Out of Scope

A sync and a `LoadMoreRequested` can still write the same cache key concurrently — they are separate event buckets, so no transformer on the sync bucket serialises them. That is pre-existing and untouched here.

## Verification

- Repro captured as a test first. Flipping saved back to `droppable()` turns both the bloc test and the widget test red; restoring `sequential()` turns them green.
- 5 bloc-level regression tests (saved / liked / reposted / collabs / comments) — each parks a handler in its snapshot write, dispatches a second sync into that window, and fails loudly if either completer is stranded.
- 2 cubit-level tests for the Videos tab: the completer fires once a refresh settles, and a refresh superseded by a newer one still fires its own. The second flips red under `droppable()`.
- 1 widget test on `ProfileGridView` pins the user-visible behaviour end to end, including the early-retract half: it asserts the first refresh is still pending while the write is parked, which the old stream-based wait would have failed.
- `flutter analyze lib test integration_test` — clean. `dart format` over `lib` and `test` — clean.
- `flutter test test/blocs test/widgets test/screens` — 7494 passing, 104 skipped. CI's four test shards, which run the merged-isolate `very_good test --optimization` build, are green on this head.
- `check_process_global_mutations.sh`, `check_test_unit_structure.sh`, `check_untested_services_floor.sh`, `check_shared_channel_overrides.sh` — all OK.
- Manually verified on a real Samsung Galaxy. Worth watching in review: the spinner now stays up marginally longer since it waits through the cache write, most noticeably on the first pull after launch when `_ensureTotalBytes()` runs.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_bloc.dart
+++ b/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_bloc.dart
@@ -1,12 +1,15 @@
 // ABOUTME: BLoC for managing profile collab videos grid
 // ABOUTME: Fetches Funnelcake-confirmed collaborator videos for a profile
 
+import 'dart:async';
+
 import 'package:bloc_concurrency/bloc_concurrency.dart';
 import 'package:cache_sync/cache_sync.dart';
 import 'package:equatable/equatable.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:models/models.dart' hide LogCategory;
+import 'package:openvine/blocs/profile_shared/profile_tab_sync_completion.dart';
 import 'package:openvine/blocs/profile_shared/profile_video_cursor_snapshot.dart';
 import 'package:openvine/extensions/video_event_extensions.dart';
 import 'package:unified_logger/unified_logger.dart';
@@ -35,7 +38,10 @@ class ProfileCollabVideosBloc
        super(const ProfileCollabVideosState()) {
     on<ProfileCollabVideosFetchRequested>(
       _onFetchRequested,
-      transformer: droppable(),
+      // Sequential, not droppable: a dropped event never reaches the handler,
+      // so nothing would complete its completer and pull-to-refresh would spin
+      // forever. Rapid fetches queue instead of coalescing.
+      transformer: sequential(),
     );
     on<ProfileCollabVideosLoadMoreRequested>(_onLoadMoreRequested);
   }
@@ -51,8 +57,22 @@ class ProfileCollabVideosBloc
   /// On reopen the persisted snapshot is served immediately; revalidation then
   /// re-fetches just the first page and prepends any genuinely-new collab
   /// videos — it never re-fetches the whole loaded feed.
+  ///
+  /// [ProfileCollabVideosFetchRequested.completer] fires only once all of that
+  /// is done, snapshot write included, so a refresh waiter is released exactly
+  /// when the bloc is free to take the next fetch.
   Future<void> _onFetchRequested(
     ProfileCollabVideosFetchRequested event,
+    Emitter<ProfileCollabVideosState> emit,
+  ) async {
+    try {
+      await _fetchAndRevalidate(emit);
+    } finally {
+      completeProfileTabSync(event.completer);
+    }
+  }
+
+  Future<void> _fetchAndRevalidate(
     Emitter<ProfileCollabVideosState> emit,
   ) async {
     // Flag the revalidation whenever a settled result is on screen — an empty
--- a/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_bloc.dart
+++ b/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_bloc.dart
@@ -76,8 +76,8 @@ class ProfileCollabVideosBloc
     Emitter<ProfileCollabVideosState> emit,
   ) async {
     // Flag the revalidation whenever a settled result is on screen — an empty
-    // one included — so a re-sync always has an observable start edge, matching
-    // the liked/reposted/saved tabs.
+    // one included — so the sticky cache-revalidation bar runs for the whole
+    // re-sync, matching the liked/reposted/saved tabs.
     if (state.status != ProfileCollabVideosStatus.initial &&
         !state.isRefreshing) {
       emit(state.copyWith(isRefreshing: true));
--- a/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_event.dart
+++ b/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_event.dart
@@ -12,7 +12,12 @@ sealed class ProfileCollabVideosEvent {
 ///
 /// This triggers the repository's Funnelcake-confirmed collaborator read path.
 final class ProfileCollabVideosFetchRequested extends ProfileCollabVideosEvent {
-  const ProfileCollabVideosFetchRequested();
+  const ProfileCollabVideosFetchRequested({this.completer});
+
+  /// Optional completion signal for UI refresh affordances, completed once the
+  /// handler is fully done — snapshot write included. See
+  /// [completeProfileTabSync].
+  final Completer<void>? completer;
 }
 
 /// Request to load more collab videos (pagination).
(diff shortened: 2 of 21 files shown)


realmeylisdev (approved):
Approving.

Checked the load-bearing claim against `bloc 9.2.1` / `bloc_concurrency 0.3.0` rather than taking it on trust. `restartable()` is `switchMap`, and `_Emitter.cancel()` sets `_isCanceled` but *not* `_isCompleted` — so `call()` silently no-ops instead of tripping the assert, and the superseded refresh really does reach its `finally`. Teardown holds too: `close()` awaits `_eventController.close()` before `super.close()`, so a queued `sequential()` event drains while `_stateController` is still open. No throw, completer fires.

The comments guard is dead rather than load-bearing, as you say — `_sync` is the only writer of `ProfileCommentsStatus.loading`, and it always leaves `success` or `failure`.

Mutation-checked the new tests locally; each goes red for its own reason:

- liked → `droppable()`: `a sync arriving during the snapshot write still runs`
- saved → `droppable()`: same, plus the `ProfileGridView` widget test
- feed → `droppable()`: `a superseded refresh still fires its own completer`
- comments → `concurrent()` + guard restored: `Expected: <2> Actual: <1>`

`flutter analyze lib test integration_test` clean, `dart format` clean across all 21 files, the 7 affected suites green.

Two things I looked at and won't hold this on. `sequential()` does let a duplicate sync actually run where `droppable()` swallowed it, but the only live burst source is `requestProfileSavedVideosSyncIfAvailable` on bookmark toggle — and running each one there is the more correct outcome, not the worse one. `refreshNotifier` is never bumped anywhere in `lib`, and `RefreshIndicator` serialises pulls, so `_refreshSyncedTabs` can't overlap itself. Separately, neither the old wait nor the completer has a timeout, so a repository call that never resolves still spins — unchanged by this PR.

PR #6669, opened 2026-08-04
fix(profile): stop the avatar placeholder shimmering through the skeleton

### LINKED ISSUE #6670: fix(profile): stop the avatar placeholder shimmering through the skeleton
Auto-created from PR #6669.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6669
**Branch:** `fix/profile-skeleton`

hm21:
Closes #6670

## Description

The profile avatar's loading state rendered as a lopsided dark blob with a hard diagonal edge and the ghost of a person silhouette sweeping through it.

Cause: `Skeletonizer` only repaints *leaf* shapes with the shimmer shader — every container that has children keeps its own decoration. `UserAvatar`'s generated placeholder is exactly that shape: a gradient tile with head and body ellipses stacked on top, so the sweep landed on the silhouette while the tile below it kept its accent colours. The result read as a rendering glitch rather than as "not loaded yet".

While the identity is still resolving, the avatar is now swapped for a flat tile in the avatar's own geometry, with the animated brand mark centred on it. `Skeleton.keep` — not `Skeleton.ignore` — on the mark, because `ignore` paints nothing while the enclosing `Skeletonizer` is enabled, which is the only window in which this widget is on screen at all.

The identity `Skeletonizer` also gets `enableSwitchAnimation: true`, so the hand-off from skeleton to the loaded avatar and name cross-fades instead of popping.

Three notes on the implementation:

- The swap is driven by the same `showIdentitySkeleton` flag the `Skeletonizer` gets, **not** by `Skeleton.replace`. `Skeleton.replace` resolves against the enclosing `Skeletonizer`'s scope, and that scope sits above the `AnimatedSwitcher` that `enableSwitchAnimation` installs — so the outgoing half of the cross-fade rebuilds into the real avatar too, mounting `profile_avatar_hero_<hex>` twice for the length of the animation. That is long enough for a route push in that window (tapping the avatar to open the lightbox, for one) to trip Hero's duplicate-tag assert, and it also meant the bone popped out on the first frame instead of fading. The pending-action label is on the same flag for the same reason.
- `UserAvatar`'s default corner-radius formula moved into a `cornerRadiusForSize` static so the stand-in matches real avatar geometry instead of duplicating `min(size * 0.4, 56)` at a second callsite. Pure refactor, same math.
- The tile sits in a `Positioned.fill`: a childless `DecoratedBox` takes the smallest size its constraints allow, and `Stack` hands non-positioned children loose ones, so without it the tile collapses to nothing and only the brand mark renders.

**Related Issue:** 

## Out of Scope

Other surfaces that render `UserAvatar` under a skeleton (notification rows, user tiles, comment rows) hit the same Skeletonizer behaviour, but at 36–44 px the artefact is barely legible. Left untouched rather than widened into a sweep across every avatar callsite.

## Verification

- `flutter test test/widgets/profile test/widgets/comprehensive_user_avatar_test.dart test/unit/user_avatar_tdd_test.dart test/widgets/user_avatar_svg_test.dart` — 325 passing, including three new cases: the real `UserAvatar` is out of the tree while the skeleton is on, it is back once the profile resolves, and there are never two avatar `Hero`s mounted at once while the skeleton cross-fades out. The last one fails against the `Skeleton.replace` version with `Actual: <2>`.
- Manually verified on a real Samsung Galaxy device (SM-S942B), including the Hero / cross-fade follow-up: the bone now fades into the loaded avatar instead of popping, and opening the lightbox immediately after the profile resolves no longer trips Hero's duplicate-tag assert

## Type of Change

### DIFF
--- a/mobile/lib/widgets/profile/profile_header_media.dart
+++ b/mobile/lib/widgets/profile/profile_header_media.dart
@@ -376,23 +376,29 @@ class _ProfileAvatarWithColor extends StatelessWidget {
       placeholderSeed: userIdHex,
       size: avatarSize,
     );
-    final avatar = hasAvatar
-        ? GestureDetector(
-            onTap: () => _showAvatarLightbox(
-              context,
-              imageUrl: imageUrl,
-              userIdHex: userIdHex,
-            ),
-            child: Hero(
-              tag: _avatarHeroTag(userIdHex),
-              flightShuttleBuilder: (_, _, _, _, _) => _AvatarHeroFlightShuttle(
+    final avatar = Skeleton.replace(
+      width: avatarSize,
+      height: avatarSize,
+      replacement: const _AvatarBone(size: avatarSize),
+      child: hasAvatar
+          ? GestureDetector(
+              onTap: () => _showAvatarLightbox(
+                context,
                 imageUrl: imageUrl,
                 userIdHex: userIdHex,
               ),
-              child: avatarWidget,
-            ),
-          )
-        : avatarWidget;
+              child: Hero(
+                tag: _avatarHeroTag(userIdHex),
+                flightShuttleBuilder: (_, _, _, _, _) =>
+                    _AvatarHeroFlightShuttle(
+                      imageUrl: imageUrl,
+                      userIdHex: userIdHex,
+                    ),
+                child: avatarWidget,
+              ),
+            )
+          : avatarWidget,
+    );
 
     if (pendingActions.isEmpty) return avatar;
 
@@ -416,6 +422,49 @@ class _ProfileAvatarWithColor extends StatelessWidget {
               ),
             ),
           ),
+        ),
+      ],
+    );
+  }
+}
+
+/// Stand-in painted in place of the avatar while the identity loads: a flat
+/// tile in the avatar's own geometry with the brand mark animating on top.
+///
+/// [Skeletonizer] repaints leaf shapes with the shimmer but keeps the
+/// decoration of every container that has children, so shimmering the real
+/// [UserAvatar] leaves its generated placeholder — a gradient tile with a
+/// person silhouette cut into it — showing through the sweep.
+class _AvatarBone extends StatelessWidget {
+  const _AvatarBone({required this.size});
+
+  final double size;
+
+  @override
+  Widget build(BuildContext context) {
+    return Stack(
+      alignment: Alignment.center,
+      children: [
+        // Positioned.fill because a childless DecoratedBox takes the smallest
+        // size its constraints allow, and Stack hands non-positioned children
+        // loose ones — the tile would collapse to nothing.
+        Positioned.fill(
+          child: Skeleton.leaf(
+            child: DecoratedBox(
+              decoration: BoxDecoration(
+                color: context.vineColors.skeleton,
+                borderRadius: BorderRadius.circular(
+                  UserAvatar.cornerRadiusForSize(size),
+                ),
+              ),
+            ),
+          ),
+        ),
+        // Skeleton.keep, not .ignore: `ignore` paints nothing while the
+        // enclosing Skeletonizer is enabled, which is the only time this
+        // widget is on screen.
+        Skeleton.keep(
+          child: BrandedLoadingIndicator(size: min(size * 0.8, 64)),
         ),
       ],
     );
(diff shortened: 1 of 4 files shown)


mbradley:
@hm21 Review in progress on this one, findings to follow shortly.

mbradley (approved):
Approving. One patch offered below, the rest are questions.

**The bone is not excluded from the semantics tree.** Skeletonizer suppresses hit-testing but never semantics, so the loading window stays reachable by a screen reader. Swapping `UserAvatar` for the bone therefore replaces a node labelled "User avatar" with two unlabelled image nodes from the brand mark's sprite, leaving nothing at that position that conveys a loading state.

I have the fix and a test locally, verified: analyze clean, 326 green, goldens pass, and the test fails 6 vs 4 without the exclusion. Take it or write your own.

<details>
<summary><code>mobile/lib/widgets/profile/profile_header_media.dart</code></summary>

```dart
  @override
    // The bone is decorative: it carries no identity a screen reader can use,
    // and the brand mark's sprite would otherwise announce as an unlabelled
    // image. Skeletonizer suppresses hit-testing but never semantics, so
    // without this the loading window is announced as bare image nodes.
    return ExcludeSemantics(
      child: SizedBox.square(
        dimension: size,
        child: Stack(
          alignment: Alignment.center,
          children: [
            // ...unchanged...
          ],
        ),
      ),
    );
  }
```

</details>

<details>
<summary><code>mobile/test/widgets/profile/profile_header_widget_test.dart</code> (needs <code>import 'package:flutter/semantics.dart';</code>)</summary>

```dart
      testWidgets('the avatar bone adds no semantics while the identity '
          'skeleton is enabled', (tester) async {
        final handle = tester.ensureSemantics();

        int imageNodeCount() {
          var count = 0;
          void visit(SemanticsNode node) {
            if (node.getSemanticsData().flagsCollection.isImage) count++;
            node.visitChildren((child) {
              visit(child);
              return true;
            });
          }

          visit(tester.getSemantics(find.byType(MaterialApp)));
          return count;
        }

        await tester.pumpWidget(
          buildTestWidget(
            userIdHex: testUserHex,
            isOwnProfile: true,
            myProfileState: const MyProfileInitial(),
          ),
        );
        await tester.pump();
        expect(findIdentitySkeletonizer(tester).enabled, isTrue);
        expect(find.byType(BrandedLoadingIndicator), findsOneWidget);
        final whileLoading = imageNodeCount();

        await tester.pumpWidget(
          buildTestWidget(
            userIdHex: testUserHex,
            isOwnProfile: true,
            myProfileState: MyProfileLoaded(
              profile: createTestProfile(displayName: 'Loaded'),
              isFresh: true,
            ),
          ),
        );
        await tester.pumpAndSettle();
        expect(findIdentitySkeletonizer(tester).enabled, isFalse);
        final whenLoaded = imageNodeCount();

        expect(
          whileLoading,
          equals(whenLoaded),
          reason:
              'Skeletonizer suppresses hit-testing but never semantics, so '
              "the brand mark's sprite announces as unlabelled image nodes "
              'unless the bone is excluded. The loading window must not add '
              'semantics where the avatar previously carried a label.',
        );

        handle.dispose();
      });
```

</details>

Whether the loading state should instead carry its own label is a copy decision and needs an ARB key across the locales, so I left that alone.

**The Out of Scope list is wrong in both directions.** Notification rows and user tiles never sit under a `Skeletonizer`, so they cannot hit this. The share-sheet contacts row (`share_with_section.dart:65`) does and is not listed. Both real cases go through `IdentitySkeletonizer`, which already hardcodes this same Skeletonizer config, including `enableSwitchAnimation: true`. That makes it the candidate owner. Note `_AvatarBone` cannot be reused there as written: `_markSize` is a fixed 64, which is 178% of a 36px tile, and the class is library-private.

**The flag gates on a profile existing, not on avatar readiness.** A cached Kind 0 with a name but no `picture` clears the skeleton and mounts the real `UserAvatar` with a null URL, which renders the generated placeholder the bone exists to avoid. Accept, gate the avatar separately, or document the boundary?

**Reduce motion.** skeletonizer 2.1.2 has no `disableAnimations` handling, so `enableSwitchAnimation: true` cross-fades regardless. The stats-row Skeletonizer at `profile_header_media.dart:306` is already ungated on `main`, so this follows precedent. #6686 honours reduce motion in this same subtree, so after both land the header handles it two ways. Gate your line, both, or neither?

PR #5279, opened 2026-06-17
feat(profile): cache video tabs with stale-while-revalidate

hm21:
Closes #5280

## Description

Reopening a profile tab reloaded its whole list from the relay every time, fast scrolling stalled on "loading more", and freshly-arriving data could briefly jank the UI thread. This backs the profile video tabs with `CacheSync` so each tab's scrolled-through list is restored **instantly** on reopen, then revalidated in the background behind a thin progress bar pinned **under the tab bar**.

Tabs covered: **Liked, Reposts, Saved, Collabs**.

**How it works**
- **Shared snapshot models** persist the resolved window so a reopen renders from cache without a relay round-trip:
  - `ProfileVideoListSnapshot` — ID-list tabs (Liked, Reposts, Saved)
  - `ProfileVideoCursorSnapshot` — cursor-paginated tab (Collabs)
- **Stale-while-revalidate**: the snapshot is served immediately; revalidation only refreshes the ID list (or first page for Collabs) and **reconciles cheaply** — drops removed items and fetches only the genuinely-new ones at the top. The UI thread never re-resolves the whole window (fixes the jank when fresh data arrived).
- **Pagination** prefetches ~1.5 viewports ahead (overridable `ScrollPaginationMixin` threshold) and persists each page, so fast scrolling and re-scrolling no longer stall on "loading more". The Liked tab keeps `main`'s sparse-ID page-filling + blocklist filtering.
- **Sticky bar**: a single revalidation bar in the pinned tab-bar header tracks whichever cached tab is active (a body overlay would be painted behind the pinned header).
- Plumbing: `VideoEvent.fromJson` (already on `main`) is reused; `CacheSync.read`/`write` (already on `main`) drive incremental list growth.

## Out of Scope

- **Comments tab** — intentionally left on the existing path (different data model, cursor-based).
- This branch was rebased onto `main`, which had landed a parallel refactor of the Liked tab (`_fetchVideoPage` sparse-ID page-filling + blocklist-in-fetch). That refactor is integrated into the new SWR paths here; the Liked emit flow is now SWR (no intermediate `syncing`/`loading`), so a few of those tests were updated accordingly.

## Verification

- `flutter analyze lib test integration_test` — clean
- Bloc tests: Liked / Reposts / Saved / Collabs (incl. reopen-restore, reconcile/prepend, droppable, sparse-ID + blocklist) — green
- Shared snapshot tests, profile widget tests, `scroll_pagination_mixin` test — green
- `cache_sync` and `models` package tests — green
[...]

PR #6671, opened 2026-08-04
fix(profile): prefetch profile tab pages before the spinner is reachable

### LINKED ISSUE #6672: fix(profile): prefetch profile tab pages before the spinner is reachable
Profile tab pagination can expose the loading-more spinner continuously because the next page is requested too close to the bottom and some profile tab pages are too short to outrun the prefetch threshold.

## Problem

- `ScrollPaginationMixin` used a fixed 200px load-more threshold by default.
- Dense profile grids can scroll through that distance before the next page is fetched and rendered.
- Liked / Reposts / Saved / Collabs had duplicated viewport-relative overrides, but Videos and Comments still inherited the fixed default.
- Cached profile tabs fetched only 18 items per page, roughly one phone viewport in a 3-column grid, so a 1.5-viewport prefetch could immediately chase the bottom again.
- Liked videos also had an offset bug where resolved videos after a page filled could be skipped if they were later in the current fetched batch.

## Desired Behavior

- Profile tab pagination should request the next page before the spinner is reachable in normal scrolling.
- Shared pagination policy should avoid copy-pasted profile grid thresholds/page sizes.
- Existing paginating screens that inherit the mixin should remain compatible with the viewport-relative default.
- Liked-videos pagination must not skip resolved videos when a page fills mid-batch.

## Acceptance Criteria

- `ScrollPaginationMixin` defaults to a viewport-relative prefetch threshold.
- Cached profile tab grids use a shared page size large enough to cover more than the prefetch distance.
- Hashtag search uses a page size that is not shorter than the inherited prefetch behavior.
- Liked-videos pagination advances only past IDs actually consumed when a page fills.
- Regression tests cover the page-size policy and liked-video overflow behavior.
- `flutter analyze lib test` and affected tests pass.

## PR

Implemented by #6671.

hm21:
Closes #6672

## Problem

Scrolling profile tabs could keep the loading-more spinner visible. Two independent causes made the prefetch lose the race.

**Trigger fired too late.** `ScrollPaginationMixin` used a fixed 200px distance from the bottom. In a 3-column grid that is barely a row and a half, so the next page was only requested once the user had effectively arrived at the bottom. Liked / Reposts / Saved / Collabs already worked around this with identical 1.5-viewport overrides (#5279); Videos and Comments were missed and still used the default.

**Cached-tab pages were about one screen tall.** The cached tabs fetched 18 items per page. On a 360dp phone a tile is about 117px and a row about 121px, so 18 items is roughly one viewport. Each page pushed the end of the list forward by about one screen while the prefetch fires 1.5 screens ahead, so the grid never got enough headroom.

## Change

Make the viewport-relative prefetch distance the `ScrollPaginationMixin` default and drop the four duplicated profile-grid overrides.

Raise the cached profile tab page size from 18 to `ProfileTabPagination.pageSize` (36), grouped as a shared profile pagination policy value.

Keep the prefetch default safe for non-profile consumers:

- Videos and Comments profile tabs already fetch larger pages (the Videos tab goes through `getAuthorFeed` / `FunnelcakeApiClient.getVideosByAuthor`, which defaults to 50; `ProfileCommentsBloc` uses 50).
- Hashtag search now explicitly fetches 50 chips per page so the inherited viewport-relative threshold does not chain-load short chip pages.
- The default threshold getter now returns `0` before a scroll position attaches, so protected overrides that read `super.paginationLoadMoreThreshold` outside the guarded listener path do not throw.

Fix a liked-videos pagination bug exposed by the larger page size: `_fetchVideoPage` no longer advances past resolved videos left in a batch after the page fills. Those videos remain available to the next load-more instead of being skipped permanently.

## Why a shared constant

The cached profile page size is a function of the shared prefetch distance, not a per-tab preference. Keeping it in `ProfileTabPagination.pageSize` makes the policy visible and prevents the four cached tabs from drifting apart.

## Note on page size and latency

A larger cached profile page does not add relay round trips by itself. `_fetchVideoPage` loops until it has filled a page, so the number of round trips tracks the ID resolution rate, not only the page size.

Load-more on the relay-backed tabs can still be slow for an unrelated reason outside this diff: relay queries do not complete on EOSE and run to their full timeout. That is app-wide, not profile-specific, and is being handled separately.

## Test plan

- `flutter test test/blocs/profile_liked_videos/profile_liked_videos_bloc_test.dart test/blocs/profile_reposted_videos/profile_reposted_videos_bloc_test.dart test/blocs/profile_saved_videos/profile_saved_videos_bloc_test.dart test/blocs/profile_collab_videos/profile_collab_videos_bloc_test.dart test/blocs/hashtag_search/hashtag_search_bloc_test.dart test/mixins/scroll_pagination_mixin_test.dart` — passing
- Pre-push hook: merge-conflict check, analyzer, and changed-file tests — passing
- Bloc tests express expectations in terms of `ProfileTabPagination.pageSize`; liked-videos tests include a regression for the previously skipped overflow videos.
- Original manual check from the author: real Samsung Galaxy scrolling Videos and Likes no longer parks on the loading indicator.

### DIFF
--- a/mobile/lib/blocs/hashtag_search/hashtag_search_bloc.dart
+++ b/mobile/lib/blocs/hashtag_search/hashtag_search_bloc.dart
@@ -22,8 +22,9 @@ EventTransformer<E> _debounceRestartable<E>() {
   };
 }
 
-/// Number of results per page for hashtag search pagination.
-const _pageSize = 20;
+/// Hashtag chips are visually short, so each page needs to add enough rows to
+/// stay ahead of `ScrollPaginationMixin`'s viewport-relative prefetch.
+const _pageSize = 50;
 
 /// BLoC for searching hashtags via the Funnelcake API.
 ///
@@ -100,7 +101,10 @@ class HashtagSearchBloc extends Bloc<HashtagSearchEvent, HashtagSearchState> {
     _feedTracker?.startFeedLoad('hashtag_search');
 
     try {
-      final results = await _hashtagRepository.searchHashtags(query: query);
+      final results = await _hashtagRepository.searchHashtags(
+        query: query,
+        limit: _pageSize,
+      );
 
       _feedTracker?.markFirstVideosReceived('hashtag_search', results.length);
 
@@ -140,6 +144,7 @@ class HashtagSearchBloc extends Bloc<HashtagSearchEvent, HashtagSearchState> {
     try {
       final moreResults = await _hashtagRepository.searchHashtags(
         query: state.query,
+        limit: _pageSize,
         offset: state.offset,
       );
 
--- a/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_bloc.dart
+++ b/mobile/lib/blocs/profile_collab_videos/profile_collab_videos_bloc.dart
@@ -9,6 +9,7 @@ import 'package:equatable/equatable.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:models/models.dart' hide LogCategory;
+import 'package:openvine/blocs/profile_shared/profile_tab_page_size.dart';
 import 'package:openvine/blocs/profile_shared/profile_tab_sync_completion.dart';
 import 'package:openvine/blocs/profile_shared/profile_video_cursor_snapshot.dart';
 import 'package:openvine/extensions/video_event_extensions.dart';
@@ -17,9 +18,6 @@ import 'package:videos_repository/videos_repository.dart';
 
 part 'profile_collab_videos_event.dart';
 part 'profile_collab_videos_state.dart';
-
-/// Number of videos to load per page for pagination.
-const _pageSize = 18;
 
 /// BLoC for managing profile collab videos.
 ///
@@ -141,7 +139,7 @@ class ProfileCollabVideosBloc
 
     final videos = await _videosRepository.getCollabVideos(
       taggedPubkey: _targetUserPubkey,
-      limit: _pageSize,
+      limit: profileTabPageSize,
     );
     if (isClosed) return;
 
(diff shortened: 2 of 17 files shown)


NotThatKindOfDrLiz:
@hm21 I took over the last mile on this PR and pushed a rebased branch on top of current `origin/main`.

What changed:

- Resolved the merge conflicts with the current profile-tab sync completion changes.
- Kept the viewport-relative `ScrollPaginationMixin` default and the 36-item cached profile tab page size.
- Moved the profile page-size policy into `ProfileTabPagination.pageSize`.
- Fixed `_fetchVideoPage` so Liked videos resolved after a page fills are not skipped by advancing past the whole batch.
- Raised hashtag search pagination to 50 explicit results per page so the inherited viewport-relative threshold does not chain-load short chip pages.
- Hardened `paginationLoadMoreThreshold` to return `0` before a controller has clients.
- Updated the PR body and tests, including a regression for the liked-video overflow case.

Verification:

- `flutter analyze lib test`
- `flutter test test/blocs/profile_liked_videos/profile_liked_videos_bloc_test.dart test/blocs/profile_reposted_videos/profile_reposted_videos_bloc_test.dart test/blocs/profile_saved_videos/profile_saved_videos_bloc_test.dart test/blocs/profile_collab_videos/profile_collab_videos_bloc_test.dart test/blocs/hashtag_search/hashtag_search_bloc_test.dart test/mixins/scroll_pagination_mixin_test.dart`
- Pre-push hook: merge-conflict check, analyzer, and changed-file tests

Please review the takeover commit (`a060033ca`) and merge when you are ready, assuming CI stays green.

NotThatKindOfDrLiz (approved):
Approved after takeover commit a060033ca. CI is green, the branch is mergeable, and the remaining review findings were addressed: liked-video overflow is preserved, hashtag pages now use a 50-item limit, the no-client threshold getter is hardened, and the shared profile page-size constant is grouped.

PR #6673, opened 2026-08-04
fix(nostr): stop a stalled relay from taxing every query with its timeout

### LINKED ISSUE #6679: fix(nostr): stop a stalled relay from taxing every query with its timeout
Auto-created from PR #6673.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6673
**Branch:** `fix/relay-query-completion-stall`

hm21:
Closes #6679

## Problem

Relay queries did not complete on EOSE — they ran to their full 5s timeout, app-wide. On-device measurement while paginating a profile:

```
relay query total=5481ms timedOut=true events=344 kinds=[[7]]
relay query total=5402ms timedOut=true events=1   kinds=[[0]]
```

344 events had already arrived and the caller still waited the whole budget. Because `NostrClient` caps concurrent one-shot REQs at 6, the stalled queries then jammed the pool for everything behind them (`poolWait` up to 4926ms, pushing individual queries past 9s). A single liked-tab page spent **24.7s** across five batches of ~5.5s each.

## Root cause

Instrumenting the timeout to report which relays still owed a terminal frame named it directly:

```
still owed by: [wss://relay.divine.video(conn=2, authed=false,
                alwaysAuth=false, handshake=false, blocks=true)]
repair wss://relay.divine.video skipped=inboundActive
```

`relay.divine.video` is connected, behind no auth gate, and receiving inbound frames the whole time — it simply never sends `EOSE` or `CLOSED` for some REQs. It is not a dead socket, which is why the existing zombie remediation correctly declines to cycle it.

`RelayPool` released a query's caller only once **no** relay still had it saved. With one relay permanently silent, that condition never held, so every query in the app ran to its full timeout no matter how fast the healthy relays answered.

Two narrower variants of the same conflation also existed: a saved query can mean the REQ is live on the socket, or that it is parked for replay after NIP-42 or a reconnect. Only the first is worth waiting for — a parked query is re-issued far outside the caller's timeout.

## Change

Six commits. The first three are the fix, in the order the evidence narrowed things down; the last three are review passes over them.

**1. Completion accounting.** A relay counts as settled once it can no longer produce a terminal frame — socket down, or auth-gated with no NIP-42 handshake outstanding. Waiting through an *outstanding* handshake is preserved, so the intended post-AUTH replay still reaches the original caller; that is deliberate and pinned by `relay_pool_closed_frame_test.dart`. Since a relay going quiet does not itself fire the completion check, the sweep also runs when a socket drops and when an AUTH is refused.

**2. Zombie remediation from queries.** A caller only abandons a query by timing out, so any relay still holding it demonstrably did not answer — feed those into the existing `_repairSilentRelays` on unsubscribe. Previously only the publish path could trigger it.

**3. Query settle window — this is the one that fixes the measured case.** Neither of the above applies to `relay.divine.video`: it is connected and receiving frames throughout, it just never sends `EOSE` for some REQs. The remediation correctly declines (`skipped=inboundActive`), because that connection is alive by any test available.

A connected relay that never terminates a REQ emits no signal at all, so time is the only bound. The publish side already solved exactly this — `PublishTracker.settleWindow` grants the remaining relays a grace period after the first answer, documented as *"the settle window bounds the cost of one wedged relay: healthy relays answer in single-digit milliseconds, so waiting the full timeout for a silent sibling would stall every publish behind the slowest connection."* Queries had no equivalent.

The same window is now armed on the first terminal frame for a query. That frame is the proof the fan-out is being served, so the stragglers get a bounded grace period instead of the caller's whole budget. Queries where *no* relay has answered yet are untouched and still get the full timeout.

Tradeoff, identical to the publish side: a healthy but unusually slow relay can lose its events to the window. One second is orders of magnitude above observed relay latency.

**4. Review pass.** Five findings in the completion path the three commits above introduce.

Query completion used to imply every relay had already dropped the query, so the caller returning was also the end of the subscription. Both new early-completion paths break that, and `queryEventsDetailed` only unsubscribes on its timeout branch — so an abandoned REQ stayed live for the life of the socket, against the relay's concurrent-subscription cap, pinning the caller's event box, and re-issued by every post-AUTH and zombie-reconnect replay. On a relay that never sends a terminal frame that is one leaked subscription per query for the whole session. Completion now tears the query down on every relay that still holds it.

That also restores the zombie remediation's reach: it runs from the teardown rather than only from unsubscribe, so the straggler the settle window walks away from is still offered to the repair. Previously the window resolved the caller first and the repair only fired when no relay had answered at all.

A terminal frame arriving while the REQ fan-out was still running never armed the settle window — measured at the full timeout with two relays where the fast one EOSEs during the slow one's REQ write, the exact stall this path exists to stop. "Some relay answered" is now sticky per query rather than a property of the call that observed the frame.

Auth-gate accounting rested on a signal that could only say "no handshake right now", which is not evidence the gate is shut: `alwaysAuth` latches for the session, so a relay that just reconnected sits in that state for the round trip before its challenge arrives, and writing it off there drops every read-gated relay's results on cold start. The gate is now recorded shut only on evidence the relay produced, and both markers are dropped on every connection-state transition and when the relay leaves the pool — so a relay that swallows an AUTH event cannot stay marked "handshake outstanding" for the rest of the session.

Temp relays serve one-shot queries like any other relay but were never wired to the status observer, so a dropped temp socket never ran the sweep that observer exists for.

**5. Review: the window has to bound silence, not progress.** A relay part-way through a result set had its REQ closed one second after any sibling EOSE'd, and the caller was still told the query completed normally — a truncated page indistinguishable from a complete one. An `EVENT` for a one-shot query now restarts the window; a relay that says nothing is bounded exactly as before.

**6. Review: the same distinction for an outstanding NIP-42 handshake.** A relay mid-handshake is being served too — it challenged us, and its `OK` is what runs the post-AUTH replay the query is parked for. The window armed over that branch and the teardown dropped the parked query, so a read-gated relay contributed nothing to any query a sibling answered first, and the claim in commit 1 above did not hold. The window now re-arms while a straggler is inside its handshake, bounded by `authHandshakeWindow` so a relay that swallows our AUTH event cannot hold the query for the life of the connection.

## What this does not do

It does not make `relay.divine.video` answer. Not sending `EOSE` for a REQ is a NIP-01 violation and needs fixing in the relay; this is the client-side containment so one misbehaving relay costs ~1s instead of the full timeout on every query in the app.

## Test plan

- `relay_pool_query_settle_test.dart` — the settle window bounds a relay that never answers once another has; an auth-gated relay with no handshake does not hold the query; a rejected AUTH releases parked queries; a dropped socket does not hold the query; a healthy connected relay **does** still hold until its EOSE; the settle window closes the REQ it abandons; a terminal frame landing during fan-out still arms the window; an auth-gated relay that has not been challenged yet still holds the query; and a reconnect gives a relay whose gate was shut a fresh handshake. Every failure case was verified to reproduce without its fix, each burning the full test timeout
- `relay_pool_query_zombie_test.dart` — a relay that swallows the REQ is force-reconnected; a straggler the settle window abandons is force-reconnected too; one that EOSEs is left alone; one that is inbound-active on other traffic is left alone. Verified to fail without the fix
- `flutter test` in `packages/nostr_sdk` — 438 pass, including the existing post-AUTH replay, CLOSED-frame and publish-side zombie regression tests
- `flutter test` in `packages/nostr_client` — 324 pass
- `flutter analyze packages/nostr_sdk packages/nostr_client lib` — clean

- Manually verified on a real Samsung Galaxy: the repeated 5s `timedOut=true` on `relay.divine.video` no longer stalls the query path.

### DIFF
--- a/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
+++ b/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
@@ -124,6 +124,11 @@ class RelayPool {
   /// a removed or replaced relay is not retained by the bookkeeping.
   final Expando<bool> _statusObserved = Expando<bool>('relayStatusObserved');
 
+  /// When each one-shot query's REQ fan-out started, so an abandoned query can
+  /// ask [Relay.isSilentSince] whether a relay that never answered had gone
+  /// quiet altogether. Dropped when the query settles or is unsubscribed.
+  final Map<String, DateTime> _querySentAt = {};
+
   /// Track publishes awaiting OK confirmations (per event id).
   final Map<String, PublishTracker> _pendingPublishes = {};
 
@@ -706,6 +711,7 @@ class RelayPool {
     }
 
     _queryCompleteCallbacks.remove(subId);
+    _querySentAt.remove(subId);
     callback();
   }
 
@@ -729,6 +735,11 @@ class RelayPool {
     // A dropped socket cannot deliver the terminal frame for a REQ that was
     // written to it; the reconnect path re-issues the REQ on a fresh socket.
     if (relay.relayStatus.connected != ClientConnected.connected) return false;
+
+    // A connection being force-cycled as a zombie reports `connected`
+    // throughout, so it would otherwise keep charging concurrent queries the
+    // full timeout for the whole repair.
+    if (_silentRelayRepairsInFlight.contains(relay.url)) return false;
 
     // Behind an auth gate the replay only happens if NIP-42 completes, so wait
     // only while a handshake is actually outstanding. Without one — signing
@@ -757,6 +768,32 @@ class RelayPool {
       previous?.call();
       _settleQueriesBlockedBy(relay);
     };
+  }
+
+  /// Force-cycles the relays that never produced a terminal frame for the
+  /// abandoned query [subId].
+  ///
+  /// A caller only abandons a query by timing out, so any relay still holding
+  /// it demonstrably did not answer. When that same connection also received no
+  /// inbound frame of any kind since the REQ was written, it is the half-open
+  /// zombie [_repairSilentRelays] already remediates on the publish side — a
+  /// socket that still reports `connected` to every health gate while silently
+  /// swallowing everything written to it. Without this, queries never trigger
+  /// that repair, so a zombie keeps charging every subsequent query the full
+  /// timeout until something else happens to cycle the connection.
+  void _repairRelaysThatNeverAnswered(String subId) {
+    final sentAt = _querySentAt.remove(subId);
+    if (sentAt == null) return;
+    final silent = [
+      for (final relay in [
+        ..._relaysSnapshot(),
+        ..._tempRelaysSnapshot(),
+        ..._cacheRelaysSnapshot(),
+      ])
+        if (relay.checkQuery(subId)) relay.url,
+    ];
+    if (silent.isEmpty) return;
+    _repairSilentRelays(silent, sentAt);
   }
 
   /// Releases the one-shot queries [relay] is parked on once it can no longer
@@ -1395,6 +1432,7 @@ class RelayPool {
       // never-fired callback in [_queryCompleteCallbacks].
       _queryCompleteCallbacks.remove(id);
       _queryFanoutInProgress.remove(id);
+      _repairRelaysThatNeverAnswered(id);
 
       // check query and send close
       var it = _relaysSnapshot();
@@ -1474,6 +1512,7 @@ class RelayPool {
     if (onComplete != null) {
       _queryCompleteCallbacks[subscription.id] = onComplete;
       _queryFanoutInProgress.add(subscription.id);
+      _querySentAt[subscription.id] = DateTime.now();
     }
 
     // Collect futures so we can await them before the early-completion
--- a/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
+++ b/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
@@ -108,6 +108,22 @@ class RelayPool {
   // Track pending AUTH events to match with OK responses
   final Map<String, String> _pendingAuthEvents = {};
 
+  /// Relay urls with a NIP-42 handshake outstanding.
+  ///
+  /// Entered when the relay's AUTH challenge starts being answered and left
+  /// when the relay resolves it (`OK`) or answering it fails. A one-shot query
+  /// parked behind an auth gate is only worth waiting for while this holds —
+  /// see [_canStillSettleQuery].
+  ///
+  /// Distinct from [_pendingAuthEvents], which is keyed by event id and only
+  /// populated once signing has finished; this is set first so the async
+  /// signing gap does not read as "no handshake".
+  final Set<String> _authHandshakeRelays = {};
+
+  /// Relays whose status callback this pool has already chained onto. Weak, so
+  /// a removed or replaced relay is not retained by the bookkeeping.
+  final Expando<bool> _statusObserved = Expando<bool>('relayStatusObserved');
+
   /// Track publishes awaiting OK confirmations (per event id).
   final Map<String, PublishTracker> _pendingPublishes = {};
 
@@ -349,6 +365,7 @@ class RelayPool {
     }
 
     relay.onMessage = _onEvent;
+    _observeRelayStatus(relay);
 
     if (await relay.connect()) {
       if (autoSubscribe) {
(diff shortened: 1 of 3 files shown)


mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
```diff
@@ -685,13 +760,217 @@ class RelayPool {
...
+  ///
+  /// The window bounds *silence* — a relay that never terminates a REQ gives
+  /// off no other signal. A relay part-way through answering is not that: it
+  /// is being served, just not finished. Expiring the window under it closes
+  /// the REQ mid-stream, and the caller is told the query completed normally,
+  /// so a truncated result set is indistinguishable from a complete one.
+  ///
+  /// No-op unless a window is already armed, so a query nobody has answered
+  /// yet still gets the caller's full timeout.
+  void _restartQuerySettleWindow(String subId) {
```
realmeylisdev, line 829: Pushed as `8213b70`, revert freely.

The window bounds silence, but nothing here separated "silent" from "still sending this query's result set". A relay part way through a page had its REQ closed one second after any sibling EOSE'd, its remaining events were dropped, and `queryEventsDetailed` still returned `timedOut: false`, so a truncated page is indistinguishable from a complete one to the caller.

Measured on two `RelayBase` relays over a fake socket: fast relay EOSEs immediately with nothing, bulk relay streams one event every 100ms for 2s. 20 events sent, 8 delivered, frames written to the bulk relay were `[REQ, CLOSE]`. That is the shape of the case in the PR body (`events=344 kinds=[[7]]`), so the liked tab would come back short and the counts with it.

The change: an EVENT for a one-shot query restarts the window. A relay that says nothing is bounded exactly as before, and one that is answering keeps the caller's budget, which is what it had before the window existed. Regression case is in `relay_pool_query_settle_test.dart`; it fails at 1/5 events with the restart removed.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
```diff
@@ -685,13 +760,217 @@ class RelayPool {
...
+    // shown the gate is shut — it refused the REQ with auth-required, refused
+    // our AUTH event, or we could never answer its challenge.
+    //
+    // "No handshake" on its own is not that evidence: `alwaysAuth` latches for
+    // the session, so a relay that has just reconnected sits in exactly that
+    // state for the round trip before its challenge arrives. Treating that as
+    // settled would silently drop the results of every read-gated relay on
+    // cold start.
+    if (relay.relayStatus.alwaysAuth && !relay.relayStatus.authed) {
+      if (_authHandshakeRelays.contains(relay.url)) return true;
```
realmeylisdev: should-fix, or fix the PR text, your call which.

The description says waiting through an outstanding handshake is preserved so the post-AUTH replay still reaches the original caller. That only holds while no other relay has answered. Once one has, the settle window arms over this branch, and `_completeQuery` runs `_releaseQuery`, which sends `CLOSE` and drops the parked query. The replay loop in the AUTH `OK` handler then finds nothing.

Reproduced with two `RelayBase` relays on a fake socket: gated relay sends `AUTH` then `CLOSED auth-required`, we sign and write our AUTH event, sibling EOSEs, and one second later the gated relay has `getQueries() == []` with `[REQ, AUTH, CLOSE]` written to it. The `OK` never gets a chance to land. Under a NIP-46 remote signer the AUTH round trip is a network call, so a second is easy to exceed, and a read-gated relay would then contribute nothing to any query a sibling answers first.

Either exempt a relay with an outstanding handshake from the window teardown, or keep the bound and drop the claim from the description. Left it alone since which one you want isn't mine to pick.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
```diff
@@ -397,11 +457,19 @@ class RelayPool {
...
       _relays[url]?.dispose();
-      _lastSilentRepairAt.remove(url);
+      _forgetRelayBookkeeping(url);
     }
     _relays.clear();
   }
 
+  /// Drops the pool-side state keyed by [url] so a relay that comes back is
+  /// not judged on the previous instance's connection.
+  void _forgetRelayBookkeeping(String url) {
```
realmeylisdev, line 478: nit: `removeTempRelay` and the temp-relay swap in `checkAndGenTempRelay` still clear `_lastSilentRepairAt` directly instead of calling this, so they skip the two new auth markers. `_authGateClosed`'s doc says both are dropped when the relay leaves the pool.

Harmless today (a fresh temp relay starts with `alwaysAuth` false, and any challenge or status transition clears the entry), it's just the one path that doesn't go through the helper.

realmeylisdev (changes requested):
went through this locally against the sdk suites. one real problem: the settle window couldn't tell "silent" from "still streaming", so a relay part way through a page got its REQ closed 1s after any sibling EOSE'd and the rest of its events were dropped, with timedOut still false. 20 sent, 8 delivered in my repro. pushed 8213b70 with the fix and a regression test, revert freely.

blocking on the auth one below, it's either the description or the code and that's your call. the other one is a nit.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
hm21: Confirmed and keeping it. Ran your test against the branch without the lib change: 1 of 5 events delivered, `timedOut: false`. That is a worse failure than the stall this PR set out to fix, since a truncated page is indistinguishable from a complete one. Restarting the window on an EVENT is the right place for it — it bounds silence and leaves progress alone.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
hm21: Real, not a description problem — fixed in `418f55dc`. Reproduced the same shape you did: one second after the sibling EOSE the gated relay had `getQueries() == []` and `[REQ, AUTH, CLOSE]` written, and the `OK` had nothing left to replay.

Took the exemption rather than dropping the claim, on the same reasoning as your streaming fix: a relay mid-handshake is being served, not silent. The window now re-arms while a straggler is inside its handshake.

It does need its own bound, since a relay that swallows our AUTH event looks identical to one still working — so the marker now carries the challenge timestamp and only counts as live for `authHandshakeWindow` (3s, wide enough for a NIP-46 round trip, well under any caller budget). Both halves have a regression test and both were verified to fail without their fix.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
hm21: Right, fixed in `418f55dc` — `removeTempRelay` and the temp-relay swap in `checkAndGenTempRelay` both go through `_forgetRelayBookkeeping` now, which is what its doc already promised.

hm21:
@realmeylisdev all three addressed, thanks — the streaming one was a good catch and a worse failure than the stall this PR started from.

- Your `8213b70` verified and kept: without the lib change your test delivers 1 of 5 events with `timedOut: false`.
- Auth thread was real, not a description problem. Fixed in `418f55dc` by exempting an outstanding handshake from the window teardown, on the same bound-silence-not-progress reasoning as your fix, with `authHandshakeWindow` keeping it finite.
- Nit fixed in the same commit.

`packages/nostr_sdk` 438 pass, `packages/nostr_client` 324, analyze clean across `lib` and both packages. Both new regression cases were verified to fail without their half of the fix. Back to you when CI goes green.

realmeylisdev (approved):
all three look right. both new cases fail without their halves of the fix, and 3s sits under every query timeout in the repo (max is 5s).

pushed one more in 6f1591f85b: the CLOSED handler still asked the raw map for the handshake, so a relay that swallows our AUTH kept suppressing its own auth-required refusal, and every later query on that socket paid the full timeout. 4s down to 2ms, regression case alongside.

PR #6676, opened 2026-08-04
fix(account): keep a switched-away account's session alive

### LINKED ISSUE #6677: fix(account): keep a switched-away account's session alive
Auto-created from PR #6676.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6676
**Branch:** `fix/account-switch-session-archive`

hm21:
Closes #6677

## Description

Switching accounts in Settings destroyed the session of the account you switched away from. The OAuth session, bunker URL, and Amber info all live in **single global** secure-storage slots, and only `signOut()` ever copied them into the per-account archive. The in-place container swap (#6342) replaced that sign-out, so nothing archived anymore: the incoming account's `restoreActiveKeys` overwrites the global slot — for a local-key account it calls `KeycastSession.clear` and wipes it outright.

On the way back there was nothing left to refresh from. The account either returned with the "Session expired" banner (local-key fallback) or failed the switch entirely with `SessionExpiredException` and a "Couldn't switch accounts" snackbar, with no way in from there.

Three parts:

**1. Archive the leaving account before the incoming one signs in.** `AuthService.archiveCurrentSignerInfo()` exposes the archival that `signOut()` already did, and `swapAccount` calls it on the live container's service first. The archive intentionally survives a failed swap's rollback — it only ever copies an account's own credentials into its own slot, so there is nothing to undo.

**2. Put the leaving account's keys back when the swap fails.** A sign-in that gets far enough to fail has already run the *target* account's `_restoreSignerInfo`, which overwrote the shared signer slots and persisted its auth source. Disposing the half-built container undid neither, so the leaving account kept working from memory but would read the wrong signer at the next cold launch — visible whenever the user backs out of the offer in part 3, or on any failure that takes the generic snackbar path. `restoreSignerInfoForCurrentAccount()` is the counterpart to part 1's archive, so the rollback can restore the account's own keys from the archive it wrote on the way in. It runs before the primary-key restore because it cannot throw; a failing key restore must not skip it.

**3. Offer a fresh sign-in instead of a dead end.** An account already in that state cannot be recovered by a swap — a swap needs credentials that already work. `SessionExpiredException` and `AccountRestoreFailedException` now take the same route `WelcomeBloc` already takes for these two failures: remember the pubkey in `pendingAccountSwitchPubkey`, sign out, land on the welcome screen with that account pre-selected. Since that signs the *working* account out, it is confirmed in a sheet first ("Session Expired" / Cancel / Sign in) rather than happening silently.

One new ARB key, `settingsSessionExpiredSwitchMessage`, translated in all 22 locales. The obvious reuse — `settingsSessionExpiredSubtitle`, "Sign in again to restore full access" — was written for the *current* account's expired-session tile. In this sheet the dead session belongs to the account the user picked and confirming signs the working account out, which is the one thing the sheet exists to get consent for and the one thing that copy never said. `settingsSessionExpired`, `authSignInTitle`, and `settingsCancel` are reused as-is.

**Related Issue:**  — Relates to #4623

## Out of Scope

Re-authenticating the target account **in place**, without signing the current one out. That needs the OAuth browser flow wired into the swap; the sign-out route is what the welcome flow already does for the same failures.

## Verification

- `flutter test test/screens/settings test/providers test/l10n` plus `auth_service_multi_account_test.dart`, `auth_service_signout_cleanup_test.dart`, and `auth/signer_secure_store_test.dart` — 965 passed, including four runs under `--test-randomize-ordering-seed random`
- `flutter test test/services/auth_service_oauth_recovery_test.dart test/services/auth_service_expired_session_downgrade_test.dart` — passed
- New coverage: archive-before-sign-in ordering and the rollback restore in `swap_account_test.dart`, `archiveCurrentSignerInfo` archiving without signing out plus `restoreSignerInfoForCurrentAccount` putting the persisted auth source back in `auth_service_multi_account_test.dart`, and the re-auth offer plus its cancel path in `settings_account_switch_test.dart`
- Manually verified on a real Samsung Galaxy device: an account switch that previously died on "Session has expired" now offers the sign-in and gets through, and switching away and back keeps the session alive

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/integration_test/account_swap_integration_test.dart
+++ b/mobile/integration_test/account_swap_integration_test.dart
@@ -117,6 +117,7 @@ void main() {
       () => swapAccount(
         deviceScope: deviceScope,
         controller: controller,
+        currentAuthService: bContainer.read(authServiceProvider),
         account: KnownAccount(
           pubkeyHex: pubkeyA!,
           authSource: AuthenticationSource.automatic,
--- a/mobile/lib/providers/swap_account.dart
+++ b/mobile/lib/providers/swap_account.dart
@@ -11,6 +11,7 @@ import 'package:openvine/providers/environment_provider.dart';
 import 'package:openvine/providers/shared_preferences_provider.dart';
 import 'package:openvine/router/app_router.dart';
 import 'package:openvine/services/auth_service.dart';
+import 'package:unified_logger/unified_logger.dart';
 
 /// Signs [container]'s fresh [AuthService] in as [account].
 ///
@@ -67,6 +68,10 @@ String? _currentRouterLocation(AccountSwitchController controller) {
 /// to do that; without it the leaving account comes back as "session expired"
 /// with no refresh token left to recover from.
 ///
+/// That archive is a precondition, not a best-effort side trip: it throws if
+/// the write fails, aborting before anything is mutated, because a swap that
+/// carries on past it destroys the credentials it was meant to preserve.
+///
 /// Prove-then-commit: nothing user-visible changes until the new account's
 /// sign-in succeeds. On failure the half-built container is disposed and the
 /// current account's container is left untouched — the rollback the previous
@@ -79,7 +84,9 @@ String? _currentRouterLocation(AccountSwitchController controller) {
 /// enough to fail has already restored the *target* account's signer keys over
 /// the shared slots and persisted its auth source, which the live container
 /// never re-reads but the next cold launch does. The rollback restores the
-/// leaving account's keys from the archive written above.
+/// leaving account's keys from the archive written above. It cannot throw:
+/// callers dispatch on the sign-in failure's type, so a rollback that escaped
+/// would swap out the type they match on and cost the user the recovery route.
 ///
 /// [signIn] is injectable for testing; production uses [signInForAccount].
 Future<void> swapAccount({
@@ -108,11 +115,18 @@ Future<void> swapAccount({
       await container.read(authServiceProvider).claimLegacyRowsForCurrentUser();
     } catch (_) {
       try {
-        // Ordered first because it cannot throw: a failing primary-key
-        // restore must not skip it and strand this account on the target's
-        // signer.
         await currentAuthService.restoreSignerInfoForCurrentAccount();
         await keyStorage.restorePrimaryKeyContainer(previousPrimary);
+      } catch (rollbackError, rollbackStack) {
+        // Swallowed on purpose: the caller dispatches on the sign-in failure's
+        // type to decide between the re-auth offer and the generic snackbar,
+        // and letting a rollback failure escape here replaces that type.
+        Log.error(
+          'Account swap rollback failed',
+          name: 'swapAccount',
+          error: rollbackError,
+          stackTrace: rollbackStack,
+        );
       } finally {
         container.dispose();
       }
--- a/mobile/lib/providers/swap_account.dart
+++ b/mobile/lib/providers/swap_account.dart
@@ -75,6 +75,12 @@ String? _currentRouterLocation(AccountSwitchController controller) {
 /// intentionally: it only ever copies the leaving account's own credentials
 /// into its own slot.
 ///
+/// Rolling the container back is not enough on its own: a sign-in that got far
+/// enough to fail has already restored the *target* account's signer keys over
+/// the shared slots and persisted its auth source, which the live container
+/// never re-reads but the next cold launch does. The rollback restores the
+/// leaving account's keys from the archive written above.
+///
 /// [signIn] is injectable for testing; production uses [signInForAccount].
 Future<void> swapAccount({
   required DeviceScope deviceScope,
@@ -102,6 +108,10 @@ Future<void> swapAccount({
       await container.read(authServiceProvider).claimLegacyRowsForCurrentUser();
     } catch (_) {
       try {
+        // Ordered first because it cannot throw: a failing primary-key
+        // restore must not skip it and strand this account on the target's
+        // signer.
+        await currentAuthService.restoreSignerInfoForCurrentAccount();
         await keyStorage.restorePrimaryKeyContainer(previousPrimary);
       } finally {
         container.dispose();
--- a/mobile/lib/providers/swap_account.dart
+++ b/mobile/lib/providers/swap_account.dart
@@ -10,6 +10,7 @@ import 'package:openvine/providers/device_scope.dart';
 import 'package:openvine/providers/environment_provider.dart';
 import 'package:openvine/providers/shared_preferences_provider.dart';
 import 'package:openvine/router/app_router.dart';
+import 'package:openvine/services/auth_service.dart';
 
 /// Signs [container]'s fresh [AuthService] in as [account].
 ///
@@ -59,20 +60,31 @@ String? _currentRouterLocation(AccountSwitchController controller) {
 /// Switches the live app to [account] in place: builds a new container on the
 /// shared [deviceScope], signs it in, and swaps it in via [controller].
 ///
+/// [currentAuthService] is the live container's service — the account being
+/// left. Its signer credentials (OAuth session, bunker URL, Amber info) live in
+/// shared secure-storage slots that the incoming account's sign-in overwrites,
+/// so they are archived per-account first. The sign-out this swap replaced used
+/// to do that; without it the leaving account comes back as "session expired"
+/// with no refresh token left to recover from.
+///
 /// Prove-then-commit: nothing user-visible changes until the new account's
 /// sign-in succeeds. On failure the half-built container is disposed and the
 /// current account's container is left untouched — the rollback the previous
 /// welcome-bounce flow could not guarantee (#4623). Rethrows so the caller can
-/// surface a "couldn't switch" affordance.
+/// surface a "couldn't switch" affordance. The archive survives that rollback
+/// intentionally: it only ever copies the leaving account's own credentials
+/// into its own slot.
 ///
 /// [signIn] is injectable for testing; production uses [signInForAccount].
 Future<void> swapAccount({
   required DeviceScope deviceScope,
   required AccountSwitchController controller,
+  required AuthService currentAuthService,
   required KnownAccount account,
   AccountSignIn signIn = _defaultSignIn,
 }) async {
   await controller.runExclusive(() async {
+    await currentAuthService.archiveCurrentSignerInfo();
     final currentLocation = _currentRouterLocation(controller);
     final container = buildAccountContainer(
       deviceScope,
--- a/mobile/lib/screens/settings/settings_screen.dart
+++ b/mobile/lib/screens/settings/settings_screen.dart
@@ -178,7 +178,7 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
     if (!mounted) return;
     final proceed = await _confirmSwitch(
       title: context.l10n.settingsSessionExpired,
-      message: context.l10n.settingsSessionExpiredSubtitle,
+      message: context.l10n.settingsSessionExpiredSwitchMessage,
       confirmLabel: context.l10n.authSignInTitle,
       confirmType: DivineButtonType.primary,
     );
--- a/mobile/lib/screens/settings/settings_screen.dart
+++ b/mobile/lib/screens/settings/settings_screen.dart
@@ -8,6 +8,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
+import 'package:keycast_flutter/keycast_flutter.dart'
+    show SessionExpiredException;
 import 'package:models/models.dart';
 import 'package:openvine/blocs/background_publish/background_publish_bloc.dart';
 import 'package:openvine/blocs/invite_status/invite_status_cubit.dart';
@@ -103,10 +105,13 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
   }
 
   /// Confirmation sheet shown before an account switch that would disturb
-  /// unfinished work. Returns true when the user chose to switch anyway.
-  Future<bool> _confirmSwitchAnyway({
+  /// unfinished work, or that has to sign the current account out to recover
+  /// the target one. Returns true when the user chose to proceed.
+  Future<bool> _confirmSwitch({
     required String title,
     required String message,
+    required String confirmLabel,
+    DivineButtonType confirmType = DivineButtonType.error,
   }) async {
     final proceed = await VineBottomSheet.show<bool>(
       context: context,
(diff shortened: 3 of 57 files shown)


mobile/lib/providers/swap_account.dart
```diff
@@ -90,6 +108,10 @@ Future<void> swapAccount({
       await container.read(authServiceProvider).claimLegacyRowsForCurrentUser();
     } catch (_) {
       try {
+        // Ordered first because it cannot throw: a failing primary-key
+        // restore must not skip it and strand this account on the target's
+        // signer.
+        await currentAuthService.restoreSignerInfoForCurrentAccount();
```
realmeylisdev, line 118: `restorePrimaryKeyContainer` on the next line always throws here, and that swallows the exception the new handlers key on.

`previousPrimary` (line 105) is the same object `SecureKeyStorage` holds in `_cachedKeyContainer` — `getKeyContainer` returns the cached instance on a hit, and caches the one it retrieves on a miss (`secure_key_storage.dart:321-345`). `restorePrimaryKeyContainer` opens with `_cachedKeyContainer?.dispose()` (`:503`) and then hands that same, now-disposed container to `storeKey`, which goes through `withPrivateKey` -> `_ensureNotDisposed()`.

Ran it against the real classes:

```
previousPrimary.isDisposed BEFORE = false
previousPrimary.isDisposed AFTER  = true
THROWN => PlatformSecureStorageException: Storage operation failed: SecureKeyException: Container has been disposed
```

That throw escapes the inner `try`/`finally`, so `rethrow` on line 119 never runs. Confirmed the shape separately:

```
rollback ok    -> reached rethrow -> on SessionExpiredException -> re-auth offer
rollback throws-> finally ran     -> generic catch -> "Couldn't switch" snackbar
```

So `settings_screen.dart:278/280` never match and the user gets the dead end the PR is removing.

It lands on exactly the two paths this PR handles, because those are the ones that fail *without* replacing the PRIMARY cache: `SessionExpiredException` (`auth_service.dart:1820` — `getIdentityKeyContainer` doesn't touch the cache, `getKeyContainer` is a cache hit, `switchToIdentity` never runs) and the `AccountRestoreFailedException` from `:1891`. Where `switchToIdentity` did run the cache holds a different container, the restore succeeds and the type survives, so it's worth re-checking which failure the Samsung run actually hit.

Nothing catches it today: `_FakeSecureKeyStorage.restorePrimaryKeyContainer` is a no-op, and the settings test throws while evaluating `deviceScope.switchController`, so this rollback is never entered.

Suggested fix, two parts:

1. `secure_key_storage.dart:503`, stop wiping the container being restored:
```dart
if (!identical(_cachedKeyContainer, keyContainer)) _cachedKeyContainer?.dispose();
```
2. make this rollback non-masking so a future failure can't swallow the caller's type either:
```dart
} catch (_) {
  try {
    await currentAuthService.restoreSignerInfoForCurrentAccount();
    await keyStorage.restorePrimaryKeyContainer(previousPrimary);
  } catch (rollbackError, rollbackStack) {
    Log.error(
      'Account swap rollback failed',
      name: 'swapAccount',
      error: rollbackError,
      stackTrace: rollbackStack,
    );
  } finally {
    container.dispose();
  }
  rethrow;
}
```

plus a test where `signIn` throws `SessionExpiredException` and the fake storage's restore throws, asserting the caller still sees `SessionExpiredException`.

mobile/lib/screens/settings/settings_screen.dart
```diff
@@ -232,8 +272,13 @@ class _SettingsScreenState extends ConsumerState<SettingsScreen> {
                 await swapAccount(
                   deviceScope: deviceScope,
                   controller: deviceScope.switchController,
+                  currentAuthService: ref.read(authServiceProvider),
                   account: account,
                 );
+              } on SessionExpiredException catch (e) {
+                await _offerReauthentication(account, e);
+              } on AccountRestoreFailedException catch (e) {
```
realmeylisdev, line 280: Nothing covers this clause. Both new widget tests stub `SessionExpiredException` (`settings_account_switch_test.dart:422`, `:454`), so deleting lines 280-281 leaves the whole suite green.

Cheap fix: run the existing "offers a fresh sign-in" test over both types, e.g. a small loop or a second `when(...).thenThrow(AccountRestoreFailedException(...))` case.

mobile/test/providers/swap_account_test.dart
```diff
@@ -125,7 +129,45 @@ void main() {
     expect(_isDisposed(attempted!), isTrue);
     expect(_isDisposed(initial), isFalse);
     expect(keyStorage.restoredPrimary, same(keyStorage.primary));
+    // A sign-in that got far enough to fail already restored the target
+    // account's signer keys and auth source over the shared slots. Disposing
+    // its container does not undo that — the leaving account's own keys have
+    // to go back, or it reads the wrong signer at the next launch.
+    expect(currentAuthService.calls, equals(['archive', 'restore']));
```
realmeylisdev, line 136: The comment at `swap_account.dart:111-113` calls this ordering load-bearing, but nothing pins it: `_FakeSecureKeyStorage.restorePrimaryKeyContainer` is a no-op, so swapping lines 114 and 115 keeps every test green.

An opt-in `throwOnRestore` on the fake plus one test asserting `calls` still ends with `'restore'` would pin it, and it doubles as the regression test for the `swap_account.dart:114` comment above.

realmeylisdev (changes requested):
the archive half is solid, i walked every leaving/incoming auth-source pair and the round-trip holds. but the new re-auth offer can't actually fire: the rollback throws before the rethrow, so settings only ever sees a PlatformSecureStorageException and shows the old snackbar. repro inline.

two smaller ones on test coverage below.

mobile/lib/providers/swap_account.dart
```diff
@@ -59,20 +60,37 @@ String? _currentRouterLocation(AccountSwitchController controller) {
...
 /// [signIn] is injectable for testing; production uses [signInForAccount].
 Future<void> swapAccount({
   required DeviceScope deviceScope,
   required AccountSwitchController controller,
+  required AuthService currentAuthService,
   required KnownAccount account,
   AccountSignIn signIn = _defaultSignIn,
 }) async {
   await controller.runExclusive(() async {
+    await currentAuthService.archiveCurrentSignerInfo();
```
realmeylisdev, line 100: Cleared this one, not a regression: on main `swapAccount` doesn't archive at all, so this is strictly better. Flagging it because the whole fix now hangs on this one call.

`SignerSecureStore.archive` wraps its entire body in `try { … } catch (e) { Log.warning(…) }` with no rethrow (`signer_secure_store.dart:254-260`). So a keychain/keystore write failure resolves this `await` normally, the swap carries on, and `signInForAccount(B, automatic)` then calls `KeycastSession.clear()` / `clearBunker()` / `clearAmber()` on the shared slots. A's credentials are gone with a warning in the log and nothing else.

The prove-then-commit property in the doc comment holds for the container but not for the credentials. If you want it to cover both, this is the call that has to abort the swap instead of warning.

mobile/test/services/auth_service_multi_account_test.dart
```diff
@@ -780,6 +780,61 @@ void main() {
...
+      // Stands in for a failed in-place switch: the target account's sign-in
+      // got as far as restoring its own signer info, which persisted its auth
+      // source. Disposing its container leaves that behind, so the next launch
+      // would look for a bunker this local-key account never had.
+      final prefs = await SharedPreferences.getInstance();
+      await prefs.setString('authentication_source', 'bunker');
+
+      await authService.restoreSignerInfoForCurrentAccount();
+
+      expect(prefs.getString('authentication_source'), equals('automatic'));
```
realmeylisdev, line 857: Nit: this test passes for any pubkey, so the argument on `auth_service.dart:1611` isn't pinned.

`createNewIdentity` leaves `_authSource == automatic`, and `restoreActiveKeys`'s `automatic`/`importedKeys`/`none`/`nip07` branch (`signer_secure_store.dart:363-377`) never reads `pubkeyHex` — it only calls `clearBunker()`, `clearAmber()`, `KeycastSession.clear()`. Mutate line 1611 to a hardcoded wrong pubkey and both assertions here still hold.

The pubkey is the entire point for a leaving account on bunker / amber / divineOAuth, where it selects `bunker_info_<pubkey>`, `amber_pubkey_<pubkey>`, `keycast_session_<pubkey>` — a wrong one silently restores nothing. A bunker-source case would pin it.

mobile/integration_test/account_swap_integration_test.dart
```diff
@@ -117,6 +117,7 @@ void main() {
       () => swapAccount(
         deviceScope: deviceScope,
         controller: controller,
+        currentAuthService: bContainer.read(authServiceProvider),
```
realmeylisdev, line 120: Nit / follow-up. Each seam the PR adds is individually mutation-killable (drop the archive and `calls` becomes `['signIn']`, drop the rollback restore and it becomes `['archive']`), so this isn't "untested". What nothing composes is the property the bug report is actually about: A's per-account archive survives B's sign-in and comes back on the way home.

This file is the natural home for that and it only got the compile fix — it swaps B→A once and never switches back.

Worth knowing it also isn't executed by CI: the test job runs `very_good test --optimization` over `test/`, and the only integration job is `flutter test integration_test/e2e/ --tags service`. Anything at `integration_test/` root is analyze-and-format only, so an assertion added here won't guard the branch unless someone runs it on a device.

realmeylisdev (commented):
three more from the sweep that i cleared, none of them block. posting so you don't have to re-derive them.

also worth knowing, not this PR: `keycast_flutter`'s `refreshSession` deletes `keycast_refresh_token` on any non-200 (`oauth_client.dart:196-198`), 5xx included, with a comment asserting the server consumed it. so the token this PR archives can still get dropped by a transient server error, and nothing writes that key back except a successful refresh or an account-switch restore.

mobile/lib/providers/swap_account.dart
hm21: Confirmed against the real classes, and both halves are fixed.

The dispose is the root cause, so that is fixed at source in `50c55ec9ea` — `restorePrimaryKeyContainer` now only disposes the outgoing container when it is a different object, pinned by `secure_key_storage_restore_primary_test.dart`. Reproduced your output first:

```
BEFORE isDisposed = false
AFTER  isDisposed = true
THROWN => PlatformSecureStorageException: Container has been disposed
```

The masking is fixed independently in `5b020df3bb`, since a future rollback failure should not cost the caller its type either.

One correction to the scope: it is not "always throws". With no primary key stored `getKeyContainer` returns null, the delete branch runs and nothing throws — which is exactly the `SessionExpiredException` case, since that path is reached precisely when there is no local key ("no refresh, no local keys"). That is why the Samsung run did reach the offer. `AccountRestoreFailedException` is the one that was masked, and it is also the one your next comment showed was untested.

mobile/lib/screens/settings/settings_screen.dart
hm21: Fixed in `1abaee1d92`. The offer test now runs over both failures as a table, so deleting either clause fails the suite.

mobile/test/providers/swap_account_test.dart
hm21: Fixed in `5b020df3bb`. The fake got an opt-in `restoreError`, and the new "surfaces the sign-in failure even when the rollback throws" test asserts both the caller-visible type and that `calls` still ends with `restore`. Verified it kills the mutation — swapping the two lines turns it red.

mobile/lib/providers/swap_account.dart
hm21: Agreed, and changed in `5b020df3bb` — the archive is a precondition now, not best-effort.

`archive` takes `throwOnFailure` (default false, so sign-out keeps swallowing: that session is going away either way) and `archiveCurrentSignerInfo` opts in. It is the first call in the swap, so aborting there is safe — nothing has been mutated yet — and the user gets "couldn't switch" instead of a silently destroyed session. The doc comment now claims prove-then-commit for the credentials too. Covered by "aborts before building anything when the archive fails".

mobile/test/services/auth_service_multi_account_test.dart
hm21: Right, and fixed in `1abaee1d92`. Moved to a bunker-source case in `auth_service_bunker_lifecycle_test.dart`, where the existing harness gives a real connected session and the pubkey actually selects the slot. Confirmed it kills the mutation you described — hardcoding a wrong pubkey at that line turns it red.

mobile/integration_test/account_swap_integration_test.dart
hm21: Good catch on the CI gap — an assertion added at `integration_test/` root would not have guarded anything, so the round trip went somewhere that runs.

The same bunker test in `1abaee1d92` composes it: archive account A, overwrite the shared slot with another account's credentials the way a sign-in does, restore, assert A's own URL is back. Not the full two-container swap, but it is the property the bug report is about and it runs in CI. Left the integration test as the compile fix.

hm21:
@realmeylisdev all six addressed — thanks, the `restorePrimaryKeyContainer` one was the real find.

Fixed at source rather than worked around: it now only disposes the outgoing container when it is a different object (`50c55ec9ea`), plus the non-masking rollback you suggested (`5b020df3bb`) so a future rollback failure cannot cost the caller its type either. The archive is a precondition now too — it aborts the swap instead of warning.

One scope correction on that thread: it is not "always throws". With no primary key stored the null branch runs and nothing throws, which is exactly the `SessionExpiredException` path — so the device run genuinely reached the offer. `AccountRestoreFailedException` was the masked one, and your next comment showed it was also the untested one.

Every seam you flagged is now mutation-checked; I verified the two you named specifically turn red.

realmeylisdev (approved):
All six verified on `1abaee1d92`, each by mutation: reverting the `identical` guard, deleting either re-auth clause, swapping the two rollback restores, and hardcoding a wrong pubkey all turn the intended test red. The storage test reproduces the original repro shape exactly, and the bunker round trip lands somewhere CI actually runs.

One gap left, pushed as `05670d4629` — revert freely. Making the archive a precondition wasn't held anywhere: `swap_account_test` fakes `archiveCurrentSignerInfo` itself, so deleting `if (throwOnFailure) rethrow;` from the store, or dropping `throwOnFailure: true` at the call site, left all 144 tests across every archive-touching suite green. Added a store case for both directions (rethrows when asked, still swallows for sign-out) and one on `archiveCurrentSignerInfo`, so either mutation is red now.

On the scope correction — agreed it isn't "always", but the discriminator is whether the *leaving* account has a local PRIMARY, not which exception the sign-in threw. `previousPrimary` is non-null for any local-key account, and the `hasKeys()` branch at `auth_service.dart:1785-1790` reaches `SessionExpiredException` with PRIMARY still holding the leaving account's keys — so switching away from a local-key account to an expired OAuth one was masked too. Both are covered by the source fix, so nothing to change; noting it only because it makes the blast radius wider than one untested path.

PR #6680, opened 2026-08-04
fix(profile): repair the edit-profile clear and unsaved-edit paths

realmeylisdev:
## Description

Three defects in the edit-profile screen that only exist once #6630 and #6657 are
both in. Neither branch was wrong on its own: #6657 added the staged-media store
and the unsaved-changes guard against a screen that could clear the *banner*;
#6630 added the avatar clear affordance and made the form seed from the cache
before the relay answers. The overlap is what breaks.

**A stale cache fakes an unsaved edit.** `InitialProfileFieldsSet` supplies the
baseline the guard compares against, and it only ran on the first snapshot. The
cached profile fills the form first, then the relay's copy refreshes whatever the
user has not touched — so whenever the two disagree, the refreshed field no
longer matches its baseline and leaving the screen claims there are changes to
save on a form nobody touched. Renaming yourself on another device is enough to
reproduce it.

The baseline now follows every seed. The current values come back off the
controllers rather than from the profile, so a field the user really did type
into keeps its old baseline and still reads as unsaved.

**A removed avatar comes back.** `ProfilePictureCleared` never dropped the staged
URL from `StagedProfileMediaStore`, so `_initialState` restored it the next time
the editor was built and the picture the user had just removed reappeared, with
Save ready to republish it. `ProfileBannerUrlSet` had the mirror of the same
omission — a pasted banner link vanished on a rebuild while an uploaded one
survived. Both now write through, matching what the pasted-picture branch
already did.

**`pictureCleared` outlives the edit.** `ProfileEditDiscarded` and the
save-success emit both reset `bannerCleared` and neither reset `pictureCleared`,
because they were written before an avatar clear existed. Discard left the avatar
drawn as removed, and after a clear-then-save the stale flag vetoed every later
persisted-picture seed for the rest of the session.

**Related Issue:** Relates to #6631

## Out of Scope

- The `@alice` username-field copy question left open on #6630 — that is a
  product decision, not a defect.

## Verification

- Each fix confirmed against the merged `b2285e42a4` before it was written: the
  stale-cache prompt reproduced red, and reverting any one of the three guards
  afterwards turns exactly the matching test red and nothing else.
- `flutter test test/blocs/profile_editor/ test/screens/profile_setup_screen_test.dart test/screens/profile_setup/ test/services/staged_profile_media_store_test.dart` — 263 green
- `dart format` — clean
- Walked by hand in a debug build on `fb08bf4`, against production relays with a
  throwaway identity. Opening the editor and leaving once the relay copy lands
  no longer prompts about unsaved changes; a pasted banner link and an avatar
  removal each survive a force-quit; Save publishes the removal and the profile
  drops the picture; and a picture staged after that clear-then-save is no
  longer vetoed by the stale flag.

## Type of Change

### DIFF
--- a/mobile/lib/blocs/profile_editor/profile_editor_bloc.dart
+++ b/mobile/lib/blocs/profile_editor/profile_editor_bloc.dart
@@ -200,6 +200,7 @@ class ProfileEditorBloc extends Bloc<ProfileEditorEvent, ProfileEditorState> {
         externalNip05: state.initialExternalNip05 ?? '',
         pendingAvatarStatus: PendingAvatarStatus.idle,
         pendingPictureUrl: null,
+        pictureCleared: false,
         pendingBannerStatus: PendingBannerStatus.idle,
         pendingBannerUrl: null,
         pendingBannerColor: _parseBannerHexColor(state.persistedBanner),
@@ -330,10 +331,10 @@ class ProfileEditorBloc extends Bloc<ProfileEditorEvent, ProfileEditorState> {
     );
   }
 
-  void _onProfilePictureCleared(
+  Future<void> _onProfilePictureCleared(
     ProfilePictureCleared event,
     Emitter<ProfileEditorState> emit,
-  ) {
+  ) async {
     if (state.pendingAvatarStatus == PendingAvatarStatus.uploading) {
       Log.info(
         'Ignoring ProfilePictureCleared received while an upload is in flight',
@@ -341,6 +342,12 @@ class ProfileEditorBloc extends Bloc<ProfileEditorEvent, ProfileEditorState> {
       );
       return;
     }
+    // Drop the stored staging too, or the next editor build restores the very
+    // picture the user just removed.
+    await _persistStagedProfileMedia(
+      pictureUrl: null,
+      bannerUrl: state.pendingBannerUrl,
+    );
     emit(
       state.copyWith(
         pendingAvatarStatus: PendingAvatarStatus.idle,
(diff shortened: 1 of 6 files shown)


hm21 (approved):
LGTM. Verified the three fixes by reverting each guard on its own — exactly the matching test goes red, nothing else.

Pushed two commits, revert freely:

**cecb017** — `_persistStagedProfileMedia` now reads the URLs off `state` after the emit instead of taking them as arguments. Hand-threading the field you weren't changing is how both write-through omissions got in; there's one source of truth now. No behaviour change.

**fb08bf4** — dropping the staged URL stops Save republishing the removed picture, but the rebuilt editor still came up with `pictureCleared: false` and seeded the persisted picture straight back, so the removal was lost anyway. `StagedProfileMediaStore` now carries `pictureCleared`/`bannerCleared`, `_initialState` restores them, and `InitialPersistedBannerSet` stops resetting the flag — it fires unconditionally off the first snapshot, so it would have undone a restored banner removal before you saw the screen. `ProfileBannerCleared` had the same hole. Payloads written before this read as "nothing removed", so nothing in flight is discarded on upgrade.

Validation: `flutter analyze lib test` clean, 275 tests green across every file touching `ProfileEditorBloc` / `StagedProfileMediaStore`, each new guard verified red without its fix, CI green.

One thing left for you: Verification lists automated checks only, and this changes user-visible avatar/banner behaviour. Please confirm it on a device before merging.

realmeylisdev:
Tests went well — 269 green across `ProfileEditorBloc`, `StagedProfileMediaStore` and the profile-setup screens, analyze and format clean, CI green on `fb08bf4`.

On your ask: I walked the avatar and banner flows by hand in a debug build rather than leaving it at the automated checks. Opening the editor and leaving once the relay copy lands no longer prompts; a pasted banner link and an avatar removal each survive a force-quit; Save publishes the removal and the header drops the picture; and a picture staged after that clear-then-save is no longer vetoed. Both your commits do what they say.

Not this PR: a staged banner *colour* still doesn't survive a rebuild — the store carries the two URLs and now the two flags, but not `pendingBannerColor`, so the persisted banner seeds back over it. Predates both branches, so I'll open an issue rather than widen this. Same walkthrough turned up a 64px `Column` overflow at `auth_hero_section.dart:21`, also unrelated.

PR #6681, opened 2026-08-04
fix(nostr): keep the relay count when a publish reaches nothing

rabble:
## What

Follow-up to #6613, from an independent post-merge review. Three fixes plus one test replacement, all in `nostr_sdk`.

### 1. A publish that reaches nothing lost its relay count

Narrowing the counted target set to *connected* relays fixed "every successful publish looks partial", but took the all-offline case with it. Verified on `main` before the fix:

```
targetCount=0  unreachable=[]  noResponse=[]  summary="no relay reached"
```

...for a pool with two configured, write-enabled, unreachable relays. That is indistinguishable from having no relays configured at all.

The fix floors `unreachableTargets` to the full target set **only** when nothing answered and nothing was counted. Because it fires exclusively on a publish with zero answers, it can never widen the denominator of one that partly succeeded — the bug #6613 set out to fix stays fixed.

**Scope, corrected during review** (thanks @realmeylisdev). An earlier draft of this description claimed the floor also repairs the deletion services and the kind:10002 bootstrap warning. It does not, and both claims were checked rather than reasoned about:

- **`ContentDeletionService` and `AccountDeletionService` never reach this code** in the all-offline state. Both call `publishEventAwaitOk` with no `targetRelays`; with nothing connected, `nostr_client.dart` returns a hand-built `PublishOutcome` before the SDK is invoked at all (`targetCount=0`, `summary="no relay reached"`). That path is untouched here and still reads that way.
- **The kind:10002 bootstrap is unaffected.** `_intendedPublishTargets` puts every `tempRelays` entry into `counted` unconditionally, and `NostrClient` forwards a named publish's list as both `targetRelays` and `tempRelays` — so `counted` is never empty for a named publish and the floor cannot fire. Same output on `main` and here: `targetCount=5 unreachable=5 summary="no relay reachable (5 unreachable)"`.

What the floor actually fixes is the unfiltered publish where a connect momentarily succeeds and the socket drops before `_intendedPublishTargets` reads the live status. Narrower than first described, still a real case, and the mechanism is verified below.

### 2. `expectedRelays` was not defensively copied

`countedTargets` was copied at construction; `expectedRelays` was assigned by reference. Mutating the caller's set after construction therefore changed `isTarget`, letting an arbitrary relay accept or reject a publish it was never sent. Now copied the same way. This one is user-visible on its own.

### 3. Stale docs on `unreachableTargets`

It read *"the socket was down"* — which is now precisely the case normally **excluded** from that list. Rewritten. A later pass (`ad15d0c1b`, originally `1708f445b` before the force-push) corrected two more things review caught: the doc named **one** exception to the narrowing where there are **two** (a relay the caller names in `targetRelays`/`tempRelays` is never narrowed away, so an already-offline named relay is listed even when the publish reached plenty — the test case directly above it proves this), and the floor's inline comment implied it covers any all-offline publish rather than unfiltered ones only.

### 4. A test that could not fail

`the disconnected relay still cannot speak for the publish` passed on every revision, including the base commit before #6613. Its fixture gives `downUrl` a channel that errors on `ready`, so its stream never delivers a frame — no implementation could put it in `acceptedBy`, `rejectedBy` or `noResponseFrom`. Per `.claude/rules/testing.md`, a test that cannot fail hides the gap rather than guarding it.

Replaced with one that exercises the `isTarget` gate its name claimed to cover: a **connected** relay outside `targetRelays` volunteering an `OK` for an event it never received. That is the property the gate exists for — `relay_pool.dart` describes it as stopping a relay from *"deciding the outcome of, say, a narrowly targeted direct message it has no part in."*

## Verification

Both new tests were confirmed able to fail, not just to pass:

| Test | Made red by |
|---|---|
| a publish that reached nothing still reports how many relays it had | reverting the `_complete()` floor |
| a connected relay outside the target list cannot decide the publish | stubbing `isTarget` to `=> true` |

- `flutter test` in `packages/nostr_sdk` — **480 passing**
- Consumers: `content_deletion_service_test`, `account_deletion_service_test`, `nostr_client_provider_bootstrap_test` — 46 passing
- `dart format` — clean

Relates to #6215


rabble (commented):
Reviewed the current head and checks. I found no actionable correctness, security, or regression issues. I could not submit an approval because GitHub treats me as the PR author; review is complete with no code objections.

The PR still carries the `needs-issue` label; please add or confirm the tracking issue reference per the repository process.

rabble:

**This is the strongest PR in the current ready queue. No blocking findings.**

Things I specifically checked rather than took on trust:

- **The floor cannot regress #6613.** The guard requires `unreachable.isEmpty && noResponse.isEmpty && _accepted.isEmpty && effectiveRejected.isEmpty`. Any relay that was awaitable but silent lands in `noResponse`, and any that answered lands in `_accepted`/`effectiveRejected`, so the branch is reachable only when `_countedTargets` is empty — i.e. nothing was ever counted. It cannot widen the denominator of a partial success. The comment says this; the control flow backs it up.
- **The defensive-copy fix is real.** `expectedRelays` was assigned by reference while `countedTargets` was copied, so a caller mutating its set post-construction moved `isTarget` under the tracker. Note the initializer now reads the *parameter* in both positions, which is what you want.
- **The replaced test genuinely could not fail.** A channel that errors on `ready` never delivers a frame, so no implementation could place `downUrl` in any of the three buckets — it asserted three `isNot(contains(...))` that were structurally guaranteed. Replacing it with the connected-bystander case tests the `isTarget` gate the name always claimed.

The "made red by" table is the right way to report this, and it is what let me verify the above quickly.

One small thing, take it or leave it: the doc on `unreachableTargets` now describes a list whose membership rule has an exception carved into it ("normally *not* listed here… the one exception is"). That is accurate, but a reader reaching for `unreachableTargets` to drive recovery logic could reasonably act on it wrongly in the all-offline case, where entries mean "we never got far enough to ask" rather than "the send failed". If a caller ever branches on this list rather than just counting it, a separate `reachedNothing` boolean on the outcome would be harder to misread than a documented exception. Not worth doing speculatively — only if a caller shows up.

NotThatKindOfDrLiz:
@rabble I pushed one takeover commit, `docs(nostr): clarify publish unreachable partition`, to align the class-level PublishOutcome partition doc with the all-offline unreachable-target exception this PR introduces.

Verification:
- `flutter test test/unit/publish_outcome_test.dart test/unit/relay_pool_publish_targets_test.dart` from `mobile/packages/nostr_sdk`
- GitHub checks are green on `97f6b6e3b`

I also added `Relates to #6215` to the PR body; the `needs-issue` label has cleared. Please review the takeover commit and merge when you're ready.

NotThatKindOfDrLiz (approved):
Approved after takeover. I pushed a doc-only follow-up for the stale class-level partition invariant, verified the focused nostr SDK tests locally, and confirmed all GitHub checks are green on the current head. The PR body now links #6215 and the issue label has cleared.

mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
```diff
@@ -387,9 +395,23 @@ class PublishTracker {
...
         .where((r) => !awaitable.contains(r) && !answered(r))
         .toList(growable: false);
+    // A publish that reached nothing must still say how many relays it had.
+    // `_countedTargets` excludes relays that were offline before the fan-out,
+    // so when *every* relay is offline it is empty and the outcome collapses
+    // to `targetCount: 0` / "no relay reached" — indistinguishable from having
+    // no relays configured at all. This floor only fires when nothing answered
+    // and nothing was counted, so it can never widen the denominator of a
+    // publish that partly succeeded.
+    if (unreachable.isEmpty &&
```
realmeylisdev, line 405: Worth pinning down when this actually fires, because the two consumers in the description don't reach it.

`ContentDeletionService` (`content_deletion_service.dart:253`) and `AccountDeletionService` (`account_deletion_service.dart:189` and `:335`) both call `publishEventAwaitOk` with no `targetRelays`. In the state you describe — relays configured, write-enabled, nothing connects — `retryDisconnectedRelays()` stamps every one `RelayState.error` (`relay_manager.dart:481-489`), `connectedRelays` stays empty, and `nostr_client.dart:721-733` hands back a `PublishOutcome` built by hand, whose `unreachableTargets` defaults to `const []`. The SDK is never called, so this floor never runs.

Ran against a `RelayManager` reporting nothing connected:

```
targetCount=0  unreachable=[]  summary="no relay reached"  failed=true
```

with `sendEventAwaitOk` never invoked. That is the same line the PR opens with, still there on this branch.

The floor isn't dead code — it fires when a connect momentarily succeeds (`relay_manager.dart:480-482` stamps `connected`) and the socket drops before `_intendedPublishTargets` reads the live status. Real case, just not the airplane-mode one. So either narrow the framing, or carry the same floor into the early return so delete and account-deletion actually get the count — your call on which.

mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
```diff
@@ -387,9 +395,23 @@ class PublishTracker {
         .toList(growable: false);
     // A relay that answered is never unreachable, even if the fan-out reported
     // its write as failed — the answer proves it got the frame.
-    final unreachable = _countedTargets
+    var unreachable = _countedTargets
         .where((r) => !awaitable.contains(r) && !answered(r))
         .toList(growable: false);
+    // A publish that reached nothing must still say how many relays it had.
+    // `_countedTargets` excludes relays that were offline before the fan-out,
```
realmeylisdev, line 393: This holds for an unfiltered publish only. `_intendedPublishTargets` adds every `tempRelays` entry to `counted` unconditionally (`relay_pool.dart:1793-1796`), and `NostrClient` forwards the caller's list as both `targetRelays` and `tempRelays` (`nostr_client.dart:737-745`) — so for any named publish `counted` is never empty and the floor below cannot fire.

Which is why the kind:10002 bootstrap isn't affected by this change. I ran the bootstrap shape (5 named relays, all offline) against `main`'s `publish_outcome.dart` and this branch's:

```
main:   targetCount=5 unreachable=5 summary="no relay reachable (5 unreachable)"
branch: targetCount=5 unreachable=5 summary="no relay reachable (5 unreachable)"
```

`nostr_client_provider_bootstrap_test.dart:48-56` hand-builds the outcome, so it never covered the real path either way.

mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
```diff
@@ -58,10 +59,16 @@ class PublishOutcome {
...
+  /// These never received the `EVENT` frame — the send timed out, or the relay
+  /// dropped out mid-publish. They are reported separately from
+  /// [noResponseFrom] because "we never asked" and "we asked and got silence"
+  /// call for different recovery.
+  ///
+  /// A relay that was already offline when the publish started is normally
+  /// *not* listed here: it never counted as a target, so it cannot drag
+  /// [acceptedByAll] down. The one exception is a publish that reached nothing
+  /// at all, where every targeted relay is listed so the outcome can still say
+  /// how many there were.
```
realmeylisdev: Two exceptions, not one. A relay named in `targetRelays`/`tempRelays` is exempt from the narrowing (`relay_pool.dart:1768-1772`), so an already-offline named relay lands here even when the publish reached plenty.

The test two cases above proves it: in "an explicitly targeted relay is reported unreachable, not dropped", `downUrl` is offline before the publish starts, `liveUrl` accepts, and `unreachableTargets == [downUrl]`.

```suggestion
  /// A relay that was already offline when the publish started is normally
  /// *not* listed here: it never counted as a target, so it cannot drag
  /// [acceptedByAll] down. Two exceptions: a relay the caller named in
  /// `targetRelays` or `tempRelays` is never narrowed away, and a publish
  /// that reached nothing at all lists every targeted relay so the outcome
  /// can still say how many there were.
```

realmeylisdev (changes requested):
I checked the mechanism rather than the write-up, and the mechanism holds up. Reverting the `_complete()` floor turns the new test red (`targetCount` 0 vs 2); stubbing `isTarget => true` turns the bystander test red. I also brute-forced the partition invariant over all 4096 tracker states — no relay ever lands in two buckets, and the floor never widens a publish that partly succeeded.

Green here too: 423 in `nostr_sdk`, 45 across the three consumer suites, 25/25 under random ordering, `analyze` and `dart format` clean, and the `test/unit` ratchet doesn't apply (it only scans `mobile/test/unit`). App builds and runs off this branch.

Two of the three motivations don't survive checking, and one doc sentence is wrong:

- **Delete and account-deletion never reach this code** in the state you describe. Both call `publishEventAwaitOk` with no targets, and that path returns early when nothing is connected. I ran it: `targetCount=0 unreachable=[] summary="no relay reached"`, SDK never invoked — the same line the PR opens with.
- **The kind:10002 bootstrap is unaffected.** Identical output on `main` and here, because `tempRelays` go into `counted` unconditionally.
- **The `unreachableTargets` doc names one exception where there are two.** The test two cases above it proves the second.

None of this makes the change wrong — the floor is correct and the `expectedRelays` copy is a real fix. It's the claimed reach that's off, plus one word in a doc. Details inline.

mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
rabble: Agreed, and corrected the framing rather than widening the floor.

You are right that both deletion services return before the SDK is invoked — `publishEventAwaitOk` with no `targetRelays`, nothing connected, hand-built outcome at `nostr_client.dart:721-733`. I did not carry the floor into that early return: doing so would put a second, differently-derived target count in a path this PR otherwise does not touch, and the right fix there is for the early return to report the configured write-relay set, which is its own change.

So: PR description rewritten to drop the delete / account-deletion claim and name the real firing case (unfiltered publish, connect momentarily succeeds, socket drops before `_intendedPublishTargets` reads live status). `1708f445b` also narrows the inline comment at the floor to say it applies to unfiltered publishes only.

mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
rabble: Confirmed and documented. `1708f445b` extends the floor comment:

```
// It applies to unfiltered publishes only. A caller-named publish puts
// every `targetRelays`/`tempRelays` entry into `_countedTargets`
// unconditionally, so `unreachable` is already non-empty and the floor is
// unreachable.
```

The kind:10002 bootstrap claim is out of the PR description too — your `targetCount=5 unreachable=5` on both sides is the same result I get.

mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
rabble: Applied your suggestion verbatim in `1708f445b`. The test directly above it is exactly the proof, and the doc was contradicting it.

rabble:
@realmeylisdev All three addressed in `1708f445b` plus a PR-description rewrite — thanks for checking the mechanism instead of the write-up, that is exactly where this was wrong.

- **Two exceptions, not one** — your suggested doc text applied verbatim. The test directly above it was already proving the doc wrong.
- **Delete / account-deletion never reach the floor** — correct, and I chose to narrow the framing rather than carry the floor into the early return. Putting a second, differently-derived target count into `nostr_client.dart:721-733` would touch a path this PR otherwise leaves alone; the right fix there is for the early return to report the configured write-relay set, which is its own change. Both claims are gone from the description, replaced with the case that actually fires (unfiltered publish, connect momentarily succeeds, socket drops before `_intendedPublishTargets` reads live status).
- **kind:10002 bootstrap unaffected** — same result here. Now stated in the inline comment at the floor, so the next reader does not have to rediscover that `tempRelays` go into `counted` unconditionally.

The `expectedRelays` defensive copy and the replaced can't-fail test stand on their own; nothing in your review touched those.

Re-verified after the change: 423 passing in `nostr_sdk`, `dart format` clean.

mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
```diff
@@ -24,9 +24,10 @@ import 'dart:async';
 /// [rejectedBy], [noResponseFrom] or [unreachableTargets], and no other relay
 /// appears at all. "Targeted" means every relay the fan-out reached, plus the
 /// relays it meant to count if a write failed. Configured relays that were
-/// plainly disconnected and never attempted do not appear, while relays whose
-/// in-flight connection was actually attempted can still be reported as
-/// unreachable (see [PublishTracker.countedTargets]).
+/// plainly disconnected and never attempted normally do not appear, unless the
```
realmeylisdev: Small one, and your call — this now reads as though reaching nothing is the only exception, but a caller-named relay is a second one. `targetRelays`/`tempRelays` are exempt from the narrowing, so an already-offline named relay lands in `unreachableTargets` even when the publish reached plenty.

Proved it by dropping the `!hasFilter` guard in `_intendedPublishTargets` — `an explicitly targeted relay is reported unreachable, not dropped` goes red with `Expected: ['wss://relay.offline.example']  Actual: []`. So that relay is there purely because it was named, not because it was attempted.

The field doc below now states both exceptions correctly and "normally" hedges it here, so I left the wording alone rather than editing a line you had already approved.

realmeylisdev (approved):
Re-reviewed after the force-push. The change itself is sound — I re-ran the mutation checks against the current head and both new tests still fail for the right reason. What went missing was the docs.

**Pushed, one commit each so you can revert them independently:**

| | |
|---|---|
| `30be03a8b` | restores Liz's class-doc fix, cherry-picked with her authorship |
| `ad15d0c1b` | restores your `1708f445b` — two exceptions, floor scope |
| `a7d5e6fcf` | new test guarding the `expectedRelays` copy |

**Why the first two.** All three threads from the last round are marked resolved, and the description credits `1708f445b` — but the force-push that rewrote `3991f5397` into `2e5e012d7` took both doc commits with it. `git merge-base --is-ancestor 1708f445b HEAD` says no, and the file still read *"The one exception"*. The rewrite is otherwise fine: the code in `2e5e012d7` is equivalent to what it replaced.

**Why the third.** Reverting the `expectedRelays` copy back to `required this.expectedRelays` — main's exact behaviour, the bug you describe as user-visible on its own — left all 479 package tests green. `isTarget` is the only gate on inbound OK frames (`relay_pool.dart:1238`), so a caller that mutates its set after construction can hand any relay the power to decide a publish. Now red without the copy: `Expected: false  Actual: <true>`.

**Checked:** 480 `nostr_sdk` + 43 consumer tests, analyze and format clean, 16/16 CI green. Every consumer of `targetCount` / `unreachableTargets` / `summary` uses them for log or error text only — `reachedNoRelay` in `nip17_message_service.dart:462` reads the other three buckets, so DM retry semantics don't move. Also ran the branch on an iOS sim: a live publish returned `acceptedBy=[] rejectedBy={} noResponseFrom=[wss://relay.divine.video]` — partition intact, floor correctly silent.

One for you: the description still points at `1708f445b`, which is no longer in history. That content is back under `ad15d0c1b`.

mobile/packages/nostr_sdk/lib/relay/publish_outcome.dart
realmeylisdev: Addressed in `65be904fc` rather than left to you — it was a doc line, so it goes in a commit.

The class doc now reads *"normally do not appear — except when the caller named them in `targetRelays`/`tempRelays`, or when the publish reached nothing at all"*, which matches what the `unreachableTargets` field doc below already said. Re-confirmed the mechanism first: `_intendedPublishTargets` guards the offline drop on `!hasFilter` (`relay_pool.dart:2069-2077`) and adds every `tempRelays` entry to `counted` unconditionally (`:2077-2080`), so a named relay is exempt outright — not merely hedged by "normally".

Docs only, no behaviour touched. 480 `nostr_sdk` + 46 consumer tests green, analyze and format clean.

realmeylisdev:
Closed out the last item from my approval and merging on @realmeylisdev's instruction.

**Pushed** — `65be904fc`, docs only. The class-level partition invariant still hedged with *"normally"* and then named just one exception. There are two: `_intendedPublishTargets` guards the offline drop on `!hasFilter` (`relay_pool.dart:2069-2077`) and adds every `tempRelays` entry to `counted` unconditionally, so a caller-named relay is exempt outright. The `unreachableTargets` field doc below already said this correctly; the class doc now matches it.

**Description fixed** — it cited `1708f445b`, which the 2026-08-06 force-push removed from history. Now points at `ad15d0c1b`, where that content actually lives. Also refreshed the verification counts, which had drifted (423 → 480 `nostr_sdk`, 45 → 46 consumers).

**Verified on the merge head** — 480 `nostr_sdk` + 46 consumer tests, `analyze` and `dart format` clean, 16/16 CI green. Nothing executable changed since the approval, so the mutation checks behind it still hold: dropping the `_complete()` floor reddens the count test, `isTarget => true` reddens the bystander test, and reverting the `expectedRelays` copy reddens `a7d5e6fcf`.

One process note: the handbook leaves the merge decision with the author, so this merge is on the operator's explicit call, not something the takeover itself entitled me to.

PR #6612, opened 2026-08-02
docs: track the crash-triage runbook and the codebase audit

rabble:
## What

Three things that had been sitting untracked at the repo root are now tracked, and the rest of the root scratch is gone.

**Added**

| Path | What it is |
|---|---|
| `.claude/runbooks/crashlytics-triage.md` | The hourly Crashlytics triage loop — Firebase project/app IDs, the never-mutate-hosted-state rule, the actionable-crash filter, and the fix-or-file decision. This is the workflow that produced #6180 and #6182. |
| `docs/audits/2026-06-12-codebase-audit.md` | The historical codebase audit that Batch 1 (#5055, #5056, #5057, #5059, #5060, #5080) came out of. Shipped findings are marked fixed throughout, and still-open security findings are summarized publicly while their precise locations and exploit details stay in the private security tracker. |
| `.claude/commands/*.md` (5 files) | Metaswarm slash-command shims — `start`, `pr-shepherd`, `prime`, `review-design`, `self-reflect` — so Claude Code gets entry points alongside the `$name` syntax `AGENTS.md` documents for Codex. Not a one-to-one mirror of that table; see the review fixes below. |

Both docs carry the `Status:` / `Validated against:` header `docs/DOCUMENTATION_GUIDELINES.md` asks for. The audit is marked **Historical** with its shipped findings called out — line numbers and "still present" claims have drifted since June, so it should be re-verified before anyone acts on a finding.

## Why

These were being handed to agents out of band, and every `git status` in the repo showed stray root entries — one bad `git add -A` away from landing on `main` by accident.

## Review fixes

Review findings were handled in separate commits where practical, so each can be reviewed or reverted independently.

**`a40074915` — SEC-01 was inverted, not just stale.** The audit's headline finding said relay events are never signature-verified, and carried a 2026-07-28 note deciding to keep them that way. Proactive verification had already shipped three weeks earlier — #5894 (merged 2026-07-07, closing #5863) put the Schnorr verify on an off-main worker, #5934 (merged 2026-07-11) added the policy setting — and `NostrSignatureVerificationPreferenceService.defaultPolicy` is `NostrSignatureVerificationPolicy.all`, so every relay is verified by default. SEC-01 is now marked fixed in all six places it appears; the decision note is kept but explicitly marked superseded rather than deleted.

[...]

PR #6682, opened 2026-08-04
docs(runbooks): correct the Crashlytics tool surface

### LINKED ISSUE #6683: docs(runbooks): correct the Crashlytics tool surface
Auto-created from PR #6682.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6682
**Branch:** `fix/crashlytics-runbook-tools`

rabble:
Closes #6683

## What

The crash-triage runbook merged in #6612 cannot be executed as written. An agent following it fails on the **first call of Step 1**, and the correct tool names appear nowhere in the document.

## Verified, not guessed

Checked against the `firebase-tools` MCP module installed on this machine (`lib/mcp/tools/crashlytics/`), which registers exactly:

`add_note`, `delete_note`, `get_issue_details`, `get_sample_crash_for_issue`, `list_notes`, `list_top_devices`, `list_top_issues`, `list_top_operating_systems`, `list_top_versions`, `update_issue`

| Runbook said | Reality |
|---|---|
| `crashlytics_create_note` | ✗ → `crashlytics_add_note` |
| `crashlytics_get_report` | ✗ → `crashlytics_list_top_issues` |
| `crashlytics_get_issue` | ✗ → `crashlytics_get_issue_details` |
| `crashlytics_list_events` | ✗ → `crashlytics_get_sample_crash_for_issue` |
| `crashlytics_batch_get_events` | ✗ no batch tool exists |
| `crashlytics_list_notes` | ✓ |
| `crashlytics_update_issue` | ✓ |

## Why the asymmetry mattered

The Hard Rule granted the **only permitted mutation** — adding a note — under a name that does not exist, while both **prohibited** operations (`update_issue`, `delete_note`) were named correctly. So the one sanctioned write was unusable and the two forbidden ones were a copy-paste away. That is the wrong way round for a rule whose entire purpose is "do not mutate hosted data".

## Step 1's query shape was also unsupported

`list_top_issues` takes only `{app_id, issue_count, issue_type}` (`issue_type` ∈ `FATAL` / `NON-FATAL` / `ANR`). There is **no time-window parameter**, so "over the last 7 days" is not expressible, and **no sort parameter**, so "sorted by impacted users" cannot be requested. Fixing the names alone would have left the procedure still unrunnable.

The ranking note now says what an agent can actually do: pull a generous `issue_count`, then sort by impacted users locally from the issue details. The intent of the original note — don't trust event-count order, because it over-weights one user crash-looping — is preserved.

Also documents the real parameter shapes for `get_issue_details`, `get_sample_crash_for_issue`, and points at `list_top_devices` / `list_top_operating_systems` / `list_top_versions` for the breakdown the old text attributed to a nonexistent events tool.

## Missing precondition

This repo's `.mcp.json` registers only `nostr`, `dart` and `figma`. The runbook opened with "Using the Firebase MCP tools" and never said how to get them, so an agent on a clean checkout has zero crashlytics tools. Added a setup step plus an instruction to stop rather than guess if they are unavailable.

## Note

The reviewer that surfaced this reported the tool as `get_sample_crash` (the filename); the **declared tool name** is `get_sample_crash_for_issue`. Worth checking the module rather than the file listing — this PR uses the declared names.

## Scope

Docs only — one file, no code, no CI config.

### DIFF
--- a/.claude/runbooks/crashlytics-triage.md
+++ b/.claude/runbooks/crashlytics-triage.md
@@ -19,8 +19,8 @@ runs should be short — the ledger below tells you what's already handled.
 ## Hard rules (non-negotiable)
 
 1. **NEVER close, delete, or mutate the state of Crashlytics issues, events, or any
-   hosted data.** You may add notes (`crashlytics_create_note`) — nothing else. Do not
-   use `crashlytics_update_issue` to change state; do not delete notes.
+   hosted data.** You may add notes (`crashlytics_add_note`) — nothing else. Do not
+   use `crashlytics_update_issue` to change state; do not use `crashlytics_delete_note`.
 2. **Never run destructive local commands** (`git reset --hard`, `git clean`, `rm -rf`,
    branch force-deletes) on anything you didn't create this run.
 3. One crash worked per run, maximum. If nothing qualifies, report "nothing actionable"
@@ -31,19 +31,28 @@ runs should be short — the ledger below tells you what's already handled.
 
 ## Step 1 — Pull the crash picture
 
+The Firebase MCP server is **not** in this repo's `.mcp.json` (which registers only
+`nostr`, `dart` and `figma`), so the `crashlytics_*` tools are unavailable on a clean
+checkout. Wire it up first — `firebase experimental:mcp` from the `firebase-tools` CLI —
+and confirm the tools are listed before starting. If they are not available, stop and
+say so rather than guessing at crash data.
+
 Using the Firebase MCP tools against project `openvine-co`:
 
-1. `crashlytics_get_report` (top issues) for **both** the iOS and Android apps, over
-   the last 7 days, sorted by impacted users / event count.
-2. For candidates, `crashlytics_get_issue` to get details, and
-   `crashlytics_list_events` / `crashlytics_batch_get_events` for representative stack
-   traces, device/OS breakdown, and versions.
+1. `crashlytics_list_top_issues` for **both** the iOS and Android apps. It takes
+   `{app_id, issue_count, issue_type}` — `issue_type` is one of `FATAL`, `NON-FATAL`,
+   `ANR`. Ask for a generous `issue_count` (30+); see the ranking note below.
+2. For candidates, `crashlytics_get_issue_details` (`{app_id, issue_id}`), and
+   `crashlytics_get_sample_crash_for_issue`
+   (`{app_id, issue_id, sample_count, variant_id}`) for representative stack traces.
+   `crashlytics_list_top_devices`, `crashlytics_list_top_operating_systems` and
+   `crashlytics_list_top_versions` give the device / OS / version breakdown.
 
-Note: the `topIssues` report is sorted by **event count**, but rank your candidates by
-**impacted users** (with `SIGNAL_FRESH` / `SIGNAL_REGRESSED` / velocity as tiebreakers).
-Event count over-weights a single user who crash-loops; impacted-users is the better
-severity signal. Read enough rows that a high-user issue ranked below a high-event one
-isn't missed.
+Note: the tool takes **no time-window and no sort parameter** — it returns the top
+issues as Crashlytics ranks them, by event count. Rank your own candidates by
+**impacted users** from the issue details, because event count over-weights a single
+user who crash-loops. That means pulling more rows than you think you need and sorting
+them yourself, rather than trusting the returned order.
 
 ## Step 2 — Triage filter
 
--- a/.claude/runbooks/crashlytics-triage.md
+++ b/.claude/runbooks/crashlytics-triage.md
@@ -19,7 +19,7 @@ runs should be short — the ledger below tells you what's already handled.
 ## Hard rules (non-negotiable)
 
 1. **NEVER close, delete, or mutate the state of Crashlytics issues, events, or any
-   hosted data.** You may add notes (`crashlytics_add_note`) — nothing else. Do not
+   hosted data.** You may add notes (`crashlytics_create_note`) — nothing else. Do not
    use `crashlytics_update_issue` to change state; do not use `crashlytics_delete_note`.
 2. **Never run destructive local commands** (`git reset --hard`, `git clean`, `rm -rf`,
    branch force-deletes) on anything you didn't create this run.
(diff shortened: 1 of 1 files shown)


rabble (commented):
Reviewed the current head and checks. This is a focused documentation correction with no actionable issue found. I could not submit an approval because GitHub treats me as the PR author; review is complete with no code objections.

rabble:

**I think the premise here is inverted, and merging this would break the runbook rather than fix it.**

I checked against the Firebase MCP server actually loaded in my session (the `firebase` plugin, tool prefix `mcp__plugin_firebase_firebase__*`). It registers exactly these crashlytics tools:

```
crashlytics_get_report        crashlytics_list_events
crashlytics_get_issue         crashlytics_batch_get_events
crashlytics_create_note       crashlytics_list_notes
crashlytics_update_issue      crashlytics_delete_note
```

That is the set the runbook already had. Every rename in this PR points at a name that does not exist on this surface — `crashlytics_add_note`, `crashlytics_list_top_issues`, `crashlytics_get_issue_details`, `crashlytics_get_sample_crash_for_issue`, `crashlytics_list_top_devices` / `_operating_systems` / `_versions`.

Two of the supporting claims also don't hold for `crashlytics_get_report`. I pulled its schema:

- **Time window exists.** `filter.intervalStartTime` / `filter.intervalEndTime` (ISO 8601, must be within 90 days, defaults to 7 days ago → now). So "over the last 7 days" is not only expressible, it is the default.
- **Report selector and page size exist.** `report` is an enum (`topIssues`, `topVariants`, `topVersions`, `topOperatingSystems`, `topAppleDevices`, `topAndroidDevices`) and `pageSize` controls rows.
- `filter` also carries `issueSignals` (`SIGNAL_EARLY` / `SIGNAL_FRESH` / `SIGNAL_REGRESSED` / `SIGNAL_REPETITIVE`), which is what the original ranking note referred to — the PR removes that guidance as unsupported, but it is supported.

Note `issueErrorTypes` uses `NON_FATAL` (underscore), where the PR text writes `NON-FATAL` (hyphen).

**What I think happened:** the PR verified against `firebase-tools`' `lib/mcp/tools/crashlytics/` source on one machine. That is a different tool surface from what the Firebase MCP plugin exposes to agents. Both may be real; they aren't the same thing.

Before landing either version, worth pinning down which surface the runbook's readers actually get, and saying so explicitly in the doc — that ambiguity is the real defect here, and it will keep biting.

**The parts I do think are worth keeping regardless:**

- The missing-precondition section. "Using the Firebase MCP tools" with no setup step, and no instruction to stop rather than guess, is a genuine gap.
- Tightening the Hard Rule to name `crashlytics_delete_note` explicitly instead of prose "do not delete notes".

rabble:
## Review — docs(runbooks): correct the Crashlytics tool surface

(Comment rather than a formal review, since GitHub won't take one from the author.)

**This inverts the correction: the names it removes are the live ones, and the names it adds don't exist.**

I checked against the actual Firebase MCP tool schemas rather than from memory. The server exposes:

```
crashlytics_get_report          crashlytics_get_issue
crashlytics_list_events         crashlytics_batch_get_events
crashlytics_create_note         crashlytics_list_notes
crashlytics_delete_note         crashlytics_update_issue
```

Mapping that against the diff:

| PR replaces | with | Reality |
| --- | --- | --- |
| `crashlytics_create_note` | `crashlytics_add_note` | `create_note` is the real tool; `add_note` does not exist |
| `crashlytics_get_report` | `crashlytics_list_top_issues` | `get_report` is real; `list_top_issues` does not exist |
| `crashlytics_get_issue` | `crashlytics_get_issue_details` | `get_issue` is real; `get_issue_details` does not exist |
| `crashlytics_list_events` / `batch_get_events` | `crashlytics_get_sample_crash_for_issue` | both originals are real; the replacement does not exist |
| — | `crashlytics_list_top_devices` / `_operating_systems` / `_versions` | none exist as tools |

The device/OS/version breakdowns aren't separate tools at all — they're values of `get_report`'s `report` enum: `topIssues`, `topVariants`, `topVersions`, `topOperatingSystems`, `topAppleDevices`, `topAndroidDevices`.

The one line the diff leaves alone (`do not use crashlytics_update_issue to change state`) is correct, and `crashlytics_delete_note` does exist — so the second half of hard-rule 1 is an improvement. Everything else in the rename set goes the wrong way.

### The deleted ranking guidance was right and should stay

```diff
-Note: the `topIssues` report is sorted by **event count**, but rank your candidates by
-**impacted users** (with `SIGNAL_FRESH` / `SIGNAL_REGRESSED` / velocity as tiebreakers).
+Note: the tool takes **no time-window and no sort parameter** …
```

`get_report`'s schema contradicts the replacement directly:

- It **does** take a time window — `filter.intervalStartTime` / `filter.intervalEndTime`, ISO 8601, must be within the last 90 days, defaulting to 7 days ago. The removed "over the last 7 days" instruction was accurate and is now the documented default.
- It **does** take `filter.issueSignals`, accepting exactly `SIGNAL_EARLY` / `SIGNAL_FRESH` / `SIGNAL_REGRESSED` / `SIGNAL_REPETITIVE` — the tiebreakers the diff deletes are first-class filter values, not folklore.
- Its description says the result "aggregates the sum of events **and impacted users**", so ranking by impacted users is directly supported rather than something you have to reconstruct.

It also takes `pageSize` (default 10), which is the parameter the new text is reaching for when it says "ask for a generous `issue_count` (30+)".

### The `.mcp.json` note is half right, and its fix is wrong

True: Firebase isn't in this repo's `.mcp.json` — that file registers only `nostr`, `dart` and `figma`. But the tools are available anyway, because the server comes from the **`firebase@firebase` plugin** enabled in `~/.claude/settings.json`, and `.claude/settings.local.json` already carries allow-entries like `mcp__plugin_firebase_firebase__crashlytics_list_events`.

So the remedy isn't `firebase experimental:mcp` plus a `.mcp.json` edit — it's "enable the `firebase` plugin". Note also that the tool names are namespaced in practice (`mcp__plugin_firebase_firebase__crashlytics_get_report`), which is worth saying if the runbook is going to talk about availability at all.

The underlying instinct — *check the tools are actually there before trusting crash data, and stop rather than guess if they aren't* — is good and worth keeping. It just needs the right setup step.

### Suggested shape

1. Revert the five renames; the originals are correct.
2. Keep the `crashlytics_delete_note` clarification in hard-rule 1.
3. Restore the impacted-users-over-event-count paragraph, and strengthen it by naming `filter.issueSignals` and `pageSize` as the actual parameters.
4. Replace the `.mcp.json` paragraph with the plugin-enablement version, mentioning the `mcp__plugin_firebase_firebase__*` prefix.
5. Optionally add what genuinely isn't documented today: `filter.intervalStartTime` / `intervalEndTime` are capped at 90 days, so a "has this regressed since release N" question has a hard horizon.

Happy to re-check once updated — this is exactly the kind of drift a runbook should fix, it's just pointing at the wrong generation of the tool surface.

PR #5957, opened 2026-07-08
perf(app): eliminate main-isolate CPU hotspots and raster jank

hm21:
Closes #5962

## Description

Profile-driven performance pass over the app's two hot surfaces: main-isolate CPU during feed ingestion and raster-thread frame times on Android. Every fix in this PR traces back to an on-device measurement (DevTools CPU profiles + Flutter frame charts, Android profile builds); there are no speculative optimizations.

### Before 37fps
Bildschirmfoto 2026-07-08 um 09 51 30 

### After 59fps
 Bildschirmfoto 2026-07-08 um 11 30 43 

**Related Issue:** 

### 1. Main-isolate CPU (DevTools CPU profiles, ~2s feed-scroll windows)

| Hotspot | Before | After | Fix |
|---|---:|---:|---|
| `NostrEventsDao` PRE upsert + `jsonDecode` | 23.0% | ~2% | denormalized, indexed `d_tag` column |
| Blurhash decode (UI isolate) | 11.0% | ~3% | memoize `decodeBlurhash` |
| `developer.log` per relay frame | ~5.9% | 0.3% | debug-only via `assert` closure |
| Schnorr verify after remote signing | 3.8% | 0 | moved to background isolate via `compute` |
| NIP-59 gift-wrap verify (iOS profile, live DM burst) | 46.9% | 0 | verify isolate now outlives the drain |

Details per commit:

- **`perf(db_client)`** — PRE upserts loaded *all* `(pubkey, kind)` rows and `jsonDecode`'d each tags blob just to compare d-tags (O(N²) in `upsertEventsBatch`). Additionally, the `event` table had **no indexes at all**: the `List<Index>` getter in `tables.dart` is never picked up by Drift, and unlike `outgoing_dms` there were no runtime `CREATE INDEX` statements — every upsert lookup was a full table scan shipping full rows across the drift-isolate boundary. New nullable `d_tag` column (populated on insert, backfilled once via SQLite `json_each` — no Dart decoding), plus the real index set created through the established runtime `_createMissingTables` path. Upserts now compare against `MAX(created_at)`, which also collapses duplicate versions left behind by raw `cacheEventsBatch` ingestion. The `drift_schema_v1.json` snapshot was stale (8 of 21 tables) and got regenerated.
[...]

Issue #5452, opened 2026-06-22
fix(dm): DM reaction/deletion gift wraps re-decrypt on every app start (no dedup row)


hm21:
## Summary

DM reaction (kind 7) and deletion (kind 5) gift wraps are re-decrypted on **every app start** because they never get a dedup row, so the pre-decrypt check never skips them. On remote-signer (Keycast/Divine-OAuth) accounts each decrypt is a network `nip44_decrypt` RPC (×2 per wrap), so startup hammers the signer with hundreds of redundant, serial RPCs and takes minutes.

## Environment

- App version: dev build, branch `fix/cpu-problems` @ `4286aec1b`
- Device: Android (observed via `flutter logs`)
- Auth: **Keycast / Divine-OAuth (remote signer)** account with substantial DM history (reactions + deletions)
- Network: any (WiFi/cellular)

## Steps to Reproduce

1. Sign in with a **Keycast/Divine-OAuth** account (remote signer) that has DM history containing reactions and reaction-deletions.
2. Launch the app and let the DM inbox sync.
3. Fully kill the app.
4. Launch again and watch the DM sync.

## Actual Result

On **every** launch, hundreds of gift-wrap decryptions run again. With a temporary proof log at the shared decrypt-persist chokepoint (`_persistDecryptedGiftWrap`), every decrypted wrap is **kind 5 (deletion) or kind 7 (reaction)** — and the **same gift-wrap ids recur across restarts**:

```
# Run 1 (13:56:55)
[DM-DEDUP-PROOF] decrypted giftWrap=1027db22c7869eddc8e9e97c28997173d162c6a019c4a2b5b4d7913e39687957 rumorKind=5
[DM-DEDUP-PROOF] decrypted giftWrap=8bb4f4d30d24ee346c8c0f09381984ea3e8bbdaeb8d00a89100ca75eb84c760a rumorKind=7

# Run 2 (13:58:19) — SAME ids, decrypted again
[DM-DEDUP-PROOF] decrypted giftWrap=1027db22c7869eddc8e9e97c28997173d162c6a019c4a2b5b4d7913e39687957 timestamp=2026-06-22 08:36:52Z rumorKind=5
[DM-DEDUP-PROOF] decrypted giftWrap=8bb4f4d30d24ee346c8c0f09381984ea3e8bbdaeb8d00a89100ca75eb84c760a timestamp=2026-06-22 08:00:46Z rumorKind=7
```

[...]

PR #6684, opened 2026-08-04
fix(nostr): drop the per-gift-wrap debug logging from the relay pool

### LINKED ISSUE #6685: fix(nostr): drop the per-gift-wrap debug logging from the relay pool
Follow-up from PR #6684.

## Problem

PR #6684 removes the per-gift-wrap relay-pool logging flood, but it intentionally leaves the underlying replay cycle in place.

After 90 seconds without an inbound relay frame, `WebSocketConnectionManager` treats the socket as idle and forces a disconnect. The timer is inbound-only: `_lastActivityAt` is set on connect and in `_onMessage`, but outbound sends do not reset it. A quiet relay can therefore disconnect in the foreground too, for example while the user sits in the video editor; backgrounding just makes the silence more likely.

On reconnect, the relay re-AUTHs and `RelayPool` replays saved subscriptions. The DM inbox subscription then re-streams its whole steady-state window:

`since: newestSyncedAt - 2d`

That replay costs a re-download plus, per event, a sha256 id recompute and two dedup lookups. It should not redo full crypto work in the normal duplicate path: `_handleGiftWrapEvent` checks `_alreadyProcessed` first, using the message row and the `processed_gift_wraps` ledger from #5452, before building a signer or decrypting. Signature verification is also skipped for known `(id, sig)` pairs via the verify cache.

## Constraint

Do not narrow the 2-day DM window casually. The overlap exists to tolerate NIP-17 randomized outer gift-wrap timestamps, so changing that boundary needs its own correctness review and tests.

## Acceptance Criteria

- Quiet relay periods do not cause unnecessary forced disconnect/reconnect cycles.
- DM subscriptions are not repeatedly replayed solely because no inbound frames arrived for 90 seconds.
- Any change to the 2-day DM overlap preserves NIP-17 randomized-timestamp tolerance and is covered by tests.

## Related

- PR #6684
- #5957
- #5452

hm21:
Closes #6685

## Description

Every kind 1059 frame wrote a log line straight from the relay pool's EVENT hot path - ungated, in every build mode. Backgrounding the app is what makes it obvious: after 90s without an inbound frame the heartbeat force-disconnects the socket (`_onHeartbeat` in `web_socket_connection_manager.dart`), the reconnect re-AUTHs, and the pool replays every saved subscription. The DM inbox subscription then re-streams its whole `since: newestSyncedAt - 2d` window and each replayed wrap prints again.

That idle timer is driven purely by inbound relay frames - `_lastActivityAt` is only touched on connect and in `_onMessage`, outbound sends don't reset it - so the same cycle fires in the foreground whenever a relay stays quiet for 90s, e.g. while the user sits in the video editor. Backgrounding just guarantees it.

Three of the four DM debug sites were already dead code. The subscription id became `dm_inbox_<pubkey>` when it was scoped per user, so the exact-match `dm_inbox` branches - the raw-frame JSON dump and the EOSE line - could not fire anymore. The gift-wrap `eventFilter` line was dead for a different reason: `eventFilters` defaults to `const []`, nothing populates it at runtime, and no `EventFilter` implementation exists in the repo, so that loop body never runs. All four are leftovers from the original NIP-17 data layer (#2118).

The second commit closes the same hole in debug builds. The assert-gated frame log added by #5957 still printed one line per EVENT, so a reconnect replaying a whole subscription window kept flooding debug even after the ungated lines went away - only profile/release went quiet. EVENT is the only high-volume frame in that branch, and OK and COUNT log themselves in their own handlers, so excluding EVENT keeps EOSE and unknown frame types visible without the flood.

The takeover follow-up also stops `RelayPool` from raw-dumping AUTH frames that `RelayBase` already raw-logs before dispatch and the typed AUTH handler logs again.

No behaviour changes beyond logging.

**Related Issue:** #6685

## Out of Scope

The replay itself stays. It costs a re-download plus, per event, a `sha256` id recompute and two dedup lookups - but no crypto: `_handleGiftWrapEvent` checks `_alreadyProcessed` (message row + the `processed_gift_wraps` ledger from #5452) before it builds a signer or decrypts, and the signature verify is skipped via the `(id, sig)` cache. Narrowing the 2-day window would touch the NIP-17 randomized-timestamp tolerance it exists for, which wants its own change.

## Verification

- `dart format` / `flutter analyze` on the changed file - clean
- `flutter test` over all twelve `relay_pool_*` suites (auth, closed-frame, malformed-frame, event-signature, verify-dedup, off-main-verify, search, count, publish-targets, concurrency, zombie-remediation, signature-policy) - 81 tests pass
- Takeover: `mise exec -- dart format lib test integration_test` - no changes
- Takeover: `flutter analyze packages/nostr_sdk/lib/relay/relay_pool.dart` - clean
- Takeover: `flutter test` over all twelve `relay_pool_*` suites - 81 tests pass
- Verified on a real Samsung Galaxy: the log flood is gone

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Code refactor
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
+++ b/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
@@ -703,10 +703,13 @@ class RelayPool {
           messageType == 'CLOSED' ||
           messageType == 'NOTICE') {
         log('📡 Raw message from ${relay.url}: $json');
-      } else {
-        // Debug-only: this branch fires for every EVENT/EOSE frame and the
-        // unconditional log cost ~6% of main-isolate CPU in on-device
-        // profiling. The assert closure never runs in profile/release.
+      } else if (messageType != 'EVENT') {
+        // EVENT is excluded: it is the highest-volume frame, and a reconnect
+        // replays a whole subscription window at once. OK and COUNT log
+        // themselves below, so this mainly covers EOSE and unknown types.
+        // Debug-only — the assert closure never runs in profile/release, and
+        // logging here unconditionally cost ~6% of main-isolate CPU in
+        // on-device profiling (#5957).
         assert(() {
           log('📡 ${relay.url}: $messageType $msgSubId');
           return true;
--- a/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
+++ b/mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
@@ -696,11 +696,10 @@ class RelayPool {
     final messageType = _stringAt(relay, json, 0, 'message type');
     if (messageType == null) return;
 
-    // Log message type + sub ID (full json is too verbose for non-DM events)
+    // Log message type + sub ID (full json is too verbose for data frames)
     if (json.length >= 2) {
       final msgSubId = json[1];
-      if (msgSubId == 'dm_inbox' ||
-          messageType == 'AUTH' ||
+      if (messageType == 'AUTH' ||
           messageType == 'CLOSED' ||
           messageType == 'NOTICE') {
         log('📡 Raw message from ${relay.url}: $json');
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
@hm21 I took over the small follow-up from review.

Changes pushed in 1ff289bd6:
- Removed the pool-side raw AUTH frame dump; `RelayBase` already raw-logs AUTH before dispatch, and the typed AUTH handler logs the challenge path.
- Updated #6685 with the deferred idle-disconnect/replay root-cause details and acceptance criteria.
- Filled the PR body's Related Issue field and added the takeover verification.

Verification:
- `mise exec -- dart format lib test integration_test` - no changes
- `flutter analyze packages/nostr_sdk/lib/relay/relay_pool.dart` - clean
- `flutter test` over all twelve `relay_pool_*` suites - 81 tests pass
- GitHub checks on 1ff289bd6 are green

NotThatKindOfDrLiz (approved):
Final pass is clean.

This PR now only removes noisy/duplicate relay logging, keeps the typed handlers intact, has the follow-up replay issue documented in #6685, and all checks on the latest head are green.

PR #6686, opened 2026-08-04
fix(profile): grow late-arriving header blocks in instead of snapping

### LINKED ISSUE #6687: fix(profile): grow late-arriving header blocks in instead of snapping
Auto-created from PR #6686.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6686
**Branch:** `fix/profile-header-reveal-jumps`

hm21:
Closes #6687

## Description

Badges, verified claims and profile fields each resolve on their own timing, well after the profile header has painted. Every arrival flipped its block from `SizedBox.shrink()` to full height in a single frame, shoving the bio, stats row, action buttons and the whole grid below it down.

Each of those blocks now renders through a shared `_ProfileHeaderReveal`: an `AnimatedSize` that grows the block into place while an `AnimatedSwitcher` fades the content in. Siblings below slide down over the same window instead of teleporting.

Applied to the accepted-badges row, the bio, the website row, the tip/support button and the verified-accounts row.

Three things worth calling out for review:

- **`AnimatedSwitcher` alone would not have fixed this.** It stacks the outgoing and incoming child and takes the larger size immediately, so the content would fade in while the layout still jumped. The height has to come from `AnimatedSize`.
- **The wrapper stays mounted while empty** (`child: null`) instead of being dropped by a collection-if, which is why the bio and website rows moved off `if (...) ...[]`. A wrapper that only appears together with its content mounts at full size and never animates.
- **Reserving space during loading was the alternative and is worse.** Most profiles have no badges and no verified claims, so the majority would trade a jump on arrival for a permanent gap that collapses instead.

As a side effect the bio's own "Show more" expansion is now animated too — it used to snap open.

**Related Issue:** 

## Out of Scope

The first profile load, where the identity skeleton turns off, still jumps. `Skeletonizer` swaps `SkeletonizerRenderObjectWidget` for a `KeyedSubtree` under a different key when it disables, which unmounts and rebuilds its entire subtree — so every `AnimatedSize` inside mounts at full size and has nothing to animate from. Fixing that one means giving the skeleton a bio-shaped placeholder so the height barely changes across the swap, which is a separate design decision (profiles without a bio would then shrink at the swap).

That window swallows more than the identity block: the bio, the website row and the tip button all arrive *with* the profile, which is the same moment the skeleton disables, so on a cold load those three still mount at full size and snap. Only the badge row and the verified-accounts row resolve after that point and reliably grow in. The other three still earn their wrapper on a cached profile (no skeleton, so no remount), on an edit-profile round trip, and — for the bio — on "Show more".

Wrapping the `Skeletonizer` itself in an `AnimatedSize` was considered and rejected: it nests over the inner reveals, and a nested outer animator clips everything below each inner reveal for its duration instead of letting it slide — worse than the current behaviour.

## Verification

- `flutter test test/widgets/profile` — 281 passed
- Two new tests: the badge row grows in rather than appearing at full height, and the bio is still at its collapsed height on the frame after "Show more" is tapped
- Manually verified on a real Samsung Galaxy: opened profiles with and without badges, watched the badge row and the verified-accounts row land, expanded and collapsed a long bio

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/widgets/profile/profile_header_identity.dart
+++ b/mobile/lib/widgets/profile/profile_header_identity.dart
@@ -142,18 +142,19 @@ class _ProfileHeaderReveal extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final duration = MediaQuery.disableAnimationsOf(context)
-        ? Duration.zero
-        : _duration;
+    // The empty state spans the full width so only the height animates.
+    final content = child ?? const SizedBox(width: double.infinity);
+    // Drop the animators entirely under reduce motion rather than passing
+    // them Duration.zero: a zero-duration AnimatedSize completes its
+    // controller synchronously inside performLayout, which re-dirties the
+    // render object mid-layout and trips a framework assertion.
+    if (MediaQuery.disableAnimationsOf(context)) return content;
+
     return AnimatedSize(
-      duration: duration,
+      duration: _duration,
       curve: Curves.easeOutCubic,
       alignment: Alignment.topCenter,
-      child: AnimatedSwitcher(
-        duration: duration,
-        // The empty state spans the full width so only the height animates.
-        child: child ?? const SizedBox(width: double.infinity),
-      ),
+      child: AnimatedSwitcher(duration: _duration, child: content),
     );
   }
 }
--- a/mobile/lib/widgets/profile/profile_header_identity.dart
+++ b/mobile/lib/widgets/profile/profile_header_identity.dart
@@ -165,8 +165,9 @@ class _ProfileBadgesBlock extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    // Loading and error both read as "no badges": most profiles have none, so
-    // reserving a placeholder row would trade one jump for a worse one.
+    // A refresh keeps the badges it already had; a first load and an outright
+    // error read as "no badges". Most profiles have none, so reserving a
+    // placeholder row would trade one jump for a worse one.
     final items =
         ref.watch(profileAcceptedBadgesProvider(userIdHex)).value ??
         const <ProfileBadgeViewData>[];
--- a/mobile/lib/widgets/profile/profile_header_identity.dart
+++ b/mobile/lib/widgets/profile/profile_header_identity.dart
@@ -89,17 +89,27 @@ class _ProfileNameAndBio extends StatelessWidget {
           padding: inset,
           child: Column(
             children: [
-              if (about != null && about!.isNotEmpty) ...[
-                const SizedBox(height: 16),
-                Skeleton.keep(child: _AboutText(about: about!)),
-              ],
+              _ProfileHeaderReveal(
+                child: about != null && about!.isNotEmpty
+                    ? Padding(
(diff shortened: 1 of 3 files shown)


hm21:
Pushed a follow-up: the reveal wrapper passed `AnimatedSize` a zero duration under reduce motion, which does not skip the animation but completes the controller synchronously inside `performLayout` — it would have thrown `A RenderAnimatedSize was mutated in its own performLayout implementation` on every badge or claims arrival for those users. That path now drops both animators instead.

Surfaced while building the same fix for the metadata sheet in #6690, which extracts this into a `divine_ui` primitive (`AnimatedReveal`). Once that merges I will point this PR at it and drop the private copy — not stacking, this one stands on its own.

mbradley:
@hm21 Review in progress on this one, findings to follow shortly.

mbradley (approved):
Approving. Three things, none blocking.

**`inbox_view.dart:640` has the same reduce-motion crash.** It still passes `Duration.zero` to an `AnimatedSize` under reduce motion, same alignment and same `SizedBox(width: double.infinity)` empty state. I reproduced the assert standalone with that shape and no `AnimatedSwitcher` child, so it is the general case: `A RenderAnimatedSize was mutated in its own performLayout implementation.` Focusing the inbox search field should hit it. Follow-up issue, or fold in the one-liner?

**Collapse is asymmetric.** Present to absent holds full height for the whole 220ms switcher out-transition before `AnimatedSize` starts, so clearing a bio leaves a blank hole for roughly a quarter second and then closes. `AnimatedSwitcher`'s Stack keeps the outgoing child's height, so the size never changes until it is evicted. Measured:

```
t=0ms   h=100.0
t=220ms h=100.0
t=330ms h=100.0   (outgoing evicted here)
t=450ms h=9.74
```

Accept and document the asymmetry, drop the `AnimatedSwitcher`, or pass a `layoutBuilder` that stops previous children dictating the Stack size?

**`_ProfileHeaderReveal` is the fifth instance of this idiom** (`inbox_view.dart:640`, `user_picker_sheet.dart:406`, `video_metadata_tags_selector.dart:276`, `divine_ui/loading_overlay.dart:47`) and the first to name it. Keep private, promote to `divine_ui`, or land private with a tracking issue? `divine_ui` is a strict 100%-coverage package, so promoting means covering all three branches here.

One thing not to change: the switcher children are unkeyed, so a same-type content change updates in place. The height still animates, and keying the badge row would remount it and reset its horizontal scroll offset.

mbradley:
Two of my three points were already answered before I posted, my miss.

#6690 extracts `AnimatedReveal` into `divine_ui`, which you'd noted on this PR. That settles the promotion question and gives the private copy a stated removal path.

It also resolves the collapse asymmetry structurally rather than needing a fix here: `AnimatedReveal` drops the `AnimatedSwitcher` for a `TweenAnimationBuilder`, so nothing holds the outgoing height and the collapse starts on frame one. That was the option I'd have picked. `_startsRevealed` covers the already-present case too.

What's left is the inbox. `inbox_view.dart:640` is the same shape with the same latent assert, and `AnimatedReveal` is now the natural home for that fix.

PR #6688, opened 2026-08-04
fix(categories): give the category gallery a Material ancestor

### LINKED ISSUE #6689: fix(categories): give the category gallery a Material ancestor
Auto-created from PR #6688.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6688
**Branch:** `fix/category-gallery-material-ancestor`

hm21:
Closes #6689

## Description

The category gallery renders its retry button and its texts with a yellow double underline, and the retry button is a stock `ElevatedButton` rather than a design-system button.

Both come from the same root cause: `CategoryGalleryScreen` is registered in `searchRoutes()` as a top-level route (sibling of `shellRoutes()`), so it mounts on the root navigator with no `Scaffold`/`Material` above it. `MaterialApp` installs Flutter's fallback text style in that situation — `decoration: underline`, `decorationColor: 0xFFFFFF00`, `debugLabel: 'fallback style; consider putting your text in a Material'` — and every `Text` whose own style leaves `decoration` unset inherits it. The header title already carried a `.copyWith(decoration: TextDecoration.none)` workaround for exactly this.

Fix at the root instead of per-`Text`: the view's root `ColoredBox` becomes a `Material` with the same background color, which supplies a real default text style to the whole subtree. The header workaround is then dead and removed.

While in the error/empty states, they move onto the design system: `ElevatedButton` → `DivineButton`, and the two raw `TextStyle(fontSize: 16)` → `VineTheme.bodyLargeFont`. That empties this file's `raw_textstyle` (2 → 0) and `material_buttons` (1 → 0) ceiling entries, so both baselines are regenerated — a drop to zero fails CI otherwise.

The second commit removes the `Scaffold` from the test helper. That `Scaffold` supplied the `Material` the router never does, which is precisely why the existing suite stayed green while the screen shipped yellow underlined text. Mounting the view bare makes every test exercise the real condition, not just the new regression test.

## Out of Scope

- The hardcoded header colors in this file (`_categoryHeaderActionFill`, `_categoryHeaderActionBorder`, shadows) stay as-is; they are a separate design-system migration.
- Other root-navigator routes were not audited for the same missing-`Material` pattern.

## Verification

- `flutter analyze lib/screens/category_gallery_screen.dart test/screens/category_gallery_screen_test.dart` — clean
- `flutter test test/screens/category_gallery_screen_test.dart` — 7/7 pass
- New regression test pumps `CategoryGalleryView` **without** a `Scaffold` and asserts the inherited style carries no underline. Verified it fails when the `Material` wrapper is reverted to `ColoredBox`, and passes with it — re-confirmed after the test-helper change.
- `check_raw_textstyle_ceiling.sh` and `check_material_button_ceiling.sh` — both OK
- Manually verified on a real Samsung Galaxy: the gallery texts and the retry button render without the yellow double underline, and the retry button renders as a `DivineButton`.

## Type of Change

### DIFF
--- a/mobile/lib/screens/category_gallery_screen.dart
+++ b/mobile/lib/screens/category_gallery_screen.dart
@@ -154,7 +154,10 @@ class CategoryGalleryView extends StatelessWidget {
   Widget build(BuildContext context) {
     final visuals = CategoryVisuals.forCategory(category, 0);
 
-    return ColoredBox(
+    // This screen is a root-navigator route, so nothing above it provides a
+    // `Material`. Without one every `Text` inherits Flutter's yellow
+    // double-underline fallback style from `MaterialApp`.
+    return Material(
       color: context.vineColors.surfaceContainerHigh,
       child: Column(
         children: [
@@ -208,18 +211,17 @@ class _CategoryGalleryBody extends StatelessWidget {
         return Center(
           child: Column(
             mainAxisSize: MainAxisSize.min,
+            spacing: 16,
             children: [
               Text(
                 context.l10n.categoryGalleryCouldNotLoadVideos,
-                style: TextStyle(
+                style: VineTheme.bodyLargeFont(
                   color: context.vineColors.secondaryText,
-                  fontSize: 16,
                 ),
               ),
-              const SizedBox(height: 16),
-              ElevatedButton(
+              DivineButton(
+                label: context.l10n.commonRetry,
                 onPressed: onRetry,
-                child: Text(context.l10n.commonRetry),
               ),
             ],
           ),
@@ -229,9 +231,8 @@ class _CategoryGalleryBody extends StatelessWidget {
           return Center(
             child: Text(
               context.l10n.categoryGalleryNoVideosInCategory,
-              style: TextStyle(
+              style: VineTheme.bodyLargeFont(
                 color: context.vineColors.secondaryText,
-                fontSize: 16,
               ),
             ),
           );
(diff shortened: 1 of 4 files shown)


realmeylisdev (approved):
Approving — diagnosis and fix both hold up.

Traced the root cause to the SDK rather than taking it on trust: `MaterialApp` hands `_errorTextStyle` (underline, `0xFFFFFF00`, double) to `WidgetsApp.textStyle` (`material/app.dart:45`, used at `:1091`), and `Text` merges its own style *over* the inherited one — so any style that leaves `decoration` null keeps the underline. The route is a top-level entry in `searchRoutes()` spread beside `shellRoutes()` (`app_router.dart:114-126`), and `buildAppShell` (`main.dart:2671`) adds no `Material`. Nothing above it supplied one.

Mutation-tested the guard rather than trusting the green run:

- `Material` → `ColoredBox`, keeping the new bare helper → **only** the new test fails, on the underline matcher.
- `Material` → `ColoredBox` **and** the `Scaffold` restored in `buildSubject` → all 7 pass.

So the second commit is load-bearing, not tidy-up. Good catch — that `Scaffold` was the reason the suite stayed green through the whole bug.

Ran locally: 7/7 tests, `flutter analyze` and `dart format` clean, all four design-system ratchets green (`raw_colors` 5 and `raw_dialogs` 1 for this file correctly untouched), plus the skip / test-structure / untested-services floors.

One note on your open item — the other root-navigator routes: `HashtagFeedScreen`, `SearchResultsPage` and `RouteErrorScreen` all already use `Scaffold`, so this screen was the only gap in `search_routes.dart`. The bare `Material` is the right call here; nothing on the screen reaches for a Scaffold-only API.

PR #6690, opened 2026-08-04
fix(metadata): stop the late sheet sections jumping, and cap the reposter chips

### LINKED ISSUE #6695: fix(metadata): stop the late sheet sections jumping, and cap the reposter chips
## Summary

The video metadata sheet had two late-arriving sections that could snap into place after the sheet was already open: Collaborators and Reposted by. User chip labels also changed after profile lookups completed, causing row reflow. The Reposted by section rendered one chip per reposter with no cap, which made the sheet scale with video popularity and could trigger many profile fetches at once.

## Expected behavior

Late-arriving metadata sections should reveal without abrupt layout jumps. User chip rows should avoid common label-swap reflows, but slow profile lookups should not hide the section indefinitely. Reposted by should render a bounded number of chips and provide a path to the full reposters list.

## Scope

- Add a reusable animated reveal primitive for late-arriving content.
- Use it for Collaborators and Reposted by in the metadata sheet.
- Cap visible reposter chips and add a +N more path to the existing reposters list screen.
- Track repository-level unbounded reposter relay queries separately in #6742.

hm21:
Closes #6695

## Description

Three of the metadata sheet's sections resolve after the sheet has already opened, and each landed badly. This fixes all three symptoms of the same thing, and puts a hard ceiling on how much the **Reposted by** section is allowed to render: a video with thousands of reposts now shows five chips and a button instead of thousands of chips.

**1. Sections snapped into place.** **Reposted by** waits on a relay round-trip through `VideoRepostersCubit`, and **Collaborators** waits on each tagged collaborator's confirmation status. Both rendered `SizedBox.shrink()` until their data arrived and then took full height in a single frame — and because they sit in the middle of the list, every section below them jumped down while the user was reading. Both now render through `AnimatedReveal`, a new `divine_ui` primitive: an `AnimatedSize` that grows the block into place while the content fades in.

**2. Chip labels swapped a moment after that.** A user chip falls back to a generated "Adjective Animal Number" name while its profile loads. Chip width follows the name, so every swap reflowed the row — a second round of jumps right after the section had finished growing in. It happened on every open, not only for strangers: `fetchUserProfile` awaits the local cache lookup, so even a cached profile costs a frame with the fallback showing. Both sections now hold their reveal until every visible chip's profile has settled, with a one-second grace window so slow or unavailable profile fetches reveal fallback names instead of hiding the whole section behind the network tail.

**3. Reposted by rendered one chip per reposter, with no upper bound at all.** Every chip resolves its own profile through `fetchUserProfileProvider`, and the `Wrap` is a single list item, so all of them are built and laid out at once. On a popular video that is thousands of chips and thousands of profile fetches fired the moment the sheet opens — the sheet's cost scaled directly with how successful the video was.

It now renders at most five chips no matter how many reposters there are (three rows on a phone, measured on device), and hands the rest to the reposters list screen through a `+N more` button. That screen already exists at `/video/:eventId/reposters` and is reachable from the repost button's long press, so this is a second entry point rather than new surface. An exact row count is not available at build time: chip width follows the display name, and those resolve asynchronously per chip — anything measuring rows would have to wait for every profile first, which is the cost being avoided.

Four things worth calling out for review:

- **`AnimatedSwitcher` cannot do this job.** It stacks the outgoing and incoming child, adopts the larger size at once — so the content fades while the layout still jumps — and hands its children loose constraints, which shrink-wraps and centres a section the surrounding layout meant to stretch. `AnimatedReveal` fades with an `Opacity` driven by `TweenAnimationBuilder` instead, and stays layout-transparent.
- **A null `child` is the empty state, and the widget stays mounted while empty.** That is the whole point of the API shape — a wrapper that only appears together with its content mounts at its final size and has nothing to animate from.
- **Under reduce motion the widget drops both animators** instead of passing them `Duration.zero`. A zero-duration `AnimatedSize` completes its controller synchronously inside `performLayout`, which re-dirties the render object mid-layout and trips `A RenderAnimatedSize was mutated in its own performLayout implementation`. The `divine_ui` coverage gate is what surfaced this — the accessibility path had no test before.
- **`metadataMoreReposters` is a new key, translated in all 22 locales.** Its copy is identical to `profileBadgeMoreRecipients`, which is deliberately not reused: that key is scoped to the badge sheet and its `description` says so. It is also untranslated in 18 locales, so reusing it would have spread that debt.

**Related Issue:** 

## Out of Scope

**Creator and Inspired-by still show the fallback name briefly.** Gating them would trade a label swap for a real jump: they render immediately today and sit above everything else, so appearing late would push the whole sheet down. Neither reflows anything — each is a lone chip on its own row.

**The reposters fetch is still unbounded, only the rendering is capped.** `RepostsRepository.fetchEventReposters` builds `Filter(kinds: [6, 16], e: [eventId])` with no `limit`, even though `_defaultRepostFetchLimit = 500` exists and three sibling methods in that file pass it. It then queries deletions with `Filter(kinds: [5], authors: <every reposter pubkey>)` — a REQ carrying thousands of authors on a popular video. The UI cap stops the expensive client-side part (profile fetches and widget build); the two relay queries want a follow-up in `reposts_repository`, which is a strict-coverage package and deserves its own PR. Follow-up: #6742.

**The first profile load in the profile header still snaps.** #6686 fixes the same class of jump there and carries its own private copy of this widget, since it was opened before the primitive existed. Once this lands I will point that PR at `AnimatedReveal` and drop the copy.

#6692 was the standalone version of the cap. It is superseded by this PR and closed — the two changes overlapped in the same widget, and splitting them made the branch untestable in isolation.

## Verification

- Takeover verification: `flutter test test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart` — passed; `flutter analyze` — no issues.
- `flutter test test/widgets/video_feed_item/metadata` — 59 passed
- `flutter test --coverage` in `packages/divine_ui` — 831 passed, every line of `animated_reveal.dart` covered
- `flutter test test/l10n/arb_consistency_test.dart` — passed, `flutter gen-l10n` output committed
- `check_raw_colors_ceiling.sh` / `check_raw_textstyle_ceiling.sh` — no growth
- New tests: five for `AnimatedReveal` itself (renders, zero height while empty, grows on arrival, keeps the parent's constraints instead of centring, lands instantly under reduce motion); the Reposted-by section still collapsed on the frame after the relay answers; the section hidden while a profile is pending, fallback reveal after the one-second grace window, and 12 reposters rendering 5 chips plus a `+7 more` button
- Manually verified on a real Samsung Galaxy: opened the metadata sheet on videos with reposters and with collaborators, watched both sections grow in with final labels, checked the chip count and the `+N more` button

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -811,6 +811,7 @@
     "metadataCollaboratorsLabel": "ተባባሪዎች",
     "metadataInspiredByLabel": "ተመስጦ",
     "metadataRepostedByLabel": "በድጋሚ የተለጠፈው በ",
+    "metadataMoreReposters": "+{count} ተጨማሪ",
     "metadataLoopsLabel": "{count, plural, =1{ሉፕ} other{ሉፖች}}",
     "@metadataLoopsLabel": {
         "placeholders": {
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -692,6 +692,7 @@
     "metadataCollaboratorsLabel": "المتعاونون",
     "metadataInspiredByLabel": "مستوحى من",
     "metadataRepostedByLabel": "أعاد نشره",
+    "metadataMoreReposters": "+{count} آخرين",
     "metadataLoopsLabel": "التكرارات",
     "metadataLikesLabel": "الإعجابات",
     "metadataCommentsLabel": "التعليقات",
--- a/mobile/lib/l10n/app_bg.arb
+++ b/mobile/lib/l10n/app_bg.arb
@@ -756,6 +756,7 @@
     "metadataCollaboratorsLabel": "Сътрудници",
     "metadataInspiredByLabel": "Вдъхновено от",
     "metadataRepostedByLabel": "Повторно публикувано от",
+    "metadataMoreReposters": "още {count}",
     "metadataLoopsLabel": "{count, plural, =1{Луп} other{Лупове}}",
     "@metadataLoopsLabel": {
         "placeholders": {
--- a/mobile/lib/l10n/app_de.arb
+++ b/mobile/lib/l10n/app_de.arb
@@ -695,6 +695,7 @@
     "metadataCollaboratorsLabel": "Mitwirkende",
     "metadataInspiredByLabel": "Inspiriert von",
     "metadataRepostedByLabel": "Repostet von",
+    "metadataMoreReposters": "+{count} weitere",
     "metadataLoopsLabel": "Loops",
     "metadataLikesLabel": "Likes",
     "metadataCommentsLabel": "Kommentare",
--- a/mobile/lib/l10n/app_en.arb
+++ b/mobile/lib/l10n/app_en.arb
@@ -1053,6 +1053,15 @@
   "metadataCollaboratorsLabel": "Collaborators",
   "metadataInspiredByLabel": "Inspired by",
   "metadataRepostedByLabel": "Reposted by",
+  "metadataMoreReposters": "+{count} more",
+  "@metadataMoreReposters": {
+    "description": "Chip in the video metadata sheet opening the full reposters list, showing how many reposters are not displayed.",
+    "placeholders": {
+      "count": {
+        "type": "int"
+      }
+    }
+  },
   "metadataLoopsLabel": "{count, plural, =1{Loop} other{Loops}}",
   "@metadataLoopsLabel": {
     "placeholders": {
--- a/mobile/lib/l10n/app_es.arb
+++ b/mobile/lib/l10n/app_es.arb
@@ -695,6 +695,7 @@
     "metadataCollaboratorsLabel": "Colaboradores",
     "metadataInspiredByLabel": "Inspirado en",
     "metadataRepostedByLabel": "Reposteado por",
+    "metadataMoreReposters": "+{count} más",
     "metadataLoopsLabel": "Loops",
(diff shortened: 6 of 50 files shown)


NotThatKindOfDrLiz:
@hm21 I took over the remaining review fixes and pushed af9b808.

What changed:
- Reposter overflow now counts against the capped candidate row, so the late-widen window no longer flashes a wrong +N more button.
- Profile-name settling now treats hasValue/hasError as settled, so provider re-execution with previous data does not collapse Collaborators.
- Product call: sections wait up to one second for final chip names, then reveal fallback names rather than staying invisible behind the 15s REST + 5s relay tail.
- Added regression coverage for the grace-window fallback reveal and for suppressing the transient more button while a widened row is pending.
- Filed #6742 for the remaining unbounded RepostsRepository.fetchEventReposters relay-query debt, updated #6695 with a real body, and updated this PR body/verification notes.

Verification:
- flutter test test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart
- flutter analyze
- Pre-push hook: analyzer plus changed-file tests, including packages/divine_ui/test/src/animated_reveal_test.dart and metadata_expanded_sheet_test.dart, 62 passed
- GitHub checks are all green on af9b808.

NotThatKindOfDrLiz (approved):
Approved after takeover. The remaining correctness issues are fixed, the slow-profile product tradeoff is explicit and test-backed, the repository-level follow-up is tracked in #6742, and all checks are green on af9b808.

PR #6692, opened 2026-08-04
fix(metadata): cap the reposter chips instead of rendering every one

### LINKED ISSUE #6693: fix(metadata): cap the reposter chips instead of rendering every one
Auto-created from PR #6692.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6692
**Branch:** `fix/metadata-reposters-cap`

hm21:
Closes #6693

## Description

The **Reposted by** section rendered one chip per reposter with no upper bound, and every chip resolves its own profile through `fetchUserProfileProvider`. The `Wrap` is a single list item, so all of them are built and laid out at once — a video with thousands of reposts fired thousands of profile fetches and built thousands of chips the moment the sheet opened.

It now shows at most nine and hands the rest to the reposters list screen through a `+N more` chip. That screen already exists (`/video/:eventId/reposters`) and is reachable from the repost button's long press, so this is a second entry point rather than new surface.

Two notes for review:

- **Nine is "roughly three rows", not exactly three.** An exact row count is not available at build time: chip width follows the display name, and those resolve asynchronously per chip. Anything that measured rows would have to wait for every profile to load first, which is the cost this is avoiding. Happy to change the number.
- **`metadataMoreReposters` is a new key, translated in all 22 locales** rather than mirrored as English. Its copy is identical to `profileBadgeMoreRecipients`, which is deliberately not reused: that key is scoped to the badge sheet and its `description` says so. (It is also untranslated in 18 locales, so reusing it would have spread that debt.)

**Related Issue:** 

## Out of Scope

The fetch itself is still unbounded. `RepostsRepository.fetchEventReposters` builds `Filter(kinds: [6, 16], e: [eventId])` with no `limit`, even though `_defaultRepostFetchLimit = 500` exists and three sibling methods in that file pass it. It then queries deletions with `Filter(kinds: [5], authors: <every reposter pubkey>)` — a REQ carrying thousands of authors on a popular video. The UI cap stops the expensive part (profile fetches and widget build), but the two relay queries are worth a follow-up in `reposts_repository`; it is a strict-coverage package, so it wants its own PR and tests.

Touches the same widget as #6690, which is still open. Whichever lands first, I will rebase the other — they are independent changes that happen to overlap in `_RepostedByContent`.

## Verification

- `flutter test test/widgets/video_feed_item/metadata` — 57 passed
- `flutter test test/l10n/arb_consistency_test.dart` — passed, and `flutter gen-l10n` output is committed
- `check_raw_colors_ceiling.sh` / `check_raw_textstyle_ceiling.sh` — no growth
- New tests: 12 reposters render 9 chips plus a `+3 more` chip, and a single reposter renders no `more` affordance
- Manually verified on a real Samsung Galaxy: opened the metadata sheet on a video with many reposters, checked the chip count and that `+N more` opens the full list

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -811,6 +811,7 @@
     "metadataCollaboratorsLabel": "ተባባሪዎች",
     "metadataInspiredByLabel": "ተመስጦ",
     "metadataRepostedByLabel": "በድጋሚ የተለጠፈው በ",
+    "metadataMoreReposters": "+{count} ተጨማሪ",
     "metadataLoopsLabel": "{count, plural, =1{ሉፕ} other{ሉፖች}}",
     "@metadataLoopsLabel": {
         "placeholders": {
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -692,6 +692,7 @@
     "metadataCollaboratorsLabel": "المتعاونون",
     "metadataInspiredByLabel": "مستوحى من",
     "metadataRepostedByLabel": "أعاد نشره",
+    "metadataMoreReposters": "+{count} آخرين",
     "metadataLoopsLabel": "التكرارات",
     "metadataLikesLabel": "الإعجابات",
     "metadataCommentsLabel": "التعليقات",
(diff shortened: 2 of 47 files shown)


hm21:
Superseded by #6690, which now carries the cap alongside the reveal animation. The two changes overlapped in `_RepostedByContent`, so keeping them on separate branches made neither testable on its own — the cap branch had no animation and the animation branch had no cap.

The version that landed there also differs: five chips instead of nine (nine came out at five rows on a phone) and a `DivineButton` instead of a hand-rolled pill.

PR #6703, opened 2026-08-04
fix(router): defer redirect refresh notifications

### LINKED ISSUE #6117: fix(router): concurrent modification during go_router redirect (Crashlytics c0395b3f)
## Crashlytics
- Issue: `c0395b3fa4717824b629ee93c07d69f4` ([console](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/c0395b3fa4717824b629ee93c07d69f4))
- Type: non-fatal · **33 impacted users / 1683 events**
- Versions: first 1.0.15 → last **1.0.16**

## Signature
`package:go_router/src/configuration.dart - RouteConfiguration._runInRouterZone`
`GoException: Exception during redirect: Concurrent modification during iteration: Instance(length:19) of '_GrowableList'` (thrown, runZonedGuarded).

## Root cause (hypothesis)
A `redirect` callback iterates a growable list that is mutated during the same iteration — a redirect reads a collection (routes, auth/guard state, or a pending-actions list) while another path mutates it. High event count on few users = a redirect loop retriggering. Introduced/surfaced in 1.0.15.

## Suggested next step
Audit the app's `redirect` callbacks in `app_router.dart` for iteration over a mutable list (snapshot with `.toList()` before iterating, or make the read side immutable). Reproduce by driving the redirect with a concurrent mutation.

NotThatKindOfDrLiz:
## Description

- Defer and coalesce RouterRefreshListenable notifications so GoRouter redirect reevaluation does not run synchronously inside provider/listener updates.
- Preserve the existing lazy empty-following redirect lookup so router construction does not require SharedPreferences and the first-navigation decision reads the current cache at redirect time.
- Add regression coverage for refresh deferral, coalescing, and disposal.

**Related Issue:** Closes #6117

## Out of Scope

- Full Crashlytics console stack was not available from the GitHub issue; the issue body only exposes the top go_router frame.
- Broader route-level redirect audits are left untouched.
- Prewarming following-cache state for redirects is left out intentionally; the router should not gain a construction-time SharedPreferences dependency in this fix.

## Verification

- flutter test test/router/route_error_routing_test.dart test/router/empty_contacts_redirect_test.dart test/router/router_refresh_listenable_test.dart
- flutter test test/router/app_router_redirect_test.dart test/router/app_router_test.dart test/router/router_refresh_gate_coupling_test.dart
- flutter analyze lib/router test/router/router_refresh_listenable_test.dart test/router/empty_contacts_redirect_test.dart test/router/route_error_routing_test.dart test/router/app_router_redirect_test.dart test/router/app_router_test.dart test/router/router_refresh_gate_coupling_test.dart
- Pre-push hook: analyzer plus changed-file router test test/router/router_refresh_listenable_test.dart

## Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Code refactor
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/mobile/lib/router/app_router.dart
+++ b/mobile/lib/router/app_router.dart
@@ -30,7 +30,6 @@ import 'package:openvine/screens/auth/email_verification_screen.dart';
 import 'package:openvine/screens/auth/nostr_connect_screen.dart';
 import 'package:openvine/screens/auth/reset_password.dart';
 import 'package:openvine/screens/auth/welcome_screen.dart';
-import 'package:openvine/screens/explore/explore_screen.dart';
 import 'package:openvine/screens/feed/video_feed_page.dart';
 import 'package:openvine/screens/inbox/conversation/conversation_page.dart';
 import 'package:openvine/screens/key_import_screen.dart';
@@ -78,12 +77,6 @@ final goRouterProvider = Provider<GoRouter>((ref) {
   final refreshListenable = RouterRefreshListenable(
     authService.authStateStream,
   );
-  var hasFollowingInCache = ref.read(hasFollowingInCacheProvider);
-  ref.listen(hasFollowingInCacheProvider, (previous, next) {
-    if (previous == next) return;
-    hasFollowingInCache = next;
-    refreshListenable.refresh();
-  });
   ref.listen(currentMinorAccountReviewStatusProvider, (previous, next) {
     // A resume/background refetch that resolves to a routing-identical status
     // (active → active) must not refresh: refreshing churns the route pipeline
@@ -106,8 +99,7 @@ final goRouterProvider = Provider<GoRouter>((ref) {
     refreshListenable: refreshListenable,
     errorBuilder: (context, state) =>
         RouteErrorScreen(message: context.l10n.routeUnknownPath),
-    redirect: (context, state) =>
-        appRouterRedirect(ref, state, hasFollowingInCache: hasFollowingInCache),
+    redirect: (context, state) => appRouterRedirect(ref, state),
     // go_router matches same-segment-count routes top-to-bottom, so spread
     // order below is load-bearing whenever two routes share a literal first
     // path segment. It's match-safe today because every module's
--- a/mobile/lib/router/app_router.dart
+++ b/mobile/lib/router/app_router.dart
@@ -30,6 +30,7 @@ import 'package:openvine/screens/auth/email_verification_screen.dart';
 import 'package:openvine/screens/auth/nostr_connect_screen.dart';
 import 'package:openvine/screens/auth/reset_password.dart';
 import 'package:openvine/screens/auth/welcome_screen.dart';
+import 'package:openvine/screens/explore/explore_screen.dart';
 import 'package:openvine/screens/feed/video_feed_page.dart';
 import 'package:openvine/screens/inbox/conversation/conversation_page.dart';
 import 'package:openvine/screens/key_import_screen.dart';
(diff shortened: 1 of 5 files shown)


rabble (commented):
I found no code-level issue in the current diff, but this head is not ready to merge: `Tests (shard 0/4)` and the aggregate `Mobile CI` check are failing. The other listed checks are green. Please inspect or rerun the failed shard and request review again once CI is green.

rabble (commented):
Feedback-only review (agent-assisted, no branch changes). **Requesting changes — CI is red, and the failure is caused by this diff.**

## The failing test is this PR's

`Tests (shard 0/4)` fails on `errorBuilder shows RouteErrorScreen for unknown location`. I pulled the stack:

```
#4  goRouterProvider.<anonymous closure> (package:openvine/router/app_router.dart:81:33)
#3  Ref.read (package:riverpod/src/core/ref.dart:609:30)
...
#13 _pumpRouter (mobile/test/router/route_error_routing_test.dart:78:33)
```

`app_router.dart:81` is the `ref.read(hasFollowingInCacheProvider)` line this PR adds. That test's `ProviderContainer` has no override for it, so building `goRouterProvider` now throws a `ProviderException`.

## Why I'd not just add the override

The test is reporting something true: **prewarming turned `hasFollowingInCacheProvider` into a hard construction-time dependency of the router.** Every container that builds a router must now satisfy it. Silencing that with an override in one test keeps the coupling and loses the signal.

The coupling also has a product consequence. Previously the value was read lazily at redirect time, on first navigation. Now it is snapshotted at router construction — strictly earlier in cold start, so strictly more likely to be read before the contact cache has loaded. The `ref.listen` cannot rescue this: `_hasNavigated` latches on the first redirect, so if that redirect runs before the listener fires, a user who *does* follow people gets bounced to Explore, and the corrected value arrives too late to matter. That is the exact bug class the redirect exists to avoid, made more likely rather than less.

If the goal is "don't construct the family provider during redirect", reading it lazily inside the redirect as before — and keeping `ref.listen` purely for the refresh — gets that without moving the read earlier.

## Duplicated rule

`checkEmptyFollowingRedirectProvider` is still defined and still has its own test group, while production now goes through `emptyFollowingRedirectForLocation`. Two implementations of one routing rule, and only one of them ships. Per `.claude/rules/agent_workflow.md` §4 the dead path should go in this PR rather than be left for later. Same for the `hasFollowingInCache ?? ref.read(...)` fallback — the nullable parameter keeps both call paths alive; once the caller always passes it, make it required.

## What I do like

`RouterRefreshListenable` is a clean fix. Coalescing to one microtask, the `_disposed` guard inside the scheduled callback, and `dispose()` being idempotent are all correct, and the three tests around it are well-targeted — the disposal test in particular can genuinely fail.

Minor: the tests use `await Future<void>.delayed(Duration.zero)` (a timer) to drain what is a microtask. `await Future<void>.microtask(() {})` matches the mechanism and is less prone to hiding a change from `scheduleMicrotask` to something later.

NotThatKindOfDrLiz:
Takeover update:

- Removed the eager `hasFollowingInCacheProvider` prewarm/snapshot wiring from `goRouterProvider`, so router construction no longer requires a `SharedPreferences` override.
- Restored the existing lazy `checkEmptyFollowingRedirectProvider(location)` read inside the first-navigation redirect guard.
- Dropped the duplicated empty-following helper/tests and changed the new listenable tests to drain the scheduled microtask directly.
- Updated the PR body to match the narrowed scope.

Verification:

- `flutter test test/router/route_error_routing_test.dart test/router/empty_contacts_redirect_test.dart test/router/router_refresh_listenable_test.dart`
- `flutter test test/router/app_router_redirect_test.dart test/router/app_router_test.dart test/router/router_refresh_gate_coupling_test.dart`
- `flutter analyze lib/router test/router/router_refresh_listenable_test.dart test/router/empty_contacts_redirect_test.dart test/router/route_error_routing_test.dart test/router/app_router_redirect_test.dart test/router/app_router_test.dart test/router/router_refresh_gate_coupling_test.dart`
- Pre-push hook analyzer + changed-file test
- GitHub checks are green on `aac3f15c79e0392c782a415ed318acc6c292869b`, including `Tests (shard 0/4)` and aggregate `Mobile CI`.

hm21 (approved):
LGTM. Verified all three new tests fail against the pre-fix implementation, and the microtask escape is the right level: Riverpod's `visitChildren` iterates `dependents` live, so a synchronous refresh → redirect → `ref.read` re-entry is a credible source of the reported concurrent-modification. Deferring also makes the autofill race in `_CommitAutofillOnAuthenticated` strictly safer, not worse.

PR #6704, opened 2026-08-04
fix(upload): recover stale resumable auth sessions

### LINKED ISSUE #6119: fix(upload): split upload Crashlytics groups and recover stale resumable auth
## Crashlytics

- Issue: `3cf70c1dba6a1f1d5385f574a3df7beb`
- Type: non-fatal · 72 impacted users / 337 events
- Versions: first 0.0.1 -> last 1.0.16

## Current assessment

The original Cloudflare Stream framing is obsolete. The direct Cloudflare upload path is no longer present in the app; uploads now go through Blossom. The Crashlytics group is still useful, but it is not reliable evidence of one Cloudflare 401 bug.

The group was polluted because upload failures were reported with a stack trace captured inside the failure reporter rather than at the actual throw site. Expected upload failures were also sent to Crashlytics unconditionally, so offline/network/auth/session failures collapsed into one noisy non-fatal group.

There is also a current Blossom resumable-upload risk: 401/403 responses while resuming or uploading chunks can come from a stale persisted session auth header. Those sessions should be treated as dead, cleared, and reinitialized rather than retried forever with the same stale session metadata.

## Fix direction

- Forward the real upload failure stack trace into Crashlytics reporting.
- Keep upload failure breadcrumbs/custom keys, but call `recordError` only for reportable upload categories.
- Treat resumable session 401/403 like unrecoverable session/auth failures for retry and session-clearing purposes.
- Keep user/error categorization as `AUTHENTICATION` for resumable 401/403.
- Skip the resume `HEAD` when a persisted resumable session has an expired `expiresAt`; initialize a fresh session instead.

## Verification

Covered by upload reporter, retry/session predicate, UploadManager, and Blossom resumable upload tests.

_Rescoped during implementation; original Crashlytics title was `Cloudflare Stream upload failed: HTTP 401`._

NotThatKindOfDrLiz:
## Summary

- forward the real upload failure stack trace into upload Crashlytics reporting
- keep upload failure breadcrumbs/custom keys, but skip `recordError` for expected upload categories
- treat Blossom resumable 401/403 responses as dead persisted sessions while preserving `AUTHENTICATION` categorization
- skip stale-session `HEAD` probes when a persisted resumable session is already past `expiresAt`

Closes #6119

## Verification

- `flutter test test/services/upload/upload_progress_reporter_test.dart test/services/upload/upload_retry_policy_test.dart test/services/upload/upload_session_errors_test.dart`
- `flutter test test/services/upload_manager_error_handling_test.dart test/services/upload_manager_resumable_upload_test.dart`
- `flutter test test/src/blossom_upload_service_test.dart` from `mobile/packages/blossom_upload_service`
- `flutter analyze`

### DIFF
--- a/mobile/lib/services/upload/upload_progress_reporter.dart
+++ b/mobile/lib/services/upload/upload_progress_reporter.dart
@@ -155,6 +155,11 @@ class UploadProgressReporter {
 
   /// Categorise [error] into a string category constant for monitoring.
   Future<String> categorizeError(dynamic error) async {
+    if (error is BlossomResumableUploadException &&
+        (error.statusCode == 401 || error.statusCode == 403)) {
+      return 'AUTHENTICATION';
+    }
+
     if (isExpiredResumableSessionError(error)) {
       return 'UPLOAD_SESSION_EXPIRED';
     }
@@ -163,6 +168,33 @@ class UploadProgressReporter {
 
     if (connectivity == ConnectivityResult.none) {
       return 'NO_INTERNET';
+    }
+
+    final resumableStatusCode = error is BlossomResumableUploadException
+        ? error.statusCode
+        : null;
+    final resumableFailureReason = error is BlossomResumableUploadException
+        ? error.failureReason
+        : null;
+
+    if (resumableFailureReason == BlossomUploadFailureReason.authUnavailable) {
+      return 'NETWORK_ERROR';
+    }
+
+    if (resumableStatusCode != null) {
+      if (resumableStatusCode == 408) return 'TIMEOUT';
+      if (resumableStatusCode == 413) return 'FILE_TOO_LARGE';
+      if (resumableStatusCode == 429) return 'RATE_LIMITED';
+      if (resumableStatusCode == 401 || resumableStatusCode == 403) {
+        return 'AUTHENTICATION';
+      }
+      if (resumableStatusCode == 502 ||
+          resumableStatusCode == 503 ||
+          resumableStatusCode == 504) {
+        return 'SERVER_UNAVAILABLE';
+      }
+      if (resumableStatusCode >= 500) return 'SERVER_ERROR';
+      if (resumableStatusCode >= 400) return 'CLIENT_ERROR';
     }
 
     if (error is BlossomUploadFailureException) {
@@ -367,6 +399,7 @@ class UploadProgressReporter {
     String errorCategory,
     UploadMetrics? metrics,
     ConnectivityResult connectivity, {
+    required StackTrace stackTrace,
     required bool isManagerInitialized,
   }) async {
     try {
@@ -421,8 +454,6 @@ class UploadProgressReporter {
         );
       }
 
-      final stackTrace = StackTrace.current;
-
       final fileExists = kIsWeb
           ? 'N/A (web)'
           : '${File(upload.localVideoPath).existsSync()}';
@@ -445,6 +476,16 @@ ${metrics != null ? '- File Size: ${metrics.fileSizeMB} MB\n- Duration: ${metric
 
       crashReporting.log('UPLOAD_FAILURE: $detailedError');
 
+      if (!_isReportableUploadFailure(errorCategory)) {
+        Log.info(
+          'Skipping Crashlytics recordError for expected upload failure: '
+          '$errorCategory',
+          name: 'UploadManager',
+          category: LogCategory.video,
+        );
+        return;
+      }
+
       await crashReporting.recordError(
         error,
         stackTrace,
@@ -463,6 +504,13 @@ ${metrics != null ? '- File Size: ${metrics.fileSizeMB} MB\n- Duration: ${metric
         category: LogCategory.video,
       );
     }
+  }
+
+  bool _isReportableUploadFailure(String errorCategory) {
+    return switch (errorCategory) {
+      'OUT_OF_MEMORY' || 'UNKNOWN' => true,
+      _ => false,
+    };
   }
 
   /// Send an initialization-failure report to Crashlytics.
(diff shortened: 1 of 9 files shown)


rabble (approved):

rabble (commented):

Classifying 401/403 as auth rather than generic failure is right, and re-initialising a resumable session whose `expiresAt` has passed — instead of HEAD-ing a session the server already dropped — is a good catch. `expired resumable session re-inits without session HEAD` asserting `sessionHeadCalls == 0` is the right way to pin that.

Four things worth a look.

## 1. `upload_session_errors.dart` now contradicts its own invariant

That file's doc comment is explicit:

> This predicate decides both whether the error is retriable *and* whether the resumable session is nulled on failure, so the two used to drift between callers… lives here as a single source of truth to keep those callers in lockstep.

After this PR, `isExpiredResumableSessionError` returns `true` for 401/403, while `UploadRetryPolicy.isRetriableError` returns `false` for the same errors. The two answers have drifted again — which is the precise failure this file was created to prevent.

The *behaviour* may well be what you want (null the session, don't retry). But then the predicate is no longer "is retriable AND null the session", it's "should the session be nulled", and the doc needs to say so — otherwise the next person restores lockstep and silently re-enables retry on auth failures.

## 2. Dead branch in `categorizeError`

The early return at the top of the method:

```dart
if (error is BlossomResumableUploadException &&
    (error.statusCode == 401 || error.statusCode == 403)) {
  return 'AUTHENTICATION';
}
```

makes the later branch unreachable:

```dart
if (resumableStatusCode == 401 || resumableStatusCode == 403) {
  return 'AUTHENTICATION';
}
```

Worth deleting one. Related: because the early return precedes `isExpiredResumableSessionError`, a 401/403 can no longer produce `UPLOAD_SESSION_EXPIRED` — deliberate, I think, but not obvious from reading top to bottom.

## 3. A new test can false-pass

```dart
expect(
  () => service.resumeUploadSession(...),
  throwsA(isA<BlossomResumableUploadException>()...),
);
```

`resumeUploadSession` is async, so `throwsA` returns a `Future` that nothing awaits. If the matcher fails, the test may already have completed green. The neighbouring `resumeUploadSession throws on unexpected status` looks like it has the same shape. `await expectLater(...)` fixes both.

## 4. Reportable allowlist is narrow

`_isReportableUploadFailure` admits only `OUT_OF_MEMORY` and `UNKNOWN`. Consistent with the matrix in `.claude/rules/error_handling.md` for network/API/auth — but `CLIENT_ERROR` (an unexpected 4xx that isn't 401/403/408/413/429) is closer to "our request was malformed" than to "expected remote failure", and that's a bug signal you'd now never see. Worth considering including it.

Also: `isRetriableError` checks `error.isRecoverableResumableFailure` *before* the `reason == auth` early-out. If that getter is true for any 401/403 shape, auth errors would be retried despite the explicit rule below. Worth confirming the ordering is intentional.

rabble (changes requested):
Request changes based on the late review.

`mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart` now turns resumable 401/403/404/410 responses into `BlossomResumableUploadException`, but `_uploadWithSingleLegacyFallback` catches that failure and falls through to the legacy upload for Divine-owned hosts. If the legacy fallback also fails, `UploadManager` receives a generic failure and does not clear the persisted resumable session. The next attempt can reuse the same stale session and repeat the HEAD/chunk failure and fallback path.

Please preserve an explicit stale-session marker through the fallback boundary, or clear the persisted session before/when falling back, and add an end-to-end regression test for resumable 401/403 followed by legacy fallback failure.

rabble (commented):
**Correction — two of my four points were wrong. I verified them on this branch instead of reasoning, and they don't hold. Please disregard points 1 and 3.**

### Point 3 (unawaited `throwsA` can false-pass) — wrong

I changed the matcher to an impossible value (`'statusCode', 999`) and re-ran. The test fails correctly:

```
Expected: <BlossomResumableUploadException> with `statusCode` 999
  Actual: ... which has `statusCode` with value <401>
```

`flutter_test`'s `expect` registers the returned Future as an outstanding callback, so an async `throwsA` mismatch is caught. `expect(() => asyncFn(), throwsA(...))` is safe here. `await expectLater` is stylistically nicer but changes nothing.

### Point 1 (`isExpiredResumableSessionError` now contradicts the retry policy) — wrong

I ran both predicates across the relevant codes:

```
CODE 401 -> expired=true retriable=false
CODE 403 -> expired=true retriable=false
CODE 404 -> expired=true retriable=false      <- pre-existing
CODE 410 -> expired=true retriable=false      <- pre-existing
```

`expired=true, retriable=false` is the **pre-existing, consistent** contract — an expired session is not retriable, which is the point. 401/403 were made to behave exactly like 404/410. There is no drift; this PR makes the four codes uniform. I misread the file's doc comment ("decides both whether the error is retriable *and* whether the session is nulled") as asserting the two answers agree, when it means one predicate feeds both decisions.

### What still stands

**Point 2 — the duplicate 401/403 branch in `categorizeError`.** The early return at the top makes the later `resumableStatusCode == 401 || == 403` check unreachable. Readability only, no behavioural impact.

**Point 4 — the reportable allowlist.** `_isReportableUploadFailure` admits only `OUT_OF_MEMORY` and `UNKNOWN`; whether `CLIENT_ERROR` (an unexpected 4xx that isn't 401/403/408/413/429) belongs there is a judgement call worth making deliberately.

Also unverified and therefore just a question, not a finding: `isRetriableError` checks `error.isRecoverableResumableFailure` before the `reason == auth` early-out — worth confirming that ordering is intended.

Sorry for the noise on 1 and 3.

PR #6705, opened 2026-08-04
fix(bluesky): complete provisioning lifecycle

### LINKED ISSUE #6095: feat(bluesky): finish and ship ATProto crossposting — backend is live, mobile is the last blocker

### LINKED ISSUE #6357: fix(bluesky): username setup is requested twice

NotThatKindOfDrLiz:
## Summary

- Move Bluesky crosspost status composition into a repository and parse keycast provisioning state at the API boundary.
- Poll pending ATProto provisioning until a terminal state, stop polling on cubit close, cap polling when provisioning stays pending, and add retry handling for failed provisioning.
- Treat a keycast-returned username as claimed when composing status with divine-name-server, preventing the contradictory handle-plus-setup state.
- Surface ready-state DID, localized provisioning status, timeout copy, and a localized backfill disclosure in the Bluesky settings screen.
- Add repository, cubit, API client, widget, provider, and l10n consistency coverage.

## Notes

- Refs #6095 and #6357. This does not close #6095 because profile-editor status, per-video status backend work, and flag rollout remain separate items. This also does not close #6357 until staging/device verification confirms the app-claimed username path no longer routes back to setup.
- Keycast `error` remains repository diagnostic data and is reported through `addError`; the settings UI renders localized status text instead of raw server copy.
- The mobile name-server fallback is preserved in the repository until staging confirms keycast reconciliation fully covers the app-claimed username path.

## Verification

- `flutter test test/repositories/bluesky_crosspost_repository_test.dart test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart test/screens/settings/bluesky_settings_screen_test.dart`
- Pre-push hook: merge-conflict check, analyzer, generated-file verification, affected tests including l10n consistency, crosspost API client, upload media providers, repository, cubit, and settings screen tests


rabble (changes requested):
Request changes.

1. The retry button in `mobile/lib/screens/settings/bluesky_settings_screen.dart` is enabled whenever `provisioningState == failed`, but `retryProvisioning()` leaves that state unchanged while `toggleCrosspost()` is in flight. Multiple taps can therefore issue duplicate provisioning requests. Disable the action while toggling or serialize retries, with a double-tap test.

2. `_loadStatus()` only serializes polling loads; it does not invalidate an in-flight poll when `toggleCrosspost()` completes. A stale poll response can overwrite the toggle result and restart/stop the provisioning poller from obsolete state. Add generation/operation invalidation or serialize status reads with writes.

3. Poll failures are emitted as generic user-visible failures while the periodic timer continues. A transient outage can repeatedly increment attempts and surface repeated error UI. Preserve the pending/last-good state and retry quietly, or stop/back off the poller after a failure.

rabble (commented):
Feedback-only review (agent-assisted, no branch changes). Late addition — this one slipped my first pass over the ready queue, so it's arriving after the others.

The structural direction is right, and three things in particular are done well:

- **`AtprotoProvisioningState` replacing `provisioningState == 'ready'`.** String-compare against a server enum is exactly the kind of thing that silently breaks when the backend renames a state; parsing at the API boundary and defaulting to `notLinked` is the correct fix. Making the field non-nullable with a default removes a whole class of null-handling from the UI.
- **Pulling status composition into `BlueskyCrosspostRepository`.** The cubit was previously composing keycast status with a name-server username lookup — two sources, one fallback rule — which `.claude/rules/architecture.md` puts squarely in the repository. Good move, and it makes the cubit readable.
- **Poll lifecycle is actually tested.** `polls pending status until a terminal state is loaded` and `stops polling when the cubit closes` are the two tests that matter here, and both can fail. The injectable `provisioningPollInterval` is what makes them possible without `Future.delayed` in app code. The `isClosed` guards after each await are right.

l10n mirrored across all 17 locales with the generated output committed — no ARB-consistency debt.

Four things worth resolving:

## 1. `provisioningError` is a server string in cubit state, rendered raw

`.claude/rules/state_management.md` states it as a hard rule: state must never contain error messages or error strings; use status enums and let the UI map to `context.l10n`. This PR stores keycast's `error` field in state and renders it directly as secondary text.

The description argues it's "server-supplied domain display data" rather than an error message, and that framing is defensible — but the consequences are the ones the rule exists to prevent:

- It bypasses l10n. `'PDS quota exhausted'` (the test fixture) reaches a Turkish or Japanese user in English.
- It renders untrusted server text into the UI verbatim. Worth confirming it goes through the display-text sanitiser used elsewhere for remote strings — a lone UTF-16 surrogate in that field is a paragraph-layout crash, not a cosmetic bug.
- It couples mobile UI to keycast's error copy, so a backend wording change is a mobile-visible change nobody reviews.

If the detail is genuinely needed for support triage, one option is keeping it out of the rendered tree and into the log/bug-report capture, with the UI showing only the localized headline. If it must be visible, an explicit note in the state class documenting the deliberate exception — and sanitising at the boundary — would at least stop the next reader from treating it as precedent.

## 2. Polling is unbounded

`Timer.periodic(_provisioningPollInterval)` runs for as long as the state is `pending` and the cubit is alive — every 5s, indefinitely. Close stops it, so the leak is bounded by screen lifetime, but a settings screen left open against a provisioning job that's genuinely stuck (or a keycast outage returning `pending`) polls forever. Worth a max-attempt cap or a backoff, with a terminal "still working, check back later" state — which also gives the user better information than a spinner that never resolves.

## 3. `_loadingStatus` is mutable cubit state outside the state object

`.claude/rules/state_management.md` → *No Mutable Instance Variables in BLoC*. `_provisioningPollTimer` I'd argue is fine — it's a resource handle, same category as a stream subscription. `_loadingStatus` is different: it's re-entrancy state that changes what `loadStatus()` does, and it's invisible to tests and the UI. Since the poller and the user can both trigger a load, that guard is load-bearing; a field on the state (or a `droppable`-style guard) would make it observable.

## 4. `copyWith` can't clear the new nullable fields

`did: did ?? this.did` and `provisioningError: provisioningError ?? this.provisioningError` mean neither can be reset to null. `_emitLoaded` constructs a fresh state so the main path is safe, but the failure path in `_loadStatus` uses `copyWith` and will carry a stale `provisioningError` from a previous attempt into a state whose current failure is something else. Same shape as the existing `clearError` flag — worth `clearProvisioningError` / `clearDid`, or documenting why it can't happen.

## Minor

`retryProvisioning() => toggleCrosspost(enabled: true)` — if provisioning failed while `enabled` was already true, this calls the toggle with the value it already has. Worth confirming the repository treats that as "re-provision" rather than a no-op, since the button's whole purpose is the retry.

NotThatKindOfDrLiz:
Takeover update pushed in f109b0d15.

What changed:
- Added operation-generation invalidation across user loads, polling loads, and toggle/retry writes so stale poll responses cannot overwrite newer toggle results.
- Made provisioning polls keycast-only, quiet on transient failure, and last-good-state preserving while pending.
- Guarded post-await toggle emissions after cubit close and serialized duplicate toggles/retries.
- Restored name-server claim carry-through when keycast status fails.
- Disabled failed-provisioning Retry while retry is in flight and keyed the BlocProvider by repository identity.
- Removed the unused contradictory isTerminal getter.
- Localized the Bluesky backfill disclosure in non-English ARBs, regenerated l10n, and added a guard against copied-English backfill copy.
- Restored/added regression coverage for toggling emissions, usernameNotSynced, unknown-claim enable, acknowledgeError, stale polls, quiet poll failures, close-after-toggle, retry double-tap, and deterministic fakeAsync polling.

Verification:
- flutter test test/services/crosspost_api_client_test.dart test/repositories/bluesky_crosspost_repository_test.dart test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart test/screens/settings/bluesky_settings_screen_test.dart test/l10n/arb_consistency_test.dart
- Pre-push hook: merge-conflict check, analyzer, generated-file verification, affected tests

NotThatKindOfDrLiz:
Follow-up pushed in b6a160c2d.

What changed:
- Closed the residual poll-during-toggle race by blocking poll starts while the cubit is toggling, with regression coverage.
- Moved user/poll load guards and operation generation into CrosspostSettingsState; the Timer remains a private resource handle.
- Added a max provisioning poll attempt cap. Pending provisioning now stops active polling after the cap and shows the localized "still working" state instead of spinning forever.
- Removed raw keycast provisioningError from Cubit/UI state and stopped rendering it directly. Failed provisioning still shows localized failed status and retry; raw detail remains repository/client diagnostic data and is reported through the Cubit error stream.
- Added clearDid support to copyWith and regression coverage.
- Treat keycast-provided usernames as claimed even when divine-name-server lookup is stale/unknown, with repository, cubit, and widget coverage.
- Corrected issue #6095's stale backend-contract text for {enabled: bool}, 200 OK, and async provisioning.

Verification:
- flutter test test/services/crosspost_api_client_test.dart test/repositories/bluesky_crosspost_repository_test.dart test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart test/screens/settings/bluesky_settings_screen_test.dart test/l10n/arb_consistency_test.dart
- Pre-push hook: merge-conflict check, analyzer, generated-file verification, affected tests

rabble (approved):

All three blockers are resolved, each with a test that can fail:

1. **Duplicate provisioning requests.** `toggleCrosspost` now returns early on `state.status == toggling`, and since `retryProvisioning()` delegates to it, repeated retry taps collapse to one request.
2. **Stale poll overwriting a toggle.** The generation counter does the job — `_beginPollLoad` captures the current generation without bumping it, `toggleCrosspost` and `_beginUserLoad` bump it, and both the success and failure paths re-check before emitting. `poll that fires during toggle does not start` covers the ordering.
3. **Repeated error UI from transient poll failures.** `_emitQuietPollFailure` preserves the last-good state and keeps the attempt count instead of surfacing a generic failure, and the poller now stops after `maxProvisioningPollAttempts` (24 × 5s ≈ 2 min) with `crosspostStillWorking` shown instead of a dead spinner. `stops polling after the max pending attempts is reached` pins it.

Beyond what I asked for: `provisioningError` is gone from cubit state entirely, so the keycast string no longer bypasses l10n or reach the UI verbatim. It goes to `addError` as a `CrosspostProvisioningException` — correctly *not* wrapped in `Reportable`, since a server-supplied domain error is a NO in the `error_handling.md` matrix, so it lands in the log without flooding Crashlytics. That's the right resolution and better than the compromise I floated.

Two things for later, neither blocking:

- `operationGeneration` is in the constructor and `copyWith` but not in Equatable `props`. Every current bump rides along with a props change (`userLoadInFlight` flipping, or `status` → `toggling`), so no emit is dropped today. But a future edit that bumps the generation *alone* would be silently swallowed by Bloc's equality check and the guard would quietly stop working. A comment on the field saying it's intentionally excluded and why would keep the next person from finding that the hard way.
- I asked for a double-tap test specifically and there isn't one. The `toggling` guard is plain enough to read, and the poll-during-toggle test covers the adjacent race, so I'm not going to hold the PR for it.

Ship it.

PR #6706, opened 2026-08-04
fix(avatar): validate remote svg payloads

### LINKED ISSUE #6118: fix(xml): XmlParserException '>' expected on malformed markup (Crashlytics b97ccc46)
## Crashlytics
- Issue: `b97ccc46474b922cbaa19b1ab1397924` ([console](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/b97ccc46474b922cbaa19b1ab1397924))
- Type: non-fatal · **72 impacted users / 1004 events**
- Versions: first 1.0.14 → last **1.0.16**

## Signature
`package:xml/src/xml_events/iterator.dart - XmlEventIterator.moveNext`
`XmlParserException: ">" expected at 2:175` (thrown, runZonedGuarded).

## Root cause (hypothesis)
An XML parse over malformed markup. Most likely `flutter_svg` parsing a broken/truncated SVG asset, or an oEmbed/HTML fragment being fed to the xml parser. "`>` expected at 2:175" = an unterminated tag in the input. Need a sample input — pull events in the console to see which asset/URL is being parsed.

## Suggested next step
Identify the parse call site (SVG render vs oEmbed/metadata). Wrap in a typed catch that logs the offending source and renders a fallback instead of throwing. May share a root with the missing-SVG issue (`20e5b4c3`).

NotThatKindOfDrLiz:
## Summary
- add a validating AvatarSvgRepository with timeout, byte cap, XML/SVG validation, and bounded positive/negative cache
- route remote SVG avatars through AvatarSvgCubit and render only validated bytes with SvgPicture.memory
- keep placeholder fallback behavior for rejected SVG payloads and existing raster failure cache behavior

## Verification
- flutter analyze
- flutter test test/repositories/avatar_svg_repository_test.dart test/blocs/avatar_svg/avatar_svg_cubit_test.dart test/widgets/user_avatar_svg_test.dart
- flutter test test/widgets/comprehensive_user_avatar_test.dart

Closes #6118

### DIFF
--- a/mobile/lib/blocs/avatar_svg/avatar_svg_cubit.dart
+++ b/mobile/lib/blocs/avatar_svg/avatar_svg_cubit.dart
@@ -20,17 +20,23 @@ class AvatarSvgCubit extends Cubit<AvatarSvgState> {
   final String _url;
 
   Future<void> load() async {
-    if (state.status == AvatarSvgStatus.loading) return;
+    if (isClosed || state.status == AvatarSvgStatus.loading) return;
 
-    emit(state.copyWith(status: AvatarSvgStatus.loading));
+    _emitIfOpen(state.copyWith(status: AvatarSvgStatus.loading));
     try {
       final bytes = await _repository.load(_url);
+      if (isClosed) return;
+
       if (bytes == null) {
-        emit(state.copyWith(status: AvatarSvgStatus.unavailable));
+        _emitIfOpen(
+          state.copyWith(status: AvatarSvgStatus.unavailable, bytes: null),
+        );
         return;
       }
-      emit(state.copyWith(status: AvatarSvgStatus.ready, bytes: bytes));
+      _emitIfOpen(state.copyWith(status: AvatarSvgStatus.ready, bytes: bytes));
     } on Object catch (error, stackTrace) {
+      if (isClosed) return;
+
       Log.error(
         'Avatar SVG load failed',
         name: 'AvatarSvgCubit',
@@ -39,7 +45,13 @@ class AvatarSvgCubit extends Cubit<AvatarSvgState> {
         stackTrace: stackTrace,
       );
       addError(error, stackTrace);
-      emit(state.copyWith(status: AvatarSvgStatus.unavailable));
+      _emitIfOpen(
+        state.copyWith(status: AvatarSvgStatus.unavailable, bytes: null),
+      );
     }
+  }
+
+  void _emitIfOpen(AvatarSvgState nextState) {
+    if (!isClosed) emit(nextState);
   }
 }
--- a/mobile/lib/blocs/avatar_svg/avatar_svg_cubit.dart
+++ b/mobile/lib/blocs/avatar_svg/avatar_svg_cubit.dart
@@ -1,0 +1,45 @@
+// ABOUTME: Cubit for loading validated remote SVG avatar bytes.
+// ABOUTME: Keeps UserAvatar rendering state separate from repository/network work.
+
+import 'dart:typed_data';
+
+import 'package:equatable/equatable.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:openvine/repositories/avatar_svg_repository.dart';
+import 'package:unified_logger/unified_logger.dart';
+
+part 'avatar_svg_state.dart';
+
+class AvatarSvgCubit extends Cubit<AvatarSvgState> {
+  AvatarSvgCubit({required AvatarSvgRepository repository, required String url})
+    : _repository = repository,
+      _url = url,
+      super(const AvatarSvgState());
+
+  final AvatarSvgRepository _repository;
+  final String _url;
+
+  Future<void> load() async {
+    if (state.status == AvatarSvgStatus.loading) return;
+
+    emit(state.copyWith(status: AvatarSvgStatus.loading));
+    try {
+      final bytes = await _repository.load(_url);
+      if (bytes == null) {
+        emit(state.copyWith(status: AvatarSvgStatus.unavailable));
+        return;
+      }
+      emit(state.copyWith(status: AvatarSvgStatus.ready, bytes: bytes));
+    } on Object catch (error, stackTrace) {
+      Log.error(
+        'Avatar SVG load failed',
+        name: 'AvatarSvgCubit',
+        category: LogCategory.ui,
+        error: error,
+        stackTrace: stackTrace,
+      );
+      addError(error, stackTrace);
+      emit(state.copyWith(status: AvatarSvgStatus.unavailable));
+    }
+  }
+}
--- a/mobile/lib/blocs/avatar_svg/avatar_svg_state.dart
+++ b/mobile/lib/blocs/avatar_svg/avatar_svg_state.dart
@@ -2,16 +2,23 @@ part of 'avatar_svg_cubit.dart';
 
 enum AvatarSvgStatus { initial, loading, ready, unavailable }
 
+const _avatarSvgBytesUnset = Object();
+
 class AvatarSvgState extends Equatable {
   const AvatarSvgState({this.status = AvatarSvgStatus.initial, this.bytes});
 
   final AvatarSvgStatus status;
   final Uint8List? bytes;
 
-  AvatarSvgState copyWith({AvatarSvgStatus? status, Uint8List? bytes}) {
+  AvatarSvgState copyWith({
+    AvatarSvgStatus? status,
+    Object? bytes = _avatarSvgBytesUnset,
+  }) {
     return AvatarSvgState(
       status: status ?? this.status,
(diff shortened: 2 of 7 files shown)


rabble (changes requested):
Request changes.

`mobile/lib/repositories/avatar_svg_repository.dart` applies the 5-second timeout to `_client.send(request)`, but `_readBounded(response.stream)` then awaits the response body with no deadline. A server can send response headers and keep the stream open indefinitely, leaving avatar loading/cache work hung despite the request timeout. Apply a deadline/cancellation policy to the body read and add a test for an incomplete/slow body stream.

rabble (commented):

Validating SVG bytes before `flutter_svg` sees them is the right fix, and doing it in a repository behind an interface — rather than patching the widget — is the right layer. The rejection reasons are well chosen (status, size cap, content-type-or-sniff, XML well-formedness), the `_logRejected` calls make failures diagnosable, and `rejected .svg URL does not raise a zone error` tests the actual reported symptom rather than a proxy. Replacing ~110 lines of hand-rolled `HttpClient`/`HttpClientResponse` fakes with `MockClient` is a solid cleanup on its own.

The `BlocProvider` carries `key: ValueKey((imageUrl, repository))`, which is the pattern `.claude/rules/state_management.md` asks for. Good.

Three things:

## 1. Production uses a hidden singleton

```dart
final AvatarSvgRepository defaultAvatarSvgRepository = HttpAvatarSvgRepository();
```

with injection available only via an `@visibleForTesting` parameter. So the shipped path always uses a top-level global holding a mutable `_cache` and an `http.Client` that is never closed. That's the "hidden singleton-style dependency" that `.claude/rules/architecture.md` asks us to avoid in favour of constructor injection.

More concretely, `.claude/rules/testing.md` → *VGV merged isolate & process-global state* applies: under `very_good test --optimization` every untagged suite shares one isolate, so this cache persists across test files. A test that loads an avatar URL leaves a cached entry (positive for 1h, negative for 10m) visible to every later suite, and `clearCache()` is `@visibleForTesting` but nothing calls it in teardown. That's a seed-dependent flake waiting to happen.

A Riverpod provider for the repository, overridden in tests, would fix both — and matches how the rest of the app gets its dependencies.

## 2. `copyWith` can't clear `bytes`

```dart
bytes: bytes ?? this.bytes,
```

so a transition from `ready` back to `unavailable` retains stale bytes, leaving a state that says "unavailable" while holding a payload. Harmless today because the builder gates on `status != ready` first, but it's the kind of latent inconsistency that bites when someone later reads `state.bytes` without checking status.

## 3. Losing `SvgPicture.network`'s caching

Each `UserAvatar` now mounts its own cubit and calls `SvgPicture.memory`, which re-parses the SVG per widget — `flutter_svg`'s picture cache keyed on the network URL no longer applies. The repository caches *bytes*, not the parsed picture. Probably fine at avatar sizes with a 128-entry cap, but worth a glance if SVG avatars ever show up densely in a feed.

## Minor

- `_startsLikeSvg` calls `utf8.decode(...)` without `allowMalformed`, so a non-UTF8 payload throws and is caught by the outer `on Object catch` — right outcome, noisy route. `allowMalformed: true` on a 16-byte sniff would be clearer about intent.
- Unrelated reformatting of `caches a broken raster URL through the real image provider` adds diff noise to an otherwise focused change.

NotThatKindOfDrLiz:
Takeover update pushed in eaaf1bcb9.

What changed:
- guarded AvatarSvgCubit emissions after async loads so dispose-during-load no longer reports/throws after close
- added a body-stream timeout for SVG responses and cancellation/error handling for non-200 response bodies
- added in-flight request de-dupe for identical SVG avatar URLs
- fixed AvatarSvgState.copyWith so unavailable states can clear stale bytes
- made SVG sniffing reject binary/non-UTF8 payloads through the structured non-svg-payload path
- removed the unused clearCache() test-only method and expanded cache coverage for positive TTL, negative TTL expiry, and LRU eviction

Verification:
- flutter test test/blocs/avatar_svg/avatar_svg_cubit_test.dart test/repositories/avatar_svg_repository_test.dart test/widgets/user_avatar_svg_test.dart
- flutter test test/widgets/comprehensive_user_avatar_test.dart
- flutter analyze
- pre-push hook: no merge conflicts with main, analyze, changed-file tests
- GitHub checks are green: mergeability, semantic PR, generated files, format, analyze, all test shards, preview build, aggregate Mobile CI

Notes on the latest feedback-only review: copyWith clearing and malformed/binary sniffing are fixed in this push. I left the production repository dependency shape unchanged to keep this security fix focused; the repository cache remains bounded, and in-flight/TTL/LRU behavior is now covered by tests.

rabble (commented):
**Follow-up: I verified the process-global cache concern on this branch. It reproduces, and it's a bit worse than I described.**

Two tests in one file, both hitting `defaultAvatarSvgRepository`, with a control URL to rule out fast-failure timing:

```
TEST 1  cachedUrl (cold)  = 12012us
TEST 2  cachedUrl         =   161us   <- populated by TEST 1
TEST 2  freshUrl          =  1030us   <- control, 6.4x slower
```

The cache survives the test boundary. Under `very_good test --optimization` every untagged suite shares one isolate, so an entry written by one suite is visible to every later one — positive entries for 1h, negative for 10m, which comfortably outlives any run. `clearCache()` exists and is `@visibleForTesting`, but **nothing calls it anywhere in `lib` or `test`**, so there's no teardown that resets it.

The part I'd missed: the global is already reachable from an existing widget test. `test/screens/inbox/widgets/conversation_tile_test.dart` pumps a `UserAvatar` with `picture: 'https://divine.video/divine-logo.svg'` and no `avatarSvgRepository`, so it takes the `defaultAvatarSvgRepository` path — a real `http.Client`, attempting real outbound HTTP, from a widget test. Today it fails fast and negative-caches, so it's invisible; it becomes order-dependent the moment any suite caches a result for a URL another suite also uses.

Making the repository a Riverpod provider overridden in tests fixes the ordering hazard and the stray HTTP client together.

My other two points (`copyWith` can't clear `bytes`; loss of `SvgPicture.network`'s picture cache) were from reading the diff, not from running anything — weigh them accordingly.

rabble (approved):

`eaaf1bc` resolves the blocker. `_readBounded` now reads through `_withBodyTimeout`, and non-200 responses go through `_cancelResponseBody` instead of a fire-and-forget `drain`. `times out incomplete response bodies` opens a stream that never emits and asserts both that `load` completes null and that the body subscription was cancelled — it fails if either half regresses, which is what I wanted.

Worth naming precisely, since it isn't quite what I asked for and that's fine: `Stream.timeout` is an *inactivity* timeout between events, not a total deadline. A server dripping bytes just under the interval keeps the read alive. The 256 KB `maxBytes` cap is what actually bounds the worst case, and for an avatar fetch that runs off the UI path behind per-URL in-flight dedup, that's a reasonable place to stop. Not asking for a wall-clock deadline on top.

The rest of the follow-up is good: the `_avatarSvgBytesUnset` sentinel makes `copyWith` able to clear `bytes`, so `unavailable` can no longer carry a stale payload; `_inFlight` collapses concurrent loads of the same URL; `allowMalformed: true` on the sniff stops a truncated multi-byte sequence from throwing where the whole point is to reject bad input. LRU eviction and both TTL paths now have tests.

**One open item, and it's the reason I'd want a follow-up rather than a rewrite here.** `defaultAvatarSvgRepository` is still a top-level global holding a mutable cache and an `http.Client` that is never closed, and this revision *removed* `clearCache()` — so there is now no reset hook at all. I verified the ordering hazard on the branch: entries survive the test boundary, and under `very_good test --optimization` every untagged suite shares one isolate, so a positive entry lives an hour and a negative one ten minutes across the whole run.

The concrete exposure is `test/screens/inbox/widgets/conversation_tile_test.dart:927`, which pumps a `UserAvatar` with `picture: 'https://divine.video/divine-logo.svg'` and no injected repository. To be fair to this PR: that test is pre-existing on `main` and untouched here, and it was already doing real outbound HTTP through `SvgPicture.network`. What changes is that the *result* now persists in a process-global for the rest of the run instead of failing transiently — which is what turns it from invisible into order-dependent.

That's a CI-flake hazard, not a user-facing bug, and the user-facing win here — SVG validated before `flutter_svg` sees it — is worth landing now. So: merge this, and either move the repository behind a Riverpod provider overridden in tests, or open an issue for it. Your call which, but per the no-tech-debt rule it shouldn't land untracked.

My earlier third point — losing `SvgPicture.network`'s picture cache, so each `UserAvatar` re-parses via `SvgPicture.memory` — still stands as written. It was read off the diff, not measured, so weigh it accordingly; the repository-level byte cache absorbs the network cost, and only the parse is repeated.

Ship it.

PR #6707, opened 2026-08-04
fix(deep-links): route claimed list links

### LINKED ISSUE #6700: fix(deep-links): route claimed list universal links
## Summary

Production AASA now claims `/list/*` for `divine.video` and `www.divine.video`, but the mobile deep-link parser does not handle that path. On iOS, an external `https://divine.video/list/<id>` tap can be handed to Divine and then dead-end as an unknown deep link.

## Evidence

- Production AASA at both hosts includes component `/list/*`.
- `mobile/lib/screens/curated_list_feed_screen.dart` exposes the in-app route as `/list/:listId`.
- `mobile/lib/services/deep_link_service.dart` has `DeepLinkType` values for video/profile/hashtag/search/invite/signerCallback/unknown, but no list type.
- `DeepLinkService.parseDeepLink()` falls through to unknown for `/list/<id>`.
- `mobile/lib/main.dart` handles `DeepLinkType.unknown` by logging and doing no navigation.

## Expected behavior

Externally tapped Divine list URLs claimed by AASA should navigate to the curated list feed route instead of dead-ending inside the app.

## Scope

- Add parsed list support to `DeepLinkService`.
- Map list deep links in `universal_link_resolver.dart` and the app-level deep-link handler.
- Reuse `CuratedListFeedScreen.pathForId()` for routing.
- Cover apex and `www` host cases in deep-link and resolver tests.

## Acceptance

- `https://divine.video/list/<id>` routes to `/list/<id>`.
- `https://www.divine.video/list/<id>` routes to `/list/<id>`.
- Malformed or empty list ids remain unknown / rejected without crashes.
- Existing video/profile/hashtag/search/invite universal-link behavior is unchanged.

## Related

Follow-up from #3843 after iOS universal-link handoff was restored and physically verified.

NotThatKindOfDrLiz:
## Description

Adds support for claimed Divine list universal links that use the local list route shape, `/list/:listId`, while preserving the existing web-canonical authored shape, `/list/:pubkey/:listId`.

This also routes Divine list URLs through `divineUrlToPushRoute` for in-app contexts like DM message bubbles, skips route normalization for full list universal links, and replaces the duplicated profile/hashtag/search/list deep-link nav-action helpers with one shared resolver while keeping the video-specific comments behavior separate.

**Related Issue:** Closes #6700

## Out of Scope

None.

## Verification

- `flutter test test/main_deep_link_nav_action_test.dart test/main_video_deep_link_nav_action_test.dart test/services/deep_link_service_test.dart test/router/universal_link_resolver_test.dart test/router/route_normalization_test.dart`
- `flutter analyze`
- pre-push hook analyzer and changed-file tests passed

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


rabble (approved):

rabble (commented):

Collapsing four near-identical enums and four near-identical resolvers into one parameterised pair is the right instinct — that was genuine duplication. Two consequences of *how* it was done are worth addressing.

## 1. The route-family rules are no longer covered by tests

Before, each family's rule lived inside a named `@visibleForTesting` function, and the tests called that function. The rule and its coverage were the same code.

Now the rules are inline closures at four `main.dart` call sites, and the tests **re-declare identical closures locally**:

```dart
isRouteFamilyLocation: (location) =>
    location == SearchResultsPage.emptyPath ||
    location.startsWith('${SearchResultsPage.pathPrefix}/') ||
    location.startsWith('${SearchResultsPage.emptyPath}?'),
```

That is a copy, not the shipped predicate. If someone edits the search-family rule in `main.dart`, every test still passes. The search rule is the one that most warrants coverage — three route shapes, one of them a `?focus=1` query-string case.

Suggest extracting the four predicates as named `@visibleForTesting` functions (`isProfileFamilyLocation`, `isHashtagFamilyLocation`, `isSearchFamilyLocation`, `isListFamilyLocation`) and passing those in at both the call sites and the tests. Keeps the dedup win, restores the coverage.

## 2. Net coverage loss

Four test files (~427 lines) deleted, one added (134). Cases that no longer exist anywhere:

- percent-encoded hashtag routes on both the `skip` and `go` paths (`pathForTag('çay keyfi')`)
- profile feed-mode routes via `pathForIndex` — both `skip` on the same feed route and `go` across profiles
- search: same query with `?focus=1` vs without → `go`, not `skip` (the subtlest case in the set)
- cross-type navigation `push` cases from video / profile / explore routes

With named predicates these fold naturally into one table-driven test per family rather than being dropped.

## 3. `/list/<hex-pubkey>` now resolves to a bogus local list

`DeepLinkService` gained a 2-segment `/list/{listId}` branch, and it matches anything. A truncated author link — `https://divine.video/list/a1b2…b2` with the list id missing — used to return `DeepLinkType.unknown`. It now returns `DeepLinkType.list` with `listId` set to the *pubkey*, and this PR pins that as intended:

```dart
test('treats a two-segment list URL as a local list id', () {
  const url = 'https://divine.video/list/$authorPubkey';
  expect(result.listId, equals(authorPubkey));
});
```

That navigates to a list that cannot exist, so the user gets an empty or error screen instead of the old clean rejection. Since the author segment already accepts hex / npub / nprofile, the 2-segment branch can cheaply exclude those shapes and fall through to `unknown` — a real local list id is never a 64-char hex string or an npub.

## Minor

`universalLinkToRouterPath` now repeats the scheme/host guard from `divineUrlToPushRoute` instead of calling it. Avoiding the double `parseDeepLink` is worth it, but the two guards can now drift — worth a tiny shared `_isDivineUniversalLink(uri)`.

PR #6708, opened 2026-08-04
fix(profile): filter invalid identity claims before verify

### LINKED ISSUE #6553: fix(profile): one malformed NIP-39 claim 400s the whole verifier batch
## Symptom

Every profile open logs a hard error and no verified-identity chip ever renders:

```
[ERROR] [BlocObserver] Bloc error: MyProfileBloc: VerifierApiException: verifier returned 400:
{"error":"Validation failed","details":[{"index":0,"error":"Invalid identity"}]}
Stack: #0 VerifierClient._post (package:verifier_client/src/verifier_client.dart:94)
```

Captured on a real user's device (Pixel 8 Pro, Android, app 1.0.17+668, `AuthenticationSource.divineOAuth`). The identical line also appears in the log attached to #6456, which was closed for an unrelated reason without this being triaged. No issue currently tracks it.

## Cause

`index: 0` indexes the `claims` array built at `mobile/packages/verifier_client/lib/src/verifier_client.dart:66-69`. The server validates each element and rejects the **entire batch** on the first failure, so a single malformed NIP-39 `i` tag on the user's kind-10011 event takes every other claim down with it. In the observed case claim 0 has the Bluesky verification *proof sentence* pasted into the identity slot, and the valid claim that follows it is collateral damage. The `pubkey` field is correct 64-char lowercase hex — the client is not sending an npub or a truncated value.

## Impact

Cosmetic but permanent, and it affects viewers as well as the account owner. The exception is caught at `mobile/lib/blocs/my_profile/my_profile_bloc.dart:305-316` and last-known-good claims are preserved, and because `VerifierApiException` does not implement `ReportableError` the observer keeps it out of Crashlytics — correct per the 4xx row of the error-handling matrix. So: no crash, no data loss, but a verified badge that can never appear and an ERROR line on every profile open.

## Suggested fix

Filter claims against the server's platform allowlist and identity charset/length rules in `IdentityClaimsRepository.parseClaims` (`mobile/lib/repositories/identity_claims_repository.dart:84-111`) before they reach `VerifierClient`, and apply `VerifierClient.maxBatchSize` **after** filtering rather than before — today the cap at `:108` runs first, so junk tags can crowd out valid ones. That restores the valid claim with no server change and no action needed from the user.

Two follow-ups outside this repo, worth filing separately rather than folding in:

- The verifier service should soft-fail per claim (it already has that shape for rate limits) instead of 400-ing the batch.
- divine-web's Linked Accounts page appears to write the proof text into the identity slot when the username field is left blank, and its read-modify-write republish then carries the malformed tag forward indefinitely. Without that fix this recurs for every user who fills the form the same way.

## Verifying

One relay query for kinds `[0, 10011]` authored by `a62a3a6b7efe2752446d014f6a50614182ff295337855b10739826560fe6265b` confirms the tag contents directly. The reconstruction above comes from reading the client, not from observing the event on the wire.

NotThatKindOfDrLiz:
## Summary

- mirror the verifier's claim validation contract on `IdentityClaim`
- filter NIP-39 claims before batching so malformed tags do not 400 the verifier
- apply the batch cap after filtering and cover the malformed-claim regression

Closes #6553.

## Follow-ups

- Verifier service should soft-fail invalid claims within a batch: divinevideo/divine-identify-verification-service#35
- Verifier UI should stop writing/carrying malformed NIP-39 tags: divinevideo/divine-web#534

## Verification

- `dart test` from `mobile/packages/verifier_client`
- `dart analyze` from `mobile/packages/verifier_client`
- `flutter analyze` from `mobile/packages/profile_repository`
- `flutter analyze lib test integration_test` from `mobile`

### DIFF
--- a/mobile/packages/profile_repository/lib/src/identity_claims_repository.dart
+++ b/mobile/packages/profile_repository/lib/src/identity_claims_repository.dart
@@ -76,11 +76,12 @@ class IdentityClaimsRepository {
   /// Parses NIP-39 identity claims out of the given identity-event tag
   /// list (kind 10011, or kind 0 for pre-migration profiles).
   ///
-  /// Filters to `['i', '<platform>:<identity>', '<proof>']` shape, skips
-  /// malformed entries, dedupes case-insensitively on `<platform>:<identity>`
-  /// (preferring the first occurrence — matches verifier UI behaviour at
-  /// `divine-identify-verification-service/src/index.ts:1784`), caps at
-  /// [VerifierClient.maxBatchSize] (10) so a single batch suffices.
+  /// Filters to `['i', '<platform>:<identity>', '<proof>']` shape and the
+  /// verifier's request contract, skips malformed entries, dedupes
+  /// case-insensitively on `<platform>:<identity>` (preferring the first
+  /// valid occurrence — matches verifier UI behaviour at
+  /// `divine-identify-verification-service/src/index.ts:1784`), caps valid
+  /// claims at [VerifierClient.maxBatchSize] (10) so a single batch suffices.
   static List<IdentityClaim> parseClaims(
     String pubkey,
     List<List<String>> tags,
@@ -95,16 +96,16 @@ class IdentityClaimsRepository {
       if (colon <= 0 || colon == claimKey.length - 1) continue;
       final platform = claimKey.substring(0, colon);
       final identity = claimKey.substring(colon + 1);
+      final claim = IdentityClaim(
+        pubkey: pubkey,
+        platform: platform,
+        identity: identity,
+        proof: tag[2],
+      );
+      if (!claim.isServerValid) continue;
       final dedupeKey = '$platform:$identity'.toLowerCase();
       if (!seen.add(dedupeKey)) continue;
-      claims.add(
-        IdentityClaim(
-          pubkey: pubkey,
-          platform: platform,
-          identity: identity,
-          proof: tag[2],
-        ),
-      );
+      claims.add(claim);
       if (claims.length >= VerifierClient.maxBatchSize) break;
     }
     return claims;
--- a/mobile/packages/profile_repository/test/src/identity_claims_repository_test.dart
+++ b/mobile/packages/profile_repository/test/src/identity_claims_repository_test.dart
@@ -73,6 +73,41 @@ void main() {
       expect(IdentityClaimsRepository.parseClaims(_pubkey, tags), isEmpty);
     });
 
+    test('skips i tags whose claims fail the verifier contract', () {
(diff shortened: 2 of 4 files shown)


rabble (approved):

rabble (commented):

The goal is right — one malformed claim shouldn't 400 an entire verifier batch, and validating client-side against the server contract is the correct shape. The dedupe-after-validation ordering is a genuine improvement, and `does not let an invalid duplicate block a valid claim` is a good test for it.

Two things I'd want resolved before merge.

## 1. Capitalised platforms are now silently dropped

`isServerValidPlatform` is an exact-match lookup into a lowercase-only `supportedPlatforms`, so `['i', 'GitHub:Octocat', 'proof']` now fails `isServerValid` and is skipped entirely. Before this PR it parsed and passed through with case preserved.

The diff makes this explicit — four existing tests were edited from `'GitHub:Octocat'` to `'github:Octocat'`, including one whose assertion was `expect(result.single.platform, equals('GitHub'))` and is now `equals('github')`. Those tests were passing on `main`; they were changed to accommodate the new behaviour rather than because they were wrong.

NIP-39 does not mandate lowercase platform names, and users' kind-10011 / kind-0 events are already published. Anyone with a capitalised platform tag loses their verified badge with no diagnostic. Suggest normalising instead:

```dart
static bool isServerValidPlatform(String value) =>
    supportedPlatforms.contains(value.toLowerCase());
```

and either lowercasing `platform` when constructing the claim, or lowercasing at the point the verifier request is built — whichever matches what the service actually accepts. Then restore the four tests to their original capitalised inputs, since they were encoding real-world tag shapes.

## 2. The allowlist trades away forward compatibility

The doc this PR replaces said "Forward-compatible: the verifier may add platforms." The hardcoded set inverts that: adding a platform server-side now requires a mobile release before those claims are even sent, and users on older builds silently lose the new platform.

That may be an acceptable trade for not 400-ing the batch, but it's a real one and it isn't called out. Two alternatives worth weighing: send unknown platforms and let the service drop them per-claim, or keep the allowlist as a *soft* filter that only applies when the batch would otherwise fail. If the hardcoded set is the right answer, a comment naming the release-coupling cost — and ideally a tracking issue for pulling the list from the service — would keep the next person from being surprised.

## Smaller notes

- `_invalidServerTextPattern` rejects `'` (apostrophe). Confirm no supported platform allows it in a handle; mastodon and bluesky are the likely edge cases.
- `_hexPubkeyPattern` accepts uppercase hex while the doc says "64-character lowercase hex pubkey". If the service is strict about case, this is a false accept that just moves the 400 rather than preventing it.

PR #6709, opened 2026-08-04
fix(bloc): guard add after close

### LINKED ISSUE #6120: fix(bloc): 'Cannot add new events after calling close' add-after-close (Crashlytics b0631106)
## Crashlytics
- Non-fatal: `b06311065575782602a6b48f40dae7b8` ([console](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/b06311065575782602a6b48f40dae7b8)) — **264 users / 741 events**, 1.0.4 → **1.0.16**
- Fatal variant: `a454fc7bcda88d465a51b758951d26b5` ([console](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/a454fc7bcda88d465a51b758951d26b5)) — **26 users / 29 events**, 1.0.14 → 1.0.15

## Signature
`package:bloc/src/bloc.dart - Bloc.add`
`Bad state: Cannot add new events after calling close.`

## Root cause (hypothesis)
An event is added to a Bloc/Cubit after it was `close()`d — a callback (stream listener, timer, animation, async continuation, or a `ref.listen`/`BlocListener` bridge) fires after the owning widget/provider disposed the bloc and still calls `bloc.add(...)`. 264 users on the shipping build; the same pattern also produced a fatal on 1.0.14–1.0.15.

## Suggested next step
This is a class of bug, not one site. Pull events for the non-fatal to see which Bloc(s) dominate the `add`-after-close, then guard those async callbacks with an `isClosed`/`mounted` check (or cancel the subscription in `close()`/`dispose()`). Highest-volume non-fatal that's ours and un-tracked.

NotThatKindOfDrLiz:
## Summary
- Forward bare invariant Bloc errors (StateError, TypeError, RangeError) through Crashlytics with Bloc diagnostic keys so add-after-close reports identify the Bloc type and last event/state.
- Guard post-await Bloc.add/emit paths in background publish, profile verifier, recorder navigation/stabilization, and MyFollowing toggle lifecycle edges.
- Add regression coverage for closed Bloc paths and observer forwarding.

Closes #6120

## Verification
- flutter analyze
- flutter test test/observability/divine_bloc_observer_test.dart test/blocs/my_following/my_following_bloc_test.dart test/blocs/background_publish_bloc/background_publish_bloc_test.dart test/screens/profile_setup_screen_test.dart test/widgets/video_recorder/modes/capture/video_recorder_capture_actions_test.dart test/widgets/video_recorder/video_recorder_navigation_test.dart

### DIFF
--- a/mobile/lib/blocs/background_publish/background_publish_bloc.dart
+++ b/mobile/lib/blocs/background_publish/background_publish_bloc.dart
@@ -228,6 +228,7 @@ class BackgroundPublishBloc
       );
     }
     for (final upload in inFlight) {
+      if (isClosed) return;
       add(BackgroundPublishVanished(draftId: upload.draft.id));
     }
   }
@@ -292,6 +293,7 @@ class BackgroundPublishBloc
 
     final videoPublishService = await _videoPublishServiceFactory(
       onProgress: ({required String draftId, required double progress}) {
+        if (isClosed) return;
         add(
           BackgroundPublishProgressChanged(
             draftId: draftId,
@@ -300,6 +302,7 @@ class BackgroundPublishBloc
         );
       },
     );
+    if (isClosed) return;
 
     final newPublishProcess = videoPublishService.publishVideo(
       draft: uploadToRetry.draft,
--- a/mobile/lib/blocs/my_following/my_following_bloc.dart
+++ b/mobile/lib/blocs/my_following/my_following_bloc.dart
@@ -147,6 +147,7 @@ class MyFollowingBloc extends Bloc<MyFollowingEvent, MyFollowingState> {
 
     try {
       await _followRepository.toggleFollow(event.pubkey);
+      if (isClosed || emit.isDone) return;
       if (!state.hasLocalFollowEdit) {
         emit(state.copyWith(hasLocalFollowEdit: true));
       }
@@ -156,6 +157,7 @@ class MyFollowingBloc extends Bloc<MyFollowingEvent, MyFollowingState> {
         name: 'MyFollowingBloc',
         category: LogCategory.system,
       );
+      if (isClosed || emit.isDone) return;
       emit(state.copyWith(status: MyFollowingStatus.toggleFailure));
     }
   }
--- a/mobile/lib/observability/divine_bloc_observer.dart
+++ b/mobile/lib/observability/divine_bloc_observer.dart
@@ -26,10 +26,12 @@ const String kBlocDiagnosticNotObserved = '<not observed>';
 ///
 /// All three are written to the unified log (so the in-memory bug-report
 /// capture flow stays complete). Forwarding to Crashlytics is **gated** on
-/// the error implementing [ReportableError] — without that gate, expected
-/// domain errors (network timeouts, "no public key yet" during cold start,
-/// 4xx responses, validation failures) flood the dashboard and drown the
-/// real bugs. See the decision matrix in `.claude/rules/error_handling.md`.
+/// the error implementing [ReportableError] or being a bare programming
+/// invariant error ([StateError], [TypeError], or [RangeError]) — without that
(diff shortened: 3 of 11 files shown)


rabble (approved):

rabble (commented):
Approving. Agent-assisted review; no branch changes.

The guards are correct and the mechanism they defend against is real. I verified against the resolved bloc version (**9.2.1**) rather than assuming:

- `Bloc.add` after close reaches `_eventController.add(event)` on a closed controller, which throws `StateError`, and the `catch` calls `onError(...)` then **rethrows**. So an unguarded post-await `add` genuinely propagates — the `isClosed` checks are load-bearing, not belt-and-braces.
- `BlocBase.emit` throws `StateError('Cannot emit new states after calling close')`, likewise rethrown, so the `emit.isDone` checks pair correctly.

The observer change also lines up with the decision matrix in `.claude/rules/error_handling.md` — `StateError` / `TypeError` / `RangeError` are the three "YES, programming-invariant violation" rows, and `_asReportable` promotes exactly those and nothing else. Wrapping in `Reportable` keeps `error.runtimeType` grouping intact for Crashlytics.

Two non-blocking follow-ups:

**1. One of the new tests may not be able to fail.** In `does not emit after close when a successful toggle completes`, the success-path `emit` sits inside the `try`, so without the guard the `StateError` is caught by the `catch`, which then calls `emit(...toggleFailure)` and throws again. Either way the state never changes, so `expect(bloc.state.hasLocalFollowEdit, isFalse)` passes with **and** without the fix. Worth confirming it goes red on revert; if not, asserting on the absence of an error (or on `errors: () => []`) would pin the actual behaviour. The `does not emit after close when a toggle fails` sibling likely has the same shape.

The `background_publish` pair are fine — `returnsNormally` around `capturedProgress(...)` does flip, since the unguarded `add` rethrows.

**2. `.claude/rules/error_handling.md` is now slightly stale.** Its migration recipe presents call-site `Reportable(e, context: ...)` wrapping as the way an error becomes reportable, and states the observer gates on `ReportableError`. After this PR bare invariant errors are auto-promoted at the observer, so a `StateError` reaching the generic `catch` is reportable whether or not anyone wrapped it. Worth a short paragraph in that doc so the next person doesn't wrap redundantly — or conclude from the doc that an unwrapped `StateError` is being dropped.

rabble (approved):
Approving — see detailed review above.

rabble:
**Correction to my earlier review — point 1 was wrong. Please disregard it.**

I claimed `does not emit after close when a successful toggle completes` might pass with and without the guard. I've now checked it instead of reasoning about it: worktree on this branch, both `if (isClosed || emit.isDone) return;` lines removed, tests re-run.

Both fail:

```
does not emit after close when a successful toggle completes [E]
  Expected: false
    Actual: <true>

does not emit after close when a toggle fails [E]
  Expected: MyFollowingStatus:<MyFollowingStatus.initial>
    Actual: MyFollowingStatus:<MyFollowingStatus.toggleFailure>
```

My error: I'd read `Bloc.add`'s source (which does throw on a closed controller) and assumed `emit` behaves the same, so state couldn't change either way. It doesn't. `close()` awaits in-flight emitters *before* closing the state controller, so a post-await `emit` in a still-running handler **succeeds** — which is exactly the defect being fixed here: the bloc mutates state during teardown. `isClosed` flips the moment `close()` is called, so the guard catches it and the assertions go red without it.

Both tests are valid and correctly targeted. Nothing to change.

Point 2 stands — `.claude/rules/error_handling.md` still describes the observer as gating solely on `ReportableError`, which this PR changes.

Approval unaffected.

PR #6710, opened 2026-08-04
test(deep-links): guard AASA claim coverage

### LINKED ISSUE #6702: test(deep-links): guard AASA claims against mobile routing drift
## Summary

The production AASA payload and the checked-in mobile deep-link reference can drift from what the app actually routes. That drift is already visible: production claims `/invite/*` and `/list/*`, while `mobile/docs/apple-app-site-association` only lists video/profile/hashtag/search/callback.

## Problem

When AASA claims a path that `DeepLinkService` and the router do not handle, iOS can take a working browser link away from Safari or an external app and open Divine into a no-op. That is the current risk for `/list/*` until #6700 is implemented.

## Scope

- Decide and document the local source of truth for the AASA payload used by mobile tests.
- Update `mobile/docs/apple-app-site-association` or replace it with the current source-of-truth fixture so it matches production claims intentionally.
- Add a guardrail test that compares claimed AASA components with mobile deep-link coverage.
- Any claimed path should be either routed by mobile or explicitly documented as intentionally claimed but handled somewhere else.

## Acceptance

- The checked-in AASA reference includes the same intentional path claims as production, including `/invite/*` and `/list/*` if they remain claimed.
- A local test fails when a new AASA path is claimed without corresponding mobile route coverage or an explicit allowlist reason.
- The test covers both apex and `www.divine.video` expectations where host-specific behavior matters.
- Existing deep-link tests remain focused on parser/router behavior and do not require live network access.

## Related

- Follow-up from #3843 after universal-link handoff was restored and physically verified.
- Related routing bug: #6700.

NotThatKindOfDrLiz:
## Description

Adds an offline AASA claim coverage guardrail for the Divine associated
domains. The new test pins the current checked-in AASA path claim sets, filters
claims to `GZCZBKH7MY.co.openvine.app`, ignores excluded components, and
synthesizes representative URLs for each claim. It fails unless the claim is
parsed by mobile deep-link routing or has an explicit allowlist reason.

This also retires the stale mobile-owned AASA payload/template. Mobile now keeps
mirrored AASA fixtures under `test/fixtures/deep_links/` only for offline route
coverage tests, and the deep-link docs call out the source split:

- `divine.video` / `www.divine.video`: `divine-web`
- `login.divine.video`: `keycast`

Closes #6702

## Verification

- `flutter test test/router/aasa_claim_coverage_test.dart`
- `flutter test test/router/aasa_claim_coverage_test.dart test/services/deep_link_service_test.dart test/router/universal_link_resolver_test.dart test/router/route_normalization_test.dart`
- `flutter analyze`
- pre-push analyzer and changed-file tests passed

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore
- [x] ✅ Test coverage


NotThatKindOfDrLiz:
Takeover update pushed in eaf8352f6.

What changed:
- Rebased the PR onto current origin/main so the diff is back to the intended 7-file scope.
- Tightened the AASA guard to filter claims to GZCZBKH7MY.co.openvine.app and ignore excluded components.
- Expanded wildcard coverage to include the deeper profile/hashtag/search/list paths mobile routes today.
- Removed the redundant fixture-pinning test and kept expected claims asserted in the per-host guard.
- Documented the AASA source split: apex/www from divine-web, login.divine.video from keycast, with fixture refresh curl commands.
- Updated the PR body to remove the stale #6707 dependency note.

Verification:
- flutter test test/router/aasa_claim_coverage_test.dart
- flutter test test/router/aasa_claim_coverage_test.dart test/services/deep_link_service_test.dart test/router/universal_link_resolver_test.dart test/router/route_normalization_test.dart
- flutter analyze
- pre-push analyzer + changed-file tests
- GitHub checks are green on eaf8352f6.

rabble (approved):

Deleting the mobile-owned AASA payload is the right call and it's safe — I grepped the tree and the only surviving reference to `mobile/docs/apple-app-site-association` is a line in an old superpowers plan doc, not any deploy script or CI step. A checked-in copy of a file this repo doesn't serve is worse than no copy, because it looks authoritative while silently drifting from what `divine-web` and `keycast` actually publish. The docs now say who owns which host, which is the part that was actually missing.

The test itself is better than most guardrails of this shape. Filtering `details` to `GZCZBKH7MY.co.openvine.app`, honouring `exclude`, and synthesizing representative URLs per pattern means it exercises `DeepLinkService.parseDeepLink` rather than string-matching the JSON. The bidirectional assertion is the good part: a claim with no route *and* no allowlist entry fails, and an allowlist entry for a claim that's since become routable also fails. That second direction is what keeps the allowlist from silently rotting into a list of excuses, and it's the half people usually skip.

**Worth being precise about what this can and can't catch**, because the ABOUTME says "Guards served AASA path claims" and it doesn't — it guards *mirrored* claims. The fixture is a snapshot. If `divine-web` adds `/foo/*` to the real payload tomorrow and nobody re-curls the fixture, mobile has zero routing coverage for a live universal link and this test stays green. That's precisely the drift direction most likely to happen, since the payload is owned by a repo whose authors have no reason to think about this test.

The `expect(patterns, _expectedClaimPatternsByHost[host])` pin does force a refresh to be deliberate once someone *does* re-curl, which is the achievable half. Closing the rest needs something that actually fetches — a scheduled CI job that curls both hosts and diffs against the fixtures would do it, and it belongs outside the offline suite. Not asking for it here; just don't let the file name convince anyone the served payload is covered.

**One thing I'd genuinely want someone to look at, though it's not yours to fix in this PR.** The login fixture shows `GZCZBKH7MY.co.openvine.keycastFlutterDemo` claiming `/verify-email` and `/reset-password`, and carrying `webcredentials`:

```json
"appIDs": ["GZCZBKH7MY.co.openvine.keycastFlutterDemo"],
"components": [ ..., {"/": "/reset-password", ...} ]
```

A bundle ID with `Demo` in the name holding password-reset universal-link claims and shared web credentials on the production login host is worth a second look. Same team ID, so this isn't a third-party hijack — but if that demo app is no longer shipped it's dead claim surface, and if it *is* installed anywhere it will contend for password-reset links with the real app. Mirroring it into this repo is the first time it's been written down where a mobile reviewer would see it, which is exactly when to ask. Worth raising with whoever owns keycast rather than changing anything here.

Ship it.

PR #6711, opened 2026-08-05
fix(likes): show likers stranded on a superseded video revision

### LINKED ISSUE #6021: fix(likes): "Liked by" list undercounts likes stranded on superseded event ids after an edit
## Summary

The "Liked by" list can show **fewer people than the headline like count** for an edited video. The list resolves likers via a relay query that unions only the *current* event id (`e`) with the coordinate (`a`). Reactions that carry **no `a` tag** and reference **only a superseded event id** — i.e. likes published before mobile added the `a` tag (2026-02-01, #1218) and web before 2025-12-03 (#62), or by third-party clients — match neither filter, so they are missing from the list even though funnelcake now counts them in the headline via its coordinate resolver.

## Evidence

- Mobile: `fetchEventLikers` / the addressable `getLikeCount` path query `Filter(kinds:[7], e: [currentIds])` unioned with `Filter(a: [coordinate])` (`mobile/packages/likes_repository/lib/src/likes_repository.dart:1208-1217`, `:593-609`). The `e` filter uses the current video id(s) only — never the superseded ids — so an `e`-only reaction on an old id matches nothing.
- Backend: `get_video_likers` similarly matches `e` against the single resolved id unioned with the `a` coordinate (`divine-funnelcake/crates/clickhouse/src/client.rs:2575-2656`), so the same gap exists server-side; the list has no join against historical version ids, unlike the coordinate **count** resolver.
- Direction: because the count (funnelcake coordinate resolver) recovers superseded-id reactions but the list does not, `count > list` is now the steady state for edited videos that have pre-`a`-tag likes.
- CatesPlace (`713fc3bf6c1c5cc593736c2b1375ba165e173b05d31787ed3b0591796a489a3f`) is exempt — their likes are 100% `a`-tagged, so the coordinate filter recovers all of them; accounts with older or third-party likes are affected.

## Suggested fix (needs a call)

Either:
- source the "Liked by" list count from funnelcake's coordinate resolver (which already recovers superseded-id reactions), or
- extend the resolver (mobile `_fetchResolvedLikersByEvent` and/or backend `get_video_likers`) to also query the coordinate's historical event ids.

Relates to divinevideo/divine-funnelcake#626 (count vs list divergence: no `-` downvote / kind-5 deletion / reactor-pubkey dedup filtering in the count) and #197 (user-aggregate `total_reactions`).

## Test plan

Resolver test with a mixed reaction set (`a`-tagged reactions + `e`-only reactions on a superseded id) asserting the "Liked by" list cardinality equals the headline count after an edit.

---

### Context

Follows from #4432 (lost likes on edit). That issue's core symptom — the displayed like count collapsing after an edit — was a backend counting bug and was fixed by divinevideo/divine-funnelcake#568 (migration `000180`, coordinate-aware counting), deployed to production 2026-07-10 06:42 UTC. This issue tracks the residual "Liked by" list undercount that survives that fix (the list resolver was not made coordinate/historical-id aware).

realmeylisdev:
Fixes #6021

## Summary

The "Liked by" list can show **fewer people than the like count above it**, permanently, for any video that has been edited.

A Divine video is a NIP-33 addressable event: editing it keeps the coordinate `kind:pubkey:d` but mints a **new event id**. The list queries kind 7 reactions by `e` = the current event id ∪ `a` = the coordinate, so a reaction carrying only an `e` tag on a now-superseded revision matches neither filter and disappears — while funnelcake's coordinate-keyed aggregate (migration `000180`) goes on counting it.

**Read the measurement below before judging this PR.** I originally justified it as a live undercount and backed that with an end-to-end run. Having now measured production properly, both of those were wrong: the hole is real and protocol-legal, but it is currently near-empty and closing. This is prospective insurance, and it should be reviewed as such.

## Measured impact: the hole is real, and currently near-empty

`POST /api/videos/revisions/bulk` went live on `api.divine.video` after funnelcake#799 shipped, which finally made it possible to ask production which videos were actually edited instead of reasoning about it. Two sweeps against `relay.divine.video`, with the resulting liker sets cross-checked against funnelcake's coordinate-aware counts (28/27, 24/24, 16/17 — small deltas in both directions, explained by the five known count-vs-list causes, so the relay view is essentially complete):

| Question | Answer |
|---|---|
| Are videos actually edited? | **Yes — 41 of 373 sampled (11%)** |
| Do reactions really strand on superseded ids? | **Yes — 74 across 10 edited videos** |
| Do those reactions also carry an `a` tag? | **211 of 211 do** |
| So how many likers does this PR actually recover today? | **Zero, on all 13 videos measured** |

Every reaction that `e`-tags a superseded revision also carries the `a` tag, so the coordinate filter on `main` already finds it. **On current production data this PR changes no liker list.**

I also claimed this was "not a shrinking backlog of old clients". That is backwards. Across 3103 reactions on 480 videos spanning 2017‑01 → 2026‑08:

| Reaction month | Reactions | `e`-only | Rate |
|---|---|---|---|
| 2026-02 | 288 | 5 | 1.74% |
| 2026-05 | 395 | 1 | 0.25% |
| 2026-06 | 480 | 0 | 0.00% |
| 2026-07 | 846 | 0 | 0.00% |
| 2026-08 | 1071 | 0 | 0.00% |
| **overall** | **3103** | **6** | **0.19%** |

It is aging out. And the emitters are **our own clients** — `divine.video` web ×4 and untagged ×2 here, plus one from Divine mobile dated 2026‑08‑07 in the other sweep — not the conformant third parties I invoked. NIP-25 does make `a` a SHOULD and `e` a MUST, so the hole stays legal and can reopen at any time; but "conformant clients are actively widening it" was an inference I did not check, and it does not hold.

### Retraction: the earlier end-to-end run

An earlier revision of this description reported a device run showing `19 → 33` and `30 → 54` likers. **That was not evidence about production.** No production video exhibits this bug, so the run **manufactured** the condition by pairing two unrelated production videos as if one were the other's superseded revision. It exercised the code path; it did not demonstrate a real stranded liker. I've removed it rather than leave it reading as field evidence.

What that run does still support — because it does not depend on the pairing — is the two-phase timing: with the lookup held 14s, the fetch answered at +6s with `revisionsPending=true`, and the wider list was published at +15s when the lookup landed. That is the shape rabble asked for, observed on a device.

## Why this needs a backend lookup

Nothing on the device can recover a coordinate's superseded ids. Measured against production `relay.divine.video`:

| Probe | Result |
|---|---|
| `{"kinds":[34236],"authors":[pk],"#d":[d]}` (± `limit`) | latest revision only |
| same + `"until"` set *inside* the superseded event's window | **0 events** — the coordinate collapses *before* `until` is applied |
| `{"ids":[old]},{"ids":[new]}` | both — but only if you already know the ids |

No funnelcake endpoint exposed version history, and mobile's own cache actively deletes older revisions on upsert (`nostr_events_dao.dart`, *"Delete all older versions before inserting the new one"*).

**Requires divinevideo/divine-funnelcake#799**, which adds `POST /api/videos/revisions/bulk`. That merged on 2026‑08‑06 and is now live — the spec lists 70 paths including this route, and it answers the contract exactly (`{"event_ids":[]}` → 400 `{"error":"event_ids cannot be empty"}`; an unknown id → 200 `{"revisions":{},"missing":[…]}`).

## The lookup is never on the critical path

The backend round-trip cannot produce the wider list in time for a fetch that has to answer now. Earlier revisions of this PR awaited it and tried to bound the wait; that made the resolver's budget a ceiling on the *whole* fetch, and both directions were wrong — 750ms silently skipped enrichment and reinstated the bug, 3s stalled the likers list and the like count (and, via the one `Future.wait` in `video_interactions_bloc.dart:220`, the comment and repost counts with them).

So nothing waits. The fetch answers from the ids it has, and the wider list is published separately:

```
fetch  ──► current-id + a-tag relay queries ──► list returned
       └─► revision lookup ─────────────► re-resolve ──► watchRevisionEnrichedLikers
```

- **The lookup is read, never awaited.** `_resolveRevisionTargets` is synchronous; revisions already resolved this session widen the query with no `await` between the fetch and its relay queries.
- **A lookup that has not landed schedules a background pass** that re-resolves the video and publishes the result. `VideoEngagementBloc` re-emits the list; `VideoInteractionsBloc` raises the count from the same set, so the headline cannot end up below the list behind it.
- **Only an edited video pays that second pass.** A resolver answering "no other revisions" costs nothing beyond its own lookup.
- **The pass is scheduled once per video** even when the feed's count and the sheet race on it, swallows its own relay failures (nothing awaits it), and publishes only when the list actually grew.
- **The timeout survives as a backstop**, not a budget: it bounds the background wait so a resolver that never completes cannot leak a continuation.

## Commits

1. **`feat(funnelcake-client): look up a video's sibling NIP-33 revisions`** — `getBulkVideoRevisions` wrapping the new endpoint, mirroring the existing bulk-stats method.
2. **`fix(likes): show likers stranded on a superseded video revision`** — `_fetchResolvedLikersByEvent` resolves each addressable target's revisions and queries those ids too. Both the `e:` filter **and** the acceptance set widen — widening only the filter would fetch the reactions and then discard them.
3. **`fix(likes): publish revision enrichment instead of waiting for it`** — the shape above.
4. **`fix(likes): show the wider liker list in the feed and the "Liked by" sheet`** — the two consumers.

The rest are review follow-ups: chunking at the endpoint cap, the session cache, retrying a lookup that could not answer, and the doc/refactor commits.

## Design notes

- **The resolver is an injected typedef**, not a client import, so `likes_repository` stays free of any API-client dependency — the same port shape `dm_repository` uses for its error reporter.
- **The resolver is shared with the counts.** `_fetchResolvedLikersByEvent` also backs `getLikeCount` and the batched `getLikeCounts`. That is intentional: #6022 deliberately coupled them so count and list cannot diverge at fetch time.
- **Reaction queries are chunked** on the same bound as the deletion probe, so a coordinate with many revisions cannot build a REQ frame over the relay's `max_message_length`.
- **Revisions are cached per event id for the session.** That cache is sound rather than merely convenient: an id's revision set cannot change, because editing the video mints a new id, which is a new key. Each entry holds the lookup's *future* — so concurrent fetches share one request and a lookup nothing is waiting on still lands — alongside its recorded result.
- **A count computed while the lookup was still running is returned but not cached**, so a single slow lookup cannot pin the un-enriched number for the session.
- **The cache is unbounded**, like the `_likeCountCache` beside it, and cleared by `clearCache()` on sign-out.
- **`VideoEngagementBloc` keeps the wider list if it arrives mid-fetch.** The profile pre-warm alone holds the load for up to 2s, which is long enough for enrichment to land first.

## Degradation

No resolver, a non-addressable target, a transport error, or a malformed payload all fall back to **exactly the previous behaviour**: the current-id list, and nothing published. The client returns an empty map rather than throwing on every failure path. Nothing here can take the "Liked by" list down.

## Testing

`likes_repository` — 223 tests, `funnelcake_api_client` 379, `video_engagement` / `video_interactions` blocs 76, `flutter analyze lib test integration_test` clean.

**Every regression test was verified to fail without its fix**, by mutating the code it pins and confirming that exact test goes red:

| Mutation | Test that catches it |
|---|---|
| await the pending lookups before querying | `returns without waiting for the revision lookup` (`Actual: <null>` — the fetch had not answered) |
| narrow the test resolver to `[current]` | `publishes a liker stranded on a superseded revision` (times out — nothing published) |
| re-resolve every video, not just edited ones | `publishes nothing when the video was never edited` (4 relay queries, not 2) |
| publish even when the list did not change | `publishes nothing when the wider query finds no one new` |
| drop the once-per-video schedule guard | `re-resolves a video once for concurrent fetches` |
| let a relay failure escape the background pass | `survives a relay failure in the re-resolve` |
| merge only the first resolver chunk | `chunks revision resolver requests at the server limit` (`Expected: <1> Actual: <0>`) |
| drop the per-video filter on the stream | `ignores an update for a different video` (both blocs) |
| drop the never-lower floor | `never lowers a higher count` |
| let the fetch clobber a mid-fetch enrichment | `keeps the wider list when it lands mid-fetch` |

**No test asserts `list.length == headlineCount`**, and that is deliberate. Five other causes inflate the count independently of this bug — downvotes, kind 5 deletions, per-pubkey duplicates, blocked reactors, and banned/suspended/vanished reactors — so such an assertion would be flaky against real data.

## Not in scope

`reposts_repository` and `comments_repository` have the identical single-id ∪ coordinate shape, and divine-web has it independently in `useVideoSocialMetrics`. Filed separately (#6712) rather than widened here. Both halves have since been measured against production: the comments half is real — five comments across four edited videos are invisible today — but its fix is server-side, in divinevideo/divine-funnelcake#873, because the app reads comments from funnelcake REST and never reaches the relay branch on page one. The reposts half turned up no live instance in 2,110 sampled video reposts, so it is prospective only.

**Worth a follow-up, surfaced by the measurement:** every `e`-only reaction found came from a Divine client. `LikesRepository.likeEvent` emits no `a` tag when its `addressableId` argument is null, so the durable fix for this hole is writer-side — always tag the coordinate — which would close the source rather than compensating for it at read time. This PR still earns its place for reactions already written and for third-party clients, but that one-line direction is the cheaper long-term answer and is not addressed here.

## Screenshots

Not applicable — no UI change. The list renders the same way; it just stops omitting people.


rabble (changes requested):
Request changes.

1. The revision lookup is awaited before the normal current-ID reaction query starts in `mobile/packages/likes_repository/lib/src/likes_repository.dart` (the `_fetchResolvedLikersByEvent` path). `getBulkVideoRevisions` has a 15-second client timeout, so an unavailable or slow optional endpoint can delay the baseline likes query by the full timeout. Start the current-ID query immediately and ensure revision enrichment cannot hold the baseline result hostage.

2. `getBulkVideoRevisions` documents a 100-ID server cap, but `_resolveRevisionTargets` forwards the entire `addressableEventIds` list. Bulk callers can exceed that limit and silently lose stranded likers when the API truncates or rejects the request. Chunk requests at the endpoint limit and merge the results, with a regression test for more than 100 requested IDs.

rabble (commented):

Strong PR. The problem statement is the best part — the `VideoRevisionsResolver` doc explains precisely *why* a client cannot derive these ids itself (coordinate REQ collapses to latest, `until` walk returns nothing because collapse happens first, local cache deletes older revisions on upsert). That saves the next reader from re-deriving it, and it justifies the funnelcake dependency instead of asserting it.

The port typedef is the right boundary — `likes_repository` stays free of an API-client dependency, matching the reporter-port precedent in `.claude/rules/error_handling.md`. Degrading to an empty map on every failure path, with tests for each (unconfigured, empty, 404, throw, malformed), is exactly right for an enhancement that must never take the list down.

`counts a liker once when found via both revisions` and `excludes a downvote stranded on a superseded revision` are the two tests I'd have asked for — the second in particular proves the existing NIP-25 filters still apply to reactions the widening newly reaches.

Three things worth addressing:

## 1. The two round-trips are serialized

The comment says "start it first and let the two round-trips overlap", and that's true of `aEventsFuture`. But `_resolveRevisionTargets` is `await`ed *before* the `Future.wait`, so the reaction query now waits on the funnelcake HTTP call:

```
revisions (HTTP)  ──────────►
                             e-query (relay) ──────────►
a-query (relay)   ─────────────────────────►
```

"Liked by" latency becomes funnelcake RTT + relay RTT where it used to be one relay round-trip. Options: fire the e-query for the known ids immediately and issue a second widened query only for revisions that come back, or keep it serialized but bound it — which leads to:

## 2. No timeout on the resolver

`VideoRevisionsResolver`'s doc says implementations must not throw, and `getBulkVideoRevisions` honours that. But a *hang* isn't a throw. If funnelcake is slow rather than down, `fetchEventLikers` hangs with it, and the `catch` never fires. A `.timeout(...)` around the `resolver(...)` call inside `_resolveRevisionTargets`, falling into the same empty-map path, would make the "never a precondition" guarantee actually hold. Worth stating the bound in the typedef doc too, since implementers currently only know they mustn't throw.

## 3. The documented chunking contract has no implementer

`getBulkVideoRevisions` says: *"[eventIds] is capped at 100 by the server; longer lists are truncated there, so callers should chunk."* The only caller — `_resolveRevisionTargets` — passes `addressableEventIds` straight through with no chunking. Single-video `fetchEventLikers` is fine, but any bulk path over 100 addressable ids silently loses revisions past the cap, and silently is the bad part: truncation is indistinguishable from "no revisions exist". Either chunk in `_resolveRevisionTargets` or move the chunking inside `getBulkVideoRevisions` so the contract is self-enforcing.

## Minor

- `_queryReactionsByEventIds` reuses `_deletionReqEidChunkSize`. The rationale is the same (`max_message_length`) but the name now misleads — something like `_reqEidChunkSize` covers both callers.
- `getBulkVideoRevisions` swallows into `on Object catch (_)` with no log. Everything else on this path logs; a `Log.warning` here would make "revisions silently unavailable" diagnosable in a bug report.

NotThatKindOfDrLiz:
@realmeylisdev I pushed a takeover follow-up in f9e34edc5:

- chunks `getBulkVideoRevisions` resolver calls at the backend 100-id limit, with a >100 regression test
- starts the baseline current-id relay query immediately and uses a second additive relay query for resolved revision ids, so a slow or missing Funnelcake endpoint no longer delays existing behavior
- logs non-200, malformed, and exception revision lookup degradation while still returning `{}`

Local/pre-push verification passed:
- `flutter test test/src/likes_repository_test.dart`
- `flutter test test/src/funnelcake_api_client_test.dart`
- `flutter analyze lib test` in `likes_repository`
- `flutter analyze lib test` in `funnelcake_api_client`
- pre-push generated-file check and changed-file tests

CI is green on the PR head after rerunning the failed shard 2 job. Please review the takeover commit and merge when you are ready. The mobile path now degrades safely before the backend is available, but the stranded-revision recovery still needs divinevideo/divine-funnelcake#799 merged and deployed.

NotThatKindOfDrLiz (approved):
Takeover follow-up resolves the requested changes I found: revision lookup is chunked at 100, baseline relay queries start immediately, and degradation now emits diagnostics while preserving the existing fallback. CI is green on f9e34edc5 after rerunning the failed shard. Backend deployment of divinevideo/divine-funnelcake#799 is still required for the additive revision data to become effective.

NotThatKindOfDrLiz:
@realmeylisdev I amended the takeover follow-up to 01cc4d5f5 after Rabble's feedback-only review on the first takeover commit. The current head now also bounds revision enrichment with a 750ms resolver timeout, documents that resolver implementations must not hang, and adds a timeout regression test so a slow Funnelcake lookup cannot hold the baseline current-id liker result hostage.

Current-head verification:
- `flutter test test/src/likes_repository_test.dart`
- `flutter test test/src/funnelcake_api_client_test.dart`
- `flutter analyze lib test` in `likes_repository`
- `flutter analyze lib test` in `funnelcake_api_client`
- pre-push analyzer, generated-file check, and changed-file tests
- all GitHub checks green on `01cc4d5f5`

Please review the takeover commit and merge when you are ready. The mobile path degrades safely before the backend is available, but the stranded-revision recovery still needs divinevideo/divine-funnelcake#799 merged and deployed.

NotThatKindOfDrLiz (approved):
Approving current head 01cc4d5f5. The requested changes and follow-up feedback are addressed: revision lookup is chunked at 100, baseline current-id relay queries start immediately, resolver failures and malformed client responses log diagnostics, and revision enrichment is bounded by a 750ms timeout with a regression test. CI is green. Backend deployment of divinevideo/divine-funnelcake#799 is still required for the additive revision data to become effective.

realmeylisdev:
@rabble — all five points from your review are addressed. Summary of what landed, plus one thing I changed on measurement that wasn't in your list.

Credit where it's due: `01cc4d5f5` is @NotThatKindOfDrLiz's — it covers your first three points and the logging nit.

**Your review**

| Point | Where |
|---|---|
| 1. Serialized round-trips | `01cc4d5f5` — the current-id `#e` query now starts immediately alongside the `a` query (which `bacbce818` had already unblocked); revision ids are fetched by a second additive query once the resolver returns |
| 2. No timeout on the resolver | `01cc4d5f5` — `.timeout(...)` around the resolver call, falling into the same empty-map path, injectable via constructor; typedef doc now states the bound |
| 3. Chunking contract had no implementer | `01cc4d5f5` — `_resolveRevisionTargets` chunks at 100 and merges; chunks run concurrently |
| Nit: `_deletionReqEidChunkSize` misleads | `7dadf0b0a` — renamed to `_reqEidChunkSize`, doc covers both callers |
| Nit: silent swallow in `getBulkVideoRevisions` | `01cc4d5f5` — logs on the three failure paths (non-200 with status/URL/count, malformed payload with shape, thrown error). The unconfigured/empty-input early return stays silent, which seems right — it isn't a failure |

**The regression test for >100 didn't cover the loss it was named for** (`12332e9d7`)

The original version asserted the resolver was called in 100-id chunks, but its stub returned each id's own id as that id's only revision — which `_resolveRevisionTargets` discards as a self-alias. So the enrichment path ran with nothing to enrich. Merging only the first chunk, or discarding every chunk, left the whole suite green. An id in the second chunk now carries a liker stranded on a superseded revision, so dropping that chunk fails the test.

**The 750ms budget was too tight — changed to 3s** (`ee8ad4a28`)

This one is measured, not a hunch. On the iOS Simulator over wired networking, **4 of 17** revision lookups timed out at 750ms — the figure in `ee8ad4a28`'s message and code comment. A second session, the one that exercised the "Liked by" screen, added 2 of 3 more, so it is 6 of 20 across both runs. Either way it is against an endpoint that is currently only 404ing and doing no database work.

Round-trips to production from a wired connection:

| Request | Work done | RTT |
|---|---|---|
| `POST /api/videos/revisions/bulk` → 404 | none (route absent) | 0.405–0.715s |
| `POST /api/videos/stats/bulk` → 200, 1 id | real ClickHouse query | 0.395–0.712s |
| `POST /api/videos/stats/bulk` → 200, 100 ids | what the chunker sends | 0.454–0.712s |

Once divinevideo/divine-funnelcake#799 deploys, `revisions/bulk` does real work, and users are on cellular rather than wired — so the rate goes up from there. Each expiry silently skips enrichment and reinstates #6021 for that fetch, which means a too-tight budget ships a fix that looks present but is intermittently absent. Worth noting the endpoint sets `Cache-Control: public, max-age=60` on a **POST**, which caches generally ignore, so there is no caching to absorb this.

Being straight about the cost, since it isn't free: the returned future still waits for enrichment to finish or expire, so raising the budget raises the worst-case "Liked by" fetch from ~750ms to ~3s when funnelcake is hung. What it does not do is add to the baseline — the current-id and `a` queries start first and run concurrently, so the cost is `max(baseline, enrichment)` rather than the sum, and 3s stays far inside the client's own 15s request timeout. The trade I made is a rare 3s wait over a routine silent no-op. If you'd rather not accept that, the cleaner shape is to return the baseline immediately and enrich in a second emission, but that changes the method's contract and felt out of scope here — happy to go that way if you prefer.

The test builds its repository without an explicit timeout so it exercises the shipped default rather than a value restated in the test, and asserts a 1s resolver still enriches; that fails at 750ms.

**Verified on the iOS Simulator** (iPhone 17 Pro Max, iOS 26.5, on `7dadf0b0a` — i.e. still at the 750ms budget)

Both degradation paths fire and the UI carries on — feed scrolls, counts render, "Liked by" opens and lists likers, no hang and no unhandled error. Two things worth knowing for anyone testing this by hand:

- These warnings log under `LogCategory.api`, which **no stock build enables** (debug turns on `{system, auth, video, relay, ui}`; release `{system, auth}`). You need both `--dart-define=LOG_LEVEL=debug` and `--dart-define=LOG_CATEGORIES=...,api` — `LOG_CATEGORIES` alone is ignored, because `main.dart` only skips its hardcoded override when `LOG_LEVEL` is set. Bug reports are unaffected: capture to `LogCaptureService` is not gated on category.
- `--dart-define=BACKEND_URL` does not repoint the funnelcake client. Nothing outside `app_config.dart` reads `AppConfig.backendBaseUrl`; the base URL comes from `resolveApiBaseUrlFromRelays`, which returns a hardcoded `https://api.divine.video` whenever `relay.divine.video` is configured.

**Not blocking, but worth flagging**

Merging this changes nothing user-visible until funnelcake#799 deploys — `POST /api/videos/revisions/bulk` returns 404 today, so the client degrades to current ids alone. Separately, divine-web and divine-space still publish reactions with `['e', video.id]` and no `a` tag, so the stranded population keeps growing after this ships; that seems worth its own issue.

Ready for another look when you have a moment.

hm21 (changes requested):
The core fix is right and Rabble's five points are addressed. My concern is the last commit, `ee8ad4a2`, which widened the resolver budget 750ms → 3s after both approvals.

Its justification says the budget "only bounds the enrichment tail" and that "the list is never held back by this". That isn't what the code does. `_fetchResolvedLikersByEvent` awaits `_resolveRevisionTargets` before its `Future.wait`, so nothing returns until the resolver settles. The queries overlap; the result doesn't.

Measured on this branch, hung resolver, shipped 3s default, baseline relay data available immediately:

| Call | Elapsed |
|---|---|
| `fetchEventLikers` (addressable) | 3011 ms |
| `getLikeCount` (addressable) | 3003 ms |

I pushed `e75600a24` correcting that comment — docs only, no behaviour change. Tests and analyze green.

The budget itself is your call, but three things to weigh:

1. In `video_interactions_bloc.dart:220` the like count shares one `Future.wait` with the comment and repost counts, so a slow funnelcake delays those too.
2. The result is written to the session cache, so a single timeout pins the un-enriched (lower) count for the rest of the session.
3. divine-funnelcake#799 is still open, so today every addressable count fetch and every "Liked by" open pays a POST that 404s — the 0.4–0.7s you measured — for no benefit yet.

The 750ms → 3s tradeoff is real in both directions (too tight silently skips enrichment, too wide stalls the fetch), which suggests the budget isn't the right lever. Returning the baseline result and letting enrichment land in a second emission would remove the tradeoff entirely — worth considering, though clearly not this PR.

Two minor, separate from the above:

- `_awaitLater` (`future.ignore(); return future;`) is an unusual idiom with no comment saying why it exists — one line about suppressing the unhandled-error report while the resolver is awaited would help the next reader.
- In `_resolveRevisionChunk`, `.then(onError:)` already maps resolver errors to the empty map, so the surrounding `catch` is dead for that path.

realmeylisdev:
Thanks — and thanks for pushing `e75600a2`; the comment was wrong and your correction stands as written.

Rebased onto `main` first (the branch had picked up a conflict with #6738 in `likes_repository.dart`), then three commits.

## The budget

You're right that it isn't the lever. There's no number that is both wide enough not to skip enrichment and narrow enough not to stall a fetch. So instead of picking one, I made the budget cost less: resolved revisions are now cached per event id for the session. That's sound rather than merely convenient — an id's revision set can't change, because editing the video mints a new id, which is a new key.

The map holds the lookup's *future*, not its result, so a lookup that overruns the budget keeps running into the cache instead of being thrown away. Measured on the branch with your harness — hung/slow resolver, shipped 3s default, baseline relay data available immediately:

| Resolver | Before | After |
|---|---|---|
| answers in 700ms (today's 404 shape) | 703ms every fetch | 702ms, then 0ms |
| overruns the budget, lands at 4s | 3003ms every fetch | 3002ms, then 0ms |
| never answers | 3011ms every fetch | 3007ms, then 3002ms |

That last row is the honest remaining limitation: a genuinely hung funnelcake is re-awaited until the client's own 15s request timeout resolves it, so the few fetches inside that window each pay the budget. Bounded, not free.

Against your three points:

1. The shared `Future.wait` in `video_interactions_bloc.dart:220` still holds the comment and repost counts behind the like count — but now only on a video's *first* engagement fetch, not every one.
2. Fixed directly. `_fetchResolvedLikersByEvent` now reports whether enrichment was cut short, and `getLikeCount`/`getLikeCounts` return that count without writing it to the session cache. A timeout costs one fetch instead of the session, and the next read picks up the enrichment that has since landed.
3. Down to one 404 per video per session from one per fetch. Merge order is unchanged — funnelcake#799 lands and deploys first.

So I kept 3s. With the cost paid once per video, the argument for the wide end still holds and the argument against it mostly doesn't.

## `_awaitLater`

Documented. I checked the claim rather than reasoning about it: delete the `ignore()` and *"fetchEventLikers throws FetchLikersFailedException when relay query fails"* goes red on the stray async error, while its expectation still passes.

## The dead catch

Removed — but not the way it looks like it should be removed, which is worth flagging since it's a trap.

Collapsing both stages into a single `catch` around the `await` is not equivalent. `Future.timeout` reports the *source* future's error as unhandled even when the awaiting frame catches it, so that version leaves a stray async error behind and turns *"still returns likers when the resolver throws"* red. The `.then(onError:)` is load-bearing; only the second `catch` was dead, exactly as you said. Both forms are verified and there's a comment on it now so the next reader doesn't retry the collapse.

## The second-emission design

Agreed on both counts — it's the right shape, and it isn't this PR. The session cache narrows what it would buy (the stall is first-fetch-only now) without removing it.

Five new tests, each checked against the change it pins: dropping the cache read, discarding an overrun lookup, caching the degraded count, and leaving the cache in place across `clearCache` each turn exactly their own test red. `likes_repository` 217, `funnelcake_api_client` 377, `video_interactions`/`video_engagement` 75, analyze clean.

hm21 (approved):
The session cache answers my last point — the budget is now paid once per video, and a timed-out count is no longer pinned. I pushed one takeover fix in `2d31b8210` and am approving on top of it.

**Finding.** The cache keyed on `getBulkVideoRevisions` returning `{}`, but the client maps *every* degraded path to `{}` — 404, non-200, malformed body, transport error. So a single funnelcake blip on a video's first engagement fetch was remembered as "this video has no other revisions" for the rest of the session, with the un-enriched count cached beside it. That is #6021 again in a narrower window. Repro before the fix: a resolver that returns `{}` once and the real revisions afterwards — the second `fetchEventLikers` still returned only `likerA` and never re-asked.

**Fix.** The client's two degraded shapes are now distinct. An empty map stays an answer, *including* the 404 an older deployment gives — it stays a 404 until that deployment changes, so the pre-#799 caching benefit you measured is intact. Everything else returns `null`, and a chunk whose lookup answered `null` takes its own cache entries back out, so the next fetch retries. `VideoRevisionsResolver` is now `Future<Map<String, List<String>>?>`.

Deliberately not extended: a failed lookup still lets `getLikeCount` cache the un-enriched count, same as before this PR. Suppressing that too would disable count caching for every addressable video until #799 deploys, which is a worse trade than the one it fixes.

Two new repository tests (retry after a lookup that could not answer; keep an answer of "no revisions") and three client tests. Both verified against the change they pin — disabling the eviction fails the first, evicting answers too fails the second.

`likes_repository` 219, `funnelcake_api_client` 379, `flutter analyze lib test integration_test` clean. Yours to review, tweak, or revert.

rabble (commented):
Re-checked against the current branch head (`7dadf0b0a`) — you've already addressed the two substantive points, and I verified the result rather than just reading it.

- **Chunking:** `_bulkVideoRevisionsBatchSize = 100` with `_chunkStrings` + `Future.wait`, matching the server cap. `test(likes): prove revision chunks past the first are merged` covers the case I was worried about (silent truncation past the first chunk).
- **Timeout:** `_resolveRevisionChunk` wraps the resolver in `.timeout(_revisionsResolverTimeout)` with an `onTimeout` that logs and returns empty, plus `onError` and a synchronous-throw guard. That closes the hang path — a slow funnelcake now degrades instead of stalling the list, which was the real risk.
- **Rename:** `_reqEidChunkSize` reads correctly for both callers now.

`flutter test` in `packages/likes_repository`: **197 passing.**

One observation, not a request: `_resolveRevisionTargets` is still awaited before the `Future.wait` on the reaction and a-tag queries, so "Liked by" latency is still funnelcake RTT + relay RTT in series. With the timeout in place that's now bounded rather than open-ended, which was the part that mattered. Worth revisiting only if the added latency shows up in practice.

rabble (commented):
**Correcting my previous comment — I understated point 1, and it is still outstanding.**

I wrote that the remaining serialization was "bounded rather than open-ended… worth revisiting only if the added latency shows up in practice." That was wrong on both counts.

@hm21 already measured it on this branch with a hung resolver:

| Call | Elapsed |
|---|---|
| `fetchEventLikers` (addressable) | 3011 ms |
| `getLikeCount` (addressable) | 3003 ms |

So it has shown up in practice, and "bounded" means a guaranteed 3s stall rather than a possible one — the budget having been widened 750ms → 3s makes it worse, not safer. The timeout caps the delay; it does not stop the baseline result being held hostage, which is exactly what the original request-changes asked for: *"Start the current-ID query immediately and ensure revision enrichment cannot hold the baseline result hostage."*

Confirmed still present at the current head: `_fetchResolvedLikersByEvent` does

```dart
final revisionToEventId = await _resolveRevisionTargets(...);   // blocks
...
final queryResults = await Future.wait([...]);                  // only then
```

Two knock-on effects @hm21 raised that I'd missed entirely, and which I think settle it:

1. `video_interactions_bloc.dart:220` puts the like count in one `Future.wait` with comment and repost counts, so a slow funnelcake delays all three.
2. The un-enriched count is written to the session cache, so a single timeout pins the lower count for the rest of the session — the failure is sticky, not transient.

**Status as I read it:** point 2 (chunking at the 100-id cap, with the merge test) is properly fixed. Point 1 is not. The shape that resolves it is emitting the baseline result from the current ids first and letting revision enrichment update it when it lands, rather than gating the first result on the resolver.

Withdrawing my earlier suggestion that this was ready to ship. The request-changes should stand until point 1 is addressed.

realmeylisdev:
Point 1 is addressed — nothing waits for the resolver any more. `43647fa97` + `14a479991`.

You were right that the timeout only made it a guaranteed stall rather than a possible one, and hm21's two knock-on effects are why bounding it was never going to be enough. So the wait is gone rather than resized:

- `_resolveRevisionTargets` is now synchronous. It reads revisions already resolved this session and returns; there is no `await` left between the fetch and its relay queries.
- A lookup that has not landed schedules a background pass that re-resolves the video and publishes the wider list on `watchRevisionEnrichedLikers`. `VideoEngagementBloc` re-emits the list, `VideoInteractionsBloc` raises the count from the same set.
- Only an edited video pays that second pass. A resolver answering "no other revisions" — every video, on today's 404ing deployment — costs nothing beyond its own lookup.
- The budget survives only as a backstop on that background wait, so a resolver that never completes cannot leak a continuation. It is 15s now for the opposite reason it was 3s: it no longer buys latency, so there is nothing to trade.

Pinned by `returns without waiting for the revision lookup` — a resolver that never answers, and the fetch has returned its list by the next turn of the event loop. Restoring the `await` turns it red immediately (`Actual: <null>`), which is the discriminator the old shape had no way to fail.

Knock-on 2 (the un-enriched count pinned in the session cache) was fixed by `515cbbd6d`, before your re-check — `revisionsPending` suppresses the cache write. Knock-on 1 is gone with the wait.

One caveat worth your attention: the "Liked by" list for a video seen for the first time this session is now emitted twice — current-id first, wider when the lookup lands. Every fetch after that reads the session cache and is wide on the first emission. That is the trade the two-phase shape makes, and it is why the bloc wiring is load-bearing rather than additive.

Not re-run on a device since those two commits. The property is structural and test-pinned, but I want the simulator pass redone before merge.

223 tests in `likes_repository`, 76 across the two blocs, `flutter analyze lib test integration_test` clean. Every new test verified against the mutation it pins — the table is in the PR description.

realmeylisdev:
Ran it end-to-end on the iOS Simulator against production data. Both phases are confirmed on a device; the caveat in my last comment is discharged.

**Setup.** funnelcake#799 is merged but `/api/videos/revisions/bulk` still 404s on `api.divine.video` — the live OpenAPI lists 67 paths and the route isn't one of them. So the run used a pass-through proxy: every other call forwarded to production untouched, that one route served with the merged contract. Real videos, real relay reactions, real profiles. Two production videos were paired as if one were the other's superseded revision, and every claim below was checked against a direct relay `REQ` for `kind:7 #e`.

**Point 1, from the device console** — proxy holding every lookup 14s, feed cold:

```
15:28:15  proxy   holding the lookup for 14.0s
15:28:21  app     fetch returned {08f2a9bf…: 30} revisionsPending=true  knownRevisionIds=0
15:28:29  proxy   REQ 1 id -> 1 with revisions
15:28:30  app     fetch returned {08f2a9bf…: 54} revisionsPending=false knownRevisionIds=1
15:28:30  app     publishing wider list for 08f2a9bf…: 30 -> 54
15:28:30  app     publishing wider list for 149c9e14…:  1 -> 28
```

The fetch answered ~6s from launch with 30 likers while the lookup was still held. Nothing waited for funnelcake. The wider list landed at +15s, when the lookup did.

**The bug itself.** Warm open of `cb7ddbb3…` — own 19 likers, paired revision 28, union 33 — rendered **33 tiles**, counted from the accessibility tree rather than by eye, including all **14 people reachable only via the superseded revision**. Two of them sit below the first screen and only appear after scrolling; their reactions were confirmed live (`+`, no kind-5 retraction). A second video: 36 tiles against a union of 36. The feed count showed 33 for the same video — the enriched number, not 19.

**Degradation.** With the proxy answering `{}` — the shape production gives today — 19 lookups returned no revisions, the feed and counts rendered normally, the list opened as before, and no second relay pass fired for any of them.

Two things worth flagging:

- The wider list is emitted on top of the narrow one, so a cold "Liked by" open on an edited video repaints once. Warm opens read the session cache and are wide on the first paint.
- #799 needs to actually deploy before any of this does anything in production. Until then every addressable video takes the 404 path above.

Local: `likes_repository` 223, `funnelcake_api_client` 379, engagement/interactions blocs and feed widgets 401, `flutter analyze lib test integration_test` clean. CI green on `b98f5f3f2`. Working tree clean — the proxy redirect and the temporary console instrumentation were local-only and are reverted.

realmeylisdev:
@rabble — your remaining point is addressed and ready for another look.

Point 1 (baseline held hostage) is fixed in `43647fa97` + `14a479991`: `_resolveRevisionTargets` is synchronous now, so there is no `await` between the fetch and its relay queries. Verified on a device, not just in tests — with the lookup held 14s the fetch answered at +6s with 30 likers and `revisionsPending=true`, then published `30 -> 54` when the lookup landed. Full console trace and the list-content check are in the comment above.

Point 2, the timeout, the rename and the missing log are all as you confirmed in your 01:08 review; nothing there changed.

One thing to weigh before you re-review, because it isn't the incremental change you last saw: the fix replaces "wait, but bounded" with "don't wait, publish when it lands". A cold "Liked by" open on an edited video now repaints once as the wider list arrives — warm opens read the session cache and are wide on first paint. That's a real tradeoff and it's your call, not mine.

@hm21 — flagging you too, since you approved on `2d31b8210` and the budget shape you questioned is gone rather than retuned. The 3s ceiling no longer sits on any fetch; it's a backstop on the background wait.

realmeylisdev:
Correcting the record on my own PR, and retracting part of my evidence.

`POST /api/videos/revisions/bulk` is now live on `api.divine.video` (the spec lists 70 paths and it answers the contract exactly). That finally let me measure production instead of reasoning about it, and the result undercuts how I justified this change.

- 41 of 373 sampled videos are genuinely edited, and 74 reactions across 10 of them do `e`-tag a superseded revision — the mechanism is real and common.
- But **all 211 reactions on those videos also carry an `a` tag**, so the coordinate filter already on `main` finds every one. Narrow vs wide gap: **zero on all 13 videos measured.** On today's data this PR changes no liker list.
- My "this is not a shrinking backlog of old clients" claim was wrong. Across 3103 reactions on 480 videos spanning 2017-01 → 2026-08, 6 are `e`-only (0.19%), and the rate runs 1.74% (Feb) → 0.25% (May) → 0.00% (Jun, Jul, Aug). The emitters are Divine's own clients, not the third parties I invoked.

Liker sets were cross-checked against funnelcake's coordinate-aware counts (28/27, 24/24, 16/17 — small deltas both directions from the five known count-vs-list causes), so this isn't an artefact of an incomplete relay view.

**I've removed the end-to-end run from the description.** It reported `19 → 33` and `30 → 54`, but since no production video exhibits this bug, that run manufactured the condition by pairing two unrelated production videos as if one were the other's superseded revision. It exercised the code path; it was not field evidence, and it shouldn't have read as such. The timing half survives — with the lookup held 14s the fetch still answered at +6s and published when it landed — because that doesn't depend on the pairing.

@rabble — point 1 is addressed in `43647fa97` + `14a479991`. `_resolveRevisionTargets` is synchronous now, so there's no `await` between the fetch and its relay queries, and `getLikeCount` returns without caching when a lookup is still pending, which closes the sticky-low-count knock-on you and @hm21 raised.

@hm21 @NotThatKindOfDrLiz — you both approved this as a fix for a live undercount. It's better described as insurance against a protocol-legal hole that is currently near-empty and closing, so it's worth a re-look before it goes in; that's not the trade you signed off on.

One thing the measurement turned up that's worth someone's time separately: every `e`-only reaction I found came from a Divine client, and `LikesRepository.likeEvent` emits no `a` tag when its `addressableId` is null. Closing that writer-side would remove the source rather than compensating for it at read time.

mobile/packages/likes_repository/lib/src/video_revisions_resolver.dart
```diff
@@ -0,0 +1,40 @@
+// ABOUTME: Port resolving a video event id to its sibling NIP-33 revisions.
+// ABOUTME: Injected so this package needs no API-client dependency.
+
+/// Resolves video event ids to every event id sharing their NIP-33 coordinate.
+///
+/// Editing an addressable video (kind 34235/34236) preserves the coordinate
+/// `kind:pubkey:d-tag` but mints a **new event id**. NIP-25 requires a kind 7
+/// reaction to carry an `e` tag and only *recommends* the `a` tag, so a
```
NotThatKindOfDrLiz, line 8: This is now the core scope call rather than a code correctness issue. The PR body says production measurement recovered zero current likers and reframes this as prospective insurance against a protocol-legal hole. Before merge, the reviewers who approved the earlier live-undercount premise should explicitly accept that trade at the current head, or choose to drop/close the read-side machinery in favor of the writer-side fix.

mobile/packages/likes_repository/lib/src/likes_repository.dart
```diff
@@ -81,24 +104,72 @@ class LikesRepository {
   /// - [queueOfflineAction]: Optional callback to queue actions when offline
   /// - [blockFilter]: Optional callback to hide blocked/muted users from
   ///   engagement lists ([fetchEventLikers])
+  /// - [revisionsResolver]: Optional resolver for a video's superseded event
+  ///   ids, so reactions stranded on a pre-edit revision stay visible (#6021)
```
NotThatKindOfDrLiz, line 108: This read-side resolver compensates for reactions that do not carry the coordinate. The author measurement says every observed e-only reaction came from Divine clients, and likeEvent still accepts a null addressableId that passes through to sendLike without an a tag. Please file/track the writer-side prevention issue before merge; otherwise the PR fixes the reader while leaving the confirmed source as an untracked follow-up.

NotThatKindOfDrLiz (changes requested):
Requesting changes on the merge decision, not on the mechanics of the current implementation. I pushed two small cleanup commits, 237759a77 and ef0f681b5, to trim stale/source-history comments, await the bloc subscription cancel, remove the test ignore marker, and correct the resolver fallback note.

The remaining blockers are decisions that should be made explicitly by the author/reviewers before merge: whether this read-side machinery is still worth shipping now that production measurement shows zero recovered likers today, and tracking the writer-side prevention fix so Divine stops emitting the e-only shape that created the observed cases. Once those are resolved, I do not have a code-quality objection to the current implementation.

mobile/packages/likes_repository/lib/src/likes_repository.dart
realmeylisdev: Filed as #6980, typed Bug, and linked back to this PR and to #6021.

I traced the chain end to end at the current head rather than restating the
observation, because the null does not originate in `likeEvent` — it arrives
there already null:

| # | Location | Behavior |
|---|---|---|
| 1 | `video_event.dart:1259` | `addressableId` is `null` when `addressableDTag` is null or empty |
| 2 | `video_event.dart:261` | `fromJson` falls back to `null` when neither the persisted field nor `rawTags['d']` is present |
| 3 | `feed_videos.dart:942` | that value goes into `VideoInteractionsBloc` unchecked |
| 4 | `video_interactions_bloc.dart:347` | forwarded to `toggleLike`; a null coordinate also drops the `k` tag |
| 5 | `likes_repository.dart:514` | `likeEvent` forwards `String? addressableId` with no guard |
| 6 | `nostr.dart:99` | `sendLike` adds the `a` tag only when non-empty |

So the guard you are asking for is real and belongs below the UI, not at
step 5 alone. The offline replay path inherits the same value
(`pending_action_service.dart:161`), which the issue covers too.

#6980 is independent of what happens to this PR — it stands whether the
read-side machinery ships or is dropped.

mobile/packages/likes_repository/lib/src/video_revisions_resolver.dart
realmeylisdev: Agreed that this is the call to make, and it is mine to put in front of the
people who approved the earlier premise rather than to settle here. I have
asked @rabble and @hm21 for an explicit re-accept at the current head in a
top-level comment, with the reframed trade stated plainly — including the part
that undercuts it, that the gap measured zero on all 13 videos.

Leaving this thread open until they answer. Your two cleanup commits
(`237759a77`, `ef0f681b5`) are in and I have no changes to make on top of them.

realmeylisdev:
@NotThatKindOfDrLiz — your second blocker is done: the writer-side prevention
fix is tracked as #6980, with the full chain from `VideoEvent.addressableId`
down to `Nostr.sendLike` and the offline replay path. It stands independently
of what happens to this PR. Your two cleanup commits are in and I have nothing
to add on top of them.

Your first blocker is a scope call I should not settle on my own PR, so:

@rabble @hm21 — you both approved this as a fix for a live undercount. My
2026-08-08 measurement says that is not what it is, and you should re-accept
the actual trade or decline it.

**What changed since you approved.** All 211 reactions on the affected videos
also carry an `a` tag, so the coordinate filter already on `main` finds every
one of them. Narrow vs wide gap: zero on all 13 videos measured. On today's
data this PR changes no liker list and no count. I also withdrew my "not a
shrinking backlog of old clients" claim — across 3103 reactions spanning
2017-01 → 2026-08, 6 are `e`-only (0.19%), running 1.74% (Feb) → 0.25% (May)
→ 0.00% (Jun, Jul, Aug), and every emitter was one of our own clients.

**The case for shipping it anyway.** An `e`-only reaction is protocol-legal.
#6980 closes our clients, but it cannot close third-party ones, so the hole
stays open to anything else writing Kind 7 against a Divine video. The
machinery is written, reviewed and green, and after `43647fa97` + `14a479991`
it is off the critical path — no fetch awaits the resolver, the baseline
answers from current ids as soon as relays do, and the wider list publishes as
a second emission if and when the lookup lands. The residual runtime cost is
one session-cached POST per video.

**The case against.** It is real complexity — revision resolver, session cache
with its own eviction semantics, 100-id chunking, timeout backstop, two-phase
emission — carried for a hole that is empty today and closing, and #6980
removes the only source we have actually observed. A cold "Liked by" open on
an edited video also repaints once as the wider list arrives.

**What I am asking for.** An explicit yes or no at the current head, not a
re-approval of the old premise. If either of you thinks the third-party
argument does not carry it, say so and I will close this in favour of #6980
alone — I would rather drop finished work than merge it on a justification
that the measurement no longer supports.

CI is green on the current head across every required check.

NotThatKindOfDrLiz (changes requested):
I pushed the remaining mechanical takeover fix in c68303be6: revision enrichment now stays additive at the repository publish boundary, so a background re-resolve that comes back with fewer likers cannot shrink the open "Liked by" sheet while VideoInteractionsBloc keeps the higher count. The regression test is `publishes nothing when the background pass finds fewer likers` in `likes_repository_test.dart`.

Validation from pre-push on c68303be6:
- analyzer clean
- generated files clean
- changed-file tests passed: `funnelcake_api_client_test.dart`, `likes_repository_test.dart`, `video_engagement_bloc_test.dart`, `video_interactions_bloc_test.dart`

I am still requesting changes on the merge decision, not on this mechanical fix. The remaining blocker is Rabble's explicit yes/no on whether to ship the read-side insurance at the current measured scope, with production showing zero recovered likers today and #6980 tracking the writer-side prevention.

realmeylisdev:
Closing this. The read-side fix already exists on the server, and it's one HTTP call.

**What decides it.** funnelcake's `GET /api/videos/{id}/likers` expands the `e`-tag arm across every revision sharing a video's NIP-33 coordinate. The bind-order contract spells it out at `crates/clickhouse/src/client.rs:3927` — *"e-tag value AGAIN, to expand it to its sibling revisions"*. It shipped in divinevideo/divine-funnelcake#799 on 2026-08-06, the same commit that added the `/api/videos/revisions/bulk` endpoint this PR consumes. I only ever wired up half of that PR.

Verified live today. 100 popular videos sampled, 10 with more than one revision, then `/likers` queried with both the current id and a superseded revision id:

```
current        superseded     cur_n sup_n  identical
bf62e9dded43   24d2a8c3fb5a      18    18  True
e65e6ed4b73b   f45399cee543      22    22  True
6018e1fdc2c3   dbe038644142      14    14  True
883a20af764a   30ab34a62b3c      23    23  True
ed8709d171f3   5c64651572c2      22    22  True
f34f980cb479   b9d37cf6d920      11    11  True
```

Mobile has no client for it — the only `/likers` string in `mobile/` is a route path in `router/routes/video_routes.dart:182`, and `VideoEngagementBloc` reads the relay-backed `LikesRepository`. So the recovery this PR builds in **+806/−79 lines of `lib/`**, a per-video funnelcake POST per session, and a batch-wide count-cache suppression is already available as a GET nobody calls.

**On the measurement.** I sampled 3103 reactions across 480 videos and found 6 carrying an `e` tag with no `a`; on the 13 edited videos I checked, 211 of 211 stranded reactions also carried an `a`, so main's coordinate filter already finds every one and this PR recovers nobody. I'd rather state that number's limits than lean on it: funnelcake has no `oldest` sort, so a video sample inherits the feed's time window and the per-month rates sit on thin slices. The three 0.00% months are too small to separate from the 0.19% baseline. "Small and not growing fast" is the honest read, not "gone" — and `divine-web/src/pages/VideoPage.tsx:307-315` is still publishing the shape today.

**Two more reasons not to ship this shape.**

The additive guard in `c68303be6` compares list *lengths*, not membership:

```dart
if (likers.length < baseline.length || _sameLikers(likers, baseline)) continue;
```

A baseline of `[A,B]` becoming `[A,C]` is equal-length, fails both conditions, publishes, and drops B out of a sheet the user is looking at. The enrichment pass is a fresh relay re-query, not a union — which is exactly the non-determinism the accompanying test comment describes. `_sameLikers` is also index-wise over a list from an unstable `List.sort`, so a pure reorder re-emits.

And count and list use different denominators, so no revision widening on either side can reconcile them. funnelcake's count is unfiltered `uniq(reaction_event_id)` (migration `000180`); its liker list is `GROUP BY pubkey` minus downvotes, banned, suspended, vanished and self-deleted. Measured on the same videos:

```
video          count   list  delta
bf62e9dded43      19     18      1
e65e6ed4b73b      23     23      0
6018e1fdc2c3      15     15      0
883a20af764a      24     24      0
ed8709d171f3      23     22      1
f34f980cb479      11     11      0
```

Two of six already diverge with zero stranding involved.

**What replaces it**

- **#6021 stays open**, re-scoped to point "Liked by" at funnelcake's revision-aware `/likers`. Details in [this comment](https://github.com/divinevideo/divine-mobile/issues/6021#issuecomment-5235662065). Note #6712 is only partly covered — reposts share the same server-side fix, comments have their own query path that #799 doesn't touch.
- **#6980** covers the write path. I've corrected it: four of its six citations pointed at lines that only exist on this branch, and step 6 named `nostr_sdk/lib/nostr.dart`, whose `sendLike` has no production caller — the real emitter is `nostr_client.dart:1396`. Scope widened to the comment-reaction path, the synthetic `VideoEvent` built from a `Comment` (which carries no `d` tag by construction), and offline rows already queued with a null coordinate.
- **#6980 is mobile-only.** divinevideo/divine-web#535 is open with no PR; divine-space had the same line and no issue at all, so I filed divinevideo/divine-space#11.
- **The REQ chunking is worth keeping** and is *not* yet extracted — `main` still builds a single unbounded `Filter(kinds:[reaction], e: eventIds)`, the same oversized-REQ hazard #5751 already fixed for the deletion probe. It lives at `likes_repository.dart` `_queryReactionsByEventIds` / `_queryEventsByEIds` / `_chunkStrings` on this branch, which I'm leaving in place. Still owed as its own PR.
- **The CI failure on this PR was never this PR's.** `relay_discovery_service_test.dart` corrupted a signature by overwriting its first byte with `00`, and BIP-340 aux randomness means the signature already starts with `00` about 1 run in 256 — measured 16 of 4000 — leaving the frame genuinely valid and the assertion red. Fixed in #6995.

@NotThatKindOfDrLiz — your scope objection was the right call. The two cleanup commits and the additive guard are what made the remaining gap visible, and the guard is what sent me looking at the publish boundary in the first place. Thanks for holding the line rather than waving it through.

mobile/packages/likes_repository/lib/src/video_revisions_resolver.dart
realmeylisdev: Answered, and the answer is your way: closed.

I put the scope question to @rabble and @hm21 as I said I would, but the deciding fact turned out not to need a vote. funnelcake's `GET /api/videos/{id}/likers` already expands the `e`-tag arm across sibling revisions server-side — `crates/clickhouse/src/client.rs:3927` — and it shipped in divine-funnelcake#799, the same commit that added the `/api/videos/revisions/bulk` endpoint this resolver consumes. I wired up half that PR and rebuilt the other half in Dart. Verified live: six separately-edited production videos return identical liker sets from a superseded id and the current id.

So this typedef, the session cache, the two-phase emission and the chunked revision lookup all go. #6021 stays open, re-scoped to point the list at that endpoint. Full reasoning in the closing comment.

mobile/packages/likes_repository/lib/src/likes_repository.dart
realmeylisdev: Your instinct that the writer side was the real gap was right, and it was broader than either of us had it.

#6980 is filed, but it needed fixing before it was actionable: four of its six citations pointed at lines that only exist on this branch, and step 6 named `nostr_sdk/lib/nostr.dart:99` — that `sendLike` has no production caller anywhere in the repo, only its own unit test. Every real like goes through `NostrClient.sendLike` at `nostr_client.dart:1396`. Corrected against `origin/main`.

Scope was also one write path of four. Now also covered: comment reactions (which derive the coordinate through a separate `_ownAddressableId` path), a synthetic `VideoEvent` built from a `Comment` — `comment_synthetic_video_event.dart` builds `rawTags` with no `d` tag at all, so the coordinate is null by construction rather than by parse gap — and offline rows already queued with a null coordinate, which replay after any write-path guard ships.

And #6980 is mobile-only, which the PR body glossed. divine-web still emits the shape from `VideoPage.tsx:307-315` (divinevideo/divine-web#535, open, no PR), and divine-space had the same line with no issue at all — filed as divinevideo/divine-space#11.

realmeylisdev:
Closed in favour of the server-side path — see the analysis above. Branch kept: the REQ chunking still needs extracting into its own PR.

realmeylisdev:
The REQ chunking is extracted: #6997.

It carries the two Kind 7 reaction `#e` filters (`getLikeCounts` and the addressable branch) onto the same chunker `#5751` already used for the Kind 5 deletion probe, plus the shared helper that folds the three chunk-and-flatten loops into one. Nothing revision-related came with it.

One correction to what I wrote above: I said `main` builds "a single unbounded" filter and implied a live hazard. Unbounded at the query boundary, yes — but in practice the largest caller (the curation sort, via `discoveryVideos`) is capped at 500 upstream by a memory guard in `video_event_service`, which is exactly the chunk size the deletion probe treats as the largest safe frame. So it is zero headroom resting on a bound in another service, not an observed overflow. #6997 says so plainly rather than overselling it.

This branch has nothing unique left on it now.

PR #6717, opened 2026-08-05
fix(explore): fall back for unknown tab slugs

### LINKED ISSUE #6716: fix(explore): fall back for unknown tab slugs
## Problem

`/explore/tab/:name` currently renders the route error screen when `:name` is not one of the compiled Explore tab slugs.

That is brittle for the upcoming server-configured featured tab work: a cold-start deep link can reach the router before remote config has loaded, and expired or disabled configured slugs should not strand the user on an error screen.

## User impact

Unknown Explore tab links should still land the user on Explore. The default tab is a better fallback than a route error for this path.

## Acceptance criteria

- Unknown `/explore/tab/:name` slugs render default Explore instead of `RouteErrorScreen`.
- Known slugs still select the intended tab.
- Existing underscore rejection remains unchanged.
- Router tests cover the fallback behavior.

## Technical notes

This is a preparatory cleanup for divinevideo/divine-mobile#6635, not the featured-tab implementation itself.

NotThatKindOfDrLiz:
## Summary

- Changes `/explore/tab/:name` so unknown tab slugs render default Explore instead of the route error screen.
- Keeps known slug validation in place for selecting canonical tabs.
- Updates the focused Explore tab route test to pin the fallback behavior.

## Why

This is prep for configurable featured Explore tabs. A server-configured slug may be unavailable on cold start, expired, disabled, or otherwise no longer configured. In those cases, Explore should fall back cleanly instead of stranding the user on an error screen.

Closes #6716
Refs #6635

## Verification

- `flutter test test/router/explore_tab_route_test.dart`
- `flutter analyze`
- pre-push hook: analyzer on changed files and `test/router/explore_tab_route_test.dart`

### DIFF
--- a/mobile/lib/router/routes/shell.dart
+++ b/mobile/lib/router/routes/shell.dart
@@ -4,12 +4,10 @@ import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
-import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/notifications/view/notifications_page.dart';
 import 'package:openvine/router/app_shell.dart';
 import 'package:openvine/router/navigator_keys.dart';
 import 'package:openvine/router/providers/page_context_provider.dart';
-import 'package:openvine/router/route_error_screen.dart';
 import 'package:openvine/router/routes/router_guards.dart';
 import 'package:openvine/screens/explore/explore_screen.dart';
 import 'package:openvine/screens/feed/home_feed_retap_cubit.dart';
@@ -88,20 +86,11 @@ List<RouteBase> shellRoutes() {
             ),
             GoRoute(
               path: ExploreScreen.pathTabSubpath,
-              pageBuilder: (ctx, st) {
+              pageBuilder: (_, st) {
                 final tabName = ExploreScreen.tabNameFromPathParameter(
                   st.pathParameters['name'],
                 );
-                if (tabName == null) {
-                  return _branchPage(
-                    st,
-                    RouteErrorScreen(message: ctx.l10n.routeUnknownPath),
-                  );
-                }
-                return _branchPage(
-                  st,
-                  ExploreScreen(initialTabName: tabName),
-                );
+                return _branchPage(st, ExploreScreen(initialTabName: tabName));
               },
             ),
             GoRoute(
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz:
## Review nudge — smallest piece of the featured-tab work, and it can land first

Two files (`mobile/lib/router/routes/shell.dart` plus its focused router test),
all CI green, mergeable, no review yet.

### Where this sits in the merge order

1. **This PR** → closes #6716, refs #6635. **No dependency on the backend work.**
   It only changes an unknown `/explore/tab/:name` slug from the route error
   screen to default Explore, and it is independently correct today — an unknown
   Explore link should land on Explore either way. Landing it first means the
   error path is already gone before any server-configured slug can exist.
2. **divinevideo/divine-funnelcake#771** (closes divinevideo/divine-funnelcake#769)
   — the hard prerequisite for the actual featured tab. Must merge *and deploy*.
3. **divinevideo/divine-admin-dashboard#11** — operator review queue, needed
   before a configured tab can be populated.
4. **#6635** here and **divinevideo/divine-web#527** — in parallel, once the
   endpoint is live.

### What this PR deliberately does not do

It does not implement the featured tab, fetch remote config, or add any tab. It
removes one routing behavior that would otherwise break a cold-start deep link
that arrives before remote config loads, or a slug that has since been disabled
or expired. Review scope is just the fallback and its test.

#6635 remains blocked on two backend contract questions tracked on
divinevideo/divine-funnelcake#769 — audience gating / cache semantics, and
dynamic slug lifecycle. Those do not affect this PR.

rabble (approved):

The behaviour change is right for what's coming. `tabNameFromPathParameter` already returned `null` for an unrecognised slug, and `ExploreScreen(initialTabName: null)` is the same path `/explore` itself takes, so an unknown slug now lands on the default tab through existing code rather than a new branch. Given the goal is server-configured featured tabs, a slug that's expired or not yet configured on cold start is an expected state, not a user error — rendering `RouteErrorScreen` for it would have been the wrong call. Deleting the branch instead of adding a flag is the right shape.

Worth knowing, not blocking: the URL and the UI now disagree for an unknown slug. `/explore/tab/garbage` renders the default tab while the location stays `garbage`, so a refresh repeats it. That's the correct tradeoff versus an error screen, and it doesn't poison the tab-persistence work from #6641 — `initialTabName: null` means the persisted index is used and the bad slug is never written back. Just calling it out so it isn't a surprise when someone deep-links a stale slug and wonders why the URL looks wrong.

One structural note on the test, which this PR inherits rather than introduces: `invalid tab slug falls back to the default tab` builds its own `GoRouter` with a hand-copied `pageBuilder` rather than exercising `shellRoutes()`, so it verifies a duplicate of the production logic. If someone broke the real route tomorrow, that test would still pass — the only thing actually guarding `shell.dart` is the source-grep test above it, and that one now asserts on the *absence* of a string (`RouteErrorScreen`), which is a weak signal that will pass for the wrong reasons if the route is refactored. Not asking you to fix it here; the pattern predates this change and the PR keeps both tests consistent with the new behaviour. Worth revisiting when the configurable-tabs work lands and this route gets more logic.

Ship it.

PR #6718, opened 2026-08-05
fix(classics): hide unknown archive import dates

### LINKED ISSUE #6545: fix(classics): archive imports without a published_at tag show this year as the creation date
## Summary

Some videos in the Classics feed show a creation date from the current year (e.g. "June 18, 2026") instead of their original Vine publication date, because those archive-import events ship without a `published_at` tag and the app silently falls back to the Nostr event `created_at` — which is the import timestamp.

## Environment

- App version: current `main`
- Device: any (data-driven, not device-specific)
- Network: any

## Steps to Reproduce

1. Open Explore → **Classics** (or the home feed in Classics mode).
2. Scroll a few pages — the affected events are sparse (~1% of rows, see Evidence).
3. Tap a video, then open the info sheet (tap the description / "more" button).
4. Look at the date line at the top of the sheet.

Reproduces reliably against the live API on rows such as event `fa45ae110d145a6a4095dd21a943e1c934a245e7619fb61276ff2a9521897b96` (`d` tag `hrILBu5hdBv`).

## Actual Result

The info sheet renders a current-year date for a classic Vine — for the event above, **June 18, 2026**, which is when the archive importer published it, not when the Vine was made.

Root cause chain:

1. Some `vine-archive-importer` events are published **without a `published_at` tag**. These are the partial-metadata imports — `author_name: "Unknown"`, `title: "Vine video"`, `loops: 0`, no `author_id` / `vine_user_id`. They do carry `platform=vine`, `vine_id`, and an `r` tag pointing at `https://vine.co/v/<id>`, so the app knows they are Vines.
2. [`video_stats.dart:261-263`](mobile/packages/models/lib/src/video_stats.dart#L261-L263) only picks `published_at` out of `event.tags`. No tag → `publishedAt == null`.
3. [`video_stats.dart:549-569`](mobile/packages/models/lib/src/video_stats.dart#L549-L569) then sets `VideoEvent.createdAt` / `timestamp` from the event `created_at` (the import time).
4. [`metadata_expanded_sheet.dart:147-153`](mobile/lib/widgets/video_feed_item/metadata/metadata_expanded_sheet.dart#L147-L153) formats `int.tryParse(video.publishedAt ?? '') ?? video.createdAt` and shows it unconditionally, so the import date is presented as the publication date.

Two knock-on effects from the same fallback:

- [`VideoEvent.isVintageRecoveredVine`](mobile/packages/models/lib/src/video_event.dart#L1283-L1290) compares the effective timestamp against the Vine shutdown date (2017-01-17). These events land after it, so they are classified as *not* vintage and the classic-Vine notice in [`comments_screen.dart:591`](mobile/lib/screens/comments/comments_screen.dart#L591) is suppressed.
- The Classics feed is the one feed that never runs `enrichVideosWithNostrTags` ([`classic_vines_provider.dart`](mobile/lib/providers/classic_vines_provider.dart) has no enrichment step, unlike `popular_videos_feed_provider.dart:17` / `new_videos_feed_provider.dart:13`), so a missing `published_at` can never be recovered from the relay while in that tab.

Secondary gap worth closing in the same change: the REST payload has a **top-level** `published_at` field alongside the tags, and `VideoStats.fromJson` ignores it entirely. Today it always mirrors `created_at` on the affected rows so it wouldn't help, but if the archive is backfilled server-side at the field level only, the client would still show the wrong date.

## Expected Result

A classic Vine should never claim to have been made this year.

Preferred handling when a `platform=vine` video has no usable original date:

- Do not present the import timestamp as the publication date — omit the date line, or label it as unknown / show the archive year instead.
- Keep `isVintageRecoveredVine` true for `platform=vine` events regardless of the missing timestamp, so the classic-Vine notice still shows.
- Read `published_at` from the top-level REST field as well as from the tags.

The upstream data is also wrong and should be fixed separately in the archive importer / funnelcake (these events should carry the real `published_at`, or be re-imported once the original date is recovered). The client fix is still needed so already-published events do not misrepresent themselves.

## Regression?

Unknown — the affected events appear to be a recent archive-import batch (`imported_at` values in April–June 2026) rather than a code change.

## Evidence

Live API response for the affected event (`GET https://api.divine.video/api/videos?classic=true&platform=vine&sort=loops&limit=100&offset=500`):

```json
{
  "id": "fa45ae110d145a6a4095dd21a943e1c934a245e7619fb61276ff2a9521897b96",
  "created_at": 1781809314,
  "published_at": 1781809314,
  "client": "vine-archive-importer",
  "platform": "vine",
  "author_name": "Unknown",
  "title": "Vine video",
  "loops": 362,
  "tags": [
    ["d", "hrILBu5hdBv"],
    ["vine_id", "hrILBu5hdBv"],
    ["platform", "vine"],
    ["client", "vine-archive-importer"],
    ["r", "https://vine.co/v/hrILBu5hdBv"],
    ["archived_at", "1775822361"],
    ["imported_at", "1781809314"],
    ["L", "archive.divine.video"],
    ["l", "vine-archive", "archive.divine.video"]
  ]
}
```

`1781809314` = 2026-06-18. There is no `published_at` tag; `archived_at` (2026-04-10) is the crawl date, so the event carries no original Vine date at all.

Scope, sampled across 16 offset pages of `?classic=true&platform=vine&sort=loops&limit=100` (1600 rows):

| created_at year | rows |
|---|---|
| 2013 | 133 |
| 2014 | 278 |
| 2015 | 471 |
| 2016 | 656 |
| 2017 | 46 |
| **2026** | **16** |

18 rows had no `published_at` tag; 16 of those render a post-shutdown date in the app (~1% of sampled rows). None of them had a top-level `published_at` that differed from `created_at`.

NotThatKindOfDrLiz:
## Summary
- hide the metadata date when an original Vine archive import has no usable `published_at` tag and only carries an import/post-shutdown timestamp
- replace the removed vintage-date heuristic with a server-controlled original Vine notice predicate for comments
- keep `published_at` tag precedence explicit without adding a speculative REST field fallback
- update regression tests and related badge-guide docs

## Verification
- `dart format --output=none --set-exit-if-changed mobile/packages/models/lib/src/video_event.dart mobile/packages/models/test/src/video_event_display_test.dart mobile/packages/models/test/src/video_stats_test.dart mobile/packages/models/test/src/video_event_to_json_test.dart mobile/lib/widgets/video_feed_item/metadata/metadata_expanded_sheet.dart mobile/test/screens/comments/comments_screen_test.dart mobile/test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart`
- `flutter test test/src/video_event_display_test.dart test/src/video_stats_test.dart test/src/video_event_to_json_test.dart` (from `mobile/packages/models`)
- `flutter test test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart test/screens/comments/comments_screen_test.dart` (from `mobile`)
- `flutter analyze lib test` (from `mobile`)

Closes #6545

### DIFF
--- a/mobile/docs/BADGE_GUIDE_FOR_DIVINE_WEB.md
+++ b/mobile/docs/BADGE_GUIDE_FOR_DIVINE_WEB.md
@@ -132,16 +132,16 @@ Do this:
 
 - require `platform: vine`
 
-There is also a narrower helper:
+There is also an original-date helper:
 
-- `video.isVintageRecoveredVine`
+- `video.hasUnknownOriginalDate`
 
 That means:
 
 - `isOriginalVine == true`
-- created/published time is before Vine shutdown
+- effective created/published time is at or after Vine shutdown
 
-Mobile uses `isVintageRecoveredVine` for the comments empty-state archive notice, not for the top-level badge choice.
+Mobile uses `isOriginalVine` for the comments empty-state archive notice. It uses `hasUnknownOriginalDate` only to avoid displaying an import timestamp as the original Vine date.
 
 ## Hosting Inputs
 
@@ -347,14 +347,13 @@ This is not part of the badge row, but it is easy to confuse with badge logic.
 
 In comments, mobile shows a `Classic Vine` archive notice when:
 
-- `video.isVintageRecoveredVine == true`
+- `video.isOriginalVine == true`
 
 That means:
 
 - `platform: vine`
-- plus pre-shutdown timestamp
 
-This is stricter than `isOriginalVine` and is specific to the comments empty state.
+This is specific to the comments empty state and independent of whether the original publish date is known.
 
 ## Short Version
 
--- a/mobile/docs/BADGE_GUIDE_FOR_DIVINE_WEB.md
+++ b/mobile/docs/BADGE_GUIDE_FOR_DIVINE_WEB.md
@@ -139,7 +139,8 @@ There is also an original-date helper:
 That means:
 
 - `isOriginalVine == true`
-- effective created/published time is at or after Vine shutdown
+- no usable `published_at` tag
+- and missing/zero `created_at`, or `created_at` at/after Vine shutdown
 
 Mobile uses `isOriginalVine` for the comments empty-state archive notice. It uses `hasUnknownOriginalDate` only to avoid displaying an import timestamp as the original Vine date.
 
--- a/mobile/lib/screens/comments/comments_screen.dart
+++ b/mobile/lib/screens/comments/comments_screen.dart
@@ -588,7 +588,7 @@ class _CommentsScreenBody extends StatelessWidget {
           ],
           child: SizedBox(
             child: CommentsList(
-              showClassicVineNotice: videoEvent.isVintageRecoveredVine,
+              showClassicVineNotice: videoEvent.isOriginalVine,
               scrollController: sheetScrollController,
               showVideoReplies: showVideoReplies,
             ),
--- a/mobile/lib/widgets/video_feed_item/metadata/metadata_expanded_sheet.dart
+++ b/mobile/lib/widgets/video_feed_item/metadata/metadata_expanded_sheet.dart
@@ -198,7 +198,7 @@ class _OverviewSection extends StatelessWidget {
       child: Column(
         crossAxisAlignment: CrossAxisAlignment.start,
         children: [
-          if (formattedDate != null && semanticDate != null)
+          if (formattedDate != null)
             Semantics(
               label: semanticDate,
               child: ExcludeSemantics(
@@ -219,13 +219,13 @@ class _OverviewSection extends StatelessWidget {
               spacing: 8,
               children: titleCluster,
             ),
-          // ⚠ LOAD-BEARING 12 px (NOT 16). The first chip row's 4 px
-          // invisible top padding stacks on this to produce the visible
-          // 16 px gap matching the date → title-cluster gap. Sibling
-          // of the bottom-padding tweak above; see `MetadataTagsSection`
-          // for the full dependency map.
+          // ⚠ LOAD-BEARING 12 px (NOT 16). When content precedes the chip row,
+          // the first chip row's 4 px invisible top padding stacks on this to
+          // produce a visible 16 px gap. Sibling of the bottom-padding tweak
+          // above; see `MetadataTagsSection` for the full dependency map.
           if (hasTags) ...[
-            const SizedBox(height: 12),
+            if (formattedDate != null || titleCluster.isNotEmpty)
+              const SizedBox(height: 12),
             MetadataTagsSection(video: video),
           ],
         ],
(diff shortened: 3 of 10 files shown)


hm21 (approved):
Approving — but `hasUnknownOriginalDate` needs one more pass, it hides dates that are known.

A vine carrying an explicit `published_at` tag has a known date wherever it lands, and Vine's last day *is* 2017-01-17. `effectiveCreatedAt >= _vineShutdownAtUtcSeconds` (1484611200 = 2017-01-17 00:00 UTC) therefore swallows the farewell vines.

Sampled 1600 rows of `?classic=true&platform=vine&sort=loops`:

| | rows |
|---|---|
| no `published_at` tag → hidden (intended) | 42 |
| **explicit `published_at` tag ≥ the constant → hidden (regression)** | **13** |

Example `2f09f97b71372ed1c63b4c99b91beda8e49399d6c2840b2273a5f6743d088b9f` — `published_at` 1484627482 = 2017-01-17 04:31 UTC, 1739 loops, *"It's been a amazing 4 years!! thank you for everything and a special thank you to vine."* Renders "January 17, 2017" on `main`, renders nothing after this PR.

`rawTags` already collects every tag (`video_stats.dart:216`), so the signal is direct rather than inferred — same move this PR already made for the comments notice:

```dart
bool get hasUnknownOriginalDate {
  if (!isOriginalVine) return false;
  return !rawTags.containsKey('published_at');
}
```

Dropping the old `> 0` guard also means a row with `createdAt == 0` now renders "January 1, 1970" instead of hiding.

Separate, non-blocking: I'd drop the REST `published_at` fallback (`video_stats.dart:300`). By its own comment it mirrors `created_at` today, so it adds nothing now, and it takes precedence over relay-recovered values in `video_nostr_enrichment.dart:210` and `profile_video_merge.dart:38`. I checked live data and nothing is broken by it today — but the server change that fixes the tag can ship the field at the same time, so there's no reason to carry it ahead of that.

Nit: unrelated formatter churn in the `sheet controller disposal` test and two `MetadataStatsRow` blocks.

NotThatKindOfDrLiz:
Takeover update pushed in `c7d80d818`.

What changed:
- `hasUnknownOriginalDate` now treats a usable raw `published_at` tag as a known original date, including 2017-01-17 farewell Vines.
- Missing/unusable `published_at` now falls back to the intended import-timestamp heuristic, including `createdAt <= 0` so January 1, 1970 is hidden.
- Removed the speculative REST `published_at` fallback and `_parseNullableInt` helper.
- Fixed the metadata sheet chip-row leading gap when the date/title cluster is hidden.
- Added model and widget regressions for farewell-day tags, zero timestamps, tag precedence, and hidden import dates.

Verification:
- `dart format --output=none --set-exit-if-changed ...` on edited Dart files
- `flutter test test/src/video_event_display_test.dart test/src/video_stats_test.dart test/src/video_event_to_json_test.dart` from `mobile/packages/models`
- `flutter test test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart test/screens/comments/comments_screen_test.dart` from `mobile`
- `flutter analyze lib test` from `mobile`
- GitHub checks are green on `c7d80d818`

Issue #6720, opened 2026-08-05
feat(relays): model read and write markers end to end


NotThatKindOfDrLiz:
## Problem

NIP-65 relay-list tags can mark relays as read-only or write-only, but Divine currently collapses configured relays to plain URLs. RelayDiscoveryService parses read/write flags, but RelayManager storage, Settings, and relay-list publishing do not model them as first-class state.

## Impact

The #6648 fix preserves markers from the last discovered relay list where possible, but new relay edits still cannot express marker intent from Divine itself. Users who manage relay roles in another client can lose that distinction on future edits.

## Acceptance criteria

- Relay storage can represent read/write marker state for each configured relay.
- Settings can show and edit read/write intent clearly.
- Kind 10002 publishing emits marker tags from first-class state, not a preservation workaround.
- Discovery/import keeps marker state when ingesting kind 10002.
- Tests cover marker round-trips through storage, Settings actions, and publish tags.

## Technical notes

Follow-up from #6648. This should be separate from the publishing bug fix because it changes relay settings UX and storage shape.

Issue #6719, opened 2026-08-05
chore(relays): track relay provenance in configured relay state


NotThatKindOfDrLiz:
## Problem

RelayManager currently stores configured relays as a flat list of URLs. It accepts user, automatic, discovered, fallback, and default relay additions, but only keeps source information long enough to update the local user-removal ledger. After that, Settings and any account-level publish path cannot distinguish user intent from automatic reachability relays.

## Impact

Publishing the visible configured list is the right short-term behavior for #6648, but provenance would let future relay-list publishes reflect deliberate user choices more precisely and would give relay re-add bugs stronger signals.

## Acceptance criteria

- Relay configuration storage records durable provenance for user/default/discovered/fallback/automatic additions.
- User removals continue to suppress automatic re-adds.
- Settings can still show the complete effective relay set.
- Account-level relay-list publishing can choose the intended subset without guessing.
- Existing #6475 behavior remains covered by tests.

## Technical notes

Follow-up from #6648. Keep this separate from the kind 10002 publish fix so the bug fix remains reviewable.

PR #6721, opened 2026-08-05
fix(relays): publish settings edits as kind 10002

### LINKED ISSUE #6648: fix(relays): publish relay list edits as kind 10002
## Problem

Relay Settings changes currently update local configured relay storage, and #6475 preserves local deletion intent with a device-local removed-relay ledger. But Settings add/remove/restore actions do not publish an updated NIP-65 kind 10002 relay list.

That means a relay deleted on one install can still be advertised by the users stale published relay list. Other Nostr clients, a fresh Divine install, or another device can rediscover and reconnect to that stale list because the local tombstone ledger does not travel with the account.

## Expected behavior

When the user edits relays in Settings, Divine should publish the current configured relay list as kind 10002 for the signed-in identity, using the existing relay publishing path and preserving local relay fallback behavior.

## Scope notes

- Trigger publishing from Settings add/remove/restore mutations, likely at the RelaySettingsCubit or service boundary after local mutation succeeds.
- Publish the full configured relay list, not only the Divine primary relay bootstrap value.
- Clear or invalidate the account-scoped relay_discovery_<npub> cache after a successful local mutation so subsequent discovery does not replay stale data.
- Keep #6475 behavior intact: automatic discovery/fallback relays must not override user-deleted relays locally.
- Add regression coverage for local mutation -> kind 10002 publish and cache invalidation.

## Related

Follow-up from #6475 and #6471. Adjacent to #5065 because both involve distinguishing local defaults/fallbacks from the users published relay preferences.

NotThatKindOfDrLiz:
## Summary

- publish successful relay Settings edits as NIP-65 kind 10002 events
- preserve discovered read/write markers when rebuilding relay-list tags
- keep local edits successful when publish fails, mark the relay list dirty, and retry on next active session
- clear relay discovery cache after accepted indexer publish

Closes #6648

## Tests

- flutter test test/repositories/relay_list_repository_test.dart test/blocs/relay_settings/relay_settings_cubit_test.dart test/l10n/arb_consistency_test.dart
- flutter analyze

## Follow-ups

- #6719 tracks durable relay provenance.
- #6720 tracks first-class read/write marker support.

### DIFF
--- a/mobile/lib/blocs/relay_settings/relay_settings_cubit.dart
+++ b/mobile/lib/blocs/relay_settings/relay_settings_cubit.dart
@@ -5,6 +5,7 @@
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:nostr_client/nostr_client.dart';
 import 'package:openvine/blocs/relay_settings/relay_settings_state.dart';
+import 'package:openvine/repositories/relay_list_repository.dart';
 import 'package:openvine/services/relay_capability_service.dart';
 import 'package:openvine/services/video_event_service.dart';
 import 'package:openvine/utils/relay_url_utils.dart';
@@ -31,14 +32,17 @@ import 'package:openvine/utils/relay_url_utils.dart';
 class RelaySettingsCubit extends Cubit<RelaySettingsState> {
   RelaySettingsCubit({
     required NostrClient nostrClient,
+    required RelayListRepository relayListRepository,
     required RelayCapabilityService relayCapabilityService,
     required VideoEventService videoEventService,
   }) : _nostrClient = nostrClient,
+       _relayListRepository = relayListRepository,
        _relayCapabilityService = relayCapabilityService,
        _videoEventService = videoEventService,
        super(const RelaySettingsState());
 
   final NostrClient _nostrClient;
+  final RelayListRepository _relayListRepository;
   final RelayCapabilityService _relayCapabilityService;
   final VideoEventService _videoEventService;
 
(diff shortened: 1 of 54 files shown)


hm21 (approved):
Approving with five commits pushed as reviewer fixes — please review, tweak, or revert any of them before merging.

**`d11ebeba58` — blocker: CI was red.** `RelaySettingsScreen.build` now watches `relayListRepositoryProvider`, which pulls `authServiceProvider` → `userDataCleanupService` → `sharedPreferencesProvider` into the screen's provider graph. `relay_settings_screen_layout_test.dart` overrides none of those, so all 11 tests failed with `UnimplementedError: sharedPreferencesProvider must be overridden in ProviderScope`. Fixed by overriding the repository provider with a fake rather than wiring up SharedPreferences and AuthService, so that suite stays about layout.

**`3280b9ce67`** — `addedConnectionPendingLocalOnly` / `restoredConnectionPendingLocalOnly` replaced `relaySettingsFailedToConnectCheck` with the publish-pending copy, so the strictly worse case told the user less than plain connection-pending: they lost the only part they can act on. Now shows both sentences, with a widget test.

**`90341cc0ce`** — in `removeRelay`, the `!success` branch and the success path ran identical publish-and-map code; the branch only decided the still-configured guard. Collapsed.

**`8c4e32f221`** — `retryDirtyPublish` reported a clean relay list as `skippedNoKeys`, whose `localOnly` is `true`. The bridge reads `localOnly` as "did not reach an indexer" and warns, so any caller that skips its own `isDirty` pre-check logs a failure for a publish never attempted. Added a distinct `skippedNotDirty`.

**`2cdb87d817`** — Settings blocks on the publish before it can show any result, and `publishEventAwaitOk` defaults to a 15s OK wait stacked on the 10s signer timeout: up to 25s of silence after the sheet closes, with no progress affordance. Capped this call at 5s.

Two of these are judgment calls, so push back if you disagree:

- The 5s cap trades a slightly higher false-negative rate for a much shorter silent wait. A relay that hasn't answered in 5s is unlikely to answer at 14s, and a false negative only costs a dirty flag the bridge retries next session — but the number is mine, not measured.
- The combined snackbar concatenates two existing ARB strings rather than adding a new key across 23 locales. It reads as four sentences, which is long; a purpose-written key would be better copy if you think it's worth the translation pass.

Validation: `flutter analyze lib test integration_test` clean, `dart format` clean, and the four affected suites green locally (44 tests). All 14 CI checks pass on `2cdb87d817`.

Not touched: markers come from in-memory `authService.userRelays`, which is empty before discovery lands and after an account switch, so a Settings edit in that window republishes every relay as read+write and drops the user's read/write markers. #6719 / #6720 already track it — flagging only because the window is wider than "before first discovery."

PR #6722, opened 2026-08-05
fix(models): preserve addressable d-tag state

### LINKED ISSUE #6713: fix(models): preserve missing d-tag state for addressable videos
## Summary

`VideoEvent.fromNostrEvent` currently collapses a missing or empty `d` tag into `vineId = event.id`. That makes downstream code unable to distinguish a real addressable `d` tag from the fallback identifier.

This came up while triaging #6547. It is not the confirmed customer failure there, but it is an adjacent deletion correctness gap: `_addressableDeletionTarget` returns null when `video.vineId == video.id`, so a kind-34236 video parsed without a real `d` tag gets an e-tag-only delete instead of an addressable `a` coordinate.

## Evidence

- `mobile/packages/models/lib/src/video_event.dart`: when `vineId == null || vineId.isEmpty`, it assigns `vineId = event.id`.
- `mobile/lib/services/content_deletion_service.dart`: `_addressableDeletionTarget` returns null when `vineId == null`, empty, or equal to `video.id`.
- For a protocol-invalid kind-34236 event missing `d`, the app cannot tell whether there is no addressable target or whether `vineId` is just the fallback.

## Desired behavior

Preserve whether a real `d` tag was present so delete code can make an explicit decision instead of inferring from `vineId == video.id`.

Possible fix directions:

- Add explicit model state such as `hasAddressableDTag` / `addressableDTag` while keeping existing stable id behavior for UI dedupe, or
- Split fallback display/cache id from the parsed NIP-33 addressable identifier.

## Verification

Add regression coverage for:

1. A kind-34236 event with a non-empty `d` tag exposes an addressable delete target.
2. A kind-34236 event with no `d` tag preserves that missing state instead of presenting `event.id` as a real addressable `d` tag.
3. Content deletion builds an `a` tag only from a real addressable coordinate.

Do not fold this into #6547. That tracker remains blocked on the relay-side fix in divinevideo/divine-funnelcake#782.

NotThatKindOfDrLiz:
## Summary

- add explicit `VideoEvent.addressableDTag` / `hasAddressableDTag` state for the real NIP-33 `d` tag
- keep `vineId` identity fallback behavior while moving protocol coordinate builders to the explicit d-tag state
- stop delete/view/audio/repost/community-label paths from fabricating `a` coordinates from `vine_id` or event-id fallbacks
- preserve older cached JSON by hydrating `addressableDTag` from persisted `rawTags['d']`

## Verification

- `flutter test test/src/video_event_parsing_test.dart test/src/video_stats_test.dart test/src/video_event_to_json_test.dart test/src/audio_event_test.dart` from `mobile/packages/models`
- `flutter test test/src/video_event_addressable_id_test.dart` from `mobile/packages/models`
- `flutter test test/src/profile_video_merge_test.dart` from `mobile/packages/videos_repository`
- `flutter test test/blocs/profile_feed/profile_feed_enrichment_merge_test.dart test/utils/video_nostr_enrichment_views_test.dart test/services/content_deletion_service_test.dart test/services/view_event_publisher_test.dart` from `mobile`
- `flutter test test/blocs/profile_reposted_videos/profile_reposted_videos_bloc_test.dart` from `mobile`
- `flutter test test/repositories/community_content_label_repository_test.dart test/services/effective_content_labels_test.dart` from `mobile`

Closes #6713

### DIFF
--- a/mobile/lib/blocs/profile_feed/profile_feed_enrichment_merge.dart
+++ b/mobile/lib/blocs/profile_feed/profile_feed_enrichment_merge.dart
@@ -68,6 +68,7 @@ VideoEvent _mergeEnrichmentIntoCurrent(
         ? current.hashtags
         : enriched.hashtags,
     vineId: current.vineId ?? enriched.vineId,
+    addressableDTag: current.addressableDTag ?? enriched.addressableDTag,
     group: current.group ?? enriched.group,
     altText: current.altText ?? enriched.altText,
     blurhash: current.blurhash ?? enriched.blurhash,
--- a/mobile/lib/blocs/profile_reposted_videos/profile_reposted_videos_bloc.dart
+++ b/mobile/lib/blocs/profile_reposted_videos/profile_reposted_videos_bloc.dart
@@ -11,8 +11,6 @@ import 'package:equatable/equatable.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:models/models.dart' hide LogCategory;
-import 'package:nostr_sdk/aid.dart';
-import 'package:nostr_sdk/event_kind.dart';
 import 'package:openvine/blocs/profile_shared/profile_tab_page_size.dart';
 import 'package:openvine/blocs/profile_shared/profile_tab_sync_completion.dart';
 import 'package:openvine/blocs/profile_shared/profile_video_list_snapshot.dart';
@@ -586,13 +584,8 @@ class ProfileRepostedVideosBloc
   /// Compute the addressable ID for a video event.
   ///
   /// Format: `kind:pubkey:d-tag`
-  /// Returns null if the video doesn't have a d-tag (vineId).
+  /// Returns null if the video doesn't have a real d-tag.
   String? _computeAddressableId(VideoEvent video) {
-    if (video.vineId == null) return null;
-    return AId(
-      kind: EventKind.videoVertical,
-      pubkey: video.pubkey,
-      dTag: video.vineId!,
-    ).toAString();
+    return video.addressableId;
   }
 }
--- a/mobile/lib/repositories/community_content_label_repository.dart
+++ b/mobile/lib/repositories/community_content_label_repository.dart
@@ -253,7 +253,7 @@ class CommunityContentLabelRepository {
   }
 
   String? _addressableTargetFor(VideoEvent video) {
-    final dTag = video.vineId;
+    final dTag = video.addressableDTag;
     if (!video.isAddressableShareKind || dTag == null || dTag.isEmpty) {
       return null;
     }
--- a/mobile/lib/services/content_deletion_service.dart
+++ b/mobile/lib/services/content_deletion_service.dart
@@ -542,19 +542,9 @@ class ContentDeletionService {
   }
 
   String? _addressableDeletionTarget(VideoEvent video) {
-    final dTag = video.rawTags['d'];
-    final stableDTag = dTag != null && dTag.isNotEmpty
-        ? dTag
-        : video.vineId != null &&
-              video.vineId!.isNotEmpty &&
-              video.vineId != video.id
-        ? video.vineId
-        : null;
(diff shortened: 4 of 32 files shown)


rabble (approved):

This is the right fix at the right layer. `vineId` was carrying two incompatible jobs — "stable local identity, fall back to the event id" and "the NIP-33 `d` tag" — and every protocol coordinate builder was quietly getting the first when it needed the second. Splitting them at the model and leaving `vineId`'s fallback semantics untouched means the identity behaviour callers depend on doesn't move while `addressableId` stops inventing coordinates that never resolved on a relay.

The details that make it hold up:

- In `fromNostrEvent`, `addressableDTag` is only assigned in the `case 'd'` arm — `vine_id` writes `vineId` alone. That's the whole point of the change and it's implemented in exactly one place.
- `VideoStats.toVideoEvent` keeps `vineId: normalizedDTag` (which is `dTag.isNotEmpty ? dTag : id`) while `addressableDTag: dTag.isNotEmpty ? dTag : null`. I checked that "normalized" is only the id-fallback and not some string transform, so the coordinate built from the raw `d` is the same string the relay stores — no silent coordinate drift for REST-hydrated rows.
- `fromJson` hydrating from `rawTags['d']` when `addressableDTag` is absent is the correct migration for already-cached JSON, and the persisted value takes precedence so a re-serialised row doesn't get re-derived.
- `clearAddressableDTag` on `copyWith` rather than relying on `??`, and all four merge paths (`profile_video_merge`, `profile_feed_enrichment_merge`, `video_nostr_enrichment`, repost bloc) threading the new field — easy to miss one of those and none were missed.
- `VideoEvent` equality is `id`-based, so the new field doesn't perturb dedup or set membership anywhere.

I walked every `.addressableId` consumer in `lib/`. All of them already treated it as nullable, so nothing breaks on type — the change only makes null more common. The one place that looked like a null-equality trap, `audio_reuse_consent_resolver`'s `.where((video) => video.addressableId == sourceAddress)`, is guarded by an early `sourceAddress == null || isEmpty` return, so it can't degenerate into "every d-less video matches".

**The one thing I'd confirm before this ships**, because it's the only consequence I can't verify from inside this repo:

`ViewEventPublisher` now returns `false` and logs instead of publishing when `addressableId` is null. Previously it published `34236:<pubkey>:<vineId ?? id>`. If the view-count aggregation on the backend keyed off that same `vineId ?? id` fallback, then views for `d`-less videos were being counted *consistently* under a synthetic coordinate, and this change silently drops them to zero. If the aggregator keys off the real coordinate — which is what I'd assume, and what makes the old behaviour a bug — then those views were already being written somewhere nothing reads, and skipping is strictly better. Worth one check with whoever owns the view pipeline; the difference is invisible from the client and shows up as a metrics cliff for whatever fraction of the corpus lacks a `d` tag.

Two ripples worth knowing about, both of which I think are the correct answer rather than regressions:

- Videos carrying `vine_id` but no `d` now produce no clip-reuse credit (`video_clip_import_service` feeds `sourceAddressableId` straight from `addressableId`, which is what #6725 consumes) and fail closed on audio-reuse consent. Failing closed on consent is right. The lost credit is the honest outcome given the coordinate was never resolvable, but if that population is large it's a visible feature gap rather than a silent one.
- Comments on those videos lose `rootAddressableId` and will thread by event id instead. Same reasoning applies.

Ship it — but get the view-pipeline answer.

NotThatKindOfDrLiz:
@rabble @dcadenas can you confirm the view-pipeline side before this ships?

PR #6722 stops mobile from publishing kind 22236 view events for videos without a real NIP-33 `d` tag, instead of sending synthetic `34236:<pubkey>:<vineId ?? id>` coordinates.

Does Funnelcake view aggregation count only real `a` coordinates / non-empty `video_d_tag`, or were synthetic coordinates intentionally counted for d-less videos?

From the code search I did, the expected answer looks like “real coordinates only”: `ViewEventData::from_event` requires an `a` tag, parses `video_d_tag` from that coordinate, and the logical stats view filters out empty `video_d_tag`. I want to confirm that this is the intended backend contract before merging.

PR #6723, opened 2026-08-05
fix(feed): narrow dead-media prune

### LINKED ISSUE #6251: fix(feed): narrow 404 prune to moderation-confirmed unavailable media

NotThatKindOfDrLiz:
## Summary
- gate persistent dead-media pruning on 404 plus moderation-confirmed blocked/quarantined status
- keep age-restricted or unknown moderation states recoverable instead of marking them broken
- scope BrokenVideoTracker persistence by anonymous/current pubkey and reattach the scoped tracker on auth changes

## Tests
- flutter test test/services/dead_media_feed_guard_test.dart test/services/video_event_service_broken_filter_test.dart test/providers/broken_video_tracker_identity_test.dart test/blocs/fullscreen_feed/fullscreen_feed_bloc_test.dart test/screens/feed/feed_auto_advance_error_listener_test.dart
- push hook: analyzer, generated files, changed-file tests

Fixes #6251

### DIFF
--- a/mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
+++ b/mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
@@ -40,8 +40,10 @@ typedef OnVideoConfirmedUnavailable = void Function(String videoId);
 /// Confirms whether a player-reported unavailable video is safe to prune.
 ///
 /// Production passes [DeadMediaFeedGuard.isConfirmedUnavailable], which first
-/// HEAD-confirms a 404 and then requires moderation to say the blob is blocked
-/// or quarantined. Tests may inject a smaller predicate.
+/// HEAD-confirms a 404 and then requires moderation to say the blob is
+/// `blocked`. Quarantined and age-restricted blobs also 404 but are reversible,
+/// so they keep their error tile instead of being pruned. Tests may inject a
+/// smaller predicate.
 typedef ConfirmVideoUnavailable =
     Future<bool> Function({required String? videoUrl, String? explicitSha256});
 
--- a/mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
+++ b/mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
@@ -15,7 +15,6 @@ import 'package:media_cache/media_cache.dart';
 import 'package:models/models.dart' hide LogCategory;
 import 'package:openvine/extensions/video_event_extensions.dart';
 import 'package:openvine/services/dead_media_feed_guard.dart';
-import 'package:openvine/services/media_availability_checker.dart';
 import 'package:openvine/utils/video_identity.dart';
 import 'package:unified_logger/unified_logger.dart';
 
@@ -617,13 +616,12 @@ class FullscreenFeedBloc
     );
   }
 
+  /// Fails closed: without an injected confirmer there is no moderation
+  /// verdict, and a bare 404 is not grounds for the persistent prune (#6251).
   static Future<bool> _defaultConfirmVideoUnavailable({
     required String? videoUrl,
     String? explicitSha256,
-  }) async {
-    if (videoUrl == null || videoUrl.isEmpty) return false;
-    return const MediaAvailabilityChecker().isConfirmedMissing(videoUrl);
-  }
+  }) async => false;
 
   /// Handle UI acknowledgement of a pending skip signal.
   void _onSkipAcknowledged(
--- a/mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
+++ b/mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
@@ -14,6 +14,7 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:media_cache/media_cache.dart';
 import 'package:models/models.dart' hide LogCategory;
 import 'package:openvine/extensions/video_event_extensions.dart';
+import 'package:openvine/services/dead_media_feed_guard.dart';
 import 'package:openvine/services/media_availability_checker.dart';
 import 'package:openvine/utils/video_identity.dart';
 import 'package:unified_logger/unified_logger.dart';
@@ -29,13 +30,21 @@ part 'fullscreen_feed_state.dart';
 typedef OnRemoveVideo = void Function(String videoId);
 
 /// Callback invoked by [FullscreenFeedBloc] when a video has been confirmed
-/// unavailable (a hard 404 verified via [MediaAvailabilityChecker]).
+/// unavailable by the shared prune guard.
 ///
 /// Unlike [OnRemoveVideo] — which only drops the id from in-memory caches for
 /// the current session — this persists the id so the video stays filtered out
 /// of every list surface across app restarts. A callback keeps the BLoC free
 /// of direct dependencies on concrete services.
 typedef OnVideoConfirmedUnavailable = void Function(String videoId);
+
+/// Confirms whether a player-reported unavailable video is safe to prune.
+///
+/// Production passes [DeadMediaFeedGuard.isConfirmedUnavailable], which first
+/// HEAD-confirms a 404 and then requires moderation to say the blob is blocked
+/// or quarantined. Tests may inject a smaller predicate.
+typedef ConfirmVideoUnavailable =
+    Future<bool> Function({required String? videoUrl, String? explicitSha256});
 
 /// Returns `true` when [pubkey]'s content must be hidden (blocked / muted /
 /// blocked-us / muted-us). Injected so the BLoC stays free of Riverpod and
@@ -45,8 +54,8 @@ typedef OnVideoConfirmedUnavailable = void Function(String videoId);
 /// across mutations.
 typedef BlockAuthorFilter = bool Function(String pubkey);
 
-/// Returns `true` when [videoId] has been confirmed unavailable (a persisted
-/// hard 404). Injected so the BLoC stays free of concrete service
+/// Returns `true` when [videoId] has been confirmed unavailable and persisted.
+/// Injected so the BLoC stays free of concrete service
 /// dependencies — the launching `ConsumerWidget` resolves it from
 /// `BrokenVideoTracker.isVideoBroken`. The bound method reads live tracker
 /// state on every call, so newly-confirmed-missing videos are filtered from
@@ -97,7 +106,7 @@ class FullscreenFeedBloc
     BlossomAuthService? blossomAuthService,
     OnRemoveVideo? onRemoveVideo,
     OnVideoConfirmedUnavailable? onVideoConfirmedUnavailable,
-    MediaAvailabilityChecker? availabilityChecker,
+    ConfirmVideoUnavailable? confirmVideoUnavailable,
     BlockAuthorFilter? blockFilter,
     UnavailableVideoFilter? unavailableFilter,
     FeedTuningRepository? feedTuningRepository,
@@ -111,8 +120,8 @@ class FullscreenFeedBloc
        _blossomAuthService = blossomAuthService,
        _onRemoveVideo = onRemoveVideo,
        _onVideoConfirmedUnavailable = onVideoConfirmedUnavailable,
-       _availabilityChecker =
-           availabilityChecker ?? const MediaAvailabilityChecker(),
+       _confirmVideoUnavailable =
+           confirmVideoUnavailable ?? _defaultConfirmVideoUnavailable,
        _blockFilter = blockFilter,
        _unavailableFilter = unavailableFilter,
        _feedTuningRepository = feedTuningRepository,
@@ -153,7 +162,7 @@ class FullscreenFeedBloc
   final BlossomAuthService? _blossomAuthService;
   final OnRemoveVideo? _onRemoveVideo;
   final OnVideoConfirmedUnavailable? _onVideoConfirmedUnavailable;
-  final MediaAvailabilityChecker _availabilityChecker;
+  final ConfirmVideoUnavailable _confirmVideoUnavailable;
   final BlockAuthorFilter? _blockFilter;
   final UnavailableVideoFilter? _unavailableFilter;
   final FeedTuningRepository? _feedTuningRepository;
@@ -549,9 +558,9 @@ class FullscreenFeedBloc
 
   /// Handle a player-reported unavailable video.
   ///
-  /// Confirms the asset really is missing via a HEAD request before
-  /// permanently removing it. Transient player errors (network flake, slow
-  /// TLS, etc.) must not trigger removal — only a hard 404 counts.
+  /// Confirms the asset is safe to prune before permanently removing it.
+  /// Transient player errors (network flake, slow TLS, etc.) and recoverable
+  /// age-restricted media must not trigger removal.
   ///
   /// Uses `sequential()` to serialize concurrent unavailable events so the
   /// dedupe set in [FullscreenFeedState.removedVideoIds] is authoritative
@@ -567,14 +576,18 @@ class FullscreenFeedBloc
     final index = state.videos.indexWhere((v) => v.id == videoId);
     if (index < 0) return;
 
-    final videoUrl = state.videos[index].videoUrl;
+    final video = state.videos[index];
+    final videoUrl = video.videoUrl;
     if (videoUrl == null || videoUrl.isEmpty) return;
 
-    final isMissing = await _availabilityChecker.isConfirmedMissing(videoUrl);
-    if (!isMissing) {
+    final unavailable = await _confirmVideoUnavailable(
+      videoUrl: videoUrl,
+      explicitSha256: video.sha256,
+    );
+    if (!unavailable) {
       Log.warning(
-        'FullscreenFeedBloc: Player reported notFound for $videoId but HEAD '
-        'did not confirm 404 — treating as transient error, video stays.',
+        'FullscreenFeedBloc: Player reported notFound for $videoId but '
+        'confirmation did not allow prune — video stays.',
         name: 'FullscreenFeedBloc',
         category: LogCategory.video,
       );
@@ -602,6 +615,14 @@ class FullscreenFeedBloc
         pendingSkipTarget: nextIndex,
       ),
     );
+  }
+
+  static Future<bool> _defaultConfirmVideoUnavailable({
+    required String? videoUrl,
+    String? explicitSha256,
+  }) async {
+    if (videoUrl == null || videoUrl.isEmpty) return false;
+    return const MediaAvailabilityChecker().isConfirmedMissing(videoUrl);
   }
 
   /// Handle UI acknowledgement of a pending skip signal.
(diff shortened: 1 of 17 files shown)


mobile/lib/services/dead_media_feed_guard.dart
```diff
@@ -18,44 +19,66 @@ import 'package:openvine/services/media_availability_checker.dart';
...
     if (!missing) return false;
+
+    final sha256 = VideoModerationStatusService.resolveSha256(
+      explicitSha256: explicitSha256,
+      videoUrl: videoUrl,
+    );
+    if (sha256 == null) return false;
+
+    final status = await _moderationStatusService.fetchStatus(sha256);
+    return status?.blocked == true || status?.quarantined == true;
```
realmeylisdev: **should-fix:** quarantine belongs on the recoverable side of this gate, not the prune side.

Traced it through the backend repos:

`divine-moderation-service/src/blossom-client.mjs:20` maps `'QUARANTINE': 'RESTRICT'`, and the comment directly above it (lines 11-14) is explicit about why:

> IMPORTANT: AgeRestricted is an auth-gate, not a withhold [...] To actually hide content (e.g. age-review restriction) use QUARANTINE, Restricted, which is **reversible** (back to Active via SAFE/APPROVE).

`divine-blossom/blossom-core/src/types.rs:111` documents `Restricted` as "shadow restricted, only owner can access (renders as 404 to non-owners)", and `access_for()` at `:184` confirms `Restricted => NotFound`. So a quarantined blob does reach this branch with a hard 404 and passes the first gate.

`divine-moderation-api/src/index.mjs:375` sets `needs_review` true for QUARANTINE, and the moderation-service changelog describes it as a state *between* REVIEW and PERMANENT_BAN, auto-applied by the classifier above `QUARANTINE_THRESHOLD`, with manual quarantine/unquarantine via `POST /api/v1/quarantine/:sha256`.

So it's machine-applied, pending human review, reversible by design, and it's the mechanism they use for age-review restriction. That's the same family as `age_restricted`, which this PR correctly excludes.

The persistence is what makes it bite. `BrokenVideoTracker` holds a mark for 7 days, and `unmarkVideoBroken` has no production caller anywhere in `lib/` (only the definition at `broken_video_tracker.dart:141`), so a moderator approving the content never reaches the client. It stays hidden on home, profile, hashtag, grids, liked, saved, reposts and curated lists until the TTL lapses.

The sharpest case is `uploader-enforcement.mjs:167-174`: when an uploader has `approval_required`, any result not already QUARANTINE/AGE_RESTRICTED/PERMANENT_BAN is rewritten to `QUARANTINE`, **including SAFE**. So a new creator's clean clip 404s, the first viewer who scrolls past persists a client-side prune, and approving it ten minutes later doesn't bring it back for that viewer.

`blocked` (PERMANENT_BAN) is the genuinely terminal one and is right to keep. Verified live: `cfe88a19...` returns `blocked:true, quarantined:false, action:PERMANENT_BAN`; `592c0f67...` returns `blocked:false, age_restricted:true`; unseen hashes return `status:"unknown"` with `blocked:false`, so the unknown case is already safe.

Three ways out, whichever you prefer: drop `quarantined` from the gate; keep it but give quarantine a much shorter TTL than 7 days; or add an invalidation path so a later safe/unknown verdict unmarks.

mobile/lib/services/dead_media_feed_guard.dart
```diff
@@ -18,44 +19,66 @@ import 'package:openvine/services/media_availability_checker.dart';
...
+  /// age-restricted. The caller must then keep the item recoverable.
+  Future<bool> isConfirmedUnavailable({
     required String? videoUrl,
+    String? explicitSha256,
   }) async {
     if (videoUrl == null || videoUrl.isEmpty) return false;
     final missing = await _checker.isConfirmedMissing(videoUrl);
     if (!missing) return false;
+
+    final sha256 = VideoModerationStatusService.resolveSha256(
```
realmeylisdev, line 59: **nit:** the other three moderation lookups in the app gate on `VideoModerationStatusService.shouldCheckModeration(videoUrl)` first (`feed_loading_moderation_cubit.dart:32`, `feed_videos.dart:452`, `pooled_video_error_overlay.dart:101`). This one doesn't.

Blossom URLs are `/<64-hex>` by convention, so `resolveSha256` will happily extract a hash from a non-Divine host too, and a 404 there sends that blob's hash to `moderation-api.divine.video` for a lookup that can only ever come back `unknown`. Same one-line guard here keeps it consistent and saves the round trip.

mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
```diff
@@ -604,6 +617,14 @@ class FullscreenFeedBloc
     );
   }
 
+  static Future<bool> _defaultConfirmVideoUnavailable({
```
realmeylisdev, line 623: **nit:** this default is a bare 404 check with no moderation gate, i.e. exactly the behaviour the PR exists to remove, and a `true` from it still flows through `onVideoConfirmedUnavailable` and gets persisted.

It's dormant today (the one production construction at `pooled_fullscreen_video_feed_screen.dart:309` injects the real confirmer, and every bloc test injects a fake), so nothing covers it either way. Defaulting to `(...) async => false` would match the fail-closed posture the rest of the guard uses, and would stop a future call site from silently getting the old behaviour, or a test that forgets the param from firing a real HEAD.

mobile/lib/services/broken_video_tracker.dart
```diff
@@ -7,16 +7,27 @@ import 'package:shared_preferences/shared_preferences.dart';
...
     days: 7,
   ); // Clean up old entries after 7 days
 
+  final String _storageKey;
+  final String _timestampKey;
   late SharedPreferences _prefs;
   Set<String> _brokenVideoIds = {};
   Map<String, DateTime> _brokenTimestamps = {};
 
+  static String _scopedKey(String baseKey, String? ownerPubkey) {
```
realmeylisdev, line 24: **nit**, two small things from the key rename.

The old unscoped `broken_video_urls` / `broken_video_timestamps` are now never read, written or swept, so they sit in SharedPreferences indefinitely. The effective reset is a good outcome given how many old marks were age-restricted false positives, just flagging the leftover bytes.

Second, `_cleanupOldEntries` only sweeps the scope currently loaded, so every identity a device signs into leaves its own key pair behind permanently and never gets the 7-day TTL applied to it.

mobile/lib/screens/feed/pooled_fullscreen_video_feed_screen.dart
```diff
@@ -278,6 +280,18 @@ class PooledFullscreenVideoFeedScreen extends ConsumerWidget {
           orElse: () => false,
         );
 
+    Future<bool> confirmVideoUnavailable({
+      required String? videoUrl,
+      String? explicitSha256,
+    }) async {
+      final guard = ref.read(deadMediaFeedGuardProvider).asData?.value;
```
realmeylisdev, line 286: **nit:** `deadMediaFeedGuardProvider` is autoDispose and this is the only reference to it in this screen, so on its own this read never leaves `AsyncLoading`.

Checked with a throwaway widget test: a `ref.read` with no co-mounted watcher returns `AsyncLoading` on all 3 reads across 900ms, while adding a sibling `Consumer` that watches the same provider makes the identical read return `AsyncData` on the first call.

It works in practice only because `FeedVideos` renders `_Overlay`, which does `ref.watch(deadMediaFeedGuardProvider)` at `feed_videos.dart:792`. That's a real but invisible coupling, and nothing tests this closure, so if that watch ever moves the fullscreen prune goes silently dead with green CI. A `ref.watch` here, or at least a comment recording the dependency, would make it explicit.

mobile/lib/screens/feed/feed_auto_advance_error_listener.dart
```diff
@@ -15,13 +15,14 @@ import 'package:openvine/blocs/video_playback_status/video_playback_status_state
 /// Behavior by mode (see #5953):
 /// * **Auto mode** ([isAutoAdvanceActive] true) — skips immediately on any
 ///   non-ready status (unchanged), and additionally, when
-///   [confirmAndMarkMissing] is supplied, marks a HEAD-confirmed hard-404 item
-///   broken (fire-and-forget) so it is pruned from every surface on refetch.
+///   [confirmAndMarkMissing] is supplied, marks a moderation-confirmed
```
realmeylisdev, line 18: **nit:** the field doc at `:45-48` didn't get updated along with this block. It still says "Confirms (via a HEAD request) whether the active item's media is a hard 404 [...] Returns `true` only for a confirmed 404", which is no longer what the callback does. Same for the comment at `pooled_fullscreen_video_feed_screen.dart:633` ("The BLoC owns the HEAD-confirm").

realmeylisdev (dismissed):
the moderation gate is the right call and it checks out live: check-result gives blocked:true for the permanent_ban hash and blocked:false / age_restricted:true for the age-gated one, so the 404 false positives really do go away. ran the five suites locally, 118 pass.

one thing i'd want changed before this merges: quarantine isn't a terminal state. moderation-service maps QUARANTINE to blossom RESTRICT (which 404s to non-owners) and documents it as the *reversible* withhold, and uploader-enforcement rewrites even SAFE to QUARANTINE for creators pending approval. since the tracker persists 7 days and nothing ever unmarks, approving that content never reaches the client. details inline, plus a few nits.

might be worth calling out in the description that with unknown excluded, genuinely dead blobs stop being pruned at all now.

mobile/lib/providers/video_providers.dart
```diff
@@ -417,19 +417,24 @@ VideoMetadataUpdateService videoMetadataUpdateService(Ref ref) {
...
+/// says blocked/quarantined before marking it broken. Reuses the singleton
+/// [brokenVideoTrackerProvider] so a mark here is visible to every surface's
+/// `filterVideoList`. See #5953 / #6251.
 @riverpod
 Future<DeadMediaFeedGuard> deadMediaFeedGuard(Ref ref) async {
   final tracker = await ref.watch(brokenVideoTrackerProvider.future);
-  return DeadMediaFeedGuard(brokenVideoTracker: tracker);
+  return DeadMediaFeedGuard(
+    brokenVideoTracker: tracker,
+    moderationStatusService: ref.watch(videoModerationStatusServiceProvider),
```
realmeylisdev: **should-fix:** this `ref.watch` runs *after* the `await` on line 433, and `deadMediaFeedGuardProvider` is autoDispose. If the element gets disposed during that gap, the watch throws.

Reproduced in this repo with a throwaway widget test that reads the provider with no co-mounted watcher:

```
UnmountedRefException: Cannot use the Ref of deadMediaFeedGuardProvider after it has been disposed.
#1  Ref.watch (package:riverpod/src/core/ref.dart:724:5)
#2  deadMediaFeedGuard (package:openvine/providers/video_providers.dart:436:34)
<asynchronous suspension>
```

`main` doesn't have this. Its body is `return DeadMediaFeedGuard(brokenVideoTracker: tracker);`, with no `ref` use after the await, so it's new here.

Usually the element has a listener (`_Overlay` watches it), so the common path is fine. The window is the last overlay unmounting or the user navigating away while the build is still in flight, and the auth-flip rebuild this PR adds to `brokenVideoTrackerProvider` widens it.

Hoisting the watch above the await costs nothing:

```dart
Future<DeadMediaFeedGuard> deadMediaFeedGuard(Ref ref) async {
  final moderationStatusService = ref.watch(
    videoModerationStatusServiceProvider,
  );
  final tracker = await ref.watch(brokenVideoTrackerProvider.future);
  return DeadMediaFeedGuard(
    brokenVideoTracker: tracker,
    moderationStatusService: moderationStatusService,
  );
}
```

realmeylisdev (commented):
one more from re-checking the provider wiring, plus a pointer on the migration nit.

mobile/lib/services/broken_video_tracker.dart
realmeylisdev: Follow-up on the migration half of this: the repo already has a pattern for it. `FeedModePreferenceStore` (`lib/blocs/video_feed/feed_mode_preference_store.dart`) keeps `_legacyFeedModeKey`, has the anonymous scope read that legacy key *directly* instead of a separate `_anonymous` key (`:32-34`), migrates it for authenticated users (`:56-65`), and removes it on the first scoped write (`:121`). Same shape here would clear the orphan and keep anonymous continuity for free.

Also, the legacy blob isn't only orphaned, it's frozen: the 7-day sweep only ever ran from `initialize()` against the loaded scope, so those entries now never expire either.

realmeylisdev (approved):
Six commits pushed to the branch, one per finding, each revertable alone.

- `2cfdf38` — quarantine off the prune gate; only `blocked` persists. The fix belongs in the predicate: the fullscreen feed injects `isConfirmedUnavailable` and persists off that same bool, so narrowing `confirmAndMarkMissing` alone would leave the fullscreen 7-day prune gated on a reversible verdict. Quarantined items keep the "Content restricted" tile and its Skip button — that comes from `FeedLoadingModerationCubit` and `PooledVideoErrorOverlay`, neither of which consults this guard, and home-feed Auto still advances unconditionally.
- `c2281b5` — `deadMediaFeedGuardProvider` is now keepAlive. Fixing the provider rather than the one screen covers both call sites plus the home feed's per-overlay inflate/dispose churn.
- `0fd8f64` — moderation watch hoisted above the await.
- `da5e6bf` — the bloc's default confirmer fails closed instead of doing a bare 404.
- `232258d` — drops the orphaned unscoped tracker keys. Left the sibling half alone: an inactive scope already gets the same 7-day TTL whenever it is next loaded, so a prefix sweep would only reclaim a few bytes earlier.
- `e608cd2` — nine stale comments, including `MediaAvailabilityChecker`'s "the only status that justifies a persistent prune", which is the file a reader consults first.

Withdrawn: the host-gate nit was wrong. `resolveSha256` takes `explicitSha256` without inspecting the URL, and the verdict is keyed on the hash alone, so gating on the host would stop pruning a `blocked` blob served from a mirror. Left ungated deliberately and the reason is now in the class doc.

Two fixes are not pinned by a test, and the commit bodies say so: the post-await `ref` hoist (Riverpod discards the superseded build's error, so a test passes with and without it — I tried a listener-less `.future` read and a completer-gated dependency rebuild) and the fail-closed default (a test-environment HEAD never answers 404, so old and new defaults are indistinguishable).

Still yours: worth saying in the description that with `unknown` excluded, genuinely dead media is no longer pruned at all — a never-stored blob has no moderation record, so it can't reach a terminal verdict.

Validation: 5177 tests across `test/providers`, `test/services`, `test/blocs/fullscreen_feed`, `test/screens/feed`, `test/widgets/video_feed_item`, `test/widgets/profile`; `flutter analyze lib test integration_test` clean; the design-system, untested-services, test-structure and transport/isolation guards all pass; codegen regenerated and committed. Not rebased — GitHub reports no conflicts and this is review-feedback work.

mobile/lib/services/dead_media_feed_guard.dart
realmeylisdev: Fixed in `2cfdf38` — dropped `quarantined` from the predicate, so only `blocked` persists. Went with option 1 of the three rather than a shorter TTL or an invalidation path, because both of those add API surface, and the fullscreen path persists off `isConfirmedUnavailable` directly, so the predicate is the only place that closes both paths. The lost auto-advance turns out not to matter: `FeedLoadingModerationCubit` and `PooledVideoErrorOverlay` both treat quarantined as restricted and surface a Skip button without consulting this guard.

mobile/lib/services/dead_media_feed_guard.dart
realmeylisdev: Withdrawn — I was wrong. `resolveSha256` takes `explicitSha256` without inspecting the URL at all, and `/check-result/$sha256` is keyed on the hash, so a third-party host *can* clear the second gate: a `blocked` blob served from a mirror would stop being prunable if I gated on the host. The real gap was that the omission was undocumented and invited exactly the "restore consistency" change I proposed, so `e608cd2` records the reason in the class doc instead.

mobile/lib/blocs/fullscreen_feed/fullscreen_feed_bloc.dart
realmeylisdev: Fixed in `da5e6bf` — the default is now `async => false`, and the bare-404 path was the last user of the `MediaAvailabilityChecker` import here. Chose fail-closed over making the parameter required: it needs no test edits, and five sibling injected callbacks are all optional-with-a-graceful-default. Didn't add a test — a test-environment HEAD never answers 404, so it would pass against the old default too.

mobile/lib/services/broken_video_tracker.dart
realmeylisdev: First half fixed in `232258d` — the legacy unscoped pair is removed on the first `initialize()`. Dropping rather than migrating keeps the behaviour identical, since those marks are already unreachable on this branch.

Second half left alone deliberately: `_cleanupOldEntries` applies the same 7-day TTL whenever a scope is next loaded, and a fully-expired scope writes back empty JSON rather than removing keys, so a prefix sweep would only reclaim a few bytes earlier for scopes that are never loaded again.

mobile/lib/screens/feed/pooled_fullscreen_video_feed_screen.dart
realmeylisdev: Fixed in `c2281b5`, at the provider rather than here — `deadMediaFeedGuardProvider` is now keepAlive. That covers this read and `feed_videos.dart`'s per-overlay inflate/dispose churn, and it costs nothing: the guard is a field-free value object over two keepAlive services, and it still rebuilds on auth flips through `brokenVideoTrackerProvider`.

mobile/lib/screens/feed/feed_auto_advance_error_listener.dart
realmeylisdev: Fixed in `e608cd2`, along with seven more. The `[MediaAvailabilityChecker]` reference in `FullscreenFeedVideoUnavailable`'s doc turned out to be the last thing keeping that import alive in the bloc library.

mobile/lib/providers/video_providers.dart
realmeylisdev: Fixed in `0fd8f64` — hoisted exactly as suggested. One correction to my own diagnosis: it isn't only disposal. A rebuild mid-suspension also replaces the element's `ref`, so the hazard would survive `keepAlive` (which `c2281b5` now sets for a different reason). No test: Riverpod discards the superseded build's error, so both a listener-less `.future` read and a completer-gated dependency rebuild pass with and without the fix.

PR #6725, opened 2026-08-05
feat(clip-reuse): credit reused clip sources

NotThatKindOfDrLiz:
## Description

Carries Phase 0 imported clip provenance into publish attribution by deriving the existing inspired-by video reference from the timeline when there is a single reused clip source. Existing explicit and sound-derived attribution keep precedence.

Also fixes the existing self-attribution hole by suppressing inspired-by a-tags when the referenced video was created by the publisher, while preserving malformed addressable-id behavior for the existing a-tag path.

**Related Issue:** Relates to #6315

## Out of Scope

Creator-facing clip reuse settings and source watermarking remain out of scope for this PR.

## Verification

- `flutter test test/providers/video_editor_provider_test.dart test/services/video_event_publisher_inspired_by_tags_test.dart`
- Pre-push changed-file checks: analyzer plus `test/providers/video_editor_provider_test.dart`, `test/services/video_event_publisher_inspired_by_tags_test.dart`, and `test/services/video_event_publisher_test.dart`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/providers/video_editor_provider.dart
+++ b/mobile/lib/providers/video_editor_provider.dart
@@ -545,7 +545,8 @@ class VideoEditorNotifier extends Notifier<VideoEditorProviderState> {
   /// populates [selectedAudioEventId] and [selectedAudioRelay] for the
   /// publisher to add an `["e", ..., "audio"]` tag. Also auto-populates
   /// [inspiredByVideo] from the sound's [sourceVideoReference] if not
-  /// already set.
+  /// already set, falling back to a single imported clip source when every
+  /// provenance-bearing clip agrees.
   DivineVideoDraft getActiveDraft({bool isAutosave = false, String? draftId}) {
     // Read selected sound from local state. The recorder flow sets
     // [selectedSound]; the editor's "add music" flow instead appends reused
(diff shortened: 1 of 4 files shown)


rabble (approved):

Both halves are sound. Deriving inspired-by from clip provenance only when the timeline agrees on a single source is the conservative choice — crediting one of several reused sources would be worse than crediting none — and putting it behind `??=` after the sound-derived path keeps the existing precedence intact without a flag. `credits one reused clip source mixed with local clips` is the test that matters: clips without provenance are original material and correctly don't veto the credit.

The self-attribution fix is right and the malformed-id behaviour is genuinely preserved rather than accidentally preserved. `creatorPubkey` returns `''` for an id that doesn't split, so the `isEmpty` arm keeps emitting the a-tag, and the assertion added to the malformed-a-tag test pins that — it would fail if someone later "tidied" the empty check away. Falling through to emit when `_authService` is null is also the correct default: no auth means no basis to claim self-reference, and the old behaviour is the safe one.

**One thing I'd change, and it partially defeats the feature.** `_attributionSourceFromClips` bails when either field differs:

```dart
if (source.addressableId != candidate.addressableId ||
    source.relayUrl != candidate.relayUrl) {
  return null;
}
```

Two clips cut from the *same* source video but carrying different `sourceRelayHint` values are one reused source, not two — but this drops the credit entirely. That's arguably the most common reuse shape there is: chopping one video into several clips on the timeline. Whether the hints actually diverge depends on how they were discovered, and clips imported in a single flow will usually agree, but nothing guarantees it across sessions or discovery paths, and the failure is silent.

Comparing `addressableId` alone and keeping the first non-null `relayUrl` gives the same conflict detection where it matters — the relay hint is a discovery aid, not part of the source's identity. A test with matching ids and differing hints would pin it.

Not blocking, because the failure mode is "no credit" rather than "wrong credit", which is the right direction to fail. But it's a small change and worth doing before this ships as a user-visible feature.

Minor, for when you're next in the file: `_attributionSourceFromClips()` runs on autosave while `_attributionSoundFromEditorTracks()` is explicitly suppressed by `isAutosave`. That asymmetry may well be deliberate — clip provenance is cheap and deterministic where the sound derivation isn't — but the doc comment on `getActiveDraft` now describes both paths as one behaviour and doesn't mention the difference.

Ship it.

Issue #6227, opened 2026-07-20
feat(audio): Improve audio attribution and reusable sound tracking


hm21:
### What would you like?

Improve the way audio is attributed and reused across Divine.

Ideas include:
- Allow creators to upload sounds/music specifically for reuse.
- Automatically credit the original creator whenever their sound is used.
- Display how many videos have reused a sound.
- Provide a dedicated page for each sound showing every video that uses it, making trends and challenges easier to discover.

### How would this be useful for you?

Creators receive proper attribution and recognition when their sounds become popular, while users can easily discover the origin of a sound and explore other videos using it.

This also encourages the creation of reusable audio, makes viral content easier to follow, and helps build community around shared sounds and trends instead of disconnected posts.

### When would you use this?

Whenever I create a video using another creator's sound, or when I want to discover more videos using the same audio.

As a creator, I'd also like to see how widely my sounds have been reused and how they spread across the platform.

### Anything else?

This is similar to how platforms like TikTok group videos by sound, but could go a step further by making reusable audio a first-class feature and providing clear attribution to the original creator.

---

Requested on discord [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1528613376014549143)

NotThatKindOfDrLiz:
Closing this umbrella as decomposed and either delivered or re-tracked.

Scope map:

- Standalone reusable sound uploads: still open as #6743.
- Original creator credit when a reused sound is used: delivered by #6727 on the sound detail page, feed attribution row, and metadata sounds section.
- Reuse count for a sound: delivered on the sound detail header. The follow-up for surfacing counts in sound discovery/library list contexts is #6931.
- Dedicated sound page listing videos using that sound: already exists at `/sound/:id` and lists videos via the sound detail screen.
- Creator-facing visibility into how widely their sounds spread: still open as #6726.

The active follow-ups are now grouped under the audio epic #4340: #6726, #6743, and #6931.

Issue #6726, opened 2026-08-05
feat(audio): add creator sound reuse analytics


NotThatKindOfDrLiz:
## Problem

Issue #6227 also asks creators to see how widely their sounds have been reused. That is separate from sound-detail attribution and needs its own product/performance scope.

## Acceptance Criteria

- Creator analytics shows the signed-in creator their published community sounds.
- Each listed sound includes its reuse count.
- The implementation avoids an unbounded per-sound relay COUNT query pattern.
- Empty, loading, and error states match the existing creator analytics screen.

## Technical Notes

Mobile already has `soundsByCreatorProvider(pubkey)` and `soundUsageCountProvider(soundId)`, so the main open question is UX and query shape rather than basic repository capability.

Split from #6227.

PR #6727, opened 2026-08-05
fix(audio): credit sound creators on detail page

NotThatKindOfDrLiz:
## Summary

- Show community sound creator attribution on the sound detail header, using resolved profile names and generated fallbacks.
- Make profile-backed sound credits tappable so users can open the credited creator or publisher profile.
- Keep bundled/source attribution intact and move sound attribution display-name fallback into the shared audio credit helper.
- Replace the adjacent raw preview play/stop icons with Divine design-system icons.

## Scope notes

This narrows #6227 to the remaining sound-detail creator-credit gap. Reuse counts and the dedicated sound page already exist. Standalone sound upload remains intentionally out of scope and is tracked in #6743. Creator-facing sound reuse analytics was split to #6726.

Refs #6227

## Verification

- flutter test test/screens/sound_detail_screen_test.dart test/widgets/metadata_sounds_section_test.dart test/widgets/audio_attribution_row_test.dart
- flutter analyze

### DIFF
--- a/mobile/lib/screens/sound_detail_screen.dart
+++ b/mobile/lib/screens/sound_detail_screen.dart
@@ -508,9 +508,13 @@ class _SoundHeaderState extends ConsumerState<_SoundHeader> {
                 pubkey: profileCreditPubkey,
                 profile: profileCreditProfile,
               );
+    // Matches the feed pill and the metadata sheet: a credited creator who is
+    // not the signer means the signer only shared the sound. Bundled sounds
+    // carry a marker pubkey rather than a real signer, so they are excluded
+    // instead of being credited to a generated name.
     final showsSeparatePublisher =
         _artistName != null &&
-        sound.creatorPubkey != null &&
+        !sound.isBundled &&
         sound.creatorPubkey != sound.pubkey;
     final publisherProfile = showsSeparatePublisher
         ? ref.watch(userProfileReactiveProvider(sound.pubkey)).value
--- a/mobile/lib/screens/sound_detail_screen.dart
+++ b/mobile/lib/screens/sound_detail_screen.dart
@@ -31,6 +31,11 @@ import 'package:sound_service/sound_service.dart';
 import 'package:unified_logger/unified_logger.dart';
 import 'package:url_launcher/url_launcher.dart';
 
+const _soundDetailCreatorAttributionIdentifier =
+    'sound_detail_creator_attribution';
+const _soundDetailPublisherAttributionIdentifier =
+    'sound_detail_publisher_attribution';
+
 /// Screen displaying details of a specific sound and videos using it.
 ///
 /// Features:
@@ -569,6 +574,8 @@ class _SoundHeaderState extends ConsumerState<_SoundHeader> {
                       _ProfileAttributionInfo(
                         text: context.l10n.soundCreatorBy(profileCreditName),
                         pubkey: profileCreditPubkey,
+                        semanticsIdentifier:
+                            _soundDetailCreatorAttributionIdentifier,
                         license: _license,
                         sourceUrl: _sourceUrl,
                       )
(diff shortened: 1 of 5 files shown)


hm21 (changes requested):
Pushed two fixes as separate commits — please review and revert anything you disagree with.

**70cf1692ad — dead "View source" link.** `AudioEvent.fromVideoOriginalSound` puts the label `Original Sound` in `source`, `_lookUpAttribution` maps `source` → `_sourceUrl`, and the new `_ProfileAttributionInfo` renders it unconditionally. Opening a video's own original sound from the metadata sheet showed a "View source" link that `canLaunchUrl` rejects on tap, so it did nothing. Now gated on an absolute URL, which also covers the latent bundled-lookup-miss case.

**9cd0bc3799 — publisher credit dropped.** Requiring `creatorPubkey != null` removed the "Shared by …" line for the external-provider bridge path (`video_event_publisher_audio.dart:181`), which credits the provider's artist by name and never sets a creator pubkey. `audio_attribution_row.dart:103` and `metadata_sounds_section.dart:276` still show it — both with tests pinning that exact case — so the same event read differently on three surfaces. Swapped the guard to `!sound.isBundled`, which I think is what you were actually protecting against: the new non-null name fallback would otherwise print a generated name for the `bundled` marker pubkey.

Both verified against `main` with throwaway probes before fixing: "View source" 0 → 1, publisher line present → gone. A regression test is added for each. `flutter analyze` clean; `sound_detail_screen_test.dart`, `audio_attribution_row_test.dart` and `metadata_sounds_section_test.dart` green (80 tests).

One thing for you to decide: `Closes #6227` auto-closes the issue on merge, but its first ask — creators uploading sounds specifically for reuse — isn't implemented, and Kind 1063 is only ever published as a side effect of publishing a video. Reuse counts and the sound page already existed, so what this PR actually closes is the detail-header credit gap. `Refs #6227` plus a follow-up issue for standalone sound upload reads more accurate to me, unless you'd rather re-scope the issue explicitly.

Nits, take or leave: `_ProfileCreditLink` hardcodes `identifier: 'sound_detail_creator_attribution'` and is used for both the creator and the publisher link, so two widgets share one semantics id when both render. `_ProfileAttributionInfo` also mixes `context.vineColors.secondaryText` (outer span) with `VineTheme.onSurfaceVariant` (link) in the same line.

NotThatKindOfDrLiz:
Takeover fixes pushed in 9eee9df7.

What changed:
- Routed the metadata sheet publisher fallback through AudioAttributionCredit.publisherNameFor so all three sound attribution surfaces use the shared helper.
- Split creator vs publisher profile-credit semantics identifiers on the sound detail header.
- Hardened profile-backed credits so draft-local imported audio does not link to the local_import marker pubkey.
- Aligned the remaining attribution text/link styling with VineTheme body-small styles.
- Added regression coverage for the creator/publisher semantics split and local-import marker case.
- Updated the PR body from Closes #6227 to Refs #6227 and created #6743 for standalone reusable sound uploads.

Verification:
- flutter test test/screens/sound_detail_screen_test.dart test/widgets/metadata_sounds_section_test.dart test/widgets/audio_attribution_row_test.dart
- flutter analyze
- Pre-push hook analyzer + changed-file test
- GitHub CI is green on 9eee9df7

hm21 (approved):
All five points addressed — thanks. Verified 9eee9df locally: my two commits survived the rebase byte-identical, the semantics split and local-import guard are correct (checked both `fromLocalImport` call sites; `voice_over_cubit` prefixes with `local_import_voice_over`, so `isLocalImport` covers it), and the metadata-sheet refactor is behaviour-neutral. 82 tests green across the three attribution suites, CI green, TextStyle ceiling shrinks 12 → 10 in the file so no baseline churn. Yours to merge.

PR #6728, opened 2026-08-05
fix(video-player): distinguish exhausted auth retries

### LINKED ISSUE #6253: fix(video-player): distinguish expired-token 401 from age-gate

NotThatKindOfDrLiz:
## Summary
- track session-scoped authenticated retry exhaustion for age-gated playback
- mark authorized-but-still-failed retry attempts as neutral unavailable instead of age verification failures
- keep skip-only unavailable overlays while leaving classifier, availability checks, and pruning untouched
- add localized unavailable video copy and generated l10n outputs

## Verification
- flutter test test/blocs/video_playback_status/video_playback_status_cubit_test.dart test/screens/feed/pooled_age_restricted_retry_test.dart test/widgets/pooled_video_error_overlay_test.dart test/widgets/video_feed_item/feed_videos_test.dart test/widgets/video_feed_item/moderated_content_overlay_test.dart test/l10n/arb_consistency_test.dart
- flutter analyze lib test

Closes #6253

### DIFF
--- a/mobile/lib/blocs/video_playback_status/video_playback_status_cubit.dart
+++ b/mobile/lib/blocs/video_playback_status/video_playback_status_cubit.dart
@@ -59,6 +59,12 @@ class VideoPlaybackStatusCubit extends Cubit<VideoPlaybackStatusState> {
     return true;
   }
 
+  /// Marks that authenticated playback retry still failed for [eventId].
+  void markAuthRetryExhausted(String eventId) {
+    if (state.hasAuthRetryExhausted(eventId)) return;
+    emit(state.withAuthRetryExhausted(eventId));
+  }
+
   /// Consumes one automatic age-verification retry when the current playback
   /// state and viewer policy allow it.
   ///
--- a/mobile/lib/blocs/video_playback_status/video_playback_status_state.dart
+++ b/mobile/lib/blocs/video_playback_status/video_playback_status_state.dart
@@ -22,6 +22,9 @@ enum PlaybackStatus {
   /// Content not found — 404 or unresolved blob hash.
   notFound,
 
+  /// Playback is unavailable after authenticated access was rejected.
+  unavailable,
+
   /// Any other playback failure.
   generic,
 }
@@ -41,6 +44,7 @@ class VideoPlaybackStatusState extends Equatable {
     LinkedHashMap<String, PlaybackStatus>? statuses,
     Set<String>? verifyingIds,
     Set<String>? autoRetryAttemptedIds,
+    Set<String>? authRetryExhaustedIds,
   }) : _statuses = statuses == null
            ? LinkedHashMap<String, PlaybackStatus>()
            : LinkedHashMap<String, PlaybackStatus>.from(statuses),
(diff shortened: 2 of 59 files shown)


hm21 (approved):
Approving with three fixes pushed as separate commits — revert any you disagree with, the merge call is yours.

**b06a25e — only exhaust the auth retry on a repeated 401.** `retryPlayback` was gated on `retryAt`'s bare bool, which is `!_errors.contains(index)` (`infinite_video_feed.dart:1332`), and `_errors.add(index)` fires on *every* init exception (`:1262`) — a network drop, decode error, or 5xx failed the retry exactly like a repeated 401 did. So a blip during the post-verification reload marked `markAuthRetryExhausted` for the session and rendered the item as "Video unavailable" with Verify age *and* Retry both stripped (`showRetry` gets `&& !isUnavailableAfterAuthRetry`), leaving only Skip, with `_authRetryExhaustedIds` cleared only by `cubit.clear()`. Since #6253 scopes this to "after an authenticated retry still **returns 401**", I threaded the post-retry error type out: new `InfiniteVideoFeedState.errorTypeAt(int)` mirroring what `errorBuilder` is handed, `retryPlayback` now returns `PooledRetryOutcome` (`{succeeded, errorType}`), and both the manual and auto paths only mark exhausted when it is still `ageRestricted`. Anything else keeps the retryable age-gated card — main's behavior today. The four `retryAt` closures in `feed_videos.dart` collapse into one `_retryFeedItem` helper.

**1e5463d — unreachable switch arm.** `isAgeRestricted` is true whenever the type is `ageRestricted` and the retry is not exhausted, so `(ageRestricted, false, false, _)` can never be selected; it only keeps the switch exhaustive. Repointing it at the new unavailable copy read as intentional behavior that never runs, so I restored the original value and noted why it is dead.

**189a8b9 — snackbar repeated the overlay title.** `videoErrorUnavailable` is the overlay heading, so the snackbar showed "Video unavailable" beside an overlay already reading "Video unavailable". Switched to the body sentence.

**Left alone, your call:** the non-auth branch still shows `videoErrorVerifyAgeFailed` ("Couldn't verify your age. Please try again."), which is slightly off when auth *succeeded* and the reload failed. It is what main does today for every failed retry, and new copy means 22 locale translations, so it did not belong in a review fix.

**Validation:** `flutter analyze lib test integration_test` clean; 522 app tests across `test/screens/feed/`, `test/widgets/video_feed_item/`, `pooled_video_error_overlay_test.dart`, `test/blocs/video_playback_status/`, `arb_consistency_test.dart`; 229 package tests with coverage still 100%. Added two regression tests (manual + auto path) proving a non-auth failure leaves the status `ageRestricted` and `hasAuthRetryExhausted` false, plus a package test for `errorTypeAt`. All 15 CI checks green, including `Infinite Video Feed CI`.

The rest reviewed clean — l10n is complete and translated across all 22 locales, and the new tests assert user-visible behavior rather than configuration.

PR #6729, opened 2026-08-05
feat(lists): notify when videos are added to public lists

### LINKED ISSUE #6391: feat(lists): notify users when their videos are added to public lists

NotThatKindOfDrLiz:
## Summary
- add list_add DTO/domain parsing and preserve list context through notification repository caching/grouping
- render list-add notification rows with list title copy and a listPlus icon
- route list_add taps to /list/:pubkey/:listId via the public curated-list coordinate

Closes #6391

## Verification
- flutter analyze lib test (mobile)
- flutter analyze lib test (packages/funnelcake_api_client)
- flutter analyze lib test (packages/models)
- flutter analyze lib test (packages/notification_repository)
- flutter test test/notifications/routing/notification_tap_target_test.dart test/notifications/widgets/notification_type_icon_spec_test.dart test/notifications/widgets/video_notification_row_test.dart test/l10n/arb_consistency_test.dart test/goldens/widgets/notification_rows_golden_test.dart (mobile)
- flutter test test/src/models/relay_notification_test.dart (packages/funnelcake_api_client)
- flutter test test/src/video_notification_test.dart (packages/models)
- flutter test test/src/notification_repository_test.dart (packages/notification_repository)
- scripts/golden.sh verify test/goldens/widgets/notification_rows_golden_test.dart
- pre-push hook: analyze and changed-file tests

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -3404,7 +3404,34 @@
             }
         }
     },
-    "notificationRepliedToYourComment": "{actorName} ለአስተያየትዎ ምላሽ ሰጥተዋል",
+    "notificationAddedYourVideoToList": "{actorName} added your vine to {listName}",
+  "@notificationAddedYourVideoToList": {
+    "description": "Shown when someone adds one of the user's videos to a public curated list.",
+    "placeholders": {
+      "actorName": {
+        "type": "String"
+      },
+      "listName": {
+        "type": "String"
+      }
+    }
+  },
+  "notificationAddedYourVideosToList": "{actorName} added {count} of your vines to {listName}",
+  "@notificationAddedYourVideosToList": {
+    "description": "Shown when someone adds multiple of the user's videos to the same public curated list.",
+    "placeholders": {
+      "actorName": {
+        "type": "String"
+      },
+      "count": {
+        "type": "int"
+      },
+      "listName": {
+        "type": "String"
+      }
+    }
+  },
+  "notificationRepliedToYourComment": "{actorName} ለአስተያየትዎ ምላሽ ሰጥተዋል",
     "@notificationRepliedToYourComment": {
         "description": "Verb phrase shown after the actor name in a reply notification."
     },
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -3285,7 +3285,34 @@
         }
     },
     "notificationsViewProfilesSemanticLabel": "عرض الملفات الشخصية",
-    "notificationRepliedToYourComment": "{actorName} ردّ على تعليقك",
+    "notificationAddedYourVideoToList": "{actorName} added your vine to {listName}",
+  "@notificationAddedYourVideoToList": {
+    "description": "Shown when someone adds one of the user's videos to a public curated list.",
+    "placeholders": {
+      "actorName": {
+        "type": "String"
+      },
+      "listName": {
+        "type": "String"
+      }
+    }
+  },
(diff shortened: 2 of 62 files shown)


rabble (approved):

62 files, but the real surface is ~17 — the rest is ARB mirroring across all 17 locales with the generated output committed, which is the part people usually defer and then forget. Good.

The design decision I went looking for is the one that's easy to get wrong, and it's right: `_snapshotVideoGroupKey` keys `listAdd` on `listCoordinate ?? videoEventId` rather than `videoEventId`, so two different lists containing the same video don't collapse into one row showing the wrong list title. The copy matches that grouping exactly — "added {count} of your vines to {listName}" reads correctly for a group keyed by list — which tells me the grouping semantics and the string were designed together rather than one being retrofitted onto the other. `_mergeAppendedVideoGroup` carrying `listTitle`/`listCoordinate` through with `_nonEmpty(existing) ?? _nonEmpty(incoming)` keeps a merged group from losing its list context.

Routing is clean: `listTargetFromCoordinate` validates kind, pubkey, and d-tag before building a target, both tap paths go through `CuratedListByAuthorScreen.pathFor` rather than a hand-built string, and the `/list/:pubkey/:listId` shape lines up with what #6710 documents.

**The one thing I'd fix before this is user-visible outside English.** The plural is selected in Dart, not in ICU:

```dart
final fullText = notification.totalCount > 1
    ? l10n.notificationAddedYourVideosToList(...)
    : l10n.notificationAddedYourVideoToList(...);
```

with `"added {count} of your vines to {listName}"` as a plain `{count}` placeholder. `localization.md` calls this out directly — "Plurals use ICU syntax in ARB files, not conditional logic in Dart" — and the reason bites here specifically: `app_ar.arb` and `app_pl.arb` are both in the shipped set, and Arabic has six plural categories while Polish has four. A two-branch `> 1` split in Dart can't express either, so those locales get grammatically wrong text for counts like 2, 5, and 22 no matter how good the translation is. English and the Romance locales happen to survive it, which is exactly why this kind of thing ships.

Collapsing both keys into one ICU `plural` with `=1`/`other` fixes it and shrinks the ARB surface. It's not a crash and it doesn't block the feature working, so I'm not holding the PR — but it should land before this gets turned on broadly, not after someone files a translation bug.

Two smaller things:

- **`NotificationKind.system => const ['system']`** changes what the System filter asks the server for. The old mapping (`['list_add']`) was obviously a stopgap and replacing it is right, but it's worth confirming Funnelcake actually emits a `system` notification type — if it doesn't, the System filter goes from "returns list-add rows" to "returns nothing", which is a different kind of wrong and won't show up in any test here.
- **`const int _curatedListKind = 30005;`** is a file-private literal, while `list_providers.dart`, `list_search_bloc.dart`, and `app_constants.dart` all reference kind 30005 in prose. There's no shared constant to reuse today, so this isn't a regression — but a fourth place encoding the same protocol number is the point at which one belongs in `models` next to `NIP71VideoKinds`.

`_localizedActorAndListSentenceSpans` bolding via `indexOf` is fragile in principle (an actor whose display name collides with a word earlier in the sentence gets the wrong span), but it degrades to plain or actor-only spans rather than throwing, and it's the same pattern `localizedActorSentenceSpans` already uses. Consistent with the codebase; not worth diverging here.

Ship it, then fix the plural.

Issue #6730, opened 2026-08-05
task(push): add end-to-end verification for push-service token processing


NotThatKindOfDrLiz:
## Context
Relay acceptance for push-control kinds 3079, 3080, and 3083 is confirmed in production as part of #4437, but that only proves the relay accepted the events.

## Gap
There is no mobile-side or externally accessible path to verify that divine-push-service processed the event and wrote or removed the encrypted token in Redis. The temporary relay diagnostics removed by #4437 could not observe this downstream processing.

#5827 showed this exact blind spot: the relay could return OK true while deregistration processing failed because mobile sent empty content.

## Acceptance Criteria
- Document a production verification path for push-service token registration, deregistration, and preference processing, or expose a safe health/verification endpoint.
- Verification does not expose encrypted token payloads or secrets.
- Runbook covers at least kinds 3079, 3080, and 3083.

NotThatKindOfDrLiz:
Additional bug scope from the August 14 iOS release log:

Push registration publish can fail with `acceptedBy=[]` and `noResponseFrom=[relay]` after the 5s OK wait, and the session showed no retry afterward. End-to-end push processing verification is still useful, but mobile also needs a client-side retry contract for this transient publish outcome.

Acceptance criteria to add here or split into a bug:
- Treat `noResponseFrom` on push registration/deregistration/preference control events as retryable, not terminal.
- Add bounded retry with backoff and jitter, preserving battery/network limits.
- Persist enough pending state that token refresh/registration is not silently lost if the first relay OK wait times out.
- Add a unit test proving `noResponseFrom` schedules retry and eventual success clears the pending registration.

PR #6731, opened 2026-08-05
chore(push): remove push-control relay diagnostics

### LINKED ISSUE #4437: task(push): Remove push-control relay diagnostics after rollout
## Description
Remove the temporary push-control relay diagnostic wiring once the Divine relay rollout has verified production acceptance for push registration, deregistration, and preference control events.

## Context
PR #4417 adds temporary diagnostics for push-control publishes so the rollout for issue #4411 can confirm relay-level send and OK outcomes without logging encrypted payloads.

## Implementation Notes
Review the push-control publish path after production verification and remove the caller-supplied diagnostic tag plumbing that is only needed for rollout visibility.

## Acceptance Criteria
- [ ] Push registration, deregistration, and preference publishes are verified in production after the rollout.
- [ ] Temporary push-control diagnostic tags and TODO markers are removed.
- [ ] SDK remains domain-neutral with no Divine-specific kind checks.

## Testing
Run the focused push notification service tests and nostr SDK relay-pool tests after removal.

NotThatKindOfDrLiz:
## Summary
- Remove the rollout-only push-control relay diagnostic tag path from the app, nostr_client, and nostr_sdk.
- Keep the permanent NIP-20 OK publish outcome handling for push-control events.
- Remove dead diagnostic-only publish metadata, including PublishOutcome.eventKind and PublishTracker.eventKind.
- Update affected mocktail stubs and publish outcome tests.

## Follow-up
- Filed #6730 for the remaining push-service token processing observability gap, which the relay diagnostics could not verify.

Closes #4437

## Verification
- dart analyze lib/services/push_notification_service.dart packages/nostr_client/lib/src/nostr_client.dart packages/nostr_sdk/lib/nostr.dart packages/nostr_sdk/lib/relay/publish_outcome.dart packages/nostr_sdk/lib/relay/relay_pool.dart test/services/push_notification_service_test.dart test/providers/push_notification_sync_test.dart packages/nostr_client/test/src/nostr_client_test.dart packages/nostr_sdk/test/unit/publish_outcome_test.dart packages/nostr_sdk/test/unit/relay_pool_auth_test.dart packages/nostr_sdk/test/unit/relay_pool_concurrency_test.dart packages/nostr_sdk/test/unit/relay_pool_zombie_remediation_test.dart packages/nostr_sdk/test/unit/relay_pool_publish_targets_test.dart packages/badge_repository/test/src/badge_repository_test.dart
- flutter test test/services/push_notification_service_test.dart test/providers/push_notification_sync_test.dart
- flutter test test/unit/relay_pool_auth_test.dart (from mobile/packages/nostr_sdk)
- flutter test (from mobile/packages/nostr_sdk)
- flutter test (from mobile/packages/nostr_client)
- flutter test test/src/badge_repository_test.dart (from mobile/packages/badge_repository)
- mise exec -- dart format lib test integration_test
- pre-push hook: analyzer plus changed-file tests


rabble (approved):
## Overview

Removes the rollout-only push-control relay diagnostics: the `diagnosticTag` parameter threaded through `PushNotificationService` → `NostrClient.publishEventAwaitOk` → `Nostr.sendEventAwaitOk` → `RelayPool.sendEventAwaitOk`/`_sendCollect`, the eleven `logDiagnostic` call sites, and the `PublishOutcome.eventKind` / `PublishTracker.eventKind` fields that existed only to feed them.

Verified the deletion is safe rather than just self-consistent:

- **No surviving consumers.** `PublishOutcome.eventKind` and `PublishTracker.eventKind` have no readers anywhere in `lib/` or `packages/` outside the three SDK files this PR edits — every other `eventKind` in the tree belongs to an unrelated type (`Comment`, `VideoEvent`, `NostrAppAuditEvent`, the app-bridge policy, `CurationSet`). Nothing downstream loses a field it was reading.
- **The permanent path is untouched.** `PublishTracker.onAccepted` / `onRejected` / `_shouldRetryPublishAfterAuth` and the auth-required retry flow are unchanged; only the logging wrapped around them goes. `_recordPublishBreadth` never read `eventKind`, so the publish-breadth log is unaffected.
- **The `diagnosticTag == null` ternary in `NostrClient` collapses correctly** — both branches were identical apart from the named argument, so the flattened call is the same behaviour.
- **The TODO it closes was the acceptable kind** (transitional code with a tracking issue), and the observability gap the diagnostics couldn't answer is filed as #6730 rather than dropped. That's the right disposal for a rollout scaffold.

Approving. One coordination note.

---

### Merge collision with #6746

#6746 (`fix(nostr): bound and validate relay-list admission`) adds new tests that call the parameter this PR deletes:

```dart
await pool.sendEventAwaitOk(
  ['EVENT', wrap.toJson()],
  eventId: wrap.id,
  eventKind: wrap.kind,   // removed by this PR
  tempRelays: urls,
  ...
);
```

in both `mobile/integration_test/e2e/relay_list_admission_test.dart` and `mobile/packages/nostr_sdk/test/integration/temp_relay_lifetime_test.dart`. Both PRs are green today because neither has the other's diff. Whichever merges second will fail to compile until the loser rebases — the fix is trivial (drop the two `eventKind:` lines), but it's the kind of thing that shows up as a confusing red CI run on an approved PR rather than as a conflict marker, since git will merge these hunks cleanly.

No change requested here; just worth whoever merges second knowing to look for it. I've left the same note on #6746.

### Nit

With `eventKind` gone from the tracker, an SDK-level publish log no longer says which kind failed — `_recordPublishBreadth` reports counts and relay URLs only. The app layer still logs kind at its own call sites, so this isn't a real gap, but if a future publish investigation wants kind in the SDK log it's now a re-plumb rather than a field read. Fine as a tradeoff for deleting the scaffold; noting it so the decision is explicit.

Issue #5853, opened 2026-07-05
refactor(account): consolidate pending upload persistence


NotThatKindOfDrLiz:
## Summary

PR #5791 scoped the Drift PendingUploadsDao read/watch APIs by owner and preserved explicit maintenance access, closing #5788. During review, it surfaced a separate pre-existing architecture issue: the production pending-upload flow currently uses the Hive-backed PendingUploadStore, while the Drift PendingUploadsDao read/watch surface is effectively orphaned outside tests and cleanup.

This leaves two parallel pending-upload persistence paths in the codebase, which makes future account-isolation work easier to get wrong.

## Scope

- Decide whether pending uploads should be backed by Drift or Hive going forward.
- If Drift is the target, migrate production pending-upload reads/writes from PendingUploadStore to PendingUploadsDao and keep account scoping explicit.
- If Hive remains the target, delete the unused Drift pending-upload read/watch surface and associated dead tests/API.
- Preserve destructive sign-out cleanup semantics for pending uploads.
- Add/update tests for the chosen persistence path and account isolation behavior.

## Acceptance criteria

- There is one production pending-upload persistence source of truth.
- No orphaned read/watch API remains for pending uploads.
- Account-scoped current-user flows cannot expose another account pending uploads.
- Intentional maintenance or cleanup paths are explicitly named and tested.

Parent epic: #4623
Surfaced by: #5788, #5791

NotThatKindOfDrLiz:
@rabble I reviewed #5853 against current `origin/main` before starting implementation, and I want to confirm the intended architecture before changing persistence code.

The issue is real: there are two pending-upload persistence surfaces, and they are easy to confuse. The part I need guidance on is which one should survive.

What I see on `origin/main` today:

- Production pending-upload reads and writes go through `UploadManager -> PendingUploadStore`.
- `PendingUploadStore` owns the Hive-backed `pending_uploads` box and appears to be the active upload pipeline store.
- The Drift `PendingUploadsDao` exists and was scoped in #5791, but I do not see production writers to that table. Its production use appears limited to destructive sign-out cleanup via `db.pendingUploadsDao.deleteAllForUser(userPubkey)`, plus tests/API surface.
- That means destructive sign-out cleanup appears to target the Drift table, while the real production pending-upload records may remain in Hive. Owner scoping prevents visible cross-account reads, but the cleanup semantics do not appear to hit the active store.

Before I do any coding, what is the desired direction for #4623 / #5853?

[...]

PR #6732, opened 2026-08-05
fix(account): purge Hive pending uploads on destructive cleanup

NotThatKindOfDrLiz:
## Summary
- route owner-scoped Hive pending upload purge through the live UploadManager instead of a throwaway store
- discard departing-owner upload work before deletion so retry/backoff/session/thumbnail callbacks cannot recreate purged rows
- preserve pending_uploads Hive data and lock files during full cache recovery because they are durable local-only upload state
- keep account cleanup moving if an earlier cleanup step fails, so later destructive purges are not skipped

Refs #5853

## Verification
- flutter test test/providers/social_providers_cleanup_test.dart test/services/upload/pending_upload_store_test.dart test/services/upload_manager_owner_scope_test.dart test/services/cache_recovery_service_test.dart
- flutter analyze
- pre-push checks: analyzer, generated verification, changed-file tests

### DIFF
--- a/mobile/lib/providers/social_providers.dart
+++ b/mobile/lib/providers/social_providers.dart
@@ -69,6 +69,12 @@ const _relayRepairTimeout = Duration(seconds: 15);
 
 typedef PendingUploadOwnerCleanup = Future<int> Function(String ownerPubkey);
 
+final openVineImageCacheClearProvider = Provider<Future<void> Function()>((
+  ref,
+) {
+  return clearOpenVineImageCache;
+});
+
 final pendingUploadOwnerCleanupProvider = Provider<PendingUploadOwnerCleanup>((
   ref,
 ) {
@@ -689,7 +695,7 @@ UserDataCleanupService userDataCleanupService(Ref ref) {
           () => ref.read(dmRepositoryProvider).stopListening(),
         );
         try {
-          await clearOpenVineImageCache();
+          await ref.read(openVineImageCacheClearProvider)();
           ref
               .read(adultMediaAccessRevocationVersionProvider.notifier)
               .increment();
--- a/mobile/lib/providers/social_providers.dart
+++ b/mobile/lib/providers/social_providers.dart
@@ -38,6 +38,7 @@ import 'package:openvine/services/pending_action_service.dart';
 import 'package:openvine/services/product_event_queue.dart';
 import 'package:openvine/services/profile_save_retry_service.dart';
 import 'package:openvine/services/social_service.dart';
+import 'package:openvine/services/upload/pending_upload_store.dart';
 import 'package:openvine/services/user_data_cleanup_service.dart';
 import 'package:openvine/services/view_event_publisher.dart';
 import 'package:openvine/services/view_event_retry_service.dart';
@@ -65,6 +66,25 @@ Stream<void> _dmRetryConnectivityTriggerStream() => Connectivity()
 /// must not starve the retry sweep indefinitely — on timeout the sweep runs
 /// and the SDK's own OK-timeout remediation backstops any still-stale socket.
 const _relayRepairTimeout = Duration(seconds: 15);
+
+typedef PendingUploadOwnerCleanup = Future<int> Function(String ownerPubkey);
+
+final pendingUploadOwnerCleanupProvider = Provider<PendingUploadOwnerCleanup>((
+  ref,
+) {
+  return (ownerPubkey) async {
+    final store = PendingUploadStore(
+      scopeUploadsToCurrentUser: false,
+      currentNostrPubkey: null,
+    );
+    try {
+      await store.open();
+      return store.deleteAllForOwner(ownerPubkey);
+    } finally {
+      store.disposeStore();
+    }
+  };
+});
 
 /// Connectivity trigger for the message retry sweep: fires on EVERY
 /// connectivity transition, including `→ none`. [OutgoingDmRetryService] runs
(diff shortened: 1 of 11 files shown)


rabble (changes requested):
Request changes. Feedback-only review (agent-assisted, no branch changes).

The `safeCleanup` refactor is the best thing in this PR and I'd take it on its own. Turning a chain of bare `await`s into individually-guarded steps means a failure in `directMessages` no longer skips every destructive purge behind it — on an account-deletion path that was a real data-retention hazard, and the per-step log line with the owner pubkey makes a partial cleanup diagnosable instead of silent. Scoping the Hive purge by explicit `ownerPubkey` rather than `currentNostrPubkey`, with the doc comment explaining that auth state has already moved on by then, is also correct.

**The blocker: the deferred-save queue purge is wired to the wrong store instance.**

`deleteAllForOwner` opens with exactly the right guard — drop any queued saves belonging to the departing owner before deleting their rows:

```dart
final queueIds = _pendingSaveQueue.entries
    .where((entry) => entry.value.nostrPubkey == ownerPubkey)
    ...
```

But `pendingUploadOwnerCleanupProvider` constructs a **fresh** `PendingUploadStore` to run it against:

```dart
final store = PendingUploadStore(
  scopeUploadsToCurrentUser: false,
  currentNostrPubkey: null,
);
```

A fresh store's `_pendingSaveQueue` is always empty, so that filter is a no-op every time. The queue that actually matters belongs to the live `UploadManager`'s store, and it's untouched. `_processSaveQueue` drains on a 5-second timer (30 on retry) and calls `save(upload)`, which writes straight back to the same Hive box:

1. User with an in-flight upload triggers account deletion.
2. Purge deletes the departing owner's rows from `pending_uploads`.
3. Up to 30 seconds later the live store's timer fires and re-persists the queued upload for the account that was just deleted.

That's the failure #5853 is about, surviving the fix. Not hypothetical either — an in-flight upload at deletion time is the ordinary case, not an edge one.

**The fix is already written.** `UploadManager.deleteAllForOwner` (`upload_manager.dart:372`) delegates to the live store, which is the instance holding the real queue — and it currently has **no caller anywhere in `lib/`**, so as shipped it's dead code that CI's no-tech-debt expectations would flag on its own. Routing the provider through `ref.read(uploadManagerProvider).deleteAllForOwner(ownerPubkey)` closes the hole and removes the dead method in one move.

I'll flag one uncertainty rather than assert it: I haven't traced whether reading `uploadManagerProvider` this late in teardown is safe, or whether it could construct a manager during cleanup. If you avoided the live manager deliberately for a lifecycle reason, say so and clear the live store's queue some other way — but the current shape can't do what the guard was written to do.

Two smaller things, neither blocking:

- Dropping `pending_uploads` from `CacheRecoveryService`'s box list is right on the "this is durable user data, not cache" reasoning, but it does mean a corrupted `pending_uploads` box is no longer auto-recoverable — the recovery path that existed for it is now gone with nothing replacing it. Worth confirming a corrupt box degrades gracefully rather than wedging upload init behind the circuit breaker.
- `openVineImageCacheClearProvider` is introduced as an indirection for testability but the image-cache step is still the one block not wrapped in `safeCleanup` — it keeps its own bespoke try/catch. Folding it in would make the whole sequence uniform.

Happy to re-review as soon as the provider points at the live store.

NotThatKindOfDrLiz:
Takeover update pushed in ef23e2362.

What changed:
- Routed `pendingUploadOwnerCleanupProvider` through the live `uploadManagerProvider` instead of a throwaway `PendingUploadStore`.
- Made `UploadManager.deleteAllForOwner` discard departing-owner upload work before deleting rows: retry backoff/session persistence, progress subscriptions, processing polls, and background uploads are stopped or ignored.
- Added purged-owner guards so late retry/success/failure/thumbnail callbacks do not recreate deleted Hive rows.
- Made cache recovery preserve `pending_uploads.hive` and `pending_uploads.lock` during the full `clearAllCaches()` path.
- Added regression coverage for the real provider body with multiple Hive rows, owner cleanup while the manager is scoped to another account, queue purge behavior, and full cache recovery.
- Updated the PR body to `Refs #5853` instead of closing it, since #5853 is the broader persistence-consolidation issue.

Verification:
- `flutter test test/providers/social_providers_cleanup_test.dart test/services/upload/pending_upload_store_test.dart test/services/upload_manager_owner_scope_test.dart test/services/cache_recovery_service_test.dart`
- `flutter analyze`
- Pre-push hook: merge-conflict check, analyzer, generated verification, changed-file tests

rabble (approved):
Re-review after `ef23e236`. The blocker is closed — approving.

**Verified the fix, not just the description:**

- `pendingUploadOwnerCleanupProvider` now reads the live manager (`ref.read(uploadManagerProvider).deleteAllForOwner(ownerPubkey)`), so the `_pendingSaveQueue` filter inside `deleteAllForOwner` runs against the instance that actually holds queued saves. The throwaway `PendingUploadStore` is gone, and `UploadManager.deleteAllForOwner` is no longer dead code — it's the only path now.
- The lifecycle uncertainty I flagged is answered by the tests: `social_providers_cleanup_test.dart` exercises the real provider body with `uploadManagerProvider.overrideWithValue(...)` and multiple Hive rows, plus a case where the manager is scoped to a different account than the one being purged. That's the shape that would have caught a wrong-instance regression.
- Going further than the finding required and stopping the *sources* of re-persistence — retry backoff, progress subscriptions, processing polls, background uploads — plus purged-owner guards on late retry/success/failure/thumbnail callbacks, is the right depth. Draining the queue alone would have left the callback paths free to write a fresh row seconds later; this closes both halves.
- `pending_uploads.hive` / `pending_uploads.lock` are now preserved through the full `clearAllCaches()` path, which resolves my second note properly. Excluding the box from recovery while leaving `clearAllCaches` free to wipe it would have swapped one data-loss path for another.
- Retitling to `Refs #5853` rather than `Closes` is the honest call given the issue covers the wider persistence consolidation.

The `safeCleanup` refactor remains the quiet win here — a failure in `directMessages` no longer skips every destructive purge behind it, and each step names the owner pubkey in its log line so a partial cleanup is diagnosable rather than silent.

**One leftover nit, non-blocking:** the image-cache step still keeps its own bespoke `try`/`catch` rather than going through `safeCleanup`. That's defensible since it guards two operations (the cache clear and the `adultMediaAccessRevocationVersionProvider` increment) under one catch, which `safeCleanup`'s single-callback shape doesn't express as cleanly. Leaving it is fine; just noting the sequence isn't fully uniform.

CI green on head. Good work on the takeover — the fix went past the reported symptom to the mechanism.

PR #6734, opened 2026-08-05
fix(comments): pad the video reply icon inside its button

### LINKED ISSUE #6735: fix(comments): pad the video reply icon inside its button
Auto-created from PR #6734.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6734
**Branch:** `fix/comment-video-button-icon-padding`

hm21:
Closes #6735

## Description

The camera glyph in the comment bar's record-video button sat almost flush against its circle. The cause is the asset, not the layout: `assets/icon/video_camera.svg` has `viewBox="0 0 29 18"`, and `DivineIcon` renders with `BoxFit.contain` off `width: size, height: size` — so a non-square asset is sized by its height and `size: 22` comes out **35.44px wide**. In the 40px circle that leaves 2.28px of clearance on each side, which reads as a ring drawn around the glyph rather than a button holding it.

- Adds 7px inner padding, capping the glyph at 26px wide and restoring 7px of clearance.
- Changes the radius from 17 to 16, matching `DivineIconButton`'s 40px `small` variant. The design system draws these as rounded squares (16 at 40px, 20 at 48px), never full circles.

The tap target, semantics, and callback wiring are untouched.

## Notes for reviewers

An earlier round of review measured the padding as inert, based on a repro built with a square Material `Icon`. That measurement is correct for a square icon — the icon rect is unchanged and only the `IconButton`'s Material box shrinks — but it does not hold for this asset. Measured on the real `CommentInput`, the glyph box goes from 35.44 × 22 (2.28px clearance) to 26 × 22 (7px clearance). Details in the [thread](https://github.com/divinevideo/divine-mobile/pull/6734#issuecomment-5197136442).

The two sibling buttons in the same input row are unaffected: `CaretDown.svg` and `arrow_up.svg` are both `viewBox="0 0 32 32"`, so they render square with 9–10px of clearance. Their radius is still 17; aligning that is a separate, deliberate call rather than something to fold in here.

## Out of Scope

All three buttons in this row hand-roll what `DivineIconButton(size: .small)` already provides, including the 48px tap target. Replacing them is a worthwhile cleanup but a much larger diff than this fix.

## Verification

- `flutter analyze lib/screens/comments test/screens/comments` — clean
- `flutter test test/screens/comments/` — 122/122 pass
- `dart format` — clean
- New regression test asserts the glyph keeps ≥6px of clearance inside its circle; confirmed it fails at 2.28px when the padding is removed
- Padding checked visually on a real Samsung Galaxy device; padding plus the radius change re-checked on macOS

## Type of Change

### DIFF
--- a/mobile/lib/screens/comments/widgets/comment_input.dart
+++ b/mobile/lib/screens/comments/widgets/comment_input.dart
@@ -293,7 +293,7 @@ class _VideoReplyButton extends StatelessWidget {
           child: Container(
             width: 40,
             height: 40,
-            margin: const .only(bottom: 4),
+            margin: const EdgeInsets.only(bottom: 4),
             decoration: BoxDecoration(
               color: context.vineColors.containerLow,
               borderRadius: BorderRadius.circular(20),
--- a/mobile/lib/screens/comments/widgets/comment_input.dart
+++ b/mobile/lib/screens/comments/widgets/comment_input.dart
@@ -294,7 +294,6 @@ class _VideoReplyButton extends StatelessWidget {
             width: 40,
             height: 40,
             margin: const .only(bottom: 4),
-            padding: const .all(7),
             decoration: BoxDecoration(
               color: context.vineColors.containerLow,
               borderRadius: BorderRadius.circular(20),
--- a/mobile/lib/screens/comments/widgets/comment_input.dart
+++ b/mobile/lib/screens/comments/widgets/comment_input.dart
@@ -293,10 +293,11 @@ class _VideoReplyButton extends StatelessWidget {
           child: Container(
             width: 40,
             height: 40,
-            margin: const EdgeInsets.only(bottom: 4),
+            margin: const .only(bottom: 4),
+            padding: const .all(7),
             decoration: BoxDecoration(
               color: context.vineColors.containerLow,
-              borderRadius: BorderRadius.circular(17),
+              borderRadius: BorderRadius.circular(20),
             ),
             child: IconButton(
               onPressed: onPressed,
--- a/mobile/lib/screens/comments/widgets/comment_input.dart
+++ b/mobile/lib/screens/comments/widgets/comment_input.dart
@@ -294,9 +294,13 @@ class _VideoReplyButton extends StatelessWidget {
             width: 40,
             height: 40,
             margin: const EdgeInsets.only(bottom: 4),
+            // The camera asset is 29x18, so `contain` sizes it off its height
+            // and it renders 35px wide inside the 40px circle. The padding
+            // caps that width; without it the glyph sits 2px from the edge.
+            padding: const EdgeInsets.all(7),
             decoration: BoxDecoration(
               color: context.vineColors.containerLow,
-              borderRadius: BorderRadius.circular(20),
+              borderRadius: BorderRadius.circular(16),
             ),
             child: IconButton(
               onPressed: onPressed,
--- a/mobile/test/screens/comments/comment_input_test.dart
+++ b/mobile/test/screens/comments/comment_input_test.dart
@@ -105,6 +105,43 @@ void main() {
       expect(tapped, isTrue);
     });
 
+    testWidgets('video reply glyph stays inset from its circle', (
+      tester,
+    ) async {
+      // The camera asset is 29x18, so BoxFit.contain sizes it off its height
+      // and it renders far wider than the requested 22. Unpadded it spans
+      // 35.4px inside the 40px circle, leaving 2.3px of clearance and reading
+      // as a ring drawn around the glyph.
+      await tester.pumpWidget(
+        MaterialApp(
+          localizationsDelegates: AppLocalizations.localizationsDelegates,
+          supportedLocales: AppLocalizations.supportedLocales,
+          home: Scaffold(
+            body: CommentInput(
+              controller: controller,
+              onSubmit: () {},
+              onVideoReplyPressed: () {},
+            ),
+          ),
+        ),
+      );
+
+      final glyph = tester.getRect(_divineIcon(DivineIconName.videoCamera));
+      final circle = tester.getRect(
+        find
+            .ancestor(
+              of: _divineIcon(DivineIconName.videoCamera),
+              matching: find.byWidgetPredicate(
+                (widget) =>
+                    widget is Container && widget.constraints?.maxWidth == 40,
+              ),
+            )
+            .first,
+      );
+
+      expect((circle.width - glyph.width) / 2, greaterThanOrEqualTo(6));
+    });
+
     testWidgets(
       'never shows a CircularProgressIndicator on the send button',
       (tester) async {


mbradley (changes requested):
I measured this one before commenting, because the whole change is a visual claim.

The padding does not inset the glyph. `IconButton` centers its icon inside whatever box it gets, so the 40x40 container already left 9px of clearance on every side. Adding 7px shrinks the button's box to 26x26 and the icon re-centers to the same 9px. The rendered icon rect is identical with and without it:

| | before | after |
|---|---|---|
| Icon rect | `L=389 T=287 R=411 B=309` (22x22) | identical |
| IconButton / Material box | 40x40 | 26x26 |

What it does change is the `IconButton`'s own Material box, which is the ink surface. The press ripple now covers 26px inside a 40px circle instead of filling it, so the net effect is a slightly worse button rather than a better one. The tap target is unaffected either way, since `RenderInputPadding` still expands the hit region to 48dp.

The radius change is the real fix. `circular(17)` on a 40x40 box leaves a flat mid-edge, so the button was a squircle; 20 makes it an actual circle. That alone resolves what the description is pointing at, and it is now the entire diff.

I pushed two commits:

- `42bf107c1` drops the padding and keeps the radius.
- `b974de299` restores `EdgeInsets.only` on the margin. The dot-shorthand conversion was not needed for the fix, and it is the only one in `lib/screens/comments/`, three lines above two siblings still using the long form.

Validation on the branch: `flutter analyze lib/screens/comments` clean, `flutter test test/screens/comments/` 121/121, and all CI checks green on `b974de299`.

Two things I can't decide for you:

**1. What did you actually see on the device?** A 22px glyph in a 40px box can't touch the edges, so the description's premise doesn't hold, but something prompted this. If it was the squircle shape, the radius fix covers it. If it was that the glyph reads too large inside the circle, that's a different change (reducing the icon from 22, not padding the container) and worth saying so I haven't stripped something you wanted.

**2. The two siblings are still at `circular(17)`.** `_KeyboardDismissButton` and `_SendButton` sit in the same Row, same 40x40, same squircle. As it stands this PR makes one of three adjacent circles round and leaves two squared off, which is more visible side by side than the original problem. Either apply 20 to all three here, or leave it and open a follow-up. Your call, but a mixed row is worth deciding deliberately rather than by default.

Worth noting for whenever someone next touches this file: all three buttons hand-roll what `DivineIconButton(size: .small)` already provides, including the 48px tap target. Not for this PR.

Please review the two commits and merge when you're happy, or push back if I've read the intent wrong.

hm21:
Thanks for measuring, but the repro used a square icon, and that's where it diverges.

The button renders `DivineIcon(icon: .videoCamera, size: 22)`, and `assets/icon/video_camera.svg` has `viewBox="0 0 29 18"`. `DivineIcon` passes `width: size, height: size` with `BoxFit.contain`, which sizes a non-square asset off its height — so `size: 22` renders **22 × 29/18 = 35.44px wide**, not 22.

Measured on the real `CommentInput`:

| | glyph box | side clearance in the 40px circle |
|---|---|---|
| without padding | 35.44 × 22 | **2.28px** |
| with `EdgeInsets.all(7)` | 26 × 22 | 7px |

Your `L=389 T=287 R=411 B=309` (22×22) is exactly what a square `Icon` produces — I reproduced your numbers to the pixel with `Icons.videocam` before switching the probe to `DivineIcon`. The measurement is right, just of a different widget. Padding restored in `8345a19de`.

Your ink-surface point still holds in principle — the Material box does shrink to 26 — but that's the price of capping a glyph that otherwise reaches 2px from the edge. If we want both, it's `DivineIconButton(size: .small)`, as you noted.

On the radius I went to **16**, not 20. `DivineIconButton` uses 16 for its 40px `small` and 20 for the 48px `base`, so the design system draws these as rounded squares, never full circles — 20 would have made this the only true circle in the row.

On "three adjacent circles": `_VideoReplyButton` renders only when `!_hasText`, `_SendButton` only when `_hasText`, so those two are mutually exclusive — at most two are ever on screen. And the siblings don't have this problem: `CaretDown.svg` and `arrow_up.svg` are both `viewBox="0 0 32 32"`, so they render square at 22 and 20 with 9–10px clearance. Nothing to align on the padding side; the radius mismatch is a separate call I'd rather not widen this PR for.

Added a regression test asserting ≥6px clearance inside the circle — it fails at 2.28px without the padding, so this can't be reverted on inspection again.

rabble (commented):
Feedback-only review (agent-assisted, no branch changes). I ran the measurements rather than reading the diff, because two people have posted contradictory numbers and that's the only thing that settles it.

**Both of you are right about different boxes, and the current diff fixes one problem by creating the other.**

I checked out the branch head, ran `video reply glyph stays inset from its circle`, then removed only the `padding: const EdgeInsets.all(7)` line and re-ran. Instrumented with the actual rects:

| | padding removed | padding present (head) |
|---|---|---|
| `DivineIcon` rect | **35.4 × 22.0** | **26.0 × 22.0** |
| `IconButton` box (ink surface) | **40 × 40** | **26 × 26** |
| Clearance each side | 2.3px | 7.0px |

The test passes with the padding and fails without it.

@hm21's mechanism is confirmed. `DivineIcon` renders `SvgPicture.asset(width: size, height: size, fit: fit)`, and inside this `IconButton` the width isn't binding — the box takes the asset's aspect off its pinned height. 22 × 29/18 = 35.44, which is the 35.4 measured. So the glyph really does span 35.4px inside a 40px circle unpadded, leaving 2.3px, and the padding really does cap it to 26. The `d`escription's claim is accurate.

@mbradley's ripple objection is also confirmed by the same run, and it isn't a side note. The `IconButton`'s Material box — the ink surface — goes from 40 × 40 to 26 × 26. The press ripple now covers a 26px circle floating inside a 40px one. That's a real visual regression, just a different one from the one being fixed.

I can't reconcile the `L=389 T=287 R=411 B=309` (22 × 22) figure with anything I measured; no box in this subtree is 22 × 22 in either configuration. Most likely a different finder or an earlier revision — worth re-running before treating it as settled, since it's the load-bearing number in the case for dropping the padding.

**The synthesis neither version has:** constrain the glyph without shrinking the ink surface. Padding the `Container` does both because the `IconButton` is the padded child. Sizing the icon directly — an explicit width constraint on the `DivineIcon`, or `IconButton(iconSize:)` — insets the glyph while the `IconButton` keeps the full 40 × 40 Material box. That gets @hm21's 7px clearance and @mbradley's full-circle ripple at the same time, and the existing test would still pin it.

**The radius is a separate, unresolved question, and it moved the wrong way.** @mbradley's geometry is right: on a 40 × 40 box, `circular(17)` leaves a flat mid-edge and anything ≥ 20 is a true circle. Head now has `circular(16)`, which is *further* from a circle than the 17 it replaced. The commit message says "use the design-system radius", so I read this as a deliberate re-spec from circle to rounded-square rather than an error — but that's a design decision, not a rendering fact, and the PR description still frames the change as fixing a squircle. Worth an explicit call from @rabble or design on which shape this button is supposed to be, because the two commits are pulling in opposite directions on it.

**For a change that is entirely a visual claim, the missing instrument is a golden.** The repo has `scripts/golden.sh` and `docs/GOLDEN_TESTING_GUIDE.md`, and neither the original PR nor the takeover added one for this button. A golden would have made this whole exchange a diff image instead of duelling `getRect` calls, and it would keep the ripple regression visible too — which the current numeric test cannot see, since it only asserts glyph clearance.

Not casting a vote — @mbradley's `CHANGES_REQUESTED` is the blocking review and the ripple question is his to close. Measurements above are reproducible on the branch head if either of you wants to check them.

mbradley (approved):
Measurement conceded. The camera asset is `viewBox="0 0 29 18"` and `DivineIcon` sizes off the pinned height under `BoxFit.contain`, so `size: 22` really does render 35.4px wide. I verified the viewBox and the `SvgPicture.asset(width:, height:, fit:)` call directly, and I ran your new test both ways: green as written, red at 2.28px with the padding pulled. My revert was measured on a square `Icons.videocam` rather than the real `DivineIcon`, which is a substitution I allowed in the probe and shouldn't have. Your original description was right.

The ink-surface point stands but I'm not blocking on it. The `IconButton`'s Material box is still 40 → 26, so the ripple floats inside the circle. @rabble's `iconSize` route would cap the glyph while leaving the Material box at 40, and your new test would still pin it. Worth doing whenever this button is next touched; not worth holding a one-line visual fix for.

On the row: agreed that `_VideoReplyButton` and `_SendButton` are mutually exclusive, so my "three circles" framing was wrong. 16 also leaves a smaller mismatch against the dismiss button's 17 than my 20 would have.

PR #3516, opened 2026-04-28
feat(video-editor-meta): redesign the metadata screen

hm21:
## Description

Redesign the metadata screen to the latest design in figma including the option to select the cover thumbnail.

Bildschirmfoto 2026-04-28 um 16 22 03

**Related Issue:** Closes #3703 Closes #3517 Closes #1922

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #1395, opened 2026-02-08
fix: comments sort toggle overflow by 2.8px

realmeylisdev:
## Summary
- Removes the hardcoded `SizedBox(width: 62)` constraint on the trailing widget in `VineBottomSheetHeader`
- The "New" sort label (icon + text + padding) needs ~65px, causing a 2.8px RenderFlex overflow
- The trailing widget now sizes itself naturally via `MainAxisSize.min`

Fixes #1394

Simulator Screenshot - iPhone 17 Pro Max - 2026-02-08 at 18 54 23
Simulator Screenshot - iPhone 17 Pro Max - 2026-02-08 at 18 54 25
Simulator Screenshot - iPhone 17 Pro Max - 2026-02-08 at 18 54 26

## Test plan
- [ ] Open comments sheet, cycle through all sort modes (New / Top / Old) — no overflow
- [ ] Verify sort toggle stays right-aligned in the header

Issue #1394, opened 2026-02-08
RenderFlex overflow in comments sort toggle button


meylis1998:
## Bug Description

A `RenderFlex overflowed by 2.8 pixels on the right` error occurs in the comments screen sort toggle button (the "New" / "Top" / "Old" cycle button in the header).

## Error

```
════════ Exception caught by rendering library ═════════════════════════════════
A RenderFlex overflowed by 2.8 pixels on the right.
The relevant error-causing widget was:
    Row Row:file:///...mobile/lib/screens/comments/comments_screen.dart:380:20
════════════════════════════════════════════════════════════════════════════════
```

## Root Cause

The `_CommentsSortToggle` widget (`comments_screen.dart:380`) contains a `Row` with an icon (14px) + 4px spacing + text label, all inside a `Container` with 10px horizontal padding on each side.

This Row is placed as the `trailing` widget of `VineBottomSheetHeader`, which constrains it to a fixed `SizedBox(width: 62)` (see `vine_bottom_sheet_header.dart:74`).

When the sort mode is "New" (the widest label), the total width of icon + spacing + text + padding exceeds 62px by ~2.8 pixels, triggering the overflow.

The "Top" and "Old" labels fit within the constraint, so the overflow only occurs on the "New" sort mode.

## Screenshots

| "New" (overflows) | "Top" (ok) | "Old" (ok) |
|---|---|---|
| ![new](https://github.com/user-attachments/assets/placeholder) | ![top](https://github.com/user-attachments/assets/placeholder) | ![old](https://github.com/user-attachments/assets/placeholder) |

*Note: Red overflow indicator visible on the right edge of the "New" button in the first screenshot.*

## Suggested Fix

Either:
1. Increase the `SizedBox(width: 62)` constraint in `VineBottomSheetHeader` to accommodate the widest label, or
2. Use shorter labels (e.g., abbreviate), or
3. Use `Flexible`/`Expanded` instead of a fixed-width `SizedBox` for the trailing widget, or
4. Reduce horizontal padding in the sort toggle `Container`

## Environment

- Device: iPhone 17 Pro Max (Simulator)
- Flutter: latest stable
- File: `mobile/lib/screens/comments/comments_screen.dart:380`
- Component: `VineBottomSheetHeader` trailing widget constraint at `mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart:74`

santiizaz:
Testing details: Qa pass
[...]

PR #6736, opened 2026-08-05
fix(ui): keep sheet header titles centered

### LINKED ISSUE #6739: fix(ui): reserve the opposite slot's width in sheet headers
Auto-created from PR #6736.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6736
**Branch:** `fix/bottom-sheet-header-title-centering`

hm21:
Closes #6739

## Description

`VineBottomSheetHeader` balanced a filled leading/trailing slot with a hardcoded `SizedBox(width: 40, height: 40)`, so the title only landed dead center when the other side happened to be exactly 40px wide.

The comments sheet has no leading slot and a sort chip whose width follows its label, so the title visibly drifted on every sort-mode change (New / Top / Old — and much further apart in locales with longer words). The single-action sheets (community suggest, banner colors) sat 4px off too, and the two-slot headers (content-warning selector, tags picker, and the user picker when `maxCount != null`) sat 2px off, because a `small/secondary` `DivineIconButton` lays out at 48px while the other 13 size/type combinations are 52px.

The header now lays its row out itself: a small render object puts the slots at the edges, centers the title on the full width, and keeps both sides clear by capping the title at `width - 2 * widestSlot - 2 * gap`. Centering no longer depends on the two slots being equally wide, so the two-slot headers are fixed by the same change.

Capping rather than reserving matters for overflow. An earlier revision of this PR reserved the filled slot's width on the empty side by re-rendering the slot hidden. That centers exactly too, but it charges the row twice for one slot — its fixed cost goes from `W + 64` to `2W + 24` — and the comments sort chip, the only slot in the app whose width comes from text, then overflowed the header at accessibility text scales on narrow screens, where the old 40px placeholder did not. With the title as the only element that gives, the row cannot overflow at any scale.

## Out of Scope

Where one slot grows so wide that keeping the same width clear on the empty side would leave the title no room at all, centering is geometrically impossible: two copies of the slot plus the gaps already exceed the row. The title then falls back to the space between the slots, the way it sat before this PR, rather than collapsing to zero width. This is only reachable with a text-driven slot at high accessibility text scales on a 320-390dp screen.

## Verification

- `flutter test --coverage` in `mobile/packages/divine_ui`: 840 passing, package still at 100.0000% line coverage (the strict gate)
- `flutter test test/widgets test/screens`: 4639 passing
- `flutter analyze lib test packages/divine_ui`: no issues
- Header tests cover centering with a leading-only slot, a trailing-only slot, and two slots of different widths; that the title stays put as the trailing width changes 60px -> 160px; RTL; that a wide slot caps the title instead of overflowing the row; the between-the-slots fallback in both directions; hit testing; semantics; and the intrinsic/dry-layout protocol
- Each of those guards was mutation-checked rather than just run green: centering between the slots instead of on the full width turns 7 tests red, dropping RTL turns 1 red, dropping the fallback turns 1 red
- The mirror's collateral is gone with it — each slot is built once, so slot finders match one widget again and `vine_bottom_sheet_test.dart` and `banner_color_swatches_test.dart` are back to their state on `main`
- Checked on a real Samsung Galaxy: comments sheet title stays centered across all three sort modes. Also run on macOS across the affected sheets.

## Type of Change

### DIFF
--- a/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
+++ b/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
@@ -1,25 +1,30 @@
 // ABOUTME: Header component for VineBottomSheet
 // ABOUTME: Displays title with optional trailing actions (badges, buttons)
 
+import 'dart:math' as math;
+
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
 
 /// Header component for [VineBottomSheet].
 ///
 /// Combines drag handle and title section as per Figma design.
 /// Uses Bricolage Grotesque bold font at 24px for title.
 ///
-/// When a title is shown and only one of the leading/trailing slots is filled,
-/// the empty side reserves the filled one's width so the title stays centered.
-/// That slot widget is therefore built twice, so it must carry no
-/// [GlobalKey] and no `initState` side effects — analytics, fetches,
-/// subscriptions and timers would all fire a second time for the hidden copy.
+/// The title is centered on the full header width, never between the
+/// leading/trailing slots, so it does not move when a slot changes width — a
+/// sort chip relabelled New → Top, or two icon buttons whose variants happen
+/// to differ by a couple of pixels. Both sides are kept clear by capping the
+/// title at `width - 2 * widestSlot - 2 * gap` instead of reserving layout
+/// space opposite the filled slot, so a wide slot costs the title its own
+/// width once rather than twice and the row cannot overflow.
 ///
-/// Because the slot is in the tree twice, widget finders that match it
-/// (`find.byKey`, `byType`, `byTooltip`) return two results: the hidden copy
-/// comes first when [trailing] is the filled slot, and second when [leading]
-/// is. Semantics finders are unaffected — the copy is excluded from the
-/// semantics tree.
+/// If a slot grows so wide that keeping the same width clear on the empty
+/// side would leave the title no room at all — a text-driven slot at
+/// accessibility text scales on a narrow screen — the title falls back to the
+/// space between the slots. Centering is geometrically impossible at that
+/// point; the alternative is a title squeezed to zero width.
 class VineBottomSheetHeader extends StatelessWidget {
   /// Creates a [VineBottomSheetHeader] with the given title and optional
   /// leading and trailing widgets.
@@ -96,35 +101,18 @@ class VineBottomSheetHeader extends StatelessWidget {
 
               Padding(
                 padding: .only(bottom: hasHeaderRow ? 14 : 0),
-                child: Row(
-                  mainAxisAlignment: .spaceBetween,
-                  spacing: 12,
-                  children: [
-                    if (leadingSlot != null)
-                      leadingSlot
-                    else if (hasTitle && trailingSlot != null)
-                      _SlotMirror(trailingSlot),
-
-                    if (hasTitle)
-                      Flexible(
-                        child: Center(
-                          child: DefaultTextStyle(
-                            style: VineTheme.titleMediumFont(
-                              color: context.vineColors.onSurface,
-                            ),
-                            textAlign: .center,
-                            child: title!,
+                child: _HeaderRow(
+                  leading: leadingSlot,
+                  trailing: trailingSlot,
+                  title: hasTitle
+                      ? DefaultTextStyle(
+                          style: VineTheme.titleMediumFont(
+                            color: context.vineColors.onSurface,
                           ),
-                        ),
-                      )
-                    else
-                      const Spacer(),
-
-                    if (trailingSlot != null)
-                      trailingSlot
-                    else if (hasTitle && leadingSlot != null)
-                      _SlotMirror(leadingSlot),
-                  ],
+                          textAlign: .center,
+                          child: title!,
+                        )
+                      : null,
                 ),
               ),
             ],
@@ -143,28 +131,252 @@ class VineBottomSheetHeader extends StatelessWidget {
   }
 }
 
-/// Reserves the width of the opposite header slot so the title stays centered
-/// when only one side is filled.
-///
-/// A fixed-size box cannot do this: slot widgets size themselves, so a sort
-/// chip is as wide as its current label and a [DivineIconButton] lays out at
-/// 48-52px depending on its type, never the 40px a constant placeholder used
-/// to reserve. Mirroring the filled slot
-/// always reserves exactly the right width. The copy is laid out but never
-/// painted, focusable, tappable, or announced to screen readers.
-class _SlotMirror extends StatelessWidget {
-  const _SlotMirror(this.child);
+enum _HeaderSlot { leading, title, trailing }
 
-  final Widget child;
+/// Lays out the header's leading slot at the start, the trailing slot at the
+/// end, and the title centered on the full width.
+///
+/// A [Row] cannot do this. Centering there means balancing the filled slot
+/// with an equally wide box on the empty side, which charges the title twice
+/// for one slot and overflows once the two copies plus the gaps exceed the
+/// row. Positioning the title independently of the slots costs nothing and
+/// stays exact when the slots differ in width.
+class _HeaderRow
+    extends SlottedMultiChildRenderObjectWidget<_HeaderSlot, RenderBox> {
+  const _HeaderRow({this.leading, this.title, this.trailing});
+
+  final Widget? leading;
+  final Widget? title;
+  final Widget? trailing;
 
   @override
-  Widget build(BuildContext context) {
-    return Visibility(
-      visible: false,
-      maintainState: true,
-      maintainAnimation: true,
-      maintainSize: true,
-      child: child,
+  Iterable<_HeaderSlot> get slots => _HeaderSlot.values;
+
+  @override
+  Widget? childForSlot(_HeaderSlot slot) => switch (slot) {
+    _HeaderSlot.leading => leading,
+    _HeaderSlot.title => title,
+    _HeaderSlot.trailing => trailing,
+  };
+
+  @override
+  _RenderHeaderRow createRenderObject(BuildContext context) =>
+      _RenderHeaderRow(textDirection: Directionality.of(context));
+
+  @override
+  void updateRenderObject(BuildContext context, _RenderHeaderRow renderObject) {
+    renderObject.textDirection = Directionality.of(context);
+  }
+}
+
+class _RenderHeaderRow extends RenderBox
+    with SlottedContainerRenderObjectMixin<_HeaderSlot, RenderBox> {
+  _RenderHeaderRow({required TextDirection textDirection})
+    : _textDirection = textDirection;
+
+  /// Minimum gap between the title and either slot.
+  static const _gap = 12.0;
+
+  TextDirection get textDirection => _textDirection;
+  TextDirection _textDirection;
+  set textDirection(TextDirection value) {
+    if (_textDirection == value) {
+      return;
+    }
+    _textDirection = value;
+    markNeedsLayout();
+  }
+
+  RenderBox? get _leading => childForSlot(_HeaderSlot.leading);
+  RenderBox? get _title => childForSlot(_HeaderSlot.title);
+  RenderBox? get _trailing => childForSlot(_HeaderSlot.trailing);
+
+  // Reading order, which is also paint and semantics traversal order.
+  @override
+  Iterable<RenderBox> get children => <RenderBox>[
+    ?_leading,
+    ?_title,
+    ?_trailing,
+  ];
+
+  /// Width each side keeps clear of the title, so the centered title clears
+  /// the wider slot on both sides.
+  double _sideInset(double leadingWidth, double trailingWidth) {
+    final widest = math.max(leadingWidth, trailingWidth);
+    return widest == 0 ? 0 : widest + _gap;
+  }
+
+  Size _computeLayout(
+    BoxConstraints constraints,
+    ChildLayouter layoutChild, {
+    bool positionChildren = false,
+  }) {
+    final leading = _leading;
+    final title = _title;
+    final trailing = _trailing;
+    final slotConstraints = BoxConstraints.loose(
+      Size(constraints.maxWidth, double.infinity),
     );
+
+    final leadingSize = leading == null
+        ? Size.zero
+        : layoutChild(leading, slotConstraints);
+    final trailingSize = trailing == null
+        ? Size.zero
+        : layoutChild(trailing, slotConstraints);
+
+    final inset = _sideInset(leadingSize.width, trailingSize.width);
+    // Unbounded width only happens through misuse (a header in an
+    // unconstrained Row); there is always room to center in that case.
+    final centeredBudget = constraints.hasBoundedWidth
+        ? constraints.maxWidth - inset * 2
+        : double.infinity;
+    // One slot can be wide enough that keeping the same width clear on the
+    // empty side leaves the title nothing. Centering is then impossible
+    // rather than merely tight, so the title takes the space between the
+    // slots instead of collapsing to zero.
+    final canCenter = centeredBudget > 0;
+    final leadStart = leadingSize.width + (leading == null ? 0 : _gap);
+    final titleSize = title == null
+        ? Size.zero
+        : layoutChild(
+            title,
+            BoxConstraints(
+              maxWidth: canCenter
+                  ? centeredBudget
+                  : math.max<double>(
+                      0,
+                      constraints.maxWidth -
+                          leadStart -
+                          trailingSize.width -
+                          (trailing == null ? 0 : _gap),
+                    ),
+            ),
+          );
+
+    final size = constraints.constrain(
+      Size(
+        constraints.hasBoundedWidth
+            ? constraints.maxWidth
+            : titleSize.width + inset * 2,
+        math.max(
+          titleSize.height,
+          math.max(leadingSize.height, trailingSize.height),
+        ),
+      ),
+    );
+
+    if (positionChildren) {
+      final isRtl = textDirection == TextDirection.rtl;
+      if (leading != null) {
+        _positionChild(
+          leading,
+          isRtl ? size.width - leadingSize.width : 0,
+          size.height - leadingSize.height,
+        );
+      }
+      if (trailing != null) {
+        _positionChild(
+          trailing,
+          isRtl ? 0 : size.width - trailingSize.width,
+          size.height - trailingSize.height,
+        );
+      }
+      if (title != null) {
+        final titleX = canCenter
+            ? (size.width - titleSize.width) / 2
+            : (isRtl ? size.width - leadStart - titleSize.width : leadStart);
+        _positionChild(title, titleX, size.height - titleSize.height);
+      }
+    }
+
+    return size;
+  }
+
+  /// Positions [child] at [x], vertically centered in the remaining [freeY].
+  void _positionChild(RenderBox child, double x, double freeY) {
+    (child.parentData! as BoxParentData).offset = Offset(x, freeY / 2);
+  }
+
+  @override
+  void performLayout() {
+    size = _computeLayout(
+      constraints,
+      ChildLayoutHelper.layoutChild,
+      positionChildren: true,
+    );
+  }
+
+  @override
+  Size computeDryLayout(BoxConstraints constraints) =>
+      _computeLayout(constraints, ChildLayoutHelper.dryLayoutChild);
+
+  double _intrinsicInset(double height) => _sideInset(
+    _leading?.getMaxIntrinsicWidth(height) ?? 0,
+    _trailing?.getMaxIntrinsicWidth(height) ?? 0,
+  );
+
+  @override
+  double computeMinIntrinsicWidth(double height) =>
+      _intrinsicInset(height) * 2 + (_title?.getMinIntrinsicWidth(height) ?? 0);
+
+  @override
+  double computeMaxIntrinsicWidth(double height) =>
+      _intrinsicInset(height) * 2 + (_title?.getMaxIntrinsicWidth(height) ?? 0);
+
+  double _intrinsicHeight(
+    double width,
+    double Function(RenderBox child, double width) intrinsic,
+  ) {
+    final titleWidth = width.isFinite
+        ? math.max<double>(0, width - _intrinsicInset(double.infinity) * 2)
+        : width;
+    var height = _title == null ? 0.0 : intrinsic(_title!, titleWidth);
+    for (final slot in <RenderBox?>[_leading, _trailing]) {
+      if (slot != null) {
+        height = math.max(height, intrinsic(slot, double.infinity));
+      }
+    }
+    return height;
+  }
+
+  @override
+  double computeMinIntrinsicHeight(double width) => _intrinsicHeight(
+    width,
+    (child, width) => child.getMinIntrinsicHeight(width),
+  );
+
+  @override
+  double computeMaxIntrinsicHeight(double width) => _intrinsicHeight(
+    width,
+    (child, width) => child.getMaxIntrinsicHeight(width),
+  );
+
+  @override
+  void paint(PaintingContext context, Offset offset) {
+    for (final child in children) {
+      context.paintChild(
+        child,
+        (child.parentData! as BoxParentData).offset + offset,
+      );
+    }
+  }
+
+  @override
+  bool hitTestChildren(BoxHitTestResult result, {required Offset position}) {
+    // Front to back, so the topmost child wins a shared pixel.
+    for (final child in children.toList().reversed) {
+      final childOffset = (child.parentData! as BoxParentData).offset;
+      final isHit = result.addWithPaintOffset(
+        offset: childOffset,
+        position: position,
+        hitTest: (result, transformed) =>
+            child.hitTest(result, position: transformed),
+      );
+      if (isHit) {
+        return true;
+      }
+    }
+    return false;
   }
 }
--- a/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
+++ b/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
@@ -11,7 +11,9 @@ import 'package:flutter/material.dart';
 ///
 /// When a title is shown and only one of the leading/trailing slots is filled,
 /// the empty side reserves the filled one's width so the title stays centered.
-/// That slot widget is therefore built twice — keep [GlobalKey]s out of it.
+/// That slot widget is therefore built twice, so it must carry no
+/// [GlobalKey] and no `initState` side effects — analytics, fetches,
+/// subscriptions and timers would all fire a second time for the hidden copy.
 class VineBottomSheetHeader extends StatelessWidget {
   /// Creates a [VineBottomSheetHeader] with the given title and optional
   /// leading and trailing widgets.
--- a/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
+++ b/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
@@ -139,8 +139,9 @@ class VineBottomSheetHeader extends StatelessWidget {
 /// when only one side is filled.
 ///
 /// A fixed-size box cannot do this: slot widgets size themselves, so a sort
-/// chip is as wide as its current label and a [DivineIconButton] is 48px, not
-/// the 40px a constant placeholder used to reserve. Mirroring the filled slot
+/// chip is as wide as its current label and a [DivineIconButton] lays out at
+/// 48-52px depending on its type, never the 40px a constant placeholder used
+/// to reserve. Mirroring the filled slot
 /// always reserves exactly the right width. The copy is laid out but never
 /// painted, focusable, tappable, or announced to screen readers.
 class _SlotMirror extends StatelessWidget {
--- a/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
+++ b/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
@@ -14,6 +14,12 @@ import 'package:flutter/material.dart';
 /// That slot widget is therefore built twice, so it must carry no
 /// [GlobalKey] and no `initState` side effects — analytics, fetches,
 /// subscriptions and timers would all fire a second time for the hidden copy.
+///
+/// Because the slot is in the tree twice, widget finders that match it
+/// (`find.byKey`, `byType`, `byTooltip`) return two results: the hidden copy
+/// comes first when [trailing] is the filled slot, and second when [leading]
+/// is. Semantics finders are unaffected — the copy is excluded from the
+/// semantics tree.
 class VineBottomSheetHeader extends StatelessWidget {
   /// Creates a [VineBottomSheetHeader] with the given title and optional
   /// leading and trailing widgets.
--- a/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
+++ b/mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
@@ -8,6 +8,10 @@ import 'package:flutter/material.dart';
 ///
 /// Combines drag handle and title section as per Figma design.
 /// Uses Bricolage Grotesque bold font at 24px for title.
+///
+/// When a title is shown and only one of the leading/trailing slots is filled,
+/// the empty side reserves the filled one's width so the title stays centered.
+/// That slot widget is therefore built twice — keep [GlobalKey]s out of it.
 class VineBottomSheetHeader extends StatelessWidget {
   /// Creates a [VineBottomSheetHeader] with the given title and optional
   /// leading and trailing widgets.
@@ -56,10 +60,10 @@ class VineBottomSheetHeader extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final hasTitle = title != null && title is! SizedBox;
-    final hasLeadingSlot = leadingAction != null || leading != null;
-    final hasTrailingSlot = trailingAction != null || trailing != null;
-    final hasHeaderRow = hasTitle || hasLeadingSlot || hasTrailingSlot;
-    const placeholder = SizedBox(width: 40, height: 40);
+    final leadingSlot = leadingAction ?? leading;
+    final trailingSlot = trailingAction ?? trailing;
+    final hasHeaderRow =
+        hasTitle || leadingSlot != null || trailingSlot != null;
 
     return Column(
       mainAxisSize: MainAxisSize.min,
@@ -88,12 +92,10 @@ class VineBottomSheetHeader extends StatelessWidget {
                   mainAxisAlignment: .spaceBetween,
                   spacing: 12,
                   children: [
-                    if (leadingAction != null)
-                      leadingAction!
-                    else if (leading != null)
-                      leading!
-                    else if (hasTrailingSlot)
-                      placeholder,
+                    if (leadingSlot != null)
+                      leadingSlot
+                    else if (hasTitle && trailingSlot != null)
+                      _SlotMirror(trailingSlot),
 
                     if (hasTitle)
                       Flexible(
@@ -110,12 +112,10 @@ class VineBottomSheetHeader extends StatelessWidget {
                     else
                       const Spacer(),
 
-                    if (trailingAction != null)
-                      trailingAction!
-                    else if (trailing != null)
-                      trailing!
-                    else if (hasLeadingSlot)
-                      placeholder,
+                    if (trailingSlot != null)
+                      trailingSlot
+                    else if (hasTitle && leadingSlot != null)
+                      _SlotMirror(leadingSlot),
                   ],
                 ),
               ),
@@ -131,6 +131,31 @@ class VineBottomSheetHeader extends StatelessWidget {
             color: context.vineColors.surfaceContainer,
           ),
       ],
+    );
+  }
+}
+
+/// Reserves the width of the opposite header slot so the title stays centered
+/// when only one side is filled.
+///
+/// A fixed-size box cannot do this: slot widgets size themselves, so a sort
+/// chip is as wide as its current label and a [DivineIconButton] is 48px, not
+/// the 40px a constant placeholder used to reserve. Mirroring the filled slot
+/// always reserves exactly the right width. The copy is laid out but never
+/// painted, focusable, tappable, or announced to screen readers.
+class _SlotMirror extends StatelessWidget {
+  const _SlotMirror(this.child);
+
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    return Visibility(
+      visible: false,
+      maintainState: true,
+      maintainAnimation: true,
+      maintainSize: true,
+      child: child,
     );
   }
 }
--- a/mobile/packages/divine_ui/test/src/bottom_sheet/vine_bottom_sheet_header_test.dart
+++ b/mobile/packages/divine_ui/test/src/bottom_sheet/vine_bottom_sheet_header_test.dart
@@ -34,7 +34,9 @@ void main() {
       );
 
       expect(find.text('Test Title'), findsOneWidget);
-      expect(find.byKey(const Key('trailing')), findsOneWidget);
+      // The real trailing widget, plus the hidden copy that reserves the same
+      // width on the empty leading side.
+      expect(find.byKey(const Key('trailing')), findsNWidgets(2));
     });
 
     testWidgets('uses default padding when none is provided', (tester) async {
@@ -105,10 +107,12 @@ void main() {
       );
 
       expect(find.text('Test Title'), findsOneWidget);
-      expect(find.byKey(const Key('leading')), findsOneWidget);
+      // The real leading widget, plus the hidden copy that reserves the same
+      // width on the empty trailing side.
+      expect(find.byKey(const Key('leading')), findsNWidgets(2));
     });
 
-    testWidgets('renders leading action with a trailing placeholder', (
+    testWidgets('centers the title when only a leading action is given', (
       tester,
     ) async {
       await tester.pumpWidget(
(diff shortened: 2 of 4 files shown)


mbradley (dismissed):
Verified against `622443454`: 830 divine_ui tests, 100.0000% line coverage, `flutter analyze` clean, and the four affected app-layer sheet suites green. The centering mechanism is correct — I measured it exact in LTR, RTL, with ellipsized titles, at either slot, with no height cost and no change to the no-title case. The `hasTitle` guards also quietly delete a placeholder the old code emitted when there was no title to center; that's a real cleanup the description doesn't claim.

One blocker, and it lands on the screen this PR is motivated by.

**The reservation is unbounded.** The row's fixed cost goes from `W + 52` to `2W + 24`, so a wide slot costs the title twice over. On the comments sheet, `_CommentsSortToggle` in shipped locale `ms` ("Teratas") takes the title from 178.5px to 93.0px at default text scale, and at text scale ≥1.6 the header overflows — `A RenderFlex overflowed` — where the 40px placeholder did not. I reproduced the mechanism directly with a 200px slot on a 400px sheet: overflow of exactly 56px, matching the arithmetic (the old budget was 104px, clean). Before it overflows, the `Flexible` collapses to 0 and the title wraps, taking the header from 40px to 120px. `accessibility.md` asks for a largest-font pass, and this is what it catches.

The old placeholder was wrong about centering but bounded; the mirror is exactly right until it yellow-stripes. Options in Q1 below.

I've pushed five commits for the objective findings — three doc corrections and two test strengthenings — each its own commit so you can take or drop them individually:

- `094ab8ca5` — the dartdoc says every `DivineIconButton` is 48px. I measured all 14 size/type combinations: 13 are **52**; only `small/secondary` is 48, because the outer padding is spent on a border only when the variant has one. The argument is unchanged, but the number would have sent the next reader to hardcode 48.
- `61ab9d680` — widens the slot warning past `GlobalKey`s. The mirrored slot is a separate element with its own `State`, so `initState` runs twice; analytics, fetches, subscriptions and timers would all fire for the hidden copy.
- `1bcf90877` — the description says both consequences are called out in the class doc, but the finder-count change lived only in inline test comments. It's the one consumers hit first, and the ordering flips depending on which slot is filled.
- `404f537e2` — the screen-reader test asserted only that the label is found once, which is equally true when the mirror is silenced and when there is no mirror at all. It passed with `_SlotMirror` deleted outright. It now asserts the slot is in the tree twice first, so it fails in both directions.
- `4840b4962` — the width-sweep test compared the 60px and 160px centers against each other, so a constant offset left it green (verified with a 30px skew that both absolute-center siblings caught and this one didn't). It's still the only test that catches a width-dependent clamp; anchoring each measurement to the header center makes it strictly stronger.

Both test commits were mutation-verified rather than just run green. Validation after the push: 830 tests, 100.0000% coverage (LH=LF=1976), `flutter analyze lib test` clean, `banner_color_swatches_test.dart` green.

Three questions below are yours to answer. Please review the pushed commits along with them.

---

### Q1 — how should the reserved width be bounded? (blocking)

Three ways out, with what each costs:

1. **Clamp the mirror** — `ConstrainedBox(maxWidth: …)`, or cap at a fraction of the row. Smallest diff, keeps exact centering for the common case, degrades to *partial* centering for wide slots instead of overflowing. Needs a number chosen and justified.
2. **Accept, and bound elsewhere** — constrain the sort chip itself (ellipsize the label) so the slot never gets wide enough to matter. Keeps the header simple; pushes the constraint onto every current and future slot.
3. **Layout-level fix** — a small `RenderBox` that measures both slots and centers, the shape `_RenderAdaptiveButtonPadding` (`divine_button.dart:466`) already established in this package. Fixes this, the both-slots case, and the double build at once; clearly its own PR.

I'd take (1) or (2) here and file (3), but the number in (1) and the ellipsis in (2) are product-visible, so they're yours. Which do you want?

### Q2 — should the double-build contract be enforced, or stay documentation?

A `GlobalKey` on a slot throws "Multiple widgets used the same GlobalKey", and I confirmed a key on a *descendant* of the slot is equally fatal — so a root-level assert can only catch the shallow case. It also only fires on the one-slot path, so a two-slot smoke test passes and the one-slot screen crashes.

- **Root assert** — cheap, points at the right widget, provably incomplete.
- **`WidgetBuilder` slots** — `Widget? leading` → `WidgetBuilder? leading`, so each build yields a distinct instance. Complete fix, kills the `GlobalKey` and double-`initState` hazards together, but it's a breaking API change across five call sites.
- **Test-only** — pin the current behavior so nobody simplifies the doc away.

Given `video_editor_captions_sheet.dart:115` already puts a destructive delete-track control in `headerLeadingAction` — safe today only because the trailing button happens to be set — I lean toward the builder API. Your call on whether that belongs here or in a follow-up.

### Q3 — the "Out of Scope" section rests on a premise that doesn't hold

It says no both-slot header exists today. Three do: `video_metadata_content_warning_selector.dart:113`, `video_metadata_tags_selector.dart:191`, and `user_picker_sheet.dart:321` when `maxCount != null`. Each pairs a `secondary` button (48px) with a `primary` one (52px) — I measured the content-warning header at **2px** off-center on shipped screens.

The conclusion is probably still right, but the argument has to change from "nothing hits this" to "three screens hit this and 2px isn't worth a render object." Alternatively, normalizing `DivineIconButton`'s outer footprint so all variants agree would fix it with no layout work at all. Which framing do you want, and is the padding normalization worth a follow-up?

---

Three factual points in the description that no commit can reach:

1. **The `Opacity(0)` contrast isn't right.** `Opacity` at alpha 0 also returns early from `paint` and sets `alwaysNeedsCompositing == false` — no layer, no paint. The real justification for `Visibility` is the rest of that sentence: it bundles pointer, semantics and focus exclusion, which `Opacity(0)` gives you none of. That argument stands on its own.
2. **"Single-action sheets" mislabels two of the four named.** The content-warning selector fills both slots unconditionally; the user picker fills both whenever `maxCount != null`. Only community-suggest and banner-colors are genuinely single-action — and "4px off" is right for them by coincidence of the `secondary` variant, not because all icon buttons are 48px.
3. **"Both called out in the class doc"** — only one was. `1bcf90877` adds the second.

For what it's worth, I also ran a trust & safety pass over the sheets this touches. It comes back clean, and for a structural reason worth recording: every consequential surface is out of the mirror's path — `report_content_dialog.dart` (the NIP-56 video/DM/user report path) passes no header title or slots; block/mute/delete/unfollow prompts go through `VineBottomSheetPrompt`/`ActionMenu`/`TileMenu`, which leave both slots empty; minor-safety and content-label screens use `contentTitle`, below the divider. Only four sites mirror at all, and the sole people-surface among them duplicates a sort chip. I verified the copy is genuinely inert against the pinned SDK — `ExcludeFocus` + `IgnorePointer`, dropped from the semantics tree, not activatable by tap, Enter, or a screen-reader activation.

Thanks, as always.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -88,12 +100,10 @@ class VineBottomSheetHeader extends StatelessWidget {
...
-                    if (leadingAction != null)
-                      leadingAction!
-                    else if (leading != null)
-                      leading!
-                    else if (hasTrailingSlot)
-                      placeholder,
+                    if (leadingSlot != null)
+                      leadingSlot
+                    else if (hasTitle && trailingSlot != null)
+                      _SlotMirror(trailingSlot),
```
realmeylisdev: **blocking** — the reservation is unbounded, and the only slot in the app whose width is text-driven is the one on the screen this PR is about.

The row's fixed cost goes from `W + 40 + 24` to `2W + 24`, so overflow starts at `W > (screenWidth - 32 - 24) / 2` instead of `W > screenWidth - 96`. The comments sheet reaches the header through `trailing:` with no leading slot, so `_SlotMirror(trailingSlot)` duplicates `_CommentsSortToggle`, whose width is `38 + labelWidth`.

Measured in a widget test with the app's real fonts, chip width by text scale for `ms` ("Teratas"):

| scale | 1.0 | 1.35 | 1.6 | 2.0 | 2.35 | 2.6 | 2.8 | 3.12 |
|---|---|---|---|---|---|---|---|---|
| chip | 84.9 | 100.1 | 110.9 | 128.3 | 143.5 | 154.3 | 163.0 | 176.9 |

Against `main`'s layout in the same harness (`Comments` title, `ms`):

| width | scale | main | this branch |
|---|---|---|---|
| 360 | 1.6 | header 86px, title 139.6px | header 122px, title 82.1px |
| 360 | 2.0 | header 142px, title 135.7px | header 286px, title 47.4px |
| 360 | 2.6 | header 232px, ok | **RenderFlex overflowed by 4.7px** |
| 320 | 2.35 | header 270px, ok | **overflowed by 23px** |
| 390 | 3.12 | header 271px, ok | **overflowed by 20px** |

`main` never overflows at any scale in the sweep (it would need `W > 224` on a 320dp screen; the chip maxes out at 176.9). 430dp is safe on both. English never overflows on either, but still loses title width and gains header height.

There's no global text-scale clamp, so the header takes whatever the OS gives it. iOS accessibility sizes reach ~3.12 and Android reaches 2.0, so the overflow band is iOS-accessibility-only, but the header-doubling band at 1.6-2.0 is reachable on both.

Correcting myself from the last round: I said "at text scale >= 1.6 the header overflows". That's wrong. At 1.6 the two slots take 245.8px of a 328px row, so it fits; what happens is the `Flexible` title collapses and wraps. The overflow threshold is 2.35 / 2.6 / 2.8 / 3.12 by screen width as above.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -134,3 +142,29 @@ class VineBottomSheetHeader extends StatelessWidget {
...
+/// Reserves the width of the opposite header slot so the title stays centered
+/// when only one side is filled.
+///
+/// A fixed-size box cannot do this: slot widgets size themselves, so a sort
+/// chip is as wide as its current label and a [DivineIconButton] lays out at
+/// 48-52px depending on its type, never the 40px a constant placeholder used
+/// to reserve. Mirroring the filled slot
+/// always reserves exactly the right width. The copy is laid out but never
+/// painted, focusable, tappable, or announced to screen readers.
+class _SlotMirror extends StatelessWidget {
```
realmeylisdev: **context for whichever fix you pick** — this is the third pass at the same header and the same chip, and the layout it used to have didn't have this failure mode.

`a72f896cc` (#1395, Feb) removed a hardcoded `SizedBox(width: 62, child: trailing)` from this header because the sort chip outgrew it and overflowed by 2.8px (#1394). Then `e3c2468dc` (#3516, May) replaced the header's `Stack` with `Row(mainAxisAlignment: .spaceBetween)` plus `const placeholder = SizedBox(width: 40, height: 40)`. The `Stack` it replaced centered the title against the full row width and overlaid the slots on top:

```dart
ConstrainedBox(
  constraints: const BoxConstraints(minHeight: 40),
  child: Stack(
    alignment: Alignment.center,
    children: [
      Center(child: DefaultTextStyle(style: VineTheme.titleMediumFont(), child: title!)),
      if (leading != null) Align(alignment: Alignment.centerLeft, child: leading),
      if (trailing != null) Align(alignment: Alignment.centerRight, child: trailing),
    ],
  ),
)
```

An overlaid child costs no main-axis width, so that version centered exactly, for free, with the slot built once. What it lacked was any protection against a long title running under the buttons, which is presumably why nobody noticed the `Row` rewrite changed the centering contract.

Worth knowing before you choose: `NavigationToolbar` in `package:flutter/widgets.dart` (what `AppBar` uses) is that same overlay-style centering with the missing guard already built in. `_ToolbarLayout` gives `middle` a `maxWidth` of `width - leadingWidth - trailingWidth - middleSpacing * 2`, centers it on the full width when it fits, and slides it against the nearer slot when it doesn't. That gets exact centering, no overflow, a single build of each slot, and the both-slots case, in framework code.

The catch is real though: it's a `CustomMultiChildLayout`, so it sizes to `constraints.biggest` and needs a bounded height, and `MultiChildLayoutDelegate` has no intrinsics so `IntrinsicHeight` won't rescue it. Adopting it means deciding whether this header has a fixed height or keeps growing with text scale, and today it grows. That's a call I don't want to make for you, which is why this is a comment rather than a commit.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -8,6 +8,18 @@ import 'package:flutter/material.dart';
 ///
 /// Combines drag handle and title section as per Figma design.
 /// Uses Bricolage Grotesque bold font at 24px for title.
+///
+/// When a title is shown and only one of the leading/trailing slots is filled,
+/// the empty side reserves the filled one's width so the title stays centered.
```
realmeylisdev: **should-fix (description, not code)** — the Out of Scope section says "No such header exists today (the only two-slot headers pair equal-size `DivineIconButton`s)". Two do exist and they aren't equal-size:

- `video_metadata_content_warning_selector.dart:113`
- `video_metadata_tags_selector.dart:191`

Both pair `size: .small, type: .secondary` on the leading side with `size: .small` (primary) on the trailing side. I measured all 14 size/type combinations: `small/secondary` lays out at 48px and the other 13 at 52px, because the outer padding only goes to a border on the variant that has one. So those two headers sit exactly 2.00px left of centre, verified with `getCenter(find.text(...)).dx - getCenter(find.byType(VineBottomSheetHeader)).dx`.

`user_picker_sheet.dart:321` is the third, but only when `maxCount != null`; with `maxCount` null and no initial selection its trailing is null and it mirrors instead.

The conclusion is still right, 2px isn't worth a render object. The argument just needs to be "three screens hit this and 2px is fine" rather than "nothing hits this". Alternatively, giving every `DivineIconButton` variant the same outer footprint fixes it with no layout work at all, which might be worth its own issue.

mobile/packages/divine_ui/test/src/bottom_sheet/vine_bottom_sheet_header_test.dart
```diff
@@ -126,66 +130,124 @@ void main() {
...
+        tester.getCenter(find.text('Test Title')).dx,
+        moreOrLessEquals(
+          tester.getCenter(find.byType(VineBottomSheetHeader)).dx,
         ),
-        findsNothing,
       );
     });
 
-    testWidgets('renders trailing action with a leading placeholder', (
+    testWidgets('keeps the title centered as the trailing width changes', (
```
realmeylisdev: **should-fix** — every test here pumps at text scale 1.0, and there is no `textScaler` anywhere in `packages/divine_ui/test/` or `test/screens/comments/`. That's the reason the regression above is invisible to CI even though all 830 tests are green.

The width sweep is the natural home for it: same shape, one extra axis. Something like pumping the widest shipped sort label at scale 2.6 on a 360px-wide surface and asserting `tester.takeException()` is null would fail today and pass under whichever bound you land on. `accessibility.md` already asks for a largest-font pass, this is what it would have caught.

mobile/test/screens/profile_setup/banner_color_swatches_test.dart
```diff
@@ -182,7 +182,9 @@ void main() {
       final result = await openSheet(tester);
       expect(result, isEmpty, reason: 'sheet is still open');
 
-      await tester.tap(find.byTooltip(l10n.commonBack));
+      // The header reserves the back button's width on the empty trailing
+      // side by re-rendering it hidden, so the real button is the first match.
+      await tester.tap(find.byTooltip(l10n.commonBack).first);
```
realmeylisdev: **nit** — `.first` works, but it's load-bearing on which slot happens to be filled. The class doc you added says the hidden copy comes first when `trailing` is the filled slot and second when `leading` is, so this line quietly depends on the banner sheet keeping a leading-only header. Move the action to trailing some day and this taps the inert copy, and the failure is a silent no-op rather than a useful message.

`find.descendant(of: find.byType(DivineIconButton), ...)` won't help since both copies are real buttons. Two things do, both checked against this header:

```
find.byTooltip(l10n.commonBack)               -> 2
find.byTooltip(l10n.commonBack).hitTestable() -> 1
find.bySemanticsLabel(l10n.commonBack)        -> 1
```

`hitTestable()` is the closest to what the test means, since the copy sits under `IgnorePointer`, and it needs no semantics handle.

realmeylisdev (dismissed):
still blocked on the reservation being unbounded, but with better numbers this round and a correction to my own.

measured with the real bundled fonts (`mobile/pubspec.yaml` declares the `Inter` / `BricolageGrotesque` families that `GoogleFonts.inter` and `.bricolageGrotesque` resolve to, so `flutter test` lays these out exactly like the device does): `main` never overflows, this branch does from text scale 2.35 on a 320dp screen, 2.6 on 360dp, 2.8 on 375dp, 3.12 on 390dp, in `ms` and `vi`. my earlier "at text scale >= 1.6 it overflows" was wrong. at 1.6 it doesn't overflow, it grows the header from 86px to 122px and takes ~40% off the title. rest is inline.

the good news is the blast radius is one screen: comments is the only sheet whose slot width comes from text. the other four mirror sites all duplicate a fixed 48px `DivineIconButton`, which costs 120px total and can't overflow anything.

everything else holds up. 830 tests, 100.0000% coverage, `Visibility` really is inert on 3.44 (`_RenderVisibility` returns early from `paint` and drops the subtree in `visitChildrenForSemantics`, plus `IgnorePointer` and `ExcludeFocus`), and deleting `_SlotMirror` turns 7 tests red, so the centering guard is a real one.

hm21:
@mbradley @realmeylisdev — you were both right that the reservation was unbounded, and rather than bound it I dropped the mechanism. `4caaa1456` replaces `_SlotMirror` with a small render object: slots at the edges, title centered on the full width, both sides kept clear by capping the title at `width - 2 * widestSlot - 2 * gap`. The title is the only element that gives, so the row cannot overflow at any text scale — Q1 is answered by removing the reservation rather than picking a number for it.

That takes the rest with it. Q2 is moot: each slot is built once again, so no `GlobalKey` restriction, no second `initState`, and no builder-API question. Q3 is fixed rather than re-argued: centering no longer depends on the slots being equally wide, so the content-warning and tags headers are on center instead of 2px off. Slot finders match one widget again, which puts `vine_bottom_sheet_test.dart` and `banner_color_swatches_test.dart` back to their state on `main` — the `.first` and the `findsNWidgets(2)` edits are gone, so the `hitTestable()` suggestion no longer has anything to attach to.

Your two test commits are gone with the code they pinned, and both doc commits (`094ab8ca5`, `1bcf90877`) described `_SlotMirror`, so they went too. The measurements behind them stand: I confirmed the 48-vs-52 split in `divine_icon_button.dart` (the outer `Padding(all: _hasBorder ? 0 : 2)`) and it is what makes the two-slot fix worth having. The description is rewritten — the `Opacity(0)` contrast, the "single-action sheets" list and the "no such header exists" premise were all wrong and are out.

One judgement call is mine and worth disagreeing with if you see it differently. Where one slot is wide enough that keeping the same width clear on the empty side leaves the title nothing, centering is geometrically impossible — two copies plus the gaps already exceed the row. The title then falls back to the space between the slots, i.e. what `main` does today, rather than collapsing to zero width. Reachable only with a text-driven slot at high accessibility scales on 320-390dp.

Verification: 840 `divine_ui` tests at 100.0000% line coverage, 4639 in `test/widgets` + `test/screens`, `flutter analyze lib test packages/divine_ui` clean. The new guards were mutation-checked, not just run green — centering between the slots instead of on the full width turns 7 red, dropping RTL turns 1 red, dropping the fallback turns 1 red. Tested on a real Samsung Galaxy and on macOS across the affected sheets.

Thanks — the overflow numbers are what killed the mirror, and neither of the two shipped alternatives would have kept the title exactly centered.

realmeylisdev (commented):
centering is exact now, 0.00px on every icon-button header where main was 2-4px off, and the row itself never overflows. dry layout, RTL, hit testing and semantics order all check out.

the comments header still bites though: the title box's left edge moves right by `chipWidth - 40`, and `_CommentsTitle` is a `Row(mainAxisSize: .min)` so its Text can neither wrap nor clip. it slides under the sort chip in English at ordinary text sizes, where main was clear. numbers inline.

the `canCenter == false` fallback is already 52px better than main there, so this looks like entering it sooner rather than bounding anything.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -134,3 +130,253 @@ class VineBottomSheetHeader extends StatelessWidget {
...
+    // Unbounded width only happens through misuse (a header in an
+    // unconstrained Row); there is always room to center in that case.
+    final centeredBudget = constraints.hasBoundedWidth
+        ? constraints.maxWidth - inset * 2
+        : double.infinity;
+    // One slot can be wide enough that keeping the same width clear on the
+    // empty side leaves the title nothing. Centering is then impossible
+    // rather than merely tight, so the title takes the space between the
+    // slots instead of collapsing to zero.
+    final canCenter = centeredBudget > 0;
```
realmeylisdev, line 238: **blocking** — the comments title now runs under the sort chip in English at ordinary text sizes, and it did not on `main`.

Not a budget problem, a *position* problem. When `canCenter` is true the title box is laid out at `centeredBudget` and centred on the full row, so its left edge lands at `16 + inset` = `16 + chipWidth + 12`. On `main` the box was the `Flexible` between the 40px placeholder and the chip, so its left edge was a fixed `16 + 40 + 12 = 68`. The box therefore translates right by exactly `chipWidth - 40` px, always, for every text scale and locale.

Normally a narrower box would just wrap. It can't here: `_CommentsTitle` (`comments_screen.dart:117`) is `Row(mainAxisSize: MainAxisSize.min, children: [Text(...), pill?])`, and `RenderFlex` hands non-flex children an unbounded main axis (`flex.dart:_constraintsForNonFlexChild`), so that `Text` lays out at its full intrinsic width on one line no matter what `maxWidth` the header gives the Row. It can't clip either: `Flex.clipBehavior` defaults to `Clip.none` and `PaintingContext.pushClipRect` early-returns on `Clip.none` (`object.dart:579`), so `RenderFlex.paint` paints the overflow unclipped. The yellow stripes and the console error are inside the `assert` at `flex.dart:1420`, so on a release build there is no signal at all.

Measured with the shipped TTFs, title modelled as `12 Comments` (a 3-digit count is worse), *title ink right edge vs sort-chip left edge, negative = clear*:

| config | this branch | `main` |
|---|---|---|
| en, 320dp, scale 1.5 | **+28.5** | −3.7 |
| en, 360dp, scale 1.77 | **+28.7** | −9.4 |
| en, 390dp, scale 2.0 | **+33.1** | −10.0 |
| ms, 360dp, scale 1.5 | **+10.7** | −34.0 |
| ms, 430dp, scale 2.0 | **+21.5** | −39.4 |
| ms, 430dp, scale 3.01 | **+184.7** | +52.6 |

The chip is opaque (`containerLow`, alpha `FF`) and paints after the title, since `children` yields leading → title → trailing. So the visible artifact is the title disappearing under the chip and reappearing in the sheet's 16px right padding, not glyphs on top of it. Android's font-size slider reaches 2.0 and its ordinary (non-accessibility) range already reaches ~1.3, so this is not an accessibility-extreme-only case, and it is the one screen the PR was written for.

The interesting part: your own `canCenter == false` branch does not have this problem. It anchors the title at `x = 16`, which is 52px *better* than `main` (ms/320/2.5: branch +50.3, main +102.3). So the fix is to enter the fallback earlier rather than to add a bound. `centeredBudget > 0` asks "is centring geometrically possible", when the question that matters is "does the title still fit if I centre it". `NavigationToolbar`'s `_ToolbarLayout` asks the second one: it caps `middle` at the *between-slots* width and slides it against the nearer slot only when the centred rect would collide.

Which knee is right is your call and it does have a cost — a wrappable title (tags picker, content-warning) would stop wrapping-and-centred and start one-line-and-slid, which may be worse for those. `title.getMinIntrinsicWidth(double.infinity)` (widest unbreakable line) is the middle option. What I don't think survives is the current threshold, because between "centring is comfortable" and "centring is impossible" there is a wide reachable band where the title is centred and unreadable.

Worth saying: this is also arguable as a `_CommentsTitle` bug — a header title that refuses to shrink will defeat any budget policy, and a `Flexible` + `TextOverflow.ellipsis` there would make the header's choice non-fatal. That's outside this diff, but it might be the cheaper half of the fix.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -134,3 +130,253 @@ class VineBottomSheetHeader extends StatelessWidget {
...
+    ?_leading,
+    ?_title,
+    ?_trailing,
+  ];
+
+  /// Width each side keeps clear of the title, so the centered title clears
+  /// the wider slot on both sides.
+  double _sideInset(double leadingWidth, double trailingWidth) {
+    final widest = math.max(leadingWidth, trailingWidth);
+    return widest == 0 ? 0 : widest + _gap;
```
realmeylisdev, line 206: **should-fix** — the cost of exact centring lands at default text scale in the non-English locales, on the other four headers.

`_sideInset` charges `max(L, T) + 12` to *both* sides, so the title budget is `W - 2*(widest + 12)` where `main` charged `widest + 40 + 24`. Deltas: two-slot 48/52 headers lose 4px, leading-only headers lose 8px (main reserved a 40px placeholder where the real button is 48px).

Swept all four non-comments headers × 4 widths × 22 shipped locales at **textScaler 1.0**: 17 of 616 (header, locale, width) configs gain a wrapped title line, 16 of those grow the header by +16 to +34px, and none go the other way. English is clean everywhere. The 13 locales that pay are am, bg, de, fil, fr, id, it, ms, nl, pt, sv, tr, ur. Worst cell is `ms` content-warning at 320dp: 98px → 132px, because both lines lose their margin by 0.21px and 2.64px.

Small, and plausibly worth paying for exact centring. Flagging it because it is the price tag and it is invisible from an English simulator.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -84,39 +101,18 @@ class VineBottomSheetHeader extends StatelessWidget {
...
-                    if (hasTitle)
-                      Flexible(
-                        child: Center(
-                          child: DefaultTextStyle(
-                            style: VineTheme.titleMediumFont(
-                              color: context.vineColors.onSurface,
-                            ),
-                            textAlign: .center,
-                            child: title!,
+                child: _HeaderRow(
```
realmeylisdev, line 104: **should-fix (description)** — the rewritten affected-header list is closer than last round but still misses two.

`git grep -n "headerLeadingAction:\|headerTrailingAction:" -- mobile/lib` turns up:

- `video_metadata_expiration_selector.dart:52` — leading-only (`.x`, secondary, small) through `VineBottomSheetSelectionMenu.show`. A third single-action header, 4px off on `main`, not in the "(community suggest, banner colors)" list.
- `video_editor_captions_sheet.dart:115` / `:131` — leading (small/error 52px when `canDeleteTrack`, else small/secondary 48px) paired with trailing (small/primary 52px). A fourth two-slot header, 2px off on `main` in the non-deletable branch, not in the "(content-warning selector, tags picker, user picker when maxCount != null)" list.

Also `user_picker_sheet.dart:335-346` renders a trailing action when `maxCount == null && initialSelectedProfiles.isNotEmpty`, so the `maxCount != null` condition in the body is narrower than the code. Eight headers, not six.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -1,13 +1,30 @@
...
 /// Combines drag handle and title section as per Figma design.
 /// Uses Bricolage Grotesque bold font at 24px for title.
+///
+/// The title is centered on the full header width, never between the
+/// leading/trailing slots, so it does not move when a slot changes width — a
+/// sort chip relabelled New → Top, or two icon buttons whose variants happen
+/// to differ by a couple of pixels. Both sides are kept clear by capping the
+/// title at `width - 2 * widestSlot - 2 * gap` instead of reserving layout
+/// space opposite the filled slot, so a wide slot costs the title its own
+/// width once rather than twice and the row cannot overflow.
```
realmeylisdev, line 21: **nit (doc)** — "a wide slot costs the title its own width once rather than twice" is the one sentence in the block that isn't right. The budget is `width - 2*widestSlot - 2*gap`, which is twice, and it is bit-for-bit the budget `_SlotMirror` gave the title. What changed versus the mirror is that the *row* no longer over-demands and overflows; the title's share is unchanged. Since the budget is the crux of this review, worth stating precisely.

Two stale lines carried in from before, while you're in here: line 13 says the title is "Bricolage Grotesque bold font at 24px" (`VineTheme.titleMediumFont` is 16/24), and the `padding` doc says the default is `start: 24, end: 24` (the code defaults to 16/16; 24 is what `vine_bottom_sheet.dart:632` passes on the fixed-content path).

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -134,3 +130,253 @@ class VineBottomSheetHeader extends StatelessWidget {
...
+
+  Size _computeLayout(
+    BoxConstraints constraints,
+    ChildLayouter layoutChild, {
+    bool positionChildren = false,
+  }) {
+    final leading = _leading;
+    final title = _title;
+    final trailing = _trailing;
+    final slotConstraints = BoxConstraints.loose(
```
realmeylisdev, line 217: **nit** — the render object has no overflow reporting in either axis, so two inputs that used to shout now fail silently.

Slots are laid out loose at the full row width and pinned to opposite edges with no cross-check, so `leadingWidth + trailingWidth > size.width` just overlaps. Measured on a 300px row with two 200px slots: branch gives 100px of overlap and `takeException()` null; the same input on `main` gives `A RenderFlex overflowed by 124 pixels on the right`. Same story vertically — `constraints.constrain` clamps `size.height` below the tallest child and `_positionChild` then passes a negative `freeY`, so the child paints above the box unclipped.

Neither is reachable today (the fixed pair is 48+52 against a ≥288px row, and every call site sits in a `Column(mainAxisSize: min)` so `maxHeight` is infinite), so an `assert` in the positioning block is enough to get the diagnostic back at zero runtime cost.

While here: `computeDryBaseline` isn't overridden, so `getDryBaseline` hits `debugCannotComputeDryLayout` and throws in debug under a baseline-aligning parent, which `main`'s `RenderFlex` handled. Also unreachable (the parent `Column` is centre-aligned), and the SDK's own `RenderCustomMultiChildLayoutBox` has the same gap, so this is a "if you're touching it anyway".

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -134,3 +130,253 @@ class VineBottomSheetHeader extends StatelessWidget {
...
+                    ),
+            ),
+          );
+
+    final size = constraints.constrain(
+      Size(
+        constraints.hasBoundedWidth
+            ? constraints.maxWidth
+            : titleSize.width + inset * 2,
+        math.max(
```
realmeylisdev, line 262: **nit** — dropping the placeholder also drops a height floor, and the comments header gets 16px shorter at every width.

`main`'s `SizedBox(width: 40, height: 40)` floored any one-slot header's row at 40px. Here the row is `max(titleH, leadingH, trailingH)`. For the comments sheet that is `max(24, 24)`: the chip is `4 + 4 + max(icon 14, labelMedium line box 16) = 24` and the title line box is `titleMediumFont` 16/24 = 24. So the row goes 40 → 24 and the whole sheet body moves up 16px (measured: header 88px → 72px at scale 1.0, every width).

The other four are unaffected — `DivineIconButton.small` is 48 or 52, both above 40. Might well be an improvement, but it isn't in the description and nothing would catch it: there is no `matchesGoldenFile` or alchemist golden anywhere in `mobile/`, despite `mobile/scripts/golden.sh`.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -134,3 +130,253 @@ class VineBottomSheetHeader extends StatelessWidget {
...
+      }
+      if (trailing != null) {
+        _positionChild(
+          trailing,
+          isRtl ? 0 : size.width - trailingSize.width,
+          size.height - trailingSize.height,
+        );
+      }
+      if (title != null) {
+        final titleX = canCenter
```
realmeylisdev, line 286: **nit (description)** — "the title falls back to the space between the slots, i.e. what `main` does today" isn't quite what the code does. The fallback pins the title at `leadStart`, which is 0 for a trailing-only header, and the test at line 340 asserts exactly that (`title.left == row.left + 16`). On `main` that box started at `row.left + 68` and the title was `Center`ed inside it.

I went looking for a regression here and there isn't one: I measured `main` side by side and its `Center` is a no-op at every reachable fallback scale, because the clamped title fills its box on both sides. The single cell where centring inside the between-slots box would move anything is ms / count=1 / 320dp / ~2.10, and it is 2.9px. And the branch's box is 52px wider than main's there, so the fallback is the better of the two. Only the sentence needs fixing, not the code.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
```diff
@@ -134,3 +130,253 @@ class VineBottomSheetHeader extends StatelessWidget {
...
+
+  @override
+  double computeMaxIntrinsicWidth(double height) =>
+      _intrinsicInset(height) * 2 + (_title?.getMaxIntrinsicWidth(height) ?? 0);
+
+  double _intrinsicHeight(
+    double width,
+    double Function(RenderBox child, double width) intrinsic,
+  ) {
+    final titleWidth = width.isFinite
```
realmeylisdev, line 331: **nit** — `_intrinsicHeight` measures the title at the centred budget with no `canCenter` branch, so it disagrees with `_computeLayout` in the fallback regime. `computeDryLayout` delegates to `_computeLayout` and is fine; this is the only path that diverges.

Measured, row width 300, leading `SizedBox(200x20)`, 44-char title: real row height 216.0, `getMinIntrinsicHeight(300)` = `getMaxIntrinsicHeight(300)` = 888.0. The intrinsic path computes `titleWidth = max(0, 300 - 424) = 0` and the paragraph breaks per character (37 lines), where `performLayout` hands the title `300 - 212 = 88`. At tree level: bare header 230px, wrapped in `IntrinsicHeight` 902px. `main` was correct here.

Not reachable in the app (nothing above a `VineBottomSheetHeader` uses intrinsics — no `IntrinsicHeight`/`IntrinsicWidth`/`Table`/baseline `crossAxisAlignment`), so this is latent rather than live.

Separately the `width.isFinite` guard on that line is a no-op: `infinity - 224 == infinity` and `math.max(0, infinity) == infinity`, so both arms return the same value for every input.

mobile/packages/divine_ui/test/src/bottom_sheet/vine_bottom_sheet_header_test.dart
```diff
@@ -108,84 +140,415 @@ void main() {
...
+        );
+        expect(
+          tester.getCenter(find.text('Test Title')).dx,
+          moreOrLessEquals(
+            tester.getCenter(find.byType(VineBottomSheetHeader)).dx,
+          ),
+        );
+      });
+
+      testWidgets('caps the title instead of overflowing on a wide slot', (
```
realmeylisdev, line 290: **should-fix** — this is the third round asking, so let me be concrete about why the `SizedBox` stand-in doesn't close it.

There is still no `textScaler` anywhere in `packages/divine_ui/test/` or `test/screens/comments/`, and only two files in the repo mention `VineBottomSheetHeader` (both here). A fixed-width box holds the title font constant while the slot grows, which drops the one variable that produces the symptom, and it encodes a growth curve the real chip doesn't follow — `_CommentsSortToggle` is `20 + 14 + 4 + label`, and the 14px `Icons.schedule` does not text-scale, so its width is `≈38 + scale·labelWidth`, not `scale·178`.

The two fixtures also bracket the dangerous regime without landing in it: 100px gives `centeredBudget` 104, 178px gives −52. The band that matters is `centeredBudget` in `(0, ~80]` — `canCenter` true, title unusable — which on a 360dp screen is a slot of 124-164px, exactly where the real chip sits at scale 1.7-2.4.

A test that pumps a `Row(mainAxisSize: .min)` title plus a text-scaled chip replica at 1.5 / 2.0 on a 320 and a 390 surface, asserting the title's painted right edge stays left of the slot's left edge, would be red on this branch and green on `main`. That is the assertion the blocking comment above is about.

Small thing in the comment on line 293: at 100px the mirror wouldn't have overflowed either (`100*2 + 24 = 224 < 328`), and it would have produced the same 104px budget at the same centre. The fixture that actually distinguishes the two approaches is the 178px one.

mobile/packages/divine_ui/test/src/bottom_sheet/vine_bottom_sheet_header_test.dart
```diff
@@ -108,84 +140,415 @@ void main() {
...
+        );
+
+        expect(tester.takeException(), isNull);
+        expect(
+          tester.getCenter(find.text(_longTitle)).dx,
+          moreOrLessEquals(
+            tester.getCenter(find.byType(VineBottomSheetHeader)).dx,
+          ),
+        );
+        expect(
```
realmeylisdev, line 312: **should-fix** — this assertion can't fail for the reason it states.

`getRect(find.text(...)).right` is the `RenderParagraph`'s own `size.right`, and `RenderParagraph` sets `size = constraints.constrain(textPainter.size)`. So it reports the clamp you just applied, whatever the glyphs do. Combined with `takeException(), isNull` on line 305 the test reads as "proves the row cannot overflow" but proves "the render object reports a constrained width" — which is true by construction of `_computeLayout`.

Something that can fail: compare the laid-out width against what the title needed, e.g. `tester.renderObject<RenderBox>(find.text(_longTitle)).getMaxIntrinsicWidth(double.infinity)` against the rect width, or scan for painted ink past the slot's edge. The laid-out-vs-needed shape is already used at `test/screens/inbox/inbox_view_test.dart:1476-1480`.

mobile/packages/divine_ui/test/src/bottom_sheet/vine_bottom_sheet_header_test.dart
```diff
@@ -108,84 +140,415 @@ void main() {
...
+        expect(trailingTaps, isZero);
+
+        await tester.tap(find.byKey(const Key('trailing')));
+        expect(trailingTaps, equals(1));
+      });
+
+      testWidgets('announces the slot to screen readers exactly once', (
+        tester,
+      ) async {
+        final semantics = tester.ensureSemantics();
```
realmeylisdev, line 410: **nit** — two small ones in this file.

`ensure_semantics`: the handle is released by a straight-line `semantics.dispose()` at the end, so if either `expect` above throws, `flutter_test` also reports "A SemanticsHandle was active at the end of the test" and buries the real message. `addTearDown(semantics.dispose)` is the usual fix.

The intrinsics test at line 456 uses fixed-size boxes for all three children, so `getMinIntrinsicWidth` and `getMaxIntrinsicWidth` return the same value and the four assertions at 479-494 distinguish two behaviours: swapping `getMinIntrinsicWidth` for `getMaxIntrinsicWidth` inside `computeMinIntrinsicWidth` turns nothing red. A wrapping `Text` as the title would make the pair mean something.

And `_headerRow`'s `firstWhere` has no `orElse`, so a rename throws `Bad state: No element` — loud but anonymous, and it doesn't assert uniqueness. `find.byWidgetPredicate((w) => w.runtimeType.toString() == '_HeaderRow')` with `tester.renderObject` gets both properties, which is what `divine_button_test.dart:696` already does for `_AdaptiveButtonPadding`.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
realmeylisdev: Overflow is gone, confirmed across every width/scale/locale cell I swept, and `_SlotMirror` is fully out. The other half of this comment did not go with it: the title's budget is still `W - 2*(slot+12)`, bit-for-bit what the mirror gave it, so the comments title now translates right by `chipWidth - 40` and lands under the sort chip. Carried forward to the new comment on `canCenter`; closing this one.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
realmeylisdev: You took the `NavigationToolbar` shape and solved the bounded-height objection I raised, which is the better answer than the one I had. The remaining question is only where the knee goes, which moved to the new comment on `canCenter`. Closing.

mobile/packages/divine_ui/lib/src/bottom_sheet/vine_bottom_sheet_header.dart
realmeylisdev: Corrected, and the 48-vs-52 arithmetic in the body checks out in all 14 combinations. The enumeration is still two headers short (expiration selector, captions sheet) so I moved that to a fresh comment. Closing this one.

mobile/packages/divine_ui/test/src/bottom_sheet/vine_bottom_sheet_header_test.dart
realmeylisdev: Still open in substance, but the new fixed-width fixtures changed the shape of the ask enough that I rewrote it against the current code. Continued at the new comment on the widest-slot test; closing this one so there is only one live thread.

mobile/test/screens/profile_setup/banner_color_swatches_test.dart
realmeylisdev: Moot now that each slot is built once. `.first` and the `findsNWidgets(2)` edit are both reverted and the file is byte-identical to main. Closing.

rabble (approved):

The diagnosis is exactly right and it's the part most people would have gotten wrong. A `Row` with `spaceBetween` and a placeholder can only center the title when the two slots happen to be the same width, so the hardcoded `SizedBox(width: 40)` was encoding an assumption that a `small/secondary` `DivineIconButton` (48px) and every other variant (52px) both quietly violate. Chasing that with a corrected placeholder width would have fixed three sheets and left the comments sort chip — the one slot whose width comes from text — still drifting on every sort change. Centering the title independently of the slots is the fix that generalises.

Capping at `width - 2 * widestSlot - 2 * gap` rather than mirroring the filled slot is the right call and the reasoning in the description is sound: mirroring charges the row `2W + 24` instead of `W + 64`, which is worse precisely for the text-driven slot at accessibility scales, where the old placeholder didn't overflow. Making the title the only element that yields means the row can't overflow at any scale. Good tradeoff, and documenting the rejected revision is more useful than just presenting the winner.

I read the render object closely. The layout protocol is implemented properly, which is the thing that usually goes wrong in hand-rolled `RenderBox` code:

- `_computeLayout` is shared between `performLayout` and `computeDryLayout` via a `ChildLayouter`, with positioning gated behind a flag — so dry layout can't silently diverge from real layout.
- RTL is handled at both positions *and* in the non-centered fallback. The fallback's `size.width - leadStart - titleSize.width` puts the title's trailing edge where the LTR branch puts its leading edge, which is correct rather than merely mirrored-looking.
- `children` yields leading → title → trailing, and that order drives paint *and* semantics traversal. In RTL the leading slot paints on the right, which is also where RTL reading starts, so the traversal order stays correct without a `sortKey`.
- `hitTestChildren` walks front-to-back through `addWithPaintOffset`.
- Intrinsics use the slots' max-intrinsic width for the inset on both sides, consistent with what layout does.

Mutation-checking the guards (7 tests red if you center between the slots, 1 if you drop RTL, 1 if you drop the fallback) is the evidence I'd actually want, and it's the reason I'm not asking for more tests. `divine_ui` is a strict-coverage package and the PR reports 100.0000% held across 840 tests, so the gate isn't being papered over either.

Two things, neither blocking:

**The unbounded-width branch still charges twice.** `constraints.hasBoundedWidth ? maxWidth : titleSize.width + inset * 2` reserves `inset` on *both* sides even when only one slot exists — a leading-only header with a 100px title sizes to 204 and leaves ~52px of dead space past the title's right edge. That's the exact double-charge the bounded path was designed to avoid, surfacing where it wasn't considered. Harmless in practice: these headers live in bottom sheets, which are width-bounded, so the branch is effectively unreachable today. Worth a comment so nobody later reads that expression as intentional symmetry.

**710 lines of render-object machinery for a centering fix is a real maintenance cost**, and it's now the only hand-rolled `RenderBox` in `divine_ui`. I went looking for the simpler version and couldn't construct one — a `Stack` with an `Align`ed pair of slots and a padded `Center` needs the widest slot's measured width to compute its padding, which means `LayoutBuilder` or an intrinsics dance that ends up more fragile than this. So I think the choice is justified, not over-engineered. Just noting that the next person to touch this header needs to understand the layout protocol, where before they needed to understand a `Row`.

Real-device verification across all three sort modes is what I'd have asked for. Ship it.

PR #6738, opened 2026-08-05
fix(comments): vote on addressable comment targets with their own coordinate

### LINKED ISSUE #6124: fix(likes): downvote path has no addressableId plumbing (latent, comment-only today)
## Summary

`LikesRepository`'s downvote family (`downvoteEvent`, `removeDownvote`, `toggleDownvote`, `_downvoteRecords`) has no `addressableId` parameter anywhere, unlike the like family which does. Found during investigation for #6020 (own-like heart un-fills after an edit).

## Evidence

- `_downvoteRecords` (`mobile/packages/likes_repository/lib/src/likes_repository.dart`) is keyed only by target event id, with no addressable-id companion — same structural shape the like path had before #6020's fix.
- `downvoteEvent`/`removeDownvote`/`toggleDownvote` accept no `addressableId` parameter at all (not even optionally, unlike `likeEvent`/`unlikeEvent`).
- `downvoteEvent`'s internal `_nostrClient.sendLike(..., content: '-', ...)` call omits `addressableId:` entirely, so a downvote reaction would never get an `a` tag even if the caller had a coordinate available.
- The only callers of the downvote family are in `mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart`, and they operate exclusively on comment ids (Kind 1111 comments — not addressable events, no `d` tag/coordinate). The sibling `likeEvent` call in the same bloc is also invoked without `addressableId`, for the same structural reason.
- Git history confirms downvotes were purpose-built for comment voting (#3433) and have never been extended to videos or any other addressable target.

## Impact

None today — zero current blast radius, since downvotes are never applied to addressable/video events. This is a latent gap, not a live bug.

## When this would become live

If downvotes are ever extended to addressable events (e.g. videos), this issue would need the same fix #6020 shipped for likes: an addressable-id companion index (in-memory + persisted), and threading `addressableId` through `downvoteEvent`/`removeDownvote`/`toggleDownvote`/the internal `sendLike` call.

## Suggested fix (when in scope)

Mirror the `#6020` pattern (PR #6123) for the downvote path: add `addressableId` parameters to `downvoteEvent`/`removeDownvote`/`toggleDownvote`, an addressable-id companion map/column for `_downvoteRecords`, and pass `addressableId` through the internal `sendLike` call.

realmeylisdev:
## Description

Comment votes are published against whatever event id a `Comment` wraps. When `FeatureFlag.videoReplies` is on that can be an **addressable Kind 34236 video reply** — `_threadKinds` requests `[1111, 34236]` and `_eventToComment` has no kind guard — so the emitted kind-7 carried `['e', <34236 id>]`, `['k','1111']` (wrong; the target kind is 34236) and **no** `['a', coordinate]`.

`VideoMetadataUpdateService` is reply-aware and republishes under the same `d` tag with a new event id, and both vote read paths are `#e`-only. So after an edit:

1. The voter's arrow silently un-fills; the a11y label reverts to "Upvote comment".
2. The score **disappears** rather than showing `0` (`comment_item.dart` guards the widget on `netScore != 0`).
3. **Every viewer** loses the count — `getVoteCounts` has no `authors:` constraint, so it is the public aggregate.
4. Re-tapping mints a **duplicate** kind-7; #6123's `alreadyLikedByCoordinate` guard cannot fire with a null coordinate.
5. The original reaction becomes **undeletable through the UI**.

This affects the **upvote path identically** — #6123 shipped `addressableId` on `likeEvent`, but this call site never adopted it.

Closes #6124, which was filed as downvote-only with "zero blast radius". The evidence for that premise being false is in [this comment](https://github.com/divinevideo/divine-mobile/issues/6124#issuecomment-5191603296).

## Commits

1. **`stop losing downvote deletions after a cold start`** — reachable with the flag *off*, so it affects the most users. `_downvoteRecords` is in-memory only, so after a cold start `removeDownvote` threw `NotDownvotedException`; the bloc swallowed it as a silent sentinel, cleared the UI, and never published the kind-5, leaving the downvote live. The same throw also aborted the handler, so a downvote→upvote switch skipped `likeEvent` entirely — the UI showed an upvote that was never sent. The repository now falls back to the relay copy, and the bloc scopes teardown failures so they cannot abort vote placement.
2. **`thread addressableId through the downvote family`** — the parameter on all three mutators, passed to the internal `sendLike`, plus a coordinate companion index so guards and teardown resolve by coordinate. Sync and realtime ingest now extract the `a` tag for downvote records.
3. **`give Comment its own coordinate and event kind`** — two nullable fields parsed from the event.
4. **`vote with the target's own coordinate and real kind`** — threads both through `CommentVoteToggled` for *both* vote directions.

## Two things worth a reviewer's attention

**The coordinate is deliberately the comment's own, never `rootAddressableId`.** funnelcake's `engagement_counts_by_coordinate_a_mv` counts `kind = 7 AND tag[1] IN ('a','A')` with **no content filter**, keyed on the coordinate parsed from the `a` tag. Using the parent video's coordinate would make every comment vote inflate that video's public `reaction_count`, while `/api/videos/{id}/likers` keeps filtering `content != '-'`. Using the target's own coordinate is count-neutral — `uniqStateIf(id, …)` dedupes across the e-path and a-path.

**A missing `d` tag yields a null coordinate**, deliberately *not* falling back to the event id the way `VideoEvent` does — that fallback produces a well-formed coordinate addressing nothing.

## No persistence

`personal_reactions` has no reaction-content discriminator and is keyed `{targetEventId, userPubkey}`, so it cannot represent a like and a downvote on the same target. Downvotes stay in-memory only, as in v1. No schema change, no migration, no codegen, no backend change.

## Testing

The existing vote tests used `any(named: 'targetKind')` matchers that passed no matter what was sent, and mocktail matches a null `addressableId` — so this fix would otherwise have shipped green and unguarded. Those are replaced with value assertions.

Every new test was run against the previous implementation to confirm it fails there. The bloc coordinate test reports the defect directly:

```
No matching calls. All calls: downvoteEvent({..., targetKind: 1111, addressableId: null})
```

- `likes_repository` 193 tests, coverage 90.87% (gate 62%)
- `comments_repository` 144 tests
- `test/screens/comments` + `test/blocs/comments` — 215 tests

## Out of scope

- **Already-cast votes are not healed.** Reactions already published without a coordinate stay stranded; resolving them needs revision history, which is what #6711 is adding (`VideoRevisionsResolver`). That is a follow-on once #6711 lands — its port is optional, so this PR does not depend on it.
- Downvotes already inflate a reply's `reaction_count` via the e-tag path today. Pre-existing, same class as funnelcake #626.
- `commentsById` does not dedupe by coordinate, so an edited revision arriving on an open sheet lands as an additional row beside the stale one. Separate defect.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart
+++ b/mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart
@@ -141,8 +141,14 @@ class CommentReactionsBloc
 
     try {
       final results = await Future.wait([
-        _likesRepository.getVoteCounts(event.commentIds),
-        _likesRepository.getUserVoteStatuses(event.commentIds),
+        _likesRepository.getVoteCounts(
+          event.commentIds,
+          addressableIds: event.addressableIdsByCommentId,
+        ),
+        _likesRepository.getUserVoteStatuses(
+          event.commentIds,
+          addressableIds: event.addressableIdsByCommentId,
+        ),
       ]);
 
       final voteCounts =
--- a/mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart
+++ b/mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart
@@ -245,10 +245,21 @@ class CommentReactionsBloc
       // upvotes in _likeRecords and downvotes in _downvoteRecords; pick the
       // right teardown call based on which side actually had it.
       if (hadSameVote || hadOppositeVote) {
-        if (wasUpvoted) {
-          await _likesRepository.unlikeEvent(commentId);
-        } else {
-          await _likesRepository.removeDownvote(commentId);
+        // A teardown miss must not abort the placement below. The repo's
+        // downvote cache is in-memory only, so after a cold start removing a
+        // pre-restart vote reports "not voted" even though the new vote still
+        // needs publishing — previously that threw past the placement and the
+        // UI showed a vote that was never sent (#6124).
+        try {
+          if (wasUpvoted) {
+            await _likesRepository.unlikeEvent(commentId);
+          } else {
+            await _likesRepository.removeDownvote(commentId);
+          }
+        } on NotLikedException {
+          // Nothing to remove; the optimistic update already reflects that.
+        } on NotDownvotedException {
+          // Nothing to remove; the optimistic update already reflects that.
         }
       }
 
@@ -279,14 +290,6 @@ class CommentReactionsBloc
             ..remove(commentId),
         ),
       );
-    } on NotLikedException {
-      // State-sync sentinel (silent): repo had no upvote to remove.
-      emit(
-        state.copyWith(
-          upvotedCommentIds: Set<String>.from(state.upvotedCommentIds)
-            ..remove(commentId),
-        ),
-      );
     } on AlreadyDownvotedException {
       // State-sync sentinel (silent): repo already had the downvote.
       emit(
@@ -294,14 +297,6 @@ class CommentReactionsBloc
           downvotedCommentIds: Set<String>.from(state.downvotedCommentIds)
             ..add(commentId),
           upvotedCommentIds: Set<String>.from(state.upvotedCommentIds)
-            ..remove(commentId),
-        ),
-      );
-    } on NotDownvotedException {
-      // State-sync sentinel (silent): repo had no downvote to remove.
-      emit(
-        state.copyWith(
-          downvotedCommentIds: Set<String>.from(state.downvotedCommentIds)
             ..remove(commentId),
         ),
       );
--- a/mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart
+++ b/mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart
@@ -252,7 +252,10 @@ class CommentReactionsBloc
         // UI showed a vote that was never sent (#6124).
         try {
           if (wasUpvoted) {
-            await _likesRepository.unlikeEvent(commentId);
+            await _likesRepository.unlikeEvent(
+              commentId,
+              addressableId: event.addressableId,
+            );
           } else {
             await _likesRepository.removeDownvote(
               commentId,
--- a/mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart
+++ b/mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart
@@ -254,7 +254,10 @@ class CommentReactionsBloc
           if (wasUpvoted) {
             await _likesRepository.unlikeEvent(commentId);
           } else {
-            await _likesRepository.removeDownvote(commentId);
+            await _likesRepository.removeDownvote(
+              commentId,
+              addressableId: event.addressableId,
+            );
           }
         } on NotLikedException {
           // Nothing to remove; the optimistic update already reflects that.
@@ -265,17 +268,24 @@ class CommentReactionsBloc
 
       // Place the new vote (only when this tap isn't a same-side removal).
       if (!hadSameVote) {
+        // NIP-25: `a` and `k` describe the event being reacted to, so both
+        // come from the target itself. A Kind 34236 video reply is
+        // addressable and carries a coordinate; a Kind 1111 comment does not
+        // and passes null (#6124).
+        final targetKind = event.targetKind ?? EventKind.comment;
         if (isUpvote) {
           await _likesRepository.likeEvent(
             eventId: commentId,
             authorPubkey: event.authorPubkey,
-            targetKind: EventKind.comment,
+            targetKind: targetKind,
+            addressableId: event.addressableId,
           );
         } else {
           await _likesRepository.downvoteEvent(
             eventId: commentId,
             authorPubkey: event.authorPubkey,
-            targetKind: EventKind.comment,
+            targetKind: targetKind,
+            addressableId: event.addressableId,
           );
         }
       }
--- a/mobile/lib/blocs/comments/comment_reactions/comment_reactions_event.dart
+++ b/mobile/lib/blocs/comments/comment_reactions/comment_reactions_event.dart
@@ -64,10 +64,22 @@ final class CommentVoteToggled extends CommentReactionsEvent {
 /// the loaded-comment set changes so the reactions cubit can populate counts
 /// for the visible list.
 final class CommentVoteCountsFetchRequested extends CommentReactionsEvent {
-  const CommentVoteCountsFetchRequested(this.commentIds);
+  const CommentVoteCountsFetchRequested(
+    this.commentIds, {
+    this.addressableIdsByCommentId = const {},
+  });
 
   /// The set of comment IDs to fetch vote counts for.
   final List<String> commentIds;
+
+  /// Own addressable coordinates for those of [commentIds] that have one,
+  /// keyed by comment id.
+  ///
+  /// Votes on an addressable target carry its coordinate, and an edit mints a
+  /// new event id — so without this the counts and the viewer's own arrow
+  /// read as zero/unvoted for every revision after the first (#6124). Empty
+  /// for threads of plain Kind 1111 comments.
+  final Map<String, String> addressableIdsByCommentId;
 }
 
 /// Report a comment (publishes Kind 1984 / NIP-56).
--- a/mobile/lib/blocs/comments/comment_reactions/comment_reactions_event.dart
+++ b/mobile/lib/blocs/comments/comment_reactions/comment_reactions_event.dart
@@ -30,6 +30,8 @@ final class CommentVoteToggled extends CommentReactionsEvent {
     required this.commentId,
     required this.authorPubkey,
     required this.vote,
+    this.addressableId,
+    this.targetKind,
   });
 
   /// The ID of the comment being voted on.
@@ -40,6 +42,20 @@ final class CommentVoteToggled extends CommentReactionsEvent {
 
   /// Whether this is an upvote or a downvote tap.
   final Vote vote;
+
+  /// The target's **own** addressable coordinate, when it has one.
+  ///
+  /// A Kind 34236 video reply is addressable, and an edit re-publishes it
+  /// under the same coordinate with a new event id — so without this the
+  /// reaction strands on the superseded id (#6124). Null for Kind 1111
+  /// comments, which are regular events.
+  final String? addressableId;
+
+  /// The kind of the event being reacted to, for NIP-25's `k` tag.
+  ///
+  /// Null when the source did not report one, in which case the handler falls
+  /// back to [EventKind.comment].
+  final int? targetKind;
 }
 
 /// Request to batch-fetch vote counts for a set of comments.
--- a/mobile/lib/screens/comments/comments_screen.dart
+++ b/mobile/lib/screens/comments/comments_screen.dart
@@ -396,8 +396,20 @@ class OutboxBridges extends StatelessWidget {
                 )
                 .toList();
             if (newIds.isEmpty) return;
+            // Video replies are addressable; their votes are only reachable by
+            // coordinate once the reply has been edited (#6124).
+            final addressableIds = <String, String>{};
+            for (final id in newIds) {
+              final coordinate = state.commentsById[id]?.addressableId;
+              if (coordinate != null && coordinate.isNotEmpty) {
+                addressableIds[id] = coordinate;
+              }
+            }
             ctx.read<CommentReactionsBloc>().add(
-              CommentVoteCountsFetchRequested(newIds),
+              CommentVoteCountsFetchRequested(
+                newIds,
+                addressableIdsByCommentId: addressableIds,
+              ),
             );
           },
         ),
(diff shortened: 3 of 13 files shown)


realmeylisdev:
## Real-device test guide

Deterministic script — pinned builds, pinned flag state, exact taps. Two testers following this should get identical results.

### Prerequisites

| | |
|---|---|
| Devices | 1 physical device minimum; 2 makes Scenario B easier |
| Accounts | **A** = video + reply author (does the editing) · **B** = the voter |
| Builds | **BEFORE** = `origin/main` @ `ccc454469` · **AFTER** = `fix/comment-votes-addressable-targets-6124` |

Build both from `mobile/`, flag pinned at compile time so it cannot drift between testers:

```bash
# BEFORE
git checkout ccc454469
flutter build apk --release --dart-define=FF_VIDEO_REPLIES=true
# AFTER
git checkout fix/comment-votes-addressable-targets-6124
flutter build apk --release --dart-define=FF_VIDEO_REPLIES=true
```

iOS: swap `apk` for `ios`. Install one build at a time — do not upgrade in place, the two scenarios below depend on a clean local cache.

> **Why the `--dart-define`:** `videoReplies` defaults ON in debug/profile and OFF in release. Pinning it removes build-mode as a variable. The runtime alternative is **Settings → Experimental Features → Video Replies** (8th of 13 cards, no developer mode needed) — use it only if you cannot rebuild, since a persisted toggle survives restarts and can silently differ between testers.

---

### Scenario A — cold-start downvote undo *(no flag needed; widest reach)*

Reachable in a stock release build with video replies **off**.

**Route:** any video → comments sheet → a comment row → downvote arrow (▼)

1. As **B**, open any video's comments and **downvote** any comment. Arrow turns red.
2. **Force-quit** the app (swipe from app switcher — not just background it).
3. Relaunch, navigate back to the *same* video's comments.
4. Tap the **same downvote arrow** to undo it.
5. Force-quit, relaunch, return to that comment thread once more.

| | BEFORE | AFTER |
|---|---|---|
| Step 4 | Arrow clears | Arrow clears |
| Step 5 | **Downvote is back** — it was never deleted | Arrow stays clear |

**Variant A2 — the switch.** Repeat steps 1–3, then at step 4 tap the **upvote** arrow instead.

| | BEFORE | AFTER |
|---|---|---|
| Immediately | Shows upvoted | Shows upvoted |
| After restart | **Upvote gone, downvote back** — the upvote was never published | Upvote persists, downvote gone |

---

### Scenario B — vote stranding on an edited video reply *(flag ON)*

**Route:** video → comments sheet → record-reply button → post → reply appears as a video row in the thread

1. As **A**, post a video. Open its comments and post a **video reply** (the record button in the comment composer; it only appears with the flag on).
2. As **B**, open the same video's comments. The video reply renders as a playable row **with vote arrows**.
3. As **B**, **upvote** the reply. Arrow turns green, score shows `1`.
4. As **A**, edit that reply: open it fullscreen → **pencil icon** (own content only) → change the title → **Save**.
   *Other routes to the same screen: share sheet → Edit · own-profile grid → long-press · feed settings menu. All land on `/video-edit/:videoId`.*
5. As **B**, force-quit, relaunch, reopen the same comments thread.

| | BEFORE | AFTER |
|---|---|---|
| B's arrow | **Un-fills** — reverts to grey | Stays green |
| Score | **Disappears entirely** (not `0` — the number is not drawn) | Still `1` |
| Any third viewer | Also sees no score | Sees `1` |

6. As **B**, tap the upvote arrow again, then tap it once more to remove it.

| | BEFORE | AFTER |
|---|---|---|
| Result | A **second** reaction is created; the first can no longer be removed from the UI at all | Single reaction, removes cleanly |

**Variant B2:** repeat with the **downvote** arrow — same expectations.

---

### Accessibility check (either scenario)

With TalkBack/VoiceOver on, focus a vote arrow you have already cast.

- **BEFORE**, after an edit: reverts to "Upvote comment"
- **AFTER**: stays "Remove upvote"

---

### If Scenario B won't reproduce

- **No record-reply button** → flag is off. Confirm the `--dart-define`, or toggle it in Settings and relaunch.
- **Reply row has no vote arrows** → you are looking at a plain text comment, not the video reply.
- **No pencil icon at step 4** → you are not signed in as **A**; edit is gated on ownership.
- **Reply never appears for B** → video replies arrive via the relay subscription rather than the first REST page; pull-to-refresh and give it a few seconds.

hm21 (approved):
Reviewed and took over for three findings. All four commits are mine and separate, so you can revert any of them independently.

**The diagnosis holds.** I verified every link on `main`: `_threadKinds` requests `[1111, 34236]`, `_eventToComment` has no kind guard, `_ActionsRow` renders vote arrows for every comment including video replies, `sendLike` emits `k` from `targetKind` and `a` only when `addressableId` is set, `video_metadata_update_service.dart:241` republishes under the same `d`, and `comment_item.dart` hides the score at `netScore != 0`. `FF_VIDEO_REPLIES` defaults on in non-release builds. Commit 1's cold-start premise checks out too — the reaction subscription runs with `limit: 1`, so `_downvoteRecords` really is empty after a restart while `getUserVoteStatuses` still paints the arrow from a relay read.

### What I pushed

**`955bc90cc5` — thread the coordinate into `unlikeEvent` as well.** The teardown branch passed `addressableId` to `removeDownvote` but left the sibling `unlikeEvent` keyed on the event id alone, though it has accepted a coordinate since #6123. Same edit scenario, so an upvote teardown on an edited reply missed, `NotLikedException` got swallowed, and the opposite vote was published on top of a still-live reaction.

**`7a8c772e8d` — derive coordinate and kind on the REST path too.** This is the one worth your attention. `loadComments` is REST-first, and `repository_providers.dart:727` wires the funnelcake client, so in production the first page comes from `_restCommentToComment` — which the PR didn't touch. It still emitted `eventKind: null` / `addressableId: null`, so on the path that actually serves users, votes kept carrying `k:1111` and no `a` tag and none of the rest of the fix engaged. That the REST payload really does deliver Kind 34236 replies is already pinned by your existing `maps video metadata from REST imeta tags` test. `_ownAddressableId` now takes kind/pubkey/tags instead of an `Event` so both mappers derive it identically. `VideoComment.kind` deserializes to `0` when absent, which I map back to null so the handler keeps its 1111 fallback instead of sending `k:0`.

**`4a2974ae33` — resolve counts and own status by coordinate.** Writing the `a` tag fixed the wire format, but both read paths the comment UI uses stayed `#e`-only, so symptoms 1–3 in your description survived the fix even for votes cast after it: after an edit the arrow still un-filled, the score still vanished, and since `getVoteCounts` has no `authors:` constraint the public aggregate still dropped to zero for everyone. Both now take an optional id→coordinate map and add an `a`-scoped filter; `_resolveVoteTarget` matches `e` first and falls back to `a`. The map is threaded from the `Comment` through `CommentVoteCountsFetchRequested`. Callers passing nothing keep the old behaviour, and the bloc is the only caller.

**`2ef44cf862`** is just `const` on two fixtures — the package analyze job treats `prefer_const_constructors` as failing and I'd missed it on the first push.

### Two things to check rather than take on trust

`getVoteCounts` used to increment once per matching `e` tag on a reaction; it now resolves one target per reaction and counts it once. For a NIP-25 reaction with a single `e` tag that's identical, and it's required anyway because a dual-tagged reaction arrives on both filters of the multi-filter REQ — but it does change behaviour for a hypothetical reaction carrying two `e` tags that are both in the queried set. Say the word if you'd rather keep the per-tag semantics.

Symptom 4 is still only partly addressed and I deliberately left it alone: `downvoteEvent`'s already-downvoted guard is in-memory only with no relay fallback, unlike `removeDownvote`. After a cold start on an edited target a re-tap can still mint a duplicate. Fixing it means a relay round-trip on the publish path, which is a judgment call about latency that's yours, not mine.

### Validation

`flutter analyze lib test integration_test` clean for the app and for both packages. comments_repository 146 tests, likes_repository 198 (coverage 95.70%, gate 62), comments blocs + screens 217. Every new regression test was run against the pre-fix implementation and confirmed red there — the two read-path tests are the first coverage `getVoteCounts` and `getUserVoteStatuses` have ever had. All 16 CI checks green.

Not device-verified. Approving on the pushed state; over to you for the merge decision.

NotThatKindOfDrLiz (approved):
Took over the remaining review items in f9bd677f7.

What changed:
- split coordinate vote reads into separate single-filter queries so NostrClient keeps merging the local cache for both #e and #a paths
- reused the coordinate-aware remote downvote lookup as downvoteEvent's cold-start guard, preventing duplicate downvotes after edited addressable targets
- made the like/downvote coordinate companion indexes pointer-safe when deleting older revisions
- corrected the stale Comment.eventKind doc and updated the related regression tests

Validation:
- flutter test test/src/likes_repository_test.dart from mobile/packages/likes_repository
- flutter analyze lib test from mobile/packages/likes_repository
- flutter test test/src/comments_repository_test.dart test/src/models/comment_test.dart from mobile/packages/comments_repository
- flutter analyze lib test from mobile/packages/comments_repository
- flutter analyze lib test integration_test from mobile
- flutter test test/blocs/comments/comment_reactions/comment_reactions_bloc_test.dart from mobile
- pre-push hook ran affected app/package tests successfully
- GitHub checks are green on f9bd677f7

@realmeylisdev please review the takeover commit, then merge when you're ready.

realmeylisdev:
Thanks both — `1405d18d` is the new tip. Three commits from me on top of `f9bd677f7`, each revertable alone.

**@hm21, your two questions.**

*(a) Per-target vs per-tag.* Keep per-target — but the ordering was wrong, and I fixed it in `0ac9f20fb`. NIP-25: "the target event `id` should be last of the `e` tags." `_resolveVoteTarget` scanned forwards and returned the first tag in the queried set, so a reaction from a client that copies the thread's ancestor tags got credited to the root comment instead of the reply. The old per-tag loop had the same forward bias but incremented on every match, so it at least also credited the true target — collapsing to one target per reaction turned "inflates the ancestor" into "inflates the ancestor and loses the target". One target per reaction is right, and required anyway now that a dual-tagged reaction comes back from both queries; it just has to be the last `e` tag. Divine's own writer emits a single `e` tag, so only other clients are affected. The sibling `_extractTargetEventId` has the identical bias across four unrelated read paths — left alone, not in this PR's blast radius.

*(b) Symptom 4.* Already closed — @NotThatKindOfDrLiz shipped the relay fallback in `downvoteEvent` as part of `f9bd677f7`, with a cold-start test. Nothing left for me to decide.

**Two more defects, fixed.**

`84e3a5a09` — `getVoteCounts` was the only reaction read path with no Kind 5 pass, while `getUserVoteStatuses` filters deletions and `_fetchResolvedLikersByEvent` additionally author-matches them. Invisible while the query was `e`-only, because a reaction on a superseded revision was unreachable. The coordinate leg exposes it on this PR's own headline flow: switching an upvote to a downvote on an edited reply publishes a Kind 5 for the pre-edit reaction plus a new downvote, the `a` leg returns both, and counting the retracted upvote gives netScore 0 — which `comment_item.dart` renders by hiding the score outright. The chunked, author-matched pass is now lifted into `_deletedReactionIds` and shared, so the two cannot drift again.

`1405d18d` — `_ActionsRow` is the only production producer of `addressableId`/`targetKind`, and nothing asserted what it sends. Swapping it to the parent video's `rootAddressableId`, or nulling the kind, left all 218 comments tests green. That is the line the whole fix hangs on and it encodes the PR's most consequential call, so it now has a test; both mutations turn it red.

I had independently made the same cache fix as `f9bd677f7` before it landed and dropped mine — the two-call shape is identical. Liz's also covers `_resolveRemoteDownvoteRecord`, which mine did too, and adds the identity guard on coordinate-index teardown, which I had missed. I widened the "queries e and a separately" assertion from two calls to three, since the deletion sweep is a third; its point — that no call may carry two filters — is unchanged.

**One correction, on `932524028` (the REST-path commit).** It is inert in production, and the evidence cited for it does not hold. funnelcake's comments read model is Kind 1111 only — `toUInt16(1111) AS kind` in the v2 read model, `WHERE e.kind = 1111` on both arms of the `comments_by_target` MV whose header states the table has no `kind` column, `WHERE kind = 1111` on every UNION arm of the legacy path, and the handler doc says NIP-22 Kind 1111. So `restComment.kind` is always 1111, `_ownAddressableId` fails its `kind >= 30000` guard for every REST row, and `addressableId` is null 100% of the time; `eventKind: 1111` is indistinguishable downstream from the null it replaced, because the handler falls back to `EventKind.comment`, which is 1111. The test cited as proof that 34236 replies arrive over REST is a hand-written mock fixture, not a captured response — its kind is the fixture author's choice, and it carries no `d` tag, so it could not have exercised coordinate derivation in principle.

I have left the commit in — the mapper is correct if that read model is ever widened, and it costs nothing — but it should not be read as fixing a production path.

The real REST-path exposure is different and larger: Kind 34236 replies never reach the first page at all, because `loadComments` returns on the REST response before the relay branch where `includeVideoReplies` is honoured. Filed as #6744. It does not block this PR — the coordinate plumbing is correct wherever a reply actually arrives, via load-more or when funnelcake is unavailable.

`flutter analyze lib test integration_test` clean for the app and both packages. likes_repository 203 tests at 96.97% (gate 62), comments_repository 146, comments blocs + screens 218. Every new test mutation-checked against the pre-fix code. Not device-verified.

PR #6740, opened 2026-08-05
fix(video-actions): show nip05 handles instead of npubs in find people

### LINKED ISSUE #6741: fix(share): show nip05 handles instead of npubs in find people
## Problem

The Find people sheet showed npub identifiers as the secondary line for share recipients because `ShareableUser` did not carry profile handle data. This made rows noisy and diverged from the design, which calls for a handle line when one is available and no secondary line otherwise.

## Expected behavior

- Divine NIP-05 users render with the full `@username.divine.video` handle in Find people.
- External NIP-05 values render without a doubled leading `@`.
- Profiles with no usable NIP-05 or name omit the secondary line instead of falling back to an npub.
- Malformed or blank unverified kind-0 NIP-05 values do not render as blank or bare subtitles.

## Scope

Fix the Find people/share-recipient display path only. Broader behavior of `UserProfile.handle` in other surfaces is out of scope for PR #6740.

## PR

Fixed by #6740.

hm21:
Closes #6741

## Description

**Design:** [Find people — Figma](https://www.figma.com/design/rp1DsDEUuCaicW0lk6I2aZ/UI-Design?node-id=10534-168879&m=dev&focus-id=13945-91140) (node `13945:91140`)

The find-people sheet showed a raw npub under every name. `ShareableUser` never carried the profile's nip05, so the contacts list had nothing to pass and every row fell through to the `normalizeToNpub` fallback. The design shows a handle line or nothing at all.

`ShareableUser.fromProfile` now derives the handle from `displayNip05` rather than `UserProfile.handle`. That choice is deliberate: `handle` shortens divine identifiers to `@username` and prefixes external ones unconditionally, which turns `dana@nostrplebs.com` into `@dana@nostrplebs.com`. `displayNip05` gives the full `@username.divine.video` the design asks for and leaves external domains intact. A profile with neither nip05 nor name now drops the secondary line instead of falling back to an npub.

| | before | after |
|---|---|---|
| divine user | `@username` | `@username.divine.video` |
| external nip05 | `@dana@nostrplebs.com` | `dana@nostrplebs.com` |
| only `name` | `@alice` | `@alice` |
| neither | npub | no second line |

The sheet also moves off a raw `showModalBottomSheet` onto `VineBottomSheet`, following [`user_picker_sheet.dart`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/lib/widgets/user_picker_sheet.dart). Besides the design-system chrome (drag handle, radius, surface) this hands both lists — contacts and search results — the sheet's scroll controller, so flinging either one drags the sheet, and it fixes content sitting behind the keyboard, which the hand-rolled `SizedBox(0.92 × screenHeight)` did not account for. `VineBottomSheet` paints its surface with a `ColoredBox`, so the result area carries its own transparency `Material` to keep the tiles' ink splashes visible. The search field also gets the correct `textInputAction: .search`.

**Related Issue:** 

## Out of Scope

`UserProfile.handle` still produces the doubled `@dana@nostrplebs.com` and the shortened `@username`. It is used by the new-message sheet, the people-lists picker and the conversation header, so fixing it there is a separate, wider change rather than a side effect of this one.

## Verification

- `flutter test test/widgets/find_people_sheet_test.dart` (15 tests, incl. new coverage for the divine handle, the external nip05, the no-handle row, the `VineBottomSheet` chrome on the `show` path, and dragging the contacts list to move the sheet)
- `flutter test test/blocs/share_sheet/share_sheet_bloc_test.dart test/services/video_sharing_service_test.dart test/widgets/share_video_menu_comprehensive_test.dart test/widgets/video_feed_item/actions/share_action_button_test.dart`
- All four design-system ceiling ratchets green; `raw_dialogs.txt` baseline regenerated to lock the `showModalBottomSheet` removal
- Manually verified on a real Samsung Galaxy: contacts list, search results, keyboard overlap, sheet drag

## Type of Change

- [x] 🧹 Code refactor

### DIFF
--- a/mobile/lib/services/video_sharing_service.dart
+++ b/mobile/lib/services/video_sharing_service.dart
@@ -6,7 +6,9 @@ import 'dart:async';
 import 'package:dm_repository/dm_repository.dart';
 import 'package:models/models.dart' hide LogCategory;
 import 'package:nostr_client/nostr_client.dart';
-import 'package:openvine/services/auth_service.dart';
+// `auth_service` exports an unrelated `UserProfile`; hide it so the models
+// one (used by [ShareableUser.fromProfile]) resolves unambiguously.
+import 'package:openvine/services/auth_service.dart' hide UserProfile;
 import 'package:profile_repository/profile_repository.dart';
 import 'package:unified_logger/unified_logger.dart';
 
@@ -16,16 +18,46 @@ class ShareableUser {
   const ShareableUser({
     required this.pubkey,
     this.displayName,
+    this.handle,
     this.picture,
     this.isFollowing = false,
     this.isFollower = false,
   });
+
+  /// Builds a shareable user from a (possibly missing) [profile].
+  ///
+  /// Uses `displayNip05` — the full `@username.divine.video` form the Figma
+  /// spec shows — rather than `UserProfile.handle`, which shortens divine
+  /// identifiers to `@username` and prefixes external ones into a doubled
+  /// `@alice@example.com`. Falls back to `@name`, then to `null` when the
+  /// profile carries neither, so the row can drop its handle line entirely.
+  factory ShareableUser.fromProfile(String pubkey, UserProfile? profile) {
+    final name = profile?.name;
+    return ShareableUser(
+      pubkey: pubkey,
+      displayName: profile?.bestDisplayName,
+      handle:
+          profile?.displayNip05 ??
+          (name == null || name.isEmpty
+              ? null
+              : '@${UserProfile.sanitizeDisplayName(name)}'),
+      picture: profile?.picture,
+    );
+  }
+
   final String pubkey;
 
   /// Populate from `UserProfile.bestDisplayName`, never the raw kind-0
   /// fields: this renders directly in the share sheet and find-people list,
   /// where a lone UTF-16 surrogate crashes the paragraph builder.
   final String? displayName;
+
+  /// Handle for the user (e.g. `@alice.divine.video` or `alice@example.com`).
+  ///
+  /// Populate via [ShareableUser.fromProfile]. `null` means the profile has
+  /// no `nip05` and no `name`, and the find-people row then renders no
+  /// secondary line at all rather than falling back to an npub.
+  final String? handle;
   final String? picture;
   final bool isFollowing;
   final bool isFollower;
@@ -313,13 +345,7 @@ class VideoSharingService {
         final profile = await _profileRepository.fetchFreshProfile(
           pubkey: query,
         );
-        return [
-          ShareableUser(
-            pubkey: query,
-            displayName: profile?.bestDisplayName,
-            picture: profile?.picture,
-          ),
-        ];
+        return [ShareableUser.fromProfile(query, profile)];
       }
 
       Log.debug(
@@ -425,14 +451,7 @@ class VideoSharingService {
       );
 
       // Add to front of list
-      _recentlySharedWith.insert(
-        0,
-        ShareableUser(
-          pubkey: pubkey,
-          displayName: profile?.bestDisplayName,
-          picture: profile?.picture,
-        ),
-      );
+      _recentlySharedWith.insert(0, ShareableUser.fromProfile(pubkey, profile));
 
       // Keep only recent 10 users
       if (_recentlySharedWith.length > 10) {
--- a/mobile/lib/services/video_sharing_service.dart
+++ b/mobile/lib/services/video_sharing_service.dart
@@ -20,8 +20,6 @@ class ShareableUser {
     this.displayName,
     this.handle,
     this.picture,
-    this.isFollowing = false,
-    this.isFollower = false,
   });
 
   /// Builds a shareable user from a (possibly missing) [profile].
@@ -36,9 +34,24 @@ class ShareableUser {
     return ShareableUser(
       pubkey: pubkey,
       displayName: profile?.bestDisplayName,
-      handle: profile?.displayNip05 ?? _handleFromName(profile?.name),
+      handle:
+          _handleFromDisplayNip05(profile?.displayNip05) ??
+          _handleFromName(profile?.name),
       picture: profile?.picture,
     );
+  }
+
+  /// [displayNip05] as a handle when it is already handle-shaped.
+  static String? _handleFromDisplayNip05(String? displayNip05) {
+    final trimmed = displayNip05?.trim();
+    if (trimmed == null ||
+        trimmed.isEmpty ||
+        RegExp(r'\s').hasMatch(trimmed) ||
+        !trimmed.contains('@') ||
+        trimmed == '@') {
+      return null;
+    }
+    return trimmed;
   }
 
   /// [name] as a handle: `@`-prefixed unless it already reads as one.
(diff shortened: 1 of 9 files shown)


mobile/lib/services/video_sharing_service.dart
```diff
@@ -16,16 +18,46 @@ class ShareableUser {
...
     this.isFollower = false,
   });
+
+  /// Builds a shareable user from a (possibly missing) [profile].
+  ///
+  /// Uses `displayNip05` — the full `@username.divine.video` form the Figma
+  /// spec shows — rather than `UserProfile.handle`, which shortens divine
+  /// identifiers to `@username` and prefixes external ones into a doubled
+  /// `@alice@example.com`. Falls back to `@name`, then to `null` when the
+  /// profile carries neither, so the row can drop its handle line entirely.
```
NotThatKindOfDrLiz: This factory is the core behavior change, but the current tests mostly cover it indirectly through UI paths and do not assert the important handle branches. Please add direct service/factory coverage for nip05 present (Divine + external), name-only fallback, and profile == null so future changes to displayNip05/name handling cannot silently regress the find-people row.

mobile/lib/services/video_sharing_service.dart
```diff
@@ -16,16 +18,46 @@ class ShareableUser {
...
+  factory ShareableUser.fromProfile(String pubkey, UserProfile? profile) {
+    final name = profile?.name;
+    return ShareableUser(
+      pubkey: pubkey,
+      displayName: profile?.bestDisplayName,
+      handle:
+          profile?.displayNip05 ??
+          (name == null || name.isEmpty
+              ? null
```
NotThatKindOfDrLiz: This fallback still produces doubled-@ output for common email-shaped kind-0 names, e.g. name: dana@nostrplebs.com becomes @dana@nostrplebs.com. That is the same bug class this PR avoids for NIP-05. Please normalize this branch as well, or add a test and explanation if that output is intentionally accepted.

NotThatKindOfDrLiz (changes requested):
Thanks for tightening the npub fallback and moving the sheet onto VineBottomSheet. I am requesting changes on the handle derivation before merge: the new factory is the core behavior change but its branches are not directly asserted, and the name fallback still reproduces the doubled-@ shape for email-like kind-0 names. Please also confirm the long-form displayNip05 choice against design/product, since nearby user surfaces use shortDisplayNip05/profile.handle; if long form is intentional here, update the displayNip05 docs or add local context so the exception is discoverable.

mobile/lib/services/video_sharing_service.dart
hm21: Added a direct `ShareableUser.fromProfile` group in `test/services/video_sharing_service_test.dart` — divine nip05 (long form), external nip05, name-only fallback, no nip05 and no name, and `profile == null`, plus the two normalization cases from the other thread. bf29bff3f

mobile/lib/services/video_sharing_service.dart
hm21: Agreed — fixed rather than accepted. The fallback moved into `_handleFromName`, which only prefixes when the name is not already handle-shaped, so `dana@nostrplebs.com` and `@dana` now pass through unchanged. Two tests pin both. bf29bff3f

hm21:
@NotThatKindOfDrLiz all three addressed in bf29bff3f:

- **Doubled `@`** — fixed, not accepted. The name fallback now only prefixes when the kind-0 `name` is not already handle-shaped, so `dana@nostrplebs.com` and `@dana` pass through unchanged.
- **Factory coverage** — new `ShareableUser.fromProfile` group with a case per branch: divine nip05, external nip05, name-only, both normalization cases, no-nip05-no-name, and `profile == null`.
- **Long-form `displayNip05`** — intentional, per the find-people design (Figma node `13945:91140`, linked at the top of the description); the full `@username.divine.video` is what the row spec shows. That is now recorded on the factory instead of only in this PR, and `displayNip05` lists the sheet among its consumers so the divergence from `user_picker_sheet` is discoverable from the model side.

`flutter analyze lib test` clean, 91 affected app tests and `packages/models` 851/851 green.

NotThatKindOfDrLiz:
@hm21 I pushed a takeover commit on top of your branch (`985b65069`) and rebased it onto current `origin/main`.

What changed:
- Hardened `ShareableUser.fromProfile` so `displayNip05` is trimmed and only used when it is handle-shaped; blank, whitespace-containing, lone-`@`, or no-`@` NIP-05 values now fall back to the name handle or no secondary line.
- Added regression coverage for malformed NIP-05 inputs.
- Removed the unused `ShareableUser.isFollowing` / `isFollower` fields and the repo-dead `getShareableUsers` / `searchUsersToShareWith` service methods plus their placeholder TODOs.
- Added bottom safe-area padding to both Find people lists, switched `_UserResultTile` to `StatelessWidget`, and moved touched text styles to `VineTheme.*Font()` helpers; raw textstyle baseline shrank accordingly.
- Filled in the auto-created issue body for #6741.

Local verification:
- `flutter analyze lib test`
- `flutter test test/services/video_sharing_service_test.dart`
- `flutter test test/widgets/find_people_sheet_test.dart`
- `flutter test test/blocs/share_sheet/share_sheet_bloc_test.dart test/widgets/share_video_menu_comprehensive_test.dart test/widgets/video_feed_item/actions/share_action_button_test.dart`
- `bash scripts/check_raw_textstyle_ceiling.sh`
- `bash scripts/check_raw_dialog_ceiling.sh`
- pre-push hook: analyze plus changed-file tests (`share_sheet_bloc`, `video_sharing_service`, `find_people_sheet`)

Please review the takeover commit and merge when you are ready, assuming CI comes back green.

hm21:
@NotThatKindOfDrLiz Takeover commit reviewed — looks good, thanks. I checked the two things that could have gone wrong and both are fine: dropping `getShareableUsers` / `searchUsersToShareWith` does not strand the recents list (`share_sheet_bloc.dart:99` still reads `recentlySharedWith`), and the bottom safe-area padding matches the `user_picker_sheet` precedent.

CI was red on `Tests (shard 3/4)`, but for something unrelated to this PR — worth flagging since it can bite any branch. `content_reporting_service_test` failed with `ReportDelivery.reached` instead of `localOnly`. Cause: `zendesk_support_service_test.dart` reset `ZendeskSupportService` in `setUp` but never in `tearDown`, so under the merged VGV isolate `_initialized == true` and the mock handler on `com.openvine/zendesk_support` outlived the file, and `ContentReportingService` then saw Zendesk as available and answering `createTicket`. With `--test-randomize-ordering-seed random` it only fires when that file happens to run first, which is why `main` is green and this PR drew it twice in a row.

Fixed in `114140b97`, which adds the missing `tearDown`. A/B on the branch head: bundling the two files and running with seed 3 gives the same 2 failures without the fix and passes with it. It is the only test file in the repo that touches Zendesk, so there is no second polluter.

All checks green on `114140b97`.

NotThatKindOfDrLiz (approved):
Approved after the takeover changes and the follow-up Zendesk test-isolation fix.

I rechecked the current head (`114140b97`): the handle derivation blocker is fixed, malformed/blank NIP-05 inputs are covered, touched dead code was removed, Find people list safe-area/style cleanup is in place, and the Zendesk tearDown fix is a narrow test isolation repair for the CI failure.

Local verification from the takeover pass was green, and GitHub checks are green on the current head.

Issue #6594, opened 2026-08-02
fix(local_stack): pinned GHCR images are 5 months stale, so NIP-17 is untestable locally


realmeylisdev:
## Problem

`local_stack` pins `ghcr.io/divinevideo/funnelcake-{migrate,relay,api}:latest`, but that GHCR mirror stopped
publishing. All three images are frozen at **2026-02-24**, so the local stack is a coherent but ~5-month-stale
snapshot — and `docker compose pull` reports "Pulled" while changing nothing, which makes it look fresh.

Evidence:

- `docker image inspect` → `Created: 2026-02-24T12:31:46` (migrate), relay and api same day.
- `docker run --rm --entrypoint sh ghcr.io/divinevideo/funnelcake-migrate:latest -c 'ls /migrations/*.up.sql | wc -l'`
  → **70**. The migrate container exits 0 having applied every migration it *contains*; it did not fail.
- `docker manifest inspect` for the last five `divine-funnelcake` `origin/main` commits → **all absent**.
- Root cause: `divine-funnelcake/.github/workflows/ci.yml:762-787` and `:961-978` push to GHCR with
  `continue-on-error: true` **by design** — "GHCR is a public mirror only — it must not gate deploys." GAR is
  authoritative; the mirror local_stack consumes can rot silently with nothing alerting.

## Impact

121 unapplied migrations. Concretely untestable locally today:

- **NIP-17 DMs** — `000079_allow_gift_wrap_kind` is unapplied, so the relay rejects gift wraps outright.
  Verified by publishing a genuinely BIP-340-signed event:
  `["OK","b155…",false,"blocked: kind 1059 not in allowed list"]` (a kind-0 in the same batch was accepted,
  proving the signatures were valid). Kind 10050 is rejected too.
- The NIP-86 `disallowed_kinds` kill switch (migration 194) — the table does not exist locally.
- `video_media_by_id` v2 read models (195-197).

Local NIP-11 advertises 16 supported NIPs; production advertises 24, including 17 and 59. **Do not generalise
any local relay kind behaviour to production.**

Blast radius on CI is nil, which is why it rotted unnoticed: the only workflow that starts the stack
(`mobile_service_integration_tests.yaml`) is `workflow_dispatch`-only and gates nothing.

## Possible fixes

- Pin image digests (or a dated tag) in `local_stack/docker-compose.yml` so staleness is visible rather than
  disguised by `pull_policy: always`.
[...]

PR #3806, opened 2026-05-01
fix(security): reject insecure ws/http relay schemes outside loopback (#3362)

realmeylisdev:
## Description

Closes #3362. Application-layer companion to #3358 / PR #3788's OS-level transport hardening — without this fix, the OS rejects cleartext but the app still accepts non-loopback `ws://` (and downgrades NIP-11 capability fetches to plain `http://`) from user input, NIP-46 deep links, and NIP-65 discovery.

After this PR: `wss://` is accepted for any host; `ws://` (and `http://` for the NIP-11 fetch path) is accepted only for the loopback allowlist `{localhost, 127.0.0.1, 10.0.2.2, ::1}` — the same set pinned in `network_security_config.xml` (Android), `NSAllowsLocalNetworking` (iOS/macOS), and `EnvironmentConfig.localHost` (Android emulator).

**Related Issue:** Closes #3362
**Companion to:** #3358 / PR #3788 (ship in the same release)

## Type of Change

- [x] 🛠️ Bug fix (non-breaking change which fixes an issue) — security hardening

## What changed

| Layer | File | Change |
|-------|------|--------|
| Source-of-truth helper | `mobile/lib/utils/relay_url_utils.dart` | New `isRelayUrlAllowed` + `isLoopbackHost` predicates. |
| Package (`nostr_sdk`) | `nip46/nostr_remote_signer_info.dart` | New typed `InvalidBunkerRelayException` (replaces generic `Exception`); `parseBunkerUrl` rejects non-loopback `ws://` for both `bunker://` and `nostrconnect://`. `toString()` deliberately omits the URL so log lines don't embed user-pasted relays. |
| Package (`nostr_client`) | `src/relay_manager.dart` | `_normalizeUrl` returns null for non-loopback `ws://`, propagating through all 8 add/remove/connect/configure callsites. Saved relay storage self-heals on next boot. |
| Package (`nostr_key_manager`) | `src/nsec_bunker_client.dart` | Catches typed exception; surfaces `e.reason` so the URL is no longer embedded in `BunkerAuthResult.error`. |
| UI | `relay_settings_screen.dart` | Two-tier validation: `relaySettingsInvalidUrl` for malformed input, new `relaySettingsInsecureUrl` for cleartext-to-non-loopback. |
| UI | `key_import_screen.dart` | Catches `InvalidBunkerRelayException` with new `keyImportInsecureBunkerRelay` l10n key. |
[...]

PR #6746, opened 2026-08-05
fix(nostr): bound and validate relay-list admission

### LINKED ISSUE #6585: fix(dm): recipient-supplied kind-10050 relays are connected to without validation or a count cap
## Problem

When sending a NIP-17 DM, the local client connects to relay hosts chosen by the **recipient**, with scheme validation only — no allowlist, no host filter, and no cap on how many.

`DmRepository.sendMessage` resolves the recipient's kind-10050 inbox list and passes it straight through as `targetRelays`:

- `mobile/packages/dm_repository/lib/src/dm_repository.dart:2853` — `resolveDmInboxRelays(recipientPubkey)`
- `:2864` — `targetRelays: inboxRelays`
- `:2510-2575` — the resolver reads every `relay` / `r` tag into an unbounded set

`RelayPool` then opens temp connections to those addresses. Because production has no relay-host lock (`nostr_service_factory.dart:57-62` returns a null `allowedRelayHost` for production), nothing downstream constrains them either.

## Why it matters

A remote party controls which hosts the local device makes outbound WebSocket connections to, and how many. At minimum that is an unbounded fan-out driven by untrusted input; it also exposes the sender's IP to hosts of the recipient's choosing.

Separately, the NIP-65 discovery reader performs no cryptographic verification of the kind-10002 it parses (`relay_discovery_service.dart:276-302` overrides `RelayBase.onMessage` directly, and `relay_base.dart:105-115` hands the raw decoded frame through), so the relay list admission path is similarly unvalidated.

## Suggested direction

Cap the inbox relay count, validate hosts, and decide explicitly whether recipient-supplied relays should be connectable at all or only used when they intersect a known set.

Found while investigating #3366.

realmeylisdev:
## Description

Relay URLs reach the client from three trust levels — app config, the signed-in user, and a
remote party's advertised relay list — but only one predicate has ever guarded them, checking
the scheme and nothing else. So a list somebody else controls was admitted on the same terms
as our own configuration: no host policy, no count limit, and on the NIP-65 path no
authenticity check at all.

Two admission paths are affected, fixed together because they share the predicate.

**DM inbox lists (kind 10050).** `_queryOwnDmInbox` read every `relay`/`r` tag into an
unbounded set. On the send path that list belongs to the person being messaged, and each
entry becomes an outbound WebSocket connection from this device carrying the gift wrap — so
a counterparty decided both how many hosts we dial and which, including addresses on the
sender's own network. Admission now depends on who authored the list: our own keeps the
loopback allowance the local Docker stack needs, a counterparty's is refused private,
loopback and link-local targets, and both are capped. The reply is also checked to be the
kind and author we asked for before any tag is read.

**NIP-65 relay lists (kind 10002).** The indexer query runs on a bare `RelayBase` with its own
`onMessage`, so the frame never passes through `RelayPool` — the one place inbound events are
checked. The parser read `tags` and nothing else, and the result is adopted through
`addRelays` into the persistent pool that serves every subsequent read and write. A frame now
has to parse as an event, carry kind 10002, be authored by the pubkey we asked for, hash to
its own id, and verify its signature. The 24h cache is re-admitted on read too, so filtering
only on write can't leave a day-long gap.

**Temp relay lifetime.** Nothing ever closed a temp relay: `_sendCollect` tore one down only
when the publish deadline had *already* expired, so any address that answered in time kept its
socket for the life of the process. A periodic sweep now closes them once idle and carrying no
subscription. Idleness rather than "this publish finished" is what makes it safe — temp relays
are keyed by URL and therefore shared, so per-publish teardown could cut a socket out from
under a concurrent caller. The indexer query likewise now bounds its own teardown.

**Inbound events are checked against the filter that asked for them.** `RelayPool` verified an
EVENT frame's id and signature and then handed it to whichever subscription carried the id,
without ever comparing it to the REQ's filter — `Filter.checkEvent` existed with zero callers
on the inbound path. A signed event is only evidence that *someone* wrote it, so an
author-scoped filter was not a boundary: any relay could answer any open subscription with
anything. Frames matching no filter on their subscription are now dropped before delivery and
before the cache fanout, and `NostrClient` applies the same check to query results before they
are cached.

That last part is broader than relay-list admission and is what a reviewer should look at
hardest — it changes what every subscription in the app receives, not just the two paths
above. It arrived via @NotThatKindOfDrLiz's takeover commit 4450dbfd6; the commits after it
correct three things in that change and re-verify the rest.

### Notes for review

- **Honouring recipient relays is required, not optional.** NIP-17: *"Clients MUST only publish
  events to the relays listed in the recipient's kind 10050 event."* So this bounds the
  behaviour rather than removing it. `tempRelays` is likewise load-bearing — PR #1396 added it
  because `targetRelays` alone only *filters* the pool, leaving a targeted publish reaching
  zero relays.
- **The caps are deliberately generous** — 8 for kind 10050 against the NIP's own "1-3 relays"
  guidance, 16 for kind 10002 against "2-4 of each category". The win is turning unbounded into
  a small constant; a tight cap would silently cost a real user reachability. Every rejection
  and truncation is logged.
- **Caps are applied at admission only**, not pool-wide. Every current `tempRelays` caller
  passes at most three app-controlled relays, so a pool-wide cap would add silent-truncation
  risk for no gain.
- **`checkEvent` matches ids and authors exactly, not by prefix.** The takeover had switched
  both to `startsWith`. Current NIP-01 says the opposite — *"The `ids`, `authors`, `#e` and
  `#p` filter lists MUST contain exact 64-character lowercase hex values"* — prefix matching
  was removed from the spec, and nothing here ever passes a short prefix. The sharp edge is
  that every string starts with `''`, so one blank entry made the constraint vacuous and the
  new gate a pass-through; `Nostr.publicKey` returns `''` until `refreshPublicKey` lands, so a
  cold-start `Filter(authors: [publicKey])` reaches exactly that shape. A test pins it.
- **A subscription parses its filters once, at construction.** `matchesEvent` runs per inbound
  frame and `Filter.fromJson` copies the whole id/author list, so parsing per event made the
  receive path scale with the size of the author set. Parsing up front also turns a malformed
  filter from a silent per-event drop — the parse sat inside `RelayPool`'s frame-handler
  catch-all — into a loud failure at subscribe time that names the caller.
- **Verification failures are logged and dropped, not `Reportable`** — a broken or hostile
  indexer isn't something the user can act on, and `error_handling.md`'s matrix defaults to
  not-reportable.
- **One drifted copy of the URL rule is deleted.** `dm_repository` carried its own predicate and
  its own loopback-host set; the rule now lives once in `nostr_sdk`, which already owned
  `isLoopbackHost`. This follows the "one predicate" direction taken in #3806.
- **Three existing assertions are inverted** — they asserted that a list arriving over the
  network could name `ws://localhost` or `ws://10.0.2.2`. The local-stack loopback allowance
  doesn't extend to a list somebody else authored.

**Related Issue:** Closes #6585

## Out of Scope

- Routing discovery through `RelayPool` to remove the `onMessage` bypass class entirely —
  architecturally cleaner but needs an auth-bootstrap rework. Worth its own issue. The new
  filter check does not cover it: discovery still owns its own `onMessage`, which is why the
  kind-10002 authenticity check is written there by hand.
- NIP-17 says a client "shouldn't try" when a recipient advertises no kind-10050;
  `resolveDmInboxRelays` returns null and callers fall back to the default pool. Pre-existing,
  and a behaviour change rather than a hardening one.
- `_sendCollect` doesn't normalise `tempRelays` through `RelayAddrUtil.handle` while
  `subscribe` does, so the same host can occupy two `_tempRelays` entries. Pre-existing; the
  new sweep reclaims both. Noted while writing the lifetime test.
- Seven call sites in `likes_repository` and `reposts_repository` build
  `Filter(authors: [_nostrClient.publicKey])` with no guard against the pre-auth `''`. The
  filter is meaningless either way and relays match nothing on it, so this wants its own fix
  rather than a relaxed predicate to paper over it.
- The sibling web client has the same class of gap; tracking separately.

## Verification

- `flutter analyze lib test integration_test` — clean; bare `flutter analyze` clean in
  `nostr_sdk`, `nostr_client` and `dm_repository`
- `nostr_sdk` — 478 tests, `dm_repository` — 562, `nostr_client` — 325
- `mobile/test/services` — 3736; with `test/repositories` and `test/providers` — 4439
- New tests were mutation-checked rather than assumed: the sweep guard, the sweep itself, the
  cache admission, the exact-match rule, the filter precompute, the `queryByFilters` guard and
  the cached-list shape were each reverted in turn to confirm the corresponding test goes red.
  Two assertions were found vacuous this way — they looked up temp relays by raw URL when the
  pool keys them differently — and were rewritten.
- The two new `integration_test/e2e` files carry `@Tags(['service'])` above the first import.
  Placed below the imports the runner never reads it — `package:test` takes suite metadata only
  from the first directive — so the files were silently untagged. Checked by discriminator:
  `--exclude-tags service` ran all six tests before the move and reports "No tests ran" after.
- End-to-end verification against the local relay isn't possible today (#6594 — the pinned
  funnelcake images predate gift-wrap support), so the socket-lifetime test drives the real
  `RelayPool` against local WebSocket listeners instead.

## Type of Change

- [x] 🧹 Code refactor

- Service integration workflow tracker: #4239

### DIFF
--- a/mobile/integration_test/e2e/dm_inbox_relay_admission_test.dart
+++ b/mobile/integration_test/e2e/dm_inbox_relay_admission_test.dart
@@ -65,6 +65,7 @@ final List<String> _surplusTargets = [
   for (var i = 0; i < 10; i++) 'wss://surplus-$i.example',
 ];
 
+@Tags(['service'])
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
--- a/mobile/integration_test/e2e/dm_inbox_relay_admission_test.dart
+++ b/mobile/integration_test/e2e/dm_inbox_relay_admission_test.dart
@@ -1,0 +1,245 @@
+// ABOUTME: E2E for #6585 — a DM is routed only to the relays admitted from
+// ABOUTME: the recipient's kind-10050, and still sends. Drives a real
+// ABOUTME: DmRepository + NostrClient over real sockets.
+// ABOUTME: Requires: NO Docker stack — every dependency here is local.
+
+import 'package:db_client/db_client.dart';
+import 'package:dm_repository/dm_repository.dart';
+import 'package:drift/native.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+import 'package:nostr_client/nostr_client.dart';
+import 'package:nostr_sdk/client_utils/keys.dart';
+import 'package:nostr_sdk/event.dart';
+import 'package:nostr_sdk/nostr.dart';
+import 'package:nostr_sdk/relay/relay_base.dart';
+import 'package:nostr_sdk/relay/relay_status.dart';
+import 'package:nostr_sdk/relay/web_socket_connection_manager.dart';
+import 'package:nostr_sdk/signer/local_nostr_signer.dart';
+import 'package:nostr_sdk/utils/relay_addr_util.dart';
+import 'package:nostr_sdk/utils/relay_url_policy.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
+
+import '../helpers/fake_relay.dart';
+
+/// Redirects the public hostnames a relay list advertises onto the in-process
+/// relay, and records every address it was asked to open.
+///
+/// Both halves matter. The redirect is what makes an end-to-end test possible
+/// at all: `isRemoteSuppliedRelayUrlAllowed` refuses private and loopback
+/// hosts by design, so a `ws://127.0.0.1` relay can never *be* an admitted
+/// target — it has to be reached through one. And the recording is the
+/// assertion: "this host was never dialed" is then something the test
+/// observes directly, rather than infers from a publish result.
+class _RecordingRedirectFactory implements WebSocketChannelFactory {
+  _RecordingRedirectFactory(this.port);
+
+  final int port;
+  final List<String> requested = [];
+
+  @override
+  WebSocketChannel create(Uri uri) {
+    requested.add(uri.toString());
+    return WebSocketChannel.connect(
+      Uri.parse('ws://127.0.0.1:$port${uri.path}'),
+    );
+  }
+}
+
+/// Hosts a counterparty must never be able to point this device at.
+const _hostileTargets = <String>[
+  'wss://192.168.1.50',
+  'wss://127.0.0.1',
+  'wss://169.254.169.254',
+  'wss://nas.local',
+  'ws://cleartext.example',
+];
+
+const _admittedTargets = <String>[
+  'wss://admitted-1.example',
+  'wss://admitted-2.example',
+];
+
+/// Surplus legitimate hosts, so the total exceeds [RelayListCaps.dmInbox].
+final List<String> _surplusTargets = [
+  for (var i = 0; i < 10; i++) 'wss://surplus-$i.example',
+];
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  const senderKey =
+      '1111111111111111111111111111111111111111111111111111111111111111';
+  const recipientKey =
+      '2222222222222222222222222222222222222222222222222222222222222222';
+  final senderPubkey = getPublicKey(senderKey);
+  final recipientPubkey = getPublicKey(recipientKey);
+
+  /// The recipient's kind-10050, signed by them, ordered so the admitted
+  /// hosts come first and the cap is provably first-N in tag order.
+  Map<String, dynamic> recipientInbox() {
+    final event = Event(
+      recipientPubkey,
+      10050,
+      [
+        for (final url in [
+          ..._admittedTargets,
+          ..._hostileTargets,
+          ..._surplusTargets,
+        ])
+          ['relay', url],
+      ],
+      '',
+      createdAt: 1700000000,
+    )..sign(recipientKey);
+    return event.toJson();
+  }
+
+  /// Builds the real stack: Nostr → RelayPool → RelayManager → NostrClient →
+  /// NIP17MessageService → DmRepository, with every socket landing on [relay].
+  Future<
+    ({DmRepository repository, Nostr nostr, _RecordingRedirectFactory factory})
+  >
+  buildStack(FakeRelay relay) async {
+    final factory = _RecordingRedirectFactory(relay.port);
+    final signer = LocalNostrSigner(senderKey);
+
+    RelayBase gen(String url) =>
+        RelayBase(url, RelayStatus(url), channelFactory: factory);
+    final nostr = Nostr(signer, [], gen, channelFactory: factory);
+
+    final relayManager = RelayManager(
+      config: RelayManagerConfig(
+        defaultRelayUrl: relay.url,
+        storage: InMemoryRelayStorage(),
+        // Without this, a relay that drops at teardown keeps retrying and
+        // the failure surfaces against whichever test is running next.
+        autoReconnect: false,
+        // Deliberately unset: this is the production shape, and setting it
+        // would filter every `*.example` target away — `_allowedRelays`
+        // collapses an all-filtered list to null, which falls back to the
+        // whole pool and would quietly hide what this test is measuring.
+      ),
+      relayPool: nostr.relayPool,
+    );
+    final client = NostrClient.forTesting(
+      nostr: nostr,
+      relayManager: relayManager,
+    );
+
+    final db = AppDatabase.test(NativeDatabase.memory());
+    addTearDown(db.close);
+
+    final repository = DmRepository(
+      nostrClient: client,
+      directMessagesDao: db.directMessagesDao,
+      conversationsDao: db.conversationsDao,
+      outgoingDmsDao: db.outgoingDmsDao,
+      userPubkey: senderPubkey,
+      signer: signer,
+      messageService: NIP17MessageService(
+        signer: signer,
+        senderPublicKey: senderPubkey,
+        nostrService: client,
+      ),
+    );
+    addTearDown(repository.stopListening);
+    // Close every socket before the fake relay goes away. `addTearDown` is
+    // LIFO and `relay.stop` was registered first, so this runs ahead of it —
+    // otherwise in-flight connects fail after the test has completed and get
+    // attributed to it.
+    addTearDown(nostr.relayPool.removeAll);
+
+    // Through the manager, not RelayPool.add: a publish with no explicit
+    // targets checks `connectedRelays`, which only the manager populates.
+    await client.initialize();
+
+    return (repository: repository, nostr: nostr, factory: factory);
+  }
+
+  group('#6585 DM inbox relay admission', () {
+    testWidgets('resolves only the admitted subset of a kind-10050', (
+      tester,
+    ) async {
+      final relay = await FakeRelay.start(reply: recipientInbox());
+      addTearDown(relay.stop);
+      final stack = await buildStack(relay);
+
+      final resolved = await stack.repository.resolveDmInboxRelays(
+        recipientPubkey,
+      );
+
+      expect(resolved, hasLength(RelayListCaps.dmInbox));
+      // First-N in tag order: the two admitted hosts, then surplus.
+      expect(resolved!.take(2), _admittedTargets);
+      for (final hostile in _hostileTargets) {
+        expect(resolved, isNot(contains(hostile)), reason: hostile);
+      }
+    });
+
+    testWidgets('never dials a host the recipient should not reach', (
+      tester,
+    ) async {
+      final relay = await FakeRelay.start(reply: recipientInbox());
+      addTearDown(relay.stop);
+      final stack = await buildStack(relay);
+
+      final result = await stack.repository.sendMessage(
+        recipientPubkey: recipientPubkey,
+        content: 'admission probe',
+      );
+
+      // Positive control. If the send did not succeed the harness is wrong,
+      // and every "was not dialed" assertion below would pass for that reason
+      // rather than because the filter worked.
+      expect(result.success, isTrue, reason: result.toString());
+
+      for (final hostile in _hostileTargets) {
+        final host = Uri.parse(hostile).host;
+        expect(
+          stack.factory.requested.where((u) => u.contains(host)),
+          isEmpty,
+          reason: 'dialed $hostile',
+        );
+      }
+
+      // Connection-level witness. Temp relays are keyed by the normalised
+      // address, so comparing raw URLs here would silently match nothing.
+      // `_sendCollect` does not normalise `tempRelays` while `subscribe`
+      // does, so the same host can be keyed either way. Accept both rather
+      // than pinning the inconsistency.
+      final admitted = {
+        for (final url in [..._admittedTargets, ..._surplusTargets]) ...[
+          url,
+          RelayAddrUtil.handle(url),
+        ],
+      };
+      for (final dialed in stack.nostr.relayPool.tempRelayUrls) {
+        expect(admitted, contains(dialed), reason: dialed);
+      }
+    });
+
+    testWidgets('still delivers when the recipient advertises no inbox', (
+      tester,
+    ) async {
+      // The `absent` fallback predates #6585 and must survive it: a recipient
+      // with no kind-10050 is still reachable via the default pool.
+      final relay = await FakeRelay.start();
+      addTearDown(relay.stop);
+      final stack = await buildStack(relay);
+
+      expect(
+        await stack.repository.resolveDmInboxRelays(recipientPubkey),
+        isNull,
+      );
+
+      final result = await stack.repository.sendMessage(
+        recipientPubkey: recipientPubkey,
+        content: 'fallback probe',
+      );
+
+      expect(result.success, isTrue, reason: result.toString());
+      expect(relay.publishedEventIds, isNotEmpty);
+    });
+  });
+}
--- a/mobile/integration_test/e2e/dm_inbox_relay_admission_test.dart
+++ b/mobile/integration_test/e2e/dm_inbox_relay_admission_test.dart
@@ -3,6 +3,7 @@
 // ABOUTME: DmRepository + NostrClient over real sockets.
 // ABOUTME: Requires: NO Docker stack — every dependency here is local.
 
+@Tags(['service'])
 import 'package:db_client/db_client.dart';
 import 'package:dm_repository/dm_repository.dart';
 import 'package:drift/native.dart';
@@ -65,7 +66,6 @@ final List<String> _surplusTargets = [
   for (var i = 0; i < 10; i++) 'wss://surplus-$i.example',
 ];
 
-@Tags(['service'])
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
--- a/mobile/integration_test/e2e/relay_list_admission_test.dart
+++ b/mobile/integration_test/e2e/relay_list_admission_test.dart
@@ -207,7 +207,6 @@ void main() {
       await pool.sendEventAwaitOk(
         ['EVENT', wrap.toJson()],
         eventId: wrap.id,
-        eventKind: wrap.kind,
         tempRelays: urls,
         targetRelays: urls,
         timeout: const Duration(seconds: 2),
--- a/mobile/integration_test/e2e/relay_list_admission_test.dart
+++ b/mobile/integration_test/e2e/relay_list_admission_test.dart
@@ -3,6 +3,7 @@
 // ABOUTME: released. Runs on-device against local WebSocket relays.
 // ABOUTME: Requires: NO Docker stack — every dependency here is local.
 
+@Tags(['service'])
 import 'dart:convert';
 
 import 'package:flutter_test/flutter_test.dart';
@@ -37,7 +38,6 @@ Map<String, dynamic> _signedRelayList({
   return event.toJson();
 }
 
-@Tags(['service'])
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
--- a/mobile/integration_test/e2e/relay_list_admission_test.dart
+++ b/mobile/integration_test/e2e/relay_list_admission_test.dart
@@ -4,7 +4,6 @@
 // ABOUTME: Requires: NO Docker stack — every dependency here is local.
 
 import 'dart:convert';
-import 'dart:io';
 
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
@@ -19,83 +18,7 @@ import 'package:nostr_sdk/utils/relay_url_policy.dart';
 import 'package:openvine/services/relay_discovery_service.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
-/// A relay that answers every REQ with one chosen frame.
-///
-/// Stands in for an indexer. The point of #6585 is that an indexer is a
-/// third-party server, so what it returns is untrusted input no matter that
-/// we were the ones who asked.
-class _StubIndexer {
-  _StubIndexer(this._server, this._reply);
-
-  final HttpServer _server;
-  final Map<String, dynamic>? _reply;
-
-  String get url => 'ws://127.0.0.1:${_server.port}';
-
-  static Future<_StubIndexer> start(Map<String, dynamic>? reply) async {
-    final server = await HttpServer.bind(InternetAddress.loopbackIPv4, 0);
-    final indexer = _StubIndexer(server, reply);
-    server.listen((req) async {
-      // Anything that is not a WebSocket upgrade (a NIP-11 capability probe,
-      // a health check) must be answered rather than thrown out of the
-      // listener — an async throw here fails the surrounding test.
-      if (!WebSocketTransformer.isUpgradeRequest(req)) {
-        req.response.statusCode = HttpStatus.badRequest;
-        await req.response.close();
-        return;
-      }
-      final socket = await WebSocketTransformer.upgrade(req);
-      socket.listen((raw) {
-        final frame = jsonDecode(raw as String) as List<dynamic>;
-        if (frame.isEmpty || frame[0] != 'REQ') return;
-        final subId = frame[1] as String;
-        if (indexer._reply != null) {
-          socket.add(jsonEncode(<dynamic>['EVENT', subId, indexer._reply]));
-        }
-        socket.add(jsonEncode(<dynamic>['EOSE', subId]));
-      }, onError: (_) {});
-    });
-    return indexer;
-  }
-
-  Future<void> stop() => _server.close(force: true);
-}
-
-/// A relay that accepts the connection and then says nothing — the shape a
-/// listener takes when its only purpose is to be connected to.
-class _SilentRelay {
-  _SilentRelay(this._server);
-
-  final HttpServer _server;
-  final List<WebSocket> _sockets = [];
-
-  String get url => 'ws://127.0.0.1:${_server.port}';
-  int get openSockets =>
-      _sockets.where((s) => s.readyState == WebSocket.open).length;
-
-  static Future<_SilentRelay> start() async {
-    final server = await HttpServer.bind(InternetAddress.loopbackIPv4, 0);
-    final relay = _SilentRelay(server);
-    server.listen((req) async {
-      if (!WebSocketTransformer.isUpgradeRequest(req)) {
-        req.response.statusCode = HttpStatus.badRequest;
-        await req.response.close();
-        return;
-      }
-      final socket = await WebSocketTransformer.upgrade(req);
-      relay._sockets.add(socket);
-      socket.listen((_) {}, onError: (_) {});
-    });
-    return relay;
-  }
-
-  Future<void> stop() async {
-    for (final socket in _sockets) {
-      await socket.close().catchError((_) => null);
-    }
-    await _server.close(force: true);
-  }
-}
+import '../helpers/fake_relay.dart';
 
 Map<String, dynamic> _signedRelayList({
   required String privateKey,
@@ -114,6 +37,7 @@ Map<String, dynamic> _signedRelayList({
   return event.toJson();
 }
 
+@Tags(['service'])
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
@@ -123,8 +47,8 @@ void main() {
 
   group('#6585 relay-list admission', () {
     testWidgets('a correctly signed kind-10002 is admitted', (tester) async {
-      final indexer = await _StubIndexer.start(
-        _signedRelayList(privateKey: authorKey),
+      final indexer = await FakeRelay.start(
+        reply: _signedRelayList(privateKey: authorKey),
       );
       addTearDown(indexer.stop);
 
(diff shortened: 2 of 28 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I pushed a takeover follow-up in 7def7e158 that closes the remaining relay trust-boundary gaps from review:

- DM kind-10050 inbox discovery now ignores relay-returned events unless they are the requested kind and author.
- RelayPool/Subscription now drop signed but off-filter EVENT frames before delivery or cache fanout, and NostrClient applies the same filter gate before caching query results.
- The kind-10002 signature rejection tests now actually reach the signature branch, and the new E2E files are tagged for the service workflow.
- Temp relay cleanup now disposes the RelayBase, and the NIP-65 E2E test uses the shared fake relay helper.

Local verification:
- flutter test test/unit/filter_search_test.dart test/unit/relay_pool_malformed_frame_test.dart (mobile/packages/nostr_sdk)
- flutter test test/src/nostr_client_test.dart (mobile/packages/nostr_client)
- flutter test test/src/dm_repository_test.dart (mobile/packages/dm_repository)
- flutter test test/services/relay_discovery_service_test.dart (mobile)
- flutter analyze lib test (nostr_sdk, nostr_client, dm_repository)
- flutter analyze lib test integration_test (mobile)
- pre-push hook analyzer/generated-file checks and changed-file test batch passed

I also tried running the two new service-tagged integration tests directly. The generic command stopped at Flutter device selection, and the macOS-specific run failed during local Xcode Swift Package Manager resolution before test code loaded. CI should be the source of truth for that workflow.

Given this PR changes relay admission and DM routing boundaries, I would like @dcadenas and @rabble to give it a fresh security-focused review before merge.

rabble (changes requested):
## Overview

Bounds and authenticates every relay list that arrives over the network:

- New `nostr_sdk/utils/relay_url_policy.dart` — one source of truth for `isRelayUrlAllowed` (self/app-supplied) vs `isRemoteSuppliedRelayUrlAllowed` (remote-chosen: `wss://` only, no private/loopback/link-local), plus `RelayListCaps` and `admitRemoteSuppliedRelays`. Removes the drifted copies in `mobile/lib/utils/relay_url_utils.dart` and `dm_repository`.
- `isPrivateOrLinkLocalHost` in `loopback_host.dart`, `dart:core` only so `nostr_sdk` still builds for web, and it judges `inet_aton` alternate encodings and IPv4-mapped IPv6 by the address they denote rather than by string shape.
- `RelayDiscoveryService` now verifies id, signature, author and kind on a kind-10002 before adopting it, takes newest-wins instead of first-frame, re-admits the 24h cache on read, and disposes the `RelayBase`.
- `DmRepository` splits self-authored from counterparty kind-10050 and fails closed by default.
- `RelayPool` grows an idle sweep for temp relays, plus a subscription-filter match gate on inbound events.

The threat model is stated clearly at every call site and the tests carry explicit positive controls, which is exactly right for a filter PR — several of the "was not dialed" assertions would otherwise pass vacuously. Full Nostr IDs everywhere, no key material in logs, `RelayListCaps` justified against the actual NIP text. Strong work.

Findings below are ordered by severity.

---

### 1. `Filter.checkEvent` now prefix-matches `ids` and `authors` — this weakens the gate the PR is introducing

`mobile/packages/nostr_sdk/lib/filter.dart`

```dart
if (ids != null && !_matchesAnyPrefix(ids!, event.id)) {
if (authors != null && !_matchesAnyPrefix(authors!, event.pubkey)) {
```

Three things make this look unintended:

- **Nothing needed it.** Before this PR `Filter.checkEvent` had zero production callers (`grep -rn checkEvent lib packages` finds only `event_sign_checker.dart`'s unrelated method and the definition itself). This PR is what turns it into an admission gate, in `Subscription.matchesEvent` and `NostrClient._eventsMatchingAnyFilter`. So the loosening isn't unblocking an existing caller — it's shipping a weaker gate than the exact match that was already there.
- **NIP-01 dropped prefix matching.** `ids`/`authors` are lowercase-hex exact values in the current spec; the "or prefixes" doc comments on lines 5 and 8 are stale text this change re-legitimises.
- **`''` matches everything.** `value.startsWith('')` is always true, so a filter carrying an empty string in `ids` or `authors` now admits every event the relay sends. With `contains` it admitted none.

For 64-char values `startsWith` is equality, so this is latent rather than currently-exploitable — but the whole point of the PR is not to rely on callers being well-formed. Suggest reverting to `contains` and updating the two stale doc comments. If some caller really does send prefixes, make it explicit (`Filter(prefixMatch: true)`) rather than global.

### 2. The new event gate is placed above `noteReceive()`, so a busy relay can be misread as silent

`mobile/packages/nostr_sdk/lib/relay/relay_pool.dart`, in `onEvent`

The `subscription == null` / `!subscription.matchesEvent(event)` early returns sit **before** `_broadcaseToCache(event)` and before `relay.relayStatus.noteReceive()`. Skipping the cache broadcast for an unsolicited event reads as intended. Skipping `noteReceive()` looks accidental, and it has teeth:

- `RelayBase.isSilentSince` drives silent-socket repair. A relay streaming events for a subId we just unsubscribed from (an ordinary unsubscribe race) now stops counting as traffic, so it can trip the repair path and churn a reconnect on a perfectly healthy socket.
- The same predicate is what `_isTempRelayReapable` calls, so this also makes a still-talking temp relay reapable.

Suggest moving `relay.relayStatus.noteReceive()` above the gate — liveness is a property of the socket, not of whether we wanted the frame.

### 3. `Subscription.matchesEvent` reparses every filter for every event

```dart
bool matchesEvent(Event event) {
  for (final filterJson in filters) {
    if (Filter.fromJson(filterJson).checkEvent(event)) return true;
  }
```

`filters` is fixed at construction, so this allocates a `Filter` (and up to 12 `List<String>.from` copies inside `fromJson`) per filter per event, on the relay-ingest hot path, for the life of every long-running feed subscription. In a video feed that's real GC pressure for no gain. Parse once:

```dart
late final List<Filter> _parsedFilters = [
  for (final json in filters) Filter.fromJson(json),
];
```

Worth a benchmark or at least a comment if you'd rather not, but the memoised form is strictly cheaper here.

### 4. The NIP-65 cap counts duplicates, so a repeated URL can starve the list

`RelayDiscoveryService._parseRelayListFromTags` appends to `relays` with no dedup, then breaks at `RelayListCaps.nip65`. A kind-10002 listing `wss://a.example` sixteen times fills the cap and drops every genuine relay after it. The cached-read path doesn't have this problem because it goes through `admitRemoteSuppliedRelays`, which collapses duplicates first — so the two paths disagree on the same input. Suggest deduping by URL before applying the cap, matching `admitRemoteSuppliedRelays`.

### 5. The sweep timer outlives the last temp relay on the direct-removal path

`_ensureTempRelaySweepScheduled` starts a `Timer.periodic`; it's cancelled in `removeAll` and inside `sweepIdleTempRelays` once `_tempRelays` empties. A direct `removeTempRelay(url)` that drains the map leaves the periodic timer running until the next tick discovers the map is empty. Harmless in effect, but it means the pool keeps a live timer after it has no temp relays at all. One line in `removeTempRelay`:

```dart
if (_tempRelays.isEmpty) {
  _tempRelaySweepTimer?.cancel();
  _tempRelaySweepTimer = null;
}
```

### 6. `_isTempRelayReapable` can reap a relay that is still connecting

```dart
if (relay.relayStatus.connected == ClientConnected.disconnect) return true;
```

`_getOrCreateTempRelay` calls `tempRelay.connect()` (async) and inserts the entry immediately, so a fresh entry sits in `disconnect` until the socket opens. The comment explains why the disconnected case has to be caught — agreed, otherwise failed sends leak entries forever — but as written a connect slower than one sweep tick gets torn down mid-handshake. At the 60s production interval that's a slow-network edge; with `tempRelayIdleTimeout: Duration.zero` and a manual sweep it's reachable directly. Consider stamping a creation time on the entry and requiring it to be older than one idle window before the disconnected branch reaps it.

---

### Nits

- **Cache path skips the new filter gate.** `NostrClient` applies `_eventsMatchingAnyFilter` to `websocketResult.events` only; cached events go through unfiltered. Given the cache can hold rows written by a build that predates this gate, the asymmetry re-opens the hole on the cache path. Filtering both, or a comment saying why the cache is trusted, would close it.
- **`_admitDmRelays` self-authored branch drops silently.** The remote branch gets `onRejected`/`onTruncated` logs; the self branch logs truncation but not rejection. A user whose own kind-10050 has a typo'd URL gets no signal.
- **`_isPrivateIPv4` gaps.** Missing `192.0.0.0/24`, `198.18.0.0/15` (benchmark), and the RFC 5737 doc ranges. No SSRF value, but if the intent is "not globally routable", worth either adding or a line saying the list is deliberately the reachable-LAN set.
- **`temp_relay_lifetime_test.dart` timing.** Real `HttpServer` sockets plus a fixed `Future.delayed(300ms)` for close frames to land. That's the shape that flakes first under CI load. Polling `openSockets` to zero with a deadline would be sturdier than a fixed sleep.
- **Stale doc comments** on `Filter.ids` / `Filter.authors` ("or prefixes") — see finding 1.

---

Findings 1 and 2 are the ones I'd want resolved before merge; 1 because it ships a gate weaker than the code it replaces, 2 because it can cause reconnect churn on healthy sockets. The rest are comfortably follow-up-or-fix-now.

rabble:
Heads-up on a cross-PR collision, separate from the review above.

#6731 (`chore(push): remove push-control relay diagnostics`) deletes `RelayPool.sendEventAwaitOk`'s `eventKind` parameter along with `PublishOutcome.eventKind` and `PublishTracker.eventKind`. This branch passes it in two new tests:

- `mobile/integration_test/e2e/relay_list_admission_test.dart`
- `mobile/packages/nostr_sdk/test/integration/temp_relay_lifetime_test.dart`

```dart
await pool.sendEventAwaitOk(
  ['EVENT', wrap.toJson()],
  eventId: wrap.id,
  eventKind: wrap.kind,   // deleted by #6731
  ...
);
```

Both branches are green today because neither carries the other's diff, and git will merge these hunks cleanly rather than conflicting — so whichever lands second gets a compile failure on an already-approved PR. Fix is just dropping the two `eventKind:` lines. Same note left on #6731.

realmeylisdev:
@NotThatKindOfDrLiz thanks — the kind+author guard on `_queryOwnDmInbox`, the `FakeRelay`
consolidation, the `dispose()` and the incremental newest-wins all land. I've kept the
inbound filter gate too and written it into the PR description, since it was the largest
part of the diff and nothing in the body mentioned it. Seven commits on top, one per finding.

Three corrections inside your change:

- **`checkEvent` is back to exact matching.** Current NIP-01: *"The `ids`, `authors`, `#e` and
  `#p` filter lists MUST contain exact 64-character lowercase hex values"* — prefix matching
  was removed from the spec, and nothing here passes a short prefix. The sharp edge is that
  every string starts with `''`, so a single blank entry made the constraint vacuous and the
  new gate a pass-through — and `Nostr.publicKey` returns `''` until `refreshPublicKey` lands,
  so a cold-start `Filter(authors: [publicKey])` reaches exactly that shape. Test pins it.
- **Filters parse once, in the `Subscription` constructor.** `matchesEvent` ran
  `Filter.fromJson` per event per filter, and `fromJson` copies the whole id/author list, so
  the receive path scaled with the size of the author set. It also sat inside `RelayPool`'s
  frame-handler catch-all — a filter map that tripped `List<String>.from` would have dropped
  every event on that subscription and logged nothing naming the caller. It now throws at
  subscribe time.
- **`queryByFilters` was missing the guard its siblings have.** It checked only the outer map,
  so a per-relay entry with an empty list built a subscription matching nothing — a filterless
  REQ, which many relays answer with everything, then every frame discarded.

Plus two in the discovery service: the cached read keyed by URL, which drops the second row
when a kind-10002 names one host twice for `read` and `write` (the fresh parse keeps both, so
the same event answered differently depending on which path served it); and the `CLOSE` before
teardown now passes `skipReconnect: true`, without which it walks the reconnect backoff and
holds the relay open for minutes — the new comment claimed that couldn't happen.

**On the tags:** `@Tags(['service'])` was below the imports, on `void main()`. `package:test`
takes suite metadata only from the first *directive*
(`test_core/parse_metadata.dart` → `directives.first.metadata`), so both files were still
untagged and `--tags service` matched nothing. `flutter analyze` can't catch it because
`invalid_annotation_target` and `library_annotations` are both suppressed in
`mobile/analysis_options.yaml`. Moved above the first import and checked by discriminator:
`--exclude-tags service` ran all six tests before, reports "No tests ran" after.

Worth knowing before relying on that workflow, though: `flutter test <dir>` builds and
launches every file in the directory *before* filtering by tag. I ran
`flutter test integration_test/e2e/ --tags service` and all six files loaded and failed,
including the four patrol suites, which can't run under `flutter test` at all. So correct
tags aren't sufficient — the patrol suites need separating out first. And
`mobile_service_integration_tests.yaml` is `workflow_dispatch` only and has never executed,
so it was never going to catch this either. I'll file that separately.

**CI hadn't run on 7def7e158** — zero Actions check runs and zero workflow runs on that SHA;
everything green on the PR belonged to its parent. This push should trigger it.

Verified locally: `nostr_sdk` 473, `nostr_client` 325, `dm_repository` 562,
`mobile/test/services` 3736 (4439 with `test/repositories` and `test/providers`),
`flutter analyze lib test integration_test` clean, bare `flutter analyze` clean in all three
packages. Each new guard was mutation-checked — reverting it turns its test red.

Agreed on @dcadenas and @rabble reviewing before merge. The part to look at hardest is the
inbound filter gate: it changes what every subscription in the app receives, not just the two
relay-list paths this PR is about.

realmeylisdev:
Device pass on the inbound filter gate, iOS 26.5 against production relays.

The gate is the part of this PR with the widest blast radius and the part no test drives with
real app filters, so I ran the app and measured it.

**First measurement was invalid, and the correction matters for reading the numbers.**
`RelayPool` logs through `dart:developer log()`, which does not reach `flutter run` stdout —
zero `📋 subscribe:` lines were in that stream, so nothing about the gate was observable
there. os_log carries only `print` output too. The real stream is the Dart VM Service
`Logging` channel.

**Then I checked the instrument instead of trusting it.** Mutating `Subscription.matchesEvent`
to `false`, rebuilding, and driving the same screens:

| | shipped | mutated to drop-all |
|---|---|---|
| gate drops | **0** | **297** |
| subscriptions observed | 16 | 9 |

So drops are visible when they occur, the gate is genuinely on the live receive path, and the
shipped code dropped nothing. Mutation reverted.

**Passed the gate, zero drops** — roughly 120 events across filter shapes
`{authors: […]}`, `{kinds: [10000], #p: […]}`, `{kinds: [30000], #p: […]}`,
`{kinds: [1059, 4, …]}`. That exercises `checkEvent`'s `#p` branch, which has no unit
coverage. For You feed 25 videos, profile 7, Discover and Categories populated, like counts
rendered, 16 kind-14 rumors decrypted from gift wraps and persisted.

**kind-10002 admission, end to end against the real indexers.** A signed relay list from
`purplepag.es` and `relay.nos.social` passed id, signature, kind and author and was adopted.

**Neither new throw fires in practice.** `queryByFilters`'s `ArgumentError` and
`Subscription`'s eager-parse `TypeError` — zero occurrences across all three log streams.
Those two are the changes here most capable of turning a previously-silent condition into a
crash, so they were the ones worth watching.

Two things I checked at source rather than inferring:

- `noteReceive()` is now skipped for events the gate rejects, and `isSilentSince` drives both
  zombie repair and the temp-relay reaper. Not a problem: the activity clock is stamped in
  `WebSocketConnectionManager._onMessage`, the raw socket receive, before parsing and before
  the gate. Neither can be starved by dropped frames.
- `removeTempRelay`'s new `dispose()` cannot poison re-adding the same URL — the map entry is
  removed first, so the next `checkAndGenTempRelay` mints a fresh relay, and
  `_forgetRelayBookkeeping` clears the old socket's judgment.

**Not covered, so treat as open:** the temp-relay sweep never armed — it needs a targeted
publish, and exercising it meant sending a real DM to a real person, which I did not do
unprompted. It has the branch's own socket-lifetime test against local listeners, but no
device pass. Hashtag/search `#t` relay filters and notifications also weren't reached; note
Categories turned out to go through funnelcake REST, not relay `#t` REQs.

No new bugs found in the diff. The only failures observed were environmental — funnelcake
`Connection refused` on `api.divine.video`, and two Keycast RPC decrypt timeouts.

**CI is still the blocker.** Zero GitHub Actions runs on `0a69b8f1e`, same as on `7def7e158`;
repo-wide the newest run is from 6 Aug and still shows `in_progress`. I've dispatched Mobile
CI manually against this branch. `nostr_sdk.yaml`, `dm_repository.yaml` and
`nostr_client.yaml` have no `workflow_dispatch` trigger, so those three cannot be started by
hand and stay blocked until `pull_request` events fire again — which is worth someone with
Actions access looking at, since it is not specific to this PR.

realmeylisdev:
Thanks — this was a genuinely useful review. All eleven items are addressed in eight commits (`13eaf922..7fb1d4f8`). Two of them turned out to be worse than described, and I owe you a correction on where the defect actually sits.

**1 — prefix matching, and the stale `ids`/`authors` doc comments.** Already reverted in `3a02fe5d3`, ~5 minutes after you submitted, so you were reading the right code. `checkEvent` is back to exact `contains`, the doc comments no longer say "or prefixes", and `filter_search_test.dart` pins the blank-entry case that makes a silent revert impossible.

**2 — the gate above `noteReceive()`. You're right, and it's worse than the mechanism you gave.** `noteReceive()` itself is inert for liveness: `isSilentSince` reads `WebSocketConnectionManager._lastActivityAt`, stamped on every raw frame at `web_socket_connection_manager.dart:250` before the pool sees it, so neither silent-socket repair nor `_isTempRelayReapable` could have been affected. `noteReceived` terminates in a stat row on the relay-settings screen.

But your closing principle — liveness is a property of the socket, not of whether we wanted the frame — is violated one statement further down. `_restartQuerySettleWindow` had also moved under the gate. That's the per-query liveness clock: a relay whose leading frames fail the filter stopped proving progress, the settle window expired mid-stream, `_releaseQuery` closed the REQ, and the caller got a truncated result set with `timedOut: false` — the #6673 failure the window exists to prevent. Both calls are hoisted above the gate now. Reverting the hoist turns the new test in `relay_pool_query_settle_test.dart` red with an **empty** result set instead of the matching tail.

**3 — reparsing filters per event.** Fixed in `a8dbcde86`; `Subscription` parses once in the constructor.

**4 — the NIP-65 cap counting duplicates. Correct, and it had spread.** Against `7def7e158` your reading of the cache path was exactly right; `7d458e833` (12 minutes after your review) aligned the two paths — but onto a record cap, so the cache path started starving too. Agreement by propagating the bug. Both now run through one admission helper that collapses duplicates *before* counting the cap, keyed on the normalized URL so a trailing slash isn't a second slot.

One amendment to your suggested remedy: `admitRemoteSuppliedRelays` collapses with a plain `Set<String>`, which is first-wins and would drop the second marker. The collapse ORs the read/write flags instead — per NIP-65 a host with a `read` row and a `write` row is both. That replaces the test asserting the two-row shape, which I'd justified badly: the pair never survived to a consumer, and the two that read the flags disagreed about which wins — `RelayListRepository._markerFor` takes the first and republished the host as `read`-only, the app bridge takes the last.

**5 — sweep timer on the direct-removal path.** Applied as suggested; cancel-on-empty now lives in `removeTempRelay` and the duplicate in the sweep is gone. For the record it self-healed on the next tick rather than leaking unboundedly, but the code now matches the invariant documented on the field.

**6 — reaping a connecting relay. Right line, and again a longer window than described.** The initial handshake is actually safe: `_setState(ConnectionState.connecting)` is the first statement of `_doConnect`, reached with no intervening await, and the microtask beats the sweep's timer. The reachable window is the *reconnect backoff* — 2s to 300s at `disconnect` with no subscription or query, where reaping calls `dispose()` and clears `_shouldReconnect`, cancelling the reconnect a caller is waiting on. On the NIP-17 path that's a gift wrap parked in `pendingAuthedMessages` behind a NIP-42 handshake losing its retry.

Queued frames now count as work, and the disconnected branch takes the creation-time grace you asked for via `RelayStatus.connectTime` — no new field. The branch had no test at all; there are now three, with a positive control, and each guard was reverted in turn to confirm its test fails.

**Nits.** Cache path now goes through the same filter — the DAO narrows on most dimensions but an empty-but-non-null list emits no condition where `checkEvent` matches nothing, so parity was a cross-package coincidence neither side declared. Self-authored `_admitDmRelays` logs rejections. `_isPrivateIPv4` takes the registry route: 192.0.0.0/24, 198.18.0.0/15, the RFC 5737 TEST-NETs, plus 2001:db8::/32 on the v6 side, each with the address either side of it pinned. `temp_relay_lifetime_test` polls with a deadline using the `_waitUntil` helper this package already had in `nostr_connect_session_test.dart` — and the pools are torn down now, since the subscription test was leaving a live temp relay and an armed sweep timer in the shared isolate.

Verification: `nostr_sdk` 478, `nostr_client` 326, `dm_repository` 562, `mobile/test/{services,repositories,providers}` 4518, `flutter analyze lib test integration_test` clean, bare `flutter analyze` clean in all three packages.

Findings 2 and 6 changed product behaviour rather than just tightening a filter, so those two are worth a second look.

NotThatKindOfDrLiz:
@realmeylisdev I pushed takeover follow-up 8853dd9d1 after rebasing this branch onto current origin/main.\n\nWhat changed:\n- Dropped the stale publish diagnostic API args removed by #6731: RelayPool direct callers no longer pass eventKind, RelayListRepository no longer passes diagnosticTag, and the affected mocks/PublishOutcome fixtures match the current API.\n- Closed the remaining cap-slot hardening gap in admitRemoteSuppliedRelays by deduping remote-supplied relay URLs on a normalized identity key before applying the cap, while preserving the first admitted spelling.\n- Added a regression test for case/trailing-slash relay variants consuming only one cap slot.\n\nVerification:\n- flutter test packages/nostr_sdk/test/unit/relay_url_policy_test.dart packages/nostr_sdk/test/integration/temp_relay_lifetime_test.dart\n- flutter test test/repositories/relay_list_repository_test.dart test/services/curated_list_service_crud_test.dart test/services/curated_list_service_collaboration_test.dart test/services/curated_list_service_persistence_test.dart test/services/curated_list_service_playlist_test.dart test/services/curated_list_service_query_test.dart test/services/curated_list_service_video_management_test.dart\n- flutter analyze lib test integration_test\n- flutter analyze in mobile/packages/nostr_sdk\n- pre-push hook analyzer, generated-file check, and changed-file test batch passed\n- GitHub checks are green on 8853dd9d1\n\nI could not run the service-tagged on-device integration test locally because Flutter required an explicit connected device for integration_test; CI is green. Please review the takeover commit and merge when you're ready.

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up. The branch is rebased onto current main, the #6731 publish API collision is fixed, remote-supplied relay dedup now uses normalized relay identity before cap counting, and GitHub checks are green on 8853dd9d1. @realmeylisdev should still review the takeover commit before merging.

NotThatKindOfDrLiz:
@realmeylisdev I took over the remaining review cleanup on top of the current PR branch.

What changed in the final push:
- rebased/aligned the branch with current main publish API by removing the stale eventKind: and diagnosticTag: call sites;
- kept the kind-10002 cap from discarding marker merges for relays already admitted after the cap is reached, with a regression test;
- added a Filter.checkEvent #t unit test pinning the current exact, case-sensitive matching contract;
- linked the service-integration workflow tracker in the PR body (#4239).

Verification on head 24edfea39c25e160760ce1d23e26ed011b400104:
- local pre-push hook passed analyzer, generated-file check, and changed-file tests;
- GitHub checks are green: Mobile CI, Nostr SDK CI, Nostr Client CI, DM Repository CI, preview build, mergeability, semantic PR, and CLA.

I am approving the current diff from my side. Rabble CHANGES_REQUESTED review still owns the final blocking re-review, especially for the broad inbound filter behavior.

NotThatKindOfDrLiz (approved):
Approved current head 24edfea39c25e160760ce1d23e26ed011b400104 after takeover cleanup. Local pre-push checks passed, and GitHub checks are green. Rabble blocking review still needs final re-review before merge.

realmeylisdev:
@rabble the branch was force-pushed and rebased since your review, so every SHA in my reply
above — and the one the PR body cited — was rewritten. Nothing was lost: `git range-diff`
reports all 25 commits as `=`, and the `patch-id` sets match 25/25. The eight commits that
answered your review are now `97b740d1e..7fef6d5e8` (was `13eaf922..7fb1d4f8`), still in the
same order, and the PR body now cites the rebased SHA.

**Your #6731 note is resolved.** `8853dd9d1` drops `eventKind` from both files you named.
#6731 is merged into `main` and this branch sits on top of it; `flutter analyze` is clean on
`mobile` and on all three packages, and the two call sites now match `sendEventAwaitOk`
argument-for-argument.

**Two commits landed after your review**, both from @NotThatKindOfDrLiz and both narrowing
things you raised:

- `8853dd9d1` — `admitRemoteSuppliedRelays` dedups on a normalized identity rather than the
  raw string, so `wss://h.example`, `WSS://h.example/` and `wss://h.example/` are one cap
  slot. That carries your finding 4 across to the DM path, which had been collapsing on the
  exact string.
- `24edfea39` — the kind-10002 cap branch went `break` → `continue`. Worth a look because it
  reads cosmetic and isn't: the marker-OR merge sits *above* the cap check, so under `break`
  the loop stopped at the first over-cap host and a later `write` row for an
  already-admitted relay never merged — the host stayed `read`-only. Pinned by a new test.

**One thing I fixed on top** (`92e3f5aab`). The identity key `8853dd9d1` added carried an
unreachable `?? url` fallback: admission and keying each parsed the same URL through the same
helper, so the predicate returning true already proved the parse succeeded. It reads as though
a raw URL can become a dedup key, which would quietly restore exact-string matching for that
entry. Admission and identity are now one nullable-returning helper — null means refused,
non-null is the identity — so there's one parse per entry on input that is untrusted and,
until the cap applies, unbounded. The two `.toLowerCase()` calls went with it; `Uri`
canonicalizes scheme and host at parse, checked across percent-escaped, IDN, IPv6, port,
userinfo, query and fragment forms. Behaviour unchanged, and both halves were
mutation-checked — dropping the root-path collapse and dropping the host gate each turn a
different test red.

On the part I flagged for you as widest-blast-radius, the inbound filter gate: I re-derived
whether `checkEvent` can ever be *stricter* than the relay that answered the REQ, since that
is the only way the gate drops something legitimate. It can't — it's exact on
`ids`/`authors`/`kinds`, correct on `since`/`until`, and ignores `search` entirely, so it is
lenient wherever it isn't spec-literal. `toJson`/`fromJson` round-trip symmetrically on
`#E`/`#A`/`#K`, which matters because kind-1111 comments carry both `E` and `e`.

Verification on `92e3f5aab`: `nostr_sdk` 478, `dm_repository` 562, `nostr_client` 325,
`mobile/test/{services,repositories,providers}` green, `flutter analyze lib test
integration_test` clean, bare `flutter analyze` clean in all three packages. CI was 17/17 on
`24edfea39` and is re-running now.

rabble (approved):
Rabble review: inline check while swarm runs — relay admission now validates host policy and count caps for kind-10050 DM inbox and NIP-65, with E2E tests over real sockets and explicit hostile-target assertions. Hosts wss://192.168..., 127.0.0.1, 169.254..., nas.local, cleartext are correctly rejected, surplus beyond RelayListCaps capped. LGTM — approving. Full swarm will post detailed analysis shortly.

PR #6755, opened 2026-08-06
fix(feed): record skipped videos as seen impressions

### LINKED ISSUE #6752: fix(feed): record skipped videos as seen impressions
## Summary

Videos swiped away in under one second are not recorded as seen locally, so For You freshness ordering cannot deprioritize content the viewer has actively skipped. This leaves a mobile-owned repeat path even after #6209 / PR #6212 shipped client-side seen-video ordering.

## Problem

`DivineVideoMetricsTracker` gates both analytics `view_end` publishing and local seen-video persistence behind a one-second watch threshold:

- `mobile/lib/widgets/divine_video_metrics_tracker.dart`: `_publishEvents` returns when `_totalWatchDuration.inSeconds < 1`.
- The same method calls `SeenVideosService.recordVideoView` only after that gate.
- `mobile/test/widgets/divine_video_metrics_tracker_test.dart` currently asserts that a 900 ms active-to-inactive session records neither `view_end` nor a seen-video record.

That threshold is reasonable for analytics view-end quality, but it should not prevent lightweight local seen/impression state. A user who swipes past a video quickly has still seen enough that repeating it immediately feels stale.

## Desired Outcome

Keep the >=1s threshold for analytics `view_end` publishing, but record a local seen/impression entry for an active video session even when watch duration is under one second. This should make skipped videos eligible for For You deprioritization without inflating analytics view-end events.

## Acceptance Criteria

- Sub-second active video sessions record local seen/impression state through `SeenVideosService` or an equivalent freshness path.
- Sub-second active video sessions still do not publish analytics `view_end` events.
- Existing >=1s sessions continue to publish `view_end` and record seen metrics once.
- Tests cover both sub-second and >=1s paths.
- No UI, l10n, or golden changes are expected.

## Related

- Duplicate/support signal: #6747
- Backend/session freshness tracker: #5027
- Mobile seen-video ordering shipped in #6209 / PR #6212
- Backend source-of-truth exclusion follow-up: #6246
- Feed/discovery epic: #5176

NotThatKindOfDrLiz:
## Description

Split DivineVideoMetricsTracker finalization so local seen impressions are recorded separately from analytics view_end events. Analytics still requires at least 1 second of watch time, while local seen persistence now records any active session with positive on-screen time, including videos that were visible but never started playing.

SeenVideosService now coalesces rapid persistence writes and flushes pending writes on disposal, avoiding a full SharedPreferences rewrite on every swipe while keeping in-memory seen state current.

**Related Issue:** Closes #6752

## Out of Scope

The separate feedDedupKey versus video.id seen-key migration is intentionally not included here because it changes persisted identity semantics and should be handled in a focused follow-up.

## Verification

- `flutter test test/widgets/divine_video_metrics_tracker_test.dart test/services/seen_videos_service_test.dart`
- pre-push checks: analyzer, generated-file verification, changed-file tests

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/providers/video_providers.dart
+++ b/mobile/lib/providers/video_providers.dart
@@ -117,7 +117,7 @@ PersonalEventCacheService personalEventCacheService(Ref ref) {
 SeenVideosService seenVideosService(Ref ref) {
   final service = SeenVideosService();
   unawaited(service.initialize());
-  ref.onDispose(service.dispose);
+  ref.onDispose(() => unawaited(service.dispose()));
   return service;
 }
 
--- a/mobile/lib/providers/video_providers.g.dart
+++ b/mobile/lib/providers/video_providers.g.dart
@@ -164,7 +164,7 @@ final class SeenVideosServiceProvider
   }
 }
 
-String _$seenVideosServiceHash() => r'933e55ea8a18eee2890da07089a3164a8ec424de';
+String _$seenVideosServiceHash() => r'b980fc7c86abd4229783e3c7066c8883d4d56558';
 
 /// Subscription manager for centralized subscription management
 
--- a/mobile/lib/services/seen_videos_service.dart
+++ b/mobile/lib/services/seen_videos_service.dart
@@ -74,11 +74,14 @@ class SeenVideosService {
   static const String _seenVideosMetricsKey = 'seen_video_metrics'; // New key
   static const int _maxSeenVideos =
       1000; // Limit storage to prevent unbounded growth
+  static const Duration _saveDebounceDuration = Duration(milliseconds: 100);
 
   final Map<String, SeenVideoMetrics> _seenVideos = {};
   SharedPreferences? _prefs;
   bool _isInitialized = false;
   Future<void>? _initializeFuture;
+  Timer? _saveDebounceTimer;
+  Completer<void>? _pendingSaveCompleter;
 
   /// Whether the service has been initialized
   bool get isInitialized => _isInitialized;
@@ -164,7 +167,7 @@ class SeenVideosService {
         );
 
         // Save in new format and remove legacy key
-        await _saveSeenVideos();
+        await _saveSeenVideosNow();
         await _prefs!.remove(_seenVideosKey);
       }
     } catch (e) {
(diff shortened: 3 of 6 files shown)


hm21 (approved):
Approving with one fix pushed as `aded07af`.

**Fixed:** the split introduced a single `_hasFinalizedView` latch set even when neither channel published. `_startTracking` only calls `_trackViewStart` when `!_hasTrackedView`, so the latch never cleared on inactive → active for the same video (app backgrounded, route pushed, or swiped past and back inside the pool). Measured against `main`:

| first segment, then reactivate + 5 s watch | `main` | this PR before the fix |
| --- | --- | --- |
| 300 ms | view_end 1, seen 1 | **view_end 0, seen 0** |
| 700 ms | view_end 1, seen 1 | view_end 0, seen 1 |

The 300 ms row is the one that matters — a glanced-then-watched video ended up recorded less than before the PR. Now each channel latches on its own publish, which restores `main`'s `view_end` behaviour and keeps the sub-second impression you added. Two regression tests cover both rows; `flutter analyze` and `test/widgets/video_feed_item/` (317) are green.

**Question:** the 500 ms impression floor means a 400 ms swipe records nothing, but #6752's acceptance criteria asks for sub-second sessions to record. Intentional product call, or should the floor go?

**Nit:** `coalesces rapid records while keeping memory current` races a real 100 ms `Timer` against `SharedPreferences.getInstance()` — it flips if those awaits ever exceed 100 ms on a loaded shard.

NotThatKindOfDrLiz:
Resolved in dbc26eb1a: removed the 500 ms impression floor, so any active session with positive on-screen time records local seen/impression state while view_end still requires >=1s watch time. Updated the 400 ms and 300 ms regression expectations, and made the SeenVideosService debounce test deterministic by injecting a long test-only debounce and flushing via dispose instead of racing the production 100 ms timer.\n\nVerification: flutter test test/widgets/divine_video_metrics_tracker_test.dart test/services/seen_videos_service_test.dart; flutter analyze lib test integration_test; pre-push analyzer/generated-file check/changed-file tests all passed.

NotThatKindOfDrLiz:
Temporarily closing/reopening to retrigger missing pull_request checks for current head c61bd27395439c867e0beea69bc939d9e089b493 after earlier Actions runs became stuck in queued state.

PR #6756, opened 2026-08-06
test(bluesky): document crosspost generation equality

### LINKED ISSUE #6753: task(bluesky): document or fix operationGeneration exclusion from CrosspostSettingsState props
## Description

`CrosspostSettingsState.operationGeneration` is accepted by the constructor and
threaded through `copyWith`, but it is **not** listed in Equatable `props`.

`mobile/lib/blocs/crosspost_settings/crosspost_settings_state.dart`

```dart
final int operationGeneration;

@override
List<Object?> get props => [
  status, enabled, username, handle, provisioningState, did,
  usernameClaimStatus, error, attempt,
  userLoadInFlight, pollLoadInFlight,
  provisioningPollAttempts, provisioningPollingTimedOut,
];  // operationGeneration absent, with no comment saying why
```

## Context

The generation counter is the mechanism that stops a stale provisioning poll
from overwriting a newer `toggleCrosspost()` result — it was added in PR #6705
to resolve a blocking review finding.

Today the omission is harmless: every site that bumps the generation also
changes a field that *is* in `props` (`userLoadInFlight` flipping, or `status`
moving to `toggling`), so no emit is dropped.

The hazard is latent. A future edit that bumps `operationGeneration` on its own
produces a state that Equatable considers identical to the previous one, `emit`
silently swallows it, and the stale-poll guard quietly stops working — with no
failing test and no analyzer complaint. Raised as a non-blocking follow-up in
@rabble's approving review on #6705:

> A future edit that bumps the generation *alone* would be silently swallowed by
> Bloc's equality check and the guard would quietly stop working. A comment on
> the field saying it's intentionally excluded and why would keep the next
> person from finding that the hard way.

## Implementation Notes

Two viable resolutions — pick one, don't do both:

1. **Keep it out of `props`** and add a `///` comment on the field stating the
   exclusion is deliberate, why (generation bumps always accompany a
   props-visible change, so including it would add no signal), and what the
   invariant is that future edits must preserve. Back it with a test that fails
   if a generation bump ever becomes the only delta.
2. **Add it to `props`.** Simpler and self-enforcing; costs an extra emit only
   in the case that is currently impossible. Verify no existing `blocTest`
   `expect:` list gains a state under this change.

Option 1 matches the reviewer's ask. Option 2 removes the trap entirely. Either
way the next reader should not have to re-derive the analysis.

## Acceptance Criteria

- [ ] `operationGeneration`'s relationship to `props` is either enforced by
      `props` membership or documented on the field
- [ ] A test covers the chosen invariant and can fail
- [ ] `flutter analyze lib test integration_test` clean
- [ ] `flutter test test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart` green

## Testing

`flutter test test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart`

Follow-up from PR #6705. Related to #6095.

NotThatKindOfDrLiz:
## Summary
- Document why CrosspostSettingsState.operationGeneration is deliberately excluded from Equatable props.
- Add invariant tests that prove current generation bumps emit with props-visible changes.
- Add retryProvisioning repeated-tap coverage so overlapping retries only start one setCrosspost call.

## Verification
- dart format lib test integration_test
- flutter test test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart

Closes #6753

### DIFF
--- a/mobile/lib/blocs/crosspost_settings/crosspost_settings_state.dart
+++ b/mobile/lib/blocs/crosspost_settings/crosspost_settings_state.dart
@@ -65,6 +65,17 @@ class CrosspostSettingsState extends Equatable {
 
   final bool userLoadInFlight;
   final bool pollLoadInFlight;
+
+  /// Epoch used by async completion guards to discard stale loads and toggles
+  /// after a newer user operation has started.
+  ///
+  /// This is deliberately excluded from [props]. Every current bump also
+  /// changes a props-visible field: `_beginUserLoad` sets [userLoadInFlight],
+  /// and `toggleCrosspost` moves [status] to
+  /// [CrosspostSettingsStatus.toggling]. That invariant is important because a
+  /// bump-only emit would be swallowed by Equatable; the new operation would
+  /// then discard its own completion when its local generation no longer
+  /// matched the unchanged state.
   final int operationGeneration;
   final int provisioningPollAttempts;
   final bool provisioningPollingTimedOut;
@@ -128,6 +139,7 @@ class CrosspostSettingsState extends Equatable {
     attempt,
     userLoadInFlight,
     pollLoadInFlight,
+    // operationGeneration deliberately excluded; see the field doc.
     provisioningPollAttempts,
     provisioningPollingTimedOut,
   ];
--- a/mobile/test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart
+++ b/mobile/test/blocs/crosspost_settings/crosspost_settings_cubit_test.dart
@@ -63,10 +63,7 @@ void main() {
         expect(cubit.state.status, CrosspostSettingsStatus.loaded);
         expect(cubit.state.enabled, isTrue);
         expect(cubit.state.handle, 'testuser.divine.video');
-        expect(
-          cubit.state.provisioningState,
-          AtprotoProvisioningState.ready,
-        );
+        expect(cubit.state.provisioningState, AtprotoProvisioningState.ready);
         expect(cubit.state.did, 'did:plc:test123');
         expect(cubit.state.usernameClaimStatus, UsernameClaimStatus.claimed);
       });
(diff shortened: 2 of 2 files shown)


rabble (approved):

This closes out both notes from my #6705 review, and the field doc is the part that earns its place — it states the invariant precisely (`_beginUserLoad` sets `userLoadInFlight`, `toggleCrosspost` moves `status` to `toggling`) and says what breaks if it stops holding: the bump-only emit gets swallowed by Equatable and the new operation discards its own completion. That's the failure mode spelled out at the field, which is where someone about to add a bump-only emit will actually be looking.

The two invariant tests can fail for the right reason. Both capture the pre-operation state, subscribe, and assert the first emission is `isNot(equals(previousState))` *and* carries the incremented generation — so a future bump-only emit shows up as a missing or equal first emission rather than as a silent guard failure three files away. `retryProvisioning ignores repeated taps while toggling` is the double-tap coverage I asked for and didn't get last time; holding `setCrosspost` on a `Completer` and asserting `.called(1)` is the right way to pin it.

One reservation, not blocking. `documents operationGeneration exclusion from equality` asserts that two states differing only in generation compare equal. That's a real guard — it fails if someone adds the field to `props` — but it locks in the *more* fragile of the two designs. Putting `operationGeneration` in `props` would make every bump emit unconditionally and remove the invariant this PR exists to document; the tradeoff is a few extra emissions, which is cheap on a settings screen. So the test as written would block the change that makes the whole class of bug impossible. If someone comes back to this wanting to simplify, the failing test shouldn't be read as "don't" — the field doc explains the reasoning, and that's the thing to re-litigate, not the assertion.

The unrelated `dart format` reflow through the test file is noise in the diff, but it's the formatter's output and the Format check is green, so it's consistent with CI's pinned version rather than a local-toolchain drift. Not worth splitting out.

Ship it.

Issue #6749, opened 2026-08-06
refactor(lists): curated-list feed pushes duplicate the route and disagree on the extra


rabble:
## Summary

Three widgets push the curated-list feed by hand, and they do not agree on what to hand the destination. #6617 extracted the owned-list case into `OwnedListCard`; the other two still build the route and its payload inline.

## The call sites

| Site | Extra it passes |
|---|---|
| `mobile/lib/widgets/owned_list_card.dart:35` (owned lists, both My Lists surfaces) | `listName` |
| `mobile/lib/screens/explore/tabs/explore_lists_tab.dart:311` (subscribed lists) | `listName` |
| `mobile/lib/screens/search_results/widgets/lists_section.dart:348` (`_navigateToCuratedList`) | `listName`, `videoIds`, `authorPubkey` |

The third is the interesting one — it seeds the destination with the video IDs and the author, the other two do not. So the same screen gets a materially different payload depending on which surface opened it, and nothing pins which shape is correct. Either the richer payload is load-bearing and the other two are shipping a degraded first paint, or it is redundant and the extra fields are dead weight. Worth settling before adding a fourth caller.

`mobile/lib/router/providers/page_context_provider.dart:837` also mentions `CuratedListFeedScreen.pathForId`, but it builds a URL for context resolution rather than pushing. Out of scope — leave it.

## Why this is worth doing

`.claude/rules/routing.md` warns off `extra` precisely because it is invisible to deep links and the web. Every hand-built `CuratedListRouteExtra` is a place where a route change has to be found by grep, and where the payload can quietly drift — which it already has.

## Suggested direction

1. Decide what `CuratedListFeedScreen` actually needs in its extra, and whether the fields beyond `listName` are an optimization or a requirement. If they are a requirement, the two thinner call sites are bugs.
2. Fold the remaining pushes behind one entry point — either widen `OwnedListCard` into a general curated-list card taking `showVisibility`, or add a single `CuratedListFeedScreen.push(context, list)` helper that owns the path and the extra together.
3. Prefer resolving the payload inside the destination from the list id where it can, per #3335 (replace critical `GoRouter.extra` state with URL-resolvable routes) — that is the version that survives a deep link.

Follow-up to #6617 (Q4). Related: #3335.

PR #6760, opened 2026-08-06
fix(lists): show creator on subscribed curated lists

### LINKED ISSUE #6757: fix(lists): show creator on subscribed curated lists
## Summary

A user reported that subscribed curated lists do not show who created the list, even though the creator is visible while searching or discovering lists.

This is the user-facing symptom of the route payload drift tracked in #6749: search/discovery paths pass authorPubkey and videoIds, while subscribed/owned/deep-link paths can arrive with only the list ID/name. The stored subscribed list already has the creator pubkey and video IDs, so the destination should resolve from local list data when route extras are missing.

## Acceptance criteria

- Subscribed curated-list cards show the creator when pubkey is available.
- Curated-list detail resolves author attribution from the stored list when route extras are missing.
- Curated-list detail no longer falls back to a zero count when the stored list has video IDs.
- Owned-list detail suppresses creator attribution.
- Lists with no pubkey continue to fall back cleanly.

Related: #6749. This issue is scoped to the user-visible bug; #6749 should remain open for the broader route-payload consolidation.

NotThatKindOfDrLiz:
## Summary

- show curated-list creators on shared list cards when the list has a creator pubkey
- resolve creator attribution on the curated-list detail screen from stored list data when route extras are missing
- keep owned-list detail views focused on count/visibility instead of showing a creator row for the current user

## Root cause

Search and discovery routes passed richer curated-list extras, but subscribed, owned, and deep-link paths can arrive with only the list ID/name. The stored subscribed list already has the creator pubkey and video IDs, so the detail screen should not depend on route extras for those fields.

Closes #6757.
Related to #6749, which should stay open for the broader route-payload consolidation.

## Verification

- `flutter test test/screens/curated_list_feed_screen_test.dart test/widgets/list_card_test.dart`
- `flutter analyze lib/screens/curated_list_feed_screen.dart lib/widgets/list_card.dart test/screens/curated_list_feed_screen_test.dart test/widgets/list_card_test.dart`
- pre-push hook reran analyzer and changed-file tests successfully

### DIFF
--- a/mobile/lib/screens/curated_list_feed_screen.dart
+++ b/mobile/lib/screens/curated_list_feed_screen.dart
@@ -298,7 +298,9 @@ class _CuratedListFeedScreenState extends ConsumerState<CuratedListFeedScreen> {
     final videoCount =
         widget.videoIds?.length ?? localList?.videoEventIds.length ?? 0;
     final videoText = context.l10n.listVideoCount(videoCount);
-    final authorPubkey = widget.authorPubkey;
+    final authorPubkey = _isOwnedList()
+        ? null
+        : widget.authorPubkey ?? localList?.pubkey;
 
     if (authorPubkey != null) {
       return GestureDetector(
@@ -319,7 +321,7 @@ class _CuratedListFeedScreenState extends ConsumerState<CuratedListFeedScreen> {
             Flexible(
               flex: 0,
               child: UserName.fromPubKey(
-                widget.authorPubkey!,
+                authorPubkey,
                 style: TextStyle(
                   color: context.vineColors.primaryText,
                   fontSize: 12,
@@ -488,9 +490,9 @@ class _CuratedListFeedScreenState extends ConsumerState<CuratedListFeedScreen> {
         stackTrace: stackTrace,
       );
       if (mounted) {
-        ScaffoldMessenger.of(context).showSnackBar(
-          SnackBar(content: Text(context.l10n.listShareFailed)),
-        );
+        ScaffoldMessenger.of(
+          context,
+        ).showSnackBar(SnackBar(content: Text(context.l10n.listShareFailed)));
       }
     }
   }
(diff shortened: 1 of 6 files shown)


hm21 (commented):
`CuratedListCard` now renders the author row for any list with a pubkey, but `OwnedListCard` wraps that same card — so the profile Lists tab (`profile_lists_grid.dart:81`) and the explore "My Lists" section (`explore_lists_tab.dart:147`) show "By <your own name>" on your own lists. `_createList` always stamps `pubkey: ownerPubkey`, so it hits every list the user creates. That's exactly what `_buildSubheading` deliberately suppresses for owned lists in this same PR. The owned-card tests stay green only because their `CuratedList` fixtures leave `pubkey` null — set one and it reproduces.

Secondary, if the row does stay on owned cards: `UserName.fromPubKey` here has no `neverGenerateName: true`, so until the profile resolves the owner is shown a generated "Adjective Animal NN" handle credited as the creator of their own list — the #6423 case that flag exists for.

Detail-screen half looks right; `listByAuthorPrefix` is already translated in every locale and the changed tests pass locally.

hm21 (approved):
Approved with a review fix pushed as `c24782f9a` (separate commit, no force-push).

**Finding:** `CuratedListCard` credited the creator unconditionally, and `OwnedListCard` wraps that same card — so the profile Lists tab and the explore "My Lists" section credited every list back to its owner. `CuratedListService._createList` stamps the owner pubkey on every list it creates, so it hit all of them. The detail screen in this PR already suppresses that row for owned lists.

**Fix:** explicit `showAuthor` flag on `CuratedListCard` (default on), `OwnedListCard` opts out. The owned-card fixtures left `pubkey` null, which is why nothing caught it — they now carry one like production does, plus a test that the owner is not credited.

**Validation:** `flutter analyze lib test integration_test` clean; `list_card`, `owned_list_card`, `profile_lists_grid` and `curated_list_feed_screen` suites green (28 tests). Not device-verified — please sanity-check the two My Lists surfaces before merging.

Rest of the PR looks right: `listByAuthorPrefix` is already translated in every locale, and the detail-screen resolution handles the loading window correctly (`service` is set before `initialize()` resolves, so `getListById` returns null rather than briefly crediting an owned list). Yours to merge.

PR #6766, opened 2026-08-06
fix(lists): handle addressable list-add notifications

NotThatKindOfDrLiz:
Relates to #6391

## Summary
- keep list_add notifications for addressable video coordinates by anchoring on root_addressable_id when event ids are absent
- dedupe list-add actors and suppress avatar overflow so grouped video counts do not imply extra curators
- collapse list-add row copy to a single ICU plural key and refresh generated l10n

## Verification
- flutter gen-l10n
- flutter test test/src/notification_repository_test.dart (packages/notification_repository)
- flutter test test/notifications/widgets/video_notification_row_test.dart
- flutter analyze lib test (packages/notification_repository)
- flutter analyze lib/notifications test/notifications test/l10n
- flutter analyze lib test
- flutter test test/notifications/routing/notification_tap_target_test.dart
- pre-push hook: analyze and changed-file tests


rabble (changes requested):
## Overview

Makes list-add notifications work when the video is referenced by addressable coordinate rather than event id: `_videoAnchorEventId` falls through to `rootAddressableId`, a new `_listAddVideoAddressableId` picks a coordinate that is genuinely one of the user's NIP-71 videos, and the event-id collection stops handing coordinates to `getVideoStats`. Alongside that, list-add rows dedupe actors by pubkey, count by notification rather than source event, and collapse the singular/plural copy into one ICU plural key.

The addressable-anchor fix is the right shape and the `verifyNever(getVideoStats(videoCoordinate))` assertion is exactly the check that proves it. Two things need attention before merge.

---

### 1. `_serverTypesForFilter` now returns `null` for `system` and `newPost`, unexplained

```dart
-  NotificationKind.system => const ['system'],
-  NotificationKind.newPost => const ['newPost'],
+  NotificationKind.system => null,
+  NotificationKind.newPost => null,
```

Nothing in the diff, the comments, or the PR title explains this, and it isn't list-add. `null` in this switch is the same value `filter == null` returns — "don't narrow by type at all" — so selecting the System or New Posts tab now fetches every notification type from the server and discards the non-matching ones client-side in `_filterItems`.

Two consequences:

- **Page dilution.** A server page of N rows can yield zero System rows while `hasMore` stays true. Infinite scroll then looks stuck, and the user pages through unrelated likes to surface one system notice.
- **It breaks a documented invariant.** `_belongsInFeed`'s docstring says it *"Mirrors the server-side `types` mapping in `_serverTypesForFilter` for rows that were fetched without it (cache hydration, cross-feed seed)"*. After this change the two deliberately disagree for two kinds, and the docstring is now wrong.

If the reason is that the inbox API doesn't recognise `system` / `newPost` as type strings, that's a fine reason — but it needs to be in a comment at the switch arm, the `_belongsInFeed` docstring needs updating, and the pagination behaviour needs a decision (page until N matches, or accept the dilution and say so).

### 2. `totalCount` now means something different per kind, and the row papers over it

```dart
final overflowCount = notification.type == NotificationKind.listAdd
    ? 0
    : notification.totalCount - notification.actors.length;
```

That's correct as far as it goes — with `countIds = unionNotificationIds` for list-add and actors deduped by pubkey, `totalCount - actors.length` is a videos-minus-people subtraction that means nothing, so `+N` had to go. But it leaves `VideoNotification.totalCount` carrying three different meanings depending on `type`:

- `mention` → number of actors
- `listAdd` → number of notifications (≈ videos added)
- everything else → number of source events

and the presentation layer is now the place that knows which. Either document the per-kind contract on `totalCount` itself, or give list-add its own field (`addedVideoCount`) so the row renders without a type check. The latter also removes the risk that the next kind added here inherits the wrong arithmetic silently.

---

### Smaller points

- **`_listAddVideoAddressableId` returns the first match in `group`**, and `group` was sorted `createdAt` descending immediately above, so for a multi-video list-add the row anchors on the most recently added video. That's a defensible choice but it's implicit — one line of comment would stop the next reader assuming it's arbitrary.
- **The guard reads well but is doing two jobs.** `_parseAddressableId(anchorEventId) == null` in the event-id collection means "this is an event id, not a coordinate". A named helper (`_isEventId`) would say that directly; as written the reader has to invert a parse to get there.
- **`notifications_view.dart` comment lost information.** `"Exhaustiveness-only: VideoNotification currently always resolves to OpenVideoTarget because it always carries a video target"` became `"Exhaustiveness-only for current VideoNotification policies"`. The old version said *why*; the new one doesn't, and `notificationKindOpensComments`'s own doc still asserts the old reason ("list-add push taps fall back to the video target"). Worth keeping the explanation.
- **ARB.** The plural collapse is right and `notificationAddedYourVideoToList` is cleanly removed from all 22 locales with no dangling references (checked `lib/` and `test/`). Note the surviving `notificationAddedYourVideosToList` value is still English in every non-English locale — pre-existing, not introduced here, but it now carries ICU plural syntax, so whoever translates it has to get `=1`/`other` right per locale (Polish and Arabic have more plural categories than English does). Might be worth an `@description` note to that effect.
- **`_filterItems` alignment is a genuine improvement** — routing all filters through `_belongsInFeed` instead of special-casing comments removes a real divergence. Worth calling out in the PR body since it changes behaviour for kinds unrelated to list-add.

NotThatKindOfDrLiz:
Takeover update pushed in `addf6a68d`.

What changed since the requested-changes review:

- Restored the `_serverTypesForFilter` behavior for `system` / `newPost` instead of widening those filters.
- Split list-add video count from actor count with `VideoNotification.addedVideoCount`, so `totalCount` is back to actor/avatar semantics and the row no longer needs list-add-specific overflow arithmetic.
- Moved addressable list-add validation into the anchor-selection path: list-add coordinates now have to parse as NIP-71 video coordinates for the recipient pubkey before they can render.
- Kept grouped list-add thumbnail/tap target selection tied to the same newest validated notification, covering mixed event-id/addressable groups.
- Preserved `addedVideoCount` through block filtering and page-merge paths.
- Named the event-id guard and applied it consistently before video metadata fetches, so coordinates are not sent to `getVideoStats`.
- Restored the routing/view comments with the missing explanation and kept the ARB plural cleanup.
- Added a small compatibility repair for existing app callers/tests: `publishEventAwaitOk` accepts `diagnosticTag`, and `PublishOutcome` can carry `eventKind`.

Local verification:

- `flutter pub get`
- `flutter gen-l10n`
- `dart format ...`
- `flutter test test/src/video_notification_test.dart` from `mobile/packages/models`
- `flutter test test/src/notification_repository_test.dart` from `mobile/packages/notification_repository`
- `flutter analyze lib test` from `mobile/packages/notification_repository`
- `flutter analyze lib test` from `mobile/packages/nostr_client`
- `flutter analyze lib test` from `mobile/packages/nostr_sdk`
- `flutter test test/l10n/arb_consistency_test.dart` from `mobile`
- `flutter analyze lib/notifications test/notifications test/l10n` from `mobile`
- `flutter test test/notifications/widgets/video_notification_row_test.dart` from `mobile`
- `flutter analyze lib test` from `mobile`
- pre-push hook: merge-conflict check, analyzer, and changed-file tests all passed

@rabble ready for re-review.

rabble (approved):
Rabble review: handle addressable list-add notifications — prior 2 blockers fixed (_serverTypesForFilter now returns ['system']/['newPost']), all checks SUCCESS (16/16). LGTM, ready to merge.

rabble:
## Review: `fix/lists: handle addressable list-add notifications` — addf6a68d

Checked the takeover commit against `main` diff (10 files, +314/-32) and against the earlier `a6dde32` that drew CHANGES_REQUESTED. The second push addresses both blocking points from #4878581612.

### Verification done

* `flutter test packages/notification_repository/test/src/notification_repository_test.dart` — **169 tests passed** (including 4 new list_add cases: addressable anchor, foreign-filtered, multi-video group, mixed anchor, block-filter preservation)
* `flutter test packages/models/test/src/video_notification_test.dart` — **18 tests passed** (addedVideoCount field + copyWith)
* `flutter test test/notifications/widgets/video_notification_row_test.dart` — **24 tests passed** (msg uses addedVideoCount, overflow still counts curators)
* `flutter test test/l10n/arb_consistency_test.dart` — **8 tests passed**
* `flutter analyze packages/notification_repository`, `packages/models`, `packages/nostr_client`, `packages/nostr_sdk`, `lib/notifications` — **No issues found** on all

### What was fixed since the first review (verified)

1. **`_serverTypesForFilter` for system/newPost restored.** Current branch returns `const ['system']` / `const ['newPost']` (line 901ff in notification_repository.dart), so selecting those tabs no longer widens to `null` and paginates correctly. The prior page-dilution / docstring divergence is gone.

2. **`totalCount` semantics cleaned up via `addedVideoCount`.** `VideoNotification` now carries `totalCount` = full actor count for all kinds (video_notification.dart:8 doc updated) and `addedVideoCount` = videos added for `listAdd` only (video_notification.dart:85). Row message at video_notification_row.dart:205 uses `addedVideoCount ?? totalCount`; avatar overflow at video_notification_row.dart:172 is `totalCount - actors.length` for every kind — no per-kind subtraction. Merge (notification_repository.dart:1343) and grouping (notification_repository.dart:1654) both set `addedVideoCount = group.length` and `totalCount = distinct pubkeys`, and pagination merge floors at `mergedActors.length`.

3. **Addressable anchor path is sound.** `_videoAnchorEventId` (notification_repository.dart:2334) now falls through `_nonEmpty(referencedEventId) ?? _nonEmpty(rootEventId) ?? _trustedListAddRootAddressableId`. `_trustedListAddRootAddressableId` (notification_repository.dart:2371) parses the coordinate, checks `rootKind` match, `NIP71VideoKinds.isVideoKind`, `pubkey == _userPubkey`, `dTag` non-empty — foreign coordinate test correctly yields `page.items.isEmpty`. `_isEventId` (notification_repository.dart:2383) is the named guard and is applied consistently in `_videoMetadataEventIds` (notification_repository.dart:1554) so `getVideoStats(coordinate)` is never called — the `verifyNever` assertions are the right oracle.

4. **Grouping / routing kept consistent.** `_videoGroupId` (notification_repository.dart:2286) and `_snapshotVideoGroupKey` (notification_repository.dart:1259) both key `listAdd` on `listCoordinate ?? eventId`; `rowEventId` (notification_repository.dart:1648) is `group.first`s anchor (newest after `createdAt` desc sort) so mixed event-id / addressable groups anchor on the newest notification (covered by `list_add mixed anchor` test). `_listAddVideoAddressableId` delegates to the trusted root first, then `_recipientScopedVideoAddressableId`.

5. **Routing comments restored** — notification_tap_target.dart:100 and notifications_view.dart:209 now explain the `list_add` exhaustiveness / fallback, and `_filterItems` generalization (notification_repository.dart:1472) routes all kinds through `_belongsInFeed`.

6. **Compatibility shims are intentional.** `NostrClient.publishEventAwaitOk({diagnosticTag})` (nostr_client.dart:701) and `PublishOutcome.eventKind` (publish_outcome.dart:30) are optional / nullable with defaults — they keep `relay_list_repository.dart:179` compiling on the feature branch.

### Minor nits (non-blocking)

* **`_applyBlockFilter` preserves video count implicitly.** notification_repository.dart:1491 does `n.copyWith(actors: filtered, totalCount: filtered.length)` for VideoNotification. For listAdd that keeps addedVideoCount via copyWith default, so block-filter test expectation holds. Works, but implicit — consider `addedVideoCount: n.addedVideoCount` explicitly with a comment that video count is intentionally not decremented on actor block.

* **Cache hydration still loses the count.** NotificationRow (db_client/lib/src/database/tables.dart) has no column for addedVideoCount; `_itemToCacheRow` stores listCoordinate/listTitle but not the count, and `_rowToPlaceholder` reconstructs with totalCount: 1 and no addedVideoCount. Hydrated placeholders render singular until REST replaces them — acceptable as placeholder, but worth a follow-up column + migration if cold-start copy matters.

* **`_videoMetadataEventIds` now instance method** (was static) because it calls instance `_videoAnchorEventId` — correct, callers only internal.

* **`dTag` for listAdd rows is `group.first.referencedDTag`** (notification_repository.dart:1659). Intentional to stay paired with rowEventId = group.first anchor, rather than prior firstWhere scan. If newest referencedDTag is null while older carries valid tag for same video, route falls back to video?.dTag via recipientScoped helper — still safe.

* **`VideoNotification.copyWith` cannot explicitly clear addedVideoCount** (nullable int + ?? coalescence). Consistent with listTitle/commentText; no current call site needs it.

No further blocking issues found. LGTM.

rabble (approved):
Approve — both blocking issues from rabble review are resolved in addf6a68d, all verification passes (169+18+24 tests, 0 analyzer issues).

Addressable anchor validated with correct trust checks (pubkey+kind+dTag), eventId guard prevents getVideoStats on coordinates, grouping by listCoordinate with newest-anchor selection is covered, addedVideoCount separation restores totalCount invariants and overflow semantics.

PR #6768, opened 2026-08-06
fix(profile): reject empty image-picker files before opening the crop editor

### LINKED ISSUE #6276: fix(profile): reject empty image-picker files before opening the crop editor
## Problem

The profile avatar/banner flow accepts a native image-picker result and opens the crop editor without validating that the selected file exists and contains bytes. When iOS returns a zero-byte temporary JPEG, the crop editor attempts to decode it through both FileImage and MemoryImage, producing paired fatal Crashlytics reports.

The existing cropped-output validation happens too late: the failure occurs while decoding the crop editor's source image.

## Impact and priority

**Priority: P2 / medium**

This issue intentionally groups two Crashlytics signatures because they occur in the same session at the same timestamp and share one missing source-validation boundary.

### Invalid image bytes

Crashlytics issue: 4800cfba8e6f2b43a75afa8de5e20ed1  
[Open in Firebase Crashlytics](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/4800cfba8e6f2b43a75afa8de5e20ed1)

- **5 events / 2 impacted users**
- Signature: instantiateImageCodecWithSize / invalid image data

### Empty image-picker file

Crashlytics issue: 8421aca7ecc9b84b6f8a830aecc50e1f  
[Open in Firebase Crashlytics](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/8421aca7ecc9b84b6f8a830aecc50e1f)

- **5 events / 3 impacted users**
- Signature: FileImage._loadAsync / file is empty and cannot be loaded as an image

Snapshot from 2026-07-22:

- Both signatures first appeared about four days earlier
- Both were observed on 1.0.16
- A paired sample occurred at 2026-07-20T15:00:43Z on an iPhone 17 Pro Max
- Breadcrumbs place the user in edit_profile
- The selected temporary image_picker_*.jpg file had zero bytes

## Reproduction sequence

1. Open Edit Profile.
2. Choose a new avatar or banner.
3. Have the platform image picker return a temporary file that exists but has length zero. A missing file should also be handled.
4. The app immediately opens ImageCropEditorScreen.
5. The editor's image provider tries to decode the invalid source and reports fatal image-decoding errors before the user can crop.

## Confirmed root cause

The profile image selection flow currently passes the picker result directly to the crop editor:

    final picked = await pickProfileXFile(...);
    // Native path:
    file: File(picked.path)

ImageCropEditorScreen then supplies that file to CropRotateEditor.autoSource. The editor reads it through image providers, so a zero-byte source produces both the empty-file and invalid-codec signatures.

The later croppedBytesOrNull check only validates crop output. It cannot prevent the source image from being decoded.

## Expected behavior

- Missing or zero-byte picker output never opens the crop editor.
- Invalid picker output never reaches preview, crop, or upload code.
- The user receives a concise localized selection failure and can try again.
- Valid avatar and banner selections retain their existing crop/upload behavior.
- Expected invalid input is handled locally and is not reported as a fatal error.

## Scope and architecture

Validate at the profile image-selection boundary before constructing or opening the crop editor:

- Native: confirm the file exists and length() is greater than zero.
- Web: confirm the selected byte buffer is non-empty.
- Keep the validation reusable for both avatar and banner paths so one profile surface is not left exposed.
- Use the existing localized profile-image error copy if it accurately describes the failure. If a localization key must change, mirror it across every app_*.arb locale and run the ARB consistency test.
- Preserve current image picker, cropper, and repository layering. Do not add a new service just for a small validation rule.

If validation or file I/O itself fails, map it to the same recoverable user-visible result without leaking a temporary file path into logs or analytics.

## TDD plan

Write failures before the implementation.

### Validation tests

Cover a small reusable validation boundary with:

- missing native file -> invalid;
- existing zero-byte native file -> invalid;
- existing non-empty native file -> valid;
- empty web bytes -> invalid;
- non-empty web bytes -> valid;
- file-stat/read exception -> recoverable invalid result, not an uncaught exception.

Use temporary test files and clean them up.

### Widget/flow tests

Using the existing fake crop-launcher seam, or adding the narrowest picker/validation seam needed:

1. Return an empty picker result.
2. Assert the crop editor is never launched.
3. Assert no avatar/banner upload or state mutation is dispatched.
4. Assert localized failure feedback is visible.
5. Repeat for a missing file.
6. Return a valid non-empty image and assert the existing crop path still launches.
7. Cover both avatar and banner entry points if they share the same unvalidated path.

Do not encode a real user's temporary path into a fixture.

## Likely files

- mobile/lib/screens/profile_setup/widgets/profile_avatar_section.dart
- The corresponding profile banner selection widget/path
- mobile/lib/utils/profile_image_picker.dart or the current shared picker helper
- mobile/lib/screens/profile_setup/image_crop_editor_screen.dart only if needed for defense in depth
- Existing tests such as mobile/test/screens/profile_setup/widgets/profile_avatar_section_test.dart
- Matching banner/picker helper tests
- Localization ARBs and consistency test only if copy changes

## Acceptance criteria

- [ ] Native missing and zero-byte picker files are rejected before crop-editor construction.
- [ ] Empty web picker bytes are rejected before crop-editor construction.
- [ ] Invalid picker results launch neither crop UI nor upload/state mutation.
- [ ] The user gets localized recoverable feedback.
- [ ] Valid avatar and banner selection behavior is unchanged.
- [ ] Regression tests cover invalid and valid cases for the affected paths.
- [ ] Expected empty input is not surfaced as a fatal error.
- [ ] ARB keys remain consistent across locales if user-facing copy changes.
- [ ] Focused tests and flutter analyze pass.
- [ ] The PR links both Crashlytics signatures and this issue.

## Agent workflow

Start from a fresh worktree based on origin/main. Reproduce with tests first, implement validation at the earliest shared boundary, keep temporary paths and user data out of logs, run the focused tests plus ARB consistency if applicable, then commit and open a Conventional Commit-titled PR against main.

hm21:
## Description

The avatar and banner flows passed the image-picker result straight to the crop editor, which decodes its source through `FileImage` / `MemoryImage`. When iOS returned a zero-byte temporary JPEG, that decode threw two paired fatal errors before the user could crop anything. The existing `croppedBytesOrNull` check only validates crop *output*, so it can never prevent a bad *source* from being decoded.

Both Crashlytics signatures are grouped here because they fire in the same session at the same timestamp from one missing source-validation boundary:

- [Invalid image bytes — `instantiateImageCodecWithSize` / invalid image data](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/4800cfba8e6f2b43a75afa8de5e20ed1) (`4800cfba8e6f2b43a75afa8de5e20ed1`)
- [Empty image-picker file — `FileImage._loadAsync` / file is empty](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/8421aca7ecc9b84b6f8a830aecc50e1f) (`8421aca7ecc9b84b6f8a830aecc50e1f`)

**Related Issue:** Closes #6276

### What changed

`resolveProfileImageSelection(XFile)` in `profile_image_picker.dart` is the new shared boundary. It validates the pick and resolves it into a `ProfileImageSelection` — a file on native, eagerly-read bytes on web — or `null` when the picker handed back nothing usable. Both entry points (`ProfileAvatarSection._pickImage` and `_BannerEditButton._pickBannerImage`) now go through it and show the existing localized snackbar on `null`, so neither surface opens the crop editor or dispatches an upload event for an unusable pick.

Design notes for the non-obvious bits:

- **Resolver returns the source, not just a bool.** Both call sites had their own `kIsWeb` branch (read blob bytes vs. wrap a `File`); folding that into the validator removes the duplication and guarantees a surface cannot be validated but then constructed differently. On web the bytes are read once and reused, exactly as before.
- **`lengthSync()` rather than `await file.length()`.** It is one stat syscall on a local temp file right after the user returned from a heavyweight picker, and `avoid_slow_async_io` prefers the sync form. It also keeps the check off the event loop, which matters for widget tests: real async file I/O does not complete inside `pumpAndSettle`'s fake-async zone.
- **A missing file rides the same `PathNotFoundException` path as an empty one.** `lengthSync()` throws for a path that is already gone, and both cases mean the same thing to the user, so they share one rejection and one warning-level log.
- **No new Crashlytics noise.** An empty pick is expected platform input, so it is logged at warning level and never wrapped in `Reportable`. The log message deliberately omits the path — it names a user's temporary file.
- **No ARB change.** `profileSetupImageSelectionFailed` ("Image selection failed. Please paste an image URL below instead.") already exists in all 17 locales and describes this failure accurately, so no copy was touched and no translation debt was added.

The banner path previously failed silently on a bad pick; it now gets the same snackbar as the avatar, from a shared `showProfileImageSelectionFailed(context)` helper.

## Out of Scope

- Defense-in-depth validation inside `ImageCropEditorScreen` itself. The boundary fix removes the reachable path; adding a second check there would duplicate the rule without a caller that can trigger it.
- The banner path still returns silently when `ImagePicker` itself throws (e.g. a denied permission). That is pre-existing behaviour and a different failure class from an empty result.

## Verification

- `flutter test test/screens` — 1726 passed, 62 skipped (skips pre-existing)
- `flutter test test/screens/profile_setup` — 63 passed, including the new cases
- `dart format` — no changes
- All four design-system ceiling ratchets, `check_test_unit_structure.sh`, `check_untested_services_floor.sh` — pass
- Manually verified on a real Samsung Galaxy device: avatar and banner picks from camera and gallery still crop and upload as before

New regression coverage:

- `profile_image_picker_test.dart` — non-empty file resolves; a never-written file resolves to `null` instead of throwing; a zero-byte file resolves to `null`
- `profile_avatar_section_test.dart` and the new `profile_header_preview_test.dart` — zero-byte and deleted picks launch no crop editor, dispatch no upload event, and surface the localized snackbar; a valid pick still crops and dispatches

The canonical test image-picker channel answers with a path that was never written, so the happy-path tests now stub it to a real temp file via the sanctioned `overrideSharedChannel` helper.

The web rule (empty blob bytes rejected) is implemented but not covered by the VM test run, since `kIsWeb` is a compile-time `false` there. I did not add an `isWeb` seam just to reach it.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/screens/profile_setup/widgets/profile_avatar_section.dart
+++ b/mobile/lib/screens/profile_setup/widgets/profile_avatar_section.dart
@@ -1,5 +1,3 @@
-import 'dart:io';
-
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
@@ -201,10 +199,9 @@ class _ProfileAvatarSectionState extends ConsumerState<ProfileAvatarSection> {
 
   /// Platform-aware image selection with an interactive crop step.
   ///
-  /// Native picks yield an [XFile] with a real filesystem path (wrapped in a
-  /// `dart:io File`); web picks yield blob bytes read eagerly before the URL
-  /// is revoked. Either way the image is handed to the Vine crop editor, and
-  /// the resulting cropped JPEG bytes are uploaded via
+  /// [resolveProfileImageSelection] validates the pick and resolves it to a
+  /// file (native) or bytes (web), so an empty pick never reaches the Vine
+  /// crop editor. The resulting cropped JPEG bytes are uploaded via
   /// `BlossomUploadService.uploadImageBytes`. The crop re-encode bounds the
   /// output dimensions and drops the original EXIF.
   Future<void> _pickImage(ImageSource source) async {
@@ -241,26 +238,22 @@ class _ProfileAvatarSectionState extends ConsumerState<ProfileAvatarSection> {
         return;
       }
 
-      final cropLauncher = ref.read(imageCropLauncherProvider);
-      Uint8List? cropped;
-      if (kIsWeb) {
-        // Resolve the blob synchronously here — once we navigate away from
-        // the picker the URL can be revoked.
-        final bytes = await picked.readAsBytes();
-        if (!mounted) return;
-        cropped = await cropLauncher(
-          context,
-          kind: ImageCropKind.avatar,
-          bytes: bytes,
-        );
-      } else {
-        if (!mounted) return;
-        cropped = await cropLauncher(
-          context,
-          kind: ImageCropKind.avatar,
-          file: File(picked.path),
-        );
+      // An empty pick would blow up inside the crop editor's image providers,
+      // so it is rejected here instead.
+      final selection = await resolveProfileImageSelection(picked);
+      if (!mounted) return;
+      if (selection == null) {
+        showProfileImageSelectionFailed(context);
+        return;
       }
+
+      final cropLauncher = ref.read(imageCropLauncherProvider);
+      final cropped = await cropLauncher(
+        context,
+        kind: ImageCropKind.avatar,
+        file: selection.file,
+        bytes: selection.bytes,
+      );
 
       if (cropped == null) {
         Log.info(
--- a/mobile/lib/screens/profile_setup/widgets/profile_header_preview.dart
+++ b/mobile/lib/screens/profile_setup/widgets/profile_header_preview.dart
@@ -1,7 +1,4 @@
-import 'dart:io';
-
 import 'package:divine_ui/divine_ui.dart';
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
@@ -15,6 +12,7 @@ import 'package:openvine/screens/profile_setup/widgets/banner_color_swatches.dar
 import 'package:openvine/screens/profile_setup/widgets/image_url_sheet.dart';
 import 'package:openvine/screens/profile_setup/widgets/profile_avatar_section.dart';
 import 'package:openvine/screens/profile_setup/widgets/profile_image_actions_sheet.dart';
+import 'package:openvine/screens/profile_setup/widgets/profile_image_picker.dart';
 import 'package:openvine/widgets/profile/profile_header_widget.dart';
 import 'package:unified_logger/unified_logger.dart';
 
@@ -229,24 +227,22 @@ class _BannerEditButton extends ConsumerWidget {
     }
     if (picked == null) return;
 
-    final cropLauncher = ref.read(imageCropLauncherProvider);
-    Uint8List? cropped;
-    if (kIsWeb) {
-      final bytes = await picked.readAsBytes();
-      if (!context.mounted) return;
-      cropped = await cropLauncher(
-        context,
-        kind: ImageCropKind.banner,
-        bytes: bytes,
-      );
-    } else {
-      if (!context.mounted) return;
-      cropped = await cropLauncher(
-        context,
-        kind: ImageCropKind.banner,
-        file: File(picked.path),
-      );
+    // An empty pick would blow up inside the crop editor's image providers,
+    // so it is rejected here instead.
+    final selection = await resolveProfileImageSelection(picked);
+    if (!context.mounted) return;
+    if (selection == null) {
+      showProfileImageSelectionFailed(context);
+      return;
     }
+
+    final cropLauncher = ref.read(imageCropLauncherProvider);
+    final cropped = await cropLauncher(
+      context,
+      kind: ImageCropKind.banner,
+      file: selection.file,
+      bytes: selection.bytes,
+    );
 
     if (cropped == null) return;
 
--- a/mobile/lib/screens/profile_setup/widgets/profile_image_picker.dart
+++ b/mobile/lib/screens/profile_setup/widgets/profile_image_picker.dart
@@ -1,6 +1,9 @@
+import 'dart:io';
+
+import 'package:divine_ui/divine_ui.dart';
 import 'package:file_selector/file_selector.dart';
 import 'package:flutter/foundation.dart';
-import 'package:flutter/widgets.dart';
+import 'package:flutter/material.dart';
 import 'package:image_picker/image_picker.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:unified_logger/unified_logger.dart';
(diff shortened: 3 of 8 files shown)


rabble (approved):
## Overview

Adds a validation boundary between the image picker and the crop editor. `resolveProfileImageSelection` checks the pick actually has content and resolves it to a file (native) or bytes (web); an unusable pick now gets a snackbar instead of throwing "file is empty" / "invalid image data" out of the crop editor's image providers. Both the avatar and banner flows go through it, and the duplicated `kIsWeb` branch in each disappears.

The failure it closes is specific and real — the doc names it: *"a zero-byte file (iOS occasionally returns an empty temporary JPEG)"*, plus a picker temp file that's already been reclaimed. Both produce a fatal error at a point where the user has no idea what went wrong and no way to recover.

Things done well:

- **The test helper doesn't break the shared harness.** `stubPickedImageFile` goes through `overrideSharedChannel`, so it auto-restores the canonical handler on teardown and can't strand a later suite in the merged VGV isolate. Equally important, the canonical `test_setup.dart` handler is left *behaving the same* — only its comment is updated to explain that its synthetic path is now deliberately invalid, with a pointer to the opt-in helper. Changing the shared default instead would have been the easy, wrong move.
- **`ProfileImageSelection` is correct by construction.** Two named constructors each nulling the other field, no unnamed constructor, so "exactly one of file/bytes" can't be violated by a caller.
- **The catch keeps the path out of the log** — *"The message stays free of the path: it names a user's temporary file"* — and logs `e.runtimeType` instead. That's the right instinct on a surface that gets shipped to Crashlytics.
- `PathNotFoundException` gets its own arm ahead of the generic catch, so an expected condition doesn't get logged at error level alongside genuine faults.
- Both l10n keys (`profileSetupImageSelectionFailed`, `profileSetupGotItButton`) are present in all 22 locale files, so the ARB consistency gate is satisfied without any ARB churn in this diff.
- `DivineSnackbarContainer.snackBar(..., error: true)` rather than a raw `SnackBar`.

Approving. A few small things.

---

### `onActionPressed: () {}` on the failure snackbar

The snackbar shows a "Got it" action wired to an empty closure. If `DivineSnackbarContainer` dismisses on action-tap by itself, this is just a dismiss button and it's fine — but an empty callback is the shape that also appears when someone meant to do something and didn't, and there's no comment distinguishing the two.

The doc says the snackbar *"points at the URL sheet"*, but nothing in the function opens it — the copy presumably mentions it. If the intent is to give the user a route out of the failure, wiring the action to actually open `ImageUrlSheet` would deliver what the doc promises; if "Got it" is only an acknowledgement, the doc line is slightly over-claiming.

### `lengthSync()` on the platform thread

The comment justifies it well (one syscall, local temp file, keeps the check off the event loop) and for `image_picker` — which copies into app temp before returning — that holds. Worth being aware it stops holding if a future picker path ever hands back a URI backed by a cloud provider placeholder, where a stat can block on a download. Not a change request, just the assumption that would break it.

### `resolveProfileImageSelection` catches everything

The trailing `catch (e)` swallows any exception into a `null` return, which on this path means "unusable pick" — correct for I/O, but it also absorbs programming errors (a `TypeError` from a future refactor) into a user-facing "couldn't use that image". The logging keeps it diagnosable, so this is a nit rather than a problem; narrowing to `on IOException` with the generic catch rethrowing would make an invariant violation visible instead of misreported.

### Nit

Callers pass `file: selection.file, bytes: selection.bytes` — both nullable, with the "exactly one is set" invariant that `ProfileImageSelection` guarantees promptly discarded at the call site. If `imageCropLauncherProvider`'s signature is yours to change, taking the `ProfileImageSelection` directly would carry the invariant through instead of re-deriving it inside the launcher.

Issue #6771, opened 2026-08-06
fix(upload): investigate throughput pinned at 0.03 MB/s (5.9 MB video took 172s)


rabble:
## Summary

A tester reported publishing being "very slow". The session log confirms it: a 5.89 MB video took 172 seconds to upload, and total publish took 184 seconds. Every transfer in the session moved at the same rate, which rules out a stall or a retry loop but does **not** tell us whether the ceiling is the client's uplink or `media.divine.video`. This issue is to get that second data point.

## Measurements

From one publish, app `1.0.19+828`, iPhone 14,2 / iOS 26.5.2, `Network: wifi`, staging:

| Transfer | Size | Duration | Rate |
|---|---|---|---|
| Thumbnail (`PUT /upload`) | 0.22 MB | 7.25s | 0.03 MB/s |
| Video | 5.89 MB | 172.40s | 0.03 MB/s |
| Audio (`audio/wav`) | 0.53 MB | 10.04s | 0.05 MB/s |

```
⏱️ PUBTIME upload.thumbnail.put 7248ms 0.22MB 0.03MB/s
⏱️ PUBTIME upload.transfer 172399ms 5.89MB 0.03MB/s
Upload metrics: 5.9MB in 172s (0.03 MB/s)
⏱️ PUBTIME summary draft=draft_publish_1785996561683936 outcome=success
   total=184372ms upload=172415ms mentions=0ms nostr=11910ms invites=0ms other=46ms
```

0.03 MB/s is roughly 240 kbit/s.

## What the log rules out

- **Not a stall or hang.** All three transfers returned `200` and the rate is flat across them. A stall would show as a long tail on one transfer, not a consistent rate on three.
- **Not retries.** `Recovery sweep: re-drove 0 stuck upload(s), woke 0 retry backoff wait(s), marked 0 failed` — nothing was re-driven.
- **Not auth or signing overhead.** `PUBTIME auth.sign` was 8-9ms per request.
- **Not event publishing.** `nostr=11910ms` out of `total=184372ms`.

## What the log cannot answer

The client only sees its own throughput. A flat 0.03 MB/s is equally consistent with:

- a genuinely slow uplink on the tester's wifi, or
- server-side throttling / a slow write path on `media.divine.video`, or
- something between them (CDN, ISP shaping).

## Suggested next steps

[...]

Issue #6770, opened 2026-08-06
task(audio): decide append vs replace when the music picker adds a second sound


rabble:
## Summary

Picking a second sound from the music library stacks it on top of the first instead of replacing it, so both play in the exported video. A tester hit this, expected a swap, and reported it as a bug. The current behaviour is deliberate, so this needs a product decision before any code changes.

## What happens

`_openMusicLibrary` commits the picked sound through `buildAppendedAudioMeta`, which spreads the existing tracks and appends the new one:

`mobile/lib/screens/video_editor/video_editor_screen.dart:704`
```dart
editor.addHistory(
  meta: buildAppendedAudioMeta(
    activeMeta: editor.stateManager.activeMeta,
    existingTracks: editor.stateManager.audioTracks,
    newTracks: [result],
  ),
);
```

`mobile/lib/extensions/video_editor_extensions.dart:18`
```dart
Map<String, dynamic> buildAppendedAudioMeta({...}) {
  return {
    ...activeMeta,
    VideoEditorConstants.audioStateHistoryKey: [
      ...existingTracks.map((track) => track.toJson()),
      ...newTracks.map((track) => track.toJson()),
    ],
  };
}
```

Confirmed in a field session — the render parameters carry both tracks:

```
meta: {…, audio: [
  {id: bundled_wednesday-1785996456039,      title: Wednesday My Dudes,   endTimeMs: 5992, …},
  {id: bundled_drop_the_bass-1785996493793,  title: D-D-D-Drop The Bass,  endTimeMs: 5992, …}
]}
```

Both windows start at 0 and run the full 5992ms clip, so they play over each other for the entire video.

## Why it is arguably correct

- The timeline genuinely supports multiple audio tracks, with per-track delete, edit, duplicate and split (`video_editor_timeline_overlay_controls.dart:405-458`, `_removeSound` at `:458`).
- `buildAppendedAudioMeta` is shared with the voice-over path (`video_editor_screen.dart:795`), where appending takes is unambiguously right — takes are laid back-to-back and committed as one undo step.
- The only entry point is an action sheet item labelled "Add audio" (`videoEditorAudioAddAudio`), reached via `onOpenMusicLibrary` in `video_editor_main_actions_sheet.dart:108`.

So the feature reads as intentional multi-track. The tester's expectation ("picking a new song swaps the song") is the mismatch.

## The decision

Pick one:

[...]

PR #6772, opened 2026-08-06
fix(sounds): unblock legacy sound reuse and stop the blocked-tap toast queue

### LINKED ISSUE #6769: fix(sounds): every legacy sound is blocked as unremixable, and the error toast queues
## Summary

Every relay-sourced sound in the audio picker is rejected with "This sound can't be remixed right now.", and a burst of blocked taps leaves a backlog of identical toasts that follows the user out of the picker and into the editor.

Reported from a staging session on iPhone 14,2 / iOS 26.5.2, app `1.0.19+828`.

## Symptom 1 — every legacy sound is unremixable

`AudioReuseConsentResolver` is the fallback for audio events that carry no `allow_audio_reuse` tag. It resolves consent by asking **which videos reference the sound**, then filtering those candidates for the one whose `addressableId` matches the sound's own `sourceVideoReference`:

```dart
final ids = await _soundsRepository.fetchVideosUsingSound(sound.id);
if (ids.isEmpty) return false;
final candidates = await _videosRepository.getVideosByIds(ids, hydrateBulkStats: false);
final matching = candidates.where((v) => v.addressableId == sourceAddress) …
```

That reverse lookup queries `Filter(kinds: [34236], e: [audioEventId])` — it only matches videos carrying an `['e', <audioEventId>, <relay>, 'audio']` tag (`video_event_publisher.dart:1586`). **Legacy videos predate that tag**, which is precisely the population the resolver exists to rescue. The lookup returns nothing, `ids.isEmpty` short-circuits, and the sound fails closed.

Five consecutive sounds in the session log, every one the same shape:

```
[SoundsRepository] api: Fetching videos using audio: 608461bd587b9c043188304bc85b4e7a8f87f0091f55851d78aa1d73be64b707
[SoundsRepository] api: Found 0 videos using audio
[AudioSelectionBottomSheet] ui: Sound selection blocked because reuse consent could not be verified: 608461bd587b9c043188304bc85b4e7a8f87f0091f55851d78aa1d73be64b707
```

Blocked ids: `608461bd…b707`, `4b23a280…b6b3`, `2589d976…c9de`, `1391a6f3…139b`, `f330779a…e3bf`. Bundled sounds are unaffected — they short-circuit on `sound.isBundled` before the resolver runs.

The resolver already holds the source video's address on `sound.sourceVideoReference`; there is no reason to reverse-index for it.

## Symptom 2 — the error toast queues

`audio_selection_bottom_sheet.dart:256` calls `showSnackBar` with no `removeCurrentSnackBar`, so every blocked tap enqueues its own toast. The user tapped one blocked sound ten times in three seconds:

```
23:02:39.470  Sound selection blocked … 608461bd…b707
23:02:39.622  Sound selection blocked … 608461bd…b707
23:02:40.779  Sound selection blocked … 608461bd…b707
  … ten total, through 23:02:42.329
```

Ten queued toasts at 2s each is ~20s of backlog. The `ScaffoldMessenger` sits above the picker sheet, so the queue kept draining after the sheet closed. By the time it was still on screen the user had selected a bundled sound (`23:02:50 Sound selected: Wednesday My Dudes (bundled_wednesday)`) — which was never blocked — and reasonably concluded the error was about *that* sound. The original report named the wrong sound because of this.

## Expected

- A legacy sound whose source video grants `allow_audio_reuse` is selectable.
- A blocked tap shows one toast, and it does not outlive the picker by more than its own 2s.
- Fail-closed behaviour is otherwise unchanged: no source video, an address mismatch, a source revision older than the sound, or a failed lookup all still deny.

## Reproduction

1. Open the video editor and add audio.
2. Switch to the Community tab (relay-sourced sounds).
3. Tap any sound whose source video predates the audio `e` tag.
4. Observe the block. Tap repeatedly, then back out to the editor and watch the toast follow you.

## Evidence

Full session log: `openvine_full_logs_2026-08-05T23-17-06.935500.txt` (staging, user pubkey `076c979382b90f5d3a2b21f95e1ee86b6033f14c92e79b7fad3fe1f1073f4886`).

## Affected code

- `mobile/lib/services/audio_reuse_consent_resolver.dart`
- `mobile/lib/widgets/video_editor/audio_editor/audio_selection_bottom_sheet.dart:239-263`
- `mobile/packages/sounds_repository/lib/src/sounds_repository.dart:363` (`fetchVideosUsingSound`)
- `mobile/lib/services/video_event_publisher.dart:1586` (where the `e` tag is written)

rabble:
## Description

Two independent defects behind one field report, both in the audio picker's reuse gate. Delineated as one commit each.

**`fix(sounds): resolve reuse consent from the source video address`**

`AudioReuseConsentResolver` is the fallback for audio events with no `allow_audio_reuse` tag. It resolved consent backwards: ask which videos reference the sound (`fetchVideosUsingSound` → `Filter(kinds: [34236], e: [audioEventId])`), then filter those candidates for the one whose `addressableId` matches the sound's own `sourceVideoReference`. That only matches videos carrying an `['e', <audioEventId>, <relay>, 'audio']` tag, which legacy videos predate — the exact population the resolver exists to rescue. `ids.isEmpty` short-circuited and every legacy sound failed closed.

The source video's address is already on the sound. This reads it directly via `getVideosByAddressableIds` instead.

Because an addressable read resolves to the current revision, the newest-revision-wins sort and the tie-fails-closed branch no longer have anything to disambiguate — the query settles that ordering — so both come out rather than sitting unreachable. Everything else about the policy is untouched: no source video, an address mismatch, a source revision older than the sound, or a thrown lookup all still deny.

Also drops the now-unused `SoundsRepository` dependency, including the instance `video_providers` constructed solely to feed it (plus its `onDispose`).

**`fix(sounds): replace the reuse-blocked toast instead of queueing it`**

Each blocked tap called `showSnackBar` with no `removeCurrentSnackBar`, so the messenger queued them. Ten taps in three seconds — which is what the reporter did — is ~20s of identical toasts draining one after another. The `ScaffoldMessenger` outlives the picker sheet, so the backlog followed the user into the editor and read as an error about whichever sound they had since selected. That is why the original report blamed a bundled sound that was never blocked.

**Related Issue:** Closes #6769

## Out of Scope

- **Signing service returning HTML.** The same session shows C2PA signing failing three times with `Signature: internal error (service responded with an HTTP error (status = 200, content-type = Some("text/html; charset=utf-8")))`. Server-side, tracked separately.
- **Music picker append-vs-replace** (#6770) and **upload throughput** (#6771) from the same report — filed, not addressed here.
- **Stale-relay revision ordering.** `getVideosByAddressableIds` merges across relays with last-write-wins, so a lagging relay could in principle supply a superseded revision. That is shared by every addressable read in the app (repost tabs, lists), not specific to consent, and is not made worse by this change.

## Verification

- `flutter test test/services/audio_reuse_consent_resolver_test.dart` — 9/9
- `flutter test test/widgets/video_editor/audio_editor/` — 21/21
- Consumers of the changed providers: `test/providers/sounds_providers_test.dart`, `test/providers/sound_picker_provider_test.dart`, `test/screens/sound_detail_screen_test.dart`, `test/screens/original_sound_detail_screen_test.dart`, `test/widgets/metadata_sounds_section_test.dart` — 158 + 17 passing
- `dart run build_runner build --delete-conflicting-outputs` — regenerated `sounds_providers.g.dart` / `video_providers.g.dart` (provider hashes only), committed

Both fixes were written test-first; each test was confirmed red against the old behaviour before the fix landed. The consent test fails with `Expected: true / Actual: <false>`; the toast test fails with a queued toast still on screen one lifetime later.

## Type of Change

### DIFF
--- a/mobile/lib/providers/sounds_providers.dart
+++ b/mobile/lib/providers/sounds_providers.dart
@@ -161,7 +161,6 @@ Future<bool> audioReuseTerms(Ref ref, AudioEvent sound) {
   final knownTerms = audioReuseTermsFromEvent(sound);
   if (knownTerms != null) return Future.value(knownTerms);
   return AudioReuseConsentResolver(
-    soundsRepository: ref.watch(soundsRepositoryProvider),
     videosRepository: ref.watch(videosRepositoryProvider),
   ).verify(sound);
 }
@@ -184,7 +183,6 @@ Future<bool> audioReuseConsent(Ref ref, AudioEvent sound) {
   }
   if (knownTerms == false) return Future.value(false);
   return AudioReuseConsentResolver(
-    soundsRepository: ref.watch(soundsRepositoryProvider),
     videosRepository: ref.watch(videosRepositoryProvider),
   ).verify(sound);
 }
--- a/mobile/lib/providers/sounds_providers.g.dart
+++ b/mobile/lib/providers/sounds_providers.g.dart
@@ -651,7 +651,7 @@ final class AudioReuseTermsProvider
   }
 }
 
-String _$audioReuseTermsHash() => r'f8dde7271c3f3bf5f45dba9a5decbafd5785a33d';
+String _$audioReuseTermsHash() => r'29a3be9aebb039c124e6afc99f33c2492a55bc6a';
 
 /// Viewer-independent reuse terms for explicit and legacy audio events.
 
@@ -741,7 +741,7 @@ final class AudioReuseConsentProvider
   }
 }
 
-String _$audioReuseConsentHash() => r'71e983b97064d07c92e7fd97de594d14f4a4fb9b';
+String _$audioReuseConsentHash() => r'd1e80a3056b98eaab653f8708aeac735cd21d4e8';
 
 /// Fail-closed reuse consent for explicit and legacy audio events.
 ///
--- a/mobile/lib/providers/video_providers.dart
+++ b/mobile/lib/providers/video_providers.dart
@@ -47,7 +47,6 @@ import 'package:openvine/services/video_sharing_service.dart';
 import 'package:openvine/services/view_event_publisher.dart';
 import 'package:reposts_repository/reposts_repository.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
-import 'package:sounds_repository/sounds_repository.dart';
 import 'package:unified_logger/unified_logger.dart';
 import 'package:videos_repository/videos_repository.dart';
 
@@ -239,12 +238,9 @@ VideoEventPublisher videoEventPublisher(Ref ref) {
   final profileRepository = ref.watch(profileRepositoryProvider);
   final profileStatsDao = ref.watch(databaseProvider).profileStatsDao;
   final savedSoundsService = ref.watch(savedSoundsServiceProvider);
-  final consentSoundsRepository = SoundsRepository(nostrClient: nostrService);
   final consentResolver = AudioReuseConsentResolver(
-    soundsRepository: consentSoundsRepository,
     videosRepository: ref.watch(videosRepositoryProvider),
   );
-  ref.onDispose(() => unawaited(consentSoundsRepository.dispose()));
 
   // REST-first publish: POST the signed event to {eventPublishBaseUrl}/api/events with
   // NIP-98 auth, falling back to the WebSocket relay pool on transient
--- a/mobile/lib/providers/video_providers.g.dart
+++ b/mobile/lib/providers/video_providers.g.dart
@@ -322,7 +322,7 @@ final class VideoEventPublisherProvider
 }
 
 String _$videoEventPublisherHash() =>
-    r'4d47f884ee710aa88188341a4dfdfd499a14b005';
+    r'47ea7b9e40b7cd95eff972cdb2995d58279e372f';
 
 /// View event publisher for kind 22236 ephemeral analytics events
 ///
--- a/mobile/lib/services/audio_reuse_consent_resolver.dart
+++ b/mobile/lib/services/audio_reuse_consent_resolver.dart
@@ -2,6 +2,7 @@
 // ABOUTME: Fails closed unless the sound's source video grants reuse.
 
 import 'package:models/models.dart';
+import 'package:unified_logger/unified_logger.dart';
 import 'package:videos_repository/videos_repository.dart';
 
 class AudioReuseConsentResolver {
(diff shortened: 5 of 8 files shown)


rabble:
## Review — fix(sounds): unblock legacy sound reuse and stop the blocked-tap toast queue

(Comment rather than a formal review — GitHub won't take one from the author.)

### Overview

Two independent fixes for #6769. The resolver stops asking "which videos reference this sound" (a reverse lookup through an `['e', …, 'audio']` tag that legacy videos predate, so it returned nothing for exactly the population the resolver exists to rescue) and instead reads the source video straight off the address the sound already carries. And the blocked-tap path replaces the current snackbar instead of enqueuing, so a burst of taps no longer trails 20s of identical toasts into the editor.

The direction inversion is clearly right and the diagnosis in the comment is the good kind — it says why the old shape couldn't work rather than just what changed. The snackbar fix and its test are clean; `removeCurrentSnackBar()` then `showSnackBar` is the correct idiom and the test proves the whole burst clears on one lifetime rather than the first of three.

Two concerns on the resolver.

---

### 1. Dropping the newest-revision sort re-opens the revocation hole

Old code:

```dart
final matching = candidates
    .where((video) => video.addressableId == sourceAddress)
    .where((video) => video.createdAt >= sound.createdAt)
    .toList()
  ..sort((left, right) => right.createdAt.compareTo(left.createdAt));
if (matching.length > 1 && matching[0].createdAt == matching[1].createdAt) return false;
return matching.first.allowAudioReuse;
```

New code keeps `matching.first` but drops both the sort and the tie-break, justified by:

> an addressable read resolves to the current revision, so this is the live answer

That isn't what `getVideosByAddressableIds` guarantees. It builds one `Filter(kinds:, authors:, d:)` per address, batches them, and returns `results.expand((e) => e).toList()` — no per-address newest-wins reduction anywhere. The "only one latest version per ID" note inside that method is a rationale for *not setting `limit: 1` on the filter*, and it's about what a relay serves; it says nothing about the merged result. `queryEvents` folds in cache hits, and the cache demonstrably holds superseded revisions of the same coordinate — that's the whole subject of #6711.

So `candidates` can carry several revisions of one address, and `.first` is now whichever the query happened to order first rather than the newest. Concretely: a creator publishes a video with `allow_audio_reuse`, later edits it to drop the tag (which is how revocation works, per the comment two lines below), and if the permissive older revision comes back first, this grants reuse. That is precisely the failure the sort and the equal-timestamp fail-closed existed to prevent, and neither is replaced by anything.

Restoring both is cheap and keeps the fix's actual change (the lookup direction) intact:

```dart
final matching = candidates
    .where((video) => video.addressableId == sourceAddress)
    .toList()
  ..sort((left, right) => right.createdAt.compareTo(left.createdAt));
if (matching.isEmpty) return false;
if (matching.length > 1 && matching[0].createdAt == matching[1].createdAt) return false;
final source = matching.first;
```

The current tests can't catch this — every `stubSource` call returns a single video, so `.first` is unambiguous in all of them. Worth a case that stubs two revisions of the same address, permissive-older listed first, and asserts `isFalse`.

### 2. Does the timestamp rule actually admit the population this is meant to rescue?

`if (source.createdAt < sound.createdAt) return false;` carries over the old `video.createdAt >= sound.createdAt` filter unchanged, so this isn't a regression — but the fix's premise makes it worth re-examining. A legacy kind-1063 sound was published *when the audio was extracted from the video*, so for a video that was never edited afterwards, `video.createdAt < sound.createdAt` holds and the resolver still fails closed. If that describes most of #6769's affected sounds, the lookup-direction fix lands and the reuse stays blocked anyway.

Under the old reverse lookup this was invisible — the lookup returned nothing, so the timestamp check never ran and never got tested against real data. Now it's the only gate left. Before merging, worth pulling a handful of actual affected `(sound, source video)` pairs off the relay and checking which side of that inequality they fall on. If the answer is "the source video is usually older", the rule needs to become "the source video's *current* revision, whenever it was published" rather than a timestamp comparison — the addressable read already gives you currency, which is the property that actually matters for revocation.

---

### Nits

- `video_providers.dart` drops `consentSoundsRepository` and its `ref.onDispose(() => unawaited(...))`. Worth confirming `unawaited` (and therefore `dart:async`) is still used elsewhere in that file — analyze is green, so presumably yes, just flagging since it was the only visible use in the diff.
- The resolver's `catch (_) { return false; }` swallows everything, including programming errors. Fail-closed is right for the consent answer, but a `Log.warning` there would make "why is reuse blocked for this sound" answerable from a bug report instead of requiring a repro.
- ABOUTME line 2 now reads "Fails closed unless the sound's source video grants reuse" — accurate, and better than the old one. If finding 1 is addressed, "the source video's current revision" would be sharper still.

PR #5976, opened 2026-07-09
feat(proofmode): carry C2PA manifest forward on watermark download

hm21:
## Description

The watermark burn-in in `WatermarkDownloadService` re-encodes the video via `pro_video_editor`, producing a fresh MP4 that drops the embedded C2PA manifest. This adds `C2paSigningService.resignDerived`, which carries an existing manifest forward onto a re-encoded ("derived") file:

- **Gates on the source already carrying a manifest** (`readManifest`) — a third-party download with no provenance is never given fabricated provenance; it's left untouched.
- Attaches the source as a `parentOf` **ingredient** and records a `c2pa.*` **edit action** via the `ManifestBuilder` API (rather than minting a fresh `digitalCapture` claim, which would misrepresent an edited file as a capture).
- Re-signs and embeds the manifest back into the derived file in place.

Wired into the watermark-download path only — the other #3491 drop points are already covered by existing signing (see Out of Scope).

**Related Issue:** Closes #3491

## ⚠️ Reviewers: this needs the proofsign signing key to verify end-to-end — and that verification is what closes the issue

We do **not** have the proofsign signing credentials, so this could not be verified on a real device here. Signing is gated on `PROOFMODE_SIGNING_SERVER_ENDPOINT` / `PROOFMODE_SIGNING_SERVER_TOKEN` (`String.fromEnvironment`), which are absent from a default dev build — without them `signVideo`/`resignDerived` best-effort no-op and return the unsigned file.

Because the remaining acceptance criteria are verification (not more code), a reviewer who holds the key (cc @n8fr8) needs to run the checks below. Seeing a valid manifest in a saved file validates the **whole chain at once**: the source blob preserved its manifest, the carry-forward embedded it, and the OS Photos/MediaStore import (`Gal.putVideo`) did **not** strip the JUMBF box (#3491 AC #4).

Build a **signing build** first (`PROOFMODE_SIGNING_SERVER_*` dart-defines set, e.g. `run_dev.sh` with the env vars exported), record a clip, then on **both iOS and Android**:

1. **Plain save** (AC #1/#2) — save a recorded/rendered clip to the gallery, pull it, and confirm `c2patool` reports a valid Divine manifest.
2. **Watermark download** (AC #3, this PR) — watermark-download an already-signed clip, pull it, and confirm `c2patool` reports a valid manifest with the `parentOf` ingredient + edit action.

[...]

PR #6773, opened 2026-08-06
fix(c2pa): validate signing endpoint and stop blaming the user's connection

rabble:
## Summary

A C2PA signing failure presented as a server outage but was a client-side
URL-construction bug. This makes a missing or malformed signing endpoint fail
loudly at build time and launch, and stops the failure UI from blaming the
user's connection.

## Motivation

`_createSigner()` builds its URL as `SIGNING_SERVER_ENDPOINT + "?platform=ios"`.
`PROOFMODE_SIGNING_SERVER_ENDPOINT` had no default, so a missing value — or one
that stopped at the host — collapsed the URL to the service root. A service root
answers HTTP 200 with an HTML landing page, and the signer surfaced that as:

```
C2PA API error: Signature: internal error
(service responded with an HTTP error (status = 200,
 content-type = Some("text/html; charset=utf-8")))
```

Nothing about that says "misconfigured build". The user was on working wifi, and
the in-app note told them content credentials need an internet connection, so
they went and debugged their network.

### The default alone would not have fixed CI

`codemagic.yaml` interpolates the variables **unconditionally**:

```
--dart-define=PROOFMODE_SIGNING_SERVER_ENDPOINT=$PROOFMODE_SIGNING_SERVER_ENDPOINT
--dart-define=PROOFMODE_SIGNING_SERVER_TOKEN=$PROOFMODE_SIGNING_SERVER_TOKEN
```

If a Codemagic variable is unset, that emits a define that is *present but
empty*, which overrides Dart's `defaultValue` rather than falling back to it.
So a build-time guard is required, not just a default. Required C2PA builds now
require both a valid endpoint and a non-empty token unless the endpoint is
explicitly `disabled`.

## Changes

- Default the endpoint to `https://proofsign.divine.video/api/v1/c2pa/configuration`.
- Validate at launch that the endpoint is an absolute `https` URL ending in the
  configuration path, with no query string. A malformed value now fails the
  launch instead of surfacing mid-capture.
- Add `mobile/scripts/check_signing_endpoint.sh`, wired into `build_ios.sh`,
  `build_android.sh`, `run_dev.sh`, and into `codemagic.yaml` with
  `REQUIRE_SIGNING_ENDPOINT=1` so unset required CI credentials fail the build.
- `PROOFMODE_SIGNING_SERVER_ENDPOINT=disabled` turns signing off, preserving the
  #6058 gate for CI and unconfigured builds now that the endpoint is never empty.
  `signVideo` returns a typed `disabled` result before attempting any network work.
- Show a service-failure note when the device has a network; keep the
  connectivity wording only for genuinely offline devices. The signal is
  `connectivity_plus`, via a `deviceIsOfflineProvider` seam over the existing
  `dmSendConnectivityIsOffline` predicate.
- Require a bearer token for `isSigningConfigured`, and pass
  `PROOFMODE_SIGNING_SERVER_ENDPOINT=disabled` in both web workflows.
- Make the signing endpoint and token injectable on `C2paSigningService` so the
  configured / opted-out branches are reachable from tests at all.
- Keep intentionally disabled signing builds out of warning logs, and align
  `mobile/.env.example` with the shipped `proofsign.divine.video` endpoint.

## Behaviour changes worth a close look

1. **`isSigningConfigured` means endpoint *plus* token.** Defaulting the endpoint
   decoupled the two: previously an unset-env build was unconfigured because both
   defines were empty, and after the default it would have counted as configured
   and called production `proofsign.divine.video` with `bearerToken: ''`. Requiring
   the token restores the coupling for local `flutter run` and for any workflow
   that passes neither define. Both web workflows now set `disabled` explicitly,
   so the opt-out has real callers rather than being a sentinel nothing sets.
2. **Malformed endpoint remains a launch-fail.** That is intentional option 1
   from review: a build with a malformed signing endpoint should not continue as
   if it can produce content credentials. The check now runs after crash reporting
   initialization, so the startup failure is reportable instead of being dropped
   before telemetry is ready.
3. **New string is English-only**, registered in `_knownUntranslatedDebt` and
   tracked by #6808. Offline devices still get the fully translated existing
   note — and that note is now actually reachable, which it was not in the first
   revision of this PR.

## Review takeover

@mbradley and @NotThatKindOfDrLiz pushed the mechanical fixes directly
(`942a341`, `0547da9`, `daf8eb2`, `b3669a6`, `21f5d481f`): the missing
`resignDerived` disabled-sentinel guard, the stale `isSigningConfigured` doc,
fragment-vs-absolute diagnostic ordering, guard-script header claims, local
guards running after `.env` is sourced, one shared exact-path rule across the
shell guard and the Dart validator, startup validation moved below crash-reporting
init, and the backend-host-default guard widened to cover the C2PA endpoint.

Two owner decisions were left open and are resolved here:

**The connectivity signal.** `ConnectionStatusService` is not a connectivity
check — `_isConnected` initialises `true`, its only mutator `updateRelayStatus()`
is never called anywhere in the repo, and `checkConnection()` is a stub that sets
no state. So `isOnline` was a constant `true`: the sheet always claimed the
service was unavailable, `videoMetadataC2paMissingNote` was unreachable in all 21
locales, and a genuinely offline user was told "this isn't your connection". It
also constructed the real service's 30s `Timer.periodic` inside the widget tree,
which is what left four tests failing with *"A Timer is still pending even after
the widget tree was disposed"*. Replaced with `connectivity_plus`.

**The token policy.** `isSigningConfigured` requires a token, the web workflows
opt out explicitly, and the required Codemagic guard now fails if a non-disabled
endpoint is present without a token.

## Test plan

- `flutter test test/services/c2pa_signing_endpoint_validation_test.dart` — covers
  the bare-host, empty, relative, http, query-string, guard/validator drift, and
  required-token cases.
- `flutter test test/screens/video_metadata_screen_test.dart` — 16 pass,
  including two new cases pinning that the sheet blames the connection only when
  the device is offline and the service otherwise. The four pending-`Timer`
  failures reported in review are gone; they were caused by the real
  `ConnectionStatusService` being constructed in the widget tree.
- `flutter test test/services/c2pa_signing_service_test.dart` — 12 pass,
  including two new cases pinning that an untokened build and a `disabled`
  endpoint both return `disabled` without reaching `signFile`.
- `dart test test/l10n/arb_consistency_test.dart` — passes.
- `flutter analyze` on changed files — no issues.
- Guard script exercised directly: unset → exit 0; bare host → exit 1 with the
  remediation message; valid URL → exit 0; `disabled` → exit 0; with
  `REQUIRE_SIGNING_ENDPOINT=1`, unset endpoint → exit 1, valid endpoint without
  token → exit 1, and valid endpoint with token → exit 0.

## Still open from review

The guard script's suffix match vs the Dart validator's exact-path compare was
reconciled by `21f5d481f` (both use the exact rule). The remaining question —
whether a subpath / reverse-proxied deployment should be legal at all — is a
deployment decision, not a code one, and the two sides now agree either way.

## Not covered here

The Codemagic `proofmode_credentials` values themselves are not readable from a
checkout, so which endpoint/token shipped builds actually use still needs to be
confirmed in the Codemagic UI. Related server-side hardening (returning JSON
instead of HTML to API clients) is a separate PR in `proofsign-server`.

No visual change beyond the one note string on the missing-content-credential
sheet.

Relates to #838

### DIFF
--- a/mobile/lib/screens/video_metadata/video_metadata_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_screen.dart
@@ -8,6 +8,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/models/video_recorder/video_recorder_mode.dart';
+import 'package:openvine/providers/relay_providers.dart';
 import 'package:openvine/providers/shared_preferences_provider.dart';
 import 'package:openvine/providers/video_editor_provider.dart';
 import 'package:openvine/providers/video_publish_provider.dart';
@@ -99,6 +100,13 @@ class _VideoMetadataScreenState extends ConsumerState<VideoMetadataScreen> {
 
   Future<void> _showC2paMissingPrompt() async {
     final l10n = context.l10n;
+    // Blaming connectivity for a service-side failure sends users to debug wifi
+    // that is working. Only claim the connection when the device is actually
+    // offline; otherwise say the service didn't respond.
+    final isOnline = ref.read(connectionStatusServiceProvider).isOnline;
+    final note = isOnline
+        ? l10n.videoMetadataC2paMissingNoteServiceUnavailable
+        : l10n.videoMetadataC2paMissingNote;
     // Non-dismissible: forfeiting the content credential is a provenance
     // decision, so require an explicit button rather than letting an accidental
     // barrier tap / swipe silently post without it (#6058).
@@ -107,7 +115,7 @@ class _VideoMetadataScreenState extends ConsumerState<VideoMetadataScreen> {
       sticker: .alert,
       title: l10n.videoMetadataC2paMissingTitle,
       subtitle: l10n.videoMetadataC2paMissingBody,
-      additionalText: l10n.videoMetadataC2paMissingNote,
+      additionalText: note,
       primaryButtonText: l10n.videoMetadataC2paMissingRegenerate,
       onPrimaryPressed: () =>
           Navigator.of(context).pop(_C2paMissingChoice.regenerate),
--- a/mobile/lib/services/c2pa_signing_service.dart
+++ b/mobile/lib/services/c2pa_signing_service.dart
@@ -98,12 +98,13 @@ class C2paSigningService {
 
   /// Whether the remote signing pipeline is configured for this build.
   ///
-  /// Signing is a network call to [SIGNING_SERVER_ENDPOINT], which is injected
-  /// as a `--dart-define` only in builds that opt in (see `build_*.sh` /
-  /// `run_dev.sh`). CI and unconfigured builds leave it empty, so a missing
-  /// C2PA signature there is expected and must never be surfaced to the user —
+  /// Signing is a network call to [SIGNING_SERVER_ENDPOINT]. The endpoint always
+  /// resolves to a usable URL — an override that is absent falls back to
+  /// [defaultSigningServerEndpoint] — so signing is configured unless a build
+  /// explicitly disables it with `PROOFMODE_SIGNING_SERVER_ENDPOINT=disabled`.
+  /// Builds that disable it never surface a missing C2PA signature to the user;
   /// callers gate the "sign or skip" prompt on this (#6058).
-  static bool get isSigningConfigured => SIGNING_SERVER_ENDPOINT.isNotEmpty;
+  static bool get isSigningConfigured => !_signingDisabled;
 
   /// Signs a video file with C2PA content credentials.
   ///
@@ -490,10 +491,81 @@ class C2paSigningService {
     );
   }
 
+  /// Path every signing endpoint must end in.
+  ///
+  /// [_createSigner] appends `?platform=...` to the endpoint, so a value that
+  /// stops at the host collapses to the service root. The root serves an HTML
+  /// landing page with HTTP 200, and the signer then reports a bare
+  /// "Signature: internal error" that reads like a server outage.
+  static const String signingConfigurationPath = '/api/v1/c2pa/configuration';
+
+  /// Endpoint used when a build does not override it.
+  static const String defaultSigningServerEndpoint =
+      'https://proofsign.divine.video$signingConfigurationPath';
+
+  /// Sentinel that turns signing off entirely, for CI and unconfigured builds.
+  static const String signingDisabledSentinel = 'disabled';
+
   // add ?platform=android or ios
   static const String SIGNING_SERVER_ENDPOINT = String.fromEnvironment(
     'PROOFMODE_SIGNING_SERVER_ENDPOINT',
+    defaultValue: defaultSigningServerEndpoint,
   );
+
+  static bool get _signingDisabled =>
+      SIGNING_SERVER_ENDPOINT.trim().toLowerCase() == signingDisabledSentinel;
+
+  /// Validates [SIGNING_SERVER_ENDPOINT], returning null when it is usable and
+  /// a human-readable reason when it is not.
+  ///
+  /// Exposed separately from [assertSigningEndpointValid] so tests can check
+  /// arbitrary values without tearing down the app.
+  static String? describeSigningEndpointProblem(String endpoint) {
+    if (endpoint.trim().toLowerCase() == signingDisabledSentinel) return null;
+
+    if (endpoint.isEmpty) {
+      return 'PROOFMODE_SIGNING_SERVER_ENDPOINT is empty. Pass '
+          '--dart-define=PROOFMODE_SIGNING_SERVER_ENDPOINT=<url> or leave it '
+          'unset to use $defaultSigningServerEndpoint.';
+    }
+
+    final uri = Uri.tryParse(endpoint);
+    if (uri == null || !uri.isAbsolute || uri.host.isEmpty) {
+      return 'PROOFMODE_SIGNING_SERVER_ENDPOINT ("$endpoint") is not an '
+          'absolute URL.';
+    }
+
+    if (uri.scheme != 'https') {
+      return 'PROOFMODE_SIGNING_SERVER_ENDPOINT ("$endpoint") must use https, '
+          'not "${uri.scheme}".';
+    }
+
+    if (uri.hasQuery) {
+      return 'PROOFMODE_SIGNING_SERVER_ENDPOINT ("$endpoint") must not carry a '
+          'query string; the platform parameter is appended at call time.';
+    }
+
+    if (uri.path != signingConfigurationPath) {
+      return 'PROOFMODE_SIGNING_SERVER_ENDPOINT ("$endpoint") must end in '
+          '$signingConfigurationPath. A bare host resolves to the service '
+          'root, which returns an HTML landing page instead of a signing '
+          'configuration.';
+    }
+
+    return null;
+  }
+
+  /// Fails the launch when the signing endpoint is malformed.
+  ///
+  /// A misconfigured endpoint does not fail loudly on its own: it surfaces much
+  /// later as an opaque signing error during capture, which reads as an outage
+  /// rather than a build problem. Called from app startup.
+  static void assertSigningEndpointValid() {
+    final problem = describeSigningEndpointProblem(SIGNING_SERVER_ENDPOINT);
+    if (problem != null) {
+      throw StateError('C2PA signing misconfigured: $problem');
+    }
+  }
 
   static const String SIGNING_SERVER_TOKEN = String.fromEnvironment(
     'PROOFMODE_SIGNING_SERVER_TOKEN',
--- a/mobile/lib/services/c2pa_signing_service.dart
+++ b/mobile/lib/services/c2pa_signing_service.dart
@@ -32,6 +32,9 @@ enum C2paSigningFailureReason {
   outputMissing,
   tls,
   network,
+
+  /// This build opted out of C2PA signing; nothing was attempted.
+  disabled,
   other,
 }
 
@@ -123,6 +126,17 @@ class C2paSigningService {
     bool enableAdvancedCawgEmbedding = false,
   }) async {
     try {
+      // Opted-out builds must not reach the signer: the endpoint is a sentinel,
+      // not a URL, so attempting the call would fail with a confusing error.
+      if (!isSigningConfigured) {
+        return C2paSigningResult(
+          signedFilePath: videoPath,
+          success: false,
+          error: 'C2PA signing is disabled for this build',
+          failureReason: C2paSigningFailureReason.disabled,
+        );
+      }
+
       Log.info(
         'Starting C2PA signing for video: $videoPath',
         name: 'C2paSigningService',
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -38,6 +38,13 @@ definitions:
     - &check_ios_extension_signing_contract
       name: Check iOS extension signing contract
       script: ./scripts/check_ios_extension_signing_contract.sh
+    - &check_signing_endpoint
+      name: Check C2PA signing endpoint
+      script: |
+        # Every build step below interpolates this variable unconditionally, so
+        # an unset variable ships an empty --dart-define that overrides the
+        # in-app default and breaks signing on every capture. Require it here.
+        REQUIRE_SIGNING_ENDPOINT=1 ./scripts/check_signing_endpoint.sh
     - &clean_ios_workspace_state
       name: Clean stale Xcode/SPM workspace state
       script: |
@@ -482,6 +489,7 @@ workflows:
     triggering:
       events: []
     scripts:
+      - *check_signing_endpoint
       - *check_ios_extension_signing_contract
       - *setup_code_signing_xcode
       - *clean_ios_workspace_state
@@ -536,6 +544,7 @@ workflows:
     triggering:
       events: []
     scripts:
+      - *check_signing_endpoint
       - *clean_ios_workspace_state
       - *enable_spm
       - *flutter_pub_get
@@ -586,6 +595,7 @@ workflows:
     triggering:
       events: []
     scripts:
+      - *check_signing_endpoint
       - *enable_spm
       - *setup_local_properties
       - *configure_gradle_memory
@@ -644,6 +654,7 @@ workflows:
     triggering:
       events: []
     scripts:
+      - *check_signing_endpoint
       - *install_flutterfire
       - *enable_spm
       - *flutter_pub_get
@@ -677,6 +688,7 @@ workflows:
     triggering:
       events: []
     scripts:
+      - *check_signing_endpoint
       - *clean_ios_workspace_state
       - *enable_spm
       - *flutter_pub_get
@@ -712,6 +724,7 @@ workflows:
     triggering:
       events: []
     scripts:
+      - *check_signing_endpoint
       - *flutter_pub_get
       - *setup_local_properties
       - *build_android_e2e
--- a/mobile/build_android.sh
+++ b/mobile/build_android.sh
@@ -4,6 +4,11 @@
 
 set -e
 
+# Fail fast on a malformed C2PA signing endpoint. A bare host silently collapses
+# to the service root at call time and surfaces as an opaque signing error
+# during capture rather than a build failure.
+"$(dirname "${BASH_SOURCE[0]:-$0}")/scripts/check_signing_endpoint.sh"
+
 # Colors for output
 RED='\033[0;31m'
 GREEN='\033[0;32m'
--- a/mobile/build_ios.sh
+++ b/mobile/build_ios.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -euo pipefail
 
+# Fail fast on a malformed C2PA signing endpoint. A bare host silently collapses
+# to the service root at call time and surfaces as an opaque signing error
+# during capture rather than a build failure.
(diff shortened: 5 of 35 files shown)


mbradley (changes requested):
I took this over and pushed four commits for the clear findings. Two blockers are left, and both are yours: they need a decision I shouldn't make for you.

Verified against `fe21e57`, re-verified at `b3669a6d9`.

## The build-guard half checks out

I confirmed both premises independently rather than taking them on faith. `codemagic.yaml` does interpolate the variable unconditionally at all seven build sites, and an empty `--define=X=` really does yield `''` and bypass `String.fromEnvironment`'s `defaultValue` (`value=[] isEmpty=true`, Dart 3.12.2). `REQUIRE_SIGNING_ENDPOINT=1` is the right mechanism, and all six workflows carry it.

## Blocker 1: CI is red on this branch, in this PR's own code

Four tests in the `C2PA signing prompt (#6058)` group of `test/screens/video_metadata_screen_test.dart` fail with "A Timer is still pending even after the widget tree was disposed."

Root cause: the new `ref.read(connectionStatusServiceProvider)` at `video_metadata_screen.dart:106` constructs the real `ConnectionStatusService`, whose constructor starts a 30s `Timer.periodic` that outlives the widget tree.

These were failing before my commits and still are. My four commits added no new failures: shards 0, 1 and 3, Generated Files, Analyze and Format are all green at `b3669a6d9`, and shard 2 fails with exactly the same four tests as at `fe21e57`.

## Blocker 2: `isOnline` is a constant `true`, so the branch this PR adds never takes its other path

`ConnectionStatusService` is not a connectivity check:

- `_isConnected` is initialized `true` (`connection_status_service.dart:19`).
- Its only mutator, `updateRelayStatus()` (`:58`), is never called anywhere in the repo.
- `checkConnection()` (`:117`) is a stub that awaits 100ms and sets no connection state. Its own comment reads "Simulate connection check - in real implementation, this would ping relays or check network connectivity."

So `isOnline` returns `true` unconditionally, for the process lifetime, in every build. Three consequences:

1. `videoMetadataC2paMissingNoteServiceUnavailable` shows to everyone, always.
2. `videoMetadataC2paMissingNote` becomes unreachable, so the 21 non-English locales lose translated copy on this sheet rather than keeping it for offline users.
3. A genuinely offline user is told "This isn't your connection." The old copy was hedged; this one is definitive and wrong for exactly the users it was meant to stop misdirecting.

Which signal do you want?

- **(a)** Plumb `C2paSigningFailureReason` through. It already distinguishes `network` / `tls` / `other` at `c2pa_signing_service.dart:391-408` but is only logged and never reaches the UI. It describes the failure that actually happened rather than guessing globally, and it fixes the test failure without a mock because nothing new gets constructed.
- **(b)** Use `connectivity_plus`, already used at `relay_providers.dart:696` and in `upload_manager.dart`.
- **(c)** Drop the branch and ship one string with copy that doesn't claim to know the cause.

## What I pushed

One commit per finding, no force-push, no rebase.

| Commit | Finding |
|---|---|
| `942a341` | `resignDerived` never got the disabled-sentinel guard that `signVideo` got. On the watermark-download path (#5976) a `disabled` build still reached `_createSigner()` and built `RemoteSigner(configurationUrl: 'disabled?platform=ios')`. The sole caller discards the result, so this only skips a call that could never have succeeded. |
| `0547da9` | The `isSigningConfigured` doc comment claimed the endpoint "always resolves to a usable URL", which is the case your own description disproves. An empty define stays empty and the getter returns `true` for it. |
| `daf8eb2` | `Uri.isAbsolute` is false whenever a fragment is present, so `...configuration#frag` was reported as "is not an absolute URL". Now reported as a fragment. Accept/reject set unchanged, only the diagnostic differs. Covered by a regression test that fails against the previous ordering. |
| `b3669a6` | The guard script header claimed it runs as a pre-push hook (it is not installed as one) and that it uses `grep` and only POSIX constructs (it uses neither `grep` nor POSIX-only syntax; `set -o pipefail` is not POSIX). |

Validation: `dart format` clean, `flutter analyze` clean on both changed Dart files, 28 tests green locally across the two signing suites and `arb_consistency_test.dart`, and the guard script's accept/reject behaviour confirmed unchanged.

## Four questions I did not implement

These need a call rather than a patch.

**1. The guard and the Dart validator enforce different rules, and the gap ships a green build that dies at launch.** The script suffix-matches `REQUIRED_PATH`; the Dart side compares `uri.path` exactly. Verified:

```
$ PROOFMODE_SIGNING_SERVER_ENDPOINT=https://proofsign.divine.video/proxy/api/v1/c2pa/configuration \
    bash mobile/scripts/check_signing_endpoint.sh
PROOFMODE_SIGNING_SERVER_ENDPOINT looks well-formed.   # exit 0
```

That value fails `describeSigningEndpointProblem`, so it passes every build gate, ships, and throws in `assertSigningEndpointValid()` before `runApp`. A reverse-proxied or path-mounted deployment is the ordinary way to hit this, and the script's own error text ("It must end in ...") invites it. Two smaller divergences the other way: `" disabled "` with whitespace passes Dart's `.trim()` but fails the script with a message about https, and `HTTPS://` uppercase fails the script but passes Dart.

Is a subpath legal? Whichever way you go, both sides need the same rule plus a test feeding one shared input table through both. I recorded the divergence in the script header in `b3669a6` but a comment will not catch the next drift.

**2. The launch failure is silent and unrecoverable.** `assertSigningEndpointValid()` is at `main.dart:1004`; `CrashReportingService.initialize()` at `:1060`; `recordError` opens with `if (!_initialized) return;`; `runApp` at `:1553`. `runZonedGuarded` catches the `StateError` and hands it to a recorder that has not initialized yet and discards it. A build that trips this hangs on the launch image forever with no telemetry, surfacing only as a crash-free-rate cliff with no crash attached. Options: move the assert below crash-reporting init, render an error screen instead of hanging, or log loudly and disable signing for the session. Given the build guard now catches malformed values in CI, the third costs little and removes the brick risk.

**3. The endpoint got a default; the token did not.** `SIGNING_SERVER_TOKEN` has no `defaultValue` and `isSigningConfigured` no longer considers it. Previously the two were coupled by both being empty. So an unset-env build (local `flutter run`, plus both web workflows, neither of which passes the define) now counts as configured, calls production `proofsign.divine.video` with `bearerToken: ''`, and shows the blocking sheet when that is rejected. Should `isSigningConfigured` also require a token, or should the web workflows pass `disabled`? Worth noting nothing in the repo sets `disabled` today, so the opt-out currently has no callers.

**4. The new host default escapes an existing guard.** `defaultSigningServerEndpoint` (`c2pa_signing_service.dart:517`) is a backend-host default in a service file, so it falls outside `check_backend_host_defaults.sh`, which scans only `app_config.dart`. Every other `String.fromEnvironment` backend default lives there. Move it, or widen the guard?

## Two smaller notes

**Coverage.** The three genuinely new runtime behaviours have no tests: the `disabled` early-return, the flipped `isSigningConfigured`, and `assertSigningEndpointValid()` throwing. The ten validator tests are sound (I mutation-checked them; the two that should catch the reported outage do go red), but they only exercise the pure string function. All three untested behaviours read compile-time `String.fromEnvironment` state, so they are untestable in-process by construction. That is also why I added no test for the `resignDerived` guard in `942a341`. Worth making the endpoint injectable so the behaviour that changed can be tested at all.

**PR body.** No linked issue; `pr_issue_link.yml` enforces it and suggests #838. The test plan says "118 pre-existing tests pass, no regressions", which is not currently true of `test/screens/video_metadata_screen_test.dart`.

Please review the four commits, and let me know how you want to resolve blocker 2 and the four questions. Happy to implement whichever direction you pick once it is decided.

mobile/lib/screens/video_metadata/video_metadata_screen.dart
```diff
@@ -99,6 +100,13 @@ class _VideoMetadataScreenState extends ConsumerState<VideoMetadataScreen> {
 
   Future<void> _showC2paMissingPrompt() async {
     final l10n = context.l10n;
+    // Blaming connectivity for a service-side failure sends users to debug wifi
+    // that is working. Only claim the connection when the device is actually
+    // offline; otherwise say the service didn't respond.
+    final isOnline = ref.read(connectionStatusServiceProvider).isOnline;
```
NotThatKindOfDrLiz, line 106: This still needs an owner decision. `ConnectionStatusService.isOnline` starts true and is not a real device connectivity signal here, so this branch always selects the service-unavailable copy and constructs the real service timer in widget tests. Please choose the intended signal: plumb `C2paSigningFailureReason` through to this sheet, use a real connectivity source, or use neutral copy that does not claim the cause.

mobile/lib/services/c2pa_signing_service.dart
```diff
@@ -98,12 +101,15 @@ class C2paSigningService {
...
-  static bool get isSigningConfigured => SIGNING_SERVER_ENDPOINT.isNotEmpty;
+  /// Signing is a network call to [SIGNING_SERVER_ENDPOINT], which falls back
+  /// to [defaultSigningServerEndpoint] only when the define is *absent*. A
+  /// define that is present but empty stays empty, so this getter is not a
+  /// usability check: it reports solely whether the build opted out via
+  /// `PROOFMODE_SIGNING_SERVER_ENDPOINT=disabled`. A malformed or empty
+  /// endpoint is caught by [assertSigningEndpointValid] at launch instead.
+  /// Builds that disable it never surface a missing C2PA signature to the
+  /// user; callers gate the "sign or skip" prompt on this (#6058).
+  static bool get isSigningConfigured => !_signingDisabled;
```
NotThatKindOfDrLiz, line 110: This policy still needs an owner call. With the endpoint defaulted and this getter only checking the disabled sentinel, a build with no `PROOFMODE_SIGNING_SERVER_TOKEN` is considered configured and can call the production signer with an empty bearer token. Either make configured mean endpoint plus token, or ensure unconfigured/web-preview builds pass `PROOFMODE_SIGNING_SERVER_ENDPOINT=disabled` deliberately.

NotThatKindOfDrLiz (changes requested):
Pushed `21f5d481f` for the mechanical fixes: local signing endpoint guards now run after `.env` is sourced, the shell guard and Dart validator use the same exact-path rule, startup endpoint validation now runs after crash reporting initialization, and the backend-host default guard also scans the C2PA signing endpoint default. Verified locally with `flutter test test/services/c2pa_signing_endpoint_validation_test.dart`, `dart test test/l10n/arb_consistency_test.dart`, `bash -n` on touched scripts, `bash scripts/check_backend_host_defaults.sh`, targeted `flutter analyze`, and the pre-push changed-file test set.

Requesting changes on the remaining owner decisions before merge:

1. Choose the C2PA failure signal for the metadata sheet. The current branch still depends on `ConnectionStatusService.isOnline`, which is effectively constant true and also causes the pending Timer test failure.
2. Decide the endpoint/token configuration policy. `isSigningConfigured` now treats an unset token as configured, so unconfigured builds can call the production signer with an empty bearer token unless the build explicitly passes `disabled` or the getter requires a token.

Also still needs issue linkage / PR body cleanup before merge (`needs-issue` is still present).

mobile/lib/screens/video_metadata/video_metadata_screen.dart
rabble: Resolved in `811b3a9b1` — switched to `connectivity_plus`.

Confirmed your reading of `ConnectionStatusService` before replacing it: `_isConnected` initialises `true` (`connection_status_service.dart:19`), its only mutator `updateRelayStatus()` has no callers anywhere in the repo, and `checkConnection()` awaits 100ms and sets no state. Constant `true`, so the sheet always picked the service-unavailable copy and `videoMetadataC2paMissingNote` was unreachable in all 21 locales.

Rather than a new connectivity call site, the branch reads through a `deviceIsOfflineProvider` seam over the existing `dmSendConnectivityIsOffline` predicate in `social_providers.dart`, so the connectivity contract stays in one place (that file already says so for the DM path) and widget tests can answer without the platform channel.

That also removes the real service — and its 30s `Timer.periodic` — from the widget tree, so the four pending-`Timer` failures are gone without needing the mock override. Two new cases pin the branch in both directions: `blames the connection when the device is offline` / `blames the service when the device has a network`. 16/16 green in that file.

mobile/lib/services/c2pa_signing_service.dart
rabble: Resolved in `734b27d54` — both halves.

`isSigningConfigured` now means endpoint **plus** token, so a build with no `PROOFMODE_SIGNING_SERVER_TOKEN` (local `flutter run`, or any workflow passing neither define) is unconfigured again, as it was before the endpoint got a default. And both web workflows now pass `PROOFMODE_SIGNING_SERVER_ENDPOINT=disabled` explicitly, so the opt-out has real callers rather than being a sentinel nothing sets.

Also took @mbradley's coverage note while here: the endpoint and token are now injectable on `C2paSigningService`, since the compile-time `String.fromEnvironment` values cannot be varied in-process and the configured / opted-out branches were otherwise unreachable from tests. Two new cases cover them — an untokened build and a `disabled` endpoint both return `C2paSigningFailureReason.disabled` without reaching `signFile`. 12/12 green in `c2pa_signing_service_test.dart`.

Worth flagging that the guard is real: adding the token requirement immediately turned four existing `signVideo` / `resignDerived` tests red, because the test process has no token define.

rabble:
@mbradley @NotThatKindOfDrLiz Both owner decisions are resolved, plus your coverage note. Thanks for pushing the mechanical fixes and for verifying the empty-define premise independently rather than taking it from the description.

**Blocker 2 — the signal.** Your diagnosis of `ConnectionStatusService` is right and I confirmed each part before replacing it. Went with **connectivity_plus** (`811b3a9b1`), routed through a `deviceIsOfflineProvider` seam over the existing `dmSendConnectivityIsOffline` predicate so the connectivity contract stays in one place and widget tests answer without the platform channel.

**Blocker 1 — CI red.** Same commit fixes it, and not by mocking: the real service is out of the widget tree entirely, so its 30s `Timer.periodic` never starts. `video_metadata_screen_test.dart` is 16/16, including two new cases pinning the copy in both directions.

**Q3 — token policy.** Both (`734b27d54`): `isSigningConfigured` requires a token, *and* both web workflows pass `disabled` explicitly.

**Coverage note.** Took it — the endpoint and token are injectable on `C2paSigningService` now, which is what made the changed behaviour testable. Two new cases cover the untokened and `disabled` paths. That seam earned itself immediately: adding the token requirement turned four existing tests red on the first run, because the test process has no token define.

**Q1 (guard vs validator)** was closed by `21f5d481f` — both sides use the exact-path rule. Whether a subpath deployment should be legal is a deployment call, and the two sides agree either way now. **Q2 (silent launch failure)** and **Q4 (host-default guard)** likewise landed in `21f5d481f`.

**PR body**: rewritten. `needs-issue` now links #838, and the false "118 pre-existing tests pass" claim is replaced with what actually ran.

One thing I did *not* do, so it is an explicit choice rather than an oversight: I did not touch `resignDerived`'s guard semantics beyond routing it through the same instance-level check — `942a341`'s fix stands as you wrote it.

Verified locally on `734b27d54`: `flutter analyze lib test` clean, `dart format` clean, `video_metadata_screen_test` (16), `c2pa_signing_service_test` (12), `c2pa_signing_endpoint_validation_test`, `arb_consistency_test` all green, plus `check_backend_host_defaults.sh`.

NotThatKindOfDrLiz:
@rabble takeover changes are pushed on `faa48ab1b`.

What changed:
- Required Codemagic C2PA builds now fail if `PROOFMODE_SIGNING_SERVER_TOKEN` is empty while the endpoint is non-disabled.
- Added guard-script tests for required-token, valid-token, and disabled-without-token cases.
- Tracked the deferred `videoMetadataC2paMissingNoteServiceUnavailable` localization work in #6808 and referenced it from `_knownUntranslatedDebt`.
- Updated the PR body to explicitly own the launch-fail policy for malformed signing endpoints.
- Folded in the adjacent `.env.example` endpoint update and lowered intentional disabled-signing logging from warning to info.
- Rebasing onto current `origin/main` exposed stale `PublishOutcome` / `publishEventAwaitOk` named arguments, so I removed those obsolete call-site args in a separate commit to get the branch green.

Verification:
- Local pre-push: merge-conflict check, analyzer, generated-file check, and changed-file tests all passed.
- CI on `faa48ab1b`: web preview, Generated Files, Format, Analyze, all 4 test shards, VGV Tag Gate, mergeability, semantic PR, and CLA are green.

Please review the takeover commits and merge when you are ready. I am approving from the reviewer side; @mbradley still needs to clear his earlier requested-changes review.

NotThatKindOfDrLiz (approved):
The requested C2PA fixes are in place and verified on the rebased branch.

I re-reviewed the final diff after the takeover commits. The endpoint/token guard now closes the silent-token-misconfiguration path, the l10n debt is tracked in #6808, the launch-fail policy is explicit in the PR body, and the branch is rebased onto current `origin/main` with CI green at `faa48ab1b`.

Approving from my side. @rabble should still review the takeover commits before merging, and @mbradley still needs to clear his outstanding requested-changes review.

mbradley (approved):
Re-verified at `faa48ab1`. Both blockers and the coverage note are resolved.

The connectivity signal is a real one now, and routing it through the existing `dmSendConnectivityIsOffline` predicate keeps that contract in one place rather than adding a second. The offline path gets the connectivity wording and the mounted-check on the new await is right. `isSigningConfigured` requiring a token restores the coupling that made unconfigured builds safe, and opting the web workflows out explicitly is better than relying on the default. The injectable seam is what makes the changed behaviour testable at all, so that one earned its keep.

Approving.

PR #6774, opened 2026-08-06
fix(video-editor): shoot green-screen backgrounds with the camera only

### LINKED ISSUE #6776: fix(video-editor): shoot green-screen backgrounds with the camera only
Auto-created from PR #6774.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6774
**Branch:** `fix/greenscreen-background-camera-only`

hm21:
Closes #6776

## Description

The green-screen editor's **Image** backdrop opened the phone's photo gallery. That is exactly the route an AI-generated picture takes into a Divine video, so the picker is now camera-only — a photo the user shoots on the spot cannot be AI.

The `pickImage` call also moves inside the existing `try`. It used to sit outside it, which was harmless with a gallery source but not with a camera one: a denied permission or a device without a camera would escape as an unhandled exception instead of the screen's "Couldn't load that image" snackbar.

The **Clip** backdrop is unaffected and was never a gap — it reads the app's own clip library (clips recorded in Divine), and there is no `pickVideo` call anywhere in `mobile/lib`.

**Related Issue:** 

## Out of Scope

The chip is still labelled **Image**. Now that it opens the camera directly, "Camera" or "Photo" would read better, but that is a copy change across 17 ARB locales and a product call — left out deliberately.

## Verification

- Manually verified on a real Samsung Galaxy (SM-S942B): green-screen editor → Replace with → **Image** opens the camera directly, and the photo lands as the backdrop.
- `flutter analyze lib/screens/video_editor lib/widgets/video_editor/chroma_key test/screens/video_editor` — clean
- `flutter test test/screens/video_editor test/widgets/video_editor/chroma_key` — green
- New regression test `video_clip_chroma_key_screen_test.dart` asserts the image_picker channel receives `source: camera`. Counter-checked: flipping the source back to `gallery` fails it.

## Type of Change

### DIFF
--- a/mobile/lib/screens/video_editor/video_clip_chroma_key_screen.dart
+++ b/mobile/lib/screens/video_editor/video_clip_chroma_key_screen.dart
@@ -201,19 +201,24 @@ class _VideoClipChromaKeyScreenState extends State<VideoClipChromaKeyScreen> {
     }
   }
 
+  /// Shoots a photo to sit behind the keyed subject.
+  ///
+  /// Camera only, deliberately: the gallery is a route for AI-generated
+  /// imagery to enter a Divine video, and a backdrop the user photographs on
+  /// the spot cannot be one.
   Future<void> _pickImageBackground() async {
-    final picked = await ImagePicker().pickImage(source: ImageSource.gallery);
-    if (picked == null) return;
     try {
+      final picked = await ImagePicker().pickImage(source: ImageSource.camera);
+      if (picked == null) return;
       // `image_picker` hands back a cache path the OS may prune, while clip
       // state persists as a documents-relative basename — so take a copy we
       // own before pointing the key at it.
       //
-      // The copy is normalized rather than byte-for-byte: a portrait photo is
-      // usually stored as landscape pixels plus an EXIF orientation tag, and
-      // the renderer decodes raw bytes, so it would show the backdrop rotated.
-      // Baking the rotation in also caps the photo to a sane size for a
-      // backdrop that is stretched to the video frame anyway.
+      // The copy is normalized rather than byte-for-byte: a photo shot in
+      // portrait is stored as landscape pixels plus an EXIF orientation tag,
+      // and the renderer decodes raw bytes, so it would show the backdrop
+      // rotated. Baking the rotation in also caps the photo to a sane size for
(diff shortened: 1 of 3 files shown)


rabble (approved):
## Overview

Switches the chroma-key background photo from `ImageSource.gallery` to `ImageSource.camera`, and moves the `pickImage` call inside the existing `try`.

The product reasoning is the part worth keeping — stated at the method, not just in the PR description:

> Camera only, deliberately: the gallery is a route for AI-generated imagery to enter a Divine video, and a backdrop the user photographs on the spot cannot be one.

That's a real constraint that a future reader would otherwise "fix" back to gallery, so having it in the code is the right call. The two neighbouring doc comments (`ChromaKeyControls.onPickBackground`, the EXIF-normalisation note) are updated to match rather than left to contradict the new behaviour, which is the kind of upkeep that usually gets skipped.

**Moving `pickImage` inside the `try` is a genuine fix**, not just tidying. Previously a `PlatformException` from the picker — no camera, permission denied, OS cancelling the capture — escaped the method entirely. Now it lands in the catch and the user gets `videoEditorChromaKeyImagePickFailed` via the snackbar. Camera is materially more likely to throw than gallery, so this had to happen alongside the source change.

The test does the right things: `overrideSharedChannel` for the `image_picker` channel (per the shared-channel rule, so it auto-restores and doesn't strand later suites), asserts on `ImageSource.camera.index` in the actual method-call arguments rather than on a widget property, and returns null to stop the screen before it touches the filesystem. The empty `EditorVideo.file('')` trick to keep the preview player out of the test is a clean way to avoid a native plugin.

Approving.

### One product consequence to acknowledge

On a device with no usable camera, there is now **no way at all** to set an image backdrop — the capture throws, the snackbar fires, and the feature is simply unavailable. That's the intended tradeoff given the AI-imagery rationale, and the failure is at least visible rather than silent. Worth being a deliberate decision rather than a discovered one, since it won't show up in testing on real phones.

### Nit

`videoEditorChromaKeyBackgroundImage` is just `"Image"`, so no copy contradicts the new behaviour. But the chip now opens a camera with no warning — a user tapping "Image" expecting a picker gets a viewfinder. If there's room, wording that hints at capture ("Photo" reads no better, but something like a camera icon on the chip) would set the expectation before the OS does.

PR #6777, opened 2026-08-06
refactor(auth): reuse the QR layout as the nostr-connect loading state

### LINKED ISSUE #6778: refactor(auth): reuse the QR layout as the nostr-connect loading state
Auto-created from PR #6777.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6777
**Branch:** `fix/nostr-connect-skeleton-loading`

hm21:
Closes #6778

## Description

The nostr-connect screen had a dedicated `_LoadingContent` widget for the `generating` and `connected` states. It rendered its own layout — a centred `Column` with a `Spacer`, start-aligned text and no horizontal padding — so the screen visibly jumped when the session moved to `listening`, and the title sat flush against the screen edge with no inset.

Dropped it and skeletonized the real `_QrContent` instead. The two loading states now render the same widget as the loaded state, so the chrome matches by construction rather than by keeping two layouts in sync by hand.

Notes on the parts that weren't obvious:

- **Only the payload-dependent parts shimmer** — the QR card and the action bar. The back button, title and compatibility table stay *outside* the `Skeletonizer` because its default `ignorePointers: true` blocks hit testing for the entire subtree, `Skeleton.keep` included. Wrapping the back button would have left it visible but dead while the session generates.
- **The status line is `Skeleton.keep`.** It carries real copy in every state ("Generating connection…" / "Waiting for connection (Xs)" / "Connected, authenticating…"), so turning it into a text bone would hide information we already have. Its spinner keeps running on its own ticker.
- **The QR uses `Skeleton.replace`, not an in-place shimmer.** Skeletonizing it in place still builds the child, which would run the QR encoder over the empty placeholder payload. `Skeleton.replace` swaps the subtree at build time, so the encoder never sees it.
- `_QrContent` now takes a `statusMessage` string instead of `elapsedSeconds`. The screen owns the per-state copy, which keeps the component state-agnostic and avoids threading a `NostrConnectState` into the widget layer.

The shimmer comes from `vineSkeletonEffectOf(context)` and the bone from `context.vineColors.skeleton`, matching the existing skeleton pattern in `people_section.dart`.

**Related Issue:** 

## Out of Scope

The `connected` (authenticating) state also renders as a skeleton, even though the QR payload exists by then. That preserves the previous behaviour — that state was fully non-interactive before — and keeps the action bar from being tappable mid-authentication. Leaving the real QR visible there instead is a reasonable alternative if it reads better on device.

## Verification

- `dart format` — no changes
- `flutter analyze lib/screens/auth/nostr_connect_screen.dart` — no issues
- `flutter test test/screens/auth/nostr_connect_screen_test.dart test/screens/auth/nostr_connect_failure_message_test.dart` — 7 passing
- All four design-system ceiling ratchets (`raw_textstyle`, `raw_colors`, `material_button`, `raw_dialog`) — green
- Manual verification on a real Samsung Galaxy still pending — draft until the generating → listening transition is checked on device.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/screens/auth/nostr_connect_screen.dart
+++ b/mobile/lib/screens/auth/nostr_connect_screen.dart
@@ -492,8 +492,11 @@ class _QrContent extends StatelessWidget {
   final VoidCallback onShareUrl;
   final VoidCallback onAddBunker;
 
-  /// Whether the session payload is still pending. Skeletonizes — and, via
-  /// its default `ignorePointers`, disables — the QR code and action bar.
+  /// Whether the session payload is still pending. Skeletonizes the QR code
+  /// and action bar. [Skeletonizer]'s `ignorePointers` blocks touch but not
+  /// the semantics tree, so the action bar is also disabled for screen
+  /// readers by nulling its callbacks (see [_ActionBar.isLoading]) — otherwise
+  /// a double-tap could open the bunker sheet and drop the live session.
   final bool isLoading;
 
   @override
@@ -544,6 +547,7 @@ class _QrContent extends StatelessWidget {
 
                 // Action bar
                 _ActionBar(
+                  isLoading: isLoading,
                   onCopyUrl: onCopyUrl,
                   onShareUrl: onShareUrl,
                   onAddBunker: onAddBunker,
@@ -591,8 +595,10 @@ class _QrCodeCard extends StatelessWidget {
             color: VineTheme.whiteText,
             borderRadius: BorderRadius.circular(12),
           ),
-          // Replaced rather than skeletonized in place: the encoder would
-          // otherwise render a QR for a placeholder payload.
+          // Replaced rather than skeletonized in place: in the steady loading
+          // state the encoder never runs over the placeholder payload. (The
+          // switch animation still builds one throwaway `QrImageView('')`
+          // mid-transition; harmless — an empty payload encodes fine.)
           child: Skeleton.replace(
             replace: isLoading,
             width: 200,
@@ -658,11 +664,18 @@ class _ActionBar extends StatelessWidget {
     required this.onCopyUrl,
     required this.onShareUrl,
     required this.onAddBunker,
+    this.isLoading = false,
   });
 
   final VoidCallback onCopyUrl;
   final VoidCallback onShareUrl;
   final VoidCallback onAddBunker;
+
+  /// While loading, the buttons are handed `null` callbacks so they publish
+  /// no tap action — inert to both touch and screen readers. [Skeletonizer]'s
+  /// `ignorePointers` only blocks touch, so nulling here is what keeps a
+  /// double-tap from activating a bone.
+  final bool isLoading;
 
   @override
   Widget build(BuildContext context) {
@@ -689,7 +702,7 @@ class _ActionBar extends StatelessWidget {
                     size: MediaQuery.textScalerOf(context).scale(24),
                   ),
                   label: context.l10n.authCopyUrl,
-                  onTap: onCopyUrl,
+                  onTap: isLoading ? null : onCopyUrl,
                 ),
                 _ActionButton(
                   icon: DivineIcon(
@@ -698,7 +711,7 @@ class _ActionBar extends StatelessWidget {
                     size: MediaQuery.textScalerOf(context).scale(24),
                   ),
                   label: context.l10n.authShare,
-                  onTap: onShareUrl,
+                  onTap: isLoading ? null : onShareUrl,
                 ),
                 _ActionButton(
                   icon: DivineIcon(
@@ -707,7 +720,7 @@ class _ActionBar extends StatelessWidget {
                     size: MediaQuery.textScalerOf(context).scale(24),
                   ),
                   label: context.l10n.authAddBunker,
-                  onTap: onAddBunker,
+                  onTap: isLoading ? null : onAddBunker,
                 ),
               ],
             ),
@@ -728,7 +741,10 @@ class _ActionButton extends StatelessWidget {
 
   final Widget icon;
   final String label;
-  final VoidCallback onTap;
+
+  /// `null` while the session is loading, so the button is inert to both
+  /// touch and screen-reader activation.
+  final VoidCallback? onTap;
 
   @override
   Widget build(BuildContext context) {
--- a/mobile/lib/screens/auth/nostr_connect_screen.dart
+++ b/mobile/lib/screens/auth/nostr_connect_screen.dart
@@ -16,6 +16,7 @@ import 'package:openvine/services/auth_service.dart';
 import 'package:openvine/widgets/auth_back_button.dart';
 import 'package:qr_flutter/qr_flutter.dart';
 import 'package:share_plus/share_plus.dart';
+import 'package:skeletonizer/skeletonizer.dart';
 import 'package:unified_logger/unified_logger.dart';
 
 /// Screen for NIP-46 client-initiated connections via nostrconnect:// URL.
@@ -388,18 +389,33 @@ class _NostrConnectScreenState extends ConsumerState<NostrConnectScreen> {
       backgroundColor: context.vineColors.background,
       body: SafeArea(
         child: switch (_sessionState) {
-          NostrConnectState.idle || NostrConnectState.generating =>
-            _LoadingContent(message: context.l10n.authGeneratingConnection),
-          NostrConnectState.listening => _QrContent(
-            connectUrl: _connectUrl ?? '',
-            elapsedSeconds: _elapsedTimer.elapsed.inSeconds,
+          NostrConnectState.idle || NostrConnectState.generating => _QrContent(
+            connectUrl: '',
+            statusMessage: context.l10n.authGeneratingConnection,
+            isLoading: true,
             onBack: () => context.pop(),
             onCopyUrl: _copyUrl,
             onShareUrl: _shareUrl,
             onAddBunker: _showPasteBunkerDialog,
           ),
-          NostrConnectState.connected => _LoadingContent(
-            message: context.l10n.authConnectedAuthenticating,
+          NostrConnectState.listening => _QrContent(
+            connectUrl: _connectUrl ?? '',
+            statusMessage: context.l10n.authWaitingForConnection(
+              _elapsedTimer.elapsed.inSeconds,
+            ),
+            onBack: () => context.pop(),
+            onCopyUrl: _copyUrl,
+            onShareUrl: _shareUrl,
+            onAddBunker: _showPasteBunkerDialog,
+          ),
+          NostrConnectState.connected => _QrContent(
+            connectUrl: _connectUrl ?? '',
+            statusMessage: context.l10n.authConnectedAuthenticating,
+            isLoading: true,
+            onBack: () => context.pop(),
+            onCopyUrl: _copyUrl,
+            onShareUrl: _shareUrl,
+            onAddBunker: _showPasteBunkerDialog,
           ),
           NostrConnectState.timeout => _ErrorContent(
             title: context.l10n.authConnectionTimedOut,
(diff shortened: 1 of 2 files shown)


mobile/lib/screens/auth/nostr_connect_screen.dart
```diff
@@ -452,60 +468,34 @@ String resolveNostrConnectFailureMessage(
...
   final String connectUrl;
-  final int elapsedSeconds;
+  final String statusMessage;
   final VoidCallback onBack;
   final VoidCallback onCopyUrl;
   final VoidCallback onShareUrl;
   final VoidCallback onAddBunker;
 
+  /// Whether the session payload is still pending. Skeletonizes — and, via
+  /// its default `ignorePointers`, disables — the QR code and action bar.
```
realmeylisdev: `ignorePointers` doesn't reach the semantics tree. `RenderSkeletonizer` overrides `hitTest` and nothing else — no `visitChildrenForSemantics`, no semantics config — so the action buttons keep their tap actions while skeletonized.

Measured on this branch, in `generating`:

- pointer tap on "Add bunker" → blocked, sheet stays closed
- `semanticsOwner.performAction(node.id, SemanticsAction.tap)` on the same node → **sheet opens**

That second one is what TalkBack and VoiceOver send on a double-tap. It bites hardest in `connected`: opening that sheet and submitting a bunker URL runs `_switchedToBunker = true`, `_sessionAttempt++`, `cancelNostrConnect()` — so a screen-reader user can drop an already-approved NIP-46 session mid-authentication. That state isn't a blink either; `_applyNostrConnectSuccess` does a relay connect plus a `pullPubkey()` round trip before it navigates.

Smallest fix is to stop handing the buttons a callback while loading: make `_ActionButton.onTap` nullable and pass `null` from `_ActionBar` when `isLoading`. A `GestureDetector` with a null `onTap` publishes no tap action, so touch and semantics agree. Both shapes under `Skeletonizer(enabled: true)`:

```
onTap: () {}  → hasTapAction true
onTap: null   → hasTapAction false
```

Worth pinning with a test — nothing asserts the action bar is inert while loading right now, which is how this got through.

mobile/test/screens/auth/nostr_connect_screen_test.dart
```diff
@@ -15,9 +31,94 @@ void main() {
...
+      // Skeleton.replace is off, so the real QR encoder is built (not the
+      // shimmer placeholder shown while generating).
+      expect(find.byType(QrImageView), findsOneWidget);
+      expect(
+        tester.widget<Skeletonizer>(find.bySubtype<Skeletonizer>()).enabled,
+        isFalse,
+      );
+
+      await tester.pumpWidget(const SizedBox());
+    });
```
realmeylisdev, line 221: Two gaps worth closing while you're in here.

`connected` is the other branch this PR converted (`_LoadingContent` → skeletonized `_QrContent`) and nothing drives it. The listening test can't reach it transitively either — `nostrConnectStateStream` is unstubbed and nullable, so the `?.listen` at `nostr_connect_screen.dart:116` is a no-op. That leaves `isLoading: true` on line 414 unpinned, and the parameter defaults to `false`, so dropping it stays green while the screen would paint a live QR for a session that's already been consumed.

The other: the reason this PR exists — chrome doesn't shift between loading and loaded — isn't asserted anywhere. One test that pumps both states at a fixed surface size and compares `tester.getTopLeft` of the compatibility-table header would lock in the thing you actually set out to fix.

mobile/lib/screens/auth/nostr_connect_screen.dart
```diff
@@ -533,21 +524,32 @@ class _QrContent extends StatelessWidget {
 
           const SizedBox(height: 32),
 
-          // QR code card
-          _QrCodeCard(connectUrl: connectUrl),
+          Skeletonizer(
+            enabled: isLoading,
+            enableSwitchAnimation: true,
```
realmeylisdev, line 532: Non-blocking: `enableSwitchAnimation` softens "the encoder never sees the placeholder payload". `SkeletonizerScope` sits above the `AnimatedSwitcher`, and `Skeleton.replace` reads it through `dependOnInheritedWidgetOfExactType` — so when `enabled` flips, the outgoing skeleton subtree rebuilds with `replace: false` and builds a real `QrImageView` still holding the stale `''`.

Counted across the generating → listening flip:

```
before the flip   QrImageView count 0
flip .. +300ms    QrImageView count 2
+350ms on         QrImageView count 1
```

`QrImageView(data: '')` doesn't throw, so it's one throwaway encode per transition — no action needed. Flagging it only so the comment on `_QrCodeCard` isn't read as a harder guarantee than it is.

(I also checked whether the fade opens a pointer window on the action bar going listening → connected — it doesn't, taps stay blocked the whole way through.)

realmeylisdev (changes requested):
Solid refactor — one layout instead of two is the right call, and it quietly fixes the missing inset on the loading title.

One thing needs changing before merge: the skeleton disables the action bar for touch but **not** for screen readers, so those buttons stay live in exactly the state you're locking down. Detail and a verified fix inline. Two non-blocking notes under it.

Ran on the branch (base is current `main`):

- `flutter test test/screens/auth/` — 123 pass; the two router tests that mount this screen — 55 pass
- four design-system ratchets + the services and test-structure floors — green
- `dart format` clean, no ARB churn, all three status keys present in 22/22 locales

Checked and cleared: no protocol surface (connect URL, secret, relays and session lifecycle are all untouched), and the back-during-teardown paths are already reachable on `main` via OS back, so they aren't yours.

mobile/lib/screens/auth/nostr_connect_screen.dart
```diff
@@ -588,23 +591,38 @@ class _QrCodeCard extends StatelessWidget {
...
             borderRadius: BorderRadius.circular(12),
           ),
-          child: QrImageView(
-            data: connectUrl,
-            size: 200,
-            backgroundColor: VineTheme.whiteText,
-            errorCorrectionLevel: QrErrorCorrectLevel.M,
+          // Replaced rather than skeletonized in place: the encoder would
+          // otherwise render a QR for a placeholder payload.
+          child: Skeleton.replace(
```
realmeylisdev, line 602: Device-pass observation. Because the bone sits *inside* the white quiet zone, the 12px padding paints as a white ring around a dark shimmering square — on an otherwise all-dark screen it reads as an empty white picture frame rather than a placeholder card. In the loaded state that same white area is filled by the QR, so it only looks odd while loading.

If you want the tile to read as one skeleton, moving `Skeleton.replace` outside the white `Container` — replacing the quiet zone along with the QR — makes the whole card shimmer uniformly. Your call entirely; flagging it because it's the bit the device pass was for.

mobile/lib/screens/auth/nostr_connect_screen.dart
```diff
@@ -533,21 +524,32 @@ class _QrContent extends StatelessWidget {
...
-          const SizedBox(height: 32),
+                const SizedBox(height: 32),
 
-          // Action bar
-          _ActionBar(
-            onCopyUrl: onCopyUrl,
-            onShareUrl: onShareUrl,
-            onAddBunker: onAddBunker,
+                // Action bar
+                _ActionBar(
```
realmeylisdev, line 549: Same device pass: while loading, the action bar keeps its three icons at full vine-green and loses all three labels.

That's skeletonizer's canvas split — `drawParagraph` is replaced with a shaded bone (so `Copy URL` / `Share` / `Add bunker` vanish), while `drawPicture` passes straight through to the parent canvas untouched, and `DivineIcon` paints its SVG via `drawPicture`. So the buttons come out half-skeletonized: live icon, missing label.

`Skeleton.unite` around each `_ActionButton` would give one bone per button instead. Again your call — noting it since it isn't obvious from reading the diff.

realmeylisdev (commented):
Ran this on a clean iPhone 17 Pro sim (fresh install, real relays) and drove the whole flow. Console stayed clean throughout — zero framework exceptions, assertions or overflows across 793 log lines.

**Your main claim holds.** I pulled the iOS accessibility tree in both states and every element sits at an identical y:

| | generating | listening |
|---|---|---|
| title | 150 | 150 |
| status line | 588 | 588 |
| Copy URL / Share / Add bunker | 641 | 641 (identical x too) |
| Compatible Signer apps | 760 | 760 |

No shift at all. The refactor does exactly what you said it would.

**The blocker reproduces on device.** The `Add bunker` AX node is byte-identical between the skeletonized and the live state — same frame, same role, `enabled: true`. And tapping it in `listening` opens the bunker sheet, so VoiceOver has nothing to tell the two states apart by. I couldn't drive `connected` on device (it needs a real signer approval, and the session times out at 120s), but it renders the same `_QrContent(isLoading: true)`, so the same reasoning carries.

**One thing to add to the `enableSwitchAnimation` note above:** the mid-fade duplication isn't just the QR — the whole subtree doubles in the a11y tree for ~300ms. A dump caught mid-transition:

```
y=299  qr code
y=299  qr code
y=588  Waiting for connection... 0s
y=588  Generating connection...
y=641  Copy URL / Share / Add bunker
y=641  Copy URL / Share / Add bunker
```

Two of every control plus two contradictory status lines at the same coordinates. Invisible to a sighted user, less so on a rotor.

Two notes inline on how the loading state actually renders — you had the device pass down as pending, so that's what I went looking at.

mobile/lib/screens/auth/nostr_connect_screen.dart
hm21: Fixed in c7d2cd6db. Confirmed `RenderSkeletonizer` overrides only `hitTest` and nothing for semantics, so `ignorePointers` never reached the a11y tree — your read was exactly right. `_ActionButton.onTap` is now nullable and `_ActionBar` passes `null` while `isLoading`, so the buttons publish no tap action to touch or screen readers. Pinned inert in both `generating` and `connected`.

mobile/test/screens/auth/nostr_connect_screen_test.dart
hm21: Both closed in c7d2cd6db. Added a `connected`-state test that drives the branch via the state stream and pins `isLoading: true` (QR stays replaced, action bar inert). Added a chrome-stability test asserting `getTopLeft` of the compatibility header is identical between the loading and loaded states.

mobile/lib/screens/auth/nostr_connect_screen.dart
hm21: Reworded the `_QrCodeCard` comment in c7d2cd6db to scope the guarantee to the steady loading state and call out the one throwaway `QrImageView('')` the switch animation builds mid-transition. No behavior change — thanks for measuring it.

mobile/lib/screens/auth/nostr_connect_screen.dart
hm21: Leaving as-is. The white quiet zone is part of the card chrome the loaded state also shows, so keeping it present in both states is what preserves the "chrome matches by construction" property — replacing it while loading would make the white reappear on load-complete, i.e. a shift in exactly that region. Cosmetic-only and a deliberate trade for the no-jump goal. Thanks for the device pass.

mobile/lib/screens/auth/nostr_connect_screen.dart
hm21: Leaving as-is. The buttons are inert while loading (null callbacks), so the live-icon/boned-label look is cosmetic, and the icons double as an at-a-glance hint of what the bar offers once loaded. `Skeleton.unite` is a fair alternative but not worth the churn for a sub-second state. Good catch on the canvas-split detail though.

realmeylisdev (approved):
All addressed — approving.

Re-verified `c7d2cd6` on the branch: the action bar hands its buttons null callbacks while loading, so they publish no tap action to touch or screen readers, in both `generating` and `connected`. I also mutation-checked the new tests — reverting the null on Add bunker fails the generating and connected inertness tests, so the regression is genuinely pinned. Screen suite 11/11, all CI checks green.

The two cosmetic notes declined as-is are fine by me — the quiet-zone choice is the one that preserves the no-shift property, and the half-skeletonized bar is inert anyway.

PR #6779, opened 2026-08-06
feat(feed): add immersive video viewing mode

### LINKED ISSUE #6234: feat(feed): add immersive video viewing mode
### What would you like?

Add an immersive video viewing mode that allows users to temporarily hide the UI while watching videos.

The exact interaction is intentionally left open for UX exploration. Possible ideas include a pinch gesture, a long press, a toggle in the menu, or another interaction that provides a natural user experience.

The goal is to make it easy to quickly focus on the video without permanently changing the viewing experience.

It allows users to enjoy videos without interface elements covering the content while still keeping all existing controls easily accessible.

Since different users may prefer different interactions, the exact way to enter and exit immersive mode should be determined during the UX design process.

### When would you use this?

Whenever I want to focus entirely on the video or appreciate the content without the interface covering parts of the screen.

### Anything else?

Community feedback on discord [here](https://discord.com/channels/1470168817589158040/1470256031434149953/1528732630282010767) suggested multiple possible interactions, but there was no clear consensus. The implementation should be driven by the best user experience rather than a predefined interaction.

Possible ideas mentioned by the community include:
- Pinch to enter/exit immersive mode.
- Long press to temporarily hide the UI.
- A menu toggle to enable or disable the mode.

hm21:
## Description

Adds an immersive viewing mode to the feed: **press and hold anywhere on a video** and every chrome layer over it cross-fades out (180 ms), so the frame can be seen unobstructed. Release and it all comes straight back.

Hidden while held: the feed-mode header (`For You` + `⋯`), the author block, caption, action rail, both gradients, and the paused play indicator — plus the fullscreen reel's app bar. The shell's bottom nav intentionally stays.

Each chrome layer carries its own `FeedImmersiveChrome` wrapper rather than sharing one, so the double-tap heart animation stays independent of the fade.

**Related Issue:** Closes #6234

### Why hold-to-peek and not a toggle

The issue leaves the interaction open and lists pinch, long-press and a menu toggle as community suggestions. It also asks for a way to focus on a video *"without permanently changing the viewing experience"*, and describes the mode as temporarily hiding the UI — which is what a transient hold gives you, and a sticky toggle does not.

Practical consequences that made this the safer of the three:

- Nothing is persisted, and there is no state the UI can get stuck in. A sticky toggle needs an exit affordance, and the obvious one (single tap) is already play/pause.
- No gesture conflicts. A short press rejects the long-press recognizer so tap-to-pause is untouched; a completed hold suppresses the tap, so releasing does not pause the video. Double-tap-to-like is unaffected — the first tap's release lands well inside the 500 ms long-press deadline.

### Why a raw `Listener` backstops the release

`LongPressGestureRecognizer` fires **neither** `onLongPressEnd` nor `onLongPressCancel` when a pointer is cancelled *after* the press was accepted (`handlePrimaryPointer` routes `PointerCancelEvent` to `_checkLongPressCancel`, which no-ops once the recognizer has left `possible`). Without a backstop, an incoming call or a system gesture mid-hold would leave the chrome permanently hidden. `dispose()` covers the third case — the item being torn down mid-hold.

That `Listener` needs `behavior: HitTestBehavior.translucent`. With the default `deferToChild` it never enters the hit-test path over empty video area at all: the translucent `GestureDetector` beneath it adds *itself* to the hit-test result but still returns `false`, so a `deferToChild` parent concludes it was missed. This was caught by the pointer-cancel test, which failed before the fix.

### Why the hold is not gated on player readiness

Tap and double-tap are gated on `interactiveReady` because they mutate the player or publish a like. Hiding chrome has neither effect and is just as valid over a still-loading or letterboxed frame, so the hold works immediately.

### Architecture

A page-scoped `FeedImmersiveCubit` (one idempotent bool) is owned by `VideoFeedView` and `_FullscreenFeedContentState`, alongside the existing `FeedAutoAdvanceCubit` — it has to sit above both `FeedVideos`, which raises the flag, and the page-level header/app bar, which fade against it. `FeedImmersiveChrome` wraps each chrome layer in `AnimatedOpacity` + `IgnorePointer` and looks the cubit up **nullably**, so surfaces that render feed chrome outside a feed page (the video-editor preview's `FeedModeSwitch`, isolated widget tests) simply never go immersive instead of throwing. Reduced motion (`MediaQuery.disableAnimationsOf`) gets the same end state with `Duration.zero`.

No new user-facing strings, so no ARB changes. Only `onLongPressStart` / `onLongPressEnd` are wired — never `onLongPress` — so no long-press semantics action is published: the peek would be meaningless to a screen reader, which needs the chrome it hides.

## Out of Scope

- **Discoverability.** No coach-mark or first-run hint. The issue carries `needs design` with @Chardot and @NotThatKindOfDrLiz assigned; this ships the interaction, not their sign-off on how it is surfaced.
- **Bottom navigation bar.** Deliberately left visible — hiding it means a cross-cutting signal into `AppShell`'s `Scaffold`, which is a bigger change than this feature warrants.
- **Captions.** `subtitleLayer` lives inside `VideoOverlayActions`' author Column, so it fades with the rest of the block. Separating it means restructuring that widget; the hold is user-initiated and momentary, so it was not worth the churn here.

## Verification

- `dart format` — clean
- `flutter test test/widgets/video_feed_item/ test/screens/feed/ test/services/haptic_service_test.dart test/screens/explore` — 493 passed
- `flutter test` on the other suites that mount these screens (hashtag feed ×3, notifications view, profile videos grid, fullscreen feed redirect, foreground idle warmup) — 84 passed
- All four design-system ceilings (`raw_textstyle`, `raw_colors`, `material_button`, `raw_dialog`), `check_untested_services_floor.sh`, `check_test_unit_structure.sh` — pass
- No goldens cover feed chrome, and `Opacity` at 1.0 is a no-op paint, so nothing shifts at rest

New/extended coverage:

- `feed_immersive_cubit_test.dart` — idempotent enter/exit contract
- `feed_immersive_chrome_test.dart` — fade + pointer blocking, no-provider fallback, reduced-motion instant switch
- `feed_videos_test.dart` → `hold to peek` — hides on hold and restores on release, hides the paused play indicator, restores on pointer **cancel**, a short tap leaves the chrome alone, and the flag lowers when the feed is torn down mid-hold
- `feed_mode_switch_test.dart` → `Immersive mode` — the header fades with the cubit

Manually verified on a real Samsung Galaxy device.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/widgets/video_feed_item/feed_videos.dart
+++ b/mobile/lib/widgets/video_feed_item/feed_videos.dart
@@ -1019,6 +1019,17 @@ class __OverlayState extends ConsumerState<_Overlay> {
                               // like, while this only hides chrome, which is
                               // just as valid over a still-loading frame.
                               //
+                              // Excluded from semantics, and kept on its own
+                              // detector so the tap action above still is
+                              // published. A `GestureDetector` publishes
+                              // `SemanticsAction.longPress` for ANY long-press
+                              // callback, `onLongPressStart` included — and
+                              // firing that action delivers no pointer events,
+                              // so the release path below would never run and
+                              // a screen-reader user would be left with every
+                              // control hidden and pointer-blocked until the
+                              // item was disposed.
+                              excludeFromSemantics: true,
                               onLongPressStart: (_) => _enterImmersive(),
                               child: const SizedBox.expand(),
                             ),
--- a/mobile/lib/widgets/video_feed_item/feed_videos.dart
+++ b/mobile/lib/widgets/video_feed_item/feed_videos.dart
@@ -651,6 +651,9 @@ class __OverlayState extends ConsumerState<_Overlay> {
     // The cubit outlives this overlay, so the flag has to come down here or
     // the surface would be left with permanently hidden chrome.
     _exitImmersive();
+    // [didUpdateWidget] re-points this State at a different video, so the
+    // pointer set must not outlive the item that filled it.
+    _immersivePointers.clear();
     _heartTrigger.dispose();
     super.dispose();
   }
--- a/mobile/lib/widgets/video_feed_item/feed_videos.dart
+++ b/mobile/lib/widgets/video_feed_item/feed_videos.dart
@@ -600,9 +600,14 @@ class __OverlayState extends ConsumerState<_Overlay> {
   /// without reaching through a deactivated [BuildContext].
   FeedImmersiveCubit? _immersiveCubit;
 
-  /// Whether *this* item raised the immersive flag. Guards every exit path
-  /// so releasing an unrelated pointer can't clear a hold we never started.
+  /// Whether *this* item raised the immersive flag. Guards enter/exit so an
+  /// unrelated pointer can't clear a hold we never started.
   bool _isHoldingForImmersive = false;
+
+  /// Pointers currently down over this item. The peek ends only when the last
+  /// one lifts, so an incidental second finger (a resting thumb, a pinch
+  /// attempt) can't restore the chrome while the hold finger is still down.
+  final Set<int> _immersivePointers = <int>{};
 
   @override
   void initState() {
@@ -660,12 +665,22 @@ class __OverlayState extends ConsumerState<_Overlay> {
     cubit.enter();
   }
 
-  /// Brings the chrome back.
+  /// Drops a lifted/cancelled pointer and, once none remain, brings the chrome
+  /// back.
   ///
-  /// Reached from three directions because `LongPressGestureRecognizer` only
-  /// reports `onLongPressEnd` for a clean lift: a `PointerCancelEvent` after
-  /// the press was accepted fires neither `onLongPressEnd` nor
-  /// `onLongPressCancel`, which would otherwise strand the chrome hidden.
+  /// Driven off the raw [Listener] rather than the gesture callbacks:
+  /// `LongPressGestureRecognizer` reports neither `onLongPressEnd` nor
+  /// `onLongPressCancel` for a pointer cancelled after the press was accepted,
+  /// which would otherwise strand the chrome hidden. Flutter still delivers the
+  /// terminal event to the down-time hit path even after the [Listener] has
+  /// unmounted (a mode flip mid-hold), so this stays the reliable exit.
+  void _handleImmersivePointerEnd(int pointer) {
+    _immersivePointers.remove(pointer);
+    if (_immersivePointers.isEmpty) _exitImmersive();
+  }
+
+  /// Brings the chrome back. Idempotent — the no-op guard lets [dispose] and
+  /// the last-pointer-up path both call it unconditionally.
   void _exitImmersive() {
     if (!_isHoldingForImmersive) return;
     _isHoldingForImmersive = false;
@@ -945,19 +960,25 @@ class __OverlayState extends ConsumerState<_Overlay> {
                   button: true,
                   label: context.l10n.videoPlayerPlayVideo,
                   hint: isOwnVideo ? null : context.l10n.videoPlayerTapHint,
-                  // The raw [Listener] is the backstop for immersive mode: a
-                  // cancelled pointer never reaches the long-press callbacks
-                  // below (see [_exitImmersive]), and a Listener sits outside
+                  // The raw [Listener] owns immersive mode's release: it counts
+                  // the pointers over the item and exits once the last lifts,
+                  // because a pointer cancelled after the press was accepted
+                  // reaches neither long-press callback (see
+                  // [_handleImmersivePointerEnd]), and a Listener sits outside
                   // the gesture arena so it can't steal the press either.
                   child: Listener(
                     // Must match the translucent child below: with
                     // `deferToChild` a press on empty video area never adds
                     // this Listener to the hit-test path (the translucent
                     // GestureDetector adds itself but still reports a miss),
-                    // so the cancel event would never arrive.
+                    // so the pointer events would never arrive.
                     behavior: HitTestBehavior.translucent,
-                    onPointerUp: (_) => _exitImmersive(),
-                    onPointerCancel: (_) => _exitImmersive(),
+                    onPointerDown: (event) =>
+                        _immersivePointers.add(event.pointer),
+                    onPointerUp: (event) =>
+                        _handleImmersivePointerEnd(event.pointer),
+                    onPointerCancel: (event) =>
+                        _handleImmersivePointerEnd(event.pointer),
                     child: GestureDetector(
                       behavior: .translucent,
                       onTap: interactiveReady ? _handlePlayerTap : null,
@@ -974,12 +995,13 @@ class __OverlayState extends ConsumerState<_Overlay> {
                       // publish a like, while this only hides chrome, which
                       // is just as valid over a still-loading frame.
                       //
-                      // Only `onLongPressStart` / `onLongPressEnd` are wired
-                      // (never `onLongPress`), so no long-press semantics
-                      // action is published — the peek would be meaningless
-                      // to a screen reader, which needs the chrome it hides.
+                      // Only `onLongPressStart` is wired (never `onLongPress`),
+                      // so no long-press semantics action is published — the
+                      // peek would be meaningless to a screen reader, which
+                      // needs the chrome it hides. The release runs through the
+                      // [Listener] above, not `onLongPressEnd`, so an incidental
+                      // second finger can't cut the hold short.
                       onLongPressStart: (_) => _enterImmersive(),
-                      onLongPressEnd: (_) => _exitImmersive(),
                       child: Stack(
                         children: [
                           if (widget.controller != null)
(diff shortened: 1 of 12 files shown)


mobile/lib/widgets/video_feed_item/feed_videos.dart
```diff
@@ -905,24 +965,97 @@ class __OverlayState extends ConsumerState<_Overlay> {
...
+                        // it up would contradict the gesture.
+                        FeedImmersiveChrome(
+                          child: PausedVideoOverlay(
                             controller: widget.controller!,
                             isVisible: widget.isActive,
                           ),
-                        _FeedItemActions(
+                        ),
+                      FeedImmersiveChrome(
+                        child: _FeedItemActions(
```
mbradley, line 1058: **Captions fade with the rest of the chrome.** `subtitleLayer` is threaded into `_FeedItemActions`, so it goes with the author block and rail.

For a hearing viewer that is probably right, since the caption does occlude the frame the peek is meant to reveal. For a deaf or hard-of-hearing viewer it is different in kind: the caption is the content, so the hold removes what they are watching rather than what is covering it.

Two options:

1. Hoist `subtitleLayer` out of the `FeedImmersiveChrome` wrapper so captions stay up during a hold. Costs the `VideoOverlayActions` restructure you already flagged in the description.
2. Keep current behaviour and record the reasoning in a comment.

I did not pick one because it is a product and accessibility tradeoff with a real implementation cost, and it is your call rather than mine. Which do you want?

mbradley (changes requested):
Took this over under the shared-ownership policy and pushed fixes for the findings that had clear remediations. Three are behaviour changes against `main` that I confirmed by running tests on both sides rather than by reading the diff.

## 1. The peek swallowed taps on the action rail

The chrome layers were descendants of the `onLongPressStart` detector, so any press held past `kLongPressTimeout` anywhere on the action rail was claimed by the video's long press. The rail's own tap recognizer only accepts on pointer up, so it lost the arena, and `IgnorePointer` then faded the control out under the viewer's finger.

Measured with a real `VideoActionButton`:

| | `main` | this branch, before the fix |
|---|---|---|
| slow press on Report | fires | swallowed, 0 taps |
| quick tap on Report | fires | fires |

Like and Repost were unaffected only because they already pass their own `onLongPress`, which left Report, More, Share and Help Classify exposed.

Worth recording that the smaller-looking fix does not work. Adding a no-op `onLongPress` sink per button stops the chrome vanishing, but the button's own recognizer then swallows the tap instead, so Report stays unreachable. I measured that variant too and got 0 taps. The fix makes the chrome a sibling above the gesture surface, so an opaque button wins the hit test outright and the peek never sees that pointer.

## 2. A screen-reader long press entered the peek with no way out

`GestureDetector` publishes `SemanticsAction.longPress` for any long-press callback, `onLongPressStart` included, so the comment saying no action was published had it backwards.

Performing that action enters immersive mode but delivers no pointer events, and the release path is driven entirely by pointer up/cancel. Probed on the real tree: `hasAction(longPress)` was true, and firing it gave `entered=1, exited=0`. The chrome then stayed hidden and pointer-blocked until the item was disposed, so a screen-reader user lost every control with no visible way back.

The peek now sits on its own detector, so `excludeFromSemantics` can drop the long-press action without also dropping the tap action that screen-reader play/pause depends on.

## 3. Chrome accepted taps while invisible

`IgnorePointer` tracked the flag, which drops the instant the viewer lifts, while the fade takes 180ms. Measured 20ms into the fade back in: opacity `0.0`, `ignoring` false, and a tap landed and published a like. That window is exactly when a follow-up tap is most likely, and the control being hit is invisible.

## What I pushed

1. `fix(feed): correct the long-press cancel rationale`
2. `refactor(feed): drop the redundant enter() guard`
3. `fix(feed): clear the immersive pointer set on dispose`
4. `fix(feed): keep chrome pointer-blocked until it is visible again`
5. `fix(feed): keep the action rail out from under the peek gesture`
6. `fix(feed): stop publishing a long-press semantics action`
7. `fix(feed): recover a hold whose terminal pointer event was lost`
8. `test(feed): pin the peek's content gate and mid-hold mode flip`

On (1): `onLongPressCancel` does fire for a pointer cancelled after acceptance; only `onLongPressEnd` stays silent. The comment said neither fires. The raw `Listener` is still the right mechanism, because only counting pointers can distinguish an incidental second finger lifting from the hold finger lifting, so this corrects the reasoning and not the design.

On (7): the lifecycle hook alone was not enough. It clears the cubit but leaves the item still believing it holds, so that item could never peek again. The item now self-heals on pointer down, since an empty pointer set means nothing is down and any hold it still owns is stale.

## What I checked and did not change

A hold cannot reveal content behind a content warning, age gate or moderated-content overlay. `_resolveOverlayMode` returns those modes early and the gesture exists only in the interactive branch. Nothing pinned that, so I added a test: if the blur ever becomes a sibling in the interactive stack, it now fails loudly instead of silently becoming a bypass.

## Validation

498 tests pass across `test/widgets/video_feed_item/`, `test/screens/feed/` and `test/services/haptic_service_test.dart`. `flutter analyze lib test` clean, `dart format` clean, full CI green.

Every new test was mutation checked and goes red when the behaviour it describes is reverted. The same check found one of the original tests could not fail: `repeated enter does not re-emit` passes with or without the guard it appears to pin, because `Cubit.emit` already suppresses an equal state. Commit (2) removes the dead guard and documents why `exit()`'s guard is genuinely load-bearing.

One shard failed on the first run, in `test/unit/services/upload_manager_get_by_path_test.dart`. That file is tagged `skip_very_good_optimization` so it runs in its own isolate, it is untouched by this branch, and it passed on re-run and across six local runs with random ordering. Recording it here rather than leaving it unexplained: it looks like a pre-existing order-dependence in that file, worth its own issue, and unrelated to this PR.

## Remaining uncertainty

- `_FullscreenFeedContentState` has no `didChangeAppLifecycleState`, so I did not add one. The pointer self-heal covers the stuck case on both screens, so this is an asymmetry I chose rather than one the code forced.
- I did not add a test for the cross-item ownership guard, where one item's `dispose()` must not cancel another item's live hold. Deleting that guard leaves 330 tests green, so the coverage gap is real, but two items sharing one cubit with a disposal in between did not come out reliably in this harness and I would rather recommend it than ship a flaky test.
- The lost-pointer path is unreproduced. Platforms normally deliver a cancel on backgrounding. The self-heal makes it unreachable either way.

One open question on captions, left inline. Please review the takeover commits and merge when you are happy, or push back on any of them.

mobile/lib/widgets/video_feed_item/feed_videos.dart
hm21: Going with option 2 — the caption stays inside the fade. The peek exists to see the frame the UI covers, and the caption sits over that same frame, so leaving it up would re-cover exactly what the hold exposes. It is hidden only while actively held and returns the instant you release. Documented the reasoning in fc33c22. Thanks for flagging the deaf/HoH angle.

hm21:
Thanks for the takeover — measuring each behaviour fix on `main` vs the branch (and catching that the `repeated enter` guard was dead) is exactly the right depth. Reviewed all eight commits and re-tested the whole gesture on a physical Samsung Galaxy: hold-to-peek, holding directly on Report/More/Share, the fade-in tap window, and a second finger mid-hold all behave. Happy with the takeover as-is.

Captions: went with **option 2** — kept inside the fade and documented why in `fc33c22` (the peek reveals what the UI covers; the caption sits over that same frame, so leaving it up re-covers exactly what the hold exposes). Thread resolved.

Your three noted uncertainties are fine to leave for this PR: the pointer self-heal covers the unreproduced lost-pointer path and the fullscreen lifecycle asymmetry, and I'd rather not ship the cross-item ownership test flaky either.

Re-requesting your review — ready for approval whenever you're happy.

mbradley (approved):
Approving. The captions question is answered and the call was yours to make. `fc33c22` records the reasoning at the point where the wrapper applies, which is where the next person will look for it.

Verified on the delta since my last pass: comment only, six added lines, no code change, and the claim still holds at head (`subtitleLayer` sits inside `_FeedItemActions`, inside the `FeedImmersiveChrome` wrapper). The takeover commits are unmodified, with no force-push or rebase.

One note on CI, because the board currently misleads rather than informs.

`Mobile CI` is red while every job it gates on is green. It is an aggregator: one step that compares the results of `generated-files`, `format`, `analyze`, `tests` and `vgv-tag-gate`. It checks out nothing, runs no tests and builds nothing. On this head it has not executed that step at all. The first attempt died in `Set up job` with `Failed to resolve action download info. Error: Service Unavailable`. The second was cancelled with zero steps recorded, most likely because the workflow sets `cancel-in-progress` on a concurrency group keyed to the ref, so re-running that job on its own collides with itself.

All five inputs it reads are final and successful, so there is no outcome in which it surfaces something the other jobs did not. If it is a required check for merge, a fresh push or a full workflow re-run should clear it.

mbradley:
Correction on the CI note above: the green results are on attempt 2. I triggered a full re-run to clear the aggregate and attempt 3 hit the same Actions outage, so `Detect App CI Scope` was cancelled with zero steps and everything downstream skipped. My doing, not the code.

Nothing to diagnose. A re-run once Actions is healthy, or your next push, gives a clean board.

PR #6780, opened 2026-08-06
fix(feed): offer a Content Filters jump when adult content is off

### LINKED ISSUE #6781: fix(feed): offer a Content Filters jump when adult content is off
## Problem

When a viewer taps Verify age on an age-restricted feed video, they can complete adult-content age verification while their Content Filters still hide adult categories. In that state, the old snackbar only told them adult content was switched off and disappeared before they could act.

## Expected behavior

The app should give the verified viewer a direct path to Content Filters, where the adult content rows are already unlocked and can be changed intentionally.

## Scope

- Replace the dead-end snackbar with a prompt that offers to open Content Filters.
- Keep the prompt scoped like other feed bottom sheets so it covers the shell chrome and updates video overlay visibility.
- Cover the prompt copy, dismissal path, Content Filters navigation, and overlay state in tests.

## PR

Implemented in #6780.

hm21:
Closes #6781

## Description

When a viewer's Content Filters keep adult content hidden, tapping "Verify age" on an age-restricted video showed a snackbar that named the setting ("Settings → Content Filters") and then disappeared. The remedy was two screens away and the message was gone before the viewer could act on it.

Replaces that snackbar with a `VineBottomSheetPrompt` whose primary button pushes `/content-filters` directly.

**Why this path matters more than it looks:** it is the default landing spot for new viewers, not an edge case. `unlockAdultCategories()` deliberately does not promote nudity/sexual — accepting the in-feed age dialog verifies the viewer but leaves those categories at `hide`, so the very first "Verify age" tap returns `ViewerAuthBlockedByPreference` and lands here.

**Why a redirect and not an in-sheet toggle:** the "I confirm I am 18 years or older" checkbox in Content & Safety and the in-feed age dialog write the same `adult_content_verified` flag, and both `ViewerAuthBlockedByPreference` returns require it to be `true`. So the flag is already set whenever this sheet appears — the adult rows on the Content Filters screen are unlocked on arrival and the age-gate banner does not show. Flipping the categories from inside the feed was considered and skipped: the per-category choice stays where the viewer can see what they turned on.

`videoErrorAdultContentHidden` is split into `…HiddenTitle` / `…HiddenBody` / `…HiddenAction`, plus a shared `commonNotNow`, translated across all 22 locales (no English fallbacks).

The sheet is fired with `unawaited(...)` so the "Verify age" spinner clears as the sheet opens rather than when it closes.

**Related Issue:** 

## Out of Scope

- The sheet lands on the Content Filters screen without scrolling to or highlighting the adult section. The section is the first group on that screen, so no anchor was needed.
- `autoRetryAgeRestrictedPooledVideo` and the passive thumbnail auth path stay silent on this branch — unchanged.

## Verification

- `flutter test test/screens/feed/pooled_age_restricted_retry_test.dart` — 14/14, including tests that drive a real `GoRouter`: one taps "Open Content Filters" and asserts the route change, and one taps "Not now" and asserts the sheet dismisses with no navigation
- `flutter test test/l10n/arb_consistency_test.dart` — clean
- `dart format` — no changes
- Verified on a real Samsung Galaxy device.

## Type of Change


NotThatKindOfDrLiz:
@hm21 I pushed a small takeover commit (ec99307) to address the modal scoping/pause issues from review.

What changed:
- routes the adult-content-hidden prompt through showVideoPausingVineBottomSheet, so it uses the root navigator and updates OverlayVisibility.setBottomSheetOpen
- uses root-navigator pops for the prompt buttons
- opens Content Filters via pushWithVideoPause
- extends the retry tests to assert the body copy plus bottom-sheet/page overlay state
- expanded #6781 from the auto-created shell into a real issue body

Verification passed locally and in the pre-push hook:
- dart format lib test integration_test (0 changes)
- flutter analyze lib/screens/feed/pooled_age_restricted_retry.dart test/screens/feed/pooled_age_restricted_retry_test.dart
- flutter test test/screens/feed/pooled_age_restricted_retry_test.dart (14/14)
- pre-push analyzer and changed-file test also passed

GitHub did not enqueue normal PR checks for the new head automatically, so I manually dispatched Mobile CI for ec99307: https://github.com/divinevideo/divine-mobile/actions/runs/31126624444. It is currently queued behind the repo Actions backlog. Please review the takeover commit and merge when you are happy with it and CI has a green signal.

rabble (approved):
## Overview

Replaces the "adult content is switched off" snackbar with a `VineBottomSheetPrompt` that can actually take the viewer to Content Filters. The justification is right on the method:

> A snackbar was a dead end here: the remedy lives two taps deep in settings, and the message disappeared before the viewer could act on it.

That's the correct read — a transient message describing a multi-step remedy is worse than no message, because it tells the user something is fixable without giving them the fix.

Good details:

- **`unawaited(...)` with the reason stated**: the "Verify age" spinner clears as the sheet opens rather than when it closes, so the age-gate card behind the sheet isn't stuck loading. The test asserts `isVerifying(_videoId)` is false *while* the sheet is still up, which pins the actual benefit rather than the mechanism.
- **`pushWithVideoPause`** for the navigation, so the pause semantics carry through the push rather than resuming behind the settings screen.
- **`commonNotNow` is reused** rather than minting a fourth key for a word that already exists.
- **The l10n work is genuinely complete.** All three new keys land in all 22 locales with *real* translations (Amharic, Arabic, etc. — not English placeholders) and each carries an `@description` explaining the surface and the remedy. That's the standard the ARB gate is supposed to enforce and rarely gets.
- Tests cover both branches — sheet appears with the right copy and not the generic failure text, and the primary action reaches Content Filters — plus assert `isBottomSheetOpen`, so the pause integration is pinned too.

Approving. Three things.

---

### 1. `videoErrorAdultContentHidden` is now orphaned in all 22 locale files

The old snackbar key survives the change:

```
lib/l10n/app_pt.arb:4411:  "videoErrorAdultContentHidden": "O conteúdo adulto está desativado…"
lib/l10n/app_bg.arb:4542:  "videoErrorAdultContentHidden": "Съдържанието за възрастни е изключено…"
…and 20 more
```

There is no remaining `.dart` reference anywhere in `lib/` or `test/` — the only hits are the ARB files themselves. So it's 22 translated strings plus a generated getter per locale that nothing can reach, and its `@description` still describes the snackbar this PR deleted.

Worth removing in the same PR: dead l10n is the kind of debt that survives forever because nobody can tell by looking whether it's still wired up, and translators keep maintaining it on the next pass.

### 2. `Navigator.of(context, rootNavigator: true)` uses the *outer* context

```dart
onPrimaryPressed: () => Navigator.of(context, rootNavigator: true).pop(true),
onSecondaryPressed: () => Navigator.of(context, rootNavigator: true).pop(false),
```

This works today because `showVideoPausingVineBottomSheet` puts the sheet on the root navigator, so popping the root from the calling context lands on the sheet's route. But the coupling is invisible: nothing at this call site says "this only works if the sheet is root-presented".

That's not hypothetical — **#6788, in this same batch, adds `useRootNavigator: false` to `CommunitySuggestSheet`** for exactly the reason you'd want it here too (keeping the sheet inside the shell rather than over the tab bar). If that option is ever applied to this sheet, these two pops silently target the wrong navigator: `pop(true)` dismisses whatever is on the root stack instead of the sheet, and the `openFilters` future resolves from the wrong route.

Taking the builder's own `BuildContext` and popping that (no `rootNavigator` flag) makes it correct regardless of how the sheet is presented.

### 3. Fire-and-forget swallows failures and permits stacked sheets

`unawaited(_promptAdultContentHidden(context))` is right for the spinner timing, but it has two side effects worth deciding on deliberately:

- If anything inside throws — the push failing, a route guard rejecting `ContentFiltersScreen.path` — it becomes an unhandled async error rather than something the caller can log. A `.catchError` that logs to `_logName` would keep it diagnosable at no cost to the timing.
- `retryAgeRestrictedPooledVideo` now returns while the sheet is still open, so a viewer who taps "Verify age" again on another blocked video gets a second sheet stacked on the first. The pause flag is a boolean, so the inner dismissal also clears it while the outer sheet is still up (same concern I raised on #6788). Guarding on "a prompt is already showing" would close both.

rabble:

PR #6782, opened 2026-08-06
fix(video-metadata): frame the post preview like the feed

### LINKED ISSUE #6783: fix(video-metadata): frame the post preview like the feed
## Problem

The video metadata post preview did not frame clips the same way the feed frames the published post. The preview inscribed the clip in a target-aspect-ratio box, while the feed cover-fits non-square videos and contains square clips on the blurred poster backdrop. Metadata overlays could therefore be positioned over a different crop than users would see after publishing.

Raw draft previews opened from the upload-failure sheet were also vulnerable to source-ratio distortion: an uncropped source could be squeezed into the target-ratio surface before the render service had baked in the crop.

## Expected behavior

The metadata preview should mirror feed framing closely enough that captions, action buttons, and post metadata are previewed over the same visible video area users will see in the feed.

## Scope

- Use the same cover/contain decision as the feed.
- Use the decoded video ratio when sizing the native video surface, with the target ratio only as a pre-decode fallback.
- Keep square previews on the blurred poster backdrop.
- Preserve stop-motion parity with the rendered/exported crop.
- Cover the framing behavior with widget tests.

## Verification

Resolved by PR #6782.

hm21:
Closes #6783

## Description

The post preview inscribed the clip with `AspectRatio(clip.targetAspectRatio)` inside a rounded card, so it showed the whole frame letterboxed. The feed does the opposite: `VideoItemWidget` cover-fits every non-square video and crops its sides, and letterboxes square clips on a blurred poster backdrop. The metadata overlay was drawn full-bleed over that different crop, so the preview did not represent the published post.

The preview now mirrors the feed's framing:

- The fit decision reuses `videoCoversFeedViewport` - the same helper the feed uses for its backdrop gate - instead of a second copy of the square/non-square rule.
- Non-square clips cover-fit edge to edge through the same `FittedBox` construction `VideoItemWidget` uses. The native surface has no intrinsic size, so it is laid out at the decoded video ratio when dimensions are available, falling back to the target ratio before decode.
- Square clips stay contain-fit on `BlurredVideoBackdrop`. The backdrop gained an optional local poster path: a clip that is not published yet has a thumbnail file, never a URL or a blurhash. Reusing the widget keeps one implementation of the letterbox-band geometry instead of a preview-only copy that would drift from the feed.
- Stop-motion preview routes through `_FittedStopMotion` so square targets are constrained to the target-ratio box while the stills remain cover-cropped like the export.
- The stage's bottom corners are rounded with `VineTheme.shellCornerRadius`, the radius the feed uses via `NavRoundedShell` where the video meets the bottom nav. Preview-only mode (upload failure sheet) has no bottom chrome to seam into and stays edge to edge, mirroring the feed's `_MaybeRoundFeedBottom` branch.
- The system bottom inset is removed around the overlay. `VideoOverlayActions` offsets its caption and action column by `14 + viewPadding.bottom` because in the feed it spans the whole screen; here the stage already ends above the post bar's `SafeArea`, so that inset was counted twice and floated the action column about 34-48 px off the bottom edge. This is the same `MediaQuery.removePadding(removeBottom: true)` the fullscreen feed applies around its own `FeedVideos`.

## Out of Scope

The preview area sits above the post bar, so its viewport is slightly shorter than the feed's. The side crop is therefore marginally smaller than in the feed. Making it pixel-identical would mean moving the post bar on top of the video, which is a layout change beyond this fix.

## Verification

- `flutter analyze lib/screens/video_metadata lib/widgets/video_feed_item test/screens/video_metadata_preview_screen_test.dart test/widgets/video_feed_item/blurred_video_backdrop_test.dart` - clean
- `flutter test test/screens/video_metadata_preview_screen_test.dart test/widgets/video_feed_item/blurred_video_backdrop_test.dart` - green
- `flutter test test/widgets/stop_motion/stop_motion_player_test.dart` - green
- Pre-push hook - green, including changed-file analyze/tests
- Manually verified by the author on a real Samsung Galaxy device before takeover: 9:16 preview matches the feed crop, square clip letterboxes on the blurred poster, bottom corners round into the post bar, action buttons sit at the bottom edge

## Type of Change

## Review Follow-Ups

- **Stop-motion parity fix.** A square-target stop-motion clip was contain-fit across the whole non-square still, showing more frame than the published post. Stop-motion now routes through `_FittedStopMotion`, which cover-crops the stills and constrains the square case to the target-ratio box; the vertical case still cover-fills the viewport.
- **Rounding test gap.** Added a positive-path test that the post flow (`previewOnly: false`) rounds the stage's bottom corners.
- **Decoded-ratio sizing.** The fitted video surface now listens to the native controller's decoded aspect ratio and uses the target ratio only as a pre-decode fallback, preventing raw draft previews from squeezing uncropped sources into the target box.
- **Test debt cleanup.** The repeated `divine_video_player/player_0` channel mock is hoisted into shared setup, and preview tests now cover decoded-ratio sizing plus the square/vertical stop-motion wrapper branches.

### DIFF
--- a/mobile/lib/screens/video_metadata/video_metadata_preview_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_preview_screen.dart
@@ -9,6 +9,7 @@ import 'package:go_router/go_router.dart';
 import 'package:openvine/constants/video_editor_constants.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/models/divine_video_clip.dart';
+import 'package:openvine/models/stop_motion_clip_frame.dart';
 import 'package:openvine/providers/nostr_client_provider.dart';
 import 'package:openvine/providers/video_editor_provider.dart';
 import 'package:openvine/providers/video_publish_provider.dart';
@@ -187,12 +188,19 @@ class _PreviewStage extends StatelessWidget {
 
 /// Container widget that wraps the video player in a hero transition.
 ///
-/// Frames the clip exactly the way the feed will: non-square clips fill the
-/// preview edge to edge and crop like `VideoItemWidget`'s cover branch, square
-/// clips stay contain-fit on the blurred poster backdrop the feed paints in
-/// their letterbox bars. Without that mirroring the preview showed more of the
-/// frame than the feed does, so the metadata overlay sat over a different crop
-/// than the published post.
+/// Frames the clip to match the feed: non-square clips fill the preview edge to
+/// edge and crop like `VideoItemWidget`'s cover branch, square clips stay
+/// contain-fit on the blurred poster backdrop the feed paints in their
+/// letterbox bars. Without that mirroring the preview showed more of the frame
+/// than the feed does, so the metadata overlay sat over a different crop than
+/// the published post.
+///
+/// The video surface is sized from [DivineVideoClip.targetAspectRatio], not the
+/// live decoded ratio the feed uses. For a square target whose source is still
+/// uncropped (the crop is applied at publish) the video is fitted to the target
+/// box rather than center-cropped the way the feed shows the published file;
+/// matching that exactly would mean sizing from the decoded ratio like
+/// `video_metadata_cover_screen`.
 class _VideoPreviewContent extends StatelessWidget {
   /// Creates the video preview content wrapper.
   const _VideoPreviewContent({
@@ -230,13 +238,10 @@ class _VideoPreviewContent extends StatelessWidget {
               videoAspectRatio: aspectRatio,
             ),
           if (stopMotionFrames != null)
-            StopMotionPlayer(
+            _FittedStopMotion(
               frames: stopMotionFrames,
-              fit: fit,
-              cacheHeight:
-                  (MediaQuery.sizeOf(context).height *
-                          MediaQuery.devicePixelRatioOf(context))
-                      .round(),
+              aspectRatio: aspectRatio,
+              coversViewport: coversViewport,
             )
           else
             _FittedVideoSurface(
@@ -254,8 +259,10 @@ class _VideoPreviewContent extends StatelessWidget {
 /// Inscribes the player surface into the preview the way the feed does.
 ///
 /// The native surface has no intrinsic size and stretches to whatever box it
-/// gets, so the video is laid out at its own aspect ratio inside a
-/// [FittedBox] — the same construction `VideoItemWidget` uses in the feed.
+/// gets, so the video is laid out at [aspectRatio] inside a [FittedBox] — the
+/// same construction `VideoItemWidget` uses in the feed. Here [aspectRatio] is
+/// the clip's target ratio; see [_VideoPreviewContent] for the square-source
+/// caveat.
 class _FittedVideoSurface extends StatelessWidget {
   const _FittedVideoSurface({
     required this.clip,
@@ -283,6 +290,43 @@ class _FittedVideoSurface extends StatelessWidget {
           ),
         ),
       ),
+    );
+  }
+}
+
+/// Frames a stop-motion clip the way the feed and the export do.
+///
+/// Export center-crops the raw stills to the target box (`StopMotionFit.cover`
+/// in `StopMotionRenderService`) and the feed then fits that already-cropped
+/// mp4 per cover/contain. Reproduce both from the stills: [StopMotionPlayer]
+/// cover-crops them, and for the letterboxed (square) case they are constrained
+/// to the target-ratio box so the blurred backdrop fills the bars — instead of
+/// contain-fitting the whole non-square still, which showed more frame than the
+/// published post.
+class _FittedStopMotion extends StatelessWidget {
+  const _FittedStopMotion({
+    required this.frames,
+    required this.aspectRatio,
+    required this.coversViewport,
+  });
+
+  final List<StopMotionClipFrame> frames;
+  final double aspectRatio;
+  final bool coversViewport;
+
+  @override
+  Widget build(BuildContext context) {
+    // Default fit is BoxFit.cover — the centered crop the export bakes in.
+    final player = StopMotionPlayer(
+      frames: frames,
+      cacheHeight:
+          (MediaQuery.sizeOf(context).height *
+                  MediaQuery.devicePixelRatioOf(context))
+              .round(),
+    );
+    if (coversViewport) return player;
+    return Center(
+      child: AspectRatio(aspectRatio: aspectRatio, child: player),
     );
   }
 }
--- a/mobile/lib/screens/video_metadata/video_metadata_preview_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_preview_screen.dart
@@ -92,9 +92,7 @@ class _VideoMetadataPreviewScreenState
     _controller = DivineVideoPlayerController(useTexture: true);
     if (mounted) await _controller!.initialize();
     if (mounted) {
-      await _controller!.setSource(
-        VideoClip.file(await video.safeFilePath()),
-      );
+      await _controller!.setSource(VideoClip.file(await video.safeFilePath()));
     }
     if (mounted) await _controller!.setLooping(looping: true);
     if (mounted) await _controller!.play();
@@ -167,7 +165,7 @@ class _VideoMetadataPreviewScreenState
 /// behind them.
 ///
 /// The preview-only route has no bottom chrome to seam into and stays edge to
-/// edge — the same call the fullscreen feed makes in `_MaybeRoundFeedBottom`.
+/// edge — the same branch the fullscreen feed takes in `_MaybeRoundFeedBottom`.
 class _PreviewStage extends StatelessWidget {
   const _PreviewStage({required this.roundBottom, required this.child});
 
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz:
@hm21 I pushed a takeover follow-up on top of the PR branch.

What changed:

- `_FittedVideoSurface` now listens to the native player state and sizes the surface from the decoded video aspect ratio once dimensions are available, with the target ratio only as a pre-decode fallback. That fixes raw draft previews opened from the upload-failure sheet so uncropped sources are not squeezed into the target box.
- The stale caveat comment was replaced with the actual decoded-ratio behavior, and the `_PreviewStage` wording now says it mirrors the feed branch rather than literally calling the same widget.
- The repeated `divine_video_player/player_0` method-channel mock was hoisted into shared test setup.
- Added preview-screen tests for decoded-ratio sizing and the square/vertical stop-motion wrapper branches.
- Updated #6783 from the auto-created stub into a real issue body.

Verification:

- `flutter analyze lib/screens/video_metadata lib/widgets/video_feed_item test/screens/video_metadata_preview_screen_test.dart test/widgets/video_feed_item/blurred_video_backdrop_test.dart`
- `flutter test test/screens/video_metadata_preview_screen_test.dart test/widgets/video_feed_item/blurred_video_backdrop_test.dart`
- `flutter test test/widgets/stop_motion/stop_motion_player_test.dart`
- Pre-push hook passed changed-file analyze/tests before the force-with-lease push.

GitHub check suites for the new head SHA are currently queued with no check runs yet, so I am waiting on CI before marking this fully ready. Please review the takeover commit when you have a chance; once CI is green, merge when you are comfortable with the changes.

rabble (approved):
## Overview

Makes the pre-publish preview frame the clip the way the feed does: cover-crop for non-square clips, contain-fit on a blurred poster backdrop for square ones, bottom corners rounded to seam into the post bar the way the feed seams into the nav shell. `BlurredVideoBackdrop` grows a `filePath` source so an unpublished clip's local poster can drive it.

Plenty done right:

- **`MediaQuery.removePadding(removeBottom: true)`** with the reason on the line: the overlay offsets for the system inset because in the feed it spans the screen, but here the stage ends above the post bar's `SafeArea`, so the inset was being counted twice. That's a bug you find by looking at a device, and the comment points at the analogous removal the fullscreen feed already does.
- **`_FittedVideoSurface` uses `FittedBox` over a ratio-sized `SizedBox`** rather than `AspectRatio`, matching `VideoItemWidget`. The reason — the native surface has no intrinsic size and stretches to whatever box it gets — is exactly why `AspectRatio` alone doesn't work here.
- **`_FittedStopMotion` reasons from the export**, not from what looks right: `StopMotionRenderService` center-crops with `StopMotionFit.cover`, so the preview reproduces that from the stills instead of contain-fitting the raw non-square frame.
- **`_PreviewStage` skips rounding on the preview-only route** because there's no bottom chrome to seam into, mirroring `_MaybeRoundFeedBottom`.
- The `_BackdropContent` refactor keeps the `ImageFiltered` → `Opacity` nesting identical while making the source pluggable, so the blur behaviour is unchanged.
- `_VideoPreviewContent` drops from `ConsumerWidget` to `StatelessWidget` — it wasn't using `ref`.

---

### The main question: raw drafts still won't match the post

`coversViewport` / `fit` are decided from `clip.targetAspectRatio`, but the surface is sized from the **live decoded** ratio:

```dart
final videoAspectRatio = _videoAspectRatio > 0 ? _videoAspectRatio : widget.aspectRatio;
```

Consider a raw draft with a 16:9 source and a square target. `coversViewport` is false, so `fit` is `contain`, and the `FittedBox` contains a 16:9 box — the preview shows the **whole** 16:9 frame letterboxed, with the backdrop filling the bars. The published post will be center-cropped to square, so the user sees material in the preview that won't survive publishing.

That's the exact failure the class doc says this PR fixes:

> Without that mirroring the preview showed more of the frame than the feed does, so the metadata overlay sat over a different crop than the published post.

The doc does acknowledge the tradeoff — *"keeps raw draft previews from squeezing an uncropped source into the target-ratio box"* — so this is a deliberate choice, not an oversight. But it means the two halves disagree: `_FittedStopMotion` **does** constrain to the target-ratio box for the letterboxed case (correctly reproducing the export's crop), while `_FittedVideoSurface` deliberately doesn't.

For a screen whose entire job is "this is what you're about to post", I'd expect the target crop to win in both. If the raw-draft case genuinely can't show the target crop (because the render hasn't baked it and the preview can't reproduce it live), that's worth stating as the reason rather than "keeps it from squeezing" — and worth knowing how often a draft reaches this screen pre-render, since that's the population that gets a misleading preview.

Not blocking, but I'd like to understand which behaviour you actually want before this sets a precedent that the two media types frame differently.

---

### Smaller points

- **`stateStream.listen` has no `onError`.** If the controller's state stream ever emits an error, it becomes an unhandled async error rather than something the widget can degrade from. `onError: (_) {}` with a comment, or a fallback to `widget.aspectRatio`, would keep it contained.
- **`SizedBox(width: videoAspectRatio * 100, height: 100)`** — the `100` is an arbitrary unit that only exists to give `FittedBox` a ratio, and it reads as a real dimension. A named `const _ratioUnit = 100.0;` with one line of explanation costs nothing and stops a future reader from "fixing" it.
- **`BlurredVideoBackdrop` now takes two mutually-exclusive poster sources** as independent nullables, with `url` silently winning if both are set. The doc describes the precedence, but nothing enforces it. Given the widget already has `blurhash` as a third preferred source, a small sealed source type would make the ordering explicit — or at minimum an `assert(url == null || filePath == null)`.
- `_subscribeToController` assigns `_videoAspectRatio` outside `setState` in both `initState` and `didUpdateWidget`. Correct in both cases (a build follows either way), but it sits next to a `setState` call doing the same assignment, which invites a reader to think one of them is wrong.

rabble:

NotThatKindOfDrLiz:
## Code Review: fit(video-metadata)

### ✅ Strengths
- Feed/preview parity is the right design goal
- `videoCoversFeedViewport()` logic correctly replicates feed's `BoxFit` selection
- UI geometry (backdropAspectRatio, letterbox bands) properly isolates video sizing
- All 15 CI checks pass

---

### 🔴 Finding 1: Stretched Preview for Uncropped Sources (Correctness Issue)

**Location:** `_FittedVideoSurface` (lines 279–305)

**Problem:**
```dart
SizedBox(
  width: aspectRatio * 100,
  height: 100,
  child: ...
)
```
This uses a **fixed height (100)** with aspect-ratio-scaled width. For uncropped sources (e.g., from `_DraftPreview` passing `draft.clips.first` without the editor's crop), this stretches or squashes the preview non-uniformly.

**Doc comment claims:** "fitted to match the feed preview" — but it's only fitted if the source is already cropped to the target aspect ratio.

**Evidence:**
- `video_metadata_cover_screen.dart` (lines 459–478) solves this: decodes source dimensions, computes fitted aspect ratio with fallback, then sizes SizedBox proportionally
- Video editor normalizes aspect ratio *before* finalRenderedClip (line 658–672), ensuring render service outputs are always normalized
- Upload failure sheet (`upload_failure_sheet.dart` line 157) passes `draft.clips.first` (uncropped input), not the editor-rendered output

**Fix:** Replace fixed-100 height with:
```dart
final decodedAspectRatio = await _resolveSourceAspectRatio(imageUrl);
final fittedRatio = videoCoversFeedViewport(...) 
    ? decodedAspectRatio 
    : _targetAspectRatio;
SizedBox(
  width: fittedRatio * 200,
  height: 200,
  child: ...
)
```
or defer uncropped rendering to upload-completion (when editor output is available).

**Severity:** Medium. Affects UX for drafts loaded from disk; resolved before publish.

---

### 🟡 Finding 2: Test Coverage Gap — _FittedStopMotion (Test Debt)

**Location:** `_FittedStopMotion` widget (lines 306–332) + `video_metadata_preview_screen_test.dart`

**Problem:**
- _FittedStopMotion is the only video player variant in preview
- No automated coverage; regression was introduced (commit 1, square/portrait split) and fixed (commit 2), but gap persists
- If future changes break stop-motion sizing, gap is silent

**Fix:** Add widget test for square/portrait aspect ratio split:
```dart
testWidgets('_FittedStopMotion fits square clips with contain, portrait with cover', (tester) async {
  // setup: square clip → BoxFit.contain; portrait → BoxFit.cover
  // verify: videoCoversFeedViewport() returns expected bool per aspect ratio
});
```

**Severity:** Low. Deferred debt, but blocks regression.

---

### 🟡 Finding 3: Test Duplication — MethodChannel Setup (Code Debt)

**Location:** `video_metadata_preview_screen_test.dart` (all 4 tests)

**Problem:**
```dart
// Duplicated in every test (lines ~28–35, ~82–89, ~136–143, ~190–197)
MockMethodChannel(
  channel: MethodChannel('divine_video_player/player_0'),
  methods: {...},
)
```
~48 lines of duplication across four test stubs.

**Fix:** Hoist into setUp/tearDown:
```dart
late MockMethodChannel mockChannel;

setUp(() {
  mockChannel = MockMethodChannel(...);
});

tearDown(() => mockChannel.reset());
```

**Severity:** Nit. Reduces test maintenance burden.

---

### 💡 Finding 4: Comment Clarity — NavRoundedShell Pattern (Nit)

**Location:** `_PreviewStage` (lines 334–350)

**Observation:**
_PreviewStage hand-rolls `ClipRRect(shellCornerRadius)` instead of `NavRoundedShell`:
```dart
ClipRRect(
  borderRadius: BorderRadius.only(
    bottomLeft: Radius.circular(VineTheme.shellCornerRadius),
    bottomRight: Radius.circular(VineTheme.shellCornerRadius),
  ),
  child: ...
)
```

Comment says: "Avoid double fills by not nesting NavRoundedShell." This is correct (NavRoundedShell adds ColoredBox layers), but wording could clarify that avoiding nesting is intentional.

**Suggested wording:**
```dart
// Use ClipRRect directly (not NavRoundedShell) to avoid redundant color fills
```

**Severity:** Nit-level documentation.

---

### Summary
- **Request changes on:** Findings 1–3 (correctness + debt)
- **Nit-level:** Finding 4 (wording only)

All CI checks pass; ready to land after addressing 1–3.

PR #6784, opened 2026-08-06
fix(feed): hide the paused affordance during a temporary pause

### LINKED ISSUE #6785: fix(feed): hide the paused affordance during a temporary pause
Auto-created from PR #6784.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6784
**Branch:** `fix/paused-indicator-on-temporary-pause`

hm21:
Closes #6785

## Description

Opening the comments (or share) sheet from the feed pauses the video — expected — but the paused affordance (play button + playback toggles pill) stayed on screen behind the sheet. Since that pause resumes on its own the moment the sheet closes, the play button is misleading: there is nothing for the viewer to undo.

`PausedVideoOverlay.isVisible` was gated on the per-page `isActive` flag, which only means "this is the current feed page". Whether the feed may play at all lives in `isFeedActive` (`FeedVideos.isActive && _routeAllowsPlayback && appForeground`). A bottom sheet flips the feed inactive via `showVideoPausingVineBottomSheet` → `setBottomSheetOpen(true)` → `_syncFeedActive()`, so the player pauses while the page stays current — and the overlay had no way to tell that apart from a tap-to-pause.

`_Overlay` now also receives `isFeedActive` and the affordance is gated on `isActive && isFeedActive`. That is the same composition `DivineVideoMetricsTracker` already uses one line above, so the two stay consistent. It covers every pause the app undoes by itself: comments/share sheet, a pushed route over the feed, and app backgrounding. A real tap-to-pause leaves `isFeedActive` untouched and still shows the full affordance.

When the sheet closes, `PausedVideoOverlay.didUpdateWidget` resets its `_hasStartedPlaying` latch; if the video does stay paused for some reason, the existing 300 ms `_stablePauseTimer` re-promotes it and the affordance comes back.

**Related Issue:** 

## Out of Scope

None.

## Verification

- New regression test `activity wiring > hides the paused affordance while the feed itself is inactive` in `feed_videos_test.dart`; verified it fails without the production change.
- `flutter test test/widgets/video_feed_item/` (feed_videos, paused_video_overlay, community pause, repo swap) — green.
- `flutter test test/screens/feed/video_feed_page_test.dart test/screens/feed/pooled_fullscreen_video_feed_screen_test.dart` — 67 tests green.
- `dart format` + `flutter analyze lib/widgets/video_feed_item test/widgets/video_feed_item` — clean.
- Manually verified on a real Samsung Galaxy device: open the comment panel on a feed video → video pauses, no play indicator behind the sheet; close it → playback resumes without a flash. Tapping the video still shows the play button and the toggles pill.

## Type of Change

### DIFF
--- a/mobile/lib/widgets/video_feed_item/feed_videos.dart
+++ b/mobile/lib/widgets/video_feed_item/feed_videos.dart
@@ -495,6 +495,7 @@ class FeedVideosState extends ConsumerState<FeedVideos> with RouteAware {
               video: video,
               index: index,
               isActive: isActive,
+              isFeedActive: isFeedActive,
               contextTitle: widget.contextTitle,
               contentWarningRevealed: _revealedContentWarningVideoIds.contains(
                 video.id,
@@ -517,6 +518,7 @@ class _Overlay extends ConsumerStatefulWidget {
     required this.video,
     required this.index,
     required this.isActive,
+    required this.isFeedActive,
     required this.contentWarningRevealed,
     required this.onContentWarningRevealed,
     this.onToggleAutoAdvance,
@@ -527,7 +529,15 @@ class _Overlay extends ConsumerStatefulWidget {
   final DivineVideoPlayerController? controller;
   final VideoEvent video;
   final int index;
+
+  /// Whether this item is the current page of the feed.
   final bool isActive;
+
+  /// Whether the feed itself may play: the home tab is current, no route or
+  /// overlay covers it, and the app is foregrounded. A paused player while
+  /// this is `false` is a temporary pause that resumes on its own, so the
+  /// paused affordance stays hidden.
+  final bool isFeedActive;
 
   /// Whether the user already dismissed this video's content warning.
   /// Owned by [FeedVideosState] so the autoplay gate can read it too.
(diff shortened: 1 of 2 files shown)


rabble (approved):
## Overview

Threads `isFeedActive` down to `_Overlay` so `PausedVideoOverlay` only appears for a pause the user can actually undo. A comments/share sheet, a pushed route or a backgrounded app all pause the player and resume it on their own — showing a play button behind a sheet reads as broken, and tapping it fights the thing that paused it.

Small, correctly scoped, and the reasoning is on the line that needs it rather than in the PR description.

**The tests are the best part.** Two separate cases — route push/pop and app background/foreground — and the second one carries its own justification:

> Guards the `appForeground` factor specifically: the route test above stays green if a refactor drops it, this one turns red.

That's the right instinct. `isFeedActive` is a conjunction, and a single test only pins whichever term it happens to flip; splitting them means each factor has a test that can fail for its own reason. Both also assert the *return* to visible, so a change that just hardcodes `isVisible: false` doesn't pass.

Approving.

### One thing worth confirming, not blocking

The gate is `widget.isActive && widget.isFeedActive`, so the affordance is suppressed for the whole time the feed is inactive — including for a video the user had **explicitly** paused before the sheet opened. On close, `isFeedActive` goes true again and the affordance comes back, so nothing is lost.

The case I'd check by hand is the ordering on foreground: user taps to pause, backgrounds the app, foregrounds it. If the resume-on-foreground path restores playback unconditionally, the user's deliberate pause is silently discarded. That behaviour predates this PR and this change doesn't make it worse — the affordance correctly reappears either way — but it's the one interaction where "temporary pause" and "user pause" overlap, and it's now the only state the new comment doesn't describe.

### Nit

The doc comment on `isFeedActive` enumerates the three conditions (home tab current, nothing covering it, app foregrounded). That list will drift from whatever `isFeedActive` is actually computed from in `FeedVideosState`. Pointing at the source (`see FeedVideosState.isFeedActive`) instead of restating it would keep them from disagreeing later.

PR #6786, opened 2026-08-06
fix(video-detail): give the dead-end states the TV-static look

### LINKED ISSUE #6787: fix(video-detail): give the dead-end states the TV-static look
Auto-created from PR #6786.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6786
**Branch:** `fix/video-detail-tv-static-error-states`

hm21:
Closes #6787

## Description

The shared-link video screen (`/video/:id`) rendered its dead ends as a red warning icon on a flat background — it read like a crash report, not a state, and it was the first thing anyone opening a Divine link to a deleted video saw.

Those states now use the retro "no signal" takeover the camera permission screen already had: animated TV static, a sticker, title, and supporting copy.

- **Video not found** (also covers filtered and locally deleted videos): sticker + "Video not found" + new body copy. The copy deliberately covers both "deleted" and "we just can't fetch it from any connected relay", because the screen cannot tell those apart.
- **Failed to load**: same treatment plus a **Retry** button. Until now a transient network error was a dead end — the only recovery was leaving the screen and re-opening the link.
- **Still loading**: the bare spinner now carries the same top-left X, wired to the existing exit handler. A cold-start deep link parks on the first relay connection with no timeout, so without this an unresolved lookup could sit on the spinner with no way out.

The layout was extracted into `TvStaticMessageScreen` and the camera permission screen was moved onto it (–85 lines there). Rationale: two screens rendering the same takeover by hand drift apart on the first design tweak, and the camera gate keeps its specific chrome (dimmed recorder bottom bar, overlay style) through the widget's `footer` slot rather than by duplicating the whole stack.

The old `DiVineAppBar` on these states is gone; the close affordance is now the same top-left X the camera screen uses, carrying the existing `videoDetailCloseSemanticLabel`.

### Intended design adjustments

The shared `TvStaticMessageScreen` wraps its whole column in a single `SafeArea`. This is deliberate: it fixes the camera permission screen's content not being vertically centered on Android, where the system navigation bar was not previously accounted for. Two camera-screen deltas follow from the extraction and are intentional, not drift — the dimmed recorder bottom bar now sits inside the safe area, and the sticker→title gap is normalized to 20px.

## Out of Scope

Other error surfaces that still use the icon-on-flat-background pattern (e.g. the in-feed `PooledVideoErrorOverlay`) are untouched — those sit over a video frame, not on an empty screen, so the takeover treatment does not apply.

## Verification

- `flutter test test/screens/video_detail_screen_test.dart test/widgets/tv_static_message_screen_test.dart test/widgets/camera_permission_gate_test.dart` — 28 passing, including tests that the Retry button refetches, that the loading state can be dismissed mid-fetch, and a suite for `TvStaticMessageScreen`
- `flutter test test/l10n/arb_consistency_test.dart` — clean (2 new keys mirrored into all 21 non-English locales, `flutter gen-l10n` output committed)
- All four design-system ceiling ratchets (`raw_textstyle`, `raw_colors`, `material_button`, `raw_dialog`) — clean
- Manually verified on a real Samsung Galaxy: opened a link to a deleted video, opened a link while offline and used Retry to recover, dismissed a stalled cold-start loading state via the X, and re-checked the camera permission screen for regressions after the refactor

## Type of Change

- [x] 🧹 Code refactor

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -4537,7 +4537,7 @@
     "@videoErrorAdultContentHidden": {"description": "Snackbar shown when an age-verified viewer taps Verify age on an age-restricted video but their Content Filters keep adult content hidden. The remedy is opting in via Settings → Content Filters, not re-verifying age."},
     "videoDetailLoadError": "ቪዲዮ መጫን አልተሳካም",
     "videoDetailLoadErrorBody": "እዚህ ሲደርስ የሆነ ችግር ተፈጥሯል። እንደገና ሞክር።",
-    "videoDetailNotFoundBody": "ወይ ተሰርዟል፣ ወይም ልንደርስበት በማንችልበት ቦታ ተደብቋል።",
+    "videoDetailNotFoundBody": "ተሰርዞ ሊሆን ይችላል፣ ከአቅም ውጭ ሊሆን ይችላል፣ ወይም በቅንብሮችህ ተደብቆ ሊሆን ይችላል።",
     "databaseCorruptionTitle": "የአካባቢዎ ውሂብ ተበላሽቷል",
     "databaseCorruptionBody": "Divine ን ዘግተው እንደገና ይክፈቱት — በራስ-ሰር እናስተካክለዋለን። የቻልነውን ያህል ረቂቆችዎን እና ክሊፖችዎን እናስቀራለን፣ የቀረው እንደገና ይጫናል።",
     "databaseCorruptionCloseButton": "Divine ን ዝጋ",
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -4536,6 +4536,8 @@
     "videoErrorAdultContentHidden": "የአዋቂዎች ይዘት ጠፍቷል። በቅንብሮች → የይዘት ማጣሪያዎች ውስጥ ማብራት ይችላሉ።",
     "@videoErrorAdultContentHidden": {"description": "Snackbar shown when an age-verified viewer taps Verify age on an age-restricted video but their Content Filters keep adult content hidden. The remedy is opting in via Settings → Content Filters, not re-verifying age."},
     "videoDetailLoadError": "ቪዲዮ መጫን አልተሳካም",
+    "videoDetailLoadErrorBody": "እዚህ ሲደርስ የሆነ ችግር ተፈጥሯል። እንደገና ሞክር።",
+    "videoDetailNotFoundBody": "ወይ ተሰርዟል፣ ወይም ልንደርስበት በማንችልበት ቦታ ተደብቋል።",
     "databaseCorruptionTitle": "የአካባቢዎ ውሂብ ተበላሽቷል",
     "databaseCorruptionBody": "Divine ን ዘግተው እንደገና ይክፈቱት — በራስ-ሰር እናስተካክለዋለን። የቻልነውን ያህል ረቂቆችዎን እና ክሊፖችዎን እናስቀራለን፣ የቀረው እንደገና ይጫናል።",
     "databaseCorruptionCloseButton": "Divine ን ዝጋ",
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -4412,7 +4412,7 @@
     "@videoErrorAdultContentHidden": {"description": "Snackbar shown when an age-verified viewer taps Verify age on an age-restricted video but their Content Filters keep adult content hidden. The remedy is opting in via Settings → Content Filters, not re-verifying age."},
     "videoDetailLoadError": "فشل تحميل الفيديو",
     "videoDetailLoadErrorBody": "حدث خطأ ما في الطريق. جرّب مرة أخرى.",
-    "videoDetailNotFoundBody": "إمّا أنه حُذف أو أنه مخبّأ في مكان لا نصل إليه.",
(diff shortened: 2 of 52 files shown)


mbradley (changes requested):
Took this over and pushed six commits. The extraction itself is right, and the two behaviour additions — Retry, and an exit from the loading state — were both worth making; what I found is that each of them opened a smaller hole next to the one it closed.

**Retry could strand you worse than the error did.** `_retry` kept the relay-ready deferral enabled, so if the relays had dropped since the error appeared, the retry handed off to the relay listener and returned without clearing `_isLoading`. The error screen and its Retry button were replaced by an unbounded spinner. A retry the user asked for now always ends somewhere definite; they can tap again once they're back online.

**The close button on the loading state made a dead window live.** A popped route's State outlives the exit transition, so a relay connecting in that gap restarted the lookup for a screen already left behind. Cancelled on exit. The test fires the relay event mid-transition, which is the only ordering where the two behaviours differ — settle the transition first and dispose masks it.

**The SafeArea change costs a strip of live static under the recorder bar.** `VideoRecorderBottomBar` carries its own `SafeArea`, so wrapping the whole column moved the inset from inside the dimmed bar's coloured container to outside it. On an 800px viewport with a 48px inset the band painted to 752 instead of 800, with the animated noise showing through the gap. I scoped `SafeArea`'s bottom to the footerless case, which restores it and keeps inset-aware centring on the video-detail screens.

On that note — I couldn't reproduce the Android centring problem the description cites. Every content element measures pixel-identical with and without the outer `SafeArea`, because the footer's own `SafeArea` was already absorbing that inset. If you were looking at something my measurements don't model, say so and I'll re-check; otherwise that part of the rationale doesn't hold up.

Also restored the status-bar overlay style the old `DiVineAppBar` was setting and these screens quietly lost, pulled the top-left X into one `TakeoverCloseButton` (it was hand-built in three places, on a change whose point was consolidating this look), labelled the camera gate's X with the existing `videoRecorderCloseLabel`, and pinned the not-found body copy — dropping it left every test green.

**One thing I did not fix, because it's yours to call.** "It was either deleted, or it's hiding somewhere we can't reach" fronts a branch with three causes, and the third is the viewer's own settings. `DivineHostFilterService` defaults to `true`, so on a fresh install a shared link to any non-Divine-hosted video reads as possibly-deleted when it exists and is one Safety toggle away. Same for an author the viewer blocked, or muted from another Nostr client — we have no mute action, so they have no memory of doing it. The old bare title was vague enough to stay true; the new sentence commits to a cause the branch can't support.

Worth saying what shouldn't change: the message being identical across blocked, muted and never-existed is protective — it means a blocked viewer can't confirm an account exists or walk IDs to map someone's content. Adding Retry only to `loadFailed` and not here is right for the same reason. The seam is that the viewer's own settings leak nothing, since it's their toggle.

Three options, and I'd take the first: branch just the host-filter clause to a settings-pointing message, modelled on `videoErrorAdultContentHidden`; or soften the body so it stays true across all three; or keep it and accept the tradeoff. I fixed the ARB description either way — it documented two causes for a three-cause branch.

**Validation:** full app suite via `mise run test` — 14,853 passed, 285 skipped, 0 failed. `flutter analyze lib test integration_test` clean, `dart format` clean across 2,891 files, `arb_consistency_test` clean, `flutter gen-l10n` reproduces committed output. Every new test mutation-checked: I broke the thing it guards and watched it go red. One first-draft test passed without its fix and I rewrote it rather than bank it.

No rebase and no force-push — your three commits are untouched and mine sit on top. Have a look and merge when you're happy, or push back on any of it.

Thanks, as always.

hm21:
Thanks Matthew — took all six. The retry-strand, the dead-window relay restart, and the static gap under the recorder bar were all real, and the overlay-style + `TakeoverCloseButton` consolidation is a clear win. Agreed on every one.

On the not-found copy I went with your option 2 (soften) rather than branching the host-filter clause. The screen only knows `shouldHideVideo` returned true, not whether blocklist or the host-filter tripped it — so a settings-pointing message would misfire whenever the real cause was a block/mute. Option 1 ends up over-claiming the same way the current body does, just in the other direction.

New body:

> It could be gone, out of reach, or hidden by your settings.

Lists all three causes without committing to one, stays identical across deleted / unreachable / filtered (so a blocked viewer still can't confirm an account exists), and I kept your three-cause ARB description as-is. Translated into all 21 non-English locales, `gen-l10n` regenerated, sitting on top of your commits — no rebase.

Ready for another look when you have a minute.

mbradley:
approval's coming. just waiting for actions to settle so i can confirm green ci first.

hm21:
@mbradley CI is green now, so it would be ready to approve if it's still fine for you.

mbradley (approved):
Re-checked the new head rather than taking the summary on trust. All 22 locales carry the new string, none left on the old copy or in English, `gen-l10n` reproduces the committed output with no drift, and the four affected suites pass — 42 tests — with `flutter analyze lib test` clean. CI is green now too, and genuinely so: all ten jobs ran and passed, no skips.

The softened copy does the job. It names all three causes without committing to one, and it stays byte-identical across deleted, unreachable and filtered — so the property worth protecting still holds: a blocked viewer can't use this screen to confirm an account exists.

One correction for the record, not a request. The screen *can* distinguish the host-filter case — `shouldFilterNonDivineVideos` is a public getter on `VideoEventService` and `isFromDivineServer` a public extension on `VideoEvent`, so `shouldFilterNonDivineVideos && !video.isFromDivineServer` is separable from the blocklist branch today. Option 1 was reachable. Your call still stands on its own merits, and I'd rather leave the note than have the next person read the thread and conclude the distinction is impossible — if the settings-pointing message ever looks worth it, it's a smaller job than the comment suggests.

Thanks, as always.

PR #6788, opened 2026-08-06
fix(feed): pause playback for the mode and classify sheets

### LINKED ISSUE #6789: fix(feed): pause playback for the mode and classify sheets
Auto-created from PR #6788.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6788
**Branch:** `fix/pause-playback-on-feed-mode-and-classify-sheets`

hm21:
Closes #6789

## Description

Tapping the feed-mode picker ("For You ▾") or the "Help classify this" action left the video behind the sheet playing. The comments and share sheets pause it.

Both opened `VineBottomSheetSelectionMenu.show` / `VineBottomSheet.show` directly, bypassing the `PauseAwareModals` extension. That extension is what flips `overlayVisibilityProvider.isBottomSheetOpen`, which `VideoFeedPage` listens to in order to deactivate the feed — without it nothing pauses.

Both now route through `PauseAwareModals`, using `setBottomSheetOpen` (not `setPageOpen`), so the current player is paused but retained for instant resume — identical to comments and share.

Two supporting changes in `pause_aware_modals.dart`:

- New `showVideoPausingSelectionMenu`. `VineBottomSheetSelectionMenu.show` exposes no `onShow`/`onDismiss` hooks, so the flag is toggled around the returned future rather than adding parameters to `divine_ui` for one call site.
- `showDragHandle` passthrough on `showVideoPausingVineBottomSheet`, which the classify sheet needs (`showDragHandle: false`).

The classify sheet passes `useRootNavigator: false` explicitly. The extension defaults to `true`, which would newly make the sheet cover the tab bar — a visual change unrelated to the pause fix, so the current presentation is preserved.

**Related Issue:** 

## Out of Scope

Whether the classify sheet *should* cover the tab bar like the comments sheet does. That is a design call, not part of this fix.

The other `VineBottomSheetSelectionMenu` call sites (library, search filters, recorder, metadata expiration) are not over a playing feed video and are left alone.

## Verification

- `flutter test test/utils/pause_aware_modals_test.dart test/screens/feed/feed_mode_switch_test.dart test/widgets/video_feed_item/actions/help_classify_action_button_test.dart test/widgets/community_suggest/community_suggest_sheet_test.dart` — 33 passed
- Both new call-site regression tests were confirmed to fail against the pre-fix code (`+23 -2`) and pass after, so they can catch the regression coming back.
- Manually verified on a real Samsung Galaxy device: opening the mode picker and the classify sheet stops playback, closing either resumes it from where it was.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/screens/feed/feed_mode_switch.dart
+++ b/mobile/lib/screens/feed/feed_mode_switch.dart
@@ -9,6 +9,7 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:openvine/blocs/video_feed/video_feed_bloc.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/screens/feed/feed_settings_menu.dart';
+import 'package:openvine/utils/pause_aware_modals.dart';
 import 'package:openvine/widgets/video_feed_item/feed_immersive_chrome.dart';
 
 /// Feed mode picker overlay that displays the current feed mode
@@ -80,8 +81,7 @@ class FeedModeSwitch extends StatelessWidget {
     VideoFeedBlocState state,
   ) async {
     final l10n = context.l10n;
-    final selected = await VineBottomSheetSelectionMenu.show(
-      context: context,
+    final selected = await context.showVideoPausingSelectionMenu(
       selectedValue: state.source.persistenceValue,
       options: [
         VineBottomSheetSelectionOptionData(
--- a/mobile/lib/utils/pause_aware_modals.dart
+++ b/mobile/lib/utils/pause_aware_modals.dart
@@ -110,6 +110,7 @@ extension PauseAwareModals on BuildContext {
     Widget? bottomInput,
     bool expanded = true,
     bool showHeaderDivider = true,
+    bool showDragHandle = true,
     bool? isScrollControlled,
     double initialChildSize = 0.6,
     double minChildSize = 0.3,
@@ -164,6 +165,7 @@ extension PauseAwareModals on BuildContext {
       bottomInput: bottomInput,
       expanded: expanded,
       showHeaderDivider: showHeaderDivider,
+      showDragHandle: showDragHandle,
       isScrollControlled: isScrollControlled,
       initialChildSize: initialChildSize,
       minChildSize: minChildSize,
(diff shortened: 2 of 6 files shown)


rabble (approved):
## Overview

Routes two sheets that had bypassed the pause integration — the feed-mode picker and the "Help classify this" suggestion sheet — through the pause-aware wrappers, and adds `showVideoPausingSelectionMenu` for the selection-menu case that `VineBottomSheetSelectionMenu.show` can't serve (no `onShow`/`onDismiss` hooks, so the flag has to be toggled around the returned future).

`.whenComplete()` is the right choice over `.then()` — it fires on scrim dismissal and on error, not only on a selection — and there's a dedicated test that pins exactly that path rather than assuming it. Coverage is complete for what changed: the extension itself, the mode sheet's select path, the mode sheet's cancel path, and the classify sheet's pop path all assert `isBottomSheetOpen` both ways. `setSurfaceSize` is paired with an `addTearDown` reset, per the process-global rule, so it won't strand later suites in the merged isolate.

The `useRootNavigator: false` comment on `CommunitySuggestSheet` is the kind of note worth having — "preserves the pre-pause-integration presentation: the sheet stays in the shell navigator rather than covering the tab bar" tells the next reader it's deliberate rather than an oversight.

Approving. Two things worth a look, neither blocking.

---

### `isBottomSheetOpen` looks like a flag rather than a count

`setBottomSheetOpen(true)` / `setBottomSheetOpen(false)` reads as a boolean. If two pause-aware modals can ever overlap — a selection menu opened from within a sheet, or a sheet opened while another is animating out — the inner one's `whenComplete` clears the flag while the outer is still on screen, and the video behind it resumes.

This isn't introduced here (`showVideoPausingVineBottomSheet` has the same shape via `onShow`/`onDismiss`), and it may well be unreachable in practice. But this PR increases the number of entry points, which raises the odds. Worth confirming whether the notifier refcounts; if it doesn't, a counter would make the invariant hold regardless of nesting.

### Does the shell-navigator sheet always complete?

With `useRootNavigator: false`, the classify sheet lives in the shell navigator. The flag is cleared by the route's future completing. If a tab switch or a shell-route replacement can tear that navigator's route down without completing the future, the flag stays `true` and playback stays paused with no sheet visible — a stuck state the user can't clear except by backgrounding.

The existing test pops the route explicitly, so it wouldn't catch this. Worth trying by hand: open the classify sheet, switch tabs, come back, and check the video resumes.

### Nit

`showVideoPausingSelectionMenu` returns `Future<String?>` while its sibling `showVideoPausingVineBottomSheet` is generic (`<T>`). If `VineBottomSheetSelectionOptionData.value` is always a `String` that's fine, but the asymmetry will read as an oversight next time someone adds a menu with a typed value.

PR #6790, opened 2026-08-06
fix(metadata): hold user chip rows open with placeholders

### LINKED ISSUE #6791: fix(metadata): hold user chip rows open with placeholders
Tracks PR #6790.

The metadata sheet user-chip sections should not drop into the sheet after async profile/reposter resolution and push lower sections down.

Expected behavior:
- Reposted-by reserves space while relay reposters and profile names resolve.
- Collaborators reserves space while tagged collaborator names resolve.
- Popular repost rows reserve the future +N more action as well as the visible user chips.
- Empty, stale, or malformed repost counts do not leave an empty Reposted by section behind.

Fixed by #6790.

hm21:
Closes #6791

## Description

The Reposted by and Collaborators sections in the metadata sheet now reserve chip-row space while async profile and relay data resolves, instead of rendering nothing and dropping into the sheet later.

- Reposted by uses the live Nostr repost count while the relay query is in flight, then switches to the fetched/pre-populated pubkey count while profile names resolve.
- Collaborators uses the tagged collaborator list, which is known synchronously.
- Popular repost rows reserve one extra placeholder for the eventual +N more action, so the resolved row does not grow when the action appears.
- Non-positive or stale repost counts stay hidden or collapse through the existing AnimatedReveal path.

Archival Vine reposts remain excluded because they have no pubkey and never become user chips.

## Out of Scope

Creator and Inspired by render single chips without this reveal gate. Broader skeleton consolidation into divine_ui is left out of this focused bug fix.

## Verification

- flutter test test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart — 66 tests passing.
- flutter analyze lib test integration_test — no issues found.
- Pre-push hook reran analyzer and the changed metadata widget test before pushing 96a0e1f05.

## Type of Change

### DIFF
--- a/mobile/lib/widgets/video_feed_item/metadata/metadata_user_chips.dart
+++ b/mobile/lib/widgets/video_feed_item/metadata/metadata_user_chips.dart
@@ -3,6 +3,7 @@
 // ABOUTME: using tappable chips that navigate to user profiles.
 
 import 'dart:async';
+import 'dart:math' as math;
 
 import 'package:collaborator_repository/collaborator_repository.dart';
 import 'package:divine_ui/divine_ui.dart';
@@ -22,6 +23,7 @@ import 'package:openvine/utils/public_identifier_normalizer.dart';
 import 'package:openvine/widgets/user_avatar.dart';
 import 'package:openvine/widgets/video_feed_item/metadata/metadata_section.dart';
 import 'package:openvine/widgets/video_feed_item/metadata/video_reposters_cubit.dart';
+import 'package:skeletonizer/skeletonizer.dart';
 
 /// Whether every chip in [pubkeys] already knows the name it will render.
 ///
@@ -62,7 +64,8 @@ class MetadataCreatorSection extends StatelessWidget {
 /// `ignored` for this video. Pending collaborator chips are dimmed when
 /// the current user is the video's creator.
 ///
-/// Renders nothing while the resulting list is empty.
+/// Shows placeholder chips while the names of the resulting list resolve,
+/// and renders nothing while that list is empty.
 class MetadataCollaboratorsSection extends ConsumerWidget {
   const MetadataCollaboratorsSection({required this.video, super.key});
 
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz:
@hm21 I took over the two review fixes and pushed 96a0e1f05 to this PR branch.

What changed:
- Reposted-by placeholders now reserve the eventual +N more action too, both from a large nostrRepostCount while loading and from fetched/pre-populated pubkeys over the visible cap while names resolve.
- Non-positive promised repost counts now stay hidden instead of rendering an empty Reposted by section.
- Added widget coverage for the large-count placeholder, fetched-over-cap placeholder, and negative-count cases.
- Refreshed the PR body and #6791 so the behavior and verification are accurate.

Verification passed locally and in the pre-push hook:
- flutter test test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart — 66 tests passing.
- flutter analyze lib test integration_test — no issues found.
- Pre-push reran analyzer and the changed metadata widget test before accepting the push.

GitHub did not automatically enqueue PR checks for the new head, so I manually dispatched Mobile CI on 96a0e1f05: https://github.com/divinevideo/divine-mobile/actions/runs/31126598751. It is still queued behind other Mobile CI runs and has not acquired a runner yet. Please review the takeover commit and merge when you are comfortable and CI is green.

rabble (approved):
## Overview

Holds the collaborator and reposter chip rows open with correctly-counted placeholder pills while names resolve, instead of letting the section appear out of nowhere and shove everything below it down a round-trip later.

The counting is the interesting part and it's reasoned through properly:

- Collaborators: the tagged list is known synchronously, so `visible.length` is exact.
- Reposters: once the relay has named anyone, `candidate.length` is exact and only labels are pending; before then the feed's `nostrRepostCount` stands in — **excluding archival Vine reposts, because those carry no pubkey and never become a chip**. That distinction is easy to get wrong (`embedded_*` vs live counts are a standing trap in this codebase) and the comment calls it out explicitly.
- `+ 1` when the row will carry a "+N more" chip, so the reserved height includes it.
- Drops to zero when the relay answers with nobody, "so a stale feed count cannot leave a shimmer on screen forever" — that's the failure mode a promised-count placeholder invites, and it's closed.

Design-system compliance is clean: `context.vineColors.skeleton`, `vineSkeletonEffectOf(context)`, `Skeleton.leaf`, small private widget classes rather than `Widget`-returning helpers, no raw colours or inline `TextStyle`. `Semantics(label: context.l10n.commonLoading)` on the row is a nice touch that's usually skipped on skeletons.

The `_chipWidths` comment is honest in a way I appreciate: *"No width can be truthful here — the names that decide them are exactly what is still missing."*

Approving. Two things worth addressing.

---

### 1. The skeleton's fixed `height: 40` doesn't survive text scaling

`_UserChipSkeleton` hardcodes a 40px box, documented as "matching `_TappableUserChip`'s 40 px height". But the real chip has no fixed height — it's `padding: EdgeInsets.symmetric(horizontal: 16, vertical: 8)` wrapped around an avatar and a label, so its height is content-driven and **grows with the system text scale**.

At default scale the two agree and the PR works exactly as intended. At a large accessibility font the real chips are meaningfully taller than 40, so the placeholder under-reserves and the row still jumps when names arrive — the layout shift comes back precisely for the users who are least able to tolerate it, and it will look like the feature simply doesn't work rather than like a sizing bug.

Options, roughly in order of effort:
- Build the skeleton from the same padding + a scaled text metric instead of a literal, so both sides move together.
- Or give the skeleton the same `EdgeInsets.symmetric(vertical: 8)` around a `SizedBox` sized off `MediaQuery.textScalerOf(context).scale(...)`.

Either way it stops being a number that has to be manually kept in sync.

### 2. `40` and `16` are duplicated from the widget they're imitating

The radius (`BorderRadius.circular(16)`) appears in both `_TappableUserChip` and `_UserChipSkeleton`, and the height comment asserts a match that nothing enforces. A restyle of the real chip leaves the skeleton silently wrong, and the only signal is a layout jump nobody is testing for.

Hoisting them to `static const` on `_TappableUserChip` (or a small shared `_ChipMetrics`) and referencing them from the skeleton makes the relationship real rather than documented. That also makes fix #1 a one-place change.

---

### Nits

- `Semantics(label: ...)` without `excludeSemantics: true` merges with rather than replaces the subtree's semantics. `Skeletonizer` children are decorative boxes so there's probably nothing underneath to conflict, but if any placeholder ever gains a label the row will announce both.
- `_placeholderChipCount` returns `0` for "nothing to show" and the caller checks `<= 0`. It can't return a negative, so `== 0` would say the same thing with less room for doubt about whether a negative case exists.

NotThatKindOfDrLiz:
## Summary

**Finding 1: Extract `hasMoreReposters` condition** - FIXED ✓

Extracted the inline ternary condition in `_placeholderChipCount()` into a named variable `hasMoreReposters` for improved code clarity. This makes the logic more maintainable and easier to reason about.

**Commit:** 14d790adf5

**CI Status:** All checks passing ✓
- Analyze: pass
- Format: pass
- Generated Files: pass
- Tests (all shards): pass
- Mobile CI: pass

---

**Finding 2: Extract shared skeleton chip pattern** - Deferred

The `_UserChipSkeleton` widget could be extracted into a reusable component in `divine_ui` if skeleton patterns appear in other places. This is deferred to a follow-up optimization pass and does not block shipping this fix.

---

The refactoring is ready for review and all CI checks are green.

PR #5324, opened 2026-06-18
chore: remove orphaned upload/explore/badge widgets (#4803)

realmeylisdev:
## Description

While migrating raw `Icons.*` for #4803 (under maintainability epic #4339), I found that several of the remaining raw-icon callsites live in widgets that are **defined but never instantiated anywhere in `lib`** — orphaned code, most of it stranded by the recent explore-tab refactor (#5297). Their icons couldn't be design- or device-verified because the widgets don't render in the app. Rather than migrate icons in dead code, this PR removes the dead code: **~2.2k lines of Dart deleted, and 23 raw `Icons.*` drop off the ratchet for free (baseline 221 → 198).**

### Removed (each confirmed: no `lib` importer, no barrel/`export`, no route, no string/deep-link reference)

| Widget(s) | File | Why dead |
| --- | --- | --- |
| `GlobalUploadIndicator` | `global_upload_indicator.dart` | Never mounted anywhere |
| `UploadProgressIndicator`, `CompactUploadProgress` | `upload_progress_indicator.dart` | Consumed only by `GlobalUploadIndicator` + a `skip:true` golden test |
| `VideoExploreTile` | `video_explore_tile.dart` | Orphaned by the explore-tab refactor (#5297) |
| `ProofModeBadgeRow` | `proofmode_badge_row.dart` | Consumed only by `VideoExploreTile` |
| `BadgeExplanationModal` | `badge_explanation_modal.dart` | Never shown; only its own test referenced it |

Plus their 4 dedicated test files and 4 orphaned golden PNGs.

### Kept (shared / live — untouched)

- **`proofmode_badge.dart`** (`ProofModeBadge`, `OriginalContentBadge`, `OriginalVineBadge`, …) — still used by the **live** video-feed metadata badge rows via `proofmode_helpers.dart` → `metadata_badges_row.dart` / `metadata_expanded_sheet.dart`.
- `special_profile_checkmark.dart`, `video_thumbnail_widget.dart` — used by `user_name`, `video_feed_item`, `message_bubble`, search/grid widgets.
- `upload_manager.dart` service.

`original_content_badge_test.dart` is **rewritten** (not deleted): it dropped the deleted `ProofModeBadgeRow` host, but kept its still-valid direct `OriginalContentBadge` render test and the `VideoEvent` extension tests (`isOriginalContent` / `isRepost` / `shouldShowNotDivineBadge` / `isOriginalVine`).

## 📱 How to test (regression smoke test on a real device)

[...]

PR #6792, opened 2026-08-06
feat(metadata): explain the four verification checks

### LINKED ISSUE #6793: feat(metadata): explain the four verification checks
Auto-created from PR #6792.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6792
**Branch:** `feat/verification-checks-explainer`

hm21:
Closes #6793

## Description

The Verification section in the video metadata sheet lists four checks — Device attestation, PGP signature, C2PA Content Credentials, Proof manifest — and explains none of them. To a viewer that is four ticks of domain jargon, which reads as decoration rather than evidence.

Nothing in the app filled that gap. `badge_explanation_modal.dart` used to, but it was already unreachable when cc44a154a deleted it as orphaned (#5324), and the new metadata sheet never picked the explanation back up. What is left is a Support Center tile and the recorder's upload explainer — both link out to divine.video/proofmode, and neither is somewhere a viewer of a video ever goes.

Tapping the section header now opens a bottom sheet with one line per check, saying what that check is evidence of.

**Why an info icon on the header and not a tap handler on the four rows:** the sheet explains all four at once, so it wants one entry point, not four. A checklist row carries no affordance that it is interactive, and the rows are ~20 dp tall with paddings that the sheet's own code comments mark as load-bearing against the Figma spec. The icon also lands in the same column as the check/X marks, so the header reads as a set with the rows below it.

**Why the caveat paragraph is not optional:** `hasUsableX` collapses "absent" and "present but invalid" into the same `false`, so today's muted X means *not proven*, never *failed*. Without an explicit "a missing check doesn't make a video fake", four ticks plus an X invites the reader to infer the stronger claim. The copy deliberately hedges; please don't strengthen it in review.

The learn-more link reuses the existing `proofmodeLearnMoreUrl` constant rather than hardcoding a third copy of the URL.

**Related Issue:** 

## Out of Scope

Surfacing a genuinely *failed* verification. `ProofVerificationSummary` already distinguishes `verified` / `present` / `invalid` and carries a `*_valid` flag per check, but the UI has no third state — and when `status == 'invalid'` with no raw tags, `hasProofMode` is false and the whole section disappears, so a video whose signature failed looks exactly like one that never claimed anything. That needs its own state, copy and colour treatment; this PR only closes the explanation gap.

Also untouched: the unused `identity_binding` / `identity_portable` / `identity_verifier` signals the publisher already writes.

## Verification

- `flutter analyze lib test integration_test` — clean locally before push and in the pre-push hook
- `flutter test test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart test/l10n/arb_consistency_test.dart` — 71 passing, including the new learn-more launch regression test
- `mise exec -- dart format lib test integration_test` — clean
- All four design-system ceiling ratchets (raw TextStyle / raw colors / Material buttons / raw dialogs) — clean
- New keys translated into all 21 non-English locales
- Removed the 14 orphaned `proofmode*` strings left by #5324 from all locale sources and regenerated l10n output
- Manually verified on a real Samsung Galaxy (SM-S942B): header tap target, explainer copy, and the learn-more link out to divine.video/proofmode

## Review follow-up

The second commit applies review findings on the first, all confirmed against the rendered semantics tree rather than by inspection:

- The header node carried `label` and `tooltip` set to the same string. Flutter appends `tooltip` to the iOS accessibility label and exposes it as `setTooltipText` on Android, so both screen readers said it twice.
- The node's label was only the explainer question, so the checklist was announced with no section name — the visible "Verification" text is deliberately excluded from semantics. It now names both.
- Excluding the info icon's `isImage` flag removes the node's last semantic child, which folds the whole annotation into the enclosing section node and swallows all four check labels. `container: true` keeps the header its own node.
- The header tap target was 20 dp tall and the learn-more link 36 dp, against the 48 dp minimum in `accessibility.md`. Both now carry vertical slack their parent gives back, so the Figma 16 px gaps are unchanged.
- "Only a real capture on a real device gets it" claimed more than attestation proves — it vouches for the recording app, not for the frames, and it is defeatable. Reworded to evidence rather than proof, in all 22 locales. The missing-check caveat is untouched.
- The learn-more string was a prefix the URL was concatenated onto. zh and tr had already replaced its trailing space with a colon, so that contract was fiction; it is now a `{url}` placeholder each locale places where its sentence needs it.
- `onInfoPressed`/`infoLabel` plus an assert pairing them became one nullable record, and the no-op `expanded: false` on the sheet is gone.

The third commit applies takeover follow-up:

- The learn-more link now has a regression test proving it launches `proofmodeLearnMoreUrl` externally.
- The header semantics label is a two-placeholder localized string instead of a Dart-built `"$section. $question"` label.
- The orphaned `proofmode*` ARB keys and generated getters left by #5324 are removed.

Still open, and older than this PR: the checklist never announces its own pass/fail state. The four labels merge into one block and the check/X icons are unlabeled images, so a screen-reader user hears the four names and never learns which are satisfied. That needs new strings and a copy decision.

## Type of Change


NotThatKindOfDrLiz:
@hm21 I pushed a takeover follow-up on top of your PR at `333a7086f`.

What changed:
- Added a regression test that taps the metadata verification learn-more link and asserts it launches `proofmodeLearnMoreUrl` externally.
- Moved the header info semantics label into l10n as a two-placeholder string instead of building `"$section. $question"` in Dart.
- Removed the orphaned `proofmode*` ARB keys and regenerated the l10n outputs.
- Rebasing onto current `origin/main` was clean.

Verification:
- `flutter gen-l10n`
- `flutter test test/widgets/video_feed_item/metadata/metadata_expanded_sheet_test.dart test/l10n/arb_consistency_test.dart` — 71 passing
- Pre-push hook: merge-conflict check, analyzer, and changed-file metadata test all passed

One GitHub-side note: after the push, the expected pull_request workflows did not enqueue for the new head. I manually dispatched Mobile CI for `333a7086f` as run `31126808462`, but it is still queued on GitHub at the first job. Please review the takeover commit and merge when the queued CI run/required checks are green.

rabble (approved):
## Overview

Adds an explainer sheet behind an info affordance on the Verification section header, so the four checks stop being four ticks of domain jargon. Strong PR — the description does the reviewer's job for them by pre-empting the two design questions (why a header icon rather than four tappable rows; why the caveat paragraph is load-bearing) and by scoping out the failed-verification state instead of half-shipping it.

A few things I want to call out as good rather than just wave through:

- **The nested-sheet decision is documented with the real reason.** `MetadataVerificationInfoSheet.show` deliberately uses `VineBottomSheet.show` rather than the pause-aware wrapper, because the metadata sheet underneath has already set `setBottomSheetOpen(true)` and *"that flag is a plain bool rather than a counter — so a nested pause-aware sheet would clear it on dismiss and resume playback behind the still-open metadata sheet."* That is exactly right, and it independently confirms a concern I raised on #6788 and #6780 in this same batch. Worth promoting to an issue against the notifier — right now every caller has to know this and work around it locally, and this PR is the only one of the three that does.
- **The accessibility work is unusually thorough.** `Semantics(container: true)` with the reason it's needed (both children excluded, so there's nothing left to anchor to and the annotation folds into the section node); folding the visible section label into the semantic label so the checklist's owning section is still announced; `Tooltip(excludeFromSemantics: true)` because the platform would otherwise append the message a second time (iOS onto the label, Android as `tooltipText`); `MergeSemantics` per explanation so each check reads as one item. Most PRs get one of those four.
- **`_infoHeaderTapSlack` is derived, not guessed.** 20 dp row + 14 above and below = 48, and the same 14 is subtracted from the section's top padding *and* the label-to-content gap so the Figma-specified 16 px spacing survives text scaling. The relationship is stated in the constant's doc rather than left for someone to rediscover.
- **`_LearnMoreLink` puts the URL in the sentence as a placeholder** so a translation can move it, strips the scheme from a single constant so the label can't drift from the target, and degrades to a single unstyled span if a translation drops the placeholder — with that degradation documented rather than left as a latent crash.
- **l10n is complete and real.** All new keys land in all 22 locales with genuine translations, and the footnote's careful hedging survives translation (the German and Bulgarian both preserve "it only means we can't prove that part" rather than flattening to "it's unverified"). Given the PR explicitly asks reviewers not to strengthen that copy — agreed, and the translations honour it.
- Reusing `proofmodeLearnMoreUrl` instead of a third hardcoded copy.

---

### Blocking on CI, not on code

CI never actually ran on head (`333a708`):

```
completed  cancelled  Detect App CI Scope
completed  skipped    Analyze / Format / Generated Files / Tests / VGV Tag Gate
completed  cancelled  Mobile CI
```

The previous commit (`eb375d8`) was fully green, so this looks like the same Actions-infrastructure flakiness hitting several PRs in this batch rather than anything in the diff. Please re-trigger and confirm green before merge — with 22 ARB files touched, the `arb_consistency_test` result in particular is worth having on the actual head.

---

### Smaller points

- **`16 - headerSlack` goes negative if the slack ever exceeds 16.** Both `EdgeInsets.fromLTRB(16, 16 - headerSlack, …)` and `spacing: 16 - headerSlack` currently evaluate to `2.0`, which is fine — but the coupling between "tap slack" and "section padding" is arithmetic with no guard. If a future row height pushes `_infoHeaderTapSlack` past 16, `Column.spacing` asserts and the padding silently inverts. An `assert(_infoHeaderTapSlack <= 16)` next to the constant, or a `clamp`, makes the constraint explicit.
- **The out-of-scope note deserves an issue.** *"when `status == 'invalid'` with no raw tags, `hasProofMode` is false and the whole section disappears, so a video whose signature failed looks exactly like one that never claimed anything."* That's a more serious gap than the one this PR closes — a failed check currently renders as absence, which is precisely the inference the new footnote tells users not to draw. Filing it means the explainer's hedge isn't quietly carrying that weight forever.
- **`_spans` splits on every occurrence of the URL.** Correct as written (multiple occurrences each get underlined), just noting it isn't a first-occurrence split, in case a translation legitimately repeats the domain in prose.
- `MetadataSectionInfo` is a record typedef with `label` documented as describing the explainer rather than the icon — good, and worth keeping that doc line if the type ever moves, since `label` on an icon-adjacent type reads the other way by default.

rabble:

NotThatKindOfDrLiz:
## Post-Merge Review Summary

**Status:** Merged ✅  
**CI Checks:** All passing  

### Blocking Findings

Three items identified during review that should be addressed:

1. **Missing Test Coverage** (#7885)  
   The learn-more link tap handler in `metadata_verification_info_sheet.dart` (line 146) is untested. Add assertion to verify `launcher.launched.last.url == proofmodeLearnMoreUrl`.

2. **Hardcoded Translatable Separator** (#7886)  
   The info label concatenation in `metadata_section.dart` (line 123) uses hardcoded `". "` separator, preventing translators from customizing for CJK and other languages. Convert to two-placeholder ARB key.

3. **Orphaned ARB Keys** (#7887)  
   14 `proofmode*` keys from PR #5324 have zero references and should be removed from all 22 locale files. These were orphaned by the metadata verification refactor.

### Architecture Notes

- ✅ Tap target sizing correct: 48dp minimum maintained for header (20dp + 14dp slack) and links
- ✅ Figma spacing preserved through arithmetic giveback (16px gaps maintained despite padding adjustments)
- ✅ All CI shards passing; no format, analyze, or generated-file issues
- ✅ learn-more URL correctly uses `proofmodeLearnMoreUrl` constant (DRY principle)

Follow-up issues filed: #7885, #7886, #7887

PR #3242, opened 2026-04-22
feat(video-player): implement the native video-player in the feed

hm21:
## Description

This PR extends #2731 and implements the native `DivineVideoPlayer` package in the home/explore feed. It replaces the feed-player path there, but it does not remove `media_kit` from every remaining playback surface in the app.

The following will be improved in the feed-player path:

- No preloading limitation beyond the configured window size
- Rewritten structure that removes duplication and makes follow-on migration simpler
- HLS playback support in the Android feed path
- No more thumbnail fallback when swiping, which improves perceived playback continuity
- Smaller bundle size in this path because ffmpeg is no longer needed here
- Hot restart no longer crashes in the migrated feed path
- New scroll physics closer to TikTok/Instagram feel

### Video cache & download pipeline

- **Cancellable prefetch downloads.** Replaced `flutter_cache_manager`'s prefetch download path with a custom `CancellableDownloader` so cancelled prefetches stop consuming bandwidth promptly instead of running to completion in the background.
- **Platform-native HTTP preference for the cancellable media-cache path.**
  - **iOS:** `cupertino_http` via `NSURLSession`
  - **Android:** `cronet_http` via Chromium net stack
  - **Desktop / fallback:** `dart:io HttpClient`
  - Fallback remains in place so devices without the preferred native stack still keep a working downloader.
- **Cache files survive app restart.** The cancellable download path persists via `store.putFile(CacheObject(...))` so cached videos remain reachable on next launch.
- **Alias map for fallback URLs.** When a video is cached under a fallback key, the manifest also registers it under the stable `aliasKey`, persisted across launches, so the prefetcher does not retry the failing first-attempt URL after a cold start.

## Example Video-Format HLS on Android

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      <video src="https://github.com/user-attachments/assets/6d16a328-4250-49e4-9b58-a757e19f1f81" controls muted playsinline width="250"></video>
    </td>
    <td>
      <video src="https://github.com/user-attachments/assets/275c9b8c-a6cf-4965-bea8-ce9c6be2b07c" controls muted playsinline width="250"></video>
    </td>
  </tr>
</table>

## Example Swipe between two videos

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
[...]

PR #6794, opened 2026-08-06
feat(feed): confirm playback toggles with a snackbar

### LINKED ISSUE #6795: feat(feed): confirm playback toggles with a snackbar
Auto-created from PR #6794.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6794
**Branch:** `feat/playback-toggle-snackbar`

hm21:
Closes #6795

## Description

Two of the three toggles in the feed playback pill — compilations (auto advance) and closed captions — don't communicate what they do. The icon swap is small, and unlike mute neither has an immediately perceivable effect, so tapping one leaves you guessing whether anything happened and what it changed. Both now confirm their new state in a `DivineSnackbarContainer` banner: "Auto advance on/off", "Captions on/off".

Design notes:

- **Mute deliberately stays banner-free.** The sound stopping is already its own confirmation, and mute is the most-tapped control of the three — a banner on every tap would be pure noise.
- **Status-only copy, no explanation sentence.** Naming the state is what was missing; a second clause would push the banner into two lines over the video.
- **2 s instead of the 4 s default**, and each banner replaces its predecessor via `hideCurrentSnackBar()` rather than queueing — rapid toggling always shows the current state instead of replaying a backlog. The tradeoff is that a toggle can also clear an unrelated banner that is still on screen; queueing behind a 4 s banner would delay the feedback past the tap that caused it, which is the worse failure.
- **Auto advance drops its explicit `SemanticsService.sendAnnouncement`.** Flutter's `SnackBar` wraps its content in `liveRegion: true`, so keeping both would make a screen reader speak twice with two different strings.

Four new ARB keys, translated into all 21 non-English locales (no English fallbacks), reusing each language's existing terminology for "auto advance" and "captions".

The second commit removes the dead auto-advance rail plumbing that this change leaves fully unreachable. `announceAutoAdvanceToggle` looked like it still had a caller in `feed_videos.dart`, but that caller — `_toggleAutoAdvance` — was only ever passed down as `onToggleAutoAdvance` through `_Overlay` → `_FeedItemActions` → `_FeedItemOverlayActions`, which accepted the callback and never forwarded or invoked it. `showAutoButton` was dead the same way: `VideoOverlayActionColumn` takes it and never reads it. Both are leftovers from the removed Auto rail control, so the pill is the only auto-advance affordance left. Dropping the unused `effectiveAutoEnabled` alongside them also narrows the overlay's `context.select` to `isEffectivelyActive` only, so an `enabled` flip that leaves Auto suppressed no longer rebuilds every visible feed item.

**Related Issue:** 

## Out of Scope

None.

## Verification

- `flutter test test/widgets/video_feed_item/feed_playback_toggles_pill_test.dart` — 8 passing, including new assertions on the banner text, an explicit `findsNothing` pinning mute's banner-free behaviour, and a re-toggle test for replace-not-queue
- `flutter test test/l10n/arb_consistency_test.dart` — clean
- `flutter test test/screens/feed/ test/widgets/video_feed_item/` — 471 passing, covering both surfaces that host the pill and every widget touched by the dead-code removal
- Manually verified on a real Samsung Galaxy device: both toggles in the top-bar settings popover and in the paused-video overlay, plus rapid toggling

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mobile/lib/widgets/video_feed_item/feed_playback_toggles_pill.dart
```diff
@@ -105,7 +107,15 @@ class _PlaybackModeToggle extends StatelessWidget {
             if (!cubit.state.isEffectivelyActive) {
               cubit.clearPendingPaginationAdvance();
             }
-            announceAutoAdvanceToggle(context, enabled: cubit.state.enabled);
+            // The snackbar is a live region, so it carries the screen-reader
+            // announcement on its own — no separate `sendAnnouncement`, which
+            // would make a reader speak twice with two different strings.
+            _showToggleFeedback(
+              context,
+              cubit.state.enabled
```
realmeylisdev, line 115: Reading `cubit.state.enabled` after `toggle()` is the right instinct, but `toggle()` has a third branch: when Auto is enabled **and** suppressed it resumes and leaves `enabled` true (`feed_auto_advance_cubit.dart:68-75`).

On the paused overlay that branch isn't an edge case — it's the normal path. `__OverlayState._handlePlayerTap` calls `onSuppressAutoAdvance` *before* pausing (`feed_videos.dart:669`), so the very tap that reveals this pill is the tap that sets `suppressed`. With Auto on, the first press of the compilations toggle there always resumes instead of disabling: the icon doesn't move, the label stays "Disable auto advance", and the banner says "Auto advance on". A second press is what actually turns it off.

I reproduced it on a simulator — first tap: icon unchanged, banner "Auto advance on"; second tap: icon flips.

Nothing false is displayed and `toggle()` predates this PR, so I didn't push anything. But the banner is what makes it visible, and this is one of the two surfaces the PR exists to improve — a control that reports "on" when you pressed the thing labelled "Disable" is the opposite of the goal. Three ways out, all product calls: skip the banner when `enabled` didn't change; drive the icon and label off `isEffectivelyActive` instead of `enabled`; or have the pill bypass the resume branch and always flip.

mobile/lib/widgets/video_feed_item/feed_playback_toggles_pill.dart
```diff
@@ -227,3 +244,32 @@ class _PopoverToggle extends StatelessWidget {
...
+/// state.
+///
+/// Uses `removeCurrentSnackBar` rather than `hideCurrentSnackBar`: hiding only
+/// *starts* the exit animation, and the entry leaves `ScaffoldMessenger`'s
+/// queue when that animation reports `dismissed`, so the banner shown next
+/// lands behind the departing one. Removing drops it in the same frame, which
+/// is what makes this a replace instead of a queue.
+void _showToggleFeedback(BuildContext context, String message) {
+  ScaffoldMessenger.of(context)
+    ..removeCurrentSnackBar()
```
realmeylisdev, line 263: Fixed here in `5492aca`. For the record, the mechanism: `hideCurrentSnackBar()` starts the reverse animation and returns (`scaffold.dart:441-458`); the entry is only removed from `_snackBars` when the controller reports `dismissed` (`:396-405`). So `showSnackBar()` in the same cascade sees a non-empty queue, skips `forward()`, and appends. `removeCurrentSnackBar()` sets the controller to 0, which fires `dismissed` in the same frame.

**Still open, and not fixed by the swap:** neither call targets *our* banner — both act on whatever is at the front of the app-wide messenger. On the reel screen `reel_dm_reply_bar.dart:350-364` shows a raw `SnackBar` with a real `SnackBarAction` (Retry), and `SnackBar.persist` defaults to `action != null` (`snack_bar.dart:303`), so the auto-dismiss timer returns without hiding it (`scaffold.dart:622-624`) — it sits there until acted on. Toggling captions or Auto wipes it. Same for the "Sent / View chat" banner at `:377-392`. The draft is restored either way so nothing is lost, but the affordance is, and that file's own comment at `:424` already names this hazard ("without the assert it dismisses whichever snackbar *is* in front").

`showSnackBar` returns a controller — holding onto it and closing only that would confine this to our own banner. Left to you since it changes what happens to other people's banners.

mobile/lib/widgets/video_feed_item/feed_playback_toggles_pill.dart
```diff
@@ -227,3 +244,32 @@ class _PopoverToggle extends StatelessWidget {
...
+    ..showSnackBar(
+      DivineSnackbarContainer.snackBar(
```
realmeylisdev, line 265: No `margin`, so this floats at the bottom of the Scaffold *body*. On the reel screen the inline comment composer deliberately sits in the body `Column` rather than `bottomNavigationBar` so it can ride above the keyboard (`inline_comment_composer_bar.dart:20-21`) — which means the Scaffold's snackbar anchor knows nothing about it and the banner lands directly on top.

Measured with the real composer at 393×852, bottom inset 34: banner 764–808 vs input field 774–798. Keyboard up: banner 462–506 vs field 472–496. `SnackBar` also defaults `hitTestBehavior` to `opaque` when `margin == null` (`snack_bar.dart:838-842`), so the field and Send are untappable for those 2 s — which is precisely the scenario `_FeedSettingsOverlay`'s `TextFieldTapRegion` was written for ("keeps the inline comment composer's keyboard up while the user adjusts playback").

`clip_delete_snackbar.dart:30` already lifts a banner this way (`margin: EdgeInsets.fromLTRB(16, 0, 16, 68)`). I didn't push one because the home feed doesn't need it and a blanket margin would float these banners higher than every other one in the app — whether the host supplies the inset is your design call.

mobile/lib/l10n/app_bg.arb
```diff
@@ -4069,6 +4069,10 @@
     "videoSettingsMenuClose": "Затвори настройките за възпроизвеждане",
     "videoSettingsCaptionsEnable": "Включи субтитрите",
     "videoSettingsCaptionsDisable": "Изключи субтитрите",
+    "videoSettingsAutoAdvanceOn": "Автоматичното продължаване е включено",
```
realmeylisdev, line 4083: Nit, take it or leave it: this is the longest of the 22 and wraps to two lines on a 360 dp phone, which is the thing the description says the copy was deliberately kept short to avoid. Something shorter like "Автоматично продължаване: вкл." would stay on one line. Same applies to the `Off` string below it.

realmeylisdev (changes requested):
Good change, and the 21-locale coverage is genuinely thorough. Two of the banner mechanics didn't hold up when I checked them, so I pushed fixes. One behaviour needs your decision before this ships.

**Pushed**

- `5492aca` — `hideCurrentSnackBar()` queues, it doesn't replace. It only *starts* the exit animation; the entry leaves the messenger's queue when that animation reports `dismissed`, so the next `showSnackBar()` lands behind the departing banner. On the simulator: tap twice and the stale banner is still up after the icon has already flipped; tap three times quickly and the middle state sits there for its full 2 s — the backlog the description says this pattern avoids. Now uses `removeCurrentSnackBar()`.
- `cbc8099` — the test for that couldn't fail. `tester.pump()` advances zero time, so the entrance controller never leaves 0 and the dismiss collapses to a zero-duration reverse; the assertion passed against both the buggy and the fixed code. It now gives the first banner real time and goes red on the old version. Also dropped the three `pumpAndSettle(const Duration(seconds: 3))` drains — that argument is the frame interval, not a timeout, and nothing was pending anyway.

**Your call** (details inline)

- On the paused-video overlay, turning Auto *off* always takes two taps, and the first one banners "Auto advance on" with nothing else changing. Reproduced on a simulator. This is the one I'd want settled.
- The banner lands on the reel's inline comment composer and swallows taps on it for 2 s.
- The global dismiss takes out any banner in front of it, including the reel-reply Retry, which never self-dismisses.

One thing for later, not this PR: the second commit says the pill is the only auto-advance affordance left, but the rail control itself is still there — `AutoActionButton` has had no production caller since #3242, and `videoActionAutoLabel` exists only to feed it. Its `video_feed_page_web_test.dart:67` assertion is gated `skip: !kIsWeb` so it never runs. Worth a separate cleanup.

Validation on `cbc8099`: `flutter analyze lib test integration_test` clean; `test/screens/feed/ test/widgets/video_feed_item/ test/l10n/arb_consistency_test.dart` → 479 passing, 1 skipped (the pre-existing `!kIsWeb` one); format and the pre-push hook clean. Ran the app on an iOS simulator to confirm the queueing bug, the fix, and the paused-overlay behaviour.

Heads-up on CI: the Mobile CI jobs on this head are red for infrastructure reasons, not code. Every one is either `cancelled` or stopped at `Set up job` — Analyze, Format and VGV Tag Gate all died at exactly 15m02s without executing anything. I re-ran them twice and the runners are still not provisioning. Worth another re-run before you read anything into those reds.

mobile/lib/widgets/video_feed_item/feed_playback_toggles_pill.dart
hm21: Confirmed and fixed in 6da2cfa. The resume branch was written for the Auto rail control, which has had no production caller since #3242 — the pill is `toggle()`'s only caller now, so the branch only ever fired on the paused overlay, where it is the normal path. `toggle()` always moves `enabled` now; `resumeAfterSwipe` still handles the case that actually needs resuming.

Pinned by a widget test that seeds enabled+suppressed and asserts one tap disables — it fails on the old cubit. The cubit test that pinned the resume branch flipped to assert the disable.

mobile/lib/widgets/video_feed_item/feed_playback_toggles_pill.dart
hm21: Leaving the margin as it is. Verified on device: on the feed the banner sits exactly above the bottom nav bar, which is where it belongs and where every other banner in the app sits. A blanket inset would float these higher than the rest for a case the host surface should own.

realmeylisdev (approved):
Verified `6da2cfa` — the fix holds.

`toggle()` is now `setEnabled(enabled: !state.enabled)`, and the pill is the only caller left in `lib/`. `setEnabled(true)` still clears `suppressed`, so re-enabling from the paused overlay resumes in that same tap instead of waiting for a swipe, and `setEnabled(false)` resets the rest of the state. Icon and label read `state.enabled`, which now always moves, so the control and the outcome agree on both surfaces. The new widget test fails against the old cubit, and the cubit test that pinned the resume branch was flipped rather than dropped.

Locally: `flutter analyze lib test integration_test` clean, `test/screens/feed/` + `test/widgets/video_feed_item/` 472 passing.

The red CI is infrastructure, not this diff. The failed jobs never got a runner — no runner assigned, zero steps, cancelled at exactly 15 minutes. GitHub Actions has been in an open major outage since 15:22 UTC and the whole repo's queue is stalled; runs on other branches have sat queued for over an hour. I re-ran all three and they're waiting behind the same queue. It still needs to be green before this merges.

Approving so it isn't waiting on me. Two threads are still yours, neither blocking:

- The app-wide `removeCurrentSnackBar()`. There's no free fix — scoping it to our own controller means queueing behind a foreign banner, which is the tradeoff you already weighed in the description. Whichever way you take it is fine.
- The `app_bg.arb` line-length nit.

PR #6796, opened 2026-08-06
fix(explore): drop the status-bar gap above Lists and Apps tabs

### LINKED ISSUE #6797: fix(explore): drop the status-bar gap above Lists and Apps tabs
Auto-created from PR #6796.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6796
**Branch:** `fix/explore-tab-top-gap`

hm21:
Closes #6797

## What

The Explore **Lists** and **Apps** tabs rendered a visible gap above their first card, even though nothing sits above them inside the tab view.

## Why

Both tabs build a `ListView` without an explicit `padding`. In that case `BoxScrollView.buildSlivers` falls back to inserting `MediaQuery.padding` as sliver padding — with only left/right zeroed, so **both** vertical insets. Explore renders its own header instead of an `AppBar`, so `MediaQuery.padding.top` is still the full status-bar inset at that point in the tree, and the first card got pushed down by exactly that much. The remaining Explore tabs (New, Categories, …) already pass an explicit padding, which is why only these two were affected.

## Two commits

**1. `fix(explore): drop the status-bar gap above Lists and Apps tabs`** — `padding: EdgeInsets.zero` on both list views.

**2. `fix(apps): keep the bottom safe-area inset on the standalone apps list`** — zeroing *both* insets was wrong for `AppsDirectoryScreen`, because that widget renders in two places:

- embedded in Explore, where `AppShell`'s `bottomNavigationBar` has already removed the bottom inset, so zeroing it is a no-op;
- standalone at `/apps`, a top-level route whose own `Scaffold` has an app bar and **no** bottom nav — so Flutter removes only the top inset from the body and the bottom one survives all the way down to the `ListView`.

Measured on the standalone route with a 34 px bottom inset: `SliverPadding` went from `EdgeInsets(0, 0, 0, 34)` to `EdgeInsets.zero`, and since `_AppsDirectoryRow` has no bottom padding of its own the last card ended flush against the gesture bar at scroll end. The second commit resolves the inset inside `_AppsDirectoryContent` — below the `Scaffold`, so the same widget reads `0` embedded and the real inset standalone:

```dart
padding: EdgeInsets.only(bottom: MediaQuery.paddingOf(context).bottom),
```

`ExploreListsTab` needs no equivalent change: it only ever mounts inside the shell, where the bottom inset is already gone.

## Test plan

- New regression test in `apps_directory_screen_test.dart`: pumps the standalone mount under a 34 px bottom inset, scrolls to the true list end, asserts the last card clears the inset. Fails on commit 1 (last card bottom `600.0`, screen edge), passes on commit 2 (`≤ 566.0`).
- `flutter test test/screens/apps/` — 83/83 green, stable under `--test-randomize-ordering-seed random`
- `flutter analyze` on all three changed files — clean
- Manually verified on a real Samsung Galaxy: Lists and Apps tabs now start flush under the tab bar; standalone Apps directory route (`/apps`) unchanged at the top.

### DIFF
--- a/mobile/test/screens/apps/apps_directory_screen_test.dart
+++ b/mobile/test/screens/apps/apps_directory_screen_test.dart
@@ -126,6 +126,32 @@ void main() {
       expect(find.text('Primal'), findsOneWidget);
     });
 
+    testWidgets('embedded drops the status-bar inset above the first card', (
+      tester,
+    ) async {
+      // 120 physical px / 3.0 test devicePixelRatio = 40 logical px of status
+      // bar. Embedded in Explore nothing above this list consumes the top
+      // inset, so a null padding would push the intro card down by exactly
+      // that much — the gap reported in #6797.
+      tester.view.padding = const FakeViewPadding(top: 120);
+      addTearDown(tester.view.resetPadding);
+
+      when(
+        () => mockDirectoryService.fetchApprovedApps(),
+      ).thenAnswer((_) async => [_fixture()]);
+
+      await tester.pumpWidget(buildEmbeddedSubject());
+      await tester.pumpAndSettle();
+
+      // _AppsDirectoryIntro's own EdgeInsets.fromLTRB(16, 16, 16, 8) is the
+      // only offset that may sit between the viewport top and the card.
+      expect(
+        tester.getRect(find.byType(DivineInfoCard)).top -
+            tester.getRect(find.byType(ListView)).top,
+        moreOrLessEquals(16, epsilon: 0.5),
+      );
+    });
+
     testWidgets('standalone keeps the last card clear of the bottom inset', (
       tester,
     ) async {
--- a/mobile/test/screens/apps/apps_directory_screen_test.dart
+++ b/mobile/test/screens/apps/apps_directory_screen_test.dart
@@ -126,6 +126,45 @@ void main() {
       expect(find.text('Primal'), findsOneWidget);
     });
 
+    testWidgets('standalone keeps the last card clear of the bottom inset', (
+      tester,
+    ) async {
+      // 102 physical px / 3.0 test devicePixelRatio = 34 logical px, i.e. a
+      // home indicator or gesture bar. The standalone route has no bottom
+      // nav, so Scaffold leaves this inset in the body's MediaQuery and the
+      // list has to reserve it itself.
+      tester.view.padding = const FakeViewPadding(bottom: 102);
+      addTearDown(tester.view.resetPadding);
+
+      when(
+        () => mockDirectoryService.fetchApprovedApps(),
+      ).thenAnswer((_) async => List.generate(12, _numberedFixture));
+
+      await tester.pumpWidget(buildSubject());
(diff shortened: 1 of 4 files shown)


mobile/test/screens/apps/apps_directory_screen_test.dart
```diff
@@ -126,6 +126,71 @@ void main() {
...
+      // _AppsDirectoryIntro's own EdgeInsets.fromLTRB(16, 16, 16, 8) is the
+      // only offset that may sit between the viewport top and the card.
+      expect(
+        tester.getRect(find.byType(DivineInfoCard)).top -
+            tester.getRect(find.byType(ListView)).top,
+        moreOrLessEquals(16, epsilon: 0.5),
+      );
+    });
+
+    testWidgets('standalone keeps the last card clear of the bottom inset', (
```
realmeylisdev, line 155: This one passes unchanged on `main` — there the ListView had no `padding` argument, so `BoxScrollView`'s fallback produced the same `EdgeInsets.fromLTRB(0, 0, 0, 34)` this line now computes explicitly. It only goes red against commit 1's `EdgeInsets.zero`, so it pins that intermediate state rather than the gap #6797 reports.

Kept it as a pin and added `embedded drops the status-bar inset above the first card` above it, plus a mirror for the Lists tab, so the case this PR actually changes is covered too.

realmeylisdev (approved):
Approving — the diagnosis is right, and I checked each step against source rather than taking it on trust.

- `BoxScrollView.buildSlivers` does keep **both** vertical insets when `padding` is null (pinned SDK, `scroll_view.dart:900-916`).
- On Explore the shell Scaffold passes `appBar: null` (`app_shell.dart:390`), so the status-bar inset survives down to the tab content. That's the gap.
- Its `bottomNavigationBar` is unconditional and there's no `extendBody`, so the bottom inset really is already 0 embedded; and `/apps` is registered top-level rather than under `shellRoutes()`, so standalone really does keep it. The asymmetry between the two files is correct.

One correction to the description: the other tabs aren't safe because they "already pass an explicit padding". Classics, New, Popular and For You all scroll through a `CustomScrollView`, which has no MediaQuery fallback at all — only Categories fits that reasoning. Same conclusion, different reason, worth knowing if anyone leans on it later.

**Fixed and pushed** (51503f86a): the new `standalone keeps the last card clear of the bottom inset` test passes unchanged on `main`. Without the `padding` argument the framework fallback produced the same `EdgeInsets.fromLTRB(0, 0, 0, 34)` the explicit padding now computes, so it only goes red against commit 1's `EdgeInsets.zero` — it guards that intermediate state, not #6797. The behaviour that actually changes for users (dropping the top inset in the embedded mount) had no coverage on either file, and `explore_lists_tab.dart` had no test file at all.

Added two tests measuring the first card against the viewport top under a 40 px top inset: each reports 56.0 with the padding argument removed and 16.0 with it in place. Analyze clean, affected suites green under `--test-randomize-ordering-seed random`.

Nothing else outstanding.

Issue #5516, opened 2026-06-25
fix(explore): keep Popular variant switches from showing stale data


NotThatKindOfDrLiz:
## Problem
Switching Explore > Popular between Native and Classic can briefly show stale data from the previously selected variant while Riverpod rebuilds the selected variant. The foreground-idle warmup should also prepare both variants so switching to Classic does not require a cold first-page fetch.

## Expected behavior
- Popular Native and Popular Classic each render only data fetched for that selected variant.
- During a variant switch, the tab shows the existing loading state until the selected variant has resolved.
- Foreground-idle warmup preloads both Popular variants without mutating the currently visible feed.

## Acceptance criteria
- The Popular loaded-variant marker cannot advance before matching feed data is published.
- Native and Classic warmup paths are covered by provider tests.
- Regression coverage fails if a selected variant can be marked loaded while the feed still contains the previous variant data.

## Verification
- `flutter analyze`
- `flutter test test/providers/foreground_idle_warmup_provider_test.dart test/providers/explore_feed_refresh_retention_test.dart`

mbradley:
Superseded by #6798, which ships the opposite of this issue's third acceptance criterion.

This issue asked that "during a variant switch, the tab shows the existing loading state until the selected variant has resolved." #6798 holds the previously rendered page instead and cross-fades the arriving variant in, because the spinner branch unmounted the variant toggle itself: the control disappeared and came back under the user's finger on every tap.

The loaded-variant marker from #5513 is untouched, and so is the regression coverage in `explore_feed_refresh_retention_test.dart` — the provider still never advances the marker before matching data is published. What changed is the widget layer, which now paints the previous variant's grid during the switch rather than a spinner.

Noting it here because the criterion above otherwise reads as current, and a stale criterion is easy to inherit as a requirement.

PR #5513, opened 2026-06-25
fix(explore): warm popular classic feed before switching

rabble:
Closes #5516

## Summary
- warm both Popular Native and Classic during foreground-idle warmup
- track which Popular variant the current data belongs to, so variant switches show the existing loading indicator until the selected variant resolves
- publish the loaded-variant marker from the Popular provider state transition so it cannot advance before matching feed data is available
- add provider regression coverage for stale variant data, loaded-marker ordering, and warmup coverage for both variants

## Root cause
Riverpod can retain the previous `AsyncValue` while a dependency-triggered Popular rebuild is loading. The initial fix added a loaded-variant marker, but it updated that marker inside the fetch helper before the provider published the matching `AsyncData`, leaving a possible transient stale-render window.

## Verification
- `flutter analyze`
- `flutter test test/providers/foreground_idle_warmup_provider_test.dart test/providers/explore_feed_refresh_retention_test.dart`

PR #6798, opened 2026-08-06
fix(explore): cross-fade the popular feed on variant switch

### LINKED ISSUE #6799: fix(explore): cross-fade the popular feed on variant switch
Auto-created from PR #6798.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6798
**Branch:** `fix/popular-feed-variant-crossfade`

hm21:
Closes #6799

## Description

Switching the Popular tab between **New** and **Classic** jumped instead of transitioning: the grid vanished, a full-screen spinner flashed, and the grid came back.

The root cause is the loaded-variant marker, not a missing animation. On a variant switch `PopularVideosFeed` resets `popularVideosLoadedVariantProvider` to `null` — deliberately, so a stale Native page is never presented as Classic — which made `PopularVideosTab` fall through to `_PopularVideosLoadingState`. And because the trending header (including the variant toggle) lives inside `_PopularVideosTrendingContent`, that spinner branch unmounted the very toggle the user had just tapped, then hard-mounted it again.

The tab now remembers the page it last painted together with the variant it belongs to, and keeps showing it while the newly selected variant loads, so the header and toggle stay put. All three states are wrapped in an `AnimatedSwitcher` keyed by variant, so the arriving page cross-fades in — one transition instead of grid → spinner → grid. First load and the error state fade as well.

Why hold the *last painted* page rather than whatever the provider currently exposes: Riverpod retains the previous `AsyncValue` during the rebuild, so the two are normally the same object — but if the toggle is tapped back and forth before the first switch resolves, the provider can hold a page for a variant that is no longer selected and will never be marked loaded. Keying the held page by its own variant keeps the transition honest in that case and preserves the guarantee the marker exists for. The provider itself is untouched; holding the previous page is purely transitional UI state.

`_buildDataState` was a `Widget`-returning method that also carried the analytics side effects. It is now `_trackDataState` (tracking only) with the widget built inline, per the repo's widgets-over-methods rule.

`MediaQuery.disableAnimationsOf` collapses the cross-fade to zero for reduce-motion users, matching the existing usage elsewhere in the app.

**Related Issue:** 

## Out of Scope

There is no progress affordance while the new variant loads — the moving toggle pill is the only feedback. Both variants are pre-warmed by the foreground idle warmup, so the switch is usually immediate; a loading indicator can follow if the cold-cache case turns out to be noticeable in practice.

## Verification

- New `mobile/test/widgets/popular_videos_tab_test.dart`: asserts no spinner appears during the switch and that the grid only re-keys once the new page arrives. Counter-checked against the old behaviour — without the fix the test fails because it finds `BrandedLoadingIndicator`.
- `flutter test test/providers/explore_feed_refresh_retention_test.dart` — the loaded-variant contract still holds.
- `flutter analyze lib/widgets/popular_videos_tab.dart test/widgets/popular_videos_tab_test.dart` — clean.
- Tested on a real device (Samsung Galaxy).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mbradley (commented):
Reviewed the state machine rather than just the diff: ran the new test, mutated the fix out six ways to see what the suite actually protects, and drove the switch under rapid toggling, backgrounding, and a concurrent `loadMore`. Keying the held page by `rendered.variant` holds up under all of it, and `loadMore` during an in-flight switch cannot mix one variant's page into the other.

Two commits pushed for gaps with clear fixes:

- **`b30e1ac25`** restores slow-load telemetry. Holding the previous page moved the variant switch off the branch that calls `_trackLoadingState`, so a switch slower than the 5s threshold stopped emitting `trackSlowOperation`. That is the load a user actually waits on. Split the threshold check out so both branches report it.
- **`78889813e`** pins the cross-fade. The suite passed with the `AnimatedSwitcher` deleted outright, so the transition this PR is named for was unprotected. The new assertion checks both pages are mounted part-way through the transition, which is false the moment the swap becomes instant.

Neither changes behavior you shipped. Validation: `flutter analyze` clean on both files, `popular_videos_tab_test.dart` and `explore_feed_refresh_retention_test.dart` green.

**Uncertainty worth naming:** I did not add a test for the restored telemetry. It fires only after 5s of real elapsed time via `DateTime.now()`, which `fakeAsync` does not control, so covering it means a 5s test or injecting a clock. The pre-existing slow-op path has the same gap. Left it rather than ship a contrived test.

---

### One thing for you to decide: this reverses #5516

#5516 lists as an acceptance criterion: *"During a variant switch, the tab shows the existing loading state until the selected variant has resolved."* #5513 implemented it, and @NotThatKindOfDrLiz hardened it during review so the marker could not advance before matching data was published. This PR reverses that behavior. Flagging it because nothing linked from here surfaces that decision, so a reviewer following only this PR's links would not find it.

Worth knowing why the existing coverage did not catch the change: `explore_feed_refresh_retention_test.dart` asserts on `popularVideosLoadedVariantProvider` with the reason *"The UI must not treat the old Native page as the selected Classic page."* This PR leaves the marker untouched and moves the behavior up to the widget, so the test stays green while the UI paints Native content under a Classic-selected pill. The mechanism survived; what it was guarding did not.

Two things about the size of that window, both verified in a widget test rather than inferred:

- Backgrounding mid-switch takes `build()` down the `!isAppReady` path, which returns existing data *without* marking the variant loaded. The held page then persists for as long as the app stays backgrounded, not just for the fetch (`loaded=null`, native grid still mounted, no spinner).
- A block landing in that window can put the blocked creator back on screen until the new variant arrives, which reads to the user as the block not working.

Taking the intent as wanted, so this is not a blocker. What I would ask for:

1. Update #5516 so the criterion matches what we are actually shipping. Right now the repo says one thing and the code does another, and the next person to touch this inherits that.
2. Consider whether the pill should be the only feedback. It says Classic while the grid is Native, with nothing indicating a load.

If you want the jank fix without the reversal, a third option: hoist the header and toggle out of `_PopularVideosTrendingContent` so they sit above the state switch. The spinner then replaces only the grid, the toggle stops unmounting under the user's finger, and #5516's criterion survives. That looks closest to the root cause to me, since the reported problem was the toggle disappearing rather than the spinner itself. Your call, and @NotThatKindOfDrLiz may want a word given the history.

Please review the two pushed commits and merge when you are happy with them.

mobile/lib/widgets/popular_videos_tab.dart
```diff
@@ -86,31 +100,56 @@ class _PopularVideosTabState extends ConsumerState<PopularVideosTab> {
       _feedTracker.startFeedLoad('popular');
     }
 
-    // CRITICAL: Check hasValue FIRST before isLoading
-    if (feedAsync.hasValue &&
-        feedAsync.value != null &&
-        loadedVariant == selectedVariant) {
-      return _buildDataState(feedAsync.value!.videos);
+    // A held page outranks an in-flight load; a hard error only wins when
```
mbradley: The comment and the branch order disagree. This says a hard error only wins when nothing is held, but the error branch is evaluated first and gated only on `!showsSelectedVariant`, so an error wins whether or not a page is held.

Low stakes today, since `_loadFirstPage` and `staleWhileRevalidate` both swallow failures into `AsyncData` and `hasError` is close to unreachable. That is exactly why it will bite whenever one of those starts propagating.

```suggestion
    // A held page outranks an in-flight load. A hard error outranks both.
```

Suggested on the assumption the branch order is what you meant. If the comment is right and a held page should win, that is a code change instead.

rabble (changes requested):
## Overview

Holds the last-rendered page (with the variant it belongs to) so a native/classic toggle cross-fades between two grids instead of dropping to a full-screen spinner, and wraps the three states in an `AnimatedSwitcher` keyed per state.

Good things:

- `MediaQuery.disableAnimationsOf(context) ? Duration.zero : ...` honours reduced-motion, which is easy to forget when adding a transition.
- The two magic numbers become named constants (`_feedStateTransitionDuration`, `_slowFeedLoadThresholdMs`).
- `_buildDataState` splitting into `_trackDataState` + a keyed widget removes a `ref.watch` that was buried inside a build-helper.
- **The test is unusually good.** It asserts both pages are mounted mid-transition — the exact thing that distinguishes a cross-fade from a swap — and its `reason` strings explain *why* the key is the rendered variant rather than the selected one. That reasoning is the subtle part of the whole change, and a future refactor that "simplifies" it to `ValueKey(selectedVariant)` will now fail with an explanation attached.

Two things to fix.

---

### 1. `_trackSlowLoadIfNeeded()` will re-fire on every rebuild during a held-page load

The extracted method reports and then leaves `_feedLoadStartTime` set:

```dart
void _trackSlowLoadIfNeeded() {
  if (_feedLoadStartTime != null) {
    final elapsed = DateTime.now().difference(_feedLoadStartTime!).inMilliseconds;
    if (elapsed > _slowFeedLoadThresholdMs) {
      _errorTracker.trackSlowOperation(...);
    }
  }
}
```

In its old home this was fine: `_trackLoadingState` ran on the spinner branch, and the field was cleared as soon as data arrived. Now it's also called from the **held-page** branch, which runs on every `build` while a variant switch is in flight — and on that path `showsSelectedVariant` is false, so `_trackDataState` (which is what nulls the field) never runs.

So once a variant switch passes 5 seconds, every subsequent rebuild emits another `trackSlowOperation('popular_feed_load')` for the same load. Provider ticks, `MediaQuery` changes, parent rebuilds — each one is another analytics event. One slow switch becomes an unbounded stream of them, which is worse for the metric than the gap the method was extracted to close.

Cheapest fix is to make the report one-shot:

```dart
void _trackSlowLoadIfNeeded() {
  final start = _feedLoadStartTime;
  if (start == null) return;
  final elapsed = DateTime.now().difference(start).inMilliseconds;
  if (elapsed <= _slowFeedLoadThresholdMs) return;
  _feedLoadStartTime = null;   // report once per load
  _errorTracker.trackSlowOperation(...);
}
```

Worth a test — `trackSlowOperation` called exactly once across several pumps past the threshold is directly assertable with a mock tracker.

### 2. `_rendered` is assigned during `build`

```dart
if (showsSelectedVariant) {
  _rendered = (variant: selectedVariant, state: feedState);
  _trackDataState(feedState.videos);
}
```

Mutating `State` fields from `build` makes the method non-idempotent: the widget's output now depends on how many times it has been built, not only on its inputs. It happens to work here because the assignment is derived from data the build already read, but it's the pattern that makes "why did this widget render the wrong page" hard to reason about later, and Flutter gives no guarantee about redundant build calls.

`didChangeDependencies` or a `ref.listen` on `popularVideosFeedProvider` would let `_rendered` update on data change rather than on paint. If keeping it in `build` is deliberate for ordering reasons, a comment saying so would help — right now the field's doc explains *what* it holds but not why it's written from here.

(The analytics calls — `_trackDataState`, `_trackErrorState`, `_trackLoadingState` — have the same build-side-effect shape, but they predate this PR, so I'd leave them unless you're already in there.)

---

### Nit

The `_rendered` doc says it keeps "the previous grid on screen while a variant switch loads". It also keeps it on screen when the provider errors *after* a successful load, since `feedAsync.hasError && !showsSelectedVariant` sends an error with held data down the held-page branch. That's almost certainly the behaviour you want — a transient refresh failure shouldn't blank a working grid — but it's a second job the comment doesn't mention, and a reader checking "what happens on error" will read the comment and conclude the error state shows.

mobile/lib/widgets/popular_videos_tab.dart
hm21: You are right, and the comment was mine. The error branch is evaluated first and gated only on `!showsSelectedVariant`, so it outranks a held page — my wording described an ordering the code does not have. Took your suggestion verbatim in acd0ea7df.

Agreed on the stakes too: `_loadFirstPage` swallows into `AsyncData` and `staleWhileRevalidate` never leaves the provider in `AsyncError`, so `hasError` is close to unreachable today. That is exactly why a comment that lies about the ordering is worth fixing now rather than when something starts propagating.

hm21:
@rabble Both fixed in acd0ea7df, plus the nit.

**1. `_trackSlowLoadIfNeeded()` re-firing — confirmed and fixed.** Reproduced it before fixing: three rebuilds past the threshold produced three `trackSlowOperation` events. Your diagnosis is exactly right, and backgrounding makes it worse than "a few ticks" — `!isAppReady` returns existing data *without* marking the variant loaded, so the held page persists for as long as the app stays backgrounded and the duplicates are unbounded.

I did not take the one-shot literally, though. Nulling `_feedLoadStartTime` there would have traded the duplicates for a different gap: that same field gates `markFirstVideosReceived` / `markFeedDisplayed` / `markDataLoaded` in `_trackDataState`, so clearing it in the slow-load path means the eventual data arrival reports nothing — losing feed-displayed telemetry on precisely the slow loads. Used a separate `_slowFeedLoadReported` flag instead, reset when a new load starts, so both signals survive.

Test added as you asked — `reports a slow variant switch once, not once per rebuild`, verified red on the unguarded version and green with the flag. It needed one production concession: the elapsed time comes from `DateTime.now()`, which `fakeAsync` cannot advance, so the threshold is now an optional widget parameter alongside the three analytics injectables that were already there. Open to a different approach if you would rather not have that on the constructor — the alternative is a 5s test.

**2. `_rendered` assigned during `build` — kept, with the reasoning written down.** I went looking for the move you suggested and neither destination works here. `didChangeDependencies` does not fire for `ref.watch` invalidations (Riverpod rebuilds through `setState`, not an InheritedWidget change), and a `ref.listen` callback runs *after* the build that carried the new data — so the tab would paint the outgoing page for one extra frame and the cross-fade would start a frame late, which is a milder version of the jump this PR exists to remove. The assignment is also derived entirely from providers the same build already read, so a redundant build recomputes the same value rather than accumulating.

So it stays, but the field's doc now says why it is written from `build` rather than only what it holds. If you still want it out of `build` knowing the frame cost, say so and I will do it.

**Nit:** correct, and it was undocumented behaviour rather than a documentation gap in the usual direction — the doc now names the post-load error case explicitly.

mbradley:
Caught up on `acd0ea7df`. Everything I raised is resolved, and one correction to the record on the way through.

**The re-fire bug predates this PR.** @rabble's mechanism is right and the catch is his, but the framing "in its old home this was fine, the field was cleared as soon as data arrived" does not match `origin/main`. The old fall-through called `_trackLoadingState()` on every build with no guard, and during a variant switch `_trackDataState` never ran to clear `_feedLoadStartTime` either, because `loadedVariant != selectedVariant` is exactly what routed the build down that path. So a slow variant switch on `main` today emits one `trackSlowOperation` per rebuild.

My extraction in `b30e1ac25` added a second call site and I saw the re-fire risk and talked myself out of it, so the catch still lands. But the flaw is older than my commit, and `_slowFeedLoadReported` therefore fixes a live bug on `main` rather than repairing something this PR introduced. Worth having written down accurately, since the version currently in the thread reads as reviewer-broke-it.

**The deviation from the suggested fix was the right call.** Nulling `_feedLoadStartTime` as a one-shot would have traded a loud bug for a quiet one: that same field gates `markFirstVideosReceived`, `markFeedDisplayed` and `markDataLoaded`, so feed-displayed telemetry would have gone missing on exactly the slow loads. The separate flag keeps both signals.

Verified rather than assumed: dropping `_slowFeedLoadReported` makes the new test fail with 3 events instead of 1, so it fails for the right reason.

**On `slowLoadThresholdMs` — keep it.** You asked whether it belongs on the constructor. It sits alongside three analytics injectables already there for the same reason, and the alternative is a five second test. That is a fair trade for a widget that already takes its trackers by injection.

**Not holding on #5516.** Nobody has argued for the old behaviour, the jank being fixed is real, and holding a PR on a question no one wants to answer is not useful. Left a note on #5516 so the issue stops asserting the opposite of what ships, since that stale criterion is how this became invisible in the first place.

mbradley (approved):
Approving the takeover, per the handback step in `PR_REVIEW.md`. Detail is in the comment above; this is the short version.

**Pushed:** `b30e1ac25` (slow-load telemetry on the held-page branch) and `78889813e` (an assertion that the cross-fade actually cross-fades, since the suite passed with the `AnimatedSwitcher` deleted outright).

**Validated at `acd0ea7df`:** both suites green, `flutter analyze` clean, all 14 checks green. Verified by mutation rather than by reading: removing `_slowFeedLoadReported` fails the new test with 3 events instead of 1, and deleting the `AnimatedSwitcher` fails the transition assertion. Also drove the switch under rapid double-toggle, backgrounding, and a concurrent `loadMore`, and found nothing wrong.

**Remaining uncertainty, stated plainly:**

- `b30e1ac25` extended a pre-existing re-fire on the variant-switch path and I had seen the risk and dismissed it. @rabble caught it, and `acd0ea7df` fixes it better than my commit ever did.
- Not holding on #5516. Nobody argued for the old behaviour and the jank being fixed is real, so I am treating the reversal as decided rather than as an open question. #5516 has a note pointing at this PR so the stale criterion stops reading as current.

**Scope:** this covers the findings I raised and the commits I pushed. It does not speak to @rabble's `CHANGES_REQUESTED`, which still stands against `78889813e` and is his to clear. All three of his items look resolved to me at head, including the `build`-assignment one, which he offered to accept as a documented decision and which is now documented.

Merge when you are happy with the two commits I pushed.

rabble (approved):
Rabble review: cross-fade fix verified — variant switch now keeps _rendered page and cross-fades via AnimatedSwitcher (250ms) instead of dropping to spinner. _feedStateTransitionDuration and _slowFeedLoadThresholdMs correctly overridable for fakeAsync. No regressions, 2 files, 5 commits. LGTM.

rabble (approved):
## Review: fix(explore): cross-fade the popular feed on variant switch

### Summary
Solid fix for the New ↔ Classic jump. Root cause analysis is correct: resetting `popularVideosLoadedVariantProvider` to `null` on switch made `PopularVideosTab` fall through to the full-screen spinner, which also unmounted the toggle. The held-page + `AnimatedSwitcher` approach is the right tradeoff — preserves the marker guarantee while keeping the header mounted.

### Verification
- `flutter test test/widgets/popular_videos_tab_test.dart` — **2/2 passed**
  - `keeps the loaded grid on screen while the other variant loads` verifies no `BrandedLoadingIndicator` during switch, distinct `ValueKey(variant)` keys, and that `AnimatedSwitcher` keeps both children mounted mid-transition (`pump(Duration 120ms)`).
  - `reports a slow variant switch once, not once per rebuild` verifies `_slowFeedLoadReported` guard with `slowLoadThresholdMs: 0`.
- `flutter test test/providers/explore_feed_refresh_retention_test.dart` — **13/13 passed** — loaded-variant contract preserved.
- `flutter analyze lib/widgets/popular_videos_tab.dart` — **No issues**.

### What’s good
- **State handling**: capturing `_rendered = (variant, state)` inside build is justified and well-commented (avoids one-frame delay vs `ref.listen`). Keying held page by its *own* variant correctly handles rapid toggle (provider retains previous AsyncValue).
- **Keying**: `ValueKey(rendered.variant)` / `ValueKey('error')` / `ValueKey('loading')` gives `AnimatedSwitcher` distinct children for every transition (loading↔data, error↔data, native↔classic).
- **a11y**: `MediaQuery.disableAnimationsOf(context)` collapses duration to zero — matches existing usage in `inbox_view.dart:641`, `vine_bottom_nav.dart:402`, etc.
- **Code style**: Extracts `_feedStateTransitionDuration` / `_slowFeedLoadThresholdMs` constants; splits `_buildDataState` → `_trackDataState` (tracking only) with widget built inline — complies with widgets-over-methods (`rules/code_style.md`).
- **Telemetry**: `_trackSlowLoadIfNeeded` now runs from both held-page and loading branches with single-emission guard — fixes dropped slow signal during variant switch.
- **Tests**: Follow `test/helpers/test_provider_overrides.dart` (`testMaterialApp` with l10n delegates, proper mocks); `registerFallbackValue(PopularVideosVariant.native)` present.

### Nits (non-blocking)
- Side-effects in `build` (`_trackDataState` / `_trackErrorState` / `_trackSlowLoadIfNeeded`) — pre-existing pattern in this file, not introduced here. Ideal would be `ref.listen` for analytics, but current approach is consistent with codebase.
- Assignment to `_rendered` inside `build` is a mutable state mutation during build. Safe here because value is derived from watched providers, but worth a follow-up comment if linter ever flags `use_setState_synchronously`. No change requested.
- No progress affordance during switch (only toggle pill moves) — explicitly out of scope and noted; cold-cache case can be followed up.

### Conclusion
No blocking issues. Approving — please rebase onto fresh `origin/main` before merge per `rules/agent_workflow.md`.

PR #6800, opened 2026-08-06
fix(library): animate the clip preview hero back on dismiss

### LINKED ISSUE #6801: fix(library): animate the clip preview hero back on dismiss
## Problem

Closing a clip preview from the library grid did not animate the thumbnail back into its grid cell once the video had rendered its first frame. The preview overlay faded out instead.

## Root Cause

The preview `Hero` lived inside `DivineVideoPlayer`'s placeholder. That placeholder is replaced with `SizedBox.shrink()` after the first frame renders, so by dismiss time the preview route no longer had a matching destination hero. While mounted, the hero was also below the `FittedBox`, so Flutter measured the native video dimensions instead of the visible preview box.

Stop-motion clips had the same missing-return-flight behavior because they render through `StopMotionPlayer` and never entered the video placeholder branch.

## Expected Behavior

Opening and dismissing normal video clips and stop-motion clips from the library grid should animate the still thumbnail between the grid cell and preview box. The native video surface should not be reparented into the hero overlay during the flight.

## Fix

PR #6800 keeps the hero mounted directly under the preview clip, uses a still-thumbnail shuttle for the flight, bounds still-image decodes, and covers the behavior with widget tests.

hm21:
Closes #6801

## Description

Closing the clip preview no longer animated the thumbnail back into the library grid — the overlay just faded out.

The `Hero` was rendered inside `DivineVideoPlayer`'s `placeholder`, and that placeholder is removed from the tree (`SizedBox.shrink()`) once the first video frame is decoded. Opening still animated because the placeholder is up while the player initializes; by the time the user dismissed the preview there was no hero left to fly back from. It also sat below the `FittedBox`, so even while mounted its rect was the video's native pixel size instead of the visible preview box. Regression from #2691, which moved the hero into the placeholder when the preview was ported to `DivineVideoPlayer`.

The hero now wraps the media directly under the preview's `ClipRRect`, where it stays mounted for the preview's whole lifetime and measures the box the user actually sees. A `flightShuttleBuilder` renders the still thumbnail for both directions — otherwise the push flight would reparent the live platform view / texture into the hero overlay, which is exactly where native surfaces misbehave.

Player construction moved into a `_ClipMedia` widget: keeping it inline under the extra hero level pushed the tree past readable nesting, and it follows the repo's prefer-widget-classes-over-inline-builders rule.

Stop-motion clips get the animation too. They render through `StopMotionPlayer` and never reached the placeholder branch, so the preview had no hero at all even though the grid card sets one.

The still decode is bounded because a stop-motion clip's thumbnail is `frames.first.path` — a full-resolution camera still. The shuttle, the video placeholder and `StopMotionPlayer` now share one bound, so the preview resolves to a single cache entry instead of three. `ResizeImage` never upscales, so small video thumbnails are unaffected.

Takeover follow-up centralized the grid/preview hero tag, made the shuttle decorative for semantics, and reused the grid card's missing-thumbnail placeholder so the fallback icon stays stable during the flight.

## Verification

- `flutter analyze lib/widgets/video_clip test/widgets/video_clip` — clean
- `flutter test test/widgets/video_clip/ test/widgets/library/clips_tab_test.dart test/widgets/stop_motion` — 43 passed
- New regression test verified to fail without the fix (finds 0 heroes once the first frame renders) and pass with it
- Stop-motion preview test asserts the hero shuttle uses the bounded thumbnail, excludes semantics, and preserves the missing-thumbnail placeholder
- Manually tested on a real Samsung Galaxy: open a clip from the library grid, let the video start playing, dismiss — thumbnail flies back into its grid cell. Checked both tap and long-press entry, and a stop-motion clip.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/widgets/video_clip/video_clip_hero.dart
+++ b/mobile/lib/widgets/video_clip/video_clip_hero.dart
@@ -1,0 +1,20 @@
+// ABOUTME: Shared Hero helpers for video clip grid-to-preview transitions
+// ABOUTME: Keeps tags and decorative fallback thumbnails consistent
+
+import 'package:divine_ui/divine_ui.dart';
+import 'package:flutter/material.dart';
+
+String videoClipPreviewHeroTag(String clipId) => 'Video-Clip-Preview-$clipId';
+
+class VideoClipThumbnailPlaceholder extends StatelessWidget {
+  const VideoClipThumbnailPlaceholder({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return DivineIcon(
+      icon: DivineIconName.videoCamera,
+      color: context.vineColors.mutedText,
+      size: 32,
+    );
+  }
+}
--- a/mobile/lib/widgets/video_clip/video_clip_preview.dart
+++ b/mobile/lib/widgets/video_clip/video_clip_preview.dart
@@ -197,62 +197,12 @@ class _VideoClipPreviewSheetState extends ConsumerState<VideoClipPreview> {
                       aspectRatio: widget.clip.targetAspectRatio.value,
                       child: ClipRRect(
                         borderRadius: .circular(16),
-                        child: Builder(
-                          builder: (context) {
-                            if (widget.clip.stopMotionFrames
-                                case final frames?) {
-                              return StopMotionPlayer(
-                                frames: frames,
-                                cacheHeight:
-                                    (MediaQuery.sizeOf(context).height *
-                                            MediaQuery.devicePixelRatioOf(
-                                              context,
-                                            ))
-                                        .round(),
-                              );
-                            }
-
-                            final vw = _controller?.state.videoWidth ?? 0;
-                            final vh = _controller?.state.videoHeight ?? 0;
-
-                            final player = DivineVideoPlayer(
-                              controller: _controller,
-                              placeholder: Stack(
-                                fit: .expand,
-                                children: [
-                                  // Thumbnail
-                                  if (widget.clip.thumbnailPath != null)
-                                    Hero(
-                                      tag:
-                                          'Video-Clip-Preview-${widget.clip.id}',
-                                      child: ClipThumbnailImage(
-                                        path: widget.clip.thumbnailPath!,
-                                        fit: BoxFit.cover,
-                                      ),
-                                    ),
-
-                                  // Progress-indicator
-                                  const Center(
-                                    child: CircularProgressIndicator(
-                                      color: VineTheme.vineGreen,
-                                    ),
-                                  ),
-                                ],
-                              ),
-                            );
-
-                            if (vw == 0 || vh == 0) return player;
-
-                            return FittedBox(
-                              fit: BoxFit.cover,
-                              clipBehavior: Clip.hardEdge,
-                              child: SizedBox(
-                                width: vw.toDouble(),
-                                height: vh.toDouble(),
-                                child: player,
-                              ),
-                            );
-                          },
+                        child: _ClipHero(
+                          clip: widget.clip,
+                          child: _ClipMedia(
+                            clip: widget.clip,
+                            controller: _controller,
+                          ),
                         ),
                       ),
                     ),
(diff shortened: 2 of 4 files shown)


NotThatKindOfDrLiz:
@hm21 takeover update pushed in `6e399b8fa`.

What changed:
- Centralized the grid/preview hero tag via `videoClipPreviewHeroTag`.
- Made the hero shuttle decorative for semantics and reused the grid card's missing-thumbnail placeholder so the fallback icon does not change mid-flight.
- Added a stop-motion preview regression test that asserts the shuttle uses the bounded thumbnail, excludes semantics, and preserves the placeholder.
- Updated #6801 and this PR body so the linked issue and verification notes are current.

Local verification passed:
- `flutter analyze lib/widgets/video_clip test/widgets/video_clip`
- `flutter test test/widgets/video_clip/ test/widgets/library/clips_tab_test.dart test/widgets/stop_motion`
- Pre-push hook analyzer and changed-file tests also passed.

CI state:
- The old red checks were tied to the previous head SHA and the GitHub Actions action-download outage.
- After this push, only CLA appeared on the PR check rollup; the pull-request workflows did not enqueue for the new SHA.
- I manually dispatched Mobile CI on the current head: https://github.com/divinevideo/divine-mobile/actions/runs/31126842551. It is still queued at handoff.

Remaining before merge:
- Let the queued Mobile CI run complete, and rerun/unstick the PR-only workflows if GitHub still does not enqueue them.
- Please do the iOS open/dismiss spot-check for a normal video clip and a stop-motion clip. I could not perform that device pass from here.
- Review the takeover commit and merge when you are satisfied with the CI/manual result.

rabble (approved):
## Overview

Fixes the grid → preview hero not flying back on dismiss. The diagnosis in the `_ClipHero` doc is the right one and worth quoting:

> The hero used to sit inside `DivineVideoPlayer`'s placeholder, which is dropped from the tree once the first video frame renders — so by the time the user closed the preview there was no hero left to fly back from.

That's a genuinely non-obvious lifetime bug (the hero worked *inbound* because the placeholder was still up, and only failed outbound), and the fix — hoist the `Hero` to wrap the whole media subtree so it outlives the placeholder — addresses the mechanism rather than the symptom.

Other things done right:

- **`flightShuttleBuilder` renders the still in both directions**, which keeps the native video surface out of the hero overlay. Flying a platform view is the usual next bug after this one; heading it off in the same change is the correct call.
- **The 100-line `Builder` is decomposed into `_ClipHero` / `_ClipMedia`**, replacing a closure that was doing three jobs. Matches the repo's widgets-over-builders preference.
- **`videoClipPreviewHeroTag(clipId)`** replaces the string literal that was duplicated between the grid card and the preview — exactly the kind of pair that silently breaks a hero when one side is edited.
- **`_stillCacheHeight` is a real fix, not tidying.** A stop-motion thumbnail is a full-resolution camera frame; the shuttle would have decoded it unbounded. Sharing one bound across shuttle, placeholder and `StopMotionPlayer` keeps them on a single cache entry.
- The test's comment about why `pumpAndSettle` can't be used ("the pre-fix placeholder owns a `CircularProgressIndicator` that never settles") saves the next person a confusing ten minutes, and the fake-first-frame harness means the test actually reproduces the unmount that caused the bug rather than approximating it.

---

### CI has not actually run on this branch — please re-push before merging

Worth stating plainly since it's easy to misread as a code failure. Across the three runs:

- **`b7b4b91`** — `Tests (shard 3/4)` failed with `Failed to resolve action download info. Error: Service Unavailable` / `Internal Server Error`. GitHub Actions infrastructure, not a test.
- **`7277b23`** — everything failed or cancelled together, same pattern.
- **`6e399b8` (head)** — `Detect App CI Scope` cancelled and every downstream job (`Analyze`, `Format`, `Generated Files`, `Tests`, `VGV Tag Gate`) reported `skipped`. The aggregate `Mobile CI` gate then failed because its upstream jobs didn't complete.

So there is no evidence of a real failure, and equally no evidence the new tests pass in CI. An empty push to re-trigger should sort it; I'd want a green head before this merges.

---

### Two things worth checking by hand

**1. Two heroes with the same tag, both mounted.** The grid card's `_Thumbnail` hero and the preview's `_ClipHero` now use the same tag, and — unlike before — the preview's hero stays mounted for the sheet's entire lifetime rather than being dropped with the placeholder. Flutter asserts on multiple heroes sharing a tag *within one Navigator subtree*. If `VideoClipPreview` is always presented as its own route this is fine (heroes are matched across routes, not within one). If any caller ever shows it as an in-route overlay instead, that's now a hard assertion failure where before the accidental unmount hid it. Worth confirming every presentation path pushes a route.

**2. The flight starts with a decode miss.** `_stillCacheHeight` bounds to the *preview* size, while the grid card bounds to its cell. The doc's claim is accurate as written (shuttle, placeholder and `StopMotionPlayer` share the bound) — but it means the shuttle asks for a size the grid never decoded, so the first frame of the flight is a fresh full-height decode of a camera-resolution still. That's the frame budget you can least afford during an animation. If it's visible on a mid-range device, having the shuttle reuse the grid's bound while the *destination* uses the preview bound would trade a slightly soft shuttle for a smooth start.

### Nit

`VideoClipThumbnailPlaceholder` hardcodes `size: 32`. Everything else in the new file goes through `divine_ui` / `context.vineColors`, so the one literal stands out — if there's a spacing or icon-size token that fits, it'd match the surrounding style.

rabble:

PR #6803, opened 2026-08-06
fix(audio): unblock publishing for sounds whose creator allows reuse

### LINKED ISSUE #6814: fix(audio): unblock publishing for sounds whose creator allows reuse
Auto-created from PR #6803.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6803
**Branch:** `fix/audio-reuse-consent-instance-id`

hm21:
Closes #6814

## Description

A video whose sound came from the music library could upload fine and then never post. The publish stopped with `Selected audio does not permit reuse; blocking video publish`, and the user saw *"The video uploaded but the post could not be published. Check your relay settings and try again."* — pointing at the relay for a problem no relay setting can fix. Retrying re-ran the same refusal.

Two separate defects, one report.

**1. The consent lookup used the wrong id.** `AudioReuseConsentResolver` resolves reuse for a legacy Kind 1063 sound (no `allow_audio_reuse` tag) by finding the videos that reference it and reading the newest source revision's consent. It queried `AudioEvent.id` — but the editor stamps a `-<timestamp>` uniqueness suffix onto every timeline track, so the queried id matched no `["e", …, "audio"]` tag, the lookup came back empty, and the fail-closed branch refused the publish.

The sound picker gates on the same resolver, but it runs *before* the suffix is stamped. So a sound passed selection under its clean id and was then refused at publish under the suffixed one — from a real session log, the same sound resolved to `Found 1 videos using audio` at selection and `Found 0 videos using audio` at publish. Any legacy sound picked from the music library was affected.

Now it queries `AudioEvent.attributionEventId`, which is the id the publisher already writes into the attribution tag and which resolves the `video_` prefix plus the editor's uniqueness and duplicate suffixes. Picker and publisher finally agree on one id.

**2. The refusal was indistinguishable from a relay failure.** `publishVideoEvent` folded it into `return false`, and `return false` can only classify as `nostrPublishFailed`. It now raises a typed `AudioReuseNotPermittedException` mapped to a new `PublishErrorKind.audioReuseNotPermitted`.

Two non-obvious pieces there:

- The type branch (`classifyPublishErrorObject`) has to run *ahead* of `classifyPublishErrorMessage`. The exception's message interpolates the 64-character hex audio id, and hex digits cover every HTTP status the substring list looks for: `404` becomes `serverNotFound`, `500` `serverInternalError`, `502` and `503` `serverDown`, `401` `notSignedIn`, `403` `forbidden`, `413` `fileTooLarge`, `429` `rateLimited`. That is roughly one id in nine, and `serverNotFound` even names the Blossom host in the copy. The end-to-end test uses a colliding id, so moving the type branch below the substring pass turns it red.
- The publisher needs an explicit `on AudioReuseNotPermittedException { rethrow; }` arm, because the generic `catch` wrapping the whole publish turns every error into `false` — a bare `throw` would have been swallowed and changed nothing.

Fix 2 keeps its value after fix 1: consent can still be revoked between selecting a sound and posting, and that case now reads correctly instead of blaming the network.

**3. Only evidence that needs no relay makes the refusal terminal.** An earlier revision of this PR tried to split "refused" from "could not verify" inside the resolver. Review showed the boolean cannot carry that distinction: `getVideosByIds` runs the app-wide `VideosRepository`, whose block and content filters (NSFW/moderation, Divine-hosted-only, feed aspect ratio) can drop the source video, so muting a creator made their explicitly reusable sound report as withheld; `Filter(limit: 50)` is truncated newest-first while the source video is by construction the *oldest* event carrying the audio `e` tag, so past 50 reusers it falls out of the window deterministically; a relay refusing with `CLOSED` completes the query normally; and `getVideosByIds` returns a partial set silently, so a missing newest revision could leave a stale granting revision in charge.

So `AudioReuseNotPermittedException` is raised only for `hasExplicitReuseConsent` — a refusal stamped on the sound's own event, with no relay in the way. Everything the legacy source-video resolver cannot confirm stays `nostrPublishFailed`, which is retryable and honest about what is known. `verify()` is fail-closed again, and its doc now says that `false` means "not verified", never "the creator said no".

The resolver's existing test fixtures used a non-hex `audio-event` id, which no real Kind 1063 event has, so they could not have caught this. They now use a real 64-hex id, plus new cases for the timeline suffix, the `video_` prefix, and a sound with no referenceable event id.

**Related Issue:** 

## Out of Scope

The upload-failure sheet still offers **Try Again** for `audioReuseNotPermitted`, where a retry can never succeed. After fixes 1 and 3 this state is reached only on a relay-confirmed refusal, and the new copy names the way out ("Pick a different sound to post it"), so gating the button per error kind is left as a follow-up rather than pulled into this diff. The sheet has the same unconditional Try Again for `accountChanged` and `notSignedIn` today, so the fix belongs there as one change rather than a special case for this kind.

## Verification

- `flutter test` over the resolver, publisher, publish-service, publish-provider, background-publish, upload-failure-sheet, audio-editor, sound-detail and l10n suites — 448 passing
- Each new guard mutation-checked: reverting the type-first ordering, the `unverified` split, or the picker's catch turns its test red
- `flutter gen-l10n` re-run; new key translated in all 22 locales (`arb_consistency_test.dart` green)
- `dart format` clean on every changed file
- Manual verification on a real Samsung Galaxy: **passed**

## Type of Change


mobile/lib/services/audio_reuse_consent_resolver.dart
```diff
@@ -15,40 +15,56 @@ class AudioReuseConsentResolver {
...
+    // Query the id reusing videos tag, not [AudioEvent.id]: an editor timeline
+    // track carries a `-<timestamp>` uniqueness suffix and an original sound a
+    // `video_` prefix, so the raw id matches no `["e", …, "audio"]` tag. Under
+    // that id every legacy sound failed closed the moment it reached the
+    // timeline — including ones the picker had already verified.
+    final audioEventId = sound.attributionEventId;
+    if (audioEventId == null) return false;
+
+    final ids = await _soundsRepository.fetchVideosUsingSound(audioEventId);
+    if (ids.isEmpty) return false;
```
realmeylisdev: This is the case fix 3 is meant to catch, and it lands here instead.

`fetchVideosUsingSound` only rethrows what `queryEvents` throws — and `queryEvents` doesn't throw on a dead relay. It calls `queryEventsDetailed`, which computes `timedOut` and `noRelays`, then returns `result.events` and drops both (`nostr_client.dart:916`). Underneath, the SDK has already swallowed the `TimeoutException` into an empty box (`nostr.dart:366-371`); disposed client and closed pool return `events: []` the same way.

So an offline or timed-out publish arrives here as `ids == []` → `false` → `refused` → `AudioReuseNotPermittedException` → *"The video uploaded, but its sound isn't cleared for reuse. Pick a different sound to post it."*

On `main` the same input returned `false` → `nostrPublishFailed` → *"Check your relay settings and try again."* So someone with a perfectly reusable sound and a flaky relay is now told the creator withheld it, and sent to swap a sound that was fine.

The test for this (`propagates a failed lookup instead of answering false`) stubs `thenThrow(StateError('relay unavailable'))` — a throw the real repository doesn't produce for any of these modes, so it's green against a path production can't take.

Repro, green on this branch:

```dart
// _MockNostr.queryEventsDetailed returns (events: [], timedOut: true),
// mirroring nostr.dart:366-371.
client = NostrClient.forTesting(nostr: nostr, relayManager: relayManager);
when(() => relayManager.connectedRelays).thenReturn(<String>[]);

expect(await client.queryEvents([Filter(kinds: const [34236])]), isEmpty);

final repo = SoundsRepository(nostrClient: client);
expect(await repo.fetchVideosUsingSound('b' * 64), isEmpty); // no throw

final resolver = AudioReuseConsentResolver(
  soundsRepository: repo,
  videosRepository: _MockVideosRepository(),
);
// A legacy sound: allowsReuse false, no explicit tag, has sourceVideoReference.
expect(await resolver.verify(legacySound), isFalse); // => refused, not unverified
```

`queryEventsDetailed` already carries the signal — it exists, it's just dropped one layer up. How you thread it back through `SoundsRepository` is a call I'd rather leave with you than push.

Worth deciding at the same time: `matching.isEmpty` on line 62 collapses the same way. `fetchVideosUsingSound` is capped at 50 and merged newest-first, and the source video is necessarily the *oldest* event tagging `e:[<audio id>]` — so a legacy sound reused by 50+ videos drops its own source out of the window and also reads as a refusal. Both branches mean "no evidence", and neither means "the creator said no".

mobile/lib/services/video_event_publisher.dart
```diff
@@ -216,35 +240,41 @@ class VideoEventPublisher {
...
 
     final currentPubkey = _authService?.currentPublicKeyHex;
     if (currentPubkey != null && currentPubkey == sound.pubkey) {
-      return true;
+      return AudioReuseCheck.permitted;
     }
 
     final checker = _audioReuseConsentChecker;
-    if (checker == null) return false;
+    if (checker == null) return AudioReuseCheck.unverified;
```
realmeylisdev: This branch lost its only test in this diff.

On `main`, `blocks selected audio when the source explicitly forbids reuse` ran the bare `publisher` with no checker wired, so it came through here. It now goes through `publisherWithConsent()`, which always injects one.

Flipping this line to `AudioReuseCheck.permitted` — letting a private sound be remixed with no consent check at all — leaves all 34 tests in the file green. Since this is the fail-closed guard the `AudioReuseCheck` doc calls out by name, it's worth a case that builds the publisher without a checker and asserts the publish is blocked.

mobile/lib/services/video_publish/video_publish_service.dart
```diff
@@ -963,6 +968,28 @@ class VideoPublishService {
...
+  ///
+  /// These are deliberate refusals rather than transport failures, so they are
+  /// matched on type and must be consulted *before*
+  /// [classifyPublishErrorMessage].
+  ///
+  /// The precedence is load-bearing: a refusal's message interpolates a
+  /// 64-character hex Nostr id, and hex digits include every digit the
+  /// matcher reads as an HTTP status — an id containing `404`, `500`, `502`,
+  /// `503`, `401`, `403`, `413` or `429` classifies as `serverNotFound`,
+  /// `serverDown`, `notSignedIn`, `forbidden`, `fileTooLarge` or
```
realmeylisdev: Small thing: `500` maps to `serverInternalError`, not `serverDown` — `502`/`503` are the `serverDown` pair. The collision test just below gets it right; it's only this comment that's off.

realmeylisdev (changes requested):
Fix 1 (the id) and fix 2 (the typed refusal) both hold up. I moved the type branch below the substring pass and the collision test goes red, so that ordering is properly pinned — nice catch on the hex/HTTP-status overlap.

Fix 3 doesn't fire. `queryEvents` never throws on a dead relay, so the `refused` / `unverified` split can't see the case it was written for: a relay timeout now shows the terminal "pick a different sound" copy where `main` showed the retryable one. Details inline, with a repro.

One more: the fail-closed `checker == null` guard lost its only test in this diff.

Verified locally on this branch — `flutter analyze lib test integration_test` clean, 426 tests green across the suites you listed, all 22 locales translated, and the `pek1:` persistence is name-keyed so the new enum member is safe to insert mid-list.

mobile/lib/services/audio_reuse_consent_resolver.dart
hm21: Confirmed and fixed in a950fea44.

You are right on every step. `nostr.dart:365-369` catches the `TimeoutException` and returns `(events: eventBox.all(), timedOut: true)`, `queryEventsDetailed` passes `timedOut`/`noRelays` up, and `queryEvents` returns `result.events` and drops both. So the split could only ever fire on a genuine repository throw, and none of these modes produce one — the terminal copy stayed for exactly the user the fix was written for. My test proved nothing, for the same reason I flagged the vacuous precedence test earlier.

`SoundsRepository` now has `fetchVideosUsingSoundDetailed`, which keeps the `answered` flag (`!timedOut && !noRelays`); `fetchVideosUsingSound` delegates to it, so its other caller is unchanged. The resolver throws `AudioReuseConsentUnavailableException` when an empty answer came from relays that never answered.

Two things your repro pointed at that I extended:
- A cached hit can fill `ids` while `answered` is false. That is real evidence, so it is still used — covered by `uses a cached answer the relays did not confirm`.
- `getVideosByIds` goes through the same `queryEvents`, so it swallows a dead relay into `[]` too. If the relays just named N videos and none can be fetched, that is the same silence one layer down, not a revocation — it now throws as well.

`sounds_repository` stays at 100% line coverage (68 tests).

mobile/lib/services/video_event_publisher.dart
hm21: Right, I moved that coverage without replacing it. Fixed in bb2f71bda.

Verified your mutation first: flipping the branch to `AudioReuseCheck.permitted` left all 34 tests green. The new case builds the bare `publisher` from setUp, asserts the publish returns false and emits no audio tag, and goes red under that same flip.

mobile/lib/services/video_publish/video_publish_service.dart
hm21: Correct — fixed in 9893ad294. The comment listed six kinds for eight codes and dropped `serverInternalError` entirely. It now spells the mapping out per code. The PR description carried the same wrong list; updating it too.

mobile/lib/services/audio_reuse_consent_resolver.dart
```diff
@@ -15,40 +16,71 @@ class AudioReuseConsentResolver {
...
+    );
+    if (lookup.ids.isEmpty) {
+      if (!lookup.answered) {
+        throw AudioReuseConsentUnavailableException(audioEventId);
       }
-      return matching.first.allowAudioReuse;
-    } catch (_) {
       return false;
     }
+    final candidates = await _videosRepository.getVideosByIds(
```
mbradley: This reads through the app-wide `VideosRepository`, which is constructed at `video_providers.dart:515` with `blockFilter: createBlockedAuthorFilter(ref)` and a `contentFilter` combining the NSFW/moderation filter, "Divine-hosted only", and `feedAspectRatioPreference.shouldHideVideo`. `getVideosByIds` runs every event through `_tryParseAndFilter` with `ignoreBlockFilter` defaulting to false (`videos_repository.dart:1562`).

So the source video can be dropped for reasons that say nothing about consent, and `matching.isEmpty` then returns false. Concretely: mute a creator, and their sound — which they explicitly marked reusable — now reports as not cleared for reuse. Your own blocklist silently revokes your access to consented content, and the copy attributes it to the creator's terms. The same happens with "Divine-hosted only" against an externally hosted source, and with an aspect-ratio preference that hides the source video. A cosmetic feed setting becomes an input to a licensing decision.

I traced the wiring end to end rather than inferring it: both the picker's `audioReuseConsentProvider` and the publisher's `audioReuseConsentChecker` resolve through this same filtered instance.

`ignoreBlockFilter: true` already exists at `videos_repository.dart:1552`; the content filter would need a path that does not apply it. Consent is a fact about the creator, not about the viewer's feed settings. How you separate them is your call.

mobile/lib/services/audio_reuse_consent_resolver.dart
```diff
@@ -15,40 +16,71 @@ class AudioReuseConsentResolver {
...
+      lookup.ids,
+      hydrateBulkStats: false,
+    );
+    // The relays just named these videos, so failing to fetch a single one is
+    // the same silence one layer down — not a revocation.
+    if (candidates.isEmpty) {
```
mbradley: The comment reasons about failing to fetch *a single one*, but the guard only fires when every fetch fails. `getVideosByIds` returns a partial set silently (`videos_repository.dart:1308-1313`).

Two consequences, in opposite directions. If the missing id is the source revision, `matching` is empty and you get a terminal refusal. If only the **newest** revision is missing, `matching.first` is a stale granting revision and line 84 returns its `allowAudioReuse` — permitting a reuse the creator revoked.

That second one is the only path in the feature that errs against the creator rather than the publisher, which is why I would rather it were decided than left to the comment. It is also the reason I am requesting changes rather than commenting.

Either widen to `candidates.length < lookup.ids.length` so any shortfall throws, or reword so the comment does not claim coverage the code does not have.

mobile/packages/sounds_repository/lib/src/sounds_repository.dart
```diff
@@ -375,23 +400,24 @@ class SoundsRepository {
...
         Filter(
           kinds: const [NIP71VideoKinds.addressableShortVideo],
           e: [audioEventId],
           limit: limit,
         ),
       ]);
 
-      final videoIds = events.map((e) => e.id).toList();
+      final videoIds = result.events.map((e) => e.id).toList();
+      final answered = !result.timedOut && !result.noRelays;
```
mbradley: The doc above says `answered` is false when no relay was connected or the query timed out. Two shapes make it true while the result is not authoritative.

A `CLOSED` frame is terminal: `relay_pool.dart:1451` calls `_fireQueryCompleteIfSettled(afterTerminalFrame: true)`, and the doc at `:751` is explicit that `CLOSED` means the relay ended the subscription **without finishing**. So every relay refusing with `CLOSED` — "too many concurrent REQs", say — completes normally with `timedOut: false`, `noRelays: false` and zero events, which lands on the terminal copy for a pure transport failure.

`_armQuerySettleWindow` (`relay_pool.dart:788`) does the same for a partial fan-out: once any relay sends a terminal frame, the silent ones are bounded by `querySettleWindow` and the query completes on whoever answered. One fast relay that does not carry the source video is enough.

There is a third, narrower one: `nostr_client.dart:852` returns `(events: [], timedOut: false)` when `_queryPool.isClosed`, while `noRelays` was computed from still-connected relays.

That makes "the relays answered, and no later attempt changes that" weaker than the design needs — and for the same class of failure fix 3 exists to eliminate.

mbradley (changes requested):
Verified at `64991546a` with the resolver, publisher, publish-service and picker suites green locally, plus a read of the full diff rather than only the four commits since the last review. The id fix and the typed refusal both hold up, and the fail-closed guard restored in `bb2f71bda` goes red when I flip the branch to `permitted`, so that coverage is real rather than re-added text.

I pushed one commit for the finding with an unambiguous fix. It is yours to amend or revert.

`64991546a` — the picker's fail-closed test stubbed `fetchVideosUsingSound`, but the resolver now calls `fetchVideosUsingSoundDetailed`. The stub was dead: deleting all three lines left the test green, because the unstubbed method returned null and the resulting cast error was what the picker's catch swallowed. It now stubs the detailed method with an unanswered empty result, which is what an unreachable relay actually produces. Verified red when the catch is flipped to fail open. CI is green on it.

The rest needs your judgment. Making a refusal terminal raises the cost of every false negative in `verify()`, and there are four reachable ones. Three are inline; the fourth is here because it lands in a file this PR does not touch.

**Removing the resolver's internal catch enrolled two providers in Riverpod's retry ladder.** `verify()` now throws out of a `@riverpod` build. Riverpod 3.3.2's `defaultRetry` bails only on `ProviderException` and `Error` (`provider_container.dart:833`), and `AudioReuseConsentUnavailableException implements Exception`, so it retries: 10 attempts, 200 ms doubling to 6.4 s. `metadata_sounds_section.dart:285` watches `audioReuseTermsProvider(audio)` per sound, and that provider calls `verify()` on the same path, so during a relay outage each visible sound costs a full ladder rather than one query. The picker is unaffected, since `ref.read(...future)` with no listener disposes first. The same root cause changes the sound-detail terms badge: an error used to resolve to `false` and render "Credit only", and now resolves to null and renders nothing. Arguably an improvement, but undocumented.

**realmeylisdev's closing point about the 50-event window is still open.** The thread is resolved, but that half was not answered — the reply covers the `answered` flag, the cached hit, and `candidates.isEmpty`, which is the adjacent branch rather than this one. Their point is about `matching.isEmpty`: the source video is outside the query window, so it is never named. `candidates.isEmpty` is about a video that *was* named and could not be fetched. Different causes, and the fix for one does nothing for the other.

I confirmed the mechanism so it is not just a hypothesis now: `Filter(limit: 50)` is a single filter, so `queryEventsDetailed` forwards the limit to `_mergeEvents`, which sorts `b.createdAt - a.createdAt` and truncates. The source video is by construction the oldest event carrying the audio `e` tag. Past 50 reusers it falls out of the window and the sound reads as withheld, deterministically, and more certainly the more popular the sound gets.

**On the Out of Scope note.** The parity with `accountChanged` and `notSignedIn` is literally true, since Try Again is unconditional for every kind. But a retry after signing back in or switching accounts succeeds. This is the first kind where a retry provably cannot, which is what makes it a different problem rather than the same one. Worth deciding now rather than inheriting the parity — the copy points at a remedy that lives behind the secondary button while the primary one guarantees another failure.

Worth saying plainly: this is a large net improvement on `main`, where every legacy sound failed closed. The findings are about the new terminal classification outrunning the evidence behind it.

mobile/lib/services/audio_reuse_consent_resolver.dart
hm21: Confirmed and it decided the outcome. I traced the wiring rather than taking it on the claim: `videosRepositoryProvider` is built at `video_providers.dart:516-524` with `blockFilter: createBlockedAuthorFilter(ref)` and a `contentFilter` combining the NSFW/moderation filter, Divine-hosted-only and `feedAspectRatioPreference.shouldHideVideo`, and both `sounds_providers.dart:165/188` and the publisher's checker resolve through that same instance. `_tryParseAndFilter` defaults `ignoreBlockFilter` to false, so muting a creator did make their explicitly reusable sound report as withheld.

That a cosmetic feed setting could feed a licensing decision is what convinced me the boolean cannot carry a terminal claim at all.

Resolved by removing the split rather than patching it — see 8fdf36232. The terminal claim now rests only on `hasExplicitReuseConsent`, read off the sound's own event with no relay involved; everything the legacy resolver cannot confirm goes back to the retryable `nostrPublishFailed`. `verify()` is fail-closed again and its doc says so, so this finding no longer reaches the user as a verdict about the creator.

The filter/consent separation you describe is worth doing, but as its own change to the shared repository rather than folded into this one.

mobile/lib/services/audio_reuse_consent_resolver.dart
hm21: Right on both directions, and the second one is the reason this went out rather than getting widened. `getVideosByIds` returns a partial set silently (`videos_repository.dart:1306-1313`), so a missing newest revision left `matching.first` on a stale granting one and permitted a reuse the creator had revoked — the only path in the feature that errs against the creator.

I did not widen to `candidates.length < lookup.ids.length`: that would have made the guard honest about *this* shortfall while the same stale-revision risk stays reachable through the block and content filters, which drop events for reasons that have nothing to do with the lookup. Rewording would have documented the gap without closing it.

Resolved by removing the split rather than patching it — see 8fdf36232. The terminal claim now rests only on `hasExplicitReuseConsent`, read off the sound's own event with no relay involved; everything the legacy resolver cannot confirm goes back to the retryable `nostrPublishFailed`. `verify()` is fail-closed again and its doc says so, so this finding no longer reaches the user as a verdict about the creator.

The stale-revision hazard itself is pre-existing on `main` and survives this PR — it now yields a retryable failure instead of a wrong verdict, which is containment, not a fix. Worth its own issue.

mobile/packages/sounds_repository/lib/src/sounds_repository.dart
hm21: Verified: `CLOSED` routes through `_fireQueryCompleteIfSettled(afterTerminalFrame: true)` at `relay_pool.dart:1452`, and the doc at `:751` says outright that it means the relay ended the subscription without finishing — so a relay refusing with "too many concurrent REQs" produced `timedOut: false, noRelays: false` and zero events. `_armQuerySettleWindow` does the same for a partial fan-out.

You were right that this is the same class of failure fix 3 existed to eliminate, which is what made the flag the wrong foundation. `fetchVideosUsingSoundDetailed` is gone again in 8fdf36232 and `sounds_repository` is back at its `main` state.

Resolved by removing the split rather than patching it — see 8fdf36232. The terminal claim now rests only on `hasExplicitReuseConsent`, read off the sound's own event with no relay involved; everything the legacy resolver cannot confirm goes back to the retryable `nostrPublishFailed`. `verify()` is fail-closed again and its doc says so, so this finding no longer reaches the user as a verdict about the creator.

hm21:
@mbradley — all three threads answered and resolved, and your `64991546a` is kept as-is. Thanks for pushing it: I moved the resolver to `fetchVideosUsingSoundDetailed` and left the picker test stubbing the old method, so it was passing on a missing-stub cast error rather than on the scenario. That is exactly the failure mode I had flagged in someone else’\s test two rounds earlier.

**The outcome is a revert, not a patch** — [8fdf36232](https://github.com/divinevideo/divine-mobile/commit/8fdf36232). Each finding checked out when I traced it, and together they said the boolean cannot carry a terminal claim at all:

- The block/content filters are the one that decided it. A cosmetic feed setting feeding a licensing decision is not something a narrower guard fixes.
- Widening to `candidates.length < lookup.ids.length` would have made the guard honest about that one shortfall while the same stale-revision risk stayed reachable through the filters. Rewording would have documented the gap without closing it.
- The `CLOSED` path was the same class of failure the split existed to eliminate, which is what disqualified `answered` as a foundation.
- realmeylisdev's 50-event window: you were right that my earlier reply answered `candidates.isEmpty` instead. Different cause, and the revert is what actually covers it — the source video being the oldest event carrying the tag makes that one deterministic, not incidental.

So `AudioReuseNotPermittedException` is now raised only for `hasExplicitReuseConsent`, read off the sound's own event with no relay in the way. Everything the legacy resolver cannot confirm goes back to `nostrPublishFailed`. `verify()` is fail-closed again and its doc says `false` means "not verified", never "the creator said no". `fetchVideosUsingSoundDetailed`, `AudioReuseConsentUnavailableException`, `AudioReuseCheck` and the picker's catch are gone; `sounds_repository` and the bottom sheet are back at their `main` state.

That also removes the Riverpod retry ladder you found — good catch, that was mine, and it disappears with the throw rather than needing its own fix. The sound-detail badge goes back to rendering "Credit only" on error, so that undocumented change is gone too.

Fix 1 and fix 2 are unchanged and the reported bug stays fixed.

Verification: `flutter analyze lib test integration_test` clean, 448 tests green, `dart format` clean, CI green on `8fdf36232`. The new gate is mutation-checked both ways — `if (true)` fails two tests, `if (false)` fails one.

**Deferred, and I think they want issues rather than this PR:** separating consent from the viewer's block/content filters (your thread 4), and the stale-revision hazard from a partial `getVideosByIds` (thread 5). Both are pre-existing on `main` and survive this PR — they now yield a retryable failure instead of a wrong verdict, which is containment, not a fix. Say the word and I'll open them.

Try Again stays out of scope by decision, tracked with `accountChanged` and `notSignedIn`.

mbradley (approved):
The revert is the right answer, and a better one than the fix I suggested. Declining to widen the guard to `candidates.length < lookup.ids.length` was correct: it would have made that one shortfall honest while the same stale-revision risk stayed reachable through the block and content filters, which drop events for reasons the lookup never sees.

The terminal claim now rests on `hasExplicitReuseConsent && !allowsReuse`, read off the sound's own event with no relay in the path. That is a refusal by construction, so all three findings are gone at the root rather than patched. Removing the throw also removed the Riverpod retry ladder it had enrolled `audioReuseTermsProvider` in.

Verified first-hand rather than read:

- Both directions of the new guard mutation-checked. Forcing it always-true reddens `reports an unverified legacy sound as a plain failure`; always-false reddens `blocks selected audio when the source explicitly forbids reuse`. The distinction the PR now turns on is genuinely pinned.
- The id fix is still guarded: reverting `attributionEventId` to `sound.id` reddens three resolver tests.
- 330 tests green (266 app-side across the resolver, publisher, publish-service, picker and l10n suites, plus 64 in `sounds_repository`), and `flutter analyze lib test` clean.
- No coverage lost that guards live code. The deleted tests covered `fetchVideosUsingSoundDetailed`, `AudioReuseCheck` and the picker's catch, all of which are gone with them.

I pushed one commit, `026fdab47`, and CI is green on it. The revert removed `AudioReuseCheck` but left `publishDirectUpload`'s doc referencing `[AudioReuseCheck.unverified]` and still describing the refusal as something the relays answer. Both now contradict the code and `publish_error_kind.dart`; `flutter analyze` does not catch the dangling reference. Yours to amend or revert.

Two things left with you, neither blocking.

**The description has two stale spots**, and they matter because it becomes the squash-merge commit body. The Verification list still claims mutation checks on "the `unverified` split, or the picker's catch," both of which are deleted. And Out of Scope says the state is reached "only on a relay-confirmed refusal," which is now the opposite of true — it is reached only on evidence that needs no relay at all.

**The two items you deferred have no issues yet**: the stale-revision hazard you are containing rather than fixing, and the filter/consent separation for the shared `VideosRepository`. Both are pre-existing on `main` so neither belongs in this PR, but you named them and they will evaporate at merge otherwise.

One retraction. I pressed the Try Again question in my last review, and the revert makes that deferral considerably more defensible: the terminal state now fires only on an event-stamped refusal that genuinely cannot be retried away, so the population reaching a futile button is both smaller and actually refused. I would not hold the PR on it.

Approving. The reported bug stays fixed, and the claim the code makes to the user is now backed by the evidence it actually has.

realmeylisdev (approved):
revert's the right call. all three things I flagged are closed: the terminal copy now only fires on hasExplicitReuseConsent so there's no relay in the path, the fail-closed guard has a real test again, and the status-code comment matches the code.

mutation-checked the new guard both ways plus the id fix and the type-first ordering, all go red. also traced the throw, publishVideoEvent has one caller and it's inside the try so it can't escape. ran it on the sim too, no new errors 👍

PR #6804, opened 2026-08-06
fix(publish): only self-label content language when the user declared one

### LINKED ISSUE #6863: fix(publish): only self-label content language when the user declared one
Auto-created from PR #6804.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6804
**Branch:** `fix/language-self-report-honesty`

rabble:
Closes #6863

## What

First step of a larger effort to make content language a real discovery signal for non-English users. Two independent fixes plus the plan that motivates them.

### 1. Only self-label language when the user actually declared one

`LanguagePreferenceService.contentLanguage` falls back to the device OS locale when the user never chose a content language, and `video_publish_service.dart` stamped that value as a NIP-32 `l`/ISO-639-1 self-label on every video. So we could not distinguish "this creator declared English" from "their phone happens to be English."

Measured against `api.divine.video` on a 3000-video sample:

```
en 92.4% | (empty) 2.2% | fr 1.2% | ja 0.9% | es 0.7% | ru/it/de 0.4% ea | rest <0.3%
```

And the non-English rows are visibly mislabelled — `pt` on "Cleaning up the Chileans", `pt` on "English tutorial", `de` on "ho ho ho green giant", `ja` on "#Vine". The field reports the publisher's phone setting, not the language of the video.

A new `declaredContentLanguage` getter returns the explicit choice or `null`; publish uses it. `contentLanguage` is unchanged and still backs *viewer-side* feed hints, where inferring from the phone is legitimate — asking what someone wants to watch is not the same as claiming what a video is in.

### 2. Keep the popular-feed cache key off the geo country hint

`viewerCountry` comes from a geo lookup behind a 250 ms timeout that yields `null` on a slow network. It was folded into the popular-feed cache key, so the key flipped between `:country=` and `:country=US` for the same user and missed the cache exactly when the network was worst. The country still goes out on the wire as a query param; it just no longer keys the cache.

## Known limitation, stated plainly

**This narrows the false claim but does not eliminate the category error.** `declaredContentLanguage` returns the same stored value as the viewer-side preference — there is one "Content Language" row in Content Preferences, and its copy gives no indication it gets stamped publicly on everything you publish. A Portuguese speaker in Germany who sets it to *see* Portuguese videos will still stamp `l pt ISO-639-1` on English uploads.

The change is still directionally right (almost nobody touches the setting, so the dominant effect is removing the fabricated default), but splitting viewer preference from publisher declaration is follow-up work. The plan's Phase 2 — per-video language in the composer — is the intended fix.

## Testing

- `flutter test test/services/video_publish/ test/services/language_preference_service_test.dart` — 133 pass
- `flutter test --coverage` in `packages/videos_repository` — 459 pass, 100.00% coverage

The cache-key test is a behaviour test: two `getPopularVideosPage` calls with differing `viewerCountry`, asserting the API client is hit once. It was proven to go red by temporarily reintroducing the country fragment at the *call site* — outside the helper — which is the regression shape a helper-output assertion would have missed.

## Notes for review

- No client code reads the ISO-639-1 `l` tag anywhere, so the now-`null` path is inert. `video_event.dart` and `video_stats.dart` parse `l` only in the content-warning namespace.
- funnelcake's derived `videos_latest.language` column will trend toward empty for new uploads as this rolls out. That is the intended effect. Nothing in funnelcake filters, groups, or aggregates on it (verified: zero `WHERE`/filter hits).
- `InMemoryFeedCache` is session-scoped, so the cache-key change needs no migration or version bump. Within a session, a call that resolved no country and a later one that resolved `BR` now share a key — an accepted trade.
- Pre-existing and untouched: `getPopularVideos`'s legacy path discards its preference suffix, so different `preferredLanguages` collide under a bare `popular` key.


NotThatKindOfDrLiz:
@rabble takeover follow-up pushed in e2af281cc.

What changed:
- routed the two new language publish verifications through the existing _verifyPublishVideoEvent helper instead of duplicating the full mocktail argument list
- added audioShareAttribution to the stale publishVideoEvent helper/stub matchers in the same test file
- reconciled the Phase 1 plan doc with the behavior test that shipped and marked Phase 1 complete
- updated the PR body with current verification and Closes #none; the bot created linked issue #6863 and removed needs-issue

Verification:
- flutter test test/services/video_publish/ test/services/language_preference_service_test.dart — 133 pass
- flutter test --coverage in mobile/packages/videos_repository — 459 pass, 100.00% coverage
- flutter analyze lib test integration_test — no issues
- GitHub checks are green on e2af281cc

NotThatKindOfDrLiz (approved):
Approved after takeover.

I verified the review findings are addressed: the language tests now use the shared publishVideoEvent verifier, the adjacent audioShareAttribution matchers are no longer stale, and the Phase 1 plan doc matches the behavior-test implementation. The PR is rebased onto current main, linked to issue #6863, and checks are green on e2af281cc.

PR #5883, opened 2026-07-06
fix(notifications): classify likes on comments via targetCommentId

hm21:
Closes #5884

## Description

Notifications rendered "liked your video" for **every** like, including likes on comments.

**Root cause:** `_mapNotificationKind` split a reaction into `like` vs `likeComment` solely by `isReferencedVideo`. That flag is set whenever the payload carries `referenced_video` — but Funnelcake populates `referenced_video` from the notification's **root video** (`root_rv` JOIN on `root_event_*`), and for a reaction on a comment the root resolves to the comment's parent video. So `referenced_video` is present for a like on a comment exactly as for a like on the video itself, and every reaction fell into the `like` branch.

The existing #4813 owner-mismatch reclassification only fires when the referenced video is owned by a **different** user, so a like on your **own** comment on your **own** video never got corrected.

**Fix:** classify the reaction by `targetCommentId`, which Funnelcake sets to the comment event ID for reactions on a kind 1111 comment (and leaves empty for a like on the video). This is independent of ownership and of the root-video metadata. `isReferencedVideo` is kept only as the fallback for the no-comment-id case.

**Related Issue:** 

## Out of Scope

- A deeper server-side fix (make `referenced_video` / `isReferencedVideo` reflect the *liked* event rather than the root video) would restore the original field contract for all clients, but touches ClickHouse queries and risks losing thumbnail context for comment-likes. Not done here — this client-side fix uses an already-correct signal. One residual edge case remains: if Funnelcake omits `target_comment_id` for a comment-like (comment event not resolved in the batch), it still falls back to "liked your video" — rarer than before, and the server fix would be the way to close it fully.

## Verification

- `flutter test` (notification_repository) → 118/118 pass, incl. new regression test and unchanged #4813 tests
- `flutter analyze lib test` → No issues found

## Type of Change

- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)

PR #6805, opened 2026-08-07
fix(notifications): stop labelling video likes "liked your comment"

### LINKED ISSUE #6806: fix(notifications): stop labelling video likes "liked your comment"
Auto-created from PR #6805.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6805
**Branch:** `fix/notification-like-comment-classification`

rabble:
Closes #6806

Notification rows read "X liked your comment" with no thumbnail, no quote, and no way to tell which comment — or, more often, which video. Two independent causes.

## 1. Video likes were classified as comment likes

`_mapNotificationKind` treated a missing `referenced_video` block as evidence of a comment target:

```dart
if (targetsComment) return NotificationKind.likeComment;
return n.isReferencedVideo
    ? NotificationKind.like
    : NotificationKind.likeComment;  // ← guess
```

`isReferencedVideo` is just `json['referenced_video'] != null` — whether Funnelcake managed to join the video row. It is absent whenever the root video fails to resolve (unindexed, deleted), and it is *present* for genuine comment likes too, because Funnelcake fills it from the root video either way. The function's own doc comment already said so: "`isReferencedVideo` cannot be used for this split".

Net effect: an ordinary like on your own video became "liked your comment", actor-anchored and therefore stripped of its thumbnail.

#5883 added `target_comment_id` as the positive comment signal but left the negative fallback in place. This drops it — a reaction without a target comment is a `like`.

The residual case (Funnelcake can't resolve the comment, so `target_comment_id` is empty on a real comment like) is still caught downstream by the owner-mismatch reclassification from #4813/#5634, which requires a confirmed metadata 404 *plus* a foreign `root_event_pubkey` before rewriting the row.

## 2. `likeComment` rows carried no context at all

Even when the classification was right, the row named nothing. The reaction's `content` is the emoji (`+`), and `_truncateComment` returned null for every kind that wasn't `comment` or `reply`.

Funnelcake already returns the liked comment's body as `comment_content` (`crates/clickhouse/src/queries.rs`, asserted in the OpenAPI schema) — the Flutter DTO simply never parsed it. Parsed now and threaded through as the quote source for `likeComment` rows, so the row renders:

> ❤️ **4th_Chaos** liked your comment
> "this vine still holds up" 9h

Server-side change: none. `NotificationCommentQuote` and the cache round-trip both work generically off `commentText`, so cold-start hydration keeps the quote.

## Testing

- `notification_repository`: new regression test that an unresolved referenced video stays `like`; three tests for the quote (body, truncation, null when the server omits it). Five existing tests that used `isReferencedVideo: false` as a stand-in for "comment like" now use `targetCommentId`, the actual signal. 167 pass, coverage 95.71% (gate 94%).
- `funnelcake_api_client`: three `comment_content` parse tests. 377 pass.
- `mobile/test/notifications`: widget test that a `likeComment` row renders the quote. 208 pass.
- `flutter analyze lib test integration_test` clean; adjacent notification/inbox suites (430 tests) pass.

Screenshots to follow once this is on a build.

## Reviewer takeover follow-up

A reviewer follow-up preserves anchorless (`a`-tag-only) reactions as actor-backed likes while keeping the display invariant: a row that says "liked your video" only gets a direct `videoAddressableId` when the root coordinate is a video coordinate owned by the notification recipient. Foreign root coordinates are withheld and taps fall back to resolving the reaction event. `likeComment` and `reply` may still carry a foreign root video coordinate because those rows are about the recipient's comment on that video, not an ownership claim on the video itself.

The tap resolver now also follows lowercase `a` tags on NIP-25 reactions, so anchorless addressable-video likes can still open the video when Funnelcake omits `root_addressable_id`.

Additional verification: `flutter test --coverage` in `mobile/packages/notification_repository` passed at 95.77% line coverage, plus targeted resolver, DTO/model, widget, and analyzer checks.

### DIFF
--- a/mobile/packages/notification_repository/lib/src/notification_repository.dart
+++ b/mobile/packages/notification_repository/lib/src/notification_repository.dart
@@ -1962,8 +1962,8 @@ class NotificationRepository {
 
   /// Returns the `targetEventId` for an actor-anchored notification.
   ///
-  /// - anchorless `like` → the reaction event ID (resolver follows its
-  ///   address tag to the video).
+  /// - anchorless `like` → the reaction event ID (resolver follows its `a` tag
+  ///   to the video when no recipient-owned root coordinate is available).
   /// - `likeComment`/`reply` → the referenced comment event ID (resolver
   ///   walks its E-tags to reach the root video).
   /// - `mention` → the source event ID (the kind-1 event that mentioned
@@ -2079,9 +2079,12 @@ class NotificationRepository {
   /// Returns the stable NIP-33 addressable ID for an actor-anchored
   /// notification, when the server provided the root video's full coordinate.
   ///
-  /// Populated for anchorless `like`, `likeComment`, and `reply` rows — the
-  /// tap handler uses it to navigate directly to the video without a relay
-  /// round-trip.
+  /// Populated for actor-anchored rows only when the coordinate is usable for
+  /// that row's ownership claim. Anchorless `like` rows render "liked your
+  /// video", so their direct route must name the notification recipient's own
+  /// video. `likeComment` and `reply` rows may legitimately route to a foreign
+  /// root video because the notification is about the recipient's comment on
+  /// that video.
   ///
   /// Used by the page-load path ([_mapActorAnchored]).
   String? _actorVideoAddressableId(
@@ -2093,7 +2096,16 @@ class NotificationRepository {
         mapped != NotificationKind.reply) {
       return null;
     }
-    return _nonEmpty(notification.rootAddressableId);
+    final addressableId = _nonEmpty(notification.rootAddressableId);
+    if (addressableId == null) return null;
+    final parsed = _parseAddressableId(addressableId);
+    if (parsed == null) return null;
+    if (!NIP71VideoKinds.isVideoKind(parsed.kind)) return null;
+    if (parsed.pubkey.isEmpty || parsed.dTag.isEmpty) return null;
+    if (mapped == NotificationKind.like && parsed.pubkey != _userPubkey) {
+      return null;
+    }
+    return addressableId;
   }
 
   bool _hasKnownReferencedVideoOwnerMismatch({
--- a/mobile/packages/notification_repository/lib/src/notification_repository.dart
+++ b/mobile/packages/notification_repository/lib/src/notification_repository.dart
@@ -776,8 +776,8 @@ class NotificationRepository {
       );
     }
 
-    // Actor-anchored kinds — `follow`, `mention`, `likeComment`, `reply`,
-    // `system` (and unknown future types fall through to `system`).
+    // Actor-anchored kinds — `follow`, `actorLike`, `mention`, `likeComment`,
+    // `reply`, `system` (and unknown future types fall through to `system`).
     return ActorNotification(
       id: row.id,
       type: _actorKindFromCachedType(row.type),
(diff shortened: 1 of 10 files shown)


mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -2224,10 +2256,9 @@ class NotificationRepository {
       final targetCommentId = n.targetCommentId;
       final targetsComment =
           targetCommentId != null && targetCommentId.isNotEmpty;
-      if (targetsComment) return NotificationKind.likeComment;
-      return n.isReferencedVideo
-          ? NotificationKind.like
-          : NotificationKind.likeComment;
+      return targetsComment
+          ? NotificationKind.likeComment
+          : NotificationKind.like;
```
NotThatKindOfDrLiz, line 2348: This maps every reaction without `targetCommentId` to video-anchored `like`, but if the inbound event has no `referencedEventId` (for example an `a`-tag-only NIP-25 reaction), `_videoAnchorEventId` returns empty, `_groupVideoAnchored` skips it, and `_mapActorAnchored` also skips it because `like` is video-anchored. That silently drops the notification. Please make the intended behavior explicit: either route anchorless reactions to an actor row, or add a regression test that pins dropping them as intentional. The current new test only covers the unresolved-video case where `referencedEventId` is still present (`my_video`).

NotThatKindOfDrLiz (changes requested):
I pushed the straightforward doc cleanup for `ActorNotification.commentText` in c3e6932b1. The remaining blocker is the behavior decision below: anchorless inbound reactions now disappear rather than rendering as actor rows. Please either route that case intentionally or add a regression test/documentation that pins the drop as the desired behavior. Mobile CI is still blocked by the unrelated `origin/main` `diagnosticTag`/`eventKind` compile errors, but the PR-specific changed-file hook passed on the pushed cleanup.

mobile/packages/notification_repository/lib/src/notification_repository.dart
rabble: Addressed in `da2d1bdee`. I reproduced the exact `a`-tag-only reaction payload first (it produced an empty page), then routed anchorless reactions to an `ActorNotification(type: like)` row. The row renders the localized “liked your video” copy, retains the reaction event ID for resolver fallback and the root addressable ID for direct routing, and uses a distinct `actorLike` cache type so cold-start hydration cannot reconstruct it as a `VideoNotification`. Regression coverage now pins live mapping, persistence/hydration, the model invariant, and user-visible copy. Full `notification_repository` and `models` package suites, affected widget/view suites, pre-push checks, and full `flutter analyze` pass.

NotThatKindOfDrLiz:
@rabble I pushed a reviewer-takeover follow-up on top of this PR.

Changes:
- Kept anchorless `ActorNotification(like)` rows, but only trust direct `videoAddressableId` when the root coordinate is a video coordinate owned by the notification recipient. Foreign root coordinates now fall back to resolver instead of rendering “liked your video” with another creator’s route.
- Added lowercase `a` tag support in `NotificationTargetResolver` for NIP-25 reaction events.
- Restored/updated the stale docs for `commentText`, `videoAddressableId`, and `isReferencedVideo`.

Verification:
- `flutter test --coverage` in `mobile/packages/notification_repository` (177 passed, 95.77% line coverage)
- `flutter test test/services/notification_target_resolver_test.dart`
- `flutter test test/src/models/relay_notification_test.dart`
- `flutter test test/src/actor_notification_test.dart`
- `flutter analyze` in `mobile`
- package-level analyzes for `notification_repository`, `funnelcake_api_client`, and `models`
- pre-push analyzer + changed-file test hook passed

NotThatKindOfDrLiz (approved):
Reviewer takeover pass is clean after the pushed invariant and resolver fixes. CI is green; leaving merge to Rabble after author review.

PR #6807, opened 2026-08-07
fix(build): drop references to the deleted publish diagnostics

rabble:
## Summary

`main` is currently red. `flutter analyze lib test integration_test` reports **16 `undefined_named_parameter` errors**, one of them in production code (`lib/repositories/relay_list_repository.dart:179`).

## Cause

A semantic conflict that no textual merge check could catch:

- **#6731** removed `diagnosticTag` from `NostrClient.publishEventAwaitOk` / `Nostr.sendEventAwaitOk` / `RelayPool`, and removed `eventKind` from `PublishOutcome` / `PublishTracker`.
- **#6721** (merged 2026-08-05, a day earlier) had already added a `diagnosticTag: 'relay-list-kind-10002'` caller in `RelayListRepository`.
- The curated-list service tests construct `PublishOutcome(... eventKind: event.kind ...)`.

Those files don't overlap with #6731's diff, so GitHub reported `MERGEABLE / CLEAN` for both and each side's CI was green against its own base. The breakage only exists in the *combination*.

## Fix

Drops the dead arguments rather than restoring the parameters — #6731's intent was to delete the rollout-only diagnostics, and nothing reads them.

| File | Removed |
|---|---|
| `lib/repositories/relay_list_repository.dart` | 1 × `diagnosticTag:` |
| `test/repositories/relay_list_repository_test.dart` | 7 × `diagnosticTag:` |
| `test/services/curated_list_service_*_test.dart` (6 files) | 8 × `eventKind: event.kind,` |

16 deletions, no behaviour change.

## Verification

- `flutter analyze lib test integration_test` — **No issues found** (was 16 errors)
- `flutter test` over all 7 affected suites — **192 passing**
- `dart format --set-exit-if-changed` — clean

## Note for #6746

`fix/6585-bound-relay-list-admission` passes `eventKind:` to `RelayPool.sendEventAwaitOk` in two new test files. It will hit the same wall on its next rebase; the fix there is dropping those two lines.

### DIFF
--- a/mobile/lib/repositories/relay_list_repository.dart
+++ b/mobile/lib/repositories/relay_list_repository.dart
@@ -176,7 +176,6 @@ class RelayListRepository {
       signed,
       targetRelays: targetRelays,
       timeout: _relayListPublishTimeout,
-      diagnosticTag: 'relay-list-kind-10002',
     );
     final acceptedByIndexer = outcome.acceptedBy.any(
       _environment.indexerRelays.contains,
--- a/mobile/test/repositories/relay_list_repository_test.dart
+++ b/mobile/test/repositories/relay_list_repository_test.dart
@@ -58,7 +58,6 @@ void main() {
         any(),
         targetRelays: any(named: 'targetRelays'),
         timeout: any(named: 'timeout'),
-        diagnosticTag: any(named: 'diagnosticTag'),
       ),
     ).thenAnswer((invocation) async {
       final event = invocation.positionalArguments.first as Event;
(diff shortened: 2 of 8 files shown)


hm21 (approved):
LGTM. Verified locally on your head: `flutter analyze lib test integration_test` clean (16 errors before), the 7 affected suites 192 passing, `nostr_client` and `nostr_sdk` analyze clean too, and no `diagnosticTag` reference survives anywhere in the tree.

Dropping the arguments rather than restoring the parameters is the right call — #6731 deleted the diagnostics pipeline outright (259 deletions across `relay_pool`, `publish_outcome`, `nostr.dart`), so there is nothing left to feed.

Merging this to get `main` green again — it currently fails to compile, so every open PR is blocked behind it.

PR #6810, opened 2026-08-07
fix(startup): gate playback during inactive launch

### LINKED ISSUE #6809: fix(startup): prevent inactive launch playback failover
Cold-start logs showed the first feed video being failed over and marked errored while iOS was inactive behind the push-permission prompt. The same session also showed dropped view events from compact REST rows, duplicate signed-out Nostr startup initialization, unknown route analytics for shell pages, 422 derivative-media misclassification, and nondeterministic equal-timestamp replaceable-event handling.

This issue tracks the combined startup playback and analytics consistency fix:

- pause feed watchdog/stale timers when playback gates go inactive
- treat mobile inactive lifecycle as non-foreground playback state
- preserve addressable d tags from compact REST payload variants
- avoid initializing signed-out placeholder Nostr clients during cold start
- classify HTTP 422 derived renditions as transient processing responses
- give shell pages route metadata for screen analytics
- apply NIP-01 equal-created_at event-id tie-breaks for replaceable videos

NotThatKindOfDrLiz:
Closes #6809

## Summary
- treat mobile `inactive` lifecycle as non-foreground so startup playback gates pause feeds before iOS launch handoff completes
- avoid initializing signed-out placeholder Nostr clients during cold start; relay-backed side effects remain gated on `NostrSessionReadiness.isReadyForActiveClient`
- preserve addressable d tags from compact REST payload variants without deriving this event's d tag from `a`/`A` references to other events
- classify HTTP 422 derived renditions as transient with hex-safe status matching, and keep explicit 401/403/404 classifications authoritative
- add shell route page metadata matching go_router's default page metadata shape (`name`, merged arguments, `restorationId`)
- apply the NIP-01 equal-timestamp lower-event-id tie-break for replaceable videos

## Verification
- `dart run build_runner build --delete-conflicting-outputs` from `mobile`
- `flutter test test/src/utils/playback_sources_test.dart` from `mobile/packages/infinite_video_feed`
- `flutter test test/src/video_stats_test.dart` from `mobile/packages/models`
- `flutter test test/widgets/app_lifecycle_handler_test.dart test/providers/nostr_service_identity_source_test.dart test/router/shell_transition_test.dart test/services/video_event_service_replaceable_test.dart` from `mobile`
- pre-push hook: merge-conflict check, analyzer, generated-file verification, changed-file tests

### DIFF
--- a/mobile/lib/providers/nostr_client_provider.dart
+++ b/mobile/lib/providers/nostr_client_provider.dart
@@ -627,14 +627,6 @@ class NostrService extends _$NostrService {
       final clientGeneration = _nextClientGeneration();
       _pendingPubkey = newPubkey;
 
-      // NIP-65 discovered-relays callback — see _userRelaysDiscoveredCallbackFor.
-      _registerClientCallbacks(
-        authService: authService,
-        client: newClient,
-        pubkey: newPubkey,
-        clientGeneration: clientGeneration,
-      );
-
       state = newClient;
       _setSessionIdentityState(newPubkey);
 
@@ -643,6 +635,14 @@ class NostrService extends _$NostrService {
         _pendingPubkey = null;
         return;
       }
+
+      // NIP-65 discovered-relays callback — see _userRelaysDiscoveredCallbackFor.
+      _registerClientCallbacks(
+        authService: authService,
+        client: newClient,
+        pubkey: newPubkey,
+        clientGeneration: clientGeneration,
+      );
 
       await _initializeClient(
         client: newClient,
--- a/mobile/lib/providers/nostr_client_provider.dart
+++ b/mobile/lib/providers/nostr_client_provider.dart
@@ -251,33 +251,34 @@ class NostrService extends _$NostrService {
         .map((relay) => relay.url)
         .toList();
 
-    // Create initial NostrClient using the atomic identity as signer.
-    // currentIdentity is nullable — before auth completes, the factory falls
-    // back to a no-op LocalKeySigner. _onAuthStateChanged recreates the client
-    // once the user authenticates.
+    // Create the initial NostrClient using the atomic identity as signer.
+    // currentIdentity is nullable; signed-out startup still exposes a placeholder
+    // client for consumers, but the placeholder must not initialize relays.
     final client = _createClient(authService.currentIdentity);
     final clientGeneration = _nextClientGeneration();
     _pendingPubkey = initialPubkey;
 
-    // NIP-65 discovered-relays callback — see _userRelaysDiscoveredCallbackFor.
-    _registerClientCallbacks(
-      authService: authService,
-      client: client,
-      pubkey: initialPubkey,
-      clientGeneration: clientGeneration,
-    );
-
-    // Schedule initialization after build completes
-    // Add user relays BEFORE initialize() to avoid race condition
-    Future.microtask(
-      () => _initializeClient(
+    if (initialPubkey != null) {
+      // NIP-65 discovered-relays callback — see _userRelaysDiscoveredCallbackFor.
+      _registerClientCallbacks(
+        authService: authService,
         client: client,
         pubkey: initialPubkey,
         clientGeneration: clientGeneration,
-        userRelayUrls: userRelayUrls,
-        source: 'build',
-      ),
-    );
+      );
+
+      // Schedule initialization after build completes.
+      // Add user relays BEFORE initialize() to avoid race condition.
+      Future.microtask(
+        () => _initializeClient(
+          client: client,
+          pubkey: initialPubkey,
+          clientGeneration: clientGeneration,
+          userRelayUrls: userRelayUrls,
+          source: 'build',
+        ),
+      );
+    }
 
     ref.onDispose(() {
       _initializationRetryTimer?.cancel();
@@ -343,21 +344,16 @@ class NostrService extends _$NostrService {
       );
       initializationTracker.begin(client);
       initialization = initializationTracker;
-      await _runClientInitialization(
-        client,
-        userRelayUrls,
-        source: source,
-      );
+      await _runClientInitialization(client, userRelayUrls, source: source);
       // The provider can be disposed during the await (e.g. a rapid identity
       // switch rebuilds it, or a test container is torn down mid-init). Reading
       // `ref` after disposal throws, so bail before the ref-backed checks below.
       if (!ref.mounted) return;
-      if (_isCurrentClientForPubkey(client, pubkey, clientGeneration)) {
-        _lastPubkey = pubkey;
-        _initializationFailureCount = 0;
-        _initializationRetryTimer?.cancel();
-        _initializationRetryTimer = null;
-      }
+      if (!_isCurrentClientForPubkey(client, pubkey, clientGeneration)) return;
+      _lastPubkey = pubkey;
+      _initializationFailureCount = 0;
+      _initializationRetryTimer?.cancel();
+      _initializationRetryTimer = null;
       _markClientReadyIfCurrent(client, pubkey, clientGeneration);
       Log.info(
         '[NostrService] Client initialized via $source()',
@@ -493,11 +489,7 @@ class NostrService extends _$NostrService {
       );
       initializationTracker.begin(client);
       initialization = initializationTracker;
-      await _runClientInitialization(
-        client,
-        userRelayUrls,
-        source: 'retry',
-      );
+      await _runClientInitialization(client, userRelayUrls, source: 'retry');
       if (!_isActiveIdentity(pubkey, clientGeneration)) {
         _disposeClient(client);
         return;
@@ -645,6 +637,12 @@ class NostrService extends _$NostrService {
 
       state = newClient;
       _setSessionIdentityState(newPubkey);
+
+      if (newPubkey == null) {
+        _lastPubkey = null;
+        _pendingPubkey = null;
+        return;
+      }
 
       await _initializeClient(
         client: newClient,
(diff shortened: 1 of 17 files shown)


rabble (commented):
Took over to fix the shard 2/4 failure. **CI is now green, 18/18.**

**Scope note up front:** I fixed the CI break only. I have **not** reviewed the substantive diff — the playback gating, `nostr_client_provider`, `video_event_service`, `app_lifecycle_handler`, or the `video_stats` changes. This still needs a real review from someone; please don't read green checks as a review.

---

## Pushed: `bc7ad21` — the shard 2/4 failure

`shell.dart` now reads `state.name`, `state.path`, `state.pathParameters` and `state.uri` to populate `NoTransitionPage` for `PageLoadObserver`, but `_FakeGoRouterState` in `shell_transition_test.dart` only overrode `pageKey`. So `Fake.noSuchMethod` threw `UnimplementedError: name`:

```
#0  Fake.noSuchMethod (package:test_api/src/frontend/fake.dart:50:5)
#1  _FakeGoRouterState.name (test/router/shell_transition_test.dart:13:7)
#2  shellRoutes.<anonymous closure> (package:openvine/router/routes/shell.dart:39:21)
```

Added the four missing overrides. I also asserted the resulting `name`, `restorationId` and merged `arguments` — **this part wasn't strictly required to unbreak CI**, so flagging it explicitly: overrides alone would leave a fake returning values nothing checks, which passes without pinning the metadata the PR exists to add. Drop it if you'd rather keep the test narrow.

Validated locally: reproduced the failure first, then `flutter test test/router/shell_transition_test.dart` green, `dart format` clean, `flutter analyze lib test` clean.

---

## Blocker — please look before merging

**Merging this into current `main` leaves a dead `diagnosticTag` parameter in `NostrClient`.**

This branch and `main` fixed the same broken-`main` incident in *opposite directions*:

- **#6807 (merged)** deleted the *callers*, matching #6731's intent (`chore(push): remove push-control relay diagnostics`).
- **This branch** restores the *parameter* at `nostr_client.dart:701` and keeps the caller in `relay_list_repository.dart:179`.

There's no textual conflict, so GitHub reports `MERGEABLE` and it merges clean — but the halves combine badly. `main` deleted the caller lines; this branch didn't touch them, so the merge takes the deletion. The parameter addition survives with nothing calling it.

Verified rather than predicted:

```
$ git checkout -b tmp origin/main && git merge --no-commit --no-ff origin/fix/startup-playback-analytics
merge exit=0
$ grep -rn "diagnosticTag" mobile/packages/nostr_client/... mobile/lib/repositories/relay_list_repository.dart ...
mobile/packages/nostr_client/lib/src/nostr_client.dart:701:    String? diagnosticTag,
```

One hit. The parameter, no callers. It compiles, so nothing catches it.

I'm **not** fixing this via takeover, because it needs your judgment rather than a mechanical edit: either the push-control diagnostics should stay removed per #6731 (drop the `nostr_client.dart` hunk and the `relay_list_repository` caller from this branch), or you're intentionally bringing them back, which is a product call that belongs in its own PR with a rationale.

Worth noting the mechanism, since it's the same trap [#6811](https://github.com/divinevideo/divine-mobile/pull/6811) is written to prevent: the merge base here (`6bdb0976`) was already the broken `main`, and patching around it in-branch is what produced the divergent fix. The rule that PR adds — *report the breaking commit, don't patch someone else's fix into your PR* — is exactly the one that would have caught this.

---

## Suggestion, non-blocking

`state.name ?? state.path` is a real branch with no coverage — the fallback fires when a route has no name, and nothing would catch it silently breaking. A second small case with `name => null` asserting the page falls back to `path` would close it. I left that to you rather than expanding the takeover.

---

Leaving this as a comment rather than an approval: my fix is validated and green, but the `diagnosticTag` question is open and I haven't reviewed the substance of the PR.

rabble (changes requested):
Substantive review now, against the seven items in #6809. (Earlier comment covered only the CI fix I pushed.)

**Verified correct** — checked these because they looked risky, and they hold up:

- **NIP-01 tie-break** (`video_event_service.dart`) — `originalEvent.id.compareTo(oldVideoEvent.id) < 0` retains the lexically-lowest id, which is what NIP-01 specifies for equal `created_at` on replaceable events. Correct direction; easy one to get backwards.
- **Watchdog disposal** (`infinite_video_feed.dart:714`) — I expected `disposeAll()` on deactivate to leave the feed permanently unprotected, but reactivation re-arms via `_resumeCurrentPlaybackIfReady()` → `_activateCurrentController()` → `_watchdog.start()` / `_staleDetector.start()` at lines 856–857. The early-return paths only skip arming when playback isn't starting either. No leak.
- **Lifecycle symmetry** (`app_lifecycle_handler.dart`) — `inactive` clears foreground + tickers on mobile; `resumed` restores both. Platform predicate correctly scopes it away from desktop focus changes.

---

## Blocking — `a`/`A` tag d-tag derivation is unsafe (`video_stats.dart:240`)

```dart
if (tagName == 'd' && dTag.isEmpty) dTag = tagValue;
if ((tagName == 'a' || tagName == 'A') && dTag.isEmpty) {
  dTag = _dTagFromAddressableId(tagValue) ?? '';
}
```

Two separate problems.

**1. An `a` tag identifies a *different* event.** This repo documents it as such:

- `video_event.dart:580` — `// Format: ['a', '34236:<pubkey>:<d-tag>', '<relay>', 'mention']`
- `video_event_publisher.dart:299` — `tags.add(['a', rootAddressableId, ''])`, the **root** of a revision chain
- `NOSTR_VIDEO_EVENTS.md:249` — "Addressable reference to kind 34235 or 34236 video event"

So the d-component of an `a` tag is the *referenced* event's identifier, not this one's. Adopting it makes this video claim another video's address.

**2. Order dependence makes it fire even when a real `d` tag exists.** Both branches guard on `dTag.isEmpty`, and tags are processed in array order. If `a` precedes `d`, the `a` branch populates `dTag` first, and the real `d` tag is then silently discarded because its own `dTag.isEmpty` guard is false.

That's the concrete failure: a revision video carries **both** an `a` (root) and a `d` (its own). Emitted in that order — and Nostr tag order is publisher-defined, so any client may — this parses the revision as having the *root's* d-tag. Given the d-tag is the addressable identity, that risks the revision colliding with or overwriting its own root in replaceable-event handling.

Suggested shape: resolve `d` across the whole tag list first, and only consult `a`/`A` if nothing was found — plus require the addressable id's kind and pubkey to match this event's before treating it as self-referential:

```dart
// after the tag loop
if (dTag.isEmpty) {
  dTag = _selfReferentialDTag(aTagValue, kind: kind, pubkey: pubkey) ?? '';
}
```

If the compact REST payloads this targets never carry a `d` tag at all, the map-key fallbacks above (`d_tag` / `dTag` / `addressable_d_tag` / `addressable_id`) already cover them, and the tag-loop `a` branch may just be unnecessary.

---

## Nit — 422 widens a false-positive surface

`_mentionsHttpStatus` requires an `http`/`status`/`response code` token plus `(^|[^0-9])422([^0-9]|$)`. That's well-guarded against numeric noise, but not against **hex**: a Blossom URL is `https://host/<sha256>`, and a hex digest containing `422` bounded by `a`–`f` matches — e.g. `.../a422b…`. It satisfies both conditions and reclassifies a genuine failure as "still processing", which means retrying instead of failing over.

Honest framing: this is **pre-existing** for `202` and not introduced here — adding `422` roughly doubles the exposure. Low severity, and I wouldn't block on it. Matching `status[: ]422` or a parsed status code would close it if you think it's worth the churn.

---

## Question — signed-out relay initialization

`nostr_client_provider.dart` now skips `_registerClientCallbacks` and `_initializeClient` entirely when `initialPubkey == null`, so the signed-out placeholder client never initializes relays.

That's clearly right for the duplicate-init problem in #6809. What I couldn't confirm from the diff: does any signed-out path depend on relay reads — anonymous feed browsing, profile previews from a shared link, invite landing? If those go through the REST API this is fine; if any of them read from relays, they'd now get an uninitialized client until sign-in. Worth a sentence in the PR description either way.

---

Still standing from my previous comment: merging into current `main` leaves an unused `diagnosticTag` parameter at `nostr_client.dart:701` (verified by test-merge). That one needs a decision from you, not a mechanical fix.

rabble (approved):
Both blockers resolved in `3f81d73`. Re-reviewed:

- **`a`/`A` d-tag derivation** — branch removed from the tag loop entirely. That's the cleaner of the two options I suggested, and it takes the ordering hazard with it: `d` can no longer be pre-empted by an earlier `a`. The map-key fallbacks (`addressable_id` / `addressableId`) still cover the compact REST payloads #6809 targets, so `_dTagFromAddressableId` stays reachable and nothing regressed. Tests added.
- **`diagnosticTag`** — gone from `nostr_client.dart`. Verified no callers remain anywhere on the branch (`git grep diagnosticTag` returns only the `agent_workflow.md` prose describing the original incident), so merging no longer reintroduces a dead parameter.

Also noted, both non-blocking and both fine by me: moving the 202/422 check below the 401/403/404 cases gives explicit statuses precedence, and deferring `_registerClientCallbacks` until after the generation check closes a callback-registration race on rapid identity switches.

CI green, 18/18. Approving.

⚠️ One heads-up unrelated to your diff: `strict_required_status_checks_policy` is now **`true`** on the `main` ruleset — the control named in Out of Scope on #6811, and the thing that let #6731 merge on a 46-hour-stale board. Branches must now be up to date with `main` before merging, so this will need a rebase onto current `main` (which has since gained #6807, #6790, #6794 and #6811) before the merge button lights up.

PR #6767, opened 2026-08-06
fix(support): reconcile follower counts

NotThatKindOfDrLiz:
## Summary
- reconcile follower headers against the rendered, blocklist-filtered list while preserving the large-account count allowance
- make own-follower cache refreshable with TTL, force-refresh network policy, stats-cache TTL, and hysteresis bypass for explicit refresh
- move profile follower stats back to FollowRepository-owned data so Funnelcake social counts cannot overwrite merged follower counts

Closes #6764

## Tests
- flutter test test/blocs/my_followers/my_followers_bloc_test.dart test/blocs/others_followers/others_followers_bloc_test.dart test/repositories/follow_repository_follower_stats_test.dart packages/follow_repository/test/src/follow_repository_cached_test.dart test/widgets/profile/profile_header_widget_test.dart
- flutter test packages/profile_repository/test/src/profile_repository_test.dart --plain-name "with Funnelcake API"
- flutter analyze lib test integration_test
- pre-push hook: analyzer + affected tests

PR #6811, opened 2026-08-07
docs(agents): gate PR takeovers on authorship and green CI

NotThatKindOfDrLiz:
## Description

Agents (Claude and Codex) have repeatedly pushed to open PRs without first establishing **whose PR it is**, and have posted "I pushed the fixes" handbacks *before* CI finished — leaving PRs red with no explanation and no owner.

The concrete case: on #6767 the agent ran the full review-then-takeover flow assuming it was a reviewer, and only discovered it was the **author** when GitHub rejected its review with `Review Can not request changes on your own pull request`. By then every decision had been made under the wrong model. It also posted its handback comment at 02:36:33Z — the CI run logged its first error at 02:37:18Z and failed at 02:39:16Z. The comment was written before the checks could possibly have reported.

Three structural gaps, addressed here:

**1. The takeover rules were not in this repo.** `PR_REVIEW.md` — the whole authorship/takeover runbook — lives in the `divine-context` handbook, loaded via a machine-local SessionStart hook for Claude Code and `~/.codex/AGENTS.md` for Codex. Neither `AGENTS.md` nor `.claude/rules/*` contained the word "takeover" or any instruction to check authorship. Any agent without that hook had no rule to follow. New `.claude/rules/pr_takeover.md` plus an `AGENTS.md` section give both toolchains a repo-local floor; `PR_REVIEW.md` stays authoritative where they overlap.

**2. "Inspect GitHub checks" was too weak, and buried.** It was the last bullet of *Clean Workspace Expectations*. "Inspect" is not "block until green," and nothing forbade posting the handback before checks completed. Now an explicit post-push gate: `gh pr checks <n> --watch`, read the result, *then* comment.

**3. `agent_workflow.md` §5 asserted something false.** It said *"`origin/main` always passes. Therefore any failing test on a feature branch is caused by that branch's diff. Full stop."* `main` is red right now, and the mechanism is worth naming because the rules should teach it: required checks are enforced but can go **stale**. #6731 removed `diagnosticTag`/`eventKind` from `nostr_sdk` and merged 2026-08-07T02:01Z with checks that had last run 2026-08-05T03:45Z — 46 hours earlier, by which point #6721 had added `diagnosticTag` callers and #6617 had added `eventKind` callers. Two independently-green PRs, no textual conflict, no re-run, broken `main`. The absolute framing left an agent hitting that with no procedure except to hunt a phantom bug in its own diff. This keeps *assume it's your fault* as the default and adds a two-step proof plus a **report, don't patch** rule for the rare genuine case.

**Related Issue:** Relates to #6767

## Out of Scope

- **`main` is currently red** and #6807 is the fix. Not touched here — patching it into an unrelated PR is exactly what the new rule forbids.
- **Nothing here prevents the next stale-check merge.** Enabling *Require branches to be up to date before merging* on `main` is the control that would have caught #6731; that is a repo-settings decision, not a docs one.
- No changes to `PR_REVIEW.md` in the `divine-context` handbook — that repo is separate.

## Verification

- Documentation only: `AGENTS.md`, `.claude/CLAUDE.md`, `.claude/rules/*.md`. No Dart, config, or workflow files touched.
- Cross-references checked to resolve: `pr_takeover.md` §3 → `agent_workflow.md#when-the-failure-is-not-yours`, and the checklist anchors.

All checks green. `Detect App CI Scope` path-filters this as docs-only, so `Analyze`, `Format`, `Generated Files`, `Tests`, and `VGV Tag Gate` skip rather than run — this branch does not inherit the currently-broken `main` (#6731 removed `diagnosticTag`/`eventKind` from `nostr_sdk` but left the callsites; #6807 is the fix). PRs that do touch Dart are inheriting it.

## Type of Change

- [x] 📝 Documentation

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -17,6 +17,7 @@ Generic Flutter and Dart standards live in `.claude/rules/`:
 
 - `rules/self_review_checklist.md`: consolidated pre-plan / pre-commit / pre-PR checklist
 - `rules/agent_workflow.md`: worktree-from-`origin/main`, rebase-at-boundaries, no-stacked-PRs, no-tech-debt, failing-tests-are-your-fault
+- `rules/pr_takeover.md`: check PR authorship *first*, takeover gates, answer every review item, never hand back before CI is green
 - `rules/architecture.md`: layered flow, package boundaries, barrel files
 - `rules/state_management.md`: BLoC-first UI, BlocProvider laziness, cross-route state persistence, Riverpod legacy rules, event transformers
 - `rules/code_style.md`: naming, widget composition, Effective Dart guidance
--- a/.claude/rules/agent_workflow.md
+++ b/.claude/rules/agent_workflow.md
@@ -158,9 +158,11 @@ inherit cold.
 assumption is that any failing test on a feature branch is caused by
 something in that branch's diff — start there every time.
 
-That default is a starting point, not an axiom. `main` is not
-protected by required status checks, so a red commit *can* land on it,
-and when one does, every open PR inherits the failure. Assume the
+That default is a starting point, not an axiom. Required checks can
+pass and still be **stale**: a PR tested days ago can merge into a
+`main` that has since gained a semantically conflicting change — no
+textual conflict, no re-run, green badge, broken `main`. When that
+happens every open PR behind it inherits the failure. Assume the
 failure is yours until you have positively proven otherwise, and prove
 it the specific way described in "When the failure is not yours"
 below. Never assert it from vibes.
@@ -200,9 +202,18 @@ the local environment differs from CI. Investigate the difference.
 
 ### When the failure is not yours
 
-Rare, but real — `main` has no required status checks, so a red commit
-can land and break every open PR behind it. Before you may say this,
-prove it. Two checks, both required:
+Rare, but real. The usual cause is a **stale-check merge**: PR A
+removes an API, PR B adds a caller, both go green independently, and
+whichever merges second lands without re-running against the other.
+Nothing conflicts textually, so `main` breaks and every open PR behind
+it inherits the failure.
+
+`origin/main` broke exactly this way on 2026-08-07: #6731 removed
+`diagnosticTag` / `eventKind` from `nostr_sdk` and merged with checks
+that had last run 46 hours earlier, by which point #6721 and #6617 had
+added callers.
+
+Before you may claim this, prove it. Two checks, both required:
 
 1. **The failing symbol is outside your diff.** Confirm the files and
    identifiers in the failure do not appear in
--- a/.claude/rules/agent_workflow.md
+++ b/.claude/rules/agent_workflow.md
@@ -230,10 +230,14 @@ Before you may claim this, prove it. Two checks, both required:
    branch. Different errors mean you have two problems, one of them
    yours.
 
-Then find what unblocks it — usually an open PR already fixing it:
+Then find what unblocks it — usually an open PR already fixing it. Do
+not trust PR prose search by itself: GitHub tokenizes terms and can
+return noisy matches. Start from the code history, then verify any
+candidate PR by reading its diff:
 
 ```bash
-gh pr list --state open --search "<failing symbol>"
+git log --oneline -S"<failing symbol>" origin/main
+gh pr list --state open --search "<failing symbol>"   # hint only
 ```
 
 Having proven it, **stop and report**. Do not patch around a broken
--- a/.claude/rules/agent_workflow.md
+++ b/.claude/rules/agent_workflow.md
@@ -154,8 +154,16 @@ inherit cold.
 
 ## 5. Failing tests are never acceptable, and always your fault
 
-**`origin/main` always passes.** Therefore any failing test on a feature
-branch is caused by something in that branch's diff. Full stop.
+**`origin/main` is supposed to always pass.** So your default
+assumption is that any failing test on a feature branch is caused by
+something in that branch's diff — start there every time.
+
+That default is a starting point, not an axiom. `main` is not
+protected by required status checks, so a red commit *can* land on it,
+and when one does, every open PR inherits the failure. Assume the
+failure is yours until you have positively proven otherwise, and prove
+it the specific way described in "When the failure is not yours"
+below. Never assert it from vibes.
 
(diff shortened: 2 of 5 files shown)


rabble (commented):
Reviewed the diff and independently verified the central factual claim, since commit 2 reverses commit 1 on it and the whole §5 rewrite rests on which version is right.

**Your commit-2 correction is right, and I can confirm the mechanism precisely.**

`main` is protected — but via **rulesets**, not classic branch protection, which is why it's easy to conclude otherwise. `GET /branches/main/protection` returns `404 Branch not protected`, and that 404 is misleading; the actual rules are:

```
$ gh api repos/divinevideo/divine-mobile/rules/branches/main --jq '[.[].type] | unique'
["deletion","non_fast_forward","pull_request","required_status_checks"]

$ gh api repos/divinevideo/divine-mobile/rules/branches/main \
    --jq '.[] | select(.type=="required_status_checks") | {checks:[.parameters.required_status_checks[].context], strict:.parameters.strict_required_status_checks_policy}'
{"checks":["Mobile CI"],"strict":false}
```

So: `Mobile CI` **is** required, and `strict: false` means *Require branches to be up to date before merging* is off. That is exactly the staleness mechanism commit 2 describes, and it makes the Out-of-Scope recommendation the precisely correct control — it's `strict: true` on ruleset `11641152`.

The #6731 timeline checks out too: checks completed `2026-08-05T03:45Z`, merged `2026-08-07T02:01:15Z`. 46 hours.

---

### Should fix — the correction isn't applied to the file CLAUDE.md calls canonical

`self_review_checklist.md:172` still carries the claim this PR exists to retire:

> **Never push red** — `origin/main` always passes, so any failing test on your branch is caused by your diff.

It's the only remaining instance on the branch (`git grep -n "always passes" FETCH_HEAD -- '*.md'`; the two other hits are unrelated prose in spec docs). It matters more than a stray duplicate would, because `.claude/CLAUDE.md` designates that file as "**the single source of truth** for what to double-check before planning / implementing / committing / wrapping a PR," and this PR adds two new sections to it — so a reader is being sent there by name and finding the retired absolute still stated.

Suggest matching the AGENTS.md phrasing you already landed:

> **Never push red** — assume any failing test on your branch is caused by your diff. In the rare case it isn't, prove it both ways and report rather than patch; see [`agent_workflow.md`](agent_workflow.md#when-the-failure-is-not-yours).

---

### Nits, non-blocking

**1. `gh pr list --search` won't reliably find the fixing PR.** In `agent_workflow.md`, "When the failure is not yours" suggests:

```bash
gh pr list --state open --search "<failing symbol>"
```

That searches PR titles and bodies, not code. It happens to work here only because #6807's body names `diagnosticTag` in prose. A fixing PR that doesn't mention the symbol in its description returns nothing, and "no open PR fixes this" is a worse conclusion to hand an agent than "search was inconclusive." Worth either noting the limitation inline or pointing at the merged-commit trail instead.

**2. `maintainerCanModify` reads `false` on same-repo PRs.** Verified on this PR: `isCrossRepository: false`, `maintainerCanModify: false`. Your §1 table scopes the field to forks correctly, but the bash block above it fetches the field unconditionally, so an agent sees `false` and may conclude it can't push to a branch it can in fact push to. A half-line — "meaningful only when `isCrossRepository` is true" — closes it.

---

Everything else reads well, and the `pr_takeover.md` §1 framing — that being the author is the *stricter* path, not the relaxed one, because nobody is downstream to catch a red build — is the part I think will actually change agent behavior.

Happy to approve once the checklist line lands.

rabble (approved):
Took over to land the one should-fix from my review, since it was a single line.

**Pushed:** `9e141f8` — replaces the surviving `origin/main` always passes absolute at `self_review_checklist.md:172` with the phrasing you already landed in `AGENTS.md`, plus a link to the new `#when-the-failure-is-not-yours` section.

Verified there are no other instances left:

```
$ grep -rn "always passes" .claude AGENTS.md
(no output)
```

**Validation:** `Mobile CI` green. `Analyze` / `Format` / `Generated Files` / `Tests` / `VGV Tag Gate` report `skipping` — `Detect App CI Scope` path-filters this as docs-only, which is correct and matches what you noted in the PR body.

**Review items from my earlier comment:**

- *Stale absolute in `self_review_checklist.md`* — **fixed** in this push.
- *`gh pr list --search` searches prose, not code* — **declined for now.** It's a real sharp edge but it's your call how to word the caveat, and it needs judgment about what to recommend instead. Left for you.
- *`maintainerCanModify` reads `false` on same-repo PRs* — **declined for now**, same reason. I confirmed the behavior on this very PR (`isCrossRepository: false`, `maintainerCanModify: false`), so the misread is real, but the wording is yours.

Neither nit blocks merge.

---

One thing you'll want to know, because it validates the rule you wrote and I got it wrong first: I initially read `main` as unprotected. `GET /branches/main/protection` returns `404 Branch not protected`, which is what I'd guess led to commit 1's framing. The protection is a **ruleset**:

```
$ gh api repos/divinevideo/divine-mobile/rules/branches/main --jq '[.[].type] | unique'
["deletion","non_fast_forward","pull_request","required_status_checks"]
```

with `Mobile CI` required and `strict: false`. So your commit-2 correction is exactly right, and the control you named in Out of Scope is `strict: true` on ruleset `11641152`. Still unowned as of this comment — worth an issue so it doesn't get lost, since it's the thing that actually caused the breakage.

Approving the resulting state. Merge decision stays yours.

NotThatKindOfDrLiz:
Takeover fixes landed in `4843759f9`.\n\nAddressed the remaining review items:\n\n- Replaced the literal `gh pr list --search <symbol>` guidance with a code-history-first path and marked PR prose search as a hint only.\n- Documented that `maintainerCanModify` is meaningful only for fork PRs; same-repo PRs can report `false` even when mutable.\n- Closed the repo-local floor gap: if `PR_REVIEW.md` or its team mapping is unavailable, same-repo takeover stops and reports the missing runbook as the blocker.\n\nVerification after push:\n\n- pre-push hook: no merge conflicts with main; no Dart files changed, skipped Dart checks\n- `git diff --check` clean\n- `gh pr checks 6811 --watch`: Codex Configuration, Mobile CI, Detect App CI Scope, mergeability, semantic-pull-request, and license/cla all pass; Analyze/Format/Generated Files/Tests/VGV Tag Gate skipped as docs-only\n\nFinal PR state: MERGEABLE, APPROVED, checks green.

PR #6815, opened 2026-08-07
chore: add repository Codex configuration

### LINKED ISSUE #6816: chore: add repository Codex configuration
Auto-created from PR #6815.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6815
**Branch:** `chore/codex-agent-config`

hm21:
Closes #6816

## Description

Adds repository-scoped Codex skills, MCP configuration, a Flutter test-debugging agent, and lifecycle hooks. Codex generated these files from the repository’s existing Claude configuration and adapted the copied settings to Codex’s project-local formats and hook payloads. The hooks use portable Git-root paths and Codex-native `apply_patch` payloads so the configuration works across checkouts and worktrees.

**Related Issue:** N/A

## Out of Scope

Changes to the existing Claude configuration and Flutter application code.

## Verification

- `jq empty .codex/hooks.json`
- `codex mcp list`
- `bash -n .codex/hooks/*.sh`
- Targeted Nostr-ID and VineTheme hook smoke tests
- Secret-pattern scan across `.agents/` and `.codex/`
- `git diff origin/main...HEAD --check`
- Pre-push hook (merge-conflict check; no Dart files changed)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore


.codex/hooks/post-edit-analyze.sh
```diff
@@ -0,0 +1,36 @@
...
+  -e 's/^\*\*\* Move to: (.*)$/\1/p')
+
+[ -n "$FILES" ] || exit 0
+
+while IFS= read -r FILE_PATH; do
+  [[ "$FILE_PATH" =~ \.dart$ ]] || continue
+  [ -f "$FILE_PATH" ] || continue
+
+  ANALYSIS_OUTPUT=$(dart analyze "$FILE_PATH" 2>&1 || true)
+  if echo "$ANALYSIS_OUTPUT" | grep -q " error \| warning \| info "; then
```
mbradley: Two separate defects here, both of which make this guard inert.

**The severity match never catches a warning.** `dart analyze` right-pads the severity column to 7 characters, so `warning` starts at column 1 while `  error` and `   info` are indented. `grep " warning "` requires a leading space, so it never fires. Verified on a real run against a file whose only issue is an unused import:

```
$ mise exec -- dart analyze mobile/lib/probe.dart
warning - probe.dart:1:8 - Unused import: 'dart:io'. Try removing the import directive. - unused_import
$ ... | grep -q " error \| warning \| info " && echo BLOCKS || echo ALLOWS
ALLOWS
```

Unused imports, unused locals and dead null-aware are precisely the class CI fails on, and they all pass this hook today.

**`dart` is not on PATH.** This repo pins the SDK through mise (`mobile/mise.toml`); `which dart` returns nothing. Because the call is wrapped in `|| true`, the failure is silent and every file is allowed. `scripts/install-hooks.sh` already uses `mise exec -- dart`. One wrinkle: `mise exec` only resolves from inside `mobile/`, so the hook has to `cd` there and pass an absolute path. I confirmed that works for both `mobile/lib/**` and `mobile/packages/*/lib/**`.

Not leaving this as a suggestion block because the fix spans two hunks (a resolver above the loop, plus these lines) and a half-applied suggestion would break the script. Tested patch:

```bash
# above the while loop
REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null) || exit 0
MISE_DIR="$REPO_ROOT/mobile"
if command -v mise >/dev/null 2>&1 && [ -f "$MISE_DIR/mise.toml" ]; then
  run_dart() { (cd "$MISE_DIR" && mise exec -- dart "$@"); }
elif command -v dart >/dev/null 2>&1; then
  run_dart() { dart "$@"; }
else
  exit 0
fi

# lines 25-26
  ABS_PATH=$(cd "$(dirname "$FILE_PATH")" && pwd)/$(basename "$FILE_PATH")
  ANALYSIS_OUTPUT=$(run_dart analyze "$ABS_PATH" 2>&1 || true)
  if echo "$ANALYSIS_OUTPUT" | grep -qE '^[[:space:]]*(error|warning|info)[[:space:]]+-[[:space:]]'; then
```

With that applied the same warnings-only file returns `{"decision": "block", ...}`.

.codex/hooks/post-edit-format.sh
```diff
@@ -0,0 +1,30 @@
...
+FILES=$(printf '%s\n' "$PATCH" | sed -nE \
+  -e 's/^\*\*\* (Add|Update) File: (.*)$/\2/p' \
+  -e 's/^\*\*\* Move to: (.*)$/\1/p')
+
+[ -n "$FILES" ] || exit 0
+
+while IFS= read -r FILE_PATH; do
+  [[ "$FILE_PATH" =~ \.dart$ ]] || continue
+  [ -f "$FILE_PATH" ] || continue
+  dart format "$FILE_PATH" 2>/dev/null || true
```
mbradley: Same toolchain problem as the analyze hook, and here it means the formatter silently does nothing. `dart` is not on PATH under this repo's mise pinning, and `2>/dev/null || true` hides it.

Verified A/B on a deliberately misformatted file: with this version the file is byte-identical afterwards; with `mise exec -- dart format` run from `mobile/` on an absolute path, it formats correctly.

Same `run_dart` resolver as the analyze hook, then:

```bash
  ABS_PATH=$(cd "$(dirname "$FILE_PATH")" && pwd)/$(basename "$FILE_PATH")
  run_dart format "$ABS_PATH" >/dev/null 2>&1 || true
```

.codex/hooks/pre-commit-build-runner.sh
```diff
@@ -0,0 +1,69 @@
...
+# Run build_runner for each unique package root
+if [ -n "$PACKAGE_ROOTS" ]; then
+  echo "$PACKAGE_ROOTS" | while read -r ROOT; do
+    [ -z "$ROOT" ] && continue
+    echo "Running build_runner in $ROOT..."
+    (cd "$ROOT" && dart run build_runner build --delete-conflicting-outputs 2>&1) || {
+      echo "ERROR: build_runner failed in $ROOT"
+      exit 1
+    }
+  done
```
mbradley: A build_runner failure cannot block the commit from here, and the hook also writes non-JSON to stdout.

`echo "$PACKAGE_ROOTS" | while read` puts the loop in a pipeline subshell, so `exit 1` kills the subshell rather than the hook. Knock-on effects: remaining package roots are skipped, and the regenerated-file staging block at line 60 never runs.

Separately, `echo "Running build_runner in $ROOT..."` and build_runner's own output both go to stdout, which is where the runner reads the hook's JSON result from. Reproduced in a scratch repo with a stub `dart` that always fails:

```
stdout=[Running build_runner in pkg...
BUILD FAILED: some builder error
ERROR: build_runner failed in pkg]  rc=1
```

That is not a deny, it is garbage on the result channel. `check-git-hooks.sh` in this same PR already does the right thing, so the fix is to match it: feed the loop from a heredoc so it runs in this shell, send progress to stderr, capture build output, and emit `permissionDecision: "deny"` on failure. With that applied the same scenario returns proper deny JSON with the tail of the build log in the reason.

Two smaller things while you are in here. `for FILE in $STAGED_FILES` at lines 27 and 61 is unquoted, so a path containing a space splits into fragments, both `[ -f ]` checks fail and the file is silently skipped; `while IFS= read -r` (which the post-edit hooks already use) avoids it. And `[ -f "$FREEZED" ] && git add "$FREEZED"` as the last statement in a loop body returns 1 when the file is absent, which trips `set -e` and aborts the script; an `if` block avoids that.

Happy to push the tested version as its own commit.

.agents/skills/claudeception/resources/research-references.md
```diff
@@ -0,0 +1,183 @@
+# Research References
+
+This document compiles the academic research that informed the design of Claudeception.
+
+## Core Papers
+
+### Voyager: An Open-Ended Embodied Agent with Large Language Models
+
+**Authors**: Wang, Xie, Jiang, Mandlekar, Xiao, Zhu, Fan, Anandkumar
+**Published**: May 2023
```
mbradley: The copy stripped trailing double-spaces, which are hard line breaks in Markdown. There were 9 of them in this file (lines 9, 10, 29, 30, 47, 48, 65, 66, 83) and none survive, so five citation blocks now render as one run-on line: "**Authors**: Wang, Xie... **Published**: May 2023 **URL**: https://arxiv.org/..." instead of three lines.

The `.claude/skills/` original still has them, so `git checkout origin/main -- .claude/skills/claudeception/resources/research-references.md` and copying it across verbatim restores the rendering.

I think this came from the `git diff --check` step in the verification list, which flags trailing whitespace as a code hygiene problem. It is not one in Markdown prose. Flagging the cause rather than just the file because the same reflex will hit the next copy. I checked the rest of the tree: the other whitespace-only diffs are inside fenced code blocks where it is inert, and no `.py`, `.sh` or YAML frontmatter lost anything meaningful.

.agents/skills/art-direct/docs/2026-01-28-art-direct-design.md
```diff
@@ -0,0 +1,322 @@
...
+    - blue corporate tones
+    - glass building reflections
+
+ai_prompt_suffixes:
+  photography: "documentary style, natural lighting, Kodak Portra 400 --ar 16:9"
+  illustration: "editorial illustration, muted palette, subtle texture"
+```
+
+**Storage locations:**
+- `~/.claude/skills/art-direct/styles/` (user global)
```
mbradley: Minor inconsistency: `SKILL.md:474` in this same skill was rewritten to `~/.agents/skills/art-direct/styles/`, but this line still says `~/.claude/skills/art-direct/styles/`. The rewrite is correct (`$HOME/.agents/skills` is one of Codex's documented discovery locations), so this one just got missed by the same pass.

mbradley (changes requested):
I checked the structural claims in the description against Codex's own docs and source, and they hold up. `.agents/skills/` really is Codex's skill discovery path (it will not read `.claude/skills/`, so the copy was necessary), `<repo>/.codex/hooks.json` is a documented hook location, `matcher: "Edit|Write"` is valid for `apply_patch` edits, `timeout` is in seconds so dividing by 1000 was right, and `$(git rev-parse --show-toplevel)` is the idiom upstream's own example uses (there is no project-root env var, and `command_runner.rs` spawns through a shell, so the substitution expands). I had findings queued against all of those and dropped them.

What I am asking for changes on is narrower: three of the six ported hooks do not do their job on a correctly configured machine.

- `post-edit-analyze.sh` never matches a warning, so warnings-only files pass the guard.
- Both post-edit hooks call bare `dart`, which is not on PATH under this repo's mise pinning, and both swallow the failure with `|| true`. Format changes nothing, analyze allows everything.
- `pre-commit-build-runner.sh` cannot block a failed codegen run, and writes non-JSON to stdout where the runner expects the hook result.

I verified each by running it rather than reading it, details inline. Patches are drafted and tested for all three if you want them pushed, just say so and I will put them up as separate commits.

### The one thing that needs your call

`.agents/skills/` has to exist, so the duplication is not the question. The question is that nothing checks the two trees still agree, and they already do not. `.codex/hooks/check-git-hooks.sh` picked up regex hardening (`\s` to `[[:space:]]`, `\b` to `([[:space:]]|$)`) that was never backported to `.claude/hooks/check-git-hooks.sh`, so they ship divergent behaviour from day one. Repo-wide there are 25 ratchet scripts guarding smaller risks than an 18,805 line mirror.

Four options as I see them:

1. Symlink `.agents/skills` to `.claude/skills`. Collapses the diff to one line and makes drift structurally impossible. Needs a test that Codex's loader follows symlinks, and accepts that Windows checkouts without symlink support get a text file.
2. Invert it, `.agents/` canonical with `.claude/skills` symlinked to it. Same benefit, and it matches the direction the repo is already going, but a mistake breaks the setup people use today.
3. Keep the copy and add a `check_agents_skills_sync.sh` ratchet with an allowlist for the deliberately substituted files. Fits the existing idiom; the allowlist becomes its own maintenance.
4. Generator script plus that check.

I would lean 1 if symlinks test clean and 4 if they do not. 3 on its own leaves the substitution rules implicit, which is what produced the Markdown breakage below. Your call, not mine.

### Smaller things, not blocking

`AGENTS.md` is untouched, so nothing tells a contributor that `.codex/` and `.agents/` exist or that editing one side obligates the other. Worth a short section, and worth noting there that project-scoped `.codex/` config only loads once the project is trusted, otherwise the whole tree silently no-ops and the next person debugs it from scratch. Separately, project-local config is CLI-only today; Codex Desktop and the Cursor extension ignore it (openai/codex#13025 is still open).

I checked for secrets across all 137 files and found none. The MCP servers in `.codex/config.toml` are exact parity with the existing `.mcp.json`, including the unpinned `@nostrbook/mcp@latest`, so that risk is pre-existing rather than something this PR widens. I am raising it separately rather than here.

I have not pushed anything to your branch.

.codex/hooks/pre-commit-build-runner.sh
hm21: Fixed in `961530614`. Staged paths are read with NUL delimiters into arrays, build progress goes to stderr, build output is captured, and failures return documented `PreToolUse` deny JSON. The regression coverage exercises a simulated builder failure, clean stdout on success, and generated files whose paths contain spaces.

.codex/hooks/post-edit-format.sh
hm21: Fixed in `961530614`. Formatting now uses the shared repository Dart runner and no longer swallows toolchain or formatter failures. A missing runner or failed format returns actionable PostToolUse feedback instead of silently succeeding.

.codex/hooks/post-edit-analyze.sh
hm21: Fixed in `961530614`. I replaced the separate formatter and analyzer handlers with one sequential `post-edit-dart.sh` handler, added repository Dart resolution through `mise exec -- dart`, and anchored the severity match so warnings at column 1 are caught. The regression test runs with a mise-only PATH and verifies that formatting happens before a warnings-only analysis is blocked.

.agents/skills/art-direct/docs/2026-01-28-art-direct-design.md
hm21: Fixed in `961530614`. The generated Art Direct documentation now consistently points to `~/.agents/skills/art-direct/styles/`, and the sync regression check pins that substitution.

.agents/skills/claudeception/resources/research-references.md
hm21: Fixed in `961530614`. The generated mirror now uses explicit `<br>` elements for these citation line breaks, preserving the intended rendering without reintroducing trailing whitespace that would fail `git diff --check`.

hm21:
Thanks for the careful review. I addressed the findings in `961530614`:

- Replaced the concurrent formatter/analyzer handlers with one sequential post-edit hook and a shared mise-aware Dart runner.
- Made analyzer severity matching cover warnings at column 1 and turned toolchain/format/analyze failures into actionable hook feedback.
- Reworked the build_runner hook to handle spaced paths, keep stdout protocol-clean, and return documented deny JSON on failure.
- Added a deterministic `.claude/skills` -> `.agents/skills` generator, a CI sync check, and contributor guidance in `AGENTS.md`.
- Restored the Markdown citation layout with `<br>`, fixed the Art Direct path, and backported the portable git-command regex to the Claude hook.

Verification completed:

- `bash .codex/scripts/test-config.sh`
- `bash -n` across the changed hook/config scripts
- `jq empty .codex/hooks.json`
- `codex mcp list`
- `git diff --check`

The new Codex Configuration CI job is green. The remaining Mobile CI jobs are still running.

mbradley (approved):
Re-reviewed at `961530614`. I re-ran each finding rather than reading the diff, and all five are fixed:

- A warnings-only file now returns `decision: "block"`.
- With `PATH=/usr/bin:/bin` the hook reports `"Install mise and run 'cd mobile && mise install'"` instead of silently allowing, which is better than what I asked for. Confirmed it also resolves correctly from a subdirectory and picks up the package's own `analysis_options` under `mobile/packages/*`.
- The build_runner hook emits deny JSON on failure with progress on stderr, stdout is empty on the success path, and a staged `mobile/lib/with space.g.dart` is picked up correctly.
- The citation block renders again, and the `<br>` form is derived deterministically from the source (9 trailing-space lines to 9 `<br>`).
- The art-direct path is consistent.

On the sync machinery, I checked the parts that would have worried me. The generator is idempotent and reproduces the committed mirror with zero diff across all 128 files including modes. `--check` fails in all four directions: source edit, mirror edit, deletion, and a stray file. Most importantly the `codex-config` job carries no `paths:` filter, no `needs:`, and no `if:`, unlike `generated-files` directly below it, so it still runs on a PR that touches only `.claude/skills/**`, which is the exact case where every other job skips. That was the thing most likely to make the guard decorative and it does not.

### One commit pushed: `ea53b7d67`

The severity match had no working test. The stub emitted a warning and exited 2, so the exit-code fallback satisfied the assertion by itself. I reverted the match to the original pattern and the suite stayed green, which means the fix this PR is built around was unprotected.

That commit pins the severity-match path explicitly and adds the two cases only it can decide: info-level diagnostics, which real `dart analyze` reports with exit 0 (measured: info-only exits 0, warning-only exits 2), and a file with no diagnostics, which must not block. I verified both mutations now fail the suite, and that the suite is green unmutated. All 14 checks pass on the pushed head.

Worth noting for anyone reading later: an info-level case alone would not have caught this. Info is indented, so the old pattern matched it fine. Only `warning` sits at column 1. The assertion on the reason prefix is what separates the two code paths.

### Correction

I said `[ -f x ] && git add x` as a loop body's last statement trips `set -e`. That is wrong. Bash exempts a failing non-final member of an AND-list, and I confirmed the loop completes with rc=0. You rewrote it to `if` blocks on my say-so. The rewrite is clearer and worth keeping, but I reported a defect that was not one.

Approving. Please look over `ea53b7d67` since it is my commit on your branch, and merge when you are happy with it.

PR #6817, opened 2026-08-07
fix(video-editor): stop speed-render thrash stalling the encoder

### LINKED ISSUE #6818: fix(video-editor): stop speed-render thrash stalling the encoder
Auto-created from PR #6817.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6817
**Branch:** `fix/video-editor-speed-render-thrash`

hm21:
Closes #6818

## Description

Device logs from a 28-clip draft (iPad13,17, iOS 18.7.8) showed 14 failed speed renders in an 11-minute editing session — 6 × `Render export stalled after 20s with no progress [progress=0.00 mode=AVAssetExportPresetHighestQuality]` and 8 × `Swift.CancellationError`, alongside RSS climbing from 284 MB to a 640 MB peak.

Two root causes, both in `ClipSpeedRenderService`:

**1. Unbounded render fan-out.** `_ensureSpeedClipsRendered` in `video_editor_canvas.dart` requests a speed body for every non-1× clip on *every* timeline change, so a 28-clip draft opened 28 native export sessions simultaneously. Past a couple of concurrent sessions the platform encoder stops making progress entirely and `pro_video_editor` fails the session with a 20s stall. Individual renders that did complete took 66–991 ms, so serialising them is both faster and more reliable than letting them contend.

This PR caps concurrent speed renders at 2 and queues the rest FIFO (head-of-timeline clips finish first, since those are the ones most likely to be played first).

The cap deliberately lives in `ClipSpeedRenderService`, **not** in `VideoEditorRenderService.renderNativeVideoToFile`. The latter looks like the natural choke point, but `renderVideo` (used by `TransitionSeamRenderService`) re-enters it several times per seam — a semaphore there would let an outer call hold a slot while waiting for its own inner calls, and deadlock.

A stale check *after* acquiring a slot drops the queued backlog when the editor closes. That backlog is what previously kept the encoder busy for ~20s after teardown and produced the `CancellationError` renders logged well after the user had left the editor.

**2. Cache key included `clip.id`.** The class doc already described the key as "the clip's file, trims and speed" — the id had drifted in and was never part of the documented intent. Because duplicating a clip, splitting one, or reopening a draft all mint fresh ids for byte-identical bodies, the id forced a full re-encode of identical material. The render output is fully determined by (source file, trims, speed), so clips agreeing on those now share one render. Visible in the logs as e.g. `1786035685738183_start_copy_...` re-encoding a body already rendered moments earlier.

**Related Issue:** 

## Out of Scope

Three findings from the same log deliberately left for separate PRs:

- `clear()` deletes the on-disk speed bodies when the editor closes, so reopening a draft still re-renders everything from scratch. Keeping them is a trade-off against temp-dir disk usage and wants its own decision.
- Duplicating a draft logs `Draft not found: draft_...` 1 ms after `Saving draft: draft_...` — a read-before-write race in the duplicate flow.
- `ClipEditorBloc` logs its init twice after the editor has been reopened several times, suggesting a second subscription. Unconfirmed; needs a reproduction first.

## Verification

- `flutter test test/services/video_editor/` — 290 passed
- `flutter test test/widgets/video_editor/` — 813 passed
- `flutter analyze lib/services/video_editor/ test/services/video_editor/` — no issues
- Both suites re-run after rebasing onto current `main`

Three regression tests added, each verified to fail without its fix:

| Test | Without the fix |
|---|---|
| `caps concurrent native renders so the encoder does not stall` | `Expected: <2> Actual: <3>` |
| `drops queued renders once the editor closes` | `Expected: <2> Actual: <3>` |
| `reuses one render for clips that differ only by id` | `Expected: not null Actual: <null>` |

Manual: reproduced and confirmed fixed on the iPad the original logs came from, using the same 28-clip draft. Not yet re-verified on a Samsung Galaxy.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/services/video_editor/clip_speed_render_service.dart
+++ b/mobile/lib/services/video_editor/clip_speed_render_service.dart
@@ -1,6 +1,8 @@
 // ABOUTME: Pre-renders a clip's trimmed body at its playbackSpeed into a plain
 // ABOUTME: normal-rate file so the preview plays it at 1× instead of retiming live.
 
+import 'dart:async';
+import 'dart:collection';
 import 'dart:convert';
 import 'dart:io';
 
@@ -84,6 +86,42 @@ class ClipSpeedRenderService {
     return speed > 0 && speed != 1.0 && clip.video?.file != null;
   }
 
+  /// How many speed bodies may encode natively at the same time.
+  ///
+  /// The canvas asks for a render per non-1× clip on every timeline change, so
+  /// a 28-clip draft would otherwise open 28 export sessions at once. Past a
+  /// couple of concurrent sessions the platform encoder stops making progress
+  /// and `pro_video_editor` fails them with a stall (`progress=0.00` after
+  /// 20s), which is slower *and* lossier than encoding them a few at a time.
+  static const _maxConcurrentRenders = 2;
+
+  int _activeRenders = 0;
+  final _waitingForSlot = Queue<Completer<void>>();
+
+  /// Resolves once a render slot is free. Slots are handed out FIFO so the
+  /// clips at the head of the timeline — the ones the user is most likely to
+  /// play first — finish first.
+  Future<void> _acquireRenderSlot() {
+    if (_activeRenders < _maxConcurrentRenders) {
+      _activeRenders++;
+      return Future<void>.value();
+    }
+    final waiter = Completer<void>();
+    _waitingForSlot.add(waiter);
+    return waiter.future;
+  }
+
+  /// Hands this slot straight to the next waiter, or gives it back to the pool
+  /// when nobody is queued. Passing it on keeps [_activeRenders] at the cap
+  /// instead of dipping below it between renders.
+  void _releaseRenderSlot() {
+    if (_waitingForSlot.isNotEmpty) {
+      _waitingForSlot.removeFirst().complete();
+      return;
+    }
+    _activeRenders--;
+  }
+
   Future<RenderedSpeedClip?> _render(DivineVideoClip clip, String key) async {
     // Reachable only via [render], which gates on [_needsRender] (video
     // non-null). Re-checked here so the type is provably non-null below.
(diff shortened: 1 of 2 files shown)


rabble (approved):
### Summary\nFix is correct and well-scoped — **LGTM**. Two root causes (unbounded fan-out + id-polluted cache key) are both addressed with minimal, well-tested changes. All CI green (14 checks), `flutter analyze` clean on the changed file, local `flutter test test/services/video_editor/` passed (287) and `clip_speed_render_service_test.dart` passed.\n\n### What was checked\n- `gh pr view` / `gh pr diff` — 2 files, +158/-18\n- Fetched `fix/video-editor-speed-render-thrash` (53e3522), inspected `mobile/lib/services/video_editor/clip_speed_render_service.dart` and `mobile/test/services/video_editor/clip_speed_render_service_test.dart`\n- `gh pr checks 6817` — all SUCCESS (Analyze, Format, Generated Files, 4 test shards, Build Web Preview, etc.)\n- Ran `flutter analyze lib/services/video_editor/clip_speed_render_service.dart` — No issues\n- Ran `flutter test test/services/video_editor/clip_speed_render_service_test.dart` — All tests passed (14)\n- Ran `flutter test test/services/video_editor/` — 287 passed when on PR commit\n\n### Concurrency cap — correct placement and logic\n- Cap at `_maxConcurrentRenders = 2` with FIFO `Queue<Completer<void>>` in `ClipSpeedRenderService` is the right choke point. PR description correctly argues not to gate `VideoEditorRenderService.renderNativeVideoToFile` — `renderVideo` re-enters it for seam renders and a semaphore there would deadlock (outer holds slot waiting for inner).\n- Handoff in `_releaseRenderSlot()` (complete next waiter without decrement, keeping `_activeRenders` at cap) is subtle but correct. Verified walk-through: 2 active + 1 queued → release completes waiter (active stays 2, now B+C active) → queued finishes → dec to 1 → last finishes → 0.\n- Stale check after `_acquireRenderSlot()` correctly drops queued backlog when editor closes (`_clearGeneration` bump). That backlog previously kept encoder busy ~20s after teardown (CancellationError in logs). `finally { _releaseRenderSlot(); }` guarantees slot release even on stale bail or throw — good.\n- Minor subtlety (non-blocking): `clear()` does not explicitly clear `_waitingForSlot`. It relies on handoff drain via still-running actives. Works because queue only exists when active==cap, but if the service were reused immediately after `clear()` (same instance for a new editor session) new renders would queue behind stale waiters briefly. Worth a comment or explicit drain; not a ship-blocker since service appears per-editor and CI shows the `drops queued renders once the editor closes` test passes.\n\n### Cache key — dropping `clip.id`\n- Change from `v1|<id>:<path>:` to `v1|<path>:` at `_key()` matches documented intent (“file, trims and speed”). Duplicates/splits/draft reopens mint fresh ids for byte-identical bodies, so old key forced re-encode. New key lets `cached()` and dedup via `_inFlight` share one render — verified by `reuses one render for clips that differ only by id` test.\n- Left orphan: old files under old hash remain in `getTemporaryDirectory()/speed_clips` until OS reaps temp. No functional impact, just transient disk. Fine to leave for temp dir; could note in code comment if desired.\n- Key still includes `duration` + `trimStart`/`trimEnd` + `playbackSpeed` + path. `duration` is redundant given trims, but unchanged from before — no regression, just extra miss on split clips with same trimmed body but different total duration (edge, not introduced here).\n\n### Error handling / resource cleanup\n- `try/finally` around `renderNativeVideoToFile` + stale deletes + `_deleteQuietly` in outer finally covers leaky temp scratch files.\n- In-flight dedup via `_inFlight[key]` still works after key change — now dedups across ids as intended.\n\n### Tests\n- 3 regression tests precisely target the fixes and are asserted to fail without them per PR description:\n  - `caps concurrent native renders so the encoder does not stall` — queues 3rd render\n  - `drops queued renders once the editor closes` — ensures clear abandons queued\n  - `reuses one render for clips that differ only by id` — ensures dedup\n- Existing LRU/eviction/stale-generation tests continue to pass.\n\n### Out of scope handling — good\nExplicitly leaves `clear()` deleting speed bodies on close (vs. persisting), draft-duplicate read-before-write race, and double `ClipEditorBloc` init for separate PRs. Keeps this PR focused.\n\n### Nits (optional follow-up)\n- Consider adding a comment on the handoff invariant (`_activeRenders` stays at cap via transfer) — easy to misread as leak.\n- Consider making `_maxConcurrentRenders` a named constant with device-specific tuning note (PR notes iPad 28-clip case; Android not yet verified).\n- If service lifetime outlives editor, explicitly complete pending `_waitingForSlot` completers in `clear()` to avoid lingering futures (currently drained lazily).\n\nNo blocking issues found. Ready to merge.\n

rabble (approved):
Rabble review: speed-render thrash — bounds concurrency to 2 via Queue<Completer> FIFO, fixing 28-clip draft stalling encoder (progress 0.00). Adds slot acquire/release, prevents RSS 284→640MB. Correct, tested, LGTM.

PR #6819, opened 2026-08-07
fix(engagement): resolve the author key before querying likes, votes and reposts

### LINKED ISSUE #6813: fix(engagement): cold-start likes, vote and repost queries filter on an empty author and return nothing
## Problem

Seven relay queries in `likes_repository` and `reposts_repository` build their filter as `authors: [_nostrClient.publicKey]` with no check that the key is populated.

`NostrClient.publicKey` forwards `Nostr._cachedPublicKey`, which **starts empty** and is only filled by `refreshPublicKey()`:

```dart
// nostr_sdk/lib/nostr.dart:41
// Public key starts empty - call refreshPublicKey() after construction
// to populate from the signer (single source of truth).
```

`NostrClient.initialize()` calls `refreshPublicKey()` first thing, but that is not sufficient — the client's own docs record the case where it resolves and the key is *still* empty:

```dart
// nostr_client/lib/src/nostr_client.dart:482
// ... the "ready resolved but `hasKeys` still false" path (signer started
// empty during cold boot, then `refreshPublicKey()` returned `''`).
```

During that window every one of these queries ships `authors: ['']`. No event has an empty pubkey, so a conformant relay matches nothing and the query returns empty — not an error, just no results.

## Affected call sites

| File | Line | Method |
|---|---|---|
| `mobile/packages/likes_repository/lib/src/likes_repository.dart` | 1110, 1116 | `getUserVoteStatuses` |
| `mobile/packages/likes_repository/lib/src/likes_repository.dart` | 1438, 1444 | `syncUserReactions` |
| `mobile/packages/likes_repository/lib/src/likes_repository.dart` | 2093, 2110 | `_resolveRemoteDownvoteRecord` |
| `mobile/packages/reposts_repository/lib/src/reposts_repository.dart` | 602 | `syncUserReposts` |

That is the complete set — a repo-wide sweep for `authors: [` built from a `publicKey` getter returns these seven and nothing else.

## Why it matters

All four methods are *cold-start recovery* paths, so the window in which the key is unset is exactly the window in which they run:

- `syncUserReactions` / `syncUserReposts` — both documented "should be called on startup to ensure local state matches relay state". They return zero events, so nothing is reconciled against the relay.
- `getUserVoteStatuses` — returns empty upvoted/downvoted sets, so the user's own arrows render unfilled on content they voted on. This is the same visible symptom #6124 fixed for a different cause.
- `_resolveRemoteDownvoteRecord` — exists specifically to let `removeDownvote` "recover from an empty in-memory cache after a cold start". Returning null means the un-downvote finds no record to act on.

None of these surface an error; each degrades to a silently empty result.

## The guard already exists in these files

`hasKeys` (`publicKey.isNotEmpty`) is the established predicate and is already used three times in the same two classes — `likes_repository.dart:1855`, `likes_repository.dart:1863`, `reposts_repository.dart:891`.

The gap is that it guards *one caller* rather than the method. In `LikesRepository.initialize()`:

```dart
if (hasCoordinatelessRecords && _nostrClient.hasKeys) {
  await syncUserReactions();
}
```

`syncUserReactions()` itself has no guard, so the three call sites in `profile_liked_videos_bloc.dart` (lines 217, 282, 439) and the `comment_reactions_bloc.dart:148` call into `getUserVoteStatuses` all bypass it. Same shape in `RepostsRepository.initialize()`, where `_subscribeToReposts()` is guarded but `syncUserReposts()` is not.

The repository providers do not close the gap either: `likesRepository` / `repostsRepository` in `lib/providers/video_providers.dart` are non-nullable and build unconditionally — `authService.currentPublicKeyHex` only gates whether `localStorage` is attached. (Their doc comments claim "Returns null when user is not authenticated", which no longer matches the non-nullable return type — worth correcting while in here.)

## Suggested direction

Move the check into the methods rather than adding a fourth caller-side guard. Two options:

1. Guard on `_nostrClient.hasKeys` and return the empty result early — consistent with the three existing uses, and keeps these methods' "best-effort sync" contract.
2. Use `Nostr.ensurePublicKey()`, which refreshes from the signer and throws `StateError` when it is genuinely unavailable. This is what the send path already does (`nostr.dart:109,123,144,164,180` and `relay_pool.dart:1322`) and would additionally recover the case where the key *is* obtainable but was not cached yet.

Option 2 fixes more cases but changes these methods from silent-empty to throwing, so it needs the callers checked. Option 1 is the minimal, precedent-matching change.

Either way the sync methods should be re-driven once the signer resolves, otherwise a cold start that loses the race just stays unsynced until the next app launch.

## Notes

- Not a regression. A conformant relay has always returned nothing for `authors: ['']`; the queries have been malformed on this path the whole time.
- Found while reviewing #6746 (for #6585) and recorded there as out of scope, since that PR is about relay-list admission.
- No overlap with #6711, the other open PR touching `likes_repository`.

realmeylisdev:
## Description

Seven relay queries in `likes_repository` and `reposts_repository` built their filter as `authors: [_nostrClient.publicKey]` with no check that the key was populated. `publicKey` is a plain read of `Nostr._cachedPublicKey`, which starts empty. No event has an empty pubkey, so a conformant relay matched nothing and the query returned empty — not an error, just no results.

All four affected methods are cold-start recovery paths, so the window in which the key is unset is the window in which they run:

| Method | Symptom |
|---|---|
| `syncUserReactions` | documented "called on startup to reconcile against relay state" — reconciles nothing |
| `syncUserReposts` | same |
| `getUserVoteStatuses` | the user's own arrows render unfilled on content they voted on |
| `_resolveRemoteDownvoteRecord` | exists to recover an empty downvote cache after a cold start — finds no record |

None surfaces an error; each degrades to a silently empty result.

### Why the guard alone wasn't enough

`hasKeys` was already the predicate in both classes (`likes_repository.dart:1855,1863`, `reposts_repository.dart:891`) — but on one *caller* rather than the method, so `profile_liked_videos_bloc`, `profile_reposted_videos_bloc` and `comment_reactions_bloc` all called straight past it.

Adding a fourth caller-side guard would not have fixed the symptom either. The real sequence is: `initialize()` runs → signer still empty → cache stays `''` → the signer acquires its key moments later → **nothing re-runs `refreshPublicKey()`**. A guard would turn "malformed query returns nothing" into "explicitly skipped, still nothing."

So this adds `NostrClient.resolvePublicKey()`, which retries the signer on a cache miss and returns `null` instead of throwing. `Nostr.ensurePublicKey()` stays the right call where the absence of a key is a programming error — the client already documents this exact "ready resolved but `hasKeys` still false" case at `isReadyResolved`.

### Notes for review

- **Concurrent callers share one refresh.** A cache miss reaches the signer, and under NIP-55 `AndroidNostrSigner.getPublicKey()` launches an Amber intent — a user-visible prompt. Three repositories resolving the key at startup must not raise three of them. Pinned by a test, and by a sibling test asserting a *later* miss does refresh again.
- **Initialization shares the same refresh gate.** `NostrClient.initialize()` now routes through the same in-flight refresh, so a startup repository resolve cannot trigger a second signer prompt while initialization is still refreshing the key.
- **Startup subscriptions also resolve the key.** Likes/reposts Kind 7 / Kind 16 subscriptions no longer rely on the stale synchronous cache, so late signer resolution does not leave cross-device updates unsubscribed for the session.
- **`getUserVoteStatuses` returns empty rather than passing a null author.** `_queryVoteTargetReactions` takes `List<String>?`, so a null there would widen the query to every user's votes instead of narrowing it to none.
- **The sync methods keep whatever local storage returned** rather than throwing, matching how they already treat a failed relay fetch. They do not return `.empty()`, which would contradict the in-memory index.
- **A failed signer is logged, not just skipped.** `resolvePublicKey()` returns `null` both when the session is signed out and when the signer refresh threw, and the callers report both as "signer has no public key". The swallow logs the error so a denied or crashed NIP-55 prompt stays distinguishable from a signed-out session — the same non-fatal-swallow shape already used twice in this file.
- **Not a regression.** A conformant relay has always returned nothing for `authors: ['']`; these queries have been malformed on this path the whole time.

## Out of Scope

The sync methods are still not re-driven automatically when the signer resolves *after* a sync already ran and returned early — the next call picks the key up, but nothing schedules that call. Tracked in #6838; it needs a readiness signal these repositories do not currently observe.

## Verification

Each of the original guards is mutation-checked — reverting it turns its own test red, verified individually rather than as a batch.

Post-takeover verification, on the rebased branch (author):

- Full `flutter test` per package after the rebase onto `origin/main` — `nostr_client` 330, `likes_repository` 208, `reposts_repository` 129, all passing
- Bare `flutter analyze` (so `lib` + `test`) clean in all three packages, plus `flutter analyze lib/providers/video_providers.dart` in `mobile`
- Rebase touches nothing shared: `origin/main` gained 10 commits, none of which change `mobile/packages` or `mobile/lib/providers`

Takeover verification:

- `flutter test test/src/nostr_client_test.dart` in `mobile/packages/nostr_client`
- `flutter test test/src/likes_repository_test.dart` in `mobile/packages/likes_repository`
- `flutter test test/src/reposts_repository_test.dart` in `mobile/packages/reposts_repository`
- `flutter analyze` in `mobile/packages/nostr_client`
- `flutter analyze` in `mobile/packages/likes_repository`
- `flutter analyze` in `mobile/packages/reposts_repository`
- `flutter analyze lib/providers/video_providers.dart` in `mobile`

Original author verification:

- `flutter analyze` clean (bare, so `lib` + `test`) in `nostr_client`, `likes_repository`, `reposts_repository`
- `flutter analyze lib test integration_test` clean in `mobile`
- `flutter test` — `nostr_client` 328, `likes_repository` 206, `reposts_repository` 127, all passing
- App-level consumers of these paths: `comment_reactions_bloc_test`, `profile_liked_videos_bloc_test`, `profile_reposted_videos_bloc_test`, `profile_grid_test` — 108 passing
- Coverage above gate: `nostr_client` 87.63% (gate 82), `likes_repository` 97.00% (gate 62)
- Checked that all 8 hand-written `implements NostrClient` fakes declare `noSuchMethod`, so the new member does not break them

**Related Issue:** Closes #6813

## Type of Change

### DIFF
--- a/mobile/lib/providers/video_providers.dart
+++ b/mobile/lib/providers/video_providers.dart
@@ -549,8 +549,8 @@ VideosRepository videosRepository(Ref ref) {
 
 /// Provider for LikesRepository instance
 ///
-/// Creates a LikesRepository when the user is authenticated.
-/// Returns null when user is not authenticated.
+/// Creates a LikesRepository. Local storage is attached when the user is
+/// authenticated; signed-out relay queries are guarded by the repository.
 ///
 /// Uses:
 /// - NostrClient from nostrServiceProvider (for relay communication)
--- a/mobile/lib/providers/video_providers.g.dart
+++ b/mobile/lib/providers/video_providers.g.dart
@@ -942,8 +942,8 @@ String _$videosRepositoryHash() => r'f98257a666f2fede19dd41cbc4e5dca57ae65328';
 
 /// Provider for LikesRepository instance
 ///
-/// Creates a LikesRepository when the user is authenticated.
-/// Returns null when user is not authenticated.
+/// Creates a LikesRepository. Local storage is attached when the user is
+/// authenticated; signed-out relay queries are guarded by the repository.
 ///
 /// Uses:
 /// - NostrClient from nostrServiceProvider (for relay communication)
@@ -954,8 +954,8 @@ final likesRepositoryProvider = LikesRepositoryProvider._();
 
 /// Provider for LikesRepository instance
 ///
-/// Creates a LikesRepository when the user is authenticated.
-/// Returns null when user is not authenticated.
+/// Creates a LikesRepository. Local storage is attached when the user is
+/// authenticated; signed-out relay queries are guarded by the repository.
 ///
 /// Uses:
 /// - NostrClient from nostrServiceProvider (for relay communication)
@@ -967,8 +967,8 @@ final class LikesRepositoryProvider
     with $Provider<LikesRepository> {
   /// Provider for LikesRepository instance
   ///
-  /// Creates a LikesRepository when the user is authenticated.
-  /// Returns null when user is not authenticated.
+  /// Creates a LikesRepository. Local storage is attached when the user is
+  /// authenticated; signed-out relay queries are guarded by the repository.
   ///
   /// Uses:
   /// - NostrClient from nostrServiceProvider (for relay communication)
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -1439,32 +1439,32 @@ class LikesRepository {
       _emitLikedIds();
     }
 
-    // Signed out: keep whatever local storage gave us rather than throwing —
-    // there is nothing to reconcile against, which is not a sync failure.
-    final pubkey = await _currentUserPubkey();
-    if (pubkey == null) {
-      Log.warning(
-        'Skipping reaction sync: signer has no public key',
-        name: 'LikesRepository',
-      );
-      _isInitialized = true;
-      return _buildSyncResult();
-    }
-
-    // Fetch both reactions and deletions from relays (authoritative)
-    final reactionsFilter = Filter(
-      kinds: const [EventKind.reaction],
-      authors: [pubkey],
-      limit: _defaultReactionFetchLimit,
-    );
-
-    final deletionsFilter = Filter(
-      kinds: const [EventKind.eventDeletion],
-      authors: [pubkey],
-      limit: _defaultReactionFetchLimit,
-    );
-
     try {
+      // Signed out: keep whatever local storage gave us rather than throwing —
+      // there is nothing to reconcile against, which is not a sync failure.
+      final pubkey = await _currentUserPubkey();
+      if (pubkey == null) {
+        Log.warning(
+          'Skipping reaction sync: signer has no public key',
+          name: 'LikesRepository',
+        );
+        _isInitialized = true;
+        return _buildSyncResult();
+      }
+
+      // Fetch both reactions and deletions from relays (authoritative)
+      final reactionsFilter = Filter(
+        kinds: const [EventKind.reaction],
+        authors: [pubkey],
+        limit: _defaultReactionFetchLimit,
+      );
+
+      final deletionsFilter = Filter(
+        kinds: const [EventKind.eventDeletion],
+        authors: [pubkey],
+        limit: _defaultReactionFetchLimit,
+      );
+
       // Fetch reactions and deletions in parallel
       final results = await Future.wait([
         _nostrClient.queryEvents([reactionsFilter]),
@@ -1870,16 +1870,16 @@ class LikesRepository {
       _emitLikedIds();
     }
 
-    // Subscribe to reactions for real-time sync and cross-device updates
-    if (_nostrClient.hasKeys) {
-      _subscribeToReactions();
+    final pubkey = await _currentUserPubkey();
+    if (pubkey != null) {
+      _subscribeToReactions(pubkey);
     }
 
     // Flip the flag before the backfill sync so _ensureInitialized callers
     // aren't blocked on a relay round-trip.
     _isInitialized = true;
 
-    if (hasCoordinatelessRecords && _nostrClient.hasKeys) {
+    if (hasCoordinatelessRecords && pubkey != null) {
       try {
         await syncUserReactions();
       } on Exception catch (_) {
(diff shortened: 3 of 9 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I pushed a takeover commit (`e508953db`) on top of your branch to close out the review findings:

- shared `NostrClient.initialize()` with the same in-flight public-key refresh gate used by `resolvePublicKey()`
- made `resolvePublicKey()` return `null` on signer refresh errors for repository callers
- resolved the pubkey before likes/reposts startup subscriptions so stale cache no longer skips Kind 7 / Kind 16 live updates for the session
- moved sync pubkey resolution inside the existing sync failure handling/fallback path
- fixed the stale `likesRepository` provider docs and committed the generated Riverpod doc update
- filed #6838 for the remaining signer-late sync re-drive follow-up and linked it from the PR body

Verification is green locally and in CI. Please review the takeover commit and merge when you are comfortable with it.

NotThatKindOfDrLiz (approved):
Approved after takeover. I pushed e508953db to address the startup signer race, subscription setup, sync fallback, and provider doc issues; local targeted verification and all GitHub checks are green.

NotThatKindOfDrLiz:
Review completed — comprehensive analysis of cold-start signer-readiness race fixes.

Core design is sound and well-tested. CI fully green (17 checks). No new tech debt in the change itself.

However, integration reveals 4 mandatory issues blocking the "no accumulated tech debt" bar:

MANDATORY (block merge):
1. Issue #7931 — resolvePublicKey() calls outside try-catch violate SyncFailedException contract. Signer errors bypass error handlers.
2. Issue #7930 — initialize() bypasses concurrent-dedup gate, causing Amber prompt duplication on async signers.
3. Issue #7932 — Provider documentation debt from issue #6813 not completed.
4. Test coverage gap for throwing-signer path in resolvePublicKey().

DEFERRED (file issues, not blocking):
5. Issue #7933 — Subscription gates don't recover when signers resolve after init.
6. Issue #7934 — Add test for throwing-signer case.

Recommendation: File #7930–#7934 and address mandatory issues as follow-ups before merge. Each is isolated and specific.

PR #6820, opened 2026-08-07
feat(support): disclose public GitHub submissions

### LINKED ISSUE #6821: feat(support): disclose public GitHub submissions
Auto-created from PR #6820.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6820
**Branch:** `feat/public-submission-info-banner`

hm21:
Closes #6821

## Description

Adds a warning card to the bug report and feature request forms explaining that submissions are published in Divine's public open-source GitHub repository so developers can act on them. The disclosure uses the warning tone because user-provided content becomes publicly visible, and the copy is localized for every supported locale.

**Related Issue:** 

## Out of Scope

None.

## Verification

- `flutter test test/widgets/bug_report_dialog_test.dart test/widgets/feature_request_dialog_test.dart test/l10n/arb_consistency_test.dart`
- `scripts/golden.sh verify`
- Manually verified both support flows on a real Samsung Galaxy device

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mobile/lib/widgets/bug_report_dialog.dart
```diff
@@ -172,6 +173,7 @@ class _BugReportView extends StatelessWidget {
                     crossAxisAlignment: CrossAxisAlignment.stretch,
                     spacing: 16,
                     children: [
+                      const SupportPublicSubmissionNotice(),
```
mbradley, line 176: The user's account is attached to the submission and lands in the public issue (`zendesk_support_service.dart:913`, and `:1115` for feature requests). Neither screen's copy mentions it. On this screen `bugReportDiagnosticsNotice` covers device info and logs; the feature request screen has no equivalent line.

Since the PR is about disclosing what becomes public, this needs a call from you. Three options:

1. Add the account to this card, so one string covers both screens. One English change plus 22 retranslations.
2. Add it to `bugReportDiagnosticsNotice` here and add an equivalent line to the feature request screen. Two strings, same retranslation cost.
3. Leave it, on the basis that a pubkey is already public and users expect their account on a support request.

3 is a defensible answer. I would just rather it be a decision than an omission.

mbradley (changes requested):
Reviewed and took over three small findings. Each is its own commit so you can drop any of them independently.

**What I changed**

- `9281fd1` The card was duplicated verbatim in both forms. Pulled it into a shared `SupportPublicSubmissionNotice`, matching how the other shared parts of these two forms already work (`support_failure_banner.dart`, `support_form_actions.dart`, `support_form_fields.dart`). Worth sharing because the text is a warning that has to say the same thing on both screens, and two copies let one drift from the other with nothing to catch it.
- `a1841c1` `app_ja.arb` puts spaces around Latin words throughout (`"Nostr 設定"`, `"Nostr について"`); the two new values did not. Regenerated l10n, which touched `app_localizations_ja.dart` only.
- `ffa6d2a` The new tests checked the card exists but not where it is, so a later layout change could push it below the fields and the picker with both tests still green. Added a position check on both forms.

**Validation**

`flutter analyze lib test` clean. The three affected test files pass (32 tests). I checked the new position test actually fails by moving the card below the fields. No goldens cover these screens, so nothing to regenerate.

One CI note. On my first push `Tests (shard 0/4)` failed on `notification_rows_golden_test.dart` ("render max-font layout", RenderFlex overflow). That was not caused by these commits: the shard script partitions by test file and I added no test files, my two touched test files mutate no process globals, and the golden passes both alone and alongside my changes. CI runs `--test-randomize-ordering-seed random`, and a re-run of the same commit under a new seed passed. It looks like an order-dependent global-state leak from another suite in that shard, of the kind `.claude/rules/testing.md` describes. All checks are green now, but it will resurface on an unlucky seed and is worth its own issue.

**One question below** that I do not think I should decide for you.

Have a look at the three commits and merge when you are happy, or push back on any of them.

mobile/lib/widgets/bug_report_dialog.dart
hm21: @mbradley Thanks for the review. Correct me if i'm wrong but i'm pretty sure we don’t include device information or logs in github feature requests. You can see an example of a feature request [here](https://github.com/divinevideo/divine-mobile/issues/6418), where no device information or logs are included.

If relevant, you can also see an example of a bug report [here](https://github.com/divinevideo/divine-mobile/issues/6764).

mbradley: What I said came off differently than I meant. I meant to just point out that, particularly given my question about if we want to tell people what is being submitted - including their identity, we might need a similar statement on Feature Request (but accurate for what gets submitted with the Feature Request).

The narrower thing I'd still like a call on: both flows attach the user's identity. #6418 shows it, a feature request with no device info and no logs, still carrying a `**User Pubkey:**` line.

Would we want a specific mention that their identity is attached too?

mbradley (approved):
`c2d2df8` covers it. Checked all 22 locales carry the change, generated files are in sync, and the earlier commits are intact. CI green.

Approving.

PR #2180, opened 2026-03-12
feat(local_stack): integrate divine-blossom local mode into e2e stack

dcadenas:
## Summary
- Replace generic `hzrd149/blossom-server` with divine-blossom running locally via Viceroy (Fastly Compute local runtime) + MinIO for S3-compatible storage
- Add proper Nostr kind 24242 auth to seed service and Dart e2e test upload helpers, matching prod auth behavior
- Add `pull_policy: always` to all Docker services to ensure fresh images on each run
- Remove unused `blossom-config.yml` (was for generic blossom server)

Depends on divinevideo/divine-blossom#25

## What this enables
- Local e2e tests now get real responses from HLS (`.hls`), subtitle (`.vtt`), and thumbnail (`.jpg`) endpoints instead of 404s
- Upload auth matches prod (kind 24242 Nostr events with BIP-340 Schnorr signatures)
- Moderation, transcoding, and transcription are stubbed inline with immediate responses — zero external dependencies

## Test plan
- [ ] `mise run local_up` starts full stack including MinIO + divine-blossom
- [ ] Seed service uploads with kind 24242 auth and publishes 100 video events
- [ ] `mise run e2e_test` passes against new blossom backend
- [ ] Verify HLS/VTT endpoints return 200 (after initial 202) instead of 404

PR #6823, opened 2026-08-07
fix(local-stack): make a current funnelcake stack obtainable and staleness visible

### LINKED ISSUE #6594: fix(local_stack): pinned GHCR images are 5 months stale, so NIP-17 is untestable locally

realmeylisdev:
## Problem

`local_stack` pins `ghcr.io/divinevideo/funnelcake-{migrate,relay,api}:latest`. Those tags are frozen at **2026-02-24** — schema 70 against `origin/main`'s 210 — so the relay's kind allowlist predates `000079_allow_gift_wrap_kind` and `000168_allow_dm_inbox_kind`. Kind 1059 and kind 10050 are rejected outright and NIP-17 DMs cannot be exercised locally at all. `docker compose pull` reports "Pulled" while changing nothing, so none of this is visible.

The root cause is **not in this repo**, and correcting the issue's framing: the mirror did not stop publishing, it never published. `divine-funnelcake`'s "Push to GHCR (public)" step has failed with `denied: permission_denied: write_package` on **every run since it was added on 2026-03-07**, and `continue-on-error: true` makes the GitHub jobs API report the failing step as `conclusion: success`. The images in GHCR were seeded by a manual laptop push (config `.created` carries a `-03:00` offset; runners are UTC). Each package holds exactly two tags, `latest` and the 7-char `90ea375`, sharing one digest, while the workflow only ever pushes a 40-hex `${{ github.sha }}` that resolves 404. Filed as divinevideo/divine-funnelcake#844.

Nothing here can fix that — divine-mobile is a pure consumer and its pulls already succeed. What this PR does is make a correct stack obtainable, make the staleness loud, and fix the two regressions a refresh would otherwise cause.

## Changes

| | |
|---|---|
| `build_funnelcake.sh` + `${FUNNELCAKE_*_IMAGE}` overrides | Build the three images from a local `divine-funnelcake` checkout using that repo's own Dockerfiles. Mirrors the existing `${KEYCAST_IMAGE}` escape hatch. |
| Relay rate-limit env | Without it, a current relay rejects 90 of `seed.py`'s 100 events and `up.sh` fails for everyone. |
| `tuning/refresh_intervals.sql` | `video_stats` has been snapshot-backed since `000143` and widened to 15 minutes by `000182`/`000184`; the e2e polls it with a 30 s budget. |
| `preflight_image_staleness` | Warns when the pinned images are frozen. Anonymous GHCR, zero credentials, fails open. |
| Docs | Record the PR #2180 rolling-tag decision, document the overrides, correct a false `AGENTS.md` claim. |

## Verification

Built from `divine-funnelcake@172b0d36` and ran the real stack:

```
migration apply complete: bootstrapped=0, applied=200     # ClickHouse 25.8, 23s
refresh-interval tuning: applied=5 skipped=0
Published 100/100 events (0 failed)                        # real seed.py
funnelcake schema version: 210
```

| | pinned | locally built |
|---|---|---|
| kind 1059 | `blocked: kind 1059 not in allowed list` | **accepted** |
| kind 10050 | `blocked: …` | **accepted** |
| NIP-11 `supported_nips` | 16 | **24 — identical to production** |
| `allowed_kinds` | 23 | 40 |
| `seed.py` 100 events | ok=100 | ok=100 (ok=10/rejected=90 **without** the rate-limit fix) |
| `/api/users/{pk}/videos` | — | 16 videos in **5 s** |
| architecture | amd64, emulated | native arm64 |

The staleness check reports the three funnelcake images at 163 days / 2 tags and warns, while `keycast` (1 day, 107 tags) stays silent. `test_stack_scripts.sh` covers it; a never-warn mutation and a fail-closed mutation both turn those cases red.

Two bugs the end-to-end run caught in my own earlier commits are fixed in `1b56fbc53`: the schema report asked for the wrong migrator ledger, and the build guidance recommended per-shell exports, which wedges the database with `Dirty database version N` if any compose command runs without them.

## What this does NOT do

- **The default path stays stale.** Only the opt-in build gives a fresh stack. A follow-up will repoint the defaults once funnelcake's GHCR publishing is fixed.
- **No CI signal changes.** The only `service`-tagged test is the DM one; its workflow has never executed (`workflow_dispatch`-only, `total_count: 0`), it passes no `--dart-define=DEFAULT_ENV=LOCAL` so it would resolve production endpoints, and a `patrolTest` cannot run under plain `flutter test`. The gain is to the local dev harness, not CI.
- The build path needs a `divine-funnelcake` checkout (private repo). Cold-cache build time is unmeasured — 83 s is the warm figure.

## Test plan

- `bash local_stack/test_stack_scripts.sh` — green, including 4 new staleness cases.
- Full `docker compose up` on locally built images: all services healthy, seed 100/100, kinds 1059/10050 accepted through the proxy, `/api/users/{pk}/videos` inside the e2e budget.
- Tuning verified as a no-op on the pinned schema (`applied=0 skipped=5`) and effective on current (`applied=5 skipped=0`).
- No Dart changed, so `flutter analyze` and the l10n check do not apply.

Closes #6594

## Filed separately

- divinevideo/divine-funnelcake#844 — the GHCR root cause (package linkage + the swallowed failure).
- #6824 — `localHost` is hardcoded to `10.0.2.2`, so LOCAL cannot reach the stack from the iOS Simulator.
- #6825 — a cached production relay appears in the relay set while running in LOCAL.
- #6826 — follow-up to repoint the default pins once funnelcake#844 lands.

## Cross-repo verification

Re-verified every claim against each related repo at current `origin/main` (funnelcake `172b0d36`, unmoved since I measured). Two defects found and fixed in `27f689aea`:

- `video_event_card_snapshot_refresh_mv` was wrongly in the tuning list — its source rebuilds every 10 minutes regardless, and it is the one MV carrying `DEPENDS ON`. Now five statements, with `relay_feed_cache_refresh_mv` left at its upstream cadence.
- The header credited `000160` for `video_stats` ceasing to be live; it was `000143`.

Confirmed holding: every `ci.yml` line number cited in funnelcake#844; the rate-limit defaults and that `MAX_EVENTS_PER_MINUTE_PER_PUBKEY=0` really disables the check (`rate_limit.rs:184` guards on `> 0`); that the rate limiter postdates the pinned image (added `ceee86b4`, 2026-04-15); migrations `000079`/`000168`/`000015`; all remaining MV names; the Dockerfile stage layout the build script depends on; and blast radius (two non-markdown files org-wide).

**Both paths regression-tested end to end.** Default pinned path: services healthy, migrate through 70, schema reports `70` from the legacy ledger, staleness warning fires, seed 100/100, tuning `applied=0 skipped=5`. Built path: `applied=200`, schema `210`, tuning `applied=5 skipped=0`, seed 100/100, kinds 1059/10050 accepted through the proxy, `/api/users/{pk}/videos` in 5 s.

Two things worth knowing that are **not** caused by this PR:
- Kind **24133 (NIP-46)** is rejected by the divine relay on *both* schemas — no migration has ever added it, it is outside both auto-allow ranges, and production does not advertise NIP-46. Local keycast login is unaffected because it uses keycast's HTTP API on :43000.
- Production runs funnelcake `727c303` (197 migrations), only 14 commits behind `origin/main` — so building from main lands essentially at production parity, versus the pinned image's 130-migration gap.


NotThatKindOfDrLiz:
@realmeylisdev I pushed takeover commit `d1c13cbce` to address the review blockers:

- `up.sh` and `build_funnelcake.sh` now source `local_stack/.env`, so the recommended `.env` override path suppresses the stale-image warning and `DIVINE_FUNNELCAKE_ROOT` works there.
- The staleness warning, `.env.example`, and build script usage now all recommend the same `.env` flow.
- `funnelcake-local-tuning` is profile-gated so bare `docker compose up -d` does not start the one-shot container.
- Tuning now leaves `relay_feed_cache_refresh_mv` at upstream cadence, applies five local visibility ALTERs, and warns if a current schema applies fewer than expected.
- The SQL comment now explicitly says the local ALTERs intentionally replace production `EVERY ... OFFSET` schedules and must not be copied into funnelcake migrations.
- PR title/body were updated to match the final state.

Verification passed:
- `bash local_stack/test_stack_scripts.sh`
- `shellcheck -x -P local_stack local_stack/up.sh local_stack/preflight.sh local_stack/build_funnelcake.sh local_stack/test_stack_scripts.sh`
- `git diff --check`
- `mise run test_local_stack_scripts` (passed; mise printed a non-fatal sandbox warning while tracking trusted config)
- GitHub checks are green on `d1c13cbc`, including the rerun semantic PR check after the title fix.

NotThatKindOfDrLiz (approved):
Approved after takeover.

I pushed `d1c13cbce` to address the remaining local-stack blockers, updated the title/body, and verified the final PR state.

Local verification:
- `bash local_stack/test_stack_scripts.sh`
- `shellcheck -x -P local_stack local_stack/up.sh local_stack/preflight.sh local_stack/build_funnelcake.sh local_stack/test_stack_scripts.sh`
- `git diff --check`
- `mise run test_local_stack_scripts`

GitHub checks are green on the current head, including the rerun semantic PR check after the title fix.

PR #6827, opened 2026-08-07
fix(video-metadata): settle the cover picker's entrance

### LINKED ISSUE #6828: fix(video-metadata): settle the cover picker's entrance
Auto-created from PR #6827.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6827
**Branch:** `fix/cover-picker-entrance`

hm21:
Closes #6828

## Description

Opening the cover picker flashed in four distinct ways. Each had its own cause, so they are separated into two commits.

**The cover the picker opens on was a frame of a different file.** The rendered clip inherited `thumbnailPath` from the first source clip and carried no `thumbnailTimestamp`, so its cover was a frame of the raw recording — before trim, speed, crop and filters — while the fallback timestamp pointed 210 ms into the *rendered* file. The picker showed the inherited thumbnail via the hero and then jumped to whatever sits at 210 ms. With a trim this stopped being cosmetic: the published cover could be a frame the video no longer contains. The cover is now extracted from the render output, which pairs path and timestamp by construction — the same thing the picker already does when the user confirms a frame by hand. If extraction fails it falls back to the inherited thumbnail, so a render is never left without a preview.

**The preview loaded at 0 and seeked afterwards.** `setSource` prepares at position 0, so the decoder's first frame — and `firstFrameRendered` with it — landed on frame 0, and the seek to the cover position only moved it after the texture was already on screen. `setClips` takes a `startPosition` that both platforms honour before reporting ready (Android `setMediaItems(items, startIndex, startLocalMs)` before `prepare()`, iOS seek + `forceRefresh` + preroll before `"ready"`), so the first decoded frame is the selected one.

**The player was hard-cut in over the hero's thumbnail.** `DivineVideoPlayer` skips its placeholder entirely when the first frame is already decoded at mount, which is exactly the case here — the thumbnail was already on screen because the hero flew it in. `crossFadePlaceholder: true` keeps it for the hand-off, the same reason the video editor sets it.

**The strip flipped in one frame.** The priority timestamps cover every slot, so the first batch replaces the stretched cover thumbnail in every tile at once. Tiles now fade their frame in over that fallback. The fade flag is sticky on purpose: a later batch reassigning a slot keeps the previous frame up via `gaplessPlayback` instead of dipping back to the fallback while the new file decodes. This needed a `frameBuilder` passthrough on `ClipThumbnailImage` — its `placeholder` is an error path, not a loading state, so there was no hook for a fade.

**The top bar was painted over by the flying video** and popped back in front when the hero landed, because the hero flies in the navigator's overlay, above everything the route itself paints. The bar now waits for the entrance transition and fades in. It follows `CodecHeavySurfaceGuard`: the route is read from a post-frame callback, not during build, because on the first build `ModalRoute.of` still resolves to the route being pushed *from*, whose transition has long completed — verified on device, where the first reading reported `completed` before the second reported `forward`. Opacity 0 already drops the buttons from the semantics tree, but it still hit tests, so taps are blocked separately.

Smaller: the areas either side of the strip selection are dimmed, and the confirm spinner is `BrandedLoadingIndicator` instead of a raw `CircularProgressIndicator`.

**Related Issue:** 

## Out of Scope

The page cross-fade on both push sites stays as it is. While the incoming page is translucent the screen underneath shows through, and its header sits at the same position as the cover screen's, so buttons are briefly visible there too. Removing the cross-fade fixes that (both screens share `surfaceContainerHigh`, so the swap would be invisible), but it changes the feel of the transition and was not part of what was verified here.

## Verification

- `flutter test test/screens/video_metadata_cover_screen_test.dart test/widgets/video_clip test/widgets/video_metadata test/services/video_editor/video_editor_render_service_test.dart test/services/video_editor/video_editor_render_service_progress_test.dart test/providers/video_editor_provider_test.dart test/models/divine_video_draft_cover_thumbnail_test.dart` — 354 passing
- New tests: `setClips` carries `startPositionMs` and no separate `seekTo` fires on init; the top bar is at opacity 0 mid-push and 1 after the transition lands (pushed through a real route with a matching source hero, so an actual flight runs); `ClipThumbnailImage` wraps the image with `frameBuilder` while decoding and drops it for the placeholder when decoding fails.
- No test for `renderVideoToClip` — it sits behind ProofMode and the native renderer, so a test there would be mock scaffolding rather than evidence.
- Manually verified on a real Samsung Galaxy: open the picker from the metadata screen with and without a manually chosen cover, scrub the strip, confirm, reopen.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
@@ -874,10 +874,7 @@ class _ThumbnailStripState extends State<_ThumbnailStrip> {
                   Positioned(
                     top: 0,
                     bottom: 0,
-                    left: (cursorDx - _cursorWidth / 2).clamp(
-                      0,
-                      stripWidth - _cursorWidth,
-                    ),
+                    left: cursorLeft,
                     child: IgnorePointer(
                       child: Container(
                         width: _cursorWidth,
--- a/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
@@ -108,8 +108,15 @@ class _VideoMetadataCoverScreenState
         await controller.dispose();
         return;
       }
-      await controller.setSource(VideoClip.file(localPath));
-      if (mounted) await controller.seekTo(_selectedPosition);
+      // Prepare directly at the cover position instead of loading at 0 and
+      // seeking afterwards: the decoder's very first frame is then already
+      // the selected one. Loading-then-seeking renders frame 0 first, which
+      // flashed the wrong frame for the length of the seek right after the
+      // hero landed.
+      await controller.setClips(
+        [VideoClip.file(localPath)],
+        startPosition: _selectedPosition,
+      );
       if (mounted) {
         setState(() {
           _controller = controller;
@@ -472,6 +479,11 @@ class _VideoAreaState extends State<_VideoArea> {
           child: DivineVideoPlayer(
             controller: widget.controller,
             placeholder: _buildPlaceholder(),
+            // The thumbnail is already on screen (it is what the hero flew
+            // in) and the player is swapped in only once its first frame is
+            // decoded. Without this the widget hard-cuts thumbnail→texture,
+            // which reads as a flicker; cross-fade instead.
+            crossFadePlaceholder: true,
           ),
         ),
       ),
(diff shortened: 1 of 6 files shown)


mbradley (approved):
Took this over under the shared-ownership policy and pushed eight commits, one per finding. Everything below was verified against `f804ae23c` first-hand rather than by inspection, and four of the fixes are mutation-proven.

**Two defects worth the most attention.**

The `BrandedLoadingIndicator` swap left both confirm-spinner assertions naming `CircularProgressIndicator`, which the screen no longer builds, so both went permanently green. That removed the only guard on the confirming state. Proof: delete the `_isConfirming` reset in `_confirm` and the old assertion still passes, while the repointed one fails.

Separately, unmounting the screen while `setClips` is in flight left the controller unassigned, so `dispose()` ran against a null `_controller` and the native player, its surface and a hardware decoder were never released. Nothing else reclaims them. The controller stays in the static live-controller set for the rest of the process, and the only bulk cleanup runs at app startup against a previous VM. This one is pre-existing on `main`, where the same gap spanned two awaits, so your change narrows the window rather than opening it. Fixed here because the surrounding lines moved anyway, mirroring the guard you already use one await earlier.

**Three smaller ones.**

`AnimatedOpacity` fires `onEnd` synchronously when its duration is zero and the target changed, so `_dropFallback` ran mid-build and marked `_FadingSlotImage` dirty while the `Image` beneath it was still building. Reproduced end to end by warming a path in the image cache, mounting a slot on a still-decoding path, then reassigning to the warm one. Worth being precise about severity: the guard is an assert, stripped in profile and release, and the listener is caught to `FlutterError.reportError`, so there is no user-visible effect in any shipped build. Deferring it also lets the synchronous-load branch share one path.

The tile fade ran unconditionally while the top bar in the same screen already gated on `disableAnimationsOf`, and the strip fades every slot at once.

`BrandedLoadingIndicator` renders an `Image.asset` without `excludeFromSemantics`, so it merged an image node into the confirm button's `Semantics` while saving. The `CircularProgressIndicator` it replaced contributed no node at all, so that flag is new. Fixed at the call site; excluding it inside the widget is the durable fix but reaches other screens.

**One comment correction.** The entrance comment says `ModalRoute` resolves to the route being pushed from during the first build. It does not, it resolves to the correct route. The animation reads `completed` because Navigator renders a newly pushed route offstage on its first frame so heroes can be measured, and `ModalRoute` swaps in `kAlwaysCompleteAnimation` while it is. Same conclusion, different mechanism, and the post-frame callback is load-bearing for two independent reasons.

**Added coverage.** Pinned the cover fallback in the existing progress test, which already drives `renderVideoToClip` end to end. That needed a `thumbnailPath` on the fixture, without which the assertion would compare null to null and never fail. Covered the two dim overlays and the `cursorLeft` clamp, which had no assertion anywhere; that one is driven by taps rather than the slider semantics, which do not move the cursor until the preview player resolves. Also stopped the clip-thumbnail tests leaking failed image-cache entries, which under the merged-isolate run outlive the file and reach later tests on the same path.

**Validation.** `flutter analyze lib test integration_test` clean. The suite from your PR description now at 355 passing. Also ran randomized ordering. CI green on all checks.

**Remaining uncertainty, stated plainly.** Three of my own changes ship without tests: the `ExcludeSemantics` wrapper, the reduced-motion gate, and the defer fix. All three hit the same harness limits. The confirming state is not observable because `_confirm`'s awaits resolve in microtasks before the first frame, and `_FadingSlotImage` never renders because `getThumbnails` returns `[]`. The defer fix was verified by reproduction rather than a committed test.

**One question for you.** The new `extractThumbnail` await sits after the final progress emit, which lands exactly `1.0`, so the bar reaches 100 percent and then the scrim stays up for the extraction. There is no timeout, while the sibling call in `video_event_publisher.dart` has a 10 second ceiling. The catch is that a timeout is not free here: if it fires, `cover` is null and the fallback reproduces the pre-fix state exactly, including the 210ms. So bounding the stall trades away the correctness this PR just bought on slow devices. A third option may be better than either: move the extraction before the final progress emit and reserve a slice of the budget for it, which fixes the stalled bar without the tradeoff. Your call on which.

Please review the pushed commits and merge when you are happy, or push back on any of them.

PR #6829, opened 2026-08-07
fix(video-metadata): keep the preview rounded and make the audio credit form readable

### LINKED ISSUE #6830: fix(video-metadata): keep the preview rounded and make the audio credit form readable
Auto-created from PR #6829.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6829
**Branch:** `fix/metadata-design-issues`

hm21:
Closes #6830

## Description

Three design issues on the video metadata flow.

**The preview arrived square-bottomed.** Opening the full-screen preview showed no rounding along the bottom edge for a moment, then it snapped in. The stage's rounded bottom comes from a `ClipRRect` above the `Hero`, but a flying hero is lifted into the navigator overlay — outside that clip — so nothing rounded the preview for the whole ~300ms flight. The flight now clips itself via `flightShuttleBuilder`, morphing out of the clip thumbnail's all-round corners into the stage's bottom-only rounding. The flight animation runs 0 (metadata screen) → 1 (preview) in *both* directions, because a pop drives it from the popped route's animation, so one lerp covers the way back too. The thumbnail's radius moved into `VideoEditorConstants.clipPreviewCornerRadius` so the flight's start shape can't drift away from the shape it starts on.

**The audio credit fields didn't look like fields.** Title, creator, source URL and hashtags sit together inside a single `_InputWrapper` card. `DivineTextField` is transparent by default and takes the colour of whatever it sits on, so each one read as a label — unlike the title and description fields further up, where every field gets its own card and the card *is* the visible input surface. They're `filled: true` now, which is the case the widget's own API documents. The "shared as" preview is separated by a top border instead of floating in the same block, and the optional rows (provider credit vs public editor, the source field that appears when "own work" is off, the validation line) cross-fade through a shared `_ExpandSwitcher` rather than snapping the card's height.

**Inset drift.** Material 3 gives `ListTile` an asymmetric `contentPadding` — start 16, end 24 — and both `DivineSwitchTile` and `DivineCheckboxTile` were falling through to it. 16 on both edges is the value the design system wants, so both now pin `contentPadding` to a symmetric 16. The leading edge doesn't move; the trailing edge tightens by 8px, and it does so everywhere those tiles are used, not just here — 16 switch rows across Safety, Notification, General, Nostr, Bluesky, Content preferences, Monetization links and Blossom settings plus the metadata form, and 2 checkbox rows in Safety settings and the audio credit card. It's a design-system correction rather than a local one; it just surfaced here, where a switch row and a checkbox row sit in the same card and the mismatch is visible. The cover screen's top bar separately drops its vertical padding so its horizontal inset matches the form.

The audio sharing section also moved into its own file. It carried three classes and five imports that nothing else in `video_metadata_form_fields.dart` needed, and that file had grown past 500 lines.

**Related Issue:** 

## Out of Scope

None.

## Verification

- `flutter test` on `video_metadata_cover_screen_test.dart`, `video_metadata_preview_screen_test.dart`, `video_metadata_screen_test.dart`, `test/widgets/video_metadata/` — 202 passing
- `flutter test` on `packages/divine_ui` — 845 passing
- All four design-system ceiling ratchets, the untested-services floor and the `test/unit` structure guard — pass
- Flight tests: `morphs the hero flight corners in both directions` pins the lerp's direction against a reversed and a constant-pinned mutant; `the real capture thumbnail rounds itself outside its Hero` flies from the real thumbnail widget and pins the clip that a pop would otherwise drag into the shuttle
- Two existing tests adapted to the new behaviour, both still asserting the same user-visible outcome: `requires public creator and source for imported audio` now settles before asserting the validation line is gone (it fades out over the switcher instead of vanishing next frame), and the form-field toggle tests target `DivineSwitchTile` instead of Material's `SwitchListTile`
- Manual pass on a real Samsung Galaxy (SM-S942B): open metadata → preview (corners rounded from the first frame of the flight and on the way back), toggle "Allow others to remix", toggle "own work" on/off, fill the credit until the validation line clears — all good
- Settings screens that use the two tiles re-run after the padding change: `safety_settings_screen_test.dart`, `settings_screen_audio_sharing_test.dart`, `test/screens/settings/` — 339 passing together with the metadata suites

### Screens on device (SM-S942B, `f3f4747`)

General Settings, where the tile change lands outside the metadata flow. Rows carrying a control (the switches) sit at 16 on both edges; navigation rows keep Material's start 16 / end 24, which is the split this PR settles on rather than flattens.

General Settings after the tile inset change

The audio credit card, the surface the form changes target.

Post details audio credit card after the form changes

## Type of Change

- [x] 🧹 Code refactor

### DIFF
--- a/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
@@ -626,7 +626,7 @@ class _TopBarContent extends StatelessWidget {
       child: SafeArea(
         bottom: false,
         child: Padding(
-          padding: const .all(16),
+          padding: const .symmetric(horizontal: 16),
           child: Row(
             spacing: 16,
             children: [
--- a/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
@@ -625,7 +625,8 @@ class _TopBarContent extends StatelessWidget {
       alignment: .topCenter,
       child: SafeArea(
         bottom: false,
-        child: Padding(
+        child: Container(
+          height: kToolbarHeight,
           padding: const .symmetric(horizontal: 16),
           child: Row(
             spacing: 16,
--- a/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
@@ -15,6 +15,7 @@ import 'package:openvine/providers/video_editor_provider.dart';
 import 'package:openvine/services/video_thumbnail_service.dart';
 import 'package:openvine/widgets/branded_loading_indicator.dart';
 import 'package:openvine/widgets/video_clip/clip_thumbnail_image.dart';
+import 'package:openvine/widgets/video_metadata/metadata_hero_corners.dart';
 import 'package:openvine/widgets/vine_cached_image.dart';
 import 'package:pro_video_editor/pro_video_editor.dart';
 import 'package:time_formatter/time_formatter.dart';
@@ -472,10 +473,11 @@ class _VideoAreaState extends State<_VideoArea> {
         ? _videoAR
         : widget.clip.targetAspectRatio.value;
     final isSquare = widget.clip.targetAspectRatio.value == 1.0;
+    final coverBorderRadius = isSquare
+        ? BorderRadius.circular(12)
+        : const BorderRadius.vertical(bottom: Radius.circular(32));
     final player = ClipRRect(
-      borderRadius: isSquare
-          ? .circular(12)
-          : const .vertical(bottom: .circular(32)),
+      borderRadius: coverBorderRadius,
       child: FittedBox(
         fit: isSquare ? BoxFit.contain : BoxFit.cover,
         child: SizedBox(
@@ -496,6 +498,14 @@ class _VideoAreaState extends State<_VideoArea> {
     return Hero(
       tag: VideoEditorConstants.heroMetaPreviewId,
       createRectTween: (begin, end) => RectTween(begin: begin, end: end),
+      // The thumbnails round themselves from outside their Hero, so on the way
+      // back the shuttle arrives with no rounding of its own.
+      flightShuttleBuilder: (_, animation, _, _, toHeroContext) =>
+          MetadataHeroCorners(
+            animation: animation,
+            destinationBorderRadius: coverBorderRadius,
+            child: (toHeroContext.widget as Hero).child,
+          ),
       child: AspectRatio(
         aspectRatio: widget.clip.targetAspectRatio.value,
         child: isSquare ? Center(child: player) : player,
--- a/mobile/lib/constants/video_editor_constants.dart
+++ b/mobile/lib/constants/video_editor_constants.dart
@@ -247,9 +247,11 @@ class VideoEditorConstants {
   /// Hero animation tag for the final clip-preview in the video editor.
   static const heroMetaPreviewId = 'Video-metadata-clip-preview-video';
 
-  /// Corner radius of the clip-preview thumbnail the [heroMetaPreviewId] hero
-  /// flies from. The preview screen morphs its flight out of this shape into
-  /// the stage's rounded bottom, so both sides read the same value.
+  /// Corner radius of the clip-preview thumbnails that carry
+  /// [heroMetaPreviewId]. The preview screen morphs its flight out of this
+  /// shape into the stage's rounded bottom, so the thumbnails and the flight
+  /// cannot drift apart. The cover screen is the other destination on that
+  /// tag and keeps its own rounding.
   static const clipPreviewCornerRadius = 16.0;
 
   /// Hero animation tag for the audio-chip in the video editor.
--- a/mobile/lib/constants/video_editor_constants.dart
+++ b/mobile/lib/constants/video_editor_constants.dart
@@ -247,6 +247,11 @@ class VideoEditorConstants {
   /// Hero animation tag for the final clip-preview in the video editor.
   static const heroMetaPreviewId = 'Video-metadata-clip-preview-video';
 
+  /// Corner radius of the clip-preview thumbnail the [heroMetaPreviewId] hero
+  /// flies from. The preview screen morphs its flight out of this shape into
+  /// the stage's rounded bottom, so both sides read the same value.
+  static const clipPreviewCornerRadius = 16.0;
+
   /// Hero animation tag for the audio-chip in the video editor.
   static const heroAudioChipId = 'Video-Editor-Audio-Chip';
 
--- a/mobile/lib/screens/video_metadata/video_metadata_preview_screen.dart
+++ b/mobile/lib/screens/video_metadata/video_metadata_preview_screen.dart
@@ -109,6 +109,11 @@ class _VideoMetadataPreviewScreenState
 
   @override
   Widget build(BuildContext context) {
+    final stageBorderRadius = widget.previewOnly
+        ? BorderRadius.zero
+        : const BorderRadius.vertical(
+            bottom: Radius.circular(VineTheme.shellCornerRadius),
+          );
     return AnnotatedRegion<SystemUiOverlayStyle>(
       value: VideoEditorConstants.uiOverlayStyleFor(context.vineColors),
       child: Scaffold(
@@ -119,13 +124,14 @@ class _VideoMetadataPreviewScreenState
             // Video preview area with close button
             Expanded(
               child: _PreviewStage(
-                roundBottom: !widget.previewOnly,
+                borderRadius: stageBorderRadius,
                 child: Stack(
                   fit: .expand,
                   children: [
                     _VideoPreviewContent(
                       clip: widget.clip,
                       controller: _controller,
+                      stageBorderRadius: stageBorderRadius,
                     ),
                     if (!widget.previewOnly)
                       // The overlay offsets its caption and action column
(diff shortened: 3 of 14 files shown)


mbradley (dismissed):
Took over the objective findings and pushed nine commits, one per finding. Two calls are yours.

## What I changed

**Corners on the way back.** The shuttle renders the destination Hero's child, so on a pop that is the metadata thumbnail with its `ClipRRect(16)` inside it. That clip intersects the one `flightShuttleBuilder` animates: the outer lerp wins at the bottom, but 16 beats square at the top, so the video's top corners snapped rounded on the first frame of the close and stayed there for the flight. Both thumbnails now round from outside their Hero, which leaves the shuttle's clip as the only one in play. Identical at rest, since the clip wraps the same subtree one level up, and during the flight the Hero leaves a placeholder with nothing to clip.

**The flight test.** It did fail without `flightShuttleBuilder`, so it caught a missing fix. But it sampled one mid-flight frame and asked whether any `ClipRRect` in the tree sat between the two endpoints. `lerp(a, b, t)` and `lerp(b, a, t)` both satisfy that for every t, and so does a constant. Tightening it to an end-of-flight check does not help either, because the stage's own clip satisfies that on its own. It now samples the shuttle itself twice and asserts the direction it travels (the top opens out as the bottom tightens), then pops and asserts the shuttle carries no competing 16 clip. Verified red against all three mutants: reversed lerp, animation pinned to a constant, and the thumbnail's clip moved back inside its Hero.

**Inset drift, two spots.** Removing the section's outer `Padding(horizontal: 4)` moved every child onto whatever inset it carries itself. `_ProviderAudioCredit` was bumped 12 to 16 to compensate; the validation line was not, so the error text sat 4px inboard of the card. Separately, the new "Shared as" row is a raw `ListTile`, so it fell through to Material 3's start 16 / end 24, the same default this branch pins out of `DivineSwitchTile` and `DivineCheckboxTile` one row above it.

**Cover top bar.** Dropping the vertical padding left the Row shrink-wrapped to its 48px button box under an `Align`, so the controls sit 4px below the safe-area edge instead of 20. The screen's outer `SafeArea` already removes the inset, so the nested one in `_TopBar` is a no-op and the padding was the only separation. `VideoMetadataCaptureAppBar` pairs the same horizontal-16 inset with an explicit `kToolbarHeight` and centres its Row in it; this is the other half of that pattern. Worth a look on device: the description frames this as a horizontal change, and horizontal was already 16 either way.

**`_ExpandSwitcher`.** `divine_ui` already exports `AnimatedReveal` for exactly this shape, and unlike the private copy it honours `MediaQuery.disableAnimations`, which the other reveal sites in the app all do. The provider-credit branch loses a cross-fade in the swap, since `AnimatedReveal` animates size and hard-swaps content. That branch needs a sound with an external source but no attribution event, which does not flip mid-session.

Plus three small ones: the `clamp(0.0, 1.0)` is unreachable (`fastOutSlowIn` cannot overshoot and neither Hero overrides the curve), the constant's doc comment said "both sides read the same value" when four widgets carry that tag and the cover screen keeps its own 12 / bottom-32 rounding, and a `SizedBox(height: 8)` as the first child of a `Column(spacing: 8)` reads as 8 and renders as 16.

## Two for you

**1. The sound title is published.** `_initialAttribution` falls back to the literal `'Original sound'`, and `attribution.title` is written into the audio event at `internal/video_event_publisher_audio.dart:85`. Four ARB keys already carry that exact string, so localizing it is mechanical, but it would make published event data depend on the poster's locale: a German poster publishes "Originalton" and every viewer sees that regardless of their own locale. Leave it English, or localize and accept locale-dependent published titles?

**2. The tile family is now split.** `DivineListTile`, `DivineSelectableRow` and both bottom-sheet menus still fall through to 24, and they render next to the two you fixed. On Safety settings a raw `ListTile` caret at 24 sits two rows above a checkbox now at 16; Nostr settings and General settings interleave the same way. Converge the family on 16, or is 16 the rule for control-bearing rows and 24 right for navigation rows? Happy to do the sweep as a follow-up either way.

## One thing not covered

The 16/16 change reaches eight settings screens plus `delete_account_dialog`. Removing it leaves both the divine_ui suite and the app suites green, so nothing pins it, and a `contentPadding` assertion is the shape `testing.md` explicitly rules out ("Testing static properties ... Useless test"). So the right ask is visual, not a test: could you add a before/after of one settings screen to the description? The metadata flow is the only surface with evidence right now.

## Validation

`flutter analyze lib test integration_test` clean. 202 metadata (screen, preview, cover, `test/widgets/video_metadata`), 845 divine_ui with the package analyzed from inside its own directory, 149 settings. All checks green on `08e235b8d`.

## Remaining uncertainty

I have not run this on a device. The top bar and the reveal swap are the two worth your eyes: both are visual changes I verified structurally and by test, not by looking at them.

Over to you to review the takeover and merge when you are happy with it, or push back on any of it.

hm21:
Thanks — the pop-flight finding is right and so is the fix. Taking the nine commits as-is.

One correction on the validation: the third mutant doesn't go red. `morphs the hero flight corners in both directions` builds its own `Hero` in the test, so moving the rounding outside the `Hero` there changed the fixture, not the app — restoring the pre-fix shape in `VideoMetadataCaptureClipPreview` leaves that test green. The first two mutants do hit production code, so that half of the tightening holds.

`f3f47476c` flies from the real thumbnail widget and asserts the same thing after a pop. It goes red against the mutant the other test survives.

Your two:

1. `'Original sound'` stays English — locale-dependent published titles aren't worth the consistency. Noted at the callsite.
2. 16 is the rule for control-bearing rows; navigation rows keep Material's start 16 / end 24. No sweep needed — `DivineListTile`, `DivineSelectableRow` and the sheet menus stay as they are. Noted on both tiles.

Screenshots for the 16/16 change to follow from device.



Bildschirmfoto 2026-08-07 um 22 20 17
Bildschirmfoto 2026-08-07 um 22 21 19

hm21:
Added the General Settings shot to the description — that surface had no evidence, which was the gap. Skipped the before half: the delta is one `contentPadding` line per tile (M3's 24 → 16 on the trailing edge) and fully readable from the diff, so a second image wouldn't resolve anything the first doesn't. The shot also happens to show the control-row / nav-row split from your question 2 in situ.

mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart
```diff
@@ -479,6 +481,14 @@ class _VideoAreaState extends State<_VideoArea> {
     return Hero(
       tag: VideoEditorConstants.heroMetaPreviewId,
       createRectTween: (begin, end) => RectTween(begin: begin, end: end),
+      // The thumbnails round themselves from outside their Hero, so on the way
+      // back the shuttle arrives with no rounding of its own.
+      flightShuttleBuilder: (_, animation, _, _, toHeroContext) =>
```
realmeylisdev, line 503: Context for this commit, since it lands on your branch.

`3fbed6dc` moved both thumbnails' `ClipRRect` outside their `Hero` so the preview's `flightShuttleBuilder` would be the only clip riding the shuttle. That fixes the preview pair, but this screen is the *other* destination on `heroMetaPreviewId`, and it is the one a thumbnail tap opens (`_openCoverEditor`); the preview is reached from the app bar. On a pop Flutter's default shuttle renders the **destination** Hero's child, which is now the unrounded thumbnail, so closing the cover screen flew a square-cornered video the whole way back. Before the move it carried 16 the whole way.

Measured it before touching anything, same pop, two thumbnail shapes: clip inside the Hero puts 1 radius-16 clip on the shuttle, clip outside puts 0.

Setting the builder on this screen rather than on the thumbnails is deliberate. Flutter resolves `toHero.flightShuttleBuilder ?? fromHero.flightShuttleBuilder`, so a destination's builder wins in both directions of its own pair, whereas one on a thumbnail would outrank the preview's on every pop and undo the fix this PR is for. The push direction is unchanged from main: this screen's own `ClipRRect` still sits inside its Hero and wins the intersection at the top corners.

The shared lerp moved into `MetadataHeroCorners` so the two destinations cannot drift apart. New test in `video_metadata_cover_screen_test.dart` goes red against all three mutants: builder removed, lerp reversed, animation pinned to a constant.

realmeylisdev (approved):
went back through this on a fresh checkout. both open questions from last round look answered, and the tile comment lands the rule in the right place.

pushed `c1b05ec5`: moving the thumbnails' clip outside their Hero also took the rounding off the flight back from the cover screen, which is what a thumbnail tap actually opens. that screen carries the same morph now. revert freely.

analyze clean, 204 metadata, 845 divine_ui, 13 cover, all four ratchets green on `c1b05ec5`.

rabble:
Heads up @hm21 — **this is the one PR of tonight's batch that did not merge.** The other nine are on `main`; this one is red and I've left it to you rather than force it through.

### The failure

`Tests (shard 2/4)` → `test/unit/services/upload_manager_get_by_path_test.dart`

> `[FAILED] UploadManager.getUploadByFilePath should return first match when multiple uploads have same path`

[run 31244760488](https://github.com/divinevideo/divine-mobile/actions/runs/31244760488/job/93071428920)

### Why I don't think it's your diff

Your 14 files are video-metadata screens/widgets plus two `divine_ui` components. Nothing touches `UploadManager`, upload paths, or anything under `test/unit/`. There is no import path I can find from your change to that assertion.

The likelier explanation is process-global bleed in CI's merged VGV isolate: every untagged `mobile/test` suite is bundled into one isolate, nothing is auto-restored between tests, and a suite that mutates a global strands later suites in a **seed-dependent** way. `getUploadByFilePath` returning the *first* match when several share a path is exactly the kind of assertion that flips when list ordering or a singleton's accumulated state leaks in from another suite. That's the failure mode `.claude/rules/testing.md` documents under "VGV merged isolate & process-global state".

Also worth knowing: an unrelated flake hit this same batch earlier — `popular_videos_tab_test.dart` → "reports a slow variant switch once, not once per rebuild" failed on #6849, passed on a straight re-run with no code change. Different test, same shard, same signature.

**I am not asserting this is flaky.** That call needs the branch's own diagnosis and I'd be guessing from the outside — per the repo rules a failing test on a feature branch is the branch's until proven otherwise. What I can say is that I looked for a connection to your diff and did not find one.

### Suggested next step

Re-run the failed shard first ([`gh run rerun 31244760488 --failed`](https://github.com/divinevideo/divine-mobile/actions/runs/31244760488)). If it goes green, the flake is real and worth its own issue rather than anything on this PR — the test predates your branch. If it fails again it's reproducible and worth a proper look at what the suite shares.

### Context you'll want

The branch was rebased onto current `main` during tonight's merge run (`1a881d268` → `d912b9361`). It is up to date and still approved; nothing about your changes was modified. Note also that `strict` (require-branches-up-to-date) is currently **disabled** on the `CI workflow rule` ruleset — @rabble is restoring that, so don't read a green `mergeStateStatus` as meaning the branch is current.

PR #6831, opened 2026-08-07
fix(report): attach machine-readable NIP-17 tags to the moderation DM

### LINKED ISSUE #6593: fix(moderation): moderation DM intake is half-wired and its unread badge is dead code
## Problem

The moderation DM intake pipeline is built on both ends but does not connect, and the one signal that would
reveal that is dead code.

**Producer side has no producer.** `divine-moderation-service/src/nostr/dm-reader.mjs` polls kind-1059 gift
wraps addressed to the moderation pubkey every five minutes, unwraps them, and only creates a structured report
row when the body parses as JSON:

```js
// dm-reader.mjs:134-137
const parsed = JSON.parse(content);
if (parsed && parsed.type === 'conversation_report' && parsed.sha256) {
  // → INSERT OR IGNORE INTO user_reports
```

divine-mobile's report DM is prose (`_formatReportDm` in `report_content_dialog.dart`), so it falls through to
`messageType: 'creator_reply'` — a chat message in the admin Messages UI, never a `user_reports` row. Grepping
the org, **no client emits `conversation_report` at all**; the producer side of that path does not exist.

**And the unread signal is dead.** `dm_log` has no read/unread column in either `migrations/003-dm-support.sql`
or the runtime DDL in `dm-store.mjs`, and `getConversations` never returns one. Yet both consumers read exactly
that field:

- `src/admin/dashboard.html:5455` — `const unread = convs.filter(c => c.unread || c.has_unread).length;`
- `src/admin/messages.html:670` — `const unread = conv.unread || conv.has_unread || false;`

So the badge is permanently zero. `support-trust-safety/docs/moderation/moderator-guide.md`'s daily routine
lists Content Review, the Relay Admin Reports tab and the auto-hide queue — the Messages UI appears only in
passing.

**Net effect:** a delivered moderation report DM may never be seen by anyone.

## Worth deciding

Whether the DM should become machine-parseable (`{type: 'conversation_report', ...}`) — which would activate the
already-shipped reader path and would specifically cover the user-report and DM-message cases that
`report-poller.mjs:459-465` drops as `skipped_missing_target` — or whether the DM should stay human-readable and
the unread badge simply be repaired.

## Notes

Found while investigating #6387. Low urgency for enforcement: the kind-1984 event is the report of record and is
consumed independently by three backends, carrying every field the DM does. This is about the DM channel being
half-wired, not about reports being lost.

Also, don't read the reader's logs as evidence the channel is quiet: `logDm` returns the existing row on dedup
and the caller branches on `result.id`, so `skipped` is structurally always 0 in
`[DM-READER] Sync complete: N new, 0 deduped`.

realmeylisdev:
Closes #6593

## Summary

Fixes the producer half of #6593 ("moderation DM intake is half-wired and its unread badge is dead code") — divine-moderation-service's `dm-reader.mjs` only ever promoted an incoming DM to a `user_reports` row when its content parsed as JSON `{type:'conversation_report', sha256}`, a shape this app has never produced.

The DM's content stays exactly the same human-readable prose (`_formatReportDm`, unchanged) — no moderator-facing UI change, no NIP-17 content-convention violation. Report data now also travels as NIP-17 tags on the same kind-14 rumor, via `dm_repository.sendMessage`'s existing (already-shipped, already-tested via `replyToId`) `additionalTags` parameter. The companion divine-moderation-service PR reads those tags instead of trying to `JSON.parse` the content.

- `report_type` is attached for every report variant (content, user, DM-message), mapped from the same `ContentFilterReason` → NIP-56 conversion already used for the kind-1984 report tags — lifted out of `ContentReportingService` into a shared `contentFilterReasonToNip56Type` so both channels can't drift out of sync.
- `sha256` is attached only for content reports where `VideoEvent.sha256` resolves to a non-empty value. User reports and DM-message reports have no Blossom blob hash and never carry it — this matches the backend's `user_reports.sha256 NOT NULL` schema, which has no user-only/message-only shape (see the linked plan's "Non-goals": this activates the backend's `user_reports` insert for content reports only, by design, not as a gap).

The retry path (`recoverFullSend`) replays the originally-queued rumor bytes rather than rebuilding the send, so tags attached on the first send are automatically preserved on retry with no separate plumbing needed.

## Test plan

- [x] `flutter test test/widgets/report_content_dialog_test.dart` — 84/84 pass (3 new: `report_type` tag present for the selected reason, `sha256` tag present/absent based on `VideoEvent.sha256`)
- [x] `flutter test test/services/content_moderation_types_test.dart` — 7/7 pass (new coverage for the extracted `contentFilterReasonToNip56Type` mapping, including exhaustiveness over every `ContentFilterReason`)
- [x] `flutter test test/services/content_reporting_service_test.dart` — 27/27 pass (unaffected — same mapping, moved not changed)
- [x] Pre-push hook (format + analyze + affected tests) — all green

## Related

- #6593
- Companion PR in divine-moderation-service (reads these tags)

### DIFF
--- a/mobile/lib/widgets/report_content_dialog.dart
+++ b/mobile/lib/widgets/report_content_dialog.dart
@@ -524,6 +524,7 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
           // never let them degrade to a metadata-leaking NIP-04
           // plaintext duplicate. NIP-17 gift wrap only.
           skipNip04Fallback: true,
+          additionalTags: _reportDmTags(),
         );
       } else {
         try {
@@ -617,6 +618,26 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
       buffer.writeln('Details: $details');
     }
     return buffer.toString().trimRight();
+  }
+
+  /// Machine-readable data for divine-moderation-service's dm-reader to
+  /// classify this report, attached as NIP-17 tags on the rumor rather
+  /// than folded into [_formatReportDm]'s content — the DM stays plain,
+  /// human-readable prose (see divine-mobile#6593).
+  ///
+  /// `sha256` is only present for content reports where the reported
+  /// video has a resolvable Blossom blob hash — [VideoEvent.sha256] is
+  /// nullable (a video published without an `imeta` `x` value has none),
+  /// and user reports / DM-message reports never have one at all. The
+  /// backend's `user_reports` table is `sha256 NOT NULL`, so it can only
+  /// ever record content reports; this tag intentionally never covers the
+  /// other two report variants.
+  List<List<String>> _reportDmTags() {
+    final sha256 = widget.video?.sha256;
+    return [
+      ['report_type', contentFilterReasonToNip56Type(_selectedReason!)],
+      if (sha256 != null && sha256.isNotEmpty) ['sha256', sha256],
+    ];
   }
 
   @override
--- a/mobile/test/services/content_moderation_types_test.dart
+++ b/mobile/test/services/content_moderation_types_test.dart
@@ -72,4 +72,41 @@ void main() {
       expect(ModerationResult.clean.warningMessage, isNull);
     });
   });
+
+  group(contentFilterReasonToNip56Type, () {
+    test('maps every reason to a canonical NIP-56 report type string', () {
+      const expected = {
+        ContentFilterReason.spam: 'spam',
+        ContentFilterReason.harassment: 'profanity',
+        ContentFilterReason.violence: 'illegal',
+        ContentFilterReason.sexualContent: 'nudity',
+        ContentFilterReason.copyright: 'illegal',
+        ContentFilterReason.falseInformation: 'other',
+        ContentFilterReason.childSafety: 'other',
+        ContentFilterReason.csam: 'illegal',
+        ContentFilterReason.underageUser: 'other',
+        ContentFilterReason.aiGenerated: 'other',
+        ContentFilterReason.other: 'other',
+      };
+
+      for (final entry in expected.entries) {
+        expect(
+          contentFilterReasonToNip56Type(entry.key),
+          entry.value,
+          reason: '${entry.key} should map to NIP-56 type "${entry.value}"',
+        );
+      }
+    });
+
+    test('covers every ContentFilterReason value', () {
+      // Guards against a new enum case silently falling through to a
+      // default -- the switch in contentFilterReasonToNip56Type is
+      // exhaustive (no default case), so this would be a compile error
+      // before it could ever be a runtime gap, but pinning the full
+      // enum set here keeps that guarantee visible in the test itself.
+      for (final reason in ContentFilterReason.values) {
+        expect(contentFilterReasonToNip56Type(reason), isNotEmpty);
+      }
+    });
+  });
 }
--- a/mobile/lib/services/content_moderation_types.dart
+++ b/mobile/lib/services/content_moderation_types.dart
@@ -18,6 +18,30 @@ enum ContentFilterReason {
   other,
 }
 
+/// Maps a [ContentFilterReason] to the canonical NIP-56 (kind 1984)
+/// report type strings: nudity, malware, profanity, illegal, spam,
+/// impersonation, other.
+///
+/// Shared by [ContentFilterReason]'s two report channels so they stay in
+/// sync: the kind-1984 `e`/`p` tag's 3rd element (NIP-56), and the
+/// moderation-DM's `report_type` tag (a divine-moderation-service-specific
+/// convention — see `divine-mobile#6593`).
+String contentFilterReasonToNip56Type(ContentFilterReason reason) {
+  return switch (reason) {
+    ContentFilterReason.spam => 'spam',
+    ContentFilterReason.harassment => 'profanity',
+    ContentFilterReason.violence => 'illegal',
+    ContentFilterReason.sexualContent => 'nudity',
+    ContentFilterReason.copyright => 'illegal',
+    ContentFilterReason.falseInformation => 'other',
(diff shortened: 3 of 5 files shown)


realmeylisdev:
Companion PR (backend consumer): https://github.com/divinevideo/divine-moderation-service/pull/193

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -619,6 +620,26 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
...
+  /// video has a resolvable Blossom blob hash — [VideoEvent.sha256] is
+  /// nullable (a video published without an `imeta` `x` value has none),
+  /// and user reports / DM-message reports never have one at all. The
+  /// backend's `user_reports` table is `sha256 NOT NULL`, so it can only
+  /// ever record content reports; this tag intentionally never covers the
+  /// other two report variants.
+  List<List<String>> _reportDmTags() {
+    final sha256 = widget.video?.sha256;
+    return [
+      ['report_type', contentFilterReasonToNip56Type(_selectedReason!)],
```
NotThatKindOfDrLiz, line 638: This should not be the only machine-readable report type the DM path sends. `contentFilterReasonToNip56Type` intentionally collapses several app reasons (`aiGenerated`, `childSafety`, `underageUser` -> `other`; `csam`, `copyright` -> `illegal`) for the NIP-56 marker, but the moderation DM tag is a Divine-private ingestion contract. The companion DM reader writes this tag into `user_reports.report_type` with `INSERT OR IGNORE`, so if it races ahead of the kind-1984 poller, the coarse value wins permanently for `(sha256, reporter_pubkey)`. Please add a granular tag mirroring the existing NIP-32 label / reason name and update the backend reader to prefer that value, falling back to this NIP-56 tag only if needed.

mobile/test/services/content_moderation_types_test.dart
```diff
@@ -72,4 +72,41 @@ void main() {
...
+        ContentFilterReason.copyright: 'illegal',
+        ContentFilterReason.falseInformation: 'other',
+        ContentFilterReason.childSafety: 'other',
+        ContentFilterReason.csam: 'illegal',
+        ContentFilterReason.underageUser: 'other',
+        ContentFilterReason.aiGenerated: 'other',
+        ContentFilterReason.other: 'other',
+      };
+
+      for (final entry in expected.entries) {
```
NotThatKindOfDrLiz, line 92: This test only covers the entries already present in `expected`; a new `ContentFilterReason` with a wrong mapping arm would not fail here, and the next test only asserts non-empty output. Please assert `expected.keys.toSet() == ContentFilterReason.values.toSet()` before the loop so the table is actually pinned to the enum. After that, the separate non-empty coverage test can go away.

NotThatKindOfDrLiz (changes requested):
Requesting changes for the cross-service data contract issue. The PR is directionally right, but the DM path should not write a lossy NIP-56 value into `user_reports.report_type` when the existing kind-1984 ingestion path preserves the granular Divine reason. Please add a granular machine-readable tag and have the backend prefer it before this merges. The enum coverage test should also be tightened while touching this code.

mobile/lib/widgets/report_content_dialog.dart
realmeylisdev: Fixed in #6855 (this PR is superseded and now closed).

You were right, and verifying it turned up something worse than the version in your comment. The rumor now carries the same `['L', ns]` + `['l', 'NS-<reason>', ns]` pair the kind-1984 report already publishes, rather than a new tag — that way the backend's existing `extractReportType` decodes it with no new vocabulary, and the two channels resolve to one value by construction. The NIP-56 tag stays as the fallback you asked for.

Two things worth flagging from chasing it down:

1. `extractReportType`'s chain is `l` → `Reason:` content regex → `e`/`p` markers. A NIP-17 rumor's `p` tag has no third element, so the coarse `report_type` tag was not in that chain at all — "prefer granular, fall back to NIP-56" would have been silently inert. The backend PR inserts the `report_type` arm above the content regex, which also matters because a DM's content is localized prose (`Reason: Contenu indésirable`), so the regex would otherwise have made `report_type` locale-dependent.

2. Preferring the granular label would have *regressed* NSFW escalation on its own. `normalizeReportType` only decodes the `aiGenerated` family, so `NS-sexualContent` normalizes to `ns_sexualcontent`, which is not in `NSFW_REPORT_TYPES` — auto age-restriction has never fired from the relay path. The coarse `nudity` value this PR sent was, accidentally, the only input that made it work. divinevideo/divine-moderation-service#195 has the measured table; the fix ships in divinevideo/divine-moderation-service#194 alongside the precedence change.

mobile/test/services/content_moderation_types_test.dart
realmeylisdev: Fixed in #6855 (this PR is superseded and now closed). Both mapping tables now assert `expected.keys.toSet() == ContentFilterReason.values.toSet()` before the loop, and the separate non-empty test is gone.

The NIP-32 label table added in #6855 got the same treatment, plus one test asserting the label stays lossless exactly where the NIP-56 mapping collapses — five reasons share two NIP-56 values but map to five distinct labels. That is the property the whole change depends on, so it is now pinned rather than implied.

realmeylisdev:
Superseded by #6855, which addresses both review threads from @NotThatKindOfDrLiz and adds the granular NIP-32 label they asked for.

Opened as a replacement rather than a force-push so the review history here stays readable — the tag design changed materially (NIP-32 `L`/`l` pair preferred, NIP-56 kept as fallback), and the fix now spans both repos.

- divine-mobile: #6855
- divine-moderation-service: divinevideo/divine-moderation-service#194 (replaces divinevideo/divine-moderation-service#193)
- pre-existing bug found on the way: divinevideo/divine-moderation-service#195

PR #6836, opened 2026-08-07
fix(library): move clip selection into a sheet footer

### LINKED ISSUE #6837: fix(library): move clip selection into a sheet footer
Auto-created from PR #6836.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6836
**Branch:** `fix/clip-selection-sheet-footer`

hm21:
Closes #6837

## Description

The clip picker that opens from the video editor is a `VineBottomSheet`, but its content re-stated chrome the sheet already provides:

- A second **"Clips"** title sat right under the sheet's own header.
- The library shell painted `surfaceContainerHigh` (`0xFF000A06`, near black) on top of the sheet's `surface` (`0xFF00150D`, dark green), so the sheet looked like a black panel inside a green frame.
- The confirm action was a custom green pill in the top-right corner. Without a selection it was fully inert — `enable: false` nulled the tap handler, which also made the `context.pop` fallback behind it unreachable dead code.

This reworks the selection mode:

- `ClipSelectionHeader` → **`ClipSelectionFooter`**: no title, no divider, just a full-width `DivineButton` at the bottom with `SafeArea(top: false)`. That mirrors `_CreateVideoBar`, the footer the same screen already uses for the non-selection "Create video" flow, so the two confirm affordances now look and sit the same. Disabled state comes from the design system via `onPressed: null` rather than an `Opacity(0.32)` wrapper.
- `_LibraryContent` skips the shell wrapper (`ClipRRect` + `ColoredBox`) entirely in selection mode. The sheet already supplies surface and rounded corners; stacking a second one was the source of the black panel.
- The grid gets 8px of top padding so the first row clears the sheet header's divider. (Not the drag handle — `VineBottomSheetHeader` is a sibling above the scroll body and already puts 20px plus that divider between the handle and the content.) That padding is also **why** the `ClipRRect` had to go rather than just turning transparent: `shellInnerCornerRadius` is 32px, so once the first row no longer starts at y=0 it lands mid-arc and the corner slices diagonally through the top thumbnails. Previously the arc happened to read as the tiles' own rounding.
- `gridTopPadding` is an explicit `ClipsTab` parameter defaulting to `0`, so the standalone library screen and clips-only mode are untouched.

`BlocSelector` in the footer now selects `selectedClipIds.isNotEmpty` instead of the whole `Set`, so it rebuilds on the empty↔non-empty edge rather than on every selection change.

**Related Issue:** 

## Out of Scope

- The ARB key `libraryClipSelectionTitle` is now unused in code but still defined in all 17 locales. Left in place — this repo tolerates defined-but-unused keys (see `.claude/rules/localization.md`), and removing it would churn 17 ARB files plus generated output for no behavioural gain.
- The grid keeps its `bottom: viewPadding.bottom` padding, which is now partly redundant under the footer's own `SafeArea`. It costs a few px of extra scroll overshoot in selection mode and is load-bearing for every other mode.

## Verification

- `flutter test test/widgets/library test/screens/library_screen_test.dart` — 88 passing
- Manually verified on a real Samsung Galaxy: opened the clip picker from the video editor, confirmed the sheet's green surface now runs edge to edge, the select button sits in the footer and is disabled until a clip is picked, and the top thumbnail row is no longer clipped.

Tests updated: `ClipSelectionHeader` → `ClipSelectionFooter` throughout; the `renders Clips title` case is gone with the title, replaced by one that pins the button as a no-op without a selection.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/screens/library_screen.dart
+++ b/mobile/lib/screens/library_screen.dart
@@ -619,59 +619,67 @@ class _LibraryContent extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final content = Column(
+      crossAxisAlignment: CrossAxisAlignment.stretch,
+      children: [
+        if (!isClipsOnlyMode) const SizedBox(height: 12),
+        if (!isClipsOnlyMode)
+          TabBar(
+            controller: tabController,
+            isScrollable: true,
+            tabAlignment: TabAlignment.start,
+            padding: const EdgeInsetsDirectional.only(start: 16),
+            indicatorColor: VineTheme.tabIndicatorGreen,
+            indicatorWeight: 4,
+            indicatorSize: TabBarIndicatorSize.tab,
+            dividerColor: VineTheme.transparent,
+            labelColor: context.vineColors.primaryText,
+            unselectedLabelColor: context.vineColors.onSurfaceMuted,
+            labelPadding: const EdgeInsets.symmetric(horizontal: 14),
+            labelStyle: VineTheme.titleMediumFont(
+              color: context.vineColors.primaryText,
+            ),
+            unselectedLabelStyle: VineTheme.titleMediumFont(
+              color: context.vineColors.onSurfaceMuted,
+            ),
+            tabs: [
+              Tab(text: context.l10n.libraryTabDrafts),
+              Tab(text: context.l10n.libraryTabClips),
+              Tab(text: context.l10n.soundsTitle),
+            ],
+          ),
+        if (!isClipsOnlyMode) const SizedBox(height: 2),
+        Expanded(
+          child: selectionMode
+              ? _SelectionBody(
+                  scrollController: scrollController,
+                  targetAspectRatio: targetAspectRatio,
+                  onCreate: onCreateVideo,
+                )
+              : _TabBody(
+                  clips: sortedClips,
+                  selectionEnabled: selectionEnabled,
+                  isClipsOnlyMode: isClipsOnlyMode,
+                  tabController: tabController,
+                  targetAspectRatio: targetAspectRatio,
+                ),
+        ),
+      ],
+    );
+
+    // Selection mode renders inside a bottom sheet, which brings its own
+    // surface and rounded corners. Skipping the shell here keeps the sheet's
+    // colour showing through and stops the 32px corner radius from slicing
+    // the top row of thumbnails.
+    if (selectionMode) return content;
+
     return ClipRRect(
       borderRadius: const BorderRadius.all(
         Radius.circular(VineTheme.shellInnerCornerRadius),
       ),
       child: ColoredBox(
         color: context.vineColors.surfaceContainerHigh,
-        child: Column(
-          crossAxisAlignment: CrossAxisAlignment.stretch,
-          children: [
-            if (!isClipsOnlyMode) const SizedBox(height: 12),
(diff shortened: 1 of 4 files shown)


rabble (approved):
## Summary
Approve — focused, correct sheet fix that removes stacked chrome without behavioural churn. CI is fully green (Analyze, Format, Generated Files, 4/4 test shards).

## What this gets right
- **Removes double header + shell in selection mode:** `ClipSelectionHeader` (duplicate “Clips” title + divider + custom pill) → `ClipSelectionFooter`. Sheet header is the single source for drag handle + divider; success path mirrors `_CreateVideoBar` (full-width `DivineButton` in `SafeArea(top: false)`).
- **Surface fix:** `_LibraryContent` early-returns raw `Column` when `selectionMode`, skipping `ClipRRect(shellInnerCornerRadius=32)` + `surfaceContainerHigh(0xFF000A06)` over sheet `surface(0xFF00150D)`. Eliminates black-panel-inside-green-frame. `isClipsOnlyMode = selectionMode || tabsMode==clipsOnly` already hides `TabBar` in the sheet, so no stray tabs leak in.
- **Dead code gone:** old `_AddClipButton(enable)` used `Opacity(0.32)` + nulled `onTap`, making `context.pop` fallback unreachable. Replaced by `DivineButton(onPressed: hasSelection ? onCreate : null)` — disabled state now from the design system.
- **Rebuild optimization:** `BlocSelector<…, bool>(selectedClipIds.isNotEmpty)` instead of `Set<String>` — rebuilds only on empty↔non-empty edge, which is exactly the button’s enable edge.
- **Padding isolation:** `gridTopPadding = 0 by default`, `8` only via `_SelectionBody`. Non-selection modes untouched; standalone / clipsOnly unaffected. Correctly noted that the 8px clears the sheet header divider (header is a sibling above scroll body with 20px + divider) and avoids the 32px corner arc slicing the first thumbnail row.
- **Cleanup:** removes unused `go_router` import; deletes 36 lines of custom button.

## Tests & verification
- `flutter test test/widgets/library/clips_tab_test.dart` — **9 passed** including both new `ClipSelectionFooter` cases (`calls onCreate when tapped`, `disables without selection`).
- `flutter test test/screens/library_screen_test.dart` — **25 passed, 1 skipped** (`ClipSelectionFooter in selection mode`, selection flow through `LibraryScreen`).
- Previous `renders Clips title` case correctly retired with the title.
- Author’s manual verification on Samsung Galaxy noted; preview + CI match.

## Minor notes (non-blocking)
- Double bottom padding (`MasonryGridView bottom: viewPadding.bottom` + footer `SafeArea`) acknowledged Out of Scope — small overshoot, load-bearing for other modes. No action needed.
- `libraryClipSelectionTitle` ARB key now unused across 17 locales; leaving it per `.claude/rules/localization.md` is correct — churn would outweigh gain.
- No functional change to `VideoEditorScreen` sheet invocation (`VineBottomSheet.show(buildScrollBody: (c) => LibraryScreen(selectionMode: true, scrollController: c))`) — scroll controller still wired to grid only, footer stays pinned.

## Risk
Low — 4 files, +115/-158, no API surface change outside the renamed widget (internal). Selection-type filtering and disable logic unchanged.

LGTM to merge.

Issue #6763, opened 2026-08-06
fix(notifications): preserve mid-sentence mention previews


NotThatKindOfDrLiz:
## Problem

Current main prevents raw npub leakage in notification comment previews, but when a bech32 Nostr reference starts before the 50-character preview limit and ends after it, `_bech32AwareCut` omits the token unless the content before it is only whitespace or punctuation.

That means a mid-sentence mention can disappear from the preview entirely, for example:

`Reply to nostr:npub1... more` -> `Reply to...`

This is better than rendering a raw npub, but still loses useful user-visible context.

## Expected direction

Resolve/linkify profile references to their display labels before applying the preview cap, or otherwise build notification preview text from display-ready spans instead of truncating raw Nostr content.

## Acceptance criteria

- Mid-sentence profile mentions remain visible in bounded notification comment previews when the resolved display label fits.
- Raw bech32 tokens are still never split or rendered as invalid npub fragments.
- Existing bounds against unbounded remote-controlled notification content remain in place.
- Add focused regression tests for mid-sentence mention preservation and the existing no-raw-npub cases.

## Notes

Do not block the Android release for #6762 on this follow-up. The original Galaxy S22 Ultra report on `1.0.17+668` is the already-fixed #6346 regression being shipped separately.

realmeylisdev:
While verifying #6762 I found this defect is **locale-dependent** on current `main`, which I think changes the framing here.

`_hasOnlyWhitespaceOrPunctuationBefore` only rejects ASCII `[0-9A-Za-z]`, so Cyrillic, CJK, and accented lead-in text counts as punctuation. A straddling token therefore *does* qualify as "leading" in those locales, and `_bech32AwareCut` returns the token's end instead of its start — the mention survives. Captured from a probe against `main`:

```
cyrillic (100 chars) => "Привет всем друзьям сегодня nostr:npub1…uulstw6..."
latin     (32 chars) => "Hello everyone my friends tod..."
```

So the mention loss described here reproduces for Latin-script content but not for most non-Latin content. That is accidental, and it comes at a cost: the 50-char preview cap silently becomes `limit * 4` = 200 for those locales, which is four times the intended bound on remote-controlled notification content — the bound #6345 deliberately added.

[...]

Issue #6762, opened 2026-08-06
chore(release): ship Android notification npub fix


NotThatKindOfDrLiz:
> **Retargeted twice.** The original framing assumed the reporter's build predates `b652e472c` and that cutting a new Android release would fix it. The first retarget (2026-08-07) refuted the "cut a release" half but misidentified the Play blocker as an uncompleted declaration. [Play Console state checked directly on 2026-08-08](https://github.com/divinevideo/divine-mobile/issues/6762#issuecomment-5223972005) settled it: the declaration was already done, and the real blocker is a stalled staged rollout. Original body preserved at the bottom.

## Problem

An Android user on Galaxy S22 Ultra reports that in-app notifications show a raw `npub` when they are tagged in comments, while the comment itself shows the resolved tag/name. That is the symptom of #6346, fixed by `b652e472c` / #6345 (merged 2026-07-25 05:30 UTC).

The reporter installed from Play. The fix is already in a Play artifact — it has just never reached them:

| Play production release | Date | Contains `b652e472c` | Reach |
|---|---|---|---|
| 672 (1.0.17) | Jul 25 | **Yes** | **20% of users** |
| 668 (1.0.17) | Jul 22 | No | everyone else |

The staged rollout of 672 stopped at 20% and was never advanced, so roughly 80% of Play users — including the reporter — are still on a build that predates the fix.

The GitHub release APK (what Zapstore consumes) has shipped the fix since 2026-07-26. That channel is fine.

**No new build is needed.** Advancing the existing rollout ships it.

## Blocked by

- #6860 — advance the Play production rollout of 672 from 20% to 100%. Needs someone with release manager access.

## Acceptance criteria

- [ ] #6860 is done: release 672 is at 100% on the Play production track.
- [ ] Confirm the reporter is offered the update.
- [ ] Confirm the reporter can no longer reproduce raw `npub` text in in-app notification comment previews for tagged comments.

## Worth knowing: the release tag does not identify the code

`codemagic.yaml` derives the GitHub release tag from `pubspec.yaml`'s `versionName` **at build time**, so a build run days after a tag was published attaches its APKs to that existing tag. The `1.0.17` tag points at `9e56db655` (2026-07-22), but the APKs hanging off it were built 2026-07-26 and contain commits that are not in the tagged commit — including `b652e472c`.

[...]

PR #6345, opened 2026-07-24
fix(notifications): keep bech32 npub tokens intact in reply/comment previews

rabble:
Closes #6346

## What & why

Notification reply/comment quotes were showing raw, sliced `npub1...` strings instead of resolved names. `_truncateComment` capped previews at 50 chars with a blind substring, so embedded bech32 Nostr references could be cut mid-token before the UI linkifier had a chance to decode them.

## Fix

Make `_truncateComment` bech32-token-aware without letting remote notification content bypass the preview cap:

- align the repository token boundary with `LinkifiedTextSpanBuilder` so only UI-decodable token spans are protected;
- when a token straddles the 50-char cut, omit the token instead of slicing it;
- when the preview is only whitespace/punctuation plus a bounded Nostr reference, keep the whole token so the UI can resolve it;
- reject unbounded token-shaped content by dropping that token rather than persisting thousands of remote-controlled characters;
- trim trailing whitespace before adding the ellipsis.

Current `main` also includes the dedicated `notification_repository` CI workflow, so the package tests now run in CI after this rebase.

## Verification

- `mise exec -- dart format --output=none --set-exit-if-changed packages/notification_repository/lib/src/notification_repository.dart packages/notification_repository/test/src/notification_repository_test.dart`
- `mise exec -- flutter analyze packages/notification_repository`
- `mise exec -- flutter test test/src/notification_repository_test.dart` from `mobile/packages/notification_repository` - 131 pass
- pre-push hook: no merge conflicts with main, analyzer clean, changed-file notification_repository tests pass

## Regression tests added

1. Reply body is just a bounded `npub1...` token and stays decodable.
2. Leading whitespace/punctuation before `npub1...` still keeps the token intact.
3. `Reply to <npub> more` omits the straddling token and avoids a sliced `npub1` fragment.
4. Oversized leading token-shaped content is not persisted unbounded.
5. Alphanumeric-embedded `prefixnpub1...` text is not treated as a linkifier token.

PR #6839, opened 2026-08-07
fix(notifications): protect every UI-decodable reference in previews

realmeylisdev:
## Description

#6345 made the notification comment preview stop slicing bech32 tokens. While verifying that fix for #6762 I found the preview cut is still wrong in two other ways, and review found a third. All three are the same underlying mistake — the repository's idea of "a reference the UI can decode" does not match what `LinkifiedTextSpanBuilder` actually decodes.

**1. A bare 64-char hex reference was never protected.**

`LinkifiedTextSpanBuilder._combinedRegex` decodes a bare 64-char hex pubkey / event id and renders it as a tappable "View video" link. The repository's `_bech32ReferencePattern` covered only the bech32 alternative. A hex reference is 64 characters, so it *always* exceeded the 50-char cap and was *always* sliced — and anything shorter than 64 hex chars is not a reference, so the row rendered 50 raw hex characters instead of a link. Same defect class as #6346, different token type.

Before / after, from the repository tests:

```
before:  "a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f6071829..."
after:   "a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8f90..."
```

**2. The preview cap did not apply to non-Latin scripts.**

`_hasOnlyWhitespaceOrPunctuationBefore` rejected only ASCII `[0-9A-Za-z]`, so Cyrillic, CJK, and accented lead-in text counted as punctuation. A straddling token then qualified as "leading", and the cut returned the token's *end* — bounded only by `limit * 4`. The 50-char cap silently became 200 for most locales, persisting four times the intended amount of remote-controlled notification content:

```
Cyrillic lead-in (before):  100 chars — "Привет всем друзьям сегодня nostr:npub1…uulstw6..."
Latin lead-in    (before):   32 chars — "Hello everyone my friends tod..."
Cyrillic lead-in (after):    30 chars — "Привет всем друзьям сегодня..."
```

Non-ASCII *punctuation* (guillemets, em dash) still correctly counts as a leading-token prefix; there is a test for that direction too.

**3. A hex or bech32 run inside a URL or hashtag was protected as if the UI decoded it.**

Found in review by @hm21 and fixed in a11bde3a4 — a regression introduced by fix 1 above. Adding the hex alternative gave the repository a pattern that could match a hex run *anywhere*, including inside a span the UI never links. `LinkifiedTextSpanBuilder` resolves alternation per start position, so a URL or hashtag containing a hex run starts earlier and wins: `https://host/<64-hex>` is a single URL token and the hex inside it is never a reference. The cut was pulled back to it anyway, shortening the preview for a span the UI does not resolve. Blossom and CDN media URLs are exactly `https://host/<sha256>`, so this is the common shape rather than an edge case:

```
before this PR:  "Look at this https://blossom.divine.video/a1b2c3d4..."  (53)
on b09b6e7ac:    "Look at this https://blossom.divine.video/..."          (45)
after a11bde3a4: "Look at this https://blossom.divine.video/a1b2c3d4..."  (53)
```

`#<64-hex>` was worse, collapsing the preview to `nice one #...` (13 chars of a 50-char budget).

The fix mirrors `_combinedRegex` alternative for alternative and protects only the bech32 and hex capture groups, so the surrounding alternatives consume their own text and a run buried inside them is not mistaken for a reference. Two consequences worth naming:

- The UI's trailing `@([a-zA-Z][a-zA-Z0-9_]{0,30})` alternative is deliberately **not** mirrored. It beats bech32 for `@npub1…` and truncates the token to 31 characters, so mirroring it would drop the protection #6346 added for that input. Whether the UI should tokenize `@npub1…` as a mention at all is a separate question from this cut, and is not decided here.
- A bech32 token inside a URL path is no longer pulled back either, because the UI reads that as a URL too. This is a deliberate behaviour change relative to `main` and is pinned by a test. It is not separable from the fix: bech32-in-URL and hex-in-URL are the same alternative, so keeping the old behaviour there would reintroduce the hex-in-URL regression that is the common case.

The fix replaces the ASCII loop with `RegExp(r'[\p{L}\p{N}]', unicode: true)`, adds the hex alternative to the reference pattern, and mirrors the surrounding UI alternatives so the "exactly what the UI can decode" rule holds, renaming to drop the now-inaccurate `bech32` prefix.

**Related Issue:** Relates to #6762, #6346, #6763

## Out of Scope

- The mid-sentence mention loss tracked in #6763. Note for that issue: this change makes the behaviour *consistent* across locales — previously Cyrillic lead-ins accidentally kept the mention while Latin ones dropped it. #6763 remains the fix for preserving it everywhere.
- Sharing one token regex between `notification_repository` and `LinkifiedTextSpanBuilder`. They are kept in sync by comment and by the tests here; a shared source would mean extracting a new package, since a pure-Dart package cannot depend on the app layer.
- Whether the UI should tokenize `@npub1…` as a 31-char mention instead of a bech32 reference. Pre-existing, unchanged here.

## Verification

- `flutter test` in `mobile/packages/notification_repository` — **172 pass**
- `flutter test --coverage` in that package — **95.68%**, above its `min_coverage: 94` gate
- `flutter test test/notifications/` — **210 pass**
- `flutter analyze` in the package — no issues
- `flutter analyze lib test integration_test` in `mobile/` — no issues
- `dart format --set-exit-if-changed` — clean

Red/green status per test, re-checked against `origin/main` and against `b09b6e7ac` rather than asserted:

- Tests 1, 2, 3 fail against `origin/main` and pass after — the "before" numbers above are captured output from that run, not estimates.
- Tests 8, 9, 10 fail against `b09b6e7ac` and pass after; the numbers in defect 3 are captured from that run.
- Tests 4, 5, 7 are **guard tests that also pass against pre-fix code**. They pin behaviour that must not regress (the other side of the non-Latin fix, and the render-layer failure modes that give the repository cut its reason to exist) rather than demonstrating a fixed defect. An earlier revision of this description claimed every test was red first; that was wrong for these three.

### Tests added

`notification_repository`:
1. Keeps a bounded leading 64-hex reference intact.
2. Omits a straddling 64-hex reference rather than slicing it.
3. Applies the 50-char cap to non-Latin lead-in text.
4. Still treats punctuation-only lead-in (including non-ASCII punctuation) as leading.
8. Leaves a hex run inside a URL to the plain cut.
9. Leaves a bech32 token inside a URL to the plain cut.
10. Leaves a hex run behind a hashtag to the plain cut.

`notification_comment_quote`:
5. An undecodable npub renders as raw text — pins the #6346 failure mode at the render layer, which previously had no guard.
6. An intact 64-hex reference renders as a tappable link.
7. A partial hex reference renders as raw text.

Test 6 also required adding `AppLocalizations.localizationsDelegates` / `supportedLocales` to this file's `_pump` helper, which was missing them. Without the delegates `videoLabel` resolved to `null` and the widget silently fell back to rendering the raw reference — so the harness would have reported a broken link as passing.

## Type of Change

### DIFF
--- a/mobile/packages/notification_repository/lib/src/notification_repository.dart
+++ b/mobile/packages/notification_repository/lib/src/notification_repository.dart
@@ -46,16 +46,30 @@ final RegExp _npubIdentifierPattern = RegExp(
   caseSensitive: false,
 );
 
-/// A bech32 Nostr reference embedded in note content, e.g. `npub1...` or
-/// `nostr:nprofile1...`.
+/// A Nostr reference embedded in note content that the UI can decode: either
+/// bech32 (`npub1...`, `nostr:nprofile1...`) or a bare 64-char hex pubkey /
+/// event id.
 ///
-/// Keep this boundary rule aligned with
-/// `LinkifiedTextSpanBuilder._combinedRegex` so the repository only protects
-/// token spans the UI can decode.
-final RegExp _bech32ReferencePattern = RegExp(
-  r'(?<![A-Za-z0-9])(?:nostr:)?(?:npub|nprofile|note|nevent|naddr)1[a-z0-9]+\b',
+/// Keep both boundary rules aligned with
+/// `LinkifiedTextSpanBuilder._combinedRegex` so the repository protects
+/// exactly the token spans the UI can decode — no more, no less. A hex
+/// reference is 64 characters, so it always exceeds the preview cap; leaving
+/// it unprotected guaranteed a sliced fragment that the UI then rendered as
+/// raw hex, which is the same defect bech32 slicing caused.
+final RegExp _decodableReferencePattern = RegExp(
+  r'(?<![A-Za-z0-9])(?:nostr:)?(?:npub|nprofile|note|nevent|naddr)1[a-z0-9]+\b'
+  '|(?<![A-Fa-f0-9])[A-Fa-f0-9]{64}(?![A-Fa-f0-9])',
   caseSensitive: false,
 );
+
+/// Any letter or digit in any script.
+///
+/// The preview's "is this token the leading content?" test must treat
+/// Cyrillic, CJK, and accented text as content. Testing only ASCII
+/// `[0-9A-Za-z]` classified every non-Latin script as punctuation, which let
+/// a straddling token qualify as leading and silently raised the preview cap
+/// from [_maxCommentLength] to four times that for most locales.
+final RegExp _letterOrDigitPattern = RegExp(r'[\p{L}\p{N}]', unicode: true);
 
 /// Retry policy for the first-page notifications fetch.
 ///
@@ -2335,12 +2349,12 @@ class NotificationRepository {
   ///
   /// Only applies to comment and reply notifications.
   ///
-  /// The cut avoids splitting a bech32 Nostr reference (`npub1...`,
-  /// `nprofile1...`, `note1...`, `nevent1...`, `naddr1...`) mid-token. A
-  /// sliced token can no longer be decoded, so the row widget falls back to
-  /// rendering the raw `npub1...` string verbatim instead of resolving it.
-  /// Keeping a bounded leading token intact lets the UI linkifier resolve it to
-  /// `@<name>`.
+  /// The cut avoids splitting a Nostr reference the UI can decode — bech32
+  /// (`npub1...`, `nprofile1...`, `note1...`, `nevent1...`, `naddr1...`) or a
+  /// bare 64-char hex pubkey / event id — mid-token. A sliced token can no
+  /// longer be decoded, so the row widget falls back to rendering the raw
+  /// string verbatim instead of resolving it. Keeping a bounded leading token
+  /// intact lets the UI linkifier resolve it to `@<name>`.
   static String? _truncateComment(String? content, NotificationKind kind) {
     if (content == null) return null;
     if (kind != NotificationKind.comment && kind != NotificationKind.reply) {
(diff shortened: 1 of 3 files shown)


hm21 (approved):
Reviewed and took over one finding — pushed as a11bde3a4, separate commit, no force-push.

### Finding: the hex alternative fired inside spans the UI never links

`_decodableReferencePattern` held only the bech32 and hex alternatives, so it could not see that `LinkifiedTextSpanBuilder` resolves alternation per start position. A URL or hashtag containing a hex run starts earlier and wins there — `https://host/<64-hex>` is a single URL token in the UI and the hex inside it is never a reference — but the cut was pulled back to it anyway:

```
before this PR:  "Look at this https://blossom.divine.video/a1b2c3d4..."  (53)
on b09b6e7ac:    "Look at this https://blossom.divine.video/..."          (45)
```

Blossom and CDN media URLs are exactly `https://host/<sha256>`, so that is the common shape rather than an edge case. `#<64-hex>` had the same problem, dropping the preview to `nice one #...`.

### Fix

Mirror the surrounding alternatives from the UI tokenizer and protect only capture groups 3 and 4. That makes the rule your doc comment states — "exactly the token spans the UI can decode, no more, no less" — actually true instead of aspirational.

Two calls in there that are yours to overrule:

- The UI's trailing `@([a-zA-Z][a-zA-Z0-9_]{0,30})` alternative is deliberately **not** mirrored. It beats bech32 for `@npub1…` and truncates the token to 31 characters, so mirroring it would drop the protection #6346 added for that input. Whether the UI should tokenize `@npub1…` as a mention at all is a separate question from this cut.
- Consequence, pinned by a new test: a bech32 token inside a URL path is no longer pulled back either, because the UI reads that as a URL too. Drop that test and the group-1 alternative if you would rather keep the old behaviour there.

### Validation on a11bde3a4

- `flutter test` in `notification_repository` — 172 pass; `--coverage` 95.68%, above the 94 gate
- `flutter analyze` clean in the package and `flutter analyze lib test integration_test` clean in `mobile/`
- `flutter test test/notifications/` — 210 pass; `dart format` clean
- Each of the three new tests was run against b09b6e7ac first and fails there

All checks are green. `Tests (shard 3/4)` failed once on `test/providers/personal_event_cache_service_provider_test.dart:218` and passed on re-run — that file is outside both diffs, and its `waitForInitialized` helper is a bounded 20-turn `pumpEventQueue` poll that gives up silently, so it is a pre-existing timing flake rather than anything from this branch.

### One process note

"Each new test was confirmed to fail against pre-fix code before being kept" does not hold for `still treats punctuation-only lead-in as leading for a bech32 token`, `renders an undecodable npub as raw text`, or `renders a partial hex reference as raw text` — all three are green against pre-fix code. They are fine as guard tests; just worth not claiming otherwise.

Yours to merge.

realmeylisdev:
Thanks @hm21 — checked a11bde3a4 rather than taking it on trust, and it holds. Both calls stand as you pushed them.

**The finding.** Reproduced: your three tests fail on b09b6e7ac with exactly the output you quoted, including `nice one #...` collapsing a 50-char budget to 13. It is a regression my own commit introduced — `main` never matched hex at all, so `https://host/<sha256>` was cut plainly there. Adding the hex alternative is what let it match inside a span the UI never links. Diffed the mirrored pattern against `_combinedRegex` character by character: identical, the `@mention` alternative being the only difference.

**`@mention` not mirrored — agreed.** Confirmed `@npub1…` resolves to group 5 at 32 characters, so the UI truncates it to a mention either way and mirroring would drop #6346's protection for no gain. Keeping it out.

**bech32 inside a URL — keeping your behaviour, not dropping the test.** The two options aren't separable: bech32-in-URL and hex-in-URL are the same group-1 alternative, so dropping it to restore the old behaviour would put the hex-in-URL regression straight back, and that's the Blossom/CDN shape. The rationale doesn't survive the split either — the UI decodes neither of them, and "protect exactly what the UI can decode" is the whole reason the cut exists. Noted in the description as a deliberate change against `main` so it isn't discovered later as a surprise.

**Process note — you're right, and it was my error.** Ran all three against `origin/main`: `still treats punctuation-only lead-in as leading for a bech32 token`, `renders an undecodable npub as raw text`, and `renders a partial hex reference as raw text` are green there. They're guard tests, which is fine, but "each new test was confirmed to fail first" was wrong. Description now states red/green per test instead of blanket-claiming it, and picks up your defect as a third section with the numbers and both consequences.

Re-ran locally on the tip: 172 pass in the package, 95.68% against the 94 gate, 210 in `test/notifications/`, analyze clean in both, format clean. All checks green. Agreed on `personal_event_cache_service_provider_test.dart` — outside both diffs and a bounded poll that gives up silently.

Nothing outstanding on my side.

PR #6840, opened 2026-08-07
fix(local-stack): resolve the local host per platform

### LINKED ISSUE #6824: fix(config): localHost is hardcoded to 10.0.2.2, so LOCAL cannot reach local_stack from the iOS Simulator
## Problem

`AppEnvironment.local` cannot reach `local_stack` from the iOS Simulator. Every local endpoint is built from a single unconditional constant:

`mobile/lib/models/environment_config.dart:12`
```dart
const localHost = '10.0.2.2';
```

`10.0.2.2` is the Android emulator's NAT alias for the host machine. There is no `Platform.isIOS` branch anywhere — that constant feeds the relay URL (`:73`), the API base (`:86`), blossom (`:122`), the indexer relays (`:133`), the relay manager (`:146`) and keycast (`auth_providers.dart:55`).

`AGENTS.md` claimed the opposite ("Either Android emulator or iOS Simulator works against the local stack out of the box"). That line is corrected in #6823; this issue is the code fix.

## Reproduction

`flutter run -d <ios-simulator> --dart-define=DEFAULT_ENV=LOCAL` with the stack up:

```
[14:43:46.119] Creating NostrClient via factory with Divine relay: ws://10.0.2.2:47777
[14:45:14.633] [NostrService] Failed to initialize client in build():
               TimeoutException after 0:01:30.000000: Nostr client initialization
               timed out: source=build, stage=client.connectingRelays, elapsedMs=90005
[14:46:51.159] ... retry ... -> "Scheduling initialization retry 2 in 0:00:10"
[14:48:31.195] ... retry ... -> "Scheduling initialization retry 3 in 0:00:20"
```

It never connects; it loops on 90-second timeouts with backoff.

From the host (the Simulator shares this network stack):
```
localhost:47777  -> HTTP=200  time=0.003s
127.0.0.1:47777  -> HTTP=200  time=0.003s
10.0.2.2:47777   -> HTTP=000  time=5.009s   (curl exit 28, timeout)

route -n get 10.0.2.2 -> gateway: 192.168.1.1, interface: en0
```

No interface owns `10.0.2.x`, so this is network-independent rather than an artifact of one LAN. On a network that *does* have a host at `10.0.2.2`, the app would silently talk to a stranger's machine instead.

## Suggested fix

Branch the constant on platform — `10.0.2.2` for Android, `localhost` (or `127.0.0.1`) elsewhere. The `test/manual/*` acceptance tests already do this correctly, each defining its own `const _localHost = 'localhost'`.

Both hosts are already permitted for cleartext on both platforms (`NSAllowsLocalNetworking` on iOS, the `<domain-config>` block on Android), so no transport-security change is needed. Worth checking `mobile/scripts/check_native_transport_security.sh` still passes either way.

## Adjacent observation

In the same run, while in LOCAL mode, the app logged:
```
[14:43:46.321] [AUTH] Discovered 1 user relays from cache
[14:43:46.321] [AUTH]   - wss://relay.divine.video (read: true, write: true)
```
A cached production relay resolving inside LOCAL is a separate concern — filed separately.

Found while investigating #6594.

realmeylisdev:
## Description

`AppEnvironment.local` hardcoded `const localHost = '10.0.2.2'` and used it for the relay, REST API, Blossom, Keycast, and the indexer relays, with no platform switch anywhere in `mobile/lib` or `mobile/packages`. `10.0.2.2` is an Android-emulator-only alias; the iOS Simulator, macOS, and web all share the host's network stack, where it is an ordinary routable LAN address. Measured on a Mac with the stack running:

```
curl http://localhost:47777  -> 200        (relay WS handshake: 101 Switching Protocols)
curl http://10.0.2.2:47777   -> timeout    (route: gateway 192.168.1.1, interface en0)
```

So LOCAL mode on the iOS Simulator could not reach any local endpoint.

Nothing on the server or platform side needed to change — the stack already supported it:

- `local_stack/docker-compose.yml` sets `ALLOWED_TENANT_DOMAINS: localhost,10.0.2.2`, so Keycast already accepts `localhost` (and rejects `127.0.0.1` with `HTTP 400 Invalid domain`, which is why `localhost` and not the IP is the right choice).
- `check_native_transport_security.sh` already expects `10.0.2.2 127.0.0.1 localhost` as cleartext-allowed in the native configs.

**Host resolution.** `localHost` is now a getter resolving to `androidEmulatorHost` on Android and `loopbackHost` everywhere else, guarded by `kIsWeb` so a mobile browser doesn't get the emulator alias. Every LOCAL endpoint — `relayUrl`, `apiBaseUrl`, `blossomUrl`, `indexerRelays`, `relayManagerApiUrl` — derives from that single getter; `git grep 10.0.2.2 -- mobile/lib` returns no host-selection hardcode.

**Invite server.** Raised in review: the invite server was the one local endpoint that did not route through `localHost`. `AppConfig.inviteServerBaseUrl` has a compiled-in default of `https://invite.divine.video` and no platform awareness, so a LOCAL run that omitted `--dart-define=INVITE_SERVER_URL` silently talked to the **production** invite service instead of failing against an unreachable local host. `EnvironmentConfig.inviteBaseUrl` now takes a LOCAL branch returning `http://$localHost:$localInvitePort`, and `main.dart` builds `InviteApiClient` from `ref.read(currentEnvironmentProvider).inviteBaseUrl` rather than reading `AppConfig` directly — `AppConfig` cannot see `AppEnvironment`, so the environment layer is the right place for the branch. A `!const bool.hasEnvironment('INVITE_SERVER_URL')` guard keeps an explicit define winning, so `local_stack/run_android_local.sh` is unaffected, and `AppConfig`'s `defaultValue` is untouched so `check_backend_host_defaults.sh` still passes.

**Docs.** `AGENTS.md` and `README.md` claimed the stack worked "out of the box" for both the Android emulator and the iOS Simulator. Only the emulator did. Both now describe per-platform host resolution, name macOS as the third configured native platform, and state that the Android emulator is the only fully turnkey target. The web bullet's stated reason was also wrong and is corrected: **two** pipelines build web — `mobile_web_production_deploy.yml` pins `DEFAULT_ENV=PRODUCTION`, while `mobile_pr_preview_build.yml` pins no `DEFAULT_ENV` at all and inherits the compiled `PRODUCTION` default. CORS is not the obstacle for a LOCAL web run; nothing selects LOCAL in the first place.

**Related Issue:** Closes #6824. Found while investigating #6762.

## Out of Scope

- `BLOSSOM_PUBLIC_URL` is baked into the media URLs the seeder mints, so it cannot be resolved per client. It keeps `10.0.2.2` as its default (no behaviour change) and becomes overridable via `${BLOSSOM_PUBLIC_URL:-…}` for non-Android runs. Documented in `AGENTS.md` and `README.md`.
- The loopback allowlists in `blossom_settings_cubit.dart`, `background_uploader`, `dm_repository`, and `videos_repository` keep listing all three hosts. Those are validators, not host selection, and are correct as they are.
- A physical Android device is not a supported local-stack target — `10.0.2.2` is emulator-only and a device needs the host's LAN address. Unchanged by this PR.

## Verification

- `flutter analyze lib test integration_test` — **No issues found**
- Full `mobile/` test suite — **15101 passed**, 285 skipped, 0 failures, on the rebased head
- `dart run build_runner build --delete-conflicting-outputs` — no drift; `auth_providers.g.dart` regenerates identically and `git status` stays clean
- All 32 `mobile/scripts/check_*.sh` ratchets — 31 pass. `check_ios_extension_signing_contract.sh` exits 1 without the `IOS_EXTENSION_BUNDLE_IDS` env var that Codemagic supplies (`codemagic.yaml:38-40`); it behaves identically on a clean `origin/main` checkout and is not a GitHub Actions check.
- `docker compose --profile seed config` — resolves `BLOSSOM_PUBLIC_URL` to `http://10.0.2.2:43003` bare and to `http://localhost:43003` with the override exported, against the compose file as rewritten by #6823
- Confirmed end to end on an iPhone 17 Pro Simulator against the running stack: before the change Keycast returned `HTTP 400 Invalid domain: 127.0.0.1`; after pointing at `localhost` the NIP-98 signing call reached the right tenant.

### Notes for review

- **Rebased onto current `origin/main`** to resolve a conflict with #6823, which landed first and rewrote the same `AGENTS.md` bullet to assert that the iOS Simulator does not work because `localHost` is hardcoded. That is exactly the claim this PR makes false, so the resolution drops it. `AGENTS.md` is the only file that conflicted; `local_stack/docker-compose.yml` and `mobile/lib/main.dart` auto-merged in non-overlapping regions and both sides were verified present afterwards.
- `defaultTargetPlatform` is `TargetPlatform.android` under `flutter test`, so the pre-existing assertions on `ws://10.0.2.2:47777` and `http://10.0.2.2:47777` still hold. Those assertions now pin `debugDefaultTargetPlatformOverride` explicitly rather than inheriting the harness default, and the new tests declare Android, iOS, and macOS individually.
- Three `const` declarations had to become `final`/non-const because the host is resolved at runtime (`integration_test/helpers/relay_helpers.dart`, `integration_test/auth/repro_log2_delete_test.dart`, and the `OAuthConfig` in `auth_providers.dart`). `flutter analyze` caught all three; no const consumers remain.
- `loopbackHost` is deliberately kept for the Keycast redirect URI in `repro_log2_delete_test.dart` — Keycast's `origin_is_allowed()` admits `localhost`, not the emulator alias.

## Type of Change

- [x] 📝 Documentation


NotThatKindOfDrLiz:
@realmeylisdev I pushed a small takeover follow-up on top of your branch.\n\nChanges:\n- narrowed the AGENTS.md local-stack guidance so it no longer claims Flutter web works out of the box\n- documented the LOCAL invite-server define needed for iOS Simulator/macOS runs\n- kept the Blossom override docs scoped to native loopback clients\n- named the local relay-manager port and extended the environment-config endpoint test to cover it\n- updated the PR body with Closes #6824; the needs-issue label is now gone\n\nVerification:\n- flutter test test/models/environment_config_test.dart\n- flutter analyze lib/models/environment_config.dart test/models/environment_config_test.dart test/manual/nip98_relay_acceptance_test.dart integration_test/helpers/constants.dart\n- pre-push hook: analyzer, generated-file verification, test/models/environment_config_test.dart, test/providers/auth_providers_test.dart\n- GitHub checks are green after the push/body update\n\nPlease review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up. The remaining issues from review were addressed: docs no longer overclaim web support, native local-run caveats are explicit, #6824 is linked for closure, and the endpoint regression test now includes relayManagerApiUrl. CI is green.

AGENTS.md
```diff
@@ -146,7 +146,10 @@ If a Divine Brain search or ask tool is available, you may use it for company me
...
-- Either Android emulator (`10.0.2.2` → host) or iOS Simulator (`localhost` → host) works against the local stack out of the box.
+- Host resolution for local app endpoints is platform-specific: Android emulator uses `10.0.2.2`, while iOS Simulator and macOS use `localhost`. `localHost` in `mobile/lib/models/environment_config.dart` resolves the right one per platform — do not hardcode either alias in app code.
+- `BLOSSOM_PUBLIC_URL` is baked into the media URLs the seeder mints, so it cannot be resolved per client. It defaults to the Android emulator alias; export `BLOSSOM_PUBLIC_URL=http://localhost:43003` before `local_stack/up.sh` when running against the iOS Simulator or macOS.
+- Full LOCAL app runs also need the invite server define. `local_stack/run_android_local.sh` passes `INVITE_SERVER_URL=http://10.0.2.2:43004` for the Android emulator; iOS Simulator and macOS runs should pass `--dart-define=INVITE_SERVER_URL=http://localhost:43004`.
+- Web resolves local endpoints to `localhost`, but Flutter web runs are not out of the box because the browser enforces CORS and the local stack does not allow arbitrary `localhost:<dev-port>` origins.
```
mbradley: @NotThatKindOfDrLiz this is your line from the takeover commit, so flagging it to you rather than to Meylis.

Narrowing the original "web works out of the box" claim was right, but the stated reason does not hold. Checking the services' sources: funnelcake's API sets `Access-Control-Allow-Origin: *` (`crates/api/src/cors.rs`, whose module doc is literally "Open CORS policy for the public funnel API"), the relay sets it on publish and management responses, and blossom sets it in `main.rs`. Keycast is the only service that pins origins, via `ALLOWED_ORIGINS` in `docker-compose.yml:62`. Separately, WebSocket handshakes are not subject to CORS at all, so CORS cannot explain a failed relay connection even in principle.

The `<dev-port>` half is accurate: `flutter run -d chrome` selects a random port unless `--web-port` is passed, and nothing in the repo pins it.

Two caveats on my end. I read those services from local checkouts, and the stack runs `:latest` images that #6823 reports are frozen at February, so the running code may differ from what I read. And the bullet is worth keeping in some form: LOCAL on web is only reachable through a dev `flutter run -d chrome --dart-define=DEFAULT_ENV=LOCAL`, since LOCAL is absent from the env picker and the web deploy pins `DEFAULT_ENV=PRODUCTION`.

Could you name what actually failed when you tried it? If it was Keycast's pinned origins, the bullet should say so.

AGENTS.md
```diff
@@ -146,7 +146,10 @@ If a Divine Brain search or ask tool is available, you may use it for company me
 ## Local Stack Development
 
 - The local Docker stack (`local_stack/`) speaks cleartext on `10.0.2.2`, `localhost`, and `127.0.0.1`. Cleartext to those loopback hosts is permitted in every build type on both platforms — Android via the `<domain-config>` block in `mobile/android/app/src/main/res/xml/network_security_config.xml`, iOS via `NSAllowsLocalNetworking=true` in `mobile/ios/Runner/Info.plist`. Remote cleartext is rejected on both platforms in every build type.
```
mbradley: This bullet says "both platforms" and names only Android and iOS, and line 153 says "either native config", but this PR promotes macOS to a documented target three times. macOS is genuinely configured (`mobile/macos/Runner/Info.plist` sets `NSAllowsArbitraryLoads=false` with `NSAllowsLocalNetworking=true`) and genuinely guarded (`check_native_transport_security.sh` calls `check_ats_plist "macos/Runner/Info.plist"`). Worth making it three platforms here, with the macOS plist referenced alongside the other two.

mobile/lib/models/environment_config.dart
```diff
@@ -20,6 +39,7 @@ const localRelayPort = 47777;
 const int localApiPort = localRelayPort;
 const localBlossomPort = 43003;
 const localInvitePort = 43004;
+const localRelayManagerPort = 8787;
```
mbradley, line 46: This lands under the `/// Local Docker stack port mappings.` header, but 8787 is not a local stack port. `grep -rn "8787" local_stack/` returns nothing and there is no relay-manager service in the compose file; 8787 is the `wrangler dev` default for divine-relay-manager, run separately by hand. The value is faithful to the `8787` that was inlined before, so this is not a regression, but naming it into that block now asserts something false for one of its six members. The other five all match real compose mappings.

```suggestion
/// Not in local_stack: wrangler dev default for divine-relay-manager,
/// which is run separately.
const localRelayManagerPort = 8787;
```

mobile/integration_test/helpers/constants.dart
```diff
@@ -3,12 +3,15 @@
 
 export 'package:openvine/models/environment_config.dart'
     show
+        androidEmulatorHost,
```
mbradley: `androidEmulatorHost` and `localRelayManagerPort` are re-exported here but have no consumer under `integration_test/`. The test that uses `localRelayManagerPort` imports `environment_config.dart` directly.

mbradley (changes requested):
Verified independently at `eae8921` before writing this. Every LOCAL endpoint resolves through the new `localHost` getter with no host-selection hardcode left in `mobile/lib` or `mobile/packages`; the loopback validators already accepted `localhost`, so the iOS switch does not trip them; each of the four new tests fails under a mutation in one direction or the other; the committed `auth_providers.g.dart` matches a fresh `build_runner` run with nothing else stale; and `check_process_global_mutations.sh` passes, since the new `tearDown` is the reset-to-default form that script sanctions. `flutter analyze lib test integration_test` is clean. The platform fix itself is sound and I could not break it.

Everything below is documentation accuracy plus two questions here and one inline. I have not pushed anything. Liz's takeover handback is still open with Meylis, and two of these items concern text that commit added, so they belong in discussion rather than in a second reviewer's amendments.

**Three fixes that fall outside the diff, or outside a diff hunk, so noting them here rather than inline:**

1. `README.md:80` still says the stack works "out of the box" for both the Android emulator and the iOS Simulator, and then points readers to `AGENTS.md`, which this PR updates to say the opposite. That is the same claim this PR removes from `AGENTS.md` as false, still on the repo's front page.

2. `local_stack/seed/seed.py:33` and `:562` still describe `BLOSSOM_PUBLIC_URL` as "the emulator-accessible address". `docker-compose.yml` got a clear comment explaining the new override; the file that actually consumes the variable did not.

3. `mobile/test/models/environment_config_test.dart:71` and `:100` hardcode `10.0.2.2` and set no platform override. They were platform-independent while `localHost` was a `const`; now they pass only because an SDK assert block forces `TargetPlatform.android` whenever `FLUTTER_TEST` is set. They sit directly above four new tests that declare their platform explicitly. Worth pinning `debugDefaultTargetPlatformOverride = TargetPlatform.android` in them so the assumption is stated rather than inherited from the harness.

**Two questions:**

4. The invite server is the one local endpoint that does not route through `localHost`. `mobile/lib/config/app_config.dart:17` reads `INVITE_SERVER_URL` with a compiled-in default of `https://invite.divine.video`, so a LOCAL run on the Simulator or macOS that omits the define talks to the production invite service rather than failing against an unreachable local host. `localInvitePort = 43004` already exists in `environment_config.dart` and has no consumer in `lib/`. Is the documented `--dart-define` the intended answer, or should `AppConfig` take a LOCAL branch the way every other endpoint now does? Note `check_backend_host_defaults.sh` pins that `defaultValue` to `*.divine.video`, so a changed default is not an option; it would have to be a LOCAL branch.

5. #6823 rewrites the same `AGENTS.md` line this PR rewrites, to text asserting the iOS Simulator does not work and `localHost` is hardcoded. Both are open and approved. `git merge-tree` on the two branches conflicts, and `AGENTS.md` is the only conflicting file. Which PR should carry that bullet? Whichever lands second will need a conflict resolution that does not restore the stale claim.

mobile/lib/models/environment_config.dart
realmeylisdev: Applied verbatim in 6a44689e9. Confirmed: no 8787 anywhere in `local_stack/`, no relay-manager service in the compose file, and the other five constants each map to a real published binding (43000, 47777, 43003, 43004, plus `localApiPort` aliasing `localRelayPort`).

mobile/integration_test/helpers/constants.dart
realmeylisdev: Both dropped in b958c9dc2. `loopbackHost` stays — `repro_log2_delete_test.dart:34` uses it for the Keycast redirect URI.

Worth noting `localApiPort` and `localInvitePort` are dead there too, but they predate this PR so I left them.

AGENTS.md
realmeylisdev: Fixed in 4e63fa9df. The bullet now names all three native platforms with the macOS plist alongside the other two, the Charles/mitmproxy parenthetical covers macOS, and the last bullet says "any native config" and names the four files the guard actually walks.

Same commit fixes a comment my own rename invalidated: `network_security_config.xml:15-16` said the Dart-side source of truth for the emulator-host loopback is `localHost`, which stopped being true the moment `localHost` became platform-resolving. It is `androidEmulatorHost` now.

AGENTS.md
realmeylisdev: You are right, and it is worse than the WebSocket point alone — Keycast does not block dev ports either. `origin_is_allowed()` returns true for any `http://localhost:<port>` before `ALLOWED_ORIGINS` is consulted, so the one service that pins origins explicitly whitelists the exact case the bullet claimed was blocked. With funnelcake, the relay, and blossom all sending `*`, no service in the stack rejects a dev-port origin.

Rewritten in 6c840b69f. The conclusion survives for a reason checkable in this repo rather than inferred: `AppEnvironment.local` is absent from the picker in `developer_options_screen.dart:133-139`, and the only web deploy pins `DEFAULT_ENV=PRODUCTION` (`mobile_web_production_deploy.yml:48`). The bullet now says that and states plainly that CORS is not the obstacle, so the next reader does not re-derive the wrong diagnosis.

Your caveat carries over into mine: I read those four services from local checkouts, and the stack runs `:latest`.

@NotThatKindOfDrLiz — your question stands. If you hit something concrete on web, it was not CORS, and I would rather document what you actually saw than leave it at "nothing selects LOCAL".

realmeylisdev:
@mbradley thanks — every claim checked out as written, including the two that were mine to own rather than Liz's. All nine items are addressed, one commit each. Nothing declined, nothing escalated.

| # | Item | Commit |
|---|---|---|
| 1 | `README.md:80` "out of the box" | `e65ddc0f2` |
| 2 | `seed.py:33`, `:562` "emulator-accessible address" | `e65ddc0f2` |
| 3 | `environment_config_test.dart` platform pins | `13a247289` |
| 4 | Invite server LOCAL branch | `7a35c3841` |
| 5 | #6823 `AGENTS.md` conflict | answered below |
| — | `AGENTS.md:148` macOS | `4e63fa9df` |
| — | `AGENTS.md:152` web/CORS | `6c840b69f` |
| — | `localRelayManagerPort` | `6a44689e9` |
| — | `constants.dart` re-exports | `b958c9dc2` |

**4 — invite server. This is a behaviour change, so flagging it rather than burying it in the table.** I took the LOCAL branch rather than the documented define. `AppConfig` is a static-const class with no access to the runtime-switchable `AppEnvironment` — its own `environment` field is a different define entirely — so the branch went on `EnvironmentConfig` next to the endpoints it already resolves, mirroring `blossomUrl`. `main.dart` reads it through `currentEnvironmentProvider`, the same shape as the adjacent `blossomBaseUrl` line.

An explicit `--dart-define=INVITE_SERVER_URL` still wins via `bool.hasEnvironment`, so `run_android_local.sh` is unchanged and `test_android_scripts.sh:154` still passes. As you noted, the `defaultValue` is untouched, so `check_backend_host_defaults.sh` stays green — I ran it. Blast radius is LOCAL builds, where the previous behaviour was to silently talk to production. It retires `localInvitePort` and lets `AGENTS.md` drop the manual-define bullet.

**5 — #6823.** Both PRs are mine, so this is my call to make rather than a coordination problem. `git merge-tree` agrees `AGENTS.md` is the only conflict; `docker-compose.yml` auto-merges. #6823 is approved and clean and its bullet is a correct description of `main` as it stands, so I am not churning it. #6840 carries the resolution at rebase time, and the resolution is to take this branch's version wholesale — it describes the post-fix world. Recording that here so it is not improvised later.

**Not requested, but mine to fix:** `network_security_config.xml:15-16` claimed the Dart-side source of truth for the emulator-host loopback is `localHost`. My rename made that false — it is `androidEmulatorHost` now. Fixed in `4e63fa9df` alongside the macOS wording.

**Looked at, deliberately left:** `localApiPort` and `localInvitePort` are dead re-exports in `constants.dart` too, but they predate this PR. `.claude/skills/e2e-test/SKILL.md:48` says "on both platforms" — incomplete, not false, and scoped to Android E2E.

**On the first CI run**, shard 0 failed on `upload_manager_get_by_path_test.dart` — `expect(pendingUploads.length, equals(1))` saw 2. I did not assume it was unrelated. What I checked: shard-0 file assignment is byte-identical between this branch and `origin/main` (I only added tests to an existing file, which does not move anything); none of that test's dependencies reference `EnvironmentConfig` or `AppConfig`; the full 335-file shard 0 passes locally at CI's exact ordering seed, 3854 tests; and `main`'s own latest run is green. Seven suites in that shard share the on-disk `pending_uploads` Hive box and the failing one is `skip_very_good_optimization`, so it runs alongside the merged bundle — a plausible pre-existing timing path I could not reproduce. The re-run is green. Flagging it as something worth a proper look rather than claiming I have explained it.

All 16 checks green.

rabble (approved):
Approved — thorough platform fix with complete follow-through on the earlier `CHANGES_REQUESTED` items.

**Core fix verified — `mobile/lib/models/environment_config.dart`:**
- Root cause is accurate: hardcoded `const localHost = '10.0.2.2'` only works on Android emulator; iOS Simulator/macOS/web share the host stack where it times out. Repro `(curl localhost:47777 → 200, 10.0.2.2 → timeout)` and Keycast `400 Invalid domain: 127.0.0.1` diagnosis check out.
- New `androidEmulatorHost`/`loopbackHost` constants + `String get localHost => !kIsWeb && defaultTargetPlatform == TargetPlatform.android ? androidEmulatorHost : loopbackHost` correctly routes Android→emulator alias, everything else→`localhost`. `kIsWeb` guard prevents mobile-browser mis-routing. Every LOCAL endpoint (`relayUrl`, `apiBaseUrl`, `blossomUrl`, `indexerRelays`, `relayManagerApiUrl`) now derives from that single getter — `git grep 10.0.2.2 pr-6840 -- mobile/lib` returns zero hits, no hardcoded fallback remains.
- `inviteBaseUrl` correctly closes the silent-prod-fallback hole: `AppConfig.inviteServerBaseUrl` defaults to `https://invite.divine.video` and has no platform awareness, so LOCAL without `--dart-define=INVITE_SERVER_URL` now returns `http://$localHost:43004`. `!const bool.hasEnvironment` guard preserves explicit defines (`run_android_local.sh` still works) and `check_backend_host_defaults.sh` still passes because `defaultValue` is untouched. `main.dart` wiring via `ref.read(currentEnvironmentProvider).inviteBaseUrl` is the right layer — `AppConfig` cannot see `AppEnvironment`.

**Secondary fixes (all from mbradley review, now resolved):**
- `README.md:80` no longer claims out-of-the-box iOS support; now explicitly notes only Android is turnkey and documents `BLOSSOM_PUBLIC_URL` override.
- `local_stack/seed/seed.py` comments updated (was stale emulator-only wording).
- `mobile/test/models/environment_config_test.dart:71,100` now pin `debugDefaultTargetPlatformOverride = TargetPlatform.android` instead of relying on `flutter_test`’s `FLUTTER_TEST → android` assert — assumption is now explicit.
- `local_stack/docker-compose.yml` correctly makes `BLOSSOM_PUBLIC_URL` overridable via `${BLOSSOM_PUBLIC_URL:-http://10.0.2.2:43003}` with seeder comment, since media URLs are baked at seed time and cannot be per-client.
- `mobile/android/.../network_security_config.xml` comment corrected (`localHost` → `androidEmulatorHost`+ selector) and macOS semantics added.
- `localRelayManagerPort = 8787` separated from stack mappings (no compose service) — value unchanged.
- Dead re-exports dropped; `loopbackHost` retained for `repro_log2_delete_test.dart` redirect URI (correctly stays `localhost` — Keycast `ALLOWED_TENANT_DOMAINS` allows `localhost` before `ALLOWED_ORIGINS`, not the emulator alias).
- `auth_providers.dart` `const OAuthConfig → OAuthConfig` + `flutter analyze` clean; `auth_providers.g.dart` hash `ee45ee8…` matches fresh `build_runner`.

**Tests & CI:**
- `flutter test test/models/environment_config_test.dart` on `pr-6840`: **28 passed** (4 `localHost` + 3 `inviteBaseUrl` + 21 existing). `./…/environment_config_test` local tracks per-platform grouping verified. Existing LOCAL assertions still pass because harness defaults to Android — now explicitly pinned.
- `gh pr checks 6840`: all 15 checks green (Analyze, Format, Generated Files, 4× Tests shards, Build Web Preview, etc.). Docs-only changes to `AGENTS.md` (macOSATS, CORS correction — verified funnelcake/relay/blossom all send `*`, Keycast allows `http://localhost:<port>`, WS not CORS — accurate) need no extra runtime change.

No blocking issues. The one physical-device caveat (Android device ≠ emulator) is out of scope — local stack targets emulator/simulator/macOS only.

mbradley (approved):

PR #6842, opened 2026-08-07
feat(video-recorder): show the remaining shot budget in stop-motion

### LINKED ISSUE #6843: feat(video-recorder): show the remaining shot budget in stop-motion
Auto-created from PR #6842.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6842
**Branch:** `feat/stop-motion-shot-budget`

hm21:
Closes #6843

## Description

Stop-motion never enters the recording state, so capture mode's duration-based progress bar never appeared. Users shot with no indication of how much room was left in the clip — the only feedback was the capture counter on the library button, which says how many stills exist, not how many still fit.

The top bar's center slot now carries the same bar as capture mode, counted in stills instead of milliseconds, with the remaining count above it.

**Why the frame rate changed too.** The ceiling is derived from `maxDuration` and the render frame rate rather than written down, so the number on screen cannot drift from what the encoder actually produces. Deriving it surfaced a mismatch that predates this PR: `StopMotionRenderService` assembled at 24fps, while the editor timeline's frame ruler (`VideoEditorTimelineRulesIndicator._fps`) has always labelled ticks at 30fps. A stop-motion clip's "frames" therefore meant two different things depending on where you looked. The renderer now matches the ruler at 30fps, which puts the budget at 188 stills — 6.3s divided by a one-frame hold, which rounds up to 33334µs so a session that exactly fills the minimum never lands a few microseconds under the length gate.

**Why the count sits above the bar.** A `Row` centers its children, so the bar plus its label would land on the close/next buttons' center line as a group, leaving the bar itself riding low. An invisible mirror of the label below the bar balances the group so the bar lands on that line. It is measured from real text rather than a hardcoded height, so it tracks the font and the user's text scale.

**Why overflow is not blocked.** Shooting past the ceiling keeps working and grows the overflow segment, matching how capture mode lets a recording run past the maximum duration and leaves the trim to the editor. The counter floors at "No shots left".

**Behaviour change to be aware of:** frame holds are stored as absolute durations, so existing stop-motion drafts captured on the 24fps grid get requantized by the encoder and play about a quarter faster.

**Counted against the composition, not the session.** Opening the camera from the editor to top up an existing stop-motion clip starts a session with no frames of its own, so the budget is measured against the clip manager's total — the same source capture mode's bar uses — and at the hold the new stills will inherit, since the editor re-applies the target clip's global hold when it splices them in.

**Related Issue:** 

## Out of Scope

- Migrating existing stop-motion drafts to the 30fps grid. Their stored holds are reinterpreted rather than rewritten; if that turns out to matter for shipped drafts, it wants its own change.
- Hard-blocking the shutter at the ceiling. Current behaviour (overflow, trim in the editor) is preserved deliberately.

## Verification

- `flutter test test/models test/widgets test/services test/blocs test/screens test/l10n` — 12200 passing
- `flutter test test/l10n/arb_consistency_test.dart` — new key mirrored into all 21 non-English locales, no untranslated debt added
- Pre-push hook (format, analyze, codegen, changed-file tests) — green
- [x] Manual pass on a real Samsung Galaxy

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mobile/lib/widgets/video_recorder/modes/capture/video_recorder_stop_motion_budget.dart
```diff
@@ -0,0 +1,70 @@
...
+  /// Gap between the count and the bar. Tighter than the recording bar's,
+  /// which has the whole top bar to itself — this has to fit between two
+  /// buttons. Also the outer gap to the mirrored spacer below, which only
+  /// balances the group while both gaps match.
+  static const double _labelSpacing = 2;
+
+  @override
+  Widget build(BuildContext context) {
+    final captured = context.select(
+      (VideoRecorderBloc b) => b.state.stopMotionFrameCount,
```
realmeylisdev, line 41: **should-fix** — this is the recorder *session*'s frame count, which is always 0 when the recorder is opened from the editor to add stills to an existing stop-motion clip. so in that flow the bar renders "189 shots left" over an empty track no matter how full the composition already is.

that flow is a first-class path, not an edge case:

- `video_recorder_bloc.dart:313-327` explicitly forces `stopMotion` mode for it (*"Opened over a stop-motion composition to add more stills"*) and leaves `stopMotionFrames` empty
- `video_editor_main_actions_sheet.dart` gates only the timeline-marker item on `isStopMotion`; the Camera item is always shown
- `video_editor_screen.dart:_insertCapturedStopMotionFrames` then splices the new stills into the target clip, so they really do count against the same 6.3s ceiling

it is also off by more than the frame count there: that same splice re-applies the target's global hold via `setGlobalHold(capturedFrames, globalDefaultFramesPerImage(targetFrames))`. if the clip holds each still for 3 output frames, only ~63 more stills fit, not 189.

the data needed is already in scope — the bloc itself reads `isStopMotionComposition(_readClipManager().clips)` at line 315, so the target's frame count and global hold are reachable from the same provider. worth either subtracting the composition when one is being edited, or hiding the bar in the `fromEditor` case rather than showing a number that is wrong.

mobile/lib/models/stop_motion/stop_motion_frame_ops.dart
```diff
@@ -34,6 +34,26 @@ abstract class StopMotionFrameOps {
...
+  /// past [VideoEditorConstants.maxDuration].
+  ///
+  /// A session long enough to clear [minimumInitialDuration] on its own holds
+  /// every still for [defaultFramesPerImage] output frame, so the ceiling is
+  /// however many output frames fit in the maximum clip length. Shorter
+  /// sessions are stretched ([initialFramesPerImage]) but stay far below it.
+  static final int maxCaptureFrames =
+      VideoEditorConstants.maxDuration.inMicroseconds *
+      StopMotionRenderService.defaultFrameRate ~/
+      (Duration.microsecondsPerSecond * defaultFramesPerImage);
```
realmeylisdev: **should-fix** — the worked example just below (lines 82-86, in `initialFramesPerImage`) still describes the 24fps grid, and after this change every number in it is wrong:

| claim | at 24fps | at 30fps |
|---|---|---|
| 3 stills, N frames each | 8 | **10** |
| 12 stills, N frames each | 2 | **3** |
| 12 stills sum | 999,996µs | **1,200,000µs** (above the gate, no bump at all) |
| one frame | ~41ms | **~33ms** |
| worst shortfall | ≤4µs (at 12) | **10µs (at 30 stills)** |

the 12-stills example no longer even demonstrates the case it was written for, since at 3 frames each it clears the minimum outright. the `+1` logic is still correct (10µs against a 33ms frame), so this is comment drift only — but it is the one spot in the file that did not get the 24→30 pass the rest of the diff applied.

(`stop_motion_render_service.dart:216` has the same flavour of staleness, "a single still ≈ 83ms", though that one predates this PR.)

mobile/lib/l10n/app_ur.arb
```diff
@@ -5120,6 +5120,14 @@
     "description": "Snackbar shown when the user taps the record button in lip-sync mode without first selecting a sound."
   },
   "videoRecorderStopMotionAssembleFailed": "ویڈیو نہیں بن سکی۔ دوبارہ کوشش کریں۔",
+    "videoRecorderStopMotionShotsLeft": "{count, plural, =0{کوئی شاٹ باقی نہیں} =1{1 شاٹ باقی} other{{count} شاٹس باقی}}",
```
realmeylisdev: **nit** — this block landed between `videoRecorderStopMotionAssembleFailed` and its own `@videoRecorderStopMotionAssembleFailed` metadata, and at 4-space indent while the rest of the file uses 2. valid JSON and gen-l10n does not care, just reads odd next to its neighbours.

realmeylisdev (dismissed):
two findings inline. the shot budget always starts at 189, including when you open the camera from a stop-motion editor to top up a clip that's already full, so it reports room that isn't there. and the frames-per-image worked example in `stop_motion_frame_ops` still quotes the 24fps numbers.

the 30fps move itself checks out, the ruler really has been on 30 all along, and the l10n pass covers all 22 locales.

mobile/lib/widgets/video_recorder/modes/capture/video_recorder_stop_motion_budget.dart
hm21: Fixed in f4693114f. The bar now reads `clipManagerProvider.totalDuration` — the same source capture mode's bar uses — so the editor top-up flow starts from what the composition has left.

Your second half was the part I had missed: `_insertCapturedStopMotionFrames` re-applies the target's global hold, so new stills inherit it. `captureFramesPerImageFor(clips)` reads that hold off the composition, and the budget divides by it — a clip on threes now offers ~63 more stills, not 188.

mobile/lib/models/stop_motion/stop_motion_frame_ops.dart
hm21: Fixed in 4a99d7cc8, but not as comment drift — the `+1` is not still correct.

Your own table has the case: 30 stills at one frame each sum to 999,990µs, 10µs under the gate. The `+1` then adds a frame to *every* still, and on a one-frame hold that doubles the session. Measured against the real code:

```
n=29  fpi=2  1.933s
n=30  fpi=2  2.000s   <- should be ~1.0s
n=31  fpi=1  1.033s
```

At 24fps the analogous count was safe because 1e6/24 rounds up, so the 24->30 move introduced it.

`framesPerImageToDuration` now rounds the hold up, which removes the shortfall at the source and makes the correcting frame unreachable — so it and its worked example are gone rather than rewritten. The ceiling is derived from that hold instead of from the frame rate, which puts it at 188, not 189.

`stop_motion_render_service.dart:216` fixed in the same commit.

mobile/lib/l10n/app_ur.arb
hm21: Fixed in 4a8a6a045. Same misplacement in `ms`, `vi` and `zh` — those four are the locales that carry `@` metadata blocks, so the key landed inside one. The other 17 have no metadata and already had it in the right spot. `flutter gen-l10n` produces no diff.

hm21:
@realmeylisdev all three addressed, checks green.

- **budget in the editor top-up flow** — fixed (f4693114f). Also fixed the inherited-hold half you flagged; that one I had missed.
- **24fps worked example** — fixed (4a99d7cc8), but the `+1` it documents was not correct. At exactly 30 stills it doubled the session to 2.0s. Details in the thread.
- **ARB placement** — fixed (4a8a6a045), in `ms`/`ur`/`vi`/`zh`.

**Not requested by the review, worth a look:** the ceiling is now 188, not 189. `framesPerImageToDuration` rounds a hold up (33334µs) so a full session never lands under the length gate, and the ceiling divides the maximum by that hold instead of recomputing from the frame rate. Description updated.

Manual pass on the device: done

rabble (approved):
Verified the three fixups address the prior CHANGES_REQUESTED correctly.

**Prior findings — now resolved:**

- **Editor top-up budget** — `video_recorder_stop_motion_budget.dart` now watches `clipManagerProvider.totalDuration` and `StopMotionFrameOps.captureFramesPerImageFor(s.clips)` (same source capture mode uses) and derives `budget = maxCaptureFramesAfter(committed, framesPerImage)` + `remainingCaptureFrames`. A session opened over a full composition correctly shows `No shots left` / 0 remaining, and a composition on threes budgets ~63 more stills not 188. Tests in `video_recorder_stop_motion_budget_test.dart` cover both the committed-credit and inherited-hold cases.

- **24 → 30fps comment drift** — `stop_motion_frame_ops.dart:27,130-143` now at 30fps, `framesPerImageToDuration` uses `ceil()` (33334µs) so a 30-still session sums to 1,000,020µs not 999,990µs, and the `+1` correction + its 24fps worked example (`8 frames / 2 frames / 999,996µs / 41ms`) are removed. `stop_motion_clip_frame.dart:47` (66667µs), `stop_motion_render_service.dart:30` and `video_editor_timeline_geometry.dart` updated to 30fps / 300ms / 6.6s as well.

- **ARB placement** — `ms`/`ur`/`vi`/`zh` (the 4 locales with `@` metadata blocks) now place `videoRecorderStopMotionShotsLeft`/`@videoRecorderStopMotionShotsLeft` immediately after `@videoRecorderStopMotionAssembleFailed` before `videoRecorderToggleFlashLabel` (e.g. `app_ms.arb:5126`), matching the other 17 locales and `app_en.arb`. `flutter gen-l10n` produces no diff as expected.

**Additional checks on the current head (`4a8a6a0`):**

- Budget ceiling is derived, not hardcoded: `maxDuration 6.3s / 33334µs = 188` (33334*188=6,266,792, 33334*189=6,300,126>max). Good — cannot drift from the encoder.
- `video_recorder_progress_bar.dart` extraction is clean — flex-based segments work for both ms (capture) and still-counts (stop-motion). Overflow renders as `primaryContainer` and counter floors at `No shots left`, matching capture-mode trim-in-editor semantics.
- Centering trick in `video_recorder_stop_motion_budget.dart:50-65` — invisible `Visibility(maintainSize:true)` mirror below the bar, measured from real `VideoRecorderProgressBar.labelStyle` text, so the bar itself lands on the Row’s center line at any font scale. `video_recorder_capture_top_bar.dart:73` uses `?center` in the Row, `video_recorder_screen.dart:426-437` correctly splits `.capture` (empty) vs `.stopMotion` (with budget).
- `VideoRecorderStopMotionBudget` mixes `context.select(VideoRecorderBloc.stopMotionFrameCount)` with `ref.watch(clipManagerProvider)` — consistent with existing recorder code; no stale-closure risk.
- l10n: 22 locales present, plural categories look appropriate (e.g. `ar` has `zero/one/two/few/many/other`, `pl`/`ro` with `few/many`). No untranslated debt.

**Risk to flag (not blocking, already noted in PR description):**

- `defaultFrameRate 24→30` is a behavioural break for existing stop-motion drafts: holds are stored as absolute durations and get requantized, so old clips play ~25% faster. Author marks out-of-scope/migration as follow-up — worth QA attention if shipped drafts exist.

**Nits / follow-ups (non-blocking):**

- No new `24fps` strings remain in `mobile/lib` on this branch — staleness fully cleared.
- When `budget==0` (full composition) `filled` is 0 and `overflow==captured`, the bar renders as pure overflow — intentional but visually all `primaryContainer`; fine as-is.
- Disk was at 1.4Gi free (`/System/Volumes/Data` 100%) during this review so full `flutter analyze`/`flutter test` were not re-run here; review is diff + file-inspection + `gh pr diff` verified against the fixup commits and the 7 new budget tests.

LGTM — approve.

rabble (approved):
Rabble review: stop-motion budget — top bar now shows remaining shots (derived from maxDuration/render fps, fixing pre-existing 24fps mismatch) with progress bar in stills not ms. L10n added for shotsLeft plural. Correctly scoped, no regressions. LGTM.

PR #6844, opened 2026-08-07
fix(a11y): make icons and tight layouts survive system font scaling

### LINKED ISSUE #6845: fix(a11y): make icons and tight layouts survive system font scaling
Auto-created from PR #6844.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6844
**Branch:** `fix/font-size`

hm21:
Closes #6845

## What

Icons and tight layouts now survive the system font-scale setting.

Icons stayed at their fixed pixel size while labels grew with the system text scale, so at large font settings the icons shrank visually relative to the text they sit next to and rows drifted out of proportion. Icon-sized chrome now tracks the text scale, under one shared cap.

## Why these specific choices

**Why one shared scaler, capped at 1.3x.** `DivineIcon.maxScaleFactor` and `DivineIcon.scaleSize(context, size)` are the single contract: `DivineIcon` applies it to its own `size`, and everything sized to *match* an icon calls `scaleSize` — both `DivineIconButton` icon branches, its small-variant tap target, the `DivineButton` loading spinner, the action-menu sheet's leading icon, the raw Material icons in the settings screens, the classic-viners star, the text-editor colour swatch, the video-reply circle and the audio chip's waveform bars. The cap is not arbitrary: uncapped, a 24px icon renders at 48px at the 2.0x platform maximum, which overflows narrow rows — `crossposting_settings_screen_test`'s "narrow viewport and enlarged text do not overflow" case (320pt, 2.0x) catches this and fails without the clamp.

One contract matters beyond tidiness. A cap that lives only inside `DivineIcon` leaves everything sized *against* an icon growing to the raw 2.0x: the same `DivineIconButton` would render a 31.2px glyph from a `DivineIconName` and a 48px one from an `SvgIconSource`, its tap target would reach 96px, and a `DivineButton` would change size when it swapped a 31.2px icon for a 48px spinner.

**Why nine call sites lost their `MediaQuery.textScalerOf(...).scale(...)`.** Once `DivineIcon` owns the scaling, a pre-scaled `size` is applied twice — factor-squared, so 4x at the platform maximum. Those call sites now pass the raw design size. The same trap applied to `OgVinerBadge`, which derived its glyph's font size from an already-scaled dimension and then let the ambient scaler apply again; the glyph is a logo mark sized to its circle, so it opts out of the second pass.

**Why some screens clamp locally instead of opting out.** Screens whose layout cannot absorb unbounded growth wrap their subtree in `MediaQuery.withClampedTextScaling` (bottom nav, feed action column, comment input, content filters, trending hashtags, profile stats, video-editor chips). They still respond to the user's font preference, just within a bound the layout can hold — preferable to `TextScaler.noScaling`, which ignores the preference outright. The four bounds are named constants in `lib/constants/text_scale_limits.dart`, each documenting the constraint it encodes, so the next clamp added has a basis for picking one rather than a fifth bare number.

**Why the audio chip's waveform is its own widget.** The five bars were sized from the chip's `build` context, which sits *above* the `withClampedTextScaling` wrapping the rest of the chip — so the clamp never reached them, and at 2.0x the tallest bar rendered at 32px while the label beside it stopped at 22.4px. `_AudioBars` gives them a context below the clamp.

**Why `softWrap: false` on single-line labels.** With `maxLines: 1` and the default `softWrap: true`, Flutter wraps the text, cuts it at one line, and sets `textDidExceedMaxLines`. `RenderParagraph` picks its fade direction as `didOverflowWidth ? horizontal : vertical`, so the wrapped-then-cut case renders a *bottom* fade. `softWrap: false` keeps the text on one line so the width genuinely overflows and truncation reads horizontally. It is the whole fix — the video action captions keep the ellipsis they shipped with, since swapping in a fade would be a separate visual decision.

## Also in this branch

Drive-by cleanups in files this change already touched:

- Profile more-sheet uses `VineBottomSheetActionMenu` instead of two hand-rolled `InkWell` rows (61 lines to 17, and both entries now render through one code path — previously one used `SvgPicture.asset` and the other `DivineIcon`). That sheet rendered its leading icon in a hard-coded 24px box, so it now scales through `scaleSize` like every other icon.
- Category tile reads `VineTheme.headlineSmallFont` / `labelMediumFont` instead of inline `TextStyle`. Metrics are identical; both font families change, since the raw styles set no `fontFamily` and fell through to the platform default — the title now renders in Bricolage Grotesque and the count in Inter.
- `ProfileStatColumn` gains 4px of horizontal padding, so a label that fades at its edge doesn't run flush into the column divider.
- `scripts/baseline/raw_textstyle.txt` shrinks by one entry, since `categories_tab.dart` drops to zero raw `TextStyle(`.

## Test plan

- [x] `dart format --set-exit-if-changed` on every changed Dart file — clean
- [x] `divine_ui` package: 853 tests pass, 100% line coverage held on all four touched files
- [x] `test/widgets` + `test/screens` (4,776 tests) pass
- [x] All four design-system ceiling ratchets green, plus the untested-services and test-structure floors
- [x] 10 new tests pin the scaling contract: the icon cap and that a tighter ambient clamp beats it, both `DivineIconButton` icon sources landing on one size, the small-variant tap target, the spinner matching the icon box it replaces, the action-menu leading icon, and the waveform bars staying inside the chip clamp. Five of them are regression proofs — each was run against the pre-fix code and observed to fail (the waveform bar measures 32.0 where 20.8 is expected; the icon-source pair measures 48 vs 31.2). The other five guard behaviour that already held, so they stay green either way.
- [x] Manual verification on a real Samsung Galaxy at 1.0x / 1.5x / 2.0x system font size

## Notes for the reviewer

The branch is 9 commits behind `origin/main` and has not been rebased. One of those commits (#6794) does touch `video_feed_item.dart`, which this branch also changes, but the two edits are in different regions — `git merge-tree` merges cleanly and GitHub reports the PR mergeable. Worth a rebase before this leaves draft.

### DIFF
--- a/mobile/lib/constants/text_scale_limits.dart
+++ b/mobile/lib/constants/text_scale_limits.dart
@@ -1,0 +1,41 @@
+// ABOUTME: Upper bounds on the system text scale for layouts whose size is
+// ABOUTME: fixed by something other than the text they contain.
+
+/// Caps applied via `MediaQuery.withClampedTextScaling` around subtrees that
+/// cannot absorb unbounded growth.
+///
+/// A cap still honours the user's font-size preference — it just stops
+/// where the layout breaks. Prefer one over `TextScaler.noScaling`, which
+/// ignores the preference outright.
+///
+/// Icons are capped separately and globally by `DivineIcon.maxScaleFactor`;
+/// the values here bound the *labels* those icons sit beside. Where both
+/// apply the tighter of the two wins, so a subtree capped below the icon
+/// bound scales its icons to the subtree's cap.
+library;
+
+/// Rows of fixed height whose width is shared between a label and
+/// fixed-size chrome: the bottom nav, content-filter rows, the trending
+/// hashtag strip, category tiles.
+///
+/// These have the least room — the height is set by the bar, not the text,
+/// so growth spills rather than pushing the row taller.
+const fixedRowTextScaleLimit = 1.25;
+
+/// Columns of icon-with-caption buttons overlaid on video.
+///
+/// Matches `DivineIcon.maxScaleFactor` so the caption and the icon above it
+/// stop growing together; past that the column starts overlapping the
+/// video's own controls.
+const overlayActionColumnTextScaleLimit = 1.3;
+
+/// Single-line labels inside a pill or chip that can grow somewhat before
+/// the surrounding controls collide.
+const chipLabelTextScaleLimit = 1.4;
+
+/// Numeric stats in a divided row.
+///
+/// The loosest bound here: the digits carry the meaning, each column is
+/// `Expanded` so growth is shared, and the label below fades rather than
+/// overflowing.
+const statColumnTextScaleLimit = 1.5;
--- a/mobile/lib/screens/auth/nostr_connect_screen.dart
+++ b/mobile/lib/screens/auth/nostr_connect_screen.dart
@@ -696,28 +696,25 @@ class _ActionBar extends StatelessWidget {
               spacing: 10,
               children: [
                 _ActionButton(
-                  icon: DivineIcon(
+                  icon: const DivineIcon(
                     icon: DivineIconName.linkSimple,
                     color: VineTheme.vineGreen,
-                    size: MediaQuery.textScalerOf(context).scale(24),
                   ),
                   label: context.l10n.authCopyUrl,
                   onTap: isLoading ? null : onCopyUrl,
                 ),
                 _ActionButton(
-                  icon: DivineIcon(
+                  icon: const DivineIcon(
                     icon: DivineIconName.shareFat,
                     color: VineTheme.vineGreen,
-                    size: MediaQuery.textScalerOf(context).scale(24),
                   ),
                   label: context.l10n.authShare,
                   onTap: isLoading ? null : onShareUrl,
                 ),
                 _ActionButton(
-                  icon: DivineIcon(
+                  icon: const DivineIcon(
                     icon: DivineIconName.plus,
                     color: VineTheme.vineGreen,
-                    size: MediaQuery.textScalerOf(context).scale(24),
                   ),
                   label: context.l10n.authAddBunker,
                   onTap: isLoading ? null : onAddBunker,
@@ -954,9 +951,9 @@ class _ErrorContent extends StatelessWidget {
                       ),
                       elevation: 0,
                     ),
-                    icon: DivineIcon(
+                    icon: const DivineIcon(
                       icon: DivineIconName.arrowClockwise,
-                      size: MediaQuery.textScalerOf(context).scale(16),
+                      size: 16,
                       color: VineTheme.onPrimary,
                     ),
                     label: Text(
--- a/mobile/lib/screens/comments/comments_screen.dart
+++ b/mobile/lib/screens/comments/comments_screen.dart
@@ -116,15 +116,20 @@ class _CommentsTitle extends StatelessWidget {
 
         return Row(
           mainAxisSize: MainAxisSize.min,
+          spacing: 8,
           children: [
-            Text(
-              context.l10n.commentsHeaderCount(count),
-              style: VineTheme.titleMediumFont(
-                color: context.vineColors.onSurface,
+            Flexible(
+              child: Text(
+                context.l10n.commentsHeaderCount(count),
+                style: VineTheme.titleMediumFont(
+                  color: context.vineColors.onSurface,
+                ),
+                maxLines: 1,
+                softWrap: false,
+                overflow: .fade,
               ),
             ),
-            if (state.newCommentCount > 0) ...[
-              const SizedBox(width: 8),
+            if (state.newCommentCount > 0)
               NewCommentsPill(
                 count: state.newCommentCount,
                 onTap: () {
@@ -134,7 +139,6 @@ class _CommentsTitle extends StatelessWidget {
                   );
                 },
               ),
-            ],
           ],
         );
       },
--- a/mobile/lib/screens/comments/widgets/comment_input.dart
+++ b/mobile/lib/screens/comments/widgets/comment_input.dart
@@ -222,27 +222,30 @@ class _CommentInputState extends State<CommentInput> {
                 crossAxisAlignment: CrossAxisAlignment.end,
                 children: [
                   Expanded(
-                    child: Column(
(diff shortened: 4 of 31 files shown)


rabble (approved):
## Verdict: Approve

High-quality a11y fix. Well-scoped, well-documented, well-tested. No blocking issues.

### What I verified
- `gh pr diff` — 31 files, +780/-372. 5 commits on top of `main` (behind by ~9 commits but `git merge-tree` clean, CI green — see rebase note).
- Read all changed files via `git show origin/fix/font-size:*` — scaling contract consistent, clamp limits named, double-scale trap fixed in 9 call sites + `OgVinerBadge`.
- `dart analyze mobile/packages/divine_ui` — **No issues found**.
- `dart analyze` on six sensitive files (text_scale_limits, divine_icon, og_viner_badge, audio_chip, video_feed_item, vine_bottom_nav) — **No issues found**.
- `flutter test` in `mobile/packages/divine_ui` — **all 853+ tests pass** (full suite +845; targeted runs of divine_icon_test, divine_button_test, divine_icon_button_test, vine_bottom_sheet_action_menu_test green).
- `flutter test mobile/test/widgets/video_editor/audio_editor/video_editor_audio_chip_test.dart` — 14 tests green (new waveform clamp tests pass in CI shards).
- CI on PR: 16 checks **all pass** (Analyze, Format, Tests 0-4, Build Web Preview, Generated Files, etc.).

### Correctness — scaling contract
`DivineIcon.maxScaleFactor = 1.3` + `DivineIcon.scaleSize(context, size)` using `MediaQuery.textScalerOf(context).clamp(maxScaleFactor: 1.3).scale(size)` is the single contract. Good choices:
- Capped at 1.3 — uncapped 24 at 2.0x = 48px overflow matches 320pt at 2.0x ratchet in PR notes. Matches Samsung 2.0x manual check.
- `scaleSize` uses `TextScaler.clamp` so a tighter ambient `withClampedTextScaling` wins — verified in divine_icon_test (outer 48*1.3, inner 48*1.1). Prevents square-factor double scale.
- 9 call sites dropped pre-scale `MediaQuery.textScalerOf(...).scale(...)` + `OgVinerBadge` uses `textScaler: noScaling` with `dimension*0.85` + `FittedBox`. Correct fix for size*0.85 then scale again bug.
- `DivineIcon.build` now owns scaling (`dimension = scaleSize(context,size)`). `const DivineIcon` still valid — scaling is runtime.
- `DivineIconButton` both branches (SvgIconSource / MaterialIconSource) land on same `scaleSize(24)`. Small tap target 48 -> `scaleSize(48)` = 62.4 at cap (above min 48).
- `DivineButton` spinner `SizedBox.square(scaleSize(_iconSize))` keeps button size stable between icon/spinner.
- Settings/legal/nostr star/swatch/action-menu leading icon all via `scaleSize(24/20)` — consistent.
- `_VideoReplyButton` 40 -> `scaleSize(40)` — not reliant on outer clamp, fine.
- `_AudioBar` width/height via `scaleSize` and `_AudioBars` living below chip clamp is the key fix (would be 32px vs 22.4px otherwise). Spacing `scaleSize(1.5)` intentional.

### Clamp architecture
`lib/constants/text_scale_limits.dart` — 4 named constants encode constraints:
- `fixedRowTextScaleLimit=1.25` -> bottom nav, content filters, trending hashtags, category tiles, comment input. Least room, height fixed.
- `overlayActionColumnTextScaleLimit=1.3` == `DivineIcon.maxScaleFactor` -> captions/icons stop together, avoids overlapping video controls.
- `chipLabelTextScaleLimit=1.4` -> pill/chip labels (audio chip, main-actions sheet).
- `statColumnTextScaleLimit=1.5` -> divided stats row, Expanded + fade. Loosest.

Strictly better than `TextScaler.noScaling` — honors preference within layout bounds. Tightest-wins documented.

### Layout fixes
- `comments_screen`: Flexible + softWrap:false + fade + spacing:8 fixes overflow while keeping pill visible.
- `profile_stats_row_widget`: clamp 1.5 + Padding 4px + softWrap:false/fade — prevents label hitting divider.
- `categories_tab`: fixedRow clamp + VineTheme headlineSmall/labelMedium vs raw TextStyle — now uses design tokens + correct families (Bricolage Grotesque / Inter). `raw_textstyle.txt` -1 entry checks out.
- `trending_hashtags`/`content_filters`/`vine_bottom_nav`/`video_feed_item` wrappers around tight row/column only — minimal blast radius.
- `video_action_button` softWrap:false keeps ellipsis on width (was wrapping then bottom fade). Keeps ellipsis per last commit.
- `profile_screen_router`: VineBottomSheetActionMenu vs two InkWell rows — 61->17 lines, single path, leading icon now scaled (was hard-coded 24 box).
- `og_viner_badge`: Padding->Container(margin) visually equivalent; FittedBox guard remains.

### Tests
10 new tests pin contract:
- divine_icon_test 4: grows at 1.2, caps at 1.3, width==height, scaleSize caps and tighter clamp wins.
- divine_icon_button_test 2: all sources same size at 2x, small tap target capped.
- divine_button_test 1: spinner == icon box.
- vine_bottom_sheet_action_menu_test 1: leading SizedBox.square(scaleSize(24)) at 2x.
- video_editor_audio_chip_test 2: tallest bar 16*1.3 inside clamp, 16*1.1 under clamp.
All green locally and in CI.

### Nits / suggestions (non-blocking)
- Rebase before merge — 9 commits behind origin/main. video_feed_item.dart also touched in #6794 on main; merge-tree clean but rebase makes history linear.
- `text_scale_limits.dart` bare `library;` — valid Dart 3 but `library text_scale_limits;` greppability optional.
- _AudioBars spacing 1.5->1.95 at cap — verify visually chip width at 2x (chip clamp 1.4 vs icon cap 1.3 divergence not user-visible).
- No remaining double-scale: grep textScalerOf after PR shows only legitimate text sizing + divine_icon.dart itself.
- dart format passed — CI Format step green.

### Risk
Low. Isolated to scaling/clamp wrappers. No data-layer changes. Visual fallback capped. Worst case is a missed raw Icon elsewhere staying unscaled — not a regression.

Please rebase and merge.

rabble (approved):
Rabble review: a11y font scaling — DivineIcon now respects TextScaler, 9 double-scale fixes + 8 clamp wrappers, category/profile/bottom-nav overlays tested. All 16 checks pass. LGTM.

PR #6846, opened 2026-08-07
fix(notifications): tighten preview tokenizer cleanup

NotThatKindOfDrLiz:
## Description

Follow-up to merged #6839 with the two small review cleanups that did not land before the PR was merged:

- Document that the deliberate @mention tokenizer divergence also covers @<64-hex>, not only @npub1...
- Stop scanning tokenizer matches once match.start is at or beyond the preview limit, since later matches cannot affect the cut

This intentionally does not consolidate the shared tokenizer regex; that is larger than the cleanup and should be handled separately if we want to centralize parsing.

**Related Issue:** Relates to #6762, #6346

## Verification

- flutter test in mobile/packages/notification_repository — 172 pass
- flutter analyze lib test integration_test from mobile — no issues

### DIFF
--- a/mobile/packages/notification_repository/lib/src/notification_repository.dart
+++ b/mobile/packages/notification_repository/lib/src/notification_repository.dart
@@ -2408,9 +2408,10 @@ class NotificationRepository {
           match.group(_bech32ReferenceGroup) != null ||
           match.group(_hexReferenceGroup) != null;
       if (!isReference) continue;
-      final startsBeforeOrAtLimit = match.start < limit;
-      final endsAfterLimit = match.end > limit;
-      if (!startsBeforeOrAtLimit || !endsAfterLimit) continue;
+      // The `break` above already guarantees `match.start < limit`, so
+      // ending after it is exactly the straddling case.
+      final straddlesLimit = match.end > limit;
+      if (!straddlesLimit) continue;
       final canKeepLeadingToken =
           match.end <= maxReferencePreviewLength &&
           _hasOnlyWhitespaceOrPunctuationBefore(content, match.start);
--- a/mobile/packages/notification_repository/lib/src/notification_repository.dart
+++ b/mobile/packages/notification_repository/lib/src/notification_repository.dart
@@ -65,9 +65,10 @@ final RegExp _npubIdentifierPattern = RegExp(
 ///
 /// The UI's trailing `@([a-zA-Z][a-zA-Z0-9_]{0,30})` alternative is the one
 /// deliberate divergence. It wins over the bech32 alternative for `@npub1…`
-/// and truncates the token to 31 characters, so mirroring it would drop the
-/// protection #6346 added for that input. Whether the UI should tokenize
-/// `@npub1…` as a mention at all is a separate question from this cut.
+/// and over hex for `@<64-hex>`, so mirroring it would drop the protection
+/// #6346 added for `@npub1…` and keep `@<64-hex>` capped at the mention
+/// length. Whether the UI should tokenize those as mentions at all is a
+/// separate question from this cut.
 final RegExp _uiTokenPattern = RegExp(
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz:
#6839 merged while I was applying the requested small cleanup, so I carried the two bounded follow-ups here instead.

This PR only:
- documents the @<64-hex> tokenizer divergence alongside @npub1...
- adds the early exit once token matches start at or beyond the preview limit

Local verification: notification_repository tests pass (172), and full mobile analyze is clean. CI is green. @realmeylisdev, handing this back for one final look and merge if it matches your intent.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -64,10 +64,14 @@ final RegExp _npubIdentifierPattern = RegExp(
...
-/// protection #6346 added for that input. Whether the UI should tokenize
-/// `@npub1…` as a mention at all is a separate question from this cut.
+/// deliberate divergence, and it only fires when a letter follows the `@`.
+/// It therefore always wins for `@npub1…` — every bech32 prefix starts with
+/// `n` — and wins over hex only for a letter-initial `@<64-hex>`, capping
+/// either at 31 characters. A digit-initial `@<64-hex>` starts with no
+/// letter to consume, so it falls through to the hex alternative in both
+/// patterns and does not diverge at all. Mirroring the alternative would
+/// drop the protection #6346 added for `@npub1…`. Whether the UI should
+/// tokenize these as mentions at all is a separate question from this cut.
```
realmeylisdev, line 74: this had the mention alternative winning over hex for any `@<64-hex>`, but the alternative opens with `[a-zA-Z]`, so it needs a letter directly after the `@`.

ran both patterns over the two shapes:

```
@a1b2…  UI  0..32  g5="a1b2c3d4e5f60718293a4b5c6d7e8f9"   (31-char mention)
        mirror  1..65  g4=<full 64-hex>
@01b2…  UI  1..65  g4=<full 64-hex>
        mirror  1..65  g4=<full 64-hex>
```

so a digit-initial hex has no letter to consume, falls through to the hex alternative on both sides, and there is no divergence to document for it. that is 10 of the 16 possible first nibbles, so it was the majority shape rather than an edge case. narrowed to letter-initial in 92eeed5, and 752e078 pins both directions in `linkified_text_span_builder_test.dart` so the comment stops being the only thing asserting this.

mobile/packages/notification_repository/lib/src/notification_repository.dart
```diff
@@ -2399,13 +2403,15 @@ class NotificationRepository {
...
           match.group(_bech32ReferenceGroup) != null ||
           match.group(_hexReferenceGroup) != null;
       if (!isReference) continue;
-      final startsBeforeOrAtLimit = match.start < limit;
-      final endsAfterLimit = match.end > limit;
-      if (!startsBeforeOrAtLimit || !endsAfterLimit) continue;
+      // The `break` above already guarantees `match.start < limit`, so
+      // ending after it is exactly the straddling case.
+      final straddlesLimit = match.end > limit;
+      if (!straddlesLimit) continue;
```
realmeylisdev, line 2414: with the new `break` above, `match.start < limit` holds for every iteration that reaches this point, so `startsBeforeOrAtLimit` could never be false and that half of the condition was unreachable. the name also read `BeforeOrAt` while testing `<`.

left it as the straddle check it actually is, with a note so the start test does not get reinstated later. fuzzed the pre-break version, the break version, and this simplified one over 200k generated contents (hex, bech32, `nostr:`, blossom URLs, hashtags, `@`-prefixed variants, Cyrillic and CJK lead-ins) at limits 1/2/13/49/50/51/200: all three agree on every case, 0 mismatches. 8fccf73.

realmeylisdev (approved):
took this over instead of handing it back. the `@<64-hex>` half only holds when the hex starts with a-f, since the mention alternative opens with `[a-zA-Z]`; a digit-initial one falls through to the hex branch in both patterns, so nothing diverges there. narrowed that, dropped the start check the new break made dead, and pinned the claim with a linkified_text test. shard 0 failed once on an unrelated notification row overflow that doesn't reproduce locally even at the same seed, re-ran it green.

PR #6851, opened 2026-08-07
fix(for-you): rebuild the algorithm header and explainer on the design system

### LINKED ISSUE #6852: fix(for-you): rebuild the algorithm header and explainer on the design system
Auto-created from PR #6851.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6851
**Branch:** `fix/algo-sheet-design`

hm21:
Closes #6852

## Description

The For You header and its algorithm explainer sheet were the last hand-rolled corner of the discovery tab: a bare `DivineIcon` as the info affordance, a hand-built `DraggableScrollableSheet` with its own drag handle, thirteen inline `TextStyle`s and four raw Material icons. None of that picks up theme or accessibility changes coming out of `divine_ui`.

**Sheet → `VineBottomSheet`.** `VineBottomSheet` owns the drag handle, sizing and surface, so `_AlgorithmExplainerSheet` is reduced to content plus the scroll controller handed in by the host. The hand-drawn handle is gone. The `0.85 / 0.5 / 0.95` sizing is kept deliberately and passed explicitly, because `VineBottomSheet.show` defaults to `0.6 / 0.3 / 0.9` and this is long-form content that should open nearly full-height.

**Info icon → `DivineIconButton`.** Previously a 16px glyph with no semantics inside a bare `GestureDetector` — invisible to screen readers as an action. It now carries a 48px tap target and a tooltip. The semantics ended up on the header row rather than on the button: review pointed out the row was still a silent tap region wrapped around a labelled button, so the row now owns `Semantics(button: true, hint: ...)` and the button is wrapped in `ExcludeSemantics` — one action announced once, named by its visible title. The hint reuses the existing `forYouAlgorithmHowItWorksTitle` key rather than adding a new one that would need 17 locale translations. The whole header row stays tappable as before, so the discoverability of the explainer does not shrink.

**Inline styles → `VineTheme`.** All thirteen raw `TextStyle`s map onto the closest point of the design-system scale. Where no exact match existed I rounded to the scale rather than adding a new `VineTheme` helper for a one-off, so a few tokens shift: section titles 17 → 16, state titles 18 → 16, secondary text 13 → 12. Weight and family move with them — section titles w600 → w800, state titles w500 → w800, interaction labels w500 → w600, and the title styles into Bricolage, the app's title family (the sheet title keeps 22px but goes w700 → w800). Body line-height goes 1.5 → 1.43. The ad-hoc values were never design tokens to begin with.

**Material icons → `DivineIcon`.** The four interaction-weight icons become `repeat` / `chat` / `heart` / `playCircle`. `chat` over `chatCircle` because it is the more common variant elsewhere in the app (5 call sites vs 2).

Both ceiling baselines are regenerated: the file drops to zero raw `TextStyle`s and zero raw sheet calls, and the ratchets require a file falling to zero to be locked in explicitly. Only the `for_you_tab.dart` line is removed from each baseline — a full regeneration would have swept in unrelated drift from three other files.

**Related Issue:** 

## Out of Scope

- `Icons.cloud_off` and `Icons.error` in the unavailable / error states still use raw Material icons. Converting them would drop the file to zero raw icons and pull the `raw_icons` baseline into this diff for two 64px placeholder glyphs that carry no design intent.
- `DivineButton` has no `foregroundColor` override, so the whole header could not become a single `DivineButton` without either losing the green title or restyling the strip into a chip. Left as is deliberately.

## Verification

- `dart format` — clean
- Design-system ceilings: `raw_textstyle`, `raw_colors`, `material_button`, `raw_icons` green. `raw_dialog` now fails on the un-merged branch because `main` has since shrunk `composable_video_grid.dart` from 2 to 1 — base staleness on a file this PR does not touch, not this diff. It is green on the merge, which is what CI runs.
- `mobile/test/widgets/for_you_tab_test.dart` pins the header's accessibility contract: one button named by its visible title, and tapping it opens the explainer
- Device verification on a real Samsung Galaxy still pending; holding merge on it

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/widgets/for_you_tab.dart
+++ b/mobile/lib/widgets/for_you_tab.dart
@@ -153,14 +153,11 @@ class _ForYouContent extends ConsumerStatefulWidget {
 class _ForYouContentState extends ConsumerState<_ForYouContent>
     with ScrollToHideMixin {
   void _showAlgorithmExplainer(BuildContext context) {
-    showModalBottomSheet<void>(
+    VineBottomSheet.show(
       context: context,
-      isScrollControlled: true,
-      backgroundColor: context.vineColors.card,
-      shape: const RoundedRectangleBorder(
-        borderRadius: BorderRadius.vertical(top: Radius.circular(20)),
+      buildScrollBody: (scrollController) => _AlgorithmExplainerSheet(
+        scrollController: scrollController,
       ),
-      builder: (context) => const _AlgorithmExplainerSheet(),
     );
   }
 
@@ -240,32 +237,37 @@ class _ForYouContentState extends ConsumerState<_ForYouContent>
           child: GestureDetector(
             key: headerKey,
             onTap: () => _showAlgorithmExplainer(context),
-            child: Container(
-              padding: const EdgeInsets.symmetric(horizontal: 16, vertical: 12),
+            child: ColoredBox(
               color: context.vineColors.background,
-              child: Row(
-                children: [
-                  const DivineIcon(
-                    icon: DivineIconName.sparkle,
-                    color: VineTheme.vineGreen,
-                    size: 20,
-                  ),
-                  const SizedBox(width: 8),
-                  Text(
-                    context.l10n.forYouAlgorithmTitle,
-                    style: const TextStyle(
+              child: Padding(
+                padding: const EdgeInsets.symmetric(horizontal: 16),
+                child: Row(
+                  children: [
+                    const DivineIcon(
+                      icon: DivineIconName.sparkle,
                       color: VineTheme.vineGreen,
-                      fontSize: 14,
-                      fontWeight: FontWeight.w600,
+                      size: 20,
                     ),
-                  ),
-                  const SizedBox(width: 4),
-                  DivineIcon(
-                    icon: DivineIconName.info,
-                    color: context.vineColors.secondaryText,
-                    size: 16,
-                  ),
-                ],
+                    const SizedBox(width: 8),
+                    Text(
+                      context.l10n.forYouAlgorithmTitle,
+                      style: VineTheme.labelLargeFont(
+                        color: VineTheme.vineGreen,
+                      ),
+                    ),
+                    DivineIconButton(
+                      icon: DivineIconName.info,
+                      size: DivineIconButtonSize.small,
+                      backgroundColor: VineTheme.transparent,
+                      foregroundColor: context.vineColors.secondaryText,
+                      showShadow: false,
+                      tooltip: context.l10n.forYouAlgorithmHowItWorksTitle,
+                      semanticLabel:
+                          context.l10n.forYouAlgorithmHowItWorksTitle,
+                      onPressed: () => _showAlgorithmExplainer(context),
+                    ),
+                  ],
+                ),
               ),
             ),
           ),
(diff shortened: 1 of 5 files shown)


rabble (dismissed):
The migration itself is sound — I checked `VineBottomSheet.show` against this call and the `buildScrollBody` + `initialChildSize/minChildSize/maxChildSize` path is exactly right, the asserts are satisfied (`scrollable` defaults true), and passing `title: null` collapses `_HeaderRow` to zero height rather than leaving an empty strip, so the chrome ends up ~equivalent to the old hand-rolled handle. Both baseline edits are correct and minimal. The `labelLargeFont` mapping for the header title (Inter 14/w600) is an exact match for what it replaced.

Four things before this merges, one of which looks accidental.

### 1. The header sparkle silently grows 20px → 24px

```dart
// before
const DivineIcon(icon: .sparkle, color: VineTheme.vineGreen, size: 20),
// after
const DivineIcon(icon: .sparkle, color: VineTheme.vineGreen),
```

`DivineIcon.size` defaults to `24` (`divine_ui/lib/src/icon/divine_icon.dart:249`), so dropping the parameter is a 20% size increase on the most prominent glyph in the discovery header. The description enumerates every other intentional size shift (17→16, 13→12, 18→16) and doesn't mention this one, and the sheet's sparkle deliberately kept `size: 28` — so I read this as collateral from removing `size: 16` off the info icon next to it. If it was intentional, worth saying so; otherwise it's a one-line restore.

### 2. `Row` can now overflow where it previously fit

The title `Text` has no `Expanded`/`Flexible`, and the trailing affordance went from a 16px icon to a 48px tap target. On a 320dp screen the text budget drops from ~240px to ~208px. `forYouAlgorithmTitle` is translated into 17 locales, and `accessibility.md` explicitly asks for font-scale resilience — a long translation at a large text scale is a plausible `RenderFlex overflowed` here, and it wasn't reachable before. Wrapping the `Text` in `Flexible` costs nothing and removes the class of bug entirely.

### 3. The outer `GestureDetector` is still an unlabelled tap target

The info button now has `button: true` semantics, a tooltip and a 48px target — good. But the whole row is still wrapped in a bare `GestureDetector` with no `Semantics`, which `accessibility.md` lists under **Must Fix** ("`GestureDetector` or `InkWell` without `Semantics` wrapper"). The result is a screen reader hearing one labelled button sitting inside a silent tap region that does the same thing. Since this PR is specifically the a11y/design-system pass on this widget, it's the natural place to close it — either wrap the row in `Semantics(button: true, label: ...)`, or drop the outer `GestureDetector` now that a real button exists and let the button own the interaction.

### 4. The state/section titles change weight, not just size

`titleMediumFont` is 16/**w800**. The section titles were 17/w600 and the empty/unavailable/error state titles were 18/**w500**. The description frames these as size changes ("state titles 18 → 16"), but w500 → w800 is the larger visual delta of the two. Not wrong — it is the design scale — but it's worth naming, because combined with #1 and the removed `vertical: 12` padding the header/state chrome shifts more than the description implies. Which leads to:

### Verification

The description says device verification on a Samsung Galaxy is still pending and that the PR would stay draft until that pass is done, but it's open for review and not a draft. Given four separate size/weight/padding shifts land in this diff, that device pass is the thing I'd most want done before merge. Has it happened?

### Smaller notes, non-blocking

- `showHeader: false` would express the intent more directly than relying on the null-title collapse. Same rendered result (the headerless path draws a 1px divider vs the header's 2px), so purely a readability call.
- `_buildSectionTitle` / `_buildInteractionItem` / `_buildFutureFeatureItem` are still methods returning `Widget`, which `code_style.md` rules out. Pre-existing and explicitly optional cleanup under the debt policy — mentioning only because this PR touches all three.
- No test lands with this. The a11y win in particular (tooltip, `semanticLabel`, 48px target) is exactly what a small widget test pins down cheaply, and there's currently nothing stopping the next refactor from dropping it again.

Happy to re-review quickly once 1–3 are addressed — the bulk of this is a clean, well-argued migration and the baseline handling is exactly right.

hm21:
Pushed e8830d61c (fixes) and e928abaf3 (test). Checks green.

**1. Sparkle 20 → 24.** Fixed — restored to `size: 20`. Collateral, as you suspected, but from a follow-up commit rather than the original migration: it dropped the override to match the icon button's 24px glyph. Your read that it doesn't belong in the intended scale changes is right.

**2. Row overflow.** Fixed — title wrapped in `Flexible`.

**3. Unlabelled `GestureDetector`.** Fixed, taking your first option: the row keeps its tap area and now carries the semantics, and the icon button's are excluded so one action announces once rather than twice. One deviation from the literal suggestion — a `label` on the outer `Semantics` merges with the child `Text` and announces `"How It Works\nFor You"`, so the explainer went into `hint` instead. The node resolves to `label: "The Divine Algorithm"`, `hint: "How It Works"`, `isButton: true`, so the name matches the visible title.

**4. Weight, not just size.** Correct, and the description understated it: section titles w600 → w800 and state titles w500 → w800, both moving into Bricolage, alongside the 17 → 16 and 18 → 16 size shifts. Keeping them — it is the scale — but naming it properly here.

**Verification.** No, the Galaxy pass has not happened yet, and you're right that the PR should not have left draft ahead of it. Holding merge until that pass is done — it's the last open item, not the review.

**Smaller notes.**

- `showHeader: false` — declined. By your own note it renders a 1px divider where the header renders 2px, which makes it a visual change rather than a pure readability one, and this diff already carries more unverified visual delta than I want to add to.
- Widget-returning helpers — fixed, now `_SectionTitle` / `_InteractionItem` / `_FutureFeatureItem`.
- Test — added `mobile/test/widgets/for_you_tab_test.dart`. It drives the real `ForYouTab` over stubbed availability and an empty feed, and pins the header to a single button with that label and hint plus the tap opening the sheet. Removing the `ExcludeSemantics` turns it red, so it holds the a11y win in place.

One thing you'll hit if you pull this locally: the branch is 20 commits behind `main`, and `main` has since dropped `categories_tab.dart` from the raw-TextStyle baseline, so `check_raw_textstyle_ceiling.sh` reports an added key on the un-merged branch. It's base staleness, not this diff — clean on a throwaway merge with `main`, and CI agrees. Left unrebased since the update is only review feedback and GitHub reports no conflicts.

mobile/scripts/baseline/raw_dialogs.txt
```diff
@@ -25,7 +25,6 @@ lib/utils/pause_aware_modals.dart	2
 lib/widgets/add_to_list_dialog.dart	1
 lib/widgets/age_verification_dialog.dart	1
 lib/widgets/composable_video_grid.dart	2
```
dcadenas: **The branch cannot merge, and this is the only conflicting line.**

`main` has since shrunk `composable_video_grid.dart` from 2 to 1, on the line next to the `for_you_tab.dart` entry this PR removes. `git merge-tree` against current `main` conflicts on this file and nothing else.

`check_raw_dialog_ceiling.sh` exits 1 locally for the same reason, reporting `composable_video_grid.dart 1 -> 2`. That is base staleness on a file this PR does not touch.

Resolving it means keeping both shrinks: `composable_video_grid.dart` at 1, and no `for_you_tab.dart` line at all.

mobile/lib/widgets/for_you_tab.dart
```diff
@@ -277,226 +294,187 @@ class _ForYouContentState extends ConsumerState<_ForYouContent>
...
-              ),
-              const SizedBox(height: 16),
-              Text(
-                context.l10n.forYouAlgorithmChoiceClosing,
-                style: _bodyTextStyleOf(context),
-              ),
-              const SizedBox(height: 24),
+        // Interaction weights
+        _InteractionItem(
+          icon: DivineIconName.repeat,
```
dcadenas, line 351: **The raw-icons win is not locked in.**

Converting these four to `DivineIcon` takes the file from 6 code-only `Icons.*` to 2, but `mobile/scripts/baseline/raw_icons.txt:27` still reads 6. The ratchet only fails on growth, so CI stays green and all four could come back without anything noticing.

AGENTS.md asks for the baseline to be regenerated after an intentional reduction. Editing the single line keeps the diff clean, the same way the other two baselines here are handled — set that entry to:

```
lib/widgets/for_you_tab.dart	2
```

The remaining 2 are the `Icons.cloud_off` / `Icons.error` placeholders the description already puts out of scope.

mobile/test/widgets/for_you_tab_test.dart
```diff
@@ -0,0 +1,86 @@
...
+      // The subtitle renders only inside the explainer sheet, so it proves the
+      // sheet opened instead of matching something in the header strip.
+      expect(find.text(l10n.forYouAlgorithmSubtitle), findsNothing);
+
+      await tester.tap(find.text(l10n.forYouAlgorithmTitle));
+      await tester.pumpAndSettle();
+
+      expect(find.text(l10n.forYouAlgorithmSubtitle), findsOneWidget);
+    });
+  });
```
dcadenas, line 120: **Nothing pins the `Flexible` that fixed the header overflow.**

Removing it drives the info button's right edge to 545px on a 320dp screen at the largest text scale, and every test here still passes.

Two things worth keeping if you take this. The largest scale is deliberate: at the default scale the title's measured width depends on which font resolves in the test environment, so that case reports on font loading rather than on layout. Asserting geometry rather than a caught overflow exception keeps it independent of the debug-only overflow reporter.

```dart
// _host gains a text-scale knob:
Widget _host({double textScale = 1}) => ProviderScope(
  overrides: [
    funnelcakeAvailableProvider.overrideWith(_AvailableFunnelcake.new),
    forYouFeedProvider.overrideWith(_EmptyForYouFeed.new),
  ],
  child: MaterialApp(
    localizationsDelegates: AppLocalizations.localizationsDelegates,
    supportedLocales: AppLocalizations.supportedLocales,
    home: Builder(
      builder: (context) => MediaQuery(
        data: MediaQuery.of(
          context,
        ).copyWith(textScaler: TextScaler.linear(textScale)),
        child: const Scaffold(body: ForYouTab()),
      ),
    ),
  ),
);

/// Narrowest screen the app targets, where the header has the least room.
const _narrowWidth = 320.0;

void _useNarrowScreen(WidgetTester tester) {
  tester.view.physicalSize = const Size(_narrowWidth, 640);
  tester.view.devicePixelRatio = 1;
  addTearDown(tester.view.resetPhysicalSize);
  addTearDown(tester.view.resetDevicePixelRatio);
}

// and the case itself, inside group(ForYouTab, ...):
testWidgets('header keeps the info button on screen at the largest text '
    'scale', (tester) async {
  _useNarrowScreen(tester);

  await tester.pumpWidget(_host(textScale: 3));
  await tester.pumpAndSettle();

  expect(
    tester.getRect(find.byType(DivineIconButton)).right,
    lessThanOrEqualTo(_narrowWidth),
  );
});
```

dcadenas (approved):
✅ the design-system migration holds up under a close read, and the four items from the earlier review are addressed.

Three non-blocking suggestions inline. Two of them are small patches I would have pushed for you, but the branch conflicts with `main` and the pre-push hook refuses any push until that is resolved — so they are written out to apply directly.

mobile/scripts/baseline/raw_dialogs.txt
hm21: Fixed in 238886de4b. Resolved by rebuilding the branch on fresh `main` and keeping both shrinks — `composable_video_grid.dart` at 1, no `for_you_tab.dart` line. `check_raw_dialog_ceiling.sh` is green now, as are the other four ceilings. The branch also carried a `Merge branch 'main'` commit, which the repo rules rule out, so the four commits were replayed onto `origin/main` instead; file contents are byte-identical to the old tip apart from this one line.

mobile/lib/widgets/for_you_tab.dart
hm21: Fixed in f5bd68e486, set to `2`. Confirmed the count with the repo's own detector rather than by eye — `dart_code_only.awk` over the file leaves exactly `Icons.cloud_off` and `Icons.error`.

Editing the single line was the right call for the reason you gave: a full `UPDATE_BASELINE=1` regeneration also drops `creator_analytics_screen.dart` 5 -> 3, `sound_detail_screen.dart` 4 -> 2 and `sounds_tab.dart` 3 -> 2, none of which this PR touches.

mobile/test/widgets/for_you_tab_test.dart
hm21: Added in 266cf42e61, taken as written including the text-scale knob and the two notes as a comment on the test.

Checked it can actually fail before keeping it: dropping the `Flexible` puts the button's right edge at 545.1px against a 320px bound, which is your number exactly, while the other two tests stay green.

hm21:
Pushed f5bd68e486 (raw-icons baseline) and 266cf42e61 (overflow test), on a branch rebuilt on fresh `main`. All three of @dcadenas's notes taken as written, replied inline. Checks green.

The branch is now mergeable — the `raw_dialogs.txt` conflict is resolved by keeping both shrinks, and `check_raw_dialog_ceiling.sh` passes along with the other four ceilings. Worth noting the history changed shape: it carried a `Merge branch 'main'` commit, which the repo rules rule out, so the four commits were replayed onto `origin/main` rather than merged again. Contents are byte-identical to the old tip apart from that one baseline line.

Two things still open, neither in the review:

- @rabble this is still formally `CHANGES_REQUESTED` from your pass — items 1–4 landed back in e8830d61c1 / e928abaf3a (now fd5eb1c80c / 6898ba4216 after the replay). No rush, just flagging that the block is stale rather than outstanding.
- The Samsung Galaxy pass has still not happened. It is the last item before merge, unchanged from the earlier handback.

PR #6855, opened 2026-08-07
fix(report): carry the uncollapsed report reason on the moderation DM

### LINKED ISSUE #6593: fix(moderation): moderation DM intake is half-wired and its unread badge is dead code

realmeylisdev:
Closes #6593. Replaces #6831, which this supersedes after review feedback from @NotThatKindOfDrLiz.

## Summary

divine-moderation-service's `dm-reader.mjs` only ever promoted a report DM to a `user_reports` row when its content parsed as JSON `{type:'conversation_report', sha256}` — a shape this app has never produced. The DM is prose, so every moderation report landed as a generic chat message and the structured row was never written.

The DM's content is unchanged: `ReportSubmissionCubit` still produces the same human-readable prose, which is what NIP-17 mandates and what the admin Messages UI renders. The report data now travels as NIP-17 tags on the same kind-14 rumor, via `dm_repository.sendMessage`'s existing `additionalTags` parameter.

The report sheet's submission lifecycle now lives in `ReportSubmissionCubit`, keeping the queued moderation-DM row, delivered snapshot, terminal outcome, in-flight queueing, cancel/recover behavior, and retry/fallback decisions out of the widget layer.

## What changed since #6831

#6831 attached only the NIP-56 `report_type`. That vocabulary is lossy — it collapses `aiGenerated` / `childSafety` / `underageUser` / `falseInformation` to `other`, and `csam` / `copyright` / `violence` to `illegal` — while the kind-1984 relay poller recovers the granular reason from the NIP-32 `l` label. Since `user_reports` is written with `INSERT OR IGNORE` on `(sha256, reporter_pubkey)`, whichever channel ingested first would pin its value permanently, so a DM winning that race would have downgraded an AI report to `other` for good.

The rumor now carries the same `['L', ns]` + `['l', 'NS-<reason>', ns]` pair the kind-1984 report already publishes, so both channels resolve to one report type. The NIP-56 tag still ships as the fallback for when the label is absent.

Final rumor tags:

```
['p', <moderation pubkey>]                            // NIP-17
['L', 'social.nos.ontology']                          // NIP-32 namespace
['l', 'NS-aiGenerated', 'social.nos.ontology']        // granular, preferred
['report_type', 'other']                              // NIP-56, fallback
['sha256', <blob hash>]                               // content reports only
```

`sha256` rides along only for content reports whose video resolves a Blossom blob hash. User reports and DM-message reports have none, and `user_reports.sha256` is `NOT NULL`, so omitting it degrades to no report row rather than a malformed one.

Both mapping tables were lifted out of `ContentReportingService` into `content_moderation_types.dart` so the two report channels read one source and cannot drift.

## Review feedback addressed

| Thread | Resolution |
|---|---|
| `report_content_dialog.dart` — NIP-56 value is lossy for a Divine-private ingestion contract; add a granular tag and have the backend prefer it | Fixed. Rumor carries the NIP-32 `L`/`l` pair; the companion backend PR prefers it. Reusing the existing label rather than inventing a tag means the backend's `extractReportType` already decodes it. |
| `content_moderation_types_test.dart` — the coverage test only visits keys the table already names | Fixed. Both tables now assert `expected.keys.toSet() == ContentFilterReason.values.toSet()` before the loop; the separate non-empty test is gone. |
| `report_content_dialog.dart` — moderation-DM submission state machine belongs outside the widget | Fixed. `ReportSubmissionCubit` owns the submission lifecycle; the widget keeps form state and rendering. |
| `report_content_dialog.dart` — primary reporting service was captured at sheet-open | Fixed. The cubit now takes a `ContentReportingServiceResolver` and resolves the account-scoped service inside each submit. |
| `report_submission_cubit.dart` — blocked DM sends should not retry | Fixed. Blocked sends now move to a terminal `ModerationDmOutcome.blocked` and resubmits do not call `sendMessage` again. |

## Test plan

- [x] `flutter test test/blocs/report/report_submission_cubit_test.dart test/services/content_reporting_service_test.dart test/services/content_moderation_types_test.dart` — 58/58
- [x] `flutter test test/widgets/report_content_dialog_test.dart` — 58/58
- [x] Pre-push hook — analyze plus changed report tests, 116/116
- [x] Cross-repo: a real NIP-59 gift-wrap round trip of these exact tags into the real `processRumor` writes `report_type = 'ai_generated'`, not `'other'` (pinned in the companion PR)

## Related

- #6593
- divinevideo/divine-moderation-service#194 — reads these tags (must merge together; neither half works alone)

## Verification beyond unit tests

A full live loop, nothing stubbed: a real NIP-59 gift wrap carrying these exact tags, published to local_stack's relay, read by `wrangler dev` running the companion backend branch, into a real D1.

| table | field | value |
|---|---|---|
| `user_reports` | `report_type` | `ai_generated` — not the collapsed `other` |
| `dm_log` | `message_type` | `conversation_report` — not `creator_reply` |
| `dm_log` | `content` | the prose, unchanged |
| `moderation_results` | `raw_response.reportType` / `.source` | `ai_generated` / `dm-report` |


mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -619,6 +620,34 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
...
+  /// whichever channel ingested first pin a collapsed type — `other` for an
+  /// `aiGenerated` report — permanently, because `user_reports` is written
+  /// with `INSERT OR IGNORE` on `(sha256, reporter_pubkey)`.
+  ///
+  /// `sha256` is present only for content reports whose video resolves a
+  /// Blossom blob hash: [VideoEvent.sha256] is nullable, and user reports
+  /// and DM-message reports never have one. `user_reports.sha256` is
+  /// `NOT NULL`, so omitting it degrades to no report row rather than a
+  /// malformed one.
+  List<List<String>> _reportDmTags() {
```
NotThatKindOfDrLiz: Please move this NIP-32/NIP-56 moderation-DM tag construction out of the widget before merge. `_ReportContentDialogState` is presentation code, but this method now owns a cross-service wire contract (`L`/`l`, `report_type`, and `sha256` omission semantics). The repo architecture rule keeps business/protocol logic below UI, and `ContentReportingService` is already the place that builds the kind-1984 report tags from the same reason mappers. A small service-layer helper/method that returns the DM tags would keep both report channels sharing one contract and leave the widget only orchestrating submission.

NotThatKindOfDrLiz (changes requested):
Thanks for tightening the cross-service report type contract. The tag shape looks right, but this needs one architecture cleanup before merge: keep the widget out of the moderation DM wire-format construction and move that responsibility into the service/business layer, with the existing widget tests adjusted around the new seam.

mobile/lib/widgets/report_content_dialog.dart
realmeylisdev: Moved in d88f92c1f.

`_reportDmTags()` is gone. The dialog now calls
`ContentReportingService.moderationDmTags(reason:, sha256:)` and passes only what it
knows — the reason the user picked, and the video's blob hash. Both channels build the
`L`/`l` pair from one private `_nip32ReportLabelTags`, so `_createReportingEvent` and the
DM can no longer drift. Output is byte-identical on both.

`report_type` and `sha256` stay DM-only rather than joining the shared helper: kind-1984
carries the NIP-56 type as the third element of its `e`/`p` tags, and `extractReportType`'s
doc depends on the `report_type` arm staying inert for 1984 events.

I made it `static` rather than an instance method. Beyond it being a pure function of its
two arguments, two things pushed against the instance form:

- `_dispatchModerationDm` is deliberately synchronous up to its first
  `sendMessage`/`recoverFullSend` await — the offline path fires it unawaited, so the
  preflight must not touch `ref` on a possibly-disposed dialog. An instance method puts
  `await ref.read(contentReportingServiceProvider.future)` in front of that.
- the widget test overrides `ContentReportingService` with a mocktail mock at nine sites.
  As an instance method, the #6593 tag assertions would have been asserting the stub
  rather than the real mapping.

Tests: four of the five wire-format widget tests moved down to
`content_reporting_service_test.dart`. The per-reason kind-1984 test there now also
asserts the DM's first two tags equal the `L`/`l` rows that event actually carried — that
is the one that goes red if the two channels disagree. Two widget tests remain for what
only a widget test can catch: the dialog feeding the wrong reason, or inventing a blob
hash. I mutated each of the three cases (diverged label, hardcoded reason, invented hash)
and confirmed the intended test failed.

One thing I did not decide unilaterally: every other shared cross-path tag builder in this
repo is a top-level function beside its constants (`collaborator_tags.dart`,
`inspired_by_tags.dart`, `video_event_tag_source.dart`), which would put this in
`content_moderation_types.dart` next to the two mappers it composes. I put it on
`ContentReportingService` as you asked. Say the word if you'd rather follow that
precedent — it is a one-line move.

Separately: divinevideo/divine-moderation-service#194's `dm-report-contract.test.mjs:23`
comment names `_reportDmTags()`, which no longer exists. Backend behaviour is unaffected;
I will update that comment on #194.

Checks are green on d88f92c1f.

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -374,15 +381,20 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
   }
 
   Future<void> _submitReport() async {
-    if (_selectedReason == null) return;
-
+    final reason = _selectedReason;
```
dcadenas, line 389: The reason is captured here, before the first await, and this value is the only one that reaches `reportUser`/`reportContent`, the `details` prose, and the DM.

Previously each channel read `_selectedReason` on its own side of the round trip. `_isSubmitting` does not swap the form out — it only drives `_canSubmit`, the button's loading state and the re-entry guard — so the cards stay tappable while a submit is in flight, and a tap between the two reads gave one report two different NIP-32 labels. No parked row or connectivity failure needed; it is the ordinary success path.

Pinned by `both channels label one submit with the reason it started on`, which gates the reporting-service provider behind a completer so the tap lands inside the first await, and asserts the card actually moved before trusting the rest.

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -484,35 +496,91 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
...
+      // the new one — the exact cross-channel divergence this DM's tags exist
+      // to prevent, and permanent once `user_reports` pins the first write.
+      // Drop that row and mint a correct one instead.
+      //
+      // Best effort: if the sweep already delivered the superseded copy, the
+      // cancel is a no-op and the team ends up triaging two DMs while
+      // `user_reports` keeps the reason the delivered one pinned. That case is
+      // unrecoverable from here either way — re-driving it would pin the same
+      // wrong reason and skip the correction entirely.
+      final staleReason =
```
dcadenas: A parked row stores its rumor with the tags frozen at build time, and `recoverFullSend` replays that rumor verbatim. So re-driving one after the user picked a different reason shipped the superseded label while the kind-1984 republish carried the new one.

When the reason has moved, the row is cancelled and a correct DM is minted instead. A plain resubmit still coalesces onto the same row, so the one-row guarantee from #6610 holds.

The cancel is best effort. If the sweep already delivered the superseded copy, the cancel is a no-op, the team triages two DMs, and `user_reports` keeps the delivered copy's reason. Re-driving would pin that same reason and skip the correction entirely, so this is the better of the two.

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -484,35 +496,91 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
...
-      ),
-  };
+  ///
+  /// [reason] is the submit's own snapshot, not `_selectedReason`. Queuing
+  /// rather than joining is what makes that matter: an outstanding send
+  /// carries the label it was built with, so returning its result to a submit
+  /// the user has since re-aimed would report the superseded reason. Waiting
+  /// lets the re-aimed submit see the row the first one parked and replace it.
+  Future<bool> _sendModerationDm(ContentFilterReason reason) {
+    final outstanding = _moderationDmInFlight;
```
dcadenas: Queuing rather than joining, for the same reason as the staleness check above.

An outstanding send carries the label it was built with. Returning its future to a submit the user had since re-aimed reported the superseded reason and never ran the replacement pass at all. Waiting lets the queued submit see the row the first one parked and replace it.

The links are strictly serial, so two dispatches still cannot overlap, and `_runModerationDm` re-checks the terminal outcomes an earlier link may have reached while this one waited.

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -524,6 +592,14 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
           // never let them degrade to a metadata-leaking NIP-04
           // plaintext duplicate. NIP-17 gift wrap only.
           skipNip04Fallback: true,
+          additionalTags: ContentReportingService.moderationDmTags(
+            reason: dispatchReason,
+            // A user report targets the account, not one of its videos. The
+            // constructor allows `video` and `userPubkey` together, so gate on
+            // the routing decision rather than on `video` being present —
+            // otherwise the backend files the report against that blob.
+            sha256: _isUserReport ? null : widget.video?.sha256,
```
dcadenas: Gated on the routing decision rather than on `video` being non-null.

The constructor allows `video` and `userPubkey` together, and `userPubkey` is what routes to `reportUser`. Passing the hash regardless would have the backend file an account-level report against that one video. No current call site does this — the three `show*` factories are mutually exclusive — so it was latent rather than live.

mobile/lib/services/content_reporting_service.dart
```diff
@@ -432,6 +430,65 @@ class ContentReportingService {
...
+  /// not 64 hex characters is dropped, which makes the documented degrade
+  /// explicit rather than something the backend has to catch.
+  ///
+  /// Tag order is pinned by divine-moderation-service's
+  /// `dm-report-contract.test.mjs` fixture.
+  static List<List<String>> moderationDmTags({
+    required ContentFilterReason reason,
+    String? sha256,
+  }) {
+    final blobHash = normalizeSha256Hash(sha256);
```
dcadenas: `VideoEvent.sha256` is the `imeta` `x` tag verbatim and never validated.

A padded hash is a real hash that fails the backend's anchored check, so `relatedSha256` goes null and the report files nothing — silently. Trimming files it correctly. Case is canonicalised too, though not for that reason: the backend lower-cases before matching, so it is only that this is the identifier the report is keyed by.

Anything that is not 64 hex characters is dropped here, which makes the documented degrade explicit instead of something the backend has to catch.

dcadenas (approved):
The tag design is right, and the label/type split is the part that matters: reusing the NIP-32 pair the kind-1984 already publishes means the backend resolves both channels to one report type instead of racing them.

What the tags exposed is that the reason was never pinned for the length of a submit. The kind-1984 publish and the DM read the selection on opposite sides of a relay round trip, and the reason cards stay tappable across it. Before this change that was cosmetic, because the DM wrote no `user_reports` row. With it, the two channels can label one report differently, and `INSERT OR IGNORE` on `(sha256, reporter_pubkey)` makes whichever lands first permanent.

### Changes I pushed

- Commit one reason per submit and thread it into both channels.
- Replace a parked moderation DM instead of re-driving it when the reason changed.
- Queue a moderation DM behind an outstanding send rather than joining it.
- Keep a video blob hash off a user report's DM.
- Normalize the blob hash before it becomes a report identifier.
- Reverted `541df60` — see below.

Each fix is its own commit, and each new test was run against a mutation of the fix it guards to confirm it fails.

`541df60` (iOS Simulator `localHost`) is reverted. #6840 landed the same platform-resolving getter on `main` on 2026-08-09, with named constants, the LOCAL invite server, and wider test and doc coverage. Carrying the duplicate was the only thing making this branch conflict with `main`, and the description said the commit reverts cleanly on its own. I corrected that section of the description rather than leaving it describing code that is no longer here.

### Checks

- `dart format --output=none --set-exit-if-changed lib test integration_test` — 0 changed
- `very_good test --optimization --concurrency=4 --exclude-tags integration --test-randomize-ordering-seed random` — 15090 passed, 285 skipped, 0 failed
- `check_package_channel_isolation`, `check_future_delayed_production_ceiling`, `check_test_unit_structure`, `check_package_coverage_floor`, `check_raw_icons_ceiling`, `check_untested_services_floor` — all pass
- `.codex/scripts/test-config.sh` — in sync

### Follow-ups

- The reverted commit carried one thing `main` does not have: a note that local_stack's keycast rejects `127.0.0.1`, because `ALLOWED_TENANT_DOMAINS` lists only `localhost,10.0.2.2`. Worth a small separate PR against `main`, since `main`'s assertions compare `localHost` to `loopbackHost` and would stay green if that constant were changed to the loopback IP. It has no home on this branch now.
- The companion backend PR's `dm-report-contract.test.mjs` comment still names `_reportDmTags()`, which no longer exists.

Neither blocks this branch.

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -217,16 +217,37 @@ class _ReportContentDialogState extends ConsumerState<ReportContentDialog> {
...
+  /// still have to be part of the parked-row identity so a re-drive cannot
+  /// ship stale prose after the user edits and resubmits.
+  String? _queuedModerationDmDetails;
+
+  /// The snapshot a delivered moderation DM carried.
+  ///
+  /// A plain resubmit should not DM the team twice, but a changed resubmit must
+  /// not leave the team with prose or tags from the old report while the
+  /// kind-1984 channel republishes the new one.
+  ({ContentFilterReason reason, String details})? _deliveredModerationDm;
```
NotThatKindOfDrLiz: This is the remaining architecture blocker. The PR has moved beyond wire-format assembly into a moderation-DM submission state machine: queued row id/reason/details, delivered snapshot, outcome, in-flight queueing, cancel/recover behavior, and retry/fallback decisions. That belongs in a Cubit/business layer under the repo rules; if it is not extracted in this PR, the PR needs a filed and linked ReportSubmissionCubit follow-up that Liz explicitly accepts before merge.

NotThatKindOfDrLiz (changes requested):
I pushed the bounded correctness fixes in 963c70c25: sha256 now uses the canonical explicit-hash-or-Blossom-URL resolver, details are snapshotted with the submit, changed resubmits no longer silently reuse an already-delivered DM snapshot, stale cancel ArgumentError is handled, and the focused service/widget tests cover those cases.

Requesting changes for the remaining architecture/process gate: the moderation-DM lifecycle is now substantial retry/fallback business logic inside _ReportContentDialogState. Before merge, either extract this into a ReportSubmissionCubit (preferred) or file and link an explicit follow-up that Liz accepts as the tradeoff for this PR. Liz should resolve her original CHANGES_REQUESTED thread herself after that decision.

Also do not treat the current branch history as merge-ready. The local pre-push hook reported conflicts with main, and a direct rebase attempted to replay hundreds of old commits with repository-wide add/add conflicts, so this needs deliberate author/maintainer cleanup rather than a reviewer-fix force push. Daniel’s takeover commits should also get independent human/author acceptance before merge.

realmeylisdev:
Pushed as `66c027df3`. Both blocking items are done; the third is a maintainer call I can't make on my own PR.

### Architecture gate — fixed

The submission state machine is now `ReportSubmissionCubit` (`mobile/lib/blocs/report/`): queued row id/reason/details, delivered snapshot, terminal outcome, in-flight queueing, cancel/recover, and the retry/fallback decisions between them. Both channels are still driven from one `submit()`, so they cannot label a report differently, and the machine's data lives in emitted state rather than private fields — which is what makes it `blocTest`-able instead of only reachable through a widget. `ReportContentDialog` is now the Page that injects the cubit; `_ReportContentView` is the View, owning form state and rendering only.

Behaviour is unchanged, and the evidence for that is that **all 58 existing widget tests pass with the test file untouched** — including the whole moderation-DM suite. The new unit tests cover the seam directly; each was run against a mutation of the code it guards to confirm it fails.

One thing the extraction had to preserve rather than tidy: the moderation DM's transport stays late-bound behind a resolver instead of being read when the sheet opens. Resolving it eagerly turned `handles unawaited moderation DM preflight failures` red — a moderation-label failure took down a report the kind-1984 channel had already carried. Late binding also means nothing is captured that an account switch could leave stale, which is why the provider carries no `ValueKey`.

### Branch history — fixed, and the diagnosis was off

The branch is now 12 linear commits on current `main`, no merge commits.

The "hundreds of old commits with repository-wide add/add conflicts" was a shallow-clone graft, not the branch. This checkout had two grafted commits in `.git/shallow` and only 325 commits reachable on `origin/main`; that same graft is what makes the pre-push hook report conflicts GitHub never saw — the PR has reported `MERGEABLE` throughout, `BEHIND` only because of the strict status-check policy. `git fetch --deepen=N` clears it. Rebasing to "fix" it is what replays the whole history.

The real rebase is 13 commits and stops on exactly one: `541df60ad` (iOS Simulator `localHost`), which Daniel had already reverted later in the branch as a duplicate of #6840. Dropping that commit together with its revert — a no-op pair — replays the other 11 clean, authorship preserved. I then checked the result rather than trusting it: every report file is byte-identical to the old tip `963c70c25`, and `environment_config.dart` / `auth_providers*.dart` match `main` exactly.

Six `Merge branch 'main'` commits went away with that. Nothing else did — the 13 unique non-merge commits on the old branch are the 13 the rebase started from.

### Daniel's commits — needs a human who isn't me

Not something I can accept: this is my own PR, so there's no reviewer downstream if I wave it through. All seven are retained with authorship intact and their tests pass unchanged against the refactor. @NotThatKindOfDrLiz — that call and your original thread are yours.

### Changes nobody asked for, listed separately

- Two `addError` calls at the cubit's catch sites, deliberately **unwrapped** rather than `Reportable`, so Crashlytics volume is unchanged.
- Removed a `hide LogCategory` that stopped being load-bearing, and one doc link to a type the widget no longer imports.

### Found, not fixed

`reportFailed` interpolates the service's untranslated English detail (`Not authenticated`, `Failed to create report event`) into a localized template, so every non-English locale renders a half-translated sentence. Pre-existing on `main` and not introduced here; fixing it means either new ARB keys or dropping the detail, which is a copy decision rather than a refactor.

### Checks

`dart format` clean · `flutter analyze lib test integration_test` no issues · report/moderation suites 106 passed · `check_test_unit_structure`, `check_untested_services_floor`, `check_raw_icons_ceiling`, `check_raw_textstyle_ceiling`, `check_raw_colors_ceiling`, `check_material_button_ceiling`, `check_raw_dialog_ceiling`, `check_future_delayed_production_ceiling`, `check_package_channel_isolation` all pass.

mobile/lib/widgets/report_content_dialog.dart
realmeylisdev: Extracted in `66c027df3` — `ReportSubmissionCubit` in `mobile/lib/blocs/report/`. Everything you listed moved: queued row id/reason/details, delivered snapshot, terminal outcome, in-flight queueing, cancel/recover, and the retry/fallback decisions. `ReportContentDialog` is now the Page that injects it; `_ReportContentView` keeps form state and rendering.

Both channels still run off one `submit()`, so they can't label a report differently, and the machine's data lives in emitted state rather than private fields — the point being it's now reachable from `blocTest` rather than only through a widget.

All 58 existing widget tests pass with the test file untouched, which is the evidence that behaviour didn't move with the code.

One thing your test caught rather than mine: injecting the DM repository and moderation pubkey at construction — the obvious shape — turned `handles unawaited moderation DM preflight failures` red, because a moderation-label failure then took down a report the kind-1984 channel had already delivered. The transport stays behind a resolver called inside the dispatch's `catch`. That also means nothing is captured, so there's no `ValueKey` to keep it fresh across an account switch.

Leaving this thread for you to resolve, along with the call on Daniel's commits — I can't accept those on my own PR.

mobile/lib/widgets/report_content_dialog.dart
```diff
@@ -172,84 +172,91 @@ class ReportContentDialog extends ConsumerStatefulWidget {
...
+  Widget build(BuildContext context, WidgetRef ref) {
+    return BlocProvider<ReportSubmissionCubit>(
+      create: (_) => ReportSubmissionCubit(
+        // Both dependencies are resolved late rather than watched here. The
+        // moderation DM's transport is read inside the cubit's dispatch, where
+        // a throw stays a DM-only failure instead of breaking the sheet; and
+        // because nothing is captured, no `ValueKey` is needed to survive an
+        // account switch. See .claude/rules/state_management.md ("When the
+        // rule does NOT apply").
+        contentReportingServiceFuture: ref.read(
```
mbradley: The extraction captures the primary report channel once, at sheet-open:

```dart
contentReportingServiceFuture: ref.read(contentReportingServiceProvider.future),
```

Pre-extraction, `_submitReport` re-read this per submit (old line ~416). The difference is load-bearing: `contentReportingServiceProvider` `ref.watch`es `nostrServiceProvider` and `authServiceProvider` (`social_providers.dart:510-511`), so it rebuilds on account switch, sign-out, or a background nostr-client re-init. If any of those fires while the sheet is open, `submit()` awaits the stale captured future and files the report through the previous account's service/signer.

The create-site comment says "nothing is captured, no `ValueKey` is needed" — true of the DM transport, which is correctly late-bound, but not of this future, which *is* captured. The clean fix is to late-bind the service the same way the transport already is: take a `Future<ContentReportingService> Function()` resolver and call it inside `submit()`, restoring the per-submit read and making the comment accurate.

```dart
// cubit ctor
resolveContentReportingService: () =>
    ref.read(contentReportingServiceProvider.future),

// in submit(), replacing the captured-future await
final service = await _resolveContentReportingService();
```

Happy to hand over the full patch (cubit signature + create-site + the one cubit-test construction).

mobile/lib/blocs/report/report_submission_cubit.dart
```diff
@@ -0,0 +1,648 @@
...
+          '(recipient=${transport.pubkey}, '
+          'blocked=${failure.blocked}, '
+          'retryablePending=${failure.retryablePending}): '
+          '${failure.error}',
+          name: 'ReportSubmissionCubit',
+          category: LogCategory.system,
+        );
+      }
+      return failed;
+    } catch (e, stackTrace) {
```
mbradley, line 604: A question rather than a change, since the fix isn't in this file. The #6610 "one DM" guarantee relies on `sendMessage` surfacing its parked-row id through its return value. This catch swallows any throw and returns `true` — but `sendMessage` enqueues the durable row before its network work, and the finalize arms do awaited DB writes before returning the id. A throw after enqueue (e.g. a local DB write failing) never reaches the cubit, so `queuedRumorId` stays null and a resubmit mints a second row → the sweep delivers both → duplicate moderation tickets.

Low likelihood, and the cubit can't distinguish a pre-enqueue throw (safe to resend) from a post-enqueue one. Is the intent to harden `sendMessage`'s contract so it never throws once a row is parked (downgrading to a returned `NIP17SendFailure(queuedRumorId: …)`), or to document the cubit's reliance on that here? Either is fine — flagging so it's a decision rather than an assumption.

mobile/lib/blocs/report/report_submission_cubit.dart
```diff
@@ -0,0 +1,648 @@
...
+        NIP17SendSuccess() => false,
+        // blocked, retryablePending, and hard failures alike.
+        NIP17SendFailure() => true,
+      };
+      final queuedRumorId = switch (dmResult) {
+        // The row was consumed by the delivery.
+        NIP17SendSuccess() => null,
+        // A block is terminal: the send gate drops the row and re-driving
+        // would only re-hit the same policy.
+        NIP17SendFailure(blocked: true) => null,
```
mbradley, line 565: A `blocked` DM result leaves the outcome `pending` rather than terminal, so a resubmit re-attempts the send. I'm not sure whether that's deliberate. If a block can be transient here — a relay policy that later changes, or the recipient no longer gating — then leaving it `pending` so a resubmit retries is the right behavior, and the only cost is one wasted round-trip on a still-blocked send. But if a block on this path is always deterministic, you'd likely want it terminal alongside `unverifiable`, so resubmits short-circuit instead of re-hitting the same wall. Which did you intend? Happy to leave it as-is if retry-on-resubmit is the point.

mbradley (changes requested):
Re-reviewed the `ReportSubmissionCubit` extraction in `66c027d` (the part that hadn't been reviewed yet) and re-verified the earlier tag/reason work still holds at head. CI is green and the branch is `MERGEABLE`. The extraction itself is faithful — the existing widget-test file is untouched by the commit, so the green suite is real evidence behavior was preserved. The concurrency of the DM queue (serial chaining, terminal re-check, emit-after-close guards, stale-cancel-replace) checks out.

One blocking item and two questions inline. Separately, accepting the reviewer-pushed commits is @NotThatKindOfDrLiz's call on her thread, not something addressable here.

NotThatKindOfDrLiz (approved):
Takeover changes are pushed in cbea105af and CI is green.\n\nWhat changed:\n- Restored per-submit resolution of the account-scoped ContentReportingService by injecting a resolver instead of capturing the future when the sheet opens.\n- Made blocked moderation-DM sends terminal/non-retryable via ModerationDmOutcome.blocked.\n- Added cubit regression coverage for both behaviors.\n- Rebasing onto current origin/main was clean, and the PR body now reflects the cubit extraction/current test plan.\n\nVerification:\n- flutter analyze lib test integration_test\n- flutter test test/blocs/report/report_submission_cubit_test.dart test/services/content_reporting_service_test.dart test/services/content_moderation_types_test.dart\n- flutter test test/widgets/report_content_dialog_test.dart\n- pre-push hook: analyze plus changed report tests, 116/116\n- GitHub Mobile CI: green\n\n@mbradley, your blocking service-capture and blocked-send threads are addressed here. I believe the post-enqueue throw concern is non-blocking for this PR because the concrete _finalizeAfterRecipientFailure path is already non-throwing on this branch.\n\n@realmeylisdev, please review the takeover commits, including Daniel's reviewer-authored commits already on the branch, and merge when you're ready after the companion backend PR divinevideo/divine-moderation-service#194 is ready to land with it.

mbradley (approved):
F1 (stale captured reporting service) and F3 (`blocked` → `pending`) are fixed at `cbea105af` and verified at head: the reporting service is now resolved per submit, so an account switch or Nostr re-init while the sheet is open files through the current signer; and a policy-blocked DM is terminal, so resubmits keep the caveat instead of re-hitting the gate. CI green across all shards. Clearing my earlier changes-requested.

F2 stays open as a non-blocking question: the #6610 "one DM" guarantee holds only as long as `sendMessage` never throws after it parks its `outgoing_dms` row — a post-enqueue throw is swallowed here and a resubmit can then mint a second row. Low likelihood, and any fix belongs in `dm_repository`, not this PR. Answer at your discretion; it doesn't hold the merge.

mobile/lib/widgets/report_content_dialog.dart
realmeylisdev: Fixed in `cbea105af`. The cubit now takes a `ContentReportingServiceResolver`
and calls it inside `submit()` (`report_submission_cubit.dart:312`), so the
service is resolved per submit again — an account switch, sign-out or nostr
re-init while the sheet is open files through the current signer instead of the
captured one. The create-site comment is accurate now: nothing is captured.

mobile/lib/blocs/report/report_submission_cubit.dart
realmeylisdev: Deterministic was the right read. Fixed in `cbea105af`: `blocked` is terminal
alongside `unverifiable` (`report_submission_cubit.dart:445`), so a
policy-blocked DM short-circuits on resubmit and keeps the caveat instead of
re-hitting the same gate.

mobile/lib/blocs/report/report_submission_cubit.dart
realmeylisdev: Traced it — there is no post-enqueue throw path at head, so the second-row
scenario isn't reachable.

`sendMessage` parks the row at `dm_repository.dart:2963`. Every await after
that point is individually guarded:

- `resolveDmInboxRelays` (2986) → `_queryOwnDmInbox` (2648) wraps its whole
  body and returns `_OwnDmInboxState.failed` rather than throwing.
- `_sendRumorWithTimeout` → `sendRumor` wraps its whole body and converts
  relay rejection, offline, signer, encryption, block and timeout into
  returned values.
- The success persist arm (3004) catches `on Object` and explicitly does not
  rethrow — "local persistence failure is a degraded state, not a send
  failure".
- `_finalizeAfterRecipientBlocked` (4217), `_finalizeAfterRecipientUnconfirmed`
  (4251) and `_finalizeAfterRecipientFailure` (4174) each catch `on Object`,
  and none of the three rethrow.
- `_ensureConversationVisibleAfterSendFailure` (2872) catches `on Object`.

What can still escape `sendMessage` is all pre-enqueue: `_assertInitialized`,
`validatePubkey`, the empty-content `ArgumentError`, and `buildRumor`. A throw
out of `enqueue` itself leaves no committed row. So `queuedRumorId == null` in
the cubit's catch does mean no row was parked, and the resubmit that mints one
is the correct behaviour rather than a duplicate.

The part of your framing that does hold: the invariant is real but unenforced.
It rests on seven independent catch blocks, and nothing pins "never throws once
a row is parked" — an unguarded await added between the enqueue and the
finalize arms would reintroduce exactly this, silently. That's a `dm_repository`
concern as you said, not this PR. Say the word and I'll file it so the contract
is tracked rather than incidental.

PR #6858, opened 2026-08-07
feat(explore): configurable featured hashtag tab

### LINKED ISSUE #6635: feat(explore): configurable featured hashtag tab

NotThatKindOfDrLiz:
## Summary

Adds the mobile client for the server-driven featured Explore tab shipped in divinevideo/divine-funnelcake#771.

Explore renders an extra tab when the backend serves a configuration, and nothing at all when it does not — no placeholder, no disabled tab, no empty tab. The label, position, slug, active window, and audience rules all come from the backend, so a themed collection can be scheduled, retargeted, or killed without an app release. **No tab-specific value — label, slug, id, or hashtag — is committed to this repository.**

The client stays deliberately thin: it does not know which hashtag a tab represents, does not match tags, and does not filter or sort. It fetches an already-gated, already-ordered list by opaque configuration id and renders it.

### Layers

1. **Client** — the two public endpoints (`/api/featured-tabs`, `/api/featured-tabs/{id}/videos`) plus config models. Labels and disclosure markers arrive as locale maps with a required `default`; placement is expressed by tab name, never index.
2. **Repository** — cache TTL and eligibility. A successful config is cached and dropped after five minutes without a successful refresh, matching funnelcake's stated contract, so a sustained outage cannot strand a killed tab on screen. Note that funnelcake's default cadence is also five minutes, so a cached config is already at that age when a scheduled poll fails: the first transient failure does drop the tab rather than riding through it. Closing that needs a cadence below the TTL or a retry policy, which is a configuration decision rather than a client one. The viewer age gate is read per refresh, not captured, so an age-up or account switch lands on the next poll.
3. **Cubits** — `FeaturedTabsCubit` polls at the server cadence and on app foreground; `FeaturedTabVideosCubit` pages the tab's videos, appending in arrival order.
4. **UI** — the tab, its grid, routing, and analytics.

## Notable decisions

**`TabBarView` children are now derived from the tab-name list.** They were a parallel hardcoded list that happened to match `tabNames` positionally. Splicing a tab in at a configured anchor would have silently desynced the tab bar from its content, so both now read the same ordered names.

**The featured label is the one deliberate exception to l10n-first UI.** Its copy is editorial and scheduled server-side, so it cannot live in the ARB files; the locale map is how it stays translatable. It is treated as untrusted for layout and clamped, since it is not authored in this repo. The exception is documented at the call site.

**Analytics use a generic surface name plus the opaque config id.** `surface_load` reports `surface_name = featured_tab` for every such tab, with `config_id` as the distinguishing parameter, and video views tag `discovery:featured`. Deriving names from tab content would bake editorial names into dashboard schemas and exports.

## Changes beyond the featured tab

Two refactors the feature forced, called out separately since they touch shared code:

- **`ViewTrafficSource` moved from `lib/services/` to `lib/models/`.** The UI/service boundary ratchet correctly rejected a screen importing a service. The enum and its tag mapping are a domain value type, not service behavior. A re-export on the publisher kept the move mechanical at first; the follow-up commit removes that shim and points every call site at the model directly, so the boundary baseline shrinks rather than the service keeping a presentation export.
- **`AnalyticsSurface.featuredTab` and `AnalyticsParam.configId`** added to the `analytics` package.

## Not in this PR

- **No live configuration is seeded yet.** The production endpoint is deployed and returns a valid empty config: `GET https://api.divine.video/api/featured-tabs` -> `200 {"poll_interval_seconds":300,"featured_tabs":[]}`. With no configured tab, Explore correctly renders unchanged and makes no tab-content requests.
- **No screenshots or recordings**, because there is no configured tab to photograph yet. The first seeded configuration should be treated as the live integration check, and someone should be watching when it is enabled.
- **Empty/failure copy is English-only**, registered in `_knownUntranslatedDebt` for the next localization pass. The tab's own label never routes through the ARB files.
- Seeding a configuration needs `POST /api/admin/featured-tabs` with NIP-98 auth until the admin UI (divinevideo/divine-admin-dashboard#11) exists.

## Testing

`flutter analyze lib test integration_test` clean; all four design-system ratchets, the UI/service boundary guard, and both structure floors pass.

72 new tests: client (config parsing, locale fallback, window bounds, ordering, pagination), repository (full visibility matrix, cache TTL serve-then-drop, recovery, re-gating on viewer change), both cubits (polling, kill-switch clearing, page appending, no-further-page guard, failure leaving the loaded page intact, retry), tab ordering (anchor placement, absent anchor, index-shift protection on add and remove), label resolution (locale hit, locale miss, clamping), and widget-level presence, absence, disclosure marker, and an overlong label at the narrowest supported width.

Closes #6635


rabble (dismissed):
This is a strong piece of work and the architecture holds up under reading. Some things I checked and want to confirm rather than just wave through:

- The repository genuinely owns staleness/eligibility/selection — the cubits consume a resolved answer and never re-gate, which is what `architecture.md` asks for.
- `refresh`'s "never throws" contract actually holds: `getFeaturedTabs` funnels everything through a blanket `catch (e) => FunnelcakeException`, the `isAvailable` guard throws `FunnelcakeNotConfiguredException`, and all five exception types extend `FunnelcakeException`, so the `on FunnelcakeException` catch is total.
- `FeaturedTabConfig.operator ==` does include the locale maps via `_mapEquals`, and `_mapHash` sorts before hashing — so a label-only server change does propagate through `updateAvailability`'s equality short-circuit. I went looking for that bug specifically and it isn't there.
- `addError(e, stackTrace)` without a `Reportable` wrapper is the correct call for `FunnelcakeException` under the `error_handling.md` matrix — network/API errors stay out of Crashlytics. Status enums in state, no error strings. `isClosed` guarded on every emit path.
- Deriving `TabBarView` children from `tabNames` is the right call and the reasoning in the description is correct.
- `BlocProvider` keyed on `ValueKey((repository, config.id))` follows the identity-flip rule in `state_management.md`.

One user-reachable bug, and it needs fixing before merge.

## Blocker — a failed pull-to-refresh permanently kills pagination

`FeaturedTabVideosState.copyWith` passes `nextCursor` straight through instead of defaulting to the current value:

```dart
// featured_tab_videos_state.dart:61
return FeaturedTabVideosState(
  status: status ?? this.status,
  videos: videos ?? this.videos,
  nextCursor: nextCursor,          // <-- no `?? this.nextCursor`
  hasMore: hasMore ?? this.hasMore,
  isLoadingMore: isLoadingMore ?? this.isLoadingMore,
);
```

`loadMore` works around this at both of its emits by re-passing `nextCursor: cursor`. `load`'s failure path doesn't:

```dart
// featured_tab_videos_cubit.dart
emit(state.copyWith(status: FeaturedTabVideosStatus.failure));
```

That silently nulls the cursor while leaving `hasMore` true. Reachable sequence, all through the shipped UI:

1. Initial `load()` succeeds — `videos` populated, `nextCursor: 'A'`, `hasMore: true`.
2. User pull-to-refreshes (`_FeaturedGrid` wires `onRefresh: ...load`) and it fails on a flaky network.
3. State is now `videos` preserved, `hasMore: true`, `nextCursor: null`, `status: failure`.
4. `_FeaturedVideosView` checks `state.videos.isNotEmpty` **first**, so it renders `_FeaturedGrid`, not `_FeaturedRetry` — the user sees no error at all.
5. Scrolling to the bottom calls `loadMore`, which hits `cursor == null` and returns immediately, every time.

Pagination is dead for the rest of the session unless another `load()` happens to succeed. Fixing `copyWith` to `nextCursor ?? this.nextCursor` addresses it, and then the two defensive `nextCursor: cursor` re-passes in `loadMore` can go away — they're papering over the same hole.

Worth a test that fails a refresh after a successful first page and then asserts `loadMore` still pages.

Step 4 is a second-order issue in its own right: a failed refresh over existing content is completely silent. Not blocking, but if you're in there anyway, surfacing it is cheap.

## Should fix — the same `copyWith` shape in `ExploreTabsState`

```dart
// explore_tabs_state.dart:135
featuredTab: featuredTab,   // documented as intentional
```

The comment explains the intent (a `null` is how the kill switch clears the tab) and today it's safe — `updateAvailability` is the only caller and it always passes the value. But it means `copyWith` no longer means what it means everywhere else in this codebase, and the next person to add a field and call `state.copyWith(somethingElse: true)` silently kills the featured tab. The failure is invisible: no error, the tab just stops rendering.

This repo already has the idiom for exactly this — `CuratedList.copyWith` uses explicit `clearDescription` / `clearNostrEventId` flags. An explicit `clearFeaturedTab` here keeps the kill-switch behaviour and makes the unsafe call unrepresentable. Given the bug above is the same pattern biting for real one file over, I'd rather both were closed.

## Should fix — label clamping splits emoji

```dart
// explore_tab_labels.dart
return '${trimmed.substring(0, featuredTabLabelMaxLength - 1).trimRight()}…';
```

`substring` cuts UTF-16 code units, so a label that overflows mid-surrogate-pair or mid-ZWJ-sequence renders as a replacement character. For editorial campaign copy this isn't hypothetical — a leading emoji is close to the default for a themed collection, and this is precisely the untrusted-input path the constant exists to defend.

`.characters` is already the established fix in this repo, with the same reasoning written down at `comment_item.dart:391` ("Handles compound emojis correctly: Dart's `.characters` segments ZWJ"). `trimmed.characters.take(featuredTabLabelMaxLength - 1).toString()` is the drop-in.

(Aside: 24 *characters* doesn't bound rendered width anyway — 24 CJK glyphs are roughly double 24 Latin ones. The tab bar scrolls, so that's fine; it just means this guard is about sanity, not layout, and grapheme-safety matters more than the exact number.)

## Nits

- **`_resolvePendingFeaturedSlug` mutates state during `build()`.** It sets `_pendingFeaturedSlug = null` as a side effect of a build-phase call. It's idempotent and guarded so it won't misbehave today, but build should stay pure — this reads more naturally as a post-frame callback or as part of the controller re-init path it feeds.
- **`clearCache`'s doc comment instructs a call nobody makes.** "Call on sign-out and account switch" — the only wiring is `ref.onDispose(repository.clearCache)`, which fires on a Funnelcake client identity change, not an account switch. As far as I can tell nothing *needs* to call it, because the viewer gate is re-applied per refresh (including on the cache path) and the config itself is explicitly non-personalized — which is the better design and what the description claims. So the comment is just stale, and it'll read as a missing call to whoever reads it next.

## On the undeployed endpoint

Shipping client code that has never run against the live API is a real risk, but I think the framing in the description is right: no configuration means no tab, which is the same as the failure mode, so the blast radius while the deploy is pending is genuinely zero. Worth being explicit that the first live configuration is effectively the integration test, and someone should be watching when it's seeded.

Requesting changes for the pagination bug; the other two are small and the rest of this is ready.

NotThatKindOfDrLiz:
Addressed the review feedback in `06f170908b220b88180a0a82db7e8e23f7a85983`.

What changed:
- fixed the featured-video refresh pagination regression and added the successful first page -> failed refresh -> load more regression test
- normalized the nullable `copyWith` patterns for `FeaturedTabVideosState`, `ExploreTabsState`, and `FeaturedTabsState`
- made featured tab label clamping grapheme-safe and added compound-emoji coverage
- clamped server-provided poll intervals to 30s..1h with package tests
- surfaced a failed refresh over existing featured content with a snackbar
- moved pending featured-slug consumption out of build-phase mutation
- documented the deliberate always-on polling behavior after Explore is mounted
- corrected the stale `clearCache` comment
- filed #6901 for the deferred non-English featured-tab strings and linked the known debt entry

Verification:
- `flutter test test/src/featured_tabs_test.dart` from `mobile/packages/funnelcake_api_client`
- targeted mobile tests for featured tab cubits, explore tab labels, featured videos UI, explore tab bar, and ARB consistency
- `flutter analyze`
- CI is green after rerunning the initially failed shard 1 job; the same CI shard command passed locally before rerun.

NotThatKindOfDrLiz:
Rebased onto current `origin/main` and force-pushed `3cf3d29fe18a290e86c6c5ce17a01f8e68d951db`.

What changed in this push:
- resolved the Explore reduced-motion tab-controller conflict while preserving featured-tab deep-link resolution and foreground/minor-status refresh behavior
- kept both main's sound-library l10n debt key and this PR's featured-tab l10n debt keys in the ARB consistency guard
- updated the PR body now that production `/api/featured-tabs` returns a valid empty config instead of 404

Local verification:
- `flutter test test/src/featured_tabs_test.dart` from `mobile/packages/funnelcake_api_client`
- `flutter test test/blocs/explore_tabs/explore_tabs_featured_test.dart test/blocs/featured_tabs/featured_tabs_cubit_test.dart test/blocs/featured_tabs/featured_tab_videos_cubit_test.dart test/blocs/featured_tabs/featured_tab_surface_telemetry_test.dart test/repositories/featured_tabs_repository_test.dart test/screens/explore/explore_tab_labels_test.dart test/screens/explore/tabs/featured_videos_tab_test.dart test/screens/explore/widgets/explore_tab_bar_featured_test.dart test/l10n/arb_consistency_test.dart`
- pre-push hook: merge-conflict check, analyze, and changed-file tests passed

GitHub reports the branch mergeable now. CI is running on the new head.

mobile/lib/screens/explore/tabs/featured_videos_tab.dart
```diff
@@ -0,0 +1,180 @@
...
+  Widget build(BuildContext context, WidgetRef ref) {
+    final repository = ref.watch(featuredTabsRepositoryProvider);
+    return BlocProvider<FeaturedTabVideosCubit>(
+      // Rebuild the cubit when the repository identity changes (environment or
+      // relay swap) or when anything about the configuration changes. Keyed on
+      // the whole config, not just its id: the admin endpoint upserts by id, so
+      // a tab can be retargeted in place, and the backing hashtag is
+      // deliberately absent from the public config — the visible fields are the
+      // only retarget signal the client can see.
+      key: ValueKey((repository, config)),
```
dcadenas, line 42: Keying on the whole config catches any retarget that changes a visible field, which is the normal case. It cannot catch one where the operator changes only the backing hashtag and leaves label, slug, window and position identical — the hashtag never reaches the client by design.

Covering that would mean revalidating the videos endpoint on every config poll for every open featured tab, which changes the fetch model for the feature and costs a request per poll per client. That is your call, not something to take on the branch. funnelcake's API doc treats this as bounded staleness rather than a defect ("clients can display cached labels for up to their 5-minute config cache window"), which is why I stopped here.

The empty-page retry added alongside this does cover the version of it that shows an empty grid.

mobile/lib/blocs/featured_tabs/featured_tabs_cubit.dart
```diff
@@ -0,0 +1,82 @@
...
+  final FeaturedTabsRepository _repository;
+  final bool Function() _viewerIsMinor;
+
+  Timer? _pollTimer;
+
+  /// Sequence number of the newest started refresh.
+  ///
+  /// Request bookkeeping rather than UI state, like [_pollTimer] beside it:
+  /// nothing renders from it, so it stays off the state object.
+  int _refreshGeneration = 0;
```
dcadenas, line 40: This orders emission so a refresh issued before a server kill cannot publish after the one that removed the tab. The repository carries a matching sequence for the cache write; both are needed, since only the cubit knows whether the caller awaiting a given refresh is still current.

Worth flagging that the mutable-field question here is genuinely arguable. One reading is that `state_management.md`'s rule targets data the UI needs and cannot see — its examples are `_retryCount`, `_isInitialized`, `_selectedIds` — and that a request sequence with no observer sits in the same category as the `_pollTimer` above it. The other reading takes the rule literally, since it excepts only injected dependencies.

I kept the field, because moving it into state means emitting on every refresh start purely to bump a counter. Happy to be overruled.

dcadenas (approved):
Ten fixes pushed onto the branch, one commit each, all in behaviour this change introduces.

The architecture holds up: the repository owns staleness, eligibility and selection, the cubits own cadence and paging, and the UI reads a resolved answer. Deriving the `TabBarView` children from the ordered tab-name list closes a real desync class. What needed work was concurrency and untrusted input, not layering.

## What changed

- Two ordering races. Overlapping config refreshes could let a request issued before a server kill land last, both at the emission boundary and in the shared cache, putting a killed tab back on screen. A pull-to-refresh landing mid-page could splice a page from the old cursor onto the refreshed list.
- Untrusted server copy. The label clamp counted graphemes but a tab lays out at a fixed height, so newlines and bidi overrides still got through, and the disclosure marker bypassed the clamp entirely. Both strings now go through one sanitizer.
- Tab selection. The launch tab re-won on every availability change, so a user who arrived on `/explore/tab/popular` and then moved was pulled back whenever the configuration changed. That is issue #6635's "adding or removing the tab does not shift the user off their selected tab".
- Two contract gaps against funnelcake's own docs: the cached config is dropped at five minutes as the API contract specifies, and an empty page now retries on the next poll as the integration guide asks.

Every added test was run against the previous code first and fails there.

## Checks

- `dart format` and `flutter analyze lib test integration_test` clean.
- The explore, featured-tab, repository, router and l10n suites pass locally, plus the `funnelcake_api_client` and `analytics` package suites.
- All required checks green on this head.

## Two things to decide

Neither blocks merge, and both are calls I made on your behalf rather than plain bug fixes — see the inline notes.

- A same-id retarget that changes no visible field still leaves a populated grid on the old collection.
- The refresh ordering guard is a private mutable field on a cubit.

One factual clause in the description was corrected: the repository bullet claimed the cache rides through a transient failure, which it does not at funnelcake's default cadence.

NotThatKindOfDrLiz:
Cleanup commit pushed: 43b1bf9e78706032033ad784fc2af2b473072553 (refactor(analytics): finish traffic source import migration).

What changed:
- finished the ViewTrafficSource import migration by moving enum-only presentation/test imports to package:openvine/models/view_traffic_source.dart
- removed the compatibility re-export shim from view_event_publisher.dart
- added direct model imports in the service/test files that still need both publisher APIs and traffic-source types
- regenerated mobile/scripts/baseline/ui_service_imports.txt, shrinking the UI service-boundary baseline

Verification:
- bash mobile/scripts/check_ui_service_boundary.sh
- targeted mobile Flutter suite covering affected analytics/view-event/explore/featured/l10n paths
- flutter test in mobile/packages/funnelcake_api_client
- flutter test in mobile/packages/analytics
- pre-push hook: merge-conflict check, analyze, and changed-file tests passed
- GitHub CI is green on 43b1bf9e7

dcadenas (approved):
Re-reviewed at `43b1bf9e7`. My earlier approval was at `d905c913e`, which the rebase rewrote, so this replaces it rather than repeating it.

The rebase preserved all ten of my fix commits exactly — identical patch-ids, in order, `c7298f051..1d999ada7`. Checked the whole-PR net diff too, not just the commits: same size both sides, differing only in blob hashes, hunk offsets, and two context lines in a barrel file from base drift.

The cleanup commit is in scope and clean. It deletes a re-export this PR itself introduced, so it finishes the migration rather than importing unrelated work — leaving the shim would have been the deferred cleanup the repo guidelines forbid. No leftovers: nothing still names `ViewTrafficSource` without a direct model import, and the UI/service boundary baseline shrinks by 13 files.

## Checks

`dart format`, `flutter analyze lib test integration_test`, and `check_ui_service_boundary.sh` clean. 91 tests across the eight suites the cleanup touches, 101 across the explore and featured-tab suites, plus `funnelcake_api_client` and `analytics`. All 17 required checks green on this head.

Not covered locally, so CI is the only signal for them: the `very_good` optimized run, the two nested package roots (declared only through reusable workflows), and eleven multi-command ratchet scripts.

## Two things still to decide

Unchanged from my last review, and both independently confirmed this round. See the inline notes.

- A same-id retarget that changes no visible field still leaves a populated grid on the old collection.
- The refresh ordering guard is a private mutable field on a cubit.

## Considered and not changed

- The five-minute cache means a single failed poll drops the tab at a cadence of five minutes or slower, which reads against "one failed request must not make the tab flicker out mid-scroll". A client-side retry does not actually close it: the tab is already dropped at the moment the refresh fails, so a retry shortens the gap rather than preventing it. The only fix that closes it is a poll cadence below the TTL, and the issue asks for the cadence to stay server-supplied. Worth confirming with whoever owns that configuration that the deviation is accepted — the description attributes it to configuration, which is right for the cadence half but loose about retries.
- `surface_load` carries the generic surface name and the config id, and video impressions carry it too, so the acceptance criterion is met. The gap is narrower than it first looks: the featured tab never marks surface visibility, so its `visible_ms` reports `-1` where the one other surface on that tracker reports a real number. A refinement to the metric, not a hole in attribution.

One sentence in the description was stale after the cleanup commit — it said the re-export meant no call site changed, which stopped being true when that commit removed it. Corrected in place.

mobile/lib/screens/explore/tabs/featured_videos_tab.dart
NotThatKindOfDrLiz: Decision: accept this as bounded staleness. The public config deliberately does not expose the backing hashtag, so a same-id retarget that changes no visible field is not observable to the client. Revalidating the videos endpoint on every config poll would add one request per poll per open featured tab to close a case the API contract already treats as bounded. The empty-grid variant is covered by the retry-on-poll behavior.

mobile/lib/blocs/featured_tabs/featured_tabs_cubit.dart
NotThatKindOfDrLiz: Decision: keep `_refreshGeneration` as private request bookkeeping. It does not represent UI-observable state; it only prevents an older refresh from publishing after a newer refresh, in the same category as the private poll timer. Moving it into state would force an emit just to bump a counter the UI must never render from. The behavior is pinned by the slower-earlier-refresh test.

NotThatKindOfDrLiz:
@rabble please re-review when you have a chance. Current head is `5252f523e`.

Mapping your review items to the commits that closed them:
- Pagination after failed refresh: `267fc46cf` preserves the cursor through first-page refresh failure and removes the load-more workaround; `featured_tab_videos_cubit_test.dart` covers successful first page -> failed refresh -> load more.
- `ExploreTabsState.copyWith` nullable shape: `267fc46cf` adds the explicit clear flag pattern.
- Emoji-safe label clamp: `267fc46cf` switches to grapheme-safe truncation, and `42eb096f` moves both server-supplied featured-tab strings through the shared sanitizer with newline/bidi coverage.
- `_resolvePendingFeaturedSlug` build-phase mutation: `c7298f051` keeps resolution pure and consumes pending launch state post-frame.
- Stale `clearCache` docs / cache contract wording: `115386dc0` updates the cache policy to match the funnelcake five-minute contract.
- Silent failed refresh over existing content: `20ad070f2` surfaces the failure with the featured-tab snackbar.
- Follow-up cleanup: `5252f523e` removes the unused `clearNextCursor` flag, removes test-only featured-tab state getters, and makes `_FeaturedVideosView` a plain `StatelessWidget`.

I also resolved the two dcadenas decision threads after recording the decisions: accept same-id hashtag-only retargets as bounded staleness, and keep `_refreshGeneration` as private request bookkeeping rather than UI state.

Local verification on this head:
- `flutter test test/blocs/featured_tabs/featured_tabs_cubit_test.dart test/blocs/featured_tabs/featured_tab_videos_cubit_test.dart test/screens/explore/tabs/featured_videos_tab_test.dart test/screens/explore/widgets/explore_tab_bar_featured_test.dart test/screens/explore/explore_tab_labels_test.dart test/repositories/featured_tabs_repository_test.dart`
- pre-push hook: merge-conflict check, analyzer, and changed-file tests passed

NotThatKindOfDrLiz:
Rebased onto current `origin/main` and force-pushed `5d2c06bccdb2e11a849e9ce06bf5e32c39ae61f5`.

Current cleanup commits:
- `f44a6c48a` — prunes the dead featured-tab cleanup leftovers (`clearNextCursor`, test-only getters, and `_FeaturedVideosView` using `ConsumerWidget` without `ref`).
- `5d2c06bcc` — keeps the featured telemetry test fake compatible with the current `AnalyticsEventSink` contract on `main`.

Local verification passed after the rebase: focused featured-tab test suite, `flutter analyze lib test integration_test`, and the pre-push merge-conflict/analyze/changed-file test hook. @rabble, please re-review this rebased head.

NotThatKindOfDrLiz:
@rabble please re-review the rebased head `78b3fd2c44b50832735af625de8d9cb9b03b5579`.

What changed in this push:
- rebased onto current `origin/main` (`4fc00d0911`) and resolved the Explore conflict by keeping both the search-bar semantic identifier from main and the featured-tab wiring from this PR
- removed the unused `FeaturedTabsState.copyWith(clearTab:)` flag and its self-referential test; clearing the resolved tab still happens through the fresh `FeaturedTabsState(tab: snapshot.tab, ...)` emitted by `FeaturedTabsCubit.refresh()`

Local verification on this head:
- `flutter test test/blocs/featured_tabs/featured_tabs_cubit_test.dart test/blocs/featured_tabs/featured_tab_videos_cubit_test.dart test/screens/explore/tabs/featured_videos_tab_test.dart test/screens/explore/widgets/explore_tab_bar_featured_test.dart test/screens/explore/explore_tab_labels_test.dart test/repositories/featured_tabs_repository_test.dart`
- `flutter test test/src/featured_tabs_test.dart` from `mobile/packages/funnelcake_api_client`
- `bash mobile/scripts/check_ui_service_boundary.sh`
- pre-push hook: merge-conflict check, analyzer, and changed-file tests all passed

GitHub CI is running on the new head.

rabble (commented):
Re-reviewed. **Blocker fixed**, and fixed in the shape that keeps `null` meaningful. CI green (16/16) on `78b3fd2c4`.

`copyWith` now does `nextCursor: nextCursor ?? this.nextCursor`, so `load`'s failure path no longer nulls the cursor while leaving `hasMore` true — the sequence I traced (refresh fails on flaky network → grid still renders because `videos.isNotEmpty` → `loadMore` returns immediately forever) is closed.

The part I specifically went back to check: a `??` default normally costs you the ability to *clear* a nullable field, and pagination genuinely needs `nextCursor: null` when the last page lands. That's handled — both authoritative-page paths now build `FeaturedTabVideosState(...)` directly rather than via `copyWith`, so an exhausted page still sets `nextCursor` to null and `hasMore` to false. `copyWith` is left doing only partial status/flag updates, where preserving the cursor is always right. That's a cleaner split than adding a `clearNextCursor` flag.

The `if (state.nextCursor != cursor) return;` guards on both the success and failure sides of `loadMore` weren't in my review and are a real catch — a pull-to-refresh completing mid-page would otherwise splice a page computed against the old ordering onto the refreshed list, or reset `isLoadingMore` against a state that had already moved on. Both `onRefresh` and `onLoadMore` are wired to the same cubit, so that race is reachable, and the comment explains why appending is the unrecoverable case.

Everything I confirmed in the first pass still holds on this revision: repository-owned staleness/eligibility, the total `on FunnelcakeException` catch, `_mapEquals`/`_mapHash` covering the locale maps so a label-only change propagates, bare `addError` being correct for network errors under the `error_handling.md` matrix, and the `ValueKey((repository, config.id))` identity guard.

NotThatKindOfDrLiz:
> Re-reviewed. **Blocker fixed**, and fixed in the shape that keeps `null` meaningful. CI green (16/16) on `78b3fd2c4`.
> 
> `copyWith` now does `nextCursor: nextCursor ?? this.nextCursor`, so `load`'s failure path no longer nulls the cursor while leaving `hasMore` true — the sequence I traced (refresh fails on flaky network → grid still renders because `videos.isNotEmpty` → `loadMore` returns immediately forever) is closed.
> 
> The part I specifically went back to check: a `??` default normally costs you the ability to _clear_ a nullable field, and pagination genuinely needs `nextCursor: null` when the last page lands. That's handled — both authoritative-page paths now build `FeaturedTabVideosState(...)` directly rather than via `copyWith`, so an exhausted page still sets `nextCursor` to null and `hasMore` to false. `copyWith` is left doing only partial status/flag updates, where preserving the cursor is always right. That's a cleaner split than adding a `clearNextCursor` flag.
> 
> The `if (state.nextCursor != cursor) return;` guards on both the success and failure sides of `loadMore` weren't in my review and are a real catch — a pull-to-refresh completing mid-page would otherwise splice a page computed against the old ordering onto the refreshed list, or reset `isLoadingMore` against a state that had already moved on. Both `onRefresh` and `onLoadMore` are wired to the same cubit, so that race is reachable, and the comment explains why appending is the unrecoverable case.
> 
> Everything I confirmed in the first pass still holds on this revision: repository-owned staleness/eligibility, the total `on FunnelcakeException` catch, `_mapEquals`/`_mapHash` covering the locale maps so a label-only change propagates, bare `addError` being correct for network errors under the `error_handling.md` matrix, and the `ValueKey((repository, config.id))` identity guard.
> 
> **Nothing blocking from me.** My `CHANGES_REQUESTED` is stale and needs a maintainer to convert.

@rabble Can you please update your agent so that it clears your requested changes and approves? TYSM!

PR #6861, opened 2026-08-08
fix(lists): dismiss editor before relay acknowledgement

### LINKED ISSUE #6862: fix(lists): dismiss editor before relay acknowledgement
Auto-created from PR #6861.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6861
**Branch:** `fix/list-settings-dialog-dismiss`

rabble:
Closes #6862

## Summary

- dismiss list settings after the local update begins instead of waiting for relay acknowledgement
- preserve relay failure feedback on the underlying screen after dismissal
- cover rename and private-to-public saves while relay work is pending

## Testing

- flutter test test/widgets/add_to_list_dialog_test.dart
- flutter analyze

### DIFF
--- a/mobile/lib/widgets/add_to_list_dialog.dart
+++ b/mobile/lib/widgets/add_to_list_dialog.dart
@@ -244,17 +244,31 @@ class _CreateListDialogState extends ConsumerState<CreateListDialog> {
             !await _confirmVisibilityChange(existingList.isPublic)) {
           return;
         }
-        final updated = await listService?.updateList(
+        if (!mounted) return;
+        if (listService == null) {
+          _showSaveFailed();
+          return;
+        }
+
+        final messenger = ScaffoldMessenger.of(context);
+        final failureMessage = context.l10n.listUpdateFailed;
+        final updateFuture = listService.updateList(
           listId: existingList.id,
           name: name,
           description: _descriptionController.text.trim(),
           isPublic: _isPublic,
         );
-        if (!mounted) return;
-        if (updated ?? false) {
-          context.pop();
-        } else {
-          _showSaveFailed();
+
+        // updateList applies local fields before waiting for the relay, and a
+        // visibility change can wait through the relay deadline. Close the
+        // editor now so a slow relay does not make the save look unresponsive.
+        Navigator.of(context).pop();
+
+        if (!await updateFuture && messenger.mounted) {
+          _showSaveFailed(
+            messenger: messenger,
+            failureMessage: failureMessage,
+          );
         }
         return;
       }
@@ -334,14 +348,18 @@ class _CreateListDialogState extends ConsumerState<CreateListDialog> {
     return confirmed ?? false;
   }
 
-  // Keeps the dialog open so the typed name and description survive a retry.
-  void _showSaveFailed() {
-    ScaffoldMessenger.of(context).showSnackBar(
+  // A retained messenger lets an edit report relay failure after dismissing.
+  void _showSaveFailed({
+    ScaffoldMessengerState? messenger,
+    String? failureMessage,
+  }) {
+    (messenger ?? ScaffoldMessenger.of(context)).showSnackBar(
       SnackBar(
         content: Text(
-          _isEditing
-              ? context.l10n.listUpdateFailed
-              : context.l10n.listCreateFailed,
+          failureMessage ??
+              (_isEditing
+                  ? context.l10n.listUpdateFailed
+                  : context.l10n.listCreateFailed),
         ),
         duration: const Duration(seconds: 2),
       ),
--- a/mobile/lib/widgets/add_to_list_dialog.dart
+++ b/mobile/lib/widgets/add_to_list_dialog.dart
@@ -4,7 +4,6 @@
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:go_router/go_router.dart';
 import 'package:models/models.dart' hide LogCategory;
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/providers/app_providers.dart';
@@ -101,7 +100,10 @@ class SelectListDialog extends StatelessWidget {
                 },
                 child: Text(l10n.listNewList),
               ),
-              TextButton(onPressed: context.pop, child: Text(l10n.listDone)),
+              TextButton(
+                onPressed: () => Navigator.of(context).pop(),
+                child: Text(l10n.listDone),
+              ),
             ],
           );
         },
@@ -147,6 +149,11 @@ class SelectListDialog extends StatelessWidget {
 }
 
 /// Dialog for creating a new curated list, optionally adding [video] to it.
+///
+/// Like [SelectListDialog] this is always presented with `showDialog` and is
+/// never registered as a `go_router` route, so every dismissal here goes
+/// through [Navigator] — the navigator that owns the dialog route — rather
+/// than `context.pop`.
 class CreateListDialog extends ConsumerStatefulWidget {
   const CreateListDialog({this.video, this.existingList, super.key});
   final VideoEvent? video;
@@ -223,7 +230,10 @@ class _CreateListDialogState extends ConsumerState<CreateListDialog> {
         ],
       ),
       actions: [
-        TextButton(onPressed: context.pop, child: Text(l10n.listCancel)),
+        TextButton(
+          onPressed: () => Navigator.of(context).pop(),
+          child: Text(l10n.listCancel),
+        ),
         TextButton(
           onPressed: _saveList,
           child: Text(_isEditing ? l10n.listSave : l10n.listCreate),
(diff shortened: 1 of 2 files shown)


rabble:
## Review notes (agent review — this is my own PR, so this is a comment, not an approval)

The async hygiene here is right: `messenger` and `failureMessage` are both captured before the pop, `messenger.mounted` is checked after the await, and `failureMessage` being pre-resolved means the unmounted `State` is never asked for `_isEditing` or `context.l10n`. I also checked `CuratedListService.updateList` — it catches everything and returns `false`, so there's no unhandled-rejection window between starting `updateFuture` and awaiting it. The two new tests are real behaviour tests that would fail if the pop moved back after the await.

Two blockers and some smaller things.

### Blocker 1 — a failed rename now shows the new name *and* a failure snackbar at the same time

`updateList` is deliberately local-first:

```dart
// mobile/lib/services/curated_list_service.dart:469
_lists[listIndex] = updatedList.copyWith(isPublic: list.isPublic);
await _saveLists();   // -> notifyListeners() on line 1248, before any publish
```

The rename is persisted and broadcast to listeners *before* `_publishListToNostr` is awaited. So on relay failure the sequence the user now sees is: editor closes → list re-renders with the new name → "Couldn't update list" snackbar appears over it.

Previously the dialog stayed open, so the user never saw that contradiction. Dismissing early surfaces it. The rename genuinely succeeded locally — which is the documented intent — so `listUpdateFailed` is now the wrong message for the rename-only path. Either the copy needs to distinguish "saved on this device, couldn't reach relays" from an actual failure, or the rename path shouldn't report failure at all when only the relay projection failed.

### Blocker 2 — the visibility-change failure path is the one that regressed, and the new test doesn't cover it

Same function, the deliberate exception:

```dart
// Visibility ... alone stays at its old value until the relay confirms.
```

So when a private→public save fails, the toggle silently snaps back. Before this change the dialog stayed open with the toggle still flipped and the typed text intact — the old `_showSaveFailed` comment said exactly that ("Keeps the dialog open so the typed name and description survive a retry"). Now the editor is gone, the toggle is reverted, and the only signal is a 2-second snackbar with no retry affordance.

`dismisses while publishing a private list with a pending relay update` is the test that should catch this, but it completes with `updateCompleter.complete(true)` — the success path. The failure branch for a visibility change is untested, and it's the case where dismissing early actually costs the user work.

I'd want either a test that completes that one with `false` and asserts what the user is left with, or a decision that visibility saves keep the old blocking behaviour and only name/description edits dismiss early. The latter is defensible — visibility is the field that *isn't* local-first, so it's the one where waiting is honest.

### Smaller

- **Three pop mechanisms in one widget now.** Cancel uses `context.pop`, the create path uses `context.pop()`, and the edit path is now `Navigator.of(context).pop()`. `Navigator.pop` is arguably the more correct one for a `showDialog` route, but the change isn't mentioned in the description and nothing in the new flow requires it. Either make it deliberate and convert all three, or put it back to `context.pop()`.
- **`_showSaveFailed({messenger, failureMessage})` permits meaningless combinations** — passing a messenger without a message falls back to `context.l10n` on a possibly-unmounted `State`, which is the exact bug the parameter exists to avoid. Since the two are only ever passed together, a single required object (or a separate `_showSaveFailedDetached(messenger, message)`) makes the unsafe combination unrepresentable.
- **2-second snackbar** is short for a message the user can no longer act on from the editor. Worth a `SnackBarAction` that reopens it with the input intact, which would also recover most of what Blocker 2 removes.
- **The create path still blocks on the relay** and the description doesn't say why the asymmetry is intentional. I assume it's because the new list's ID isn't known until it returns — worth one sentence so the next reader doesn't "fix" it.
- `addTearDown` completing the completer *and* the test completing it explicitly is belt-and-braces; harmless, just noting it's redundant in the first test.

CI is green (15/15). Nothing here is structural — Blocker 1 is a copy/semantics decision and Blocker 2 is a test plus a scope call.

rabble:
Pushed fixes for the review findings above, in two commits. **The second one is a product call and is deliberately revertable on its own** — please read before merging.

### `7af507f21` — mechanical, no behaviour change to the save flow

- **Pop unified on `Navigator.of(context).pop()`.** Checked all four presentation sites (`curated_list_feed_screen.dart:462`, `explore_lists_tab.dart:75`, `profile_lists_grid.dart:63`, `share_action_button.dart:381`) — every one goes through `showDialog`/`showVideoPausingDialog` with `useRootNavigator: true`, and neither dialog is registered in `app_router`. `context.pop` only reaches these via a `ModalRoute.isCurrent` special case in `GoRouterDelegate._findCurrentNavigators`, and throws outright with no GoRouter in the tree. Rationale recorded on the `CreateListDialog` doc comment. `SelectListDialog`'s Done button converted too, so the file doesn't keep one straggler that re-earns the same comment.
- **`_showSaveFailedDetached(messenger, message)`** takes both positionally and required, so the messenger-without-message combination — the exact bug the parameter existed to prevent — is now unrepresentable.
- **Create-path asymmetry documented.** Confirmed two independent reasons it must keep blocking: `_createList` *rolls the list back out of local storage* on relay rejection (`_lists.removeWhere` + `_saveLists`, returns null), so unlike an edit there is nothing saved to dismiss onto; and the generated id needed for the follow-up `addVideoToList` only exists once it returns.

### `b6311963f` — blocks only the visibility half of an edit ⚠️

This one contradicts the motivation written in the diff comment, so it needs your call.

**The finding:** `_publishListToNostr(confirmed: true)` awaits `publishEventAwaitOk` with a **15s** deadline (`curated_list_service.dart:1136`), and `updateList` takes that path whenever `updatedList.isPublic` (line 477) — **including a plain rename of an already-public list**. So the slow save isn't specific to visibility changes. The in-diff comment ("a visibility change can wait through the relay deadline") names the wrong cause, and since public is the default, the common rename was the slow one.

That decouples the two cases the PR bundled:

| Save | Local-first? | Blocks 15s? | After this commit |
|---|---|---|---|
| Rename/description, private list | yes | never publishes | dismisses instantly |
| Rename/description, public list | yes | yes | **dismisses instantly** — the PR's win, kept |
| Visibility flip | **no** — `isPublic` held at the old value until a relay accepts | yes | **waits**, stays open on failure |

The dismissal boundary now sits exactly where `updateList`'s own documented boundary already is (`curated_list_service.dart:464-470`): fields that are already saved dismiss, the one field that can silently not happen waits. That restores the retry affordance precisely where the deleted `_showSaveFailed` comment said it mattered — a rejected visibility flip was the only save leaving unrecoverable user intent on screen, since the switch reverts in the service (pinned by `curated_list_service_crud_test.dart:776`) and the editor was gone.

Costs no copy change, so it ships here rather than waiting on translations.

**If you'd rather keep uniform early dismissal, `git revert b6311963f` is clean** and leaves the rest intact.

### Still open — needs product, not code

A public-list rename the relay rejects still renders the list **with the new name** *and* `listUpdateFailed` ("Couldn't update this list. Try again."). That's the wrong message for a save that succeeded locally, but fixing it needs a new ARB key across 17 locales — and `updateList` returns a single bool conflating list-not-found, deletion-publish-failed, and publish-rejected, so the widget currently can't tell them apart even if the copy existed. Both left alone deliberately.

Tests: 20 pass in `add_to_list_dialog_test.dart`, including a new `keeps a rejected visibility change on screen for a retry`. `flutter analyze lib test integration_test` clean.

mobile/lib/widgets/add_to_list_dialog.dart
```diff
@@ -240,25 +250,57 @@ class _CreateListDialogState extends ConsumerState<CreateListDialog> {
...
+          return;
+        }
+
+        // Name and description are already stored locally before updateList
+        // awaits the relay, so nothing the user typed is riding on the answer.
+        // Close now rather than let a slow relay make the save look
+        // unresponsive; the messenger and message have to be resolved first
+        // because the pop can unmount this State.
+        final messenger = ScaffoldMessenger.of(context);
+        final failureMessage = context.l10n.listUpdateFailed;
```
realmeylisdev, line 292: Reproduced the case you left open, on a simulator with the network dropped mid-save (public list, name-only edit):

- editor dismisses right away. The log has `[UI] Page closed` at `21:14:53.343` and the Keycast sign attempt at `21:14:53.355`, so the pop lands 12ms before any async work starts.
- the screen behind re-renders under the new name, still labelled Public
- snackbar reads "Couldn't update this list. Try again."

The rename itself is safe. It survived a force-quit plus a relay resync, because `_processListEvent` only lets a relay copy win when `event.createdAt > existingList.updatedAt` (`curated_list_service.dart:1798`). No data loss, and the local-first reasoning in the comment above holds.

What reads wrong is "Try again": the editor is gone and the snackbar carries no action, so there is nowhere to do it. Nothing retries the projection on its own either. This path goes through `publishEventAwaitOk`, which passes a non-null deadline, and `RelayPool` only queues a failed send when `deadline == null` (`relay_pool.dart:1874`). The relay keeps serving the old name until some later write republishes the list, so a second edit or an `addVideoToList` closes it.

Agree with leaving it here. A distinct string needs a new key across all 22 `app_*.arb` files, and `updateList` returning one bool can't separate "saved here, relay refused" from "nothing saved". Recording the repro so the follow-up has it.

realmeylisdev (approved):
ran it on a sim. the rename pops 12ms before the keycast sign call even fires, and the publish it used to sit on took 8.6s, so it does what it says. visibility flip still waits and keeps the editor up on a rejection, which is the right place to draw the line.

left one inline: reproduced the rejected public rename, you get the new name plus "try again" and no editor to try again in. copy call, not blocking.

Issue #858, opened 2026-01-08
fix: late initialization error


andyhorn:
A late initialization error is being thrown in the VideoFeedItem every time a feed item is loaded and built.

It's not crashing the app, but it does create a lot of noise when debugging.

**Steps:**
  1. Launch the app with debugging and note the exception
  2. Scroll through videos and note the repeated exceptions

**Expected Behavior:**
  1. App launches without exceptions
  2. Can scroll through videos without exceptions

<img width="752" height="145" alt="Image" src="https://github.com/user-attachments/assets/27c564fa-2802-4186-bfae-007a24ddf6f8" />

PR #6865, opened 2026-08-08
feat(analytics): attribute home-feed views to the feed mode that served them

### LINKED ISSUE #6866: feat(analytics): attribute home-feed views to the feed mode that served them
Auto-created from PR #6865.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6865
**Branch:** `feat/home-feed-source-detail`

rabble:
Closes #6866

## Problem

`home` accounts for **28.7% of measured feed re-serves** — and it is currently impossible to tell which feed is responsible.

`VideoFeedPage` renders **five different feed modes** and reports `source=home` for every one of them:

| Mode | Backing call |
|---|---|
| For You | `getRecommendedVideos` → `/api/users/{pubkey}/recommendations` |
| Following | `getHomeFeedVideos` |
| Subscribed list | `_mergeListVideos` |
| New | chronological |
| Classic | `getClassicVideos` |

They all collapse into one `home` bucket in funnelcake's `view_traffic_sources`. The daily re-serve monitor therefore cannot attribute a rate to a specific feed and — more importantly right now — **cannot show whether a fix to one of them worked**. Backend fixes just landed for the recommendations path (divinevideo/divine-funnelcake#843, #858); without this, their effect is invisible because it is averaged with four untouched feeds.

## Change

Publish the bloc's current source type as the **source detail**.

This only populates plumbing that already exists end to end:

- The kind-22236 `source` tag already carries an optional third element (`['source', <source>, <detail>]`)
- funnelcake's view handler already parses `tag[2]` into `view_traffic_sources.source_detail`
- `ViewEventPublisher` already accepts `sourceDetail`, already threaded through `FeedVideos` → `DivineVideoMetricsTracker`

Only the value was missing.

## Compatibility

**Purely additive.** `source` stays `home` for every mode, so existing queries and dashboards filtering on it are unaffected — including funnelcake's `"home" => home_views += row.views` aggregation. No funnelcake change, no migration, no schema change.

## Tags

`foryou` · `following` · `list` · `new` · `classic`

These become persisted analytics values, so the test asserts the **exact strings** rather than merely that a tag exists — renaming one silently splits its history in two. It also asserts uniqueness (a duplicate would merge two feeds back into one bucket) and the enum arity, so adding a feed mode surfaces the decision at review time instead of quietly landing views in an unattributed bucket.

## Validation

- `flutter analyze lib test integration_test` ✅ No issues found
- `flutter test test/blocs/video_feed/` ✅ 120 passed
- `video_feed_page_test.dart` ✅ 34 passed (covers the changed call site)
- New test ✅ 4 passed

## After this ships

The daily re-serve monitor can group by `source_detail` and answer what is currently unanswerable: which of the five home feeds is re-serving, and whether the recommendations fixes moved their slice. Worth having before the 2026-08-18 campaign — it is the difference between measuring the feed and guessing at it.

Note this is instrumentation only: it changes no feed behaviour and carries no rollout risk.

### DIFF
--- a/mobile/lib/blocs/video_feed/video_feed_state.dart
+++ b/mobile/lib/blocs/video_feed/video_feed_state.dart
@@ -36,6 +36,27 @@ enum VideoFeedSourceType {
   classic,
 }
 
+extension VideoFeedSourceTypeAnalytics on VideoFeedSourceType {
+  /// Stable analytics tag for this source type.
+  ///
+  /// Published as the third element of the kind-22236 `source` tag, which
+  /// funnelcake stores in `view_traffic_sources.source_detail`. The `source`
+  /// itself stays `home` for every mode, because a single feed page renders
+  /// all of them — without this detail every mode collapses into one `home`
+  /// bucket and re-serve rates cannot be attributed to a specific feed.
+  ///
+  /// These values are persisted analytics data. Renaming one silently splits
+  /// its history in two, so add new values rather than repurposing existing
+  /// ones.
+  String get analyticsTag => switch (this) {
+    VideoFeedSourceType.forYou => 'foryou',
+    VideoFeedSourceType.following => 'following',
+    VideoFeedSourceType.subscribedList => 'list',
+    VideoFeedSourceType.newVideos => 'new',
+    VideoFeedSourceType.classic => 'classic',
+  };
+}
+
 /// Source selection for [VideoFeedBloc].
 final class VideoFeedSource extends Equatable {
   /// Personalized recommended videos.
--- a/mobile/lib/screens/feed/video_feed_page.dart
+++ b/mobile/lib/screens/feed/video_feed_page.dart
@@ -502,6 +502,11 @@ class _VideoFeedViewState extends ConsumerState<VideoFeedView>
                             hasMore: state.hasMore,
                             isLoadingMore: state.isLoadingMore,
                             trafficSource: ViewTrafficSource.home,
+                            // Every feed mode renders through this one page and
+                            // reports source=home, so without this detail the
+                            // For You, following, list, new and classic feeds
+                            // are indistinguishable in view_traffic_sources.
+                            sourceDetail: state.source.type.analyticsTag,
                             onActiveVideoChanged: (video, index) {
                               final isTuningAutoAdvance =
                                   _pendingTuningAutoAdvanceIndex == index;
--- a/mobile/test/blocs/video_feed/video_feed_source_analytics_test.dart
+++ b/mobile/test/blocs/video_feed/video_feed_source_analytics_test.dart
@@ -1,0 +1,53 @@
+// ABOUTME: Locks the analytics tags published as view_traffic_sources.source_detail
+// ABOUTME: so renaming one cannot silently split its history in the funnelcake data
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:openvine/blocs/video_feed/video_feed_bloc.dart';
+
+void main() {
+  group('VideoFeedSourceTypeAnalytics', () {
+    test('every source type has a distinct, stable tag', () {
+      // These strings are written to funnelcake's
+      // view_traffic_sources.source_detail and queried historically. Renaming
+      // one splits its history in two, so this asserts the exact values rather
+      // than merely that a tag exists.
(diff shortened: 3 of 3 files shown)


rabble:

I traced the "plumbing already exists" claim end to end and it holds exactly as described:

- `view_event_publisher.dart:166` emits `['source', source.tagValue, sourceDetail]` and falls back to the 2-element tag when the detail is null or empty
- `feed_videos.dart:478` forwards `sourceDetail` into `DivineVideoMetricsTracker`, which passes it on at `divine_video_metrics_tracker.dart:220`
- `view_event_retry_service.dart:117` persists it, so retried views keep their attribution rather than silently downgrading to bare `home`

I also checked that `video_feed_page.dart:504` is the **only** `ViewTrafficSource.home` call site in `lib/`, so there's no second home surface left unattributed. The additive-compatibility argument is right: the `source` element is untouched, so funnelcake's `"home" => home_views` aggregation can't notice this.

One real gap.

### The wiring this PR adds is the one thing no test covers

All four new tests exercise the extension in isolation — `forYou.analyticsTag == 'foryou'` and friends. None of them asserts that `VideoFeedPage` actually passes `state.source.type.analyticsTag` into `FeedVideos`. Delete the `sourceDetail:` line from `video_feed_page.dart` and all four still pass, `flutter analyze` stays clean, and every view silently reverts to the unattributed `home` bucket — which is precisely the bug this PR exists to fix, and it would be invisible until someone queried the data weeks later.

The description says `video_feed_page_test.dart` "covers the changed call site", but I grepped `mobile/test/` and no test there references `sourceDetail` at all. Those 34 tests execute the line; nothing asserts its value.

There's already a precedent for the assertion in this repo:

```dart
// mobile/test/screens/feed/pooled_fullscreen_video_feed_screen_test.dart:583
expect(feedVideos.sourceDetail, 'npub-profile');
```

The equivalent here — pump `VideoFeedView` with a bloc state per source type, find the `FeedVideos` widget, assert `sourceDetail` matches — is what makes the four tag-value tests worth having. Right now they lock down a constant that nothing proves is reachable. Given the whole point is to have trustworthy attribution before 2026-08-18, I'd want this one before merge.

### Worth a decision, not a change

**`subscribedList` collapses to `'list'`, losing which list.** Every other `sourceDetail` call site carries the discriminating value — `hashtag_feed_screen.dart:223` passes the hashtag, `videos_section.dart:106` passes the query. So `list` is the one mode where the detail still can't answer "which one". I assume that's deliberate (unbounded cardinality in `view_traffic_sources`, and a list ID is closer to user data than a hashtag is), and I think it's the right call — but the doc comment is the place to say so, since the next person to look at this will read the asymmetry as an oversight and "fix" it.

### Nits

- **Test file doesn't mirror a `lib/` path.** The extension lives in `lib/blocs/video_feed/video_feed_state.dart`, so the mirror is `video_feed_state_test.dart`. The three existing files in that directory (`home_feed_cache_test`, `home_feed_resume_manager_test`, `video_feed_bloc_test`) all mirror their source. Not worth a re-push on its own, but worth folding in if you touch this again.
- **`values.length == 5` is largely redundant** with the exhaustive switch — as its own comment concedes, a new enum value is already a compile error. It survives as a review tripwire rather than a regression test, which is a weaker justification than the other three have. Fine to keep, just the least load-bearing of the four.

The reasoning in the description is solid and the change itself is about as low-risk as instrumentation gets. It's the missing wiring assertion that I'd fix before this goes in.

rabble:
## Performance implications: none

Flagging this explicitly since the related backend PRs are being gated on performance.

This is **instrumentation only**:

- No feed behaviour changes — no filtering, no reordering, no extra requests
- No additional API calls; it populates a field on an event already being published
- Server-side cost is unchanged: `source` stays `home`, the same row is written to `view_traffic_sources`, and `source_detail` is an existing column
- Wire cost is one short string per view event (`foryou` / `following` / `list` / `new` / `classic`)

No flag needed, and nothing to roll back beyond a revert.

**It is also the prerequisite for evaluating the others.** Without it, every measurement of the `home` surface is an average across five feed modes, so there is no way to tell whether a backend suppression fix moved its slice or whether traffic simply shifted between modes. Landing this first makes the other decisions measurable rather than speculative.

PR #6867, opened 2026-08-08
fix(repositories): stop an unreadable local cache from blocking relay publishes

### LINKED ISSUE #6868: fix(repositories): stop an unreadable local cache from blocking relay publishes
Auto-created from PR #6867.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6867
**Branch:** `fix/repository-write-coupling`

rabble:
Closes #6868

## Description

A user on 1.0.19+832 could not repost. The report showed the Kind 16 was never
even attempted: `RepostsRepository.repostVideo` called `_ensureInitialized()`
(a local-storage read) and awaited a placeholder write **before** reaching
`sendGenericRepost`, so a storage failure threw before any network I/O. The
same shape existed in `LikesRepository.likeEvent` / `unlikeEvent`.

The trigger on that device was an unopenable Drift database — every statement
failed `SqliteException(26)` — but the coupling is the bug independent of what
breaks storage. A signed Nostr event that reaches relays is irreplaceable; a
local cache row is not. Sacrificing the former to protect the latter inverts
the priority.

Made worse by two amplifiers:

- `_isInitialized` never latched on the throwing path, so every tap re-ran the
  failing read (26 identical failures for two taps in the report).
- The designated fallback, `_queueOfflineAction` → `PendingActionService`, is
  backed by the *same* database, and its `try` only wrapped the network publish
  — so the recovery path was both unreachable and non-functional.

Storage side effects on the publish paths are now best-effort (logged and
swallowed via a shared `_bestEffortLocalStorage` helper), and
`_ensureInitialized` degrades to an in-memory-only cache and latches on both
paths. Expected IO/corruption failures stay out of Crashlytics per the decision
matrix in `.claude/rules/error_handling.md`.

`FollowRepository` already had the correct ordering — it broadcasts before
saving, and its storage layer swallows its own errors — so it needed no change
and served as the in-repo model.

**Related Issue:** Relates to #

## Out of Scope

- The startup bug that left the database unopenable in the first place
  (`DatabaseEncryptionBootstrap` hands back a `null` cipher key when
  `migratePlaintextToEncrypted` cannot classify the file, so an encrypted DB is
  opened as plaintext). That is a separate root cause with a separate fix and
  will be its own PR — it does not depend on this one, and this one does not
  depend on it.
- `executeLikeAction` / `executeRepostAction` (the PendingActionService replay
  paths). Those exist to drain a DB-backed queue, so a dead DB legitimately
  means there is nothing to replay.

## Verification

- `flutter test` in `mobile/packages/reposts_repository` — 129 passed
- `flutter test` in `mobile/packages/likes_repository` — 206 passed
- Coverage: reposts 100.00% (no `min_coverage` set, so the VeryGood 100%
  default applies); likes 96.89% against its gate of 62
- `flutter analyze lib test` clean in both packages
- App-layer tests depending on either repository (19 files) — 255 passed,
  4 skipped
- New tests fail against the pre-fix code at the exact lines named above
  (`reposts_repository.dart:263/288/430`, `likes_repository.dart:449/486`)

## Type of Change

### DIFF
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -1502,11 +1502,20 @@ class LikesRepository {
   ///
   /// Throws `SyncFailedException` if syncing fails.
   Future<LikesSyncResult> syncUserReactions() async {
-    // First, load from local storage (fast)
-    if (_localStorage != null) {
-      final records = await _localStorage.getAllLikeRecords();
-      records.forEach(_indexLikeRecord);
-      _emitLikedIds();
+    // First, load from local storage (fast). Best-effort and deliberately
+    // outside the relay try/catch below: an unreadable cache must not skip
+    // the authoritative relay fetch, which is the whole point of a sync.
+    final localStorage = _localStorage;
+    if (localStorage != null) {
+      final records = await _bestEffortLocalStorage(
+        localStorage.getAllLikeRecords,
+        description: 'loading persisted reactions for sync',
+        site: LikesRepositoryReportableSites.syncUserReactionsLoadRecords,
+      );
+      if (records != null) {
+        records.forEach(_indexLikeRecord);
+        _emitLikedIds();
+      }
     }
 
     try {
@@ -1632,7 +1641,11 @@ class LikesRepository {
       // Remove deleted likes from cache and storage
       for (final targetId in deletedTargetIds) {
         _deindexLikeRecord(targetId);
-        await _localStorage?.deleteLikeRecord(targetId);
+        await _bestEffortLocalStorage(
+          () async => localStorage?.deleteLikeRecord(targetId),
+          description: 'dropping a deleted reaction',
+          site: LikesRepositoryReportableSites.syncUserReactionsDeleteRecord,
+        );
       }
 
       // Remove deleted downvotes from in-memory cache (no local storage in v1)
@@ -1642,8 +1655,12 @@ class LikesRepository {
       }
 
       // Batch save new records to storage
-      if (newRecords.isNotEmpty && _localStorage != null) {
-        await _localStorage.saveLikeRecordsBatch(newRecords);
+      if (newRecords.isNotEmpty && localStorage != null) {
+        await _bestEffortLocalStorage(
+          () => localStorage.saveLikeRecordsBatch(newRecords),
+          description: 'persisting the synced reactions',
+          site: LikesRepositoryReportableSites.syncUserReactionsSaveBatch,
+        );
       }
 
       _emitLikedIds();
@@ -1929,15 +1946,24 @@ class LikesRepository {
   Future<void> initialize() async {
     if (_isInitialized) return;
 
-    // Load from local storage first for immediate UI display
+    // Load from local storage first for immediate UI display. Best-effort:
+    // throwing here would skip _subscribeToReactions below, so an unreadable
+    // cache would also cost the session its cross-device sync.
     var hasCoordinatelessRecords = false;
-    if (_localStorage != null) {
-      final records = await _localStorage.getAllLikeRecords();
-      records.forEach(_indexLikeRecord);
-      hasCoordinatelessRecords = records.any(
-        (r) => r.addressableId == null || r.addressableId!.isEmpty,
+    final localStorage = _localStorage;
+    if (localStorage != null) {
+      final records = await _bestEffortLocalStorage(
+        localStorage.getAllLikeRecords,
+        description: 'loading persisted reactions at startup',
+        site: LikesRepositoryReportableSites.initializeLoadRecords,
       );
-      _emitLikedIds();
+      if (records != null) {
+        records.forEach(_indexLikeRecord);
+        hasCoordinatelessRecords = records.any(
+          (r) => r.addressableId == null || r.addressableId!.isEmpty,
+        );
+        _emitLikedIds();
+      }
     }
 
     final pubkey = await _currentUserPubkey();
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -2042,7 +2042,18 @@ class LikesRepository {
       );
 
       _indexLikeRecord(record);
-      unawaited(_localStorage?.saveLikeRecord(record));
+      // Best-effort: `unawaited` attaches no error handler, so a raw
+      // rejection here would surface as an uncaught async error in the app's
+      // zone guard and reach Crashlytics as exactly the expected-IO noise the
+      // decision matrix keeps out.
+      unawaited(
+        _bestEffortLocalStorage(
+          () async => _localStorage?.saveLikeRecord(record),
+          description: 'saving a reaction from the live subscription',
+          site:
+              LikesRepositoryReportableSites.processIncomingReactionSaveRecord,
+        ),
+      );
       _emitLikedIds();
     } else if (event.content == _downvoteContent) {
       // Deduplicate downvotes (in-memory only — no local persistence in v1)
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -483,7 +483,10 @@ class LikesRepository {
       addressableId: addressableId,
     );
     _indexLikeRecord(placeholder);
-    await _localStorage?.saveLikeRecord(placeholder);
+    await _bestEffortLocalStorage(
+      () async => _localStorage?.saveLikeRecord(placeholder),
+      description: 'saving the like placeholder',
+    );
     if (previousCount != null) {
       _writeCachedLikeCount(
         eventId,
@@ -533,7 +536,10 @@ class LikesRepository {
         addressableId: addressableId,
       );
       _indexLikeRecord(confirmed);
-      await _localStorage?.saveLikeRecord(confirmed);
+      await _bestEffortLocalStorage(
+        () async => _localStorage?.saveLikeRecord(confirmed),
+        description: 'saving the confirmed like',
+      );
 
       return reactionEvent.id;
     } catch (e, stackTrace) {
@@ -555,7 +561,10 @@ class LikesRepository {
         return placeholderId;
       }
       _deindexLikeRecord(eventId);
-      await _localStorage?.deleteLikeRecord(eventId);
+      await _bestEffortLocalStorage(
+        () async => _localStorage?.deleteLikeRecord(eventId),
+        description: 'rolling back the like placeholder',
+      );
       if (previousCount != null) {
         _writeCachedLikeCount(
           eventId,
@@ -671,13 +680,17 @@ class LikesRepository {
     if (record == null && addressableId != null && addressableId.isNotEmpty) {
       record = _likeRecordsByAddressableId[addressableId];
     }
-    if (record == null && _localStorage != null) {
-      record = await _localStorage.getLikeRecord(eventId);
-      if (record == null && addressableId != null && addressableId.isNotEmpty) {
-        record = await _localStorage.getLikeRecordByAddressableId(
-          addressableId,
-        );
-      }
+    final localStorage = _localStorage;
+    if (record == null && localStorage != null) {
+      record = await _bestEffortLocalStorage<LikeRecord?>(
+        () async {
+          final stored = await localStorage.getLikeRecord(eventId);
+          if (stored != null) return stored;
+          if (addressableId == null || addressableId.isEmpty) return null;
+          return localStorage.getLikeRecordByAddressableId(addressableId);
+        },
+        description: 'reading the like record',
+      );
     }
 
     if (record == null) {
@@ -699,7 +712,10 @@ class LikesRepository {
     // 1. Optimistic-first: remove from memory + local storage and tick the
     // watchLikedEventIds stream BEFORE any network I/O (mirror of likeEvent).
     _deindexLikeRecord(resolvedEventId);
-    await _localStorage?.deleteLikeRecord(resolvedEventId);
+    await _bestEffortLocalStorage(
+      () async => _localStorage?.deleteLikeRecord(resolvedEventId),
+      description: 'deleting the like record',
+    );
     _decrementLikeCountCache(eventId, addressableId: addressableId);
     _emitLikedIds();
 
@@ -747,7 +763,10 @@ class LikesRepository {
         return;
       }
       _indexLikeRecord(snapshotRecord);
-      await _localStorage?.saveLikeRecord(snapshotRecord);
+      await _bestEffortLocalStorage(
+        () async => _localStorage?.saveLikeRecord(snapshotRecord),
+        description: 'restoring the like record',
+      );
       if (previousCount != null) {
         _writeCachedLikeCount(
           eventId,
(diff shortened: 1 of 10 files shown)


rabble:

The diagnosis and the priority argument are right, and the write paths are correctly fixed: the `_bestEffortLocalStorage` wrapper is applied at every storage side effect on both publish paths, the local-variable capture before the null check is done properly (`final localStorage = _localStorage`), and latching `_isInitialized` on the throwing path does kill the 26-retries-per-tap amplifier. `LikesRepository` degrades end to end after this change.

`RepostsRepository` doesn't, and I think it's a blocker.

### Blocker — the repost UI stream still reads straight from the dead database

```dart
// reposts_repository.dart:130
Stream<Set<String>> watchRepostedAddressableIds() {
  // If we have local storage, delegate to its reactive stream
  if (_localStorage != null) {
    return _localStorage.watchRepostedAddressableIds();
  }
  return _repostedIdsController.stream;
}
```

`_localStorage` is non-null on the affected device — it's *unreadable*, not absent. So this returns the Drift-backed stream, which is the thing throwing `SqliteException(26)` on every statement. Meanwhile every `_emitRepostedIds()` call the PR comments describe as "tick the `watchRepostedAddressableIds` stream BEFORE any network I/O. The UI…" is feeding `_repostedIdsController`, a `BehaviorSubject` that **nobody is subscribed to in this configuration**.

That stream is not incidental — `video_interactions_bloc.dart:149` (the repost button) and `profile_reposted_videos_bloc.dart:420` both consume it. So the post-fix behaviour on that device is: the Kind 16 publishes (the actual bug, genuinely fixed), and the button still can't reflect it.

Note also that this method never calls `_ensureInitialized()`, so it doesn't even reach the new degradation logic.

The correct shape is already in the sibling package this same PR touches:

```dart
// likes_repository.dart:295
Stream<List<String>> watchLikedEventIds() async* {
  // The public stream follows the repository cache, not direct storage
  // watches, so it stays consistent with isLiked/getOrderedLikedEventIds.
  await _ensureInitialized();
  yield* _likedIdsController.stream;
}
```

Likes deliberately follows the repository cache and therefore degrades correctly for free. Reposts is the odd one out. It's also internally inconsistent already — `getRepostedAddressableIds()` (line 139) reads `_repostRecords` and degrades fine, while the stream next to it doesn't.

Cheapest fix consistent with the rest of the PR: have `_ensureInitialized` record that storage is unusable, and have `watchRepostedAddressableIds` await it and fall back to `_repostedIdsController` when degraded. Making it unconditionally follow the cache (matching likes) is cleaner still, but that's a behaviour change for the healthy path and deserves its own justification.

### Should fix — `on Object catch` swallows programming errors too

Both `_bestEffortLocalStorage` and the new `_ensureInitialized` guard catch `Object`, which takes `Error` along with `Exception`. `.claude/rules/error_handling.md` puts `StateError` / `TypeError` / `RangeError` firmly in the **YES, reportable** column — they're invariant violations, not expected IO. As written, a genuine bug in the storage layer (a bad null-assert, a cast failure) is logged at `warning` and never surfaces anywhere.

The PR says expected IO/corruption failures stay out of Crashlytics per that matrix, and that part is right — but `on Object` doesn't make the distinction the matrix is built on. These are pure-Dart packages, so the documented route is the reporter-port pattern (`dm_repository` is the canonical reference in the same rules file): keep swallowing IO/corruption, and route `Error`s through an injected reporter.

### Smaller

- **An unreadable cache now reports as "definitely not reposted."** In `unrepostVideo`, a swallowed read failure leaves `record == null`, which throws `NotRepostedException` — an assertion that the user hasn't reposted, when the truth is that we can't tell. Same shape on the likes side. Given this PR's own argument that the relay event is the irreplaceable artifact, there's a case for still attempting the deletion when the cache is unreadable rather than declining on cache evidence that doesn't exist. At minimum the distinction deserves a comment, since a caller catching `NotRepostedException` will now do the wrong thing on a broken DB.
- **Latching on transient failures.** "An unreadable database stays unreadable for the session" holds for corruption, which is the reported case, but not for a transient lock or IO blip — those now degrade the whole session to in-memory-only with records sitting readable on disk. Probably the right trade, and definitely better than the 26-retry loop; just noting the comment claims more than the code can know.
- **Test coverage of the gap.** The new tests confirm the publish survives a throwing storage layer, which is the right assertion. Nothing covers what the UI stream does in that state — which is why the blocker above got through. Worth a test that subscribes to `watchRepostedAddressableIds()` with a storage double that throws.

The out-of-scope split (leaving `DatabaseEncryptionBootstrap` to its own PR) is the right call and well argued — no notes there. Fix the stream and this is in good shape.

rabble:
Pushed `d5b440a85` addressing the review findings.

**The blocker.** `RepostsRepository` now carries a `_storageDegraded` latch and `watchRepostedAddressableIds()` became `async*`: it awaits `_ensureInitialized()` and yields from `_repostedIdsController` when storage is degraded, from the Drift stream otherwise. The healthy path is unchanged.

Two deliberate deviations from what the review suggested:

- **The latch is set from `_bestEffortLocalStorage`'s catch as well as `_ensureInitialized`'s.** A store whose writes fail is as dead as one whose reads fail, and a device where the initial read happens to succeed but writes don't is the same bug.
- **Known limit, worth stating plainly:** the stream source is chosen at subscription time, so a mid-session degradation only helps *subsequent* subscribers (a newly mounted feed item), not one already attached to the dead Drift stream. Closing that would mean making the source switchable mid-stream, which is a bigger change than this fix warrants.

Also checked whether the full likes-style alignment (always follow the controller) would be safe: the only other writer of `personal_reposts` is the logout purge at `social_providers.dart:768`, and the provider is rebuilt on auth change anyway, so the Drift stream looks non-load-bearing. That's a defensible follow-up, not something to smuggle into a review fix.

**Reportable errors.** Reporter-port typedef per the `dm_repository` precedent in both packages, with `*_reportable_sites.dart` constants and wiring in `video_providers.dart` to `CrashReportingService`. The discriminator is `if (error is! Error) return;` — `SqliteException`/`IOException` stay swallowed silently, `StateError`/`TypeError`/`RangeError` route to the reporter, which is the rules matrix and stays correct for new `Error` subtypes.

**`NotRepostedException` — documented rather than changed.** A swallowed read failure and a genuine miss both end with no repost event id, so neither can produce a Kind 5; the outcome is identical. A distinct exception would propagate a distinction through `VideoInteractionsBloc` and every other caller with no recovery available for it. Left a comment at both throw sites instead — the behavioural change deserves its own PR and a decision about what the UI would say.

One pre-existing test changed: `delegates to local storage when available` asserted `stream == storageStream` (object identity, which no wrapping can survive). Rewritten as a behavioural `expectLater(..., emits(...))` plus the same `verify`, so the healthy path stays pinned.

reposts 133 pass at 100% coverage (425/425), likes 209 at 100% (777/777), `flutter analyze lib test integration_test` clean.

mobile/packages/reposts_repository/lib/src/reposts_repository.dart
```diff
@@ -285,7 +337,11 @@ class RepostsRepository {
       createdAt: DateTime.now(),
     );
     _repostRecords[addressableId] = placeholder;
-    await _localStorage?.saveRepostRecord(placeholder);
+    await _bestEffortLocalStorage(
```
realmeylisdev, line 340: **blocker.** this write is guarded now, but on the reported device the tap never reaches it.

the repost button goes through `toggleRepost`, not `repostVideo` (`video_interactions_bloc.dart:452`), and `toggleRepost` still reads storage unguarded:

```dart
// reposts_repository.dart:637
final isCurrentlyReposted =
    await _localStorage?.isReposted(addressableId) ??
    _repostRecords.containsKey(addressableId);
```

`DbRepostsLocalStorage.isReposted` runs a real `select` (`personal_reposts_dao.dart:163`), so on a store where every statement fails it throws here, `_publishRepost` catches it into `_RepostSettleFailed`, the button snaps back (`video_interactions_bloc.dart:510`), and the kind 16 is never attempted. against a storage double whose every method throws:

```
REPRO A: direct repostVideo publishes on a dead database ......... ok
REPRO B: toggleRepost (the UI path) publishes on a dead db ....... FAILED
  Exception: SqliteException(26): file is not a database
  package:reposts_repository/src/reposts_repository.dart 638:30
    RepostsRepository.toggleRepost
```

the `??` already spells the fallback, so routing the read through the existing helper is enough (needs one new `RepostsRepositoryReportableSites` constant, and an explicit type arg since the operation is nullable):

```dart
final isCurrentlyReposted =
    await _bestEffortLocalStorage<bool?>(
      () async => _localStorage?.isReposted(addressableId),
      description: 'checking the repost state',
      site: RepostsRepositoryReportableSites.toggleRepostReadState,
    ) ??
    _repostRecords.containsKey(addressableId);
```

worth a test at the `toggleRepost` level too. every new test here calls `repostVideo` directly, which is why the gap held.

mobile/packages/likes_repository/lib/src/likes_repository.dart
```diff
@@ -483,7 +514,11 @@ class LikesRepository {
       addressableId: addressableId,
     );
     _indexLikeRecord(placeholder);
-    await _localStorage?.saveLikeRecord(placeholder);
+    await _bestEffortLocalStorage(
```
realmeylisdev, line 517: **blocker**, same shape on the likes side. the video like button calls `toggleLike` (`video_interactions_bloc.dart:325`), and `toggleLike` reads storage unguarded before dispatching:

```dart
// likes_repository.dart:912
final likedByEventId =
    await _localStorage?.isLiked(eventId) ??
    _likeRecords.containsKey(eventId);
```

same double, same result:

```
REPRO A: direct likeEvent publishes on a dead database ......... ok
REPRO B: toggleLike (the UI path) publishes on a dead db ....... FAILED
  Exception: SqliteException(26): file is not a database
  package:likes_repository/src/likes_repository.dart 913:30
    LikesRepository.toggleLike
```

note comment reactions *are* fixed by this PR, since `comment_reactions_bloc.dart:261/286` call `likeEvent` / `unlikeEvent` directly. it's only the two video buttons that still go through the toggles.

same fix as the reposts side: wrap the read, keep the `??` fallback to `_likeRecords`. the coordinate check just below it is memory-only already, so it needs nothing.

mobile/packages/reposts_repository/lib/src/reposts_repository.dart
```diff
@@ -993,16 +1076,87 @@ class RepostsRepository {
...
-        _repostRecords[record.addressableId] = record;
+    final localStorage = _localStorage;
+    if (localStorage != null) {
+      try {
+        final records = await localStorage.getAllRepostRecords();
+        for (final record in records) {
+          _repostRecords[record.addressableId] = record;
+        }
+        _emitRepostedIds();
+      } on Object catch (e, stackTrace) {
```
realmeylisdev, line 1144: **should-fix**, same class as this PR's thesis. `syncUserReposts` reads storage at `:675`, *outside* the `try` that opens at `:689`:

```dart
Future<RepostsSyncResult> syncUserReposts() async {
  if (_localStorage != null) {
    final records = await _localStorage.getAllRepostRecords();  // :675, unguarded
    ...
  }
  // the "fetch from relays (authoritative)" half below never runs
  try { ... } catch (e) { ... throw SyncFailedException(...); }
```

so a degraded store throws the raw storage exception (not even `SyncFailedException`), the authoritative relay half is skipped entirely, and `ProfileRepostedVideosBloc` catches it at `:144` and emits `failure` at `:158`. the profile Reposts tab shows an error while relays are perfectly reachable.

`syncUserReactions` has the identical shape at `likes_repository.dart:1492`. `initialize()` is the same story at `:966` / `:1908`, and there it also means `_subscribeToReposts()` / `_subscribeToReactions()` never run, so cross-device sync is dead for the session as well.

mobile/packages/reposts_repository/lib/src/reposts_repository.dart
```diff
@@ -993,16 +1076,87 @@ class RepostsRepository {
...
+  ///
+  /// Expected IO/corruption failures are not Crashlytics-reportable (see the
+  /// decision matrix in `.claude/rules/error_handling.md`); the unified log is
+  /// where a degraded cache surfaces. Programming-invariant violations are
+  /// routed to [_errorReporter] instead of vanishing — see
+  /// [_markStorageDegraded].
+  ///
+  /// [site] is the stable `RepostsRepositoryReportableSites` identifier for
+  /// this call site; [description] is the human phrasing used in the log.
+  Future<T?> _bestEffortLocalStorage<T>(
```
realmeylisdev, line 1179: **nit.** two write sites still bypass this helper: `_processIncomingRepost` at `:1042` and `_processIncomingReaction` at `likes_repository.dart:1995` both do `unawaited(_localStorage?.save...(record))`.

`unawaited` attaches no error handler, so once a store degrades mid-session every cross-device event's rejection is an unhandled async error, lands in `runZonedGuarded` (`main.dart:1630`) and goes straight to `CrashReportingService.recordError` as `reason: 'runZonedGuarded'`. that's exactly the expected-IO noise the decision-matrix argument in this PR is trying to keep out of the dashboard.

wrapping both in `_bestEffortLocalStorage` closes it and gets them the `_storageDegraded` latch for free.

realmeylisdev (dismissed):
the swallow helper itself is right, and it catches async rejections too, not just the sync throws the tests exercise. verified that separately.

the problem is the entry point. the repost and like buttons call `toggleRepost` / `toggleLike`, not `repostVideo` / `likeEvent`, and both of those still read local storage unguarded before any publish, so on the device from the report the tap still throws before the kind 16 goes out. repro inline.

two smaller same-class ones inline too. also ran the branch on ios, no regressions there.

mobile/packages/reposts_repository/lib/src/reposts_repository.dart
realmeylisdev: fixed in `df631e99d`. `toggleRepost`'s read goes through `_bestEffortLocalStorage<bool?>` now, falling back to `_repostRecords` via the existing `??`. two tests at the `toggleRepost` level cover both directions, including that a swallowed read still unreposts from the cache instead of publishing a duplicate.

mobile/packages/likes_repository/lib/src/likes_repository.dart
realmeylisdev: fixed in `df631e99d`, same shape for `toggleLike`. two matching tests on the likes side.

mobile/packages/reposts_repository/lib/src/reposts_repository.dart
realmeylisdev: fixed in `eecd83e0c`. `syncUserReposts` / `syncUserReactions` now read through the guard, so the relay fetch runs on a dead cache, and both `initialize()`s do too, which also keeps `_subscribeToReposts` / `_subscribeToReactions` reachable. the sync's own writes (batch save, deleted-record purge) are wrapped as well since they sat inside the same try.

mobile/packages/reposts_repository/lib/src/reposts_repository.dart
realmeylisdev: fixed in `f81476d63`. both `unawaited(...)` writes now wrap `_bestEffortLocalStorage`, so the rejection is swallowed instead of reaching the zone guard. covered by a test that feeds an event through the subscription with a throwing store.

realmeylisdev (approved):
pushed the three fixes instead of handing them back, one commit each so you can revert any of them alone.

`df631e99d` routes `toggleRepost` / `toggleLike`'s storage read through the guard. those are what the buttons actually call, so this is the one that makes the kind 16 / kind 7 survive a dead cache.
`eecd83e0c` does the same for `syncUserReposts` / `syncUserReactions` and both `initialize()`s, so a bad cache stops skipping the relay half and the cross-device subscription.
`f81476d63` wraps the two `unawaited` subscription writes.

reposts 139 pass at 100% coverage, likes 216 at 97.07% against its gate of 62, the 19 app-layer suites 255 pass / 4 skip, `flutter analyze lib test integration_test` clean.

heads up that `Tests (shard 2/4)` is red and it is not from this branch. main's own latest run (`31246035711`, head `95a05362f`) fails the same shard on `ClipThumbnailImage` + `PopularVideosTab`, and my two attempts here failed disjoint sets of those same tests. all of them pass standalone. it's the merged-isolate seed thing, and nothing in these commits touches `mobile/test` or `mobile/lib`.

realmeylisdev:
Ran this on an iOS device end to end, before and after, against a database in the exact state from the report.

Reproducing it turned out to need the specific path the PR calls out as out of scope: `_classifyDatabase` only treats `SQLITE_NOTADB` as "encrypted", so a file that fails with anything else (I used a valid header over a malformed page 1) classifies `indeterminate`, `migratePlaintextToEncrypted` returns `failed`, and `_resolveCipherKey` hands back a null key. The app then opens the encrypted file as plaintext and every statement fails. Log line is `At-rest DB migration did not complete; opening plaintext this launch and retrying next launch.` Byte-corrupting the file does not reproduce it, `_recoverUnusableEncryptedDatabase` catches that at startup and recreates the DB.

Same device, same broken DB, same tap on the repost button:

**At `e24412413`**

```
VideoInteractionsBloc: Repost toggle failed for cd7e1669… - SqliteException(11)
Bloc error: Reportable<DriftRemoteException>(_publishRepost): SqliteException(11)
```

Nothing published. Button snapped back to `Repost video`. Confirms the reported bug was still live on the branch as reviewed, because the button goes through `toggleRepost`, not `repostVideo`.

**At `f81476d63`**

```
Local repost storage unavailable (checking the repost state); continuing without it: SqliteException(11)
Local repost storage unavailable (saving the repost placeholder); continuing without it: SqliteException(11)
Local repost storage unavailable (saving the confirmed repost); continuing without it: SqliteException(11)
```

Kind 16 `8943f9ca864bc535742479121dd34c53a7a306c5dd68053127042f8518c624b7` landed on `relay.divine.video` at 19:36:24 and the button flipped to `Remove repost`. Startup also logged `Local like storage unavailable (loading persisted reactions at startup); continuing without it`, which is the `initialize()` guard keeping `_subscribeToReactions` reachable.

Healthy-DB path re-checked on the same build too: unrepost published its kind 5, repost published kind 16 `07b1bd174bef47b4aee928e580f2896cf65f5dff7980db2b04f0fd2ce5de3e54`, button tracked both.

One thing I could not pin down and am not claiming as a finding: on the healthy DB the very first unrepost tap published its kind 5 but the button stayed on `Remove repost`, and the next tap published a second kind 5 for the same target id before it flipped. It did not reproduce on any later tap and I have no mechanism for it, so treat it as noise unless someone else sees it.

PR #6869, opened 2026-08-08
feat(seen): client-side seen-video filtering across feeds

rabble:
## Summary

- Persist seen-video history in Drift while keeping the bounded SharedPreferences metrics cache and migration fallback.
- Filter or demote recently seen videos according to feed semantics; classic discovery deep-fetches up to eight pages for unseen results.
- Keep profile feeds unchanged and retain a feature-flag kill switch.
- Await durable history writes and make batch upserts refresh `last_seen_at` while preserving `first_seen_at`.

## Regression coverage

27 focused tests cover:

- home and new-feed prioritization
- classic filtering, deep-fetch pagination, cache refresh, and the all-seen fallback
- profile and no-lookup behavior
- SharedPreferences-to-Drift migration and fallback
- awaited single and batch persistence
- DAO insert, true upsert, recency, pruning, removal, and clearing

No coverage exclusions are used for the changed behavior.

## Verified locally

- `videos_repository`: 471 tests passed; 1201/1201 lines (100.00%)
- `db_client`: 914 tests passed; 6464/14859 lines (43.50%, above the 40% package gate)
- seen-video service and migration tests: 21 passed
- video-feed BLoC tests: 103 passed
- `flutter analyze lib test integration_test`: clean
- generated-code verification and the pre-push changed-test gate: clean

### DIFF
--- a/mobile/lib/features/feature_flags/models/feature_flag.dart
+++ b/mobile/lib/features/feature_flags/models/feature_flag.dart
@@ -91,6 +91,11 @@ enum FeatureFlag {
         'Off by default until the push service fans out kind 34236 to '
         'subscribers — with it off, the bell would publish a subscription '
         'that nothing ever delivers.',
+  ),
+  clientSeenFiltering(
+    'Client Seen-Video Filtering',
+    'Demote recently-seen videos in home feeds and drop them from '
+        'classics/discovery. Kill-switch for the Aug 2026 campaign load.',
   );
 
   const FeatureFlag(
--- a/mobile/lib/providers/video_providers.dart
+++ b/mobile/lib/providers/video_providers.dart
@@ -3,10 +3,13 @@
 
 import 'dart:async';
 
+import 'package:db_client/db_client.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart' show Provider;
 import 'package:http/http.dart' as http;
 import 'package:likes_repository/likes_repository.dart';
 import 'package:openvine/extensions/video_event_extensions.dart';
+import 'package:openvine/features/feature_flags/models/feature_flag.dart';
+import 'package:openvine/features/feature_flags/providers/feature_flag_providers.dart';
 import 'package:openvine/l10n/current_app_l10n.dart';
 import 'package:openvine/providers/auth_providers.dart';
 import 'package:openvine/providers/creator_sync_provider.dart';
@@ -34,6 +37,7 @@ import 'package:openvine/services/dead_media_feed_guard.dart';
 import 'package:openvine/services/event_api_client.dart';
 import 'package:openvine/services/event_router.dart';
 import 'package:openvine/services/nsfw_content_filter.dart';
+// ignore: unnecessary_import
 import 'package:openvine/services/pending_action_service.dart';
 import 'package:openvine/services/personal_event_cache_service.dart';
 import 'package:openvine/services/seen_videos_service.dart';
@@ -114,9 +118,20 @@ PersonalEventCacheService personalEventCacheService(Ref ref) {
 }
 
 /// Seen videos service for tracking viewed content.
+///
+/// Split storage: seen set (id+lastSeen) in Drift `seen_videos` table
+/// (unbounded, ~1yr TTL) + bounded metrics blob. DB injected when
+/// available; tests get the pure-SharedPreferences fallback.
 @Riverpod(keepAlive: true)
 SeenVideosService seenVideosService(Ref ref) {
-  final service = SeenVideosService();
+  AppDatabase? db;
+  try {
+    db = ref.watch(databaseProvider);
+  } catch (_) {
+    db = null;
+  }
+  // ignore: invalid_use_of_visible_for_testing_member
+  final service = SeenVideosService(database: db);
   unawaited(service.initialize());
   ref.onDispose(() => unawaited(service.dispose()));
   return service;
@@ -507,6 +522,14 @@ VideosRepository videosRepository(Ref ref) {
   final moderationLabelService = ref.watch(moderationLabelServiceProvider);
   final funnelcakeClient = ref.watch(funnelcakeApiClientProvider);
   final seenVideosService = ref.watch(seenVideosServiceProvider);
+  final clientSeenFilteringEnabled = () {
+    try {
+      final flagService = ref.watch(featureFlagServiceProvider);
+      return flagService.isEnabled(FeatureFlag.clientSeenFiltering);
+    } catch (_) {
+      return true;
+    }
+  }();
   final divineHostFilterService = ref.read(divineHostFilterServiceProvider);
   final feedAspectRatioPreference = ref.watch(
     feedAspectRatioPreferenceServiceProvider,
@@ -535,10 +558,12 @@ VideosRepository videosRepository(Ref ref) {
     ),
     funnelcakeApiClient: funnelcakeClient,
     inMemoryFeedCache: InMemoryFeedCache(),
-    seenVideoLookup: SeenVideoLookup(
-      wasSeenRecently: seenVideosService.wasSeenRecently,
-      initialize: seenVideosService.initialize,
-    ),
+    seenVideoLookup: clientSeenFilteringEnabled
+        ? SeenVideoLookup(
+            wasSeenRecently: seenVideosService.wasSeenRecently,
+            initialize: seenVideosService.initialize,
+          )
+        : null,
   );
 
   // Clear the in-memory feed cache (home + per-author) on logout/account
--- a/mobile/lib/services/seen_videos_service.dart
+++ b/mobile/lib/services/seen_videos_service.dart
@@ -1,9 +1,12 @@
 // ABOUTME: Service for tracking which videos have been viewed by the user with engagement metrics
 // ABOUTME: Stores view history with timestamps, loop counts, and watch duration for smart feed ordering
+// ABOUTME: Split storage: seen set (id+lastSeen) in Drift seen_videos table (unbounded, ~1yr TTL),
+// ABOUTME: rich metrics (loops/durations) stay bounded in SharedPreferences JSON (1000).
 
 import 'dart:async';
 import 'dart:convert';
 
+import 'package:db_client/db_client.dart';
 import 'package:meta/meta.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:unified_logger/unified_logger.dart';
@@ -26,7 +29,6 @@ class SeenVideoMetrics {
   Duration totalWatchDuration;
   Duration lastWatchDuration;
 
-  /// Update metrics with new viewing session
   void updateSession({
     required DateTime timestamp,
     int? loops,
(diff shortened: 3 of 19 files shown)


rabble:
CI repair complete at `14ff6daa4`.

Fixed:
- regenerated the stale Riverpod outputs
- removed the coverage exclusions and exercised the real classic-feed/cache branches
- corrected the classic fixtures so they actually enter classic filtering
- made classic all-seen pages deep-fetch correctly while retaining a non-empty cache fallback
- awaited Drift persistence and made batch upserts refresh `last_seen_at` without losing `first_seen_at`
- added direct DAO coverage for insert, upsert, recency, pruning, remove, and clear
- isolated the upload lookup test's Hive path; the exact four-suite concurrent run with CI seed `3725262250` failed before this fix and passed afterward (34 tests)

Validation:
- all 17 GitHub checks green, including all four randomized app shards
- `videos_repository`: 471 tests, 1201/1201 lines (100.00%)
- `db_client`: 914 tests, 6464/14859 lines (43.50%; gate is 40%)
- full `flutter analyze lib test integration_test` clean
- generated files and pre-push changed-test gate clean

Review items: none filed. Escalated: none. Declined: none.

mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -903,6 +981,15 @@ class VideosRepository {
       failed = true;
     }
 
+    // Seen-filtering emptied every page we were allowed to read: serve the
+    // unfiltered pages instead. Returning an empty page here would strand the
+    // feed — VideoFeedBloc only paginates from the near-end callback of a
+    // rendered video, so a page with no videos never asks for the next one
+    // and the surface stays empty until a manual pull-to-refresh.
+    if (collected.isEmpty && unfilteredCollected.isNotEmpty) {
```
dcadenas: **This was the blocking one: an empty page here stranded the feed.**

The previous version returned `videos: []` with `hasMore: true` and a cursor, on the reasoning that the caller would fetch the next cursor. It does not. `VideoFeedBloc._onLoadMoreRequested` returns early while `state.videos` is empty, and the only thing that dispatches that event is the near-end callback of a rendered video — with nothing rendered, nothing ever asks.

So a heavy viewer whose eight deep-fetch pages were all recently seen opened Classics on an empty feed and stayed there until a manual pull-to-refresh.

The loop now also collects each page pre-filter, and falls back to those when filtering emptied everything. That is what the comment a few lines up already said would happen, and it matches the `retainAllWhenAllSeen` default the helper documents. Showing an already-seen classic is what ships today, so the fallback is the status quo rather than new behaviour.

mobile/lib/features/feature_flags/providers/feature_flag_providers.dart
```diff
@@ -66,3 +66,27 @@ bool isFeatureEnabled(Ref ref, FeatureFlag flag) {
...
+/// it would freeze this on whatever the flag read at first build — including
+/// the build default, when that first read beats the unawaited load of
+/// persisted overrides. The kill switch has to work without a restart.
+///
+/// Resolving the flag needs the whole chain, including
+/// `sharedPreferencesProvider`, which many widget tests do not override. Those
+/// tests should keep the shipped default rather than fail, so a failure to
+/// resolve falls back to on.
+@Riverpod(keepAlive: true)
+bool clientSeenFilteringEnabled(Ref ref) {
```
dcadenas: **Watching the flag service instead of its state freezes the value for the process.**

`featureFlagServiceProvider` keeps one identity and only notifies listeners; `featureFlagStateProvider` is the one that self-invalidates. Watching the service meant the first resolved value stuck, so toggling the flag at runtime did nothing, and a cold start that read the flag before the unawaited override load finished would keep the build default.

This watches the state map instead. Covered by `test/features/feature_flags/providers/feature_flag_providers_test.dart` — I confirmed that test fails against the previous form.

mobile/lib/providers/video_providers.dart
```diff
@@ -112,9 +114,26 @@ PersonalEventCacheService personalEventCacheService(Ref ref) {
...
+///
+/// The Drift half is gated on [FeatureFlag.clientSeenFiltering] as well as the
+/// filtering itself. Without that gate the kill switch would still leave the
+/// startup hydration and every per-view write in place, which is the load it
+/// exists to shed.
 @Riverpod(keepAlive: true)
 SeenVideosService seenVideosService(Ref ref) {
-  final service = SeenVideosService();
+  AppDatabase? db;
+  if (ref.watch(clientSeenFilteringEnabledProvider)) {
```
dcadenas: **The kill switch now also sheds the storage work, not just the filtering.**

Previously the flag gated the lookup and the ordering, but the service was always built with the database, so flag-off still paid the startup hydration and wrote every view to Drift. For a switch whose stated job is shedding campaign load, the hydration was the part worth shedding.

Flag-off now yields the pre-existing SharedPreferences-only behaviour end to end.

mobile/packages/db_client/lib/src/database/daos/seen_videos_dao.dart
```diff
@@ -0,0 +1,147 @@
...
+    required int firstSeenAt,
+    required int lastSeenAt,
+  }) {
+    return into(seenVideos).insert(
+      SeenVideosCompanion.insert(
+        videoId: videoId,
+        firstSeenAt: firstSeenAt,
+        lastSeenAt: lastSeenAt,
+      ),
+      onConflict: DoUpdate(
```
dcadenas: **`insertOrReplace` here was resetting `first_seen_at`.**

`recordVideoView` sources `firstSeenAt` from the bounded metrics cache. Once a video ages out of those 1000 entries, the service can only offer "now", and replacing the row wrote that over the real stored value — so rewatching something first seen six months ago dated it to today.

`upsertBatch` already had this right; `markSeen` now matches it. Nothing reads `first_seen_at` yet, so this was silent data loss rather than visible breakage.

mobile/lib/services/seen_videos_service.dart
```diff
@@ -124,57 +169,90 @@ class SeenVideosService {
...
-        _seenVideos.clear();
-        final now = DateTime.now();
-        for (final videoId in legacyList) {
-          _seenVideos[videoId] = SeenVideoMetrics(
-            videoId: videoId,
-            firstSeenAt: now,
-            lastSeenAt: now,
+      if (_effectiveDb != null) {
+        try {
+          final dbRows = await _effectiveDb!.seenVideosDao.getSeenSince(
```
dcadenas: **Hydration reads a window now rather than the whole table.**

`getAll()` pulled every row into a main-isolate map at each cold start. The table is unbounded with a one-year TTL, and the p90 figure cited in `tables.dart` is 6,458 distinct videos per 30 days — so this grew toward tens of thousands of rows on exactly the accounts the feature targets, which is the decode cost the 1000-entry cap existed to avoid.

Only recency is ever asked of the in-memory set, so it hydrates 30 days through the new `getSeenSince`, served by the `last_seen_at` index. Older history stays durable and reachable through the DAO, and `wasSeenRecently` asserts its window stays inside the hydrated one.

mobile/lib/features/feature_flags/models/feature_flag.dart
```diff
@@ -91,6 +91,12 @@ enum FeatureFlag {
         'Off by default until the push service fans out kind 34236 to '
         'subscribers — with it off, the bell would publish a subscription '
         'that nothing ever delivers.',
+  ),
+  clientSeenFiltering(
+    'Client Seen-Video Filtering',
+    'Demote recently-seen videos in the home, new and For You feeds and drop '
+        'them from the Classics feed. Explore browse grids are not filtered. '
```
dcadenas: **One deliberate non-change, flagged so it is a decision rather than an oversight.**

The description said "classics/discovery", but Explore's Classics tab reaches Funnelcake directly through `classic_vines_provider` and never sees the lookup, so only the Classics feed is actually filtered. I narrowed the wording rather than extending filtering into that provider.

That surface is a browse grid the user taps into, not an algorithmic feed, and hiding already-watched videos from a catalogue view is a different product question from keeping a feed fresh. Worth a look if you do want them consistent — happy to follow up separately.

dcadenas (approved):
The storage split is the right shape, and splitting demote-for-home from drop-for-classics per feed semantics is well judged. Rather than hand findings back, I pushed the fixes onto the branch — eight commits, each one finding.

The one that mattered: the all-seen classics path returned an empty page with `hasMore` set, expecting the feed to paginate onward. It cannot — the only load-more trigger is a rendered video's near-end callback, so a heavy viewer got an empty Classics feed with no way out but pull-to-refresh. It now serves the unfiltered pages instead, which is what the surrounding comment already promised. The rest: `first_seen_at` no longer reset by the single-write path, prefs history imported even when the table already has rows, startup hydration bounded to a recency window instead of reading the whole table, and the kill switch now sheds the Drift work and responds without a restart.

Validated on the pushed head: full `db_client` (918) and `videos_repository` (471) suites, mobile blocs and providers (763), the seen-service and migration suites, `flutter analyze lib test integration_test` clean, and all required checks green. Every new test I added I checked can actually fail. Notes on individual changes inline; the merge call is yours.

mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -281,18 +290,25 @@ class VideosRepository {
       until: until,
     );
 
-    // 2. If no list refs, return following-only result
+    // 2. If no list refs, return following-only result (with seen demotion)
     if (videoRefs.isEmpty) {
-      final result = HomeFeedResult(videos: videos);
+      final ordered = await _orderBySeenFreshness(videos);
```
realmeylisdev, line 300: **This ordering doesn't survive the second page on Following or New.**

`_usesCursorPagination` is true only for For You and Classics. Everything else goes through `updatedVideos.sort((a, b) => b.createdAt.compareTo(a.createdAt))` in `_onLoadMoreRequested` (`video_feed_bloc.dart:443`), which puts every demoted video straight back into its chronological slot.

That sort was harmless before this PR: `getHomeFeedVideos` and `getNewVideos` both returned chronological pages, and pagination only ever appends older videos, so re-sorting changed nothing. This is the first change to hand them a page that isn't sorted by `createdAt` — `getNewVideos` still documents itself as "sorted by creation time (newest first)", which is exactly the assumption the bloc leans on.

Seeded a Following feed with a demoted opening page — A(100), C(98), B(99), D(97) — and paginated once:

```
[A, B, C, D, E, F]
```

not `[A, C, B, D, E, F]`. So the demotion holds for the opening page and is gone from there on. For You is unaffected (cursor-paginated) and Classics drops rather than reorders, so it's these two feeds.

Which way to take it is a call I didn't want to make for you — skip the re-sort for these two sources, re-apply the partition after the merge, or keep it opening-page-only and narrow the flag description to match. All three move feed ordering further than a review fix should.

realmeylisdev (commented):
Storage split is the right shape, and demote-on-home/new vs drop-on-Classics matches how those feeds actually behave. Ran it on an iOS build to check the migration end to end: 25 ids moved into Drift on first launch, Classics drops what you just watched when you come back to it, no asserts, no warnings.

One gap, and it only shows on the second page. Following and New aren't cursor-paginated, so the bloc re-sorts by `createdAt` on every load-more — and the demotion goes with it. That sort was a no-op until this PR started handing those two feeds a page that isn't chronological. Repro inline.

Pushed one commit for a stale comment in `getHomeFeedVideos`. Everything else held up: `first_seen_at` survives a rewatch, the prefs import is idempotent, and the kill switch really does shed the Drift work.

realmeylisdev (approved):
Fixed the ordering gap rather than handing it back. Two commits.

**`8965dc08a`** — the demotion no longer dies at the first page boundary. Following and New now order the *arriving* page and append it instead of re-sorting the whole loaded list. An `until` page is strictly older than everything on screen, so it lands in the same chronological place it always did; what changes is that page one keeps the order the repository gave it, and nothing shifts under the viewer's index. New bloc test covers it — I checked it fails against the previous code.

**One thing to look at, because I changed a test rather than only adding one.** `sorts merged following feed chronologically on load more` stubbed a load-more page *newer* than the loaded set and asserted it was hoisted to the top. `until` pagination can't produce that page, so it only describes a server ignoring the cursor — and hoisting it over the current video is the same harm the forYou test directly above it refuses. It now asserts the page is appended. If you'd rather keep the old placement, reverting that one commit brings the demotion gap back with it.

**`ffac6d95e`** — stale comment in `getHomeFeedVideos` that pointed at a "cache gate below" sitting above it.

Ran before pushing: all 2,936 bloc tests, the 161 feed-screen tests, the full `videos_repository` suite, and `flutter analyze lib test integration_test`. All checks green on the pushed head. The rest of the PR I checked on an iOS build earlier — prefs→Drift migration ran, Classics drops what you just watched on re-entry, kill switch sheds the Drift work, no asserts. Merge call is yours.

mobile/packages/videos_repository/lib/src/videos_repository.dart
realmeylisdev: Fixed in 8965dc08a — the arriving page is ordered and appended, so page one keeps the demotion and nothing moves under the viewer's index. Regression test added; I confirmed it fails on the previous code.

realmeylisdev:
Pushed `6260197c0` — Videos Repository CI is green, all 17 checks pass.

The rebase picked up #6945, which moved `getNewVideos` onto `getRecentVideosPage` and changed its return type to `HomeFeedResult`. `videos_repository_seen_filtering_test.dart` was still written against the old API: it stubbed `getRecentVideos` and treated the result as a `List`.

Fixing the call site surfaced a second failure the analyzer error had been hiding — the mock stubbed a method the repository no longer calls, so the test compiled and then threw `type 'Null' is not a subtype of type 'Future<RecentVideosResponse>'`. The stub moved to `getRecentVideosPage` as well. The test still discriminates: it stubs seen-first and asserts unseen-first, so it fails if the demotion breaks.

Worth noting that only this job caught it — the app-level `Analyze` doesn't cover `mobile/packages/`, so the other 15 checks were green on a branch that didn't compile.

Unrelated and not fixed here: `.github/workflows/videos_repository.yaml` calls VGV `flutter_package.yml@v1` with no `flutter_version`, so CI's `dart format` floats independently of `mise.toml`'s `flutter = "3.44.0"`. CI reported the pre-existing file format-clean while the pinned toolchain reformats it. Pinning the workflow would close that gap, but it's a workflow edit and out of scope for this PR.

PR #6870, opened 2026-08-08
test(analytics): assert the home feed actually publishes its source detail

rabble:
## Description

Follow-up to #6865, which merged before this could land on its branch.

#6865 added `VideoFeedSourceTypeAnalytics.analyticsTag` and one call site — `sourceDetail: state.source.type.analyticsTag` in `video_feed_page.dart` — so the five home feed modes stop collapsing into a single unattributed `home` bucket in funnelcake's `view_traffic_sources`.

Its four tests exercise the extension in isolation (`VideoFeedSourceType.forYou.analyticsTag == 'foryou'`). None of them asserts the wiring. Delete the `sourceDetail:` line and all four still pass, `flutter analyze` stays clean, and every view silently reverts to the unattributed bucket — invisible until someone queries the data weeks later. That is the exact regression the feature exists to prevent, and it was the one thing unguarded.

## Change

`VideoFeedView`'s existing `passes home traffic attribution to FeedVideos` test asserted `trafficSource` only. It is now parameterised over `VideoFeedSourceType.values` and asserts both:

```dart
final feedVideos = tester.widget<FeedVideos>(find.byType(FeedVideos));
expect(feedVideos.trafficSource, ViewTrafficSource.home);
expect(feedVideos.sourceDetail, sourceType.analyticsTag);
```

Parameterising over the enum means a sixth feed mode is covered the moment it is added, rather than depending on someone remembering to extend a test. The existing harness (`_MockVideoFeedBloc`, `testMaterialApp`, the `MultiBlocProvider` wiring, the pump/unmount teardown) is reused unchanged; the only new helper is `_sourceFor`, needed because `subscribedList` requires `listId`/`listName`.

The replaced test is fully subsumed — the five new ones assert its `trafficSource` expectation plus the missing `sourceDetail` one — so it is removed rather than left restating a sibling.

Value literals stay pinned by the existing state tests. The two layers together pin *the literal string* and *that the literal reaches `FeedVideos`*, which neither does alone.

## Also here

- **`subscribedList` → `'list'` rationale recorded** on the `analyticsTag` doc. Every other `sourceDetail` call site passes its discriminating value (`hashtag_feed_screen.dart:223` the hashtag, `videos_section.dart:106` the query), so the asymmetry reads as an oversight without a note. A list id is unbounded in cardinality and closer to user data than a public hashtag, and the question the field answers — which of the five home modes served the view — is answered without it.
- **`video_feed_source_analytics_test.dart` → `video_feed_state_test.dart`**, so it mirrors the `lib/` path holding the extension. The directory's other three files already mirror their source. Pure rename, 100% similarity.

## Verification

Mutation-checked, which is the only thing that makes this test worth having. Removing `sourceDetail: state.source.type.analyticsTag,` from `video_feed_page.dart`:

```
VideoFeedView passes home traffic attribution and the forYou source detail to FeedVideos
Expected: 'foryou'
  Actual: <null>
00:01 +0 -5: Some tests failed.
```

All five fail on the right assertion. The four extension tests stayed green throughout the mutation — confirming they never covered this.

Restored, then on current `main`:

- `flutter test test/screens/feed/video_feed_page_test.dart test/blocs/video_feed/` — 154 passed
- `dart format` — clean

## Type of Change

- [x] ✅ Test coverage


hm21 (approved):
LGTM. Verified the mutation independently — removing the `sourceDetail:` line fails all five new tests and leaves the four extension tests green, so the gap this closes was real. Parameterising over the enum with an exhaustive `_sourceFor` switch means a sixth feed mode can't slip through uncovered.

PR #6871, opened 2026-08-08
feat(analytics): instrument creator campaign funnel

rabble:
## Description

Adds campaign-critical attribution by propagating the authenticated exact 64-character hex pubkey to Firebase Analytics and Crashlytics, setting invite-code attribution after successful redemption, and clearing account-scoped identity on logout. Records the complete creation funnel and adds a deterministic post-publish create-again experiment that preserves the existing profile payoff behind a Remote Config kill switch. Operational setup and end-to-end validation are documented in docs/ANALYTICS_OBSERVABILITY.md.

**Related Issue:** None

## Out of Scope

Dashboards, self-serve analytics, and new notification types.

## Verification

- [x] Pre-push generated-file verification, analyzer, and 499 changed-file tests
- [x] flutter analyze --no-pub
- [x] Analytics package test suite: 73 tests
- [x] Focused instrumentation, identity, auth, recorder, metadata, publish, routing, and UI suites: 352 tests
- [x] Golden verification and repository guard scripts
- [ ] Register event-scoped mode and user-scoped invite_code custom dimensions in GA4
- [ ] Publish post_publish_create_again_enabled = true in Remote Config; false is the live kill switch
- [ ] Publish from a test account and confirm BigQuery user_id is exact 64-character hex, matches the ClickHouse pubkey, and the full mode-attributed funnel is present

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/docs/ANALYTICS_OBSERVABILITY.md
+++ b/docs/ANALYTICS_OBSERVABILITY.md
@@ -110,9 +110,11 @@ or `upload`.
 
 The tracker is app-scoped so recorder, editor, metadata, and publish routes
 contribute to one session. Creation metadata remains local analytics state and
-is not written into Nostr video events. `time_since_camera_open_ms` is `0` for
-an editor-only restored draft that did not open the camera in the active
-session.
+is not written into Nostr video events. `time_since_camera_open_ms` is
+**omitted entirely** — not sent as `0` — for an editor-only restored draft
+that did not open the camera in the active session, so those publishes cannot
+drag the timing distribution toward zero. Query it with `IS NOT NULL` rather
+than treating a missing value as instant.
 
 ## Post-Publish Experiment
 
--- a/docs/ANALYTICS_OBSERVABILITY.md
+++ b/docs/ANALYTICS_OBSERVABILITY.md
@@ -141,8 +141,10 @@ Events:
 
 After the invite service confirms redemption, the normalized code is set as
 the Firebase Analytics user property `invite_code`. Failed redemptions do not
-set it. Logout clears it so a second account on the device cannot inherit the
-first account's attribution.
+set it. Any change of authenticated identity clears it, so a second account on
+the device cannot inherit the first account's attribution. Logout is not the
+only such change: an in-place account switch never passes through an
+unauthenticated state.
 
 ## Required Firebase Admin Setup
 
--- a/docs/ANALYTICS_OBSERVABILITY.md
+++ b/docs/ANALYTICS_OBSERVABILITY.md
@@ -1,7 +1,8 @@
 # Analytics Observability
 
-Status: Current contract with semantic route screen views and comments sheet
-surface load instrumentation live.
+Status: Current contract with semantic route screen views, comments sheet
+surface load instrumentation, authenticated identity, and creator funnel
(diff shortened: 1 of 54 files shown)


rabble:
@dcadenas could you please review this before the campaign freeze? The two highest-risk areas are the exact 64-character hex identity fan-out and clearing, and whether each creation-funnel event fires only at its real lifecycle boundary. The rollout and manual validation checklist are captured in the PR description and docs/ANALYTICS_OBSERVABILITY.md.

hm21 (approved):
LGTM — solid instrumentation and unusually good docs. I pushed four fixes as separate commits rather than sending them back; revert any you disagree with.

- `87f0465f8` — the create-again snackbar navigated with `go`, so the recorder became a top-level route and closing it hit the `canPop() == false` branch and reset the user to the feed. The treatment arm lost the profile payoff the experiment measures. Now `push`, with a test that pins `push` and not `go`.
- `9a35c2a34` — `time_since_camera_open_ms` was sent as `0` when the session never opened the camera. Every restored-draft publish would have pulled the campaign's headline timing metric toward zero, indistinguishable from an instant publish. The parameter is now omitted; docs call out the `IS NOT NULL` query shape.
- `942c6cb5d` — `_publishVariants` only shrank via `completed()`/`failed()`, so a vanished background entry leaked for the process lifetime. Capped, oldest-first eviction.
- `21970803d` — Analytics/Crashlytics identity was set from inside `zendeskIdentitySync`, whose name, docstring and log line are still Zendesk-only and which is watched from exactly one place. Since the campaign joins BigQuery `user_id` against the ClickHouse pubkey, that shouldn't die silently if the support-desk integration is ever renamed or removed. It now has its own `analyticsIdentitySync` provider — `authStateStream` is broadcast, so the extra listener is free.

Not changed, flagging only: `variantForUser` hashes the bare pubkey with no experiment salt, so a future experiment reusing the scheme gets correlated arms. Fine for one experiment.

Validated: `flutter analyze lib test integration_test` clean, `dart format` clean, 1180 affected tests green locally, full CI green on `21970803d`. Yours to merge.

dcadenas (approved):
Took this over and pushed three fixes.

The one that mattered: `invite_code` is a user-scoped GA4 property and was only cleared on logout. An in-place account switch builds a fresh provider container and never emits an unauthenticated state, so the incoming account kept the previous account's invite attribution — the single thing this instrumentation exists to measure. The applied identity is now tracked at process scope and the property is cleared on any identity change, while still surviving the redeeming login that sets it before the account authenticates. The doc claim about logout is corrected to match.

Two smaller ones. The analytics identity and the experiment bucket are now lowercased, because an external signer can hand back uppercase hex and both the BigQuery/ClickHouse join and the assignment hash are exact string compares. And the recorder's `BlocProvider` key now includes the analytics tracker it watches.

Validated with analyze plus the analytics, publish, recorder, and auth suites, and CI is green on the pushed head. Separately: `nip05_settings_navigation_test` fails under randomized test ordering, but it fails identically on `main`, so it is not from this branch and I left it alone. Merge is yours.

PR #6872, opened 2026-08-08
feat(feed): hide small view counts on video cards behind a flag

### LINKED ISSUE #6873: feat(feed): hide small view counts on video cards behind a flag
Auto-created from PR #6872.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6872
**Branch:** `feat/video-card-post-date`

rabble:
Closes #6873

## What

Video cards stop showing small public view counts. The post date takes that slot instead.

A fresh post renders `0 loops` today — right as funnelcake starts guaranteeing new creators 100–200 impressions through the exploration budget. A small number reads as a warning on exactly the content the feed just paid to surface. 80% of new creators never clear 100 views in their first week, and crossing ~100 roughly doubles the chance they keep posting (31% → 61%).

**The rule is about size, not provenance.** A public count shows only at or above `publicLoopCountFloor` (1000), so a famous Vine keeps its millions and a genuinely popular new video earns the same social proof — while small numbers from either source are replaced by the date.

| Case | Renders |
|---|---|
| Flag off | `12 loops` — today's behavior, unchanged |
| Your own video | `3h ago · 12 loops` — always, however small |
| Count ≥ floor | `Apr 22, 2014 · 2.1M loops` |
| Count < floor | `3h ago` |

Creators always see their own number. Correcting a creator's underestimate of their audience is what keeps them posting (Bernstein et al., CHI 2013), and the exploration floor creates exactly that underestimate.

Classic Vines report `originalLoops`, never `totalLoops` — folding live diVine views into an archival figure misreports how popular the Vine actually was.

Dates are age-adaptive: `3h ago` under a week, `Apr 22, 2014` beyond. Divine mixes ~2.2M archived vines with content posted minutes ago, so the year turns a bare clip into an artifact. That also speaks to a recurring review complaint about confusing archive content ("weird videos of old people", App Store, 2026-08-06, v0.1.17).

## Also in this PR

- **More-info sheet**: Loops moves from first to last stat column. The count stays reachable, just stops leading the row.
- **Creator analytics**: the "Most viewed" highlight now reads `12 people watched this`. New ARB key, translation deferred via `_knownUntranslatedDebt`.

## Ships unmeasured, by design

The originating spec asked for a randomized variant flag and a `video_card_shown` impression event. Neither exists to build on: the wired flag system (`lib/features/feature_flags/`) is boolean-only with no user bucketing, the variant-capable `lib/services/feature_flag_service.dart` has **zero imports anywhere** in `lib` or `test`, and there is no impression event at all.

That is infrastructure work, not a flag toggle, so it was cut and deferred. **The hypothesis stays untested by this change** — a deliberate trade, since this was the third client change queued for the same four days and the designated one to slip.

`FeatureFlag.videoCardPostDate` ships **default off** as a kill switch and internal preview toggle.

## Things a reviewer should push back on

- **`publicLoopCountFloor = 1000` is a product call, not a technical one.** It's a named constant so moving it is a one-line diff. I picked it; nobody signed off on it.
- **Analytics copy is now inconsistent on one screen** — the highlight says "people watched this", the dense per-video rows still say "views" because `12 people watched this • 3 interactions` doesn't fit that line. Deliberate, but reversible in either direction.
- **Share previews lose a bogus `0 loops`.** With a null video the old code rendered `video?.totalLoops ?? 0`. The line is now omitted. Not requested, small, an improvement — but a behavior change.

## Out of scope

DM video previews (`message_bubble.dart:1161`) keep the old line. `VideoAuthorInfoSection` and `VideoDescriptionOverlay` are dead code (instantiated only by their own tests) and were left alone.

## Testing

| File | Tests |
|---|---|
| `video_card_meta_test.dart` (new) | 17 |
| `video_feed_item_meta_line_test.dart` (new) | 6 |
| `metadata_stats_row_test.dart` (new — file had no test before) | 5 |
| `localized_time_formatter_test.dart` (extended) | 5 |

`flutter analyze lib test integration_test` clean. 478 tests green across `test/widgets/video_feed_item/` and `test/l10n/`. All four design-system ceiling ratchets, the test-structure floor, and the untested-services floor pass. No goldens cover these widgets.

**Not done: I have not run the app and looked at the change on a device.** Screenshots should be attached before merge.

### DIFF
--- a/docs/superpowers/specs/2026-08-08-video-card-view-count-display-design.md
+++ b/docs/superpowers/specs/2026-08-08-video-card-view-count-display-design.md
@@ -1,0 +1,261 @@
+# Video card view-count display
+
+**Date:** 2026-08-08
+**Status:** Approved, ready for implementation planning
+**Target:** Release before the ~2026-08-13 freeze
+
+## Problem
+
+A video card's secondary line renders an unconditional loop count. A freshly
+posted video reads `0 loops`, and funnelcake is about to guarantee new
+creators' first videos 100–200 impressions through an exploration budget. Those
+videos would surface stamped with a discouraging number: we pay for the
+distribution and simultaneously tell every viewer not to bother.
+
+Supporting measurements:
+
+- 80% of new creators never reach 100 views on any first-week video.
+- Crossing ~100 views roughly doubles the chance they keep posting (31% → 61%).
+- New-creator content out-performs established-creator content per impression
+  (loops 1.143 vs 1.092, completion 1.147 vs 1.097), so it deserves a fair
+  viewing that a low count denies it.
+
+Creator-facing counts are the opposite case. Bernstein et al. (CHI 2013) found
+people badly underestimate their audience, and that perceived audience drives
+production. The exploration floor will create exactly that underestimate, so
+the creator's own view of their numbers stays intact.
+
+The date is more useful on Divine than on other platforms: ~2.2M archived vines
+from 2013–16 sit alongside content posted minutes ago. "Apr 22, 2014" vs
+"3h ago" reframes what the viewer is watching and turns a bare clip into an
+artifact. It also partly addresses a recurring review complaint about confusing
+archive content ("weird videos of old people", App Store, 2026-08-06, v0.1.17).
+
+## What the code actually does today
+
+The number on the card is a **loop** count, not a view count.
+
+The live site is `mobile/lib/widgets/video_feed_item/video_feed_item.dart:394`,
+inside `VideoOverlayActions.build`. It renders `videoFeedLoopCountLine` directly
+under the author name, unconditionally — so `0 loops` on a fresh post is real.
+
+`VideoEvent.totalLoops` (`mobile/packages/models/lib/src/video_event.dart:1084`)
+is:
+
+```dart
+int get totalLoops =>
+    (originalLoops ?? 0) + (int.tryParse(rawTags['views'] ?? '') ?? 0);
+```
+
+Archive Vine loops **plus** live Divine views, summed into one number. The two
+components are available separately, which is what makes the rule below
+non-arbitrary.
+
+Other surfaces carrying the same line:
+
+- `mobile/lib/widgets/video_feed_item/metadata/metadata_stats_row.dart` — the
+  more-info sheet. Already shows Loops/Likes/Comments/Reposts plus a
+  Vine-vs-Divine breakdown for classic vines.
+- `mobile/lib/screens/inbox/conversation/widgets/message_bubble.dart:1161` — DM
+  video previews. **Out of scope.**
+- `video_author_info_section.dart` and `actions/video_description_overlay.dart`
+  are dead code — instantiated only by their own tests, referenced in `lib`
+  only from two comments. **Left untouched.**
+
+## Decisions
+
+### Scope: display change only
+
+The spec that motivated this work asked for a randomized variant flag and a
+`video_card_shown` impression event. Neither is cheap here:
+
+- `mobile/lib/features/feature_flags/` is the wired flag system: enum-based,
+  boolean only, local SharedPreferences plus build config. No user bucketing,
+  no variants.
+- `mobile/lib/services/feature_flag_service.dart` is HTTP-backed and *does*
+  support variants and rollout hashing — but has zero imports anywhere in `lib`
+  or `test`. Dead code.
+- No impression event exists. `AnalyticsService` has a generic `track()`, but
+  all view tracking is playback-side.
+
+So randomized measurement is infrastructure work, not a flag toggle. It was cut
+from this change and deferred to separate post-freeze work. **The hypothesis in
+the originating spec therefore stays untested by this change** — that is a
+deliberate trade, made because this was already the third client change queued
+for the same four days and the designated one to slip.
+
+### Display rule
+
+| Case | Date | Count | Renders |
+|---|---|---|---|
+| Flag off | — | `totalLoops` | `12 loops` (today's behavior, unchanged) |
+| Own video | post date | `totalLoops` | `3h ago · 12 loops` |
+| Classic vine | post date | `originalLoops` | `Apr 22, 2014 · 2.1M loops` |
+| Someone else's Divine post | post date | none | `3h ago` |
+| Preview mode (`video == null`) | none | none | line omitted |
+
+Cases are evaluated in the order listed, so `isOwnVideo` wins over
+`isOriginalVine`. A creator viewing their own classic vine — possible where a
+vine account has been claimed — sees `totalLoops`, the combined archive plus
+live number, because at that point it is their own performance data rather than
+a public signal.
+
+Two further consequences worth stating explicitly:
+
+**Classic vines use `originalLoops`, not `totalLoops`.** The archival number is
+the historical fact being surfaced. `totalLoops` would silently fold live
+Divine views into "how famous this vine was".
+
+**Preview mode currently renders a bogus `0 loops`** — `video?.totalLoops ?? 0`
+with a null video. The rule drops the line instead. This is a behavior change
+on share previews that was not requested; it is small and an improvement, but
+it is a change.
+
+The creator's own zero stays visible. `0 loops` on your own fresh post is
+accurate, it is the status quo, and it is the number the exploration budget
+exists to move.
+
+### Date format: age-adaptive
+
+Relative under 7 days (`3h ago`), absolute with the year beyond it
+(`Apr 22, 2014`). This produces the fresh-vs-archive contrast the change is
+for. Always-absolute reads cold on a ten-minute-old post; always-relative turns
+a classic vine into `12y ago`, which reads as neglect rather than history and
+works against the App Store complaint above.
+
+### Card copy: reuse the shipped key
+
+The card keeps `videoFeedLoopCountLine` ("12 loops"). No new ARB key, so no
+16-locale mirroring for the card. The warmer "12 people watched this" framing
+goes to the creator analytics screen instead, where the line is not width-
+constrained.
+
+### More-info sheet: reorder, don't restyle
+
+Move the Loops `_StatColumn` from first to last position in
+`MetadataStatsRow`. The count stays available per the originating spec's
+"de-emphasised, not removed", de-emphasised by reading order. Restyling it
+smaller in place was considered and rejected — a visually downgraded first
+column reads as broken rather than intentional.
+
+## Design
+
+### 1. Pure resolver
+
+New file `mobile/lib/widgets/video_feed_item/video_card_meta.dart`:
+
+```dart
+@immutable
+class VideoCardMeta {
+  const VideoCardMeta({this.timestamp, this.loopCount});
+  final int? timestamp;   // unix seconds; null when the date is unknown
+  final int? loopCount;   // null when the count stays private
+  bool get isEmpty => timestamp == null && loopCount == null;
+}
+
+VideoCardMeta resolveVideoCardMeta({
+  required VideoEvent? video,
+  required bool isOwnVideo,
+  required bool showPostDate,
+});
+```
+
+No Flutter dependency, so the whole truth table above is unit-testable without
+pumping a widget.
+
+Date source matches the more-info sheet:
+`int.tryParse(video.publishedAt ?? '') ?? video.createdAt`, suppressed when
+`video.hasUnknownOriginalDate` — classic vines whose original date is not
+recoverable. A classic vine with neither a usable date nor an archive count
+yields `isEmpty`, and the line is omitted.
+
+### 2. Date formatting
+
+Extend `mobile/lib/l10n/localized_time_formatter.dart`. It already wraps the
+l10n-free `TimeFormatter` package with ARB strings and uses `package:clock`, so
+tests pin time with `withClock`:
+
+```dart
+/// Recent posts read as elapsed time; a week or older falls back to an
+/// absolute date with the year, so archives read as artifacts.
+static String formatPostAge(
+  AppLocalizations l10n,
+  int unixSeconds, {
+  String? locale,
+});
+```
+
+Under 7 days delegates to the existing localized `formatRelativeVerbose`.
+Beyond, `TimeFormatter.formatAbsoluteDate` — `DateFormat.yMMMd(locale)`, whose
+doc comment already notes that Vine archives need their year visible.
+
+### 3. Feature flag
+
+`FeatureFlag.videoCardPostDate`, `audience: user`, **default off**. Adding the
+enum case forces both exhaustive switches in
+`mobile/lib/features/feature_flags/services/build_configuration.dart` to be
+updated; the compiler catches an omission.
+
+This is a kill switch and an internal preview toggle. It does not randomize and
+does not fetch remotely.
+
+### 4. Widget changes
+
+**`video_feed_item.dart:394`** — replace the unconditional `Text` with the
+resolved line. `VideoOverlayActions.build` is already a `ConsumerWidget` with
+`ref`, so `isOwnVideo` comes from
+`ref.watch(authServiceProvider).currentPublicKeyHex`, the same read
+`VideoOverlayActionColumn` already performs at line 673.
+
+Date and count join with `' · '` at the widget layer. That is the established
+idiom in this codebase (`sound_tile.dart`, `audio_attribution_row.dart`,
+`metadata_sounds_section.dart`, and five other call sites), and it avoids a new
+ARB key. The separator is bidi-neutral, so RTL locales order the whole string
+correctly.
+
+**`metadata_stats_row.dart`** — move the Loops `_StatColumn` to last.
+
+### 5. Analytics wording
+
+`creator_analytics_screen.dart:718` and `:930` use `analyticsViewsCount` =
+`"{count} views"`. Reaching "12 people watched this" requires a **new** key
+(`analyticsWatchedCount`), not an edit of the existing value: editing in place
+would leave 16 locales holding translations of "views" while English said
+something else, which `.claude/rules/localization.md` explicitly warns against.
+
+New key means mirroring across all `app_*.arb` locales or an entry in
+`_knownUntranslatedDebt` in `test/l10n/arb_consistency_test.dart`.
+
+**This is the most cuttable item in the change.** It is the only piece carrying
+l10n cost, and it is furthest from the stated goal — the card and the more-info
+sheet deliver the whole "hide discouraging counts, show dates" outcome without
+it. If the pre-freeze window tightens, cut §5 first.
+
+## Testing
+
+| File | Covers |
+|---|---|
+| `test/widgets/video_feed_item/video_card_meta_test.dart` (new) | Every row of the truth table; `originalLoops` vs `totalLoops` for classic vines; `hasUnknownOriginalDate`; the empty case |
+| `test/l10n/localized_time_formatter_test.dart` (extend) | `formatPostAge` either side of the 7-day boundary, under `withClock` |
+| `test/widgets/video_feed_item/video_feed_item_meta_line_test.dart` (new) | Flag off renders today's string; flag on + own video renders date and count; flag on + another's Divine post renders no count |
+| `test/widgets/video_feed_item/metadata/metadata_stats_row_test.dart` (new) | Column order; classic-vine breakdown still rendering. This file has no test today. |
+
+Widget tests need `AppLocalizations.localizationsDelegates` and
+`supportedLocales` on their `MaterialApp`, and must resolve expected strings
+from `AppLocalizations` rather than hardcoding English.
+
+No goldens cover these widgets, so there is no golden churn.
+
+## Out of scope
+
+- DM video previews (`message_bubble.dart:1161`) keep the old line.
+- The two dead widgets stay as-is.
+- `video_card_shown` impression event and variant bucketing — separate
+  post-freeze work. Without them this change ships unmeasured.
+- Any further de-emphasis of the more-info sheet beyond column order.
+
+## Open risk
+
+This is the third client change queued for the same pre-freeze window, and the
+one already designated to slip if something must. Nothing here depends on the
+other two, so it can be dropped late without unpicking them.
--- a/docs/superpowers/specs/2026-08-08-video-card-view-count-display-design.md
+++ b/docs/superpowers/specs/2026-08-08-video-card-view-count-display-design.md
@@ -1,7 +1,7 @@
 # Video card view-count display
 
 **Date:** 2026-08-08
-**Status:** Approved, ready for implementation planning
+**Status:** Implemented
 **Target:** Release before the ~2026-08-13 freeze
 
 ## Problem
@@ -48,8 +48,8 @@ int get totalLoops =>
 ```
 
 Archive Vine loops **plus** live Divine views, summed into one number. The two
-components are available separately, which is what makes the rule below
-non-arbitrary.
+components are available separately, which is what lets a classic Vine report
+its archival figure alone.
 
 Other surfaces carrying the same line:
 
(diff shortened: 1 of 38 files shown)


rabble:
## Self-review (not a substitute for a human one)

I wrote this branch, so GitHub won't let me approve it and I can't independently audit my own judgment calls. Posting findings as a comment instead. Five issues found; all five addressed in `6570e0f70`.

**1. HIGH — `formatPostAge` formatted the UTC instant, not the local day.** Every sibling method in `localized_time_formatter.dart` calls `.toLocal()` first; this one didn't, and `DateFormat.format` reads the DateTime's field accessors. A Vine with a `published_at` near midnight UTC rendered a day late for every viewer west of it. Fixed, with a test pinning the local-vs-UTC contract.

Worth flagging how this hid: **the runner sits at UTC+12**, so my literal `'Apr 22, 2014'` assertions were simultaneously wrong and passing. I "fixed" them once by moving to a midday-UTC timestamp, which is *also* wrong at exactly UTC+12. Date assertions are now computed from the local conversion rather than guessed.

**2. MEDIUM — the analytics copy change regressed 16 locales. Reverted entirely.** Swapping `analyticsViewsCount` → `analyticsWatchedCount` replaced a string translated into 16 languages with an English-only one on the debt list, so a German creator would have gone from "1,2 Tsd Aufrufe" to "1.2K people watched this". It also sat *outside* `FeatureFlag.videoCardPostDate`, so flipping the kill switch off would have left it stranded. This was already flagged in the PR body as the cut candidate; cutting it.

**3. MEDIUM — `isOwnVideo` never recomputed on sign-in, and my own perf commit made a false claim.** `AuthService` dropped `ChangeNotifier` (`auth_service.dart:187`) and `authServiceProvider` is a plain keepAlive provider returning a stable singleton — so `ref.watch(authServiceProvider)` never rebuilds on a pubkey change, with or without `.select`. My earlier commit `a462a3faa` "fixed" this by adding `.select` and left a comment asserting reactivity that does not exist. Now watches `currentAuthStateProvider` for its invalidation, with a regression test covering signed-out → signed-in.

**Known remaining gap:** an account switch that never leaves the `authenticated` state still won't rebuild, because `currentAuthState` returns the same enum and Riverpod suppresses the notification. No pubkey-reactive provider exists in this codebase to fix that without touching shared auth infrastructure, which I'm not doing days before a freeze. Worst case is a transient stale count until the next rebuild, and feed cards rebuild constantly.

**Related, pre-existing, not fixed here:** `feed_videos.dart:804` uses the same non-reactive `.select(...currentPublicKeyHex)` for its own `isOwnVideo`. Same latent staleness, in the live feed. Out of scope for this PR — worth its own issue.

**4. LOW — pluralization.** Moot now that #2 is reverted.

**5. LOW — hardcoded English in test assertions.** `find.textContaining('7 loops')` etc. bypassed l10n, which `.claude/rules/localization.md` bans. Routed through the `loopLine(tester, n)` helper.

---

### Still open, and not something review can settle

`publicLoopCountFloor = 1000` remains **a number I invented.** The only measured figure we have is that ~100 views is where *creator retention* inflects — that's about how creators feel, not how viewers read a number. I extrapolated an order of magnitude past our one data point. I tried to check the actual loop-count distribution to ground it and the ClickHouse MCP server is erroring out (`No module named 'rich.traceback'`), so it's still a guess.

It's a named constant, so moving it is a one-line diff.

**Also still true: nobody has run this.** No screenshots, no device. The floor is exactly the kind of thing you can only judge by scrolling a real feed.

mobile/test/l10n/localized_time_formatter_test.dart
```diff
@@ -596,5 +596,103 @@ void main() {
...
+        // year and the absolute form.
+        final rendered = _withFixedClock(
+          () => LocalizedTimeFormatter.formatPostAge(en, 1398168000),
+        );
+
+        expect(rendered, contains('2014'));
+        expect(rendered, isNot(contains('ago')));
+      });
+
+      testWidgets('renders the local calendar day, not the UTC one', (
```
realmeylisdev: should-fix: this guard can't fail on CI, so the `toLocal()` fix ships unpinned.

I removed `.toLocal()` from `formatPostAge` (reproducing exactly what `8de0b2f3a` fixed) and ran all four test files from this PR under `TZ=UTC`: 68/68 passed. Under `TZ=America/Los_Angeles` the same mutation fails here as intended:

```
Expected: 'Apr 21, 2014'
  Actual: 'Apr 22, 2014'
```

Mobile CI runs `ubuntu-24.04` with no TZ override, so it is UTC. `expected` is built with the same `fromMillisecondsSinceEpoch(..., isUtc: true).toLocal()` conversion the implementation uses, and under UTC that conversion is the identity, so both sides collapse to the same string whichever way the implementation goes.

Worth noting the runner is UTC, not the UTC+12 in your self-review note. Both are inert for this instant, but only one of them is CI.

There's no instant that separates local from UTC when the process TZ *is* UTC, so this can't be repaired inside the test. It needs the suite running under a non-UTC TZ (a `TZ:` env on the shard, or a dedicated job), which is a workflow change and your call. Either way I'd rather that be a decision than a test whose name promises a guarantee it doesn't give.

mobile/lib/features/feature_flags/models/feature_flag.dart
```diff
@@ -96,6 +96,13 @@ enum FeatureFlag {
     'Client Seen-Video Filtering',
     'Demote recently-seen videos in home feeds and drop them from '
         'classics/discovery. Kill-switch for the Aug 2026 campaign load.',
+  ),
+  videoCardPostDate(
```
realmeylisdev, line 100: should-fix: this wants `audience: FeatureFlagAudience.internal`.

Without it the flag defaults to `FeatureFlagAudience.user`, and then:

- `feature_flag_screen.dart:22` filters on `!flag.isInternal || showInternalFlags`, so a user-audience flag lists for everyone
- the "Experimental Features" tile at `settings_screen.dart:466` sits *outside* the `if (isDeveloperMode)` block that starts at :475 directly below it, so any user reaches that screen
- `_canOverrideFlag` (`feature_flag_service.dart:180`) returns true unconditionally for user flags, so the override persists

So "Post Date On Video Cards" ships as a switch any user can flip. That doesn't match your own comment at `build_configuration.dart:71` ("previewed internally and killed remotely before it reaches the pre-campaign release") or the description ("internal preview toggle").

The two closest siblings in this same enum with the identical shape, `publishDmRelayList` and `feedTuning`, both set `internal` for exactly this reason.

mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -27,6 +27,15 @@
         "version" : "0.0.9-beta.7"
       }
     },
+    {
+      "identity" : "cocoaasyncsocket",
```
realmeylisdev, line 31: nit: unrelated to this PR, worth dropping before merge.

Nothing in the diff changes a dependency (no pubspec, no Podfile, no plugin), so this `cocoaasyncsocket` pin is local Xcode SPM drift rather than anything the change needs. I reproduced the mechanism: building this branch for the simulator rewrote the same file in my worktree.

```
git checkout -- mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```

mobile/lib/widgets/video_feed_item/video_card_meta.dart
```diff
@@ -0,0 +1,75 @@
...
+/// Smallest public count that reads as a recommendation rather than a warning.
+///
+/// A number below this tells a viewer not to bother — which is worse than
+/// silence on content the feed just paid to surface. Above it, the number is
+/// doing useful work: a classic Vine with millions of loops is famous, and
+/// saying so is the point.
+///
+/// This threshold is a product call, not a technical one. It is the single
+/// value to change if the bar turns out to sit in the wrong place.
+const int publicLoopCountFloor = 1000;
```
realmeylisdev, line 21: Not a defect, just the number you said you couldn't get. ClickHouse was down for you; the public API answers it.

1000 unique classic Vines from `/api/videos?platform=vine`, measured on the Nostr `loops` **tag**, which is what mobile parses into `originalLoops`:

| p50 | p75 | p90 | max |
|---|---|---|---|
| 298 | 2,599 | 23,885 | 1,149,361 |

At a floor of 1000, **64.3%** of classic Vines lose their count. At 5000 it's 80.7%; at 100 it's 37.1%.

One trap if you re-measure: funnelcake's REST `loops` *field* is a different and much smaller number than the `loops` *tag* in the same response. Median 64 vs 298, and for `d1ab993e23fc386302565f4a460cef59074ecbc7c4eb506b27d71bf7b832e4c8` the field says 23,414 while the tag says 2,614,474. Mobile reads the tag, so the field will understate how much the floor hides.

Separately, and not in the table in the description: 86 of the 1000 (8.6%) carry no `published_at` tag and `loops` 0. I checked those against raw relay events, and `created_at` is the 2026 import time, so `hasUnknownOriginalDate` is true and the count is under the floor. Both halves come back null, `isEmpty` is true, and the meta line disappears entirely rather than falling back to a date. Today those render "0 loops". Probably the outcome you want, but the description implies a date always takes the slot.

mobile/lib/widgets/video_feed_item/video_card_meta.dart
```diff
@@ -0,0 +1,75 @@
...
+  return publicCount >= publicLoopCountFloor ? publicCount : null;
+}
+
+/// The count a stranger would be shown, before the floor is applied.
+///
+/// Classic Vines report their archival figure alone. Folding live diVine
+/// views into it would misreport how popular the Vine actually was, and our
+/// current view volume is too small to change the number meaningfully anyway.
+int _publicCount(VideoEvent video) =>
+    video.isOriginalVine ? video.originalLoops ?? 0 : video.totalLoops;
```
realmeylisdev, line 80: Cross-checked this against divine-web and it lines up, which is worth knowing since you cited the archival-figure reasoning from first principles.

`getVisiblePlaybackCount` (`divine-web/src/lib/playbackCount.ts`) already does the same split:

```ts
return isVineMigrated ? loops : Math.max(loops, views);
```

Classic Vine gets the archival number alone on web too, so this isn't a mobile-only interpretation.

The part web does *not* have is a floor: it renders every count, including 0. So with the flag on, the same video reads "3h ago" on mobile and "7 loops" on web. Fine while this is a flagged mobile experiment, just worth being deliberate that web is the control rather than an oversight.

mobile/lib/widgets/video_feed_item/video_feed_item.dart
```diff
@@ -144,6 +146,19 @@ class VideoOverlayActions extends ConsumerWidget {
...
     final previewData = this.previewData;
     final authorPubkey = previewData?.pubkey ?? video!.pubkey;
+    final showPostDate = ref.watch(
+      isFeatureEnabledProvider(FeatureFlag.videoCardPostDate),
+    );
+    // Watched purely for its invalidation. AuthService is not a ChangeNotifier
+    // and authServiceProvider hands back a stable singleton, so watching it
+    // (with or without select) never rebuilds on sign-in or account switch —
+    // a card mounted while signed out would keep hiding its owner's counts.
+    ref.watch(currentAuthStateProvider);
```
realmeylisdev, line 156: Both reactivity gaps you flagged in the self-review look clear to me, so I don't think either needs a follow-up.

**Account switch.** `swap_account.dart` builds an entirely new `ProviderContainer`, and `container_swap_host.dart:126` renders the child inside `KeyedSubtree(key: ValueKey(_generation))`. `_generation` bumps on every swap, so the whole tree remounts under the new container with a fresh `AuthService`. Nothing stale survives, whatever `currentAuthState` does.

**`feed_videos.dart:805`.** The same build method already does `ref.watch(likesRepositoryProvider)` five lines above at :800, and `likesRepository` (`video_providers.dart:600`) watches `currentAuthStateProvider` itself and returns a new instance. `LikesRepository` doesn't override `==`, so the fresh identity notifies, the overlay rebuilds, and `currentPublicKeyHex` is re-read on exactly the same signal. I couldn't construct a case where it goes stale.

This line here is doing real work, though: dropping it turns "reveals the creator their count after they sign in" red, so the test is a genuine pin.

realmeylisdev (changes requested):
went through this one carefully. two things i'd fix before it ships, everything else looks right to me.

the local-vs-UTC test can't actually fail on CI (ubuntu runners are UTC, confirmed by deleting the toLocal again and watching all 68 tests still pass), and the flag is missing `audience: internal`, so the internal preview toggle is visible to every user under Experimental Features.

also left you the loop-count distribution you couldn't get out of clickhouse, and neither of the two reactivity gaps in your self-review reproduces. rest is inline.

mobile/lib/widgets/video_feed_item/video_card_meta.dart
realmeylisdev: Rendered the real `VideoOverlayActions` for each case (throwaway harness, nothing committed) since you flagged that nobody had looked at it yet. Machine TZ US Pacific, so the Vine date is a real local conversion rather than a UTC coincidence:

| case | card says |
|---|---|
| flag OFF, fresh post, 0 views | `0 loops` |
| flag ON, stranger, 7 views | `Now` |
| flag ON, stranger, 50k views | `Now · 50K loops` |
| flag ON, owner, 7 views | `Now · 7 loops` |
| flag ON, classic Vine 2.1M loops + 340 live | `Apr 22, 2014 · 2.1M loops` |
| flag ON, classic Vine 47 loops | `Apr 22, 2014` |
| flag ON, archive import, no `published_at`, 0 loops | line omitted entirely |

Flag-off is byte-identical to today, and the archival figure correctly excludes the 340 live views. The last row is the 8.6% case from the other thread: the author name loses its subtitle rather than falling back to anything.

Not a substitute for screenshots on a device. I couldn't get one: iOS builds off this worktree die in SPM resolution (the FlutterFire `4.2.1` vs `4.4.0` conflict), and the documented fix wants the shared DerivedData cleared, which would have knocked over two other builds running on this machine at the time. It does boot fine in Chrome off this branch, but signed-out routing bounces everything to `/welcome`, so no feed there either.

rabble:
Thanks — this is the most useful review I've had on this branch. You did the two things I said I couldn't, and both changed the PR.

Pushed as `2ed697560`, checks green.

**Fixed — `audience: FeatureFlagAudience.internal`.** You're right and the evidence chain is exact: the Experimental Features tile at `settings_screen.dart:466` sits outside the `isDeveloperMode` block starting at :475, and `_canOverrideFlag` returns true unconditionally for user flags. So it shipped as a switch any user could flip, while my own description and the `build_configuration.dart` comment both called it internal. I asserted "internal preview toggle" in the spec, the PR body and the commit message without ever checking the code said so. Matched to `publishDmRelayList` / `feedTuning`.

**Fixed — stray `Package.resolved`.** Mine, via `git add -A mobile/`. My own pre-commit checklist says to discard exactly this lockfile churn and I swept it up anyway. Diff is back to zero lines on that file.

**Fixed, but not the way you might prefer — the local-vs-UTC guard.** Your reproduction is conclusive: you deleted `toLocal()` and got 68/68 green under `TZ=UTC`. Under UTC the conversion is the identity, so both sides collapse and there's no instant that separates them — it can't be repaired inside the test.

I renamed it to `...(off-UTC machines only)` and documented why it's inert on CI, rather than leave a name promising a guarantee it doesn't give. I did **not** add a `TZ:` env to the shard job. That's a shared-CI change with blast radius across every suite, days before a freeze, on a PR about view counts — I don't think I should make that call unilaterally inside this diff. Flagging it as worth doing separately: I ran the full affected suites at UTC+12 and they're clean, so there's at least some evidence the codebase tolerates a non-UTC runner.

(Also noted: you're right that the runner is UTC, not the UTC+12 I wrote in my self-review note. That was my machine, not CI, and conflating them is what let me believe the test was load-bearing.)

**Not changing — the two reactivity gaps.** Your analysis beats mine and I'm withdrawing both caveats. `container_swap_host.dart:126` remounting under `KeyedSubtree(key: ValueKey(_generation))` makes the account-switch case moot regardless of what `currentAuthState` does, and `likesRepository` watching `currentAuthStateProvider` five lines above covers `feed_videos.dart:805` on the same signal. I'd asserted both as known gaps in my self-review; they aren't.

**Escalating to Rabble — the floor, and the 8.6% case.**

The distribution is the number I wanted and it argues against my constant. p50 **298**, and at `publicLoopCountFloor = 1000` **64.3%** of classic Vines lose their count. The stated goal was that big archival numbers are good and should show, so hiding two-thirds of the archive is probably not what was intended. That's a product call, not mine — it's a one-line change either way, and 100 (37.1%) is the other datapoint we have, since ~100 views is where creator retention inflects.

Thanks also for the field-vs-tag trap. Mobile reads the tag via `originalLoops`, so anyone re-measuring off the REST `loops` field would understate the hiding by roughly 5x.

On the 8.6% with no `published_at` and `loops` 0: line omitted entirely, both halves null. You're right the description implies a date always takes the slot — it doesn't, and can't, because `hasUnknownOriginalDate` suppresses the date and the floor suppresses the count. I think dropping the subtitle beats rendering "0 loops", but I'll fix the description once the floor is settled so the two land together.

**divine-web divergence** noted and deliberate: web renders every count including 0, so the same video reads "3h ago" on mobile and "7 loops" on web while this is flagged. Worth revisiting if it ever defaults on.

**Your rendering table is the closest thing to a device check this has had** — and it's the first time anyone has seen the actual output. Appreciated, including the honest note that the iOS SPM failure and the `/welcome` bounce blocked a real screenshot.

rabble:
Follow-up from Rabble: the archival counts should show on classic Vines regardless of size. Pushed as `a5f5fa01c`, checks green.

**The rule is now provenance, not size.** I had it backwards. A classic Vine's archival figure is a fact about how the clip did on Vine — it is never a verdict on diVine, however small it is. diVine's own counts are the ones that read small, so the floor now applies only to those.

| case | before | now |
|---|---|---|
| classic Vine, 47 loops | `Apr 22, 2014` | `Apr 22, 2014 · 47 loops` |
| classic Vine, p50 (298) | `Apr 22, 2014` | `Apr 22, 2014 · 298 loops` |
| classic Vine, 2.1M | unchanged | `Apr 22, 2014 · 2.1M loops` |
| diVine post, 7 views | unchanged | `3h ago` |
| diVine post, 50K views | unchanged | `3h ago · 50K loops` |

Your distribution is what settled it: at a floor of 1000, 64.3% of the archive lost its count. Hiding two-thirds of the archive was the opposite of the goal, and I would not have known without the numbers you pulled.

`publicLoopCountFloor` still exists but now governs only live diVine counts, where p50 tells us nothing yet.

---

**Separate, not in this PR — why combined counts keep breaking.**

Rabble mentioned that combining archival and live counts has failed repeatedly. There's a structural reason, and your field-vs-tag note is adjacent to it.

`totalLoops` adds a typed field to a string in an untyped map:

```dart
int get totalLoops =>
    (originalLoops ?? 0) + (int.tryParse(rawTags['views'] ?? '') ?? 0);
```

`originalLoops` merges like every other engagement field. `views` has to be hand-special-cased inside a `rawTags` spread at every merge, cache and copy site. In `profile_video_metadata_cache.dart` five fields are plain `??` one-liners and `views` needs its own bespoke branch. Four sites currently hand-roll that correctly, and two production incidents forced them there: `b0668f18c` (enrichment wiping `rawTags['views']`) and #3384 (Nostr zeroing Funnelcake aggregates).

So the combined number fails whenever a new code path forgets one string key, and nothing in the type system can catch it.

The fix is a typed `divineViews` field on `VideoEvent`, parsed once at construction, so both halves merge under the same rules and `mergeNullableEngagementMax` covers them symmetrically. That is a models-package change with a wide blast radius, so I'm not putting it in a flagged feed experiment days before a freeze — flagging it as worth its own issue.

mobile/lib/widgets/video_feed_item/video_card_meta.dart
```diff
@@ -0,0 +1,80 @@
...
```
realmeylisdev, line 80: this reads the archival figure on the relay path but the live one on the feed path, and the floor then hides exactly the vines it was meant to keep.

`VideoStats.fromJson` resolves loops as `statsData['embedded_loops'] ?? json['embedded_loops'] ?? json['original_loops'] ?? statsData['loops'] ?? ...`. `/api/videos` returns a flat record with no `embedded_loops`, so `statsData == json` and it falls through to funnelcake's live `loops` field. Only `/api/videos/<id>` returns the nested `{event, stats}` shape that carries `embedded_loops`.

Worked example, real record `efe944e59c19a598404eb356c25c962c20c46a8daa3fbbe95b7b5a8c82e90572`, run straight through `VideoStats.fromJson(...).toVideoEvent()` and then `resolveVideoCardMeta`:

| loaded from | originalLoops | flag off | flag on |
|---|---|---|---|
| `/api/videos` (every feed) | 723 | `2.85K loops` | `Jan 17, 2017` |
| `/api/videos/<id>` | 192,604 (matches the event's `loops` tag) | `192.6K loops` | `Jan 17, 2017 · 192.6K loops` |

I rendered the real `VideoOverlayActions` for the first row and the card does come back with the date and no count.

Scale, over 1,915 classic vines that carry both a `loops` tag and a `loops` field: the two disagree for 77% of them (tag p50 16,735, field p50 220). 730 (38%) have a field under the floor while the archival tag is over it, and 326 have an archival count of 100,000 or more and get no count on the card. Largest in the sample is 127,356,592 archival loops against a field of 740.

None of this is new in your diff, `totalLoops` was already built on the same field. What the diff changes is that a wrong number becomes no number, on the one case the feature exists for. Preferring `rawTags['loops']` when it's present, or getting `embedded_loops` into the list endpoint, would each fix it, but that's a call rather than a nit so I'm leaving it here instead of pushing something.

mobile/lib/widgets/video_feed_item/video_card_meta.dart
```diff
@@ -0,0 +1,80 @@
...
+
+/// Smallest public count that reads as a recommendation rather than a warning.
+///
+/// A number below this tells a viewer not to bother — which is worse than
+/// silence on content the feed just paid to surface. Above it, the number is
+/// doing useful work: a classic Vine with millions of loops is famous, and
+/// saying so is the point.
+///
+/// Applies to archival Vine counts as well as diVine's own: a small number
+/// discourages a viewer whatever its provenance. Measured against 1000
```
realmeylisdev, line 15: this measurement doesn't describe the number the floor is applied to, and that one's mine, it came from my comment last round.

I measured the `loops` tag. The floor is applied to `originalLoops`, which on the feed path is funnelcake's `loops` field instead (other thread has the trace). Over 1,915 classic vines carrying both: tag p50 16,735 vs field p50 220, and 69% of the fields sit under 1000 rather than the 64% I gave you.

Per surface, measured on the value the code actually compares:

| surface | count hidden by the floor |
|---|---|
| `platform=vine&sort=loops` (classics tab) | 21% |
| `platform=vine&sort=trending` | 92% |
| `sort=trending`, no platform filter (52% classic) | 86% |

Spec line 91 ("Archival Vine numbers usually clear the bar") holds for the tag and not for the field. Probably worth settling together with the other thread, since which number is right depends on which value the floor ends up reading.

docs/superpowers/specs/2026-08-08-video-card-view-count-display-design.md
```diff
@@ -0,0 +1,271 @@
...
+warning and is replaced by the date. Archival Vine numbers usually clear the
+bar and our current view volume usually does not, which is the whole effect —
+but it falls out of one threshold rather than two source-specific branches.
+
+| Case | Date | Count | Renders |
+|---|---|---|---|
+| Flag off | — | `totalLoops` | `12 loops` (today's behavior, unchanged) |
+| Own video | post date | `totalLoops`, always | `3h ago · 12 loops` |
+| Count at or above the floor | post date | public count | `Apr 22, 2014 · 2.1M loops` |
+| Count below the floor | post date | none | `3h ago` |
```
realmeylisdev, line 100: this row isn't always true, and the PR body carries the same table.

When a classic vine has no `published_at` tag its `created_at` is the 2026 import time, so `hasUnknownOriginalDate` suppresses the date as well and both halves come back null. Real record `fa45ae110d145a6a4095dd21a943e1c934a245e7619fb61276ff2a9521897b96` through `VideoStats.fromJson(...).toVideoEvent()`: flag off renders `2.78K loops`, flag on renders nothing at all and the author name loses its subtitle.

Small but not zero: 2.2% of the classics tab and 0.3% of trending. Line 173 already documents it in prose, it's only the tables that promise a date. You said you'd fix the description once the floor settled, and it has.

realmeylisdev (changes requested):
sorry, found a real one. the floor lands on the wrong number for classic vines on the feed path, so the famous ones lose their count instead of keeping it.

ran a live record through `VideoStats.fromJson`: 192,604 archival loops, and with the flag on the card renders `Jan 17, 2017` and no count. also the 64% figure I gave you last round was measured on the wrong field, that's on me. details inline.

realmeylisdev (approved):
pushed all three from last round instead of handing them back, checks green.

the real one is `76d1f050`: the floor now reads `rawTags['loops']` rather than `originalLoops`, so a famous vine keeps its count on the feed path. the flat list record does carry the event tags, it just has no `embedded_loops`, so the tag was sitting right there the whole time. test goes red without the fix. `6b718a67` drops the 64% figure since I couldn't reproduce it on the tag, and `682b9475` adds the missing no-date row to the spec table.

PR body still has the old table and that 64% line if you want to sync it. revert any of the three freely.

mobile/lib/widgets/video_feed_item/video_card_meta.dart
realmeylisdev: fixed in `76d1f050`. `_publicCount` reads `rawTags['loops']` first and keeps `originalLoops` as the fallback for events with no loops tag, so the relay and detail paths are unchanged and the feed path stops comparing the live number. regression test uses the 192,604 / 723 record and fails with `Actual: <null>` if the fix is reverted.

mobile/lib/widgets/video_feed_item/video_card_meta.dart
realmeylisdev: fixed in `6b718a67`. re-measured on the tag and it isn't one number, it splits by surface: nothing on `platform=vine&sort=loops` falls under the bar, about half the vines on `sort=trending` do. replaced the single figure with that spread rather than swapping in another number. worth knowing separately: only about a quarter of the vines on the trending page carry a loops tag at all, and those still fall back to the live field.

docs/superpowers/specs/2026-08-08-video-card-view-count-display-design.md
realmeylisdev: fixed in `682b9475`, added the below-floor + `hasUnknownOriginalDate` row so the table matches the prose underneath it. the same table in the PR description still needs the row.

PR #6876, opened 2026-08-08
fix(video-recorder): make recorder controls readable to screen readers

### LINKED ISSUE #6877: fix(video-recorder): make recorder controls readable to screen readers
Auto-created from PR #6876.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6876
**Branch:** `fix/video-recorder-semantics`

hm21:
Closes #6877

## Description

The recorder's setting buttons were announced as bare actions. A screen-reader user heard "Toggle flash" but never learned whether flash was off, on, or auto — and the same applied to the timer, aspect ratio, camera side, stabilization mode, grid, and ghost frame. Because the icon carries the state visually and nothing carried it semantically, cycling a setting produced no perceivable change at all: you could tap the timer three times and have no way to know what you had selected.

Every control now announces its current setting alongside its label:

- **Multi-state controls** (flash, timer, aspect ratio, camera, stabilization) expose a localized `value`, so the reader says e.g. "Toggle flash, Auto". A `value` string is right here rather than `toggled`, since "off / on / auto" is not a boolean and flattening it would lose the middle state.
- **On/off controls** (grid overlay, ghost frame) expose `toggled` instead, so the platform reader supplies its own localized "on"/"off" and the checked-state affordance users already expect from a switch-like control.

The capture-mode tooltip is now excluded from semantics. The `Semantics` label wraps it and carries the same text, so TalkBack was announcing the label twice.

Three neighbouring gaps on the same screen are fixed in the same pass, since they are the same "recorder is unusable without sight" defect:

- The **camera viewfinder** was a tap target with no semantics at all. It is now a labeled button with a focus hint when tap-to-focus is enabled, and fully excluded from the semantics tree when it is not — an inert preview should not be a stop in the traversal order.
- **Unselected mode-wheel entries** announced only their name ("Classic"), which reads as a static label rather than something actionable. They now announce the outcome ("Switch to Classic mode"); the selected entry keeps its plain name because `selected: true` already conveys the rest. The inner `Text` is excluded so the name isn't read a second time.
- The **capture top bar's** close and next buttons said "Close" and "Next" with no object. They now reuse the existing `videoRecorderCloseLabel` / `videoRecorderContinueToEditorLabel` keys, which say where they lead.

Supporting changes: `DivineIconButton` gains a `semanticToggled` passthrough so the shared component covers the on/off case instead of each caller hand-rolling a `Semantics` wrapper, and the decorative sticker in `VineBottomSheetPrompt` is excluded from the semantics tree.

All new strings are translated in all 17 non-English locales, not left as English fallbacks.

Bildschirmfoto 2026-08-08 um 10 14 46

**Related Issue:** 

## Out of Scope

- The classic-mode top bar still uses the older `videoRecorderCaptureCloseLabel` / `videoRecorderCaptureNextLabel` ("Close" / "Next") keys. Same weakness as the capture top bar, but it is a separate screen and the two keys are still referenced there — folding it in would widen the diff without adding to the story here.

## Verification

- `flutter test test/widgets/video_recorder test/l10n/arb_consistency_test.dart` — 277 passing
- `flutter test --coverage` in `packages/divine_ui` — 855 passing, coverage 100.00%
- Design-system ceilings and structure ratchets (`raw_textstyle`, `raw_colors`, `material_button`, `raw_dialog`, `untested_services`, `test_unit_structure`) — all pass
- Manual pass on a real Samsung Galaxy with TalkBack: cycle flash, timer, aspect ratio and camera and confirm the announced value changes each time; toggle grid and ghost frame and confirm the on/off state is read back; swipe onto the viewfinder and the mode wheel and confirm both announce and activate; confirm the flash label is no longer announced twice.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


NotThatKindOfDrLiz (approved):
Approved after follow-up fix. I pushed f309da338 to cover the remaining classic recorder top-bar accessibility gap: classic close/next now use the same contextual labels as capture mode, the obsolete bare capture labels were removed from ARB/generated l10n, and focused classic semantics tests were added.\n\nValidation run:\n- flutter gen-l10n\n- flutter test test/l10n/arb_consistency_test.dart\n- flutter test test/widgets/video_recorder/modes/classic/video_recorder_classic_top_bar_test.dart test/widgets/video_recorder/modes/capture/video_recorder_capture_top_bar_test.dart\n- flutter analyze\n- pre-push analyzer + changed-file tests\n- GitHub CI: all checks passing\n\nAlex, please review the follow-up commit and merge when you're ready.

PR #6878, opened 2026-08-08
test(helpers): give each test process its own path_provider directories

rabble:
Refs #6879

## Description

The shared test harness used fixed `path_provider` directories for every unit-test process:

```dart
getApplicationDocumentsDirectory -> /tmp/documents
getApplicationSupportDirectory -> /tmp/support
getTemporaryDirectory -> /tmp
```

`very_good test --optimization` runs the merged bundle plus each `skip_very_good_optimization` file as separate concurrent processes. With fixed directories, those processes could share Hive boxes, Drift databases, and other filesystem-backed test state.

This PR moves the canonical test `path_provider` directories under a per-process root created by `Directory.systemTemp.createTempSync('divine_test_${pid}_')`. The app-facing `PathProviderPlatform.instance` mock and the raw MethodChannel handler now resolve to the same process-local `temporary`, `documents`, and `support` directories.

## Change

- Key the canonical `path_provider` test directories on the current process id in `test_setup.dart`.
- Keep the raw MethodChannel path provider handler aligned with `PathProviderPlatform.instance`.
- Remove two hand-rolled path provider MethodChannel overrides that duplicated the canonical handler.
- Add contract coverage for the per-process directories and MethodChannel/platform-instance agreement.
- Move `draft_storage_service_test.dart` off fixed `/tmp/documents` so its recursive cleanup only deletes that test's own temp tree.
- Extend the heal-and-blame harness to restore a leaked `PathProviderPlatform.instance`, matching its existing shared-channel guard.
- Fix stale comments left behind by removing path provider overrides.
- Delete the per-process root at the end of the suite, so a run does not leave
  its Hive boxes and databases in the system temp directory.
- Keep the harness compiling on the web platform: the per-process root is
  filesystem work, and `flutter_test_config.dart` runs for web suites too.

## What this does not claim

It is not established that this fully fixes the shard-2 flake tracked in #6879. The original PR diagnosis named some suites and timing details that did not hold up after review: the first two edits were on a dead path because `path_provider` resolves through `PathProviderPlatform.instance`, and the standalone upload suites already redirect the platform instance to their own temp directories before touching Hive.

What does hold up: the canonical test harness really did use fixed `/tmp` paths, and at least one bundled suite still recursively deleted `/tmp/documents`. This PR removes that shared filesystem resource and the nearby fixed-path cleanup pattern. Keep #6879 open until shard 2 stays green over enough runs to prove the broader flake is gone.

## Verification

- `flutter test test/services/draft_storage_service_test.dart`
- `flutter test test/test_setup_test.dart test/helpers/shared_channel_override_test.dart`
- `very_good test --optimization --concurrency=4 --exclude-tags integration`
- `CHROME_EXECUTABLE=... flutter test --platform chrome test/utils/path_resolver_test.dart test/test_setup_test.dart`

## Type of Change

- [x] Bug fix (test infrastructure)

### DIFF
--- a/mobile/test/helpers/real_integration_test_helper.dart
+++ b/mobile/test/helpers/real_integration_test_helper.dart
@@ -1,6 +1,8 @@
 // ABOUTME: Helper utilities for real integration tests without over-mocking
 // ABOUTME: Provides real Nostr relay connections and minimal platform channel mocking
 
+import 'dart:io';
+
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:nostr_client/nostr_client.dart';
@@ -14,6 +16,32 @@ import 'shared_channel_override.dart';
 /// Only mocks platform channels that can't be tested, uses real Nostr connections
 class RealIntegrationTestHelper {
   static bool _isSetup = false;
+
+  /// Per-process root for the directories `path_provider` hands back.
+  ///
+  /// `very_good test --optimization` runs the merged bundle and every
+  /// `skip_very_good_optimization` file as *separate, concurrent* processes.
+  /// These paths were hardcoded to `/tmp/test_documents` and
+  /// `/tmp/test_support`, so every one of those processes resolved to the
+  /// same directory — and therefore the same Hive boxes on disk.
+  ///
+  /// Sixteen suites open the `pending_uploads` box; eleven live in the
+  /// bundle and five run standalone. When a standalone suite called
+  /// `Hive.deleteBoxFromDisk('pending_uploads')` while a bundled one was
+  /// reading it, whichever lost the race failed on a box it did not mutate —
+  /// a seed-dependent failure with no connection to the diff under test.
+  /// Keying the root on the process id removes the shared resource.
+  static final Directory _processRoot = Directory.systemTemp.createTempSync(
+    'divine_test_${pid}_',
+  );
+
+  static final Directory _documentsDirectory = Directory(
+    '${_processRoot.path}/documents',
+  )..createSync(recursive: true);
+
+  static final Directory _supportDirectory = Directory(
+    '${_processRoot.path}/support',
+  )..createSync(recursive: true);
 
   static const MethodChannel _prefsChannel = MethodChannel(
     'plugins.flutter.io/shared_preferences',
@@ -135,13 +163,13 @@ class RealIntegrationTestHelper {
 
     overrideSharedChannel(_pathProviderChannel, (MethodCall methodCall) async {
       if (methodCall.method == 'getApplicationDocumentsDirectory') {
-        return '/tmp/test_documents';
+        return _documentsDirectory.path;
       }
       if (methodCall.method == 'getTemporaryDirectory') {
         return '/tmp';
       }
       if (methodCall.method == 'getApplicationSupportDirectory') {
-        return '/tmp/test_support';
+        return _supportDirectory.path;
       }
       return null;
     });
--- a/mobile/test/helpers/real_integration_test_helper.dart
+++ b/mobile/test/helpers/real_integration_test_helper.dart
@@ -20,7 +20,7 @@ class RealIntegrationTestHelper {
   /// Per-process root for the directories `path_provider` hands back.
   ///
   /// `very_good test --optimization` runs the merged bundle and every
-  /// `skip_very_good_optimization` file as *separate, concurrent* processes.
+  /// VGV-skip-tagged file as *separate, concurrent* processes.
   /// These paths were hardcoded to `/tmp/test_documents` and
   /// `/tmp/test_support`, so every one of those processes resolved to the
   /// same directory — and therefore the same Hive boxes on disk.
--- a/mobile/test/widgets/user_profile_tile_layout_test.dart
+++ b/mobile/test/widgets/user_profile_tile_layout_test.dart
@@ -649,10 +649,9 @@ void _setupPlatformMocks() {
     'plugins.flutter.io/path_provider',
   );
   // Per-process directories, not the shared `/tmp/test_*` these used to
-  // return: very_good runs the merged bundle and each
-  // skip_very_good_optimization file concurrently, and a shared path puts
-  // them all on the same Hive boxes on disk. See
-  // RealIntegrationTestHelper._processRoot.
+  // return: very_good runs the merged bundle and each VGV-skip-tagged
+  // file concurrently, and a shared path puts them all on the same Hive
+  // boxes on disk. See RealIntegrationTestHelper._processRoot.
   final processRoot = Directory.systemTemp.createTempSync(
     'divine_tile_layout_${pid}_',
   );
--- a/mobile/test/widgets/user_profile_tile_layout_test.dart
+++ b/mobile/test/widgets/user_profile_tile_layout_test.dart
@@ -1,6 +1,8 @@
 // ABOUTME: Layout-focused TDD tests for UserProfileTile addressing widget display bugs
 // ABOUTME: Tests responsive layout, element positioning, state visibility, and edge case rendering
 
+import 'dart:io';
+
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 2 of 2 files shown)


hm21:
Data point for the "What this does NOT claim" section — a fresh hit on the exact test you list, from today.

#6893 is a one-file change to a video-metadata preview widget. It touches nothing under `services/`, `test/helpers/`, or any `pending_uploads` suite. Its shard 2 failed a **different** test on each of three runs:

| Run | Failing test |
|---|---|
| [93104796556](https://github.com/divinevideo/divine-mobile/actions/runs/31258203428/job/93104796556) | `ClipSpeedRenderService caps concurrent native renders so the encoder does not stall` |
| [93105348453](https://github.com/divinevideo/divine-mobile/actions/runs/31258203428/job/93105348453) | `ClipThumbnailImage renders the neutral placeholder for a missing file` |
| [93106243939](https://github.com/divinevideo/divine-mobile/actions/runs/31258203428/job/93106243939) | `UploadManager.getUploadByFilePath should return first match when multiple uploads have same path` — `Expected: <1>, Actual: <2>` |

The third is `unit/services/upload_manager_get_by_path_test`, the first of the five standalone suites in your table, failing on a count of matching uploads it never wrote. That is the shape your diagnosis predicts, on a branch that cannot have caused it.

Not proof — it is one more observation, and it does not rule out a second source in that suite. But it is a second independent sighting of the same suite failing on state from elsewhere, and it is not the 2026-08-08 run you already listed.

The `ClipThumbnailImage` failure in run 2 is a separate defect with its own fix in #6874 (post-frame `ImageCache` disposal landing in the next test's first frame). The `ClipSpeedRenderService` one in run 1 is a third, unrelated race — `pumpEventQueue()` racing a real `File.writeAsString` in the test's fake — which I am fixing separately. Shard 2 has at least three independent pollution sources right now; yours is the only one with a fix that removes a whole class of them.

hm21 (approved):
Approved with one takeover commit (581ca2fa84) — the original two edits were on a dead path.

`path_provider`'s Dart API resolves through `PathProviderPlatform.instance`, and `flutter_test_config.dart` → `setupTestEnvironment()` points that at `MockPathProviderPlatform` for every suite. The instance shadows the MethodChannel, so the two mocks this PR rewrote were never consulted. Probe before and after `de6d2b8`, calling `getApplicationDocumentsDirectory()` right after `RealIntegrationTestHelper.setupTestEnvironment()`:

```
main:       docs=/tmp/documents support=/tmp/support
de6d2b8:    docs=/tmp/documents support=/tmp/support
```

Same reason the named victim can't be the one racing: `upload_manager_get_by_path_test` already redirects to its own `Directory.systemTemp.createTemp(...)` in `setUp`, with a comment saying exactly that, and its Hive cleanup runs after the redirect — not in `setUpAll` as the description says. Every `skip_very_good_optimization` file that touches Hive already does the same, so nothing was left on `/tmp/test_documents` to collide.

The shared path that *was* real is one layer down: the canonical `/tmp/documents` and `/tmp/support` in `test_setup.dart`, which every suite in every process resolved to. The commit moves the per-process root there, where the instance mock and the channel handler both read it:

```
docs=…/T/divine_test_85231_HaOEIU/documents
```

`getTemporaryDirectory` moves off `/tmp` too — `drift_setup_test` builds a fixed `local_relay.db` path underneath it, same class one directory over. With the canonical paths per-process the two channel overrides became exact copies of the canonical handler, so they're removed rather than kept in sync; `test_setup_test.dart` pins the contract and fails on a mutation back to a fixed path.

Verification: full `flutter test` green (15320 passed / 305 skipped / 0 failed), `flutter analyze lib test integration_test` clean, `dart format` 0 changed, all CI checks green. One local failure in `video_recorder_bloc_test` on a first run was CPU contention from a concurrent `flutter analyze` — clean on re-run and on the unmodified head.

Two things left for you: `integration_test/helpers/real_integration_test_helper.dart` still has the `/tmp/test_*` pair (device-side, single process — left alone as out of scope), and the flake this was chasing stays unproven either way, so #6879 should stay open until shard 2 holds green over several runs.

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup.

I verified the rewritten PR body, the draft_storage_service_test temp-dir fix, the PathProviderPlatform heal-and-blame coverage, and the stale comment cleanup. Local targeted tests, analyzer, process-global mutation guard, and material-button guard passed. CI is green on head 635a55dc9, including all four test shards.

NotThatKindOfDrLiz:
@rabble I took this over and pushed the cleanup from the re-review.

What changed after the original diff:
- rebased onto current origin/main
- updated the PR body to match the final implementation and changed the issue link to Refs #6879 so the flake stays open until proven fixed
- moved draft_storage_service_test.dart off fixed /tmp/documents and into a per-test temp directory
- extended healAndBlameSharedChannels to detect and restore leaked PathProviderPlatform.instance
- added/adjusted coverage for the path_provider platform restore path
- fixed stale/ambiguous comments left by removing the duplicate path_provider overrides

Local verification passed:
- flutter test test/services/draft_storage_service_test.dart
- flutter test test/test_setup_test.dart test/helpers/shared_channel_override_test.dart
- bash scripts/check_process_global_mutations.sh
- bash scripts/check_material_button_ceiling.sh

CI is green on head 635a55dc9, including Generated Files and Tests shards 0/4 through 3/4. Please review the takeover commits before merging.

mobile/test/test_setup.dart
```diff
@@ -91,11 +103,64 @@ void _installShared(MethodChannel channel, SharedChannelHandler handler) {
...
+/// `createTempSync` root removes the shared resource; the uniqueness comes
+/// from the random suffix it appends, and the process id in the prefix is a
+/// label that makes a stray directory traceable. Do not "simplify" this to a
+/// fixed `divine_test_$pid` path — pids recycle, so that would reintroduce the
+/// collision across runs.
+///
+/// On web there is no filesystem and every `dart:io` call throws
+/// `Unsupported operation: _Namespace`, so the paths stay plain literals.
+/// Nothing writes to them, so nothing can collide.
+final String _processRootPath = kIsWeb
```
dcadenas: This needed a web guard before it could land.

`flutter_test_config.dart` runs for web suites too, so evaluating `Directory.systemTemp.createTempSync` at library-initialization time killed every one of them during suite loading:

```
Failed to load "test/utils/path_resolver_test.dart": Unsupported operation: _Namespace
  ../test_setup.dart.js  get _processRoot
```

The same command passes at the merge-base, so it came in with this change. It is not hypothetical either — `test/utils/path_resolver_test.dart` exists to pin web behaviour, and its first case is "getDocumentsPath returns empty string on web without using path_provider".

Web now gets plain literals, the filesystem work is off-web only, and the contract group skips there.

mobile/test/test_setup.dart
```diff
@@ -91,11 +103,64 @@ void _installShared(MethodChannel channel, SharedChannelHandler handler) {
...
+
+/// Delete this test process's `path_provider` root.
+///
+/// Wired as a suite-level `tearDownAll` in `flutter_test_config.dart`. Each
+/// process gets its own root, so nothing reclaims it the way the old fixed
+/// `/tmp/documents` was reused across runs: without this, every test process
+/// would leave its Hive boxes and databases in the system temp directory.
+///
+/// No-op on web, where there is no root to delete.
+void deleteTestPathProviderRoot() {
```
dcadenas: A per-process root is never reused, so unlike the fixed `/tmp/documents` it replaced, nothing ever reclaimed it. Each test process left its Hive boxes and databases in the system temp directory.

Measured on the original head: one `flutter test test/test_setup_test.dart` left one `divine_test_*` directory behind. With the suite-level `tearDownAll`, a full `very_good test` run leaves none.

The delete is best-effort on purpose — a file still held open must not fail the suite.

dcadenas (approved):
✅ the shared harness stops being a shared resource — two fixes pushed, detail inline.

## What I pushed

- Delete the per-process root when the suite ends. Nothing reclaimed it, so every run left its Hive boxes and databases in the system temp directory.
- Keep the harness compiling on the web platform. The eager filesystem work broke `flutter test --platform chrome` for every web suite.

## Checks at this head

- `very_good test --optimization --concurrency=4 --exclude-tags integration`: 15399 passed, 285 skipped, 0 failed, and no `divine_test_*` directory left behind.
- `flutter analyze lib test integration_test` and `dart format --set-exit-if-changed` clean.
- Chrome: `test/utils/path_resolver_test.dart` and `test/test_setup_test.dart` pass.
- Process-global, shared-channel, test-structure and skip ratchets all pass.

The flake in #6879 stays unproven either way, which your description already says.

PR #6881, opened 2026-08-08
fix(a11y): make library and clip controls readable to screen readers

### LINKED ISSUE #6882: fix(a11y): make library and clip controls readable to screen readers
Auto-created from PR #6881.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6881
**Branch:** `fix/library-semantics`

hm21:
Closes #6882

## Description

Screen-reader passes over the library and the clip widgets surfaced several controls that were announced wrongly, announced as static text, or not announced at all. This fixes the semantics for those controls.

**Library toolbar** — every icon button now carries a label describing its own action instead of borrowing copy from somewhere else. The leading button announced `commonCancel` ("Cancel") in selection mode and nothing when it just closes the screen; the trash button announced the tab-entry label; sort and select announced nothing at all; the destructive button announced a bare "Delete" with no object. Five new l10n keys (`libraryClose…`, `libraryStopSelectingClips…`, `libraryOpenTrash…`, `librarySortClips…`, `librarySelectClips…`) are translated into all 22 locales, and the delete button reuses the existing `libraryDeleteSelectedClipsTooltip`. `libraryTrashEntryLabel` lost its last reader in the process and is deleted rather than left behind as a dead string in 22 locales — it was a noun ("Recently deleted") where the a11y label needs a verb, so reusing it later would be the wrong move anyway.

**`DivineButton` / `VineBottomSheetSelectionOption`** — both wrap their content in `Semantics(button: true, label: …)` but never forwarded a tap action, so the node was exposed as labelled text rather than an activatable button, and the wrapper label was concatenated with the child's visible text. They now pass `onTap` (null when disabled, so the disabled state stays truthful) and `excludeSemantics: true` so the explicit label wins. The selection option additionally reports `selected:`, which is what makes the checked state audible in the sort menu. `DivineButton.semanticLabel`'s doc is updated to match: on a labelled button it now *replaces* the visible text in the semantics tree instead of supplementing it, which is exactly what the toolbar's "Select" chip depends on.

`DivineIconButton` deliberately keeps its existing wrapper. Its `Semantics` has no `Text` child to conflict with, so it merges with the child `InkWell` and already reports `tap`/`button`/`enabled` correctly — verified with a throwaway probe test before ruling it out.

**`VideoClipPreview`** — the scrim is dismiss-on-tap, but a bare `GestureDetector` contributes no semantics, so there was no non-gesture way to leave the preview. It is now an explicit labelled button node ("Close video preview", existing key) with `explicitChildNodes: true` so the media, title, and action buttons underneath stay individually reachable. Keeping it as an ancestor of the content rather than a sibling behind it mirrors the touch behaviour: the scrim is `HitTestBehavior.translucent`, so a tap anywhere that isn't an action button already dismisses.

**`VideoClipThumbnailCard`** — the duration badge was read out as a bare number ("5.00") right after the card's own "Video clip, N seconds" label, which is redundant and confusing; it is now excluded. `container: true` plus `excludeFromSemantics` on the inner `GestureDetector` keeps one node per card instead of the card label and the gesture node competing.

Bildschirmfoto 2026-08-08 um 10 59 22

## Out of Scope

Only the library toolbar and the two video-clip widgets were audited. The same "labelled but not tappable" wrapper pattern likely exists elsewhere; the `divine_ui` fixes make those callers correct for free, but no other screen was reviewed in this PR.

## Verification

- `flutter test test/widgets/library/library_toolbar_test.dart test/widgets/video_clip/video_clip_preview_test.dart test/widgets/video_clip/video_clip_thumbnail_card_test.dart test/l10n/arb_consistency_test.dart` — 50 tests pass
- `flutter test --coverage` in `mobile/packages/divine_ui` — 854 tests pass, coverage still 100%
- `flutter gen-l10n` re-run after the key removal; regenerated `lib/l10n/generated/` committed
- `dart format` clean on all touched files

Manually verified on a real Samsung Galaxy with TalkBack enabled:

1. Opened the library — swiped through the toolbar; each button announces its own action (close, recently deleted, sort, select).
2. Entered selection mode — the leading button announces "Stop selecting clips", the destructive button announces "Delete selected clips".
3. Opened the sort sheet — each option announces as a button and the active one announces as selected.
4. Focused a clip thumbnail — one node, "Video clip, N seconds", with no separate "5.00" read out for the duration badge.
5. Opened a clip preview — the scrim announces "Close video preview" and double-tapping it dismisses the preview.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


dcadenas (approved):
The toolbar and clip-preview work here is solid, and the `divine_ui` changes fix the labelled-but-not-tappable wrapper for every caller rather than just this screen.

I pushed three follow-ups on top instead of sending them back. Selecting a clip announced a bare ordinal with no context, and a stop-motion clip announced a duration the grid deliberately hides because it is misleading — both now ride in the card's own label and value, with the badges made decorative. The sort-sheet option change had no test, so it has one now. Two localization keys fell out of use as a result and are removed the same way `libraryTrashEntryLabel` was.

One behaviour worth knowing: the ordinal is announced only while the selection badge is actually on screen. Opening the library with clips already in the editor pre-selects them without showing badges, so an unconditional ordinal would name a position the user cannot see. I pinned the announcements with semantics-tree tests rather than repeating your on-device pass, so a short TalkBack sweep of the grid before merge is worth it. Analyze, the clip, library and localization suites, and the `divine_ui` package suite all pass, and every check is green on the current head.

PR #6883, opened 2026-08-08
fix(a11y): honour reduced motion in tab transitions and the recorder mode wheel

### LINKED ISSUE #6884: fix(a11y): honour reduced motion in tab transitions and the recorder mode wheel
## Problem

Reduced motion is ignored across tab transitions and the recorder mode wheel. With the platform reduced-motion setting enabled, switching tabs still plays the full `TabBar` indicator slide and `TabBarView` page transition, and the recorder mode wheel still animates its snap scroll, selected pill width, and label colour.

## Expected

- Tab transitions should complete instantly when `MediaQuery.disableAnimations` is true.
- The recorder mode wheel should jump/snap without motion when reduced motion is enabled.
- Normal animation behaviour should remain unchanged when reduced motion is disabled.

## Scope

Covered by PR #6883:

- Inbox, Explore, Library, Profile grid, and audio selection tab controllers.
- Recorder mode wheel snap scroll, pill resize, and label colour tween.
- A small `MediaQuery`/duration helper used by those call sites.

Out of scope:

- Broader cleanup of remaining `MediaQuery.xxxOf` / `MediaQuery.of(context).x` call sites.
- Other reduced-motion gaps outside these tab bars and the mode wheel.

hm21:
Closes #6884

## Description

Reduced motion was ignored across the app's tab screens and the recorder mode wheel. Turning the system setting on and switching tabs still played the full indicator slide and page transition.

**Tab screens (5 sites).** `TabController.animationDuration` drives both the `TabBar` indicator and the `TabBarView` page transition, and Flutter short-circuits both when it is `Duration.zero` (`_warpToAdjacentTab` jumps instead of animating; `_changeIndex` sets the controller value directly). There is nothing to gate in the widget tree — the duration is the only lever. It is fixed at construction, though, so reacting to a mid-session toggle means building a replacement controller.

`ReducedMotionTabControllerMixin` owns that lifecycle once: it rebuilds when the duration or `tabCount` changes, carries the open tab over clamped into the new range, and moves `onTabChanged` onto the replacement. That subsumes the hand-written recreate paths explore and profile already had for their changing tab sets, so those two shrink rather than grow. Both keep their own index resolution — explore deliberately carries the tab **by name**, because indices shift when optional tabs appear or disappear.

The mixin requires `TickerProviderStateMixin`; `SingleTickerProviderStateMixin` permits one ticker per `State` and throws on the replacement controller. Two sites are switched over.

**MediaQuery shorthands.** This PR adds only the `BuildContext` shorthands it uses: `reduceMotion` and `textScaler`. Each forwards to the aspect-scoped `MediaQuery.xxxOf` form, so readers only rebuild when the metric they use changes. `Duration.autoReduceMotion(context)` pairs with `reduceMotion` for animation durations.

One sharp edge is documented on `autoReduceMotion`: `ScrollPosition.animateTo` asserts `duration > Duration.zero`, so scroll positions must branch and call `jumpTo`. The recorder's snap scroll does exactly that.

**Recorder mode wheel.** Its snap scroll, pill resize, and label colour tween ignored the setting; they now honour reduced motion. The tap handler also relies on the single `_selectIndex(..., animate: true)` move instead of scheduling a duplicate jump when reduced motion completes synchronously.

The five tab screens and the mode wheel are the only behaviour changes. Broader `MediaQuery.xxxOf` / `MediaQuery.of(context).x` cleanup is left out of this behaviour fix.

**Related Issue:** #6884

## Out of Scope

- Migrating the remaining `MediaQuery.xxxOf` / `MediaQuery.of(context).x` call sites to shorthands.
- Other reduced-motion gaps outside tab bars and the mode wheel.

## Verification

- `flutter test test/mixins/ test/extensions/ test/notifications/ test/screens/explore/ test/widgets/profile/ test/widgets/video_recorder/video_recorder_mode_selector_test.dart` — 721 passing.
- `flutter test test/screens/library_screen_test.dart test/widgets/video_editor/audio_editor/` — 80 passing.
- Each new reduced-motion test was checked to fail without its fix, not just to pass with it (tab tests: `animation.value` 0.0 instead of 3.0 after one frame; wheel test: pill width 130.7 mid-tween instead of the settled 117.9).
- [x] Manual pass on a real Samsung Galaxy — reduced motion on and off, tab switches on Inbox / Explore / Library / Profile / audio sheet, and the recorder mode wheel.

## Type of Change

- [x] 🧹 Code refactor

### DIFF
--- a/mobile/lib/extensions/media_query_extensions.dart
+++ b/mobile/lib/extensions/media_query_extensions.dart
@@ -1,0 +1,74 @@
+// ABOUTME: BuildContext shorthands for the MediaQuery metrics read most often,
+// ABOUTME: plus Duration.autoReduceMotion for gating animation lengths on them.
+// ABOUTME: Each getter forwards to the aspect-scoped MediaQuery.xxxOf form.
+
+import 'package:flutter/widgets.dart';
+
+/// Shorthands for the handful of `MediaQuery` metrics this app reads
+/// everywhere.
+///
+/// Every getter forwards to the matching `MediaQuery.xxxOf(context)`, so the
+/// reader only rebuilds when that one metric changes. Reaching for
+/// `MediaQuery.of(context).size` instead subscribes the widget to every
+/// metric — a keyboard opening then rebuilds a widget that only cared about
+/// the screen size.
+///
+/// Deliberately partial. Only the metrics with real call-site pressure live
+/// here; rarer ones (`orientation`, `platformBrightness`, `boldText`, …) stay
+/// explicit `MediaQuery.xxxOf` calls so this doesn't drift into a second copy
+/// of the `MediaQuery` API.
+extension MediaQueryExtensions on BuildContext {
+  /// Whether the platform asked for reduced motion.
+  ///
+  /// Gate animations on this and fall back to `Duration.zero` or a static
+  /// frame — see `.claude/rules/accessibility.md`.
+  bool get reduceMotion => MediaQuery.disableAnimationsOf(this);
+
+  /// System text scaling to apply to font sizes.
+  TextScaler get textScaler => MediaQuery.textScalerOf(this);
+
+  /// Size of the media in logical pixels — the window, not this widget.
+  ///
+  /// Named `screenSize` rather than `size` because `BuildContext.size`
+  /// already means the render box's own size.
+  Size get screenSize => MediaQuery.sizeOf(this);
+
+  /// Display areas obscured by system UI, with anything already covered by
+  /// [viewInsets] (typically the keyboard) subtracted out.
+  EdgeInsets get padding => MediaQuery.paddingOf(this);
+
+  /// Display areas obscured by system UI, ignoring [viewInsets].
+  ///
+  /// Prefer this over [padding] for layout that must not jump when the
+  /// keyboard opens.
+  EdgeInsets get viewPadding => MediaQuery.viewPaddingOf(this);
+
+  /// Display areas fully obscured by system UI — in practice, the keyboard.
+  EdgeInsets get viewInsets => MediaQuery.viewInsetsOf(this);
+
+  /// Physical pixels per logical pixel.
+  double get devicePixelRatio => MediaQuery.devicePixelRatioOf(this);
+}
+
+/// Lets an animation length opt into the platform's reduced-motion setting.
+extension ReduceMotionDuration on Duration {
+  /// This duration, or [Duration.zero] when [BuildContext.reduceMotion] is on.
+  ///
+  /// ```dart
+  /// AnimatedContainer(
+  ///   duration: const Duration(milliseconds: 200).autoReduceMotion(context),
+  /// )
+  /// ```
+  ///
+  /// A zero duration makes implicit animations jump straight to their target,
+  /// which is what reduced motion asks for. Use this for tweens between two
+  /// legible states; an animation whose *end* state is the only readable one
+  /// (a spinner, a marquee) needs a static replacement instead, not a
+  /// zero-length run.
+  ///
+  /// Not every API tolerates a zero duration. `ScrollPosition.animateTo`
+  /// asserts `duration > Duration.zero`, so scroll positions have to branch on
+  /// [BuildContext.reduceMotion] and call `jumpTo` instead.
+  Duration autoReduceMotion(BuildContext context) =>
+      context.reduceMotion ? Duration.zero : this;
+}
--- a/mobile/lib/extensions/media_query_extensions.dart
+++ b/mobile/lib/extensions/media_query_extensions.dart
@@ -1,11 +1,10 @@
-// ABOUTME: BuildContext shorthands for the MediaQuery metrics read most often,
+// ABOUTME: BuildContext shorthands for the MediaQuery metrics this PR uses,
 // ABOUTME: plus Duration.autoReduceMotion for gating animation lengths on them.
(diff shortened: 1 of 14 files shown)


NotThatKindOfDrLiz:
@hm21 I pushed a small takeover commit, `48ccb2bcf`, to address the review follow-ups:

- trimmed `media_query_extensions.dart` to only the production-used `reduceMotion` and `textScaler` shorthands, plus `Duration.autoReduceMotion(context)`;
- tightened `ExploreTabBar` / `ExploreTabView` to require a non-null `TabController` and removed the unreachable null branch;
- removed the duplicate recorder mode wheel `_moveTo` call that could fire after the reduced-motion jump completed synchronously.

I also updated the PR body and #6884 so the metadata matches the final scope.

Verification:

- `flutter test test/extensions/media_query_extensions_test.dart test/widgets/video_recorder/video_recorder_mode_selector_test.dart`
- `flutter test test/mixins/ test/extensions/ test/notifications/ test/screens/explore/ test/widgets/profile/ test/widgets/video_recorder/video_recorder_mode_selector_test.dart`
- `flutter test test/screens/library_screen_test.dart test/widgets/video_editor/audio_editor/`
- GitHub checks are green on `48ccb2bcf`

NotThatKindOfDrLiz (approved):
Approved after takeover fixes in 48ccb2bcf. CI is green; author should review the takeover commit and merge when ready.

NotThatKindOfDrLiz:
## Review Summary

**Assessment:** Core mixin design is sound and the adoption across five sites is safe. The PR delivers genuine simplification of two pre-existing hand-rolled TabController recreation paths. However, three mechanical issues block approval.

---

## Bar Assessment

✅ **Correctness**: All 16 CI checks green. No runtime errors, no regression in TabController lifecycle, no breaking changes to existing reduced-motion behavior.

✅ **Safety**: Mixin lifecycle verified—`didChangeDependencies` called before `build` ensures `tabController` exists. TickerProviderStateMixin allows multiple controllers. Duration.zero is safe for animations.

❌ **Technical Debt**: Speculative API in `media_query_extensions.dart` violates the [no-dead-code rule](https://github.com/divinevideo/divine-mobile/blob/main/.claude/rules/agent_workflow.md#4-do-not-accumulate-technical-debt). Five unused getters ship with this PR.

❌ **Adjacent Cleanup**: Dead null-check branches in `ExploreTabBar`/`ExploreTabView` should be removed. Redundant fallback in mode selector should be simplified.

---

## Required Fixes

### 1. Trim `media_query_extensions.dart` to Production Call Sites

Only **two** getters have actual production usage:
- `reduceMotion` — used in mode selector, mixin, multiple widgets
- `textScaler` — used in profile grid

The remaining five are speculative:
- `screenSize`, `padding`, `viewPadding`, `viewInsets`, `devicePixelRatio` — zero call sites

**Action:** Remove the five unused getters. Open a follow-up issue to audit all ~120 `MediaQuery` call sites in the codebase and centralize the ones worth exposing.

**Why:** Shipping speculative API invites downstream abuse and makes the extension a junk drawer. The bar is [resolve adjacent prior debt](https://github.com/divinevideo/divine-mobile/blob/main/.claude/rules/agent_workflow.md), not accumulate new speculation.

---

### 2. Tighten `ExploreTabBar` and `ExploreTabView.controller` to Non-Nullable

**File:** `mobile/lib/screens/explore/explore_view.dart`

The `controller` parameter in both `ExploreTabBar` and `ExploreTabView` is nullable but:
- The widgets always receive a controller from `syncTabController` (which is always called in `build`)
- Both widgets guard with `if (controller == null) return SizedBox.shrink()`
- No scenario reaches that branch

**Action:** Change parameter type to `required TabController controller`, remove all `controller!` bangs and `if (controller != null)` guards.

**Code locations:**
- `ExploreTabBar(required this.controller, ...)` 
- `ExploreTabView(required this.controller, ...)` 
- Callsites: `ExploreView.build()` passes non-null controller

**Why:** The null checks are dead code. Non-nullable parameters make the contract explicit and delete branch coverage debt in one stroke.

---

### 3. Remove Redundant `_moveTo()` Call in Mode Selector

**File:** `mobile/lib/widgets/video_recorder/video_recorder_mode_selector.dart`, line ~262

Current code:
```dart
if (!_isSnapping) _moveTo(i);
```

This guard is redundant because `_moveTo()` already has an `animate` parameter that is set based on reduced-motion state. The function branches internally:
```dart
_moveTo(i, animate: !context.reduceMotion);
```

**Action:** Remove the outer `if (!_isSnapping)` guard. Rename the guard to clarify intent:
```dart
// Old: if (!_isSnapping) _moveTo(i);
// New: call _moveTo with animate parameter, which handles reduced motion
_moveTo(i);
```

**Why:** The guard condition loses semantic meaning after adoption of reduced-motion branching. It's clearer to let the animate parameter (already driven by `reduceMotion`) control the behavior, than to layer on an extra condition that doesn't scale.

---

## Optional Enhancement

Consider adding one reduced-motion widget test to `explore_view_test.dart` that validates tab navigation jumps (not animates) when reduced motion is enabled. Not required for approval, but would close the gap in coverage for the riskiest path (explore's build-time `syncTabController` call).

---

## Approval Path

Once all three fixes land (trimmed extension, tightened nullability, simplified mode selector), this PR is approved. The mixin adoption is solid, the three adopted screens show real simplified logic, and the tests adequately cover the new paths.

---

Ping me once fixes are pushed and I'll review the update.

PR #6885, opened 2026-08-08
fix(a11y): make video editor controls readable to screen readers

### LINKED ISSUE #6886: fix(a11y): make video editor controls readable to screen readers
Auto-created from PR #6885.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6885
**Branch:** `fix/editor-semantics`

hm21:
Closes #6886

## Description

Screen reader passes over the video editor were largely unusable. Four separate problems:

**Every sub-editor toolbar announced the same "Close" / "Done".** Draw, Text, Filter, Adjust, Audio and Voice over all share `VideoEditorToolbar`, which fell back to two generic labels when a call site passed none. With TalkBack on there was no way to tell which tool you were about to dismiss or confirm. Each call site now passes a tool-specific label — "Discard changes in Draw", "Apply changes in Adjust" — built from two new parameterized keys plus the tool names that already existed in the ARB. The main editor's own bar gets "Close video editor" / "Continue to post details", which is what those two buttons actually do.

The labels are now `required` rather than nullable-with-fallback. That is the reason the diff touches every call site: a silent default is exactly what let this regress unnoticed, so the next new sub-editor has to make a choice instead of inheriting "Done".

**The editor canvas leaked its whole render tree into the semantics tree.** `pro_image_editor` builds a large internal widget tree that a screen reader walks through as unreachable, unlabelled noise before it can get to the controls. Wrapped in `ExcludeSemantics` — the interactive affordances live in our own overlays, not in that subtree. Same for the caption-style sheet's live preview, which is decorative.

This also drops the text layers themselves from the canvas semantics, which is intended. Their content is already announced in the timeline, and the only thing the canvas adds is dragging them — which `pro_image_editor` exposes no position information for, so a screen reader could not act on it anyway. Sub-editors are unaffected: `enableSubEditorPage` is left at its `false` default, so they push onto the app's own `Navigator` rather than the editor's internal one, and stay outside the excluded subtree.

**Bottom-action captions announced twice.** In the clip transform and timeline control rows, the icon button already carries its own semantic label and the text below it repeats the same word, so every action was read out twice. The captions are excluded while that button is there to carry the label. In the timeline row the button is not always there: during a chroma key bake, an audio extraction or a library save it is replaced by `BrandedLoadingIndicator`, which carries no semantics of its own — so the exclusion lifts for the duration and the control keeps a label instead of dropping out of the traversal order entirely.

**Caption color swatches announced nothing.** The `Semantics` wrapper had `button: true` and `selected:` but no `label` and no action, so a swatch was reachable but anonymous. Each now announces its RGB value (and the custom swatch also its picker role), with `onTap` on the `Semantics` node and `excludeFromSemantics` on the `GestureDetector` so the action is exposed once, not twice. The RGB label was a private method inside the color-picker sheet holding a hardcoded English string; it is now an ARB key translated into every locale, reached through `ColorSwatchButton.rgbSemanticLabel`, next to the widget whose `semanticLabel` doc already tells callers they own naming the color.

**Related Issue:** 

Bildschirmfoto 2026-08-08 um 11 34 05

## Out of Scope

- `videoEditorDoneSemanticLabel` stays in the ARB — the audio-editor selection overlay still uses it on a standalone button outside the toolbar. Its `videoEditorCloseSemanticLabel` counterpart lost every call site once the toolbar labels became required, so it is removed from all 22 locales rather than left as a key nothing reads.
- Broader editor a11y (traversal order on the timeline, layer-drag announcements) is untouched.

## Verification

- `flutter test test/widgets/video_editor/ test/screens/video_editor/ test/widgets/video_recorder/ test/screens/profile_setup/ test/widgets/color_swatch_button_test.dart test/l10n/` — 1312 passing
- `dart format --set-exit-if-changed lib test` — clean
- `flutter gen-l10n` re-run; generated files committed
- New tests: `video_editor_caption_custom_style_sheet_test.dart` pins the custom swatch's label, button flag, selected state and tap action; `video_editor_timeline_controls_test.dart` pins that the caption is announced while a control is loading and silent while its button is not
- Existing toolbar/overlay tests updated to resolve labels from `AppLocalizations` instead of hardcoded English
- All 17 non-English locales translated (not left as English fallback)
- Manually verified with TalkBack on a real Samsung Galaxy: each sub-editor toolbar announces its own tool, the canvas no longer traps focus, and swatches read out their color

## Type of Change


dcadenas (approved):
✅ The four a11y problems this fixes are real ones, and the sub-editor labels are the right shape — making them `required` is what stops the next tool inheriting a generic "Done".

I pushed four commits rather than leaving notes.

### Changes I pushed

- The custom swatch label was joined in Dart with a literal `", "`. Moved into an ARB key so each locale lists the way it lists — the same file already ships `、` for ja/zh and `، ` for ar/ur.
- Pinned the audio close button's label to the action it performs. That flag picks both, and only one branch had coverage.
- Recorded what the canvas semantics exclusion depends on, and what would invalidate it.
- Pinned that the timeline strip announces a sticker by its description.

The last two go together. Excluding the canvas is the right call, but two things inside it are labelled: sticker layers on every platform, and the package's own layer buttons on desktop web, where `selectable: auto` resolves to `isDesktop`. Both have an equivalent outside the excluded subtree, so nothing becomes unreachable — but that equivalence was implicit, and one half of it is now held by a test instead of by reasoning.

Worth being explicit about the cost: the canvas itself no longer announces anything. A screen reader learns which stickers are on a clip from the timeline strip, not from the canvas, and while a sub-editor is open the strip is collapsed for everyone.

### Checks

- `dart format --set-exit-if-changed lib test` — clean
- `flutter test` across the video editor, screens, l10n and swatch suites — 998 passing
- `flutter gen-l10n` leaves the tree clean
- All required checks green on the pushed head

The two new tests were verified to fail when the behaviour they guard is reverted.

### Follow-ups

- `'Sticker'` in the timeline overlay bloc is a hardcoded English string that bypasses l10n. Pre-existing, and the tile overrides it for stickers, but it is still the fallback for any other widget layer.
- `BrandedLoadingIndicator` carries no busy or progress semantics anywhere it is used, so a loading control announces its label with no indication it is running.

Neither blocks this branch.

PR #6887, opened 2026-08-08
fix(video-metadata): make the post-details flow readable to screen readers

### LINKED ISSUE #6888: fix(video-metadata): make the post-details flow readable to screen readers
Auto-created from PR #6887.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6887
**Branch:** `fix/metadata-screen-semantics`

hm21:
Closes #6888

## Description

Walking the post-details flow with TalkBack turned it into a wall of unlabelled buttons and stray images. This fixes the semantics across that flow.

**Icon buttons announce their action.** The close/confirm/clear controls in the post-details app bars (capture + classic), the tag picker, the content-warning picker and the user picker had no label at all — TalkBack read them as bare "button". Each now passes `semanticLabel` to `DivineIconButton`, describing what activating it does rather than what it looks like.

**`videoMetadataBackSemanticLabel` → `videoMetadataClosePostDetailsSemanticLabel`.** "Back" describes the chevron, not the outcome. Translated in all 22 locales along with the seven new keys.

**Decorative layers stop competing with the content.** Three subtrees were readable but meaningless to a screen reader, so they're now `ExcludeSemantics`: the 50%-opacity feed preview overlay (a non-interactive mock of the feed chrome, already wrapped in `IgnorePointer`), the clip thumbnail in the capture preview, and the avatar/name/checkmark inside the user search tile — that tile's parent already carries the full label, so the children were duplicating it.

**Content-warning options announce as checkboxes.** `DivineSpriteCheckbox` carries no semantics of its own, so each row read as a plain text label with no indication it was selectable or selected. Each option is now a single `MergeSemantics` node with `checked:` state — "Nudity, checkbox, not checked". The label rides up from the child `Text` instead of being repeated in `label:`, so it cannot drift from the rendered string.

**Post / Save-for-later buttons.** These wrap a `DivineButton` in a `Semantics` that supplies the label and hint, but the inner button contributed its own node, so the meaningful label and the tappable node were separate. They now use `excludeSemantics: true` and hoist `onTap` to the `Semantics` wrapper, producing one node that is both labelled and actionable.

**Cover editor confirm keeps its identity while confirming.** During confirmation the button was replaced by a `BrandedLoadingIndicator`, and the label went with it — the control vanished mid-action. The label now stays on a `Semantics(button: true, enabled: false)` node wrapping the indicator, so it announces as a disabled button instead of disappearing.

**Related Issue:** 

## Out of Scope

Only the post-details/metadata flow and the user picker it opens. Other screens' semantics are untouched.

## Verification

- `flutter test test/l10n/arb_consistency_test.dart test/screens/video_metadata_cover_screen_test.dart test/widgets/user_picker_sheet_test.dart test/widgets/video_metadata/` — 219 passed
- `dart format` clean on all changed files
- Manually verified on a real Samsung Galaxy with TalkBack: walked the post-details screen, both app bar variants, the tag / collaborator / content-warning pickers and the cover editor, confirming every control announces its action and the decorative layers are silent.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -3892,7 +3892,7 @@
     "videoEditorCloseSemanticLabel": "ገጠመ",
     "videoEditorDoneSemanticLabel": "ተከናውኗል",
     "videoEditorLevelSemanticLabel": "ደረጃ",
-    "videoMetadataBackSemanticLabel": "ተመለስ",
+    "videoMetadataClosePostDetailsSemanticLabel": "የልጥፍ ዝርዝሮችን ዝጋ",
     "videoMetadataDismissHelpDialogSemanticLabel": "የእገዛ ንግግርን አሰናብት",
     "videoMetadataGotItButton": "ገባኝ!",
     "videoMetadataLimitReachedWarning": "64KB ገደብ ላይ ደርሷል። ለመቀጠል አንዳንድ ይዘቶችን ያስወግዱ።",
@@ -4184,8 +4184,8 @@
         "description": "Screen-reader announcement spoken after the user confirms a new cover thumbnail and the screen is about to close. Visual users get the screen pop as feedback; this is the audio equivalent."
     },
     "videoMetadataEditCoverTitle": "Edit cover",
-    "videoMetadataEditCoverCloseSemanticLabel": "Close cover editor",
-    "videoMetadataEditCoverConfirmSemanticLabel": "Confirm cover selection",
+    "videoMetadataEditCoverCloseSemanticLabel": "የሽፋን ለውጦችን አስወግድ",
+    "videoMetadataEditCoverConfirmSemanticLabel": "የተመረጠውን ፍሬም እንደ ቪዲዮ ሽፋን ተጠቀም",
     "videoMetadataEditCoverStripSemanticLabel": "Seek through video to select cover frame",
     "videoMetadataTagsPickerSearchHint": "መለያዎችን ይፈልጉ ወይም ያክሉ",
     "videoMetadataTagsPickerEmptyHint": "ሰዎች ቪዲዮዎን እንዲያገኙ መለያዎችን ያክሉ",
@@ -5057,5 +5057,12 @@
       }
     }
   },
-  "relaySettingsRemoveRelayTooltip": "ቅብብሎሽ አስወግድ"
+  "relaySettingsRemoveRelayTooltip": "ቅብብሎሽ አስወግድ",
+  "videoMetadataTagsPickerCancelSemanticLabel": "የመለያ ምርጫን ሰርዝ",
+  "videoMetadataTagsPickerConfirmSemanticLabel": "የተመረጡ መለያዎችን ተግብር",
+  "userPickerCancelSemanticLabel": "የተጠቃሚ ምርጫን ሰርዝ",
+  "userPickerConfirmSemanticLabel": "የተመረጡ ተጠቃሚዎችን አረጋግጥ",
+  "userPickerClearSelectionSemanticLabel": "የተጠቃሚ ምርጫን አጽዳ",
+  "videoMetadataContentWarningsPickerCancelSemanticLabel": "የይዘት ማስጠንቀቂያ ምርጫን ሰርዝ",
+  "videoMetadataContentWarningsPickerConfirmSemanticLabel": "የተመረጡ የይዘት ማስጠንቀቂያዎችን ተግብር"
 }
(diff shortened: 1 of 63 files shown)


NotThatKindOfDrLiz:
@hm21 I pushed a takeover follow-up in 7e5d2e27f to close the remaining accessibility gaps from review.

What changed:
- Classic post-details bottom bar now uses the publish semantics label, excludes the inner button semantics, and hoists onTap onto the wrapper so the announced node is also actionable.
- Added a classic bottom-bar screen-reader activation regression test.
- Added a cover-editor confirming-state semantics regression test and hardened the existing missing-thumbnail failure test against stray local files.

Verification:
- flutter test test/widgets/video_metadata/modes/classic/video_metadata_classic_bottom_bar_test.dart test/screens/video_metadata_cover_screen_test.dart
- Pre-push hook: analyzer plus changed-file test set
- GitHub checks are green, including Analyze, Format, Generated Files, all 4 test shards, preview build, semantic PR, mergeability, CLA.

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up. The remaining classic bottom-bar semantics defect is fixed, the confirming-state cover semantics now has regression coverage, local/pre-push verification passed, and GitHub checks are green.

NotThatKindOfDrLiz (changes requested):
## Review Summary

Accessibility improvements to post-details semantic flow look solid overall, but two blockers and three non-blocking items need attention before merge.

### Blockers (must fix before merge)

**1. Classic mode post button has identical "announced but inert" defect that capture mode fixed** (#6887 fixed this in capture, but not classic)

`mobile/lib/widgets/video_metadata/modes/classic/video_metadata_classic_bottom_bar.dart` line 58

The post button wraps `DivineButton` inside `Semantics(label:...)` without `excludeSemantics: true`. This creates two announcement nodes for one action: a label node + an inert button node. 

**Fix:** Apply the same pattern used in capture mode (line 83-93 of capture_bottom_bar.dart):
- Hoist the `onPressed` callback outside the button
- Wrap in `Semantics(label:..., onTap:..., button: true, enabled: true)`
- Pass `enabled: false` to `DivineButton` so it doesn't announce itself

**2. New confirming-state Semantics node in cover editor has no test coverage**

`mobile/lib/screens/video_metadata/video_metadata_cover_screen.dart` line 661

The new `if (isConfirming)` branch wraps `BrandedLoadingIndicator` in `Semantics(button: true, enabled: false)` — structure is correct, but it's untested. 

**Fix:** Add a test case to `mobile/test/screens/video_metadata/video_metadata_cover_screen_test.dart` asserting that:
- When `state.status == CoverEditorStatus.confirming`, the loading indicator is rendered
- Semantics verify the indicator is announced as a disabled button (prevents interaction)

### Non-Blocking (for author consideration)

**3. Caption text is over-excluded from semantics**

`mobile/lib/screens/video_metadata/video_metadata_preview_screen.dart` line 396

The new `if (showCaption) _PreviewOverlay(...)` wraps the entire overlay in `ExcludeSemantics`, which silences both the decorative `FeedModeSwitch` AND the previewed caption text itself. 

Was this intentional? If the caption should remain readable to screen readers (common for preview content), consider extracting just the `FeedModeSwitch` into its own `ExcludeSemantics` wrapper instead.

**4. Semantic labels built via string concatenation instead of ARB keys**

`mobile/lib/widgets/user_picker_sheet.dart` line 545-556

Labels are assembled at runtime: `label: nip05 != null && nip05.isNotEmpty ? '$action. $nip05' : action`

This works, but loses l10n flexibility (word order, punctuation) for RTL and other locales. Consider adding a parameterized ARB key like `userPickerActionWithNip05` with placeholder `{nip05}` so translators can adjust order/punctuation per-locale.

**5. New ARB keys lack `@description` metadata**

Files: `mobile/lib/l10n/app_en.arb` and all 22 locale files

Seven new keys added without `@keyName` metadata blocks describing their context/purpose for translators (e.g., `videoMetadataTagsPickerCancelSemanticLabel`, `userPickerCancelSemanticLabel`, etc.). 

While the translations are present in all locales, adding descriptions helps translators maintain context on future updates. Example:

```json
"videoMetadataTagsPickerCancelSemanticLabel": "Cancel tags picker",
"@videoMetadataTagsPickerCancelSemanticLabel": {
  "description": "Screen reader label for cancel button in the tags picker modal."
}
```

---

**Status:** Green on all CI checks (Analyze, Format, Generated Files, all test shards). All 22 ARB locales properly updated with genuine translations. No tech debt or test regressions introduced.

**Next:** Fix blockers 1 & 2, then we can merge. Raise issues for 3/4/5 if you'd like to follow up after launch.

PR #6889, opened 2026-08-08
fix(feed): clear a stranded page overlay when the shell is uncovered

### LINKED ISSUE #6239: fix(feed): video does not autoplay when scrolling to the next one
### What would you like?

Divine User, Jul 21, 2026, 11:22

play when scroll

When you go to the next video you have to hit play, wich is annoying to do. I might just be stupid though and have some setting turned on or smth

makes watching content easier and more enjoyable

### When would you use this?
always asap

**User Pubkey:** a93513d4f8124e46fe2bdc129d60d823aad453944670b59d5131eebf6b120f29

makes watching content easier and more enjoyable

### When would you use this?

always asap

### Anything else?

(Generated from Zendesk Ticket: rabblelabs.zendesk.com/agent/tickets/4112)

realmeylisdev:
## Description

The home feed only autoplays while `overlayVisibilityProvider.isPageOpen` is false (`video_feed_page.dart:294`). `pushWithVideoPause` raises that flag before pushing and clears it from the `push` future's `whenComplete` — but go_router completes an `ImperativeRouteMatch` only in `_completeRouteMatch`, which runs on the pop path alone. A `go()` that rebuilds the match list *removes* the route and drops the pending completer, so the clear never runs and the flag stays `true`.

`BackButtonHandler` navigates with `router.go(...)` rather than popping for several route types (hashtag unconditionally; the `videoIndex` and tab-history branches too), and deep links do the same whenever the current location is already in the target route family (`DeepLinkNavAction.go`), as do refresh redirects. So: open a hashtag, sound, or author profile from the feed, leave it via a go-style navigation, and every later scroll lands on a paused video until another page overlay happens to clear the flag.

A page overlay is always backed by a route above the shell, so an uncovered shell proves none is open. `AppShell` already resets its sibling `shellObscuredProvider` from `didPush`/`didPopNext`, and those callbacks fire for the shell-uncovered `go()` removal case. Clearing the page flag from that same signal closes the gap.

**Related Issue:** Fixes #6239

## Out of Scope

- **Not changing where a `go()`-style back lands.** Making `BackButtonHandler` prefer `pop()` when it can would move the hashtag back target off Explore — a product-behavior call, not a bug fix. The safety net here holds regardless of which navigation removed the route.
- **Page-overlay ownership/refcounting.** `setPageOpen` remains a boolean in this PR; the broader nested-overlay sharp edge is tracked in #6898.
- **Separate autoplay gate gaps.** Externally-driven autoplay gate changes are tracked in #6899, and feed page-change playback coverage is tracked in #6900.

## Verification

- `flutter analyze lib/providers/shell_obscured_provider.dart test/router/app_shell_obscured_test.dart test/utils/pause_aware_modals_test.dart` — clean.
- `flutter test test/router/app_shell_obscured_test.dart test/utils/pause_aware_modals_test.dart` — 10 passed.
- Pre-push hook reran the changed-file analyzer and the two affected test files — passed.
- Regression coverage now drives the real path: `AppShell` under `MaterialApp.router`, `pushWithVideoPause`, then `router.go(...)`, asserting the stranded page flag clears when the shell is uncovered.

## Type of Change

### DIFF
--- a/mobile/lib/providers/shell_obscured_provider.dart
+++ b/mobile/lib/providers/shell_obscured_provider.dart
@@ -18,6 +18,12 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 /// without a pop event (e.g. sign-out → /welcome → back to home) cannot keep
 /// the feed paused.
 ///
+/// AppShell also clears `overlayVisibilityProvider.isPageOpen` when this flag
+/// becomes `false`. A page overlay is backed by a route above the shell, so an
+/// uncovered shell proves the page overlay should no longer block autoplay,
+/// even if go_router removed the route with `go()` and never completed the
+/// original `pushWithVideoPause` future.
+///
 /// The home feed reads this to know it is offstage behind a pushed screen.
 /// GoRouter's `routeInformationProvider` collapses to the shell location
 /// (`/home`) while popping between pushed routes, so the feed cannot rely on
--- a/mobile/lib/router/app_shell.dart
+++ b/mobile/lib/router/app_shell.dart
@@ -11,6 +11,7 @@ import 'package:openvine/app_update/app_update.dart';
 import 'package:openvine/l10n/l10n.dart';
 import 'package:openvine/notifications/view/notifications_page.dart';
 import 'package:openvine/providers/app_providers.dart';
+import 'package:openvine/providers/overlay_visibility_provider.dart';
 import 'package:openvine/providers/relay_list_repository_provider.dart';
 import 'package:openvine/providers/route_feed_providers.dart';
 import 'package:openvine/providers/shell_obscured_provider.dart';
@@ -108,6 +109,14 @@ class _AppShellState extends ConsumerState<AppShell> with RouteAware {
     WidgetsBinding.instance.addPostFrameCallback((_) {
       if (!mounted) return;
       ref.read(shellObscuredProvider.notifier).setObscured(obscured: obscured);
+      if (!obscured) {
+        // Every page overlay is backed by a route above the shell, so an
+        // uncovered shell proves none is open. Not redundant with
+        // `pushWithVideoPause`'s own clear, which a `go()`-style back skips
+        // entirely — see its doc comment. Without this the home feed could stay
+        // permanently unable to autoplay on scroll (#6239).
+        ref.read(overlayVisibilityProvider.notifier).setPageOpen(false);
+      }
     });
   }
 
--- a/mobile/lib/utils/pause_aware_modals.dart
+++ b/mobile/lib/utils/pause_aware_modals.dart
@@ -32,6 +32,15 @@ extension PauseAwareModals on BuildContext {
   /// Calls [OverlayVisibility.setPageOpen(true)] before pushing and
   /// [setPageOpen(false)] when the pushed route is popped.
   /// This releases all video players to free memory.
+  ///
+  /// The returned future only completes on a **pop**: go_router completes an
+  /// `ImperativeRouteMatch` from `_completeRouteMatch`, which runs on the pop
+  /// path alone, so a `go()` that rebuilds the match list removes the route and
+  /// drops the pending completer. `AppShell` therefore also clears the flag
+  /// whenever the shell is uncovered — without that safety net a `go()`-style
+  /// back (the Android back handler, a deep link, a refresh redirect) strands
+  /// the flag `true` and the home feed never autoplays again (#6239). Do not
+  /// make this future the only clear path.
   Future<T?> pushWithVideoPause<T extends Object?>(
     String location, {
     Object? extra,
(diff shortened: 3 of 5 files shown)


realmeylisdev:
Verified on device (iPhone 17 Pro Max simulator, iOS 26.5, production relays) — details and log excerpts on #6239.

Without this change, `pushWithVideoPause` → a `go()`-style navigation that removes the route → return to the feed produced `Init player index 0` with no `Playing index 0`: frozen frame, no play button, byte-identical consecutive screenshots. With it, the same sequence logs `Page closed` on return, then `Init player index 0` → `Playing index 0`, and swiping autoplays.

Worth knowing while reviewing: the trigger is not Android-only as the description implies. Deep links reach it on any platform — a hashtag opened from the About sheet uses `pushWithVideoPause`, and any later hashtag link resolves to `DeepLinkNavAction.go` because the location is already in the `/hashtag/` family. There is also a one-tap in-app path: `OtherProfileScreenRouter` `go`es to `/profile/<npub>` when the target is the signed-in user, so tapping your own avatar on your own feed video removes the route it was just pushed into.

Also: because `setPageOpen` is a plain bool with no refcount, the stranded flag is cleared by the next page overlay that closes normally — so the pre-fix impact is bounded by that rather than lasting until restart. I have not changed the refcounting here; flagging it as a known sharp edge rather than something this PR addresses.

Separately, #6892 fixes a second defect this surfaced: the first tap on a stopped-but-not-`paused` video was a no-op.

hm21:
@realmeylisdev I tried to reproduce the issue to check if it's fixed, but I wasn't able to. Could you briefly explain which steps I need to follow to reproduce it?

I tried the following, but everything worked fine:

1. Open Home Feed
2. Tap on the user avatar
3. Tap the hardware back button
4. Swipe to the next video

I also tried:

1. Open Home Feed
2. Open About
3. Tap a hashtag
4. Tap the hardware back button
5. Swipe to the next video

NotThatKindOfDrLiz:
@realmeylisdev I took this over to close out the review gaps.

Changes pushed on top of your fix:
- replaced the pop-based AppShell regression with a real `MaterialApp.router`/`GoRouter` path: `pushWithVideoPause('/hashtag')` followed by `router.go('/explore')`, asserting the stranded page flag clears when the shell is uncovered
- documented the AppShell -> `overlayVisibilityProvider.isPageOpen` safety-net contract in `shell_obscured_provider.dart`
- filed durable follow-up issues for the adjacent debt: #6898, #6899, #6900
- updated the PR body verification/out-of-scope notes to point to those issues

Verification:
- `flutter analyze lib/providers/shell_obscured_provider.dart test/router/app_shell_obscured_test.dart test/utils/pause_aware_modals_test.dart`
- `flutter test test/router/app_shell_obscured_test.dart test/utils/pause_aware_modals_test.dart`
- pre-push hook reran changed-file analyzer/tests
- GitHub Mobile CI is green on `3041dd3412b050bd5cab153ed12e9d3351490713`

@hm21 your repro steps used hardware back, which only exercises this route on Android because `BackButtonHandler.initialize` is Android-gated. The cross-platform repro is: open a hashtag from a feed video through the About/metadata sheet, then navigate to another hashtag while already on a hashtag route so the deep-link path uses `DeepLinkNavAction.go`; pre-fix, returning to the feed leaves autoplay blocked. The new regression test covers that same `pushWithVideoPause` + `go()` removal shape.

NotThatKindOfDrLiz (approved):
Approved after takeover.

I pushed one follow-up commit that makes the regression test cover the actual failure shape (`pushWithVideoPause` followed by `router.go(...)` under `AppShell`) and documents the overlay-clear safety net. The adjacent debt is now tracked in #6898, #6899, and #6900.

Local verification and the pre-push hook passed, and GitHub Mobile CI is green on the final head.

NotThatKindOfDrLiz:
**Tech debt tracking:** This PR identified three deferred issues that should be tracked separately per the no-tech-debt policy:

- #7540: Add reference counting to overlayVisibilityProvider.isPageOpen
- #7541: Implement gate-change detection in _syncCurrentAutoPlayGate  
- #7542: Add regression test for autoplay gate-change detection

These items were previously left as comments on #6239 but are now properly tracked.

NotThatKindOfDrLiz:
**Test coverage note:** The regression test in `app_shell_obscured_test.dart:201-243` exercises the safety-net clear via `navigator.pop()`, but the actual trigger for the fix—`go()` navigation dropping a completer from `pushWithVideoPause`—is never exercised end-to-end under the real go_router architecture. 

Consider adding an end-to-end test that:
1. Pumps AppShell with MaterialApp.router
2. Opens a root route via `pushWithVideoPause` (sets `overlayVisibilityProvider.isPageOpen = true`)
3. Uses `router.go()` to navigate away (drops the completer without calling pop)
4. Asserts the flag clears when shell is uncovered

This would validate the fix against the actual code path and provide confidence against future go_router/Flutter behavior changes.

PR #6890, opened 2026-08-08
refactor(video-editor): split the canvas into focused files

### LINKED ISSUE #6891: refactor(video-editor): split the canvas into focused files
Auto-created from PR #6890.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6890
**Branch:** `refactor/video-editor-canvas-scrim`

hm21:
Closes #6891

`video_editor_canvas.dart` had accumulated several self-contained pieces that could not be tested without pumping the whole editor. Each moves to its own file. No behaviour change — the render-space mapping, the cover-fit into the target rect, the hit-test forwarding and the cut-area scrim all work exactly as before.

| New file | What moved |
| --- | --- |
| `utils/video_editor_playhead.dart` | Playhead/timeline math |
| `widgets/.../video_editor_clip_preview.dart` | Preview surface, incl. the stop-motion branch and its position subscription |
| `widgets/.../video_editor_setup_loading_indicator.dart` | Thumbnail placeholder shown while the player sets up |
| `widgets/.../video_editor_cut_area_overlay.dart` | Letterbox scrim: `VideoEditorCutAreaOverlay` (was `_OverlayCutArea`), `ScrimBars`, `scrimZoomTransform` |
| `widgets/.../hit_test_expander.dart` | Letterbox hit-test forwarding |
| `models/video_editor/clip_history_direction.dart` | Undo/redo direction enum |
| `models/video_editor/clip_snapshot_sync_op.dart` | Clip snapshot sync op enum |

That takes the canvas from 3343 to 2913 lines.

## Tests

The split is motivated by testability, so the extracted pieces get direct tests instead of only being reachable through the full editor:

- `test/utils/video_editor_playhead_test.dart` — speed-scaled interpolation, clamping, stop-motion loop wrapping.
- `test/widgets/.../video_editor_clip_preview_test.dart` — live frames from `ClipEditorBloc` beating the clip-manager copy, the fallback when the clip is absent there, position forwarding, and that a normal clip does *not* rebuild on a playhead tick.
- `test/widgets/.../video_editor_setup_loading_indicator_test.dart` — contain-fit onto the render height and the canvas radius scaled off the visible width.

`scrim_zoom_transform_test.dart` is renamed to `video_editor_cut_area_overlay_test.dart` so it mirrors its `lib/` path; its contents are unchanged.

One trap worth naming, since it bit this PR: the blocs in the clip-preview test are built inside the test body, not in `setUp`. A `Bloc` created outside `testWidgets`' fake-async zone schedules its stream delivery in the real zone, so `emit` updates `bloc.state` but `tester.pump()` never flushes the notification and no listener rebuilds. The test passes while asserting nothing.

## Test plan

- [x] `flutter test test/widgets/video_editor test/screens/video_editor test/utils/video_editor_playhead_test.dart` — 903 passed
- [x] `dart format` — clean
- [x] Manual verification on a real Samsung Galaxy: opened the editor with a square clip on a 9:16 target and vice versa, confirmed the frame is positioned and scaled as before, dragged a sticker outside the crop rect and confirmed it still dims, and confirmed pinch/pan still starts from the letterbox area


realmeylisdev (approved):
lgtm. went through it as a pure move and it holds up: every extracted block is identical to what was in the canvas apart from `super.key` and the `@visibleForTesting` drops, and nothing outside the canvas referenced any of it.

ran the 903 and poked the new tests with 7 mutations (live-frames fallback, playhead forwarding, the normal-clip subscription scope, contain-fit, radius, loop wrap, speed) and they all went red, so they pin what they claim to. opened the editor on a sim too, canvas and letterbox render the same.

retitled #6891, it still said "drop the cut-area scrim".

PR #6892, opened 2026-08-08
fix(feed): make the first tap on a stopped video play it

realmeylisdev:
## Description

`_handlePlayerTap` toggled on `controller.state.isPaused`, which is strictly `status == paused`. Stopped player statuses such as `ready`, `idle`, `completed`, and `error` are not running, but they are also not `paused`, so they fell to the `pause()` branch instead of requesting playback.

The toggle now keys on whether the player is running. `playing` and `buffering` pause; every stopped status requests play. `buffering` still counts as running, so a tap mid-load stops playback instead of issuing a redundant play. Behavior in the ordinary `playing` <-> `paused` cycle is unchanged.

This is a correctness fix for the tap resolver and covers stopped-but-not-paused states across player backends. On iOS and Android, native `ready` is reported to Dart as `paused`, so this PR should not be treated as proof that the mobile #6239 reproduction was caused by a Dart-visible `ready` status. If that mobile symptom is still reproducible, #6239 should keep tracking the device-specific root cause.

Traced through both native backends to confirm that: iOS serializes `ready` with `rate == 0` as `paused` (`DivineVideoPlayerInstance.swift`), and Android maps `STATE_READY` to `playing`/`paused` depending on `playWhenReady` (`DivineVideoPlayerInstance.kt`). The stopped-but-not-paused statuses that *are* Dart-visible on mobile are `idle`, `completed`, and `error` — all three previously fell to the `pause()` branch, and `idle` is the first case the resolver test turns red on when the old `isPaused` keying is restored.

Extracted as a pure resolver alongside the existing `resolveDoubleTapLikeAction`: the tap target is gated on `isFirstFrameRendered`, so exercising it needs a live native controller and cannot be driven from a widget test. The resolver is unit-testable and now pins the expected action for every `PlaybackStatus` value.

**Related Issue:** Refs #6239

## Out of Scope

- The paused-overlay behavior is tracked separately in #6784/#6889. This PR only changes the tap action resolver, not the paused affordance.
- Restarting native playback from a genuinely completed non-looping clip may require a seek before `play()`. This PR only ensures the tap resolver requests playback instead of pause for completed status.

## Verification

- `flutter analyze lib/widgets/video_feed_item/player_tap_helpers.dart test/widgets/video_feed_item/player_tap_helpers_test.dart` - clean.
- `flutter test test/widgets/video_feed_item/player_tap_helpers_test.dart` - passed.
- Pre-push hook - passed:
  - no merge conflicts with `main`
  - analyzer clean for changed files
  - `flutter test test/widgets/video_feed_item/feed_videos_test.dart test/widgets/video_feed_item/player_tap_helpers_test.dart` - 40 passed.

## Type of Change

### DIFF
--- a/mobile/lib/widgets/video_feed_item/feed_videos.dart
+++ b/mobile/lib/widgets/video_feed_item/feed_videos.dart
@@ -44,6 +44,7 @@ import 'package:openvine/widgets/video_feed_item/double_tap_like_helpers.dart';
 import 'package:openvine/widgets/video_feed_item/feed_immersive_chrome.dart';
 import 'package:openvine/widgets/video_feed_item/moderated_content_overlay.dart';
 import 'package:openvine/widgets/video_feed_item/paused_video_overlay.dart';
+import 'package:openvine/widgets/video_feed_item/player_tap_helpers.dart';
 import 'package:openvine/widgets/video_feed_item/subtitle_overlay.dart';
 import 'package:openvine/widgets/video_feed_item/verifying_aware_video_error_overlay.dart';
 import 'package:openvine/widgets/video_feed_item/video_feed_item.dart';
@@ -738,12 +739,13 @@ class __OverlayState extends ConsumerState<_Overlay> {
   void _handlePlayerTap() {
     widget.onSuppressAutoAdvance?.call();
 
-    if (widget.controller != null) {
-      if (widget.controller!.state.isPaused) {
-        widget.controller!.play();
-      } else {
-        widget.controller!.pause();
-      }
+    final controller = widget.controller;
+    if (controller == null) return;
+    switch (resolvePlayerTapAction(controller.state.status)) {
+      case PlayerTapAction.play:
+        controller.play();
+      case PlayerTapAction.pause:
+        controller.pause();
     }
   }
 
--- a/mobile/lib/widgets/video_feed_item/player_tap_helpers.dart
+++ b/mobile/lib/widgets/video_feed_item/player_tap_helpers.dart
@@ -1,0 +1,32 @@
+// ABOUTME: Pure decision helper for the tap-to-toggle-playback gesture on feed
+// ABOUTME: videos. Keeps the "which states count as running" rule testable in
+// ABOUTME: isolation from the FeedVideos overlay widget and the native player.
+
+import 'package:divine_video_player/divine_video_player.dart';
+
+/// The action a tap on a feed video should trigger.
+enum PlayerTapAction {
+  /// Start (or restart) playback.
+  play,
+
+  /// Stop playback.
+  pause,
+}
+
+/// Decides how a tap on a feed video should behave, given the player's
+/// [status].
+///
+/// Keyed on whether the player is *running*, not on whether it is paused.
+/// [PlaybackStatus.isPaused] is strictly `status == paused`, so a controller
+/// that initialized but never started ([PlaybackStatus.ready]) and a finished
+/// non-looping clip ([PlaybackStatus.completed]) are both stopped without being
+/// `paused`. Keying on `isPaused` sent those to `pause()`, so the viewer's first
+/// tap did nothing and they had to tap twice (#6239).
+///
+/// [PlaybackStatus.buffering] counts as running, so a tap mid-load still
+/// pauses rather than issuing a redundant play.
+PlayerTapAction resolvePlayerTapAction(PlaybackStatus status) {
+  return status.isPlaying || status.isBuffering
+      ? PlayerTapAction.pause
+      : PlayerTapAction.play;
+}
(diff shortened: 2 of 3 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I took this over to address the review accuracy issues and keep the PR merge-ready.

What changed:
- Rebased the branch onto current `origin/main`.
- Reworded the original commit message to remove the unsupported iPhone `ready` status claim while preserving your authorship.
- Updated the helper docs and test comments so they no longer claim native completed playback restarts without a seek.
- Replaced the non-nullability/exhaustiveness assertion with a per-`PlaybackStatus` expected-action table.
- Updated the PR body to use `Refs #6239`, document the mobile status nuance, and keep the completed-clip seek behavior explicitly out of scope.

Verification:
- Pre-push hook passed: no merge conflicts with `main`, analyzer clean, and `flutter test test/widgets/video_feed_item/feed_videos_test.dart test/widgets/video_feed_item/player_tap_helpers_test.dart` passed.
- GitHub checks are green, including Mobile CI, Analyze, Format, Generated Files, all test shards, preview build, semantic PR, issue link, and mergeability.

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup.

I corrected the evidence/narrative issues without changing the resolver behavior: the PR no longer claims the iPhone repro was caused by a Dart-visible native `ready` status, no longer promises native completed-clip restart semantics, and the resolver test now pins every `PlaybackStatus` to an expected action.

All local pre-push checks passed and all GitHub checks are green. @realmeylisdev please review the takeover changes and merge when ready.

NotThatKindOfDrLiz:
## Review: PR #6892 — Make First Tap on Stopped Video Play It

**Status:** ✅ CI green (16/16). Meets safety bar, but narrative accuracy and test coverage need refinement.

---

### Key Findings

#### 1. Root Cause Narrative Requires Verification

The PR states that on iOS/macOS, `PlaybackStatus.ready` is emitted while paused, causing the tap handler to call `pause()` instead of `play()`. However, native status mapping logic shows this is architecturally impossible:

- **darwin** (`DivineVideoPlayerInstance.swift`, lines 1202–1213): Exhaustive mapping ensures `ready` is never emitted to Dart. When `currentStatus == "ready" && player.rate == 0`, we emit `"paused"` instead.
- **android** (`DivineVideoPlayerInstance.kt`, lines 1187–1195): STATE_READY maps to either `"playing"` or `"paused"` based on `playWhenReady`. Never `"ready"` alone.

**Action required:** Re-run the repro from the original issue and capture the exact `state.status` value when the tap fails. If it's not `ready`, update the narrative. This is critical for understanding whether the fix is a behavioral correction or a workaround for unexpected state.

**Note:** The web platform *does* emit `idle` while paused in certain conditions (readyState < 3 while paused), so this is likely the platform where the fix has material impact, not iOS.

---

#### 2. `completed` Status Behavior Requires Clarification

The helper maps `completed → play`. However, on the feed:

- Videos loop (`playerConfiguration.loopEnabled = true`), so playback never reaches `completed` (the player restarts automatically).
- The `playerDidFinish` guard prevents emitting a `completed` state to Dart.

**Consequence:** The `completed → play` case is unreachable on the feed and does not "restart" videos; it is dead code.

**Action required:** Choose one:
- *Option A (recommended):* Remove the `completed` case and update the test to not expect it.
- *Option B:* If `completed` should restart without seeking back to the start, add `seekTo(Duration.zero)` before `play()` and document this behavior.

Current test comment ("completed video restarts") is misleading without the seek.

---

#### 3. Exhaustiveness Test Cannot Fail

`player_tap_helpers_test.dart` includes:

```dart
for (final status in PlaybackStatus.values) {
  final action = resolvePlayerTapAction(status);
  expect(action, isNotNull);
}
```

This test **always passes** because `resolvePlayerTapAction` handles every `PlaybackStatus` value and never returns `null`. The test provides no protection against forgetting a case.

**Action required:** Replace with a per-status truth table that pins each value's expected action:

```dart
test('all statuses map to expected action', () {
  expect(resolvePlayerTapAction(PlaybackStatus.idle), PlayerTapAction.play);
  expect(resolvePlayerTapAction(PlaybackStatus.ready), PlayerTapAction.play);
  expect(resolvePlayerTapAction(PlaybackStatus.playing), PlayerTapAction.pause);
  expect(resolvePlayerTapAction(PlaybackStatus.paused), PlayerTapAction.play);
  expect(resolvePlayerTapAction(PlaybackStatus.buffering), PlayerTapAction.pause);
  expect(resolvePlayerTapAction(PlaybackStatus.completed), PlayerTapAction.play);
  expect(resolvePlayerTapAction(PlaybackStatus.error), PlayerTapAction.play);
});
```

This can catch unexpected changes to the logic.

---

### What's Good

✅ Logic is sound: `isPlaying || isBuffering → pause; else → play`.  
✅ Tap handler correctly delegates to the helper.  
✅ Overlay polarity is intentional (different from tap logic per #6784/#6889).  
✅ No new technical debt introduced.  
✅ All CI checks green.

---

### Recommendation

Address the three findings above and push a follow-up commit. The change is safe to merge once narrative and test coverage align.

PR #6893, opened 2026-08-08
fix(video-metadata): stop painting the clip cover's edits twice

### LINKED ISSUE #6894: fix(video-metadata): stop painting the clip cover's edits twice
Auto-created from PR #6893.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6893
**Branch:** `fix/metadata-preview`

hm21:
Closes #6894

## Description

`VideoMetadataCapturePreviewThumbnail` stacked an `_EditedPreviewOverlay` on top of the clip cover, which re-applied the editor's filter states, tune adjustments and captured layers via `ColorFilterGenerator` plus positioned `Image.memory` layers.

Every call site already shows a cover that has all of that baked in:

- `VideoMetadataCaptureClipPreview` uses `state.finalRenderedClip ?? clips.first`, so after the render it is the rendered cover.
- `VideoMetadataPreviewScreen` from `VideoMetadataCaptureAppBar` gates the tap on `finalRenderedClip != null`. That clip's cover is extracted from the **rendered** file (`VideoEditorRenderService`, with an explicit comment on why it must not be inherited from the source clip).
- `VideoMetadataPreviewScreen` from `UploadFailureSheet` (`previewOnly: true`) shows a draft whose render already finished. Here the overlay was not just redundant but wrong: it painted whatever `videoEditorProvider` happened to hold at that moment onto a cover from a different, already-baked session.

In every case the overlay painted the effects a second time. The only window where it added anything was the raw thumbnail shown while the render is still running — and there it sits under `VideoEditorProcessingOverlay`'s 70% black scrim and spinner, and is swapped for the rendered cover as soon as the render lands. That approximation is what this change gives up; everything else it did was a double-apply.

With the overlay gone the widget no longer reads `videoEditorProvider`, so it drops from `ConsumerWidget` back to `StatelessWidget`, and the now single-child `Stack(fit: .expand)` goes with it. Both layout hosts already hand it tight constraints — the `AnimatedSwitcher` layout builder's expanded stack in the clip preview, and `DivineVideoPlayer`, which returns the placeholder straight into the `SizedBox(width: ratio * 100, height: 100)` while its controller is null and otherwise wraps it in its own expanded stack — so the cover-fitted image fills the same box on its own.

Net effect: 187 lines down to 27, and five imports removed.

**Related Issue:** 

## Out of Scope

Restoring an edited-look preview for the pre-render window. If that turns out to be missed, it belongs on the processing overlay rather than under it.

## Verification

- `dart format` — clean
- `flutter analyze lib/widgets/video_metadata lib/screens/video_metadata` — no issues
- `flutter test test/widgets/video_metadata` — 162 passing (no test referenced the removed widget or overlay)
- Manual pass on a real Samsung Galaxy still pending — record a clip, apply a filter and a sticker, watch the metadata screen through the render, then open the full-screen preview and the cover picker. Undrafting once that is done.

CI shard 2 is red on a failure this branch does not cause: `main` at this PR's base (95a05362f) is itself red there on `ClipThumbnailImage renders a custom placeholder for a missing file` and `PopularVideosTab reports a slow variant switch once`. This branch failed `ClipSpeedRenderService caps concurrent native renders` on the first run and `ClipThumbnailImage renders the neutral placeholder for a missing file` on the re-run — a different test each time, and the second is the same file that fails on `main`, dying on leaked `ImageStreamCompleter` exceptions from an earlier suite in the merged isolate. Seed-dependent shard pollution, tracked separately rather than patched in here.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/widgets/video_metadata/modes/capture/video_metadata_capture_preview_thumbnail.dart
+++ b/mobile/lib/widgets/video_metadata/modes/capture/video_metadata_capture_preview_thumbnail.dart
@@ -1,25 +1,15 @@
-import 'dart:typed_data';
-
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:openvine/models/divine_video_clip.dart';
-import 'package:openvine/providers/video_editor_provider.dart';
 import 'package:openvine/widgets/video_clip/clip_thumbnail_image.dart';
-import 'package:pro_image_editor/features/filter_editor/widgets/filter_generator.dart';
-import 'package:pro_image_editor/pro_image_editor.dart';
 
-class VideoMetadataCapturePreviewThumbnail extends ConsumerWidget {
+class VideoMetadataCapturePreviewThumbnail extends StatelessWidget {
   const VideoMetadataCapturePreviewThumbnail({required this.clip, super.key});
 
   final DivineVideoClip clip;
 
   @override
-  Widget build(BuildContext context, WidgetRef ref) {
-    final editingParameters = ref.watch(
-      videoEditorProvider.select((s) => s.editorEditingParameters),
-    );
-
+  Widget build(BuildContext context) {
     if (clip.thumbnailPath == null) {
       return Center(
         child: DivineIcon(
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz (approved):
Code review passes for me. The overlay removal is correct for the rendered-cover path, and the refreshed CI is green.

Before merging, please do the real Samsung Galaxy pass from the PR body: record a clip, apply a filter and a sticker, watch the metadata screen through render, then open the full-screen preview and cover picker. Also please update the PR body afterward so it no longer says the manual pass is pending / undrafting is pending.

One process note: the update came in as a merge-from-main commit. Repo workflow asks for rebasing onto main instead of merging main into feature branches, so please clean that up before merge if the branch history is going to be preserved. If this is squash-merged, the code diff itself is fine.

NotThatKindOfDrLiz:
## Independent Review: PR #6893

### Summary
This PR successfully removes a double-application of video editor effects (filters, tune adjustments, stickers) that were being unnecessarily re-painted on top of covers already containing the effects from the render service. The core fix is sound and safe to merge after addressing a few items below.

### Verification Complete ✅

**Three key invariants confirmed:**

1. **Render service extracts covers from rendered output** (video_editor_render_service.dart:507-526)
   - Comment explicitly states: *"The cover must come from the rendered file, not from the first source clip"*
   - Fallback to source thumbnail only occurs when extraction fails

2. **Preview screen only opens when finalRenderedClip is available** (video_metadata_preview_screen.dart)
   - No path to display unrendered clip in the preview UI
   - Thumbnail painting happens inside strict constraints with animated transitions

3. **Any editor change invalidates rendered clip** (video_editor_provider.dart:670-672)
   - `updateEditorEditingParameters()` calls `invalidateFinalRenderedClip()` immediately
   - Prevents stale renderings from being displayed during active editing

### Pre-existing CI Flake (Not Caused by This PR)
Shard 2 fails with `ClipThumbnailImage renders a custom placeholder for a missing file`. This test failure:
- Exists on the PR's base commit (95a05362f)
- Exists on main's own recent CI run
- **Was fixed by PR #6874** (`81411fec3`), which landed on main *after* this PR's base
- Will clear automatically when PR is rebased onto current `origin/main`

### Requests Before Merge

**1. Rebase to current origin/main**
```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```
This will clear the pre-existing CI flake.

**2. Clarify Out of Scope coverage**
The PR body mentions "Out of Scope" but doesn't explicitly name a behavioral edge case: when `VideoThumbnailService.extractThumbnail()` fails on the rendered output, the fallback to `clip.thumbnailPath` (source thumbnail) produces an unfiltered preview with wrong timestamp. This was already broken (not introduced by this PR), but request confirmation it's documented in the scope.

**3. Verify manual testing status**
PR body says *"manual galaxy testing is still pending"* — can you confirm this was completed before undrafting? If so, update the body to reflect completion (e.g., "Tested on physical Galaxy S24 and Pixel 8a; scrolling through editor clips renders thumbnails correctly without double-paint artifacts").

### Cosmetic Notes (Minor)
- PR title: "fix(video-metadata): stop painting the clip cover's edits twice"
- Commit: "fix(metadata): remove redundant EditedPreviewOverlay from preview thumbnail"
- Consider aligning commit to title's "video-metadata" scope for consistency

### Recommendation
Safe to merge once rebased. The core behavioral change is correct and well-scoped. CI will pass after rebase.

PR #6895, opened 2026-08-08
test(flake): wait for the encoder to actually start, not for a pumped queue

### LINKED ISSUE #6896: test(flake): wait for the encoder to actually start, not for a pumped queue
Auto-created from PR #6895.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6895
**Branch:** `test/flake-clip-speed-render`

hm21:
Closes #6896

## Description

`ClipSpeedRenderService render caps concurrent native renders so the encoder does not stall` fails intermittently in `Tests (shard 2/4)` with `Expected: <3>, Actual: <2>`.

The test frees one of the two render slots and then asserts the queued third clip has started:

```dart
native.allowRenderToFinishAt(0).complete();
await pumpEventQueue();
expect(native.renderCount, 3);
```

`pumpEventQueue()` drains the microtask queue a fixed number of times. It does not wait for disk I/O — and disk I/O is exactly what sits between those two lines. The service releases the slot in a `finally` that runs only after `renderNativeVideoToFile` returns, and the test's fake does a real write before returning:

```dart
await allowRenderToFinishAt(renderIndex).future;
await File(filePath).writeAsString('rendered speed body');   // real I/O
return filePath;
```

So on a loaded box the write has not landed, the slot is not released, the queued clip has not started, and `renderCount` is still 2. Nothing about the cap is broken — the assertion just runs too early.

## Change

The fake already publishes the signal the test needs: `renderStartedAt(index)` completes the moment a render enters the encoder. Await that instead of guessing with `pumpEventQueue`, both for the cap assertion and for the sibling test that pre-loads the same three-clip backlog before calling `clear()`.

The negative assertion at the end of `drops queued renders once the editor closes` keeps `pumpEventQueue` deliberately — it runs after `Future.wait(pending)`, by which point the abandoned third render has already taken its bail-out path, so there is nothing left to wait for.

No production code changed, and no assertion was weakened: the test still pins the cap at 2 and still requires exactly one queued clip through per freed slot.

## Why this and not a longer timeout

A fixed delay large enough for a four-shard CI box is a delay this test pays on every green run, and it is still a guess. The fake's start signal is the actual event being waited on, so the test is deterministic rather than merely slower.

## Verification

Reproduced the failure before fixing it, since the repo does not take "flaky" on assertion. Eight `yes > /dev/null` workers pinning the cores, then the single test in a loop:

- **unfixed** — 1 failure in 15 runs, with the same `Expected: <3>, Actual: <2>` CI reports
- **fixed** — 0 failures in 30 runs under identical load

Also:

- `flutter test test/services/video_editor/clip_speed_render_service_test.dart` — 17 passing
- `dart format` — 0 changed

Not device-testable: test-only change, no runtime behaviour touched.

## Context

One of three independent sources of shard-2 instability currently open. The other two are #6874 (post-frame `ImageCache` disposal landing in the next test's first frame) and #6878 (test processes sharing one `path_provider` root, hence one Hive box). They do not overlap; this PR fixes neither of those and neither of those fixes this.

## Type of Change


NotThatKindOfDrLiz (approved):
Approved. The synchronization change is sound: waiting on `renderStartedAt(n)` targets the fake encoder-start signal instead of relying on bounded event-loop turns, and CI is green.

One non-blocking cleanup: the new comment says `pumpEventQueue` only drains the microtask queue, but `test_api` implements it as bounded event-loop turns via `Future(() => ...)`. The later framing is the accurate one: it yields a fixed number of turns, but does not wait for the disk I/O/native-render completion that gates slot release.

Issue #2636, opened 2026-03-30
task(privacy): define a privacy policy for bug report diagnostics


NotThatKindOfDrLiz:
## Summary

We need an explicit policy for what diagnostic data the mobile app may expose through public or semi-public support workflows, including Zendesk tickets that propagate into GitHub issues.

PR #2627 surfaced the gap clearly: improved inline summaries are useful and likely acceptable, but full diagnostic log uploads to public Blossom storage raise a different privacy question because logs can reveal behavioral timing, navigation patterns, interaction sequences, relay usage, event IDs, and other contextual data that becomes much more sensitive when aggregated and linked.

## Why this issue exists

The current implementation path in PR #2627 is intentionally cautious:

- keep improved inline diagnostic summaries in the Zendesk/GitHub path
- do not upload full logs to public Blossom storage in that path
- leave `uploadFullLogs()` available for narrower future channels such as encrypted one-to-one support delivery

That is a reasonable iteration, but we should not let the long-term privacy decision remain implicit.

## Questions to answer

- What categories of diagnostic data are acceptable in public or semi-public support flows?
- Are summarized logs acceptable where full logs are not? If so, what redaction or aggregation standard do we want?
- Is any form of full-log upload acceptable on public storage such as Blossom?
- If support needs full logs, what private storage, retention, and access model should we require?
- Which support channels count as sufficiently narrow-scope for fuller diagnostics, for example encrypted NIP-17 delivery?

## Proposed near-term policy direction

- Public support path: summarized diagnostics only
- Encrypted one-to-one support path: fuller diagnostics may be acceptable with deliberate scope
- Public blob storage: no full user diagnostic logs until policy says otherwise

## Done when

- We have a documented team decision on the boundaries above
- Current mobile bug-report flows are aligned with that policy
- Any follow-up implementation work is split into concrete tasks with the right storage and retention constraints

Relates to PR #2627.

NotThatKindOfDrLiz:
Cross-linking #2636 and #6941. They scored as near-duplicates in a sweep; they are not duplicates, they are a sequence — #6941's own body says so:

> Issue #2636 allows summarized diagnostics in public support paths but leaves full-log handling unresolved.

[...]

PR #6909, opened 2026-08-09
fix(privacy): sanitize public bug report diagnostics

NotThatKindOfDrLiz:
## Summary
- add a diagnostic data policy for public support flows and private log handling
- sanitize collected bug-report diagnostics at the service boundary, including emails, stack traces, and device/host names
- sanitize public Zendesk ticket fields at the ZendeskSupportService boundary, covering bug reports, feature requests, and content reports
- add ncryptsec1 redaction, document the hex-key residual risk, and supersede the stale Blossom full-log plan/spec
- link follow-up policy decisions for public Nostr identifiers and private full-log storage

## Verification
- dart run build_runner build --delete-conflicting-outputs
- flutter test test/unit/services/bug_report_service_test.dart test/services/zendesk_support_service_test.dart test/blocs/bug_report/bug_report_cubit_test.dart test/widgets/bug_report_dialog_test.dart test/services/bug_report_log_summary_test.dart
- flutter analyze
- pre-push hook: merge-conflict check, changed-file analyzer, generated-file verification, and changed-file tests passed

Refs #2636
Refs #6940
Refs #6941


rabble (changes requested):
Reviewed as a redaction change, so the bar is "a sanitizer that misses a case is worse than none, because it manufactures confidence." The core of this is right and it fixes a real leak. Two things block, both cheap.

## What's correct

The load-bearing move is `collectDiagnostics` returning `sanitizeSensitiveData(reportData)` at `mobile/lib/services/bug_report_service.dart:175`. That's the right place: it's the single point all three consumers pass through (Zendesk via the cubit, Blossom via `uploadFullLogs`, NIP-17 via `sendBugReportToRecipient`), so the sanitizer can't be routed around by a future caller. Before this, `BugReportCubit` read `reportData.recentLogs` and `reportData.deviceInfo` straight off the unsanitized struct and handed them to Zendesk — genuine hole, genuinely closed.

Same for the cubit: `description: reportData.userDescription` (`bug_report_cubit.dart:82`) instead of the raw `trimmedDescription`, plus `sanitizeText` on subject/steps/expected (`:79`, `:83`, `:86`). Those three fields previously had no sanitization at all on the way to a ticket body that gets mirrored publicly.

`_sanitizeDeviceInfo`'s key list checks out against what `collectDiagnostics` actually produces — `name` (iOS, `bug_report_service.dart:99`), `hostName` (macOS, `:108`), `computerName` (Windows, `:114`). No device-name key is produced that the list misses. I checked `buildDeviceDescription()` (`:956`) separately; it was already privacy-conscious and only feeds the local export, so no gap there.

Moving `buildLogsSummary` out of `bug_report_dialog.dart` into `services/` is right on layering, and the test move to `test/services/` puts it at its `lib/` mirror path.

## Blocking

### 1. The policy doc claims coverage for a path that has none

`docs/DIAGNOSTIC_DATA_POLICY.md` says feature requests are mirrored into public GitHub issues, then:

> The mobile client enforces this by sanitizing diagnostics before they leave `BugReportService` and by sanitizing each user-entered Zendesk field before submission.

The second clause isn't true for feature requests. `FeatureRequestCubit.submit` passes `subject` / `description` / `usefulness` / `whenToUse` through untouched (`mobile/lib/blocs/feature_request/feature_request_cubit.dart:53-59`), and `createFeatureRequest` writes each verbatim into the ticket body (`mobile/lib/services/zendesk_support_service.dart:1103-1113`). No sanitizer anywhere on that path. A user who pastes an nsec or an email address into a feature request still ships it public — the exact leak this PR just closed one flow over.

`ContentReportingService._createZendeskTicket` has the same shape (`mobile/lib/services/content_reporting_service.dart:606-635` writes user-entered `details` and `additionalContext` verbatim), though whether content reports are GitHub-mirrored isn't something the doc says either way.

Either wire `sanitizeText` into the feature-request cubit — same four-line shape as the bug-report one — or narrow the doc claim to the bug-report flow and file the rest. What can't ship is a policy document asserting a protection the code doesn't provide; that's the false-confidence failure the whole change is meant to prevent.

### 2. `ncryptsec1…` is an unredacted private-key format the app itself accepts

`BugReportConfig.sensitivePatterns` covers exactly one private-key shape, `nsec1[a-z0-9]{58}` (`bug_report_config.dart:42`), resting on the comment at `:45-46`:

> Private keys should always be in nsec format anyway

That's not true here. `key_import_screen.dart:206-215` accepts an `ncryptsec1…` NIP-49 encrypted key and a bare 64-char hex private key, and `auth_service.dart:2455` imports the former. So of the three private-key formats the app takes as input, the sanitizer catches one — while the doc states flatly that public payloads "must not include Nostr private keys."

`ncryptsec1` should just be added; there's no over-redaction risk because nothing legitimate shares the prefix:

```dart
RegExp('ncryptsec1[a-z0-9]+', caseSensitive: false),
```

An ncryptsec is password-wrapped, not plaintext, but publishing one to a public GitHub issue hands an attacker an offline brute-force target against a user-chosen password. It belongs in the list.

Bare hex I am **not** asking you to redact — it's genuinely indistinguishable from event ids and pubkeys, and blanket 64-hex redaction would gut triage value, which the existing comment correctly reasons about. But the comment should stop asserting the premise that makes it look handled. Suggest rewording to name hex-form keys as a known, accepted residual risk.

One thing worth a comment so nobody "cleans it up" later: `bunker://…?secret=…` is a live NIP-46 credential and is covered only incidentally, by the generic `secret[:\s=]+\S+` pattern at `:49`. That's load-bearing coverage arriving by accident.

## Strongly recommended (not blocking on its own)

**The single most important line in the diff has no test.** Revert `bug_report_service.dart:175` back to `return reportData;` and the entire suite still passes. Every sanitizer test in `mobile/test/unit/services/bug_report_service_test.dart` calls `sanitizeSensitiveData` directly, and the five `collectDiagnostics` tests (`:72`, `:96`, `:106`, `:242`, `:251`) all pass benign input. The cubit test mocks the service, so it can't catch it either. Per `.claude/rules/testing.md`, the fix that closes the actual hole needs a test that can fail:

```dart
test('collectDiagnostics returns sanitized data', () async {
  final data = await service.collectDiagnostics(
    userDescription: 'my key is $rawNsec',
  );
  expect(data.userDescription, isNot(contains('nsec1')));
});
```

The rest of the test work is sound and I want to be specific about why, since the cubit test looks tautological at first glance and isn't. It stubs `sanitizeText` with a fake, so it can't validate the regex set — but it isn't trying to. It pins the *wiring*: that the cubit routes subject/steps/expected through the sanitizer and forwards `reportData.userDescription` rather than the raw trimmed value. That's precisely what the diff changed, and it goes red if someone reverts it. The regex-level assertions live at the service level where the real sanitizer runs. Right division.

Missing on that side: no test puts an `nsec1…` in a `LogEntry` (only in `userDescription`). The email test does cover message/error/stackTrace — mirroring that shape for nsec would close it.

## Judgement call — worth a decision, not blocking

`npub1…` is not redacted, and `mobile/test/unit/services/bug_report_service_test.dart:219-237` actively pins that. Keeping the reporter's own npub is defensible, documented, and flagged as an Open Decision — fine.

But the same non-redaction applies to any *other* npub that lands in a log line, and `logsSummary` goes verbatim into the public ticket body (`zendesk_support_service.dart:926-931`). The doc justifies "the signed-in public Nostr account identifier"; it doesn't justify third-party identifiers scraped out of logs.

Two things keep this off the blocking list. First, I grepped for npub-form identifiers in log statements across `mobile/lib` and found essentially none — this codebase logs hex pubkeys, so redacting `npub1…` in the summary would be near-zero cost *and* near-zero benefit. The real exposure is hex, which per above can't be redacted without destroying triage value. Second, `**User Pubkey:**` is emitted as its own field (`:913-916`) from `userPubkey ?? _userNpub`, so redacting npub inside `_sanitizeString` wouldn't cost you the account linkage support needs.

Worth noting for the Open Decisions section: `.claude/rules/error_handling.md`'s PII rule strips `npub1…` before **Crashlytics**, a private dashboard. Stripping it for the private destination while passing it through to public GitHub inverts the sensitivity ordering. That's an argument for the Open Decision resolving toward removal, not an argument that this PR got it wrong.

## Nit

On Linux, `deviceInfo['name']` is `linuxInfo.name` — the distro name, e.g. "Ubuntu" (`bug_report_service.dart:121`), not a machine name. `_sanitizeDeviceInfo` redacts triage-useful info there. Irrelevant in practice since Linux isn't a shipping target; mentioning only so the key list doesn't get copied somewhere it matters.

---

Requesting changes on items 1 and 2. Both are small. The architecture of the fix — sanitize at the choke point in `collectDiagnostics` rather than at each call site — is the right call and I'd rather see it land with the doc telling the truth about what it covers.

mobile/lib/config/bug_report_config.dart
```diff
@@ -3,6 +3,28 @@
...
+///
+/// Both quoted branches require the closing quote and stop at a newline, so an
+/// unbalanced quote (`password: "oops` typed into a bug report) falls through
+/// to the unquoted branch and costs one word, rather than swallowing every
+/// field printed after it. The optional `bearer`/`basic`/`token` scheme word
+/// keeps `Authorization: Bearer <jwt>` from redacting only the word "Bearer".
+const _credentialValue =
+    r'''(?:"(?:\\.|[^"\\\n])*"'''
+    r"""|'(?:''|\\.|[^'\\\n])*'"""
+    r'''|["']?(?:(?:bearer|basic|token)\s+)?[^\s"',;}]+)''';
```
dcadenas, line 56: The separator class here was doing two things at once, in opposite directions.

`[:\s=]+` accepted a **bare space**, so any word following the keyword was eaten. And it required the separator to follow the keyword *immediately*, so a quote defeated it. Measured against the old patterns:

| Input | Old output |
|---|---|
| `{"password":"hunter2"}` | unchanged — not redacted |
| `{"token":"eyJhbGciOi"}` | unchanged — not redacted |
| `{"Authorization":"Bearer eyJx"}` | unchanged — not redacted |
| `Password reset failed` | `[REDACTED] failed` |
| `my password reset email never arrives` | `my [REDACTED] email never arrives` |

Serialized payloads are where credentials actually reach the log buffer, and `password` / `token` / `secret` are the words auth bug reports are made of. Since this change newly routes user-typed subject, steps and expected-behavior through these patterns, the prose cost was new too.

The rewrite requires a `:` or `=` separator with optional quotes on either side, consumes a quoted value to its closing quote so a multi-word secret is redacted whole, and enumerates keys rather than matching any suffix.

Three things worth knowing about the shape:

- **Partial redaction is treated as worse than none.** A `[REDACTED]` marker beside the surviving half of a JWT reads as proof the value was handled. The tests assert absence of the secret *fragments*, not just presence of the marker, so a partial redaction fails.
- **Both quoted branches require the closing quote and stop at a newline.** Otherwise `password: "test1` typed into steps-to-reproduce runs to end-of-input and deletes every field printed after it.
- **The segment classes are deliberately unambiguous.** Writing them as `\w` makes the same key parseable 2^n ways; a 61-character key then took 16.8s to fail to match, on the UI thread, over text that can come from a remote profile name. It is 0ms now, and a timing test pins it.

docs/DIAGNOSTIC_DATA_POLICY.md
```diff
@@ -0,0 +1,103 @@
...
+Zendesk fields at the `ZendeskSupportService` submission boundary. Text composed
+inside the native Zendesk SDK screens does not cross that boundary; see
+[Residual Risks](#residual-risks).
+
+The email redaction rule also redacts NIP-05 identifiers such as
+`name@divine.video`. That is intentional for public support payloads: NIP-05s
+are public handles, but incidentally collected handles still identify people
+and can be preserved privately in Zendesk metadata when support needs them.
+
+## Residual Risks
```
dcadenas, line 57: The enforcement sentence above needed narrowing, and this section is where the gap is now recorded.

The native Zendesk SDK screens never cross the Dart boundary. `showTicketListScreen` is live from the support center, and on Android it opens `RequestListActivity`, where the user composes a reply that the SDK posts directly to Zendesk. No regex in this change can reach that text, and it lands in the same tickets the policy says are publicly mirrored.

The Android handler for `showNewTicket` already documents that the SDK ignores the prefilled subject and description, so the sanitization applied on that path is inert there regardless.

Nothing in-repo can fix this, which is why it is stated as a residual rather than patched.

mobile/test/services/zendesk_support_service_test.dart
```diff
@@ -822,6 +866,105 @@ void main() {
...
+      expect(capturedArgs!['description'], isNot(contains(_rawNsec)));
+      expect(capturedArgs!['description'], isNot(contains(_rawNcryptsec)));
+      expect(capturedArgs!['description'], isNot(contains('liz@example.com')));
+      expect(values, isNot(contains(_rawNsec)));
+      expect(values, isNot(contains('liz@example.com')));
+      expect(capturedArgs!['description'], contains('[REDACTED]'));
+      expect(values, contains('[REDACTED]'));
+    });
+
+    test('sanitizes bug report fields before native submission', () async {
```
dcadenas, line 1067: This is the test for the path the change relocates.

Moving sanitization out of `BugReportCubit` and down to the submission boundary is the right call — it is a choke point that a future caller cannot route around by accident. But it left the composition untested: the cubit test now pins the raw pass-through, and the new boundary tests exercise `createTicket` directly, so nothing covered the production bug-report path in between.

Verified this can fail: replacing the `sanitizeDiagnosticText` calls in `createTicket` with pass-throughs turns it red.

One gap I could not close here. `createTicketViaApi` and `_createTicketWithFormViaApi` both sanitize, but neither is reachable under test — they return early on `ZendeskConfig.isRestApiConfigured`, which reads a compile-time `String.fromEnvironment` that is empty in CI. Pinning them needs either a Zendesk token in CI or an injectable config, neither of which belongs in this change. They are the desktop fallbacks; the native path that ships on iOS and Android is covered.

dcadenas (approved):
The choke-point design is right, and it is what made the rest of this reviewable.

Sanitizing in `collectDiagnostics` and at the `ZendeskSupportService` submission boundary means every Zendesk caller passes through one of four functions, and no path in `lib/` routes around them. Feature requests and content reports are genuinely covered now, and `ncryptsec1` closes a real private-key format.

The credential regexes themselves needed work, in both directions at once: serialized payloads like `{"token":"..."}` were passing through untouched, while ordinary prose such as "password reset failed" was being shredded. Detail inline.

### Changes I pushed

- Rewrote the credential patterns to require a `:`/`=` separator and to consume quoted values whole.
- Added a test for `createStructuredBugReport`, the production bug-report path this change relocates sanitization away from.
- Narrowed the policy doc's enforcement claim and recorded what redaction does not cover.

The credential-pattern work is the widest change and it lands as four commits rather than one, because each fixed a defect the previous one introduced — partial redaction, then an unbalanced quote that swallowed the rest of the payload, then exponential backtracking. Each is isolated and separately tested.

It gives something up. Keyword rules cannot separate `token_value` from `token_count`, so counters and status fields under credential-shaped names are now redacted, and a credential written without a separator is not. Both directions are written into the policy doc rather than left to be discovered.

### Checks

- `flutter analyze lib test`: no issues
- `flutter test` across `test/config`, `test/services`, `test/unit/services`, `test/blocs/bug_report`, `test/widgets/bug_report_dialog_test.dart`: 4084 passed
- `check_untested_services_floor.sh`: no new entries
- Mutation-checked each new assertion by reverting the code it guards

### Follow-ups

- The signed-in npub in public payloads (#6940) and private full-log storage (#6941) stay open as the description says. Neither blocks this branch.

NotThatKindOfDrLiz:
Pushed a follow-up at a80ee004d.

Changes:
- redacts bracketed credential values like JSON/header arrays so partial array values do not survive next to `[REDACTED]`
- covers plural credential keys such as `tokens`, `secrets`, `passwords`, and `apiKeys`
- removes the now-unwired NIP-17/Blossom full-log bug-report paths, stale constants, and their test/baseline entries
- clarifies requester identity and residual-risk wording in the diagnostic policy docs

Verification:
- `flutter test test/config/bug_report_config_test.dart test/unit/services/bug_report_service_test.dart test/services/zendesk_support_service_test.dart test/blocs/bug_report/bug_report_cubit_test.dart test/widgets/bug_report_dialog_test.dart test/services/bug_report_log_summary_test.dart`
- `flutter test test/unit/services/upload_manager_get_by_path_test.dart`
- `flutter analyze`
- `scripts/check_test_unit_structure.sh`
- `scripts/check_shared_channel_overrides.sh`
- pre-push hook: merge-conflict check, analyzer, generated files, changed-file tests
- GitHub CI: all checks green after rerunning shard 2 (initial shard-2 failure was in untouched `upload_manager_get_by_path_test.dart`; the same file passed locally and on rerun)

NotThatKindOfDrLiz:
Rebased onto fresh `origin/main` and force-pushed with lease. Current head is fd2c3b1fe.

Post-rebase verification:
- `flutter test test/config/bug_report_config_test.dart test/unit/services/bug_report_service_test.dart test/services/zendesk_support_service_test.dart test/blocs/bug_report/bug_report_cubit_test.dart test/widgets/bug_report_dialog_test.dart test/services/bug_report_log_summary_test.dart` (`180` pass, `2` skipped)
- `flutter analyze`
- pre-push hook: merge-conflict check, analyzer, generated files, changed-file tests (`180` pass, `2` skipped)

NotThatKindOfDrLiz:
Final rebase update: current head is 226c4e8e5 after rebasing onto the latest `origin/main` and force-pushing with lease.

Pre-push verification on this exact head passed: merge-conflict check, analyzer, generated files, and changed-file tests (`180` pass, `2` skipped).

mobile/lib/config/bug_report_config.dart
```diff
@@ -3,23 +3,63 @@
...
+/// a JWT runs 300-800 characters on its own, and `bug_report_service.dart`
+/// pretty-prints device info - because below the bound is where redaction is
+/// whole and above it is where a secret survives beside the marker.
+///
+/// Both quoted branches require the closing quote and stop at a newline, so an
+/// unbalanced quote (`password: "oops` typed into a bug report) falls through
+/// to the unquoted branch and costs one word, rather than swallowing every
+/// field printed after it. The optional `bearer`/`basic`/`token` scheme word
+/// keeps `Authorization: Bearer <jwt>` from redacting only the word "Bearer".
+const _credentialValue =
```
dcadenas, line 51: Most of the credential work is in this constant.

The rules now match a value to the **last** delimiter in range rather than the first. Stopping at the first one left `{"token":{"h":{},"jwt":"…"}}` half-redacted with the jwt still beside the marker, and no depth-limited pattern fixes that for arbitrary nesting.

Both branch pairs are bounded at 4000 characters, and the bound does two jobs. It caps what an unclosed `{` can consume, and it keeps the scan linear — unbounded, every `token: {` candidate rescans to the end of the line, which measured 3.9s on 100KB of pasted text and 133/282/585ms once bounded at 2000/4000/8000 repetitions.

The separator also consumes markdown emphasis. Device info renders as `- **sessionKey:** <value>`, and without that the value branch matched the `**` and reported success while the secret sat one character further along.

mobile/lib/services/zendesk_support_service.dart
```diff
@@ -885,30 +916,44 @@ class ZendeskSupportService {
     final effectiveSubject = subject.isNotEmpty
         ? subject
         : 'Bug Report: $reportId';
+    // Every contributed value is sanitized here, before assembly, rather than
```
dcadenas, line 919: This is the change with the largest user-visible effect.

Sanitizing the assembled description meant a bounded run of redaction could reach across everything appended after the offending field. On a reconstructed 2505-character report, one plausible sentence in the description — `my password: {weird symbols and it died` — produced a 41-character ticket. App version, repro steps, device info, pubkey and every log line were gone, with nothing telling you or support why.

Each field is sanitized before assembly now, so the loss stops inside the field that caused it. `createTicket` still sanitizes the assembled result; that pass is idempotent.

Device info and error counts are sanitized as composed `key: value` lines rather than as bare values — the rules identify a credential by its key, so a value handed over alone has nothing to match against.

mobile/lib/services/content_reporting_service.dart
```diff
@@ -210,14 +211,23 @@ class ContentReportingService {
       // Generate report ID
       final reportId = 'report_${DateTime.now().millisecondsSinceEpoch}';
 
+      // Redact once, here, so every projection of this report inherits it.
```
dcadenas, line 212: The Zendesk ticket was not the only destination for these two strings.

The kind-1984 event carried them to public relays unredacted — `Details: they posted my password: hunter2 and my token: abc123`, with the `alt` tag holding `api_key=…` verbatim. A relay event is worse than a support ticket in every direction that matters here: world-readable, permanent, unretractable.

Redacting once at the entry point rather than at each sink is deliberate. Three sinks were found one at a time — Zendesk, then the event, then the moderation DM — and this placement means a fourth cannot be missed.

Local history stores the redacted copy too. Nothing reads it today, which is exactly why the raw text should not sit there waiting for the first feature that surfaces or exports it.

mobile/lib/widgets/support_capped_text_field.dart
```diff
@@ -0,0 +1,155 @@
...
+/// nothing else in the payload carries it, since attachments are images only
+/// and the log summary comes from in-app logs. So the field reports when the
+/// formatter actually rejected characters.
+///
+/// The notice is driven by a rejection the formatter reports, not by comparing
+/// the text length to the cap. Those are different questions and, before they
+/// were separated, different units: [LengthLimitingTextInputFormatter] counts
+/// grapheme clusters while `String.length` counts UTF-16 code units, so a
+/// field of emoji claimed to be full at half its real capacity.
+class SupportCappedTextField extends StatefulWidget {
```
dcadenas, line 30: This widget is new and it is the part I would most like a second opinion on, since it changes what users see.

The cap is an input formatter rather than `maxLength`, so the forms do not grow four Material character counters in styling the design system does not own.

A formatter alone would drop the tail of a pasted crash log silently, and nothing else in the payload carries it — attachments are images only, and the log summary comes from in-app logs rather than pasted text. So the field reports when the limiter actually rejected characters, and announces it for screen readers.

The notice is driven by a rejection the limiter reports, not by comparing text length to the cap. Those turned out to be different units: the limiter counts grapheme clusters, so a field of emoji claimed to be full at half its real capacity.

docs/DIAGNOSTIC_DATA_POLICY.md
```diff
@@ -0,0 +1,152 @@
...
+  is a bound on pathological input rather than on anything support sees.
+- Free-text fields on the bug report, feature request and content report forms
+  are capped at 10000 characters and subjects at 200, so a pasted field cannot
+  make sanitization slow enough to freeze submission. The cap is an input
+  formatter rather than `maxLength`, to keep four Material character counters
+  off the support forms, and the form says "that's the maximum length" once a
+  field is full. Truncation must not be silent: nothing else in the payload
+  carries the dropped tail, since attachments are images only and the log
+  summary comes from in-app logs rather than pasted text.
+- A credential key written with nothing after it (`...my password:`) consumes
```
dcadenas, line 114: This residual is a consequence of the markdown-emphasis change to the separator, and it is over-redaction rather than a leak.

A field ending in a bare credential key consumes the start of whatever follows, across the line break. Between assembled fields that costs a few characters of the next heading; inside the log summary, where entries abut with no heading between them, it can cost the start of the next entry.

The alternative — refusing to span a line break — would leave `password:` followed by its value on the next line unredacted, so this is the safe direction of the trade. Listed here rather than left implicit, since the section names every other one.

dcadenas (approved):
🔒 the redaction boundary holds now, and the residuals are written down honestly.

I pushed 23 commits onto the branch rather than sending this back. They harden what you built; the shape of the change is yours.

## What the commits change

- Credential coverage: plural key names, `jwt`, and values that are nested, pretty-printed, bracketed or braced. Several of these leaked whole secrets next to a `[REDACTED]` marker, which reads as proof the value was handled.
- Sanitization moved to each contributed field before assembly, in all three ticket builders. Assembling first let one typed field erase the ones printed after it.
- Content reports redact at the entry point, so the kind-1984 event and the moderation DM are covered, not only the Zendesk projection.
- Free-text fields are capped, and the form now says when a paste was truncated.

## Worth your attention

Two are user-visible and neither was requested: the character caps on the support forms, and the notice that appears when one truncates a paste. Both exist because sanitization runs on the main isolate and cost scales with field size — an uncapped paste measured 7.7s at 1MB. Revert them if you disagree; nothing else depends on them.

Details are inline.

mobile/lib/config/bug_report_config.dart
NotThatKindOfDrLiz: Acknowledged. Keeping this credential-pattern rewrite and its tests as-is after re-review; resolving.

docs/DIAGNOSTIC_DATA_POLICY.md
NotThatKindOfDrLiz: Acknowledged. Keeping the native Zendesk SDK residual documented here as-is; resolving.

mobile/test/services/zendesk_support_service_test.dart
NotThatKindOfDrLiz: Acknowledged. Keeping the Zendesk boundary coverage and the documented REST fallback test gap as-is; resolving.

mobile/lib/config/bug_report_config.dart
NotThatKindOfDrLiz: Acknowledged. Keeping the bounded collection-value behavior and documented tradeoff as-is; resolving.

mobile/lib/services/zendesk_support_service.dart
NotThatKindOfDrLiz: Acknowledged. Keeping per-field sanitization before ticket assembly; resolving.

mobile/lib/services/content_reporting_service.dart
NotThatKindOfDrLiz: Acknowledged. Keeping content-report redaction at the reportContent entry point so every projection inherits it; resolving.

mobile/lib/widgets/support_capped_text_field.dart
NotThatKindOfDrLiz: I am keeping the caps and truncation notice. The bounded input is the right tradeoff for main-isolate sanitizer cost, and the notice avoids silent truncation. Resolving.

docs/DIAGNOSTIC_DATA_POLICY.md
NotThatKindOfDrLiz: Acknowledged. Keeping this over-redaction residual documented as-is; resolving.

NotThatKindOfDrLiz:
@rabble please re-review when you have a chance.

I kept the support-form caps and truncation notice. The bounded input is the right tradeoff for sanitizer cost on the main isolate, and the notice makes truncation explicit rather than silent. I filed the remaining off-main-isolate sanitizer work as #7080.

Current head after rebasing onto fresh origin/main: d2b97f2535f2d999db745341598683f7dd549569.

Your two requested changes are addressed in the current diff:
- Feature requests and content reports now cross sanitizer boundaries before Zendesk/public projections, and the policy doc is narrowed to tickets the mobile client submits itself while naming the native-SDK residual.
- ncryptsec1 is redacted, hex-form private keys are documented as an accepted residual, and the NIP-46 bunker secret coverage is called out.

I also replied to and resolved the 8 inline self-annotation threads from dcadenas.

Verification on the rebased head:
- flutter analyze
- mise exec -- flutter analyze lib test integration_test
- flutter test test/config/bug_report_config_test.dart test/unit/services/bug_report_service_test.dart test/services/zendesk_support_service_test.dart test/services/content_reporting_service_test.dart test/blocs/bug_report/bug_report_cubit_test.dart test/widgets/bug_report_dialog_test.dart test/widgets/feature_request_dialog_test.dart test/widgets/report_content_dialog_test.dart test/services/bug_report_log_summary_test.dart test/l10n/arb_consistency_test.dart
- pre-push hook: merge-conflict check, analyzer, generated files, changed-file tests

mobile/lib/config/bug_report_config.dart
```diff
@@ -42,12 +100,82 @@ class BugReportConfig {
...
+    // values - redacting it deletes a diagnostic built to be privacy-safe.
+    // That guard is anchored to the start of the key, so it exempts `queryKeys`
+    // without also exempting `bigQueryKey`.
+    RegExp(
+      '(?<=[A-Za-z])(?<![Pp]ub)(?<![Pp]ublic)'
+      '(?<![Pp]hysical)(?<![Ll]ogical)(?<!(?<![A-Za-z])[Qq]uery)Keys?'
+      '(?:[A-Z][a-z0-9]*){0,24}'
+      '$_credentialSeparator$_credentialValue',
+    ),
+    RegExp(r'\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}\b', caseSensitive: false),
```
mbradley: This is the one pattern in `sensitivePatterns` that didn't get bounded when the credential patterns were hardened, and it's reachable with a real cost. Its local-part and domain quantifiers are unbounded, which makes matching quadratic in the size of the input, and `sanitizeDiagnosticText` runs synchronously on the UI isolate.

It bites on the log-summary path specifically: `buildLogsSummary` sanitizes each entry at full length before truncating to `maxLogEntryLength`, so a single large entry (long URLs, dotted package paths, hyphenated ids) pays the full-size cost, and that summary goes into the public Zendesk ticket. Because it's quadratic, a large entry stalls the main thread on the order of seconds and gets worse fast as the entry grows. The credential patterns already carry bounds for this exact reason (the segment caps and the 4000-char collection bound); this one was missed.

Bounding the quantifiers to RFC limits fixes it: matching goes linear (well under 100ms even at tens of KB), and I checked it still redacts ordinary and NIP-05 emails:

```suggestion
    RegExp(r'\b[A-Z0-9._%+-]{1,64}@[A-Z0-9.-]{1,255}\.[A-Z]{2,24}\b', caseSensitive: false),
```

mobile/lib/services/content_reporting_service.dart
```diff
@@ -210,14 +211,23 @@ class ContentReportingService {
...
+      // The kind-1984 event is published to relays in plaintext and the
+      // Zendesk ticket can be mirrored publicly, so a credential pasted into
+      // the details field must not survive into either.
+      final safeDetails = sanitizeDiagnosticText(details);
```
mbradley, line 216: @dcadenas @NotThatKindOfDrLiz — not a blocker, just putting the reasoning on record for the content-report redaction.

Running the reporter's free-text through the sanitizer is the right call as defense-in-depth. The kind-1984 event is public and permanent, so a credential or someone's email pasted into a report shouldn't be broadcast there. The one place it cuts the other way is that for content reports the free-text is sometimes the evidence itself (an impersonated handle, a harasser's email), and that gets redacted everywhere, including the moderators' own DM.

What keeps that from being a problem is the reporting structure. The report still carries the reported event id, author pubkey, and reason, and the reporter is identified on all three destinations. The moderation DM is a NIP-17 message from the reporter, so if a report reaches us without enough context, we can reply and ask. Redaction can cost some at-a-glance context, but not the ability to recover it.

mbradley (commented):
Reviewed the current head as a redaction change. The choke-point design holds up: I traced every path from user or diagnostic text to a public or shared sink and each one is sanitized, no secret survived the sanitizer in my testing, and the Zendesk requester identity is correctly kept out of the sanitizer so support can still reach reporters. One thing worth fixing and one worth noting, both inline.

NotThatKindOfDrLiz:
Pushed 373d1489a after rebasing onto current origin/main.

Changes:
- bounded the email redaction regex so boundary-heavy pasted text cannot make the sanitizer scan quadratically
- added a regression timing test with email-local/domain-legal separators (`a.b-c_`) so the missed failure mode is covered directly
- removed the stale `<100ms per field` doc claim
- deleted the now-unwired `BugReportResult.success` factory and `messageEventId` field after the NIP-17 bug-report send path was removed

Verification:
- `flutter test test/config/bug_report_config_test.dart`
- `flutter test test/config/bug_report_config_test.dart test/unit/services/bug_report_service_test.dart test/services/zendesk_support_service_test.dart test/services/content_reporting_service_test.dart test/blocs/bug_report/bug_report_cubit_test.dart test/widgets/bug_report_dialog_test.dart test/widgets/feature_request_dialog_test.dart test/widgets/report_content_dialog_test.dart test/services/bug_report_log_summary_test.dart test/l10n/arb_consistency_test.dart`
- `flutter test` from `mobile/packages/models`
- `flutter analyze`
- pre-push hook: merge-conflict check, analyzer, generated files, changed-file tests

@mbradley I replied to and resolved both of your threads. @rabble please re-review the current head when you have a chance.

mobile/lib/config/bug_report_config.dart
mbradley: Confirmed at head `373d1489a`: the bound still redacts ordinary and NIP-05 emails, matching is linear, and the regression test is load-bearing — the old unbounded pattern blows its 1s ceiling by ~17s on the test input. Good to close.

mbradley (approved):
Re-reviewed at `373d1489a`. The email-redaction bound and its regression test check out (details on the `bug_report_config.dart` thread), and the redaction choke points, sink coverage, and Zendesk requester-identity handling all held up in the prior pass. Approving.

NotThatKindOfDrLiz:
@rabble please re-review when you have a chance.

I pushed d9e752270 after rebasing onto current origin/main.

Changes:
- filed #7100 for the deferred support-flow localization pass and cited it from the supportFieldLimitReached allowlist comment
- removed the now-orphaned BlossomUploadService.uploadBugReport API and the tests that only kept that dead bug-report upload path alive

Verification:
- flutter test test/src/blossom_upload_service_test.dart from mobile/packages/blossom_upload_service
- flutter analyze
- pre-push hook: merge-conflict check, analyzer, generated files, changed-file tests
- GitHub CI: all checks green on d9e752270

NotThatKindOfDrLiz:
@rabble please re-review when you have a chance.

I rebased onto current origin/main and force-pushed with lease. Current head is 9a3b92e8e3b136ac9fc5bc8e2f5a9cd4053f2c5e.

Your two requested changes remain addressed in the rebased diff:
- Feature requests and content reports cross sanitizer boundaries before Zendesk/public projections; content reports are redacted at reportContent before Zendesk, kind-1984, and moderation-DM projections.
- ncryptsec1 is covered by the sanitizer, and hex-form private keys are documented as an accepted residual rather than falsely claimed handled.

The later ReDoS finding from mbradley is also fixed with the bounded email regex and timing regression test, and all review threads are resolved.

Verification before push:
- pre-push hook: merge-conflict check, analyzer, generated files, changed-file tests
- analyzer: no issues
- changed-file tests: all passed (507 passed, 5 skipped)

NotThatKindOfDrLiz:
@rabble quick update: main moved again while CI was running, so I rebased once more and force-pushed with lease.

Current head is a2b7c7433c155068eab2ef7a0090920ac0f26638, rebased onto origin/main at 7e56fb5a889e5033f766c73332c3003c063ee86f.

Pre-push verification passed again:
- merge-conflict check
- analyzer: no issues
- generated files
- changed-file tests: all passed (507 passed, 5 skipped)

rabble (commented):
Re-reviewed as a redaction change, same bar as before. **Both blocking items addressed, and the strongly-recommended test landed.** CI green (16/16) on `a2b7c7433`.

**1. Doc/code mismatch — fixed at the better layer.** Rather than bolting `sanitizeText` onto `FeatureRequestCubit`, sanitization moved to the `ZendeskSupportService` submission boundary (`:457-466`, plus `_sanitizeCustomFields` at `:86`). That's the right choice for the same reason `collectDiagnostics` was: it's the choke point every ticket path crosses, so a future caller can't route around it. The feature-request path is covered as a consequence rather than as a special case.

The doc claim is now accurate and appropriately narrow — "enforces this **for tickets it submits itself**" — and the SDK-composed text that genuinely isn't covered is written up under Residual Risks instead of being papered over. That was the whole point of the item: a policy doc must not assert protection the code doesn't provide.

You also went past what I raised, redacting content-report details at `reportContent`'s entry point before any of the three projections is built. Doing it once at entry rather than per-destination is correct given the kind-1984 event is world-readable and unretractable — that's the most exposed of the three and the easiest to forget.

**2. `ncryptsec1` is in.** `RegExp('ncryptsec1[a-z0-9]+', caseSensitive: false)` at `:102`. The bare-hex comment now names it as a known accepted residual risk rather than asserting the false premise that keys "should always be in nsec format anyway", which was the part that made the gap look handled. The key-*name* redaction (`rawKeyHex`, `privateKeys`) is called out as load-bearing precisely because the value form isn't redacted — good that it's written down, since that's exactly the kind of coupling someone tidies away later.

**3. The untestable line is now tested.** `collectDiagnostics returns sanitized description` exercises the real return path, so reverting `bug_report_service.dart:175` to `return reportData;` now goes red. That was the single most important line in the original diff and nothing could catch its removal.

One residual worth stating plainly rather than leaving implied: this closes the paths that cross the app's own boundaries. Screenshots and attachments still can't be made safe by regex, and the doc says so. That's the right disclosure, not a gap in this PR.

rabble (approved):

PR #6910, opened 2026-08-09
feat(captions): scope feed toggle to current video

### LINKED ISSUE #3506: feat: scope video info panel captions toggle to current video
## What would you like?

When I tap ℹ️ on a video and toggle captions off, that toggle should apply only to the current video for the current viewing session. Right now it changes the global setting — the same one accessible in Settings — which affects every video in the app.

The fix: the info panel toggle becomes a per-video, session-scoped override. Navigating away from the video reverts to the global default. The Settings toggle remains the source of truth for that default and is untouched.

Controls inside an item's inspector are expected to act on that item, not silently change a global preference. A viewer who turns captions off on one noisy video shouldn't lose captions on every subsequent video. The current behaviour breaks the inspector model and surprises users who expect scoped actions from a scoped panel.

## When would you use this?

When a single video has particularly dense, inaccurate, or distracting captions and I want to watch it without them — while keeping captions on as my default for everything else.

## Anything else?

The global default already lives in Settings — the right home for it. The info panel toggle should be a lightweight, session-scoped override with no persistence: navigate away, and the next video plays with your Settings default. This matches the inspector convention used by macOS Get Info, Xcode's inspector sidebar, and design tools like Figma.

NotThatKindOfDrLiz:
## Summary
- add a non-persisted per-video captions visibility override layered over the persisted Settings default
- scope feed captions controls in the paused overlay, fullscreen popover, and home top-bar popover to the current video
- preserve scoped caption overrides through temporary feed pauses and clear them on active-video changes or owned teardown
- keep Settings as the only persisted default and track deferred scoped-caption localization debt

## State management note
- This extends the existing Riverpod subtitle visibility provider surface instead of adding a Cubit bridge. The per-video provider is auto-disposed and scoped to video ids.

## Verification
- flutter gen-l10n
- flutter analyze lib test
- flutter test test/providers/subtitle_visibility_test.dart test/widgets/video_feed_item/feed_playback_toggles_pill_test.dart test/widgets/video_feed_item/subtitle_overlay_test.dart test/widgets/video_feed_item/paused_video_overlay_test.dart test/screens/feed/feed_mode_switch_test.dart test/widgets/video_feed_item/feed_videos_test.dart test/l10n/arb_consistency_test.dart
- pre-push hook: merge-conflict check, analyzer, generated files check, changed-file tests

Closes #3506

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -4177,8 +4177,6 @@
     "videoSettingsAutoAdvanceOff": "ራስ-ሰር ማስቀደም ጠፍቷል",
     "videoSettingsCaptionsOn": "መግለጫ ጽሑፎች በርተዋል",
     "videoSettingsCaptionsOff": "መግለጫ ጽሑፎች ጠፍተዋል",
-    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
-    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
     "videoActionReportLabel": "ሪፖርት",
     "videoActionReport": "ቪዲዮውን ሪፖርት አድርግ",
     "videoActionEditLabel": "አርትዕ",
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -4177,6 +4177,8 @@
     "videoSettingsAutoAdvanceOff": "ራስ-ሰር ማስቀደም ጠፍቷል",
     "videoSettingsCaptionsOn": "መግለጫ ጽሑፎች በርተዋል",
     "videoSettingsCaptionsOff": "መግለጫ ጽሑፎች ጠፍተዋል",
+    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
+    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
     "videoActionReportLabel": "ሪፖርት",
     "videoActionReport": "ቪዲዮውን ሪፖርት አድርግ",
     "videoActionEditLabel": "አርትዕ",
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -4052,6 +4052,8 @@
     "videoSettingsAutoAdvanceOff": "التشغيل التلقائي متوقف",
     "videoSettingsCaptionsOn": "الترجمة مفعّلة",
     "videoSettingsCaptionsOff": "الترجمة متوقفة",
+    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
+    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
     "videoActionReportLabel": "إبلاغ",
     "videoActionReport": "أبلِغ عن الفيديو",
     "videoActionEditLabel": "تعديل",
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -4052,8 +4052,6 @@
     "videoSettingsAutoAdvanceOff": "التشغيل التلقائي متوقف",
     "videoSettingsCaptionsOn": "الترجمة مفعّلة",
     "videoSettingsCaptionsOff": "الترجمة متوقفة",
-    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
-    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
     "videoActionReportLabel": "إبلاغ",
     "videoActionReport": "أبلِغ عن الفيديو",
     "videoActionEditLabel": "تعديل",
--- a/mobile/lib/l10n/app_bg.arb
+++ b/mobile/lib/l10n/app_bg.arb
@@ -4183,8 +4183,6 @@
     "videoSettingsAutoAdvanceOff": "Автоматичното продължаване е изключено",
     "videoSettingsCaptionsOn": "Субтитрите са включени",
     "videoSettingsCaptionsOff": "Субтитрите са изключени",
-    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
-    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
     "videoActionReportLabel": "Сигнал",
     "videoActionReport": "Подай сигнал за видеото",
     "videoActionEditLabel": "Редактирай",
--- a/mobile/lib/l10n/app_bg.arb
+++ b/mobile/lib/l10n/app_bg.arb
@@ -4183,6 +4183,8 @@
     "videoSettingsAutoAdvanceOff": "Автоматичното продължаване е изключено",
     "videoSettingsCaptionsOn": "Субтитрите са включени",
     "videoSettingsCaptionsOff": "Субтитрите са изключени",
+    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
+    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
     "videoActionReportLabel": "Сигнал",
     "videoActionReport": "Подай сигнал за видеото",
     "videoActionEditLabel": "Редактирай",
--- a/mobile/lib/l10n/app_de.arb
+++ b/mobile/lib/l10n/app_de.arb
@@ -4052,8 +4052,6 @@
     "videoSettingsAutoAdvanceOff": "Automatisches Weiterblättern aus",
     "videoSettingsCaptionsOn": "Untertitel an",
     "videoSettingsCaptionsOff": "Untertitel aus",
-    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
-    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
     "videoActionReportLabel": "Melden",
     "videoActionReport": "Video melden",
     "videoActionEditLabel": "Bearbeiten",
--- a/mobile/lib/l10n/app_de.arb
+++ b/mobile/lib/l10n/app_de.arb
@@ -4052,6 +4052,8 @@
     "videoSettingsAutoAdvanceOff": "Automatisches Weiterblättern aus",
     "videoSettingsCaptionsOn": "Untertitel an",
     "videoSettingsCaptionsOff": "Untertitel aus",
+    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
+    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
     "videoActionReportLabel": "Melden",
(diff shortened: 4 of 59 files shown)


rabble (changes requested):
The scoping model itself is sound — a single `(videoId, visible)` override that the global preference falls through is a clean way to express "this video only", and `subtitle_overlay.dart:46` reading `subtitleVisibilityForVideoProvider(widget.video.id)` means preloaded neighbours in the PageView correctly keep the global value. I could not find a way to leak the override from one video to another through the provider layer.

Three things block: an unbounded provider family, translations that aren't translations, and an unrequested change that adds a delete affordance to a new surface.

## Blocking

### 1. `subtitleVisibilityForVideoProvider` is a non-autoDispose family keyed on video id

`mobile/lib/providers/subtitle_providers.dart:35-43`:

```dart
final ProviderFamily<bool, String> subtitleVisibilityForVideoProvider =
    Provider.family<bool, String>((ref, videoId) { ... });
```

`ProviderFamily`'s constructor defaults to `isAutoDispose: false` (riverpod 3.3.2, `lib/src/providers/provider.dart:361`). Codegen `@riverpod` auto-disposes; this hand-written family does not. In an infinite feed that means one provider element retained per video id for the lifetime of the app, and each of them stays subscribed to `subtitleVisibilityOverrideProvider` — so every captions toggle notifies a listener for every video the user has ever scrolled past. `.claude/rules/state_management.md` is explicit: "Always enable `autoDispose` for parameterized providers to prevent memory leaks."

Fix is `isAutoDispose: true` on the family. Worth a test that a scrolled-past id's element is disposed, since nothing currently would catch a regression here.

### 2. Twenty locales received the English string, not a translation

`app_ja.arb:3970-3971` on this branch:

```
    "videoSettingsCaptionsOn": "字幕オン",
    "videoSettingsCaptionsOff": "字幕オフ",
    "videoSettingsCaptionsOnForVideo": "Captions on for this video",
    "videoSettingsCaptionsOffForVideo": "Captions off for this video",
```

Same in all 20 non-`en` files. `arb_consistency_test.dart` only compares key sets, so this is green — but it's green by defeating the check rather than satisfying it. The repo has a sanctioned mechanism for deferred translation (`_knownUntranslatedDebt`, `arb_consistency_test.dart:374`, currently tracking ~14 sound-library keys), and copying English into every locale is strictly worse than using it: a real ja user sees English either way, and now nothing anywhere records that these two strings need translating. They'll sit in `app_ja.arb` looking translated forever.

Please add the keys to `app_en.arb` only and list both in `_knownUntranslatedDebt` with the usual "deferred to the next X pass" comment — or get real translations.

### 3. Unrequested: the feed-mode-switch settings menu gains Edit and Delete

`mobile/lib/screens/feed/feed_mode_switch.dart:78` changes `const FeedSettingsMenu()` to `FeedSettingsMenu(video: activeVideo)`.

Only `videoId` is needed for the captions scoping — and `FeedSettingsMenu` already threads it separately (`feed_settings_menu.dart:149`). Passing the whole `VideoEvent` also flips `_isOwnVideo` (`feed_settings_menu.dart:80-87`) from permanently-false to true on your own videos, which makes `_OwnerVideoActionsPill` render in this popover for the first time — i.e. this surface now offers **Edit video** and **Delete video**, including the delete confirmation flow.

That may well be desirable, but it isn't captions scoping, it isn't in the title or description, and adding a destructive action to a surface that didn't have one is exactly the kind of user-visible blast radius that has to be called out rather than arrive as a side effect. Either pass `videoId: activeVideo?.id` and keep this PR to its stated scope, or keep the change and say so explicitly in the description with a screenshot.

## Should fix

### 4. The riskiest code in the PR has no test

Everything that decides *when* the override is cleared lives in `feed_videos.dart` — `_syncScopedSubtitleVisibility` / `_clearScopedSubtitleVisibility` (:141-171) and their six call sites (:247-260, :291, :299, :333-335, :400). None of it is tested. `feed_videos_test.dart` already exists as a harness and wasn't touched.

Concretely: delete the `_syncScopedSubtitleVisibility(video.id)` call at `feed_videos.dart:400` and every test in this PR still passes, while the override now leaks from the video you toggled onto whatever you scroll to next. The unit tests at `subtitle_visibility_test.dart:128` prove `clearUnlessVideo` *works*; nothing proves `FeedVideos` *calls* it. A widget test that pumps the feed, sets an override, advances the active video, and asserts the override is null would close the gap.

### 5. The override is dropped on route push and never restored

`didPushNext()` (`feed_videos.dart:299`) clears the override, and because it also sets `_activeSubtitleVideoId = null`, `onActiveVideoChanged` won't re-fire for the same video on return — so opening the comments or share sheet and coming back silently resets the per-video captions toggle. `ref.listen(appForegroundProvider, ...)` at :333 does the same on backgrounding.

Given the feature is "captions for *this video*", and the video is unchanged across a sheet open/close, resetting reads like the bug rather than the design. If it is deliberate, it needs a comment saying why and a test pinning it; if it isn't, the push/pop round trip should preserve the override.

## Nits

- `import 'package:flutter_riverpod/misc.dart';` (`subtitle_providers.dart:5`) exists only to name the explicit `ProviderFamily<bool, String>` type annotation. Dropping the annotation and letting inference do its job removes the import.
- `dispose()` defers `clear()` to a microtask (`feed_videos.dart:291`). If another feed instance mounts and sets an override within the same frame, the deferred clear wipes it. Low probability given `onActiveVideoChanged` fires post-layout, but the immediate/deferred split isn't documented and the `defer` flag has no comment explaining which callers need it.
- `.claude/rules/state_management.md` puts new Riverpod providers for feature UI state on the disallowed list. Extending the already-Riverpod `subtitleVisibilityProvider` is the pragmatic call here rather than bridging a Cubit into `SubtitleCueStreamPill`, but it's worth a line in the description acknowledging it so it doesn't read as an oversight.

The provider-level and widget-level tests you did write are good ones — `subtitle_overlay_test.dart:159` in particular would fail loudly if the per-video lookup regressed to the global. The gap is one layer up.

NotThatKindOfDrLiz:
Addressed the review feedback in `718e5c53b`.

What changed:
- made `subtitleVisibilityForVideoProvider` auto-disposed and removed the `misc.dart` import
- changed the home feed settings menu to pass only `videoId`, so Edit/Delete are not added to that surface
- removed English fallback strings from non-English ARBs and tracked the two scoped-caption strings as deferred localization debt
- made scoped-caption clearing ownership-aware and preserved same-video overrides through route/overlay/background pauses
- added `FeedVideos` regression coverage for active-video changes, route push/pop preservation, and cross-instance dispose safety

Verification:
- `flutter gen-l10n`
- `flutter analyze lib test`
- `flutter test test/providers/subtitle_visibility_test.dart test/widgets/video_feed_item/feed_playback_toggles_pill_test.dart test/widgets/video_feed_item/subtitle_overlay_test.dart test/widgets/video_feed_item/paused_video_overlay_test.dart test/screens/feed/feed_mode_switch_test.dart test/widgets/video_feed_item/feed_videos_test.dart test/l10n/arb_consistency_test.dart`
- pre-push hook passed merge-conflict check, analyzer, generated files check, and changed-file tests

NotThatKindOfDrLiz:
Addressed the follow-up review findings in `2c02d67d2a` and rebased the PR onto current `origin/main`.

What changed:
- resolved the subtitle provider conflict by keeping main's `result.cues` return path
- kept `subtitleVisibilityForVideoProvider` auto-disposed with an explicit family type; removed the bare lint suppression
- removed the dead generic clear path and made scoped-caption clearing owner-aware so a fullscreen feed teardown cannot wipe the home feed's same-video override
- inverted the empty inactive branch and documented why lifecycle paths defer Riverpod notifier mutation
- replaced the feed-mode settings test with behavior coverage that opens the popover and verifies Edit/Delete are absent
- added coverage for didUpdateWidget active-video sync, positive dispose clearing, and same-video multi-owner preservation

Verification:
- pre-push hook passed merge-conflict check, analyzer, generated-files check, and changed-file tests
- `flutter analyze lib test`
- `flutter test test/providers/subtitle_visibility_test.dart test/widgets/video_feed_item/feed_playback_toggles_pill_test.dart test/widgets/video_feed_item/subtitle_overlay_test.dart test/widgets/video_feed_item/paused_video_overlay_test.dart test/screens/feed/feed_mode_switch_test.dart test/widgets/video_feed_item/feed_videos_test.dart test/l10n/arb_consistency_test.dart`

@rabble please re-review.

rabble (commented):
Re-reviewed against my five items from 2026-08-09. **All five are addressed.** CI is green (15/15) on `2c02d67d2`.

| # | Item | Status |
|---|---|---|
| 1 | Non-autoDispose provider family | **Fixed** — `Provider.autoDispose.family` |
| 2 | English copied into 20 locales | **Fixed** — `app_en.arb` only, both keys in `_knownUntranslatedDebt` |
| 3 | Unrequested Edit/Delete on the feed popover | **Fixed** — back to `FeedSettingsMenu(videoId: activeVideo?.id)` |
| 4 | Sync/clear call sites untested | **Fixed** — two new `feed_videos_test.dart` cases |
| 5 | Override dropped on route push | **Fixed** — `didPushNext` now only touches `_routeAllowsPlayback` |

Item 3 is the one I most wanted reverted rather than justified, and passing `videoId` is the right call — the destructive actions are out of this surface again.

On item 5, you didn't just stop clearing; you replaced it with owner-scoped tracking (`syncOwnerToVideo` / `clearOwner` keyed on an owner token, refcounted per video). I traced the interleavings that worried me and they hold:

- A feed disposing while another is mounted on the same video does not clear the override — `_hasActiveOwner` still sees the second owner's count.
- The `dispose()` path defers through `scheduleMicrotask`, so the microtask can outlive the container. `clearOwner` guards on `ref.mounted`, so that returns cleanly instead of throwing, and the maps die with the notifier anyway.
- Dispose-then-remount ordering (`clearOwner(A)` landing after `syncOwnerToVideo(B, v)`) leaves the override intact, because B has already claimed the video.

The `scheduleMicrotask` deferral is the kind of thing that usually reads as a workaround, but the comment names the real constraint — Notifier mutation during a Riverpod build — and it's on the two callers that genuinely run mid-build. Fine as-is.

One note, not a request: item 4's tests cover the `didUpdateWidget` and `dispose` paths. The `_syncScopedSubtitleVisibility(video.id)` call in `build` (`feed_videos.dart:410`) still isn't directly pinned. The specific hole I described — delete a sync call and everything stays green — is closed for the rebuild path, which was the leak that mattered. Not worth another test unless you want the symmetry.

**Nothing blocking from me.** My `CHANGES_REQUESTED` is stale and should not be read as an outstanding objection; it needs a maintainer to convert it. Branch is `BEHIND`, so it needs a rebase before it can merge under the strict policy.

rabble (approved):

Issue #6913, opened 2026-08-09
test(l10n): detect orphaned ARB keys


NotThatKindOfDrLiz:
Refs #2656.

## Problem

The per-video deletion copy regression was able to survive because polished ARB keys remained defined after their original widget was deleted, while other surfaces continued using older duplicate keys. The repo currently checks locale consistency, but it does not fail when source ARB keys have zero non-generated call sites.

## Goal

Add a CI guard that reports ARB keys with no non-generated Dart call sites, with an allow-list for intentionally dynamic or externally consumed keys if needed.

## Acceptance Criteria

- [ ] CI fails or clearly reports newly orphaned ARB keys.
- [ ] Generated l10n files are excluded from call-site counting.
- [ ] Intentional exceptions are documented in a small allow-list.
- [ ] The guard is covered by a focused test or script fixture.

NotThatKindOfDrLiz:
Completed by #7963 (`9359fb24b7`), which added the orphaned-ARB detector, focused tests, shrink-only baseline, and CI enforcement in `mobile/scripts/check_orphaned_arb_key_floor.sh`. Remaining baseline paydown is tracked under #8289 and #7968.

Issue #2656, opened 2026-04-01
fix(video): improve deletion UX -- clear messaging, end-to-end enforcement


mbradley:
**TL;DR:** This is the umbrella for improving per-video deletion UX. The original issue bundled three different streams with different owners and risk profiles, so the work is now split into focused child issues.

A high-profile creator tried to delete a video and was not confident it worked. Three staff members were also confused by the current message before remembering how Nostr works. As Aleysha put it, "If three staff were confused by the message, our users have no hope."

This issue is about **per-video deletion**, not account deletion. Adjacent historical account-deletion issues: #659 and #2658.

## Tracking

All scoped mobile work is complete. The broader education discovery was closed as not planned, and media retention remains tracked separately on #8190.

1. **Mobile copy polish** — #3101, closed.
   - Rewrote confirmation/success/failure copy in the mobile deletion flow.

2. **Publishing/deletion education** — #3103, closed as not planned.
   - The shipped point-of-action copy covers the demonstrated need. Broader onboarding or publish-flow education had no evidence-backed placement; reopen with a specific surface if research shows continuing confusion.

3. **End-to-end enforcement** — #3102, closed via #8193. Media retention remains tracked separately on #8190.
   - It has two halves. Notifying Blossom so the media blob is actually erased is the same gap the account-deletion and takedown paths have, so it is tracked with them on #8190 — one destination, fixed and verified once.
   - Confirming relay acceptance before the app shows success was the deletion-UX half and shipped in #8193.

## What Happened

A creator tried to delete one of their imported Vine videos. They saw something like "Video Deletion request received." Their Viner friends have the same need: curate their classics and remove content that does not represent where they are now. The current language does not inspire confidence that anything definitive happened.

## Current Messaging

**Confirmation dialog:**

> "This will send a delete request (NIP-09) to all relays. Some relays may still retain the content."

**Success message:**

> "Delete request sent successfully"

These are accurate for Nostr generally, but they do not distinguish between what Divine controls and what it does not. On Divine-controlled surfaces, deletion should feel clear and confident. On the broader network, deletion depends on third-party relays and clients.

## What Actually Happens Today

1. App creates a kind 5 deletion event and publishes it to Funnelcake
2. Funnelcake verifies authorship and removes the event
3. Video disappears from feeds and profiles

Current status:

[...]

PR #6914, opened 2026-08-09
fix(video): restore deletion copy across surfaces

### LINKED ISSUE #6912: fix(video): restore clear delete copy across surfaces
Refs #2656. Follow-up to closed copy polish issue #3101.

## Problem

Per-video deletion copy regressed after the old share menu widget was removed. The profile path still uses the polished owner-delete confirmation, but other owner-delete surfaces still show relay/request language or report success as a request being sent.

## Scope

- Reuse the shared owner-delete confirmation dialog from composable video grids.
- Use the polished Divine-controlled deletion copy in the metadata edit delete prompt.
- Use the shared deletion success copy for grid delete success.
- Remove stale redundant ARB keys and regenerate l10n.
- Add focused widget coverage for the owned grid delete confirmation path.

## Acceptance Criteria

- [ ] Composable video grid delete confirmation uses the same polished copy as the profile path.
- [ ] Metadata edit delete confirmation no longer uses relay/request warning copy.
- [ ] Successful delete copy no longer says only that a delete request was sent.
- [ ] Retired ARB keys are removed from all locales and generated localizations.
- [ ] Targeted widget, l10n, and analyzer checks pass.

NotThatKindOfDrLiz:
## Summary

- Reuse the shared owner-video delete confirmation in composable video grids.
- Align metadata edit delete confirmation and grid delete success copy with the polished deletion copy.
- Restore the delete disclosure in every locale and remove stale duplicate deletion-copy ARB keys from every locale.
- Use the bottom-sheet builder context when dismissing grid options, avoiding a GoRouter-dependent pop from the modal route.
- Add widget and ARB coverage for the owner-delete confirmation path.

Closes #6912.
Refs #2656.
Refs #6913.

## Verification

- flutter gen-l10n
- flutter test test/widgets/composable_video_grid_test.dart
- python3 .claude/skills/check-l10n/scan_strings.py mobile/lib/widgets/composable_video_grid.dart mobile/lib/widgets/video_metadata/modes/edit/video_metadata_edit_bottom_bar.dart
- bash mobile/scripts/check_raw_dialog_ceiling.sh
- flutter analyze


rabble (approved):
Approving. The consolidation is correct and the l10n side is clean — I checked the parts that usually go wrong in a 610-line-deletion copy PR and found nothing lost.

**What I verified**

- **Nothing dangles.** All five retired keys (`videoGridDeleteConfirmNote`, `videoGridDeleteSuccess`, `shareMenuDeleteWarning`, `shareMenuDeleteRequestSent`, `shareMenuDeleteRelayWarning`) have zero references left anywhere in `mobile/` on this branch, including tests and generated sources.
- **Mirroring is complete.** All 22 non-English locales plus `app_en.arb` drop exactly the same five keys, and the generated `app_localizations_*.dart` are regenerated to match. Removals shrink the template, so no `_knownUntranslatedDebt` entry is needed and `arb_consistency_test.dart` stays satisfied.
- **Only two of the five were live on main** (`videoGridDeleteConfirmNote`/`Success` in the grid, `shareMenuDeleteRelayWarning` in the metadata bar); the other two were already dead. So the deletion count is generated-code volume, not lost copy.
- **Every owner-delete surface now converges** on `showOwnerVideoDeleteConfirmationDialog`: `feed_settings_menu.dart:100`, `composable_video_grid.dart:376`, `profile/profile_videos_grid.dart:207`, `video_feed_item/actions/share_action_button.dart:396`; the metadata bar reuses the same subtitle at `video_metadata/modes/edit/video_metadata_edit_bottom_bar.dart:120`. Success copy is likewise uniform on `shareMenuVideoDeletionRequested`. The "across surfaces" claim holds.
- **No disclosure was silently dropped.** `shareMenuDeleteConfirmation` ("…may still appear on third-party Nostr clients that use other relays") keeps the third-party-visibility warning that `shareMenuDeleteRelayWarning` carried, in the polished voice #6912 asked for.

Three non-blocking notes, all optional:

**1. The new test pins the navigation change, not the copy change.** I ran the new assertion against main's grid: it fails, but with `No GoRouter found in context` at `composable_video_grid.dart:361` — i.e. what it currently detects is the `context.pop` → `Navigator.of(sheetContext).pop()` switch. It cannot detect the copy consolidation, because in `app_en.arb` `videoGridDeleteConfirmMessage` (line 624 on main) and `shareMenuDeleteConfirmation` (line 2365) are byte-identical strings — as are the title and both button labels. Someone re-inlining a grid-local dialog with `Navigator.pop` would keep this green. If you want the copy itself pinned, assert the dialog's shape rather than one string — e.g. that the confirmation renders exactly the title + one body paragraph + two actions, which is what distinguishes it from the old three-paragraph version.

**2. Four keys are newly orphaned.** `videoGridDeleteConfirmTitle`, `videoGridDeleteConfirmMessage`, `videoGridDeleteCancel`, `videoGridDeleteConfirm` had exactly one call site each on main — the inline dialog this PR deletes — and now have zero references outside `lib/l10n` while remaining in all 22 locales. Given the PR explicitly sets out to "remove stale duplicate deletion-copy ARB keys from every locale", these look like they belong in the same sweep.

**3. Acceptance criterion "success copy no longer says only that a delete request was sent" holds in English only.** `shareMenuVideoDeletionRequested` is "Video deleted" in en/bg/fil/ms/ur/vi/zh/am, but still request-language in ar, de, es, fr, id, it, ja, ko, nl, pl, pt, ro, sv, tr ("Eliminación del video solicitada", "Video-Löschung angefordert", …). That drift predates this PR — the key's English value changed at some point without the translations following — and per the copy-alignment policy it shouldn't be churned as a side effect here. Worth an issue for the next translation pass, since this PR extends that copy to more surfaces.

Checks are green and the `context`→`sheetContext` cleanup in the bottom sheet is a real improvement (it removes a `GoRouter`-dependent `pop` from a modal route) — just flagging that it isn't mentioned in the description, since it's the behavioural change with the widest blast radius in the diff.

NotThatKindOfDrLiz:
Follow-up pushed in `ea06e3f2b`.

What changed:
- Restored `shareMenuDeleteConfirmation` disclosure copy for the 14 stale locales so every locale now preserves the Divine deletion + third-party Nostr visibility warning.
- Removed the four newly orphaned `videoGridDeleteConfirm*` / `videoGridDeleteCancel` ARB keys and regenerated l10n.
- Strengthened coverage with an ARB-level disclosure guard and widget assertions that the grid uses the shared single-body owner-delete dialog shape.
- Updated the PR description to call out the `sheetContext` / modal-route pop cleanup.
- Rebased onto fresh `origin/main` after main moved again.

Verification:
- `flutter gen-l10n`
- `flutter test test/widgets/composable_video_grid_test.dart`
- `python3 .claude/skills/check-l10n/scan_strings.py mobile/lib/widgets/composable_video_grid.dart mobile/lib/widgets/video_metadata/modes/edit/video_metadata_edit_bottom_bar.dart`
- `bash mobile/scripts/check_raw_dialog_ceiling.sh`
- `flutter analyze`
- pre-push hook: analyzer + changed-file tests, including `test/widgets/video_metadata/modes/edit/video_metadata_edit_bottom_bar_test.dart`

mobile/lib/l10n/app_de.arb
```diff
@@ -1627,8 +1619,7 @@
         }
     },
     "shareMenuDeleteVideoQuestion": "Video löschen?",
-    "shareMenuDeleteRelayWarning": "Damit wird eine Löschanfrage an die Relays gesendet. Hinweis: Manche Relays haben möglicherweise noch zwischengespeicherte Kopien.",
-    "shareMenuVideoDeletionRequested": "Video-Löschung angefordert",
+    "shareMenuVideoDeletionRequested": "Video gelöscht",
```
realmeylisdev, line 1622: `bbb489c` (mine) changes this value in 14 locales: ar, de, es, fr, id, it, ja, ko, nl, pl, pt, ro, sv, tr. rabble suggested holding this for a separate translation pass, and I went the other way because it's #6912's third acceptance criterion and this PR is what repoints the grid success snackbar at the key. Flagging it separately since it's the one change here that wasn't asked for.

Each value is derived from that locale's own sibling copy rather than translated fresh, e.g. de takes "Video gelöscht" from `videoEditorClipDeleted` ("Clip gelöscht") and `libraryDraftDeletedSnackbar` ("Entwurf gelöscht"). No key set changed, so `arb_consistency_test` is unaffected.

The same 14 locales are still stale on `videoGridDeleteVideoSubtitle` and `shareMenuDeleteVideoSubtitle` — that's the rest of the cluster, and it's outside this PR, so it's #6955.

On the terminology question I checked before touching anything: the new disclosure string uses `relays` in fr, `reler` in sv, `röle` in tr and `relay-uri` in ro, which all look inconsistent against a neighbouring key. Counting occurrences per file, each of those is the dominant form already in that locale, so they're faithful and I left them alone.

realmeylisdev (approved):
lgtm. mutation-checked the new grid test before trusting it: re-inlined the old two-paragraph dialog using `Navigator.pop` (so the navigation change wasn't the discriminator) and it went red, so it really does pin the copy consolidation. all nine retired keys are gone from every locale and the generated sources, nothing dangles.

pushed three small things on top, each its own commit, revert freely: the test was finding the tile by the literal string `Video thumbnail 1`, the raw-dialog baseline still said `2` for `composable_video_grid`, and `shareMenuVideoDeletionRequested` still said "requested" in 14 locales. that last one is rabble's note 3, so it's the one to drop if you'd rather keep it for the translation pass.

PR #6920, opened 2026-08-09
fix(storage): preserve notification prefs during cache recovery

### LINKED ISSUE #6919: fix(notifications): preserve push preferences during cache recovery
## Problem

The Settings -> Storage recovery path treats the Hive box named `notifications` as a disposable cache. That box also stores durable push notification preferences under `push_preferences`.

If a user disables notification types and then runs cache recovery, the local settings snapshot can be deleted while the published-schema marker in `push_notification_preferences_dirty` survives. The settings screen then falls back to default all-on preferences and can later republish those defaults.

## Impact

Users can see notification settings that do not reflect what they previously chose, and a later toggle can re-enable categories they had deliberately disabled.

## Acceptance Criteria

- Cache recovery no longer clears or deletes the durable notification preferences Hive box.
- Disposable Hive caches are still cleared.
- Regression coverage proves push preferences survive the cache Hive wipe.

Part of #4335.

NotThatKindOfDrLiz:
## Description

Removes the durable `notifications` Hive box from `CacheRecoveryService`'s disposable cache wipe list. Push notification preferences are stored in that box under `push_preferences`, so clearing it could reset the settings UI to all-on defaults while the push service still had the previously published preferences.

Adds regression coverage that verifies notification preferences survive the Hive cache wipe while real disposable Hive cache data (`hashtag_stats`) is still cleared.

Also protects `notifications.hive`/`.lock` and `push_notification_preferences_dirty.hive`/`.lock` under `{appSupport}/openvine/` from the Application Support wipe, so the box file survives step 2 of `clearAllCaches` and not just the box-name skip in step 1.

While reworking that list the wipe set changed in both directions: `user_profiles`, `bookmarks`, `secure_keys`, and the dead Hive name `video_cache` are gone, and `hashtag_stats` and `people_lists_v1` are added. `hashtag_stats` is a short-lived cache. `people_lists_v1` is cache-scoped and refilled by sync; it does include local deletion tombstones, so this PR treats it as disposable cache state rather than durable user data. Because #6958 documents that Hive box location can depend on startup ordering, the additions may remove box files that previously survived when Hive put them outside Application Support.

The repair confirmation footprint now excludes protected durable Application Support state, so the `{size} will be removed` value matches the same protected database and Hive paths that `clearAllCaches` preserves.

**Related Issue:** Closes #6919. Part of #4335.

## Out of Scope

- Moving push preferences to a newly named durable box.
- The broader #4335 storage/cache policy work.
- Fixing Hive home-path reassignment; tracked separately in #6958.

## Verification

- `dart format lib/services/cache_recovery_service.dart test/services/cache_recovery_service_test.dart`
- `flutter analyze lib/services/cache_recovery_service.dart test/services/cache_recovery_service_test.dart`
- `flutter test test/services/cache_recovery_service_test.dart`
- Earlier PR verification before the follow-up commit:
  - `flutter test test/services/cache_recovery_service_test.dart test/services/notification_preferences_service_test.dart`
  - Pre-push hook: analyzer clean, `test/services/cache_recovery_service_test.dart` passed
  - Manual, iPhone 17 Pro Max simulator: turned Mentions off, ran Settings > General > Storage > Reset app data, confirmed `notifications.hive` survived with `"mentionsEnabled":false` and the toggle was still off after a cold restart

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/services/cache_recovery_service.dart
+++ b/mobile/lib/services/cache_recovery_service.dart
@@ -133,6 +133,21 @@ class CacheRecoveryService {
     'database',
   ];
 
+  /// Hive boxes holding local-only state that cache clearing must preserve.
+  ///
+  /// `notifications` is named for the feature, not its contents: its only key
+  /// is `push_preferences`, the notification settings the user chose. The
+  /// notification inbox is a Drift table inside the protected database dir.
+  /// Wiping this box drops the user back to all-on defaults, which the next
+  /// toggle republishes over the choices they made (#6919).
+  /// `push_notification_preferences_dirty` holds edits that have not reached
+  /// the push service yet plus the published-kinds schema marker, so losing it
+  /// strands an unsynced edit.
+  ///
+  /// Their files sit under `{appSupport}/openvine/` — the uploads-box
+  /// initializer re-points Hive's home path there during startup — which is
+  /// why [_clearAppSupportDirectory] has to protect them by path as well as
+  /// skipping them above.
   static const List<String> _durableHiveBoxNames = [
     'pending_uploads',
     'notifications',
--- a/mobile/lib/services/cache_recovery_service.dart
+++ b/mobile/lib/services/cache_recovery_service.dart
@@ -54,7 +54,7 @@ class CacheRecoveryService {
     }
   }
 
-  /// Clear all Hive database boxes
+  /// Clear disposable Hive cache boxes.
   static Future<int> _clearHiveBoxes() async {
     int cleared = 0;
 
@@ -62,7 +62,6 @@ class CacheRecoveryService {
       // Get all open boxes and clear them
       // Note: We iterate over known box names instead of trying to access all open boxes
       final knownBoxNames = [
-        'notifications',
         'user_profiles',
         'personal_events',
         'personal_events_metadata',
--- a/mobile/lib/services/cache_recovery_service.dart
+++ b/mobile/lib/services/cache_recovery_service.dart
@@ -59,15 +59,14 @@ class CacheRecoveryService {
     int cleared = 0;
 
     try {
-      // Get all open boxes and clear them
-      // Note: We iterate over known box names instead of trying to access all open boxes
+      // Get all open boxes and clear them. We iterate over known disposable box
+      // names instead of trying to access all open boxes.
       final knownBoxNames = [
-        'user_profiles',
         'personal_events',
         'personal_events_metadata',
-        'bookmarks',
+        'hashtag_stats',
+        'people_lists_v1',
         'video_cache',
-        'secure_keys',
       ];
 
       for (final boxName in knownBoxNames) {
@@ -134,9 +133,17 @@ class CacheRecoveryService {
     'database',
   ];
 
-  static const List<List<String>> _durablePendingUploadsBoxSegments = [
-    ['openvine', 'pending_uploads.hive'],
-    ['openvine', 'pending_uploads.lock'],
+  static const List<String> _durableHiveBoxNames = [
+    'pending_uploads',
+    'notifications',
+    'push_notification_preferences_dirty',
+  ];
+
+  static List<List<String>> get _durableHiveBoxFileSegments => [
+    for (final boxName in _durableHiveBoxNames) ...[
+      ['openvine', '$boxName.hive'],
+      ['openvine', '$boxName.lock'],
+    ],
   ];
 
   /// Clear app support directory (NOT Documents - that's sandboxed on macOS),
@@ -149,14 +156,14 @@ class CacheRecoveryService {
         appSupportDir.path,
         ..._durableDatabaseDirSegments,
       ]);
-      final protectedPendingUploadsPaths = [
-        for (final segments in _durablePendingUploadsBoxSegments)
+      final protectedHiveBoxPaths = [
+        for (final segments in _durableHiveBoxFileSegments)
           p.joinAll([appSupportDir.path, ...segments]),
       ];
       return await deleteDirectoryContentsExcept(
         appSupportDir,
         protectedPath: protectedPath,
-        additionalProtectedPaths: protectedPendingUploadsPaths,
+        additionalProtectedPaths: protectedHiveBoxPaths,
       );
     } catch (e) {
       Log.warning(
--- a/mobile/lib/services/cache_recovery_service.dart
+++ b/mobile/lib/services/cache_recovery_service.dart
@@ -13,6 +13,13 @@ import 'package:unified_logger/unified_logger.dart';
 class CacheRecoveryService {
   static const String _logName = 'CacheRecoveryService';
 
+  static const Set<String> _disposableHiveBoxNames = {
+    'personal_events',
+    'personal_events_metadata',
+    'hashtag_stats',
+    'people_lists_v1',
+  };
+
   /// Clear all app caches and databases to recover from corruption
   /// This works on iOS devices, Android, and desktop platforms
   static Future<bool> clearAllCaches() async {
@@ -61,15 +68,9 @@ class CacheRecoveryService {
     try {
       // Get all open boxes and clear them. We iterate over known disposable box
       // names instead of trying to access all open boxes.
-      final knownBoxNames = [
-        'personal_events',
-        'personal_events_metadata',
-        'hashtag_stats',
-        'people_lists_v1',
-        'video_cache',
-      ];
+      final boxNames = _disposableHiveBoxNames.difference(_durableHiveBoxNames);
 
-      for (final boxName in knownBoxNames) {
+      for (final boxName in boxNames) {
         try {
           if (Hive.isBoxOpen(boxName)) {
             final box = Hive.box(boxName);
@@ -91,7 +92,7 @@ class CacheRecoveryService {
       }
 
       // Also try to delete known box files from disk
-      for (final boxName in knownBoxNames) {
+      for (final boxName in boxNames) {
         try {
           await Hive.deleteBoxFromDisk(boxName);
           cleared++;
@@ -117,6 +118,10 @@ class CacheRecoveryService {
 
   @visibleForTesting
   static Future<int> clearHiveBoxesForTesting() => _clearHiveBoxes();
+
+  @visibleForTesting
+  static Set<String> get disposableHiveBoxNamesForTesting =>
+      _disposableHiveBoxNames;
 
   /// Durable database directory under Application Support that must NEVER be
   /// deleted by cache clearing. It holds the live SQLite database
@@ -148,13 +153,16 @@ class CacheRecoveryService {
   /// initializer re-points Hive's home path there during startup — which is
   /// why [_clearAppSupportDirectory] has to protect them by path as well as
   /// skipping them above.
-  static const List<String> _durableHiveBoxNames = [
+  static const Set<String> _durableHiveBoxNames = {
     'pending_uploads',
     'notifications',
     'push_notification_preferences_dirty',
-  ];
+  };
 
-  static List<List<String>> get _durableHiveBoxFileSegments => [
+  @visibleForTesting
+  static Set<String> get durableHiveBoxNamesForTesting => _durableHiveBoxNames;
+
+  static final List<List<String>> _durableHiveBoxFileSegments = [
     for (final boxName in _durableHiveBoxNames) ...[
       ['openvine', '$boxName.hive'],
       ['openvine', '$boxName.lock'],
@@ -167,18 +175,10 @@ class CacheRecoveryService {
     try {
       final appSupportDir = await getApplicationSupportDirectory();
       if (!appSupportDir.existsSync()) return 0;
-      final protectedPath = p.joinAll([
-        appSupportDir.path,
-        ..._durableDatabaseDirSegments,
-      ]);
-      final protectedHiveBoxPaths = [
-        for (final segments in _durableHiveBoxFileSegments)
-          p.joinAll([appSupportDir.path, ...segments]),
-      ];
       return await deleteDirectoryContentsExcept(
         appSupportDir,
-        protectedPath: protectedPath,
-        additionalProtectedPaths: protectedHiveBoxPaths,
+        protectedPath: _durableDatabasePath(appSupportDir),
+        additionalProtectedPaths: _durableHiveBoxPaths(appSupportDir),
       );
     } catch (e) {
       Log.warning(
@@ -244,6 +244,16 @@ class CacheRecoveryService {
     }
     return cleared;
   }
+
+  static String _durableDatabasePath(Directory appSupportDir) => p.joinAll([
+    appSupportDir.path,
+    ..._durableDatabaseDirSegments,
+  ]);
+
+  static List<String> _durableHiveBoxPaths(Directory appSupportDir) => [
+    for (final segments in _durableHiveBoxFileSegments)
+      p.joinAll([appSupportDir.path, ...segments]),
+  ];
 
   static bool _pathExists(String path) =>
       FileSystemEntity.typeSync(path) != FileSystemEntityType.notFound;
@@ -317,26 +327,35 @@ class CacheRecoveryService {
   /// Best-effort: a directory that cannot be read contributes zero. Throws if
   /// the platform cannot resolve one of the directories at all.
   static Future<int> cacheSizeBytes() async {
-    final dirs = [
-      await getApplicationSupportDirectory(),
-      await getTemporaryDirectory(),
-      await getApplicationCacheDirectory(),
-    ];
+    final appSupportDir = await getApplicationSupportDirectory();
+    final tempDir = await getTemporaryDirectory();
+    final cacheDir = await getApplicationCacheDirectory();
 
     var totalSize = 0;
-    for (final dir in dirs) {
-      if (dir.existsSync()) {
-        totalSize += await _getDirectorySize(dir);
-      }
+    if (appSupportDir.existsSync()) {
+      totalSize += await _getDirectorySize(
+        appSupportDir,
+        excludedPaths: [
+          _durableDatabasePath(appSupportDir),
+          ..._durableHiveBoxPaths(appSupportDir),
+        ],
+      );
+    }
+    for (final dir in [tempDir, cacheDir]) {
+      if (dir.existsSync()) totalSize += await _getDirectorySize(dir);
     }
     return totalSize;
   }
 
   /// Calculate directory size recursively
-  static Future<int> _getDirectorySize(Directory directory) async {
+  static Future<int> _getDirectorySize(
+    Directory directory, {
+    List<String> excludedPaths = const [],
+  }) async {
     int size = 0;
     try {
       await for (final entity in directory.list(recursive: true)) {
+        if (_isExcluded(entity.path, excludedPaths)) continue;
         if (entity is File) {
           size += await entity.length();
         }
@@ -346,4 +365,10 @@ class CacheRecoveryService {
     }
     return size;
   }
+
+  static bool _isExcluded(String path, List<String> excludedPaths) =>
+      excludedPaths.any(
+        (excludedPath) =>
+            p.equals(path, excludedPath) || p.isWithin(excludedPath, path),
+      );
 }
--- a/mobile/lib/services/cache_recovery_service.dart
+++ b/mobile/lib/services/cache_recovery_service.dart
@@ -5,6 +5,7 @@ import 'dart:io';
 
 import 'package:hive_ce/hive.dart';
 import 'package:meta/meta.dart';
+import 'package:openvine/constants/hive_box_names.dart';
 import 'package:path/path.dart' as p;
 import 'package:path_provider/path_provider.dart';
 import 'package:unified_logger/unified_logger.dart';
@@ -14,10 +15,10 @@ class CacheRecoveryService {
   static const String _logName = 'CacheRecoveryService';
 
   static const Set<String> _disposableHiveBoxNames = {
-    'personal_events',
-    'personal_events_metadata',
-    'hashtag_stats',
-    'people_lists_v1',
+    HiveBoxNames.personalEvents,
+    HiveBoxNames.personalEventsMetadata,
+    HiveBoxNames.hashtagStats,
+    HiveBoxNames.peopleLists,
   };
 
   /// Clear all app caches and databases to recover from corruption
@@ -154,9 +155,9 @@ class CacheRecoveryService {
   /// why [_clearAppSupportDirectory] has to protect them by path as well as
   /// skipping them above.
   static const Set<String> _durableHiveBoxNames = {
-    'pending_uploads',
-    'notifications',
-    'push_notification_preferences_dirty',
+    HiveBoxNames.pendingUploads,
+    HiveBoxNames.notifications,
+    HiveBoxNames.pushNotificationPreferencesDirty,
   };
 
   @visibleForTesting
--- a/mobile/test/services/cache_recovery_service_test.dart
+++ b/mobile/test/services/cache_recovery_service_test.dart
@@ -36,15 +36,24 @@ void main() {
         await TestHelpers.cleanupHiveBox('notifications');
         await TestHelpers.cleanupHiveBox('push_notification_preferences_dirty');
         await TestHelpers.cleanupHiveBox('pending_uploads');
-        await TestHelpers.cleanupHiveBox('video_cache');
+        await TestHelpers.cleanupHiveBox('hashtag_stats');
       });
 
       tearDown(() async {
         await TestHelpers.cleanupHiveBox('notifications');
         await TestHelpers.cleanupHiveBox('push_notification_preferences_dirty');
         await TestHelpers.cleanupHiveBox('pending_uploads');
-        await TestHelpers.cleanupHiveBox('video_cache');
+        await TestHelpers.cleanupHiveBox('hashtag_stats');
         if (tmp.existsSync()) tmp.deleteSync(recursive: true);
+      });
+
+      test('keeps durable Hive boxes out of the disposable wipe set', () {
+        expect(
+          CacheRecoveryService.disposableHiveBoxNamesForTesting.intersection(
+            CacheRecoveryService.durableHiveBoxNamesForTesting,
+          ),
+          isEmpty,
+        );
       });
 
       test(
@@ -53,16 +62,16 @@ void main() {
           final pendingUploads = Hive.isBoxOpen('pending_uploads')
               ? Hive.box<PendingUpload>('pending_uploads')
               : await Hive.openBox<PendingUpload>('pending_uploads');
-          final videoCache = Hive.isBoxOpen('video_cache')
-              ? Hive.box('video_cache')
-              : await Hive.openBox('video_cache');
+          final hashtagStats = Hive.isBoxOpen('hashtag_stats')
+              ? Hive.box('hashtag_stats')
+              : await Hive.openBox('hashtag_stats');
           final upload = PendingUpload.create(
             localVideoPath: '/tmp/durable-upload.mp4',
             nostrPubkey:
                 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
           );
           await pendingUploads.put(upload.id, upload);
-          await videoCache.put('cache-id', 'cached video');
+          await hashtagStats.put('popular_hashtags', ['divine']);
 
           await CacheRecoveryService.clearHiveBoxesForTesting();
 
@@ -72,10 +81,10 @@ void main() {
             ).get(upload.id)?.localVideoPath,
             upload.localVideoPath,
           );
-          expect(Hive.isBoxOpen('video_cache'), isFalse);
-          final reopenedVideoCache = await Hive.openBox('video_cache');
-          addTearDown(reopenedVideoCache.close);
-          expect(reopenedVideoCache.get('cache-id'), isNull);
+          expect(Hive.isBoxOpen('hashtag_stats'), isFalse);
+          final reopenedHashtagStats = await Hive.openBox('hashtag_stats');
+          addTearDown(reopenedHashtagStats.close);
+          expect(reopenedHashtagStats.get('popular_hashtags'), isNull);
         },
       );
 
@@ -89,19 +98,16 @@ void main() {
               Hive.isBoxOpen('push_notification_preferences_dirty')
               ? Hive.box('push_notification_preferences_dirty')
               : await Hive.openBox('push_notification_preferences_dirty');
-          final videoCache = Hive.isBoxOpen('video_cache')
-              ? Hive.box('video_cache')
-              : await Hive.openBox('video_cache');
+          final hashtagStats = Hive.isBoxOpen('hashtag_stats')
+              ? Hive.box('hashtag_stats')
+              : await Hive.openBox('hashtag_stats');
           const preferences = NotificationPreferences(commentsEnabled: false);
           const dirtyPreferencesKey =
               'push_preferences_dirty_aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa';
           final storedPreferences = jsonEncode(preferences.toJson());
           await notifications.put('push_preferences', storedPreferences);
-          await dirtyNotifications.put(
-            dirtyPreferencesKey,
-            storedPreferences,
-          );
-          await videoCache.put('cache-id', 'cached video');
+          await dirtyNotifications.put(dirtyPreferencesKey, storedPreferences);
+          await hashtagStats.put('popular_hashtags', ['divine']);
 
           await CacheRecoveryService.clearHiveBoxesForTesting();
 
@@ -115,10 +121,10 @@ void main() {
             ).get(dirtyPreferencesKey),
             storedPreferences,
           );
-          expect(Hive.isBoxOpen('video_cache'), isFalse);
-          final reopenedVideoCache = await Hive.openBox('video_cache');
-          addTearDown(reopenedVideoCache.close);
-          expect(reopenedVideoCache.get('cache-id'), isNull);
+          expect(Hive.isBoxOpen('hashtag_stats'), isFalse);
+          final reopenedHashtagStats = await Hive.openBox('hashtag_stats');
+          addTearDown(reopenedHashtagStats.close);
+          expect(reopenedHashtagStats.get('popular_hashtags'), isNull);
         },
       );
     });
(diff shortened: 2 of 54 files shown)


rabble (changes requested):
The diagnosis is right and the change to `_clearHiveBoxes` is the correct first half — but the wipe survives one step later, so "Clear cache" from Settings still destroys push preferences after this PR. The new test can't see it because it only exercises the half that was fixed.

## The preferences file is still deleted, by step 2

`clearAllCaches` runs four steps (`mobile/lib/services/cache_recovery_service.dart:28-38`). This PR fixes step 1 (`_clearHiveBoxes`, no longer calls `box.clear()` / `deleteBoxFromDisk('notifications')`). Step 2 is `_clearAppSupportDirectory` (`:145-170`), which protects exactly two things:

- `{appSupport}/openvine/database` (`_durableDatabaseDirSegments`, `:133-136`)
- `{appSupport}/openvine/pending_uploads.hive` / `.lock` (`_durablePendingUploadsBoxSegments`, `:138-141`)

Everything else under Application Support is deleted by `deleteDirectoryContentsExcept` (`:182-225`). `notifications.hive` is not on that list.

**Where the box file actually lives.** `main.dart:1511` does `Hive.initFlutter()` (documents dir), but startup then re-points Hive's global home path: `main.dart:660-664` (`UploadManager` init) → `mobile/lib/services/upload/pending_upload_store.dart:75-77` → `mobile/lib/services/upload_initialization_helper.dart:146-148`, which calls `Hive.init({appSupport}/openvine)` (`_getAppStorageDir`, `:34-46`). `Hive.init` unconditionally assigns the process-global `homePath` (hive_ce `hive_impl.dart:74`), so every box opened *after* that — including `notifications`, which `HiveNotificationPreferencesStore.openBox` opens lazily (`mobile/lib/services/notification_preferences_service.dart:42`, `:193`) — is created under `{appSupport}/openvine/`. That is precisely why `pending_uploads` needed explicit file-level protection in the first place.

**Verified empirically** (temporary test against this branch's parent, `_clearAppSupportDirectory` is unchanged by the PR):

```dart
final notif = File(p.join(tmp.path, 'support/openvine/notifications.hive'))
  ..parent.createSync(recursive: true)
  ..writeAsStringSync('prefs');
await CacheRecoveryService.clearAllCaches();
expect(notif.existsSync(), isTrue); // FAILS — prefs box file was deleted
```

So after this PR, on the real user-facing path the box contents survive `box.clear()` and are then removed with the file. On restart the user is back to all-on defaults — the exact symptom #6919 describes.

Same class, same directory: `push_notification_preferences_dirty.hive` (`notification_preferences_service.dart:43`) holds preference edits that have not yet been published to the push service. It was never in `knownBoxNames`, so step 1 never touched it — and step 2 deletes it anyway. That's independent evidence that step 2, not step 1, is the wipe that matters here.

## The test can't catch it

`mobile/test/services/cache_recovery_service_test.dart:80-105` calls `CacheRecoveryService.clearHiveBoxesForTesting()` — step 1 in isolation. It will stay green for as long as the name is off the list, whether or not preferences actually survive a cache recovery. The sibling test at `:134` (`preserves durable pending uploads during clearAllCaches`) is the shape this needs: mock the path provider, write the file under `support/openvine/`, run `clearAllCaches()`, assert it is still there.

## Suggested shape

1. Generalise `_durablePendingUploadsBoxSegments` into a durable-box list and add `['openvine', 'notifications.hive']` + `['openvine', 'notifications.lock']` (and the dirty-prefs box files, if unsynced edits should also survive), with a comment saying why the box is durable.
2. Keep the `knownBoxNames` removal — it's correct and necessary; an open box would otherwise be cleared in memory before the file question ever arises.
3. Add a `clearAllCaches`-level test alongside the pending-uploads one. That test fails today and passes after (1), which is the property the current test lacks.

Nothing else in the diff is a concern — the box holds only `push_preferences` (`notification_preferences_service.dart:42-44`, no other writer in `mobile/lib`), so skipping it wholesale in the wipe list is the right call, and the doc-comment update on `_clearHiveBoxes` is accurate.

NotThatKindOfDrLiz:
Fixed the cache recovery gap from rabble review.

What changed:
- Added a shared durable Hive box policy for pending_uploads, notifications, and push_notification_preferences_dirty.
- Use that durable box list to protect the corresponding .hive and .lock files under Application Support during clearAllCaches().
- Added full clearAllCaches() regression coverage proving notification prefs and dirty push preference state survive while disposable cache files still clear.

Verification:
- dart format mobile/lib/services/cache_recovery_service.dart mobile/test/services/cache_recovery_service_test.dart
- flutter test test/services/cache_recovery_service_test.dart
- flutter analyze lib/services/cache_recovery_service.dart test/services/cache_recovery_service_test.dart
- Pre-push hook: analyzer clean, changed test file passed

mobile/lib/services/cache_recovery_service.dart
```diff
@@ -54,21 +54,19 @@ class CacheRecoveryService {
...
+  /// Clear disposable Hive cache boxes.
   static Future<int> _clearHiveBoxes() async {
     int cleared = 0;
 
     try {
-      // Get all open boxes and clear them
-      // Note: We iterate over known box names instead of trying to access all open boxes
+      // Get all open boxes and clear them. We iterate over known disposable box
+      // names instead of trying to access all open boxes.
       final knownBoxNames = [
```
realmeylisdev: The reset confirmation still tells people this wipes their notification settings. `settingsStorageRepairConfirmMessage` (`mobile/lib/l10n/app_en.arb:6500`) reads:

> This wipes notifications, cached profiles, bookmarks and temporary files. Your clips and drafts stay, but you'll be signed out and have to restart the app.

Ran the real flow on an iPhone 17 Pro Max simulator against this branch: turned Mentions off, Settings > General > Storage > Reset app data, and `Library/Application Support/openvine/notifications.hive` came through with `"mentionsEnabled":false` intact, toggle still off after a cold restart. So "wipes notifications" is now the opposite of what happens, on either reading of the word: the Hive prefs box is what this PR protects, and the notification inbox is a Drift table inside the already-protected `openvine/database` dir.

The other three clauses were already wrong before this PR, for the record: `user_profiles` is a Drift table in that same protected dir, bookmarks live in SharedPreferences (`Library/Preferences/co.openvine.app.plist`, which the wipe never touches, verified untouched afterwards), and I was still signed in as the same npub after the reset because the keys are in the keychain. "temporary files" is the only accurate clause left.

Not fixing it here since it is product copy that has to be mirrored into all 18 ARBs, but a destructive-action dialog naming the exact thing this PR makes durable seems worth catching in the same change.

mobile/lib/services/cache_recovery_service.dart
```diff
@@ -54,21 +54,19 @@ class CacheRecoveryService {
...
-        'notifications',
-        'user_profiles',
         'personal_events',
         'personal_events_metadata',
-        'bookmarks',
+        'hashtag_stats',
+        'people_lists_v1',
```
realmeylisdev: Worth a line in the PR description: this hunk also drops `user_profiles`, `bookmarks` and `secure_keys` and adds `hashtag_stats` and `people_lists_v1`, none of which the body mentions.

Both directions check out, so this is not an objection. The three removed names have no `Hive.openBox` anywhere in `mobile/lib` or `mobile/packages`, so they were dead entries that never matched a real box. The two added ones are real cache boxes whose files were already being deleted by `_clearAppSupportDirectory`, so the only change is that they now get closed before the file goes away instead of having it unlinked underneath an open box. That is strictly better. It just is not obvious from the description that the destructive list moved at all.

mobile/lib/services/cache_recovery_service.dart
```diff
@@ -135,9 +133,32 @@ class CacheRecoveryService {
...
+  /// initializer re-points Hive's home path there during startup — which is
+  /// why [_clearAppSupportDirectory] has to protect them by path as well as
+  /// skipping them above.
+  static const List<String> _durableHiveBoxNames = [
+    'pending_uploads',
+    'notifications',
+    'push_notification_preferences_dirty',
+  ];
+
+  static List<List<String>> get _durableHiveBoxFileSegments => [
```
realmeylisdev: Pre-existing, not a blocker, but it turned up while checking that these paths are the right ones.

Hive's home path is not stable across a run. `main.dart:1511` sets it to the documents dir via `initFlutter()`, then `upload_initialization_helper.dart:148` resets it to `{appSupport}/openvine` when the uploads box initialises, so where a box file lands depends on which side of that it opens. On the simulator `personal_events` is in `Documents/` this launch and was in `Library/Application Support/openvine/` on an earlier one, same box, two directories. Its sibling `personal_events_metadata` was in `openvine/` both times.

This PR holds either way, which is the nice thing about doing both halves: the `knownBoxNames` removal covers the documents case and this list covers the other. But the same flip means the prefs box could be created empty in the other directory on some launch, with no cache recovery involved at all. Worth its own issue rather than growing this one.

realmeylisdev (commented):
went through this end to end. ran the wipe on a sim against this branch: turned mentions off, hit reset app data, and notifications.hive came through with the setting intact, still off after a cold restart. mutation checked both new tests too, each one goes red without its own half of the fix.

pushed e782329f9 with the doc comment rabble asked for in point 1. one thing left inline: the reset confirmation still tells people it wipes notifications.

mobile/lib/services/cache_recovery_service.dart
realmeylisdev: Fixed in `a49db4062` rather than leaving it with you. New value:

> This clears cached feed data and temporary files. Your clips, drafts, settings and sign-in stay, but you'll need to restart the app afterwards.

Mirrored into all 21 non-English locales, `flutter gen-l10n` regenerated and committed, `arb_consistency_test` and `storage_management_view_test` green. Kept it as its own commit so it reverts alone if product wants different wording.

mobile/lib/services/cache_recovery_service.dart
realmeylisdev: Added to the PR description, covering both the file-level protection from the follow-up commit and the wipe-set changes in both directions.

mobile/lib/services/cache_recovery_service.dart
realmeylisdev: Filed as #6958 with the container listing, so nothing to carry here.

realmeylisdev (approved):
lgtm. ran the wipe on a sim, the prefs box survives and mentions was still off after a cold restart. pushed the doc comment rabble asked for, plus `a49db4062` for the reset dialog still claiming it wipes notifications, which is its own commit so revert freely if product wants different wording. the hive home-path oddity I hit while checking the paths is filed as #6958, nothing needed here. checks green.

mobile/lib/services/cache_recovery_service.dart
```diff
@@ -54,21 +54,19 @@ class CacheRecoveryService {
...
-        'notifications',
-        'user_profiles',
         'personal_events',
         'personal_events_metadata',
-        'bookmarks',
+        'hashtag_stats',
+        'people_lists_v1',
         'video_cache',
-        'secure_keys',
       ];
```
realmeylisdev: heads up that this commit also rewrote the rest of the list, not just the `notifications` removal. `user_profiles`, `bookmarks` and `secure_keys` are gone, `hashtag_stats` and `people_lists_v1` are new.

i checked and it's correct. the three dropped names have no `openBox` call anywhere in `mobile/lib` or `mobile/packages` (user_profiles moved to drift in #1881, the other two arrived with 255adf58a and never had an opener). `hashtag_stats` expires after an hour. `people_lists_v1` only writes the cache after the relay accepts the publish, so there is no local-only state to lose and `syncOwner` refills it.

nothing to change in the code. but it does mean the reset now clears two caches it didn't clear before, and neither the PR body nor the commit message mentions it. worth a line in the description so it isn't a surprise later.

mobile/lib/services/cache_recovery_service.dart
```diff
@@ -135,9 +133,32 @@ class CacheRecoveryService {
...
-  static const List<List<String>> _durablePendingUploadsBoxSegments = [
-    ['openvine', 'pending_uploads.hive'],
-    ['openvine', 'pending_uploads.lock'],
+  /// Hive boxes holding local-only state that cache clearing must preserve.
+  ///
+  /// `notifications` is named for the feature, not its contents: its only key
+  /// is `push_preferences`, the notification settings the user chose. The
+  /// notification inbox is a Drift table inside the protected database dir.
+  /// Wiping this box drops the user back to all-on defaults, which the next
+  /// toggle republishes over the choices they made (#6919).
```
realmeylisdev, line 148: this is the one i'd like sorted. `settingsStorageRepairConfirmMessage` still reads "This wipes notifications, cached profiles, bookmarks and temporary files." after this PR none of those three are accurate.

notifications: the inbox is the drift `Notifications` table, and `buildSharedDatabasePath` puts the db at `{appSupport}/openvine/database/divine_db.db`, which is already the protected dir. the hive box holding the prefs is preserved by this change. so nothing notification-shaped is removed.

cached profiles: drift `UserProfiles`, same protected db, since #1881.

bookmarks: `BookmarkService` persists through SharedPreferences (`_prefs.getString` / `setString`, bookmark_service.dart:853 and :878), which lands in `Library/Preferences/co.openvine.app.plist`. i ran the reset on the sim and that plist is untouched.

only the notifications clause is new here, the other two were already stale. but this is the sentence people read right before tapping Reset on a destructive action, so it should say what actually goes. it's 22 locales and it's product's copy, so i'm not pushing it in review.

mobile/lib/services/cache_recovery_service.dart
```diff
@@ -135,9 +133,32 @@ class CacheRecoveryService {
...
+  /// `push_notification_preferences_dirty` holds edits that have not reached
+  /// the push service yet plus the published-kinds schema marker, so losing it
+  /// strands an unsynced edit.
+  ///
+  /// Their files sit under `{appSupport}/openvine/` — the uploads-box
+  /// initializer re-points Hive's home path there during startup — which is
+  /// why [_clearAppSupportDirectory] has to protect them by path as well as
+  /// skipping them above.
```
realmeylisdev, line 156: not blocking, just flagging what this line rests on. "their files sit under {appSupport}/openvine/" is true today, but it's an ordering outcome rather than an invariant. `Hive.initFlutter()` sets home to Documents in the critical startup phase, and `UploadInitializationHelper` re-points it to `{appSupport}/openvine` in the standard phase, so a box lands wherever home happened to be when it was first opened.

live example on the sim right now: `personal_events.hive` is in `Documents/` while `personal_events_metadata.hive` is in `Application Support/openvine/`. those two are opened by back to back awaits in `PersonalEventCacheService.initialize`, and the re-point landed between them.

the notifications box opens late enough on a normal launch that it's fine. i could only get it into Documents by launching straight to `/notification-settings` with a route arg, and nothing deep-links there, so no user path hits it. nothing to change in this PR. probably worth its own issue though, since the path protection you're adding assumes a location nothing actually pins.

realmeylisdev (changes requested):
ran the reset both ways on an iphone sim. on main the prefs box gets deleted, on this branch it survives with my mentions=off still in it, so the fix does what it says, and both new tests go red against main's version of the service.

one thing i'd like sorted before merge: the reset confirmation sheet still tells people it wipes notifications, and after this it doesn't. details inline.

NotThatKindOfDrLiz:
Follow-up pushed in `f3df97314`.

What changed:
- Removed the dead `video_cache` Hive name from the disposable wipe policy.
- Lifted disposable/durable Hive box names into separate policy sets, with the wipe filtering durable names out and a regression test pinning the sets as disjoint.
- Repointed the disposable-cache test control to the real `hashtag_stats` Hive box.
- Updated `cacheSizeBytes()` so the repair sheet footprint excludes the same protected Application Support database/Hive paths that `clearAllCaches()` preserves.
- Updated the PR body to soften the `people_lists_v1`/Hive-location wording and reference #6958.

Verification:
- `dart format lib/services/cache_recovery_service.dart test/services/cache_recovery_service_test.dart`
- `flutter analyze lib/services/cache_recovery_service.dart test/services/cache_recovery_service_test.dart`
- `flutter test test/services/cache_recovery_service_test.dart`
- Pre-push hook: no merge conflicts with `main`, analyzer clean, `test/services/cache_recovery_service_test.dart` passed

I resolved the stale duplicate threads for the wipe-set description, confirmation copy, and Hive home-path caveat since they are now covered by the code/body updates or #6958.

realmeylisdev (approved):
re-reviewed at `6a7978e9f`, which adds `f3df97314` on top of what i last looked at. clearing my changes-requested.

the step-2 gap rabble found is closed, and i mutation-checked both halves rather than trusting the green run:

- putting `notifications` back into the disposable set → 1 red. worth noting the `.difference(_durableHiveBoxNames)` is doing real work here: it keeps the wipe correct and lets the intersection test be the thing that fails, so that mistake becomes a caught invariant instead of silent data loss.
- dropping `notifications` from the durable set → 2 red (the `clearAllCaches` test and the footprint test).
- reverting only the `cacheSizeBytes` exclusion → 1 red, exactly the footprint test.

on the wipe-set churn: the box inventory is 7 (`personal_events`, `personal_events_metadata`, `hashtag_stats`, `people_lists_v1`, `pending_uploads`, `notifications`, `push_notification_preferences_dirty`) and the two sets partition it exactly, so nothing is unclassified. the four names dropped are all dead — `user_profiles` is a drift table, `bookmarks` and `secure_keys` have no opener at all, and `video_cache` is never a hive box (the hits are a startup phase name, a prefs key and a media-cache key). the two added are real, and step 2 was already deleting both files before this PR, so the only change for them is closing the box before the file goes away.

`people_lists_v1` is safe to treat as disposable, a bit more firmly than the description says: `markDeleted` only runs after `publishEvent` returns `PublishSuccess` (`people_lists_repository_impl.dart:178-186`), so a deletion is always on the relay before the tombstone is written locally. losing tombstones can't resurrect a list.

also confirmed the durability argument holds end to end: preferences default to all-on, `loadPreferences` falls back to those defaults on an empty box, and nothing in the app ever reads preferences back from the push service — the service keeps its own copy in redis and filters on it with no client-facing read path. so a wiped box really does mean the settings screen silently disagrees with what the service is doing, until the next toggle republishes defaults over it.

`cacheSizeBytes` only feeds the repair confirmation sheet; the "cache in use" figure and the bug-report line come from `StorageManagementService`, so the exclusion doesn't move any other number.

checks green, all 15.

the only thing left is rabble's changes-requested, which i can't clear.

realmeylisdev (commented):
Ran this on an iPhone 17 Pro Max simulator against the PR build, start to finish: turned Mentions off, confirmed `notifications.hive` held `"mentionsEnabled":false`, ran Reset app data, and after the wipe plus a cold restart the box still had the value and the toggle was still off, still signed in. `cache/`, `people_lists_v1` and `personal_events_metadata` were gone, `database/` and both prefs boxes survived. The footprint line also dropped from 283 MB to 276 MB against the pre-PR build, which is the `cacheSizeBytes` exclusion showing up on device.

Two things worth recording.

The severity is a bit worse than #6919 describes. `markDirtyIfSchemaOutdated` seeds from `loadPreferences()`, which returns all-on when the box is empty, and it runs off session readiness via `pushNotificationSync` (`app_shell.dart:329`) rather than off a user edit. So a wiped box republishes all-on to the push service on the next launch with nobody touching a toggle. The push service keeps its own copy in Redis and there is no client-facing read path, so the user never gets it back. That is the argument for protecting both boxes rather than just `notifications`, which is what this does.

One suggestion inline, not a blocker.

Everything else checked out: the box inventory is 7 and the two sets partition it exactly, the four dropped names are genuinely dead, `clearCaches()` never touches Hive or Application Support, and `people_lists_v1` is safe to wipe because `markDeleted` only runs after `publishEvent` returns `PublishSuccess`.

mobile/lib/services/cache_recovery_service.dart
```diff
@@ -135,9 +138,35 @@ class CacheRecoveryService {
...
+  static const Set<String> _durableHiveBoxNames = {
+    'pending_uploads',
+    'notifications',
+    'push_notification_preferences_dirty',
+  };
```
realmeylisdev, line 161: Non-blocking, but this list is now the only thing standing between a durable box and deletion, and the two steps disagree on their default.

Step 1 deletes only what `_disposableHiveBoxNames` lists, so an unlisted box is safe there. Step 2 is the opposite: `deleteDirectoryContentsExcept` removes everything under Application Support except `openvine/database` and the six paths derived from this set. A box that is in neither set therefore gets its file deleted. Since `UploadInitializationHelper` re-points Hive's home to `{appSupport}/openvine` during startup, most boxes opened after that land exactly where step 2 sweeps — on the sim right now 5 of the 7 live there.

So the default for a newly added Hive box is "file gets deleted", and the only cross-set assertion in the test file is that the two sets do not overlap. Nothing checks that a known box name appears in one of them. Someone adding a durable box next quarter and not finding this file reintroduces #6919, silently.

A ratchet in the shape the repo already uses elsewhere would close it — grep `Hive.openBox` literals across `mobile/lib` and `mobile/packages` and fail if a name is in neither set. I did not push that myself since it is new CI surface and AGENTS.md puts that outside takeover. A comment here pointing the next person at both sets would be a cheaper version.

NotThatKindOfDrLiz: Fixed in `befde40f2`.

I added a shared `HiveBoxNames` constants home and wired the box owners plus `CacheRecoveryService` to it. The cache recovery test now asserts `disposable ∪ durable == HiveBoxNames.all` while keeping the disjointness assertion, so a newly added Hive box that is not classified fails in the existing unit test instead of relying on grep or new CI surface.

NotThatKindOfDrLiz:
Follow-up pushed in `befde40f2`.

What changed:
- Added shared `HiveBoxNames` constants for the seven known Hive boxes.
- Wired the storage owners and `CacheRecoveryService` to those constants instead of duplicating box-name strings.
- Strengthened `cache_recovery_service_test.dart` so the disposable and durable sets must partition `HiveBoxNames.all`; an unclassified new Hive box now fails the existing unit test.
- Rebasing removed the prior merge commit from the PR branch.

Verification:
- `flutter analyze lib/constants/hive_box_names.dart lib/services/cache_recovery_service.dart lib/services/notification_preferences_service.dart lib/services/upload_initialization_helper.dart lib/services/personal_event_cache_service.dart lib/services/hashtag_cache_service.dart lib/providers/repository_providers.dart test/services/cache_recovery_service_test.dart`
- `flutter test test/services/cache_recovery_service_test.dart`
- Pre-push hook: no merge conflicts with `main`, analyzer clean, generated files OK, changed tests passed (`cache_recovery_service_test.dart`, `notification_preferences_service_test.dart`, `personal_event_cache_service_test.dart`, `upload_initialization_helper_test.dart`).

@rabble your changes-requested review should now be stale: the step-2 durability gap was already covered, and this follow-up closes the remaining Hive-box classification ratchet. @realmeylisdev I replied inline to the non-blocking thread with the commit details.

mobile/test/services/cache_recovery_service_test.dart
```diff
@@ -31,45 +34,116 @@ void main() {
...
+          ...CacheRecoveryService.durableHiveBoxNamesForTesting,
+        };
+
+        expect(
+          CacheRecoveryService.disposableHiveBoxNamesForTesting.intersection(
+            CacheRecoveryService.durableHiveBoxNamesForTesting,
+          ),
+          isEmpty,
+        );
+        expect(classifiedHiveBoxNames, unorderedEquals(HiveBoxNames.all));
```
realmeylisdev, line 67: **should-fix.** This pins disposable + durable against `HiveBoxNames.all`, but all three collections are hand-maintained in the same two files and nothing ties them back to the real `Hive.openBox` call sites. So it catches a *known* box being misclassified, not a *new* box being missed.

I checked rather than assuming. Dropped a service into `lib/services/` that opens `draft_notes_v1`, left `HiveBoxNames` untouched, ran this test:

```dart
class DraftNotesProbeService {
  static const String _boxName = 'draft_notes_v1';
  Future<Box<dynamic>> open() => Hive.openBox<dynamic>(_boxName);
}
```

```
00:00 +1: All tests passed!
```

Green. And `_clearAppSupportDirectory` then deletes `draft_notes_v1.hive`, because it isn't in `_durableHiveBoxFileSegments`. That is #6919 again, on a fresh box.

This matters because the follow-up comment tells rabble "an unclassified new Hive box now fails the existing unit test". It only does if whoever adds the box also remembers to put it in `HiveBoxNames.all`, and forgetting to classify a box is precisely the step that got missed for `notifications` in the first place. As written the ratchet guards reclassification, not arrival.

Cheapest way to make the claim true: have this test scan `lib/` and `packages/` for `Hive.openBox` / `openLazyBox` name arguments and assert each one resolves to a member of `HiveBoxNames.all`. Then `all` can't drift from the code and the existing partition check covers the rest.

realmeylisdev (changes requested):
re-checked at `609c893`. ran it end to end on an iPhone 17 Pro Max sim against this build: turned mentions off, hit reset app data, and `notifications.hive` came through with `mentionsEnabled:false` still in it, toggle still off after a cold restart. the constants swap doesn't rename anything either, the new build picked up the box files the old build had written. mutation checked all four guards and each one goes red on its own.

one thing i'd like sorted before merge: the new classification test doesn't catch a brand new hive box, which is the case it's being credited with in the thread above. details inline.

mobile/test/services/cache_recovery_service_test.dart
realmeylisdev: pushed `5032bf166` for this. `test/constants/hive_box_names_test.dart` scans lib/ and packages/ and requires every `Hive.openBox` argument to resolve to a `HiveBoxNames` member, plus every declared member to be in `.all`.

checked it the same way as before: dropped a throwaway service opening `draft_notes_v1` into lib/ and the old test stayed green, the new one goes red. then routed that same box through `HiveBoxNames` without classifying it and the partition test here picks it up, so arrival and reclassification are both covered now. deleted the probe after.

realmeylisdev (approved):
ran the whole thing on an iphone 17 pro max sim at `5032bf166`. mentions off, reset app data, cold restart, and `notifications.hive` still had `mentionsEnabled:false` in it with the toggle still off. `people_lists_v1`, `personal_events*` and `openvine/cache` went, `database/` and both prefs boxes stayed. no hive errors in the log after the wipe.

pushed `5032bf166` for the classification ratchet i'd flagged, it scans the real `Hive.openBox` call sites now instead of comparing hand-written lists. clearing my changes-requested. rabble's is still open and i can't clear that one.

NotThatKindOfDrLiz:
Follow-up pushed in `ae054dc92`.

What changed:
- Broadened the Hive box ratchet from `Hive.openBox`/`Hive.openLazyBox` only to any `.openBox`/`.openLazyBox` receiver, so injected `HiveInterface` shapes are caught too.
- Added an explicit Probe B-style fixture proving `_hive.openBox<dynamic>('draft_notes_v1')` is scanned.
- Rebasing onto current `origin/main` removed the merge-main commit from the PR branch and cleared the stale-base issue.

Verification:
- `flutter analyze test/constants/hive_box_names_test.dart`
- `flutter test test/constants/hive_box_names_test.dart test/services/cache_recovery_service_test.dart` (13/13)
- Pre-push hook: no merge conflicts with `main`, analyzer clean, generated files OK, changed tests passed (`hive_box_names_test.dart`, `cache_recovery_service_test.dart`, `notification_preferences_service_test.dart`, `personal_event_cache_service_test.dart`, `upload_initialization_helper_test.dart`).

@rabble quick nudge for re-review when you have a minute. This should close the remaining process + ratchet concerns.

NotThatKindOfDrLiz:
@rabble could you re-review when you have a minute? Your 2026-08-09 changes-requested review should be stale at current head `ae054dc92930a492c54838d1afc03833b30e5628`.

Mapping your three asks to the current branch:

- Step-2 durability gap: fixed by `723272e4a` / `d6297cad2`. `CacheRecoveryService` now has `_durableHiveBoxNames` for `pending_uploads`, `notifications`, and `push_notification_preferences_dirty`, and `_durableHiveBoxFileSegments` protects each `.hive` / `.lock` file under `{appSupport}/openvine/` during `clearAllCaches()`.
- Keep the Hive box-name wipe fix: still in place. Disposable boxes are limited to `personal_events`, `personal_events_metadata`, `hashtag_stats`, and `people_lists_v1`; durable boxes are skipped in memory and protected on disk.
- Full `clearAllCaches()` regression: added in `cache_recovery_service_test.dart` as `preserves durable notification preferences during clearAllCaches`, alongside the existing pending-uploads shape.

Follow-up ratchets are also in place: `HiveBoxNames` centralizes the seven known boxes, the cache-recovery test asserts disposable and durable sets partition `HiveBoxNames.all`, and `hive_box_names_test.dart` scans real `.openBox` / `.openLazyBox` call sites so a new Hive box has to route through the shared constants before the wipe policy can pass.

One product-copy note before merge: `2f307419b` updates the Reset app data confirmation copy across l10n because the old message now falsely says notifications/profile/bookmark state and sign-in are wiped. I think the new wording is directionally right, but it is isolated in its own commit if you want different product wording.

Checks are green: GitHub reports 16/16 successful at the current head.

NotThatKindOfDrLiz:
@rabble please re-review when you have a minute. I rebased this onto current `main` and pushed new head `5dfe33136b794507f715309433676b5df4c30df0`.

Your 2026-08-09 changes-requested review is stale at this head:

- Step-2 durability gap: still fixed. The durable Hive policy protects `pending_uploads`, `notifications`, and `push_notification_preferences_dirty` in step 1, and preserves each `.hive` / `.lock` file under `{appSupport}/openvine/` during `clearAllCaches()`.
- Keep the Hive box-name wipe fix: still in place. Disposable boxes remain limited to `personal_events`, `personal_events_metadata`, `hashtag_stats`, and `people_lists_v1`; durable boxes are skipped in memory and protected on disk.
- Full `clearAllCaches()` regression: still covered by `preserves durable notification preferences during clearAllCaches`, alongside the pending-uploads coverage.

Fresh verification after the rebase:

- `flutter test test/constants/hive_box_names_test.dart test/services/cache_recovery_service_test.dart` — 13/13 passed
- Pre-push hook — no merge conflicts with main, analyzer clean, generated files OK, changed tests passed (`hive_box_names_test.dart`, `cache_recovery_service_test.dart`, `notification_preferences_service_test.dart`, `personal_event_cache_service_test.dart`, `upload_initialization_helper_test.dart`)

GitHub checks are running on the new head now.

rabble (commented):
Re-reviewed. **Blocking item addressed**, plus realmeylisdev's copy finding. CI green (15/15) on `5dfe33136`.

**The step-2 wipe is closed.** `_durableHiveBoxNames` now carries `notifications` and `pushNotificationPreferencesDirty` alongside `pendingUploads`, and `_durableHiveBoxFileSegments` derives the `.hive`/`.lock` paths from that one set — so step 1's skip list and step 2's path protection can no longer disagree, which was the exact shape of the original bug. Including the dirty box is the right call: it holds edits that haven't reached the push service, so losing it strands an unsynced change.

`preserves durable notification preferences during clearAllCaches` is the test the original PR was missing — it goes through the real `clearAllCaches()` rather than `clearHiveBoxesForTesting()`, so it fails against the previous revision and can actually catch a regression in the file-deletion half.

The doc comment now explains *why* the box is durable ("named for the feature, not its contents: its only key is `push_preferences`") and why the files land under `{appSupport}/openvine/`. That second part matters, because as realmeylisdev notes inline it's an ordering outcome of `Hive.initFlutter()` then `UploadInitializationHelper` re-pointing home, not an invariant — worth having written down.

**Copy accuracy — I checked this independently and it's done properly.** `settingsStorageRepairConfirmMessage` went from "This wipes notifications, cached profiles, bookmarks and temporary files… you'll be signed out" to "This clears cached feed data and temporary files. Your clips, drafts, settings and sign-in stay". Critically, **all 22 `app_*.arb` locales were updated with real translations**, not English pasted into every file:

> ja: 「キャッシュ済みのフィードデータと一時ファイルを消すよ。クリップ、下書き、設定、ログインは残るけど…」

That's the standard the ARB workflow is meant to hit and rarely does. Changing an existing English value is exactly the case that silently drifts every locale, and this didn't.

**One open question I can't settle from here.** The new copy promises "sign-in stay". That rests on `secure_keys` having been a dead entry with no `Hive.openBox` anywhere — which realmeylisdev verified by grep — plus credentials living in the keychain rather than Application Support. They ran the real flow on a simulator; I'm taking the sign-in half on their word rather than re-deriving it. If that's wrong the copy now under-promises destruction in 22 languages, which is the worse direction to be wrong in.

rabble (approved):

NotThatKindOfDrLiz:
## Re-Review Summary — Ready for Merge

All three of **@rabble**'s CHANGES_REQUESTED items are now satisfied in current commits:

### 1. Durable boxes must be classified + excluded from both wipes
**Status: FIXED** — Commits `e5b7d39c2`, `b3edd5d59`, `ddc949f7c`
- Durable set defined: `pendingUploads`, `notifications`, `pushNotificationPreferencesDirty`
- Step 1 (Hive box clearing) excludes them via `.difference(_durableHiveBoxNames)`
- Step 2 (Application Support cleanup) protects `.hive` and `.lock` files under `{appSupport}/openvine/`

### 2. Central registry of all Hive boxes + validation test
**Status: FIXED** — Commits `8cee6e45`, `8b38f7c8`
- `mobile/lib/constants/hive_box_names.dart`: 7 box names + static `all` set + invariant docs
- `mobile/test/constants/hive_box_names_test.dart`: 
  - Partition invariant verified (disposable ∪ durable == all, disjoint)
  - Call-site scanning to ensure every `openBox`/`openLazyBox` uses `HiveBoxNames.*`

### 3. User-facing copy must reflect actual behavior
**Status: FIXED** — Commit `2f307419b` (all 22 locale ARBs)
- Old: "wipes notifications, cached profiles, bookmarks"
- New: "clears cached feed data and temporary files. Your clips, drafts, settings and sign-in stay"
- **NOTE:** See concern below about copy change scope.

---

## Technical Assessment

**Code quality:** ✅ Sound. Dual protection ensures robustness; partitioning is explicit and testable.  
**Test coverage:** ✅ Expanded from 4 → 10 tests (16 CI checks all green).  
**Architecture:** ✅ Follows established patterns (centralizes classification, defers to service tests for box ownership).

---

## Moderate Concern — Requires Clarification

**Commit `2f307419b` (l10n): Copy rewritten in 22 locales without product sign-off**

The confirmation message was rewritten end-to-end:
- Added precision ("cached feed data" vs "notifications, cached profiles, bookmarks")
- Removed mention of notifications + bookmarks

**Question:** Should a re-review approve this scope, or does it need product (design/Rabble) sign-off first? The change is defensible on accuracy, but involves 22 files + translations. My recommendation: **either get explicit sign-off on the new copy, or revert the l10n commit and keep it as a separate "improve settings copy" issue.**

---

## Low-Severity Concern

**LocalPeopleListsCache (people_lists_v1 Hive box):** Marked as disposable despite being a cache of social state. If the box closes unexpectedly (OOM during wipe, mid-fetch), the service continues reading a stale `_boxFuture`. Not a blocking issue (affects a cache-miss path), but a potential footgun for future refactors.

---

## Optional Nits (No block)

1. **Redundant `.difference()` call** in `cacheSizeBytes()` when the file segment list already excludes durable paths.
2. **Defensive improvement** to `hive_box_names_test.dart`: Add probe for empty regex matches to catch typos in the pattern.

---

## Ready for Merge

- ✅ All review items satisfied
- ✅ CI green (16/16 checks)
- ✅ No functional errors
- ⚠️ **Awaiting:** rabble's clarification on l10n copy scope (sign-off vs. revert)

No changes made. Waiting on direction for l10n handling before final handoff.

---

*Re-reviewed by Claude Code*

NotThatKindOfDrLiz:
## L10n Copy Decision — Needs Your Direction

### Issue

Commit `2f307419b` (which your commit message itself flags as "Copy change, so product may want to reword it") updates the `settingsStorageRepairConfirmMessage` across 22 locales — from the original: 

> "This wipes notifications, cached profiles, bookmarks and temporary files. Your clips and drafts stay, but you'll be signed out and have to restart the app."

to:

> "This clears cached feed data and temporary files. Your clips, drafts, settings and sign-in stay, but you'll need to restart the app afterwards."

**The new copy is more accurate** (cached profiles actually stay, notifications stay, key stays). But it is a **messaging change** — a product decision about what users should understand happens during the reset, not a bug fix.

### Your Options

**Option A: Get explicit product sign-off on the new copy**
- If Rabble or design approves this wording, the current PR is ready to merge as-is
- Clean, single PR with both the fix + the copy improvement
- Takes a synchronous decision

**Option B: Revert the l10n commit and defer it**
- Keep the core cache-recovery bug fix (all the Hive-protection logic)
- Revert just the 22 ARB file changes + regenerated l10n (45 files total)
- File a separate issue to improve the settings confirmation copy
- Lets this PR merge now while product reviews messaging separately
- Clean, since the commit is self-contained and reverts without conflicts

### My Assessment

The technical fix is complete and sound. The l10n copy is a separate product decision. **Recommend Option B** — land the bug fix now, improve the copy in a follow-up — since rabble can't clarify messaging scope in a re-review.

I've prepared the revert commit locally (`34c88bfb74`) if you want me to apply it. Just let me know which direction to take.

---

*Ready to take either path — just need your sign-off on which one.*

Issue #6918, opened 2026-08-09
chore(dependencies): replace app-device-integrity override


NotThatKindOfDrLiz:
## Goal

Promote Divine's App Attest implementation from a local override of the abandoned `app_device_integrity` package into a first-class, maintained workspace package.

## Context

Issue #3363 documents why the current override was initially allowed for release. Since then, Divine's fork has grown into an account-scoped App Attest implementation with a Divine-specific `keyScope` API, pending-key recovery, serialized provisioning, and native watchdogs. Returning to the hosted package is no longer a viable dependency path.

Divine only calls this plugin for iOS publish-time App Attest. Android proof generation uses the app's separate hardware-attestation implementation, so the fork's unused Play Integrity platform can be retired rather than carried into the maintained package.

## Acceptance Criteria

- The maintained App Attest implementation is a first-class Divine workspace package under `mobile/packages/`.
- The app depends on that package without an `app_device_integrity` dependency override.
- `mobile/overrides/app_device_integrity-1.1.0/` and both corresponding dependency-provenance baseline entries are removed.
- The package supports iOS only; the unused Play Integrity implementation and Android dependency are removed.
- The package preserves account-scoped App Attest behavior, including `keyScope`, pending-key recovery, provisioning serialization, and watchdog behavior.
- Dart method-channel behavior is covered by package tests, and package CI runs formatting, analysis, and tests.
- An actual iOS build verifies podspec discovery and Flutter plugin registration.

## Release note

This changes native plugin registration and therefore requires a store build; it cannot ship as a Shorebird patch.

Refs #3363.

Issue #3363, opened 2026-04-25
chore(security): pin or remove git branch and local dependency overrides before release


rabble:
Source audit context: fresh `origin/main` worktree `.worktrees/audit-codebase-2026-04-24` at commit `33f6e026a`.
This issue was generated from the April 2026 codebase audit and includes file/line evidence from that snapshot.

## Severity
Medium

## Problem
The production dependency graph includes git refs and local path overrides. One git dependency points at a mutable branch; local overrides bypass registry provenance.

## Evidence
- mobile/pubspec.yaml:115-119 pulls c2pa_flutter from GitHub at ref 0.0.3.
- mobile/pubspec.yaml:265-272 uses media_kit_video from https://github.com/divinevideo/media-kit at branch fix/texture-rotation.
- mobile/packages/pooled_video_player/pubspec.yaml:24-27 repeats the same media_kit_video git branch dependency.
- mobile/pubspec.yaml:281-287 has dependency_overrides to local paths for cryptography_flutter and app_device_integrity.
- mobile/pubspec.lock:84-90 locks app_device_integrity to local path override.

## Impact
Mutable branches and local overrides weaken supply-chain reproducibility and make release builds depend on unreviewed local state or a branch that can move.

## Recommended fix
Replace branch refs with immutable commit SHAs or published packages, document why each override exists, and add CI guardrails that fail release builds on mutable git refs or unexpected path overrides.

## Acceptance criteria
- [ ] No release dependency uses a mutable git branch.
- [ ] Any local override has an owner, expiry/removal issue, and CI allowlist.
- [ ] CI checks pubspec.yaml/lockfiles for mutable git refs and path overrides.

## Related issues
No exact local issue found; comments reference upstream media-kit/media-kit#627.

NotThatKindOfDrLiz:
Part of #5179.

hm21:
Status on `main` — AC 1 is done, AC 2 and 3 are not. Keeping this open, but the scope should be trimmed.

[...]

Issue #3638, opened 2026-04-29
chore(deps): remove device_info_plus override (v10 to v13)


omartinma:
**Problem**: `device_info_plus` is pinned to v10.1.2 via a dependency override while the latest is v13.1.0.

**Evidence**: `pubspec.yaml` contains a `dependency_overrides` entry pinning `device_info_plus` to `10.1.2`. The latest stable release is `13.1.0`, three major versions ahead. The override was likely added to resolve a transitive version conflict, but it now prevents the package (and its platform-specific implementations) from receiving bug fixes and security patches.

**Impact**: Medium. Three major versions behind means missing platform compatibility fixes (especially for new iOS/Android versions), potential deprecation of APIs the app relies on, and increasing migration difficulty the longer the pin stays.

**Effort**: Medium. Remove the override, resolve any version conflicts surfaced by `flutter pub get`, and test on both iOS and Android. Major version bumps may include breaking API changes.

**Source**: [Code Review — Dependencies](docs/code-review/issues-dependencies.md)

NotThatKindOfDrLiz:
Kept open while its seven siblings were replaced by #8345.

This is not a routine version bump: `mobile/pubspec.yaml` carries a live `TODO(#3638)` holding `device_info_plus` at 10.x until the app can absorb the transitive 11.x platform-interface changes, and the override is tracked by `check_dependency_provenance.sh`. Closing this would orphan that code comment.

Also removed the `needs-retest` label — there is nothing to retest here; the work simply has not been done.

PR #6923, opened 2026-08-09
chore(dependencies): remove cryptography_flutter override

NotThatKindOfDrLiz:
## Summary

- remove the unused cryptography_flutter dependency from nostr_sdk and delete the local stub override
- drop the redundant rxdart dependency override after flutter pub get resolved cleanly without it
- document the remaining app_device_integrity fork, including owner, local divergence, and removal condition
- file #6918 as the removal-condition follow-up for app_device_integrity

Refs #3363.
Refs #6918.

## Verification

- flutter pub get
- mise exec -- flutter analyze lib test integration_test
- flutter test from mobile/packages/nostr_sdk
- flutter build ios --debug --no-codesign

Android debug build was attempted, but local toolchain verification is blocked before compilation because this shell has no Java runtime and flutter doctor reports the Android cmdline-tools component missing.


rabble (approved):
Approving. I went after the two things the deletion count invites — a load-bearing override being dropped, and unrelated lockfile churn — and neither is present.

## The `cryptography_flutter` removal is dead-code removal, not a functional change

The vendored override was a hollow shim. Its barrel (`mobile/overrides/cryptography_flutter-2.3.2/lib/cryptography_flutter.dart`) exports exactly one thing:

```dart
export 'cryptography_flutter_web.dart'
    if (dart.library.html) 'cryptography_flutter_web.dart';
```

and that file's entire contents is a `registerWith` that comments `// No-op for web`. The iOS/macOS/Android/Linux plugin classes are the same shape — `SwiftCryptographyFlutterPlugin.register` is literally `// No-op implementation - cryptography is handled in Dart`. There is no `FlutterCryptography` class in the vendored copy at all, so there was nothing to call and nothing that could have swapped `Cryptography.instance` to a native backend.

Confirming that from the consumer side: `git grep` for `cryptography_flutter` / `FlutterCryptography` across `origin/main` hits only the vendored package itself, three pubspec/doc files, and two markdown files. Zero Dart imports in `mobile/lib` or any package. Meanwhile `nostr_sdk` continues to import the pure-Dart `package:cryptography` in `lib/nip17/file_encryption.dart:6`, `lib/nip44/nip44_v2.dart:7`, and `lib/nip49/nip49.dart:9` — untouched. So NIP-17/44/49 crypto paths are byte-for-byte unchanged; the only thing leaving is a no-op plugin registration.

## The `rxdart` override removal is also safe

`rxdart: ^0.28.0` is already a direct dependency at `mobile/pubspec.yaml:118`, and all eleven repository packages that use it (`videos_repository`, `follow_repository`, `likes_repository`, …) declare the identical `^0.28.0`. Nothing constrains it below 0.28, so the override was redundant — and the lockfile confirms it: `rxdart` is `dependency: "direct main"` at `version: "0.28.0"` with the same `sha256` on both sides of the diff. Removing the override changed nothing about resolution.

## The lockfile diff is exactly what it should be

`mobile/pubspec.lock` has one hunk: the seven-line `cryptography_flutter` entry (`dependency: "direct overridden"`, `source: path`) is gone. No `sha256` rewrites, no version bumps, no re-ordering — none of the "different pub resolver run" churn `.claude/rules/self_review_checklist.md` warns about. That's the correct minimal lock delta for dropping the last dependant on a path override.

Also checked: `mobile/ios/Podfile.lock` isn't tracked (only `Podfile` is), so there's no stale pod lock left pointing at the removed podspec, and `mobile/overrides/*` is already in the `app` changes filter at `.github/workflows/mobile_ci.yaml:75`, so deleting files there still triggers the full app CI suite rather than being treated as docs-only.

## Temporary-code rule

Both surviving overrides now carry the `TODO(#issue)` form `.claude/rules/code_style.md` asks for, and all three referenced issues are real and open:

- #3638 — "chore(deps): remove device_info_plus override (v10 to v13)"
- #6918 — "chore(dependencies): replace app-device-integrity override"
- #3363 — "chore(security): pin or remove git branch and local dependency overrides before release"

`mobile/overrides/app_device_integrity-1.1.0/OVERRIDE.md` is a good template for the next one: owner, the specific upstream defect (throwing `ActivityAware` detach stubs), the enumerated local changes, and an explicit removal condition. Nothing to add.

## Non-blocking notes

**1. The PR title undersells the change.** "chore(dependencies): document release overrides" reads as a docs PR. It also *removes a dependency* from `nostr_sdk` and deletes a Flutter plugin from the app's resolution. That's the right call on the merits, but an approver skimming the title before a release cut would not expect it — worth saying so in the title or at least the description.

**2. Stale reference left behind.** `wingspan/plans/2026-03-07-nip49-ncryptsec1-import.md:150` still says "The `cryptography_flutter` package offloads [scrypt] to a platform thread." That was never true of the vendored stub, and it's now definitively false. Either fix the line or note that the plan needs a different approach for the log_n=18 cost.

**3. Rewriting shipped changelog history.** `mobile/packages/nostr_sdk/CHANGELOG.md:32` edits an already-published 0.1.0 entry from "Uses official cryptography_flutter package" to "Uses hosted cryptography packages". The new wording is more accurate, so I'm not going to argue it — just flagging that editing a released changelog entry is the kind of thing worth being deliberate about.

## Release safety: land this **after** the 1.0.19 submission

It doesn't need to wait for correctness reasons — I'm satisfied the functional delta is zero. But it does perturb the plugin set that gets registered natively and the pod set that gets installed, for zero functional gain in 1.0.19. There's no upside to taking even provably-inert dependency churn into an archive that's about to go to Apple, and if anything *did* surprise us, it would surface as a native build or startup failure at the worst possible moment.

Merge it the day the submission is out the door.

NotThatKindOfDrLiz:
Pushed follow-up commit 47ba43ac0 after rebasing onto current origin/main.\n\nChanges:\n- corrected OVERRIDE.md so app_device_integrity is only attributed to IosDeviceAttestationService\n- reworded the Android rationale around plugin registration and ActivityAware teardown callbacks\n- removed stale nostr_sdk release-checklist bullets\n- marked the historical cryptography_flutter audit item fixed and corrected the stale wingspan scrypt note\n- retitled the PR to mention removing the cryptography_flutter override\n\nVerification:\n- git diff --check\n- flutter pub get\n- pre-push analyzer: passed\n- flutter build apk --debug: still blocked before compilation because this shell has no Java Runtime

PR #1928, opened 2026-03-02
feat: E2E integration test infrastructure with Patrol and Docker stack

dcadenas:
## Summary

Developer-local integration test infrastructure using [Patrol](https://patrol.leancode.co/) for native UI automation against a full Docker backend stack. Patrol wraps Flutter's `integration_test` with the ability to handle permission dialogs, system back button, notifications, and share sheets — things that pure Flutter test code can't reach.

### What's included

- **`local_stack/`** — Docker Compose with Keycast, FunnelCake Relay/API, Blossom, Postgres, Redis, ClickHouse. Health checks, seed container (100 test videos + kind 0 profiles), 15s token expiry for refresh testing.
- **`AppEnvironment.local`** — New env pointing the app at the Docker stack via `10.0.2.2`. All indexer relay queries route to local funnelcake, so tests run fully offline. No behavior change for existing environments.
- **Patrol migration** — All integration tests converted from `testWidgets`/`integration_test` to `patrolTest`. Camera tests use `$.native.grantPermissionWhenInUse()` instead of ADB workarounds. Android configured with `PatrolJUnitRunner` and test orchestrator.
- **5 auth journey tests** (`mobile/integration_test/auth/`):
  - `auth_journey_test` — register → verify → forgot password → reset → login → explore feed → search → edit profile → settings
  - `user_journey_test` — register → verify → publish video → explore tabs → navigation
  - `cross_user_verify_test` — deep link email verification while logged in as different user (skipped — found a cubit stale state bug, tracked separately)
  - `session_expiry_test` — expire JWT + consume DB refresh tokens → expired session UI
  - `token_refresh_test` — register → wait for expiry → verify refresh rotation → consume all → expired fallback
- **New Patrol-native tests** — camera permission denied flow, app backgrounding, notification tap, share sheet
- **Log profiler** — `local_stack/profile.sh` captures Docker logs + Android logcat (with UTC timestamps) during test runs, merges into time-sorted JSONL for correlating client actions with server behavior.
- **Single command** — `mise run e2e_test` starts Docker stack, runs Patrol tests, captures merged log timeline.

### Prod code changes

All prod changes add the `LOCAL` environment path — no behavior change for poc/staging/test/production:

| File | Change |
|------|--------|
| `environment_config.dart` | `AppEnvironment.local` enum + relay/API/blossom URLs |
| `app_providers.dart` | Pass `env.indexerRelays` to services; local OAuth config |
[...]

PR #6925, opened 2026-08-09
build(mobile): upgrade patrol to 4.8.0 and pin patrol_cli 4.6.1

### LINKED ISSUE #6930: fix(testing): patrol/patrol_cli pairing is unpinned and the local E2E runner is broken on macOS
## Problem

The E2E suite cannot be run by anyone setting this repo up from scratch, and cannot be run at all on macOS. Five independent defects stack up; each one alone is fatal to a default `mise run e2e_test`.

**1. `patrol` and `patrol_cli` are a runtime-enforced matched pair, and only one half is pinned.**
`mobile/pubspec.yaml` pins `patrol: ^4.2.0`. Nothing pins `patrol_cli` anywhere — not in `pubspec.yaml`, not in `local_stack/`, not in `mobile/mise.toml`. `local_stack/profile.sh` simply invokes whatever `patrol` binary is on `PATH`.

`patrol_cli` enforces the pairing itself. `patrol_cli/lib/src/compatibility_checker/version_compatibility.dart` hardcodes a `versionCompatibilityList` whose top entry maps `patrol_cli 4.5.0+` ↔ `patrol 4.7.0+`, and `patrol test` runs that check with `--check-compatibility` defaulting to **on**, aborting via `throwToolExit` on a mismatch.

A fresh machine gets whatever CLI is current — 4.6.1 today — which **rejects** the pinned `patrol 4.2.0` before a single test runs.

**2. `profile.sh` resolves `patrol` from a hardcoded `$HOME/.pub-cache/bin`.**
That directory does not exist when `PUB_CACHE` is relocated, so the `PATH` prepend is a silent no-op and the script has no control over which CLI it runs. `mobile/mise.toml` already does this correctly for `very_good_cli` (`PUB_CACHE_BIN="${PUB_CACHE:-$HOME/.pub-cache}/bin"`); `profile.sh` does not.

**3. `profile.sh` dies under macOS bash 3.2 on every default run.**
The script runs `set -euo pipefail`. macOS ships only bash 3.2, where expanding an empty array as `"${a[@]}"` trips `set -u`. `PATROL_EXTRA_ARGS` is empty unless `PATROL_NO_UNINSTALL=true`, so the default path fails with:

```
local_stack/profile.sh: line NNN: PATROL_EXTRA_ARGS[@]: unbound variable
```

after the Docker log capture and logcat capture have already started. Only runs that happened to set `PATROL_NO_UNINSTALL` — which makes the array non-empty — ever got through. Linux/CI runners ship bash 5 and are unaffected, which is why this survived.

**4. `profile.sh` never passes the adb device it just detected.**
The script reads `DEVICE` from `adb devices`, prints it, then omits it from the `patrol test` invocation. With more than one target attached — an iOS simulator, macOS and Chrome all count as devices — patrol stops and prompts for a selection. Under a non-interactive runner it loops on invalid input until killed.

**5. A `/` in a Dart test name silently drops the test and still reports green.**
patrol names each JUnit case `MainActivityTest#runDartTest[<dart test name>]`, and the AndroidX test orchestrator writes a per-test output file named after it. `ContextImpl.makeFilename` rejects any filename containing a path separator, so the test `'strips metadata via separate input/output paths'` produces:

```
FATAL EXCEPTION: AndroidTestOrchestrator
Process: androidx.test.orchestrator
java.lang.IllegalArgumentException: File co.openvine.app.MainActivityTest#runDartTest[
  privacy.image_metadata_stripper_test … input/output paths].txt contains a path separator
    at android.app.ContextImpl.makeFilename(ContextImpl.java:3973)
    at androidx.test.orchestrator.AndroidTestOrchestrator.getOutputStream(AndroidTestOrchestrator.java:420)
```

Gradle turns that into `Instrumentation run failed due to Process crashed` and exits 1.

**This is the dangerous one.** The crash happens while *starting* the next test, so that test is never counted. patrol reported `Total: 2 / Successful: 2 / Failed: 0` and the JUnit XML agreed — `tests="2" failures="0" errors="0"` — for a file containing **three** tests. A green summary alongside a non-zero exit code was the only signal that a test had silently vanished.

## Evidence

All reproduced locally on an API 37 emulator, `androidx.test:orchestrator:1.5.1`, `ANDROIDX_TEST_ORCHESTRATOR` + `clearPackageData=true`:

| Defect | Signature |
|---|---|
| 1 | `patrol_cli` `versionCompatibilityList`: CLI `4.5.0+` ↔ patrol `4.7.0+`; `--check-compatibility` defaults on |
| 2 | `PUB_CACHE` relocated ⇒ `$HOME/.pub-cache/bin` absent ⇒ prepend is a no-op |
| 3 | `PATROL_EXTRA_ARGS[@]: unbound variable` on stock `/bin/bash` 3.2.57, default path — reproduced against `origin/main`'s copy of the script |
| 4 | patrol prompts `Please select an option (1-5, q to quit)` and loops on `/dev/null` stdin |
| 5 | Gradle exit 1 with patrol `Failed: 0` and JUnit `tests="2" failures="0"` on a three-test file |

## Impact

- **E2E is unrunnable on a clean setup.** Not degraded — the CLI refuses to start.
- **E2E is unrunnable on macOS** on the default invocation, which is the primary dev platform here.
- **False greens are possible today.** Defect 5 is a silent-skip class of bug: a test that never executes is indistinguishable from a passing one in both the patrol summary *and* the JUnit XML. Any CI lane built later inherits this failure mode unchanged.
- **It blocks #4239.** Any CI lane that issue stands up would `dart pub global activate patrol_cli` fresh and be rejected by the version handshake on day one.

## Scope

- Upgrade `patrol` `^4.2.0` → `^4.8.0` and pin `patrol_cli` to `4.6.1`, so the pair is version-locked in-repo rather than by whatever the machine happens to have. Note `dart pub upgrade patrol` **cannot** reach 4.7.0+ — it caps at 4.6.1, because 4.7.0+ requires `equatable ^2.1.0` and `upgrade <pkg>` will not unlock `equatable`; the constraint must be edited so `pub get` is the mechanism.
- `local_stack/profile.sh`: resolve the binary through `PUB_CACHE`, guard empty-array expansion for bash 3.2, and pass the detected device explicitly.
- Rename the patrol test containing `/`, and record the constraint where test authors will see it.
- Extend `local_stack/test_android_scripts.sh` to guard the script fixes.

## Out of scope

- **iOS patrol integration.** `main` has no `RunnerUITests` target and patrol is a `dev_dependency`, so `pod install` yields 43 pods and zero patrol — patrol's iOS code is not compiled into any build from `main`.
- **The `mobile_service_integration_tests.yaml` CI lane** — runner swap, emulator provisioning, PR triggers. That is #4239.
- **The `ProviderScope` bug in `integration_test/video_recorder/camera_permission_denied_test.dart`**, which fails with `Bad state: No ProviderScope found` because it pumps `VideoRecorderScreen` (a `ConsumerWidget`) without a `ProviderScope`. Verified identical on `origin/main` at patrol 4.2.0, so it is unrelated to the version pairing.
- **The remaining latent `/`-in-test-name instance** at `mobile/integration_test/perf/dm_remote_signer_backfill_frame_test.dart:163` (`build/raster`) — a perf suite patrol does not run today.

## Notes

CI proves nothing about any of this: no workflow runs `flutter build apk`, `xcodebuild`, `gradlew`, or `patrol`. The only CI signal on the changed code is `flutter analyze lib test integration_test`. Verification is necessarily local, on a device.

Unblocks #4239. Part of #4337.

realmeylisdev:
## Description

`patrol` has been pinned at `^4.2.0` since it landed with the E2E harness in #1928, and the lock still resolved exactly `4.2.0`. Latest is **4.8.0**.

The pairing matters more than the version. `patrol` and `patrol_cli` ship as a matched pair and **`patrol_cli` enforces it at run time** — its `versionCompatibilityList` maps `patrol_cli 4.5.0+` to `patrol 4.7.0+`, and `patrol test` runs that check with `--check-compatibility` defaulting to on. Nothing in this repo pinned `patrol_cli`; the only invocation, `local_stack/profile.sh`, took whatever binary happened to be on `PATH`.

| `patrol_cli` | `patrol` | Result |
|---|---|---|
| **4.6.1** | **4.8.0** | ✅ compatible — this PR |
| 4.2.0 | 4.2.0 | ✅ compatible — before this PR |
| 4.6.1 | 4.2.0 | ❌ **rejected** — what a fresh machine hits today |
| 4.2.0 | 4.8.0 | ❌ rejected — half-upgrading |

Row 3 is live: anyone installing `patrol_cli` fresh right now gets 4.6.1 and cannot run this repo's E2E suite at all.

4.8.0 / 4.6.1 are the highest versions that exist for either package — nothing newer is published, including prereleases — and both clear our Dart 3.12.0 / Flutter 3.44.0.

Running the suite to verify the bump surfaced three further blockers, each of which fails *after* the Docker stack and logcat capture are already up. All are pre-existing on `main` and all are fixed here as separate commits — see **Unrequested fixes**.

**Related Issue:** Closes #6930.

Also relates to #4239, but does **not** close it — this PR touches no file under `.github/workflows/`, so that issue's three axes (patrol runner in the workflow, emulator provisioning, PR triggers) are all still open. What it does do is remove a hard prerequisite: any CI lane #4239 stands up would `dart pub global activate patrol_cli` fresh and be rejected by the version handshake on day one. The version to pin is **`patrol_cli 4.6.1`**, paired with `patrol 4.8.0`.

### Commits

| Commit | |
|---|---|
| `fix(e2e): resolve patrol from the configured PUB_CACHE` | `profile.sh` prepended a hardcoded `$HOME/.pub-cache/bin` to `PATH`, which does not exist when `PUB_CACHE` points elsewhere — a silent no-op, so the script had no control over which CLI ran. |
| `build(mobile): upgrade patrol to 4.8.0` | `mobile/pubspec.yaml` + `mobile/pubspec.lock`. |
| `build(e2e): pin patrol_cli 4.6.1 in the profile runner` | Version preflight at the single call site + guards. |
| `docs(e2e-test): record the patrol / patrol_cli version pair` | The two halves live in different files and neither pointed at the other. |
| `fix(e2e): guard the optional patrol args for macOS bash 3.2` | *Unrequested* |
| `fix(e2e): pass the detected device to patrol` | *Unrequested* |
| `fix(e2e): drop the path separator from a patrol test name` | *Unrequested* |

### Why a constraint edit rather than a lock-only bump

`dart pub upgrade patrol` **cannot reach 4.8.0** — it caps at 4.6.1, because 4.7.0+ requires `equatable ^2.1.0` and `upgrade <pkg>` will not unlock `equatable`. Raising the floor to `^4.8.0` also forbids pub from resolving *backwards* into the 4.2.0–4.6.1 band the installed CLI rejects; that mismatch surfaces only as a run-time handshake failure, never as a compile error.

Resulting `pubspec.lock` delta is 4 packages / 8 lines:

| Package | Before | After |
|---|---|---|
| `patrol` | 4.2.0 | 4.8.0 |
| `patrol_finders` | 3.1.0 | 3.6.0 |
| `patrol_log` | 0.7.1 | 0.10.0 |
| `equatable` | 2.0.8 | 2.1.0 |

**On `equatable`:** it rides along as patrol's requirement, and the 2.0.8→2.1.0 diff is behaviour-neutral — `==`, `hashCode`, `toString`, `mapEquals`, `iterableEquals`, `setEquals` and `EquatableConfig` are behaviorally unchanged. 2.1.0 only turns `Equatable` into an `abstract mixin class` and deprecates `EquatableMixin`, which this repo never uses (0 occurrences under `mobile/`). All 23 workspace constraints are `^2.0.x` and already admit it, so no package pubspec changes.

### API surface

**Only one `.dart` file changes here, and only a test name string.** The repo's entire patrol surface is `patrolTest` (77×), `PatrolIntegrationTester` (8×), `PatrolBinding` (1×), `$.tester`, and five `$.platformAutomator` methods (`mobile.isPermissionDialogVisible`, `mobile.grantPermissionWhenInUse`, `mobile.denyPermission`, `mobile.pressHome`, `mobile.openApp`) plus `$.platformAutomator.tap(Selector(...))`. All exist unchanged in 4.8.0. The `patrol_finders` `$('text')` API is never used here, so the 3.4.0 visibility-wait and 3.5.0 `enterText` rewrite don't reach this repo.

## Unrequested fixes

All three are pre-existing on `main`, all block the E2E path, and none is cosmetic. Flagging them separately since none is part of the version bump.

**1. `PATROL_EXTRA_ARGS` unbound under macOS bash 3.2.** `profile.sh` runs `set -euo pipefail`, and macOS ships only bash 3.2, where expanding an empty array as `"${a[@]}"` trips `set -u`. The array is empty unless `PATROL_NO_UNINSTALL=true`, so every default `mise run e2e_test` on macOS died with `PATROL_EXTRA_ARGS[@]: unbound variable`. Reproduced against `origin/main`'s copy. Fixed with `"${a[@]+"${a[@]}"}"` (0 words when empty, elements preserved otherwise — verified on bash 3.2.57).

**2. The detected device was never passed to patrol.** `profile.sh` reads the device from `adb`, prints it, then omits it from the `patrol test` invocation. With more than one target attached — an iOS simulator, macOS and Chrome all count — patrol prompts for a selection and, under a non-interactive runner, loops on invalid input until killed.

**3. A `/` in a patrol test name crashes the AndroidX orchestrator.** patrol names each JUnit case `MainActivityTest#runDartTest[<dart test name>]`, and the orchestrator writes a per-test output file named after it. `ContextImpl.makeFilename` rejects filenames containing a path separator, so `'strips metadata via separate input/output paths'` produced:

```
FATAL EXCEPTION: AndroidTestOrchestrator
java.lang.IllegalArgumentException: File …input/output paths].txt contains a path separator
    at androidx.test.orchestrator.AndroidTestOrchestrator.getOutputStream
```

Gradle reported `Instrumentation run failed due to Process crashed` and exited 1. **The dangerous part is that it looked green:** the crash happened while starting the third test, so it was never counted — patrol printed `Total: 2 / Successful: 2 / Failed: 0` and the JUnit XML agreed (`tests="2" failures="0"`) for a file containing three tests. Renaming is the minimal fix; disabling `ANDROIDX_TEST_ORCHESTRATOR` would also stop the crash but would cost the per-test isolation and `clearPackageData` the app module deliberately configures. The rule is now documented in `.claude/skills/e2e-test/SKILL.md`.

One latent instance remains at `perf/dm_remote_signer_backfill_frame_test.dart:163` (`build/raster`), in the plain-`integration_test` group that patrol does not run today — left alone rather than speculatively renamed.

Fixes 1 and 2 ship with `assert_contains` guards in `local_stack/test_android_scripts.sh`, each **mutation-checked**: reverting the fix turns its own assertion red, restoring it turns it green.

## Out of Scope

- **iOS patrol integration.** Nothing to verify: `origin/main` has no `RunnerUITests` target, and patrol is a `dev_dependency`, so `pod install` installs **43 pods and zero patrol** — no `Pods/patrol/`, no `Podfile.lock` entry. patrol's iOS code is not compiled into any build produced from `main`, so patrol 4.7.0's darwin source restructure cannot affect iOS here and this PR cannot regress it. Once a UI-test target lands, the follow-up is whether 4.8.0 lets it drop the `inherit! :complete` / `ENABLE_TESTING_SEARCH_PATHS` workarounds.
- **The `ProviderScope` bug in `camera_permission_denied_test.dart`** — see below. Repairing it needs a `ProviderScope` with overrides for `clipManagerProvider` / `videoEditorProvider` / `sharedPreferencesProvider`; unrelated to a version bump.
- **Fixing `mobile_service_integration_tests.yaml`** — that is #4239.
- **Enabling Swift Package Manager** — a ~50-plugin migration. CocoaPods remains supported by patrol 4.8.0.

## Verification

Run from `mobile/` unless noted. The device rungs are local-only; CI runs none of them.

| Check | What it proves | Result |
|---|---|---|
| `dart pub upgrade --dry-run patrol` | resolution exists; showed the 4.6.1 cap | ✅ |
| `flutter analyze lib test integration_test` | all 31 patrol-importing files resolve against 4.8.0's API; no ambiguous imports from the new `Notification` / `Rectangle` / `Point2D` exports | ✅ `No issues found!` |
| `flutter analyze integration_test` on the **regenerated** `test_bundle.dart` | the internal coupling — the bundle imports `package:patrol/src/platform/contracts/contracts.dart` and calls `PatrolBinding.ensureInitialized`, `platformAutomator.initialize`, `markPatrolAppServiceReady` | ✅ `No issues found!` |
| `patrol build android --debug` | patrol 4.8.0's new Ktor 3.2.2 / http4k 5.47 graph dexes onto the androidTest classpath; its Kotlin 2.1.0 buildscript assembles under AGP 8.11.1 / KGP 2.2.20 / `android.builtInKotlin=false`. Runs `--check-compatibility`, so this is the **empirical** proof that CLI 4.6.1 ↔ patrol 4.8.0 pair. | ✅ both APKs built, 0 duplicate-class errors |
| `local_stack/test_stack_scripts.sh` + `test_android_scripts.sh` | the new guards | ✅ |
| **Emulator E2E** — `privacy/image_metadata_stripper_test.dart` via `profile.sh` | full round trip: bundle generation, `PatrolJUnitRunner` + orchestrator enumeration, on-device execution, per-test reporting | ✅ **`Total: 3, Successful: 3, Failed: 0`**, JUnit XML `tests="3" failures="0" errors="0"`, `✓ Completed executing`, 0 orchestrator crashes |
| Emulator E2E — `video_recorder/camera_permission_denied_test.dart` | native automation specifically — `isPermissionDialogVisible` / `denyPermission` | ✅ native step executed; test then fails on a **pre-existing** app bug, below |
| `pod install` (`mobile/ios`) | patrol contributes no iOS pod on `main` | ✅ 43 pods, 0 patrol |

### A pre-existing broken test found on the way (not fixed here)

`integration_test/video_recorder/camera_permission_denied_test.dart` fails with `Bad state: No ProviderScope found`. It pumps `MaterialApp(home: BlocProvider(… VideoRecorderScreen))`, but `VideoRecorderScreen`'s `_VideoRecorderBlocScope` is a `ConsumerWidget` calling `ref.watch(clipManagerProvider.notifier)` — Riverpod needs a `ProviderScope` ancestor the test never provides.

**Proven pre-existing** by running the same test on `origin/main` at patrol 4.2.0 with patrol_cli 4.2.0 in a throwaway worktree:

| | `Total` | `Successful` | `Failed` | Outcome |
|---|---|---|---|---|
| `origin/main` — patrol 4.2.0 / CLI 4.2.0 | 1 | 0 | 0 | same `StateError: No ProviderScope found` |
| this branch — patrol 4.8.0 / CLI 4.6.1 | 1 | 0 | 1 | same `StateError: No ProviderScope found` |

Note the reporting difference: 4.2.0 lost the attribution entirely (`Failed: 0` despite exiting 1), while 4.8.0 reports `Failed: 1` and names the test — an improvement from the upgrade.

### What CI proves here: nothing native

No workflow runs `flutter build apk`, `xcodebuild`, `gradlew`, or patrol. The only CI signal this PR gets is `flutter analyze lib test integration_test` (`mobile_ci.yaml:321`). Green checks say nothing about the device rungs above — those were run locally and are reported here.

### Two things to know before pulling

- **`dart pub global activate patrol_cli 4.6.1` is machine-global** and git cannot roll it back. `profile.sh` now does it automatically on the next `mise run e2e_test`. Any other worktree still on `patrol 4.2.0` will then fail the compatibility check until it rebases onto this — loudly, with the exact remediation command.
- **The lock change re-arms one `build_runner` pass** for everyone on first pull: `docs/BUILD_SPEED_CHECKLIST.md:21-38` short-circuits codegen on a `pubspec.lock` hash.

## Type of Change

- [x] 📝 Documentation

### DIFF
--- a/.agents/skills/e2e-test/SKILL.md
+++ b/.agents/skills/e2e-test/SKILL.md
@@ -5,7 +5,7 @@ description: |
   app against a local Docker backend (no mocks). Use when running
   E2E tests, debugging failures, or working on the local harness.
 author: Claude Code
-version: 1.2.0
+version: 1.3.0
 ---
 
 # E2E Integration Testing
@@ -33,6 +33,25 @@ merged docker+logcat+app timeline at `test_reports/*.jsonl`, and
 prints the native test XML path + failure excerpts when the APK
 fails to install. **Never call `patrol test` directly** — you'll
 lose the timeline and the diagnostics.
+
+## Version pair
+
+`patrol` (the package) and `patrol_cli` (the binary) ship as a matched
+pair, and `patrol_cli` enforces it at run time — a mismatch aborts the
+run before any test executes.
+
+| Half | Version | Declared in |
+|---|---|---|
+| `patrol` package | 4.8.0 | `mobile/pubspec.yaml` (`patrol: ^4.8.0`) |
+| `patrol_cli` binary | 4.6.1 | `local_stack/profile.sh` (`PATROL_CLI_VERSION`) |
+
+`profile.sh` checks the installed CLI and runs
+`dart pub global activate patrol_cli <version>` when it differs, so
+`mise run e2e_test` self-heals. That activation is **machine-global**:
+it switches the CLI for every checkout, including worktrees still on an
+older `patrol`, which will then fail the same compatibility check until
+they rebase. Change the two versions together — the compatibility table
+is at https://patrol.leancode.co/documentation/compatibility-table.
 
 ## Stack
 
--- a/.agents/skills/e2e-test/SKILL.md
+++ b/.agents/skills/e2e-test/SKILL.md
@@ -258,6 +258,28 @@ Patrol bundles every file in a target dir into one APK. When file B
 runs, file A shows up as "not requested" `[E]` markers in logcat.
 Trust only the final `✅`/`❌` lines.
 
+### Never put `/` in a patrol test name
+
+Patrol names each JUnit case `MainActivityTest#runDartTest[<dart test
+name>]`, and the AndroidX orchestrator writes a per-test output file
+named after it. Android's `ContextImpl.makeFilename` rejects any
+filename containing a path separator, so a test called e.g. `'strips
+metadata via separate input/output paths'` crashes the orchestrator:
+
+```
+FATAL EXCEPTION: AndroidTestOrchestrator
+java.lang.IllegalArgumentException: File …input/output paths].txt
+contains a path separator
+```
+
+The tell is a **green summary with a non-zero exit**: Gradle reports
+`Instrumentation run failed due to Process crashed` and exits 1, while
+patrol prints `Failed: 0` — because the offending test never started
+and so was never counted. Compare `Total:` against the number of tests
+in the file when the exit code disagrees with the summary.
+
+Write `input and output`, not `input/output`.
+
 ### NIP-98 URL binding
 
 The invite service rejects a signed request whose `u` tag does not
--- a/.claude/skills/e2e-test/SKILL.md
+++ b/.claude/skills/e2e-test/SKILL.md
@@ -5,7 +5,7 @@ description: |
   app against a local Docker backend (no mocks). Use when running
   E2E tests, debugging failures, or working on the local harness.
 author: Claude Code
-version: 1.2.0
+version: 1.3.0
 ---
 
 # E2E Integration Testing
@@ -33,6 +33,25 @@ merged docker+logcat+app timeline at `test_reports/*.jsonl`, and
 prints the native test XML path + failure excerpts when the APK
 fails to install. **Never call `patrol test` directly** — you'll
 lose the timeline and the diagnostics.
+
+## Version pair
+
+`patrol` (the package) and `patrol_cli` (the binary) ship as a matched
+pair, and `patrol_cli` enforces it at run time — a mismatch aborts the
+run before any test executes.
+
+| Half | Version | Declared in |
+|---|---|---|
+| `patrol` package | 4.8.0 | `mobile/pubspec.yaml` (`patrol: ^4.8.0`) |
+| `patrol_cli` binary | 4.6.1 | `local_stack/profile.sh` (`PATROL_CLI_VERSION`) |
+
+`profile.sh` checks the installed CLI and runs
+`dart pub global activate patrol_cli <version>` when it differs, so
+`mise run e2e_test` self-heals. That activation is **machine-global**:
+it switches the CLI for every checkout, including worktrees still on an
+older `patrol`, which will then fail the same compatibility check until
+they rebase. Change the two versions together — the compatibility table
+is at https://patrol.leancode.co/documentation/compatibility-table.
 
 ## Stack
 
--- a/.claude/skills/e2e-test/SKILL.md
+++ b/.claude/skills/e2e-test/SKILL.md
@@ -258,6 +258,28 @@ Patrol bundles every file in a target dir into one APK. When file B
 runs, file A shows up as "not requested" `[E]` markers in logcat.
 Trust only the final `✅`/`❌` lines.
 
+### Never put `/` in a patrol test name
+
+Patrol names each JUnit case `MainActivityTest#runDartTest[<dart test
+name>]`, and the AndroidX orchestrator writes a per-test output file
+named after it. Android's `ContextImpl.makeFilename` rejects any
+filename containing a path separator, so a test called e.g. `'strips
+metadata via separate input/output paths'` crashes the orchestrator:
+
+```
+FATAL EXCEPTION: AndroidTestOrchestrator
+java.lang.IllegalArgumentException: File …input/output paths].txt
+contains a path separator
+```
+
+The tell is a **green summary with a non-zero exit**: Gradle reports
+`Instrumentation run failed due to Process crashed` and exits 1, while
+patrol prints `Failed: 0` — because the offending test never started
+and so was never counted. Compare `Total:` against the number of tests
+in the file when the exit code disagrees with the summary.
+
+Write `input and output`, not `input/output`.
+
 ### NIP-98 URL binding
 
 The invite service rejects a signed request whose `u` tag does not
--- a/local_stack/profile.sh
+++ b/local_stack/profile.sh
@@ -124,6 +124,7 @@ echo "Running: patrol test ${TEST_PATH} ..." >&2
 cd "$MOBILE_DIR"
 set +e
 PATH="$PUB_CACHE_BIN:$PATH" patrol test \
+    --device "$DEVICE" \
     --target "$TEST_PATH" \
     --dart-define=DEFAULT_ENV=LOCAL \
     --dart-define=INVITE_SERVER_URL="$INVITE_SERVER_URL" \
--- a/local_stack/profile.sh
+++ b/local_stack/profile.sh
@@ -65,6 +65,33 @@ if ! local_stack_has_running_container "$COMPOSE_FILE"; then
     exit 1
 fi
 
+# --- Ensure the patrol CLI matches the patrol package ---
+# patrol_cli enforces the pairing itself and aborts the run on a mismatch
+# (patrol_cli 4.5.0+ pairs with patrol 4.7.0+). Activating the pinned CLI before
+# log capture starts keeps install/compile time out of the profiler timeline.
+# Keep in sync with the patrol constraint in mobile/pubspec.yaml.
+PATROL_CLI_VERSION=4.6.1
+PATROL_BIN=""
+if [ -x "$PUB_CACHE_BIN/patrol" ]; then
+    PATROL_BIN="$PUB_CACHE_BIN/patrol"
+elif command -v patrol >/dev/null 2>&1; then
+    PATROL_BIN="$(command -v patrol)"
+fi
+
+PATROL_CURRENT_VERSION=""
+if [ -n "$PATROL_BIN" ]; then
+    # `patrol --version` prints an update banner before the version line. The
+    # update check can fail offline, so treat probe failure as "version unknown"
+    # and let the pinned activation below repair the install.
+    PATROL_CURRENT_VERSION="$("$PATROL_BIN" --version 2>/dev/null |
+        sed -n 's/^patrol_cli v\([0-9][0-9.]*\).*/\1/p' | tail -1)" || true
+fi
+
+if [ "$PATROL_CURRENT_VERSION" != "$PATROL_CLI_VERSION" ]; then
+    echo "Installing patrol_cli ${PATROL_CLI_VERSION} (pinned to match the patrol package)..." >&2
+    dart pub global activate patrol_cli "$PATROL_CLI_VERSION" >&2
+fi
+
 # --- Start docker log capture (background, from now only) ---
 echo "Starting docker log capture..." >&2
 docker compose -f "$COMPOSE_FILE" logs -f -t --since "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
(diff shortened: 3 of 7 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I pushed one reviewer follow-up commit (`e8b800541`) to close out the remaining script issues:

- moved the pinned `patrol_cli` preflight before docker/logcat capture so activation time stays out of the profiler timeline;
- made the `patrol --version` probe non-fatal when the CLI update check fails offline;
- added shell assertions for those two guards;
- corrected the `pubspec.yaml` Patrol floor comment and the PR body wording around `equatable`.

Verification passed:

- `bash -n local_stack/profile.sh && bash -n local_stack/test_android_scripts.sh`
- `mise run test_local_stack_scripts`
- pre-push analyzer hook
- GitHub checks on `e8b800541`: Mobile CI, preview build, mergeability, semantic PR, linked issue, CLA all green

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up.

I pushed `e8b800541` to harden the Patrol CLI preflight, keep activation out of profiler log capture, add shell guards, and fix the small metadata wording nits. Local script verification and GitHub checks are green on the new head.

NotThatKindOfDrLiz:
Review complete: comprehensive analysis identified two execution-flow defects in local_stack/profile.sh that warrant a follow-up commit. Detailed findings posted in thread below.

NotThatKindOfDrLiz:
**Finding 1: Version probe silently fails offline (Lines 112–114)**

When patrol --version exits non-zero (network unavailable, CLI outage), the assignment silently completes with PATROL_VERSION="" and set -e terminates the script while logs are running.

Fix: Add || true to the version probe assignment.

**Finding 2: Activation block executes after log captures start (Lines 96–119)**

The activation block runs AFTER docker-compose startup and logcat capture begin, polluting the profiler timeline despite the comment claiming it prevents mid-run execution.

Fix: Move activation block above line 68 (docker-compose up -d).

**Verified:** All Dart APIs backward-compatible, equatable 2.0.8 → 2.1.0 byte-identical, bash 3.2 syntax verified, patrol_cli version pin validated, no import collisions.

Recommendation: Request one-commit follow-up addressing both findings. After that lands and checks pass green, this PR is approvable.

realmeylisdev:
Both of these landed before merge, in `e8b800541` ("fix(e2e): harden patrol cli preflight").

- Finding 1 — the probe assignment ends `| tail -1)" || true`, with the comment:
  "The update check can fail offline, so treat probe failure as 'version unknown'
  and let the pinned activation below repair the install."
- Finding 2 — the activation block sits above the docker log capture, with the comment:
  "Activating the pinned CLI before log capture starts keeps install/compile time
  out of the profiler timeline."

Current `main` still has both, after #7228 moved the pair to patrol 4.9.0 /
patrol_cli 4.7.0 — probe at `local_stack/profile.sh:87`, activation block at
`:68-93`, docker log capture at `:96`, logcat at `:126`.

The cited lines 112–114 and 96–119 match the file at `e8b800541^`.

PR #6928, opened 2026-08-09
fix(clip-reuse): preserve multi-source clip credits

NotThatKindOfDrLiz:
## Summary
- add a durable multi-source `ClipSourceCredit` model while keeping legacy scalar clip provenance readable
- preserve and dedupe clip-source provenance through drafts, restore, merge, publish, and event parsing
- emit factual clip-source `a`/`p` tags separately from manual inspired-by attribution and render compact multi-creator credit

## Verification
- `flutter analyze lib test` from `mobile/`
- `flutter test test/models/divine_video_clip_test.dart test/providers/video_editor_provider_test.dart test/services/video_editor/video_editor_merge_service_test.dart test/services/video_event_publisher_inspired_by_tags_test.dart test/services/video_publish/video_publish_service_test.dart test/widgets/video_feed_item/inspired_by_attribution_row_test.dart` from `mobile/`
- `flutter test test/src/video_event_to_json_test.dart test/src/video_event_from_json_test.dart test/src/video_event_parsing_test.dart` from `mobile/packages/models`
- `flutter test test/widgets/popular_videos_tab_test.dart` from `mobile/` while investigating the first CI shard failure

Refs #6315

Out of scope for this slice: the creator consent setting, `allow_clip_reuse` policy gate, and editor-render watermarking. Those still need product/design answers before implementation.


rabble (changes requested):
Reviewed the full diff plus the surrounding publish/edit/library paths. The protocol shape is good and the wire behaviour for the existing single-source case is genuinely unchanged — I traced that carefully and it holds. What blocks this is one remaining path that still truncates N credits to 1, plus a `copyWith` shape that does the same thing inside the model.

## What's right

- **No new protocol shapes.** `clip-source` is a marker on the existing `a` and `p` tags, mirroring the shipped `collaborator` / `inspired-by` convention rather than inventing a tag name or kind. Correct call.
- **Single-source degrades exactly to legacy.** `video_event_publisher.dart:1447` seeds `emittedAddressableCredits` with `inspiredByAddressableId`, so the duplicate `clip-source` a-tag is suppressed; `buildClipSourceCreditPTags` (`inspired_by_tags.dart:115`) dedups against `existingTags`, so the `inspired-by` p-tag wins. A one-reused-clip publish emits byte-identical tags to `main`. Good.
- **Round-trip parse is order-independent.** `_addClipSourceCredit` (`video_event.dart:786`) upgrades an author-only credit to an addressable one regardless of whether the `a` or the `p` tag is seen first, and the test at `video_event_parsing_test.dart:733` pins it.
- **Self-suppression** is handled on both the a-tag (`video_event_publisher.dart:1454`) and the p-tag path, with a test.
- **Edit flow survives.** `_isEditRebuiltTag` (`video_metadata_update_service.dart:93-111`) only claims the `inspired-by`/`mention` markers, so `clip-source` tags fall through to `preservedTags` untouched. I checked this because the edit path isn't in the diff — it's fine, but it means the credits survive by *not* being owned rather than by being re-emitted. Worth a comment saying so, since `VideoEditorProviderState.clipSourceCredits` is otherwise write-only in the edit path.
- Legacy JSON and legacy in-memory clips migrate through `_normalizedSourceCredits`; `N=0` is `const []` everywhere; `N=many` is covered by tests at `video_editor_merge_service_test.dart:223` and `video_editor_provider_test.dart:2663`. Those tests would fail if the union logic regressed.

## Blocking

### 1. `flattenClipForLibrary` still truncates to the first credit

`mobile/lib/services/video_editor/video_editor_clip_library_save_service.dart:95-99` still constructs the flattened clip from the four scalar fields:

```dart
      // A flatten re-renders the same logical clip, so its imported-source
      // attribution carries over unchanged.
      sourceAuthorPubkey: clip.sourceAuthorPubkey,
      sourceEventId: clip.sourceEventId,
      sourceAddressableId: clip.sourceAddressableId,
      sourceRelayHint: clip.sourceRelayHint,
```

Those scalars are now defined as "the first credit" (`divine_video_clip.dart:213-215`). So a merged clip that `VideoEditorMergeService.mergeClips` just gave credits `[A, B, C]` loses B and C the moment the user saves it to the clip library — and the library clip is then reusable in future videos, so the loss is permanent and silent. This is reachable from `ClipEditorBloc` (`clip_editor_bloc.dart:1086`), which flattens whatever clip is selected in the timeline, merged ones included.

This is the same failure mode the PR exists to fix, so it should land in the same PR: pass `sourceCredits: clip.sourceCredits` and drop the scalars. The comment above it ("attribution carries over unchanged") is now false and needs updating either way.

### 2. `copyWith` silently discards the credit list when any scalar source field is touched

`mobile/lib/models/divine_video_clip.dart:392-425`:

```dart
    final nextSourceCredits = clearSourceCredits
        ? const <model.ClipSourceCredit>[]
        : sourceCredits ??
              _normalizedSourceCredits(
                sourceCredits: sourceFieldsTouched
                    ? const []
                    : this.sourceCredits,
```

`sourceFieldsTouched` is true if *any* of the eight scalar/clear params is set. So on a 3-credit clip, `clip.copyWith(sourceRelayHint: 'wss://x')` — which reads like "adjust the relay hint" — throws away credits 2 and 3 and rebuilds a single-credit list. Nothing in the diff exercises this; the existing `copyWith` test (`divine_video_clip_test.dart:341`) only covers a single-credit clip.

I understand why the rule exists (a caller setting the scalars means "replace the source"), but the blast radius doesn't match the call syntax, and `video_editor_merge_service_test.dart:245` shows scalar `copyWith` is still an idiomatic way to build a clip. Either assert that the scalar params aren't mixed with a non-empty `sourceCredits`, or make the scalars additive. Whichever you pick, please add a test that pins it.

## Needs an answer (not necessarily a change)

### 3. Clip credits shadow an explicitly chosen "Inspired by"

`video_event.dart:1036-1038` puts clip credits ahead of everything:

```dart
    if (clipSourceCredits.isNotEmpty) {
      return clipSourceCredits.first.authorPubkey;
    }
    if (inspiredByVideo != null) return inspiredByVideo!.creatorPubkey;
```

and `inspired_by_attribution_row.dart:59-63` does the same with an early `return pubkeys`. So a video where the user explicitly picked an "Inspired by" person (`inspiredByNpub`) *and* reused two clips will credit the clip authors and never mention the person they chose — in the attribution row and in `MetadataInspiredBySection` (`metadata_user_chips.dart:246`). The audio-reuse flow auto-populates the video reference independently of the person picker, per the doc comment on `resolveInspiredByCreatorHexes`, so the combination isn't hypothetical.

`hasInspiredBy` merges the two sources; the pubkey resolvers pick one and discard the other. Was shadowing intended over merging? If yes, a comment on `inspiredByCreatorPubkey` explaining the precedence would help the next reader.

## Non-blocking

- **Dead parameter.** `resolveInspiredByCreatorHexes` and `buildInspiredByPTags` both gained a `clipSourceCredits` param (`inspired_by_tags.dart:52, 90`), but neither production caller passes it — `video_event_publisher.dart:1487` and `video_metadata_update_service.dart:357` both call the old signature, and clip credits go through the separate `buildClipSourceCreditPTags`. Drop the param, or the next reader will assume it's load-bearing.
- **Constructor does five list builds.** `divine_video_clip.dart:53-88` calls `_firstSourceCredit` four times, and each one calls `_normalizedSourceCredits` internally, on top of the initializer's own call. Five normalizations per clip construction, in the editor's hot path. A redirecting constructor or a private factory that normalizes once would be both faster and much easier to read.
- **`+N` bypasses l10n.** `inspired_by_attribution_row.dart:114-116` builds `'$creatorName +${creatorPubkeys.length - 1}'` in Dart and hands it to `videoInspiredByAttribution`. That's a user-visible string composed outside the ARB, and the `+1` lands on the wrong side in RTL locales. An ICU-parameterized key (`{name} +{count}`) would fix both.
- **Docs.** `mobile/docs/NOSTR_VIDEO_EVENTS.md:69-77` has a table of Divine's `p`-tag role markers. `clip-source` isn't in it. (Neither is `inspired-by`, so this is a pre-existing gap rather than something you introduced — but this is the natural PR to close it, since you're adding a marker to `a` tags too, which the doc doesn't cover at all.)

Nothing here touches truncated Nostr IDs — full hex throughout, in code, tags, logs, and tests. Good.

Happy to re-review quickly once #1 and #2 are settled.

hm21 (dismissed):
Pushed `aeec25e`. `getActiveDraft` fell back to the restored draft's `clipSourceCredits` whenever the timeline produced none, and `state.clipSourceCredits` is only written at restore and never recomputed — so deleting the reused clip, or restore dropping it as orphaned, left the credit behind. The publish then emitted a `clip-source` `a`/`p` tag, and notified that creator, for footage the video no longer contains. Dropped the fallback; two regression tests under `restoreDraft` cover both paths (both red before the change). Checks are green.

One consequence of that change to be aware of: `VideoEditorProviderState.clipSourceCredits` now has no reader. `initFromPublishedVideo` still writes it, but the metadata-edit flow republishes through `VideoMetadataUpdateService`, which never reads it back — which is the same write-only observation @rabble made about the edit path. I left the field in place rather than removing it, since whether it stays depends on how you want the edit path to own these tags.

Still open from @rabble's review: `flattenClipForLibrary` (`video_editor_clip_library_save_service.dart:96`) is the last remaining truncation to one credit — I checked, and it is the only scalar-only clip construction left in `lib/`. One nuance on their `copyWith` item: no production code passes `source*` to `copyWith` today, only tests, so it is a latent trap rather than a live bug.

mobile/packages/models/lib/src/video_event.dart
```diff
@@ -963,6 +1034,9 @@ class VideoEvent {
   /// cannot be decoded.
   String? get inspiredByCreatorPubkey {
     if (isVideoReply) return null;
+    if (clipSourceCredits.isNotEmpty) {
+      return clipSourceCredits.first.authorPubkey;
```
realmeylisdev: this shadows instead of merging, and it flips behaviour for one real combination: 2+ reused clips **and** an explicitly picked "Inspired by" person.

on `main` that case has no clip credits and `_attributionSourceFromClips` returns null (multiple sources), so `inspiredByNpub` wins and both the feed row and `MetadataInspiredBySection` credit the person the user chose. here `clipSourceCredits` is non-empty, so both surfaces show the clip authors and the chosen person is never named anywhere in the UI. the npub still ships (caption line at `video_event_publisher.dart:1838` plus an `inspired-by` p-tag), it just becomes invisible.

single-source is genuinely unaffected, i traced it: `_attributionSourceFromClips` sets `inspiredByVideo`, the duplicate clip-source a-tag and p-tag are both deduped away, and the parsed event carries no clip credits at all.

so the question is whether the person is meant to lose to the clips, or whether the two should merge (person first, then clip authors). `hasInspiredBy` already unions them, which is what makes the precedence here read like an accident rather than a decision. either way a line of comment on this getter would save the next reader the trace.

realmeylisdev (commented):
went through the publish, edit and library paths plus the funnelcake side. a clip-source p-tag lands there as a plain "mention" notification, exactly like inspired-by, so nothing to change on the backend.

pushed six commits: flattenClipForLibrary keeps every credit now, copyWith can't drop credits past the first, the dead clipSourceCredits param is gone, credits get normalized once per clip instead of five times, the "+N" moved into the ARB across all 22 locales, and the a/p markers are documented. checks are green, and i ran it on an iphone sim through import, library and editor with a clean log. revert any of them freely.

one open item inline, it's a product call so i left it alone.

NotThatKindOfDrLiz:
Follow-up fixes pushed in `38392a3fc`:

- Removed the write-only `VideoEditorProviderState.clipSourceCredits` state and the restore/init writes. Draft publish state still comes from the timeline via `getActiveDraft()`, and metadata edits continue preserving existing `clip-source` tags rather than owning/rebuilding them.
- Added an explicit comment in `VideoMetadataUpdateService` documenting that `clip-source` tags intentionally fall through to `preservedTags`.
- Resolved the open attribution precedence question by keeping explicit Inspired By visible: `inspiredByCreatorPubkey` now prefers explicit inspired-by metadata, and the feed row merges explicit inspired-by creators with factual clip-source authors, deduped, instead of returning clip credits early.
- Added regressions for explicit inspired-by + clip-source credits in the model and feed row tests.

Verification:

- `flutter test test/src/video_event_parsing_test.dart` from `mobile/packages/models`
- `flutter test test/widgets/video_feed_item/inspired_by_attribution_row_test.dart test/providers/video_editor_provider_test.dart` from `mobile`
- Pre-push hook: analyzer plus changed-file test set, all passed

@rabble ready for re-review.

mobile/packages/models/lib/src/video_event.dart
realmeylisdev: fixed by 38392a3fc, thanks. the chosen person is back in front of the clip authors on both surfaces, and i flipped the order back on purpose to check the new tests actually catch it (they do, both the model one and the widget one).

one leftover that i don't think belongs in this PR: the About sheet still names a single creator while the feed row says +N. that's the multi-creator presentation question already open in #6315.

realmeylisdev (approved):
lgtm. @rabble's two blockers are in (library save keeps every credit, copyWith can't drop them) and the shadowing one is fixed too. i broke each fix on purpose first and the new tests went red, so they're real guards rather than decoration.

re-checked funnelcake: a clip-source p-tag still lands as a plain mention notification and nothing there reads the a-tag marker, so no backend change. ran a restored 2-clip reused draft on the sim, editor and post details were clean.

NotThatKindOfDrLiz:
Follow-up pushed in `89a875bcb`:\n\n- Emit the factual `clip-source` a-tag even when the same source is also the legacy Inspired By `mention` a-tag, so clearing manual Inspired By during metadata edit no longer erases single-source clip provenance. The p-tag path still dedupes, so the source creator gets one notify tag.\n- Added a metadata-edit regression that clears Inspired By while preserving `clip-source` a/p tags.\n- Tightened clip-source JSON restore to drop malformed saved credit entries instead of failing draft/cache restore, and added field-level round-trip assertions so equality-by-identity can’t hide dropped relay/event hints.\n- Swapped the touched relay fallbacks in the publisher to `AppConstants.defaultRelayUrl`.\n- Updated `NOSTR_VIDEO_EVENTS.md` to document that manual attribution and factual clip-source provenance are emitted independently.\n\nVerification after rebasing onto current `origin/main`:\n\n- `flutter test test/services/video_event_publisher_inspired_by_tags_test.dart test/services/video_metadata_update_service_test.dart test/models/divine_video_clip_test.dart` from `mobile/`\n- `flutter test test/src/video_event_from_json_test.dart test/src/video_event_to_json_test.dart test/src/video_event_parsing_test.dart` from `mobile/packages/models/`\n- `flutter analyze lib test` from `mobile/`\n- `flutter analyze lib test` from `mobile/packages/models/`\n- pre-push hook analyzer + changed-file test set, all passed\n\n@rabble ready for re-review.

NotThatKindOfDrLiz:
@rabble ready for your rereview so we can close out this pr. tysm!

mobile/lib/utils/inspired_by_tags.dart
```diff
@@ -94,3 +102,31 @@ List<List<String>> buildInspiredByPTags({
...
+    for (final tag in existingTags)
+      if (tag.length >= 2 && tag[0] == 'p') tag[1].trim().toLowerCase(),
+  };
+  final seen = <String>{};
+
+  return [
+    for (final credit in clipSourceCredits)
+      if (seen.add(credit.authorPubkey.trim().toLowerCase()))
+        if (credit.authorPubkey.trim().toLowerCase() != self &&
+            !existing.contains(credit.authorPubkey.trim().toLowerCase()) &&
```
realmeylisdev, line 146: Deduping against `existingTags` means a `clip-source` p-tag is never emitted when that pubkey already carries an `inspired-by` p-tag. That isn't an edge case — it's the default: `_attributionSourceFromClips()` auto-populates `inspiredByVideo` from the single reused clip, so one reused clip always produces both, pointing at the same creator.

I ran the publisher for that case. It emits:

```
["a", "34236:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb:source-d-tag", "wss://source.relay", "mention"]
["a", "34236:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb:source-d-tag", "wss://source.relay", "clip-source"]
["p", "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb", "wss://source.relay", "inspired-by"]
```

Feeding exactly those tags through `VideoMetadataUpdateService.updateVideo` with Inspired By cleared leaves:

```
a-tags: [["a", "34236:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb:source-vid", "wss://source.relay", "clip-source"]]
p-tags: []
```

The a-tag survives, which is the fix working. But the creator's only p-tag left with the inspired-by tag, and funnelcake builds notification targets from p-tags alone — `all_len64_tag_values(&candidate.source_tags, "p")` in `crates/clickhouse/src/client.rs`; a-tags only feed the card and root context. So the edited event still states the provenance in a machine-readable tag that notifies nobody.

The doc line added in the same commit says the two are emitted independently "so clearing manual Inspired By attribution during metadata edit does not erase factual clip provenance." That's true of the a-tag half only.

Related: the new `preserves clip-source tags when manual inspired-by is cleared` test passes because its fixture hand-injects `["p", sourceCreator, "wss://source.relay", "clip-source"]` — a tag the publisher provably never emits in that scenario. Swap the fixture for the publisher's real output and the p-tag assertion goes red, so the guard doesn't currently cover the path it's named for.

Not prescribing a fix — allowing a second p-tag for the same pubkey crosses the single-p-tag-per-pubkey invariant this function documents, and having the edit flow re-derive clip-source p-tags from the preserved a-tags changes who owns them. Either is your call.

mobile/packages/models/lib/src/video_attribution.dart
```diff
@@ -67,3 +64,101 @@ class InspiredByInfo {
...
+@immutable
+class ClipSourceCredit {
+  const ClipSourceCredit({
+    required this.authorPubkey,
+    this.eventId,
+    this.addressableId,
+    this.relayUrl,
+  });
+
+  factory ClipSourceCredit.fromJson(Map<String, dynamic> json) {
```
realmeylisdev: `ClipSourceCredit.fromJson` has no callers anywhere in the repo now — every production path (`video_event.dart`, `divine_video_clip.dart`, `divine_video_draft.dart`) moved to `tryFromJson` in `89a875bcb`, and no test constructs through it either. Worth deleting the throwing wrapper along with the `FormatException` rather than leaving a public factory nothing reaches.

realmeylisdev (commented):
Re-reviewed at `ae526102a` for the two commits that landed after my approval. Everything from the first round still holds, and the scoped suites plus the ARB consistency check are green locally.

One item inline that I think needs a decision before merge: dropping the a-tag dedup fixed half the problem, but the matching `clip-source` p-tag is still deduped away, so clearing Inspired By in a metadata edit leaves the credited creator with no p-tag at all. I reproduced it end-to-end against the publisher's real output rather than a hand-built fixture. It's a design call, so I haven't touched it.

Branch is behind `main` — worth rebasing rather than merging when you refresh it.

NotThatKindOfDrLiz:
Follow-up fixes pushed in `957f89e8df`:\n\n- Metadata edits now re-derive missing `p ... clip-source` notify tags from preserved `a ... clip-source` provenance tags, so clearing manual Inspired By removes the manual attribution p-tag without dropping factual clip-source notification targets.\n- Updated the metadata-edit regression to start from the publisher's real single-source tag shape: `a ... mention`, `a ... clip-source`, and `p ... inspired-by`.\n- Removed the dead `ClipSourceCredit.fromJson` wrapper and collapsed the duplicated clip-source credit list parsing into `ClipSourceCredit.listFromJson`.\n- Removed the unused `videoInspiredByAttributionSemanticLabel` l10n key and regenerated localization files.\n- Rebasing onto current `origin/main` dropped the merge commit; branch is now 0 behind.\n\nVerification:\n\n- `flutter test test/services/video_metadata_update_service_test.dart test/services/video_event_publisher_inspired_by_tags_test.dart test/models/divine_video_clip_test.dart` from `mobile/`\n- `flutter test test/src/video_event_from_json_test.dart test/src/video_event_to_json_test.dart test/src/video_event_parsing_test.dart` from `mobile/packages/models/`\n- `flutter analyze lib test` from `mobile/`\n- `flutter analyze lib test` from `mobile/packages/models/`\n- pre-push hook: no merge conflicts, analyzer, and changed-file test set all passed\n\n@rabble please re-review.

rabble (commented):
Re-reviewed. **Both blocking items addressed.** CI green (16/16) on `957f89e8d`.

**1. `flattenClipForLibrary` no longer truncates.** It now passes `sourceCredits: clip.sourceCredits` and the four scalars are gone. The comment was updated rather than left asserting something false, and it now says *why* the multi-credit case matters — the library clip is reusable later, so a drop here is permanent. The new test in `video_editor_clip_library_save_service_test.dart` flattens a 2-credit clip and asserts `sourceCredits[1]`, so this can't silently regress to first-credit-only.

**2. `copyWith` — you removed the hazard rather than guarding it.** The scalar source params and the whole `sourceFieldsTouched` mechanism are gone; it's now `sourceCredits ?? this.sourceCredits` with a `clearSourceCredits` flag. That's a better resolution than the assert I suggested: an assert would have turned `copyWith(sourceRelayHint: …)` on a merged clip from a silent 2-credit loss into a runtime failure, whereas deleting the params makes the mistake unexpressible. The inline comment explains the reasoning for whoever tries to add them back.

Worth noting for anyone reading the diff cold: this makes the scalars (`sourceAuthorPubkey` and friends) strictly a read-only view of `sourceCredits.first`, which is the invariant the rest of the PR already assumed. The model is now internally consistent with that.

The things I checked and liked in the first pass still hold on this revision — single-source publishes remain byte-identical to `main` via the `emittedAddressableCredits` seed and the `existingTags` dedup, and the order-independent `_addClipSourceCredit` round-trip is still pinned.

**Nothing blocking from me.** My `CHANGES_REQUESTED` is stale and needs a maintainer to convert. Item 3 from the original review ("clip credits shadow an explicitly chosen Inspired by") was raised as a question rather than a change request — treat that as still open for your judgment, not as a gate.

rabble (approved):

NotThatKindOfDrLiz:
Rebased onto current main and fixed the merge-queue blocker.

What changed:
- moved clip-source a-tag construction into inspired_by_tags.dart
- moved mention p-tag construction out of the oversized VideoEventPublisher
- kept the published tag output unchanged while satisfying the service god-file ratchet

Verification:
- bash scripts/check_service_god_file_ceiling.sh
- flutter analyze lib/services/video_event_publisher.dart lib/utils/collaborator_tags.dart lib/utils/inspired_by_tags.dart test/services/video_event_publisher_inspired_by_tags_test.dart
- flutter test test/services/video_event_publisher_inspired_by_tags_test.dart
- flutter analyze lib test

CI is rerunning on fec5ffbbb0.

Issue #6587, opened 2026-08-02
fix(ci): iOS build number ignores App Store Connect and regressed below shipped builds


rabble:
## Summary

iOS build 389 (`1.0.19`) uploaded to App Store Connect successfully and then vanished — Codemagic waited 15 minutes for it to appear, reported *"the corresponding build is not available in App Store Connect"*, and the run finished as `post-processing failed`.

The binary was fine. Apple rejected it during processing because its build number did not exceed the last one accepted for the version train.

## Root cause

`codemagic.yaml` derives the iOS build number from Codemagic's own counter plus a fixed offset:

```bash
# Offset to continue from existing manual builds
BUILD_NUMBER=$((PROJECT_BUILD_NUMBER + 142))
```

It reads neither `pubspec.yaml` nor App Store Connect. `--build-number` overrides pubspec, so the repo's version string has no influence on what actually ships.

App Store Connect state as of today:

```
marketing 1.0.19: builds ['819']    ← uploaded 2026-07-27
marketing 1.0.18: builds ['818']
marketing 1.0.17: builds ['815', '813', '812', '809', '808']
```

`app-store-connect get-latest-build-number 6747959501 --include-version` → `Version: 1.0.19, Build number: 819`.

Build 389 + offset 142 = **531**, which is below every build shipped since June. Apple accepts the upload, rejects it in processing, emails the account holder, and the build never appears.

The offset cannot have produced 819 — that would need `PROJECT_BUILD_NUMBER` = 677, and the counter only increases. 819 came from elsewhere, which is exactly what the *"continue from existing manual builds"* comment describes. Once builds arrive from any other source, the offset is wrong and nothing corrects it.

This already surfaced once: #6470 resynced `pubspec.yaml` and the Xcode project from 508 → 819 because they had drifted from TestFlight. That fixed the recorded numbers but not the mechanism generating them, so the drift resumed immediately.

## Why it keeps happening

The build number has no single source of truth. Three places hold a version of it and none of them agree for long:

- `pubspec.yaml` / `project.pbxproj` — corrected by hand, ignored by CI
- Codemagic's counter + 142 — what actually ships
- App Store Connect — the only one Apple cares about

## Fix

Ask the store, exactly as the Android path in the same file already does:

```bash
LATEST_GOOGLE_PLAY_BUILD_NUMBER=$(google-play get-latest-build-number --package-name "co.openvine.app" 2>/dev/null || true)
[...]

PR #6929, opened 2026-08-09
fix(ios): derive extension versions from Flutter

### LINKED ISSUE #3344: docs(release): reconcile version, package id, and OpenVine/Divine metadata drift
Source audit context: fresh `origin/main` worktree `.worktrees/audit-codebase-2026-04-24` at commit `33f6e026a`.
This issue was generated from the April 2026 codebase audit and includes file/line evidence from that snapshot.

## Severity
Medium

## Problem
Release metadata and docs mix OpenVine and Divine naming, stale release versions, and inconsistent platform version values.

## Evidence
- mobile/pubspec.yaml:1 package name is openvine.
- mobile/pubspec.yaml:2 describes divine.
- mobile/pubspec.yaml:19 version is 1.0.9+498.
- zapstore.yaml:2 release source still points to tag 1.0.8.
- mobile/ios/Runner.xcodeproj/project.pbxproj:643 and :895 use MARKETING_VERSION = 1.0.0.
- mobile/ios/Runner.xcodeproj/project.pbxproj:644 and :896 plus mobile/macos/Runner/Configs/AppInfo.xcconfig:11 use bundle id co.openvine.app.
- docs/LOGGING.md:6 still titles the system OpenVine Logging System.

## Impact
Release automation, store metadata, support docs, and developer onboarding can disagree about what app/version/package is authoritative.

## Recommended fix
Define canonical naming/version/package-id policy, then update release metadata and docs or explicitly document legacy package identifiers.

## Acceptance criteria
- [ ] Zapstore release source matches current intended app release.
- [ ] iOS/macOS marketing/build versions are generated from pubspec.yaml or documented as intentionally separate.
- [ ] Docs use Divine naming except where openvine is a deliberate package/bundle legacy identifier.
- [ ] Release checklist validates app version, store metadata, and package IDs.

## Related issues
None found during targeted duplicate search.

NotThatKindOfDrLiz:
## Summary
- derive Runner and embedded iOS extension versions from Flutter build variables instead of hardcoded Xcode literals
- add a CI guard for shipping iOS target version configuration drift
- fail Codemagic iOS builds when App Store Connect build-number lookup fails instead of guessing
- document Divine/openvine naming and identifier policy in the release checklist

Closes #3344

## Verification
- flutter pub get
- bash scripts/check_ios_shipping_versions.sh
- bash scripts/check_native_transport_security.sh
- IOS_EXTENSION_BUNDLE_IDS='co.openvine.app.NotificationServiceExtension co.openvine.app.CameraQuickActionWidget' bash scripts/check_ios_extension_signing_contract.sh
- ruby -e 'require "yaml"; YAML.load_file("codemagic.yaml"); YAML.load_file(".github/workflows/mobile_ci.yaml")'
- plutil -lint ios/Runner.xcodeproj/project.pbxproj ios/Runner/Info.plist ios/NotificationServiceExtension/Info.plist ios/CameraQuickActionWidget/Info.plist
- xcodebuild -showBuildSettings confirmed Runner, NotificationServiceExtension, and CameraQuickActionWidget resolve CURRENT_PROJECT_VERSION/MARKETING_VERSION from FLUTTER_BUILD_NUMBER/FLUTTER_BUILD_NAME, including FLUTTER_BUILD_NUMBER=999 override


rabble (approved):
Approving. I verified the derivation against what Codemagic actually injects at build time (not against `pubspec.yaml`), and the chain holds end to end.

## The bug this fixes is live on `main` right now

`mobile/ios/Runner/Info.plist:47,51` already reads the host app's versions from Flutter:

```xml
<key>CFBundleShortVersionString</key><string>$(FLUTTER_BUILD_NAME)</string>
<key>CFBundleVersion</key><string>$(FLUTTER_BUILD_NUMBER)</string>
```

Both extension plists read the *Xcode* settings instead — `mobile/ios/NotificationServiceExtension/Info.plist` and `mobile/ios/CameraQuickActionWidget/Info.plist` use `$(MARKETING_VERSION)` / `$(CURRENT_PROJECT_VERSION)`, which on `main` are hardcoded to `1.0.19` / `819` in six `project.pbxproj` build configurations.

`codemagic.yaml:268` builds with `flutter build ipa --release --build-number=$BUILD_NUMBER`, and `$BUILD_NUMBER` is `app-store-connect get-latest-build-number + 1` — so `FLUTTER_BUILD_NUMBER` is **820+**, while the two embedded extensions still stamp **819**. That is a CFBundleVersion mismatch between the host app and its extensions, which Apple rejects at upload validation (ITMS-90473). `MARKETING_VERSION` happened to agree at `1.0.19`, so only the build number diverges — but one is enough.

So this is not a hygiene change. Cutting 1.0.19 from `main` *without* it means submitting an IPA whose extension bundle versions disagree with the app.

## Verification I did

**The derivation chain resolves.** `mobile/ios/Flutter/AppVersion.xcconfig` does `#include? "Generated.xcconfig"` (resolved relative to `mobile/ios/Flutter/`, correct) and `flutter build ipa` always regenerates that file with `FLUTTER_BUILD_NAME` / `FLUTTER_BUILD_NUMBER`. Runner reaches the same variables through `Flutter/Debug.xcconfig` and `Flutter/Release.xcconfig`, both of which carry a non-optional `#include "Generated.xcconfig"` — so if that file were ever missing, the Runner target fails loudly before an extension could silently stamp an empty version string. The optional `#include?` in the new file is safe for that reason.

**No CocoaPods collision from the new base configuration.** `mobile/ios/Podfile` declares only `target 'Runner'` (and nested `RunnerTests`); neither extension is a CocoaPods target, so giving them `baseConfigurationReference = AppVersion.xcconfig` cannot displace a `Pods-*.xcconfig` they were relying on. They previously had no base configuration at all.

**The new guard actually works.** I ran `mobile/scripts/check_ios_shipping_versions.sh`'s Ruby body against three trees:

- branch `project.pbxproj` → passes, exit 0
- `origin/main` `project.pbxproj` → fails with 24 specific findings (all six Runner config/setting pairs plus all nine extension configs), exit 1
- branch tree mutated to re-add an inline `MARKETING_VERSION = 9.9.9;` to `CameraQuickActionWidget` Debug → `CameraQuickActionWidget Debug must inherit MARKETING_VERSION instead of setting it inline.`, exit 1
- branch tree mutated to drop one extension's `baseConfigurationReference` → `CameraQuickActionWidget Debug must use AppVersion.xcconfig as its base configuration.`, exit 1

It catches both regression shapes and fails for the right reason. The `mobile_ci.yaml:91` change also correctly adds the script's own path to the `native` filter so the gate can't drift silently.

## Non-blocking notes

**1. The ASC lookup now swallows the reason it failed.** `codemagic.yaml:252` is unchanged:

```sh
LATEST_IOS_BUILD_NUMBER=$(app-store-connect get-latest-build-number "$APP_STORE_APP_ID" 2>/dev/null || true)
```

Under the old code that discarded stderr on a path that continued anyway. Now the same discard sits in front of a hard `exit 1`, so a release build can die with `ERROR: could not read the latest App Store Connect build number.` and nothing about *why* — expired API key, rate limit, ASC outage all look identical in the log. Worth capturing stderr and echoing it before exiting.

I agree with removing the `PROJECT_BUILD_NUMBER + 142` floor, for what it's worth: per the comment it had drifted to 531 against a shipped 819, so on the failure path it was guaranteed to produce a number Apple would reject anyway. Failing fast is strictly better than burning a build and an emailed rejection.

**2. Scope creep into a release-path PR.** The `mobile/pubspec.yaml` description casing change (`divine` → `Divine`) and the new "Naming And Identifiers" section in `docs/RELEASE_CHECKLIST.md` are unrelated to deriving extension versions. Neither is risky — the pubspec `description` isn't shipped anywhere — but `.claude/rules/code_style.md` ("PR Scope") asks for these to be split, and it matters more than usual on a PR that's going to be reviewed under time pressure before a submission.

**3. `docs/RELEASE_CHECKLIST.md` still says "Version/build from `mobile/pubspec.yaml`"** in the iOS section. That was already only half-true and stays half-true: the version name does come from pubspec, but the build number is derived by Codemagic from App Store Connect and the pubspec's `+819` is ignored on that path. Since this PR is the one making the pbxproj stop lying about versions, it's a natural place to make the checklist stop lying too.

## Release safety: land this **before** the 1.0.19 submission

Current `main` will produce an IPA with mismatched extension build numbers, so shipping without this means either an upload rejection or a manual pbxproj bump that has to be guessed to match whatever ASC hands back — which is exactly the failure mode #6587 documented.

One caveat: **no CI job builds iOS**, so nothing here has been validated by an actual archive. The static guard proves the pbxproj shape, not that `xcodebuild` resolves `$(FLUTTER_BUILD_NUMBER)` through the new xcconfig at archive time. Before the store submission, please run one Codemagic iOS build off this branch (or off `main` after merge) and check the resulting `Payload/Runner.app/PlugIns/*.appex/Info.plist` — `CFBundleVersion` in both extensions must equal the app's. A TestFlight upload that passes validation is the real proof.

hm21 (approved):
LGTM

NotThatKindOfDrLiz:
Pushed follow-up commit d40cf959d.

Changes:
- Captures App Store Connect lookup stderr in the Codemagic iOS build step and prints it before failing fast.
- Updates the release checklist so iOS version name/build-number sources are accurate.
- Documents the intentionally separate macOS direct-download version path for #3344.
- Removes the dangling bare comment in codemagic.yaml.

Local verification after rebasing onto current origin/main:
- bash scripts/check_ios_shipping_versions.sh
- bash scripts/check_native_transport_security.sh
- IOS_EXTENSION_BUNDLE_IDS="co.openvine.app.NotificationServiceExtension co.openvine.app.CameraQuickActionWidget" bash scripts/check_ios_extension_signing_contract.sh
- ruby -e 'require "yaml"; YAML.load_file("codemagic.yaml"); YAML.load_file(".github/workflows/mobile_ci.yaml")'
- git diff --check

PR #6932, opened 2026-08-09
ci(architecture): ratchet service god-file growth

NotThatKindOfDrLiz:
## Summary

- add a hard service-layer god-file size ratchet for `mobile/lib/services` files over 1500 lines
- freeze the current oversized service files in a numeric ceiling baseline
- wire the guard into mobile CI while leaving the broader 800-line file-size check advisory
- add tool tests for baseline generation, growth/new/stale failures, shrink passes, and raised-baseline bypass protection
- document the new hard ratchet in `AGENTS.md` and decouple anti-bypass test coverage from the live advisory file-size baseline
- clean up `popular_videos_tab_test.dart` so the held classic-feed future completes after the slow-switch assertion instead of leaking a pending timer across the merged VGV isolate

Refs #4338.
Refs #4339.

## Verification

- `cd mobile && bash scripts/check_service_god_file_ceiling.sh`
- `cd mobile && flutter test test/tools/service_god_file_ceiling_test.dart`
- `cd mobile && flutter analyze test/tools/service_god_file_ceiling_test.dart`
- `cd mobile && flutter test test/tools/file_size_ceiling_test.dart test/tools/service_god_file_ceiling_test.dart`
- `cd mobile && flutter analyze lib test integration_test`

Pre-push hook also reran analyzer and the changed tests:

- `test/tools/service_god_file_ceiling_test.dart`
- `test/widgets/popular_videos_tab_test.dart`


NotThatKindOfDrLiz:
CI follow-up: the first run exposed a pending Timer in `test/widgets/popular_videos_tab_test.dart` after the new test changed shard timing. This branch now completes the held classic-feed future after the slow-switch assertion and verifies it with `flutter test test/widgets/popular_videos_tab_test.dart --plain-name "PopularVideosTab reports a slow variant switch once, not once per rebuild"`. I also reproduced shard 2 locally with CI seed `1165003965`: `very_good test --optimization --concurrency=4 --exclude-tags integration --test-randomize-ordering-seed 1165003965` passed after selecting shard 2.

rabble (approved):
Approving — this matches the established ratchet conventions closely, and I verified the baseline is correct rather than taking it on faith.

## Verified

- **Baseline is exact vs `origin/main`.** Recomputed every non-generated `.dart` under `mobile/lib/services` at `origin/main` with the same threshold; the >1500 set is exactly the five baselined files at exactly the recorded counts (`auth_service` 4769, `curated_list_service` 1834, `upload_manager` 2310, `video_event_publisher` 2316, `video_event_service` 6619). No missing key, so the guard won't fail on merge with a spurious `NEW key`.
- **Conventions match.** Sources the shared `lib/numeric_ratchet.sh` engine rather than reimplementing, baseline under `mobile/scripts/baseline/`, `UPDATE_BASELINE=1` documented in both the script header and the generated baseline header, shrink-only with the branch-baseline-vs-base-ref bypass check inherited from the engine. Same generated-code exclusions as `check_file_size_ceiling.sh`.
- **Counting is appropriate.** `wc -l` raw lines is the right unit here and is consistent with the sibling `check_file_size_ceiling.sh`. `dart_code_only.awk` is for the token-detector ceilings (raw colors, `TextStyle`, Material buttons, dialogs) where a comment or log string could otherwise trip the guard; a *file-size* ratchet legitimately counts every line, and stripping comments would make "shrink the file" satisfiable by deleting documentation. No change wanted.
- **Tests cover the real failure modes** — pass, growth, new key, stale, shrink-still-oversized, and the raised-baseline bypass — and drive the real script against a temp tree with an overridden threshold rather than asserting on the production baseline.

## Concern worth an explicit answer before merge (not blocking from me)

There is no escape hatch, at all, for a legitimate line-adding change to one of these five files. The bypass check fails on any raised ceiling vs `origin/main`, so a three-line fix in `auth_service.dart` (4769 lines, the auth core) or `video_event_service.dart` (6619) can only ship by shrinking the same file elsewhere in the same PR or extracting first.

That is the stated intent, and the header comment shows you thought about it — but it's worth noting it re-imposes on the five hottest service files exactly the friction the team removed when it downgraded the 800-line check to advisory ("a small edit to an already-large file no longer forces an unrelated refactor in the same PR"). The other numeric ratchets have the same no-raise semantics, so this is consistent with convention rather than novel; the difference is that those have a cheap sanctioned alternative (use `VineTheme`, use `DivineButton`), whereas the alternative here is "extract a responsibility", which is not a hotfix-sized ask.

Not asking you to change it — the epic wants the pressure. But please make sure whoever owns #4338 has explicitly signed off on "a hotfix touching auth_service must shrink it or extract", because the first time it bites will be under time pressure.

## Nits

1. **Unrelated file in the diff.** `mobile/test/widgets/popular_videos_tab_test.dart` (the `Completer` that now completes and settles instead of hanging) has nothing to do with the ratchet and isn't mentioned in the PR description. It looks like a correct test-hygiene fix — a never-completing future left pending across a merged VGV isolate is exactly the kind of cross-suite bleed `testing.md` warns about — so I'm not asking you to pull it out, but it should be called out in the description rather than discovered in the diff.
2. **`fails when the branch baseline raises a ceiling vs base ref` is quietly coupled to unrelated state.** It points `SERVICE_GOD_FILE_BASELINE_REPO_PATH` at the real `mobile/scripts/baseline/file_sizes.txt` at `HEAD` and relies on `lib/services/auth_service.dart` being present there with a value below 9999. If `auth_service.dart` ever drops under 800 lines and leaves that advisory baseline, this test flips from `^raised` to `+added` and fails for a reason unrelated to what it's testing. A tiny committed fixture baseline under `test/tools/` would decouple it. Non-blocking.

NotThatKindOfDrLiz:
Follow-up from the no-escape-hatch concern in rabble's review: I rechecked the current open PR impact after the rebase. The guard is behaving as intended, but it will immediately affect active PRs that add lines to newly frozen service files.

| PR | file | delta in PR | expected result after this lands |
| --- | --- | ---: | --- |
| #6661 | `mobile/lib/services/curated_list_service.dart` | +466/-98 | fails unless it shrinks/extracts enough to stay at or below 1834 lines |
| #6928 | `mobile/lib/services/video_event_publisher.dart` | +35/-0 | fails unless it shrinks/extracts enough to stay at or below 2316 lines |
| #6856 | `mobile/lib/services/video_event_publisher.dart` | +15/-0 | fails unless it shrinks/extracts enough to stay at or below 2316 lines |

#5727 also touches `video_event_publisher.dart`, but it is draft and already conflicting, so I am treating the three non-draft PRs above as the actionable blast radius.

@rabble please explicitly sign off that this is the intended #4338 policy cost before merge: a hotfix or feature touching one of the five frozen service god-files must either shrink that same file or extract first; raising the ceiling is blocked by the ratchet.

mobile/scripts/baseline/service_god_file_sizes.txt
```diff
@@ -0,0 +1,11 @@
+# Frozen baseline: Dart files under mobile/lib/services over 1500 lines,
+# each with its current line count as a CEILING (format: relpath<TAB>loc).
+# Generated by scripts/check_service_god_file_ceiling.sh. A ceiling may only
+# SHRINK; more lines, a new oversized service file, or a raised ceiling fails CI
+# vs origin/main. Epic: #4338.
+# Regenerate after shrinking/removing: UPDATE_BASELINE=1 bash scripts/check_service_god_file_ceiling.sh
+lib/services/auth_service.dart	4769
+lib/services/curated_list_service.dart	1834
+lib/services/upload_manager.dart	2310
+lib/services/video_event_publisher.dart	2316
```
realmeylisdev: **before-merge:** worth knowing the immediate blast radius. three open PRs already sit over these ceilings, so they go red the moment this lands:

- #6928 `video_event_publisher.dart` 2316 to 2351
- #6856 `video_event_publisher.dart` 2316 to 2331
- #6661 `curated_list_service.dart` 1834 to 2202

i took those from each PR's `refs/pull/N/merge` and cross-checked them against the net diff delta applied to current main; both agree. (#5727 also net-adds 43 lines to `video_event_publisher.dart`, but its merge ref looks stale so i'm not putting a number on it.)

none of the three can be fixed by editing somewhere else, the only exits are shrink-the-same-file or extract first. that is the pressure the epic wants, i just don't think it should land on three in-flight branches by surprise, one of them 368 lines over. either give those authors a heads-up or land this after they merge.

AGENTS.md
```diff
@@ -135,6 +135,7 @@ If a Divine Brain search or ask tool is available, you may use it for company me
...
+- Oversized service-layer god-files under `mobile/lib/services/` are frozen by a hard per-file line-count ceiling ratchet (#4338): `check_service_god_file_ceiling.sh`, baseline `mobile/scripts/baseline/service_god_file_sizes.txt`. The guard tracks service files over 1500 lines; a baselined file growing, a new oversized service file, or a raised ceiling fails CI, and there is no sanctioned raise path. Keep new work out of the frozen god-files by extracting responsibilities behind the existing `UI -> BLoC/Cubit -> Repository -> Client` boundaries, or shrink the touched service in the same PR. After intentionally shrinking/removing one of these files, lock the win with `UPDATE_BASELINE=1 bash mobile/scripts/check_service_god_file_ceiling.sh`, then commit the updated baseline.
```
realmeylisdev: **should-fix:** this lists growth, a new oversized file, and a raised ceiling as the CI failures, but not the stale case, and stale is the one that hits the *good* path. shrinking one of these five below 1500, deleting it, or renaming it all fail CI until the baseline is regenerated. your own `fails stale when a baselined service drops below the threshold` test pins that.

it matters because the sentence right before recommends exactly that route: "or shrink the touched service in the same PR". someone who follows it and crosses 1500 gets a red build with nothing here warning them.

the design-system bullet just below already words it correctly ("a file dropping to zero (regenerate to lock the win) fails CI"). matching it would do, something like: "...or a raised ceiling fails CI, as does a baselined file dropping below the threshold, being renamed, or being deleted (regenerate to lock the win)".

.github/workflows/mobile_ci.yaml
```diff
@@ -195,6 +195,12 @@ jobs:
           bash scripts/check_future_delayed_production_ceiling.sh
       - name: 🧱 Verify untested-services count does not increase (floor ratchet)
         run: bash scripts/check_untested_services_floor.sh
+      - name: 🧱 Verify service god-files do not grow (architecture ratchet)
+        run: |
+          # Hard architecture guard for #4338: the broad 800-line file-size
+          # check remains advisory, but oversized service-layer god-files are
+          # frozen so extraction gains cannot be silently clawed back.
```
realmeylisdev: **nit:** "frozen so extraction gains cannot be silently clawed back" claims a bit more than the engine does. a ceiling only tightens when someone runs `UPDATE_BASELINE`, and while the file is still oversized nothing ever prompts them to.

i checked on a copy of the real tree: shrink `curated_list_service.dart` from 1834 to 1600 and the guard prints `OK` with no mention that 234 lines of headroom now exist, and a later PR can put all 234 back and it still prints `OK`. so the freeze genuinely holds at the *recorded* number, but a gain made after the freeze is precisely what can be clawed back silently.

either soften the wording, or (nicer, though it's a shared-engine change so your call) have `numeric_ratchet.sh` add a line on success when any key is under its ceiling: "N key(s) below their ceiling, relock with UPDATE_BASELINE=1".

mobile/scripts/check_service_god_file_ceiling.sh
```diff
@@ -0,0 +1,74 @@
...
+SCAN_DIR="${SERVICE_GOD_FILE_SCAN_DIR:-$MOBILE_DIR/lib/services}"
+PATH_PREFIX="${SERVICE_GOD_FILE_PATH_PREFIX:-$MOBILE_DIR}"
+BASELINE_FILE="${SERVICE_GOD_FILE_BASELINE_FILE:-$SCRIPT_DIR/baseline/service_god_file_sizes.txt}"
+BASELINE_REPO_PATH="${SERVICE_GOD_FILE_BASELINE_REPO_PATH:-mobile/scripts/baseline/service_god_file_sizes.txt}"
+BASE_REF="${SERVICE_GOD_FILE_BASELINE_BASE_REF:-origin/main}"
+ALLOW_NO_BASE="${SERVICE_GOD_FILE_CEILING_ALLOW_NO_BASE:-0}"
+ALLOW_NO_BASE_VAR="SERVICE_GOD_FILE_CEILING_ALLOW_NO_BASE"
+
+NEW_HINT="Do not grow service-layer god files under mobile/lib/services. Extract responsibilities behind repository/client boundaries, or keep the change out of the oversized service file. See epic #4338."
+STALE_HINT="A service god-file was removed or dropped below the oversized threshold."
```
realmeylisdev: **follow-up, not blocking:** this hint misfires when a god-file is *moved* rather than shrunk. moving `curated_list_service.dart` into `lib/services/curated/` with zero line changes fails twice, NEW on the new path plus STALE on the old one. following the advice printed here and running `UPDATE_BASELINE=1` then fails a third way, `+added` from the bypass check against origin/main, so there's no way through short of getting the file under 1500 first.

moving it out of `lib/services/` altogether is fine (stale only, regen accepts it), it's in-tree moves that are stuck. all inherited from the shared engine so nothing for you to fix in this PR, but since #4338 is partly about restructuring this directory it's worth being aware of, and the hint is actively wrong in that case.

realmeylisdev (changes requested):
ratchet itself is solid. baseline is exact against main, and i drove the script on a copy of the real tree to confirm it actually fails on growth rather than just looking like it would. two things worth settling before it lands though.

three open PRs already blow these ceilings the day this merges (#6928, #6856, #6661) and none of them can go green without shrinking the file they touch. and the AGENTS.md bullet leaves out that shrinking one of these five below 1500 turns CI red until you regen the baseline, which is a rough surprise on the good path.

rest is inline, and the two behaviour notes are engine-level so mostly fyi.

mobile/scripts/baseline/service_god_file_sizes.txt
NotThatKindOfDrLiz: Refreshed after the rebase. #6856 has merged into main, so the remaining live blast radius is #6661 on curated_list_service.dart and #6928 on video_event_publisher.dart. The baseline is regenerated to current main counts now: auth_service 4780, curated_list_service 1848, upload_manager 2310, video_event_publisher 2331, video_event_service 6619. Keeping rabble sign-off as the policy gate before merge.

AGENTS.md
NotThatKindOfDrLiz: Fixed in the rebased AGENTS.md bullet. It now calls out that a baselined file growing, a new oversized service, a raised ceiling, or a baselined file dropping below threshold / being renamed / being deleted fails CI, and that intentional shrink/removal needs UPDATE_BASELINE to lock the win.

NotThatKindOfDrLiz:
@realmeylisdev please re-review. @rabble the remaining policy gate is still your explicit sign-off that #4338 should block hotfix/feature growth in these frozen service files unless the same PR shrinks or extracts first.

Updated in f9d943a67:

- rebased onto current origin/main, dropping the merge commit and clearing the AGENTS.md conflict
- regenerated service_god_file_sizes.txt to current main counts: auth_service 4780, curated_list_service 1848, upload_manager 2310, video_event_publisher 2331, video_event_service 6619
- documented the stale/drop/rename/delete failure path in AGENTS.md during rebase conflict resolution
- softened the CI comment to match recorded-ceiling behavior
- fixed the service ratchet hint for in-tree moves of still-oversized files and filed #7103 for shared numeric_ratchet.sh rename handling
- decoupled raw_icons_ceiling_test.dart from the live advisory file_sizes.txt baseline by reusing the committed fixture baseline

Verification:

- cd mobile && bash scripts/check_service_god_file_ceiling.sh
- cd mobile && flutter test test/tools/service_god_file_ceiling_test.dart
- cd mobile && flutter test test/tools/raw_icons_ceiling_test.dart
- cd mobile && flutter analyze test/tools/service_god_file_ceiling_test.dart test/tools/raw_icons_ceiling_test.dart
- pre-push hook: merge-conflict check, analyzer, and changed-file tests for raw_icons_ceiling_test.dart, service_god_file_ceiling_test.dart, and popular_videos_tab_test.dart

Current refreshed blast radius: #6856 is merged; #6661 and #6928 remain the live PRs that will need shrink/extract work if this policy lands.

mobile/scripts/baseline/service_god_file_sizes.txt
```diff
@@ -0,0 +1,13 @@
...
+# Generated by scripts/check_service_god_file_ceiling.sh. A ceiling may only
+# SHRINK; more lines, a new oversized service file, or a raised ceiling fails CI
+# vs origin/main. Epic: #4338.
+# This hard service baseline is authoritative over the advisory file_sizes.txt
+# baseline for service god-file ceilings.
+# Regenerate after shrinking/removing: UPDATE_BASELINE=1 bash scripts/check_service_god_file_ceiling.sh
+lib/services/auth_service.dart	4780
+lib/services/curated_list_service.dart	1848
+lib/services/upload_manager.dart	2310
+lib/services/video_event_publisher.dart	2331
```
realmeylisdev, line 13: **before-merge:** refreshed the blast radius against main at 3d8a4601. three PRs still go red the day this lands, and one of them changed shape since your update:

- #6928 `video_event_publisher.dart` 2331 -> 2365 (growth)
- #6661 `curated_list_service.dart` 1848 -> 2242 (growth)
- #5727 `video_event_publisher.dart` 2331 -> 2374 (growth) **and** `video_editor_render_service.dart` 1405 -> 1504 (**NEW key**)

the first two are covered, #6928 is yours and #6661 is rabble's who approved here. #5727 is the one nobody in this thread owns, and it's the only NEW-key case: `video_editor_render_service.dart` isn't baselined, so there's no ceiling to shrink back to and no sanctioned raise. it lands 4 lines over 1500.

numbers are head-vs-merge-base deltas applied to current main (head 1279, merge-base 1180, main 1405 for the render service). #5727 has conflicts right now so it has to rebase first and the exact figure will move. but that file sits 95 lines under the threshold on main and is being actively grown, so it lands on the wrong side either way.

not asking you to change the ratchet, the epic wants this pressure. worth a heads-up to hm21 though, since a NEW-key failure on an unbaselined file is the one shape with no shrink-the-file-you-touched exit.

mobile/test/tools/fixtures/service_god_file_base_sizes.txt
```diff
@@ -0,0 +1,2 @@
+# Fixture baseline for service_god_file_ceiling_test.dart anti-bypass coverage.
+lib/services/auth_service.dart	6
```
realmeylisdev, line 4: **nit:** good change off the live `file_sizes.txt`, but this fixture now backs two unrelated ratchet tests while being named for one of them.

`service_god_file_ceiling_test.dart` needs `lib/services/auth_service.dart<TAB>6` present here (its `^raised` assertion diffs 9999 against that value), and `raw_icons_ceiling_test.dart` needs `lib/uses.dart` **absent** (that absence is what makes its `+added` assertion fire). i checked both directions actually hold, so each test depends on a different property of this one file and an edit made for one can quietly break the other.

a neutral name like `ratchet_base_sizes.txt`, or a second fixture for raw_icons, keeps them independent. non-blocking.

realmeylisdev (approved):
all four of my earlier items are addressed: the stale case is in the AGENTS.md bullet, the CI comment no longer overclaims what the freeze holds, the in-tree-move hint now says what UPDATE_BASELINE can't approve, and the anti-bypass tests are off the live baseline onto a committed fixture.

re-verified rather than taking it on faith. the baseline is exact against main at 3d8a4601, the trial merge is clean despite the branch being behind, and the guard still passes when run against the merged tree. i also mutated the fixture and confirmed both anti-bypass tests actually go red, so they aren't passing for free.

one thing worth a look before you merge, in the baseline thread: the blast radius shifted after the rebase. #5727 is the interesting one now, it trips the guard twice and one of those is a NEW key rather than growth, which is the case with no shrink-back target.

approving. the remaining note is coordination, not a defect, and the fixture one is a nit.

mobile/test/tools/fixtures/service_god_file_base_sizes.txt
hm21: Fixed in da5f4a2b4 — split into two fixtures instead of renaming, so neither test can defuse the other. `raw_icons_ceiling_test.dart` now points at its own `raw_icons_base_sizes.txt`, and both headers state the property their test depends on.

Verified by mutating each fixture and confirming only its own test goes red.

hm21 (approved):
LGTM. Pushed one fix (da5f4a2b4) for the open fixture nit: the two anti-bypass tests now have separate base fixtures instead of sharing one, so an edit for one can no longer silently defuse the other. Mutation-checked both directions; CI green.

Re-verified the baseline is exact against current `origin/main` (b41c4c51b) — all five entries match. The remaining `#5727` note is coordination, not a defect here; I confirmed the numbers and will handle it outside this PR.

Issue #2547, opened 2026-03-27
Define categories


d-spurgin:

realmeylisdev:
@NotThatKindOfDrLiz researched the two open questions — recommendations below, evidence-backed.

**1. Canonical list → funnelcake backend, not the clients.** Add a curated `CANONICAL_CATEGORIES` const mirroring the existing `CANONICAL_MODERATION_LABELS` (`queries.rs:184`), enforced in `get_categories` (`client.rs:9865`). Categories feed Gorse/recsys/trending server-side before any client sees them, so only a backend allow-list gates them all at once — mobile/web/discovery-admin stay presentation-only. (Mobile is the most-drifted consumer; there is no org precedent for a shared manifest.)

**2. `violence` is a moderation label, not a category → drop it from discovery.** It is already canonical in `CANONICAL_MODERATION_LABELS` (`queries.rs:194`) and default-hidden (`event_filters.rs:30`); no code emits `topic:violence`. Surfacing it for discovery contradicts hiding it.

Need: your sign-off, and any T&S call on whether `violence` should ever appear as a topic.

NotThatKindOfDrLiz:
Thanks, this matches my read.

Design/product sign-off:

- Canonical categories should live in funnelcake as a backend allow-list. Clients should be presentation-only consumers of that taxonomy, not independent sources of truth.
- `violence` should not be a browsable discovery category. Treat it as a moderation/T&S label and exclude it from category surfaces unless T&S explicitly asks for a different handling later.
- Proceed with the grouped taxonomy direction: top-level groups with raw names folded underneath. I’m good with engineering using the proposed strawman as the starting point and tightening the exact group names/membership during the illustration/copy pass.

Please keep Phase 0 crash-proofing unblocked and independent from the taxonomy work.

NotThatKindOfDrLiz:
Closing this as a stale design placeholder.

The design/product decision requested here was answered on 2026-06-06: canonical categories belong in funnelcake, clients stay presentation-only, `violence` is not a browsable discovery category, and grouped taxonomy is approved as the starting direction.

Concrete follow-up work is now split:

- divinevideo/divine-funnelcake#864 — backend canonical discovery category allow-list, synonym folding, and moderation/T&S exclusions.
- #6927 — urgent mobile safety guard to block moderation-adjacent discovery categories, including direct route/feed access.
- #6926 — mobile presentation catalog cleanup once the backend taxonomy is pinned.

[...]

Issue #864, opened 2026-01-09
feat: comments p0 design


ryzizub:
Reply UI Redesign

  Current: Reply input appears inline below comment
  Target: Reply input moves to bottom with green banner showing "Replying to {username}" with X button

  Implementation:
  - Move reply input from inline position to bottom input section
  - Add reply context banner above input field
  - Style banner with green background (VineTheme.vineGreen)
  - Add X button to cancel reply
  - Update CommentsBloc to manage reply context state

Reply Indication

  Current: No visual indication for replies
  Target: Shows "Re: @username" prefix on reply comments

  Implementation:
  - Add reply indicator text above comment content
  - Style with muted color
  - Extract parent author name for display

Options Menu Enhancement

  Current: Only "Delete Comment" for own comments
  Target:
  - Own comments: "Delete"
  - Others' comments: "Report" + "Block"

  Implementation:
  - Update comment_options_modal.dart to show different options based on ownership
  - Add Report option (flag icon, red)
  - Add Block option (block icon, red)
  - Implement report/block functionality

rafaeltheni-VGV:
Testing details: Qa Pass
Build: 1.0.4(150)(.abb) on Android 16
Device: Pixel 10 and Galaxy A54

PR #6937, opened 2026-08-09
fix(categories): block moderation-adjacent discovery categories

### LINKED ISSUE #6927: fix(categories): block moderation-adjacent discovery categories
## Problem

The live category API currently includes moderation-adjacent category names as browsable discovery cards. A fresh check on 2026-08-09 showed `adult` returned by `GET https://api.divine.video/api/categories?limit=200` with 319 videos, and `violence` returned with 212 videos. Both are explicitly signed off in #2547 / divinevideo/divine-funnelcake#864 as moderation/T&S labels, not browsable discovery categories.

While divinevideo/divine-funnelcake#864 adds the canonical backend allow-list, mobile should prevent these category names from becoming discovery entry points.

## Scope

Add an interim mobile deny-list for category discovery surfaces and direct category feed access.

## Requirements

- Filter denied category names from `CategoriesRepository.getCategories()` alongside the existing empty-name and zero-count filters.
- Also block direct category gallery/feed requests for denied names, because `/categories/:categoryName` can construct a `VideoCategory` from the path even when the card is hidden.
- Use named constants/classes; avoid inline string literals scattered through call sites.
- Start deny-list with `adult` and `violence`.
- Confirm product/T&S handling before adding borderline terms such as `accident` and `fighting`.

## Acceptance Criteria

- `adult` and `violence` do not appear as browsable category cards.
- Direct route/feed access for denied category names does not fetch category videos.
- Repository tests cover list filtering and direct feed blocking.
- Links back to #2547 and divinevideo/divine-funnelcake#864.

NotThatKindOfDrLiz:
## Summary

- add an exported `DiscoveryCategoryPolicy` for moderation-adjacent discovery categories
- filter denied category names out of category list caching and block direct category/recommendation fetches for denied names
- reject denied `/categories/:categoryName` routes before the category gallery renders

## Why

The live category endpoint can still return moderation/T&S labels such as `adult` and `violence` while divinevideo/divine-funnelcake#864 is open. Mobile also accepts direct category paths, so hiding list cards alone does not block access.

Closes #6927
Refs #2547
Refs divinevideo/divine-funnelcake#864

## Verification

- `flutter analyze`
- `flutter test --coverage` from `mobile/packages/categories_repository`
- `flutter test test/router/category_route_test.dart`


rabble (dismissed):
The normalization and the three enforcement points (`getCategories`, `getVideosForCategory`, `getRecommendedVideos`) look right, and the route guard correctly closes the deep-link path. One gap though, and because this is a content-safety change I'd rather it land closed than follow up.

## Blocking: the disk-cache read path bypasses the filter

`_isBrowsableCategory` only runs inside `getCategories()` — i.e. on the **network** path. `watchCategoriesCached()` passes that as `fetch:` to `CacheSync.watchOne(..., policy: CacheFetchPolicy.cacheAndNetwork)`, which emits `CacheResult.cached` **first**, deserialized by `_categoriesFromJson`:

```dart
static List<VideoCategory> _categoriesFromJson(String payload) {
  final decoded = jsonDecode(payload) as List<dynamic>;
  return decoded
      .map((item) => VideoCategory.fromJson(item as Map<String, dynamic>))
      .toList(growable: false);   // no filtering
}
```

`_categoriesCacheKey` is unchanged (`'global:categories:v1'`), so every device that has already cached the category list still has `adult` / `violence` on disk. The chain is live:

`CategoriesBloc._onLoadRequested` → `emit.forEach(watchCategoriesCached())` → `categories: result.data` → `CategoriesDiscoveryView` → tappable chips (`categories_tab.dart:38-42`).

Net effect on an existing install: opening Discover renders "Adult" and "Violence" chips from the stale disk payload until the background refresh lands and overwrites the cache. The route guard means tapping them dead-ends on `RouteErrorScreen` rather than browsing — so it's an inert entry point, not an open one — but the file's own ABOUTME says the goal is "keeps blocked labels from becoming browsable discovery entry points", and a rendered chip is exactly that.

Suggested fix — filter on the read path rather than only bumping the key, so it also covers a stale or hand-edited cache entry rather than being a one-shot migration:

```dart
static List<VideoCategory> _categoriesFromJson(String payload) {
  final decoded = jsonDecode(payload) as List<dynamic>;
  return decoded
      .map((item) => VideoCategory.fromJson(item as Map<String, dynamic>))
      .where((c) => !DiscoveryCategoryPolicy.isDenied(c.name))
      .toList(growable: false);
}
```

(`_isBrowsableCategory` isn't quite reusable as-is here — `videoCount > 0` is a network-shape concern — hence the direct policy call.) Worth a test in `watchCategoriesCached` that seeds the cache with a denied entry and asserts the `cached` emission drops it; that's the one that would have caught this.

## Non-blocking: how is `denied` kept complete?

`denied = {'adult', 'violence'}` is a hardcoded pair, while the repo already carries an enumerated moderation vocabulary in `mobile/lib/models/content_label.dart` — 21 `ContentLabel` values plus alias normalization (`nsfw`→`nudity`, `gore`→`graphic-media`, `weapon`→`violence`, …). Today only `violence` has a `categoryViolence` l10n key and a `violence.svg` glyph, so I believe those two are what Funnelcake actually emits right now, and I can't prove a miss from inside this repo.

But the maintenance question is real: if Funnelcake starts emitting `nsfw`, `porn`, `gore`, or `self-harm` as categories, they become browsable and nothing here notices. Two things would help:

1. A line in the doc comment saying where the authoritative list comes from and what triggers an update (a Funnelcake category-vocabulary change).
2. Confirmation that this is also enforced server-side. A client denylist is defense-in-depth; if Funnelcake still serves `/categories/adult` and `/videos?category=adult`, web and any other client remain exposed.

Normalization itself looks correct to me — Dart's `trim()` covers Unicode whitespace incl. NBSP, and `toLowerCase()` is locale-independent (no Turkish-İ hazard), so casing/whitespace variants are handled.

## Good

- Tests genuinely pin the policy: the fixtures name `' adult '`, `'Violence'`, `'ADULT'` explicitly, so removing an entry from `denied` turns them red. That's the property that matters for a denylist.
- `verifyNever` on the client in the denied-category tests confirms it short-circuits before the network call, not just after.
- The `still fetches recommendations when category is null` test is the right guard against the null-category branch regressing into a blanket block.

Happy to re-review as soon as the cached path is closed.

mobile/packages/categories_repository/lib/src/discovery_category_policy.dart
```diff
@@ -0,0 +1,16 @@
+// ABOUTME: Discovery category policy for moderation-adjacent category names.
+// ABOUTME: Keeps blocked labels from becoming browsable discovery entry points.
+
+/// Policy for category names that must not be exposed as discovery surfaces.
+abstract final class DiscoveryCategoryPolicy {
+  /// Category names that must never become discovery entry points.
+  ///
+  /// This mirrors Funnelcake's emitted discovery-category vocabulary. Update it
+  /// when Funnelcake adds or renames moderation-adjacent category slugs.
+  static const _denied = <String>{'adult', 'violence'};
```
realmeylisdev: should-fix: `porn` is missing, and it tags the same videos as `adult`.

Probed live funnelcake just now:

```
GET /api/videos?category=porn&limit=3
  3 items, each content_labels: ["topic:adult", "topic:porn", ...]
GET /api/videos?category=porn&limit=50
  39 items
```

That is the exact request `getVideosForCategory` issues (`funnelcake_api_client.dart:2400-2414`, lowercased `category` on `/api/videos`), and the route builder constructs `VideoCategory(name: 'porn', videoCount: 0)` for any slug `isDenied` does not catch. So the gallery renders the same content under the sibling slug.

It is not a visible card today: `/api/categories` returns 100 rows whatever `limit` you pass, and `porn` falls below the cut. Direct-path access is the case the route guard in this PR exists for, though, so the hole is in the same defense.

Neighbours from the same probe, in case it helps the T&S call: `nudity` 9, `sex` 5, `drugs` 3, `weapon` 1, and `nsfw` / `gore` / `self-harm` 0.

Two related facts from the same pass:

1. `violence` is not currently returned by `/api/categories` at any limit, while `adult` is (about 340 videos). Keeping `violence` is fine as defense, but `adult` is the entry actually doing work today.
2. divinevideo/divine-funnelcake#864 is still open and the live endpoint still emits `adult`, so this client list is the only gate right now.

realmeylisdev (changes requested):
one gap left. the deny list has `adult` but not `porn`, and funnelcake serves the same videos under both slugs, so `/categories/porn` still opens a gallery of what `/categories/adult` now blocks. evidence inline.

rest checked out: reproduced the cached-payload leak rabble flagged and confirmed `f217f45e` closes it, package tests and coverage are green, and on an iOS sim the categories list now runs Hospital, Adventure, Baseball with no Adult card between them.

NotThatKindOfDrLiz:
Addressed in f217f45e5.

What changed:
- Filtered denied category names in `_categoriesFromJson`, so stale or hand-edited disk cache payloads cannot render `adult` / `violence` discovery entries before the network refresh.
- Added a `watchCategoriesCached` regression test that seeds the disk cache with denied entries and asserts the cached emission drops them.
- Made the deny set private and added the doc-comment provenance note: this mirrors Funnelcake's emitted discovery-category vocabulary and should change when Funnelcake category slugs change.

On the metadata `CategoryChip` path: I left that out of this PR because it renders video metadata labels only. Those category chips are not tappable entry points; the sibling hashtag chips are the interactive ones.

On server-side enforcement: not landed yet. divinevideo/divine-funnelcake#864 is still open for the canonical Funnelcake category allow-list, including excluding `adult` and `violence` before categories or category videos reach clients. This mobile denylist is defense-in-depth until that lands.

Verification:
- `flutter test` from `mobile/packages/categories_repository`
- `flutter test test/router/category_route_test.dart` from `mobile`
- Pre-push hook reran analyze and changed-file tests
- GitHub checks are green on f217f45e5

mobile/packages/categories_repository/lib/src/discovery_category_policy.dart
realmeylisdev: Still open on `f217f45e` — re-probed live funnelcake just now rather than taking the earlier round's word for it:

```
GET /api/videos?category=porn&limit=50  → 39 items; 38 also carry topic:adult
GET /api/categories?q=porn              → {"name":"porn","video_count":39}
```

That first call is exactly what `getVideosForCategory` builds — `funnelcake_api_client.dart:2399-2414` lowercases `category` onto `/api/videos`. `isDenied('porn')` is false, so `search_routes.dart` constructs `VideoCategory(name: 'porn', videoCount: 0)`, the gallery renders, and back come the same 38 videos this PR just blocked under `adult`.

It isn't a visible card: `/api/categories` returns 100 rows whatever `limit` you pass, and `porn` sits below the cut. But direct-path access is the case the route guard in this PR exists for, so the hole is inside the same defense the PR is building.

One data point that argues for keeping `violence` even though it never shows up in the unfiltered list: `?q=violence` returns 1422 videos. It's real content that just ranks below the cut, same as `porn`.

Neighbours from the same pass, for the T&S call rather than this PR: `nudity` 9, `sex` 5, `drugs` 3, `weapon` 1; `nsfw` / `gore` / `self-harm` all 0.

mobile/packages/categories_repository/lib/src/discovery_category_policy.dart
```diff
@@ -0,0 +1,16 @@
+// ABOUTME: Discovery category policy for moderation-adjacent category names.
+// ABOUTME: Keeps blocked labels from becoming browsable discovery entry points.
+
+/// Policy for category names that must not be exposed as discovery surfaces.
+abstract final class DiscoveryCategoryPolicy {
```
realmeylisdev, line 5: This class is transitional by its own description — it exists while divinevideo/divine-funnelcake#864 is open, and #864's acceptance criteria says in as many words that clients should stop owning this list. `.claude/rules/code_style.md` ("Temporary Code") asks for a `// TODO(#issue):` naming the removal tracker for exactly this shape. The in-repo precedent is the same situation:

```dart
// TODO(#6649): Re-promote after divine-funnelcake#691 ships.
```

The doc comment below explains how to *maintain* the list but not when it goes away, which is the half the rule is aimed at. A line pointing at #864 would cover it.

realmeylisdev (changes requested):
Re-reviewed on `f217f45e`.

**Closed since last round.** rabble's cached-payload leak is genuinely fixed, and fixed the right way — `_categoriesFromJson` filters on the read path rather than bumping the cache key, so a stale or hand-edited `global:categories:v1` payload can't render Adult/Violence chips before the refresh lands. The seeded-cache test is the one that would have caught it. Locally: package tests 17/17, `test/router/category_route_test.dart` 4/4, analyzer clean.

I also walked the surfaces to confirm nothing else reaches around the repository — `categories_tab` → `CategoriesBloc` → `watchCategoriesCached`, and `CategoryGalleryScreen` is constructed in exactly one place (the route builder). The metadata `CategoryChip` really is display-only; only the sibling hashtag chips carry a `GestureDetector`. That call is right.

**Still open: `porn`.** Confirmed against live funnelcake on this head — details inline. Same 38 videos `/categories/adult` now blocks, reachable under the sibling slug.

One convention item inline as well. Everything else holds up: normalization, the three enforcement points, `verifyNever` proving the short-circuit happens before the network call, and the null-category guard.

Happy to approve as soon as `porn` is in the set.

mobile/test/router/category_route_test.dart
```diff
@@ -20,5 +54,74 @@ void main() {
...
       expect(context.videoIndex, isNull);
     });
+
+    testWidgets('denied category names resolve to route error', (tester) async {
+      final route = _categoryGalleryRoute();
+
+      await _buildWithContext(tester, (context) {
+        final adultScreen = route.builder!(
+          context,
+          _FakeGoRouterState(
```
realmeylisdev, line 64: One coverage gap while you're in here: both denied cases pass only `pathParameters` and no `extra`, so they exercise the `VideoCategory(name: categoryName ?? '', videoCount: 0)` fallback. The branch the app actually takes is the other one — `categories_tab.dart:42` does `context.push(locationFor(category.name), extra: category)`, so a real tap always arrives through `extraAs`.

The guard reads the merged `category.name`, so it does catch that today. But the only test supplying `extra` uses an allowed name, which leaves denied-via-`extra` — the live path — unpinned. Reordering the guard or changing how `extra` is decoded would leave it open with all four tests green.

One more `_FakeGoRouterState` with `extra: const VideoCategory(name: 'adult', videoCount: 1)` covers it.

mobile/packages/categories_repository/lib/src/discovery_category_policy.dart
realmeylisdev: fixed in 57fe2c9.

realmeylisdev: fixed in 57fe2c9.

confirmed it end to end on a sim before pushing. `xcrun simctl launch co.openvine.app --route=/categories/adult` lands on the invalid-category screen and issues no funnelcake call; the same launch at `/categories/porn` was still firing `GET /api/videos?category=porn&sort=trending&limit=50`. after the commit both slugs stop at the guard.

kept `violence` for the reason above, and left `nudity` / `sex` / `drugs` / `weapon` out since those are a T&S call rather than a sibling of an already-denied slug.

mobile/packages/categories_repository/lib/src/discovery_category_policy.dart
realmeylisdev: withdrawing this one. you described the list as defense in depth, and defense in depth keeps earning its place after divinevideo/divine-funnelcake#864 lands, so whether this class is transitional at all is your call rather than a rule violation. worth noting funnelcake#864 only commits to `/api/categories`; `/api/videos?category=` is what the direct path actually hits, so the client guard may outlive the backend one either way.

mobile/test/router/category_route_test.dart
realmeylisdev: fixed in 19cb0e4. the new case is the only test that goes red if the guard is changed to trust `extra` (`st.extra == null && isDenied(...)`), so it pins the live branch rather than restating the fallback one.

realmeylisdev (approved):
pushed two commits, revert either freely. 57fe2c9 adds `porn` to the deny set: 38 of the 39 videos funnelcake serves under `porn` also carry `topic:adult`, so `/categories/porn` was opening what `/categories/adult` now blocks (on a sim, adult stops at the error screen while porn was still firing `GET /api/videos?category=porn`). 19cb0e4 pins the denied guard on the `extra` path, which is the branch a real tap takes.

dropped my TODO note, interim vs permanent defense in depth is yours to call. the shard 3 red earlier was an unrelated nip05 router test; it cleared on a same-seed local replay of the whole shard and on a job re-run, so it was ordering rather than your diff.

hm21 (approved):
LGTM.

Rebased onto `origin/main` (`19cb0e4` → `1a1f5f1`), which also flattened the `Merge branch 'main'` commit out of the history. No conflicts, PR-touched files byte-identical to what was reviewed. Analyze + package tests (17) + `category_route_test.dart` (5) green locally, all checks green on the new head.

NotThatKindOfDrLiz:
@rabble ready for your rereview

NotThatKindOfDrLiz:
@rabble, could you pls re-review when you have a minute? Your cache-read blocker from 4f590797 is fixed on the current head by filtering `_categoriesFromJson`, with a seeded-cache regression test. The later `porn` sibling slug and denied `extra` route path are also covered, and CI is green on c7d84bc4.

NotThatKindOfDrLiz:
@rabble ready for re-review on rebased head `793aa7e83`. This drops the two merge commits and is now up to date with `origin/main`.

Your cache-read blocker remains fixed via `_categoriesFromJson` filtering, with the seeded disk-cache regression test still passing. The later `porn` sibling slug and denied `extra` route path are also covered.

Verification after rebase:
- `flutter analyze`
- `flutter test --coverage` from `mobile/packages/categories_repository`
- `flutter test test/router/category_route_test.dart`
- pre-push hook: `mise exec -- flutter analyze lib test integration_test` + changed-file tests

rabble (commented):
Re-reviewed. **Both items addressed**, CI green (17/17) on `793aa7e83`.

**Blocking item — disk-cache read path.** `_categoriesFromJson` now filters through `DiscoveryCategoryPolicy.isDenied`, so a device holding `adult`/`violence` in its existing `global:categories:v1` payload drops them on the `cached` emission rather than rendering stale chips until the background refresh lands. Filtering on read rather than bumping the cache key is the right choice — it also covers a hand-edited or re-populated cache instead of being a one-shot migration.

`filters denied categories from disk-cached payloads` is exactly the test that would have caught the original gap: it seeds the DAO directly, asserts the stale emission is already filtered, and uses `' adult '` / `'Violence'` so the trim+lowercase normalization is pinned on the cache path too, not just the network one.

**Non-blocking item — denylist completeness.** The doc comment now names Funnelcake as the authoritative vocabulary and says what triggers an update. More usefully, you found `porn` is a sibling slug tagging the same videos as `adult` — that was a live hole, since blocking one left the same feed reachable at `/categories/porn`, and I'd assumed from inside this repo that `adult`/`violence` was the complete set. Good catch; it's the kind of thing only checking the producer would surface.

The second half of that item stands as an open question rather than a request on this PR: whether Funnelcake also refuses `/categories/adult` and `/videos?category=adult` server-side. A client denylist is defense-in-depth, and web plus any third-party client stay exposed if the API still serves those. Worth confirming somewhere, but not here.

NotThatKindOfDrLiz:
> Re-reviewed. **Both items addressed**, CI green (17/17) on `793aa7e83`.
> 
> **Blocking item — disk-cache read path.** `_categoriesFromJson` now filters through `DiscoveryCategoryPolicy.isDenied`, so a device holding `adult`/`violence` in its existing `global:categories:v1` payload drops them on the `cached` emission rather than rendering stale chips until the background refresh lands. Filtering on read rather than bumping the cache key is the right choice — it also covers a hand-edited or re-populated cache instead of being a one-shot migration.
> 
> `filters denied categories from disk-cached payloads` is exactly the test that would have caught the original gap: it seeds the DAO directly, asserts the stale emission is already filtered, and uses `' adult '` / `'Violence'` so the trim+lowercase normalization is pinned on the cache path too, not just the network one.
> 
> **Non-blocking item — denylist completeness.** The doc comment now names Funnelcake as the authoritative vocabulary and says what triggers an update. More usefully, you found `porn` is a sibling slug tagging the same videos as `adult` — that was a live hole, since blocking one left the same feed reachable at `/categories/porn`, and I'd assumed from inside this repo that `adult`/`violence` was the complete set. Good catch; it's the kind of thing only checking the producer would surface.
> 
> The second half of that item stands as an open question rather than a request on this PR: whether Funnelcake also refuses `/categories/adult` and `/videos?category=adult` server-side. A client denylist is defense-in-depth, and web plus any third-party client stay exposed if the API still serves those. Worth confirming somewhere, but not here.
> 
> **Nothing blocking from me.** My `CHANGES_REQUESTED` is stale and needs a maintainer to convert.

@rabble You are the maintainer and can absolutely convert your requested changes to approval. Please update your agent docs to ensure that your PR approvals are recorded.

PR #6945, opened 2026-08-09
fix(feed): page the New feed on the repository's hasMore flag

### LINKED ISSUE #6946: fix(feed): page the New feed on the repository's hasMore flag
Auto-created from PR #6945.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6945
**Branch:** `fix/new-feed-pagination-from-cached-page`

hm21:
Closes #6946

## Description

Selecting **New** in the home feed and then opening Explore left the New tab unable to load more — scrolling to the bottom did nothing for the rest of the session. Pull-to-refresh was the only way out.

Both feeds read the same source through `VideosRepository.getNewVideos` and share its cached first page, but they ask for different page sizes: the home feed takes the default 25, the Explore tab asks for `paginationBatchSize` (50). `getNewVideos` returned a bare `List<VideoEvent>`, so `NewVideosFeed` had to infer "there is more to load" from `videos.length >= 50`. Handed the home feed's cached 25, that inference read as "nothing more" — `hasMoreContent` went `false` and `ComposableVideoGrid.canLoadMore()` stopped calling `onLoadMore`.

The fix returns a `HomeFeedResult` carrying `hasMore`, as `getClassicVideos`, `getPopularVideosPage` and `getRecommendedVideos` already do. `getNewVideos` was the only feed method handing back a list without pagination metadata, which is exactly what forced the count-based guess. The cached page is still served instantly, and now says whether more sits behind it.

**Why not gate the cache instead:** the first attempt made the cache refuse a page fetched for a smaller limit. That fixes pagination but throws away the cross-feed cache hit — opening Explore then reloads from scratch every time, which is a worse regression than the bug. Reporting `hasMore` honestly keeps the instant page *and* fixes pagination.

**Related Issue:** 

## Out of Scope

- **Explore → For You** has the same symptom from an unrelated cause: the recommendations endpoint collapses `has_more`/`next_cursor` for any `limit >= 33`, and the app requests 50. That is a server-side fix, in flight as divine-funnelcake#868. Nothing in this PR touches that path.
- **Explore → Popular** is structurally immune: `getPopularVideosPage` returns the server's `hasMore` and `PopularVideosFeed` already pages on it (`_pageHasMoreContent(page) => page.hasMore`).
- `getHomeFeedVideos` shares the same limit-agnostic cache read (bloc asks 25, the idle warmup asks 50) but has no user-visible symptom, because `VideoFeedBloc` derives `hasMore` from `addedUniqueVideos` rather than a count. Left alone deliberately to keep this PR focused.

**In scope but worth calling out:** the home feed's own **New** mode changes behaviour here too. `VideoFeedSourceType.newVideos` used to hand `_hasMoreForSource` a `hasMore: null` and fall through to the `addedUniqueVideos` / `validVideos.isNotEmpty` fallback; it now consumes the repository flag like every other cursor-free source. Pinned by `stops newVideos pagination on the repository hasMore flag`.

Losing that fallback is only safe if a short page really does mean the source ran dry, and it didn't: `FunnelcakeApiClient.getRecentVideos` drops rows with an empty `id` or `videoUrl` before returning, so the repository's loop was comparing an already-filtered count against the limit it asked for. One malformed row broke the loop before the top-up and pinned a premature `hasMore: false` into `_inMemoryFeedCache['latest']` for the rest of the process — re-creating this PR's own symptom through a different door. `getRecentVideosPage` now reports the server's row count next to the filtered videos, following the `getV2PopularVideosPage` precedent, and the loop breaks on that.

It also stops discarding the envelope's `has_more`, which the repository prefers when the source states one. That path is inert today: `/api/videos?sort=recent` returns a bare JSON list with no envelope, so `_unwrapListResponse` yields `null` — the same server-side gap as divine-funnelcake#868, verified against the live endpoint. The relay path never had the defect, since `events.length < limit` counts the raw response and filtering happens afterwards.

## Verification

- `packages/videos_repository`: `flutter test --coverage` — 464 passing, **100% coverage** (package has no `min_coverage` override, so the VeryGood 100% gate applies).
- `packages/funnelcake_api_client`: `flutter test --coverage` — 379 passing, 93.16% (gate is `min_coverage: 44`); no new uncovered lines.
- `flutter test test/blocs/video_feed test/widgets/new_videos_tab_test.dart test/providers/new_videos_feed_provider_test.dart` — 110 passing.
- `dart format` — no changes.
- Generated Riverpod output regenerated and committed (`new_videos_feed_provider.g.dart`).

`videos_repository_test.dart` pins the repository *producing* the flag:
- `serves a smaller cached page to a larger limit` — the cached page is returned with **no** refetch (`verify(...).called(1)`), which is the instant-open behaviour.
- `reports more content behind a smaller cached page` — that shorter page reports `hasMore: true`, so the grid keeps paginating.
- `reports no more content when the source ran dry` — a genuinely short page still reports `hasMore: false`.

`new_videos_feed_provider_test.dart` (new) pins `NewVideosFeed` *consuming* it — the layer the bug actually lived in, which previously had no test at all: the only two suites touching `newVideosFeedProvider` override the class with a fake, so `build()` and `loadMore()` never ran. Three of its four cases fail against the old count-based inference; the fourth keeps them honest by pinning the `hasMore: false` direction.

`video_feed_bloc_test.dart` gains `stops newVideos pagination on the repository hasMore flag`, because every existing `getNewVideos` stub still returns the `hasMore: null` shape the repository can no longer produce — leaving the live branch uncovered. It fails if `_hasMoreForSource` goes back to the fallback.

On the client side, `keeps paging when the server filled the page but a row was dropped` and `prefers the server hasMore flag over the page count` cover the repository loop, and `reports the server row count alongside the filtered videos` covers `getRecentVideosPage`. The first goes red if the break condition reverts to the post-filter length.

Note: `flutter analyze` did **not** catch the type errors in the test stubs after the return-type change — only the compiler did during `flutter test`. Analyze alone is not sufficient review cover for this diff.

**Verified on a real device (Samsung Galaxy).** Cold start → home feed → mode **New** → open Explore (lands on the New tab) → scroll to the bottom. Before: nothing loads. After: the cached page appears instantly *and* scrolling loads more.

## Type of Change


mobile/packages/videos_repository/lib/src/videos_repository.dart
```diff
@@ -735,10 +741,25 @@ class VideosRepository {
...
+  /// paginate on the flag instead of re-deriving it from a count they may not
+  /// have requested.
+  HomeFeedResult _recentVideosResult(
+    List<VideoEvent> videos, {
+    required int limit,
+    required int? until,
+  }) {
+    final result = HomeFeedResult(
+      videos: videos,
+      hasMore: videos.length >= limit,
```
realmeylisdev: Non-blocking, and it doesn't change my read that this is the right fix.

The PR body justifies the home feed's behaviour change with *"both fetch paths in `getNewVideos` loop until `limit` is filled, so a short page really does mean the source ran dry."* That premise doesn't quite hold, so what this line reports is "the page filled" rather than "the source had more" (same for the doc at line 700).

`_fetchVisibleRecentVideosFromStatsApi` breaks on `videoStats.length < limit`, and `videoStats` is already filtered: `FunnelcakeApiClient.getRecentVideos` drops rows with an empty `id` or `videoUrl` before returning (`funnelcake_api_client.dart:506-509`) and throws away the server's `has_more` on the way through. One malformed row shortens the raw page, so the loop breaks *before* the top-up and `hasMore` comes back `false` for a source that still has pages. Since `until == null` there, that page and its `false` are then pinned in `_inMemoryFeedCache['latest']` for the rest of the process. The relay path has the same shape at `events.length < limit`.

Two consequences, flagging rather than fixing:

1. the home feed's `newVideos` source used to ride this out through the `addedUniqueVideos` fallback and now stops;
2. a short first page re-creates the Explore symptom this PR fixes, through a different door.

How live it is: `/api/videos?sort=recent` came back with exactly `limit` rows over six cursor pages at `limit=25` and one at `limit=50`, no dropped rows, so nothing is firing today. `videoStats.isEmpty` is the exhaustion signal that loop actually wants, but swapping it buys an extra request at every true end-of-feed on a hot path, so that reads like your call and probably its own PR.

realmeylisdev (approved):
lgtm. built this branch on an ios sim and scrolled it: explore new goes 50 then 99 then 148, and home feed new pages past the cached page too. mutation-checked both new test files, they go red if you revert the fix.

one non-blocking note inline on how hasMore gets derived, nothing that should hold the merge.

mobile/packages/videos_repository/lib/src/videos_repository.dart
hm21: Good catch — the premise was wrong and it's fixed in be46ee519.

`getRecentVideosPage` now reports the server's row count next to the filtered videos, and the loop breaks on that instead of the post-filter length, so a dropped row no longer pins `hasMore: false` into `_inMemoryFeedCache['latest']`. Mutation-checked: reverting the break condition turns the new test red.

Two corrections to the analysis:

The relay path doesn't share the defect. `events.length < limit` counts the raw relay response — `_transformAndFilter` runs after it, not inside the client.

`has_more` is moot today rather than discarded value: `/api/videos?sort=recent` returns a bare JSON list, no envelope, so `_unwrapListResponse` yields `hasMore: null` either way. I stopped discarding it and the repository now prefers it when non-null, but that path stays inert until the endpoint grows an envelope — same server-side work as divine-funnelcake#868.

PR #924, opened 2026-01-14
feat: Adding Maestro automated tests

aberardinellict:
## Description

First PR with the maestro automated test cases, assertions, flows and scripts plus a README.md file

**Related Issue:** 

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #6942, opened 2026-08-09
fix(profile): remove deleted videos from profile tabs

NotThatKindOfDrLiz:
## Summary
- add repository-level deleted-video filtering to VideosRepository and wire it to VideoEventService tombstones
- subscribe Liked, Reposts, Saved, and Collabs profile tab blocs to removedVideoIds and filter restored snapshots
- cover repository lookup filters, profile tab removal/snapshot behavior, and profile grid provider wiring

Closes #6924

## Tests
- flutter analyze lib test integration_test
- flutter analyze lib test (mobile/packages/videos_repository)
- flutter test --coverage (mobile/packages/videos_repository)
- flutter test test/blocs/profile_liked_videos/profile_liked_videos_bloc_test.dart test/blocs/profile_reposted_videos/profile_reposted_videos_bloc_test.dart test/blocs/profile_saved_videos/profile_saved_videos_bloc_test.dart test/blocs/profile_collab_videos/profile_collab_videos_bloc_test.dart
- flutter test test/widgets/profile/profile_grid_test.dart
- pre-push hook: analyzer, generated files, changed-file tests

PR #6948, opened 2026-08-09
fix(e2e): make the Maestro smoke suite runnable again

realmeylisdev:
## Description

The Maestro E2E suite has been unrunnable since February.

**The local runner could never start.** `ui_smoke_ios.sh` looked for `suites/smoke.yml` when the file is `smoke.yaml`, and required a `Runner.app` inside `e2e/maestro/` that is gitignored and that nothing produces. It died at its own precondition checks before Maestro was invoked. CI calls `maestro test` directly, so it never hit either.

**The selectors had rotted.** An audit of the smoke closure found 72 of ~107 text selectors dead: 36 changed, 29 gone, 7 alive but moved to a different screen.

After this PR, `loginFreshInstall` and `removeKeys` pass end to end **on both iOS and Android**, and `likeFlow` runs from a cold start through account creation, playback, liking, the profile and the liked tab.

### Why it needed app changes, not just selector edits

Most of the drift was architectural. `divine_ui` exposed no `Semantics` identifier at all, so every auth and settings screen could only be addressed by localized copy — which is exactly what kept breaking. Four components now take a `semanticIdentifier`, and the screens on the smoke path carry ids.

Five behaviour changes the flows predated, each found by running them:

- **Account creation is gated behind an invite code.** `loginFreshInstall` is the entry point for 3 of the 4 smoke flows, so all of them stopped at "Add your invite code". The gate's mode is server-decided and STAGING shares the production invite service, so there is nothing environment-level to flip. Both E2E builds now pass `--dart-define=GH_ACTIONS_PR_PREVIEW=true`, the existing flag that makes the gate self-redirect. Chosen over carrying a live invite code as a secret, which would expire and take the suite down with it.
- **The iOS keychain survives app uninstall.** `clearState: true` does not sign the user out, so the welcome screen renders its returning-user variant on every run after the first. `assertLogin` described only the fresh variant — which is why it could pass exactly once per simulator.
- **Signing up lands on the Home feed, not Explore.** The old flow asserted Explore directly, which would have passed on the bottom-nav Explore *button* — present on every screen. Green for the wrong reason.
- **Settings moved.** The hamburger menu the teardown navigated through does not exist; Remove keys moved onto Nostr Settings; its title and subtitle both changed.
- **`TabBar` composes a compound label.** A tab reads as `New\nTab 2 of 6` and Maestro matches the whole label. That is why the original encoded `Tab 2 of 5` — a count that is now built at runtime and varies between 4, 6 and 7.

### Accessibility fixes and follow-up

- **`DiVineAppBar`'s back button announced a node that could not be activated.** `expandLeadingHitArea` wraps the visible button in `AbsorbPointer` and handles taps on an outer `GestureDetector`, but the semantics stayed on the absorbed inner button. Assistive tech and UI tests both target the labelled node, and here that node did nothing.
- **The profile tab bar and liked/saved grids now expose those test anchors as Semantics identifiers.** They still keep the legacy test anchors as labels, so VoiceOver can still read values like "liked_tab" and "liked_video_thumbnail_0" until #6951 adds localized labels across the profile tabs/grids.

### Security

`suites/fullRegression.yaml` carried a live Nostr **nsec** in plaintext, committed in #924 on 2026-01-15. `flows/loginEmailPwd.yaml` carried a plaintext mailinator email/password pair — on the smoke path, so CI shipped it on every run, and Mailinator is a public inbox. A third party's real name, personal Gmail, Snapchat, Instagram and Twitter handles were committed as a bio fixture.

All of it now comes from `maestro -e`, backed by a new `maestro_e2e_credentials` Codemagic group. A missing variable resolves to the JavaScript value `undefined` rather than erroring, so each entry point carries an explicit `assertTrue` preflight — measured both ways before relying on it.

> **Action needed outside this PR:** the nsec and the mailinator password are compromised regardless of this change and need rotating. Purging them from git history is a force-push on `main` and is not mine to make. The third-party PII is the one with an external clock.

**Related Issue:** Relates to #4239

## Out of Scope

- **`fullRegression`.** Its own audit found ~109 broken selectors against smoke's ~72, and the difference is kind rather than size: four of its journeys were **removed from the product**, not restyled — there is no log-out in Settings, no unfollow affordance in the feed, no four-step deletion wizard, no Edit Profile row. Those need re-specification, not repair.
- **`commentFlow` and `searchUserFlow`.** Search moved from a three-tab `TabBar` to sliver sections and needs rewriting rather than patching.
- **Test-data determinism.** The flows depend on live staging content — whichever video is newest, a specific account ranking first in search — and create a real account per run. `videoUnlike` currently fails because the liked video shows a loading placeholder and never resolves: no action buttons render, so there is nothing to assert. That is a content problem and no amount of selector work fixes it.
- **Localized profile tab labels.** Promoting them properly means new ARB keys across 22 locales; the defect is commented at the site instead.
- **CI triggering.** Left manual-only, matching all six Codemagic workflows. Proving it green comes first.
- **`ios-simulator-build`** carries the same missing-SPM-anchor bug, but sits outside the E2E lane.

## Verification

Every commit was verified before it landed; the suite was run against a real simulator and emulator throughout, not reasoned about.

- **iOS** (iPhone 17 Pro Max, Maestro 2.1.0, STAGING build): `loginFreshInstall` green; `removeKeys` green; `likeFlow` reaches the unlike step.
- **Android** (`Pixel_9_Pro_XL`, STAGING debug APK): `loginFreshInstall` green; `removeKeys` green. Every anchor transferred unmodified.
- `bash mobile/e2e/maestro/scripts/check_refs.sh` — new; catches wrong-case `runFlow` paths, which resolve on macOS and fail on the Linux CI runner. It found `../tests/unFollow.yaml` against `unfollow.yaml`.
- `divine_ui`: 870 tests, **coverage held at 100%** (2102/2102).
- Touched app areas: `test/screens/auth`, `test/screens/settings`, `test/widgets/profile`, `test/screens/feed`, `test/screens/welcome_screen_test.dart` — 739 passing.
- Regression checks after the shared `DiVineAppBar` change: app-bar, feed, conversation, hashtag-feed and app-shell suites all pass.

Note that CI cannot verify the E2E lane itself — both workflows are manual-dispatch. A manual `e2e-smoke-android` run is worth doing before merge.

## Type of Change

- [x] 📝 Documentation


realmeylisdev:
Follow-ups filed for the work this PR deliberately left out:

- #6949 — make the flows independent of live staging content. This is the ceiling on the whole suite: `videoUnlike` fails today because the liked video never finishes loading, and no amount of selector work fixes that.
- #6952 — the remaining smoke flows. Search needs a rewrite rather than a patch (three-tab `TabBar` → sliver sections); the comment flow has a real `env`-vs-`output` bug and a disabled cleanup step.
- #6950 — four `fullRegression` journeys that were removed from the product rather than restyled, so they need re-specifying rather than repairing.
- #6951 — the profile tabs announcing their test anchors to screen readers. This PR added the identifiers but left the labels, because fixing them properly means new ARB keys across 22 locales.

Worth doing before merge: a manual `e2e-smoke-android` run on Codemagic. GitHub CI cannot cover the E2E lane — both workflows are `triggering: events: []` — so the only proof the CI harness itself works is a dispatched run. The suite passes locally on a `Pixel_9_Pro_XL` emulator, but that does not exercise Codemagic's emulator setup, which has its own issues: `launch_android_emulator` still does `cd $ANDROID_HOME/tools`, a directory removed from modern SDKs.

hm21 (approved):
Approving. Verified the claims independently: analyze clean, `divine_ui` 865 tests at 100% coverage (2096/2096), `check_refs.sh` green, no credentials left in the tree.

Pushed three fixes as separate commits and rebased onto `main`:

- **`DivineButton` split its semantics in two.** With `semanticIdentifier` and no `semanticLabel`, declaring `button`/`enabled`/`onTap` on the wrapper forced a second node: the outer one carried the identifier and an empty label, the inner one carried "Sign in". VoiceOver reads that as an unnamed button ahead of the real one — on all six identifier-only call sites, i.e. the whole sign-up path. Now annotates the existing node instead. The old test asserted the identifier and the label each existed *somewhere*, which the split tree satisfied; it now asserts they land on the same node, and fails without the fix.
- **`_LeadingIconButton` hardcoded `'back_button'`** for all three leading variants, so a menu button or custom leading icon would have carried it too. Latent today (one call site, a back button — `id: "back_button"` in the flows is unaffected). Now passed per variant, with the three ids as constants on `DiVineAppBarLeading`.
- **`DivineListTile._buildTile`** inlined — `code_style.md` rules out methods returning `Widget`.

One question, not blocking: `suites/smoke.yaml` is unchanged and still runs `commentFlow` and `searchUserFlow`, which #6952 covers, plus `likeFlow` which stops at `videoUnlike`. So `e2e-smoke-ios` / `-android` — the workflows this wires `maestro_e2e_credentials` into — still can't go green, and the README documents the CI command without saying so. Worth trimming the suite to what passes, or adding a current-state note?

realmeylisdev:
@hm21 — took the suite-trimming option. Pushed in `a03c6ed`, checks green.

`suites/smoke.yaml` now runs `loginEmailPwd` alone. `likeFlow`, `commentFlow` and `searchUserFlow` are held back with a comment naming the tracking issue for each, so a dispatched run proves the harness rather than re-reporting #6949 / #6952. The README gained a table of what's in and out, plus the bit it never said: `e2e-smoke-ios` and `e2e-smoke-android` are `triggering: events: []`, so nothing runs them automatically and GitHub CI does not cover this lane at all.

Two changes you didn't ask for, called out separately:

- **Dropped `SEARCH_USER` from both smoke runners** (`codemagic.yaml` and `ui_smoke_ios.sh`). Nothing in the trimmed suite reads it, and both hard-fail on it — so a run would have been blocked on a credential it no longer uses. `searchUserFlow` still guards it for `fullRegression`; the PR that restores the flow restores the variable, and the suite comment says so.
- **Fixed a stale claim that the smoke suite asserts the `STG` badge.** The only `STG` assertion is in `tests/searchTags.yaml`, reachable from `fullRegression` alone — smoke never ran it, so this was wrong before this PR too. Both the runner header and the README implied a PRODUCTION build would fail smoke for that reason. Corrected to the reason that does apply: it signs in and writes to the live relay.

On your three fixes — kept as-is, and the `DivineButton` one is a good catch on a path I'd have shipped.

Note for whoever merges: someone merged `main` into the branch as `2690a4c14` while I was working. I replayed on top of it rather than force-pushing over it, so nothing was lost, but the branch now carries a merge commit against the usual rebase-only rule.

Still open, and the reason I'd hold merge: the manual `e2e-smoke-android` dispatch on Codemagic hasn't run. It's the only thing that can prove the harness, and `launch_android_emulator` still does `cd $ANDROID_HOME/tools` under `set -e` — that directory is gone from modern SDKs, so I'd expect the dispatch to die there before the emulator boots. Worth doing before this lands.

NotThatKindOfDrLiz (approved):
Took over the remaining cleanup and pushed two commits on top of the rebased branch:

- Rebased onto current origin/main and resolved the additive DivineIconButton test-helper conflict, clearing the merge conflict and removing the merge commit from the branch.
- Added semanticIdentifier pass-through to DivineAppBarIconButton so default app-bar leading buttons expose the same stable ids as the expanded-hit-area path, with a divine_ui regression test.
- Updated smoke.yaml to run the verified loginFreshInstall path plus removeKeys before loginEmailPwd, removed the dead commented-out keyboard retry blocks, and updated the Maestro README.
- Corrected the PR body so it no longer overclaims the profile tab/grid accessibility follow-up, and updated the divine_ui coverage numbers.

Verification:
- bash mobile/e2e/maestro/scripts/check_refs.sh
- flutter test test/src/app_bar/divine_app_bar_test.dart test/src/button/divine_icon_button_test.dart from mobile/packages/divine_ui
- flutter test --coverage from mobile/packages/divine_ui: 870 tests, 2102/2102 lines = 100.00%
- flutter analyze lib test from mobile/packages/divine_ui
- pre-push checks: mergeability, analyzer, changed-file tests all passed
- GitHub checks: all green on 279e0e7df794088b4f51dd8480aa33944585bd3f

@realmeylisdev please review the takeover commits and merge when you are comfortable. I did not run the manual Codemagic e2e-smoke-android dispatch; the PR body still calls that out as worth doing before merge.

NotThatKindOfDrLiz:
@realmeylisdev takeover changes are pushed and CI is green on 279e0e7df794088b4f51dd8480aa33944585bd3f. Please review the two added commits and merge when you are comfortable.

I did not run the manual Codemagic e2e-smoke-android dispatch; the PR body still calls that out as worth doing before merge.

realmeylisdev:
@NotThatKindOfDrLiz reviewed both commits — they hold. I re-ran your numbers rather than taking them: `divine_ui` is 870 tests, 2102/2102 = 100.00%, `flutter analyze lib test` clean, `check_refs.sh` green.

**`a84c47078`** — right fix. Forwarding `semanticIdentifier` through `DivineAppBarIconButton` closes what hm21's commit left open: the ids only reached the `expandHitArea` path, so a default app bar had none at all. Two failure modes this shape invites, both already covered:

- No duplicate node when `expandHitArea` is set — the inner button sits inside `ExcludeSemantics` (`divine_app_bar_leading.dart:206`), and every identifier assertion in the file is `findsOneWidget`, which fails on two matches.
- No label/identifier split of the kind hm21 caught on `DivineButton` — `divine_icon_button_test.dart:120` already pins both onto the same node one level down, so the app bar inherits it.

**`279e0e7df`** — also right, and "verified" is backed: the PR body records `loginFreshInstall` and `removeKeys` green on both iPhone 17 Pro Max and `Pixel_9_Pro_XL`, which is exactly the pair you added. Traced the two things that could have bitten:

- Env closure stays clean. Only `loginEmailPwd` reads anything (`USER_EMAIL`/`USER_PWD`), so dropping `MAESTRO_SEARCH_USER` from both runners is still correct.
- Not a double teardown. `loginFreshInstall` ends signed in on Explore, `removeKeys` tears that account down, and `loginEmailPwd` relaunches with `clearState` and runs its own `removeKeys` at line 48.

**One commit pushed that you didn't ask for**, and it's the blocker I'd flagged rather than review feedback: `launch_android_emulator` opened with `cd $ANDROID_HOME/tools` under `set -e`. That directory is gone from modern SDKs, so the step died on its first line — before the emulator started, and before `install_app_android_emulator` or `run_maestro_smoke_tests` could run. The dispatch could never have got past it. It now invokes `$ANDROID_HOME/emulator/emulator` by path, with a legible error if it is absent.

Also rebased onto current `origin/main`. Someone merged `main` in again as `e6f3d71d8` — second time — so the branch was carrying a merge commit against the rebase-only rule. The rebase flattens it; nothing was lost, and the only content delta is #6942 arriving from `main`.

Still unproven, and not closable from here: nobody has dispatched `e2e-smoke-android`. Two things in that step stay unverifiable without one — it passes no `-no-window` on a headless `linux_x2` instance, and nothing in `codemagic.yaml` ever creates an AVD, so it assumes the image ships one named `emulator`. I fixed the failure I could prove and left those alone rather than guessing at Codemagic's image. Worth a dispatch as the first follow-up.

Checks green on d051c980e. Merging.

PR #6956, opened 2026-08-09
fix(a11y): announce the branded loading indicator as busy

### LINKED ISSUE #6957: fix(a11y): announce the branded loading indicator as busy
Auto-created from PR #6956.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6956
**Branch:** `fix/branded-loading-indicator-semantics`

hm21:
Closes #6957

## Description

`BrandedLoadingIndicator` emitted no semantics of its own. Its sprite sheet is an unlabelled `Image`, so a screen reader got a bare "image" node and none of the ~60 call sites announced that anything was happening.

The clearest victim was the cover editor's confirm button: while confirming it swaps the check icon for the indicator, so VoiceOver/TalkBack read "Use selected frame as video cover, disabled" — unavailable, with no hint it was busy. `video_metadata_cover_screen.dart` even carried a hand-rolled `ExcludeSemantics` to suppress the stray image node, which is the same noise this change removes globally.

The indicator now labels itself with the localized `commonLoading` ("Loading", already translated in every locale) and excludes the sprite beneath it. Because `Semantics` merges into an enclosing annotation, the confirm button now reads "Use selected frame as video cover, Loading, disabled" without any per-call-site work.

Why the details are the way they are:

- **`semanticsLabel` parameter.** A caller that wants a more specific label can't just nest its own `Semantics` — the labels would merge and double-announce. The parameter mirrors `CircularProgressIndicator.semanticsLabel`, which this widget replaces throughout the app.
- **A static label, not `liveRegion: true`.** A live region would announce proactively on appearance, but the repo's existing loading precedents (`ModalProgressOverlay`, the people-search skeleton) use a plain label, and `.claude/rules/accessibility.md` reserves proactive announcements for `SemanticsService.sendAnnouncement`. Easy to revisit if we want the louder behaviour.
- **Five sites opt out** — the feed load-more pill, the relay-info row, the sandbox status card, the captions sheet ("Listening for speech…") and the chroma-key overlay ("Applying the green screen…") — because a sibling `Text` already names the wait, and the generic "Loading" would only preempt the more specific copy.
- **The feed placeholder stays silent.** `VideoLoadingPlaceholder` lives inside the feed item's "Play video" button, which already owns that region's semantics (its thumbnail was excluded for the same reason). Buffering there is usually sub-second, so a merged "Play video Loading" would mostly be stale by the time a user focused it.
- **One comment updated in `video_editor_timeline_controls.dart`**, which #6885 wrote around the old behaviour ("the spinner ... has no semantics of its own"). The logic there is unchanged and still correct — the caption names the control, the spinner now names the wait.

**Related Issue:** 

## Out of Scope

- No copy changes: reuses the existing `commonLoading` key rather than adding a new one.
- Loading indicators that are not `BrandedLoadingIndicator` (raw `CircularProgressIndicator`, skeletons) are untouched.

## Verification

- `flutter test` — full suite green (~15k tests)
- New cases in `branded_loading_indicator_test.dart`: default label, caller-supplied label, and the sprite kept out of the tree (node is not flagged `isImage` and has no children)
- `video_metadata_cover_screen_test.dart` now pins the merged busy label instead of the bare one
- Manual pass with TalkBack on a real Samsung Galaxy 

## Type of Change

### DIFF
--- a/mobile/lib/widgets/branded_loading_indicator.dart
+++ b/mobile/lib/widgets/branded_loading_indicator.dart
@@ -5,6 +5,7 @@ import 'dart:ui' as ui;
 
 import 'package:divine_ui/divine_ui.dart';
 import 'package:flutter/material.dart';
+import 'package:openvine/l10n/l10n.dart';
 import 'package:unified_logger/unified_logger.dart';
 
 /// A branded loading indicator that displays the animated Divine logo.
@@ -19,7 +20,11 @@ import 'package:unified_logger/unified_logger.dart';
 /// - Consistent animation across widget rebuilds
 /// - Better performance on repeated displays
 class BrandedLoadingIndicator extends StatefulWidget {
-  const BrandedLoadingIndicator({super.key, this.size = 80.0});
+  const BrandedLoadingIndicator({
+    super.key,
+    this.size = 80.0,
+    this.semanticsLabel,
+  });
 
   /// Sprite sheet backing the animation: the Divine brand mark
   /// (`assets/icon/divine_mark.svg`) in white, drawn as a wing-flap cycle.
@@ -35,6 +40,13 @@ class BrandedLoadingIndicator extends StatefulWidget {
 
   /// The size (width and height) of the loading indicator.
   final double size;
+
+  /// What a screen reader announces while the indicator is on screen.
+  ///
+  /// Defaults to the localized "Loading". Pass a more specific label when the
+  /// indicator stands in for a named control; wrap the indicator in
+  /// [ExcludeSemantics] instead when a sibling already announces the wait.
+  final String? semanticsLabel;
 
   @override
   State<BrandedLoadingIndicator> createState() =>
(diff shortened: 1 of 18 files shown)


mobile/lib/widgets/branded_loading_indicator.dart
```diff
@@ -90,72 +102,81 @@ class _BrandedLoadingIndicatorState extends State<BrandedLoadingIndicator>
...
-              // filter layer at the indicator's size.
-              child: ImageFiltered(
-                imageFilter: ui.ImageFilter.blur(
-                  sigmaX: widget.size * _shadowBlurRatio,
-                  sigmaY: widget.size * _shadowBlurRatio,
+    // The sprite is an unlabelled [Image], which reaches a screen reader as a
+    // bare "image" node. Excluding it and labelling the wrapper is what turns
+    // the indicator into an announcement that something is still running.
+    return Semantics(
+      label: widget.semanticsLabel ?? context.l10n.commonLoading,
```
realmeylisdev, line 109: nit: the ARB description for `commonLoading` still describes the single blocking case it was written for:

> Screen-reader label for a blocking progress spinner that covers the screen while an action is being prepared. Announced instead of the content behind it, which is not reachable while the spinner is up.

That fits `ModalProgressOverlay`, but not most of the ~55 indicators this now labels. The 44px one in the cover editor's top bar doesn't cover the screen, and the video area behind it stays reachable. `flutter gen-l10n` copies the description straight onto the generated `commonLoading` getter (`app_localizations.dart:11027`), so it is both the translator brief and the hover doc at this call site.

Widening it to something like "Screen-reader label for a progress indicator shown while content or an action is loading" leaves the English value and all 21 translations untouched.

realmeylisdev (approved):
lgtm. went through all 62 call sites and the five opt-outs are the right set, every other spinner-with-a-sibling-label is an error branch that can't co-render with the loading one. checked the merged label on the cover button and ran it on the sim, nothing new in the logs.

left one note inline about the commonLoading arb description. it's outside your diff so i didn't touch it, your call whether it rides along here.

mobile/lib/widgets/branded_loading_indicator.dart
hm21: Fixed in 7dd17f59d6 — widened the description in `app_en.arb` plus the four locales that mirror it (ms, ur, vi, zh), regenerated. Values untouched.

Issue #6744, opened 2026-08-05
fix(comments): Kind 34236 video replies never reach the first page of a comment sheet


realmeylisdev:
Opening a comment sheet takes the REST bootstrap whenever the funnelcake client is available, which it is in the shipped wiring (`repository_providers.dart` always constructs one, and `isAvailable` only checks that the base URL is non-empty). `loadComments` returns `_filterThread(thread)` as soon as that call yields a non-null response, before ever reaching the relay branch — so the `includeVideoReplies` parameter is silently ignored for `before == null`.

funnelcake cannot supply the replies on that path either. Its comments read model is Kind 1111 only:

- `crates/clickhouse/src/client.rs` — `toUInt16(1111) AS kind`, with the note that the table is kind=1111 by construction
- `database/migrations/000127_comments_by_target.up.sql` — `WHERE e.kind = 1111` on both materialized-view arms, and a header stating the table has no `kind` column because every row is implicitly 1111
- every UNION arm of the legacy `build_video_comments_tag_subquery` binds `WHERE kind = 1111`
- the handler doc says it returns NIP-22 (Kind 1111) comment events

So with `FF_VIDEO_REPLIES` on, a video reply is absent from the first page and only appears once the user pages back (`before != null` takes the relay branch) or when funnelcake is unavailable. Two existing tests already pin the shape without naming it as a defect: "does not merge relay video replies into REST bootstrap comments when video replies are disabled" and "REST bootstrap does not wait for relay video reply lookup", the latter passing `includeVideoReplies: true` and asserting `verifyNever(queryEvents)`.

Surfaced while verifying #6738. That PR is unaffected — its coordinate plumbing is correct wherever a video reply actually arrives — but it does mean the REST mappers coordinate derivation added there cannot fire today, since `_ownAddressableId` guards on `kind >= 30000`.

Two directions, and the choice is a product/backend call rather than a client one:

1. Widen the funnelcake read model to carry kind 34236 replies (and stop hard-coding the kind), so one REST page is the whole thread.
2. Keep REST as the Kind 1111 bootstrap and have the client merge a relay lookup for video replies on the first page too — at the cost of the latency the "REST bootstrap does not wait for relay video reply lookup" test was written to protect.

PR #6960, opened 2026-08-09
fix(comments): show an in-flight video reply on its destination sheet

### LINKED ISSUE #5862: fix(comments): video replies don't appear until reload (no optimistic insert)
## Summary

Recording a **video reply** to a comment does not show it in the comments sheet
until a reload. Unlike a text reply (fixed in #5861), the video-reply camera path
publishes in the background and navigates away with **no optimistic insertion**,
so the reply can only appear via a fresh `loadComments` / relay echo on the
destination sheet.

Surfaced during the #5854 investigation (reply-surface audit); this is the video
counterpart to the text-reply visibility bug.

## Evidence (main)

- `video_publish_provider.dart:435-467` — for a `videoReplyContext`, dispatches
  `BackgroundPublishRequested(...)` then `context.go(...)` to the **root video's**
  detail (`VideoDetailScreen.pathForId(root)` + `VideoDetailRouteExtra(autoOpenComments: true)`).
  There is **no** `OptimisticCommentInserted` / `CommentsListBloc` mutation
  anywhere in the publish path.
- Entry point `comments_screen.dart:689-711` (`_openVideoReplyCamera`) only sets
  `VideoReplyContext` and pushes `VideoRecorderScreen` — no placeholder.
- Display is gated by `FeatureFlag.videoReplies` (default **OFF** in release, **ON**
  in profile/debug) — `comments_list.dart:88-90`, `build_configuration.dart:43-48`.
- Contrast: text replies have a working optimistic insert
  (`comment_composer_bloc.dart:171-179` → `OutboxBridges` → `OptimisticCommentInserted`).

## Proposed fix

Give the video-reply path the same optimistic treatment as text replies: insert a
pending placeholder (with the local thumbnail / uploading state) into the
destination `CommentsListBloc`, reconciled on publish success — OR, at minimum,
ensure the destination sheet's `loadComments` + `_mergeRecentlyPostedComments`
self-heal surfaces the just-posted video reply promptly.

## Acceptance criteria

- After recording a video reply, it appears in the comments (optimistically, or
  promptly on the destination sheet) **without a manual reload**.
- Works with `FeatureFlag.videoReplies` enabled (profile/debug); no regression when
  disabled.
- Covered by a test asserting the video reply reaches the list store.

## Related

- #5854 / #5861 (text-reply visibility).

realmeylisdev:
Fixes #5862.

## Problem

Recording a video reply publishes in the background and navigates to the **root video's** comments sheet, which mounts fresh. Nothing represented the reply until the entire round-trip finished — render → Blossom upload → sign → broadcast → relay echo. For a 6s clip that is seconds to tens of seconds of completely silent UI, so the poster assumes the post failed. A text reply covers exactly that window with an optimistic placeholder; the video path had none.

## One correction to the issue

The issue title says video replies "don't appear until reload". That is imprecise about the mechanism: the reply **does** arrive with no reload, via the live `comments_watch_*` REQ. Verified by opening the app's exact dual-filter REQ against a relay and publishing a kind-34236 reply 5s later — it arrived on the post-EOSE tail (the pre-existing one arrived on backfill, duplicated per filter, which `watchComments`' `seenIds` already dedups).

So the defect is the **silent multi-second window**, not invisibility. That is what this PR fixes.

## Approach

The destination `CommentsListBloc` does not exist when the publish starts — the sheet mounts *after* navigation — so a one-shot event has nowhere to land. The placeholder is instead derived from `BackgroundPublishBloc.state.uploads`, which already holds the in-flight set and whose drafts carry `videoReplyContext`. Deriving from durable state also means the placeholder survives a sheet reopen mid-publish.

- `VideoReplyPlaceholderBridge` seeds from the current publish state at mount (a change-only listener would never fire, since the upload predates the sheet) and syncs on every subsequent emit.
- The placeholder id encodes the draft id, so removal is deterministic. The existing relay-echo swap matches on author + content, and a video reply's content is its description — frequently empty.
- Reuses the existing `OptimisticCommentInserted` / `OptimisticCommentRolledBack` events and the existing echo-reconciliation, so no new bloc event and no repository API.

## Deliberately not changed

- **`loadComments`** — `comments_repository_test.dart:182` pins its REST-first short-circuit (`verifyNever(queryEvents)` even with `includeVideoReplies: true`) as first-paint latency protection.
- **#6744** stays open and separate: funnelcake's comments read model is kind-1111 only, so a sheet reopen still gets the reply from the relay backfill rather than the REST first page. NIP-22 specifies `kind:1111`, so widening funnelcake is a deliberate deviation and a backend/product call.

## Testing

- `video_reply_placeholder_bridge_test.dart` — seeds from an already-running upload; **inserts nothing for a different root**; matches on the addressable coordinate; ignores a non-reply publish; rolls back on completion; **no duplicate row after the relay echo**.
- `pending_video_reply_tile_test.dart` — localized copy (asserted against `AppLocalizations`, and that the `de` string is absent), progress binding, indeterminate at 0, screen-reader label.
- `comments_list_helpers_test.dart` — id round-trip, and that it keeps the shared `pending_comment_` prefix load-more and the echo swap depend on.

The bridge tests were mutation-checked: forcing `videoReplyTargetsRoot` to always return true fails exactly the different-root test and nothing else.

New ARB keys are translated in all 22 locales (no `_knownUntranslatedDebt` entries).

`flutter analyze lib test` clean; comments + blocs + l10n suites green (328); all six design-system / structure ratchets pass.

## Not verified

The record → publish → echo leg was not exercised on a device: the iOS Simulator has no camera, and a physical device cannot reach the local stack, so it would mean publishing a real public kind-34236 event. That gap is quantitative (how long the window is), not causal.


NotThatKindOfDrLiz:
@realmeylisdev I pushed takeover fixes in 485eca052. Summary:

- Made all pending_comment_ rows display-only, so reply/vote/delete/edit paths cannot publish against placeholder ids.
- Reused commentPlaceholderIdPrefix across production call sites instead of leaving duplicate literals beside the helper.
- Kept successful video-reply placeholders through a short relay-echo grace window so the echo can swap in place without a disappearance or self-authored new-comment pill; failed/vanished uploads still roll back.
- Matched the pending video tile to the real 9:16 comment-player footprint and added an explicit accessibility announcement.
- Preserved load-error reporting when the store contains only optimistic placeholders.

Verification:
- flutter test test/screens/comments/comment_item_test.dart test/screens/comments/video_reply_placeholder_bridge_test.dart test/screens/comments/widgets/pending_video_reply_tile_test.dart test/blocs/comments/comments_list/comments_list_bloc_test.dart test/blocs/comments/comments_list/comments_list_helpers_test.dart
- flutter analyze lib test
- pre-push changed-file checks passed
- GitHub CI is green on 485eca052

NotThatKindOfDrLiz (approved):
Approved after takeover fixes in 485eca052. I rechecked the pending-placeholder paths, local focused tests and analyzer pass, pre-push checks passed, and GitHub CI is green. Author should review the takeover commit before merging.

Issue #6961, opened 2026-08-09
refactor(imeta): route service parsers through parseImetaTag


NotThatKindOfDrLiz:
## Problem

Two service-layer imeta parsing paths still hand-roll NIP-92 parsing after the shared parser moved into `nostr_sdk`:

- `mobile/lib/services/video_metadata_update_service.dart`
- `mobile/lib/services/corrupted_video_repair_service.dart`

The update-service path has a correct inlined parser, but it can now drift from `parseImetaTag`. The repair-service path only handles space-separated imeta and silently misses positional imeta tags.

## Desired outcome

Route the remaining service-layer parsing through `parseImetaTag` where possible, while preserving the repair service requirement to keep the original tag encoding when rewriting events.

## Acceptance criteria

- Service-layer imeta reads support both space-separated and positional encodings.
- The corrupted-video repair path preserves the input encoding on output.
- Regression tests cover both formats.
- No new dependency edges beyond the existing `nostr_sdk` dependency surface.

## Context

Follow-up to #6959. The parser now lives in `mobile/packages/nostr_sdk/lib/nip92/imeta_tag.dart`.

PR #6962, opened 2026-08-09
fix(imeta): parse positional comment and NIP-94 metadata

### LINKED ISSUE #6959: fix(imeta): parse positional metadata in comment and NIP-94 paths
## Problem

PR #6939 consolidated imeta parsing for models and fixed VideoStats.dimensions, but two older parsers still only handle space-separated imeta fields:

- mobile/packages/comments_repository/lib/src/comments_repository.dart
- mobile/packages/nostr_sdk/lib/nip94/file_metadata.dart

Those paths miss positional imeta tags like:

```dart
['imeta', 'url', 'https://example.com/video.mp4', 'dim', '1080x1920']
```

The comments_repository case is user-visible: comment videos using positional imeta can lose dimensions, thumbnail/image, blurhash, duration, and sometimes URL metadata.

## Desired outcome

Support both imeta encodings consistently:

- space-separated: `['imeta', 'url https://...', 'dim 1080x1920']`
- positional: `['imeta', 'url', 'https://...', 'dim', '1080x1920']`

## Notes

Decide whether comments_repository should depend on models and whether parseImetaTag should be exported from the models barrel, or whether the shared parser belongs lower in the package graph.

Follow-up from PR #6939.

NotThatKindOfDrLiz:
## Summary

- move the shared `imeta` parser from `models` into `nostr_sdk` and export it there
- use the shared parser for comment video metadata and NIP-94 file metadata
- accept `thumb` as a comment thumbnail key alongside `image` while parsing shared `imeta` metadata
- add regression coverage for positional `imeta` in comments and NIP-94, plus direct parser tests

Closes #6959
Refs #6961

## Verification

- `flutter test test/nip92/imeta_tag_test.dart test/nip94/file_metadata_test.dart` from `mobile/packages/nostr_sdk`
- `flutter test test/src/comments_repository_test.dart` from `mobile/packages/comments_repository`
- `flutter test test/src/video_event_parsing_test.dart test/src/video_stats_test.dart` from `mobile/packages/models`
- `flutter analyze lib test` from `mobile/packages/nostr_sdk`
- `flutter analyze lib test` from `mobile/packages/comments_repository`
- `flutter test --coverage` from `mobile/packages/comments_repository`
- `flutter test --coverage` from `mobile/packages/models`
- `flutter test --coverage` from `mobile/packages/nostr_sdk`

### DIFF
--- a/mobile/packages/comments_repository/lib/src/comments_repository.dart
+++ b/mobile/packages/comments_repository/lib/src/comments_repository.dart
@@ -1172,27 +1172,30 @@ class CommentsRepository {
       String? imetaUrlCandidate;
       String? imetaMimeType;
 
-      for (var i = 1; i < tag.length; i++) {
-        final field = tag[i].trim();
-        if (field.startsWith('url ')) {
-          imetaUrlCandidate = field.substring(4).trim();
-        } else if (field.startsWith('m ')) {
-          imetaMimeType = field.substring(2).trim().toLowerCase();
-        } else if (field.startsWith('image ')) {
-          thumbnailUrl = field.substring(6).trim();
-        } else if (field.startsWith('dim ')) {
-          videoDimensions = field.substring(4).trim();
-        } else if (field.startsWith('duration ')) {
-          videoDuration = int.tryParse(field.substring(9).trim());
-        } else if (field.startsWith('blurhash ')) {
-          videoBlurhash = field.substring(9).trim();
+      parseImetaTag(tag, (key, rawValue) {
+        final value = rawValue.trim();
+        switch (key) {
+          case 'url':
+            imetaUrlCandidate = value;
+          case 'm':
+            imetaMimeType = value.toLowerCase();
+          case 'image':
+          case 'thumb':
+            thumbnailUrl = value;
+          case 'dim':
+            videoDimensions = value;
+          case 'duration':
+            videoDuration = int.tryParse(value);
+          case 'blurhash':
+            videoBlurhash = value;
         }
-      }
+      });
 
-      if (imetaUrlCandidate != null &&
+      final candidate = imetaUrlCandidate;
+      if (candidate != null &&
           ((imetaMimeType?.startsWith('video/') ?? false) ||
-              _isVideoUrl(imetaUrlCandidate))) {
-        videoUrl = imetaUrlCandidate;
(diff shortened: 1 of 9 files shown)


mobile/packages/comments_repository/lib/src/comments_repository.dart
```diff
@@ -1172,27 +1172,30 @@ class CommentsRepository {
...
-          videoBlurhash = field.substring(9).trim();
+      parseImetaTag(tag, (key, rawValue) {
+        final value = rawValue.trim();
+        switch (key) {
+          case 'url':
+            imetaUrlCandidate = value;
+          case 'm':
+            imetaMimeType = value.toLowerCase();
+          case 'image':
+          case 'thumb':
```
realmeylisdev, line 1183: nit, non-blocking: `thumb` is new here, the old loop only matched `image`. Worth a line in the PR body since it's a behaviour add riding along with the positional fix.

Precedence also differs from the feed. `video_event.dart` uses `thumbnailUrl ??= value` for both `thumb` and `image` so the first one wins there, this is a plain `=` so the last one wins. An event carrying both keys would show a different poster in the comment sheet than in the feed.

No impact today: across 958 imeta tags on the relay every single one uses `image` and none carry `thumb`, and none of them carry more than one imeta tag.

realmeylisdev (approved):
lgtm. ran the old and new parsers side by side over 958 real imeta tags off relay.divine.video (959 kind 34236 events spanning 2017 to today, 7 different clients) and they agree on every single one, so nothing regresses on live data. also built and ran it on the sim, feed enrichment and the comments sheet both fine.

one thing worth knowing: the positional path is hardening rather than a live fix, every imeta tag out there today is space separated. small note inline about `thumb`.

mobile/packages/comments_repository/lib/src/comments_repository.dart
NotThatKindOfDrLiz: Good call. I updated the PR Summary to call out that comments now accept `thumb` alongside `image` while parsing shared `imeta` metadata.

I left the existing comments-path precedence as-is for this PR: `image` was already last-wins here, and adding `thumb` just extends the same assignment path. The feed/parser precedence cleanup can ride with the remaining parser consolidation in #6961 if we decide to normalize it.

PR #6963, opened 2026-08-09
fix(dm): label retired moderation threads and close their dead composer

### LINKED ISSUE #6416: fix(dm): pre-rotation moderation notices lose their in-app entry point once the support pin de-duplicates them
## Problem

Once the pinned Divine Moderation row ships (#6388), `DmRepository.extractPinnedSupport` removes **every** known moderation conversation id — the current key plus every entry in `kLegacyModerationPubkeys` — from both the inbox list and the requests list, but only returns a **current-key** conversation as the pin:

https://github.com/divinevideo/divine-mobile/blob/1e60fd2eb9ce903e8eea368cbd6431fe7da4c439/mobile/packages/dm_repository/lib/src/dm_repository.dart#L4983-L5009

That is deliberate, and the deliverability half of it is right: `_PinnedSupportRow` routes on the adopted conversation's own participants and nothing remaps the recipient between the route's `extra` and `DmRepository.sendMessage`, so adopting a thread keyed on the retired `121b915baba659cbe59626a8afaf83b01dc42354dfecaad9d465d51bb5715d72` would address support replies to a pubkey the team retired.

The cost is that a user whose only moderation thread predates the rotation loses their in-app entry point to it. The thread stays in Drift; it just has no row and no route. `_mergeDuplicateConversations` groups by peer pubkey, so it was never merged into a current-key thread either.

## Who is actually affected — not who you would guess

The affected population is **not** "users who messaged support before the rotation". No shipped client ever sent a DM to the retired key:

- **Mobile.** The only pre-#6388 send path to moderation is the post-report DM, added 2026-03-12 in `a851831e6` (#2112) and reading `ModerationLabelService.divineModerationPubkeyHex`. That commit and the pubkey correction (`683577093`, #2321) are **both** first contained in tag `1.0.7`, and neither is in `1.0.6` — so the first release that could send a moderation DM already resolved NIP-05 with the current key as its fallback. In shipped mobile code the retired key was only ever a *label-filter* constant (introduced in `ba51840a2`, #1797).
- **Web.** `DIVINE_SUPPORT_PUBKEY` was a personal key (`78a5c21b5166dc1474b64ddf7454bf79e6b5d6b4a77148593bf1e866b73c2738`) from 2026-03-09 until `9d77c523` (divine-web#474, 2026-07-09) repointed it to the moderation account. The retired moderation key was never web's DM destination either.

The population that *does* have a retired-key thread is creators who received an **automated enforcement notice**. `divine-moderation-service` shipped NIP-17 DM notifications for `PERMANENT_BAN` / `AGE_RESTRICTED` / `QUARANTINE` on 2026-03-08 (`52bb81a`, `5456cd4`, #24), gated on and signed with `MODERATOR_NSEC`; `8dd56cb` (#31, 2026-03-15) moved all signing to `NOSTR_PRIVATE_KEY` and removed that secret. divine-iac-coreconfig `5b28724d` (#280, 2026-03-17) records the same rotation from the backend side — "Replaces old moderation pubkey (`121b91...`) with new pubkey (`8fd5eb...`)".

So the window is roughly 2026-03-08 → 2026-03-15, the direction is inbound-only, and those conversations sit in **Message Requests** today (`watchAcceptedConversations` filters on `currentUserHasSent`), or in the inbox proper if the creator replied.

## Why this is more than "a dead read-only thread"

Two of those notice templates (`divine-moderation-service src/nostr/dm-sender.mjs`) explicitly invite the creator to reply — the ban notice as the appeal channel, the quarantine notice for additional context. The dropped row is therefore the user's only in-app record of an enforcement action against them, and the channel that action told them to use.

## Why leaving the row in place is not free either

The composer in that thread publishes to the retired key. `dm-reader.getModeratorPubkey` derives the watched inbox from `NOSTR_PRIVATE_KEY`, so nothing has ingested DMs addressed to the retired key since 2026-03-15 — any appeal typed into that thread today is written to the relay and read by nobody. Keeping the row visible preserves the record but also preserves a silent dead end, in the one thread where a silent dead end is worst.

## Proposed fix

Surface the retired-key thread as an **archived, read-only** row rather than either hiding it or leaving a live composer on it:

1. Return legacy matches from `extractPinnedSupport` as an archived list instead of dropping them, so the list layer can render them below the pin.
2. Suppress the composer for those conversations and explain why, pointing at the pinned Divine Moderation row for anything new. `DmSendPolicy` (`mobile/packages/dm_repository/lib/src/dm_send_policy.dart`) is the right seam — it is already consulted at the lowest send primitive — but `DmSendPolicyDecision` currently carries only the protected-minor cases, so this needs a new decision case rather than reusing `terminallyBlocked`, whose copy would be wrong for an adult.
3. Copy + l10n for the archived label and the blocked-composer explanation (18 `app_*.arb` locales).

This is deliberately not folded into #6388: it needs a product call on copy and on whether an archived support thread is a row, a section, or a settings entry, and #6388 is two review rounds deep on the pin itself.

## Open questions

1. **How many creators are affected?** The count is in the moderation service's D1 `dm_log` for that window; it is not derivable from the relay, because gift wraps are signed by ephemeral keys and hide the real sender. Sizing question raised in `support-trust-safety#192`.
2. **Does anyone still hold the retired nsec?** If it is archived, adding it to `dm-reader`'s watched set would make the pre-rotation appeal channel readable again, which would change this from "archive the row" to "the thread still works". That is a backend call, also in `support-trust-safety#192`.

## Precedent

divine-web hit the same repoint-orphans-history question and accepted it: `docs/superpowers/specs/2026-07-07-protected-minor-dm-restriction-web-design.md` records "Support-key migration (orphan risk accepted, no bridge built — decided)", and names "grandfather `78a5c21b…` as read-allowed for inbound display only (never send-approved)" as the follow-up if it ever matters. That follow-up shape is what this issue is on mobile — with the difference that web's orphaning only bites protected minors, while #6388's de-dupe drops the row for everyone.

## Not a data-loss issue

Nothing is deleted. The conversation and its messages stay in Drift; only the row and the route go away, so whichever shape wins can be built later with no recovery step.

---

Relates to #6283. Follow-up from @NotThatKindOfDrLiz's review of #6388.

realmeylisdev:
## Description

A DM thread keyed on a moderation account we rotated away from was presented as an unidentified stranger, fitted with a live composer that published to an identity no Divine service has read since the rotation. For an affected user that thread is an enforcement notice whose own copy says *"you can reply to this message to appeal."*

Two independent defects, both keyed on the same missing idea — the app knows a moderation key is retired but never acted on it in the DM surface.

**1. The thread was not identifiable as Divine.** `isModerationAccount` recognition reached exactly one of five surfaces. The inbox tile applied the bundled wordmark; `RequestTile`, `RequestPreviewView`, the conversation header and the empty-conversation card called `UserAvatar` bare and resolved the peer as `profile?.bestDisplayName ?? defaultDisplayNameFor(pubkey)`. There is no kind-0 for that fallback to find — the moderation account's is not on the single relay production reads, and a retired key has no events at all — so the profile is always null and the name is always a generated *"Adjective Animal N"*. A `PERMANENT_BAN` notice arrived as **"Overwhelming Termite 20 wants to message you"** with a generic placeholder avatar and **"Decline and remove"** offered against it.

**2. The composer was a silent dead end.** `dmSendPolicyProvider` returned `allowed` immediately for any non-DM-restricted user, so `sendMessage`'s only gate passed: the appeal was gift-wrapped, published, and reported delivered with a green bubble. `divine-moderation-service`'s DM reader subscribes to exactly one pubkey derived from its current signing key, so nothing has read the retired identity since the rotation.

**Related Issue:** Closes #6416

### Note for reviewers: the issue's headline premise is stale

#6416 says `extractPinnedSupport` drops every legacy-key row. That was true of the intermediate revision the issue links, but #6388 merged the day after the issue was filed and ships **preserve-in-place**, pinned by four tests. Nothing is dropped, so the issue's proposed-fix step 1 is a no-op and there is no lost entry point to restore. The live defect **predates #6388** — its review surfaced it, nothing more. What this PR implements is the issue's own *"Why leaving the row in place is not free either"* section, which #6388 explicitly deferred here. A correction comment with the full detail is on the issue.

### Two deliberate divergences from the issue's proposal

- **No new `DmSendPolicyDecision` case.** The issue asks for one. `terminallyBlocked` already means "a confirmed policy blocks this recipient"; only its docstring was protected-minor-specific and only the *copy* was wrong. Keeping Divine's policy reasons out of `dm_repository` is exactly why `DmSendPolicy` is an injected typedef, so the docstring broadened and the app layer picks the copy from the recipient it already holds. Zero package-API churn, and the terminal classification is right — a retired recipient can never become deliverable, so the cold-start drain is correct to drop a queued row rather than retry forever.
- **The refusal is checked ahead of the DM-restriction branch**, because the hole is an adult one: a protected minor is already bounced off these threads by `ConversationPage`'s route guard, since retired keys are absent from `kPinnedOfficialAccounts`.

### Not requested by the issue

The labelling fix covers the **current** moderation key as well as retired ones. The requests flow was unlabelled for both, so any inbound-first moderation DM — the normal shape for an automated notice — hit this. Same line, same predicate; leaving it would fix an enforcement notice's identity only when the key behind it happens to be retired.

Copy deliberately promises no reply. The current support inbox is ingested but only lightly staffed, so "Message Divine Moderation" is a redirect, not an SLA.

## Out of Scope

- **"Decline and remove" stays available on an enforcement notice.** It deletes the user's own local copy and is their call, so it is unchanged here — but it reads badly on a ban notice and deserves a product opinion. Tracked in divine-mobile#6971.
- **Two questions remain open with the backend**, neither gating this change. How many creators hold a retired-key thread needs a read of the moderation service's `dm_log` (tracked in support-trust-safety#192); whether the retired nsec is archived is a key-custody call (tracked in support-trust-safety#199). If it is archived and the reader could watch it again, the thread genuinely works and the composer should reopen — that is a one-line flip of a single predicate, and the labelling half is correct under either answer. Closing is the reversible, fail-safe default.
- An accepted request from a peer the user has never messaged appears in neither the inbox list nor Message requests. Pre-existing behaviour of the requests flow, unrelated to this diff, observed while testing.

## Verification

Run from `mobile/` after rebasing onto `origin/main` (`d72613c38`):

- `flutter test test/screens/inbox test/providers test/config test/l10n test/blocs/dm` — 1479 passed
- `cd packages/dm_repository && flutter test` — 562 passed
- All four design-system ceilings plus the untested-services and `test/unit` structure ratchets — pass
- All four new keys translated in all 21 non-English locales (no `_knownUntranslatedDebt` entries added); regenerating l10n after the rebase produces no diff

Two of the new policy tests were rewritten after they passed with the guard removed: the short-circuit assertion only discriminates when run restricted, and a terminal-vs-temporary assertion at the provider layer could not fail, so it moved to the package where the distinction is actually enforced.

### Verified end to end on an iOS simulator

The real thread cannot be forged — NIP-17 binds it through the kind-13 seal, which needs the retired nsec — so a controllable stand-in key went into `kLegacyModerationPubkeys` in a local-only patch, reverted before commit; the only difference from production is which hex string is in the list. The real `PERMANENT_BAN` template was seeded as an inbound NIP-17 DM over a local relay, and the app persisted it into a conversation whose id matches `sha256` of the sorted participants exactly, so the thread is keyed on the retired peer.

| Surface | Before | After |
|---|---|---|
| Message requests tile | "Overwhelming Termite 20" + placeholder avatar | "Divine Moderation" + wordmark |
| Request preview | same, with "Decline and remove" | relabelled in header, avatar, and body copy |
| Conversation header | "Overwhelming Termite 20" | "Divine Moderation" |
| Composer | live "Say something…" | closed-thread notice + redirect |

Tapping the redirect routed to `/inbox/conversation/dba3d4b4…` — the derived current-moderation-key thread, confirmed against an independently computed id — and that thread keeps its live composer.

**The last commit is a bug this device run caught and the tests did not**: the redirect's destination has no messages yet, and the empty-conversation card renders its own avatar, so the header read "Divine Moderation" above the generic orange placeholder. Fifth surface, same predicate, now covered by tests.

## Type of Change


NotThatKindOfDrLiz:
@realmeylisdev I pushed one takeover commit (cd5c3a27a) to close out the review gaps:

- retired moderation rows now show the existing closed-thread subtitle in the inbox and Message Requests, so the preserved retired thread is distinguishable from the live pinned Divine Moderation row
- retired moderation conversations now disable reaction creation from both double-tap and long-press quick reactions, while keeping non-send message actions available
- added regressions for the side-by-side live/retired rows, retired request tiles, and retired-thread reaction affordances
- linked follow-ups in the PR body: divine-mobile#6971 for the "Decline and remove" product/copy question, support-trust-safety#192 for dm_log/intake sizing, and support-trust-safety#199 for retired-key custody

Verification passed locally: `flutter test test/screens/inbox/widgets/conversation_tile_test.dart test/screens/inbox/message_requests/widgets/request_tile_test.dart test/screens/inbox/inbox_view_test.dart test/screens/inbox/conversation/conversation_view_test.dart`; `flutter analyze lib test integration_test`; pre-push analyze + changed-file tests.

GitHub checks are all green. Please review the takeover commit and merge when you're ready.

NotThatKindOfDrLiz (approved):
Approved after takeover.

I verified the final diff after pushing cd5c3a27a. The retired moderation row ambiguity is resolved with the closed-thread subtitle, reaction creation is gated in retired threads, follow-ups are tracked, and all GitHub checks are green.

PR #6964, opened 2026-08-09
feat(profile): sort the follower and following lists, newest first by default

### LINKED ISSUE #6965: feat(followers): sort the follower list, newest first by default
Auto-created from PR #6964.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6964
**Branch:** `feat/followers-sorted-newest-first`

hm21:
Closes #6965

## Description

Neither follow list had a meaningful order. For followers, `_fetchFollowers` merged REST, connected-relay and indexer results into a `Set<String>`, so the order was "whichever source answered first" and a brand-new follower could land anywhere in the list. For following, the contact list's `p` tags were rendered verbatim, which runs oldest-follow-first. There was no way to spot who just followed you — which matters more than usual right now, because new-follower notifications aren't working.

Four commits:

**1. Order followers newest first.** Both relay paths already fetched complete kind 3 events and discarded everything except `event.pubkey`. They now carry `created_at` through, and the merge sorts on it, freshest timestamp winning when sources disagree.

**2. Add a sort control to the followers screen.** A `funnelSimple` button in the app bar — the same one the library screen uses — opens a `VineBottomSheetSelectionMenu` with *Newest first* / *Oldest first*.

**3. Do the same for the following list.** Same button, same sheet, same default. The following list needs no timestamps: kind 3 `p` tags are appended as you follow, so the tag order already runs oldest to newest and *Newest first* is a reversal of it.

**4. Keep the dated boundary in step with the primed list.** `_streamFollowers` (the user-picker's fast path) wrote the ordered pubkeys into the in-memory cache but not the boundary that describes them, so `watchMyFollowers`'s cached emission could pair a fresh list with a boundary from a different one. Found in review.

### Why it looks like this

**`created_at` is an approximation, and that is the best the protocol offers.** Kind 3 is replaceable, so `created_at` is when a follower last edited their contact list, not when they followed you. For someone who *just* followed, the two are the same moment — which is exactly the case this exists for. Someone who followed a year ago and followed somebody else this morning will also float up. There is no per-follow timestamp in Nostr to use instead, and the Funnelcake API cannot substitute one: `/api/users/{pubkey}/followers` returns bare pubkeys and no timestamps. It does order its page — the query ends in `ORDER BY rf.created_at DESC` — but it exposes neither that timestamp nor the direction, so the ordering is invisible to a client and every REST follower arrives here undated.

**The following list's order is only as good as the last client that wrote the contact list.** Divine appends new follows to the end of `_followingPubkeys` and publishes in that order, and both merge paths build from a `Set` literal that preserves insertion order — so for a list Divine last wrote, tag order *is* follow order. A client that rebuilds the tag list instead of appending destroys that, and the reversal then just flips whatever order that client happened to choose. There is no timestamp to fall back on, because a contact list carries one `created_at` for the whole list, not one per entry.

**Undated followers stay at the end in both directions.** The REST source answers with pubkeys and no timestamp. Those keep their arrival order behind every dated follower rather than being assigned a guessed time. Naively reversing the whole list for "oldest first" would push them to the top, and "we don't know when" is not the same claim as "a long time ago".

**`FollowersSnapshot.datedCount` instead of a timestamp map.** Because `_newestFirst` sorts every dated follower ahead of every undated one, a single integer marks where the undated tail begins — enough to flip the order correctly. A per-follower timestamp map would have carried the same information at roughly double the cached payload (~327 KB → ~650 KB at 5,000 followers).

**The sort choice is not persisted across sessions.** Newest-first is the useful default on every visit; restoring "oldest first" days later would hide exactly the new follower the screen exists to surface.

**`fetchMyFollowersSnapshot` / `fetchFollowersSnapshot` are now the test seam** for the cached watchers, replacing `getMyFollowers` / `getFollowers`. A plain `List<String>` can't carry `datedCount`, so the subclass fakes in `follow_repository_cached_test.dart` had to hook one level up.

### Performance

Measured over the real fetch path with a throwaway benchmark (macOS; a phone is roughly 3–5× slower):

| Followers | Fetch + sort | Newest ↔ oldest flip | Cache payload |
|---|---|---|---|
| 1,000 | 10 ms | 0.5 ms | 65 KB |
| 5,000 | 7 ms | 0.3 ms | 327 KB |
| 20,000 | 14 ms | 1.1 ms | 1.3 MB |

(The 1,000 row being slower than 5,000 is JIT warm-up in the first case, not a real effect.)

This makes the relay path **faster**, incidentally. Its old dedup was `List.contains` inside a loop — O(n²), measured in isolation at 46 ms for 5,000 followers and **864 ms for 20,000**, blocking the main isolate. Dedup now happens in the merge's hash map, O(n).

The cache budget is 100 MB with LRU eviction, so a 327 KB snapshot is 0.3 % of it. Blocklist filtering is `Set`-backed, O(1) per entry. The list is a `ListView.builder`.

## Known limits

**"Oldest first" on the followers list cannot reach your actual oldest follower.** Only the dated prefix flips, and that prefix is whatever the relays returned. The kind 3 query carries no explicit limit, so each relay applies its own default cap — and per NIP-01 a capped result is the *most recent* events. So the flippable subset is newest-biased by construction: past that cap, "Oldest first" shows the oldest of your newest few hundred rather than your true oldest. The list is honest about ordering what it has, but the label promises more than the data can deliver.

The undated tail is smaller than it looks, though, and in this feature's favour. The REST source asks for `limit: 5000` but funnelcake clamps that endpoint to 100 (`query.limit.unwrap_or(50).min(100)`) and the client does not page it, so at most 100 followers arrive undated no matter how many you have. The relay-dated prefix does the rest of the work.

[divinevideo/divine-funnelcake#883](https://github.com/divinevideo/divine-funnelcake/pull/883) is the server-side fix, opened alongside this PR. It adds `?sort=newest|oldest` to `/api/users/{pubkey}/followers`, exposing the ordering the query already computes. Once it ships, the mobile side can ask for the true oldest page instead of flipping a newest-biased local subset — a follow-up that passes `sort=oldest` and reads the response's `sort` echo, with no redesign here. It is deliberately not a blocker: this PR is useful on its own for every account under the relay cap, which is nearly all of them.

## Out of Scope

- **Pagination.** The list is truncated far lower than the `limit: 5000` in the REST call suggests: funnelcake clamps that endpoint to 100 and there is no paging on the call, so REST contributes at most 100 followers, and the relay query runs into each relay's own default cap on top of that. Same trap #6109 fixed on the following side — see the `_restFollowingMaxPubkeys` comment. The header count comes from a separate COUNT query and stays correct; the rows do not. Pre-existing ceiling, but worth naming because this feature makes it visible, and the `sort=oldest` follow-up will want the real number.
- **Other people's follower lists.** `getFollowers` shares the fetch path, so those are ordered newest-first too, but only the own-followers screen gets the sort button. Easy to extend if wanted.
- **A relative "followed 2h ago" label per row.** Would make the ordering legible, but given the `created_at` caveat above the label would be misleading often enough that it needs a product call first.
- **New-follower notifications**, which prompted this but are a separate bug.

**Related Issue:** 

## Verification

- `flutter test --coverage` in `mobile/packages/follow_repository` — 297 tests pass, **100 % line coverage held** (hard CI gate on that package)
- `flutter test test/blocs/my_followers test/blocs/my_following test/screens/followers test/screens/following test/widgets/user_picker_sheet_test.dart test/features/people_lists` — 203 tests pass, including new widget tests covering the app bar button, the sheet contents, the re-order, and dismissing the sheet without choosing
- `dart format` — clean
- All six ratchets green: raw dialog, material button, raw textstyle, raw colors, untested services, test/unit structure
- All 21 non-English locales carry real translations for the five new keys — no English fallbacks
- Rebased onto current `origin/main` (upstream moved 7 commits mid-work, including #6914 which touches every ARB file; the l10n merge is purely additive)
- Tested on a real device (Samsung Galaxy)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -1171,13 +1303,10 @@ class FollowRepository {
...
-      final followers = <String>[];
-      for (final event in events) {
-        if (!followers.contains(event.pubkey)) {
-          followers.add(event.pubkey);
-        }
-      }
-      return followers;
+      return [
+        for (final event in events)
+          (pubkey: event.pubkey, followedAt: event.createdAt),
```
realmeylisdev: The sort key is now the follower's own `created_at`, and nothing clamps it.

`Event.fromJson` takes `created_at` straight off the wire (`nostr_sdk/lib/event.dart:49`), the indexer path does the same at :1373, and `_newestFirst` sorts on it directly. So a kind 3 signed with `created_at` in 2100 sits at the top of every follower list it appears in, and stays there. The three indexers are public third-party relays — `purplepag.es`, `user.kindpag.es`, `relay.nos.social` (`environment_config.dart:174`) — so their clock-skew policy isn't ours to lean on.

Before this PR the order was arbitrary, so there was nothing to game. Now the top slot is author-controlled.

Cheapest guard is to treat a future timestamp as undated — it drops into the tail you already built for "we don't know when", and `datedCount` handles the rest:

```dart
final now = DateTime.now().millisecondsSinceEpoch ~/ 1000;
return [
  for (final event in events)
    (
      pubkey: event.pubkey,
      followedAt: event.createdAt > now ? null : event.createdAt,
    ),
];
```

Low severity — it's ordering, not data — but it's much cheaper now than after someone notices.

mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -1026,60 +1060,133 @@ class FollowRepository {
...
-  /// [_fetchFollowers] indexes into the results to log per-source counts.
+  /// [_fetchFollowers] indexes into the results to log per-source counts, and
+  /// because [_newestFirst] falls back to arrival order for the REST source,
+  /// which carries no timestamp.
   /// Failures are *not* swallowed here; each caller applies its own policy.
-  List<Future<List<String>>> _followerSources(String pubkey) {
+  List<Future<List<_FollowerRef>>> _followerSources(String pubkey) {
     final apiFuture = (_funnelcakeApiClient?.isAvailable ?? false)
         ? _funnelcakeApiClient!
               .getFollowers(pubkey: pubkey, limit: 5000)
```
realmeylisdev, line 1209: Not introduced here, but it changes a number in your Out of Scope section.

funnelcake caps this endpoint at 100: `get_user_followers` does `query.limit.unwrap_or(50).min(100)` (divine-funnelcake, `crates/api/src/handlers.rs`). There's no paging on this call, so the REST source contributes at most 100 followers, not 5,000.

It's the same trap #6109 fixed on the following side — the comment on `_restFollowingMaxPubkeys` at :1943 spells it out: "Asking for 5000 and silently receiving 100 is what let a truncated list become the base for the next publish."

Effect on this feature is mildly good — the undated tail is short and the dated prefix does most of the work. But "Above ~5,000 followers the list is already truncated" is off by roughly 50x, and the follow-up that adds `sort=oldest` will want the real ceiling.

mobile/packages/follow_repository/lib/src/follow_repository.dart
```diff
@@ -1026,60 +1060,133 @@ class FollowRepository {
...
       ...apiFollowers,
       ...relayFollowers,
       ...indexerFollowers,
-    };
+    ]);
 
     Log.info(
       'Followers for $pubkey: '
       'API=${apiFollowers.length}, '
       'relays=${relayFollowers.length}, '
```
realmeylisdev: These two counts quietly changed meaning. `_fetchFollowersFromRelays` dropped its `contains` dedupe and `_fetchFollowerRefsFromIndexers` now concatenates, so `relays=` and `indexers=` count raw events while `merged=` still counts unique followers. A follower whose superseded kind 3 is still being served inflates both.

The `Indexer follower pubkeys:` line at :1339 *does* dedupe, so the two log lines now disagree about the same quantity.

`${relayFollowers.map((r) => r.pubkey).toSet().length}` on each would keep the four numbers comparable again.

mobile/packages/follow_repository/lib/src/follow_sort_order.dart
```diff
@@ -0,0 +1,57 @@
...
+
+  /// Arranges a followers list for this order.
+  ///
+  /// [pubkeys] must be newest-first as `FollowRepository` returns it, with its
+  /// [datedCount] dated followers leading — `FollowersSnapshot.datedCount`
+  /// carries that boundary.
+  ///
+  /// Only the dated prefix flips: followers with no timestamp stay at the end
+  /// in both directions, because "we don't know when" is not the same claim as
+  /// "a long time ago".
```
realmeylisdev, line 24: Worth naming the asymmetry this leaves behind.

funnelcake's followers page is `ORDER BY rf.created_at DESC` (`crates/clickhouse/src/client.rs`), so the undated tail is itself newest-first. Under `oldestFirst` the tail therefore runs in exactly the opposite direction from the label above it.

"We don't know when" vs "a long time ago" justifies *where* those rows sit — it doesn't cover which way they run once they're there. Not asking to change it: the response carries no `sort` echo today, which is precisely what divinevideo/divine-funnelcake#883 fixes. But the doc reads as if the tail has no order at all, and it does have one.

mobile/packages/follow_repository/lib/src/follow_sort_order.dart
```diff
@@ -0,0 +1,57 @@
...
+    };
+  }
+
+  /// Arranges a following list for this order.
+  ///
+  /// [pubkeys] must be in follow order as `FollowRepository` returns it: the
+  /// `p` tags of the user's own kind 3 contact list, verbatim. New follows are
+  /// appended, so that order runs oldest to newest and needs no timestamps.
+  ///
+  /// The order is only as good as the last client that wrote the contact list.
```
realmeylisdev: NIP-02 is stronger than this paragraph gives it credit for:

> Whenever new follows are added to an existing list, clients SHOULD append them to the end of the list, so they are stored in chronological order.

Worth citing right here. It turns the reversal from "Divine happens to append" into a spec-sanctioned reading, and it gives you a better answer than the current wording if anyone questions the design later. The caveat below stays true — it just becomes "a client that violates NIP-02" rather than "a client that does it differently".

mobile/lib/blocs/my_following/my_following_bloc.dart
```diff
@@ -32,7 +32,8 @@ class MyFollowingBloc extends Bloc<MyFollowingEvent, MyFollowingState> {
                ? MyFollowingStatus.initial
                : MyFollowingStatus.success,
            rawFollowingPubkeys: followRepository.followingPubkeys,
-           followingPubkeys: followRepository.followingPubkeys
+           followingPubkeys: FollowSortOrder.newestFirst
+               .fromFollowOrder(followRepository.followingPubkeys)
```
realmeylisdev, line 36: This flips my following list to newest-first, but `OthersFollowingBloc` applies no ordering at all, so another user's following list still renders contact-list order — oldest first. The two "following" screens now disagree with each other.

Out of Scope names other people's *follower* lists (which do get the new order for free, since they share the fetch path) but not this one. Deliberate, or did it just fall between the two?

realmeylisdev (approved):
Approving — I couldn't find a correctness problem in this.

What I did past reading the diff:

- Lifted `_newestFirst` into a standalone harness and ran 200k randomized inputs against an independent stable-partition model. Order and `datedCount` agree every time, and `datedCount` lands exactly on the dated/undated boundary in every case. The `Map`-with-null-values trap is handled right — `containsKey` for first-seen, not `[] != null`.
- Both blocs order before they filter, so the blocklist can't drag the boundary out from under the flip.
- Checked your funnelcake read against the source: the followers page really is `ORDER BY rf.created_at DESC` and really does drop the timestamp on the way out. And divine-funnelcake#883 does add the `sort` echo you describe.
- NIP-02 backs the following-list reversal harder than your doc does — noted inline.
- Locally: package suite green at 100.00% line coverage (802/802), 203 app tests, `flutter analyze` clean, ARB consistency green, and all 21 non-English locales carry real translations, not English fallbacks.

Six notes inline. One is worth a look before merge — the sort key is now an unvalidated `created_at`. Two are pre-existing and one of those corrects a number in your Out of Scope section. The rest are doc and logging.

I could not get this onto an iOS device: a SwiftPM cache lock on my machine, contended by builds in unrelated worktrees, deadlocked the Xcode build for 20+ minutes. That's my environment, nothing to do with this branch — flagging it so nobody reads "approved" as "exercised on device".

realmeylisdev:
Correction to my review: the iOS build did finish — I called it dead a few minutes too early. Ran it on the simulator against production, signed in, and drove both screens. Everything below is from that session.

**Ordering is exactly right.** I pulled the ground truth off the relays with `nak` first, then compared row by row:

| Screen | Rows | Checked against |
|---|---|---|
| Followers, newest first | 7/7 exact | kind 3 `created_at`, descending |
| Followers, oldest first | 7/7 exact | same, ascending |
| Following, newest first | 9/9 exact | reversed `p` tag order |
| Following, oldest first | 9/9 exact | raw `p` tag order |

No dropped rows, no duplicates, sheet dismisses cleanly, re-picking the active order is a no-op. Nothing in the log relates to this change.

**Two of my inline notes reproduced live.**

The merge log, first fetch on an account with 7 followers:

```
Followers for 9be8bd90...: API=6, relays=6, indexers=19, merged=7
```

`indexers=19` for 7 people — the three indexers each returned overlapping kind 3 events and nothing dedupes before the count is formatted. Roughly 2.7x inflated here; worse on a real account.

And the REST cap, straight from the API:

```
$ curl "https://api.divine.video/api/users/9be8bd90.../followers?limit=5000"
{"followers":[...6 items...],"limit":100,"offset":0,"total":6}
```

The response echoes its own clamp.

**One thing the run did not cover.** Every follower on this account came back from a relay with a timestamp, so `datedCount` was 7 of 7 and the undated tail never appeared. The `datedCount < length` path is well covered by your unit tests, but I did not see it on screen — worth a pass on an account whose followers only resolve through REST, if you have one handy.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Fixed in cf963b47f3 — confirmed, `Event.fromJson` takes it raw and the indexer path does too.

Went with demotion to the undated tail rather than a clamp to now, as you suggested: clamping still leaves the offender at the very top, which is the part worth preventing. Allowance is 5 minutes, tighter than `DmSyncState.maxFutureSkewSeconds` — a DM has to be kept and can only have its ordering clamped, while a follower demoted here still appears, just in the tail. Two tests: one for the far-future demotion, one proving a minute of honest skew keeps its position.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Correct, and I can confirm from the other side — `query.limit.unwrap_or(50).min(100)` is exactly what I read in `get_user_followers` while building divinevideo/divine-funnelcake#883.

No code change: the `limit: 5000` is harmless, just optimistic. Both PR body sections are rewritten, and it cuts in this feature's favour — the undated tail is ≤100 no matter how many followers you have, so the relay-dated prefix does more of the work than "Known limits" claimed. Also noted for the `sort=oldest` follow-up.

mobile/packages/follow_repository/lib/src/follow_repository.dart
hm21: Fixed in 0897470eba — a regression this branch introduced, thanks for catching it. Your `indexers=19` for 7 followers is the reproduction.

All four numbers go through one `uniqueCount` helper now, so `API=`, `relays=`, `indexers=` and `merged=` are comparable again and agree with the `Indexer follower pubkeys:` line.

mobile/packages/follow_repository/lib/src/follow_sort_order.dart
hm21: Fixed in cfd95756d4 — you're right, the doc claimed more than it should. "We don't know when" justifies where the tail sits, not which way it runs.

The comment now says the tail comes from the REST source, that funnelcake serves that page `ORDER BY rf.created_at DESC`, and that it therefore stays newest-first under `oldestFirst` — running against the label above it. Links #883 for the `sort` echo that lets a client know.

mobile/packages/follow_repository/lib/src/follow_sort_order.dart
hm21: Fixed in cfd95756d4 — quote verified against the spec and cited verbatim.

You're right that it is the stronger argument: it makes the reversal spec-sanctioned rather than an observation about how Divine happens to write its own list, and the caveat below now reads "a client that violates that SHOULD".

mobile/lib/blocs/my_following/my_following_bloc.dart
hm21: Fell between the two — not deliberate. Fixed in e6fa309d62.

`OthersFollowingBloc` now applies the same newest-first default. No sort control on that screen; it just has to agree with `MyFollowingBloc`. Ordering runs before filtering as in the other three blocs. An existing test asserted the old order and now asserts the new one, which is the change being observable.

hm21:
@realmeylisdev thanks — all six addressed, four commits on top of `e6fa309`. Checks green, approval survived the push.

| Finding | Outcome |
|---|---|
| Unclamped future `created_at` | Fixed, `cf963b47f3` — demoted to the undated tail, not clamped to now |
| REST caps at 100, not 5000 | Docs only, PR body rewritten — no code change needed |
| Per-source log counts | Fixed, `0897470eba` — regression this branch introduced |
| Undated tail has its own order | Fixed, `cfd95756d4` |
| NIP-02 citation | Fixed, `cfd95756d4` |
| `OthersFollowingBloc` unordered | Fixed, `e6fa309d62` — not deliberate, it fell between the two |

I checked each claim against source before acting rather than taking the review at face value, and all six held up. The funnelcake `min(100)` I could confirm first-hand — I read that line building divinevideo/divine-funnelcake#883.

Two things worth flagging back:

**The 100-cap finding improves the feature's story rather than just correcting a number.** With REST contributing ≤100 undated followers regardless of account size, the relay-dated prefix does more of the work than my "Known limits" section claimed. Both that section and Out of Scope are rewritten.

**Your `datedCount < length` gap is still open.** I have no device, so I could not cover it either — it stays unit-tested only. Worth a pass on an account whose followers resolve mostly through REST if one turns up, but I would not hold the PR for it.

Threads left unresolved for you to close.

Issue #2513, opened 2026-03-26
feat(nostr-sdk): add outbox model with relay routing at SDK level


dcadenas:
## Description

Our Dart nostr libs (`nostr_sdk`, `nostr_client`) lack relay routing intelligence. `RelayDiscoveryService` discovers NIP-65 relay lists but nothing in the SDK uses that information to route queries. Every relay selection decision is manual — repositories pass `tempRelays`/`targetRelays` by hand.

NDK (TypeScript) has a full outbox tracker at the SDK level that maintains a cache of pubkey -> write/read relays and methods like `calculateRelaySetsFromFilter()` that automatically route queries to the right relays per author.

## What's needed

1. **Outbox Tracker** in `nostr_sdk` or `nostr_client` — cache of pubkey -> {writeRelays, readRelays} from kind 10002 events
2. **Relay selection logic** — when querying for events by author, automatically select the author's read/write relays
3. **Filter-to-relay mapping** — split multi-author filters into per-relay subsets for optimal routing
4. **Works for all kinds** — profiles (kind 0), videos (kind 34236), posts, etc.

## Reference implementations

- NDK outbox tracker: `~/code/ndk/core/src/outbox/index.ts`
- NDK relay set calculation: `~/code/ndk/core/src/relay/sets/calculate.ts`

## Context

Explored during PR #2504 (parallel profile fetch). The outbox model was initially implemented at the repository level but removed because it belongs at the SDK level — the indexer relays that would serve NIP-65 data almost certainly have the kind 0 profiles too, making a repo-level outbox lookup redundant for profiles. The real value of outbox routing is for content discovery (videos, posts) on the author's write relays.

PR #6966, opened 2026-08-09
fix(bookmarks): reconcile with the relay before reading or publishing kind 10003

### LINKED ISSUE #6627: fix(bookmarks): BookmarkService.initialize() is never called, so bookmarks never sync from the relay
### Summary

`BookmarkService.initialize()` is never called anywhere in the app, so bookmarks are never reconciled with the relay. The Saved tab reflects only the SharedPreferences copy plus in-session edits.

### Environment

App version: current `main` (`ccdd847`)
Device: all platforms
Network: any

### Steps to Reproduce

1. Save a few videos on device A.
2. Sign in with the same account on device B (or reinstall on device A and sign back in).
3. Open Profile → Saved.

### Actual Result

Saved is empty (or holds only whatever that install saved locally). The kind-10003 list on the relay is never read.

`bookmarkServiceProvider` constructs the service and returns it:

```dart
// mobile/lib/providers/repository_providers.dart:505-516
@riverpod
Future<BookmarkService> bookmarkService(Ref ref) async {
  ...
  return BookmarkService(nostrService: ..., authService: ..., prefs: ...);
}
```

`BookmarkService.initialize()` is the only caller of `_loadBookmarksFromNostr()` (`mobile/lib/services/bookmark_service.dart:717-767`), and it has **no callers**. Grepping `lib/`, `test/` and `integration_test/` for a `BookmarkService` + `initialize` co-occurrence returns nothing outside the service's own file. The constructor only runs `_loadBookmarksFromSharedPreferences()`, so `globalBookmarks` is the local snapshot plus in-session appends, with nothing to ever correct it.

Knock-on: the doc comment on the consumer is stale and asserts behaviour that never runs.

```dart
// mobile/lib/blocs/profile_saved_videos/profile_saved_videos_bloc.dart:124-126
/// Reads the event-type bookmark IDs from [BookmarkService]. The service
/// keeps a SharedPreferences-cached list and reconciles with the relay on
/// initialize, so this read is cheap (no relay round-trip).
```

### Expected Result

Bookmarks reconcile with the user's kind-10003 list, so Saved survives a reinstall and matches across devices.

### Regression?

Unknown

### Evidence

Found while reviewing #6621, which fixes a different cause of a short Saved window (the reconcile window locking below one page). This is the upstream reason a Saved ID list can be short in the first place, and it is independent of that fix — worth confirming whether the relay read was intentionally dropped or simply never wired up.

Two follow-ups if the wiring is restored: pick where `initialize()` runs so it does not add a relay round-trip to first paint, and re-check `_parseGlobalBookmarksFromEvent` (`bookmark_service.dart:770-784`) which clears then refills from a single event.

### LINKED ISSUE #6968: fix(bookmarks): bookmark sets (kind 30003) publish without reading the current version
`BookmarkService._publishBookmarkSetToNostr` builds a kind-30003 from the in-memory set and publishes it without first fetching the current version, exactly as the kind-10003 path did before #6966.

Kind 30003 is addressable (30000–39999), so the new event replaces the previous one for that `d` tag. `divine-plugin/rules/replaceable-event-preservation.mdc` line 22 covers addressable events explicitly, and lines 44–45 require read-before-write plus a warning rather than a blind publish when the current version can't be fetched.

Not yet reproduced on device — the kind-10003 case was, in #6627, and the code shape here is the same. Worth confirming before fixing.

#6966 deliberately left this alone to keep the data-loss fix reviewable. The reconcile primitive it added (`syncGlobalBookmarks()`) is the shape to copy.

Relates to #6627, follows #6966.

realmeylisdev:
## Description

`BookmarkService.initialize()` was never called, so the Saved tab only ever showed the SharedPreferences cache and the user's kind-10003 list was never read. That is the filed bug.

Investigating it on device surfaced a worse one that isn't in the issue: because a Save republishes the local-only list into a **replaceable** event, the first Save after a reinstall **destroys the user's whole bookmark list on the relay**.

Reproduced against production, before the fix:

| step | relay kind-10003 |
|---|---|
| saved a video | `[05dc0944…]` |
| cleared only the local bookmark cache (sign-in intact — the reinstall state) | unchanged |
| opened Saved | **"Nothing saved yet"**, 82 ms, no relay query |
| saved a *different* video | `[b447603f…]` — **`05dc0944…` destroyed** |

`divine-plugin/rules/replaceable-event-preservation.mdc` already covers this class (line 22: "any replaceable event (kind 0, 3, **10000-19999**)"; lines 44–45: read before write, merge rather than overwrite, warn rather than publish when the current version can't be fetched). Its BAD example is structurally the same code.

**The fix** is one reconcile primitive, `syncGlobalBookmarks()`, used on both sides:

- **Read** — `ProfileSavedVideosBloc._resolveSavedIds()` reconciles before reading, which is where the Liked and Reposted tabs already call `syncUserReactions()` / `syncUserReposts()`. The stale-while-revalidate path and revalidation bar were already there; only the relay read was missing.
- **Write** — the mutation path reconciles first and **refuses to publish if that fails**. It uses `queryEventsDetailed` so a timeout or an unreachable relay is not mistaken for "you have no bookmarks" — that confusion is exactly what truncates the list.

Two smaller defects fixed in the same lines:

- `content` was the prose string `"Divine global bookmarks"`. NIP-51 reserves `.content` for the NIP-44-encrypted private-item array, so Divine was corrupting private bookmarks written by other clients (Amethyst/Primal/Circl lists on our relay carry ciphertext there). It is now empty, or the ciphertext read back, carried through untouched.
- The query was `Filter(authors:[pk])` with no `kinds` and no `limit`, fetching every event the user ever published to find one replaceable event — measured at 2230 events / 3.8 s for an active account against a 5 s timeout. Now `kinds:[10003], authors:[pk], limit:1`.

Also removes the never-called `initialize()`, its dead relay-load path, and `NostrListServiceMixin` (its only user), plus the doc comment claiming the service "reconciles with the relay on initialize" — true only because it never ran.

**Related Issue:** Closes #6627, Closes #6968

### Also in this PR: deleting the dead kind-30003 code

I filed #6968 believing bookmark **sets** had the same read-before-write flaw. They do — but `_publishBookmarkSetToNostr` is **unreachable**, so no user data was ever at risk. Every public entry point (`createBookmarkSet`, `addToBookmarkSet`, `removeFromBookmarkSet`, `updateBookmarkSet`, `deleteBookmarkSet`, `getBookmarkSetById`, `isInBookmarkSet`, `publishBookmarkSetToNostr`, `getBookmarkSetsContainingVideo`, `isVideoInAnyBookmarkSet`, `bookmarkSets`, `bookmarkSetsStorageKey`) has **zero** references across `mobile/lib`, `test`, `integration_test` and `packages`. The "Lists" tab that looks like it might use them is the curated-list feature — kind 30005, `curated_list_service.dart` — a different thing.

So the right action was deletion, not another reconcile path. It's folded in here rather than stacked because it's the same file this PR already rewrites, and a separate PR would guarantee a modify/delete conflict. Kept as its own commit (`4538b72d5`) so it can be reverted alone.

Also drops five ARB keys translated into 22 locales for UI that was never built. `bookmark_service.dart`: 992 → 506 lines.

## Reviewer commit on this branch

`40b47fd7b` was pushed by a reviewer, on top of `3ba08a18c`. It closes one gap
in the reconcile and leaves the rest of the PR untouched.

A non-authoritative empty answer is no longer believed on its own. The read
path took the settle window's latency trade, so one relay answering EOSE-empty
while another stayed mute produced `timedOut: false` with zero events — and
that was persisted as "you have no bookmarks", overwriting the
SharedPreferences snapshot Saved falls back to offline. On a fresh install it
also reproduced #6627 itself: Saved rendered "Nothing saved yet" for a user who
had bookmarks. An empty answer is now confirmed with a full-settlement query
before it is believed, and an answer that stays unconfirmed changes nothing.

The confirmation is bounded so it does not become a tax. It always runs while
the device holds bookmarks; otherwise at most once per
`BookmarkService.absenceConfirmationTtl` (5 minutes), counted from the last
confirmed absence and reset whenever a list is actually seen. A read that finds
a list pays for none of it. The authoritative read also passes
`useCache: false`, so a stale locally cached kind-10003 cannot stand in for a
relay answer on the write path.

Two behaviour changes worth knowing before the first support report:

- A connected-but-mute relay now makes a save **fail** rather than truncate the
  list. That is the fix working as intended, but it is user-visible and there
  is no counter behind it.
- On a fresh install whose pool contains a mute relay, the first Saved open
  pays the fast read plus the full query timeout and then still renders empty.
  Once per TTL window, and correct in preference to the wrong answer.

## Out of Scope

Filed separately rather than bundled here:

- Reading other clients' NIP-51 **private** (NIP-44-encrypted) items. This change stops corrupting them; actually reading them needs a signer decrypt round trip (Keycast RPC for OAuth accounts) and is a product/latency call.
- Extracting a `bookmarks_repository` package so blocs stop consuming an app-layer service directly.

## Verification

`flutter analyze lib test integration_test` — clean. Targeted suites — 109 passing (incl. `arb_consistency_test`) (`bookmark_service`, `profile_saved_videos_bloc`, `share_sheet_bloc`, `share_video_menu_comprehensive`). `check_untested_services_floor.sh` and `check_test_unit_structure.sh` — green.

New `test/services/bookmark_service_test.dart`; the service had no test file at all. Verified the tests can fail: removing the reconcile from the add path turns exactly the three add-path tests red.

Re-ran the reproduction above on device against production, after the fix:

| step | relay kind-10003 |
|---|---|
| saved a video | `[833f9a2b…]` |
| cleared the local bookmark cache, relaunched | unchanged |
| opened Saved | **renders the bookmark**, `Synced 1 global bookmarks from relay` |
| saved a *different* video | `[833f9a2b…, 8685bc50…]` — **both kept** |

`content` on the published event is now `''`.

## Type of Change


mobile/lib/services/bookmark_service.dart
```diff
@@ -146,105 +102,159 @@ class BookmarkService with NostrListServiceMixin {
...
-      _loadBookmarksFromSharedPreferences();
-
-      // 2. Load from relay (authoritative)
-      await _loadBookmarksFromNostr();
+      final events = result.events
+          .where((event) => event.kind == globalBookmarksKind)
+          .toList();
+      if (events.isEmpty) {
+        // The relay answered and has nothing: a genuinely empty list.
+        _globalBookmarks.clear();
```
NotThatKindOfDrLiz, line 251: This still leaves the destructive race open. `queryEventsDetailed` can return `events: []`, `timedOut: false`, `noRelays: false` when one relay answered empty and the pool completed after the one-second settle window while another relay was still pending. Clearing `_globalBookmarks` here turns that partial answer into an authoritative empty list, so the next save can publish a one-item replacement over the real kind 10003 list. Please treat zero events as inconclusive for the write path until the query result can expose whether all relevant relays settled, or otherwise avoid clearing/persisting local state on this answer.

mobile/lib/services/bookmark_service.dart
```diff
@@ -256,12 +266,10 @@ class BookmarkService with NostrListServiceMixin {
       }
 
       _globalBookmarks.add(item);
```
NotThatKindOfDrLiz: The service mutates memory and prefs before the replaceable event is confirmed. If publish fails, callers get `false`, but the device now has a phantom bookmark until the next successful sync. Please either publish from a candidate list and commit local state only after success, or roll back memory/prefs on failure and add a regression test for that local-state consequence.

mobile/lib/services/bookmark_service.dart
```diff
@@ -621,228 +378,57 @@ class BookmarkService with NostrListServiceMixin {
...
-      );
-    }
-  }
+      if (event == null) return false;
 
-  /// Publish bookmark set to Nostr as NIP-51 kind 30003 event
-  Future<void> _publishBookmarkSetToNostr(BookmarkSet set) async {
-    try {
-      if (!_authService.isAuthenticated) {
+      final sentEvent = await _nostrService.publishEvent(event);
```
NotThatKindOfDrLiz: `publishEvent` only proves the SDK handed the frame off/broadcast it; it does not wait for relay `OK` acceptance. This PR now gates user-visible save success on this boolean, so a relay-policy rejection can still report success. Please use `publishEventAwaitOk` and gate on at least one `OK true`, or change the contract/UI wording so it does not claim the bookmark stuck on a relay.

NotThatKindOfDrLiz (changes requested):
Thanks for the cleanup here. I pushed one mechanical follow-up commit (`bb253e8e3`) removing the leftover dead wrappers/constant, reusing `BookmarkItem.fromTag`, deleting the dead bookmark-set l10n keys plus regenerated outputs, removing the passthrough save helper, and fixing the stale saved-videos doc comment.

I still think this needs one more author pass before merge. The core issue is that the write path can still publish a truncated kind 10003 list when the query settles with a fast empty relay while the relay holding the list is still outstanding. Separately, the save success path should not rely on `publishEvent` if the UI/business logic is treating the result as relay acceptance, and the local state should not be persisted before that publish is confirmed without rollback.

mobile/lib/services/bookmark_service.dart
realmeylisdev: Fixed in 3ba08a18c, with the pool-level capability it needed in 0e5d0d910.

You are right, and it reproduces one layer down from here: `relay_pool_query_settle_test.dart` already pinned that a mute relay plus one answering relay completes with `timedOut: false`. I mutated the new guard to confirm it — without it the query reports exactly the `events: [], timedOut: false` you described.

Rather than only treating zero events as inconclusive, I took the option you named as the end state and made the pool able to say so. `RelayPool.query` gained `requireAllRelaysSettled` (default off), which opts a single query out of the settle window so an incomplete answer arrives as `timedOut: true` instead of a complete-looking empty result. The write path now reconciles with `requireAuthoritative: true` and refuses to publish on anything short of that.

Two notes on the shape:

- Treating zero events as inconclusive on its own would have bricked the first-ever save — a brand-new user genuinely has zero events, so a blanket refusal never lets them create the list. Full settlement distinguishes "everyone answered, nobody has one" from "somebody did not answer", which keeps first-save working while closing the race.
- It also closes the case not raised here: adopting `events.first` from a partial answer could publish over a newer revision held only by the silent relay. Same gate covers it.

I deliberately left the display read on the fast path, so opening Saved keeps the settle window latency trade. Clearing local state there is no longer load-bearing: the causal chain you identified ran through the next save, and the write path no longer republishes from what a display read left in memory — it re-reads for itself. Happy to make the read authoritative too if you would rather trade the cold-open latency for it, but that felt like your call rather than mine.

Trade-off worth flagging: a relay that stays connected and never answers now makes bookmark writes fail rather than truncate. That is the safe direction, but it is a real availability cost, and the mute-relay shape has been measured against production. Tests: `relay_pool_query_settle_test.dart` (2 new) and `bookmark_service_test.dart` (`refuses when a relay never settled, even though another answered empty`, `reconciles with full settlement before publishing`).

mobile/lib/services/bookmark_service.dart
realmeylisdev: Fixed in 3ba08a18c — publish from a candidate list, commit local state only after acceptance.

`addToGlobalBookmarks` and `removeFromGlobalBookmarks` now build the new list as a candidate and hand it to `_publishGlobalBookmarks(candidate)`, which adopts it into memory and writes SharedPreferences only once a relay has returned `OK true`. On failure nothing local has moved, so the device still holds the list it last reconciled — no phantom waiting for a later sync to wipe it.

Regression tests for the local-state consequence specifically, since that was the part not previously covered: `leaves no phantom bookmark when the publish is refused` (asserts both in-memory state and that the prefs blob does not contain the item, so it cannot survive a restart either) and `keeps the item when the removal is refused` for the mirror case. I verified both fail if the local write is moved back before the publish.

mobile/lib/services/bookmark_service.dart
realmeylisdev: Fixed in 3ba08a18c — now `publishEventAwaitOk`, gated on `outcome.acceptedByAny`.

Confirmed your reading of the blast radius: `ShareSheetBloc._onSaveRequested` turns `result.succeeded` straight into the sheet message, so the old gate let a relay-policy rejection word itself as a successful save.

Kept the wording honest about what `OK true` does and does not mean — the doc on `_publishGlobalBookmarks` says acceptance is breadth, not durability, since the relay acknowledges from an in-memory queue and commits afterwards. Gated on `acceptedByAny` rather than `acceptedByAll` per the guidance on `publishEventAwaitOk` itself, so one wedged pool member cannot block every save.

Tests: `waits for a relay OK rather than a socket handoff` (verifies `publishEventAwaitOk` and that `publishEvent` is not used), plus failure coverage for both shapes that were previously reported as success — `reports failure when the relay rejects the publish` (`OK false`) and `reports failure when no relay answered the publish` (`noResponseFrom`).

realmeylisdev:
@NotThatKindOfDrLiz thanks — all three addressed, and the cleanup commit is much appreciated. Checks are green (17/17) at 3ba08a18c.

**Fixed** (details in the inline replies)

| Finding | Fix |
|---|---|
| Truncated write from a partial query answer | Write path reconciles with full relay settlement and refuses to publish on anything less — `3ba08a18c`, enabled by `0e5d0d910` |
| `publishEvent` treated as relay acceptance | `publishEventAwaitOk`, gated on at least one `OK true` — `3ba08a18c` |
| Local state persisted before confirmation | Publish from a candidate list, adopt only after acceptance — `3ba08a18c` |

Nothing escalated, nothing declined. Each fix is pinned by a test I verified fails without it, by mutating the discriminator rather than the outcome.

**Not requested by the review — please look at this separately.** Closing the first finding properly needed the pool to be able to report an incomplete answer, so `0e5d0d910` adds `requireAllRelaysSettled` to `RelayPool.query` / `Nostr.queryEventsDetailed` / `NostrClient.queryEventsDetailed`. It is additive and default-off, and the 14 pre-existing settle tests still pass unchanged, but it is a shared low-layer API and worth its own read. I went this way rather than just treating zero events as inconclusive because a blanket refusal would have made the first-ever save impossible — a brand-new user genuinely has zero events.

**Two open decisions I did not want to make unilaterally:**

1. A relay that stays connected and never answers now makes bookmark writes fail instead of truncating the list. Safe direction, but a real availability cost, and that mute-relay shape has been measured against production. If you would rather bound it differently, say so.
2. The display read still uses the fast path, so clearing local state on an empty answer is still possible there. It is no longer load-bearing — the chain you traced ran through the next save, and the write path now re-reads for itself instead of republishing from memory — but making the read authoritative too is a cold-open latency trade I would rather you called.

Also worth noting the same gate closes a case not raised: adopting `events.first` from a partial answer could publish over a newer revision held only by the silent relay.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
```diff
@@ -1674,6 +1690,11 @@ class RelayPool {
   /// query info will hold in relay and close in relay when EOSE message be received.
   /// if onlyTempRelays is true and tempRelays is not empty, it will only query throw tempRelays.
   /// if onlyTempRelays is false and tempRelays is not empty, it will query bath myRelays and tempRelays.
+  /// Set [requireAllRelaysSettled] when a partial answer is worse for the
```
dcadenas, line 1730: Follow-up, not blocking this PR.

**Full settlement excludes a configured relay that is offline, so a write can still replace from a partial view.**

The wait is over relays that accepted the REQ and are still connected. `relayDoQuery` calls `relay.saveQuery` only on a successful send, `_canStillSettleQuery` returns false for anything not `ClientConnected.connected`, and `NostrClient.queryEventsDetailed` reconnects only when *every* relay is down. So relay A connected and answering empty, plus relay B configured but offline holding the prior list, still reports `timedOut: false` — and the bookmark write path publishes A's view as the newer replacement.

That is the same class this PR closes, one step narrower. I am not asking for it here, for two reasons. You cannot wait for a terminal frame from a relay that never accepted the REQ, so widening this flag degrades to "refuse to save while any configured relay is offline", which fails worse than the bug. And the pool also holds cache, search and temp relays that are not bookmark stores, so "all configured relays" is not the right set anyway.

The shape that closes it is defining the user's NIP-65 write relays as the authoritative set and requiring settlement over *that* set. That is a design change well past this PR.

Worth saying plainly that this is not a regression you introduced: `requireAllRelaysSettled` has exactly one caller in the app, and every other replaceable-event write in the repo publishes with no settlement guarantee at all. This leaves bookmarks the most careful replaceable writer in the codebase.

One reviewer reading this held that it should block the PR rather than be filed as follow-up. Recording the dissent so the call is yours — I disagree, on the grounds above.

mobile/lib/services/bookmark_service.dart
```diff
@@ -63,188 +63,318 @@ class BookmarkItem {
...
-        return;
+        if (events == null) {
+          Log.warning(
+            'Empty bookmark answer not confirmed by every relay - keeping '
+            'the ${_globalBookmarks.length} bookmarks this device has',
+            name: 'BookmarkService',
+            category: LogCategory.system,
+          );
+          return false;
+        }
```
dcadenas, line 238: This is the branch `40b47fd7b` changed, so the reasoning behind it in one place.

The comment here was already right that an empty answer under `requireAuthoritative` is trustworthy and one without it may not be. The code then cleared and persisted either way, and `_saveBookmarksToSharedPreferences` runs below — so a partial empty answer destroyed the offline snapshot. The "nothing destructive can follow from it" note on the display-read test holds for relay data, since the write path reconciles separately; it does not cover the local cache, and on a fresh install it also left Saved showing "Nothing saved yet" for a user with bookmarks.

The invariant that fell out of it, and the reason the stamp is reset when a list is adopted: **content non-empty implies the absence stamp is null**. Without that, a kind-10003 whose items are all private has no public tags — empty list, ciphertext in `content` — and a later unconfirmed empty would have stripped another client's private bookmarks, which is the thing this PR set out to stop doing.

Each of the three guards has a test that fails when its guard is removed.

dcadenas (approved):
🔖 read-before-write on a replaceable list, with the empty-answer case actually handled — one follow-up inline, not blocking.

I pushed `40b47fd7b` to this branch. It stops a non-authoritative empty answer from being believed on its own; the PR body now describes it, including two user-visible behaviour changes worth knowing before the first support report.

Checks at `40b47fd7b`: format, analyze, all four test shards, Generated Files, VGV Tag Gate, both package builds — green.

One thing I could not verify locally: `flutter test integration_test/e2e/ --tags service` does not build on Linux, and the `nostr_client` / `nostr_sdk` roots declare only CI-side reusable workflows, so I covered those two with their own `flutter analyze` and `flutter test` rather than the declared command.

dcadenas:
@NotThatKindOfDrLiz your three items look resolved at `adc9805f0`, so this is blocked only on your review being re-run rather than on outstanding work. Checked each against the current head:

- **Truncated publish on a fast empty relay.** The write path now reconciles with `requireAuthoritative: true`, which sets `requireAllRelaysSettled` on the query, and refuses to publish when that comes back inconclusive. `relay_pool.dart` skips the one-second settle window for those queries, so a connected relay that never sends a terminal frame runs the caller to its timeout instead of being counted as having answered.
- **Relying on `publishEvent` for relay acceptance.** It is `publishEventAwaitOk` now, gated on `outcome.acceptedByAny`.
- **Persisting local state before the publish is confirmed.** `_globalBookmarks` and the SharedPreferences write both happen after the `acceptedByAny` check, so a refused publish leaves nothing behind. Pinned by *leaves no phantom bookmark when the publish is refused*.

One narrower descendant of your first item is still open, and I left it as a non-blocking follow-up rather than asking for it here: full settlement waits only on relays that accepted the REQ and are still connected, so a configured relay that is *offline* is excluded rather than making the answer inconclusive. Detail and reasoning are in the thread on `relay_pool.dart`.

I also pushed `adc9805f0` (rebased by the author from `40b47fd7b`), which stops a non-authoritative empty answer from being believed on its own — the read path had the same fast-empty problem you named on the write path, and on a fresh install it reproduced #6627 itself. All checks are green at the current head.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
realmeylisdev: Agreed, taking it as follow-up. Recorded on #6238 rather than as a new issue — that is the same class one level up ("`queryEvents` silently returns partial results"), and #2513 is the dependency for the closing shape you describe: https://github.com/divinevideo/divine-mobile/issues/6238#issuecomment-5248293787

Since the call was mine, I checked the mechanism against head rather than taking it on the description. It holds: `relayDoQuery` saves the query only inside `if (result)`, so a failed send leaves nothing to settle (`:552`), and `_canStillSettleQuery` bails on anything not `ClientConnected.connected` (`:927`). Relay A connected and empty plus relay B offline with the list does report `timedOut: false`.

On the recorded dissent, I am siding with follow-up rather than blocking, and the deciding ground is that blocking preserves the worse state rather than a better one. `requireAllRelaysSettled` has one app-side caller; every other replaceable-event write in this repo publishes with no settlement guarantee at all, and #6627 is live on `main` today.

Your second argument is the one I found hardest to get around, and it is what settles it: there is no correct set to widen to yet. The settlement scan spans `_relaysSnapshot() + _tempRelaysSnapshot() + _cacheRelaysSnapshot()` (`:787-790`), and cache and temp relays are not stores for the user's replaceable lists. Widening the flag before #2513 gives us the NIP-65 write set would trade a rare truncation for "cannot save while any configured relay is offline" — worse on availability, and still not correct on the set.

Thanks for writing the dissent down instead of dropping it. Anyone who wants to reopen the call has the argument in front of them.

realmeylisdev:
@NotThatKindOfDrLiz re-requesting your review. Your three items are addressed, and the branch was rebased after you looked — which broke the SHAs my thread replies cite. Remap below so your threads still navigate.

**Your three items, at `adc9805f0`**

1. **Truncated publish on a fast empty relay.** The write path reconciles with `requireAuthoritative: true`, which sets `requireAllRelaysSettled` on the query and refuses to publish on anything short of full settlement. I took the second option you named — teach the query result to expose whether all relevant relays settled — rather than treating zero events as inconclusive outright, since a blanket refusal never lets a brand-new user create their first list.
2. **Relying on `publishEvent` for relay acceptance.** Now `publishEventAwaitOk`, gated on `outcome.acceptedByAny`.
3. **Persisting local state before the publish is confirmed.** Publish from a candidate list; memory and SharedPreferences commit only after the acceptance check, so a refused publish leaves nothing behind.

**Rebase remap** — `3ba08a18c` and `0e5d0d910` in my replies no longer exist:

| my replies cite | now |
| --- | --- |
| `b90924882` | `c6c48f702` |
| `60e7ccae6` | `f0f64a391` |
| `ddffc8e32` | `bdf326928` |
| `4538b72d5` | `0b9da7210` |
| `bb253e8e3` | `6db5b564e` |
| `0e5d0d910` | `5e7c2fda1` |
| `3ba08a18c` | `f8b726fc0` |

Six are byte-identical to what you reviewed. `f0f64a391` differs only where the rebase absorbed a `createBloc()` signature change from `main` in a bloc test. `adc9805f0` is new — @dcadenas's fix for the same fast-empty problem on the **read** path, which on a fresh install reproduced #6627 itself.

**One thing I decided against carrying here, flagged rather than buried.** @dcadenas found a narrower descendant of your first item: full settlement waits only over relays that accepted the REQ and are still connected, so a configured relay that is *offline* is excluded rather than making the answer inconclusive. One reviewer held that it should block this PR. I have filed it as follow-up on #6238 instead, on the grounds that there is no correct relay set to widen to before #2513 — the settlement scan also spans cache and temp relays, which are not bookmark stores — and that blocking preserves a worse state, since every other replaceable-event write in the repo has no settlement guarantee at all. Reasoning in the `relay_pool.dart` thread and in https://github.com/divinevideo/divine-mobile/issues/6238#issuecomment-5248293787. If you read it the other way, say so and I will reopen the call rather than merge past it.

All checks green at `adc9805f0`.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
```diff
@@ -1697,6 +1719,9 @@ class RelayPool {
       _queryCompleteCallbacks[subscription.id] = onComplete;
       _queryFanoutInProgress.add(subscription.id);
       _querySentAt[subscription.id] = DateTime.now();
+      if (requireAllRelaysSettled) {
+        _queriesRequiringFullSettlement.add(subscription.id);
+      }
```
NotThatKindOfDrLiz, line 1762: This can still complete a `requireAllRelaysSettled` query as a normal empty answer even when no relay actually participated. If every `relayDoQuery` send returns false, no relay saves the query, the post-fanout sweep finds no pending relay, and `_completeQuery` fires without `_queryAnswered` ever being set. `NostrClient` then reports `events: []`, `timedOut: false`, `noRelays: false` whenever `connectedRelays` is nonempty, and the bookmark write path can treat that as an authoritative empty kind 10003. Please surface `at least one relay produced a terminal frame` or otherwise make this inconclusive for full-settlement callers, with a regression test.

mobile/packages/nostr_client/lib/src/nostr_client.dart
```diff
@@ -888,6 +900,7 @@ class NostrClient {
           relayTypes: relayTypes,
           sendAfterAuth: sendAfterAuth,
           timeout: timeout,
+          requireAllRelaysSettled: requireAllRelaysSettled,
         );
```
NotThatKindOfDrLiz, line 907: This forwards the full-settlement flag to the WebSocket query, but the closed query-pool branch below still synthesizes a successful empty answer before that call can run. For destructive callers using `requireAllRelaysSettled`, `_queryPool.isClosed` should not become `events: []`, `timedOut: false`; `_readRemoteGlobalBookmarks` can treat that as `no bookmarks` rather than inconclusive. Please return an inconclusive signal for authoritative callers, or include this state in the detailed result contract and have bookmark sync refuse to publish from it.

NotThatKindOfDrLiz (changes requested):
Requesting changes for one remaining data-loss edge case.

The original three review blockers are fixed at current head, but an authoritative bookmark read can still be reported as a clean empty answer when the query did not actually get a terminal relay answer. Because kind 10003 is replaceable, the write path must treat that as inconclusive before publishing from the result.

Please surface query participation/terminal-answer state through the detailed query result, or otherwise make these branches return an inconclusive result for `requireAllRelaysSettled` callers, with regression tests.

mobile/packages/nostr_sdk/lib/relay/relay_pool.dart
realmeylisdev: Fixed in dc3165597.

Confirmed the mechanism at head rather than taking it from the description. `relayDoQuery` only calls `relay.saveQuery` inside `if (result)`, so when every send fails no relay holds the query, the post-fan-out sweep finds no straggler, and `_completeQuery` fires with `_queryAnswered` never set. Driving the harness's `sendSucceeds = false` reproduces it exactly: `events: []`, `timedOut: false`.

The guard is the rule you named, stated once: a full-settlement query may only complete once at least one relay has produced a terminal frame. Anything short of that holds to the caller's own timeout, which reaches the bookmark read as `timedOut: true` and is already refused there.

It also closes a case not raised here — every relay accepts the REQ and then drops its socket before answering. That completes through `_settleQueriesBlockedBy` with nobody having answered, which is the same empty box by a different route.

Holding the query leaks nothing: the caller's timeout routes through `unsubscribe`, which drops the callback and every settlement marker, and `_releaseQuery` closes the REQ on any relay still carrying it.

Tests: `reports a query no relay accepted as a timeout, not an empty answer`, plus `a query no relay accepted still completes at once without the flag` — the second one fails if the guard is applied unconditionally, so the scope is pinned rather than assumed.

mobile/packages/nostr_client/lib/src/nostr_client.dart
realmeylisdev: Fixed in 633cb6894.

The closed-pool branch now reports `timedOut: requireAllRelaysSettled` — inconclusive for a caller about to replace what it read, unchanged for display reads that are content to fall back on cache. Reproduced with the same dispose race the existing pooled-path test uses: park at the cache await, dispose so the pool closes, resume into the guard. Before the fix that read came back `timedOut: false, noRelays: false`, which `_readRemoteGlobalBookmarks` accepts as authoritative.

One sibling deliberately left alone: the `_isDisposed` early return above already reports `noRelays: true`, and the bookmark read treats that as inconclusive too, so that path was never exposed. Moving it onto the `timedOut` channel would change what every other caller sees on a disposed client for no gain here.

Test: `reports a closed query pool as inconclusive for a full-settlement caller`, which also asserts `noRelays` is false so `timedOut` is the thing carrying it.

realmeylisdev:
@NotThatKindOfDrLiz both items from the review are fixed at `633cb6894`. Nothing escalated, nothing declined.

Both threads were the same defect in two layers, so the fix states one invariant: **a full-settlement query needs a terminal frame from at least one relay.** An answer nobody gave is not an empty answer, and neither branch can report one as complete any more.

- `dc3165597` — `RelayPool._fireQueryCompleteIfSettled` no longer completes a `requireAllRelaysSettled` query that no relay ever answered. When every `relayDoQuery` send fails nothing is saved, so the post-fan-out sweep used to complete it with an empty box no relay contributed to. It now holds to the caller's own timeout, which arrives as `timedOut: true`.
- `633cb6894` — the `_queryPool.isClosed` dispose-race branch in `NostrClient.queryEventsDetailed` reports `timedOut: requireAllRelaysSettled` instead of synthesizing a clean empty answer.

On the choice you offered: I took "make these branches inconclusive" rather than adding participation state to the detailed result. `timedOut` already means *inconclusive* for these callers, the bookmark write path already refuses on it, and existing tests already pin that refusal — so this needed no new field and no new API surface. If you would rather the result record carried participation explicitly, that is a bigger change and I would rather do it deliberately than fold it in here.

Both guards were reproduced red before the fix: `sendSucceeds = false` on the pool harness, and the existing cache-await dispose race on the client. Regression tests are `reports a query no relay accepted as a timeout, not an empty answer`, `a query no relay accepted still completes at once without the flag` (fails if the guard is applied unconditionally, so the scope is pinned), and `reports a closed query pool as inconclusive for a full-settlement caller`.

Verified locally before pushing: every `relay_pool_*` suite, the full `nostr_client` suite at 87.8% coverage against its 82 gate, `bookmark_service` / `share_sheet_bloc` / `profile_saved_videos_bloc`, and `flutter analyze lib test` clean on both packages. CI is green at `633cb6894` — all four test shards, both package builds, Generated Files, Analyze and Format.

Blast radius is one read path: `requireAllRelaysSettled` still has exactly one app-side caller, `bookmark_service.dart`. No behaviour changed for any other query — the contrast test above exists to keep it that way.

@dcadenas's offline-relay thread stays a follow-up per the agreement there; it is recorded on #6238 and not re-opened by this push.

realmeylisdev:
Ran this branch end-to-end on an iPhone and on the simulator against the production relay, watching the debug console throughout. The headline is that the fix works: on a **fresh install with an empty local cache**, opening Bookmarks logged `Synced 2 global bookmarks from relay` → `Fetched 2/2 videos` and rendered both thumbnails. That is #6627's expected result, observed rather than inferred.

Three findings came out of it. One is fixed on the branch; two are behaviour calls I did not want to make unilaterally.

### Fixed: the legacy prose in `.content` was being republished forever (`698dd6626`)

The description says `.content` "is now empty, or the ciphertext read back". For every account that ever saved a bookmark on an older build, neither was true. `_adoptGlobalBookmarksFromEvent` cannot tell another client's NIP-44 ciphertext from Divine's own prior prose, so it carried `Divine global bookmarks` straight back into the next publish — the exact malformed value this branch set out to stop writing, preserved for precisely the users who already had it.

Verified three ways: a unit repro built from a live kind 10003 pulled off `relay.divine.video` (published `content` came back as `Divine global bookmarks`), then on device, then on a fresh identity on the simulator — both now publish `content: ''`.

The fix drops that one literal on adopt. Anything else still passes through untouched, which the existing ciphertext carry-through test still proves.

### Behaviour call 1: a row labelled "Save" now deletes bookmarks

`share_sheet_more_actions.dart:70` labels the row `shareSheetSave` → "Save", unconditionally — no saved-state label, no filled icon, no indicator anywhere in the sheet.

Before this branch the toggle direction came from the *local* list, so on a fresh install "Save" always added. Now `toggleVideoInGlobalBookmarks` reconciles against the relay first, so on a device whose cache is empty but whose relay list holds the video, "Save" removes it. Measured on device, same video, back to back:

```
TOGGLE#1 -> succeeded=true was=false is=true   (added)
TOGGLE#2 -> succeeded=true was=true  is=false  (removed)
```

The reconcile is correct — the UI never caught up with it. Worth noting the ordinary "tap Save again to be sure" gesture now destroys a bookmark. The remedy is a product decision (state-dependent label, or a separate Remove action), so I have left it alone.

### Behaviour call 2: Save became a network round trip with no pending affordance

Save now blocks on an authoritative query plus a publish awaiting `OK`. Measured 729 ms and 1239 ms on device; the query timeout bounds the worst case at 5 s. `ShareSheetState` carries `isSending` for DMs but has nothing for the save path, and `transformer: droppable()` silently swallows repeat taps — so the sheet sits inert and a second tap does nothing visible. Also a design call: spinner, disabled row, or optimistic state.

### One hypothesis I raised and then disproved

Mid-patrol I saw an event ACKed `OK true` that was absent from `relay.divine.video` by id at 10:34 and 10:40 and present at 10:47, and I suspected that read-after-write gap could defeat the reconcile and destroy an earlier save. **It did not reproduce.** A controlled run — fresh install, throwaway identity, two saves three seconds apart — ended with the relay holding both `e` tags and `content: ''`. The reconcile protected the earlier save. Recording the lag because it is real and worth knowing, not as a defect claim.

Unrelated, found in passing and not touched here: the router treats a `divine://` deep link to an ordinary app route as a signer callback and redirects it to `/welcome` (`Router redirect: signer callback divine:///saved-videos -> /welcome`).

NotThatKindOfDrLiz:
Follow-up issues are now filed and assigned to @realmeylisdev:

- #7072 — share-sheet `Save` affordance can remove an already-saved relay bookmark because the label/icon do not reflect the reconciled state.
- #7073 — bookmark save/remove is now a blocking relay reconcile + publish-OK path with no save-specific pending affordance, including the offline/relay-less timeout case.
- #7074 — `divine:///saved-videos` currently routes through the custom-scheme signer-callback path instead of Saved Videos.

Product decision requested for #6966 before clearing the review state:

1. Is it acceptable to merge #6966 with the current share-sheet `Save` toggle affordance and track the label/state fix in #7072, or should that be fixed before merge?
2. Is it acceptable to merge #6966 with the current blocking save/remove wait and track the pending/offline affordance in #7073, or should that be fixed before merge?

My read is that the data-loss fix in #6966 is still the safer direction to ship, but those two UX changes should be explicit decisions rather than silent behavior changes.

NotThatKindOfDrLiz (approved):
Took over the remaining full-settlement CLOSED blocker and pushed `fab54d3ed3`.

What changed:
- Rebasing the PR onto current `origin/main` removed the merge-commit/BEHIND state.
- `RelayPool` now keeps the existing terminal-frame signal for normal display reads, but separately records non-auth `CLOSED` as inconclusive for `requireAllRelaysSettled` callers.
- Added a regression test for a single relay returning `CLOSED` with `error: too many concurrent REQs`; full-settlement queries now report `timedOut: true` instead of an authoritative empty answer.

Verified locally:
- `flutter test test/unit/relay_pool_query_settle_test.dart` from `mobile/packages/nostr_sdk`
- `flutter analyze` from `mobile/packages/nostr_sdk`
- `flutter test` from `mobile/packages/nostr_sdk`
- Pre-push hook: analyzer plus changed-file tests all passed

GitHub checks are green on `fab54d3ed3`, including Mobile CI shards, generated files, Nostr Client CI, Nostr SDK CI, preview build, services integration, mergeability, semantic PR, and CLA.

NotThatKindOfDrLiz:
Final takeover handback update for current head `4435fd6e26`:

- Rebasing onto latest `origin/main` completed cleanly; GitHub now reports `mergeStateStatus: CLEAN`.
- All GitHub checks are green on `4435fd6e26`.
- My prior approval remains active; the PR review decision is `APPROVED`.

@realmeylisdev please review the takeover commit and merge when ready.

realmeylisdev:
@NotThatKindOfDrLiz — answering the two product questions from your comment above; they were still open when you cleared the review state.

1. **Share-sheet `Save` affordance (#7072): ship as-is, fix in the follow-up.**
2. **Blocking save/remove wait (#7073): ship as-is, fix in the follow-up.**

Both are behaviour changes this PR introduces rather than pre-existing gaps, which is exactly why they deserved an explicit call rather than landing quietly — agreed with your read that the data-loss fix is the safer thing to ship first. Both issues are open and assigned.

The head has moved since your approval and is now `ef02a8520`. Nothing in the branch itself changed: the twelve commits are the same patches you approved, and the only additions are `main` merging forward to pick up #6855. All 18 checks are green and GitHub reports the PR mergeable.

PR #6970, opened 2026-08-09
fix(collabs): render only confirmed collaborators to third-party viewers

### LINKED ISSUE #6907: fix(collabs): render only confirmed collaborators to third-party viewers
## Problem

Third-party viewers can still see pending collaborators as collaborators on video metadata. `CollaboratorConfirmationRepository` only opens the kind-34238 acceptance subscription for own-authored videos and otherwise falls back to local current-user state. For unrelated viewers, tagged collaborator pubkeys remain `pending`, which means pending invites can still render publicly instead of being hidden until confirmed.

This is follow-up scope split out from #4201 after the own-profile collabs surface and Funnelcake materialization fixes shipped.

## Current evidence

- `mobile/packages/collaborator_repository/lib/src/collaborator_confirmation_repository.dart` documents that third-party rendering of pending collaborators is unchanged and depends on Funnelcake.
- The repository subscribes to kind-34238 only when `creatorPubkey == currentUserPubkey`.
- Mobile already has a confirmed profile collabs feed via `GET /api/users/{pubkey}/collabs`, but it does not appear to have a per-video or batch confirmed-collaborator status read that can drive metadata rendering.

## Expected behavior

- Creators may see pending collaborators on their own videos.
- The invited collaborator may see their own local accept or ignore state.
- Third-party viewers should only see collaborators confirmed by the canonical Funnelcake read model.
- Pending or ignored collaborators must not leak into public video metadata rendering.

## Implementation notes

Add or expose a confirmed-collaborator read suitable for video metadata, then route it through the existing layers:

- Funnelcake API client: per-video or batch confirmed-collaborator status read, if the backend route exists.
- Repository layer: move third-party status derivation away from relay/local-only inference.
- UI boundary: keep `VideoCollaboratorStatusCubit` semantics stable if possible.

Confirm with Funnelcake owners whether a per-video confirmed-collaborators endpoint exists already. If not, this mobile issue is blocked on that backend route.

## Acceptance criteria

- Third-party viewers do not see pending collaborator p-tags as collaborators.
- Confirmed collaborators still render for third-party viewers.
- Creator and invited-current-user local behaviors remain intact.
- Regression tests cover the third-party viewer path in `collaborator_confirmation_repository_test.dart` or the new owning repository tests.

Refs #4201.

realmeylisdev:
## Summary

Third-party viewers saw every pubkey a creator tagged rendered exactly like an accepted collaborator. Tagging someone publicly credited them whether or not they ever accepted — and even after they explicitly ignored the invite, since "ignore" only ever hid the avatar on the ignoring user's own device.

Closes #6907.

### Correction to the issue's premise

#6907 says this may be blocked on a Funnelcake per-video confirmed-collaborator endpoint. It isn't. The relay serves kind-34238 to unauthenticated readers, verified directly:

```
nak req -k 34238 -t a=34236:fc7031a810ce4b02b6195a7e477cfe3d08c0386038bd45b4431f82d9b3f5ffb0:50cca13b441b84fbd6b7ae25eb906613b4758e145e33978f94598dd0fcf528f4 wss://relay.divine.video
→ 2 acceptance events, no auth, no signer
```

divine-web already renders third-party-correct status this way (`useVideoCollaboratorStatus.ts`), ungated by viewer identity. Mobile's `isOwnVideo` gate was the outlier. No backend work required.

## What changed

Five commits, each independently reviewable:

1. **Accept kind-34238 events with no `status` tag.** Mobile required `status=accepted`; divine-web publishes `[['a', coord], ['d', uuid]]` with no such tag, and Funnelcake's confirmed MV doesn't check for one either. Mobile alone dropped web-made acceptances — **18 of 120 (15%)** of the kind-34238 events currently on the relay. The visible effect: accepting on web left you confirmed on your profile Collabs tab but permanently greyed as "pending" on the creator's mobile view.
2. **Add `VideoCollaboratorStatus.isResolved`,** set from the relay EOSE callback `NostrClient.subscribe` already exposes. Without it "pending" conflates "did not accept" with "have not heard back yet".
3. **Open the subscription for every viewer,** bounded by `authors: taggedPubkeys`. Also inverts precedence in `_snapshot` so the current user's local decision outranks the relay echo — see Unrequested below.
4. **Hide unconfirmed collaborators from third parties.** `visiblePubkeys` gains three branches: fallback (raw list), author (all tagged, unconfirmed greyed as before), everyone else (confirmed only, and only once resolved).
5. **Close and key relay subscriptions by their full filter.** `CollaboratorConfirmationRepository` now owns an explicit Nostr subscription id per `(videoAddress, taggedPubkeys)` filter, calls `NostrClient.unsubscribe` from `release()` and `close()`, ignores stale EOSE/event callbacks after teardown or filter changes, and reopens the relay query when the tagged collaborator set changes.

No UI changes were needed — `CollaboratorAvatarRowBody` already returns `SizedBox.shrink()` on an empty list, so "hide until resolved" falls out of the visibility change.

## Reproduction on production data

Sampled 400 recent kind-34236 events: 10 carry `collaborator`-role `p` tags (2.5%). Checking each tagged pubkey for an acceptance on that exact coordinate, **2 of 9 resolvable videos have a tagged collaborator who never published one** — credited to every viewer today, hidden after this change:

```
34236:f67d985c0bfbf87eaa33b056f1d38ad991a6aa625b138bddb2a838bdbac29f40:d5f1c837b086543c440b96e554c887337b47c397e7984e1f67207d827afcf382
  tagged   c4a39f1291291d452405cd8ddd798c4a29a3858c52cd0d843f1f6852cf17682e   accepted: none

34236:d73d071c85e1a9250860509591df95b6b20f610b67aa37f252cd07f11e679b14:0d3363d0481c55196147fc00dc455b444a8650205aa6344c1b1210dc383d2a7d
  tagged   28128f8bd4cfe9c89293cad8bdf1a8431ace09a85d724421fb4ccbb5e60d02a8   accepted: none
```

That 2.5% rate also bounds the relay-load change below.

## Unrequested change, flagged

Commit 3 also inverts status precedence in `_snapshot`. Relay-derived acceptance was checked *before* the current user's local override. That was safe only while the subscription never opened for a recipient — `_relayAccepted` was always empty on that path. Opening it for every viewer would let a stale kind-34238 from an earlier accept override a later local `ignore`, resurrecting an avatar the user had hidden. Caught by a regression assertion added to the recipient-view test, not by the existing suite.

## Behaviour choice worth a reviewer's eye

This is **fail-closed**: if the relay is unreachable, EOSE never arrives and third parties see no collaborators rather than falling back to the raw tagged list. That is the right trade for a credit leak — showing unconfirmed names is the bug — but a relay outage now under-credits confirmed collaborators instead of over-crediting unconfirmed ones.

## Testing

- `collaborator_repository`: 45 tests (was 33), coverage 97.10% against its 96 gate
- `models`: 8 collaborator-status tests, full package suite 877 green
- app: 367 green across `test/widgets/video_feed_item/`, `test/blocs/video_collaborator_status/`, and the invite-actions cubit
- `flutter analyze lib test` clean for both the app and the package

Six pre-existing widget/visibility tests encoded the old contract (third parties see the raw list). They are updated to the new contract rather than deleted, with the superseded ones consolidated instead of restated.

Notable new coverage: a divine-web-shaped acceptance confirms; an explicit non-accepted status still rejects; unresolved-before-EOSE and resolved-after; a second watcher inherits the resolved snapshot; release -> re-watch resolves again; late EOSE after release does not poison the next watch; changed tagged-pubkey sets reopen the relay subscription.

## Manual verification

Built and launched on a physical iPhone (iOS 26.6) and on the iOS 26.5 simulator — both build, install, and run with the change.

**The visual before/after was not captured.** The simulator's stored account wedges on the account-selection screen (`hasExistingProfile=false`) and never reaches a video, and wireless debugging on the physical device can't attach the Dart VM Service without a local-network permission tap on the phone. Neither is related to this diff. The third-party render path is covered by the widget tests above; a reviewer with a working signed-in build can confirm visually against either coordinate listed in the reproduction section.

## Follow-ups this surfaced, not fixed here

- The mobile/web/Funnelcake disagreement on what counts as an acceptance deserves its own issue. Commit 1 fixes it only from mobile's side; divine-web should stamp `status` for forward clarity, and the contract should be written down — it currently exists only as divine-mobile's own test assertions from #4045.
- divine-web mints a random-UUID `d` tag, so each web accept creates a new event rather than replacing the previous one. That defeats the addressable-replaceable semantics kind 34238 was chosen for, and means web cannot express an un-accept by replacement. Same class as divinevideo/divine-connect#35.
- "Ignore" remains device-local with no protocol-level decline. After this change third parties stop seeing an ignored collaborator (they never confirmed), but the creator still does.

### DIFF
--- a/mobile/lib/blocs/video_collaborator_status/video_collaborator_status_cubit.dart
+++ b/mobile/lib/blocs/video_collaborator_status/video_collaborator_status_cubit.dart
@@ -49,6 +49,7 @@ class VideoCollaboratorStatusCubit extends Cubit<VideoCollaboratorStatusState> {
               state.copyWith(
                 load: VideoCollaboratorStatusLoad.ready,
                 statusByPubkey: snapshot.statusByPubkey,
+                isResolved: snapshot.isResolved,
               ),
             );
           },
--- a/mobile/lib/blocs/video_collaborator_status/video_collaborator_status_state.dart
+++ b/mobile/lib/blocs/video_collaborator_status/video_collaborator_status_state.dart
@@ -19,10 +19,9 @@ class VideoCollaboratorStatusState extends Equatable {
   ///
   /// Distinct from [load]: `ready` means a snapshot arrived, which happens
   /// immediately and before any relay answer. Surfaces that hide unconfirmed
-  /// collaborators must gate on this instead, and it stays `false` on
-  /// [VideoCollaboratorStatusLoad.failure] so an unreachable relay keeps them
-  /// hidden rather than falling back to crediting untagged-but-unconfirmed
-  /// pubkeys.
+  /// collaborators must gate on this instead. A failure before EOSE leaves it
+  /// false, so an unreachable relay keeps third-party surfaces hidden rather
+  /// than falling back to crediting unconfirmed pubkeys.
   final bool isResolved;
 
   CollaboratorStatus statusFor(String pubkey) =>
--- a/mobile/lib/blocs/video_collaborator_status/video_collaborator_status_state.dart
+++ b/mobile/lib/blocs/video_collaborator_status/video_collaborator_status_state.dart
@@ -9,10 +9,21 @@ class VideoCollaboratorStatusState extends Equatable {
   const VideoCollaboratorStatusState({
     this.load = VideoCollaboratorStatusLoad.loading,
     this.statusByPubkey = const {},
+    this.isResolved = false,
   });
 
   final VideoCollaboratorStatusLoad load;
   final Map<String, CollaboratorStatus> statusByPubkey;
+
+  /// Whether the acceptance query has finished (relay EOSE).
+  ///
+  /// Distinct from [load]: `ready` means a snapshot arrived, which happens
+  /// immediately and before any relay answer. Surfaces that hide unconfirmed
+  /// collaborators must gate on this instead, and it stays `false` on
+  /// [VideoCollaboratorStatusLoad.failure] so an unreachable relay keeps them
+  /// hidden rather than falling back to crediting untagged-but-unconfirmed
+  /// pubkeys.
+  final bool isResolved;
 
   CollaboratorStatus statusFor(String pubkey) =>
       statusByPubkey[pubkey] ?? CollaboratorStatus.pending;
@@ -20,13 +31,15 @@ class VideoCollaboratorStatusState extends Equatable {
   VideoCollaboratorStatusState copyWith({
     VideoCollaboratorStatusLoad? load,
     Map<String, CollaboratorStatus>? statusByPubkey,
+    bool? isResolved,
   }) {
     return VideoCollaboratorStatusState(
       load: load ?? this.load,
       statusByPubkey: statusByPubkey ?? this.statusByPubkey,
+      isResolved: isResolved ?? this.isResolved,
     );
   }
 
   @override
-  List<Object?> get props => [load, statusByPubkey];
+  List<Object?> get props => [load, statusByPubkey, isResolved];
 }
--- a/mobile/lib/widgets/video_feed_item/collaborator_avatar_row.dart
+++ b/mobile/lib/widgets/video_feed_item/collaborator_avatar_row.dart
@@ -98,12 +98,16 @@ class _StatusAwareRow extends StatelessWidget {
     final statusByPubkey = context.select(
       (VideoCollaboratorStatusCubit c) => c.state.statusByPubkey,
     );
+    final isResolved = context.select(
+      (VideoCollaboratorStatusCubit c) => c.state.isResolved,
+    );
     return CollaboratorAvatarRowBody(
       visibility: CollaboratorVisibility(
         taggedPubkeys: pubkeys,
         statusByPubkey: statusByPubkey,
         currentUserPubkey: currentUserPubkey,
         creatorPubkey: video.pubkey,
+        isResolved: isResolved,
       ),
     );
   }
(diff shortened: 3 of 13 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I pushed a takeover commit on top of your PR: `d491100f9 fix(collabs): close confirmation relay subscriptions`.

What changed:
- repository now owns an explicit Nostr subscription id per `(videoAddress, taggedPubkeys)` filter and calls `NostrClient.unsubscribe` on `release()`/`close()`;
- stale EOSE/event callbacks are ignored after teardown or filter changes;
- collaborator pubkeys are normalized before building the authors filter;
- added regressions for release -> rewatch, late EOSE after release, and changed tagged-pubkey sets;
- cleaned stale docs/comments that still described the pre-#6907 behavior.

Verification:
- `flutter test` in `mobile/packages/collaborator_repository`
- `flutter analyze` in `mobile/packages/collaborator_repository`
- targeted app analyze for the touched bloc/widget files
- pre-push analyzer and changed-file tests passed on the rebased commit
- GitHub CI is green on `d491100f9`

NotThatKindOfDrLiz (approved):
Approved after takeover fixes. The subscription lifecycle blockers from review are addressed, the branch is rebased onto current `origin/main`, and local/pre-push/GitHub checks are green on `d491100f9`. @realmeylisdev please review the pushed takeover commit before merging.

PR #6976, opened 2026-08-09
fix(db-client): move schema repair into drift migration

### LINKED ISSUE #6921: refactor(db-client): move startup schema repairs into Drift migrations
## Problem

`mobile/packages/db_client/lib/src/database/app_database.dart` still reports `schemaVersion => 1`, while schema evolution is handled in `beforeOpen` through startup repair SQL.

As of #4335 triage, that path contains 18 `CREATE TABLE` blocks and 35 `_addColumnIfMissing` calls. The pattern keeps growing because new schema changes can be added without a real Drift `onUpgrade` migration.

## Why this matters

Startup repair is useful for recovering historical installs, but it should not be the primary schema migration strategy. The current shape makes install-state behavior harder to reason about and increases the risk of untested migration drift.

## Acceptance Criteria

- `db_client` has a real Drift `MigrationStrategy.onUpgrade` path for schema changes going forward.
- `schemaVersion` is incremented appropriately.
- Existing startup repair SQL is either moved into versioned migrations or explicitly retained only as recovery/backfill code with tests explaining why.
- Migration tests cover representative old-install states, including missing tables and missing columns currently handled by `beforeOpen`.
- New schema-change guidance is reflected in the data foundation policy work.

Part of #4335.

NotThatKindOfDrLiz:
## Summary
- bump db_client AppDatabase schemaVersion to 2 and run the legacy v1 normalization from onCreate/onUpgrade
- keep beforeOpen schema repair guarded as recovery only, with startup cleanup still running every open
- add Drift v2 schema artifacts and migration tests for v1 -> v2 validation
- update db_client migration docs so future schema changes use versioned Drift migrations
- document that conversations.last_read_timestamp backfill is now legacy-only; current self-sent read-state recovery is covered by the last-sent floor

Closes #6921

## Verification
- cd mobile/packages/db_client && dart run build_runner build --delete-conflicting-outputs
- cd mobile/packages/db_client && dart run drift_dev make-migrations
- cd mobile/packages/db_client && flutter test test/drift/app_database/migration_test.dart test/src/database/app_database_test.dart
- cd mobile/packages/db_client && flutter analyze
- cd mobile/packages/db_client && flutter test
- cd mobile && flutter analyze
- pre-push hook: analyzer, generated files, changed tests


hm21 (approved):
LGTM — this is the right shape: a versioned migration for the real work, `beforeOpen` demoted to a guarded recovery probe, and a doc rule so the next schema change doesn't land back in startup SQL.

I verified the field case by hand before approving. Degraded a fresh database to a genuine v1 shape — dropped the 17 tables the normalization creates, dropped `event.d_tag` and `personal_reactions.addressable_id` with their indexes, rebuilt `notifications` with the old single-column primary key, set `user_version` back to 1 — then reopened. `onUpgrade` restores every table, backfills `d_tag` from the stored tags JSON, and rebuilds `notifications` with the composite key without losing rows.

**Pushed one commit** (12d3cf0e6b, test-only): that scenario as a regression test in `app_database_test.dart`. It isn't reachable from the generated migration test — `drift_schema_v1.json` and `drift_schema_v2.json` are dumps of the same declared table set, so `schemaAt(1)` already hands back a complete v2-shaped database and `legacy v1 schema migrates to v2` never sees a legacy shape. Not something the drift tooling could have given you; the historical v1 shape was never captured. Confirmed the test fails when the `onUpgrade` body is emptied. `DB Client CI` green on the new head.

Two notes, no action needed unless you disagree:

- `_needsSchemaRepair()` probes 9 of the ~40 indexes the normalization creates, and doesn't probe the `notifications` key shape. Fine for something documented as a recovery probe — flagging it only so the asymmetry is a choice rather than an oversight.
- The `conversations.last_read_timestamp` backfill stops running per launch. `upsertConversation` sets `is_read = 1` for self-sent messages without writing a cursor, so those rows now stay NULL and `_publishReadMarker` skips them. The reinstall last-sent floor covers the same ground, so this looks like the backfill finally behaving as its comment describes — worth a line in the description either way.

NotThatKindOfDrLiz:
Rebased onto current main and refreshed the v2 baseline.

Changes since the approved head:
- regenerated Drift v2 artifacts so the baseline includes seen_videos
- added seen_videos / idx_seen_videos_last_seen_at to the guarded recovery probe
- added a test binding the legacy normalization fixtures to the recovery probe lists
- fixed the handwritten seen_videos primary-key SQL to match Drift's schema shape
- corrected MIGRATIONS.md to describe the actual MigrationStrategy.onUpgrade path
- updated the PR description with the last_read_timestamp note

Verification:
- cd mobile/packages/db_client && dart run build_runner build --delete-conflicting-outputs
- cd mobile/packages/db_client && dart run drift_dev make-migrations
- cd mobile/packages/db_client && flutter test test/drift/app_database/migration_test.dart test/src/database/app_database_test.dart
- cd mobile/packages/db_client && flutter analyze
- cd mobile/packages/db_client && flutter test
- cd mobile && flutter analyze
- pre-push hook: analyzer, generated files, changed tests

Filed follow-up #7040 for wiring Drift to own the index set instead of growing this PR.

@hm21 please re-review when CI finishes.

hm21 (approved):
Re-approved. Checked the two changes since my last pass.

The `seen_videos` primary-key fix is load-bearing, and now provably so: `drift_schema_v1.json` predates that table while v2 has it, so `legacy v1 schema migrates to v2` finally has a real delta to migrate rather than validating an already-complete database. I reverted your fix locally and the test fails with `Expected the table to have 1 table constraints, it actually has 0` — so the handwritten inline-`PRIMARY KEY` form really was drifting from `createAll()`, and the migration test is what catches it from here on.

The fixture/probe binding test is a better answer to my note than the note was. Lifting the lists into `@visibleForTesting` constants means the drop list in the legacy-upgrade test and the recovery probe can't diverge silently, which is the actual failure mode.

Locally: 919 tests green in `db_client`, `flutter analyze` clean, and `DB Client CI` green on 988eebe. Nothing outstanding from my side — #7040 is the right home for the index-ownership cleanup. Yours to merge.

PR #6978, opened 2026-08-09
ci(e2e): guard Maestro suite against copy drift

MartinMontero:
Closes #7477

## Description

Follow-up to #6948 (Maestro smoke suite repair). The suite rotted silently between 2026-03 and 2026-08 partly because copy in `lib/l10n/app_en.arb` and the flows' asserted literals diverged with nothing failing. This PR adds a blocking copy-drift guard so a copy change that invalidates a Maestro assertion fails CI at the cause.

- `mobile/scripts/check_maestro_copy_drift.sh` extracts literals the flows assert/tap (`assertVisible`, `assertNotVisible`, `tapOn`, `longPressOn`, `text:` props; skips `${}` interpolations and ids), binds ARB-backed literals to `app_en.arb` keys via the manifest, and fails with named DRIFT / UNREGISTERED / ERODED messages.
- `mobile/scripts/baseline/maestro_copy_manifest.txt` records 91 TAB-separated bindings (`arb_key⇥flow path[⇥rendered:literal]`); `rendered:` rows are hand-maintained for ICU-parameterized copy.
- The guard now refuses `UPDATE_BASELINE=1` when regeneration would silently erase a binding whose ARB key and flow file still exist. Intentional removals require `ACCEPT_REMOVALS=1`.
- Once this manifest lands on `main`, the guard ratchets against `origin/main` by default (`MAESTRO_COPY_DRIFT_BASE_REF`) so a branch cannot bless away a base binding while the key and flow still exist. It bootstrap-skips only until the manifest exists on the base ref.
- The existing `mobile_ci` Generated Files job runs this guard plus `e2e/maestro/scripts/check_refs.sh` as a blocking static check. `check_refs.sh` landed in #6948 but was not previously wired into CI.
- `mobile/test/tools/maestro_copy_drift_detector_test.dart` pins the detector behavior against synthetic ARB/flow fixtures.

Refs #4337. Refs #6948. Follow-up hardening is tracked in #7213.

## Known v1 Limits

- The guard only binds literals that already match `app_en.arb`, so already-drifted flow copy is visible in the denominator but not blocked yet.
- Substring matching can miss copy shortening until the planned exact-literal inverse check lands.
- #7213 tracks those limits plus the remaining extractor nits.

## Out of Scope

- Repairing or re-spec'ing `suites/fullRegression.yaml` (needs re-spec, tracked separately).
- The committed credential in `suites/fullRegression.yaml` — human-gated action, not touched here.
[...]

Issue #6952, opened 2026-08-09
task(e2e): repair the remaining smoke flows (search, comment)


realmeylisdev:
## Problem

#6948 repaired the smoke suite. Three flows remain broken, each for a different reason.

### `searchUserFlow` — needs a rewrite, not a patch

The search screen moved from a three-tab `TabBar` to sliver sections plus a filter pill (`search_results_view.dart`). Every selector keyed on `Users (n) / Tab 2 of 3` describes a UI that no longer exists. The result rows carry a pubkey-keyed id (`search_user_tile_$pubkey`), which a flow cannot know in advance — which is why the current test taps a fixed screen coordinate, `point: 50%,32%`.

**Needed:** an index-keyed `search_user_tile_$index` alongside the existing id, section-header identifiers, and a rewrite of `assertSearch.yaml` / `searchUsers.yaml`.

Note `utils/searchWorkaround.yaml` — a stop-app-relaunch-retype ritual someone wrote to make search pass. Do not port it forward; if search needs it, search has a bug and that deserves its own issue.

### `commentFlow` — blocked on cleanup and a real bug

- `tests/commentVideo.yaml` writes `output.FULL_COMMENT` (`:6`), types `${output.FULL_COMMENT}` (`:29`), then asserts the bare `${FULL_COMMENT}` (`:32`) — the *env* default, which is `""`. `env` is read-only inside `evalScript`; the writable namespace is `output.*`.
- The `deleteComment` cleanup in `flows/commentFlow.yaml:22-27` is commented out, so each run leaves a real comment behind. Re-enabling it also means moving it **before** the back-navigation block above it — `deleteComment.yaml` starts by tapping `comments_button` and needs the video still open.
- The comments header dropped its `\d+ Comments` count, which was also the swipe anchor for dismissing the sheet. `close_comments_button` and `comments_drag_handle` already exist and should be used instead.

### `videoUnlike` — blocked on content, tracked in #6949

Opening a liked video in the fullscreen feed shows the loading placeholder and never resolves, so no action buttons render. Not a selector problem.

## Approach

Follow the pattern #6948 established: target `Semantics(identifier:)`, add ids to the app where they are missing, and verify each flow against a real simulator or emulator before moving on. `mobile/e2e/maestro/README.md` documents the loop, including the trick that most `tests/` and all `asserts/` files run against whatever is already on screen, so a single assert re-runs in about three seconds.

Depends on #6949 for anything that touches like state or comments.

Issue #6949, opened 2026-08-09
task(e2e): make Maestro flows independent of live staging content


realmeylisdev:
## Problem

The Maestro E2E flows assert against live STAGING content and mutate it. That is now the binding constraint on the suite, not selector rot — #6948 repaired the selectors and the remaining failures are all data.

Concretely, the suite depends on:

- **"the first video in the New grid"** — whichever video is newest at run time. `videoLike` likes it, `videoUnlike` unlikes it, `commentVideo` comments on it. It currently fails: opening that video in the fullscreen feed shows the loading placeholder and never resolves, so no action buttons render and there is nothing to assert.
- **a specific account existing and ranking first in search** (`searchUserFlow`).
- **a video with no prior likes**, so the liked tab holds exactly one entry — `videoUnlike`'s comment says as much: *"given the user ... only has 1 video liked"*.

It also **creates a real account per run** and publishes real likes. `commentFlow` publishes a real comment and its cleanup step is commented out, so every run leaves one behind on a stranger's video.

## Why it matters

An E2E gate that fails for reasons unrelated to the diff trains people to ignore it — which is roughly how this suite ended up unmaintained for six months. No amount of selector work fixes it.

## Possible directions

Not a recommendation, just the options as they look from here:

1. **Seed known fixtures on STAGING** — a stable account and a video guaranteed playable, referenced by id rather than "whatever is newest".
2. **Point the suite at `local_stack`**, which already seeds deterministic data. Bigger: Codemagic would need to run the stack, and it overlaps with what Patrol covers.
3. **Narrow the suite** to flows that do not depend on other people's content.

## Context

`mobile/e2e/maestro/README.md` documents the fragility under "Environments and test data". Surfaced while repairing the suite in #6948.

PR #6979, opened 2026-08-09
ci(e2e): run iOS Maestro smoke on PRs to main, pinned and JUnit-reported

MartinMontero:
## ⚠️ Prerequisite before merge (confirmed 2026-08-10)

**Create the `maestro_e2e_credentials` variable group in Codemagic first** (vars: `MAESTRO_USER_EMAIL`, `MAESTRO_USER_PWD`, both marked Secret). Reviewer verification: the group does not exist at app level, and triggering `e2e-smoke-ios` on `main` was rejected by Codemagic's config validator — `references to unknown variable group(s): maestro_e2e_credentials` — before any machine was provisioned. Consequences:

- Until the group exists, **no run of either e2e workflow can start at all** — including manual dispatch of the lanes #6948 repaired. The e2e lane on `main` is currently dead; creating the group revives it independently of this PR.
- If this PR merged first, every PR to `main` would show an instant red **config error** — not the credential guard's explanatory message (the script never runs), and no JUnit report.

## Description

Follow-up to #6948. The repaired smoke suite still only runs when someone starts it by hand. This PR makes `e2e-smoke-ios` run on every pull request targeting `main`, and hardens the things that would otherwise make an automatic suite flaky, opaque, or expensive:

- **Pin Maestro**: `MAESTRO_VERSION=2.8.0` in `&install_maestro` — verified green on Codemagic itself: probe builds on the current image (Xcode 26.4.1, 17E202) passed the credential-free smoke flows 2/2 (review round 3, 2026-08-11). The #6948 baseline (2.1.0) predates Maestro's iOS-26 runner fixes (2.5.0, 2.7.0, 2.8.0); its driver never starts on any Xcode 26 toolchain (upstream mobile-dev-inc/Maestro#2932), so the pin moves up rather than the toolchain moving down. An unpinned `get.maestro.mobile.dev` install lets a Maestro release break CI with no commit on our side — bump deliberately, with a probe build, then merge the bump.
- **Pin Xcode**: `xcode: 26.4` on `e2e-smoke-ios`, paired with the Maestro pin so neither side of the compatibility contract moves alone. 26.4 is the CI-verified image (same probes); `latest` resolving past it is the only unverified move left. The image ships iOS 26 runtimes only, so the simulator runtime follows the Xcode image either way.
- **JUnit reporting, per-flow**: the anchor invokes the three active flows directly (`loginFreshInstall`, `removeKeys`, `loginEmailPwd` — kept in sync with `suites/smoke.yaml`) with `--format junit --output report.xml`, so `test_report` shows one testcase per flow with name, file, timing and tags (probe-verified) instead of suite mode's single unnamed testcase. `report.xml` is an artifact and `test_report` is wired on both `e2e-smoke-ios` and `e2e-smoke-android`.
- **PR trigger (iOS only)**: `events: [pull_request]`, `branch_patterns: [{pattern: 'main', include: true, source: false}]`, `cancel_previous_builds: true`. Android stays `events: []` — one platform proves the trigger mechanics before we spend linux_x2 minutes on every PR.
- **Cost controls** (review round 1): `max_build_duration: 20` on `e2e-smoke-ios` (the only full run on record took 7m49s end to end, 2026-07-29 — 20 min is ~2.5x with cold-cache headroom; a hang bills 20 minutes, not 60), and a `when: changeset: includes: ['mobile/']` scope so docs/website/CI-only PRs never spin up a Mac Mini.
- **Fork-aware failure message**: the credential guard in `&run_maestro_smoke_tests` explains that Codemagic withholds secure variables from outside-contributor (fork) builds by design, so a fork PR stops with a clear "not a code failure — a maintainer can re-run it" instead of a bare missing-variable error. The same red-by-design behavior and the maintainer re-run path are documented in `mobile/e2e/maestro/README.md`.

**Non-blocking by design**: per team decision this workflow must NOT be marked as a required check until a flake baseline exists (target: ≥30 runs, ≥95% pass) — and that baseline is gated on **#7104**: the probes found a fixed ~3m31s hierarchy-fetch stall on the Explore screen that makes `loginFreshInstall` a coin flip, so a baseline collected before that fix measures the stall, not the app. #7104 also tracks `~/.maestro/tests/**` artifact collection (minus the 153 MB/flow simulator log) and the simulator-selection triage note.

**If PR builds do not fire after merge**: check the GitHub webhook for the Codemagic app (repo Settings → Webhooks) before touching this YAML — a misconfigured webhook presents exactly like a YAML mistake.

## Out of Scope

- Android PR trigger (follow after iOS baseline).
- `suites/fullRegression.yaml` re-spec and its committed credential (human-gated).
- Copy-drift guard (separate PR #6978).
- The Explore stall fix, artifact glob, and simulator-selection docs (#7104).
- Codemagic spend sign-off: with the 20-min ceiling + path scoping, worst case is one ~20-minute m2 build per relevant PR push (bounded by `cancel_previous_builds`) — the yes/no on that budget remains a team call.

## Verification

- `codemagic.yaml` parses (anchors resolve); the diff touches only the e2e lanes.
- Trigger block matches the Codemagic YAML schema: `events`, `branch_patterns` with `source: false` (target-branch match), `cancel_previous_builds`, `when: changeset`.
- Suite health verified by the reviewer with probe builds on Codemagic (round 3): Maestro 2.8.0 / Xcode 26.4.1 (17E202), credential-free flows 2/2 green; per-flow invocation yields per-flow JUnit testcases. 2.1.0's driver never starts on the same toolchains (#2932).
- The credential path cannot be end-to-end verified until the group exists and a full run fires — expected; the non-blocking posture plus the prerequisite above bound the blast radius.

## Type of Change

- [x] CI/Tooling
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change


rabble (commented):
The mechanics here are right, and a couple of things I expected to be wrong aren't. Checking the parts that are easy to get subtly wrong:

- **The version pin is real, not cargo-culted.** `#6948`'s description says "Maestro 2.1.0", so `MAESTRO_VERSION=2.1.0` pins to the version the repaired suite was actually verified against on device. That's the correct reason to pin, and the comment says so.
- **JUnit output reaches both platforms.** `maestro test` appears exactly once in the file (a shared anchor), so adding `--format junit --output report.xml` there covers iOS and Android, which is why adding `report.xml` + `test_report` to both workflows is consistent rather than copy-paste.
- **The fork-secrets behaviour is correctly understood.** Codemagic does withhold secure variables from outside-contributor builds, and putting that explanation in the failure message — rather than letting Maestro resolve an empty variable and report it as a selector failure — is exactly right.
- **"Non-blocking" is accurate.** The `CI workflow rule` ruleset requires only the `Mobile CI` context, so this workflow cannot block a merge no matter what it reports. The comment telling future readers not to promote it to required yet is well-placed.
- `cancel_previous_builds: true` is the right default here and limits the damage from the point below.

## The thing I'd want a decision on before this merges

This flips `events: []` (manual) to running on **every PR targeting main**, on:

```yaml
instance_type: mac_mini_m2
max_build_duration: 60
```

That is the expensive instance, with a 60-minute ceiling, per PR — and this repo currently has ~30 open PRs with several pushes a day. A hung simulator or a suite that stops making progress burns the full hour rather than failing fast.

None of that is a defect in your change; it's a spend decision that this PR quietly makes on the team's behalf. Two things would de-risk it without giving up the coverage:

1. **Drop `max_build_duration` to something near the suite's real runtime** (plus headroom). 60 minutes is a backstop for a build that has already gone wrong; there's no reason to pay for the whole hour to learn that.
2. **Consider scoping the trigger** — e.g. only PRs that touch `mobile/lib`, `mobile/ios`, or `e2e/` — so docs and CI-only PRs don't each spin up a Mac Mini. Several of the repo's other jobs already gate on a changed-paths filter (`Detect App CI Scope`).

Worth confirming with whoever watches the Codemagic bill before this lands.

## One consequence worth naming

Fork PRs will fail this check by design, at the credential guard. That includes **your own** — both of these PRs are from `MartinMontero/divine-mobile`, so once this merges, every PR you open will show a red E2E check that you cannot make green yourself, and you won't be able to self-verify changes to the very suite you own.

That's not an argument against the change, but it's an argument for a maintainer re-run being a documented step rather than folklore — your failure message already points at it, which is more than most repos manage. Might be worth a line in the E2E docs too.

No code changes requested — the implementation is sound. Holding at comment rather than approve only because the cost/trigger scope is a team call, not mine.

MartinMontero:
@rabble Both de-risking items are in, plus the docs line — two commits pushed to the branch:

**`fa1262d9` — cost controls on the trigger:**
- `max_build_duration: 60 → 40` on `e2e-smoke-ios` only, with a comment to tighten again once the baseline produces real timings. 40 is ~2x the observed build-plus-smoke runtime; the 60-minute backstop was exactly the "pay for the whole hour to learn the simulator hung" case you flagged.
- Trigger scoped with `when: changeset: includes: ['mobile/']` — app code, native config, pubspec, and the suite itself all live there, so docs/website/CI-only PRs no longer spin up a Mac Mini. Codemagic semantics worth knowing (now in the YAML comment): `codemagic.yaml` changes always run regardless, the first build after this lands runs regardless, the comparison is against the last successful build, and a "skipped" run still starts and stops early — so it costs seconds, not a full build.

**`6277c7b2` — the fork-PR consequence, documented:** `mobile/e2e/maestro/README.md` now states that fork PRs fail the e2e check by design (credentials withheld → credential guard → red check the contributor can't clear), names it not-a-code-failure, and gives the maintainer path (pull the branch into the repo, or re-run on their behalf). It also fixes the now-stale "both workflows are manual" paragraph. You're right that my own PRs are the standing example of this — noted and accepted; the re-run step being written down is what makes it workable.

**Left as a team call, as you framed it:** the Codemagic spend sign-off. With the 40-min ceiling + path scoping, the worst case is one ~40-minute m2 build per relevant PR push (bounded further by `cancel_previous_builds`), and zero for docs/CI-only PRs — but the yes/no on that budget still belongs to whoever watches the bill, not to this PR.

rabble:
Verified the mechanics against the Codemagic API and a local simulator run. The wiring in this PR is correct, but three things should be settled before the PR trigger is enabled.

### 1. `maestro_e2e_credentials` does not exist (blocking)

The `divine-mobile` Codemagic app defines exactly four variable groups: `firebase`, `google_play_credentials`, `proofmode_credentials`, `zendesk_credentials`. A raw search of the app payload returns **0** occurrences of `maestro_e2e_credentials`, `MAESTRO_USER_EMAIL`, and `MAESTRO_USER_PWD`, against 4 for `zendesk_credentials` — so the payload does carry group names and the absence is real.

Caveat stated rather than glossed: Codemagic also supports team-level shared groups, and the `/teams` endpoints are not reachable with the token I used, so this rules out an app-level group definitively and a team-level one only presumptively.

As it stands, enabling `events: [pull_request]` would make `&run_maestro_smoke_tests` exit 1 at the guard on every PR to `main`, before a single test runs.

Separately and out of scope here: `github_credentials` is missing the same way, and it is referenced by three production workflows (lines 530, 641, 703). Worth its own issue.

### 2. `xcode: latest` should be pinned too

The reasoning behind `MAESTRO_VERSION=2.1.0` is right — an unpinned install lets an upstream release break CI with no commit on our side. But the compatibility contract has two sides and this workflow pins one.

Running the suite locally against **Xcode 26.2**, Maestro 2.1.0 never starts its iOS driver: nothing binds port 7001, and raising `MAESTRO_DRIVER_STARTUP_TIMEOUT` to 180s changes nothing because the process is absent rather than slow. That is upstream [Maestro #2932](https://github.com/mobile-dev-inc/Maestro/issues/2932) — `xcodebuild test-without-building` returns instantly instead of holding the runner alive.

So if Codemagic's image resolves `latest` to 26.2, this suite cannot go green no matter what the credentials say — and the Maestro pin makes recovery harder, since moving to a fixed Maestro breaks the "2.1.0 is the verified baseline" claim from #6948. Suggest pinning `xcode:` to an explicit known-good version alongside the Maestro pin.

### 3. `max_build_duration: 60` is ~7x the real run

The only `E2E Smoke Tests (iOS)` build in history (`6a6975794e492a319cf46a23`, 2026-07-29) took **7m49s** end to end. A 60-minute ceiling on an 8-minute job turns a hang into an hour of billed `mac_mini_m2`.

Worth noting that build failed on `You have 0 devices connected` — the defect #6948 repaired — and its recorded command had no credential guard, so it predates this anchor and says nothing about credentials. Every step through `Install app on iOS Simulator` was green, so the pipeline mechanics are proven.

On cost: at ~8 minutes, per-PR spend is comfortable. The remaining argument against every-PR is coverage, not money — `suites/smoke.yaml` currently runs three login flows, with `likeFlow`/`commentFlow`/`searchUserFlow` held back on #6949 and #6952.

rabble:
Follow-up to my comment above — one correction and one stronger result.

### Correction: `github_credentials` is fine

I said it was missing and suggested tracking it as a defect. That was wrong; withdrawing it.

It is referenced by `ios-build`, `android-build`, and `macos-build`, and those workflows succeed — 24 of the last 28 production builds finished, most recently `iOS Build` on 2026-08-08. Since Codemagic rejects a workflow at validation when it names an unknown group (see below), the group must exist. The app-payload API query I used does not enumerate team-level shared groups, so it produces false negatives and can't support an absence claim on its own.

### `maestro_e2e_credentials` is confirmed missing, by Codemagic itself

I triggered `e2e-smoke-ios` on `main` through the API. It was rejected before any machine was provisioned (`startedAt: null`):

```
Codemagic.yaml references to unknown variable group(s): maestro_e2e_credentials
```

That is authoritative at every scope and closes the team-level caveat from my previous comment. It also cost nothing — no instance was allocated.

### This changes the failure mode

I previously described this as the credential guard exiting 1. It's earlier than that: **config validation rejects the workflow before step one runs.**

With `events: [pull_request]` enabled, every PR to `main` would get an instant red config error rather than a test failure. Two consequences worth weighing:

- The guard's helpful message (`Missing $var — add it to the maestro_e2e_credentials group`) never prints, because the script never executes. The developer sees a bare config error instead.
- The JUnit reporting this PR adds never produces a report, so there's nothing in `test_report` to look at.

The fix is unchanged — create the group before enabling the trigger — but the blast radius if it merges first is worse than my earlier comment implied.

The Xcode pinning point (#2 above) is unaffected and still stands; it can't be tested until the group exists.

MartinMontero:
@rabble Round two addressed — commit `1667c406`, and the PR body now leads with the prerequisite. Point by point:

**1. `maestro_e2e_credentials` missing (blocking) — confirmed and made the merge gate.** Your validator probe settles it: if the workflow is rejected at config validation, the credential guard's message never prints and there's no JUnit report — the failure mode is worse than either of us initially described, and you're right that merging first would paint every PR red with a bare config error. The PR body now opens with a **Prerequisite before merge** section: create the group (`MAESTRO_USER_EMAIL`, `MAESTRO_USER_PWD`, both Secret) first. One consequence worth stating plainly: since `main` already references the group, **manual dispatch of both e2e lanes is failing validation today** — the #6948 repair is un-runnable in CI until the group exists, independent of this PR. Creating it is the single unblocking action for everything.

**2. Xcode pinned — `xcode: 26.1`.** Verified the upstream issue (#2932: driver never binds port 7001 on 26.2, `test-without-building` exits instantly) and Codemagic's current image ladder: `latest` has resolved to 26.4 since 2026-04-26, which is unverified for Maestro 2.1.0 and plausibly carries the same regression. 26.1 is the newest image with no reported breakage (the iOS 26 beta issue #2609 was fixed in 2.0.5; the reported regression begins at 26.2), it keeps SDK parity with the production builds, and its image ships iPhone 16 Pro / 17 Pro simulators the boot script prefers. The YAML comment says to verify on the first real run and, if the driver still fails, to drop to 16.4 or move both pins past the upstream fix together. Deliberately scoped to `e2e-smoke-ios` — production workflows keep `latest` and Apple's Xcode-26 upload requirement is untouched.

**3. Ceiling tightened to measured reality — `max_build_duration: 20`.** Your 7m49s number beats my estimate: 20 is ~2.5x with cold-cache headroom, and a hang now bills 20 minutes, not 60. Comment cites the 2026-07-29 run so the next person knows where the number came from.

On the coverage point you closed with: agreed — per-PR spend at ~8 minutes is comfortable either way, and the real argument for the trigger is catching harness/build regressions while the social flows are held back on #6949/#6952, not pretending the suite covers them.

rabble:
Ran the suite locally. **Good news: the suite is fine — the pinned Maestro version is the problem.** This changes two of my earlier recommendations.

### The suite passes end to end on Maestro 2.8.0

Both credential-free smoke flows go green on macOS 15.7.4 / Xcode 26.2:

```
 > Flow loginFreshInstall
Launch app "co.openvine.app" with clear state... COMPLETED
  Assert that id: create_account_button is visible... COMPLETED
  Assert that id: home_tab is visible... COMPLETED
Run ../asserts/assertExploreNew.yaml... COMPLETED

 > Flow removeKeys
Run ../utils/openSettings.yaml... COMPLETED
  Tap remove_keys_tile / Cancel / remove_keys_tile / Remove from device
Run ../asserts/assertLogin.yaml... COMPLETED
```

On **Maestro 2.1.0**, the version this PR pins, the same flows on the same machine never start at all — the iOS driver never binds port 7001 ([upstream #2932](https://github.com/mobile-dev-inc/Maestro/issues/2932)), and `MAESTRO_DRIVER_STARTUP_TIMEOUT` doesn't help because the process is absent rather than slow.

### Revised recommendation: raise the Maestro pin, don't lower Xcode

My earlier comment suggested pinning `xcode:` back to something 2.1.0 works with. That was the wrong direction. 2.1.0 predates every iOS-26 fix in Maestro's changelog — 2.5.0 (XCTest UI-interruption preflight hang in the iOS runner), 2.7.0 (iOS 26 hierarchy retrieval), 2.8.0. The pairing #6948 verified has aged out rather than the toolchain having gone wrong.

Suggest `MAESTRO_VERSION=2.8.0`, plus an explicit `xcode:` pin so neither side of the contract can move on its own.

For reference, **Codemagic's `latest` currently resolves to Xcode 26.4.1** (build 17E202) — from the `Installing SDKs` log of a successful `iOS Build` on 2026-08-10, which prints `== Use Xcode 26.4 ==`. I could not verify 2.8.0 against 26.4.1 locally; that needs one real build. ([Maestro #3218](https://github.com/mobile-dev-inc/Maestro/issues/3218) mentions Xcode 26.4, but it's physical-device only and its author says the simulator path still works, so it likely doesn't apply here.)

### `--format junit` doesn't name the failing flow

Worth knowing before relying on `test_report`. Running the real `suites/smoke.yaml` with the exact flag this PR adds produces one testcase for the entire suite:

```xml
<testsuite name="Test Suite" tests="1" failures="1" time="25.979">
  <testcase id="Divine Smoke regression test" name="Divine Smoke regression test"
            file="e2e/maestro/suites/smoke.yaml" status="ERROR">
    <failure>Assertion is false: false is true</failure>
```

No per-flow granularity, and the failure text doesn't name `loginEmailPwd` — the flow that actually failed (expected, since I supplied no credentials). So the Codemagic UI will show "the suite failed" and nothing more, which is most of what log spelunking was costing in the first place.

Invoking Maestro per-flow (or in directory mode) so each flow becomes its own `<testcase>` would fix that, though I haven't verified it.

### One methodological warning

My first local run failed on `create_account_button is visible` and looked like a real regression. It wasn't — Maestro booted a *different* simulator than the one I'd installed the build on, and that device had a stale bundle. Re-installing onto the device Maestro actually selects turned it into a full pass.

`&install_app_ios_simulator` picks the first booted device and `&boot_ios_simulator` picks from a preference list, so they should agree — but if this suite ever reports a mystery selector failure in CI, check which binary is on the device Maestro chose before believing it.

MartinMontero:
@rabble Round 3 — your local run settles the direction, and the pairing argument is convincing: the #6948-verified pairing aged out rather than the toolchain going wrong. Patch below (applies with `git am` on this branch).

**1. Raise the Maestro pin, don't lower Xcode — done.** `MAESTRO_VERSION: 2.1.0 → 2.8.0`, with the rationale rewritten in the anchor comment: 2.1.0 predates every iOS-26 runner fix (2.5.0, 2.7.0, 2.8.0) and its driver never starts on any Xcode 26 toolchain (#2932), so the pin now records *your* verified pairing — 2.8.0 green end-to-end on Xcode 26.2, both credential-free flows.

**2. Xcode pinned as the other half of the contract — `xcode: 26.2`** (the version you verified against, not my earlier 26.1 guess in the opposite direction). `latest` = 26.4.1 stays out until one real build verifies 2.8.0 against it; the comment notes #3218 is physical-device-only so the simulator path is expected-but-not-verified, and that the pins move together or not at all.

**3. JUnit granularity — documented, deferred.** Confirmed your reading: suite mode emits one `<testcase>` for the whole suite, so `test_report` will say "the suite failed" without naming the flow. That's now a comment at the test anchor so nobody over-trusts the report on day one. Per-flow / directory-mode invocation is a real change to the shared anchor (both lanes), and unverified — I'd rather track it as a follow-up than sneak it in here. Filing it once this lands.

**4. Simulator-selection warning — noted for the verification run.** `&install_app_ios_simulator` takes the first booted device while `&boot_ios_simulator` walks a preference list; if the first real CI run reports a mystery selector failure, checking which binary is on the device Maestro chose is now the documented first move (it goes in the follow-up issue too).

The patch:

```diff
From ece85e139f03e743f2e98a82c72f52c25ca9840b Mon Sep 17 00:00:00 2001
From: Martin Montero <scoutmontero@gmail.com>
Date: Tue, 11 Aug 2026 23:38:17 +0000
Subject: [PATCH] ci(e2e): Maestro 2.8.0 + Xcode 26.2 paired pins, JUnit
 granularity note
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

Round 3 on #6979: local verification showed the smoke suite passes
end-to-end on Maestro 2.8.0 with Xcode 26.2 (both credential-free
flows), while 2.1.0's driver never starts on any Xcode 26 toolchain —
the #6948 pairing aged out, so raise the Maestro pin instead of
lowering Xcode. Pin xcode 26.2 explicitly so neither side of the
contract moves alone; latest (26.4.1) stays unverified until one real
build runs after the credentials group exists. Document the suite-mode
JUnit limitation (one testcase for the whole suite) at the test anchor;
per-flow granularity is a follow-up.
---
 codemagic.yaml | 35 +++++++++++++++++++++--------------
 1 file changed, 21 insertions(+), 14 deletions(-)

diff --git a/codemagic.yaml b/codemagic.yaml
index 86b92ec1a..d3ca51c04 100644
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -379,11 +379,14 @@ definitions:
       name: Install Maestro CLI
       script: |
         set -e
-        # Pinned: 2.1.0 is the version the smoke suite was repaired and
-        # verified against on real devices (#6948). An unpinned install lets
-        # a Maestro release break CI with no commit on our side. Bump
-        # deliberately and re-verify the suite before merging the bump.
-        export MAESTRO_VERSION=2.1.0
+        # Pinned: 2.8.0 verified green end-to-end against Xcode 26.2 on a
+        # local simulator run (both credential-free smoke flows, #6979
+        # review). 2.1.0 — the #6948 repair baseline — predates Maestro's
+        # iOS-26 runner fixes (2.5.0, 2.7.0, 2.8.0) and its driver never
+        # starts on any Xcode 26 toolchain (upstream #2932). An unpinned
+        # install lets a Maestro release break CI with no commit on our
+        # side; bump deliberately and re-verify the suite before merging.
+        export MAESTRO_VERSION=2.8.0
         curl -Ls "https://get.maestro.mobile.dev" | bash
         export PATH="$PATH:$HOME/.maestro/bin"
         maestro --version
@@ -405,6 +408,10 @@ definitions:
             exit 1
           fi
         done
+        # JUnit caveat: in suite mode Maestro emits ONE testcase for the
+        # whole suite — test_report will show "the suite failed" without
+        # naming the flow (#6979 review). Per-flow granularity needs
+        # per-flow (or directory-mode) invocation; tracked as a follow-up.
         maestro test \
           --format junit --output report.xml \
           -e USER_EMAIL="$MAESTRO_USER_EMAIL" \
@@ -741,15 +748,15 @@ workflows:
     instance_type: mac_mini_m2
     environment:
       flutter: 3.44.0
-      # Pinned, not latest: Maestro 2.1.0's iOS driver never starts on Xcode
-      # 26.2 (upstream mobile-dev-inc/Maestro#2932 — xcodebuild
-      # test-without-building exits instantly and port 7001 never binds), and
-      # latest now resolves to 26.4, which is unverified. 26.1 is the newest
-      # Codemagic image with no reported Maestro breakage. Verify on the
-      # first run after the maestro_e2e_credentials group exists; if the
-      # driver still fails, drop to 16.4 or move the Maestro pin past the
-      # upstream fix together with this one.
-      xcode: 26.1
+      # Pinned as a pair with MAESTRO_VERSION so neither side of the
+      # compatibility contract moves alone: Maestro 2.8.0 passed the smoke
+      # suite end-to-end on Xcode 26.2 in local verification (#6979 review).
+      # latest resolves to 26.4.1 — unverified with 2.8.0 (upstream #3218
+      # concerns physical devices only, so the simulator path is expected
+      # to work, but expected is not verified). Verify with one real build
+      # after the maestro_e2e_credentials group exists; move both pins
+      # together when bumping toward latest.
+      xcode: 26.2
       cocoapods: default
       groups:
         - zendesk_credentials
-- 
2.39.5
```

Validated with `git apply --check` against the current branch head (`3ae9ef0`). Still outstanding and unchanged: the `maestro_e2e_credentials` group (manual dispatch of both e2e lanes on main fails config validation until it exists) — once it's in, one manual `e2e-smoke-ios` run on this branch verifies 2.8.0+26.2 in CI for real.

rabble:
Ran two probe builds on Codemagic to settle the toolchain question. **Maestro 2.8.0 works on the current image, and the suite passes.**

### Verified on Codemagic

Throwaway branch with `MAESTRO_VERSION=2.8.0`, per-flow invocation, and no credential-group reference (so config validation passes without the missing group):

```
[Passed] loginFreshInstall (3m 57s)
[Passed] removeKeys (16s)
2/2 Flows Passed in 4m 13s
```

Runner reported `maestro 2.8.0` / `Xcode 26.4.1 (17E202)`. So bumping the pin from 2.1.0 to 2.8.0 is sufficient — no Xcode pin is needed for compatibility, and the driver-startup failure 2.1.0 hits on Xcode 26 is gone.

For reference, the image ships **only iOS 26 runtimes** (iPhone 17 Pro/17/Air/16e @ iOS 26.3 and 26.4). There is no iOS 18 available, so the simulator runtime cannot be pinned without changing the Xcode image itself. Good news is that it doesn't need to be.

### The real problem: a fixed 3m31s stall

An earlier probe of the same branch **failed**, and the difference is worth understanding before this goes on every PR.

One hierarchy fetch on the Explore screen stalls for the same interval in both runs:

| Build | Stalled assertion | Interval |
|---|---|---|
| `6a7b94b6` | `"Search.*"` (text) | 21:51:57.704 → 21:55:28.946 = **3m31.2s** |
| `6a7bb0d4` | `explore_search_bar` (id) | 23:51:25.990 → 23:54:57.002 = **3m31.0s** |

Identical to within 200ms, across different assertions and different runs — a timeout expiring, not device load. Whichever assertion arrives first absorbs it. Once past, both id and text assertions match in ~250ms:

```
23:54:58.356  Assert id: explore_search_bar  RUNNING
23:54:58.606  Assert id: explore_search_bar  COMPLETED    <- 250ms
23:54:58.869  Assert "Search.*"              COMPLETED    <- 262ms
```

**This makes `loginFreshInstall` a coin flip.** It survives only when the stall lands on an assertion inside the `Retry 3 times` wrapper — probe 2 got lucky there, probe 1 didn't (the stall hit `"Search.*"` in `assertExploreNew.yaml`, which is outside the wrapper).

It's also the entire gap between 27s locally and ~4m in CI.

Two implications for this PR:

1. **A flake baseline collected before this is fixed would measure the stall, not the app.** The proposed ≥30-run / ≥95% threshold can't mean much while one flow is a coin flip for reasons unrelated to the diff.
2. `max_build_duration: 60` remains ~7x the real runtime; worth lowering regardless.

### Also confirmed

- `~/.maestro/tests/**` artifact collection works well — per-step screenshots, screen-hierarchy JSON, `maestro.log`, `commands.json`. Richer than JUnit for triage. One caveat: `logs/device-simulator.log` is **153 MB per flow**, worth excluding from the glob before this runs on every PR.
- Per-flow invocation (passing flow files instead of `suites/smoke.yaml`) yields one `<testcase>` per flow with name, file, timing and tags — the suite form collapses to a single unnamed testcase, which makes `test_report` much less useful than intended.

rabble (changes requested):
Requesting changes for one blocking Codemagic schema issue. test_report belongs on an expanded script item, not at workflow root. The shared run_maestro_smoke_tests anchor is already the right place: move test_report: report.xml into that script object so both iOS and Android inherit it, and remove both workflow-root test_report keys. As written, Codemagic can reject the workflow schema before either lane runs. The official YAML testing example places test_report on the script item: https://docs.codemagic.io/yaml-testing/testing/ . The ensure-linked-issue check is also currently red and needs a linked issue. This touches protected CI configuration, so I am leaving feedback only rather than pushing a takeover edit.

MartinMontero:
@rabble The probes settle everything outstanding — thank you for running them on real Codemagic iron. Round 3 **v2** patch below (supersedes the v1 patch in my previous comment, which chose `xcode: 26.2` before your probe data existed; v1 was never applied to the branch, so v2 applies cleanly to head `3ae9ef0` with `git am`).

**1. Maestro 2.8.0 + Xcode 26.4, paired.** Your probe builds (`6a7b94b6` / `6a7bb0d4`) verified 2.8.0 on the *current* image — 26.4.1 (17E202), 2/2 flows green — which beats both my 26.1 guess and the 26.2 local pairing: it's CI-verified. So the pin moves to `xcode: 26.4`, and the comment records the probe evidence and your point that the image ships iOS 26 runtimes only, so the simulator runtime follows the image either way. Keeping the pin explicit (rather than `latest`) so a future image move fails loudly instead of silently — bump both pins together, after a probe build, per your pairing rule.

**2. Per-flow invocation — adopted.** Since your probe confirms passing flow files yields one `<testcase>` per flow (name, file, timing, tags), the anchor now invokes the three active flows directly instead of `suites/smoke.yaml`. The comment says to keep the list in sync with the suite — and the copy-drift guard (#6978) binds the same literals, so a forgotten flow surfaces as manifest drift rather than silently dropping out of CI.

**3. The 3m31s stall — tracked as #7104, and it gates the flake baseline.** A fixed 3m31.0–3m31.2s interval across different assertions and runs is a timeout expiring, full stop. Agreed with both implications: `loginFreshInstall` is a coin flip until it's fixed, and a ≥30-run baseline collected now would measure the stall, not the app. Filed #7104 with your table, the artifact-collection win (`~/.maestro/tests/**`, minus the 153 MB/flow `device-simulator.log`), and the simulator-selection triage note. The PR's non-blocking-until-baselined posture now explicitly sits behind that issue.

The patch (validated with `git apply --check` against `3ae9ef0`):

```diff
From e1bbb2967c14fcd2c20d2817929f637579bb22dd Mon Sep 17 00:00:00 2001
From: Martin Montero <scoutmontero@gmail.com>
Date: Wed, 12 Aug 2026 02:56:52 +0000
Subject: [PATCH] ci(e2e): Maestro 2.8.0 + Xcode 26.4 paired pins, per-flow
 JUnit
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

Round 3 on #6979, revised after probe builds on Codemagic itself:
Maestro 2.8.0 passes the credential-free smoke flows 2/2 on the current
image (Xcode 26.4.1, 17E202), so raise the Maestro pin from 2.1.0 (its
iOS driver never starts on any Xcode 26 toolchain, upstream #2932) and
pin xcode 26.4 as the CI-verified other half of the contract — latest
moving past it is the only unverified move left. Invoke Maestro
per-flow instead of via suites/smoke.yaml so the JUnit report carries
one testcase per flow (name, file, timing, tags) instead of a single
unnamed suite-level testcase; the flow list mirrors smoke.yaml's three
active flows. The fixed ~3m31s Explore hierarchy-fetch stall and
~/.maestro/tests artifact collection (minus the 153MB/flow simulator
log) are tracked as follow-ups — the flake baseline stays gated until
the stall is fixed, since a baseline measured now measures the stall.
---
 codemagic.yaml | 42 +++++++++++++++++++++++---------------
 1 file changed, 27 insertions(+), 15 deletions(-)

diff --git a/codemagic.yaml b/codemagic.yaml
index 86b92ec1a..ad67a6ffa 100644
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -379,11 +379,14 @@ definitions:
       name: Install Maestro CLI
       script: |
         set -e
-        # Pinned: 2.1.0 is the version the smoke suite was repaired and
-        # verified against on real devices (#6948). An unpinned install lets
-        # a Maestro release break CI with no commit on our side. Bump
-        # deliberately and re-verify the suite before merging the bump.
-        export MAESTRO_VERSION=2.1.0
+        # Pinned: 2.8.0 verified green on Codemagic itself — probe builds on
+        # the current image (Xcode 26.4.1, 17E202) passed the credential-free
+        # smoke flows 2/2 (#6979 review, 2026-08-11). 2.1.0, the #6948 repair
+        # baseline, predates Maestro's iOS-26 runner fixes (2.5.0, 2.7.0,
+        # 2.8.0) and its driver never starts on any Xcode 26 toolchain
+        # (upstream #2932). Bump deliberately, with a probe build, and
+        # re-verify the suite before merging the bump.
+        export MAESTRO_VERSION=2.8.0
         curl -Ls "https://get.maestro.mobile.dev" | bash
         export PATH="$PATH:$HOME/.maestro/bin"
         maestro --version
@@ -405,11 +408,20 @@ definitions:
             exit 1
           fi
         done
+        # Per-flow invocation, not suites/smoke.yaml: one JUnit <testcase>
+        # per flow with name, file, timing and tags (probe-verified on
+        # Codemagic, #6979 review); suite mode collapses the whole run into
+        # a single unnamed testcase, which makes test_report near-useless.
+        # Keep this list in sync with suites/smoke.yaml — the copy-drift
+        # guard (#6978) binds the same literals, so a forgotten flow shows
+        # up as manifest drift on the next ARB change.
         maestro test \
           --format junit --output report.xml \
           -e USER_EMAIL="$MAESTRO_USER_EMAIL" \
           -e USER_PWD="$MAESTRO_USER_PWD" \
-          e2e/maestro/suites/smoke.yaml
+          e2e/maestro/tests/loginFreshInstall.yaml \
+          e2e/maestro/tests/removeKeys.yaml \
+          e2e/maestro/flows/loginEmailPwd.yaml
     - &boot_ios_simulator
       name: Boot iOS Simulator
       script: |
@@ -741,15 +753,15 @@ workflows:
     instance_type: mac_mini_m2
     environment:
       flutter: 3.44.0
-      # Pinned, not latest: Maestro 2.1.0's iOS driver never starts on Xcode
-      # 26.2 (upstream mobile-dev-inc/Maestro#2932 — xcodebuild
-      # test-without-building exits instantly and port 7001 never binds), and
-      # latest now resolves to 26.4, which is unverified. 26.1 is the newest
-      # Codemagic image with no reported Maestro breakage. Verify on the
-      # first run after the maestro_e2e_credentials group exists; if the
-      # driver still fails, drop to 16.4 or move the Maestro pin past the
-      # upstream fix together with this one.
-      xcode: 26.1
+      # Pinned as a pair with MAESTRO_VERSION so neither side of the
+      # compatibility contract moves alone. Probe builds on this exact
+      # image (2026-08-11, #6979 review) ran Maestro 2.8.0 / Xcode 26.4.1
+      # (17E202) with the smoke flows 2/2 green — 26.4 is CI-verified, and
+      # latest resolving past it is the only unverified move left. The
+      # image ships iOS 26 runtimes only, so the simulator runtime follows
+      # the Xcode image either way. Move both pins together, after a probe
+      # build, when bumping toward a newer image.
+      xcode: 26.4
       cocoapods: default
       groups:
         - zendesk_credentials
-- 
2.39.5
```

Unchanged blocker, still the single unblocking action for everything: the `maestro_e2e_credentials` group. Your probes dodged it by dropping the group reference on a throwaway branch, but this PR's lanes reference it for real — and manual dispatch on main still fails config validation until it exists.

rabble:
I've opened #7208, which carries this PR's intent through to a verified green run. Per `PR_REVIEW.md` the findings here were codeable, so coding them was the right call rather than leaving a fourth round of review comments — apologies for the three that came first.

**This PR had the right diagnosis.** Pinning Maestro, JUnit reporting, PR triggering, non-blocking posture, the fork-credentials caveat, and the webhook warning are all correct and all carried over.

What it could not have caught, because **its branch lives on a fork Codemagic cannot reach** — I confirmed this by trying: `Cannot access the repository: 'ci' does not exist in divinevideo/divine-mobile`:

1. **`test_report` at workflow level is rejected by Codemagic's schema** — `extra fields not permitted`. It is a script-level key. The config fails validation before provisioning, so the JUnit reporting this PR adds would never have produced a report.
2. **Maestro 2.1.0 cannot start its iOS driver on Xcode 26.x at all** ([#2932](https://github.com/mobile-dev-inc/Maestro/issues/2932)) — the XCTest runner installs, reports success, and never binds port 7001. Codemagic's image is Xcode 26.4.1. The pin was the right instinct but landed on a version that cannot run there; 2.8.0 is green on that image.
3. **`maestro_e2e_credentials` does not exist**, and Codemagic validates the whole file before provisioning — so referencing it fails *every* workflow, not just this one. Same mechanism that took the pipeline down for ~21 hours via `supporters_credentials` (#7203). #7208 drops the reference and runs the credential-free flows instead.
4. **`xcode: latest` left half the compatibility contract floating**, so an image bump could break the suite through exactly the failure the Maestro pin prevents.

Two things worth taking from this beyond the config:

**Working from a fork is costing you.** You have `write` on this repo, and every other contributor pushes branches to it. Because `ci/maestro-smoke-trigger` only exists on your fork, Codemagic cannot build it — so this PR was unverifiable, and once the gate is enabled your own PRs would get a permanent red check from the credential guard, on the very suite you maintain. Pushing branches to `divinevideo/divine-mobile` fixes all of that.

**#7204 is the real unlock, not the credential group.** `likeFlow` and `commentFlow` need no credentials at all — each creates a throwaway account and cleans up. They are blocked only on a fixed hierarchy stall: measured at 9m12s for `likeFlow` before a driver snapshot timeout. Fixing that buys the actual product coverage.

Happy for you to take #7208 over, or to close this in its favour — your call, it is your work that got it here.

MartinMontero:
Superseded by #7208, which carries this PR's intent through to a verified green run — closing in its favour.

What this PR diagnosed is all carried over: the paired Maestro/Xcode pins (2.8.0 + 26.4, probe-verified on Codemagic), per-flow JUnit invocation, PR triggering with a non-blocking-until-baselined posture, the changeset gate, and the fork-credentials caveat. What it could not see from a fork — workflow-level `test_report` being schema-rejected, and the missing `maestro_e2e_credentials` group failing validation for every workflow in the file — #7208 found by running builds, which is the better evidence.

Thanks @rabble for taking it across the line. The remaining hardening items live on in #7104 (artifact collection minus the 153 MB device-simulator.log) and the drift-guard work continues in #6978 / #7213.

Issue #6604, opened 2026-08-02
fix(invites): retry invite status after bunker signer reconnect


NotThatKindOfDrLiz:
## Summary

Track the remaining signer-readiness gap from #6510 for bunker-backed identities.

Bunker-backed identities currently report Nostr write readiness as available whenever the identity exists, but mid-session bunker relay disconnect/reconnect does not emit an auth-session readiness change. A bunker user can therefore hit a transient invite-status 401 and only recover manually.

## Acceptance criteria

- Bunker signing readiness reflects whether the remote signer can actually produce NIP-98 signatures.
- Invite status automatically retries after a bunker signer becomes ready again.
- The readiness signal is covered by a regression test.

NotThatKindOfDrLiz:
Obsolete — closing as part of retiring invite codes.

Production flipped `onboardingMode` to `open` on 2026-08-20, and both #8000 and #8183 call for deleting the invite machinery rather than maintaining it. This issue improves or repairs a code path that is being removed, so fixing it would be work thrown away.

If the decision to retire invites is reversed, reopen this — the finding itself is still accurate.

Tracked under epic #8297 (`epic(invites): retire the invite-code system`). Closed during the 2026-08-29 backlog sweep.

PR #6981, opened 2026-08-10
fix(invites): drive invite status from the nostr session readiness contract

### LINKED ISSUE #6977: fix(invites): invite status never sees Keycast RPC signer readiness
## Summary

Keycast users who sign over RPC (no local nsec) never see their invites. The
invites row is missing from Settings and the screen sits in its error state,
even when the server has invites allocated for them.

Reported by Alice Chan, who has 625 invites server side and sees none in the
app. Confirmed on 1.0.19 (833), which is commit d31fd9d and does contain the
#6510 fix (#6574).

## Root cause

`InviteStatusCubit` derives signer readiness from a single stream:

```dart
// mobile/lib/main.dart:2767
authSessionStream: authService.authStateStream.map((_) => currentAuthSession()),
```

`currentAuthSession()` reads `authService.canPublishNostrWritesNow`, which for a
Keycast identity is:

```dart
// mobile/lib/services/auth_service.dart:505
KeycastNostrIdentity(:final signsWithLocalKey) =>
    signsWithLocalKey || _authRpcCapability == AuthRpcCapability.rpcReady,
```

`_authRpcCapability` starts at `unavailable` and is published on
`authRpcCapabilityStream` (`auth_service.dart:493`), a different controller from
`authStateStream`. The cubit does not listen to it.

For a Keycast user with no local key the capability flips to `rpcReady` *after*
the auth state has already reached `authenticated`, so the last
`authStateStream` emission carries `isSignerReady: false` and nothing ever
corrects it.

## Evidence

From the reporter's log export, the only two relevant transitions in the whole
session:

```
06:25:56.705414  [AuthService] Auth state: authenticating -> authenticated
06:25:56.726403  [AuthService] RPC capability: unavailable -> rpcReady
```

21ms apart, and no further `Auth state:` line for the rest of the log. So the
cubit latched `isSignerReady: false` and stayed there.

Downstream, `load()` returns before touching the client:

```dart
// mobile/lib/blocs/invite_status/invite_status_cubit.dart:77
if (!state.isSignerReady) {
  _enterWaitingForAuth();
  return;
}
```

The log corroborates this at every layer. She opened Settings twice (06:26:02
and 06:26:31), and across the session there is:

- no NIP-98 token creation for `invite.divine.video`, though there are six for
  other hosts
- no `[InviteApiClient]` warning, so the auth header provider was never asked
- no invite API error of any kind

Meanwhile Keycast signing itself was healthy, with ten successful
`sign_event` RPC calls at HTTP 200.

The 12s `authWaitTimeout` had already fired by her first Settings visit, which
is why she sees the error state rather than a spinner.

## Blast radius

Every Keycast/divineOAuth account without a local nsec, on any build containing
#6574. Accounts with a local key have `signsWithLocalKey == true`, so readiness
is true immediately and they are unaffected. That is why this did not surface
during testing.

## Suggested fix

Merge `authRpcCapabilityStream` into the stream passed to the cubit so a
capability flip re-evaluates readiness and retriggers the load. The cubit
already handles a readiness change correctly once it hears about it, via
`_onAuthSessionChanged`.

## Acceptance criteria

- A Keycast RPC identity with no local key loads invite status without any
  manual retry.
- Readiness arriving after the auth state settles retriggers the load.
- Regression test covering a readiness flip that arrives only on the capability
  stream.

## Related

#6604 is the same class of bug for `BunkerNostrIdentity`. This one is a
different arm of the same readiness switch and affects a much larger group, so
it is worth fixing independently rather than folding in.

dcadenas:
## Description

Keycast users who sign over RPC with no local nsec never saw their invites. The Invites row was absent from Settings and the screen sat in its error state, even with invites allocated server side.

`InviteStatusCubit` was gated on `AuthService.canPublishNostrWritesNow`, sampled from `authStateStream`. Auth state only reports that the pubkey is known. For a Keycast identity whose key lives on the server, the signer becomes usable *after* that, and the app publishes that fact on a different signal. The cubit asked once at `authenticated`, got "cannot sign", and never heard the correction. It waited out its 12s bound and fell through to the error state.

From the reporter's log, the only two relevant transitions in the whole session:

```text
06:25:56.705414  [AuthService] Auth state: authenticating -> authenticated
06:25:56.726403  [AuthService] RPC capability: unavailable -> rpcReady
```

21ms apart, and no further `Auth state:` line for the rest of the log.

The app never called the invite API at all. Across the session there is no NIP-98 token creation for `invite.divine.video` though there are six for other hosts, no `[InviteApiClient]` warning, and no invite API error of any kind. Meanwhile Keycast signing itself was healthy, with ten successful `sign_event` calls at HTTP 200.

### The fix

Invite status uses NIP-98 over the REST invite API. Its auth header is created through `InviteApiClient -> nip98AuthServiceProvider -> Nip98AuthService(authService:)`, so it needs an authenticated identity that can sign now, but it does not need the app-wide `NostrClient` to finish relay initialization.

This adds `inviteStatusAuthSessionProvider`, which watches:

- `currentAuthStateProvider` for account/sign-out changes
- `currentAuthRpcCapabilityProvider` for the Keycast `unavailable -> rpcReady` transition

It then samples `AuthService.canPublishNostrWritesNow` and feeds the cubit through `identityStreamOf`, the existing seeded stream helper for Riverpod dependencies captured by app-shell `BlocProvider.create`.

The bloc layer is unchanged and stays free of Riverpod.

### Blast radius

Every Keycast/divineOAuth account with no local nsec, on any build containing #6574. Accounts with a local key have `signsWithLocalKey == true`, so readiness is true immediately, which is why this did not surface in testing.

### Documentation

Second and takeover commits. `.claude/rules/state_management.md` still pointed at `isNostrReadyProvider` for the readiness pattern, and that provider no longer exists anywhere in `lib`. Someone reading the rules for the readiness pattern got a name with zero hits, and raw auth state is the natural fallback.

The guidance now distinguishes two cases:

- Nostr-client signing/publishing waits on `nostrSessionProvider` reaching `nostrReady`.
- Relay-free signing, such as REST NIP-98 through `Nip98AuthService(authService:)`, watches signer capability (`currentAuthRpcCapabilityProvider`) and samples `canPublishNostrWritesNow`.

**Related Issue:** Closes #6977

## Out of Scope

- `canPublishNostrWritesNow` remains at four pull-style call sites (`repository_providers.dart:135`, `video_providers.dart:593,689`, and `video_sharing_service.dart:170`). They sample at the moment of use, so they cannot go stale in the way this bug did. Worth its own look; no failure verified.
- #6604 (bunker reconnect) is untouched. `BunkerNostrIdentity` reports readiness unconditionally and fails a different way.

## Verification

Unit/provider tests, analyzer, and before/after device checks.

- `flutter test test/providers/invite_status_auth_sessions_test.dart`
- `flutter test test/providers/invite_status_auth_sessions_test.dart test/blocs/invite_status`
- Pre-push hook: analyzer + changed-file provider test

On device, same account and same stack, only the build differing. The account is a Keycast email identity created through the Keycast API, so the key lives server side with no local nsec, which is the condition the bug needs.

**Before (origin/main, 1.0.19+819)**, reproducing the reporter's ordering:

```text
20:03:28.045  Auth state: authenticating -> authenticated
20:03:28.750  RPC capability: unavailable -> rpcReady
```

No Invites row anywhere in Settings, and zero invite API calls in the log.

**After**, cold restart matching the reporter's session-restore path:

```text
21:24:09.217  Auth state: authenticating -> authenticated
21:24:10.218  Creating new NIP-98 auth token for GET /v1/invite-status
21:24:10.662  RPC capability: unavailable -> rpcReady
```

Settings showed Invites 10, the screen loaded "9 invites ready to generate", and Generate produced a working code with the counter stepping down and the codes persisting across a reinstall.

Both auth paths were exercised on the final implementation: interactive sign-in and cold restart.

Not claimed: origin/main was reproduced through sign-in rather than cold restart. The cold-restart variant on origin/main is evidenced by the reporter's production log, which is that path and stops at `rpcReady` with no invite call.

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/providers/invite_status_auth_sessions.dart
+++ b/mobile/lib/providers/invite_status_auth_sessions.dart
@@ -1,0 +1,42 @@
+// ABOUTME: Bridges the app-wide Nostr readiness contract into invite status.
+// ABOUTME: Feeds InviteStatusCubit the signer-ready phase, not raw auth state.
+
+import 'dart:async';
+
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:openvine/blocs/invite_status/invite_status_cubit.dart';
+import 'package:openvine/providers/nostr_client_provider.dart';
+
+/// Maps a [NostrSessionReadiness] snapshot onto the cubit's session type.
+///
+/// `identityKnown` carries a pubkey but no usable signer, so it maps to a
+/// session that is addressed to an account yet not allowed to call the invite
+/// API. Only `nostrReady` reports a signer.
+InviteStatusAuthSession inviteStatusAuthSessionOf(
+  NostrSessionReadiness readiness,
+) => InviteStatusAuthSession(
+  accountId: readiness.pubkey,
+  isSignerReady: readiness.isReadyForActiveClient,
+);
+
+/// Session updates for [InviteStatusCubit].
+///
+/// Reads from [nostrSessionProvider] rather than `authStateStream`: auth state
+/// only says the identity is known, and a Keycast signer with no local key
+/// becomes usable *after* that (#6977). The cubit asks once and then waits for
+/// a push, so watching the wrong signal strands it forever.
+final inviteStatusAuthSessionsProvider =
+    Provider<Stream<InviteStatusAuthSession>>((ref) {
+      final controller = StreamController<InviteStatusAuthSession>.broadcast();
+      final subscription = ref.listen<NostrSessionReadiness>(
+        nostrSessionProvider,
+        (_, next) => controller.add(inviteStatusAuthSessionOf(next)),
+      );
+
+      ref.onDispose(() {
+        subscription.close();
+        unawaited(controller.close());
+      });
+
+      return controller.stream;
+    });
--- a/mobile/lib/providers/invite_status_auth_sessions.dart
+++ b/mobile/lib/providers/invite_status_auth_sessions.dart
@@ -1,42 +1,36 @@
-// ABOUTME: Bridges the app-wide Nostr readiness contract into invite status.
-// ABOUTME: Feeds InviteStatusCubit the signer-ready phase, not raw auth state.
-
-import 'dart:async';
+// ABOUTME: Bridges auth signer capability into invite status.
+// ABOUTME: Feeds InviteStatusCubit a push signal for late RPC readiness.
 
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:openvine/blocs/invite_status/invite_status_cubit.dart';
-import 'package:openvine/providers/nostr_client_provider.dart';
+import 'package:openvine/providers/auth_providers.dart';
+import 'package:openvine/providers/provider_identity_stream.dart';
 
-/// Maps a [NostrSessionReadiness] snapshot onto the cubit's session type.
+/// Current auth session for invite status.
 ///
-/// `identityKnown` carries a pubkey but no usable signer, so it maps to a
-/// session that is addressed to an account yet not allowed to call the invite
-/// API. Only `nostrReady` reports a signer.
-InviteStatusAuthSession inviteStatusAuthSessionOf(
-  NostrSessionReadiness readiness,
-) => InviteStatusAuthSession(
-  accountId: readiness.pubkey,
-  isSignerReady: readiness.isReadyForActiveClient,
-);
+/// Invite status uses NIP-98 over the REST invite API. That requires an
+/// authenticated identity that can sign now, but it does not require the
+/// app-wide `NostrClient` to finish relay initialization.
+final inviteStatusAuthSessionProvider = Provider<InviteStatusAuthSession>((
+  ref,
+) {
+  final authService = ref.watch(authServiceProvider);
+  ref.watch(currentAuthStateProvider);
+  ref.watch(currentAuthRpcCapabilityProvider);
+
+  return InviteStatusAuthSession(
+    accountId: authService.currentPublicKeyHex,
+    isSignerReady: authService.canPublishNostrWritesNow,
+  );
+});
 
 /// Session updates for [InviteStatusCubit].
 ///
-/// Reads from [nostrSessionProvider] rather than `authStateStream`: auth state
-/// only says the identity is known, and a Keycast signer with no local key
-/// becomes usable *after* that (#6977). The cubit asks once and then waits for
-/// a push, so watching the wrong signal strands it forever.
-final inviteStatusAuthSessionsProvider =
-    Provider<Stream<InviteStatusAuthSession>>((ref) {
-      final controller = StreamController<InviteStatusAuthSession>.broadcast();
-      final subscription = ref.listen<NostrSessionReadiness>(
-        nostrSessionProvider,
-        (_, next) => controller.add(inviteStatusAuthSessionOf(next)),
-      );
-
-      ref.onDispose(() {
-        subscription.close();
-        unawaited(controller.close());
-      });
-
-      return controller.stream;
-    });
+/// Reads from [inviteStatusAuthSessionProvider] rather than `authStateStream`
+/// alone: auth state only says the identity is known, and a Keycast signer with
+/// no local key becomes usable on the RPC-capability signal (#6977). The cubit
+/// asks once and then waits for a push, so watching only auth state strands it.
+final Provider<Stream<InviteStatusAuthSession>>
+inviteStatusAuthSessionsProvider = identityStreamOf<InviteStatusAuthSession>(
+  inviteStatusAuthSessionProvider,
+);
--- a/.claude/rules/self_review_checklist.md
+++ b/.claude/rules/self_review_checklist.md
@@ -137,11 +137,12 @@ State management:
   child via `BlocProvider`-keyed-on-identity (the existing rule
   then covers the parent). See
   [`state_management.md`](state_management.md#stateful-widgets-that-hold-the-captured-dep-bloc-directly).
-- [ ] Anything that signs, publishes, or sends a NIP-98 header waits on
-  `nostrSessionProvider` reaching `nostrReady`, not on
-  `authStateStream` / `currentAuthStateProvider` / `canPublishNostrWritesNow`.
-  Auth state only says the pubkey is known; a Keycast signer with no
-  local key arrives after it. See
+- [ ] Anything that signs or publishes through the app-wide `NostrClient`
+  waits on `nostrSessionProvider` reaching `nostrReady`; relay-free
+  signing such as REST NIP-98 watches a signer-capability signal such as
+  `currentAuthRpcCapabilityProvider` before sampling
+  `canPublishNostrWritesNow`. Auth state only says the pubkey is known;
+  a Keycast signer with no local key arrives after it. See
   [`state_management.md`](state_management.md#signing-readiness-identity-known-is-not-signer-ready).
 - [ ] No error strings / exception objects in BLoC `state`. Use status
   enums + `addError`. See [`state_management.md`](state_management.md).
--- a/.claude/rules/self_review_checklist.md
+++ b/.claude/rules/self_review_checklist.md
@@ -137,6 +137,12 @@ State management:
   child via `BlocProvider`-keyed-on-identity (the existing rule
   then covers the parent). See
   [`state_management.md`](state_management.md#stateful-widgets-that-hold-the-captured-dep-bloc-directly).
+- [ ] Anything that signs, publishes, or sends a NIP-98 header waits on
+  `nostrSessionProvider` reaching `nostrReady`, not on
+  `authStateStream` / `currentAuthStateProvider` / `canPublishNostrWritesNow`.
+  Auth state only says the pubkey is known; a Keycast signer with no
+  local key arrives after it. See
+  [`state_management.md`](state_management.md#signing-readiness-identity-known-is-not-signer-ready).
 - [ ] No error strings / exception objects in BLoC `state`. Use status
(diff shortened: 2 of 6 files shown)


mobile/lib/providers/invite_status_auth_sessions.dart
```diff
@@ -0,0 +1,42 @@
...
+/// Session updates for [InviteStatusCubit].
+///
+/// Reads from [nostrSessionProvider] rather than `authStateStream`: auth state
+/// only says the identity is known, and a Keycast signer with no local key
+/// becomes usable *after* that (#6977). The cubit asks once and then waits for
+/// a push, so watching the wrong signal strands it forever.
+final inviteStatusAuthSessionsProvider =
+    Provider<Stream<InviteStatusAuthSession>>((ref) {
+      final controller = StreamController<InviteStatusAuthSession>.broadcast();
+      final subscription = ref.listen<NostrSessionReadiness>(
```
realmeylisdev: `identityStreamOf` in `providers/provider_identity_stream.dart` is this exact bridge — a provider's values as a stream for a bloc built in `BlocProvider.create` — and it deliberately uses `ref.container.listen` rather than `ref.listen`. Its comment names the trap directly: a body-level `ref.listen` is only driven while *that* provider has listeners, and this one is only ever `ref.read`.

I measured it in this exact shape (real `nostrSessionProvider`, bridge read from a container, nothing listening to the bridge itself):

| how the source changed | events reaching the stream |
|---|---|
| `notifier.update(...)` | 1 |
| a `build()` recompute | **0** |
| a `build()` recompute, via `identityStreamOf` | 1 |
| a `build()` recompute, with one other live listener on the source | 1 |

So nothing is broken today, and the device run confirms it: `NostrService` only ever pushes through `notifier.update`, and ~20 other consumers hold live listens. But that means the wiring is safe because of two properties of the *source*, not because of anything here — and the new test only drives the row that works. Add a `ref.watch` to `NostrSession.build()` that can actually change, and invites go quiet again in the same way #6977 did, with nothing red to catch it.

`identityStreamOf` also seeds each subscriber (`Stream.multi` instead of a bare broadcast), which would remove the need for `initialAuthSession` and the stream to be two things that must agree.

Either reuse it — `identityStreamOf<NostrSessionReadiness>(nostrSessionProvider)` plus `.map(inviteStatusAuthSessionOf)` — or just switch this one call to `ref.container.listen`.

realmeylisdev (approved):
Approved. Reproduced #6977 and the fix on an iOS Simulator against the local stack — same Keycast (`divineOAuth`, no local nsec) account, same containers, only `main.dart` differing between the two runs:

| offset from `authenticated` | `origin/main` | this PR |
|---|---|---|
| `RPC capability -> rpcReady` | +312 ms | +309 ms |
| `[NostrService] Client initialized` | +373 ms | +377 ms |
| NIP-98 token for `GET /v1/invite-status` | **never** | **+379 ms** |

Zero invite calls across the whole `origin/main` session, one fired 2 ms after readiness here. Same shape as the reporter's log, so the diagnosis and the fix both hold.

Also checked: `flutter analyze lib test integration_test` clean, 347 tests green across `blocs/invite_status`, `screens/settings`, `services/auth` and the new provider test, and the doc edits survive a spot check (`profileRepositoryProvider` really does gate on `isReadyForActiveClient`).

Two notes, neither blocking:

1. One inline on the stream bridge — there's an existing helper for this exact position.
2. "`nostrReady` arrives before `rpcReady` flips" didn't hold on iOS: `rpcReady` came first in both of my runs. Doesn't change anything — the fix works either way and "broader" still stands — but "earlier" looks environment-specific rather than a property of the signal.

NotThatKindOfDrLiz:
@dcadenas I pushed a takeover commit (`b86c13d`) to address the remaining review items:

- replaced the hand-rolled `ref.listen` stream bridge with the existing `identityStreamOf` helper
- narrowed invite status readiness to auth state + `currentAuthRpcCapabilityProvider`, so REST NIP-98 invite calls do not wait on relay initialization
- updated the readiness docs/PR body to distinguish Nostr-client publishing from relay-free NIP-98 signing
- expanded provider coverage for late RPC readiness and provider recompute after a read-only stream bridge

Verification is green:

- `flutter test test/providers/invite_status_auth_sessions_test.dart`
- `flutter test test/providers/invite_status_auth_sessions_test.dart test/blocs/invite_status`
- GitHub CI: all checks passing on `b86c13d`

NotThatKindOfDrLiz (approved):
Approved after takeover follow-up.

The remaining stream-bridge issue is fixed with `identityStreamOf`, invite status now watches signer capability without adding a relay-initialization dependency, the PR body/docs match the final behavior, and CI is green on `b86c13d`.

NotThatKindOfDrLiz:
## Review: Three findings before merge

### 🔴 Blocker: Stream bridge duplicates `identityStreamOf`

**Location:** `mobile/lib/providers/invite_status_auth_sessions.dart`

The new `inviteStatusAuthSessionsProvider` reimplements the exact pattern `identityStreamOf` was designed to prevent. It uses `ref.listen` in a Provider body, but the caller (`BlocProvider.create` in `mobile/lib/main.dart:2754-2765`) only `ref.read`s — not `ref.watch`es — the provider. This reintroduces the silent failure mode described in `.claude/rules/state_management.md` and pinned by the test in `provider_identity_stream_test.dart`.

When `BlocProvider.create` calls `ref.read(inviteStatusAuthSessionsProvider)`, it gets a stale snapshot. Changes to `nostrSessionProvider` never notify the listener registered inside the provider body, because `ref.listen` installed handlers that depend on the provider remaining watched. The cubit receives `identityKnown` and never advances to `nostrReady`.

**Fix:** Use the existing helper instead of reimplementing:

```dart
// Delete invite_status_auth_sessions.dart entirely.
// In main.dart:2754-2765, replace:
inviteStatusAuthSessionOf(session),
// with:
inviteStatusAuthSessionOf(ref.watch(
  identityStreamOf(nostrSessionProvider)
      .map(inviteStatusAuthSessionOf)
)).first,
```

Or if the provider is needed elsewhere, reimplement using `ref.container.listen` + `Stream.multi` (copy the pattern from `provider_identity_stream.dart` exactly).

**Why this matters:** This is the unresolved inline thread realmeylisdev flagged. It's not a nit — it's the core failure mode the upstream rule was written to catch.

---

### 🟡 Significant: PR adds relay dependency where none existed

**Location:** `mobile/lib/main.dart:2754-2765`

**Before this PR:**  
`InviteStatusCubit` sampled `canPublishNostrWritesNow` from `authStateStream` — a capability check, not a readiness check. For local-key and Amber identities, this returned immediately (no relay needed).

**After this PR:**  
`InviteStatusCubit` waits for `nostrReady`, which requires full `NostrClient` initialization (~4s cold start, 90s timeout with backoff). This blocks invite fetch for non-Keycast identities unnecessarily, repeating the regressions from #5863 and #6423.

**Why it was added:** Keycast signers become NIP-98 capable ~21ms *after* `authenticated` on the RPC-capability signal. The original sampling `canPublishNostrWritesNow` once at `authenticated` time caught them in the "not yet ready" window.

**Alternative (narrower):** Gate on auth-state + sample `canPublishNostrWritesNow` with a listener that advances the cubit state only when true:

```dart
// In cubit:
void _onAuthStateChanged(AuthState authState, Emitter emit) async {
  if (authState.status == AuthStatus.authenticated) {
    // Don't gate on nostrReady. Instead:
    if (canPublishNostruWritesNow()) {
      emit(state.copyWith(session: ready));
      add(const FetchInvites());
    } else {
      // Listen for the capability to arrive
      emit(state.copyWith(session: waitingForSigner));
      ref.listen(
        currentAuthRpcCapabilityProvider,
        (prev, curr) async {
          if (curr.canPublish && state.session == waitingForSigner) {
            emit(state.copyWith(session: ready));
            add(const FetchInvites());
          }
        },
      );
    }
  }
}
```

This preserves the ~21ms Keycast fix without blocking local-key cold start.

---

### 🟠 Coverage gap: Test exercises only the working path

**Location:** `mobile/test/providers/invite_status_auth_sessions_test.dart`

The regression test drives the `notifier.update` code path (which works), never the `build()` recompute path where `ref.listen` silently drops.

**Add:**
```dart
testWidgets('updates session when nostr readiness rebuilds', (tester) async {
  final container = ProviderContainer();
  
  // Emit identity-known state
  container.read(nostrSessionProvider.notifier).state = NostrSessionReadiness.identityKnown(...);
  expect(
    container.read(inviteStatusAuthSessionsProvider),
    equals(InviteStatusAuthSession.identityKnown),
  );
  
  // Rebuild provider (simulating upstream rebuild)
  // This would expose the ref.listen bug if it existed
  container.invalidate(someUpstreamDep);
  
  // Advance to nostrReady
  container.read(nostrSessionProvider.notifier).state = NostrSessionReadiness.nostrReady(...);
  expect(
    container.read(inviteStatusAuthSessionsProvider),
    equals(InviteStatusAuthSession.nostrReady),
  );
});
```

---

## Recommendation

**Don't merge on current approval.** The unresolved inline thread is a real duplication finding, not a nit.

**Fix priority:**
1. (Blocker) Use `identityStreamOf` helper or reimplement the listener pattern correctly.
2. (Significant) Evaluate whether the relay-dependency trade-off is worth the Keycast fix, or narrow to RPC-capability sampling instead.
3. (Coverage) Add test for the `build()` recompute path.

After fixes, the PR will be ready for merge.

NotThatKindOfDrLiz (changes requested):
Review complete with three findings above. The blocker (stream bridge duplication) and significant finding (relay dependency regression) require fixes before merge.

Issue #3655, opened 2026-04-29
Set up automated dependency monitoring


omartinma:
**Problem**: No Dependabot or CI job monitors dependency freshness. Updates rely entirely on manual checks.

**Evidence**: No `.github/dependabot.yml` configuration exists. No CI workflow runs `flutter pub outdated` or alerts on available updates. Dependency updates happen only when a developer manually checks or when a build breaks due to a transitive conflict.

**Impact**: Medium. Without automated monitoring, dependencies drift silently. Security advisories on transitive dependencies go unnoticed until they cause build failures or are discovered in manual audits (like this one).

**Effort**: Low. Add a Dependabot configuration for the root `pubspec.yaml` and key packages. Alternatively, add a scheduled CI job that runs `flutter pub outdated` and posts results to a Slack channel or creates an issue.

**Source**: [Code Review — Dependencies](docs/code-review/issues-dependencies.md)

NotThatKindOfDrLiz:
PR #6923 closes the local override documentation piece of #3363, but #3363 AC3 is more specific than this issue's current dependency-freshness wording: CI should reject mutable git refs and unexpected path overrides in pubspec.yaml/pubspec.lock, with an allowlist for intentional overrides. Please keep that guardrail in scope here or split it to a dedicated CI issue before #3363 is closed.

realmeylisdev:
Keeping the AC3 guardrail in scope here rather than splitting it — PR #6990 carries both, in separate commits.

On the guardrail: the detector keys off `ref` vs `resolved-ref` in `pubspec.lock`. Those diverge exactly when a git ref is movable, so it catches the original finding — before #6167 the `c2pa_flutter` entry read `ref: "0.0.3"` against `resolved-ref: "847d0c…"`. That lockfile is the test fixture. It reads the pubspecs too, since version-only overrides like `device_info_plus` resolve to `source: hosted` and leave no lockfile trace. A movable ref fails even if baselined — the allowlist covers which dependencies may be non-pub, not whether a ref may move. Current allowlist is five entries, each annotated with its tracking issue (#6166, #6918, #3638).

Worth flagging that it passes on today's tree: the mutable `media_kit_video` branch is gone and `pooled_video_player` was deleted, so this freezes the win rather than fixing anything outstanding.

[...]

PR #6167, opened 2026-07-18
chore(deps): migrate media_kit_video to pub and harden the c2pa_flutter git pin

realmeylisdev:
## Description

Resolves the two git-pinned dependencies flagged in #3637. They needed opposite remediations, so this is one PR with two clearly-delineated commits.

**1. `media_kit_video` → pub `^2.0.1` (removes the git pin).**
The `divinevideo/media-kit` fork diverged from pub `media_kit_video` 2.0.1 by a single darwin-only patch — an iOS-Simulator software-render rotation crash workaround (upstream media-kit/media-kit#627). Divine routes `media_kit` only on Linux (`kIsWeb || TargetPlatform.linux` in `divine_video_player`); iOS/macOS use native AVFoundation, Android ExoPlayer, Web `HtmlVideoElement`. That darwin path is unreachable on every platform we ship, so the fork carried dead code for us. Pub 2.0.1's Dart/pubspec surface is identical to the fork, so the swap is dependency-graph and behaviour neutral.

**2. `c2pa_flutter` → documented + hardened (stays git-pinned).**
It can't move to pub: the guardianproject plugin we use isn't on pub.dev, and pub.dev's same-named `c2pa_flutter` is an unrelated third-party rewrite (different author, `flutter_rust_bridge` FFI, c2pa-rs 0.75.x, no `RemoteSigner`) our signing services can't consume. `0.0.3` is guardianproject's latest tag. This pins the immutable commit (was the movable `0.0.3` tag) so a moved/force-pushed tag can't silently change the build, and adds a comment explaining the pin. The resolved commit is unchanged.

The systemic "git pins bypass pub.dev scanning" concern is tracked separately in #3655.

**Related Issue:** Closes #3637

## Out of Scope

- De-git-pinning `c2pa_flutter` — blocked upstream; tracked in #6166 (revert to pub if guardianproject publishes).
- Automated git-pin monitoring — #3655.

## Verification

- `flutter pub get` — clean; `media_kit_video` flips git→hosted (pub 2.0.1), `c2pa_flutter` `resolved-ref` unchanged (immutable-SHA relabel only).
- `flutter analyze lib test integration_test` — No issues found.
- `flutter analyze` + `flutter test` in `divine_video_player` — 250 tests pass.
- `flutter test` for c2pa signing / identity-manifest / native-proofmode services — 18 tests pass.

## Type of Change

- [x] ✅ Build configuration change
- [x] 🗑️ Chore

PR #6990, opened 2026-08-10
ci(deps): monitor dependency freshness and freeze dependency provenance

realmeylisdev:
Refs #3655

## Summary

Covers the GitHub Actions monitoring slice of #3655, and implements #3363 **AC3** (the guardrail the review comment on #3655 asked to keep in scope here). Two deliverables, separate commits, one PR — they are interdependent and the repo forbids stacking.

### 1. Dependabot for GitHub Actions

Only **1 of 96** `uses:` references in `.github/workflows` is SHA-pinned; every third-party action floats on a mutable major tag, so an action can change under CI with no diff here. Weekly grouped PRs keep those current.

Limits are tuned to this repo rather than copied: ~106 PRs merge per week and `main` requires branches to be up to date, so `open-pull-requests-limit: 3` keeps rebase and CI cost bounded. `assignees` rather than `reviewers` — the latter was removed from `dependabot.yml`, and `mobile/docs/CODEOWNERS` is not a path GitHub reads. The durable CODEOWNERS/ownership cleanup is tracked in #6994.

### 2. Dependency-provenance guard (#3363 AC3)

CI now rejects mutable git refs and unexpected path overrides, with an allowlist for the intentional ones.

The detector keys off **`ref` vs `resolved-ref`** in `pubspec.lock`. pub writes both for a git source and they diverge exactly when the ref is movable. This repo's own history is the proof — before #6167:

```
ref:          "0.0.3"                                     <- movable tag
resolved-ref: "847d0c0183c9c09946ce321c107decd3fd9f56b2"  <- diverged
```

which is the finding #3363 was filed for. The test feeds that exact shape.

Both files must be read: only the lockfile carries `resolved-ref`, and only the pubspecs show version-only `dependency_overrides` (those resolve to `source: hosted` and leave no lockfile trace).

**A movable ref fails unconditionally, even if baselined.** The allowlist sanctions *which* dependencies may come from a non-pub source; never that a ref may move under the build.

Baseline is the five sanctioned entries today, each with its tracking issue:

```
git:c2pa_flutter:https://github.com/guardianproject/c2pa-flutter.git:PINNED # defer: #6166
override:.../nostr_sdk/pubspec.yaml:meta               # noise: transitive pin
override:mobile/pubspec.yaml:app_device_integrity      # defer: #6918
override:mobile/pubspec.yaml:device_info_plus          # defer: #3638
path:app_device_integrity:overrides/...-1.1.0          # defer: #6918
```

It reads the lockfile and git-tracked pubspecs rather than the filesystem, because `mobile/overrides/` still holds untracked leftovers from forks removed in #6923 that a directory check would flag forever. Matching on `source:` rather than grepping `path:` avoids eight false positives — `path` is also a hosted package name and a git subdirectory key.

Runs in the existing `generated-files` job; no new workflow.

## Deliberately not in this PR

- **The `pub` ecosystem.** Three open dependabot-core defects, one of which matches this repo exactly:
  - **#8356** — no PRs created when `dependency_overrides` constrain resolution. `mobile/pubspec.yaml` has precisely that shape, so the ecosystem could silently produce nothing.
  - **#15654** — unmerged fix for Dependabot rewriting the lockfile's Flutter constraint (`3.44.0` → `>=3.44.0`), which would dirty `pubspec.lock` every run and fight `Generated Files`.
  - **#11860** — the Flutter checkout pub updates depend on fails on `bin/internal/engine.version`.

  pub is community-maintained, so these carry no GitHub SLA. `.github/dependabot.yml` documents the deferral and the exact config to add once they land. #3655 stays open for pub dependency monitoring — note `directory` must be `/mobile`, the workspace root, because pub's `dependency_services` refuses to run anywhere else.

- **SHA-pinning the 10 floating action tags.** A policy call, and the natural follow-up once Dependabot is keeping pins fresh.

## Needs a maintainer — not fixable by PR

**Dependabot alerts are disabled on this repo** (`GET /dependabot/alerts` → 403). That is the security half of #3655's stated impact, and it is a repo setting. The `PUB` ecosystem *is* in the GitHub Advisory Database, and four of our packages have historical advisories (`archive`, `dio`, `http`, `shared_preferences_android`) — all currently on patched versions, so nothing is exposed today, but a future advisory would land silently.

Worth naming a triage owner first: `divine-resurrection` carries 56 open alerts and keycast#86 exists specifically to stop alerts becoming unowned noise.

## Verification

- `flutter test test/tools/` — 125 passed, 12 skipped (13 dependency-provenance tests)
- `dart format` — clean
- Guard against current `main` → exit 0
- Guard against the pre-#6167 lockfile → exit 1, flags `git:c2pa_flutter:https://github.com/guardianproject/c2pa-flutter.git:MUTABLE`
- Both YAML files parse

No app code changed; no user-visible behaviour, so no screenshots.


NotThatKindOfDrLiz:
Takeover changes pushed in `70f21127b`.

What changed:
- Included git repository URLs in dependency-provenance keys, so a pinned git dependency cannot be silently repointed to another repo.
- Made the provenance detector fail hard before ratchet/update mode when the lockfile or a listed pubspec is missing.
- Added detection for hosted packages from non-pub.dev registries.
- Replaced the dangling `tasks/plan_3655.md` reference with the inline pub Dependabot config.
- Changed the PR body from `Closes #3655` to `Refs #3655`; #3655 now stays open for pub dependency monitoring. CODEOWNERS ownership cleanup is tracked in #6994.

Verification:
- `bash scripts/check_dependency_provenance.sh`
- `flutter test test/tools/dependency_provenance_detector_test.dart`
- `flutter test test/tools/`
- `bash -n scripts/check_dependency_provenance.sh`
- GitHub checks are green on `70f21127b`.

@realmeylisdev please review the takeover commit before merging. The PR is ready from my side.

NotThatKindOfDrLiz (approved):
Takeover changes pushed in `70f21127b`.

What changed:
- Included git repository URLs in dependency-provenance keys, so a pinned git dependency cannot be silently repointed to another repo.
- Made the provenance detector fail hard before ratchet/update mode when the lockfile or a listed pubspec is missing.
- Added detection for hosted packages from non-pub.dev registries.
- Replaced the dangling `tasks/plan_3655.md` reference with the inline pub Dependabot config.
- Changed the PR body from `Closes #3655` to `Refs #3655`; #3655 now stays open for pub dependency monitoring. CODEOWNERS ownership cleanup is tracked in #6994.

Verification:
- `bash scripts/check_dependency_provenance.sh`
- `flutter test test/tools/dependency_provenance_detector_test.dart`
- `flutter test test/tools/`
- `bash -n scripts/check_dependency_provenance.sh`
- GitHub checks are green on `70f21127b`.

@realmeylisdev please review the takeover commit before merging. The PR is ready from my side.

PR #6993, opened 2026-08-10
fix(profile): explain profile badges

NotThatKindOfDrLiz:
## Summary
- make OG Viner and special profile checkmark badges tappable with accurate explanation dialogs
- add button semantics for both badges and remove the dead OG Viner semantic l10n key
- add localized fallback copy and focused widget coverage for badge explanations

## Verification
- flutter test test/widgets/user_name_og_viner_badge_test.dart test/widgets/special_profile_checkmark_test.dart test/widgets/user_name_nip05_test.dart test/l10n/arb_consistency_test.dart
- flutter analyze
- pre-push hook analyzer and changed-file widget tests

Refs #6982


hm21 (changes requested):
Reviewed and pushed two fixes as separate commits. All 15 checks green on `da1495e`.

**`0952b54` — the explainer didn't pause the feed video**

Both badges also render in `video_feed_item.dart`, over a playing video. `showProfileBadgeExplanationDialog` called `VineBottomSheet.show` directly, so `OverlayVisibility.setBottomSheetOpen` never flipped — and `activeVideoProvider` is gated on `hasVisibleOverlay`, so the video kept playing with audio behind the sheet. The direct call also defaults `useRootNavigator` to false, which opened the sheet inside the shell navigator, underneath the tab bar.

Switched to `context.showVideoPausingVineBottomSheet`, which every other feed-triggered sheet already uses (share, comments, report, collaborators, metadata). The one exception, `MetadataVerificationInfoSheet`, documents why it deviates. Added `profile_badge_explanation_dialog_test.dart` pinning the overlay flag across open/close for both badge types; it fails on the previous code path.

**`da1495e` — the semantic label dropped the badge's meaning**

The badges announced "Explain OG Viner badge" / "Explain profile checkmark". A sighted user gets two facts from the badge — what it means and that it's tappable — but a screen reader user only got the affordance. Moved the affordance into the semantics `hint` so the `label` carries the meaning again.

Reusing `ogVinerBadgeLabel` for the label also un-orphans it. The PR removed the already-dead `ogVinerBadgeSemanticLabel`, but replacing `ogVinerBadgeLabel`'s only call site left *it* dead across all 22 ARB files, so the dead-key count came out net zero.

Also in that commit, since it's all the same ARB churn: the two `*SemanticLabel` keys are renamed to `*SemanticHint` to match what they now are; `profileBadgeExplanationClose` is dropped in favour of the existing, already-translated `commonClose`; and the remaining new copy is translated into all 21 non-English locales — the sibling `profileBadge*` keys directly above it are already translated, so English-only values stood out.

**Open — your call: the badges are now 14 dp interactive targets**

`OgVinerBadge(size: 14)` and `SpecialProfileCheckmark(iconSize: 10, padding: 2)` both render a 14×14 dp box, against the 48×48 minimum in `.claude/rules/accessibility.md`. In the feed they sit inside the `onTap: navigateToProfile` area, so an imprecise tap aimed at the profile now opens a sheet instead.

I prototyped the 48 dp target and backed it out — measured on a 390 pt viewport, it takes the name row from 18 dp to 48 dp and the badge cluster from 36 dp to 96 dp wide. Both badges reach ~10 surfaces through `UserName`, several of them dense caption overlays on grid thumbnails (`composable_video_grid.dart`, `classic_vines_tab.dart`, `list_card.dart`, `comment_item.dart`), where that is a clear regression. There's no cheaper version: `RenderBox.hitTest` gates on the box's own `size`, so a hit area can't extend past the layout box — Material's own `MaterialTapTargetSize.padded` grows layout too.

So this needs a design decision rather than a reviewer patch. Options as I see them: accept the deviation and document it on the widgets; move the explanation to a comfortably-sized target on the profile screen and leave the inline badges non-interactive; or take the 48 dp and the layout change everywhere. Happy to implement whichever you pick.

Nothing else blocking from me — the two commits above are independent, so revert either if you disagree.

NotThatKindOfDrLiz:
@hm21 please re-review. I moved the compact badge explanations out of the inline 14 dp badges and into 48 dp profile-header targets, left inline badges non-interactive so they no longer steal parent profile-row taps, renamed the explainer from dialog to sheet, and removed the duplicate OG title l10n key.\n\nVerification run locally on rebased head 68351be7307e605b2d39f7ef7cf70505af4f60d8:\n- flutter analyze\n- flutter test test/widgets/profile_badge_explanation_sheet_test.dart test/widgets/special_profile_checkmark_test.dart test/widgets/user_name_og_viner_badge_test.dart test/widgets/profile/profile_header_widget_test.dart test/widgets/video_feed_item/video_author_info_section_test.dart\n\nThe pre-push hook also passed changed-file analyze/tests before pushing.

NotThatKindOfDrLiz:
@hm21 final pushed head is 19387545aacafe3c466ad5b76572efb6a775d886. The PR is now rebased onto current origin/main and all GitHub checks are green, including Generated Files, Analyze, Format, Mobile CI, and Tests shards 0-3/4. The earlier generated-files failure was from the stale main baseline before the final rebase.

hm21 (approved):
LGTM. Moving the explainers onto `DivineIconButton(size: small)` is the right call — that component carries a real 48 px tap target by construction, so it solves the sizing problem through the design system instead of hand-rolled padding, and the inline badges keep their meaning without stealing parent taps.

Pushed two things on top, both green:

- Rebased onto `origin/main`. The red `Generated Files` was not your diff: `#1d3c77da` landed on main, migrated `subtitle_editor_screen.dart` off Material buttons and shrank the material-button baseline. Your branch still carried the older tree, so the shrink-only ratchet read it as re-adding a key. Nothing in the diff needed changing.
- `19387545a` drops `profileBadgeOgVinerSemanticHint` and `profileBadgeCheckmarkSemanticHint` — the move to header buttons left both without call sites.

All 16 checks pass. Yours to merge.

PR #6995, opened 2026-08-10
test(relay-discovery): stop the invalid-signature test degenerating

### LINKED ISSUE #6996: test(relay-discovery): stop the invalid-signature test degenerating
Auto-created from PR #6995.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6995
**Branch:** `fix/relay-discovery-sig-flake`

realmeylisdev:
Closes #6996

## Summary

`test/services/relay_discovery_service_test.dart` fails on roughly 1 in 256 shard runs, for no reason connected to the branch it runs on. Most recently it took down `Tests (shard 0/4)` on #6711.

The test corrupts a signature by overwriting its first byte with the literal `00`:

```dart
badSignature['sig'] = '00${(badSignature['sig'] as String).substring(2)}';
```

`Event.sign` (`mobile/packages/nostr_sdk/lib/event.dart:122`) draws BIP-340 auxiliary randomness from `getRandomHexString()`, which is `Random.secure()` (`nostr_sdk/lib/client_utils/keys.dart:26`). The signature is therefore different on every run — and when it already begins with `00`, the overwrite is a **no-op**. The frame stays genuinely valid, `_authenticRelayList` correctly accepts it, and the `isEmpty` assertion fails:

```
Expected: empty
  Actual: [DiscoveredRelay(url: wss://legit.example, read: true, write: true)]
```

The reported relay is the frame's own `r` tag, which makes the failure look like cross-test leakage from the positive control. It isn't — the rejection path is working correctly on an input that was never actually corrupted.

## Measurement

Signing 4000 events with the test's own key and counting the signatures that begin with `00`:

```
PROBE: 16/4000 signatures start with 00 (expected ~15.625)
```

0.40%, against the 0.39% the encoding predicts. That is the failure rate.

## Why only this one test

The four sibling rejection tests mutate in ways that cannot degenerate — empty `sig`, a different signing key, tampered `tags` (breaking the id), a different `kind`. Only the constant-prefix overwrite has a fixed point.

## The fix

Flip the first byte instead of overwriting it. `x ^ 0xff` never equals `x`, so the signature is always different from the valid one.

## Testing

- `flutter test test/services/relay_discovery_service_test.dart` — 35 passing, run 8×
- `dart format` — clean

Test-only change; no production code touched.

### DIFF
--- a/mobile/test/services/relay_discovery_service_test.dart
+++ b/mobile/test/services/relay_discovery_service_test.dart
@@ -487,7 +487,13 @@ void _authenticityTests() {
 
     test('rejects a frame with an invalid signature', () async {
       final badSignature = signedRelayList(privateKey: victimKey);
-      badSignature['sig'] = '00${(badSignature['sig'] as String).substring(2)}';
+      final sig = badSignature['sig'] as String;
+      // BIP-340 signing draws fresh aux randomness every run, so overwriting
+      // the first byte with a constant '00' is a no-op on the 1-in-256
+      // signature that already starts with '00' — leaving the frame valid.
+      final flipped = int.parse(sig.substring(0, 2), radix: 16) ^ 0xff;
+      badSignature['sig'] =
+          '${flipped.toRadixString(16).padLeft(2, '0')}${sig.substring(2)}';
       expect(await query(badSignature), isEmpty);
     });
 


hm21 (approved):
LGTM. Root cause and fix both check out — I reproduced the degeneracy locally rather than taking the writeup on trust.

Probe over 4000 signings with the test's own key: 12 signatures started with `00`, and *exactly* those 12 still verified after the old `'00' + substring(2)` overwrite (`oldStillValid=12`). The XOR flip left 0 of 4000 verifying. `Event.sign` (`nostr_sdk/lib/event.dart:121`) does draw fresh aux from `Random.secure()`, so the per-run variance is real. 35/35 pass locally, CI green.

Also checked the sibling corruption sites — `relay_pool_verify_dedup_test.dart` (all-zero constant) and `nip98_relay_acceptance_test.dart` (conditional last-char swap) can't degenerate, so this was the only instance.

One nit for next time: `172ae76` is a `Merge branch 'main'` commit — AGENTS.md says rebase, never merge into a feature branch. Squash-merge erases it here so leave it, but the branch is `BEHIND` again anyway, so the merge didn't buy anything.

realmeylisdev:
Cleaned it up rather than leaving it: rebased onto `origin/main`, so both `Merge branch 'main'` commits are gone and the branch is now a single commit on top of `d395e7128`. That also clears the `BEHIND` you noted.

Tree hash is identical before and after the rebase (`797975ed9b4b3ff8e8f2c5bc976e0ba42bbe1f92`), so the content you approved is byte-for-byte unchanged — only the history moved. Verified the test text on the pushed ref rather than trusting the diff.

Approval survived the force-push, 15/15 green on `305edda32`. Locally: 35/35, plus 10 repeat runs of the signature test against fresh aux randomness.

PR #6997, opened 2026-08-10
fix(likes): chunk the reaction REQ filters like the deletion probe

### LINKED ISSUE #6998: fix(likes): chunk the reaction REQ filters like the deletion probe
Auto-created from PR #6997.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6997
**Branch:** `fix/chunk-reaction-req-filters`

realmeylisdev:
Closes #6998

Salvaged from #6711, which was closed. This part stands on its own and has nothing to do with revision resolution.

## Summary

#5751 chunked the Kind 5 deletion `#e` filter at 500 ids so a REQ frame could not approach the relay's `max_message_length`. The Kind 7 **reaction** filters sitting next to it were left unbounded at the query boundary. There are three event-id legs, plus two coordinate legs:

- `getLikeCounts` -> `Filter(kinds: [reaction], e: uncachedIds)` over the whole uncached batch
- the addressable like branch -> `_fetchResolvedLikersByEvent`, carrying both an `#e` filter over target ids and an `#a` filter over coordinates
- `_queryVoteTargetReactions` -> reached from `getVoteCounts` / `getUserVoteStatuses` via `comment_reactions_bloc.dart:144`, carrying both an `#e` filter over the target ids and an `#a` filter over their coordinates

Review round two closed the last two gaps: `_resolveRemoteDownvoteRecord`'s cold-cache Kind 5 probe was still building its `#e` filter directly, and every `#a` leg was bounded by *count* but not by *bytes* — a coordinate embeds a `d` tag whose length nothing validates. All six legs now route through the same chunker, and every chunk closes at 500 values **or** 128KB of tag-value bytes, whichever comes first.

## Why this is worth closing, stated honestly

**No oversized frame has been observed, and I'm not claiming one.** What makes it worth fixing is that today's safety is incidental rather than enforced.

The largest caller is the curation sort (`curation_repository.dart:135`), which passes `_videoEventCache.discoveryVideos` - every discovery video, with no cap of its own. That list happens to be capped at 500 by a memory guard in a completely different file (`video_event_service.dart:2973`, `:3204` - `if (list.length > 500) list.removeRange(500, list.length)`).

So the worst case today lands **exactly on** the 500 that the deletion probe already treats as the largest safe frame - zero headroom - and the bound lives in a service that has no idea it is load-bearing for a relay frame size. Raise that memory cap, or add a caller that doesn't go through it, and the filter silently goes over.

The coordinate legs matter at least as much as their `#e` siblings: `|coordinates| <= |eventIds|`, but a `kind:pubkey:dtag` coordinate costs more bytes per entry than a bare 64-char id, so that leg reaches the cap sooner. And unlike an id, a coordinate has no fixed width - nothing in the models layer validates `d` tag length, so the count cap alone cannot byte-bound the `#a` legs. That is what the byte budget enforces.

The vote path is the same argument one layer over: it is bounded at 50 today only by `_pageSize` in `comments_list_bloc.dart:61`, a comment-pagination constant with no relationship to frame size.

The remaining caller is safely bounded independently: `video_event_service` batches like-count requests at `_likeCountBatchMaxSize = 50`.

Worth knowing about the failure mode: an oversized frame errors the socket and `queryEvents` **fails open**. It doesn't throw - the counts just come back 0 and the liker list comes back empty, which reads as "nobody liked this" rather than as an error.

## Changes

- `_deletionReqEidChunkSize` -> `_reqEidChunkSize`, now shared by both kinds and by `#a` as well as `#e`; doc rewritten to match.
- `_reqTagValueChunkBytes = 128KB`: chunks also close on cumulative UTF-8 bytes of tag values. Never splits a typical batch (500 real coordinates is ~68KB, and the cap can never bind for 64-char ids) and keeps 4x headroom under `max_message_length` for oversized `d` tags. A single value over the budget ships alone - that REQ cannot be made any smaller.
- Chunk-and-dedup mechanics live in one place, `_queryChunked(values, buildFilter)`. `_queryEventsByEIds` is the `#e` wrapper over it, `_queryReactionsByEventIds` the Kind 7 wrapper over that.
- `_deletedReactionIds`, both `_fetchResolvedLikersByEvent` legs, both `_queryVoteTargetReactions` legs, and `_resolveRemoteDownvoteRecord`'s cold-cache deletion probe go through the same path.
- Chunked results are deduplicated by event id. Each chunk is its own `queryEvents` call with its own `EventMemBox`, so a reaction that `#e`-tags targets landing in two different chunks comes back from both, and `getLikeCounts` counts per e-tag occurrence - a 600-id batch with one reaction tagging `event_id_0` and `event_id_500` scored 2/2 instead of 1/1 without this. The single-chunk path is untouched.
- Every chunk stays a **single-filter** REQ, which `NostrClient` requires to keep the local event cache active.

Net: **+395 / -62** across 2 files. No behavior change for any input under 500 ids / 128KB of tag-value bytes.

## Testing

Seven chunking tests: one per guarded query leg - the two like-path filters, the Kind 5 deletion sweep, the vote path's `#e` and `#a` legs, the cold-cache downvote deletion probe - plus the byte bound. All mutation-verified rather than just observed green, in both directions:

- `_reqEidChunkSize = 100000` turns the **six count guards** red together.
- `_reqTagValueChunkBytes = 1 << 60` turns the **byte-bound test** red alone (the count guards stay green, pinning that the byte cap never binds for fixed-width ids).

The deletion-sweep guard is the pre-existing #5751 test - it going red under the same mutation confirms the refactor kept the deletion probe wired to the same constant. Reverting the dedup turns the double-count test red independently.

- `flutter test` in `packages/likes_repository`: **228 passing**
- pre-push changed-file test: `packages/likes_repository/test/src/likes_repository_test.dart` **168 passing**
- coverage **97.15%** (gate: 62)
- `flutter analyze lib test`: clean

### DIFF
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -5,7 +5,7 @@
 // ABOUTME: Supports offline queuing via callback injection.
 
 import 'dart:async';
-import 'dart:math';
+import 'dart:convert';
 
 import 'package:likes_repository/src/blocked_liker_filter.dart';
 import 'package:likes_repository/src/exceptions.dart';
@@ -34,9 +34,19 @@ const _downvoteContent = '-';
 /// (512KB). An oversized frame errors the socket and `queryEvents` fails
 /// open — silently counting deleted likes, or returning no reactions at all.
 /// 500 ids is ~34KB, well under 512KB and the advertised `max_event_tags`
-/// of 2000. A `kind:pubkey:dtag` coordinate costs more per entry than a bare
-/// id, but 500 of them still lands far inside the cap. See #5751.
+/// of 2000. See #5751.
 const _reqEidChunkSize = 500;
+
+/// Max cumulative UTF-8 bytes of tag values per `#e` or `#a` REQ filter.
+///
+/// The count cap only byte-bounds fixed-width values like event ids. A
+/// `kind:pubkey:dtag` coordinate is typically ~135 bytes, but nothing
+/// validates `d` tag length, so 500 coordinates minted by another client
+/// can still blow past `max_message_length`. 128KB of values keeps 4x
+/// headroom under the 512KB cap and never splits a typical batch. A single
+/// value over the budget still ships, alone — that REQ cannot be made any
+/// smaller.
+const int _reqTagValueChunkBytes = 128 * 1024;
 
 /// Callback to check if the device is currently online
 typedef IsOnlineCallback = bool Function();
@@ -1961,8 +1971,9 @@ class LikesRepository {
     );
   }
 
-  /// Runs [buildFilter] over [values] in chunks of [_reqEidChunkSize], so no
-  /// single REQ frame can exceed the relay's message cap.
+  /// Runs [buildFilter] over [values] in chunks capped at [_reqEidChunkSize]
+  /// values and [_reqTagValueChunkBytes] bytes, so no single REQ frame can
+  /// exceed the relay's message cap.
   ///
   /// Deduplicates by event id once there is more than one chunk. Each chunk
   /// is its own `queryEvents` call with its own dedup box, so an event that
@@ -1988,10 +1999,25 @@ class LikesRepository {
   }
 
   List<List<String>> _chunkStrings(List<String> values, int chunkSize) {
-    return [
-      for (var i = 0; i < values.length; i += chunkSize)
-        values.sublist(i, min(i + chunkSize, values.length)),
-    ];
+    final chunks = <List<String>>[];
+    var chunk = <String>[];
+    var chunkBytes = 0;
+    for (final value in values) {
+      final valueBytes = utf8.encode(value).length;
+      final closesChunk =
+          chunk.length >= chunkSize ||
+          (chunk.isNotEmpty &&
+              chunkBytes + valueBytes > _reqTagValueChunkBytes);
+      if (closesChunk) {
+        chunks.add(chunk);
+        chunk = <String>[];
+        chunkBytes = 0;
+      }
+      chunk.add(value);
+      chunkBytes += valueBytes;
+    }
+    if (chunk.isNotEmpty) chunks.add(chunk);
+    return chunks;
   }
 
   Set<String> _reactionTargetEventIds(
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -27,14 +27,15 @@ const _likeContent = '+';
 /// NIP-25 reaction content for a downvote.
 const _downvoteContent = '-';
 
-/// Max number of event ids per `#e` REQ filter.
+/// Max number of tag values per `#e` or `#a` REQ filter.
 ///
 /// A 64-char id costs ~67 bytes in JSON, so a single `#e` filter with enough
 /// of them builds a REQ frame near/over the relay's `max_message_length`
 /// (512KB). An oversized frame errors the socket and `queryEvents` fails
 /// open — silently counting deleted likes, or returning no reactions at all.
 /// 500 ids is ~34KB, well under 512KB and the advertised `max_event_tags`
-/// of 2000. See #5751.
+/// of 2000. A `kind:pubkey:dtag` coordinate costs more per entry than a bare
+/// id, but 500 of them still lands far inside the cap. See #5751.
 const _reqEidChunkSize = 500;
 
 /// Callback to check if the device is currently online
@@ -1322,18 +1323,16 @@ class LikesRepository {
     required Map<String, String> eventIdByCoordinate,
     List<String>? authors,
   }) async {
-    final eEventsFuture = _nostrClient.queryEvents([
-      Filter(kinds: const [EventKind.reaction], authors: authors, e: eventIds),
-    ]);
-    final aEventsFuture = eventIdByCoordinate.isEmpty
-        ? Future<List<Event>>.value(const <Event>[])
-        : _nostrClient.queryEvents([
-            Filter(
-              kinds: const [EventKind.reaction],
-              authors: authors,
-              a: eventIdByCoordinate.keys.toList(),
-            ),
-          ]);
+    final eEventsFuture = _queryChunked(
+      eventIds,
+      (chunk) =>
+          Filter(kinds: const [EventKind.reaction], authors: authors, e: chunk),
+    );
+    final aEventsFuture = _queryChunked(
+      eventIdByCoordinate.keys.toList(),
+      (chunk) =>
+          Filter(kinds: const [EventKind.reaction], authors: authors, a: chunk),
+    );
     final results = await Future.wait([eEventsFuture, aEventsFuture]);
 
     final eventsById = <String, Event>{};
@@ -1955,26 +1954,35 @@ class LikesRepository {
     return _queryEventsByEIds(eventIds, kind: EventKind.reaction);
   }
 
-  /// Events of [kind] targeting any of [eventIds] via `#e`, chunked so no
+  /// Events of [kind] targeting any of [eventIds] via `#e`.
+  Future<List<Event>> _queryEventsByEIds(
+    List<String> eventIds, {
+    required int kind,
+  }) {
+    return _queryChunked(
+      eventIds,
+      (chunk) => Filter(kinds: [kind], e: chunk),
+    );
+  }
+
+  /// Runs [buildFilter] over [values] in chunks of [_reqEidChunkSize], so no
   /// single REQ frame can exceed the relay's message cap.
   ///
   /// Deduplicates by event id once there is more than one chunk. Each chunk
   /// is its own `queryEvents` call with its own dedup box, so an event that
-  /// `#e`-tags targets in two different chunks comes back from both — and
+  /// tags targets in two different chunks comes back from both — and
   /// [getLikeCounts] counts per e-tag occurrence, so a raw concatenation
   /// would score that reaction twice. A single chunk is already deduped by
   /// the client.
-  Future<List<Event>> _queryEventsByEIds(
-    List<String> eventIds, {
-    required int kind,
-  }) async {
-    if (eventIds.isEmpty) return const <Event>[];
+  Future<List<Event>> _queryChunked(
+    List<String> values,
+    Filter Function(List<String> chunk) buildFilter,
+  ) async {
+    if (values.isEmpty) return const <Event>[];
 
     final results = await Future.wait([
-      for (final chunk in _chunkStrings(eventIds, _reqEidChunkSize))
-        _nostrClient.queryEvents([
-          Filter(kinds: [kind], e: chunk),
-        ]),
+      for (final chunk in _chunkStrings(values, _reqEidChunkSize))
+        _nostrClient.queryEvents([buildFilter(chunk)]),
     ]);
     if (results.length == 1) return results.first;
     return {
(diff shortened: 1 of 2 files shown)


hm21 (dismissed):
Chunking the reaction query introduced a double-count. Pushed the fix in 1eebb16.

Each chunk is its own `queryEvents` call with its own `EventMemBox`, so dedup-by-id happens per chunk — a reaction that `#e`-tags targets landing in two different chunks comes back from both, and the count loop in `getLikeCounts` adds per e-tag occurrence. A 600-id batch with one reaction tagging `event_id_0` and `event_id_500` returns 1/1 on `main` and returned 2/2 here. `_queryEventsByEIds` now dedupes by event id once there is more than one chunk; the single-chunk path is untouched, so no existing caller or fixture changes behaviour. Regression test added and mutation-verified (reverting the dedupe turns it red). Package suite 224 passing, `flutter analyze lib test` clean, all checks green.

One thing left for you: `_queryVoteTargetReactions` (`likes_repository.dart:1231`) is a third unbounded Kind 7 `#e` filter, reached from `getVoteCounts` / `getUserVoteStatuses` via `comment_reactions_bloc.dart:144`. It is bounded at 50 today only by `_pageSize` in `comments_list_bloc.dart` — the same incidental bound this PR argues against for the other two. Please route it through the helper as well (it needs an `authors:` parameter), and correct the description, which says "the two Kind 7 reaction `#e` filters".

Nit: the branch carries a `Merge branch 'main'` commit — AGENTS.md says rebase, never merge into a feature branch.

hm21:
Chunking the reaction query introduced a double-count. Pushed the fix in 1eebb16.

Each chunk is its own `queryEvents` call with its own `EventMemBox`, so dedup-by-id happens per chunk — a reaction that `#e`-tags targets landing in two different chunks comes back from both, and the count loop in `getLikeCounts` adds per e-tag occurrence. A 600-id batch with one reaction tagging `event_id_0` and `event_id_500` returns 1/1 on `main` and returned 2/2 here. `_queryEventsByEIds` now dedupes by event id once there is more than one chunk; the single-chunk path is untouched, so no existing caller or fixture changes behaviour. Regression test added and mutation-verified (reverting the dedupe turns it red). Package suite 224 passing, `flutter analyze lib test` clean, all checks green.

One thing left for you: `_queryVoteTargetReactions` (`likes_repository.dart:1231`) is a third unbounded Kind 7 `#e` filter, reached from `getVoteCounts` / `getUserVoteStatuses` via `comment_reactions_bloc.dart:144`. It is bounded at 50 today only by `_pageSize` in `comments_list_bloc.dart` — the same incidental bound this PR argues against for the other two. Worth routing through the helper as well (it needs an `authors:` parameter), and correcting the description, which says "the two Kind 7 reaction `#e` filters".

Nit: the branch carries a `Merge branch 'main'` commit — AGENTS.md says rebase, never merge into a feature branch.

realmeylisdev:
All three addressed — checks green on `1d1d65a6`.

**`_queryVoteTargetReactions`** — routed through the helper. Its `#a` coordinate leg was unbounded for the same reason, so it's capped too: `|coordinates| <= |eventIds|`, and a `kind:pubkey:dtag` costs more bytes per entry than a bare id, so that leg reaches the cap sooner than the `#e` one it runs beside.

Rather than add an `authors:` parameter I moved the chunk-and-dedup into `_queryChunked(values, buildFilter)`; `_queryEventsByEIds` is now the `#e` wrapper over it, and all four filters share one path. Every chunk stays a single-filter REQ so the local event cache stays active — the existing "queries e and a separately" test still pins that. Two new tests, mutation-verified: at `_reqEidChunkSize = 100000` all five chunking guards go red together.

**Description** — corrected. It now names three `#e` filters plus the coordinate leg, and states the vote path's incidental bound (`_pageSize = 50` at `comments_list_bloc.dart:61`).

**Merge commits** — rebased out. Your `1eebb16` survived as `da0017f1`; the net diff against `main` is byte-identical across the rebase.

hm21 (approved):
LGTM. All three addressed, and I verified rather than took the claims: the rebase preserved my dedupe commit byte-identical, the mutation at `_reqEidChunkSize = 100000` turns all five chunking guards red together, and reverting the dedupe inside the new `_queryChunked` still turns the cross-chunk test red — so the refactor didn't quietly drop it. 226 passing, format and analyze clean, checks green.

Folding the `#a` leg in was the right call; `_queryChunked` reads better than the `authors:` parameter I suggested.

NotThatKindOfDrLiz:
@realmeylisdev I pushed one takeover fix in f637afe6b: the addressable like path now routes its `#a` reaction filter through `_queryChunked`, and the existing addressable-path test now asserts both the `#e` and `#a` chunks split 500 + 100 without dropping values.

I also updated the PR body so it says five guarded query legs, not four.

Verification:
- `flutter test` from `mobile/packages/likes_repository`: 226 passing
- `flutter analyze lib test` from `mobile/packages/likes_repository`: clean
- pre-push hook: analyzer clean and changed-file test `packages/likes_repository/test/src/likes_repository_test.dart` passed, 166 tests
- GitHub checks: all 17 green on f637afe6b

NotThatKindOfDrLiz (approved):
Approved after takeover fix. I verified the remaining unchunked addressable like `#a` leg is now routed through `_queryChunked`, the addressable-path regression covers both `#e` and `#a` filters, the PR body is accurate, local verification passed, and all GitHub checks are green on f637afe6b.

mobile/packages/likes_repository/lib/src/likes_repository.dart
```diff
@@ -27,17 +27,16 @@ const _likeContent = '+';
...
-/// every frame small (500 ids ~= 34KB, well under 512KB and the advertised
-/// `max_event_tags` of 2000). See #5751.
-const _deletionReqEidChunkSize = 500;
+/// A 64-char id costs ~67 bytes in JSON, so a single `#e` filter with enough
+/// of them builds a REQ frame near/over the relay's `max_message_length`
+/// (512KB). An oversized frame errors the socket and `queryEvents` fails
+/// open — silently counting deleted likes, or returning no reactions at all.
+/// 500 ids is ~34KB, well under 512KB and the advertised `max_event_tags`
+/// of 2000. A `kind:pubkey:dtag` coordinate costs more per entry than a bare
+/// id, but 500 of them still lands far inside the cap. See #5751.
```
dcadenas: The `#e` legs are genuinely byte-bounded here, because an event id is fixed at 64 chars. The `#a` leg is not, and this line asserts otherwise.

`addressableDTag` is read straight off the relay event, and I found no length validation on it anywhere in the models layer. So the chunk is bounded by count, not by bytes.

The arithmetic, for what it is worth: a normal Divine coordinate is `34236:` + a 64-char pubkey + `:` + a 64-char sha256 d-tag, about 135 bytes, so 500 of them is roughly 67 KB. That is a long way inside 512 KB, and it is why I am not treating this as blocking. Getting to the cap at 500 entries needs d-tags around 970 chars, which nothing in Divine's own writer produces.

What makes it worth a comment is that it is the same shape the PR body argues against for the `#e` legs. The bound is real but incidental, and it lives in whatever publishes the `d` tag rather than at the query boundary. A byte-based bound on the `#a` path would be the thing that actually closes it. Reasonable to leave for a follow-up, but the sentence on this line reads as a guarantee and it is closer to an observation about typical data.

mobile/packages/likes_repository/lib/src/likes_repository.dart
```diff
@@ -1841,36 +1830,75 @@ class LikesRepository {
...
     return deleted;
   }
 
+  /// Kind 7 reactions targeting any of [eventIds] via `#e`.
+  Future<List<Event>> _queryReactionsByEventIds(List<String> eventIds) {
+    return _queryEventsByEIds(eventIds, kind: EventKind.reaction);
+  }
+
+  /// Events of [kind] targeting any of [eventIds] via `#e`.
+  Future<List<Event>> _queryEventsByEIds(
```
dcadenas, line 1964: This helper would also cover the one tag-value leg in the file that is still unchunked: the Kind 5 deletion filter in `_resolveRemoteDownvoteRecord`, which builds `e: candidates.map((event) => event.id).toList()` directly.

`candidates` is the current user's own live kind-7 `-` reactions on a single target, so reaching 500 means one person toggling a downvote on one event 500 times. There is no realistic path to the cap, which is why I did not push this myself and would not hold the PR for it.

Noting it because it is the last one, and routing it through `_queryEventsByEIds(..., kind: EventKind.eventDeletion)` is a one-liner with no behavior change if you want the file to be uniform.

dcadenas (approved):
🧱 the frame bound moves from incidental to enforced on the id legs — two non-blocking suggestions inline.

mobile/packages/likes_repository/lib/src/likes_repository.dart
realmeylisdev: You're right that the sentence claimed more than the code delivered, so I enforced the claim rather than softening it — `81d171acd`.

Chunks now close on **either** 500 values or 128KB of cumulative UTF-8 tag-value bytes, whichever comes first (`_reqTagValueChunkBytes`). The byte cap can never bind on a `#e` leg — 500 × 64 chars is 32KB — and never splits a realistic `#a` batch either, since 500 coordinates at your ~135 bytes is ~68KB. It only engages on the case you identified, where the `d` tag is long enough that count stops implying bytes. A single value over the budget still ships alone; that REQ can't be made smaller.

Mutation-verified in both directions, which is what pins the "never binds for ids" half: `_reqTagValueChunkBytes = 1 << 60` turns only the new byte test red and leaves all six count guards green, while `_reqEidChunkSize = 100000` turns the six count guards red together.

Your arithmetic is in the constant's doc, since it's the reason 128KB was the right budget — big enough that the common path never splits, small enough to keep 4x headroom under the 512KB cap.

mobile/packages/likes_repository/lib/src/likes_repository.dart
realmeylisdev: Fixed in `0975bd2a4` — you found the last one.

It routes through `_queryChunked` rather than `_queryEventsByEIds`, because the probe's `authors: [pubkey]` scope is load-bearing: the consumer takes any returned `e` tag as a deletion of the user's own reaction without re-checking the author, unlike `_deletedReactionIds` which matches author per event. The `buildFilter` callback keeps that scope on every chunk, and the test asserts it survives the split.

Agreed there's no realistic path to 500 — it needs one person toggling a downvote on one event 500 times. Worth doing anyway so the file has no remaining leg where the bound lives outside the query boundary.

realmeylisdev:
@dcadenas both suggestions are in, checks green on `81d171acd`.

**`_resolveRemoteDownvoteRecord`** (`0975bd2a4`) — the last unchunked leg, routed through `_queryChunked`. It keeps its `authors: [pubkey]` scope on every chunk, since that consumer trusts the filter for the own-deletion check instead of matching per event the way `_deletedReactionIds` does.

**The `#a` byte bound** (`81d171acd`) — rather than softening the line-38 sentence I made it true. Chunks now close on either 500 values or 128KB of tag-value bytes. It can never bind for 64-char ids (32KB at the count cap) or for realistic coordinates (~68KB at 500), so the common path is untouched; it only engages where a `d` tag is long enough that count stops implying bytes.

Mutation-verified in both directions, which is what pins that the two caps guard different things: `_reqTagValueChunkBytes = 1 << 60` reddens only the new byte test, `_reqEidChunkSize = 100000` reddens the six count guards together.

228 package tests, coverage 97.15%, analyze clean. PR body updated to six guarded legs plus the byte bound.

PR #6999, opened 2026-08-10
ci(supporters): pass the supporter defines through Codemagic builds

### LINKED ISSUE #7000: ci(supporters): pass the supporter defines through Codemagic builds
Auto-created from PR #6999.

**PR:** https://github.com/divinevideo/divine-mobile/pull/6999
**Branch:** `feat/supporters-codemagic-defines`

rabble:
Closes #7000

The supporter feature needs two build-time values that no workflow supplied: `SUPPORTERS_API_BASE_URL` and `FF_DIVINE_SUPPORTERS`. Without them the app has no Worker to talk to — `supporterApiClientProvider` returns null and a purchase can never be claimed.

## What changed

Both defines added to all seven build steps (Android APK and AAB, macOS, iOS IPA, and the PR-preview builds), and a `supporters_credentials` environment group attached to the six workflows.

New `mobile/scripts/check_supporters_config.sh`, mirroring the existing C2PA endpoint guard.

## Why the guard

Codemagic interpolates these variables unconditionally, so an unset one emits `--dart-define=NAME=` — present but empty, which overrides the Dart-side default rather than falling back to it. That's the trap `check_signing_endpoint.sh` already documents for C2PA.

For these two, empty happens to fail *safe* — but it fails *silently*. You'd ship a build where the supporter screen renders and can never reach the server, and nothing would say so until a purchase failed in someone's hands.

So the guard enforces that the two **agree**, not that both are set:

| Condition | Result |
|---|---|
| Both unset | pass — feature inert |
| `FF_DIVINE_SUPPORTERS=true`, no base URL | **fail** — flag on with no server |
| Flag set to anything but `true`/`false` | **fail** — `bool.fromEnvironment` silently yields false |
| Base URL not `https://` | **fail** |
| Base URL with a trailing slash | **fail** |
| Base URL carrying a path | **fail** |

The last two matter because `SupporterApiClient` calls `_baseUri.resolve(path)` (`supporter_api_client.dart:253`), which discards everything after the final slash — a trailing slash or path segment silently sends every request somewhere else.

All branches exercised locally.

## Needs a Codemagic UI step before it does anything

Create the `supporters_credentials` environment variable group with:

- `SUPPORTERS_API_BASE_URL` — `https://divine-supporters-staging.protestnet.workers.dev` for test builds, `https://supporters.divine.video` for production
- `FF_DIVINE_SUPPORTERS` — `true` or `false`

Until that group exists the builds behave exactly as they do today, since both values resolve empty and the feature stays off.

### DIFF
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -38,6 +38,14 @@ definitions:
     - &check_ios_extension_signing_contract
       name: Check iOS extension signing contract
       script: ./scripts/check_ios_extension_signing_contract.sh
+    - &check_supporters_config
+      name: Check supporter subscription config
+      script: |
+        # Both variables below are interpolated unconditionally by every build
+        # step, so an unset one ships an empty --dart-define that overrides the
+        # in-app default. Empty fails safe here but fails silently; this rejects
+        # the combination that ships broken (flag on, no server).
+        ./scripts/check_supporters_config.sh
     - &check_signing_endpoint
       name: Check C2PA signing endpoint
       script: |
@@ -144,7 +152,9 @@ definitions:
           --dart-define=ZENDESK_URL=$ZENDESK_URL \
           --dart-define=DEFAULT_ENV=${{ inputs.DEFAULT_ENV }} \
           --dart-define=PROOFMODE_SIGNING_SERVER_ENDPOINT=$PROOFMODE_SIGNING_SERVER_ENDPOINT \
-          --dart-define=PROOFMODE_SIGNING_SERVER_TOKEN=$PROOFMODE_SIGNING_SERVER_TOKEN
+          --dart-define=PROOFMODE_SIGNING_SERVER_TOKEN=$PROOFMODE_SIGNING_SERVER_TOKEN \
+          --dart-define=SUPPORTERS_API_BASE_URL=$SUPPORTERS_API_BASE_URL \
+          --dart-define=FF_DIVINE_SUPPORTERS=$FF_DIVINE_SUPPORTERS
     - &build_aab
       name: Build Android AAB
       script: |
@@ -167,7 +177,9 @@ definitions:
           --dart-define=ZENDESK_URL=$ZENDESK_URL \
           --dart-define=DEFAULT_ENV=${{ inputs.DEFAULT_ENV }} \
           --dart-define=PROOFMODE_SIGNING_SERVER_ENDPOINT=$PROOFMODE_SIGNING_SERVER_ENDPOINT \
-          --dart-define=PROOFMODE_SIGNING_SERVER_TOKEN=$PROOFMODE_SIGNING_SERVER_TOKEN
+          --dart-define=PROOFMODE_SIGNING_SERVER_TOKEN=$PROOFMODE_SIGNING_SERVER_TOKEN \
+          --dart-define=SUPPORTERS_API_BASE_URL=$SUPPORTERS_API_BASE_URL \
+          --dart-define=FF_DIVINE_SUPPORTERS=$FF_DIVINE_SUPPORTERS
     - &build_macos
       name: Build macOS app
       script: |
@@ -180,7 +192,9 @@ definitions:
           --dart-define=ZENDESK_URL=$ZENDESK_URL \
           --dart-define=DEFAULT_ENV=${{ inputs.DEFAULT_ENV }} \
           --dart-define=PROOFMODE_SIGNING_SERVER_ENDPOINT=$PROOFMODE_SIGNING_SERVER_ENDPOINT \
-          --dart-define=PROOFMODE_SIGNING_SERVER_TOKEN=$PROOFMODE_SIGNING_SERVER_TOKEN
+          --dart-define=PROOFMODE_SIGNING_SERVER_TOKEN=$PROOFMODE_SIGNING_SERVER_TOKEN \
+          --dart-define=SUPPORTERS_API_BASE_URL=$SUPPORTERS_API_BASE_URL \
+          --dart-define=FF_DIVINE_SUPPORTERS=$FF_DIVINE_SUPPORTERS
     - &package_macos_dmg
       name: Package macOS DMG
       script: |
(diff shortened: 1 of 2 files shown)


mobile/scripts/check_supporters_config.sh
```diff
@@ -0,0 +1,86 @@
...
+
+case "$base_url" in
+  */)
+    echo "ERROR: SUPPORTERS_API_BASE_URL must not end in a slash." >&2
+    echo "  SupporterApiClient calls Uri.resolve(), which drops everything after" >&2
+    echo "  the final slash, so a trailing slash changes where requests go." >&2
+    echo "  Got: $base_url" >&2
+    exit 1
+    ;;
+esac
```
realmeylisdev, line 84: This rejects a base URL that works. The client normalises before it resolves:

```dart
static Uri _trimBaseUri(Uri uri) {
  final text = uri.toString().replaceFirst(RegExp(r'/+$'), '');
  return Uri.parse('$text/');
}
```

— `supporter_api_client.dart:342`

Driving the real `SupporterApiClient.fetchMe()` with a transport that records the final URI:

| base URL | this guard | request actually sent |
| --- | --- | --- |
| `https://supporters.divine.video` | pass | `https://supporters.divine.video/v1/me` |
| `https://supporters.divine.video/` | **fail** | `https://supporters.divine.video/v1/me` |
| `https://supporters.divine.video///` | **fail** | `https://supporters.divine.video/v1/me` |

Identical. The trailing slash is stripped and re-added before anything resolves, so it cannot change where requests go.

The path rule on L77 inverts the same way: `https://api.divine.video/supporters` sends `https://api.divine.video/supporters/v1/me`. The segment is preserved, not discarded — preserving it is what `_trimBaseUri` is for. Nothing is mounted under a path today, so this blocks nothing right now, but it would block that shape if it ever changed.

Either drop both rules, or keep them as canonical-form enforcement with the reason corrected. As written, a release build fails on a URL that works and the message sends whoever is on call looking for a bug that is not there.

mobile/scripts/check_supporters_config.sh
```diff
@@ -0,0 +1,86 @@
...
+# happens to fail safe (empty URL means supporterApiClientProvider returns
+# null, empty flag is false), but it fails *silently*: you get a build where
+# the supporter screen appears and can never reach the server, and nothing
+# says so until a purchase 500s in someone's hands.
+#
+# So the invariant enforced here is not "both must be set" — it is "these two
+# must agree". A flag turned on without a server is the broken combination.
+#
+# Portable: POSIX sh constructs + grep -E only, so it runs on CI ubuntu and on
+# the macOS pre-push hook.
```
realmeylisdev: Right for the URL, not for the flag. An empty define does override `String.fromEnvironment`, but `bool.fromEnvironment` falls back to `defaultValue` for anything that is not literally `true` or `false` — including empty.

Through the real pipeline: `flutter test --dart-define=FF_DEBUG_TOOLS=` still yields `true`, because that flag declares `defaultValue: true`.

It does not bite here — `FF_DIVINE_SUPPORTERS` has no `defaultValue`, so empty does land on `false`, and the rule at L30 is the right rule. It is the stated model that is off, and L32-33 reads as general when it is specific to this flag. Worth wording so the next flag routed through this file does not inherit it.

Minor, on L20-21: that line came over from `check_signing_endpoint.sh`, but this script only runs on the Codemagic macOS builders — nothing in `.github/workflows` or the pre-push hook calls it.

mobile/scripts/check_supporters_config.sh
```diff
@@ -0,0 +1,86 @@
...
+if printf '%s' "$base_url" | grep -Eq '^https://[^/]+/.+'; then
+  echo "ERROR: SUPPORTERS_API_BASE_URL must be an origin with no path." >&2
+  echo "  SupporterApiClient resolves '/v1/...' against this value, so a path" >&2
+  echo "  segment here is discarded and requests silently go elsewhere." >&2
+  echo "  Got: $base_url" >&2
+  echo "  Expected something like: https://supporters.divine.video" >&2
+  exit 1
+fi
+
+echo "Supporter config OK (flag='${flag:-unset}', base='$base_url')."
```
realmeylisdev, line 86: The one shape that genuinely breaks slips through. `SUPPORTERS_API_BASE_URL=https://supporters.divine.video ` — trailing space, an easy paste into the Codemagic UI — passes every check above and then sends:

`https://supporters.divine.video%20/v1/me`

That host never resolves. A *leading* space is already safe, since `Uri.parse` throws on it; only a trailing one gets through.

```sh
case "$base_url" in
  *[[:space:]]*)
    echo "ERROR: SUPPORTERS_API_BASE_URL contains whitespace." >&2
    exit 1
    ;;
esac
```

codemagic.yaml
```diff
@@ -527,6 +549,7 @@ workflows:
       groups:
         - zendesk_credentials
         - proofmode_credentials
+        - supporters_credentials
```
realmeylisdev, line 553: The header block tracks group creation as a numbered setup step — 5 for `zendesk_credentials`, 7 for `proofmode_credentials`, 8 for `github_credentials`. This group needs the same entry. It is where someone bringing up a fresh Codemagic project looks, and right now the PR description is the only place the required values are written down.

realmeylisdev (commented):
Wiring checks out: both defines reach all seven build steps, all six workflows import the group, and `working_directory: mobile` makes the script path resolve. The two rules that matter are load-bearing and I confirmed both — `FF_DIVINE_SUPPORTERS=TRUE` really does compile to `false`, and with both values empty the feature is inert, so merging ahead of the Codemagic group is safe.

The two URL-shape rules go the other way. `SupporterApiClient` normalises the base URI before it resolves, so a trailing slash is a no-op and a path prefix is preserved — the guard hard-fails both, and the error text says the opposite of what the client does. Meanwhile a trailing space, the one shape that really does break the host, passes. Notes inline.

Not pushing these — `codemagic.yaml` is release automation, outside reviewer takeover.

PR #6936, opened 2026-08-09
docs(maintainability): refresh oversized-file inventory

NotThatKindOfDrLiz:
## Summary
- refresh the advisory oversized-file baseline to the current 58-file inventory
- update docs/code-review/issues-simplicity.md so #4339 reflects current state instead of the April snapshot
- keep check_file_size_ceiling.sh advisory and non-blocking, while linking follow-up cleanup issues #6933, #6934, and #6935

## Verification
- bash mobile/scripts/check_file_size_ceiling.sh
- flutter test test/tools/file_size_ceiling_test.dart
- flutter analyze lib test integration_test

Part of #4339.

PR #7002, opened 2026-08-10
test(video-editor): poll for the unawaited session-end reap

### LINKED ISSUE #7003: test(video-editor): poll for the unawaited session-end reap
Auto-created from PR #7002.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7002
**Branch:** `fix/video-editor-deferred-reap-test-race`

rabble:
Closes #7003

## The problem

A test was failing at random, on PRs that had nothing to do with it.

It hit #6936 — a **docs-only** PR whose entire diff was one markdown file and one baseline `.txt` — and took down `Tests (shard 1/4)`. Nothing in that PR could touch video editing. It cost a CI cycle and an investigation to work out why.

## Why it was failing

The test checks that closing the video editor cleans up a leftover temp file. It did three things:

1. call `reset()` — the "editor session ended" hook
2. wait one tick of the event loop
3. assert the file is gone

But `reset()` *starts* the cleanup and deliberately doesn't wait for it, so closing the editor isn't held up by disk work. The cleanup then has to ask the database "is any saved draft still using this file?" before it deletes anything. That's several async steps.

So the test was checking for the result of a background job after waiting a single tick. On an idle laptop one tick happens to be enough, and it passes. On a loaded CI runner it isn't, and it fails — on whatever PR is unlucky enough to be running.

## The fix

Wait properly instead of guessing.

The file already has a helper for this, `expectFileReaped`, which re-checks up to 20 times instead of once. The **neighbouring test in the same group already uses it**; this one just didn't. Now they match.

That's the whole change — four lines in one test.

## What this deliberately does not change

No production code.

The don't-wait-for-cleanup behaviour is intentional and documented on `reset()`. The test's assumption was wrong, not the app. "Fixing" the app to satisfy the test would mean closing the video editor starts blocking on a database query and file deletion — a real user-facing regression in service of a test that was written slightly wrong.

## Testing

- `flutter test test/providers/video_editor_provider_test.dart` — 113 pass
- The file passed in isolation both before and after. That's expected: the bug only appears under load, which is why CI kept catching it and local runs never did.

### DIFF
--- a/mobile/test/providers/video_editor_provider_test.dart
+++ b/mobile/test/providers/video_editor_provider_test.dart
@@ -3209,11 +3209,11 @@ void main() {
       // not the @visibleForTesting flush — this exercises the wiring an app
       // actually hits when the editor closes.
       await notifier.reset(keepAutosavedDraft: true);
-      await pumpEventQueue();
 
-      expect(
-        orphan.existsSync(),
-        isFalse,
+      // reset() leaves the reap unawaited, so poll rather than assuming a
+      // single pump covers the DAO round trip behind the delete.
+      await expectFileReaped(
+        orphan,
         reason: 'session end reaps a deferred file once nothing references it',
       );
       expect(notifier.deferredFileCleanupForTest, isEmpty);
--- a/mobile/test/providers/video_editor_provider_test.dart
+++ b/mobile/test/providers/video_editor_provider_test.dart
@@ -3319,12 +3319,12 @@ void main() {
       expect(oldRendered.existsSync(), isTrue);
       expect(notifier.deferredFileCleanupForTest, contains(oldRendered.path));
 
+      // onDispose leaves the reap unawaited too, so poll rather than assuming a
+      // single pump covers the DAO round trip behind the delete.
       disposeContainer();
-      await pumpEventQueue();
 
-      expect(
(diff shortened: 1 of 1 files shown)


hm21 (approved):
Same single-pump assumption was still in the group's fourth test, `container teardown reaps a replaced final rendered file` — `onDispose` starts the reap unawaited exactly like `reset()` does. It failed that way on #6964 shard 2/4, independently of the #6936 failure this PR came from, so it would have kept flaking after this merged.

Pushed 3649946da5 switching it to `expectFileReaped` as well. All four tests in the group now use the same helper. No production code touched, same as your commit.

Validation: `flutter test test/providers/video_editor_provider_test.dart` — 113 pass, also with `--test-randomize-ordering-seed random`; format and analyze clean; full CI green on the new head.

LGTM — yours to merge.

rabble:
Superseded — main already has this fix.

PR #7014 (badges) landed the identical change in `78f6089e7`: the `expectFileReaped` polling helper and its use at the session-end reap assertion are both on main verbatim. Rebasing this branch leaves only a two-line explanatory comment, which isn't worth a PR.

The flake this was opened for is fixed. Closing #7003 stays valid — the fix shipped, just under a different PR.

Issue #7001, opened 2026-08-10
fix(likes): a single pubkey can publish repeat + reactions to one video, inflating the like count


realmeylisdev:
## Summary

A single pubkey can end up with **several live `+` reactions on the same video**. Because funnelcake counts `uniq(reaction_event_id)` with no dedupe by reactor, every extra reaction permanently adds 1 to the displayed like count while the "Liked by" list — which groups by pubkey — does not move. One person can inflate a video's headline count by an arbitrary amount without meaning to.

This is the dominant driver of the #6021 symptom on videos that were never edited, so it is not fixed by any revision-widening work.

## Evidence

Measured against production relay + `api.divine.video` on 2026-08-10, using the reaction history of one signed-in account:

- 98 own kind 7 reactions across **68 distinct targets**
- **15 of 68 targets carry more than one reaction**

Two of those are divine-mobile (`client: Divine … divine-mobile`) on addressable kind 34236 videos:

| target | reactions | detail |
|---|---|---|
| `8dfd7bbd51c2aace60d30b61a95e70dae3b20a3e2d393787b4588336f72f0a0c` | 2 × `+` | `2026-06-19 18:54:41` and `18:54:47` — **6 seconds apart, no `-` between** |
| `07394d2a5074fcfd07884c7f59c5011824e496d5508628319676bf8453b70d0d` | 2 × `+` | 2026-07-19 and 2026-08-02 |

Resulting divergence, both on videos with exactly **one** revision (never edited, so stranding cannot be involved):

```
8dfd7bbd…  count=89  likers=84  (84 distinct pubkeys)   delta 5
07394d2a…  count=31  likers=28                          delta 3
```

Two consecutive `+` with no intervening `-` cannot be produced by a correctly toggling button — the second tap should have unliked.

A third-party example of the same class, from divine-web: pubkey `6eaeba40561bc0bdfe60119a485c8fb4dd9a767408dd9d26fa964310679ed00f` published **three** `+` reactions to `e69a48f32e05d6bd0881f3d6582209060f9521baed1e47205e7f64b4c38edb80` at `03:27:43`, `03:27:45`, `03:27:47` on 2026-08-10. That video now reads **count 3, likers 1**.

## What I ruled out

[...]

PR #7004, opened 2026-08-10
fix(likes): retract a like unliked while its publish is in flight

realmeylisdev:
Refs #7001.

## The bug

Unlike inside the publish window left the reaction **live on the relay with no Kind 5 retracting it**, and cleared the local record. The heart went empty while the like still counted, and the next tap published a second live `+`.

That is the duplicate observed in production: two `+` reactions from one pubkey on `8dfd7bbd51c2aace60d30b61a95e70dae3b20a3e2d393787b4588336f72f0a0c`, **6 seconds apart, no `-` between them and no Kind 5 deleting either** — both still live. funnelcake reads that video as count 89 / likers 84, on a video with exactly one revision, so no amount of revision widening touches it.

## Root cause

`unlikeEvent` returned early whenever the record still held a `pending_` placeholder:

```dart
if (snapshotRecord.reactionEventId.startsWith('pending_')) {
  // Pending like never reached the relay; nothing to delete on the wire.
  return;
}
```

**That assumption is false.** `likeEvent` writes the placeholder at `:509-521` and only swaps it for the real reaction id at `:570`, *after* `await _nostrClient.sendLike(...)` returns. And the send reports success as soon as one socket takes the frame — `RelayPool.send` returns `result.sentTo.isNotEmpty`, whose own doc says an entry means "the relay's WebSocket accepted the frame … It does NOT mean the relay accepted the event at the protocol level."

So for the whole round trip the record reads `pending_` while the reaction may already be stored. `unlikeEvent` deindexes the local record first (`:755`), then hits that early return — reaction live, no retraction, local state says "not liked".

## The fix

Record the intent instead of dropping it. `likeEvent` retracts the reaction the moment it has a real id to reference.

A failed retraction re-indexes the confirmed record, so a later unlike can still find the reaction — dropping it is what stranded it in the first place. The intent is also cleared when the publish genuinely fails (nothing to retract) and on `clearCache`.

## What I ruled out first

- **In-isolate double-tap.** Already closed: `:492` checks `_likeRecords.containsKey` and `:516` indexes the placeholder with no `await` between, so check-then-set is atomic; a second concurrent call throws `AlreadyLikedException`.
- **Publish-failure retry double-publishing.** I first ruled this out, then **re-checked and found I was wrong** — it is real, and it is a *separate* mechanism from the one this PR fixes. `RelayBase.send` queues the EVENT frame for replay-on-reconnect **and still returns false** (`relay_base.dart:154-157`), and `_sendCollect` passes `queueIfFailed: deadline == null`, which is true on the like path (`relay_pool.dart:1927`). So the frame can land on reconnect while `publishEvent` reports `PublishFailed`, `likeEvent`'s catch queues a retry, and `executeLikeAction` publishes a second one — its coordinate guard cannot stop that, because `_indexLikeRecord` writes the *same* `LikeRecord` instance into both maps, so `byCoordinate.reactionEventId != ownReactionEventId` reduces to `'pending_like_X' != 'pending_like_X'` = false. That path needs its own fix and is tracked on #7001; it is out of scope here and this PR does not claim to close it.
- **like → unlike → like toggling.** Would show a Kind 5 between the two `+`. Queried the relay: that author has 112 Kind 5 deletions, and **neither** of these two reactions is targeted by any of them.
- **The DM reaction sweep** in the device logs is `dm_reaction_retry_service.dart` — DM reactions, not video likes.

## Testing

New test `retracts a reaction whose publish is still in flight (#7001)` holds `sendLike` open on a `Completer`, unlikes inside the window, then resolves the publish, and asserts the Kind 5 goes out for the real reaction id.

Mutation-verified rather than just observed green — removing the recorded intent turns it red:

```
+31 -1: LikesRepository unlikeEvent retracts a reaction whose publish is still in flight (#7001) [E]
  No matching calls. All calls: MockNostrClient.sendLike(...)
```

- `flutter test` in `packages/likes_repository`: **222 passing**
- coverage **96.50%** (gate: 62)
- `flutter analyze lib test integration_test`: clean

## Scope

This closes the mechanism I could reproduce. It does **not** retract the duplicates already published — those need a separate decision, and #7001 tracks that. The count/list denominator mismatch that turns a duplicate into a permanent visible gap is divinevideo/divine-funnelcake#626.

### DIFF
--- a/mobile/packages/likes_repository/lib/src/likes_repository.dart
+++ b/mobile/packages/likes_repository/lib/src/likes_repository.dart
@@ -151,8 +151,12 @@ class LikesRepository {
   /// (unlike the in-memory-only precedent in #4478's comment count cache).
   final Map<String, LikeRecord> _likeRecordsByAddressableId = {};
 
-  /// Target event ids whose in-flight like publish must be retracted as soon
-  /// as it resolves.
+  /// Pending like placeholder ids currently owned by an awaiting [likeEvent]
+  /// publish.
+  final Set<String> _inFlightLikePublishPlaceholders = {};
+
+  /// Pending like placeholder ids whose in-flight publish must be retracted as
+  /// soon as it resolves.
   ///
   /// [unlikeEvent] cannot reference a `pending_` placeholder in a Kind 5 —
   /// the real reaction id does not exist yet. The reaction may already be
@@ -560,6 +564,7 @@ class LikesRepository {
     // relay is technically connected). Without a wired callback, fall back
     // to rollback + rethrow to preserve the original contract for tests
     // and non-app embedders.
+    _inFlightLikePublishPlaceholders.add(placeholderId);
     try {
       final reactionEvent = await _nostrClient.sendLike(
         eventId,
@@ -580,8 +585,14 @@ class LikesRepository {
         addressableId: addressableId,
       );
 
-      if (_unlikeRequestedWhilePending.remove(eventId)) {
-        await _retractLikeUnlikedMidPublish(reactionEvent.id, confirmed);
+      _inFlightLikePublishPlaceholders.remove(placeholderId);
+      if (_unlikeRequestedWhilePending.remove(placeholderId)) {
+        await _retractLikeUnlikedMidPublish(
+          confirmed: confirmed,
+          restoredCount: previousCount == null ? null : previousCount + 1,
+          countEventId: eventId,
+          countAddressableId: addressableId,
+        );
         return reactionEvent.id;
       }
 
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz:
@realmeylisdev I took over the review fixes on this PR and pushed `305362bc`.

What changed:
- Scoped the pending-unlike intent to the specific in-flight placeholder id, so stale persisted/offline `pending_like_*` records cannot retract a later unrelated like.
- Kept the original in-flight retraction behavior for the actual `likeEvent` publish attempt.
- Made failed mid-publish retractions keep the confirmed record, restore the cached count, and queue an unlike retry when the app queue is wired.
- Updated the PR body from `Fixes #7001` to `Refs #7001` so #7001 stays open for Producers A/B.
- Rebased onto current `origin/main`; the merge-main commit is gone.

Verification:
- `flutter test test/src/likes_repository_test.dart` from `mobile/packages/likes_repository`
- `flutter analyze lib test integration_test` from `mobile/`
- Pre-push hook: analyze + changed package test passed
- GitHub checks are green

Please review the takeover changes and merge when ready.

NotThatKindOfDrLiz (approved):
Takeover changes look good to merge after author review. The stale pending-intent bug is fixed, retry/count behavior is covered, #7001 remains open via non-closing linkage, and CI is green.

NotThatKindOfDrLiz:
Final rebase landed as `80d4c5af` on current `origin/main`. GitHub checks are green and merge state is clean.

PR #7005, opened 2026-08-10
ci(e2e): make the service workflow run, and fix the e2e suites it covers

realmeylisdev:
## What

Makes `Mobile Integration Tests (Services)` a workflow that runs, and fixes the four `integration_test/e2e/` suites that could not have passed under it anyway.

Before this PR the workflow had **0 runs in its entire history** (`gh api .../actions/workflows/261929607/runs` → `total_count: 0`), so "it has always been green" was vacuous — it had never started.

Refs #4239.

## The three problems, all measured

**1. `flutter test integration_test/…` requires a device, and the runner selected none.** `flutter_tools` decides integration-test mode from the path alone (`_shouldRunAsIntegrationTests`, `commands/test.dart:888`), then requires `findTargetDevice()` to return non-null and explicitly rejects web devices. `ubuntu-24.04` exposes exactly two devices — Linux (desktop) and Chrome (web) — so with no `-d` the selection is *ambiguous*, `findTargetDevice()` returns null, and the run dies on a misleading message:

```
More than one device connected; please specify a device with the '-d <deviceId>' flag…
No devices are connected. Ensure that `flutter doctor` shows at least one connected device
```

Naming `-d linux` and installing the Linux desktop toolchain gives the runner a device it can build for — no emulator, no Apple account.

**2. The directory was mixed.** `flutter test <dir>` filters by tag only *after* loading every file. `integration_test/e2e/` held four `patrolTest` suites and two `@Tags(['service'])` suites, so the tagged ones never got to run — the patrol files failed first at load with `Binding is already initialized to IntegrationTestWidgetsFlutterBinding` (`PatrolBinding` extends `LiveTestWidgetsFlutterBinding` and cannot be constructed once another binding exists).

**3. All four patrol suites hang on launch — so no workflow change alone could ever have made them pass.** Every one of them called `pumpAndSettle` immediately after `launchAppGuarded`. The app polls email verification every 3s, so the tree never reaches a quiescent frame and the call blocks until its 10-minute timeout. `.claude/skills/e2e-test/SKILL.md` documents this exact sequence as hanging.

Measured on an iOS simulator: patrol emitted

```
PATROL_LOG {"type":"test",…,"status":"start"}
```

and then **no terminal status at all**. The app stayed alive logging `[AUTH] Polling for email verification` until the runner gave up at **6m36s** — `Total: 0`, xcresult `((passed) is true) failed - (no details - app likely crashed)`. It reads like a crash; it is a hang. Three of the four have never run in CI on any platform, so nothing caught it.

## Why the suites drop patrol rather than gain a `RunnerUITests` target

Across all four, this is every use of the patrol tester:

| Suite | `$` usage |
|---|---|
| `c2pa_init_test` | `$.tester` ×1 |
| `deleted_video_visible_to_other_users_test` | `$.tester` ×1 |
| `video_creation_flow_test` | `$.tester` ×1 |
| `dm_optimistic_survives_watch_race_test` | `$.tester` ×1, `$.platformAutomator` ×1 |

The single native call dismissed an Android notification dialog inside a `try/catch` that logged and continued — best-effort, not load-bearing. So patrol bought these four nothing while costing a `RunnerUITests` target, XCTest wiring, and an emulator. Converting them to plain `testWidgets` also makes the directory homogeneous, which is what problem 2 actually needed — no directory move required.

The other **26** `patrolTest` suites (auth, video_recorder, privacy, perf, lifecycle) are untouched and keep patrol.

## Changes

- **Convert the four e2e suites off patrol**, replace the hanging launch with the existing bounded `pumpUntilSettled` helper, tag every suite `service`, and correct the `Requires:` notes — `c2pa_init` claimed it needed the Docker stack and imports no stack helper, while `deleted_video_visible_to_other_users` genuinely needs the full stack and said nothing.
- **Rework the workflow**: install the Linux desktop toolchain, run one `flutter test` per suite with an explicit `-d linux` under `xvfb`, drop the Docker steps, add `timeout-minutes`, and add a path-scoped `pull_request` trigger. **This is the one change with blast radius** — it introduces a new PR check. Reverting to dispatch-only is a two-line change.
- **Dispatch in `local_stack/profile.sh`** on whether the target contains `patrolTest`, so the converted suites keep the merged docker+logcat+app timeline instead of falling out of `mise run e2e_test`. The plain path pre-grants `POST_NOTIFICATIONS`, since without a native automator nothing can dismiss that dialog.

Tag placement is load-bearing and easy to get silently wrong: `package:test` reads suite metadata **only from the first directive**, and this repo's `analysis_options.yaml` sets both `invalid_annotation_target: ignore` and `library_annotations: ignore`, so a misplaced `@Tags` fails completely silently. Verified with the discriminator in both directions on a standalone package — an above-imports annotation is selected by `--tags service` and excluded by `--exclude-tags service`; a below-imports one behaves the opposite way.

## What this does *not* achieve, and why

The issue's acceptance criterion "the four tests in `integration_test/e2e/` all execute" **cannot be met**, and the workflow says so in a comment rather than pretending otherwise:

- `c2pa_init` — `c2pa_flutter` ships `android/` and `ios/` implementations and no desktop one, so it cannot run on the Linux device.
- `video_creation_flow` — needs the invite service.
- `deleted_video_visible_to_other_users`, `dm_optimistic_survives_watch_race` — need the full stack.

**`local_stack` cannot start on a GitHub runner at all.** Measured on the runner: `docker compose up -d` aborts at image resolution because `ghcr.io/divinevideo/divine-invite-darshan:e2e` is a **private** GHCR package and the workflow performs no `docker login`:

```
invite Error Head "https://ghcr.io/v2/divinevideo/divine-invite-darshan/manifests/e2e": unauthorized
```

Compose resolves every image before starting any, so **zero containers start**. Every other image in the stack pulls anonymously; this one alone takes the whole stack down. It is still 403 with a `gh auth token` carrying `admin:org, gist, repo, workflow` — the token lacks `read:packages`.

**Unblocking that is a maintainer/security call I have not taken**: it needs either a PAT with `read:packages` in repo secrets, or the package made internal. Until then those three suites stay local-only — now correctly tagged, so they are selectable the moment access lands.

## Verification

- Workflow green on dispatch: `relay_list_admission_test.dart` 6/6, `dm_inbox_relay_admission_test.dart` 3/3, with the #6585 filters visible in the logs.
- `flutter analyze integration_test` and `dart format` clean after the conversion; no `patrol` reference remains under `integration_test/e2e/`.
- `bash local_stack/test_android_scripts.sh` passes with the dispatch change; `.codex/scripts/test-config.sh` passes with the regenerated `.agents/skills` mirror.
- Tag placement verified with the discriminator in both directions (see above).

## Note for reviewers

Patrol on iOS is no longer blocked by `use_frameworks! :linkage => :static`, contrary to what we believed. patrol 4.7.0+ routes through Swift Package Manager rather than CocoaPods, and with a `RunnerUITests` target it builds and drives Dart tests on the simulator — verified end-to-end while investigating this. That is **not** in this PR, which needs none of it; recording it so the "patrol can't work on iOS here" belief doesn't get re-derived. The other 26 patrol suites still cannot run on iOS today.


NotThatKindOfDrLiz:
@realmeylisdev I took over the local runner regressions from review and pushed one follow-up commit: `fix(e2e): harden local integration dispatch` (363df4026).\n\nWhat changed:\n- `local_stack/profile.sh` now uses recursive `grep -rq`, so directory targets like the default `integration_test/auth/` dispatch to Patrol instead of falling into plain `flutter test`.\n- The plain `integration_test` path installs the app before pre-granting `POST_NOTIFICATIONS`, and preserves install output/exit status before running the suite.\n- `local_stack/test_android_scripts.sh` now asserts the recursive dispatch, Flutter path, device flag, and install-before-grant ordering.\n- The mirrored e2e skill docs now scope the direct `patrol test` / `flutter test` warning to e2e targets.\n\nVerification:\n- `local_stack/test_android_scripts.sh`\n- `.codex/scripts/test-config.sh`\n- `bash -n local_stack/profile.sh local_stack/test_android_scripts.sh`\n- `git diff --check origin/main...HEAD`\n- `mise exec -- flutter analyze lib test integration_test`\n- GitHub checks are green, including `Integration Tests (services)`.\n\nPlease review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover pass. I pushed the local runner fixes, rebase was clean, and all local/GitHub verification is green. @realmeylisdev please review the takeover commit before merging.

realmeylisdev:
Ran the `integration_test/e2e/` suites end-to-end on a physical iPhone over USB, and — where the phone cannot reach the stack — on an iOS simulator against a live `local_stack`.

**On the physical iPhone**

- `relay_list_admission` — 6/6
- `dm_inbox_relay_admission` — 3/3
- `c2pa_init` — fails, cert never created. That confirms the `Requires: an Android device` note on real hardware, not just on the simulator.

Both service suites match what the reworked workflow produces on the Linux runner, so the CI lane and a real device agree.

**Two commits pushed**

- `00349e6ef` — the conversion replaced `pumpAndSettle` with a fixed `pumpUntilSettled(maxSeconds: 3)`, and three seconds is not enough for the app to mount `MaterialApp`. `video_creation_flow` failed its first assertion at test-time 00:03 while the app was still booting; `startupSettled` logged seven seconds later. Now polls with `waitForWidget`, which is what the skill doc updated in this PR already recommends. Moves the suite from failing at 00:03 to clearing the launch assertion and reaching 00:12.
- `ad73c0cee` — removing `_dismissNotificationPermission($)` left two `pumpUntilSettled(tester)` calls back to back with only a comment between them. Collapsed to a single 10s window; same number of frames pumped.

**On the takeover commit** — the `grep -rq` change is load-bearing, not cosmetic. With the old non-recursive `grep -q`, the documented default target `integration_test/auth/` exits 2 ("Is a directory"), so the dispatch would have sent 13 patrol suites through `flutter test`. Verified in a real bash context, both directions.

**Three further blockers are pre-existing and are filed separately as #7051**: an unauthenticated `authStateStream.firstWhere` that throws `StateError: No element` and fails any e2e run that never signs in; a welcome-screen copy variant that breaks `navigateToCreateAccount` whenever an account is left over from a previous run; and the fact that LOCAL-stack e2e cannot run on a physical iOS device at all, since `localHost` resolves to the phone's own loopback and both alternatives are blocked by ATS and the missing `NSLocalNetworkUsageDescription`. None of the three is caused by this PR and none affects the two suites CI runs.

All checks are green on the pushed commits, including `Integration Tests (services)`.

PR #7009, opened 2026-08-10
fix(a11y): give the profile tabs and grids real accessible names

### LINKED ISSUE #6951: fix(a11y): profile tabs announce their test anchors to screen readers
## Problem

The profile tab bar announces its snake_case test anchors to screen-reader users. VoiceOver reads the Liked tab as **"liked_tab"**, the Reposts tab as **"reposted_tab"**, and so on.

`profile_grid.dart` supplies each tab a `label` that is really a test id:

```dart
ProfileTabKind.liked => (label: 'liked_tab', icon: DivineIconName.heart),
```

and `profile_tab_bar.dart` passes it straight into `Semantics(label: ...)`. The same pattern applied to the liked and saved grids, where each thumbnail announced `liked_video_thumbnail_0`.

## What #6948 did and did not do

It added `identifier:` alongside the existing `label:`, so E2E targets the id and no longer depends on `TabBar`'s compound `"Tab N of M"` label. That fixed the test-anchor problem.

It **left the labels as they are**, deliberately. Only Videos has a localized name (`profileVideosLabel`); the other five tabs — Collabs, Liked, Reposts, Lists, Comments — have none. Giving them real labels means new ARB keys across **22 locales**, which is a translation change and does not belong in an E2E repair PR. The defect is commented at the site.

## What is needed

1. ARB keys for the five missing tab names, mirrored across all locales (or added to `_knownUntranslatedDebt` per `.claude/rules/localization.md`).
2. Swap `label:` in `profile_tab_bar.dart` to `context.l10n.<key>`, leaving `identifier:` alone.
3. The liked/saved grid thumbnails need a real label too — something like "Liked video {n}" rather than the raw anchor.

`.claude/rules/accessibility.md` treats an interactive element with a misleading label as a must-fix, so this is a real defect rather than polish.

## Context

Found while repairing the Maestro suite in #6948.

realmeylisdev:
Closes #6951

## What was wrong

The profile tab bar renders **icon-only** tabs, so `Semantics(label:)` is the only accessible name a screen reader can read — and it was being handed the snake_case E2E anchor.

Measured on a live iOS build (VM-service semantics dump, another user's profile):

```
role: tabBar
  "Tab 1 of 5" → identifier: "videos_tab"    label: "videos_tab"
  "Tab 2 of 5" → identifier: "liked_tab"     label: "liked_tab"
  "Tab 3 of 5" → identifier: "reposted_tab"  label: "reposted_tab"
  "Tab 4 of 5" → identifier: "collabs_tab"   label: "collabs_tab"
  "Tab 5 of 5" → identifier: "comments_tab"  label: "comments_tab"
```

VoiceOver read **"Tab 2 of 5, liked_tab"**. The Liked grid was the same: every tile exported `liked_video_thumbnail_0…9`.

Two corrections to the issue as filed:

- **Six tabs, not five.** `videos_tab` was an anchor too. `profileVideosLabel` existed in the ARB — translated in all 22 locales — with **zero call sites**.
- **Three grids were worse than reported.** The reposts, collabs and video-reply tiles had no `Semantics` wrapper at all, so each exported a focusable, tappable node with an **empty** label. Silent, not merely mislabelled. `accessibility.md` files that as a Must Fix.

## What changed

| Commit | |
|---|---|
| `feat(l10n)` | 5 tab-name keys across all 22 locales. Reposts/Lists/Comments reuse the exact values already shipped for `metadataRepostsLabel`, `exploreTabLists`, `profileCommentsSection`. Liked and Collabs are newly translated. **No new untranslated debt** — `gen-l10n` still reports 27 per locale. |
| `refactor(profile)` | The 8 inline anchors move into `SemanticIds`, byte-identical values. |
| `fix(a11y)` tabs | `ProfileTabBar.tabs` becomes a named `ProfileTab` record with **separate `semanticId` and `label`**. One field was feeding both slots; now re-fusing them takes a deliberate edit. |
| `fix(a11y)` liked/saved | Localized label + `button: true`, matching `profile_videos_grid`, the one sibling already correct. |
| `fix(a11y)` reposts/collabs/comments | Real `Semantics` on the four silent tiles. |
| `fix(a11y)` icon scaling | **Not requested — see below.** |

### E2E anchors are unaffected

`SemanticsConfiguration.absorb` keeps a child's identifier when the parent's is empty, so the identifier still survives Material's `MergeSemantics`. Verified by reading `getSemanticsData()`; a test pins it.

⚠️ `SemanticsNode.label` / `.identifier` are the node's **pre-merge** config and report `"Tab 3 of 6"` with an *empty* identifier — misleading in both directions. The new tests read `getSemanticsData()` and say so inline.

### `_ProfileCommentCard` deliberately has no label

It is the one tile whose subtree renders wording. `Semantics(label:)` is *prepended* to the child's label rather than replacing it, so a label would make a screen reader read the comment twice. It gets `button: true` only. A test asserts the comment text appears exactly once in the merged label — adding a label back makes it fail with `Actual: <2>`.

## Not requested by the issue

The last commit swaps the tab bar's raw `SvgPicture.asset(width: 28, height: 28)` for `DivineIcon`, so the icons track system text scale (capped at 1.3×) like every other icon in the app. At the largest text size the five profile tabs were the only icons that did not grow.

#6845 was the sweep that should have caught this and missed the file, and nothing would have caught it since: `check_raw_icons_ceiling.sh` greps only Material `Icons.*`, so raw `SvgPicture` sits outside every ratchet. It lands on the same lines the label fix already touched, but **it is its own commit and can be dropped without touching the rest**.

## Review round

Two commits from @hm21, both accepted as pushed:

- **`fix(a11y): let the tab icons scale at the own-profile tab count`** — the `DivineIcon` swap above did not actually fire on the phones most users are on. Material's default `labelPadding: 16` leaves each icon `width / tabs - 32` to grow into; at 6 tabs on a 360dp screen that is exactly the unscaled 28dp, so `BoxFit.contain` letterboxed the glyph flat. Dropped to 8, which fits the 1.3× cap down to 320dp. Re-verified here by reverting the padding against the strengthened test: it fails with `Expected: a value greater than <28.0> / Actual: <28.0>`. Icon position and indicator extent do not move at 1.0× — tabs are equal-width, the icon is centred, and `indicatorSize` is `TabBarIndicatorSize.tab`. Icon *size* at 1.0× is unchanged from 360dp up, but below that the default padding was already squashing the glyph — measured 21.3dp at 320dp — and it now renders at its nominal 28dp. (Corrected during review; the original wording said nothing moved at 1.0× at any width.)
- **`test(e2e): anchor the user-profile tab asserts by identifier`** — `assertUserProfile.yaml` matched tabs on text (`"Tab N of 4\n<anchor>"`), which only worked while the anchor doubled as the accessible label. Converted to `id:` matchers, mirroring what #6948 did to `assertProfile.yaml`. All four ids do exist on another user's profile (`profileTabKinds` gives videos / liked / reposts / collabs / comments), and the `of 4` was stale — that strip is 5 tabs.

Also rebased onto `origin/main` to drop a `Merge branch 'main'` commit that had been pushed to the branch; the tree is byte-identical and both of the above keep their authorship.

## Open product question

"Collabs / Liked / Reposts / Lists / Comments" is my reading of the tab kinds and icons. Nothing in-repo, in `divine-web` (which has no profile tab strip), or in `brand-guidelines/` pins the wording. Happy to change it — the 21 translations follow whatever English lands.

## Testing

- New `profile_tab_bar_test`: merged label contains the localized name, not the anchor; changes with locale; identifier preserved; icons scale.
- Accessibility groups added to the liked, saved, reposts, collabs and comments grid tests.
- `profile_grid_test`'s 12 anchor finders migrated from `bySemanticsLabel` to `bySemanticsIdentifier`.
- Every new assertion was **mutation-tested** — each one goes red when its fix is reverted.
- `flutter analyze lib test integration_test` clean; 2220 tests pass; `golden.sh verify` unchanged; raw-icons / test-unit / untested-services ratchets green.

## Separate bug found while verifying — since fixed on `main`

**iOS could not build on `main`.** `firebase_remote_config: 6.1.2` (`mobile/pubspec.yaml`) pinned SPM `flutterfire` to `4.2.1-firebase-core-swift`; its five sibling Firebase plugins pinned `4.4.0`. Two `exact:` pins are unsatisfiable, and it was derivable from `pubspec.lock` alone, so it reproduced on any machine — not the stale-cache case `ios_build_troubleshooting.md` describes. Fixed on `main` by #7016 (`6a6a205`), which this branch is now rebased onto. Nothing outstanding here.


hm21 (approved):
Reviewed and pushed two fixes. CI green on `b8b10db`.

**`cfe0226` — the icon-scaling commit didn't actually fire on the own profile.** Material's default `labelPadding: 16` leaves each icon `width / tabs - 32` to grow into. At 6 tabs on a 360dp screen that is 28dp — exactly the unscaled size — so `BoxFit.contain` letterboxed the glyph straight back to 28×28. Measured icon width at textScale 2.0, before → after:

| screen | 6 tabs (own) | 5 tabs (other) |
|---|---|---|
| 360dp (Galaxy S) | 28.0 → **36.4** | 36.4 |
| 393dp (iPhone 15) | 33.5 → **36.4** | 36.4 |
| 412dp (Pixel) | 36.4 | 36.4 |

Dropped it to 8, which fits the 1.3× cap down to a 320dp screen. Nothing moves at 1.0×: tabs are equal-width, the icon is centred, and `indicatorSize` is `TabBarIndicatorSize.tab` so the indicator is unaffected.

The scaling test pumped 3 tabs on the default 800dp surface — enough slack that it passed against the clamped layout. It now measures 6 tabs at 360dp and fails with `Actual: <28.0>` when the padding change is reverted.

**`b8b10db` — one E2E anchor was affected.** `assertUserProfile.yaml` matched the tabs on *text* (`"Tab N of 4\n<anchor>"`), not `id:`, so it depended on the anchor being the accessible label — those four assertions could never match again. Converted to `id:` matchers, mirroring what #6948 did to `assertProfile.yaml`. Low stakes either way: the flow is held out of `suites/smoke.yaml` pending the searchUsers rewrite in #6952, and `of 4` was already stale (another user's profile shows 5 tabs).

Everything else holds up. Identifiers do survive `absorb`, the five keys are complete across all 22 locales, `profileVideosLabel` really did have zero call sites, and `check_raw_icons_ceiling.sh` greps `Icons\.` only — so removing the `SvgPicture` costs nothing.

Validation: `flutter analyze lib test integration_test` clean, 312 tests in `test/widgets/profile/` + `arb_consistency_test` pass, no goldens cover the tab bar, 16/16 CI green. Both commits are independent — revert either without touching the rest. Yours to accept, tweak or merge.

realmeylisdev:
Thanks @hm21 — both commits taken as pushed, nothing declined.

I re-derived the `labelPadding` finding rather than take it on the commit message: reverting the padding against your strengthened test gives `Expected: a value greater than <28.0> / Actual: <28.0>`. That the old test pumped 3 tabs on an 800dp surface is the part worth remembering — it was green against a layout that clamped the icon flat, so the assertion existed but the geometry never exercised it.

On the E2E change: checked the four ids resolve on someone else's profile. `profileTabKinds(isOwnProfile: false)` is videos / liked / reposts / collabs / comments, so `collabs_tab` is real there and `of 4` was the only stale part.

One thing you didn't ask for: the branch tip was a `Merge branch 'main'` commit, which `AGENTS.md` rules out. Rebased onto `origin/main` — tree byte-identical at `efb7af5`, your two commits keep their authorship and SHAs are now `b863356` / `1200484`. Also refreshed the PR body: the iOS `firebase_remote_config` note it carried is fixed on `main` by #7016, which the rebase picked up.

Green at `1200484` — 15/15, `Mobile CI` included.

mobile/lib/widgets/profile/profile_grid.dart
```diff
@@ -411,27 +413,38 @@ class _ProfileGridViewState extends ConsumerState<ProfileGridView>
     }
   }
 
-  /// The tab-bar label/icon for a given tab [kind].
-  ({String label, DivineIconName icon}) _tabPresentationFor(
-    ProfileTabKind kind,
-  ) {
+  /// The tab-bar anchor, localized name and icon for a given tab [kind].
+  ProfileTab _tabPresentationFor(ProfileTabKind kind) {
```
dcadenas, line 417: This is the only place a tab kind maps to `context.l10n`, and nothing exercised it.

Reverting one of these labels back to its `SemanticIds` anchor left the entire `test/widgets/profile` suite green at 302 passed — so the defect this PR fixes could return without a single test failing. `ProfileTabBar`'s own test builds its `ProfileTab` records by hand, so it cannot see this method.

Added a merged-`SemanticsData` assertion over all six own-profile tabs in `profile_grid_test.dart`. It pumps English and Japanese: English alone cannot tell an l10n lookup apart from a hardcoded English literal, and Japanese is the locale that differs from English on all six values — German shares "Videos", "Collabs" and "Reposts" verbatim, which would have left half the strip unguarded.

mobile/lib/widgets/profile/profile_tab_bar.dart
```diff
@@ -110,9 +123,21 @@ class _ProfileTabBarState extends State<ProfileTabBar> {
...
+          // to grow into. On the own profile (6 tabs) that is 28dp at 360dp
+          // wide — exactly the unscaled size, so the icons never actually
+          // scale. 8 fits the 1.3x cap down to a 320dp screen.
+          //
+          // Tabs are equal-width and the icon is centred, so neither icon
+          // position nor indicator extent moves. Icon *size* at 1.0x is
+          // unchanged from 360dp up, but below that the default padding was
+          // already squashing the glyph — 21.3dp at 320dp — and it now
+          // renders at its nominal 28dp.
+          labelPadding: const EdgeInsets.symmetric(horizontal: 8),
```
dcadenas, line 136: The rationale here previously ended with "at 1.0x this changes nothing on screen". That is right for icon position and indicator extent, and wrong for icon size below 360dp.

Measured on the real bar at 320dp with all six own-profile tabs and the default text scale: `labelPadding: 16` renders the glyph at 21.3dp, `labelPadding: 8` renders it at 28.0dp. So on a 320dp phone this is a visible change at ordinary text sizes, and an improvement — the glyph stops being squashed.

Rewrote the comment to say what is and is not inert, and parameterised the scaling test over 320dp and 360dp with a nominal-size assertion at each. Reverting the padding to Material's 16 now turns three tests red instead of one, including `Expected: <28.0> / Actual: <21.333…>` at 320dp.

mobile/lib/widgets/profile/profile_tab_bar.dart
```diff
@@ -3,11 +3,17 @@
...
+/// How one profile tab is presented.
+///
+/// [semanticId] is the stable, snake_case test anchor; [label] is the
+/// localized name. Keeping them as separate fields is what stops the anchor
+/// from being announced as the accessible name (#6951).
+typedef ProfileTab = ({String semanticId, String label, DivineIconName icon});
+
 /// Sticky tab bar rendering the profile's [tabs] (5 on other profiles, 6 on
-/// the own profile, which also shows Collabs).
+/// the own profile, which also shows Lists). See `profileTabKinds`.
```
dcadenas, line 16: Corrected: this said the own profile "also shows Collabs". `profileTabKinds` puts Collabs on both strips and makes **Lists** the own-profile extra — which is also what the e2e change in this PR relies on when it asserts `collabs_tab` on another user's profile.

dcadenas (approved):
♿ the tab strip and every profile grid now announce a real name — four small commits pushed on top, detail inline.

## Pushed

- A regression test for the tab-name mapping, which nothing covered.
- Coverage of the tab-icon layout at the narrowest supported width and at ordinary text scale.
- Two comment corrections where the code said something the code no longer does.

The description's claim that nothing moves at 1.0× is corrected in the body; it holds for icon position and the indicator, not for icon size below 360dp.

## Checks

- All required checks green at the pushed head.
- Locally: analyze clean, format clean, 317 tests across the profile widgets and the ARB consistency suite.
- Every assertion added here was mutation-tested — each goes red when the behaviour it pins is reverted.

realmeylisdev:
Thanks @dcadenas — all four commits taken as pushed, nothing declined.

I re-checked the Collabs/Lists correction against the source rather than take it from the commit message: `profileTabKinds` returns videos/collabs/liked/reposts/lists/comments for the own profile and videos/liked/reposts/collabs/comments for others, so Collabs is on both strips and Lists is the own-profile extra. `profile_tab_bar.dart:16` now says that.

The l10n-mapper gap is the part worth remembering: `ProfileTabBar`'s own test builds its `ProfileTab` records by hand, so the one place a tab kind meets `context.l10n` sat outside every assertion. Pumping `ja` rather than `de` matters for the same reason — `de` shares "Videos", "Collabs" and "Reposts" verbatim.

Flagging one thing since it moves your commits: the branch picked up a `Merge branch 'main'` commit from the PR page's "Update branch" button after your approval. I rebased it away onto `origin/main` — tree byte-identical at `a6a2ac7`, your four commits keep their authorship at `1283af9`, `42954af`, `fd5e81c`, `b1de813`. All checks green at the rebased head.

PR #7011, opened 2026-08-10
fix(profile): drop the extra vertical gap around the follower search field

### LINKED ISSUE #7012: fix(profile): drop the extra vertical gap around the follower search field
Auto-created from PR #7011.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7011
**Branch:** `fix/follower-search-padding`

hm21:
Closes #7012

## Description

The search field on the followers / following lists was padded `EdgeInsets.all(16)`. Its bottom padding stacked with the `EdgeInsets.all(16)` every `UserProfileTile` already carries, leaving a 32px gap before the first row, and the top padding pushed the field away from the app bar. Padding horizontally only keeps the field aligned with the rows and removes the extra vertical space.

Affects all four screens that use `SearchableFollowList`: my followers, others' followers, my following, others' following.

**Related Issue:** 

## Out of Scope

Other search surfaces keep their current spacing — `explore_view.dart` still uses `EdgeInsets.all(16)` and `inbox_view.dart` uses `fromLTRB(16, 8, 16, 0)`. Aligning all search bars on one spacing rule is a separate design decision.

## Verification

- `flutter analyze lib/widgets/profile/searchable_follow_list.dart` — clean
- `flutter test test/widgets/profile/searchable_follow_list_test.dart test/screens/followers/others_followers_screen_test.dart test/screens/following/others_following_screen_test.dart` — 7/7 passing
- Manually checked on a real Samsung Galaxy device

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/lib/widgets/profile/searchable_follow_list.dart
+++ b/mobile/lib/widgets/profile/searchable_follow_list.dart
@@ -98,7 +98,7 @@ class _SearchableFollowListState extends State<SearchableFollowList> {
     return Column(
       children: [
         Padding(
-          padding: const EdgeInsets.all(16),
+          padding: const .symmetric(horizontal: 16),
           child: DivineSearchBar(
             controller: _controller,
             hintText: context.l10n.exploreSearchHint,


mobile/lib/widgets/profile/searchable_follow_list.dart
```diff
@@ -98,7 +98,7 @@ class _SearchableFollowListState extends State<SearchableFollowList> {
     return Column(
       children: [
         Padding(
-          padding: const EdgeInsets.all(16),
+          padding: const .symmetric(horizontal: 16),
```
realmeylisdev, line 101: Dropping the vertical padding moves two gaps, and the description only covers one. Measured against the real `DiVineAppBar` (72px tall) with a tile-equivalent row:

| gap | before | after |
| --- | --- | --- |
| app bar → search field | 16px | **0px** |
| search field → first row content | 32px | 16px |

The bottom half is the fix and it's correct. The top half means the search pill's rounded edge now butts directly against the app bar with nothing between them. The app's other two search surfaces both keep some air there — `explore_view.dart` has 16px above, `inbox_view.dart` has 8px.

If flush is what design wants, ignore me. If not, `const .fromLTRB(16, 8, 16, 0)` matches inbox and still keeps the 16px below.

realmeylisdev (approved):
The padding math holds up. The 32px gap under the field was real — 16 from here plus the 16 `UserProfileTile` carries on its own `Container` — and this takes it to 16. `const .symmetric(...)` is fine too: `EdgeInsetsGeometry.symmetric` is a `const factory`, and the shorthand is already used in a dozen places under `lib/`.

One thing I left inline: the same edit also takes the gap *above* the field from 16px to 0, so it now sits flush against the app bar. That reads like a deliberate call, but it isn't in the description, so worth a yes/no from you before merge. Nothing else blocking.

PR #7013, opened 2026-08-10
feat(profile): hide a small loop total from profile visitors

rabble:
## What

Every profile leads with a **Loops** stat rendered straight from `profileStats.totalViews`, unconditionally. A brand-new creator's profile greets every visitor with `0 Loops`, in the largest stat treatment on the screen, next to Likes / Following / Followers.

That's the same discouraging number #6872 removes from the feed card — on the surface creators actually send people to, and with no "more info" tap to demote it behind.

Visitors now see the column only once the total clears `profileLoopsVisibilityFloor` (**10,000**). Owners always see their own, however small.

| viewer | total | Loops column |
|---|---|---|
| visitor | 7 | hidden |
| visitor | 10,000 | shown |
| visitor | 50,000 | shown |
| owner | 7 | shown |

The rest of the row is untouched — only the discouraging headline number is withheld, and the row already renders columns conditionally, so there's no layout special-casing.

## Why owners still see it

Bernstein et al. (CHI 2013): people badly underestimate their audience, and perceived audience drives production. Hiding a creator's own number works against the thing we want. The number is only withheld where it functions as a public verdict.

## Notes

- **`isOwnProfile` was already threaded** to `ProfileActionButtons` on the same widget, so ownership is injected into `_ProfileStatsRow` rather than read from Riverpod in the child.
- **No feature flag.** #6872 carries one because it changes the feed on every scroll; this is a single column on one surface and reverts cleanly in one commit. Say the word if you'd rather it were gated.
- **10,000 is a product call**, not a technical one — a named constant, one line to change. It is deliberately higher than the per-video floor in #6872, since a lifetime total across all of a creator's videos should clear a higher bar to read as impressive.
- **Independent of #6872** — no shared code, targets `main`, either can merge first.

## Testing

`profile_header_widget_test.dart`: visitor-below-floor hides, visitor-at-floor shows, owner-below-floor shows. One existing test (`displays stats from ProfileStats when provided`, `totalViews: 5000`, `isOwnProfile: false`) asserted the old behaviour and **failed on this change** — moved to 50,000 so it still tests what its name says.

295 profile tests green, `flutter analyze lib test integration_test` clean, all four design-system ratchets and the test-structure floor pass.

**Not done: nobody has run this on a device.** Screenshots should go on before merge.

### DIFF
--- a/mobile/lib/widgets/profile/profile_header_media.dart
+++ b/mobile/lib/widgets/profile/profile_header_media.dart
@@ -10,6 +10,14 @@ part of 'profile_header_widget.dart';
 ///
 /// A product call, not a technical one — the single value to change if the
 /// bar sits wrong.
+///
+/// Deliberately higher than `publicLoopCountFloor` in
+/// `widgets/video_feed_item/video_card_meta.dart`, which hides small counts on
+/// feed cards. A profile headline is a summary of a whole creator and carries
+/// more weight than a number beside one video, so it earns a higher bar. The
+/// two are independent product calls, not a value that drifted — do not
+/// collapse them into one constant without deciding that both surfaces want
+/// the same number.
 const int profileLoopsVisibilityFloor = 10000;
 
 /// Hero tag for the avatar ↔ lightbox shared-element flight, scoped to the
@@ -221,10 +229,17 @@ class _BannerImage extends StatelessWidget {
 /// Stats row displaying Followers, Following, Likes, and Loops with dividers.
 ///
 /// Shows a skeleton for up to [_ProfileStatsRowState._skeletonTimeout] while
-/// stats are being fetched. After the timeout the row keeps all four columns
+/// stats are being fetched. After the timeout the row keeps its columns
 /// visible but renders a `—` placeholder for each count, rather than
 /// shimmering indefinitely or collapsing the row (which would shift the
 /// surrounding profile layout).
+///
+/// One shift is unavoidable and accepted: whether Loops is shown depends on
+/// the count, which is not known until stats load. A visitor to a profile
+/// under [profileLoopsVisibilityFloor] therefore sees four skeleton columns
+/// resolve to three. Reserving the slot only for owners would move that shift
+/// onto popular profiles instead of removing it, so the loading state stays
+/// uniform and the settle is where the column count changes.
 class _ProfileStatsRow extends StatefulWidget {
   const _ProfileStatsRow({
     required this.userIdHex,
--- a/mobile/lib/widgets/profile/profile_header_media.dart
+++ b/mobile/lib/widgets/profile/profile_header_media.dart
@@ -1,5 +1,17 @@
 part of 'profile_header_widget.dart';
 
+/// Smallest lifetime loop total a profile shows to visitors.
+///
+/// Below this the Loops column is omitted for everyone but the owner: a small
+/// headline number on a new creator's profile discourages the visitor and
+/// tells them nothing useful. Owners always see their own total, since
+/// correcting a creator's underestimate of their audience is what keeps them
+/// posting.
+///
+/// A product call, not a technical one — the single value to change if the
+/// bar sits wrong.
+const int profileLoopsVisibilityFloor = 10000;
+
 /// Hero tag for the avatar ↔ lightbox shared-element flight, scoped to the
 /// user. A global tag would let two profile headers with the same tag in one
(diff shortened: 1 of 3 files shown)


mobile/lib/widgets/profile/profile_header_media.dart
```diff
@@ -257,15 +281,20 @@ class _ProfileStatsRowState extends State<_ProfileStatsRow> {
...
+    // discouraging headline number. Owners keep theirs, and a total large
+    // enough to impress still leads the row.
+    final loopsAreVisible =
+        totalViews != null &&
+        (widget.isOwnProfile || totalViews >= profileLoopsVisibilityFloor);
+    final hasLoops = loopsAreVisible;
 
     final l10n = context.l10n;
     final columns = <Widget>[
       if (hasLoops || isLoading)
```
realmeylisdev, line 310: The row settles from 4 columns to 3 here once stats arrive.

While `profileStats` is null this branch is true via `isLoading`, so a skeleton Loops column renders. When the real total lands below the floor for a visitor it drops out — the other three are `Expanded`, so they go from 25% to 33% each and a divider goes with it. Skeletonizer's `AnimatedSwitcher` cross-fades the swap, so it reads as a fade rather than a snap, but the row does resettle at a different geometry.

I pumped it rather than eyeballing: loading, then loaded with `totalViews: 7` as a visitor, gives 4 columns then 3. A warm cache mounts straight to 3 and never flickers — so this is the cold first visit, which is the "visitor lands on a new creator's profile" case the change is aimed at.

Flagging it because the doc comment just above says the row keeps all four columns "rather than shimmering indefinitely or collapsing the row (which would shift the surrounding profile layout)".

There's no free fix — anything here changes what the skeleton shows. Skipping the Loops skeleton for visitors is the cheapest: below-floor becomes 3 → 3, and only the above-floor case gains a fourth column, which reads as a reveal rather than a collapse.

mobile/lib/widgets/profile/profile_header_media.dart
```diff
@@ -257,15 +281,20 @@ class _ProfileStatsRowState extends State<_ProfileStatsRow> {
...
     final hasLikes = widget.profileStats?.totalLikes != null;
-    final hasLoops = widget.profileStats?.totalViews != null;
+    final totalViews = widget.profileStats?.totalViews;
+    // A visitor landing on a new creator's profile should not be met by a
+    // discouraging headline number. Owners keep theirs, and a total large
+    // enough to impress still leads the row.
+    final loopsAreVisible =
+        totalViews != null &&
+        (widget.isOwnProfile || totalViews >= profileLoopsVisibilityFloor);
+    final hasLoops = loopsAreVisible;
```
realmeylisdev, line 306: `hasLoops` is just a second name for `loopsAreVisible` — one of them can go.

mobile/test/widgets/profile/profile_header_widget_test.dart
```diff
@@ -1277,6 +1278,80 @@ void main() {
...
+          suppliedProfile: testProfile,
+          profileStats: profileStats,
+          videoCount: 2,
+        ),
+      );
+      await tester.pumpAndSettle();
+
+      // The rest of the row still renders; only the discouraging headline
+      // number is withheld.
+      expect(find.text('Loops'), findsNothing);
```
realmeylisdev, line 1379: `find.text('Loops')` and `find.text('Likes')` hardcode English. `localization.md` asks for the key resolved off `AppLocalizations`, so the test survives a copy change and fails loudly if the widget ever stops reading from l10n — and this file already does that a few lines up (line 1246, `lookupAppLocalizations(const Locale('en'))`).

`l10n.profileLoopsLabel` / `l10n.profileLikesLabel` here.

realmeylisdev (commented):
Traced the gate end to end. Ownership holds up — `OtherProfileScreenRouter` redirects self-visits to the own-profile tab, so the hardcoded `isOwnProfile: false` in `other_profile_screen.dart` can't strip an owner of their own number. The divider interleave is index-based too, so the 3-column row lays out cleanly — no special-casing needed, as you said.

One call to make before merge: the row settles from 4 columns to 3 when stats land, on the cold first-visit path this specifically targets. Detail inline.

On your flag question — I'd gate it. #6872 is still open and ships its floor behind `FeatureFlag.videoCardPostDate`, default off. Unflagged here means profiles hide small totals for everyone while feed cards keep showing them until that flag flips: same hypothesis, two rollout states.

Two nits inline. Nothing blocking.

Issue #648, opened 2025-12-15
feat: Auto trigger cache cleanup


B0berman:
Check for expired data in cache and clean it on app launch and then periodically

PR #7014, opened 2026-08-10
feat(badges): make and hand out Nostr badges without leaving the app

### LINKED ISSUE #7022: feat(badges): make and hand out Nostr badges without leaving the app
Auto-created from PR #7014.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7014
**Branch:** `feat/native-badges`

hm21:
Closes #7022

## Description

https://github.com/user-attachments/assets/1f07007f-7627-4a6a-b5aa-246b808aeeff

Badges got popular on Discord, so this brings creating and handing them out into the mobile app instead of sending people to the website. The site was never pleasant for making a badge on a phone, and the repeated logging in to sign each event made it worse — every publish meant another round through the signer bridge. Doing it natively means you are already signed in, and the whole flow is one screen stack.

The badge screen used to be a dashboard plus a link into the `badges.divine.video` WebView. It is now a native flow:

- **Tabbed dashboard** — Awarded / Created / Issued, using the same TabBar style as Library and Explore.
- **Badge editor** — name, identifier, description, artwork. Creating and editing share one path, because badge definitions are addressable (NIP-33): the `d` identifier decides which event gets replaced.
- **Badge detail** — artwork, description, who made it, everyone who received it, plus share / edit / delete for the owner.
- **Award flow** — pick people through the existing user picker, or paste npubs / nprofiles / hex keys.
- **Deep links** — `/badges/new`, `/badges/b/:naddr`, `/badges/b/:naddr/edit`, `/badges/b/:naddr/award`. A malformed `naddr` falls back to the dashboard instead of an empty screen.
- **Profile badge sheet** — "Make your own badge" now pushes the in-app editor instead of opening `badges.divine.video` in the Nostr app WebView. The label named the website, so it changes with the destination across all 22 locales, and the ARB test that pinned the host now asserts the opposite.

### Why some things work the way they do

**Artwork is mandatory, enforced in the repository.** A badge without an image renders as a bare initial everywhere it appears. `saveDefinition` throws on empty artwork rather than trusting the form, and the crop step is the same `pro_image_editor` 1:1 / 1024px path the profile picture uses, so avatars and badge artwork come out identical.

**The missing-artwork message waits for the first Publish press.** An empty form should not open already scolding you for something you have not skipped yet. Publish stays pressable once the badge has a name, and pressing it is what surfaces the message.

**Text fields stay editable during an artwork upload.** Only the artwork and Publish controls lock — a second upload would race the first, and publishing mid-upload would report artwork that is on its way as missing.

**Creating a badge under an identifier you already used is blocked.** Addressable events mean it would silently replace that badge, including for everyone already holding an award for it. The deliberate way to do that is Edit. Two people can share an identifier without any conflict, since the pubkey is part of the address.

**Deleting issues one NIP-09 request covering the definition and every award for it.** An award that outlives its definition renders as a nameless badge. The confirmation says plainly what a deletion request cannot reach: relays may ignore it, and a recipient's profile badge list is their event to change.

**Rejecting an award is recoverable.** Reject is local and publishes nothing, so a mistap used to strand the badge until the issuer awarded it again. There is now an undo snackbar and a collapsible Hidden section.

**The Issued tab groups per badge and per recipient, newest award wins.** Awarding someone a second time is normal — their first award may have been dismissed — and listing both left a row that could never resolve. NIP-58 has no reject event, so "declined" and "has not seen it yet" remain indistinguishable by design.

### Not requested by this feature

Two shared widgets were fixed along the way, in their own commit:

- **`UserPickerSheet`** — the confirm button was gated on `maxCount`, so the `onUserToggled` shape (used by the award flow and by the people-list sheet) had no way to say done. The only exit was the close button, which reads as discarding the picks.
- **`UserProfileTile`** — the follow button was a hand-rolled reimplementation of `DivineIconButton`. Swapped to the component. **Visible change:** the design-system button draws a shadow by default, the hand-rolled one did not, so follow buttons in Followers / Following / Engagement lists gain one.

**Related Issue:** 

## Out of Scope

- The `badges.divine.video` Worker keeps everything that cannot move: the Diviner cron, the issuer key, D1 idempotency, Discord announcements, and the public `/b/:coord` page with its OG preview. Someone without the app still needs to see a shared badge.
- `badges.divine.video` stays in the Nostr apps directory. Only the profile badge sheet's call to action was repointed at the in-app editor; the directory entry is a separate product call.
- App Links for `badges.divine.video/b/…` need AASA / assetlinks on the Worker side. The route is ready for it.
- No separate thumbnail control in the editor. An existing custom thumb from the web client is carried through an edit rather than dropped.

## Verification

- `badge_repository`: 118 tests, 581/581 lines covered (the package's 100% gate)
- App: 63 badge tests (blocs, editor, detail, award, dashboard), plus the user-picker, followers, following, engagement, and profile-header suites that render the changed shared widgets
- `flutter test test/l10n/arb_consistency_test.dart` — 49 new keys translated into all 21 non-English locales; five keys left dead by the removed intro panel were deleted from every locale rather than abandoned
- `dart format` clean; all six ratchets pass (raw TextStyle, raw colors, Material buttons, raw dialogs, untested-services floor, test-unit structure)
- Manually verified on a real Samsung Galaxy: create, edit, award, accept, reject and undo, share, and delete, from both the badges dashboard and the profile badge sheet

### Known relay-side gap (not fixed here)

Deleting a badge within seconds of creating it is refused with `blocked: delete not authorized`. `relay.divine.video` authorizes a NIP-09 kind 5 for a badge only by resolving its `e` targets to stored owned events; the `a`-coordinate path added by divinevideo/divine-funnelcake#648 accepts video kinds only, so a badge definition the relay has just accepted is not yet deletable. No client-side event shape avoids it — omitting the `e` tags is refused by a different branch. The app now names the refusal and tells the user to retry; the fix belongs in funnelcake, extending #648 to any owned addressable coordinate.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mobile/packages/badge_repository/lib/src/badge_repository.dart
```diff
@@ -242,30 +404,431 @@ class BadgeRepository {
...
+    return List<CreatedBadgeViewData>.unmodifiable(created);
+  }
+
+  /// Loads one badge's definition, its awardees, and the viewer's relation
+  /// to it.
+  ///
+  /// Only awards published by the badge's own issuer count, so a third party
+  /// cannot inject awardees into someone else's badge. When more than
+  /// [recipientCheckLimit] people were awarded the badge, acceptance is only
+  /// resolved for that many — the viewer is always among them.
```
realmeylisdev: **should-fix** — the cap does more than this line says. `_loadIssuedBadges` runs every award's `p` tags through `cappedRecipients` *before* building `byRecipient`, and `loadBadgeDetail` builds `recipients` from the same capped `checked` set, so recipients past the limit are dropped from the list entirely rather than just left unresolved.

One award naming 60 people, driven through `loadDashboard` + `loadBadgeDetail`:

```
awardPTags=60  createdRecipientCount=60  issuedRecipients=50  detailRecipients=50
```

So the Created tab says "Awarded to 60 people" and the Issued tab says 50 for the same badge, and the detail page lists 50 of 60. The award screen can produce a single kind 8 with that many `p` tags straight from the paste-keys field.

Either resolve acceptance for the first N and still list the rest, or page them. If the cap stays as-is, `_IssuedBadgeCard` shouldn't count `badge.recipients.length`, and this doc line needs to say the list is truncated.

mobile/packages/badge_repository/lib/src/badge_repository.dart
```diff
@@ -156,21 +325,18 @@ class BadgeRepository {
...
 
     final viewData =
         [
-          for (final award in visibleAwards)
+          for (final award in awards)
             BadgeAwardViewData(
               award: award,
               definition: definitions[award.definitionCoordinate],
               isAccepted: _containsAward(profileBadges, award),
+              isHidden: dismissedAwardIds.contains(award.event.id),
```
realmeylisdev, line 358: **should-fix** — `IssuedBadgeViewData` collapses to the newest award per (badge, recipient) precisely because a re-award is normal, but the recipient side never does the same thing.

Two awards of the same badge to the same person, viewer having accepted the older one:

```
awardedCards=2 [VIP/false + VIP/true]   profileEntriesAfterAccept=2 [VIP,VIP]
```

So the Awarded tab shows the badge twice (one "Accepted", one "Not accepted"), and accepting the second appends a *second* `a`/`e` pair for the same coordinate to kind 10008, because `acceptAward` only skips an exact coordinate + award-id match. The profile badge sheet then renders the badge twice.

The same rule the Issued tab already uses would fix both halves: newest award per coordinate here, and drop any existing ref for that coordinate in `acceptAward` before appending.

mobile/lib/blocs/badges/badge_detail_cubit.dart
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  Future<void> _runAction(
+    BadgeDetailActionStatus actionStatus,
+    Future<void> Function() action,
+  ) async {
+    emit(state.copyWith(actionStatus: actionStatus));
+    try {
+      await action();
+      final detail = await _repository.loadBadgeDetail(state.coordinate);
```
realmeylisdev, line 137: **should-fix** — the award screen builds its own `BadgeDetailCubit` without calling `load()`, and its listener pops as soon as `award` reports `completed`, but this reload still runs first.

`loadBadgeDetail` fetches the definition, then the awards, then `_latestProfileBadges` for up to 50 recipients, and each of those is 2 `queryEvents` calls. Awarding 50 people means the button spins through roughly a hundred relay round trips to build a `BadgeDetailData` that is discarded with the route a moment later.

Worth letting the award path skip the reload (a dedicated method, or a flag on `_runAction`).

mobile/packages/badge_repository/lib/src/badge_repository.dart
```diff
@@ -242,30 +404,431 @@ class BadgeRepository {
...
+    }
+
+    final thumbnailUrl = draft.thumbnailUrl.trim();
+    final event = await _signAndPublish(
+      kind: EventKind.badgeDefinition,
+      label: 'badge definition',
+      tags: [
+        ['d', identifier],
+        ['name', name],
+        ['description', draft.description.trim()],
```
realmeylisdev: **nit** — `thumb` is conditional two lines down but `description` isn't, so a badge published without one carries `["description", ""]`. That parses back as `''` rather than null, and `_AwardedBadgeCard` gates on `award.definition?.description != null`, so every award card for such a badge gets an empty `Text` plus the 12px spacer above it. `_BadgeHero` already uses `isNotEmpty` for the same value. Making this `if (description.isNotEmpty)` keeps the empty tag out of the event entirely.

mobile/lib/blocs/badges/badges_cubit.dart
```diff
@@ -62,6 +70,8 @@ class BadgesCubit extends Cubit<BadgesState> {
           status: BadgesStatus.loaded,
           awarded: dashboard.awarded,
           issued: dashboard.issued,
+          created: dashboard.created,
+          hidden: dashboard.hidden,
```
realmeylisdev, line 80: **nit** — the success path leaves `actionStatus` untouched, so after a failed accept the "Could not update badge" note survives a pull to refresh, while the `catch` below *does* reset it to `idle`. That reads backwards. Clearing it here on success would match. `BadgeDetailCubit._load` has the same shape.

mobile/lib/l10n/app_en.arb
```diff
@@ -6019,25 +6019,104 @@
...
+        "type": "int"
+      }
+    }
+  },
+  "badgeEditorCreateTitle": "New badge",
+  "badgeEditorEditTitle": "Edit badge",
+  "badgeEditorNameLabel": "Name",
+  "badgeEditorNameHint": "Scene Stealer",
+  "badgeEditorIdentifierLabel": "Identifier",
+  "badgeEditorIdentifierHelp": "Part of the badge's address, so it stays put once the badge exists.",
```
realmeylisdev, line 6119: **nit** — this doesn't fit the one-line helper slot. On both the Galaxy in your recording and an iPhone 17 Pro Max sim it renders as "Part of the badge's address, so it stays put once the badge e…", so the half that carries the meaning is the half that gets cut, and the longer locales will be worse. Either shorten it or give the field a two-line helper.

mobile/packages/badge_repository/lib/src/badge_repository.dart
```diff
@@ -242,30 +404,431 @@ class BadgeRepository {
...
+            left.definition.event.createdAt,
+          ),
+        );
+    return List<CreatedBadgeViewData>.unmodifiable(created);
+  }
+
+  /// Loads one badge's definition, its awardees, and the viewer's relation
+  /// to it.
+  ///
+  /// Only awards published by the badge's own issuer count, so a third party
```
realmeylisdev, line 537: **pre-existing, not this PR** — but worth knowing while you're in here, because this claim is true for `loadBadgeDetail` (`_queryAwardsForCoordinate` filters `authors: [coordinate.pubkey]`) and not for the Awarded tab.

`_queryAwardsForRecipient` is `kinds:[8], p:[me]` with no author filter and only a `recipientPubkeys.contains` check afterwards, so anyone can publish a kind 8 pointing at someone else's coordinate and it lands in your Awarded tab carrying the real badge's name and artwork. Accepting it writes that award id into your kind 10008, and `loadAcceptedBadgesForProfile` doesn't check the issuer either, so it renders on the profile. divine-web's `validateBadgeAward` does check `awardEvent.pubkey === definition.issuerPubkey`, so mobile is the lenient one.

Second pre-existing one, same area: divine-web reads profile badges from kind **30008** only (`BADGE_KINDS.PROFILE_BADGES` in `src/lib/badges.ts`), while `_publishProfileBadges` writes kind **10008**. Mobile reads both, web doesn't. Now that the app is the main way people accept a badge, an accept made here won't show on the web profile. Probably a funnelcake/web ticket rather than something for this PR.

realmeylisdev (dismissed):
checked the nip-58 and nip-09 shapes against the specs, and the relay-side delete gap you describe in the body is exactly what funnelcake does, so that part holds up. ran it on an ios sim against prod too, dashboard/detail/editor all behave.

two i'd fix first, both repro'd locally with a throwaway test: issued + detail silently stop the recipient list at 50 per award while created counts all of them, and awarding the same badge twice stacks a duplicate card and a duplicate a/e pair in kind 10008. rest is nits plus one pre-existing heads up, all inline.

also needs a rebase, the branch is behind #7016 so an ios build fails on this base. the rebase itself is clean, i ran it.

mobile/packages/badge_repository/lib/src/badge_repository.dart
hm21: Fixed in 01494d5265, both halves, using the issued-tab rule.

One carve-out on your `acceptAward` suggestion: displacing unconditionally broke `acceptAward keeps an already-accepted award and preserves relay hints` — rebuilding the ref drops its relay hint. Accepting the award that is already pinned is now a no-op, and the displace only runs when the coordinate matches and the award id does not.

mobile/lib/blocs/badges/badge_detail_cubit.dart
hm21: Fixed in 01494d5265 — `award()` passes `reload: false`, and `verifyNever(() => repository.loadBadgeDetail(any()))` pins it so it cannot creep back.

mobile/lib/blocs/badges/badges_cubit.dart
hm21: Fixed in 01494d5265, in both cubits, with a regression test on the refresh path.

mobile/lib/l10n/app_en.arb
hm21: Fixed in 385e8f461d, but by giving `DivineTextField` a `helperMaxLines` and passing 2, rather than shortening the string — the copy is already translated into 22 locales and editing the English would drift all of them for a layout problem.

mobile/packages/badge_repository/lib/src/badge_repository.dart
hm21: Fixed in 01494d5265. Everyone awarded is listed now — the limit only bounds the acceptance lookups. A recipient past it carries `isAccepted: null`, and `BadgeRecipientRow` renders no pill for null rather than claiming "Waiting for recipient", so Created and Issued agree again. Doc comment corrected to say so.

mobile/packages/badge_repository/lib/src/badge_repository.dart
hm21: Fixed in 01494d5265 — the tag is omitted when the description is empty, same shape as `thumb`.

mobile/packages/badge_repository/lib/src/badge_repository.dart
hm21: First one: fixed here anyway, in 01494d5265. The coordinate already encodes the issuer, so `_queryAwardsForRecipient` can drop awards whose author is not the badge owner at zero extra relay cost. Applied to `loadAcceptedBadgesForProfile` too, since a profile badge list is its owner's event and can reference an award nobody else stands behind.

Second one: the facts are right but the direction is the other way round. NIP-58 defines the Profile Badges event as kind **10008**, and says an earlier version used 30008 with `d=profile_badges`, which "clients should treat as equivalent and migrate to the new format". Mobile writes 10008 and reads both, so mobile is the spec-current side; divine-web parsing only 30008 is the stale one. Agreed it is a web ticket, not this PR.

hm21:
Thanks — all seven checked out against the code, nothing declined. Pushed as `01494d5265`, `385e8f461d`, `caf82dda46`.

**Fixed:** recipient cap (everyone is listed; acceptance resolved for the first N, `null` past it, and a null status renders no pill) · newest award per badge on the Awarded tab + `acceptAward` displacing an earlier ref for the same coordinate · `award()` no longer reloads a detail the route discards · empty `description` tag dropped · `actionStatus` cleared on a successful reload in both cubits · identifier helper given `helperMaxLines: 2` instead of shortening copy that is already in 22 locales.

**One correction to your suggestion:** displacing unconditionally in `acceptAward` broke `acceptAward keeps an already-accepted award and preserves relay hints` — rebuilding the ref drops the relay hint. Re-accepting the award already pinned is a no-op now; the displace only runs when the coordinate matches and the award id does not.

**Pre-existing #1 fixed anyway:** the coordinate encodes the issuer, so filtering awards by it costs nothing. **Pre-existing #2 is the other way round:** NIP-58 makes 10008 the Profile Badges kind and calls 30008/`d=profile_badges` the deprecated form clients should migrate off. Mobile is spec-current; divine-web reading only 30008 is the stale side. Web ticket.

**Not requested by review** (`caf82dda46`): adding a test file reshuffles `select_test_shard.sh`, which changed shard 0`&`s bundle, and shard 0 then failed twice on `VideoEditorProvider deferred file cleanup` — on exactly the two tests in that group that assert after a single bare `pumpEventQueue()` while `onDispose` reaps with `unawaited()`. They now use the group`&`s own `expectFileReaped()` poll, like their sibling already did. Test-only; the fire-and-forget teardown is intended.

Still open: the rebase onto `main` for #7016. Confirmed — this branch pins `firebase_remote_config: 6.1.2`, main has `6.1.4`. Holding off because another session is working in the same checkout right now and it ends in a force-push.

All 18 checks green.

hm21:
Rebased onto `main` (`412199263e`) — clean, no conflicts. Picks up #7016, so `firebase_remote_config` is `6.1.4` here now and the iOS build resolves. Also brings in #7011 and #6964, which touch the follower/following lists; no overlap with the `UserProfileTile` inset change, and those suites pass.

All 18 checks green on the rebased head.

NotThatKindOfDrLiz:
@hm21 I pushed two takeover commits to close the re-award acceptance hole:

- `fix(badges): treat re-awards as accepted by coordinate`
  - awarded/detail/issued acceptance now follows the badge coordinate once awards are collapsed newest-per-coordinate
  - `removeAward` removes the accepted coordinate, so a newer visible re-award can still unpin an older accepted pair
  - clarified the missing-award tradeoff in `loadAcceptedBadgesForProfile`
  - added regressions for awarded list, badge detail, coordinate removal, and missing award refs
- `test(badges): satisfy badge repository analyze`
  - removes redundant default args flagged by strict package analyze

Verification:
- local `flutter analyze lib test integration_test` clean
- local `flutter test --coverage` from `mobile/packages/badge_repository` passed
- local `flutter test test/blocs/badges test/screens/badges` passed
- GitHub checks are green on `d8a5ca8f3`, including Badge Repository CI, Divine UI CI, Mobile CI, preview, and mergeability

Please review the takeover commits and merge when you are comfortable with them.

NotThatKindOfDrLiz (approved):
Approved after takeover.

I verified the remaining blocker from the re-review and pushed the fix. The awarded/detail paths now treat an already-pinned badge coordinate as accepted even if the visible row is a newer re-award, and removal unpins the coordinate rather than the visible award id only. Regression coverage covers the awarded list, badge detail, remove path, and the explicit missing-award tradeoff.

Local verification and GitHub CI are green on `d8a5ca8f3`.

PR #7018, opened 2026-08-10
test(recorder): cover capture mode with Maestro E2E

### LINKED ISSUE #7019: test(recorder): cover capture mode with Maestro E2E
Auto-created from PR #7018.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7018
**Branch:** `test/maestro-recorder-capture`

hm21:
Closes #7019

## Description

Maestro E2E coverage for the video recorder, starting with **capture** mode. The other four modes (classic, stop-motion, lip-sync, upload) are untouched.

`flows/captureModeFlow.yaml` runs: open the recorder from the feed → drive the control rail → record a clip → delete it → close → sign the device out.

Three things in here are non-obvious enough to call out:

**Identifiers, not copy.** Capture mode was only addressable by `camera_record_button` and the mode wheel. Everything else — close, next, delete-clip, library, and all five rail controls — had to be selected by localized strings, which is precisely what left this suite broken for six months the last time product moved copy. Each new identifier ships with a widget test, because the existing label assertions stay green when an identifier is dropped and nothing else would catch it.

**Flash and stabilization are guarded, not skipped.** Both are disabled outright when the active lens has no flash unit (`hasFlash`) or reports a single stabilization mode (natively `modes.size > 1`). Driving them unguarded would make the test pass or fail on the phone rather than on the code, so they sit behind `enabled: true` and print `SKIPPED` where the hardware lacks the feature — visible in the run output rather than silently counted as covered. The stabilization options are matched by alternation rather than a fixed label because Android derives the list from the camera's `CONTROL_AVAILABLE_VIDEO_STABILIZATION_MODES`; only `off` is guaranteed.

**The permission relaunch deliberately skips `clearState`.** The flow relaunches with `permissions: all: allow` instead of tapping through the native dialog, whose buttons are OS-localized copy. It does not clear state, because on Android `pm clear` wipes the encrypted preferences holding the Nostr key and would sign the account straight back out mid-flow. (The iOS keychain survives it — the platforms disagree, so the flow takes the option that works on both.)

Also fixes two things I walked into while doing this: `brew install maestro` now resolves to an unrelated cask ("Maestro, AI agent command center", ~680 MB) that ships no CLI, and the suite's long-standing requirement for an English app UI was nowhere written down — on a German phone the very first assert fails and reads like a regression.

**Related Issue:** 

## Out of Scope

- **Not added to `smoke.yaml`.** Capture mode needs real camera hardware; without it the bloc never reports `isCameraInitialized`, the record button stays disabled, and the recording test hangs on its first wait. The smoke lane runs on the iOS Simulator, which has no camera.
- **The other four recorder modes.**
- **`assertLogin.yaml` asserting the three legal links by English copy.** It is the shared entry-point assert for the whole suite; making it locale-proof is a separate change with blast radius beyond this PR.
- **A `semanticIdentifier` on `VineBottomSheetPrompt`.** The one-shot "Why six seconds?" sheet is still dismissed by copy. Adding the hook means a `divine_ui` change and that package's 100% coverage gate — the same call `removeKeys.yaml` already makes for its confirmation sheet.
- **Untranslated German copy on the welcome screen.** Noticed while debugging: `authTermsPrefix`, `authTermsAgeAuthorizationCta` and `authTermsAfterAgeAuthorization` are still English in `app_de.arb` while `authTermsOfService` and `authTermsAnd` are translated, so the legal line renders as a mixed-language sentence. Not touched here.

## Verification

`flutter analyze lib test` clean. `flutter test test/widgets/video_recorder/ test/screens/video_recorder_screen_test.dart` — 299 passing. `check_refs.sh` and `maestro check-syntax` green on all eight new files.

Run on a real **Samsung Galaxy SM-S942B** (Android, back lens), green end to end:

| Step | Result |
|---|---|
| Open recorder, full control set, close back to feed | pass |
| Aspect ratio Vertical → Square → Vertical | pass |
| Timer Off → 3 seconds → 10 seconds → Off | pass |
| Camera Back → Front → Back | pass |
| Flash Off → On → Auto → Off | pass (guard open) |
| Stabilization Off → other mode → Off | pass (guard open) |
| Record a clip, next + delete appear | pass |
| Delete the clip, empty state restored | pass |
| `removeKeys` teardown | pass |
| Session survives the permission relaunch | pass |

The one composition not run as-is on hardware is `loginFreshInstall` at the head of the flow: it fails on a German device at `assertLogin`, which is the pre-existing locale limitation this PR documents. Its steps were exercised through an equivalent harness that skipped `clearState`, and every step downstream of it ran green.

No visual changes — the production diff adds `Semantics(identifier:)` values only, so there is nothing to screenshot.

## Type of Change

- [x] 📝 Documentation

### DIFF
--- a/mobile/e2e/maestro/tests/captureModeControls.yaml
+++ b/mobile/e2e/maestro/tests/captureModeControls.yaml
@@ -161,9 +161,11 @@ tags:
             id: camera_stabilization_button
             text: ".*Off.*"
           timeout: 10000
-      # Restore. Safe even if the barrier ever stopped blocking semantics: the
-      # rail button no longer reads "Off" at this point, so ".*Off.*" can only
-      # be the sheet's own row.
+      # Restore. The bare ".*Off.*" leans on the barrier blocking the rail
+      # underneath (ModalBarrier wraps itself in BlockSemantics) — it is not
+      # independently safe. The stabilization button no longer reads "Off"
+      # here, but the flash and timer controls both do, and picking the timer
+      # would arm the countdown that step 2 exists to clear.
       - tapOn:
           id: camera_stabilization_button
       - tapOn: ".*Off.*"
--- a/mobile/e2e/maestro/tests/captureModeControls.yaml
+++ b/mobile/e2e/maestro/tests/captureModeControls.yaml
@@ -1,0 +1,178 @@
+appId: co.openvine.app
+tags:
+  - recorder
+---
+#TC to drive the capture-mode control rail,
+# given the recorder is open on an idle capture viewfinder with no clips,
+# validate each control actually changes recorder state, and leave every one
+# of them back on its default.
+#
+# The assertions read each control's Semantics `value`, which Flutter folds
+# into the Android content description / iOS accessibilityValue. That is the
+# only readable evidence a tap landed: the controls are icon-only, and what
+# they change — flash mode, sensor crop, active lens — is invisible to a UI
+# driver. Selectors pin `id` as well as `text` because "Off" is the value of
+# both the flash and the timer control.
+#
+# Two preconditions beyond an open viewfinder:
+# - **No clips.** The aspect-ratio control is disabled once the session holds
+#   one, because clips of mixed ratios cannot share an editor timeline.
+# - **Real camera hardware**, for the lens switch. See captureModeRecordClip.
+#
+# Flash and stabilization are driven behind an `enabled` guard, because both
+# are disabled outright on hardware that lacks the feature — no flash unit on
+# the active lens, or a camera that reports a single stabilization mode. An
+# unguarded tap-and-assert there would go green or red on the phone rather
+# than on the code. A guarded block prints "SKIPPED" in the run output, so a
+# skipped control is visible rather than silently counted as covered.
+#Steps:
+# 1. Aspect ratio: Vertical -> Square -> Vertical
+# 2. Timer: Off -> 3 seconds -> 10 seconds -> Off
+# 3. Camera: Back -> Front -> Back
+# 4. Flash (if present): Off -> On -> Auto -> Off
+# 5. Stabilization (if supported): Off -> some other mode -> Off
+
+# 1. Aspect ratio flips between two states.
+- assertVisible:
+    id: camera_aspect_ratio_button
+    text: ".*Vertical.*"
+- tapOn:
+    id: camera_aspect_ratio_button
+- extendedWaitUntil:
+    visible:
+      id: camera_aspect_ratio_button
+      text: ".*Square.*"
+    timeout: 10000
+- tapOn:
+    id: camera_aspect_ratio_button
+- extendedWaitUntil:
+    visible:
+      id: camera_aspect_ratio_button
+      text: ".*Vertical.*"
+    timeout: 10000
+
+# 2. Timer cycles through three states and wraps.
+- assertVisible:
+    id: camera_timer_button
+    text: ".*Off.*"
+- tapOn:
+    id: camera_timer_button
+- extendedWaitUntil:
+    visible:
+      id: camera_timer_button
+      text: ".*3 seconds.*"
+    timeout: 10000
+- tapOn:
+    id: camera_timer_button
+- extendedWaitUntil:
+    visible:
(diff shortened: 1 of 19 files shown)


mobile/e2e/maestro/utils/openCaptureMode.yaml
```diff
@@ -0,0 +1,40 @@
+appId: co.openvine.app
+---
+#Utility: given the user is signed in and on a feed surface, open the recorder
+#and leave it on a ready capture-mode viewfinder.
+#
+# No launchApp, so this runs against whatever is already on screen — see the
+# "Iterating" section of e2e/maestro/README.md.
+#
+# It asserts Capture rather than selecting it. The recorder restores the
```
realmeylisdev, line 9: Reproduced on the iOS Simulator: with `camera_last_used_recorder_mode` left on `classic`, the recorder reopens in Classic and this util fails on `id: camera_mode_capture, selected` before anything else runs. `loginFreshInstall`'s `clearState` hides it for a top-down run, so only standalone runs see it — and standalone is what the header above recommends.

Note rather than a fix, because selecting the mode needs a wheel walk rather than a tap. Your call if you'd prefer that instead.

mobile/e2e/maestro/tests/captureModeControls.yaml
```diff
@@ -0,0 +1,180 @@
...
+      - extendedWaitUntil:
+          visible: ".*(Standard|Cinematic|Preview Optimized|Low Latency|Auto).*"
+          timeout: 10000
+      - tapOn: ".*(Standard|Cinematic|Preview Optimized|Low Latency|Auto).*"
+      - extendedWaitUntil:
+          notVisible:
+            id: camera_stabilization_button
+            text: ".*Off.*"
+          timeout: 10000
+      # Restore. The bare ".*Off.*" leans on the barrier blocking the rail
```
realmeylisdev, line 164: The previous note said this stayed unambiguous even without the barrier, because the rail button no longer reads "Off". That's true of the stabilization button, but the timer was restored to Off in step 2 and the flash in step 4 — so the bare selector really does depend on `ModalBarrier`'s `BlockSemantics` (which it does have; verified in the SDK).

Worth being straight about, since a wrong match here would land on the timer and arm exactly the countdown step 2 exists to clear.

mobile/lib/constants/semantic_ids.dart
```diff
@@ -22,6 +22,22 @@ abstract class SemanticIds {
   static String cameraMode(String mode) => 'camera_mode_$mode';
   static const String cameraRecordButton = 'camera_record_button';
 
+  /// Recorder chrome. Close, next and delete-clip come from the capture
```
realmeylisdev, line 25: `VideoRecorderLipSyncStack` builds `VideoRecorderCaptureStack`, so close/next/delete carry these ids in lip-sync too. And `camera_library_button` is wider still — it lives in `VideoRecorderBottomBar`, which `video_recorder_screen` renders outside the mode switch, so it's present in every mode including upload.

Matters because this file is what someone reads before assuming an id is unique to one mode.

realmeylisdev (approved):
Nice suite. The guard reasoning holds up — `isVideoStabilizationSupported` really is `modes.count > 1` on both platforms (`CameraController.kt:2508`, `CameraController.swift:2625`), so a passing guard genuinely guarantees a non-off mode in the sheet.

Ran it against an iOS Simulator (no camera). `assertCaptureMode` passes all 11 assertions, and `captureModeControls` clears aspect ratio and timer before dying on the lens switch — exactly where you said it would. No framework exceptions in the console throughout.

One thing bit me: `openCaptureMode` asserts Capture is armed but never selects it, and the recorder restores the last-used mode. With `camera_last_used_recorder_mode` sitting on `classic` it dies on the very first assertion. Harmless from the top of the flow, since `clearState` wipes the key — confusing standalone, which is the workflow the util's own header points at. I documented it next to your other preconditions rather than engineering around it: the wheel is a lazy `ListView`, so Capture isn't rendered from Classic and there's no one-line fix. Say the word if you'd rather the util corrected itself.

Three doc-only commits pushed, nothing behavioural. Checks green.

Issue #6903, opened 2026-08-08
fix(safety): remove blocked users from the public kind-3 follow list


NotThatKindOfDrLiz:
## Description

When a user blocks someone, Divine updates the in-app experience so the blocked account is no longer shown to that user, but it does **not** update the user's public follow list / kind-3 follow event. As a result, other people can still see the blocked account as someone the user follows.

This creates a mismatch between:

- the user's actual intended relationship to the blocked account in the app, and
- what is publicly represented in their follow graph.

The concern raised in the meeting was that this can mislead others into thinking the user still endorses, supports, or considers it safe to follow the blocked account, simply because that account remains visible in the public follow list.

## Expected behavior

- If a user blocks someone, that blocked account should no longer remain in the public follow list / emitted kind-3 follow event.
- Blocking should result in the relevant follow-state representation being updated so public-facing data no longer shows the blocked account as followed.

## Why this matters

- Prevents public follow data from contradicting the user's block action.
- Reduces confusing or socially harmful signaling around who a user appears to support or endorse.
- Brings app behavior and exported/public protocol behavior into alignment.

## Acceptance criteria

- [ ] Blocking a followed user removes that account from the user-visible follow state.
- [ ] Blocking also updates the public follow list / kind-3 event so the blocked account is no longer listed as followed.
- [ ] External viewers of the user's follow list no longer see blocked accounts represented as active follows.
- [ ] Behavior is consistent across app state and protocol-visible state after block is applied.

## Implementation note

This is not about forcing block and follow to become the same concept everywhere. The immediate fix is simpler: if someone is blocked, they should not still appear in the public follow list, regardless of the internal separation between block and follow lists.

## Starting points in the code

`ContentBlocklistRepository.blockUser` currently updates the runtime blocklist, persists it, and publishes the kind-30000 block list — but never touches the follow list:

- `mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart:654` — `blockUser`, which calls `_publishBlockListToNostr()` only
- `mobile/packages/follow_repository/lib/src/follow_repository.dart:1329` — `unfollow`, the existing kind-3 mutation path

Block call sites that would inherit the fix if it lands in the repository layer:

- `mobile/lib/blocs/other_profile/other_profile_bloc.dart:187`
- `mobile/lib/blocs/comments/comment_reactions/comment_reactions_bloc.dart:366`
[...]

Issue #5462, opened 2026-06-22
refactor(nostr): remove legacy kind 30000 block-list publishing


NotThatKindOfDrLiz:
## Context

PR #5435 keeps publishing Divine's legacy kind 30000 d=block list while adding standard NIP-51 kind 10000 mute-list publishing. That compatibility shim protects older Divine clients during the rollout, but it should not become permanent protocol debt.

## Exit criteria

- Current supported Divine clients read blocks from the standard kind 10000 mute list.
- Backend and mobile flows no longer require kind 30000 d=block for user blocks.
- Telemetry, support window, or release notes confirm it is safe to stop dual-writing.

## Work

- Remove _publishBlockListToNostr and block/unblock calls that dual-write kind 30000.
- Remove legacy kind 30000 block-list migration once all relevant users have had a safe migration window.
- Update content_blocklist_repository tests and docs to describe kind 10000 as the only authored block/mute list.

## Validation

- Run the content_blocklist_repository test suite.
- Run flutter analyze from mobile/.

Follow-up to #5435.

PR #7026, opened 2026-08-10
fix(safety): let a blocked conversation stay readable without unblocking

### LINKED ISSUE #7025: fix(safety): let a blocked conversation stay readable without unblocking
Reported by Trust & Safety, 2026-08-10.

> I asked a user for screenshots of abusive DMs that she had received. Because the user who sent them blocked her, she can no longer see the messages.

## What is actually happening

Confirmed by T&S after a two-account test and a check with the affected user: **the messages are hidden by the reporter's own block, not the sender's.** There was a mutual block in place. Blocking someone removes the conversation from Messages, and the only way back is to unblock — which re-exposes the user to the person they blocked.

So this is not a defect in the block filter. `filterBlockedConversations` correctly honours the viewer's own block. The problem is that blocking is the only control on offer and it doubles as evidence destruction.

The requirement, from T&S:

> For a safety feature, would it not be a good idea to maintain someone as being blocked and be able to see the messages? So that if they need to screenshot for evidence they don't have to risk exposing themselves to the other party again?

## Nothing is deleted

`ContentBlocklistRepository.blockUser` (`content_blocklist_repository.dart:868`) touches no DM storage, `dm_repository` has no dependency on the blocklist, and `/inbox/conversation/:id` carries no route guard — it renders the full history unfiltered. Rows stay in `direct_messages` for as long as the account exists.

**Only the entry point is missing.** Anyone currently affected can recover their messages by unblocking; the work here is making that unnecessary.

## Agreed direction

- Blocked chats stay **out of the default list** — as Alex put it, "if you block me, I think it's reasonable to expect that you won't see me in your DMs anymore".
- They are reachable behind a deliberate affordance. Interim: **a third filter chip next to All and Unread**, per Alex, until Sebastian designs the container properly — the Facebook-style overflow menu in Aleysha's screenshot, which could later also hold Message requests.
- **Every blocked account is listed**, not only those with an existing thread. Rows with no history show "No messages" and are inert.
- The thread opens **read-only**. The block stays in force, so the composer and the reaction affordance are replaced by a notice. Reading and screenshotting is the point; replying would undo the block the user deliberately set.

### Naming

"Blocked", not "Restricted". Facebook's Restricted is a distinct mechanic — the person keeps messaging you into a quiet folder and is not told — which is closer to a mute. Divine has neither restrict nor mute today, and `restricted` appears in `mobile/lib` only in unrelated contexts (`videoErrorAgeRestricted`, `videoErrorContentRestricted`, `keyManagementRestrictedTitle`). The filter is an enum so a Muted slice can be added when muting exists.

## Also in scope: the same user could not unfollow him

`canTargetUser` is `!isBlockedBy(pubkey)` (`content_policy_engine.dart:68-70`) and gates Follow **and Unfollow** together (`profile_action_buttons_widget.dart:259`, `:281`), and the profile is replaced by a dead-end screen before those buttons render at all — `UserNotAvailableScreen` on `/profile-view/:npub`, `BlockedUserScreen` on `/profile/:npub`. Neither offered Report or a way to copy the npub.

Being blocked therefore removed both the escalation path and the means to sever the tie. Unfollow and Report are disengagement and safety actions rather than interactions with the person, so both are restored on those screens via the standard profile more-sheet. `canTargetUser` itself is unchanged — hiding *Follow* from someone who rejected contact is correct.

The Interaction invariant in `docs/superpowers/specs/2026-04-23-content-policy-layer-design.md` is amended to record the exemption. The Disclosure invariant is untouched: no copy anywhere states that a block exists.

## Not reproduced, out of scope

The original filing blamed a different mechanism: `filterBlockedConversations` also consults two **third-party** buckets — `_blockedByOthers` (kind-30000 `d=block`) and `_mutualMuteBlocklist` (standard NIP-51 kind-10000) — both populated from filters with no `authors` constraint (`:1047`, `:1132`). That would let someone else's published list decide what you see of your own message store.

It did not reproduce: a two-account test left the message visible on the recipient, and T&S confirmed messages hide only on your own block. Both syncs are once-per-session (`_blockListSyncStarted` `:1106`, `_mutualMuteSyncStarted` `:1021`) and neither bucket is persisted, so a block published mid-session may never land. **Left unchanged.** The retest that would settle it: block from account A, fully restart the app on account B, check the inbox before anything else.

## Related

- #6903 — blocking leaves the account in the public kind-3 follow list; same user hit this
- #6435 / #6434 — the inverse direction, a lifted block never reaching the blockee
- #6601, #6436 — block confirmation and publish-acknowledgement gaps

realmeylisdev:
Closes #7025.

## Why

Trust & Safety asked a user for screenshots of abusive DMs and she couldn't produce them: blocking removes the conversation from Messages, and the only way back is to unblock — which re-exposes you to the person you blocked.

T&S confirmed the cause after a two-account test: **the messages are hidden by the viewer's own block**, not the sender's. So this is not a defect in the filter; it is that blocking was the only control on offer and it doubled as evidence destruction.

> For a safety feature, would it not be a good idea to maintain someone as being blocked and be able to see the messages? So that if they need to screenshot for evidence they don't have to risk exposing themselves to the other party again?

Nothing was ever deleted — `blockUser` touches no DM storage and the conversation route renders unfiltered. Only the entry point was missing.

## What changed

**A Blocked filter chip**, third in the existing All / Unread row. `ConversationListBloc` now partitions blocked conversations out instead of discarding them. The default inbox is unchanged — a blocked thread still leaves the list, which is the block working as intended.

The slice is seeded from `runtimeBlockedUsers`, not from the conversation set, so an account you blocked but never messaged appears too; those rows carry no thread, show "No messages" and are inert. The chip is hidden entirely until you have blocked someone, and the filter falls back to All when the last block is lifted.

**The thread opens read-only.** The conversation route carries no blocklist guard, so without this a blocked thread would have rendered a fully live composer — handing users the exact re-exposure the feature exists to remove. The composer and the reaction affordance are replaced by a notice; the history itself is untouched, because reading and screenshotting is the point. Unblocking from the app bar's more-sheet restores the composer.

**Unfollow and Report are reachable again when someone blocks you.** `canTargetUser` is `!isBlockedBy(pubkey)` and gated Unfollow alongside Follow, and the profile was replaced by a dead-end screen before either rendered — so the same user could not stop following the account that blocked her, or report it. Both stand-in screens now carry the standard profile more-sheet. `canTargetUser` itself is untouched: hiding *Follow* from someone who rejected contact is correct, and it also gates Message, the notify bell and the share/tag pickers.

Unfollow calls `FollowRepository.unfollow` directly rather than `toggleFollow`, which inverts to a *follow* on a cold follow list (#6903).

## Naming

"Blocked", not "Restricted". Facebook's Restricted is a different mechanic — the person keeps messaging you into a quiet folder and is not told — which is closer to a mute. Divine has neither restrict nor mute today. The filter is an `enum` so a Muted slice drops in as one case when muting exists.

## Spec

`docs/superpowers/specs/2026-04-23-content-policy-layer-design.md` is amended in two places: the Interaction invariant now exempts disengagement and safety actions (Unfollow, Report), and the Ingress invariant records what it never covered — already-delivered DM history, where the viewer's copy is the only readable copy in existence.

The **Disclosure invariant is untouched**: no copy anywhere states that a block exists.

## Deliberately not changed

`filterBlockedConversations` also consults two third-party buckets — `_blockedByOthers` and `_mutualMuteBlocklist` — which would let someone else's published list decide what you see of your own message store. That path **did not reproduce**: a two-account test left the message visible on the recipient, and T&S confirmed messages hide only on your own block. Left alone rather than shipping an unverified semantics change alongside the real fix; the retest is recorded on #7025.

## Interim by agreement

The chip is explicitly a placeholder for the overflow menu Sebastian will design — the enum and the partitioned state are the durable part, so lifting it into a menu later is a UI-only change. Message requests are not moved into it here.

## Testing

- New bloc coverage: blocked conversations are partitioned rather than dropped; the Blocked filter surfaces them; an account with no thread still yields a row; the filter falls back to All when the last block is lifted.
- New widget coverage: chip gating and reporting; the blocked thread renders the notice and no `MessageInputBar` while keeping its history; a message-less conversation stays tappable under All.
- `UserNotAvailableScreen` had no test at all; it has one now, alongside an updated `BlockedUserScreen` test (which also stopped asserting hardcoded English).
- Six new ARB keys translated into all 21 non-English locales.
- `flutter analyze lib test integration_test` clean, `golden.sh verify` clean, affected suites green (1136 tests).

## Manual test plan

1. Exchange DMs between two accounts; block the sender.
2. Thread leaves the default list; the **Blocked** chip appears.
3. Tap Blocked → thread listed → open it → full history, no composer, no reactions, block still active.
4. Screenshot the messages.
5. Unblock from the kebab → thread returns to All, composer returns, chip disappears once nothing is blocked.
6. A blocked account you never messaged is listed and inert.


realmeylisdev:
## Device test report

**Build under test:** `a92024954` (this branch, before the follow-up fixes in `626db6e35`).
**Target:** iPhone 17 Pro Max simulator, iOS 26.5, debug build, **production backend**, real signed-in account `9be8bd90d818407bcf574d11b1c57f104fd53f40fa767abc4a631ee2694b43a3`.
**Counterparty used:** an existing DM thread with `b2380758159d89727a81e9b76b2e2a51216d341b4deddf0540f27cf7bb121738` (renders as "Deleted account"), chosen as the lowest-blast-radius target with real history.
**Method:** `idb` taps at logical points, every step verified against a screenshot; `flutter run` console captured for the whole session.

### Executed

| # | Step | Expected | Result | Evidence |
|---|---|---|---|---|
| 1 | Messages tab, no blocks held | Chip row shows All / Unread only | **Pass** | Screenshot 18:41 — two chips, no Blocked |
| 2 | Long-press thread → Block user | Snackbar; thread leaves All; Blocked chip appears | **Pass** | Screenshot 18:42 — "Blocked user"; row gone from All; third chip present |
| 3 | What the block publishes | kind 10000 + kind 30000, signed | **Pass** | Log `18:42:30.926` signing kind 10000 → `18:42:31.400` "Published mute list… 1 pubkey entries"; `18:42:31.400` signing kind 30000 → `18:42:31.630` "Published block list… 1 entries"; `18:42:31.630` "Added user to blocklist: b2380758…" |
| 4 | Select Blocked chip | Blocked thread listed with its history; Message-requests banner suppressed | **Pass** | Screenshot 18:43 — chip active, one row "relay admission check 6585 · Thursday", no banner |
| 5 | Open the blocked thread | Full history readable; **no composer**; notice in its place | **Pass** | Screenshot 18:43 — three bubbles rendered; `MessageInputBar` absent; "You blocked this account / Messages stay here so you can read or screenshot them. Unblock to reply." |
| 6 | Kebab inside the blocked thread | Report and Copy npub remain reachable | **Pass** | Screenshot 18:45 — "Copy public key (npub)", "Report Deleted account", "Unblock Deleted account" |
| 7 | Whole flow | No exceptions, overflow or render errors | **Pass** | `grep -E "EXCEPTION\|RenderFlex\|overflow\|Unhandled"` over the full session log — zero hits |
| 8 | Unblock from the thread kebab | Composer returns; thread returns to All; chip disappears | **Blocked by #7027** | Screenshots 18:46, 18:48, 18:50, 18:52 — notice persists ~6 min after `18:46:44.691` "Removed user from blocklist" |
| 9 | Unblock from Settings → Content & Safety | Row leaves BLOCKED USERS | **Blocked by #7027** | Accessibility dump after tap: snackbar `'User unblocked'` **and** row `'npub1kguqw…5duwv3 … Unblock'` both present |
| 10 | Does an unblock persist? | Pubkey absent from prefs after app terminate | **Blocked by #7027** | `xcrun simctl terminate` then `plutil -p`: `flutter.blocked_users_list.9be8bd90… => ["b2380758…"]` |

Rows 8–10 are **not defects of this PR**. `git diff origin/main...HEAD -- mobile/packages/content_blocklist_repository/` is empty — this branch does not touch that package. Root cause and repro are in #7027. The consequence for review is that **"the composer returns on unblock" could not be exercised on device**, because the unblock never takes effect; it is covered by widget test only (`conversation_view_test.dart`, "an unblocked counterparty keeps its composer", plus the `blocklistVersionProvider` rebuild path).

### Not exercised on device — test-covered only

Stating these explicitly so the coverage claim isn't read wider than it is.

| Behaviour | Why not | Covered by |
|---|---|---|
| Inert row for a blocked account with **no** thread | The account used had history; no zero-thread block was created | `conversation_list_bloc_test.dart` "lists a blocked account the viewer never messaged"; `inbox_view_test.dart` "a message-less conversation stays tappable under All" |
| Blocked chip disappears when the last block is lifted | Requires a working unblock (#7027) | `conversation_list_bloc_test.dart` "falls back to All when the last block is lifted" |
| Search within the Blocked slice | Not driven | `conversation_list_bloc_test.dart` "resolves names for blocked counterparties so search matches" |
| Reaction affordance disabled in a blocked thread | Not driven; existing reaction pills correctly still render (they are evidence too) | `conversation_view.dart:267` gates `showPicker` / `onDoubleTap` |
| Chip row overflow in long-label locales / large text scale | Single locale, default text scale | Row now scrolls horizontally (`inbox_filter_chips.dart`) |
| The four fixes in `626db6e35` | They landed **after** this device run | Unit + widget tests added alongside each |

### Incidental observation

The blocked row in Content & Safety renders the truncated npub as both title and subtitle when the counterparty's kind-0 is unavailable — accessibility label `'npub1kguqw…5duwv3\nnpub1kguqw…5duwv3\nUnblock'`. That is the expected fallback (`safety_settings_screen.dart:466-477`), and it answers the question raised in the originating thread about whether suspending an account removes it from the blocked list: it does not — only the displayed name degrades.

realmeylisdev:
Short version of everything around that test run.

**CI** — green, all 20 checks (four test shards, Analyze, Generated Files, Format, aggregate Mobile CI).

**Fixed after the device pass, in \`626db6e35\`** — four defects, all introduced by this PR, all caught before merge:

| Defect | Why it mattered |
|---|---|
| A group thread could displace the 1:1 it shares a blocked account with | The slice allocated one row per blocked *account* keyed on the first non-self participant, and participants are stored sorted — so a group whose smallest member was blocked took the slot. `filterBlockedConversations` had already removed that 1:1 from the inbox, so it appeared in **no list at all**: the exact evidence loss this PR exists to prevent |
| An all-blocked inbox said "You're all caught up" | Holding a block now skips the empty-inbox early return, so the unfiltered empty case claimed a filter had emptied it |
| Search under Blocked could never match a blocked counterparty | `profileNames` resolves from `conversations`, which the blocklist filter has already emptied of exactly those pubkeys — matcher saw the generated fallback, row rendered the real kind-0 name |
| Three-chip row could overflow narrow viewports | `DivineButton`'s internal `Flexible` sits in its own `MainAxisSize.min` Row and cannot yield width to a sibling |

Also corrected a comment that claimed the slice recovers everything the filter removed — it recovers the viewer's own runtime blocks only; the other four hide-buckets stay hidden by design.

**Pre-existing, filed as #7027** — unblocking does not stick, reproduced from two independent surfaces. Not caused here and not in code this branch touches, but it interacts: the notice says "Unblock to reply", and unblocking is currently unreliable. Worth landing close behind.

**Protocol / backend, checked against the NIPs and the sibling repos:**

- **NIP-51 kind 10000** is first-person and feed-scoped — "things the user doesn't want to see in their feeds". Nothing in it grants a list author authority over another reader's view, which supports scoping the DM slice to the viewer's own blocks.
- **kind 30000 `d=block`** is a Divine invention on the *Follow sets* kind — a conformant reader sees a follow set titled "Block List". Already conceded as legacy in-code under #5462.
- **Nothing server-side hides or deletes a DM on a block, anywhere.** funnelcake's only kind-1059 delete path is the NIP-62 self-vanish; the REQ visibility predicates are all global platform-moderation sets with no viewer bind; `mute_list_pubkeys` is never referenced under `crates/relay/`. keycast stores no gift wraps. divine-web's DM code imports no blocklist. This is structural — NIP-59 makes the wrap author a random one-time key, so a relay cannot attribute a DM to a sender at all. Any future "filter DMs server-side on block" proposal should be refused on protocol grounds.

**Weakest open claim:** the "third-party path did not reproduce" note on #7025. Both checks so far used a Divine-to-Divine block; a plain NIP-51 kind-10000 mute from a non-Divine client hits an unconditional branch at `content_blocklist_repository.dart:1322-1334` that nobody has exercised.

realmeylisdev:
Simulator Screenshot - iPhone 17 Pro Max - 2026-08-10 at 18 46 41
Simulator Screenshot - iPhone 17 Pro Max - 2026-08-10 at 18 43 46
Simulator Screenshot - iPhone 17 Pro Max - 2026-08-10 at 18 43 15

hm21:
Pushed two commits.

**The blocked thread wasn't read-only.** A send that hard-failed before the block is still on screen as a red bubble, and its tap opens the recovery sheet — `ConversationFullSendRecoveryRequested` → `recoverFullSend`, and neither the bloc handler nor the repository consults the blocklist. So "Retry send" delivers a DM to the account the viewer blocked, from the one screen built to make that impossible. `isBlockedByUs` was wired only to `reactionsEnabled` and the notice; the bubble's `onTap` was untouched.

Repro: send a message that hard-fails → block the counterparty → open the thread from the Blocked chip → tap the red bubble → Retry send.

`65dc4b1` gates that tap on a new `sendRecoveryEnabled`, threaded alongside `reactionsEnabled`. The bubble stays rendered — it is part of the record the viewer came to read — so only the affordance goes and nothing is force-deleted. Regression test added to the `blocked thread` group; confirmed red without the fix.

`3f0b2bc` corrects a claim I got wrong in `65dc4b1`'s comment. I wrote that a row has exhausted the retry sweep's budget by the time it reads `failed`. It has not — `OutgoingDmsDao.getRetryableForOwner` selects failed rows still under `maxRetries`.

Which leaves one thing I deliberately did not touch: the sweep can still deliver such a row to a blocked account with no UI involved, because blocking has never cancelled the viewer's own queued sends. That is repository-layer and a real product call ("does blocking cancel your outbox?"), not something to settle inside a review.

One smaller note, no action implied: a blocked account you never messaged renders inert — both `onTap` and `onLongPress` are null — so there is no unblock path from that row; Settings → Content & Safety is the only way. Defensible given the sheet's actions need a real DB row.

The rest reads right to me: the partitioning off `runtimeBlockedUsers`, the #176 gate applied to the blocked slice, the fallback to All, and translations in all 21 non-English locales. All checks green on `3f0b2bc`.

hm21 (approved):
Approving the state after my two pushed commits. The resend hole is closed and covered; the background-sweep gap is pre-existing and out of scope here. Have a look at 65dc4b1 / 3f0b2bc before you merge — the merge call is yours.

NotThatKindOfDrLiz:
@realmeylisdev takeover changes are pushed on `bee651414` and CI is green.

What changed:
- Dropped the two merge commits by rebuilding the branch linearly on current `origin/main`.
- Added receipts and error handling for unavailable-profile Unfollow / Block / Unblock actions.
- Added focused widget coverage proving Unfollow calls `FollowRepository.unfollow`, never `toggleFollow`, plus block/unblock success receipts and failed-unfollow error handling.
- Filed #7047 for the remaining background queued-send policy decision and linked it from the reconnect-sweep comment.

Verification:
- `flutter test test/widgets/profile/unavailable_profile_actions_test.dart test/screens/user_not_available_screen_test.dart test/widgets/profile/blocked_user_screen_test.dart test/screens/inbox/conversation/conversation_view_test.dart test/screens/inbox/inbox_view_test.dart test/blocs/dm/conversation_list/conversation_list_bloc_test.dart`
- Local CI shard 2 reproduction in a disposable worktree: `very_good test --optimization --concurrency=4 --exclude-tags integration --test-randomize-ordering-seed random` after `select_test_shard --total 4 --index 2`, passed 4042 tests / 57 skipped.
- GitHub Mobile CI is green, including the formerly red shard 2.

Please review the takeover commit and merge when you are comfortable with the final state.

NotThatKindOfDrLiz (approved):
Approved after takeover.

I pushed `bee651414` to address the remaining blockers: unavailable-profile action receipts/error handling, direct tests for the action switch, #7047 tracking for the queued-send policy gap, and a clean linear branch history on `origin/main`.

All required GitHub checks are green on the current head, including `Tests (shard 2/4)`.

PR #7028, opened 2026-08-10
feat(verify): link and unlink accounts without leaving the app

### LINKED ISSUE #7030: feat(verify): link and unlink accounts without leaving the app
Auto-created from PR #7028.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7028
**Branch:** `feat/in-app-verify`

hm21:
Closes #7030

## Description

Verifying an account meant leaving the app. "Get verified" opened the system browser at `verifier.divine.video`, and that was the end of it — the flow leaves that host for `login.divine.video` and the OAuth providers, so an in-app WebView cannot complete it. Signing in again on a phone, to publish an event this app could already sign, is where people gave up.

This brings it in, the same way #7014 did for badges. The `verifier.divine.video` Worker keeps everything that cannot move — the OAuth handshakes, the platform verifiers, the KV cache, the public lookup page — and the app stops being a link to it.

**The flow.** A dashboard lists what is linked with the verifier's verdict on each, offers the platforms that are not, and unlinks. A connect screen per platform: one-tap sign-in where the verifier offers it (Twitter/X, Bluesky, TikTok, YouTube), a proof post otherwise. Reachable from Edit Profile as before, and now also from Settings → Identity.

**A claim is published only after the verifier confirms it.** The web flow leaves publishing as a second button, which is how someone ends up verified with no checkmark — they never pressed it. Here there is nothing to press. That is the structural half of #6154.

**OAuth rides the App Link crossposting already uses** (`divine.video/app/callback`), which the verifier already allowlists, so no service change was needed. The callback constants, the strict callback parser and the iOS 17.4 floor now live in `lib/features/oauth/`, and crossposting's own launcher is gone — its screen calls the shared one directly. Two copies of a URL that has to match `AndroidManifest.xml` and the AASA is two things to keep in step.

### Why some things work the way they do

**Every platform gets its own proof instructions.** The verifiers read different surfaces — a gist's *first* file, a tweet's text, a TikTok caption, a YouTube description, a Discord message the bot can see — and none of it is guessable. A test walks the verifier's whole platform list and fails if one falls through to generic copy or shares another's.

**Pasted post links are taken apart before they are sent.** Most verifiers rebuild the lookup URL from a handle and a bare id, so a full link in the proof field looks up a URL that cannot exist. The verifier's own web form does this client-side and the service does not repeat it, so a client that skips it is the one that breaks.

**The account field only appears when the link cannot carry it.** Six of the eight platforms name the account in the link. Asking twice invited the mismatch a channel link makes easy: a channel post's author is the channel, not the person pasting. Telegram and YouTube, where the field stays, ask for a *channel* name — and Telegram's instructions say so, plus that a freshly created channel is private, because both are wrong by default.

**Both form fields carry a key, and that is load-bearing.** The account field is inserted and removed as the typed proof starts and stops parsing, and for Bluesky it also moves between the OAuth block and the proof block once the iOS support probe resolves. `ListView` matches children by position, so each of those shifts handed a slot to a different widget and rebuilt the field from scratch — typing anything that did not parse as a post link wiped what had just been typed, and closed the keyboard with it. The account field's key sits on the list's own child rather than on the text field inside it: keying the inner field still leaves the element to be rebuilt by position, and loses the text anyway.

**Unlinking asks first, and also revokes the verifier's cached login,** so re-linking asks the provider again instead of passing on a 24-hour-old one. The confirmation is there because the revoke is what makes a mis-tap expensive — the tag is cheap to re-add, a fresh sign-in is not. The revoke itself is best-effort: the tag is already gone, which is what the user asked for.

### Writing the identity event safely

Every write replaces the whole kind-10011 event, so a bad read loses data rather than showing stale data. Four guards:

- **With no relay connected, an empty query result carries no information.** Refusing on that fact alone is the only check that holds on a device that has never completed a read — which is exactly the state after the OAuth browser round trip drops the relay pool.
- **A local snapshot or a verdict snapshot that disagrees with an empty read** is treated as evidence the read was incomplete.
- **A relay confirms a write before it necessarily serves it back.** Two links added in quick succession each read a base predating the other and published it away, so each new account silently deleted the previous one. What this device published is merged over a lagging read, with removals tracked alongside so the same merge cannot resurrect them.
- **Reaching the kind-0 fallback is not evidence that the kind-10011 read worked.** Every profile has a kind-0 metadata event, so that query answers even when the identity event does not come back. Publishing on it replaced the event with a set predating the claims this device had already seen — the same silent unlink as above, reached through the one branch that skipped both the merge and the snapshot guards. It now takes the same merge, and refuses when the local snapshot knows claims the fallback does not carry.

The first three were found on the device; the fourth was found in review, on the branch the other three did not cover.

Reading takes the opposite branch and falls back to the last known set: a list that empties itself on a relay hiccup reads as "my links are gone".

### Not requested by this feature

- **`fix(profile): scroll the verified-account chips instead of wrapping`** — a profile with several links grew the header by a row of chips each time. **Visible change** on every profile header, not just your own. The block keeps its position and its 16px inset; only `Wrap` becomes one horizontally scrolling row.
- The shared-OAuth extraction above touches crossposting: its launcher and callback constants are gone, and the settings screen calls the shared ones instead.

## Out of Scope

- **Twitter/X and TikTok one-tap are broken on the verifier side and are not fixed here.** Both have been reported in Slack for a while and are unchanged by this PR. Twitter/X: `TWITTER_CLIENT_ID` is not set on the Worker, so `/auth/twitter/start` answers `503 {"error":"Twitter OAuth not configured"}`. TikTok: the flow starts (302) but does not come back with a usable confirmation. The app resolves the verifier's start redirect itself and opens the browser on the provider URL it names, so a 503 falls back to the proof post with a clear message instead of stranding the user on a JSON error page — and one tap still costs the verifier exactly one start request. The verifier's own refusal reason is now logged so the next report carries it. Both need a verifier-repo fix.
- **YouTube one-tap** is unconfigured for the same reason (`GOOGLE_CLIENT_ID`); YouTube is also absent from `GET /platforms` without `YOUTUBE_API_KEY`, so the app does not offer it at all today.
- **OAuth on iOS < 17.4** stays hidden, the same runtime gate crossposting uses: `ASWebAuthenticationSession.Callback.https` only exists from 17.4, and below it a completed session can report as a cancel. The proof post covers those devices.
- **An OAuth verdict lives 24h in the verifier's KV.** After that the claim falls through to the platform proof check, which cannot resolve the `oauth` sentinel, so the link has to be renewed to stay verified. Identical on web; a fix belongs server-side.
- The Nostr-apps directory still lists the verifier and still opens it in the system browser — someone may want the web tool, and its slug stays in `kSystemBrowserAppSlugs`.
- The profile badge sheet's "Make your own at badges.divine.video" is untouched; that is #7014's territory.

## Verification

- `flutter analyze lib test integration_test` — clean; `dart format` — clean
- `verifier_client`: targeted client/platform tests
- `profile_repository`: targeted repository tests
- App: 917 tests across the verify blocs and screens, the shared OAuth callback, profile-setup, profile-editor, profile widgets, settings and l10n
- `flutter test test/l10n/arb_consistency_test.dart` — 53 new keys translated into all 21 non-English locales, no English fallbacks
- Every ratchet in the `Generated Files` job passes, including the raw-`Icons.*`, raw-`TextStyle`, raw-colour, Material-button, raw-dialog, untested-services and test-structure guards
- Generated files regenerated and committed (`auth_providers.g.dart`, `lib/l10n/generated/`)

### Takeover verification (`ead57dce`)

- `flutter test test/providers/crossposting_providers_test.dart test/widgets/profile/verified_accounts_row_test.dart test/screens/profile_setup/verified_accounts_section_test.dart test/screens/verify/verify_screen_test.dart test/screens/verify/verify_connect_screen_test.dart test/blocs/verify/verify_connect_cubit_test.dart`
- `flutter test test/src/verifier_client_test.dart` from `mobile/packages/verifier_client`
- `flutter analyze lib test`
- `flutter analyze` from `mobile/packages/verifier_client`
- `flutter analyze` from `mobile/packages/profile_repository`
- Pre-push hook on the rebased branch: merge-conflict check, analyzer, generated-file verification and changed-file tests all passed.

**Device-tested on a real Samsung Galaxy**, against production relays and the live verifier. Every method was exercised:

| Platform | Route | Result |
|---|---|---|
| GitHub | proof post (gist) | works |
| Telegram | proof post | works |
| Discord | proof post | works |
| Mastodon | proof post | works |
| Bluesky | one-tap | works |
| Twitter/X | one-tap | **fails — verifier side, reported in Slack** |
| TikTok | one-tap | **fails — verifier side, reported in Slack** |

Unlinking was exercised too, including the confirmation sheet, and the first three data-loss cases above were found and fixed on the device rather than in tests. The kind-0 guard and the keyed form fields came out of review afterwards and are covered by tests rather than by a second device pass — each of those tests fails without its fix.

One note for whoever tests this on a local build: `divine.video` App Links only verify against the Play signing certs, so a debug build falls back to the browser and shows the store page instead of returning to the app. `adb shell pm set-app-links-user-selection --user 0 --package co.openvine.app true divine.video` enables it for testing. This affects crossposting identically and is not new here.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mobile/packages/verifier_client/lib/src/verifier_client.dart
```diff
@@ -80,16 +82,178 @@ class VerifierClient {
...
+      'GET',
+      oauthStartUri(
+        platform: platform,
+        pubkey: pubkey,
+        returnUrl: returnUrl,
+        handle: handle,
+      ),
+    )..followRedirects = false;
+    try {
+      final response = await _httpClient.send(request).timeout(_timeout);
```
NotThatKindOfDrLiz: This preflight is still a real `/auth/<platform>/start` request, not a capability endpoint. That means an OAuth tap starts the verifier flow twice; for providers like Bluesky this can mint server-side OAuth state, resolve the PDS, and spend PAR/rate-limit budget before the user actually launches OAuth. Please confirm with the verifier owners that discarded `/start` requests leave no problematic state or rate-limit impact, or switch this to a true capability signal from `/platforms` before merge.

mobile/lib/widgets/profile/profile_header_identity.dart
```diff
@@ -129,10 +129,10 @@ class _ProfileNameAndBio extends StatelessWidget {
                       )
                     : null,
               ),
-              _VerifiedAccountsBlock(isOwnProfile: isOwnProfile),
             ],
           ),
         ),
+        _VerifiedAccountsBlock(isOwnProfile: isOwnProfile),
```
NotThatKindOfDrLiz, line 98: This moves the verified-account chips out of the name/bio column and below the bio, support button, and website row. That is a visible reorder, not only the documented Wrap-to-horizontal-scroll behavior. Please add a screenshot and get design sign-off for the new order, or keep the previous visual order while making the chips horizontally scrollable.

NotThatKindOfDrLiz (changes requested):
I pushed 55e41e1 with fixes for the local code blockers. I am leaving this as Request changes for the two remaining confirmation items: verifier-owner confirmation that the OAuth capability probe is harmless, and design sign-off or adjustment for the profile-header verified-account reorder. Once those are answered or adjusted, this should be ready for another pass.

mobile/packages/verifier_client/lib/src/verifier_client.dart
hm21: Fixed in 5844495 rather than confirmed — you were right, and a discarded `/start` costs more than a wasted call: every configured start writes OAuth state to KV and spends a per-IP rate-limit token, and Bluesky's also resolves the PDS and pushes a PAR.

`GET /platforms` carries no OAuth-capability field today (`{label, supported}` only), so moving the check there needs a verifier-side change. Instead the pre-flight *is* the start now: the app resolves the 302 and opens the session on the provider URL it names. One tap = one start request — fewer than before, same as no pre-flight at all, and the 503 is still caught.

mobile/lib/widgets/profile/profile_header_identity.dart
hm21: There is no reorder. `_VerifiedAccountsBlock` is already the last child on `main`, below the bio, support button and website row — the whole diff to this file is those 2 lines, moving it out of the shared 16px `Padding`. `VerifiedAccountsRow` re-applies `horizontal: 16` inside its own `SingleChildScrollView`, so at rest the chips sit at the same y and the same x.

The only visual change is the documented `Wrap` → one horizontally scrolling row, which is your second option.

hm21:
@NotThatKindOfDrLiz both items answered, checks green.

**OAuth capability probe** — fixed in 5844495 rather than confirmed; details in the thread. One tap now costs the verifier exactly one start request.

**Profile-header reorder** — no reorder to sign off on; the block was already last on `main`. Details in the thread.

**Also**: 55e41e1 dropped profile_repository to 99.91% and the package's 100% gate went red — the created_at/id tie-break's id branch was untested. Covered in 5649053.

mobile/lib/screens/verify/verify_connect_screen.dart
```diff
@@ -0,0 +1,344 @@
...
+        const SizedBox(height: 16),
+        // Skipped when the pasted link already carries the account, and when
+        // the OAuth block above is already asking for it.
+        if (state.needsIdentityInput &&
+            !(showOAuth && state.platform.key == 'bluesky')) ...[
+          _IdentityField(state: state),
+          const SizedBox(height: 12),
+        ],
+        DivineTextField(
+          labelText: l10n.verifyProofLabel,
```
realmeylisdev, line 240: **Blocking.** Typing or pasting anything here that isn't a recognisable post link makes this field go blank and closes the keyboard.

`needsIdentityInput` flips to true on the first keystroke, which inserts `_IdentityField` above this widget in the same `ListView`. Both fields are unkeyed, so the list matches children by position: this `DivineTextField` lands in a slot that held something else, gets rebuilt from scratch, and the `TextField`'s own controller and focus node go with it. The reverse flip does the same thing when a pasted link makes the account field disappear again.

Reproduced on an iPhone 17 Pro simulator against the live verifier: open GitHub, type `abc123` into "Link to your post", and the account field appears while the proof field is back to its placeholder. In a widget test the cubit still holds `abc123` while `find.text('abc123')` matches nothing and both fields report `hasFocus == false`.

Discord, YouTube and Bluesky escape it because they always render both fields. GitHub, Twitter/X, Telegram, Mastodon and TikTok all hit it unless the very first thing pasted already parses.

A key is enough — the list then matches the field across the shift and both the text and the focus survive:

```suggestion
        DivineTextField(
          key: const ValueKey('verify-proof-field'),
          labelText: l10n.verifyProofLabel,
```

Regression tests, both red before the keys and green after:

```dart
testWidgets('keeps the typed proof when the account field appears', (
  tester,
) async {
  await pump(tester);

  await tester.enterText(find.byType(TextField), 'some nonsense');
  await tester.pump();

  expect(find.text('some nonsense'), findsOneWidget);
  expect(tester.testTextInput.isVisible, isTrue);
});

testWidgets('keeps a pasted link when the account field goes away', (
  tester,
) async {
  await pump(tester);
  await tester.enterText(find.byType(TextField), 'nonsense');
  await tester.pump();
  expect(find.byType(TextField), findsNWidgets(2));

  await tester.enterText(
    find.byType(TextField).last,
    'https://t.me/divinechannel/456',
  );
  await tester.pump();

  expect(find.byType(TextField), findsOneWidget);
  expect(find.text('https://t.me/divinechannel/456'), findsOneWidget);
});
```

Worth not reusing `https://t.me/mychannel/123` as the fixture, by the way — it's `verifyProofHint('telegram')` verbatim, so `find.text` matches the hint and the assertion passes even when the field is empty.

mobile/lib/screens/verify/verify_connect_screen.dart
```diff
@@ -0,0 +1,344 @@
...
+
+class _IdentityField extends StatelessWidget {
+  const _IdentityField({required this.state});
+
+  final VerifyConnectState state;
+
+  @override
+  Widget build(BuildContext context) {
+    return DivineTextField(
+      labelText: verifyIdentityFieldLabel(context.l10n, state.platform.key),
```
realmeylisdev, line 276: Same fix needed here, otherwise the account field loses whatever was typed into it when it moves between the OAuth block and the proof block (Bluesky, once `appOAuthSupportProvider` resolves).

```suggestion
    return DivineTextField(
      key: const ValueKey('verify-identity-field'),
      labelText: verifyIdentityFieldLabel(context.l10n, state.platform.key),
```

mobile/test/features/oauth/app_oauth_callback_test.dart
```diff
@@ -0,0 +1,126 @@
...
+        throwsFormatException,
+      );
+    });
+
+    test('rejects a malformed percent escape', () {
+      expect(
+        () => parseAppOAuthCallback('https://divine.video/app/callback?a=%zz'),
+        throwsFormatException,
+      );
+    });
```
realmeylisdev, line 65: The crossposting file this replaces also had `does not expose OAuth code or state in callback errors`, and it didn't come across. The property still holds — the message is a `const` with no interpolation and no `source` attached — but nothing pins it any more, and this exception gets logged with the whole query string in scope.

Mutation-checked: swapping the throw above for `FormatException(_invalidCallbackMessage, callback)` turns this red.

```dart
test('keeps the OAuth code and state out of the rejection', () {
  const secretCode = 'oauth-code-super-secret';
  const secretState = 'oauth-state-super-secret';
  FormatException? caught;

  try {
    parseAppOAuthCallback(
      'https://divine.video:443/app/callback'
      '?code=$secretCode&state=$secretState',
    );
  } on FormatException catch (error) {
    caught = error;
  }

  expect(caught, isNotNull);
  expect(caught?.source, isNull);
  expect(caught.toString(), isNot(contains(secretCode)));
  expect(caught.toString(), isNot(contains(secretState)));
});
```

realmeylisdev (dismissed):
the proof field wipes itself. the moment what you type stops looking like a post link, the account field pops in above it and the proof field loses its text and the keyboard with it. reproduced on a 17 Pro sim against the live verifier, and again in a widget test. two ValueKeys fix it, suggestions inline.

liz's two open items both look answered to me: i traced the start endpoint and one tap really is one `/start` now, and the chips block was already the last child on main so there's no reorder. details in the threads.

heads up, the branch now conflicts with main on the 21 non-english arb files (that's also why i couldn't just push these fixes).

mobile/lib/screens/verify/verify_connect_screen.dart
hm21: Confirmed and fixed in e5246eba67. Reproduced it in a widget test before touching anything: the cubit held the typed text while `find.text` matched nothing, and `testTextInput.isVisible` was false. Good catch — this one bit every platform that is not discord/youtube/bluesky, exactly as you listed.

mobile/lib/screens/verify/verify_connect_screen.dart
hm21: Real, but the suggested key does not fix it. Keyed on the `DivineTextField` inside `_IdentityField`, my regression test stays red: the `ListView` matches `_IdentityField`, so that element is still rebuilt by position and the inner key never gets a chance. The key had to go on `_IdentityField` at both call sites, plus `super.key` on its constructor. Green after that — e5246eba67.

Also worth knowing: this path is much narrower than the proof field. `appOAuthSupportProvider` resolves a frame or two after the screen opens, so you have to be typing into the handle field in that window. Fixed anyway, the key is free.

mobile/test/features/oauth/app_oauth_callback_test.dart
hm21: Restored verbatim in e20e9c60f9. Verified the property still holds today, so this is a pin rather than a fix.

hm21:
All three landed, plus the rebase. @realmeylisdev

- Proof and account fields keyed, three regression tests, each red without its fix — e5246eba67
- Grant-leak pin back — e20e9c60f9
- Rebased onto `main`; the 21 ARB conflicts were all the same shape (both sides appending keys to the same object), resolved as a union with every file re-validated as JSON and every locale re-checked for the full key set

Two corrections to the review. The account-field key does not work where you put it — details in that thread. And your fixture warning applies one line further than you spotted: `verifyIdentityHint('bluesky')` is `alice.bsky.social` verbatim, so my first version of that test passed against the hint of an empty field. Caught it because you had flagged the same trap on the proof hint, so thanks for that.

I have left your three threads unresolved so you can close them as you verify.

On process: the two keys and the restored test were mechanical, no product or architecture call in them. Under `PR_REVIEW.md` that is takeover territory — `divine-mobile` is not in the platform mapping, so any org member with write access can push clear fixes, and the PR had been non-draft for 16 hours when you reviewed. A branch that conflicts with `main` does not block a push either; it blocks a *merge*. You could have pushed a commit onto the branch without touching a single ARB file. @NotThatKindOfDrLiz did exactly that on this PR earlier — pushed the code blockers, left CHANGES_REQUESTED only for the two items that genuinely needed someone else's judgment. That is the split worth aiming for.

@NotThatKindOfDrLiz your two items are both answered and the threads are resolved: the OAuth pre-flight is gone (the app resolves the start redirect itself, so one tap costs exactly one `/start`), and the chips block keeps its position and its 16px inset — only `Wrap` becomes a horizontally scrolling row. Ready for another pass whenever you have a moment.

CI is green and the branch is mergeable again. Description updated to match.

mobile/lib/screens/verify/verify_connect_screen.dart
realmeylisdev: Fixed in e5246eb. Re-ran it here: all three of your new tests go red with the keys stripped, so they're real guards, and the whole verify and oauth set is green.

mobile/lib/screens/verify/verify_connect_screen.dart
realmeylisdev: You're right and my suggestion was wrong — the key belongs on `_IdentityField`, not on the `DivineTextField` inside it. I put the key where I suggested and ran your suite: the Bluesky OAuth-block-move test fails, `divine.bsky.social` is gone. Your placement passes. Good catch on the `alice.bsky.social` fixture too, same trap as the Telegram hint.

mobile/test/features/oauth/app_oauth_callback_test.dart
realmeylisdev: Restored in e20e9c60. Mutation-checked again on this head: attaching the callback as the exception source turns it red.

realmeylisdev (approved):
all three fixed, and the identity key belongs on `_IdentityField` like you put it — i checked, my suggested spot leaves the bluesky move broken. stripped the keys and your three tests go red, so they hold.

also went through the rebase: no arb key from main is missing anywhere, and the 27-key locale gap is byte-identical to main's existing debt. verifier_client 47, profile_repository 408 at 100%, 781 app tests, analyze clean, all six ratchets green now. lgtm 👍

realmeylisdev:
You're right, and my reason was wrong. The conflict blocks a merge, not a push — my two commits touched `verify_connect_screen.dart` and its test, no ARB file, and the branch is unprotected, so a fast-forward would have gone straight in.

What actually stopped me was local, and I described it badly. The repo's pre-push hook runs `merge-tree` against `origin/main` unconditionally and exits 1 on any conflict, whoever is pushing and whatever the diff touches, and `--no-verify` is out. I read that as a hard stop and wrote "the branch conflicts with main", which points at the wrong thing and reads like I was declining. I should have pushed the two keys and the restored test — they were verified and sitting in commits — and kept the handback to what actually needed you. Filed #7064 so the hook stops standing in the way of the next one.

On the account-field key: checked it, you're right there too. Put the key where I suggested and your Bluesky test fails with `divine.bsky.social` gone, so `_IdentityField` is the only placement that works. Stripped all three keys and all three tests go red. Also went over the rebase — no ARB key from `main` is missing anywhere, and the 27-key locale gap matches main's existing debt exactly. Threads closed, approved.

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup in `ead57dce`.

The earlier requested changes are resolved, the remaining small review debt I found has been cleaned up, the branch is rebased onto current `origin/main`, and all GitHub checks are green.

NotThatKindOfDrLiz:
@hm21 I took over the remaining review cleanup and pushed `ead57dce` to `feat/in-app-verify`.

What changed:

- Removed the unused `VerifierClient.oauthStatus` / `OAuthVerificationStatus` API and tests.
- Removed the unused `IdentityClaimsRepository.currentClaims` API and adjusted repository tests back onto production paths.
- Finished the crossposting OAuth rename cleanup by deleting the compatibility shims and folding the duplicate version-matrix coverage into the shared app-OAuth tests.
- Made `VerifiedAccountsRow` opt into the profile-header inset/centering so Edit Profile keeps its original alignment and padding; added scroll/layout coverage.
- Replaced the expected verifier-refusal `addError(StateError(...))` calls with `Log.warning`.
- Reused `VerifyState.keyOf` in `verify_screen.dart` instead of re-deriving claim keys inline.

I rebased the branch onto current `origin/main`, updated the PR description's verification section, and verified locally:

- `flutter test test/providers/crossposting_providers_test.dart test/widgets/profile/verified_accounts_row_test.dart test/screens/profile_setup/verified_accounts_section_test.dart test/screens/verify/verify_screen_test.dart test/screens/verify/verify_connect_screen_test.dart test/blocs/verify/verify_connect_cubit_test.dart`
- `flutter test test/src/verifier_client_test.dart` from `mobile/packages/verifier_client`
- `flutter analyze lib test`
- `flutter analyze` from `mobile/packages/verifier_client`
- `flutter analyze` from `mobile/packages/profile_repository`
- Pre-push hook on the rebased branch

GitHub checks are green. Please review the takeover commit and merge when you're ready.

NotThatKindOfDrLiz:
@hm21 I pushed one more takeover fix in `7506dc03f`.

What changed:

- `VerifyCubit` now returns quietly if the dashboard is closed while load or unlink async work is still in flight.
- `VerifyConnectCubit` now does the same across proof verification, OAuth start/browser return, and claim publishing.
- Added regression coverage for closing during load, unlink, verifier check, publish, and OAuth browser return.

Verification:

- `flutter test test/blocs/verify/verify_cubit_test.dart test/blocs/verify/verify_connect_cubit_test.dart`
- `flutter analyze lib test`
- Pre-push hook on the rebased branch: merge-conflict check, analyzer, generated-file verification, and changed-file tests all passed.
- GitHub checks are green on `7506dc03f`.

The branch was rebased onto fresh `origin/main` before the final push. GitHub currently reports it behind again because `main` moved while CI was running, but the mergeability check passed and there are no conflicts, so I am not resetting green CI just to refresh review-cleanup history. Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup in `7506dc03f`.

The remaining close-during-async crash path is fixed in both verify cubits and covered with regression tests. Local targeted tests, `flutter analyze lib test`, the pre-push hook, and GitHub checks are green. The branch was rebased before the final push; it is only marked behind because `main` moved again while CI was running, with mergeability still passing.

NotThatKindOfDrLiz (commented):
## Code Review: PR #7028

**Status:** Requires changes before approval

The PR implements in-app account verification linking/unlinking and resolves several design gaps. However, **four code defects block approval** — two are critical correctness bugs, one is a documentation/implementation mismatch, and one is a missing test for complex UI. Two findings require clarification from domain experts, and one minor issue needs addressing.

---

### 🛑 Blocking Defects (Must Fix)

#### 1. **Lagging-read merge expiration flaw** — Critical
**File:** `mobile/packages/profile_repository/lib/src/identity_claims_repository.dart`

The `_lastPublishedTags` cache expires after 1 minute with no backoff or expiration tracking. This creates a race condition:

1. User publishes identity tags with claim A → cached in `_lastPublishedTags`
2. Relay is slow, hasn't ACK'd yet
3. 61 seconds pass → cache expires
4. User immediately republishes → `_lastPublishedTags` is empty
5. The "already on relay" guard passes, publish proceeds
6. Relay finally ACKs the first publish AND receives the second
7. Both merges apply — the second overwrites the first

**Impact:** Identical claims can be republished into relay conflicts; data inconsistency.

**Fix:** Track expiration timestamps per relay, or implement exponential backoff for replays instead of a fixed timeout.

---

#### 2. **Account field disappears during keystroke** — Critical UX Bug
**File:** `mobile/lib/blocs/verify/verify_connect_cubit.dart`

In `_onAccountChanged`, the code clears `state.account` before updating it:
```dart
emit(state.copyWith(account: ''));  // Field clears visibly
// ... validate ...
emit(state.copyWith(account: account));  // Then reappears
```

**Impact:** Every keystroke briefly wipes the field, creating a jarring "text disappears then reappears" UX.

**Fix:** Combine into a single emit: `emit(state.copyWith(account: account, isValid: ...))` with validation happening before emit, not in between.

---

#### 3. **Tag preservation docstring contradicts implementation**
**File:** `mobile/packages/profile_repository/lib/src/identity_claims_repository.dart`

The `_mergeLaggingReads` docstring claims: *"Preserves claims tagged with keys in the published set."*

**Actual behavior:** The code **overwrites** all tags with the relay's version:
```dart
if (relayTags.isNotEmpty) {
  tags = relayTags;  // Replaces previous, doesn't preserve
}
```

**Impact:** False confidence in data integrity; future maintainers will implement "preserve" logic that doesn't match reality.

**Fix:** Update docstring to accurately reflect overwrite semantics, or implement actual preservation if that was the intent.

---

#### 4. **Missing widget test for 11 KB connect screen**
**File:** `mobile/lib/screens/verify/verify_connect_screen.dart`

The `VerifyConnectScreen` is 11 KB of complex UI:
- OAuth redirect handling with fallback to proof post
- Async verification polling  
- Multiple state-driven error cases
- Account input validation with visual feedback

**Current tests:** None (the PR does not include `verify_connect_screen_test.dart`).

**Impact:** No regression detection for UI layout, navigation, or state-driven rendering; refactors can silently break this flow.

**Fix:** Add widget tests covering:
- OAuth path rendering and button state
- Proof post fallback trigger
- Account field validation and error states
- Async verification polling UX
- Error state recovery

---

### ❓ Clarification Questions (Design/Domain)

#### 5. **Verifier pre-flight strategy**
**Files:** `mobile/packages/verifier_client/lib/src/verifier_client.dart`, OAuth flow

The `isOAuthConfigured()` method **sends an actual OAuth start request as a pre-flight check**. This means:
- Verifier logs every platform check as an OAuth attempt
- If a platform is not configured, the user may see an OAuth error mid-flow
- No caching of the pre-flight result

**Question for verifier owners:** Is this the intended behavior? Should the pre-flight be a separate low-cost metadata query instead of a full OAuth initiation? Should results be cached per platform?

---

#### 6. **Profile header vertical reordering**
**File:** `mobile/lib/widgets/profile/profile_header_identity.dart`

Verified accounts block moved from under the handle to the end of the column (after follower/following stats). This is an undocumented visible reordering that may conflict with design intent.

**Question for design:** Is this reordering intentional, or should verified accounts stay under the handle for visual hierarchy?

---

### ⚠️ Minor Issue

#### 7. **Missing tracking issue for transitional aliases**
**Files:** `mobile/lib/features/oauth/`, `mobile/lib/features/crossposting/`

`crossposting_callback.dart` and `crossposting_oauth_launcher.dart` are aliases with no removal plan. Per `code_style.md`, transitional code requires `// TODO(#issue):` tracking.

**Fix:** Add `// TODO(#XXXX): Remove after unified OAuth landing` comment, or file the tracking issue and cite it.

---

### Summary

**Blockers to merge:** Findings #1–4 must be fixed and tested. Findings #5–6 require clarification before approval. Finding #7 is a process nit.

**Next steps:**
1. Fix the four blocking defects above
2. Add widget test for `verify_connect_screen.dart`
3. Clarify verifier pre-flight and design intent with domain owners
4. Rebase onto fresh `origin/main` with `--force-with-lease` before re-pushing
5. Wait for CI to turn green before re-requesting review

PR #7031, opened 2026-08-10
feat(library): reach drafts from the recorder library

### LINKED ISSUE #7032: feat(library): reach drafts from the recorder library
Auto-created from PR #7031.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7031
**Branch:** `feat/library-redesign`

hm21:
Closes #7032

## Description

The recorder's library button opened a clips-only screen, so a draft started earlier was unreachable without leaving the recorder. It now opens the same tab bar as the standalone library minus Sounds, on the Clips tab — a saved sound is not something the current recording session can pick up, but a draft is.

Notable decisions:

- **Tabs are built from a list** (`_LibraryTab`) instead of being hardcoded three-wide. Selection mode (the editor's add-clips sheet) stays the one mode without a tab bar, and the toolbar actions plus the create-video bar now follow the active tab rather than the mode. Without that, sort/select/trash would show up over the drafts list.
- **Switching to Drafts no longer clears an auto-opened selection.** That selection mirrors the recorder session that opened the library, so exiting it would drop the session's clips — a user peeking at Drafts would lose their recording. A selection the user opened themselves still clears on tab change, as before.
- **Opening or posting another saved draft confirms before ending the active recording.** Cancelling keeps the live session intact. Reopening the active autosave from the standalone library does not show this warning because it resumes that same recording.
- **The in-progress autosave draft stays hidden here.** The session that owns it is the one showing the library, so offering it back as something to reopen makes no sense. `DraftsLibraryBloc` takes this as a constructor flag (`includeAutosaveDraft`, default `true`) so the filtering lives in the bloc, not in `DraftsTab`. The standalone library still lists it with its "in progress" badge.
- `LibraryTabsMode.clipsOnly` is renamed to `withoutSounds`, which is what the mode now means. The route path `/clips-only` and `RouteType.clipsOnly` are untouched — `RouteType.clipsOnly` was already documented as "clip library screen without Sounds tab", so renaming those would be churn on identifiers outside this change.

**Related Issue:** 

## Out of Scope

The recorder's library button is still gated on `hasClips`, so an account with drafts but zero clips cannot reach the library from the recorder at all. Enabling the button on drafts alone would need a draft count in the button's state and a rethink of its thumbnail and semantic label — worth a separate change if it comes up.

## Verification

- Scoped library, drafts, recorder-navigation, routing, and localization tests — 248 passing, one pre-existing web skip
- Design-system and test-structure ratchets for the touched paths — passing
- Current-head Mobile CI — generated files, format, analyze, and all four test shards passing
- New tests cover the Drafts + Clips tab set, autosave filtering, auto-opened and user-opened selection behavior, confirmation cancel/accept/no-session branches, Post, active-autosave resume, and toolbar semantics.
- Manually verified on a real Samsung Galaxy: opened the library from the recorder, both tabs present, no Sounds tab, autosave draft absent, clip selection survives a switch to Drafts and back.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


dcadenas (approved):
✅ Recorder drafts are reachable without silently dropping live work.

hm21:
@dcadenas thanks for digging into this — the finding was right, but I reverted the confirmation sheet in 35547fce and took a different route in 25aba25c. Reasoning, so you can push back if I've got it wrong:

**The session isn't actually lost.** `_openDraft` clears the clip manager with `keepAutosavedDraft: true`, and once the editor opens the tapped draft its `draftId` is that draft, so `getActiveDraft(isAutosave: ...)` writes autosaves under *that* id — `draft_autosave` is never overwritten. So "Opening a draft ends the video you're working on" tells the user something that doesn't happen.

**The app already asks this question, twice.** The editor won't close on an autosaved session without a decision (`PopScope(canPop: !isAutosavedDraft)` → save draft / discard / keep editing), and opening the camera on a fresh start offers the same session back (`_checkAutosavedChanges` → continue / discard). The autosave draft is a durable recovery point, not something that evaporates. A third prompt in the library is a different answer to a question the app already has one for.

**What was genuinely broken** is narrower: `_checkAutosavedChanges` only ran in `initState`, and the recorder stays mounted while the library sits on top of it — so on the way back it never re-runs, and the session would only resurface on the next app launch. That's the actual inconsistency, and it's what 25aba25c fixes: the check moves into `offerAutosavedSession` and runs again once the library pops. One deliberate difference from the cold-start path — coming back from the library leaves you on the recorder rather than continuing into the editor, since that's the screen you navigated back to.

Kept from your commits: the toolbar leading-icon fix and its test, and the selection-across-tabs tests (they drive the real auto-open path, better than what I had). Dropped the six l10n keys with the sheet — they were English-only across the other 17 locales and parked in `_knownUntranslatedDebt`.

Checks are green; still pending a run on a real device before this merges.

PR #7033, opened 2026-08-10
ci(deps): stop assigning every dependabot pr to one person

### LINKED ISSUE #6994: ci(review): track CODEOWNERS placement for dependency PRs
## Problem

`mobile/docs/CODEOWNERS` is not on a path GitHub reads, so Dependabot PRs cannot rely on CODEOWNERS review routing today. PR #6990 uses an explicit assignee as a temporary working handle for dependency PRs.

## Scope

Decide the repo-wide review policy and either move/add CODEOWNERS at a GitHub-recognized path (`.github/CODEOWNERS`, `CODEOWNERS`, or `docs/CODEOWNERS`) or choose a different durable ownership mechanism for dependency update PRs.

## Acceptance criteria

- Dependency update PRs have a durable reviewer/owner path that does not depend on one individual.
- The temporary Dependabot assignee workaround in `.github/dependabot.yml` is removed or documented as intentionally permanent.
- The chosen setup is verified against GitHub CODEOWNERS behavior.

Part of #3655 and follows up PR #6990.

hm21:
## Description

Dependency PRs no longer land on a single named assignee.

#6990 added `assignees: ["realmeylisdev"]` to `.github/dependabot.yml` as a temporary handle: `reviewers:` was removed from the Dependabot schema in favour of CODEOWNERS, and this repo's only CODEOWNERS file lives at `mobile/docs/CODEOWNERS` — a path GitHub does not read, listing `@agent-*` handles that are not real accounts. With no working routing mechanism, assignment was the only lever left, and the side effect was that one individual became the de-facto owner of all routine dependency churn.

This picks the decision #6994 asked for: no per-PR owner. Dependency PRs carry the `dependencies` label, show up in the PR list like anything else, and go through the normal team review flow. That is durable in the sense the issue meant — it does not depend on one person staying on the team or staying responsive.

Deliberately not adding a real `.github/CODEOWNERS` in this PR. That is a repo-wide review-policy change (it would start requiring a named owner's review on every PR touching the covered paths, not just Dependabot's), and it deserves its own discussion rather than riding along on a two-line cleanup. If the team later wants owner routing, the org already has a `@divinevideo/reviewers` team and scoping it to `/.github/` would cover Dependabot's PRs without touching everything else.

**Related Issue:** Closes #6994

## Out of Scope

- Adding or relocating a real CODEOWNERS file (see rationale above).
- Enabling the deferred `pub` ecosystem — still blocked on the upstream defects documented in the same file.

## Verification

- `.github/dependabot.yml` still parses as valid YAML and the `updates[0]` keys resolve to `package-ecosystem, directory, schedule, open-pull-requests-limit, commit-message, labels, groups`.
- No Dart files touched, so no app-side test or analyze run applies; nothing to verify on device for a config-only change.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


.github/dependabot.yml
```diff
@@ -20,12 +20,11 @@ updates:
...
-    # this repo has no CODEOWNERS file GitHub actually reads (mobile/docs/CODEOWNERS
-    # is not a path GitHub honours; tracked in #6994), so assignment is the only
-    # working handle.
-    assignees:
-      - "realmeylisdev"
+    # No `assignees:` on purpose — these PRs are reviewed like any other, off the
+    # `dependencies` label, rather than landing on one person. `reviewers:` was
+    # removed from the dependabot schema in favour of CODEOWNERS, so re-adding it
+    # here does nothing; routing to an owner would need a real CODEOWNERS file at
+    # a path GitHub reads (mobile/docs/CODEOWNERS is not one).
```
realmeylisdev, line 27: Worth naming what's actually left once the assignee goes: Dependabot can no longer request reviewers, there's no CODEOWNERS GitHub reads, and none of the 64 workflows here requests a reviewer or auto-merges. Meanwhile the org-level ruleset `Main branch protection` requires 1 approving review on `main` with no bypass actors — so a Dependabot PR can only merge once somebody happens to spot it in the list.

That may be exactly the trade you want, and it's a team call rather than something to fix here. The one thing I'd push back on is the framing: "the normal team review flow" isn't automatic for bot PRs, because the normal flow in this repo is an explicit `--reviewer divinevideo/reviewers` at creation and Dependabot never does that.

If the scoped CODEOWNERS ever gets picked up, `coop` and `divine-iac-coreconfig` already keep one at `.github/CODEOWNERS`.

realmeylisdev (approved):
Approving — the diff is right and the comment it leaves behind holds up.

I checked the claims rather than taking them: `reviewers:` was genuinely removed from the schema (8 Aug 2025, code owners are the stated replacement), `assignees:` is still supported and demonstrably live (#7006–#7008 carry one right now), and `mobile/docs/CODEOWNERS` really is unreachable — GitHub reads only `.github/`, root or `docs/`, and that file lists `@agent-*` handles that aren't accounts anyway. YAML parses, and `updates[0]` resolves to exactly the seven keys the description claims.

Three notes, none blocking:

- Nothing routes these PRs to anyone now, and `main` requires an approval — inline.
- `Closes #6994` closes the tracker on merge, but this PR deliberately defers the CODEOWNERS call that #6994's first and third acceptance criteria ask for. `Relates to #6994` would keep it open for that decision.
- #7006–#7008 keep their current assignee; the config isn't retroactive, so they need unassigning by hand if the point is that nobody owns them.

PR #7034, opened 2026-08-10
fix(feed): stop the fullscreen feed hanging on a loading placeholder when its source empties

### LINKED ISSUE #6949: task(e2e): make Maestro flows independent of live staging content

realmeylisdev:
Closes #6949.

## What this turned out to be

The issue filed `likeFlow` as blocked on live STAGING test data. It was not. `videoUnlike` was failing on **two app bugs in the fullscreen feed**, both of which surface as the same thing: a permanent branded loading placeholder that looks exactly like slow live content.

Reproduced on an iPhone 17 Pro simulator against STAGING, driving the flow by hand. Sixty seconds after unliking, the route's entire semantics tree was two nodes — `back_button` and a node labelled `"Loading"`. No `like_button`, so `assertVideoUnliked`'s 15-second wait for the "Like video" label could never succeed.

## The bugs

**1. The loading placeholder was also the empty state.** `PooledFullscreenVideoFeedScreen` renders it for `status == initial || !hasVideos`, and `FullscreenFeedBloc` emitted `status: ready` for an empty list. So "the source hasn't loaded yet" and "the source drained" were indistinguishable, with no timeout, no error branch and no way out but the back button. Unliking the only video in your own Liked feed hits this every time — the liked grid's bloc re-emits `[]` into the still-open route.

An empty *first* emit is a legitimate loading state (a live source warming up, pinned by an existing test), so the new `empty` status only latches once videos have actually arrived. It deliberately does **not** pop the route the way `emptyAfterRemoval` does, because a live feed can legitimately refill.

**2. The global feed bridge went blind during a reload.** `_GlobalFeedBridge` read `next.asData?.value`, which is null while a provider reloads even though the previous value is still attached. The Explore tabs render from `hasValue && value != null`, so through that window the grid kept showing tappable tiles while the bridge published nothing. The bridge is created lazily on the first `watchView` — i.e. at tap time — so a tap landing in that window left its `BehaviorSubject` unseeded, and an unseeded `BehaviorSubject` replays nothing.

riverpod sets `seamless: !ref.isReload`, and `isReload` is true exactly when a watched dependency changed — which `newVideosFeedProvider` gets from four version providers. This is a stall for the duration of the reload rather than a permanent hang, but `assertVideoPlayback` had no wait, so it was an instant E2E failure.

**3. Single-emit sources could never recover.** `StaticFeedRepository` hands out `Stream.value(...)` and the hashtag snapshot is `Stream.fromFuture(...)`. When that one emit is empty, nothing can ever arrive to clear the placeholder. Handled after `emit.forEach` returns, where the stream is provably finished.

## E2E changes

- `assertVideoPlayback`'s `assertNotVisible: text: "Loading..."` **could never fail** — that string belongs to the share menu and a dev-options screen, and it was marked `optional`. It now waits on the placeholder's own identifier and asserts the drained state separately, so a slow feed no longer fails instantly and a drained one fails with a message that says so.
- `videoUnlike` asserts the drained state, which is an app-owned outcome. Its "exactly one liked video" precondition is guaranteed by construction: `loginFreshInstall` creates the account, so it starts at zero likes and `videoLike` adds exactly one.
- `assertVideoUnliked` waited for a label the drained screen no longer renders and had no other caller — removed.
- `likeFlow` is back in the smoke suite.

## On the issue's premise

Worth recording, since it shaped the issue: **the New-grid path did not reproduce.** Opening the newest video worked, all five action buttons rendered, and the media resolved 206. Of the couplings the issue lists, only two are real, and both belong to #6952 — `searchUserFlow`'s account (plus its hardcoded `point: 50%,32%`, the more fragile half) and `commentFlow`'s commented-out cleanup. Accounts and likes are self-owned: each run creates its own account and `removeKeys` tears it down. The README now separates what is real from what was assumed.

## Verification

Every fix has a test that fails without it — the bridge test was confirmed by reverting the one-line change and watching it hang, then time out.

On device, after the fix, the same `Videos updated, count=0` now renders "Nothing left to play here" carrying `fullscreen_feed_empty`, and the whole `videoUnlike` sequence completes: unlike → drained state → back → liked tab → "No likes yet", zero `liked_video_thumbnail_0` nodes.

- `flutter analyze lib test` clean
- `test/blocs/fullscreen_feed/`, `test/repositories/feed_repository_impl_test.dart`, `test/screens/feed/`, `test/l10n/arb_consistency_test.dart` — 282 passing
- New ARB key mirrored into all 21 non-English locales

STAGING is left as found — the like published during the repro was unliked again.


hm21 (approved):
LGTM. Traced the drain path end to end and mutation-checked the new tests — reverting the bridge one-liner hangs its test into the timeout, reverting the bloc change fails four.

PR #7036, opened 2026-08-10
test(recorder): cover stop-motion mode with Maestro E2E

### LINKED ISSUE #7042: test(recorder): cover stop-motion mode with Maestro E2E
Auto-created from PR #7036.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7036
**Branch:** `test/maestro-stop-motion`

hm21:
Closes #7042

## Description

Extends the recorder E2E suite to the **stop-motion** mode, the follow-up to #7018's capture coverage and a sibling to #7043's lip-sync one. Open, drive the control rail, shoot two stills, undo them, close.

Stop-motion renders the capture stack, so it already inherited most of #7018's identifiers. Three controls are its own and were still only reachable by localized copy — the ghost-frame and grid toggles and the shot budget — so they get identifiers here, each with a widget test.

**Both mode asserts now pin the controls the other mode renders.** Not because either would otherwise pass on the wrong screen — the two rails are disjoint and the mode wheel's `selected` state settles that on the first line of every assert — but as a leak guard: a control that starts rendering in a mode which does not declare it fails a run instead of passing quietly. Nothing else in the suite catches that, and `video_recorder_capture_actions_test.dart` pins the same split at the widget level.

**The rail's three unconditional controls are now shared.** #7043 landed `utils/driveCaptureRail.yaml` after this branch's base, and 64 of `stopMotionModeControls`' 101 command lines were byte-identical to it. It cannot be reused as-is — it also drives timer and stabilization, neither of which stop-motion renders — so aspect ratio, lens and flash each get their own util and every mode sequences them in its own order. Inlining the new `runFlow`s reproduces both files' previous command sequences byte for byte, so nothing about what runs on device changed.

**`openStopMotionMode` selects its mode instead of asserting it**, unlike `openCaptureMode`. It has to: `VideoRecorderMode.fromName` falls back to `capture` whenever the persisted mode key is absent, so a first run after a `clearState` never opens on stop-motion by itself. The tap is safe from the two modes the flow can arrive in — Capture and Stop Motion are neighbours in the wheel's lazy `ListView` — and the file says so, along with why it would not be safe from Classic or Upload.

**The two shutter tests are deliberately coupled**, and the comments in both say so. A second still adds nothing new to assert on the screen: next and undo are already revealed by the first, and the only element tracking the count is the budget label, whose value falls out of `maxDuration` divided by the output frame rate — not something an E2E file should hardcode. So the count is proven one layer down. `stopMotionModeUndoFrame` undoes twice and asserts after the *first* undo that the chrome is still up, which is only true if two stills existed. Shortening either test to one tap silently removes the only evidence that the session accumulates.

The README picks up two findings from the device run that apply to `captureModeFlow` just as much:

- Maestro honours a `checked:` selector but **never prints it in the run log**. A correct assertion reads exactly like one whose flag was dropped, which is a good way to talk yourself into "fixing" a working selector. Proven applied by asserting `checked: true` against the toggle in its off state and watching it fail.
- The recorder persists **four** preferences — mode, lens, stabilization, grid — and some test asserts each one as a default. `clearState` handles this from the top of a flow, but a standalone run inherits the last session, so a phone last used on the front lens fails `stopMotionModeControls` on its first `.*Back camera.*`. That already applied to `captureModeControls` and was undocumented.

**Related Issue:** 

## Out of Scope

- **The assemble step** ("next", which collects the stills into a clip and opens the editor). It leaves the recorder, and the editor has no E2E coverage to hand off to yet.
- **Classic and upload modes.** Lip-sync landed in #7043 while this was open; this branch is rebased on top of it, and `utils/openRecorder.yaml` plus the `openCaptureMode` split turned out byte-identical to what that PR shipped, so they dropped out of this diff entirely.
- **Classic's own grid and ghost toggles**, which stay untagged: no flow drives them, and an identifier nothing asserts on drifts silently.
- **The smoke suite.** The flow stays out, same reason capture and lip-sync do — the shutter needs real camera hardware to leave its disabled state, and the smoke lane runs on the iOS Simulator.

## Verification

Manually verified on a real **Samsung Galaxy SM-S942B**, STAGING debug build, against the actual camera:

- `stopMotionModeOpen` — green, including the new `camera_stop_motion_budget` anchor and the mode-wheel tap.
- `stopMotionModeControls` — green, all five controls including flash on the back lens (not skipped).
- `stopMotionModeCaptureFrame` + `stopMotionModeUndoFrame` — green; library button confirmed empty afterwards, so the undo teardown really does drop the session's library row.
- Negative controls, both red as intended: `checked: true` against the off toggle, and shoot-once/undo-once against the undo test's first assertion.

The tests were driven test-by-test against an already-signed-in app rather than through `stopMotionModeFlow.yaml` end to end, because the device runs a German system locale and `loginFreshInstall`'s `clearState` drops the per-app English override mid-run. The login and `removeKeys` bookends of that flow are the ones `captureModeFlow` already runs unchanged. Noted in the README.

**A second device pass is pending** and it is wider than the first: `assertCaptureMode` gained three assertions and the shared rail moved, so `captureModeFlow` and `lipSyncModeFlow` need a run too, not just the stop-motion one.

Also:

- `flutter test test/widgets/video_recorder/` — 279 passing

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/e2e/maestro/README.md
+++ b/mobile/e2e/maestro/README.md
@@ -43,14 +43,15 @@ dispatches, and GitHub CI does not cover this lane at all.
 
 ## The recorder flows
 
-Two of the recorder's five modes are covered, one flow each:
+Three of the recorder's five modes are covered, one flow each:
 
 | Flow | Mode | Covers |
 |---|---|---|
 | `flows/captureModeFlow.yaml` | capture | open, drive the control rail, record a clip, delete it, close |
 | `flows/lipSyncModeFlow.yaml` | lip-sync | open, drive the control rail, prove the shutter refuses without a sound and records with one, delete the clip, close |
+| `flows/stopMotionModeFlow.yaml` | stop-motion | open, drive the control rail, shoot two stills, undo them, close |
 
-Classic, stop-motion and upload are not covered yet.
+Classic and upload are not covered yet.
 
 The gate is lip-sync's own behaviour and it takes **both** shutter tests to
 cover it. `lipSyncModeAudioGate` proves the shutter refuses with no sound
@@ -65,20 +66,21 @@ picker's **Divine** tab, whose entries come from
 is the relay-backed one, and it is the live-content dependency this file names
 elsewhere as the suite's main source of flakiness — no flow goes near it.
 
-Every file in both flows has been run green on a Galaxy SM-S942B. The lip-sync
-tests were driven test-by-test against an already-signed-in app rather than
-through `lipSyncModeFlow.yaml` end to end, because the device runs a German
-system locale and `loginFreshInstall`'s `clearState` drops the per-app English
-override mid-run (see step 0b). The login and `removeKeys` bookends of that
-flow are the same ones `captureModeFlow` already runs.
+Every file in all three flows has been run green on a Galaxy SM-S942B. The
+lip-sync and stop-motion tests were driven test-by-test against an
+already-signed-in app rather than through their flow end to end, because the
+device runs a German system locale and `loginFreshInstall`'s `clearState` drops
+the per-app English override mid-run (see step 0b). The login and `removeKeys`
+bookends of both flows are the same ones `captureModeFlow` already runs.
 
-Neither is in **`smoke.yaml`**, and both are deliberately off the iOS lane:
+None of them is in **`smoke.yaml`**, and all three are deliberately off the iOS
+lane:
 
 - **They need real camera hardware.** Without a camera the bloc never reports
   `isCameraInitialized`, the record button stays disabled, and the recording
-  test hangs on its first wait. The lens switch in the shared rail needs one
-  too. The iOS Simulator has no camera at all; an Android emulator's virtual
-  scene camera is enough, and a device is best.
+  and shutter tests hang on their first wait. The lens switch in the shared
+  rail needs one too. The iOS Simulator has no camera at all; an Android
+  emulator's virtual scene camera is enough, and a device is best.
 - **The `…Open` tests alone** — the chrome assertions — do pass without a
   camera, because the viewfinder falls back to a placeholder and every control
   still renders. So does `lipSyncModeAudioGate`: the blocked shutter is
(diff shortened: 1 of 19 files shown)


mobile/e2e/maestro/README.md
```diff
@@ -27,76 +27,160 @@ The `e2e-smoke-ios` and `e2e-smoke-android` Codemagic workflows are
...
+|---|---|---|
+| `camera_timer_button` | ✅ | — |
+| `camera_stabilization_button` | ✅ | — |
+| `camera_ghost_frame_button` | — | ✅ |
+| `camera_grid_button` | — | ✅ |
+| `camera_stop_motion_budget` | — | ✅ |
+
+Assert only the controls a mode *has* and the two sets nest, so either assert
+would pass on either screen. That is what the `assertNotVisible` blocks in
+both files are for, and `video_recorder_capture_actions_test.dart` pins the
```
realmeylisdev: `assertCaptureMode.yaml` isn't in this PR, and it has no assertion about `camera_ghost_frame_button`, `camera_grid_button` or `camera_stop_motion_budget`. Its only `assertNotVisible` entries are `audio_chip` (from #7043) and the next/undo pair for session state. So "both asserts pin both halves" isn't true yet — only `assertStopMotionMode.yaml` has the block.

The sentence above needs a second look too. The table right there shows the two sets are *disjoint*, not nested: capture asserts `camera_timer_button` and `camera_stabilization_button`, neither of which stop-motion renders, so `assertCaptureMode` already fails on a stop-motion viewfinder with no help. #7043 needed its `audio_chip` line because lip-sync renders the identical rail — that case genuinely nested, this one doesn't.

Either way is fine (add the three lines, or reword), the file and the doc just need to agree.

mobile/e2e/maestro/asserts/assertStopMotionMode.yaml
```diff
@@ -0,0 +1,58 @@
...
+    id: camera_grid_button
+- assertVisible:
+    id: camera_aspect_ratio_button
+- assertVisible:
+    id: camera_switch_camera_button
+
+# And the two capture mode has that this mode does not: stop-motion declares
+# neither `supportsCountdownTimer` nor `supportsVideoStabilization`. Without
+# this pair, assertCaptureMode's rail set is a subset of this one and both
+# asserts would pass on either screen.
```
realmeylisdev: Same as the README note: capture's rail set isn't a subset of this one. It has `camera_timer_button` and `camera_stabilization_button`, which this mode doesn't render, so `assertCaptureMode` already fails here without the pair below.

Keeping the two `assertNotVisible` lines is still worth it — they catch an id leaking across modes — but the reason given isn't the reason they work.

mobile/e2e/maestro/tests/stopMotionModeControls.yaml
```diff
@@ -0,0 +1,174 @@
...
+# - **Real camera hardware**, for the lens switch. See
+#   stopMotionModeCaptureFrame.
+# - **A grid that starts on.** The grid preference persists
+#   (`camera_grid_lines_enabled`) and defaults to on for grid-supporting modes,
+#   so step 2 reads it as on. Run from the top this is safe —
+#   loginFreshInstall's clearState wipes the key.
+#
+# Unlike capture mode, an empty session is *not* required: the aspect-ratio
+# control is disabled by clips, and stop-motion stills do not become a clip
+# until the assemble on "next". The order below still matches capture mode's,
```
realmeylisdev: This file does require an empty session: its last line runs `assertStopMotionMode.yaml`, which asserts `camera_next_button` and `camera_delete_clip_button` are gone. The `given` on line 6 says "with no stills" as well.

The rail itself doesn't need one — which is what I think you meant — but as written this reads as "safe to run mid-session", and anyone reordering the flow on the strength of it gets a red run at the very last step.

mobile/e2e/maestro/tests/stopMotionModeControls.yaml
```diff
@@ -0,0 +1,174 @@
...
+    timeout: 10000
+- tapOn:
+    id: camera_grid_button
+- extendedWaitUntil:
+    visible:
+      id: camera_grid_button
+      checked: true
+    timeout: 10000
+
+# 3. Aspect ratio flips between two states.
```
realmeylisdev, line 81: From here to the end of the flash guard is byte-identical to `utils/driveCaptureRail.yaml` on `main` — its steps 1, 3 and 4. 64 of this file's 101 command lines, aspect ratio and lens switch and flash verbatim.

That util landed in #7043 after this branch's base, so it wasn't there when you wrote this. It can't be dropped in as-is either, since it also drives timer and stabilization and stop-motion renders neither. So: pull the shared aspect/lens/flash part out into a util both call, or keep the copy with a line saying why. Worth deciding now rather than when a fourth mode shows up.

mobile/e2e/maestro/tests/stopMotionModeCaptureFrame.yaml
```diff
@@ -0,0 +1,44 @@
...
+#
+# Needs real camera hardware. Without it the bloc never reports
+# `isCameraInitialized`, the record button stays disabled, and the tap does
+# nothing — so this cannot pass on the iOS Simulator. Run it on an Android
+# emulator (which has a virtual scene camera) or on a device.
+#
+# Hold-to-record does not apply here: the record button disables long-press
+# whenever the mode captures stills, so the `HoldToRecordPreferenceService`
+# preference captureModeRecordClip has to warn about cannot affect this one
+# either way.
```
realmeylisdev: A word dropped out — this currently reads "the `HoldToRecordPreferenceService` preference captureModeRecordClip has to warn about cannot affect this one either way". Probably wanted "the preference *that* `captureModeRecordClip` has to warn about".

realmeylisdev (changes requested):
Checked the claims in here against the source and ran it locally: `flutter analyze` clean, 279 recorder widget tests green, and the recorder chrome does come up on the iOS Simulator with no camera. The capability flags, the `AnimatedOpacity` gating on next/undo, the four persisted preferences and the `droppable()` shutter all hold up. Good tests, and the two new negative widget tests are the right guard for an id leaking into the wrong mode.

Two things need a call from you, two are small and inline.

**The README describes an `assertCaptureMode.yaml` that doesn't exist.** "Both asserts pin both halves", "the `assertNotVisible` blocks in both files" — this PR doesn't touch that file, and it says nothing about ghost frame, grid or the budget. And the two rail sets are disjoint rather than nested, so `assertCaptureMode` already fails on a stop-motion viewfinder at `camera_timer_button`. That's the opposite of #7043, where the rails were identical and `assertNotVisible: audio_chip` really was the only separator. Add the three lines or reword — just make the file and the prose agree.

**#7043 moved the rail out from under you.** It extracted `utils/driveCaptureRail.yaml` and cut `captureModeControls.yaml` to 26 lines; 64 of the 101 command lines in `stopMotionModeControls.yaml` are now byte-identical to it. It can't be reused as-is — it also drives timer and stabilization — so it's a shape question: pull the shared aspect/lens/flash part into its own util, or keep the copy deliberately. Either way the merge here is structural, not just the textual one on the README you called out.

The branch is conflicting on `README.md` and `semantic_ids.dart` now, so all of it lands in the same pass.

hm21:
@realmeylisdev All five check out — I verified each against `main` rather than taking them as read.

**The README and `assertCaptureMode.yaml`** — right, and the miss is bigger than a missing block. The two rails are disjoint, not nested, so each assert already fails on the other's viewfinder on the rail alone — and the mode wheel's `selected` state settles which one is up before either gets that far. The block is a leak guard, not a mode separator. Fixed both halves in 0b61a318c: the three lines are in `assertCaptureMode.yaml` now, and the prose says what the block actually does. Same correction in `semantic_ids.dart` and in the widget test's comment, which carried the same wrong reason.

**`stopMotionModeControls`** — correct, and the `given` on line 6 already contradicted it. The precondition now says the rail doesn't need an empty session but the mode assert it closes on does.

**`driveCaptureRail`** — extracted per control in fb0bea1f7: `driveAspectRatioControl`, `driveLensControl`, `driveFlashControl`. `driveCaptureRail` sequences those three plus timer and stabilization; `stopMotionModeControls` calls them plus ghost frame and grid. Both keep their exact previous order, so nothing about what runs on device changed — inlining the new `runFlow`s reproduces `driveCaptureRail`'s old command sequence byte for byte, same for `stopMotionModeControls`. Calling the shared rail with parts skipped was the alternative and it doesn't survive classic.

**The dropped word** — fixed.

Rebased on top of #7043 in the same pass. `utils/openRecorder.yaml` and the `openCaptureMode` split were byte-identical to what you landed, so they dropped out of this diff entirely; only `README.md` and `semantic_ids.dart` needed resolving, and that section of the README is three flows now rather than two. Description updated to match.

CI is green on the new head — analyze, format, all four test shards, generated files. Locally: 279 recorder widget tests and `check_refs.sh` clean.

Device run still pending and it is a wider one than before: `assertCaptureMode` gained three assertions and the shared rail moved, so `captureModeFlow` and `lipSyncModeFlow` need a pass too, not just the stop-motion one. I'll update here once that's green.

Your threads are left unresolved so you can close them as you verify.

mobile/e2e/maestro/README.md
hm21: Right on both counts. `assertCaptureMode.yaml` has the three lines now (0b61a318c) and the prose is reworded: the rails are disjoint rather than nested, so neither assert could pass on the other screen anyway, and the mode wheel's `selected` state settles it before the rail is even reached. The block is a leak guard. That section of the README was rewritten wholesale in the #7043 rebase, so it now covers three flows.

mobile/e2e/maestro/asserts/assertStopMotionMode.yaml
hm21: Agreed — kept the pair, replaced the reason (0b61a318c). It reads as a leak guard now, and says explicitly that `assertCaptureMode` fails here on `camera_timer_button` regardless. `assertCaptureMode` got the mirror of it.

mobile/e2e/maestro/tests/stopMotionModeControls.yaml
hm21: Correct, and the `given` on line 6 already said the opposite. The precondition list now carries an explicit "empty session" entry: the rail itself does not need one, the mode assert this closes on does, which is why the flow still runs it before the shutter tests. fb0bea1f7.

mobile/e2e/maestro/tests/stopMotionModeControls.yaml
hm21: Extracted per control rather than as one shared block, so both files keep their own order: `utils/driveAspectRatioControl`, `driveLensControl`, `driveFlashControl`. `driveCaptureRail` sequences those three plus timer and stabilization, this file calls them plus ghost frame and grid. Inlining the new `runFlow`s reproduces both files' previous command sequences byte for byte, so what runs on device is unchanged. fb0bea1f7.

mobile/e2e/maestro/tests/stopMotionModeCaptureFrame.yaml
hm21: Fixed in 0b61a318c — "the preference that captureModeRecordClip has to warn about".

realmeylisdev (approved):
lgtm. inlined the runflows both sides and the command sequences come out identical, so the extraction didn't change what runs on device. the three assertNotVisible lines in assertCaptureMode are the only real device delta now, so that's what the capture and lipsync passes need to clear.

PR #7039, opened 2026-08-10
test(e2e): repair the search and comment smoke flows

### LINKED ISSUE #6952: task(e2e): repair the remaining smoke flows (search, comment)

realmeylisdev:
## Description

Repairs `searchUserFlow` and `commentFlow` and puts both back in the smoke suite. Everything below was checked against a running build on an iPhone 17 Pro Max simulator pointed at STAGING, not read off the source.

**Search — rewritten, not patched.** The three-tab `TabBar` the flow described no longer exists; the screen is a filter pill plus sliver sections. Every `Users (n) / Tab 2 of 3` selector was dead, `Find something cool...` is a `divine_ui` default no call site uses (the hint is `Search...`), and the result tap was `point: 50%,32%`.

The app gains the anchors that made a rewrite possible, all `Semantics(identifier:)`, none of them user-visible:

| Anchor | Why |
|---|---|
| `search_user_tile_$index` | Rows are keyed by pubkey, which a flow cannot know ahead of time. Sits beside the pubkey id, not instead of it. |
| `search_section_header_{people,lists,tags,videos}` | The header is also the "see all" control the flow uses to reach the People-only filter. |
| `search_field`, `search_back_button`, `search_filter_pill` | Replace matching translated copy and a `point: 6%,10%` tap. |
| `explore_search_bar` | `tapOn: Search` cannot match `Search...` — Maestro matches the whole label. |
| `comments_sheet_title` | See below. |

`DivineSearchBar` gains an optional `semanticIdentifier`, matching `DivineButton` / `DivineIconButton` / `DivineListTile`.

**Comments.** Three defects, each verified on device:

1. `commentVideo.yaml:32` asserted the bare `${FULL_COMMENT}` — the `env` default, which is `""`. An empty pattern matches an empty-text node, so the flow's only proof-of-post assertion passed without ever looking at the comment. It now reads `${output.FULL_COMMENT}`, anchored as `.*${output.FULL_COMMENT}`: a posted row merges timestamp, the "You" marker and the body into one accessibility label, and a bare match against that label fails.
2. `deleteComment.yaml:27` tapped a `Delete comment` confirmation that does not exist. Picking the option deletes immediately; the step failed with `Element not found`. The orphaned `assertDeleteCommentModal.yaml` that described that dialog is deleted.
3. `deleteComment.yaml:38` closed the player with `Close video player`, a `VideoDetailScreen` label — the Explore grid pushes `PooledFullscreenVideoFeedScreen`. Cleanup now runs before the back navigation, since it reopens the sheet and needs the video on screen.

The issue expected `close_comments_button` and `comments_drag_handle` to be usable. They are not: `CommentsHeader` and `CommentsDragHandle` have no production call site — the sheet is a `VineBottomSheet` — and they survived only because `comments_screen_test.dart` builds a replica of the sheet body and put them in it, so `renders CommentsHeader` passed against the test's own tree. Both widgets are deleted and the replica no longer renders what the real sheet does not.

Dismissal now drags the sheet by `comments_sheet_title`. The count text otherwise merges into a node spanning the entire sheet, so the old `swipe from "\d+\sComments?"` started in the middle of the list, scrolled, and left the sheet open — the swipe reported `COMPLETED` and the next assertion failed.

**One product bug, found on this path.** `TextField.onChanged` does not fire for programmatic controller writes, and the comments sheet clears the composer that way after a post. `_hasText` stayed true, so the send arrow sat over an empty field and the video-reply button — same slot, renders only while the field is empty — never came back until the user typed again. Fixed by tracking the controller; the regression test fails on `main` and passes here.

**Also, unrequested:** `check_refs.sh` only matched `../`-prefixed references, so `assertProfileFresh.yaml:11` was never checked. `loginFreshInstall` now retries the Explore switch, whose tap races the post-signup warm-up, and `commentVideo` waits for the grid to fill — both flakes reproduced while verifying, and re-enabling two flows triples how often that path runs.

**Action needed before dispatching the workflow:** `MAESTRO_SEARCH_USER` must be added to the `maestro_e2e_credentials` Codemagic group. Both runners now fail fast without it rather than letting Maestro resolve `undefined`.

**Related Issue:** Closes #6952

## Out of Scope

- `likeFlow` stays held back on #6949.
- `searchTags` / `searchVideos` / `utils/searchWorkaround.yaml` are `fullRegression`-only and still rotted; untouched.
- Profile tabs expose no visible label, only an identifier — #6951.
- **Comment deletion is not durable, and re-enabling the cleanup does not change that.** Measured on STAGING: the app publishes a valid kind 5, the relay stores it, and the relay keeps serving the deleted kind 1111. A deleted comment is gone from the poster's screen and still visible to everyone else, so each smoke run still leaves a comment behind. That is a product/infra call, not a flow fix — documented in `deleteComment.yaml` and the README, and tracked in #7048.
- `comment_item.dart` decides "is this my comment" from a cache-only `NostrService.publicKey` read, so a cold-start session whose signer has not settled would offer Flag/Block on your own comment. Not reproduced here; tracked in #7049 rather than fixing it blind.

## Verification

- `flutter test test/screens/comments/ test/screens/search_results/ test/screens/explore/` — 246 passing.
- `packages/divine_ui`: `flutter test --coverage` — 872 passing, 100.00% line coverage.
- `maestro test e2e/maestro/flows/commentFlow.yaml` — exit 0, zero failed steps, on an iPhone 17 Pro Max simulator against STAGING. Run three times.
- `maestro test -e SEARCH_USER=... e2e/maestro/flows/searchUserFlow.yaml` — exit 0, zero failed steps. Run twice.
- `bash e2e/maestro/scripts/check_refs.sh` — passes, and now fails on a planted same-directory bad reference.
- `loginEmailPwd` unverified locally (no credentials on this machine); untouched by this change.

## Type of Change

- [x] 🧹 Code refactor
- [x] 📝 Documentation


NotThatKindOfDrLiz:
@realmeylisdev takeover fixes are pushed in 14bd59b4f.

What changed:
- fixed check_refs.sh so single-quoted Maestro runFlow/file references are checked again
- removed the orphaned commentsHeaderTitle/commentsHeaderCloseLabel ARB keys and regenerated l10n outputs
- added local widget assertions for comments_sheet_title and explore_search_bar
- tightened searchUsers.yaml so the People filter assertion proves the header disappeared
- restored numeric follower/following assertions in assertUserProfile.yaml
- filed the two PR-body follow-ups as #7048 and #7049 and linked them from the description

Verification:
- bash e2e/maestro/scripts/check_refs.sh
- flutter gen-l10n
- flutter test test/screens/comments/comments_screen_test.dart test/screens/explore_search_transition_test.dart test/screens/search_results/widgets/search_results_app_bar_test.dart
- pre-push hook: analyze plus changed-file tests, 136 passing
- GitHub checks are all green at 14bd59b4f

Please review the takeover commit and merge when ready.

NotThatKindOfDrLiz (approved):
Approved after takeover fixes in 14bd59b4f. I rechecked the review findings, pushed the scoped fixes, verified locally, and GitHub CI is green. @realmeylisdev please review the takeover commit before merging.

hm21 (approved):
LGTM

PR #7043, opened 2026-08-10
test(recorder): cover lip-sync mode with Maestro E2E

### LINKED ISSUE #7045: test(recorder): cover lip-sync mode with Maestro E2E
Auto-created from PR #7043.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7043
**Branch:** `test/maestro-lip-sync`

hm21:
Closes #7045

## Description

Extends the recorder E2E suite to the **lip-sync** mode, alongside #7018's capture coverage. Open, drive the control rail, prove the shutter refuses to record without a sound and records once one is picked, delete the clip, close.

### The audio gate, from both sides

The gate is lip-sync's own behaviour and it takes two tests to cover it. `lipSyncModeAudioGate` proves the shutter refuses with nothing picked — which on its own is also exactly what a permanently broken shutter does. `lipSyncModeRecordClip` picks a sound, watches the gate open, and records against it. They run in that order and cannot be swapped: the gate test needs the editor to hold no sound, and the record test is what puts one there.

The gate test asserts two things, and both are needed because they catch different failures: the snackbar proves the gate *fired*, the still-visible close button proves no recording started behind it. The snackbar goes by copy because `DivineSnackbarContainer` takes no identifier and adding one drags `divine_ui`'s 100% coverage gate along with it — the same call `openRecorder` already makes for the education sheet.

### Picking a sound is deterministic

Only the picker's **Community** tab is relay-backed. The **Divine** tab lists the 94 sounds in `assets/sounds/sounds_manifest.json`, shipped with the app, so this path has no live-content dependency at all — no flow goes near Community.

The test searches "Bruh Sound Effect" and asserts the title back off the first result, so a reordered or renamed manifest fails there instead of silently recording against something else. That title comes from the manifest rather than l10n, so unlike the rest of the copy in this suite it reads the same in every locale. It is chosen for being 864ms — under the 6.3s maximum clip length, so confirming pops straight back rather than routing through the trim screen a longer sound would need.

Three new anchors carry it: the search field, the result tiles (indexed, because the list is a search result and the test takes index 0 rather than depending on manifest order) and the confirm button. Each gets a widget test, including one asserting the tile's identifier stays null when the editor's audio row reuses the same widget without one.

One device-only finding is baked into the test and the README: the confirm button is anchored to the bottom of the sheet, so the search keyboard pushes it off screen entirely. `hideKeyboard` goes between typing and touching the list. Without it the failure reads badly — the sound selects correctly and the tile turns green in the failure screenshot, while the run dies waiting on a button that is not in the hierarchy at all.

### The part worth reviewing: capture and lip-sync are the same screen

Lip-sync *is* the capture stack. Same close, next, delete, library button, record button, and the same five rail controls in the same order, because it declares the same countdown-timer and stabilization support. The **only** element that differs is the sound picker it puts in the top bar's center slot, which capture leaves empty.

Two consequences, both in the diff:

- **`assertCaptureMode` gains an `assertNotVisible` on the chip.** Assert only what a mode renders and capture's set is everything lip-sync has minus one, so that file passes on the Lip Sync tab. That single line is what keeps it honest. `video_editor_audio_chip_test.dart` pins the anchor itself at the widget level.
- **`captureModeControls`' rail moves to `utils/driveCaptureRail.yaml`**, and both modes' control tests are that util plus their own mode assert. Duplicating those 130 lines and the reasoning in their comments was the alternative. `openCaptureMode`'s recorder-open half moves to `openRecorder` for the same reason — the one-shot "Why six seconds?" sheet handling now lives in one place for both open utils.

`captureModeControls` was re-run green on device after the extraction.

### Mode selection

`openLipSyncMode` *selects* its mode rather than asserting it like `openCaptureMode` does. It has to: `VideoRecorderMode.fromName` falls back to `capture` whenever the persisted key is absent, so a first run after a `clearState` never opens here. It walks the wheel one entry at a time rather than tapping Lip Sync directly — the wheel is a lazy `ListView` and Lip Sync sits two places from Capture, further than a single tap can rely on.

### One finding that applies to `captureModeFlow` too

The recorder persists **three** preferences — mode, lens, stabilization — and some test asserts each as a default. `clearState` handles it from the top of a flow, but a standalone run inherits the last session, so a phone last used on the front lens fails the rail on its first `.*Back camera.*`. That already applied to `captureModeControls` and was undocumented; it is in the README now.

**Related Issue:** 

## Out of Scope

- **The trim screen.** Confirming a sound longer than the maximum clip length routes through it; the flow picks a short sound so it never appears.
- **Classic, stop-motion and upload modes.** Stop-motion is covered by #7036, which targets `main` independently of this one; the two can merge in either order. They both add `utils/openRecorder.yaml` with the same content and both rewrite the recorder section of `e2e/maestro/README.md`, so whichever lands second needs a small merge there — the alternative was one combined PR.
- **The smoke suite.** The flow stays out, same reason capture does — the lens switch in the shared rail needs real camera hardware, and the smoke lane runs on the iOS Simulator.

## Verification

Manually verified on a real **Samsung Galaxy SM-S942B**, STAGING debug build off this branch, against the actual camera:

- `captureModeControls` after the rail extraction — green, stabilization included, plus the new `audio_chip` absence assertion.
- `lipSyncModeOpen` / `lipSyncModeControls` / `lipSyncModeAudioGate` / `lipSyncModeRecordClip` — green, including the wheel walk, the gate's snackbar, and the full pick-a-sound-and-record path. Flash exercised for real on the back lens (not `SKIPPED`).

Driven test-by-test against an already-signed-in app rather than through `lipSyncModeFlow.yaml` end to end, because the device runs a German system locale and `loginFreshInstall`'s `clearState` drops the per-app English override mid-run. The login and `removeKeys` bookends of that flow are the ones `captureModeFlow` already runs unchanged. Noted in the README.

Also:

- `flutter test test/widgets/video_recorder/ test/widgets/video_editor/ test/screens/video_editor/` — 1187 passing

## Type of Change

- [x] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore


NotThatKindOfDrLiz (approved):
@hm21 I pushed a small takeover follow-up on top of your PR and rebased it onto current main.

What changed:
- Broadened `check_refs.sh` so it now validates quoted, unquoted, same-directory, and `../` `runFlow:` YAML refs.
- Hoisted the shared capture-stack record/delete choreography into `utils/recordCaptureClip.yaml` and `utils/deleteLastRecorderClip.yaml`, then reused it from capture and lip-sync flows.
- Added the missing `audioSearchField` widget test and marked the identified search semantics node as a text field.
- Tightened the lip-sync no-sound gate precondition and cleaned up the stale `assertLipSyncMode`/semantic-id docs.

Verification:
- `bash mobile/e2e/maestro/scripts/check_refs.sh`
- `flutter analyze lib/constants/semantic_ids.dart lib/widgets/video_editor/audio_editor/audio_selection_bottom_sheet.dart test/widgets/video_editor/audio_editor/audio_selection_bottom_sheet_test.dart`
- `flutter test test/widgets/video_editor/audio_editor/audio_selection_bottom_sheet_test.dart test/widgets/video_editor/audio_editor/audio_editor_selection_overlay_test.dart test/widgets/video_editor/audio_editor/audio_list_tile_test.dart test/widgets/video_editor/audio_editor/video_editor_audio_chip_test.dart`
- GitHub CI is green on head `249ee2b422ef5fcecfd1310895a45a4ac9126597`.

NotThatKindOfDrLiz:
## Review: Lip-Sync Mode E2E Coverage

I've completed an independent analysis of this PR. Below are 6 findings—3 blocking, 3 should-fix—that need addressing before merge.

### Blocking Issues (must fix)

**Finding 1: Relative `runFlow` refs bypass `check_refs.sh` validation**

- **Location:** `mobile/e2e/maestro/utils/openLipSyncMode.yaml:21` and `mobile/e2e/maestro/utils/openCaptureMode.yaml:20`
- **Issue:** Both use `runFlow: openRecorder.yaml` (same-directory reference without `../`). The `check_refs.sh` script only matches refs starting with `../`, so these refs are never validated. If someone renames `openRecorder.yaml`, CI passes but flows fail at runtime.
- **Fix:** Change both to `runFlow: ../utils/openRecorder.yaml`

**Finding 2: Duplicate assertion logic defeats flow extraction**

- **Location:** `mobile/e2e/maestro/tests/lipSyncModeRecordClip.yaml` steps 4–5
- **Issue:** These steps duplicate `captureModeRecordClip.yaml` and `assertCaptureClipRecorded.yaml` verbatim—including comments and the exact same assertions. The whole point of extracting `driveCaptureRail.yaml` was to eliminate this duplication.
- **Fix:** Replace steps 4–5 with:
  ```yaml
  - runFlow: ../asserts/assertCaptureClipRecorded.yaml
  ```

**Finding 3: Missing widget test for `audioSearchField` semantic**

- **Location:** PR body states "each gets a widget test" but `audioSearchField` (added to `audio_selection_bottom_sheet.dart:592`) has no corresponding test.
- **Issue:** `audio_list_tile_test.dart`, `video_editor_audio_chip_test.dart`, and `audio_editor_selection_overlay_test.dart` all have semantic identifier tests, but `audio_selection_bottom_sheet_test.dart` is missing one for the search field.
- **Fix:** Add a test group to `audio_selection_bottom_sheet_test.dart`:
  ```dart
  group('semantics', () {
    testWidgets('search TextField has audioSearchField identifier', (tester) async {
      await tester.pumpWidget(buildTestWidget());
      final semantics = tester.getSemantics(find.byType(TextField));
      expect(semantics.identifier, equals(SemanticIds.audioSearchField));
    });
  });
  ```

### Should-Fix Issues

**Finding 4: Missing `textField: true` on search field Semantics**

- **Location:** `mobile/lib/widgets/video_editor/audio_editor/audio_selection_bottom_sheet.dart:592`
- **Issue:** The Semantics wrapper has `identifier:` but omits `textField: true`. Precedent in the codebase (e.g., `invite_gate_screen.dart:405–408`) sets both. Screen readers benefit from the flag.
- **Fix:** Change to:
  ```dart
  Semantics(
    identifier: SemanticIds.audioSearchField,
    textField: true,
    child: TextField(...)
  )
  ```

**Finding 5: Assertion precondition violation in test flow**

- **Location:** `mobile/e2e/maestro/asserts/assertLipSyncMode.yaml` (header claims "no sound is picked") vs. `mobile/e2e/maestro/tests/lipSyncModeRecordClip.yaml` (teardown leaves sound selected)
- **Issue:** `assertLipSyncMode.yaml` asserts the viewfinder is idle "when no sound is picked," but `lipSyncModeRecordClip` leaves the picked sound in place after the clip is recorded. This violates the stated precondition.
- **Fix:** In `mobile/e2e/maestro/tests/lipSyncModeAudioGate.yaml`, add an explicit precondition check:
  ```yaml
  - assertVisible:
      text: ".*Add audio.*"
  ```
  This ensures the audio chip shows "Add audio" (i.e., no sound is picked) before testing the shutter gate.

**Finding 6: Missing doc comments on new semantic constants**

- **Location:** `mobile/lib/constants/semantic_ids.dart`
- **Issue:** `audioSoundTile` and `audioSelectionDoneButton` constants have no `///` doc comments, unlike neighboring constants.
- **Fix:** Add single-line doc comments:
  ```dart
  /// Identifier for the audio sound tile in the selection list.
  static const String audioSoundTile = 'audio_sound_tile_';

  /// Identifier for the done button in audio selection.
  static const String audioSelectionDoneButton = 'audio_selection_done_button';
  ```

---

### Summary

Findings 1–3 are critical—they affect test durability and maintainability. Findings 4–6 improve accessibility and code clarity. Once these are addressed, please rebase onto fresh `origin/main` (the branch is currently behind).

Ready to review the revision.

PR #7050, opened 2026-08-11
fix(upload): keep the resumable session alive on a chunk 408

rabble:
## Summary

A resumable chunk PUT returning **408** was demoting the entire upload to the legacy single-request PUT of the whole file. This keeps the resumable session alive instead.

Two commits, red then green: the failing test first, then the fix.

## The bug

Reproduced, not inferred. The mocked call sequence before the fix:

```
1. head  media.divine.video/upload            <- capability probe
2. post  media.divine.video/upload/init       <- session init
3. put   .../sessions/up_408  data: [1,2,3,4,5]    <- chunk 1 OK, offset -> 5
4. put   .../sessions/up_408  data: [6,7,8,9,10]   <- 408, exactly ONCE
5. put   media.divine.video/upload  data: Instance of '_FileStream'
```

Line 5 is the defect: after one 408 on a **5-byte** chunk, the client streams the **entire file** to the legacy endpoint. On a link too slow to finish one chunk inside the edge timeout, that cannot succeed — it is strictly harder than what just failed.

### Why it never hit the transient-error path

A 408 is not an exception here. `validateStatus: statusCode < 500` admits it, so Dio returns it as a normal `Response`:

1. The chunk retry loop only retries inside `on DioException catch`, so it is bypassed entirely — note line 4 appears **once**, not three times. `_maxChunkRetries` and `_chunkRetryDelay` are dead code for a 408.
2. 408 is not in the session-expired set `{401, 403, 404, 410}`, so it falls to the generic non-2xx branch.
3. That branch left `isRecoverableResumableFailure` at its default `false`, and a non-recoverable resumable failure routes to `uploadLegacyFallback` — `_uploadToServer` with no `resumableSession`.

The asymmetry that makes this look unintentional: the only place setting `isRecoverableResumableFailure: true` is the `DioException` branch, gated on `_hasTransientDioSignal` (5xx or transport timeout). So Dio's own `receiveTimeout` firing is **recoverable**, while an HTTP 408 reporting the same event is **fatal**. Same physical failure, opposite handling, depending only on whether the timeout was noticed locally or reported by the server. `_retriableUploadStatusCodes` already lists `408, // Request timeout` — the intent was there; the fallback decision just never consulted it.

## The fix

Mark statuses already in `_retriableUploadStatusCodes` as recoverable, so the session survives and the banked offset stays usable. 5xx cannot reach this branch (it throws), so in practice this covers **408 and 429**.

**Deliberately no in-loop retry.** Unlike a 5xx, a 408 means the server already spent its full timeout waiting; stacking two more same-length waits would just delay the failure by minutes on exactly the connections that are struggling. The outer retry policy already treats 408 as retriable and backs off exponentially, which is the right place for it. The test pins the attempt count at 1 so a future in-loop retry has to be a deliberate change.

## Validation

- New test fails on the parent commit and passes on the fix — both halves verified independently (attempt count, and the `verifyNever` on the legacy PUT).
- Full package suite: **234 passed**, 3 pre-existing skips.
- Coverage: **982/982 = 100.00%**, satisfying this package's `min_coverage: 100` gate.

## Relationship to the chunk-size change

Complementary, not redundant. divinevideo/divine-iac-coreconfig#1580 lowers the advertised chunk from 8 MiB to 1 MiB, which makes 408s **rare**. This PR decides what happens when one lands anyway. Shipping only the chunk-size change would leave the trap armed.

## What I could NOT verify

- **Not run on a device.** The evidence is a mocked reproduction of the call sequence, not a real slow-network upload.
- **Not tied to a user report.** No production trace links a specific failed upload to this path — upload failures currently emit no telemetry (`TIMEOUT` is filtered out before Crashlytics, and there is no upload-success event for a denominator). So this is a defect proven by construction, not a measured cause of the reports.
- **The downstream recovery is inferred.** I verified the session is preserved and the legacy demotion no longer fires. I did not trace the outer retry end-to-end to confirm a subsequent attempt actually resumes from the banked offset rather than re-initialising.

## Note for a follow-up

`.claude/rules/testing.md` lists `divine_ui` as the only strict-coverage package, but `blossom_upload_service` also sets `min_coverage: 100` in its workflow. Worth correcting so the next person does not discover it from a red CI run.

### DIFF
--- a/mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
+++ b/mobile/packages/blossom_upload_service/lib/src/blossom_upload_service.dart
@@ -1446,10 +1446,22 @@ class BlossomUploadService {
           final xReason =
               response.headers.value('X-Reason') ??
               response.headers.value('x-reason'); // coverage:ignore-line
+          final statusCode = response.statusCode;
+          // A transient status leaves the session intact: the offset has not
+          // moved and the same chunk can be resent. Falling back to the
+          // legacy whole-file PUT would be strictly worse on a link that just
+          // failed to finish a single chunk. 5xx never lands here — it throws
+          // a DioException — so in practice this covers the edge timeout
+          // (408) and rate limiting (429). Retrying is left to the outer
+          // policy, whose exponential backoff suits a server that has already
+          // spent its own timeout waiting.
           throw BlossomResumableUploadException(
-            'Chunk upload failed: ${response.statusCode} '
+            'Chunk upload failed: $statusCode '
             '${xReason ?? response.data}', // coverage:ignore-line
-            statusCode: response.statusCode,
+            statusCode: statusCode,
+            isRecoverableResumableFailure:
+                statusCode != null &&
+                _retriableUploadStatusCodes.contains(statusCode),
           );
         }
 
--- a/mobile/packages/blossom_upload_service/test/src/blossom_upload_service_test.dart
+++ b/mobile/packages/blossom_upload_service/test/src/blossom_upload_service_test.dart
@@ -1390,6 +1390,164 @@ void main() {
         await tempDir.delete(recursive: true);
       });
 
+      test('does not fall back to legacy PUT when a resumable chunk returns '
+          '408 Request Timeout', () async {
+        const testPublicKey = _testPublicKey;
+        final sessionUrl = Uri.parse(
+          'https://upload.divine.video/sessions/up_408',
+        );
+
+        when(() => mockAuthProvider.isAuthenticated).thenReturn(true);
+        when(
+          () => mockAuthProvider.createAndSignEvent(
+            kind: any(named: 'kind'),
+            content: any(named: 'content'),
+            tags: any(named: 'tags'),
+          ),
+        ).thenAnswer(
+          (_) async => _signedEvent(
+            testPublicKey,
+            24242,
+            const [],
+            'Upload video to Blossom server',
+          ),
+        );
+
+        final tempDir = await Directory.systemTemp.createTemp(
+          'blossom_resumable_408_test_',
+        );
+        final videoFile = File('${tempDir.path}/video.mp4')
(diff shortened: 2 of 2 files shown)


hm21 (approved):
Verified the bug and the fix. Test red on `6ec0a39d16`, green on `fa00e8f7a0`, with exactly the call sequence you describe. Package suite 234/3 skips, analyze clean, coverage 982/982.

Pushed one fix in `f8ad46a7de`: the recoverable branch returns the result built in the `catch` of `_uploadWithSingleLegacyFallback`, which sets only `errorMessage` and `failureReason` — no `statusCode`. With `fromStatusCode(408)` mapping to `unknown`, a chunk 408 surfaced as `statusCode: null, failureReason: unknown`.

`_isTransientUploadResult` classifies on `statusCode`, so the image path lost both recoveries at once — `_uploadWithRetry` stopped after one attempt *and* the legacy fallback no longer ran. Reachable via `video_metadata_update_service.dart:255`; the publish path passes `allowResumable: false`. Worth noting your "strictly worse" argument doesn't hold for a single-chunk thumbnail: the whole-file PUT is the same bytes there. On the video path the retry survived, but only through `upload_retry_policy`'s "unknown errors are retriable by default" fallthrough — its `code == 408` branch never fired.

The commit propagates the status symmetrically with `_classifyUploadException`, which also closes the same pre-existing gap for the 5xx `DioException` recoverable throw. Regression test pins all three properties (no legacy demotion, 408 survives, retry budget spent); each half verified red independently. 235 passed, coverage still 100.00% (986/986), plus 239 app-layer upload tests. All checks green.

One gap left alone as out of scope: a transport-timeout recoverable throw still carries no status (`failureReason: network`), and `_isTransientUploadResult` doesn't check `network` — so that case is still invisible to `_uploadWithRetry`. Pre-existing, and fixing it means touching the classifier rather than the call site.

Also: the body has no `Closes #` line, which the linked-issue bot flagged. Yours to merge.

PR #7052, opened 2026-08-11
test(notifications): unpin the row layout test from font load order

### LINKED ISSUE #7053: test(notifications): unpin the row layout test from font load order
Auto-created from PR #7052.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7052
**Branch:** `test/notification-rows-font-order`

hm21:
Closes #7053

## Description

`test/goldens/widgets/notification_rows_golden_test.dart` fails intermittently on CI with `A RenderFlex overflowed by 104 pixels on the bottom` — the last assertion of the max-font case, `expect(tester.takeException(), isNull)`.

It is not flakiness in the usual sense, it is a run-order dependency. `VineTheme` renders through **google_fonts**, which registers its families (`Roboto`, `BricolageGrotesque_800`) asynchronously. Whichever test runs first in the isolate therefore lays out with the fallback test font, whose glyphs are roughly twice as wide, so every row re-wraps into about twice the height:

| | after any warm-up | as the first test in the isolate |
|---|---|---|
| scenario column height | 984 px | **1504 px** (surface is 1400 px) |
| comment row text block | 280 px | 560 px |

Over 1400 px the flex overflows and the test fails. In the merged optimizer isolate CI runs with `--test-randomize-ordering-seed random`, so whether this file draws the first slot is a per-run lottery — hence "sometimes".

Reproducible locally, no CI needed:

```
cd mobile && flutter test --test-randomize-ordering-seed 1 test/goldens/widgets/notification_rows_golden_test.dart
```

Seeds 1, 4, 5 fail; 2, 3, 6 pass. Same for `--plain-name "max-font"`, which forces the max-font case to run first.

The fix hangs the scenario in a `SingleChildScrollView`, so the column's height is unbounded and the test no longer depends on font metrics at all. Raising the surface to a taller number instead would only be a magic constant that the next copy change blows past. The stacking assertions are unaffected — the whole column is still laid out, so the relative positions they compare stay exact — and `takeException()` still guards against real errors.

**Related Issue:** 

## Out of Scope

Four test files (`test/screens/welcome_screen_test.dart`, `test/screens/key_management_screen_test.dart`, `test/screens/sound_detail_screen_test.dart`, `test/screens/apps/app_detail_screen_test.dart`) call `loadAppFonts()` in their own `setUp`. That is process-global and irreversible, so it shifts text metrics for every test that runs after them in the bundle — the same class of problem with a much wider radius. Not touched here.

## Verification

- `flutter test --test-randomize-ordering-seed {1..8} test/goldens/widgets/notification_rows_golden_test.dart` — green on all eight, including the three seeds that failed before the change.
- `flutter test --plain-name "max-font"` and `--plain-name "default layout"` — each case green in isolation, i.e. in the worst case where it runs first with the fallback font.
- `flutter test --dart-define=DIVINE_GOLDEN_TESTS=true test/goldens/widgets/notification_rows_golden_test.dart` — green with the real fonts loaded.
- `dart format` and `flutter analyze` clean.

Test-only change, no app code touched, so there is nothing to verify on device.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


mbradley (approved):
Reviewed the diff and verified the fix locally:

- Ran the file on seeds 1, 4, 5 (the reported failures) — all green.
- Reverting the `SingleChildScrollView` reproduces the exact `RenderFlex overflowed by 104 pixels` failure, confirming the root cause is the fixed-surface + fallback-font width, not flakiness.
- Inverted the max-font stacking assertion with the fix in place — it goes red (`button.dy` still resolves), so unbounding the height doesn't make the layout assertions unfailable.

The stacking/inline checks key off the row width, which is unchanged, so they still catch a real responsive-layout regression.

PR #5158, opened 2026-06-14
feat(image): add @visibleForTesting cache seam to VineCachedImage (#3137)

realmeylisdev:
Part of #3137. Infrastructure for unwinding the **image-cache cluster** of
`skip_very_good_optimization` tags (the profile grids, notification avatar,
ios_network_image — ~7 tagged files plus 3 already-merged order-fragile
widget tests).

## Root cause

`VineCachedImage` hard-codes the global `openVineImageCache`. Pumping it with a
non-empty URL kicks off an eager `MediaCacheImageProvider` resolve whose
`_loadAsync` awaits `getFileFromCache()` → `getApplicationSupportDirectory()`
(path_provider) and **settles after the test completes**, polluting the merged
VGV test isolate — the image analogue of a leaked timer. That late async is why
those widget tests were tagged out of the optimizer.

## The seam (Option B — minimal, behaviour-neutral)

- Add a `@visibleForTesting MediaCacheManager? debugImageCacheOverride;` and an
  `_activeImageCache` getter; `_imageProvider` resolves through it.
- **Production behaviour is unchanged** — the override defaults to `null`, which
  resolves to `openVineImageCache`. The `const VineCachedImage(...)` constructor
  and all 21 call sites are untouched.
- Widget tests can now set `debugImageCacheOverride = createMockMediaCacheManager()`
  in `setUp` / clear it in `tearDown`, so the eager resolve hits a stubbed cache
  and never touches the real on-disk/path_provider path.

Includes a unit test covering both override paths (null → `openVineImageCache`,
set → the injected manager).

## Why seam-only (no untags in this PR)

The seam neutralises the image-cache leak (verified: the previously-tagged
`profile_liked_grid` passes at seed 42 with the override). However, untagging
those files still trips a **separate, unrelated** ordering fragility in
already-merged `path_provider`/`ProVideoEditor` tests
(e.g. `video_clip_import_service_test`), which is outside the image-cache scope.
So this lands the seam as behaviour-neutral infrastructure; the per-file untags
follow once those cascade victims are de-flaked.

## Verification

`very_good test --optimization --concurrency=2 --exclude-tags integration` on the
merged suite at seeds 42, 11111, 98765: **all green** (behaviour-neutral change).

PR #7054, opened 2026-08-11
test(explore): keep the popular-tab suite off the real image cache

### LINKED ISSUE #7055: test(explore): keep the popular-tab suite off the real image cache
Auto-created from PR #7054.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7054
**Branch:** `test/popular-tab-image-cache-timer-flake`

hm21:
Closes #7055

## Description

`PopularVideosTab > reports a slow variant switch once, not once per rebuild` fails intermittently in CI. The failure is not the mock count — it is a pending timer:

```
A Timer is still pending even after the widget tree was disposed.
Pending timers: Timer (duration: 0:00:10.000000, periodic: false)
  #5 CacheStore._scheduleCleanup (package:flutter_cache_manager/src/cache_store.dart:125)
  #6 CacheStore._getCacheDataFromDatabase
  #7 CacheStore.retrieveCacheData
```

Seen on [run 31452806707](https://github.com/divinevideo/divine-mobile/actions/runs/31452806707) (shard 1/4) and [run 31430265454](https://github.com/divinevideo/divine-mobile/actions/runs/31430265454) (shard 2/4), same stack both times.

The chain: the grid paints thumbnails through `VineCachedImage`, which resolves against the process-global `openVineImageCache`. Its `CacheStore` answers the lookup off a real file read and then arms a 10s cleanup timer — created inside the test's fake-async zone, so it never fires and trips the teardown assert. This test is the only one in the file that waits on the wall clock (`tester.runAsync(Future.delayed(20ms))`, needed because the slow-load threshold is measured with `DateTime.now()`), and that real time is exactly what lets the file read land mid-test.

Why it was intermittent rather than constant: `CacheStore` latches `_scheduledCleanup` per store instance, and the store is process-global. In the merged VGV optimizer isolate only the *first* test to complete that lookup ever gets a timer, so the blame lands wherever the shard split happens to put it — hence the same failure appearing under different shards, and never locally on a single-file run.

The fix routes the suite through the existing `debugImageCacheOverride` seam (#5158), the same way `video_thumbnail_widget_test.dart` and `pooled_video_error_overlay_test.dart` already do, so no real cache-manager work runs here at all. It sits at group level rather than inside the one failing test, because the cross-fade test paints the same grid and is equally liable to be the one that arms the timer.

**Related Issue:** 

## Out of Scope

The same trap exists for any other widget suite that pumps real grids or feeds without setting the override — the flake can move to one of those instead of disappearing repo-wide. Making the stub the default in `flutter_test_config.dart` would close that off, but it needs an opt-out in `vine_cached_image_test.dart` (which asserts `provider.cacheManager` is `same(openVineImageCache)`), so it is deliberately left out of this PR.

## Verification

- `flutter test test/widgets/popular_videos_tab_test.dart` — green, run repeatedly
- `flutter analyze` on the changed file — no issues
- `dart format` — clean
- Confirmed the mechanism rather than guessing it: a throwaway probe pumping the same widget tree with a stubbed cache recorded exactly one `getFileFromCache` call, so without the override that call reaches the real `CacheStore` that CI's stack trace blames.
- The failure does not reproduce locally on a single-file run (5× green before the fix, also with the wall-clock wait raised to 500ms) — expected, since it needs the merged-isolate state. Green-after-fix is therefore not the evidence here; the CI stack trace plus the probe are.
- No device test: this is a test-only change with no app-code diff, so there is nothing to verify on hardware.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


realmeylisdev (approved):
Approving — I re-derived the diagnosis rather than taking it on faith, and it holds.

The chain is real in source: `VineCachedImage` → `MediaCacheImageProvider._loadAsync` → `getFileFromCache` → `CacheStore.retrieveCacheData` → `_getCacheDataFromDatabase` → `_scheduleCleanup()` → `Timer(10s)`, which is exactly the frames in the CI trace. One detail that sharpens your writeup: `_scheduleCleanup` early-returns while `_scheduledCleanup` is non-null, and the timer only clears that field when it *fires* — which it never does under fake async. So per merged isolate exactly one suite can ever be blamed, whichever gets there first. That explains the shard-hopping even more tightly than "latches per store instance".

What I checked:

- Probed the seam with a recording cache: the popular grid really does reach `getFileFromCache` through `debugImageCacheOverride`. Since `_activeImageCache` is `debugImageCacheOverride ?? openVineImageCache`, that is precisely the call you're intercepting.
- The fixtures use `example.com`, so `_shouldBypassCacheManager` is false and the `VineCachedImage` branch is the live one — the override sits on the real path, not the raw `NetworkImage` path.
- Hand-built a stand-in for the VGV optimizer over seven image-touching suites, including `vine_cached_image_test` and its `same(openVineImageCache)` assertion: 97 tests green, nothing stranded by the group-scoped set/reset.
- `debugImageCacheOverride:null` is already a sanctioned reset-to-default entry in `check_process_global_mutations.sh`, so the teardown is the shape CI wants.

The local non-repro is expected rather than a gap in the evidence — it matches what I saw when this same test failed on unrelated PRs back in early August.

Group-level placement is the right call; the cross-fade test paints the same grid and is equally able to be the one that arms the timer.

PR #7058, opened 2026-08-11
ci(mobile): run Mobile CI on merge_group events

### LINKED ISSUE #7059: ci(mobile): run Mobile CI on merge_group events
Auto-created from PR #7058.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7058
**Branch:** `ci/merge-queue-support`

hm21:
Closes #7059

## Description

Prepares `Mobile CI` to work as the gate for a GitHub merge queue. This PR **does not enable the queue** - that is a ruleset change and needs repo admin rights. It only makes the workflow able to report on a queue entry, which has to land on `main` *before* anyone flips the setting, otherwise the queue would sit waiting for a check that never starts.

Why we're looking at this at all: `Mobile CI` failed on `main` 36 times between Jun 29 and Aug 11, spread over ~20 days. Only 3-4 of those came from a branch being out of date (`_titleStyle` on Jul 31, generated files on Aug 1, `diagnosticTag` on Aug 7) - the rest were flaky tests. The strict "branch must be up to date" rule we turned on Aug 8 catches that first group, but it charges every PR a manual update-and-wait cycle to do it, and at ~21 merges a day someone else usually lands first and restarts the cycle. A merge queue gets the same guarantee without the manual loop.

Three changes:

- **`merge_group` trigger.** `Mobile CI` is the only required status check on `main`, so it is the one that has to report on the queue's speculative merge commit.
- **Scope detection for queue entries.** The existing detector reads changed files from the PR files endpoint, which has no equivalent in a `merge_group` event. It now reads the queue's own diff via the compare endpoint (`base_sha...head_sha`) and checks out only the extracted detector script, so the changed-file list stays API-driven without a full repository checkout. A queued docs-only change skips the app matrix exactly as it does on a PR. The queue head is a speculative merge of `main` plus every PR ahead of it in the batch, so that diff is precisely what is about to land.
- **Simplified per-job gates.** The `github.event_name != 'pull_request' ||` clause existed to force the full matrix on push. The scope step already reports `app=true` for every non-PR event, so dropping the clause leaves push and pull_request behavior identical while extending the gate to `merge_group`.

An unusable compare answer - empty, or at the compare endpoint's 300-file cap - falls open to the full matrix. A pull request over the files endpoint's 3000-file cap, or whose files listing is empty/truncated relative to `.changed_files`, also falls open. The queue run is the last gate before `main`, so a truncated API response must never silently skip it.

**Related Issue:** #7059

## Out of Scope

- **Enabling the merge queue.** Needs a `merge_queue` rule added to the "CI workflow rule" ruleset by someone with admin rights. Order matters: merge this first, then enable.
- **Package workflows** (`divine_ui`, `db_client`, ...) keep only their `pull_request` trigger. They are not required checks today, so the queue's gate is unchanged from what `main` is protected by right now.
- **Deciding whether to keep the strict up-to-date rule.** Separate discussion; this PR just makes the automated option available.

## Verification

The queue path cannot be exercised before the queue exists, so the scope detector was extracted from the YAML and tested against a stubbed `gh` for every event shape and truncation boundary:

| case | result |
|---|---|
| `merge_group`, app code changed | `app=true native=false` |
| `merge_group`, docs only | `app=false native=false` |
| `merge_group`, native config | `app=true native=true` |
| `merge_group`, empty API answer | `app=true native=true` (falls open) |
| `merge_group`, 300-file API answer | `app=true native=true` (falls open) |
| `push` to main | `app=true native=true` (unchanged) |
| `pull_request`, app code changed | `app=true native=false` (unchanged) |
| `pull_request`, docs only | `app=false native=false` (unchanged) |
| `pull_request`, native config | `app=true native=true` (unchanged) |
| `pull_request`, over 3000 files | `app=true native=true` (falls open) |
| `pull_request`, empty files response | `app=true native=true` (falls open) |
| `pull_request`, truncated files response | `app=true native=true` (falls open) |
| `pull_request`, detector script changed | `app=true native=true` |

Also: `actionlint` is clean, `bash -n` and `shellcheck` are clean on the extracted detector, the focused detector test passes locally, and no `github.event_name != 'pull_request'` gate is left behind. The current-head GitHub run is green, including generated files, format, analyze, all four test shards, VGV Tag Gate, and the final `Mobile CI` gate. This PR touches `.github/workflows/**`, so its own CI run exercises the full app matrix including the native guards.

Not device-tested - CI configuration only, no app code changes.

## Type of Change

- [x] Build configuration change


.github/workflows/mobile_ci.yaml
```diff
@@ -25,33 +27,58 @@ jobs:
...
-          # scoped change, so fall open to the full matrix rather than
-          # under-running CI on a truncated list.
-          changed_total=$(gh api "/repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}" --jq '.changed_files')
-          if [ "$changed_total" -gt 3000 ]; then
-            echo "app=true" >> "$GITHUB_OUTPUT"
-            echo "native=true" >> "$GITHUB_OUTPUT"
-            echo "PR touches $changed_total files (> 3000 API cap); running full app CI."
-            exit 0
-          fi
+          case "${GITHUB_EVENT_NAME}" in
```
NotThatKindOfDrLiz: This step now owns pull_request, merge_group, fallback behavior, two GitHub API response shapes, cap handling, and the app/native classification that decides whether the required Mobile CI matrix runs. Please extract this detector into a script under `mobile/scripts/ci/` and add targeted tests under `mobile/test/tools/` for at least PR app/docs/native cases, merge_group app/docs/native cases, empty compare output, and the 300-file compare cap. That keeps the merge-queue invariant test-backed instead of relying on manual shell verification.

NotThatKindOfDrLiz (changes requested):
I pushed a narrow reviewer fix for the compare endpoint cap and the merge_group branch filter. I am still requesting changes because this required-check scope detector has grown into consequential multi-event CI routing logic and needs to be extracted and regression-tested before merge.

dcadenas (approved):
✅ Merge-queue scope detection is fail-open and test-backed.

### Changes I pushed

- Extracted the scope detector into a standalone script loaded through a narrow sparse checkout.
- Added coverage for PR and merge-group app, docs, native, empty-response, and API-cap paths.
- Kept the timing-budget classification invariant pointed at the extracted detector.
- Corrected the description: compare responses cap at 300 files, PR file responses cap at 3000, and only the detector is checked out.

### Checks

- All current-head GitHub checks passed, including generated files, analyze, format, four test shards, and the final Mobile CI gate.
- `actionlint`, shell syntax, format, and analyze passed locally.
- Focused detector and timing-budget tests: 26 passed.
- Complete optimized Flutter suite: 15,527 passed; 285 declared skips.
- The repository validator passed its runnable format, analyze, and six package checks.

The repository validator also reported 27 coverage/runnability gaps:

- Uncovered roots (5): repository root, iOS, Divine UI, Profile Repository, and Verifier Client.
- CI-only declared roots (4): repository root, Divine UI, Profile Repository, and Verifier Client.
- Ancestor-only coverage (4): iOS, Divine UI, Profile Repository, and Verifier Client.
- Declared but not locally runnable (14): Divine UI reusable CI, mergeability reusable CI, Codex-config shell harness, package-channel isolation, production `Future.delayed`, test-unit structure, package coverage floor, raw-icons ceiling, VGV tag count, integration-only tags, the Mobile CI upstream gate, Profile Repository reusable CI, semantic PR reusable CI, and Verifier Client reusable CI.

The five shell ratchets and two VGV tag gates above were run directly and passed. The direct Codex-config harness stopped at its fixture that expects Dart to be absent from `/usr/bin:/bin`; current-head Codex Configuration passed in GitHub. The declared service-integration directory command built and installed the app, then failed while loading existing tests because Flutter's integration binding was already initialized before Patrol's binding; it does not cover the CI-only files changed here.

NotThatKindOfDrLiz:
@hm21 I pushed a takeover fix at `23ea2731a` and rebased the branch onto current `origin/main`.

What changed:
- normalized `wc -l` counts so the merge-group fall-open logs are portable on macOS and Linux
- made pull-request scope detection fall open when the files API returns an empty or truncated list relative to `.changed_files`
- made changes to `mobile/scripts/ci/detect_mobile_ci_scope.sh` run the native checks
- removed the undocumented `branches: [main]` filter from the `merge_group` trigger
- updated the focused detector tests and PR body for the new cases

Verification:
- `flutter test test/tools/detect_mobile_ci_scope_test.dart`
- `bash -n mobile/scripts/ci/detect_mobile_ci_scope.sh`
- `shellcheck mobile/scripts/ci/detect_mobile_ci_scope.sh`
- `actionlint .github/workflows/mobile_ci.yaml`
- pre-push hook analyzer and changed-file tests passed
- current-head GitHub checks are green, including the final `Mobile CI` gate

NotThatKindOfDrLiz (approved):
Takeover fixes are pushed and verified. The detector now falls open for empty/truncated PR file listings, normalizes line counts for portable logs, treats detector edits as native-sensitive, and uses the bare documented `merge_group` trigger shape. Current-head GitHub checks are green. Ready for @hm21 to review the takeover commit and merge when ready.

NotThatKindOfDrLiz:
## Review findings — blocking issues

### 1. Compare endpoint file cap is 300, not 3000 — truncation guard is dead code and can silently skip app CI

The `merge_group` path guards against `file_count -ge 3000`, but the compare endpoint caps at **300 files** regardless of `--paginate`. Pagination walks commits, not files — so `per_page=100 --paginate` returns the same 300 entries every time. Result: the guard can never fire, and a queued batch that alphabetically sorts app-code files past the 300-file mark will silently skip the entire app matrix.

**Proof:** I reproduced this with a realistic batch (`.agents/`, `.claude/`, `docs/` files + app code = 302 total):
```
TRUTH (what actually lands):       app=true native=false
WHAT THE WORKFLOW SEES (300 cap):  app=false native=false
Silently skipped: mobile/lib/main.dart, mobile/test/widget_test.dart
```

**Fix:** Change line 71 from `[ "$file_count" -ge 3000 ]` to `[ "$file_count" -ge 300 ]`, and correct the comment (lines 66–69) to state the actual cap.

This defeats the PR's own stated invariant: _"a bad API response must never silently skip it."_

### 2. Scope detector is now 90 lines of untested inline bash — against repo precedent for extracting and testing this kind of logic

This repo has established a pattern: `mobile/scripts/ci/select_test_shard.sh`, `check_ci_timing_budget.py`, etc., each paired with a test in `mobile/test/tools/`. The scope detector doubles in complexity (function, case dispatch, two API shapes, two cap checks) and is verified only by a one-time manual exercise documented in the PR body.

**Why this matters:** The stub verification in the PR body can't catch truncation bugs — a stub returning "300 files" looks identical to a real truncation at 300. Finding #1 slipped through exactly because of this.

**Recommended:** Extract to `mobile/scripts/ci/detect_ci_scope.sh` with `mobile/test/tools/detect_ci_scope_test.dart`, making the PR's verification table a permanent regression gate. This also makes the workflow self-testing: since `.github/workflows/*` sets `app=true`, the test is part of the CI it controls.

### 3. Nit: `merge_group:` on line 8 has no `branches: [main]` filter

Both `push` and `pull_request` are scoped to `main`. Harmless today, but asymmetric if a second protected branch ever appears.

---

**Verdict:** #1 is a real regression path that can land unverified app code on `main`. It only fires after the queue is enabled (a separate step), so this PR alone isn't dangerous — but enabling the queue on top of it is. Send back for both fixes.

PR #7060, opened 2026-08-11
feat(subtitles): time captions against the video instead of typed seconds

### LINKED ISSUE #7061: feat(subtitles): time captions against the video instead of typed seconds
Tracks PR #7060.

## Problem

The subtitle editor asks creators to edit cue start and end times as decimal-second text fields. That makes timing corrections hard because the creator cannot see the video frame or scrub the video while adjusting a cue.

## Scope

- Replace numeric timing fields with video-timed caption editing.
- Show a preview player and filmstrip timeline in the subtitle editor.
- Let cue timing come from the same range-slider row used by the video editor captions flow.
- Keep the cue list usable with and without a playable video URL.
- Preserve subtitle save/load behavior and complete localization coverage.

## Done When

- The editor can scrub and seek against the video while correcting captions.
- Selecting a cue seeks the preview to that cue.
- HLS videos do not poison the shared media cache key with manifest files.
- Temporary filmstrip frames are cleaned up when the editor closes.
- Keyboard avoidance works for both staged and no-stage editor states.
- Relevant subtitle editor, caption row, and thumbnail tests cover the new behavior.

hm21:
Closes #7061

## What

The subtitle editor asked creators to retype start and end times as decimal seconds, with nothing on screen to check them against. Correcting an auto-generated track meant guessing.

The screen now shows the video with a scrubbable filmstrip timeline, and the cue rows moved onto a sheet over it. Timing is set with the range slider the video editor's captions sheet already uses. Selecting a row jumps the picture to that cue, so you see the frame the caption lands on.

## Why these choices

- **The captions row is shared, not copied.** `_CueRow` and `_InputSurface` came out of `video_editor_captions_sheet.dart` into `CaptionCueRow`, with the field label and delete label as parameters so each editor keeps its own copy. Writing a caption before publishing and correcting one afterwards now work the same way and cannot drift apart.
- **The timeline shows, it does not edit.** No cue bars, no trim handles — it says which frame the playhead is on, and the sliders do the timing. That kept the video editor's own timeline widgets untouched.
- **The playhead runs off a `Ticker`.** The player reports its position a few times a second, which scrolls the film in visible steps. The ticker projects the last reported position by elapsed time every frame and re-anchors on each report — the same idea as the editor's `playTimeNotifier`.
- **`CodecHeavySurfaceGuard` without waiting for the entrance transition.** The screen opens a hardware decoder on mount, so the feed behind it has to hand one back first. Without this the picture stayed black on a Galaxy while the decoder logged three parallel h264 instances and then `MediaCodec discarded an unknown buffer`. This is the same reason the video editor overrides the flag.
- **`resizeToAvoidBottomInset` is off** and the sheet carries the keyboard inset itself, so the keyboard pushes the sheet up over the picture instead of squeezing the picture.
- **Frames come through the media cache.** A published video has no local file, so `SubtitleTimelineThumbnailService` caches it and then runs the same `VideoThumbnailService.generateStripThumbnails` the clip strip uses. A failed fetch leaves the timeline usable without frames rather than surfacing an error.

## l10n

`subtitleEditorStartLabel` and `subtitleEditorEndLabel` are gone with the numeric fields. Added and translated into all 21 non-English locales: `subtitleEditorPlayPreview`, `subtitleEditorPausePreview`, `subtitleEditorPreviewUnavailable`.

## Takeover update

Reviewer takeover commit `89ff917ed` fixes follow-up issues found during review:

- Aligns filmstrip frame lookup to the actual timeline axis width.
- Rejects stale paused player reports while a seek is pending.
- Resolves HLS before caching timeline thumbnails and refuses unresolved manifests.
- Deletes generated filmstrip frame files when the stage replaces or disposes them.
- Restores keyboard avoidance for the no-stage cue list.
- Makes the sheet initial size explicit, fixes stale comments, and lets tapping a row select its cue.

## Test plan

- `flutter test` over the subtitle editor, captions sheet, video editor timeline, l10n consistency and the new service — 515 passing
- All four design-system ceiling ratchets, the untested-services floor and the test-structure floor — green
- Manually verified on a real Samsung Galaxy: video plays in the preview, timeline scrubs smoothly, keyboard pushes the sheet over the picture without resizing the video, sheet drags between its bounds, cue text and timing edits reach the cubit

## Takeover verification

- `flutter test test/widgets/subtitle_editor/subtitle_timeline_thumbnails_test.dart test/widgets/subtitle_editor/subtitle_editor_stage_test.dart test/services/subtitle_timeline_thumbnail_service_test.dart test/widgets/captions/caption_cue_row_test.dart test/screens/subtitle_editor/subtitle_editor_screen_test.dart`
- `flutter test test/widgets/video_editor/timeline_editor/controls/video_editor_captions_sheet_test.dart test/widgets/captions/caption_cue_row_test.dart`
- Pre-push hook: merge-conflict check, analyzer, and changed-file tests — all passed

## Screenshots

Screen recording to follow.


NotThatKindOfDrLiz:
@hm21 I took over the review fixes and pushed two commits on this branch: `89ff917ed` and `d8386c13b`.

What changed:
- Fixed filmstrip frame timing so thumbnails map to the actual visible timeline axis instead of the rounded raster width.
- Added stale seek-report filtering so scrub and row-focus seeks do not snap the playhead back to old native player positions.
- Resolved playable/cacheable video URLs before timeline thumbnail extraction and avoided poisoning the shared `video.id` media cache with HLS manifests.
- Added lifecycle cleanup for extracted timeline frame files, with logged cleanup failures.
- Restored keyboard avoidance for the no-stage editor path, made the sheet initial size explicit, fixed stale comments, and made row body taps select the cue.
- Added regression coverage for the thumbnail timing, stale-report guard, cache URL resolution, frame cleanup, shared caption row behavior, no-stage keyboard inset, and row-tap selection.

Verification:
- `flutter test test/widgets/subtitle_editor/subtitle_timeline_thumbnails_test.dart test/widgets/subtitle_editor/subtitle_editor_stage_test.dart test/services/subtitle_timeline_thumbnail_service_test.dart test/widgets/captions/caption_cue_row_test.dart test/screens/subtitle_editor/subtitle_editor_screen_test.dart`
- `flutter analyze` on the touched files
- `flutter test test/widgets/video_editor/timeline_editor/controls/video_editor_captions_sheet_test.dart test/widgets/captions/caption_cue_row_test.dart`
- `bash scripts/check_empty_catch_ceiling.sh`
- GitHub CI is green, including Mobile CI, generated-files, test shards, preview build, semantic PR, CLA, and mergeability.

Please review the takeover commits and merge when you are ready.

NotThatKindOfDrLiz (approved):
I re-reviewed the final diff after the takeover commits. The blocking review findings are addressed: thumbnail timing now uses the actual timeline axis, stale seek reports are filtered, timeline thumbnail downloads resolve playable/cacheable URLs and skip unresolved HLS, extracted frame files are cleaned up, and the no-stage keyboard path is restored. The smaller follow-ups around sheet sizing, stale comments, row body selection, and missing tests are covered too.

Local targeted verification passed, broader `flutter analyze lib test integration_test` passed, and GitHub CI is green. Approved.

NotThatKindOfDrLiz (changes requested):
## PR Review: feat(subtitles) — Time Captions Against Video Instead of Typed Seconds

**Summary:** This PR adds video preview and timeline-based frame extraction to the subtitle editor, enabling visual feedback during caption editing. The implementation is mostly sound but has **4 blocking issues** that break functionality or violate project invariants, plus 3 cleanup items.

---

## Blocking Issues (Must Fix Before Merge)

### 1. **Frame-to-Time Mapping Formula Is Incorrect** 🔴
**File:** `mobile/lib/widgets/subtitle_editor/subtitle_timeline_thumbnails.dart`, `_frameForSlot()` method

**Problem:**
```dart
final timeMs = (slot + 0.5) / slotCount * timelineMs;
```

This uses a uniform distribution (`(slot + 0.5) / slotCount`), which is **wrong** for video scrubbing. The video editor's clip strip uses:
```dart
final timeMs = slotIndex * (pixelsPerSecond / 1000) * pixelWidth;
```

The two formulas produce **different frame sequences** when the same video is previewed side-by-side in the subtitle editor and video editor. Users will see mismatched timelines.

**Impact:** Visual consistency and scrubbing fidelity.

**Fix:** Align the slot-to-time formula with the video editor's density calculation, or extract to a shared constant.

---

### 2. **Seek-in-Flight Race Condition** 🔴
**File:** `mobile/lib/widgets/subtitle_editor/subtitle_editor_stage.dart`, `_onPlayheadChanged()`

**Problem:**
The playhead can emit seek events faster than the player finishes seeking. If the user scrubs rapidly:
1. User seeks to 5s → player starts seek
2. User seeks to 6s → emits new seek **before** 5s completes
3. Player reports position at 5s
4. Timeline UI now shows 5s but playback is actually at 6s → desync

The video editor guards this with `shouldAcceptPlayerReport()`. This PR's implementation lacks that guard.

**Impact:** Playhead position diverges from actual playback during rapid scrubbing.

**Fix:** Add a `_seekInFlight` flag (like `video_editor_canvas.dart`):
```dart
if (_seekInFlight) return; // Ignore stale player reports
_seekInFlight = true;
_videoController.seekTo(Duration(...));
// Set _seekInFlight = false in player's completion callback
```

---

### 3. **Cache Key Poisoning from Raw Video URL** 🔴
**File:** `mobile/lib/providers/subtitle_timeline_thumbnail_provider.dart`

**Problem:**
```dart
thumbnailService: SubtitleTimelineThumbnailService(
  mediaCache: mediaCache,
  videoUrl: videoUrl,  // ← Raw URL, not resolution-aware
),
```

The raw `videoUrl` is passed directly to `SubtitleTimelineThumbnailService`, which uses it as a cache key. This:
1. Bypasses platform-specific URL resolution (see `getCacheableVideoUrlForPlatform()`)
2. Poisons the media cache with inconsistent keys if the same video is accessed via different URLs (HTTP redirects, domain aliases, etc.)
3. Violates the cache contract established in `media_cache_package`

**Impact:** Cache misses and disk duplication.

**Fix:** Call `getCacheableVideoUrlForPlatform(videoUrl)` before passing to the service.

---

### 4. **Frame Extraction Disk Leak — Extracted JPEGs Are Never Cleaned** 🔴
**File:** `mobile/lib/services/subtitle_timeline_thumbnail_service.dart`

**Problem:**
```dart
Future<TimelineFrame> extractFrame(int frameIndex, ...) async {
  final ffmpegOutput = '${tempDir.path}/frame_$frameIndex.jpg';
  await FFmpegKit.execute(...output: ffmpegOutput);
  return TimelineFrame(imagePath: ffmpegOutput, timestamp: ...);
}
```

The extracted JPEGs at `ffmpegOutput` are created but **never deleted**. Every subtitle editor session that extracts N frames leaves N orphaned files in `tempDir`.

**Impact:** Unbounded disk growth over app lifetime.

**Fix:** Implement cleanup:
```dart
@override
Future<void> dispose() async {
  // Delete extracted frames from tempDir
  final pattern = RegExp(r'frame_\d+\.jpg$');
  for (final file in tempDir.listSync()) {
    if (file is io.File && pattern.hasMatch(file.path)) {
      await file.delete();
    }
  }
}
```

Call `dispose()` in the provider's `onDispose` callback.

---

## Important Issues (Should Fix)

### 5. **Keyboard Avoidance Regression on Sheet** 🟡
**File:** `mobile/lib/widgets/subtitle_editor/subtitle_editor_screen.dart`

**Problem:**
```dart
Scaffold(resizeToAvoidBottomInset: false, ...)
DraggableScrollableSheet(
  builder: (...) => Padding(
    padding: EdgeInsets.only(bottom: viewInsets.bottom),
    ...
  )
)
```

The Scaffold disables keyboard avoidance globally, then only the sheet applies inset compensation. The **video stage sits below the sheet** and receives **no inset padding** — the keyboard will overlap the video player controls if opened from the sheet.

**Impact:** Hidden player buttons when keyboard is visible.

**Fix:** Apply inset padding to the video stage as well, or wrap both stage + sheet in a single keyboard-aware container.

---

### 6. **Stale Comments Indicate Design Drift** 🟡
**Files:**
- `subtitle_editor_screen.dart`: `// TODO: Handle navigation and loading states` — not addressed in this PR
- `subtitle_editor_stage.dart`: `// TODO: Add more keyboard controls` — incomplete feature

These suggest design decisions were punted; they should either be removed or tracked in an issue.

**Fix:** File an issue for each, remove TODOs, or complete the feature.

---

## Cleanup (Minor — No Functional Impact)

### 7. **Duplicate Helper Logic in CaptionCueRow** 🟢
**File:** `mobile/lib/widgets/captions/caption_cue_row.dart`

The `_InputSurface` helper extracted from video editor is duplicated here. If the design system component changes, both copies need updates. Consider adding this to `divine_ui` as a reusable `CaptionInputField`.

---

## Verification Checklist

- [x] No new tech debt (**mostly met** — TODOs are pre-existing; disk leak is new)
- [x] Tests cover new state + UI paths
- [x] Follows `flutter_bloc` + repository pattern
- [x] No hardcoded values or magic numbers
- [x] Localization keys added to ARB
- [x] Golden files updated
- [x] Pre-push checks pass locally

---

## Recommendation

**Send back to author with findings 1–5 as blocking.** Findings 6–7 are cleanup and can be tracked separately. The visual timeline concept is solid; the fixes are straightforward layering / guard clause additions that the author is well-positioned to complete.

---

## Reference Implementation Pointers

- **Frame formula:** `mobile/lib/widgets/video_editor/timeline_editor/clip_strip.dart`, `slotForTime()` / `timeForSlot()`
- **Seek guard:** `mobile/lib/widgets/video_editor/canvas/video_editor_canvas.dart`, `_onPlayerPositionChanged()` → `shouldAcceptPlayerReport()`
- **Cache key resolution:** `mobile/lib/models/video.dart`, `getCacheableVideoUrlForPlatform()`
- **Temp file cleanup patterns:** `mobile/lib/services/subtitle_timeline_thumbnail_service.dart` (apply to this service's own extraction)

PR #5328, opened 2026-06-19
fix(db): force SQLCipher runtime on iOS

rabble:
Closes #5329

## Summary
- force iOS/macOS sqlite3 loading to resolve SQLCipher from the app process and force-link SQLCipher in Runner configs
- remove stale CSQLite SwiftPM pins so plain SQLite cannot win the runtime loader path
- keep automatic encrypted-cache repair for allowlisted SQLite cache corruption / `SQLITE_NOTADB` failures only
- keep SQLCipher-link and secure-storage bootstrap failures fail-closed so transient keychain errors cannot wipe or rotate the DB cipher key
- deduplicate post-reset DM sync cleanup and avoid recording the same bootstrap failure twice during the one retry
- keep the repair classifier web-safe so preview builds do not import native sqlite FFI declarations
- show a release-safe database bootstrap diagnostic code and add CI guard tests for iOS SQLCipher linkage regressions

## Root cause
Crashlytics for `1.0.15+751` showed `SQLCipher is not linked`, not a stale or missing DB key. Wiping the local DB cannot fix that class of failure. iOS had the SQLCipher pod present, but the runtime could still resolve plain SQLite through stale `simolus3/CSQLite` SwiftPM pins / sqlite3.framework / other dependencies. The existing runtime probe correctly failed closed instead of opening an unencrypted database.

The follow-up commit also fixes the review-blocking repair policy bug: the first version repaired on every bootstrap error except SQLCipher-unavailable, which could destructively back up/remove the local DB after a transient secure-storage read failure. Repair is now an allowlist, and secure-storage failures explicitly do not repair.

## Verification
- `flutter analyze lib/database/sqlcipher_runtime_io.dart lib/startup/database_bootstrap_failure_app.dart lib/services/database_encryption_bootstrap.dart lib/main.dart test/startup/database_bootstrap_failure_app_test.dart test/services/database_encryption_bootstrap_test.dart test/ios/ios_sqlcipher_linkage_test.dart`
- `flutter test test/startup/database_bootstrap_failure_app_test.dart test/services/database_encryption_bootstrap_test.dart test/ios/ios_sqlcipher_linkage_test.dart`
- `flutter test test/startup/database_bootstrap_failure_app_test.dart`
[...]

PR #7065, opened 2026-08-11
build(ios): refresh the SPM lockfiles for patrol's Swift Package

### LINKED ISSUE #7067: build(ios): refresh the SPM lockfiles for patrol's Swift Package
Auto-created from PR #7065.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7065
**Branch:** `build/ios-spm-patrol-pin`

hm21:
Closes #7067

## Description

Both tracked iOS SwiftPM lockfiles show up as modified after every local build, and the churn has already leaked into unrelated PRs (two commits on `feat/video-card-post-date`). They are not machine-specific artifacts — the checked-in files are simply stale.

The timeline:

- Both lockfiles were last written on 2026-06-19 (`3b2005ebfa`, the SQLCipher fix), and they are the same blob on `main` (`92baa9372`).
- #6925 moved patrol from 4.2.0 to 4.8.0 on 2026-08-09.
- patrol 4.2.0 shipped as a CocoaPods pod. 4.8.0 is the first version with a `Package.swift`, and it depends on CocoaAsyncSocket 7.6.5.

Nobody has committed a resolve since that bump, so Xcode re-adds the pin on every build and the files have been coming back dirty for two months.

Both files are updated because Xcode writes them from two separate resolves: flutter's prefetch runs `xcodebuild -resolvePackageDependencies` with no `-project` and no `-workspace` (`xcodeproj.dart:208-227`), which resolves `Runner.xcodeproj` and writes the inner file, while the build itself passes `-workspace Runner.xcworkspace` and writes the outer one. The repo already treats them as a pair — `ios/.gitignore` keeps the inner one tracked "alongside the outer workspace lock", `sqlite3mc_configuration_test.dart` asserts over both, #5328 dropped the `csqlite` pin from both symmetrically, and `codemagic.yaml:306` reads the outer one for the Crashlytics symbol upload.

The revision (`dbdc00669c1ced63b27c3c5f052ee4d28f10150c`, 7.6.5) is exactly the one patrol pins in its own `darwin/patrol/Package.resolved`, so this adds no independent version decision.

**Related Issue:** 

## Out of Scope

An earlier revision of this description claimed release builds would prune the pin again, because Flutter filters dev-dependency plugins out in release mode. That filter does not apply to Darwin: `injectPlugins` re-resolves from the unfiltered plugin list under `if (iosPlatform || macOSPlatform)`, commented "iOS and macOS doesn't yet support filtering out dev dependencies" (flutter/flutter#163874). iOS never filters patrol out, in any mode, so the pin is stable and a release `flutter build ipa` will not remove it.

The flip side is that patrol — and CocoaAsyncSocket with it — is linked into release iOS builds today. That arrived with #6925 and is out of scope here; this PR only records a resolve that already happens.

## Verification

- `flutter test test/database/sqlite3mc_configuration_test.dart` — 4/4 pass (this is the test that reads both lockfiles)
- Both files are valid JSON with 25 pins, and byte-identical to each other, as they are on `main`
- Diff is additive: one pin object per file, no existing entry touched or reordered
- Pinned revision compared against `~/.pub-cache/hosted/pub.dev/patrol-4.8.0/darwin/patrol/Package.resolved` — identical

## Type of Change

### DIFF
--- a/mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -28,6 +28,15 @@
       }
     },
     {
+      "identity" : "cocoaasyncsocket",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/robbiehanson/CocoaAsyncSocket",
+      "state" : {
+        "revision" : "dbdc00669c1ced63b27c3c5f052ee4d28f10150c",
+        "version" : "7.6.5"
+      }
+    },
+    {
       "identity" : "dkcamera",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/zhangao0086/DKCamera",
--- a/mobile/ios/Runner.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/mobile/ios/Runner.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -28,6 +28,15 @@
       }
     },
     {
+      "identity" : "cocoaasyncsocket",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/robbiehanson/CocoaAsyncSocket",
+      "state" : {
+        "revision" : "dbdc00669c1ced63b27c3c5f052ee4d28f10150c",
+        "version" : "7.6.5"
+      }
+    },
+    {
       "identity" : "dkcamera",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/zhangao0086/DKCamera",


mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -27,6 +27,15 @@
         "version" : "0.0.9-beta.7"
       }
     },
+    {
```
realmeylisdev, line 30: this pins the inner lockfile, but the outer one is tracked too and still churns.

one `flutter run` / `flutter build ios` writes both, from two different xcodebuild calls:

- flutter's prefetch runs `xcodebuild -resolvePackageDependencies` from `mobile/ios` with no `-project` and no `-workspace`, so it resolves `Runner.xcodeproj` and writes this file
- the build itself runs `-workspace Runner.xcworkspace`, which writes `Runner.xcworkspace/xcshareddata/swiftpm/Package.resolved`

checked it by deleting each file and running only the matching resolve: each one comes back, the other stays untouched. on main both files are the same blob (`92baa9372`), so once this merges the outer one still shows up modified after every debug build, and the two now differ.

the repo already treats them as a pair:

- `mobile/ios/.gitignore:30` says to keep this one tracked "alongside the outer workspace lock"
- `mobile/test/database/sqlite3mc_configuration_test.dart` runs its CSQLite assertion over both
- 3b2005ebf (#5328) dropped the `csqlite` pin from both, symmetrically, at this exact spot
- `codemagic.yaml:306` reads the outer file for the crashlytics symbol upload

they're identical on main so a straight copy does it:

```
cp mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved \
   mobile/ios/Runner.xcworkspace/xcshareddata/swiftpm/Package.resolved
```

happy to push that for you if you'd rather not.

mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -27,6 +27,15 @@
...
+      "identity" : "cocoaasyncsocket",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/robbiehanson/CocoaAsyncSocket",
+      "state" : {
+        "revision" : "dbdc00669c1ced63b27c3c5f052ee4d28f10150c",
+        "version" : "7.6.5"
```
realmeylisdev, line 36: about the description rather than this line.

the dev-dependency filter you quote is real, but it only feeds android/linux/windows. darwin is explicitly exempt in 3.44.0's `flutter_plugins.dart`:

```dart
// iOS and macOS doesn't yet support filtering out dev dependencies.
// See https://github.com/flutter/flutter/issues/163874.
```

that's 1344-1346, and 1359-1361 hands `DarwinDependencyManagement` the unfiltered `plugins` list. no `isDevDependency` hits anywhere under `lib/src/macos`.

so ios never filters patrol out, in any mode. mostly good news: it answers your out-of-scope note, since a release `flutter build ipa` keeps patrol in the generated package and won't prune the pin. flip side is that "CocoaAsyncSocket is not in the release dependency graph" doesn't hold on ios. that came in with #6925 rather than here, but might deserve its own look.

realmeylisdev (changes requested):
pin itself is right, matches patrol 4.8.0's own `Package.resolved` byte for byte. but it only lands in one of the two tracked lockfiles, so `git status` still comes back dirty after this.

- `Runner.xcworkspace/xcshareddata/swiftpm/Package.resolved` needs the same 9 lines. it's the same blob as this one on main, and one `flutter run` rewrites both
- the release-mode reasoning in the description is backwards for ios. doesn't break the change, and it actually settles your out-of-scope question

rest is inline.

mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
hm21: Confirmed and fixed in 3077c0d25 — same 9 lines in the outer lockfile, both files byte-identical again.

Verified the pairing myself before copying: `git ls-files` shows both tracked, both are blob `92baa9372` on main, `3b2005ebfa` changed both symmetrically, and `sqlite3mc_configuration_test.dart:52-73` iterates both (4/4 pass with the change). Also confirmed the resolve split — `_xcodebuildProjectCommandArguments` builds `xcrun xcodebuild -clonedSourcePackagesDirPath <cache>` with no `-project` and no `-workspace` (`xcodeproj.dart:208-227`).

One nuance: on my machine the outer file was still clean while the inner one had been dirty for days, so `flutter run` alone had not rewritten both here. Does not change anything about the fix.

realmeylisdev (approved):
both lockfiles carry the pin now and they're back to the same bytes. that blob is exactly what xcode writes on my machine too, so the dirty-file thing is genuinely gone rather than halved. nice catch rewriting the out-of-scope note as well, and calling out that the release linkage came from #6925. checks are green 👍

mobile/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
hm21: You are right, and I checked it in the SDK this repo pins rather than taking it on trust: `injectPlugins` filters dev dependencies for android/linux/windows, then the `if (iosPlatform || macOSPlatform)` branch re-resolves from the unfiltered `plugins` list with exactly that comment, and hands `DarwinDependencyManagement` the unfiltered list too. So iOS never filters patrol out, in any mode.

Description rewritten: the release-mode causality is gone, the out-of-scope caveat is resolved (a release `flutter build ipa` will not prune the pin), and the "not in the release dependency graph" claim is removed since it is false on iOS.

patrol being linked into release iOS builds is a real thing worth its own look, but it came in with #6925 and I am not folding it into this PR.

hm21:
Both findings were right — verified each one against the pinned SDK source and the repo before acting, details in the threads.

- **Outer lockfile** — fixed. `3077c0d25` carries the same pin in both files, byte-identical again.
- **Release-mode reasoning** — fixed. Darwin is exempt from the dev-dependency filter, so the pin is stable in every mode; description rewritten and the out-of-scope caveat resolved by your finding.
- **patrol in release iOS builds** — real, but it arrived with #6925. Not folded in here.

Checks are green. `Mobile CI` went red once on a timing budget (Generated Files 431s vs 420s); the job itself passed and the previous commit on this same PR ran it in 218s, with recent `main` runs at 192-235s, so I re-ran it — 200s, green.

Thanks @realmeylisdev, the second one was the useful kind of catch: it did not change the diff but it settled the open question in it.

PR #7069, opened 2026-08-11
feat(video-editor): crop, rotate and flip a stop-motion frame

### LINKED ISSUE #7070: feat(video-editor): crop, rotate and flip a stop-motion frame
Auto-created from PR #7069.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7069
**Branch:** `feat/stop-motion-frame-transform`

hm21:
Closes #7070

## Description

A stop-motion still could be deleted, duplicated, reordered and held longer, but never reframed — while a video clip in the same editor could be. Selecting a still now offers **Transform**, which opens it in `pro_image_editor`'s `CropRotateEditor`.

**Why the still does not reuse `VideoClipTransformScreen`.** That screen drives `CropRotateEditor.video` off a `ProVideoController`, and a frames-only stop-motion clip has no mp4 to give it — the file only exists after `materialize()` at publish, so the existing screen bails immediately on `requireVideo.file?.path == null`. The results also differ in kind: a clip transform hands back an `ExportTransform` for the bloc to bake into a re-rendered video, whereas the image editor rasterizes the still itself, so the frame screen returns finished pixels and only has to write them to a file. What both screens *show* is now one shared `transformEditorConfigs`, so the two can no longer drift apart.

**Why the crop is locked to the composition ratio.** The stop-motion encoder cover-crops every still into exactly the target frame (`StopMotionFit.cover`), so a crop to any other ratio would simply be cropped again on render.

**Why the result is always a new file.** Editor undo history keeps pointing at the previous path, and duplicated stills share one file — rewriting in place would silently transform every copy and make undo a no-op. Only the frame's `path` changes; the hold and its override flag belong to the slot in the sequence, not to the pixels.

Two bugs found while testing on device are fixed in the refactor commit, both caused by the crop-rotate editor inheriting package defaults:

- **Yellow processing text.** The editor mounts its loading dialog in an `OverlayEntry`, which sits outside every route's `Material`, so text there inherited `MaterialApp`'s unstyled-text fallback — red monospace under a yellow double underline. The shared overlay now carries its own `Material`.
- **"Oops, something went wrong" on close.** Both editors defaulted to `kImageEditorHeroTag`, so pushing or popping a transform route flew a hero between them. A hero flight re-parents the main editor's content into the navigator overlay, above the route owning `BlocProvider<ClipEditorBloc>` — and `VideoEditorClipPreview` reads that bloc *only* for a stop-motion clip, so it threw `ProviderNotFoundException` mid-build and the global error widget blanked the screen. That is also why the existing clip transform never showed it. The transform editors now use their own hero tag; the routes already fade rather than hero-animate.

**Related Issue:** 

## Out of Scope

- Frame multi-select still offers hold / reverse / delete only. Applying one transform across N stills needs a batch path and is not attempted here.
- The clip thumbnail is not refreshed when the first still is transformed, matching the existing behaviour of every other frame op (delete, reorder, hold).

## Found reviewing the branch

Three fixes on top of the feature, all caught reading the diff rather than on device:

- **A failed rasterization was silent.** `CropRotateEditor.done` retries its screenshot pass and, when all attempts fail, hands the callback `Uint8List(0)` and closes anyway. Those empty bytes were dropped alongside a Cancel — the user waited through the loading overlay, tapped Done, and got an unchanged still with no message and nothing logged. They now reach the bloc, which raises the same failure snackbar a bad write does without writing a frame that would point at a broken image.
- **The stale-index guard did not guard.** It re-resolved the clip and re-checked the frame bounds, but a delete or reorder landing during the write leaves `frameIndex` naming a different still — and the crop would have been applied to that one. The still's path is captured before the write and compared after it; a slot that moved abandons the result.
- **Errors were not classified.** The `catch` passed everything to `addError` raw, so a broken invariant in the write path never reached Crashlytics. It now matches `_onClipTransformRequested` directly below it: `StateError` / `TypeError` / `RangeError` wrapped as `Reportable`, IO failures left out.

Plus two notes that were wrong: the `.jpg` filename rested on an undeclared `pro_image_editor` default (the configs test pins it now), and the write doc claimed an orphan sweep that does not exist — the superseded still leaks, the clip transform leaks its input the same way, and #7077 tracks both.

## Verification

- `flutter test test/l10n/arb_consistency_test.dart test/widgets/video_editor test/screens/video_editor test/services/video_editor test/models/stop_motion` — 1295 passing, 3 skipped
- Ratchets: `untested_services_floor`, `raw_textstyle_ceiling`, `raw_colors_ceiling`, `material_button_ceiling`, `raw_dialog_ceiling`, `test_unit_structure` — all OK
- New tests: `StopMotionFrameOps.setFramePath`; the frame-transform service's write / new-path behaviour; the shared chrome's rotate / flip / reset / cancel / apply wiring; the loading overlay's resistance to the unstyled-text fallback; the hero tag being distinct from the main editor's; the Transform button's presence, absence without a selection, and frame-specific semantic label; the bloc rejecting bytes the editor failed to rasterize and abandoning a still whose slot moved during the write; the editor's output format staying JPEG
- `StopMotionFrameTransformScreen` itself carries no direct test. The one written for it could only be kept green by letting real wall-clock time pass inside `runAsync` — `pro_image_editor`'s fit-to-screen loop drives itself off `DateTime.now()` while awaiting `Future.delayed`, so pumping fake time never ends it and the binding reports a pending timer at teardown. That is the wall-clock wait the `future_delayed_ceiling` ratchet exists to freeze, so it was dropped rather than reworded past the grep. What it covered is not lost: the shared chrome, the loading overlay and the editor configuration keep their own tests, and the frame swap now lives in `ClipEditorBloc`, where it is covered without an editor at all.
- l10n: three new keys translated into all 21 non-English locales
- Manually exercised on a real Samsung Galaxy — that device run is how the yellow processing text and the `ProviderNotFoundException` blank screen were found. **The two fixes for them landed after that run and still want a final pass on device, which is why this is a draft.**

## Type of Change

- [x] 🧹 Code refactor


rabble (approved):
Approved at acd8ae00b after takeover verification. I resolved the main conflict by rebasing the contributor commits and preserving the newer crop-corner accent in the shared transformEditorConfigs path, so both the video and stop-motion editors inherit it. After the fresh rebase: 260 focused transform/editor tests passed, the full analyzer was clean, the pre-push hook passed 290 changed-file tests, and every GitHub check is green. GitHub has since marked the branch behind another moving-main update, but it remains conflict-free and mergeable; there is no review blocker in this diff.

realmeylisdev (approved):
lgtm. drove the frame screen in a harness with a real jpeg at acd8ae0: Done comes back with real JPEG bytes, Cancel comes back null, and the loading overlay is up while it rasterizes.

also reproduced the hero thing rather than take it on trust. with the same tag on both routes the pop throws "Could not find the correct Provider", and with your tag the transform screen pushes and cancels clean while the main editor keeps its bloc. so the on-device pass you were still waiting on is covered from my side.

nit, the description still says this is a draft.

Issue #4124, opened 2026-05-08
DM: auto-sweep recipient-sent / self-wrap-failed rows in outgoing_dms


realmeylisdev:
## Summary

`outgoing_dms` queue rows can persist in `recipient: sent / self: failed` state when the user dismisses the partial-delivery SnackBar without tapping Retry, kills the app before retry, or both bookkeeping writes fail in `recoverSelfWrap` after a successful publish. Today these rows only progress when the user manually re-taps Retry — there is no automatic sweep.

## Problem

The user-facing UX assumes Retry is the only recovery surface, but the queue row is the durable source of truth and outlives the SnackBar. Without a sweep:

- A self-wrap publish that fails silently (relay outage, transient network) stays unfixed across app launches.
- Sender cross-device sync — the entire purpose of the self-wrap — is silently broken until the user reopens that conversation, sees no SnackBar, and never knows to retry.

## Desired Outcome

Periodically sweep `outgoing_dms` for rows where `recipient_wrap_status == sent && self_wrap_status == failed` and call `DmRepository.recoverSelfWrap(rumorId)` for each.

## Scope

- Decide cadence: app foreground, conversation enter, or a periodic timer. App-foreground is the cheapest signal that maps to "user is around if a relay round-trip is needed."
- Bound the work: rate-limit attempts per row (e.g., min N seconds between retries), cap total attempts before backing off (the row already records `lastError`).
- Verify the sweep cooperates with `recoverSelfWrap`'s idempotent `selfWrapStatus == sent` guard so a manual Retry mid-sweep doesn't republish.

## Context

- Recovery primitive landed in #4106 (closes #4102); what's missing is automatic invocation.
- Called out during PR #4106 review as a follow-up gate before merge.

Issue #4977, opened 2026-06-08
fix(dm): DM read state lost after reinstall — all conversations show unread (no cross-install read sync)


realmeylisdev:
### Summary

After uninstall → reinstall → re-login, **every** recovered DM conversation shows as unread — including ones the user had already read on the previous install. Reported by @hm21 while testing #4973.

### Root cause (verified in code)

DM read state is **local-only**. A conversation's `isRead` is set to `isSentByMe` when its latest message is processed (`dm_repository.dart:825` and `:991`) and is only flipped to `true` by the local `markConversationAsRead` → `ConversationsDao.markAsRead` write (`conversations_dao.dart:258`). There is **no relay-side read marker / read receipt** published or read anywhere (grep for read-receipt / read-marker / publishRead returns nothing in `dm_repository` or the DM blocs).

On reinstall the local Drift DB is wiped, so all read markers are gone. The #4953 recovery drain (#4973) then re-materializes the full conversation list from relays with `isRead = isSentByMe`, i.e. every conversation whose most-recent message was *received* shows unread. Given the current model this is technically correct — on this fresh install there is no record that the messages were read elsewhere.

This is **not a regression** introduced by #4973; the drain just makes far more (correctly-unread-on-this-install) conversations visible than the pre-fix ~4.

### Expected (product decision needed)

Read state should ideally survive reinstall and sync across the user's devices.

### Fix direction (needs design)

Requires a cross-device read-state mechanism over Nostr — e.g. publishing per-conversation "last read" pointers to the user's own relays and reconciling them on login. This carries protocol/privacy choices (which kind; encrypted payload vs plaintext pointers; per-conversation vs a single rolling marker; interaction with NIP-17) and is a real feature, not a quick fix.

Flagging for design (@NotThatKindOfDrLiz).

### Context

Surfaced by @hm21 testing #4973 (closes #4953). Pairs with the badge-undercount follow-up (separate issue). Related receive-side completeness work: #4974.

NotThatKindOfDrLiz:
Flagging for @Chardot as this needs design input

realmeylisdev:
Dug into this. Root cause checks out — read state is local-only and the reinstall drain re-marks everything unread. Same thing can happen on an app update too, not just reinstall.

The catch: the "I read this" flag got wiped with the DB, so there's no way to bring it back locally after a reinstall. So it's a product call:

[...]

PR #7075, opened 2026-08-11
fix(dm): re-derive the DM publish backstop and bound the crypto chain

### LINKED ISSUE #6586: fix(dm): the 90s publish backstop is derived from a Keycast timeout that has since changed
## Problem

`dm_repository`'s 90s message-publish backstop documents its derivation from a Keycast RPC bound of **12s**. That constant is now **30s**, so the arithmetic in the comment no longer holds.

`mobile/packages/dm_repository/lib/src/dm_repository.dart:160-169` still reads:

> Keycast RPCs 12s each (`KeycastRpc.defaultRequestTimeout`) ...

The actual bound is `KeycastRpc` at 30s per round trip (`mobile/packages/keycast_flutter/lib/src/rpc/keycast_rpc.dart`).

Recomputing the documented worst case with the real value takes it from ~68s to ~140s — past the 90s backstop. The backstop can therefore fire mid-send, which is the exact failure it was raised to prevent.

## Suggested direction

Re-derive the backstop against the current signer bounds, or key it off the constant rather than a hard-coded number so it cannot drift again.

Found while investigating #3366.

realmeylisdev:
Closes #6586.

## Description

`dm_repository`'s 90s publish backstop documented its own derivation as "Keycast RPCs 12s each". #6046 computed 90s from that bound; #6075 raised the bound to 30s — correctly, since 12s had regressed video-publish signing, likes, reposts and follows — but never re-derived the backstop.

The stale comment turned out to be the smallest part of it.

**A 1:1 send costs four signer round trips, measured against the real send path** — `nip44Encrypt` + `signEvent` for the seal, once per wrap; the kind-1059 wrap is ephemeral-signed and never reaches the signer. So the honest chain was `(2 × 30s) + 10s + (2 × 30s) + 10s = 140s` against a 90s cap.

**The cap therefore fired on successful sends.** The self-wrap is only built *after* the recipient's `OK true` arrives, so that 140s path is the delivery path: the recipient had the message, and the sender got a stuck "sending" bubble, five duplicate republishes, then a red "failed" bubble. keycast#291 records the same failure from production mobile logs — `nip44_encrypt` at 23s and 30s, and *"a kind-14 DM send hit the mobile 30s timeout and did not deliver."*

**Amber/NIP-55 is human-gated in the way that matters here.** `AndroidNostrSigner`'s 300s timeout only bounds waiting for its serialized signer lock; the `nip44Encrypt` and `signEvent` intent calls can wait unbounded for the user's approval. That makes the app-layer chain bound the only pre-publish guard for a user who backgrounds into Amber approvals.

## Approach

Correcting the number alone would not have fixed the Amber arm, and no signer interface exposes a per-call timeout. So the fix bounds the **send chain** rather than the transport for queued durable sends, while preserving the previous uncapped behavior for `sendPrivateMessage` callers that have no retry queue.

When the recipient wrap build times out before any publish, durable DM sends now classify it as `retryablePending: true`: nothing reached a relay, so the row should stay pending and let the sweep re-drive instead of turning a slow human approval into a red bubble. A builder that actually returns `null` remains a hard failure.

The key behavioural change is *where* a slow signer fails:

| Scenario | Before | After |
|---|---|---|
| Slow queued recipient build | publishes, delivers, dies at 90s → stuck bubble, 5 duplicate republishes, red bubble on a delivered message | times out at the build bound **before any publish**; durable sends stay retryable-pending and are re-driven |
| Slow uncapped recipient build (`sendPrivateMessage`) | waits for the signer approval | still waits; no retryable-pending result with no queue row to retry from |
| Slow self-wrap build | eats the remaining budget, trips the backstop, misclassifies a **delivered** message | returns `selfWrapPublished: false` → existing `recoverSelfWrap` finishes it out of band (#4124); uncapped callers get the wider self-wrap build bound |
| Amber/NIP-55 approval | backstop is the only bound, and user approval can outlive it | queued recipient build timeout is pre-publish and retryable-pending; uncapped convenience sends preserve their prior wait behavior |

## Commits

1. **`fix(dm): derive the publish backstop from its sub-bounds`** — moves every sub-bound into `DmSendBudget` and computes totals from shared constants.
2. **`fix(dm): bound the gift-wrap builds so the backstop cannot be outrun`** — adds recipient/self wrap build bounds so the send backstop cannot fire after delivery.
3. **`fix(dm): derive the interrupted-send guard from the publish backstop`** — derives `_interruptedMinAge` from the real send budget.
4. **`fix(dm): bound self-wrap recovery on its own budget, not the send's`** — keeps out-of-band recovery off the tight in-send self-wrap bound.
5. **`fix(dm): give the wrap-build bound margin over the two transport ops`** — gives recipient-wrap build local crypto margin over `2 × 30s` and scopes the budget docs.
6. **`fix(dm): don't apply the send-sized self-wrap bound to uncapped callers`** — widens self-wrap build for `sendPrivateMessage` callers with no outer cap.
7. **`docs(dm): align self-wrap budget comments`** — fixes the self-wrap naming/docs/accounting from review.
8. **`fix(dm): keep Amber wrap timeouts retryable`** — keeps queued Amber recipient-build timeouts pending/retryable and bounds `publishSelfApplicationMarker` (#7093, now closed).
9. **`fix(dm): keep uncapped recipient wrap builds retryable`** — preserves uncapped recipient-wrap construction for `sendPrivateMessage` callers that have no retry row.

## Testing

- Recipient-build timeout → retryable-pending failure **and `publishEventAwaitOk` is never called**, proving nothing was delivered.
- `sendPrivateMessage` with a recipient build slower than `recipientWrapBuild` → succeeds instead of returning retryable-pending with no queue row.
- Recipient builder returning `null` → hard failure, preserving the true build-failure classification.
- Self-wrap-build timeout → `success == true` with `selfWrapPublished == false`, proving the partial-delivery path rather than a send failure.
- Self-application-marker wrap build timeout → returns `false` without publishing, covering the adjacent unbounded call.
- Cross-package budget invariants in the app layer (`dm_repository` cannot import `keycast_flutter`, so this is the only place the full relationship can be asserted).
- Existing tests that hardcoded the 120s boundary now derive it from the guard, so a moving budget cannot silently stop testing the boundary.

Local verification after the final takeover commit:

- `flutter test test/src/nip17_message_service_test.dart` from `mobile/packages/dm_repository`
- `flutter analyze` from `mobile/packages/dm_repository`
- `flutter test` from `mobile/packages/dm_repository`
- Pre-push hook analyze and changed-file tests, including `packages/dm_repository/test/src/nip17_message_service_test.dart` and `test/services/outgoing_dm_retry_service_test.dart`

## Out of scope — tracked separately

1. #7090 — adopt `nip17_wrap_batch` (4 signer round trips → 1) once the production Keycast verb is confirmed.
2. #7091 — bound the send segments outside the publish backstop: kind-10050 inbox resolution, the connectivity probe, and `retryDisconnectedRelays()`'s missing aggregate cap.
3. #7092 — reconcile the client's 30s per-op bound with Keycast's new 8s server bound once deployment behavior is confirmed.

## Notes for reviewers

No UI change, so no screenshots. The trade-off worth a second opinion is the backstop moving 90s → 120s: the ceiling rises, but the common queued slow path now fails at 65s before delivery and remains retryable for durable DM sends, so typical time-to-feedback drops without turning human-gated Amber approval into a red bubble.

### DIFF
--- a/mobile/packages/dm_repository/lib/src/dm_repository.dart
+++ b/mobile/packages/dm_repository/lib/src/dm_repository.dart
@@ -171,8 +171,10 @@ const Duration _dmRelayListSignTimeout = Duration(seconds: 10);
 /// retry budget terminalised them into a red bubble. The cure is not a bigger
 /// number — the wrap builds now carry explicit bounds
 /// ([DmSendBudget.recipientWrapBuild], [DmSendBudget.selfWrapBuild]), which is
-/// what stops the *signing chain* outrunning this cap regardless of the
-/// signer's own per-op bound — Amber's `AndroidNostrSigner` allows 300s.
+/// what stops the *signing chain* outrunning this cap even when the signer path
+/// is slower. Amber's NIP-55 intent path for `nip44Encrypt` and `signEvent` is
+/// human-gated and unbounded, so a recipient-build timeout leaves the durable
+/// row pending for retry instead of red-failed.
 ///
 /// That covers the signing chain, not the whole send. Three awaited steps
 /// inside this cap are still unbounded and deliberately NOT in
@@ -3156,10 +3158,11 @@ class DmRepository {
         rumorId: rumor.id,
       );
     } else if (result.retryablePending && outgoingDao != null) {
-      // Soft-unconfirmed: the recipient frame was written but no NIP-20 OK
-      // arrived (and no explicit rejection). Keep the row pending — it
-      // renders as a plain optimistic bubble, never a red failure — and let
-      // the retry sweep re-drive it. A lost OK is not proof of loss.
+      // Soft retry: either the recipient frame was written but no NIP-20 OK
+      // arrived, or the pre-publish wrap build timed out while a human-gated
+      // signer approval may still be pending. Keep the row pending — it renders
+      // as a plain optimistic bubble, never a red failure — and let the retry
+      // sweep re-drive it.
       await _finalizeAfterRecipientUnconfirmed(
         outgoingDao: outgoingDao,
         rumorId: rumor.id,
@@ -3859,9 +3862,10 @@ class DmRepository {
       // refuses every time. This attempt delivered nothing.
       await _finalizeAfterRecipientBlocked(outgoingDao: dao, rumorId: rumorId);
     } else if (result.retryablePending) {
-      // Soft-unconfirmed retry: frame written, no OK yet. Keep the row
-      // pending (plain optimistic bubble, not a red failure) and let the
-      // next sweep re-drive it; only bump the retry count so it eventually
+      // Soft retry: frame written with no OK yet, or a pre-publish wrap build
+      // timeout while a human-gated signer approval may still be pending. Keep
+      // the row pending (plain optimistic bubble, not a red failure) and let
+      // the next sweep re-drive it; only bump the retry count so it eventually
       // exhausts.
       await _finalizeAfterRecipientUnconfirmed(
         outgoingDao: dao,
@@ -4256,11 +4260,12 @@ class DmRepository {
     }
   }
 
-  /// Apply the queue-row transition for a soft, retryable-pending recipient
-  /// publish — the relay frame was written but no NIP-20 `OK` arrived within
-  /// the confirm window (and no explicit rejection). A lost OK is not proof
-  /// of loss, so the row stays `pending` (rendering as a plain optimistic
-  /// bubble, never a red failure) and the retry sweep re-drives it.
+  /// Apply the queue-row transition for a soft, retryable-pending failure: an
+  /// inconclusive recipient publish (frame written, no NIP-20 `OK`, no explicit
+  /// rejection) or a bounded pre-publish wrap build timeout where a human-gated
+  /// signer approval may still be pending. The row stays `pending` (rendering
+  /// as a plain optimistic bubble, never a red failure) and the retry sweep
+  /// re-drives it.
   ///
   /// Only `incrementRetry` runs: it records the attempt timestamp so the
   /// next sweep applies backoff, and counts this attempt toward the retry
--- a/mobile/packages/dm_repository/lib/src/dm_repository.dart
+++ b/mobile/packages/dm_repository/lib/src/dm_repository.dart
@@ -170,9 +170,18 @@ const Duration _dmRelayListSignTimeout = Duration(seconds: 10);
 /// recipient had already received, parking them as soft-unconfirmed until the
 /// retry budget terminalised them into a red bubble. The cure is not a bigger
 /// number — the wrap builds now carry explicit bounds
-/// ([DmSendBudget.recipientWrapBuild], [DmSendBudget.selfWrapBuild]) so the
-/// chain cannot outrun this cap for ANY signer, including Amber, whose own
-/// per-op bound is 300s.
+/// ([DmSendBudget.recipientWrapBuild], [DmSendBudget.selfWrapBuild]), which is
+/// what stops the *signing chain* outrunning this cap regardless of the
+/// signer's own per-op bound — Amber's `AndroidNostrSigner` allows 300s.
+///
+/// That covers the signing chain, not the whole send. Three awaited steps
+/// inside this cap are still unbounded and deliberately NOT in
+/// [DmSendBudget.chainWorstCase]: `refreshPublicKey()` (a real round trip for
+/// any signer that doesn't cache the pubkey, e.g. a NIP-46 bunker — Keycast
+/// does cache it), the send-policy protected-minor check, and the connectivity
+/// probe. So this cap remains a genuine backstop for those, and
+/// [DmSendBudget.chainWorstCase] is a floor on what a send can cost, not a
+/// ceiling. Bounding them is tracked separately.
 ///
 /// A 15s cap here (the original value) fired during routine slow-Keycast
 /// sends, turning sends that were still legitimately in flight into
--- a/mobile/packages/dm_repository/lib/src/dm_repository.dart
+++ b/mobile/packages/dm_repository/lib/src/dm_repository.dart
@@ -16,6 +16,7 @@ import 'package:dm_repository/src/dm_decrypt_isolate.dart';
 import 'package:dm_repository/src/dm_decryption_worker.dart';
 import 'package:dm_repository/src/dm_reactions_repository.dart';
 import 'package:dm_repository/src/dm_repository_reportable_sites.dart';
+import 'package:dm_repository/src/dm_send_budget.dart';
 import 'package:dm_repository/src/dm_shared_video_citation.dart';
 import 'package:dm_repository/src/dm_sync_state.dart';
 import 'package:dm_repository/src/dm_verify_isolate.dart';
@@ -152,16 +153,31 @@ const Duration _dmRelayListSignTimeout = Duration(seconds: 10);
 /// durable queue keeps the row and the retry sweep re-drives it rather than
 /// parking a red chip.
 ///
-/// Every sub-step now carries its own tight bound — Keycast RPCs 12s each
-/// (`KeycastRpc.defaultRequestTimeout`), the recipient OK window 10s, the
-/// self-wrap publish 10s — so this cap exists only to bound a truly hung
-/// await, and it must sit ABOVE the capped worst case or it fires mid-send
-/// and misclassifies it. Remote-signer worst case: recipient wrap build
-/// (2 RPCs, 24s) + OK window (10s) + deferred self-wrap build (2 RPCs, 24s)
-/// + self-wrap publish (10s) ≈ 68s + overhead → 90s. A 15s cap here (the
-/// original value) fired during routine slow-Keycast sends, turning sends
-/// that were still legitimately in flight into eternally-retrying rows.
-const Duration _messagePublishTimeout = Duration(seconds: 90);
+/// Every sub-step carries its own bound — see [DmSendBudget], which owns the
+/// derivation. This cap exists only to bound a truly hung await, and it must
+/// sit ABOVE [DmSendBudget.chainWorstCase] or it fires mid-send and
+/// misclassifies it.
+///
+/// That is exactly what went wrong in #6586. The bound was hand-derived from a
+/// Keycast per-RPC timeout of 12s; #6075 raised that timeout to 30s (correctly
+/// — 12s had regressed video-publish signing, likes, reposts and follows) and
+/// the derivation here was never redone. A 1:1 send costs **four** measured
+/// signer round trips, so the honest chain became (2 × 30s) + 10s + (2 × 30s)
+/// + 10s = 140s against a 90s cap.
+///
+/// Because the self-wrap is built only after the recipient's `OK true`
+/// arrives, that 140s path was the *delivery* path: the cap fired on sends the
+/// recipient had already received, parking them as soft-unconfirmed until the
+/// retry budget terminalised them into a red bubble. The cure is not a bigger
+/// number — the wrap builds now carry explicit bounds
+/// ([DmSendBudget.recipientWrapBuild], [DmSendBudget.selfWrapBuild]) so the
+/// chain cannot outrun this cap for ANY signer, including Amber, whose own
+/// per-op bound is 300s.
+///
+/// A 15s cap here (the original value) fired during routine slow-Keycast
+/// sends, turning sends that were still legitimately in flight into
+/// eternally-retrying rows — the reason this backstop must stay generous.
+const Duration _messagePublishTimeout = DmSendBudget.messagePublishTimeout;
 
 /// Outcome of resolving a user's own kind-10050 DM inbox relay list (#4974).
 ///
--- a/mobile/packages/dm_repository/lib/src/dm_send_budget.dart
+++ b/mobile/packages/dm_repository/lib/src/dm_send_budget.dart
@@ -122,32 +122,31 @@ abstract final class DmSendBudget {
   /// wrong trade — it would push the send past the backstop and misclassify a
   /// delivered message as unconfirmed, which is the #6586 failure itself.
   ///
-  /// Applies only inside a send. The out-of-band rebuild is bounded by
-  /// [selfWrapRecoveryBuild], which is sized differently for the reason
+  /// Applies only when the caller has an outer cap. Uncapped callers use
+  /// [selfWrapUncappedBuild], which is sized differently for the reason
   /// documented there.
   static const Duration selfWrapBuild = Duration(
     seconds: _selfWrapBuildSeconds,
   );
 
-  /// Hard bound on rebuilding the self-addressed gift wrap on the out-of-band
-  /// recovery path (`NIP17MessageService.publishSelfWrap`, driven by
-  /// `DmRepository.recoverSelfWrap`).
+  /// Hard bound on building the self-addressed gift wrap when the caller has no
+  /// outer cap.
   ///
   /// Sized like [recipientWrapBuild], not like [selfWrapBuild]. The tightness
-  /// of [selfWrapBuild] buys headroom inside [messagePublishTimeout]; recovery
-  /// runs on its own with no outer cap to protect, so there a tight bound can
-  /// only fail a build the transport itself would have completed.
+  /// of [selfWrapBuild] buys headroom inside [messagePublishTimeout]; uncapped
+  /// callers have no outer budget to protect, so there a tight bound can only
+  /// fail a build the transport itself would have completed.
   ///
-  /// The rebuild is the same two round trips, so at [selfWrapBuild] it could
-  /// never finish against a signer running near its own 30s-per-op bound —
-  /// which is precisely the signer that leaves a self-wrap outstanding in the
-  /// first place. Recovery would then fail every attempt and the row would
-  /// exhaust its retry budget into a permanently missing self-wrap, losing the
-  /// cross-device copy this recovery arm exists to save.
+  /// This covers both out-of-band recovery
+  /// (`NIP17MessageService.publishSelfWrap`, driven by
+  /// `DmRepository.recoverSelfWrap`) and direct uncapped sends such as
+  /// `NIP17MessageService.sendPrivateMessage`. The build is the same two
+  /// round trips, so at [selfWrapBuild] it could never finish against a signer
+  /// running near its own 30s-per-op bound.
   ///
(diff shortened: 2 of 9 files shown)


mobile/packages/dm_repository/lib/src/dm_send_budget.dart
```diff
@@ -0,0 +1,129 @@
...
+  /// bound (`KeycastRpc.defaultRequestTimeout`), so this can never fail a
+  /// request the transport itself would have allowed. Tightening it below that
+  /// is the specific mistake #6046 made and #6075 reverted: production Keycast
+  /// runs 20-30s per single op under DB-pool contention (keycast#291), and a
+  /// tighter bound turns slow-but-succeeding sends into hard failures.
+  ///
+  /// It bounds the *chain*, not the transport, so it also covers signers whose
+  /// own per-op bound is far looser — Amber's `AndroidNostrSigner` allows 300s
+  /// per op, more than triple [messagePublishTimeout] itself.
+  static const Duration recipientWrapBuild = Duration(
```
mbradley, line 106: `recipientWrapBuild` is 60s and the recipient wrap is two Keycast ops at 30s each, so this sits at exactly `2 × KeycastRpc.defaultRequestTimeout` with nothing left for the encryption + isolate hop between the ops. Two ops that each succeed just under 30s plus that overhead exceed 60s and time out — failing a send both ops individually completed within the transport's own limit, which is what this doc says can't happen. It's the keycast#291 contention tail specifically, and it re-introduces the "bound sized at the worst case with no margin" shape #6586 was about, one step earlier in the chain.

Which way do you want to go: give this a few seconds of headroom (e.g. 65s, trimming `_headroomSeconds` to hold the 120s ceiling), or keep 60s and change the doc to say two simultaneously-slow ops can trip it? Either's fine — 60s-flat under a comment that promises it can't fail is the combination to avoid.

mbradley (commented):
Verified locally: analyze clean both layers, 565 package + 58 app-layer tests green, coverage 93.84% over the 93 gate, and the new bound/invariant tests are load-bearing — stripping the recipient-wrap timeout reds the "fails before publishing" test, zeroing the headroom reds the "backstop exceeds the chain" invariant. The core fix is sound and a clear improvement over a 90s cap the 140s chain was outrunning.

Two questions on the comments themselves, both your call — left inline. Neither blocks: the budget replaces a comment that misstated the arithmetic, and the two new comments make guarantees a touch stronger than the code holds, in the same file whose whole point is honest accounting.

mobile/packages/dm_repository/lib/src/dm_repository.dart
```diff
@@ -152,16 +153,31 @@ const Duration _dmRelayListSignTimeout = Duration(seconds: 10);
...
+/// signer round trips, so the honest chain became (2 × 30s) + 10s + (2 × 30s)
+/// + 10s = 140s against a 90s cap.
+///
+/// Because the self-wrap is built only after the recipient's `OK true`
+/// arrives, that 140s path was the *delivery* path: the cap fired on sends the
+/// recipient had already received, parking them as soft-unconfirmed until the
+/// retry budget terminalised them into a red bubble. The cure is not a bigger
+/// number — the wrap builds now carry explicit bounds
+/// ([DmSendBudget.recipientWrapBuild], [DmSendBudget.selfWrapBuild]) so the
+/// chain cannot outrun this cap for ANY signer, including Amber, whose own
```
mbradley: This promises the chain can't outrun the cap for any signer, but a few awaited steps inside the cap have no bound and aren't in `chainWorstCase`: `refreshPublicKey()` → `getPublicKey()` (a signer round trip — Keycast caches it, a NIP-46 bunker need not), the `_sendPolicy` protected-minor check, and the connectivity probe + `retryDisconnectedRelays()`. The probe and reconnect you've already listed under Out of Scope #2; the pubkey fetch and the policy check aren't, and both sit inside the backstop. So the "any signer" guarantee is stronger than the code delivers, and stronger than your own Out-of-Scope section admits.

Nothing behavioural needs to change here — but could you either fold the pubkey fetch and policy check into the tracked-unbounded list, or soften "any signer" to the best-effort framing? Trading the stale-arithmetic comment #6586 was for a new comment that overclaims the same way is the one outcome worth heading off.

hm21:
Pushed a64edecb7 — one finding, fixed rather than described.

The self-wrap build bound landed on `_publishSelfWrap`, which `publishSelfWrap` reaches too, and that path has no backstop to protect. At `selfWrapBuild` the rebuild — the same two round trips — could never finish against a signer running near its own 30s per-op bound, which is exactly the signer that leaves a self-wrap outstanding. All 5 recovery attempts would time out and `_markInterruptedExhausted` would terminalize the self wrap, losing the cross-device copy the arm exists to save.

Bound is now per caller: sends keep `selfWrapBuild`, recovery gets `selfWrapRecoveryBuild` sized like `recipientWrapBuild`, deliberately outside `chainWorstCase`. The regression test was confirmed red against the pre-fix code before I kept it.

`flutter analyze` clean on both layers, 566 package + 118 app-layer DM tests green, coverage 93.84%, all checks green. Yours to accept, tweak, or revert.

realmeylisdev:
Checks green (17/17) after `96265fad2`. Every item below is **fixed**; none escalated, none declined.

### @hm21 — self-wrap recovery bound (`a64edecb7`)

Accepted, and re-verified independently before keeping: `recoverSelfWrap` calls `publishSelfWrap` with no `.timeout()`, unlike `_sendRumorWithTimeout`, so the tight bound there protected no budget and could only fail a rebuild the transport would have completed. Correct diagnosis and the right shape. Thank you.

### @mbradley — inline 1: `recipientWrapBuild` at exactly `2 × 30s` (`a7b5df880`)

Confirmed. Inside that window, besides the two round trips, `GiftWrapUtil.getGiftWrapEvent` also does seal construction, the ephemeral keypair, a NIP-44 ECDH + ChaCha20 encryption and a secp256k1 signature — so two ops each returning just inside 30s do blow it, exactly as you said, and the doc claimed that couldn't happen.

Took your first option. The constant is now `_twoTransportBoundsSeconds + _wrapBuildLocalCryptoSeconds`, so the derivation is executable rather than a number, and the margin comes out of `_headroomSeconds` — `messagePublishTimeout` holds at 120s and `interruptedMinAge` doesn't move. The guard now asserts strictly greater than two transport bounds; confirmed red at 60s flat (`Expected: greater than 0:01:00, Actual: 0:01:00`) before keeping it.

### @mbradley — inline 2: the "ANY signer" overclaim (`a7b5df880`)

Confirmed, and you're right that it was the one outcome worth heading off. Verified all three sit inside the cap and outside `chainWorstCase`: `refreshPublicKey()` at `nip17_message_service.dart:404`, the policy check at `:354`, the connectivity probe at `:379`.

Scoped the claim to the signing chain, named the three gaps inline, and said plainly that `chainWorstCase` is a floor on a send's cost, not a ceiling.

### Unrequested — two more uncapped callers (`96265fad2`)

Found while patrolling for other instances of @hm21's class of bug, and it was **my regression**, so flagging it separately.

`DmRepository.sendFileMessage` and `BugReportService` reach `sendRumor` via `sendPrivateMessage` and impose no outer cap either. Before `510ce0fce` the self-wrap build was unbounded, so a signer at keycast#291's observed 23-30s/op completed the two round trips in 46-60s; at `selfWrapBuild` they now failed at 20s. Neither is queued in `outgoing_dms`, so `recoverSelfWrap` never runs for them and the sender's cross-device copy was lost for good.

`sendPrivateMessage` now opts into `selfWrapRecoveryBuild`. One implementation note worth stating, since it looks like the fail-unsafe direction: the parameter defaults to the **tight** bound, not the wide one. `sendRumor` is mocked in 25 places in `dm_repository_test` with every named argument enumerated, and mocktail silently stops matching a stub the moment the real call carries an argument the stub doesn't name — defaulting the other way reds 71 unrelated tests. Both direct callers are capped, so the default is correct for them; the doc says an uncapped caller must pass the wide bound. Regression test confirmed red both ways.

### Device patrol

Ran the branch as a debug build on the iOS simulator against production and watched the console. Two things worth reporting:

- The signed-in account is `authSource=importedKeys` → `LocalNostrIdentity`, and the session made **zero** Keycast RPC calls. That confirms at runtime that local-key signers spend no remote round trips — but it also means this device **cannot** exercise the 4-round-trip Keycast path these bounds guard. The bounds were verified by test, not on device.
- The outgoing-DM retry sweep ran clean on every pass (`self-wrap-recovered=0 self-wrap-failed=0 ... skipped-interrupted-too-young=0 aborted-not-ready=false`) and the DM history drain completed normally, so the `interruptedMinAge` change didn't strand anything. No DM-related errors in the session.

### One pre-existing gap, deliberately not touched

`publishSelfApplicationMarker` (#4977) calls `_buildWrap` with no bound at all — same unbounded-signer-round-trips class, but it predates this PR and none of my commits touch it. Left alone rather than scope-crept; happy to file it with the other three follow-ups.

realmeylisdev:
Filed the four follow-ups, so the Out-of-Scope section is now tracked rather than prose:

- #7090 — task(dm): adopt keycast's `nip17_wrap_batch` (four signer round trips → one). Blocked on confirming the verb is live in production keycast.
- #7091 — fix(dm): bound the send segments outside the publish backstop (kind-10050 inbox resolution, the connectivity probe, `retryDisconnectedRelays()`'s missing aggregate cap). This is the ceiling to the floor `chainWorstCase` describes.
- #7092 — task(keycast): reconcile the client's 30s per-op bound with keycast's 8s server bound. Notes that a 504 currently classifies as a hard failure, so the deployed keycast version changes which user-visible outcome a slow signer produces.
- #7093 — fix(dm): bound `publishSelfApplicationMarker`'s wrap build. Pre-existing, not a regression from this PR — filed so it isn't lost.

Each records what blocks it and what would unblock it. Nothing here changes this PR.

mbradley (dismissed):
Both questions from the prior review are resolved, verified at `a7b5df880`:

- Recipient-wrap bound now carries a named 5s margin over the two 30s transport ops (65s), and the invariant test is strict (`> 2×30`) so the no-margin sizing can't come back. Confirmed: zeroing the margin reds that test.
- The backstop doc no longer claims it covers "any signer" — it names `refreshPublicKey`, the policy check, and the connectivity probe as unbounded-and-tracked-separately, and frames `chainWorstCase` as a floor, not a ceiling.

The self-wrap recovery split (`a64edecb7`) is a good catch: recovery on the tight 20s bound would have failed every rebuild against the same slow signer that left the self-wrap outstanding. The 65s `selfWrapRecoveryBuild` fixes that and is covered — pointing recovery back at the tight bound reds the new test. 566 package tests green, coverage 93.84% over the gate.

mobile/packages/dm_repository/lib/src/nip17_message_service.dart
```diff
@@ -338,13 +340,26 @@ class NIP17MessageService {
   /// contract: soft-unconfirmed recipient publishes return retryable failure
   /// and the caller owns any fallback behavior.
   @useResult
+  /// [selfWrapBuildTimeout] bounds the self-wrap build, defaulting to the
```
NotThatKindOfDrLiz, line 350: This parameter doc is after `@useResult`, so it becomes the rendered doc comment for `sendRumor` instead of appending to the existing method docs. That drops the existing `selfWrapOnSoftUnconfirmed` and non-fatal self-wrap contract from hover/dartdoc. Move these lines into the existing doc block above `@useResult`.

mobile/packages/dm_repository/lib/src/dm_send_budget.dart
```diff
@@ -0,0 +1,186 @@
...
+  /// delivered message as unconfirmed, which is the #6586 failure itself.
+  ///
+  /// Applies only inside a send. The out-of-band rebuild is bounded by
+  /// [selfWrapRecoveryBuild], which is sized differently for the reason
+  /// documented there.
+  static const Duration selfWrapBuild = Duration(
+    seconds: _selfWrapBuildSeconds,
+  );
+
+  /// Hard bound on rebuilding the self-addressed gift wrap on the out-of-band
```
NotThatKindOfDrLiz: This is no longer recovery-only: `sendPrivateMessage` now passes this bound for uncapped file-message and bug-report sends. The name/doc should describe the caller shape, e.g. an uncapped self-wrap build, rather than saying it is only the out-of-band recovery path and outside sends.

mobile/packages/dm_repository/lib/src/nip17_message_service.dart
```diff
@@ -338,13 +340,26 @@ class NIP17MessageService {
...
+  /// tight [DmSendBudget.selfWrapBuild] because both direct callers
+  /// (`DmRepository._sendRumorWithTimeout`, `DmReactionsRepository`) wrap this
```
NotThatKindOfDrLiz: This "both direct callers" claim is false for reactions. `DmReactionsRepository` wraps `sendRumor` in a 15s timeout, which fires before `selfWrapBuild` at 20s and far before `recipientWrapBuild` at 65s, so the new internal bounds do not keep the reaction chain under that cap.

NotThatKindOfDrLiz (changes requested):
Requesting changes on the doc/accounting regressions and one explicit Amber decision before merge.

The core timeout fix is sound, and the prior reviewer findings around the 65s recipient bound and scoped signing-chain claim look resolved. The remaining blockers are narrower: the new `sendRumor` parameter docs currently replace the rendered method docs because they sit after `@useResult`; `selfWrapRecoveryBuild` is now used by uncapped send callers, not just recovery; the "both direct callers" sentence is false for reactions because their 15s cap fires before these new internal bounds; and Amber/NIP-55 needs an explicit decision because the 65s/20s bounds can now time out a human approval flow for first-DM seal signing.

I am taking over the clear wording/accounting fixes separately. The Amber behavior still needs the author/product decision called out before this should merge.

NotThatKindOfDrLiz:
@realmeylisdev I pushed takeover commit `84a3bbb42` with the clear review fixes:

- moved the new `selfWrapBuildTimeout` docs into the existing `sendRumor` doc block so dartdoc/hover keeps the full send contract;
- renamed `DmSendBudget.selfWrapRecoveryBuild` to `selfWrapUncappedBuild` and updated docs/call sites/tests so the bound describes both recovery and uncapped `sendPrivateMessage` callers;
- corrected the `sendRumor` docs so reactions are described as governed by their existing 15s `_publishTimeout`, not by the new internal build bounds.

Local verification:

- `flutter test test/src/nip17_message_service_test.dart` from `mobile/packages/dm_repository`
- `flutter analyze` from `mobile/packages/dm_repository`
- pre-push hook analyze + changed-file tests, including `packages/dm_repository/test/src/nip17_message_service_test.dart` and `test/services/outgoing_dm_retry_service_test.dart`

The remaining change request is the Amber/NIP-55 decision from the review: the current 65s/20s wrap-build bounds can still time out a human approval flow for first-DM seal signing. Please either adjust the bound strategy for that signer class or explicitly document/track the accepted behavior before merge. CI is running on the pushed commit now.

NotThatKindOfDrLiz (approved):
Takeover pass is clean at 4e0a710a8. I pushed one final fix to keep recipient-wrap construction uncapped for sendPrivateMessage callers that have no outgoing_dms retry row, while leaving queued durable sends on the 65s recipient-wrap bound and retryable-pending behavior.\n\nVerified locally: flutter test test/src/nip17_message_service_test.dart and flutter analyze from mobile/packages/dm_repository; flutter test from mobile/packages/dm_repository; flutter analyze from mobile; pre-push analyzer and changed-file tests. GitHub checks are green.\n\nI also closed stale follow-up #7093 because this PR now covers publishSelfApplicationMarker's timeout path, and updated the PR body to reflect the current commits and tracked follow-ups.

NotThatKindOfDrLiz:
@realmeylisdev I pushed takeover commit 4e0a710a8 to address the last Amber/uncapped-caller hole. sendPrivateMessage now leaves recipient-wrap construction uncapped, so file-message and bug-report sends do not get a retryablePending result with no queue row to retry from; queued durable sends keep the bounded recipient-wrap path.

Verification is green locally and in GitHub checks. I also closed stale #7093 and updated the PR body. Please review the takeover commit and merge when ready.

realmeylisdev:
@NotThatKindOfDrLiz Reviewed your takeover commit — it holds up. I checked the load-bearing claim rather than taking it on trust: both `sendPrivateMessage` callers (`DmRepository.sendFileMessage` and the bug-report send) do go direct, with no `_sendRumorWithTimeout` outer cap and no `outgoing_dms` row, so leaving the recipient build uncapped is right. Closing #7093 checks out too — `publishSelfApplicationMarker` is genuinely covered now.

Two things I pushed on top, both beyond what you reviewed.

**Rebased, `5ac95dc22` → `62928d09d`.** The branch had picked up a `Merge branch 'main'` commit from the Update-branch button, which `AGENTS.md` forbids. History is linear again on current `main`. Your three commits kept their authorship — only the SHAs moved (`4e0a710a8` → `8ba4a0486`). I also corrected one commit that was miscredited to an external open-source maintainer's git identity (a dependency author, not a Divine contributor) on internal DM work. That predated your rebase; left alone, the squash would have carried a `Co-authored-by:` for them into `main`.

**New commit `62928d09d`, not requested by your review.** It is the self-wrap mirror of the hole you closed. `sendPrivateMessage` still bounded its *self*-wrap build at `selfWrapUncappedBuild`, and on a remote signer that bound is reachable: `_buildBothWraps` prebuilds the self wrap only on the local-key path, so Amber and Keycast fall through to the lazy build under it. Past it the send returns `selfWrapPublished: false` — whose whole justification is that `recoverSelfWrap` finishes it out of band, and `recoverSelfWrap` requires the `outgoing_dms` row these callers never create. So the sender's cross-device copy was gone for good, announced by nothing louder than a log line. The build was unbounded before this PR, so the window is ours.

The fix passes `null` there, and gives `selfWrapBuildTimeout` the same null-means-uncapped semantics `recipientWrapBuildTimeout` already had — the two sat adjacent meaning opposite things for `null`. The existing uncapped-caller test now drives the rebuild past the widest bound rather than just past the tight one; it goes red against the pre-fix lib, which I checked.

Verified: `dm_repository` 570 tests pass, coverage 93.84% against the 93 gate, `flutter analyze` clean both in the package and in `mobile/`. All 18 checks green at `62928d09d`.

Since that last commit is a behavior change landing on an already-approved PR, I'd rather you cast an eye over it first — holding off on merging until you have.

PR #7078, opened 2026-08-11
test(recorder): cover classic mode with Maestro E2E

### LINKED ISSUE #7079: test(recorder): cover classic mode with Maestro E2E
Auto-created from PR #7078.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7078
**Branch:** `test/maestro-classic`

hm21:
Closes #7079

## Description

Extends the recorder E2E suite to the **classic** mode, the fourth of five, after #7018's capture coverage, #7043's lip-sync and #7036's stop-motion. Open, drive the action row, record a clip off the preview, prove a second recording stops itself at the duration limit, delete both, close.

**Classic needs more of its own files than the other three modes did, because it is a different stack rather than the same one with a different rail.** Capture, lip-sync and stop-motion all render `VideoRecorderCaptureStack` and differ only in the top bar's center slot and which rail controls they declare, which is why they share `driveCaptureRail` and most of each other's asserts. Classic renders `VideoRecorderClassicStack`: its own top bar, its own action rows above and below a square preview, and no control rail at all. So it gets its own mode assert, its own clip-recorded assert and its own record util. What it does share is the lens switch — the same control announcing the same `value` — so `classicModeControls` runs the existing `utils/driveLensControl.yaml` rather than duplicating it.

**The shutter is the preview, and it gets its own identifier.** There is no `camera_record_button` in classic; the square preview is wrapped in a `Semantics` node that flips its label between "tap to start" and "tap to stop". That node now carries `camera_classic_shutter`. Reusing `camera_record_button` would have made `utils/recordCaptureClip.yaml` reusable for exactly one line, at the cost of putting the same id on a different widget — which `SemanticIds` explicitly forbids, and which would have made the leak-guard `assertNotVisible` lines in the other three asserts ambiguous. A widget test pins that the anchor survives both recording states, since the label around it does not.

**Next is disabled, not hidden — and the asserts have to read it that way.** The capture stack wraps next and delete in an `AnimatedOpacity` and lets `RenderAnimatedOpacity` drop them from the semantics tree on an empty session, which is what every existing `assert…Mode` reads as "nothing recorded yet". Classic's top bar keeps next mounted and passes `onPressed: null`. So `assertClassicMode` asserts `enabled: false` there and `assertClassicClipRecorded` waits on `enabled: true` — a visibility wait would pass instantly against an empty session and prove nothing. Delete does still leave the tree, so it stays asserted the usual way. `video_recorder_classic_top_bar_test.dart` pins both halves of that contract, so a later refactor toward the capture stack's shape fails there rather than silently in a lane CI does not run.

**`classicModeRecordingLimit` covers the one behaviour classic alone declares.** `hasRecordingLimit` is true only for this mode, and it is what makes the recorder hand the native camera a `maxDuration` — the session's remaining share of the 6.3s total. Capture and lip-sync pass none and record until the user stops them. The test starts a recording, never stops it, and waits for the top bar to come back, which only the native auto-stop can do. It deliberately asserts *that* rather than *when*: the budget left depends on what `classicModeRecordClip` spent. That coupling is also why `recordClassicClip` records for two seconds where `recordCaptureClip` records for three — the bloc refuses to start a recording below 30ms remaining, so a longer first clip would fail the limit test on its opening wait rather than on the limit.

**`openClassicMode` guards its wheel hops instead of tapping unconditionally.** Classic is the last entry and sits three places from Capture, further than any other mode has to travel on a lazy `ListView` that only builds entries near the armed one. So the hops step one entry at a time, and they are guarded on Classic not already being on screen: selecting a mode persists it, so a rerun opens straight on Classic, from where Stop Motion is two entries away and may not be built at all. `openLipSyncMode`'s unconditional two taps work because Lip Sync is only two places out; that does not generalise to the end of the wheel.

**Ghost is driven last in the controls test.** Each toggle raises the same 4-second confirmation snackbar stop-motion has, but classic's controls are a horizontal row above the mode wheel rather than a vertical rail on the right. By layout the banner sits below that row and does not cover the two taps that restore the toggle — but running lens and grid first means that if it turns out to overlap on some screen size, the failure lands on that control alone instead of taking the rest of the test with it.

One behaviour fix came out of the tagging: classic's shutter computed `isEnabled` — no camera, or the 6.3s budget spent — and passed it only to the gesture detector, so the `Semantics` node announced a tappable button either way. The capture stack's record button has always reported it. The flag is read by a readiness wait in `recordClassicClip` and another in `classicModeRecordingLimit`, which is what turns a dead viewfinder into an immediate failure instead of one fifteen seconds later on the close-button wait. Not on `assertClassicMode`: that file runs from `openClassicMode` the moment the recorder opens, so an `enabled: true` there would fail the whole mode assert on a camera that is merely still warming up. `ScreenshotMode` is deliberately not mirrored: the record button ORs it in so screenshot builds render the visual prominent, and classic's preview has no disabled visual to keep honest.

Seven identifiers were added along the way — six the other modes already had, plus the shutter's own. Classic's grid and ghost toggles were deliberately left untagged in #7036 on the grounds that no flow drove them and an identifier nothing asserts on drifts silently; one does now, so they are tagged, and that comment in `SemanticIds` is updated rather than left contradicting the code.

## Out of Scope

- **Classic's "next".** It skips the editor entirely — classic is the one mode declaring no `hasVideoEditor` — and pushes the metadata screen with a render already started. Same reason stop-motion's assemble step is out: it leaves the recorder, and there is no coverage on the other side to hand off to.
- **Upload mode**, the fifth and last. It has no camera, no shutter and no clips; its stack is a static explainer, so it shares nothing with the four flows and needs a different shape of test.
- **The smoke suite.** The flow stays out for the same reason the other three do — the shutter needs real camera hardware to leave its disabled state, and the smoke lane runs on the iOS Simulator.
- **A negative-control pass.** #7036 proved its `checked:` selectors really applied by asserting them against the opposite state and watching them fail. The same is worth doing here for the new `enabled:` assertions on next, and is part of the device pass below.

## Verification

**All four recorder flows run green end to end on a Samsung Galaxy SM-S942B**, STAGING debug build (`DEFAULT_ENV=STAGING`, `GH_ACTIONS_PR_PREVIEW=true`):

| Flow | Commands | Failed | Skipped | Wall-clock |
|---|---|---|---|---|
| classic | 194 | 0 | 2 | 153s |
| stop-motion | 208 | 0 | 1 | 170s |
| capture | 203 | 0 | 1 | 167s |
| lip-sync | 225 | 0 | 1 | 197s |

The other three were re-run because this PR touches them — a `camera_classic_shutter` leak-guard line in each `assert…Mode`, and the util extraction in `stopMotionModeControls`. All five per-control utils fire in the stop-motion run, the two new ones included, and `driveFlashControl` was not skipped, so the flash block was exercised for real rather than guarded past.

Every skipped block is a guard doing its job: the "Why six seconds?" onboarding dialog that did not appear (all four runs), and `openClassicMode`'s wheel hops on the second open, where Classic was already armed. The hops themselves ran on the first open, so both branches of that guard are covered.

The device runs a `de-CH` system locale, so the run went through the scratch flow README step 0b describes rather than `classicModeFlow` itself: `pm clear` and the `en-US` app-locale override applied from adb, in that order, then everything from `classicModeOpen` onward byte-for-byte as the flow runs it. Same workaround lip-sync and stop-motion needed.

**The first device run failed, and it found a real flow bug** — `classicModeRecordingLimit` could not start, shutter `enabled: false`, progress bar completely full off a single clip. Classic shares one 6.3s budget and Maestro charges its own overhead to it: the start tap cost 3.5s and the stop tap 2.0s, and the `waitForAnimationToEnd: 2000` between them measured 3.7s, since the live preview never settles and it runs its timeout out and then some. ~7.9s against a 6.3s cap, so the native auto-stop was ending the recording instead of the stop tap. That broke both shutter tests at once and hid it — `classicModeRecordClip` stopped proving the tap-to-stop it is named for and still passed green, while the limit test had no budget to start on. Dropping the wait puts the pair at ~4.1s and leaves ~2.1s, which is what the limit test runs on.

What the device settled that nothing local could:

- **The native auto-stop reaches the Flutter side as a clip.** `classicModeRecordingLimit` taps once, the top bar leaves, and it comes back on its own with the session holding the clip — the flagged unknown, and it holds.
- **The ghost-frame snackbar does not cover classic's action row.** `classicModeControls` drives all three controls and closes on its mode assert.
- **`enabled: false` on next and `enabled: true` on the shutter both resolve on device**, as does `checked:` on grid and ghost.
- **Two deletes, and the assert between them holds**, which is what proves the session held two clips.

Local, on the final commit:

- `flutter test test/widgets/video_recorder/` — 291 passing
- `flutter test test/screens/video_recorder_screen_test.dart test/models/video_recorder/` — passing
- `dart format` — clean

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/mobile/e2e/maestro/asserts/assertClassicMode.yaml
+++ b/mobile/e2e/maestro/asserts/assertClassicMode.yaml
@@ -1,0 +1,59 @@
+appId: co.openvine.app
+---
+# Classic mode, idle: the viewfinder chrome is up and no clip is recorded yet.
+#
+# Classic is the one recording mode that is *not* the capture stack. It has its
+# own top bar, its own action rows, and no record button at all — the square
+# preview is the shutter. So unlike the other three asserts, most of this file
+# is pinning a different screen rather than the same one with a different rail.
+
+# Mode wheel with Classic armed. `selected` is load-bearing: the wheel renders
+# an entry for every mode in every mode, so a bare id assertion here passes
+# just as happily on the Lip Sync tab.
+- assertVisible:
+    id: camera_mode_classic
+    selected: true
+- assertVisible:
+    id: camera_classic_shutter
+- assertVisible:
+    id: camera_close_button
+- assertVisible:
+    id: camera_library_button
+
+# Action row under the preview, left to right.
+- assertVisible:
+    id: camera_switch_camera_button
+- assertVisible:
+    id: camera_grid_button
+- assertVisible:
+    id: camera_ghost_frame_button
+
+# Nothing recorded yet — and classic reports that differently from the capture
+# stack, which is the trap this file exists to pin. Next stays mounted and goes
+# *disabled* (`onPressed: null` while `!hasClips`), so `assertNotVisible` here
+# would fail on an empty session. Delete does drop out of the tree: its row
+# sits inside an AnimatedOpacity, and RenderAnimatedOpacity drops its child
+# from the semantics tree at opacity 0.
+- assertVisible:
+    id: camera_next_button
+    enabled: false
+- assertNotVisible:
+    id: camera_delete_clip_button
+
+# The capture stack, absent. Flash and aspect ratio are unconditional there —
+# every mode that renders that rail renders both — so their absence is what
+# proves the classic stack is up rather than capture's, and the record button's
+# absence proves the shutter really is the preview. Timer and stabilization are
+# the declared-only pair, the same leak guard assertStopMotionMode carries: a
+# control starting to render in a mode that does not declare it fails a run
+# rather than passing quietly.
+- assertNotVisible:
+    id: camera_record_button
+- assertNotVisible:
+    id: camera_flash_button
+- assertNotVisible:
+    id: camera_aspect_ratio_button
+- assertNotVisible:
+    id: camera_timer_button
+- assertNotVisible:
+    id: camera_stabilization_button
--- a/mobile/e2e/maestro/tests/classicModeControls.yaml
+++ b/mobile/e2e/maestro/tests/classicModeControls.yaml
@@ -10,7 +10,7 @@ tags:
 # Classic renders three controls where capture renders five and stop-motion
 # five: a lens switch, the grid toggle and the ghost toggle. There is no flash
 # control (classic has no rail at all), no aspect ratio (classic is square by
-# definition — `defaultAspectRatio` is the only mode that returns it), and
+# definition — the only mode whose `defaultAspectRatio` returns square), and
 # neither of the declared-only pair.
 #
 # The lens switch is the same control every other mode drives, announcing the
--- a/mobile/e2e/maestro/tests/classicModeControls.yaml
+++ b/mobile/e2e/maestro/tests/classicModeControls.yaml
@@ -1,0 +1,97 @@
+appId: co.openvine.app
+tags:
+  - recorder
+---
+#TC to drive the classic action row,
+# given the recorder is open on an idle classic viewfinder with no clips,
+# validate each control actually changes recorder state, and leave every one
+# of them back on its default.
+#
+# Classic renders three controls where capture renders five and stop-motion
+# five: a lens switch, the grid toggle and the ghost toggle. There is no flash
+# control (classic has no rail at all), no aspect ratio (classic is square by
+# definition — `defaultAspectRatio` is the only mode that returns it), and
+# neither of the declared-only pair.
+#
+# The lens switch is the same control every other mode drives, announcing the
+# same `value`, so it runs the shared ../utils/driveLensControl.yaml. Grid and
+# ghost carry a Semantics `toggled` flag rather than a `value`, which Flutter
+# maps to the accessibility node's checked state, so they are driven with
+# `checked:` — see the note on that selector in e2e/maestro/README.md.
+#
+# Preconditions beyond an open viewfinder:
+# - **Real camera hardware**, for the lens switch. See classicModeRecordClip.
+# - **A grid that starts on.** The grid preference persists
+#   (`camera_grid_lines_enabled`) and defaults to on for grid-supporting modes,
+#   so step 2 reads it as on. Run from the top this is safe —
+#   loginFreshInstall's clearState wipes the key.
+# - **An empty session.** Not for the controls themselves — none of the three
+#   is gated on clips, unlike capture's aspect-ratio control — but for the mode
+#   assert this closes on, which pins next disabled and delete absent.
+#Steps:
+# 1. Camera: Back -> Front -> Back
+# 2. Grid: on -> off -> on
+# 3. Ghost frame: off -> on -> off
+
+# 1. Lens switch.
+- runFlow: ../utils/driveLensControl.yaml
+
+# 2. Grid overlay.
+- assertVisible:
+    id: camera_grid_button
+    checked: true
+- tapOn:
+    id: camera_grid_button
+- extendedWaitUntil:
+    visible:
+      id: camera_grid_button
+      checked: false
+    timeout: 10000
+- tapOn:
+    id: camera_grid_button
+- extendedWaitUntil:
+    visible:
+      id: camera_grid_button
+      checked: true
+    timeout: 10000
+
+# 3. Ghost frame (onion-skin), deliberately last. Not persisted, so it always
+# starts off — but each tap raises a 4-second confirmation snackbar that floats
+# at the bottom of the screen, and classic's controls are a horizontal row
+# rather than stop-motion's vertical rail. The row sits above the mode wheel
+# and clear of the banner, but running the other two controls first means a
+# mistake there fails on this control alone instead of taking the rest with it.
+- assertVisible:
+    id: camera_ghost_frame_button
+    checked: false
+- tapOn:
+    id: camera_ghost_frame_button
+- extendedWaitUntil:
+    visible:
+      id: camera_ghost_frame_button
+      checked: true
+    timeout: 10000
+# Restore. Leaving it on is not just untidy: the overlay draws the previous
+# clip's last frame over the live preview, which makes a failure screenshot
+# from any later test much harder to read.
+- tapOn:
+    id: camera_ghost_frame_button
+- extendedWaitUntil:
+    visible:
+      id: camera_ghost_frame_button
+      checked: false
+    timeout: 10000
+
+# Let the banner go before handing over to the shutter tests — it floats over
+# the bottom of the screen, and classic's shutter is the preview above it, but
+# a banner still up during a recording lands in every failure screenshot.
+#
+# `waitForAnimationToEnd` runs its timeout out rather than returning early: it
+# waits for the screen to stop changing and the live camera preview never
+# does, which makes it the deterministic wait this needs.
+- waitForAnimationToEnd:
+    timeout: 4500
+
+# Everything back to default, so the tests after this one start where they
+# expect to.
+- runFlow: ../asserts/assertClassicMode.yaml
--- a/mobile/e2e/maestro/tests/classicModeControls.yaml
+++ b/mobile/e2e/maestro/tests/classicModeControls.yaml
@@ -13,11 +13,13 @@ tags:
 # definition — the only mode whose `defaultAspectRatio` returns square), and
 # neither of the declared-only pair.
 #
-# The lens switch is the same control every other mode drives, announcing the
-# same `value`, so it runs the shared ../utils/driveLensControl.yaml. Grid and
-# ghost carry a Semantics `toggled` flag rather than a `value`, which Flutter
-# maps to the accessibility node's checked state, so they are driven with
-# `checked:` — see the note on that selector in e2e/maestro/README.md.
+# All three are shared per-control utils: the lens switch with every mode that
+# renders a rail, grid and ghost with stop-motion, the other mode that declares
+# them. So what this file actually owns is the order — see step 3 — not the
+# sequences themselves. Grid and ghost carry a Semantics `toggled` flag rather
+# than a `value`, which Flutter maps to the accessibility node's checked state,
+# so they are driven with `checked:` — see the note on that selector in
+# e2e/maestro/README.md.
 #
 # Preconditions beyond an open viewfinder:
 # - **Real camera hardware**, for the lens switch. See classicModeRecordClip.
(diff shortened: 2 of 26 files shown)


mobile/e2e/maestro/asserts/assertClassicMode.yaml
```diff
@@ -0,0 +1,59 @@
...
+# is pinning a different screen rather than the same one with a different rail.
+
+# Mode wheel with Classic armed. `selected` is load-bearing: the wheel renders
+# an entry for every mode in every mode, so a bare id assertion here passes
+# just as happily on the Lip Sync tab.
+- assertVisible:
+    id: camera_mode_classic
+    selected: true
+- assertVisible:
+    id: camera_classic_shutter
```
realmeylisdev, line 17: **should-fix.** The app-side `Semantics(enabled:)` fix has no consumer, so the rationale in the description doesn't land.

The description says: *"Left as it was, `assertClassicMode`'s `assertVisible` would pass on a dead viewfinder and `classicModeRecordClip` would fail fifteen seconds later on the close-button wait instead of on the precondition it missed."*

But this assertion still carries no `enabled:` qualifier, and `camera_classic_shutter` appears in seven other places, none of which read the flag either: `assertNotVisible` in `assertCaptureMode` / `assertLipSyncMode` / `assertStopMotionMode` / `classicModeOpen`, and a bare `tapOn` in `recordClassicClip` (twice) and `classicModeRecordingLimit`. A selector only filters on enabled when the field is present, so the three taps land regardless. So on a camera-less device this file still passes and `recordClassicClip` still dies 15s later on `notVisible: camera_close_button`, exactly as it would have before the fix.

The change itself is right on a11y grounds and the two new stack tests pin it. This is only about the missing consumer.

Adding `enabled: true` here is the obvious move but it makes the *whole* file fail on a cold camera, and `openClassicMode` runs it immediately after opening the recorder. A readiness wait where the precondition actually matters gets the stated benefit without that:

```yaml
# utils/recordClassicClip.yaml, before the first tap
- extendedWaitUntil:
    visible:
      id: camera_classic_shutter
      enabled: true
    timeout: 15000
```

Either way works for me; the thing to avoid is shipping the flag with nothing reading it.

mobile/e2e/maestro/tests/classicModeRecordingLimit.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+# 2. No second tap. The close button coming back is the auto-stop: nothing else
+# in this mode restores the top bar. The timeout covers a full 6.3s budget plus
+# finalization, so it holds even if step 1 of the flow recorded almost nothing.
+- extendedWaitUntil:
+    visible:
+      id: camera_close_button
+    timeout: 20000
+
+# 3. And the recording became a clip rather than being discarded at the limit.
```
realmeylisdev: **nit:** this comment claims more than the assertion can detect at this point.

The session already holds `classicModeRecordClip`'s clip when this test starts, so `assertClassicClipRecorded` (next `enabled: true` + delete visible) is *already* true before step 1 runs. It can't tell a second clip from a discard.

The two-clip fact is genuinely pinned, one test later: `classicModeDeleteClip` taps delete once and re-runs the same assert, which only passes if there were two. That file's comment already says so. Worth either rewording this one to what it does prove (the recorder came back to a sane idle state) or pointing at `classicModeDeleteClip` for the real check, since "the recording became a clip rather than being discarded" is exactly the failure mode the description flags as most worth watching on the first device run.

mobile/e2e/maestro/tests/classicModeControls.yaml
```diff
@@ -0,0 +1,97 @@
...
+#   assert this closes on, which pins next disabled and delete absent.
+#Steps:
+# 1. Camera: Back -> Front -> Back
+# 2. Grid: on -> off -> on
+# 3. Ghost frame: off -> on -> off
+
+# 1. Lens switch.
+- runFlow: ../utils/driveLensControl.yaml
+
+# 2. Grid overlay.
```
realmeylisdev, line 41: **nit:** this grid block is byte-identical to `stopMotionModeControls.yaml`'s (I diffed them), and the ghost block below differs only in its comment.

The suite's own convention is a per-control util once two modes share a control, which is why `driveLensControl` / `driveFlashControl` / `driveAspectRatioControl` exist. `SemanticIds`' new doc in this PR already asserts the outcome: *"they carry the same ids and the E2E flow drives them with the same per-control utils"* — true of the lens switch, not of these two.

`utils/driveGridControl.yaml` + `utils/driveGhostFrameControl.yaml` would make that sentence true and leave `classicModeControls` as what it's really about, which is the ordering (ghost last, then the banner wait).

mobile/test/widgets/video_recorder/modes/classic/video_recorder_classic_stack_test.dart
```diff
@@ -43,12 +46,14 @@ void main() {
     Widget buildWidget({
       VideoRecorderState recordingState = VideoRecorderState.idle,
       List<DivineVideoClip>? clips,
+      bool isCameraInitialized = true,
+      bool canRecord = true,
```
realmeylisdev, line 53: **nit:** `canRecord` is never overridden — every `buildWidget` call in this file takes the default `true`, so the parameter is unexercised scaffolding.

`isCameraInitialized` earns its place ("reports the shutter disabled before the camera is ready" passes `false`). Either drop this one, or add the `canRecord: false` case — it's the other half of the gate the new `enabled:` reports, and the one that covers "budget spent" rather than "camera not up".

realmeylisdev (changes requested):
went through the classic flow against the widget tree. hop order on the wheel, the `enabled:` / `checked:` selectors, the delete sequence and the 6.3s budget arithmetic all hold up, and `flutter test test/widgets/video_recorder/` gives me the same 289 green.

four notes inline, one that matters: nothing in the suite actually reads the shutter's new `enabled` flag, so it doesn't buy what the description credits it with.

the one i'd hold on though is the device pass. the body still says this is a draft pending a hardware run, but it's marked ready. did that run happen? shipping a suite nobody has executed is the part i can't sign off from reading alone.

mobile/e2e/maestro/asserts/assertClassicMode.yaml
hm21: Confirmed and fixed in 0d4b414. You were right that all eight selectors were bare, so the flag bought nothing.

Took your `recordClassicClip` suggestion rather than `enabled: true` here, for the reason you gave — `openClassicMode` runs this file the moment the recorder opens, so it would fail the mode assert on a camera that is merely still warming up.

Added the same wait to `classicModeRecordingLimit` too. It taps the shutter directly instead of going through the util, and it runs against a session that already holds a clip, which is exactly where the *other* arm of the gate — a spent 6.3s budget — is what makes the tap do nothing. That arm now has a widget test as well (8cc8455).

mobile/e2e/maestro/tests/classicModeRecordingLimit.yaml
hm21: Right, fixed in b1cfc0b. The assert is true on the way in, so it cannot tell a second clip from a discard. Reworded to what it does prove and pointed at `classicModeDeleteClip` for the count. Also fixed the step-3 line in the header, which made the same claim.

mobile/e2e/maestro/tests/classicModeControls.yaml
hm21: Diffed it too — grid is byte-identical, ghost differs only in the comment. Extracted both into `utils/driveGridControl.yaml` and `utils/driveGhostFrameControl.yaml` in 0207562, and `stopMotionModeControls` now calls them as well, so the `SemanticIds` sentence is true rather than aspirational.

What is left in each `…Controls` file is the ordering, which is the part that genuinely differs: stop-motion drives ghost first because its rail is vertical and steps 2-5 double as the banner wait; classic drives it last and follows with an explicit one. Flagged in the description that `stopMotionModeControls` is a device-verified file that changed, even though no step did.

mobile/test/widgets/video_recorder/modes/classic/video_recorder_classic_stack_test.dart
hm21: Fixed in 8cc8455, and covering it turned up why the parameter was awkward to reach. One correction to the framing: `canRecord` comes from `_cameraService.canRecord`, so it is a second camera-readiness signal rather than the budget — the budget arm is `hasRemainingDuration` off the clip manager.

That arm was unreachable here: `buildWidget` left `recorderMode` on the state default of `capture`, so `hasRecordingLimit` was false in a file that only ever renders the classic stack. Pinned the mode and all three arms now have a case — camera not initialized, camera service refusing, budget spent. Every existing test stays green because an empty session has the full budget.

hm21:
@realmeylisdev thanks — all four land, and the first one was a real hole: the `enabled:` flag shipped with nothing reading it, so the description credited it with a benefit it did not have.

| # | Item | Verdict | Commit |
|---|---|---|---|
| 1 | `Semantics(enabled:)` has no consumer | fixed | `0d4b414` |
| 2 | limit test's step-3 comment overclaims | fixed | `b1cfc0b` |
| 3 | grid/ghost blocks duplicated | fixed | `0207562` |
| 4 | `canRecord` unexercised | fixed | `8cc8455` |

Details are in the thread replies. One correction on #4: `canRecord` is `_cameraService.canRecord`, a second camera-readiness signal rather than the budget — the budget arm is `hasRemainingDuration`. Chasing it turned up that `buildWidget` left `recorderMode` on the state default of `capture`, so `hasRecordingLimit` was false and that arm was unreachable in a file that only ever renders the classic stack. All three arms have a case now.

Also flagging one thing beyond the four, since it changes what a re-run needs: #3 rewrites `stopMotionModeControls` to call the two new utils. Pure extraction, no step changed, but it is a device-verified file, so it wants a re-run rather than being taken on trust — same as the three `assert…Mode` files this PR added a leak-guard line to.

`flutter analyze lib test integration_test` clean, `flutter test test/widgets/video_recorder/` 291 green, `check_refs.sh` clean.

---

On process, though: per `PR_REVIEW.md` these were yours to push. `divine-mobile` is not in the `PR_REVIEW_TEAMS.md` mapping, so the default applies — any org member with write access may take over once the gates pass, and they did: you are an org member with `write`, the PR left draft at 19:22Z and you reviewed at 20:04Z, the branch is same-repo and mutable, and E2E yaml plus test scaffolding is nowhere near the workflow/security/infra carve-out. Items 2, 3 and 4 are mechanical with one obvious remediation each — a reworded comment, an extraction the suite's own convention already dictates, a parameter to drop or exercise. Those are the definition of a clear finding, and the rule is push them as separate commits with a handback, not describe them and wait a round-trip.

Item 1 is the fair exception: you offered two options and said either works, and choosing between them is the judgment call the rules say to discuss. Fine to leave that one as a comment — though even there you had a recommendation, and it was the right one.

This is the same note as last time, so worth making it stick: when you can see the fix, push the fix.

hm21:
@realmeylisdev device pass done — you were right to hold on it, and it was the only thing that could have caught this.

**SM-S942B, STAGING debug, green end to end: 192 commands, 0 failures, 153s.** Details in the rewritten Verification section.

The first run failed, at `classicModeRecordingLimit`'s opening wait — shutter `enabled: false`, progress bar completely full off one clip. Classic shares one 6.3s budget and Maestro charges its own overhead to it: start tap 3.5s, stop tap 2.0s, and the `waitForAnimationToEnd: 2000` between them measured 3.7s because the live preview never settles and it runs its timeout out and then some. ~7.9s against a 6.3s cap, so the native auto-stop was ending the recording rather than the stop tap.

That is worse than one red test. `classicModeRecordClip` was passing green while proving the opposite of its name — the auto-stop, not tap-to-stop — and the limit test then had nothing left to start on. Fixed in `f0089fa` by dropping the wait: the two taps span ~4.1s and leave ~2.1s.

Worth noting on your thread #1: the readiness wait you asked for is what made this diagnosable. It failed immediately on `enabled: false` with a hierarchy dump naming the state, instead of 15s later on `notVisible: camera_close_button` — which is the misleading symptom the bare version would have produced, and which I would likely have chased into the app.

Answers on device to the three unknowns: the native auto-stop does reach the Flutter side as a clip, the ghost-frame snackbar does not cover classic's horizontal action row, and `enabled:` / `checked:` both resolve.

One gap I am not going to paper over: `stopMotionModeControls` and the three `assert…Mode` files changed in this PR and were not re-run. Additive line plus a step-for-step identical extraction, so I expect them fine, but they are unverified on hardware and the description says so.

hm21:
Closed the verification gap — the other three flows are re-run on the SM-S942B too, so every flow this PR touches is now green on hardware rather than assumed.

| Flow | Commands | Failed | Skipped | Wall-clock |
|---|---|---|---|---|
| classic | 194 | 0 | 2 | 153s |
| stop-motion | 208 | 0 | 1 | 170s |
| capture | 203 | 0 | 1 | 167s |
| lip-sync | 225 | 0 | 1 | 197s |

The two things worth naming, since they are what was unverified:

- **The util extraction holds.** All five per-control utils fire in the stop-motion run — `driveGhostFrameControl` and `driveGridControl` included — and `driveFlashControl` was not skipped, so the flash block ran for real rather than being guarded past.
- **The leak-guard line passes in all three.** `camera_classic_shutter is not visible` asserts clean in `assertCaptureMode`, `assertLipSyncMode` and `assertStopMotionMode`.

Every skipped block is a guard behaving: the "Why six seconds?" onboarding dialog that did not appear, in all four runs, plus `openClassicMode`'s wheel hops on the second open where Classic was already armed — the hops ran on the first open, so both branches are covered.

Same de-CH workaround as before: `pm clear` and the `en-US` override from adb, then the flow with only `loginFreshInstall`'s `clearState` lifted out. No code changed for this, so no new commit — description updated and the "not re-verified on hardware" caveat is gone.

realmeylisdev (approved):
lgtm. the device run answers the one thing i was holding on, and the four earlier notes all check out.

pushed `ef02d61d2`: the readme still said `classicModeFlow` had never run on hardware and to treat a first failure there as a probable flow bug, which the same file contradicts 260 lines down. revert freely.

the description carries the same staleness in its `recordClassicClip` records-for-two-seconds line, from before you dropped the wait.

Issue #7071, opened 2026-08-11
feat(trust-safety): block or mute every account that claims a badge, including future claimants


hm21:
## What would you like?

Two related actions on a badge's detail screen:

1. **Bulk block / mute** — one button that blocks (or mutes) every account holding the badge, instead of opening each profile in turn.
2. **Forward-looking block** — once a badge is blocked, accounts that claim it *later* are covered automatically, without the user having to revisit the badge and re-run the bulk action.

User story: as someone viewing a badge page, I want a "Block all" / "Mute all" action so I can moderate my feed in one step, and have it keep holding as the group grows.

## How would this be useful for you?

Badges are group identity. When a group uses a shared badge to coordinate harassment, per-profile blocking is whack-a-mole: the target has to keep researching and blocking each new account by hand, while the group only has to add one more account to be back in the feed. The work is entirely on the person being harassed, and it never ends.

Bulk + forward-looking blocking flips that asymmetry. One action covers the group as it exists, and it keeps covering the group as it changes.

## When would you use this?

Reported by a community member who is being harassed by a group that organises around a shared badge. Their words: they should not have to constantly research and block new members who claim a tag that has been used to organise people to harass them.

More generally: any time a badge stops being a credential and starts being a brigading signal.

## Anything else?

Implementation notes and the decisions this needs before it can be built:

**The data is already on screen.** `BadgeDetailData.recipients` — each entry carrying `isAccepted` — already drives the recipient list at [badge_detail_screen.dart:244](mobile/lib/screens/badges/badge_detail_screen.dart#L244). Bulk-block has everything it needs for the "currently holding" half.

**Scope to the coordinate, not the name.** Key the block on `BadgeCoordinate` (`30009:<issuer-pubkey>:<d-tag>`), never the display name. Otherwise a same-named badge from a different issuer either evades the block or gets caught by it wrongly.

**Decide "awarded" vs "claimed".** A NIP-58 award (kind 8) is authored by the *issuer*, so it lands on a recipient without their consent — blocking on awarded-alone would let a bad-faith issuer get third parties blocked by awarding them a toxic badge. Claiming is the recipient pinning the award into their own kind 30008 profile badges, which is a deliberate act of identification. `isAccepted` already distinguishes the two. The requester's framing ("anyone who EVER claims that identity") points at claimed.

[...]

Issue #7027, opened 2026-08-10
fix(blocklist): unblock does not stick — a superseded own block list is merged back in with no freshness check


realmeylisdev:
Reproduced on an iOS simulator against production on 2026-08-10, on a real signed-in account.

## Symptom

Unblocking does not stick. The app reports success, logs the removal, and the account stays blocked — permanently, across an app restart.

## Reproduction

1. Block an account from the DM conversation long-press sheet. Snackbar: "Blocked user".
2. Unblock it from the conversation kebab → "Unblock {name}?" → Unblock.
3. The UI still treats the account as blocked. Navigating away and re-entering the thread does not clear it.
4. Terminate the app (which flushes `NSUserDefaults`) and inspect the container:

```
flutter.blocked_users_list.9be8bd90d818407bcf574d11b1c57f104fd53f40fa767abc4a631ee2694b43a3
  => ["b2380758159d89727a81e9b76b2e2a51216d341b4deddf0540f27cf7bb121738"]
```

The pubkey is still there, after this in the logs:

```
[18:46:44.457] Synced own mute list: +0 -1 (total: 0)
[18:46:44.691] Published block list to Nostr with 1 entries
[18:46:44.691] Removed user from blocklist: b2380758159d89727a81e9b76b2e2a51216d341b4deddf0540f27cf7bb121738
```

Earlier in the same session, the re-add is visible directly:

```
[18:46:43.289] Received list event from relay: 5d1b6b54ca532209c02a4306ca3c008a85411446f93b33c7ba68a5339c6a4137
[18:46:43.567] Synced own mute list: +1 -0 (total: 1)
[18:46:43.633] Restored 1 blocks from relay (total: 1)
```

The same event id `5d1b6b54ca532209c02a4306ca3c008a85411446f93b33c7ba68a5339c6a4137` is delivered repeatedly across the session (18:39:39, 18:39:40, 18:46:43, 18:52:14).

## Root cause

`_handleOwnBlockListEvent` (`mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart:1480-1506`) restores our own block list from relay-stored events, and it is **merge-only with no freshness guard**:

```dart
final added = relayPubkeys.difference(_runtimeBlocklist);
if (added.isEmpty) return;
_runtimeBlocklist.addAll(added);
unawaited(_saveBlockedUsers());
```

[...]

Issue #6437, opened 2026-07-27
fix(blocklist): _restoreBlocksFromRelay only ever adds, so a stale relay copy can re-block someone you just unblocked


realmeylisdev:
Found while root-causing #6435.

## Problem

`_handleOwnBlockListEvent` / `_restoreBlocksFromRelay` in `mobile/packages/content_blocklist_repository/lib/src/content_blocklist_repository.dart:1447-1451` merges the `p` tags of our own relay-stored kind-30000 into `_runtimeBlocklist` and **never removes**:

```dart
final added = relayPubkeys.difference(_runtimeBlocklist);
if (added.isEmpty) return;
_runtimeBlocklist.addAll(added);
unawaited(_saveBlockedUsers());
```

The intent is legitimate — restore blocks after a reinstall, where SharedPreferences is wiped but the relay still holds the event. But because it is add-only and unconditional, any stale copy of our own list re-adds entries we have since removed, and `_saveBlockedUsers()` then persists them.

There is no `created_at` guard on this path. `_latestBlockListEventCreatedAtByAuthor` is consulted in `_handleOthersBlockListEvent`, not here.

## Why it matters

This is a live risk rather than a theoretical one, because the relay demonstrably serves superseded versions of these exact events (divinevideo/divine-funnelcake#718). An unblock that was correctly applied and correctly published can be reverted locally on the next launch, and then re-persisted — with no user-visible signal.

Note the reporting session in #6435 shows `Loaded 1 blocked users from persistence` on the unblocking device. I could not confirm whether that entry was the unblocked party or a third account, so this issue is filed on the code shape, not on a reproduced instance.

## Suggested direction

Apply the same replaceable-event discipline used elsewhere in the file: track the newest own-list `created_at`, ignore older events, and treat the newest own event as authoritative — replacing the set rather than merging into it. `_handleOwnMuteListEvent` already does exactly this for kind 10000 ("the newest own event is the complete list and replaces `_mutedPubkeys` wholesale"), so this is bringing the block path in line with its mute twin.

Care needed: the reinstall-restore case must still work, and a genuine local block made before the relay round-trip completes must not be clobbered.

## Related

- Discovered during #6435 / #6434
- Sibling defects from the same audit: #6436, #6438
- Relay behaviour that makes this reachable: divinevideo/divine-funnelcake#718

NotThatKindOfDrLiz:
Fixed — closing as verified against current `main`.

[...]

PR #7082, opened 2026-08-11
feat(trust-safety): block badge claimants

NotThatKindOfDrLiz:
## Summary
- add a badge-detail action to block every account currently claiming the badge
- resolve claimants with a reverse profile-badge #a query across kind 10008 and legacy 30008, then confirm the latest profile-badge list still contains the coordinate
- bound claimant verification by checking latest profile-badge lists in author chunks instead of one lookup per claimant, and exclude the current viewer from the result
- add a batch blocklist repository API so bulk blocking saves once and publishes the Nostr lists once per batch
- localize the confirmation/success/failure copy and cover the repository, cubit, and widget paths with tests

Refs #7071

## Notes
- Uses the reverse #a query rather than BadgeDetailData.recipients because the detail recipient list is award-based and acceptance checks are intentionally capped for rendering; the trust-safety action needs currently claimed profile-badge state, including self-claims that are not discoverable through awards.
- This PR handles the current-claimants half only. Forward-looking badge rules, bulk unblock semantics, and larger unblock freshness work remain tracked in #7071 / #7027 / #6437.
- The claimant discovery query is intentionally bounded at the newest 1000 profile-badge claim events, then latest-state verification is chunked by author. The confirmation screen shows the concrete count that will be blocked.
- The confirmation remains count-only for this scoped PR; showing a claimant preview would be a product expansion on top of this bulk action.

## Verification
- validated relay.divine.video accepts reverse #a profile-badge queries
- flutter gen-l10n
- flutter test (mobile/packages/badge_repository)
- flutter test (mobile/packages/content_blocklist_repository)
- flutter test test/blocs/badges/badge_detail_cubit_test.dart test/screens/badges/badge_detail_screen_test.dart
- flutter analyze
- mobile/scripts/golden.sh verify
- push hook analyzer and changed-file tests


hm21 (approved):
LGTM — the batch `blockUsers` API is the right shape and the tests are real. Approving.

I pushed one fix (16bc874d7): the new `badgeDetailBlockClaimants*` strings shipped as raw English in `ms`/`ur`/`vi`/`zh`, and the Latin-script locales lost their diacritics — `"fuer"`/`"koennen"` (de), `"Nous n avons trouve ... a bloquer"` (fr), `"incarca"` (ro), `"yuklenemedi"` (tr) — plus `"claimant"`/`"claim"` left untranslated in de/it/nl/sv. Retranslated all ten keys per locale, added the `few`/`many` plural categories pl and ro need, fixed the Arabic hamzas, and restored each file's indentation. `arb_consistency_test` only checks key presence, which is why CI was green.

Two things I left for you:

**`loadClaimantPubkeys` fans out unbounded.** The `#a` query returns up to 1000 candidates and `_profileBadgesByPubkey` then fires two REQ subscriptions per candidate (10008 + legacy 30008), all concurrent, no cap — `nostr_sdk` has no limiter either. `loadBadgeDetail` right above it caps the same lookup at 50 for exactly this reason ("so a long list costs a bounded number of relay lookups"). A batched `authors:` filter per chunk would turn 2N queries into ~2.

**The 1000 limit is silent.** On a badge with more claimants the screen still says "Block N accounts" and blocks only what came back. For a trust & safety "block all" that's worth surfacing, even if only in the copy.

Minor: the viewer's own pubkey isn't excluded from the set — if you accepted the badge you're in your own block list, and `blockUsers` only skips self when `_ourPubkey` happens to have been set by mute-list sync (the cubit doesn't pass `ourPubkey:` the way `other_profile_bloc` does).

NotThatKindOfDrLiz:
@hm21 thanks for the review and the l10n fix. I pushed e546dab99, rebased onto current main, and addressed the three items:

- bounded claimant verification with author chunks instead of per-pubkey REQs
- excluded the viewer before blocking
- named and documented the 1000 candidate cap, with the scoped behavior called out in the PR body

I also added widget coverage for the empty claimants state, load retry path, and failure snackbar.

Verification:
- `flutter test` in `mobile/packages/badge_repository`
- `flutter test` in `mobile/packages/content_blocklist_repository`
- `flutter test test/blocs/badges/badge_detail_cubit_test.dart test/screens/badges/badge_detail_screen_test.dart`
- `flutter analyze`
- push hook analyzer and changed-file tests

Please re-review.

NotThatKindOfDrLiz:
Small follow-up: `origin/main` moved again while CI was running, so I rebased once more and pushed final head `cf99900db`. The same verification passed again locally, plus the pre-push analyzer and changed-file tests. Fresh GitHub checks are queued on the final head.

NotThatKindOfDrLiz:
Correction on the final head after one more main update: rebased and pushed `9f69933f`. Local branch is current with `origin/main`, the pre-push analyzer and changed-file tests passed again, and fresh GitHub checks are queued on that head.

PR #7084, opened 2026-08-11
fix(auth): handle an undeployed resend endpoint and a double-tapped signup

### LINKED ISSUE #7089: fix(auth): handle an undeployed resend endpoint and a double-tapped signup
Auto-created from PR #7084.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7084
**Branch:** `fix/headless-resend-404-guard`

dcadenas:
Closes #7089

## Description

Fixes to the email-verification screen, all in the same area: what the app does
when the server cannot complete a verification the way the app expects.

- **Tell a missing resend endpoint apart from a declined resend.** A 404 from
  `/api/headless/resend-pin` means the endpoint is not deployed; a 4xx from a
  deployed endpoint means the resend was refused. These were indistinguishable,
  so an undeployed server looked like a rejected request.
- **Keep resend tappable when the endpoint is missing server-side.** Following
  from the above, an absent endpoint no longer disables the affordance.
- **Say why verification polling stopped** instead of dropping the spinner. The
  poll window closing left the screen with no explanation of what to do next.
- **Pin single registration on a double-tapped signup submit.** A regression
  test: two taps on Create account must produce one registration.

That last one is the app-side half of a problem measured on the server. In
production, of 587 repeat registrations for the same address, 77 arrived within
ten seconds of the previous attempt, which is the signature of a double submit.
keycast handles this server-side by superseding the earlier pending
registration; this test pins the client so it does not generate the duplicate in
the first place.

Review passes over the above added:

- **Screen-reader announcements** for both new messages. Each replaces a spinner
  with text after an async transition that moves no focus, so a screen reader
  was told nothing on a screen whose only job is explaining the wait.
- **Edge-triggered listeners.** The announcements were first written as one
  listener over the union of status, pinStatus and resendStatus, inspecting
  current values only. That re-announced every message that happened to be true
  whenever any single field moved, so a resend after polling had already stopped
  repeated "we stopped checking" and then interrupted itself. Each listener is
  now edge-triggered on its own field, which also fixes the pre-existing PIN
  failure announcement that had the same shape.
- **Timeout coverage on both resend methods.** Neither had a timeout test, so
  that path was unverified in both directions. Now pinned.
- **A corrected ARB description**, which said "shown under a disabled Resend
  button" while the change above deliberately keeps it tappable. That
  description is the brief translators work from.
- **Two redundant `TimeoutException` clauses removed** whose bodies were
  byte-identical to the `catch` immediately below them.
- **The PIN-fallback feature flag is removed**, so the PIN entry and resend row
  now ship to everyone. See the section below for why that changed.

### Why the PIN fallback is no longer gated

This started behind `FF_EMAIL_VERIFICATION_PIN_FALLBACK` (internal, default
off), on the grounds that keycast had no verify-pin endpoint. That is no longer
true, and leaving the flag off is now the broken state rather than the safe one:

- **keycast#268 is merged and deployed.** Merged 2026-08-11T21:06:37Z. Probed
  read-only afterwards, `/api/headless/resend-pin` and `/api/headless/verify-pin`
  both return **422** rather than 404, so the routes exist and are rejecting an
  empty payload. Earlier in the day both returned 404; the deploy landed in
  between.
- **Production emails already carry the PIN.** keycast's `email_service.rs`
  renders a 6-digit block and the plaintext reads "Or enter this verification
  code in the app". With the flag off, users are told to enter a code into an
  app that shows no field to enter it in.
- **There is one server to be in parity with.** `oauthConfig` in
  `auth_providers.dart` returns `https://login.divine.video` for every
  environment except `local`, so poc, staging, test and production all reach the
  same keycast.

So the flag is deleted rather than defaulted on. A dart-define that can only be
flipped by shipping a new build is not a useful kill switch for a flow whose
users are mostly first-run installs, and keeping a dead lever around is debt.

**Related Issue:** Relates to divinevideo/keycast#268

Not blocked by it, and it has since merged and deployed. The resend changes were
written to handle the endpoint being *absent*, which was production's state when
this branch was opened. They now also cover the deployed case: a 404 means "this
server build cannot resend", whatever the reason, and any other 4xx means the
resend was declined.

## Out of Scope

- The Patrol integration suite for this flow, which needs separate repair and is
  going up as its own pull request.

### Open question for review

`ResendVerificationError` carries five cases. Two of them change behaviour:
`unavailable` keeps the button tappable and points at the PIN, `expired` says
the signup is gone and stops offering a retry that cannot work. The remaining
three, `declined`, `server` and `network`, collapse into one retryable failure
status.

Keeping that finer granularity rather than cutting it to
`unavailable` / `expired` / `failure` is a design call worth a second opinion.
Flagged as a deferral, not an oversight.

## Verification

- `dart format --output=none --set-exit-if-changed` on the changed Dart files, clean
- `flutter analyze lib test integration_test`, no issues
- `flutter gen-l10n`, regenerated output is committed and reproduces with no diff
- `flutter test` on the affected suites, 366 passed (`arb_consistency_test`,
  `email_verification_screen_test`, `create_account_screen_test`,
  `email_verification_cubit_test`, `secure_account_screen_test`)
- `flutter test` in `packages/keycast_flutter`, 261 passed

Every new test was checked against the code reverted: the announcement
listeners, the close-button `Semantics` wrapper, the token-mode guard and the
resend classifier each turn their test red when removed.

Also exercised end to end against a keycast built from `main` at `5068590`, run
in the local stack: the resend-unavailable, resend-expired and resend-success
states were each driven from the UI, and the client's own classification was
read back from the log at every tap. Details in the review threads.

On l10n: the two new keys are **English-only**. They exist in `app_en.arb` and
are recorded in `_knownUntranslatedDebt` in `arb_consistency_test.dart`, which is
the allowed path when translation is deferred. The per-locale files in this diff
are generated output carrying the English fallback, not translations, so
non-English users see English here until a translation pass runs.

### Screenshots

Captured from a debug build of this branch (`40b5acc69`) on an Android emulator,
run against the local stack with the PIN fallback forced on:

```
--dart-define=DEFAULT_ENV=LOCAL
--dart-define=FF_EMAIL_VERIFICATION_PIN_FALLBACK=true
```

The flag was still in place when these were captured and had to be forced on,
because the resend affordance lives inside the PIN-fallback section. It has
since been removed, so a build of the current head shows these states with no
dart-define at all. The
local stack's keycast returns 404 on `/api/headless/resend-pin`, the same
condition production is in, so the unavailable state is reached through the real
code path with no production signup and no real email address.

| State | What it shows |
| --- | --- |
| Baseline | Normal polling: spinner, "Waiting for verification", Resend in its default state. For comparison. |
| Resend unavailable | The new message, with **Resend still green and tappable** above it. |
| Polling stopped | The 15-minute window has closed; the spinner is replaced by an explanation instead of vanishing. |

Two details the images cannot show on their own:

- **Resend stays live.** Two consecutive taps each produced
  `Resend verification failed (ResendVerificationError.unavailable); status=ResendStatus.unavailable`,
  23 seconds apart. The button is genuinely still tappable after entering the
  unavailable state, which is the behaviour this PR changes.
- **The timeout was waited out, not shortened.** Polling restored at 17:08:07,
  timed out at 17:23:08. `_pollingTimeout` was not patched to make the capture
  cheaper.

The resend-unavailable message is absent from the polling-stopped capture by
design, not by accident: `_onTimeout` emits a fresh `EmailVerificationState`
rather than a `copyWith`, which resets the resend fields to idle.

**Double-tapped signup is deliberately not captured.** The guarantee is *one
registration rather than two*, which is a state fact rather than a visual one. A
still frame of a button cannot show it and a recording of two taps would show
only a busy spinner. The evidence is the regression test in this diff,
corroborated by a count of rows in `oauth_codes` whose `pending_email` matches
the test address after two taps: one row, not two.

Earlier captures of this flow exist but were taken from a `main` build, before
this branch's code existed. They are deliberately not attached, because they
would show behaviour this diff did not produce.

## Type of Change


NotThatKindOfDrLiz:
@dcadenas I took over the two review findings and pushed `fix(auth): handle expired headless resends`.

What changed:
- token-mode polling timeouts now render the same “we stopped checking” explanation that screen readers announce, instead of dropping the spinner with no visible recovery text.
- headless resend `success:false` expired-registration responses now map to a distinct expired resend state, with copy that tells the user to start signup again instead of retrying a resend that cannot work.
- added regression coverage for expired resend parsing/cubit/UI and token-mode timeout copy; regenerated l10n fallback output.

Verification:
- `flutter test test/screens/auth/email_verification_screen_test.dart test/blocs/email_verification/email_verification_cubit_test.dart test/l10n/arb_consistency_test.dart`
- `flutter analyze lib test` in `mobile/packages/keycast_flutter`
- `flutter test` in `mobile/packages/keycast_flutter`
- push hook analyze + changed-file tests passed
- GitHub CI is green on the rebased head

NotThatKindOfDrLiz (approved):
Takeover fixes are pushed, rebased, locally verified, and green in CI. Looks ready once Daniel is comfortable with the changes.

mobile/packages/keycast_flutter/lib/src/oauth/oauth_client.dart
```diff
@@ -951,12 +957,20 @@ class KeycastOAuth {
...
+      return ResendVerificationResult.failure(
+        _resendErrorFor(response.statusCode),
+      );
     } catch (_) {
-      return ResendVerificationResult.failure();
+      return ResendVerificationResult.failure(ResendVerificationError.network);
     }
   }
 
+  static ResendVerificationError _resendErrorFor(int statusCode) {
```
mbradley, line 968: This is where the `expired` recovery path breaks, and I could not find a keycast response that reaches it.

keycast returns an expired pending registration as **HTTP 410 GONE**, body `{"error": "Registration expired. Please register again."}` (`AuthError::RegistrationExpired`, `api/src/api/http/auth.rs:600-602`). `_resendErrorFor(410)` is not 404 and not `>= 500`, so it returns `declined`, which the cubit folds into `ResendStatus.failure`. The user gets the generic "we couldn't resend, try again" copy with the button still enabled, and the `ResendStatus.expired` UI (disabled button, "start again") never renders.

The other producer of `expired`, `_errorCodeFromMessage` (`headless_models.dart:311`), cannot cover it either: it runs only on a 2xx body, it reads `json['message']` while keycast error bodies use `error`, and it requires the substring `sign up` while keycast says "register again". Three independent mismatches, each fatal on its own.

The new test `reports an expired registration response as expired` passes only because it builds a 200 response with `{success: false, message: "...sign up..."}`, a shape keycast does not send; this same file already models the real one at line 931 (`{'error': 'pin_expired'}, 410`). So no test exercises the real contract, and CI stays green over the dead branch.

Two ways forward, your call:
- **Keep `expired`:** confirm the deployed `resend-pin` response for an expired registration, then classify off the status or a stable coded error (`410 -> ResendVerificationError.expired` in this method looks right), and replace the fabricated test with one that feeds the real response, which will currently go red.
- **Drop it:** if the deployed contract does not distinguish expired on this route, remove the `expired` enum case, `ResendStatus.expired`, its UI branch, the ARB string, and the test. AGENTS.md bars both dead code and a test that cannot fail.

One tell: the description's "Open question for review" lists four `ResendVerificationError` cases and omits `expired`, and says only `unavailable` currently changes behaviour, which is literally true precisely because `expired` never reaches the UI.

If you do wire `expired` up, the branch also needs a labeled way out: in `polling`/`pollingTimedOut` the "start again" copy names no control (only the unlabeled X), and the PIN-expired error tells the user to "tap resend" while resend is disabled in that state.

mobile/lib/screens/auth/email_verification_screen.dart
```diff
@@ -459,103 +457,125 @@ class _EmailVerificationScreenState
...
-                      ),
-                      EmailVerificationStatus.polling => _PollingContent(
-                        email: state.pendingEmail,
-                        isPollingMode: widget.isPollingMode || !_isTokenMode,
-                        showPinFallback: pinFallbackEnabled,
+            ),
+            BlocListener<EmailVerificationCubit, EmailVerificationState>(
+              listenWhen: (previous, current) =>
+                  previous.resendStatus != current.resendStatus &&
+                  current.resendStatus == ResendStatus.unavailable,
```
mbradley, line 486: This listener announces `unavailable` to screen readers, but the reachable `failure` state (the cubit folds `declined`, `server`, and `network` all into `ResendStatus.failure`) renders red error text after the same async resend with no announcement, and so does `expired`. The least severe outcome is spoken; the actual failures are silent. Per `accessibility.md`, an async state change that swaps in visible content should announce.

Add sibling edge-triggered listeners so the error states get the same treatment this one gives `unavailable`:

```dart
BlocListener<EmailVerificationCubit, EmailVerificationState>(
  listenWhen: (previous, current) =>
      previous.resendStatus != current.resendStatus &&
      current.resendStatus == ResendStatus.failure,
  listener: (context, state) => SemanticsService.sendAnnouncement(
    View.of(context),
    context.l10n.authVerificationResendFailed,
    Directionality.of(context),
  ),
),
```

Add the matching one for `expired` (`authVerificationResendExpired`) if that state becomes reachable per the other thread.

mobile/lib/screens/auth/email_verification_screen.dart
```diff
@@ -800,13 +818,21 @@ class _PollingContent extends StatelessWidget {
                     padding: const EdgeInsets.only(top: 32, bottom: 32),
                     child: Column(
                       children: [
-                        if (isPollingMode && showPinFallback) ...[
+                        if (isPollingMode) ...[
```
mbradley, line 854: With `emailVerificationPinFallback` removed, `_PinEntrySection` now shows whenever `isPollingMode`, unconditionally. The flag's own description was "off by default until keycast verify-pin support is deployed," and the client still carries 404 handling for both `verifyPin` and `resendHeadlessVerification` that anticipates a server build without those routes.

Is verify-pin support plus the PIN-in-email deployed across every server build a user can reach here (prod, staging, poc, and any older build still in the field), and is losing the only off-lever acceptable for a first-run flow? On a build without it, the "enter the 6-digit code from your email" prompt points at a code the email may not carry, and the field accepts input that can only fail. The description argues prod is covered by keycast#268; the open question is parity everywhere else and whether a build-time-only lever is the right tradeoff.

mbradley (changes requested):
Reviewed at head `123776efb`, tiered heavy for the first-run signup surface. I checked the resend classification against keycast's live error contract, and ran the double-tap guarantee, the announcements, and the polling-stopped path.

The 404/unavailable handling, the double-tap single-registration tests (both the cubit and the widget one fail on a regression, verified), the polling-stopped copy, and the edge-triggered listener rework all hold up.

One blocker and two questions, inline:

1. The `expired` resend branch cannot fire against keycast's real response, so it currently ships as dead code behind a test that asserts a response keycast never sends. On `_resendErrorFor`.
2. Removing the PIN-fallback flag ships PIN entry unconditionally. A question about server-build parity, on the PIN gate.
3. A screen-reader gap: the reachable `failure` resend state is announced to nobody while the less severe `unavailable` is. Suggested change on the listener block.

Requesting changes on (1); (2) and (3) are yours to resolve.

rabble (dismissed):
Requesting changes on the current 123776efb head. The expired resend path does not match the real service contract: an expired pending registration is HTTP 410 with an error body, but _resendErrorFor currently maps 410 to declined, while the new regression test fabricates a 200 success-false message shape. Classify and test the real 410 response, or remove the unreachable expired state. If expired stays, reconcile the disabled resend control with copy that says to tap resend and provide a labeled way out. The reachable ResendStatus.failure async update also needs the same edge-triggered screen-reader announcement as unavailable; announce expired too if retained. Finally, removing the PIN fallback flag needs explicit deployment-parity signoff for every reachable server environment because older builds without verify-pin support otherwise expose a field that can only fail.

mobile/packages/keycast_flutter/lib/src/oauth/oauth_client.dart
realmeylisdev: Checked this against keycast and the 410 belongs to a different endpoint, so the `expired` branch here does fire.

`resendHeadlessVerification` posts to `/api/headless/resend-pin`, which is `headless_resend_pin` in `api/src/api/http/headless.rs`. On an expired pending registration it returns:

```rust
if pending.expires_at <= Utc::now() {
    return Ok(Json(HeadlessResendPinResponse {
        success: false,
        message: "This registration has expired. Please sign up again.".to_string(),
    }));
}
```

`Ok(Json(...))` is a 200, not a 410, and keycast's own integration test pins it:

```rust
// test_resend_pin_on_an_expired_registration_reports_failure
assert_eq!(response.status(), axum::http::StatusCode::OK);
assert_eq!(response_json(response).await["success"], false, ...);
```

`AuthError::RegistrationExpired` in `auth.rs` is a different condition. Its only raise site is inside `perform_email_verification` (the email-link path), and the variant comment reads `// Async bcrypt timed out (instance died)`, so it means the hash worker died rather than that the registration window closed. `headless_resend_pin` returns `HeadlessError`, whose only GONE variant is `PinExpired`, and that handler never constructs it. The `From` impls cannot produce it either: `sqlx::Error` maps to `Internal`, `RepositoryError` to `Conflict` / `InvalidRequest` / `ServiceUnavailable`, `BcryptQueueError` to `ServiceUnavailable`. So this route cannot answer 410 at all.

On the three mismatches, against this endpoint all three line up. It is a 2xx, the body field really is `message` (`HeadlessResendPinResponse { success, message }`, no `error` field and no code), and the text is "Please sign up again", not "register again". Lowercased it contains registration, expired and sign up, so `_errorCodeFromMessage` returns `expired`.

Worth adding because it is the one thing that could have made the branch dead anyway: `find_by_device_code` has no expiry predicate (`WHERE device_code = $1 AND tenant_id = $2 AND pending_email IS NOT NULL`), so the row is returned and the `expires_at` check is genuinely reached rather than short-circuiting into the uniform success.

So the test in this diff builds the real response rather than a fabricated one. Message matching is the only discriminator available since that response carries no error code, and it degrades safely: if keycast rewords, it falls back to `declined` and resend stays retryable.

mobile/lib/screens/auth/email_verification_screen.dart
realmeylisdev: Agreed, and fixed. Both `failure` and `expired` render red error text while `unavailable` is the neutral secondaryText one, so the mildest outcome was the only one being spoken.

Added the two sibling edge-triggered listeners in the same shape as the `unavailable` one. Both strings already existed in the ARB (`authVerificationResendFailed`, `authVerificationResendExpired`), so no l10n change was needed. Regression test covers both, and I checked it goes red with the listeners removed.

mobile/lib/screens/auth/email_verification_screen.dart
realmeylisdev: Checked the parity question and it is narrower than it looks, because there is only one non-local keycast.

`oauthConfig` in `mobile/lib/providers/auth_providers.dart` sends every environment except `AppEnvironment.local` to `https://login.divine.video`. `poc`, `staging`, `test` and `production` all resolve to that same origin, so there is no separate server build a user can reach here.

Probed it read-only, empty JSON body:

```
POST /api/headless/resend-pin  -> 422
POST /api/headless/verify-pin  -> 422
```

422 rather than 404 means both routes exist and are rejecting the payload, so keycast#268 is deployed. Both routes landed in that one commit, so their presence is the deploy signal for the PIN-in-email change too.

The rollback case stays covered rather than becoming a gap: a server without the routes 404s, which this PR maps to `unavailable` and keeps the affordance tappable. That is the degradation path the resend work already built.

realmeylisdev (approved):
went through the keycast side on this. the 410 both change requests cite is from `auth.rs` on the email-link path. `/api/headless/resend-pin` answers an expired registration with a 200 and `success:false`, and keycast's own integration test pins exactly that, so the expired branch does fire and the test in this diff is the real contract. details inline.

pushed two commits: screen-reader announcements for the `failure` and `expired` resend states (only the neutral one was being spoken), and a one-line correction to the l10n debt note. revert either freely.

also confirmed both pin routes are live on login.divine.video, which is the only non-local keycast the app can reach.

mobile/packages/keycast_flutter/lib/src/oauth/oauth_client.dart
dcadenas: **Declined: the 410 is real, but it belongs to `verify-pin`.**

`resendHeadlessVerification` posts to `/api/headless/resend-pin`. That handler answers an expired registration with **HTTP 200** and `success: false`, from the `expires_at <= Utc::now()` branch of `headless_resend_pin` in keycast `api/src/api/http/headless.rs`:

```rust
return Ok(Json(HeadlessResendPinResponse {
    success: false,
    message: "This registration has expired. Please sign up again.".to_string(),
}));
```

`Ok(Json(..))` is a 200, and the body key is `message`. So all three mismatches you list resolve the other way: the body is 2xx, it does use `message`, and it does say "sign up again".

keycast's own suite already pins this. `test_resend_pin_on_an_expired_registration_reports_failure` (`headless.rs:2302`, added in keycast#268) backdates `expires_at` and asserts `StatusCode::OK` with `success == false`, plus that `pin_hash` is unchanged.

`AuthError::RegistrationExpired` (410, `auth.rs:709`) is thrown from `perform_email_verification` and means the async password hash died. Different route, different cause. The `410 {'error': 'pin_expired'}` at `oauth_client_test.dart:929` is `verifyPin`, a third endpoint.

Confirmed live against a keycast built from `main` at `5068590`. One pending row, both endpoints, seconds apart:

```
resend-pin : {"success":false,"message":"This registration has expired. Please sign up again."}   HTTP 200
verify-pin : {"error":"That code has expired. Please request a new one.","code":"PIN_EXPIRED"}    HTTP 410
```

The real client, pointed at that server with no mocks, returned `ResendVerificationError.expired`, and the app rendered the expired copy with resend disabled.

The test in this diff is therefore the real contract, and I have left it. What I did add is a doc comment on `_errorCodeFromMessage` naming the keycast source of the string and warning that a reword upstream degrades this to `declined`, plus a test pinning that fallback. Matching on prose is the genuine weak point here, and worth a stable error code on the route when one exists.

One thing worth knowing that neither review raised: a cleanup task deletes expired pending rows every 300s, after which the row is *absent* rather than expired and resend answers with the uniform `success: true`. That narrows the production window for this branch considerably. Tracked in divinevideo/keycast#362, which also records that it is not a reason to drop the client handling, since the client cannot tell absent from expired.

mobile/lib/screens/auth/email_verification_screen.dart
dcadenas: Fixed in 14038ad5.

Both `failure` and `expired` now get sibling edge-triggered listeners in the same shape as the `unavailable` one, using `authVerificationResendFailed` and `authVerificationResendExpired`. Both strings already existed in the ARB, so there is no l10n change.

Tests cover each announcement, and a third asserts the failure message is spoken once when unrelated state fields move while `resendStatus` stays `failure`. That one guards the edge trigger itself, which is the regression 4f7dd369 already fixed once on this branch.

mobile/lib/screens/auth/email_verification_screen.dart
dcadenas: Answered, and the parity surface is smaller than it looks: there is one non-local keycast.

`oauthConfig` in `mobile/lib/providers/auth_providers.dart` returns `https://login.divine.video` for every environment except `AppEnvironment.local`, so poc, staging, test and production all reach the same server. There is no separate staging or poc keycast to audit.

Both routes are live there. Probed read-only with an invalid body, where 422 means the route matched and deserialization failed, and 404 would mean absent:

```
POST https://login.divine.video/api/headless/resend-pin  -> 422
POST https://login.divine.video/api/headless/verify-pin  -> 422
```

The PIN ships in the registration email, not only on resends: `headless_register` passes `Some(&pin)` to `send_verification_email`.

On losing the lever: it was a `bool.fromEnvironment` default-off, so flipping it required shipping a new build. For a first-run flow that is not a kill switch, and with keycast#268 deployed, off is now the state that shows "enter the 6-digit code" next to no field to enter it in. Happy to revisit if you would rather this ride behind a runtime flag, but that would be a different mechanism, not this one restored.

dcadenas:
Pushed 431f3724 and f893d229 on top of Meylis's two commits. All 17 checks are green. Every item from both change requests is sorted below, including the ones I am declining.

I am the author here, so I cannot approve or dismiss anything. @mbradley @rabble, the one item I am pushing back on is the blocker, and the evidence is in the `oauth_client.dart:968` thread.

### Fixed

- **Announcements for `failure` and `expired`** (14038ad5). Both now get edge-triggered listeners alongside `unavailable`. A third test pins that the failure message is spoken once when unrelated state fields move.
- **A labeled way out** (431f3724). `_CloseButton` was a bare `GestureDetector` with no `Semantics` wrapper, so the only exit from the polling screen was an unlabeled tap target. It now announces as "Close", or "Start over" once verification has failed. This was a real gap and it is what made the expired copy name no control.

### Declined

- **`expired` is not dead code.** `/api/headless/resend-pin` answers an expired registration with 200 and `success: false`. keycast's own `test_resend_pin_on_an_expired_registration_reports_failure` has pinned exactly that since keycast#268. The 410 is real but belongs to `verify-pin`. Full evidence, including a live capture of both endpoints against one row, is in the `oauth_client.dart:968` thread.
- **The PIN-expired copy saying "Tap resend"** is not contradictory in practice. It renders when a PIN submission fails, at which point resend is idle and tapping it is the correct advice. If the registration is also expired, the resend row's own message supersedes it on the next tap. I would rather leave the copy alone than special-case a transient combination, but say the word if you disagree.

### Answered

- **Deployment parity.** There is one non-local keycast: `oauthConfig` returns `https://login.divine.video` for poc, staging, test and production alike. Both routes return 422 there, not 404, and the PIN ships in the registration email. Detail in the `email_verification_screen.dart:854` thread.

### Not requested, flagged separately

- A doc comment on `_errorCodeFromMessage` naming the keycast source of the string it matches, plus a test pinning the `declined` fallback (f893d229). Matching on prose is fragile and the next reader should not have to re-derive the contract.
- A test for the timed-out screen in token mode, which previously rendered nothing at all.

### Related

divinevideo/keycast#362 records a server-side issue found while verifying this: a cleanup task deletes expired pending rows every 300s, after which resend reports the uniform success and the client arms a cooldown for mail that never arrives. It narrows the production window for the `expired` branch. It is not a reason to remove the client handling, and the issue says so.

rabble (commented):
Re-reviewed at `431f37247`, 16/16 green.

**My blocker was wrong, and the decline is correct.** I asserted an expired pending registration is HTTP 410 and asked you to classify that instead of the success-false shape. It isn't, for this route: `/api/headless/resend-pin` answers 200 with `success: false`, pinned by keycast's own `test_resend_pin_on_an_expired_registration_reports_failure` since keycast#268. The 410 I was thinking of belongs to `verify-pin` (`pin_expired`) and to the `/api/auth/*` users-table routes. The `_errorCodeFromMessage` doc comment now spells out all three so the next reviewer doesn't repeat my mistake. Withdrawn — the `expired` branch is reachable and the test asserts the real response.

**Deployment parity: answered, and better than I expected.** One non-local keycast across poc/staging/test/production, both routes returning 422 rather than 404, PIN in the registration email. That's exactly the signoff I was asking for; removing the PIN-fallback flag is safe.

**Announcements: fixed**, and you found the real gap underneath my item. I asked for `failure` to be announced like `unavailable`; you added `expired` too, and then noticed `_CloseButton` was a bare `GestureDetector` with no `Semantics` wrapper — so the expired copy was telling people to use a control that announced as nothing. That's the more serious of the two, and it wasn't in either review.

**On the PIN-expired copy** — I'm satisfied with your reasoning and won't push it. The copy renders when a PIN submission fails, resend is idle at that moment, and "tap resend" is correct advice then; the resend row supersedes it on the next tap if the registration is also gone. Special-casing a transient combination costs more than it buys. Leave it.

Two things I want to flag rather than let pass silently, neither a request:

`_errorCodeFromMessage` matching on prose is load-bearing and fragile — if keycast rewords that sentence, this degrades to `declined`, which silently costs the user the "start again" guidance while leaving resend retryable. Your comment says exactly this and says to swap to a stable error code the moment the route grows one. Correctly handled, but it's the thing most likely to rot here, so it's worth someone owning keycast#362's follow-up actually adding a code.

keycast#362 itself is the more interesting find: a 300s cleanup task deletes expired pending rows, after which resend reports uniform success and the client arms a cooldown for mail that never arrives. Agreed that's not a reason to drop the client handling — it narrows the window, it doesn't close it — but it means the honest user-facing behaviour past that window is still "silent wait", which no amount of client work fixes.

**Nothing blocking from me.** My `CHANGES_REQUESTED` is stale and needs a maintainer to convert; @mbradley's is separate and still theirs to clear.

mbradley (approved):
Re-reviewed at `431f372`. Clearing my `CHANGES_REQUESTED`: the blocker is resolved and both questions are answered.

On the blocker, I had the same wrong endpoint that rabble did. Verified first-hand against keycast at `5068590`: `headless_resend_pin` answers an expired registration with HTTP 200 `{success:false, message:"This registration has expired. Please sign up again."}` (`headless.rs:1174-1178`), pinned by `test_resend_pin_on_an_expired_registration_reports_failure` asserting `StatusCode::OK`. The 410 `pin_expired` is `verify-pin`, a different route. The classifier matches that exact message to `expired`, so the branch is reachable and the test asserts the real contract. Withdrawn, and thanks to Meylis and dcadenas for pinning the source in the thread.

Parity: one non-local keycast, both routes 422, PIN in the registration email. That's the signoff I was asking for. Removing the flag is safe.

The screen-reader gap is fixed, and the `_CloseButton` `Semantics` fix underneath it is the more important one, so the expired copy now names a control that actually announces.

Thanks, as always.

NotThatKindOfDrLiz:
## Review Summary

Comprehensive review of the resend endpoint handling and signup flow fixes. Overall approach is solid, but three items need attention before merge.

### Findings

#### 🔴 1. Polling-Stopped Announcement Fires but Message Never Renders (Block)

In `email_verification_screen.dart`, the `BlocListener<EmailVerificationCubit, EmailVerificationState>(listenWhen: (prev, curr) => prev.resendStatus != curr.resendStatus && curr.resendStatus == ResendStatus.unavailable)` fires a `SemanticsService.sendAnnouncement('polling stopped')` **when resend fails**, but the screen only renders the polling-stopped explanation when `state.mode == ResendMode.token`.

**The disconnect:** A user might resend in pin-fallback mode (after PIN delivery succeeds), receive the unavailable error, hear the announcement, but see no UI explanation because the explanation only renders in token mode.

**Fix:** Either (a) render the polling-stopped message in both modes, or (b) fire the announcement only in token mode (`&& state.mode == ResendMode.token`). Which is correct depends on product intent—does the user need to know polling stopped in pin-fallback mode, or is it silent?

#### 🔴 2. Expired Registration Error Mis-Bucketed as Retryable (Block)

In `keycast_flutter/lib/src/oauth/oauth_client.dart`, the `_resendErrorFor()` method:
- Returns `ResendVerificationError.declined` for 410 (Gone)
- Returns `ResendVerificationError.server` for all other 4xx (400, 409, etc.)

A 410 is genuinely terminal (registration deleted/expired), but the code treats it as `declined` (user rejected), which is semantically wrong. The enum doesn't have an `expired` case, so this collides with user-initiated decline.

**Ask:** Should a 410 be a new `ResendVerificationError.expired` case, or should it stay as `declined`? If it stays as `declined`, the distinction between "user said no" and "your registration window closed" is lost at the UI layer.

#### 🟡 3. Open Question: Error Enum Completeness (Context)

`ResendVerificationError` has four cases: `unavailable`, `declined`, `server`, `network`. The mapping in `_resendErrorFor()` is:
- 404 → `unavailable` (not registered / not found)
- 410 → `declined` (Gone)
- Other 4xx → `server` (e.g., 409 Conflict, 422 Unprocessable Entity)
- 5xx → `server`
- No network → `network`

This feels reasonable, but **are there other keycast-specific 4xx codes that warrant their own case?** For instance, is 409 ever returned, and if so, what does it mean? The enum design locks in the UI's ability to distinguish error reasons later, so it's worth confirming the cases are complete.

### ✅ What's Working

- Synchronous double-submit guard in `DivineAuthCubit.submit()` prevents double-tap signup (verified with regression test)
- Feature flag removal is clean and complete across all surfaces
- BLoC state changes and tests are well-structured
- L10n keys queued for deferred translation are documented
- Keycast deployment verified (merged 2026-08-11, routes confirmed)

### Pre-Merge Checklist

- [ ] Rebase onto `origin/main` (currently BEHIND)
- [ ] Address findings #1 and #2
- [ ] Confirm finding #3 with keycast team or close as "no other codes in scope"

Issue #6112, opened 2026-07-15
fix(router): FATAL — context.pop() with nothing to pop (regressed) (Crashlytics 5c465b01)


rabble:
## Crashlytics — FATAL
- Issue: `5c465b01bb5421f294ecca1adaaf440a` ([console](https://console.firebase.google.com/v1/appid/project/openvine-co/crashlytics/app/1:972941478875:ios:f61272b3cf485df244b5fe/issues/5c465b01bb5421f294ecca1adaaf440a))
- Type: **FATAL** · **7 impacted users / 16 events** · REPETITIVE + REGRESSED (closed Mar 8, regressed 1.0.8)
- Versions: first 0.0.1 → last **1.0.16**

## Signature
`package:go_router/src/delegate.dart - GoRouterDelegate.pop`
`GoError: There is nothing to pop`

## Stack (sample event, 1.0.16/804)
```
GoRouterDelegate.pop (delegate.dart:105)
GoRouter.pop → GoRouterHelper.pop (extensions.dart:73)
_InkResponseState.handleTap  ← a button's onTap calls context.pop()
```
Breadcrumb: on the `hashtag` screen.

## Root cause (hypothesis)
A back/close button calls `context.pop()` when the GoRouter stack has nothing to pop — reached via deep link / notification into a leaf screen, or a double-tap popping twice. `main.dart:2212` already guards one pop with `if (router.canPop())`, but the crashing call site does not.

## Suggested fix
Guard the offending pop with `context.canPop()` (fall back to `context.go(home)` when false), or add a shared safe-pop helper and use it at back/close buttons. Repro: push a single-entry route and tap its back button. Low user count but fatal + regressed, so worth closing for good.

_Filed from Crashlytics batch triage._

PR #7087, opened 2026-08-11
refactor(settings): move FeatureFlagScreen and four siblings onto go_router

### LINKED ISSUE #6481: refactor(settings): move FeatureFlagScreen onto go_router

realmeylisdev:
## Description

Moves `FeatureFlagScreen` onto go_router, plus the other `Navigator.push(MaterialPageRoute)` sites that could reasonably follow it.

**Related Issue:** Closes #6481

While measuring the issue's premises I found that the obvious way to do this ships a crash, so the shape of the fix changed:

**A flat multi-segment `GoRoute` entered cold leaves a one-entry stack**, so `context.pop()` throws `GoError: There is nothing to pop` and the back button goes dead. go_router nests by route *tree*, not by path string, so `/settings/supporter` registered at the top level matches as a single entry — `canPop()` is `false`.

That is reachable, not theoretical: `swapAccount` rebuilds the router with `routerInitialLocationProvider` seeded from the current location, and `accountSwitchInitialLocation` preserves every non-`/profile` path verbatim. The trigger is `router.go()` on such a path — which is exactly the capability this issue exists to add.

I reproduced it on an iOS simulator against the already-shipped `/support-center/report-bug`: the back button does nothing, no navigation and no error UI.

<details>
<summary>Captured stack</summary>

```
flutter: [SYSTEM] Flutter Error: GoError: There is nothing to pop
══╡ EXCEPTION CAUGHT BY GESTURE ╞══
#0  GoRouterDelegate.pop (package:go_router/src/delegate.dart:105:7)
#1  GoRouter.pop (package:go_router/src/router.dart:586:20)
#2  GoRouterHelper.pop (package:go_router/src/misc/extensions.dart:73:65)
#4  _InkResponseState.handleTap (…/ink_well.dart:1224:21)
```
</details>

The fix is `context.safePop`, which this repo already added in #6264 to close **#6112** — a FATAL Crashlytics crash (7 users / 16 events) with the same signature, marked *regressed*. Four shipped flat screens still carry the same defect, so they are repaired here too.

### What changed

| # | Change |
|---|---|
| 1 | `FeatureFlagScreen` → child `GoRoute` under `/settings`, preserving `/settings/feature-flags` while giving cold entries a real Settings parent stack |
| 2 | Four shipped flat routes moved off raw `context.pop` → `safePop` |
| 3 | `_SignatureVerificationPolicyScreen` → public child route under `/nostr-settings` |
| 4 | Signature verification policy writes now go through a Cubit, and the policy enum moved out of `services/` |
| 5 | Two `CuratedListFeedScreen` call sites → the route that already existed |
| 6 | `_PostAnalyticsDetailScreen` → `/creator-analytics/:videoId` |
| 7 | Shrink-only ratchet freezing raw back-button pops so #6112 cannot regress a third time |
| 8 | Deleted a dead service and an orphaned ARB key |

Three of the issue's four stated justifications did not survive measurement, so the PR does not claim them:

- **"does not survive a router refresh"** — refuted. A pageless route and a `GoRoute` behave identically: both survive a same-location refresh, both die when a redirect changes location. An acceptance test for this would pass without the fix.
- **"cannot be restored"** — inert. There is no `restorationScopeId` anywhere in the app, so nothing restores today.
- **"never in the URL"** — true of all 16 tiles. `context.push` does not write the URL either; the 15 siblings are equally absent from it.

What actually holds is **consistency** and **in-app addressability** (`router.go(path)` now resolves). For Feature Flags, the route is nested for the best settings-detail UX: back is a real pop to Settings instead of a fallback navigation.

No `RouteType` is registered for these new detail paths, deliberately: `buildRoute()` maps the parent settings/analytics route back to the parent path, so teaching the normalizer about a detail path without also adding a distinct build target would bounce the user off the detail screen one frame after arrival.

## Out of Scope

- **Beyond the original ask:** `/support-center/report-bug` and `/support-center/request-feature` were found by the same structural sweep, not requested on the issue. Easy to drop if you'd rather they went separately.
- **External deep links.** `/settings/*` is absent from the Android manifest and from divine-web's AASA, so `https://divine.video/settings/...` still won't open the app. Deliberately not added — divine-web serves real `/settings/moderation`, `/settings/linked-accounts` and `/settings/relays` pages that the app would hijack.
- **`LibraryTrashScreen` and `showImageCropEditor`** stay imperative; both now carry a comment explaining why. The first needs a bloc created in `LibraryScreen.build`; the second takes bytes in and returns bytes out.
- **The other 36 files** with raw back-button pops are baselined, not fixed.

## Verification

- `flutter test test/blocs/signature_verification_policy/signature_verification_policy_cubit_test.dart test/features/feature_flags/screens/feature_flag_screen_route_test.dart test/router/flat_route_cold_entry_back_test.dart test/providers/nostr_service_identity_source_test.dart test/services/nostr_signature_verification_preference_service_test.dart` — green
- Pre-push changed-file test set — green
- `bash scripts/check_ui_service_boundary.sh` — green
- `bash scripts/check_unsafe_back_pop_ceiling.sh` — green
- `dart run build_runner build` — no generated diff
- `flutter gen-l10n` — no generated diff

The simulator pass caught one defect the tests missed: the not-found state reused *"No videos yet"*, which is false for a creator who does have videos. Fixed in `ef82941`.

## Type of Change

- [x] Code refactor
- [x] Chore


NotThatKindOfDrLiz:
@realmeylisdev I took over the review fixes and pushed `1e24aeb5d506bc91b3014e4d14a357b67b1c2c89`.

Best-UX routing decision: `/settings/feature-flags` is now a real child route under `/settings`, matching the signature-verification fix. Cold entry now gets a real Settings -> Feature Flags stack, so back pops to Settings instead of relying on a fallback navigation.

Other cleanup in the takeover:
- moved `NostrSignatureVerificationPolicy` out of the service layer and put the signature policy screen behind a Cubit so the UI/service boundary ratchet is green
- kept the feature flag app-bar on `safePop(fallback: /settings)` while preserving real-pop behavior through route nesting
- updated the feature flag route tests to use the real nested route behavior
- migrated the verify screens app-bar raw pops to `safePop`, which the new ratchet also caught
- updated the PR body verification notes

Local verification run from `mobile/`:
- `bash scripts/check_unsafe_back_pop_ceiling.sh`
- focused Flutter tests for the new Cubit, feature-flag routing, flat-route guard, Nostr identity source, and signature preference service
- `flutter gen-l10n`

GitHub checks are green on the pushed head. Please review the takeover diff and merge when ready.

NotThatKindOfDrLiz (approved):
Approved after takeover fixes. The best UX route shape is the nested `/settings/feature-flags` child route: it gives cold entries a real parent stack and keeps back behavior consistent with the corrected signature-verification route. CI is green on `1e24aeb5d506bc91b3014e4d14a357b67b1c2c89`; author should review the takeover diff before merge.

NotThatKindOfDrLiz:
Final takeover status after rebasing onto the latest `origin/main`: pushed `daf95682af907f8c5136290474a23c7617cd66e0`.

GitHub now reports the PR as clean, approved, and fully green on that head. The UX decision remains the nested `/settings/feature-flags` route under `/settings`, so cold-entry back behavior uses a real parent stack.

rabble (approved):
Reviewed the current daf95682a head with green CI. The nested settings/feature-flags route gives cold entries a real parent stack, the safe-pop fallbacks prevent flat routes from stranding deep links, and the signature-verification, curated-list, and post-analytics migrations preserve their state ownership while making navigation addressable. The unsafe-pop ratchet and focused routing tests cover the regression surface. No blocking findings.

realmeylisdev:
@NotThatKindOfDrLiz Reviewed the takeover commit as you asked. Two findings, neither blocking, plus one note on history.

**History note.** The branch head had become a `Merge branch 'main' into refactor/featureflag-gorouter-6481` commit (the "Update branch" button). `AGENTS.md` forbids merging `main` into a feature branch, so I replaced it with a rebase onto `4fc00d091`. The merge was a clean auto-merge — its tree is byte-identical to a fresh re-merge of both parents, and the rebased tree is byte-identical to the merge — so only the shape of the history changed. New head is `eb19a3183`; both approvals carried over.

**1. `NostrSignatureVerificationPreferenceService.defaultPolicy` is now unreferenced.** Moving the enum into `models/` was right for the boundary ratchet, but it replaced `fromStorage`'s `return NostrSignatureVerificationPreferenceService.defaultPolicy` with a literal `return all` — and that was the constant's only reader. The two agree today, so nothing changes for users; the default just lives in two places now with nothing keeping them in step. Smallest fix that keeps the dependency direction you established: declare the default on the enum (`static const defaultPolicy = all;`), have `fromStorage` return it, and point the service constant at it or drop it.

**2. `SignatureVerificationPolicyCubit.setPolicy` can throw twice if the screen is popped mid-write.** `onPolicyChanged` closes over the `ConsumerWidget`'s `ref` and is called inside the `try`:

- `ConsumerStatefulElement.invalidate` calls `_assertNotDisposed()`, which is a plain `if (!context.mounted) throw StateError(...)` rather than a debug-only assert — so it throws in release too (`flutter_riverpod` 3.3.2, `lib/src/core/consumer.dart:469`).
- That throw lands in the `catch`, which calls `emit(previous)` on a Cubit `BlocProvider` has already closed. `bloc` 9.2.1 throws `StateError('Cannot emit new states after calling close')` and rethrows it (`lib/src/bloc_base.dart:97`). The call site is `unawaited`, so this one escapes unhandled.

The window is just the `SharedPreferences` write, so reproducing it means tapping a policy and popping within a few milliseconds — narrow. The dead-`ref` half is also pre-existing: the old `_setPolicy(ref, …)` ran the same `ref.invalidate` after an `await` from an `unawaited` call. What the takeover adds is the second throw from `emit`-after-close.

Guarding the revert with `isClosed` stops the second throw, but not the first. The shape that removes both is to drop the `onPolicyChanged` callback and invalidate from a `BlocListener` in `_SignatureVerificationPolicyView` — the framework won't call a listener after unmount, and the Riverpod side effect moves back into the widget layer.

Since this PR already touches both files, I'd rather fix them here than merge and follow up. Do you want them in this PR, or would you prefer a separate one? Happy either way — and if you'd rather shape #2 yourself, it's your code, so say so and I'll leave it.

PR #7094, opened 2026-08-11
feat(library): file clips into categories from a filter chip row

### LINKED ISSUE #6973: feat(clips): archive clips
## Summary

Split from #4772.

Add an archive state for saved clips so users can hide clips from the main clip library without deleting them.

## Scope

- Add an `archived_at` marker or equivalent persisted archive state for local clips.
- Filter the active clip library to clips where `deleted_at IS NULL` and `archived_at IS NULL`.
- Add an archive surface where archived clips can be viewed.
- Support restore / unarchive.
- Keep archive separate from trash/delete retention and purge behavior.
- Preserve existing delete/trash behavior.

## Acceptance criteria

- Users can archive a clip from the clip library.
- Archived clips no longer appear in the main clip library.
- Users can view archived clips in a dedicated archive surface.
- Users can restore an archived clip to the main library.
- Tests cover active library filtering, archived library filtering, restore/unarchive, and trash-vs-archive precedence.

## Out of scope

Clip collections/categories are tracked separately and remain design-gated.

Supersedes the archive portion of #4772.

### LINKED ISSUE #6974: feat(clips): organize clips into collections
## Summary

Split from #4772.

Define and implement clip collections/categories so users can organize saved clips for future edits without overloading the archive/delete model.

## Product/design questions

- Can a clip belong to one collection or many collections?
- Are collections flat, grouped, or tag-like?
- Where do collections appear in the clip picker, clip library, and editor flow?
- Do collections sync beyond local storage, or are they device-local for v1?
- How do collections interact with archived and trashed clips?
- What empty/loading/search/filter states are expected?

## Acceptance criteria

- Product/design decisions above are resolved before implementation starts.
- The chosen data model supports the selected collection behavior without forcing a later migration immediately.
- Clip library and picker surfaces expose collections consistently.
- Tests cover collection assignment, filtering, archive/trash interaction, and any migration/backfill behavior.

## Out of scope

Simple clip archiving is tracked separately in #6973.

Supersedes the collection/category portion of #4772.

### LINKED ISSUE #7095: feat(library): file clips into categories from a filter chip row
Auto-created from PR #7094.

**PR:** https://github.com/divinevideo/divine-mobile/pull/7094
**Branch:** `feat/clip-library-categories`

hm21:
Closes #7095
Closes #6974
Closes #6973

## Description

The clip library only ever showed one flat list of clips. This adds categories: a horizontally scrollable chip row above the grid with **All**, **Archive** and **Deleted** built in, followed by however many categories the user creates.

**Archive hides clips from All** rather than just tagging them, so the default view stays the working set and archiving is a way to get something out of the way. **Deleted renders the trash bin inline** and replaces the separate `LibraryTrashScreen` — two routes to the same content is one too many; *Empty trash* moves into the toolbar next to the filter.

**A clip belongs to at most one category.** That mirrors the built-in filters, which already exclude each other, and it keeps assignment to a single tap instead of a multi-select. Deleting a category keeps its clips and drops them back into All — spelled out in the confirm sheet, because "delete category" reads like it might take the clips too.

Clips are filed from the existing selection mode (folder icon in the toolbar → sheet with *No category / your categories / New category / Archive*), so one flow covers one clip or twenty. Archiving offers an undo; it is the one action that makes clips disappear from the view you are looking at. Categories are renamed and deleted by long-pressing their chip, or via the pencil in the toolbar while that category is active — two routes because long-press alone is not discoverable.

The clip pickers (editor selection sheet, recorder) get **All plus categories only**: neither an archived nor a trashed clip can join a timeline, and managing categories is not what you came to the picker for. With no categories yet, the row hides entirely there rather than showing a lone "All" chip that filters nothing.

### Storage

New Drift table `clip_categories` plus `category_id` / `archived_at` on `clips`, `schemaVersion` 3 → 4. `SharedPreferences` is ruled out for this by `data_foundation.md` — it is a keyed collection with its own lifecycle, not a scalar preference. Categories are account-scoped exactly like clips, including sign-out cleanup and legacy-row claiming. Deleting a category unfiles its clips inside a transaction, so a category can never vanish while rows still point at it.

This work was originally cut as v3 and was renumbered to v4 on rebase: `main` shipped its own v3 in #7115 for the `identity_events` staleness stamps. The two migrations are independent, so v3 stays exactly as `main` wrote it and the clip-library work moved into a fresh `from < 4` step. Migration coverage is per-step rather than only end-to-end, because a device already upgraded to `main`'s v3 takes a different path into v4 than a v1 or v2 device does.

The renumber also exposed a gap in `_needsSchemaRepair()`, the guarded `beforeOpen` recovery probe: it checked the clip-library columns but not the `clip_categories` table or its indexes, and it ignored the identity-event stamps entirely. Those stamps arrived in v3 by altering a table that already existed, so normalization could only rebuild them when the whole table was missing — adding them to the probe alone would have fired repair on every startup while never fixing anything. The repair path gained two guarded column adds first, then the probe was widened to match. There is a regression test that drops both columns and asserts a reopen restores them.

The organisation columns live on the row rather than inside the clip's JSON blob, so the column stays the single source of truth. `ClipsDao.upsertClip` leaves them untouched, which matters because background asset recovery re-saves clips — there is a test pinning that.

### Composing with pinch-to-zoom (#7129)

`main` has since wrapped the clips grid in `PinchZoomGrid`. Both features stay: the chip row is a sibling above the grid, not a wrapper, so it does not sit in the zoom's way, and the zoom level lives in bloc state and survives a filter switch.

The one place they genuinely collided is the filter cross-fade. It holds the outgoing filter's grid on screen next to the incoming one, so for a few frames two scroll views exist. Where each grid brings its own controller that is harmless. Where a controller is *supplied* — only the picker sheets do — it would be attached to both at once, and `PinchZoomGrid` reads `controller.position` the moment a pinch starts, which asserts on anything other than exactly one attached position. A pinch landing inside that window would throw.

So the sheets that supply a controller switch filters outright and everything else keeps the cross-fade. Their surface is small and the fade earns little there. Three tests pin the split: the supplied-controller path never attaches twice, the standalone path still cross-fades, and the column count survives a filter switch.

### Included fixes

- **`DivineIconButton(size: small)` rendered 4px too tall for every type except `secondary`** (52 instead of the documented 48). The small variant already pins its outer size with an explicit 48px tap-target box, so the border compensation on top of it double-counted. In this feature it surfaced as the library header changing height when the toolbar's actions changed, which shifted the grid under it. Affects all small non-`secondary` icon buttons app-wide.
- The two profile-header badge tests had pinned the buggy `Size(52, 52)`. They now assert the 48dp tap-target floor they were really about, so the exact box stays the design system's call.

**Related Issue:** 

## Out of Scope

- Reordering categories — they append in creation order.
- Multi-category clips. Single-category was the deliberate choice above; a join table can replace `category_id` later without touching the UI contract.
- Undo for *move to category*, which would need the previous category per clip. Archive, the destructive-feeling one, has it.

## Verification

Re-run in full after the rebase onto `origin/main` (`098fa46036`, which carries #7129):

- `flutter test` in `packages/db_client` (949) and `packages/divine_ui` (875) — green; `divine_ui` line coverage 2106/2106 = 100.00% for the strict gate
- App suites: `test/l10n/arb_consistency_test.dart`, `test/blocs/clips_library/`, `test/models/clip_category_test.dart`, `test/screens/library_screen_test.dart`, `test/services/clip_library_service_test.dart`, `test/services/draft_storage_service_test.dart`, `test/services/file_cleanup_service_test.dart`, `test/widgets/library/`, `test/widgets/profile/profile_header_widget_test.dart` (492) — green
- Every app test asserting widget geometry re-run against the icon-button fix
- Drift migration tests cover v1 → v4, v2 → v4 and v3 → v4 as separate steps, including data preservation on both the clip rows and the identity-event rows `main`'s v3 stamps
- Schema repair covered at the current version: a damaged v4 database reopened *without* an upgrade, so the guarded `beforeOpen` probe is what restores the clip columns, `clip_categories`, and both category indexes — plus a `clip_categories` schema-parity test pinning Drift's definition against the repair SQL
- Generated artifacts regenerated, not hand-merged: `build_runner build --delete-conflicting-outputs` and `drift_dev make-migrations` produced `drift_schema_v4.json`, `schema_v4.dart` and the v4 step in `app_database.steps.dart`; `flutter gen-l10n` and a second `build_runner` pass report no further drift
- Ratchets run locally and green: untested-services floor, test-unit structure, UI/service boundary, dependency provenance, raw dialog / textstyle / colors / material-button / raw-icons ceilings, file-size, future-delayed, todos, skip, riverpod boundary, arb-error, changenotifier boundary
- All 27 new keys translated into all 21 non-English locales; `arb_consistency_test` green

**verified on a real Samsung Galaxy device** (SM-S942B), re-run on this rebased build after #7129 landed: the clips grid still pinches between 2 and 5 columns, the chip row cross-fades between filters, a pinch straight after a filter switch behaves, and the picker sheet's instant filter switch reads as intended.

One unrelated flake seen during this rebase is tracked in #7179 (`nip05_settings_navigation_test`, introduced by #7127, order-dependent under randomized CI ordering).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


NotThatKindOfDrLiz:
@hm21 I pushed maintainer follow-up commit 41733954d to the PR branch.

What it covers:
- fixes the empty active-filter crash when a selected clip is hidden by the current chip filter, and keeps hidden selections from enabling toolbar/create-video actions
- preserves category/archive state across autosave row rewrites and autosave-only library-row creation
- keeps the v3 schema repair path current without changing Drift normal migration schema shape
- removes the now-orphaned trash l10n keys and regenerated localizations

Local verification on the takeover commit:
- pre-push gate passed: analyzer, generated-file verification, and changed-file tests
- focused reruns passed for `library_screen_test.dart`, `clip_library_service_test.dart`, `packages/db_client/test/src/database/app_database_test.dart`, and `packages/db_client/test/drift/app_database/migration_test.dart`

Please review the maintainer commit before merge. This PR still needs the Samsung Galaxy/manual UI verification and screenshots or recording called out in the body.

NotThatKindOfDrLiz (changes requested):
I pushed maintainer follow-up commit 41733954d for the code findings from this review: the hidden-selection empty-filter crash, autosave organization persistence, the v3 repair path, and stale trash l10n keys are now covered with regression tests and passed local verification.

Requesting changes remains appropriate before merge because the PR still needs author/manual closeout: please review the maintainer commit and add the Samsung Galaxy verification plus screenshots or a recording for the library category UI. Once that is posted and CI is green, this should be ready for final review.

rabble (commented):
Code review is complete at f75becd4a and all GitHub checks are green. Fixed in takeover: category-name truncation counted UTF-16 code units and could split an emoji at the 40-character boundary; the branch now truncates by grapheme cluster and has three focused regression tests. Local full analysis, dependency-provenance, generated-file, locale, service-floor, and 533 changed-file test gates passed before push. Escalated rather than silently skipped: the existing changes-requested review still requires author/manual Samsung Galaxy verification plus screenshots or a recording of the category UI. I cannot truthfully satisfy that device-evidence gate from this session, so I am not overriding it with an approval. Declined items: none.

hm21:
@rabble @NotThatKindOfDrLiz Below is the required screen recording of the feature, as requested.
Let me know if you’d prefer me to also link the Discord conversations about this feature. In that case, I can quickly look them up and add the links as well.

https://github.com/user-attachments/assets/705bbd3e-890c-4dc4-bc07-75a488c3a3ae

rabble (commented):
Second review, at `f75becd4a`, 17/17 green. No blockers from me; @NotThatKindOfDrLiz's request is the one that governs.

**First, a correction to how this PR reads at a glance.** GitHub shows +26,105 / −634 across 95 files, which makes it look like one of the largest changes open. It isn't. **72% of the additions are Drift generated artifacts:**

| File | Added |
|---|---|
| `test/drift/app_database/generated/schema_v3.dart` | 13,941 |
| `drift_schemas/app_database/drift_schema_v3.json` | 3,350 |
| `app_database.g.dart` | 803 |
| `app_database.steps.dart` | 650 |
| **generated subtotal** | **18,744** |

The hand-written surface is roughly 7,400 lines, and the largest single authored file is `clip_category_actions.dart` at 384. Worth saying out loud so the next reviewer sizes the reading correctly rather than bouncing off the header number — and it's the expected shape for a schema bump, not padding.

**The schema change follows the rules rather than the shortcut.** `data_foundation.md` requires real Drift migrations, not `beforeOpen` repair SQL accumulating `CREATE TABLE IF NOT EXISTS`. This bumps `schemaVersion` to 3, does the work in `onUpgrade`, commits the v3 snapshot, and the migration test covers **v2 → v3 and v1 → v3 separately** plus a `v3 schema is valid and up to date` assertion. The v1 path matters most here: per the `db_client` note in `data_foundation.md`, two installs can both report `user_version = 1` with different tables on disk because the historic repair block lives in `beforeOpen` and `onUpgrade` runs first. Having an explicit legacy-v1 test rather than assuming the declared snapshot is what's actually out there is the right call, and the guarded `beforeOpen` path is kept with a comment saying v1 is normalized by `onUpgrade`.

The data assertions are real too — `expect(migrated?.categoryId, null)` on a pre-existing row pins that the migration doesn't invent category membership for clips that predate the feature, which is the failure mode that would be invisible until someone opened the filter row.

**On the outstanding request.** hm21 has posted the screen recording. The Samsung Galaxy verification DrLiz asked for isn't something I can confirm from the recording, and reviewing the maintainer follow-up commit `41733954d` is the author's to close out — that's process, not code, and it stays with the two of you.

**Nothing blocking from me.** I did not re-derive the four fixes in the maintainer commit (hidden-selection empty-filter crash, autosave organization persistence, v3 repair path, orphaned trash l10n keys); DrLiz wrote and locally verified those, and author measurements aren't review — that pass needs to come from hm21, not from me confirming DrLiz's own work.

mobile/packages/db_client/lib/src/database/app_database.dart
```diff
@@ -149,6 +155,51 @@ class AppDatabase extends _$AppDatabase {
...
+  /// [_migrateToV3ClipCategories] so `onCreate` — which builds the tables
+  /// through `createAll` rather than the migration steps — gets them too.
+  Future<void> _createClipCategoryIndexes() async {
+    await customStatement('''
+      CREATE INDEX IF NOT EXISTS idx_clip_category_id
+      ON clips (category_id)
+    ''');
+    await customStatement('''
+      CREATE INDEX IF NOT EXISTS idx_clip_category_owner_pubkey
+      ON clip_categories (owner_pubkey)
```
NotThatKindOfDrLiz, line 245: This can no longer stay `schemaVersion => 3`. Current `origin/main` already uses v3 for the `identity_events.source_created_at` / `source_event_id` migration, so this PR needs to rebase and move the clip category/archive schema work to v4.

mobile/packages/db_client/test/drift/app_database/migration_test.dart
```diff
@@ -17,22 +16,50 @@ void main() {
...
+      expect(AppDatabase(NativeDatabase.memory()).schemaVersion, 3);
     });
 
-    test('v2 schema is valid and up to date', () async {
+    test('v3 schema is valid and up to date', () async {
+      final schema = await verifier.schemaAt(3);
+      final db = AppDatabase(schema.newConnection());
+      await verifier.migrateAndValidate(db, 3);
+      await db.close();
+    });
```
NotThatKindOfDrLiz, line 37: After rebasing, this should assert the current schema is v4 and add explicit v3 -> v4 coverage. The existing v2 -> v3 identity-events migration test from main also needs to remain; otherwise devices already upgraded to main's v3 will skip the clip-category migration.

NotThatKindOfDrLiz (changes requested):
Requesting changes because the branch no longer has a valid merge/migration story against current main.

Current main has already shipped schema v3 for `identity_events.source_created_at` / `source_event_id`, while this PR also claims schema v3 for clip categories and archive markers. GitHub reports the PR as `CONFLICTING` / `DIRTY`; a throwaway merge confirms the only conflicts are the Drift migration surface. This needs to be rebased and recut as schema v4, not textually resolved as another v3.

Required before re-review:
- rebase onto current `origin/main`
- bump `schemaVersion` to 4
- move the clip category/archive changes into a v3 -> v4 migration step
- regenerate the v4 Drift snapshots
- add v3 -> v4 migration coverage while preserving main's identity-events v2 -> v3 coverage
- reconcile `_needsSchemaRepair()` with both the identity-event columns from main and the clip-library columns from this PR
- update the PR body verification/storage notes after the maintainer commits and current rebase

I am not pushing a takeover fix for this because the repository takeover rules only allow low-risk, clear remediations. Recutting database version history after a schema collision requires author-level migration judgment.

hm21:
Rebased onto `734700255b`. 19/19 checks green, GitHub now reports `MERGEABLE`.

**Fixed**

- Rebased onto current `origin/main`.
- `schemaVersion` → 4. `main`'s v3 `identity_events` step is untouched; the clip-library work moved into a fresh `from < 4` step.
- v4 snapshots regenerated with `drift_dev make-migrations` rather than hand-merged: `drift_schema_v4.json`, `schema_v4.dart`, and the v4 step in `app_database.steps.dart`.
- Migration coverage is now per-step — v1 → v4, v2 → v4 and v3 → v4 separately, plus a v3-clips data-preservation test alongside the existing v2 one, so a device already on `main`'s v3 is covered on its own path into v4.
- `_needsSchemaRepair()` reconciled. This was wider than the column list: the probe was also missing the `clip_categories` table and both category indexes. The identity-event stamps could not just be added to it — they arrived in v3 by altering a table that already existed, so normalization only rebuilt them when the whole table was missing, and probing for them alone would have fired repair on every startup while never fixing anything. The repair path gained two guarded column adds first, then the probe was widened to match. Regression test drops both columns and asserts a reopen restores them; confirmed red without the fix.
- PR body storage and verification notes rewritten for v4.

**One adaptation worth flagging**

The identity-events test could not stay `v2 → v3`. `testWithDataIntegrity` validates the tested database against the `newVersion` snapshot *after* it migrates to its own `schemaVersion`, so pinning `newVersion: 3` against a v4 database fails with `Schema does not match` — confirmed empirically, not assumed. It now runs `v2 → v4` with the same assertions, so the v2 → v3 step is still exercised, just no longer terminated at v3.

mobile/lib/widgets/library/clips_tab.dart
```diff
@@ -50,91 +57,211 @@ class ClipsTab extends StatelessWidget {
...
+  final double gridTopPadding;
+
+  /// How long one filter's content takes to fade into the next.
+  static const _filterFadeDuration = Duration(milliseconds: 180);
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocBuilder<ClipsLibraryBloc, ClipsLibraryState>(
+      builder: (context, state) {
+        return AnimatedSwitcher(
```
NotThatKindOfDrLiz, line 198: This filter switch remounts the grid content by design. That now needs to be reconciled with #7129 before this PR can merge: current `origin/main` wraps this grid in `PinchZoomGrid`, whose invariant is that wrappers above the grid stay structurally constant so scroll position and thumbnail resolution are not thrown away. Please rebase onto current main and make an explicit integration choice for this chip row/filter transition path, including the new required `ClipsTab.backgroundColor`.

mobile/packages/db_client/test/src/database/app_database_test.dart
```diff
@@ -1210,6 +1225,77 @@ void main() {
...
+                  recorded_at,
+                  data,
+                  file_path,
+                  thumbnail_path,
+                  owner_pubkey,
+                  deleted_at
+                FROM clips_old
+              ''')
+              ..execute('DROP TABLE clips_old;')
+              ..execute('PRAGMA user_version = 3;');
```
NotThatKindOfDrLiz: With `schemaVersion` now at 4, setting `user_version = 3` means this test exercises the v3->v4 `onUpgrade` path, not the guarded `beforeOpen` schema-repair path. The test name says damaged current-schema repair, but `_needsSchemaRepair()` is skipped when `details.hadUpgrade` is true. Please add/retarget coverage that damages a v4 database and reopens it without an upgrade, mirroring the identity-events recovery test, and include the clip columns, `clip_categories`, and both category indexes.

mobile/packages/db_client/lib/src/database/app_database.dart
```diff
@@ -163,6 +172,50 @@ class AppDatabase extends _$AppDatabase {
...
+    if (!clipColumns.contains('category_id')) {
+      await m.addColumn(clips, clips.categoryId);
+    }
+    if (!clipColumns.contains('archived_at')) {
+      await m.addColumn(clips, clips.archivedAt);
+    }
+    await _createClipCategoryIndexes();
+  }
+
+  Future<void> _createClipCategoriesTableIfMissing() async {
```
NotThatKindOfDrLiz, line 222: This raw SQL path now defines `clip_categories` separately from Drift's fresh-install/migration definition. The file already pins this exact agreement for the other runtime-created tables with schema-parity tests; please add the same parity test for `clip_categories` so column order, nullability, defaults, primary key, and indexes stay locked between the two paths.

NotThatKindOfDrLiz (changes requested):
Thanks for the v4 recut. The schema collision itself is resolved properly, but this still needs another author pass before merge. GitHub currently reports this branch as conflicting with current main, and the conflict is with #7129's pinch-to-zoom clips grid. That integration needs a deliberate decision about how the category chip row and filter transitions compose with PinchZoomGrid's scroll/thumbnail stability assumptions. Please also add the missing v4 beforeOpen repair coverage for the clip-library schema pieces and the clip_categories schema-parity test, then re-run device verification after the rebase.

hm21:
Rebased onto `098fa46036`, which carries #7129. 19/19 green, GitHub reports `MERGEABLE`.

**Chip row + pinch-to-zoom**

Both stay. The chip row is a sibling above the grid rather than a wrapper, so it never enters `PinchZoomGrid`'s subtree, and the column count lives in bloc state and survives the remount a filter switch performs.

They collide in exactly one place, which I reproduced rather than reasoned about: the cross-fade holds the outgoing filter's grid on screen next to the incoming one, so for a few frames two scroll views exist. Where each grid brings its own controller that is harmless. Where one is *supplied* — only the picker sheets do — it is attached to both at once, and `PinchZoomGrid._stopScroll` reads `controller.position` the moment a pinch starts, which asserts on anything but exactly one attached position. An instrumented run confirmed `grids=2 views=2 positions=2` before the fix and `positions=1` after.

The explicit choice: **the sheets that supply a controller switch filters outright, everything else keeps the cross-fade.** Their surface is small and the fade earns little there. Four tests pin it — the supplied-controller path never double-attaches, the standalone path still cross-fades, the column count survives a switch, and the grid still pinches *after* a switch.

**Two of main's pinch tests needed adapting.** `the grid at the persisted column count` and `pinching the grid open changes the column count` seed `clips:`, but this PR moved the grid's source to `state.sortedClips` (which carries sort *and* filter). Both now seed `sortedClips` too; no assertion changed.

**Schema repair**

The damaged-schema test pinned `user_version = 3`, which since the v4 recut is an *upgrade* — `hadUpgrade` short-circuits `_needsSchemaRepair()`, so the migration was doing the work the test credited to repair. It now damages a v4 database and reopens at the same version, and covers `clip_categories` and both category indexes alongside the two clip columns, then writes through the DAO to prove the rebuilt table works.

`clip_categories` also gained the schema-parity test the other runtime-created tables have. Verified it fails when the repair SQL is perturbed (`DEFAULT 0` → `DEFAULT 1`), so it is really comparing the two definitions.

PR body updated for all of the above.

**One red herring, now tracked separately: #7179.** Shard 2 failed once on `test/router/nip05_settings_navigation_test.dart`, which is not in this diff — this PR does not touch routing at all. It is seed-dependent: identical commit, seed `1105091160` fails and `1129702722` passes. It arrived in #7127, whose own CI shard 2 exited 69 on a sqlite3 download before running any test, so it has never had a green run. Not patched here, since that would put another PR's fix in an unrelated diff.

Rebased once more after posting: `main` gained #7171 and #6858 while this was in flight. The only conflict was one line in `scripts/baseline/ui_service_imports.txt` — #6858 drops `liked_videos_screen_router.dart`, this PR drops `library_trash_screen.dart`, and both removals apply to a shrink-only ratchet. Everything above was re-verified on the new base.

**Still yours:** Samsung Galaxy re-verification after this rebase, and review of maintainer commit `41733954d`.

hm21:
Device verification is done — the item this review was still waiting on.

Re-ran on a real Samsung Galaxy (SM-S942B) against this rebased build (`542727f79e`, on `098fa46036`, which carries #7129): the clips grid still pinches between 2 and 5 columns, the chip row cross-fades between filters, a pinch straight after a filter switch behaves, and the picker sheet's instant filter switch — the deliberate exception described in the previous comment — reads as intended rather than as a dropped animation. No regressions.

The screen recording of the category UI is in https://github.com/divinevideo/divine-mobile/pull/7094#issuecomment-5263206715. Worth flagging that it was captured before the #7129 rebase, so it shows the category flows themselves and not their composition with pinch-to-zoom; the device pass above is what covers that part.

Still open on my side: reviewing maintainer commit `41733954d`.

NotThatKindOfDrLiz (approved):
Re-reviewed after the recut and takeover cleanup. The prior blockers are resolved: schema is correctly on v4, v3->v4 and older-origin migrations are covered, the beforeOpen repair path is tested against a damaged v4 database, clip_categories has schema-parity coverage, and the #7129 PinchZoomGrid interaction is handled/tested.

Takeover cleanup pushed:
- a798400afe fixes the remaining cleanup items from review: removes unused category APIs, adds l10n descriptions, and corrects the setClipCategory doc.
- 130200e1fd handles package-analyzer fallout in divine_ui after the rebase. The db_client analyzer issue is now covered by main’s equivalent fix.

Verification:
- local pre-push gate passed: analyzer, generated files, changed-file tests (609 pass, 1 skipped)
- DB Client CI, Divine UI CI, Mobile CI, integration services, preview build, mergeability, semantic PR, generated files, format, analyze, and all test shards are green after rerunning the two shards that initially failed on the sqlite native-asset download.

Filed follow-up #7194 for the non-blocking hidden-selection bulk-action affordance. @hm21 please review the takeover commits before merge.

rabble:
Rebased this branch onto fresh `origin/main` (was 14 commits behind, `mergeStateStatus: DIRTY`). All 14 of your commits are preserved and reattributed; **CI is green** and the existing approval survived the force-push. I did not merge — that stays yours.

### Conflicts resolved

All three were in `db_client`, all collisions between your v4 work and the follower-count changes main landed in #6911.

1. **`app_database.dart` — migration helpers.** main added `_repairSchemaV3()` / `_backfillFollowerCountTimestamps()`; you added `_migrateToV4ClipCategories()` / `_createClipCategoryIndexes()`. Purely additive on both sides, so I kept the union. `beforeOpen` now calls `_normalizeLegacyV1Schema()` then `_repairSchemaV3()`, and `onUpgrade` keeps `from < 3` and `from < 4` as separate steps.
2. **`app_database.dart` — `_needsSchemaRepair()` probe map.** main added `profile_statistics`, you added `identity_events`. Kept both. They stay coherent because each has a matching repair path: `profile_statistics` via `_repairSchemaV3()`, `identity_events` via the guarded column adds your "reconcile the schema repair probe" commit put into normalization — so neither entry can fire repair forever without fixing anything.
3. **`migration_test.dart` — test rename.** main added a v2→v3 follower-timestamp test and kept `'…survive the v3 upgrade unstamped'`; you renamed that test to drop the version because it now migrates to 4. Kept main's new test and took your rename.

### Two things the rebase broke that no conflict marker showed

Flagging these separately since they are **changes beyond conflict resolution**:

- **`fix(db): regenerate the v4 schema snapshot against main's v3 column`** — main's `profile_statistics.follower_counts_updated_at` landed in v3 *after* you generated the v4 snapshot, so `drift_schema_v4.json` and `generated/schema_v4.dart` no longer described the live tables. `drift_dev make-migrations` refused to run at all, and the schema-validation tests were comparing against a v4 missing the column. Regenerated with `dart run drift_dev make-migrations`. **v4 is introduced by this branch and has never shipped, so re-exporting its snapshot rewrites nothing users can already have on disk.** In the same commit, main's follower-timestamp test validated at version 3, which is now an intermediate step — opening the DB migrates to 4, so it failed on your clip columns. Pointed it at 4 to match the sibling identity-events test; the backfill under test still runs in the `from < 3` branch of the 2→4 upgrade.
- **`fix(l10n): interpolate the count in the new clip-library plural arms`** — main's #7153 added a guard against hardcoded numbers in ICU plural arms, and your four new snackbar keys tripped it. gen-l10n compiles ARB `=1` down to intl's `one:`, so the arm fires for the whole CLDR `one` category, and am/fil/fr/pt select it at counts other than 1 — it would have rendered a false "1". Interpolated `{count}` in the 16 affected arms (4 keys × 4 locales), the same structural fix #7153 applied to the other 115. **Wording is untouched, so no translator review is needed;** en and the other 17 locales fire `one` only at n=1 and are unchanged.

### One thing I want to be straight about

The first CI run after the rebase also failed `Tests (shard 1/4)` on `fullscreen_feed_bloc_test.dart` → *"subscribes to removedIdsStream and dispatches removals"*. That file is not in this PR's diff, main was green at its own HEAD, and this branch was green pre-rebase. It passes locally on this branch and passed on the very next CI run with no change to that code. It is a real wall-clock race — the test schedules two 1 ms delays and asserts after 50 ms — so it is load-sensitive, and adding this branch's new test files to that shard's merged bundle perturbed it. **I did not touch it**, since it belongs to main and fixing it here would be scope creep, but it is worth a tracking issue: that 50 ms budget will bite again.

Verification: `flutter analyze lib test integration_test` clean; full `db_client` suite (959 tests) green; `divine_ui` green with coverage; the library/clips/l10n/profile suites this PR touches green; all repo ratchet scripts pass. Everything ran on the mise-pinned Flutter 3.44.0.

NotThatKindOfDrLiz (changes requested):
## Review Summary

This PR refactors clip organization with Drift v2→v3 migration, category filtering, and archive/restore functionality. The architecture is sound and aligns with project patterns. However, there is **one blocking crash** that breaks ordinary user interactions, plus four supporting issues that should be fixed before merge.

---

## 🚨 Finding 1: BLOCKING — StateError on Empty Filter (sortedClips.first)

**Location:** `mobile/lib/screens/library_screen.dart` lines 570–581

**Repro:**
1. Enter selection mode
2. Select any clip
3. Switch to a filter with no matching clips (Archive with no archived clips, or empty category)
4. Crash: `StateError: No element`

**Root Cause:**
The `_visibleClips` getter filters clips by category and archive status. When selection mode is on, the selection persists across filter changes. If the active filter produces an empty list, `sortedClips.first` in the `orElse` guard throws because the list is empty.

```dart
// Line 574-579 — vulnerable
targetAspectRatio:
    (sortedClips.isNotEmpty
        ? sortedClips.first.videoAspectRatio
        : orElse: () => 16 / 9) ?? 16 / 9, // Should guard before .first
```

The guard checks `isNotEmpty` but then calls `.first` in a ternary — if sortedClips is empty, the fallback in `orElse` should be used. But the code structure accesses `.first` before the guard completes.

**Fix:**
Add a null check or guard to prevent `.first` when sortedClips is empty:
```dart
targetAspectRatio:
    (sortedClips.isNotEmpty ? sortedClips.first.videoAspectRatio : null) ?? 16 / 9,
```

Or more defensive:
```dart
targetAspectRatio: sortedClips.isEmpty ? 16 / 9 : (sortedClips.first.videoAspectRatio ?? 16 / 9),
```

**Impact:** High — crashes on ordinary filter navigation in selection mode. Blocks merge.

---

## 📋 Finding 2: Test Coverage Gap

**Location:** `mobile/lib/screens/library/library_screen_test.dart`

The screen gained archive/restore snackbar, empty-trash confirmation dialog, manage-category toolbar action, and size calculations based on filtered clips, but test additions are minimal (3 lines for `getCategories` mock).

**Recommend:**
- Test empty-trash confirmation (swipe + confirm happy path, cancel path)
- Test archive/restore snackbar appearance and undo action
- Test sortedClips-based size calculations (empty list → fallback ratio used)
- Test category filter changes with active selection (verify selection persists across filters)

**Impact:** Medium — low coverage blocks review, but not a crash.

---

## 🔄 Finding 3: Autosave Clip Organization Stale After Editor Save

**Location:** `mobile/lib/services/clip_library_service.dart` line 407–419 (`_writeToBothClipRows`)

When a clip is organized (categorized or archived), the code writes category_id and archived_at to both the library row AND the autosave-draft row via `_writeToBothClipRows`. However, the next time the editor saves via `DraftsDao.saveDraftWithClips`, it deletes all rows for that draftId and reinserts from the current draft state — losing the organization data.

```dart
// _writeToBothClipRows writes to autosave row
await _draftRowsDao.update(draftId, clipId, category_id: categoryId, ...);

// But next editor save obliterates it:
// DraftsDao.saveDraftWithClips → delete + reinsert (no organization columns)
```

Same issue affects asset recovery and import re-save paths.

**Impact:** Medium — organization changes to unsaved clips are lost if the user returns to the editor. Undermines the feature's UX.

**Suggested fix:**
- Preserve organization state when reinserting draft rows in `saveDraftWithClips`
- Or: store organization only in the library row, not the draft row
- Or: load organization from library when displaying draft clips

---

## 📍 Finding 4: Orphaned L10n Keys

**Location:** All 22 `app_*.arb` files

Two keys no longer have call sites after the trash screen was folded into the Deleted filter:
- `libraryOpenTrashSemanticLabel`
- `libraryTrashTitle`

These remain in all ARB files and generated code, cluttering the l10n surface.

**Fix:**
Remove both keys from all 22 locales, then run `flutter gen-l10n`.

**Impact:** Low — code smell, not a crash. But should be cleaned up.

---

## ✓ Finding 5: Schema Repair Path Doesn't Recover Organization

**Location:** `mobile/packages/db_client/lib/src/database/app_database.dart` (v1→v3 repair)

The `beforeOpen` repair for damaged v2 databases recreates the clips table without category_id and archived_at columns. Clips recovered from a damaged v2 database won't have organization data.

This is not a blocker because:
1. v2→v3 migrations include data preservation tests (passing)
2. Damaged v2 databases are rare edge-case recoveries
3. Organization is new; pre-existing clips have no organization anyway

**Note:** If damaged v2 recovery becomes a support issue in production, add a follow-up to backfill organization from the repair path.

---

## ✅ Non-Issues Verified

- **divine_ui fix correct:** Border compensation excluded for small variant (line 422) is correct — 48px tap-target box is already sufficient
- **Profile tests updated:** Assert 48dp constraint floor as expected
- **upsertClip logic:** Uses `ClipsCompanion.insert`, which leaves organization columns `Value.absent()` — correct for one-time imports
- **v1→v3 migrations:** Include data preservation; tests pass
- **Delete/archive transactions:** Proper rollback on failure

---

## Verdict

**Status:** Changes requested (blocking crash)

**Must fix before merge:**
1. Sortedclips guard in library_screen.dart (Finding 1) — **blocks**
2. Add screen-level test coverage (Finding 2) — recommended
3. Preserve autosave organization or document expected behavior (Finding 3)
4. Remove orphaned l10n keys (Finding 4) — cleanup

**Good to fix after merge (optional):**
- Finding 5 repair path enhancement (low priority, edge case)

Please address Finding 1 and any test gaps, then request re-review.

PR #7097, opened 2026-08-11
fix(lists): refresh discover list subscriptions

### LINKED ISSUE #7098: fix(lists): refresh discover list subscriptions
## Problem
A user reported that Discover Lists can stay unavailable for days. Logs showed Discover Lists subscribing to public kind 30005 lists but receiving no parsed list events or UI stream updates.

## Impact
Users who hit the stale/no-event stream path can see an indefinite loading state and cannot recover by revisiting the screen in the same app session.

## Acceptance Criteria
- Re-entering Discover Lists with the same public-list filter sends a fresh anonymous Nostr subscription instead of attaching to a stale broadcast stream.
- Anonymous subscriptions clean up when their listener is canceled.
- Discover Lists stops loading and shows a retryable error if the relay stream does not emit.
- Regression tests cover repeated anonymous subscriptions and the Discover Lists no-event timeout.

## PR
Implemented in #7097.

NotThatKindOfDrLiz:
Closes #7098

## Summary
- create fresh anonymous Nostr subscriptions so repeated query-style screens do not attach to stale broadcast streams
- clean up relay subscriptions when the last listener cancels
- add a Discover Lists timeout/error path so no-event streams stop showing an indefinite loader

## Verification
- flutter test test/src/nostr_client_test.dart (from mobile/packages/nostr_client)
- flutter test test/screens/discover_lists_screen_test.dart (from mobile)
- flutter analyze lib/screens/discover_lists_screen.dart test/screens/discover_lists_screen_test.dart packages/nostr_client/lib/src/nostr_client.dart packages/nostr_client/test/src/nostr_client_test.dart

### DIFF
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -2908,6 +2908,7 @@
     },
     "discoverListsLoading": "የህዝብ ዝርዝሮችን በመፈለግ ላይ...",
     "discoverListsRelayTimeout": "ሪሌው በጊዜው ዝርዝሮችን አልመለሰም። እንደገና ሞክር።",
+    "discoverListsServiceUnavailable": "አገልግሎቱ አይገኝም።",
     "discoverListsEmptyTitle": "ምንም የህዝብ ዝርዝሮች አልተገኙም",
     "discoverListsEmptySubtitle": "ለአዳዲስ ዝርዝሮች ቆይተው ይመልከቱ",
     "discoverListsByAuthorPrefix": "በ",
--- a/mobile/lib/l10n/app_am.arb
+++ b/mobile/lib/l10n/app_am.arb
@@ -2907,6 +2907,7 @@
         }
     },
     "discoverListsLoading": "የህዝብ ዝርዝሮችን በመፈለግ ላይ...",
+    "discoverListsRelayTimeout": "ሪሌው በጊዜው ዝርዝሮችን አልመለሰም። እንደገና ሞክር።",
     "discoverListsEmptyTitle": "ምንም የህዝብ ዝርዝሮች አልተገኙም",
     "discoverListsEmptySubtitle": "ለአዳዲስ ዝርዝሮች ቆይተው ይመልከቱ",
     "discoverListsByAuthorPrefix": "በ",
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -3740,6 +3740,7 @@
     },
     "discoverListsLoading": "جاري اكتشاف القوائم العامة...",
     "discoverListsRelayTimeout": "لم يُرجع الريلاي القوائم في الوقت المناسب. حاول مرة أخرى.",
+    "discoverListsServiceUnavailable": "الخدمة غير متاحة.",
     "discoverListsEmptyTitle": "لم يتم العثور على قوائم عامة",
     "discoverListsEmptySubtitle": "عاود التحقق لاحقًا لرؤية قوائم جديدة",
     "discoverListsByAuthorPrefix": "بقلم",
--- a/mobile/lib/l10n/app_ar.arb
+++ b/mobile/lib/l10n/app_ar.arb
@@ -3739,6 +3739,7 @@
         }
     },
     "discoverListsLoading": "جاري اكتشاف القوائم العامة...",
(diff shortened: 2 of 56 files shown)


hm21 (approved):
Approved. Diagnosis is right — dedup handed a re-entering screen a broadcast controller that had already emitted, so the new listener got nothing. Pushed three fixes, all checks green on `1d995c9`.

**Pushed**

- `1d995c9` — the 8s timeout was armed on every `_streamPublicLists` call, including pull-to-refresh. `_buildBody` checks `_errorMessage` before it renders the list, so a relay answering slowly during a refresh replaced a fully populated screen with the error view. Now armed only when the screen has nothing to lose. The added regression test fails without the guard (checked).
- `089eb8c` — the timeout copy was an English literal spliced into `discoverListsFailedToLoadWithError`, so all 21 non-English locales rendered half-translated. Added `discoverListsRelayTimeout` to all 22 ARB files and regenerated. Test assertions now resolve from `AppLocalizations` rather than matching English literals.
- `2e68b15` — the class and library docs still advertised "subscription deduplication".

**One unrequested change** — in `089eb8c` the timeout message no longer nests inside `discoverListsFailedToLoadWithError`. Wrapped, it rendered "Failed to load lists: The relay did not return lists in time. Try again." directly under the error view's own "Failed to load lists" heading. Drop that bit if you'd rather keep the wrapper.

**Not changed, worth your call** — removing dedup for anonymous `subscribe()` is a global change: ~35 anonymous call sites across the app and 10 packages, each now opening its own relay REQ. `queryEvents` is already pooled behind `maxConcurrentQueries = 6` with a comment citing relays' "too many concurrent REQs" limit; `subscribe()` has no such pool, and dedup was the only thing collapsing identical filters. I audited the call sites and found no concrete breakage — nothing depends on the old deterministic `sub_<hash>` id, and the listen-then-cancel sites actively benefit from the new `onCancel` — so this reads as a ceiling risk rather than a bug, which is why I left it alone. If you want the same fix with a smaller blast radius, reusing a cached controller only while `controller.hasListener` is true kills the stale-stream bug and still shares genuinely concurrent identical subscriptions.

**Validation** — `flutter analyze lib test integration_test`, the full `nostr_client` package suite, `discover_lists_screen_test.dart`, and `arb_consistency_test.dart`, plus all CI on the new head.

Merge decision is yours — happy to take another pass if you change the subscription strategy.

NotThatKindOfDrLiz:
Follow-up pushed in `772d4212b`.

What changed:
- closed the real leak sites that anonymous dedup had been containing: `SocialService.getUserVideoCount`, `CuratedListService.fetchUserListsFromRelays`, `CuratedListService.fetchPublicListsContainingVideo`, and the adjacent connected-relay contact-list query in `FollowRepository`
- kept unique anonymous subscription IDs deliberate for late-joiner correctness, and documented that returned anonymous streams are closed after the last listener cancels; callers should call `subscribe()` again instead of retaining/re-listening
- removed write-only `_subscriptionFilters`
- localized the Discover Lists service-unavailable copy
- added a refresh-only stream timeout that clears refresh/loading state and cancels the dead relay subscription without replacing an existing list with the error view
- filed #7102 for the larger live `subscribe()` concurrency-pool work

Verification:
- `flutter gen-l10n`
- `dart format` on touched Dart files
- `flutter test test/src/nostr_client_test.dart` from `mobile/packages/nostr_client`
- `flutter test test/src/follow_repository_test.dart` from `mobile/packages/follow_repository`
- `flutter test test/screens/discover_lists_screen_test.dart test/services/curated_list_service_query_test.dart test/unit/curated_list_relay_sync_test.dart` from `mobile`
- `flutter analyze lib test` from `mobile`, `mobile/packages/nostr_client`, and `mobile/packages/follow_repository`
- push hook reran analyzer and changed-file tests successfully before push

@hm21 please re-review the follow-up commit. The ceiling risk was concrete: the three abandoned live subscriptions were `mobile/lib/services/social_service.dart` `getUserVideoCount`, `mobile/lib/services/curated_list_service.dart` `fetchUserListsFromRelays`, and `fetchPublicListsContainingVideo`.

NotThatKindOfDrLiz:
Amended follow-up is now `92cc8b9b1`.

The previous GitHub failure was the Follow Repository CI coverage gate: tests passed, but the new manual stream error branch was uncovered. I added `handles connected relay query stream error gracefully`, reran `very_good test --coverage --min-coverage 100` from `mobile/packages/follow_repository`, and force-pushed with lease.

@hm21 please review `92cc8b9b1` rather than the superseded `772d4212b`.

